PR #426, opened 2019-01-29
Minor updates to LV (Tank friendly, denser jungle)

Jimminian95:
Adds TONNES of foliage to the LV map, tries to complete a jungle feel by outlining the map in vines and overpasses, as well as making the jungle walls even denser and closer to the fences, as opposed to effectively being tank trap tiles scattered in a field.
Upgrades fences to have a wall every two tiles.
Adds blast doors in nexus to complement the Nexus lockdown button in sec.
Adds mini tablefort to nexus central.
Converts all nexus doors to all access.
Creates several more paths for the tank to take, including the ability to leave the damn LZs without needing the dropship to takeoff, as well as the ability to drive through cargo, nexus main entrances, around hydro and through east LZ2.


Alterist:
While I would call this anything but minor, its ready for test merge, mainly for balance

PR #436, opened 2019-01-29
Fixes preferences being unchangeable

### LINKED ISSUE #377: Game Preferences while not in Lobby
Oy i was just playing and whenever im not in the lobby screen I cant change any of my preferences when i go to OOC then Game Preferences and when i click anything it doesn't read it. I don't know if this is a bug or feature but i think its a bug in my opinion hope to fix soon so that when i die i can change my prefs and not be in lobby just to do it.

DominikPanic:
:cl: LaKiller8
fix: You can now modify preferences at any time
/:cl:
Fixes #377

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -803,7 +803,7 @@ datum/preferences
 	return 1
 
 /datum/preferences/proc/process_link(mob/user, list/href_list)
-	if(!ismob(user))
+	if(!istype(user))
 		return
 
 	switch(href_list["preference"])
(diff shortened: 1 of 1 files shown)


code/modules/client/preferences.dm
```diff
@@ -803,7 +803,7 @@ datum/preferences
 	return 1
 
 /datum/preferences/proc/process_link(mob/user, list/href_list)
-	if(!isnewplayer(user) && !isobserver(user))
+	if(!ismob(user))
```
Rohesie:
```suggestion
	if(!istype(user))
```

PR #437, opened 2019-01-29
Prevents synthetics from being able to rig and throw rigged power cells.

DominikPanic:
:cl: LaKiller8
fix: Synths can no longer use power cells like grenades.
/:cl:


Rohesie:
Ghommie made a `dangerous` tag for grenades the synths already check for. Can it be applied here?

Rohesie:
Damn snowflakes. Oh well, to fix this we'll have to wait for components.

PR #439, opened 2019-01-29
Pulled larva (and xenos) can now get up.

### LINKED ISSUE #317: Pulled larva can't get up
When pulled larva cant stand up Rest state to normal state

DominikPanic:
:cl: LaKiller8
fix: You can now get up while being pulled as a xeno.
/:cl:
Fixes #317


Ghommie:
Just make this a callback for now for livings, and check `if(is_mob_incapacitated(TRUE))` at the end for the "fail to get up" and "get up" messages.
On a side note, we can port the /tg/'s version of `do_mob()` and `do_after()` without too much hindrance now.

Rohesie:
That would be very positive.

code/__HELPERS/unsorted.dm
```diff
@@ -951,17 +951,18 @@ var/global/image/busy_indicator_hostile
...
-	for(var/i = 0 to numticks)
-		sleep(delayfraction)
+	var/endtime = world.time + delay
+	while(world.time < endtime)
+		stoplag(1)
 		if(!user || user.loc != original_loc || get_turf(user) != original_turf || user.stat || user.knocked_down || user.stunned)
 			. = FALSE
 			break
-		if(L && L.health < CONFIG_GET(number/health_threshold_crit))
+		if(L.health && L.health < CONFIG_GET(number/health_threshold_crit))
```
Rohesie:
```suggestion
		if(L?.health && L.health < CONFIG_GET(number/health_threshold_crit))
```

Rohesie:
This is a bit hacky of a solution. This will bypass the `FALSE` return with the flag, so the earlier it's broken, such as swapping hands, the better. And it complexifies a very used proc for it.
A `CALLBACK` would achieve the same, with less cost and less unintended behavior.
Failing that, a custom proc.

Rohesie:
This issue is not exclusive to xenos, so you could extend the proc rework to everyone.

If you want to preserve the icon, which would be positive, you can add `user.overlays += get_busy_icon(BUSY_ICON_GENERIC)` right before the `CALLBACK`, and `user.overlays -= get_busy_icon(BUSY_ICON_GENERIC)` at the end of it.

PR #440, opened 2019-01-29
You can now destroy much more with the tank.

### LINKED ISSUE #262: Crushers / Tanks cant crush Operating Table cant crush crates
And many other crushable items.

A-lexa:
:cl: LaKiller8
balance: Tank can now crush more things (disposals, vending machines, operating tables, airlocks)
/:cl:
Fixes #262

### DIFF
--- a/code/game/machinery/doors/shutters.dm
+++ b/code/game/machinery/doors/shutters.dm
@@ -3,6 +3,7 @@
 	icon = 'icons/obj/doors/rapid_pdoor.dmi'
 	icon_state = "shutter1"
 	power_channel = ENVIRON
+	destructible = FALSE
 
 /obj/machinery/door/poddoor/shutters/New()
 	..()


code/modules/vehicles/cm_armored.dm
```diff
@@ -408,6 +417,10 @@ var/list/TANK_HARDPOINT_OFFSETS = list(
 		var/obj/structure/table/T = A
 		T.visible_message("<span class='danger'>[root] crushes [T]!</span>")
 		T.destroy_structure(TRUE)
+	else if(istype(A, /obj/structure/closet))
```
Rohesie: If you are willing, you could also do a generic behavior for structures along the lines of
`if(breakable) then C.destroy_structure(TRUE)`

Else this should be ready to merge.

Rohesie:
Structures need a `breakable` check. I suspect for machinery we'll end up running into the same issue, such as telecomms stuff, which shouldn't be crushed.

DominikPanic:
> 
> 
> Structures need a `breakable` check. I suspect for machinery we'll end up running into the same issue, such as telecomms stuff, which shouldn't be crushed.

Done.

Issue #227, opened 2018-12-22
Surgery NEVER causes infection


DominikPanic:
Yes, another bug I introduced. Oh well, writing shit here helps me keep track of it.

PR #441, opened 2019-01-29
Fixes surgery infection chances

DominikPanic:
:cl: LaKiller8
fix: Surgery can now cause infection again.
/:cl:
A coder fixing his own bugs? Impossible!
#227

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -76,9 +76,9 @@ proc/spread_germs_to_organ(datum/limb/E, mob/living/carbon/human/user)
 	//Gloves
 	if(user.gloves)
 		if(user.germ_level && istype(user.gloves, /obj/item/clothing/gloves/latex))
-			E.germ_level += user.gloves.germ_level * 0.5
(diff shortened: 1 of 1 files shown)


code/modules/surgery/surgery.dm
```diff
@@ -76,27 +76,29 @@ proc/spread_germs_to_organ(datum/limb/E, mob/living/carbon/human/user)
 	//Gloves
 	if(user.gloves)
 		if(user.germ_level && istype(user.gloves, /obj/item/clothing/gloves/latex))
-			E.germ_level -= user.gloves.germ_level * 0.5
+			E.germ_level += user.gloves.germ_level * 0.5
```
Rohesie: The behavior is identical to the `else`, making it moot. What's the logic here? Should latex gloves pass a smaller fraction of the germs than other gloves? Or pass no germs at all?

DominikPanic: The way the infection calculations work is that you have to reduce it like this, but yeah fair point for the numbers, I've tweaked them.

PR #443, opened 2019-01-29
changes the first-aid pouch to have a green cross

Hughgent:
[why]: # the first aid pouch doesn't have a cross on it. this gives it a green one which is different than all the medic ones, which are red.


Hughgent:
from: 
image
to:
image
which is based on:
image

Ghommie:
You should make the green outline more marked, I can barely make it out without pushing my face on the monitor. All good otherwise.

Hughgent:
@Ghommie Have you considered increasing the brightness of your monitor? Otherwise when I increase the brightness of the greens it tends to blend with the white of the cross too much.

I feel the important part is the white cross which matches the red variants.

PR #446, opened 2019-01-30
Yet another admin tweaks PR.

DominikPanic:
So many small issues to hunt, it never ends.


code/game/gamemodes/gameticker.dm
```diff
@@ -7,7 +7,7 @@ var/global/datum/controller/gameticker/ticker
 
 
 /datum/controller/gameticker
-	var/const/restart_timeout = 600
+	var/const/restart_timeout = 1200
```
Rohesie:
```suggestion
	var/const/restart_timeout = 120 SECONDS
```
Or 2 MINUTES

DominikPanic:
Fixes #490

PR #453, opened 2019-01-31
Fixes xeno inability at dropship barrel rolls

### LINKED ISSUE #456: Queen cannot hijack the shuttle
## Testmerges:

#426 #443 #442 #440 #439 #438 #437 #435 #434 #433 #432 #429 #422 #419 #368 #367 #446 #451 #448 #219

## Reproduction:

The exact AHelp about it was `You don't have the dexterity to do this! <getting this while I'm trying to use the alamo console.`

Ghommie:
:cl:
fix: Fixes xeno queens being unable to maneuver the dropship consoles. add that up to my mistakes.
/:cl:

fixes #456


code/game/machinery/machinery.dm
```diff
@@ -242,7 +242,7 @@ Class Procs:
 		return 1
 	if(usr.is_mob_restrained() || usr.lying || usr.stat)
 		return 1
-	if (!ishuman(usr))
+	if (!usr.IsAdvancedToolUser() && !isxeno(usr))
```
Rohesie:
```suggestion
	if (!usr.IsAdvancedToolUser())
```
The check should be `isxenoqueen`, but restricted to the dropship.

Ghommie: Xenos aren't advanced tool users, as far as I know. Should I change it?

Ghommie: The old version was basically `if(!ishuman(usr) && !ismonkey(usr) && !issilicon(usr) && !isxeno(usr))`.
computers usually don't return `attack_hand` on `attack_alien`, with the exception of these dropship ones.

PR #459, opened 2019-01-31
Zoom fixes

### LINKED ISSUE #337: RPG fires when trying to point to something
## Reproduction:

try to point to someone or something and watch it shoot the RPG (shift middlemousebutton)

DominikPanic:
:cl: LaKiller8
fix: You can now point while zoomed in.
/:cl:
Fixes #337

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -654,6 +654,8 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 		zoom = !zoom
 		if(user.interactee)
 			user.unset_interaction()
+		else if(!istype(src, /obj/item/attachable/scope))
+			user.set_interaction(src)
 		return
 
 	//General reset in case anything goes wrong, the view will always reset to default unless zooming in.
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -654,8 +654,6 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 		zoom = !zoom
 		if(user.interactee)
 			user.unset_interaction()
-		else
-			user.set_interaction(src)
 		return
 
 	//General reset in case anything goes wrong, the view will always reset to default unless zooming in.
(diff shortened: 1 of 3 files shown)


code/game/objects/items.dm
```diff
@@ -654,7 +654,7 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 		zoom = !zoom
 		if(user.interactee)
 			user.unset_interaction()
-		else
+		else if(!istype(src, /obj/item/attachable/scope))
```
Ghommie: Make this `istype(src, /obj/item/device/binoculars/tactical)` or add the lines to a child version of this proc, since it's the only zoom item that makes use of handle_click().

code/modules/mob/mob_verbs.dm
```diff
@@ -19,37 +19,36 @@
 		next_move = world.time + 2
 	return
 
-/mob/verb/point_to(atom/A in view())
+/mob/verb/point_to(atom/A in view(client.view + client.get_offset(), loc))
```
Ghommie: That allows people to receive "pointed at" messages of far flung things they shouldn't see or arrow points from seemengly nowhere, no?

DominikPanic: If their offset is 3 in one direction, it will allow them to point 3 further in all directions, sadly there's no better way to do this.

code/game/objects/items.dm
DominikPanic: This is more of a future proofing, I don't expect someone making a new zoomable item with special interaction adding it here.

Ghommie: That's also true.

code/modules/mob/mob_verbs.dm
Ghommie: Still potentially unhelpful and may also work through cameras. The `get_offset()` proc is good though.

DominikPanic: Client.view is just the amount of tiles a client sees from their mob.

Ghommie: Umh, thanks for cleaing it out.

Rohesie:
This is some pretty nice debugging work. Kudos.

PR #380, opened 2019-01-26
Minor refactor on supply drops code.

Ghommie:
:cl:
refactor: Minor refactor on supply drops.
balance: supply pads can now load up to four unanchored objects per launch instead of a single one.
their potential debris factor has been ever so slighty reduced.
/:cl:

Supply drops code was expectedly bad and limited. This can go after #366 so I get the time to make sure it's ready.

PR #474, opened 2019-02-01
Fixing supply drops

Ghommie:
:cl:
fix: fixes supply drops.
/:cl:

Tested, it works as intended now. Fixes #380. Plus minor clean up.


Rohesie:
Is this tested locally?

Ghommie:
Yea, jumped through CiC, requisition and colony several times.

PR #478, opened 2019-02-01
Deletes changerock

Alterist:
why is this even a thing?


DominikPanic:
I mean it was used for an event by AIIA a few times.

Rohesie:
Isn't this a /tg/ item that was ported for an event?

spookydonut:
@Rohesie this is legacy from when cm forked from bay

PR #480, opened 2019-02-01
Adds more screams

Spyroshark:
:cl:
soundadd: Added two new screams and two new pain sounds for humans.
/:cl:

All of the sounds are from Half Life 1.
_Why do we all have to wear these ridiculous ties?_


Rohesie:
I'm not really convinced by the meme set of screams we've gathered.
The TF spy one is specially bad.

DominikPanic:
Agreed, this is too memey.

Spyroshark:
i mean we already had a wilhelm scream from CM

Spyroshark:
I'm going to make a new PR, because 1: rohsie said he'd be ok with the half life screams as long as they're made more rare and 2: because drawsstuff has sent me a LARGE collection of screams and xeno sounds from the same game our current ones are ripped from.

PR #485, opened 2019-02-01
Removes a couple nazi references from the code

Ghommie:
:cl:
spellcheck: Removes a couple references to the national socialistic party, as per guidelines.
remove: Removes an unused miscellanous jacket due to missing sprites.
/:cl:


Alterist:
If the great coat didn't have a worn sprite, then delete it completely

Ghommie:
Ok, great coat removed as it seems baycode didn't have the worn sprite either.

Alterist:
despite the name "Fury Fuhrer" being there, we re-enabled the random enemy name generation long ago, so it would never show up. either way, this is good to go

PR #489, opened 2019-02-02
changes allowed armor storage items for marine armor

py01:
In practice, marine's suit storage is almost exclusively used to store their gun, but random crap like shotgun shells auto-equips into the suit storage.  This PR limits marine armor to not store items that you would never want to put in there, due to having much better places to store them.


DominikPanic:
Tbh I'd rather see this solved from the opposing end (autoequip only equipping certain things to certain places) than removing the ability to manually store things in the suit storage.

Issue #486, opened 2019-02-01
Facehuggers can hide under certain barricades


py01:
Currently, facehuggers can hide beneath certain barricades like plasteel when facing south, making them impossible to see or click on. It feels cheap to get hit by one of these huggers.

Rohesie:
Just a matter of adjusting their layer.

PR #503, opened 2019-02-02
Huggers render above more things

py01:
Huggers thrown onto a barricade now render above the barricade, instead of being  hidden. This has the side effect of making huggers always render above humans and xenos.

Closes:#486

### DIFF
--- a/code/__DEFINES/layers.dm
+++ b/code/__DEFINES/layers.dm
@@ -80,7 +80,7 @@
 
 //#define MOB_LAYER 4
 
-#define FACEHUGGER_LAYER 4.1
+#define FACEHUGGER_LAYER 4.05
 
 #define ABOVE_MOB_LAYER 4.1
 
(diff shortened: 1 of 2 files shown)


Ghommie:
This is as far as the facehugger layer increasing should go honestly, going way will create other visual issues.
This should be fixed by shifting the facehugger sprite a couple pixels up, animating its tail and/or squashing the plasteel barricade height down a little. I'm not pretending you to do so though.

DominikPanic:
Agreed with Ghommie here.

spookydonut:
Make a new HUGGER_LAYER so you can put it in whatever order you want without zlevelfighting other things in ABOVE_MOB_LAYER like huggers currently do in MOB_LAYER

DominikPanic:
Right now it's the exact same since both layers are the same number.

code/__DEFINES/layers.dm
```diff
@@ -80,6 +80,8 @@
 
 //#define MOB_LAYER 4
 
+#define FACEHUGGER_LAYER 4.1
```
spookydonut:
```suggestion
#define FACEHUGGER_LAYER 4.05
```

PR #506, opened 2019-02-02
new marine and xeno sounds

Spyroshark:
:cl: New xenomorph and marine sounds
soundadd: Adds more sounds for xeno/larva roars, xeno growls, and human pain/scream noises.
tweak: The TF2 spy scream has a lower chance of playing, just like the wilhelm scream and the new joke sounds.
/:cl:

All of the new sounds (except the new joke ones) are ripped from AVP 2010, which is also where almost all of the old CM sounds are from. The wilhelm scream and spy scream are both renamed, along with the four half-life scientist screams/pain noises. Each of the joke sounds should have a 5% chance to play.


Spyroshark:
thanks to drawsstuff for ripping the sounds

PR #512, opened 2019-02-03
Nerfed effective range of dropship turrets

Surrealaser:
Per title


DominikPanic:
Since it's balance you should changelog it

PR #528, opened 2019-02-03
E to equip, now with being able to set your preferred slots!

DominikPanic:
:cl: LaKiller8
add: You can now press E to both equip and unequip, along with being able to set your preferred slot in the preferences tab!
/:cl:


Rohesie:
This is looking good. Poke me when you feel there's no more tweakings necessary. Until then we'll keep testing it.

code/game/objects/items.dm
```diff
@@ -462,43 +462,60 @@ cases. Override_icon_state should be a list.*/
 				if (!H.back || !istype(H.back, /obj/item/storage/backpack))
 					return FALSE
 				var/obj/item/storage/backpack/B = H.back
-				if(src.w_class <= B.max_w_class && B.can_be_inserted(src))
+				if(src.w_class <= B.max_w_class && B.can_be_inserted(src, warning))
 					return TRUE
 			if(SLOT_IN_B_HOLSTER)
 				if (H.back && istype(H.back, /obj/item/storage/large_holster))
```
spookydonut: early return this

code/game/objects/items.dm
```diff
@@ -462,43 +462,60 @@ cases. Override_icon_state should be a list.*/
...
 					var/obj/item/storage/S = H.back
-					if(S.can_be_inserted(src))
+					if(S.can_be_inserted(src, warning))
 						return TRUE
 				return FALSE
 			if(SLOT_IN_HOLSTER)
 				if((H.belt && istype(H.belt,/obj/item/storage/large_holster)) || (H.belt && istype(H.belt,/obj/item/storage/belt/gun)))
```
spookydonut: early return

code/game/objects/items/storage/storage.dm
```diff
@@ -328,65 +328,65 @@
...
+			if(istype(W, text2path(A)))
+				ok = TRUE
 				break
 		if(!ok)
-			if(!stop_messages)
-				if (istype(W, /obj/item/tool/hand_labeler))
-					return 0
+			if(warning)
+				if(istype(W, /obj/item/tool/hand_labeler))
+					return FALSE
```
spookydonut: if i'm correct that this stops a hand labeller being put in a bag, then just make it return FALSE if the hand labeller is turned on.

Borisvanmemes:
Being able to turn off E equiping weapons would be nice, there are times where i really don't want it to do that.

code/game/objects/items.dm
```diff
@@ -462,43 +462,62 @@ cases. Override_icon_state should be a list.*/
...
+			if(SLOT_IN_SUIT)
+				var/obj/item/clothing/suit/storage/S = H.wear_suit
+				if(!istype(S) || !S.pockets)
+					return FALSE
+				var/obj/item/storage/internal/T = S.pockets
+				if(T.can_be_inserted(src, warning))
+					return TRUE
+			if(SLOT_IN_ACCESSORY)
+				var/obj/item/clothing/under/U = H.w_uniform
+				if(!U || !U.hastie)
```
Rohesie:
```suggestion
				if(!U?.hastie)
```

PR #529, opened 2019-02-03
Probable Fix for Corrosive Acid Effects Shifting Around

Surrealaser:
Per title


Rohesie:
Is this tested to work?

Surrealaser:
@Rohesie I haven't seen any issues thus far but who knows if there's some unresolved corner case.

PR #535, opened 2019-02-04
Replaces a_intent text strings with defines.

Ghommie:
:cl:
code: Replaced a_intent text strings with defines.
/:cl:


code/modules/mob/mob_helpers.dm
```diff
@@ -294,20 +294,27 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
...
-var/list/intents = list("help","disarm","grab","hurt")
 /proc/intent_numeric(argument)
 	if(istext(argument))
 		switch(argument)
-			if("help")		return 0
-			if("disarm")	return 1
-			if("grab")		return 2
-			else			return 3
+			if(INTENT_HELP)
+				return 0
```
DominikPanic:
```suggestion
				return FALSE
```
If you're cleaning up don't half-ass it xP

code/modules/mob/mob_helpers.dm
```diff
@@ -294,20 +294,27 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
...
+				return 2
+			else
+				return 3
 	else
 		switch(argument)
-			if(0)			return "help"
-			if(1)			return "disarm"
-			if(2)			return "grab"
-			else			return "hurt"
+			if(0)
```
DominikPanic: Magic numbers.

PR #545, opened 2019-02-04
Briefing chairs can now only be deconstructed by welding.

DominikPanic:
:cl: LaKiller8
tweak: Briefing chairs can now only be deconstructed by welding them.
/:cl:


DominikPanic:
> 
> 
> I'd suggest having a different type of chair for briefing to allow for props planetside to be scrapped easily, as they can be.

There already is though - chair/reinforced. No other chairs were touched.

PR #556, opened 2019-02-05
is to is a

Hughgent:
[why]: 

Reeeeeee
grammer
to
1grammer


Hughgent:
the staffwho verb in the admin tab.

DominikPanic:
You will see `is a Admin`

Hughgent:
image

Rohesie:
Honestly something like `Name - Rank - Playing as` would look better than "is" or "is a". There will always be different cases, such as "is a mentor" and "is an admin".

DominikPanic:
Agreed with Roh here.

Hughgent:
easy enough.

Hughgent:
image

PR #562, opened 2019-02-06
Ongoing facehugger fixes.

Ghommie:
:cl:
fix: Ongoing facehugger fixes.
/:cl:

Fixes more edge cases and omissions such as synths getting knocked out, facehuggers possibly dying while attached, potential runtimes, inability to place facehuggers back into eggs, throwing/leaping vars not resetting.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
@@ -171,6 +171,8 @@
 /obj/item/clothing/mask/facehugger/proc/check_lifecycle()
 	if(sterile)
 		return TRUE
+	if(check_neighbours())
+		return FALSE
 	if(lifecycle - 4 SECONDS <= 0)
 		if(isturf(loc))
 			var/obj/effect/alien/egg/E = locate() in loc
--- a/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
@@ -48,6 +48,8 @@
 			STOP_PROCESSING(SSobj, src)
 
 /obj/item/clothing/mask/facehugger/process()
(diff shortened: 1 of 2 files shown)


Rohesie:
Is there a need to changelog this?

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -433,22 +442,19 @@
 	if(ishuman(user))
 		playsound(loc, (user.gender == MALE ?'sound/misc/facehugged_male.ogg' : 'sound/misc/facehugged_female.ogg') , 25, 0)
 	if(!sterile)
-		if(!issynth(user) || !isIPC(user)) //synthetics aren't paralyzed
+		if(!issynth(user) && !isIPC(user)) //synthetics aren't paralyzed
```
DominikPanic: Why the and here?

code/modules/mob/living/carbon/xenomorph/Facehuggers.dm
```diff
@@ -192,19 +195,19 @@
...
-	var/obj/item/clothing/mask/facehugger/F
-	var/count = 0
-	for(F in get_turf(src))
-		if(F.stat == CONSCIOUS && !F.sterile)
-			count++
-		if(count > 2) //Was 5, our rules got much tighter
-			visible_message("<span class='xenowarning'>The facehugger is furiously cannibalized by the nearby horde of other ones!</span>")
-			qdel(src)
-			return TRUE
+	if(isturf(loc))
```
spookydonut:
```suggestion
	if(isturf(loc))
```
early return

Ghommie: It is still won't grant a TRUE return value even if the location happen to be a turf, plus it's not too deeply indented, hence why I didn't do the early return.

PR #563, opened 2019-02-06
Gives the ship a Mast

Hughgent:
:cl: Hughgent & Spyroshark & Arachnidnexus
add: The Ship now has a true connection to earth.
/:cl:

[why]: 
> Small Superstitions, they never leave us no matter how far we evolve.
> 
> Sailors are no exception. In fact, you could say they’re some of the most superstitious. When all you’ve got between you and a Lonely, deadly void is a thin sheet of steel, you tend to make up a story that protects the mind.
> 
> One of these you can see, right now if you so choose, is the mast of the ship. Silly as it sounds for a ship in space, but it does have a mast. It’s thick Oak or Redwood, real wood, never fake. If it were fake the magic wouldn’t be there. It’s our link with earth and every other sailing ship. 
> 
> The constraints of physics and engineering means that it doesn’t support anything except itself, and the spirit of the crew. You can find ours in cryo. Time and many, many hands have worn it a bit. 
> 
> Ours was even part of a real sailing ship! She was called the HMS Victory. Floated and fought for over a hundred years according to the plaque.


Hughgent:
image
new spot in Briefing

code/game/objects/structures/misc.dm
```diff
@@ -49,16 +49,25 @@
...
+	icon = 'icons/obj/structures/structures.dmi'
+	icon_state = "shipmast" //Thank you to Spyroshark and Arachnidnexus
+	desc = "A piece of old earth that was. The plaque reads<br><br><span class='name'>HMS Victory Sailed 1765 to 1922.</span><br><span class='name'>Relaunched 2393.</span><br><span class='name'>On loan from the First Sea Lord.</span><br><br>"
+	layer = ABOVE_MOB_LAYER
+	density = TRUE
+	anchored = TRUE
+
+/obj/structure/shipmast/attack_hand(var/mob/user)
+	. = ..()
+	if(user.a_intent == INTENT_HELP)
```
Rohesie: Do a `switch` here.
Usage:
```
switch(user.a_intent)
	if(INTENT_HELP)
		visible_message("[usr] rubs the [src] for good luck.")
	if(INTENT_DISARM)
		...
```
And so on.

Hughgent:
Everything checks out on travis's end, but it doesn't compile on my end. says it's missing `is_special_character()` because it's an undefined proc.

Hughgent:
> 
> 
> Everything checks out on travis's end, but it doesn't compile on my end. says it's missing `is_special_character()` because it's an undefined proc.

turns out a file got unchecked.  BYOND!

PR #568, opened 2019-02-06
Briefing Huboxing 2.0

DominikPanic:
:cl: LaKiller8
tweak: Briefing chairs can no longer be rotated at lightning speeds, all briefing objects are much tougher to destroy now.
/:cl:


code/game/machinery/doors/windowdoor.dm
```diff
@@ -315,3 +324,36 @@
 	dir = SOUTH
 	icon_state = "rightsecure"
 	base_state = "rightsecure"
+
+
+/obj/machinery/door/window/briefing/northleft
+	dir = NORTH
```
spookydonut: no.

code/game/machinery/doors/windowdoor.dm
```diff
@@ -315,3 +324,36 @@
...
+	icon_state = "rightsecure"
+	base_state = "rightsecure"
+
+/obj/machinery/door/window/briefing/eastright
+	dir = EAST
+	icon_state = "rightsecure"
+	base_state = "rightsecure"
+
+/obj/machinery/door/window/briefing/westright
+	dir = WEST
```
spookydonut: very no

spookydonut:
this is a hamfisted and poorly thought out attempt to stamp out something that has been arbitrarily ruled LRP.  I am against this change.

code/game/machinery/doors/windowdoor.dm
```diff
@@ -252,6 +252,15 @@
 	health = 300.0 //Stronger doors for prison (regular window door health is 200)
 
 
+/obj/machinery/door/window/briefing
+	name = "reinforced safety windoor"
+	icon = 'icons/obj/doors/windoor.dmi'
+	icon_state = "leftsecure"
+	base_state = "leftsecure"
+	req_access = list(ACCESS_MARINE_BRIDGE)
+	health = 3000
```
Ghommie: this seems excessive, stick it to 1k or so if possible. Also give this a different description to hint the sturdiness.

code/game/objects/structures/window.dm
```diff
@@ -411,6 +411,16 @@
 	health = 300
 	reinf = TRUE
 
+
+/obj/structure/window/reinforced/briefing
+	name = "reinforced safety glass"
+	desc = "A very tough looking glass window with a special rod matrice, probably bullet proof."
+	icon_state = "rwindow"
+	basestate = "rwindow"
+	health = 3000
```
Ghommie: given how directional should have half the health of a full tile one and requisitions windows have 2k, I think 1k should be enough.

code/game/machinery/doors/windowdoor.dm
```diff
@@ -252,6 +252,19 @@
...
+	name = "reinforced safety windoor"
+	desc = "An extremely strong windoor, capable of stopping the fiercest of riots."
+	icon_state = "leftsecure"
+	base_state = "leftsecure"
+	req_access = list(ACCESS_MARINE_BRIDGE)
+	health = 1000
+
+
+/obj/machinery/door/window/briefing/get_projectile_hit_chance(obj/item/projectile/P)
+	return TRUE
```
Ghommie: this should serve no purpose as windoors don't account for energy projectiles (yet), if I recall correctly.

code/game/objects/structures/window.dm
```diff
@@ -411,6 +411,20 @@
...
+	name = "reinforced safety glass"
+	desc = "An extremely tough looking glass window with a special titanium rod matrice, most definitely bullet proof."
+	icon_state = "rwindow"
+	basestate = "rwindow"
+	health = 1500
+	reinf = TRUE
+
+
+/obj/structure/window/reinforced/briefing/get_projectile_hit_chance(obj/item/projectile/P)
+	return TRUE
```
Ghommie:
```suggestion
	if((flags_atom & ON_BORDER) && P.dir != dir && P.dir != reverse_direction(dir))
		return FALSE
	else
		return TRUE

```
So bullets parallel to it shouldn't get stopped.

code/game/machinery/doors/windowdoor.dm
Rohesie: This can be removed, it's not a must, if it makes the mappers' lives harder.

Rohesie:
People will survive with being unable to rotate a chair a hundred times per second.
It won't detract from gameplay, it won't make rotating a chair for usual purposes any noticeably harder, nor it will impair creativity, which is what the game is about. It will merely limit one localized visual effect contrasting with the rest of the game.
Slippery slope is a fallacy. Take a stand against something that actually makes a difference.

factoryman942:
local headmin can't spell "hugboxing"

Alterist:
make a note somewhere that says your snowflake windows can't be shot through with lasers

PR #571, opened 2019-02-06
MOD88 pistol stock and holster combo

Hughgent:
:cl: Hughgent
add: New MOD88 Stock and holster
/:cl:

[why]: 
[https://youtu.be/xEMTx5MNqk4?t=383](url)

image
There's one in the PO lockers on map.


Hughgent:
https://youtu.be/xEMTx5MNqk4?t=383
This is why.

Hughgent:
and I realized too late that I've got the attached stock on upside down.

Hughgent:
image
art is hard.

Borisvanmemes:
So how does it work?

Hughgent:
> 
> 
> So how does it work?

as an attachment it's equal in modification to the SMG stock of which it is based on.
in practicality, you can stick the pistol inside of it, and stick it on a pocket.
if you click it with a contained pistol while it's on your pocket, you draw the pistol.
if you click the mod88 pistol with your stock-holster, you attach it to the back as shown.
beyond that. you could store a size 1 item inside of it if you really wanted to.

(and yes, if you get a second mod88, you can have a mod88 in the stock-holster while it is attached to another mod88. No recursion beyond that though.)

Hughgent:
image
Ain't that pretty.

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -673,6 +673,56 @@ Defined in conflicts.dm of the #defines folder.
...
+	scatter_mod = -CONFIG_GET(number/combat_define/med_scatter_value)
+	movement_acc_penalty_mod = CONFIG_GET(number/combat_define/min_movement_acc_penalty)
+	pockets = new/obj/item/storage/internal(src)
+	pockets.storage_slots = 1
+	pockets.max_w_class = 1
+	pockets.bypass_w_limit = list("/obj/item/weapon/gun/pistol/vp70")
+	pockets.max_storage_space = 3
+	
+/obj/item/attachable/stock/vp70/attack_hand(mob/user)
+	if(loc == user && pockets.contents.len)
```
DominikPanic: Using length(x) instead of x.len is faster.

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -673,6 +673,56 @@ Defined in conflicts.dm of the #defines folder.
...
+	size_mod = 1
+	icon_state = "vp70stock" // Thank you to Manezinho
+	attach_icon = "vp70stock_a" // Thank you to Manezinho
+	pixel_shift_x = 39
+	pixel_shift_y = 11
+	var/obj/item/storage/internal/pockets
+
+/obj/item/attachable/stock/vp70/Initialize()
+	. = ..()
+	accuracy_mod = CONFIG_GET(number/combat_define/low_hit_accuracy_mult)
```
spookydonut: Needs to be in gun config proc

Hughgent: I presume this was what you meant.

PR #574, opened 2019-02-06
CMO letters of Commendation

Hughgent:
:cl: Hughgent & Tobinerd
add: CMO can now give letters of commendation
add: New "Letter of Commendation" "Medal" for captains to award civilians with.
/:cl:

[why]: Tobinerd came up with a great idea for giving awards to non-military staff.


Hughgent:
There's a bunch of, probably, superfluous junk in the communications/cmo section.

But at the moment, everything is working as intended, CMO can do two things, 1. give awards, and 2. send messages to TGMC high command.

TobiNerd:
Doctors need love too

Ghommie:
You should definitely try to bring some feature locks variable settings down to the parent path, even though the idea is nice.

Hughgent:
Working on a less code intensive option at the moment. no need to merge.

Hughgent:
the, "less code" option didn't work out, for some reason you can't have "pre-made" fields on paper.

now this has become a mess of attempting to re-add the biohazard setting to the ship displays that the CMO can only set.

Help, I am drowning in feature creep.

code/game/machinery/computer/communications.dm
```diff
@@ -500,6 +501,418 @@
...
+	desc = "This can be used for various important functions."
+	icon_state = "comm"
+	req_access = list(ACCESS_MARINE_CMO)
+	circuit = "/obj/item/circuitboard/computer/communications"
+	prints_intercept = FALSE
+	authenticated = 0
+	list/messagetitle = list()
+	list/messagetext = list()
+	currmsg = 0
+	aicurrmsg = 0
```
DominikPanic: I bet you're re-defining at least half of those without change.

code/game/machinery/computer/communications.dm
```diff
@@ -500,6 +501,418 @@
...
+								set_security_level(SEC_LEVEL_BIO)
+								feedback_inc("alert_comms_bio",1)
+				else
+					to_chat(usr, "<span class='warning'>You are not authorized to do this.</span>")
+				tmp_alertlevel = SEC_LEVEL_GREEN //Reset to green.
+				state = STATE_DEFAULT
+			else
+				to_chat(usr, "<span class='warning'>You need to swipe your ID.</span>")
+
+	/*	if("announce")
```
DominikPanic: Calling upper here makes commenting out things irrelevant. Also, big blocks of commented out code that shouldn't be there in the first place are bad.

code/game/machinery/computer/communications.dm
```diff
@@ -500,6 +501,418 @@
...
+
+		if("changeseclevel")
+			state = STATE_ALERT_LEVEL
+
+		else return FALSE
+
+	updateUsrDialog()
+
+/obj/machinery/computer/communications/cmo/attack_hand(var/mob/user as mob)
+	if(..())
```
DominikPanic: Calling upper here and adding another snowflake chcek for area is a bad solution, either don't call upper and have each check separate or make this into a variable (that would be better)

code/modules/paperwork/paper.dm
```diff
@@ -424,6 +424,14 @@
 /*
  * Premade paper
  */
+
+
+/obj/item/paper/commendation/
+	var/field = "<span class=\"paper_field\"></span>"
+	name = "Commendation"
+	//info = "<center><H1>Commendation<H2>of Medical Excellence</H2></H1><br>Proudly presented to<br><H2><I>[field]</I></H2><b>for superior performance and excellency in medical skills.</b><br><br><hr><br><small>They have gone above and beyond in saving human lives, treating wounded, chemical knowledge and surgical performance. Even when working in a difficult environment, they have shown outstanding courage, determination and commitment to save those in need.<br>They are a true example of the ideal doctor.</small><br><br><hr><br><b>Date</b><br><I>[field] - 2186</I><br><br><b>Granted by</b><br><I>Prof. [field]</I><br><small>Chief Medical Officer</small><br>[field]</center>"
```
DominikPanic: Why commented out?

Hughgent: At the moment, it's non-functional. If this was working, I would scrap all the changes to communications.dm.

As it is, this makes a properly formatted piece of paper, but doesn't input the fields. The fields can be put in the proper place if a person writes "[field][field][field]" with a pen. Then they can fill out the form normally.

Hughgent:
Screw you communications console!

Hughgent:
I'm going to give all this work to @TobiNerd so they can implement it in their maps PR!

Hughgent:
Implemented in #605

code/modules/paperwork/paper.dm
```diff
@@ -424,6 +427,23 @@
...
+	fields = 5
+
+/*Let this be a lesson about pre-made forms.
+
+when building your paper, use the above parsed pen code in parsepencode(). no square bracket anything in the info field.
+Specifically for the field parts, use <span class=\"paper_field\"></span>.
+then, for every time you included a field, increment fields. */
+
+/obj/item/paper/commendation/Initialize()
+	. = ..()
```
Rohesie: This essentially does nothing. Why do you need this?

code/modules/paperwork/paper.dm
```diff
@@ -56,6 +56,9 @@
 		return
 	icon_state = "paper"
 
+/obj/item/paper/commendation/update_icon() //it looks fancy and we want it to stay fancy.
```
Ghommie: You should properly index this below the item path.

Hughgent:
the timeline of events was
1. It would be nice if the CMO could give out awards to their doctors.
2. it would be even nicer if those awards showed up during the awarded medals section.
3. I'll give the cmo their own communications console so they can print medals.
4. oh god communications.dm is horrible to work with.
5. how about a paper form? that'd be very personal.
6. oh god why don't fields work in pre-made forms?
7. oh hey Tobin is doing a medbay rework, why not put the paper forms in there?
8. time goes by
9. that was a terrible idea, let's re-open this.
10. remove all the changes to communications.dm
11. pre-made forms work now! hooray!
12. letters of commendation that the CO can give out to not military is still neat, I'll leave medals_awards.dm alone.
13. just waiting for a review now.

and thus we get to now.
eventually, @Alterist said they would add in the papers to the CMO office, in their locker or on their desk.

so they are meant to be two different things. one by the captain, given out like any other medal, pins on shirts and everything, shows up at round end.
The other a more personal note from one player to the next.

Hughgent:
I do consider this complete at the moment, unless there is a better way of linking the two that I'm not sure how to make happen.

PR #578, opened 2019-02-07
Cleanup

Alterist:
Cleans up Weapons.dmi and Items.dmi a bit, placing skub into items instead of weapons
also updates esword and wirecutter sprites


spookydonut:
pro: skub

PR #588, opened 2019-02-07
Makes a new coffee vending item

Hughgent:
:cl: Hughgent
tweak: Coffee vendors got a supply of creamer
/:cl:

[why]: Cause I want milk in my coffee.


Ghommie:
You should lower the amounts of each selection by 5 units. I wonder if ordering roughly a dozen coffee cups still happen to drain the APC's charge dead anyway.

Hughgent:
can do. and it probably still does.

PR #596, opened 2019-02-08
Fixes the unilateral muhrin armor lamp overlay.

### LINKED ISSUE #603: Marine armor directional lamps broken
They have been for quite a long time, the overlay only faces south.

Ghommie:
:cl:
fix: Fixes the on-mob marine armor lamp overlay only having one direction.
/:cl:

Old issue resurfacing yet again. Closes #603.

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -727,7 +727,7 @@ var/global/list/damage_icon_parts = list()
 				for(var/i in marine_armor.armor_overlays)
 					I = marine_armor.armor_overlays[i]
 					if(I)
-						I = image(I.icon,src,I.icon_state)
+						I = image('icons/mob/suit_1.dmi',src,I.icon_state)
 						standing.overlays += I
 
 		if(wear_suit.blood_DNA)


code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -727,7 +727,7 @@ var/global/list/damage_icon_parts = list()
 				for(var/i in marine_armor.armor_overlays)
 					I = marine_armor.armor_overlays[i]
 					if(I)
-						I = image(I.icon,src,I.icon_state)
+						I = image('icons/mob/suit_1.dmi',src,I.icon_state)
```
spookydonut: Uhhhh I feel like this is the wrong way to do this

Ghommie: I could make a new file for the marine armor overlays, similarly to helmets, but I can't use the former icon, nor making it directional as it'd become an item sprite issue, with the overlay shifting position as you move the object around.

spookydonut: I mean having the dmi file path, shouldn't it be like a variable.

Ghommie: I mean, I just can think of one way to apply your indication at the moment, one moment.

PR #612, opened 2019-02-09
Set-transfer-amount automatically selects the active hand.

Hughgent:
:cl: Hughgent
refactor: set-transfer-amount is no more, it has been changed to the attack_self() proc. Activate-held-object or clicking the container also works.
fix: No more selecting transfer amounts for containers not in your active hand.
/:cl:

[why]: Because as a chemist I often hold the Bluespace beaker and a medicine bottle at the same time while changing the medicine bottle to 60. thus it would be nice to only select the medicine bottle when activating the verb.

**Or at least that would be nice to say if it actually worked. cause apparently this always brings back both hands contents as a list. Do not merge as is.**

**I need help from someone who understands why this is happening.**

### DIFF
--- a/code/game/objects/items/reagent_containers/food/cans.dm
+++ b/code/game/objects/items/reagent_containers/food/cans.dm
@@ -11,11 +11,10 @@
 		canopened = TRUE
 		container_type = OPENCONTAINER_NOUNIT
 		return
-	else
-		var/obj/item/reagent_container/H = usr.get_active_held_item()
-		var/N = input("Amount per transfer from this:","[H]") as null|anything in H.possible_transfer_amounts
-		if (N)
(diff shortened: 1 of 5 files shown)


Hughgent:
turns out the set-transfer-amount verb was tied to the items, rather than a thing a person could do. Thus activating the verb caused all items in a mobs inventory to trigger the verb.

code/game/objects/items/reagent_containers/food/cans.dm
```diff
@@ -11,6 +11,11 @@
 		canopened = TRUE
 		container_type = OPENCONTAINER_NOUNIT
 		return
+	else
```
spookydonut: dont need the else just drop this down an indent it returns above

PR #618, opened 2019-02-09
Pain Rebalance

Surrealaser:
:cl: by Surrealistik
tweak: Oxy loss pain multiplier reduced from 1 to 0.75, Tox loss increased to 0.7 to 0.75. Brute and Clone loss reduced from 1.5 to 1.
/:cl:


Rohesie:
As this is essentially yet another buff, it will have to be re-thought once stamina damage does in and the massive halloss damages go out.

PR #623, opened 2019-02-09
Update Z.01.Desert-Dam.dmm

Alterist:
Deleted removed objects, changed antitox and cryo_cell obj paths on desert dam.


Ghommie:
Remember to also change the shuttle door subtype if #606 gets merged as I didn't  handle it right there.

Alterist:
changed shuttle doors

PR #633, opened 2019-02-10
Chairs no longer rotate at the speed of light.

Rohesie:
[Changelogs]: 
tweak: Chairs no longer rotate at the speed of light. You need to be in the same tile as them to do it.
/:cl:

[why]: Objects behave more uniformly.

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -143,8 +143,6 @@
 
 /datum/config_entry/flag/bones_can_break
 
-/datum/config_entry/flag/ghost_interaction
-
 /datum/config_entry/flag/unlimited_rotate_speed
 
 /datum/config_entry/flag/allow_random_events
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -143,6 +143,10 @@
 
 /datum/config_entry/flag/bones_can_break
 
+/datum/config_entry/flag/ghost_interaction
+
+/datum/config_entry/flag/unlimited_rotate_speed
+
 /datum/config_entry/flag/allow_random_events
 
 /datum/config_entry/flag/allow_synthetic_gun_use
--- a/code/game/objects/structures/stool_bed_chair_nest/chairs.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/chairs.dm
@@ -39,19 +39,21 @@
 	set category = "Object"
 	set src in view(0)
 
-	if(CONFIG_GET(flag/ghost_interaction))
+	if(isobserver(usr) && CONFIG_GET(flag/ghost_interaction))
 		setDir(turn(dir, 90))
-		return FALSE
+		return TRUE
 	
 	var/mob/living/carbon/user = usr
 
-	if(!istype(user) || !isturf(user.loc) || user.stat || user.is_mob_restrained())
+	if(!istype(user) || !isturf(user.loc) || user.is_mob_incapacitated())
 		return FALSE
 
-	if (world.time <= user.next_move)
+	if(!CONFIG_GET(flag/unlimited_rotate_speed) && world.time <= user.next_move)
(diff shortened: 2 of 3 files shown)


code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -37,20 +37,22 @@
...
-		if(!usr || !isturf(usr.loc))
-			return
-		if(usr.stat || usr.is_mob_restrained())
-			return
+		setDir(turn(dir, 90))
+		return FALSE
+	
+	var/mob/living/carbon/user = usr
+
+	if(!istype(user) || !isturf(user.loc) || user.stat || user.is_mob_restrained())
```
Ghommie: Use `user.is_mob_incapacitated()` instead of `user.stat || user.is_mob_restrained()`, it covers stun and knockdown statuses in additions of handcuffs and mob stat.

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -37,20 +37,22 @@
 /obj/structure/bed/chair/verb/rotate()
 	set name = "Rotate Chair"
 	set category = "Object"
-	set src in oview(1)
+	set src in view(0)
```
Ghommie: Not documented in the changelog. 

Also I'd suggest adding a control statement on the buckled mob's a_intent rather than this (that implying your issue is people spinning adjacent occupied chairs), because it's quite wonky having to stand on the same turf to rotate it.

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -37,20 +37,22 @@
...
-			return
+		setDir(turn(dir, 90))
+		return FALSE
+	
+	var/mob/living/carbon/user = usr
+
+	if(!istype(user) || !isturf(user.loc) || user.stat || user.is_mob_restrained())
+		return FALSE
+
+	if (world.time <= user.next_move)
```
spookydonut: I suggest moving this to a config value

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -37,20 +37,22 @@
...
+		return FALSE
 
-		setDir(turn(src.dir, 90))
+	user.next_move = world.time + 3
```
spookydonut: I suggest moving this to a config value

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
DominikPanic: Personally I wouldn't over-config everything like /tg/ does.

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
Rohesie: Spooky suggested this change, instead of an `Adjacent()` check.
Mostly due to rotating bypassing windows and whatnot, and to avoid that check.

Rohesie: Changelog added.

Ghommie: Well, fine.

Rohesie: I don't mind either way, personally.

code/controllers/configuration/entries/game_options.dm
```diff
@@ -143,6 +143,10 @@
 
 /datum/config_entry/flag/bones_can_break
 
+/datum/config_entry/flag/ghost_interaction
```
spookydonut: per the other pr these should have the production values as defaults here

Rohesie:
This PR has complied with all change requests.

DominikPanic:
Even though I disagree with the config this PR is fine code-wise so here we go.

PR #636, opened 2019-02-10
Hand Labeler rework

### LINKED ISSUE #591: Hand Labelers Don't Remove Old Labels
## Reproduction:

1) Label something with a hand labeler.
2) Turn off the hand labeler.
3) Label that same something with a hand labeler.
4) Observe how it now says "itemname (First Label) (Second Label)", instead of relabelling as "itemname (Second Label)"

DominikPanic:
:cl: LaKiller8
fix: Hand labeler can now label backpacks without them having to be full first. Every item can only have one label now.
/:cl:
Fixes #591

### DIFF
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -351,13 +351,6 @@
 				to_chat(usr, "<span class='notice'>[src] cannot hold [W].</span>")
 			return FALSE
 
-	if(istype(W, /obj/item/tool/hand_labeler))
-		var/obj/item/tool/hand_labeler/L = W
-		if(L.on)
-			return FALSE
-		else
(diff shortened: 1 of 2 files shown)


code/game/objects/items/storage/storage.dm
```diff
@@ -347,15 +347,17 @@
...
+			if(warning)
 				to_chat(usr, "<span class='notice'>[src] cannot hold [W].</span>")
 			return FALSE
 
+	if(istype(W, /obj/item/tool/hand_labeler))
+		var/obj/item/tool/hand_labeler/L = W
+		if(L.on)
+			return FALSE
+		else
+			return TRUE
```
spookydonut: Should still check the blacklists below

PR #639, opened 2019-02-10
refactors gun default attachment code

spookydonut:
ignoring `starting_attachment_types` and messing with flags in gun Inits is hacky and bad


Ghommie:
the `var/list/starting_attachment_types` in the root comes with a dumb comment indicating it should only be just used for attachments that can be effectively removed. So if you want, go ahead and "removable" string from it.

spookydonut:
Only because most attachments are removable by default and no one added subtypes properly

code/_onclick/hud/screen_objects.dm
```diff
@@ -631,7 +631,7 @@
 
 	var/obj/item/weapon/gun/G = user.get_active_held_item()
 
-	if(!G || !G.has_ammo_counter() || !G.hud_enabled)
+	if(!G || !(G.flags_gun_features & GUN_AMMO_COUNTER) || !G.hud_enabled)
```
Rohesie:
```suggestion
	if(!G?.hud_enabled || !(G.flags_gun_features & GUN_AMMO_COUNTER))
```

code/_onclick/hud/screen_objects.dm
```diff
@@ -645,7 +645,7 @@
 
 	var/obj/item/weapon/gun/G = user.get_active_held_item()
 
-	if(!G || !istype(G) || !G.has_ammo_counter() || !G.hud_enabled || !G.get_ammo_type() || isnull(G.get_ammo_count()))
+	if(!G || !istype(G) || !(G.flags_gun_features & GUN_AMMO_COUNTER) || !G.hud_enabled || !G.get_ammo_type() || isnull(G.get_ammo_count()))
```
Rohesie:
```suggestion
	if(!istype(G) || !(G.flags_gun_features & GUN_AMMO_COUNTER) || !G.hud_enabled || !G.get_ammo_type() || isnull(G.get_ammo_count()))
```

code/__DEFINES/conflict.dm
```diff
@@ -35,22 +35,23 @@
...
+#define GUN_AMMO_COUNTER		(1 << 7)
+#define GUN_BURST_ON			(1 << 8)
+#define GUN_BURST_FIRING		(1 << 9)
+#define GUN_FLASHLIGHT_ON		(1 << 10)
+#define GUN_WIELDED_FIRING_ONLY	(1 << 11)
+#define GUN_HAS_FULL_AUTO		(1 << 12)
+#define GUN_FULL_AUTO_ON		(1 << 13)
+#define GUN_POLICE				(1 << 14)
+#define GUN_ENERGY				(1 << 15)
+#define GUN_LOAD_INTO_CHAMBER	(1 << 16)
```
DominikPanic: You should also add this into the appropriate bitfield so us admins can ~~bus~~ VV things easily.

DominikPanic:
SMG's with this PR cannot burstfire.

PR #646, opened 2019-02-11
New Runner Passive: Hit and Run

Surrealaser:
:cl: by Surrealistik
add: Hit and Run involves building momentum. For every tile moved, the Runner's slash/tackle damage increases by +5% to a maximum of +100% damage or a 2x damage multiplier. All Hit and Run bonuses are lost if the Runner is stationary for more than 0.5 seconds, and upon hitting a target.
/:cl:


Notamaniac:
Does this mean table runners is the new meta?

Or hiding under xeno corpses

code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -2413,4 +2413,27 @@
...
+	return damage
+
+/mob/living/carbon/Xenomorph/Runner/hit_and_run_bonus(damage)
+	var/last_move = last_move_intent - 10
+	var/bonus
+	if(last_move && last_move < world.time - 5) //If we haven't moved in the last 500 ms, we lose our bonus
+		hit_and_run = 1
+	bonus = CLAMP(hit_and_run, 1, 2)//Runner deals +5% damage per tile moved in rapid succession to a maximum of +100%. Damage bonus is lost on attacking.
+	switch(bonus)
+		if(2 to INFINITY)
```
Rohesie:
```suggestion
		if(1.96 to 2)
```
Or `if(2)`
It's already clamped, it can't be more than 2.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -44,6 +44,13 @@
 		attack_delay = initial(attack_delay) - round(rage * 0.05,0.01) //-0.05 attack delay to a maximum reduction of -2.5
 	return ..()
 
+/mob/living/carbon/Xenomorph/Runner/handle_status_effects()
+	if(hit_and_run)
```
Rohesie: Is there a possibility of this not being true?

Surrealaser: sanity check

PR #654, opened 2019-02-11
Ports /tg/'s old legacy map changer to hold us over.

MrStonedOne:
Works by having the server tools recompile with certain files included

:cl:
add: Added a temporary map rotator to hold us over until spooky finishes the /tg/ map loader 
add: The current map has a reduced chance of being selected again.
config: Added player preference for their preferred map. This will be used to weight the random map rotation
admin: Admins can manually load maps and manually trigger map rotation
server: Server operators should Edit maps.txt and config.txt to their liking.
/:cl:

This is the map rotation system that /tg/ had before the map loader, much of the map loader was built on top of it.

**Once test merged this can not be simply un-testmerged**: If you remove this from the test merges compiles will fail until you attempt to rotate the map again, that will fail but then compiles and updates will succeed again.

@spookydonut if this ends up getting merged in, you can just override all of these changes with your stuff.


code/modules/mapchanger/mapchanger.dm
```diff
@@ -0,0 +1,105 @@
...
+	ASSERT(instancename)
+
+	log_game("Changing map to [VM.name]([VM.friendlyname])")
+
+	var/const/filename = "..\\..\\static\\maprotate.dm"
+	fdel(filename)
+	var/file = file(filename)
+
+	file << "#include \"_maps\\[VM.name].dm\"\n#define MAP_OVERRIDE"
+	. = shell("[tgs3_path] --instance [instancename] dm compile --wait") //don't judge me cyberboss
```
MrStonedOne: @Cyberboss :eyes:

Cyberboss: _breathes slowly_

Cyberboss: iirc this... SHOULD work fine, the --wait means you'll be waiting a LONG time though

Cyberboss: Alternatively, you could NOT wait and use the callback in tgs.config.json to indicate completion

Cyberboss: Namely this:

https://github.com/tgstation/TerraGov-Marine-Corps/blob/master/code/__DEFINES/tgs.config.dm#L6

Will be called with this string: 

https://github.com/tgstation/tgstation-server/blob/V3/TGS.Server/Instance/Compiler.cs#L570

Cyberboss: This is nextlevel tool stitching lmao. I'll take a closer look later

MrStonedOne: no, --wait is the proper way. if i used callbacks i couldn't know if it failed.

it takes tgmc 25 seconds to compile on my computer, and this is just a temp thing to hold us over until we get the proper map loader ported in.

PR #663, opened 2019-02-11
Admin Fixes

DominikPanic:
Fixes faxes
Explosion logging

### DIFF
--- a/code/game/machinery/fax.dm
+++ b/code/game/machinery/fax.dm
@@ -16,6 +16,7 @@
 	var/sendcooldown = 0
 
 	var/department = "Corporate Liasion"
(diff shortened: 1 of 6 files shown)


code/game/machinery/fax.dm
```diff
@@ -16,6 +16,7 @@
 	var/sendcooldown = 0
 
 	var/department = "Corporate Liasion"
+	var/selected = "Nanotrasen"
```
spookydonut: Make the name nanotrasen a define to stop misspellings or is that too far?

DominikPanic: The user can only pick from a list so it cannot be misspelled

PR #665, opened 2019-02-12
Replaces legacy json procs with built-ins

DominikPanic:


DominikPanic:
DNM until MSO figures out why it breaks nanoui and how to fix it.

PR #671, opened 2019-02-12
Cleans up configs, makes the defaults friendlier for downstreams, fixes a few fuckups.

DominikPanic:


MrStonedOne:
The Galaxy Brain way to do configs is to have the in code default (config_entry_value) default to sane production values, and have the .txt default, default to sane development defaults.

So the repo config shouldn't necessarily be the same as production.

DominikPanic:
> 
> 
> The Galaxy Brain way to do configs is to have the in code default (config_entry_value) default to sane production values, and have the .txt default, default to sane development defaults.
> 
> So the repo config shouldn't necessarily be the same as production.

Wise, I'll do it like that

PR #673, opened 2019-02-12
A/B profile byond json procs in production

MrStonedOne:
Simply start a round and start a profile, at the end of the round, you can see how well both procs do by comparing their cpu usage.


DominikPanic:
Is this still necesssary?

PR #679, opened 2019-02-13
You can no longer draw/equip under all circumstances.

### LINKED ISSUE #675: You can draw your gun inside xenos
If you are somehow not in crit while inside a Xeno, you can draw your gun with E. If you do this, your gun will disappear the moment you get out of the Xeno, even with (or especially with) a mag harness.

DominikPanic:
:cl: LaKiller8
fix: You can no longer draw/equip while incapacitated.
/:cl:
Fixes #675

### DIFF
--- a/code/modules/mob/living/carbon/human/inventory.dm
+++ b/code/modules/mob/living/carbon/human/inventory.dm
@@ -2,6 +2,9 @@
 	set name = "quick-equip"
 	set hidden = TRUE
 
+	if(is_mob_incapacitated() || lying || buckled)
+		return
+
 	var/obj/item/I = get_active_held_item()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -2,6 +2,9 @@
 	set name = "quick-equip"
 	set hidden = TRUE
 
+	if(is_mob_incapacitated() || lying || buckled)
```
Ghommie: buckled seems unnecessary, or perhaps I'm missing out the motive.
This should also be undoable whilst inside a mecha, pretty much like other inventory interactions.

PR #682, opened 2019-02-13
Crushers get a unique roar when charging

Spyroshark:
:cl:
soundadd: Crushers now make a unique, louder roar when their charge is at max speed.
/:cl:

Since some of the new xeno roars are meant to sound more like pain noises and not battle cries, I thought the crusher should sound more aggressive when it's charging roar plays. Sounds are from Left 4 Dead 2.


Spyroshark:
A lot of people don't like the new sound and it gets too repetitive. Closing this since it's not very good.

Rohesie:
You could add the among the regular roars.

PR #687, opened 2019-02-13
Fixes Overall Damage References for Stomp and Headbutt Abilities

Surrealaser:
Per title


Ghommie:
And this is why the parent proc should have the same args of the children versions.

PR #700, opened 2019-02-15
Renames RoleAuthority to SSjob

DominikPanic:
More :atom: for spookster


Rohesie:
Speedmerged, tested and all.

PR #711, opened 2019-02-16
Cell Destabilization Fix for No Charge

Surrealaser:
:cl: by Surrealistik
tweak: Retooled power cell detonation to scale with charge more appropriately.
fix: Fixed depleted cells not detonating properly.
tweak: Fuse timer on a destabilized power cell may be as low as 0.3 seconds; there is risk of premature detonation.
tweak: Destabilized power cells immediately detonate when power is drawn from them, including when loaded into a lasgun.
tweak: Destabilized power cells now have a lightning overlay that makes their state obvious.
/:cl:


Rohesie:
Would this be necessary if there wasn't a minimum value here associated to the explosion?

DominikPanic:
This doesn't change throw mode at all

Rohesie:
Can't this still produce a (0, 0, 0) infinite explosion?

Surrealaser:
@Rohesie Not per the code as it exists, but I'll re-test after work tonight.

Rohesie:
```

**Comment by Rohesie** [comment]
```

explosion(T, devastation_range, heavy_impact_range, light_impact_range, flash_range)
```

File: code/modules/projectiles/updated_projectiles/guns/energy.dm [hunk 1]
```diff
@@ -385,10 +385,8 @@
 			playsound(user, active_attachable.activation_sound, 15, 1)
 			active_attachable.activate_attachment(src, null, TRUE)
 
-	if(!cell || cell.charge - charge_cost < 0)
+	if(!cell || !cell.use(charge_cost))
```
Rohesie:
```suggestion
	if(!cell?.use(charge_cost))
```

DominikPanic:
@spookydonut QDEL_IN is the right solution here, correct?

spookydonut:
if this is working fine then merge it

PR #716, opened 2019-02-16
Boiler Damage Improvement vs Vehicles/Structures

Surrealaser:
:cl: by Surrealistik
tweak: Boiler acid globs will now do 2x damage to most structures (barricades, sentries, etc) and 3x damage to vehicles
/:cl:


Rohesie:
We have to refactor how damage is processed, and yet more snowflake checks just add coder debt.
Also the effect here is very minor because it only applies to boiler direct hits.
Make it so the projectile covers what was hit in acid if applicable or something like that instead, but stop adding snowflake checks.

DominikPanic:
Agreed with Rohesie here. We can't afford to keep tacking on these changes forever.

Rohesie:
Why are we testing it then?

DominikPanic:
> 
> 
> Why are we testing it then?

Test-merges are temporary and undone with a single tick, we can still observe how badly coded damage tweak affects balance without polluting the codebase.

Rohesie:
Closed for now until a solution that doesn't add coder debt presents itself.

PR #730, opened 2019-02-16
Mateba and M44 belt fix

Pariah919:
Should fix the M44 and Mateba (Not that the Mateba can have a stock or slot increasing attachment I believe) being to be held in the belt while having a stock or other slot increasing attachable.
:cl: MJP
fix: M44 and mateba belts can now only hold things of proper size
/:cl:
EDIT:Cleaned up my engrish and added :Cl: and Fix:


Rohesie:
The idea of size increasing attachments is not to fit in containers. That's the whole balance logic behind them.

Pariah919:
Oops. I meant being able to be in the belt while having a stock or slot increasing attachment in them. I typoed like a idiot.

DominikPanic:
Could you changelog the change please?
@Rohesie This means the affected belts are now max_w_class 3, inherited from gun belt.

PR #737, opened 2019-02-17
Medevac stretcher no bucklee removal

Shadow-Quill:
:cl:
del: Medevac stretchers will no longer yell at you for not having something buckled in when you start the teleport process.
/:cl:

Weird that they yell at you, but still start the teleport sequence. Decided to remove it instead of adding a `return` so as to allow lone medics a chance to teleport themselves out of danger.

~~Too bad the xeno ate their bed for lunch.~~


Rohesie:
The spawn would be better replaced by a CALLBACK.

Shadow-Quill:
> The spawn would be better replaced by a CALLBACK.

Done.

PR #742, opened 2019-02-17
Fixing the not fixed energy guns description

Ghommie:
## About The Pull Request
Fixing the not properly tested energy gun description fix. Refactored the ammo count description into a different proc to do that.


DominikPanic:
I can no longer trust this

PR #755, opened 2019-02-18
fixing xeno FF staggerstun

### LINKED ISSUE #745: As xeno, getting hit by an allied spit shakes your screen
Pretty annoying, seems unintended

Ghommie:
## About The Pull Request
Closes #745.

## Why It's Good For The Game
Neurotoxin spit shouldn't affect teammates, for consistency.

## Changelog
:cl:
fix: Fixes xeno neurotoxin applying friendly fire staggerstun.
/:cl:

### DIFF
--- a/code/modules/projectiles/updated_projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/updated_projectiles/ammo_datums.dm
@@ -1293,7 +1293,7 @@
 	damage_var_high = CONFIG_GET(number/combat_define/mlow_proj_variance)
 
 /datum/ammo/xeno/toxin/on_hit_mob(mob/living/carbon/M, obj/item/projectile/P)
-	if(!istype(M) || xeno_hivenumber(M) == xeno_hivenumber(P.firer))
+	if(!istype(M) || xeno_hivenumber(M) && xeno_hivenumber(M) == xeno_hivenumber(P.firer))
 		return ..()
 	var/mob/living/carbon/C = M
 	if(C.status_flags & XENO_HOST && istype(C.buckled, /obj/structure/bed/nest) || C.stat == DEAD)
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -1410,11 +1410,12 @@
 	burst(M,P,damage_type)
 
 /datum/ammo/xeno/acid/on_hit_mob(mob/M, obj/item/projectile/P)
-	if(iscarbon(M))
-		var/mob/living/carbon/C = M
-		if(C.status_flags & XENO_HOST && istype(C.buckled, /obj/structure/bed/nest) || C.stat == DEAD)
-			return
-	..()
+	if(!iscarbon(M) || xeno_hivenumber(M) && xeno_hivenumber(M) == xeno_hivenumber(P.firer))
+		return ..()
```
DominikPanic: Are you sure you wanna do this? Shouldn't you just return here? Since there's no difference currently between this case and the mob being carbon or with a different hivenumber.

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -1293,7 +1293,7 @@
 	damage_var_high = CONFIG_GET(number/combat_define/mlow_proj_variance)
 
 /datum/ammo/xeno/toxin/on_hit_mob(mob/living/carbon/M, obj/item/projectile/P)
-	if(!istype(M))
+	if(!istype(M) || xeno_hivenumber(M) && xeno_hivenumber(M) == xeno_hivenumber(P.firer))
```
DominikPanic: The only difference right now is the staggerstun

code/modules/projectiles/updated_projectiles/ammo_datums.dm
Ghommie: Yea, actually it could just return without the ..() since the parent's is null.

PR #762, opened 2019-02-18
Warrior, Neurotox, Flechette and Lasgun Nerfs

Rohesie:
## Changelog
:cl:
balance: Warrior grab knockdown time reduced.
balance: Neurotoxin chem gets metabolized quicker. Neurotoxin spit gets a 30u threshold.
balance: Flechettes and lasgun overcharge mode both lose a little damage.
balance: Lasguns can no longer fit scopes. They can still fit mini-scopes, though.
/:cl:

### DIFF
--- a/code/modules/projectiles/updated_projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/updated_projectiles/ammo_datums.dm
@@ -646,7 +646,7 @@
 	accuracy_var_low = CONFIG_GET(number/combat_define/med_proj_variance)
 	accuracy_var_high = CONFIG_GET(number/combat_define/med_proj_variance)
 	max_range = CONFIG_GET(number/combat_define/short_shell_range)
-	damage = CONFIG_GET(number/combat_define/med_hit_damage)
+	damage = CONFIG_GET(number/combat_define/lmmed_hit_damage)
 	damage_var_low = -CONFIG_GET(number/combat_define/low_proj_variance)
 	damage_var_high = CONFIG_GET(number/combat_define/low_proj_variance)
 	damage_falloff *= 0.5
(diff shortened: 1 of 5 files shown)


Surrealaser:
Warrior knockdown reduction is probably an over-nerf.

Neuro metabolizing faster is probably only really going to hurt Sentinels and Defilers using neurospines which isn't a problem.

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -1292,16 +1292,28 @@
...
 
+
 /datum/ammo/xeno/toxin/on_hit_mob(mob/living/carbon/M, obj/item/projectile/P)
+
 	if(!istype(M))
 		return ..()
 	var/mob/living/carbon/C = M
-	if(C.status_flags & XENO_HOST && istype(C.buckled, /obj/structure/bed/nest) || C.stat == DEAD)
+
+	if(C.status_flags & XENO_HOST && istype(C.buckled, /obj/structure/bed/nest) || C.stat == DEAD || !(ishuman(C) || ismonkey(C)))
```
Ghommie: I'd suggest you move the `(ishuman(C) || ismonkey(C))` check just above the ammo reagents magic instead, because this would make #755 useless as of now.

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -1292,16 +1292,28 @@
...
 		return
+
 	staggerstun(C, P, CONFIG_GET(number/combat_define/close_shell_range), 0, 0, 1, 1, 0) //Staggers and slows down briefly
 
+	for(var/r_id in ammo_reagents)
+		var/on_mob_amount = C.reagents.get_reagent_amount(r_id)
+		var/amt_to_inject = ammo_reagents[r_id] //Never inject more than 30u through spitting
+		if(amt_to_inject + on_mob_amount > 30)
+			amt_to_inject = 30 - on_mob_amount
```
Ghommie: OD still expires whenever the reagent amount goes below the threshold rather than on deletion. That said, this makes the OD simply last a lone tick at most, is this is your plan?

Rohesie: Yep, I don't even see a need for xenos to ever OD marines on this.

PR #766, opened 2019-02-18
Sniper gains an integrated laze in its gun.

Rohesie:
## Why It's Good For The Game

The philosophy design for the sniper is the long-range support and debuffer, more than death from afar. The idea is to reduce off-screen deaths as much as possible, giving some functions now as part of the Scout to the Sniper, and making the Scout more suited for closer range engagements.

## Changelog
:cl:
add: Sniper gains an integrated laze in its gun. Use CTRL+click, same as a regular tactical binocular.
refactor: Refactors the Laser Targeting into a Target Marker.
/:cl:

### DIFF
--- a/code/_onclick/handle_click.dm
+++ b/code/_onclick/handle_click.dm
@@ -9,11 +9,10 @@
 	return FALSE
 
 /obj/item/weapon/gun/rifle/sniper/M42A/handle_click(mob/living/user, atom/A, list/mods)
-	if (mods["ctrl"])
+	if(mods["ctrl"])
 		integrated_laze.acquire_target(A, user)
 		return TRUE
-	else
-		return ..()
+	return ..()
 
 /obj/machinery/m56d_hmg/handle_click(mob/living/carbon/human/user, atom/A, list/mods)
 	if(mods["middle"] || mods["shift"] || mods["alt"] || !operator || operator != user)
--- a/code/_onclick/handle_click.dm
+++ b/code/_onclick/handle_click.dm
@@ -8,6 +8,13 @@
 		return TRUE
 	return FALSE
 
+/obj/item/weapon/gun/rifle/sniper/M42A/handle_click(mob/living/user, atom/A, list/mods)
+	if (mods["ctrl"])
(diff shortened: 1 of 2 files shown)


Surrealaser:
I like it.

Borisvanmemes:
Maybe give it a mode that makes it fire the railgun without command needing to to push the button for it?

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -96,6 +99,8 @@
 
 /obj/item/weapon/gun/rifle/sniper/M42A/Destroy()
 	laser_off()
+	qdel(integrated_laze)
```
Ghommie: `QDEL_NULL(integrated_laze)`, and remove the line below.

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -39,6 +40,8 @@
 	select_gamemode_skin(type, list(MAP_ICE_COLONY = "s_m42a"))
 	. = ..()
 	LT = image("icon" = 'icons/obj/items/projectiles.dmi',"icon_state" = "sniper_laser", "layer" =-LASER_LAYER)
+	integrated_laze = new
```
Ghommie: `integrated_laze = new(src)`, and remove the line below.

code/_onclick/handle_click.dm
```diff
@@ -8,6 +8,13 @@
 		return TRUE
 	return FALSE
 
+/obj/item/weapon/gun/rifle/sniper/M42A/handle_click(mob/living/user, atom/A, list/mods)
+	if (mods["ctrl"])
+		integrated_laze.acquire_target(A, user)
+		return TRUE
+	else
```
Ghommie: no need for `else` here.

Rohesie:
That would kill the overwatch functions, Boris, but in the long run after we give overwatch nicer tools it might be somthing to consider.

PR #775, opened 2019-02-19
Adds shutdown server admin verb.

MrStonedOne:
Requires the server be at pregame or endround.

Requires the start/end be delayed

Requires two confirmations from the same admin, 30 seconds apart.

During the first 30 second delay, all admins are notified and alerted and those with +SERVER can cancel the process within those 30 seconds

After that final confirmation, it announces the shutdown to players, and waits another 30 seconds where admins can cancel the shutdown still.

Requires the `ALLOW_SHUTDOWN` config. This config is protected from runtime editing via VV.

image

### DIFF
--- a/code/modules/admin/server_verbs.dm
+++ b/code/modules/admin/server_verbs.dm
@@ -90,12 +90,31 @@
 		return
 
 	to_chat(world, "<span class='danger'>Server shutting down.</span> <span class='notice'>Initiated by: [shuttingdown]</span>")
+	log_game("Server shutting down. Initiated by: [shuttingdown]")
 
+#ifdef TGS_V3_API
 	if (GLOB.tgs)
 		var/datum/tgs_api/TA = GLOB.tgs
-		TA.EndProcess()
+		var/tgs3_path = CONFIG_GET(string/tgs3_commandline_path)
+		if (fexists(tgs3_path))
+			var/instancename = TA.InstanceName()
+			if (instancename)
+				shell("[tgs3_path] --instance [instancename] dd stop --graceful") //this tells tgstation-server to ignore us shutting down
+			else
+				var/msg = "WARNING: Couldn't find tgstation-server3 instancename, server might restart after shutdown."
+				message_admins(msg)
+				log_game(msg)
+		else
+			var/msg = "WARNING: Couldn't find tgstation-server3 command line interface, server will very likely restart after shutdown."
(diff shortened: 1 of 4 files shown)


code/modules/admin/server_verbs.dm
```diff
@@ -17,6 +17,87 @@
...
+		return
+
+	to_chat(world, "<span class='danger'>Server shutting down.</span> <span class='notice'>Initiated by: [shuttingdown]</span>")
+
+	if (GLOB.tgs)
+		var/datum/tgs_api/TA = GLOB.tgs
+		TA.EndProcess()
+
+	sleep(world.tick_lag)
+	qdel(world) //there are a few ways to shutdown the server, but this is by far my favorite
```
MrStonedOne: ^^^

code/modules/admin/server_verbs.dm
```diff
@@ -17,6 +17,87 @@
...
+		return
+
+	if (shuttingdown != usr.ckey) //somebody cancelled but then somebody started again.
+		return
+
+	to_chat(world, "<span class='danger'>Server shutting down.</span> <span class='notice'>Initiated by: [shuttingdown]</span>")
+
+	if (GLOB.tgs)
+		var/datum/tgs_api/TA = GLOB.tgs
+		TA.EndProcess()
```
MrStonedOne: @Cyberboss this restarts the server, it doesn't shut it down like the name implies.

What should I put here if I want to shut the server down?

config/config.txt
```diff
@@ -21,6 +21,17 @@ HUBPASSWORD kMZy3U5jJHSiBQjr
 ## Add a # infront of this if you want to use the SQL based admin system, the legacy system uses admins.txt. You need to set up your database to use the SQL based system.
 ADMIN_LEGACY_SYSTEM
 
+## Add a # infront of this to disable the shutdown server verb.
+ALLOW_SHUTDOWN
```
Rohesie:
```suggestion
##ALLOW_SHUTDOWN
```
This should be disabled by default. It's more likely for downstreams to not be using TGS3.

MrStonedOne: 1 config file defaults are aimed at development defaults, not production defaults.

2 this does not need tgs to work, it just needs snowflake if you have tgs.

Rohesie: Fair

PR #785, opened 2019-02-19
Hunters Must Be on Same Z Level As Queen to Regen Plasma Off Weeds

Surrealaser:
:cl: by Surrealistik
balance: While a Queen is alive, Hunters must be on the same Z level to regain plasma off-weeds.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -223,7 +223,9 @@
 		if(recovery_aura)
 			plasma_stored += round(xeno_caste.plasma_gain * recovery_aura * 0.25 * modifier) //Divided by four because it gets massive fast. 1 is equivalent to weed regen! Only the strongest pheromones should bypass weeds
 	else
-		plasma_stored++
+		var/datum/hive_status/hive = hive_datum[hivenumber]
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -223,7 +223,9 @@
 		if(recovery_aura)
 			plasma_stored += round(xeno_caste.plasma_gain * recovery_aura * 0.25 * modifier) //Divided by four because it gets massive fast. 1 is equivalent to weed regen! Only the strongest pheromones should bypass weeds
 	else
-		plasma_stored++
+		var/datum/hive_status/hive = hive_datum[hivenumber]
+		if(!hive.living_xeno_queen || hive.living_xeno_queen.loc.z == loc.z) //We only regenerate plasma off weeds while on the same Z level as the queen; if one's alive
```
DominikPanic: If your intention is to regenerate while a queen is as you describe in the comment, this should be made into:
`hive.living_xeno_queen && hive.living_xeno_queen.loc.z == loc.z`

Ghommie: Actually this is the way it should be: their out-of-weed plasma generation should be halted only if a queen is alive and their z level location isn't the same of the src's.

Surrealaser: Ghommie's got it.

Hunters will regenerate off weeds as normal per this logic if A: there is no queen OR B: there is a queen, and it's on the same Z level.

PR #788, opened 2019-02-19
Removes icon_states from all map files

spookydonut:
## About The Pull Request
this removes all the icon_states used in map files with subtypes and dirs

## Why It's Good For The Game
makes life easier for mappers and shrinks the size of the files.  Less problems with weird hardset icons.


Alterist:
Yes, a thousand times yes

PR #789, opened 2019-02-19
Adds preferences for the two new equipment slots.

### LINKED ISSUE #714: E hotkey won't equip shotgun shells into the shotgun shell belt
Small oversight.

DominikPanic:
Fixes #714


Rohesie:
Since it holsters small items, perhaps the helmet should come first in the order. Else it becomes useless by default, as the items will always be placed anywhere else that has room for it.

PR #795, opened 2019-02-20
Razorwire cannot entangle if it's been destroyed.

Surrealaser:
Per title.


code/game/objects/structures/razorwire.dm
```diff
@@ -50,9 +50,11 @@
 	razorwire_tangle(M)
 
 /obj/structure/razorwire/proc/razorwire_tangle(mob/living/M, duration = RAZORWIRE_ENTANGLE_DELAY)
+	if(!src || health <= 0) //Sanity check so that you can't get entangled if the razorwire is destroyed; this happens apparently.
```
Ghommie:
```suggestion
	if(QDELETED(src) || health <= 0) //Sanity check so that you can't get entangled if the razorwire is destroyed; this happens apparently.
```

PR #799, opened 2019-02-20
Gun Balance Tweak

Surrealaser:
🆑 by Surrealistik
balance: Shotgun base wield slowdown reduced from 0.75 to 0.35
balance: Rifle base wield slowdown reduced from 0.75 to 0.35
balance: Non-ZX/Rocket Launcher spec weapon wield slowdowns reduced from to 1.75 to 1 (includes smartgun). Minigun slowdown increased to 1 up from 0.75. Slowdowns otherwise unchanged.
balance: Rifle base accuracy increased from 1 to 1.1
balance: Pistol/Revolver and SMG movement penalty multiplier to one-handed accuracy reduced by 1 each.
balance: Accuracy penalty for pistol ammo removed.
/🆑


DominikPanic:
I agree with the last two changes, however the rest seems like an overkill balance-wise.

Surrealaser:
@DominikPanic the idea is to improve marine performance on offense, and shore up the weakness of the rifle which is mostly due to its lack of accuracy at the ranges it should be doing well at. By increasing mobility, we achieve this without strengthening marine defense significantly.

DominikPanic:
The marine winrate is now at it's all-time high. Every time you buff marines in some way we have to keep upping the xeno numbers. I disagree with further buffs.

Surrealaser:
@DominikPanic here's the problem: the majority of marine wins are predicated on defense, where the marines hole up in a fortification, then counterattack. It's a shit meta. I've addressed this with two things: the changes to sentry guns and the increase in boiler acid gas duration.

Rohesie:
The comment on the rifle slowdown was "anything below that doesn't change anything."
It's below now.

Surrealaser:
@Rohesie It's also a lie. I tested it to confirm this.

Rohesie:
Undocumented balance changes, big no.

PR #815, opened 2019-02-20
QC bugfix

Hughgent:
## About The Pull Request

QC was infinite in a mobs system.

## Why It's Good For The Game

Bugfix

## Changelog
:cl: Hughgent
fix: Quick-Clot and Peridaxon are no longer infinite in a mobs system.
/:cl:


code/modules/reagents/chemistry_reagents/medical.dm
```diff
@@ -599,7 +599,7 @@ datum/reagent/medicine/synaptizine/overdose_crit_process(mob/living/M, alien)
 		for(var/datum/internal_organ/I in H.internal_organs)
 			if(I.damage)
```
Ghommie: `if(I.damage && I.damage <= 5 && M.bodytemperature <= 169)`, and remove the two lines below.
In the current state it'll return whenever an organ has more than 5 damage, instead of skipping over.

PR #819, opened 2019-02-21
Hunter Stealth Tweaks and Fixes

Surrealaser:
:cl: by Surrealistik
balance: Hunter max stealth transparency reduced from 95% to 90%. Subject to tweaking.
balance: Hunter sneak attack damage multipliers halved while on run intent, unless stationary for 2+ seconds.
fix: Being on fire removes stealth.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -99,17 +99,17 @@
 		return
 	//Initial stealth
 	if(last_stealth > world.time - HUNTER_STEALTH_INITIAL_DELAY) //We don't start out at max invisibility
-		alpha = HUNTER_STEALTH_RUN_ALPHA //50% invisible
+		alpha = HUNTER_STEALTH_RUN_ALPHA
 		return
 	//Stationary stealth
 	else if(last_move_intent < world.time - HUNTER_STEALTH_STEALTH_DELAY) //If we're standing still for 4 seconds we become almost completely invisible
-		alpha = HUNTER_STEALTH_STILL_ALPHA //90% invisible
+		alpha = HUNTER_STEALTH_STILL_ALPHA
 	//Walking stealth
 	else if(m_intent == MOVE_INTENT_WALK)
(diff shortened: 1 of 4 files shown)


AdamElTablawy:
The issue has never really been that hunters are terribly hard to see so much as it is that hunters can hit at maximum strength when running in and out. I crippled a marine push less by acting like an organic landmine and being virtually unnoticeable and more by just sneak attacking them every three seconds, even though it wasn't sneaky at all-because I cracked bones with every hit.

Rohesie:
Agreed with Adam. This doesn't really solve much. Making them just _a little_ visible just turns the game into a matter of which tiles they'll blend in and who squints harder at the monitor or has better pattern recognition. Not really engaging.

The problem is stealth attack and this does nothing to improve it.

Surrealaser:
@Rohesie being able to spot a Hunter if you're looking for one is definitely a significant and positive change, as does forcing them to be more intelligent about where they choose to hide.

I am investigating reducing the sneak attack multiplier while running.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -103,10 +103,10 @@
 		return
 	//Stationary stealth
 	else if(last_move_intent < world.time - HUNTER_STEALTH_STEALTH_DELAY) //If we're standing still for 4 seconds we become almost completely invisible
-		alpha = HUNTER_STEALTH_STILL_ALPHA //95% invisible
+		alpha = HUNTER_STEALTH_STILL_ALPHA //90% invisible
 	//Walking stealth
 	else if(m_intent == MOVE_INTENT_WALK)
-		alpha = HUNTER_STEALTH_WALK_ALPHA //80% invisible
+		alpha = HUNTER_STEALTH_WALK_ALPHA //85% invisible
```
spookydonut: remove these % comments and just leave the ones on the #define so you dont need to update the comments everytime the define changes

Borisvanmemes:
Hunters can no longer break bones in a single slash while running up to a marine, i think this is the perfect nerf now.

PR #821, opened 2019-02-21
Preferences code cleanup.

### LINKED ISSUE #796: Spawning Slot Error
## Reproduction: Having a period "." in the user name/ckey supposedly might have something to do with this.  

Effect: Spawning as a marine PFC no matter what the character slot preferences are customized to, including xeno only preference.

### LINKED ISSUE #812: Loadout Issue: Unable to spawn with beret/cap
## Reproduction:
If your selected class/role already spawns with headgear then your selected loadout-item won't spawn. Please change it so the loadout items spawn in your hands or something... I just want a beret!

DominikPanic:
:cl: LaKiller8
code: Preferences code has been made much better now. As a side effect you will have to set all of your preferences again because I was not able to sensibly convert that mess. The defaults should also be better for all players (no more Queen and Survivor on default).
/:cl:
Fixes #812 
Fixes #796


code/modules/client/preferences.dm
```diff
@@ -416,857 +361,473 @@ datum/preferences
...
+	jobs_high = NOFLAGS
+	jobs_medium = NOFLAGS
+	jobs_low = NOFLAGS
 
 
-/datum/preferences/proc/GetPlayerAltTitle(datum/job/job)
-	if(player_alt_titles)
-		. = player_alt_titles[job.title]
+/datum/preferences/proc/GetJobDepartment(var/datum/job/job, var/level)
+	if(!job || !level)
```
Rohesie:
```suggestion
	if(!level)
```

code/modules/client/preferences.dm
```diff
@@ -416,857 +361,473 @@ datum/preferences
...
+		return FALSE
+	if(!job.prefflag)
```
Rohesie:
```suggestion
	if(!job?.prefflag)
```

PR #824, opened 2019-02-21
Humans can no longer cheesily kill xenos from the inside in 10 seconds.

### LINKED ISSUE #763: Devoured humans can slash xenos from inside
## Testmerges:

Not affected by test-merges.

## Reproduction:

Devours a human, human can use their knife to hit you while being invulnerable until you spit them out.

DominikPanic:
:cl: LaKiller8
fix: Humans can't slash xenos from the inside anymore.
/:cl:
Fixes #763

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -26,26 +26,6 @@
 			for(var/mob/M in hearers(4, src))
 				if(M.client)
 					M.show_message("<span class='warning'> You hear something rumbling inside [src]'s stomach...</span>", 2)
-		var/obj/item/I = user.get_active_held_item()
-		if(I && I.force)
-			var/d = rand(round(I.force / 4), I.force)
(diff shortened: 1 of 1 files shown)


Alterist:
this is intended behaviour on tg tho

DominikPanic:
> 
> 
> this is intended behaviour on tg tho

@Rohesie Told me it's a bug (which is why I've reported it)

Ghommie:
Despite this possibly being a bad feature, it'd be better if this were replaced with a "resist to crawl out" do_after of 30 seconds or someting, lest we want the insider to be more of a sitting duck.

Rohesie:
The insider is already auto-spat after two minutes. We can reduce that timer if needed, but it would be a different issue.

This is a feature that bugged out in CM code and they never fixed it. So the bug became the fix.

Rohesie:
Damn, I thought this PR was older. Merged it too fast. Oh well.

PR #833, opened 2019-02-22
Tweaked evolution so that larvas don't count towards the total of xenos.

Rohesie:
Did minor refactors and micro-optimizations.

## Why It's Good For The Game

Micro optimizations and a tiny refactor to make the equation more clear, after some algebraic manipulation.

## Changelog
:cl:
balance: Larvas no longer count towards xeno numbers for evolution purposes. This makes the behavior a bit more uniform and predictable.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Evolution.dm
@@ -149,10 +149,10 @@
 						to_chat(src, "<span class='warning'>You shouldn't see this. If you do, bug repot it! (Error XE01).</span>")
 
 						continue
-		if(tier == 1 && ( (tierB + tierC) > tierA )
+		if(tier == 1 && ( (tierB + tierC) > tierA ) )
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/Evolution.dm
```diff
@@ -149,13 +149,10 @@
...
 						continue
-				if(M.client && M.ckey)
-					totalXenos++
-
-		if(tier == 1 && ((tierB + tierC) / max(totalXenos, 1))> 0.5)
+		if(tier == 1 && ( (tierB + tierC) > tierA ) )
 			to_chat(src, "<span class='warning'>The hive cannot support another Tier 2, wait for either more aliens to be born or someone to die.</span>")
 			return
-		else if(tier == 2 && (tierC / max(totalXenos, 1))> 0.25)
+		else if(tier == 2 && ( (tierC * 3) > (tierA + tierB) ) )
```
spookydonut: the 3 is a magic number thats used below, should be a define to avoid them becoming different values

Rohesie:
Turned into a define as requested.

PR #666, opened 2019-02-12
RankChanges3

Karl-Franz-I:
:cl: Rory & Franz
add: Tank Commander & Field Officer Role, replaces executive officer
tweak: Existing ranks redone.
tweak: Changed round start job messages, equipment, and skills for officers.
tweak: Item descriptions changed to match new ranking system.
add: new variant of m4a3 holster with m1911 and mags inside.
/:cl:

It makes the ranking structure more in line with a realistic standard while also not shifting jobs all over and causing chaos. Forms a consistent base for any future marine jobs added. Adds a dedicated pair of officer boots to the ground.

PR #838, opened 2019-02-22
Changes assigned_role to be a /datum/job instead of a string

spookydonut:
## About The Pull Request
This changes mind.assigned_role to be job datums instead of strings

## Why It's Good For The Game
This will simplify a lot of code that has to deal with assigned_role.  Also it will eliminate more string ids that are prone to undetectable bugs from misspellings.  Seeing the changes required by #666/#641/#539 was a heavy inspiration for this, eg; there shouldn't be hardcoded lists of names

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -129,6 +129,7 @@
 #define JOB_DISPLAY_ORDER_SUQAD_ENGINEER 21
 #define JOB_DISPLAY_ORDER_SQUAD_MARINE 22
 
+#define JOBS_MARINE_ROLES (J_FLAG_MARINE|J_FLAG_SHIP)
 
 #define JOBS_COMMAND 		list("Commander", "Executive Officer", "Staff Officer", "Pilot Officer", "Tank Crewman", "Corporate Liaison", "Requisitions Officer", "Chief Engineer", "Chief Medical Officer", "Synthetic")
 #define JOBS_POLICE			list("Chief MP", "Military Police")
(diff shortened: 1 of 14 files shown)


code/controllers/stonedmc/subsystem/job.dm
```diff
@@ -61,31 +61,30 @@ SUBSYSTEM_DEF(job)
 	return type_occupations[jobtype]
 
 
-/datum/controller/subsystem/job/proc/AssignRole(mob/new_player/player, rank, latejoin = FALSE)
-	JobDebug("Running AR, Player: [player], Rank: [rank], LJ: [latejoin]")
-	if(player?.mind && rank)
-		var/datum/job/job = GetJob(rank)
+/datum/controller/subsystem/job/proc/AssignRole(mob/new_player/player, datum/job/job, latejoin = FALSE)
+	JobDebug("Running AR, Player: [player], Rank: [job?.title], LJ: [latejoin]")
+	if(player?.mind)
```
Menshin: If we remove `rank`, we need a way to prevent abstract jobs (e.g `datum/job/civilian`) to be assigned.

code/controllers/stonedmc/subsystem/job.dm
```diff
@@ -213,7 +212,7 @@ SUBSYSTEM_DEF(job)
 					// If the job isn't filled
 					if((job.current_positions < job.spawn_positions) || job.spawn_positions == -1)
 						JobDebug("DO pass, Trying to assign Player: [player], Level:[level], Job:[job.title]")
-						if(AssignRole(player, job.title))
+						if(AssignRole(player, job))
```
Menshin: Will make everyone share the same `datum/job` (the one in `all_jobs` list). Is this ok ?

code/__DEFINES/jobs.dm
```diff
@@ -129,14 +129,15 @@
 #define JOB_DISPLAY_ORDER_SUQAD_ENGINEER 21
 #define JOB_DISPLAY_ORDER_SQUAD_MARINE 22
 
+#define JOBS_MARINE_ROLES (J_FLAG_MARINE|J_FLAG_SHIP)
```
DominikPanic: You should keep the naming scheme consistent with the rest `JOBS_REGULAR_ALL`

code/controllers/stonedmc/subsystem/job.dm
DominikPanic: Not really that different from current behavior that's just one abstraction away from sharing the same ref.

code/game/gamemodes/colonialmarines/colonialmarines.dm
```diff
@@ -65,6 +64,7 @@
 /datum/game_mode/colonialmarines/post_setup()
 	initialize_post_queen_list()
 	initialize_post_xenomorph_list()
+	initialize_starting_survivor_list()
```
DominikPanic: Curious change, the gamemode code is ran before the jobs code, won't this mean there won't be any minds to pick from?

PR #844, opened 2019-02-22
Slow debuff bugfix

Surrealaser:
Slow debuff fixed, and working as it should for humans.


code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -173,12 +173,17 @@
 
 			if(M.stealth_router(HANDLE_STEALTH_CHECK)) //Cancel stealth if we have it due to aggro.
 				if(M.stealth_router(HANDLE_SNEAK_ATTACK_CHECK)) //Pouncing prevents us from making a sneak attack for 4 seconds
-					damage *= 3.5 //Massive damage on the sneak attack... hope you have armour.
+					if(m_intent == MOVE_INTENT_RUN)
```
DominikPanic: Document this please.

PR #855, opened 2019-02-23
Fax cooldown fix

Shadow-Quill:
## About The Pull Request

Fixes the fax cooldown, and uses a callback in favor of a spawn.
Also adds the CIC as a department (CIC will no longer recieve CL faxes), which I'll PR in once/if the mapping freeze is thawed.

## Why It's Good For The Game

**FAX: ThisIsntAPlayer/(Space Man) (REPLY): Receiving 'ASSFAX' via secure connection (x121904)**
is fixed.

## Changelog
:cl:
fix: Fax machine cooldown (two minutes). No more spamming faxes!
/:cl:


code/game/machinery/fax.dm
```diff
@@ -159,9 +160,15 @@
 		to_chat(user, "<span class='notice'>You [anchored ? "wrench" : "unwrench"] \the [src].</span>")
 
 
+/obj/machinery/faxmachine/proc/end_cooldown()
+	sendcooldown = FALSE
+
+
+/obj/machinery/faxmachine/cic
+	department = "Combat Information Center"
```
DominikPanic: This won't work without changing the rest of the fax code.

Shadow-Quill: Could you explain?

DominikPanic: Check the topic.dm on faxreply and faxcreate and check the other faxes file in the admin folder.

PR #860, opened 2019-02-23
Various admin and a few runtimes tweaks.

DominikPanic:


code/modules/admin/fun_verbs.dm
```diff
@@ -1111,3 +1111,37 @@ GLOBAL_LIST_EMPTY(custom_outfits)
...
+		L.taken = FALSE
+		var/mob/dead/observer/ghost = L.ghostize(FALSE)
+		if(ghost)
+			ghost.timeofdeath = world.time
+	else if(L.taken)
+		if(alert("This mob has been offered, are you sure you want to proceed?", "Warning", "Yes", "No") != "Yes")
+			return
+		L.taken = FALSE
+
+	if(alert("Are you sure?", "Offer Mob", "Yes", "No") != "Yes")
```
Rohesie: Is this second alert necessary?

code/modules/mob/dead/observer/observer.dm
```diff
@@ -57,6 +57,27 @@
...
+			to_chat(usr, "<span class='warning'>That mob has already been taken.</span>")
+			return
+
+		target.taken = TRUE
+
+		log_admin("[key_name(usr)] has taken [key_name_admin(target)].")
+		message_admins("[ADMIN_TPMONTY(usr)] has taken [ADMIN_TPMONTY(target)].")
+
+		mind.transfer_to(target, TRUE)
+		target.fully_replace_character_name(real_name)
```
Rohesie: This should fail due to having no new name assigned.

code/modules/admin/fun_verbs.dm
DominikPanic: It's a first alert technically and yeah a popup is necessary with this dangerous verb.

PR #864, opened 2019-02-23
Crusher tweaks

Surrealaser:
:cl: by Surrealistik
balance: Crusher charge damage increased vs barricades, tanks and razorwire (it can destroy barbed barricades in 2 full speed charges now); damage is dependent on ramming speed. Crushers should be able to continue charging through razorwire if colliding at full speed.
/:cl:


code/game/objects/structures/razorwire.dm
```diff
@@ -225,14 +225,15 @@
...
+		health -= C.charge_speed * CRUSHER_CHARGE_RAZORWIRE_MULTI
 		update_health()
 
 		var/def_zone = ran_zone()
 		if(C.charge_speed >= C.charge_speed_max)
 			C.visible_message("<span class='danger'>[C] plows through the barbed wire!</span>",
 			"<span class='danger'>You plow through the barbed wire!</span>", null, 5)
-		else //If we didn't destroy the barbed wire, we get tangled up.
+
+		else if(src && health > 0) //If we didn't destroy the barbed wire, we get tangled up.
```
Rohesie: The `update_health()` should happen at the end, else this proc will have errors with its source deleted.
Then you won't need this check.

code/modules/mob/living/carbon/xenomorph/Castes/Crusher.dm
```diff
@@ -249,13 +249,13 @@
 	if(istype(target, /obj/structure/razorwire))
 		var/obj/structure/razorwire/B = target
 		if(charge_speed >= charge_speed_max) //plows right through
-			flags_pass = 0
+			flags_pass = TRUE
```
spookydonut:
```suggestion
			flags_pass |= PASSTABLE
```
passtable is defined as 1 which is why this works but use the proper flag define for the bitfield

code/modules/mob/living/carbon/xenomorph/Castes/Crusher.dm
```diff
@@ -249,13 +249,13 @@
...
 			update_icons()
 			return TRUE
 		else if(charge_speed > charge_speed_buildup * charge_turfs_to_charge)
 			visible_message("<span class='danger'>[src] rams into [B] and skids to a halt!</span>",
 			"<span class='xenowarning'>You ram into [B] and skid to a halt!</span>")
-			flags_pass = 0
+			flags_pass = FALSE
```
spookydonut:
```suggestion
			flags_pass &= ~PASSTABLE
```

Rohesie:
This is tested. When the requests are done it can be merged, if no other changes happen.

PR #873, opened 2019-02-24
Fixes a weird bug with fuel logs on start up

spookydonut:
No clue why this fixes it but it does

### DIFF
--- a/code/game/objects/effects/decals/Cleanable/fuel.dm
+++ b/code/game/objects/effects/decals/Cleanable/fuel.dm
@@ -1,4 +1,4 @@
-obj/effect/decal/cleanable/liquid_fuel
+/obj/effect/decal/cleanable/liquid_fuel
 	//Liquid fuel is used for things that used to rely on volatile fuels or phoron being contained to a couple tiles.
 	icon = 'icons/effects/effects.dmi'
 	icon_state = "fuel"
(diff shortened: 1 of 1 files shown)


DominikPanic:
>No clue why this fixes it but it does
Ninja will whip me if I merge this

code/game/objects/effects/decals/Cleanable/fuel.dm
```diff
@@ -1,60 +1,61 @@
...
+		log_game("[amt] units of liquid fuel have spilled in [AREACOORD(newLoc.loc)].")
+		message_admins("[amt] units of liquid fuel have spilled in [ADMIN_VERBOSEJMP(newLoc.loc)].")
+	amount = amt
+
+	//Be absorbed by any other liquid fuel in the tile.
+	for(var/obj/effect/decal/cleanable/liquid_fuel/other in newLoc)
+		if(other != src)
+			other.amount += src.amount
+			spawn other.Spread()
+			qdel(src)
```
Menshin: Should probably `return` after `qdel`ing.

code/game/objects/effects/decals/Cleanable/fuel.dm
```diff
@@ -1,60 +1,61 @@
...
+	for(var/d in list(turn(dir,90),turn(dir,-90), dir))
+		var/turf/O = get_step(S,d)
+		if(locate(/obj/effect/decal/cleanable/liquid_fuel/flamethrower_fuel) in O)
+			continue
+		if(O.CanPass(null, S) && S.CanPass(null, O))
+			new/obj/effect/decal/cleanable/liquid_fuel/flamethrower_fuel(O, amount * 0.25, 1)
 //					O.hotspot_expose((T20C * 2) + 380, 500) //Light flamethrower fuel on fire immediately.
 
-			amount *= 0.25
+	amount *= 0.25
```
Menshin: Intended indentation ?

spookydonut: its in line with the for loop above it still.

code/game/objects/effects/decals/Cleanable/fuel.dm
spookydonut: good catch

code/game/objects/effects/decals/Cleanable/fuel.dm
Menshin: So, whether it succeed to spread or not, it loses 75% of it content ?

Probably was intended to be :
```suggestion
			amount *= 0.75
```

code/game/objects/effects/decals/Cleanable/fuel.dm
```diff
@@ -1,60 +1,61 @@
...
+			other.amount += src.amount
+			spawn other.Spread()
+			qdel(src)
+
+	Spread()
+	. = ..()
+
+/obj/effect/decal/cleanable/liquid_fuel/proc/Spread()
+	//Allows liquid fuels to sometimes flow into other tiles.
+	if(amount < 5.0) return
```
Rohesie: Same line control statement.
There's also others down there.

code/game/objects/effects/decals/Cleanable/fuel.dm
spookydonut: Yes, if you use side-by-side diffs you'll see this line was the same indent as line 52/53 before and is still the same indent now

Rohesie: It's reduced to 25% after spreading into 3 tiles (`list(turn(dir,90),turn(dir,-90), dir)`), 25% in each, for a total of 100%.
Works differently than the other fuel.

Menshin: Except it is reduced to this amount whether it succeed in spreading or not.
So any `/obj/effect/decal/cleanable/liquid_fuel/flamethrower_fuel` is instanciated with 25% of its amount even if it didn't spread (using against a wall, throwing fuel over fuel, etc).
Anyway, it's not critical, it can stay that way for now.

PR #900, opened 2019-02-26
OB fix

Zostroll:
## About The Pull Request
This fixes a bug where you can chamber the OB tray without loading the tray causing it to not use ammo or start the timer allowing you to use the OB as a machine gun.

## Why It's Good For The Game
Stops machine gun OB

## Changelog
🆑
fix: Makes the OB not be able to be fired like a machine gun.
/:cl:

### DIFF
--- a/code/modules/cm_marines/orbital_cannon.dm
+++ b/code/modules/cm_marines/orbital_cannon.dm
@@ -142,8 +142,7 @@ var/obj/structure/ship_rail_gun/almayer_rail_gun
 	set waitfor = 0
 
 	if(!loaded_tray)
-		if(user)
-			to_chat(user, "<span class='warning'>You need to load the tray before chambering it.</span>")
+		to_chat(user, "<span class='warning'>You need to load the tray before chambering it.</span>")
 		return
 	
 	if(ob_cannon_busy)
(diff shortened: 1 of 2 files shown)


code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -141,6 +141,11 @@ var/obj/structure/ship_rail_gun/almayer_rail_gun
 /obj/structure/orbital_cannon/proc/chamber_payload(mob/user)
 	set waitfor = 0
 
+	if(!loaded_tray)
+		if(user)
+			to_chat(user, "You need to load the tray before chamber it.")
```
Shadow-Quill:
```suggestion
			to_chat(user, "<span class='warning'>You need to load the tray before chambering it.</span>")
```
Makes it red and noticible (With grammar :D)

code/modules/cm_marines/orbital_cannon.dm
```diff
@@ -141,6 +141,11 @@ var/obj/structure/ship_rail_gun/almayer_rail_gun
 /obj/structure/orbital_cannon/proc/chamber_payload(mob/user)
 	set waitfor = 0
 
+	if(!loaded_tray)
+		if(user)
```
DominikPanic: User should always exist, if not that's a problem of other pieces of the OB code and should be handled there so you can safely remove this.

Zostroll: Can I remove it in the other instances in the OB code without breaking it just to make it more streamline?

PR #903, opened 2019-02-26
Praetorian Acid Spray Now Goes Past Razorwire

Surrealaser:
:cl: by Surrealistik
fix: Acid Spray now goes past razorwire.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Powers.dm
@@ -184,15 +184,17 @@
 					var/obj/structure/barricade/B = O
 					B.health -= rand(40,60) + 8 * upgrade
 					B.update_health(1)
+					return
 				else if(istype(O, /obj/vehicle/multitile/root/cm_armored) )
 					var/obj/vehicle/multitile/root/cm_armored/A = O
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/Powers.dm
```diff
@@ -276,7 +282,11 @@
 							var/obj/vehicle/multitile/root/cm_armored/A = O
 							A.take_damage_type(rand(40,60) + 8 * upgrade, "acid", src)
 							A.healthcheck()
-						normal_density_flag = 1
+						else if(istype(O, /obj/structure/razorwire) )
```
DominikPanic: Another snowflake checks that should be handled by components.

Surrealaser: Well this is currently the norm unfortunately.

Surrealaser: If this is cause for barring the PR, we should just delete the entire ability.

spookydonut: We dont have components yet so this is fine

PR #905, opened 2019-02-26
Sulaco's first pass

Rohesie:
## About The Pull Request

You can load the maps and even play in them if you load them as ground maps.
It requires a map cleanup, landmarks and some shuttle code to be otherwise playable.

Ended up cleaning a few other files in the process, updating old object references with current ones.

## Why It's Good For The Game

Adds some options.


spookydonut:
You need to mapmerge any maps before committing changes to them

PR #928, opened 2019-02-28
Adds legacy job ban debugging

Rohesie:
* Faxes are fun.
* Changing hive-numbers as well, and micro-optimized.
* Legacy job bans can be checked and transferred across job names.


code/modules/admin/debug_verbs.dm
```diff
@@ -542,3 +496,55 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
 
 	log_admin("[key_name(usr)] updated the mob sprite of [key_name(H)].")
 	message_admins("[ADMIN_TPMONTY(usr)] updated the mob sprite of [ADMIN_TPMONTY(H)].")
+
+
+/datum/admins/proc/transfer_job_bans()
```
DominikPanic: Make this a /proc/ in ban_panel.dm. There is no reason this should be a regular button as it will at most be used once.

DominikPanic: Scratch that, this PR is on hold until my DB stuff gets merged, then it will get added into the legacy ones.

DominikPanic:
Add this to legacy.dm, don't add it to holder.dm (as it's for the legacy system not in the active rotation)

PR #933, opened 2019-03-01
Tank Now Emits a Targeting Laser When Zoomed and Firing LTB and TOW

Surrealaser:
:cl: by Surrealistik
balance: LTB fire delayed for 2 seconds while zoomed in. TOW fire delay now 0.3 seconds unzoomed and 1.5 seconds zoomed. During this time both emit a targeting laser at the target turf while zoomed.
/:cl:


code/modules/vehicles/hardpoints.dm
```diff
@@ -180,7 +180,20 @@ Currently only has the tank hardpoints
...
+
+	var/obj/vehicle/multitile/root/cm_armored/tank/C = owner
+	var/obj/effect/overlay/temp/tank_laser/TL
+	if(C.is_zoomed)
+		delay = 20
+		TL = new /obj/effect/overlay/temp/tank_laser (T)
+
+	to_chat(usr, "<span class='warning'>Preparing to fire... keep the tank still for [delay * 0.1] seconds.</span>")
+
+	if(!do_after(usr, delay, FALSE, 5, BUSY_ICON_HOSTILE))
```
Rohesie:
```suggestion
	if(!do_after(usr, delay, FALSE, 5))
```
There is no need for busy icon as it won't be seen from inside the tank.

PR #937, opened 2019-03-02
Explosive Damage Rework

Surrealaser:
:cl: by Surrealistik
balance: Explosions now apply stagger and slow to Xenos and humans. The amount is higher at higher explosion powers, and is lowered by explosion resistance/Bomb armor.
balance: Damage and debuffs from explosions is reduced by a % equal to a human's Bomb armor. Initial damage has been reduced accordingly. Explosions can no longer gib humans in keeping with policy to minimize permadeath events. X-treme explosions will still probably instantly kill you dead though.
/:cl:


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -201,38 +201,46 @@
...
-				ear_damage += 30
-				ear_deaf += 120
-			if(prob(70))
-				KnockOut(10)
+				ear_damage += 60 * armor
+				ear_deaf += 240 * armor
+
+			adjust_stagger(round(4 * armor,0.1))
+			add_slowdown(round(4 * armor,0.1))
+			KnockOut(10 * armor)
```
Rohesie: While `KnockOut` is realistic, I'd personally prefer we moved away from it in favor of `KnockDown` here. Let pain or crit handle losing consciousness instead.
```suggestion
			KnockDown(10 * armor)
```
Maybe a little feedback if they are still `CONSCIOUS` to inform them they'll be out of commission for a bit, as well, would be good.
Something like
```
if(stat)
	tochat(... "You fail to see the plate number of the truck that hit you.")
```

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -201,38 +201,46 @@
...
+			adjust_stagger(round(2 * armor,0.1))
+			add_slowdown(round(2 * armor,0.1))
+			KnockOut(7 * armor)
 
 		if(3)
-			b_loss += rand(50, 70)
-			if(prob(getarmor(null, "bomb")))
-				b_loss = b_loss/2
+			b_loss += rand(30, 40) * armor
+			b_loss += rand(30, 40) * armor
```
Rohesie: This is repeated. You can substitute it with a single `b_loss += rand(60, 80) * armor` if you want to preserve the change.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -12,21 +12,27 @@
 		if(1)
 			switch(xeno_explosion_resistance)
 				if(3)
-					apply_damage(rand(125, 175), BRUTE) //Used to be rand(200, 300) which would one-shot most things. Tanks have severity 1 explosions on direct hits...
-					updatehealth()
+					b_loss += rand(70, 80)
+					f_loss += rand(70, 80)
+					adjust_stagger(1.1)
```
Rohesie: Stagger is done in integer additions/substractions. Why the `.1`? Use `1` or `2` for the same effect.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -12,21 +12,27 @@
...
+					add_slowdown(1.1)
```
Rohesie: Same for slowdown, it's an integer paired with a `0.3` define. This would increase it in `0.33`, to little effect.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -12,21 +12,27 @@
...
 				if(2)
 					KnockDown(6)
-					apply_damage(rand(125, 175), BRUTE)
-					updatehealth()
+					adjust_stagger(2.25)
+					add_slowdown(2.25)
```
Rohesie: Same on these two.

PR #944, opened 2019-03-02
Defiler Neuroclaw Inject Amount Increased to Offset Faster Metarate

Surrealaser:
:cl: by Surrealistik
tweak: The amount of neurotox injected by the Defiler's neuroclaws has been increased by 0.5 at each upgrade level to offset the higher neuro meta rate.
/:cl:

PR #952, opened 2019-03-03
Xeno stings respect OD thresholds, abilities tweaked

Rohesie:
## Changelog
:cl:
balance: The Queen's Larval Advancement skill cooldown reduced from 60 to 15 seconds. Don't get caught near a queen.
balance: Defiling ability changed. It no longer injects neurotoxin after implanting a larva, only growth serum, besides causing minor brute damage to the chest and decent pain damage to the victim. In compensation the Defiler gains both neuro and larval advancement serum stings. More versatility, less raw power.
tweak: Xeno stings inform the user if their victim is ODing to reduce accidental kills. They will start jittering earlier as well.
balance: Larval growth serum and neurotoxin injection abilities inject 10 and 15u respectively, three times, the first one instant. ODs for them tweaked: soft OD (between 30 and 50u) has low lethality while hard ODs (over 50) is almost certain death. Damage is oxygen, though, so it's easily countered by medics.
/:cl:


Ketrai:
I wouldn't remove larval serum from the defiler sting. This will heavily encourage the tedium of stinging captured hosts. If xenos capture five marines, you have to sting each 10 times, which is annoying.

Surrealaser:
@Rohesie This needs significant changes.

#1: Make a sting shift function (like the shift spits) so the Defiler can choose whether his sting injects Neuro, Growth Toxin + Larva (these should not be separated out; too clumsy), or both.

#2: Defiler should be able to OD with his Neurosting, but receive warnings about it.

#3: Neurosting by itself should have a shorter cooldown than Growth Toxin + Larva, or both.

DominikPanic:
As per my comment in #944 talk with Surr instead of reverting his changes with PRs

Rohesie:
@Surrealaser 

`1: Make a sting shift function (like the shift spits) so the Defiler can choose whether his sting injects Neuro, Growth Toxin + Larva (these should not be separated out; too clumsy), or both. Alternately separate out the stings, but add a shared cooldown between the all-in-one and Growth + Larva and Neuro.`

Second option done. Defile ability now injects growth serum after the larva is implanted. Defilers have both Defile and Serum stings. Since the Defile proc triggers the Serum one, there will be cooldown for both, each on their own time..

`2: Defiler should be able to OD with his Neurosting, but receive warnings about it.`

Choose between instant injection and warning. Else you have accidental ODs.

`3: Neurosting by itself should have a shorter cooldown than Growth Toxin + Larva, or both.`

Give me values. For both Defiler and Sent neurostings if you want to differentiate them.

`4: Reinstate the instant sting 15u injection component immediately; removing this is a massive, unwarranted nerf.`

Reinstated.

Surrealaser:
@Rohesie Alright.

#1: To be clear, we have a neurotoxin sting, a growth toxin sting, a growth toxin + larva sting, and an all-in-one sting, right?

#2: Warning and instant injection is fine. Given the speed of the meta rate, how obvious existing OD is, and the fact that the deadly OD requires 60u, the worst case scenario of a 15U basic overdose isn't really lethal at all. Further, I have plans to add HUD indicators for the presence of neuro/growth toxin, but feel free to include those with this PR if you want.

#3: 12 seconds for Neurotox and Growth Toxin only. 20 seconds for larval injection. We can tweak these later.

Rohesie:
`1: To be clear, we have a neurotoxin sting, a growth toxin sting, a growth toxin + larva sting, and an all-in-one sting, right?`

Defiler has 3 powers:
1) Neurotox sting: injects neurotox, cooldown 12 seconds.
2) Larval growth sting: injects larval growth, cooldown 12 seconds.
3) Defile: Implants larva (will move to the next step if already a host), then calls for Larval growth sting (will fail that second part if that one is in CD), cooldown 20 seconds.

`2: Warning and instant injection is fine. Given the speed of the meta rate, how obvious existing OD is, and the fact that the deadly OD requires 60u, the worst case scenario of a 15U basic overdose isn't really lethal at all. Further, I have plans to add HUD indicators for the presence of neuro/growth toxin, but feel free to include those with this PR if you want.`

That would add nothing but killing hosts in bad shape by mistake by new xenos. It's not like if a xeno wants to kill a marine a neuroinjection would be the first, second or third choice. It's a tool for subduing instead. If the marine has 30u in them then there's plenty of ways to kill them.

Let's not discuss this forever. This system avoids accidental deaths and turns the sting into a safe tool for all the family. I won't oppose allowing it to OD with a warning as long as the first injection is timed as well. Then there's no risk of accidental ODs.

`3: 12 seconds for Neurotox and Growth Toxin only. 20 seconds for larval injection. We can tweak these later.`

Done as requested.

Surrealaser:
> That would add nothing but killing hosts in bad shape by mistake by new xenos. It's not like if a xeno wants to kill a marine a neuroinjection would be the first, second or third choice. It's a tool for subduing instead. If the marine has 30u in them then there's plenty of ways to kill them.

No.

The most common source of dangerous ODing was via neuro spit; that is no longer a thing. Right now it caps at 30U, and 15U from the instant injection isn't enough to put someone at dangerous OD levels. Hell, it takes the entirety of the injection to get someone to dangerous OD levels from 30u to under 65u, and at current meta rates, that means the victim will suffer at most ~3 ticks of suffocation.

Meanwhile, it is painfully obvious whether or not someone is ODed in corner cases like gas or another neurosting between shaking and pain emotes. Worst case we can add HUD icons that indicate the presence/status of neuro and growth toxin, which is a better solution than expanding artificial caps.

As to the utility of deliberate ODing, as I and others have stated repeatedly, it is to increase the time a host is disabled in the event that processing takes a longer period of time.

Rohesie:
All requests done, conflicts solved.

PR #957, opened 2019-03-03
Another Admin PR

### LINKED ISSUE #968: Taking an offered human makes you use your old skills, not the skills that human should have.
Basically, if you're a squad medic or doctor and you become a tank crewmember, you'll have the medical skills from being a doctor instead of the skills of a tank crewmember.

DominikPanic:
-Proper tank damage logging
-Supply pack logging
-~~Forum account linking~~ depends on MSO
-Debug controllers removed in favor of statcliks
-Rearranges a lot of the verbs
-Improves a lot of the verbs
-Adds blinking screen for PMs and new round start
-Makes interfaces pretty
-Minor bugfixes
-Pretty job slots button

Fixes #968


code/modules/admin/admin_verbs.dm
```diff
@@ -1148,7 +1237,8 @@
 		if(isobserver(C.mob) || C.mob.stat == DEAD)
 			continue
 		if(ishuman(C.mob))
-			C.mob.loc = get_turf(usr)
+			var/mob/living/carbon/human/H = C.mob
```
Rohesie: There is no need to type cast this, as `forceMove()` is defined at the mob level.
This could be improved by replacing the observer check with a `if(!ishuman(C.mob))` and thus avoiding double checks.

code/modules/admin/admin_verbs.dm
```diff
@@ -1159,7 +1249,8 @@
 		if(isobserver(C.mob) || C.mob.stat == DEAD)
 			continue
 		if(isxeno(C.mob))
-			C.mob.loc = get_turf(usr)
+			var/mob/living/carbon/Xenomorph/X = C.mob
```
Rohesie: Same here, substitute `isobserver(C.mob)` with `!isxeno(C.mob)`, remove the type cast and avoid unnecessary checks.
If the compiler complains about the proc not being defined for the variable just type cast it at the beginning and use a `!istype()`early continue or something.

code/modules/admin/debug_verbs.dm
```diff
@@ -469,30 +437,10 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
 	var/dat = "<b>Contents of [key_name(M)]:</b><hr>"
 
 	var/list/L = M.get_contents()
-	for(var/t in L)
-		dat += "[t]<br>"
+	for(var/atom/A in L)
```
Rohesie: This should only return atoms (items in fact). Type cast it outside to avoid unnecessary checks.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -76,6 +77,15 @@
 
 		mind.transfer_to(target, TRUE)
 		target.fully_replace_character_name(real_name, target.real_name)
+		if(target.job && target.mind)
```
Rohesie:
```suggestion
		if(target.job)
```
Theoretically there should be a mind by now.

code/modules/admin/fun_verbs.dm
```diff
@@ -439,25 +371,40 @@
...
+	set category = "Fun"
+	set name = "Stop Playing Music"
+
+	if(!check_rights(R_SOUND))
+		return
+
+	for(var/mob/M in GLOB.player_list)
+		if(M.client)
 			var/client/C = M.client
 			if(C?.chatOutput && !C.chatOutput.broken && C.chatOutput.loaded)
```
Rohesie: Since you added the `if(M.client)` check:
```suggestion
			if(C.chatOutput?.loaded && !C.chatOutput.broken)
```
Alternatively, to preserve the order of the checks turn these into an early continue instead.

code/modules/admin/fun_verbs.dm
```diff
@@ -439,25 +371,40 @@
...
+/datum/admins/proc/music_stop()
+	set category = "Fun"
+	set name = "Stop Playing Music"
+
+	if(!check_rights(R_SOUND))
+		return
+
+	for(var/i in GLOB.clients)
+		var/client/C = i
+		if(!C.chatOutput || C.chatOutput.broken || !C.chatOutput.loaded)
```
```suggestion
		if(!C?.chatOutput.loaded || C.chatOutput.broken)
```

PR #970, opened 2019-03-04
fixes mateba sprites

Alterist:
oh god oh fuck


Alterist:
TM with #972 for best results

PR #972, opened 2019-03-04
Field Commander Now starts with a M1911A1 Custom

Alterist:
:cl:
feature: The Field Commander Now starts with a M1911A1 Custom, a powerful variant of the .45 that is more powerful, faster firing and easier to handle. the CO gets his mateba back
tweak: the revolver belt can now hold all types of revolvers
/:cl:
Revolver belt can now hold all revolvers
The FC gets his own snowflake gun
fixes issues with gun belts in general


Alterist:
TM with #970 so i don't get yelled at

code/game/objects/items/storage/belt.dm
```diff
@@ -539,6 +539,15 @@
 	new /obj/item/ammo_magazine/pistol/ap(src)
 	new_gun.on_enter_storage(src)
 
+/obj/item/storage/belt/gun/m4a3/fieldcommander/New()
+	..()
```
DominikPanic:
```suggestion
	. = ..()
```

PR #997, opened 2019-03-06
Fixes crushers being able to crush fog.

### LINKED ISSUE #991: Crushers can smash through dense fog on LV
## Reproduction:
Just charge at the fog and not get slowed down, can be done easily on the eastern bridge.

DominikPanic:
:cl: LaKiller8
fix: Crushers can no longer crush fog.
/:cl:
Fixes #991

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/Castes/Crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/Castes/Crusher.dm
@@ -314,10 +314,6 @@
 				return FALSE
 
 
-/obj/effect/forcefield/charge_act(mob/living/carbon/Xenomorph/X)
-	X.stop_momentum(X.charge_dir)
-	return FALSE
-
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/Castes/Crusher.dm
```diff
@@ -313,6 +313,11 @@
 				X.stop_momentum(X.charge_dir)
 				return FALSE
 
+
+/obj/effect/forcefield/charge_act(mob/living/carbon/Xenomorph/X)
```
Rohesie: There's quite a few conflicting methods here, and while the fog is not `unacidable`, there's a specific check for `istype(I, /obj/effect)` in the acid proc.
Crushers can't crush unacidable objects.
Thus, we should simply remove that check and make all effects unacidable. No need for this proc redefinition then.

Issue #992, opened 2019-03-06
Huggers die when they jump onto the same tile.


MiniMeatwad:
## Reproduction: Bring someone to more than 1 egg and most likely the facehuggers will stack on the same tile, hug the guy and then kill each other. This could have something to do with old code made to prevent hugger stacking before it was coded in by CM to have huggers die from leaving eggs after a certain amount of time.
Possible solution is to have huggers not kill each other and also crawl back into eggs 1 tile nearby to make this issue less common. Also they might even kill each other even if the hugger on the ground is dead and the hugger is in your hand, haven't been able to find out but that might explain unexpected hugger deaths.

DominikPanic:
@Ghommie

Shadow-Quill:
Huggers currently die if: (from what I'm reading)

- There are more than two conscious, non-sterile (Lamarr) huggers on a tile, all but two will be qdeled.
- Or if they're out of an egg or a carrier's grasp for eight seconds(?), and they aren't on a tile with an empty egg.

Ghommie:
I'll probably slap in a `prob(30)` and a chance of them leaping to an adjacent tile, so they don't just die when leaping on the same turf. Thanks.

PR #1008, opened 2019-03-07
Removes the anti-stacking proc from facehuggers.

Ghommie:
## About The Pull Request
Fiddling furthermore with facehuggers. Removing unnecessary process from non-conscious spawn facehuggers. Removes check_neighbours().

## Why It's Good For The Game
Should resolve an issue in #992. Life cycles already kill facehuggers after a dozen seconds circa, so piling them up isn't much of a viable tactic anymore.

## Changelog
:cl:
del: Facehuggers don't die when piled up in groups of 3 and more on the same tile anymore.
/:cl:


DominikPanic:
Won't changing the default hugger type to stasis just not make them work at all? Have you tested it in-game with the alien egg structures that they actually work?

Ghommie:
Facehuggers within an egg/hugger trap/carrier storage already undergo stasis. The subtype is so they don't start processing on `Initialize()` just to stop right away afterward, perhaps it's not really a problem.
I'll let you know when this is 100% ok, self-confidence aside.

PR #1009, opened 2019-03-07
Job Prefs: High To Medium When Overwriting

Shadow-Quill:
## About The Pull Request

Currently, if you have one job set to `High` priority, and you attempt to set another to high, the first one will reset to `Never`. This PR changes it to the (I think?) more often decrease from `High` to `Medium`.

Does not affect setting `High` to `Never`.

image

## Why It's Good For The Game

Less hassle, I suppose.

## Changelog
:cl:
tweak: Job preferences will now go back to Medium if it was set to High, and you try to change another to High.
/:cl:

It seems like this is your baby @DominikPanic, so have a ping.


DominikPanic:
This is not my baby I preserved the original functionality without changing anything when porting the jobs subsystem.

PR #1011, opened 2019-03-07
Fixes some more runtimes.

DominikPanic:


code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -174,6 +174,8 @@
 
 
 /obj/effect/particle_effect/smoke/tactical/proc/uncloak_smoke_act(var/mob/living/M)
+	if(!istype(M))
```
Rohesie: This check should be happening in `/obj/effect/particle_effect/smoke/tactical/Uncrossed(var/mob/living/M)`
Which should actually be using `atom/movable/O` as per the reference, instead of assuming only `living` trigger it.
Same as `/obj/effect/particle_effect/smoke/tactical/Crossed(atom/movable/M)` really.

code/modules/admin/verbs/adminhelp.dm
```diff
@@ -668,7 +668,9 @@ GLOBAL_DATUM_INIT(ahelp_tickets, /datum/admin_help_tickets, new)
 //
 
 /client/proc/giveadminhelpverb()
-	src.verbs |= /client/verb/adminhelp
+	if(!src)
```
Rohesie: This shouldn't be necessary, as it's checked before calling this proc.

DominikPanic: Call proc, client leaves, this triggers, runtime.

PR #1012, opened 2019-03-07
Emote overlays.

DominikPanic:
:cl: LaKiller8
imageadd: New overlay icon when throwing greandes created by Sleepy Retard.
tweak: Emoting now has a slightly different icon. Emote icons should now also appear above the HUD for clarity.
/:cl:


code/modules/mob/mob.dm
```diff
@@ -794,8 +794,27 @@ mob/proc/yank_out_object()
...
+			continue
+		to_chat(M, emote_overlay)
+
+	if(remove_delay)
+		addtimer(CALLBACK(src, .proc/remove_emote_overlay, client, emote_overlay, viewers), remove_delay)
+
+
+/mob/proc/remove_emote_overlay(client/C, image/emote_overlay, list/viewers)
+	if(C)
+		C.images -= emote_overlay
```
Ghommie: `C?.images -= emote_overlay`, line above unneeded I guess.

code/modules/mob/mob.dm
```diff
@@ -794,8 +794,27 @@ mob/proc/yank_out_object()
...
+	if(!viewers)
+		viewers = viewers()
+	for(var/mob/M in viewers)
+		if(!M.client)
+			continue
+		M.client.images -= emote_overlay
```
Ghommie: M.client?.images -= emote_overlay, two lines above unneeded.

code/modules/mob/mob.dm
```diff
@@ -794,8 +794,27 @@ mob/proc/yank_out_object()
...
-/mob/proc/remove_emote_overlay(var/image/overlay_to_remove)
-	overlays -= overlay_to_remove
+/mob/proc/add_emote_overlay(image/emote_overlay, remove_delay = TYPING_INDICATOR_LIFETIME)
+	var/viewers = viewers()
+	for(var/mob/M in viewers)
+		if(M.stat == DEAD)
+			continue
+		to_chat(M, emote_overlay)
+
+	if(remove_delay)
```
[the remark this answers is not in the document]
DominikPanic: It's actually intended to be this way, the typing indicator has no delay and instead is removed whenever you close it again.

code/modules/mob/mob.dm
```diff
@@ -794,8 +794,27 @@ mob/proc/yank_out_object()
 	return TRUE
 
 
-/mob/proc/remove_emote_overlay(var/image/overlay_to_remove)
-	overlays -= overlay_to_remove
+/mob/proc/add_emote_overlay(image/emote_overlay, remove_delay = TYPING_INDICATOR_LIFETIME)
```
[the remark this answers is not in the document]
DominikPanic: This is for audible emotes only.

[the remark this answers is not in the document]
Ghommie: Yea, but if deafs can't hear the audible message, they shouldn't see the overlay either, no?

code/modules/mob/mob.dm
```diff
@@ -794,8 +794,27 @@ mob/proc/yank_out_object()
...
+	var/viewers = viewers()
+	for(var/mob/M in viewers)
+		if(M.stat == DEAD)
+			continue
+		to_chat(M, emote_overlay)
```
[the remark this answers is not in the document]
DominikPanic: Not sure tbh I copied this snippet over from the emote code that seems to work well.

code/modules/mob/mob.dm
```diff
@@ -794,8 +794,24 @@ mob/proc/yank_out_object()
...
+		if(!isobserver(M) && (M.stat != CONSCIOUS || isdeaf(M)))
+			continue
+		to_chat(M, emote_overlay)
+
+	if(remove_delay)
+		addtimer(CALLBACK(src, .proc/remove_emote_overlay, client, emote_overlay, viewers), remove_delay)
+
+
+/mob/proc/remove_emote_overlay(client/C, image/emote_overlay, list/viewers)
+	C?.images -= emote_overlay
```
Rohesie: This will require to be inside an `if(C)`

code/modules/mob/mob.dm
```diff
@@ -794,8 +794,24 @@ mob/proc/yank_out_object()
...
+	if(!viewers)
+		viewers = viewers()
+	for(var/mob/M in viewers)
+		M.client?.images -= emote_overlay
```
Rohesie: Same here for `M.client`

code/modules/mob/mob.dm
```diff
@@ -794,8 +794,24 @@ mob/proc/yank_out_object()
...
+			continue
+		to_chat(M, emote_overlay)
+
+	if(remove_delay)
+		addtimer(CALLBACK(src, .proc/remove_emote_overlay, client, emote_overlay, viewers), remove_delay)
+
+
+/mob/proc/remove_emote_overlay(client/C, image/emote_overlay, list/viewers)
+	C?.images -= emote_overlay
+	if(!viewers)
```
Rohesie: This should probably not be a thing. If no viewers are selected, it probably shouldn't be looking for more. This should be defined at the call level.

code/modules/mob/mob.dm
DominikPanic: Thanks Ghom

code/modules/mob/mob.dm
DominikPanic: Thanks Ghom

code/modules/mob/mob.dm
Rohesie: That'd be bad. If `C` doesn't exist that will create an operation that looks like "`null` equals something".

code/modules/mob/mob.dm
Rohesie: Same here.

code/modules/mob/mob.dm
Ghommie: Oh, right there, I forgot we had this kind of talk long time ago.

PR #1019, opened 2019-03-09
Scout Kit Weapon Rebalance

Surrealaser:
:cl: by Surrealistik
balance: Accuracy multiplier for ZX reduced by 30% when bursting; to be tweaked as necessary.
tweak: M4RA fire delay decreased from 0.5 to 0.4 seconds, dramatically increased the number of compatible attachments, base accuracy multiplier increased from 1 to 1.2x, burst amount decreased from 3 to 2, burst delay decreased from 0.2 to 0.1 seconds.
/:cl:


code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -946,6 +953,7 @@
 	accuracy_mult_unwielded = CONFIG_GET(number/combat_define/base_hit_accuracy_mult)
 	scatter = CONFIG_GET(number/combat_define/med_scatter_value)
 	scatter_unwielded = CONFIG_GET(number/combat_define/med_scatter_value)
+	burst_accuracy_mult = CONFIG_GET(number/combat_define/mlow_burst_accuracy_penalty)
```
InterroLouis: Shouldn't this be on the ZX-76 instead of the minigun?

Ketrai:
Yeah, the triple burst scout rifle was also something silly. Hopefully this will make the scout into less of a powerhouse.

Ketrai:
Also probably should buff single fire-rate. there's currently a moot tradeoff to bursting. Single fire is about as slow as a pump shotgun~

Surrealaser:
@Ketrai the single fire is fine, it's an M37 without the need to pump and a bigger ammo capacity which means it's slightly better as is. The double shot is the main appeal of the weapon, it was just too out there.

Ketrai:
scout shotgun is unfortunately dead now. you killed it sur. There's no point to taking a slightly better m37

PR #1033, opened 2019-03-11
Tank Rebalance

Surrealaser:
:cl: by Surrealistik
balance: Knockdown stacks for mobs crushed by the tank reduced from 12 to 1, and they are displaced behind the tank. Tanks do not apply knockdown stacks to knocked down mobs. Mobs crushed are displaced away from the tank.
balance: Tank takes no collision damage while running over objects/walls it is facing with the snowplow , and the damage it deals to them is doubled.
/:cl:

### DIFF
--- a/code/modules/vehicles/cm_armored.dm
+++ b/code/modules/vehicles/cm_armored.dm
@@ -420,11 +420,11 @@ var/list/TANK_HARDPOINT_OFFSETS = list(
 		H?.livingmob_interact(src)
 
 /turf/closed/wall/tank_collision(obj/vehicle/multitile/hitbox/cm_armored/C, facing, turf/T, turf/temp)
-	var/obj/vehicle/multitile/root/cm_armored/CA = C.root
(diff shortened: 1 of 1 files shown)


AdamElTablawy:
will this still result in mobs being displaced into walls and being hit ten thousand times

code/modules/vehicles/cm_armored.dm
```diff
@@ -357,9 +357,18 @@ var/list/TANK_HARDPOINT_OFFSETS = list(
 	return
 
 /mob/living/tank_collision(obj/vehicle/multitile/hitbox/cm_armored/C, facing, turf/T, turf/temp)
+	if(isdead(src)) //We don't care about the dead
```
Rohesie: This check is for dead mobs, such as ghosts. Since this proc is defined for the living, it is not necessary.
I assume you want something like `if(stat == DEAD)`

Surrealaser: This is done

PR #1045, opened 2019-03-13
Fixes glide_size modification code to actually work

ma44:
## About The Pull Request
This PR fixes a major problem where the code that modifies glide_size (a variable that determines the speed of a mob moving between tiles) was actually modifying the client datum's glide_size and using client datum's tick_lag (that's always equal zero), this fixes it by using the proper variables and setting them appropriately.

Gyazo is annoying and makes the thing a bit laggy, but still a nice improvement.

https://gyazo.com/1d364ff8b3faa051a1a02a3c5f542b0c

https://gyazo.com/81355d147cfd9d23860ee35e774e1c48

## Why It's Good For The Game
Makes the game movement more smooth than stuttering, making it more appealing to new and veteran players by being more 'modern'

## Changelog
:cl:
tweak: The time it takes to move between tiles now scales with your current speed.
tweak: Layman's term: Movement is now much more smoother and should stutter much less
/:cl:

### DIFF
--- a/code/datums/action.dm
+++ b/code/datums/action.dm
@@ -247,17 +247,5 @@ Useful as an alternative to TG radials and the unholy input()
 
 /datum/action/collapsible/can_use_action()
 	return TRUE //Shouldn't be too much of an issue, the actions to be displayed should be properaly sanity checked however
-
-/datum/action/collapsible/action_activate()
-	if(IsOpen)
-		Collapse()
-	else
-		Expand()
-
-/datum/action/collapsible/proc/Collapse()
-	for(var/datum/action/act in actions)
-		act.remove_action(owner)
-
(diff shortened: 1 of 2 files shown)


ma44:
Note to self: Fix following issues

1. Move glide_size to mob Move() instead of client; fixes issues like getting pulled using old glide size as the pulled mob moves really fast compared to the pulling mob

2. Sync the loader/mechs with user's glide_size

3. see about freezing when going diagonal

spookydonut:
This is just a hacky change to what was a bad refactor: https://github.com/tgstation/TerraGov-Marine-Corps/blame/065dfc70ff4fb315a343780e2610bf38f74b8532/code/modules/mob/mob_movement.dm#L190
glide_size shouldnt be set like this.

PR #1051, opened 2019-03-13
More admin changes

DominikPanic:
-map template loader
-splits admin log into admin, asay, msay for ease of reading
-cleans up some code
-buildmode
-purges old feedback system and some ancient globals
-polls


code/__HELPERS/unsorted.dm
```diff
@@ -953,22 +959,40 @@ var/global/image/busy_indicator_hostile
 			else
 				air_master.tiles_to_update += T2*/
 
-proc/DuplicateObject(obj/original, var/perfectcopy = 0 , var/sameloc = 0)
-	if(!original)
-		return null
 
-	var/obj/O = null
+/proc/DuplicateObject(atom/original, perfectcopy = TRUE, sameloc, atom/newloc)
```
Rohesie: There are no instances in either our or /tg/'s code in where this proc is called and either `perfectcopy` or `sameloc` are expected to have any values besides the standard ones, so they can be safely removed.

code/modules/buildmode/submodes/copy.dm
```diff
@@ -0,0 +1,32 @@
...
+
+/datum/buildmode_mode/copy/handle_click(client/c, params, obj/object)
+	var/list/pa = params2list(params)
+	var/left_click = pa.Find("left")
+	var/right_click = pa.Find("right")
+
+	if(left_click)
+		var/turf/T = get_turf(object)
+		if(stored)
+			DuplicateObject(stored, perfectcopy = TRUE, sameloc = FALSE, newloc = T)
```
Rohesie:
```suggestion
			DuplicateObject(stored, T)
```
After you've removed `perfectcopy` and `sameloc`.

code/__HELPERS/unsorted.dm
```diff
@@ -953,22 +959,37 @@ var/global/image/busy_indicator_hostile
 			else
 				air_master.tiles_to_update += T2*/
 
-proc/DuplicateObject(obj/original, var/perfectcopy = 0 , var/sameloc = 0)
-	if(!original)
-		return null
 
-	var/obj/O = null
+/proc/DuplicateObject(atom/original, atom/newloc)
+	if(!original || !newloc)
```
Rohesie: This `newloc` check is leaving `copy_contents_to()` broken. Granted, the proc is only used for the holodeck code, but it shouldn't be left like that. Just include `newloc` in the calls to `DuplicateObject` in `/area/proc/copy_contents_to(var/area/A , var/platingRequired = 0 )`.

DominikPanic: B-But you told me it's not used anywhere! Reeee

PR #1063, opened 2019-03-14
Fixes flamer backpack reloading.

### LINKED ISSUE #1018: 250-T FuelTank not refilling standard 250-T ammo tanks.
Discovered by accident during one of the attempts to refuel the 20/100 fuel tank and later 0/100 one only to receive a message that 'The bag is full'. To clarify, in description of the object it says that it contains 500 units of 250-T flamer fuel.

DominikPanic:
:cl: LaKiller8
fix: The pyro spec backpack now properly reloads default fuel.
/:cl:
Fixes #1018

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -699,17 +699,25 @@
 
 
 /obj/item/storage/backpack/marine/engineerpack/flamethrower/attackby(obj/item/W, mob/living/user)
-	if (istype(W, /obj/item/ammo_magazine/flamer_tank))
-		var/obj/item/ammo_magazine/flamer_tank/large/FTL = W
-		if(!FTL.current_rounds && reagents.total_volume)
-			var/fuel_available = reagents.total_volume < FTL.max_rounds ? reagents.total_volume : FTL.max_rounds
-			reagents.remove_reagent("fuel", fuel_available)
-			FTL.current_rounds = fuel_available
-			playsound(loc, 'sound/effects/refill.ogg', 25, 1, 3)
-			FTL.caliber = "UT-Napthal Fuel"
-			to_chat(user, "<span class='notice'>You refill [FTL] with [lowertext(FTL.caliber)].</span>")
-			FTL.update_icon()
 	. = ..()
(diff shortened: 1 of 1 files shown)


code/game/objects/items/storage/backpack.dm
```diff
@@ -699,17 +699,25 @@
...
-		var/obj/item/ammo_magazine/flamer_tank/large/FTL = W
-		if(!FTL.current_rounds && reagents.total_volume)
-			var/fuel_available = reagents.total_volume < FTL.max_rounds ? reagents.total_volume : FTL.max_rounds
-			reagents.remove_reagent("fuel", fuel_available)
-			FTL.current_rounds = fuel_available
-			playsound(loc, 'sound/effects/refill.ogg', 25, 1, 3)
-			FTL.caliber = "UT-Napthal Fuel"
-			to_chat(user, "<span class='notice'>You refill [FTL] with [lowertext(FTL.caliber)].</span>")
-			FTL.update_icon()
 	. = ..()
```
spookydonut: will give an error about you cant fit that in here if this is first. (probably)

DominikPanic: It won't because I don't do an early return `if(!.)`

spookydonut: Why not only call ..() if you arent refilling something?

DominikPanic: Good point, changed

code/game/objects/items/storage/backpack.dm
```diff
@@ -699,17 +699,24 @@
...
-	. = ..()
+	if(!istype(W, /obj/item/ammo_magazine/flamer_tank))
+		return ..()
+	var/obj/item/ammo_magazine/flamer_tank/FTL = W
+	if(FTL.default_ammo != /datum/ammo/flamethrower)
+		return ..()
+	if(FTL.max_rounds == FTL.current_rounds)
+		return ..()
+	if(reagents.total_volume <= 0)
+		return ..()
```
spookydonut: add feedback here

PR #1065, opened 2019-03-14
F12 now properly updates the ammo HUD.

### LINKED ISSUE #899: Multiply press F12 makes ammo count dissapear while wielding guns

DominikPanic:
:cl: LaKiller8
fix: Pressing F12 now properly updates the ammo HUD.
/:cl:
Fixes #899

### DIFF
--- a/code/_onclick/hud/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects.dm
@@ -631,7 +631,7 @@
 	screen_loc = ui_ammo
 	var/warned = FALSE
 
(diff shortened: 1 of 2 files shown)


code/_onclick/hud/screen_objects.dm
```diff
@@ -631,7 +631,7 @@
 	screen_loc = ui_ammo
 	var/warned = FALSE
 
-/obj/screen/ammo/proc/add_hud(var/mob/user)
```
spookydonut: if you're changing these should be mob/living/user

PR #1068, opened 2019-03-15
Components 2.0, now featuring atomized commits.

DominikPanic:
I'll need a lotta help with this.


code/datums/components/squeak.dm
```diff
@@ -0,0 +1,95 @@
...
+		last_use = world.time
+		play_squeak()
+
+
+/datum/component/squeak/proc/on_equip(datum/source, mob/equipper, slot)
+	return
+
+
+/datum/component/squeak/proc/on_drop(datum/source, mob/user)
+	return
```
ninjanomnom: It's not like you don't have disposals, this is implementable

code/datums/datum.dm
```diff
@@ -9,6 +10,7 @@
 	var/datum/weakref/weak_reference
 	var/hidden_from_codex = FALSE //set to TRUE if you want something to be hidden.
 
+
```
ninjanomnom: You're missing all the code that cleans up components

code/game/atoms.dm
```diff
@@ -691,12 +691,21 @@ Proc for attack log creation, because really why not
...
 	return INITIALIZE_HINT_NORMAL
 
+
 //called if Initialize returns INITIALIZE_HINT_LATELOAD
 /atom/proc/LateInitialize()
 	return
 
+
+// Put your AddComponent() calls here
+/atom/proc/ComponentInitialize()
```
ninjanomnom: Don't port this part

code/modules/clothing/shoes/miscellaneous.dm
```diff
@@ -77,10 +81,15 @@
 	name = "clown shoes"
 	icon_state = "clown"
 	item_state = "clown_shoes"
-	slowdown = SHOES_SLOWDOWN+1
-	var/footstep = 1	//used for squeeks whilst walking
```
ninjanomnom: I know you're replacing what this was for but was this not used at all? Where's the code removal of whatever is using this?

DominikPanic: I couldn't find any reference to it anywhere.

code/datums/datum.dm
DominikPanic: Added

ninjanomnom:
Seems probably fine but it needs to be tested

PR #1070, opened 2019-03-15
Updates contibution guidelines

spookydonut:


.github/CONTRIBUTING.md
```diff
@@ -156,6 +156,33 @@ Remember: although this tradeoff makes sense in many cases, it doesn't cover the
...
 While we normally encourage (and in some cases, even require) bringing out of date code up to date when you make unrelated changes near the out of date code, that is not the case for `New` -> `Initialize` conversions. These systems are generally more dependant on parent and children procs so unrelated random conversions of existing things can cause bugs that take months to figure out.
 
+### `Initialize()` must be used properly
+`Initialize()` must always call the parent proc `..()` to ensure the atom initializes and must return a valid value either by using `. = ..()` or `return ..()` or `return INITIALIZE_HINT_*`
+This ensures atoms initialize correctly and there are not false positives of failures.
+You also must not use `qdel(src)` inside `Initialize()`, use `return INITIALIZE_HINT_QDEL` instead.
+
+### Use of `..()` vs `. = ..()`
+There's no reason to not to use `. = ..()` instead of `..()`.  The exception being `return ..()` or procs where you do not want to call the parent proc.
```
ninjanomnom: This is not true

spookydonut: Do you want to expand on that?

ninjanomnom: There are valid reasons to use just `..()` instead of `return ..()` or `. = ..()` and this generalization shouldn't be a guideline

PR #1076, opened 2019-03-16
Medkit upgrade and sprite port

Alterist:
:cl:
add: Medkits now show how much is left in it's stack
imageadd: updated some item/storage sprites thanks to Bay and CEV-Eris
/:cl:


code/game/objects/items/stacks/medical.dm
```diff
@@ -147,6 +147,27 @@
 	stack_id = "Tear petals"
 
 
+/obj/item/stack/medical/advanced/
```
DominikPanic:
```suggestion
/obj/item/stack/medical/advanced
```

code/game/objects/items/stacks/medical.dm
```diff
@@ -147,6 +147,27 @@
...
+	dir = NORTH
+	flags_atom = DIRLOCK
+
+/obj/item/stack/medical/advanced/update_icon()
+	if(max_amount < 1)
+		return
+	var/percentage = round(amount / max_amount) * 100
+	if(amount >= max_amount)
```
DominikPanic: Early return, combine it with the first check.

Rohesie: Is the check even necessary? As it it, it should never be true, so this should only work for full stacks.
If you wanted it not to update the sprite if the amount is higher than the maximum possible, the why set the switch to account for `to INFINITY`?

code/game/objects/items/stacks/medical.dm
```diff
@@ -147,6 +147,27 @@
...
 
+/obj/item/stack/medical/advanced
+	dir = NORTH
+	flags_atom = DIRLOCK
+
+/obj/item/stack/medical/advanced/update_icon()
+	if(max_amount < 1)
+		return
+	var/percentage = round(amount / max_amount) * 100
+	if(max_amount >= amount)
```
DominikPanic: Combine it with the first check for that double sweet early return.

Rohesie: That would be something like `if(max_amount <1 || max_amount < amount)` for an early return.

Alterist: changed

code/game/objects/items/stacks/medical.dm
```diff
@@ -254,7 +276,7 @@
 	stack_id = "splint"
 
 /obj/item/stack/medical/splint/attack(mob/living/carbon/M, mob/user)
-	if(..()) return 1
+	if(..()) return TRUE
```
DominikPanic: Same line control statement.

code/game/objects/items/stacks/medical.dm
```diff
@@ -147,6 +147,27 @@
 	stack_id = "Tear petals"
 
 
+/obj/item/stack/medical/advanced
+	dir = NORTH
+	flags_atom = DIRLOCK
+
+/obj/item/stack/medical/advanced/update_icon()
+	if(max_amount < 1)
```
DominikPanic:
```suggestion
	if(max_amount < 1 || amount < max_amount)
```

code/game/objects/items/stacks/medical.dm
```diff
@@ -147,6 +147,27 @@
...
+		return
+	var/percentage = round(amount / max_amount) * 100
+	if(amount < max_amount)
```
DominikPanic:
```suggestion

```

PR #1084, opened 2019-03-17
Fixes hunter sprite offset

Alterist:
i'm the worst reviewer


code/modules/mob/living/carbon/xenomorph/Castes/Hunter.dm
```diff
@@ -152,8 +152,8 @@
 	plasma_stored = 50
 	tier = 2
 	upgrade = 0
-	pixel_x = -12
-	old_x = -12
+	pixel_x = 0
```
DominikPanic: It should be safe to just remove both as I couldn't find them redefined at the Xenomorph level.

Alterist: changed

PR #1091, opened 2019-03-18
Sentry Gun Rebalance 2.0

Surrealaser:
:cl: by Surrealistik
balance: Sentry Guns now cost 120 Requisition up from 100.
tweak: Sentry Gun burst accuracy penalty increased from (burst amount - 1) ^ 2 to  (burst amount - 1) ^ 2 * 1.2
tweak: AP of minisentry reduced from 20 to 10, but it now deploys anchored and activated, can be undeployed while anchored and powered, and can be secured/unsecured with a screwdriver in 1 second down from 2.
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -273,7 +273,7 @@ WEAPONS
 	contains = list(
 					/obj/item/storage/box/sentry
 					)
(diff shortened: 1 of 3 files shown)


juanototo2:
### **[AGITATED SCREECHING]**

PR #1096, opened 2019-03-18
SD Balance and wall adjustments

Alterist:
:cl:
balance: moved SD shutters forward, adds back inner walls
fix: SD walls shouldn't link with the shutters anymore
/:cl:


Metroidlover:
I dislike how this cuts off the loop at the back of the ship, I think the shutters should be moved east two tiles

francinum:
If you really want to cut it off, move it west another tile and make those internal walls hull.

Rohesie:
I've suggested Franc's idea before. Being a lower layer than the wall would solve it and make it look much smoother, really. But this is up to Alterist in the end. Let's test again and go the way they decide.

PR #1104, opened 2019-03-19
Smartgun Tweaks

Surrealaser:
:cl: by Surrealistik
tweak: Smartgun base accuracy multiplier slightly increased, from 1.05 to 1.1
/:cl:

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -322,7 +322,7 @@
 
 /datum/skills/smartgunner
 	name = "Squad Smartgunner"
-	smartgun = SKILL_SMART_EXPERT
+	smartgun = SKILL_SMART_TRAINED
 	leadership = SKILL_LEAD_BEGINNER
 
 /datum/skills/smartgunner/pmc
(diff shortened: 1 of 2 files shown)


Pariah919:
Make sure to add it to field commander if you haven't. The expert skill.

InterroLouis:
> Make sure to add it to field commander if you haven't. The expert skill considering FC spawns with a set and has spec weapon skill would make me assume they are also an expert in smartgun usage.

No, the point is to make the smartgunner themselves better at using the smartgun, not to buff up everyone who has smartgun skill. Read the thing, it's on purpose, so they're better than Commanders and Specialists at using the smartgun.

Rohesie:
This halves wield time, fire delay and adds accuracy, all of which we'll have to adjust for.
The scatter and recoil are not issues in this case.
The PR looks rushed as it is. The extent of this buff is not properly conveyed by "increased skill".

Rohesie:
I'll step in the power creep and request the skill increase to be removed. Rework the whole skill itself if you want to change it, just doubling the numbers is not really a good alternative.

Surrealaser:
> I'll step in the power creep and request the skill increase to be removed. Rework the whole skill itself if you want to change it, just doubling the numbers is not really a good alternative.

Okay, we still had people calling weak, but sure.

PR #1107, opened 2019-03-20
Fixes initial damages not being applied on flamer_fire instanciation

Menshin:
## About The Pull Request

Quickfix for initial damages on flame creation being overlooked in the fire range refactor.
Some light code cleaning, but the whole flame code is a mess and should be added on the refactor list, if not already.

## Changelog
:cl:
fix: Flame now applies initial damage correctly

/:cl:

### DIFF
--- a/code/modules/projectiles/updated_projectiles/guns/flamer.dm
+++ b/code/modules/projectiles/updated_projectiles/guns/flamer.dm
@@ -489,12 +489,11 @@
 	for(var/mob/living/C in T)
 		if(C.fire_immune)
 			continue
-		else
-			C.adjust_fire_stacks(fire_stacks)
-			var/armor_block = C.run_armor_check("chest", "energy")
-			C.apply_damage(fire_damage, BURN, null, armor_block)
-			C.IgniteMob()
-			C.visible_message("<span class='danger'>[C] bursts into flames!</span>","[isxeno(C)?"<span class='xenodanger'>":"<span class='highdanger'>"]You burst into flames!</span>")
+
+		C.flamer_fire_act(fire_stacks,100)
+		var/armor_block = C.run_armor_check("chest", "energy")
+		C.apply_damage(fire_damage, BURN, null, armor_block)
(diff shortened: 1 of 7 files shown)


code/modules/projectiles/updated_projectiles/guns/flamer.dm
```diff
@@ -457,52 +462,39 @@
...
 	return ..()
 
+/obj/flamer_fire/proc/apply_initial_damage(fire_stacks, fire_damage)
+	var/turf/T = get_turf(src)
+
+	for(var/mob/living/C in T)
+		if(C.fire_immune)
+			continue
+		else
+			C.adjust_fire_stacks(fire_stacks)
```
spookydonut: screammmmmmmmmmmmmmmmm use the flamer_fire_act() etc mob procs.

code/modules/projectiles/updated_projectiles/guns/flamer.dm
```diff
@@ -457,52 +462,38 @@
...
 /obj/flamer_fire/Destroy()
 	SetLuminosity(0)
 	STOP_PROCESSING(SSobj, src)
 	return ..()
 
+/obj/flamer_fire/proc/apply_initial_damage(fire_stacks, fire_damage)
+	var/turf/T = get_turf(src)
+
+	for(var/mob/living/C in T)
+		if(C.fire_immune)
```
spookydonut: should be handled by flamer_fire_act

Menshin: Except it doesn't, want me to add it ?

spookydonut: if it isnt it should be handled by it, xenos do fire immune differently from memory so its important that behaviour is done via override.

even if you make a new proc called like flamer_fire_direct_act or something

Menshin:
Indeed, the flame code deserves a real refactor instead of a quick fix. I'll close this PR until i have some time to do it properly (still on area/lighting/machinery refactor). Of course, if anyone wants to take the task, be my guest.

PR #1110, opened 2019-03-20
Clean up some bits of code

SpaceManiac:
:cl:
fix: Fixed the bar sign never randomly picking The Net.
/:cl:

* Removes a bunch of stray `.`s everywhere.
* `is_type_in_typecache` and `TEST_RUN_PARAMETER` are each defined twice.
* `var/tmp/local_var` is no different than `var/local_var` (in procs).
* `(X ? Y : )` and `(X ? Y : ())` are both the same as `(X ? Y : null)`, except the latter actually makes sense.
	* `""` is used instead of `null` because it produces the same end-result when used in these contexts.
* `tgs/includes.dm` already includes all the TGS files, no need to include them again (it is harmless though).
* Regarding `THIS_PROC_TYPE_WEIRD` and `.....`: https://github.com/tgstation/tgstation/pull/43045

You also have two copies of `/obj/machinery/proc/power_change`, one in machinery.dm:222 and the other in power.dm:92, I can't tell which is correct so you should look at that.


Menshin:
> You also have two copies of /obj/machinery/proc/power_change, one in machinery.dm:222 and the other in power.dm:92, I can't tell which is correct so you should look at that.

Thank you.
If you feel like updating this PR, the one in `machinery.dm` is the one to keep, otherwise i'll do it in the area/lighting/machinery refactor that is currently worked on.

SpaceManiac:
Regarding `/tmp`, should have been more clear - it has no meaning on local variables, inside procs. It does have meaning on fields.

Menshin:
Got it 👍

PR #1111, opened 2019-03-20
LV Edit & Tweaks

Alterist:
:cl:
tweak: opens a few paths in the fog on LV, removes vines on dense jungle tiles to make open paths clearer
/:cl:

### DIFF
--- a/_maps/map_files/LV624/LV624.dmm
+++ b/_maps/map_files/LV624/LV624.dmm
@@ -3789,6 +3789,12 @@
 "aod" = (
 /turf/open/floor/plating/catwalk,
 /area/lv624/ground/river2)
+"aoe" = (
+/obj/effect/landmark/lv624/fog_blocker,
+/turf/open/floor/plating/warning{
+	dir = 10
+	},
+/area/lv624/ground/river2)
 "aof" = (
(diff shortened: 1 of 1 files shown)


francinum:
I believe the fog was meant to intentionally delay marine contact with the hive. I'm perfectly fine with the dense jungle changes, but I'm not so keen on such a major change in the function of the fog.

Rohesie:
It is true, that's the function of the fog. But it's also true that's a major disadvantage for the marines to be cut off from their supply lines. To re-purpose the fog into turning the area as a xeno chokepoint could be something interesting, but a lot should be considered. For example that marines can laze and shoot the tank through the fog.

francinum:
There's a hole into the internal fog channel at 65,86. Seal it.

Alterist:
fixed

Kingtrin:
This PR has completely broken LV's balance. I understand fog is not a fun concept, but its even less fun for the xenos to be rushed down. Xenos cannot hold the chokepoint early on due to a lack of T3's and a weak queen, meaning it devolves into a cave rushing shitshow. Marines are now encouraged more than ever before to rush down the xenos and ignore everything else.

PR #1121, opened 2019-03-22
Temperature defines

InterroLouis:
## About The Pull Request

Standardizes temperature defines to the proper ALL CAPS method of LOUD TEMPERATURE PROTECTION. Also makes the luchador masks protect from cold. 

## Why It's Good For The Game

Well, I suppose it cleans up some code? Also gives another option for those fashion minded marines on ICE COLONY. 

## Changelog
:cl:
tweak: Luchador masks now protect from cold
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/rig.dm
+++ b/code/modules/clothing/spacesuits/rig.dm
@@ -11,7 +11,7 @@
 	var/rig_color = "engineering" //Determines used sprites: rig[on]-[color] and rig[on]-[color]2 (lying down sprite)
 	actions_types = list(/datum/action/item_action/toggle)
 	flags_heat_protection = HEAD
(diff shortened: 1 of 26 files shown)


code/modules/clothing/spacesuits/rig.dm
```diff
@@ -252,7 +252,7 @@
 				if(!helmet)
 					to_chat(user, "\The [src] does not have a helmet installed.")
 				else
-					to_chat(user, "You detatch \the [helmet] from \the [src]'s helmet mount.")
+					to_chat(user, "You detach \the [helmet] from \the [src]'s helmet mount.")
```
spookydonut: atomise changes like this in future

PR #1123, opened 2019-03-22
Adds order overlays for humans

Jalordon:
## About The Pull Request

Makes it so orders have a visible overlay similar to xeno pheromones 

## Why It's Good For The Game

Makes it so that its a lot easier to tell when an order is in effect while in the middle of combat, since without the overlay it's easy to miss the message of an order issued. 

Additionally makes it easier to tell if you are still in range of an order and if the order is still active.

## Changelog
:cl:
add:Humans can now visibly see who is affected by a command order.
/:cl:


code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -424,9 +424,9 @@
 		. -= round(max(30,(shooter_human.traumatic_shock) * 0.2)) //Chance to hit declines with pain, being reduced by 0.2% per point of pain.
 		if(shooter_human.stagger)
 			. -= 30 //Being staggered fucks your aim.
-		if(shooter_human.marskman_aura)
-			. += shooter_human.marskman_aura * 1.5 //Flat buff of 3 % accuracy per aura level
-			. += P.distance_travelled * 0.35 * shooter_human.marskman_aura //Flat buff to accuracy per tile travelled
+		if(shooter_human.marksman_aura)
+			. += shooter_human.marksman_aura * 1.5 //Flat buff of 3 % accuracy per aura level
```
Jalordon: didn't change the effects of focus orders since felt out of scope of PR, but focus may be giving half the intended flat accuracy bonus

spookydonut: make this a define.

code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -424,9 +424,9 @@
...
+			. += P.distance_travelled * 0.35 * shooter_human.marksman_aura //Flat buff to accuracy per tile travelled
```
spookydonut: here too

Rohesie:
Can you make this visible for ghosts?

PR #1127, opened 2019-03-23
Xenos can open/close morgue trays now

Surrealaser:
:cl: by Surrealistik
tweak: Xenos can open/close morgue trays now.
/:cl:

### DIFF
--- a/code/game/objects/structures/morgue.dm
+++ b/code/game/objects/structures/morgue.dm
@@ -52,6 +52,9 @@
 /obj/structure/morgue/attack_hand(mob/user)
 	toggle_morgue(user)
 
+/obj/structure/morgue/attack_alien(mob/living/carbon/Xenomorph/M)
+	toggle_morgue(M)
+
 /obj/structure/morgue/proc/toggle_morgue(mob/user)
 	add_fingerprint(user)
 	if(!connected) return


code/game/objects/structures/morgue.dm
```diff
@@ -52,6 +52,9 @@
 /obj/structure/morgue/attack_hand(mob/user)
 	toggle_morgue(user)
 
+/obj/structure/morgue/attack_alien(mob/living/carbon/Xenomorph/M)
+	toggle_morgue(M)
```
DominikPanic: Should most likely be a do_after so they're actually prying it open.

Surrealaser: No, because marines have been using this to hide from xenos; if they take time to pry open, and a marine can close it in an instant, the problem is obvious.

On the flipside, if we make the xenos just break a tray and make it lodge open forever, it can trap the xeno if it has no acid.

The only other option I see is the xeno can pry the morgue tray open and break it such that the tray is severed, and doesn't impede movement.

spookydonut: Add a stun on the marine here so they cant toggle it closed again.

spookydonut:
See #1145

DominikPanic:
Done in #1145

PR #1143, opened 2019-03-25
everyone who should reasonably use a powerloader now does not suck at it

AdamElTablawy:
i wrote this on my phone

CE had a worse powerloader skill than his MTs, now does not

CT and RO now can use powerloaders when that's 80% of their job

PO can now use powerloaders as well
## Changelog
:cl:
tweak: CE, PO, MT, RO, and the early model synthetic can now use powerloaders at master level.
/:cl:

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -162,7 +162,7 @@
 	melee_weapons = SKILL_MELEE_SUPER
 	pilot = SKILL_PILOT_TRAINED
 	police = SKILL_POLICE_MP
-	powerloader = SKILL_POWERLOADER_TRAINED
+	powerloader = SKILL_POWERLOADER_MASTER
 	large_vehicle = SKILL_LARGE_VEHICLE_TRAINED
 
 
@@ -203,7 +203,7 @@
 /datum/skills/pilot
 	name = "Pilot Officer"
 	pilot = SKILL_PILOT_TRAINED
-	powerloader = SKILL_POWERLOADER_PRO
+	powerloader = SKILL_POWERLOADER_MASTER
 	leadership = SKILL_LEAD_TRAINED
 
 /datum/skills/MP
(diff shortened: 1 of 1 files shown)


Alterist:
Donno how I feel about giving EVERYONE max powerloader levels. A buff sure, but having everyone at master skill make the other skills a bit pointless

DominikPanic:
Agreed with Alterist here. It makes everything else obsolete.

AdamElTablawy:
yes that's because it's intentionally so

Rohesie:
At this point it would be better to rework the powerloader skill entirely, instead.

AdamElTablawy:
@Rohesie Put it in for now for QOL purposes

DominikPanic:
> 
> 
> @Rohesie Put it in for now for QOL purposes

We don't take shortcuts and further coder debt. And especially with things like this, it is not a hard rework to do.

AdamElTablawy:
What rework do you want.

DominikPanic:
> 
> 
> What rework do you want.

Your PR, your ideas. At the current state it's not acceptable.

AdamElTablawy:
So let me get this straight. You don't want me to buff powerloaders to be functionally usable so that people can actually move quickly in powerloaders because you want a rework.

You don't want to specify what the rework is or would consist of or what I should do for that, though.

DominikPanic:
No, I don't want you to do a lazy fix instead of a proper removal or rework. As simple as that.

AdamElTablawy:
You're still not making any sense.

AdamElTablawy:
Mmm, maybe i'll rephrase.

I don't know what you want me to do for the PR, and you don't want to specify what should be done in turn. You want a rework, but you don't want to say what the rework is.

Since I don't feel like playing guessing games, you can tell me what you want or you can close the PR.

DominikPanic:
It is your PR, you have the freedom to rework it to your liking. All I'm saying is that a lazy fix like this is not acceptable.

AdamElTablawy:
Okay, then I really don't feel like playing guessing games with you if you feel like playing coy about whatever it is you want.

PR #1145, opened 2019-03-26
Morgue Trays don't accept living mobs.

spookydonut:
## About The Pull Request

Kind of hugboxing but stops living marines hiding or getting stuck inside morgue trays

## Changelog
:cl:
tweak: Living mobs can no longer be put in morgue trays
/:cl:


Rohesie:
Not ideal but works under the current circumstances.

The better fix would be giving morgue trays lights, not unlike /tg/, to signal different kinds of contents, plus ways to open and close them for both those inside (after a do_after) and those outside (xenos as well).

PR #1149, opened 2019-03-27
Slowstack Related Balance Tweaks - TESTED

Surrealaser:
:cl: by Surrealistik
tweak: Slow stacks on laser targeted hits reduced by 1.
balance: Slow stacks overlap rather than stacking; when receiving slow stacks, the amount is set to the higher of current stacks and stacks to be added.
balance: Crusher is immune to receiving further slow and stagger stacks while at half its maximum charge speed or higher.
/:cl:


Rohesie:
This PR broke a game, by making slowdown to last forever. Slugs destroyed the hive.
A quick local test would have detected the issue.

Surrealaser:
@Rohesie Tested and fixed.

PR #1151, opened 2019-03-27
Hivelord Improvements - TESTED

Surrealaser:
:cl: by Surrealistk
tweak: Tunnel cooldown reduced from 4 minutes to 2. Tunnels are easier to destroy with grenades (2 frag grenades required). Tunnel travel time is now proportionate to distance between tunnels with a minimum travel time of 2 seconds, and a maximum of 4 or 6 for large Xenos. 
tweak: If a tunnel is destroyed, its corresponding tunnel will also be destroyed.
tweak: Hivelords can dismantle their own tunnels by attacking them while on Harm intent.
balance: Hivelords may only have up to 4 tunnel sets in play at once (the current amount is displayed in the Status panel). Tunnel sets created beyond this amount will destroy the oldest tunnel set.
tweak: Increased slow down stacks from sticky spit hits.
/:cl:


code/modules/cm_aliens/XenoStructures.dm
```diff
@@ -698,6 +698,8 @@ TUNNEL
 /obj/structure/tunnel/Destroy()
 	GLOB.xeno_tunnels -= src
 	if(other)
+		other.health = 0 //No sense in keeping the defunct tunnel around.
+		other.healthcheck()
```
Rohesie: This will delete the object referenced by `other`.
Then `other.other = null` will runtime.
This is the third locally-untested PR causing errors in less than a week.

DominikPanic: Damn, I feel stupid now.

spookydonut: yeah this is problematic

PR #431, opened 2019-01-29
Adds Imperium clothing

ondrej008:
Adds code for Imperium clothing, will add icons and stuff later.

:cl: Ondrej008
add: Adds boots, helmets, armour, headset for the Imperial boys.
add: Guardsmen ERT
/:cl:

# FOR THE EMPEROR!

also probably will add an ERT so this *actually* gets used.

- [x] ERT
- [x] Sprites
- [x] Code

PR #1163, opened 2019-03-29
Doctor medical buff

Rohesie:
Skills renamed to be more generic.

## Changelog
:cl:
balance: Doctor medical skill buffed to CMO/Synth level, so they are all on par.
/:cl:

Also removed the WIP endurance skill, with no effect in game.


spookydonut:
Other way round, buff doctor skill to match cmo/synth

Rohesie:
That would leave a gap with nobody with `SKILL_MEDICAL_DOCTOR` besides the survivor doctor and early synthetic.

All this changes is that CMOs will now heal 12 burn, brute and tox damage per defib instead of 15 of each, that they will take two seconds instead of one while injecting a syringe, and that they will take 2.5 instead of 2 seconds to splint someone else and 10 seconds instead of 8 to splint themselves.

Though now that I see it from that perspective it doesn't seem really problematic to make all docs that way. It would be a doctor buff, but not likely something that would have big impact in the game.

Gather more feedback, I suppose?

sleepyretard:
Such changes are so insignificant, in my mind, that it doesn't really do any good to "nerf" them - but it would be slightly nicer to "buff" every doctor to that standard.

It's practically a medical QoL change rather than a buff, if you were to go that route. It doesn't make doctors unstoppable, unkillable gods, it just shaves a couple seconds off waiting. Not very overpowered, and not like it would change much.

DominikPanic:
> 
> 
> Such changes are so insignificant, in my mind, that it doesn't really do any good to "nerf" them - but it would be slightly nicer to "buff" every doctor to that standard.
> 
> It's practically a medical QoL change rather than a buff, if you were to go that route. It doesn't make doctors unstoppable, unkillable gods, it just shaves a couple seconds off waiting. Not very overpowered, and not like it would change much.
> 
> Just nice to have, rather than not to.

Powercreep.

Rohesie:
Let's try to get feedback here before deciding.

NonQueueingMatt:
@DominikPanic I saw this pop up in my github feed (which I browse for Bay and Aurora, maiy) and I knew I had to reply.

Labeling what sleepy said as powercreep is utterly ridicolous and beyond grasping at straws If anything, it's QoL. There's absolutely nothing of value to be gained from this nerf - I don't see the idea behind it either, really. You're leaving an unused skill level because... reasons?

Honestly I'm surprised someone saw a CMO taking 2 seconds less to do surgery and said "this is overpowered".

DominikPanic:
> 
> 
> @DominikPanic I saw this pop up in my github feed (which I browse for Bay and Aurora, maiy) and I knew I had to reply.
> 
> Labeling what sleepy said as powercreep is utterly ridicolous and beyond grasping at straws. If anything, it's QoL. There's absolutely nothing of value to be gained from this nerf - I don't see the idea behind it either, really. You're leaving an unused skill level because... reasons?
> 
> Honestly I'm surprised someone saw a CMO taking 2 seconds less to do surgery and said "this is overpowered".

And that's exactly how it gets you. Nothing but buffs for months with no nerfs.

sleepyretard:
> > @DominikPanic I saw this pop up in my github feed (which I browse for Bay and Aurora, maiy) and I knew I had to reply.
> > Labeling what sleepy said as powercreep is utterly ridicolous and beyond grasping at straws. If anything, it's QoL. There's absolutely nothing of value to be gained from this nerf - I don't see the idea behind it either, really. You're leaving an unused skill level because... reasons?
> > Honestly I'm surprised someone saw a CMO taking 2 seconds less to do surgery and said "this is overpowered".
> 
> And that's exactly how it gets you. Nothing but buffs for months with no nerfs.

I see that line of thinking, but surely if you're taking such a proactive stance on power creep, letting a single QoL feature through wouldn't be as bad as it could be. Power creep is a over time consistent thing, a one off buff is not that.

spookydonut:
Whichever way, synth should just be the same skill as doctor + cmo otherwise they end up in the same situation as cmo is.

DominikPanic:
> I see that line of thinking, but surely if you're taking such a proactive stance on power creep, letting a single QoL feature through wouldn't be as bad as it could be. Power creep is a over time consistent thing, a one off buff is not that.

Has there been anything but buffs for the past 6 months?

Rohesie:
This PR conflicts with #431 and should ideally not be tested before that one is over.

Rohesie:
The Imperium roles are in, their skills converted now. This should be ready.

Rohesie:
> @DominikPanic I saw this pop up in my github feed (which I browse for Bay and Aurora, maiy) and I knew I had to reply.

If you want not to be bothered by this kind of thing there's the Unwatch option.
Otherwise feel free to check us out, not need to justify it.

code/__DEFINES/objects.dm
```diff
@@ -185,11 +185,10 @@ var/list/RESTRICTED_CAMERA_NETWORKS = list( //Those networks can only be accesse
 #define OBJ_SKILL_MEDICAL		12
 #define OBJ_SKILL_SURGERY		13
 #define OBJ_SKILL_PILOT			14
-#define OBJ_SKILL_ENDURANCE		15
```
spookydonut: Undocumented change

PR #1166, opened 2019-03-29
Deleted objects no longer checked for Adjacency in MouseDrop

Rohesie:
Fixes some runtimes that happened whenever someone would click and drag an object who'd be deleted before releasing the mouse button.
Before the item Adjacent rework that would be culled during the proc, but this should never even check for Adjacency ideally.

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -12,7 +12,6 @@
 		return // should stop you from dragging through windows
 
 	over.MouseDrop_T(src,usr)
-	return
 
 
 // recieve a mousedrop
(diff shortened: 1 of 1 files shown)


code/_onclick/drag_drop.dm
```diff
@@ -5,15 +5,16 @@
...
+	if(!usr || !over || QDELETED(src))
+		return
+	if(!Adjacent(usr) || !over.Adjacent(usr))
+		return // should stop you from dragging through windows
 
-	spawn(0)
-		if(over)
-			over.MouseDrop_T(src,usr)
+	over.MouseDrop_T(src,usr)
 	return
```
DominikPanic: Uneccessary return

PR #1170, opened 2019-03-30
Ports SDQL2 selector arrays

JJRcop:
The source pull request is not merged yet, if changes are necessary I will mirror them here if possible.

# From tgstation/tgstation#43383

## About The Pull Request

I've added a selector array query. It's a special construct formed like this:
```bnf
'@[' <selectors> ']'
```
The `<selectors>` is the very same way that you normally use to select datums for all of the queries.
```bnf
UPDATE <selectors> SET <assignments>
SELECT <selectors>
```

I chose `@[ ]` because it's adding an @ sign to the existing SDQL list symbols, since it still generates a list.

## Why It's Good For The Game

It adds more flexibility to SDQL, allowing admins to do more with the language. I find it very convenient in testing to really narrow down the search to say, an /area, so SDQL will only look for datums in that /area's .contents, when paired with MAP.

Screenshot_20190328_123436

Screenshot_20190328_123333

## Changelog
:cl: JJRcop
add: SDQL2 now features a list constructor that takes a selector query. 
tweak: The MAP statement will unwrap lone lists that are the only result. This means the parent query begins directly iterating over the list rather than a list that contains the list.
/:cl:

i stayed up too late

### DIFF
--- a/code/modules/admin/SDQL2/SDQL_2.dm
+++ b/code/modules/admin/SDQL2/SDQL_2.dm
@@ -61,6 +61,10 @@
 
 	"SELECT /mob WHERE client MAP client WHERE holder MAP holder"
 
+	You can also generate a new list on the fly using a selector array. @[] will generate a list of objects based off the selector provided.
+
+	"SELECT /mob/living IN @[/area/crew_quarters/bar MAP contents]"
+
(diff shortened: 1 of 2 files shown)


JJRcop:
Port merged before upstream. Huh.

PR #1180, opened 2019-03-31
Armor datums

spookydonut:
## About The Pull Request
ports armor datums from /tg13/

## Why It's Good For The Game
step 2 on the path to refactoring obj attacks, allows us to use a fire armor type for fire attacks

essentially just changes it from implied strings to actual strings, converts everything currently using armor var to use the datum and sets acid and fire damage types where applicable
also everything now has an acid and fire armor rating but they're just the same value as energy so balance is unchanged
removes explosion resistance var that only the recursive explosion code used and we arent using it anyway but i modified it to use bomb armor rating of objects instead

### DIFF
--- a/_maps/WIP/Sulaco_top.dmm
+++ b/_maps/WIP/Sulaco_top.dmm
@@ -1822,9 +1822,7 @@
 /turf/open/floor/engine,
 /area/sulaco/engineering/atmos)
 "eq" = (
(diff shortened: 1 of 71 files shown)


DominikPanic:
If you add it to datumvars.dm I will :drooling_face: all over you.

PR #1187, opened 2019-04-01
Code cleanup.

Rohesie:
Periodic small code cleanup.

* Unused cloning paper-manual was giving off a warning, likely being defined before papers were. Just removed it.
* Old paint code was giving warnings. Just cleaned it up to standards.
* `var/global/obj/effect/datacore/GLOB.datacore` was leftover. Already defined by `GLOBAL_DATUM_INIT(datacore, /datum/datacore, new)`, so simply removed.
* json ancientcode was giving some errors. Cleaned it up to standards, turned a few tiny procs into defines and cleaned tidied it up a little. Worked fine in local tests.

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -1,6 +1,6 @@
 #define GAME_YEAR 2386
 
-var/global/defer_powernet_rebuild = 0		// true if net rebuild will be called manually after an event
+var/global/defer_powernet_rebuild = FALSE		// true if net rebuild will be called manually after an event
 
 var/global/list/global_map = null
 	//list/global_map = list(list(1,5),list(4,3))//an array of map Z levels.
--- a/code/global.dm
+++ b/code/global.dm
@@ -1,9 +1,5 @@
 #define GAME_YEAR 2386
 
-var/global/obj/effect/datacore/GLOB.datacore = null
-
-		//items that ask to be called every cycle
-
 var/global/defer_powernet_rebuild = 0		// true if net rebuild will be called manually after an event
 
 var/global/list/global_map = null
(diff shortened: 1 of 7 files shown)


code/modules/nano/JSON_Reader.dm
```diff
@@ -1,204 +1,210 @@
...
-					. += new/json_token/symbol(char)
-				else if(is_digit(char))
-					. += read_number()
+#define JSON_GET_CHAR(json, i) ( copytext(json, i, i + 1) )
+#define JSON_IS_WHITESPACE(char) ( char == " " || char == "\t" || char == "\n" || text2ascii(char) == 13 )
+#define JSON_GET_TOKEN(i) ( tokens[i] )
+#define JSON_NEXT_TOKEN(tokens, i) ( tokens[++i] )
+
+
+/json_token
```
spookydonut:
```Datum type paths must began with "datum"

In DM, this is optional, but omitting it makes finding definitions harder.```

These are actually /datum/json_token

File: code/global.dm [hunk 2]
```diff
@@ -1,10 +1,6 @@
 #define GAME_YEAR 2386
 
-var/global/obj/effect/datacore/GLOB.datacore = null
-
-		//items that ask to be called every cycle
-
-var/global/defer_powernet_rebuild = 0		// true if net rebuild will be called manually after an event
+var/global/defer_powernet_rebuild = FALSE		// true if net rebuild will be called manually after an event
```
spookydonut: this isnt a boolean.

Rohesie: Removed the superfluous checks.

PR #1190, opened 2019-04-02
unfucks the imperial pockets

spookydonut:


ondrej008:
Neat, that's a nice way to do it.

PR #1191, opened 2019-04-02
Slight SD Rework

Aurelien3189:
Reworked SD to be slightly more spacious while having more flanks:
Added windows to the front of the Button room, buttom room expanded by 4 tiles.
Added Docking port(s) to the back of SD, in preparation for potential future changes to SD/Evac
Removed outer hull tiles that were inside the SD room, which prevented flanks


Aurelien3189:
Well I'm having difficulties getting my master in line correctly, might just reset everything and start from scratch when I get home, but if I solution or advice can be provided, I'm all ears!

PR #1194, opened 2019-04-02
Renames is_mob_incapacitated to incapacitated

DominikPanic:
Will make porting /tg/ code much easier


Rohesie:
Curiously enough it used to be this way.
I prefer `is_mob_incapacitated` myself, but /tg/ compatibility is a plus.

PR #1205, opened 2019-04-04
Minor Ice Update

Aurelien3189:
Adds more to the area between the west underground cargo and LZ1, including the following:

More flanks and rooms at LZ1, including a garage/warehouse west of the landing zone, an extra storage room under Tcomms and a control bubble.
*DS1 Console moved to control bubble

Added lots of fences to the open area between hydro and Dorms, akin to LV624's fencing

Added small extension to tunnel in the middle of the map, small warehouse.

Opened up research with 3 openings in the walls, to serve as an alternate main path


Alterist:
The DS console used to be on an area that was always powered. That's why its not working as expected.
Also, you forgot to mapmerge

DominikPanic:
Wrong format, conflicting and stale

PR #1210, opened 2019-04-04
(un)Atomized admin PR

DominikPanic:
Yep, it's possible

-logging for windows and resin
-marked display of tickets
-rewritten offer/take mob
-roundid now attempts to set every 10 minutes in case it were to fail on startup
-topic.dm cleanup
-player panel tweaks
-mob transformation is now better
-changing squads now works properly again
-removes ancient vars from global.dm
-moved relevant globals to admin.dm
-a few minor fixes


DominikPanic:
It used to be. Such is life.

PR #773, opened 2019-02-19
[Ready] Tank-eful galore II

Ghommie:
## About The Pull Request
Extended bughunt and refactoring. Also made the tank vendor capable of repairing and refunding hardpoints and their ammo, split off multitile vehicle files into their own folder, and made the hardpoints repairable and reloadable even when not installed..

## Why It's Good For The Game
Fixing more tank issues. Some improvements. Closes #1116, closes #1101, closes #770, closes #706, closes #705.

## Changelog
:cl:
fix: Fixed more whacky tank issues.
fix: Alien effects such as eggs and resin (except floor weed) do get damaged by tanks walking over them now.
Add: Made the tank fabricator capable of repairing and refunding hardpoints and tank ammunition clips.
Slighty expanding their interface.
Add: Hardpoints can now be repaired, reloaded with spare clips and have their remaining integrity examined even if not installed in a vehicle.
/:cl:

PR #1218, opened 2019-04-05
Tank Tweak/Rebalance - TESTED

Surrealaser:
:cl: by Surrealistik
tweak: Overdrive now has nitros the driver can activate for a big, 5 second speed boost instead of a passive, continual speed increase. Has a cooldown of 20 seconds.
tweak: M56 Cupola ammo capacity increased to 1000 and fire rate increased to 3.33 shots / second up from 2 / second.
balance: Snowplow now takes half damage from wall collisions instead of no damage when facing walls, and destroys walls 50% faster down from 100% faster. Snowplow health increased by +100.
/:cl:


DominikPanic:
Should wait for #773

PR #1201, opened 2019-04-04
Refactoring Xeno Ability Buttons

spookydonut:
## About The Pull Request
Refactors xeno ability buttons to be more modular and consistent

## Why It's Good For The Game
makes behaviour more consistent and makes the code more maintainable

## Changelog
:cl:
tweak: Regurgitate will disabled if you have nothing to regurgitate
tweak: Many xeno ability buttons will now red/grey out when you can't use them.
tweak: Abilities on cooldown or that you lack plasma for are now selectable to allow quick switching
tweak: Abilities on cooldown now show a moving clock on top of the icon
tweak: Abilities on cooldown or you lack plasma for now show the green outline showing they're selected
/:cl:

- [x] Boiler bombard tested
- [x] Warrior abilities tested
- [x] Defender abilities tested
- [x] Drone abilities tested
- [x] Spitter/sent abilities tested

https://gyazo.com/95313c1c115a4b95031e83798e492377

PR #1232, opened 2019-04-06
Reduces xeno icon code duplication

spookydonut:
split off from #1201


spookydonut:
Locally tested, all special icon states work fine.

PR #1233, opened 2019-04-06
Shifts charge mob vars to defines

spookydonut:
split off from #1201


spookydonut:
Tested but its a trivial change anyway

PR #1241, opened 2019-04-06
Ports forensics component.

DominikPanic:
This was definitely easier than squeak.


Rohesie:
This is not as vital as in /tg/ due to the game mode, but ideally? Almost everything should be logged ideally.
It's what lets admins reconstruct a story.
Unless the performance loss is really significant, this is very desirable.

Rohesie:
Tested, no associated issues found.

PR #1243, opened 2019-04-06
Admin PR, this time atomized for real.

DominikPanic:


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1719,14 +1719,13 @@
 
 
 /mob/living/carbon/human/take_over(mob/M)
-	assigned_squad?.clean_marine_from_squad(src)
+	if(assigned_squad)
+		assigned_squad.clean_marine_from_squad(src)
```
Rohesie: Why the format change? The prior one worked the same, but with one less check.

DominikPanic: Because I don't trust it :cry:

Rohesie: Stop being paranoid.
D=<
This should be changing nothing.

spookydonut: i think the new code is more readible

Rohesie: Reeeeee.
Two against one, sad times. Oh well.
If you prefer it for readability I won't pry further, just don't assume it doesn't do what it's supposed to do.

code/modules/admin/verbs/datumvars.dm
```diff
@@ -677,566 +679,538 @@
...
-				H.dropItemToGround(H.head)
-				H.head = new /obj/item/clothing/head/kitty(H)
-				H.regenerate_icons()
-				H.head.update_icon(H)
-				log_admin("[key_name(usr)] has purrbated [key_name(H)].")
-				message_admins("[ADMIN_TPMONTY(usr)] has purrbated [ADMIN_TPMONTY(H)].")
+		if(!new_language)
+			return
 
+		if(!L)
```
Rohesie:
```suggestion
		if(QDELETED(L))
```

PR #1246, opened 2019-04-07
Smartgun Sight Change

InterroLouis:
## About The Pull Request

Increases the smartgun sight darkvision to the full sight range. 

## Why It's Good For The Game

Lets them see targets better, while also not making it useful for scoped fire. 
## Changelog
:cl:
balance: Smartgun NVG reached to the end of their screen now
/:cl:


Rohesie:
I'm actually adding this along with a big unatomize refactor, so I very much approve.

PR #1248, opened 2019-04-07
[READY]Adds in a basic tank with code that isn't terrible!

Kmc2000:
## About The Pull Request

Hi /tg/mc gamers! 

Rohesie asked me to take a look at tank code, I promptly took one look at it, screamed my head off, and ran.

This PR attempts to fix some of the inherent issues with it, and can hopefully be extended to encompass other vehicles as well however that's not why I'm here.

What this is is a more modular BASE. You'll need to add back the features you want from the original tank but hopefully this should make it less cancerous to work with.

What this DOES have is combat functionality. The gun turret swivels around as you aim at shit, and you can click + hold on a tile to repeatedly fire at it.

Controls:
Click to enter
Right click to see verbs

Click anywhere as gunner to fire the minigun
CTRL Click anywhere as gunner to fire the main tonk cannon

TODO:

Add a modular upgrades system []
Add a way to reload the tank [X]
Add an engine noise that plays when someone's in the tank (so benos can hear them coming) [X]
Add more sounds [X]
Add a way to exit the tank [ X ]

DEMO:
https://file.house/Zl23.mp4

Please note, not all sprites aren't mine. Rohesie sent me some. If you own those let me know so I can credit you.

Apc and tank sprites are reworked versions of arachnidnexus' tank

## Changelog
:cl: Kmc2000 (Sprites by Arachnidnexus and Kmc2000)
add: Made an experimental new tank dubbed the "Benocrusher". Testing only for now.
/:cl:


code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+	ammo = new /obj/item/ammo_magazine/tank/ltaaap_minigun
+	fire_sounds = list('sound/weapons/tank_minigun_loop.ogg')
+	cooldown = 2 //Minimal cooldown
+
+/obj/item/weapon/tank_weapon/Initialize()
+	. = ..()
+
+/obj/item/weapon/tank_weapon/proc/fire(var/atom/T,var/mob/user)
+	if(!can_fire(T))
+		return
```
Menshin:
```suggestion
		return FALSE
```
for consistency

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+	if(ammo.current_rounds <= 0)
+		return FALSE
+	if(world.time < lastfired + cooldown)
+		if(owner.gunner)
```
Menshin: This check is useless if we're going to return `FALSE` anyway.

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+		playsound(get_turf(owner), pick(fire_sounds), 60, 1)
+	ammo.current_rounds--
+	return TRUE
+
+/obj/item/weapon/tank_weapon/proc/can_fire(var/turf/T)
+	if(T in orange(owner, 2))
+		if(owner.gunner)
+			to_chat(owner.gunner, "Firing your main gun here could damage the tank!")
+			return FALSE
+	return TRUE
```
Menshin: Wouldn't it make sense to check for the existence of `owner.gunner` here instead of  in `fire()`?

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+
+/obj/item/weapon/tank_weapon/minigun/can_fire(var/turf/T)
+	return TRUE //No loc check here
+
+/obj/vehicle/tonk
+	name = "MK-1 Benocrusher x10"
+	desc = "A gigantic wall of metal designed for maximum beno destruction. Click it with an open hand to enter as a pilot or a gunner."
+	icon = 'icons/obj/tonk.dmi'
+	icon_state = "tank"
+	layer = 3
```
Menshin:
```suggestion
	layer = OBJ_LAYER
```

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+	var/minigun_dir = null
+	var/firing = FALSE //Used in autofire
+	var/turf/stored_firing_turf = null
+
+/obj/turret_overlay
+	name = "Tank gun turret"
+	desc = "Big fucking tonk gun oh lawd he comin"
+	icon = 'icons/obj/tonk_gun.dmi'
+	icon_state = "turret"
+	layer = 3.1
```
Menshin:
```suggestion
	layer = ABOVE_OBJ_LAYER
```

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+	mouse_opacity = FALSE //It's an overlay
+
+/obj/vehicle/tonk/Initialize()
+	. = ..()
+	turret_overlay = new(get_turf(src))
+	update_icon()
+	health = max_health
+	main_cannon = new(src) //Make our guns
+	minigun = new(src)
+	main_cannon.owner = src
```
Menshin: This would probably best set in the weapons `New()`.
Same for below.

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+		return
+	if(gunner && position == "gunner")
+		to_chat(user, "[gunner] is already controlling [src]!")
+		return
+	if(can_enter(user)) //OWO can they enter us????
+		to_chat(user, "You climb into [src] as a [position]!")
+		return enter(user, position) //Yeah i could do this with a define, but this way we're not using multiple things
+
+/obj/vehicle/tonk/proc/can_enter(var/mob/living/carbon/M) //NO BENOS ALLOWED
+	if(!M.IsAdvancedToolUser())
```
Menshin: Should probably `if(!istype(M))` to prevent silicon from using this ?

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+	var/time = 10 SECONDS - 2 SECONDS * M.mind.cm_skills.large_vehicle
+	if(do_after(M, time, TRUE, 5, BUSY_ICON_BUILD) && M.Adjacent(src))
+		return TRUE
+
+/obj/vehicle/tonk/proc/enter(var/mob/user, var/position) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
+	if(!user || !position)
+		return //Something fucked up. Whoops!
+	switch(position)
+		if("pilot")
+			user.forceMove(src)
```
Menshin: Move `user.forceMove(src)` and `operators += user` out of the `switch`.

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+		if("gunner")
+			user.forceMove(src)
+			gunner = user
+			operators += user
+
+/obj/vehicle/tonk/proc/exit_tank(var/mob/user) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
+	if(!user)
+		return //Something fucked up. Whoops!
+	if(user == pilot)
+		user.forceMove(get_turf(src))
```
Menshin: Same here.

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+		pilot = null
+		operators -= user
+	if(user == gunner)
+		user.forceMove(get_turf(src))
+		gunner = null
+		operators -= user
+
+/obj/vehicle/tonk/relaymove(mob/user, direction)
+	if(user.incapacitated() || user != pilot)
+		to_chat(user, "Youre not the pilot fucking idiot.")
```
Menshin:
```suggestion
		to_chat(user, "You're not the pilot.")
```

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+TONKS BY KMC
+THIS IS LIKE REGULAR CM TONK BUT IT'S LESS SHIT
+WHOEVER MADE CM TANKS: YOU ARE A BAD CODER!!!!!
+*/
+
+//generic component stuff
+/atom
+	var/tank_crash_damage = 10  //Arbitrary number. Set this to whatever you want with extra walltypes
+
+/atom/proc/tank_act(var/obj/vehicle/tonk/victim)
```
Menshin: The tank's not the victim here 😃

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+	update_icon()
+
+/obj/vehicle/tonk/Bump(atom/unluckyfuck, yes)
+	. = ..()
+	unluckyfuck.tank_act(src) //You poor motherfucker
+
+/obj/vehicle/tonk/Bumped(atom/unluckyfuck, yes)
+	return FALSE //Tonk BIG, THOG SAY NOT MOVE TONK WHEN BUMPED
+
+/client/MouseDown(object, location, control, params)
```
Menshin: Wouldn't it be better to have some kind of `vehicule` var in the mob that relay the click to any vehicule the mob is on instead of redefining these here ?
@spookydonut 
@DominikPanic 
@Rohesie

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
+/*
+TONKS BY KMC
+THIS IS LIKE REGULAR CM TONK BUT IT'S LESS SHIT
+WHOEVER MADE CM TANKS: YOU ARE A BAD CODER!!!!!
+*/
+
+//generic component stuff
+/atom
+	var/tank_crash_damage = 10  //Arbitrary number. Set this to whatever you want with extra walltypes
```
Rohesie: Could you add atom vars to `atoms.dm` ?

code/modules/vehicles/tonk.dm
Rohesie: That would be a bit saner, yes, tied to enter and leave procs to modify it.

code/modules/vehicles/tonk.dm
DominikPanic: kek

code/modules/vehicles/tonk.dm
DominikPanic: Yes, there' the click_intercept var for clients and mobs.

Kmc2000: I just did it that way for clarity

code/modules/vehicles/tonk.dm
ninjanomnom: I really dislike having to use this at all, damage from crashing into something should be based on a more generic value

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+		return FALSE
+	if(!allowed(M))
+		to_chat(M, "<span class='warning'>Access denied.</span>")
+		return FALSE
+	var/obj/item/offhand = M.get_inactive_held_item()
+	if(offhand && !(offhand.flags_item & (NODROP|DELONDROP)))
+		to_chat(M, "<span class='warning'>You need your hands free to climb on [src].</span>")
+		return FALSE
+
+	if(!M.mind || !(!M.mind.cm_skills || M.mind.cm_skills.large_vehicle >= SKILL_LARGE_VEHICLE_TRAINED))
```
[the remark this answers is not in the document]
Menshin:
```suggestion
	if(!M.mind || M.mind?.cm_skills?.large_vehicle < SKILL_LARGE_VEHICLE_TRAINED)
```
Also, doesn't lazy evaluation makes the `M.mind?` redundant ?

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+/obj/vehicle/tonk/update_icon() //To show damage, gun firing, whatever. We need to re apply the gun turret overlay.
+	var/icon/I = icon(icon,icon_state,dir)
+	bound_width = I.Width() //Adjust the hitbox in case admins want to make an OMEGATONK
+	bound_height = I.Height()
+	if(!turret_overlay) //We have to do this because of BYOND's innate dir setting habits, in byond, if you have an overlay'd object its direction can ONLY be which way the source is facing. So, we improvise.. ((yes this is kinda shit))
+		return
+	if(main_cannon_dir)
+		turret_overlay.setDir(main_cannon_dir)
+		if(main_cannon_dir == WEST || main_cannon_dir == SOUTHWEST|| main_cannon_dir == NORTHWEST)
+			turret_overlay.pixel_x = -19
```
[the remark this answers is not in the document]
Kmc2000: I dont know what brittle code is but this is the only way i can make the gun swivel properly due to how BYOND handles overlays

code/modules/vehicles/tonk.dm
ninjanomnom: This isn't helping clarity

code/modules/vehicles/tonk.dm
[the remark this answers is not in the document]
ninjanomnom: You shouldn't be setting the offset like this, handle it in a generic fashion for turning or make the overlay art itself handle it

code/modules/vehicles/tonk.dm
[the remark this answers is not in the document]
ninjanomnom: Oh, yeah with the reversed condition it's unnecessary

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+		return
+	if(modifiers["middle"])
+		return
+	if(modifiers["alt"])
+		return
+	handle_fire(A)
+
+/obj/vehicle/tonk/proc/handle_fire(var/atom/A)
+	if(!minigun)
+		if(gunner)
```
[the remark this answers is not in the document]
Menshin: Made the same mistake: it should `return FALSE` if there's no `minigun` in all cases.
The second `if` is to notify any gunner of the failure.

code/modules/vehicles/tonk.dm
Kmc2000: Copied from current tankcode so iunno

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+	if(T in orange(owner, 2))
+		if(owner.gunner)
+			to_chat(owner.gunner, "Firing your main gun here could damage the tank!")
+			return FALSE
+	return TRUE
+
+/obj/item/weapon/tank_weapon/minigun/can_fire(var/turf/T)
+	return TRUE //No loc check here
+
+/obj/vehicle/tonk
```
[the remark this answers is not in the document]
Kmc2000: I'll change this when this is ready to replace the old tank. Makes it easier that way.

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+			update_icon()
+
+/obj/vehicle/tonk/proc/handle_fire_main(var/atom/A) //This is used to shoot your big ass tank cannon, rather than your small MG
+	if(!main_cannon)
+		if(gunner)
+			to_chat(gunner, "You haven't installed a main cannon on this tank you fucking idiot")
+		return FALSE
+	firing = TRUE
+	while(firing)
+		stoplag(2)
```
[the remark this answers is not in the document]
Kmc2000: And on that day, carp fishing will finally be complete.

code/modules/vehicles/tonk.dm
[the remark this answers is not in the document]
Rohesie: Just so you know, @Kmc , the existing tank never rotated its turret on firing. Since the 3/4 perspective makes rotating the turret require one sprite for each  direction, consider this not a really necessary feature if it makes your job harder.
The way the current tank handles it is that you can only fire the main gun in the direction the tank is facing, which forces coordination between gunner and driver.
Not saying changing this is bad, merely for you to consider the cost and if it's worth the effort for you.
Though support for the functionality wouldn't be bad, for if we add a different gun whose sprite allows for smoother 360 rotation.

config/admins.txt
```diff
@@ -1,101 +1 @@
-wjohnston = Admin
```
DominikPanic: Absolutely unacceptable. Revert this change immediately.

francinum:
KMC writing code that doesn't cause actual pain is something that simply doesn't happen.

config/admins.txt
```diff
@@ -1,101 +1 @@
...
-jroinc1 = Admin
-sanchez13 = Admin
-hughgent = Admin
-fourty2 = Admin
-csolaris = Admin
-devilsaaron = Admin
-shadow_quill = Admin
-cosmicscientist = Admin
\ No newline at end of file
+Kmc2000 = Admin
```
ondrej008: If you're going to add yourself, don't remove the others.

code/modules/vehicles/tonk.dm
Rohesie: Silicons in current code are allowed to enter the tank. They can drive it, they just can't shoot due to their restrictions.
It would be necessary to check for the synth flag anyway, to prevent them, as they are carbon and human.

code/modules/vehicles/multitile/cm_armored.dm
```diff
@@ -341,7 +343,7 @@ var/list/TANK_HARDPOINT_OFFSETS = list(
 /atom/proc/tank_collision(obj/vehicle/multitile/hitbox/cm_armored/C, facing, turf/T, turf/temp)
 	return
 
-/mob/living/tank_collision(obj/vehicle/multitile/hitbox/cm_armored/C, facing, turf/T, turf/temp)
+/mob/living/tank_collision(obj/vehicle/C, facing, turf/T, turf/temp)
```
Rohesie:
```suggestion
/mob/living/tank_collision(obj/vehicle/V, facing, turf/T, turf/temp)
```
Try using variables that represent what you are after. It improves readability.
`C` usually implies `carbon`. Or the more evil `cm_armored` we are trying to get rid of.

code/modules/vehicles/tonk.dm
```diff
@@ -0,0 +1,354 @@
...
+		else
+			turret_overlay.pixel_x = 0
+	turret_overlay.forceMove(get_turf(src))
+
+/obj/vehicle/tonk/attack_hand(mob/user)
+	if(pilot && gunner)
+		to_chat(user, "You are unable to enter [src] because all of its seats are occupied!")
+		return ..()
+	if(user in operators)
+		exit_tank(user)
```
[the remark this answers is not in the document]
Kmc2000: FUCK

code/modules/vehicles/tonk.dm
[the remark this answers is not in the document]
ninjanomnom: This pr should not be merged until it's actually ready to replace the existing code, keeping both around is a terrible idea

Ghommie:
Uh uh, I don't think you merged from the master correctly. there are garbage backup map files that should be removed, and the .dme is also outdated as heck.

code/game/objects/items/radio/detpack.dm
```diff
@@ -1,349 +1,349 @@
...
-		if(isobj(plant_target))
-			plant_target = null
-			if(!istype(plant_target,/obj/vehicle/multitile/root/cm_armored))
-				qdel(plant_target)
-	qdel(src)
-
-
-/obj/item/radio/detpack/attack(mob/M as mob, mob/user as mob, def_zone)
-	return
+/obj/item/radio/detpack
```
Rohesie: Whitespace / line ending changes. Revert.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	if(user == gunner)
+		user.forceMove(get_turf(src))
+		gunner = null
+		operators -= user
+
+/obj/vehicle/tank/relaymove(mob/user, direction)
+	if(user.incapacitated() || user != pilot)
+		to_chat(user, "You can't reach the gas pedal from down here, maybe try manning the driver's seat?")
+		return
+	if(world.time > last_move_time + move_delay)
```
Rohesie: Do an early return with `if(world.time < last_move_time + move_delay)`, so the previous message doesn't get spammy either.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	var/facing = get_dir(src, A)
+	var/turf/temp = get_turf(loc)
+	var/turf/T = get_turf(loc)
+	A.tank_collision(src, facing, T, temp)
+	if(isliving(A) && pilot)
+		log_attack("[pilot] drove over [A] with [src]")
+
+/client/MouseDown(object, location, control, params)
+	. = ..()
+	if(istype(mob.loc, /obj/vehicle/tank)) //mob is a var that exists on all clients anyway
```
Rohesie: This could be sending a signal instead, which would be registered by the tank once the mob enters the gunner seat, and unregister when it leaves it.
Let's see what @ninjanomnom says about it.

Also, there's no need to redefine it here. Add that coded in `code\_onclick\drag_drop.dm` where it's defined, instead of spreading little bits on several files.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	P.generate_bullet(new ammo.default_ammo)
+	log_combat(user, user, "fired the [src].")
+	log_explosion("[user] fired the [src] at [AREACOORD(loc)].")
+	P.fire_at(T, owner, src, P.ammo.max_range, P.ammo.shell_speed)
+	if(!CONFIG_GET(flag/tank_mouth_noise))
+		playsound(get_turf(owner), pick(fire_sounds), 60, 1)
+	ammo.current_rounds--
+	return TRUE
+
+/obj/item/tank_weapon/proc/can_fire(var/turf/T)
```
Rohesie:
```suggestion
/obj/item/tank_weapon/proc/can_fire(turf/T)
```
No need to have `var/` on any proc variables. It's implied already.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	if(get_dist(T, src) <= 3 && owner.gunner)
+		to_chat(owner.gunner, "Firing your main gun here could damage the tank!")
```
Rohesie: Why are you checking for `owner.gunner`? If it's a sanity check, remove it. We want it to runtime, screech and fail if it tries to fire without a gunner owner so we can fix it.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	. = ..()
+	update_icon()
+	if(world.time > last_drive_sound + 2 SECONDS)
+		playsound(src, 'sound/ambience/tank_driving.ogg', vol = 20, sound_range = 30)
+		last_drive_sound = world.time
+
+/obj/vehicle/tank/update_icon() //To show damage, gun firing, whatever. We need to re apply the gun turret overlay.
+	var/icon/I = icon(icon,icon_state,dir)
+	bound_width = I.Width() //Adjust the hitbox in case admins want to make an OMEGAtank
+	bound_height = I.Height()
```
Rohesie: Ninjanomnom had asked to remove this one, as the `bound_height` and `bound_width` should be standard, and it should be manually changed should an admin decide to edit it in-game. Or have a custom proc for them to do it instead.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	vis_contents = null
+	if(!turret_overlay) //We have to do this because of BYOND's innate dir setting habits, in byond, if you have an overlay'd object its direction can ONLY be which way the source is facing. So, we improvise.. ((yes this is kinda shit))
```
Rohesie: By ninjanomnom: "Comment is wrong, you can make an overlay from an image to get around this bug"

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		return
+	if(main_cannon_dir)
+		turret_overlay.setDir(main_cannon_dir)
+		if(main_cannon_dir == WEST || main_cannon_dir == SOUTHWEST|| main_cannon_dir == NORTHWEST)
+			turret_overlay.pixel_x = pixel_x -19
```
Rohesie: He called this brittlecode. We'll have to look for an alternative here.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		else
+			turret_overlay.pixel_x = 0
+	vis_contents += turret_overlay
+
+/obj/vehicle/tank/attack_hand(mob/user)
+	if(pilot && gunner)
+		to_chat(user, "You are unable to enter [src] because all of its seats are occupied!")
+		return ..()
+	if(user in operators)
```
Rohesie: `Won't this make it so if pilot and gunner are full you can't exit the tank?`

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		exit_tank(user)
+		return
+	var/position = alert("What seat would you like to enter?.",name,"pilot","gunner") //God I wish I had radials to work with
```
Rohesie: `Provide options based on what's available, though you'll still need to check the position wasn't filled since the options were shown to the player`

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		return FALSE
+	if(!allowed(M))
+		to_chat(M, "<span class='warning'>Access denied.</span>")
+		return FALSE
+	var/obj/item/offhand = M.get_inactive_held_item()
+	if(offhand && !(offhand.flags_item & (NODROP|DELONDROP)))
+		to_chat(M, "<span class='warning'>You need your hands free to climb on [src].</span>")
+		return FALSE
+
+	if(!M.mind || !(!M.mind.cm_skills || M.mind.cm_skills.large_vehicle >= SKILL_LARGE_VEHICLE_TRAINED))
```
Rohesie:
```suggestion
	if(!M.mind?.cm_skills || M.mind.cm_skills.large_vehicle < SKILL_LARGE_VEHICLE_TRAINED)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		M.visible_message("<span class='notice'>[M] fumbles around figuring out how to get into the [src].</span>",
+		"<span class='notice'>You fumble around figuring out how to get into [src].</span>")
+	var/time = 10 SECONDS - (2 SECONDS * M.mind.cm_skills.large_vehicle)
+	if(do_after(M, time, TRUE, src, BUSY_ICON_BUILD))
+		return TRUE
+
+/obj/vehicle/tank/proc/enter(var/mob/user, var/position) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
+	if(!user || !position)
```
Rohesie: Remove this sanity check. Let it error nicely if it happens.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		if("gunner")
+			gunner = user
+
+
+/obj/vehicle/tank/proc/remove_all_players()
+	for(var/M in operators)
+		exit_tank(M)
+
+/obj/vehicle/tank/proc/exit_tank(var/mob/user) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
+	if(!user)
```
Rohesie: Same, remove sanity check.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+
+/obj/vehicle/tank/relaymove(mob/user, direction)
+	if(user.incapacitated() || user != pilot)
+		to_chat(user, "You can't reach the gas pedal from down here, maybe try manning the driver's seat?")
+		return
+	if(world.time > last_move_time + move_delay)
+		. = step(src, direction)
+	update_icon()
+
+/obj/vehicle/tank/Bump(atom/A, yes)
```
Rohesie:
```suggestion
/obj/vehicle/tank/Bump(atom/A)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		handle_fire_main(A) //CTRL click to fire your big tank gun you can change any of these parameters here to hotkey for other shit :)
+		return
+	if(modifiers["middle"])
+		return
+	if(modifiers["alt"])
+		return
+	handle_fire(A)
+
+/obj/vehicle/tank/proc/handle_fire(var/atom/A)
+	if(!minigun && gunner)
```
Rohesie: If we use a signal for this, then checking for `minigun` and `gunner` is not needed.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		to_chat(gunner, "[src]'s minigun hardpoint spins pathetically. Maybe you should install a minigun on this tank?")
+		return FALSE
+	firing_target = A
+	playsound(get_turf(src), 'sound/weapons/tank_minigun_start.ogg', 60, 1)
+	firing_minigun = TRUE
+	START_PROCESSING(SSfastprocess, src)
+
+/obj/vehicle/tank/proc/handle_fire_main(var/atom/A) //This is used to shoot your big ass tank cannon, rather than your small MG
+	if(!main_cannon && gunner)
```
Rohesie: Same here, no need for this check if we unregister the signal once any of these conditions is no longer true.

code/_onclick/click.dm
```diff
@@ -128,9 +128,7 @@
 	if(next_move > world.time)
 		return
 
-	if(istype(loc, /obj/vehicle/multitile/root/cm_armored))
-		var/obj/vehicle/multitile/root/cm_armored/N = loc
-		N.click_action(A, src, params)
+	if(istype(loc, /obj/vehicle))
```
Rohesie:
```suggestion
	if(SEND_SIGNAL(src, COMSIG_MOB_CLICKON_NO_MODIFIERS, A, params) & COMSIG_MOB_CANCEL_CLICKON_NO_MODIFIERS)
```
Then register it on tank enter and unregister it on tank exit.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+/obj/item/tank_weapon/minigun
+	name = "TGS 3 XENOCRUSHER tank machine gun"
+	desc = "A much better gun that shits out bullets at ridiculous speeds, don't get in its way!"
+	icon = 'icons/obj/hardpoint_modules.dmi'
+	icon_state = "m56_cupola"
+	ammo = new /obj/item/ammo_magazine/tank/ltaaap_minigun
+	fire_sounds = list('sound/weapons/tank_minigun_loop.ogg')
+	cooldown = 2 //Minimal cooldown
+
+/obj/item/tank_weapon/proc/fire(var/atom/T,var/mob/user)
```
Rohesie:
```suggestion
/obj/item/tank_weapon/proc/fire(atom/T, mob/user)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	ammo.current_rounds--
+	return TRUE
+
+/obj/item/tank_weapon/proc/can_fire(var/turf/T)
+	if(get_dist(T, src) <= 3 && owner.gunner)
+		to_chat(owner.gunner, "Firing your main gun here could damage the tank!")
+		return FALSE
+	return TRUE
+
+/obj/item/tank_weapon/minigun/can_fire(var/turf/T)
```
Rohesie:
```suggestion
/obj/item/tank_weapon/minigun/can_fire(turf/T)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		to_chat(user, "[pilot] is already controlling [src]!")
+		return
+	if(gunner && position == "gunner")
+		to_chat(user, "[gunner] is already controlling [src]!")
+		return
+	if(can_enter(user)) //OWO can they enter us????
+		to_chat(user, "You climb into [src] as a [position]!")
+		return enter(user, position) //Yeah i could do this with a define, but this way we're not using multiple things
+
+/obj/vehicle/tank/proc/can_enter(var/mob/living/carbon/M) //NO BENOS ALLOWED
```
Rohesie:
```suggestion
/obj/vehicle/tank/proc/can_enter(mob/living/carbon/M) //NO BENOS ALLOWED
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+		return FALSE
+
+	if(M.mind?.cm_skills?.large_vehicle < SKILL_LARGE_VEHICLE_TRAINED)
+		M.visible_message("<span class='notice'>[M] fumbles around figuring out how to get into the [src].</span>",
+		"<span class='notice'>You fumble around figuring out how to get into [src].</span>")
+	var/time = 10 SECONDS - (2 SECONDS * M.mind.cm_skills.large_vehicle)
+	if(do_after(M, time, TRUE, src, BUSY_ICON_BUILD))
+		return TRUE
+
+/obj/vehicle/tank/proc/enter(var/mob/user, var/position) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
```
Rohesie:
```suggestion
/obj/vehicle/tank/proc/enter(mob/user, position) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+			pilot = user
+		if("gunner")
+			gunner = user
+
+
+/obj/vehicle/tank/proc/remove_all_players()
+	for(var/M in operators)
+		exit_tank(M)
+
+/obj/vehicle/tank/proc/exit_tank(var/mob/user) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
```
Rohesie:
```suggestion
/obj/vehicle/tank/proc/exit_tank(mob/user) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+
+/client/MouseUp(object, location, control, params) ///Oh boy click code, my favourite! This adds support for click n' hold gun firing action
+	. = ..()
+	if(istype(mob.loc, /obj/vehicle/tank))
+		var/obj/vehicle/tank/our_tank = mob.loc //I could refactor this to be a var on the mob, if you'd prefer
+		our_tank.onMouseUp(object,mob)
+
+
+//Combat, tank guns, all that fun stuff
+/obj/vehicle/tank/proc/onMouseDown(var/atom/A, mob/user, params)
```
Rohesie:
```suggestion
/obj/vehicle/tank/proc/onMouseDown(atom/A, mob/user, params)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	if(modifiers["ctrl"])
+		handle_fire_main(A) //CTRL click to fire your big tank gun you can change any of these parameters here to hotkey for other shit :)
+		return
+	if(modifiers["middle"])
+		return
+	if(modifiers["alt"])
+		return
+	handle_fire(A)
+
+/obj/vehicle/tank/proc/handle_fire(var/atom/A)
```
Rohesie:
```suggestion
/obj/vehicle/tank/proc/handle_fire(atom/A)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	if(!minigun && gunner)
+		to_chat(gunner, "[src]'s minigun hardpoint spins pathetically. Maybe you should install a minigun on this tank?")
+		return FALSE
+	firing_target = A
+	playsound(get_turf(src), 'sound/weapons/tank_minigun_start.ogg', 60, 1)
+	firing_minigun = TRUE
+	START_PROCESSING(SSfastprocess, src)
+
+/obj/vehicle/tank/proc/handle_fire_main(var/atom/A) //This is used to shoot your big ass tank cannon, rather than your small MG
```
Rohesie:
```suggestion
/obj/vehicle/tank/proc/handle_fire_main(atom/A) //This is used to shoot your big ass tank cannon, rather than your small MG
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+			main_cannon_dir = get_dir(src, firing_target)
+			update_icon()
+		else
+			stop_firing()
+	if(firing_minigun)
+		if(minigun.fire(firing_target, gunner))
+			minigun_dir = get_dir(src, firing_target) //Set the gun dir
+			update_icon()
+
+/obj/vehicle/tank/proc/onMouseUp(var/atom/A, mob/user)
```
Rohesie:
```suggestion
/obj/vehicle/tank/proc/onMouseUp(atom/A, mob/user)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,381 @@
...
+	if(M.mind?.cm_skills?.large_vehicle)
+		fumbling_time -= 2 SECONDS * M.mind.cm_skills.large_vehicle
+	if(!do_after(M, fumbling_time, TRUE, 5, BUSY_ICON_HOSTILE) || !M.Adjacent(hatch))
+		return
+	exit_tank(occupant)
+	M.visible_message("<span class='warning'>[M] forcibly pulls [occupant] out of [src].</span>",
+					"<span class='notice'>you forcibly pull [occupant] out of [src].</span>", null, 6)
+	occupant.KnockDown(4)
+
+/obj/vehicle/tank/Bumped(var/atom/A) //Don't ..() because then you can shove the tank into a wall.
```
Rohesie:
```suggestion
/obj/vehicle/tank/Bumped(atom/A) //Don't ..() because then you can shove the tank into a wall.
```

code/modules/vehicles/vehicle.dm
```diff
@@ -252,3 +254,8 @@
 //-------------------------------------------------------
 /obj/vehicle/proc/update_stats()
 	return
+
+/obj/vehicle/proc/take_damage(var/damage)
```
Rohesie:
```suggestion
/obj/vehicle/proc/take_damage(damage)
```

code/modules/vehicles/vehicle.dm
```diff
@@ -252,3 +254,8 @@
...
+	if(damage)
```
Rohesie: This proc shouldn't be called without damage. No need for the check.

code/modules/vehicles/tank.dm
Kmc2000: BYOND won't let you do inputs of 1 thing, it just automatically makes you choose the one option.

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,76 @@
...
+/obj/machinery/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp)
+	. = ..()
+	var/damage = .
+	take_damage(damage)
+
+/turf/closed/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp)
+	. = ..()
+	var/damage = .
+	if(prob(15))
+		ex_act(damage)
```
Rohesie: `ex_act` uses `severity`, which goes from 1 to 3 typically, not damage.

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,76 @@
...
+
+/mob/living/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp) //If theyre alive, yeet them
+	if(stat == DEAD) //We don't care about the dead
+		return
+	if(!roadtrafficaccident.demolish_on_ram)
+		return 0 //FALSE, also 0 damage.
+
+	if(loc == roadtrafficaccident.loc) // trodden over.
```
Rohesie: Tonk brok

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,76 @@
...
+			throw_at(T, 3, 2, roadtrafficaccident, 0)
+		else
+			throw_at(T, 3, 2, roadtrafficaccident, 1)
+		if(!knocked_down)
+			KnockDown(1)
+		apply_damage(rand(10, 15), BRUTE)
+		visible_message("<span class='danger'>[roadtrafficaccident] bumps into [src], throwing [p_them()] away!</span>", "<span class='danger'>[roadtrafficaccident] violently bumps into you!</span>")
+
+/mob/living/carbon/Xenomorph/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp)
+	if(lying || loc == roadtrafficaccident.loc)
```
Rohesie: Brok

code/modules/vehicles/multitile/multitile.dm
```diff
@@ -71,23 +71,23 @@ Vehicles are placed on the map by a spawner or admin verb
...
 	desc = "You shouldn't see this"
 
-/obj/vehicle/multitile/relaymove(mob/user, direction)
+/obj/vehicle/relaymove(mob/user, direction)
 	return
 
 //Hitboxes, do notthing but move with the root object and take up space
 //All interactions like bullets or whatever should be passed up to the root object
-/obj/vehicle/multitile/hitbox
+/obj/vehicle/hitbox
```
Rohesie: Is this supposed to continue existing?

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+	desc = "A gun that works about 50% of the time, but at least it's open source! It fires tank shells."
+	icon = 'icons/obj/hardpoint_modules.dmi'
+	icon_state = "glauncher"
+	var/obj/item/ammo_magazine/ammo
+	var/obj/vehicle/tank/owner
+	var/list/fire_sounds = list('sound/weapons/tank_cannon_fire1.ogg', 'sound/weapons/tank_cannon_fire2.ogg')
+	var/default_ammo = /obj/item/ammo_magazine/tank/ltb_cannon //What do we start with?
+	var/list/accepted_ammo = list(/obj/item/ammo_magazine/tank/tank_glauncher) //Alternative ammo types that we'll accept. The default ammo is in this by default
+	var/fire_delay
+	var/cooldown = 60 //6 second weapons cooldown
```
Rohesie:
```suggestion
	var/cooldown = 6 SECONDS
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+	accepted_ammo += default_ammo
+
+/obj/item/tank_weapon/secondary_weapon
+	name = "TGS 3 XENOCRUSHER tank machine gun"
+	desc = "A much better gun that shits out bullets at ridiculous speeds, don't get in its way!"
+	icon = 'icons/obj/hardpoint_modules.dmi'
+	icon_state = "m56_cupola"
+	default_ammo = /obj/item/ammo_magazine/tank/ltaaap_minigun
+	fire_sounds = list('sound/weapons/tank_minigun_loop.ogg')
+	cooldown = 2 //Minimal cooldown
```
Rohesie:
```suggestion
	cooldown = 0.2 SECONDS //Minimal cooldown
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+	accepted_ammo = list(/obj/item/ammo_magazine/tank/towlauncher,/obj/item/ammo_magazine/tank/m56_cupola,/obj/item/ammo_magazine/tank/flamer,/obj/item/ammo_magazine/tank/tank_slauncher)
+
+/obj/item/tank_weapon/proc/fire(atom/T,mob/user)
+	if(!can_fire(T))
+		return FALSE
+	if(ammo.current_rounds <= 0)
```
Rohesie: Check for rounds after the cooldown so we can give it a click sound for feedback without spam.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+		return FALSE
+	if(world.time < lastfired + cooldown)
+		return FALSE
+	lastfired = world.time
+	var/obj/item/projectile/P = new
+	P.generate_bullet(new ammo.default_ammo)
+	log_combat(user, user, "fired the [src].")
```
Rohesie:
```suggestion
	log_combat(user, T, "fired the [src].")
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+	log_explosion("[user] fired the [src] at [AREACOORD(loc)].")
```
Rohesie: Do all weapons generate explosions?

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+// You can set max_passengers to something different (eg, 2 for a jeep) if you don't want loads of marines to pile in														//
+// I have set the layer of this tank high so that it layers over lights. This is an issue with multi-tile sprites in byond 													//
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/obj/vehicle/tank
+	name = "MK-1 'friendly fire' prototype tank"
+	desc = "A gigantic wall of metal designed for maximum Xeno destruction. Click it with an open hand to enter as a pilot or a gunner."
+	icon = 'icons/obj/tank.dmi'
+	icon_state = "tank"
+	layer = 5
```
Rohesie: Use defines.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+	anchored = FALSE
+	can_buckle = FALSE
+	req_access = list(ACCESS_MARINE_TANK)
+	move_delay = 4
```
Rohesie:
```suggestion
	move_delay = 0.4 SECONDS
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,388 @@
...
+	. = ..()
+	to_chat(user, "<b>To fire its main cannon, <i>ctrl</i> click a tile</b>")
+	to_chat(user, "<b>To fire its secondary_weapon, click a tile</b>")
+
+/obj/turret_overlay
+	name = "Tank gun turret"
+	desc = "The shooty bit on a tank."
+	icon = 'icons/obj/tank_gun.dmi'
+	icon_state = "turret"
+	layer = 5.1
```
Rohesie: Use defines.

Rohesie:
Travis check.

Rohesie:
Travis failing.

code/modules/vehicles/tank.dm
Rohesie: This is still a thing.

code/modules/vehicles/multitile/multitile.dm
Rohesie: This is still a thing.

code/modules/vehicles/tank.dm
Rohesie: This is still a thing.

code/game/objects/items/radio/detpack.dm
Rohesie: This is still a thing.

Rohesie:
The map change is broken and needs to be redone.

code/modules/admin/admin_verbs.dm
```diff
@@ -884,7 +884,7 @@
 
 	else if(istype(whom, /client))
 		recipient = whom
-		
+
```
Rohesie:
```suggestion
```

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,75 @@
...
+		var/target_dir = turn(roadtrafficaccident.dir, 180)
+		temp = get_step(roadtrafficaccident.loc, target_dir)
+		T = temp
+		target_dir = turn(roadtrafficaccident.dir, 180)
+		T = get_step(T, target_dir)
+		face_atom(T)
+		throw_at(T, 3, 2, roadtrafficaccident, 1)
+		apply_damage(rand(5, 7.5), BRUTE)
+		return
+	if(!lying)
```
Rohesie: Do an early return instead:
```
if(lying)
	return 0
```

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,75 @@
...
+			KnockDown(1)
+		var/target_dir = turn(roadtrafficaccident.dir, 180)
+		temp = get_step(roadtrafficaccident.loc, target_dir)
+		T = temp
+		target_dir = turn(roadtrafficaccident.dir, 180)
+		T = get_step(T, target_dir)
+		face_atom(T)
+		throw_at(T, 3, 2, roadtrafficaccident, 1)
+		apply_damage(rand(5, 7.5), BRUTE)
+		return
```
Rohesie: Is this not supposed to return damage?

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+
+/obj/vehicle/tank/tiny //SQUEEEE
+	name = "Mk-4 'pint pot' tank"
+	desc = "An adorable chunk of metal with an alarming amount of firepower designed to crush, immolate, destroy and maim anything that nanotrasen wants it to. This model contains advanced bluespace technology which allows a tardis like amount of room on the inside"
+	icon = 'icons/obj/tinytank.dmi'
+	turret_icon = 'icons/obj/tinytank_gun.dmi'
+	icon_state = "tank"
+	pixel_x = -16 //Stops marines from treading on it...d'aww
+	pixel_y = -8
+	max_passengers = 2 //Bluespace's one hell of a drug.
```
Rohesie:
```suggestion
	max_passengers = 0 //Bluespace's one hell of a drug.
```
Let's start small.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+	name = "M557 Armoured Personnel Carrier"
+	desc = "A heavily armoured vehicle with light armaments. Though it doesn't pack the punch of a tank its utility launcher can provide cover for troops and allow for rapid deployment with minimal casualties."
+	icon = 'icons/obj/tinytank.dmi'
+	turret_icon = 'icons/obj/tinytank_gun.dmi'
+	turret_icon_state = "apc_turret"
+	icon_state = "apc"
+	move_delay = 0.5 SECONDS //Faster because it's lightweight
+	pixel_x = -16
+	pixel_y = -8
+	max_passengers = 4 //It's a big carrier designed to get marines into the operation safely.
```
Rohesie:
```suggestion
	max_passengers = 2 //It's a small carrier designed to get marines back from the operation safely.
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+	to_chat(user, "<b><span class='notice>To fire its main cannon, <i>ctrl</i> click a tile</b></span>")
+	to_chat(user, "<b><span class='notice>To fire its secondary weapon, click a tile</b></span>")
+	to_chat(user, "<i><span class='notice>It's currently holding [passengers.len] / [max_passengers] passengers</i></span>")
+
+/obj/turret_overlay
+	name = "Tank gun turret"
+	desc = "The shooty bit on a tank."
+	icon = 'icons/obj/tank_gun.dmi'
+	icon_state = "turret"
+	layer = ABOVE_MOB_LAYER+0.1
```
Rohesie: Make a define like /tg/'s `LARGE_MOB_LAYER`

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+	. = ..()
+	turret_overlay = new()
+	turret_overlay.icon = turret_icon
+	turret_overlay.icon_state = turret_icon_state
+	update_icon()
+	if(primary_weapon_type)
+		primary_weapon = new primary_weapon_type(src) //Make our guns
+	if(secondary_weapon_type)
+		secondary_weapon = new secondary_weapon_type(src)
+	primary_weapon?.owner = src
```
Rohesie: Put this inside the `if(primary_weapon_type)`
A `null = src` operation would be bad.
Same for the secondary one.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+	secondary_weapon?.owner = src
+	GLOB.tank_list += src
+
+/obj/vehicle/tank/Destroy()
+	qdel(turret_overlay)
```
Rohesie: Null the references.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+	qdel(primary_weapon)
+	qdel(secondary_weapon)
+	for(var/x in contents) //Yeet the occupants out so they arent deleted
```
Rohesie: This should be a proc that gets called before the `qdel` of the tank, though a reference to it for safety wouldn't be bad inside of here.

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+			mob.apply_effect(6, WEAKEN)
+			mob.apply_effect(6, STUTTER)
+			to_chat(mob, "<span class='warning'>You violently dive out of [src] as it explodes behind you!</span>")
+	playsound(get_turf(src), 'sound/weapons/tank_cannon_fire1.ogg', 100, 1) //Explosion sound
+	. = ..()
+
+/obj/vehicle/tank/Move()
+	. = ..()
+	update_icon()
+	if(world.time > last_drive_sound + 2 SECONDS)
```
Rohesie: Early return:
```
if(world.time < last_drive_sound + 2 SECONDS)
	return
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+		playsound(src, 'sound/ambience/tank_driving.ogg', vol = 20, sound_range = 30)
+		last_drive_sound = world.time
+
+/obj/vehicle/tank/update_icon() //To show damage, gun firing, whatever. We need to re apply the gun turret overlay.
+	vis_contents = null
```
Rohesie: Is this supposed to be nulled on each icon update?

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+		if(30 to 59)
+			damage_overlay.icon_state = "damage_medium"
+		if(60 to 70)
+			damage_overlay.icon_state = "damage_minor_medium"
+		if(71 to 90)
+			damage_overlay.icon_state = "damage_minor"
+	add_overlay(damage_overlay)
+
+/obj/vehicle/tank/attack_hand(mob/user)
+	if(!ishuman(user) || isxeno(user)) //Aliens can't get in a tank...as hilarious as that would be
```
Rohesie:
```suggestion
	if(!ishuman(user)) //Aliens can't get in a tank...as hilarious as that would be
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+/obj/vehicle/tank/proc/remove_all_players()
+	for(var/M in operators)
+		exit_tank(M)
+
+/obj/vehicle/tank/proc/exit_tank(mob/user) //By this point, we've checked that the seats are actually empty, so we won't need to do that again HOPEFULLY
+	if(user == pilot)
+		user.forceMove(get_turf(src))
+		pilot = null
+		operators -= user
+	if(user == gunner)
```
Rohesie:
```suggestion
	else if(user == gunner)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+		user.forceMove(get_turf(src))
+		gunner = null
+		operators -= user
+	if(user in passengers)
```
Rohesie:
```suggestion
	else if(user in passengers)
```

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,491 @@
...
+		user.forceMove(get_turf(src))
+		passengers -= user
+		operators -= user
+	to_chat(user, "<span class ='notice'>You hop out of [src] and slam its hatch shut behind you.</span>")
+
+/obj/vehicle/tank/relaymove(mob/user, direction)
```
Rohesie: You could implement a check here to prevent diagonal movement.

Rohesie:
Could you also remove the NicBoone sprites in the same PR?

DominikPanic:
```
[21:20:32] Runtime in garbage.dm, line 321: /obj/vehicle/tank/apc destroy proc was called multiple times, likely due to a qdel loop in the Destroy logic
proc name: qdel (/proc/qdel)
usr: David Stormwell/(David 'Liquid' Stormwell)
usr.loc: (Hangar (77, 65, 3))
src: null
call stack:
qdel(M557 Armoured Personnel Carrie... (/obj/vehicle/tank/apc), 0)
M557 Armoured Personnel Carrie... (/obj/vehicle/tank/apc): explode()
M557 Armoured Personnel Carrie... (/obj/vehicle/tank/apc): healthcheck()
M557 Armoured Personnel Carrie... (/obj/vehicle/tank/apc): take damage(42)
M557 Armoured Personnel Carrie... (/obj/vehicle/tank/apc): take damage(42)
M557 Armoured Personnel Carrie... (/obj/vehicle/tank/apc): bullet act(the heavy pistol bullet (/obj/item/projectile))
M557 Armoured Personnel Carrie... (/obj/vehicle/tank/apc): bullet act(the heavy pistol bullet (/obj/item/projectile))
the heavy pistol bullet (/obj/item/projectile): scan a turf(the floor (75,68,3) (/turf/open/floor/almayer))
the heavy pistol bullet (/obj/item/projectile): follow flightpath(5, -3, 4, 20)
```

factoryman942:
clicking the tonk and walking away lets you get in from a distance

code/modules/mob/living/carbon/xenomorph/castes/boiler/boiler.dm
```diff
@@ -1,56 +1,57 @@
...
-		see_in_dark = 20
-
-// ***************************************
-// *********** Gibbing behaviour
-// ***************************************
-/mob/living/carbon/xenomorph/boiler/proc/gib_explode()
-	visible_message("<span class='danger'>[src] begins to bulge grotesquely, and explodes in a cloud of corrosive gas!</span>")
-	smoke.set_up(2, get_turf(src))
-	smoke.start()
+/mob/living/carbon/xenomorph/boiler
```
Rohesie: Bad rebase.

code/__DEFINES/layers.dm
```diff
@@ -1,163 +1,164 @@
...
-
-#define HUD_PLANE 21
-#define HUD_LAYER 21
-#define ABOVE_HUD_PLANE 22
-#define ABOVE_HUD_LAYER 22
-
-#define SPLASHSCREEN_LAYER 23
+
+//Filters
+#define AMBIENT_OCCLUSION filter(type="drop_shadow", x=0, y=-2, size=4, border=4, color="#04080FAA")
```
Rohesie: Bad merge.

ninjanomnom: It's accidental line ending changes

Rohesie:
Travis check.

code/_onclick/drag_drop.dm
```diff
@@ -23,6 +23,9 @@
 /client/MouseDown(object, location, control, params)
 	if(mouse_down_icon)
 		mouse_pointer_icon = mouse_down_icon
+	if(istype(mob.loc, /obj/vehicle/tank)) //This code is related to tank + vehicle autofire code, as seen in tank.dm
+		var/obj/vehicle/tank/our_tank = mob.loc
+		our_tank.onMouseDown(object,mob,params)
```
ninjanomnom: You should not be having tank code here in client code

code/_onclick/drag_drop.dm
```diff
@@ -38,4 +41,7 @@
 			middragatom = src_object
 		else
 			middragtime = 0
-			middragatom = null
\ No newline at end of file
+			middragatom = null		mouse_pointer_icon = mouse_up_icon
+	if(istype(mob.loc, /obj/vehicle/tank))//This code is related to tank + vehicle autofire code, as seen in tank.dm
+		var/obj/vehicle/tank/our_tank = mob.loc
+		our_tank.onMouseUp(object,mob)
```
ninjanomnom: Ditto

code/game/objects/machinery/doors/multi_tile.dm
```diff
@@ -4,7 +4,7 @@
 
 /obj/machinery/door/airlock/multi_tile/close() //Nasty as hell O(n^2) code but unfortunately necessary
 	for(var/turf/T in locs)
-		for(var/obj/vehicle/multitile/M in T)
+		for(var/obj/vehicle/M in T)
 			if(M) return FALSE
```
ninjanomnom: Does this need to exist? The airlock should assumingly be not closing if any dense object is blocking it

If it does need to exist then this internal loop should be replaced with
```
if(locate(/obj/vehicle) in T)
  return FALSE
```

code/modules/admin/admin_verbs.dm
```diff
@@ -1130,7 +1129,7 @@
 	if(!check_rights(R_ADMIN))
 		return
 
-	for(var/obj/vehicle/multitile/root/cm_armored/CA in GLOB.tank_list)
+	for(var/obj/vehicle/tank/CA in GLOB.tank_list)
```
ninjanomnom: There are non tanks in the tank_list?

code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -30,7 +30,7 @@
 	var/mob/living/carbon/C = owner
 	if(!length(C.stomach_contents))
 		if(!silent)
-			to_chat(C, "<span class='warning'>There's nothing in our belly that needs regurgitating.</span>")
+			to_chat(C, "<span class='warning'>There's nothing in your belly that needs regurgitating.</span>")
```
ninjanomnom: Why are you doing this change in tank pr. You have unrelated changes in a lot of places

code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -520,11 +522,11 @@
...
-	if(istype(A, /obj/vehicle/multitile/root/cm_armored))
-		var/obj/vehicle/multitile/root/cm_armored/R = A
-		R.take_damage_type( (1 / newacid.acid_strength) * 20, "acid", X)
-		X.visible_message("<span class='xenowarning'>\The [X] vomits globs of vile stuff at \the [R]. It sizzles under the bubbling mess of acid!</span>", \
-			"<span class='xenowarning'>We vomit globs of vile stuff at \the [R]. It sizzles under the bubbling mess of acid!</span>", null, 5)
+	if(istype(A, /obj/vehicle))
+		var/obj/vehicle/carcrash = A
+		carcrash.take_damage(30)
+		X.visible_message("<span class='xenowarning'>\The [X] vomits globs of vile stuff at \the [carcrash]. It sizzles under the bubbling mess of acid!</span>", \
+			"<span class='xenowarning'>You vomit globs of vile stuff at \the [carcrash]. It sizzles under the bubbling mess of acid!</span>", null, 5)
```
ninjanomnom: Yeah this looks like a bad conflict resolution?

code/modules/mob/living/carbon/xenomorph/castes/crusher/crusher.dm
```diff
@@ -1,313 +1,314 @@
...
-			return FALSE
-
-		if(anchored)
-			if(X.charge_speed < CHARGE_SPEED_BUILDUP * CHARGE_TURFS_TO_CHARGE)
-				X.stop_momentum(X.charge_dir)
-				return FALSE
-			else
-				X.visible_message("<span class='danger'>[X] crushes [src]!</span>",
+			visible_message("<span class='danger'>[src] rams into [carcrash] and skids to a halt!</span>",
+			"<span class='xenowarning'>You ram into [carcrash] and skid to a halt!</span>")
```
ninjanomnom: More bad conflict resolution

code/modules/vehicles/attachments.dm
```diff
@@ -0,0 +1,148 @@
...
+	point_cost = 5
+
+/obj/item/ammo_magazine/tank/tank_slauncher/update_icon()
+	icon_state = "slauncher_[current_rounds <= 0 ? "0" : "1"]"
+
+///////////////
+// AMMO MAGS // END
+///////////////
+
+/obj/vehicle/tank/attackby(obj/item/I, mob/user) //This handles reloading weapons, or changing what kind of mags they'll accept. You can have a passenger do this
```
ninjanomnom: This is in the wrong file

code/modules/vehicles/attachments.dm
```diff
@@ -0,0 +1,148 @@
...
+		var/time = 8 SECONDS - (1 SECONDS * M.mind.cm_skills.large_vehicle)
+		to_chat(user, "You start to swap out [primary_weapon]'s magazine...")
+		if(do_after(M, time, TRUE, src, BUSY_ICON_BUILD))
+			playsound(get_turf(src), 'sound/weapons/working_the_bolt.ogg', 100, 1)
+			primary_weapon.ammo.forceMove(get_turf(user))
+			primary_weapon.ammo = I
+			to_chat(user, "You load [I] into [primary_weapon] with a satisfying click.")
+			user.transferItemToLoc(I,src)
+		return
+	if(is_type_in_list(I, secondary_weapon?.accepted_ammo))
```
ninjanomnom: You can write this so you don't need duplicate code blocks. Really you need to break apart the code in this proc in general, you have too much happening here

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,76 @@
+/atom/proc/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp)
```
ninjanomnom: If you have this proc you need to deal with all those CanPass procs doing similar things regarding collision damage

code/modules/admin/admin_verbs.dm
Kmc2000: No

Rohesie: Then do this to avoid the typecheck:
```
for(var/i in GLOB.tank_list)
	var/obj/vehicle/tank/CA = i
```

code/_onclick/drag_drop.dm
Kmc2000: Rohesie's autofire shit should fix this anyway

ninjanomnom: This pr will have to wait until you don't have this code here one way or another

Rohesie: If you copypaste the signal from my PR you can include it in yours.
Poke me if you need help, it's pretty trivial to do.

code/game/objects/machinery/doors/airlock.dm
```diff
@@ -697,37 +697,34 @@
...
-		for(var/turf/turf in locs)
-			if(locate(/mob/living) in turf)
-			//	playsound(src.loc, 'sound/machines/buzz-two.ogg', 25, 0)	//THE BUZZING IT NEVER STOPS	-Pete
+
+	for(var/i in bounds())
+		var/atom/thing = i
+		if(isliving(thing))
+			if(safe)
 				spawn (60 + openspeed)
 					close()
```
ninjanomnom: Change this spawn to a timer while you're in this file

code/game/objects/machinery/doors/airlock.dm
```diff
@@ -697,37 +697,34 @@
...
-			if (iscarbon(M))
-				var/mob/living/carbon/C = M
-				var/datum/species/S = C.species
-				if(S?.species_flags & NO_PAIN)
-					M.emote("pain")
-			var/turf/location = src.loc
-			if(istype(location, /turf))
-				location.add_mob_blood(M)
-
+			else
```
ninjanomnom: You don't need this else since it always returns in the previous if

code/game/objects/machinery/doors/airlock.dm
```diff
@@ -697,37 +697,34 @@
...
+			return //Can't close with shit ontop
 	use_power(360)	//360 W seems much more appropriate for an actuator moving an industrial door capable of crushing people
 	if(istype(src, /obj/machinery/door/airlock/glass))
 		playsound(src.loc, 'sound/machines/windowdoor.ogg', 25, 1)
 	else
 		playsound(src.loc, 'sound/machines/airlock.ogg', 25, 0)
-	for(var/turf/turf in locs)
-		var/obj/structure/window/killthis = (locate(/obj/structure/window) in turf)
-		if(killthis)
-			killthis.ex_act(2)//Smashin windows
```
ninjanomnom: You lost the window smashing functionality

code/game/objects/structures/barricade.dm
```diff
@@ -100,7 +100,10 @@
 				return FALSE
 		return TRUE
 
-	if(istype(mover, /obj/vehicle/multitile))
+	if(istype(mover, /obj/vehicle))
+		var/obj/vehicle/roadtrafficaccident = mover
+		if(!roadtrafficaccident.demolish_on_ram)
+			return
```
ninjanomnom: You have to return true or false from this proc

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,76 @@
...
+		roadtrafficaccident.lastsound = world.time
+		roadtrafficaccident.take_damage(tank_damage)
+	return damage
+
+/obj/structure/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp)
+	. = ..()
+	var/damage = . //If it has a snowplow, add extra damage
+	take_damage(damage)
+
+/obj/machinery/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp)
```
ninjanomnom: `/obj/tank_collision(...)`? You have two procs doing the same thing here

code/modules/vehicles/collision.dm
```diff
@@ -0,0 +1,76 @@
...
+	else
+		throw_at(T, 3, 2, roadtrafficaccident, 1)
+	if(!knocked_down)
+		KnockDown(1)
+	apply_damage(rand(10, 15), BRUTE)
+	visible_message("<span class='danger'>[roadtrafficaccident] bumps into [src], throwing [p_them()] away!</span>", "<span class='danger'>[roadtrafficaccident] violently bumps into you!</span>")
+
+/mob/living/carbon/Xenomorph/tank_collision(obj/vehicle/roadtrafficaccident, facing, turf/T, turf/temp)
+	if(lying || loc == roadtrafficaccident.loc)
+		return ..()
```
ninjanomnom: Prefer to write new procs and such so that you always call parent at the top with `. = ..()`

code/modules/vehicles/interactions.dm
```diff
@@ -0,0 +1,90 @@
...
+		var/time = 8 SECONDS - (1 SECONDS * M.mind.cm_skills.large_vehicle)
+		to_chat(user, "You start to swap out [secondary_weapon]'s magazine...")
+		if(do_after(M, time, TRUE, src, BUSY_ICON_BUILD))
+			playsound(get_turf(src), 'sound/weapons/working_the_bolt.ogg', 100, 1)
+			secondary_weapon.ammo.forceMove(get_turf(user))
+			secondary_weapon.ammo = I
+			to_chat(user, "You load [I] into [secondary_weapon] with a satisfying click.")
+			user.transferItemToLoc(I,src)
+		return
+	. = ..()
```
ninjanomnom: `return ..()` when last

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,577 @@
...
+
+/obj/item/tank_weapon/Initialize()
+	. = ..()
+	ammo = new default_ammo(src)
+	accepted_ammo += default_ammo
+
+/obj/item/tank_weapon/secondary_weapon
+	name = "TGS 3 XENOCRUSHER tank machine gun"
+	desc = "A much better gun that shits out bullets at ridiculous speeds, don't get in its way!"
+	icon = 'icons/obj/hardpoint_modules.dmi'
```
ninjanomnom: This is on the base type

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,577 @@
...
+	range_safety_check = 0
+
+/obj/item/tank_weapon/apc_cannon
+	name = "MKV-7 utility payload launcher"
+	desc = "A double barrelled cannon which can rapidly deploy utility packages to the battlefield."
+	icon = 'icons/obj/hardpoint_modules.dmi'
+	icon_state = "APC uninstalled dualcannon"
+	default_ammo = /obj/item/ammo_magazine/tank/tank_slauncher
+	cooldown = 0.7 SECONDS //Minimal cooldown
+	accepted_ammo = list(/obj/item/ammo_magazine/tank/towlauncher,/obj/item/ammo_magazine/tank/flamer,/obj/item/ammo_magazine/tank/tank_slauncher, /obj/item/ammo_magazine/tank/tank_glauncher)
```
ninjanomnom: The default_ammo gets added to accepted_ammo so why is the same type in this list? Additionaly, why do both of these variables exist?

code/modules/vehicles/tank.dm
```diff
@@ -0,0 +1,577 @@
...
+	fire_sounds = list('sound/weapons/gun_shotgun_automatic.ogg', 'sound/weapons/gun_shotgun_light.ogg', 'sound/weapons/gun_shotgun_heavy.ogg')
+
+/obj/item/tank_weapon/proc/fire(atom/T,mob/user)
+	if(!can_fire(T))
+		return FALSE
+	if(world.time < lastfired + cooldown)
+		return FALSE
+	if(ammo.current_rounds <= 0)
```
ninjanomnom: Shouldn't can_fire check this?

psykzz:
@Kmc2000 are you okay to close this for #2622 ?

DominikPanic:
#2622 is now in a better state

PR #1249, opened 2019-04-07
Tweaks the previously pred-restricted snatch-facehugger-mid-air feature into an intentional one.

### LINKED ISSUE #1245: Supine marines can still grab facehuggers out of the air.
## Reproduction:
Attached is the log representing the issue:
https://gyazo.com/58ef41ddf2597bb8ad1fb9cde822dfb1

A knocked-down marine with substantial haloss buildup (he had been throttled many times by two xenos) was still able to grab a facehugger out of the air and squash it, despite being stunned. There was no sprite change, he did not momentarily stand up, he just squashed it while stunned. This doesn't seem like intended behavior. Either it bounced off his helmet (he had a standard helmet, so not likely) and the message was wrong, or this is a bug.

Ghommie:
## About The Pull Request
When the rred removal PR did go through, this was one of the few things accidentally left over. I thought this wouldn't make a bad feature with a little tweaking.
Made the "action" require throw mode on, have the active hand empty and not be incapacitated, tweaked the prob modifiers, and unrestricted this feature from the human

## Why It's Good For The Game
Improve, not remove. Closes #1245.

## Changelog
:cl:
add: Tweaked the unintended feature (left over from preds removal) of snatching and killing facehuggers mid-attaching into an intentional one. It also requires a free active hand, throw mode on, and can't be done while incapacitated.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
@@ -324,7 +324,7 @@
 
 	return TRUE
 
-/obj/item/clothing/mask/facehugger/proc/Attach(mob/living/carbon/M, self_done = FALSE)
+/obj/item/clothing/mask/facehugger/proc/Attach(mob/living/carbon/M)
 
 	throwing = FALSE
 	leaping = FALSE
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/facehuggers.dm
```diff
@@ -338,14 +338,27 @@
...
 		var/mob/living/carbon/Xenomorph/X = loc
 		X.dropItemToGround(src)
 		X.update_icons()
 
+	if(M.in_throw_mode && M.dir != dir && !M.incapacitated() && !M.get_active_held_item())
+		var/catch_chance = 50
+		if(M.dir == reverse_dir[dir])
+			catch_chance += 20
+		if(M.lying)
+			catch_chance -= 50
```
Rohesie: Chatching chance for a lying mob should be zero. Else you'll get nested hosts catching huggers.

Ghommie: Nested hosts are usually restrained/incapacitated and so can't do jack.
I were merely thinking about /tg/'s mobility code and crawling when letting this be, in the case we ever get there.

Rohesie: Hmm, fair enough.

PR #1253, opened 2019-04-08
Factions and squad rewrite

spookydonut:
## About The Pull Request
Makes factions a thing and squad generic and tied to factions

## Why It's Good For The Game
Opens up more flexible gamemode creation


code/controllers/stonedmc/subsystem/job.dm
```diff
@@ -97,6 +97,95 @@ SUBSYSTEM_DEF(job)
...
 
+/proc/handle_squad(mob/M, rank, latejoin = FALSE)
+	var/strict = FALSE
+	var/datum/squad/P = SSjob.squads[M.client.prefs.preferred_squad]
+	var/datum/squad/R = SSjob.squads[pick(SSjob.squads)]
+	if(M.client?.prefs?.be_special && (M.client.prefs.be_special & BE_SQUAD_STRICT))
+		strict = TRUE
+	switch(rank)
+		if("Squad Marine")
+			if(P && P.assign(M, rank))
```
Rohesie:
```suggestion
			if(P?.assign(M, rank))
```

code/controllers/stonedmc/subsystem/job.dm
```diff
@@ -97,6 +97,95 @@ SUBSYSTEM_DEF(job)
 	JobDebug("AR has failed, Player: [player], Mind: [player?.mind], Rank: [rank]")
 	return FALSE
 
+/proc/handle_squad(mob/M, rank, latejoin = FALSE)
+	var/strict = FALSE
+	var/datum/squad/P = SSjob.squads[M.client.prefs.preferred_squad]
+	var/datum/squad/R = SSjob.squads[pick(SSjob.squads)]
```
Rohesie: We might want to add further checks for how squads are picked, but this is a good base.

code/controllers/stonedmc/subsystem/job.dm
DominikPanic: That's my oldcode, I don't trust your ?. in procs

code/controllers/stonedmc/subsystem/job.dm
DominikPanic: Yeah it's not very modular since it has hardcoded roles, just something I hacked together to make the job subsystem work plus preserve the "prefer squad over role" idea.

code/controllers/stonedmc/subsystem/job.dm
Rohesie: What do you mean you don't trust it? Have you tested it and see it fail?

PR #1254, opened 2019-04-08
Moves frenzy damage bonus to a define

spookydonut:
## About The Pull Request
Stomp was the only thing this was different for.

## Why It's Good For The Game
Consistency also a define for easy mass change

## Changelog
:cl:
balance: frenzy bonus for stomp now adds 2*frenzy level instead of +10% per frenzy level to match all other frenzy bonuses
/:cl:


KitKetKetrai:
You should just expand on frenzy entirely. There's a load of abilities that aren't affected by frenzy. See fling, see punch. Admittedly, those just deal halloss or damage and halloss. But halloss damage should also be increased. At most in practice you'll get a hivelord emitting frenzy. Which is at strenght level three when ancient. So six extra damage/halloss.

PR #1256, opened 2019-04-08
Refactors hivelord thiccening

spookydonut:
split from #1201


code/modules/mob/living/carbon/xenomorph/powers.dm
```diff
@@ -497,15 +497,11 @@ GLOBAL_LIST_INIT(acid_spray_hit, typecacheof(list(/obj/structure/barricade, /obj
 				if(WR.walltype == "resin")
 					visible_message("<span class='xenonotice'>\The [src] regurgitates a thick substance and thickens [WR].</span>", \
 					"<span class='xenonotice'>You regurgitate some resin and thicken [WR].</span>", null, 5)
-					var/prev_oldturf = WR.oldTurf
-					WR.ChangeTurf(/turf/closed/wall/resin/thick)
-					WR.oldTurf = prev_oldturf
+					WR.thicken()
```
Rohesie: Maybe do something like this to prevent spending plasma on wrong targets?
```
if(!WR.thicken())
	to_chat(src, "<span class='xenowarning'>[DR] can't be made thicker.</span>")
	return
```

spookydonut: see #1201 this is just atomising the addition of the thicken proc off.

PR #1258, opened 2019-04-09
Moves squad color overlays to GLOBs

spookydonut:
## About The Pull Request
Code quality improvement

## Why It's Good For The Game
instead of armor overlays being made in world/New they are now done in SSjob init and squad colors no longer need a global list as the color values are in hex already.
Expect more stuff like this as I work on factions

### DIFF
--- a/code/game/jobs/squads.dm
+++ b/code/game/jobs/squads.dm
@@ -78,9 +78,9 @@
 	radio_freq = DELTA_FREQ
 
 GLOBAL_LIST_EMPTY(armormarkings)
-GLOBAL_LIST_EMPTY(armormarkings_sql)
+GLOBAL_LIST_EMPTY(armormarkings_sl)
 GLOBAL_LIST_EMPTY(helmetmarkings)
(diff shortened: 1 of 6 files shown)


code/game/jobs/squads.dm
```diff
@@ -72,11 +72,31 @@
...
+GLOBAL_LIST_EMPTY(helmetmarkings_sql)
+
+/datum/squad/New()
+	. = ..()
+	var/image/armor = image('icons/mob/suit_1.dmi',icon_state = "std-armor")
+	var/image/armorsl = image('icons/mob/suit_1.dmi',icon_state = "sql-armor")
+	armor.color = color
+	armorsl.color = color
+	GLOB.armormarkings[type] = armor
+	GLOB.armormarkings_sql[type] = armorsl
```
Rohesie: Could you use `armormarkings_sl` instead of `armormarkings_sql` ? Squad leader is not the first thing that comes to mind by SQL.

PR #1259, opened 2019-04-09
Minor cleanup of overwatch code

spookydonut:
split from #1253

### DIFF
--- a/code/game/jobs/squads.dm
+++ b/code/game/jobs/squads.dm
@@ -78,6 +78,12 @@
 	radio_freq = DELTA_FREQ
 
 
+/datum/squad/proc/get_all_members()
+	return marines_list
+
+/datum/squad/proc/get_total_members()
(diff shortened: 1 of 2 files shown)


code/game/jobs/squads.dm
```diff
@@ -78,6 +78,12 @@
 	radio_freq = DELTA_FREQ
 
 
+/datum/squad/proc/get_all_members()
+	return marines_list
+
+/datum/squad/proc/get_total_members()
```
Rohesie: How about turning these into macros, since they are not redefined anywhere?
`#define SQUAD_TOTAL_MEMBERS(squad) length(squad.marines_list)`
``#define SQUAD_ALL_MEMBERS(squad) squad.marines_list``

spookydonut: see #1253 for why not

PR #1263, opened 2019-04-09
Band-aiding this heap of sadness that's facehuggers.

Ghommie:
## About The Pull Request
Applying some rapid tweaks and adjustments to reduce facehuggers inconsistency and RNG. I'd call this band-aid because I now believe they need a subsystem to handle this at a more appropriate tick rate than once every 2 seconds, but can't do right away.

## Why It's Good For The Game
Reducing facehuggers inconsistency and RNG a little. Super optimal outcome not granted.

## Changelog
:cl:
tweak: Tweaking facehuggers once more in preparation of a possibly more appropriate subsystem for them.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
@@ -4,8 +4,8 @@
 #define MIN_IMPREGNATION_TIME 10 SECONDS //Time it takes to impregnate someone
 #define MAX_IMPREGNATION_TIME 15 SECONDS
 
-#define MIN_ACTIVE_TIME 3 SECONDS //Time between being dropped and going idle
-#define MAX_ACTIVE_TIME 5 SECONDS
+#define MIN_ACTIVE_TIME 4 SECONDS //Time between being dropped and going idle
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/facehuggers.dm
```diff
@@ -234,27 +235,25 @@
 	update_icon()
 
 /obj/item/clothing/mask/facehugger/throw_impact(atom/hit_atom, speed)
-	if(stat == DEAD)
+	if(stat != CONSCIOUS)
```
spookydonut:
```suggestion
	if(stat)
```

DominikPanic: While technically true it's MUCH less readable.

Ghommie: ultimately, this is why the CONSCIOUS define even exists.

Rohesie: I prefer `if(stat)` myself, but this is not something we should be enforcing. `if(stat != CONSCIOUS)` is fine.

PR #1268, opened 2019-04-10
Spitter plasma rebalance

KitKetKetrai:
## About The Pull Request

- Buffs heavy spit plasma cost from 100 to 75
- nerfs spitter plasma from 800 - 1150 to 650 - 925
- nerfs spitter plasma regen from 25 - 35 to 21 - 30

## Why It's Good For The Game

This makes it so the spitter doesn't have an obscene amount of plasma for seemingly no reason. This is mainly aimed at making drones and hivelords more useful as a support role. Currently at higher maturity levels spitter has so much plasma that drones and even hivelords struggle to fill them back up. They also had more plasma than praetorians, which makes little sense. This makes it so praetorians aren't necessarily worse at spitting than spitters, as their plasma remains the same.

I didn't nerf regen and plasma by 25% (about 15-20% instead) due to the existence of a 300 plasma acid spray. This ability is on a long cooldown, and shouldn't just be spammed like spit. I also felt like 30-ish seconds to recover your plasma as a spitter made gameplay a bit too slow unless you had recovery pheromones.

## Changelog
:cl:
balance:  Acid spit is now cheaper. Praetorians can spit acid longer, whilst spitters get lower plasma and regen to compensate. This should make it easier for drones and hivelords to keep spitters topped off.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/spitter/castedatum_spitter.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/spitter/castedatum_spitter.dm
@@ -19,8 +19,8 @@
 	speed = -0.5
 
 	// *** Plasma *** //
-	plasma_max = 800
-	plasma_gain = 25
+	plasma_max = 600
+	plasma_gain = 20
 
 	// *** Health *** //
 	max_health = 180
@@ -66,8 +66,8 @@
 	speed = -0.6
 
 	// *** Plasma *** //
-	plasma_max = 1000
-	plasma_gain = 30
+	plasma_max = 750
+	plasma_gain = 24
 
 	// *** Health *** //
 	max_health = 220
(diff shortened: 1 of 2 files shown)


Rohesie:
Will have to be tested.

KitKetKetrai:
If this ends up being too much free spit for praetorian, the easiest fix I can think of is making there be two different spits again. Identical damage, just slightly cheaper for spitter vs praetorian.

Dunno what is so salt induced about this? IF you think I'm salty about spitters being too good I'm definitely not. This is practically a neutral update for spitters. Making them slighlty quicker at recovering their plasma, but worse at spraying.

It's not that big of ab uff to anything but praetorian. But who uses spit as praetorian? It's useful, but most of the time you want to be using neurotoxin, which wasn't affected.

PR #1271, opened 2019-04-11
Thin Sticky Rebalance

InterroLouis:
## About The Pull Request

Lowers thin sticky health by 1 point. The result of this is that it takes 1 combat knife hit to clear it instead of 2. 
## Why It's Good For The Game

Makes it easier to clear thin sticky, which is considerably more easily spammed than it used to be. 

## Changelog
:cl:
balance: Thin Sticky Resin clears in one knife hit
/:cl:


KitKetKetrai:
severe nerf to thin resin because knives can attack a lot faster. But we'll see how it plays out. At least with a bayonet/machete you have to made a real tradeof and/or give up storage space.

PR #1279, opened 2019-04-12
[Tested]Second batch of /attackby( changes

DominikPanic:
copypasta:
This will be a long series as there's 330 files where attackby is used which will mean ~8 PRs in total

The goal is for every attackby to call parent where possible for proper fingerprinting and click delays.

I'm also cleaning up the code SLIGHTLY, changes welcome as long as they're reasonable.


spookydonut:
Mostly invalidated by #1324

PR #1282, opened 2019-04-13
Fixes click code passing parameters

DominikPanic:


spookydonut:
Uh why? If we're headed towards tg style click code then things should be using the like ctrlclick procs for those behaviours

DominikPanic:
> 
> 
> Uh why? If we're headed towards tg style click code then things should be using the like ctrlclick procs for those behaviours

If you read the actual comment in /tg/code it was supposed to pass those parameters in the first place. This is a bugfix, not a change in behavior.

PR #1286, opened 2019-04-14
Var cleanup.

Rohesie:
## About The Pull Request

Big clean-up from `mob`, `living`, `carbon` and `human` defines.
Moved some variables around to match their location in /tg/, intended ones as was their description, or where they were effectively used.
Removed several unused variables and useless initializations (unused empty lists for example).
Had to tweak procs and whatnot to make that happen.

Tested it locally, but by the size of this PR this is bound to unearth before-concealed issues in the form of runtimes. Will have to be tested live to gather those.

Things to look for:
* Merged the `living` var `tod` with the `mob` var `timeofdeath`. @DominikPanic will have to check the change to the sql schema and tell if it should be modified.
* Removed references of `mutations`, `dna` and other stuff that are in for removal (just not doing it fully in this PR to keep it atomized), and moved the legacy disease code (also in for removal), but those may end up causing runtimes due to currently-useless checks.
* Fixed `is_ventcrawling` that was never set. The variable now actually serves the intended purpose. A quick test showed no issues, but this is not my feature and time will tell if its current state is fully functional.


code/datums/disease.dm
```diff
@@ -205,3 +205,157 @@ var/list/diseases = subtypesof(/datum/disease)
...
+		//to_chat(world, "It's an advance virus.")
+		var/datum/disease/advance/A = virus
+		if(A.GetDiseaseID() in resistances)
+			//to_chat(world, "It resisted us!")
+			return
+		if(count_by_type(viruses, /datum/disease/advance) >= 3)
+			return
+
+	else
+		if(src.resistances.Find(virus.type))
```
DominikPanic:
```suggestion
		if(resistances.Find(virus.type))
```

SQL/tgmc-schema.sql
```diff
@@ -100,7 +100,7 @@ CREATE TABLE IF NOT EXISTS `death` (
   `server_ip` int(10) unsigned NOT NULL,
   `server_port` smallint(5) unsigned NOT NULL,
   `round_id` int(11) NOT NULL,
-  `tod` datetime NOT NULL COMMENT 'Time of death',
+  `timeofdeath` datetime NOT NULL COMMENT 'Time of death',
```
DominikPanic: Revert this. This is just the name of the table in the database, not affected by any in-game variables.

code/datums/disease.dm
```diff
@@ -205,3 +205,157 @@ var/list/diseases = subtypesof(/datum/disease)
...
+			//to_chat(world, "Normal virus and resisted")
+			return
+
+	if(has_disease(virus))
+		return
+
+	if(force_species_check)
+		var/fail = 1
```
DominikPanic:
```suggestion
		var/fail = TRUE
```

code/datums/disease.dm
```diff
@@ -205,3 +205,157 @@ var/list/diseases = subtypesof(/datum/disease)
...
+		if(ishuman(src))
+			var/mob/living/carbon/human/H = src
+			for(var/vuln_species in virus.affected_species)
+				if(H.species.name == vuln_species)
+					fail = 0
```
DominikPanic:
```suggestion
					fail = FALSE
```

code/datums/disease.dm
```diff
@@ -205,3 +205,157 @@ var/list/diseases = subtypesof(/datum/disease)
...
+					break
+		if(fail) return
```
DominikPanic: Single line

code/modules/mob/mob_helpers.dm
```diff
@@ -271,15 +271,22 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
 	return FALSE
 
 
-/mob/proc/abiotic(var/full_body = 0)
-	if(full_body && ((src.l_hand && !( src.l_hand.flags_item & ITEM_ABSTRACT )) || (src.r_hand && !( src.r_hand.flags_item & ITEM_ABSTRACT )) || (src.back || src.wear_mask)))
+/mob/proc/abiotic(full_body)
+	if(full_body && ((src.l_hand && !( src.l_hand.flags_item & ITEM_ABSTRACT )) || (src.r_hand && !( src.r_hand.flags_item & ITEM_ABSTRACT ))))
```
DominikPanic:
```suggestion
	if(full_body && ((l_hand && !(l_hand.flags_item & ITEM_ABSTRACT )) || (r_hand && !(r_hand.flags_item & ITEM_ABSTRACT ))))
```

code/datums/disease.dm
```diff
@@ -205,3 +205,157 @@ var/list/diseases = subtypesof(/datum/disease)
...
+
+	if(skip_this == 1)
```
DominikPanic: A lot of lines below that also need cleanup

Rohesie: This is all stuff to be removed. Only not done on this PR not to bloat it. Merely moved it.

code/__HELPERS/unsorted.dm
```diff
@@ -281,11 +281,17 @@ Turf and target are seperate in case you want to teleport some distance from a t
 		mind.name = newname
 		if(mind.key)
 			log_played_names(mind.key, newname) //Just in case the mind is unsynced at the moment.
-	if(dna)
-		dna.real_name = real_name
 
 	return TRUE
 
+/mob/living/carbon/human/fully_replace_character_name(oldname, newname)
```
spookydonut: duplicate proc, see below

code/controllers/stonedmc/subsystem/job.dm
```diff
@@ -189,16 +189,16 @@ SUBSYSTEM_DEF(job)
 //Gives the player the stuff he should have with his rank
 /datum/controller/subsystem/job/proc/EquipRank(mob/M, rank, joined_late = FALSE)
 	var/mob/new_player/N
-	var/mob/living/H
```
spookydonut: just why

code/datums/disease.dm
```diff
@@ -205,3 +205,158 @@ var/list/diseases = subtypesof(/datum/disease)
 	active_diseases -= src
 	. = ..()
 	return TA_PURGE_ME_NOW //immediate deletion
+
+
+/mob/living/carbon/proc/has_disease(var/datum/disease/virus)
```
spookydonut: where is this from

code/datums/diseases/advance/symptoms/weight.dm
```diff
@@ -27,7 +27,7 @@ Bonus
 /datum/symptom/weight_gain/Activate(var/datum/disease/advance/A)
 	..()
 	if(prob(SYMPTOM_ACTIVATION_PROB))
-		var/mob/living/M = A.affected_mob
```
spookydonut: you dont need to change this

spookydonut: you could even just directly reference A.affected_mob since thats already carbon

code/datums/disease.dm
Rohesie: `code/datums/diseases/mob_procs.dm`

code/datums/diseases/advance/symptoms/weight.dm
Rohesie: This code is in for removal. No need to waste time on improving it.

code/controllers/stonedmc/subsystem/job.dm
Rohesie: Because later into the file `H` is defined for human.

code/game/dna/genes/monkey.dm
```diff
@@ -7,7 +7,7 @@
 /datum/dna/gene/monkey/can_activate(var/mob/M,var/flags)
 	return ishuman(M) || istype(M,/mob/living/carbon/monkey)
 
-/datum/dna/gene/monkey/activate(var/mob/living/M, var/connected, var/flags)
+/datum/dna/gene/monkey/activate(mob/living/carbon/human/M, connected, flags)
```
spookydonut: in future dont change stuff like this unless you're doing the entire file or changing the proc.

Rohesie: Alright.

PR #1293, opened 2019-04-16
away_timer away from Life()

Rohesie:
## About The Pull Request

Turned `mob`'s `away_timer` `life()` tick accumulator into a variable that stores the time when the mob lost its client.

Tested this locally as much as I could, but I cannot really disconnect, so this will require a test with at least someone else.

Admin's Take Over verb (`direct_control()`) fails to reset the `away_timer` of the mob it leaves behind, but then again that's not the only issue of that proc so it would get a bit out of scope fixing it here.

* Made xeno respawn use the global respawn time variable. This makes it easier for admins to tweak it and testing, but will tie it to other kinds of respawns. Can make another var for it if needed.

## Why It's Good For The Game

Long-term goal is to make `Life()` only deal with living or revibable peeps. One step at a time.


spookydonut:
aghosting then taking over another mob or ckeying into it won't set the away timer

DominikPanic:
> 
> 
> aghosting then taking over another mob or ckeying into it won't set the away timer

To address this, make the transfer_mind proc handle it

Rohesie:
This is tested and works. The reference was wrong regarding `mob.Logout()` having a `client` still attached on disconnections, so had to processed it through `client.Del()` instead.

Check on the xeno `respawn` changes and see if you agree with them. They are pretty simple and shouldn't come into play unless with admemery.

PR #1297, opened 2019-04-17
Make life live less

spookydonut:
## About The Pull Request
Reduce amount of mob processing load

### DIFF
--- a/code/controllers/stonedmc/subsystem/mobs.dm
+++ b/code/controllers/stonedmc/subsystem/mobs.dm
@@ -17,6 +17,7 @@ SUBSYSTEM_DEF(mobs)
 /datum/controller/subsystem/mobs/proc/stop_processing(mob/M)
 	if(!M.processing)
 		return
+	M.ExtinguishMob()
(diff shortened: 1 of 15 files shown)


code/controllers/stonedmc/subsystem/mobs.dm
```diff
@@ -25,17 +40,17 @@ SUBSYSTEM_DEF(mobs)
...
-		var/mob/M = currentrun[currentrun.len]
+		var/mob/living/L = currentrun[currentrun.len]
 		currentrun.len--
-		if(M)
-			M.Life(seconds, times_fired)
+		if(L)
+			L.Life(seconds, times_fired)
 		else
-			GLOB.mob_list.Remove(M)
+			processing.Remove(L)
```
DominikPanic:
```suggestion
			LAZYREMOVE(processing, L)
```

DominikPanic: Eh, this is so minor and unclear it's not worth it.

PR #1298, opened 2019-04-17
Multitiles poddoors + doors code cleanup

Menshin:
## About The Pull Request

* Made poddoors use inherited procs instead of rewriting them when possible
* made the door base class correctly handles 4-tiles doors regarding opacity and boundaries box
* multitile opacity is now handled by an `opacifier` object instead of meddling with turfs opacity directly
* extended `UNACIDABLE` and `INDESTRUCTIBLE` flags to all multitiles poddoors (instead of only the 4 tiles horizontal one)
* minor code cleaning (`locked`, `welded` and `not_weldable` are now attributes of the door base class, lots of defines replacing)


code/game/machinery/doors/poddoor.dm
```diff
@@ -1,381 +1,73 @@
-
 /obj/machinery/door/poddoor
 	name = "\improper Podlock"
 	desc = "That looks like it doesn't open easily."
 	icon = 'icons/obj/doors/rapid_pdoor.dmi'
 	icon_state = "pdoor1"
-	id = 1.0
-	dir = 1
+	id = ""
```
spookydonut: null?

code/game/machinery/doors/poddoor.dm
```diff
@@ -1,381 +1,73 @@
...
+	dir = SOUTH
```
spookydonut: dont all objects have a default dir?

Zostroll:
I spent so long on these doors and you went and did it all better than me :(

code/game/machinery/doors/poddoor.dm
```diff
@@ -1,381 +1,71 @@
...
 		spawn(0)
-			operating = 1
-			flick("pdoorc0", src)
-			icon_state = "pdoor0"
-			SetOpacity(0)
+			operating = TRUE
 			sleep(15)
-			density = 0
-			operating = 0
+			open()
```
Rohesie: Any change of a `CALLBACK` here, to remove those `spawn` and `sleep` procs?

Menshin: Sure.

PR #1305, opened 2019-04-19
Purges zombie code

DominikPanic:
It had to be done


Ghommie:
Good riddance. I gave this a look a month or two ago ~~after breaking up some things~~, It's badly done both by design and code, so just better to port some at a later date.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -17,8 +17,7 @@
 	life_tick++
 
 	voice = GetVoice()
-	if(stat == DEAD && iszombie(src) && regenZ)
-		handle_organs()
+	if(stat == DEAD)
```
Rohesie: This is conflict with Spooky's life PR. Keep an eye on it.

DominikPanic: Also it will definitely conflict with all of my attackby PRs

DominikPanic: And done.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -18,11 +18,9 @@
 
 	voice = GetVoice()
 	if(stat == DEAD)
-		if(iszombie(src) && regenZ)
-			handle_organs()
-			return
 		if(!check_tod())
 			SSmobs.stop_processing(src)
+		return
```
spookydonut:
```suggestion

```

PR #1310, opened 2019-04-20
Made the xeno queen locator more consistent

Ghommie:
## About The Pull Request
Made the `queen_locator()` activate once every four ticks (and if conscious) instead of a `prob(25)`.
Also moved life_tick++ down to carbon `life()` for this to work.

## Why It's Good For The Game
More consistent updates without the RNG.

## Changelog
:cl:
tweak: Made the xeno queen locator more consistent, triggering once every 4 life cycles rather than a 25% probability each cycle.
/:cl:


Ghommie:
> 
> 
> should be moved to SSdirection

My thougths exactly, but I got hands full on tasks and can't do that right now.

Rohesie:
This will get deprecated by #1337

Ghommie:
> 
> 
> This will get deprecated by #1337

Good, this is closed then.

PR #1315, opened 2019-04-20
Fixes possible runtimes for people loading in too fast

DominikPanic:


Rohesie:
Could you share the runtime?

DominikPanic:
> 
> 
> Could you share the runtime?

It's mob/Stat null.stat_netry

spookydonut:
I've never seen this runtime, how do you recreate it?

DominikPanic:
> 
> 
> I've never seen this runtime, how do you recreate it?

Restart server with ~40 players

code/modules/mob/mob.dm
```diff
@@ -32,12 +32,12 @@
 				stat("CPU:", "[world.cpu]")
 				stat("Instances:", "[num2text(length(world.contents), 10)]")
 				stat("World Time:", "[world.time]")
-				GLOB.stat_entry()
-				config.stat_entry()
-				supply_controller.stat_entry()
-				shuttle_controller.stat_entry()
-				lighting_controller.stat_entry()
-				SSradio.stat_entry()
+				GLOB?.stat_entry()
```
Rohesie: stack_trace time to determine which is runtiming.
```suggestion
				if(GLOB)
					GLOB.stat_entry()
				else
					stack_trace("Null GLOB in MC panel")
```
And so on.

DominikPanic: That does not make any sense to do here. This is only present by people loading in too fast, and a very rare runtime at that.

Rohesie: Only for testing purposes until the runtime is located. Not for merging.

DominikPanic: There is nothing to test. This is objectivelly a bad solution. The problem is that one of those systems (I know which but too lazy to look up the specific runtime) is initialized or rather non-existant when people load in super-early. Stack_tracing won't make it better, the proper solution is to either do it like this or add some arbitrary check to prevent stat from updating until a certain init point but that would make the user experience terrible.

Rohesie: If you know which is then add the `?.` to it instead of to them all. That's the purpose of determining which is failing.
A check to only start updating `stat` when it's due time would be ideal, yes, but might not be worth the effort. The user experience would be unaffected here, really.

DominikPanic: Optimized (it was the shuttle one)

PR #1316, opened 2019-04-20
Fixes stack recipes on most floor turfs

DominikPanic:
@Rohesie did the big bad


DominikPanic:
> 
> 
> this will let you build floors on dirt

Still better than marines not being able to build barricades on the vast majority of maps.

Rohesie:
Good work as a patch. Let me get a more maintainable long-term solution.

DominikPanic:
> 
> 
> Good work as a patch. Let me get a more maintainable long-term solution.

Aye, not demanding a merge

PR #1317, opened 2019-04-20
Adds fuzzy cuffs 2.0, Electric Boogaloo

DominikPanic:
Arachnid died for this.

Also, Alterist didn't set it as invalid so the PR being closed was definitely not okay.
@Rohesie also said they don't mind this PR.

dreamseeker_2019-04-11_20-35-49

:cl: LaKiller8 & ArachnidNexus
imageadd: Adds fuzzy cuffs. Yes, that kind.
/:cl:


Pariah919:
why do we need these

DominikPanic:
Arachnid died for this

PR #1327, opened 2019-04-21
Simplify generating xeno hud icons

psykzz:
## About The Pull Request

Refactors xeno hud to simplify buckets using maths insteading of manual assignment.

I did read that switchs are faster than maths but the scale is on ten of thousands of calls. The benefit here is the calculation is shared instead of multiple implementations.

## Changelog
:cl:
tweak: Xeno huds are now rounded up.
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1510,7 +1510,7 @@ proc/pick_closest_path(value, list/matches = get_fancy_list_of_atom_types())
 
 // Bucket a value within boundary
 /proc/get_bucket(var/bucket_size, var/max, var/current, var/min = 0, var/list/boundary_terms = list())
-	if (boundary_terms.len == 2)
+	if (length(boundary_terms) == 2)
 		if (current >= max) 
 			return boundary_terms[1]
 		if (current < min)
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1509,7 +1509,7 @@ proc/pick_closest_path(value, list/matches = get_fancy_list_of_atom_types())
 	return "[round((powerused * 0.000000001),0.0001)] GW"
 
 // Bucket a value within boundary
-/proc/get_bucket(var/bucket_size, var/max, var/current, var/min = 0, var/list/boundary_terms = list())
+/proc/get_bucket(bucket_size, max, current, min = 0, list/boundary_terms)
 	if (length(boundary_terms) == 2)
 		if (current >= max) 
 			return boundary_terms[1]
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1507,3 +1507,13 @@ proc/pick_closest_path(value, list/matches = get_fancy_list_of_atom_types())
 	else if(powerused < 1000000000) //Less than a GW
 		return "[round((powerused * 0.000001),0.001)] MW"
 	return "[round((powerused * 0.000000001),0.0001)] GW"
+
+// Bucket a value within boundary
+/proc/get_bucket(var/bucket_size, var/max, var/current, var/min = 0, var/list/boundary_terms = list())
+	if (boundary_terms.len == 2)
+		if (current >= max) 
+			return boundary_terms[1]
+		if (current < min)
+			return boundary_terms[2]
+
+	return CEILING((bucket_size / max) * current, 1)
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -229,7 +229,6 @@
 		if(stat != DEAD)
 			if (xeno_caste.plasma_max)
 				var/bucket = get_bucket(XENO_HUD_ICON_BUCKETS, xeno_caste.plasma_max, plasma_stored, 0, list("full", "empty"))
-				message_admins("Bucket size [bucket]")
 				hud_used.alien_plasma_display.icon_state = "power_display_[bucket]"
 		else
 			hud_used.alien_plasma_display.icon_state = "power_display_empty"
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -3,6 +3,9 @@
 #define XENO_STANDING_HEAL 0.2
 #define XENO_CRIT_DAMAGE 5
 
+#define XENO_HUD_ICON_BUCKETS 17  // +1 to the number of icons you to represent health / plasma
+
+
 /mob/living/carbon/Xenomorph/Life()
 
 	if(monkeyizing || !loc)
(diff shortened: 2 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -216,100 +219,25 @@
...
-		if(interactee)
-			interactee.check_eye(src)
-		else
-			if(client && !client.adminobs)
-				reset_view(null)
+	if(interactee)
+		interactee.check_eye(src)
+	else
+		if(!client.adminobs)
+			reset_view(null)
```
psykzz: Added this in too, because I dont see why it was attach to the fact of if you have a plasma hud

code/__HELPERS/unsorted.dm
```diff
@@ -1507,3 +1507,13 @@ proc/pick_closest_path(value, list/matches = get_fancy_list_of_atom_types())
 	else if(powerused < 1000000000) //Less than a GW
 		return "[round((powerused * 0.000001),0.001)] MW"
 	return "[round((powerused * 0.000000001),0.0001)] GW"
+
+// Bucket a value within boundary
+/proc/get_bucket(var/bucket_size, var/max, var/current, var/min = 0, var/list/boundary_terms = list())
+	if (boundary_terms.len == 2)
```
DominikPanic:
```suggestion
	if (length(boundary_terms) == 2)
```

code/__HELPERS/unsorted.dm
```diff
@@ -1507,3 +1507,13 @@ proc/pick_closest_path(value, list/matches = get_fancy_list_of_atom_types())
...
+		if (current >= max) 
+			return boundary_terms[1]
+		if (current < min)
+			return boundary_terms[2]
+
+	return CEILING((bucket_size / max) * current, 1)
```
Menshin: This isn't equivalent to the old switch, for example if a mob has 90 health out of 100 maxhealth :
* the switch gives `hud_used.healths.icon_state = "health_15"`
* this formula gives `CEILING((17/ 100) * 90, 1) = CEILING(15,3, 1) = 16`, so it uses `"health_16"`. 

Also it can uses `"health_17"` which (i assume) doesn't exist.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -3,6 +3,9 @@
 #define XENO_STANDING_HEAL 0.2
 #define XENO_CRIT_DAMAGE 5
 
+#define XENO_HUD_ICON_BUCKETS 17  // +1 to the number of icons you to represent health / plasma
```
Menshin:
```suggestion
#define XENO_HUD_ICON_BUCKETS 17  // +1 to the number of icons you (use ?) to represent health / plasma
```

code/__HELPERS/unsorted.dm
psykzz: I  reduced down the buckets from 17 to 16, as you're right i didn't meant to go over 16.

The fact that the switch and proc return different results, Im fine with personally.

code/modules/mob/living/carbon/xenomorph/life.dm
Ghommie: That may have been my mistake when I reorganized some ofthese procs to /carbon last year. Good fix.

code/__HELPERS/unsorted.dm
Menshin: > The fact that the switch and proc return different results, Im fine with personally.

Then it should be clear in the changelog that these changed (even slightly), as players may be used to the old thresholds.

code/__HELPERS/unsorted.dm
```diff
@@ -1507,3 +1507,13 @@ proc/pick_closest_path(value, list/matches = get_fancy_list_of_atom_types())
 	else if(powerused < 1000000000) //Less than a GW
 		return "[round((powerused * 0.000001),0.001)] MW"
 	return "[round((powerused * 0.000000001),0.0001)] GW"
+
+// Bucket a value within boundary
+/proc/get_bucket(var/bucket_size, var/max, var/current, var/min = 0, var/list/boundary_terms = list())
```
Ghommie: Feel free to remove the `var/` prefixes, they are already implied on arguments, also remove the new `list()` from `boundary_terms`, it's not used by default.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -216,100 +218,25 @@
...
-					hud_used.healths.icon_state = "health_3"
-				if(10 to 15)
-					hud_used.healths.icon_state = "health_2"
-				if(4 to 9)
-					hud_used.healths.icon_state = "health_1"
-				if(0 to 3)
-					hud_used.healths.icon_state = "health_0"
-				else
-					hud_used.healths.icon_state = "health_critical"
+			if(maxHealth)
```
Rohesie: If you want to do a check to avoid a division by zero, do it so it doesn't silently errors, as no mob with health should have `maxHealth`
Something like this would work:
```
if(!maxHealth)
	stack_trace("[src] called handle_regular_hud_updates() while having [maxHealth] maxHealth.")
	return
```
Will require to fix indentation.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -216,100 +218,25 @@
...
-						hud_used.alien_plasma_display.icon_state = "power_display_2"
-					if(4 to 9)
-						hud_used.alien_plasma_display.icon_state = "power_display_1"
-					if(0 to 3)
-						hud_used.alien_plasma_display.icon_state = "power_display_0"
-					else
-						hud_used.alien_plasma_display.icon_state = "power_display_empty"
-			else
-				hud_used.alien_plasma_display.icon_state = "power_display_empty"
+			if (xeno_caste.plasma_max)
```
Rohesie: Same here, this is unfortunate inheritance.

code/modules/mob/living/carbon/xenomorph/life.dm
Rohesie: Are you sure every mob processing `handle_regular_hud_updates()` not interacting with an object will have a client attached?

psykzz:
@DominikPanic is this good for you now?

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -214,102 +216,36 @@
...
 			hud_used.alien_plasma_display.icon_state = "power_display_empty"
 
-		if(interactee)
-			interactee.check_eye(src)
-		else
-			if(client && !client.adminobs)
-				reset_view(null)
+	if(interactee)
+		interactee.check_eye(src)
+	else
```
DominikPanic: combine the else and if

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -214,102 +216,36 @@
 /mob/living/carbon/Xenomorph/handle_regular_hud_updates()
 	if(!client)
 		return FALSE
+
+	// Sanity checks
+	if(!maxHealth)
+		stack_trace("[src] called handle_regular_hud_updates() while having [maxHealth] maxHealth.")
+		return
+	if(!xeno_caste.plasma_max)
```
DominikPanic: This will only handle 0 and null, will also runtime if xeno_caste is null though that is not an issue as it would be a sign of something much worse happening

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -214,102 +216,36 @@
 /mob/living/carbon/Xenomorph/handle_regular_hud_updates()
 	if(!client)
 		return FALSE
+
+	// Sanity checks
+	if(!maxHealth)
```
DominikPanic: Same, will also only handle 0 and null

psykzz:
@DominikPanic regarding the talk we had on discord, should these `stack_trace` checks just go away, or should i continue check there?

Rohesie:
I'll leave La to answer on if the whole package should go or stay, but if you add sanity checks, do make them informative, as with `stack_trace()`.
My opinion is that it's not really necessary here.

DominikPanic:
*shrug, use your best judgement

psykzz:
Lets leave them for now, we can remove it, if they get noisy or break shit. If it works and helps, then maybe we can add them elsewhere.

PR #1328, opened 2019-04-21
Refactor screech into screech_act

psykzz:
## About The Pull Request

Small refactor for screech into screech_act

## Changelog

:cl:
tweak: Queen's screech is no longer blocked by LoS.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
```diff
@@ -154,28 +154,8 @@
...
-		if(dist < 8)
-			to_chat(H, "<span class='danger'>An ear-splitting guttural roar tears through your mind and makes your world convulse!</span>")
-			H.stunned += stun_duration
-			H.KnockDown(stun_duration)
-			H.apply_damage(halloss_damage, HALLOSS)
-			if(!H.ear_deaf)
-				H.ear_deaf += stun_duration * 20  //Deafens them temporarily
-			//Perception distorting effects of the psychic scream
-			addtimer(CALLBACK(GLOBAL_PROC, /proc/shake_camera, H, stun_duration * 10, 0.75), 31)
+	for(var/mob/M in oview(7, src))
```
Rohesie:
```suggestion
	for(var/mob/M in range(world.view))
```
`oview()` would make people under the queen to be unaffected.
Using view or hearing procs is a bit quirky, as it causes stuff like smoke to block the screech.
This is in the reference: `oview(Dist,Center=usr)`
It means you don't need to set `src` as it will be by default the user.

psykzz: This was mostly copied due to the fact the original code used it that way, looking into it i did see this was the case, but didn't want to make any assumptions so left it as it was. 
Happy to use `range(world.view)` though

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -258,4 +258,8 @@
 	return
 
 /mob/living/carbon/Xenomorph/reagent_check(datum/reagent/R) //For the time being they can't metabolize chemicals.
-	return TRUE
\ No newline at end of file
+	return TRUE
+
+
+/mob/living/carbon/Xenomorph/screech_act(mob/Q)
```
Rohesie:
```suggestion
/mob/living/carbon/Xenomorph/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
```

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -258,4 +258,8 @@
...
+	shake_camera(src, 10, 1)
```
Rohesie:
```suggestion
	shake_camera(src, 1 SECONDS, 1)
```
For clarity.

code/modules/mob/mob.dm
```diff
@@ -853,4 +853,8 @@ mob/proc/yank_out_object()
 		else
 			to_chat(usr, "<span class='notice'>You just did an audible emote. Wait a while.</span>")
 			return TRUE
-	return FALSE
\ No newline at end of file
+	return FALSE
+
+// When they are affected by a queens screech
+/mob/proc/screech_act(mob/Q)
```
Rohesie:
```suggestion
/mob/living/proc/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
```
This should be handled ideally at the `living` level, and not even consider ghosts.

code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
```diff
@@ -154,28 +154,8 @@
...
+		M.screech_act(src)
```
Rohesie:
```suggestion
		L.screech_act(src)
```
So ghosts are not processed without need.

code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
```diff
@@ -154,28 +154,8 @@
...
-		if(dist < 8)
-			to_chat(H, "<span class='danger'>An ear-splitting guttural roar tears through your mind and makes your world convulse!</span>")
-			H.stunned += stun_duration
-			H.KnockDown(stun_duration)
-			H.apply_damage(halloss_damage, HALLOSS)
-			if(!H.ear_deaf)
-				H.ear_deaf += stun_duration * 20  //Deafens them temporarily
-			//Perception distorting effects of the psychic scream
-			addtimer(CALLBACK(GLOBAL_PROC, /proc/shake_camera, H, stun_duration * 10, 0.75), 31)
+	for(var/mob/M in range(world.view))
```
Rohesie:
```suggestion
	for(var/mob/living/L in range(world.view))
```
Only living, no ghosts.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1750,4 +1750,21 @@
...
+	return TRUE
+
+/mob/living/carbon/human/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
+	shake_camera(src, 3 SECONDS, 1) //50 deciseconds, SORRY 5 seconds was way too long. 3 seconds now
+	var/dist = get_dist(src, Q)
+	var/reduction = max(1 - 0.1 * protection_aura, 0) //Hold orders will reduce the Halloss; 10% per rank.
+	var/halloss_damage = (max(0,140 - dist * 10)) * reduction //Max 130 beside Queen, 70 at the edge
+	var/stun_duration = max(0,1.1 - dist * 0.1) * reduction //Max 1 beside Queen, 0.4 at the edge.
+
+	if(dist < 8)
```
Rohesie: Do an early return here.
Basically if the contrary, then return.
Such as
```
if(dist > 7)
	return
```
Then you'll have to adjust the indentation of the rest.

code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
```diff
@@ -154,28 +154,8 @@
...
-			to_chat(H, "<span class='danger'>An ear-splitting guttural roar tears through your mind and makes your world convulse!</span>")
-			H.stunned += stun_duration
-			H.KnockDown(stun_duration)
-			H.apply_damage(halloss_damage, HALLOSS)
-			if(!H.ear_deaf)
-				H.ear_deaf += stun_duration * 20  //Deafens them temporarily
-			//Perception distorting effects of the psychic scream
-			addtimer(CALLBACK(GLOBAL_PROC, /proc/shake_camera, H, stun_duration * 10, 0.75), 31)
+	for(var/mob/living/L in range(world.view))
+		L.screech_act(src)
```
Rohesie: You could check for `stat`. Something like
```
if(L.stat == DEAD)
	continue
L.screech_act(src)
```
That way it wouldn't waste time processing the dead.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1750,4 +1750,21 @@
 
 	hud_set_squad()
 
-	return TRUE
\ No newline at end of file
+	return TRUE
+
+/mob/living/carbon/human/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
+	shake_camera(src, 3 SECONDS, 1) //50 deciseconds, SORRY 5 seconds was way too long. 3 seconds now
```
Rohesie: The prior behavior checked for client before shaking camera:
```
if(client)
	shake_camera(src, 3 SECONDS, 1) //50 deciseconds, SORRY 5 seconds was way too long. 3 seconds now

File: code/modules/mob/living/carbon/xenomorph/xenomorph.dm [hunk 10]
```diff
@@ -258,4 +258,8 @@
 	return
 
 /mob/living/carbon/Xenomorph/reagent_check(datum/reagent/R) //For the time being they can't metabolize chemicals.
-	return TRUE
\ No newline at end of file
+	return TRUE
+
+
+/mob/living/carbon/Xenomorph/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
+	shake_camera(src, 1 SECONDS, 1)
```
Rohesie: Same here, could use an `if(client)`

code/modules/mob/living/carbon/human/human.dm
psykzz: `shake_camera()` checks for client which is why i removed it.

code/modules/mob/living/carbon/human/human.dm
psykzz: Can this just be removed as we rely on `range(world.view)` instead?

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
psykzz: > `shake_camera()` checks for client which is why i removed it.

Rohesie: Good catch.

code/modules/mob/living/carbon/human/human.dm
Rohesie: It's more of a sanity check. Ideally, yes, there's no need for this.
Maybe do a
```
if(dist > 7)
	stack_trace("[src] called `screech_act()` with [dist] dist to the source of the screech, [Q]")
	return
```
That will inform us if it actually happens due to lag or something.
The problem with sanity checks is when they are silent, for the most part.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1750,4 +1750,21 @@
 
 	hud_set_squad()
 
-	return TRUE
\ No newline at end of file
+	return TRUE
+
+/mob/living/carbon/human/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
```
spookydonut: move this to human_defense.dm

code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
spookydonut: unless you need to handle stuff that isnt mobs theres no reason to change this from view() or viewers()

code/modules/mob/living/living.dm
```diff
@@ -616,6 +616,9 @@ below 100 is not dizzy
 	M.fully_replace_character_name(M.real_name, real_name)
 	GLOB.offered_mob_list -= src
 
+// When they are affected by a queens screech
+/mob/living/proc/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
```
spookydonut: living_defense.dm

code/modules/mob/mob.dm
```diff
@@ -812,4 +812,4 @@ mob/proc/yank_out_object()
 		else
 			to_chat(usr, "<span class='notice'>You just did an audible emote. Wait a while.</span>")
 			return TRUE
-	return FALSE
\ No newline at end of file
+	return FALSE
```
spookydonut: dont commit this stuff.

code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
psykzz: is there a difference between `view` or `viewers`?

Rohesie: Yes.
`range()` is the most efficient among them, and it will return a list of every (movable?) atom in a certain distance, regardless of conditions.

`view()` will return a list of (movable?) atoms visible by the source. It is less efficient, and it will create quirky behaviors, such as screech not affecting mobs that are next to the source but can't be seen by it (inside a locker/tarp for example) and smoke blocking screech.

`viewers()` is the least efficient among them in general, and it would create even quirkier behavior, with someone blindfolding themselves right before the screech happens to become unaffected, for example. It returns a list of mobs that can see the source.

I'd say `range()` is the best option here.

psykzz:
@spookydonut regarding view/viewers vs range. see https://github.com/tgstation/TerraGov-Marine-Corps/pull/1328#discussion_r277458085

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -366,3 +366,25 @@ Contains most of the procs that are called when a mob is attacked by something
 	if(!istype(C)) return
 	if(!(unique_access in C.access)) return
 	return 1
+
+
+/mob/living/carbon/human/screech_act(mob/living/carbon/Xenomorph/Queen/Q)
+	shake_camera(src, 3 SECONDS, 1) //50 deciseconds, SORRY 5 seconds was way too long. 3 seconds now
+	var/dist = get_dist(src, Q)
+
+	if(dist > 7)
```
DominikPanic: Clamp would also be a possible solution here

DominikPanic: Also, admins procalling may want a higher dist than 7, this is meant more of a sanity check for values like 100 or 1000 or 1000000000

Rohesie: Hmm, good point. Just remove the check, really.

code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
```diff
@@ -154,28 +154,10 @@
...
-		if(dist < 8)
-			to_chat(H, "<span class='danger'>An ear-splitting guttural roar tears through your mind and makes your world convulse!</span>")
-			H.stunned += stun_duration
-			H.KnockDown(stun_duration)
-			H.apply_damage(halloss_damage, HALLOSS)
-			if(!H.ear_deaf)
-				H.ear_deaf += stun_duration * 20  //Deafens them temporarily
-			//Perception distorting effects of the psychic scream
-			addtimer(CALLBACK(GLOBAL_PROC, /proc/shake_camera, H, stun_duration * 10, 0.75), 31)
+	for(var/mob/living/L in range(world.view))
```
Rohesie:
```suggestion
	for(var/mob/living/carbon/C in range(world.view))
```
Since you've decided not to keep the behavior of shaking the camera of all `living`, there's no reason to define `screech_act()` for `living` and call it only to return.
Define it for `carbon` instead then.

psykzz: I didn't plan to change anything and overlooked that everything should shake. Adding that back in.

PR #1351, opened 2019-04-23
Add more stats for start of round observers

psykzz:
## About The Pull Request

Add the new player stats to observers at the start of a round. This way if you observe early you can keep track of player readyness and how long until the round actually starts.


psykzz:
Tested locally because i dont want to be on the list
image

PR #1353, opened 2019-04-24
replace warding armor bonus with overhealth

spookydonut:
## About The Pull Request
as per the discussion in #balance to make warding less OP 

## Why It's Good For The Game
Make warding useful for pushes

## Changelog
:cl:
balance: Warding now generates a special bonus health instead of giving an armor bonus
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -121,6 +121,13 @@
 
 	damage = process_rage_damage(damage, src)
 
+	if(overhealth < damage)
+		damage -= overhealth
+		overhealth = 0
+	else
+		overhealth -= damage
+		damage = 0
+
 	switch(damagetype)
 		if(BRUTE)
 			adjustBruteLoss(damage, FALSE)
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -119,6 +119,8 @@
 	if(stat == DEAD)
 		return FALSE
 
+	last_damage_time = world.time
+
 	damage = process_rage_damage(damage, src)
 
 	if(overhealth < damage)
@@ -127,6 +129,7 @@
 	else
 		overhealth -= damage
 		damage = 0
+		return TRUE
 
 	switch(damagetype)
 		if(BRUTE)
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -130,8 +130,11 @@
 /mob/living/carbon/Xenomorph/proc/handle_overhealth()
 	if(overhealth > XENO_MAX_OVERHEALTH) // too much overhealth
 		overhealth = max(0, overhealth - XENO_OVERHEALTH_DECAY_RATE)
-	if(overhealth < XENO_MAX_OVERHEALTH) // not enough overhealth
-		overhealth = min(overhealth + XENO_OVERHEALTH_GEN_RATE, XENO_MAX_OVERHEALTH)
+	else if(overhealth < XENO_MAX_OVERHEALTH) // not enough overhealth
+		var/gen_rate = XENO_OVERHEALTH_GEN_RATE
+		if(last_damage_time + 10 SECONDS < world.time)
+			gen_rate *= 5
+		overhealth = min(overhealth + gen_rate, XENO_MAX_OVERHEALTH)
 
 /mob/living/carbon/Xenomorph/proc/handle_living_plasma_updates()
 	var/turf/T = loc
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -28,6 +28,7 @@
 		handle_aura_emiter()
 
 	handle_aura_receiver()
+	handle_overhealth()
 	handle_living_health_updates()
 	handle_living_plasma_updates()
 	update_action_button_icons()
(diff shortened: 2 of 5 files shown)


psykzz:
With B18's being the way they are, and the compliants people have already, how is this going to affect that?

spookydonut:
> 
> 
> With B18's being the way they are, and the compliants people have already, how is this going to affect that?

This will allow warding and gun damage to be more easily balanced.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -121,6 +121,13 @@
 
 	damage = process_rage_damage(damage, src)
 
+	if(overhealth < damage)
+		damage -= overhealth
+		overhealth = 0
+	else
+		overhealth -= damage
+		damage = 0
```
Rohesie:
```suggestion
		damage = 0
		updatehealth()
		return TRUE
```

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -120,7 +121,17 @@
 	adjustBruteLoss(-amount)
 	adjustFireLoss(-amount)
 
+#define XENO_OVERHEALTH_PER_WARDING_LEVEL 25
```
Rohesie:
```suggestion
#define XENO_OVERHEALTH_PER_WARDING_LEVEL 20
```
Let's be a little more conservative here. From 20 to 100 extra raw HP seems like a lot already.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -120,7 +121,17 @@
...
+#define XENO_OVERHEALTH_DECAY_RATE 5
+#define XENO_OVERHEALTH_GEN_RATE 2.5
 
+#define XENO_MAX_OVERHEALTH ( ( warding_aura * XENO_OVERHEALTH_PER_WARDING_LEVEL ) )
+
+/mob/living/carbon/Xenomorph/proc/handle_overhealth()
```
Rohesie: It would be better for the procs handling `overhealth` fluctuation not to let it go under or over the threshold, instead of having to police it every tick in `life()`

Ketrai:
Usually with queen warding the average xeno will get 80 extra health from this?  (queen mature pheromones at strenght 4)

As far as I can tell you've removed both the armor bonus and the flat damage reduction %.

A runner with this can get about 240 hp. A praetorian 330, and a crusher easily 430.

This seems to benefit low health castes much more than it benefits high health castes. Whilst it may benefit casts with innately high armor like crusher/warrior/queen/defender too much as they can get a lot of mileage out of that extra hp.

Interesting. I like it though, because without all the stupid reductions and calculations it'll be a hell of a lot easier to determine which guns are too strong.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -120,7 +121,17 @@
...
+	if(overhealth > XENO_MAX_OVERHEALTH) // too much overhealth
+		overhealth = max(0, overhealth - XENO_OVERHEALTH_DECAY_RATE)
+	if(overhealth < XENO_MAX_OVERHEALTH) // not enough overhealth
```
Rohesie:
```suggestion
	else if(overhealth < XENO_MAX_OVERHEALTH) // not enough overhealth
```

Rohesie:
> This seems to benefit low health castes much more than it benefits high health castes. Whilst it may benefit casts with innately high armor like crusher/warrior/queen/defender too much as they can get a lot of mileage out of that extra hp.

This is actually a serious balance concern. Overhealth should benefit from no armor reduction, it should be detracted first.

code/modules/mob/living/carbon/xenomorph/life.dm
spookydonut: given they have to build it up over time it feels fine, esp with the loss of all the damage reductions.

spookydonut:
> 
> 
> > This seems to benefit low health castes much more than it benefits high health castes. Whilst it may benefit casts with innately high armor like crusher/warrior/queen/defender too much as they can get a lot of mileage out of that extra hp.
> 
> This is actually a serious balance concern. Overhealth should benefit from no armor reduction, it should be detracted first.

no, keep it simple as part of take_damage proc otherwise it will be hell.
we can change the amount of overhealth castes get based on armor or natural health if need be but the key to overhealth is its simple and easy to balance.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
spookydonut: I figured updatehealth() needs to be called also.

Rohesie:
It's much simpler and easier to balance if it doesn't benefit from the caste's natural armor.
And it gives us more parameters to work with: armor-affecting HP and non-armor-affecting one. Former best dealt with AP, latter best dealt with raw damage.
Else we have the current AP situation, in where it's incredibly hard not to make it either useless or OP.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
Rohesie: What for? There wasn't any change in health.

code/modules/mob/living/carbon/xenomorph/life.dm
Rohesie: Small health pool that regenerates fast is better than big health pool that takes a long time regenerating.
The waiting game is not very engaging in general, nor is facing a foe that has waited a long time.

Ketrai:
After a few tests with this it seems to be benefitting ranged castes the most. Who can tank a whopping three extra bullets before really feeling the pain. If anything it is much weaker than the old warding. Especially for melee castes who pretty much relied on it. It regenerates way too slowly. It should regenerate faster if you're at full health or if you're resting. The values seem fine, it's the regeeneration that's the main issue. As it stands it takes about two to three minutes for it to heal to full?????

The lack of armor reduction bonus means that smartgunners and other low AP high damage weaponry now deal too much damage. And melt down xenomorphs stupidly fast. Even the normal m41a feels a shitload stronger. The only gun that feels relatively the same is the lasgun, because that gun wasn't affected as much by armor as the others.

psykzz:
Define out of combat. No dmg in 10 seconds and away from marines.
Increase it by 500% if out of combat.

Rohesie:
A warding nerf will necessarily unbalance things. That is to be expected. But this implementation, while it has good points, has some issues, as stated before.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -119,6 +119,8 @@
 	if(stat == DEAD)
 		return FALSE
 
+	last_damage_time = world.time
```
psykzz: I know this is currently only useful for xeno's but would it make sense to move this to living. So that other things can make use of a combat timer?

spookydonut: not currently no

Ketrai:
re-review this REEEE It'S BEEN TWO WEEKS

Rohesie:
The essence of the feature hasn't changed from my perspective. The points I made still stand.
Get more #balance people to comment on it.

PR #1356, opened 2019-04-24
Ground turfs

Rohesie:
## About The Pull Request

* Moved plating to its own file and tweaked it only a little to be closer to /tg/'s.
* Unified the different ground map open turfs into a unified `/ground` subtype.
* Separated the ones with firm terrain, in where one would expect to build (to `/turf/open/floor/plating/ground`) from the ones that serve a different purpose, such as rivers (to `/turf/open/ground`), following the example of how /tg/ handles asteroid turfs.

This is mostly to increase compatibility with /tg/, really. Their asteroid turfs could be a subtype of our ground one.

## Changelog
:cl:
balance: You can no longer build on water tiles.
/:cl:


Rohesie:
This should solve what #1316 intends to, and should be tested separately.
It's a big PR and while I did some local tests there's a good chance I missed something.
Better give it some testing and watch out for quirky behavior.

Barricades use snowflakey code. They can still be built on rivers but not on some jungle tiles. Would fix that but atomizing and all.

PR #1366, opened 2019-04-25
Removes the SMOKE_OPAQUE flag. 

Ghommie:
## About The Pull Request
Reverting back to how smoke opaqueness used to work, I was told the new one made it more difficult for defilers and (young/mature) boilers to stop the line-of-sigth of snipers and turrets.

## Why It's Good For The Game
Concerns balance.

## Changelog
:cl:
balance: Reverts the way smoke opaqueness works to make smaller clouds decent at shrouding line of sight again.
/:cl:


Rohesie:
Why not just adding `SMOKE_OPAQUE` to the gasses that people are complaining that should have it?

Ghommie:
> 
> 
> Why not just adding `SMOKE_OPAQUE` to the gasses people are complaining about?

All smokes minus camo and the flashbang particles had that flag. It's just the way it worked being different. Basically, the newly spwaned smokes would have become effectively opaque only if they did still had a non-zero range value, thus making the smoke puffs around the border unable to stop line of sight.

Ketrai:
mhm, you could see like up to three tiles into a gas cloud... which is pretty silly. Afaik one of the intented purposes of boiler and defiler gas is to provide cover.

PR #1367, opened 2019-04-25
Tweaks on xeno sting and larval growth serum.

Rohesie:
As requested by @Ketrai .
This will have to be tested for balance, as it can make a marine with tricord and tramadol quite the tank for as long as they last before bursting.

## Changelog
:cl:
balance: Xeno stings will stop injecting chems on overdosed infected hosts.
balance: Larval growth serum provides decent healing and strong analgesic properties to the host, both stacking with other chems. OD made a little softer. Still able to push you over if you are at the edge.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/powers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/powers.dm
@@ -925,6 +925,10 @@ GLOBAL_LIST_INIT(acid_spray_hit, typecacheof(list(/obj/structure/barricade, /obj
 		face_atom(C)
 		if(stagger)
 			return FALSE
+		body_tox = C.reagents.get_reagent(toxin)
+		if(CHECK_BITFIELD(C.status_flags, XENO_HOST) && body_tox && body_tox.volume > body_tox.overdose_threshold)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/powers.dm
```diff
@@ -925,6 +925,10 @@ GLOBAL_LIST_INIT(acid_spray_hit, typecacheof(list(/obj/structure/barricade, /obj
 		face_atom(C)
 		if(stagger)
 			return FALSE
+		body_tox = C.reagents.get_reagent(toxin)
+		if(CHECK_BITFIELD(C.status_flags, XENO_HOST) && body_tox && body_tox.volume > body_tox.overdose_threshold)
```
spookydonut: should this maybe be critical overdose threshold

Rohesie: That might increase the number of accidental OD deaths and change the dynamic of the chemical.
I understand the issue with basically ignoring the critical obverdose effects, because they would kind of not appear in game anymore, but I believe it's pretty clear that they are not needed and just cause salt with capture kills instead.

PR #1373, opened 2019-04-26
Moves game preference verb to preferences

psykzz:
## About The Pull Request
Moves Game Preferences to Preferences

## Changelog
:cl:
tweak: Moves Game Preferences to Preferences tab
/:cl:

image

### DIFF
--- a/code/modules/admin/preferences_verbs.dm
+++ b/code/modules/admin/preferences_verbs.dm
@@ -1,3 +1,10 @@
+/client/verb/setup_character()
+	set name = "Game Preferences"
+	set category = "Preferences"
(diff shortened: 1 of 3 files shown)


code/modules/admin/preferences_verbs.dm
```diff
@@ -1,3 +1,10 @@
+/client/verb/setup_character()
```
DominikPanic: This is the wrong place for it. This is for admin-only preference verbs.

psykzz: As a ghost you have a ton of procs there as well.
That was my inspiration.

DominikPanic: Only admin-only verbs are there.

PR #1374, opened 2019-04-26
Two new admin verbs: "Check Contents" and "Randomize Name"

DominikPanic:
Also adds a few more debug messages to topic.dm


code/modules/admin/panels/player_panel.dm
```diff
@@ -464,20 +464,24 @@
 
 
 	if(!isnewplayer(M))
-		body += {"<br><br>
-			<b>Other actions:</b>
-			<br>
-			<a href='?src=[ref];thunderdome=[REF(M)]'>Thunderdome</a> |
-			<a href='?src=[ref];gib=[REF(M)]'>Gib</a>"}
+		body += "<br><br>"
```
Rohesie:
```suggestion
		body += "<br><br>\
			<b>Other actions:</b>\
			<br>\
			<a href='?src=[ref];thunderdome=[REF(M)]'>Thunderdome</a> | \
			<a href='?src=[ref];gib=[REF(M)]'>Gib</a>"
```
A single operation while retaining clarity.

DominikPanic: I'd argue this is much easier on the eyes, plus admin code doesn't have to squeeze every last drop of performance.

code/modules/admin/debug_verbs.dm
```diff
@@ -370,38 +370,39 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
...
 
-	var/mob/M
+	var/mob/living/L
 	switch(choice)
 		if("Key")
 			var/selection = input("Please, select a key.", "Check Contents") as null|anything in sortKey(GLOB.clients)
 			if(!selection)
 				return
-			M = selection:mob
+			L = selection:mob
```
Rohesie:
```suggestion
			var/client/C = selection
			L = C.mob
```
Or, more sanely, `var/client/selection = input(yaddayadda)`

code/modules/admin/debug_verbs.dm
```diff
@@ -370,38 +370,39 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
...
 		if("Mob")
-			var/selection = input("Please, select a mob.", "Check Contents") as null|anything in sortNames(GLOB.mob_list)
+			var/selection = input("Please, select a mob.", "Check Contents") as null|anything in sortNames(GLOB.mob_living_list)
 			if(!selection)
 				return
-			M = selection
+			L = selection
 
-	if(!isliving(M))
+	if(!istype(L))
```
Rohesie:
```suggestion
	if(!istype(L))
		to_chat(usr, "<span class='warning'>Target is no longer valid.</span>")
```

PR #1375, opened 2019-04-26
Weed subsystem

psykzz:
## About The Pull Request

Weeds now regrow and are managed by a new weeds subsystem.

## Why It's Good For The Game

This means marines need to actually attack nodes to kill off the weeds.

## Changelog
:cl:
balance: weeds continue to grow back until the main node is destroyed
server: new weeds subsystem
/:cl:


code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,70 @@
...
+	var/obj/effect/alien/weeds/node/N
+	while(length(currentrun))
+		N = currentrun[currentrun.len]
+		currentrun.len--
+		if(!N)
+			weed_nodes.Remove(N)
+			continue
+
+		// PROCESS N
+		for(var/turf/T in N.node_turfs)
```
Rohesie: If you don't expect the list to have anything but turfs, it's more efficient to do
```
for(var/X in N.node_turfs) 
	var/turf/T = X
```
As it saves an implicit `istype()` check.

code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,70 @@
...
+		return FALSE
+	weed_nodes.Remove(N)
+
+/datum/controller/subsystem/weeds/proc/generate_weed_graph(obj/effect/alien/weeds/node/N, node_size)
+	var/list/node_turfs = list()
+	var/list/turfs_to_check = list()
+	turfs_to_check += get_turf(N)
+	while (node_size > 0)
+		node_size--
+		for(var/turf/T in turfs_to_check)
```
Rohesie: Same here, this can only contain turfs.

code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,70 @@
...
+			for(var/turf/AdjT in T.AdjacentTurfs())
```
Rohesie: And here.

code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,70 @@
...
+				if (AdjT == N) // Ignore the node
+					continue
+				if (AdjT in node_turfs) // Ignore existing weeds
+					continue
+
+				turfs_to_check += AdjT
+				node_turfs += AdjT
+
+
+	for(var/turf/smallT in node_turfs)
```
Rohesie: Same here.

code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,101 @@
...
+/datum/controller/subsystem/weeds/proc/remove_node(obj/effect/alien/weeds/node/N)
+	if(!N)
+		stack_trace("SSweed.remove_node called with a null obj")
+		return FALSE
+	weed_nodes.Remove(N)
+
+
+/datum/controller/subsystem/weeds/proc/create_weed(turf/T, obj/effect/alien/weeds/node/N)
+
+	if (!T.is_weedable())
```
Rohesie: Remove this from the `node_turfs`, as it's no longer a valid spot.

code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,101 @@
...
+		return
+
+	var/obj/effect/alien/weeds/W = locate() in T
+	if (W)
```
Rohesie: Is it possible for this to happen?

psykzz: Yeah if two nodes are near each other, they can have overlapping graphs

Rohesie: Make it check for parent-node as described in Discord, and either remove the turf from the graph if it does have a parent or adopt the weed into the node (set this node as its parent) if it has none.

code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,105 @@
...
+/datum/controller/subsystem/weeds/stat_entry()
+	return ..("Nodes: [length(processing)]")
+
+/datum/controller/subsystem/weeds/fire(resumed = FALSE)
+	if(!resumed)
+		currentrun = processing.Copy()
+		creating = list()
+
+	for(var/A in currentrun)
+		var/turf/T = A
```
spookydonut: dont it like this just keep it in a simple processing list

Rohesie:
Depending on how expensive this subsystem is, you could add an `if(generate_weed_graph())` check when planting weeds, make it `return TRUE` at the end, but make it `return` if it finds another node, so as to limit how close you can put purple sacs.

Would require you to store the result of `generate_weed_graph()`, namely `node_turfs`.
That would mean `return node_turfs` at the end, and since that's a list, the check going `if(length(generate_weed_graph()))`

psykzz:
The downside there @Rohesie is that by pairing the nodes if you kill the other node you can destroy both graphs or get left with a crippled graph.

Ideally I'd like all weeds destroyed within the graph of a node to regenerate. It may mean weeds become too strong lategame and might need some tweaking, but something im keeping an eye on.

Rohesie:
Ah, I didn't mean pairing the nodes. I meant not to allow two nodes to be planted in expansion distance, as to reduce the maximum number of possible nodes a little.

psykzz:
Thats still something i'd like to have as a balance reason, if you over index on weeds you have multiple nodes to destroy before the weeds stop growing.

If the system does introduce lag, i think we can implement your idea later, but for now if it can handle it, i'd like to keep it as is.

code/controllers/stonedmc/subsystem/weeds.dm
psykzz: Do you mean process a single list?

If so, that wouldn't allow me to check for the health of a node. We add turfs on weed death for regrowth and they dont have a way to find their parent (without explicitly looking which id like to avoid).

Rohesie:
Sure.

code/controllers/stonedmc/subsystem/weeds.dm
```diff
@@ -0,0 +1,106 @@
...
+		if(MC_TICK_CHECK)
+			return
+
+	// We create weeds outside of the loop to not influence new weeds within the loop
+	for(var/A in creating)
+		var/turf/T = A
+		var/obj/effect/alien/weeds/node/N = creating[T]
+		creating -= T
+		// Adds a bit of jitter to the spawning weeds.
+		spawn(rand(0,10))
```
DominikPanic: Callback perhaps?

Rohesie:
@spookydonut , please review.

PR #1378, opened 2019-04-27
Gun Rebalance for Warding

Ketrai:
## About The Pull Request

A rebalance for guns to go with the warding change of #1353 

Lowers the damage of all guns by 10%

Lowers armor penetration of all guns by 10%

Lowers smartgun and autocannon armor penetration by 25%. (these two were being hit the hardest by the old warding, and have always been balanced accordingely. So a harsher change only seems fair)

Lowers pulse rifle armor penetration by 55% for default ammo from 10 to 4.5. (Unironically regular ammo did more damage against even armored targets like warriors, defenders, praetorians, etc. This will make it so AP actually does more damage on average against armored targets.)

AP rocket and LBT armor penetration remains the same as they were already ludicrously high, it wouldn't have affected them anyways.

Specialist weapons.

Sniper rifle
- Lowered sniper rifle damage by 20%
- Increased sniper rifle fire-rate by 30%
- Lowered laser shot AP and Damage by 15% (to 35% bonus)
- Increased mag size from 15 to 18
This will allow you to keep a target slowed more effectively, whilst also toning down the obnoxiously high damage per shot allowing you to clean up high value targets too easily.

Battle rifle 
- simply removed the ability to add a barrel charger to the gun.

ZX assault shotgun
- lowered damage by 15%
- upgraded fire-rate by 10%
- Decreased the accuracy penalty on burst from 30% to 20%
- Removed recoil compensator, extended barrel, gyroscope. These attachments allow for too much min-maxing. You could just override the negatives of this gun (accuracy, spread) and make it a single target murder machine at decent range of 6-7 tiles. (the spread would easily overlap using scatter attachments where all three flechettes hit the same tile.)

Minigun
- burst increased from 10 to 30
- Scatter increased from 20% to 35% (reduced somewhat by specialist skill ofc)
- Ammo per mag increased from 300 to 420 (even number for bursts)
- Damage lowered by 20%
I feel like the minigun is a bit too good at straight up dealing damage and focusing down single targets. If I understand correctly, this weapon was intended to help hold FOBs, and to supress xenos, deter them from getting close. I feel like these changes will help meet that purpose. It's harder  to hit a single target, significantly so. And you have to commit to a burst as you won't stop firing until those thirty bullets make it out.

Flamethrower : No change needed.

## Why It's Good For The Game

To tune back the powercreep and glass-cannon mentality. With the new warding the average 25-30% overall effectiveness reduction of guns will be gone. This makes it so weapons don't just shred xenomorphs with that change in place.

As a side benefit, this also makes friendly fire less retardedly strong. And make HVH slightly more bearable. 

**DO NOT  TEST MERGE THIS WITHOUT THE WARDING REBALANCE TO OVERHEALTH MERGED AS WELL, IT WOULD MAKE GUNS TOO WEAK.**

- **IMPORTANT** : All changes were tested. 

## Changelog
:cl: Ketrai
balance: Overall gun damage and  penetration reduced by 10%
balance: ZX shotgun => Removes extended barrel, recoil compensator and gyroscope. Damage lowered by 15%. Accuracy on burst penalty decreased from 30 to 20%. Fire-rate increased by 10%
balance: Smartgun and Autocannon AP reduced by an additional 25%
balance: Pulse rifle default ammo AP reduced by 55% AP ammo remains unchanged.
balance: AP rocket and LBT Armor penetration values remain unchanged. Damage is lowered like all other guns.
balance: Battle rifle => removes barrel charger.
balance: Sniper rifle => Increases fire-rate by 30%. Decreases laser bonus by 15%, now 35% bonus. Decreases damage by 20%. Magazine size increased  fron 15 to 18
balance: Minigun => Burst increased from 10 to 30. Damage reduced by 20%. Magazine size increased by 120 to 420. Scatter increased from 20 to 35%
/:cl:


Ketrai:
As a footnote. Smartgun and sentries are still on the watch list. Being able to crit praetorians in about 16 bullets with stupidly high accuracy puts them dangerously close to a m41a who can crit a praetorian in about 11 bullets. 

Ap bullets are now always on par with normal m41a bullets. Dealing the same damage to decently armored targets like a mature praetorian. Whilst dealing more damage to medium armored targets like mature to elder warriors, defenders in crest mode/fortify, and doing more damage to crusher side and crest armor.

DominikPanic:
```
Ap bullets are now always on par with normal m41a bullets.
```
The intended role of AP is to be better only in some certain circumstances, and if this is really the case it takes away from the unique ascept of it. Now there's little reason to shoot that runner with AP when you have it.

code/__DEFINES/conflict.dm
```diff
@@ -116,8 +116,8 @@
 
 //Define sniper laser multipliers
 
-#define SNIPER_LASER_DAMAGE_MULTIPLIER	1.5 //+50% damage vs the aimed target
-#define SNIPER_LASER_ARMOR_MULTIPLIER	1.5 //+50% penetration vs the aimed target
+#define SNIPER_LASER_DAMAGE_MULTIPLIER	1.35 //+35% damage vs the aimed target
```
DominikPanic: Could turn these into combat defines.

Ketrai:
> ```
> Ap bullets are now always on par with normal m41a bullets.
> ```
> The intended role of AP is to be better only in some certain circumstances, and if this is really the case it takes away from the unique ascept of it. Now there's little reason to shoot that runner with AP when you have it.

You misunderstood me. If  you shoot a runner with Ap bullets you will do LESS damage. 15-20% less to be exact. 

I fixed the case where normal bullets did more damage to armored targets like praetorians, warriors, defenders, etc.

Rohesie:
Fighting powercreep is good, though I'm really not a fan of how we assign damage around here.
This along won't be enough in any case, we need to fix how armor happens, but it's a good step.

Do not TM this without a warding nerf PR.

code/controllers/configuration/entries/combat_defines.dm
```diff
@@ -122,55 +122,55 @@ Damage.
 Damage of the projectile, though the gun itself can modify it. This is the main source of projectile damage.
 */
 /datum/config_entry/number/combat_define/base_hit_damage
-	config_entry_value = 10
+	config_entry_value = 9
 
 /datum/config_entry/number/combat_define/min_hit_damage
-	config_entry_value = 15
+	config_entry_value = 13.5
```
Rohesie: This kind of change is not going to pass, we want the combat defines to be more or less consistent, not to add another layer of obscurity to them.
The solution is to nerf specific weapons instead. More work, yes, but less random results.

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -895,10 +894,7 @@
 						/obj/item/attachable/reddot,
 						/obj/item/attachable/verticalgrip,
 						/obj/item/attachable/angledgrip,
-						/obj/item/attachable/gyro,
 						/obj/item/attachable/flashlight,
-						/obj/item/attachable/extended_barrel,
```
Rohesie: The ZX is overpowered at the moment, there's no denying that, but this is outside of the scope of the PR and warrants its own, as it's not just de-escalating damage, but changing how the gun works.

DominikPanic:
Roh's points still stand

PR #1385, opened 2019-04-27
VV improvements: rotate, varedit updates, link to VV'd object

DominikPanic:
Rohesie's request

### DIFF
--- a/code/modules/admin/verbs/datumvars.dm
+++ b/code/modules/admin/verbs/datumvars.dm
@@ -693,6 +693,8 @@
 			return
 
 		M.fully_replace_character_name(M.real_name, new_name)
+		vv_update_display(M, "name", new_name)
+		vv_update_display(M, "real_name", M.real_name || "No real name")
 
 		message_admins("[ADMIN_TPMONTY(usr)] renamed [ADMIN_TPMONTY(M)] to [new_name].")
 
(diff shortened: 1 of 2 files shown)


code/modules/admin/verbs/datumvars.dm
```diff
@@ -927,6 +929,26 @@
...
+			return
+
+		var/atom/A = locate(href_list["rotatedatum"])
+		if(!istype(A))
+			to_chat(usr, "This can only be done to instances of type /atom")
+			return
+
+		switch(href_list["rotatedir"])
+			if("right")
+				A.setDir(turn(A.dir, -45))
```
spookydonut: some things won't play nicely with 45 degree rotations

DominikPanic: This is /tg/ code, what would you suggest?

code/modules/admin/verbs/datumvars.dm
```diff
@@ -693,6 +693,8 @@
 			return
 
 		M.fully_replace_character_name(M.real_name, new_name)
+		vv_update_display(M, "name", new_name)
```
spookydonut: shouldnt this be in #1384

DominikPanic: This is VV/admin tools related, that one is for a game proc

PR #1388, opened 2019-04-27
Makes the chat loading smoother

DominikPanic:
Improves the matching proc to be /tg/'s, moves it to the same place (client/New())
Chat loading upon login and init should be much smoother now.

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -172,7 +172,6 @@ GLOBAL_VAR_INIT(external_rsc_url, TRUE)
 	if(CONFIG_GET(flag/log_access))
 		for(var/I in GLOB.clients)
 			if(!I || I == src)
-				stack_trace("null or new client same as existing one in GLOB.clients during client /New()")
 				continue
 			var/client/C = I
 			if(C.key && (C.key != key) )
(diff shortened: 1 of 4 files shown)


code/modules/client/client_procs.dm
```diff
@@ -169,6 +169,27 @@ GLOBAL_VAR_INIT(external_rsc_url, TRUE)
 
 	var/full_version = "[byond_version].[byond_build ? byond_build : "xxx"]"
 
+	if(CONFIG_GET(flag/log_access))
+		for(var/I in GLOB.clients)
+			if(!I || I == src)
```
Rohesie: Could you trace stack null vars in `GLOB.clients`?

DominikPanic: They will never be there otherwise may as well add it to all for loops

DominikPanic: But yeah it's /tg/code so why not

Rohesie: Could just remove the check as well.

code/modules/client/client_procs.dm
```diff
@@ -172,6 +172,7 @@ GLOBAL_VAR_INIT(external_rsc_url, TRUE)
 	if(CONFIG_GET(flag/log_access))
 		for(var/I in GLOB.clients)
 			if(!I || I == src)
+				stack_trace("null or new client same as existing one in GLOB.clients during client /New()")
```
Rohesie: "client same as existing" is expected to happen, but null is not.

PR #1389, opened 2019-04-27
/tg/ machinery/button

Rohesie:
## About The Pull Request

An incredibly watered-down version of how /tg/ deals with buttons, without assemblies/devices, and still using the /tg/ icons.
The old buttons are not replaced yet. Added a door control button, which opens and closes doors, and one that only opens there for future (soon :tm: ) gameplay purposes.

## Why It's Good For The Game

By itself it doesn't do much, but it sets a base for new buttons to be made as children of the `button` parent, instead of each a different `machinery`, and also for converting old existent buttons to the new format.
Also a step into the direction of making buttons buildable and more customizable, as /tg/ does. Power to the engineers and all that jazz.

Keeping it small to avoid bloating the PR.


code/game/machinery/buttons.dm
```diff
@@ -1,3 +1,111 @@
+#define DOOR_FLAG_OPEN_ONLY (1 << 0)
```
DominikPanic: Could you move this to the machine defines?

Rohesie: It's being undefined at the end of the file. This flag won't likely survive it if we add assemblies and electronics as intended.
This is a very watered-down version of /tg/ buttons because a ton of stuff has been striped away.

PR #1400, opened 2019-04-28
Updates the selector arrays to latest /tg/ version

DominikPanic:


DominikPanic:
@JJRcop

JJRcop:
Thank you. It took a bit of a long time for it to be merged upstream. I don't have a problem with this, but the unfortunate, but also great part of it is that I was asked to kill the MAP list unwrap which I thought was going to suck but it actually was great.

I moved the (subexpression) logic out of expression and unary_expression into variable, which enables a fully recursive code chain of expression:arrow_right:value:arrow_right:variable:arrow_right:expression. This probably has a much greater impact on the possibilities that I could think of, and it's a great change.

PR #1401, opened 2019-04-28
Adds preference for window flashing

DominikPanic:


code/__HELPERS/game.dm
```diff
@@ -405,11 +405,11 @@ datum/projectile_data
...
 
 
-/proc/window_flash(client/C)
+/proc/window_flash(client/C, ignorepref = FALSE)
 	if(ismob(C))
 		var/mob/M = C
 		if(M.client)
 			C = M.client
-	if(!C)
+	if(!C || (!C.prefs.windowflashing && !ignorepref))
```
Rohesie:
```suggestion
	if(!C?.prefs.windowflashing && !ignorepref)
```

code/__HELPERS/game.dm
```diff
@@ -405,11 +405,11 @@ datum/projectile_data
 	flick_overlay(I, viewing, duration)
 
 
-/proc/window_flash(client/C)
+/proc/window_flash(client/C, ignorepref = FALSE)
 	if(ismob(C))
 		var/mob/M = C
 		if(M.client)
```
Rohesie: Given you are checking for `C` in the next line, you can remove this `if(M.client)` check.

code/__HELPERS/game.dm
```diff
@@ -405,11 +405,11 @@ datum/projectile_data
...
 			C = M.client
```
psykzz: Why is it that given a client it gets the mob, then resets the original client to the mobs client?

psykzz: Why is it that given a client it gets the mob, then resets the original client to the mobs client?

DominikPanic: Because unga devs can call it with a mob, just like to-chat

code/__HELPERS/game.dm
```diff
@@ -405,11 +405,10 @@ datum/projectile_data
...
 
-/proc/window_flash(client/C)
+/proc/window_flash(client/C, ignorepref = FALSE)
 	if(ismob(C))
 		var/mob/M = C
-		if(M.client)
-			C = M.client
-	if(!C)
+		C = M.client
+	if(!C?.prefs.windowflashing && !ignorepref)
```
Rohesie:
```suggestion
	if(!ignorepref && !C?.prefs.windowflashing)
```
Better check if the ignore pref before. Micro optimization.

PR #1404, opened 2019-04-28
Add a few more stats

psykzz:
## About The Pull Request

Add a few more stats into the stat panel.
Round ID and Time Dilation

## Why It's Good For The Game

Helps answer the question "is the lag me, or the server?"

Pregame lobby
image

Mid round
image

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -21,7 +21,8 @@
 	. = ..()
 
 	if(statpanel("Stats"))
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob.dm
```diff
@@ -21,8 +21,10 @@
 	. = ..()
 
 	if(statpanel("Stats"))
+		stat("Round ID: [GLOB.round_id ? GLOB.round_id : "NULL"]")
```
DominikPanic: Could make it only show up if there is one

PR #1414, opened 2019-04-29
Fixes resin walls cannibalizing weeds

### LINKED ISSUE #1023: making and destroying resin walls fucks with the floor beneath
if you make a resin wall and destroy it, the floor beneath gets fucked with:
snow reappears on rice colony
directional floor tiles get forced to face down

psykzz:
## About The Pull Request

Fixes #1023 

---

Not generally happy about this so happy to learn of a better way.

### DIFF
--- a/code/game/turfs/walls/wall_types.dm
+++ b/code/game/turfs/walls/wall_types.dm
@@ -368,6 +368,11 @@
 	if(!locate(/obj/effect/alien/weeds) in loc)
 		new /obj/effect/alien/weeds(loc)
 
+/turf/closed/wall/resin/ChangeTurf(new_turf_path, forget_old_turf, flags)
+	. = ..(new_turf_path, forget_old_turf, flags)
+	if(forget_old_turf)
+		new /obj/effect/alien/weeds(src)
+
 /turf/closed/wall/resin/flamer_fire_act()
 	take_damage(50)
 
--- a/code/game/turfs/walls/wall_types.dm
+++ b/code/game/turfs/walls/wall_types.dm
@@ -369,7 +369,7 @@
 		new /obj/effect/alien/weeds(loc)
 
 /turf/closed/wall/resin/ChangeTurf(new_turf_path, forget_old_turf, flags)
-	. = ..(new_turf_path, forget_old_turf, flags)
+	. = ..()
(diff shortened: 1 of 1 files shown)


code/game/turfs/walls/wall_types.dm
```diff
@@ -368,6 +368,11 @@
 	if(!locate(/obj/effect/alien/weeds) in loc)
 		new /obj/effect/alien/weeds(loc)
 
+/turf/closed/wall/resin/ChangeTurf(new_turf_path, forget_old_turf, flags)
+	. = ..(new_turf_path, forget_old_turf, flags)
```
Rohesie:
```suggestion
	. = ..()
```
Those vars will be passed by default.

psykzz: Oh cool thanks.

code/game/turfs/walls/wall_types.dm
```diff
@@ -368,6 +368,11 @@
...
+	if(forget_old_turf)
```
Rohesie: Do early return instead:
```
if(!forget_old_turf)
	return
new /obj/effect/alien/weeds(src)
```

code/game/turfs/walls/wall_types.dm
```diff
@@ -368,6 +368,11 @@
 	if(!locate(/obj/effect/alien/weeds) in loc)
 		new /obj/effect/alien/weeds(loc)
 
+/turf/closed/wall/resin/ChangeTurf(new_turf_path, forget_old_turf, flags)
+	. = ..()
+	if(forget_old_turf)
+		new /obj/effect/alien/weeds(src)
```
Ghommie: Due to how turfs work, the new turf will have deleted the current one and thus, But since the parent version has a return of said new turf:
```suggestion
		new /obj/effect/alien/weeds(.)
```
doing this should possibly work.

psykzz: Testing this it did work locally, but i agree its prob referencing the wrong turf.

Ghommie: Mh, that makes me wonder even more about how damned undocumented some standard features of byond are.

PR #1415, opened 2019-04-29
Fix Prae speed exploit

### LINKED ISSUE #1117: Praetorian Speed bug/exploit
## Testmerges:
Tested with no testmerges.

## Reproduction:
use the praetorian's acid spray ~5 ticks/seconds before upgrading. That will increase the speed to -2.15 which it shouldn't be at. 

in my testing I used var edit to not have to wait the time to get near the upgrade.

psykzz:
## About The Pull Request

Fixes #1117

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/praetorian/abilities_praetorian.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/praetorian/abilities_praetorian.dm
@@ -48,11 +48,11 @@
 
 	speed += 2
 	do_acid_spray_cone(target)
-	addtimer(CALLBACK(src, .speed_increase, 2), rand(20,30))
+	addtimer(CALLBACK(src, .reset_speed), rand(20,30))
 	addtimer(CALLBACK(src, .acid_spray_cooldown_finished), xeno_caste.acid_spray_cooldown)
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/praetorian/abilities_praetorian.dm
```diff
@@ -48,11 +48,11 @@
...
 	speed += 2
 	do_acid_spray_cone(target)
-	addtimer(CALLBACK(src, .speed_increase, 2), rand(20,30))
+	addtimer(CALLBACK(src, .reset_speed), rand(20,30))
 	addtimer(CALLBACK(src, .acid_spray_cooldown_finished), xeno_caste.acid_spray_cooldown)
 
-/mob/living/carbon/Xenomorph/proc/speed_increase(var/amount)
-	speed -= amount
+/mob/living/carbon/Xenomorph/proc/reset_speed()
+	speed = xeno_caste.speed
```
Rohesie: Does this take into account the age of the xeno?

psykzz: Yes, `xeno_caste` is set on levelup and is what is used during that process to set the original speed value.

PR #1416, opened 2019-04-29
Fixes Xeno healing without a queen present

### LINKED ISSUE #1137: No Queen means no heals which means no drones evolving.
## Testmerges:

#827 #1100 #1129 #1130 #1131 #1134 #1132 #1128 #1109 #1104 #1091 #1046 #1033 #952

## Reproduction:

Apparently, being a xeno on the Theseus during the crash phase and losing the queen means no one can heal. 

This includes drones. Who need to be at full health to evolve.

I got the feeling that drones should always be able to heal on weeds. Not sure if that would be too snowflakey codewise though.

image

psykzz:
## About The Pull Request

Fixes #1137 

The reason healing was stopped from Discord
> [21:13] Ketrai / One Spitty Boi / Beaver: @PsyKzz it's only on different z-levels, still works if the corpse is on the same z-level. And it's mainly to stop a single lurker from sneaking on the ship and killing the entire crew

:cl:
balance: Xeno's now have half healing (up from none) when a queen is not present.
/:cl:


code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -95,12 +95,15 @@
 	if(!T || !istype(T))
 		return
 
+	var/queen_healing_penalty = 0.5
 	if(!hive?.living_xeno_queen || hive.living_xeno_queen.loc.z == loc.z) //if there is a queen, it must be in the same z-level
```
Ghommie:
```suggestion
	if(!hive || (hive.living_xeno_queen && hive.living_xeno_queen.loc.z == T.z)) //if we are hiveless or the living queen's z-level is the same as ours.
```
An old mistake made by me.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -110,12 +110,15 @@
...
 		return
 
-	if(!hive?.living_xeno_queen || hive.living_xeno_queen.loc.z == loc.z) //if there is a queen, it must be in the same z-level
-		if(locate(/obj/effect/alien/weeds) in T || xeno_caste.caste_flags & CASTE_INNATE_HEALING) //We regenerate on weeds or can on our own.
-			if(lying || resting)
-				heal_wounds(XENO_RESTING_HEAL)
-			else
-				heal_wounds(XENO_STANDING_HEAL) //Major healing nerf if standing.
+	var/queen_healing_penalty = 0.5
+	if(!hive || (hive.living_xeno_queen && hive.living_xeno_queen.loc.z == T.z)) //if we are hiveless or the living queen's z-level is the same as ours.
```
spookydonut:
```suggestion
	if(hive?.living_xeno_queen?.loc?.z == T.z)) //if the living queen's z-level is the same as ours.
```
no xeno should ever be hiveless.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -110,12 +110,15 @@
...
 		return
 
-	if(!hive?.living_xeno_queen || hive.living_xeno_queen.loc.z == loc.z) //if there is a queen, it must be in the same z-level
-		if(locate(/obj/effect/alien/weeds) in T || xeno_caste.caste_flags & CASTE_INNATE_HEALING) //We regenerate on weeds or can on our own.
-			if(lying || resting)
-				heal_wounds(XENO_RESTING_HEAL)
-			else
-				heal_wounds(XENO_STANDING_HEAL) //Major healing nerf if standing.
+	var/queen_healing_penalty = 0.5
+	if(hive?.living_xeno_queen?.loc?.z == T.z)) //if the living queen's z-level is the same as ours.
```
spookydonut:
```suggestion
	if(hive?.living_xeno_queen?.loc?.z == T.z) //if the living queen's z-level is the same as ours.
```

code/modules/mob/living/carbon/xenomorph/life.dm
DominikPanic: That is bad don't do checking if something exists with comparison. This way if the left-side is null and they are in nullspace it would pass.

DominikPanic:
Travis

PR #1428, opened 2019-04-30
Shift xenos to use armor datums

spookydonut:
will require xeno signals to be done first.


DominikPanic:
Seems to be currently worked on in #1685 ,do you still plan on to continue this?

PR #1430, opened 2019-04-30
Fixes camera autoname bug

### LINKED ISSUE #1346: Dropship camera bug
## Testmerges:

None

## Reproduction:

There's only 1 camera avaliable in dropship camera controls. Not sure if this is mapping or code issue.

Alamo - 
изображение

Normandy - 
изображение

psykzz:
## About The Pull Request
Fixes #1346

### DIFF
--- a/code/game/machinery/camera/presets.dm
+++ b/code/game/machinery/camera/presets.dm
@@ -62,27 +62,32 @@
 
 // AUTONAME
 
+var/static/list/id_by_area = list()
+
 /obj/machinery/camera/autoname
 	var/number = 0 //camera number in area
 
 //This camera type automatically sets it's name to whatever the area that it's in is called.
 /obj/machinery/camera/autoname/New()
 	..()
-	addtimer(CALLBACK(src, .proc/generate_c_tag), 1 SECONDS, TIMER_OVERRIDE|TIMER_UNIQUE)
-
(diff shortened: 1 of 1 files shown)


code/game/machinery/camera/presets.dm
```diff
@@ -68,17 +68,11 @@
...
-				if(C == src) continue
-				var/area/CA = get_area(C)
-				if(CA.type == A.type)
-					if(C.number)
-						number = max(number, C.number+1)
-			c_tag = "[A.name] #[number]"
+	var/static/list/id_by_area = list()
+	var/area/A = get_area(src)
+	if(!id_by_area[A])
+		id_by_area[A] = 0
```
ninjanomnom: I took this out in the example earlier in discord because this if block is unnecessary

code/game/machinery/camera/presets.dm
```diff
@@ -68,17 +68,11 @@
 //This camera type automatically sets it's name to whatever the area that it's in is called.
 /obj/machinery/camera/autoname/New()
```
ninjanomnom: Change to initialize while you're here

psykzz: 👍 , what is the difference?

ninjanomnom: New happens outside our control at server start and switching to initialize lets us actually manage how and when it happens

PR #1432, opened 2019-05-01
sublime users dont have access to dm langserver

spookydonut:


ninjanomnom:
Why would this need to be testmerged

DominikPanic:
> 
> 
> Why would this need to be testmerged

Because spooky wants every PR that is TM'd to have this tag

spookydonut:
> 
> 
> Why would this need to be testmerged

PR #1440, opened 2019-05-01
Fixes queen being unable to launch the shuttle

DominikPanic:
Noone reported it during testing of the nanoui PR :cry:

### DIFF
--- a/code/modules/nano/interaction/default.dm
+++ b/code/modules/nano/interaction/default.dm
@@ -80,6 +80,13 @@ GLOBAL_DATUM_INIT(default_state, /datum/topic_state/default, new)
 		return STATUS_UPDATE
 
 
+/mob/living/carbon/Xenomorph/default_can_use_topic(src_object)
+	. = shared_nano_interaction(src_object)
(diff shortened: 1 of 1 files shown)


code/modules/nano/interaction/default.dm
```diff
@@ -80,6 +80,13 @@ GLOBAL_DATUM_INIT(default_state, /datum/topic_state/default, new)
 		return STATUS_UPDATE
 
 
+/mob/living/carbon/Xenomorph/default_can_use_topic(src_object)
+	. = shared_nano_interaction(src_object)
+	if(. != STATUS_CLOSE)
+		if(loc)
+			. = min(., loc.contents_nano_distance(src_object, src))
```
psykzz: Can you explain the return type here?

Above it looks like a define to explain whats happening, CLOSE, REFRESH, ETC.
However the 2nd return feels more like a distance variable.

Is that not mixing types?

DominikPanic: Both of these are a number, which is just an abstraction of the states.

PR #1451, opened 2019-05-03
Fixes EORD shuttle windows

### LINKED ISSUE #1126: Shuttle Windows act like non-full tile windows
## Reproduction:

The shuttle windows in the EORG arena can be walked through, as if they're a south-facing directional window. (IE: The ones in briefing.)

DominikPanic:
Fixes #1126

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -441,6 +441,7 @@
 	basestate = "window"
 	max_integrity = 40
 	reinf = TRUE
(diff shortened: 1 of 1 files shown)


francinum:
This should probably be generalized up to the basetype of the non-edge windows.

DominikPanic:
> 
> 
> This should probably be generalized up to the basetype of the non-edge windows.

This is the only non-edge subtype, all others have directional icon states.

PR #1452, opened 2019-05-03
Fixes lasgun underbarrel attachments

### LINKED ISSUE #1078: Masterkey On M37 Bug
## Testmerges:

No testmerges. Code from 3/13. If this has been noticed and fixed since then, feel free to delete. 

## Reproduction:

Right so, just put a masterkey shotgun on an M37, and try to reload the masterkey. It automatically switches to loading the M37's tube instead. Using buckshot, which is the ammo the masterkey takes.

DominikPanic:
Fixes #1078

### DIFF
--- a/code/modules/projectiles/updated_projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/updated_projectiles/gun_helpers.dm
@@ -292,7 +292,7 @@ should be alright.
 	else if(active_attachable?.flags_attach_features & ATTACH_RELOADABLE && check_inactive_hand(user))
 		active_attachable.reload_attachment(I, user)
 
-	else if(istype(I, /obj/item/ammo_magazine) && check_inactive_hand(user))
+	else if((istype(I, /obj/item/ammo_magazine) || istype(I, /obj/item/cell/lasgun)) && check_inactive_hand(user))
 		reload(user, I)
 
 
--- a/code/modules/projectiles/updated_projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/updated_projectiles/gun_helpers.dm
@@ -289,21 +289,11 @@ should be alright.
 			attach_to_gun(user,I)
 
  	//the active attachment is reloadable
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -289,21 +289,11 @@ should be alright.
...
-					active_attachable.activate_attachment(src, null, TRUE)
-					reload(user,MG)
-					return
-			active_attachable.reload_attachment(I, user)
-
-	else if(istype(I,/obj/item/ammo_magazine))
-		if(check_inactive_hand(user))
-			reload(user,I)
+	else if(active_attachable?.flags_attach_features & ATTACH_RELOADABLE && check_inactive_hand(user))
+		active_attachable.reload_attachment(I, user)
```
spookydonut: shouldnt this return?

code/modules/projectiles/updated_projectiles/guns/energy.dm
```diff
@@ -290,31 +290,14 @@
...
-				playsound(user, active_attachable.activation_sound, 15, 1)
-				active_attachable.activate_attachment(src, null, TRUE)
-				reload(user,MG)
-				return
-		active_attachable.reload_attachment(I, user)
-
-	else if(istype(I,/obj/item/cell/lasgun))
-		if(check_inactive_hand(user))
-			reload(user,I)
+	else if(istype(I, /obj/item/cell/lasgun) && check_inactive_hand(user))
```
spookydonut: as above, if you reload an attachment with a lasgun cell wont this also trigger or potentially try to reload with a qdelled cell if the attachment qdels the cell as part of the reload

code/modules/projectiles/updated_projectiles/gun_helpers.dm
DominikPanic: Only one else-if will trigger.

Rohesie: A `return` would add clarity to the code.
This might be subjective, though.

PR #1459, opened 2019-05-03
Aggressive changelogs

psykzz:
## About The Pull Request

Allows admins to set if they want to show the changelog on connect if something new has been added.

## Why It's Good For The Game

Keep players informed


DominikPanic:
Not sure how I feel about this - considering the fact we have 0 downstreams. Things should be optional and not forced upon people.

PR #1463, opened 2019-05-04
Removes purrbation

Rohesie:
Kind of not really fitting for our game mode and the tone of our game.

Besides, if we want to add cosmetic items, it must be for everyone, and not just for some. This set bad precedents, as some people would be denied felinid aesthetics while others would be granted, depending on admin mood, stance and preference. Too much room for problems, and really bad for RP immersion.

If Felinids are added to the code will be a larger decision, and it will follow some RP standards. Or we'll throw the RP rules out of the window. Can't really be saying much about LRP if we are doing this as an administration.

## Changelog
:cl:
del: Removes purrbation. No more felinid ears and tail on prayer-demand.
/:cl:


AffectedArc07:
***reeEEEEeEEeEEEEEeEeEeEEE***

PR #1468, opened 2019-05-04
Supply drops begone.

Rohesie:
I have no idea why this was a ever a thing.

## Changelog
:cl:
del: There is no longer a 5% chance of the distress beacon returning just a couple of supply crates instead of an ERT with actual people in it.
/:cl:

### DIFF
--- a/code/datums/emergency_calls/supplies.dm
+++ b/code/datums/emergency_calls/supplies.dm
@@ -41,7 +41,7 @@
 	name = "Supply Drop"
 	mob_max = 0
 	mob_min = 0
-	probability = 5
+	probability = 0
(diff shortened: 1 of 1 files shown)


spookydonut:
testing not required

DominikPanic:
> 
> 
> testing not required

It is not whether testing is required or not, but it helps when selecting which PRs are okay and to be TM'd. Just because it doesn't NEED to be TM'd doesn't mean it shouldn't.

Rohesie:
@spookydonut , please don't wordlessly remove tags. Else it becomes an unpleasant tug of war game.

spookydonut:
We discussed it on discord

PR #1474, opened 2019-05-04
adds an isnestedhost helper macro

spookydonut:
## About The Pull Request
increase readability.


code/modules/mob/living/carbon/xenomorph/castes/crusher/crusher.dm
```diff
@@ -248,7 +248,7 @@
 					count++
 			if(count)
 				X.charge_speed -= CHARGE_SPEED_BUILDUP / (count * 2) // half normal slowdown regardless of number of corpses.
-		else if(!(status_flags & XENO_HOST) && !istype(buckled, /obj/structure/bed/nest))
+		else if(!isnestedhost(src))
```
psykzz: Original code only hit you if you were neither, this way you will only get hit if your not nested and your not a host.

spookydonut: good catch

PR #1489, opened 2019-05-05
remove defunct old shuttle code

spookydonut:


code/modules/shuttles/shuttle.dm
```diff
@@ -134,25 +108,10 @@
...
-				spawn(0)
-					D.open()
-	else
-		for(var/obj/machinery/door/airlock/unpowered/D in area)
-			if(D.density && !D.locked)
-				spawn(0)
-					D.open()
+	for(var/obj/machinery/door/airlock/unpowered/D in area)
+		if(D.density && !D.locked)
+			spawn(0)
```
Rohesie: Is the `spawn(0)` necessary?

spookydonut: this is gonna get replaced eventually by the shuttle port but in this case yes it does because open/close use a sleep

PR #1515, opened 2019-05-07
Possibly fixes SQL connection issues, adds SQL automatic reconnecting

DominikPanic:


Rohesie:
:optimumtact_lakillerflushed:

PR #1518, opened 2019-05-07
synth repair

grinkgo:
## About The Pull Request

Changes the behavior of a synthetic using cable coil and/or welders on themselves. They can now repair their own brute and burn damage after a short delay. Having someone else repair them is still faster, though.

## Why It's Good For The Game

It gives synthetic players a bare amount of self-sufficiency, which can reduce their frustration and dependency on other players, especially during lowpop rounds where there may not be any engineers. The player experience is better overall because instead of not being able to repair themselves at all, they have the ability to albeit at a slow rate. It's important to have that self-sufficiency without eliminating any need for other players at all.

## Changelog
:cl:
tweak: Synth players rejoice! You no longer have to rely on pesky humans to repair your own body (however, they still do it faster than you can).
soundadd: repairing synths with a welder now makes the welding noise.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/cable_coil.dm
+++ b/code/game/objects/items/stacks/cable_coil.dm
@@ -135,7 +135,7 @@
 								"<span class='warning'>You repair some burn damage on \the [H]'s [S.display_name] with \the [src].</span>")
 			return
 		else
-			to_chat(user, "Nothing to fix!")
+			to_chat(user, "<span class='warning'>Nothing to fix!</span>")
 
 	else
 		return ..()
--- a/code/game/objects/items/stacks/cable_coil.dm
+++ b/code/game/objects/items/stacks/cable_coil.dm
@@ -128,7 +128,7 @@
 		
 		if(S.burn_dam > 0 && use(1))
 			if(issynth(H) && M == user)
-				if(user.action_busy || !do_after(user, 5 SECONDS, TRUE, 5, BUSY_ICON_BUILD))
+				if(user.action_busy || !do_after(user, 5 SECONDS, TRUE, src, BUSY_ICON_BUILD))
 					return
 			S.heal_damage(0,15,0,1)
 			user.visible_message("<span class='warning'>\The [user] repairs some burn damage on \the [H]'s [S.display_name] with \the [src].</span>", \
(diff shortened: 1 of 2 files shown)


code/game/objects/items/tools/maintenance_tools.dm
```diff
@@ -202,20 +202,19 @@
...
 			S.heal_damage(15,0,0,1)
 			H.UpdateDamageIcon()
 			user.visible_message("<span class='warning'>\The [user] patches some dents on \the [H]'s [S.display_name] with \the [src].</span>", \
 								"<span class='warning'>You patch some dents on \the [H]'s [S.display_name] with \the [src].</span>")
 			remove_fuel(1,user)
+			playsound(user.loc, 'sound/items/Welder2.ogg', 25, 1)
 			return
 		else
-			to_chat(user, "<span class='warning'>Nothing to fix!</span>")
+			to_chat(user, "Nothing to fix!")
```
DominikPanic: Don't remove useful formatting for no reason

code/game/objects/items/stacks/cable_coil.dm
```diff
@@ -118,20 +118,20 @@
     
 /obj/item/stack/cable_coil/attack(mob/M as mob, mob/user as mob)
 	if(hasorgans(M))
+		var/mob/living/carbon/human/H = M
 		var/datum/limb/S = M:get_limb(user.zone_selected)
+
+		if(!S) return
```
DominikPanic: Single line control statement (press enter after if(!S))

code/game/objects/items/tools/maintenance_tools.dm
grinkgo: I just did that to match the cable coil message

DominikPanic: Then you should add formatting to that. Making the experience worse is never the correct choice.

grinkgo: gotcha

code/game/objects/items/stacks/cable_coil.dm
grinkgo: :ok_hand:

code/game/objects/items/stacks/cable_coil.dm
```diff
@@ -118,23 +118,24 @@
     
 /obj/item/stack/cable_coil/attack(mob/M as mob, mob/user as mob)
 	if(hasorgans(M))
```
spookydonut: this is conceptually haslimbs(), but for now it should stay like this given theres plans to give non-humans limbs at some point.

PR #1520, opened 2019-05-08
Purges borg code

DominikPanic:
Not only is borg code currently super ancient, but this will make port of AI code and AI eye much easier, plus we can just decide to port borgs later if they are found desirable.


code/__HELPERS/game.dm
```diff
@@ -135,21 +135,15 @@
...
-					continue //No radio component (Shouldn't happen)
-				if(!borg.is_component_functioning("radio") || !borg.cell_use_power(CO.active_usage))
-					continue //No power.
-
-			var/turf/speaker = get_turf(R)
-			if(speaker)
-				for(var/turf/T in hear(R.canhear_range,speaker))
-					speaker_coverage[T] = T
+		if(!R)
+			continue
```
psykzz: Would it make sense to `radios -= R` here as well to skip it next check?

DominikPanic: If you continue, the same item doesn't get checked again

PR #1534, opened 2019-05-09
[s] Fixes protected lists being able to be shown and edited

DominikPanic:
For those rowdy varediting admins

### DIFF
--- a/code/modules/admin/verbs/varedit.dm
+++ b/code/modules/admin/verbs/varedit.dm
@@ -351,11 +351,12 @@ GLOBAL_PROTECT(VVpixelmovement)
 
 	switch(alert("Would you like to associate a value with the list entry?",,"Yes","No"))
 		if("Yes")
-			L[var_value] = mod_list_add_ass(O) //hehe
-	if (O)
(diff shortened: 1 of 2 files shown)


code/modules/admin/verbs/varedit.dm
```diff
@@ -351,23 +351,25 @@ GLOBAL_PROTECT(VVpixelmovement)
...
 	switch(alert("Would you like to associate a value with the list entry?",,"Yes","No"))
 		if("Yes")
-			L[var_value] = mod_list_add_ass(O) //hehe
-	if (O)
-		if (O.vv_edit_var(objectvar, L) == FALSE)
-			to_chat(src, "Your edit was rejected by the object.")
-			return
+			L[var_value] = mod_list_add_ass(O)
+
+	if(!O?.vv_edit_var(objectvar, L))
```
ninjanomnom: Were you intending to change behavior here so it tells you the edit was rejected if `O` is null?

DominikPanic: Not intended, goofed up, fixed

PR #1538, opened 2019-05-09
Mends a not-so-little race condition smoke runtime on rockets & co.

Ghommie:
## About The Pull Request
Damn race conditions.

## Why It's Good For The Game
Less runtiming.

## Changelog
:cl:
fix: Fixes more ammo datum smokes stuff, such as some borked rockets, apologises for the inconvenience.
fix: Fixes the halloss from dirty smartgun bullets.
/:cl:


code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -94,6 +94,8 @@
 			M.smokecloak_off()
 
 /obj/effect/particle_effect/smoke/proc/apply_smoke_effect(turf/T)
+	if(!T)
+		return
```
spookydonut:
```suggestion
		qdel(src)
		return
```
if theres no turf it should delete itself

Rohesie:
Why is `apply_smoke_effect()` being called without a turf? Is that intended to happen?

Ghommie:
> 
> 
> Why is `apply_smoke_effect()` being called without a turf? Is that intended to happen?

There was no safety check existing to prevent this runtime from happening under ever so rarely possible fallacies, race hazards and so on. Now incorporating outer safety checks from prevent caller procs from keep running too.

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -150,6 +153,8 @@
 
 //proc to check if smoke can expand to another turf
 /obj/effect/particle_effect/smoke/proc/check_airblock(turf/T)
+	if(!T)
```
Rohesie: This proc is called by `spread_smoke()` in where `get_adjacent_open_turfs()` already filters invalid values. There is no need for this check.

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -94,6 +94,9 @@
 			M.smokecloak_off()
 
 /obj/effect/particle_effect/smoke/proc/apply_smoke_effect(turf/T)
+	if(!T && !QDELING(src))
```
Rohesie: Same here for the `if(!T)` check. This could only happen if there's somehow a `/obj/effect/particle_effect/smoke` in null space.
`apply_smoke_effect()` is being called on three different procs, you might want to check that logic.

code/modules/projectiles/updated_projectiles/ammo_datums.dm
```diff
@@ -1343,21 +1308,20 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
...
+	drop_nade(T)
 
 /datum/ammo/xeno/acid/heavy/do_at_max_range(obj/item/projectile/P)
-	drop_acid(get_turf(P))
+	drop_nade(get_turf(P))
 
-/datum/ammo/xeno/acid/proc/drop_acid(turf/T) //Leaves behind a short lived acid pool; lasts for 1-3 seconds.
-	if(T.density)
+/datum/ammo/xeno/acid/drop_nade(turf/T) //Leaves behind a short lived acid pool; lasts for 1-3 seconds.
+	if(!T || T.density)
```
Rohesie: There should always be a turf. What is going on here?

code/game/objects/effects/effect_system/smoke.dm
Ghommie: What I'm doing here is specifically stop the proc from runtiming when the source ever happens to be in null space. As I said again, I'm placing one safety check inside the proc rather than three outside.
I could suppose the reason why the src was in null space due to the smoke spread affixed location, but very finnicky hazards due to the smoke being slow to set up that'll be patched on the way.
Plus the runtime report gives almost no useful information, originating from the processing subsystem ignite().

Rohesie: Having three checks outside the proc instead of one inside is not necessarily a bad thing in DM, proc overhead considered. On other languages the alternative approach might make more sense.

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -52,11 +54,13 @@
...
 
 /obj/effect/particle_effect/smoke/proc/kill_smoke()
-	if(CHECK_BITFIELD(smoke_traits, SMOKE_CAMO))
-		apply_smoke_effect(get_turf(src))
 	LAZYCLEARLIST(cloud?.smoked_mobs)
 	STOP_PROCESSING(SSobj, src)
 	INVOKE_ASYNC(src, .proc/fade_out)
+	if(CHECK_BITFIELD(smoke_traits, SMOKE_CAMO))
+		var/turf/T = get_turf(src)
+		if(T)
```
Rohesie: Alright, the idea here is to trace back to the origin of the problem, instead of silencing it later on.
In here (and in the other times) `get_turf(src)` would return `null` if this was in nullspace. Such as when it's being QDELETED. Can you think of another alternative?
If it's being QDELETED, why is that happening? That's the question.

Ghommie: I can guess the problem with the smoke being in nullspace to start with may be the holder happening to be in null space too. The safety check on start() is limited to the affixed location or holder being qdeleted or not as of now.
I'll open up another PR to add safety check and warning on the smoke spread datum start(), will need a TM to find the real origin of this issue.

Ghommie: #1596. Keeping this PR open the time being should the issue not relate to the datum's start()

Ghommie:
That runtime is elusive and Rohesie told me to close this for the time being.

Ghommie:
Necro time. I have to fix Ammo datums race conditions and don't want to open a new PR. Awaken my masters.

PR #1541, opened 2019-05-10
Fixes a runtime for non-human forced emotes

DominikPanic:
Monkeys can gasp yet again


Rohesie:
Hold on, `gasp` is defined to be a human emote. Monkeys not gasping is standard behavior unless that changes, no?

DominikPanic:
> Hold on, `gasp` is defined to be a human emote. Monkeys not gasping is standard behavior unless that changes, no?

Hardcoded .emote( calls bypass things.

Rohesie:
/tg/ has it set to `/datum/emote/living/gasp` which seems to be the more proper implementation.
Else the solution is a hacky patch.

PR #1548, opened 2019-05-10
Cleanup queen larva code

psykzz:
## About The Pull Request

Looking into why larva disappear on queen death, found this. Fixing it.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
@@ -367,7 +367,7 @@ to_chat will check for valid clients itself already so no need to double check f
 	if(!stored_larva) // no larva to deal with
 		return ..()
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/hive_datum.dm
```diff
@@ -367,7 +367,7 @@ to_chat will check for valid clients itself already so no need to double check f
 	if(!stored_larva) // no larva to deal with
 		return ..()
 
-	stored_larva = round(stored_larva * QUEEN_DEATH_LARVA_MULTIPLIER(Q)
+	stored_larva = round(stored_larva * QUEEN_DEATH_LARVA_MULTIPLIER(Q), 1)
```
Ghommie: It already rounds to the nearest integer by default, no need to explicate the latter arg.

psykzz: Personally don't agree as I feel explicit is better than implicit, that said, i made the change regardless.

PR #1555, opened 2019-05-11
Improves the "Select Rank" admin verb

DominikPanic:
Just like edit appearance, it's pretty and more functional now

### DIFF
--- a/code/modules/admin/fun_verbs.dm
+++ b/code/modules/admin/fun_verbs.dm
@@ -732,62 +732,38 @@
 	message_admins("[ADMIN_TPMONTY(usr)] changed the security level to code [sec_level].")
 
 
-/datum/admins/proc/select_rank(mob/living/carbon/human/H in GLOB.human_mob_list)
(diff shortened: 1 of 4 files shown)


code/modules/admin/fun_verbs.dm
```diff
@@ -732,62 +732,38 @@
...
+	if(!check_rights(R_FUN))
 		return
 
-	switch(alert("Modify the rank or give them a new one?", "Select Rank", "New Rank", "Modify", "Cancel"))
-		if("New Rank")
-			var/newrank = input("Select new rank for [H]", "Change the mob's rank and skills") as null|anything in sortList(SSjob.name_occupations)
-			if(!newrank || !istype(H))
-				return
+	if(!istype(H))
+		return
```
```suggestion
		CRASH("edit_rank() called for invalid type: [H]")
```

DominikPanic: No that's stupid look at literally all other procs. This just means a nonhuman was clicked there is literally zero need for this.

Issue #1560, opened 2019-05-11
Corgi runtime with new emotes


psykzz:
## Reproduction:

```
[13:42:25] Runtime in emote.dm,6: undefined proc or verb /mob/living/simple_animal/corgi/Ian/get paygrade().

  proc name: run emote (/datum/emote/living/carbon/human/run_emote)
  src: /datum/emote/living/carbon/hum... (/datum/emote/living/carbon/human/look)
  call stack:
  /datum/emote/living/carbon/hum... (/datum/emote/living/carbon/human/look): run emote(Leo (/mob/living/simple_animal/corgi/Ian), "at the area where the meat ste...", null, 0, null)
  Leo (/mob/living/simple_animal/corgi/Ian): emote("looks", null, null, 0)
  Meat steak (/obj/item/reagent_container/food/snacks/meatsteak): attack animal(Leo (/mob/living/simple_animal/corgi/Ian))
  Leo (/mob/living/simple_animal/corgi/Ian): Life(2, 33)
  Mobs (/datum/controller/subsystem/mobs): fire(0)
  Mobs (/datum/controller/subsystem/mobs): ignite(0)
```

in corgi.dm line 80 
```
emote("stares at the [movement_target] that [movement_target.loc] has with a sad puppy-face")
```
This is calling the human "stares" which checks for paygrade.

I tried to fix this myself, but don't know enough about emotes.

@DominikPanic for vis

PR #1563, opened 2019-05-11
Fixes evil emote runtime

DominikPanic:
Fixes: #1560
Fixes: #1570

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -1,5 +1,5 @@
 /datum/emote/living/carbon/human
-	mob_type_allowed_typecache = list(/mob/living/carbon/human)
+	mob_type_allowed_typecache = /mob/living/carbon/human
 
 
 /datum/emote/living/carbon/human/run_emote(mob/living/carbon/human/user, params, type_override, intentional = FALSE, prefix)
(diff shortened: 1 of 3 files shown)


Rohesie:
Is this for corgis to use `/datum/emote/living/carbon/human/run_emote()`?

DominikPanic:
> 
> 
> Is this for corgis to use `/datum/emote/living/carbon/human/run_emote()`?

the base run_emote runs can_run_emote, before that this was running it before checking if you were allowed to

Rohesie:
Why is `get_paygrade()` being defined at a mob level, though, if it's only checked in `/datum/emote/living/carbon/human/run_emote()`? Is something non-human calling that proc?

DominikPanic:
> 
> 
> Why is `get_paygrade()` being defined at a mob level, though, if it's only checked in `/datum/emote/living/carbon/human/run_emote()`? Is something non-human calling that proc?
> 
> Maybe try a rubber duck method here, explaining your change.

````
/datum/emote/proc/run_emote(mob/user, params, type_override, intentional = FALSE, prefix)
	. = TRUE
	if(!can_run_emote(user, TRUE, intentional))
		return FALSE
```
If I overwrite that proc
```
/datum/emote/living/carbon/human/run_emote(mob/living/carbon/human/user, params, type_override, intentional = FALSE, prefix)
	var/paygrade = user.get_paygrade()
	if(paygrade)
		prefix = "<b>[paygrade]</b> "
	return ..()
```
I have to ensure I either check if it can be ran, or define get_paygrade at the mob level.

PR #1577, opened 2019-05-12
Runner balance tweaks

Rohesie:
## Changelog
:cl:
balance: Runner speed got a tiny nerf. Their savage skill no longer stacks with the hit-and-run damage bonus either.
fix: Fixes a bug preventing xenos with special attacks from dealing extra damage, like ravager's rage and runner hit-and-run.
/:cl:


Ketrai:
the hit and run thing should be fixed. I don't see why runner speed needs a nerf tho? Especially when it is just a drop in the bucket. Feels like a pointless change that won't have any effect except nerfing runner by like 1%

Rohesie:
I'd like to see the impact first, but it's a tiny step on the direction of slowing it down a little.
With diagonal movement we'll have to slow down stuff, and the Runner is simply the fastest thing at the moment.

PR #1562, opened 2019-05-11
Fixes Rohesie's runtime

DominikPanic:
This is a shitty solution but hopefully Rohesie will fix it properly after seeing it

```

[23:16:05] Runtime in storage.dm, line 181: Cannot modify null.screen_loc.
proc name: slot orient objs (/obj/item/storage/proc/slot_orient_objs)
src: the survival pouch (/obj/item/storage/pouch/survival/full)
src.loc: Timothy Dugmore (/mob/living/carbon/human)
call stack:
the survival pouch (/obj/item/storage/pouch/survival/full): slot orient objs(0, 5, null)
the survival pouch (/obj/item/storage/pouch/survival/full): orient2hud()
the survival pouch (/obj/item/storage/pouch/survival/full): recalculate storage space()
the plasteel (/obj/item/stack/sheet/plasteel): update weight()
the plasteel (/obj/item/stack/sheet/plasteel): New(the survival pouch (/obj/item/storage/pouch/survival/full), 15)
the plasteel (/obj/item/stack/sheet/plasteel): New(the survival pouch (/obj/item/storage/pouch/survival/full), 15)
the survival pouch (/obj/item/storage/pouch/survival/full): Initialize(0)
Atoms (/datum/controller/subsystem/atoms): InitAtom(the survival pouch (/obj/item/storage/pouch/survival/full), /list (/list))
the survival pouch (/obj/item/storage/pouch/survival/full): New(0)
the survival pouch (/obj/item/storage/pouch/survival/full): New(Timothy Dugmore (/mob/living/carbon/human))
the survival pouch (/obj/item/storage/pouch/survival/full): New(Timothy Dugmore (/mob/living/carbon/human))
the survival pouch (/obj/item/storage/pouch/survival/full): New(Timothy Dugmore (/mob/living/carbon/human))
Distress Signal (/datum/game_mode/distress): transform survivor(Kira Tedrow (/datum/mind))
Distress Signal (/datum/game_mode/distress): pre setup()
Ticker (/datum/controller/subsystem/ticker): setup()
Ticker (/datum/controller/subsystem/ticker): fire(0)
Ticker (/datum/controller/subsystem/ticker): ignite(0)
```

PR #1527, opened 2019-05-09
increase init speed

spookydonut:
before:
Initialized Mapping subsystem within 22.1 seconds!
Initialized Atoms subsystem within 19.7 seconds!
after:
Initialized Mapping subsystem within 15.5 seconds!
Initialized Atoms subsystem within 18.9 seconds!

PR #1598, opened 2019-05-15
Update weight storage runtime fix

Rohesie:
Should do a better fix than #1562 for the specific problem.

### DIFF
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -504,9 +504,12 @@
 
 /obj/item/storage/Initialize(mapload, ...)
 	. = ..()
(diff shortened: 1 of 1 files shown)


code/game/objects/items/storage/storage.dm
```diff
@@ -507,8 +501,13 @@
 	for(var/obj/item/I in contents)
 		remove_from_storage(I, T)
 
-/obj/item/storage/New()
-	..()
+
+/obj/item/storage/Initialize(mapload, ...)
+	. = ..()
+	can_hold = typecacheof(can_hold)
```
spookydonut: https://github.com/tgstation/TerraGov-Marine-Corps/pull/1527

PR #1604, opened 2019-05-19
Fixes people not being able to respond to IRC admin PMs

DominikPanic:
Works
https://puu.sh/Dup2m/d05f29de41.png

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -1016,6 +1016,24 @@
 
 	else if(istype(whom, /client))
 		recipient = whom
+		
+
+
+	if(irc)
+		if(!ircreplyamount)	//to prevent people from spamming irc
+			return
+
+		if(!msg)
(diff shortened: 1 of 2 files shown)


code/modules/admin/admin_verbs.dm
```diff
@@ -1016,6 +1016,24 @@
 
 	else if(istype(whom, /client))
 		recipient = whom
+		
+
+
+	if(irc)
```
spookydonut: is this discord or irc

DominikPanic: The convention on /tg/ is to call anything external related IRC.

spookydonut: do we have working irc code?

DominikPanic: Yes. This both works with Discord  and irc. It only sends a message to TGS which then decides which channel to use (either irc or discord.)

PR #1600, opened 2019-05-15
Add ghost hud & alerts from TG

psykzz:
## About The Pull Request

This ports the ghost hud & alerts from /TG/

## Why It's Good For The Game

Improves ghost gameplay, which can get a little boring. 
This helps by highlighting events in the round, allowing ghosts to see more action.

## Changelog
:cl:
add: Added ghost hud
add: Added new ghost alerts, They show specific events to ghosts that happen throughout the round, allowing ghosts to see more action.
/:cl:

### Screens
<details>
<summary>Click to see some images of it all in action</summary>

**(Ghost messages with orbit / follow link)**
image

**(Alert image in the top right, up to 5)**
image

https://streamable.com/8zbkw

https://streamable.com/6sg54

</details>

Closes #1635

PR #1610, opened 2019-05-19
Makes bwoinks ignore the window flashing preference.

DominikPanic:


spookydonut:
Nothing worse than making windows flash if you dont want them to

DominikPanic:
> 
> 
> Nothing worse than making windows flash if you dont want them to

This is so people don't get noted for not responding to an admin while AFK. Not all people play with sound, this is something that everyone notices.

Rohesie:
For AdminPMs it makes sense.

Pariah919:
as someone who plays regularly without sound can approve

spookydonut:
If someone sets their client to not flash the client should respect that otherwise you should just remove the preference completely.

francinum:
You do realize this is one of the two places `window_flash` is ever actually used, the other being serverstart init completion.

DominikPanic:
> 
> 
> You do realize this is one of the two places `window_flash` is ever actually used, the other being serverstart init completion.

Psy's #1600 PR changes that. And I assume more will in the future.

Rohesie:
The flash only triggers when the window is not in focus. It is not horribly disturbing. If they are playing the game it won't disturb it. If they are alt+tabbed the window will just be a bit more insistent in that they are getting an admin message, which is kind of important here. They can always close the game if they wish to ignore admins.

In general I'd agree not to bypass preferences, but in this case it makes sense to.

PR #1587, opened 2019-05-13
Fixes benomorph plasma salvage.

Ghommie:
## About The Pull Request
Closes #1585. Also removes some an useless do_after check whose criterias will always be met, because no container should contain itself.

## Why It's Good For The Game
Bugfixing. Though the ability may need to be sped up, I originally made it too slow. Dunno if Surr will ever finish his PR.

## Changelog
:cl:
fix: Fixes drone's plasma salvage ability.
/:cl:

PR #1611, opened 2019-05-19
atomized do_after clean up.

Ghommie:
## About The Pull Request
Removing an useless check whose criteria should be always meet (as no container should contain itself) if I'm correct. Atomized from #1587.

## Why It's Good For The Game
Relatively small clean up.

## ~~Changelog~~

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -162,9 +162,8 @@ proc/age2agedescription(age)
 			break
 
 		if(!QDELETED(Tloc) && (QDELETED(target) || Tloc != target.loc))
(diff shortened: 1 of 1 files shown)


code/__HELPERS/mobs.dm
```diff
@@ -162,9 +162,8 @@ proc/age2agedescription(age)
 			break
 
 		if(!QDELETED(Tloc) && (QDELETED(target) || Tloc != target.loc))
-			if(Uloc != Tloc || Tloc != user)
```
ninjanomnom: This won't always pass

Ghommie:
Well then, consider this closed.

PR #1614, opened 2019-05-19
Fixes clicking other people to suicide

DominikPanic:


spookydonut:
Needs a valid description and changelog entry

Rohesie:
Same as said in the other PRs. Changelog entry is optional here unless we discuss this and decide otherwise, it has been so far.
Description would be desirable. In this case I know what it is about: if you targeted the mouth of someone else, on harm intent, you'd attempt to suicide with your gun.

Issue #1613, opened 2019-05-19
Vents and scrubbers are invisible under weed


MoXuT0:
## Testmerges:

Round ID:  506 
BYOND Version:  512.1464 
Server revision compiled on:  
Master commit: b0addb74d0b9f4169006a5743621ab1c08037b5c 
The following pull requests are currently test merged: 
#1596: 'Inspection on the little smoke runtime.' by Ghommie at commit 31870a2326 
#1598: 'Update weight storage runtime fix' by Rohesie at commit 09c40070f4 
#1600: 'Add ghost hud & alerts from TG' by psykzz at commit d51fa6ca4f 
#1597: 'Ports tool_act' by Jalleo at commit 137d7c918e 
#1604: 'Fixes people not being able to respond to IRC admin PMs' by DominikPanic at commit b4c45cd7d2 
#1551: 'Fixing smoke nades and buffing phosphorus smoke.' by Ghommie at commit da49786000 
#1605: 'Smartgun resprite' by MoXuT0 at commit 30fab39b0b 

## Reproduction:

Lurking in vents seems fine, problem is vents are hard to spot without right-clicking every tile or knowing where they are.

![ass](https://user-images.githubusercontent.com/34478337/57988126-025de580-7a93-11e9-8ef8-5a19b701bc74.png)

Rohesie:
Layer issue.

DominikPanic:
> 
> 
> Layer issue.

*plane issue. The layers are correct.

PR #1618, opened 2019-05-20
Fixes vents going under weeds.

DominikPanic:
:cl: LaKiller8
fix: Vents are now visible over weeds.
/:cl:
Fixes: #1613

### DIFF
--- a/code/modules/atmospherics/machinery/components/components_base.dm
+++ b/code/modules/atmospherics/machinery/components/components_base.dm
@@ -33,10 +33,8 @@
 	var/turf/T = loc
 	if(level == 2 || !T.intact_tile)
 		showpipe = TRUE
-		plane = GAME_PLANE
 	else
 		showpipe = FALSE
-		plane = FLOOR_PLANE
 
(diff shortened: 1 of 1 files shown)


DominikPanic:
> 
> 
> I would not call this a valid fix.

It works and doesn't break anything. Could you go into more detail and suggest an alternative, please?

spookydonut:
Removing plane code is not a fix, you need to fix weeds to work with planes correctly.

Rohesie:
This is a bug that severely hurts xenos and planes are not likely to happen anytime so soon.
I agree with not merging it, even though the feature shouldn't happen if incomplete, but TMing this will be kind of necessary.

spookydonut:
Planes are already mostly a thing.

Rohesie:
Alright, we'll TM it until then.

spookydonut:
> 
> 
> Alright, we'll TM it until then.

just copy the vent plane code for weeds, duh.

spookydonut:
ffs i'll fix it myself.

PR #1638, opened 2019-05-26
removes some of my own hackiness with lunge

spookydonut:
replaces a hacky list reference with a signal

:cl:
add: Ctrl-Click as a Warrior will activate lunge on whatever you click, if you're able to
/:cl:


spookydonut:
I tested this locally btw

PR #1640, opened 2019-05-27
Regex decapitalisation of xeno typepaths

spookydonut:
its finally come


Rohesie:
Good. This needed to happen.
Merge conflicts, though.

PR #1654, opened 2019-05-28
Cleans up /atom procs and defines

DominikPanic:
Removes unused procs and defines
Slightly cleans up and refactors existing one to /tg/ standard

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -2,20 +2,18 @@
 	layer = TURF_LAYER
 	plane = GAME_PLANE
 	var/level = 2
-	var/flags_atom = 0
-	var/fingerprintslast
-	var/list/blood_DNA
-	var/blood_color
-	var/last_bumped = 0
-	var/flags_pass = 0
-	var/throwpass = 0
-	var/germ_level = GERM_LEVEL_AMBIENT // The higher the germ level, the more germ on the atom.
 
-	///Chemistry.
+	var/flags_atom = NONE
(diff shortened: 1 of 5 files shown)


code/game/atoms.dm
```diff
@@ -332,45 +250,19 @@ its easier to just keep the beam vertical.
...
-			break
-//	to_chat(world, "X = [cur_x]; Y = [cur_y]")
-	if(cur_x && cur_y)
-		return list("x"=cur_x,"y"=cur_y)
-	else
-		return 0
-
 /atom/proc/checkpass(passflag)
-	return flags_pass&passflag
+	return CHECK_BITFIELD(flags_pass, passflag)
```
Rohesie: This could be a macro instead. It's never redefined. Legacy /tg/ code.

code/game/atoms.dm
```diff
@@ -76,8 +79,8 @@ directive is properly returned.
 /atom/proc/on_reagent_change()
 	return
 
-/atom/proc/Bumped(AM as mob|obj)
-	return
+/atom/proc/Bumped(atom/movable/AM)
+	set waitfor = FALSE
```
Rohesie: While this is the /tg/ standard, I'm not entirely sure that's the best for us, as our bump code is not the same as theirs, nor theirs is necessarily the best here.

DominikPanic: So what should I do

Rohesie: Leave this one doing nothing, just returning.

PR #1659, opened 2019-05-28
Implements skeleton crew access

DominikPanic:
:cl: LaKiller8
qol: Under a certain population level, access has been expanded for all jobs. Everyone gets dropship access, most non-marines get medbay and cargo access.
/:cl:


code/controllers/stonedmc/subsystem/job.dm
```diff
@@ -253,6 +261,9 @@ SUBSYSTEM_DEF(job)
 		job.radio_help_message(M)
 		if(job.req_admin_notify)
 			to_chat(M, "<span clas='danger'>You are playing a job that is important for game progression. If you have to disconnect, please head to hypersleep, if you can't make it there, notify the admins via adminhelp.</span>")
+		if(CONFIG_GET(number/minimal_access_threshold))
```
psykzz:
```suggestion
		if(CONFIG_GET(flag/jobs_have_minimal_access))
```

Why do we care about the config when its this flag that is whats important?

DominikPanic: Flag can be overriden at any time, this is the minimal number at which the flag gets set. this offers finer control for both admins and server operators.

PR #1669, opened 2019-05-29
Cleans up unused or shitty mob defines

DominikPanic:


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -883,17 +881,6 @@
 
 	return species.name
 
-/mob/living/carbon/human/proc/play_xylophone()
```
spookydonut: no, this is a blocker, this proc must remain.

psykzz: https://youtu.be/XVQOhIVcMj0

Rohesie: This is a good feature. It should stay, even though very spooky.

DominikPanic: Poe's law

Rohesie: Talking more seriously, keep it there. It's funny enough.

Rohesie:
Merge conflict. Other than that and the change request, the PR is fine.

code/modules/mob/living/carbon/human/human.dm
DominikPanic: It's not even usable and relies on non-existant dna code

Rohesie: It can be invoked by admins. /tg/ has skeletons.
This is a little gem, nothing too evil.

DominikPanic: Nothing uses it and you know our stance on that. So far it has been used exactly 0 times, procs that aren't used anywhere but are just for admins are very unlikely to be ever used.

PR #1670, opened 2019-05-29
Adds spec scaling

DominikPanic:
:cl: LaKiller8
balance: The amount of specialists is now scaled and depends on the amount of ready players.
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/game_options.dm
+++ b/code/controllers/configuration/entries/game_options.dm
@@ -181,3 +181,9 @@
 	integer = FALSE
 	min_val = 0
 	config_entry_value = 4
+
+
+/datum/config_entry/number/spec_coefficient
+	integer = FALSE
+	config_entry_value = 5
+	min_val = 0.001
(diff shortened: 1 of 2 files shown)


Mr-Worth:
Does the value in marines.dm not need to be touched? When I was trying this the total_positions value was what stumped me. Would this break the admin job slot control since it does not touch that?

DominikPanic:
> 
> 
> Does the value in marines.dm not need to be touched? When I was trying this the total_positions value was what stumped me. Would this break the admin job slot control since it does not touch that?

Technically not since assigning a job will fail if there's not a place in the squad. Though syncing them would prevent one check.

Rohesie:
Could you extend this to smartgunners?

Rohesie:
This will prioritize squads by alphabetic name. Quite an interesting change in the meta.

DominikPanic:
Should I randomize it?

Rohesie:
No real need. Predictable is good.

PR #1667, opened 2019-05-29
fuck dna

spookydonut:
fuck dna its been chopped and gutted so much it has no purpose and i want to see tg/goon genetics ported

PR #1671, opened 2019-05-29
Gets rid of the last legacy globals.

DominikPanic:
:cleanbot:


spookydonut:
big brain move duplicating my efforts with #1667

DominikPanic:
> 
> 
> big brain move duplicating my efforts with #1667

That was not my intention.

Issue #1661, opened 2019-05-29
timeofdeath = 0


Rohesie:
Shuttles going into the sunset fail to set `timeofdeath` properly, instead default it to zero. Minor issue, but this messes up a bit with respawn things.

The bug is on master.

PR #1674, opened 2019-05-29
Fixes zero timeofdeath

DominikPanic:
Fixes: #1661

### DIFF
--- a/code/modules/shuttle/shuttle.dm
+++ b/code/modules/shuttle/shuttle.dm
@@ -501,7 +501,8 @@
 			// Ghostize them and put them in nullspace stasis (for stat & possession checks)
 			L.notransform = TRUE
 			var/mob/dead/observer/O = L.ghostize(FALSE)
-			O.timeofdeath = world.time
+			if(O)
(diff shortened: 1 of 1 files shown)


code/modules/shuttle/shuttle.dm
```diff
@@ -497,14 +497,12 @@
...
-//			if(M.mind && !istype(t, /turf/open/floor/plasteel/shuttle/red) && !istype(t, /turf/open/floor/mineral/plastitanium/red/brig))
-//				M.mind.force_escaped = TRUE
+		for(var/mob/living/L in T.GetAllContents())
 			// Ghostize them and put them in nullspace stasis (for stat & possession checks)
-			M.notransform = TRUE
-			M.ghostize(FALSE)
-			M.moveToNullspace()
+			L.notransform = TRUE
+			var/mob/dead/observer/O = L.ghostize(FALSE)
+			O.timeofdeath = world.time
```
Rohesie: Do an `if(O)` because if they are living mobs without a player this will runtime.

PR #1678, opened 2019-05-29
Purges disease code.

DominikPanic:


code/_onclick/other_mobs.dm
```diff
@@ -39,12 +39,6 @@
 		ML.apply_damage(rand(1,3), BRUTE, dam_zone, armor)
 		for(var/mob/O in viewers(ML, null))
 			O.show_message("<span class='danger'>[name] has bit [ML]!</span>", 1)
-		if(armor >= 1) //Complete negation
```
spookydonut: intended?

francinum: It seems like a poorly laid out check for armor on disease spreading. Probably.

DominikPanic: Yeah the only thing below that is the disease spreading, so not removing this would keep an useless check

PR #1684, opened 2019-05-29
Adds a bad admin SQL proc

psykzz:
This will need a larger discussion before we merge im sure.

## About The Pull Request

Adds a really bad admin proc for querying raw SELECT statements against the database.

## Why It's Good For The Game

What good is data if you can't use it?

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -13,9 +13,9 @@
 		to_chat(usr, "<span class='warning'>SQL not connected?!.</span>")
 		return
 	
-	var/rawQuery = input(usr, "query?", "query?", "") as text|null
-	if(!dd_hasprefix(rawQuery, "SELECT"))
-		to_chat(usr, "<span class='warning'>SELECT Statements only?!.</span>")
+	var/rawQuery = input(usr, "What is your query", "Admin Query", "") as text|null
+	if(!findtext(rawQuery, regex("^select [\s*a-z,]+ from feedback(?:|\s+[a-z].*);$", "i")))
+		to_chat(usr, "<span class='warning'>SELECT Statements from Feedacb only!!.</span>")
 		return
 	var/datum/DBQuery/result = SSdbcore.NewQuery("[rawQuery]")
 	result.Execute()
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -1,3 +1,29 @@
+/datum/admins/proc/admin_sql()
+	set category = "Admin"
+	set name = "SQLQuery"
+	set desc = "Debug Query yo."
+
+	if(!check_rights(R_ADMIN))
+		to_chat(usr, "<span class='warning'>Not enough rights.</span>")
+		return
+	if(!CONFIG_GET(flag/sql_enabled))
+		to_chat(usr, "<span class='warning'>SQL disabled.</span>")
+		return
+	if(!SSdbcore.Connect())
+		to_chat(usr, "<span class='warning'>SQL not connected?!.</span>")
+		return
+	
+	var/rawQuery = input(usr, "query?", "query?", "") as text|null
+	if(!dd_hasprefix(rawQuery, "SELECT"))
+		to_chat(usr, "<span class='warning'>SELECT Statements only?!.</span>")
+		return
+	var/datum/DBQuery/result = SSdbcore.NewQuery("[rawQuery]")
+	result.Execute()
+	if(result.NextRow())
+		to_chat(usr, result.item[1])
+	qdel(result)
+
+
 /datum/admins/proc/admin_ghost()
 	set category = "Admin"
 	set name = "Aghost"
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -9,7 +9,7 @@
 	if(!CONFIG_GET(flag/sql_enabled))
 		to_chat(usr, "<span class='warning'>SQL disabled.</span>")
 		return
-	if(!SSdbcore.Connect())
+	if(!SSdbcore.IsConnected())
 		to_chat(usr, "<span class='warning'>SQL not connected?!.</span>")
 		return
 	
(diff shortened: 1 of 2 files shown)


psykzz:
@Rohesie @spookydonut @DominikPanic

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,29 @@
...
+	set name = "SQLQuery"
+	set desc = "Debug Query yo."
+
+	if(!check_rights(R_ADMIN))
+		to_chat(usr, "<span class='warning'>Not enough rights.</span>")
+		return
+	if(!CONFIG_GET(flag/sql_enabled))
+		to_chat(usr, "<span class='warning'>SQL disabled.</span>")
+		return
+	if(!SSdbcore.Connect())
```
DominikPanic: Don't use connect, use IsConnected

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,29 @@
+/datum/admins/proc/admin_sql()
+	set category = "Admin"
+	set name = "SQLQuery"
+	set desc = "Debug Query yo."
```
DominikPanic:
```suggestion
```

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,29 @@
...
+		return
+	if(!CONFIG_GET(flag/sql_enabled))
+		to_chat(usr, "<span class='warning'>SQL disabled.</span>")
+		return
+	if(!SSdbcore.Connect())
+		to_chat(usr, "<span class='warning'>SQL not connected?!.</span>")
+		return
+	
+	var/rawQuery = input(usr, "query?", "query?", "") as text|null
+	if(!dd_hasprefix(rawQuery, "SELECT"))
```
DominikPanic: Make it only work with the feedback table (which is not trivial, almost impossible) but I don't see it passing by Jordie/MSO otherwise.

psykzz:
The idea with this PR is not for merging but at least to figure our if we can query back data from the feedback table to make use of the SSblackbox for recording interesting gameplay stats.

Being able to track a weapons playrate, kill rate etc would be useful for balancing.

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,29 @@
...
+		to_chat(usr, "<span class='warning'>Not enough rights.</span>")
+		return
+	if(!CONFIG_GET(flag/sql_enabled))
+		to_chat(usr, "<span class='warning'>SQL disabled.</span>")
+		return
+	if(!SSdbcore.Connect())
+		to_chat(usr, "<span class='warning'>SQL not connected?!.</span>")
+		return
+	
+	var/rawQuery = input(usr, "query?", "query?", "") as text|null
```
DominikPanic: Give it some more user friendly options

DominikPanic:
You can test this locally.

code/modules/admin/admin_verbs.dm
psykzz: I see we have regex so ill make sure it captures it correctly.

psykzz: https://regex101.com/r/7Fa8qE/2/

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,28 @@
...
+		return
+	if(!CONFIG_GET(flag/sql_enabled))
+		to_chat(usr, "<span class='warning'>SQL disabled.</span>")
+		return
+	if(!SSdbcore.IsConnected())
+		to_chat(usr, "<span class='warning'>SQL not connected?!.</span>")
+		return
+	
+	var/rawQuery = input(usr, "What is your query", "Admin Query", "") as text|null
+	if(!findtext(rawQuery, regex("^select [\s*a-z,]+ from feedback(?:|\s+[a-z].*);$", "i")))
```
Jordie0608: It's essentially impossible to regex out user-input injections with regex, some examples with this regex:
`SELECT * from feedback WHERE 1=1; DROP TABLE feedback;`
`SELECT * from feedback;`
`SELECT * from feedback UNION ALL SELECT * FROM someothertable;`
`SELECT * from feedback WHERE 1=1 OR sleep(30000000000000000);`

https://regex101.com/r/sfsJGE/1

I'm not very good at sql injections so no doubt there'll be many more methods even with more restrictive parsing.

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,28 @@
...
+		to_chat(usr, "<span class='warning'>SELECT Statements from Feedacb only!!.</span>")
```
Jordie0608: Feedacb

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,28 @@
...
+		return
+	var/datum/DBQuery/result = SSdbcore.NewQuery("[rawQuery]")
+	result.Execute()
+	if(result.NextRow())
+		to_chat(usr, result.item[1])
```
Jordie0608: You allow for returning multiple columns but this only ever prints the first.

code/modules/admin/admin_verbs.dm
```diff
@@ -1,3 +1,28 @@
...
+	if(!SSdbcore.IsConnected())
+		to_chat(usr, "<span class='warning'>SQL not connected?!.</span>")
+		return
+	
+	var/rawQuery = input(usr, "What is your query", "Admin Query", "") as text|null
+	if(!findtext(rawQuery, regex("^select [\s*a-z,]+ from feedback(?:|\s+[a-z].*);$", "i")))
+		to_chat(usr, "<span class='warning'>SELECT Statements from Feedacb only!!.</span>")
+		return
+	var/datum/DBQuery/result = SSdbcore.NewQuery("[rawQuery]")
+	result.Execute()
```
Jordie0608: Should use warn_execute()

Jordie0608:
Even if only testmerged under specific circumstances I can't see this as being safe.

ninjanomnom:
This is an absolutely terrible idea and should never be done.

francinum:
This is a terrible idea, period.

PR #1693, opened 2019-05-30
Changes keybinds to use signals

### LINKED ISSUE #1647: Keybinds runtime on non-human mobs
This is on master code, due to #1602 . A Runner, for example, does not have `quick_equip()`:
```
[00:40:09] Runtime in human.dm, line 14: undefined proc or verb /mob/living/carbon/xenomorph/runner/quick equip(). 
proc name: down (/datum/keybinding/human/quick_equip/down)
usr: Psykzz/(Young Runner (53))
usr.loc: (Medical Upper (96, 181, 3))
src: quick_equip (/datum/keybinding/human/quick_equip)
call stack:
quick_equip (/datum/keybinding/human/quick_equip): down(Psykzz (/client))
Psykzz (/client): keyDown("E")
```

This could be solved by sending a signal and making humans (and whoever else we want to) register it, or by using an intermediate redefined proc, instead of the current:
```
/datum/keybinding/human/quick_equip/down(client/user)
    var/mob/living/carbon/human/H = user.mob
    H.quick_equip()
    return TRUE
```

Rohesie:
Fixes #1647 
As described in there, in the signal solution.

### DIFF
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -120,11 +120,17 @@
 #define COMSIG_XENOMORPH_FIRE_BURNING "xenomorph_fire_burning"
 #define COMSIG_XENOMORPH_TAKING_DAMAGE "xenomorph_taking_damage" // (target, damagetaken)
 
+//human signals
+#define COMSIG_CLICK_QUICKEQUIP "click_quickequip"
+
 // /obj/item/radio signals
 #define COMSIG_RADIO_NEW_FREQUENCY "radio_new_frequency"		//called from base of /obj/item/radio/proc/set_frequency(): (list/args)
 
+//keybindings
+
+#define COMSIG_KB_ACTIVATED (1<<0)
+
 // xeno abilities for keybindings
-#define COMSIG_XENOABILITY_HAS_ABILITY 1
 
 #define COMSIG_XENOABILITY_REGURGITATE "xenoability_regurgitate"
 #define COMSIG_XENOABILITY_DROP_WEEDS "xenoability_drop_weeds"
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -195,6 +195,10 @@
 #define COMSIG_XENOABILITY_FLING "xenoability_fling"
 #define COMSIG_XENOABILITY_PUNCH "xenoability_punch"
 
+// human signals for keybindings
+#define COMSIG_KB_QUICKEQUIP "keybinding_quickequip"
+#define COMSIG_KB_HOLSTER "keybinding_holster"
+#define COMSIG_KB_UNIQUEACTION "keybinding_uniqueaction"
 
 /*******Non-Signal Component Related Defines*******/
 
--- a/code/datums/keybinding/human.dm
+++ b/code/datums/keybinding/human.dm
@@ -3,7 +3,7 @@
 	weight = WEIGHT_MOB
 
 /datum/keybinding/human/down(client/user)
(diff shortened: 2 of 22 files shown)


code/datums/keybinding/human.dm
```diff
@@ -9,9 +9,9 @@
     full_name = "Quick equip"
     description = ""
 
+
 /datum/keybinding/human/quick_equip/down(client/user)
-    var/mob/living/carbon/human/H = user.mob
-    H.quick_equip()
+    SEND_SIGNAL(user.mob, COMSIG_KB_QUICKEQUIP)
```
spookydonut:
```suggestion
    if(SEND_SIGNAL(user.mob, COMSIG_KB_QUICKEQUIP) & comsig keybind handled or something) 
          Return true
```
Do the same for the others, only return on keybinds if it actually does the bind or you can't double up this key for xenos who don't need it

code/__DEFINES/components.dm
```diff
@@ -190,3 +190,8 @@
 #define COMSIG_XENOABILITY_LUNGE "xenoability_lunge"
 #define COMSIG_XENOABILITY_FLING "xenoability_fling"
 #define COMSIG_XENOABILITY_PUNCH "xenoability_punch"
+
+// human signals for keybindings
+#define COMSIG_KB_QUICKEQUIP "keybinding_quickequip"
+#define COMSIG_KB_HOLSTER "keybinding_holster"
+#define COMSIG_KB_UNIQUEACTION "keybinding_uniqueaction"
```
spookydonut:
```suggestion
#define COMSIG_KB_UNIQUEACTION "keybinding_uniqueaction"
	#define COMSIG_KB_ACTIVATED 1

```

code/datums/keybinding/human.dm
```diff
@@ -1,39 +1,30 @@
 /datum/keybinding/human
-    category = CATEGORY_HUMAN
-    weight = WEIGHT_MOB
+	category = CATEGORY_HUMAN
+	weight = WEIGHT_MOB
 
+/datum/keybinding/human/down(client/user)
+	return !isnull(SEND_SIGNAL(user.mob, keybind_signal)) //True or False is not important here. If there's a return, we've triggered a proc.
```
spookydonut:
```suggestion
	return (SEND_SIGNAL(user.mob, keybind_signal) & COMSIG_KB_ACTIVATED) 
```

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -51,6 +51,9 @@
 
 	randomize_appearance()
 
+	RegisterSignal(src, COMSIG_KB_QUICKEQUIP, .proc/quick_equip)
```
spookydonut: add wrapper procs for these and return COMSIG_KB_ACTIVATED

Rohesie:
Suggestion applied.

code/datums/keybinding/human.dm
```diff
@@ -1,39 +1,30 @@
 /datum/keybinding/human
-    category = CATEGORY_HUMAN
-    weight = WEIGHT_MOB
+	category = CATEGORY_HUMAN
+	weight = WEIGHT_MOB
 
+/datum/keybinding/human/down(client/user)
+	return CHECK_BITFIELD(SEND_SIGNAL(user.mob, keybind_signal), COMSIG_KB_ACTIVATED)
```
spookydonut: you should check that theres a valid keybind_signal value.

PR #1698, opened 2019-05-31
Tips Fix

Rohesie:
Fix for a small typo on the tooltips and micro-optimizes it.


Rohesie:
Is `user.?.client.tooltips` not a typo?

DominikPanic:
Not if you include your typical microoptimization :p

Rohesie:
Description changed. The behavior is essentially the same, there's just an added check for client, because clients are fickly things.

PR #1700, opened 2019-05-31
Cleans up preference code

DominikPanic:
Jobs are now saved as an assoc list, meaning we can support preferences for basically anything for the future.

Removes unnecessary job prefflags, department flags and flags (they were bad design in the first place since they were used as unique values rather than flags)

:cl: LaKiller8 & PsyKzz
qol: You can now right click in the job selection menu. Job preferences were reset, so don't forget to set them again!
qol: You can now use modifier keys for keybindings.
/:cl:

Fixes: #1725


code/modules/client/preferences.dm
```diff
@@ -8,9 +8,9 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	var/path
 	var/default_slot = 1
 	var/savefile_version = 0
+	var/current_tab = 0
```
Rohesie: This variable seems unused?

code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -118,32 +118,88 @@
...
-	if(!SSatoms.initialized)
-		return // clothing wont work properly
-		
-	// Set up the dummy for its photoshoot
-	var/mob/living/carbon/human/dummy/mannequin = generate_or_wait_for_human_dummy(DUMMY_HUMAN_SLOT_PREFERENCES)
-	copy_to(mannequin)
-
 	// Determine what job is marked as 'High' priority, and dress them up as such.
 	var/datum/job/previewJob
+	var/highest_pref = 0
```
Rohesie:
```suggestion
	var/highest_pref = JOB_PRIORITY_UNSET
```
And make a define up there along the others, for clarity?

code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -118,32 +118,88 @@
...
-	if(!SSatoms.initialized)
-		return // clothing wont work properly
-		
-	// Set up the dummy for its photoshoot
-	var/mob/living/carbon/human/dummy/mannequin = generate_or_wait_for_human_dummy(DUMMY_HUMAN_SLOT_PREFERENCES)
-	copy_to(mannequin)
-
 	// Determine what job is marked as 'High' priority, and dress them up as such.
 	var/datum/job/previewJob
+	var/highest_pref = JOBS_PRIORITY_NEVER
```
Rohesie: This is 4, not 0, and will break the logic.

DominikPanic: I'll just swap the defines since /tg/ uses it reversed from us

PR #1702, opened 2019-05-31
Ports supply pods from /tg/

DominikPanic:
This was pretty hard


Rohesie:
Yeeeeesssss

code/__HELPERS/icons.dm
```diff
@@ -1167,4 +1167,14 @@ world
...
 	else
-		return humanoid_icon_cache[icon_id]
\ No newline at end of file
+		return humanoid_icon_cache[icon_id]
+
+
+/proc/Get_Pixel_Angle(y, x)
+	if(!y)
+		return (x >= 0) ? 90 : 270
+	. = arctan(x / y)
```
spookydonut: beware byond trig functions have errors, see shuttle code for less buggy angle calc

DominikPanic: Which one should I use?

DominikPanic: Also this is just for a visual effect so any errors don't matter that much

PR #1704, opened 2019-06-01
Moves in a subsystem mistakenly forgotten during component port

ninjanomnom:
This is the subsystem for component management. If we end up wanting to make components process we could turn this into a processing subsystem. Otherwise it's mainly used for globally accessible signals like getting notified when someone dies or something.

Yeah we don't have any of those signals but here's the subsystem for that to match the define already there.

### DIFF
--- a/code/controllers/stonedmc/subsystem/dcs.dm
+++ b/code/controllers/stonedmc/subsystem/dcs.dm
@@ -1,0 +1,6 @@
+SUBSYSTEM_DEF(dcs)
+	name = "Datum Component System"
+	flags = SS_NO_INIT | SS_NO_FIRE
+
+/datum/controller/subsystem/dcs/Recover()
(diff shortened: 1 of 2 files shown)


code/controllers/stonedmc/subsystem/dcs.dm
```diff
@@ -0,0 +1,6 @@
+SUBSYSTEM_DEF(dcs)
```
DominikPanic: I disagree with cluttering the MC tab even more with no benefit. Either make the PR actually use this, or don't port it.

Rohesie:
Talked with  DominikPanic. He has his stance and we respect it, but the majority of us considers this to be a beneficial change.

I'd like to stress out here that in the past we've blocked things for not being fully ready and functional, like assets and other ports, but like now, when there's credible expectation of future use, we shouldn't be blocking these additions of extra tools and resources..

PR #1726, opened 2019-06-01
Start now will always work

psykzz:
## About The Pull Request

The start now button will no longer complain that your in the wrong state.

### DIFF
--- a/code/modules/admin/server_verbs.dm
+++ b/code/modules/admin/server_verbs.dm
@@ -215,7 +215,7 @@
 	if(!check_rights(R_SERVER))
 		return
 
-	if (SSticker.current_state != GAME_STATE_STARTUP && SSticker.current_state != GAME_STATE_PREGAME)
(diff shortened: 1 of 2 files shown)


code/modules/admin/server_verbs.dm
```diff
@@ -215,18 +215,26 @@
 	if(!check_rights(R_SERVER))
 		return
 
-	if(SSticker?.current_state != GAME_STATE_PREGAME)
-		to_chat(usr, "<span class='warning'>The round cannot be started early in this state.</span>")
+	if (!SSticker.current_state == GAME_STATE_STARTUP && !SSticker.current_state == GAME_STATE_PREGAME)
+		to_chat(usr, "<font color='red'>Error: Start Now: Game has already started.</font>")
```
DominikPanic:
```suggestion
		to_chat(usr, "<span class='warning'>The round has already started.</span>")
```
Don't make the user experience worse please.

PR #1729, opened 2019-06-02
Add support for a dynamic number of squads in SSdirection

psykzz:
## About The Pull Request

Add support for a dynamic number of squads in SSdirection

## Why It's Good For The Game

This is some initial work for working with an unlimited number of factions / squads.


psykzz:
PR is still a work in progress.

psykzz:
Good to go now, find out defining an arg as `proc(/datum/full/path, arg2, arg3)` no only fails silently but also blocks further args.

code/controllers/stonedmc/subsystem/direction.dm
```diff
@@ -4,8 +4,8 @@ SUBSYSTEM_DEF(direction)
 	runlevels = RUNLEVEL_GAME | RUNLEVEL_POSTGAME
 	wait = 1 SECONDS
 
-	// this is a map of defines to mob references, eg; list(CHARLIE_SL = <mob ref>, XENO_NORMAL_QUEEN = <mob ref>)
-	var/list/leader_mapping = list()
+	// this is a map of defines to mob references, eg; list(FACTION_ID = <mob ref>, FACTION_ID2 = <mob ref>)
+	var/list/leader_mapping[0]
```
spookydonut:
```suggestion
	var/list/leader_mapping = list()
```

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
```diff
@@ -35,6 +35,7 @@
 	for(var/upgrade in GLOB.xenoupgradetiers)
 		xenos_by_upgrade[upgrade] = list()
 
+
```
spookydonut: don't commit whitespace changes

PR #1731, opened 2019-06-02
Improves launch pod error handling

DominikPanic:
Better messages and safety.

### DIFF
--- a/code/modules/admin/verbs/podlauncher.dm
+++ b/code/modules/admin/verbs/podlauncher.dm
@@ -6,6 +6,9 @@
 		return
 
 	var/datum/podlauncher/P = new(usr)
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/podlauncher.dm
```diff
@@ -6,6 +6,9 @@
 		return
 
 	var/datum/podlauncher/P = new(usr)
+	if(!P.bay || !P.podarea)
+		return
```
Rohesie:
```suggestion
		to_chat(user, "No bay or podarea found.")
		return
```

DominikPanic: If you look below that's already handled and user is messaged what part is erroring out.

PR #1734, opened 2019-06-02
Sleeper Condense

Hughgent:
## About The Pull Request

You thought the sleeper window was done? think again buddy.

## Why It's Good For The Game

It's condensed even more.

## Changelog
:cl: Hughgent
tweak: Sleeper window is condensed even more.
/:cl:

image

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -82,7 +82,7 @@
 			dat += text("[]\t-Burn Severity %: []</FONT><BR>", (occupant.getFireLoss() < 60 ? "<font color='#487553'>" : "<font color='#b54646'>"), occupant.getFireLoss())
 			dat += text("<HR>Knocked Out Summary %: [] ([] seconds left!)<BR>", occupant.knocked_out, round(occupant.knocked_out / 4))
 			for(var/chemical in connected.available_chemicals)
-				dat += "<label style='width:180px; display: inline-block'>[connected.available_chemicals[chemical]] ([occupant.reagents.get_reagent_amount(chemical)] units)</label> Inject:"
+				dat += "<label style='width:180px; display: inline-block'>[connected.available_chemicals[chemical]] ([round(occupant.reagents.get_reagent_amount(chemical), 0.01)] units)</label> Inject:"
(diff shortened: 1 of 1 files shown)


Shadow-Quill:
This is the best PR.

code/game/machinery/Sleeper.dm
```diff
@@ -81,9 +81,11 @@
 			dat += text("[]\t-Toxin Content %: []</FONT><BR>", (occupant.getToxLoss() < 60 ? "<font color='#487553'>" : "<font color='#b54646'>"), occupant.getToxLoss())
 			dat += text("[]\t-Burn Severity %: []</FONT><BR>", (occupant.getFireLoss() < 60 ? "<font color='#487553'>" : "<font color='#b54646'>"), occupant.getFireLoss())
 			dat += text("<HR>Knocked Out Summary %: [] ([] seconds left!)<BR>", occupant.knocked_out, round(occupant.knocked_out / 4))
-			if(occupant.reagents)
-				for(var/chemical in connected.available_chemicals)
-					dat += "[connected.available_chemicals[chemical]]: [occupant.reagents.get_reagent_amount(chemical)] units<br>"
+			for(var/chemical in connected.available_chemicals)
+				dat += "<label style='width:180px; display: inline-block'>[connected.available_chemicals[chemical]] ([occupant.reagents.get_reagent_amount(chemical)] units)</label> Inject:"
```
Rohesie:
```suggestion
				dat += "<label style='width:180px; display: inline-block'>[connected.available_chemicals[chemical]] ([round(occupant.reagents.get_reagent_amount(chemical), 0.01)] units)</label> Inject:"
```
So we don't get 3.333333333333-like numbers.
Feel free to use a different multiplier, I thought that `3.33` was a suitable alternative.

Shadow-Quill: 3.3 might be better. I can't think of any reason to go further than that.

Rohesie: Just change that to `0.1` then.

Hughgent: Two significant digits is fine. it defaults to four, and the extra precision could be expected from a larger machine.

PR #1737, opened 2019-06-03
Gets rid of the last legacy proc definitions

DominikPanic:
Slightly updated Astar (tested with the bots, seems to work)
Unused SD alert.
Records moved to the proper file.
Also moves the legacy stylesheet to the proper location and updates it.


code/__HELPERS/astar.dm
```diff
@@ -0,0 +1,220 @@
...
+//Insert and place at its position a new node in the heap
+/datum/Heap/proc/Insert(atom/A)
+	L.Add(A)
+	Swim(length(L))
+
+
+//removes and returns the first element of the heap
+//(i.e the max or the min dependant on the comparison function)
+/datum/Heap/proc/Pop()
+	if(!L.len)
```
Rohesie:
```suggestion
	if(!length(L))
```

DominikPanic:
> 
> 
> Remove the Astar calls and just replace it with byond built-ins, Astar for games, especially byond is really slow/expensive and for the most part you don't need perfect pathing.

What calls should I use?

PR #1758, opened 2019-06-04
Adds more explosion sounds

MoXuT0:
## About The Pull Request

Per title. Small explosion sound effect now plays if devastation_range < 0.

## Changelog
:cl:
soundadd: Added more explosion sound effects.
/:cl:


Rohesie:
If you are changing the name of the explosion sound you need to replace every case.
For example, you have `if(prob(50))src << 'sound/effects/Explosion1.ogg'` in `Hallucination.dm`

MoXuT0:
Forgot about these, will check and replace if I'll find more.

Rohesie:
I was only that one and `else src << 'sound/effects/Explosion2.ogg'` in that same file.

DominikPanic:
Also use `SEND_SOUND` instead of `something << 'sound'`

MoXuT0:
Can I do that in separate pr? Cuz there's atleast 20 cases which need either `SEND_SOUND` or `to_chat`

code/modules/flufftext/Hallucination.dm
```diff
@@ -116,8 +116,8 @@ mob/living/carbon/proc/handle_hallucinations()
 				switch(rand(1,12))
 					if(1) src << 'sound/machines/airlock.ogg'
 					if(2)
-						if(prob(50))src << 'sound/effects/Explosion1.ogg'
-						else src << 'sound/effects/Explosion2.ogg'
+						if(prob(50))src << 'sound/effects/explosion1.ogg'
```
DominikPanic:
```suggestion
						if(prob(50))
						    SEND_SOUND(src, 'sound/effects/explosion1.ogg')
```

code/modules/flufftext/Hallucination.dm
```diff
@@ -116,8 +116,8 @@ mob/living/carbon/proc/handle_hallucinations()
...
+						else src << 'sound/effects/explosion2.ogg'
```
DominikPanic:
```suggestion
						else 
						    SEND_SOUND(src, 'sound/effects/explosion2.ogg')
```

PR #1762, opened 2019-06-05
Slight machinery code cleanup

DominikPanic:
Removes unused procs and defines, makes existing stuff use the better/standardized ones.
Should only improve performance.

### DIFF
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -9,7 +9,7 @@
 	layer = CONVEYOR_LAYER // so they appear under stuff
 	anchored = 1
 	var/operating = 0	// 1 if running forward, -1 if backwards, 0 if off
-	var/is_operational = 1	// true if can operate (no broken segments in this belt run)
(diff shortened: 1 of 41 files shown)


code/modules/recycling/conveyor2.dm
```diff
@@ -9,7 +9,7 @@
 	layer = CONVEYOR_LAYER // so they appear under stuff
 	anchored = 1
 	var/operating = 0	// 1 if running forward, -1 if backwards, 0 if off
-	var/operable = 1	// true if can operate (no broken segments in this belt run)
+	var/is_operational = 1	// true if can operate (no broken segments in this belt run)
```
Rohesie:
```suggestion
	var/is_operational = TRUE	// true if can operate (no broken segments in this belt run)
```

DominikPanic: Oof that's one I missed in the mass replacement. I'll likely not change this value, let me see.

Rohesie:
Merge conflict.

PR #1766, opened 2019-06-05
LOS Reduces screech effectiveness

psykzz:
## About The Pull Request

This adds an extra LOS check into the queen and reduces the effectiveness of screech for mobs not in LOS of the target.

## Why It's Good For The Game

This stops a queen running up to the side of a wall and screeching, with no counter play.

## Changelog
:cl:
balance: Queen screech is reduced by 30% when out of LOS.
/:cl:


code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
```diff
@@ -146,10 +146,15 @@
 	X.create_shriekwave() //Adds the visual effect. Wom wom wom
 	//stop_momentum(charge_dir) //Screech kills a charge
 
-	for(var/mob/living/L in range(world.view, X))
+	var/screech_range = world.view
```
DominikPanic: If you're using this just in the proc then there's zero point to having this instead of world.view. It doesn't allow customization and only further obfuscates the code.

Rohesie:
With the changes to xeno vision, `view()` may no longer reflect what it used to.
Test if this gets actually triggered.

psykzz:
Fixed the issue @Rohesie and i also updated the loop a little.

Did some profiling with `range` vs `get_dist`.
Tested 1000 loops on around 134 mobs.

```
                                               Profile results (total time)
Proc Name                                                                  Self CPU    Total CPU    Real Time     Overtime        Calls
----------------------------------------------------------------------    ---------    ---------    ---------    ---------    ---------
/proc/screech_2                                                               0.197       36.535       36.555        0.161         1000
/proc/screech_1                                                               0.312       37.100       37.086        0.005         1000
```

PR #1768, opened 2019-06-05
A greater mob define cleanup.

DominikPanic:
Ninja said that refactoring mob code brings out the madness in men. He was right. But I won't give up.


Rohesie:
From what I gather you are moving variables around and then fixing what breaks when that happens.
Could you list the variables changed, and if you are many any different changes?

DominikPanic:
> 
> 
> From what I gather you are moving variables around and then fixing what breaks when that happens.
> Could you list the variables changed, and if you are many any different changes?

Everything that had a comment like living/carbon has been moved appropriately. /tg/ has these in the proper file. There are not many different changes, this is just cleaning up old bay shitcode.

PR #1771, opened 2019-06-06
Updates the contributing standards.

DominikPanic:
Removes a lot of /tg/ references
Adds some already enforced ideas (argument naming, return value preservation, src)
Improves the language to be more friendly
Makes a lot of the things better reflect how it's already being done so people have an actual valid point of reference.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -238,7 +238,6 @@ There's no indication of what "1" and "2" mean! Instead, you'd do something like
 			(...)
 ````
 This is clearer and enhances readability of your code! Get used to doing it!
-On another note, if the define is only used in one place only with a clearly named variable (var/cost), then adding a define would further obfuscate it, use the regular data types in those cases.
 
 ### Control statements
 (if, while, for, etc)
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,45 +2,47 @@
 
 ## Reporting Issues
 
-See [this page](http://tgstation13.org/wiki/Reporting_Issues) for a guide and format to issue reports.
+See [this page](https://github.com/tgstation/TerraGov-Marine-Corps/issues/new?template=bug_report.md) for a guide and format to issue reports.
 
 ## Introduction
 
-Hello and welcome to /tg/station's TGMC contributing page. You are here because you are curious or interested in contributing - thank you! Everyone is free to contribute to this project as long as they follow the simple guidelines and specifications below; at /tg/station, we strive to maintain code stability and maintainability, and to do that, we need all pull requests to hold up to those specifications. It's in everyone's best interests - including yours! - if the same bug doesn't have to be fixed twice because of duplicated code.
+Hello and welcome to the TGMC contributing page. You are here because you are curious or interested in contributing - thank you! Everyone is free to contribute to this project as long as they follow the simple guidelines and specifications below. We strive to maintain long-term code stability and maintainability, and to do that, we need all pull requests to hold up to those specifications. It's in everyone's best interests - including yours! - if the same bug doesn't have to be fixed twice because of duplicated code.
 
-First things first, we want to make it clear how you can contribute (if you've never contributed before), as well as the kinds of powers the team has over your additions, to avoid any unpleasant surprises if your pull request is closed for a reason you didn't foresee.
+First things first, we want to make it clear how you can contribute (if you've never contributed before), as well as what the structure of the team and the processes are, to avoid any unpleasant surprises if your pull request is closed for a reason you didn't foresee.
 
 ## Getting Started
 
-/tg/station doesn't have a list of goals and features to add; we instead allow freedom for contributors to suggest and create their ideas for the game. That doesn't mean we aren't determined to squash bugs, which unfortunately pop up a lot due to the deep complexity of the game. Here are some useful starting guides, if you want to contribute or if you want to know what challenges you can tackle with zero knowledge about the game's code structure.
+We don't have a strict list of goals and features to add; we instead allow freedom for contributors to suggest and create their ideas for the game. That doesn't mean we aren't determined to squash bugs, which unfortunately pop up a lot due to the deep complexity of the game. Here are some useful starting guides, if you want to contribute or if you want to know what challenges you can tackle with zero knowledge about the game's code structure.
 
-If you want to contribute the first thing you'll need to do is [set up Git](http://tgstation13.org/wiki/Setting_up_git) so you can download the source code.
+If you'd still like some guidance to see which features are appreciated, check the thread [here](https://tgstation13.org/phpBB/viewtopic.php?f=65&t=20487).
 
-We have a [list of guides on the wiki](http://www.tgstation13.org/wiki/index.php/Guides#Development_and_Contribution_Guides) that will help you get started contributing to /tg/station with Git and Dream Maker. For beginners, it is recommended you work on small projects like bugfixes at first. If you need help learning to program in BYOND, check out this [repository of resources](http://www.byond.com/developer/articles/resources).
+If you want to contribute the first thing you'll need to do is set up Git and clone the repository. Check out [this](https://tgstation13.org/wiki/TGMC:Guide_to_contributing) helpful guide.
 
-You can of course, as always, ask for help at TGMC coding channel on /tg/station official discord (https://discord.tgstation13.org/). We're just here to have fun and help out, so please don't expect professional support.
+We have a [list of guides on the wiki](http://www.tgstation13.org/wiki/index.php/Guides#Development_and_Contribution_Guides) that will help you get started contributing to this codebase with Git and Dream Maker. For beginners, it is recommended you work on small projects like bugfixes at first. If you need help learning to program in BYOND, check out this [repository of resources](http://www.byond.com/developer/articles/resources).
+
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -2,45 +2,47 @@
...
+The Art Director controls sprites and aesthetic that get into the game. While sprites for brand-new additions are generally accepted without harsh standards, modified current art assets fall to the Art Director, who can decide whether or not a sprite tweak is both merited and a suitable replacement.
 
 They also control the general "perspective" of the game - how sprites should generally look, especially the angle from which they're viewed. An example of this is the [3/4 perspective](http://static.tvtropes.org/pmwiki/pub/images/kakarikovillage.gif), which is a bird's eye view from above the object being viewed.
 
 **Maintainers**
 
-Maintainers are quality control. If a proposed pull request doesn't meet the following specifications, they can request you to change it, or simply just close the pull request. Maintainers are required to give a reason for closing the pull request.
+Maintainers are the volunteer quality control. If a proposed pull request doesn't meet the following specifications, they can request a change, and if a proper reason is not provided or the request becomes stale, they may close it. Maintainers are required to give a reason for closing the pull request.
```
ninjanomnom: You don't really need t specify volunteer here as everyone is volunteer. This also kinda implies headcoder and design lead aren't?

.github/CONTRIBUTING.md
```diff
@@ -99,9 +101,30 @@ The previous code made compliant:
...
+```
+/datum/proc/proc1()
 	code
+	return something
+
+/datum/datum1/proc1()
+	. = ..()
 ```
+`.` is the current return value of the current proc.
+If you have the parent call at the end of the proc, use `return ..()` instead, it's slightly faster.
```
ninjanomnom: The speed isnt really why we do it, it's mostly just been maintainer preference in the past because they don't like the look of `. = ..()` as the last line

.github/CONTRIBUTING.md
```diff
@@ -202,6 +238,7 @@ There's no indication of what "1" and "2" mean! Instead, you'd do something like
 			(...)
 ````
 This is clearer and enhances readability of your code! Get used to doing it!
+On another note, if the define is only used in one place only with a clearly named variable (var/cost), then adding a define would further obfuscate it, use the regular data types in those cases.
```
ninjanomnom: What does this mean?

.github/CONTRIBUTING.md
```diff
@@ -242,31 +279,28 @@ This prevents nesting levels from getting deeper then they need to be.
 
 * Information that players could use to metagame (that is, to identify round information and/or antagonist type via information that would not be available to them in character) should be kept as administrator only.
 
-* It is recommended as well you do not expose information about the players - even something as simple as the number of people who have readied up at the start of the round can and has been used to try to identify the round type.
-
-* Where you have code that can cause large-scale modification and *FUN*, make sure you start it out locked behind one of the default admin roles - use common sense to determine which role fits the level of damage a function could do.
+* Where you have code that can cause large-scale modification and *FUN*, make sure you start it out locked behind one of the proper admin permissions - ask if you're not sure.
```
ninjanomnom: Why are you taking out the previous statement?

.github/CONTRIBUTING.md
```diff
@@ -309,6 +343,7 @@ Math operators like +, -, /, *, etc are up in the air, just choose which version
 #### Use
 * Bitwise AND - '&'
 	* Should be written as ```bitfield & bitflag``` NEVER ```bitflag & bitfield```, both are valid, but the latter is confusing and nonstandard.
+	* Feel free to use one of the helpful bitflag operation macros instead (CHECK_BITFIELD and similar).
```
ninjanomnom: I still really hate these things

.github/CONTRIBUTING.md
DominikPanic: Because the current code is already breaking that. Who will be playing or observing is public information.

DominikPanic: Though I agree, I should include something that mentions just IPs and CIDs.

ninjanomnom: sure but it's still bad to add more

DominikPanic: For private information, I agree, for the ready players and such, it's heavily context dependent for our server so putting it here wouldn't really be wise.

.github/CONTRIBUTING.md
spookydonut: it's talking about xeno abilities where instead of the cooldown var being defined as 12 SECONDS it's set to ABILITY_NAME_COOLDOWN which is a define for 12 SECONDS

.github/CONTRIBUTING.md
```diff
@@ -150,14 +186,14 @@ There are two key points here:
...
 While we normally encourage (and in some cases, even require) bringing out of date code up to date when you make unrelated changes near the out of date code, that is not the case for `New` -> `Initialize` conversions. These systems are generally more dependant on parent and children procs so unrelated random conversions of existing things can cause bugs that take months to figure out.
 
 ### `Initialize()` must be used properly
-`Initialize()` must always call the parent proc `..()` to ensure the atom initializes and must return a valid value either by using `. = ..()` or `return ..()` or `return INITIALIZE_HINT_*`
+`Initialize()` must always call the parent proc and preserve the return value or return a valid hint like `return INITIALIZE_HINT_*`
```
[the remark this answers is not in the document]
DominikPanic: Preserving the return value is talked a bit above this, this is just using the terms defined there.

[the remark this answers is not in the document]
ninjanomnom: I feel like taking out the reminder of what a parent call is reduces the usefulness of this to a first time contributor

.github/CONTRIBUTING.md
ninjanomnom: It's not though, this is an anti powergamer standard. It makes sense to have this in the contribution guidelines still

.github/CONTRIBUTING.md
ninjanomnom: I've done that before when I want a defines file to act as a source for balance regarding a feature. It makes sense to do that depending on context

.github/CONTRIBUTING.md
DominikPanic: It does not for our server.

.github/CONTRIBUTING.md
```diff
@@ -238,35 +274,34 @@ This prevents nesting levels from getting deeper then they need to be.
 
 * Calls to the database must be escaped properly - use sanitizeSQL to escape text based database entries from players or admins, and isnum() for number based database entries from players or admins.
 
+* It is recommended as well you do not expose information about the players - even something as simple as the number of people who have readied up at the start of the round can and has been used to try to identify the round type.
```
ninjanomnom: You still probably should change the example to something applicable as you brought up. Keeping information hidden like which player is playing in what role etc.

Also you put this back in the wrong place

.github/CONTRIBUTING.md
spookydonut: yeah its not something you can make a blanket rule for its a case by case thing

Rohesie:
How about maps being 255x255?

.github/CONTRIBUTING.md
Rohesie: At least we are not using this another codebase does: `#define LAZYLEN(L) length(L)`
:p

.github/CONTRIBUTING.md
```diff
@@ -244,29 +280,29 @@ This prevents nesting levels from getting deeper then they need to be.
...
+* Please attempt to clean out any dirty variables that may be contained within items you alter through var-editing. For example, due to how DM functions, changing the `pixel_x` variable from 23 to 0 will leave a dirty record in the map's code of `pixel_x = 0`. Likewise this can happen when changing an item's icon to something else and then back. This can lead to some issues where an item's icon has changed within the code, but becomes broken on the map due to it still attempting to use the old entry.
+* Areas should not be var-edited on a map to change it's name or attributes. All areas of a single type and it's altered instances are considered the same area within the code, and editing their variables on a map can lead to issues with powernets and subsystems which are difficult to debug.
+* All maps should be 255x255 at most.
```
spookydonut:
```suggestion
* All maps should be 255x255 at most because of known issues with how BYOND behaves with maps bigger than this.
```

Alterist:
Should mention something about maps being in the .tgm format, unless I totally.missed that

.github/CONTRIBUTING.md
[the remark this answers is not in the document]
DominikPanic: I've readded it.

.github/CONTRIBUTING.md
DominikPanic: Removed.

.github/CONTRIBUTING.md
DominikPanic: Should I remove it or keep it?

DominikPanic:
> 
> 
> Should mention something about maps being in the .tgm format, unless I totally missed that

Give me a the exact wording for what you want added about maps and I'll do it.

.github/CONTRIBUTING.md
ninjanomnom: There arent actualy any issues with larger maps. It's just all your maps have to be the same size and 255x255 is what's used

Alterist:
`All new maps should be converted into the .tgm format before it becomes a pull request. All edits to existing maps should be correctly map merged. See: https://tgstation13.org/wiki/Map_Merger on how to do this `

Issue #994, opened 2019-03-06
You can prevent the queen from locking down a dropship before it's too late and it can be sent back to be blocked again.


MiniMeatwad:
## Reproduction:
Place a rocket frame and ripley blocking the shuttle console and lockdown button and you can prevent the xenos from using the dropship, call it back, kill any xenos on board and block it again if possible. Can be abused to stall a round for a long time. This only works if the queen doesn't know how to access the lockdown button or shuttle console, which is to climb over the rocket frame and break the ripley, which sometimes can only be destroyed by aciding it. As said by someone else, it's like a gate for xeno queens that don't know how to climb over a rocket frame, it's very obscure and unknown as even being an option, and doesn't really have a reason to exist other than for veterans to trick newer players into dying.

DominikPanic:
This can't be feasibly fixed tbh

InterroLouis:
Eh, I mean, you could prevent people from depositing things on the floor in the dropship using a Ripley if you really wanted to. Probably, somehow. Through the magics of coding, knowledge of which is byond me.

DominikPanic:
> 
> 
> Eh, I mean, you could prevent people from depositing things on the floor in the dropship using a Ripley if you really wanted to. Probably, somehow. Through the magics of coding, knowledge of which is byond me.

Snowflake code not possible to do properly until components

DominikPanic:
It's possible to fix properly now I suppose, though has it really been an issue?

Rohesie:
This is more organically fixed by making those things destroyable by xeno melee attacks.

DominikPanic:
Very goo idea, I'll get onto that.

PR #1773, opened 2019-06-06
Makes dropship-placeable structures destructible by xenos

DominikPanic:
Should prevent the blocking.
Fixes: #994


Pariah919:
can't you use razor wire and a turret to also accomplish blocking? Also you can climb on these anyway.

DominikPanic:
> 
> 
> can't you use razor wire and a turret to also accomplish blocking? Also you can climb on these anyway.

Razorwires and turrets are slashable. These are not.

PR #1774, opened 2019-06-06
Makes machines use START_PROCESSING instead of our snowflake variant

DominikPanic:
The way /tg/ does it is that in Init it starts processing, but most machines have the default /process that returns PROCESS_KILL. This is a more elegant way to do it, and will unshitcode our codebase in the long term.

Here I'm either:
making New into Init for machines
Removing start_processing for init in machines since it's included at the base
Changing the calls from start_stop to START_STOP
If it's in process I'm returning the PROCESS_KILL

### DIFF
--- a/code/game/machinery/OpTable.dm
+++ b/code/game/machinery/OpTable.dm
@@ -145,7 +145,7 @@
 			icon_state = M.pulse ? "table2-active" : "table2-idle"
 			return 1
 	victim = null
-	stop_processing()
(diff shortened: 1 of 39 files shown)


Jalleo:
These PRs are a priority my tool_act PR can work around the conflicts.

spookydonut:
We went over this before, don't start them processing unless they need to process, that's shitcode.

PR #1778, opened 2019-06-06
Mass-replaces every density = 1 / 1.0 into TRUE

DominikPanic:
Same as my previous PR, don't give it priority, let me solve all the possible conflicts.

Edit: The rationale behind this is that new code should always be using TRUE, which means this is unlikely to conflict with what people are doing at the moment in a larger capacity. It's not just conflict bait, I think that this will be beneficial.


spookydonut:
Good atomisations, do this more often please

PR #1767, opened 2019-06-05
Refactor gamemode code

psykzz:
## About The Pull Request

Refactor gamemode code

## Why It's Good For The Game

The idea here being to make it easier to add new gamemodes.

PR #1781, opened 2019-06-07
Revert "Refactor gamemode code"

spookydonut:
Reverts tgstation/TerraGov-Marine-Corps#1767

There is an error that will cause a runtime related to jobbans, see original PR.


Rohesie:
Ideally we patch the errors instead of reverting it.

DominikPanic:
Fixed it in #1785

PR #1782, opened 2019-06-07
Ports /tg/ telekinesis

Rohesie:
Also adds `do_attack_animation()`, which is the usual attack animation in /tg/ when someone is hit by an object or similar.

It works and can be added to mobs by registering the signal and flipping the flag, but the reason to port this is because I want the `tk_grab` for the Shrike more than because of anything else.


code/game/atoms_movable.dm
```diff
@@ -469,6 +469,67 @@
 	return
 
 
+/atom/movable/proc/do_attack_animation(atom/A, visual_effect_icon, obj/item/used_item, no_effect)
```
psykzz: We already had `Mob.animation_attack_on(atom/A)`
Wont this conflict?

DominikPanic: It won't conflict, our code should be refactored to use this in the long term though.

code/__DEFINES/living_flags.dm
```diff
@@ -0,0 +1 @@
+#define LIVING_TK_USER (1<<0)
```
DominikPanic: Not really happy with define organization since there's already the mobs one but not sure how to feel about it.

Rohesie: Up to you. I'd rather keep the defines each on their own file, but can adapt at will.

Rohesie:
Conflict solved.

Issue #1787, opened 2019-06-07
Heavy armor item icon layers underneath outer clothing slot


ArcaneCarrot:
## Testmerges:
#1694: 'm39 on back' by Rohesie at commit 275aa8c733
#1757: 'Adds new firing sounds for some weapons' by MoXuT0 at commit 3e26b4e17c
#1758: 'Adds more explosion sounds' by MoXuT0 at commit 67c096955e
#1766: 'LOS Reduces screech effectiveness' by psykzz at commit 3f86655d11
#1785: 'Fixes a tiny typo that would break jobs.' by DominikPanic at commit 8524b1fee1
#1770: 'Adds imaginary friends for the mentors.' by DominikPanic at commit 9b229a079a

## Reproduction:
Can't click this shit, boss.
![Dunga](https://user-images.githubusercontent.com/32185313/59126034-e377a400-8918-11e9-83d5-65805ad79746.PNG)

PR #1789, opened 2019-06-07
Fixes layers of items related to actions

DominikPanic:
Fixes: #1787 
Otherwise this happens:
https://imgur.com/a/a9fTjAh

### DIFF
--- a/code/datums/action.dm
+++ b/code/datums/action.dm
@@ -95,11 +95,13 @@
 /datum/action/item_action/update_button_icon()
 	button.overlays.Cut()
 	var/obj/item/I = target
+	var/old_layer = I.layer
+	var/old_plane = I.plane
 	I.layer = ABOVE_HUD_LAYER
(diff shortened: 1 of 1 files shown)


code/datums/action.dm
```diff
@@ -95,11 +95,13 @@
...
 	button.overlays.Cut()
 	var/obj/item/I = target
+	var/old_layer = I.layer
+	var/old_plane = I.plane
 	I.layer = ABOVE_HUD_LAYER
 	I.plane = ABOVE_HUD_PLANE
 	button.overlays += I
-	I.layer = initial(I.layer)
-	I.plane = initial(I.plane)
+	I.layer = old_layer
```
spookydonut: wouldnt this be fixed by making an /image instead

DominikPanic: Could be, though that could be more performance intensive than this.

Rohesie: Yea, this is definitely cheaper.

PR #1790, opened 2019-06-07
Fixes latejoining

DominikPanic:
Since both browsers used the same ID, the default one was replaced since the if check didn't pass because there was no truthy value being returned.

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -490,7 +490,7 @@
 
 
 /datum/game_mode/proc/AttemptLateSpawn(mob/M, rank)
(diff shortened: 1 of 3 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -486,8 +486,11 @@
...
 	popup.open(FALSE)
 
+	return TRUE
+
+
 /datum/game_mode/proc/AttemptLateSpawn(mob/M, rank)
-	if(src != usr || !isnewplayer(M))
+	if(!isnewplayer(M))
 		return
 	var/mob/new_player/NP
```
psykzz:
```suggestion
	var/mob/new_player/NP = M
```

PR #1791, opened 2019-06-07
Fixes being unable to interact with camera computers

DominikPanic:
We don't implement the machine var yet, the cameras already check if the user is able to view them so the check is not needed.

Also removes the unnecessary cancel option since it's one of the two buttons.


Rohesie:
Conflict.

Issue #1631, opened 2019-05-25
Teleport runtime


Rohesie:
## Testmerges:

Round ID: 513
BYOND Version: 512.1464
Server revision compiled on: 
Master commit: 57045e70970150ddc33564fe427b05aa20de79c6
The following pull requests are currently test merged:
#1600: 'Add ghost hud & alerts from TG' by psykzz at commit 617471aaa4
#1596: 'Inspection on the little smoke runtime.' by Ghommie at commit 31870a2326
#1623: 'Balance tweaks.' by Rohesie at commit bc93d71d05
#1628: 'Warrior lunges grab humans in their offhand' by psykzz at commit e4b5114079
#1627: 'fixes weeds' by spookydonut at commit 1e4416a9be

## Reproduction:

Verb used: Ghost > Teleport.

```
[07:11:32] Runtime in unsorted.dm, line 927: Cannot read null.z
proc name: params2turf (/proc/params2turf)
usr: Trevorious/(Ellie 'Coffee' Weaver)
src: null
call stack:
params2turf("15:26,9:9", null, Trevorious (/client))
(/obj/screen/click_catcher): Click(null, "mapwindow.map", "icon-x=474;icon-y=265;left=1;s...")
Trevorious (/client): Click( (/obj/screen/click_catcher), null, "mapwindow.map", "icon-x=474;icon-y=265;left=1;s...")
```

psykzz:
I don't believe this is the right runtime, this seems to happen when clicking aftering the teleport to null space.

DominikPanic:
This is a race condition click runtime.

DominikPanic:
Haven't seen this happen since. Will keep this open meanwhile. Possible it could have been fixed as a side effect in anothre PR.

PR #1804, opened 2019-06-07
Cleans up some dirty perspective modifications

DominikPanic:
Fixes: #1631 
Fixes: #1824 
Uses proper code now, things should be using reset_perspective more.

As a side effect, the observe verb for ghosts is quite nice now.


Rohesie:
It has an already-reported issue of the ghost returning back to the original position instead of moving to the observed person.
Putting it here more to remember it myself and not merge it early.

DominikPanic:
This is likely not the cause of it as I can replicate it on master, but it shouldn't get merged until it's fixed regarldess.

DominikPanic:
@Rohesie ready again wink wink

PR #1807, opened 2019-06-07
Adds a varset_callback to reduce the proc amount

DominikPanic:
Quite a few callbacks are just setting one variable to true/false, this will make further unshitcoding and porting useful.

### DIFF
--- a/code/__DEFINES/callbacks.dm
+++ b/code/__DEFINES/callbacks.dm
@@ -2,3 +2,6 @@
 
 #define CALLBACK new /datum/callback
 #define INVOKE_ASYNC world.ImmediateInvokeAsync
+
+#define VARSET_LIST_CALLBACK(target, var_name, var_value) CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, ##target, ##var_name, ##var_value)
+#define VARSET_CALLBACK(datum, var, var_value) CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, ##datum, NAMEOF(##datum, ##var), ##var_value)
--- a/code/datums/callback.dm
+++ b/code/datums/callback.dm
@@ -178,3 +178,13 @@
 		sleep(resolution*world.tick_lag)
 	return CS.finished
 
+
+/proc/___callbackvarset(list_or_datum, var_name, var_value)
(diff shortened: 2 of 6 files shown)


psykzz:
Do you have an example of how to use `VARSET_LIST_CALLBACK`?

DominikPanic:
> 
> 
> Do you have an example of how to use `VARSET_LIST_CALLBACK`?

No, /tg/ doesn't use it either as far as I can tell, but it's useful to have regardless.

code/__DEFINES/callbacks.dm
```diff
@@ -2,3 +2,6 @@
 
 #define CALLBACK new /datum/callback
 #define INVOKE_ASYNC world.ImmediateInvokeAsync
+
+#define VARSET_LIST_CALLBACK(target, var_name, var_value) CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, ##target, ##var_name, ##var_value)
+#define VARSET_CALLBACK(datum, var, var_value) CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, ##datum, NAMEOF(##datum, ##var), ##var_value)
```
Rohesie:
```suggestion
#define VARSET_CALLBACK(datum, var, var_value) (CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, (##datum), (NAMEOF(##datum), (##var)), (##var_value)))
```
Make use and abuse of parenthesis on macros.

Rohesie:
This is very, very nice.

code/__DEFINES/callbacks.dm
```diff
@@ -2,3 +2,6 @@
 
 #define CALLBACK new /datum/callback
 #define INVOKE_ASYNC world.ImmediateInvokeAsync
+
+#define VARSET_LIST_CALLBACK(target, var_name, var_value) CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, ##target, ##var_name, ##var_value)
```
Rohesie:
```suggestion
#define VARSET_LIST_CALLBACK(target, var_name, var_value) (CALLBACK(GLOBAL_PROC, /proc/___callbackvarset, (##target), (##var_name), (##var_value)))
```

PR #1808, opened 2019-06-08
Combines the rank/equipment/squad admin verbs, slight PP tweaks

DominikPanic:
More admin QoL and niceness.

### DIFF
--- a/code/modules/admin/fun_verbs.dm
+++ b/code/modules/admin/fun_verbs.dm
@@ -639,14 +639,14 @@
 	message_admins("[ADMIN_TPMONTY(usr)] changed the security level to code [sec_level].")
 
 
-/datum/admins/proc/edit_rank(mob/living/carbon/human/H in GLOB.human_mob_list)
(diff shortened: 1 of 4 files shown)


code/modules/admin/fun_verbs.dm
```diff
@@ -639,14 +639,14 @@
...
+/datum/admins/proc/rank_and_equipment(mob/living/carbon/human/H in GLOB.human_mob_list)
 	set category = "Fun"
-	set name = "Edit Rank"
+	set name = "Rank and Equipment"
 
 	if(!check_rights(R_FUN))
 		return
 
-	var/dat
+	var/dat = "<br>"
```
psykzz: Does this fix the issue with long named mobs that wrap to the next line?
I found that it covers the edit rank button in those cases.

DominikPanic: Yes, that's exactly what this does :)

PR #1810, opened 2019-06-08
Fixes an admin runtime.

DominikPanic:
Avoided the ?. operator since it's comparison.
```
[06:59:19] Runtime in admin_verbs.dm, line 953: Cannot read null.tier
proc name: ticket reply (/client/proc/ticket_reply)
usr: Psykzz/(Matt Dowl)
usr.loc: (Entrance Hallway (221, 126, 2))
src: Psykzz (/client)
call stack:
Psykzz (/client): ticket reply(LaKiller8 (/client))
Some edginess: (F) \[Delta (Sp... (/datum/admin_help): Action("reply")
psykzz\'s admin datum (Admin O... (/datum/admins): Topic("_src_=holder;admin_token=82962...", /list (/list))
Psykzz (/client): Topic("_src_=holder;admin_token=82962...", /list (/list), psykzz\'s admin datum (Admin O... (/datum/admins))
Psykzz (/client): Topic("_src_=holder;admin_token=82962...", /list (/list), psykzz\'s admin datum (Admin O... (/datum/admins))
Psykzz (/client): Topic("_src_=holder;admin_token=82962...", /list (/list))
```

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -950,7 +950,7 @@
 
 	var/datum/admin_help/AH = C.current_ticket
 
-	if(AH.tier == TICKET_ADMIN && !check_rights(R_ADMINTICKET, FALSE))
+	if(AH && AH.tier == TICKET_ADMIN && !check_rights(R_ADMINTICKET, FALSE))
 		return
 
 	if(AH && !AH.marked)


code/modules/admin/admin_verbs.dm
```diff
@@ -950,7 +950,7 @@
 
 	var/datum/admin_help/AH = C.current_ticket
 
-	if(AH.tier == TICKET_ADMIN && !check_rights(R_ADMINTICKET, FALSE))
+	if(AH && AH.tier == TICKET_ADMIN && !check_rights(R_ADMINTICKET, FALSE))
```
spookydonut:
```suggestion
	if(!AH || (AH.tier == TICKET_ADMIN && !check_rights(R_ADMINTICKET, FALSE)))
```
and remove the check of AH below
Probably also add logging if a non-admin is trying to reply to an admin ticket.

spookydonut: Or change it to deal with just normal messaging them if theres no ticket first as an early return

DominikPanic: That would break the logic. You want to private_message the person in the end no matter if there was an adminhelp or not.

spookydonut: thats what i'm saying, if theres no adminhelp just call private_message in an early return then deal with the case where it is an admin help with more simplified logic

DominikPanic: The message is not yet defined at this point, changing the logic here would also break the marking behavior. Is it the most optimized code? No, but this allows for a lot of admin QoL and 3 more simple checks won't hurt.

spookydonut: Ah okay I see now, yeah would require more work

PR #1811, opened 2019-06-08
Ensure we save keybindings after all changes

psykzz:
## About The Pull Request

Ensure we save keybinding changes after all changes


code/modules/client/preferences.dm
```diff
@@ -940,6 +940,8 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 					key_bindings[old_key] -= kb_name
 					key_bindings["Unbound"] += list(kb_name)
 				user << browse(null, "window=capturekeypress")
+				save_preferences()
+				save_character()
```
DominikPanic: Keybindings are global, thus this shouldn't be required

DominikPanic: (the save_character part of all your additions)

PR #1812, opened 2019-06-08
Fixes airlock welding

DominikPanic:
Also makes it /tg/code for the future, win-win
Fixes: #1806


psykzz:
What was the actual bug that caused the welder not to work?

DominikPanic:
> 
> 
> What was the actual bug that caused the welder not to work?

Heck if I know, likely me fucking up some check in the attackby PR

PR #1815, opened 2019-06-08
Makes the pod launcher logging better.

DominikPanic:


code/modules/admin/verbs/podlauncher.dm
```diff
@@ -713,25 +713,27 @@
 		qdel(M)
 
 
-/datum/podlauncher/proc/supplypod_punish_log(list/whoDyin)
+/datum/podlauncher/proc/supplypod_log(list/targets, atom/target)
 	var/podString = effectBurst ? "5 pods" : "a pod"
 	var/whomString = ""
-	if(LAZYLEN(whoDyin))
-		for(var/mob/living/M in whoDyin)
+	if(LAZYLEN(targets))
```
Rohesie:
```suggestion
	if(length(targets))
```
I have no idea who added this monstrosity to  our codebase, but it must be purged.

DominikPanic: It must not. It's actually better to use it since if another method ever becomes more useful there's only one place to change it. All of these macros are genius.

spookydonut: you don't need to check length here the for loop will just do nothing anyway

code/modules/admin/verbs/podlauncher.dm
```diff
@@ -713,25 +713,27 @@
...
 			explosionString += "[X]|"
 
-	var/msg = "launched [podString][whomString].[delayString][damageString][explosionString][areaString]"
-	message_admins("[key_name_admin(usr)] [msg].")
-	if(!isemptylist(whoDyin))
-		for(var/mob/living/M in whoDyin)
+	var/msg = "launched [podString][whomString][delayString][damageString][explosionString][playerString][areaString]."
+	log_admin("[key_name(usr)] [msg]")
+	message_admins("[ADMIN_TPMONTY(usr)] [msg]")
+	if(!isemptylist(targets))
```
spookydonut: same deal here the for will just do nothing.

code/modules/admin/verbs/podlauncher.dm
```diff
@@ -713,25 +713,27 @@
...
 
 
-/datum/podlauncher/proc/supplypod_punish_log(list/whoDyin)
+/datum/podlauncher/proc/supplypod_log(list/targets, atom/target)
 	var/podString = effectBurst ? "5 pods" : "a pod"
 	var/whomString = ""
-	if(LAZYLEN(whoDyin))
-		for(var/mob/living/M in whoDyin)
+	if(LAZYLEN(targets))
+		for(var/mob/living/M in targets)
```
spookydonut: you can make this typeless since this proc is called with a list of /mob/living

code/modules/admin/verbs/podlauncher.dm
```diff
@@ -713,25 +713,27 @@
...
 
-	var/msg = "launched [podString][whomString].[delayString][damageString][explosionString][areaString]"
-	message_admins("[key_name_admin(usr)] [msg].")
-	if(!isemptylist(whoDyin))
-		for(var/mob/living/M in whoDyin)
+	var/msg = "launched [podString][whomString][delayString][damageString][explosionString][playerString][areaString]."
+	log_admin("[key_name(usr)] [msg]")
+	message_admins("[ADMIN_TPMONTY(usr)] [msg]")
+	if(!isemptylist(targets))
+		for(var/mob/M in targets)
```
spookydonut: same here can be made typeless

code/modules/admin/verbs/podlauncher.dm
```diff
@@ -594,10 +594,10 @@
 				return
 			//if(effectAnnounce)
 			//	deadchat_broadcast("<span class='deadsay'>A special package is being launched at the station!</span>", turf_target = target)
-			var/list/bouttaDie = list()
+			var/list/targets = list()
 			for(var/mob/living/M in target)
-				bouttaDie.Add(M)
-			supplypod_punish_log(bouttaDie)
+				targets.Add(M)
```
spookydonut: might be worth replacing this loop with `viewers(0, target)`

PR #1834, opened 2019-06-09
Xenos do not keep the skills of previous lives anymore.

DominikPanic:
Fixes: #1883

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -70,6 +70,9 @@
 	if(new_character.mind)								//disassociate any mind currently in our new body's mind variable
 		new_character.mind.current = null
 
+	if(isxeno(new_character))
+		QDEL_NULL(cm_skills)
+
 	SSnano.user_transferred(current, new_character) // transfer active NanoUI instances to new user
 
 	current = new_character								//associate ourself with our new body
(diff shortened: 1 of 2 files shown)


Rohesie:
This should happen when becoming a ghost.

DominikPanic:
> 
> 
> This should happen when becoming a ghost. The skills shouldn't carry over then.

That will have more sideeffects. Aghosting vs just ghosting, also if you die, ghost, then reenter, all your skills are gone since everything is kept in the mind. This is the cleanest solution I could come up with.

Rohesie:
Yet the mind is a mob var. Why does it have to be carried over to the ghost, and from the ghost to the living body?

DominikPanic:
> 
> 
> Yet the mind is a mob var. Why does it have to be carried over to the ghost, and from the ghost to the living body?

Because that's how minds work. The mind is the mob representation of the client in a way, it's made to be persistent.

spookydonut:
What does this have to do with xenos

DominikPanic:
> 
> 
> What does this have to do with xenos

Some things check `usr.mind?.cm_skills` meaning that if there are CM skills, it treats it like a human person.

spookydonut:
This should be done in mind transfer code not login

PR #1836, opened 2019-06-09
Optimizes pheromone's emitting

ma44:
## About The Pull Request

Makes the emitting of pheromones more better by changing some things

-No longer does plasma checks; needing plasma to activate the pheromones in the first place is already checked by ability and current_aura is already nulled VIA handle_living_plasma_updates() if there isn't enough plasma to sustain the aura
-Removed some redundant checks (on_fire, moved current_aura and leader_aura up to beginning)
-Optimized seeing if there's xenos in range from a range() to a get_dist and same z level in the outputting xeno's hivemind's list of hive xenos

## Why It's Good For The Game

Makes the pheromone CPU usage less intensive
## Changelog
:cl:
refactor: Made xeno pheromone emitting less worse in terms of performance
/:cl:


code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -168,31 +169,35 @@
...
-							if(xeno_caste.aura_strength > Z.frenzy_new)
-								Z.frenzy_new = xeno_caste.aura_strength
-						if("warding")
-							if(xeno_caste.aura_strength > Z.warding_new)
-								Z.warding_new = xeno_caste.aura_strength
-						if("recovery")
-							if(xeno_caste.aura_strength > Z.recovery_new)
-								Z.recovery_new = xeno_caste.aura_strength
+				//The reason why we don't check the hive of observed_xeno is just in case the watched xeno isn't of the same hive for whatever reason
+				for(var/mob/living/carbon/xenomorph/Z in Q.hive.get_all_xenos())
```
Rohesie:
```suggestion
				for(var/Z in Q.hive.get_all_xenos())
					var/mob/living/carbon/xenomorph/X = Z
```

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -168,31 +169,35 @@
...
+					if(get_dist(Q, Z) <= phero_range && (z == Z.z) && !Z.on_fire) //We don't need to check to see if it's dead or if it's the same hive as the list we're pulling from only contain alive xenos of the same hive
```
Rohesie: Why is this using the distance to the queen instead of to the observed xeno?
Also will this allow the queen to use xenos from other hives as pheromone emitters? Can it observe other hive xenos?

ma44: Should be alright

For the other part, if a queen was to observe a xeno of another hivemind for whatever reason, because there's no checks in place for seeing if their both the same hivemind, if the observed xeno hivemind was used instead of the queen, then the queen would output pheromones to that hivemind instead of it's own; aiding the enemy instead of it's allies

spookydonut:
Profile this please with different amounts of xenos vs the old code, I think this will be slower

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -168,31 +169,37 @@
...
-						if("warding")
-							if(xeno_caste.aura_strength > Z.warding_new)
-								Z.warding_new = xeno_caste.aura_strength
-						if("recovery")
-							if(xeno_caste.aura_strength > Z.recovery_new)
-								Z.recovery_new = xeno_caste.aura_strength
+				//The reason why we don't check the hive of observed_xeno is just in case the watched xeno isn't of the same hive for whatever reason
+				for(var/xenos in Q.hive.get_all_xenos())
+					var/mob/living/carbon/xenomorph/Z = xenos
+					if(get_dist(Q.observed_xeno, Z) <= phero_range && (Q.observed_xeno.z == Z.z) && !Z.on_fire) //We don't need to check to see if it's dead or if it's the same hive as the list we're pulling from only contain alive xenos of the same hive
```
spookydonut: get_dist() works on a square shaped radius but range() does a circular one, so you need to properly calculate the distance

ma44: Byond docs diagram seems to indicate that it's more of a square (assuming range() fully incorporates view() which has the diagram) so this should be okay

ma44:
15 young drones outputting in the same spot, alternating between the two procs (sorted by named)

https://pastebin.com/raw/YBGpRyyg

100 young drones output in one spot (sort by total cpu)

https://pastebin.com/raw/U0cbzsaH

Above but average usage per proc

https://pastebin.com/raw/GLGbp4zV

spookydonut:
Good shit, just do the 100 xeno test with the phero hud calls removed and a test with 50 xenos as we discussed and this is good to go, good shit :ok_hand:

PR #1839, opened 2019-06-09
Fixes admin ghosts not being properly deleted allowing for duplication

DominikPanic:
Thanks to Psy for poking me about it

### DIFF
--- a/code/modules/mob/dead/observer/logout.dm
+++ b/code/modules/mob/dead/observer/logout.dm
@@ -11,6 +11,6 @@
 
 	. = ..()
 
-	spawn(0)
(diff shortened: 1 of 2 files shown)


code/modules/mob/dead/observer/logout.dm
```diff
@@ -11,6 +11,6 @@
 
 	. = ..()
 
-	spawn(0)
-		if(src && (!key || isaghost(src)))	//we've transferred to another mob. This ghost should be deleted.
+	spawn(1)
```
spookydonut: use a timer

PR #1841, opened 2019-06-09
Pain hud

Rohesie:
Admin-spawn-only HUD to see pain / perceived_health.

This is for the Shrike, since it's a psychic xeno I want to give it more feedback on the marines without outright giving out their exact health.
Possibly for the Queen as well.


Rohesie:
This is supposed to conflict with it, because it is an alternative.
The Shrike and the Queen would have it, and there's no plans for them to have a human health hud, and the person wearing these hud shades wouldn't be wearing a medhud.

spookydonut:
Just keep in mind pain is a piece of shit at the moment so not sure how useful this will be

psykzz:
Can you post a screen shot of what this was meant to look like?

DominikPanic:
Literally health hud

Rohesie:
Not in for the shrike yet.

PR #1848, opened 2019-06-09
Hopefully fixes a lot of binding_client errors.

DominikPanic:
New players would do thousands of runtimes since the list is only created on certain circumstances.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -113,7 +113,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 		if(load_preferences() && load_character())
 			return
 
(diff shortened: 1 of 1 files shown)


psykzz:
I don't think this is going to be enough to fix the issue.
We'll need to create a list of every key possible and assign an empty list to it.

Otherwise we'd need to just add the `if _key in prefs.key_bindings`, but thats really expensive.

psykzz:
Just read up on disord, if this is on new_players only it makes more sense.

DominikPanic:
> 
> 
> Just read up on disord, if this is on new_players only it makes more sense.

It's on any players, creating new prefs makes the key_bindings null, and before they are created, it keeps runtiming.

spookydonut:
Speed merge this :heavy_check_mark:

PR #1851, opened 2019-06-10
Makes sure 99% of Topic calls have proper parent logic

DominikPanic:
Similar to my previous PR, atomized and not a code cleanup.
Lightly tested, shouldn't break anything in a major way.


spookydonut:
Test this for like a few days of general testing, enough time for bugs to be found and reported (if they exist)

PR #1852, opened 2019-06-10
Updates the github webhook processor

DominikPanic:
Not sure if this will work. The goal of this is to get only our PRs announced and only to our server.


DominikPanic:
Seems like this won't work due to some reasons discussed. Oh well.

Issue #1845, opened 2019-06-09
There's no delay between firing the TOW and the tank main gun.


Ketrai:
## Testmerges:#1841, #1842, #1837, #1839, #1831, #1829, #1825, #1538, #1815, #1810, #1804, #1766, #1758, #1757, #1789, #1817, #1816, #1832, #1843...

I doubt it was caused by any testmerges.
## Reproduction:

Equip a TOW onto a tank. Use arty module for a clearer effect.

Target an area with both the TOW and the tank main gun.
Watch as the shots fire almost perfectly in sync.

I don't think it's a bug, but rather an oversight. You can oneshot any xenomorph like this.

56[13:02:58] /(Elder Queen) (Main Hallway (132, 32, 2))
has lets out a waning guttural screech, green blood bubbling from its maw. /(Elder Queen) (NEWHP: -110.5) 
55[13:02:58] /(Elder Queen) (Main Hallway (132, 32, 2))
has been shot by [M34A2 Longstreet Light Tank (Artillery)] with [anti-armor rocket] (NEWHP: 90.42) 
54[13:02:57] /(Elder Queen) (Main Hallway (133, 32, 2))
has grabbed /(Isaac Clark) (NEWHP: 27.8) 
53[13:02:57] /(Elder Queen) (Main Hallway (133, 32, 2))
has been shot by [M34A2 Longstreet Light Tank (Artillery)] with [anti-armor rocket] (NEWHP: 297.92)

PR #1857, opened 2019-06-10
Fixes being able to cheese with tank weapon switching.

DominikPanic:
The reason I didn't do this as a do_after during hardpoint switching would be that it would be a huge nerf to moving and trying to switch.
Fixes: #1845

### DIFF
--- a/code/modules/vehicles/multitile/hardpoints.dm
+++ b/code/modules/vehicles/multitile/hardpoints.dm
@@ -247,6 +247,9 @@ Currently only has the tank hardpoints
 		return
 
 	next_use = world.time + owner.cooldowns["primary"] * owner.misc_ratios["prim_cool"]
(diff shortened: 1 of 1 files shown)


code/_onclick/ventcrawl.dm
```diff
@@ -117,10 +117,14 @@ GLOBAL_LIST_INIT(ventcrawl_machinery, typecacheof(list(
...
 			pipes_shown += A.pipe_vision_img
 	is_ventcrawling = TRUE
+	return TRUE
+
 
 /mob/living/carbon/xenomorph/hunter/add_ventcrawl(obj/machinery/atmospherics/starting_machine)
 	. = ..()
-	cancel_stealth()
+	if(.)
+		cancel_stealth()
```
psykzz: contains commits from the hunter fix

DominikPanic: :git:

Issue #1835, opened 2019-06-09
Xeno Praetorians can acid spray through walls.


DalekThePerson:
Praetorians Can Acid Spray Through walls

## Testmerges:
#1830: 'Fixes frustrating repairs of generators.' by DominikPanic at commit bcff745e46
#1831: 'Fixes a race condition launching humans into nullscape on nesting' by DominikPanic at commit 732de572e7
#1829: 'Flares burn weeds' by psykzz at commit 9a8b40b19f
#1825: 'Standarizing pills init.' by Ghommie at commit a13facf4ae
#1538: 'Mends a not-so-little race condition smoke runtime on rockets & co.' by Ghommie at commit d8e3156a0d
#1818: 'Fixes a to_chat runtime.' by DominikPanic at commit 8cd7553b01
#1820: 'Fixes a sleeper runtime' by DominikPanic at commit 7c2d968741
#1815: 'Makes the pod launcher logging better.' by DominikPanic at commit ce68163d5c
#1810: 'Fixes an admin runtime.' by DominikPanic at commit 553c1083ad
#1803: 'Makes the AI unable to interact with dropship doors.' by DominikPanic at commit 511e0a1b48
#1804: 'Cleans up some dirty perspective modifications' by DominikPanic at commit c182668897
#1801: 'Adds round IDs to the note panel.' by DominikPanic at commit fcc3d440bc
#1766: 'LOS Reduces screech effectiveness' by psykzz at commit 3f86655d11
#1758: 'Adds more explosion sounds' by MoXuT0 at commit 67c096955e
#1757: 'Adds new firing sounds for some weapons' by MoXuT0 at commit ac896e083a
#1789: 'Fixes layers of items related to actions' by DominikPanic at commit 1bb8cfae5a
#1808: 'Combines the rank/equipment/squad admin verbs, slight PP tweaks' by DominikPanic at commit 8995384498
#1805: 'Adds a slight delay to watchlists to prevent join race condition' by DominikPanic at commit 4d45a3604e
#1814: 'Makes AI report verbs more modular.' by DominikPanic at commit 0ac09beb93
#1812: 'Fixes airlock welding' by DominikPanic at commit 050cd10e76
#1817: 'Gives B18 tacticool gloves' by MoXuT0 at commit 8715b3a537
#1816: 'Fixes blood overlay for boots, adds different icon for pyro boots' by MoXuT0 at commit 3b44ed3f99
[...]

PR #1858, opened 2019-06-10
Fixes prae acid being able to spray through walls

DominikPanic:
Fixes: #1835

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -588,10 +588,26 @@
 		return FALSE
 	if(!A)
 		return FALSE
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/praetorian/abilities_praetorian.dm
```diff
@@ -35,6 +35,29 @@
 	add_cooldown()
 	addtimer(CALLBACK(X, /mob/living/carbon/xenomorph/.proc/reset_speed), rand(20,30))
 
+
+/datum/action/xeno_action/activable/spray_acid/cone/can_use_ability(atom/A, silent = FALSE)
```
spookydonut: move this to root path to make it apply to all acid spray.

Issue #1819, opened 2019-06-08
Weeds show up under walls and window frames


MoXuT0:
## Testmerges:

Round ID:  557 
BYOND Version:  512.1464 
Server revision compiled on:  
Master commit: 73f07c774ab876e36061dc3edbee9de002517054 
The following pull requests are currently test merged: 
#1815: 'Makes the pod launcher logging better.' by DominikPanic at commit ce68163d5c 
#1809: 'Another missed do_after' by DominikPanic at commit aab233539d 
#1810: 'Fixes an admin runtime.' by DominikPanic at commit 553c1083ad 
#1803: 'Makes the AI unable to interact with dropship doors.' by DominikPanic at commit 511e0a1b48 
#1804: 'Cleans up some dirty perspective modifications' by DominikPanic at commit 78e7ed6c21 
#1798: 'Fixes ghost forms not saving properly' by DominikPanic at commit fa0adbfd21 
#1799: 'Makes admins able to fully rename any mobs via VV, not just /living' by DominikPanic at commit 7843329b5f 
#1800: 'Fix new player references' by psykzz at commit 346ba64455 
#1801: 'Adds round IDs to the note panel.' by DominikPanic at commit fcc3d440bc 
#1796: 'Following the AI will now follow the eye as an observer' by DominikPanic at commit cdd73e1343 
#1770: 'Adds imaginary friends for the mentors.' by DominikPanic at commit c7d03f47ba 
#1766: 'LOS Reduces screech effectiveness' by psykzz at commit 3f86655d11 
#1758: 'Adds more explosion sounds' by MoXuT0 at commit 67c096955e 
#1757: 'Adds new firing sounds for some weapons' by MoXuT0 at commit ac896e083a 
#1694: 'm39 on back' by Rohesie at commit 275aa8c733 
#1789: 'Fixes layers of items related to actions' by DominikPanic at commit 1bb8cfae5a 
#1791: 'Fixes being unable to interact with camera computers' by DominikPanic at commit bde43cddee 
#1797: 'Fixes AI death state' by DominikPanic at commit c2dba365f1 
#1808: 'Combines the rank/equipment/squad admin verbs, slight PP tweaks' by DominikPanic at commit 8995384498 
#1807: 'Adds a varset_callback to reduce the proc amount' by DominikPanic at commit 77560e603f 
#1805: 'Adds a slight delay to watchlists to prevent join race condition' by DominikPanic at commit 4d45a3604e 
[...]

PR #1860, opened 2019-06-10
Fixes vent and weed layer issues.

DominikPanic:
I think that this is the proper way to go, otherwise we are going to keep finding more and more things breaking with planes.

Fixes: #1819
Fixes: #2087

### DIFF
--- a/code/game/objects/effects/weeds.dm
+++ b/code/game/objects/effects/weeds.dm
@@ -8,7 +8,6 @@
 
 	anchored = TRUE
 	density = 0
-	plane = FLOOR_PLANE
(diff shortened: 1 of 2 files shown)


spookydonut:
Not a fix, don't take us backwards on planes.

Rohesie:
Weren't these two lines added by the pipe PR, when our code wasn't using planes yet, causing issues?

What's the plan here, to add planes one at a time, as things break? Shouldn't we do one big implementation instead and proper refactor? Because having a half-implemented feature seems counter-productive, even if planes are the way to go in the long run.

Issue #1592, opened 2019-05-13
Users are able to use macros such as "rest" to bypass checks such as entering or being inside a vent.


Rohesie:
## Testmerges:

Round ID: 474
BYOND Version: 512.1464
Server revision compiled on: 
Master commit: 5c0c4281810834ea8a80f7a6ea69464c6f35ad4c
The following pull requests are currently test merged:
#1577: 'Runner balance tweaks' by Rohesie at commit cff6a9abc4
#1578: 'Xenos now break barrels properly.' by DominikPanic at commit e5729fcae4
#1580: 'Gives CLF leader ammo.' by DominikPanic at commit ed36f0c78b
#1556: 'adding tox bottles to med vendors' by Ghommie at commit 4c9591202f
#1568: 'Fixes a race condition init runtime' by DominikPanic at commit 524ae62dee
#1561: 'Fixes an ancient init runtime' by DominikPanic at commit 1ea06746e4
#1563: 'Fixes evil emote runtime' by DominikPanic at commit 51d4ef746c
#1564: 'Possibly fixes another runtime' by DominikPanic at commit cb3dd612ac
#1552: 'Fixes otherwordly sound spamming.' by Ghommie at commit c9c84c5f0f
#1540: 'Ports SSinput from /tg/' by DominikPanic at commit d44234fc0b

## Reproduction:

Set a macro for "rest", enter a vent, toggle it.

DominikPanic:
It's not about macros, you can use rest while anywhere.

PR #1861, opened 2019-06-10
You can no longer rest in vents.

DominikPanic:
Fixes: #1592
Can't add it to incapacitated since that would prevent movement and various other needed shit inside of vents.

### DIFF
--- a/code/modules/mob/living/living_verbs.dm
+++ b/code/modules/mob/living/living_verbs.dm
@@ -260,7 +260,7 @@
 	set name = "Rest"
 	set category = "IC"
 
-	if(incapacitated(TRUE))
+	if(incapacitated(TRUE) || is_ventcrawling)
 		return
 
 	if(!resting)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/living_verbs.dm
```diff
@@ -260,7 +260,7 @@
 	set name = "Rest"
 	set category = "IC"
 
-	if(incapacitated(TRUE))
+	if(incapacitated(TRUE) || is_ventcrawling)
```
Rohesie: Add it as an extra `if` inside `if(!resting)`, just in case.

code/modules/mob/living/living_verbs.dm
```diff
@@ -263,7 +263,7 @@
 	if(incapacitated(TRUE))
 		return
 
-	if(!resting)
+	if(!resting && !is_ventcrawling)
```
Rohesie: Since there's an `else`, I was thinking about this instead:
```
if(!resting)
	if(!is_ventcrawling)
		set_resting(TRUE, FALSE)
else if(action_busy)
```

PR #1868, opened 2019-06-11
Cleans up unused or shit mob procs

DominikPanic:
Fixes: #1976


code/modules/mob/living/carbon/monkey/monkey.dm
```diff
@@ -270,11 +270,6 @@
 
 	return 1
 
-/mob/living/carbon/get_standard_pixel_y_offset()
```
spookydonut: Macro this may be?

code/modules/mob/living/carbon/xenomorph/castes/warrior/warrior.dm
```diff
@@ -57,13 +57,13 @@
 	if(isxeno(L))
 		return ..()
 
-	if(lunge && ..(L, TRUE))
+	if(lunge && ..(L, supress_message))
```
spookydonut: Suppress has two ps

code/modules/mob/living/carbon/monkey/monkey.dm
DominikPanic: How would I macro that since it takes both `lying` and the `initial` of pixel_y?

code/game/objects/structures/ladders.dm
```diff
@@ -88,7 +88,30 @@
...
 		return
-	user.trainteleport(ladder_dest.loc)
+	var/obj/buckled
+	var/atom/movable/pulling
+	var/mob/pulling_mob
+	if(user.pulling)
+		pulling = user.pulling
+		if(ismob(user.pulling))
+			pulling_mob = user.pulling
+			if(pulling_mob.buckled)
```
Rohesie: What if the pulled mob isn't buckled?

Rohesie:
As discussed in private, `trainteleport()` could use some improvements, but it does what is required. No need to remove it.

PR #1877, opened 2019-06-11
Fixes stun from being released from a non-gibbed xeno while devoured

DominikPanic:
Only a gib would set it to 1, now a death does too.


Rohesie:
Could you do it via a signal?
If it's too much let me know and I'll try to do it myself.

DominikPanic:
> 
> 
> Could you do it via a signal?
> If it's too much let me know and I'll try to do it myself.

Please go ahead then :c

PR #1888, opened 2019-06-11
Fixes a transit turf issue

DominikPanic:
Likely to be fixed by the shuttle port, this is a temporary solution since we do not implement that behavior yet.


spookydonut:
We don't have behaviour where hitting the edge of the map or entering space in transit does stuff

spookydonut:
Not sure what this fixes

DominikPanic:
> 
> 
> Not sure what this fixes

It fixes the runtime, or rather supresses it until this can be properly implemented.

PR #1902, opened 2019-06-12
Fixes explosions deleting worn items

DominikPanic:


code/game/atoms.dm
```diff
@@ -252,8 +252,10 @@ directive is properly returned.
 /atom/proc/relaymove()
 	return
 
-/atom/proc/ex_act()
-	return
+/atom/proc/ex_act(severity, target)
+	set waitfor = FALSE
```
```suggestion
```
This part is not needed for our explosion.

Rohesie: If you are adding `CHECK_TICK` further down, this may be necessary.
But I would remove those as well until there's a proper explosion port.

code/__DEFINES/components.dm
```diff
@@ -57,6 +57,7 @@
 #define COMSIG_ATOM_ATTACK_GHOST "atom_attack_ghost"			//from base of atom/attack_ghost(): (mob/dead/observer/ghost)
 	#define COMPONENT_NO_ATTACK_HAND 1							//works on all attack_hands.
 #define COMSIG_PARENT_EXAMINE "atom_examine"                    //from base of atom/examine(): (/mob)
+#define COMSIG_ATOM_EX_ACT "atom_ex_act"						//from base of atom/ex_act(): (severity, target)
```
spookydonut: Not sure what this signal is for

DominikPanic: Nothing uses it yet, but the source is there now, /tg/ does it via the base atom proc, just like the attack_hands and such

PR #1917, opened 2019-06-13
TGS Debugdalus - a stripped down ship for testing purposes

Alterist:
Adds the Debugdalus, a smaller stripped down marine ship the only contains the essentials. mainly for testing new PRs and such. It will be disabled by default
Debugdalus

### DIFF
--- a/_maps/debugdalus.json
+++ b/_maps/debugdalus.json
@@ -1,0 +1,8 @@
+{
+    "map_name": "Debugdalus",
+    "map_path": "map_files/debugdalus",
+    "map_file": "tgs_debugdalus.dmm",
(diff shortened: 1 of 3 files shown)


Rohesie:
Needs to be added to the all maps list.

_maps/debugdalus.json
```diff
@@ -0,0 +1,8 @@
+{
+    "map_name": "Debugdalus",
+    "map_path": "map_files/debugdalus",
+    "map_file": "tgs_debugdalus.dmm",
+	"traits": [{"Marine Main Ship": false}],
```
DominikPanic: I don't think it's necessary to add a negative trait like this since that's already the default.

PR #1925, opened 2019-06-14
Add end of round stats to PRs

psykzz:
## About The Pull Request

This will send a HTTP GET before the end of round deathmatch. It will send a list of merged PRs, number of runtimes (unique and total) along with the current commit and round Id.

This will get sent to a server that will add it to each PR to record rounds.

See https://github.com/psykzz/tgmc-server/pull/2 for an example (and for server code)

## Why It's Good For The Game

More data and tracking around rounds played with a test merge.

## Changelog
:cl:
server: End of round stats are sent and added to test merges.
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -301,6 +301,10 @@ Master controller and performance related.
 
 /datum/config_entry/flag/use_recursive_explosions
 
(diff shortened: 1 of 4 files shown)


ninjanomnom:
TGS already has configs to do this and this should 100% not ever be done from within game code

PR #1928, opened 2019-06-14
Admin level tweaks

DominikPanic:
-Removed the unused shuttle doors that were removed since they're nonfunctional now.
-Made the meeting areas next to the shuttle locations nicer.
-Made the "server room" more efficient with machines that will actually be useful
-Added required pod areas so it works out of the box now
-The command area has been refactored, there's a meeting area now, an armory, a vault with a nuke and an entry chamber.
-Made the tomb a bit nicer since previously it was cluttered a lot
-Removed the prison area and landmarks since it's unused and a bad admeme tool regardless

From Franc's commit:
Moves cargo pods into their own wing
Adds a brief room for the ERT
Mapsize is now 255x255 (Previously 150x100)
Gave the shuttle spawn areas more than enough space with the spare room.
Moved the start area to the bottom left.
probably other minor crap I forgot


spookydonut:
Keep it small as possible for faster loading and less memory usage.

DominikPanic:
@francinum Think you could change it? Don't wanna revert your changes.

Issue #1926, opened 2019-06-14
stop pulling keybind is broken


spookydonut:
should be !M.pulling

street cred to @qwertyquerty for finding this

spookydonut:
https://github.com/tgstation/TerraGov-Marine-Corps/blob/baec3a6a230f6e8536439a990b8cb06b1f87fc13/code/datums/keybinding/mob.dm#L60

qwertyquerty:
bruh wheres muh street cred

PR #1930, opened 2019-06-14
Fixes stop pulling keybind

DominikPanic:
Fixes: #1926

### DIFF
--- a/code/datums/keybinding/mob.dm
+++ b/code/datums/keybinding/mob.dm
@@ -57,7 +57,7 @@
 
 /datum/keybinding/mob/stop_pulling/down(client/user)
 	var/mob/M = user.mob
-	if (M.pulling)
+	if(!M.pulling)
 		to_chat(user, "<span class='notice'>You are not pulling anything.</span>")
 	else
 		M.stop_pulling()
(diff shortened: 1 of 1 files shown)


code/datums/keybinding/mob.dm
```diff
@@ -57,7 +57,7 @@
 
 /datum/keybinding/mob/stop_pulling/down(client/user)
 	var/mob/M = user.mob
-	if (M.pulling)
+	if(!M.pulling)
```
psykzz:
```suggestion
	if(M.pulling)
	    M.stop_pulling()
```

Just delete the rest, I don't feel the message is helpful on keybindings.

Rohesie: Any chance of making this a signal? I can guide you through how, and it's useful knowledge.

DominikPanic: reeee sure

spookydonut: doesnt need to be a signal, pulling and stop_pulling are defined at mob or above

spookydonut: just do the suggested change and it can merge

PR #1945, opened 2019-06-14
Fixes barricade repair issues

DominikPanic:
Fixes: #1944
Not sure this is the best solution, feel free to suggest a better one that wouldn't require a complete rework of the system


Rohesie:
Ask for harm intent instead.

PR #1957, opened 2019-06-15
fixes queen calldown bug

spookydonut:
stops queen calling down if the shuttle is already down

### DIFF
--- a/code/modules/shuttle/marine_dropship.dm
+++ b/code/modules/shuttle/marine_dropship.dm
@@ -246,6 +246,13 @@
 /datum/game_mode/proc/can_summon_dropship()
 	if(SSticker.round_start_time + MIN_CALLDOWN_TIME > world.time)
 		return FALSE
+	var/obj/docking_port/mobile/marine_dropship/D
+	for(var/k in SSshuttle.dropships)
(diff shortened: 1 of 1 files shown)


code/modules/shuttle/marine_dropship.dm
```diff
@@ -246,6 +246,15 @@
 /datum/game_mode/proc/can_summon_dropship()
 	if(SSticker.round_start_time + MIN_CALLDOWN_TIME > world.time)
 		return FALSE
+	var/obj/docking_port/mobile/marine_dropship/D
+	for(var/k in SSshuttle.dropships)
+		var/obj/docking_port/mobile/M = k
+		if(M.id == "alamo")
```
DominikPanic: Right now there's only one dropship, but what about when the Normandy gets back? Or the ID changes to something else, since this is not a define it'd just silently fail.

Wouldn't something like `is_marine_dropship` tagging marine dropships with a special variable/flag be better?

spookydonut: it's designed for this gamemode which has one shuttle, no need to overcomplicate it, also outside the scope of this fix.

PR #1968, opened 2019-06-15
Kwin's Tgs canterbury

spookydonut:
this is the ship with fixed areas, can be merged and amended later.


Rohesie:
Add to all maps?

spookydonut:
To be clear, this is just the shuttle with the template in a state where it can be loaded, kwin can make whatever fixes to it later, merging this will allow psy to have something to work with for the new gamemode

spookydonut:
> 
> 
> Add to all maps?

no, because templates are done separately.

PR #1970, opened 2019-06-15
Fixes asset topic sanity

DominikPanic:
Stolen from /tg/
https://github.com/tgstation/tgstation/pull/44550


Rohesie:
Can you make it shriek if that value is negative?

DominikPanic:
> 
> 
> Can you make it shriek if that value is negative?

No. /tg/ doesn't do that.

Rohesie:
Just run a `stack_trace` so we know if it happens.

DominikPanic:
> 
> 
> Just run a `stack_trace` so we know if it happens.

If ninja says so I will. He merged it on /tg/ and I assume there's a good reason this is not present.

PR #1985, opened 2019-06-16
Resist refactor

Rohesie:
Refactored the big ugly verb that was `resist()`. Some solutions are from /tg/ to increase compatibility, others are custom, like using a signal.

## Changelog
:cl:
balance: Fixed the warrior grab move-resist delay, which was broken. To free yourself you need to move/resist twice and there's a cooldown of around 1 second between each. Button/macro and movement resists no longer stack.
tweak: Resisting works a bit differently than usual, but nothing big. If you are bucklecuffed you'll try to remove the cuffs, and if you succeed you'll just be able to unbuckle as normal. If you are hand and legcuffed, you'll resist whatever was slapped first.
/:cl:


code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -316,6 +318,42 @@
...
+		"<span class='italics'>You hear banging from [src].</span>")
+	if(do_after(user, breakout_time, target = src))
+		if(!user || user.stat != CONSCIOUS || user.loc != src || opened || (!locked && !welded) )
+			return
+		//we check after a while whether there is a point of resisting anymore and whether the user is capable of resisting
+		user.visible_message("<span class='danger'>[user] successfully broke out of [src]!</span>",
+							"<span class='notice'>You successfully break out of [src]!</span>")
+		bust_open()
+	else
+		if(user.loc == src) //so we don't get the message if we resisted multiple times and succeeded.
```
DominikPanic: combine into else if

Rohesie:
Alright, this should be ready for testing.
Not entirely happy with registering a signal for the mob on fire to listen to, but there's no fire datum.

Handcuffs and legcuffs have duplicated code that should be fixed once #2005 gets merged. Nothing big, really.

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -316,6 +318,42 @@
...
+	if(user.action_busy) //Already resisting or doing something like it.
+		return
+	//okay, so the closet is either welded or locked... resist!!!
+	user.changeNext_move(CLICK_CD_BREAKOUT)
+	user.last_special = world.time + CLICK_CD_BREAKOUT
+	user.visible_message("<span class='warning'>[src] begins to shake violently!</span>", \
+		"<span class='notice'>You lean on the back of [src] and start pushing the door open... (this will take about [DisplayTimeText(breakout_time)].)</span>", \
+		"<span class='italics'>You hear banging from [src].</span>")
+	if(do_after(user, breakout_time, target = src))
+		if(!user || user.stat != CONSCIOUS || user.loc != src || opened || (!locked && !welded) )
```
psykzz: Isn't these checks inside the `do_after` if not i believe do_after has `extra_checks` would that not work?

DominikPanic: not really, extra checks is a callback

Rohesie: To be fair some of these are excessive. This is /tg/ code, and not the best of it.

Rohesie:
Integrated it with the `/restraint/ path change.
Tweaked it to make resisting warriors a bit easier.
Before, you'd take like 6 seconds by mashing the button/macro resist and almost instantly by mashing the movement keys, both stacking. Now it should take between one and two seconds mashing any or both of them (same thing).

PR #1986, opened 2019-06-16
Fix ghost alpha for admin made ghosts

psykzz:
## About The Pull Request

When made a ghost by an admin `ghostize()` sets the alpha to 127. This is incorrect for the ghost mob which already has alpha applied. 

This now sets the alpha of admin ghosts to 255.

## Why It's Good For The Game

Consistent with other ghosts


code/modules/admin/topic.dm
```diff
@@ -270,7 +270,7 @@ Status: [status ? status : "Unknown"] | Damage: [health ? health : "None"]
 			if("Cancel")
 				return
 			if("Yes")
-				location = get_turf(oldusr)
+				location = get_turf(usr.loc)
```
DominikPanic: Are you aware of how this will affect the behavior?

psykzz: I believed so but i didn't read above, oldusr was already the admin user, i only needed to change the line below this.

psykzz:
Travis is complaining about the debug ship.

PR #1989, opened 2019-06-16
Adds double click with modifiers

psykzz:
## About The Pull Request

This allows for mobs to use modifiers while double clicking things

## Why It's Good For The Game

Allows for more control via mouse which opens up for better gameplay.


Rohesie:
Good concept.

PR #1995, opened 2019-06-16
Fixes xeno hud not showing synths and dead humans

MoXuT0:
## About The Pull Request

Per title.

## Changelog
:cl:
fix: Xeno hud now properly shows synths and dead humans.
/:cl:


Rohesie:
I was planning on doing a custom state hud for xenos and non-medical-hud mobs.

Rohesie:
Opened #2004 , which is the way I wanted to make it.

PR #2006, opened 2019-06-17
Updates to travis scripts

spookydonut:
checks templates now and some other things


tools/travis/template_dm_generator.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
+import os
+import os.path
```
psykzz: redundant line

spookydonut: same as tg

Rohesie: Seems to be redundant from what it's said, just accept the suggestion.

Rohesie:
You'll have to remove the icon state generation check because the only cure for that right now would be a nuke.

tools/travis/template_dm_generator.py
spookydonut: > 
> 
> Seems to be redundant from what it's said, just accept the suggestion.

i just mean i'll change it upstream

PR #2008, opened 2019-06-17
Unmaintained maps fix

Rohesie:
Should make all maps mappeable and not complain when we add the travis check.

Removed `hideout.dmm` because it was not merely unmaintained, it wasn't even in .tgm format. And I didn't know exactly its purpose. Due to its tiny size it's easy to be remade if needed.

Added back some fluff removed items using /tg/ paths and in one case icon.


code/modules/power/singularity/field_generator.dm
```diff
@@ -0,0 +1,11 @@
+/obj/machinery/field/generator
```
spookydonut: just delete any instances of these

code/game/machinery/computer/aifixer.dm
```diff
@@ -0,0 +1,4 @@
+/obj/machinery/computer/aifixer
```
spookydonut: same here

PR #2009, opened 2019-06-17
Remove failsafe for ssdir

psykzz:
There shouldn't be failsafes just good code, especially when we are throwing stack_traces to catch these.


Rohesie:
Fully agreed with the philosophy.

PR #2010, opened 2019-06-17
Recolors pyro backpack to be same as other gear

MoXuT0:
## About The Pull Request

Per title. Also added snow version because why not?

## Changelog
:cl:
imageadd: Pyro's backpack is now same color as other marine gear.
/:cl:


Rohesie:
Very nice.

PR #2014, opened 2019-06-17
Reduces admin code duplication, fixes grammar issues.

DominikPanic:
:cleanbot:

### DIFF
--- a/code/__DEFINES/admin.dm
+++ b/code/__DEFINES/admin.dm
@@ -100,3 +100,12 @@
 
 
 #define PR_ANNOUNCEMENTS_PER_ROUND 5 //The number of unique PR announcements allowed per round
+
+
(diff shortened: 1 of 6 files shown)


DominikPanic:
> 
> 
> In the long run try to avoid variables like LGIVE and MREC, which look like constants and are not really very clear. `given_living`, `mob_received` or the likes might be better.
> 
> Other than that, good code improvement.

That was your suggestion :p

Rohesie:
> 
> 
> > In the long run try to avoid variables like LGIVE and MREC, which look like constants and are not really very clear. `given_living`, `mob_received` or the likes might be better.
> > Other than that, good code improvement.
> 
> That was your suggestion :p

Got counter-suggested by ninjanomnom and it made sense.
Besides, a word of four or five characters all in caps is kind of the standard of a define.

PR #2018, opened 2019-06-18
REVERSE_DIR macro

Rohesie:
Profile method:

```
/proc/dir_method_one(var/times)
	var/start_time = world.timeofday
	var/usless_var
	for(var/i in 1 to times)
		for(var/j in list(1,2,4,5,6,8,9,10))
			usless_var = REVERSE_DIR(j)
	return "First = [(world.timeofday-start_time)*0.1] sec, looped [times] times, with [usless_var] as last value"

/proc/dir_method_two(var/times)
	var/start_time = world.timeofday
	var/usless_var
	for(var/i in 1 to times)
		for(var/j in list(1,2,4,5,6,8,9,10))
			usless_var = reverse_direction(j)
	return "Second = [(world.timeofday-start_time)*0.1] sec, looped [times] times, with [usless_var] as last value"
```

Results, looping 1.000.000 times each:
```
dir_method_one returned: First = 4.2 sec, looped 1e+006 times, with 5 as last value
dir_method_one returned: First = 4.2 sec, looped 1e+006 times, with 5 as last value
dir_method_one returned: First = 4.1 sec, looped 1e+006 times, with 5 as last value
dir_method_two returned: Second = 4.5 sec, looped 1e+006 times, with 5 as last value
dir_method_two returned: Second = 4.6 sec, looped 1e+006 times, with 5 as last value
dir_method_two returned: Second = 4.6 sec, looped 1e+006 times, with 5 as last value
```

Proc overhead and all.

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -893,6 +893,10 @@ var/global/list/common_tools = list(
 	)
 
 
+//Actually better performant than reverse_direction()
+#define REVERSE_DIR(dir) ( ((dir & 85) << 1) | ((dir & 170) >> 1) )
+
+
 /proc/reverse_direction(direction)
 	switch(direction)
 		if(NORTH)


code/__HELPERS/unsorted.dm
```diff
@@ -893,6 +893,10 @@ var/global/list/common_tools = list(
 	)
 
 
+//Actually better performant than reverse_direction()
+#define REVERSE_DIR(dir) ( ((dir & 85) << 1) | ((dir & 170) >> 1) )
```
DominikPanic: If it's better performant can you remove the other one and make other things use it?

Rohesie: I can substitute the other one where it makes sense, but I think I could see specific scenarios where the other might be more interesting, such as when you need to return a `null` value when the input is not a valid direction. It's a safe proc.

Also I'd like some review first.

Ghommie: I can't quite imagine a situation where you'd expect it to return null though, since all atoms are supposed to have a direction by default.

Anyhow, This macro should work as intended and is neatly wrapped between parenthesis,
so you can safely replace `reverse_direction()` uses with this already, imo.

Rohesie: Well, I know it's an unlikely place but if you run `reverse_direction(16)`, which is not an actual direction, it will return you `null`.
If you run `REVERSE_DIR(16)` it will return you `15`, which is a nonsensical value. Which could be interpreted as you facing every direction, really, as `15 & dir` will return true for any `dir`.

Ghommie: REVERSE_DIR(16) (ergo UP) should actually return 32 (DOWN), and viceversa.
Are you sure you did calculate it correctly? Or am I failing to remember how binary operations even work?

Rohesie: Uhh, I didn't do the math and gave wrong info. `REVERSE_DIR(15)` should return `15`
In short, `15` is the same as `1111`. This will take the even and odd bits and shift them. Since both the even and odd bits are the same, the result will be the same.

Still a nonsensical return to a nonsensical value, instead of a `null` return.

Rohesie: Actually yes. I didn't do the math properly and gave wrong info.
`16` is `010000`. This would shift the only bit it can to produce `100000`, aka `32`.
Which is not really implemented yet AFAIK. Just reserved for later use.

Ghommie: Good to know there is no problem then. 👏

spookydonut: things with directions of UP or DOWN shouldn't be using reverse direction imo

PR #2021, opened 2019-06-18
More travis improvements

spookydonut:
more things from /tg/


Rohesie:
I've called it Nanotransen for quite a while before I noticed it wasn't the case.

spookydonut:
Can be merged after #2042

PR #2024, opened 2019-06-18
Cleans up logging categories on most things

DominikPanic:
log_admin should be for actions an admin committed.
log_admin_private should be for punishment/exploit related stuff
log_game should be for important stuff that's good to log globally
admins should be messaged whenever something important happens
warning should be used for stuff that can happen but is bad
stack_trace should be used for stuff that shouldn't happen


Rohesie:
Very, very nice. Great standard.
Should also help with future statbus.

PR #2036, opened 2019-06-18
Adds proper paper logging and explicit round end log.

DominikPanic:
https://github.com/tgstation/tgstation/pull/43565
https://github.com/tgstation/tgstation/pull/43744

### DIFF
--- a/code/__HELPERS/_logging.dm
+++ b/code/__HELPERS/_logging.dm
@@ -133,7 +133,8 @@
 
 
 /proc/log_href(text)
-	WRITE_LOG(GLOB.world_href_log, "HREF: [text]")
+	if(CONFIG_GET(flag/log_hrefs))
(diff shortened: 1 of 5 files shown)


code/__HELPERS/_logging.dm
```diff
@@ -167,6 +168,10 @@
 	SEND_TEXT(world.log, text)
 
 
+/proc/log_paper(text)
+	WRITE_LOG(GLOB.world_paper_log, "PAPER: [text]")
```
psykzz: why are hrefs behind a config but logging paper is not?

paper seems the lesser of the two?
(although log file will be significantly smaller).

DominikPanic: not every one is behind a config. the href already existed it just did nothing.

Rohesie:
Merge conflict.

PR #2040, opened 2019-06-19
More more globals removed

spookydonut:


DominikPanic:
Could you add the new flags to bitfields.dm?

PR #2041, opened 2019-06-19
[s] blacklist version

spookydonut:
theres some security issues with this build


DominikPanic:
This was discussed in [redacted] and is a bad idea, since quite a lot of people still use this build (afaik 15%+ people on average)

spookydonut:
> 
> 
> This was discussed in [redacted] and is a bad idea, since quite a lot of people still use this build (afaik 15%+ people on average)

Was asked by Roh to do this

PR #2046, opened 2019-06-19
Add travis check

psykzz:
## About The Pull Request

Add a check to travis to find changed files (files apart of the PR) that have var/.* as an arguement.

> **This will fail your build if you change a file and do not remove all `var/` from proc arguements.**

## Why It's Good For The Game

Code standards.


psykzz:
First to make sure it compiles when no changes detected.

psykzz:
Now to add a failing test case.

psykzz:
We actually need to fix our shit first before we test.

code/__HELPERS/unsorted.dm
```diff
@@ -1140,7 +1140,7 @@ will handle it, but:
 	if(final_x || final_y)
 		return locate(final_x, final_y, T.z)
 
-/proc/animate_speech_bubble(image/I, list/show_to, duration)
+/proc/animate_speech_bubble(image/I, list/show_to, var/duration)
```
Rohesie:
```suggestion
/proc/animate_speech_bubble(image/I, list/show_to, duration)
```
Merge garbage.

psykzz: (you didn't read did you)

Rohesie: Ooops!

spookydonut:
Should only apply to lines that are changed not the whole file

spookydonut:
This matches `//      for(var/x in y)`

PR #2047, opened 2019-06-20
removes var/ from proc params

spookydonut:
applies #2046

### DIFF
--- a/code/_onclick/ai.dm
+++ b/code/_onclick/ai.dm
@@ -156,5 +156,5 @@
 //
 // Override TurfAdjacent for AltClicking
 //
-/mob/living/silicon/ai/TurfAdjacent(var/turf/T)
+/mob/living/silicon/ai/TurfAdjacent(turf/T)
 	return (GLOB.cameranet && GLOB.cameranet.checkTurfVis(T))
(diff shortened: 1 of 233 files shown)


psykzz:
The not opposed to this but the reason i did it in changed files only was to enforce this over time, instead of the sweeping change like this.

If we do want to do this ill update my PR to enforce it on every file instead.

DominikPanic:
> 
> 
> The not opposed to this but the reason i did it in changed files only was to enforce this over time, instead of the sweeping change like this.
> 
> If we do want to do this ill update my PR to enforce it on every file instead.

If your PR was merged travis would not compile until this was fixed

Rohesie:
Travis check.

spookydonut:
this is separate from #2046, this just brings the current code into compliance with what that will enforce

psykzz:
> If your PR was merged travis would not compile until this was fixed 

Why not @DominikPanic ?

DominikPanic:
> 
> 
> > If your PR was merged travis would not compile until this was fixed
> 
> Why not @DominikPanic ?

if you add a check to make something fail without fixing the stuff that makes it fail travis wouldn't be happy

PR #2055, opened 2019-06-21
Jump to Area ugly hack

Rohesie:
Reverts the code to an ugly hack because of how areas work.

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -617,12 +617,38 @@
 	if(!istype(A))
 		return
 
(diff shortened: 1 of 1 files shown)


code/modules/admin/admin_verbs.dm
```diff
@@ -617,14 +617,40 @@
 	if(!istype(A))
 		return
 
-	var/turf/T = get_turf(A)
-	N.forceMove(T)
+	var/turf/T = A.jump_to_by(N)
```
DominikPanic: Just replace this with pick(get_area_turfs(area)) instead of making 3 shitcode procs

PR #2078, opened 2019-06-23
Increase weeds size

### LINKED ISSUE #2075: weeds' range is smaller
image
image
New is first image, old is second.
This was never documented, as far as I can tell, in changelogs or anything I can find searching PRs on git. Therefore, I believe it to be an unintended change by @psykzz in #1375 .

psykzz:
Fixes #2075 

## About The Pull Request

Changes weed format from a star to a square.

## Why It's Good For The Game

This fixes a previously undocumented change, where the size was smaller..

## Changelog
:cl:
balance: Weed nodes are backing to creating squares of weeds, which should be much more similar to the original weeds before the last refactor
/:cl:

Before:
image

Current: 
image

New:

image


code/game/objects/effects/weeds.dm
```diff
@@ -205,25 +205,55 @@
...
+	// 				// Finish here, but add it to expand weeds into
+	// 				node_turfs += AdjT
+	// 				continue
+
+	// 			turfs_to_check += AdjT
+	// 			node_turfs += AdjT
+
+	for(var/direction in GLOB.alldirs) //First pass to get the adjacent ones.
+		var/turf/adjacent_turf = get_step(src, direction)
+		if(adjacent_turf.density || LinkBlocked(src, adjacent_turf) || TurfBlockedNonWindow(adjacent_turf))
```
Rohesie: This condition should produce a `continue` instead.

code/game/objects/effects/weeds.dm
```diff
@@ -205,25 +205,55 @@
...
+							to_chat(world, "new turf - steraight")
+				if(NORTHEAST, NORTHWEST, SOUTHEAST, SOUTHWEST) //These are corner cases and check both their direction and their cardinal cases, three in total.
+					for(var/next_direction in list(expansion_dir, expansion_dir & NORTH|SOUTH, expansion_dir & WEST|EAST))
+						var/turf/next_turf = get_step(current_turf, next_direction)
+						if(!next_turf.density && !LinkBlocked(current_turf, next_turf) && !TurfBlockedNonWindow(next_turf) && !(next_turf in turfs_to_check))
+							turfs_to_check += next_turf
+							if(!(next_turf in node_turfs))
+								node_turfs += next_turf
+								to_chat(world, "new turf - diag")
+			turfs_to_check -= current_turf
```
Rohesie: Not sure if this is nested in the last `for`, but it should be.

code/game/objects/effects/weeds.dm
```diff
@@ -205,25 +205,55 @@
...
-				if (AdjT in node_turfs) // Ignore existing weeds
-					continue
-				if(AdjT.density || LinkBlocked(T, AdjT) || TurfBlockedNonWindow(AdjT))
-					// Finish here, but add it to expand weeds into
-					node_turfs += AdjT
-					continue
-
-				turfs_to_check += AdjT
-				node_turfs += AdjT
+	// var/list/turfs_to_check = list()
```
spookydonut: delete commented out code.

code/game/objects/effects/weeds.dm
spookydonut: looks correct here

Rohesie:
5x5 seems like a good size.
Given the weeds regrow, 7x7 is excessive.

psykzz:
The original size was 7x7, but im happy to reduce it down if thats whats being asked?
The bug report was more about the sizing difference.

Im afraid that changing this down (even if it does regrow) will require a mapping change to ensure proper coverage of default weeds in maps.

Rohesie:
Yes, it was 7x7 minus 4 corners, but it didn't regrow.
It's fine if the map is not fully weeded at roundstart,

psykzz:
Updated weed size 

image

psykzz:
This is technically fixed, but @Rohesie didnt like the performance impacts and wanted to take a stab at improving it. It could be merged knowing the performance issues, but thats a larger discussion.

Rohesie:
My fix PR has grown stale.
This PR is fine by itself, but the feature could surely use some optimization. It's unnecessarily costly for what it does.

PR #2096, opened 2019-06-23
Recreate an autodoc radio if it disappears

psykzz:
Temp fix for the autodoc radio that sometimes just nulls out


DominikPanic:
The reason is not so unknown, it happens during a specific procedure (larva bursts, human inside dies, it g ets stuck in a runtime loop)

Rohesie:
Deprecated by #2153

PR #2112, opened 2019-06-25
Removes removed config, locally restarting is now better

DominikPanic:
One less popup for us devs

### DIFF
--- a/code/modules/admin/server_verbs.dm
+++ b/code/modules/admin/server_verbs.dm
@@ -13,11 +13,12 @@
 		return
 
 	var/message = FALSE
-	switch(alert("Send the new round message?", "Message", "Yes", "No", "Cancel"))
-		if("Yes")
-			message = TRUE
-		if("Cancel")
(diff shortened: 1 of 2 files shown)


code/modules/admin/server_verbs.dm
```diff
@@ -13,11 +13,12 @@
 		return
 
 	var/message = FALSE
-	switch(alert("Send the new round message?", "Message", "Yes", "No", "Cancel"))
-		if("Yes")
-			message = TRUE
-		if("Cancel")
-			return
+	if(CONFIG_GET(string/restart_message))
```
spookydonut: should be whether discord is configured or not.

DominikPanic: That has no relevance to this, since it's just a link the in-game button uses, this is TGS related. Perhaps some TGS check but I think that this is more appropriate since it asks about sending a message, and thus checking if the message exists makes sense.

PR #2119, opened 2019-06-26
Full jobs squadhud

Rohesie:
## Changelog
:cl: Rohesie (implementation) and Sleepy (sprites)
imageadd: Every job has gained an icon in the squad HUD, which should be default for everyone on the ship. Gold is for the ship crew (navy and attachées), silver for the deployed troops (marines). Blue is for command, light blue for medical, yellow for engineering, brown for requisitions, red for security.
/:cl:
dreamseeker_iGce3DKtV4


Rohesie:
The idea is to, with this, remove the need for the job part of the security one.

DominikPanic:
I understand the idea, I just disagree with it. In my opinion the security hud should remain as-is, and squad hud should only show squad jobs rather than all jobs. The more HUD icons there are during regular gameplay the more it clutters everything. This combined along with the status HUD can be quite disorienting with multiple people involved.

PR #2122, opened 2019-06-26
Adds popcaps, panic bunker, a few ticker tweaks

DominikPanic:


code/__HELPERS/mobs.dm
```diff
@@ -162,4 +162,12 @@
...
\ No newline at end of file
+			return M
+
+
+/proc/living_player_count()
+	. = 0
+	for(var/i in GLOB.player_list)
+		var/mob/M = i
+		if(M.stat != DEAD)
+			. += 1
```
Rohesie:
```suggestion
			.++
```

code/modules/admin/panels/ban_panel.dm
```diff
@@ -1100,11 +1100,25 @@
...
+
 	//Guest Checking
-	if(!real_bans_only && IsGuestKey(key))
+	if(!real_bans_only && !C && IsGuestKey(key))
 		if(CONFIG_GET(flag/guest_ban))
 			log_access("Failed Login: [key] - Guests not allowed")
 			return list("reason"="guest", "desc"="\nReason: Guests not allowed. Please sign in with a byond account.")
+		if(CONFIG_GET(flag/panic_bunker) && SSdbcore.Connect())
+			log_access("Failed Login: [key] - Guests not allowed during panic bunker")
+			return list("reason"="guest", "desc"="\nReason: Sorry but the server is currently not accepting connections from never before seen players or guests. If you have played on this server with a byond account before, please log in to the byond account you have played from.")
```
psykzz: Should this not use `string/panic_bunker_message `

code/controllers/stonedmc/subsystem/ticker.dm
```diff
@@ -17,25 +17,24 @@ SUBSYSTEM_DEF(ticker)
...
 	var/list/datum/mind/minds = list()		//The characters in the game. Used for objective tracking.
-	var/datum/mind/liaison
 
 	var/delay_end = FALSE					//If set true, the round will not restart on it's own
 	var/admin_delay_notice = ""				//A message to display to anyone who tries to restart the world after a delay
 
 	var/time_left							//Pre-game timer
 	var/start_at
 
-	var/gametime_offset = 432000			//Deciseconds to add to world.time for station time.
```
psykzz: Does this mean round time is in sync with normal time?
If so, is it possible to see how long the round has been running somewhere else?

DominikPanic: The variable was never used. This will not affect anything.

code/modules/admin/panels/ban_panel.dm
DominikPanic: No that's used in a different place

PR #2125, opened 2019-06-27
Signals Episode 7 The Signals Awaken

spookydonut:
take two of this

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
@@ -39,8 +39,10 @@
 	return ..()
 
 /datum/action/xeno_action/neuroclaws/proc/slash(datum/source, mob/living/carbon/human/H, damage, list/damage_mod)
-	if(!active || !H)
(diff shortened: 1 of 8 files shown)


code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
```diff
@@ -7,29 +7,47 @@
...
-	if(X.neuro_claws)
+	if(active)
 		button.overlays += image('icons/mob/actions.dmi', button, "neuroclaws_on")
 	else
 		button.overlays += image('icons/mob/actions.dmi', button, "neuroclaws_off")
 	return ..()
 
+/datum/action/xeno_action/neuroclaws/proc/slash(datum/source, mob/living/carbon/human/H, damage, list/damage_mod)
+	if(!active || !H)
+		return
```
Rohesie:
```suggestion
		CRASH("neuroclaws/proc/slash invoked with improper active ([active]) and/or H ([H])")
```
So we know when it happens and fix it if needed.

Rohesie:
Good code. Me likes.

PR #2128, opened 2019-06-27
Full auto

Rohesie:
This image is a lie, because it's for now restricted to guns. But the M56D will get it in the future as well.

PSKFU4yqDO

## Changelog
:cl:
add: Full-auto and autoburst modes added, along with actions to toggle those. Dakka-dakka!
/:cl:


code/_onclick/drag_drop.dm
```diff
@@ -21,10 +21,26 @@
 
 
 /client/MouseDown(object, location, control, params)
+	if(!control)
+		return
 	if(mouse_down_icon)
 		mouse_pointer_icon = mouse_down_icon
+	if(SEND_SIGNAL(src, COMSIG_CLIENT_MOUSEDOWN, object, location, params) & COMSIG_CLIENT_MOUSEDOWN_INTERCEPT)
```
ninjanomnom: Provide all relevant arguments for new signals you create even if you wont use it yourself

code/_onclick/drag_drop.dm
```diff
@@ -21,10 +21,26 @@
...
\ No newline at end of file
+		mouse_pointer_icon = mouse_up_icon
+	SEND_SIGNAL(src, COMSIG_CLIENT_MOUSEUP)
+	return ..()
+
+
+/client/MouseDrag(src_object, over_object, src_location, over_location, src_control, over_control, params) //The order seems to be wrong in the reference.
+	if(!src_control)
+		return
+	SEND_SIGNAL(src, COMSIG_CLIENT_MOUSEDRAG, over_object, over_location)
```
ninjanomnom: Ditto

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,129 @@
...
+	switch(.) //Here we give the chance for custom behavior.
+		if(NONE, COMSIG_CLIENT_MOUSEDOWN_INTERCEPT)
+			return //Null lets things progress. NONE lets the click pass .COMSIG_CLIENT_MOUSEDOWN_INTERCEPT doesn't.
+	src.target = target
+	start_autofiring()
+	return COMSIG_CLIENT_MOUSEDOWN_INTERCEPT
+
+
+/datum/component/full_auto_fire/proc/start_autofiring()
+	var/client/holder = parent
```
ninjanomnom: This kind of component should be on the gun not the client

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,129 @@
...
+	selected_frame = image('icons/mob/actions.dmi', null, "selected_frame")
+	selected_frame.appearance_flags = RESET_COLOR
+
+
+/datum/action/toggle_full_auto/action_activate()
+	if(!owner.client)
+		return
+	if(active)
+		var/datum/component/full_auto_fire/fullauto = owner.client.LoadComponent(/datum/component/full_auto_fire)
+		fullauto.fullauto_off()
```
ninjanomnom: Very no

Apply the component to the gun. The gun needs to handle creating its own action button and dealing with interactions with it.

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,129 @@
...
+
+
+/datum/action/toggle_full_auto/remove_action()
+	if(active)
+		action_activate()
+	return ..()
+
+
+//Full auto proc.
+/datum/proc/do_auto_fire()
```
ninjanomnom: What?

chaoko99:
Does this require the use of mousepolling to work? I kinda want to do something like this on another server.

Rohesie:
Not at all. But the implementation is not finished. I'd suggest waiting for it to be.
/tg/ and Eris already have full auto implementations, this one attempts to be more efficient than those, and less hacky.

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,222 @@
...
+		return
+	//COMSIG_HUMAN_GUN_FIRED would go here.
+	var/obj/screen/ammo/A = shooter.hud_used.ammo
+	A.update_hud(shooter) //Ammo HUD.
+
+
+/obj/item/weapon/gun/minigun/do_autofire(datum/source, atom/target, mob/shooter, params, shots_fired)
+	if(shooter.action_busy)
+		return
+	if(shots_fired == 1) //First shot.
```
Rohesie: Could actually offload this for `autofire_fail_check()`. Will do later.

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -583,6 +583,16 @@ should be alright.
 		return
 	src = G
 
+	if(flags_gun_features & GUN_HAS_FULL_AUTO) //No burst on full auto for now.
```
ninjanomnom: You can attach verbs to objects during runtime, the component should handle this instead

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -54,6 +54,7 @@
 	var/movement_acc_penalty_mult = 5				//Multiplier. Increased and decreased through attachments. Multiplies the accuracy/scatter penalty of the projectile when firing onehanded while moving.
 
 	var/fire_delay = 0							//For regular shots, how long to wait before firing again.
+	var/autofire_delay = 0.4 SECONDS			//Delay between automatic fire shots.
```
ninjanomnom: Component argument instead of var on the gun

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -911,6 +911,8 @@ and you're good to go.
 		playsound(user, dry_fire_sound, 25, 1, 5) //5 tile range
 	else
 		playsound(src, dry_fire_sound, 25, 1, 5)
+	SEND_SIGNAL(src, COMSIG_GUN_CLICKEMPTY)
```
ninjanomnom: Prefer to place signals as early as you can in procs they represent

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -1005,6 +1007,10 @@ and you're good to go.
 	return TRUE
 
 /obj/item/weapon/gun/proc/simulate_scatter(obj/item/projectile/projectile_to_fire, atom/target, turf/targloc, total_scatter_chance = 0, mob/user, burst_scatter_bonus = 0)
+	if(isnull(targloc)) //For debugging only.
```
ninjanomnom: :eyes:

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,222 @@
...
+	autofire_off()
+
+
+// There is a gun and there is a user wielding it. The component now waits for the mouse click.
+/datum/component/automatic_fire/proc/autofire_on(client/usercli)
+	clicker = usercli
+	RegisterSignal(clicker, COMSIG_CLIENT_MOUSEDOWN, .proc/on_mouse_down)
+	RegisterSignal(clicker.mob, COMSIG_MOB_LOGOUT, .proc/autofire_off)
+	RegisterSignal(parent, list(COMSIG_PARENT_QDELETED, COMSIG_ITEM_DROPPED), .proc/autofire_off)
+	parent.RegisterSignal(src, COMSIG_AUTOFIRE_STOPPED, /obj/item/weapon/gun/.proc/on_autofire_stop)
```
ninjanomnom: ...what? Just call the proc.

Also you should be passing in this as a callback during component init

code/modules/projectiles/updated_projectiles/gun_system.dm
Rohesie: This may require a larger refactor, by how the gun values are set.
I'll do that in a different PR.

code/modules/projectiles/updated_projectiles/gun_helpers.dm
Rohesie: This has been changed to a toggle between all the fire modes. It is also pending a slight tweak, but the value should stay in the gun if there are possible values unrelated to the automatic fire mode.

code/modules/projectiles/updated_projectiles/gun_system.dm
Rohesie: Happened once. But not since then. Likely due to an earlier runtime, but who knows?
The thing is now filled with debug messages because there's sometimes quirky behavior whose causes I haven't yet narrowed down.

code/datums/actions/item_action.dm
```diff
@@ -49,13 +49,33 @@
 /datum/action/item_action/firemode
 	var/action_firemode
 	var/obj/item/weapon/gun/holder_gun
+	var/action_icon
+	var/image/selected_frame
```
ninjanomnom: Can this be made static?

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,357 @@
...
+// VV wake_up() VV
+#define AUTOFIRE_STAT_IDLE 1 //Compatible firemode is in the gun. Wait until it's held in the user hands.
+// ^^ autofire_off() ^^
+// VV autofire_on() VV
+#define AUTOFIRE_STAT_ALERT	2 //Gun is active and in the user hands. Wait until user does a valid click.
+// ^^ stop_autofiring() ^^
+// VV start_autofiring() VV
+#define AUTOFIRE_STAT_FIRING 3 //Dakka-dakka-dakka.
+
+datum/component/automatic_fire
```
ninjanomnom:
```suggestion
/datum/component/automatic_fire
```

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,357 @@
+#define AUTOFIRE_STAT_SLEEPING 0 //Component is in the gun, but the gun is in a different firemode. Sleep until a compatible firemode is activated.
+// ^^ sleep_up() ^^
+// VV wake_up() VV
+#define AUTOFIRE_STAT_IDLE 1 //Compatible firemode is in the gun. Wait until it's held in the user hands.
+// ^^ autofire_off() ^^
+// VV autofire_on() VV
+#define AUTOFIRE_STAT_ALERT	2 //Gun is active and in the user hands. Wait until user does a valid click.
+// ^^ stop_autofiring() ^^
+// VV start_autofiring() VV
+#define AUTOFIRE_STAT_FIRING 3 //Dakka-dakka-dakka.
```
ninjanomnom: If these aren't going to be in a defines folder you need to undef them at the end of the file

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,357 @@
...
+		return COMPONENT_INCOMPATIBLE
+	RegisterSignal(parent, COMSIG_GUN_FIREMODE_TOGGLE, .proc/wake_up)
+	RegisterSignal(parent, COMSIG_GUN_FIREDELAY_MODIFIED, .proc/modify_firedelay)
+	RegisterSignal(parent, COMSIG_GUN_BURSTDELAY_MODIFIED, .proc/modify_burst_delay)
+	RegisterSignal(parent, COMSIG_GUN_BURSTAMOUNT_MODIFIED, .proc/modify_burst_amount)
+
+	var/obj/item/weapon/gun/shoota = parent
+	autofire_shot_delay = shoota.fire_delay
+	burstfire_shot_delay = shoota.burst_delay
+	shots_to_fire = shoota.burst_amount
```
ninjanomnom: Pass these in as initialize arguments

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,357 @@
...
+
+
+/datum/component/automatic_fire/proc/stop_autofiring(datum/source, atom/object, turf/location, control, params)
+	switch(autofire_stat)
+		if(AUTOFIRE_STAT_SLEEPING, AUTOFIRE_STAT_IDLE, AUTOFIRE_STAT_ALERT)
+			return
+	autofire_stat = AUTOFIRE_STAT_ALERT
+	if(auto_delay_timer) //Keep this at the top of the proc. If anything else runtimes or fails it would cause a potentially-infinite loop.
+		if(!deltimer(auto_delay_timer))
+			addtimer(CALLBACK(src, .proc/keep_trying_to_delete_timer, auto_delay_timer), 0.1 SECONDS) //Next tick hopefully.
```
ninjanomnom:
```suggestion
			addtimer(CALLBACK(src, .proc/keep_trying_to_delete_timer, auto_delay_timer), 1) //Next tick hopefully.
```

Really this should even work with 0

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,357 @@
...
+
+/obj/item/weapon/gun/proc/on_autofire_stop(shots_fired)
+	if(shots_fired > 1) //If it was a burst, apply the burst cooldown.
+		extra_delay = min(extra_delay+(burst_delay*2), fire_delay*3)
+
+
+/obj/item/weapon/gun/proc/autofire_bypass_check(datum/source, client/clicker, atom/target, turf/location, control, params)
+	if(clicker.mob.get_active_held_item() != src)
+		return COMSIG_AUTOFIRE_ONMOUSEDOWN_BYPASS
+	return NONE //No flags means success.
```
ninjanomnom: You don't need to return NONE on your own
```suggestion
```

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,357 @@
...
+	var/obj/item/projectile/projectile_to_fire = load_into_chamber(shooter)
+	if(!projectile_to_fire)
+		click_empty(shooter)
+		return NONE
+	var/dual_wield = FALSE
+	var/obj/item/weapon/gun/akimbo_gun = shooter.get_inactive_held_item()
+	if(istype(akimbo_gun))
+		dual_wield = TRUE
+		akimbo_gun.Fire(target, shooter, params, FALSE, dual_wield)
+	apply_bullet_effects(projectile_to_fire, shooter, shots_fired, FALSE, dual_wield)
```
ninjanomnom: Split up this proc into other more readable procs

code/datums/components/full_auto_fire.dm
Rohesie: This was meant more for clarity, but sure, it's functionally the same.

code/_globalvars/bitfields.dm
```diff
@@ -493,4 +493,10 @@ GLOBAL_LIST_INIT(bitfields, list(
 		"TURRET_IMMOBILE" = TURRET_IMMOBILE,
 		"TURRET_COOLDOWN" = TURRET_COOLDOWN
 		),
+	"autofire_stat" = list(
+		"AUTOFIRE_STAT_SLEEPING" = AUTOFIRE_STAT_SLEEPING,
+		"AUTOFIRE_STAT_IDLE" = AUTOFIRE_STAT_IDLE,
+		"AUTOFIRE_STAT_ALERT" = AUTOFIRE_STAT_ALERT,
+		"AUTOFIRE_STAT_FIRING" = AUTOFIRE_STAT_FIRING
+		),
```
DominikPanic:
```suggestion
		)
```
Muh microoptimization

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,362 @@
...
+			CRASH("Failed to get the turf under clickcatcher")
+	
+	if(SEND_SIGNAL(src, COMSIG_AUTOFIRE_ONMOUSEDOWN, source, target, location, control, params) & COMPONENT_AUTOFIRE_ONMOUSEDOWN_BYPASS)
+		return
+
+	source.click_intercepted = world.time //From this point onwards Click() will no longer be triggered.
+
+	switch(autofire_stat)
+		if(AUTOFIRE_STAT_SLEEPING, AUTOFIRE_STAT_IDLE)
+			CRASH("on_mouse_down() called with [autofire_stat] autofire_stat")
```
ninjanomnom: More unnecessary two case switches

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,362 @@
...
+	mouse_parameters = params
+	start_autofiring()
+
+
+//Dakka-dakka
+/datum/component/automatic_fire/proc/start_autofiring()
+	if(autofire_stat == AUTOFIRE_STAT_FIRING)
+		return
+	autofire_stat = AUTOFIRE_STAT_FIRING
+	if(auto_delay_timer)
```
ninjanomnom: Clean up this proc a bit and make it more readable

code/datums/components/full_auto_fire.dm
DominikPanic: 👀

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,362 @@
...
+		if(GUN_FIREMODE_SEMIAUTO, GUN_FIREMODE_BURSTFIRE)
+			switch(autofire_stat)
+				if(AUTOFIRE_STAT_IDLE, AUTOFIRE_STAT_ALERT, AUTOFIRE_STAT_FIRING)
+					sleep_up()
+			return //No need for autofire on this mode.
+	
+	component_fire_mode = fire_mode
+	
+	switch(autofire_stat)
+		if(AUTOFIRE_STAT_IDLE, AUTOFIRE_STAT_ALERT)
```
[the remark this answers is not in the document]
Rohesie: In here there are 4 options and 3 possible behaviors:
If 1 or 2, return.
If 3, stop autofiring and return.
If 4 go on.

Would require two `if` checks to substitute this, and I feel a `switch` is better.

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,362 @@
...
+	RegisterSignal(parent, COMSIG_ITEM_DROPPED, .proc/autofire_off)
+	RegisterSignal(shooter, COMSIG_MOB_LOGOUT, .proc/autofire_off)
+	parent.RegisterSignal(src, COMSIG_AUTOFIRE_ONMOUSEDOWN, /obj/item/weapon/gun/.proc/autofire_bypass_check)
+	parent.RegisterSignal(parent, COMSIG_AUTOFIRE_SHOT, /obj/item/weapon/gun/.proc/do_autofire)
+
+
+/datum/component/automatic_fire/proc/autofire_off(datum/source)
+	switch(autofire_stat)
+		if(AUTOFIRE_STAT_SLEEPING, AUTOFIRE_STAT_IDLE)
+			return
```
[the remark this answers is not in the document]
Rohesie: This would need to be replaced by two `if` checks:
```
if(autofire_stat & AUTOFIRE_STAT_SLEEPING|AUTOFIRE_STAT_IDLE)
	return
if(autofire_stat & AUTOFIRE_STAT_FIRING)
	stop_autofiring()
```
Feels like a `switch` is better. You sure you want this one changed?

code/datums/components/full_auto_fire.dm
Rohesie: Same here, the replacement would be two if checks:
```
if(autofire_stat & AUTOFIRE_STAT_SLEEPING|AUTOFIRE_STAT_IDLE)
	CRASH("on_mouse_down() called with [autofire_stat] autofire_stat")
if(autofire_stat & AUTOFIRE_STAT_FIRING)
	stop_autofiring()
```

code/datums/components/full_auto_fire.dm
```diff
@@ -0,0 +1,367 @@
...
+	autofire_off()
+	return ..()
+
+
+/datum/component/automatic_fire/proc/wake_up(datum/source, fire_mode, client/usercli)
+	switch(fire_mode)
+		if(GUN_FIREMODE_AUTOMATIC, GUN_FIREMODE_AUTOBURST)
+			component_fire_mode = fire_mode
+		else
+			if(autofire_stat & AUTOFIRE_STAT_IDLE|AUTOFIRE_STAT_ALERT|AUTOFIRE_STAT_FIRING)
```
ninjanomnom: You want to group bit operators in general when checking against values
```suggestion
			if(autofire_stat & (AUTOFIRE_STAT_IDLE|AUTOFIRE_STAT_ALERT|AUTOFIRE_STAT_FIRING))
```

code/datums/components/full_auto_fire.dm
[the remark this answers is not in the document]
ninjanomnom: Why is it better? Yes you get slight performance benefits with a switch but in most cases `if` statements are more readable and don't have as much indentation. It's fine if it's being used in this case but I'd really consider not doing it regardless because it requires more time invested in reading the code to understand exactly what you want.

PR #2129, opened 2019-06-28
General clothes and equipment vendors

Rohesie:
* Generalized the behavior of equipment vendors so we can have some that are not squad locked. The objective of this is to allow for general prep rooms like in the Sulaco, Pillar and Cant maps, and reduce squad hardcoding.

* Did some very minor updates on icons such as the ship engineer headset and the likes, which was using the same as Bravo, and the medical one, which just changed a few pixels.

* Re-added blue, purple and yellow glove sprites, which had been simply replaced by the marine ones. This is nothing but fluff.


code/game/objects/items/radio/headset.dm
```diff
@@ -349,9 +349,38 @@ GLOBAL_LIST_INIT(channel_tokens, list(
...
+				dat += " corpsman"
+				keyslot2 = new /obj/item/encryptionkey/med
+		name = dat + " radio headset"
+	return ..()
+
+
 /obj/item/radio/headset/almayer/marine/alpha
 	name = "marine alpha radio headset"
-	icon_state = "sec_headset"
+	icon_state = "Alpha_headset"
```
DominikPanic: capital names in icon states are usually a bad practice, think it would be easy to change?

Rohesie: It is done this way because it's set by `"[squad]_headset"`, with `squad` being capitalized.
It would just require the usage of `[lowertext(squad)]_headset`, which is trivial.
Want it done?

DominikPanic: Yeah that sounds better

Rohesie:
Should be ready.

PR #2131, opened 2019-06-28
Adds the ability to load and vote for ship maps.

DominikPanic:


code/game/objects/structures/crates_lockers/closets/secure/cm_closets.dm
```diff
@@ -25,9 +25,8 @@
 
 /obj/structure/closet/secure_closet/marine/PopulateContents()
 	new /obj/item/clothing/shoes/marine(src)
-	switch(SSmapping.config.map_name)
-		if(MAP_ICE_COLONY)
-			new /obj/item/clothing/mask/rebreather/scarf(src)
+	if(SSmapping.configs[GROUND_MAP].map_name == MAP_ICE_COLONY)
```
Rohesie: I believe the linter complains about this because despite BYOND accepts this, it treats it as `SSmapping.configs[GROUND_MAP]:map_name`, thus being an unsafe access.
The proper way to solve it is through this:
```
var/map = SSmapping.configs[GROUND_MAP]
if(map.map_name == MAP_ICE_COLONY)
```

Rohesie: (the variable will need to have the proper type, of course)

DominikPanic: Currently we're not using the linter but if you really want I can add it there (though it will lead to a quite a lot of code added)

PR #2132, opened 2019-06-28
Reorganizes the controllers folder

DominikPanic:
All legacy controllers are now gone, time to make this folder like the infamous /tg/ standard


code/controllers/failsafe.dm
```diff
@@ -1,102 +1,102 @@
...
-		statclick = new/obj/effect/statclick/debug(null, "Initializing...", src)
-
-	stat("Failsafe Controller:", statclick.update("Defcon: [defcon_pretty()] (Interval: [Failsafe.processing_interval] | Iteration: [Failsafe.master_iteration])"))
+/**
+* Failsafe
+*
+* Pretty much pokes the MC to make sure it's still alive.
+**/
+
+GLOBAL_REAL(Failsafe, /datum/controller/failsafe)
```
Rohesie: Think you could do this again without the CRLF/LF changes?

DominikPanic: Sure I'll try

PR #2134, opened 2019-06-28
Makes chat loading even more smooth, removes unused defines

DominikPanic:
By making chat init last it will only fire when everything is ready thus nothing will block the init of it now instead of it having the default init order.


Rohesie:
Maybe add a comment stating it should be last, and the reasons for it?

PR #2139, opened 2019-06-28
Fixes init runtime with keybindings

psykzz:
Fixes these 2 runtimes because these are just types and not meant to be initialized.


psykzz:
the `typesof` was removing every keybinding so reverted back to `list()` which works
image

PR #2140, opened 2019-06-28
Fixes ventcrawl lag

DominikPanic:


Rohesie:
Holy heck, this is unoptimized.

DominikPanic:
> 
> 
> Holy heck, this is unoptimized.

Still more optimized than the previous solution. It was literally calling image() on every single machine in the pipenet on every move.

Rohesie:
This is how /tg/ does it, but in the long term we should really refactor this.
Another stripe for the tiger, I guess.

PR #2149, opened 2019-06-28
Disables xeno basketball behind fun_allowed config

Rohesie:
## Changelog
:cl:
del: Xenos can no longer play basketball unless fun_allowed is turned on. And, by default, fun is not allowed.
/:cl:


Shadow-Quill:
Badmin :newspaper2:

Would Space Jam Drone want this?

PR #2156, opened 2019-06-29
Bodybag/Cryobag/Tarp Rework

### LINKED ISSUE #2152: camo tarp doesn't work
tarp stuck

Rohesie:
closes #2152

* Refactored bodybags, cryobags and tarps, removing useless processing and reducing qdels to lighten SSgarbage.
* Added a mob death signal, so things can react to them. Cryobags and tarps do, for example, opening if the mob dies inside them.
* What's in the changelog.

## Changelog
:cl:
del: Removed an unknown feature: cryobags actually had a lifetime of around an hour of holding mobs inside them. Maybe more.
balance: Tarps now fade out gradually instead of in janky steps. They don't accept dead bodies in them anymore (and will spit out whoever dies inside them). You can now drag them even if closed, as long as they don't contain someone. They will no longer show the name of the person inside.
/:cl:

### DIFF
--- a/code/game/objects/items/bodybag.dm
+++ b/code/game/objects/items/bodybag.dm
@@ -257,10 +257,10 @@
 	. = ..()
 	if(bodybag_occupant)
 		bodybag_occupant.in_stasis = STASIS_IN_BAG
-		RegisterSignal(bodybag_occupant, list(COMSIG_MOB_DEATH, COMSIG_PARENT_QDELETED), .proc/on_stasis_mob_death)
+		RegisterSignal(bodybag_occupant, list(COMSIG_MOB_DEATH, COMSIG_PARENT_QDELETED), .proc/on_bodybag_occupant_death)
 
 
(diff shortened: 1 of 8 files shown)


code/game/objects/items/bodybag.dm
```diff
@@ -108,41 +118,49 @@
 		overlays.Cut()
 
 
-/obj/structure/closet/bodybag/closet_special_handling(mob/living/mob_to_stuff) // overriding this
+/obj/structure/closet/bodybag/closet_special_handling(mob/living/carbon/human/mob_to_stuff) // overriding this
```
DominikPanic: This will get called on any living and will likely runtime if you're accessing human stuff

Rohesie: The first check is a type check.

DominikPanic: That still breaks the practice that arguments in procs should be the same, especially since it's `ishuman` rather than `istype`
```suggestion
/obj/structure/closet/bodybag/closet_special_handling(mob/living/mob_to_stuff) // overriding this
```

PR #2157, opened 2019-06-29
Improves embryo logging

DominikPanic:


Rohesie:
Embryo removed via autodoc, manual surgery or aheal perhaps?

PR #2159, opened 2019-06-29
Balance boiler bombard

psykzz:
## About The Pull Request

This changes the timing of the boiler bombard ability

Long range: 2sec -> 1sec
Toggle ability: 3sec -> 4sec
Fire ability: 5sec -> 3sec

## Why It's Good For The Game

Boiler is in a really bad spot right now, this should help balancing by allow it to launch attacks faster after clicking, but generally keeps the cooldown around the same.

We removed 2seconds from the whole chain because of a stealth nerf during the xeno ability refactor. As you can now only perform one action at a time, previously you could toggle long range and toggle the ability at the same time.

## Changelog
:cl:
balance: rebalanced boiler bombard, overall reduced 2 seconds from total time needed to fire, but also frontloaded most of the wait to toggling the ability, and reduced the wait when firing.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
@@ -18,7 +18,7 @@
 	else
 		X.visible_message("<span class='notice'>[X] starts looking off into the distance.</span>", \
 			"<span class='notice'>We start focusing your sight to look off into the distance.</span>", null, 5)
-		if(!do_after(X, 20, FALSE, null, BUSY_ICON_GENERIC) || X.is_zoomed)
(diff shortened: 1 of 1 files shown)


psykzz:
I know about whitespace but it was wrong to start with, you can make your reviewing easier if you use https://github.com/tgstation/TerraGov-Marine-Corps/pull/2159/files?w=1

Rohesie:
Fine to test without extensive testing. The balance change is not huge.

drawsstuff:
What makes boilers particularly bad is a combination of:

1. Slowness making it impossible to maintain a good distance
2. Vulnerability to incendiary weapons
3. Glowing in the dark making it a fat, juicy target

While this buff is nice, I'm not sure it addresses the big issues that make boiler so bad to begin with.

PR #2162, opened 2019-06-29
Allows the queen to override shuttle lockdown

DominikPanic:
:cl: LaKiller8
balance: The queen can now override the planetside shuttle lockdown if all doors are locked after 60 seconds of standing still.
/:cl:

### DIFF
--- a/code/modules/shuttle/marine_dropship.dm
+++ b/code/modules/shuttle/marine_dropship.dm
@@ -246,6 +246,8 @@
 #define ALIVE_HUMANS_FOR_CALLDOWN 0.1
 
 /datum/game_mode/proc/can_summon_dropship(mob/user)
+	if(user.action_busy)
+		return FALSE
 	if(SSticker.round_start_time + SHUTTLE_HIJACK_LOCK > world.time)
 		to_chat(user, "<span class='warning'>It's too early to call it. We must wait [DisplayTimeText(SSticker.round_start_time + SHUTTLE_HIJACK_LOCK - world.time, 1)].</span>")
 		return FALSE
@@ -279,8 +281,6 @@
 			return FALSE
 		if(locked_sides < 3)
 			to_chat(user, "<span class='warning'>At least one side is still unlocked!</span>")
-			return FALSE
-		if(user.action_busy)
 			return FALSE
 		to_chat(user, "<span class='warning'>We begin overriding the shuttle lockdown. This will take a while...</span>")
 		if(!do_after(user, 60 SECONDS, FALSE, null, BUSY_ICON_DANGER, BUSY_ICON_DANGER))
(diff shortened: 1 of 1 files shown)


code/modules/shuttle/marine_dropship.dm
```diff
@@ -255,7 +255,38 @@
...
+				continue
+			locked_sides++
+			break
+		if(!locked_sides)
+			to_chat(user, "<span class='warning'>We can't call the bird from here!</span>")
+			return FALSE
+		if(locked_sides < 3)
+			to_chat(user, "<span class='warning'>At least one side is still unlocked!</span>")
+			return FALSE
+		to_chat(user, "<span class='warning'>You begin overriding the shuttle lockdown. This will take a while.</span>")
```
Rohesie:
```suggestion
		to_chat(user, "<span class='warning'>We begin overriding the shuttle lockdown. This will take a while...</span>")
```

code/modules/shuttle/marine_dropship.dm
```diff
@@ -255,7 +255,38 @@
...
+		if(!do_after(user, 60 SECONDS, FALSE, null, BUSY_ICON_DANGER, BUSY_ICON_DANGER))
+			to_chat(user, "<span class='warning'>You stop overriding the shuttle lockdown.</span>")
```
Rohesie:
```suggestion
			to_chat(user, "<span class='warning'>We cease overriding the shuttle lockdown.</span>")
```

code/modules/shuttle/marine_dropship.dm
```diff
@@ -255,7 +255,38 @@
...
+			return FALSE
+		D.hijack_state = HIJACK_STATE_CALLED_DOWN
+		D.unlock_all()
+		to_chat(user, "<span class='warning'>You override the shuttle lockdown.</span>")
```
Rohesie:
```suggestion
		to_chat(user, "<span class='warning'>We have overriden the shuttle lockdown!</span>")
```

Rohesie: Maybe a little roar sound.

Rohesie:
Request still pending. If done I can merge.

DominikPanic:
> 
> 
> Request still pending. If done I can merge.

The change request has already been done.

Rohesie:
This one:
"Since you are adding a do_after, add a return FALSE check if the user's action_busy is TRUE, at the start of the proc."

DominikPanic:
> 
> 
> This one:
> "Since you are adding a do_after, add a return FALSE check if the user's action_busy is TRUE, at the start of the proc."

There's no point doing it much sooner than the do-after, but if you want me to move it up I will.

DominikPanic:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/2162/commits/5b1d6a910270545a98ee8ea852a0c1aefb19e011#diff-2ee49cc9969facc782293d6bc10859eeR283

Rohesie:
When you leave a proc with a `do_after` without an `busy_action` check at start, it can be clicked multiple times and start several timed action processes.

PR #2174, opened 2019-06-30
Hopefully fixes the pref saving bad output runtime

DominikPanic:
This is the only place that could have realistically been causing this, leave it up for ~2 days, if it doesn't reappear, merge it.


DominikPanic:
@Rohesie Runtime hasn't appeared since (previously it was like once per 2 rounds)

PR #2176, opened 2019-06-30
Adds argument definitions to contribution guidelines

DominikPanic:


.github/CONTRIBUTING.md
```diff
@@ -216,6 +216,25 @@ Instead of this you should;
...
+/atom/movable/do_a_thing(mob/living/carbon/human/user)
+	do something
+````
+Instead of this you should do it like this:
+````DM
+/atom/proc/do_a_thing(mob/user)
+	do something
+
+/atom/movable/do_a_thing(mob/user)
+	optional istype() check
```
Rohesie:
```suggestion
	var/mob/living/carbon/human/human_user = user
```

DominikPanic: removing the istype there would be misleading since this would be basically assuming the person is human rather than checking it

DominikPanic: also the istype is only optional if you need a different type than /mob

Rohesie: If the proc is changing the type, it's because it needs to be typecast. The type check is assumed, and may not be necessary.

.github/CONTRIBUTING.md
```diff
@@ -216,6 +216,26 @@ Instead of this you should;
...
+	do something
+````
+Instead of this you should do it like this:
+````DM
+/atom/proc/do_a_thing(mob/user)
+	do something
+
+/atom/movable/do_a_thing(mob/user)
+	if istype(user, mob/living/carbon/human)
+	typecast
```
Rohesie: If you use the human example, then give a correct example here, indented.
Else do it wholly generic.

.github/CONTRIBUTING.md
```diff
@@ -216,6 +216,26 @@ Instead of this you should;
...
+/atom/movable/do_a_thing(mob/living/carbon/human/user)
+	do something
+````
+Instead of this you should do it like this:
+````DM
+/atom/proc/do_a_thing(mob/user)
+	do something
+
+/atom/movable/do_a_thing(mob/user)
+	if istype(user, /mob/living/carbon/human)
```
Rohesie:
```suggestion
	if !(istype(user, /mob/living/carbon/human))
		return
```
Pseudocode is weird, but indentation and all.

Issue #2184, opened 2019-07-01
game thinks you're DNR if you move out of your corpse


factoryman942:
round 649
If I moved out of my corpse, I could see my hud icon in ghost view as a lightning bolt for a second, then it turned to a skull. Going back in and out made the icon swap between the two, and an attempt to defib me while I was outside of my body said I was DNR and didn't give me the prompt to move back to my body because someone was trying to defib.
testmerges #2172 #2174 #2125 #2162 #2161 #1931 #2169 #2168 #2163 #1685 #2179 #2178 #2177 #2180

Rohesie:
Started a PR to fix this, thinking it was a HUD issue, but this is about there no longer being a reference to the ghost in the dead mob. The `ckey` and `mind` references are erased when someone ghosts, and the `mind` one was the old method to link both.
I'll wait for @DominikPanic to give his input. We'll see what kind of solution he wants for this.

DominikPanic:
The mind reference should get erased when someone ghosts, otherwise a lot of quirky behavior will happen and people will show up as xenos. If they are not DNR their mind.current variable will still point to the non-ghost body.

PR #2186, opened 2019-07-01
get_ghost() cleanup

Rohesie:
Micro cleanup. Started it because of #2184 but then saw the issue was a bit deeper.

### DIFF
--- a/code/game/objects/items/defibrillator.dm
+++ b/code/game/objects/items/defibrillator.dm
@@ -72,7 +72,7 @@
 		return FALSE
 	for(var/g in GLOB.observer_list)
 		var/mob/dead/observer/ghost = g
-		if(G.mind != mind)
+		if(ghost.mind != mind)
 			continue
 		if(ghost.client && ghost.can_reenter_corpse)
 			return ghost
--- a/code/game/objects/items/defibrillator.dm
+++ b/code/game/objects/items/defibrillator.dm
@@ -68,7 +68,7 @@
 
 
 /mob/living/carbon/human/proc/get_ghost()
-	if(!mind || client) //Let's call up the correct ghost!
+	if(client) //Let's call up the correct ghost!
 		return FALSE
 	for(var/g in GLOB.observer_list)
 		var/mob/dead/observer/ghost = g
--- a/code/game/objects/items/defibrillator.dm
+++ b/code/game/objects/items/defibrillator.dm
@@ -72,7 +72,7 @@
 		return FALSE
 	for(var/g in GLOB.observer_list)
 		var/mob/dead/observer/ghost = g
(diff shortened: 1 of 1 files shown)


code/game/objects/items/defibrillator.dm
```diff
@@ -66,14 +66,18 @@
...
+
 /mob/living/carbon/human/proc/get_ghost()
-	if(mind && !client) //Let's call up the correct ghost! Also, bodies with clients only, thank you.
-		for(var/mob/dead/observer/G in GLOB.player_list)
-			if(G.mind == mind)
-				var/mob/dead/observer/ghost = G
-				if(ghost && ghost.client && ghost.can_reenter_corpse)
-					return ghost
-	return 0
+	if(!mind || client) //Let's call up the correct ghost!
```
DominikPanic: See my comment on the bug report, this won't work.

Rohesie: Yep, the issue is deeper. Was waiting on your input. This PR only refactors the currently non-working proc.

The problem is the lack of a reference to the ghost in the mob at the moment.
Suggestions?

DominikPanic: Like I said. remove the mind check, check if ghost.current == src

DominikPanic:
```suggestion
	if(client) //Let's call up the correct ghost!
```

code/game/objects/items/defibrillator.dm
```diff
@@ -66,14 +66,18 @@
...
+		return FALSE
+	for(var/g in GLOB.observer_list)
+		var/mob/dead/observer/ghost = g
+		if(ghost.mind != mind)
```
DominikPanic:
```suggestion
		if(ghost.current != src)
```

code/game/objects/items/defibrillator.dm
```diff
@@ -66,14 +66,18 @@
...
-				if(ghost && ghost.client && ghost.can_reenter_corpse)
-					return ghost
-	return 0
+	if(client) //Let's call up the correct ghost!
+		return FALSE
+	for(var/g in GLOB.observer_list)
+		var/mob/dead/observer/ghost = g
+		if(ghost.mind.current != src)
+			continue
+		if(ghost.client && ghost.can_reenter_corpse)
```
psykzz: does it matter to check the if the clients are the same?

DominikPanic: There is no check for clients being the same?

psykzz: You are right, I was being dense. 
Ignore me :)

PR #2187, opened 2019-07-01
Ranks turned into defines

Rohesie:
This would have made the ranks-name-changing PR we did before much easier. Also good so people don't have to remember that "Master at Arms" has a lowercase-starting word.

If one of these change, they should all change.


code/__DEFINES/jobs.dm
```diff
@@ -24,16 +24,41 @@
...
+#define MEDICAL_RESEARCHER "Medical Researcher"
+#define SQUAD_LEADER "Squad Leader"
+#define SQUAD_SPECIALIST "Squad Specialist"
+#define SQUAD_SMARTGUNNER "Squad Smartgunner"
+#define SQUAD_CORPSMAN "Squad Corpsman"
+#define SQUAD_ENGINEER "Squad Engineer"
+#define SQUAD_MARINE "Squad Marine"
+
+
+#define JOBS_COMMAND 		list(CAPTAIN, FIELD_COMMANDER, INTELLIGENCE_OFFICER, PILOT_OFFICER, REQUISITIONS_OFFICER, CHIEF_SHIP_ENGINEER, CHIEF_MEDICAL_OFFICER, SYNTHETIC, COMMAND_MASTER_AT_ARMS)
```
spookydonut: make these GLOBs so you arent instancing a new list everytime these defines are used.

code/datums/jobs/access.dm
```diff
@@ -263,27 +263,27 @@ proc/get_all_job_icons()
 
 /proc/get_marine_jobs()
 		return list(
-				"Captain",
```
spookydonut: same here just make a GLOB so you arent instancing a new list everytime this proc is called

Rohesie: Done.

Rohesie: Done.

code/datums/jobs/job/job.dm
```diff
@@ -1,11 +1,11 @@
 GLOBAL_LIST_INIT(exp_jobsmap, list(
```
spookydonut: move this to make_datum_refs since you're referencing a GLOB inside a GLOBAL_LIST_INIT, works now but its a ticking time bomb

code/game/objects/effects/landmarks/corpsespawner.dm
```diff
@@ -29,7 +29,7 @@
 	var/corpseback = null
 	var/corpseid = 0     //Just set to 1 if you want them to have an ID
 	var/corpseidjob = null // Needs to be in quotes, such as "Clown" or "Chef." This just determines what the ID reads as, not their access
-	var/corpseidaccess = null //This is for access. See access.dm for which jobs give what access. Again, put in quotes. Use "Captain" if you want it to be all access.
+	var/corpseidaccess = null //This is for access. See access.dm for which jobs give what access. Again, put in quotes. Use CAPTAIN if you want it to be all access.
```
psykzz:
```suggestion
	var/corpseidaccess = null //This is for access. See access.dm for which jobs give what access. Use CAPTAIN if you want it to be all access.
```

code/datums/jobs/job/job.dm
Rohesie: As said on discord, these requests are getting quite out of scope.

PR #2188, opened 2019-07-01
Fixes an issue with xeno cooldown abilities

Rohesie:
Also removes hopefully the last runtime from the Shrike.

In short, `add_cooldown()` was returning if there were any `active_timers`. Which meant that an ability with a `CALLBACK` would simply fail to add cooldown.
Since `on_cooldown` will always be `TRUE` when the timer is running and `FALSE` otherwise, this will avoid the runtimes the original check wanted to.

Fixed a few other places where it would take or delete any timer disrespecting ids.

### DIFF
--- a/code/datums/actions/xeno_action.dm
+++ b/code/datums/actions/xeno_action.dm
@@ -114,13 +114,17 @@
 /datum/action/xeno_action/proc/action_cooldown_check()
 	return !on_cooldown
 
+
 /datum/action/xeno_action/proc/clear_cooldown()
-	for(var/timer in active_timers)
-		qdel(timer)
-		on_cooldown_finish()
+	if(!cooldown_id)
+		return
+	deltimer(cooldown_id)
+	on_cooldown_finish()
+
 
 /datum/action/xeno_action/proc/get_cooldown()
 	return cooldown_timer
+
 
 /datum/action/xeno_action/proc/add_cooldown()
 	if(on_cooldown) // stop doubling up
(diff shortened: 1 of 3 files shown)


code/datums/actions/xeno_action.dm
```diff
@@ -114,27 +114,31 @@
...
 /datum/action/xeno_action/proc/cooldown_remaining()
-	for(var/i in active_timers)
-		var/datum/timedevent/timer = i
-		return (timer.timeToRun - world.time) * 0.1
-	return 0
+	return timeleft(cooldown_id) * 0.1
+
 
 //override this for cooldown completion.
 /datum/action/xeno_action/proc/on_cooldown_finish()
```
spookydonut: clear the cooldown_id here

Rohesie: Removed `on_cooldown` in favor of `cooldown_id`

code/datums/actions/xeno_action.dm
```diff
@@ -112,33 +111,36 @@
...
-	on_cooldown = FALSE
+	cooldown_id = null
```
DominikPanic:
```suggestion
	qdel(cooldown_id)
```
to actually stop the timer properly

Rohesie: This proc is called by `clear_cooldown()`, which stops the timer.

Rohesie: May be safer to clear it there, though...

DominikPanic: then it should be done like that there, it's bad to split these two lines apart

spookydonut: this suggestion is wrong since on_cooldown_finish is called when the timer ends

spookydonut: This is coded correctly no need to change anything now.

PR #2191, opened 2019-07-01
send_signal wrapper

Rohesie:
Let's check with @ninjanomnom if there's no better way, without having to pass the arguments as a list.
If not, this works, still.

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1159,5 +1159,9 @@ will handle it, but:
 	addtimer(CALLBACK(GLOBAL_PROC, /.proc/remove_images_from_clients, I, show_to), 0.5 SECONDS)
 
 
-/proc/send_global_signal(signal) //Wrapper for callbacks and the likes.
-	SEND_GLOBAL_SIGNAL(signal)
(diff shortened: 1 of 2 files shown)


code/__HELPERS/unsorted.dm
```diff
@@ -1159,5 +1159,9 @@ will handle it, but:
...
 
 
-/proc/send_global_signal(signal) //Wrapper for callbacks and the likes.
-	SEND_GLOBAL_SIGNAL(signal)
\ No newline at end of file
+/proc/send_global_signal(datum/signal, ...) //Wrapper for callbacks and the likes.
+	SEND_GLOBAL_SIGNAL(signal, args) //Passes the arguments as a list.
+
+
+/proc/send_signal(datum/source, signal, ...) //Wrapper for callbacks and the likes.
```
DominikPanic: All signals already pass their arguments properly, that's why components.dm contains the comments like this
`//from base of atom/Click(): (location, control, params, mob/user)`
This makes the entire PR invalid

spookydonut: you're wrong, you can't make callbacks with macros.

ninjanomnom:
You're going to need to remake the macro as an actual proc rather than trying to use it inside here for a callback

PR #2193, opened 2019-07-01
unfuck paygrades

spookydonut:
either this or just collapse them into the job datum but some share the same paygrades.

this is unfinished.


psykzz:
Can we not just purge this instead?

If its a prefix for the radio, then lets call it a radio prefix, It's got nothing to do with a paygrade and is honestly just confusing.

PR #2196, opened 2019-07-02
Adds weed decay

psykzz:
## About The Pull Request

This adds that when a weed node is destroyed the weeds connected within its graph slowly die out.
This happens every 3rd tick of the subsystem which makes decay much slower than growth.

Also adds some changes to ensure new nodes take over orphaned weeds, so they can continue to live on.

## Why It's Good For The Game

This helps combat weed creep, and reduces the power of the weeds.

## Changelog
:cl:
balance: Weeds will now decay over time if their node is destroyed, at a rate of about 3 times slower than the growth speed.
/:cl:

---

https://streamable.com/4d9jq

### DIFF
--- a/code/controllers/subsystem/weeds.dm
+++ b/code/controllers/subsystem/weeds.dm
@@ -7,10 +7,7 @@ SUBSYSTEM_DEF(weeds)
 	// This is a list of nodes on the map.
 	var/list/creating = list()
 	var/list/pending = list()
-	var/list/decaying = list()
 	var/list/currentrun
-
-	var/decay_tick = 1
 
 /datum/controller/subsystem/weeds/stat_entry()
 	return ..("Nodes: [length(pending)]")
@@ -19,37 +16,6 @@ SUBSYSTEM_DEF(weeds)
 	if(!resumed)
 		currentrun = pending.Copy()
 		creating = list()
-
-	// Decay weeds first, but only ever 3rd tick (x3 slower than growing weeds)
-	if ((decay_tick % 4) == 0)
-		decay_tick = 1
-		for(var/A in decaying)
-			if(MC_TICK_CHECK)
-				return
-
-			var/turf/T = A
-			if(QDELETED(T))
-				decaying -= T
-				continue
-
-			var/obj/effect/alien/weeds/W = locate() in T
-			if(QDELETED(W) || W.parent_node)
-				decaying -= T
-				continue
-
-			var/decay_chance = 100
-			for(var/direction in GLOB.cardinals) 
-				var/turf/adj = get_step(T, direction)
-				if(locate(/obj/effect/alien/weeds) in adj)
-					decay_chance -= rand(25, 40)
-
-			if(decay_chance > 0 && prob(decay_chance))
-				W.parent_node = null // So it wont try to regrow
-				addtimer(CALLBACK(GLOBAL_PROC, .proc/qdel, W), rand(0, 7 SECONDS))
-				decaying -= T
-	else
-		decay_tick++
-
 
 	for(var/A in currentrun)
 		if(MC_TICK_CHECK)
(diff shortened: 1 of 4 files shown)


code/controllers/subsystem/weeds.dm
```diff
@@ -17,7 +20,41 @@ SUBSYSTEM_DEF(weeds)
 		currentrun = pending.Copy()
 		creating = list()
 
+	// Decay weeds first, but only ever 3rd tick (x3 slower than growing weeds)
+	if ((decay_tick % 4) == 0)
```
spookydonut: just put this in a different subsystem, like SSweeddecay

psykzz:
This requires a full review but assuming all good, then mergable.
I can move subsystems as @spookydonut proposed i dont mind. but do think its a little wasteful.

spookydonut:
doing the decay_tick thing doesnt mesh well with a resumable subsystem, its much better to just make a weed decay subsystem that ticks at a different rate.

Rohesie:
If the SS doesn't have a list of things to process at all times and only gets stuff added on demand, then it won't really be costly.

psykzz:
@spookydonut @Rohesie new subsystem for decay has been added should be good to review & TM if your happy

code/controllers/subsystem/weeds_decay.dm
```diff
@@ -0,0 +1,51 @@
...
+		var/obj/effect/alien/weeds/W = locate() in T
+		if(QDELETED(W) || W.parent_node)
+			decaying -= T
+			continue
+
+		var/decay_chance = 100
+		for(var/direction in GLOB.cardinals) 
+			var/turf/adj = get_step(T, direction)
+			if(locate(/obj/effect/alien/weeds) in adj)
+				decay_chance -= rand(25, 40)
```
Rohesie: `decay_chance` is already a probability value. Don't stack probability over probability. There's 9 possible adjacent weeds, so -10% each will still keep a low 10% chance of happening on a healthy but nodeless network.

code/controllers/subsystem/weeds_decay.dm
```diff
@@ -0,0 +1,51 @@
...
+
+		if(decay_chance > 0 && prob(decay_chance))
+			W.parent_node = null // So it wont try to regrow
+			addtimer(CALLBACK(GLOBAL_PROC, .proc/qdel, W), rand(0, 7 SECONDS))
```
Rohesie: You are already adding a % chance of it not happening. Is the timer needed on top of it?

code/controllers/subsystem/weeds_decay.dm
```diff
@@ -0,0 +1,51 @@
...
+			decaying -= T
+			continue
+
+		var/decay_chance = 100
+		for(var/direction in GLOB.cardinals) 
+			var/turf/adj = get_step(T, direction)
+			if(locate(/obj/effect/alien/weeds) in adj)
+				decay_chance -= rand(25, 40)
+
+		if(decay_chance > 0 && prob(decay_chance))
```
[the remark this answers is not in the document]
psykzz: TIL

code/controllers/subsystem/weeds_decay.dm
Rohesie: Doesn't that already happen with the % chance you introduced before?

code/controllers/subsystem/weeds_decay.dm
psykzz: Im only checking cardinals because it made it a little better for the check (and was also how it grows), so instead of 10% ill do 25% each time

code/controllers/subsystem/weeds_decay.dm
psykzz: yeah but the loop will process all exteria weeds at once and they might all get marked for removal, wtihout this rnad they all get removed at the same time, this just stutters the removal for visual effects.

Rohesie: Alright, makes sense.

code/controllers/subsystem/weeds_decay.dm
```diff
@@ -0,0 +1,51 @@
...
+		var/obj/effect/alien/weeds/W = locate() in T
+		if(QDELETED(W) || W.parent_node)
+			decaying -= T
+			continue
+
+		var/decay_chance = 100
+		for(var/direction in GLOB.cardinals) 
+			var/turf/adj = get_step(T, direction)
+			if(locate(/obj/effect/alien/weeds) in adj)
+				decay_chance -= 25
```
Rohesie: Maybe something between 20 and 24 so there's a small chance for weeds in the middle of a large pattern to die as well?

PR #2197, opened 2019-07-03
Fire() refactor part I: Non-projectile gun attachables handled by CTRL+Click

Rohesie:
The final goal is to implement automatic fire by pressing the mouse and moving it around. And also cleanup the fire code some.
A few features happening in the meantime.
There's extra unrelated changes, such as the deletion of an useless file and a yet-unused signal Spooky was using for his flamer component.
The idea is to try to make it compatible as much as possible.

## Changelog
:cl:
balance: UGLs and miniflamers are now shot via ctrl+click when the attachment is active.
/:cl:


Rohesie:
Separated the cooldown checks out of `able_to_fire()` to have a little more control over it.

DominikPanic:
Poke me when you want it tested

Rohesie:
This is tested locally. It can be tested live anytime.

Rohesie:
A single little mass-replacement in where you think you don't change the logic, and it screws up. Should be a lesson for never thinking you can't not-test.

Was tested for one round live and it worked good, after the patch.

PR #2200, opened 2019-07-03
Fixes explosions not working

DominikPanic:
This one is one me :c
Fixes: #2093


code/game/objects/explosion.dm
```diff
@@ -110,6 +110,9 @@
 
 			if(dist > 0)
 				T.ex_act(dist)
+				for(var/i in T.contents)
```
Rohesie:
```suggestion
				for(var/i in T)
```
Contents is implied.

PR #2202, opened 2019-07-03
IC information on enemy presence in Distress Signal game mode.

Rohesie:
This goes in line with reducing the need of first-contact-related rules.
Both sides are to expect enemy presence incoming and prepare for it.

Also moves some variables that were on the wrong file.

:cl:
balance: A bioscan now happens at the start of the game, providing rough information for both sides on what they will be facing and giving them a chance to prepare. No locations are shown, but the scanner is a bit more precise, giving only a +/- 1 variation.
/:cl:


Pariah919:
Won't this show survivor locations at spawn? This could make survivor even more unplayable really.

Rohesie:
Addressed that. Xenos will know more or less how many survivors there are, with the delta variation included, but not their location.

PR #2206, opened 2019-07-04
flamethrower component

spookydonut:
now with no reagent changes


spookydonut:
Still has some bugs and unimplemented functionality but basically finished.

PR #2207, opened 2019-07-04
Service pistol balance patch

MoXuT0:
## About The Pull Request

Renamed M4A3 (.45) to M1911 as it should be, increased it's mag size to 10. Regular service pisol mag size also increased to 14 (24 for extended mag). VP70 magazine now also uses its original sprite, not .45 one.

## Changelog
:cl:
tweak: M4A3 (.45) is now named M1911.
balance: Increased magazine size for M4A3 up to 12 (24 for extended).
balance: Increased magazine size for M1911 up to 10.
imageadd: 88Mod4 aka VP70 magazine now uses it's own sprite.
/:cl:


Pariah919:
Why would the m1911 be used so far into the future? Doesn't make much sense. Nor does it really look like one. 
You're also making the FC gun hold 10 rounds in it.

Rohesie:
The FC handgun could use a buff.

MoXuT0:
FC gun is literally same as regular M1911, it just fires 0.5 seconds faster and slightly accurate when unwielded. I thought renaming it to M4A4 but not sure about it, people may confuse it with regular pistol.

Pariah919:
The FCs handgun should either have better ammo that the FC can access or just get a higher general damage bonus.

MoXuT0:
It's already deals around 45 damage per shot, no need to increase it further. Auto eject is tradeoff for faster firing speed and better accuracy, it also fits that FC gun is very old one.

PR #2209, opened 2019-07-04
Second part of the mob cleanup - attack hand is now living+

DominikPanic:
Ghosts have their dedicated one, no dead mobs have hands. Will lessen the mess that the defines are.

### DIFF
--- a/code/modules/clothing/shoes/marine_shoes.dm
+++ b/code/modules/clothing/shoes/marine_shoes.dm
@@ -22,11 +22,11 @@
 	. = ..()
 
 //ATTACK HAND IGNORING PARENT RETURN VALUE
-/obj/item/clothing/shoes/marine/attack_hand(mob/living/M)
-	if(knife && src.loc == M && !M.incapacitated()) //Only allow someone to take out the knife if it's being worn or held. So you can pick them up off the floor
-		if(M.put_in_active_hand(knife))
-			to_chat(M, "<span class='notice'>You slide [knife] out of [src].</span>")
-			playsound(M, 'sound/weapons/gun_shotgun_shell_insert.ogg', 15, 1)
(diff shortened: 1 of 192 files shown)


code/modules/clothing/shoes/marine_shoes.dm
```diff
@@ -22,11 +22,11 @@
 	. = ..()
 
 //ATTACK HAND IGNORING PARENT RETURN VALUE
-/obj/item/clothing/shoes/marine/attack_hand(mob/living/M)
-	if(knife && src.loc == M && !M.incapacitated()) //Only allow someone to take out the knife if it's being worn or held. So you can pick them up off the floor
-		if(M.put_in_active_hand(knife))
-			to_chat(M, "<span class='notice'>You slide [knife] out of [src].</span>")
-			playsound(M, 'sound/weapons/gun_shotgun_shell_insert.ogg', 15, 1)
+/obj/item/clothing/shoes/marine/attack_hand(mob/living/user)
```
spookydonut: this should be done by signals

DominikPanic: Out of scope. This is just a mass replace.

spookydonut: arent you trying to unfuck attack_hand? should be in scope to make this call parent, else you miss out on stuff like add_fingerprints etc

DominikPanic: No, I'm purely changing the argument it gets called with since it's always a living.

Rohesie: Signals would be positive, but this change is not negative nor adds technical debt.
We can't force contributors to do more than they want if they are not introducing bad code.

PR #2210, opened 2019-07-04
Converts all attack animations to the new proper type

DominikPanic:
begone oldcode

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -476,18 +476,30 @@
 	var/pixel_y_diff = 0
 
 	var/direction = get_dir(src, A)
-	if(direction & NORTH)
-		pixel_y_diff = 8
-	else if(direction & SOUTH)
-		pixel_y_diff = -8
-
-	if(direction & EAST)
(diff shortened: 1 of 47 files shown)


code/game/atoms_movable.dm
```diff
@@ -476,18 +476,30 @@
...
-	if(direction & NORTH)
-		pixel_y_diff = 8
-	else if(direction & SOUTH)
-		pixel_y_diff = -8
-
-	if(direction & EAST)
-		pixel_x_diff = 8
-	else if(direction & WEST)
-		pixel_x_diff = -8
+	switch(direction)
```
spookydonut: this is worse code, change it back.

DominikPanic: Why is it worse?

spookydonut: because the old code treats them like bitflags instead of enums, and direction is bitflags not enums

Rohesie: This is actually better performant, and doing bitflag comparisons adds nothing. It's neither more clear nor more efficient.

PR #2230, opened 2019-07-07
Sechud jobs begone

Rohesie:
## Changelog
:cl:
del: Now that the jobs display are handled by the squad HUD, the security one for jobs is removed. SecHUDs still allow to set people as wanted statuses and whatnot.
/:cl:

### DIFF
--- a/code/game/data_huds.dm
+++ b/code/game/data_huds.dm
@@ -364,11 +364,7 @@
 	hud_list[QUEEN_OVERWATCH_HUD] = holder
 
 
-/datum/atom_hud/security //Legacy for now, as the squad hud takes over for job slots.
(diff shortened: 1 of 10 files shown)


code/game/data_huds.dm
```diff
@@ -364,24 +364,12 @@
 	hud_list[QUEEN_OVERWATCH_HUD] = holder
 
 
-/datum/atom_hud/security
-	hud_icons = list(ID_HUD)
+/datum/atom_hud/security //Legacy for now, as the squad hud takes over for job slots.
```
DominikPanic: Just make this the advanced one, replacing where appropriate.

PR #2241, opened 2019-07-08
Limb armor datum

Rohesie:
Limb now uses the armor datum, and it updates on equip/unequip to save some processing when handling damage.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -88,8 +88,6 @@ Contains most of the procs that are called when a mob is attacked by something
 
 //this proc returns the armour value for a particular external organ.
 /mob/living/carbon/human/proc/getarmor_organ(datum/limb/affected_limb, type)
-	if(!type)
-		CRASH("[src] called getarmor_organ() for [affected_limb] with no type.")
 	return affected_limb.armor.getRating(type)
 
 
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -71,14 +71,16 @@ Contains most of the procs that are called when a mob is attacked by something
 
 
 /mob/living/carbon/human/proc/add_limb_armor(obj/item/armor_item)
-	for(var/datum/limb/limb_to_check in limbs)
+	for(var/i in limbs)
+		var/datum/limb/limb_to_check = i
 		if(!(limb_to_check.body_part & armor_item.flags_armor_protection))
 			continue
 		limb_to_check.armor = limb_to_check.armor.attachArmor(armor_item.armor)
 
 
 /mob/living/carbon/human/proc/remove_limb_armor(obj/item/armor_item)
-	for(var/datum/limb/limb_to_check in limbs)
+	for(var/i in limbs)
+		var/datum/limb/limb_to_check = i
 		if(!(limb_to_check.body_part & armor_item.flags_armor_protection))
 			continue
 		limb_to_check.armor = limb_to_check.armor.detachArmor(armor_item.armor)
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -69,17 +69,27 @@ Contains most of the procs that are called when a mob is attacked by something
 
 	return siemens_coefficient
 
+
+/mob/living/carbon/human/proc/add_limb_armor(obj/item/armor_item)
+	for(var/datum/limb/limb_to_check in limbs)
+		if(!(limb_to_check.body_part & armor_item.flags_armor_protection))
+			continue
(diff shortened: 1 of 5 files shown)


code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -595,8 +596,9 @@ Normal range for a defender's bullet resist should be something around 30-50. ~N
 
 //Deal with xeno bullets.
 /mob/living/carbon/xenomorph/bullet_act(obj/item/projectile/P)
-	if(!P || !istype(P))
-		return
+	if(!istype(P))
```
DominikPanic: Can this check realistically happen? Does checking for it every single time make sense? These are the two questions you need to answer before putting this here would make sense rather than not having it at all.

spookydonut: It can't happen.

spookydonut: If it is happening then it needs to be fixed and the check below will easily detect it

code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -496,7 +496,8 @@ sniper rifle or something similar. I suppose that's to be expected though.
 Normal range for a defender's bullet resist should be something around 30-50. ~N
 */
 /mob/living/carbon/human/bullet_act(obj/item/projectile/P)
-	if(!P) return
+	if(!istype(P))
```
spookydonut: same here

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -69,17 +69,27 @@ Contains most of the procs that are called when a mob is attacked by something
 
 	return siemens_coefficient
 
+
+/mob/living/carbon/human/proc/add_limb_armor(obj/item/armor_item)
+	for(var/datum/limb/limb_to_check in limbs)
```
spookydonut: make this typeless

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -69,17 +69,27 @@ Contains most of the procs that are called when a mob is attacked by something
...
+		if(!(limb_to_check.body_part & armor_item.flags_armor_protection))
+			continue
+		limb_to_check.armor = limb_to_check.armor.attachArmor(armor_item.armor)
+
+
+/mob/living/carbon/human/proc/remove_limb_armor(obj/item/armor_item)
+	for(var/datum/limb/limb_to_check in limbs)
```
spookydonut: same here

code/modules/projectiles/updated_projectiles/projectile.dm
Rohesie: It cannot. I removed it in favor of a big screaming feedback just in case.
I intend to remove it at some point if we never get one such error message.

DominikPanic: Then it should just be removed outright. We can also ask ninja if you have doubts, though.

Rohesie: Any reason to urgently remove it right now instead of letting it be tested a few rounds before removal?

DominikPanic: Becaus you haven't proven a reason for it to be there - a necessesity for this kind of check.

Rohesie: Do notice I'm not adding the check.

DominikPanic: You're changing the line.

Rohesie: Alright, I'll fix that.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -69,17 +69,29 @@ Contains most of the procs that are called when a mob is attacked by something
...
-	var/list/protective_gear = list(head, wear_mask, wear_suit, w_uniform, gloves, shoes)
-	for(var/gear in protective_gear)
-		if(gear && istype(gear ,/obj/item/clothing))
-			var/obj/item/clothing/C = gear
-			if(C.flags_armor_protection & def_zone?.body_part)
-				protection += C.armor.getRating(type)
-	return protection
+/mob/living/carbon/human/proc/getarmor_organ(datum/limb/affected_limb, type)
+	if(!type)
+		CRASH("[src] called getarmor_organ() for [affected_limb] with no type.")
```
spookydonut: this will already runtime when called without a type

Rohesie: Want the check removed?

spookydonut: its just extra code that doesnt actually provide extra debugging info since there would still be a stack trace.

PR #2243, opened 2019-07-09
Defender rebalance

Spyroshark:
## About The Pull Request

This gives defenders better stats and somewhat better upgrades, to make staying as a defender more viable.

## Why It's Good For The Game

The other combat T1 xenos, runner and sentinel, are useful throughout the round and are just as capable as T2s by the time they reach elder. Defenders are currently much worse then warriors and other front line xenos and need to be buffed so that they are viable later on in the round, since this caste is currently useless late game.

## Changelog
:cl:
balance: Defenders now have more health, armor, and stronger defensive abilities.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defender/castedatum_defender.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defender/castedatum_defender.dm
@@ -11,8 +11,8 @@
 	wound_type = "defender" //used to match appropriate wound overlays
 
 	// *** Melee Attacks *** //
-	melee_damage_lower = 15
-	melee_damage_upper = 25
+	melee_damage_lower = 20
+	melee_damage_upper = 30
 
 	// *** Tackle *** //
 	tackle_damage = 30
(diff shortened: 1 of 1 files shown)


Pariah919:
what even

drawsstuff:
Defender is worse than warrior in every single way. 
Ancient defender has worse armor than a **young** warrior

Spyroshark:
Defender is too squishy and unarmored for a front-line caste. Since it's abilities aren't as good as they used to be it needs to be stronger to be worth playing

spookydonut:
I dont think these stat changes will meaningfully address the perceived issues with the defender, but they are worth a try.

Rushnagar:
I don't see why the slash damage needs to be increased, they deal more damage by having increased staying power after all.
I also dislike changing too many things at once, makes hard to tell where the buffs "went too far" in cases where things go wrong.

Rohesie:
Agreed, changing armor, speed and damage all in one go is a bit too much.

The change is slow, but you've said the defender is squishy. Do you also feel it does too little damage and is too slow?

We don't want to move towards making the game faster and more rocket-tag-based in general, so only buff speed and damage in those situations, else leave them as they are.

Spyroshark:
That's fair. I'm gonna close this pr since it isn't really thought through well enough, I'll ask for some more opinions on defender in the balance channel on the discord and maybe make a better PR in the future.

PR #2246, opened 2019-07-12
Blockers rework

Rohesie:
The intent here is dealing with early rushes.
The marine shutters should give the engineers some time to build a minimal FOB.
The fog should allow xenos to retain a safe haven from where to stage attacks.
Both should be gone by around one hour into the round.
If this goes well we can expand it on other maps and remove the anti-rush rules.

## Changelog
:cl:
add: Tweaked all ground maps. They will all have LV-like fog protecting areas, but it can be crossed by xenos at will. The marines gain shutters on both landing zones only they can open (and not close), from inside.
add: Added a dropship consoles to every LZ.
/:cl:

### DIFF
--- a/_maps/map_files/BigRed_v2/BigRed_v2.dmm
+++ b/_maps/map_files/BigRed_v2/BigRed_v2.dmm
@@ -25261,6 +25261,14 @@
 /obj/machinery/door/poddoor/timed_late/containment/landing_zone/lz2,
 /turf/open/floor,
 /area/bigredv2/outside/cargo)
+"ggp" = (
+/obj/machinery/button/door/open_only/landing_zone,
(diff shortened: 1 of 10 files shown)


spookydonut:
I thought this idea was agreed to be anti-fun, for the record I am against this change but I am interested to see this tested and player feedback active sought-after. 

I think the ideas should perhaps be separated and tested in three stages;
1)  lz shutter change
2)  fog change
3)  lz shutter + fog change

The fog line being moved isnt really a big deal given the whole resin wall tfort and surrounding areas meta strat xenos do.

The fog and shutter changes could perhaps be made a config option to simplify the testing of this without having to alter the PR between tests.

PR #2247, opened 2019-07-12
Removes more unused configs

DominikPanic:
The guest stuff is already handled via a config in IsBanned rather than client/New


code/game/objects/structures/stool_bed_chair_nest/chairs.dm
```diff
@@ -37,7 +37,7 @@
 	if(!istype(user) || !isturf(user.loc) || user.incapacitated())
 		return FALSE
 
-	if(!CONFIG_GET(flag/unlimited_rotate_speed))
```
spookydonut: disagree with combining these

code/modules/vehicles/multitile/hardpoints.dm
```diff
@@ -278,7 +278,7 @@ Currently only has the tank hardpoints
 	log_combat(usr, usr, "fired the [src].")
 	log_explosion("[usr] fired the [src] at [AREACOORD(loc)].")
 	P.fire_at(T, owner, src, P.ammo.max_range, P.ammo.shell_speed)
-	if(!CONFIG_GET(flag/tank_mouth_noise))
```
spookydonut: disagree with combining these, either commit to removing the joke noises or keep it as a separate thing

code/game/objects/structures/stool_bed_chair_nest/chairs.dm
DominikPanic: I disagree with this.

code/modules/vehicles/multitile/hardpoints.dm
DominikPanic: I disagree with this.

Rohesie: To be fair if we consider the possible usage, then individual configs is better, as it gives the admins the choice of what to enable and what not to.

DominikPanic: It has never been used and I sure hope it never will. This simply doesn't fit our server and having 200 configs for every little thing is not really feasible nor maintainable.

Rohesie: Eh, remove both then.

DominikPanic: Done

PR #2252, opened 2019-07-12
Fixes the release object verb

DominikPanic:
Me dumb

### DIFF
--- a/code/modules/admin/fun_verbs.dm
+++ b/code/modules/admin/fun_verbs.dm
@@ -910,7 +910,7 @@
 	message_admins("[ADMIN_TPMONTY(usr)] changed hivenumber of [ADMIN_TPMONTY(X)] from [hivenumber_status] to [newhive].")
 
 
-/datum/admins/proc/release()
+/datum/admins/proc/release(obj/OB in world)
 	set category = null
 	set name = "Release Obj"
 


code/modules/admin/fun_verbs.dm
```diff
@@ -910,7 +910,7 @@
 	message_admins("[ADMIN_TPMONTY(usr)] changed hivenumber of [ADMIN_TPMONTY(X)] from [hivenumber_status] to [newhive].")
 
 
-/datum/admins/proc/release()
+/datum/admins/proc/release(obj/O in world)
```
spookydonut: does this get used for anything except unpossessing an object?

spookydonut:
https://github.com/tgstation/tgstation/blob/08a5b40ffb3ed407f0d2122e03e90218e31ae8da/code/modules/admin/verbs/possess.dm#L28

DominikPanic:
I like the way it's done here, but if Roh requests it I will do it that way.

spookydonut:
> 
> 
> I like the way it's done here, but if Roh requests it I will do it that way.

build is failing since its getting the object from the currently possessed one

spookydonut:
should probably shift it from a datum proc to a global proc like tg has it

Rohesie:
Got an opinion, @ninjanomnom ?

ninjanomnom:
Not really

Rohesie:
I'm indifferent as well.
This is fine, no need to change it.

code/modules/admin/fun_verbs.dm
```diff
@@ -910,7 +910,7 @@
 	message_admins("[ADMIN_TPMONTY(usr)] changed hivenumber of [ADMIN_TPMONTY(X)] from [hivenumber_status] to [newhive].")
 
 
-/datum/admins/proc/release()
+/datum/admins/proc/release(obj/OB in world)
```
psykzz: `OB` isn't used in the proc. How does this work?

DominikPanic: it doesn't matter what it's named, since it's being added as a verb, the argument specifies what it works on

PR #2253, opened 2019-07-12
Renames status procs

DominikPanic:
Roh wanted this
For an easy list of changes, see status_procs.dm


psykzz:
Wouldn't it make sense to also create the proc `apply_status(amount, type, flags)` similar to the `apply_damage` proc?

For example
```
C.apply_damage(100, BRUTE, part)
C.apply_status(2, STUN)
```

@Rohesie?

Rohesie:
Yea, it would.
Though I'm not entirely sure about this. As /tg/ handles it differently.

PR #2256, opened 2019-07-12
Imaginary friend fixes

DominikPanic:
The message should only display once now, they should be less hacky.

### DIFF
--- a/code/modules/admin/fun_verbs.dm
+++ b/code/modules/admin/fun_verbs.dm
@@ -977,40 +977,26 @@
 	if(!check_rights(R_FUN|R_MENTOR))
 		return
 
-	if(istype(usr, /mob/camera/imaginary_friend))
-		var/mob/camera/imaginary_friend/IF = usr
-		IF.deactivate()
+	var/client/C = usr.client
+
+	if(istype(C.mob, /mob/camera/imaginary_friend))
+		var/mob/camera/imaginary_friend/IF = C.mob
+		IF.ghostize()
 		return
 
-	if(is_mentor(usr.client) && !isobserver(usr))
-		to_chat(usr, "<span class='warning'>Can only become an imaginary friend while observing.</span>")
+	if(is_mentor(C) && !isobserver(C.mob))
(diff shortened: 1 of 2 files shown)


code/modules/admin/fun_verbs.dm
```diff
@@ -977,40 +977,26 @@
...
-				return
-			L = C.mob
-		if("Mob")
-			var/mob/M = input("Please, select a mob.", "Imaginary Friend") as null|anything in sortNames(GLOB.mob_living_list)
-			if(!M)
-				return
-			L = M
-
-	if(!isobserver(usr))
+	if(is_mentor(C) && !isobserver(C.mob))
```
```
if(!isobserver(C.mob))
	if(is_mentor(C)
		usr.client.holder.admin_ghost()
		to_chat(C, "<span class='warning'>Can only become an imaginary friend while observing.</span>")
		return
	C.holder.admin_ghost()
```

DominikPanic: That is changing the functionality to something completely different

Rohesie: Unless I'm missing something the behavior should remain exactly the same this way.

DominikPanic: I want to return early for mentors, not admin ghost them, i want to only admnghost admins.

Rohesie: That's what will happen with my code. Non-observer mentors will trigger `if(!isobserver(C.mob))` and `if(is_mentor(C))`, reaching the `return` before the `C.holder.admin_ghost()`.

Rohesie: Though it would be more maintainable if it was something like `!is_admin` instead.

DominikPanic:
```
	if(is_mentor(C))
		usr.client.holder.admin_ghost()
```
No it wont

Rohesie: Heck, I only saw that now. Very much unintended. Remove that line and the rest follows.

PR #2258, opened 2019-07-12
Implements FOOC (faction OOC)

DominikPanic:
Death to brackets

:cl: LaKiller8
FOOC (Faction OOC) has been added.
/:cl:


factoryman942:
tfw two brands of ooc aren't enough so you need a third one

DominikPanic:
> 
> 
> tfw two brands of ooc aren't enough so you need a third one

Pepsi, Coke, Dr Pepper

factoryman942:
under what circumstances would using fooc be preferable to using looc
and under what circumstances would using looc be preferable to using fooc

Rohesie:
When you need to interject about a recent game mechanics change, for example. Like about the Shrike, without telling all the marines the player is struggling with its mechanics.

psykzz:
I agree 3 is quite a lot, with FOOC, LOOC doesn't make as much sense, i only see it used to talk with others when you die "How much hp did you have left before you got me down?"

Spyroshark:
we already have LOOC. The only servers that use FOOC is highrp servers that have team antags trying to do gimmicks and whatnot together. There's absolutely zero reason to add it here it's just going to be shitpost central and whining about xenos

DominikPanic:
I also disagree with this PR (which is a bit ironic), but I'm picking the lesser of the two evils (FOOC vs brackets in IC chat)

psykzz:
Why do you need one or the other in the first place.

DominikPanic:
I don't want either, It's Rohesie's decision (see #policy for a LOT of reeeing)

spookydonut:
This is a feature PR with a config option, discussions about if the actual server should use it shouldn't influence the fate of this PR.

DominikPanic:
> 
> 
> This is a feature PR with a config option, discussions about if the actual server should use it shouldn't influence the fate of this PR.

I'd rather close it than let a config decide if it gets merged or not. My philosophy here is a little different.

Rohesie:
I'll be closing this PR due to the current rejection.

Issue #2239, opened 2019-07-07
Op table unbuckles AND removes anestesia tank


Shadow-Quill:
## Reproduction:
1) Buckle someone into an optable.
2) Apply mask.
3) Unbuckle the person by clicking the table.
4) Watch as you unbuckle, remove the mask _and the tank_. 
![image](https://user-images.githubusercontent.com/44811257/60774416-58d1c280-a0d9-11e9-8742-87edb1297009.png)
Also the person doing the untaking isn't correctly shown.

PR #2259, opened 2019-07-12
Fixes optable quirkiness

DominikPanic:
Fixes: #2239


DominikPanic:
> 
> 
> this is introducing new quirkiness, I assume this is about self-unbuckling in which case add an exception to always unbuckle if user == buckled_mob

This is not about self-unbuckling.

spookydonut:
Superceded by #2264

PR #2261, opened 2019-07-12
Makes fire work properly

DominikPanic:
Holy shit how did NOONE notice this huge oversight for so long (myself included)


Pariah919:
Because literally nobody used non flamer fire I guess.

Issue #2238, opened 2019-07-07
Typing indicator doesn't go away if you get stunned/knocked out while typing


MoXuT0:
## Testmerges:

Round ID:  657 
BYOND Version:  512.1464 
Server revision compiled on:  
Master commit: 291d79988c28dead09cc30cc39fc1a1744ae685d 
The following pull requests are currently test merged: 
#2234: 'Adds action buttons for voting' by DominikPanic at commit bf683d6828 
#2229: 'Disarming C4 with a multitool' by Eskjjlj at commit 397863c704 
#2228: 'Feedback for joining as larva.' by Rohesie at commit 5ffdf9d4cb 
#2223: 'A different medkit size fix' by DominikPanic at commit ebec5e44ae 
#2215: 'Fixes ASL radio for real' by DominikPanic at commit aae182b995 
#2210: 'Converts all attack animations to the new proper type' by DominikPanic at commit 2982dc4a3f 
#2214: 'Life() cleanup and microoptimizations' by DominikPanic at commit eabe5dddee 
#2211: 'Storage rebalance' by MoXuT0 at commit f5ddb6b0ce 
#2161: 'Makes the AI playable.' by DominikPanic at commit 3e6f392538 
#2220: 'Removes unused implant code, fixes CL implanter' by DominikPanic at commit 45e0614662 
#2217: 'Fixes duplicate pipe causing warnings on ice' by DominikPanic at commit e96b5b0654 
#2216: 'Fixes EORD ghost runtimes' by DominikPanic at commit fc01571efc 
#2227: 'Fixes teleporter attack_hand runtime' by DominikPanic at commit 84b2976dde 
#2226: 'Improves admin input sanitization' by DominikPanic at commit 74efda0ad6 
#2233: '.45 Ammo box' by Hughgent at commit 81179c1f21 

## Reproduction:

Try to type something > get knocked out > typing indicator doesn't go away now

![dreamseeker_AXCb3IPt1X](https://user-images.githubusercontent.com/34478337/60774118-3d39dc80-a118-11e9-8722-aecbc56aff30.png)

DominikPanic:
That is intended since the dialogue boxes doesn't get closed either

MoXuT0:
But you can’t close that dialogue box after you get knocked out, typing indicator will stay forever

PR #2262, opened 2019-07-12
Typing indicator fixes

DominikPanic:
Should help with stuff
Fixes: #2238
Fixes: #2288

### DIFF
--- a/code/modules/mob/typing_indicator.dm
+++ b/code/modules/mob/typing_indicator.dm
@@ -2,14 +2,11 @@
 	if(stat != CONSCIOUS)
 		return
 
-	var/image/I
-
 	if(emoting)
-		I = image('icons/mob/talk.dmi', src, "emoting")
+		typing_indicator = image('icons/mob/talk.dmi', src, "emoting")
 	else
-		I = image('icons/mob/talk.dmi', src, "typing")
+		typing_indicator = image('icons/mob/talk.dmi', src, "typing")
 
-	typing_indicator = I
 	add_emote_overlay(typing_indicator, remove_delay = NONE)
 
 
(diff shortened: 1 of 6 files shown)


code/modules/mob/typing_indicator.dm
```diff
@@ -1,31 +1,33 @@
-/mob/proc/toggle_typing_indicator(emoting)
-	var/image/typing_indicator
+/mob/proc/add_typing_indicator(emoting)
+	if(stat != CONSCIOUS)
+		return
+
+	var/image/I
```
Rohesie: Since `typing_indicator` is type cast, this intermediary var is not really necessary.

code/modules/mob/typing_indicator.dm
```diff
@@ -1,31 +1,30 @@
...
-		typing = null
 
-	else if(stat == CONSCIOUS)
-		typing = typing_indicator
-		add_emote_overlay(typing, remove_delay = NONE)
+/mob/proc/remove_typing_indicator(emoting)
+	if(!typing_indicator)
+		return
+
+	remove_emote_overlay(client, typing_indicator, viewers())
```
psykzz: What about the people that were around when it triggered but moved away for the removal?
Do you need to keep a list of "viewers" to ensure you remove the emote from the right people instead?

DominikPanic: So far it hasn't been an issue, I am not changing the behavior.

psykzz: There are definite issues with the indicator not going away, this seems like its an area for improvement.

DominikPanic: I am fixing one specific issue, feel free to fix more!

PR #2248, opened 2019-07-12
Allows admins to set a custom tip of the round

DominikPanic:
Also fixed da code.

Fixes: #2237

PR #2265, opened 2019-07-13
SSstrings

spookydonut:
Adds a string subsystem for loading files into globs.

See also: #2248

### DIFF
--- a/code/__DEFINES/_globals.dm
+++ b/code/__DEFINES/_globals.dm
@@ -58,3 +58,12 @@
 
 //Create an typed null global
 #define GLOBAL_DATUM(X, Typepath) GLOBAL_RAW(Typepath/##X); GLOBAL_UNMANAGED(X)
+
+GLOBAL_LIST_EMPTY(string_files)
+
+//Create a global to be populated with a file during init
+#define GLOBAL_LIST_FILE_STRING(X, Filename) GLOBAL_LIST(X); \
+/datum/controller/global_vars/InitGlobal##X(){\
+	..();\
+	GLOB.string_files[#X] = #Filename;\
+}
--- a/code/controllers/subsystem/strings.dm
+++ b/code/controllers/subsystem/strings.dm
@@ -1,0 +1,10 @@
+SUBSYSTEM_DEF(strings)
+	name = "Strings"
+	init_order = INIT_ORDER_STRINGS
+	flags = SS_NO_FIRE
+
+/datum/controller/subsystem/strings/Initialize(timeofday)
+	for(var/i in GLOB.string_files)
+		GLOB.vars[i] = file2list(GLOB.string_files[i])
+
+	return ..()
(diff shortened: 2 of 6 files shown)


psykzz:
We were also trying to enforce loading strings from the strings directory, would it make sense for this subsystem to enforce that by only loading strings pathed inside?

eg removing `string/` from each import and adding it back in within the subsystem?

code/controllers/subsystem/strings.dm
```diff
@@ -0,0 +1,10 @@
+SUBSYSTEM_DEF(strings)
+	name = "Strings"
+	init_order = INIT_ORDER_STRINGS
+	flags = SS_NO_FIRE
+
+/datum/controller/subsystem/strings/Initialize(timeofday)
+	for(var/i in GLOB.string_files)
+		GLOB.vars[i] = file2list(GLOB.string_files[i])
```
psykzz: If a file is large for whatever reason this could lock up? 
Could you add the TICK_CHECK here?

spookydonut: it's init so i dont think it matters

spookydonut:
> 
> 
> We were also trying to enforce loading strings from the strings directory, would it make sense for this subsystem to enforce that by only loading strings pathed inside?
> 
> eg removing `string/` from each import and adding it back in within the subsystem?

i'd rather just keep explicit full file paths.

code/__DEFINES/_globals.dm
```diff
@@ -58,3 +58,12 @@
 
 //Create an typed null global
 #define GLOBAL_DATUM(X, Typepath) GLOBAL_RAW(Typepath/##X); GLOBAL_UNMANAGED(X)
+
+GLOBAL_LIST_EMPTY(string_files)
```
SpaceManiac: Defining a global in this file seems odd. Maybe this could go on `SSstrings` instead?

code/__DEFINES/_globals.dm
```diff
@@ -58,3 +58,12 @@
...
+
+//Create a global to be populated with a file during init
+#define GLOBAL_LIST_FILE_STRING(X, Filename) GLOBAL_LIST(X); \
+/datum/controller/global_vars/InitGlobal##X(){\
+	..();\
+	GLOB.string_files[#X] = Filename;\
```
SpaceManiac: This "different kind of InitGlobal" that doesn't actually init that global but instead sets a key so it will be initted later seems... odd to me, though I guess it works.

code/controllers/subsystem/strings.dm
```diff
@@ -0,0 +1,11 @@
...
```
SpaceManiac: `GLOB.vars` screams to me like a bad idea. An admin could edit `GLOB.string_files` and then proccall `SSstrings.Initialize()` to overwrite protected vars, maybe if not to anything useful, but at least to garbage

code/__DEFINES/_globals.dm
spookydonut: I wasn't sure of the define order but yeah it should be there

code/__DEFINES/_globals.dm
spookydonut: protected does this i just copied that

PR #2267, opened 2019-07-13
Another do_after runtimes

DominikPanic:
I should just regex for the fourth argument being a number but I gotta learn regex better first


spookydonut:
> 
> 
> I should just regex for the fourth argument being a number but I gotta learn regex better first

`do_after\([^,]+,[^,]+,[^,]+,\s?[0-9]`

PR #2274, opened 2019-07-13
Fixes a HUD runtime

DominikPanic:
Toggling to a different style of hud means this won't always be present, and thus this check is necessary.


DominikPanic:
> 
> 
> `if(hud_used.nutrition_icon)` will runtime down there.
> Can you nest them all into an `if(hud_used)`?
> This whole proc could use some early returns, but that's out of scope.

It shouldn't runtime.

PR #2287, opened 2019-07-14
Adds a preference for observer actions

DominikPanic:


Rohesie:
I used it quite a bit. Found it much less clunkier than having to hunt for the verb in the tab.
Maybe ask more people, or make it able to be hidden?

DominikPanic:
A lot of folks share the same opinion, it was discussed quite a bit and this is not the proper way to go about things.

spookydonut:
make it a preference, i like this feature.

PR #2294, opened 2019-07-15
Fixes being able to psychic-throw someone more than once if you clicked fast enough

Rohesie:
This was caused because `throw_at()` is a sleeping-while proc instead of like /tg/'s which processes.
That prevented the cooldown from being set.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/shrike/abilities_shrike.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/shrike/abilities_shrike.dm
@@ -84,6 +84,7 @@
 	playsound(victim,'sound/weapons/alien_claw_block.ogg', 75, 1)
 
 	succeed_activate()
+	add_cooldown()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/shrike/abilities_shrike.dm
```diff
@@ -84,6 +84,7 @@
 	playsound(victim,'sound/weapons/alien_claw_block.ogg', 75, 1)
 
 	succeed_activate()
```
psykzz: Not a problem for this PR but a note. Couldn't succeed_activate just set the cooldown instead?

Rohesie: Sounds sensible. If you consume plasma in general you want the cooldown activated if there's any.
Should poke @spookydonut about it, though, since this system is his design.

PR #2295, opened 2019-07-16
Timers out of bounds.

Rohesie:
`/datum/controller/subsystem/timer/PreInit()` is ran before world.
In there it is defined that `bucket_list.len = BUCKET_LEN`
`BUCKET_LEN` is `(world.fps*1*60)`
When `PreInit` happens, `fps` will have not been set yet, so it will depend on local configs. In my case it starts at `10` before later on it is set to `20`.
By the time `bucketJoin()` is called the first times, that has already happened, so you have a list whose length is 600 in my case (10 fps times 60).
But the position for this first new bucket will use the fps of its time:
`var/bucket_pos = BUCKET_POS(src)`
`BUCKET_POS(timer)` is `(((round((timer.timeToRun - SStimer.head_offset) / world.tick_lag)+1) % BUCKET_LEN)||BUCKET_LEN)`
By that time `BUCKET_LEN` will be 1200, as `world.fps` is 20.

This adds a check that is already present in the timer process looping, to adjust in case the world fps settings change.

### DIFF
--- a/code/controllers/subsystem/timer.dm
+++ b/code/controllers/subsystem/timer.dm
@@ -429,10 +429,6 @@ SUBSYSTEM_DEF(timer)
 	//get the list of buckets
 	var/list/bucket_list = SStimer.bucket_list
 
-	if(length(bucket_list) != BUCKET_LEN)
-		SStimer.reset_buckets()
-		bucket_list = SStimer.bucket_list
-
 	//calculate our place in the bucket list
 	var/bucket_pos = BUCKET_POS(src)
 
--- a/code/controllers/subsystem/timer.dm
+++ b/code/controllers/subsystem/timer.dm
@@ -46,7 +46,9 @@ SUBSYSTEM_DEF(timer)
 
 	if(lit && lit < last_check && head_offset < last_check && last_invoke_warning < last_check)
 		last_invoke_warning = world.time
-		message_admins("No regular timers processed in the last [BUCKET_LEN * 1.5] ticks[bucket_auto_reset ? ", resetting buckets" : ""].")
+		var/msg = "No regular timers processed in the last [BUCKET_LEN * 1.5] ticks[bucket_auto_reset ? ", resetting buckets" : ""]."
+		message_admins(msg)
+		WARNING(msg)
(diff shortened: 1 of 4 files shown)


code/controllers/subsystem/timer.dm
```diff
@@ -427,6 +429,10 @@ SUBSYSTEM_DEF(timer)
 	//get the list of buckets
 	var/list/bucket_list = SStimer.bucket_list
 
+	if(length(bucket_list) != BUCKET_LEN)
+		SStimer.reset_buckets()
+		bucket_list = SStimer.bucket_list
```
ninjanomnom: This should not be done here, if it's really a problem for you then you need to fix this in init somewhere

ninjanomnom:
@MrStonedOne

Rohesie:
Set up some debugging to know how this went:
```
## WARNING: /datum/controller/master/New || FPS: 10 in code/controllers/master.dm at line 65 src: Master usr: .
## WARNING: /datum/controller/subsystem/timer/PreInit || FPS: 10 in code/controllers/subsystem/timer.dm at line 32 src: Timer usr: .
423 global variables
## WARNING: /world/New || Pre || FPS: 10 in code/game/world.dm at line 12 src:  usr: .
## WARNING: /world/New || Pos || FPS: 20 in code/game/world.dm at line 60 src:  usr: .
## WARNING: /datum/controller/master/Initialize || pre || FPS: 20 in code/controllers/master.dm at line 206 src: Master usr: .
## WARNING: /datum/controller/master/Initialize || pos || FPS: 20 in code/controllers/master.dm at line 208 src: Master usr: .
```
In short, it got changed at `/world/New` but the timer SS's `PreInit` ran before.

New solution turns changing FPS into a proc, which deals with the SS.

MrStonedOne:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/2295/files#diff-6aa021daf96230ce85097b19bacaac39L113 handles detecting this and resetting buckets already

Rohesie:
I can port it, sure.

PR #2296, opened 2019-07-16
Reboot runtime

Rohesie:
Rebooting from Dream Daemon will runtime from an invalid index if `SSmapping.next_map_configs` is `null` for example. Which is the case by default.

Went ahead and tweaked the proc a little more, nothing big.

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -138,14 +138,14 @@ GLOBAL_VAR_INIT(bypass_tgs_reboot, world.system_type == UNIX && world.byond_buil
 
 		var/datum/map_config/next_map
 
-		if(length(SSmapping.next_map_configs) >= GROUND_MAP)
+		if(length(SSmapping.next_map_configs))
 			next_map = SSmapping.next_map_configs[GROUND_MAP]
 		else
 			next_map = SSmapping.configs[GROUND_MAP]
 		if(next_map)
 			msg += "Next Ground Map: [next_map.map_name]"
 
-		if(length(SSmapping.next_map_configs) >= SHIP_MAP)
+		if(length(SSmapping.next_map_configs))
 			next_map = SSmapping.next_map_configs[SHIP_MAP]
 		else
 			next_map = SSmapping.configs[SHIP_MAP]
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -136,20 +136,21 @@ GLOBAL_VAR_INIT(bypass_tgs_reboot, world.system_type == UNIX && world.byond_buil
 		if(GLOB.round_id)
 			msg += "Round ID [GLOB.round_id] finished"
 
-		var/datum/map_config/next_map
+		var/datum/map_config/next_gound_map
+		var/datum/map_config/next_ship_map
 
-		if(length(SSmapping.next_map_configs))
-			next_map = SSmapping.next_map_configs[GROUND_MAP]
-		else
-			next_map = SSmapping.configs[GROUND_MAP]
-		if(next_map)
+		if(length(SSmapping.next_map_configs)) //To avoid a bad index, let's check if there's actually a list.
+			next_gound_map = SSmapping.next_map_configs[GROUND_MAP]
+			next_ship_map = SSmapping.next_map_configs[SHIP_MAP]
+
+		if(!next_gound_map) //The list could hold a single item, so better check each because there's no guarantee both exist.
+			next_gound_map = SSmapping.configs[GROUND_MAP]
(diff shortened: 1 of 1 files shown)


code/game/world.dm
```diff
@@ -131,12 +131,43 @@ GLOBAL_VAR_INIT(bypass_tgs_reboot, world.system_type == UNIX && world.byond_buil
...
 		send2update(CONFIG_GET(string/restart_message))
-		send2update("Round ID [GLOB.round_id] finished | Next Ground Map: [SSmapping.next_map_configs[GROUND_MAP] ? SSmapping.next_map_configs[GROUND_MAP].map_name : SSmapping.configs[GROUND_MAP].map_name] | Next Ship Map: [SSmapping.next_map_configs[SHIP_MAP] ? SSmapping.next_map_configs[SHIP_MAP].map_name : SSmapping.configs[SHIP_MAP].map_name] | Round End State: [SSticker.mode?.round_finished] | Players: [length(GLOB.clients)]")
+		var/list/msg = list()
+
+		if(GLOB.round_id)
+			msg += "Round ID [GLOB.round_id] finished"
+
+		var/datum/map_config/next_map
+
+		if(length(SSmapping.next_map_configs) >= GROUND_MAP)
```
DominikPanic:
```suggestion
		if(SSmapping.next_map_configs[GROUND_MAP])
```

code/game/world.dm
```diff
@@ -131,12 +131,43 @@ GLOBAL_VAR_INIT(bypass_tgs_reboot, world.system_type == UNIX && world.byond_buil
...
+			next_map = SSmapping.next_map_configs[GROUND_MAP]
+		else
+			next_map = SSmapping.configs[GROUND_MAP]
+		if(next_map)
+			msg += "Next Ground Map: [next_map.map_name]"
+
+		if(length(SSmapping.next_map_configs) >= SHIP_MAP)
```
DominikPanic:
```suggestion
		if(SSmapping.next_map_configs[SHIP_MAP])
```

code/game/world.dm
Rohesie: This is the original runtime cause.

code/game/world.dm
```diff
@@ -131,12 +131,43 @@ GLOBAL_VAR_INIT(bypass_tgs_reboot, world.system_type == UNIX && world.byond_buil
...
 		send2update(CONFIG_GET(string/restart_message))
-		send2update("Round ID [GLOB.round_id] finished | Next Ground Map: [SSmapping.next_map_configs[GROUND_MAP] ? SSmapping.next_map_configs[GROUND_MAP].map_name : SSmapping.configs[GROUND_MAP].map_name] | Next Ship Map: [SSmapping.next_map_configs[SHIP_MAP] ? SSmapping.next_map_configs[SHIP_MAP].map_name : SSmapping.configs[SHIP_MAP].map_name] | Round End State: [SSticker.mode?.round_finished] | Players: [length(GLOB.clients)]")
+		var/list/msg = list()
+
+		if(GLOB.round_id)
+			msg += "Round ID [GLOB.round_id] finished"
+
+		var/datum/map_config/next_map
+
+		if(length(SSmapping.next_map_configs))
```
DominikPanic: Now I am 100% sure this will runtime. Imagine only one map has the next config (since only one was voted for, the current case, meaning the length of the list will be 1, both will pass, but it will runtime with 1 case.

Rohesie: Started up, changed the map immediately, populating the list to hold a single item:
Xf2cqWq
Then restarted. It didn't runtime, worked normally.

The runtime happened when the list was empty/null, because then `SSmapping.next_map_configs[GROUND_MAP]` became `SSmapping.null[GROUND_MAP]`
If you have a single item in the list, then you can look for string indexes that don't exist (just not integer ones, because then you are trying to access the list directly).

What happens instead is that `next_map = SSmapping.next_map_configs[SHIP_MAP]` returns a `null` value, because that index is empty. Which is fine because there's an `if(next_map)` check next.

Though now that you mention it there are two `if(length(SSmapping.next_map_configs))` checks, and those could be nested into a single one.

DominikPanic: I may be stupid then, carry on

Rohesie:
This is already tested locally.
Current code is runtiming, this doesn't need further testing.

DominikPanic:
> 
> 
> This is already tested locally.
> Current code is runtiming, this doesn't need further testing.

You can get another maintainer to merge it, but I want it tested first since to me it seems possibly problematic.

PR #2298, opened 2019-07-18
Improves random name code

spookydonut:
old code had a lot of code duplication and unneeded capitalize() calls

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -147,10 +147,7 @@
 
 /proc/random_unique_name(gender, attempts_to_find_unique_name = 10)
 	for(var/i in 1 to attempts_to_find_unique_name)
-		if(gender == FEMALE)
-			. = capitalize(pick(SSstrings.get_list_from_file("names/first_female"))) + " " + capitalize(pick(SSstrings.get_list_from_file("names/last_name")))
-		else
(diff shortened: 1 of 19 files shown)


code/__HELPERS/mobs.dm
```diff
@@ -147,10 +147,7 @@
 
 /proc/random_unique_name(gender, attempts_to_find_unique_name = 10)
 	for(var/i in 1 to attempts_to_find_unique_name)
-		if(gender == FEMALE)
-			. = capitalize(pick(SSstrings.get_list_from_file("names/first_female"))) + " " + capitalize(pick(SSstrings.get_list_from_file("names/last_name")))
-		else
-			. = capitalize(pick(SSstrings.get_list_from_file("names/first_male"))) + " " + capitalize(pick(SSstrings.get_list_from_file("names/last_name")))
+		. = GLOB.namepool[/datum/namepool/clf].get_random_name(gender)
```
Rohesie: Isn't this `x[y].z` pattern the same as `x[y]:z`, and thus something we were trying to get rid of?

psykzz: No this is instance lookup and execute?

`:z` hits the base of the namepool afaik.

So `GLOB.namepool[/datum/namepool/clf].get_random_name(gender)` would've been `GLOB.namepool[/datum/namepool].get_random_name(gender)` if you were right

spookydonut: namepool is GLOBAL_LIST_EMPTY_TYPED

spookydonut: It's enforced by the linter

PR #2300, opened 2019-07-18
Fixes HUD updating runtimes

DominikPanic:
You were right @Rohesie, though I have only seen this with clientless bodies which is why it didn't show up until my local testing.


code/modules/mob/living/carbon/human/life/handle_regular_hud_updates.dm
```diff
@@ -186,7 +191,4 @@
 				else if(bodytemperature <= species.body_temperature - temp_step * 1)
 					hud_used.bodytemp_icon.icon_state = "temp-1"
 				else
-					hud_used.bodytemp_icon.icon_state = "temp0"
-
-
-	interactee?.check_eye(src)
\ No newline at end of file
+					hud_used.bodytemp_icon.icon_state = "temp0"
```
spookydonut:
```suggestion
					hud_used.bodytemp_icon.icon_state = "temp0"

```

DominikPanic: Why should I put random whitespace at the end of the file?

spookydonut: you should have a newline as the last character of the file to avoid having to change two lines to add something to the end of the file as per ninjanomnom on discord has told you in the past.

DominikPanic: I don't remember ninjanomnom saying that nor that being in the contributor guidelines.

spookydonut: well since you're touching the last line of the file just stick a newline in there and this can be merged.

DominikPanic: I disagree with this change request, but I'm not opposed to get another opinion.

spookydonut: https://discordapp.com/channels/498569646014857217/498576390753615882/599805710787936266

PR #2302, opened 2019-07-18
Fixes keybinding prefs for players who played here previously

DominikPanic:
Ask Psy as for why it's necessary in case it's not clear.


Rohesie:
Can you explain here for documentation purposes, @psykzz ?

DominikPanic:
Basically since I removed the hacky check in my previous PR that meant people who aren't 100% new but haven't played and gotten their keybindings yet would start with an empty list.

PR #2306, opened 2019-07-19
Battle rifle balance tweaks

MoXuT0:
## About The Pull Request

Adds 0.05 accuracy multiplier to M4RA so total accuracy is 1.1 with stock. Incendiary ammo is no longer straight upgrade from default, nerfs damage and accuracy bonus by 10. This should be enough not to miss on screens edge.

## Changelog
:cl:
balance: Increased M4RA accuracy multiplier slightly.
balance: Nerfed high velocity incendiary bullet damage and accuracy bonus.
/:cl:


Rohesie:
Let's see how #balance reacts to this.

PR #2310, opened 2019-07-19
Clarifies the hotkey preference name, removes obsolete SSinput things

DominikPanic:
Tested locally, Psy said most of this is done via keybindings so there's no reason to keep this code around as it only makes stuff confusing.

Fixes: #2399

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -34,7 +34,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 
 	var/show_typing = TRUE
 	var/windowflashing = TRUE
-	var/hotkeys = TRUE
+	var/focus_chat = TRUE
 
 	// Custom Keybindings
 	var/list/key_bindings = null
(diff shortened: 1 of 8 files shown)


Rohesie:
2302 in. This ready?

DominikPanic:
> 
> 
> 2302 in. This ready?

No, it needs code changes

DominikPanic:
Should technically be ready now though I cannot guarantee that this will work as testing this is difficult

code/modules/client/preferences.dm
```diff
@@ -998,8 +998,8 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 			if (choice == "Cancel")
 				ShowKeybindings(user)
 				return
-			hotkeys = (choice == "Hotkey")
-			key_bindings = (hotkeys) ? deepCopyList(GLOB.hotkey_keybinding_list_by_key) : deepCopyList(GLOB.classic_keybinding_list_by_key)
+			focus_chat = (choice == "Hotkey")
```
psykzz: If we don't want keybindings to hit focus_chat, this should be removed.

DominikPanic: It's still preferable for it to be there, as most users of classic keybindings want to combine those.

DominikPanic:
Should be ready

PR #2321, opened 2019-07-19
Fixes xeno basketball being enabled by default

DominikPanic:
No more fun.


code/game/objects/items/toys/toys.dm
```diff
@@ -481,6 +481,8 @@
 
 
 /obj/item/toy/beach_ball/basketball/attack_alien(mob/living/carbon/xenomorph/user)
+	if(!CONFIG_GET(flag/fun_allowed))
```
spookydonut: putting too many things on this config flag imo

DominikPanic: It was and already is there, but on the holoball variant. This is not being added.

PR #2326, opened 2019-07-20
Cleans up rav rage decay

psykzz:
Rav rage is being a little weird and decaying pretty quickly. 
This ensures it only decays every 3 seconds instead of every life tick.


code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
```diff
@@ -134,7 +134,7 @@
 
 	GLOB.round_statistics.ravager_second_winds++
 
-	last_rage = current_rage
+	last_rage = world.time
```
DominikPanic: Could you implement this using the new coooldown framework?

PR #2332, opened 2019-07-20
adds a newline to be the eof of all /code/ files and a travis grep fo…

spookydonut:
…r it.

avoids breaking blame for future changes when appending to file end.


spookydonut:
https://discordapp.com/channels/498569646014857217/498576390753615882/599805710787936266

Issue #2194, opened 2019-07-01
You can't put an unspent sniper round (M42A) back into its mag


Nickkoranda:
If you cycle the bolt on your M42A and retrieve your bullet off the ground you can't put it back into a magazine, it's prompts with the error of "These are not the same rounds, better not mix them up"

DominikPanic:
How did you manage to cycle the bolt? I can't seem to do it in-game.

Rohesie:
Can't seem to cock it either. Normally it would be through unique-action, but in this case it's overwritten.

InterroLouis:
Mmhmm. Figured it out. If you use the unload-weapon verb to unload an M42A without a magazine in it, it generates a stackable single bullet that can't be loaded back into the magazine.

factoryman942:
@DominikPanic close this?

PR #2333, opened 2019-07-20
Weapon/Magazine Caliber Fix

InterroLouis:
## About The Pull Request

Fixes the issue of M42A rounds ejected from the rifle being unable to be put back into the magazine. 
## Why It's Good For The Game

It fixes a bug, as outlined in #2194

### DIFF
--- a/code/modules/projectiles/updated_projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/updated_projectiles/guns/specialist.dm
@@ -17,7 +17,7 @@
 	icon_state = "m42a"
 	item_state = "m42a"
 	max_shells = 15 //codex
-	caliber = "10x28mm" //codex
+	caliber = "10x28mm"
 	origin_tech = "combat=6;materials=5"
 	fire_sound = 'sound/weapons/guns/fire/sniper.ogg'
 	dry_fire_sound = 'sound/weapons/guns/fire/sniper_empty.ogg'
@@ -206,7 +206,7 @@
 	icon_state = "m42c"
 	item_state = "m42c"
 	max_shells = 6 //codex
-	caliber = "10x99mm" //codex
+	caliber = "10x99mm"
 	origin_tech = "combat=7;materials=5"
 	fire_sound = 'sound/weapons/guns/fire/sniper_heavy.ogg'
 	dry_fire_sound = 'sound/weapons/guns/fire/sniper_empty.ogg'
--- a/code/modules/projectiles/updated_projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/updated_projectiles/guns/specialist.dm
@@ -17,7 +17,7 @@
 	icon_state = "m42a"
 	item_state = "m42a"
 	max_shells = 15 //codex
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -17,7 +17,7 @@
 	icon_state = "m42a"
 	item_state = "m42a"
 	max_shells = 15 //codex
-	caliber = "10x28mm Caseless" //codex
+	caliber = "10x28mm" //codex
```
DominikPanic: This is not a proper fix as other guns are likely to have the same issues, a codex variable shouldn't affect regular gameplay.

spookydonut: this isnt a codex variable the comment is wrong

DominikPanic: Then the comment should be removed. Either way is fine, but right now the PR is not okay.

DominikPanic:
Could you add "Fixes: #2194" in the description too? Otherwise it won't get linked via git. The title should be the name of the fix rather than the bug report, since that doesn't really tell anyone viewing it what the PR does.

psykzz:
Should the magazine get updated to be the same caliber instead?
This means the sniper is going to start sharing ammo with a lot of other types of weapons, like sentries as well.

Im not sure if caseless was intentional for the caliber or not.

InterroLouis:
You can't actually remove bullets from sentry magazines, or put them into it. So mechanically, it wouldn't make any difference. So the real question is, do we want the M42A to be caseless or not.

code/modules/projectiles/updated_projectiles/guns/specialist.dm
```diff
@@ -206,7 +206,7 @@
 	icon_state = "m42c"
 	item_state = "m42c"
 	max_shells = 6 //codex
-	caliber = "10x99mm Caseless" //codex
+	caliber = "10x99mm" //codex
```
DominikPanic:
```suggestion
	caliber = "10x99mm"
```

code/modules/projectiles/updated_projectiles/guns/specialist.dm
DominikPanic:
```suggestion
	caliber = "10x28mm"
```

Rohesie: The vars having a comment pointing they are codex-related is not a bad thing IMO.

DominikPanic: It is misleading in making it seems like they're ONLY there for codex.

Rohesie: Ah, fair.

DominikPanic: Either removing the comment or making it more clear is fine, but right now it's not ideal.

PR #2335, opened 2019-07-20
Allow admins to set a distress beacon when sending

psykzz:
## About The Pull Request
Allow admins to set a distress beacon when sending

## Changelog
:cl:
add: Admins can choose what ERT to send, instead of it always being random.
/:cl:

https://streamable.com/zzopo

### DIFF
--- a/code/datums/emergency_calls/emergency_call.dm
+++ b/code/datums/emergency_calls/emergency_call.dm
@@ -66,10 +66,10 @@
 		to_chat(M, "<span class='attack'>You cannot join if you have Ghosted before this message.</span><br>")
 
 
-/datum/game_mode/proc/activate_distress()
-	picked_call = get_random_call()
+/datum/game_mode/proc/activate_distress(datum/emergency_call/chosen_call = null)
+	picked_call = chosen_call || get_random_call()
 
-	if(!picked_call) //Something went horribly wrong
+	if(!istype(picked_call)) //Something went horribly wrong
 		return FALSE
 
 	if(SSticker?.mode?.waiting_for_candidates) //It's already been activated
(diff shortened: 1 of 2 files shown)


code/datums/emergency_calls/emergency_call.dm
```diff
@@ -66,10 +66,10 @@
 		to_chat(M, "<span class='attack'>You cannot join if you have Ghosted before this message.</span><br>")
 
 
-/datum/game_mode/proc/activate_distress()
-	picked_call = get_random_call()
+/datum/game_mode/proc/activate_distress(datum/emergency_call/chosen_call = null)
```
Rohesie:
```suggestion
/datum/game_mode/proc/activate_distress(datum/emergency_call/chosen_call)
```
No need to set it to `null`, that's so by default.

code/datums/emergency_calls/emergency_call.dm
```diff
@@ -66,10 +66,10 @@
...
+	picked_call = chosen_call || get_random_call()
 
-	if(!picked_call) //Something went horribly wrong
+	if(!istype(picked_call)) //Something went horribly wrong
```
Rohesie: `get_random_call()` should be unable to return a wrong type. But if it does, then something indeed went horribly wrong and this could use a `CRASH()` message instead of a silent error.

Or just remove the check entirely.

DominikPanic: No CRASHes on my watch :p

DominikPanic: No Crashes on my watch >:)

Rohesie: No silent checks on mine.

DominikPanic: No check = no silent check

PR #2337, opened 2019-07-20
Fixes lobby countdown when admin sets it before startup

DominikPanic:


code/controllers/subsystem/ticker.dm
```diff
@@ -58,7 +58,7 @@ SUBSYSTEM_DEF(ticker)
 		if(GAME_STATE_STARTUP)
 			if(Master.initializations_finished_with_no_players_logged_in && !length(GLOB.clients))
 				return
-			start_at = world.time + (CONFIG_GET(number/lobby_countdown) * 10)
+			start_at = time_left || world.time + (CONFIG_GET(number/lobby_countdown) * 10)
```
Rohesie:
```suggestion
			start_at = time_left || world.time + (CONFIG_GET(number/lobby_countdown) SECONDS)
```

Issue #2346, opened 2019-07-20
Overwatch camera still appears under weeds


MoXuT0:
## Testmerges:

Round ID: 698
BYOND Version: 512.1464
Server revision compiled on:
Master commit: 6a6c3c0
The following pull requests are currently test merged:
#2335: 'Allow admins to set a distress beacon when sending' by psykzz at commit 5067be0
#2336: 'Makes some fax fields optional' by DominikPanic at commit 0a99ce3
#2326: 'Cleans up rav rage decay' by psykzz at commit 938dc8d
#2329: 'Fixes plasmacutter battery removal' by DominikPanic at commit 1b81dc2
#2312: 'Add human / xeno count to distress beacons for admins' by psykzz at commit 46e4d1b
#2314: 'Removes Big Red cameras' by DominikPanic at commit e161dc7
#2306: 'Battle rifle balance tweaks' by MoXuT0 at commit c286e12
#2307: 'Adds ammo counters to most spec weapons and launchers' by MoXuT0 at commit a9b8a22
#2300: 'Fixes HUD updating runtimes' by DominikPanic at commit c453545
#2301: 'Fixes vote action runtime' by DominikPanic at commit f0ba482
#2196: 'Adds weed decay' by psykzz at commit 0505125
#2264: 'Buckle parent logic' by Rohesie at commit 1bc389a
#2246: 'LV blockers rework' by Rohesie at commit 49632b1
#2224: 'Makes reagents use typepaths instead of IDs' by DominikPanic at commit 091e838
#2320: 'Ammo datum cleanup and runtime fix' by Rohesie at commit 20325ee
#2333: 'Fixes Issue #2194' by InterroLouis at commit d7136e2
#2325: 'Fixing the runtime for ssdirection' by psykzz at commit 2ac4a15

## Reproduction:

Forgot to take a picture but camera still appears under weeds.

DominikPanic:
Please take a picture. It does not do that for me.

DominikPanic:
https://puu.sh/DVj63/43abc79f11.png

MoXuT0:
![изображение](https://user-images.githubusercontent.com/34478337/61584792-7467af00-ab56-11e9-80a2-af4031b72624.png)

MoXuT0:
![изображение](https://user-images.githubusercontent.com/34478337/61584794-7c275380-ab56-11e9-85fb-d551fac4a64a.png)

DominikPanic:
[...]

PR #2354, opened 2019-07-20
Fixes overwatch eye layer issues

DominikPanic:
/tg/ still has this problem so someone can port it there
Fixes: #2346

### DIFF
--- a/code/_globalvars/misc.dm
+++ b/code/_globalvars/misc.dm
@@ -12,7 +12,7 @@ GLOBAL_LIST_INIT(duplicate_forbidden_vars, list("tag", "datum_components", "area
 	"group", "atmos_adjacent_turfs", "comp_lookup", "boxes", "click_border_start","storage_start", "storage_continue",\
 	"storage_end", "closer", "stored_start", "stored_continue", "stored_end", "click_border_end", "opened"))
 
-GLOBAL_LIST_INIT(ignored_atoms, list(/mob/dead, /obj/effect/landmark, /obj/docking_port, /obj/effect/particle_effect/sparks, /obj/effect/DPtarget, /obj/effect/supplypod_selector, /atom/movable/lighting_object))
+GLOBAL_LIST_INIT(ignored_atoms, typecacheof(list(/mob/dead, /obj/effect/landmark, /obj/docking_port, /obj/effect/particle_effect/sparks, /obj/effect/DPtarget, /obj/effect/supplypod_selector, /atom/movable/lighting_object))) 
 
 GLOBAL_VAR_INIT(current_date_string, "[num2text(rand(1,31))] [pick("January","February","March","April","May","June","July","August","September","October","November","December")], [GAME_YEAR]")
 
--- a/code/_globalvars/misc.dm
+++ b/code/_globalvars/misc.dm
@@ -12,6 +12,8 @@ GLOBAL_LIST_INIT(duplicate_forbidden_vars, list("tag", "datum_components", "area
 	"group", "atmos_adjacent_turfs", "comp_lookup", "boxes", "click_border_start","storage_start", "storage_continue",\
 	"storage_end", "closer", "stored_start", "stored_continue", "stored_end", "click_border_end", "opened"))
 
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/computer/camera_advanced.dm
```diff
@@ -200,11 +200,21 @@
...
-				user_image = image(icon,loc,icon_state,FLY_LAYER)
-				eye_user.client.images += user_image
+		if(visible_icon && eye_user.client)
+			eye_user.client.images -= user_image
+			var/atom/top
+			for(var/i in loc)
+				var/atom/A = i
+				if(!top)
+					top = loc
+				if(A.type in GLOB.ignored_atoms)
```
Rohesie: Will this work for `/mob/dead/observer` for example? If not, how about a typecache?

DominikPanic: This would have to be multiple typecaches and affect performance negatively, I'll just make it mob/dead/observer instead of mob/dead since thats the only one we have

Rohesie: The performance impact would be only in setting it up, though.
The idea is that you make a typecache of every item on the list, merge all of them into a single global list, and then use that for the check. Then performance should be unaffected.
Makes sense?

spookydonut: Typecacheof makes a list that has all subtypes of the supplied types unless you specifically tell it not to

spookydonut: Yeah looking again just use a typecache in the glob and use is_type_in_typecache, it's faster than this, O(logn)vs O(n)

code/_globalvars/misc.dm
```diff
@@ -12,6 +12,8 @@ GLOBAL_LIST_INIT(duplicate_forbidden_vars, list("tag", "datum_components", "area
 	"group", "atmos_adjacent_turfs", "comp_lookup", "boxes", "click_border_start","storage_start", "storage_continue",\
 	"storage_end", "closer", "stored_start", "stored_continue", "stored_end", "click_border_end", "opened"))
 
+GLOBAL_LIST_INIT(ignored_atoms, list(/mob/dead, /obj/effect/landmark, /obj/docking_port, /obj/effect/particle_effect/sparks, /obj/effect/DPtarget, /obj/effect/supplypod_selector, /atom/movable/lighting_object))
```
spookydonut:
```suggestion
GLOBAL_LIST_INIT(ignored_atoms, typecacheof(list(/mob/dead, /obj/effect/landmark, /obj/docking_port, /obj/effect/particle_effect/sparks, /obj/effect/DPtarget, /obj/effect/supplypod_selector, /atom/movable/lighting_object))) 
```

code/game/objects/machinery/computer/camera_advanced.dm
spookydonut:
```suggestion
				if(is_type_in_typecache(A.type, GLOB.ignored_atoms)) 
```

code/_globalvars/misc.dm
DominikPanic: Won't this also apply to all subtypes of these, potentially being wasteful?

spookydonut: it will replicate the behaviour of a long chain of istype() calls and is faster, if you dont want it to apply to subtypes theres an option you can supply to typecacheof(), `typecacheof(list(), FALSE, FALSE)`

spookydonut: or just `typecacheof(list(), only_root_path=TRUE)`

spookydonut: Even if you include the subtypes its still faster.

spookydonut: Unless there's more than 10 million subtypes.

spookydonut: log(10,000,000) = 7

spookydonut: typecaches are FAST.

PR #2355, opened 2019-07-21
Fixes cells icon updating

DominikPanic:
Fixes: #2339

### DIFF
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -6,17 +6,16 @@
 	. = ..()
 	charge = maxcharge
 
-	updateicon()
+	update_icon()
 
-/obj/item/cell/proc/updateicon()
-	overlays.Cut()
-
+/obj/item/cell/update_icon()
+	cut_overlays()
 	if(charge < 0.01)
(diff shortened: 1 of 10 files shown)


code/modules/power/cell.dm
```diff
@@ -6,17 +6,16 @@
...
-	overlays.Cut()
+	update_icon()
 
+/obj/item/cell/update_icon()
+	cut_overlays()
 	if(charge < 0.01)
 		return
-	else if(charge/maxcharge >=0.995)
-		overlays += image('icons/obj/power.dmi', "cell-o2")
+	else if(charge / maxcharge >= 0.995)
```
spookydonut: Breaks blame

Rohesie: And?

spookydonut: just observing.

DominikPanic: From the contributor guidelines:
```
Operators that should be separated by spaces

    Boolean and logic operators like &&, || <, >, ==, etc (but not !)
```

Issue #2359, opened 2019-07-21
you cant plasmacut floodlights


Pariah919:
titles

Rohesie:
Were you ever able to?

DominikPanic:
`resistance_flags = UNACIDABLE`
It's only meant to not be melted.

PR #2361, opened 2019-07-21
Fixes plasmacutter interactions

DominikPanic:
Fixes: #2359
This may need more testing since it can break stuff, but we should move towards doing it this way in the future rather than combining the two checks.


DominikPanic:
Should be ready.

PR #2368, opened 2019-07-21
Alt+click containers

Rohesie:
The idea is to move away from these ugly toggle verbs and into the alt+click pattern.
Blessed click refactor.

## Changelog
:cl:
add: Alt+clicking storage containers will remove the last placed item in them. This works for pill bottles to make the medic life easier, ammo belts, and several other things.
/:cl:


Metroidlover:
Yes, god yes. Finally no more opening every individual pill bottle every time.

spookydonut:
Should require it to not be on the ground to avoid qol loss of alt clicking anywhere on a turf to get the contents list.
Alternatively make it a preference for the client

Rohesie:
Change request done, it no longer works on stuff whose `loc` is a `turf`.

PR #2378, opened 2019-07-21
No shrike ventcrawl

Rohesie:
## Changelog
:cl:
balance: Shrikes can no longer ventcrawl.
/:cl:

Should help with the lowpop early round ends when all xenos ventcrawl.


spookydonut:
Did you move the crawl verb from them or did someone fix that bug

Rohesie:
I'm unaware of what bug is that.

Issue #2381, opened 2019-07-22
to_chat called with invalid input type


Rohesie:
## Testmerges:

Master commit: 5c675ae77c070c9c9ef9c048b8fa71404c42d3d1
The following pull requests are currently test merged:
#2370: 'Fixes queens losing strong acid when going into ovi mode.' by Rohesie at commit 071d09ea1d
#2366: 'Fixes a slight access issue' by DominikPanic at commit 7b1dba02eb
#2361: 'Fixes plasmacutter interactions' by DominikPanic at commit 7c509c5688
#2362: 'Adds additional null protection from /tg/' by DominikPanic at commit e58f5b6112
#2363: 'Fixes graceful shutdown pinging for real' by DominikPanic at commit 3b12902235
#2358: 'Removes emergency shutter admin message' by DominikPanic at commit d1b19143cc
#2354: 'Fixes overwatch eye layer issues' by DominikPanic at commit d1f0942443
#2352: 'Fixes simple window deconstruction' by DominikPanic at commit c1b5891fce
#2310: 'Clarifies the hotkey preference name, removes obsolete SSinput things' by DominikPanic at commit aa026dad74
#2333: 'Weapon/Magazine Caliber Fix' by InterroLouis at commit 3fef03fdfc
#2224: 'Makes reagents use typepaths instead of IDs' by DominikPanic at commit 5bee0a730d
#2246: 'LV blockers rework' by Rohesie at commit 49632b1983
#2264: 'Buckle parent logic' by Rohesie at commit 1bc389af0f
#2196: 'Adds weed decay' by psykzz at commit f0533e68bb
#2301: 'Fixes vote action runtime' by DominikPanic at commit f0ba482387
#2306: 'Battle rifle balance tweaks' by MoXuT0 at commit c286e12984
#2353: 'Fixes some topic interaction issues' by DominikPanic at commit 49a36fb0a8
#2357: 'Fixes a gun action runtime' by DominikPanic at commit 96acd02108
#2360: 'Fixes a null in player list if the player disconnects at a bad time' by DominikPanic at commit fe50c5d9e5
#2128: 'Full auto' by Rohesie at commit 90e1c820c5
#2368: 'Alt+click containers' by Rohesie at commit 61e192dbb4
#2378: 'No shrike ventcrawl' by Rohesie at commit 62d9d77869

Local commit: 95918627cc566640a231ddeda532c15bd119dc15

## Reproduction:

```
[...]

PR #2386, opened 2019-07-22
Converts show_messages into visible/audible_message

DominikPanic:
Fixes: #2381


DominikPanic:
Gitkraken opened a PR twice. Whoops.

Issue #2384, opened 2019-07-22
Synths still blind/deaf at full health


grinkgo:
## Testmerges:
#2378 #2368 #2128 #2306 #2301 #2196 #2264 #2246 #2224 #2333 #2310 #2354 #2361 #2366 #2370

## Reproduction:

Unsure if this happens every time, but when a synth becomes blind or deaf, their vision/hearing does not restore even after repairing to 100% health.

PR #2387, opened 2019-07-22
Fixes synthetic blind and deafness issues

DominikPanic:
FIxes: #2384

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -45,6 +45,7 @@
 #define isvoxarmalis(H) (is_species(H, /datum/species/vox/armalis))
 #define isIPC(H) (is_species(H, /datum/species/machine))
 #define issynth(H) (is_species(H, /datum/species/synthetic) || is_species(H, /datum/species/early_synthetic))
+#define isspeciessynthetic(H) (H.species.species_flags & IS_SYNTHETIC)
 #define ismoth(H) (is_species(H, /datum/species/moth))
 #define ishumanbasic(H) (is_species(H, /datum/species/human))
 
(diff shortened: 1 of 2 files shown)


DominikPanic:
> 
> 
> Use issynth

Issynth is more wasteful, these are already human and using the bitflag is more universal.

Rohesie:
But by using `issynth(H)` we can modify how the whole thing works more easily. It's a define after all.

DominikPanic:
> 
> 
> But by using `issynth(H)` we can modify how the whole thing works more easily. It's a define after all.

issynth is used for when you have mob that you aren't sure is a human, here we are 100% sure it's a human since it's defined on the human path, also this is more efficient. That define has other purposes, this is not it.

Rohesie:
I'm not sure about it. What's the basis of your request, @spookydonut ?

spookydonut:
issynth is used plenty of places already where you already know its a human how is this any different, its more readable and easier to change the behaviour of it without hard coding behaviour like in this pr.

DominikPanic:
Just because it's already used in those places doesn't mean it's ideal. You have to agree that the performance impact of issynth is much worse than a bitflag comparison on the species.

spookydonut:
> 
> 
> Just because it's already used in those places doesn't mean it's ideal. You have to agree that the performance impact of issynth is much worse than a bitflag comparison on the species.

OH NO NOT AN EXTRA ISTYPE CHECK IN INFREQUENTLY CALLED CODE THINK OF THE CHILDREN.

its more maintainable, its more readable, this code isnt performance critical.

DominikPanic:
For the readbility and maintainability argument, a second macro could be created.

Rohesie:
A second macro would be a good compromise.

PR #2392, opened 2019-07-22
More spawn

spookydonut:
more spawns for the spawn god


Rohesie:
A quick review shows no glaring issues.
Is this tested locally or will it require server testing?
It will conflict with the reagent typepaths PR.

PR #2397, opened 2019-07-23
Fixes facehugger description

DominikPanic:
It wasn't a code issue, just the unconscious stat and dead stat were combined, and egg facehuggers start in stasis. The facehugger not curling up there is intended, since they stay like that for a while then go back into the egg.
Fixes: #2369


code/modules/mob/living/carbon/xenomorph/facehuggers.dm
```diff
@@ -121,10 +121,12 @@
 
 /obj/item/clothing/mask/facehugger/examine(mob/user)
 	. = ..()
-	if(stat != CONSCIOUS)
-		to_chat(user, "<span class='warning'>[src] is not moving.</span>")
-	else
-		to_chat(user, "<span class='danger'>[src] seems to be active.</span>")
+	if(stat == CONSCIOUS)
```
Rohesie:
```
switch(stat)
	if(CONSCIOUS)
		stuff()
	if(UNCONSCIOUS)
		stuff()
```

DominikPanic: That increases more tabs

Rohesie: Yea, `switch` does that by default, but the point is that `if-else` chains should be switches if they share the same common var as it's a single check instead of potentially several.
Tabs are not bad per se, else we would never have a `switch`.

PR #2400, opened 2019-07-23
Removes hunter games

MoXuT0:
## About The Pull Request

Per title.

## Changelog

:cl:
del: Removed hunter games.
/:cl:


Rohesie:
What is the purpose of this?

MoXuT0:
Dead game mode whose only purpose is to pollute maps with retarded landmarks. It’s not working due to it counting one additional person on alive humans list, landmarks with random loot don’t produce anything, maps are not made for it. If there would be battle royale mode, it should have unique map/mode. But for right now it’s dead weight that doesn’t gets used at all, lowpop mode is already in works.

Rohesie:
Oh well, looks like I'm outnumbered here.

DominikPanic:
:owl: :gun:

Issue #2404, opened 2019-07-23
Your groin is tinglingblistered.


Shadow-Quill:
## Testmerges:

#2402 #2327 #1780 #2310 #2361 #2398 #2397 #2396 #2333 #2224 #2394

## Reproduction:

Examine self while burned and brute damaged.

PR #2406, opened 2019-07-23
Fixes tinglingblistered

DominikPanic:
Fixes: #2404


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -190,6 +190,8 @@
 
 				if(halloss > 0)
 					status = "tingling"
+					if(brutedamage > 0 || burndamage > 0)
+						status += " and "
 
 				if(brutedamage > 0)
 					status = "bruised"
```
Rohesie: This will override the status.

Issue #2412, opened 2019-07-24
 Runtime in qdel.dm, line 14: bad index


Rohesie:
```
[20:36:30] Runtime in qdel.dm, line 14: bad index
proc name: qdel null wrapper (/proc/______qdel_null_wrapper)
src: null
call stack:
qdel null wrapper(r_hand (/datum/limb/r_hand), /datum/effect_system/spark_spr... (/datum/effect_system/spark_spread))
/datum/callback (/datum/callback): InvokeAsync()
Timer (/datum/controller/subsystem/timer): fire(0)
Timer (/datum/controller/subsystem/timer): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop()
Master (/datum/controller/master): StartProcessing(0)
```

spookydonut:
#2395

PR #2418, opened 2019-07-24
Fixes qdel_null issue

DominikPanic:
Fixes: #2412

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -937,7 +937,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 			spark_system.set_up(5, 0, owner)
 			spark_system.attach(owner)
 			spark_system.start()
-			QDEL_NULL_IN(src, spark_system, 1 SECONDS)
+			QDEL_IN(spark_system, 1 SECONDS)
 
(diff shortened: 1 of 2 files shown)


code/modules/organs/limbs.dm
```diff
@@ -937,7 +937,8 @@ Note that amputating the affected organ does in fact remove the infection from t
 			spark_system.set_up(5, 0, owner)
 			spark_system.attach(owner)
 			spark_system.start()
-			QDEL_NULL_IN(src, spark_system, 1 SECONDS)
+			spawn(1 SECONDS)
```
Rohesie: `INVOKE_NEXT_TICK` instead, to a proc that does the QDEL_NULL.

Rohesie: Actually just a callback.

DominikPanic: It can't be a callback though, that's the entire reason I'm making this PR

Rohesie:
`error: undefined field: "spark_system" on /datum/limb`

DominikPanic:
> 
> 
> `error: undefined field: "spark_system" on /datum/limb`

I have no idea how to fix this.

spookydonut:
See #2395 for the fix

PR #2421, opened 2019-07-25
Update blackbox from TG

psykzz:


DominikPanic:
I kind of wanted to do this myself. We won't just be using what /tg/ does, so expect my PR to be very different from this.

PR #2425, opened 2019-07-26
Ranks adjustment

Maciekkub:
## About The Pull Request
 Swaps some of the ranks to be more accurate without actually changing the chain of command (Makes it more visible in terms of IO>CSE/CMO)

## Why It's Good For The Game
More realistic look at the ranks as well as more visible chain of succession between the command staff

## Changelog
:cl:
add: New rank CMN - Crewman, for Cargo technicians as it's unfit for sergeants' only job to be moving crates
tweak: change in the ranks of IO to Lieutenant Commander, CSE to Lieutenant Junior Grade together with the CMP, Cargo technician to Crewman, SL to Sergeant Spec to Corporal Sgner to Lance Corporal
/:cl:


code/datums/jobs/access.dm
```diff
@@ -219,6 +219,8 @@
 			. = size ? "Dr. " : "Doctor"
 		if("CCMO")
 			. = size ? "Prof. " : "Professor"
+		if ("CMN")
```
DominikPanic:
```suggestion
		if("CMN")
```
Whitespace is usually not enforced but consistency is important, and the rest of the lines use this style

Maciekkub: I see! I'll keep an eye out for that the next time i'll PR something

grinkgo:
For clarity just change the changelog to say the ranks instead of the grade. Otherwise we gucci.

PR #2448, opened 2019-07-27
Cargo technician rework

Maciekkub:
## About The Pull Request

Gives more choices for the cargo technician to do except for supply, also increases the slots from 2 to 4

## Why It's Good For The Game
Not only it's been a suggested change, then also it's a major fluff to the cargo technician and the ship, who now becomes a Serviceman and is tasked mainly with still being supply technician, but can also take upon janitorial or cooking tasks
## Changelog
:cl:
add: Locker of a chef, requiring cargo tech access, both placed in the kitchen backrooms of upper and lower deck (2 in total)
add: Added maint access to CT so he can access janitorial and hydroponics
del: one fridge from kitchen backroom lower deck to create space for the locker
tweak: Cargo tech is now called serviceman, his armour and hat have been put into CT locker, which is now named Serviceman's locker
/:cl:

### DIFF
--- a/code/datums/jobs/access.dm
+++ b/code/datums/jobs/access.dm
@@ -219,6 +219,8 @@
 			. = size ? "Dr. " : "Doctor"
 		if("CCMO")
 			. = size ? "Prof. " : "Professor"
+		if ("CMN")
+			. = size ? "CMN" : "Crewman"
 		if("PMC1")
 			. = size ? "PMC " : "PM Contractor"
 		if("PMC2")
@@ -237,6 +239,8 @@
 			. = size ? "PVT " : "Private"
 		if("E2")
 			. = size ? "PFC " : "Private First Class"
+		if("E2.5")
+			. = size ? "SPR" : "Sapper"
 		if("E3")
 			. = size ? "LCPL " : "Lance Corporal"
 		if("E4")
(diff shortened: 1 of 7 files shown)


code/datums/jobs/access.dm
```diff
@@ -219,6 +219,8 @@
 			. = size ? "Dr. " : "Doctor"
 		if("CCMO")
 			. = size ? "Prof. " : "Professor"
+		if ("CMN")
```
DominikPanic:
```suggestion
		if("CMN")
```

code/datums/jobs/job/shipside.dm
```diff
@@ -429,23 +429,22 @@ A happy ship is a well-functioning ship."})
 //Cargo Tech
 /datum/job/requisitions/tech
 	title = CARGO_TECHNICIAN
-	paygrade = "PO"
-	comm_title = "CT"
-	total_positions = 2
+	paygrade = "CMN"
+	comm_title = "Ord"
```
DominikPanic:
```suggestion
	comm_title = "Ord"
```

Maciekkub:
> Considering the rank rework, CTs are now Crewman. Might as well just make that their role as well instead of Crewman Serviceman. We'll just call 'em Crewmen. Also if we're expanding the CT's role to be more general, we might as well just merge them with the MTs.

got you

PR #2460, opened 2019-07-27
Makes humans jitter when larval stage increases

Shadow-Quill:
## About The Pull Request

So you can somewhat tell when you're about to burst.
Also removes a check that Lak said I could remove.

## Why It's Good For The Game

It isn't fun to suddenly drop unconscious and burst without warning.

## Changelog
:cl:
tweak: Humans now shake when a larvae inside them grows older.
/:cl:


Shadow-Quill:
the mult may need adjusting, we'll see

PR #2469, opened 2019-07-28
Call of the burrowed now tries to spawn observers first

DominikPanic:
:cl: LaKiller8
qol: Call of the burrowed now tries to spawn observers first, by their time of death like other methods to do get xeno.
/:cl:

PR #2490, opened 2019-07-29
Less call of the burrowed spam.

Rohesie:
The ability can only be used when there are burrowed larvas available.

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -179,3 +179,6 @@
 
 // Shuttles
 #define isshuttleturf(T) (length(T.baseturfs) && (/turf/baseturf_skipover/shuttle in T.baseturfs))
+
+// Xeno hives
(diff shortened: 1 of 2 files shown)


code/__DEFINES/is_helpers.dm
```diff
@@ -179,3 +179,6 @@
 
 // Shuttles
 #define isshuttleturf(T) (length(T.baseturfs) && (/turf/baseturf_skipover/shuttle in T.baseturfs))
+
+// Xeno hives
+#define isnormalhive(hive) (istype(hive, /datum/hive_status/normal))
```
DominikPanic: Just implement my solution from #2469 regarding hives instead of making a hacky helper, adding such functionality to the rest of the hives is a good thing.

Rohesie: Dejavu.

DominikPanic: Hehe

PR #2493, opened 2019-07-29
Slightly condenses the AI tracking href

DominikPanic:
It's not /tg/ but it sure is a hella lot cleaner

### DIFF
--- a/code/game/say.dm
+++ b/code/game/say.dm
@@ -50,8 +50,9 @@ GLOBAL_LIST_INIT(freqtospan, list(
 	//Start name span.
 	var/spanpart2 = "<span class='name'>"
 	//Radio freq/name display
-	var/job = speaker.GetJob()
-	var/freqpart = radio_freq ? "\[[get_radio_name(radio_freq)][job ? " ([job])": ""]\] " : ""
+	var/freqpart = compose_freq(speaker, radio_freq)
+	//Job display
+	var/jobpart = compose_job(speaker, message_language, raw_message, radio_freq)
 	//Speaker name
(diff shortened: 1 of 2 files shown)


code/game/say.dm
```diff
@@ -50,8 +50,9 @@ GLOBAL_LIST_INIT(freqtospan, list(
 	//Start name span.
 	var/spanpart2 = "<span class='name'>"
 	//Radio freq/name display
-	var/job = speaker.GetJob()
-	var/freqpart = radio_freq ? "\[[get_radio_name(radio_freq)][job ? " ([job])": ""]\] " : ""
+	var/freqpart = compose_freq(speaker, radio_freq)
+	//Job display
+	var/jobpart = compose_job(speaker, message_language, raw_message, radio_freq)
```
Rohesie: Why not defines?

DominikPanic: Give an example and I'll implement it.

Rohesie:
```
#define CMSG_FREQPART compose_freq(speaker, radio_freq)
#define CMSG_JOBPART compose_job(speaker, message_language, raw_message, radio_freq)
/atom/movable/proc/compose_message(atom/movable/speaker, datum/language/message_language, raw_message, radio_freq, list/spans, message_mode, face_name = FALSE)
(...)
return "[spanpart1][spanpart2][CMSG_FREQPART][languageicon][CMSG_JOBPART][namepart][endspanpart][messagepart]"
(...)
#undef CMSG_FREQPART
#undef CMSG_JOBPART
```

DominikPanic: I don't really see it as being cleaner, but if you really want it, I can do it

PR #2495, opened 2019-07-29
Adds camera tracking to the overwatch console

DominikPanic:
QoL overtime

### DIFF
--- a/code/game/objects/machinery/computer/camera_advanced.dm
+++ b/code/game/objects/machinery/computer/camera_advanced.dm
@@ -13,6 +13,7 @@
 	var/list/actions
 	var/mob/living/tracking_target
 	var/tracking = FALSE
+	var/cameraticks = 0
 
 
 /obj/machinery/computer/camera_advanced/Initialize()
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/computer/camera_advanced.dm
```diff
@@ -158,6 +160,49 @@
...
+
+	if(!target || !target.can_track(current_user))
+		to_chat(current_user, "<span class='warning'>Target is not near any active cameras.</span>")
+		tracking_target = null
+		return
+
+	to_chat(current_user, "<span class='notice'>Now tracking [target.get_visible_name()] on camera.</span>")
+
+	var/cameraticks = 0
+	spawn(0)
```
Rohesie: Invoke async a `set waitfor = FALSE` proc.

code/game/objects/machinery/computer/camera_advanced.dm
```diff
@@ -158,6 +160,49 @@
...
+				if(!cameraticks)
+					to_chat(current_user, "<span class='warning'>Target is not near any active cameras. Attempting to reacquire...</span>")
+				cameraticks++
+				if(cameraticks > 9)
+					tracking_target = null
+					to_chat(current_user, "<span class='warning'>Unable to reacquire, cancelling track...</span>")
+					tracking = FALSE
+					return
+				else
+					sleep(10)
```
Rohesie: stoplag

code/game/objects/machinery/computer/camera_advanced.dm
DominikPanic: I'll ask ninja about this, since this is the original /tg/ camera tracking implementation that's still used.

code/game/objects/machinery/computer/camera_advanced.dm
```diff
@@ -158,6 +160,55 @@
 	eyeobj.setLoc(eyeobj.loc)
 
 
+/obj/machinery/computer/camera_advanced/proc/track(mob/living/target)
+	if(!istype(target))
+		return
+
+	INVOKE_ASYNC(src, .track_actual, target)
```
Rohesie:
```suggestion
	INVOKE_ASYNC(src, .proc/track_actual, target)
```

Rohesie: Or just call it regularly. `INVOKE_ASYNC` is to make it `set waitfor = FALSE`, but this is already done in the proc.

code/game/objects/machinery/computer/camera_advanced.dm
```diff
@@ -158,6 +160,55 @@
...
+			if(!cameraticks)
+				to_chat(current_user, "<span class='warning'>Target is not near any active cameras. Attempting to reacquire...</span>")
+			cameraticks++
+			if(cameraticks > 9)
+				tracking_target = null
+				to_chat(current_user, "<span class='warning'>Unable to reacquire, cancelling track...</span>")
+				tracking = FALSE
+				return
+			else
+				sleep(10)
```
Rohesie:
```suggestion
				stoplag(10)
```

code/game/objects/machinery/computer/camera_advanced.dm
```diff
@@ -158,6 +160,55 @@
...
+
+	if(!target || !target.can_track(current_user))
+		to_chat(current_user, "<span class='warning'>Target is not near any active cameras.</span>")
+		tracking_target = null
+		return
+
+	to_chat(current_user, "<span class='notice'>Now tracking [target.get_visible_name()] on camera.</span>")
+
+	var/cameraticks = 0
+	while(tracking_target == target)
```
ninjanomnom: This needs to be done with standard processing from a subsystem

Rohesie: You invoked the dreaded ninjanomnom reviews, which are much more demanding than mine.
May this be a lesson to never wake up this creature from its slumber, in his house at R'lyeh.

DominikPanic: ~~I should just find a way to call the ai tracking proc~~

DominikPanic: Is a subsystem really necessary for something as simple as this tracking? It already uses stoplag(), it won't be used much per round.

PR #2498, opened 2019-07-29
Adds admin notification for people who DC'd after 15 minutes

DominikPanic:
Also unshitcodes away_time

### DIFF
--- a/code/modules/mob/living/carbon/human/login.dm
+++ b/code/modules/mob/living/carbon/human/login.dm
@@ -1,6 +1,6 @@
 /mob/living/carbon/human/Login()
 	. = ..()
 	species?.handle_login_special(src)
-	if(away_time)
-		deltimer(away_time)
-		away_time = null
+	if(afk_timer_id)
+		deltimer(afk_timer_id)
+		afk_timer_id = null
--- a/code/modules/mob/living/carbon/human/login.dm
+++ b/code/modules/mob/living/carbon/human/login.dm
@@ -1,3 +1,6 @@
 /mob/living/carbon/human/Login()
-	..()
-	if(species) species.handle_login_special(src)
+	. = ..()
+	species?.handle_login_special(src)
+	if(away_time)
+		deltimer(away_time)
+		away_time = null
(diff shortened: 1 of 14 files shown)


code/modules/mob/living/carbon/human/login.dm
```diff
@@ -1,3 +1,6 @@
 /mob/living/carbon/human/Login()
-	..()
-	if(species) species.handle_login_special(src)
+	. = ..()
+	species?.handle_login_special(src)
+	if(away_time)
+		deltimer(away_time)
+		away_time = 0
```
Rohesie:
```suggestion
		away_time = null
```

Rohesie: Will need other references to be changed, since it's no longer a time reference, but a timer ID.

DominikPanic: The default value is 0, though I suppose this would be helpful for clarity that it triggered.

code/modules/mob/living/carbon/xenomorph/logout.dm
```diff
@@ -0,0 +1,3 @@
+/mob/living/carbon/xenomorph/Logout()
+	. = ..()
+	addtimer(CALLBACK(src, .proc/handle_afk_takeover), XENO_AFK_TIMER)
```
Rohesie: Add a reference same as the human one, make the timer stoppable and delete it on login?

DominikPanic: For xenos away_time is used as a number so when xenos spawn with it, they can be taken away instantly.

Rohesie: Using the same var to represent two things can get messy.

DominikPanic: This doesn't need to be a timer_id stoppable since if they get qdeleted it already gets properly unsed, and take a look at the proc it calls, it checks stuff pretty well.

Rohesie: That is fine for now. In the future things might change, so having a reference to the timer would be positive.
Even if that's not desired, if we change what the variable represents in one place it should change everywhere. A same variable representing two different things is very confusing.

PR #2503, opened 2019-07-30
Fixes Map Issues

Alterist:
Fixes some groundmap and pipenet issues

- Removed stacking turfs on Big Red and implemented /turf/open/space/basic , Map Merge also removed some unused bits

- Fixed Pipenet issues and added missing pipes on Ice Colony

- Removed Atmos and bad pipes on Prison

PR #2504, opened 2019-07-31
Silos, Nukes, Pinpointers.

Rohesie:
Atomizing the Crash game mode objects a bit, and making them generic to be used in any other.

The plan is to add them perhaps as game-mode dependent landmarks to the maps, but that will require for #2503 to be merged to avoid conflicts.

Did a few small improvements, such as showing ghosts the timer on the nuke, and making it start anchored to avoid xeno pushing memes.

Disks are no longer unacidable and indestructible by default, given that more can be made on the computers.

Advanced pinpointers can track both disks and disk computers, depending on the mode. It starts pointing at computers by default, which is the usual use one would expect for them.

The nuke kills (gives twice the max health amount in fire damage) every mob alive in the z-level when it explodes, and triggers a global signal, which can be used by the game mode to declare completion.

Resin silos work as they used to on the game mode: if there's no queen in ovi or shrike calling, then larvas can join into one of the silos.


code/game/objects/effects/countdown.dm
```diff
@@ -0,0 +1,76 @@
...
+
+	if(displayed_text)
+		maptext = "<font size = [text_size]>[displayed_text]</font>"
+	else
+		maptext = null
+
+/obj/effect/countdown/Destroy()
+	attached_to = null
+	STOP_PROCESSING(SSobj, src)
+	. = ..()
```
DominikPanic:
```suggestion
	return ..()
```

code/game/objects/effects/countdown.dm
```diff
@@ -0,0 +1,76 @@
...
+/obj/effect/countdown/ex_act(severity, target) //immune to explosions
+	return
+
+/obj/effect/countdown/nuclearbomb
+	name = "nuclear bomb countdown"
+	color = "#81FF14"
+
+/obj/effect/countdown/nuclearbomb/get_value()
+	var/obj/machinery/nuclearbomb/N = attached_to
+	if(!istype(N))
```
DominikPanic: Wouldn't it be more efficient to do `if(!attached_to) return`?

code/game/objects/items/pinpointer.dm
```diff
@@ -7,137 +11,114 @@
...
-			to_chat(usr, "You set the pinpointer to locate [locationx],[locationy]")
 
+/obj/item/pinpointer/advpinpointer/proc/do_toggle_mode(new_mode)
+	switch(new_mode)
+		if(PINPOINTER_MODE_DISK)
+			tracked_list = GLOB.nuke_disk_list
+		if(PINPOINTER_MODE_DISK_GENERATOR)
+			tracked_list = GLOB.nuke_disk_generators
+		else
+			CRASH("do_toggle_mode called with an invalid new_mode")
```
DominikPanic: Will this ever happen? Psy didn't have this in his PR.

code/game/objects/machinery/computer/nuke_disk_generator.dm
```diff
@@ -0,0 +1,169 @@
...
+	return ..()
+
+
+/obj/machinery/computer/nuke_disk_generator/process()
+	. = ..()
+	if(. || !current_timer)
+		return
+
+	deltimer(current_timer)
+	current_timer = null
```
DominikPanic: could return PROCESS_KILL so it doesn't waste another tick

code/game/objects/machinery/computer/nuke_disk_generator.dm
```diff
@@ -0,0 +1,169 @@
...
+	if(.)
+		return
+
+	if(href_list["generate"])
+		if(printing || current_timer)
+			to_chat(usr, "<span class='warning'>A program is already running.</span>")
+			return
+		if(reprintable)
+			printing = TRUE
+			addtimer(VARSET_CALLBACK(src, printing, FALSE), printing_time) // TODO Change to larger time
```
DominikPanic: No TODOs

code/game/objects/machinery/computer/nuke_disk_generator.dm
```diff
@@ -0,0 +1,169 @@
...
+
+			usr.visible_message("[usr] started a program to regenerate a nuclear disk code.", "You started a program to generate a nuclear disk code.")
+			if(!do_after(usr, printing_time, TRUE, src, BUSY_ICON_GENERIC, null, null, CALLBACK(src, /datum.proc/process)))
+				return
+
+			print_disc()
+			return
+
+		printing = TRUE
+		addtimer(VARSET_CALLBACK(src, printing, FALSE), segment_time) // TODO Change to larger time
```
DominikPanic: Same

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
```diff
@@ -550,8 +575,15 @@ to_chat will check for valid clients itself already so no need to double check f
 	if(!(mother in possible_mothers))
 		to_chat(xeno_candidate, "<span class='warning'>This mother is not in a state to receive us.</span>")
 		return FALSE
+	return do_spawn_larva(xeno_candidate, mother.loc)
```
DominikPanic: Wouldn't a get_turf be safer here?

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
```diff
@@ -500,7 +500,7 @@ to_chat will check for valid clients itself already so no need to double check f
 	qdel(L)
 
 
-// This proc checks for available mothers and offers a choice if there's more than one.
+// This proc checks for available spawn points and offers a choice if there's more than one.
 /datum/hive_status/normal/proc/attempt_to_spawn_larva(mob/xeno_candidate)
```
DominikPanic: Do the silos currently try to get an observer via get_alien_candidate first? If yes, resolve this, if not, make it that way please.

code/game/objects/machinery/nuclearbomb.dm
```diff
@@ -1,339 +1,298 @@
...
-				return
-			user.visible_message("<span class='notice'>[user] begins unwrenching the anchoring bolts on [src].</span>",
-			"<span class='notice'>You begin unwrenching the anchoring bolts.</span>")
 
-			if(!do_after(user, 50, TRUE, src, BUSY_ICON_BUILD))
-				return
+/obj/machinery/nuclearbomb/proc/BurnEveryoneOnZLevel(z)
+	if(!z)
+		return
+	for(var/mob/living/victim in GLOB.alive_mob_list)
```
DominikPanic: Ghosts and new players cannot be alive, this can be optimized.

code/game/objects/machinery/nuclearbomb.dm
```diff
@@ -1,339 +1,298 @@
...
+/obj/machinery/nuclearbomb/proc/play_cinematic()
+	GLOB.enter_allowed = FALSE
+	priority_announce("DANGER. DANGER. Planetary Nuke Activated. DANGER. DANGER. Self destruct in progress. DANGER. DANGER.", "Priority Alert")
+	SEND_SOUND(world, pick('sound/theme/nuclear_detonation1.ogg','sound/theme/nuclear_detonation2.ogg'))
 
-			if(!do_after(user, 40, TRUE, src, BUSY_ICON_BUILD, extra_checks = CALLBACK(WT, /obj/item/tool/weldingtool/proc/isOn)) || !WT.remove_fuel(5, user))
-				return
+	for(var/x in GLOB.player_list)
+		var/mob/M = x
+		if(isobserver(M))
```
DominikPanic: This will still happen for new_player

code/game/objects/machinery/nuclearbomb.dm
```diff
@@ -1,339 +1,298 @@
...
-	else if(iswirecutter(I) || ismultitool(I))
-		if(!CHECK_BITFIELD(machine_stat, PANEL_OPEN))
-			return
-
-		nukehack_win(user)
-
-	else if(extended && istype(I, /obj/item/disk/nuclear))
-		if(!user.transferItemToLoc(I, src))
+/obj/machinery/nuclearbomb/process()
+	if(timer_enabled)
```
DominikPanic: Early return?

code/game/objects/items/pinpointer.dm
Rohesie: Debug check for the future. It's within a `switch`, so it's free.

DominikPanic: Evil, but acceptable.

code/game/objects/machinery/computer/nuke_disk_generator.dm
Rohesie: This is copypasted. Let's see what @psykzz wanted to do here and I'll try to get it done.

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
Rohesie: Silos are permanent spawn points, they are passive.
What can be done is for new burrowed larva additions to try to find observers and mothers, and then spawn.
Want that?

code/game/objects/machinery/nuclearbomb.dm
Rohesie:
```
/mob/Initialize()
	GLOB.mob_list += src
	if(stat == DEAD)
		GLOB.dead_mob_list += src
	else
		GLOB.alive_mob_list += src
```
This line made me cautious.

code/game/objects/machinery/nuclearbomb.dm
Rohesie: That's how the SD explosion works. Is that a problem?
The whole proc should be discussed, though, as it's a copypaste from `initiate_self_destruct()`
As it is, the shaking won't be seen. It will be covered by the cinematic screen.
At some point a rework may have skipped a `spawn()` or something for that to happen.

code/game/objects/machinery/computer/nuke_disk_generator.dm
Rohesie: Done.

code/game/objects/machinery/nuclearbomb.dm
DominikPanic: It should be safe to implement it then

code/game/objects/machinery/nuclearbomb.dm
DominikPanic: Not really a problem, but an elegant solution may be to define the proc to just return at the mob level, and to shake at the living level, this would save us both typechecks

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
DominikPanic: 1. Try to find observer and spawn them
2. If that works, return, if not, either notify observers with a popup or spawn an empty larva

That was the idea me and psy agreed on

code/game/objects/effects/countdown.dm
```diff
@@ -0,0 +1,76 @@
+/obj/effect/countdown
```
psykzz: I did not know this was a thing

Rohesie: It's a thing in /tg/, aye.
It's in the PR text:
`Did a few small improvements, such as showing ghosts the timer on the nuke, and making it start anchored to avoid xeno pushing memes.`

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
Rohesie: I'll need clarification on where you want this behavior added.
I can't add it to resin silos because they are just passive spawn points. Nor to `attempt_to_spawn_larva()` because this is merely the hive proc called by `attempt_to_join_as_larva()`.

code/game/objects/effects/countdown.dm
psykzz: Oh i understand you doing it, i just didn't know text on screen like this was this easy. Means we could have more context for certain things instead of just icons.

For example when people shout medic, to just output medic / help /etc

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
DominikPanic: I don't care where it's added but any xeno spawning methods should try to use that proc to preserve the behavior that oldest observers get to play first, and only using popups and other ways in case there are none, to still offer ghosts without prefs to play.

code/game/objects/effects/countdown.dm
Rohesie: It doesn't look great, but yea, it's a possibility BYOND offers.

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
Rohesie: Noted.

code/game/objects/machinery/nuclearbomb.dm
Rohesie: The name of that is a signal, really. I'll see later if there's a simple way of implementing it.

code/game/objects/machinery/computer/nuke_disk_generator.dm
psykzz: This printing time is good, previously it had 3 SECONDS or so before i changed it to some standard number.

code/game/objects/machinery/computer/nuke_disk_generator.dm
```diff
@@ -0,0 +1,170 @@
...
+	if(.)
+		return
+
+	if(href_list["generate"])
+		if(printing || current_timer)
+			to_chat(usr, "<span class='warning'>A program is already running.</span>")
+			return
+		if(reprintable)
+			printing = TRUE
+			addtimer(VARSET_CALLBACK(src, printing, FALSE), printing_time) // TODO Change to larger time
```
psykzz:
```suggestion
			addtimer(VARSET_CALLBACK(src, printing, FALSE), printing_time)
```

code/game/objects/machinery/computer/nuke_disk_generator.dm
```diff
@@ -0,0 +1,170 @@
...
+
+			usr.visible_message("[usr] started a program to regenerate a nuclear disk code.", "You started a program to generate a nuclear disk code.")
+			if(!do_after(usr, printing_time, TRUE, src, BUSY_ICON_GENERIC, null, null, CALLBACK(src, /datum.proc/process)))
+				return
+
+			print_disc()
+			return
+
+		printing = TRUE
+		addtimer(VARSET_CALLBACK(src, printing, FALSE), segment_time) // TODO Change to larger time
```
psykzz:
```suggestion
		addtimer(VARSET_CALLBACK(src, printing, FALSE), segment_time)
```

PR #2505, opened 2019-07-31
Small xeno takeover refactor

Rohesie:
* Made it use `afk_timer_id` instead of `away_time` to match the changes for humans.
* Added checks for aghosting not to generate timers and messages in xenos or humans.


code/modules/admin/holder.dm
```diff
@@ -543,6 +543,8 @@ GLOBAL_PROTECT(admin_verbs_spawn)
 /proc/afk_message(mob/living/carbon/human/H)
 	if(QDELETED(H))
 		return
+	if(isaghost(H) && GLOB.directory[H.key])
```
DominikPanic: Macro this?

PR #2512, opened 2019-07-31
Locks input keys based on focus_chat

psykzz:
Frees up a set of keys that are commonly associated with either side of focus chat.
With focus chat on, we no longer look at WASD for movement.
With focus chat off, we no longer use arrows for movement.


code/modules/keybindings/bindings_atom.dm
```diff
@@ -1,17 +1,21 @@
...
-			movement_dir |= user.next_move_dir_add
-		if(user.next_move_dir_sub)
-			movement_dir &= ~user.next_move_dir_sub
-		// Sanity checks in case you hold left and right and up to make sure you only go up
-		if((movement_dir & NORTH) && (movement_dir & SOUTH))
-			movement_dir &= ~(NORTH|SOUTH)
-		if((movement_dir & EAST) && (movement_dir & WEST))
-			movement_dir &= ~(EAST|WEST)
-		user.Move(get_step(src, movement_dir), movement_dir)
+	if(user.keys_held["Ctrl"])
```
Rohesie: This is something I'd like to change, by the byes.
The ability to turn around holding CTRL is a bit less interesting than the ability to use it for secondary actions, such as grabbing someone or firing from an attachment while moving.

psykzz:
This PR is having issues with CTRL movement in general right now its blocking the keybinding and not doing the facing bindings

code/modules/keybindings/bindings_atom.dm
psykzz: I agree, i just didn't change it because a lot of people reee'd when i broke it.

code/modules/keybindings/bindings_client.dm
```diff
@@ -37,7 +38,9 @@
 	set hidden = TRUE
 
 	keys_held -= _key
-	var/movement = SSinput.movement_keys[_key]
+
+	var/focus_chat = prefs.focus_chat
```
DominikPanic: This is making an unnecessary variable since you literally only use it once

code/modules/keybindings/bindings_atom.dm
DominikPanic: Just revert the indentation changes.

code/modules/keybindings/bindings_client.dm
```diff
@@ -5,12 +5,13 @@
 	set hidden = TRUE
 
 	keys_held[_key] = world.time
-	var/movement = SSinput.movement_keys[_key]
+	var/focus_chat = prefs.focus_chat
```
DominikPanic: Actually, you should profile this, if creating a new variable is faster than accessing it twice.

psykzz: image

Original vs revised.

psykzz:
Generally hated by the people, happy to close.

PR #2517, opened 2019-08-01
Moves variable definitions to their path definition files

Rohesie:
I keep promising myself I'll do that, but I'm usually doing something else that would require stopping.


code/modules/mob/mob_defines.dm
```diff
@@ -82,3 +82,6 @@
 	var/obj/item/storage/s_active //Carbon
 	var/obj/item/clothing/mask/wear_mask //Carbon
 	var/turf/listed_turf	//the current turf being examined in the stat panel
+
+	var/datum/focus //What receives our keyboard inputs. src by default
```
DominikPanic: Remove the empty line so it's nicely organized

Rohesie: Removed extra spaces and added a comment for organization, not to mix input with interaction.
Does this work?

PR #2520, opened 2019-08-02
Allows the AI to use the autodoc

DominikPanic:
Also had to do a slight interface update cleanup since such a tiny change made me find quite a few issues.

Removed updateDialog since it was redundant.
Mass replaced all occurences with updateUsrDialog
Neither of those can nor should be used in process due to the way interface interaction works, and hacking it together is not really feasible. Interfaces that want to update constantly should be using nanoUI instead.

### DIFF
--- a/code/game/objects/machinery/autodoc.dm
+++ b/code/game/objects/machinery/autodoc.dm
@@ -1044,9 +1044,11 @@
 	if(connected.occupant && ishuman(connected.occupant))
 		// manual surgery handling
 		var/datum/data/record/N = null
-		for(var/datum/data/record/R in GLOB.datacore.medical)
-			if (R.fields["name"] == connected.occupant.real_name)
+		for(var/i in GLOB.datacore.medical)
+			var/datum/data/record/R = i
+			if(R.fields["name"] == connected.occupant.real_name)
 				N = R
+		
 		if(isnull(N))
 			N = create_medical_record(connected.occupant)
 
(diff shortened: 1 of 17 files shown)


code/game/objects/machinery/autodoc.dm
```diff
@@ -1041,176 +1038,171 @@
...
-				if(!needed)
-					N.fields["autodoc_manual"] += create_autodoc_surgery(null,LIMB_SURGERY,ADSURGERY_INTERNAL,1)
-				updateUsrDialog()
-
-			if(href_list["broken"])
-				for(var/datum/limb/L in connected.occupant.limbs)
-					if(L)
-						if(L.limb_status & LIMB_BROKEN)
-							N.fields["autodoc_manual"] += create_autodoc_surgery(L,LIMB_SURGERY,ADSURGERY_BROKEN)
+	if(connected.occupant && ishuman(connected.occupant))
```
Rohesie:
```suggestion
	if(ishuman(connected.occupant))
```

code/game/objects/machinery/autodoc.dm
```diff
@@ -1041,176 +1038,171 @@
...
+		// manual surgery handling
+		var/datum/data/record/N = null
+		for(var/datum/data/record/R in GLOB.datacore.medical)
```
Rohesie: No need for typechecking.

code/game/objects/machinery/autodoc.dm
```diff
@@ -1041,176 +1038,171 @@
...
+			N.fields["autodoc_manual"] += create_autodoc_surgery(null,EXTERNAL_SURGERY,ADSURGERY_BLOOD)
+
+		if(href_list["organgerms"])
+			N.fields["autodoc_manual"] += create_autodoc_surgery(null,ORGAN_SURGERY,ADSURGERY_GERMS)
+
+		if(href_list["eyes"])
+			N.fields["autodoc_manual"] += create_autodoc_surgery(null,ORGAN_SURGERY,ADSURGERY_EYES,0,connected.occupant.internal_organs_by_name["eyes"])
+
+		if(href_list["organdamage"])
+			for(var/datum/limb/L in connected.occupant.limbs)
```
Rohesie: No need for typechecking.

code/game/objects/machinery/autodoc.dm
```diff
@@ -1041,176 +1038,171 @@
...
+				if(L)
```
Rohesie: Check unecessary inside the `for` loop.

code/game/objects/machinery/autodoc.dm
```diff
@@ -1041,176 +1038,171 @@
...
-				for(var/datum/limb/L in connected.occupant.limbs)
-					if(L)
-						if(L.limb_status & LIMB_NECROTIZED)
-							N.fields["autodoc_manual"] += create_autodoc_surgery(L,LIMB_SURGERY,ADSURGERY_NECRO)
+			if(!needed)
+				N.fields["autodoc_manual"] += create_autodoc_surgery(null,ORGAN_SURGERY,ADSURGERY_DAMAGE,1)
+
+		if(href_list["internal"])
+			for(var/datum/limb/L in connected.occupant.limbs)
+				if(L)
```
Rohesie: Check and typecheck unnecessary,

code/game/objects/machinery/autodoc.dm
DominikPanic: Not my code

code/game/objects/machinery/autodoc.dm
DominikPanic: Same, just reduced indentation. But I'll fix it regardless.

PR #2541, opened 2019-08-03
Minor Canterbury Improvements

Alterist:
Slight improvements to the pipnet and wiring, fixed extra hull turf in the cockpit. Added a crate with some attachments to the cargo bay

:cl:
add: Added new things
/:cl:


psykzz:
If you wanna do these changes would you instead apply them to my branch for the crash-gm.
There are already a number of changes and its for sure going to conflict.

Issue #2526, opened 2019-08-02
humans don't go in body bags


factoryman942:
observe this human:
![image](https://user-images.githubusercontent.com/20996272/62400896-cf4ace80-b578-11e9-945a-d42f1b09fb94.png)
It has been stripped of all equipment, yet it refuses to go in the bag.
Clickdragging it to the bag is also ineffective.

TMs:
#2523: 'Fixes give mob mind issues' by DominikPanic at commit ff1301f724 
#2524: 'Fixes xeno join admin logs' by DominikPanic at commit ec7dfd2af0 
#2502: 'Makes you deaf while unconscious' by Maciekkub at commit 3da9b9bfdb 
#2520: 'Allows the AI to use the autodoc' by DominikPanic at commit 98c50afac5 
#2521: 'Adds an easy spatial agent verb' by DominikPanic at commit 3d8f9bf5b8 
#2522: 'Clothing restriction changes' by DominikPanic at commit 7eb6b3a5ca 
#2514: 'Fixes 2 slight interface issues' by DominikPanic at commit 5f0ab24571 
#2507: 'Fixes prayer logging' by DominikPanic at commit ed15df8611 
#2508: 'Very minor turf code cleanup' by DominikPanic at commit d17719e99a 
#2509: 'Fixes some more shutter orientation on big red' by DominikPanic at commit 6d34155b6c 
#2510: 'Fixes Lamarr runtime' by DominikPanic at commit b6816b803c 
#2379: 'Add some dragover QOL' by psykzz at commit ef4c534f4b 
#2128: 'Full auto' by Rohesie at commit fd0e583099 
#2361: 'Fixes plasmacutter interactions' by DominikPanic at commit 9cb69d24b7 
#2506: 'Reworks the automute system add a message weight option' by DominikPanic at commit b0fb71d7f8 
#2495: 'Adds camera tracking to the overwatch console' by DominikPanic at commit 1740a08a8e 
#1780: 'Gamemode: Crash (name pending)' by psykzz at commit d0020b14a7

DominikPanic:
@Rohesie Can confirm this is happening

Rohesie:
Unable to reproduce it on the live server:
![1](https://user-images.githubusercontent.com/42041276/62403494-f1862180-b563-11e9-8e27-35c23a6886ca.png)
[...]

PR #2545, opened 2019-08-04
Fixes bodybag body handling

DominikPanic:
Fixes: #2526

### DIFF
--- a/code/game/objects/items/bodybag.dm
+++ b/code/game/objects/items/bodybag.dm
@@ -124,7 +124,7 @@
 	if(mob_to_stuff.stat != DEAD) //Only the dead for bodybags.
 		return FALSE
 	var/mob/living/carbon/human/human_to_stuff = mob_to_stuff
-	if((!check_tod(human_to_stuff) || issynth(human_to_stuff)) && human_to_stuff.is_revivable())
+	if(human_to_stuff.is_revivable() && (check_tod(human_to_stuff) || issynth(human_to_stuff)))
 		return FALSE //We don't want to store those that can be revived.
 	return TRUE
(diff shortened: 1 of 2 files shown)


code/game/objects/items/bodybag.dm
```diff
@@ -124,7 +124,7 @@
 	if(mob_to_stuff.stat != DEAD) //Only the dead for bodybags.
 		return FALSE
 	var/mob/living/carbon/human/human_to_stuff = mob_to_stuff
-	if((!check_tod(human_to_stuff) || issynth(human_to_stuff)) && human_to_stuff.is_revivable())
+	if((check_tod(human_to_stuff) && human_to_stuff.is_revivable()) || issynth(human_to_stuff))
```
Rohesie: You could revive a headless synth with this change.

DominikPanic: This does not change revival code, this merely fixes closing bodybags in some situations

Rohesie: It does change behavior, is what I mean.
Old code previews two behaviors:
1) If it's synth, check `is_revivable()` (because a headless synth can't be revived)
2) Whether synth or not, check if `undefibbable` and enough time has passed, which is the `check_tod()` macro.

New code has two different behaviors.
1) If it's synth, check always succeeds. This allows for headless synths to be revived.
2) If other cases, it needs both `is_revivable()` and `s_revivable()` to pass.

This is because `&&` is evaluated before `||`

DominikPanic: The old behavior is broken, though.

PR #2550, opened 2019-08-04
Xenos can now see better in the dark with nightvision disabled

DominikPanic:
More useful to tell which place is dark or lit without compromising their ability too much.

:cl: LaKiller8
tweak: Xenos without night vision should now be able to tell if they are in a lit area better without being completely blinded.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/login.dm
+++ b/code/modules/mob/living/carbon/xenomorph/login.dm
@@ -4,9 +4,8 @@
 		deltimer(afk_timer_id)
 		afk_timer_id = null
 
-	if(see_in_dark == XENO_NIGHTVISION_ENABLED)
-		lighting_alpha = LIGHTING_PLANE_ALPHA_INVISIBLE
-		see_in_dark = XENO_NIGHTVISION_ENABLED
+	if(lighting_alpha == LIGHTING_PLANE_ALPHA_NV_TRAIT)
+		see_in_dark = 8
 		ENABLE_BITFIELD(sight, SEE_MOBS)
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/carbon/xenomorph/login.dm
```diff
@@ -4,9 +4,8 @@
 		deltimer(afk_timer_id)
 		afk_timer_id = null
 
-	if(see_in_dark == XENO_NIGHTVISION_ENABLED)
-		lighting_alpha = LIGHTING_PLANE_ALPHA_INVISIBLE
-		see_in_dark = XENO_NIGHTVISION_ENABLED
+	if(lighting_alpha == LIGHTING_PLANE_ALPHA_NV_TRAIT)
+		see_in_dark = 8
```
Rohesie:
```suggestion
		see_in_dark = XENO_NIGHTVISION_ENABLED
```

Rohesie: Ah, I see you intentionally removed it.
Shouldn't these be defines, in general?

DominikPanic: Because it doesn't change. This variable no longer controls whether their vision is enabled or disabled, it stays at 8

Rohesie: So replaced with new defines, like the planes and layers?

DominikPanic: No, it does not need new defines, that would be confusing.

PR #2555, opened 2019-08-04
Fixes chat early client DC issue

DominikPanic:
Client may get fucked during connecting which is a confirmed BYOND bug we can't really solve, which means we'll have to add checks to a few extra places relating to early client init. This fixes like 3 related runtimes


Rohesie:
I regret merging it so fast now. Think you could add a comment in the code regarding this?

PR #2556, opened 2019-08-04
Fixes Ice shutter orientation issues

DominikPanic:


Rohesie:
Hope this doesn't conflict with Crash GM...

DominikPanic:
Both were testmerged together

PR #2568, opened 2019-08-05
Removes an ancient config file

DominikPanic:
How did I not get to this before? It's not used, inaccurate, and encourages varediting.


Rohesie:
I grudgingly accept removing this guide to `THE NEW (june 2008) ACCESS LEVEL SYSTEM`, albeit I'm worried about having no guidance now for it.

DominikPanic:
Still better than a guidance that suggests varediting :p

PR #2576, opened 2019-08-10
Ports the looping sounds datum, and cleans up sound.dm

Ghommie:
## About The Pull Request
Adding the looping sound datum. The name is self explainatory, it uses timers to loop sets of sounds, plus a start and an ending set, to be played to specific targets or the surroundings depending on the settings, If I recall correctly.
I modified sound.dm a little to fit it in, then carried on a bit further, cleaning it from single line statements and updating it where viable.
Possibly boosted performance of playsound() by having it create a sound used on each playsound_local(), instead of the latter do so everytime. I'm not big into performance myself.

## Why It's Good For The Game
Making ports and original features that should be using looping sounds easier, code/performance improvement/update.


DominikPanic:
Everyone is deaf.

Rohesie:
Also travis.

Ghommie:
Ok, this is now fixed and tested locally, smoothly with no runtime. Apologises for the inconvenience before.

Rohesie:
Solve the merge conflict and we can test it.

Rohesie:
Travis check.
Please let us know when you test it locally, so we are able to test it on the server.

Ghommie:
Travis check was successful, what happened is the Map Diff Renders bot erroring its run, despite no modification to maps being applied. So all is fine now as I stated before.

Rohesie:
Little conflict now, though.

code/game/sound.dm
```diff
@@ -10,37 +7,43 @@
 //falloff: how the sound's volume decreases with distance, low is fast decrease and high is slow decrease.
 //A good representation is: 'byond applies a volume reduction to the sound every X tiles', where X is falloff.
 
-/proc/playsound(atom/source, soundin, vol, vary, sound_range, falloff, is_global)
-
-	if(!sound_range) sound_range = round(0.5*vol) //if no specific range, the max range is equal to half the volume.
-
-	soundin = get_sfx(soundin) // same sound for everyone
-
+/proc/playsound(atom/source, soundin, vol, vary, sound_range, falloff, is_global, frequency, channel = 0, pressure_affected = TRUE)
```
DominikPanic: Honestly, just get rid of all the perssure code here.

code/game/sound.dm
```diff
@@ -10,37 +7,43 @@
...
 	for(var/i in GLOB.player_list)
-		M = i
-		if(!istype(M) || !M.client) continue
-		if(get_dist(M, turf_source) <= sound_range)
-			T = get_turf(M)
-			if(T && T.z == turf_source.z) M.playsound_local(turf_source, soundin, vol, vary, frequency, falloff, is_global)
-
-/mob/proc/playsound_local(turf/turf_source, soundin, vol, vary, frequency, falloff, is_global)
+		var/mob/M = i
+		if(!istype(M) || !M.client)
```
DominikPanic: This will always be the correct type

code/game/sound.dm
```diff
@@ -10,37 +7,43 @@
...
+			continue
+		var/turf/T = get_turf(M)
+		if(T && T.z == turf_source.z && get_dist(M, turf_source) <= sound_range)
```
DominikPanic: Early return

code/game/sound.dm
```diff
@@ -10,37 +7,43 @@
...
 
-/proc/playsound(atom/source, soundin, vol, vary, sound_range, falloff, is_global)
-
-	if(!sound_range) sound_range = round(0.5*vol) //if no specific range, the max range is equal to half the volume.
-
-	soundin = get_sfx(soundin) // same sound for everyone
-
+/proc/playsound(atom/source, soundin, vol, vary, sound_range, falloff, is_global, frequency, channel = 0, pressure_affected = TRUE)
 	if(isarea(source))
 		CRASH("[source] is an area and is trying to make the sound: [soundin]")
```
DominikPanic: This is no longer necessary

code/game/sound.dm
```diff
@@ -98,28 +85,40 @@
...
 
-	if(!is_global) S.environment = 2
+	if(!is_global)
+		S.environment = 2
+
 	SEND_SOUND(src, S)
 
 
-/mob/living/playsound_local(turf/turf_source, soundin, vol, vary, frequency, falloff, is_global)
+/mob/living/playsound_local(turf/turf_source, soundin, vol, vary, frequency, falloff, is_global, channel = 0, pressure_affected = TRUE, sound/S)
```
DominikPanic: Since you've removed the pressure code, `pressure_affected` should be removed too. Also, why two variables for the sound? soundin vs S

Ghommie: `playsound()` generates the `var/sound/S` which is sent to each recipient's `playsound_local()` instead of the latter generating it for every recipient.
Though there are situations (such as hallucinations) where `playsound_local()` is used standalone, and thus soundin is still necessary for them to generate the sound datum.
I assume `istype()` would be detrimental to its performance, and thus it's better to leave both vars.

Btw thanks for warning me about the pressure_affected var leftover.

PR #2583, opened 2019-08-10
Fixes human and xeno numbers on admin distress message

psykzz:
Also now checks that xenos are more than 80% globally or more than 50% shipside instead of just checking the planet, since we added a transit level.

### DIFF
--- a/code/game/objects/machinery/computer/communications.dm
+++ b/code/game/objects/machinery/computer/communications.dm
@@ -207,7 +207,6 @@
 				var/AllMarines[] = All[1]
 				var/AllXenos[] = All[2]
 				if((AllXenos < round(AllMarines * 0.8)) && (ShipXenos < round(ShipMarines * 0.5))) //If there's less humans (weighted) than xenos, humans get home-turf advantage
-					log_game("[key_name(usr)] has attemped to call a distress beacon, but it was denied due to lack of threat.")
 					to_chat(usr, "<span class='warning'>The sensors aren't picking up enough of a threat to warrant a distress beacon.</span>")
 					return FALSE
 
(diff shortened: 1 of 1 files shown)


code/game/objects/machinery/computer/communications.dm
```diff
@@ -203,10 +203,10 @@
...
 				var/ShipXenos[] = Ship[2]
-				var/Planet[] = SSticker.mode.count_humans_and_xenos(SSmapping.levels_by_trait(ZTRAIT_GROUND))
-				var/PlanetMarines[] = Planet[1]
-				var/PlanetXenos[] = Planet[2]
-				if((PlanetXenos < round(PlanetMarines * 0.8)) && (ShipXenos < round(ShipMarines * 0.5))) //If there's less humans (weighted) than xenos, humans get home-turf advantage
+				var/All[] = SSticker.mode.count_humans_and_xenos()
+				var/AllMarines[] = All[1]
+				var/AllXenos[] = All[2]
+				if((AllXenos < round(AllMarines * 0.8)) && (ShipXenos < round(ShipMarines * 0.5))) //If there's less humans (weighted) than xenos, humans get home-turf advantage
 					log_game("[key_name(usr)] has attemped to call a distress beacon, but it was denied due to lack of threat.")
```
DominikPanic: This isn't really necessary anymore

psykzz: Is it because its covered in the following line?

DominikPanic: Yes, this was used as a debug measure back in the CM code days and used to be much more spammy. There's no need to log this now.

PR #2610, opened 2019-08-11
 Fixes a null check

psykzz:
`Runtime in defibrillator.dm, line 75: Cannot read null.current`

### DIFF
--- a/code/game/objects/items/defibrillator.dm
+++ b/code/game/objects/items/defibrillator.dm
@@ -72,7 +72,7 @@
 		return FALSE
 	for(var/g in GLOB.observer_list)
 		var/mob/dead/observer/ghost = g
-		if(ghost.mind.current != src)
+		if(ghost?.mind.current != src)
 			continue
 		if(ghost.client && ghost.can_reenter_corpse)
 			return ghost


code/game/objects/items/defibrillator.dm
```diff
@@ -72,7 +72,7 @@
 		return FALSE
 	for(var/g in GLOB.observer_list)
 		var/mob/dead/observer/ghost = g
-		if(ghost.mind.current != src)
+		if(ghost?.mind.current != src)
```
DominikPanic: Ghosts should always have a mind

Rohesie:
```suggestion
		if(ghost.mind.current != src)
```

DominikPanic: Well considering this is the entire PR, this suggestion would close it automatically

Rohesie: Oh wait...
Hold on, why is there a `null` element in `observer_list`?

DominikPanic: It's an observer with a null mind, not a null element in observer_list

psykzz:
I'll close this to avoid confusion @Rohesie @DominikPanic

PR #2621, opened 2019-08-12
Temporarily removes the tank

DominikPanic:
TC is disabled anyway and right now Synth and FC/Cap can use it


Rohesie:
There's tanks in the debugdalus and ice colony as well.

DominikPanic:
Copy, will get those

PR #2623, opened 2019-08-13
Adds xeno panel for the admemes

DominikPanic:
Noice

### DIFF
--- a/code/modules/admin/fun_verbs.dm
+++ b/code/modules/admin/fun_verbs.dm
@@ -858,10 +858,9 @@
 	message_admins("[ADMIN_TPMONTY(usr)] has offered [ADMIN_TPMONTY(L)].")
 
 
-/datum/admins/proc/change_hivenumber(mob/living/carbon/xenomorph/X in GLOB.xeno_mob_list)
+/datum/admins/proc/xeno_panel(mob/living/carbon/xenomorph/X in GLOB.xeno_mob_list)
 	set category = "Fun"
-	set name = "Change Hivenumber"
-	set desc = "Set the hivenumber of a xenomorph."
+	set name = "Xeno Panel"
 
 	if(!check_rights(R_FUN))
 		return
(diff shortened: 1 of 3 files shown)


code/modules/admin/fun_verbs.dm
```diff
@@ -858,54 +858,26 @@
...
-	if(!istype(X) || X.hivenumber != hivenumber_status)
-		to_chat(usr, "<span class='warning'>Target is no longer valid.</span>")
-		return
-		return
+	dat += "Hive: [X.hive.hivenumber] <a href='?src=[REF(usr.client.holder)];[HrefToken()];xeno=hive;mob=[REF(X)]'>Edit</a><br>"
+	dat += "Nicknumber: [X.nicknumber] <a href='?src=[REF(usr.client.holder)];[HrefToken()];xeno=nicknumber;mob=[REF(X)]'>Edit</a><br>"
+	dat += "Upgrade Tier: [X.xeno_caste.upgrade_name] <a href='?src=[REF(usr.client.holder)];[HrefToken()];xeno=upgrade;mob=[REF(X)]'>Edit</a><br>"
 
-	X.transfer_to_hive(newhivenumber)
+	var/datum/browser/browser = new(usr, "xeno_panel_[key_name(X)]", "<div align='center'>Xeno Panel [key_name(X)]</div>")
```
psykzz:
```suggestion
	var/datum/browser/browser = new(usr, "xeno_panel", "<div align='center'>Xeno Panel [key_name(X)]</div>")
```

This is causing infinite new panels and would be better to just force one at a time to avoid confusion.

psykzz: Not blocking just QOL for the QOL

DominikPanic: That means you won't be able to have panels of separate players open at the same time

psykzz: If you can make sure you close the old panel before opening the new panel, im also happy.

DominikPanic: Again, that would remove QoL. This allows you to have as many xeno panels open as you want at the same time. if you close it, it gets closed properly.

psykzz: I mean for the same player, if you decide to level one person tier 1 to tier 2 they have a new key_name which causes a new panel, but the old panel is for a mob that no longer exists.

PR #2625, opened 2019-08-13
Fixes a few small oof with AI names

DominikPanic:
I'm stoopid

### DIFF
--- a/code/datums/jobs/job/job.dm
+++ b/code/datums/jobs/job/job.dm
@@ -110,7 +110,7 @@ GLOBAL_PROTECT(exp_specialmap)
 
 	if(!ishuman(L))
 		return FALSE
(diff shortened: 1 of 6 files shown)


code/datums/jobs/job/job.dm
```diff
@@ -94,10 +94,23 @@ GLOBAL_PROTECT(exp_specialmap)
...
+
+	L.job = title
+	L.faction = faction
+
+	if(skills_type)
+		var/datum/skills/S = new skills_type
+		L.mind.cm_skills = S
+
+	if(!ishuman(L))
+		return FALSE
```
psykzz:
```suggestion
		return TRUE
```
TRUE makes more sense here because it was handled

PR #2626, opened 2019-08-13
Boiler Bombard Time "rebalance"

DalekThePerson:
## About The Pull Request
Decreases the boiler's time to dig and glob launch.

Digging changed from 4 seconds to 3 seconds.
Bombard Launch changed from 3 seconds to 2 seconds.
The cooldown between bombards was changed from 20 to 30.

Total time it takes to setup and bombard is ~6 seconds excluding the time it takes to click the longsight and find where you want to fire.

These changes will most likely change if too much or too little.

## Why It's Good For The Game
Makes the boiler not take ~10 seconds to launch the globs to give boilers an easier time when trying to launch under pressure. 

## Changelog
:cl:
balance: Decreased the time it takes to dig into the ground and acid blob launch by 1 seconds each. And increases the cooldown between bombards by 10 seconds.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
@@ -44,7 +44,7 @@
 	// *** Boiler Abilities *** //
 	bomb_strength = 1 //Multiplier to the effectiveness of the boiler glob. Improves by 0.5 per upgrade
 	acid_delay = 9 SECONDS //9 seconds delay on acid. Reduced by -1 per upgrade down to 5 seconds
-	bomb_delay = 20 SECONDS //20 seconds per glob at Young, -2.5 per upgrade down to 10 seconds
+	bomb_delay = 30 SECONDS //20 seconds per glob at Young, -2.5 per upgrade down to 10 seconds
 
 /datum/xeno_caste/boiler/young
 	upgrade_name = "Young"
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
```diff
@@ -155,4 +155,4 @@
 	// *** Boiler Abilities *** //
 	bomb_strength = 2.5
 	acid_delay = 9 SECONDS //9 seconds delay on acid. Reduced by -1 per upgrade down to 5 seconds
-	bomb_delay = 20 SECONDS //20 seconds per glob at Young, -2.5 per upgrade down to 10 seconds
+	bomb_delay = 30 SECONDS //20 seconds per glob at Young, -2.5 per upgrade down to 10 seconds
```
Rohesie: The comments were outdated before, but now they are even more.

Rohesie:
This is a serious boiler nerf. Sure you want to increase the delay between shots?

Maciekkub:
I see the boiler is more agile but shoots slower and i believe it might not be a bad change

Rohesie:
30 seconds is a pretty long time. Test this and see how the players react to it.

Maciekkub:
I got to play some boiler recently, and being honest, it's a very pleasurable caste to play as, the decreased setup time and fire time reward the increased cooldown between shots, the only thing that strikes me a bit uneccessary is longish time on turning on long sight. Nonetheless, the boiler caste is quite strong and if provided with other aliens to support it is marvelous, similarily to Shrike, it's not exactly able to fight alone, but with the help of its teammates can wreck havoc and make the marines stuck at one place for long periods of time. The spray acid ability on it is also valuable, as it provides with an escape means in case of being chased, or if a quick stun is needed, as i believe it travels very fast

PR #2642, opened 2019-08-15
Chat subsystem optimization plus a fallback solution

DominikPanic:
Will require testing


Rohesie:
Micro-optimization: if `C.chatOutput.broken` was a positive value, such as `C.chatOutput.working`, you could simply do `if(C.chatOutput?.working)` instead of `if(!C.chatOutput || C.chatOutput.broken)` each time.

PR #2643, opened 2019-08-15
Pseudo-pixel projectiles

Rohesie:
## About The Pull Request

* Projectiles now process through a subsystem.
* Their interaction is still tile-based (they will scan any tile they enter for possible collision with their contents), but movement is calculated through a pixel-based line, which makes it a bit more accurate.
* Their movement is animated for speeds up to 4. For that and over the animation is a bit too fast, so they look the same as current ones, but the flicks seen are on a correct pixel offset instead of in the middle of the tile. It shouldn't be too expensive to increase this, but over 6 it's just fails to bring any benefit.
* ~~Because these are intended to be kept cheap the animation is called only on batch moves, which makes them look janky at speeds under 2, unlike /tg/ projectiles. They are _much_ cheaper, though. And they still look much less janky than old CM projectiles at the same speed.~~ They look pretty good.
* Chance to hit living mobs was oddly divided into `scan_a_turf()` and `get_projectile_hit_chance()`. Cleared it up and unified it into `projectile_hit()`, leaving the effects of the hit in case the former succeeds for `do_projectile_hit()` (which was meshed up in `scan_a_turf()` before).
*  Did small tweaks on accuracy/hit chance, and a small rework on the sniper laser targeting, incidentally.
* `fire_at()` now allows for shooting towards the same tile, so there's no need to be super snowflakey on pointblanking.
* There is now support for super slow projectiles. Under 0.2 movement may get a bit stuttery, but it works. There are optimizations to be made if that speeds under 1 start being used.
* Removes the shell speed configs. The plan is to move in that direction and remove the obscurity those configs create. They have are no longer really useful.
* muzzle_flash is aligned to use the same angle as the fired shot, plus using visual contents for better performance (at some memory cost, each gun having a movable atom associated).

## Changelog
:cl:
add: Reworked projectiles to use pixel movement (but not collisions, those are still tile-based), tracking the clicked pixel (unless a mob is clicked, in which case it will aim for the center of the target's tile for maximum accuracy, or unless scatter happens) in a straight line. For most bullets (rifles, SMGs, pistols, revolvers) the difference will be subtle at best, given how fast they are. It is more noticeable in tasers. For lasguns and faster projectiles the moving animation is outright disabled.
fix: Fixed a speed inconsistency of old projectiles, which would make bullet speed lower if you clicked closer to your own sprite.
balance: Tweaked accuracy to make the calculations a bit more sane. This may impact on it as reproducing the exact same behavior is hard, but it can be more easily adjusted now.
balance: Once a sniper acquires a target through its laser market it will no longer need to click on it to aim it. Any shots while active will be auto-targeted to it.
balance: Projectile speeds reduced all round the board. Xeno projectiles are slow, so you can see them traveling. Non-slug shotgun shells are slow and more visible. Slugs and regular bullets are fast, and hard to see. Sniper bullets and lasers are very fast, and even harder to see.
balance: Scatter is now angle-based, so it doesn't matter if the target (where you click) is near or far. Maximum scatter is 45 degrees in either direction. Extra projectiles (additional buckshot and flechettes) have a smoother, less random scatter, firing in a cone.
/:cl:

Short range pixel shot, using values taken from testing, position adjusted:
firefox_jXKIytPgPU
```
X1: 5857
Y1: 1569
X2: 5828.2
Y2: 1554.92
X3: 5799.4
Y3: 1540.84
X4: 5770.6
Y4: 1526.76
```
Shot is from the upper-right point towards the bottom-left one.
Animation is called three times, from P1 to P2, from P2 to P3 and from P3 to P4.
In tile coordinates, the shot is fired from (4, 3), scans (3,3), (3, 2), (2, 2) and (2, 1) before finally colliding at (1, 1). Namely all the turfs it crosses.
If the shot is perfectly diagonal it will choose an appropriate cardinal tile to cross at random, to preserve behavior of no fully diagonal move allowed.

~~Despite the math seems sound, slow-moving projectiles still look janky. I assume it's an issue on the times pixel offset and loc are updated visually.
This could be easily fixed by going the /tg/ way and updating them more often, in smaller time frames, but the idea here is to keep it cheap.
Their projectiles have a speed of 1.25 by default (lasers, bullets, energy projectiles) in our measurement (tiles traveled per tick/decisecond), while our rifle bullets have speed 5, lasgun lasers 6, and tasers 2. If we don't count attachment modifiers.
Because of that, and because shooting is much less common than for us, their system is more focused towards smoothness and ours towards performance.
If /tg/ projectiles were fast as ours their cost would be prohibitive. If old CM projectiles were as slow as /tg/'s they'd look terrible. This allows us to slowdown projectiles a bit if needed, but not to the point of matching /tg/'s. Up to 3 I'd say should be pretty smooth, possibly less.~~

Actually this seems smooth enough at speed 1:
StZBclgRLv

### DIFF
--- a/code/game/objects/machinery/sentries.dm
+++ b/code/game/objects/machinery/sentries.dm
@@ -838,12 +838,19 @@
 	return
 
 /obj/machinery/marine_turret/proc/load_into_chamber()
-	if(in_chamber) return 1 //Already set!
-	if(!CHECK_BITFIELD(turret_flags, TURRET_ON) || !cell || rounds == 0 || machine_stat == 1) return 0
+	if(in_chamber)
+		return TRUE //Already set!
+	if(!CHECK_BITFIELD(turret_flags, TURRET_ON) || !cell || rounds == 0 || machine_stat & (NOPOWER|BROKEN))
+		return FALSE
 
-	in_chamber = new /obj/item/projectile(loc) //New bullet!
+	create_bullet()
+	return TRUE
+
+
+/obj/machinery/marine_turret/proc/create_bullet()
+	in_chamber = new /obj/item/projectile(src) //New bullet!
 	in_chamber.generate_bullet(ammo)
-	return 1
+
 
 /obj/machinery/marine_turret/proc/process_shot()
 	set waitfor = 0
@@ -922,7 +929,7 @@
 				in_chamber.ammo.penetration = round(in_chamber.ammo.penetration * 1.5, 0.01)
 
 			//Setup projectile
-			in_chamber.original = target
+			in_chamber.original_target = target
 			in_chamber.setDir(dir)
 			in_chamber.def_zone = pick("chest", "chest", "chest", "head")
 
--- a/code/modules/projectiles/updated_projectiles/gun_system.dm
+++ b/code/modules/projectiles/updated_projectiles/gun_system.dm
@@ -613,13 +613,10 @@ and you're good to go.
 	if(gun_on_cooldown(user))
 		return
 
-	var/turf/curloc = get_turf(user) //In case the target or we are expired.
 	var/turf/targloc = get_turf(target)
-	if (!targloc || !curloc)
+	if(!targloc)
+		CRASH("Fire called on [QDELETED(target) ? "qdeleted" : "invalid loc"] target ([target])")
 		return //Something has gone wrong...
-	var/atom/original_target = target //This is for burst mode, in case the target changes per scatter chance in between fired bullets.
-
-	
 
 	/*
 	This is where burst is established for the proceeding section. Which just means the proc loops around that many times.
@@ -647,6 +644,11 @@ and you're good to go.
 			click_empty(user)
 			break
 
+		if(QDELETED(target)) //This can happen on burstfire, as it's a sleeping proc.
+			if(QDELETED(targloc))
+				break
+			target = targloc //If the original targets gets destroyed, fire at its location.
+
 		var/recoil_comp = 0 //used by bipod and akimbo firing
 
 		//checking for a gun in other hand to fire akimbo
@@ -670,8 +672,7 @@ and you're good to go.
 			if(mouse_control["icon-y"])
 				projectile_to_fire.p_y = text2num(mouse_control["icon-y"])
 
-		target = original_target ? original_target : targloc
-		target = simulate_scatter(projectile_to_fire, target, targloc, user)
+		var/target_scatter = simulate_scatter(projectile_to_fire, target, targloc, user)
 
 		//Finally, make with the pew pew!
 		if(!projectile_to_fire || !istype(projectile_to_fire,/obj))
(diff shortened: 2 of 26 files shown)


Rohesie:
I prioritized performance over pretty-looking code. This code is pretty hot for our game mode.

code/game/objects/machinery/sentries.dm
```diff
@@ -838,12 +838,19 @@
 	return
 
 /obj/machinery/marine_turret/proc/load_into_chamber()
-	if(in_chamber) return 1 //Already set!
-	if(!CHECK_BITFIELD(turret_flags, TURRET_ON) || !cell || rounds == 0 || machine_stat == 1) return 0
+	if(in_chamber)
+		return TRUE //Already set!
+	if(!CHECK_BITFIELD(turret_flags, TURRET_ON) || !cell || rounds == 0 || machine_stat & (NOPOWER|BROKEN))
```
DominikPanic:
```suggestion
	if(!CHECK_BITFIELD(turret_flags, TURRET_ON) || !cell || !rounds || machine_stat & (NOPOWER|BROKEN))
```

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -16,7 +16,7 @@
 	flags_atom = CONDUCT
 	flags_item = TWOHANDED
 
-	var/muzzle_flash 	= "muzzle_flash"
+	var/atom/movable/vis_obj/effect/muzzle_flash/muzzle_flash = new
```
DominikPanic: Initialize?

code/modules/projectiles/updated_projectiles/gun_system.dm
```diff
@@ -1082,6 +1046,110 @@ and you're good to go.
...
+	if(!QDELETED(flash_loc))
+		flash_loc.vis_contents -= muzzle_flash
+	muzzle_flash.applied = FALSE
+
+/*
+	var/mutable_appearance/flash_img = new(muzzle_flash)
+	flash_img
+	var/matrix/rotate = matrix() //Change the flash angle.
+	rotate.Translate(0, 5)
+
```
DominikPanic: Commented out code

code/modules/projectiles/updated_projectiles/projectile.dm
```diff
@@ -315,142 +457,172 @@
...
-
+		. -= (proj.ammo.flags_ammo_behavior & AMMO_SNIPER) ? (proj.distance_travelled * 3) : (proj.distance_travelled * 5) //Snipers have a smaller falloff constant due to longer max range
 
 	#if DEBUG_HIT_CHANCE
 	to_chat(world, "<span class='debuginfo'>Final accuracy is <b>[.]</b></span>")
 	#endif
 
 	. = max(5, .) //default hit chance is at least 5%.
-	if(lying && stat) . += 15 //Bonus hit against unconscious people.
+	if(lying && stat == UNCONSCIOUS)
```
DominikPanic:
```suggestion
	if(lying && stat != CONSCIOUS)
```

code/game/objects/machinery/sentries.dm
Rohesie: I prefer this format because `!var` implies the variable is a boolean.

PR #2644, opened 2019-08-16
New Kitchen and Toilet sink sprites from TG

Alterist:
Currently unused, might use them in the summer ball map


code/game/objects/structures/watercloset.dm
```diff
@@ -474,8 +474,11 @@
 
 /obj/structure/sink/kitchen
 	name = "kitchen sink"
-	icon_state = "sink_alt"
+	icon_state = "sink2"
 
+/obj/structure/sink/bathroom
+	name = "kitchen sink"
```
psykzz:
```suggestion
	name = "bathroom sink"
```

If its a bathroom item, it wouldn't make sense being called a "kitchen sink"

DominikPanic:
You shouldn't add unused sprites, could you make a new item for them?

PR #2646, opened 2019-08-16
AI now automatically gets unachored on crash

DominikPanic:
:cl: LaKiller8
tweak: The AI can no longer toggle it's floor bolts, dropship crash disables them automatically.
/:cl:


Rohesie:
Elegant solution.
Alternatively, on hijack.

PR #2647, opened 2019-08-16
Removes focus chat

DominikPanic:
Almost every single round there's at least 2 ahelps about this mode. It's just not feasible to support this code that so few players use for little benefit. It's doing more harm than good at the moment

:cl: LaKiller8
del: Focus chat has been removed, and much confusion has been prevented.
/:cl:

### DIFF
--- a/code/__DEFINES/colors.dm
+++ b/code/__DEFINES/colors.dm
@@ -1,6 +1,3 @@
-#define COLOR_INPUT_DISABLED	"#F0F0F0"
-#define COLOR_INPUT_ENABLED		"#D3B5B5"
-
 #define COLOR_WHITE            "#EEEEEE"
 #define COLOR_GRAY             "#808080"
 #define COLOR_BLACK            "#000000"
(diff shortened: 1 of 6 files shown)


Rohesie:
What about it being disabled by default instead?

DominikPanic:
It is disabled. Doesn't change that 100 people have it enabled since they played here a long time ago and that was the default. It is not maintainable and not working as intended and impossible to fix like Franc said. There is little value in keeping this feature compared to just how many issues it brings.

Rohesie:
@francinum ?

francinum:
I did not say it is impossible to fix. I admitted that it was a difficult issue to maintain. If there is an issue with the old defaults being stuck in the system, burn a savefile version. I'm once again annoyed to wake up and check in on the server to this issue. *AGAIN.*

DominikPanic:
That was not what you said over voice Franc (or perhaps I misunderstood), and if so then I apologize. This PR wouldn't have been merged without your input nor was anyone trying to things behind your back.

francinum:
A solution to the problem exists. That's certain. TG's system does not cause the same issues, and I understand, yes, their keybindings are static and not dynamic. I'd have gladly fixed it or at least bodged it myself if I actually understood all this `.winset` engine garbage.

Issue #2659, opened 2019-08-16
You can access lz1 on big red when shutters are up 


Pariah919:
## Testmerges:

#2648: 'Makes changelog less laggy' by DominikPanic at commit 4462d3e94c
#2647: 'Removes focus chat' by DominikPanic at commit 288dd8ae44
#2477: 'The Rise of the ST' by Maciekkub at commit 406caa21db
#2635: 'Adds a boot knife to basic marine sets.' by grinkgo at commit 92e7307464
#2570: 'Object attack damage rework' by DominikPanic at commit 14b672eaab
#2645: 'Makes the pizza ERT admin only' by DominikPanic at commit 4e8f3e3978
#2364: 'Support emoji and emoji picker.' by DeadLineSS13 at commit 6e9a5c6a45
#2553: 'Insight into storage runtime' by DominikPanic at commit efda667a8a
#2599: 'Fixes a runtime when clicking with null turf in the clickcatcher' by DominikPanic at commit cb635f75c8
#2626: 'Boiler Bombard Time "rebalance"' by DalekThePerson at commit 7a3380ad0a
#2642: 'Chat subsystem optimization plus a fallback solution' by DominikPanic at commit 924df531e7
#2128: 'Full auto' by Rohesie at commit 462996e5da
#2646: 'AI fix' by DominikPanic at commit 6713e868ca
#2649: 'Fixes a few autodoc runtimes' by DominikPanic at commit 7304e1f143
step 1
be beno and go to a vent connected to lz1
step 2- Get inside and leave the pipe 
step 3- Laugh

PR #2666, opened 2019-08-16
Removes big red pipe cheesing

DominikPanic:
Fixes: #2659


Rohesie:
Do the pipes end in the entrance? Shouldn't there be a vent in there somewhere?

DominikPanic:
> 
> 
> Do the pipes end in the entrance? Shouldn't there be a vent in there somewhere?

Diffbot doesn't render pipes under floors since it can't see them, but yes it ends properly

Issue #2667, opened 2019-08-17
Cleaning yourself with soap makes you attack yourself


Maciekkub:
## Reproduction:
Take soap
click yourself
you attack and clean yourself at the same time
![Soapddead](https://user-images.githubusercontent.com/30427887/63204040-377ad380-c094-11e9-9332-225462663a47.png)

PR #2669, opened 2019-08-17
Fixes soap attacking

DominikPanic:
Fixes: #2667


Rohesie:
No, check the bug report. It would cause an attack message when cleaning oneself.

Rohesie:
Though an alternative behavior, perhaps more desirable, would be for harm intent to attack only and the other intents to clean, no?

DominikPanic:
You can't hurt anyone with a soap anyway and more intent fuckery will make reworks in the future harder once we do decide to clean up this mess

PR #2673, opened 2019-08-17
Fixes main overwatch console

DominikPanic:


psykzz:
Could you not put this in the init proc and just remove the need for it to be called in multiple places?

PR #2694, opened 2019-08-17
Shrike can now track a queen

psykzz:
Fixes an issue where the shrike would not be able to track a queen because its intelligent.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xenomorph.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xenomorph.dm
@@ -264,7 +264,7 @@
 		return
 
 	var/obj/screen/LL_dir = hud_used.locate_leader
-	if(xeno_caste.caste_flags & CASTE_IS_INTELLIGENT)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -264,7 +264,7 @@
 		return
 
 	var/obj/screen/LL_dir = hud_used.locate_leader
-	if(xeno_caste.caste_flags & CASTE_IS_INTELLIGENT)
+	if(hive.living_xeno_ruler == src || src == X) // No need to track ourselves, especially if we are the hive leader.
```
Rohesie:
```suggestion
	if(src == X) // No need to track ourselves.
```
Wouldn't this behave the same?

psykzz: Just future proof, this assume someone else could be a leader that isn't hive ruler.

PR #2721, opened 2019-08-18
Fixes misleading graceful shutdown message

DominikPanic:


goon/browserassets/js/browserOutput.js
```diff
@@ -506,6 +506,9 @@ function ehjaxCallback(data) {
 	} else if (data == 'roundrestart') {
 		opts.restarting = true;
 		internalOutput('<div class="connectionClosed internal restarting">The connection has been closed because the server is restarting. Please wait while you automatically reconnect.</div>', 'internal');
+	} else if (data == 'shutdown') {
+		opts.restarting = true;
+		internalOutput('<div class="connectionClosed internal restarting">The connection has been closed because the server is shutting down. See you next time!</div>', 'internal');
```
psykzz:
```suggestion
		internalOutput('<div class="connectionClosed internal shutdown">The connection has been closed because the server is shutting down. See you next time!</div>', 'internal');
```

restarting isn't used for styling its used to generate the "reconnected" message in JS. To avoid confusion in the future use shutdown so it clear.

PR #2736, opened 2019-08-22
Movespeed mod and datum charge/crushing

Rohesie:
* Event-based move delay support added. Should be the standard to move towards for speed mods in the future.
* Charge-crush reworked to allow for more generic behavior, and not be restricted to Crushers only. They are handled through the ability datum now. It's no longer based in `Bump()`, but on a signal sent on `CanPass()` instead, doing snowflake pre-bump checks.
* Vending machines can tip over if they take enough damage.
* Dropship structures are indestructible, not merely unacidable.
* Removed support for crushing multitile hitbox things, but those things are already unsupported.

## Changelog
:cl:
balance: Crusher charge/crushing heavily refactored. There should be some differences in behavior, such as being able to break more things, dealing more consistent damage to objects (thus no longer deleting some and being stopped by others).  Interaction with crushing mobs also changed, should do a bit less damage and have an easier time plowing through them.
balance: Xeno bump delay removed. If xenos want to escape bodyblocking from smaller creatures (such as humans), all they need is to do is be on help intent.
/:cl:


code/modules/shuttle/marine_dropship.dm
```diff
@@ -564,7 +564,7 @@
 /obj/structure/dropship_piece
 	icon = 'icons/obj/structures/dropship_structures.dmi'
 	density = TRUE
-	resistance_flags = UNACIDABLE
+	resistance_flags = UNACIDABLE|INDESTRUCTIBLE
```
DominikPanic:
```suggestion
	resistance_flags = RESIST_ALL
```

code/modules/mob/mob_movespeed.dm
```diff
@@ -0,0 +1,158 @@
...
+	if(!LAZYACCESS(movespeed_modification, id))
+		return FALSE
+	LAZYREMOVE(movespeed_modification, id)
+	UNSETEMPTY(movespeed_modification)
+	if(update)
+		update_movespeed(FALSE)
+	return TRUE
+
+///Handles the special case of editing the movement var
+/mob/vv_edit_var(var_name, var_value)
```
DominikPanic: I think that there's already a mob edit var, could you combine those for code quality?

code/modules/mob/living/carbon/xenomorph/charge_crush.dm
```diff
@@ -0,0 +1,434 @@
...
+	tip_over()
+	if(density)
+		return COMPONENT_MOVABLE_PREBUMP_STOPPED
+	charger.visible_message("<span class='danger'>[charger] slams [src] into the ground!</span>",
+	"<span class='xenowarning'>We slam [src] into the ground!</span>")
+	return COMPONENT_MOVABLE_PREBUMP_PLOWED
+
+
+/mob/living/post_crush_act(mob/living/carbon/xenomorph/charger, datum/action/xeno_action/ready_charge/charge_datum)
+	if(density && ((src.mob_size == charger.mob_size && charger.is_charging <= CHARGE_MAX) || mob_size > charger.mob_size))
```
DominikPanic:
```suggestion
	if(density && ((mob_size == charger.mob_size && charger.is_charging <= CHARGE_MAX) || mob_size > charger.mob_size))
```

code/game/atoms_movable.dm
```diff
@@ -335,6 +335,7 @@
 
 
 /atom/movable/proc/throw_at(atom/target, range, speed, thrower, spin)
+	set waitfor = FALSE
```
psykzz: I thought procs were `waitfor = FALSE` by default now?

drawsstuff:
Crusher could use a serious structure damage buff, maybe 5x what it normally is? For perspective,  It now takes 5 max charges to destroy one reinforced table which normally you coulc plow through 8 of in 1 charge

code/game/atoms_movable.dm
Rohesie: They are `set waitfor = TRUE` by default.
If you saw a line about this being deprecated, that's what it means.
It's worded in a confusing manner.

Rohesie:
Structure damage buff applied. Let's see if it's not excessive.

PR #2739, opened 2019-08-22
Allow only one SL for crash

### LINKED ISSUE #2730: Crash - Multiple Squad Leaders
So hi, began as original SL, began to gear up only to notice multiple SLs spawning as the round went on, with new one overidding the powers of the old ones.

psykzz:
Fixed #2730

Allow only one SL for crash


DominikPanic:
There is a bug report, can you link it in the description?

PR #2740, opened 2019-08-22
Fixes marking admin tickest

### LINKED ISSUE #2729: Marking tickets directly prevent answering them.
They can be answered by clicking on the person to mark and answer, but if one marks it first then one gets the _This ticket has already been marked by Rohesie, click the mark button to replace them._ message.

psykzz:
Fixes #2729


psykzz:
look who didn't update master...

code/modules/admin/admin_verbs.dm
```diff
@@ -812,17 +812,17 @@
 
 	if(AH && AH.tier == TICKET_ADMIN && !check_rights(R_ADMINTICKET, FALSE))
 		return
-
+	var/client/usr_client = usr.client
```
DominikPanic: It should be using key, not the client reference, this PR is a regression

psykzz: Reverting all this

Issue #2749, opened 2019-08-23
#1858 Makes the acid spray not work sometimes, even if you're aiming somewhere that should work.


MiniMeatwad:
## Reproduction:
Fire acid spray at this exact position with the exact setup for reproduction: 
https://i.imgur.com/Ih8PFzQ.png
Seems like the acid spray check may be failing as it's checking for obstacles. 
It has 3 possibilities when used as shown in the picture above:
Acid is sprayed correctly;
https://i.imgur.com/Ih8PFzQ.png
Acid is sprayed incorrectly; 
https://i.imgur.com/vRiAmwm.png
Acid doesn't spray at all; 
https://i.imgur.com/I41btTE.png

In gameplay this works out to be that the acid spray sometimes works and sometimes doesn't if you aim too far away from yourself, since there is almost always some amount of walls nearby. 

This causes a scenario such as the Praetorian tries to fire 5 acid spray shots however none of them work and the Praetorian dies. 
Often times you have to spam click spray acid until it works and if you're unlucky you just die trying to use it, with the acid spray either not working or firing somewhere completely bizarre.

PR #2750, opened 2019-08-23
Removes a frustrating warning when using acid spray

DominikPanic:
Fixes: #2749

If there's 3 valid tiles in front of you, then a wall, and you aim behind that wall, this would prevent you from spraying for zero benefit basically, this should and is handled by the acid itself not passing through walls and obstacles.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -602,7 +602,7 @@
 	var/turf/T2 = get_turf(A)
 	if(T == T2)
 		if(!silent)
-			to_chat(owner, "<span class='xenowarning'>That's far too close!</span>")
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -602,21 +602,9 @@
 	var/turf/T2 = get_turf(A)
 	if(T == T2)
 		if(!silent)
-			to_chat(owner, "<span class='warning'>That's far too close!</span>")
+			to_chat(owner, "<span class='xenowarning'>That's far too close!</span>")
```
Rohesie:
```suggestion
			to_chat(owner, "<span class='warning'>That's far too close!</span>")
```
For failed actions `warning` is generally better.
`xenowarning` tends to be used on other contexts, such as feedback from events happening around. Would like to switch the failed actions in the long term to the generic warning red text.

PR #2755, opened 2019-08-23
Roundstart xenos start at max upgrade

psykzz:
Roundstart xenos start at max upgrade

## Changelog
:cl:
balance: (Crash Only) Xenos now start with the first tier already unlocked.
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -160,6 +160,15 @@
 		computer_to_disable.machine_stat |= BROKEN
 		computer_to_disable.update_icon()
 
+	for(var/i in GLOB.alive_xeno_list)
+		if(isxenolarva(i)) // Larva
+			var/mob/living/carbon/xenomorph/larva/X = i
+			X.amount_grown = X.max_grown
+		else // Handles Shrike etc
+			var/mob/living/carbon/xenomorph/X = i
+			X.upgrade_stored = X.xeno_caste.upgrade_threshold
+		
+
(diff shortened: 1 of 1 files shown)


DominikPanic:
Changelog balance changes

code/datums/gamemodes/crash.dm
```diff
@@ -160,6 +160,15 @@
 		computer_to_disable.machine_stat |= BROKEN
 		computer_to_disable.update_icon()
 
+	for(var/i in GLOB.alive_xeno_list)
+		if(isxenolarva(i)) // Larva
+			var/mob/living/carbon/xenomorph/larva/X = i
+			X.amount_grown = X.max_grown
```
Rohesie: Could you handle this through a proc, so the icons and name are updated properly?

DominikPanic: They will be, just a Life tick later. But yeah, that would be better, depending on if we want them to experience the cool effects.

psykzz: the current progress proc sucks tbh
```
//Larva Progression.. Most of this stuff is obsolete.
/mob/living/carbon/xenomorph/larva/update_progression()
	if(amount_grown < max_grown)
		amount_grown++
	if(!isnull(src.loc) && amount_grown < max_grown)
		if(locate(/obj/effect/alien/weeds) in loc)
			amount_grown++ //Double growth on weeds.
```

If you want i can just call the other procs that follow the progress in life() but its just duplicating code. 

I dont think its worth adding a proc just to handle this when its handled a tick later as Lak mentioned.

Rohesie: Fair enough.

PR #2764, opened 2019-08-23
Round end after xenos defuse the nuke

psykzz:
Assuming no humans left on the planet this ends the round early so we don't randomly wait around.


Rohesie:
Maybe an announcement for the marines?

psykzz:
Added an annoucement @Rohesie

Rohesie:
Not sure if related, but xenos are winning even if the planet is nuked, as long as there's a single larva in the Canterbury's ship.
Which is making some larvas do that.

DominikPanic:
Ready to merge after Rohesie's concern is addressed.

psykzz:
@Rohesie @DominikPanic should be fixed, i believe it was caught in the boolean order of operations.

Rohesie:
Crash rounds are not ending automatically even after the Canterbury takes off and the place is nuked.

psykzz:
Any chance you can get more detail? I tested this a lot after the last batch because of the amount of times it was failing.

Rohesie:
If the marines set up the nuke and take off during Crash, then the round doesn't end and an admin has to end it.

If they take off without activating the nuke the round ends normally in a Xeno Minor.

spookydonut:
@psykzz please address roh's bugs and i can merge

psykzz:
@Rohesie so i think i figured this out, it wasn't a bug just implementation detail.
It takes 5 minutes for the evac to complete and for the marine major you had to wait that time, i suspect most people complained before then.

I've changed this now to be aslong as they have started the evac they win.

PR #2770, opened 2019-08-23
Xenos empty their stomachs when passing fog

psykzz:
## Changelog
:cl:
balance: Xenos empty their stomachs when passing fog
/:cl:

### DIFF
--- a/code/game/objects/effects/misc.dm
+++ b/code/game/objects/effects/misc.dm
@@ -111,6 +111,8 @@
 
 /obj/effect/forcefield/fog/CanPass(atom/movable/mover, turf/target)
 	if(isxeno(mover))
+		var/mob/living/carbon/xenomorph/X = mover
+		X.empty_gut()
 		return TRUE
 	return FALSE
 


code/game/objects/effects/misc.dm
```diff
@@ -111,6 +111,8 @@
 
 /obj/effect/forcefield/fog/CanPass(atom/movable/mover, turf/target)
 	if(isxeno(mover))
+		var/mob/living/carbon/xenomorph/X = mover
+		X.empty_gut()
```
DominikPanic: This should be in /Crossed, behavior shouldn't happen in a Can function

psykzz: I agree but /Crossed would be too late.

psykzz:
I thnk we should look at another way to solve this by changing the fog.

code/game/objects/effects/misc.dm
Rohesie: If it happens in `Crossed()` the marine will be expelled over the fog tile.
In here it will happen before, which is what's intended.

PR #2772, opened 2019-08-23
Smartgun aimslowdown

Rohesie:
## Changelog
:cl:
balance: Smartgun slowdown while wielded increased some to make it more defensive and less mobile..
/:cl:


code/__DEFINES/conflict.dm
```diff
@@ -103,6 +103,7 @@
 #define SLOWDOWN_ADS_SPECIALIST_LIGHT	0.75
 #define SLOWDOWN_ADS_SCOPE				1
 #define SLOWDOWN_ADS_SPECIALIST_MED		1
+#define SLOWDOWN_ADS_SMARTGUN			1.5
```
DominikPanic: In the long term, wouldn't it make sense to just use these values directly and put a comment on var/aim_slowdown explaining how it works?

Rohesie: Possibly.
On one hand it's good to have an easily accessible reference at no cost.
On the other it's nice to be able to check the code directly without having to hunt for a define.
I'm not really sure either way, and have no strong opinions here.

Issue #2724, opened 2019-08-18
Deployed shotgun ammo boxes can be moved.


Hughgent:
## Testmerges: 

None, Master

## Reproduction:

Vend a Shotgun ammo box, the 100 round ones, any type. 
Deploy it on the ground so that you can take ammo out.
click it with an open hand to take ammo out.

The box will move(slide) onto your tile after the ammo handful spawns in your hand.

![image](https://user-images.githubusercontent.com/45076386/63225597-2b2b6f00-c197-11e9-9776-d545611c91d8.png)

![image](https://user-images.githubusercontent.com/45076386/63225600-31215000-c197-11e9-8d35-945cda1502ac.png)

DominikPanic:
Unable to reproduce this https://puu.sh/E8HPx/1eeabfcb91.mp4

Rohesie:
Current master: https://i.imgur.com/BqFEDf8.gifv
Vend the box from req, click on the box, it will travel to you.

PR #2776, opened 2019-08-24
Fixes shotgunbox issues

DominikPanic:
This gets rid of the parent logic but unfortunately I don't see a cleaner solution for this

Fixes: #2724


Rohesie:
The alternative is `afterattack()` or something like that.

DominikPanic:
Afterattack is for objects, not empty hands

PR #2782, opened 2019-08-24
Fulton extraction devices

Rohesie:
Fulton extraction devices. Sprites, sounds and basic design taken from /tg/.
They can export dead xenos and closed crates, selling them for points. They return less points, but it can be done in the field (only in open areas).

First step towards a Crash-friendly Req system.


code/modules/requisitions/fulton.dm
```diff
@@ -0,0 +1,128 @@
+/obj/item/fulton_extraction_pack
+	name = "fulton extraction pack"
+	desc = "A balloon that can be used to extract equipment or personnel to a Fulton Recovery Beacon. Anything not bolted down can be moved. Link the pack to a beacon by using the pack in hand."
+	icon = 'icons/obj/items/fulton.dmi'
+	icon_state = "extraction_pack"
+	w_class = WEIGHT_CLASS_NORMAL
+	tool_behaviour = TOOL_FULTON
+	resistance_flags = UNACIDABLE|INDESTRUCTIBLE
+	var/atom/movable/vis_obj/fulton_baloon/baloon = new()
+	var/obj/effect/fulton_extraction_holder/holder_obj = new()
```
DominikPanic: Do these two in initialize

code/modules/requisitions/fulton.dm
```diff
@@ -0,0 +1,128 @@
...
+
+/obj/effect/fulton_extraction_holder
+	name = "fulton extraction holder"
+	desc = "You shouldn't see this."
+	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
+
+
+//Overrides.
+/mob/living/carbon/xenomorph/fulton_act(mob/living/user, obj/item/I)
+	if(!SSpoints)
```
DominikPanic: This should never happen

code/modules/requisitions/fulton.dm
```diff
@@ -0,0 +1,128 @@
...
+		return FALSE //Place it in.
+	to_chat(user, "<span class='warning'>Cannot extract [src].</span>")
+	return TRUE
+
+
+/obj/structure/closet/crate/fulton_act(mob/living/user, obj/item/I)
+	if(opened)
+		return FALSE //Place it in.
+
+	if(!SSpoints)
```
DominikPanic: Same

code/modules/requisitions/_supply_export.dm
```diff
@@ -0,0 +1,23 @@
...
+/obj/structure/closet/crate/supply_export()
+	. = 3
+	SSpoints.supply_points += .
+
+
+/mob/living/carbon/xenomorph/supply_export()
+	switch(tier)
+		if(XENO_TIER_ZERO)
+			SSpoints.supply_points++
+			. = 1
```
Ghommie: Change this to
```
			. = 2
```
and remove the line above.

Rohesie: Forgot to remove that when I changed the system. They were supposed to net a single point, yea.
This is for larvas.

code/modules/requisitions/fulton.dm
Rohesie: It's there for extensibility. `SSpoints` is not an essential feature like chat or whatever, which should be expected to always exist, even if it currently does.

DominikPanic: Well if it's removed, the code won't compile, bringing us to this check.

Rohesie: If `SSpoints` is disabled in a game mode that doesn't use it (which is a perfectly reasonable thing to do) and there's mapped objects like this one (also a fair assumption), the game will compile and break stuff.

The point of is not leaving the code hostage and prone to failures in case of things changing.

Rohesie:
This should be functional and work for testing on Distress, since that one has a Req system.
Things extracted with the Fulton system will yield less points than if extracted through Requisitions.

code/modules/requisitions/fulton.dm
DominikPanic: SSpoints is gamemode agnostic and works on both gamemodes currently

PR #2785, opened 2019-08-24
Fixes visual issues with girders

DominikPanic:


code/game/objects/structures/girders.dm
```diff
@@ -256,14 +257,12 @@
 
 /obj/structure/girder/update_icon()
 	if(obj_integrity <= 0)
-		icon_state = "[icon_state]_damaged"
-		ENABLE_BITFIELD(resistance_flags, UNACIDABLE)
+		icon_state = "girder_damaged"
+		ENABLE_BITFIELD(resistance_flags, UNACIDABLE|INDESTRUCTIBLE)
```
Rohesie: This would be a great place to use `integrity_failure` for the change, so it becomes non-dense and changes icon, but can still be destroyed if needed if it takes more damage (likely melee).

Are you up for it or want me to do it?

DominikPanic: That would be changing the behavior, girders here specifically work like this

Rohesie: Alright, I'll do it.

PR #2796, opened 2019-08-25
Girder refactor, barricade cleanup, projectile tweaks.

Rohesie:
As per discussed in #2785
* Reworked girders.
* Reworked bullet hit chance on structures.
* Cleaned up barricade code a little bit.
* Unanchored structures can now be hit but they offer half protection, and in the case of metal/plasteel barricades and girders they are more fragile.
* Projectiles diagonal movement made more pixel-movement compliant, ignoring the on_border objects on borders they don't cross and crossing/uncrossing the others properly..

## Changelog
:cl:
add: Girders reworked. Wall construction formula tweaked. Marines can build reinforced walls now.
tweak: Structure hitchance tweaked. The direction they are facing matters once again (barricades, flipped tables). Even unflipped tables and window frames have a chance, though lower, to stop bullets. Accuracy modifiers, distance between shooter and obstacle, and effective accuracy range of the bullet all affect it.
fix: Windoors no longer block lasers.
add: You can now repair beat-up barricades you couldn't before by adding two sheets of metal/plasteel to them and then welding them.
/:cl:


Rohesie:
This should be ready for review and testing.

PR #2797, opened 2019-08-25
Reduces the roundstart notransform lock time

DominikPanic:
Hopefully shouldn't break anything and be smoother TM


DominikPanic:
Seems to be fine

PR #2854, opened 2019-09-03
Changes the Integrated Storage to hold a gun.

Hughgent:
## About The Pull Request

Talked about how the Integrated Storage armor wasn't being picked. So this is a potential replacement to it's limited usefulness.

## Why It's Good For The Game

makes the Integrated Storage armor set a bit more competitive.

## Changelog
:cl: Hughgent
balance: The Integrated Storage armor no longer is a satchel in your armor. It is now a single slot for carrying a fully kitted weapon or machete.
/:cl:

image

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -116,8 +116,9 @@
 
 /obj/item/storage/internal/suit/marine/M3IS
 	bypass_w_limit = list(/obj/item/weapon/gun/,
-						/obj/item/storage/large_holster/machete)
-	cant_hold = list()
+						/obj/item/storage/large_holster/machete,
+						/obj/item/weapon/claymore/mercsword/machete)
+	cant_hold = list() //for if you want to not allow certain weapons.
 	storage_slots = 1
 	max_storage_space = 8 // Rifle, stock, extended barrel, grip/bipod.
 	max_w_class = 3 //Can fit larger items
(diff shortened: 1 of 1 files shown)


factoryman942:
IS as an extra satchel's worth of random stuff can be useful if you're engi w/welderpack or medic w/satchel though

davidchan13:
I'd rather it be kept satchel size with an option to hold smaller weapons like smg. I think it's less used because there is no wiki or documentation about what the different variations do.

Rohesie:
Holding a gun is fine, but you should restrict which ones.
The M41A with a stock is a bit of an overkill.
Holding up to something like a SMG seems more realistic and better for balance, for hidden guns are kind of an issue.

Hughgent:
With how storage is coded @davidchan13 if it was a satchel that could hold the SMG/weapons, then you'd be able to fit three in there. I also don't feel comfortable implementing all the special `belt holster` storage code to only allow one weapon in there at a time.

I had forgotten about grips though. So the space is increasing to 8. You're a madman with those attachments since you got a gun over 6.

Should someone want to do the satchel but only one gun thing. Then it would be a weapon plus storage space, or just storage space. I think this is the best option, but I'm not up to implementing it.

As for @Rohesie I have to politely disagree. The IS storage space isn't a quick access slot like the normal weapon holster. It's a replacement for the mag harness or it allows versatility in loadout choices at the cost of movement speed and ammo capacity.

I feel if the IS was only relegated to no (size increasing) attachment SMG's then it would never be a competitive choice for players.

all guns are size 4(BULKY), except for pistols which are size 3(NORMAL).

Rohesie:
Fair enough, @Hughgent , let's try it your way and see how it goes.

Issue #2860, opened 2019-09-04
Flamer boot sprite glitching when a knife is inserted


Henriquekill9576-taken:
## Testmerges:

Server-Side Code Modifications: https://tgstation13.org/parsed-logs/basil/server_side_modifications.dm

Round ID: 958
BYOND Version: 512.1464
Server revision compiled on: 
Master commit: 93fb8be0e57b0905a41b8c541046c43bb7c73a78
The following pull requests are currently test merged:
#2833: 'Removes MP skill and tank access from PMC' by Maciekkub at commit fa392ba420
#2854: 'Changes the Integrated Storage to hold a gun.' by Hughgent at commit fb2887bfe8
#2853: '4 Helmet Storage Slots' by Aurelien3189 at commit b7d267ebaf
#2852: 'Small M39 Mags' by Aurelien3189 at commit ca3b2fd1fb
#2814: '[W.I.P] Interaction rework' by DominikPanic at commit 13f333de0f
#2795: '[WIP] Pillar of Spring' by grinkgo at commit ffe3da4f0f
#2782: 'Fulton extraction devices' by Rohesie at commit f989905051
#2851: 'Admins can now cancel playing a cinematic' by DominikPanic at commit 585d366991
#2856: 'Fixes xeno fog cheesing' by DominikPanic at commit 9181291f04
#2855: 'Fixes powerloader attack animation' by DominikPanic at commit 048f3e92a3

Local commit: 356b3042a2146eeb6907408881da605e22fafba9
Server tools version: /tg/station 13 Server v3.2.6.2

## Reproduction:

1. Insert a knife into a flamer boot
2. Watch as it's sprite changes into a normal boot

PR #2861, opened 2019-09-04
Fixes pyro marine shoes icon

DominikPanic:
Fixes: #2860


DominikPanic:
Diffbot is broke but basically all I did was rename the icon states to match the code.

PR #2872, opened 2019-09-06
Rage no longer ticks down or gives damage resist

factoryman942:
Rage changes, as stated by drawsstuff:

> 1. Rage no longer goes down over time
> 2. Rage is now only gained from taking damage
> 3. Rage is only expended upon using abilities that are powered by rage
> 4. Rage no longer provides damage resist and fire resist
> 5. Ravager gets back fire immunity or 50% fire immunity

:cl: Drawsstuff and factoryman942
balance: Ravager no longer loses rage over time, only by using abilities.
balance: Rage no longer gives ravagers extra damage resistance or fire resistance.
balance: Ravager no longer gains rage by normal slashing.
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -582,5 +582,4 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 
 // Xeno Cooldowns
 // -- Ravager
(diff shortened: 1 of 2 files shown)


Spyroshark:
you might want to list these as tweaks instead of removal

DominikPanic:
Or even better, Balance

factoryman942:
> Or even better, Balance
k

PR #2883, opened 2019-09-07
M3 pattern Harness

Hughgent:
## About The Pull Request

Had an Idea about how it should allow defibs to work through it, due to no armor. found a nice work around by checking not if they are wearing a suit, but only if that suit conducts electricity.

## Why It's Good For The Game

Another armor choice, with armor being a subjective term. 

It provides no protection, but allows the marine to carry items and have a weapon suit slot.

The sprites have a invisible inner layer that allows for clicking. no need to really pixel hunt.

## Changelog
:cl: Hughgent
add: M3 Pattern harness, no armor plates, no slowdown.
tweak: Defibs work through some kinds of suits now, if your exo suit conducts electricity, no dice.
/:cl:

image
image
image
image
image


code/modules/clothing/suits/marine_armor.dm
```diff
@@ -107,6 +107,13 @@
 	armor = list("melee" = 30, "bullet" = 20, "laser" = 25, "energy" = 10, "bomb" = 15, "bio" = 0, "rad" = 0, "fire" = 10, "acid" = 10)
 	slowdown = SLOWDOWN_ARMOR_LIGHT
 
+/obj/item/clothing/suit/storage/marine/M3harness
+	name = "\improper M3 pattern marine harness"
+	desc = "A standard Marine M3 Pattern Harness. No encumbrance and no protection."
+	icon_state = "10"
+	armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
+	slowdown = NONE
```
DominikPanic:
```suggestion
	slowdown = 0
```
Slowdown is not a bitflag, but a number.

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -107,6 +107,13 @@
 	armor = list("melee" = 30, "bullet" = 20, "laser" = 25, "energy" = 10, "bomb" = 15, "bio" = 0, "rad" = 0, "fire" = 10, "acid" = 10)
 	slowdown = SLOWDOWN_ARMOR_LIGHT
 
+/obj/item/clothing/suit/storage/marine/M3harness
```
DominikPanic: Ideally typepaths shouldn't contain capital letters
```suggestion
/obj/item/clothing/suit/storage/marine/harness
```

Rohesie:
Very interesting. Let's see how it plays out.

Hughgent:
> 
> 
> there's actually a clothing var for this called siemens coefficient, 1 is full conductivity and 0 is completely insulated.

I argue that it should stay as is.

the defib is meant to work via direct skin contact. And although we could pretty easily implement the Siemens coefficient as a `prob(1-Siemens_Coefficient)` sort of check. I see the armor spreading out the electric shock thus making it ineffective.

The second reason is that I doubt anyone wants to balance how easy it is for random chance to allow a defib. I'd rather it be pretty cut and dry whether the defibs will work or not.

Rohesie:
This has been tested and works.

PR #2909, opened 2019-09-10
[s] Input sanity

DominikPanic:


actioninja:
just this isn't actually solving the issue, see the updated made in recent commits to the tg PR.

PR #2914, opened 2019-09-13
IFF bullets for M4RA

factoryman942:
20 bullets/mag
2 mags in kit
deal 30/50 damage
IFF

X3JetwnlRt
_(no, the ammo hud icon doesn't appear in the gif; yes, i fixed that)_

:cl:
add: M4RA scout spec now gets two magazines of Smart bullets in their kit box.
add: They deal about the same damage as rifle AP, and have IFF.
add: Each magazine holds 20 bullets, vs 15 for normal BR magazine. 
add: They can also be ordered from req, at the same price as other scout magazines.
/:cl:


Rohesie:
Realism-wise a bit odd that the IFF is placed on the bullet.

Other than that, would need testing for balance, but I suspect this may be a bit too powerful. We'll see.

factoryman942:
As far as balance, they're the same damage as smartgun bullets, with +30 AP; it's not really clear to me how much stronger that is, but it seems a fair trade for having 40 bullets total instead of 500.

factoryman942:
@Rohesie the nice spooky man fixed it

Rohesie:
There are reports of the ammo sprites being broken.

PR #2915, opened 2019-09-13
Material refactor

DominikPanic:
This PR has several effects:
-gets rid of the matter list, replacing it with datumized materials. right now they are a bare base but make further changes very trivial.
-gets rid of origin_tech, this combined with the previous change will make porting techwebs and doing a proper research system also pretty easy. the reason these two are combined is because both are one line and both are next to each other, combined with their similar purpose and conflicting nature it made sense to combine these. the research machiens are unfortunately included here due to the way they are coded.

this PR will open up several possibilities in the future, many of which i'll likely do myself:
-finish up construction refactor
-proper research system
-stock parts and their effects on machinery


code/game/objects/items/scanners.dm
```diff
@@ -18,9 +18,6 @@ REAGENT SCANNER
 	w_class = WEIGHT_CLASS_SMALL
 	item_state = "electronic"
 
-	matter = list("metal" = 150)
```
Rohesie: Is removing the custom values of these items intentional?

Rohesie:
Good cleanup. This doesn't look like it really needs much live testing.

code/game/objects/items/scanners.dm
DominikPanic: Yes, I made some items more generic, being inherited from the base /item path. Can tweak if you'd like, but these weren't used much if at all, this was done to make future balancing simpler.

Rohesie: No need if it's intentional. This is fine.

PR #2924, opened 2019-09-15
Armor and embedding reworks

Rohesie:
* Armor reworked. It soaks 10% of its value, and then reduces damage by a % equivalent to the rest of it. Penetration is still a flat reduction of armor, so damage + penetration should be kept low, but either can be sort of high while retaining a proper counter.
* Incendiary bullets respect fire armor. 100 or superior values in fire armor is equivalent to the old fire immunity. Negative values are possible, to represent fire weakness. The amount determines the number of fire stacks.
* Crest defense received a buff in the sense that it will trigger from bullets coming from the 3 tiles on the direction the mob is facing. In order to hit a crest user's weak spot, however, the bullet must come from the tile opposite to the one it's facing.
* Tweaked the feedback process on being hit by a bullet to reduce spam a tiny bit.
* Embedding process reworked, opening the possibility of better customization by using /tg/'s embedding behavior datum, tweaked to our needs. Should allow for viable thrown weapons against xenos, or even shrapnel-embedding-based bullets (or grenades).
* Several balance values tweaked. Armor and bullet values. The idea is to nerf damage some and differentiate AP from raw damage ammo.
* Removed some randomness on the damage variance for cleaner balancing.
* Fixed a small issue with projectiles bypassing structures.

## Changelog
:cl:
refactor: Armor and embedding reworked. The changes are a bit too big for the changelog, consult the PR.
balance: Armor, weapon penetration and embedding values changed quite a lot. Meta may be affected.
/:cl:


Rohesie:
Looks ready for merge. Been tested worked.

PR #2929, opened 2019-09-17
Fixes three small issues

DominikPanic:
1. the bypass is no longer necessary
2. the server operator will have to enable the hub in the config, since last I remember it wasn't there to not appear on the hub
3. new players can now finally click buttons on the hotkey popup

will port more if I discover

### DIFF
--- a/code/datums/browser.dm
+++ b/code/datums/browser.dm
@@ -154,6 +154,10 @@
 	set_content(output)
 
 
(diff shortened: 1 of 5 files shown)


DominikPanic:
Anyone else is free to port this since it's AGPL code, but I won't contribute where I'm not welcome.

PR #2939, opened 2019-09-19
Adds a heavy-duty harness to the minigun

Spyroshark:
## About The Pull Request
Adds the Heavy-Duty Harness as a unique belt in the minigun kit that functions like the mag harness attachment, but takes up the belt slot. (The gun is still re-attached to the armor, not the belt slot)

Thanks to Kuro for the sprites.

## Why It's Good For The Game
This makes it harder (but still do-able) to bring a second large gun alongside the minigun. Right now it's much harder for xenos to counter the minigun when the spec can whip out a shotgun or pulse rifle, and players will now have to decide between having the harness on the minigun, or keeping the belt slot for holding the ammo, so they can keep a second gun on their back.

## Changelog
:cl:
balance: The minigun now requires a unique belt harness to keep it attached to armor instead of the magnetic harness. The belt harness can be found in the minigunner kit.
/:cl:


code/game/objects/items/storage/marine_boxes.dm
```diff
@@ -493,10 +493,21 @@
 	new /obj/item/clothing/suit/storage/marine/specialist(src)
 	new /obj/item/clothing/head/helmet/marine/specialist(src)
 	new /obj/item/weapon/gun/minigun(src)
+	new /obj/item/minigun_harness(src)
 	new /obj/item/ammo_magazine/minigun(src)
 	new /obj/item/ammo_magazine/minigun(src)
 	new /obj/item/ammo_magazine/minigun(src)
 
+obj/item/minigun_harness
```
Rohesie:
```suggestion
/obj/item/minigun_harness
```

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -171,6 +171,16 @@ should be alright.
 				harness_return(user)
 				return 1
 
+/obj/item/weapon/gun/minigun/harness_check(mob/user)
+	if(user && ishuman(user))
```
Rohesie:
```suggestion
	if(ishuman(user))
```
This will return false if there's no `user` already.

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -171,6 +171,16 @@ should be alright.
...
+		var/mob/living/carbon/human/owner = user
+		var/obj/item/I = owner.wear_suit
+		if(istype(I,/obj/item/clothing/suit/storage/marine) || istype(I, /obj/item/clothing/suit/armor))
```
Rohesie: Same here, early return:
```
		if(!istype(I, /obj/item/clothing/suit/storage/marine) && !istype(I, /obj/item/clothing/suit/armor))
			return
```

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -171,6 +171,16 @@ should be alright.
...
+			var/obj/item/B = owner.belt
+			if(istype(B,/obj/item/minigun_harness))
```
Rohesie: And same here.

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -171,6 +171,16 @@ should be alright.
...
+				harness_return(user)
+				return 1
```
Rohesie:
```suggestion
				return TRUE
```

code/game/objects/items/storage/marine_boxes.dm
```diff
@@ -493,10 +493,25 @@
...
 
+/obj/item/minigun_harness
+	name = "\improper M45 heavy-duty harness"
+	desc = "A large belt with shoulder straps and clamps to hook onto a minigun, meant to keep the user from dropping their weapon."
+	icon = 'icons/obj/clothing/belts.dmi'
+	icon_state = "heavy_harness"
+	item_state = "heavy_harness"
+	flags_equip_slot = ITEM_SLOT_BELT
+	w_class = WEIGHT_CLASS_BULKY
+	time_to_equip = 20
```
Rohesie:
```suggestion
	time_to_equip = 2 SECONDS
```

code/game/objects/items/storage/marine_boxes.dm
```diff
@@ -493,10 +493,25 @@
...
+
+/obj/item/minigun_harness/equipped(mob/living/carbon/human/user, slot)
+	playsound(src,'sound/machines/click.ogg', 15, 0, 1)
```
Rohesie:
```suggestion
	playsound(src,'sound/machines/click.ogg', 15, FALSE, 1)
```

PR #2944, opened 2019-09-20
Removes piano plus related sounds

Rohesie:
Was kind of broken and unmaintained.
One of several ports of the authorship of LaKiller8.


Rohesie:
Fixed.

PR #2952, opened 2019-09-20
Crash xeno numbers patch

Rohesie:
## Changelog
:cl:
balance: Increased the number of xenos that will naturally spawn in Crash game mode.
/:cl:


Rohesie:
Has been tested. Is not enough to solve the problem but it helps.

PR #2936, opened 2019-09-19
Mrcalzon02maps

spookydonut:
some maps by mrcalzon02

pr for test merging

PR #2940, opened 2019-09-19
Armor resprite 2: Electric Boogaloo

Metroidlover:
## About The Pull Request

Kuro hated his previous work, so he redid all of it, welcome to yet another wave of entirely new armor sprites for the basic variants of marine armor
## Light Armor
Light
## Medium Armor
Regular
## Heavy Armor
Heavy
## Ballistic Armor
Ballistic
## Internal Storage Armor
Storage
## Edge Melee Armor
Edge
## Changelog
:cl: Kuro
imageadd: Brand new Sprites for the basic varients of the Marine armor
/:cl:

PR #2954, opened 2019-09-20
Revert suits oof

Rohesie:
#2936 had a lot of dirty/eol changes. Reverting only the sprites one that breaks them.


Rohesie:
This is also fixed by #2940 , which should be prioritized.

PR #2997, opened 2019-09-24
ZX is now an admin only weapon

Metroidlover:
## About The Pull Request

The ZX now only exists in the code, and it's spec kit is gone, reduced to atoms along with any mention of it
## Why It's Good For The Game

It may be too much work to change it at this point, we can just give the scout another weapon that isn't the ZX
## Changelog
:cl: MetroidLover
del: Removed the ZX scout kit and the ZX crate
/:cl:

### DIFF
--- a/_maps/map_files/generic/Admin_Level.dmm
+++ b/_maps/map_files/generic/Admin_Level.dmm
@@ -679,7 +679,6 @@
 /obj/item/storage/box/spec/heavy_gunner,
 /obj/item/storage/box/spec/pyro,
 /obj/item/storage/box/spec/scout,
-/obj/item/storage/box/spec/scoutshotgun,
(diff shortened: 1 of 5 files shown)


Pariah919:
BEGONE FOUL CREATE!
FRAG MAN BAD!

Kuro-ai:
Back to the pit with you abomination!fragman spec bad

Rohesie:
Things people remove rarely tend to return.
Try rebalancing the weapon and creating something better before removing.

spookydonut:
@Rohesie your call on accepting/rejecting this, either close or merge.

Rohesie:
This is a no until another proper gun substitutes it.
If we start removing every unbalanced gun and feature we'll be left with nothing in no time.
Rebalance it or substitute it with something else.

PR #2998, opened 2019-09-24
VARIANCE BEGONE

Metroidlover:
## About The Pull Request

Removes any and all melee variance from the code

## Why It's Good For The Game

RNG isn't fun nor fair, and when dealing with damage calculations can change wildly form doing nothing to actually killing things
## Changelog
:cl: MetroidLover
balance: All Melee attacks no longer have a random variance and always do the same damage
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/vending.dm
+++ b/code/game/objects/machinery/vending/vending.dm
@@ -156,7 +156,7 @@
 
 	if(M.a_intent == INTENT_HARM)
 		M.do_attack_animation(src)
-		if(prob(M.xeno_caste.melee_damage_lower))
+		if(prob(M.xeno_caste.melee_damage))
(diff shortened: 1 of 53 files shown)


Kuro-ai:
Spike damage begone!!!

Rohesie:
The principle of the change is good, but the values should be either the average or the lower ones.

Metroidlover:
There, every value that had a Vary of more than 5 originally now has been averaged for the most part, every other caste either needs to have damage higher than it's low or is such a small difference it doesn't matter.

Rohesie:
There's several straight buffs here.
Please use the lower value if you don't want to average it.

If you want to buff certain attacks, state which and why.

Metroidlover:
You know what, after doing the math I agree that everything should use it's average, I'll go change that, my bad

Issue #2982, opened 2019-09-22
AI sent as defenceless core to the middle of Deathmatch, fix plox.


Urthrem:
While most dead/civ classes get Freelancers and such the AI is sent to slaughterhouse as core that can't even see anything around it, only able to listen to the fray and wait to be killed.

PR #3004, opened 2019-09-24
makes AI not go to EORDM

factoryman942:
:cl:
fix: AI no longer goes to the end-of-round deathmatch.
/:cl:
they should be treated as ghosts are, meaning they spawn as ERT people in the arena.
Fixes: #2982

### DIFF
--- a/code/datums/gamemodes/game_mode.dm
+++ b/code/datums/gamemodes/game_mode.dm
@@ -483,7 +483,7 @@ Sensors indicate [numXenosShip ? "[numXenosShip]" : "no"] unknown lifeform signa
 			continue
 
 		var/mob/living/L
-		if(!isliving(M))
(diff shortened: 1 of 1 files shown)


code/datums/gamemodes/game_mode.dm
```diff
@@ -483,7 +483,7 @@ Sensors indicate [numXenosShip ? "[numXenosShip]" : "no"] unknown lifeform signa
 			continue
 
 		var/mob/living/L
-		if(!isliving(M))
+		if(!isliving(M)|isAI(M))
```
Rohesie:
```suggestion
		if(!isliving(M) || isAI(M))
```

factoryman942: heck

PR #3021, opened 2019-09-27
Ravagers now gain rage on slashing humans; no more rage gained from taking damage

ma44:
## About The Pull Request

Removes the component that makes ravagers gain rage upon taking damage and reintroduces the mechanic of ravagers gaining rage when slashing humans

## Why It's Good For The Game

Ravagers should be rewarded more for actually slashing humans instead of tanking damage, healing then deciding their finally going to use abilities; even in a best case scenario of a ravager hitting 3 dudes with their 1 tile ravage and getting 30 rage, they still need to take about 80-100 damage to actually get full rage and keep it there
## Changelog
:cl:
tweak: Ravagers now gain rage upon slashing humans; reward for actually getting into combat and slashing things
tweak: Ravagers no longer gain rage upon taking damage

/:cl:


Rohesie:
Needs discussion in #balance

PR #3023, opened 2019-09-28
Shuttle ripple effect icon

Rohesie:
Adds the ripple icon, the mechanic is already there, though it doesn't fully cover the intended area.


Rohesie:
This is ready for merge.

PR #3027, opened 2019-09-28
fix travis eol nl check

spookydonut:


Rohesie:
Travis will complain due to #3028 when updated

spookydonut:
@Rohesie travis seems fine

PR #3029, opened 2019-09-28
Ravager is now old ravager, rage removed

ma44:
## About The Pull Request

Removes the rage mechanic from ravager, adjusts charge of the ravager to be closer to the original and overall buffs up ravager as it was weakened when rage was first introduced
## Why It's Good For The Game

Generally rage is seen as a bad mechanic, mainly due to the now burst damage that the ravager has access to with high rage and eviscerating charge and is otherwise a mess, this reverts them to the old ravager

Features:
-Overall similar armor to pre rage ravager
-Increased speed
-Somewhat more health
-Drastic change from young to mature (mainly stun immunity from explosions is regiven)
-Charge has the old cooldown of 12 seconds, however it has less speed and a much shorter range of 4 tiles; ravagers can also only do 2 charges before running out of plasma if not on weeds

## Changelog
:cl:
del: Ravager rage and its mechanics are no longer a thing
balance: Ravager has been reverted to old ravager statline alongside a overall health buff
/:cl:


Spyroshark:
ever since rav got rage it's been useless, this is a good change

Spyroshark:
don't entirely delete rage though, just remove it from the ravenger. it might be useful if we want to try it on another caste, like warrior.

InterroLouis:
If you want it to not be stunned by explosions at mature+, give it bomb resist 1 to start and then 2 at mature, 3 is crusher level explosion ignoring bomb resist.

factoryman942:
tfw cm had it right all along

drawsstuff:
Rip in piss rageaver

Rohesie:
Spyro seems to have plans to port this to the Warrior or something like that. Copypasting my message in #balance:
```
Anyway, my point on this is simple: remove ravager rage if you want, but then rework it to add it to the warrior or something.
Try not to kill content without refitting it. It's super easy to remove features, but adding stuff is not so much.
```

Tupinambis:
Rage isn't expressly a bad mechanic on its own, but it just doesn't make sense on a hit and run caste like ravager. Good change.

spookydonut:
Pick this or #3021, close one or both.

code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
```diff
@@ -37,104 +37,3 @@
 	X.throw_at(A, RAV_CHARGEDISTANCE, RAV_CHARGESPEED, X)
 
 	add_cooldown()
-
-// ***************************************
-// *********** Ravage
-// ***************************************
-/datum/action/xeno_action/activable/ravage
```
Rohesie: Old ravager had this ability.
This one and several others could be repurposed to work differently.

Rohesie:
The PR seems to be breaking the game quite a bit.
Folk spawn as broken mobs, there are missing verbs.

Metroidlover:
Not amazing with code, but I imagine that removing lines from mob defines broke a TON of shit

InterroLouis:
Right so, Roh asked me to comment on here so here's my comment. A lot of things were broken in that round, so I'll list them the best I can. First of all, inability to move with the WASD or arrow keys. Second, the inability to use any sort of OOC communication-no ahelps, no OOC, no LOOC, nothing. Most verbs were just entirely gone, along with hotkeys also being broken. I'm going to guess that hotkeys being broken is why movement couldn't be done, but maybe I'm wrong since I don't know code. Also satchels were...tricordrazine pills for some reason, but with satchel storage. 
TRICORDRAZINE

PR #3033, opened 2019-09-29
Adds belt harnesses

Spyroshark:
## About The Pull Request

Adds a belt harness for marines, and a variant for survivors.
You can find the belt harness in marine vendors, and survivors will spawn with one instead of their revolver belt. (The revolver belt is kind of useless anyway)
When you have a harness on your belt slot, whenever you drop a weapon it will snap back to your armor, like the magnetic harness attachment currently does.

This PR also removes the minigun harness and replaces it with a belt harness, since they function the exact same way. Also, this changes survivor loadouts a bit so that they can all carry their weapons on their armor.

## Why It's Good For The Game

Right now you pretty much NEED to use a mag harness on your weapon as a marine. This adds an alternative, so you can actually use other attachments more often, at the expense of storing things in your belt slot. It also gives survivors a way to not instantly lose their gun, giving them a SLIGHTLY better chance of fighting xenos.

## Changelog
:cl:
add: Marines now have belt harnesses, an alternative to using the magnetic harness attachment. Check your vendor for one!
add: Survivors now start with their own variant of the belt harness instead of a revolver belt.
/:cl:


Spyroshark:
added different gear to survivor loadouts so they can all actually use the belt

Spyroshark:
ok multiple people told me to not give them CLF armor on ice colony so I undid that.

PR #3037, opened 2019-09-29
[FIX] Corrects the get_alien proc so it is more random.

Jalleo:
## Why It's Good For The Game

Aliens picking are now actually random instead of the first available in the observer list.

## Changelog
:cl:
fix: The first person on the observers list is no longer the next person to be a alien (If they want to be or get picked by a birthing larva maybe)
/:cl:

relevant PR and issue report to understand it.
https://github.com/tgstation/tgstation/pull/26099
https://github.com/tgstation/tgstation/issues/26098
This is likely a massive issue with poor RNG in places. (Because it might not be RNG at all)

### DIFF
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -55,7 +55,7 @@
 		var/turf/speaker = get_turf(R)
 		if(!speaker)
 			continue
-			
+
 		for(var/turf/T in hear(R.canhear_range,speaker))
 			speaker_coverage[T] = T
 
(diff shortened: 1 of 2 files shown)


Jalleo:
Those are relevant because I suddenly started to think about it when I was playing elsewhere and thought hmm maybe there are issues. Had a quick check and confirmed it. Likely the same in job selection code trying to find the critical area now.

code/__HELPERS/game.dm
```diff
@@ -82,7 +82,7 @@
 /proc/get_alien_candidate()
 	var/mob/dead/observer/picked
 
-	for(var/i in GLOB.observer_list)
+	for(var/i in shuffle(GLOB.observer_list))
```
Jalleo: The observer_list is added to and removed each time someone logs in rather than turns into a ghost. So the first person on the list is the first to log in. That means the older users get first chance for being a xeno (passing checks rather than it being the oldest death) so being truly random is more fair.

PR #3046, opened 2019-10-02
Adds and changes Voice SFX

Spyroshark:
## About The Pull Request

This replaces and adds lots of voice clips. In this PR:
-Replaced a lot of old voice clips from CM. Some of the old clips weren't that good, and most of the old female voice clips sounded more like they were from a porn actor then a marine.
-Rare joke screams are gone, Mostly because a sound that plays EVERY ROUND CONSTANTLY isn't the best place to hide an easter egg.
-More sounds for facehugger infection
-Right before a larva bursts out of someone they will make a dying scream
-Gasp and Choke emotes both have sounds now

Thanks to Drawsstuff for ripping the sounds from Alien: Isolation

## Why It's Good For The Game

More variety good, meme noise bad.

## Changelog
:cl: Spyroshark and Drawstuff
add: New voice clips have been added, and some old ones were replaced.
del: Rare joke screams like the half life scientist and TF2 spy have been removed.
/:cl:


code/modules/mob/living/carbon/xenomorph/embryo.dm
```diff
@@ -169,6 +169,9 @@
 								"<span class='danger'>You feel something ripping up your insides!</span>")
 	victim.jitter(300)
 
+	var/preburst_screams = victim.gender == FEMALE ? get_sfx("female_preburst") : get_sfx("male_preburst")
+	playsound(loc, preburst_screams, 25, 0)
+
```
psykzz: Instead of removing the *scream, could you not work these sounds into those screams?

Spyroshark: they already have screaming in them. Also having the *scream and the larva burst sound play over eachother doesn't work ingame that well, since you can't really hear both

Spyroshark:
the "preburst" sounds are REALLY violent, and are about 3 seconds long each. It's intended to play for the duration of the "violent shaking" before the larva comes out.

Spyroshark:
No need to include a second scream emote over it.

code/modules/mob/living/carbon/xenomorph/embryo.dm
```diff
@@ -169,6 +169,9 @@
 								"<span class='danger'>You feel something ripping up your insides!</span>")
 	victim.jitter(300)
 
+	var/preburst_screams = victim.gender == FEMALE ? get_sfx("female_preburst") : get_sfx("male_preburst")
```
Rohesie: How does this tie with monkeys and moths? Do they use `"male_preburst"` sound?

code/game/sound.dm
```diff
@@ -220,20 +220,36 @@
 				S = pick('sound/voice/alien_queen_command.ogg','sound/voice/alien_queen_command2.ogg','sound/voice/alien_queen_command3.ogg')
 			// Human
 			if("male_scream")
-				S = pick('sound/voice/human_male_scream_1.ogg','sound/voice/human_male_scream_2.ogg','sound/voice/human_male_scream_3.ogg','sound/voice/human_male_scream_4.ogg','sound/voice/human_male_scream_5.ogg','sound/voice/human_male_scream_6.ogg',5;'sound/voice/human_male_scream_joke1.ogg',5;'sound/voice/human_male_scream_joke2.ogg',5;'sound/voice/human_male_scream_joke3.ogg',5;'sound/voice/human_male_scream_joke4.ogg')
+				S = pick('sound/voice/human_male_scream_1.ogg','sound/voice/human_male_scream_2.ogg','sound/voice/human_male_scream_3.ogg','sound/voice/human_male_scream_4.ogg','sound/voice/human_male_scream_5.ogg','sound/voice/human_male_scream_6.ogg')
```
spookydonut: why are you removing the weightings from these options?

Spyroshark: Because I’m removing all the “joke” sounds that currently are weighted to play only on rare occasions.  None of the sounds here are meant to be rare meme sounds

PR #3047, opened 2019-10-03
m43 lens attachments

Hughgent:
## About The Pull Request

This is more of a proof of concept.

It's not balanced in the SLIGHTEST, and will sit in the draft stage until it gets sprites.

now it's all about attachments!

feel free to PR balance changes to my repo!

## Why It's Good For The Game

Now the overcharge mode is dependent on a lens, which also decides what ammo it shoots when overcharged!

old lasgun base damage was 20 damage 25 ap. 'new' is 25 damage, 25 ap.
applying a lens that gives a damage boost via overchage or autofire applies a -0.15 multiplier, which makes damage at 21.25.

old lasgun was 20/25 with 40/40 overcharge. the M43 focus lens makes it equivalent to the old lasgun. (+1.25 per normal shot)

## Changelog
:cl: Hughgent
add: The M43 Sunfury now has different lens attachments that fit to the barrel, they decide what overcharge you have, or if you have one at all!
balance: Tweaks the damage for the M43 Sunfury.
/:cl:


Hughgent:
At the moment, it needs a way of setting the lasguns ammo back to 'normal' upon removing the attachment lens.

Hughgent:
image

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -73,7 +73,7 @@ Defined in conflicts.dm of the #defines folder.
 
 	var/activation_sound = 'sound/machines/click.ogg'
 
-	var/flags_attach_features = ATTACH_REMOVABLE
+	var/flags_attach_features = ATTACH_REMOVABLE //set to NONE to make the attachment unremoveable.
```
spookydonut:
```suggestion
	var/flags_attach_features = ATTACH_REMOVABLE
```

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -83,6 +83,10 @@ Defined in conflicts.dm of the #defines folder.
 	var/attachment_action_type
 	var/scope_zoom_mod = FALSE //codex
 
+	var/ammo_mod = null			//what ammo the gun could fire if overcharge is activated, lasers usually.
```
spookydonut: why are you doing this in the attachment

Hughgent: Same Variable name, but they were both local. Changed the gun one to ammo_diff.

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -157,6 +162,13 @@ Defined in conflicts.dm of the #defines folder.
 	master_gun.movement_acc_penalty_mult	+= movement_acc_penalty_mod
 	master_gun.shell_speed_mod				+= attach_shell_speed_mod
 	master_gun.scope_zoom					+= scope_zoom_mod
+	master_gun.fire_delay					+= fire_delay_mod
+	master_gun.ammo_diff					= ammo_mod
+	if(master_gun.charge_cost)
```
Rohesie:
```suggestion
	if(charge_mod)
```

code/modules/projectiles/updated_projectiles/gun_attachables.dm
```diff
@@ -157,6 +162,13 @@ Defined in conflicts.dm of the #defines folder.
...
+		master_gun.charge_cost				+= charge_mod
+	if(length(gun_firemode_list_mod) > 0)
```
Rohesie: The check is not necessary if we have a `for` already.

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -330,6 +330,10 @@ should be alright.
 		to_chat(user, "<span class='warning'>[attachment] doesn't fit on [src]!</span>")
 		return
 
+	if(overcharge == TRUE)
```
Rohesie: How about something like this?
```
if(overcharge)
	toggle_chargemode(user)
```
That way you don't need to return here.
It would require for you to make `toggle_chargemode()` a gun proc that returns, redefined for the lasgun, though.

code/modules/projectiles/updated_projectiles/gun_helpers.dm
```diff
@@ -477,6 +481,10 @@ should be alright.
 		to_chat(usr, "<span class='warning'>You cannot conceviably do that while looking down \the [src]'s scope!</span>")
 		return
 
+	if(G.overcharge == TRUE)
+		to_chat(usr, "[icon2html(src, usr)] You need to disable overcharge mode to remove attachments.")
```
Rohesie: Same thing here, you can do this instead:
`G.toggle_chargemode(usr)`

Hughgent:
@Rohesie Although that would be nice from a quality of life and user experience perspective. I think they should stay as warnings for two reasons.
1. I'm lazy and this works.
2. It's slightly more lore friendly as you SHOULD turn the power settings down on your weapon before messing about with the 'internals'. ~~In fact I should make it so that it checks if there is a bullet in the chamber for all weapons before allowing the modification of attachments. Can't have those marines accidentally shooting their fingers off when adjusting the gun.~~

psykzz:
An issue appeared via this PR. With the lens you can add a lens that gives you full auto and remove it with another and keep the fullauto.

Can you make sure that when actions are removed the selected action is also reset be null'd?

Hughgent:
@psykzz Hey....

Could you check this proc for me. `/obj/item/weapon/gun/proc/attach_to_gun(mob/user, obj/item/attachable/attachment)`

Cause I'm fairly certain, now that you mention this issue, that this proc doesn't actually 'remove' the previous attachment stats! (keep switching a rail light and a RDS to get 'infinite' accuracy?)

Hughgent:
Scratch that. `Detach()` is called, in the `Attach()` proc. 
As such,
```
	if(gun_firemode == removed_firemode)
		gun_firemode = gun_firemode_list[1]
```
is already part of the remove_firemodes() proc that is called.
Therefore, I have no idea what you mean or how to prevent it.

Hughgent:
I've never even used signals before in my life. And this one instance of them is making me hate them.

spookydonut:
yolo work out the balance later.

PR #3048, opened 2019-10-05
HPR Rework

Metroidlover:
## About The Pull Request

The HPR rework demanded to be made by MJP
Pros:

- Burst upped to 5. 
- The guns fire delay down to 0.4.
- Slight accuracy buff of +5% instead of 0%
- Added 1-2 into the Req vendor.

Cons:

- Unable to run a QFA, BFA, or BC
- Wield Delay upped to 2 seconds
- Mags now pretty much only fit in Backpacks and Satchels
- 0.8 slowdown
- 200 round magazines

## Why It's Good For The Game

From MJP: The HPR in its current state is basically an M41A1 with 300 rounds and worse scatter rather than it's intended "LMG" status. This rework aims to make the HPR more of it's intended LMG status by making you slower but giving you 5 round bursts. The ammo capacity was reduced from 300 to 200 since it was deemed like it was too much.

## Changelog
:cl: MJP & MetroidLover
balance: HPR rework
balance: Positive changes; Burst upped to 5, The guns fire delay down to 0.4, Slight accuracy buff of +5% instead of 0%, Added 1-2 into the Req vendor.
balance: Negative Changes; Unable to run a QFA, BFA, or BC, Wield Delay upped to 2 seconds, Mags now pretty much only fit in Backpacks and Satchels, 0.8 slowdown, 200 round magazines, 
/:cl:

### DIFF
--- a/code/modules/projectiles/updated_projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/updated_projectiles/magazines/rifles.dm
@@ -87,6 +87,7 @@
 	desc = "A semi-rectangular box of rounds for the M41AE2 Heavy Pulse Rifle."
 	icon_state = "m41ae2"
 	caliber = "10x24mm caseless"
+	default_ammo = /datum/ammo/bullet/rifle
 	w_class = WEIGHT_CLASS_NORMAL
 	max_rounds = 200
 	gun_type = /obj/item/weapon/gun/rifle/lmg
--- a/code/modules/projectiles/updated_projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/updated_projectiles/magazines/rifles.dm
@@ -82,10 +82,11 @@
 //-------------------------------------------------------
 //M41AE2 HEAVY PULSE RIFLE
 
-/obj/item/ammo_magazine/rifle/lmg
+/obj/item/ammo_magazine/lmg
 	name = "\improper M41AE2 ammo box (10x24mm)"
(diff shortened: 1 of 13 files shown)


code/modules/projectiles/updated_projectiles/magazines/rifles.dm
```diff
@@ -82,11 +82,13 @@
 //-------------------------------------------------------
 //M41AE2 HEAVY PULSE RIFLE
 
-/obj/item/ammo_magazine/rifle/lmg
+/obj/item/ammo_magazine/lmg
```
spookydonut: why

Metroidlover: Because I need the mags to not be able to be held by belts and such.
As far as I know, it was either this or having to specify every child of /obj/item/ammo_magazine/rifle/ for everything with an allowed list

Rohesie:
Could you fit those balance changes in one or two lines, not to overfill the changelog?

Metroidlover:
Like that?

code/modules/projectiles/updated_projectiles/magazines/rifles.dm
spookydonut: you should probably just use w_class although its a bit of mess at the moment with regular magazines being w_class 3

spookydonut: Yeah just stick with the typepath change for now

Rohesie:
The way you did it now is good.

PR #3058, opened 2019-10-06
Increases the bonus to larva growth speed when a host is nested.

Isy232:
## About The Pull Request

Changed the bonus to larva growth speed when the host is nested from:
1 + (0.03 * host's health)
To:
3 + (0.03 * hosts health)

Before the change it took a 100% healthy host 6 minutes to burst and now it is down to 4.
The impact of this can be felt more when a host is moderately injured. These changes should make larva growth sting still good but not completely necessary to get bursts without waiting nearly 10 minutes.

## Why It's Good For The Game

No more xenos standing around for nearly 10 minutes waiting for a host to burst when no drone is around.
Also a drone is still useful, but not completely necessary. This is good for lowpop so a hive does not have to sacrifice a combat capable xeno for one that only goes around weeding and stinging hosts if some get captured.
Its extremely boring to be a drone and to just be around for the purposes of stinging hosts with larval growth toxin as the shrike is already capable of providing enough weeds and making nests.

## Changelog
:cl:
balance: Adds an extra +2 to the larva growth speed bonus from nesting a host.
/:cl:


code/modules/mob/living/carbon/xenomorph/embryo.dm
```diff
@@ -75,7 +75,7 @@
 /obj/item/alien_embryo/proc/process_growth()
 
 	if(CHECK_BITFIELD(affected_mob.restrained_flags, RESTRAINED_XENO_NEST)) //Hosts who are nested in resin nests provide an ideal setting, larva grows faster.
-		counter += 1 + max(0, (0.03 * affected_mob.health)) //Up to +300% faster, depending on the health of the host.
+		counter += 3 + max(0, (0.03 * affected_mob.health)) //Up to +300% faster, depending on the health of the host.
```
factoryman942:
```suggestion
		counter += 3 + max(0, (0.03 * affected_mob.health)) //Up to +500% faster, depending on the health of the host.
```
honk

spookydonut:
```suggestion
		counter += 3 + max(0, (0.03 * affected_mob.health))
```

spookydonut:
PR descriptions don't need to be essays, just keep it short especially 1 line changes like this, looks good though it's subject to balance discussion

Rohesie:
This makes the crit-and-nest meta viable for xenos. Might not be well received.

Spyroshark:
maybe we could have it only buff hosts who are healthy?

spookydonut:
@Rohesie I intend to merge this tomorrow.

Rohesie:
@spookydonut , this hasn't been tested enough, and Spyro's suggestion is a better idea.

PR #3064, opened 2019-10-08
Stops resin sacs and resin holes being placed on the same turf

spookydonut:
## Changelog
:cl:
balance: resin holes and resin sacs can no longer be placed on the same spot
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -49,6 +49,13 @@
 	mechanics_text = "Plant a weed node (purple sac) on your tile."
 	keybind_signal = COMSIG_XENOABILITY_DROP_WEEDS
 
+/datum/action/xeno_action/plant_weeds/can_use_action(silent, override_flags)
+	. = ..()
+	if(locate(/obj/effect/alien/resin/trap) in get_turf(owner))
+		if(!silent)
+			to_chat(owner, "<span class='warning'>There is a resin trap in the way!</span>")
+		return FALSE
+
 /datum/action/xeno_action/plant_weeds/action_activate()
 	var/turf/T = get_turf(owner)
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -49,6 +49,13 @@
 	mechanics_text = "Plant a weed node (purple sac) on your tile."
 	keybind_signal = COMSIG_XENOABILITY_DROP_WEEDS
 
+/datum/action/xeno_action/plant_weeds/can_use_action(silent, override_flags)
+	. = ..()
+	if(locate(/obj/effect/alien/resin/trap) in get_turf(owner))
```
Rohesie: This will be called all the time, constantly locating for traps under the mob, even if it's not trying to use the ability. It would be better if this was done on demand, in `action_activate()`, no?

Also, shouldn't this be on `check_alien_construction()`?

spookydonut: check_alien_construction fails on a lot of things weeds shouldnt fail on.

can_use_action should in theory be on demand, sadly it looks like theres a proc chain from life that will trigger it.

in terms of xeno actions, can_use_* is meant to check if the action can be done, and action_activate/use_ability do it without (much) checks

code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
```diff
@@ -152,6 +152,11 @@
 	if(!T.check_alien_construction(owner, silent))
 		return FALSE
 
+	if(locate(/obj/effect/alien/weeds/node) in T)
```
Rohesie: What about this one for `check_alien_construction()`?

code/modules/mob/living/carbon/xenomorph/abilities.dm
Rohesie: So keep this as it is, and plan to remove the call from `Life()`?

code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
spookydonut: you want to be able to build resin walls on top of nodes.

PR #3070, opened 2019-10-09
removes fog from distress

spookydonut:
## About The Pull Request
Per popular request

## Changelog
:cl:
del: Fog lines no longer spawn in distress gamemode
/:cl:


Kuro-ai:
Yes please

jroinc:
I think this is a very bad idea if we want rounds to consist of more than 20 min prep and 15 min combat, but let's see how it goes. Can always change later.

PR #3072, opened 2019-10-09
tweak cant to be less xeno rush friendly

spookydonut:
## About The Pull Request
removes some entrances and makes turrets face a better direction

## Changelog
:cl:
balance: removed some entrances from the canterbury and made turrets face outwards
/:cl:

### DIFF
--- a/_maps/shuttles/tgs_canterbury.dmm
+++ b/_maps/shuttles/tgs_canterbury.dmm
@@ -200,7 +200,9 @@
 /turf/open/floor/mainship/mono,
 /area/shuttle/canterbury/cic)
 "aw" = (
-/obj/machinery/marine_turret/premade/canterbury,
+/obj/machinery/marine_turret/premade/canterbury{
+	dir = 8
+	},
 /obj/machinery/light/small{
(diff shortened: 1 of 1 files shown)


Spyroshark:
this is a good change but there's 2 issues
1. You can't put cades in front of the turrets, at least until the ship lands. there's no room in front of the turrets for building. I'd suggest moving them back a tile, like this:
image

2. The outer shutters button is blocked by the shutters now.

InterroLouis:
The buttons for the outer side doors are underneath shutters, which makes it so you have to alt click just to open it. Second, it seems just as exposed as before. The sentries are incredibly easy to pick off, to the point where the xenos don't even bother with the south entrance-they just pop the turrets and go in that way. I thought outwards facing would be better, but perhaps if we focused them instead on defending the inside of the canterbury-with cover if facing them from the outside of the Cant, so there won't be spit cheese.

Rohesie:
Check #3093

spookydonut:
> 
> 
> Check #3093

@Rohesie  fixed

PR #3078, opened 2019-10-10
XenoSpit to 8

Hughgent:
## About The Pull Request

EOL changes are evil.

## Why It's Good For The Game
Xenos don't need to be perfectly on tile accurate, they can just click in line with their target.

## Changelog
:cl:
tweak: Xeno Spit is range based now. you can click in line with your target.
/:cl:


factoryman942:
still a bandaid fix but it's better without eol changes i guess

Spyroshark:
spit ending where you click is not a fun change anyway, imo this is better

Rohesie:
Deprecated by #3079 ?

Hughgent:
If ya say so.

PR #3081, opened 2019-10-11
Less xenos on Crash

Rohesie:
Little balance patch.


spookydonut:
You may as well just put these in config.

PR #3092, opened 2019-10-13
ports elements

spookydonut:
## About The Pull Request
ports elements

This adds a windowshutter element which is for the prison windows to spawn and close a shutter when they break.
To do this I also had to enforce parent calling for the deconstruct() proc


Rohesie:
Ready for testing?

PR #3099, opened 2019-10-16
Allows for the option to have ageless xenos [WIP] [HORRIFICALLY BROKEN]

Tupinambis:
## About The Pull Request

This is intended to add a special "ageless" version of xenos that have static power regardless of how long they have been alive. It will be an option that can be controlled either via gamemode or via config.

Currently it is WIP and a horrible buggy mess that doesn't work. Assistance requested.

## Why It's Good For The Game

Aging is a terrible mechanic that encourages xenos to delay and hide instead of actively participating in fights, and leads to rounds typically being decided during the first engagement between marines and xenos. After all, if a mature T3 dies early on, it is unlikely a mature T3 will be replacing it any time soon, and this weakens the xenos immediately. Consistent power should lead to easier balancing, less waiting around, and an overall less frustrating time for xenos.

Ideally, a better system could be implemented at a later time, but that would take significantly more code effort.

## Changelog
:cl:
add: Adds the option to give xenos a static amount of power, removing aging. This may be locked to gamemode or config.
/:cl:


spookydonut:
It's okay we can fix this

Tupinambis:
Abandoning this idea completely for the time being, want to focus on bigger issues

PR #3100, opened 2019-10-16
SG and M41A Reworks

Max-023:
## About The Pull Request
The PR reworks the M41A to be more of an assault rifle rather than a crappy high capacity M41A1. Designed by MJP, this should be good for testing. Putting the M41A in Marine vendors to be alongside the M41A1 will wait until the SX PR is merged that adds it.

Unlocking all attachments that the M41A1 has and giving it to the M41A will allow it to be more flexible, however the BFA will still be locked as burst fire delay will be decreased and there will be a 10% accuracy modifier (as opposed to the -5% it had) for the M41A, as well as the skeleton stock only being on the M41A1. It will now come standard with a grenade launcher attached, too. Other changes include an increased aim slowdown to 0.5 and the wield time being bumped up to 1.35 seconds.

Additionally, all M41 rifles will get scatter set to -10 to hopefully make the weapon shoot straighter and compete with the other Marine weapons.

The M56 has gotten some much needed love to make it easier for smartgunner mains to do the job they are supposed to, indirect fire support. It was a bit of a challenge trying to buff the M56 in other was besides damage, but we found that reducing the single fire delay to 0.3 seconds (from 0.4 seconds), giving it the recoil compensator and laser sight, as well as reducing the scatter to -10 will make the gun more worth while to play. We also upped the burst to be 5 rounds per burst and removed the BFA, so players who like to burst fire the M56 do not have to waste an attachment slot. Additionally, the bullet has been given 5 penetration so the gun is more applicable to armored castes, making the bullet a 30/5 round. This should bring it up with the M41E2 and the Vindicator in the indirect fire support weapons category.

## Why It's Good For The Game
Right now the M41s are all known as weaker weapons. With shotguns in the game becoming more prevalent with the SX and the M39 being a solid player choice, the M41s were falling behind, especially the M41A. The M41A was known as being exceptionally low performing due to the low amount of attachment options, the accuracy debuff, and the lack of a real difference in play style from the M41A1, its carbine counterpart. This rework should make it a pronounced weapon that fits its own category and is more flexible to maintain that jack of all trades style the M41 series is known for.

As for the M56 smartgun, where do I begin? Ever since the massive nerfs were applied to the smartgun, it was a shell of its former self. With exceptionally low damage, no penetration, slow firerate, and all the slowness it provided, the weapon was far from useful. The weapon stopped being a fear of the xenos and more of a laughing stock as runners could run through the fire and work unimpeded and armored xenos laughed off the rounds as they plinked off. As a machine gun, it felt very far from what a machine gun is. While it might not be much harder hitting as it was before this PR, this should hopefully make the M56 feel nice to play. At least, it should make current SGner mains not rush to use the M39 over it for suppression.

## Changelog
:cl: Guprei
balance: The M56B smartgun trades the burst fire assembly for the recoil compensator and the laser sight attachments. Scatter and penetration to its ammo increased. Burst amount increased and fire delay reduced.
balance: The M41A old pulse rifle is re-purposed It. can take a lot more attachments. Scatter increased. Wield delay, wield slowdown, burst delay, and accuracy rebalanced.
/:cl:


Max-023:
I did not touch movement speed for the SG.

code/modules/projectiles/updated_projectiles/guns/rifles.dm
```diff
@@ -83,6 +83,7 @@
 	fire_delay = 0.4 SECONDS
 	burst_delay = 0.15 SECONDS
 	accuracy_mult = 1.15
+	scatter = -10
```
Rohesie: Isn't this a -30 scatter buff?

PR #3103, opened 2019-10-16
Stamina

Rohesie:
* Damage-based movespeed uses the new system.
* A ton of duplicate `updatehealth()` removed.
* Some damage procs cleaned up.
* Humans handle damage through the species datum.
* handle_suit_punctures() and other ancient life stuff, or unused procs/defines, gone.

## Changelog
:cl:
add: Stamina type of damage added. It comes and goes faster than other damage types, knocksdown the victim if it has too much of it, and is otherwise harmless.
tweak: Walk mode made severely faster, and the default for humans. They can run by making use of a stamina buffer, which they recover when not running. Synths don't have stamina, so they cannot sprint/run but are unaffected by stamina damage.
balance: Xenos now have melee armor. Humans also have melee accuracy, which is perfect while walking and only 80% when running. Their tackles now  fullyrespect melee armor.
balance: Neurotoxin no longer injects chems. It now deals direct stamina-damage, same as xeno tackles, resin spit and a few other things.
add: Examine is now a keybind, defaulted to Shift. It merges old shift+click and alt+click on a turf near you. Alt defaults to run-while-holding, and you can shoot or attack while doing so. You can rebind it as well.
balance: The hunter is now much faster while stealthed, so as a consequence its sneak attack has been nerfed some. Probably needs further tweaks.
/:cl:


code/__DEFINES/conflict.dm
```diff
@@ -38,8 +38,8 @@
 //Ammo defines for gun/projectile related things.
 //flags_ammo_behavior
 #define AMMO_EXPLOSIVE 			(1<<0)
-#define AMMO_XENO_ACID 			(1<<1)
-#define AMMO_XENO_TOX			(1<<2)
+#define AMMO_XENO 				(1<<1)
+//#define AMMO_XENO_TOX			(1<<2) //Unused value.
```
spookydonut: delete

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -133,10 +133,10 @@
 		var/armor_block = L.run_armor_check("chest", "melee")
 		if(ishuman(L))
 			var/mob/living/carbon/human/H = L
```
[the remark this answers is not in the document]
Rohesie: Legacy of old code. Fixed.

code/__DEFINES/conflict.dm
spookydonut: fair

spookydonut: just uncomment it and leave the unused comment

spookydonut:
@Rohesie merge conflict

Rohesie:
This one depends on #3119

Rohesie:
This needs a lot of balancing, but the base is functional.

Eskjjlj:
After runners and hunters pounce their move intent is automatically switched to Stalk which is the xeno walk mode.

Rohesie:
> After runners and hunters pounce their move intent is automatically switched to Stalk which is the xeno walk mode.

Fixed.

GravitronBlue:
Assuming this PR is responsible I'm not a fan of how the shift-examine and alt key turf menu are combined into a single hotkey.

code/datums/keybinding/mob.dm
```diff
@@ -146,18 +146,64 @@
...
+/datum/keybinding/mob/examine/proc/examinate(datum/source, atom/A, params)
+	var/mob/user = source
+	if(!user.client || !(user.client.eye == user || user.client.eye == user.loc))
+		UnregisterSignal(user, COMSIG_MOB_CLICKON)
+		return
+	user.examinate(A)
+	var/turf/examined_turf = get_turf(A)
+	if(examined_turf && user.TurfAdjacent(examined_turf))
+		user.listed_turf = examined_turf
+		user.client.statpanel = examined_turf.name
```
GravitronBlue: Please consider separating this into its own client preference hotkey or just being used by alt-clicking like it was before.

Rohesie: Can you expand on the reasons a bit?

GravitronBlue: From a marine standpoint the new default alt key being used to examine messes with certain inventory shortcuts like alt-clicking an MRE or pill bottle to grab the first item out from it.
Though this can be avoided by changing keybindings it ruins consistency between here and TG by having to use a different key just to pull up the turf contents menu.

The turf menu also pops up when you examine inventory items, which is a bit of a nuisance when you just want to check gun stats.
There's probably some actual cases where this could be disruptive, like trying to grab a medical scan from a patient while sorting through a pile of medkits on a nearby table.

I've also been unable to examine things using shift-click while as a ghost.

Personally just think this is a specific change that wasn't asked for, though perhaps you'll find a different opinion and support on the server.

davidchan13:
Synths still can't run.

GravitronBlue:
The way stamina damage is applied to barricades and structures needs to be examined, as sentinels are dealing massage damage to fortifications.

Isy232:
Xenos are by default on "stalk" mode. This gets reset every time they evolve. Most xenos are not used to this and I myself can blame several deaths on this, please fix this.

Rohesie:
> The way stamina damage is applied to barricades and structures needs to be examined, as sentinels are dealing massage damage to fortifications using neurospit.

Fixed, objects no longer take bio damage, they have 100 bio armor.

> Xenos are by default on "stalk" mode. This gets reset every time they evolve. Most xenos are not used to this and I myself can blame several deaths on this, please fix this.
> 
> Please just remove stalk from all xenos that don't use it, it exists just to be accidentally pressed and screw you over as it is now.

Fixed. They weren't on stalk mode but changed to it on the toggle intent hotkey, which now toggles it to stalk and goes back to running when released.

PR #3119, opened 2019-10-19
Damage procs cleanup, queued health update

Rohesie:
Atomizing this out of #3103


Rohesie:
I may have overdone it on the health update removal. It's creating delayed damage and some issues.
On the other hand, the lag-upon-being-shot is being reported as smaller. Might be a false positive, we'll see in the long run.

Rohesie:
Ok, added spooky's health update queuing. This is the logic employed on when to update:

* Chem healing and life-based damage changes don't update health. That is done at the end of the life / reagents process proc.
* Combat damage sets to update health at the end of the tick, as multiple sources can happen.
* Precise damage or healing, such as surgery (successes and failures), bandages, and the likes, updates health immediately.

Rohesie:
```
Proc Name                                                                                Self CPU    Total CPU    Real Time     Overtime        Calls
/proc/updatehealthnew                                                                       0.005        0.088        0.090        0.001         1393
/proc/updatehealthnew_called                                                                0.001        0.155        0.153        0.000         1358
/proc/updatehealthold                                                                       0.000        0.202        0.200        0.000         1393
/proc/updatehealthprofile                                                                   0.005        0.294        0.297        0.000         2786
```
All in all, `updatehealthnew` + `updatehealthnew_called` are somewhat more expensive than `updatehealthold` in the bigger picture, but they are cheaper during big bursts, which is what causes client lag and what we want to fix.

Rohesie:
This one is ready.

PR #3133, opened 2019-10-20
Icy Caves Map

Isy232:
## About The Pull Request

This map has already been tested once, after that I gave the map its own areas and fixed up a few things.

This map is likely to have some issues with it, but the more testing it gets the more of these issues I can find and fix.

The areas not being cold is intentional, I may change it but I find cold just to be annoying rather than engaging.

## Why It's Good For The Game

Adds a smaller map meant for lower populations. Also just a new map so a new place to play so its not just the same 4.

## Changelog
:cl:
add: Icy Caves Test Map
/:cl:


Rohesie:
1) Resin silos are not covered by fog.
2) Cave areas have the `always_unpowered = TRUE` var set, because they are copied from LV. All areas should be their own in the map.

Until these two are fixed the map can't be tested again.
There are also overpowered items spread around that need tweaking, but that's secondary.

Rohesie:
Survivor spawns lacking.

Isy232:
Added survivor spawns

jroinc:
There are seven fully-equipped PMC corpses in an area on the map. That's a LOT of high-quality equipment and anti-hug lying around free for the taking. Maybe make the 4 in the "cryo" area naked, so there are only 3 sets available?

Also medbay being right next to the LZ's kinda cute. APC's and cryo tubes can't be melted, and are EXTREMELY useful if someone brings a cryo beaker down (or maybe it's preloaded).

_maps/icy_caves.json
```diff
@@ -0,0 +1,8 @@
+{
+    "map_name": "Icy Caves",
+    "map_path": "map_files/icy_caves",
+    "map_file": "icy_caves.dmm",
+    "shuttles": {
+    },
```
spookydonut:
```suggestion
    "environment_traits": [
        {"COLD": true}
    ],
    "armor": "ice",
```

_maps/icy_caves.json
```diff
@@ -0,0 +1,8 @@
+{
+    "map_name": "Icy Caves",
+    "map_path": "map_files/icy_caves",
+    "map_file": "icy_caves.dmm",
+    "shuttles": {
```
spookydonut:
```suggestion
```

_maps/icy_caves.json
Isy232: I personally think that cold is a shit game mechanic and is just a pain in the ass with very little benefit game play wise.

Setting armor to Ice is good.

Isy232:
I didn't want the map to be cold so I copied spooky's suggestions and pushed them so I updated the armor used to ice armor.

PR #3134, opened 2019-10-20
More xeno vision types

Rohesie:
* Incidentally `update_spits()` was using the mob that was going to be deleted in evolution. I'm not sure that should even be there.

## Changelog
:cl:
add: Xenos can switch with their toggle-nightvision hud icon between 4 light intensity levels, from marine-like no-nightvision to fully shadowless, instead of only having two options.
/:cl:


code/modules/mob/living/carbon/xenomorph/evolution.dm
```diff
@@ -1,11 +1,3 @@
-//Xenomorph Evolution Code - Apophis775 - Last Edit: 11JUN16
-
-//Recoded and consolidated by Abby -- ALL evolutions come from here now. It should work with any caste, anywhere
-// refactored by spookydonut because the above two were shitcoders and i'm sure in time my code too will be considered shit.
```
spookydonut: this should be retained and updated

PR #3137, opened 2019-10-20
M41A vending and FC alterations

Max-023:
## About The Pull Request
This PR is 1) to finish the M41A development in the coding by making it available to Marines in vendors. They have been added to both req and regular armory vendors so Marines can order them and use them alongside the M41A1, the less accurate and lower capacity sister gun. There are 25 guns and 25 mags, same as the M41A1.

Additionally, there are changes to the FC so the FC feels more complete and less like an unfinished setup. Currently, they spawn with an SG harness, SG, power pack, a megaphone, and a command pouch. This PR adds the M56 goggles to the glasses slot (which had nothing in it in the beginning) and replaced the megaphone with a full medkit pouch.

## Why It's Good For The Game
The FC not only has a very high medical skill that can go under used, but they also get a second megaphone in their office in the Pillar of Spring, meaning currently they get TWO megaphones. Additionally, many see no reason for the FC not to get the M56 goggles SGners have as standard. Yes, the FC is a command role and is thus more important, but the SGner has access to more equipment to be applicable in battle. In distress this is less noticeable, but in crash this is very noticeable, as the SGners are the only class with the M56B that has req attachments. These changes should help the FC be more prevalent on the frontline and hopefully more interactive with the Marine force.

As for the M41A, this is just so normal Marines can actually access it. Right now it is very difficult for Marines to obtain one in distress and impossible in crash.

## Changelog
:cl:
add: m56_goggles to the FC glasses slot
add: M41A to req and Marine vendors
add: M41A mags to the req and Marine vendors
tweak: traded out the megaphone for a full medkit for the FC left pouch
/:cl:


code/datums/jobs/job/shipside.dm
```diff
@@ -86,8 +86,9 @@ Make the TGMC proud!"})
 	shoes = /obj/item/clothing/shoes/marine
 	gloves = /obj/item/clothing/gloves/marine/officer
 	head = /obj/item/clothing/head/tgmcberet/fc
+	glasses = /obj/item/clothing/glasses/night/m56_goggles
```
spookydonut: strongly disagree, dont just ignore #balance lol

Max-023: No one but you said FCs with smartgoggles was bad. Not sure what you are referring to.

Max-023:
Closing so I can redo this.

PR #3140, opened 2019-10-20
Non-distress evolution_stored

Rohesie:
## Changelog
:cl:
balance: Xenos are now able to gain evolution points even if there's no ruler on Crash game mode.
/:cl:


code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
```diff
@@ -115,7 +115,7 @@
 
 	if(!(xeno_caste.caste_flags & CASTE_EVOLUTION_ALLOWED))
 		stat(null, "Evolve Progress (FINISHED)")
-	else if(!hive.living_xeno_ruler)
```
spookydonut: change this to a hive proc

code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
```diff
@@ -247,11 +247,13 @@
 		return
 	if(evolution_stored >= xeno_caste.evolution_threshold || !(xeno_caste.caste_flags & CASTE_EVOLUTION_ALLOWED))
 		return
-	if(hive?.living_xeno_ruler)
```
spookydonut: same here then have the hive proc check like a gamemode flag like XENOS_REQUIRE_RULER

PR #3143, opened 2019-10-20
B18 armor no longer autoheals, is now not heavy

Fikou:
## About The Pull Request

the b18 armor no longer autoheals, it still has a health analyzer mode
now slows down the same as regular armor

## Why It's Good For The Game

b18 armor is pretty much unfair, autoheal with 80 protection against everything, now they have to rely on another person (a medic) instead of being able to go solo against any xeno and win and then return with full health

## Changelog
:cl:
balance: b18 armor no longer autoheals, but it now slows down less
/:cl:

### DIFF
--- a/code/__DEFINES/objects.dm
+++ b/code/__DEFINES/objects.dm
@@ -131,18 +131,6 @@ GLOBAL_LIST_INIT(restricted_camera_networks, list( //Those networks can only be 
 #define SCOUT_CLOAK_MAX_ENERGY 100
 #define SCOUT_CLOAK_OFF_DAMAGE (1 << 0)
 #define SCOUT_CLOAK_OFF_ATTACK (1 << 1)
-//B18 DEFINES
-#define B18_CHEM_COOLDOWN				2.5 MINUTES
-#define B18_CHEM_MOD					0.5
-#define B18_BRUTE_CODE					1
-#define B18_BURN_CODE					2
-#define B18_OXY_CODE					3
-#define B18_TOX_CODE					4
-#define B18_PAIN_CODE					5
-#define B18_DAMAGE_MIN					50
-#define B18_DAMAGE_MAX					150
-#define B18_PAIN_MIN					50
-#define B18_PAIN_MAX					150
 
 
 //Razor wire
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -235,6 +235,7 @@
 	flags_cold_protection = CHEST|GROIN|ARMS|LEGS|FEET
 	flags_heat_protection = CHEST|GROIN|ARMS|LEGS|FEET
 	var/obj/item/healthanalyzer/integrated/B18_analyzer = null
+	var/mob/living/carbon/human/wearer = null
 	supporting_limbs = list(CHEST, GROIN, ARM_LEFT, ARM_RIGHT, HAND_LEFT, HAND_RIGHT, LEG_LEFT, LEG_RIGHT, FOOT_LEFT, FOOT_RIGHT) //B18 effectively stabilizes these.
 	resistance_flags = UNACIDABLE
 
@@ -243,8 +244,18 @@
 	B18_analyzer = new /obj/item/healthanalyzer/integrated
 
 /obj/item/clothing/suit/storage/marine/specialist/Destroy()
+	wearer = null
 	qdel(B18_analyzer)
 	. = ..()
+
+/obj/item/clothing/suit/storage/marine/specialist/dropped(mob/user)
+	. = ..()
+	wearer = null
+
+/obj/item/clothing/suit/storage/marine/specialist/equipped(mob/living/carbon/human/user, slot)
+	. = ..()
+	if(slot == SLOT_WEAR_SUIT)
+		wearer = user
 
 /obj/item/clothing/suit/storage/marine/specialist/verb/b18_automedic_scan()
 	set name = "B18 Automedic User Scan"
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -228,7 +228,7 @@
 
 /obj/item/clothing/suit/storage/marine/specialist
 	name = "\improper B18 defensive armor"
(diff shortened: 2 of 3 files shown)


Henriquekill9576-taken:
how dare you

Kuro-ai:
How to make b18 useless in one pr

Max-023:
The entire point of the B18 is to substitute the tank and be a breakthrough armor. Only two roles get it and instead of having very high defenses, it has healing and pain medicine to keep the user up. These nerfs do not fix the B18 as it is perfectly fine. The weapons the B18 gets are hard to frag with and the ultra heavy slowdown, which it already has, means you are not very mobile. There are plenty of ways to counter a B18 user in its current form, as well. I suggest re-evaluating xeno tactics as opposed to immediately going to nerf the armor.

Fikou:
> The entire point of the B18 is to substitute the tank and be a breakthrough armor. Only two roles get it and instead of having very high defenses
what do you mean b18 has better armor than SL. heavy marine armor, practically everything, its 80 across the board

Pariah919:
lol
"Oh my god the b18 didn't die in 3 clicks from acid spit when I expected him to!! Please nerf!!'
B18 armor is fine. Leave it alone.

Max-023:
80 across the board isn't bad. Not to mention xenos interact with less than half of the protections it gives. Armor rarely means anything in SS13 and that is mostly due to the fact that the heavier the armor, the slower you are. And if you are slow, you are easily clicked on and hit. The B18 has ultra heavy slowdown so that is a major drawback.

The only thing that makes the B18 actually worth it and useful is the function that treats your wounds. Mind you, this does have a cooldown and is not constant, but a good B18 player will have to learn how to balance taking hits and making sure their armor can treat their wounds. It also automatically splints broken bones, meaning you don't have to hurt yourself more once they are broken by moving. Two things the suit doesn't treat, IB and foreign objects, still require you to go all the way back to a surgical station the B18 offers no counter. Quite a few castes can instantly break bones and cause IB and a B18 user has to look out for that, but overall little will actually put you out of a round for a while.

Additionally, as long as halloss has no medicinal counter, neuro in any capacity will be a hard counter.



The issue is not that B18 armor wearers are unkillable. They are very mortal. It is that the B18 is anti-attrition and thus xenos need to learn to adapt to that. Xenos don't, so minigunners often see a lot of success. As I said before, this is not a B18 issue. The amount of resistance to these changes clearly shows that most think the B18 is balanced anyway.


As for the weapons, do actually try to use the minigun and grenade launcher. Neither are very favorable.

jroinc:
I'd be 100% fine with removing the automedical capacity. In-combat healing is a significant boon to the B-18, and it is near-impossible to capture one currently, cause they'll heal back to full repeatedly and take minimal damage. The auto-splint system is a big buff as-is, and there's nothing stopping you from popping pills before combat like the rest of the mere mortals out there.

jroinc:
Additionally, broken bones and foreign objects that are autosplinted don't affect the suit at all, due to only causing minor pain (as PFC, I don't even take tramadol for 1 splinted break, it's so light) and the suit's built-in oxy autoinjector. Hell, I could even see keeping the medical systems, just make them manually triggered.

Rohesie:
If you want to nerf the B18, do it to its speed, a little.
Removing the feature completely is not only wildly unpopular, but also not really necessary.

jroinc:
I mean, I was thinking of removing the B-18 from the spec kit entirely, because I kept hearing that the armor was "shit". Didn't think it was THAT popular then.

Rohesie:
B18 armor is great in several senses. It's the best armor around.

jroinc:
I dunno. I've heard a LOT of people call it absolute shit when it's brought up due to the slowdown not letting them be "I am speed". I think it's the best armor, but maybe it's not "meta" anymore.

Kuro-ai:
People call it shit because xenos can delimb your hand in a few hits,in some cases one hit,you know that stop playing dumb jpr

Rohesie:
Delimb random chance has been fixed by a different PR.
If you want an actual nerf to the B18 you can reduce the melee armor by like 5.
Otherwise removing the functionality is a no-go.

factoryman942:
if you remove the autodoc you'll probably have to return it to 95 melee defence

...and give it the two magic tricord injectors back

davidchan13:
Xeno mains complaining that a single enemy combatant is resistant to their primary form of lethal combat and requires a coordinated effort to pin and kill said combatant or else their foe might escape and be back on the front lines not 5 minutes later at full health.... Hmmmmmmm.

jroinc:
David, I don't give a shit about their armor or autoheal. I'm annoyed that they have both, causing only scratch damage to get through and having it heal rapidly with no thought required.

I'd like to see either the automedical capacity removed (and transferred onto something else, like a webbing), OR the armor lowered, and this PR does one of these.

It's not like he's prevented from using pill bottles like the rest of us lowly ungas, or GOD FORBID waiting for the medic!

And don't say "There are no medics". "There are no medics" because the common unga doesn't need one due to the availability of meds. I play medic, and I'm a useful bonus, but that's all.

davidchan13:
B18 isn't for common ungas so I don't know why you're trying to balance around that shit. It's a very specialized armor for a very specialized role that is instantly recognizable and Xenos have numerous things they can do shut down a B18 boi. 

Having things like B18 is good because it shakes up the meta and gives Marines a reason to push out of their FOBs and not need an entire murder ball to break T3 xeno pushes and counter pushes. If the average unga gear was buffed so Xenos couldn't just instantly stun/infect/kill any marine that gets caught out of position, then yeah I'd say B18 could be buffed or removed. As it stands if B18 stops being viable marines will just go back to FOBForts and xenos will complain the marines never push until they have enough numbers on their side form a murderball.

Compare this to xeno play, where most xenos have the freedom to run off on their own thanks recent pheremone changes so that xenos aren't 'reliant' on recovery mones to recover their wounds, but marines get a limited option to have a bruiser that can jump in and out of combat with a fast recover time? Somehow that's unfair.

Rohesie:
This PR is being superseded by other nerfs. The bottom line is that we don't want to remove the feature. If the B18 is too strong, then make the armor protect less.

PR #3149, opened 2019-10-21
Delimbs now work off of thresholds instead of rng

ma44:
## About The Pull Request
Makes delimbing only happen once the limb goes above the max_damage threshold; threshold has been increased according

Upon delimbing, 25% of the max_damage is transferred to it's parent instead of 100% (hand => arm, arm => chest etc)

Bone breaking threshold is also increased to max_damage / 2 rounded up

## Why It's Good For The Game
Removes rng from delimbs and makes it threshold based, this stops people from complaining of getting their hand yeeted by some young runner (420) and instead has to occur from a drawn out fight or a significant pile on

## Changelog
:cl: ma44
balance: Delimbs now occur based on thresholds instead of rng, see below for details
balance: Delimbs occur when the limb has 75 or 100 brute damage occurred to it if it's a hand/foot or arm/leg respectively
balance: When a limb is delimbed, only 25% of it's damage is transferred to it's "parent" (hand to arm to chest for example)
balance: Bone breaks now only occur at 37 and 50 damage for hands/feet and arms/legs respectively
balance: FOR XENOS: This means if you want to kill a marine as fast as possible, you target their chest, otherwise you can target a limb to disable a limb but expect a longer fight due to a decreased damage transfer upon limb dismembering
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -9,7 +9,7 @@
 	var/damage_state = "00"
 	var/brute_dam = 0
 	var/burn_dam = 0
-	var/max_damage = 0
+	var/max_damage = 0 //The threshold before that limb gets destroyed
 	var/max_size = 0
 	var/last_dam = -1
 	var/supported = FALSE
@@ -213,10 +213,8 @@
 		owner.updatehealth()
 		return update_icon()
 	if(CONFIG_GET(flag/limbs_can_break) && brute_dam >= max_damage * CONFIG_GET(number/organ_health_multiplier))
-		var/cut_prob = brute/max_damage * 10
-		if(prob(cut_prob))
-			droplimb()
-			return
+		droplimb() //Reached max damage threshold through brute damage, that limb is going bye bye
+		return
 
 	owner.updatehealth()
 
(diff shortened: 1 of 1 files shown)


code/modules/organs/limbs.dm
```diff
@@ -662,8 +660,8 @@ Note that amputating the affected organ does in fact remove the infection from t
 		wounds.Cut()
 		if(parent && !amputation)
 			var/datum/wound/W
-			if(max_damage < 50) W = new/datum/wound/lost_limb/small(max_damage)
-			else 				W = new/datum/wound/lost_limb(max_damage)
+			if(max_damage < 50) W = new/datum/wound/lost_limb/small(max_damage / 4)
```
Rohesie: Proper indentation here, please.
Also use `* 0.25` instead of `/ 4`

Rohesie:
I'd increase the values to make delimbs less common.

jroinc:
Not sure how I feel about the increased bone-break threshold, given that a splinted broken bone does very little besides slightly increasing pain.

Rohesie:
Try to sum up your changes a bit more in the changelog.

jroinc:
Chest and groin now break at 60, up from 40, so the chest's the strongest part of the body still. Once the super-armored gloves/boots go away, what's going to keep hands/feet from being the meta again?

PR #3158, opened 2019-10-23
M41A And FC Tweaks

Max-023:
## About The Pull Request
Again, you saw this before. M41As and mags for them are added to the vendors. FC also gets smartgunner goggles and a medkit pouch so he can do his job more effectively.

## Why It's Good For The Game
M41As are currently way too limited and this will fix that. Additionally, the FC is an underplayed class with a very weak loadout. With no vendor, locker, or any other private req area on crash, they are under threat of just being dragged away and dying and in distress they have to make massive tweaks to their starting loadout to stay effective. The medkit pouch will make it so they can treat wounds in the field and keep their Marines as well as their own self up and running for longer. The smartgunner goggles means they should also no longer fear the dark and help their Marines push without the threat of a random hunter appearing out of nowhere and temporarily removing them from the round.

## Changelog
:cl:
add: M41A rifles and mags to the vendors, even in distress
add: M56 smartgunner goggles and a full medkit pouch to the FC, taking the glasses and left pocket slots respectively
del: apparently the entire shipside.dm, marine_vending.dm, and the tgmc.dme but they all look fine to me since it adds them all back in+my additions
/:cl:


Rohesie:
The PR has two concerns:
* Nightvision for the Field Commander is an excessive buff.
* The M41A have too much ammo for a standard marine weapon to be in vendors.

Rohesie:
Closing due to staleness.

PR #3162, opened 2019-10-24
Makes rack health sensible

Metroidlover:
## About The Pull Request

Makes Rack health 150 instead of whatever insane amount it was before hand(I actually couldn't find where it was set)

## Why It's Good For The Game

Hand portable structures that take ages to break are a bad idea, and people are already starting to catch onto this fact.

## Changelog
:cl: MetroidLover
balance: Racks no longer have insane health
/:cl:


Rohesie:
The object base has 500 max integrity, so that was the former value.

PR #3171, opened 2019-10-25
Swamp Map

heffy117:
Swamp Map.

## About The Pull Request

Adds a new map. Crashed theseus in a swamp. Large caves south and a small smugglers den north. WIP but playable. Use with the new smaller starting ship.

## Why It's Good For The Game

New map. More variation.


_maps/swamp.json
```diff
@@ -0,0 +1,8 @@
+{
+    "map_name": "swamp",
+    "map_path": "map_files/Swamp",
+    "map_file": "swamp.dmm",
+    "shuttles": {
+    },
```
spookydonut:
```suggestion
```

_maps/swamp.json
```diff
@@ -0,0 +1,8 @@
+{
+    "map_name": "swamp",
+    "map_path": "map_files/Swamp",
+    "map_file": "swamp.dmm",
+    "shuttles": {
```
spookydonut:
```suggestion
    "armor": "jungle",
```

code/game/turfs/turf.dm
```diff
@@ -315,8 +315,13 @@
 	src.ChangeTurf(/turf/open/space)
 	new /obj/structure/lattice( locate(src.x, src.y, src.z) )
 
+/turf/proc/ReplaceHoleWithLattice()
+	src.ChangeTurf(/turf/open/hole_lattice)
```
Rohesie: No need for the `src.` part on any of this.

Alterist:
Pretty interesting concept, but the map is way too large.

You should just have empty space in the northern cave tiles, as that helps out with lag a bit
You should remove most airlocks on the Theseus ship itself, otherwise it's very difficult for marines to move around in it.
you also need to remove more of the machinery on the ship itself, stuff like the alarm computers, cryopods, most of the overwatch computers, all the squad vendors, replace reinforced windows and chairs in briefing and the brig. Stuff like the autodoc and QM stuff should probably ve removed for balance reasons

If I was creating the map, i would just use either one of the floors and delete the ladders. put the cave building exposed off to one side. and leave just a bit of caves for the xenos and cut all the excess space.

PR #3190, opened 2019-10-27
newton element

spookydonut:
## About The Pull Request
newton's third law, shoot a gun and go flying

## Why It's Good For The Game
memes

## Changelog
:cl:
add: buckling yourself to a roller chair and shooting now moves you
/:cl:

### DIFF
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -192,6 +192,7 @@
 #define COMSIG_MOB_ITEM_AFTERATTACK "mob_item_afterattack"		//from base of obj/item/afterattack(): (atom/target, mob/user, proximity_flag, click_parameters)
 #define COMSIG_MOB_SAY "mob_say" 								// from /mob/living/say(): (proc args list)
 #define COMSIG_MOB_LOGOUT "mob_logout"							//from /mob/Logout(): ()
+#define COMSIG_MOB_EXTINGUISHER_USE "mob_extinguisher_user"		//from /obj/item/extinguish(): (target, force)
 
 //mob/living signals
 #define COMSIG_LIVING_DO_RESIST			"living_do_resist"		//from the base of /mob/living/do_resist()
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -161,7 +161,7 @@
 	#define COMPONENT_ITEM_CLICKCTRLON_INTERCEPTED (1<<0)				//from base of mob/CtrlClickOn(): (/atom, /mob)
 
 // /obj/item/weapon/gun signals
-#define COMSIG_GUN_FIRE "gun_fire"
+#define COMSIG_GUN_FIRE "gun_fire"								//:() (target, user)
 	#define COMPONENT_GUN_FIRED 1
 #define COMSIG_GUN_AUTOFIRE "gun_autofire"
 #define COMSIG_GUN_CLICKEMPTY "gun_clickempty"
--- a/code/datums/elements/newton.dm
+++ b/code/datums/elements/newton.dm
@@ -1,0 +1,40 @@
+/datum/element/newton/Attach(datum/target)
+	. = ..()
+	if(!ismob(target))
+		return ELEMENT_INCOMPATIBLE
+
+	RegisterSignal(target, COMSIG_MOB_EXTINGUISHER_USE, .proc/exinguish)
+
+/datum/element/newton/Detach(datum/target, force)
+	. = ..()
+	UnregisterSignal(target, COMSIG_MOB_EXTINGUISHER_USE)
+
+/datum/element/newton/proc/exinguish(datum/target, atom/target_atom, force)
+	var/mob/M = target
+
+	propel(target, get_dir(M, target_atom), force)
+
+/datum/element/newton/proc/propel(mob/M, direction, force)
+	if(!M.buckled || !isobj(M.buckled) || M.buckled.anchored)
+		Detach(M)
+
+	var/rep = CLAMP((4-force)*2, 0, 8)
+
+	move_chair(M.buckled, reverse_direction(direction), rep)
+
+/datum/element/newton/proc/move_chair(obj/B, movementdirection, repetition=0)
+	step(B, movementdirection)
+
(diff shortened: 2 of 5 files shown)


Kuro-ai:
SPEED MERGE PLS

code/game/objects/items.dm
```diff
@@ -747,42 +747,9 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 
 	playsound(user.loc, 'sound/effects/extinguish.ogg', 52, 1, 7)
 
-	var/direction = get_dir(user,target)
+	SEND_SIGNAL(user, COMSIG_MOB_EXTINGUISHER_USE, target, 4)
```
Rohesie: Put this before the sound and feedback, a couple of lines up.

Hughgent:
You mentioned in the Discord that it's 1 bullet per tile. If you could make that 3 bullets per tile, for most rifles bursts, I think that would be 'slightly' more realistic.

code/datums/elements/newton.dm
```diff
@@ -0,0 +1,45 @@
...
+/datum/element/newton/proc/propel(mob/M, direction, force)
+	if(!M.buckled || !isobj(M.buckled) || M.buckled.anchored)
+		Detach(M)
+
+	var/rep = CLAMP((4-force)*2, 0, 8)
+
+	move_chair(M.buckled, reverse_direction(direction), rep)
+
+/datum/element/newton/proc/move_chair(obj/B, movementdirection, repetition=0)
+	step(B, movementdirection)
```
Rohesie: Should use something like `relaymove()`

code/game/objects/items.dm
spookydonut: Same place as the movement code was before.

spookydonut:
> 
> 
> You mentioned in the Discord that it's 1 bullet per tile. If you could make that 3 bullets per tile, for most rifles bursts, I think that would be 'slightly' more realistic.

not without making it a lot more complicated

code/__DEFINES/components.dm
```diff
@@ -161,7 +161,7 @@
 	#define COMPONENT_ITEM_CLICKCTRLON_INTERCEPTED (1<<0)				//from base of mob/CtrlClickOn(): (/atom, /mob)
 
 // /obj/item/weapon/gun signals
-#define COMSIG_GUN_FIRE "gun_fire"
+#define COMSIG_GUN_FIRE "gun_fire"								//:() (target, user)
```
ninjanomnom: Do the actual comment format

code/game/objects/items.dm
ninjanomnom: This isnt movement code though, the signal should be as early in the proc as reasonable by default

code/datums/elements/newton.dm
```diff
@@ -0,0 +1,54 @@
...
+/datum/element/newton/proc/fire_gun(datum/source, atom/target, obj/item/weapon/gun/gun, mob/living/user)
+	propel(user, get_dir(user, target), 0)
+
+/datum/element/newton/proc/propel(mob/M, direction, force)
+	if(!M.buckled || !isobj(M.buckled) || M.buckled.anchored)
+		Detach(M)
+
+	var/rep = CLAMP((4-force)*2, 0, 8)
+
+	move_chair(M.buckled, reverse_direction(direction), rep)
```
ninjanomnom: This seems very strange to be adding to the mob whenever they sit on a chair, not only that you're assuming here that the chair exists. You should instead either make the element go on the chair and use a signal for when mobs buckle or make the element more generic.

If it *is* generic you probably want to make it support going on either side, and at that point you've recreated space drift.

code/datums/elements/newton.dm
```diff
@@ -0,0 +1,54 @@
...
+
+/datum/element/newton/proc/move_chair(obj/B, movementdirection, repetition=0)
+	step(B, movementdirection)
```
ninjanomnom: You want the walk() proc

code/datums/elements/newton.dm
```diff
@@ -0,0 +1,54 @@
...
+
+/datum/element/newton/proc/exinguish(datum/source, atom/target_atom, force)
+	var/mob/M = source
+
+	propel(M, get_dir(M, target_atom), force)
+
+/datum/element/newton/proc/throw_item(datum/source, atom/target)
+	var/mob/M = source
+
+	propel(M, get_dir(M, target), 0)
```
ninjanomnom: You dont need to typecast

code/datums/elements/newton.dm
```diff
@@ -0,0 +1,54 @@
...
+	UnregisterSignal(target, list(
+		COMSIG_MOB_EXTINGUISHER_USE, 
+		COMSIG_HUMAN_ITEM_THROW, 
+		COMSIG_HUMAN_GUN_FIRED, 
+		COMSIG_HUMAN_GUN_AUTOFIRED))
+
+/datum/element/newton/proc/exinguish(datum/source, atom/target_atom, force)
+	var/mob/M = source
+
+	propel(M, get_dir(M, target_atom), force)
```
ninjanomnom: ditto

code/game/objects/objs.dm
```diff
@@ -201,13 +203,15 @@
...
 /obj/proc/do_buckle(mob/M, mob/user, silent = FALSE)
+	SHOULD_CALL_PARENT(1)
 	if(!silent)
 		send_buckling_message(M, user)
 	M.buckled = src
 	M.loc = src.loc
 	M.setDir(dir)
 	M.update_canmove()
 	src.buckled_mob = M
+	M.AddElement(/datum/element/newton)
```
ninjanomnom: Not all bucklable objects should allow drift

code/datums/elements/newton.dm
spookydonut: I ripped this from tg which uses step

ninjanomnom: then they do it wrong

ninjanomnom: then they do it wrong

PR #3198, opened 2019-10-27
Game tab

Rohesie:
Moves everything but the ping from the "Stat" tab into a "Game" tab, and renames the former to "Status" like /tg/'s
This is because the first tab is the default one, and causes lag for players with a poor connection. They can still access the information but need to opt for it.


factoryman942:
if their connection is bad enough that six lines of text are lagging them, how are they even playing the game

PR #3220, opened 2019-11-01
Queen Acid Spit Removed

InterroLouis:
## About The Pull Request

Removes Queen Acid Spit. Her toolbox was too big and this was a very strong thing. MJP wants it gone but won't do their own PRs so here I am I guess. 

## Why It's Good For The Game

Less acid spam. Yay for less face melting. 

## Changelog
:cl:
balance: Queen loses acid spit
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/queen/castedatum_queen.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/queen/castedatum_queen.dm
@@ -37,7 +37,7 @@
 
 	// *** Ranged Attack *** //
 	spit_delay = 1.5 SECONDS
-	spit_types = list(/datum/ammo/xeno/sticky, /datum/ammo/xeno/acid/heavy)
+	spit_types = list(/datum/ammo/xeno/sticky)
 
 	// *** Pheromones *** //
 	aura_strength = 3 //The Queen's aura is strong and stays so, and gets devastating late game. Climbs by 1 to 5


InterroLouis:
Curse you Travis, I'll make you work. Or at least try. I've seen people do this before.

InterroLouis:
Now to try again, Travis.

Rohesie:
Yea, the bot is having hiccups, it can be ignored.

Isy232:
To be honest, the spit and melee combo of the queen is what makes the queen a very strong opponent. If she just plain looses acid spit with nothing else in return I can see the queen being extremely boring to play, you just press button to down all nearby marines and melee.

The queens combat toolbox consists of 4 things:
1: Melee
2: Scree
3: Spit
4: Healing abilities (currently near useless, they cost about all your plasma so if you heal 1 xeno you can barely emit pheromones anymore).
I don't think this is too much, the queen has about 20 special actions but not many are combat related.

Because of things such as the SADAR existing with this sort of change the only way for a queen to be of use would be to risk getting near instantly killed.

The spit + zoom combo belongs to something more like the boiler.

I'd say make a rework, give the queen something else to work with like better overwatch abilities, don't force the queen to be a pure tank caste.

I do also think that the pain damage from acid deserves to be nerfed too, but please come up with something so playing the queen isn't an exceptionally boring caste when it comes to combat.

davidchan13:
Queen is meant to be a support/leadership role and not a heavy combat role. If queen is on the front lines it should be life or death for the xenos.

Isy232:
Thats CM, here the queen leads the xenos into battle. Thats why ovi was removed and laying eggs was added.

Rohesie:
This pure nerf cannot go while the xenos are losing so much. It will require input from xeno players and probably a rework.

Rohesie:
Already being done on the Queen Rework.

PR #3227, opened 2019-11-02
Reworks Queen (Including removing acid spit)

Isy232:
## About The Pull Request

Well I think its stupid game design that the queen, the leader of the xenos, is also one of the strongest fighting castes. In the current meta its pretty much necessary for her to fight on the front lines because without her the xenos loose a large, tanky xeno.

I think a queen that stays back and supports the hive with ranged support abilites, rarely running out to screech to assist the xenos is a better design.

I would also like to note I did test this stuff locally, most xenos have health in the hundreds. These abilities will be strong but I highly doubt they will be worse than what the queen was previous to this.

Queen Looses Acid Spit
Queen Heal Plasma Cost: 600 => 150
Queen Heal Cooldown: 15s => 8s
Queen Heal Amount: 50 => 100
Queen Give Plasma Cost: 600 => 150
Queen Give Plasma Cooldown: 15s => 8s
Queen Slash Damage: 45/50/55/60 => 30/35/40/45
Queen Tackle Damage: 45/50/55/60 => 35/40/45/50
Queen Speed: 0.6/0.5/0.4/0,3 => -0.2/-0.3/-0.4/-0.5 
Queen Max Health: 300/325/350/375 => 200/225/250/275
Queen Plasma Max: 700/800/900/1000 => 900/1000/1100/1200
Queen Plasma Gain: 30/40/50/60 => 40/50/60/70
Queen Gains Larva Grow Sting

## Why It's Good For The Game

Queen is no longer a big tanky caste that is required to be on the front lines to be useful but rather can command from the back line while using her abilities to help our fellow xenos.

## Changelog
:cl:
balance: Reworks Queen
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
@@ -452,8 +452,8 @@
 	name = "Heal Xenomorph"
 	action_icon_state = "heal_xeno"
 	mechanics_text = "Heals a target Xenomorph"
-	plasma_cost = 600
-	cooldown_timer = 15 SECONDS
+	plasma_cost = 150
+	cooldown_timer = 8 SECONDS
 	keybind_signal = COMSIG_XENOABILITY_QUEEN_HEAL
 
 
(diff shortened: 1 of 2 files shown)


Rohesie:
The basic idea here is positive.
Can you replace the old queen overwatch-healing with a modified version of the Shrike's healing, for higher QoL?

Isy232:
Queen heal is able to affect any xeno in view while shrike has a range so I'd like to keep it to queen heal so you can be futher back. Queen heal is also instant, later I'll see about adding the sound effect of the heal though.

Rohesie:
You can do the Shrike heal without a minimum range, but clicking on the xeno instead.
A subtype of the action.

We want no Queen nerfs while the marines are still winning more.

Isy232:
Queen should still be top tier with this, she can blast off support powers quickly and run in quickly for a screech.

She also is still fairly tanky.

PR #3235, opened 2019-11-03
removes antidepressants, makes reagents do more things as multiples of defines

factoryman942:
med reagents now deal/heal more of their damage as multiples of REM (0.5), instead of just numbers
all custom_metabolisms are now multiples of REAGENTS_METABOLISM (0.2)
removes antidepressants because there's no escape from the depression of being a muhreen
:cl:
del: there are no more antidepressants
/:cl:


Rohesie:
This changes the resulting metabolism of several chems. Why?

PR #3243, opened 2019-11-04
Respawn and no repeated names

Rohesie:
Also improves the "Status" and "Game" tabs some, and adds a more visible respawn timer in them, for both xeno rejoining and marine respawns. And number of burrowed larva available if they exist.
## Changelog
:cl:
add: Marines are allowed to respawn after a certain time dead. We'll see how this goes.
fix: Marines can no longer join with the same name of someone else who joined the game and didn't cryo.
/:cl:


jroinc:
It looks to me like the respawn timer is 10 minutes, could you confirm if I read it right?

Rohesie:
> It looks to me like the respawn timer is 10 minutes, could you confirm if I read it right?

`GLOBAL_VAR_INIT(respawntime, 30 MINUTES)`

Rohesie:
Change request done.

PR #3248, opened 2019-11-05
HUD fix

Rohesie:
Fixes a bug that made eyewear HUD keep affecting the users even after they remove the glasses.

### DIFF
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -31,9 +31,9 @@
 
 /obj/item/clothing/glasses/proc/toggle_glasses(mob/user)
 	if(active)
-		deactivate_glasses(user, TRUE)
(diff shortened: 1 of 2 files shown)


code/modules/clothing/glasses/glasses.dm
```diff
@@ -26,28 +26,40 @@
...
+	if(ishuman(loc))
+		var/mob/living/carbon/human/H = loc
+		if(H.glasses == src)
+			H.update_tint()
+			H.update_sight()
+
+	update_action_button_icons()
+
+
+/obj/item/clothing/glasses/proc/activate_glasses(mob/user, feedback = FALSE)
```
psykzz: Feedback is typically referenced as `silent` elswhere, but i dont mind either way.

Rohesie: Change request done.

PR #3255, opened 2019-11-05
makes explosions not explode things 4 times

factoryman942:
:cl:
fix: explosions no longer hit things on their centre tile multiple times, and rockets no longer explode twice. Xenos should be pulped less by explosions now.
/:cl:
fixes the xeno squish
HE rockets should no longer do ~400 damage to queens instantly. They now do somewhere in the area of 100.


Rohesie:
The changelog entry can be reduced from 3 to 1, not to spam it.

PR #3261, opened 2019-11-06
SpecialClick Framework added, Vaulting and holecrawling changed to fit into it

terranaut1:
## About The Pull Request

Adds an extra proc simply named Special Click, by default bound to Ctrl, that overrides regular click procs. You can use it for any special item interactions you want without writing your own proc. I was initially going to make it easier to climb over barricades, tables, frames and through acid holes because I think the drag-dropping yourself on them is really fucking annoying and sometimes difficult especially in hectic situations or when the sprite is blocked. Adding this functionality is included. Things like grab-pulling can be reworked to call this and if I'm less lazy in the future maybe I'll put container opening (what happens when you alt-click a container on /tg/ and downstreams) to this or someone else can.

## Why It's Good For The Game

Click-dragging yourself is also clunky, slow and fails half the time especially if there's weeds grown over the frame because the weed sprite takes up most of it.
Framework for future stuff I guess.

## Changelog
:cl:Terranaut
add: Specialclick framework added
tweak: You can now specialclick (default: ctrl) to get over/through waist-high obstacles and through acid holes, or to rip them open.
/:cl:

### DIFF
--- a/code/datums/keybinding/carbon.dm
+++ b/code/datums/keybinding/carbon.dm
@@ -88,24 +88,24 @@
 	user.mob?.a_intent_change(INTENT_HARM)
 	return TRUE
 
-/datum/keybinding/mob/specialclick
+/datum/keybinding/carbon/specialclick
 	key = "Control"
 	name = "specialclick"
 	full_name = "Special Click"
 	description = "Hold this key and click to trigger special object interactions."
 
 
-/datum/keybinding/mob/specialclick/down(client/user)
+/datum/keybinding/carbon/specialclick/down(client/user)
 	RegisterSignal(user.mob, list(COMSIG_MOB_CLICKON), .proc/specialclicky)
-	RegisterSignal(user.mob, list(COMSIG_MOB_MOUSEDOWN, COMSIG_MOB_MOUSEUP), .proc/intercept_mouse_special)
+	RegisterSignal(user.mob, list(COMSIG_MOB_MOUSEDOWN, COMSIG_MOB_MOUSEUP), /datum/keybinding/mob/examine/proc/intercept_mouse_special)
 	return TRUE
 
 
-/datum/keybinding/mob/specialclick/up(client/user)
+/datum/keybinding/carbon/specialclick/up(client/user)
 	UnregisterSignal(user.mob, list(COMSIG_MOB_MOUSEDOWN, COMSIG_MOB_MOUSEUP, COMSIG_MOB_CLICKON))
 	return TRUE
 
-/datum/keybinding/mob/specialclick/proc/specialclicky(datum/source, atom/A, params)
+/datum/keybinding/carbon/specialclick/proc/specialclicky(datum/source, atom/A, params)
 	var/mob/living/carbon/user = source
 	if(!user.client || !(user.client.eye == user || user.client.eye == user.loc))
 		UnregisterSignal(user, (COMSIG_MOB_CLICKON))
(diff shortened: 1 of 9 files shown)


code/game/objects/structures.dm
```diff
@@ -41,6 +41,14 @@
 
 	do_climb(usr)
 
+/obj/structure/attack_alien(mob/living/carbon/xenomorph/user)
+	. = ..()
+	do_climb(usr)
+
+/obj/structure/attack_hand(mob/living/carbon/human/user)
+	. = ..()
+	do_climb(usr)
```
Rohesie: This should would benefit from a keybind instead.
Need to make modifications to the system to allow for this.
It also needs to use `user` instead of `usr`, and check intent.

code/game/atoms.dm
```diff
@@ -675,3 +675,8 @@ Proc for attack log creation, because really why not
 		return FALSE
 
 	return TRUE
+
+
+// For special click interactions (take first item out of container, quick-climb, etc.)
+/atom/proc/specialclick(mob/living/carbon/user)
+	return
```
```suggestion
	return

```
Always end the files with a blank line.

code/datums/keybinding/carbon.dm
```diff
@@ -65,3 +65,27 @@
 /datum/keybinding/carbon/select_harm_intent/down(client/user)
 	user.mob?.a_intent_change(INTENT_HARM)
 	return TRUE
+
+/datum/keybinding/mob/specialclick
+	key = "Control"
```
Rohesie: If `ctrl` is used, then the interactions with `CtrlClick` will need to be addressed.

terranaut1:
Ignore the commit gore this is a learning experience :) Thanks to tral and Roh for helping.

code/datums/keybinding/keybinding.dm
```diff
@@ -22,3 +22,6 @@
 
 /datum/keybinding/proc/up(client/user)
 	return FALSE
+
+/datum/keybinding/proc/intercept_mouse_special(datum/source)
+	return COMSIG_MOB_CLICK_CANCELED
```
Rohesie:
```suggestion
	return COMSIG_MOB_CLICK_CANCELED

```
Always end files in a new line.

code/game/atoms.dm
Rohesie: Still an issue. Please don't resolve things on your own, leave it to the reviewers.

code/game/objects/effects/acid_hole.dm
```diff
@@ -41,10 +41,12 @@
 		use_wall_hole(user)
 
 
-/obj/effect/acid_hole/attack_alien(mob/living/carbon/xenomorph/user)
+/obj/effect/acid_hole/specialclick(mob/living/carbon/xenomorph/user)
```
Rohesie: Do we have any guarantees the user will be a xeno and not a random carbon?

code/datums/keybinding/keybinding.dm
terranaut1: I comitted this shit 4 times with definitely trailing new lines idk why it keeps being fucky just kill me

code/game/objects/effects/acid_hole.dm
terranaut1: fixed and tested

Rohesie: To follow good practices, set it as what it can actually come as:
`/obj/effect/acid_hole/specialclick(mob/living/carbon/user)`
Then on the check, do an early return instead of a big nesting proc:
```
if(!isxeno(user))
	return
```

terranaut1: donezie

PR #3264, opened 2019-11-07
Makes the Machete destroy weed nodes in one hit.

Hughgent:
## About The Pull Request

Shitcode

## Why It's Good For The Game

Makes the machete a one hit weed node removal tool. as before it was a three hit tool.
When compared to the faster hitting knife (4 hits) it was the sub optimal choice for removing weed nodes.

## Changelog
:cl: Hughgent
tweak: Machete's now remove nodes in one hit.
/:cl:

### DIFF
--- a/code/game/objects/effects/weeds.dm
+++ b/code/game/objects/effects/weeds.dm
@@ -167,4 +167,9 @@
 				turfs_to_check += AdjT
 				node_turfs += AdjT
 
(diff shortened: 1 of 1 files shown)


Rohesie:
Please, no shitcode.
Give the node less health, increase the machete's damage and decrease its attack speed.

Fikou:
why not do the same thing fireaxes do? deal  structural damage when it hits with special weapon, this is the shittiest of codes

Hughgent:
#3273

PR #3282, opened 2019-11-09
AFK, Respawn and Game tab tweaks

Rohesie:
Nothing happened here.

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -315,7 +315,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 
 //AFK status
 #define MOB_CONNECTED 0
-#define MOB_AFK 1 //Still within the grace period.
+#define MOB_RECENTLY_DISCONNECTED 1 //Still within the grace period.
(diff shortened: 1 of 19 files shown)


code/__DEFINES/mobs.dm
```diff
@@ -313,6 +313,11 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 #define IS_MOTH (1<<7)
 //=================================================
 
+//AFK status
+#define MOB_CONNECTED 0
+#define MOB_AFK 1 //Still within the grace period.
```
psykzz: AFK isn't AFK here, its disconnected but within grace.

I would personally find another name as its confusing. To me AFK is still connected, but not moved for X time.
However from your implementation it looks like its disconnected within 5/15 mins.

Rohesie: Which name would you give it?

Rohesie: Changed it to `MOB_RECENTLY_DISCONNECTED`

psykzz: Sorry i didn't have a name at the time, `DISCONNECTING` would also work

PR #3287, opened 2019-11-09
unfuck shutter element

spookydonut:
## About The Pull Request
aaaaaaaaaaaaa what shitcoder wrote this

### DIFF
--- a/code/datums/elements/windowshutter.dm
+++ b/code/datums/elements/windowshutter.dm
@@ -3,20 +3,22 @@
 	if(!istype(target, /obj/structure/window))
 		return ELEMENT_INCOMPATIBLE
 
-	RegisterSignal(target, COMSIG_OBJ_DECONSTRUCT, .proc/spawn_shutter)
+	RegisterSignal(target, ELEMENT_CLOSE_SHUTTER_LINKED, .proc/spawn_shutter)
+	RegisterSignal(target, COMSIG_OBJ_DECONSTRUCT, .proc/spawn_shutter_first)
 
 /datum/element/windowshutter/Detach(datum/target, force)
 	. = ..()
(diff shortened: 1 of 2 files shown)


code/datums/elements/windowshutter.dm
```diff
@@ -3,20 +3,22 @@
...
 /datum/element/windowshutter/Detach(datum/target, force)
 	. = ..()
-	UnregisterSignal(target, COMSIG_OBJ_DECONSTRUCT)
+	UnregisterSignal(target, list(COMSIG_OBJ_DECONSTRUCT, ELEMENT_CLOSE_SHUTTER_LINKED))
 
-/datum/element/windowshutter/proc/spawn_shutter(datum/source, from_dir=0)
-	if(!from_dir) //air escaping sound effect for original window
-		playsound(src, 'sound/machines/hiss.ogg', 50, 1)
+/datum/element/windowshutter/proc/spawn_shutter_first(datum/source)
+	playsound(source, 'sound/machines/hiss.ogg', 50, 1)
```
Rohesie:
```suggestion
	UnregisterSignal(source, ELEMENT_CLOSE_SHUTTER_LINKED)
	playsound(source, 'sound/machines/hiss.ogg', 50, 1)
```
To avoid spawning two shutters on the first one?

spookydonut: no

spookydonut: deconstruct only triggers this proc, which plays the sound effect for the first shutter.
spawn_shutter sends the element signal to adjacent windows

before the element signal is sent, the window that is in the process of spawning the shutter unregisters from the element signal.
This will stop circular cascades.

PR #3289, opened 2019-11-09
Gun repaths

Rohesie:
A little prequel to the new guns.


spookydonut:
magazine path changes intended?
Also why

Rohesie:
Yes, they are intended.
We are going to replace the main guns with different ones, and they will accept different magazines.

This is in order to, when we replace the guns soon, we can preserve the old ones easily for events or map fluff. And in case we need to change the main guns soon, we don't need to replace a million things again.

Pariah919:
M41A magazines are pathed as "/obj/item/ammo_magazine/rifle/standard_lmg"

But HPR magazines are also pathed as "/obj/item/ammo_magazine/rifle/standard_lmg"

HPR doesn't seem to have a copy of itself in the code for future-keeping. Not sure if that's intended or not but just noting.

M41A uses the lmg pathing that it's magazines use, but the HPR also has this pathing. This should probably be changed aswell once you change the pathing if you do.

Rohesie:
Travis having hiccups.

PR #3301, opened 2019-11-10
Spitting anything (acid spit, slowdown spit, boiler spit) as a Xeno slows you down for a short time.

BurgerLUA:
## About The Pull Request

Spitting anything as a xeno will slow you down for about 2 seconds. The slow effect is based on the game's slow effect, and you will progressively increase in speed as the effect reduces.

## Why It's Good For The Game

Honestly one of the worst things about TGMC right now is the 2 hour rounds that are only 2 hours long because the last few xenos consist of a shrike and two spitters/sentinels. The marines proceed to chase these 3 xenos through the winding ice caves, almost killing one each time, but only to get healed or get away at 25% health because the xenos just spam their short cooldown spit abilities.

The second worst thing about TGMC is those lone sentinels/spitters who repeatedly hide behind a corner, spit at a turret, then go back behind that corner and/or run off into the distance to repeat the process a thousand times over. They don't bother to evolve to boiler because that means they're too weak and important to be alone, which means that they, god forbid, have to work with other xenos.

This is not good gameplay. It's getting extremely stale, especially on crash rounds. Even playing spitter/sentinel doing this, it just feels wrong. Hopefully this PR will be the first of many reworks/buffs/nerfs that prevent the sort of hit and run gameplay that should only be reserved to the hunter/runner/pred castes. 

## Changelog
:cl: BurgerBB
balance: Spitting as xeno slows you down for 2 seconds. The slow's magnitude is reduced when the duration is depleted.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -681,7 +681,7 @@
 
 	newspit.fire_at(A, X, null, X.ammo.max_range, X.ammo.shell_speed)
 
-	X.add_slowdown(4) //This can stack.
+	X.add_slowdown(2)
 
 	add_cooldown()
 
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -681,6 +681,8 @@
 
 	newspit.fire_at(A, X, null, X.ammo.max_range, X.ammo.shell_speed)
 
+	X.add_slowdown(4) //This can stack.
+
 	add_cooldown()
 
 	return succeed_activate()


BurgerLUA:
It's hard to really explain the slowdown mechanic but it doesn't feel like 4 seconds. Many people may think it's a hard 4 second slow at 50% but it's more like a 50% slow at second 1, 30% at second 2, and 10% at second 3 and 0% at second 4.

terranaut1:
Spitters rely excessively on hit and run tactics when harassing marines in a fortified position, when they're moving a convoy or just to slow them down as the hive retreats. This is an awful idea.

Kuro-ai:
Are you nerfing hit and run tactics on a antagonist that is all abou hit and run tactics?this is not a good solution to end round delaying

InterroLouis:
This is an unnecessary change which will send spitting to the shadow realm. I say nay.

Rohesie:
Don't you think that slowing down these fast xenos might be a better alternative?

BurgerLUA:
> Don't you think that slowing down these fast xenos might be a better alternative?

I mean, this is what the PR does.

jroinc:
Uh. Make it a 3 second slow and I can see it. 4 seems a little too long in combat.

BurgerLUA:
Made it 2.

tgstation-server:

PR #3303, opened 2019-11-10
removes shift spits from queen

Fikou:
## About The Pull Request

removes shift spits from queen because it only has 1 spit type now


InterroLouis:
On the one hand, this makes sense. On the other hand, we would have to readd it if we ever decide to give the Queen an additional spit option in the future.

Fikou:
> On the one hand, this makes sense. On the other hand, we would have to readd it if we ever decide to give the Queen an additional spit option in the future.

what kind of argument is this thats like scrapping an entire ability but keeping the button to use it

factoryman942:
yeah, it's literally just a 1 line change to readd it later if/when needed

Rohesie:
Yes, it should be re-added only if she gets more spits.

PR #3320, opened 2019-11-11
Removes Ice Cave Memes

BurgerLUA:
image

## About The Pull Request

Makes LZ1 larger. Makes the distance between the hospital and the refinery 5 tiles instead of 3. The entirety of the refinery is locked down until the shutters are raised. LZ2 is also slightly larger, and the main entrance to the caves is also larger, and outside, so mortars can work at that checkpoint.

## Why It's Good For The Game

I observed a round and I witnessed the xenos pull some pretty stupid shit due to bad map design. This should hopefully fix LZ1 camping by creating a larger area inaccessible to both xenos AND humans until the shutters are up.

## Changelog
:cl: BurgerBB
balance: Rebalanced Icy Caves do to xeno behavior. You know what you did.
/:cl:


BurgerLUA:
lmao it adds the removal tag automatically because I had "remove" in the title

TiviPlus:
Dunno how problematic it would be but it seems xenos can still ventcrawl into refinery

Isy232:
As both the original map creator and the crusher who probably was the reason this issue was identified I agree with this change. (If this even means anything)

Crushing down the middle of of the path between refinery and medbay, hitting 5 marines right as the shutters dropped was a bit too much of a meme.

PR #3322, opened 2019-11-11
Larval toxin changes

jroinc:
## About The Pull Request
Changes Larva growth

## Why It's Good For The Game

Larval growth is weird right now. If you aren't toxined, you won't burst. This aims to change that.

## Changelog
:cl:
balance: Larval toxin now advances larva less, but larva naturally grow more. Overall, there's slower growth. It no longer heals oxygen damage.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/embryo.dm
+++ b/code/modules/mob/living/carbon/xenomorph/embryo.dm
@@ -77,10 +77,10 @@
 	if(CHECK_BITFIELD(affected_mob.restrained_flags, RESTRAINED_XENO_NEST)) //Hosts who are nested in resin nests provide an ideal setting, larva grows faster.
 		counter += 1 + max(0, (0.03 * affected_mob.health)) //Up to +300% faster, depending on the health of the host.
 	else if(stage <= 4)
-		counter++
+		counter += 2 //Doubled growth overall, lowering free burst time to ~10 min.
 
 	if(affected_mob.reagents.get_reagent_amount(/datum/reagent/toxin/xeno_growthtoxin))
-		counter += 4 //Dramatically accelerates larval growth. You don't want this stuff in your body. Larva hits Stage 5 in just over 3 minutes, assuming the victim has growth toxin for the full duration.
+		counter += 2 //Accelerates larval growth somewhat. You don't want this stuff in your body. Larva hits Stage 5 in just over 3 minutes, assuming the victim has growth toxin for the full duration.
 
 	if(stage < 5 && counter >= 120)
 		counter = 0
--- a/code/modules/mob/living/carbon/xenomorph/embryo.dm
+++ b/code/modules/mob/living/carbon/xenomorph/embryo.dm
@@ -80,7 +80,7 @@
 		counter += 2 //Free burst time in ~10 min.
 
 	if(affected_mob.reagents.get_reagent_amount(/datum/reagent/toxin/xeno_growthtoxin))
-		counter += 2 //Accelerates larval growth somewhat. You don't want this stuff in your body. Larva hits Stage 5 in exactly 5 minutes, assuming the victim has growth toxin for the full duration.
+		counter += 2 //Accelerates larval growth somewhat. You don't want this stuff in your body. Larva hits Stage 5 in exactly 5 minutes (lower if healthy and nested), assuming the victim has growth toxin for the full duration.
 
 	if(stage < 5 && counter >= 120)
 		counter = 0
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/embryo.dm
```diff
@@ -77,10 +77,10 @@
 	if(CHECK_BITFIELD(affected_mob.restrained_flags, RESTRAINED_XENO_NEST)) //Hosts who are nested in resin nests provide an ideal setting, larva grows faster.
 		counter += 1 + max(0, (0.03 * affected_mob.health)) //Up to +300% faster, depending on the health of the host.
 	else if(stage <= 4)
-		counter++
+		counter += 2 //Doubled growth overall, lowering free burst time to ~10 min.
 
 	if(affected_mob.reagents.get_reagent_amount(/datum/reagent/toxin/xeno_growthtoxin))
-		counter += 4 //Dramatically accelerates larval growth. You don't want this stuff in your body. Larva hits Stage 5 in just over 3 minutes, assuming the victim has growth toxin for the full duration.
+		counter += 2 //Accelerates larval growth somewhat. You don't want this stuff in your body. Larva hits Stage 5 in just over 3 minutes, assuming the victim has growth toxin for the full duration.
```
Rohesie: `process_growth()` is called every 2 seconds, more or less.
Make the math and update the comment appropriately.

jroinc: Oh yeah, I did change the overall. I didn't change it cause I was considering keeping it the same, one sec.

jroinc: 5 min exact

Rohesie: Add that to the comment.

code/modules/mob/living/carbon/xenomorph/embryo.dm
```diff
@@ -77,10 +77,10 @@
 	if(CHECK_BITFIELD(affected_mob.restrained_flags, RESTRAINED_XENO_NEST)) //Hosts who are nested in resin nests provide an ideal setting, larva grows faster.
 		counter += 1 + max(0, (0.03 * affected_mob.health)) //Up to +300% faster, depending on the health of the host.
 	else if(stage <= 4)
-		counter++
+		counter += 2 //Doubled growth overall, lowering free burst time to ~10 min.
```
Rohesie: This comment is referring to your current change. Future coders checking this won't really care there was a change.
Keep it informative to the current situation. Something like this:
```suggestion
		counter += 2 //Free burst time in ~10 min.
```

code/modules/mob/living/carbon/xenomorph/embryo.dm
```diff
@@ -77,10 +77,10 @@
...
 
 	if(affected_mob.reagents.get_reagent_amount(/datum/reagent/toxin/xeno_growthtoxin))
-		counter += 4 //Dramatically accelerates larval growth. You don't want this stuff in your body. Larva hits Stage 5 in just over 3 minutes, assuming the victim has growth toxin for the full duration.
+		counter += 2 //Accelerates larval growth somewhat. You don't want this stuff in your body. Larva hits Stage 5 in exactly 5 minutes, assuming the victim has growth toxin for the full duration.
```
Rohesie:
```suggestion
		counter += 2 //Accelerates larval growth somewhat. You don't want this stuff in your body. Larva hits Stage 5 in exactly 5 minutes (lower if healthy and nested), assuming the victim has growth toxin for the full duration.
```

jroinc:
Played, got 4 marine and one survivor infects the whole round (lowpop).

The survivor got hugged and captured with zero damage, was toxin'd and nested instantly. He resisted out once and while he was fighting, he hit stage 4 (Then renested and burst.) If he were a marine, that would be just enough time to stasis-bag and be medevac'd for surgery. The autodoc might be able to save you, or it might not depending on if you were bagged/hypervened/medevac'd, treated promptly.

The marines that got hugged weren't captured or stung. Over the course of the battle, they went up in stage much faster, and actually had to evac via Alamo once they hit stage 3 after ~5 min.

One stayed behind to unga. He actually burst (the only marine burst that round).

Might be too early to tell, but I think it's working as intended.

Pariah919:
20->10 is too drastic for right now. Try 12-15.

jroinc:
> 20->10 is too drastic for right now. Try 12-15.

Seems fine to me. Medevac beds and stasis bags exist. And the rounds I've seen, you can still fight for about 5 minutes before being evac'd/stasis'd. 

If I were to make it 12-15 min, I'd make being infected have a slowdown per stage, and I don't think that would be the best change for the desired effect.

psykzz:
> infected have a slowdown per stage
Really like this idea

PR #3324, opened 2019-11-11
A major marine victory in crash only requires one authentication disk to be loaded into the nuke.

BurgerLUA:
## About The Pull Request

Changes the gamemode so only one disk, and of course the nuke, is needed to blow the planet up.
Marines are notified of their objectives on round start via announcement system.

## Why It's Good For The Game

According to statistics from the World Health Organization, the average marine intelligence is roughly around 10 (Margin of error: 20). Given the local wildlife on most crash scenarios, completing difficult tasks such as defending 3 computer terminals from an near infinite onslaught of xenos (one is literally born every minute) is almost impossible. This PR fixes that so marines don't have to go through a sea of bullshit (See: Spitters, Runners) to take even a single step outside the Canterbury Cream Egg.

## Changelog
:cl: BurgerBB
balance: Reworked crash so you only need one authentication disk to activate the nuclear device.
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -147,8 +147,8 @@
 	for(var/i in GLOB.xeno_resin_silo_turfs)
 		new /obj/structure/resin/silo(i)
 
-	for(var/i in GLOB.nuke_spawn_locs)
-		new /obj/machinery/nuclearbomb(i)
+	var/obj/machinery/nuclearbomb/NB = new(pick(GLOB.nuke_spawn_locs))
+	NB.require_one_auth = TRUE
 
 	for(var/i in GLOB.shuttle_controls_list)
 		var/obj/machinery/computer/shuttle/shuttle_control/computer_to_disable = i
@@ -179,9 +179,8 @@
 
 /datum/game_mode/crash/announce()
 	to_chat(world, "<span class='round_header'>The current map is - [SSmapping.configs[GROUND_MAP].map_name]!</span>")
-	priority_announce("Scheduled for landing in T-10 Minutes. Prepare for landing. Known hostiles near LZ. Detonation Protocol Active, planet disposable. Marines disposable.", type = ANNOUNCEMENT_PRIORITY)
+	priority_announce("Scheduled for landing in T-10 Minutes. Prepare for landing. Known hostiles near LZ. Detonation Protocol Active, planet disposable. Marines disposable.<br><br>Your objective is the following:<br>1.Locate and obtain a nuclear authentication disk from one of the three terminals located in this area.<br>2. Defend the nuclear authentication terminal as it generates an aunthentication generation program.<br>3. Locate and activate the plantary destruction device.<br>4. Defend the plantary destruction device as it detonates.<br>(Optional) 5. Flee the planet.", type = ANNOUNCEMENT_PRIORITY)
 	playsound(shuttle, 'sound/machines/warning-buzzer.ogg', 75, 0, 30)
-
 
 /datum/game_mode/crash/proc/add_larva()
 	var/datum/hive_status/normal/HS = GLOB.hive_datums[XENO_HIVE_NORMAL]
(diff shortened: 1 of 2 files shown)


Fikou:
if anything this should probably be a config

Kuro-ai:
A step towards make crash less cancer to play +1

Fikou:
> A step towards make crash less cancer to play +1

i would disagree, xenos wont be able to do any pre-preparation with hives when they dont know where marines will go, and if marines split up to all the disk places, you already lose, xenos arent that great even in crash, since theres 2-4 of them usually with about 10 marines with all specs taken
this just makes crash an easy mode for marines where they can split up and get disks in groups and then win because xenos wont be able to do shit

Kuro-ai:
> > A step towards make crash less cancer to play +1
> 
> i would disagree, xenos wont be able to do any pre-preparation with hives when they dont know where marines will go, and if marines split up to all the disk places, you already lose, xenos arent that great even in crash, since theres 2-4 of them usually with about 10 marines with all specs taken
> this just makes crash an easy mode for marines where they can split up and get disks in groups and then win because xenos wont be able to do shit

We can just reduce the disk generator to one,as it is right now marines can barely leave the ship,doing the objective is impossible

terranaut1:
>xenos lose one fight
>ungas have at least 3 minutes (120 respawn, about a minute to evolve from larva if it stays on weeds) to cade up around the computer
>print a disk for everyone
>gg

Crash has issues but this is not the way to fix them.
Suggestions that come to mind:
- At roundstart, one of the three computers is chosen at random. This is the one that will provide the nuclear authentication disk. Which one is immediately visible upon getting eyes on the computer (the other two will be smashed or whatever). This still forces the marines to, at worst, check all three locations and at best they still only have to win one fight, but it'll be on the xenos turf (similar to distress).
- The Canterbury Cream Egg is transporting highly technologically advanced equipment that it or allies bought, stole, recovered, plundered or whatever from somewhere and, sadly, crashes on the way back. At least this piece of tech was fueled when the marines got their grubby hands on it: A matter-compiler large enough to produce basic household items out of an internal material storage tank. Fuel is unavailable until the eggheads back home manage to reverse-engineer it, but the display is idiot proof, so a team of marines taken together should barely be smart enough to use it. Prints any item also available through a requisitions console. I've never played RO so I'm not going to suggest point balancing here, but the idea is giving the marines a finite, but flexible way to get new and powerful gear after your roundstart selection

Rohesie:
We'll need to increase the xeno numbers, but sure.

BurgerLUA:
No. All of these ideas are bad. Making the place where you get the Nuke Disk random would be shit considering some places are worse to defend than others, and if the xenos find out which one is the correct one, they would fortify the shit out of it. Giving marines the freedom of choice where to defend and reducing it to a single nuke disk would fix Crash being shit. It is absolutely and excessively rare that xenos are ever wiped out on crash, and if that is the case, then Marines deserve that victory considering how hard it is to do that as Xenos have serious evolution buffs in crash.

terranaut1:
Crash is intended as a lowpop gamemode and lowpop suffers from the fact that a Xeno, generally speaking, is stronger than a marine; It's just one player, but 5 marines vs 1 xeno is piss easy for the ungas while 5 vs 2 can go either way.
Making the outcome of the round hinge upon one fight (or one xeno loss, rather, since marines losing a fight means next to nothing if they can medivac the dead or bring them back to the egg in time) is pretty stupid if your goal is to make the gamemode less cancer.
If you want the aliens to be unable to fortify the one nuke disk computer in advance, then don't give the not-relevant computers a broken sprite but rather just make it obvious upon interacting with the computer once that they will not get the disk here. The marines will see this and move on. It allows for tactical choices too; do they split up, with 3 groups going for a computer each, or do they move together, slowly but surely? Generally speaking I'm a fan of player choice and think it's good game design to encourage choice. If you want to go a bit easier on the marines, then make it so Nuke Computer APCs cannot be broken from the roundstart randomness selection, so at worst the marines will need a screwdriver, wirecutter and a powercell when aliens slash them all.
Opening more specialist slots for latejoiners would probably also be a good idea. I've seen multi-spec on roundstart but right now I'm in a game lobby, checked the crew manifest, saw one spec, tried to join as one, no slots, so this seems like some kind of a bug that does make Crash noticeably harder for Marines.
I don't think the matter compiler is a bad idea. Giving marines more access to otherwise extremely limited items such as claymores or incendiary ammo would increase their combat strength in general and give them more flexibility to adapt to the situation once they've had a sortie or two and know what they're up against.

BurgerLUA:
I seriously think you're overeaxgerating the issues here. Go make your own PR if you feel this way because I am not going to code this.

Rohesie:
I'll remove the need of three disks when I have the time, making it just two.

terranaut1:
I'm not overexaggerating the issues, I'm saying your PR is bad and explaining why while offering alternatives since you want to make crash a bit easier for the marines.

GravitronBlue:
Since I particularly hate how crash plays out for marines, another alternate suggestion to throw in would be having all the disk generators work towards getting a single disk, with having multiple disk generators active speeding up the process.

BurgerLUA:
> Looks fine by me, while it may no longer be valid the original idea with the 3 disks were to give marines a reason to either be cocky and spread out for a quick win or stick together to go slow and steady.
> 
> My worry with the current implementation is the "meta" disk, that will always be choosen each round because its easiest for marines.

If there is ever a meta, message me so I can crush it. I've already repositioned the disks in another PR and I can do it again.

Rohesie:
If you want this PR to be tested, then increase the need for two disks instead of a single one.
Slow changes.

BurgerLUA:
> If you want this PR to be tested, then increase the need for two disks instead of a single one.
> Slow changes.

No. Unneeded.

Rohesie:
Understood.

PR #3327, opened 2019-11-11
Keybind tweaks

Rohesie:
* Backports /tg/ keybind tweaks.
* Makes all keybinds use signals.
* Makes `T`, `M`, `O` and `L` work directly without a keybind.
* Ports a /tg/ keybind and proc or two.
* Fixes goonchat permanence.


code/modules/client/preferences.dm
```diff
@@ -552,18 +553,43 @@ GLOBAL_LIST_EMPTY(preferences_datums)
...
 	onclose(user, "keybindings", src)
 
-
 /datum/preferences/proc/CaptureKeybinding(mob/user, datum/keybinding/kb, old_key)
 	var/HTML = {"
 	<div id='focus' style="outline: 0;" tabindex=0>Keybinding: [kb.full_name]<br>[kb.description]<br><br><b>Press any key to change<br>Press ESC to clear</b></div>
 	<script>
+	var deedDone = false;
 	document.onkeyup = function(e) {
+		if(deedDone){return}
```
psykzz:
```suggestion
		if(deedDone) { return; }
```

code/modules/keybindings/bindings_client.dm
```diff
@@ -55,20 +75,22 @@
...
 	// We don't do full key for release, because for mod keys you
 	// can hold different keys and releasing any should be handled by the key binding specifically
 	for (var/kb_name in prefs.key_bindings[_key])
 		var/datum/keybinding/kb = GLOB.keybindings_by_name[kb_name]
-		if (kb.up(src))
+		if(kb.up(src))
 			break
 
-	holder?.key_up(_key, src)
-	mob.focus?.key_up(_key, src)
```
psykzz: Are you sure this isn't needed?

code/controllers/subsystem/input.dm
```diff
@@ -7,29 +7,36 @@ SUBSYSTEM_DEF(input)
...
+	setup_default_macro_sets()
 
 	initialized = TRUE
 
 	refresh_client_macro_sets()
 
 	return ..()
 
+// This is for when macro sets are eventualy datumized
+/datum/controller/subsystem/input/proc/setup_default_macro_sets()
```
psykzz: Are you sure this is intended, this hard codes in the values for `O L T M`

(Comment was missed from the review, not sure what happened)

spookydonut:
very broken

Fikou:
pr seems to cause weird stuff like this
image
the movement keys dont work without rebinding them to wsad and even then you cant use arrow keys
also keybinds werent reset when this was tm'd so when you use T, O and other stuff it pops up the menu twice

sunofang:
It also causes two say/ooc windows to come up when the verb OOC or Say is used

code/modules/keybindings/bindings_client.dm
Rohesie: Yes, these are being replaced with keybinds.

code/controllers/subsystem/input.dm
Rohesie: For now it's how /tg/ is doing it.
Keybinds have some lag, and we want chat to be instant.

Rohesie:
Broken issue fixed.

Rohesie:
Travis having hiccups.

Rohesie:
This has been ported to /tg/ and is working fine.

PR #3329, opened 2019-11-11
Boiler rework

Hughgent:
## About The Pull Request

~~Just starting out~~
~~needs some 'interface work' to be usable. some way for the player to track how much ammo they have.~~
two ways of tracking.
1. an examine proc override for the boiler itself.
2. the 'make glob' button is itself an ammo counter.

image
image
image

## Why It's Good For The Game

changes the boiler to be more 'bursty' with it's bombarding. It'll be more constrained by plasma cost rather than ability cooldown.

## Changelog
:cl: Hughgent
tweak: The boiler now runs on an ammo system for it's bombard ability.
tweak: The boiler now glows according to the number of stored globs.
/:cl:


Pariah919:
neat

jroinc:
There was some talk about giving boilers defender-tier crest armor, thoughts on if that would be desirable?

code/modules/mob/living/carbon/xenomorph/castes/boiler/boiler.dm
```diff
@@ -14,6 +14,8 @@
 	upgrade = XENO_UPGRADE_ZERO
 	gib_chance = 100
 	drag_delay = 6 //pulling a big dead xeno is hard
+	var/corrosive_ammo = 0
```
Rohesie: Can you put these on the ability datum instead?

Hughgent: I don't want to limit the amount of each globule, I only want to limit the total amount.

Hughgent:
image
image

Hughgent:
@Rohesie How can I check a 'signal' to see if it's active? I think it's important to make the glob generating ability not work in 'bombard' mode.

code/modules/mob/living/carbon/xenomorph/castes/boiler/boiler.dm
Rohesie: Mhm. Still, the ability datum can track this.

Rohesie:
If the xeno has no `selected_ability` then it's not bombarding.

code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
```diff
@@ -53,14 +53,54 @@
...
+	action_icon_state = "toggle_bomb0" //to be changed
+	mechanics_text = "Creates a Boiler Bombard of the type currently selected."
+	plasma_cost = 200
+	use_state_flags = XACT_USE_BUSY
+	keybind_signal = COMSIG_XENOABILITY_CREATE_BOMB
+
+/datum/action/xeno_action/create_boiler_bomb/action_activate()
+	var/mob/living/carbon/xenomorph/boiler/X = owner
+
+	if(X.selected_ability)
```
Rohesie:
```suggestion
	if(X.selected_ability) //Bombarding or doing something else.
```

code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
```diff
@@ -53,14 +53,54 @@
...
+		to_chat(X, "<span class='notice'>We can not prepare globules as we are now.</span>")
```
Rohesie: This needs to be a little more informative.

code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
```diff
@@ -153,12 +198,19 @@
 	if(!istype(target))
 		return
 
-	to_chat(X, "<span class='xenonotice'>We begin building up acid.</span>")
+	if(X.ammo.type == /datum/ammo/xeno/boiler_gas/corrosive)
+		if(X.xeno_caste.corrosive_ammo <= 0)
+			to_chat(X, "<span class='warning'>We have no corrosive globules available.</span>")
+			return
+	else
```
Rohesie: There's a return, the `else` is not needed.

Rohesie: Actually, this shouldn't be in the caste.
See, castes are singletons. Only one of each type exists for all xenos. They store variables like this.
Put these variables in the action datum itself.

code/datums/actions/xeno_action.dm
```diff
@@ -10,6 +10,10 @@
 	var/keybind_signal
 	var/cooldown_id
 
+	//Boiler
+	var/corrosive_ammo = 0
```
Rohesie: Oh. You have multiple actions dealing with the same ability. This is bad and requires a component.
Guess I'll have to make one for you.

Hughgent: I'm afraid I don't follow the multiple actions part. There is only one ability that makes the ammo.

Rohesie: And there is another different ability that consumes it.

Hughgent: Which is why I felt it was important to have it stored on the mob, cause it makes sense to keep things there. A pouch that stores the globs if you will.

Rohesie: Yea, usually that'd be right. But the idea is to make actions generic, so that we can apply it to anything, other xeno castes or whatever.
That's why we'll be doing a component which stores this information.
The logic is that if we slap the component to a mob it adds these actions and manages them.

Rohesie:
Changes pending, also breaks plasma.

Hughgent:
I'm going to argue pretty heavily that it doesn't break plasma on it's own.

I'm also going to argue that it shouldn't be a component because of the unique sprites.

Hughgent:
Totally exonerated on the 'Breaks plasma' front. working perfectly fine in testing.

PR #3331, opened 2019-11-12
Shotgun spread tweaks

Rohesie:
Spread calculation fixed, lack of parentheses made it non-symmetrical.

## Changelog
:cl:
balance: Flechette and buckshot (both regular and SX) stats reduced, but instead of firing 3 they fire 5 projectiles. This means higher spread, shooting in a cone. The result is that flechette is nerfed in general and buckshot is buffed at close range.
/:cl:


InterroLouis:
This feels like too much spread-it makes shotguns basically useless at anything but PB.

Rohesie:
This has been tested quite a bit. It nerfs the shotguns in long-range, which is intended. They are still fearsome in short range. Could decrease the falloff if needed to make them interesting as spray damage.

PR #3336, opened 2019-11-12
Makes the Biogenerator just a prop item

psykzz:


jroinc:
-1
plz no remove any of the very few functional machines left without replacing it with something.

francinum:
This is expressly not functional.

TiviPlus:
Infinite larva exploit got snapped

Pariah919:
things like these should just be props tbf

Rohesie:
This machine is a security concern. Unsanitized against href exploits.

PR #3353, opened 2019-11-13
fixes boiler lights going out after an aheal

### LINKED ISSUE #3333: Ahealing a boiler makes its light go out
## Reproduction:
Aheal a boiler and watch as its light goes out

terranaut1:
## About The Pull Request

Fixes #3333 without interfering with #3329 

## Why It's Good For The Game

bugfix good

## Changelog
:cl:Terranaut
fix: boilers keep on shining bright after aheal
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -165,6 +165,10 @@
 	. = ..()
 	set_light(0) //Reset lighting
 
+/mob/living/carbon/xenomorph/caste/boiler/ExtinguishMob()
+	. = ..()
+	set_light(BOILER_LUMINOSITY)
+
 /mob/living/proc/update_fire()
 	return
 
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -165,7 +165,7 @@
 	. = ..()
 	set_light(0) //Reset lighting
 
-/mob/living/carbon/xenomorph/caste/boiler/ExtinguishMob()
+/mob/living/carbon/xenomorph/boiler/ExtinguishMob()
 	. = ..()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/living_defense.dm
```diff
@@ -165,6 +165,10 @@
 	. = ..()
 	set_light(0) //Reset lighting
 
+/mob/living/carbon/xenomorph/caste/boiler/ExtinguishMob()
```
psykzz: Might be better to update the original xeno proc to just check `isxenoboiler` to avoid destroying and recreating a light source.

Either way this fixes the issue.

terranaut1: I did that first but Roh didn't like it

Rohesie: PsyKzz brings a good point from an optimization perspective, but from a maintainability one I believe this is better.
A third option to take the best of both worlds would be either a variable that tracks the natural lighting, or even better a light-emitting component.
Those ideas are way out of scope and this fixes the issue. If you want to learn how to do them, poke me on discord. Else this is very approved.

Rohesie: This type path is incorrect.

terranaut1: goddamnit

terranaut1:
unga fix

spookydonut:
This is just a bandaid for how sloppily mobs on fire light code was converted to the new lighting code.

terranaut1:
I assumed there's a better way to fix this but at my current level of understanding I'm not gonna bother trying to do that without guidance. If someone wants to guide me I'm down but this is functional and not a ressource waster.

Rohesie:
This is fine and does the job.

PR #3358, opened 2019-11-14
New attack animations

Rohesie:
This is just an excuse to remove `flick_attack_overlay()`, which is a sleeping proc that can runtime if the target gets deleted during it.

## Changelog
:cl:
add: Most attacks get an animation now. Attacking with items shows the weapon used.
/:cl:


terranaut1:
I like this but please add transparency where applicable (scout/sniper thermal cloaks), it makes stealthing much more difficult if you have to smash your way into something because your weapons sprite stands out so much.

Rohesie:
I'd argue that you should lose your cloaking if you attack/bash things while stealthed.

terranaut1:
Why? You can run at full sprint and retain (some) transparency. Have it be semi-transparent and drain charge as if moving at worst, I'd say. I really like Scout spec for trying to stealth disks on crash and this makes it significantly harder.

PR #3362, opened 2019-11-14
gives hyposprays and its subtypes melee range

### LINKED ISSUE #3360: Autoinjectors can work from a distance
Title, autoinjectors can be used on those who aren't adjacent to you, allowing for remote treatment via overwatch cameras.

## Reproduction:

- Make or get autoinjectors from medical
- Hold the autoinjector in your active hand
- Use it on anyone you can see (Note that you can't inject xenos)

Particularly powerful against B18s using a couple custom made water and potassium autoinjectors.

terranaut1:
## About The Pull Request

fixes #3360

### DIFF
--- a/code/game/objects/items/reagent_containers/hypospray.dm
+++ b/code/game/objects/items/reagent_containers/hypospray.dm
@@ -49,7 +49,7 @@
 		return
 	if(!istype(user))
 		return
-	if(!in_range(A, user))
+	if(!in_range(A, user) || !user.Adjacent(A))
 		return
 	if(inject_mode == HYPOSPRAY_INJECT_MODE_DRAW) //if we're draining
 		if(reagents.holder_full())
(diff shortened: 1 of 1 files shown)


code/game/objects/items/reagent_containers/hypospray.dm
```diff
@@ -49,6 +49,8 @@
 		return
 	if(!istype(user))
 		return
+	if(!in_range(A, user))
```
Rohesie:
```suggestion
	if(!in_range(A, user) || !user.Adjacent(A))
```
First one checks if the item and the user are both on the same location relative to the attacked thing.
Second checks if the user is not just close to the thing, but if there are no barriers between them, such as a directional window or the likes.

Rohesie: There is a `do_after` in this proc, which signals a timed action. Could you repeat this check after it, as the target may have moved while it finishes?
`if(!do_after(user, SKILL_TASK_EASY, TRUE, A, BUSY_ICON_UNSKILLED) || same checks)`

terranaut1: done

PR #3369, opened 2019-11-15
Removes recoil from all weapons

BurgerLUA:
## About The Pull Request

Recoil doesn't exist anymore. Your screen will no longer shake like a shakeweight:tm: when firing any gun.

## Why It's Good For The Game

Recoil is shit in a 2D spessmans game. This isn't a third person shooter; it's a top down shooter where the recoil system makes absolutely no sense. Recoil shaking should be used in special circumstances only, such as getting hit with something, explosions, or taking damage; not firing a gun.

Also recoil has 0 effect on the performance/damage/accuracy of the gun: it's literally just a cosmetic effect.

## Changelog
:cl: BurgerBB
del: Removes recoil from the game, aka that shitty screenshake you get when you fire a weapon sometimes.
fix: Fixes any and all past, present, and future recoil bugs.
/:cl:

### DIFF
--- a/code/datums/components/full_auto_fire.dm
+++ b/code/datums/components/full_auto_fire.dm
@@ -359,7 +359,6 @@
 		projectile_to_fire.p_x = text2num(mouse_control["icon-x"])
 	if(mouse_control["icon-y"])
 		projectile_to_fire.p_y = text2num(mouse_control["icon-y"])
-	simulate_recoil(0 , shooter)
 	play_fire_sound(shooter)
 	var/firing_angle = get_angle_with_scatter(shooter, target, get_scatter(projectile_to_fire.scatter, shooter), projectile_to_fire.p_x, projectile_to_fire.p_y)
 	muzzle_flash(firing_angle, shooter)
--- a/code/modules/codex/entries/guns_codex.dm
+++ b/code/modules/codex/entries/guns_codex.dm
@@ -73,8 +73,6 @@
 		traits += "Damage modifier: [((damage_mult - 1) * 100) > 0 ? "+[(damage_mult - 1) * 100]" : "[(damage_mult - 1) * 100]"]%"
 	if(damage_falloff_mult)
 		traits += "Damage falloff: -[damage_falloff_mult] per tile travelled."
-	if(recoil)
-		traits += "Recoil: [recoil]"
 	if(scatter)
 		traits += "Scatter chance modifier: [scatter]%"
 	if(burst_scatter_mult)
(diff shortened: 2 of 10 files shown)


Pariah919:
yey

Rohesie:
Hello, no.
Recoil is good for allowing people to shoot a gun they shouldn't normally, with a consequence. Such as having no skill, or terrible modifiers. The alternative of not letting them shoot is worse.

InterroLouis:
And what are the guns they shouldn't normally shoot? Because any guns you shouldn't normally shoot are typically...specialist guns. Which do come with having no skill. Or smartguns, which have terrible modifiers for anyone that isn't a smartgunner. Does recoil actually do anything aside from a majorly annoying screenshake? It's not even a good consequence.

BurgerLUA:
Recoil is cosmetic, annoying, and affects nothing other than your viewport, and in that case, it fucking shakes that thing like a martini.

Pariah919:
The only gun that actually has recoil is the M44 and that's mostly annoying to most people.

Spec guns recoil isn't usually noticeable since you're usually firing a gun with modifiers so bad you won't hit shit anyway like the minigun that being unable to fire it is probably honestly better or they have such a high fire delay that the recoil isn't noticeable (RPG/sniper) 

The recoil itself is cosmetic and can be outdone anyway if the player knows about attachments.

spookydonut:
weird flex but ok

Teaika:
akimbo LMGs when

jroinc:
-1

It generally only punishes people with unwielded guns, or those without the skill to use them.

It's 100% obnoxious and annoying, but has a purpose.

Additionally, as a marine firing a wielded gun I'm meant to use, I have never had recoil, ever.

Rohesie:
If the gun is recoiling, then a message is being sent. You shouldn't be shooting it without penalties.

BurgerLUA:
> If the gun is recoiling, then a message is being sent. You shouldn't be shooting it without penalties.

Recoil is a shitty clientside effect that should only be used as a "gun-feel" tool to gauge how powerful the weapon is in a top-down shooter. In this game, the effects are VASTLY over exaggerated and cartoonish and serves no purpose other than to give motion sickness to the player, worst case scenario, and a minor annoyance best case scenario. I've fired a shotgun one handed before on a table and in no way did it make my entire world shake violently. Unless it is canon that weapons are fired pressed against your forehead, there is no way recoil would be this bad.

Rohesie:
There are several levels of recoil. The mechanic is to do a soft punishment for firing weapons in certain ways. If you want it changed, then suggest a different method, not just remove this one.

spookydonut:
stale closing

PR #3381, opened 2019-11-16
Defender Buff/Rework

Isy232:
## About The Pull Request

Removes Crest Defense ability from defender.
Buffs Defender's regular armor slightly.
Buffs Defender's speed slightly.
Buffs  Younger Upgrades of Defender damage slightly.
Ancient defender no longer gets the useless lunge and keeps its forward charge.
Forward Charge activation time lowered (So when you use the ability, it takes effect quicker).

## Why It's Good For The Game

From what I've seen, new players aiming to play a T1 don't have any simple solid picks.
Defender without these changes is just trash and playing wall is boring.
Runner is extremely squishy, as it should be, but because of that this caste is not good for new players.
Sentinel is mediocre, it has a very small health pool and requires you to play very safely to avoid death.
Drone is good for extreme noobs, but it wont get you any combat experience.

This aims to do 2 things:
1: Buff defender to the point where its a viable pick for a T1 xeno.
2: Make defender even simpler to play so new players to xeno have a solid caste to play where if they die their loss wont severely impact the hive.

If in testing this proves to be too good or not enough, @Isy232 on discord or just DM me and we can see about making changes to this.

## Changelog
:cl:
balance: Reworked And Buffed Defender
/:cl:


tgstation-server:

PR #3383, opened 2019-11-16
Removes ice colony from rotation

BurgerLUA:
## About The Pull Request

Removes Ice Colony v2 from map rotation.

## Why It's Good For The Game

I used ice colony as a base map, as it was a map created DIRECTLY for TGMC, as an example for Desert Outpost. Funny enough, literally all the criticisms of Desert Outpost apply to Ice Colony, but worse. Doesn't make sense that we don't have consistent mapping standards but here we are.

image
image
image

Ice Colony is considered a bad map on CM and it's considered a bad map on here. Ice Colony absolutely does not work on a coderbase where the average number of marines playing is 20. This map was clearly designed for 60 or so marines because of its massive size and massive amount of ground to cover. This map is absolutely not new player friendly and even though I've played several rounds of the map and looked at the map file several times, I still get lost moreso than any other map in rotation. Most maps clearly have a marine or xeno bias, but this map's bias is unironically against the player.

if Ice Colony was implemented today in its current state in a PR then there would be at least 100 comments of discussion, requested changes, feedback, and tweaks to get it working to a playable map.

## Changelog
:cl: BurgerBB
del: Removes Ice Colony v2 from rotation.
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -19,9 +19,6 @@ endmap
 map bigred_v2
 endmap
 
-map ice_colony_v2
-endmap
-
 map prison_station_fop
 endmap
 


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -2. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

jroinc:
+1.

It's too fucking big.

If I really wanted to as a lone shrike, I could make the round that never ends.

TiviPlus:
+1
It's a map thats slow and annoying to play on, and since we're getting a couple new ones we might as well get rid of the bad ones for now

Rohesie:
Configs are done server-based, but we could put this to vote.

Teaika:
Add a population lock, 50-60.

Rohesie:
Removed it in the server configs for now. No need for a code change here.

PR #3385, opened 2019-11-16
bugfix for the heavy helmet in storage tint

Hughgent:
## About The Pull Request

Apparently the heavy helmet was applying stats even in storage.

## Why It's Good For The Game

bugfix, but also apparently helmets were giving you head armor in backpacks?

## Changelog
:cl: Hughgent
fix: Heavy helmet no longer applies tint while in your backpack.
/:cl:

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -219,87 +219,92 @@
 
 /mob/living/carbon/human/proc/equip_if_possible(obj/item/W, slot, del_on_fail = 1) // since byond doesn't seem to have pointers, this seems like the best way to do this :/
 	//warning: icky code
-	var/equipped = 0
+	var/equipped = FALSE
+	var/in_storage = FALSE
 	switch(slot)
 		if(SLOT_BACK)
 			if(!src.back)
 				src.back = W
-				equipped = 1
+				equipped = TRUE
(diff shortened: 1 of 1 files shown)


Hughgent:
I do have to point out that helmets are also providing armor in the backpack, which by extension includes any item with stats that fits in storage inventory slots.

code/modules/mob/inventory.dm
```diff
@@ -219,87 +219,92 @@
...
 		if(SLOT_L_HAND)
 			if(!src.l_hand)
 				src.l_hand = W
-				equipped = 1
+				equipped = TRUE
 		if(SLOT_R_HAND)
 			if(!src.r_hand)
 				src.r_hand = W
-				equipped = 1
+				equipped = TRUE
```
Rohesie: I'm fixing this using flags, but this will cause problems, because holding stuff on the hands will trigger as equipped.

Hughgent: Should hands also be considered `in_storage`? Or are we planning on having hand held shields in the future?

spookydonut: > 
> 
> I'm fixing this using flags, but this will cause problems, because holding stuff on the hands will trigger as equipped.

literally see #3391

PR #3389, opened 2019-11-16
Acid damage no longer triggers fuel tank explosions

BurgerLUA:
## About The Pull Request

Acid damage no longer triggers fuel tank explosions.

## Why It's Good For The Game

Fuel tanks are triggered to explode if a brute dealing bullet or a burn dealing bullet hits it. This assume that both sources of the bullet have a high enough heat to combust the fueltank, which is true in 99% of cases. The code thinks that since xeno acid spit deals burn damage, it must be a laser or something hot.

Xenos have been exploiting this recently and bringing fuel tanks to the LZ to destroy barricades and kill marines before the shutters have even opened. It's hilarious but once you consider the situation it's some pretty awful bullshit. This effectively makes this behavior impossible until coders make some sort of flamethrower caste or something idk.

## Changelog
:cl: BurgerBB
fix: Acid dealing projectiles no longer trigger fueltank explosions.
/:cl:

### DIFF
--- a/code/game/objects/structures/reagent_dispensers.dm
+++ b/code/game/objects/structures/reagent_dispensers.dm
@@ -146,12 +146,13 @@
 
 
 /obj/structure/reagent_dispensers/fueltank/bullet_act(obj/item/projectile/Proj)
+
 	if(exploding)
 		return FALSE
 
 	. = ..()
 
-	if(Proj.damage > 10 && prob(60) && (Proj.ammo.damage_type in list(BRUTE, BURN)))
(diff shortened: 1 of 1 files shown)


BurgerLUA:
> give the tanks high acid resistance and make them use the integrity system.

acid shouldn't cause fuel tanks to explode in the first place, this is what the pr does

terranaut1:
You have no idea what kind of acid it is and if it causes an exothermic reaction with the welder fuel. It's easy to handwave an explosion happening. The problem is the explosion going through shutters, not spitters being able to lay traps or targetting a fuel tank in a firefight.

spookydonut:
Needs to be done properly via armor and integrity

PR #3391, opened 2019-11-17
clothing_tint component

spookydonut:
## About The Pull Request
aaaaaaaaaaa i dont know how I didn't notice how badly tinting was done.

Closes #3385

### DIFF
--- a/code/datums/components/clothing_tint.dm
+++ b/code/datums/components/clothing_tint.dm
@@ -11,10 +11,6 @@
 		src.tint = tint
 	if(!isnull(tint_state))
 		src.tint_state = tint_state
-	
-	//sanity checking
-	if(tint == TINT_NONE)
-		qdel(src)
 
 /datum/component/clothing_tint/Destroy(force, silent)
 	remove_tint()
--- a/code/datums/components/clothing_tint.dm
+++ b/code/datums/components/clothing_tint.dm
@@ -1,0 +1,61 @@
+/datum/component/clothing_tint
+	var/tint = TINT_NONE
+	var/tint_state = TRUE
+	var/mob/living/tinted_mob
+
+/datum/component/clothing_tint/Initialize(tint, tint_state)
+	. = ..()
+	if(!isitem(parent))
+		return COMPONENT_INCOMPATIBLE
+	if(!isnull(tint))
+		src.tint = tint
+	if(!isnull(tint_state))
+		src.tint_state = tint_state
+	
+	//sanity checking
+	if(tint == TINT_NONE)
+		qdel(src)
+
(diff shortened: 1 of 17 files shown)


code/game/objects/items.dm
```diff
@@ -279,7 +279,12 @@
 // for items that can be placed in multiple slots
 // note this isn't called during the initial dressing of a player
 /obj/item/proc/equipped(mob/user, slot)
+	SHOULD_CALL_PARENT(TRUE) // no exceptions
 	SEND_SIGNAL(src, COMSIG_ITEM_EQUIPPED, user, slot)
+	if(flags_equip_slot & slotdefine2slotbit(slot)) // flags_equip_slot is a bitfield
+		SEND_SIGNAL(src, COMSIG_ITEM_EQUIPPED_TO_SLOT, user)
+	else
+		SEND_SIGNAL(src, COMSIG_ITEM_EQUIPPED_NOT_IN_SLOT, user, slot)
```
Rohesie: Do a single signal and make it pass if equipped or not as a variable.

spookydonut: That would completely defeat the purpose of these signals

Rohesie: No, you can still register to the signal and check for the var.

spookydonut: That's what you do with the current signal so you're asking me to just make a duplicate signal

ninjanomnom: normally I wouldn't be too much a fan of that kind of thing but there's nothing inherently wrong with it and in that case in particular that check is done in a lot of component code. In this context this is a prety nice pair of signals.

code/datums/components/clothing_tint.dm
```diff
@@ -0,0 +1,59 @@
...
+		return COMPONENT_INCOMPATIBLE
+	if(!isnull(tint))
+		src.tint = tint
+	if(!isnull(tint_state))
+		src.tint_state = tint_state
+
+/datum/component/clothing_tint/Destroy(force, silent)
+	remove_tint()
+	return ..()
+
```
ninjanomnom: You need to clean up `tinted_mob`

code/modules/codex/entries/clothing_codex.dm
```diff
@@ -79,7 +79,7 @@ GLOBAL_LIST_INIT(string_equip_flags, list("suit slot" = ITEM_SLOT_OCLOTHING,
 	if(!slowdown)
 		armor_strings += "<br>It will not slow the wearer down."
 
-	if(tint)
+	if(GetComponent(/datum/component/clothing_tint))
```
ninjanomnom: get_mechanics_info signal

PR #3411, opened 2019-11-19
Grenade Cost Change

InterroLouis:
## About The Pull Request

This alters the cost of grenades orderable via requisitions.
M15 Box of 15 goes from 100 to 60, M15 crate of 5 goes from 40 to 20
HEDP Box of 25 goes from 100 to 40, crate of 5 goes to 10
HIDP Box of 25 goes from 100 to 40, crate of 5 goes to 10

## Why It's Good For The Game

Makes resupplying B17 less of a pain for everyone involved. 

## Changelog
:cl:
balance: Requisitions Grenade Cost Reduced
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -513,7 +513,7 @@ WEAPONS
 					/obj/item/explosive/grenade/frag/m15,
 					/obj/item/explosive/grenade/frag/m15
 					)
-	cost = RO_PRICE_NORMAL
+	cost = RO_PRICE_VERY_CHEAP
 	containertype = /obj/structure/closet/crate/explosives
 	containername = "\improper explosive M15 grenades crate (WARNING)"
 	group = "Weapons"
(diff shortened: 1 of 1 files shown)


Rohesie:
I'm very much not sure about this change.

TiviPlus:
I really do not see this as a good change, as grenades are already nearly unlimited

Roundstart a B17 has Access to:
-1 M15 Box
  -2 Spare M15 nades
-2 HIDP boxes
  -2 spare nades in req
  -8 Nades in the B17 Belt
-4 HEDP boxes
  -8 Nades in the B17 Belt

This amounts to 17 M15's, 60 HIDPs and 108 HEDPs meaning a B17 has 125 Explosive, stunning shots and 50 Fire grenades which is more than enough for even an extended fight. Of course, some marines take these nades but this is still a massive amount

Rohesie:
Yep, I'm not seeing as interesting either. There needs to be a limit for nade supply.

InterroLouis:
Yeah, they get ammo at the start. So does every other specialist. But every other specialist can also resupply for less than the cost of an entire new specialist gun, from req. It's cancerously high cost to have the boxes at 100 points. I even made the M15s more expensive than the HEDPs because they're better.

Rohesie:
This buff is considered excessive. Grenades are expensive, that much is true, but they are also much more devastating. We don't want them spammed.

PR #3413, opened 2019-11-19
Grenade launcher and similar shoot through MMB or shift+click, instead of ctrl+click

Rohesie:
This is not good code, but I need to do a big gun refactor to do it like I want to, properly.
People were complaining a lot about ctrl+click, so hopefully this is better.

## Changelog
:cl:
tweak: UGLs and similar are now shot once active via middle mouse button or shift click, like xeno abilities, instead of ctrl+click. You can also load and unload rifle magazines while it's active. You can find the option to toggle this in the IC tab instead of the Alien one.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -465,7 +465,7 @@
 
 /mob/living/carbon/verb/middle_mousetoggle()
 	set name = "Toggle Middle/Shift Clicking"
-	set desc = "Toggles between using middle mouse click and shift click for selected abilitiy use."
+	set desc = "Toggles between using middle mouse click and shift click for selected ability use."
 	set category = "IC"
 
 	middle_mouse_toggle = !middle_mouse_toggle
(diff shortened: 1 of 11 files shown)


Rohesie:
I was the one adding the previous code. It's hacky, now that we have a nice keybind system.
When I refactor click and guns we can improve this, though.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -461,3 +461,15 @@
 
 	to_chat(src, "<span class='xenoannounce'>We are an old xenomorph re-awakened from slumber!</span>")
 	playsound_local(get_turf(src), 'sound/effects/xeno_newlarva.ogg')
+
+
+/mob/living/carbon/verb/middle_mousetoggle()
+	set name = "Toggle Middle/Shift Clicking"
+	set desc = "Toggles between using middle mouse click and shift click for selected abilitiy use."
```
factoryman942:
```suggestion
	set desc = "Toggles between using middle mouse click and shift click for selected ability use."
```

PR #3414, opened 2019-11-19
More darkmode tweaks

Rohesie:
The squads should look a bit better, and command is rather distinct.

### DIFF
--- a/code/modules/goonchat/browserOutput.css
+++ b/code/modules/goonchat/browserOutput.css
@@ -226,10 +226,10 @@ a.popt {text-decoration: none;}
 .colorooc				{color: #3257c6;	font-weight: bold;}
 .looc					{color: #84add6;	font-weight: bold;}
 .hostooc				{color: #323232;	font-weight: bold;}
-.projleadooc			{color: #993299;	font-weight: bold;}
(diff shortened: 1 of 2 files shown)


code/modules/goonchat/browserOutput.css
```diff
@@ -322,7 +321,7 @@ h1.alert, h2.alert		{color: #99aab5;}
 .round_setup			{color: #db0000;		font-family: impact; font-size: 1.25em;}
 .round_header			{color: #db0000; 	display: block; text-align: center; font-family: courier; font-weight: bold; font-size: 2em;}
 .round_body				{color: #6685f5; 	display: block; text-align: center; font-family: trebuchet-ms; font-weight: bold; font-size: 1.5em;}
-.event_announcement		{color: #600d48; 	font-family: arial-narrow; font-weight: bold; font-size: 1.5em;}
```
psykzz: Just to be clear its intentional to remove the `font-weight`?
I'd suspect you want that for an announcement unless its wrapped elsewhere.

Rohesie: It is. This wasn't used anywhere, and the place where it could fit already used bold for the start of it.
The body of the announcement wasn't bold and doesn't need to be with this.

PR #3415, opened 2019-11-19
Dripping component processes instead of timer-loops

Rohesie:
ninjanomnom was displeased


code/datums/components/dripping.dm
```diff
@@ -47,8 +47,17 @@
 		qdel(src)
 
 
-/datum/component/dripping/proc/timed_drip()
+/datum/component/dripping/process()
 	var/mob/living/dripper = parent
+	if(QDELETED(dripper))
+		qdel(src)
+		return
```
nemvar: This doesn't look like a needed sanity check. Components are fairly decent at cleaning themselves up. If this case ever happens, this SHOULD runtime.

PR #3419, opened 2019-11-19
Fixes Lasgun Charger double stocking.

### LINKED ISSUE #3417: Infinite high cap laser ammo
## Testmerges:

#3415 #3413 #3404 #3405 #3406 #3386 #3327 #3335 #3399 #3409 #3407 #3414

## Reproduction:

Just take a high cap laser ammo out of the charger and put it back, it will count as if you had placed 2 of them in there for some reason, giving you infinite high caps

Hughgent:
## About The Pull Request

Fixes #3417 
Unfortunately, `. = ..()` was calling `stock()` the first time and the override was calling it a second time (with special recharge parameters). Both were working just fine, which was the issue.

## Why It's Good For The Game

no more infinite lasers.

## Changelog
:cl: Hughgent
fix: You can no longer get infinite lasgun cells from the lasgun charger.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -288,7 +288,14 @@
 
 
 /obj/machinery/vending/lasgun/MouseDrop_T(atom/movable/A, mob/user)
-	. = ..()
+	if(machine_stat & (BROKEN|NOPOWER))
+		return
+
+	if(user.stat || user.restrained() || user.lying)
+		return
+
+	if(get_dist(user, src) > 1 || get_dist(src, A) > 1)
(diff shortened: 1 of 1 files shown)


Rohesie:
This is a very snowflake construction.
No way around it, so oh well, but ideally it would call parent.

terranaut1:
Isn't it possible to handle the recharging in the parent via checking for a var or an isrechargable() helper and give that to lasgun magazines, powerpacks and such?

Rohesie:
It would be better than the current situation, but it would require a little refactor.
If you are willing to do so, it'd be positive.
A flag might be better than a helper proc, though.

PR #3425, opened 2019-11-21
Xeno elusive break sound bug fix

### LINKED ISSUE #2763: Purple sac doesn't always emit sound when hit
## Reproduction:
I took out a knife, N5 Night Raider something along hte lines of this, sniped those purple sacs with it and to my surprise, sometime there was no sounds at all, sometimes only couple hits emitted a sound.

Rohesie:
Fixes #2763
Looks like one of the sounds was broken.


Ghommie:
Amazing. And I gived up on trying to fix it ages ago just to find out it the issue wasn't within playsound.

PR #3429, opened 2019-11-21
Removes the broken Defender Lunge

Rohesie:
Someone will have to test better to re-add something like this.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defender/castedatum_defender.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defender/castedatum_defender.dm
@@ -152,12 +152,3 @@
 	// *** Defender Abilities *** //
 	crest_defense_armor = 25
 	fortify_armor = 50
-
-	actions = list(
-		/datum/action/xeno_action/xeno_resting,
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/defender/castedatum_defender.dm
```diff
@@ -152,12 +152,3 @@
...
 	fortify_armor = 50
-
-	actions = list(
-		/datum/action/xeno_action/xeno_resting,
-		/datum/action/xeno_action/regurgitate,
-		/datum/action/xeno_action/toggle_crest_defense,
-		/datum/action/xeno_action/fortify,
-		/datum/action/xeno_action/activable/lunge,
-		/datum/action/xeno_action/activable/tail_sweep
-		)
```
psykzz: This is just removing everything not just lunge.

Rohesie: The actions are defined at the first stage of maturity.
This list was just redefining the actions.

PR #3430, opened 2019-11-21
Fire and Snow interactions

psykzz:
## About The Pull Request
Fire and Snow interactions

## Changelog
:cl:
add: you can now instantly put out fires by resisting on snow
add: fire melts snow
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/particle_effects.dm
+++ b/code/game/objects/effects/effect_system/particle_effects.dm
@@ -36,6 +36,8 @@
 		E.fire_act()
 	for(var/obj/structure/bed/nest/N in loc)//Nests
 		N.fire_act()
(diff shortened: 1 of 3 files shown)


Pariah919:
This doesn't really make much sense if it's a flamethrower direct hit-
Napalm doesn't really get off things that easily. Same with LZ river interaction.

psykzz:
This is one of those times i'd go for gameplay over realism as you just roll on the floor to deal with it why not make the snow  more effective?

Pariah919:
This makes flamethrowers even worse than they already are though against beans on maps like ice and icy caves. I disagree with this change.

Rohesie:
I like the concept.

What does this make flamethrower bad at?

PR #3436, opened 2019-11-22
Buffs the Range Finder

BurgerLUA:
## About The Pull Request

The tactical binoculars now has an internal GPS that can be read by examining it.
The tactical binocular acquisition time is now reduced by a second.
The tactical binoculars can now work indoors.

## Why It's Good For The Game

Tactical binoculars were shit. This makes them not shit by buffing how easy it is to get coords now because there is nothing stopping another person from getting the cords another way, such as suit sensors or literally opening the map in the map viewer/editor or preemptively writing down good mortar coordinates.

## Changelog
:cl:
balance: The tactical binoculars now has an internal GPS that can be read via examine, it can now work indoors, and it has a faster acquisition time.
/:cl:

### DIFF
--- a/code/game/objects/items/binoculars.dm
+++ b/code/game/objects/items/binoculars.dm
@@ -147,7 +147,6 @@
 
 
 	var/turf/TU = get_turf(A)
-	var/area/targ_area = get_area(A)
 	if(!istype(TU))
 		return
 	if(user.action_busy)
--- a/code/game/objects/items/binoculars.dm
+++ b/code/game/objects/items/binoculars.dm
@@ -33,8 +33,7 @@
 
 /obj/item/binoculars/tactical/examine()
 	..()
-	to_chat(usr, "<span class='notice'>They are currently set to [mode ? "range finder" : "CAS marking"] mode.</span>")
-	to_chat(usr, "<span class='notice'>The internal GPS reads: [src.x][src.y].</span>")
+	to_chat(usr, "<span class='notice'>They are currently set to [mode ? "range finder" : "CAS marking"] mode.<br>The internal GPS reads: [src.x][src.y].</span>")
 
 /obj/item/binoculars/tactical/Destroy()
 	if(laser)
--- a/code/game/objects/items/binoculars.dm
+++ b/code/game/objects/items/binoculars.dm
@@ -135,7 +135,7 @@
 
 	var/acquisition_time = target_acquisition_delay
 	if(user.mind.cm_skills)
-		acquisition_time = max(5, acquisition_time - 15*user.mind.cm_skills.leadership)
+		acquisition_time = max(0.5 SECONDS, acquisition_time - 1.5 SECONDS * user.mind.cm_skills.leadership)
 
 	var/datum/squad/S = user.assigned_squad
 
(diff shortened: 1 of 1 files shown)


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -4. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

code/game/objects/items/binoculars.dm
```diff
@@ -34,6 +34,7 @@
 /obj/item/binoculars/tactical/examine()
 	..()
 	to_chat(usr, "<span class='notice'>They are currently set to [mode ? "range finder" : "CAS marking"] mode.</span>")
+	to_chat(usr, "<span class='notice'>The internal GPS reads: [src.x][src.y].</span>")
```
Rohesie: Put it on the same `to_chat` as the previous (use `<br>` if needed), and remove the `src` references.

code/game/objects/items/binoculars.dm
```diff
@@ -134,7 +135,7 @@
 
 	var/acquisition_time = target_acquisition_delay
 	if(user.mind.cm_skills)
-		acquisition_time = max(15, acquisition_time - 25*user.mind.cm_skills.leadership)
+		acquisition_time = max(5, acquisition_time - 15*user.mind.cm_skills.leadership)
```
Rohesie:
```suggestion
		acquisition_time = max(0.5 SECONDS, acquisition_time - 1.5 SECONDS * user.mind.cm_skills.leadership)
```

code/game/objects/items/binoculars.dm
```diff
@@ -146,19 +147,8 @@
 
 
 	var/turf/TU = get_turf(A)
-	var/area/targ_area = get_area(A)
 	if(!istype(TU))
 		return
-	var/is_outside = FALSE
```
Rohesie: I'm in favor of expanding the places where lazes can happen, but not completely removing them.
There's deep caves in where we might still want to preserve the functionality.
Change the logic to a positive one, in where this becomes false if the ceiling is deep caves.

BurgerLUA: It doesn't make sense for rangefinders not to work indoors. Rangefinders IRL don't calculate distance using some magical laser-finding satellite, but rather using their internal GPS as well as the time it takes for a signal to bounce back and forth between the designated target and the binoculars.

https://precisionrifleblog.com/2013/10/29/how-do-rangefinders-work/

Rohesie: There's the rangefinder for coordinates for the mortar (which we don't use beyond that, because coordinates in this game is a pretty bad idea), and there's the laser pointer. They follow different logics, which are not entirely realistic.

The idea is to restrict some areas to be free from OBs and CAS, as safe zones. The problem is that 90% of the maps qualify for that, and the intention is to reduce that significantly.

That is why I ask for some kind of area not to be laze-able.

BurgerLUA: CAS and OB are rarely used and when they are, the "attack" is super telegraphed. 

The mortar also has build in code to prevent any area that is deep underground from being targeted. As I said before anyone could just open the map in a map editor and get the coords themselves or just do basic estimation of the coordinate. For example, I always laze LZ1 and then subtract -30 from the value to get the center.

Rohesie: Yes, this would be fine for the rangefinder mode to get coordinates (even thought that's something we want to remove in the long-term). The issue is with it generating lazes that can be interacted by CAS and OBs.

PR #3448, opened 2019-11-23
Custom squad amounts

Rohesie:
## Changelog
:cl:
tweak: The Pillar of Spring is changed to support two squads instead of four.
/:cl:


code/datums/gamemodes/game_mode.dm
```diff
@@ -860,3 +868,86 @@ Sensors indicate [numXenosShip ? "[numXenosShip]" : "no"] unknown lifeform signa
...
+	sortTim(preferred_squads, cmp=/proc/cmp_numeric_dsc, associative = TRUE)
+
+	preferred_squads.len = max_squad_num
+	for(var/s in preferred_squads) //Back from weight to type.
+		preferred_squads[s] = SSjob.squads[s]
+	SSjob.active_squads = preferred_squads.Copy()
+
+	return TRUE
+
+/datum/game_mode/proc/initialize_scales()
```
psykzz: Since this looks to be completely an SSjob proc, we could just move it there instead, and ensure its called at `pre_start()`?

Rohesie: The issue is that it is redefined depending on the game mode.

Rohesie: Turned the relevant part into an SSjob proc.

PR #3455, opened 2019-11-24
DMR minor nerf

Rohesie:
## Changelog
:cl:
balance: The DMR can no longer point-blank. Its fire delay has been slightly increased.
/:cl:


InterroLouis:
That stings. It really does.

PR #3466, opened 2019-11-25
Stale automarking

spookydonut:
## About The Pull Request
Autolabel prs with no activity after 7 days as stale.


psykzz:
can you also add `needs-review`, that resets after a push? bunch of PRs go stale because its waiting for something or another but if its a review we shouldn't be faulting the dev.

PR #3467, opened 2019-11-25
Welderpacks can be opened while worn

XElectricX:
## About The Pull Request

Allows for engineers/ship techs to open welderpacks when they are on their back.

## Why It's Good For The Game

The only benefit a welderpack has is a small fuel tank. Other than that it is only a satchel, but without being able to be opened like one. Welderpack being the niche storage option that it is, it should not punish you for trying to take advantage of its one purpose.

## Changelog
:cl:
balance: Welderpack can be opened while worn.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -672,6 +672,7 @@
 	var/max_fuel = 260
 	storage_slots = null
 	max_storage_space = 15
+	worn_accessible = TRUE
 
 /obj/item/storage/backpack/marine/engineerpack/Initialize(mapload, ...)
 	. = ..()


TiviPlus:
Dunno how a viable this is, but you could use this as a PFC to get essentially normal storage space but with added flamer fuel storage

XElectricX:
You could do that already. I certainly did. You don't need to open the backpack though to refuel your tank so this is irrelevant.

TiviPlus:
Yes, but before you would have the negative effect of it having to be taken off to access it, while with this change it'll just be a better satchel that can hold my flamer fuel at no downside. As long as you can only get the welderpack exclusively from engi vendors this shouldnt be much as a problem.

XElectricX:
Non-vendor welderpacks can only be obtained from engineering lockers. Also if someone wanted to be a ghetto pyro, this isn't a deal-breaker. Again, you don't have to open the pack to refuel so in a combat situation it does not matter. The only uses for the welderpack are engineer-type roles and maybe the odd marine who is intent on using it for flamethrowers. Having to take off a satchel just to look in it is tedious.

TiviPlus:
If they're only obtainable from engi lockers/vendors everythings fine, but either way I meant that with this change there is no downside to using the welderpack as opposed to the satchel

XElectricX:
You are correct, there is no downside. But at the same time, the upside is just a portable fuel tank. I wouldn't say it is a huge advantage, just means you don't have to go back to base as often for welding. In the field, you're not going to be welding that much either. Most welding fuel is spent repairing defenses, and there is likely already a portable tank at the FOB (the one that is white and holds 600 units). And props to anyone if they want to be creative and use it for flamethrowing purposes. Emergent gameplay is great like that.

PR #3468, opened 2019-11-25
Revert "Grenade launcher and similar shoot through MMB or shift+click, instead of ctrl+click"

spookydonut:
Reverts tgstation/TerraGov-Marine-Corps#3413

Complaints have been raised from people who don't have middle mouse buttons.


Rohesie:
People have never complained about xeno abilities and MMB before. Why would they complain about grenade launchers now?

spookydonut:
> 
> 
> People have never complained about xeno abilities and MMB before. Why would they complain about grenade launchers now?

someone complained so i made this for them.

Rohesie:
That's fair. Just tell them to make use of the verb and toggle it, as it's not a real issue.

PR #3469, opened 2019-11-26
Adds a bunch of feedback vars

psykzz:
## About The Pull Request

This mimics the current round statistics to ensure the stats are saved in the database.

## Why It's Good For The Game

Working on that statbus

## Screens
image


Hughgent:
This is it. This is the thing I always wanted yet was unable to do.

code/controllers/subsystem/blackbox.dm
```diff
@@ -174,3 +174,59 @@ SUBSYSTEM_DEF(blackbox)
 	key_type = new_key_type
 	version = new_version
 	json = list()
+
+
+/datum/controller/subsystem/blackbox/proc/ReportDeath(mob/living/L)
+	set waitfor = FALSE
+	if(sealed)
+		return
+	if(!L || !L.key || !L.mind)
```
Rohesie:
```suggestion
	if(!L?.key || !L.mind)
```

spookydonut:
Do we actually have database access?

code/controllers/subsystem/blackbox.dm
```diff
@@ -174,3 +174,59 @@ SUBSYSTEM_DEF(blackbox)
...
 	json = list()
+
+
+/datum/controller/subsystem/blackbox/proc/ReportDeath(mob/living/L)
+	set waitfor = FALSE
+	if(sealed)
+		return
+	if(!L?.key || !L.mind)
+		return
+	var/sqlname = L.real_name
```
Rohesie:
```suggestion
	var/sqlname = sanitizeSQL(L.real_name)
```
And so on. No need to define the vars before the connect check, right?

psykzz: I was thinking about that, but i assume its because they want the check as late as possible?

PR #3471, opened 2019-11-26
Requisitions Defines

InterroLouis:
Basically I've taken all the defines for the crate costs, and converted them to plain numbers for ease of reading. I'm leaving the defines in at the top here-just so it's easier to see that I haven't made any changes to what things cost. 

## Why It's Good For The Game

It's not. It's also not bad. It's simply a code thing.

## Changelog
:cl:
code: Supplypacks uses plain numbers now.
/:cl:

### DIFF
--- a/code/datums/supplypacks.dm
+++ b/code/datums/supplypacks.dm
@@ -59,7 +59,7 @@ OPERATIONS
 					/obj/item/clothing/mask/gas/swat,
 					/obj/item/clothing/tie/storage/black_vest
 					)
-	cost = RO_PRICE_NORMAL
+	cost = 40
 	containertype = /obj/structure/closet/crate
 	containername = "\improper spec ops crate"
 	group = "Operations"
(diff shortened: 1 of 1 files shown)


jroinc:
Thank you so much.

You're a true hero.

Rohesie:
Do remove the defines as well, and try not to change any values without adding it to the changelog.

InterroLouis:
I was careful not to change any numbers from what they were. They're all the same, otherwise would be out of the scope of this PR. No, I did not try and sneak grenade cost changes into this.

spookydonut:
Just use a reasonable editor that lets you show define values rather than making this into an unmanageable mess
image

InterroLouis:
Did I mention it makes it easier to change specific numbers for balance purposes? Maybe I should have led with that, so you wouldn't arbitrarily close my PR because of your better code reading tools.

PR #3473, opened 2019-11-26
Buff light armor and harness, slightly nerf heavy

XElectricX:
## About The Pull Request

Light armor has become useless because all other armors provide better protection while still being slower than harness. Perhaps it will do better if it had a specialization. Bio and acid protections were increased to be more sturdy against neurotoxin and acid. No other armor serves that kind of role yet, and it makes sense for light armor to serve that role. Neurotoxin is a death sentence for speed-oriented marines as it affects stamina. Spitter acid can quickly build up burn damage and considerably slow down an unarmored marine over time from burns. A player will not be expected to 1v1 a crusher or ravager with light armor, but they will not be severely hampered by neurotoxin or acid burn like other armors. Lowered heavy and integrated storage bio and acid armor values to allow light armor to shine in that role, since they already provide high melee and bullet armor. With them being slow, it is also not as bad if they take stamina damage (until the player falls over).

Also removed the equip delay for harnesses because they are pitiful as armor and have no slowdown, so there is no reason to quick un/equip in order to move faster.

## Why It's Good For The Game

Light armor is a meme because it is not very good at being armor compared to the others. Instead of raising melee armor, which would result in powercreep, better to give it bio and acid resistances so that it won't be crippled by ranged castes. Not as fast as harness, but will keep you running for longer when attacked by toxin/acid.

## Changelog
:cl:
balance: Light armor becomes anti-acid/toxin. Heavy and IS have less resistance.
tweak: Harness no longer has an un/equip delay.
/:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -98,14 +98,14 @@
 	name = "\improper M3-H pattern marine armor"
 	desc = "A standard Marine M3 Heavy Build Pattern Chestplate. Increased protection at the cost of slowdown."
 	icon_state = "1"
-	armor = list("melee" = 65, "bullet" = 70, "laser" = 60, "energy" = 30, "bomb" = 60, "bio" = 50, "rad" = 20, "fire" = 50, "acid" = 50)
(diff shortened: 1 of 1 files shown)


Pariah919:
The problem isn't protection. It's bump melee and stamina.

Kuro-ai:
Why are you nerfing heavy armor to buff light armor?this doesn't solve the issue and has a unecessary nerf to a balanced item

PR #3484, opened 2019-11-28
Adds an admin approval proc

psykzz:
## About The Pull Request
Adds an admin approval proc, basically expanding on the idea of how the distress call is handled.

## Why It's Good For The Game
This allows things to happen while admins are offline, but when online allows admins to gate certain activities as required.

### DIFF
--- a/code/__HELPERS/admin.dm
+++ b/code/__HELPERS/admin.dm
@@ -1,0 +1,57 @@
+GLOBAL_LIST_INIT(admin_approvals, list())
+
+/*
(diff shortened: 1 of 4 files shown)


spookydonut:
Cool idea, you could make it use a signal/callback to work generically and slap it on everything

Rohesie:
Would you say this is the finished version, at least for now, or do you intend to do the suggested changes?

psykzz:
Would personally continue for now and add signals as another proc or update.

I want this to be the admin version of do_after, unless we want to convert
those to signals too I'd continue as we are.

PR #3486, opened 2019-11-29
Underbarrel Flashlights

David-Stormwell:
## About The Pull Request

Adds in Underbarrel Flashlights!
You can now feel like a tacticool operator with Underbarrel Lights on I think every gun expect for 1 or 2! Also, you can use a wrench on a normal flashlight to make it into a UBL/ Underbarrel Light!
Also a few guns like the Lever-action rifle and the Double-barrel shotgun can now accept rail or underbarrel lights!

## Why It's Good For The Game

How do we have a "Rail Flashlight" yet not a Under-Rail light?
Also more survivor guns have the ability to have lights now.

## Changelog
:cl:
add: Adds in Under-Barrel/Rail Lights! 
balance: Adds more lights to all the guns!
/:cl:


code/game/objects/items/flashlight.dm
```diff
@@ -69,7 +70,26 @@
...
+			return
+
+		if(on)
+			to_chat(user, "<span class='warning'>Turn off [src] first.</span>")
+			return
+
+		if(loc == user)
+			user.dropItemToGround(src) //This part is important to make sure our light sources update, as it calls dropped()
 
+		var/obj/item/attachable/underbarrel_light/A = new(loc)
```
Rohesie:
```suggestion
		var/obj/item/attachable/underbarrel_light/modified_light = new(loc)
```
More informing names than`A`

code/game/objects/items/flashlight.dm
```diff
@@ -69,7 +70,26 @@
...
+		user.put_in_hands(A) //This proc tries right, left, then drops it all-in-one.
+		to_chat(user, "<span class='notice'>You modify [src]. It can now be mounted on a weapon.</span>")
```
Rohesie:
```suggestion
		to_chat(user, "<span class='notice'>You modify [src]. It can now be mounted on a weapon.<br>Use a wrench on [modified_light] to change it back.</span>")
```
You can call the proc a single time using `<br>` (line break) to add a new line.

code/game/objects/items/flashlight.dm
```diff
@@ -11,6 +11,7 @@
 	actions_types = list(/datum/action/item_action)
 	var/on = FALSE
 	var/brightness_on = 5 //luminosity when on
+	var/underbarrel_light_compatible = TRUE //Can this be turned into a underbarrel light ?
```
Rohesie: Make it `FALSE` by default, and only add it to the ones that can.
Alternatively make it a flag, poke me on discord if you want to know how. If not, that's fine.

Rohesie:
This will cause issues in its interaction with the rail flashlight, so you'll need o either refactor the gun/light code a little or do a safety check to make both incompatible.

jroinc:
Balance concerns as well.

If I put a raillight/griplight on my primary, a beltgun, and if I have 2 flashlights and armor light, I THINK that'll light up all of the screen.

adamkad1:
> Balance concerns as well.
> 
> If I put a raillight/griplight on my primary, a beltgun, and if I have 2 flashlights and armor light, I THINK that'll light up all of the screen.

thats a lot of slots you are sacrificing

Kuro-ai:
> Balance concerns as well.
> 
> If I put a raillight/griplight on my primary, a beltgun, and if I have 2 flashlights and armor light, I THINK that'll light up all of the screen.

but then you'll sacrifice undergrip and upper rail slots (the only 2 slots current guns have for the most part)you'll sacrifice all your armor storage unless it's IS and you'll sacrifice a belt slot,I think this is balanced

code/game/objects/items/flashlight.dm
David-Stormwell: I don't think I can, as I only want the flashlight itself to be true, and all the children are already false, and how would I set the flashlight back to true even though it's false?

David-Stormwell:
I'll be doing this on a later date.

PR #3502, opened 2019-12-01
Buffs Xenos upgrade/evolution times

Kuro-ai:
## About The Pull Request

Reduces evolution and upgrade times for all xenos across the board

## Why It's Good For The Game

A buff for xenos so they hopefully have a better chance against marines,this reduces all the boring wait times and elevator gameplay that makes xeno so infuriating to play as,since my stat changes to xenos made maturities not crazy jumps in stats this will be balanced.
## Changelog
:cl:
Balance: Reduced all xeno evolution and upgrade timers across the board
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
@@ -3,7 +3,7 @@
 	speak_emote = list("hisses")
 	icon_state = "Bloody Larva"
 	amount_grown = 0
-	max_grown = 100
+	max_grown = 10
 	maxHealth = 35
 	health = 35
 	see_in_dark = 8
--- a/code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
@@ -56,9 +56,9 @@
 	var/progress = "" //Naming convention, three different names
 
 	switch(amount_grown)
-		if(0 to 49) //We're still bloody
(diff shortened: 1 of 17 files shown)


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -4. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
```diff
@@ -3,7 +3,7 @@
 	speak_emote = list("hisses")
 	icon_state = "Bloody Larva"
 	amount_grown = 0
-	max_grown = 100
+	max_grown = 10
```
Rohesie: Increase this a bit. 10 is almost instant.

code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
```diff
@@ -56,9 +56,9 @@
 	var/progress = "" //Naming convention, three different names
 
 	switch(amount_grown)
-		if(0 to 49) //We're still bloody
+		if(0 to 5) //We're still bloody
 			progress = "Bloody "
-		if(100 to INFINITY)
+		if(6 to INFINITY)
 			progress = "Mature "
```
Rohesie: Instead of `amount_grown` with hardcoded values, do a new ratio variable, something like this:
`var/growth_progress = (amount_grown * 100) / max_grown`
Then you can just do a the switch using the percentage, and if you change the values it will be fine.

PR #3505, opened 2019-12-03
Shield component

Rohesie:
## Changelog
:cl:
add: Riot shields now absorb damage until they are destroyed. Component made which allows for new kinds of shields, abstract or concrete.
/:cl:


code/datums/components/shield.dm
```diff
@@ -0,0 +1,324 @@
+#define SHIELD_TOGGLE (1<<0) //Can be toggled on and off.
+#define SHIELD_PURE_BLOCKING (1<<1) //Only runs a percentage chance to block, and doesn't interact in other ways.
+#define SHIELD_PARENT_INTEGRITY (1<<2) //Transfers damage to parent's integrity.
+
+/datum/component/shield
+	var/mob/living/affected
+	var/datum/callback/intercept_damage_cb
+	var/datum/callback/transfer_damage_cb
+	var/datum/armor/armor
+	var/datum/armor/cover = list("melee" = 80, "bullet" = 100, "laser" = 100, "energy" = 100, "bomb" = 80, "bio" = 30, "rad" = 0, "fire" = 80, "acid" = 80) //Percentage damage it intercepts.
```
nemvar: Move this to Initialize as the default value of ``shield_cover``.

code/datums/components/shield.dm
```diff
@@ -0,0 +1,324 @@
+#define SHIELD_TOGGLE (1<<0) //Can be toggled on and off.
+#define SHIELD_PURE_BLOCKING (1<<1) //Only runs a percentage chance to block, and doesn't interact in other ways.
+#define SHIELD_PARENT_INTEGRITY (1<<2) //Transfers damage to parent's integrity.
```
nemvar: These defines shouldn't just be local.

code/modules/mob/living/living_defense.dm
```diff
@@ -243,3 +243,18 @@
 	if(CHECK_BITFIELD(S.smoke_traits, SMOKE_CHEM))
 		S.reagents?.reaction(src, TOUCH, S.fraction)
 	return protection
+
+/mob/living/proc/check_shields(attack_type, damage, damage_type = "melee", silent)
+	if(!damage)
+		stack_trace("check_shields called without a damage value")
+		return TRUE
```
nemvar: Should probably return 0 so it's consistent in it's return value.

Rohesie:
This should be ready for merge.

PR #3509, opened 2019-12-03
Basic port of footsteps.

nemvar:
## About The Pull Request

Ports the footstep component from tg

Full feature list:
Footsteps for humans
Footsteps for monkeys
Implements support for xeno footsteps but doesn't add them.
Potentially footsteps for any living mob.
Different footsteps based on the underlying turf. Only expection to the 'turf determines sound' rule is resin which I think is reasonable considering how core it is to the gameplay.

## Why It's Good For The Game

Adds immersion.
TODO list:
Fine tuning.
Maybe beno footsteps

## Changelog
:cl:
add: Added footsteps.
/:cl:


nemvar:
Alright if this doesn't throw errors, it's basically ready. Actually adding footstep sounds to xenos is like 5 more lines, only real problem being the charger thing.

code/game/turfs/floor.dm
```diff
@@ -346,7 +355,7 @@ GLOBAL_LIST_INIT(wood_icons, list("wood", "wood-broken"))
 	levelupdate()
 
 //This proc will make a turf into a wood floor. Fun eh? Insert the wood tile to be used as the argument
-//If no argument is given a new one will be made.
+//If no argument is given a new one will be made. SOMEONE FIX THIS COPYPASTE BULLSHIT!!
```
Rohesie: This isn't even used by anything. Feel free to clean it up on a different PR.

Rohesie:
`/turf/open/floor/mainship` is the most common turf on the ship. Think it'd be interesting to add a sound to it, or would it be too much?

Maybe we should repath the floors, we have `/turf/open/floor/tile` , `/turf/open/floor/marking` and whatnot.

I could do some of that work if you have a direction to point.

nemvar:
all turfs on the ship are /mainship turfs, that's why there's some ugly code about changing footstep sounds in update_icon().
Changing the floor to say, for example, wood does keep the turf /mainship and only changes the icon

nemvar:
turfs defo need repathing though. inhertitance tree makes no sense

PR #3515, opened 2019-12-04
Powerloader set direction fix

Rohesie:
Micro-optimizations going wrong and causing runtimes.

### DIFF
--- a/code/modules/vehicles/powerloader.dm
+++ b/code/modules/vehicles/powerloader.dm
@@ -135,7 +135,7 @@
 
 /obj/vehicle/powerloader/setDir(newdir)
 	. = ..()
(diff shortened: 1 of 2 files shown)


code/modules/vehicles/powerloader.dm
```diff
@@ -135,7 +135,7 @@
 
 /obj/vehicle/powerloader/setDir(newdir)
 	. = ..()
-	if(buckled_mob?.dir != dir)
+	if(buckled_mob && buckled_mob.dir != dir)
```
spookydonut:
```suggestion
	if(!(buckled_mob?.dir == dir))
```

Rohesie:
Deprecated.

PR #3523, opened 2019-12-05
Removes /obj/item/projectile in favour of /obj/projectile

nemvar:
## About The Pull Request

Both literally only shared a single variable, being hitsound. Item overhead is pretty big (for something that gets spawned thousands of times every round and only exists for 3 seconds at most) so I don't see a reason to them as parent.

This is comparable to https://github.com/tgstation/tgstation/pull/46692 which worked perfectly fine on tg and hasn't caused any issues since merge.

It's kinda hard to do extensive testing on this alone but considering that the whole search/replace only resulted in a single error and I didn't notice any obvious fuckery while trying it out I'd wager this is probably fine.

## Why It's Good For The Game

Die silly object tree die

## Changelog

probably not needed. purely backend


Rohesie:
Conflicting with currently TMed PRs. Will need to solve conflicts once they are merged. Likely the bullet ping fix PR.

Rohesie:
Merge conflict.

nemvar:
:(

Rohesie:
`git rebase upstream/master ; git push -f`
You won't even need to solve conflicts, this is just a repath.

nemvar:
Done

PR #3525, opened 2019-12-05
Skills refactor

Rohesie:
* Skills datums are singletons.
* Skills reference moved from human mind to mob.


Rohesie:
This is ready for merge.

PR #3526, opened 2019-12-06
Unatomized turf-related cleanup

Rohesie:
An unatomized start to cleaning the turf code a bit towards /tg/ patterns. This should not really affect regular gameplay at all, and are all innocuous changes by themselves.

* Kills the signal redirect component.
* Takes `update_icon()` and `update_overlays()` to the base and adds `update_icon_state()`
* Removes duplicate `/turf/open/space/transit` code.
* Removes lights from `/turf/open/space`, making it not have dynamic lighting by default.
* Gives mobs entering space 600 burn damage to stop space-walking memes.
* Makes space in general not use dynamic lighting, except for the transit shuttle turfs and area, so the dropship traveling looks better, without light skips.


code/game/area/areas/shuttles.dm
```diff
@@ -48,7 +48,7 @@
 /area/shuttle/transit
 	name = "Hyperspace"
 	desc = "Weeeeee"
-	dynamic_lighting = DYNAMIC_LIGHTING_DISABLED
+	//dynamic_lighting = DYNAMIC_LIGHTING_DISABLED //Different from /tg/
```
psykzz: Instead of commented commented-out code, can you just set dynamic_lighting to the default with the comment to explain that its being explicit?

Rohesie: Sure.

Issue #2828, opened 2019-09-01
Planting Eggs Underneath Objects


InterroLouis:
## Testmerges:
Server-Side Code Modifications: https://tgstation13.org/parsed-logs/basil/server_side_modifications.dm

Round ID: 949
BYOND Version: 512.1464
Server revision compiled on: 
Master commit: c2862d6f94028f8efc34581d8cb4ae35c19bc88d
The following pull requests are currently test merged:
#2821: 'Fixes a sound runtime' by DominikPanic at commit 76f205e490
#2736: 'Movespeed mod and datum charge/crushing' by Rohesie at commit 48826a261c
#2797: 'Reduces the roundstart notransform lock time' by DominikPanic at commit 83b60755ef
#2782: 'Fulton extraction devices' by Rohesie at commit f989905051
#2795: '[WIP] Pillar of Spring' by grinkgo at commit ff2a5ef8af
#2823: 'Fixes powerloader clamp runtimes' by DominikPanic at commit df72aa5602
#2825: 'Incoming mentorhelps now have a different sound' by DominikPanic at commit ec8dfe9057
#2819: 'New PFC armor sprites and new squad color armor markings' by Metroidlover at commit be5cc9c6fa
#2814: '[W.I.P] Interaction rework' by DominikPanic at commit 5f3a5d6e68

Local commit: 4c5cfbf50171dbb11b3322ba9556335a64ab82a3
Server tools version: /tg/station 13 Server v3.2.6.2
## Reproduction:
The issue here is xenos being able to plant eggs underneath obscuring objects. Mainly at this point, trees. You can't see the egg unless you right click the exact tile(which is usually difficult to do since right clicking the tree tends to get you the base tile of the tree), you can't shoot the egg without seeing it. Basically we just need to...make the eggs layer over everything, like we did to mines.

DominikPanic:
@Rohesie What do you think?

drawsstuff:
I've seen this used to great effect but wouldn't mind seeing it go. Carrier traps exist for a reason.

Alterist:
I feel like the trees I coded in are causing this issue, is there a way to make the tree semi-transparent?

Rohesie:
3/4 perspective and objects larger than 32x32 are an issue, and not a trivial one.
By something behind it you mean specifically an egg, or anything?
[...]

PR #3528, opened 2019-12-06
Makes trees transparent if important objects are hiding behind them.

nemvar:
## About The Pull Request

As per title. Also makes you unable to click the tree while something is behind it. Hopefully nobody complains about the lack of lumberjacking during battle.
Current list of important things:
Facehuggers.
Eggs.
All living things.
Nuke

## Why It's Good For The Game

I really dislike the idea of the games perspective fucking people over.
Fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/2828

I'm feeling really wonky about COMSIG_ATOM_INITIALIZED_ON but byond doesn't call Entered by default when things get spawned and I don't want to mess with the natural order of thing.

## Changelog
:cl:
tweak: Trees now turn transparent if something is behind them.
/:cl:


code/game/atoms.dm
```diff
@@ -465,10 +465,11 @@ Proc for attack log creation, because really why not
 
 	if(light_power && light_range)
 		update_light()
-
-	if(opacity && isturf(loc))
-		var/turf/T = loc
-		T.has_opaque_atom = TRUE // No need to recalculate it in this case, it's guaranteed to be on afterwards anyways.
+	if(loc)
+		SEND_SIGNAL(loc, COMSIG_ATOM_INITIALIZED_ON, src) //required since spawning something doesn't call Move hence it doesn't call Entered.
```
```suggestion
		SEND_SIGNAL(loc, COMSIG_ATOM_INITIALIZED_ON) //required since spawning something doesn't call Move hence it doesn't call Entered.
```
`src` is sent by default as the first argument.

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,80 @@
...
+/datum/component/largetransparency/proc/OnMove()
+	amounthidden = 0
+	restoreAlpha()
+	UnregisterFromParent()
+	RegisterWithParent()
+
+/datum/component/largetransparency/proc/OnTurfChange()
+	addtimer(CALLBACK(src, .proc/OnMove, 1)) //*pain
+
+/datum/component/largetransparency/proc/objectEnter(atom/source, atom/thing)
```
Rohesie:
```suggestion
/datum/component/largetransparency/proc/objectEnter(datum/source)
```
And typecast it.

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,80 @@
...
+	registered_turfs.Cut()
+
+/datum/component/largetransparency/proc/OnMove()
+	amounthidden = 0
+	restoreAlpha()
+	UnregisterFromParent()
+	RegisterWithParent()
+
+/datum/component/largetransparency/proc/OnTurfChange()
+	addtimer(CALLBACK(src, .proc/OnMove, 1)) //*pain
```
Rohesie:
```suggestion
	INVOKE_NEXT_TICK(src, .proc/OnMove) //*pain
```

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,80 @@
...
+			tu2 = get_step(tu2, WEST)
+			registered_turfs.Add(tu1)
+			registered_turfs.Add(tu2)
+	for(var/t in registered_turfs)
+		RegisterSignal(t, list(COMSIG_ATOM_ENTERED, COMSIG_ATOM_INITIALIZED_ON), .proc/objectEnter)
+		RegisterSignal(t, COMSIG_ATOM_EXITED, .proc/objectLeave)
+		RegisterSignal(t, COMSIG_TURF_CHANGE, .proc/OnTurfChange)
+		for(var/a in t)
+			var/atom/at = a
+			if(!(at.flags_atom & CRITICAL_ATOM))
```
Rohesie: Could you use a little bit more descriptive vars?

code/game/objects/structures/flora.dm
```diff
@@ -31,6 +31,12 @@
 	max_integrity = 500
 	layer = ABOVE_FLY_LAYER
 	var/log_amount = 10
+	var/y_coverage = 2
+	var/x_coverage = 0
```
Rohesie: Either a proc to change this, with a signal registered by the component, or just a proc ran on `Initialize()` that sends these values and can be redefined by children objects, instead of variables?

Rohesie:
Add the nuke as a critical atom?

code/game/atoms.dm
nemvar: It is not. loc is sent by default as the first argument.

Rohesie: Oh, I misread that, you are sending the signal to loc. Nevermind.

code/datums/components/largeobjecttransparency.dm
nemvar: COMSIG_ATOM_ENTERED, COMSIG_ATOM_EXITED and COMSIG_ATOM_INITIALIZED_ON always have atoms as source so the type isn't actually incorrect. If you are just going to cast your variable with a different type anyways, I don't see why you couldn't do that in proc arguments.
I will just quickly change this for appeasements sake and since I don't use these anyhow but in concept I see nothing wrong with this.

Rohesie: I don't have strong feelings about it, it's the standard @ninjanomnom had told me to follow. Let's see what he says.

code/datums/components/largeobjecttransparency.dm
nemvar: I'm bad at that-

Rohesie:
This is working nicely. Can you add it to `/obj/structure/table` and `/obj/structure/barricade`? When they are facing south they cover mobs laying down, and since we have plans for adding crouching/crawling, this will become a concern.

nemvar:
I'll see what I can do but it does involve reworking how getting the target area works. I should do that anyways because I am currently being really stupid about the way I do that.

nemvar:
A not small concern is that currently, being transparent completely blocks you from interacting with an object. That means you won't be able to attack/repair a barricade should it get hidden.

Rohesie:
Yes, that's a concern. But it's something we'll need to address when the crawling memes arrive.

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,83 @@
...
+
+/datum/component/largetransparency/proc/reduceAlpha()
+	var/atom/par = parent
+	par.alpha = 140
+	par.mouse_opacity = 0
+
+/datum/component/largetransparency/proc/restoreAlpha()
+	var/atom/par = parent
+	par.alpha = 255
+	par.mouse_opacity = 2
```
psykzz: You should actually store previous values to be restored so that admins can varedit things without being reset.

nemvar: If you want to varedit the alpha/mouse_opacity of something, you should just delete this component since it will always mess up your day one way or the other. Real concern is if someone were to apply this to something that does not have 255 alpha by default.

nemvar:
Alright, the getting of the turfs just got a whole lot saner.
I'm not convinced this implementation is ideal for barricades though since those are actually a lot easier to implement since movable atoms support crossed/uncrossed and only need to look at a single turf.

code/game/objects/structures/flora.dm
nemvar: Did that, doesn't look at lot better though.

code/datums/components/largeobjecttransparency.dm
nemvar: Well works now.

nemvar:
This is what multi-tile hitboxes look like in byond. Let that be a lesson to anyone who is stupid enough to even think about it.

Rohesie:
Makes sense. Let us know when you feel this is ready not just for testing.

nemvar:
Technically ready since I would deem barricades out of scope.

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,98 @@
...
+	y_size = _y_size
+	if(isnull(_initial_alpha))
+		var/atom/at = parent
+		initial_alpha = at.alpha
+	else
+		initial_alpha = _initial_alpha
+	target_alpha = _target_alpha
+	toggle_click = _toggle_click
+	registered_turfs = list()
+	RegisterSignal(parent, COMSIG_MOVABLE_MOVED, .proc/OnMove)
```
ninjanomnom: This should be in RegisterWithParent if you're going to use that proc

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,98 @@
...
+		if(!regist_tu)
+			continue
+		RegisterSignal(regist_tu, list(COMSIG_ATOM_ENTERED, COMSIG_ATOM_INITIALIZED_ON), .proc/objectEnter)
+		RegisterSignal(regist_tu, COMSIG_ATOM_EXITED, .proc/objectLeave)
+		RegisterSignal(regist_tu, COMSIG_TURF_CHANGE, .proc/OnTurfChange)
+		for(var/thing in regist_tu)
+			var/atom/at = thing
+			if(!(at.flags_atom & CRITICAL_ATOM))
+				continue
+			amounthidden++
```
ninjanomnom: You need to do all this again if parent moves so put this in its own proc

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,98 @@
...
+	if(amounthidden)
+		reduceAlpha()
+
+/datum/component/largetransparency/UnregisterFromParent()
+	for(var/regist_tu in registered_turfs)
+		UnregisterSignal(regist_tu, list(COMSIG_ATOM_ENTERED, COMSIG_ATOM_EXITED, COMSIG_TURF_CHANGE, COMSIG_ATOM_INITIALIZED_ON))
```
ninjanomnom: Make the list once

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,98 @@
...
+	registered_turfs.Cut()
+
+/datum/component/largetransparency/proc/OnMove()
+	amounthidden = 0
+	restoreAlpha()
+	UnregisterFromParent()
+	RegisterWithParent()
+
+/datum/component/largetransparency/proc/OnTurfChange()
+	INVOKE_NEXT_TICK(src, .proc/OnMove) //*pain
```
ninjanomnom: Use a unique addtimer

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,98 @@
...
+		return
+	amounthidden = max(0, amounthidden - 1)
+	if(!amounthidden)
+		restoreAlpha()
+
+/datum/component/largetransparency/proc/reduceAlpha()
+	var/atom/par = parent
+	par.alpha = target_alpha
+	if(toggle_click)
+		par.mouse_opacity = 0
```
ninjanomnom: There's a define for mouse_opacity values

code/datums/components/largeobjecttransparency.dm
```diff
@@ -0,0 +1,98 @@
...
+
+/datum/component/largetransparency/proc/restoreAlpha()
+	var/atom/par = parent
+	par.alpha = initial_alpha
+	if(toggle_click)
+		par.mouse_opacity = 2
```
ninjanomnom: Ditto

code/__DEFINES/components.dm
```diff
@@ -118,6 +118,7 @@
 #define COMSIG_ATOM_EX_ACT "atom_ex_act"						//from base of atom/ex_act(): (severity, target)
 #define COMSIG_ATOM_SET_LIGHT "atom_set_light"					//from base of atom/set_light(): (l_range, l_power, l_color)
 #define COMSIG_ATOM_BULLET_ACT "atom_bullet_act"				//from base of atom/bullet_act(): (/obj/projectile)
+#define COMSIG_ATOM_INITIALIZED_ON "atom_initialized_on"		//called from atom/Initialize() of target: (atom/target)
```
ninjanomnom: You made this as a turf signal not an atom signal

nemvar:
should be good to go

Issue #3524, opened 2019-12-05
Nukes enter stealth mode when put on the same tile as a bed (it layers under the bed)


adamkad1:
When you put nuke on same tile as bed, it layers under the bed, rendering it completely invisible

## Testmerges:

#3520: 'Tweaks footsteps' by nemvar at commit 8059315288
#3505: 'Shield component' by Rohesie at commit bec8fb9bf3
#3514: 'Duplicate tables fix' by Rohesie at commit a832c30d4b
#3516: 'Bullet ping fix' by Rohesie at commit 9f7329dc87
#3517: 'Fix new linter errors' by spookydonut at commit 2dbba055b2
#3512: 'Fixes marines being able to bypass the SD shutters' by Rohesie at commit 9f3656825a
#3513: 'Distress resin silos' by Rohesie at commit 30457818ee
#3499: 'Fluff sprites' by Rohesie at commit 0d076799a4
#3519: 'Fixes admin approval proc showing to mentors' by psykzz at commit 0a1e73d665
#3521: 'Lowpop Crash' by Rohesie at commit 5d5008b601

## Reproduction:
1-Get nuke
2-push onto bed
3-profit

PR #3531, opened 2019-12-06
Bumps the nuke up a few layers

nemvar:
## About The Pull Request

The nuke is now above beds. Alternative and preferable solution is to have beds be saner with their layers but this is the immediate fix.

## Why It's Good For The Game

Fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/3524

## Changelog
:cl:
fix: The nuke can no longer be hidden under beds.
/:cl:


Rohesie:
This probably looks weird when the Ravager is south of it, but as said before, sprites over 32 pixels tall are kind of an issue.

nemvar:
It's still below mob layer so it probably looks fine.

nemvar:
yeahfine

Rohesie:
No concerns then, nice.

PR #3538, opened 2019-12-08
buffs xeno upgrade/evo times 2: electric boogaloo

terranaut1:
## About The Pull Request

halves age timers (yes thats less of a buff than the previous one)
evo timers decreased by 20 (100 > 80, 200 > 180)
i did this with find and replace and a quick skim over the changed files if theres some oddities that would fall under the radar please point them out but i dont think there are any

## Why It's Good For The Game

apparently this is desired and needed

## Changelog
:cl:
balance: Xenos upgrade faster
/:cl:


nemvar:
Switch these numbers for defines??

PR #3541, opened 2019-12-08
Speed up larva growth

psykzz:
## About The Pull Request
Speed up larva growth
## Why It's Good For The Game
Reduces the time spent as larva, increasing response time from a respawn.

## Changelog
:cl:
tweak: Larvas now evolve twice as quuick
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
@@ -3,7 +3,7 @@
 	speak_emote = list("hisses")
 	icon_state = "Bloody Larva"
 	amount_grown = 0
-	max_grown = 100
+	max_grown = 50
 	maxHealth = 35
 	health = 35
 	see_in_dark = 8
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
```diff
@@ -3,7 +3,7 @@
 	speak_emote = list("hisses")
 	icon_state = "Bloody Larva"
 	amount_grown = 0
-	max_grown = 100
+	max_grown = 50
```
Rohesie: This has the same issue as previous PRs: it doesn't update the larva name/appearance correctly.
Ideally we refactor this to use the same system as other xenos.
See: https://github.com/tgstation/TerraGov-Marine-Corps/pull/3502

psykzz: Larva is on a different style, so i just updated that to be a percent of the max and generate the name from that instead.

PR #3542, opened 2019-12-08
larvapoints includes revivable humans

psykzz:
## About The Pull Request

When calculating the number of humans, revivable humans are included

## Why It's Good For The Game

Improves balance of humans / xenos

### DIFF
--- a/code/modules/mob/living/carbon/human/dummy.dm
+++ b/code/modules/mob/living/carbon/human/dummy.dm
@@ -5,6 +5,9 @@
 
 INITIALIZE_IMMEDIATE(/mob/living/carbon/human/dummy)
 
+/mob/living/carbon/human/dummy/Initialize()
+	return // This stops dummies being setup and registered in the human_mob_list
+
 /mob/living/carbon/human/dummy/Destroy()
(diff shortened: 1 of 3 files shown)


psykzz:
Ignore the last commits, wrong PR.

terranaut1:
i like this idea but this implementation of it causes a runtime with get_total_joblarvaworth which ultimately locks the game down since it stops creating burrowed

psykzz:
@terranaut1 it was including dummy humans, which i've now removed.

code/modules/mob/living/carbon/human/dummy.dm
```diff
@@ -5,6 +5,9 @@
 
 INITIALIZE_IMMEDIATE(/mob/living/carbon/human/dummy)
 
+/mob/living/carbon/human/dummy/Initialize()
+	return // This stops dummies being setup and registered in the human_mob_list
```
Rohesie:
```suggestion
	return INITIALIZE_HINT_NORMAL // This stops dummies being setup and registered in the human_mob_list
```

PR #3544, opened 2019-12-08
Add Fax on bioscan

psykzz:
## About The Pull Request

Marines never got the bioscan, but this way they get it but require communication.

## Why It's Good For The Game

Immersions

## Changelog
:cl:
add: Marines now get a fax in Crash with bioscan information
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -37,6 +37,7 @@
 	latejoin_larvapoints_required = 9 // to avoid division by zero if config doesn't deliver a value in :58 for some reason
 
 	var/larva_check_interval = 0
+	var/bioscan_interval = INFINITY
 
 
 /datum/game_mode/crash/initialize_scales()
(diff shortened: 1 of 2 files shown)


psykzz:
Every 10mins admin might see the admin_log twice, not sure if its a big enough deal to handle right now.

code/datums/gamemodes/crash.dm
```diff
@@ -145,8 +145,9 @@
 	shuttle_landed = TRUE
 
 	// We delay this a little because the shuttle takes some time to land, and we want to the xenos to know the position of the marines.
-	addtimer(CALLBACK(src, .proc/announce_bioscans, TRUE, 0, FALSE, TRUE), 30 SECONDS)  // Announce exact information to the xenos.
-	addtimer(CALLBACK(src, .proc/announce_bioscans, TRUE, 0, FALSE, TRUE), 5 MINUTES, TIMER_LOOP)
+	addtimer(CALLBACK(src, .proc/announce_bioscans, TRUE, 0, FALSE, TRUE, FALSE), 30 SECONDS)  // Announce exact information to the xenos.
+	addtimer(CALLBACK(src, .proc/announce_bioscans, TRUE, 0, FALSE, TRUE, FALSE), 5 MINUTES, TIMER_LOOP) // Xeno Regular, 5mins
```
Rohesie: I've started removing procs from `TIMER_LOOP`, as I was told by ninjanomnom it wasn't a good idea to use it unless as a last resource.
Could you add this to the processing loop?

nemvar: announcement subsystem doesn't sound unreasonable.

Rohesie: Wouldn't be used much, but yea, it'd be fair.

psykzz: Hard to invision how an annoucement subsystem could work, so i went with process for now.

PR #3552, opened 2019-12-10
Reduce hunter knockdown from max of 6 secs to 0.5

psykzz:
## About The Pull Request

Hunter could knock someone unconcious for up to 6 secs, 2 if running.
This replaces that with a flat 0.5 sec knockdown but only when disarming instead of on harm.

## Why It's Good For The Game
Xeno balance
## Changelog
:cl:
balance: hunter stealth attack changed to flat 0.5 seconds, and only applied on disarm.
/:cl:


Rohesie:
There should be no knockdown at all, really.

adamkad1:
is this gonna make hunter useless?

XElectricX:
No more hunter stun memes. I could cry right now.

terranaut1:
You should probably decrease the cooldown and plasmacost then, otherwise it loses its properties of being able to deal with heavier type marines with slow weapons.
Or maybe give it some AP :)

PR #3553, opened 2019-12-10
Removes M4RA double damage button (burst mode) (again)

adamkad1:
## About The Pull Request
The gun has:
1-(nearly)perfect accuracy
2-burst that fires 2 shots instantly, both going the same route, basically being a double damage button
3-almost nonexistent refire delay
If you got a better idea than this, feel free to propose it yourself

Idea for removing burst ditched, instead makes the gun slightly slower
And back to removing burst

## Why It's Good For The Game

Battle rifle is too powerfull currently, Plenty a people solo'ed multiple xenos with it (including me)

## Changelog
:cl:
balance: Removed the M4RA's burst mode and increased its recoil.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -320,7 +320,7 @@
 	attachable_offset = list("muzzle_x" = 32, "muzzle_y" = 17,"rail_x" = 12, "rail_y" = 23, "under_x" = 23, "under_y" = 13, "stock_x" = 24, "stock_y" = 13)
 	starting_attachment_types = list(/obj/item/attachable/scope/mini/m4ra, /obj/item/attachable/stock/rifle/marksman)
 
-	fire_delay = 1 SECONDS
+	fire_delay = 0.7 SECONDS
 	accuracy_mult = 1.05
 	burst_amount = 2
 	burst_delay = 0.5 SECONDS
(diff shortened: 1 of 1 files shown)


psykzz:
Instead of flat removing it, could you 
1-reduce the accuracy
2-fix the 2 shot instant burst 
3-add a refire delay
4-reduce the (if not) perfect accuracy

adamkad1:
good ideas, maybe more people could help me pick

psykzz:
Perhaps just start with the refire delay and see how it goes from there, doesn't have to be a big bang change.

adamkad1:
After a check, i think the burst delay is fine but the refire delay might be a bit slow for people used to previous one, but i could use someone else's opinion

adamkad1:
toned down refire delay a bit

adamkad1:
Im pretty sure most people would rather have burst removed than reduce its firerate

psykzz:
This PR is pretty volatile, changing frequently between approaches, can you lock down the change you want?
We'll test that specific change and either merge or close.

adamkad1:
Well, we tested slower br, a bunch of people spoke up saying its trash, some said burst removal would be better so lets just keep at that

psykzz:
No i like the idea, im just saying if we are looking to merge have you settled one what approach?

psykzz:
If you update the changelog, happy to merge.

adamkad1:
updated changelog

adamkad1:
or not, ill just increase recoil to make automode unappealing

PR #3554, opened 2019-12-10
Updated acid spray processing

psykzz:
Previously on SSobj, it was checking every 2 secs, but was on a 1 sec timer, this means a second was not checked delaying the skill.

Moved it to SSfastprocess, which checks every 2 ticks, which should trigger closer to the 1 sec cooldown.

### DIFF
--- a/code/game/objects/effects/aliens.dm
+++ b/code/game/objects/effects/aliens.dm
@@ -48,11 +48,11 @@
 
 /obj/effect/xenomorph/spray/Initialize(mapload, duration = 10 SECONDS) //Self-deletes
 	. = ..()
-	START_PROCESSING(SSfastprocess, src)
+	START_PROCESSING(SSprocessing, src)
 	QDEL_IN(src, duration + rand(0, 2 SECONDS))
 
 /obj/effect/xenomorph/spray/Destroy()
-	STOP_PROCESSING(SSfastprocess, src)
+	STOP_PROCESSING(SSprocessing, src)
 	return ..()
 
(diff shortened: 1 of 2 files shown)


code/game/objects/effects/aliens.dm
```diff
@@ -48,11 +48,11 @@
 
 /obj/effect/xenomorph/spray/Initialize(mapload, duration = 10 SECONDS) //Self-deletes
 	. = ..()
-	START_PROCESSING(SSobj, src)
+	START_PROCESSING(SSfastprocess, src)
```
Rohesie:
```suggestion
	START_PROCESSING(SSprocessing, src)
```
This one checks every second.

code/game/objects/effects/aliens.dm
```diff
@@ -48,11 +48,11 @@
...
 	QDEL_IN(src, duration + rand(0, 2 SECONDS))
 
 /obj/effect/xenomorph/spray/Destroy()
-	STOP_PROCESSING(SSobj, src)
+	STOP_PROCESSING(SSfastprocess, src)
```
Rohesie:
```suggestion
	STOP_PROCESSING(SSprocessing, src)
```

code/game/objects/effects/aliens.dm
Rohesie: Also this shouldn't be calling `Crossed()` in `process()`, but that's a different thing.
Both `Crossed()) and `process()` should be calling something like `acid_burn_feet()`

Rohesie:
Test and see if the behavior is appropriate. If so, let me know and I'll merge.

PR #3565, opened 2019-12-14
Only check larvasworth of a mob that has a job

psykzz:
```
Runtime in game_mode.dm, line 731: Cannot read null.larvaworth
proc name: get total joblarvaworth (/datum/game_mode/proc/get_total_joblarvaworth)
src: Crash (/datum/game_mode/crash)
call stack:
Crash (/datum/game_mode/crash): get total joblarvaworth(/list (/list), null)
Crash (/datum/game_mode/crash): balance scales()
Crash (/datum/game_mode/crash): process(2)
Ticker (/datum/controller/subsystem/ticker): fire(0)
Ticker (/datum/controller/subsystem/ticker): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop()
Master (/datum/controller/master): StartProcessing(0)
```

### DIFF
--- a/code/datums/gamemodes/game_mode.dm
+++ b/code/datums/gamemodes/game_mode.dm
@@ -719,7 +719,7 @@ Sensors indicate [numXenosShip ? "[numXenosShip]" : "no"] unknown lifeform signa
 
 	for(var/i in GLOB.human_mob_list)
 		var/mob/living/carbon/human/H = i
-		var/datum/job/job = SSjob.GetJob(H.mind ? H.mind.assigned_role : H.job)
+		var/datum/job/job = SSjob.GetJob(H.job)
 		if(!istype(job))
 			continue
 		if(H.stat == DEAD && !H.is_revivable())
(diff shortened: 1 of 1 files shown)


code/datums/gamemodes/game_mode.dm
```diff
@@ -719,7 +719,9 @@ Sensors indicate [numXenosShip ? "[numXenosShip]" : "no"] unknown lifeform signa
 
 	for(var/i in GLOB.human_mob_list)
 		var/mob/living/carbon/human/H = i
-		var/datum/job/job = SSjob.GetJob(H.job)
+		var/datum/job/job = SSjob.GetJob(H.mind ? H.mind.assigned_role : H.job)
```
Rohesie: We are moving away from `mind.assigned_role`, so try to avoid it.
Every job with a larvaworth should have a job referenced in the mob already.

psykzz: Got it, i was wondering why the job might be blank on some mobs, and instead of just skipping them trying extra hard to find one (by looking at `assigned_role`).

Will revert this.

code/datums/gamemodes/game_mode.dm
```diff
@@ -720,6 +720,8 @@ Sensors indicate [numXenosShip ? "[numXenosShip]" : "no"] unknown lifeform signa
 	for(var/i in GLOB.human_mob_list)
 		var/mob/living/carbon/human/H = i
 		var/datum/job/job = SSjob.GetJob(H.job)
+		if(!istype(job))
```
Rohesie:
```suggestion
		if(!job)
```
The proc should always return the correct type.

PR #3563, opened 2019-12-14
Adds global signal for dropships departing

spookydonut:
## About The Pull Request
For the FOB drones to hook

PR #3566, opened 2019-12-14
adds remote FOB construction

terranaut1:
## About The Pull Request

what it says on the tin

this is a grand theft of tg auxbase code and butchery of camera code to create an unholy frankenstein monster combining both
note: you need a newly added access level that only squad engineers and people with all access have to use the console, because fuck you specs, play something else for once.

materials need to be inserted into the computer like with a lathe, for sentry gun placement there's vouchers available you can buy at the engi prep vendor for the same cost as a regular sentry

the computer is located between the two dropship hangars

thanks to spooky for supplying the shuttle signals which i shamelessly yoinked like everything else
## Why It's Good For The Game

cuts down on setup time and gets everyone to the core gameplay faster

## Changelog
:cl:terranaut
add: added remote fob construction
add: added sentry vouchers for this purpose to engi prep vendor
/:cl:


code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+				if(istype(get_area(port), /area/shuttle/drop2/lz2))
+					spawn_spot = port
+					break
+		else
+			return
+	. = ..()
+	
+
+/obj/machinery/computer/camera_advanced/remote_fob/CreateEye()
+	if(spawn_spot)
```
psykzz: You should handle the case there isnt a spawn spot and offer feedback to the user

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+		return
+	if(!drone_creation_allowed)
+		to_chat(user, "<span class='notice'>Communication with the drone impossible due to fuel-residue in deployment zone atmosphere.</span>")
+		return
+	spawn_spot = FALSE
+	var/port
+	switch(tgalert(user, "Summon Drone in:", "FOB Construction Drone Control", "LZ1","LZ2"))
+		if("LZ1")
+			port = /obj/docking_port/stationary/marine_dropship/lz1
+			for(port in SSshuttle.stationary)
```
psykzz: use `locate(/obj/docking_port/stationary/marine_dropship/lz1) in SSshuttle.stationary` instead

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+
+/obj/machinery/computer/camera_advanced/remote_fob/attack_hand(mob/living/user)
+	if(!allowed(user))
+		to_chat(user, "<span class='warning'>Access Denied!</span>")
+		return
+	if(!drone_creation_allowed)
+		to_chat(user, "<span class='notice'>Communication with the drone impossible due to fuel-residue in deployment zone atmosphere.</span>")
+		return
+	spawn_spot = FALSE
+	var/port
```
psykzz:
```suggestion
	var/obj/docking_port/stationary/marine_dropship/port
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+	desc = "A computer console equipped with camera screen and controls for a planetside deployed construction drone. Materials or equipment vouchers can be added simply by inserting them into the computer."
+//	icon_screen = 
+//	icon_keyboard =
+	icon_state = "syndishuttle"
+	req_access = list(ACCESS_MARINE_REMOTEBUILD)
+	networks = FALSE
+	off_action = new/datum/action/innate/camera_off/remote_fob
+	jump_action = null
+	var/drone_creation_allowed = TRUE
+	var/spawn_spot
```
psykzz:
```suggestion
	var/obj/docking_port/stationary/marine_dropship/spawn_spot
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+		var/useamount = min(W.amount, (max_plasteel-plasteel_remaining))
+		plasteel_remaining += useamount
+		W.use(useamount)
+		to_chat(user, "<span class='notice'>Inserted [useamount] plateel sheets.")
+		return
+	else
+		to_chat(user, "<span class ='notice'>Invalid material type.</span>")
+	..()
+
+/obj/machinery/computer/camera_advanced/remote_fob/GrantActions(mob/living/user)
```
psykzz:
```suggestion
/obj/machinery/computer/camera_advanced/remote_fob/give_actions(mob/living/user)
```

psykzz: Also update references

psykzz: Don't forget a `remove_actions` proc

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+				break
+			else
+				to_chat(owner, "<span class='warning'>No space here for a barricade.</span>")
+				return
+	if(!do_after(fobdrone, 3 SECONDS, FALSE, buildplace, BUSY_ICON_BUILD))
+		return
+	console.plasteel_remaining -= 5
+	cade = new /obj/structure/barricade/plasteel(buildplace)
+	cade.setDir(fobdrone.dir)
+	cade.closed = 0
```
psykzz:
```suggestion
	cade.closed = FALSE
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+	cade.density = 1
```
psykzz:
```suggestion
	cade.density = TRUE
```

code/modules/shuttle/marine_dropship.dm
```diff
@@ -140,6 +140,12 @@
 	. = ..()
 	SSshuttle.dropships += src
 
+/obj/docking_port/mobile/marine_dropship/enterTransit()
+	. = ..()
+	if(!.) // it failed in parent
+		return
+	// pull the shuttle from datum/source, and state info from the shuttle itself
+	SEND_GLOBAL_SIGNAL(COMSIG_GLOB_DROPSHIP_TRANSIT)
```
psykzz: You should remove this as it was included in @spookydonut's pr #3563

code/modules/shuttle/shuttle.dm
```diff
@@ -466,6 +466,7 @@
 			WARNING("shuttle \"[id]\" could not enter transit space. Docked at [S0 ? S0.id : "null"]. Transit dock [S1 ? S1.id : "null"].")
 		else
 			previous = S0
+			return TRUE
```
psykzz: You should remove this as it was included in @spookydonut's pr #3563

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+			else
+				to_chat(owner, "<span class='warning'>No space here for a sentry.</span>")
+				return
+	if(!do_after(fobdrone, 6 SECONDS, FALSE, buildplace, BUSY_ICON_BUILD))
+		return
+	console.sentry_remaining -= 1
+	turret = new /obj/machinery/marine_turret(buildplace)
+	turret.setDir(fobdrone.dir)
+
+/obj/item/stack/sentryvoucher
```
psykzz:
```suggestion
/obj/item/stack/voucher/sentry
```

This would set up the future for other types of vouchers

psykzz: Make sure to update references.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+		console.metal_remaining -=2
+		cade.wire()
+		
+
+/datum/action/innate/remote_fob/toggle_wiring
+	name = "Toggle Razorwire"
+	action_icon_state = "wire"
+
+/datum/action/innate/remote_fob/toggle_wiring/Activate()
+	if(..())
```
psykzz:
```suggestion
    	. = ..()
	if(.)
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+
+	var/turf/buildplace = get_turf(fobdrone)
+	var/obj/structure/barricade/cade = /obj/structure/barricade
+	if(is_blocked_turf(buildplace))
+		for(var/obj/thing in buildplace)
+			if(istype(thing, cade) && thing.dir != fobdrone.dir)
+				break
+			else
+				to_chat(owner, "<span class='warning'>No space here for a barricade.</span>")
+				return
```
psykzz: Are you assuming there are more than one cade in this area?
If so, are you not going to output a bunch of messages to chat?

If not, can you not just use `locate()`

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
```
psykzz: Same as above

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+	.. ()
+	eyeobj.invisibility = INVISIBILITY_ABSTRACT
+	eyeobj.eye_initialized = FALSE
+	return PROCESS_KILL
+
+/obj/machinery/computer/camera_advanced/remote_fob/check_eye(mob/living/user)
+	if(!drone_creation_allowed)
+		to_chat(user, "<span class='notice'>Communication with the drone has been disrupted.</span>")
+		user.unset_interaction()
+	. = ..()
```
psykzz:
```suggestion
	return ..()
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+
+	if(sentry)
+		sentry.target = src
+		sentry.give_action(user)
+		actions += sentry
+
+	eyeobj.invisibility = 0
+
+/obj/machinery/computer/camera_advanced/remote_fob/remove_eye_control(mob/living/user)
+	.. ()
```
psykzz:
```suggestion
	..()
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+			to_chat(user, "<span class='notice'>Can't insert any more plasteel.")
+			return
+		var/useamount = min(W.amount, (max_plasteel-plasteel_remaining))
+		plasteel_remaining += useamount
+		W.use(useamount)
+		to_chat(user, "<span class='notice'>Inserted [useamount] plateel sheets.")
+		return
+	else
+		to_chat(user, "<span class ='notice'>Invalid material type.</span>")
+	..()
```
psykzz:
```suggestion
	return ..()
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+					break
+		if("LZ2")
+			port = /obj/docking_port/stationary/marine_dropship/lz2
+			for(port in SSshuttle.stationary)
+				if(istype(get_area(port), /area/shuttle/drop2/lz2))
+					spawn_spot = port
+					break
+		else
+			return
+	. = ..()
```
psykzz:
```suggestion
	return ..()
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+	user.see_in_dark = 7
+	return TRUE
+
+/////////////////////////////// the camera computer
+
+/obj/machinery/computer/camera_advanced/remote_fob
+	name = "FOB Construction Drone Control"
+	desc = "A computer console equipped with camera screen and controls for a planetside deployed construction drone. Materials or equipment vouchers can be added simply by inserting them into the computer."
+//	icon_screen = 
+//	icon_keyboard =
```
psykzz:
```suggestion
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+	user.lighting_alpha = LIGHTING_PLANE_ALPHA_NV_TRAIT
+	user.see_in_dark = 7
+	return TRUE
+
+/////////////////////////////// the camera computer
+
+/obj/machinery/computer/camera_advanced/remote_fob
+	name = "FOB Construction Drone Control"
+	desc = "A computer console equipped with camera screen and controls for a planetside deployed construction drone. Materials or equipment vouchers can be added simply by inserting them into the computer."
+//	icon_screen =
```
psykzz:
```suggestion
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+	var/area/curr_area = get_area(t)
+	if(istype(t, /turf/closed) || locate(/obj/machinery/door/poddoor) in t)
+		return
+	if(curr_area == starting_area || is_type_in_typecache(curr_area, GLOB.dropship_lzs))
+		return ..()
+
+
+/mob/camera/aiEye/remote/fobdrone/relaymove(mob/user, direct)
+	dir = direct //This camera eye is visible as a drone, and needs to keep the dir updated
+	..()
```
psykzz:
```suggestion
	return ..()
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+/datum/action/innate/camera_off/remote_fob
+	name = "Log out"
+
+/datum/action/innate/remote_fob/metal_cade
+	name = "Place Metal Barricade"
+	action_icon_state = "metal_cade"
+	
+
+/datum/action/innate/remote_fob/metal_cade/Activate()
+	if(..() || !check_spot())
```
psykzz:
```suggestion
	. = ..()
	if(. || !check_spot())
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,336 @@
...
+		return
+	console.do_wiring = !console.do_wiring
+	to_chat(owner, "<span class='notice'>Will now [console.do_wiring ? "do wiring" : "stop wiring"].</span>")
+
+/datum/action/innate/remote_fob/sentry
+	name = "Place Sentry"
+	action_icon_state = "sentry"
+
+/datum/action/innate/remote_fob/sentry/Activate()
+	if(..() || !check_spot())
```
psykzz:
```suggestion
	. = ..()
	if(. || !check_spot())
```

code/modules/mining/remote_fob.dm
terranaut1: we dont have give_actions. that's /tg/. we use grantactions. the same thing. removing actions is handled through parent.

code/modules/mining/remote_fob.dm
terranaut1: the case of no spawn spot being there is when attack_hand is called and the resulting popup window is closed rather than a selection being made. offering feedback for that seems useless.
the if check here is to prevent issues i've encountered during testing resulting in createeye being called to the last used spawn_spot, ie you summon a drone to lz1, cancel out, click on the computer, close the window instead of making a selection and instead it gives you a drone at lz1.

code/modules/mining/remote_fob.dm
terranaut1: i initially tried locate here which caused issues, as-is it allows to place 4 cades on the same turf as if you were hand-building them. it only outputs a single message per attempt if the turf wont allow a cade no matter how blocked it is.

code/modules/mining/remote_fob.dm
terranaut1: same as above

code/modules/mining/remote_fob.dm
psykzz: We seem to only use `GrantActions` in one place and its in eyecam code.
That said, `give_action` and `give_actions` seems clearer to me.

Also even if `give_actions` doesn't exist yet, its a good time to add it.

code/game/objects/structures/barricade.dm
```diff
@@ -122,19 +122,22 @@
...
+
+/obj/structure/barricade/proc/wire()
+	if(!closed)
+		wired_overlay = image('icons/Marine/barricades.dmi', icon_state = "[barricade_type]_wire")
+	else
+		wired_overlay = image('icons/Marine/barricades.dmi', icon_state = "[barricade_type]_closed_wire")
 
+	overlays += wired_overlay
+	modify_max_integrity(max_integrity + 50)
+	update_icon()
```
Rohesie: Put this at the end of the proc. It's the kind of thing that should go after the vars change, and should be useful for future refactors.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+/////////////////////////////// Drone Mob
+
+/mob/camera/aiEye/remote/fobdrone
+	name = "Remote Construction Drone"
+	mouse_opacity = MOUSE_OPACITY_OPAQUE
+	icon = 'icons/Marine/remotefob.dmi'
+	icon_state = "drone"
+	var/area/starting_area
+	use_static = FALSE
+	var/turf/spawnloc
```
Rohesie: Null reference variables such as `starting_area` and `spawnloc` on `Destroy()` even if you don't intend for the object to be destroyed.
Having loose references causes issues on garbage collecting, and things may change in the future on whether they are able to be deleted or not.

This is a personal preference, but I tend to place the redefined variables first, and then the ones defined for the object, for clarity. Up to you if you want to do it differently.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+
+/mob/camera/aiEye/remote/fobdrone/Initialize()
+	. = ..()
+	starting_area = get_area(loc)
+
+/mob/camera/aiEye/remote/fobdrone/setLoc(atom/t) //unrestricted movement inside the landing zone
```
Rohesie: You'll find a lot of old code using single-letter variables such as this one: `t`
Try to avoid replicating that, and instead give variables more informative names for ease of reading.
For example, here, something like `(atom/target)`

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+	var/area/curr_area = get_area(t)
+	if(istype(t, /turf/closed) || locate(/obj/machinery/door/poddoor) in t)
+		return
+	if(curr_area == starting_area || is_type_in_typecache(curr_area, GLOB.dropship_lzs))
+		return ..()
```
Rohesie: The logical progress of this proc is a series of checks to see if it should go through, and calling the parent if that is the case. In that sense it's a bit more clear if these checks are negated and the call to parent happens at the end, alternatively.
In practice:
```
if(curr_area != starting_area && !is_type_in_typecache(curr_area, GLOB.dropship_lzs))
	return
return ..()
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+
+
+/mob/camera/aiEye/remote/fobdrone/relaymove(mob/user, direct)
+	dir = direct //This camera eye is visible as a drone, and needs to keep the dir updated
```
Rohesie:
```suggestion
	setDir(direct) //This camera eye is visible as a drone, and needs to keep the dir updated
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+	desc = "A computer console equipped with camera screen and controls for a planetside deployed construction drone. Materials or equipment vouchers can be added simply by inserting them into the computer."
+	icon_state = "syndishuttle"
+	req_access = list(ACCESS_MARINE_REMOTEBUILD)
+	networks = FALSE
+	off_action = new/datum/action/innate/camera_off/remote_fob
+	jump_action = null
+	var/drone_creation_allowed = TRUE
+	var/obj/docking_port/stationary/marine_dropship/spawn_spot
+	var/lz
+	var/datum/action/innate/remote_fob/metal_cade/metal_cade = new
```
Rohesie: Assigning actions on the definition is generally necessary for a `static` type or the likes.
Not being the case, prefer doing it on `Initialize()` as that gives the MC finer control to improve the startup times.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+	RegisterSignal(SSdcs, COMSIG_GLOB_DROPSHIP_TRANSIT, .proc/disable_drone_creation)
+
+/obj/machinery/computer/camera_advanced/remote_fob/proc/disable_drone_creation()
+	drone_creation_allowed = FALSE
+	UnregisterSignal(SSdcs, COMSIG_GLOB_DROPSHIP_TRANSIT)
+
+/obj/machinery/computer/camera_advanced/remote_fob/examine(mob/user)
+	. = ..()
+	var/list/details = list()
+	details +=("It has [metal_remaining] sheets of metal remaining.</br>")
```
Rohesie:
```suggestion
	details += "It has [metal_remaining] sheets of metal remaining.</br>"
```
The parentheses are unneeded. This will be added as a string element to the  list already.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+	details +=("It has [plasteel_remaining] sheets of plasteel remaining.</br>")
+	details +=("It has [sentry_remaining] sentries ready for placement.</br>")
+	to_chat(user, "[details.Join(" ")] ")
```
Rohesie:
```suggestion
	to_chat(user, details.Join(" "))
```
Unless if you want to add a space at the end of it, `Join()` will already create a string output.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+
+/obj/machinery/computer/camera_advanced/remote_fob/attack_hand(mob/living/user)
+	if(!allowed(user))
+		to_chat(user, "<span class='warning'>Access Denied!</span>")
+		return
+	if(!drone_creation_allowed)
+		to_chat(user, "<span class='notice'>Communication with the drone impossible due to fuel-residue in deployment zone atmosphere.</span>")
+		return
+	spawn_spot = FALSE
+	switch(tgalert(user, "Summon Drone in:", "FOB Construction Drone Control", "LZ1","LZ2"))
```
Rohesie:
```suggestion
	switch(tgalert(user, "Summon Drone in:", "FOB Construction Drone Control", "LZ1", "LZ2", "Cancel"))
```
Options to return are nice to have.

code/modules/mining/remote_fob.dm
Rohesie: From what I gather if no selection is made then `attack_hand()` should return, not call parent, in current code.
If it's calling parent something is wrong.

From what you describe, you should consider nulling the `spawn_spot` at some point.

Prefer to use early returns, and to give feedback for behavior that is not supposed to happen. Silent errors are the bane of debugging.
In this case it would go:
```
if(!spawn_spot)
	CRASH("CreateEye() called without a spawn_spot designated")
```
That gives you a nice runtime and the call stack to retrace what went wrong.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+	return ..()
+	
+
+/obj/machinery/computer/camera_advanced/remote_fob/CreateEye()
+	if(spawn_spot)
+		eyeobj = new /mob/camera/aiEye/remote/fobdrone(get_turf(spawn_spot))
+		eyeobj.origin = src
+	return
+
+/obj/machinery/computer/camera_advanced/remote_fob/attackby(obj/item/stack/W, mob/user, params)
```
Rohesie:
```suggestion
/obj/machinery/computer/camera_advanced/remote_fob/attackby(obj/item/attacking_item, mob/user, params)
```
The base of `attackby` is with items. Do not worry about having to typecast it if you are sure that right now you'll only deal with stacks, but don't just retype it in the proc call, because else you are passing the wrong information to further coders that you can only get that kind of type of input on this proc.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+		if(max_plasteel >= plasteel_remaining)
+			to_chat(user, "<span class='notice'>Can't insert any more plasteel.")
+			return
+		var/useamount = min(W.amount, (max_plasteel-plasteel_remaining))
+		plasteel_remaining += useamount
+		W.use(useamount)
+		to_chat(user, "<span class='notice'>Inserted [useamount] plateel sheets.")
+		return
+	else
+		to_chat(user, "<span class ='notice'>Invalid material type.</span>")
```
Rohesie: The `else` is not needed here, as every `if` properly returns.
Do you want this message to happen always? Do consider if the user is, for example, bashing the machine with a weapon.

code/modules/mining/remote_fob.dm
terranaut1: this is done to access stack-specific stuff, notably in this case the amount var, if there's a better way to do that i'm all ears

code/modules/mining/remote_fob.dm
terranaut1: wont compile if i stuff them in initialize

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+		sentry.give_action(user)
+		actions += sentry
+
+	eyeobj.invisibility = 0
+
+/obj/machinery/computer/camera_advanced/remote_fob/remove_eye_control(mob/living/user)
+	..()
+	eyeobj.invisibility = INVISIBILITY_ABSTRACT
+	eyeobj.eye_initialized = FALSE
+	return PROCESS_KILL
```
Rohesie: Why is this value being returned, instead of the parent's? I don't see it being used anywhere.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+
+/obj/machinery/computer/camera_advanced/remote_fob/check_eye(mob/living/user)
+	if(!drone_creation_allowed)
+		to_chat(user, "<span class='notice'>Communication with the drone has been disrupted.</span>")
+		user.unset_interaction()
```
Rohesie:
```suggestion
		user.unset_interaction()
		return
```
No need for further checks, we've already unset the interaction.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+	return ..()
+
+/////////////////////////////// Placement Actions
+
+/datum/action/innate/remote_fob //Parent stuff
+	action_icon = 'icons/Marine/remotefob.dmi'
+	var/mob/living/C //the mob using the action
```
Rohesie: Do try to use more descriptive variable names, such as `user`, `remote_builder` or whatever.

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,329 @@
...
+	if(console.sentry_remaining < 1)
+		to_chat(owner, "<span class='warning'>You need to redeem a Sentry voucher to place one.</span>")
+		return
+	if(is_blocked_turf(buildplace))
+		for(var/obj/thing in buildplace)
+			if(istype(thing, cade))
+				break
+			else
+				to_chat(owner, "<span class='warning'>No space here for a sentry.</span>")
+				return
```
Rohesie: I see this repeated 3 times. Maybe make a `remote_fob` proc to avoid repeated code?

code/modules/shuttle/marine_dropship.dm
Rohesie: I recommended adding it here before that one is merged.
Personally I think it's better if it's added with functionality already, instead of solo.

code/modules/mining/remote_fob.dm
terranaut1: remains from testing other stuff

code/modules/mining/remote_fob.dm
terranaut1: the conditions are different for the sentry. sentries can be placed on any nonblocked turfs, aswell as turfs which are blocked exclusively by barricades

code/modules/mining/remote_fob.dm
terranaut1: spawn_spot is nulled every time before a selection is made precisely to prevent this kind of behavior, i encountered it during testing a bunch of times, aswell any other unforeseen behavior

code/modules/mining/remote_fob.dm
spookydonut: This is awful, do it properly in initialize

code/modules/mining/remote_fob.dm
spookydonut: this proc should be on the ai eye

spookydonut:
Drones don't use the garbage system correctly, this is DNM/DNTM until it does

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,345 @@
...
+
+/obj/machinery/computer/camera_advanced/remote_fob/give_eye_control(mob/user)
+	. = ..()
+	user.lighting_alpha = 120
+	eyeobj.name = "Remote Construction Drone"
+	if(eyeobj.eye_initialized)
+		eyeobj.setLoc(get_turf(spawn_spot))
+
+
+/obj/machinery/computer/camera_advanced/remote_fob/attack_hand(mob/living/user)
```
spookydonut: you can replace this with interact()

spookydonut:
sure whatever

code/modules/mining/remote_fob.dm
Rohesie: Alright. Then just do the early return here.

code/modules/mining/remote_fob.dm
Rohesie: How are you doing it on initialize? The correct way would be something like this:

```
/obj/machinery/computer/camera_advanced/remote_fob
	var/datum/action/innate/remote_fob/metal_cade/metal_cade

/obj/machinery/computer/camera_advanced/remote_fob/Initialize()
	. = ..()
	metal_cade = new()
```

code/modules/mining/remote_fob.dm
Rohesie: Ah, alright.

code/modules/mining/remote_fob.dm
Rohesie: Something like this:
```
/obj/machinery/computer/camera_advanced/remote_fob/attackby(obj/item/attacking_item, mob/user, params)
	if(istype(attacking_item, /obj/item/stack/voucher/sentry))
		var/obj/item/stack/attacking_stack = attacking_item
		var/useamount = attacking_stack .amount
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,356 @@
...
+			spawn_spot = locate(/obj/docking_port/stationary/marine_dropship/lz1) in SSshuttle.stationary	
+		if("LZ2")
+			spawn_spot = locate(/obj/docking_port/stationary/marine_dropship/lz2) in SSshuttle.stationary
+		else
+			return
+	return ..()
+
+/obj/machinery/computer/camera_advanced/remote_fob/CreateEye()
+	if(!spawn_spot)
+		return
```
Rohesie:
```suggestion
		CRASH("CreateEye() called without a spawn_spot designated")
```

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,356 @@
...
+	eyeobj = new /mob/camera/aiEye/remote/fobdrone(get_turf(spawn_spot))
+	eyeobj.origin = src
+	return
```
Rohesie:
```suggestion

```

code/modules/mining/remote_fob.dm
```diff
@@ -24,11 +24,11 @@ GLOBAL_LIST_INIT(dropship_lzs, typecacheof(list(/area/shuttle/drop1/lz1, /area/s
 	return ..()
 
 /mob/camera/aiEye/remote/fobdrone/setLoc(atom/target) //unrestricted movement inside the landing zone
-	var/area/curr_area = get_area(target)
-	if(istype(target, /turf/closed) || locate(/obj/machinery/door/poddoor) in target)
-		return
-	if(curr_area != starting_area && !is_type_in_typecache(curr_area, GLOB.dropship_lzs))
+	if(is_type_in_typecache(target, GLOB.blocked_remotebuild_turfs))
 		return
+	for(var/obj/thing in target)
```
Rohesie: Turn this into movable atoms, to make it generic.

_maps/map_files/LV624/LV624.dmm
```diff
@@ -10200,8 +10200,12 @@
...
 /area/lv624/ground/jungle6)
 "aNa" = (
-/obj/machinery/door/poddoor/timed_late/containment/landing_zone/lz2{
-	dir = 1
+/obj/machinery/door/airlock/mainship/engineering{
+	density = 0;
+	icon_state = "door_open";
+	name = "\improper Robotics Dome";
+	req_access_txt = "0";
+	req_one_access_txt = "0"
```
Rohesie: These var-edits are not ideal.
Make a child of the airlock with different `density`.
Also preefer `req_access` to `req_access_txt`, same for the other one.

_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
```diff
@@ -2520,6 +2520,13 @@
 	dir = 4
 	},
 /area/mainship/medical/medical_science)
+"iy" = (
+/obj/structure/bed/chair/comfy/black{
+	icon_state = "comfychair";
```
Rohesie:
```suggestion
```
This is a map editor issue. Just remove the line.

code/modules/mining/remote_fob.dm
psykzz: this comment is still relevant, unless you have a reason not to?

code/modules/mining/remote_fob.dm
```diff
@@ -0,0 +1,412 @@
+/////////For remote construction of FOB using a computer on the ship during setup phase
```
psykzz: If you get the time, its worth splitting this file into multiple for the eye / computer / actions.
The file is getting large enough that even git says its a large diff.

code/modules/mining/remote_fob.dm
terranaut1: interact causes a host of issues, trying to create an eye mob in i think nullspace and passing vision and actions to the user when you cancel out of the popup for starters, it's a headache and i dont see the benefit

AdamElTablawy:
it'd be nice if the AI could use this too.

code/modules/mining/remote_fob/remote_fob_computer.dm
```diff
@@ -50,13 +51,16 @@
...
 	. = ..()
+	icon_state = "fobpc-transfer"
 	user.lighting_alpha = 120
 	eyeobj.name = "Remote Construction Drone"
 	eyeobj.register_facedir_signals(user)
 	if(eyeobj.eye_initialized)
 		eyeobj.setLoc(get_turf(spawn_spot))
 	
 /obj/machinery/computer/camera_advanced/remote_fob/attack_hand(mob/living/user)
+	if(machine_stat & (NOPOWER|BROKEN))
```
psykzz: You should call parent here i believe if you check they do a lot of checks like if the user is concious and if the machine is operational etc.

Look into the parents.

terranaut1: they dont actually, i experimented with that before. it's just a long line of calling parent, all the way up to atom/movable which obviously doesnt have these kind of checks. i hadnt even noticed you could use this when broken/depowered before experimenting with the new sprites tbh.

code/modules/mining/remote_fob.dm
terranaut1: guess it works fine if you move stuff down the inheritance line to interact hee hoo

code/modules/mining/remote_fob/remote_fob_computer.dm
```diff
@@ -0,0 +1,178 @@
...
+	var/metal_remaining = 0
+	var/datum/action/innate/remote_fob/plast_cade/plast_cade
+	var/max_plasteel = 50
+	var/plasteel_remaining = 0
+	var/datum/action/innate/remote_fob/toggle_wiring/toggle_wiring //whether or not new barricades will be wired
+	var/do_wiring = FALSE
+	var/datum/action/innate/remote_fob/sentry/sentry
+	var/sentry_remaining = 0
+	var/datum/action/innate/remote_fob/eject_metal/eject_metal
+	var/datum/action/innate/remote_fob/eject_plasteel/eject_plasteel
```
Rohesie: Can you add a `Destroy()` proc to null or `QDEL_NULL` these several vars?

Rohesie:
Should be ready for merge after this change request, IMO.

code/modules/mining/remote_fob/remote_fob_computer.dm
```diff
@@ -0,0 +1,199 @@
...
+
+	RegisterSignal(SSdcs, COMSIG_GLOB_DROPSHIP_TRANSIT, .proc/disable_drone_creation)
+
+/obj/machinery/computer/camera_advanced/remote_fob/proc/disable_drone_creation()
+	drone_creation_allowed = FALSE
+	UnregisterSignal(SSdcs, COMSIG_GLOB_DROPSHIP_TRANSIT)
+
+
+/obj/machinery/computer/camera_advanced/remote_fob/Destroy()
+	drone_creation_allowed = null
```
Rohesie: This kind of variable needs not be nulled, it's not a reference to anything.

code/modules/mining/remote_fob/remote_fob_computer.dm
```diff
@@ -0,0 +1,199 @@
...
+	spawn_spot = null
+	lz = null
+	metal_cade = null
```
Rohesie:
```suggestion
	QDEL_NULL(metal_cade)
```
References to things that are made by the object and should be gone with it need to be qdeleted as well.
Same for the other ones created on Initialize

terranaut1: Can you elaborate further? Just for my understanding.

code/modules/mining/remote_fob/remote_fob_computer.dm
```diff
@@ -0,0 +1,198 @@
...
+
+/obj/machinery/computer/camera_advanced/remote_fob/proc/disable_drone_creation()
+	drone_creation_allowed = FALSE
+	UnregisterSignal(SSdcs, COMSIG_GLOB_DROPSHIP_TRANSIT)
+
+
+/obj/machinery/computer/camera_advanced/remote_fob/Destroy()
+	spawn_spot = null
+	lz = null
+	QDELL_NULL(metal_cade)
```
spookydonut:
```suggestion
	QDEL_NULL(metal_cade)
```
one L

Rohesie: Right, that's on me.

code/modules/mining/remote_fob/remote_fob_computer.dm
Rohesie: Of course.
The reason we null things on `Destroy` is to remove references to other existing objects so the GC processes things correctly.
Values like numbers or strings are not references to anything, so it doesn't make a difference if they get nulled.
Things we create on Initialize and that should be destroyed along with the thing that created should be deleted on the creator's `Destroy` proc, besides nulled. hence the `QDEL_NULL`

Rohesie:
Simple merge conflict, poke on discord for help resolving.

PR #3568, opened 2019-12-14
Set default larvaworth for base job

psykzz:
Fixes issue with admin spawned mobs not contributing to total larvaworth.

### DIFF
--- a/code/datums/jobs/job/job.dm
+++ b/code/datums/jobs/job/job.dm
@@ -50,7 +50,7 @@ GLOBAL_PROTECT(exp_specialmap)
 	var/display_order = JOB_DISPLAY_ORDER_DEFAULT
 	var/job_flags = NONE
 	
-	var/larvaworth = 0
+	var/larvaworth = LARVA_POINTS_REGULAR
 
 /datum/job/proc/after_spawn(mob/living/L, mob/M, latejoin = FALSE) //do actions on L but send messages to M as the key may not have been transferred_yet
 	if(!ishuman(L))


terranaut1:
do we want random mobs spawned for a variety of reasons to contribute towards the team worth though? it counts all keyed, non-ssd humans on planet and ship z iirc

Rohesie:
This could be an issue for testing admin mobs and whatnot.
I'd keep this value as zero, and manually add larvas if needed.

psykzz:
@terranaut1 generally yes, in the last round we admin spawned some mercs, but that didn't allow the xeno numbers to compensate for the extra 2 mobs.
@Rohesie I can add a check for centcom mobs and ignore them, if that helps?

terranaut1:
centcomm mobs are already ignored because it only counts ship and planet z
giving merc jobs values seems like the better solution, would handle some ERT interactions too, especially if you call balance_scales() (balance_the_scales()? i forgot what i called it and im too lazy to check) once after ERT are spawned

psykzz:
Perhaps adding a smaller amount to every mob then, and properly configuring others?
Right now i dont know what values for other jobs should be, so a blanket 3 points seemed good to me.

terranaut1:
3 for mobs that see combat is fine, thats in line with the old balance, we should just filter out mobs that are spawned somewhere but wont see combat or otherwise contribute to the marine team with supportive actions, ie. if admins spawn a human thats hostile to the marines they'd be punished twice for it

Rohesie:
I'm trying to handle every spawn through the job system with my refactor. And yes, that means xenos as well, and it implies removing hive burrowed larva in favor of xeno job slots open.

Perhaps we could extend this to this kind of admin spawned mobs, make them jobs and be joinable through their own panel/verb, handling themselves.

terranaut1:
Rather than Xeno job slots we should assign castes a point worth, but that's for discord discussion

psykzz:
Thats a good point, instead of larvaworth, maybe a value called "balance" and basically it allows new spawns of X based on the available difference?

Also its good to know how "balance" scales throughout the game, with people dying or spawning.

I could make some pretty graphs that show that.

Rohesie:
Mind closing this PR? It will conflict with the job refactor and cause issues on its new implementation.

psykzz:
Sure

PR #3569, opened 2019-12-14
Update power usage

psykzz:
## About The Pull Request

This improves the code quality for power and enables power usage for a bunch of items across the board.

## Changelog
:cl:
tweak: power across all the things
/:cl:

### DIFF
--- a/code/game/objects/machinery.dm
+++ b/code/game/objects/machinery.dm
@@ -29,6 +29,7 @@
 	. = ..()
 	GLOB.machines += src
 	component_parts = list()
+	return INITIALIZE_HINT_LATELOAD
 
 /obj/machinery/LateInitialize(mapload)
 	. = ..()
--- a/code/game/objects/machinery.dm
+++ b/code/game/objects/machinery.dm
@@ -29,7 +29,8 @@
 	. = ..()
 	GLOB.machines += src
 	component_parts = list()
-	START_PROCESSING(SSmachines, src)
+	if(use_power >= IDLE_POWER_USE)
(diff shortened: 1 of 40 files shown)


code/game/objects/machinery.dm
```diff
@@ -29,6 +29,7 @@
 	. = ..()
 	GLOB.machines += src
 	component_parts = list()
+	START_PROCESSING(SSmachines, src)
```
Rohesie: Does every machine need to process on init?

Rohesie:
There are several improvements here. My only concern is for all machines to be added to processing on init. I know /tg/ does this, but it's really not ideal.
Maybe a flag could be added to signify if the machine should be always processing, or the likes.

code/game/objects/machinery.dm
psykzz: If we want a machine to consume power we do.
If a machine doesn't it will be removed on the first tick.

Rohesie: Now that I think it a bit more, maybe starting either in LateInitialize or after the game mode starts might be the best here. The latter would be ideal IMO, but out of scope. Push it to late init?

psykzz: Sure

code/game/objects/machinery.dm
```diff
@@ -30,6 +30,11 @@
 	GLOB.machines += src
 	component_parts = list()
```
Rohesie:
```suggestion
	component_parts = list()
	return INITIALIZE_HINT_LATELOAD
```

PR #3570, opened 2019-12-15
Autodoc balancing

psykzz:
## About The Pull Request

Bunch of different power balances
This increases, massively, the power cost of the autodoc.
Right now if you use the autodoc with the generator and smes charged and active, you will not really noticed the impact.

However as soon as the generator / smes get empty, you get around 1 or 2 surguries before you run out of power.

## Why It's Good For The Game

Autodoc is very powerful, and this reflects that.

## Changelog
:cl:
balance: Increased autodocs power consumption, reduced max charge of smes.
/:cl:


psykzz:
Tested this out ingame via varediting, and in heavy combat (lots of injured) we saw the apc / smes staying regularly around half charge, slowly building up

image

psykzz:
We saw about a 2% drain from SMES per surgury.

jroinc:
So, the SMES can currently output a maximum of 200KW. With the SMES set to that, the autodoc can do surgeries without draining APC power at all.

Additionally, it takes ~15% of one fuel cell to up the SMES from 20% to 100% (400000W)

I recommend removing the fuel cell recycler and spare fuel cells, and putting a few spare ones around the map in places only marines/engis can reach. Or just make them unacidable.

jroinc:
As-is, one surgery drains about 10-15% from the SMES with the generator on, and about 3x that amount with it off.
The APC cell can also support exactly one surgery by itself, but if it runs out, it shuts the autodoc down for about 30 seconds.

It takes about 1 second/percent for the generator to charge the SMES.

Shadow-Quill:
Imagine using the autodoc for anything but simple(quik) surgeries or blood transfusions /s

jroinc:
So after looking at it during the night, I think that giving medical and the cockpit roughly a 10k drain each passively, multiplying the generator fuel usage by 5x, and removing all the spare fuel cells BUT leaving the recharger's the best way to go.

That way, power lasts about 5-7 min without the generator, the generator lasts 20 minutes, and recharging it takes 2 steps, and about 2 minutes.

Rohesie:
Do you feel this PR to be ready, balance-wise?

psykzz:
It will definitely cause some issues for marines if its merged, i dont actually know if power will be able to sustatin it for the whole round.

PR #3572, opened 2019-12-15
Update flame turfs for incen grenades

psykzz:
## About The Pull Request

Adds a new pattern taken from how we generate weed nodes.
Added a bit of customization to support circle and square sets.

This respects blocking tiles and wont continue if the tile is blocked / dense / whatever

## Why It's Good For The Game

Stops incens from destroying things on the other side of a barrier.

## Changelog
:cl:
tweak: how flame graphs are generated, they are now blocked by dense objects
/:cl:

As a note, this is kinda nerf to the grenades, so im thinking about increasing the radius by 1

video : https://streamable.com/0lc5k

### DIFF
--- a/code/__HELPERS/patterns.dm
+++ b/code/__HELPERS/patterns.dm
@@ -72,7 +72,8 @@
 				if (AdjT in results) // Ignore existing turfs
 					continue
 				if(AdjT.density || LinkBlocked(T, AdjT) || TurfBlockedNonWindow(AdjT))
-					results += AdjT
+					if(include_edge)
+						results += AdjT
 					continue
 
 				turfs_to_check += AdjT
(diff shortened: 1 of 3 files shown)


code/__HELPERS/patterns.dm
```diff
@@ -43,3 +43,39 @@
...
+			directions = GLOB.alldirs
+		if("circle")
+			directions = GLOB.cardinals
+
+	var/list/results = list()
+	var/list/turfs_to_check = list()
+	turfs_to_check += center_turf
+	var/node_size = radius
+	while (node_size > 0)
+		node_size--
```
Rohesie: `for(var/i = node_size ; node_size > 0 ; node_size--)`

terranaut1:
I like this but maybe make it ignore barriers with throwpass
Not sure if waist high barriers should block an incendiary nade

PR #3584, opened 2019-12-19
makes m4ra actually not burst & sets burst_amounts to 1 

factoryman942:
:cl:
balance: actually removes the scout rifle's burst
/:cl:

#3553 removed GUN_FIREMODE_BURSTFIRE from the gun, and removed "burst_amount = 2" from it. This caused it to inherit burst amount of 3 from gun/rifle, effectively giving it a 3 shot burst instead of 2.
It also increased the recoil _of the gun above it_, the Dragunov. 
Besides rectifying these, this sets some burst_amount s from 0 to 1 (as the default /gun has), meaning the guns will work correctly with BFA. None of these guns can take a BFA so that doesn't matter much, but still. 

Test your PRs.


code/modules/projectiles/guns/rifles.dm
```diff
@@ -176,7 +176,7 @@
 	fire_delay = 1 SECONDS
 	accuracy_mult = 1.25
 	scatter = -15
-	burst_amount = 0
+	burst_amount = 1
```
Rohesie: If the parent has a burst amount of `1`, then remove the line. Same for the rest.

factoryman942: parent has 3; their granddad has 1

Rohesie: Then it's fair.

PR #3585, opened 2019-12-19
tip changes

factoryman942:
I have altered the tips. Pray I do not alter them further.

stop-drop-roll tip removed because there's already one in meta tips.


factoryman942:
i think travis hit his head

PR #3592, opened 2019-12-20
New mob: Xeno Hivemind

psykzz:
## About The Pull Request

Adds a new mob the xeno hivemind. 
When spawned the hivemind creates a node, and is only able to move along weeds.
They level up like normal xeno, gaining more speed, plasma, and abilities.

At early ranks they are only able to plant weeds.
Later they can build structures and eventually provide pheros.

I want to expand it with two new abilities (still in progress)
 - Track xenos on weeds 
 - Track/Be notified of humans on weeds

## Why It's Good For The Game

Adds the xeno equivilant of the AI.

## Changelog
:cl:
add: the xeno hivemind. The mind is a mob spawned by the queen. It starts off weak with low plasma and only able to spawn weeds. Over time it can heal, place structures and give pheros.
/:cl:

image

image

## Issues to resolve

- The weed processing gets a datum/action entry somehow
- Density is reset at some point
- Sometimes the hivemind can't move
- When SSD the hivemind goes to sleep.


Rohesie:
Very interesting concept. How would it be destroyed?

jjpark-kb:
It could be almost exactly like an AI with a core, and get destroyed when its core is destroyed. Maybe a shrike/queen could build it so you don't need to modify any maps / prevents meta-gaming. 
I really like the concept...

psykzz:
Queen spawning it in as a "core" could work well and also be a good limit (becaue it would require a queen)

code/game/objects/effects/weeds.dm
```diff
@@ -159,4 +159,33 @@
...
+	max_integrity = 600
+
+	var/mob/living/carbon/xenomorph/hivemind/parent
+
+/obj/effect/alien/weeds/node/strong/hivemindcore/Initialize(mapload, mob/living/carbon/xenomorph/X)
+	. = ..(mapload, src)
+	parent = X
+
+/obj/effect/alien/weeds/node/strong/hivemindcore/Destroy()
+	parent.playsound_local(parent, get_sfx("alien_help"), 30, 1)
```
Rohesie:
```suggestion
	parent.playsound_local(parent, get_sfx("alien_help"), 30, TRUE)
```

code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -160,6 +160,9 @@
 
 	var/wait_time = 2 SECONDS - max(0, (1 SECONDS * X.health / X.maxHealth)) //Between 1 and 2 seconds, depending on health.
 
+	if(isxenohivemind(X)) // Hivemind has free build, it should be slower
```
Rohesie: Could you turn the `wait_time` base into an ability var, and give the hivemind a child with a higher one?
Don't think the hivemind building speed according to health is as relevant as it is for a xeno.

code/modules/mob/living/silicon/ai/freelook/eye.dm
```diff
@@ -93,7 +93,7 @@
 	if(istype(ai.current, /obj/machinery/holopad))
 		var/obj/machinery/holopad/H = ai.current
 		H.move_hologram(ai, T)
-	if(ai.camera_light_on)
+	if(ai.camera_light_on)
```
Rohesie:
```suggestion
	if(ai.camera_light_on)
```

jroinc:
What happens if the "Camera" gets trapped on an isolated weed section/the weed section it's on is destroyed?

psykzz:
@jroinc it gets teleported to some random weeds.

adamkad1:
is queen going to be a combat viable xeno now?

jroinc:
I can see this being VERY easy to get trapped as (people notice the one area with self-building weeds and just trim it back as needed with one guy/people cut a path so wide it'll take 20 min to weed through if it wasn't being watched.)

I'd recommend an ability where for the cost of ALL your plasma, you can teleport back to your core/to a random weed node/anywhere with weeds onscreen, whichever you think is more balanced.

Would also recommend an ability where you can "jump" to xenos as long as they're on weeds.
Otherwise, ungas will just track the unbroken trail of weeds back to the stationary target that can't defend itself and torch it's ass within 5 minutes of landing.

Also, otherwise, it can't get across the river on LV.

psykzz:
You can already jump to any hive member assuming they spoke and are on weeds.

jroinc:
Ok, nevermind, that covers both issues I had.

code/game/objects/effects/weeds.dm
```diff
@@ -147,16 +143,62 @@
...
+/obj/effect/alien/weeds/node/strong
+	name = "strong purple sac"
+	desc = "A weird, pulsating node. This looks pretty tough."
+	node_range = NODERANGE*2
+	max_integrity = 120
+
+
+// =================
+// hivemind core
+/obj/effect/alien/weeds/node/strong/hivemindcore
```
Rohesie: Does this need to be a child of `strong` ?

code/game/objects/machinery/sentries.dm
```diff
@@ -980,7 +980,7 @@
 	var/mob/living/M
 
 	for(M in oview(range, src))
-		if(M.stat == DEAD) //No dead or robots.
+		if(M.stat == DEAD || isxenohivemind(M)) //No dead or robots.
```
Rohesie: Instead of an istype check, some sort of trait or flag would be better.

code/game/turfs/turf.dm
```diff
@@ -533,6 +533,9 @@ GLOBAL_LIST_INIT(unweedable_areas, typecacheof(list(
 			else
 				has_obstacle = TRUE
 				break
+		if(istype(O, /obj/effect/alien/weeds/node/strong/hivemindcore))
```
Rohesie: I'm on the opinion that we should up the status of weed nodes in general and not allow constructions over them.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,103 @@
+/mob/living/carbon/xenomorph/hivemind
+	caste_base_type = /mob/living/carbon/xenomorph/hivemind
+	name = "hivemind"
+	real_name = "hivemind"
```
Rohesie: Names of other xenos are capitalized, no?

adamkad1:
was making it immobile a good idea tho?

TiviPlus:
Seeing as heal is moved from queen to the Hivemind it arguably makes the Queen obsolete since the queen can now only Screech and have strong Pheros, making the Shrike a better xeno than the queen.

jjpark-kb:
I decided to try and make a sprite for it... I'm not really good at sprite making but hopefully it works? Maybe even ideas to those who will sprite it.

weed_hivemind

[weeds.zip](https://github.com/tgstation/TerraGov-Marine-Corps/files/3991904/weeds.zip)
In this is the dmi of the weeds, with the addition of what I have labeled the "weed_hivemind"

psykzz:
Thanks @jjpark-kb ill give it a try and let you know.

psykzz:
image

code/__DEFINES/mobs.dm
```diff
@@ -128,6 +128,7 @@
 #define TK_USER			(1<<8)
 #define CANUNCONSCIOUS	(1<<9)
 #define CANCONFUSE		(1<<10)
+#define UNDETECTABLE	(1<<10) // Whether not this unit should be detectable by automated means (like turrets). Used by hivemind
```
Rohesie:
```suggestion
#define UNDETECTABLE	(1<<11) // Whether not this unit should be detectable by automated means (like turrets). Used by hivemind
```
Maybe something like `INVISIBLE` ?
Undetectable is a bit broader.

psykzz: I had thought about that, but undetectable is what i wanted to go with, for example maybe having a xeno that is completely visible but can't be detected on via computers.

Rohesie: Then `UNDETECTABLE_BY_MACHINE_`, then?
The flag value needs to change, to be different than `CANCONFUSE`

Rohesie:
Mood.

jjpark-kb:
you can grab people, and then proceed to eat them

jjpark-kb:
It would be interesting if you allowed ancient shrike to place the hivemind core. With how population goes on the server, plus the behavior of most xenos during crash, it would be nice to have some hivemind play. I have to say that it is less than thrilling having to be ghost and wait for the position to open up, if it will ever open up.

jroinc:
I'd vote for letting ancient shrike place/remove (but not behind fog, as that's an unkillable xeno).

Also grabdevour memes are pretty funny and don't happen if the unga has half-a-brain but should probably be removed.

~~I spawned in 3, and one devoured while 2 spam-grabbed to prevent escape~~

They also set off landmines.

psykzz:
I removed grab, so no more devour either.
I'll add that if a xeno places the hivemind the fog is removed.

jjpark-kb:
The hivemind should not take up xeno slots. There occurs times when all the xenos die and the only one left is the hivemind with no burrowed. 
Also, the hivemind should reach ancient a little faster. Hivemind just feels extremely counterproductive to make when it takes so long to reach ancient plus the removing fog. Who would spawn it down when they know the fog will be removed, potentially removing a spawn point.

jroinc:
I've played as it some, and I'd like to see some changes.
Lower the weed time base to 10 seconds from 12 and knock 2 seconds off the timer each age category. 12 seconds is a while to wait when you have ZERO other abilities.

Heal feels fine, though.

Pheremones feel fine, but up the plasma cost for them to the point where it's like 2/3 your plasma regen.

Lower the build timer for resin structures from 7.5 seconds to 2.5 seconds, but add a 5 second delay before a new structure can be built, so you can move around some while trying to build stuff.

Lower plasma generation and storage slightly. Once you hit mature, you have no issues with plasma because you can use abilities so rarely.

Add in a warning either to the hivemind, or to all xenos, that the hivemind core is taking damage. I've seen 3x hiveminds drop dead out of thin air cause they couldn't tell they were taking damage.


I've let players use ancient hivemind with 6 second weeding and 2.5 second build 3-4 rounds now, and that seems both reasonably fun to play and balanced for ungas.

Firefox13:
Hivemind _eye_ can be attacked and you get splashed by acid blood.
And marines can hear talking hivemind _eye_.

Hathkar:
A couple bugs I've noticed when a hivemind was in the game.

When the hivemind went SSD, weeds no longer spread around when planted. The shrike's healing ability also healed for 0.

Also, when joining the game by taking over a SSD hivemind, the hivemind is in a resting state as per usual SSD conventions. However, the hivemind can't get up to use it's abilities or move around. There's no hud button to get up, and nothing in the IC tab to get up.

psykzz:
Thanks for the feedback, Hivemind has been since removed.
I'll work on looking at the interactions with the SSD hivemind, as i feel its a likely cause for the issues we've been having.

nemvar:
You know what would be a cool ability? If it could create a shroud of darkness on its location. The lighting support negative lights and it isn't used anywhere.

nemvar:
Also why are you not using a camera mob?

psykzz:
@nemvar The issue with a cameramob was being tied to the cameranet and losing the xeno caste.
I made the decision to just recreate it like a blob, and keep it a subclass of a xeno.

Rohesie:
What's the current status of the feature?

code/_onclick/xeno.dm
```diff
@@ -21,6 +21,16 @@
...
+
+
+
+/mob/living/carbon/xenomorph/hivemind/UnarmedAttack(atom/A, proximity_flag)
+	if(lying)
+		return FALSE
+
+	A.attack_hivemind(src)
+
+/atom/proc/attack_hivemind(mob/living/carbon/xenomorph/hivemind/L)
```
Rohesie:
```suggestion
/atom/proc/attack_hivemind(mob/living/carbon/xenomorph/hivemind/hivemind_attacker)
```

code/game/objects/effects/weeds.dm
```diff
@@ -145,18 +141,79 @@
...
+	X.visible_message("<span class='danger'>[X] nudges its head against [src].</span>", \
+	"<span class='danger'>You nudge your head against [src].</span>")
+
+/obj/effect/alien/weeds/node/hivemindcore/take_damage(damage_amount, damage_type, damage_flag, sound_effect, attack_dir, armour_penetration)
+	. = ..()
+	if(isnull(parent))
+		return
+	if(prob(60))
+		return	
+	var/percent = round((max_integrity / obj_integrity) * 100)
```
Rohesie:
```suggestion
	var/health_percent = round((max_integrity / obj_integrity) * 100)
```

code/game/objects/effects/weeds.dm
```diff
@@ -145,18 +141,79 @@
...
+	switch(percent)
```
Rohesie:
```suggestion
	switch(health_percent)
```

code/game/objects/items/shards.dm
```diff
@@ -69,26 +69,28 @@
...
-		var/mob/living/M = AM
-		playsound(src.loc, 'sound/effects/glass_step.ogg', 25, 1) // not sure how to handle metal shards with sounds
-		if(!M.buckled)
-			to_chat(M, "<span class='danger'>[isxeno(M) ? "We" : "You"] step on \the [src]!</span>")
-			if(ishuman(M))
-				var/mob/living/carbon/human/H = M
-
-				if(H.species.species_flags & IS_SYNTHETIC || H.species.insulated)
+/obj/item/shard/Crossed(AM)
+	if(!isliving(AM) || isxenohivemind(AM))
```
Rohesie: Maybe an incorporeal flag instead?

code/game/turfs/turf.dm
```diff
@@ -533,6 +533,9 @@ GLOBAL_LIST_INIT(unweedable_areas, typecacheof(list(
 			else
 				has_obstacle = TRUE
 				break
+		if(istype(O, /obj/effect/alien/weeds/node))
```
Rohesie: This proc would also benefit from either a redefined proc to check the objects, or a flag.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,126 @@
...
+	hud_possible = list(PLASMA_HUD, PHEROMONE_HUD, QUEEN_OVERWATCH_HUD)
+
+	var/mob/living/carbon/xenomorph/observed_xeno
+
+/mob/living/carbon/xenomorph/hivemind/Initialize(mapload, ...)
+	. = ..()
+	GLOB.hivemind_list += src
+
+	// Remove some unused xeno things
+	verbs -= /mob/living/proc/lay_down
```
Rohesie: Could you turn the verb addition into a proc to be redefined at the hivemind level, instead of adding and removing it?

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,126 @@
...
+	if(!W)
+		var/obj/effect/alien/weeds/nearby = locate() in range("3x3", loc)
+		if(!nearby)
+			forceMove(get_turf(locate(/obj/effect/alien/weeds)))
+		return FALSE
+
+	// Don't allow them over the timed_late doors 
+	var/obj/machinery/door/poddoor/timed_late/door = locate() in NewLoc
+	if(door?.density)
+		return FALSE
```
Rohesie: This is pretty specific. Something like a `BLOCKS_LIVING` flag might be better.

Rohesie: Maybe we should add a specific blocker instead.
Or block `living` at the level of `/obj/machinery/door/poddoor/timed_late/door/CanPass(atom/movable/mover, turf/target)` instead

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,126 @@
...
+/mob/living/carbon/xenomorph/hivemind/AltClickOn(atom/A)
+	return FALSE
+
+/mob/living/carbon/xenomorph/hivemind/UnarmedAttack(atom/A, proximity_flag)
+	a_intent = INTENT_HELP //Forces help intent for all interactions.
+	return ..()
+
+/mob/living/carbon/xenomorph/hivemind/RangedAttack(atom/A, params)
+	a_intent = INTENT_HELP //Forces help intent for all interactions.
+	return ..()
```
Rohesie: Redefine the proc to change intent instead, so it starts in `INTENT_HELP` and never changes.

code/_globalvars/lists/mobs.dm
```diff
@@ -28,6 +28,7 @@ GLOBAL_LIST_EMPTY(alive_living_list)		//all alive /mob/living, including clientl
 GLOBAL_LIST_EMPTY(carbon_list)				//all instances of /mob/living/carbon and subtypes, notably does not contain brains or simple animals
 GLOBAL_LIST_EMPTY(offered_mob_list)				//all /mobs offered by admins
 GLOBAL_LIST_EMPTY(ai_list)
+GLOBAL_LIST_EMPTY(hivemind_list)
```
Rohesie: This doesn't seem to be used anymore.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
psykzz: Im defining my own move proc here, similar to blob. That wouldn't work.

code/_globalvars/lists/mobs.dm
psykzz: I use it to keep track of the hivemind in its Init and Destroy procs

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
psykzz: Could you explain this a little further?
I took this example from the larva.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,188 @@
...
+/mob/living/carbon/xenomorph/hivemind/ShiftClickOn(atom/A)
+	return FALSE
+
+/mob/living/carbon/xenomorph/hivemind/CtrlClickOn(atom/A)
+	return FALSE
+
+/mob/living/carbon/xenomorph/hivemind/AltClickOn(atom/A)
+	return FALSE
+
+/mob/living/carbon/xenomorph/hivemind/a_intent_change(new_intent)
```
Rohesie:
```suggestion
/mob/living/carbon/xenomorph/hivemind/a_intent_change()
```

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,188 @@
...
+	return FALSE //Unable to change intent, forced help intent
```
Rohesie:
```suggestion
	return //Unable to change intent, forced help intent
```

code/modules/mob/living/carbon/xenomorph/castes/larva/larva.dm
```diff
@@ -21,9 +24,8 @@
 // ***************************************
 // *********** Mob overrides
 // ***************************************
-/mob/living/carbon/xenomorph/larva/UnarmedAttack(atom/A)
-	a_intent = INTENT_HELP //Forces help intent for all interactions.
-	. = ..()
+/mob/living/carbon/xenomorph/larva/a_intent_change(new_intent)
```
Rohesie:
```suggestion
/mob/living/carbon/xenomorph/larva/a_intent_change()
```

code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -827,6 +846,51 @@
...
+		"<span class='xenowarning'>We have created a hivemind!</span>")
+
+	var/obj/effect/alien/weeds/node/hivemindcore/core = new /obj/effect/alien/weeds/node/hivemindcore(current_turf)
+	if(!isxenohivemind(core.parent))
+		return FALSE
+	var/mob/living/carbon/xenomorph/hivemind/xeno_hivemind = core.parent
+	playsound(xeno_hivemind.loc, 'sound/effects/splat.ogg', 25)
+
+	xeno_hivemind.offer_mob()
+	SEND_GLOBAL_SIGNAL(COMSIG_GLOB_OPEN_TIMED_SHUTTERS_XENO_HIVEMIND)
```
Rohesie: Why does the creation of the hivemind open the shutters?

code/_globalvars/lists/mobs.dm
Rohesie: Yea, but what for?

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
Rohesie: This one is still a thing.

code/modules/mob/living/carbon/xenomorph/abilities.dm
psykzz: because the balance issue that people were complaining you could place a hivemind behind the fog / shutters. They wanted to be able to take it out, so we agreed to open shutters / remove fog.

code/_globalvars/lists/mobs.dm
psykzz: We track over mobs, it only makes sense no?

code/modules/mob/living/carbon/xenomorph/abilities.dm
Rohesie: They shouldn't be able to put a hivemind behind shutters, as xenos can't access that area, right?

On fog, that's a bit more problematic. How are you handling it?
First solution that comes to my mind is, if the game mode is one with permafog, runnig a locate fog in range when checking if the hivemind can be placed.

psykzz: I was just removing the fog.
Xenos are opting for a hivemind with the downside of no fog.
Saved doing anything weird /  complicated for it.

code/_globalvars/lists/mobs.dm
Rohesie: If we are not using the list anywhere, then it's legacy that should be removed, really.
We already track xenos through the hive datum.

code/modules/mob/living/carbon/xenomorph/abilities.dm
Rohesie: This signal is raising the shutters, not removing the fog.

code/_onclick/xeno.dm
```diff
@@ -21,6 +21,16 @@
...
 
-
 /atom/proc/attack_larva(mob/living/carbon/xenomorph/larva/L)
 	return
+
+
+
+/mob/living/carbon/xenomorph/hivemind/UnarmedAttack(atom/A, proximity_flag)
+	if(lying)
+		return FALSE
```
Rohesie: Hivemind can never be lying, right?

code/game/objects/effects/weeds.dm
```diff
@@ -145,18 +143,79 @@
...
+	node_icon = "weed_hivemind4"
+	var/mob/living/carbon/xenomorph/hivemind/parent
+
+/obj/effect/alien/weeds/node/hivemindcore/Initialize(mapload, obj/effect/alien/weeds/node/node)
+	. = ..()
+	parent = new(loc)
+	set_light(7, 5, LIGHT_COLOR_PURPLE)
+
+/obj/effect/alien/weeds/node/hivemindcore/Destroy()
+	set_light(0, 0, null)
```
Rohesie: Aren't lights properly removed on Destroy by default?

code/_onclick/xeno.dm
```diff
@@ -21,6 +21,16 @@
...
 	A.attack_larva(src)
 
-
 /atom/proc/attack_larva(mob/living/carbon/xenomorph/larva/L)
 	return
+
+
+
+/mob/living/carbon/xenomorph/hivemind/UnarmedAttack(atom/A, proximity_flag)
+	if(lying)
```
Rohesie: Reminder for myself to change this lying to `handblock_flags` in the mobility refactor.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,188 @@
...
+	GLOB.hivemind_list -= src
+	return ..()
+
+/mob/living/carbon/xenomorph/hivemind/Move(NewLoc, Dir = 0)
+	var/obj/effect/alien/weeds/W = locate() in range("3x3", NewLoc)
+	if(!W)
+		var/obj/effect/alien/weeds/nearby = locate() in range("3x3", loc)
+		if(!nearby)
+			// If we run out of weeds just teleport to some random weeds.
+			forceMove(get_turf(locate(/obj/effect/alien/weeds)))
```
Rohesie: The ideal method to implement this is for weeds and fire to modify flags on the turf they spawn in, and to check for those flags on the turf, I believe.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -0,0 +1,188 @@
...
+	hud_possible = list(PLASMA_HUD, PHEROMONE_HUD, QUEEN_OVERWATCH_HUD)
+
+	var/mob/living/carbon/xenomorph/observed_xeno
+
+/mob/living/carbon/xenomorph/hivemind/Initialize(mapload, ...)
+	. = ..()
+	GLOB.hivemind_list += src
+
+	// Remove some unused xeno things
+	verbs -= /mob/living/proc/lay_down
```
Rohesie:
```suggestion
```

code/game/objects/effects/weeds.dm
psykzz: Not that i could see

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
psykzz: > The ideal method to implement this is for weeds and fire to modify flags on the turf they spawn in, and to check for those flags on the turf, I believe

I dont mind this, do you mind doing a sample of how you'd like turf flags to be like?

code/game/objects/effects/weeds.dm
Rohesie: It's pretty obscure but we have this:
```
/atom/Destroy()
	QDEL_NULL(light)
```
which leads to this:
```
/datum/light_source/Destroy(force)
```
So lights already get removed properly on Destroy.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
Rohesie: You don't have to do it, but I'd add a `var/turf_flags = NONE` variable to the base of turfs, and then do some wrapper procs to add and remove the turf traits. Something like this:
```
#define TURF_WEEDED (1<<0)

/turf/proc/set_weeds(weed_state)
  if(weed_state)
    if((turf_flags & TURF_WEEDED))
      return
    turf_flags |= TURF_WEEDED
    //room for changes when weeding happen
    return
  if(!(turf_flags & TURF_WEEDED))
    return
  turf_flags &= ~TURF_WEEDED
  //room for changes when de-weeding happens
```
The same logic would apply to fires and other turf phenomena.
What do you think, @ninjanomnom ?

ninjanomnom: This is one of the harder problems to solve in a good way, what you're suggesting is a form of spaghetti code imo but there aren't many better options if you want this exact behavior.

If you're willing to make the graphic for it though you can tweak how the mechanic works to make the code much cleaner. Just prevent going too far away from the last touched weed and show some tether on the screen. This way you only need to check distance to a single weed every move rather than either track every legal turf or range() check every move.

code/game/objects/effects/weeds.dm
psykzz: Alright will remove this.

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
Rohesie: I'm thinking on it a bit broader. We don't have atmos fire, but we have fire and weed objects, which are found with `locate() in turf` each cycle. Wouldn't it make more sense to treat it as a turf property, if it's on fire or weeded?

ninjanomnom: Fire is a bit more broad but something weeded is just a turf with a weed on it, that's not got anything to do with the turf really and is spaghetti due to it being code unrelated to the turf.

Rohesie:
It seems to be conflicting with the other TMs. How confident are you in this for a merge?

psykzz:
Let me fix the shutters / fog, otherwise it should be good and can simply be just not used / removed if its breaking rounds.

psykzz:
Should be good to merge @Rohesie

PR #3593, opened 2019-12-20
Buff surgery based on skills

psykzz:
## About The Pull Request

Autodoc changes along with unskilled corpsman take almost 15 seconds for every action. This reduces that down to a more reasonable number

## Why It's Good For The Game

I dont want to remove the autodoc changes just yet.
Instead buffing manual surgeries.

## Changelog
:cl:
balance: Increased the benefits of skills when performing surgery.
/:cl:

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -120,7 +120,7 @@ proc/do_surgery(mob/living/carbon/M, mob/living/user, obj/item/tool)
 	if(user.skills.getRating("surgery") < SKILL_SURGERY_PROFESSIONAL)
 		user.visible_message("<span class='notice'>[user] fumbles around figuring out how to operate [M].</span>",
 		"<span class='notice'>You fumble around figuring out how to operate [M].</span>")
-		var/fumbling_time = SKILL_TASK_FORMIDABLE - ( SKILL_TASK_AVERAGE * user.skills.getRating("surgery") ) // 20 secs non-trained, 15 amateur, 10 semi-prof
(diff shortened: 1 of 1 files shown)


code/modules/surgery/surgery.dm
```diff
@@ -120,7 +120,7 @@ proc/do_surgery(mob/living/carbon/M, mob/living/user, obj/item/tool)
 	if(user.skills.getRating("surgery") < SKILL_SURGERY_PROFESSIONAL)
 		user.visible_message("<span class='notice'>[user] fumbles around figuring out how to operate [M].</span>",
 		"<span class='notice'>You fumble around figuring out how to operate [M].</span>")
-		var/fumbling_time = SKILL_TASK_FORMIDABLE - ( SKILL_TASK_AVERAGE * user.skills.getRating("surgery") ) // 20 secs non-trained, 15 amateur, 10 semi-prof
+		var/fumbling_time = SKILL_TASK_FORMIDABLE - ( 6 SECONDS * user.skills.getRating("surgery") ) // 20 secs non-trained, 14 amateur, 8 trained, 2 prof
```
Rohesie: Could you make a new define for it?

psykzz: It felt weird here specifying a define, instead i thought about defining a curve, but not sure how to approach it.

I kinda want a non-linear lerp.

Rohesie: We should decide then whether to drop the skill defines or not.

psykzz: I like having them for the total time number, but when you can start at lower than max (in this case one less than professional) its hard to find the correct curve.

PR #3603, opened 2019-12-21
Crash logs

Rohesie:
This will potentially produce some pretty massive logs.
It's a followup of <http://www.byond.com/forum/post/2528780>
The crash dumps indicate they happened during a read/write operation, so we are going to log all of such and see if we can find out what's causing it.


Rohesie:
Solved with #3637

PR #3609, opened 2019-12-22
Turns View into a string

### LINKED ISSUE #3258: Detector blips inaccurate when using M4RA scope
## Testmerges:

None.

## Explanation:

Motion detector blips show up inaccurately when using the M4RA scope. Unknown if it has the same effect on other scopes, will update with relevant information after sufficient testing.

## Reproduction:

Step 1: Pick the Scout set from the Spec vendor. All you need is the M4RA, ammo, and a motion detector. For my test, I used the scout detector, in my helmet.

Step 2: Set it to blip on Short Range mode. Unknown if this works on further ranges, since I'm reporting this the moment I found it.

Step 3: Wield the M4RA and use the scope function. If there is something in the vicinity that'll send off a blip, the visual aid (A green, red, cyan or blue circle) will not be on the target.

Rohesie:
`world.view` seems to be a constant that can't be changed, remains as `7` even if redefined.

This fixes several type mismatch runtimes we are having.

Fixes #3258


spookydonut:
http://www.byond.com/docs/ref/index.html#/world/var/view

> Default value:
>     5 
> 
> Possible values:
>     -1 to 34 or "WIDTHxHEIGHT" 
> 
> This is the default map viewport range. The default value of 5 produces an 11x11 viewport. A value of -1 turns off the map display altogether. The client may automatically scale down icons in order to conveniently fit the map on the player's screen.
> 
> For non-square views, you can assign this to a text string of the form "WIDTHxHEIGHT". For example, "11x11" is equivalent to a view depth of 5, but you could make it wider like this: "13x11".
> 
> This setting also affects the default range of the view(), oview(), range(), and orange() procedures.
> 
> If the entire map is small enough to fit on one screen (arbitrarily defined to be 21x21 or less), the default view is automatically adjusted to fit the map. In this case, client.lazy_eye is also automatically turned on by default, since you probably don't want the map to scroll around.

PR #3611, opened 2019-12-22
Removes acid blood from all Xenos except boilers

TiviPlus:
## About The Pull Request

Removes acid blood, I was told Rohesie greenlighted this ages ago but that everyone was too lazy to do it themselves.

## Why It's Good For The Game

Getting punished for dealing damage is not cool

## Changelog
:cl:
balance: Removed acid blood damage from standing near xenos that take damage.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -173,3 +173,25 @@
 		else
 			if(decal.random_icon_states && length(decal.random_icon_states) > 0) //If there's already one, just randomize it so it changes.
 				decal.icon_state = pick(decal.random_icon_states)
+
+		var/splash_chance = 40 //Base chance of getting splashed. Decreases with # of victims.
+		var/distance = 0 //Distance, decreases splash chance.
+		var/i = 0 //Tally up our victims.
+		
+		if(!(/mob/living/carbon/xenomorph.xeno_caste.caste_flags & CASTE_ACID_BLOOD))
+			return
+		else
+			for(var/mob/living/carbon/human/victim in range(radius,src)) //Loop through all nearby victims, including the tile.
+				distance = get_dist(src,victim)
+
+				splash_chance = 80 - (i * 5)
+				if(victim.loc == loc) splash_chance += 30 //Same tile? BURN
+				splash_chance += distance * -15
+
+				if(splash_chance > 0 && prob(splash_chance)) //Success!
+					i++
+					victim.visible_message("<span class='danger'>\The [victim] is scalded with hissing green blood!</span>", \
+					"<span class='danger'>You are splattered with sizzling blood! IT BURNS!</span>")
+					if(prob(60) && !victim.stat && !(victim.species.species_flags & NO_PAIN))
+						victim.emote("scream") //Topkek
+					victim.take_limb_damage(0, rand(10, 25)) //Sizzledam! This automagically burns a random existing body part.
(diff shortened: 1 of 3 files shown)


Rohesie:
Are you able to, instead of outright remove it, turn this into a caste flag and keep it for, say, the boiler?

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -173,3 +173,25 @@
...
 			if(decal.random_icon_states && length(decal.random_icon_states) > 0) //If there's already one, just randomize it so it changes.
 				decal.icon_state = pick(decal.random_icon_states)
+
+		var/splash_chance = 40 //Base chance of getting splashed. Decreases with # of victims.
+		var/distance = 0 //Distance, decreases splash chance.
+		var/i = 0 //Tally up our victims.
+		
+		if(!(xeno_caste.caste_flags & CASTE_ACID_BLOOD))
+			return
+		else
```
Rohesie: No need for an `else` if you return, but on this kind of proc it's better to have this within the `if`, as more things can happen after.
Also the definition of variables such as `splash_chance`, `distance` and `i` should be within the `if`, as they are not used anywhere else.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -173,3 +173,25 @@
...
+			for(var/mob/living/carbon/human/victim in range(radius,src)) //Loop through all nearby victims, including the tile.
+				distance = get_dist(src,victim)
+
+				splash_chance = 80 - (i * 5)
+				if(victim.loc == loc) splash_chance += 30 //Same tile? BURN
```
Rohesie: Put the `splash_change` in a new line, properly indented.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -173,3 +173,25 @@
...
+				splash_chance += distance * -15
+
+				if(splash_chance > 0 && prob(splash_chance)) //Success!
```
Rohesie: `prob` of a negative or zero value will return false, so no need to check for if it's higher than zero.

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -177,18 +177,21 @@
...
+					i++
+					victim.visible_message("<span class='danger'>\The [victim] is scalded with hissing green blood!</span>", \
+					"<span class='danger'>You are splattered with sizzling blood! IT BURNS!</span>")
+					if(prob(60) && !victim.stat && !(victim.species.species_flags & NO_PAIN))
```
Rohesie:
```suggestion
					if(victim.stat != CONSCIOUS && !(victim.species.species_flags & NO_PAIN) && prob(60)
```

TiviPlus:
Travis complaining about a { in a place there is none, so that's nice

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -173,22 +173,22 @@
...
 			if(decal.random_icon_states && length(decal.random_icon_states) > 0) //If there's already one, just randomize it so it changes.
 				decal.icon_state = pick(decal.random_icon_states)
-
+		
+		if(!(xeno_caste.caste_flags & CASTE_ACID_BLOOD))
+			return
 		var/splash_chance = 40 //Base chance of getting splashed. Decreases with # of victims.
 		var/distance = 0 //Distance, decreases splash chance.
 		var/i = 0 //Tally up our victims.
+			for(var/mob/living/carbon/human/victim in range(radius,src)) //Loop through all nearby victims, including the tile.
```
spookydonut: you have a space at the start of this line per this: https://github.com/SpaceManiac/SpacemanDMM/issues/117

spookydonut: No wait, you indented too much

Issue #3578, opened 2019-12-17
You can't keybind defender's forward charge


Teaika:
## Testmerges:
no I don't think so

## Reproduction:
1. Enter keybindings menu
2. Try to find Forward Charge ability
3. Fail at it.

spookydonut:
fixed

PR #3613, opened 2019-12-23
Allows the keybinding of defenders forward charge

Hughgent:
## About The Pull Request

#3578

## Why It's Good For The Game

Xeno keybinds.

## Changelog
:cl:
fix: defender forward charge can now be keybound
/:cl:

### DIFF
--- a/code/__DEFINES/components.dm
+++ b/code/__DEFINES/components.dm
@@ -399,6 +399,7 @@
 
 #define COMSIG_XENOABILITY_HEADBUTT "xenoability_headbutt"
 #define COMSIG_XENOABILITY_TAIL_SWEEP "xenoability_tail_sweep"
+#define COMSIG_XENOABILITY_FORWARD_CHARGE "xenoability_forward_charge"
 #define COMSIG_XENOABILITY_CREST_DEFENSE "xenoability_crest_defense"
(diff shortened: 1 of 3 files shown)


Hughgent:
May or may not be truly functional, I am not perfectly certain on how the keybinds work in this case as I don't play xeno.

Rohesie:
Have you tested it locally?

Hughgent:
Long story short on that. In writing this, I copied the bits about the signal from tail_sweep.

In testing, the tail sweep kept getting triggered instead. So I double checked the signals, and saw I was sending the tail sweep signal instead of the newly created forward charge signal. 

BUT since I had changed up my key bind preferences, even after I had made the corrections, the key I had bound was still triggering tail sweep rather than forward charge.

Thus I think there is an issue with my local savefile. Thus I request someone else test it locally before TMing on the main game.

psykzz:
This never worked because you need to have "click" event to get the turf to charge to, otherwise you should replace this with get 4 tiles ahead and charge.

Hughgent:
As it is, right now, the keybind just selects the ability. (this is something I tested)

Considering the 'click' event needed. I think this is proper behavior for the keybind.
(Press Key, Middle Mouse Click) >> charge at clicked location.

psykzz:
in that case, yeah this just selects the ability to make middle clicking easier 👍

PR #3622, opened 2020-01-03
Delete unused/broken maps

spookydonut:
As I said a few months ago, I am deleting these maps as they haven't been worked on and arent in playable states, this will reduce the burden on contributors in having to keep these current with typepath etc changes.


spookydonut:
Other maps on the chopping block: 
sulaco
whiskey outpost

My criterion is:
must have crash gamemode markers
OR
under active development

Rohesie:
Don't think Desert Dam needs removing, or causes any issues, but whatever.

PR #2540, opened 2019-08-03
Add support for custom clientside CSS

psykzz:
## About The Pull Request

Adds a new verb in Preferences that accepts an input to input custom CSS.
The CSS is then loaded into a style element within the `<head>` of goonchat.

The a new savefile is then created and sent to the client.

On login and creation of the chatOutput datum, I check if there is a local file on the client load it into the client's head.

No savefile is stored on the server.

## Why It's Good For The Game

Customization

## Changelog
:cl:
add: A new verb has been added to preferences that allows you to set your own custom css for goonchat.
/:cl:

PR #3637, opened 2020-01-09
Removes custom CSS

Rohesie:
Mainly due to being unmaintained and crashing the server. Lummox hasn't been able to diagnose the cause properly, but it happens every time certain client CSS files are read at different points of initialization in general.
This implementation is also problematic because this file is shared with every other SS13 server, which can lead to issues if they also use the same method, and easily corrupt the file.
Since we've had light and dark mode for quite a while, this is not really very much needed. Other skin or CSS options can be added if necessary.

## Changelog
:cl:
del: Removed custom CSS option.
/:cl:

### DIFF
--- a/code/modules/goonchat/browserOutput.dm
+++ b/code/modules/goonchat/browserOutput.dm
@@ -21,7 +21,6 @@ GLOBAL_DATUM_INIT(iconCache, /savefile, new("tmp/iconCache.sav")) //Cache of ico
 	var/working      = TRUE
 	var/list/connectionHistory //Contains the connection history passed from chat cookie
 	var/adminMusicVolume = 10 //This is for the Play Global Sound verb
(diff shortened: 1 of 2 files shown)


psykzz:
From the original PR there looks to be some extra things that need to be removed - https://github.com/tgstation/TerraGov-Marine-Corps/pull/2540/files

Rohesie:
Oh, I thought LaKiller8 had implemented this, not you. Will poke you when you have time to discuss the issues, in case you want to maintain this.
It's currently crashing the server on bad read operations. If we stored the custom CSS as a savefile, properly sanitized, we could fix it.

psykzz:
I have no problem removing it. it was added because at the time we didn't want to create dark mode but that got added.

PR #3644, opened 2020-01-12
Start building dreamchecker from source

spookydonut:
~~lets see if cargo takes forever to run~~

This is fine to merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,13 @@ matrix:
             - python3-pip
             - python3-setuptools
             - pcregrep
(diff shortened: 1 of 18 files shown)


spookydonut:
if anyone test merges this they should lose keyholder

psykzz:
are we doing this from source because releases are slow or something?

spookydonut:
> 
> 
> are we doing this from source because releases are slow or something?

yes but also spacemaniac has now had sufficient pressure to do a new release soon

PR #3646, opened 2020-01-13
Adds components as an external dependency

ninjanomnom:
This allows component code to be located remotely so multiple codebases can make use of the same code. This allows for multiple codebases to collaborate on core code without as much effort being spent fixing the same issues and dealing with conflicts since it forces you to contribute to one location. It also has a nice code quality benefit in that you can't put feature specific code in core system code.

We're going to need to experiment with it a bit to see if it's all easy enough to use. I'll need to test a bit more but this seems fully functional as a first version using submodules.

Do not testmerge this


ninjanomnom:
@Cyberboss is tgs able to handle submodules or does it need an update for that

ninjanomnom:
We're going to need to make linting ignore submodules

ninjanomnom:
This should be good to go, we just need to be prepared to help anyone who has issues for the days following.

Cyberboss:
FOR THE FIFTH EMAIL

TGS DOES SUPPORT SUBMODULES

PR #3651, opened 2020-01-19
Icon and gun fixes.

Rohesie:
* Fixes an unnamed saber icon causing hand overlay issues.
* Fixes empty guns appearing as invisible when wielded.
* Cleans up the vending machine colors so that all guns appear properly highlighted with their ammo right under them.
* Adds the DMR to the cargo vendor machine.
* Cleans up a little code.


code/game/objects/items.dm
```diff
@@ -112,6 +112,10 @@
 	return ..()
 
 
+/obj/item/proc/update_item_state(mob/user)
+	item_state = "[initial(icon_state)][flags_item & WIELDED ? "_w" : ""]"
```
spookydonut: you could throw in a check here for the `iconstate_w` icon_state existing
http://www.byond.com/docs/ref/#/proc/icon_states

Rohesie: An automated check for non-existent icon states on init would be really good and show a lot of unmaintained objects. But doing a check every time an item state is updated or item wielded wouldn't make as much sense.

Rohesie:
Should be merged soon, as the game without it has a noticeable visual bug.

PR #3660, opened 2020-01-20
remove unreachable code

spookydonut:
https://github.com/SpaceManiac/SpacemanDMM/pull/123


Rohesie:
Can't review due to being blocked.
We can't work like this so you'll have to make a choice.

spookydonut:
> 
> 
> Can't review due to being blocked.
> We can't work like this so you'll have to make a choice.

@Rohesie Can't communicate or check TMs due to you picking a fight then banning me /shrug

PR #3142, opened 2019-10-20
Radial Menu Bandages

Hughgent:
## About The Pull Request

PR has been opened.
Things to do:
- [x] Add it to preferences for players to use/not use.
- [x] Check to make sure it works.
- [x] Get some sprites for the radial menu icons.

It works for surgery steps now.

## Why It's Good For The Game

People will be able to apply bandages/whathaveyou by using this radial menu.

## Changelog
:cl: Hughgent
add: Radial menu for medical bandages.
add: Preference for players to turn Radial menu on and off. Doctors may prefer off.
/:cl:

PR #3668, opened 2020-01-25
Revert "Radial Menu Bandages (#3142)"

Rohesie:
This was merged against my reviews. I wasn't notified it would happen in 24 hours because I was blocked by the author, and it shouldn't have happened at all without having been contacted about it, notification or not.


spookydonut:
All pending change requests were addressed and the pr was verified to be working.

There were no balance concerns with it.

spookydonut:
I haven't seen any justification for a revert here @Rohesie other than your unreasonable demands were ignored.

PR #3677, opened 2020-03-21
smg now does 5 round burst

Pariah919:
## Why It's Good For The Game

0.2 is 5 bullets a second in theory. It also does no AP. This is so burst isn't garbo bad, removes BFA because I'd get screeched at otherwise

## Changelog
:cl:

balance: SMG now has 5 round burst, removed BFA

/:cl:


Pariah919:
This hits the average 100-120 DPS. (this being 113 but no AP.)

PR #3688, opened 2020-03-25
Synths don't scream

psykzz:
Synths don't scream anymore when burning alive in acid spray.


Toxici11i:
@psykzz wouldnt it be prudent to address the fact it applies the pain stun to synths too (along with the slow)

Toxici11i:
@psykzz wouldnt it be prudent to address the fact it applies the pain stun to synths too (along with the slow)

PR #3689, opened 2020-03-26
Buff Synthetic burn resistances 

Toxici11i:
## About The Pull Request
Buffs synthetics to have acid resistance rather than vulnerability - ill cover why below

## Why It's Good For The Game
Currently, synths can be instantly stunlocked and killed (nearly full to 0hp) upon so much as touching an acid tile or acid smoke
https://cdn.discordapp.com/attachments/500801682058379265/692526872940118036/unknown.png
https://cdn.discordapp.com/attachments/500801682058379265/692499935983173672/unknown.png
both of these screenshots happened in less than two seconds respectively.
Furthermore, even with heavy armor this occurs (second one occurred with, the first one with medium)

Basically, with the amount of trash synths have piled ontop of their gameplay, this would be a nice change of pace.

Peoples argument for 'more health and brute resist' being the tradeoff for not being able to use guns, heavy armaments (such as dropship weapons + tank + mortar) is a terrible excuse for making a role utterly useless when faced with burn damage in general

## Changelog
:cl:
tweak: synth and old synth now have 0.9* the normal burn damage.
tweak: old synths now have extra brute armor to accompany the slowdown and extra HP compared to new synths
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -590,7 +590,7 @@
 	total_health = 150 //more health than regular humans
 
 	brute_mod = 0.75
-	burn_mod = 1.1
+	burn_mod = 0.90 //Synthetics should not be instantly melted by acid compared to humans - This is a test to hopefully fix very glaring issues involving synthetics taking 2.6 trillion damage when so much as touching acid
 
 	cold_level_1 = -1
 	cold_level_2 = -1
@@ -642,8 +642,8 @@
 	slowdown = 1.3 //Slower than later synths
 	total_health = 200 //But more durable
 	insulated = 1
-	brute_mod = 0.75
-	burn_mod = 1.1
+	brute_mod = 0.60 //but more durable
+	burn_mod = 0.90 //previous comment
 
 	cold_level_1 = -1
 	cold_level_2 = -1


Toxici11i:
basically synths are poo poo stinky and useless at the moment as soon as they touch anything that does burn due to their long repair time compared to humans (5s to heal around 15 burn) and expensive in terms of cable and welder fuel. This is compared to humans, who can premed (tramaoxy to remove paincrit and slowdown) and kelo/bicard/tricord to heal quickly.
this is not funny and only proves how the 'advantages' of synths have turned into cringe moments

psykzz:
Updated the name to make TM's easier, it was confusing with the other draft open.

Toxici11i:
> Since i couldn't get the #balance team to agree to the changes, either this needs to be closed or revised.
> https://discordapp.com/channels/498569646014857217/498576457740714001/694975488476839997

image

jroinc:
I'm down with this for now. I'd like to see synth damage output lowered, but survivability increased.

Toxici11i:
> I'm down with this for now. I'd like to see synth damage output lowered, but survivability increased.

The thing is, early synths (The synth with buffed up damage, by default even before this) already has many negatives towards doing damage with melee as discussed. Firstly, slower MS means by default (even without armor) that they have a lower chance to disengage, even vs the slowest xeno, so its extremely hard to move away from a fight you are losing
Secondly, with armor you become tanky as hell sure. But then you lose the ability to move almost completely
Thirdly, early synths have a higher healthpool (sure) but this also does NOT benefit them. They take longer to repair as a result, they stay slowed due to damage for longer (more repairs to get out of that 'damaged' slowdown) etc.

jroinc:
Here's a thought straight from the discord-

Lower synth damage to unga or below unga level. Give them bump melee. Remove combat-synth rule.

If alone, they can combat decently well while against only melee, but will get fucked once heavier ranged castes show up.

If in a group of ungas, they run the risk of hitting friendlies if bump is toggled, but they're surrounded by friendlies and should TECHNICALLY be supporting mainly.

If the group starts to thin out and be hurt, they can toggle it and use their amazing armor and decent health with consistent damage to tank melee castes away and provide breathing room. 

They're unlikely to get kills, but they'll provide breathing space for the main party and probably won't die.

Toxici11i:
bump melee is a terrible idea and i refuse to implement it, especially on melee dependent things like synth. also, giving them 77% of unga damage (for example) would actually mean buffing unga meele damage, which would be terrifying.

PR #3691, opened 2020-03-26
Synthetic skill changes & skill fixes

Toxici11i:
## About The Pull Request
This is a draft looking into how we can balance synths for potentially allowing synths to exist and use guns.

## Why It's Good For The Game
still WIP

## Changelog
:cl:
tweak: removed spooky synth skills (super specialist skills that not even DS had)
tweak: gives synths a boost to CQC and meele respectively
tweak: gives synths pistol training, but untrained firearm training for everything else
fix: now factors in melee skill for damaging objects and humans.
/:cl:


Toxici11i:
After testing in live, this may need to be held back while another solution is put in place to adjust how skills affect the way guns and recoil work - currently firearms:0 is better than 1, you experience less recoil at the expense of a slower fire rate (????).

basically fucking hell skills are broken as fuck again

Pariah919:
giving synth more melee damage is pointless, as they're not a combat role, and the gun skills themselves are moreso a meme but I believe they're fine outside of the dumb fringe cases where the damn thing breaks and lets you use guns as synth.

Toxici11i:
> giving synth more melee damage is pointless, as they're not a combat role, and the gun skills themselves are moreso a meme but I believe they're fine outside of the dumb fringe cases where the damn thing breaks and lets you use guns as synth.

Currently, synths are usually not allowed to be a combat role (self preservation is law 1, unless stated otherwise/done for testing purposes), however, yeah while it may not seem to help synths, it really does. Even if guns arent enabled permanently, the current usage of melee in self defence is staggering (nearly every synth carries a machete), and so the most effective buff to make synths easier to play (and not die instantly) is to give them a buff towards melee damage vs ranged damage (throwing knives)

Toxici11i:
> To increase differentiation between early and gen 2 synth, I'd nerf gen 2 synth melee/cqc one level each. Early synth can then still have the advantage of better health and melee damage, but less supportive skill, while gen 2 has fully buffed supportive skill but less melee damage.

Thats fair, ill bump it down after these tests

code/_onclick/item_attack.dm
```diff
@@ -81,7 +81,8 @@
 	user.visible_message("<span class='warning'>[user] hits [src] with [I]!</span>",
 		"<span class='warning'>You hit [src] with [I]!</span>")
 	log_combat(user, src, "attacked", I)
-	take_damage(I.force, I.damtype, "melee")
+	var/power = I.force + round(I.force * 0.3 * user.skills.getRating("melee_weapons")) //30% bonus per melee level
```
psykzz: If untrained this will equal 0.

Toxici11i: The code is force + force * 0.3 * skill, if no skill, no modifier is applied but the base damage is kept

Toxici11i: So no, the power would just be the force if skill is not present, e.g machete would still be 60 damage with no skill, as it is its base force

Toxici11i: To clarify @psykzz, the total damage for a machete using this method would be calculated by power = 60(force of the machete) + (60*0.3*skill)

At 0 skill, there would be 0 damage bonus so it would be 60+ 0 aka 60
At 1 skill it would be a 30% increase (18 damage) or 78 damage
At 2 skill it would be 60% increase, bringing it up to 96 damage.

There shouldnt be any case where the total force of the weapon is 0, even with no skill, and if it does occur, investigate the weapons force variable. If its a value higher than 0, it borked somehow, or armor completely negated it

psykzz: All good, this is my bad.

PR #3704, opened 2020-03-30
Buffs the RPG

Pariah919:
## About The Pull Request

Buffs the RPG majorly, it's generally agreed upon to be kind of bad.

## Why It's Good For The Game

Makes the RPG far better on direct hits, which should hurt a lot considering if you miss 2/3 rocket types don't really do much

## Changelog
:cl:

balance: RPG direct hits are far deadlier!
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -937,7 +937,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	accuracy_var_low = 7
 	accurate_range = 15
 	damage = 200
-	penetration = 200
+	penetration = 75
 	max_range = 20
 
 /datum/ammo/rocket/wp/drop_nade(turf/T, radius = 3)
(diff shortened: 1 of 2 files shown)


Toxici11i:
This is a good change, HE is now viable due to its increased on hit damage and AP, meaning that AP rockets are no longer the only rocket that is viable.

also incendiary rockets when

GravitronBlue:
HE maybe deserves the buff, AP now effectively being an instakill rifle against every xeno (with the one exception of ancient crushers, but given that it still crits it may as well be the same) feels like a mistake, queens in particular being one-shot while trying to screech is occasionally costing xenos the round.

Toxici11i:
> HE maybe deserves the buff, AP now effectively being an instakill rifle against every xeno (with the one exception of ancient crushers, but given that it still crits it may as well be the same) feels like a mistake, queens in particular being one-shot while trying to screech is occasionally costing xenos the round.

sounds like the xenos fault for standing still long enough to get AP rocketed, even though it has a clear windup and a flashing WARNING SYMBOL above the RPG guys head.

PR #3712, opened 2020-04-01
Adds TheBimmer231 as new map/icon codeowner

psykzz:


psykzz:
@TheBimmer231 can you confirm

PR #3714, opened 2020-04-01
Nested hosts don't need to breath

psykzz:
## About The Pull Request
Buff to nested marines, they will no longer recover / loss breath while nested.

## Why It's Good For The Game
Increase the change that a xeno capture turns into a successful infection.

## Changelog
:cl:
add: Buff to nested marines, they will no longer recover / loss breath while nested
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon_helpers.dm
+++ b/code/modules/mob/living/carbon/carbon_helpers.dm
@@ -12,7 +12,7 @@
 
 
 /mob/living/carbon/proc/need_breathe()
-	if(reagents.has_reagent(/datum/reagent/toxin/lexorin) || HAS_TRAIT(src, TRAIT_STASIS))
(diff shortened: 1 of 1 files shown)


Toxici11i:
This is a no from me, xenos usually have every chance to capture due to larval sting stabilising people etc, it also nerfs suicide pens into the ground, and stops people from dying in general.

bad nono and xeno cringe

jroinc:
Currently, if you slash/spit an unga, they won't go down until hardcrit, and that's not recoverable from. Stamina is an ENTIRELY DIFFERENT HEALTH BAR which is shitty as well.

On the other hand, this will just lock people into hardcrit hell.

Counter-suggestion- Make nests heal everything at 1pt/tick of ALL. Nests get a boost to evolve speed based off of unga health, critted ungas will heal up, AND this allows ungas a reasonable chance to escape once they finish resisting.

psykzz:
Will close the PR and put it under consideration, thanks for the feedback.

PR #3727, opened 2020-04-03
Icy caves rework

KwinOriginal:
Made parts of LZ2 more open so it could be used. Added shutters for the same reason. 

## Changelog
:cl:
balance: Rebalanced icy caves LZ2
/:cl:


psykzz:
It would look nice to have a fence or something between or around the pathways.

PR #3747, opened 2020-04-05
M56D IFF adding

Pariah919:
## About The Pull Request

Gives the M56D IFF and more damage/pen
## Why It's Good For The Game

Have you even seen one ran as a non-joke other than in crash where you get a free one? Giving it IFF makes it far more useable and better.

## Changelog
:cl:

balance: m56d now has iff
spellcheck:Removed some comments related to it not having IFF.
/:cl:


jroinc:
+1, it takes an unga out of being mobile and doing stuff so we should make it better to use.

PR #3753, opened 2020-04-06
DataMiner

Hughgent:
## About The Pull Request

I wanted a way of learning about each and every object/thing in the game at once. I wanted to compare them to each other. I wanted to see how many X does it take to do Y. (bullets per xeno in this case)

As it turned out, there was a way using a tool from strongDMM.

This should be codebase agnostic, so feel free to use as the license allows!

## Why It's Good For The Game

With enough data analysis, 'balance' could possibly be achieved.

## Changelog
:cl: Hughgent
add: There is a data analysis tool in the tools folder for those interested.
/:cl:


Hughgent:
For those curious. on the TGMC codebase, it creates a 3.5 mb JSON file, that converts into a 55mb CSV file. If you apply no filters.

Hughgent:
I don't see much of an issue with a person copying a filepath into a bit of python code. This tool is really not for the average user.

psykzz:
That said but you have a batch file to install the requirements which is arguable the easiest part.

You should update the batch file to finish with the generated.csv

The batch should 
 - Check python is installed, throw error if not
 - Install requirements
 - Run the parser to dump the json file
 - Run python to convert JSON to CSV

That way the only requirement is to install python

Hughgent:
Woah woah woah, the 55MB CSV file is basically useless. This is why I'm breaking it out into the steps that there are.

psykzz:
If its useless then why add it to the repo?
I assume its not, so simplifing the steps seems reasonable.

Hughgent:
I say it's useless because it's 90% empty space, all typepaths don't use all variables, which is what the 55MB CSV is.

I included a function in the python to break it down by desired typepath. And when you apply this filter, the CSV ends up being 100kb (depending on what you desire)

I'm sure that, when you want to see all the weapons, every single possible /datum/ isn't as relevant. neither are all the /world/ vars.

Hughgent:
However, please do provide a suggestion on how to make the batch file run the sdmmparser.exe

psykzz:
Good catch on the last var, this is why you never blindly trust PR's!

Also I don't mean to overshadow your work, its an amazing idea, i just felt it was easier to explain what i was suggesting via a PR of my own.

script.py
```diff
@@ -0,0 +1,47 @@
+import os
```
psykzz, line 1: Oh my bad i wasn't meant to commit this file.

psykzz: similar to sdmmparser.exe in root.

PR #3766, opened 2020-04-07
Fixes internal storage references on deletion

Rohesie:
These were forcing a hard del due to not being properly cleared.


Rohesie:
Not ready yet. I'll fix it later.

Rohesie:
Alright, this is ready.

PR #3768, opened 2020-04-07
Components cleanup fix

Rohesie:
Components were not being correctly removed.

This adds latest /tg/ code, plus a few useful procs for debugging deletion and finding uncleared references.


Rohesie:
This is ready for merge.

PR #3769, opened 2020-04-07
Projectile reference fix and cleanup

Rohesie:
Several projectiles were still referenced on `in_chamber` on deletion, causing a hard del.

Also removed the legacy `clear_jam` proc.


Rohesie:
This is ready.

PR #3770, opened 2020-04-07
mob reference cleanup

Rohesie:
Clears up mob references from lists so that they may not require being hard-deleted.
Requires #3768 to compile correctly.


Rohesie:
And this should be ready.

PR #3775, opened 2020-04-07
YouTube Links for the lobby music

Hughgent:
## About The Pull Request

We should no longer need to store music files on the server itself.

however, this isn't functioning at the moment. at least not in testing alone.
It compiles, server starts normally, but no music gets played. Probably because I'm not actually making the client perform the proc.

Any help would be appreciated.

## Why It's Good For The Game

should be no change to the actual game.

## Changelog
:cl: Hughgent
tweak: Lobby music now plays via YouTube links.
/:cl:


Hughgent:
I will admit, testing this locally is not without issue. I'm fairly confident it will work on the actual server though.

Rohesie:
Make them config entries so that can changed through the server.
And keep them old system and at least one music for when there's no youtube-dl

psykzz:
Im opposed to how this is right now.

Download the music on server start instead of during the play proc
    - You may additionally needto download music if its changed while the server is running by an admin
Ensure that there is enough backup music for when youtube-dl fails to work or download music.

Lastly, whats to stop a client spamming this proc to force the server to continuously download music from youtube?

Hughgent:
as I understand it, shouldn't this proc make the client download and play the music, rather than the server pushing the music?

Rohesie:
Yea, the server doesn't download the music. The client opens a browser and downloads the music themselves, as if they were watching a video on youtube.

code/controllers/subsystem/ticker.dm
```diff
@@ -43,13 +43,12 @@ SUBSYSTEM_DEF(ticker)
...
 
 	login_music = pick(
-		'sound/music/SpaceHero.ogg',
-		'sound/music/ManOfWar.ogg',
-		'sound/music/PraiseTheLord.ogg',
-		'sound/music/BloodUponTheRisers.ogg',
-		'sound/music/DawsonChristian.ogg',
-		'sound/music/Bomber.ogg',
-	)
+		list("https://youtu.be/fNdMC6_eUGk",1,180), //Space Hero
```
spookydonut, line 46: yeah make this a config file

Hughgent:
having trouble making this work as a config file
the to_chat world is only showing the link, when it SHOULD be showing the other two numbers as well
and I'm not sure why.

Hughgent:
I think this fails travis because it's using the server config file, which doesn't have the links.

psykzz:
This should get disabled again in configuration when your ready to merge.

code/game/sound.dm
```diff
@@ -105,11 +105,38 @@
...
 		return FALSE
 	if(prefs && (prefs.toggles_sound & SOUND_LOBBY))
-		SEND_SOUND(src, sound(SSticker.login_music, repeat = 0, wait = 0, volume = vol, channel = CHANNEL_LOBBYMUSIC)) // MAD JAMS
+		var/ytdl = CONFIG_GET(string/invoke_youtubedl)
+		if(!ytdl)
+			play_title_music_legacy()
+			return
+		to_chat(world, SSticker.login_music[1])
+		to_chat(world, SSticker.login_music[2])
+		to_chat(world, SSticker.login_music[3])
```
psykzz, line 118: Remove these

code/controllers/subsystem/ticker.dm
```diff
@@ -41,15 +41,11 @@ SUBSYSTEM_DEF(ticker)
...
-		'sound/music/SpaceHero.ogg',
-		'sound/music/ManOfWar.ogg',
-		'sound/music/PraiseTheLord.ogg',
-		'sound/music/BloodUponTheRisers.ogg',
-		'sound/music/DawsonChristian.ogg',
-		'sound/music/Bomber.ogg',
-	)
+	
+	var/all_music = CONFIG_GET(keyed_list/lobby_music)
+	var/key = pick(all_music)
```
psykzz, line 46:
```suggestion
	var/key = safepick(all_music)
```

code/controllers/subsystem/ticker.dm
```diff
@@ -41,15 +41,11 @@ SUBSYSTEM_DEF(ticker)
...
+	var/music_options = splittext(all_music[key], " ")
+	login_music = list(music_options[1], music_options[2], music_options[3])
```
psykzz, line 48:
```suggestion
	if(key)	
		var/music_options = splittext(all_music[key], " ")
		login_music = list(music_options[1], music_options[2], music_options[3])
```

code/modules/client/preferences_toggles.dm
```diff
@@ -98,7 +98,11 @@
 		to_chat(src, "<span class='notice'>You will now hear music in the game lobby.</span>")
 		if(!isnewplayer(mob))
 			return
-		playtitlemusic()
+		var/all_music = CONFIG_GET(keyed_list/lobby_music)
+		if(all_music)
+			play_title_music()
+		else
+			play_title_music_legacy()
```
psykzz, line 105:
```suggestion
		play_title_music()
```

code/modules/mob/new_player/login.dm
```diff
@@ -15,4 +15,9 @@
 	sight |= SEE_TURFS
 
 	new_player_panel()
-	client?.playtitlemusic()
+	var/all_music = CONFIG_GET(keyed_list/lobby_music)
+	if(all_music)
+		client?.play_title_music()
+	else
+		client?.play_title_music_legacy()
```
psykzz, line 22:
```suggestion
	play_title_music()
```

code/controllers/subsystem/ticker.dm
psykzz: This will check if its null as well and return safely

PR #3779, opened 2020-04-08
Comments out world/topic spam filtering

MrStonedOne:
This blocked the website game banners and has no whitelist.

Could somebody give this a whitelist.

PR #3782, opened 2020-04-08
Topic filtering whitelist

psykzz:
closes #3779 
This adds a whitelist controlled by config.txt
These are not banned when they go over length or request per second limits.

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -133,19 +133,22 @@ GLOBAL_VAR(restart_counter)
 	if (bannedsourceaddrs[addr])
 		return
 
-	if(length(T) >= MAX_TOPIC_LEN)
-		log_admin_private("[addr] banned from topic calls for a round for too long status message")
-		bannedsourceaddrs[addr] = TOPIC_BANNED
-		return
-
-	if(lasttimeaddr[addr])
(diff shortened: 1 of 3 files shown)


code/game/world.dm
```diff
@@ -133,19 +133,22 @@ GLOBAL_VAR(restart_counter)
...
-		log_admin_private("[addr] banned from topic calls for a round for too long status message")
-		bannedsourceaddrs[addr] = TOPIC_BANNED
-		return
-
-	if(lasttimeaddr[addr])
-		var/lasttime = lasttimeaddr[addr]
-		if(world.time < (lasttime + 2 SECONDS))
-			log_admin_private("[addr] banned from topic calls for a round for too frequent messages")
+	var/list/filtering_whitelist = CONFIG_GET(keyed_list/topic_filtering_whitelist)
+	var/host = splittext_char(addr, ":")
```
Rohesie, line 137: Won't `addr` here always be ASCII? If so, you can just use `splittext` instead

Also, where are you comparing the address?

psykzz: `if(!filtering_whitelist[host[1]])`

`filtering_whitelist` is a map

PR #3787, opened 2020-04-08
Fixes Pillar maint id access

grinkgo:
## About The Pull Request

Something I never noticed when I published the Pillar was that for some unknown reason (cough cough KWIN) most maintenance doors were set to only open with bridge access.
This PR changes the perms so that anyone with at least Ship Tech access can use the maintenance airlocks. Only exceptions are maintenance airlocks leading into private rooms i.e. FC, CO, and CL quarters.

## Why It's Good For The Game

Ship Techs can use maintenance tunnels.

## Changelog
:cl:
fix: Ship Techs should now be able to properly access most maintenance airlocks on the Pillar
/:cl:


tgstation-server:
Your Fix/Feature pull request delta is now above zero (0). Feel free to make Feature/Tweak/Balance PRs.

PR #3789, opened 2020-04-09
Increases bull charge steps, lowers damage

Mouseofthecake:
## About The Pull Request

Increases bull steps_for_charge from 3 to 5, 1 less than crusher's. This means it'll take 5 steps for a bull to enter charging phase. Also, lowers their damage from 40 to 30.

## Why It's Good For The Game

Crushers were nerfed from their throne of one-hit-killing every single marine that was in their path, so, naturally, bulls took up that position. This should prevent them from critting marines five seconds into the round.

## Changelog
:cl:
balance: Increased bull charge steps, lowered bull charge damage.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
+++ b/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
@@ -329,7 +329,7 @@
 	speed_per_step = 0.1
 	steps_for_charge = 5
 	max_steps_buildup = 10
-	crush_living_damage = 30 // This is multiplied by speed, which in this case will range from 0.1 to 1.
+	crush_living_damage = 30
 
 
 /datum/action/xeno_action/ready_charge/bull_charge/give_action(mob/living/L)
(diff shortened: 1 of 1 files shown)


David-Stormwell:
The Charge steps nerf is good, although there may be a bug with the gore ability activating more then 1 time on a single marine

code/modules/mob/living/carbon/xenomorph/charge_crush.dm
```diff
@@ -327,9 +327,9 @@
 /datum/action/xeno_action/ready_charge/bull_charge
 	charge_type = CHARGE_BULL
 	speed_per_step = 0.1
-	steps_for_charge = 3
+	steps_for_charge = 5
 	max_steps_buildup = 10
-	crush_living_damage = 40 // This is multiplied by speed, which in this case will range from 0.1 to 1.
+	crush_living_damage = 30 // This is multiplied by speed, which in this case will range from 0.1 to 1.
```
Rohesie, line 332: Will speed still range from 0.1 to 1 now?

Mouseofthecake: Pretty sure it will.

Rohesie: This will vary between `speed_per_step * steps_for_charge` and `max_steps_buildup * steps_for_charge`
The comment wasn't accurate before and it's even less with this change.
Could you fix it?

spookydonut, line 332:
```suggestion
	crush_living_damage = 30
```

PR #3792, opened 2020-04-09
Stack trace badly setup reagents

psykzz:
Helps to indentify issues when initalizing reagents.
Happened recently with the chem rework


jroinc:
Thank you!

PR #3793, opened 2020-04-09
Sunder 2: add icons and refactor projectile

psykzz:
## About The Pull Request
Refactor projectiles to use a refined proc instead of a typecheck

## Why It's Good For The Game
Better code, and visibility for xenos

## Changelog
:cl:
add: Xenos get a new hud to show sundering levels
/:cl:


psykzz:
https://streamable.com/z22ok4

code/game/data_huds.dm
```diff
@@ -318,8 +318,30 @@
...
+			holder.icon_state = "sundering0"
+		if(1 to 35)
+			holder.icon_state = "sundering25"
+		if(36 to 65)
+			holder.icon_state = "sundering50"
+		if(66 to 95)
+			holder.icon_state = "sundering75"
+		if(96 to INFINITY)
+			holder.icon_state = "sundering100"
+		else
```
Rohesie, line 342: Just remove the `else`, everything is covered.

code/modules/projectiles/projectile.dm
```diff
@@ -1026,6 +1025,15 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 	visible_message("<span class='danger'>[onlooker_feedback]</span>",
 	"<span class='xenodanger'>[victim_feedback]", null, 4)
 
+// Sundering procs
+/mob/living/proc/adjust_sunder(adjustment)
+	return 0
+
+/mob/living/proc/set_sunder(new_sunder)
+	return FALSE
```
Rohesie, line 1033:
```suggestion
	return 0
```

code/modules/projectiles/projectile.dm
```diff
@@ -1026,6 +1025,15 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
...
+
+/mob/living/proc/get_sunder()
+	return FALSE
```
Rohesie, line 1036:
```suggestion
	return 0
```

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -59,6 +59,10 @@
 
 	var/fire_resist = 1 //0 to 1; lower is better as it is a multiplier.
 
+	// *** Sunder *** //
+	var/sunder_recover = 0.1 // How much sunder is recovered per tick
```
Rohesie, line 63:
```suggestion
	var/sunder_recover = 3 // How much sunder is recovered per tick
```
A life tick happens in around 2 seconds each. Better increase this drastically.
With this value the xeno will need a full minute to regen from zero to full again.

code/modules/projectiles/projectile.dm
psykzz: I originally planned for this to return FALSE because then calling the parent could override children

Rohesie: The problem is returning a boolean in one proc and a numeric var in its redefinition. The type should be consistent, even if BYOND doesn't really distinguish between those two types.

psykzz: Well this is for set_sunder, i would expect this to only ever return true or false, based on if the parent handled it or not.

This should never return another value?

Rohesie: Alright, that works.

PR #3800, opened 2020-04-10
Fix xenos being able to drag and drop humans into machines

### LINKED ISSUE #3638: Xenos can cryo marines
## Testmerges:

#3634, #3637, #3631, #3635, #3633

## Reproduction:
As a xeno, take a SSD human and drag their sprite into a cryostasis chamber. You will then put in the human and they will be cryod properly.

psykzz:
fixes #3638
Drag and drop requires the user to be human and can only drag living

### DIFF
--- a/code/game/objects/machinery/Sleeper.dm
+++ b/code/game/objects/machinery/Sleeper.dm
@@ -471,7 +471,7 @@
 
 /obj/machinery/sleeper/MouseDrop_T(mob/M, mob/user)
 	if(!isliving(M) || !ishuman(user))
-		return 
+		return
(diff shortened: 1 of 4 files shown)


code/game/objects/machinery/Sleeper.dm
```diff
@@ -470,8 +470,8 @@
 		qdel(O)
 
 /obj/machinery/sleeper/MouseDrop_T(mob/M, mob/user)
-	if(!isliving(M))
-		return
+	if(!isliving(M) || !ishuman(user))
+		return
```
Rohesie, line 474:
```suggestion
		return
```

Issue #3795, opened 2020-04-10
Humans speaking while in ghost mode are bold text instead of normal text


XSlayer300:
## Testmerges:

#3793 #3782 #3715 #3742 #3765 #3777 #3786 #3607 #3717 #3775 #3774 #3759 #3754 #3751 #3744 #3648 #3787 #3788 #3772

## Reproduction:

1. Be a ghost
2. Every text that the humans speak are now in bold text regardless if Ghost Ears is on or off.

PR #3805, opened 2020-04-10
Fix always bold deadchat

psykzz:
fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/3795

Additionally removes the range check which is nice.

### DIFF
--- a/code/modules/mob/dead/observer/say.dm
+++ b/code/modules/mob/dead/observer/say.dm
@@ -21,7 +21,7 @@
 		else
 			to_follow = V.source
 
-	else if(client && get_dist(src, to_follow) < CEILING(view_size[1] / 2, 1))
+	else if(client && in_view_range(src, to_follow))
 		raw_message = "<b>[raw_message]</b>"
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/dead/observer/say.dm
```diff
@@ -12,15 +12,18 @@
...
 	if(radio_freq)
 		var/atom/movable/virtualspeaker/V = speaker
 		if(isAI(V.source))
 			var/mob/living/silicon/ai/S = V.source
 			to_follow = S.eyeobj
 		else
 			to_follow = V.source
-	else if(client && (to_follow in range(client.view)))
+
+	else if(client && get_dist(src, to_follow) < CEILING(view_size[1] / 2, 1))
```
Rohesie, line 24: Why not use `in_view_range(src, to_follow)`?

psykzz: spoke offline, in_view_range wasn't doing what it should have

PR #3809, opened 2020-04-10
Alamo Surgery Table Attachment

Hathkar:
## Why It's Good For The Game

A way for surgeons to do surgery near the field without having to deal with using tons and tons of sterilizing spray.

## Changelog
:cl:
add: New operating table attachment for the Alamo.
/:cl:


Hughgent:
You've got your code in there twice.

Hathkar:
Done some testing, and it works. The only thing needed is a sprite for the system before attaching it, and to add it to the Alamo's equipment printer thingy.

Rohesie:
Be sure to `QDEL_NULL` the `deployed_table` on `Destroy()`

David-Stormwell:
pretty epic ngl

code/game/objects/structures/dropship_equipment.dm
```diff
@@ -762,7 +762,40 @@
...
+		
+/obj/structure/dropship_equipment/operatingtable/Destroy()
+	QDEL_NULL(deployed_table)
+	return ..()
 
+/obj/structure/dropship_equipment/operatingtable/update_equipment()
+	if(!deployed_table)
+		return
+	if(ship_base)
+		deployed_table.layer = 3.03
```
psykzz, line 792: could this layer be a define?

Hughgent: It could 
`#define ABOVE_OBJ_LAYER 3.02 //just above all items`
is what's used for all the dropship structures.

Please test if it is functional using this define, and add a new one at 3.03 if it doesn't.

Hathkar: Changed it to 3.02, and it works fine.

psykzz: @Hathkar the point was less about the specific number but instead using a define.

If you check in `layers.dm` we define a ton of existing layers. `3.02` for example already exists there.

So instead of putting `layer = 3.03` consider `layer = ABOVE_OBJ_LAYER`

This is much clearer the intention and where the item should appear.

code/game/objects/structures/dropship_equipment.dm
```diff
@@ -762,7 +762,40 @@
...
+	if(!deployed_table)
+		return
+	if(ship_base)
+		deployed_table.layer = ABOVE_OBJ_LAYER
+		deployed_table.loc = loc
+		icon_state = "table2-idle"
+	else
+		deployed_table.layer = ABOVE_OBJ_LAYER
+		deployed_table.loc = src
+		icon_state = "table2-idle"
```
psykzz, line 798: What is `ship_base` doing for you here? From what i can tell you could just remove the check and do the same thing each time.
```suggestion
	deployed_table.layer = ABOVE_OBJ_LAYER
	deployed_table.loc = loc
	icon_state = "table2-idle"
```

Hathkar: I have no clue, I'm just doing what hugh suggested.

psykzz: @Hughgent

Hughgent: This is copied 'directly' from the mg nest, so it's replicated there as well I presume.

psykzz: the mg nest was updating their icon based on if it was attached or not.
In this case it looks like we are not.

I'd take my suggestion and then i think we'd be good to merge.

Rohesie:
`deployed_table.loc = loc`
This is not kosher, we use `forceMove` for this

PR #3813, opened 2020-04-11
Deleted phoron as it wasn't a part of the thermite recipe

IEatRox:
Just removed phoron from the thermite recipe since it wasn't working

### DIFF
--- a/code/modules/reagents/chemistry_reactions/other.dm
+++ b/code/modules/reagents/chemistry_reactions/other.dm
@@ -74,7 +74,7 @@
 /datum/chemical_reaction/thermite
 	name = "Thermite"
 	results = list(/datum/reagent/thermite = 3)
-	required_reagents = list(/datum/reagent/aluminum = 10, /datum/reagent/iron = 10, /datum/reagent/oxygen = 10, /datum/reagent/toxin/phoron = 1)
+	required_reagents = list(/datum/reagent/aluminum = 10, /datum/reagent/iron = 10, /datum/reagent/oxygen = 10
 
 /datum/chemical_reaction/lexorin
 	name = "Lexorin"


jroinc:
WELCOME TO THE GITHUB!

Few things. One, change the title to show what you're doing.

Two, test your changes by compiling and running them locally, I can show you how to do that if you need to.

If you had tested them, you'd see it doesn't compile or run.

Three, the reason it doesn't run is that you accidentally deleted a parentheses when making your change. Add the            )          back and it "should" work.

Rohesie:
It's mixing phoron and oxygen to produce dexalin.
I'd change it to something else to keep thermite scarce, instead.

jroinc:
> It's mixing phoron and oxygen to produce dexalin.
> I'd change it to something else to keep thermite scarce, instead.

Research starts with 50 sheets. That's like 1000u or something stupid. I'd rather it was balanced around work time/recipe than artificial scarcity. Additionally, plasma cutters are a thing.

Sometimes people ask for changes to your PR in the name of balance.

I agree with roh. I would remove phoron from the recipe and replace it with ammonia, so there's some light work required.

IEatRox:
Yeah I'm pretty much new to this, how do I compile it

Rohesie:
Follow this guide:
https://tgstation13.org/wiki/TGMC:Guide_to_contributing

IEatRox:
I compiled it and it worked

jroinc:
> I compiled it and it worked, though I might consider changing the ingredients to make thermite into 1 instead of 10

Is it on the same branch as the one this PR is from? Because this PR shows no updates or changes.

Also if you change the ingredients to 1 instead of 10, be sure to change the results from 30 to 3 to maintain proportions.

PR #3829, opened 2020-04-11
BC changes

Pariah919:
## About The Pull Request
BC delay penalty removed, added more scatter/accuracy malus. Reduced damage bonus by 10%

## Why It's Good For The Game
BC is really bad right now, outside of shotguns. I'm going to try to make it more worthwhile or just deletus it I guess.
## Changelog
:cl:

balance: BC stats changed a bit. Less damage bonus but fire delay is gone! The accuracy and scatter penalties are higher though.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -360,10 +360,9 @@ Defined in conflicts.dm of the #defines folder.
 	slot = "muzzle"
 	icon_state = "hbarrel"
 	attach_icon = "hbarrel_a"
-	accuracy_mod = -0.30
-	damage_mod = 0.30
-	delay_mod = 3
-	scatter_mod = 5
+	accuracy_mod = -0.35
+	damage_mod = 0.20
+	scatter_mod = 10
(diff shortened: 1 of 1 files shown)


MerryweatherLost:
**Overview**
It looks to me that the BC changes are in fact balanced, and they added a tradeoff for the lack of fire delay for a 10% decrease. I see no issues with this being merged.

As the code was reviewed and approved, I approve the balance of this item.

code/modules/projectiles/gun_attachables.dm
```diff
@@ -360,10 +360,9 @@ Defined in conflicts.dm of the #defines folder.
 	slot = "muzzle"
 	icon_state = "hbarrel"
 	attach_icon = "hbarrel_a"
-	accuracy_mod = -0.30
-	damage_mod = 0.30
-	delay_mod = 3
-	scatter_mod = 5
+	accuracy_mod = -0.35
+	damage_mod = 0.20
```
Rohesie, line 364:
```suggestion
	damage_mod = 0.2
```
No need for extra zeros after the decimal. They just make mass-replacements harder.

PR #3830, opened 2020-04-11
Las rifle

Hughgent:
## About The Pull Request

the multi functional lasrifle

## Why It's Good For The Game

that one weapon for someone who doesn't know what kind of style of play they like, so they choose the one that does everything.

## Changelog
:cl: Hughgent, Kuro
add: lasrifle that is a jack of all trades, master of none. This replaces the m43 Sunfury Lasgun
/:cl:


code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,129 @@
...
+						/obj/item/attachable/scope/mini)
+	flags_gun_features = GUN_AUTO_EJECTOR|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_ENERGY|GUN_AMMO_COUNTER
+	attachable_offset = list("muzzle_x" = 32, "muzzle_y" = 18,"rail_x" = 12, "rail_y" = 23, "under_x" = 23, "under_y" = 15, "stock_x" = 22, "stock_y" = 12)
+
+	accuracy_mult_unwielded = 0.5 //Heavy and unwieldy; you don't one hand this.
+	scatter_unwielded = 100 //Heavy and unwieldy; you don't one hand this.
+	damage_falloff_mult = 0.25
+	fire_delay = 3
+	var/current_mode = 1
+	var/mode_list = list(
```
Ghommie, line 451:
```suggestion
	var/list/datum/lasrifle/mode_list = list(
```
Linter check for typed lists.

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,129 @@
...
+						/obj/item/attachable/scope/mini)
+	flags_gun_features = GUN_AUTO_EJECTOR|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_ENERGY|GUN_AMMO_COUNTER
+	attachable_offset = list("muzzle_x" = 32, "muzzle_y" = 18,"rail_x" = 12, "rail_y" = 23, "under_x" = 23, "under_y" = 15, "stock_x" = 22, "stock_y" = 12)
+
+	accuracy_mult_unwielded = 0.5 //Heavy and unwieldy; you don't one hand this.
+	scatter_unwielded = 100 //Heavy and unwieldy; you don't one hand this.
+	damage_falloff_mult = 0.25
+	fire_delay = 3
+	var/current_mode = 1
+	var/list/datum/lasrifle/mode_list = list(
```
Ghommie, line 451:
```suggestion
	var/list/datum/lasrifle/base/mode_list = list(
```
I skipped a path section.

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,129 @@
...
+						/obj/item/attachable/attached_gun/shotgun,
+						/obj/item/attachable/scope/mini)
+	flags_gun_features = GUN_AUTO_EJECTOR|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_ENERGY|GUN_AMMO_COUNTER
+	attachable_offset = list("muzzle_x" = 32, "muzzle_y" = 18,"rail_x" = 12, "rail_y" = 23, "under_x" = 23, "under_y" = 15, "stock_x" = 22, "stock_y" = 12)
+
+	accuracy_mult_unwielded = 0.5 //Heavy and unwieldy; you don't one hand this.
+	scatter_unwielded = 100 //Heavy and unwieldy; you don't one hand this.
+	damage_falloff_mult = 0.25
+	fire_delay = 3
+	var/current_mode = 1
```
Rohesie, line 450: Use human-readable defines for the modes.

Hughgent: Gonna argue against that. We have already gone though a patch where every gun value was defined. And we know that got changed quite a while back.

On the point of `current_mode`, there are no switch statements, it's purely used as a counter to keep track of current position.

Rohesie: Does it track different fire modes or something purely numeric?

Hughgent: purely numeric. it tracks the position in the list of firemodes.

As nice as STANDARD_FIREMODE 1 and AUTO_FIREMODE 2 would be, we'd only ever actually have the STANDARD_FIREMODE in code.

Hughgent:
@Kuro-ai If you would be so kind as to checkout this branch and update the `items_lefthand_1.dmi` and `items_righthand_1.dmi` to be more presentable.

Secondly, how would you like this integrated? are we going to replace the m43 sunfury in it's entirety, or should it instead be a req only weapon in it's current state due to it being a direct upgrade over the m43?

Hughgent:
Integration choice was "replace the m43"

Hughgent:
welp, looks like I did something bad with mapmerging

code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -213,7 +213,7 @@
 	GLOB.cargo_ammo_vendors.Remove(src)
 
 /obj/machinery/vending/lasgun
-	name = "ColMarTech Lasgun Field Charger"
+	name = "ColMarTech lasrifle Field Charger"
```
KwinOriginal, line 216: I'm not the biggest code diver but wouldn't it sound somewhat better if you just put the full name here instead of just "lasrifle"? It looks weird but is purely cosmetic.

code/game/objects/machinery/vending/new_marine_vendors.dm
```diff
@@ -350,11 +350,6 @@ GLOBAL_LIST_INIT(marine_selector_cats, list(
 		/obj/item/attachable/suppressor = list(CAT_ATT, "Suppressor", 0,"black"),
 		/obj/item/attachable/extended_barrel = list(CAT_ATT, "Extended barrel", 0,"orange"),
 		/obj/item/attachable/compensator = list(CAT_ATT, "Recoil compensator", 0,"black"),
-		/obj/item/attachable/efflens = list(CAT_ATT, "M43 Efficient lens", 0,"black"),
-		/obj/item/attachable/focuslens = list(CAT_ATT, "M43 Focus lens", 0,"black"),
-		/obj/item/attachable/widelens = list(CAT_ATT, "M43 Wide lens", 0,"black"),
-		/obj/item/attachable/heatlens = list(CAT_ATT, "M43 Heat lens", 0,"black"),
-		/obj/item/attachable/pulselens = list(CAT_ATT, "M43 pulse lens", 0,"black"),
```
KwinOriginal, line 357: Are those attachments lost? I saw that a bunch of regular ones were re-added. Why's that? Sprite pain?

Pariah919: Bloat in the vendor probably.

Hughgent: They still exist in code, just with the total replacement of the m43, there isn't any reason to keep them. No (normally) accessible gun can use them.

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,129 @@
...
+	if(current_mode >= max_overcharge_mode)
+		current_mode = 1
+	else
+		current_mode += 1
+	
+	playsound(user, 'sound/weapons/emitter.ogg', 5, 0, 2)
+	charge_cost = initial(mode_list[current_mode].charge_cost)
+	ammo = GLOB.ammo_list[initial(mode_list[current_mode].ammo)]
+	fire_delay = initial(mode_list[current_mode].fire_delay)
+	fire_sound = initial(mode_list[current_mode].fire_sound)
```
psykzz, line 522: Why the use of `initial` all over?

Hughgent: No particular reason for the method.

I need it to get the relevant information from each datum in the list, and `initial` does that.

code/modules/projectiles/guns/energy.dm
```diff
@@ -150,7 +150,7 @@
 	ammo = /datum/ammo/energy/lasgun/M43
```
spookydonut, line 150: missed a m43 reference here?

code/modules/projectiles/guns/energy.dm
```diff
@@ -150,7 +150,7 @@
...
 	ammo_diff = null
 	cell_type = null
-	charge_cost = M43_STANDARD_AMMO_COST
+	charge_cost = TX73_STANDARD_AMMO_COST
```
spookydonut, line 153: maybe just change it to a generic name instead of changing all the defines every time someone renames this gun

code/modules/projectiles/guns/energy.dm
```diff
@@ -205,21 +205,21 @@
 			playsound(user, 'sound/machines/buzz-two.ogg', 15, 0, 2)
 			to_chat(user, "<span class='warning'>You attempt to toggle on [src]'s overcharge mode but you have no battery loaded.</span>")
 			return
-		if(cell.charge < M43_OVERCHARGE_AMMO_COST)
+		if(cell.charge < TX73_OVERCHARGE_AMMO_COST)
```
spookydonut, line 208: same

code/modules/projectiles/guns/energy.dm
spookydonut: rename it to mode_index or something

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,129 @@
...
+	fire_delay = 8
+	fire_sound = 'sound/weapons/guns/fire/laser_rifle_2.ogg'
+	message_to_user = "You set the Lasrifle's charge mode to spread."
+	icon_state = "tx73_spread"
+
+/obj/item/weapon/gun/energy/lasgun/lasrifle/unique_action(mob/user)
+	return switch_modes(user)
+
+/obj/item/weapon/gun/energy/lasgun/lasrifle/proc/switch_modes(mob/user)
+	var/max_overcharge_mode = length(mode_list)
```
spookydonut, line 512: `current_mode = WRAP(current_mode + 1, 1, length(mode_list))`

`#define WRAP(val, min, max) clamp(( min == max ? min : (val) - (round(((val) - (min))/((max) - (min))) * ((max) - (min))) ),min,max)`

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,129 @@
...
+						/obj/item/attachable/scope/mini)
+	flags_gun_features = GUN_AUTO_EJECTOR|GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_ENERGY|GUN_AMMO_COUNTER
+	attachable_offset = list("muzzle_x" = 32, "muzzle_y" = 18,"rail_x" = 12, "rail_y" = 23, "under_x" = 23, "under_y" = 15, "stock_x" = 22, "stock_y" = 12)
+
+	accuracy_mult_unwielded = 0.5 //Heavy and unwieldy; you don't one hand this.
+	scatter_unwielded = 100 //Heavy and unwieldy; you don't one hand this.
+	damage_falloff_mult = 0.25
+	fire_delay = 3
+	var/current_mode = 1
+	var/list/datum/lasrifle/base/mode_list = list(
```
spookydonut, line 451:
```suggestion
	var/static/list/datum/lasrifle/base/mode_list = list(
```

code/modules/projectiles/guns/energy.dm
spookydonut: initial is correct here to avoid having to instance the datums

code/game/objects/machinery/vending/new_marine_vendors.dm
KwinOriginal: Can't they just be renamed and used for the new Laser Rifle? I mean, laser rifle is laser rifle, regardless of the name and looks, or am I missing something? 
Bloating is an effect too, that's true.

code/modules/projectiles/guns/energy.dm
Hughgent: nah, it shoots the current lasgun bullet/bullets. I feel that copying the current lasgun bullets just to rename the typepath is just code duplication, and another thing to keep in mind when balancing.

code/modules/projectiles/guns/energy.dm
Hughgent: I guess I'll go with `ENERGY_STANDARD_AMMO_COST` or perhaps just remove the define? That would go more in line with the whole 'no more defines for weapons'

spookydonut: removing the define here is okay but below you should keep it for the OVERCHARGE_AMMO_COST because its unclear what you're checking otherwise

code/game/objects/machinery/vending/new_marine_vendors.dm
Pariah919: They're lens created to do what the current one has standard now, no point to them.

code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -228,10 +228,10 @@
 	machine_current_charge = 50000 //integrated battery for recharging energy weapons. Normally 10000.
 	machine_max_charge = 50000
 
-	product_ads = "Lasgun running low? Recharge here!;Need a charge?;Power up!;Electrifying!;Empower yourself!"
+	product_ads = "Lasrifle running low? Recharge here!;Need a charge?;Power up!;Electrifying!;Empower yourself!"
 	products = list(
-					/obj/item/cell/lasgun/M43 = 10,
-					/obj/item/cell/lasgun/M43/highcap = 2,
+					/obj/item/cell/lasgun/lasrifle = 10,
+					/obj/item/cell/lasgun/lasrifle/highcap = 2,
```
psykzz, line 234: sort out the indent

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,125 @@
...
+						/obj/item/attachable/lasersight,
+						/obj/item/attachable/gyro,
+						/obj/item/attachable/flashlight,
+						/obj/item/attachable/bipod,
+						/obj/item/attachable/magnetic_harness,
+						/obj/item/attachable/attached_gun/grenade,
+						/obj/item/attachable/scope,
+						/obj/item/attachable/attached_gun/flamer,
+						/obj/item/attachable/attached_gun/shotgun,
+						/obj/item/attachable/scope/mini)
```
psykzz, line 442: sort out the indent

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,125 @@
...
+	scatter_unwielded = 100 //Heavy and unwieldy; you don't one hand this.
+	damage_falloff_mult = 0.25
+	fire_delay = 3
+	var/mode_index = 1
+	var/static/list/datum/lasrifle/base/mode_list = list(
+						/datum/lasrifle/base/standard,
+						/datum/lasrifle/base/auto,
+						/datum/lasrifle/base/overcharge,
+						/datum/lasrifle/base/heat,
+						/datum/lasrifle/base/spread)
```
psykzz, line 456: sort out the indent

code/datums/supplypacks.dm
```diff
@@ -383,12 +383,12 @@ WEAPONS
...
 /datum/supply_packs/gun/lasrifle
 	contains = list(
-					/obj/item/weapon/gun/energy/lasgun/M43,
-					/obj/item/weapon/gun/energy/lasgun/M43,
-					/obj/item/cell/lasgun/M43,
-					/obj/item/cell/lasgun/M43
+					/obj/item/weapon/gun/energy/lasgun/lasrifle,
+					/obj/item/weapon/gun/energy/lasgun/lasrifle,
+					/obj/item/cell/lasgun/lasrifle,
+					/obj/item/cell/lasgun/lasrifle
```
psykzz, line 389: sort out the indent

code/modules/projectiles/guns/energy.dm
psykzz: Spooky is right, this reference needs to be renamed unless you can tell me of another weapon that makes use of it.

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,125 @@
...
+	else
+		SEND_SIGNAL(src, COMSIG_GUN_FIREMODE_TOGGLE, GUN_FIREMODE_SEMIAUTO, user.client)
+	
+	base_gun_icon = initial(mode_list[mode_index].icon_state)
+	update_icon()
+
+	to_chat(user, initial(mode_list[mode_index].message_to_user))
+
+
+	if(user)
```
psykzz, line 531: Just check user at the start and return early.

code/game/objects/machinery/vending/marine_vending.dm
Hughgent: What about it?

This is the exact same indent as it's always been.

code/modules/projectiles/guns/energy.dm
Hughgent: the m43 lasgun uses it. which is technically admin spawn only right now.

code/game/objects/machinery/vending/marine_vending.dm
psykzz: No point having it so far indented, it just makes reviews harder.

Hughgent: It would make the code less pretty though

image

code/modules/projectiles/guns/energy.dm
Hughgent: Done

code/game/objects/machinery/vending/marine_vending.dm
Hughgent: Secondly, how are you doing your reviewing?

image

This seems easy enough to follow from my perspective.

Pariah919: I'm pretty sure everything in that file is indented that way, would need to change everything else.

Pariah919: Everything in that file is indented like this. It'd look out of place if it's just this box. I'd argue.

Hughgent: Everything in all files are indented like this.

Pariah919: Should be a seperate PR then, if he wants it changed.

code/game/objects/machinery/vending/marine_vending.dm
Rohesie: Removing `ColMarTech` would be positive as well.

code/modules/projectiles/guns/energy.dm
Rohesie: A static list makes you lose a lot on flexibility. Use a global list instead.

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,126 @@
...
+/obj/item/weapon/gun/energy/lasgun/lasrifle/unique_action(mob/user)
+	return switch_modes(user)
+
+/obj/item/weapon/gun/energy/lasgun/lasrifle/proc/switch_modes(mob/user)
+	if(!user)
+		return FALSE
+	
+	mode_index = WRAP(mode_index + 1, 1, length(mode_list))
+	
+	playsound(user, 'sound/weapons/emitter.ogg', 5, 0, 2)
```
Rohesie, line 517:
```suggestion
	playsound(user, 'sound/weapons/emitter.ogg', 5, FALSE, 2)
```

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,126 @@
...
+	fire_sound = 'sound/weapons/guns/fire/laser_rifle_2.ogg'
+	message_to_user = "You set the Lasrifle's charge mode to spread."
+	icon_state = "tx73_spread"
+
+/obj/item/weapon/gun/energy/lasgun/lasrifle/unique_action(mob/user)
+	return switch_modes(user)
+
+/obj/item/weapon/gun/energy/lasgun/lasrifle/proc/switch_modes(mob/user)
+	if(!user)
+		return FALSE
```
Rohesie, line 513: Is this check necessary? There should always be a `user` when this is invoked, no?
And if there is not, runtiming would be good so we know the issue exists.
If you really think this is necessary to keep replace the `return FALSE` with something like `CRASH("switch_modes called for [type] with no user.")`

code/game/objects/machinery/vending/marine_vending.dm
psykzz: This is what im trying to avoid 
image

I wont block merging and would just end up doing this myself later.

code/modules/projectiles/guns/energy.dm
spookydonut: they aren't being instanced and the list doesnt change, static is the right choice here

code/modules/projectiles/guns/energy.dm
spookydonut: yeah problem hiding, add a CRASH if you think its happening

code/modules/projectiles/guns/energy.dm
Hughgent: Done

code/modules/projectiles/guns/energy.dm
Hughgent: Done

Hughgent:
I could use some advice on the following

```
//Ammo/Charge functions
/obj/item/weapon/gun/energy/lasgun/update_icon(mob/user)
	var/cell_charge = (!cell || cell.charge <= 0) ? 0 : CEILING((cell.charge / max(cell.maxcharge, 1)) * 100, 25)
	icon_state = "[base_gun_icon]_[cell_charge]"
	update_mag_overlay(user)
	update_item_state(user)


/obj/item/weapon/gun/energy/lasgun/update_item_state(mob/user)
	. = item_state
	var/cell_charge = (!cell || cell.charge <= 0) ? 0 : CEILING((cell.charge / max(cell.maxcharge, 1)) * 100, 25)
	item_state = "[base_gun_icon]_[cell_charge][flags_item & WIELDED ? "_w" : ""]"
	if(. != item_state && ishuman(user))
		var/mob/living/carbon/human/human_user = user
		if(src == human_user.l_hand)
			human_user.update_inv_l_hand()
		else if (src == human_user.r_hand)
			human_user.update_inv_r_hand()
```

Switching modes updates the `base_gun_icon` of the lasrifle, giving it the differing nozzles. and currently, if you wield or unwield the rifle in any mode other than standard, the on mob sprite disappears. These procs handle the icon state updating for the lasrifle.

Should I not have the `base_gun_icon` be updated or should I add another argument which forces the in hands for the lasrifle? What should be done in this situation?

psykzz:
Im not sure i follow why its going invisible unless the icon_state doesn't exist.
for `update_item_state` can you not just `item_state = "[icon_state][flags_item & WIELDED ? "_w" : ""]"`?

Hughgent:
You've basically got it. the multiple item_state names don't exist, and it would be somewhat of a bloating of the right/left hand 1 files to add them all.

Hughgent:
Essentially, I need for the lasrifle, for the item_state to always be `item_state = "tx73_[cell_charge][flags_item & WIELDED ? "_w" : ""]"` but this doesn't work with the lasgun pulsegun and any other future energy weapons.

Mr-Worth:
Balance request after watching some of the rounds as it is TM'd today. The ability for the bolt to set what it hits on fire is a bit too powerful. It is doing a really good job of suppressing the xenos as they try to push. 

Just a thought.

psykzz:
> You've basically got it. the multiple item_state names don't exist, and it would be somewhat of a bloating of the right/left hand 1 files to add them all.

I'd remove the name tx73 from the mode base icon and set it as the icon_state alone.
Then on `update_icons()`

Then assign icons like 
`item_state = "[initial(icon_state)][mode_icon]_[cellcharge][flags_item & WIELDED ? "_w" : ""]"`

On the modes that don't have icons don't define them, just set them all to empty or whatever works for you.

code/modules/projectiles/guns/energy.dm
```diff
@@ -411,3 +411,126 @@
...
+	icon_state = "tx73_spread"
+
+/obj/item/weapon/gun/energy/lasgun/lasrifle/unique_action(mob/user)
+	return switch_modes(user)
+
+/obj/item/weapon/gun/energy/lasgun/lasrifle/proc/switch_modes(mob/user)
+	if(!user)
+		CRASH("switch_modes called with no user.")
+	
+	mode_index = WRAP(mode_index + 1, 1, length(mode_list))
```
psykzz, line 515:
```suggestion
	if(!do_after(user, 0.25 SECONDS, TRUE, user, BUSY_ICON_GENERIC))
		return
	mode_index = WRAP(mode_index + 1, 1, length(mode_list))
```

This is my suggestion

code/modules/projectiles/guns/energy.dm
Rohesie: The ways these datums are used are basically unmanageable global lists.
What's their advantage?
Both could be substituted by a referenced global list at no cost.

Rohesie: Static gives you no advantages over a referenced global list and lowers flexibility.
It allows for no variants nor badminery.

code/modules/projectiles/guns/energy.dm
```diff
@@ -516,7 +516,7 @@
 		to_chat(user, "<span class ='warning'>You fumble the Lasrifle's charge mode.</span>")
 		return
 
-	mode_index = WRAP(mode_index + 1, 1, length(mode_list))
+	mode_index = WRAP(mode_index + 1, 1, length(mode_list)+1)
```
spookydonut, line 519: ffs whoever wrote WRAP `max is exclusive`

PR #3832, opened 2020-04-12
Github Auto generate documentation

psykzz:
## About The Pull Request

This adds a new github action that will take the latest push to master, run dmdoc against it and push the output to `gh-pages`

## Why It's Good For The Game

We plan to start enforcing documentation on vars / procs / types, what good is that if we don't  have somewhere to view it all.


psykzz:
Example of them working for me on my fork - https://psykzz.github.io/TerraGov-Marine-Corps/index.html

spookydonut:
Use `tools/travis/install_spacemandmm_suite.sh dmdoc` since thats what its there for including the correct versioning.

You can also just do this with travis like tg if thats easier

psykzz:
I liked the idea of github actions but noticed they have a limit of 2000 minutes a month. I doubt we'd ever hit that but i'd rather not get to a point where we potentially migrate other things and hit the limt.

Will convert this to use travis instead similar to tgstation.

PR #3835, opened 2020-04-12
Fixed thermite recipe

IEatRox:
- Removed phoron to fix thermite recipe
- Changed the proportions of the thermite ingredients to balance it out


Rohesie:
This makes thermite dirt cheap and reduces scarcity a ton.

jroinc:
> This makes thermite dirt cheap and reduces scarcity a ton.

Thermite is a single-use thing that burns one wall down and requires chemistry to make.

Plasma cutters do the same thing, but are 10-use. C-4 does the same thing, detpacks do the same thing, hell, ungas take down doors via knifespam every round.

I'm fine with it becoming much more common.

Rohesie:
You can splash it in 5u doses to delete walls, bypassing the need of a plasma cutter.

terranaut1:
Make the thermite amount required to burn through the wall higher, 30 at least, maybe 50 to try to find some balance between storage space used vs. walls deleted

jroinc:
> Make the thermite amount required to burn through the wall higher, 30 at least, maybe 50 to try to find some balance between storage space used vs. walls deleted

60u sounds reasonable, that's a bottle per wall, but if you large-beaker it up, you can get 2 shots, assuming you use the finicky transfer mechanics.

spookydonut:
Themite mechanics are bad, removing walls shouldnt be instant regardless of how much chems you splash around

psykzz:
I dont see this getting merged without some more positive opinions based on balance.
I poked on discord for others to give it their thumbs up, see what happens.

Rohesie:
This shouldn't have been merged, and should be reverted.

PR #3848, opened 2020-04-13
Make hiveminds right clickable again

psykzz:
This is a pain to not be able to do this as an admin.
This was added as an attempt to fix hiveminds but didn't help.


jroinc:
thank you

spookydonut:
shouldnt have been merged in the first place since it was clearly a hacky non-fix

PR #3849, opened 2020-04-13
Fixes hivemind dying to explosions

psykzz:
Additional cleansup updatehealth on godmode


code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -181,6 +181,18 @@
 /mob/living/carbon/xenomorph/hivemind/setBrainLoss(amount)
 	return FALSE
 
+/mob/living/carbon/xenomorph/hivemind/ex_act()
```
spookydonut, line 184: just add the flag checks to the xenomorph/ version of these procs, its really ugly to see empty returns like this and reduces maintainability when it comes to stuff like SHOULD_CALL_PARENT

PR #3854, opened 2020-04-13
Fix explosion not actually creating fires

psykzz:
Now the explosion proc will actually create fires

### DIFF
--- a/code/game/objects/explosion.dm
+++ b/code/game/objects/explosion.dm
@@ -119,13 +119,7 @@
 				AM.ex_act(dist)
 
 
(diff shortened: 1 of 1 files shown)


jroinc:
Explosions from grenades create a 4-tile ring of fire around a central tile. Intended?

psykzz:
Yes the only difference is the fire doesn't have a probability to spawn, and will spawn 100% of the time but with random life.

I can add the randomness back to the proc, should be easy enough

David-Stormwell:
keep it, it's pretty epic ngl

PR #3857, opened 2020-04-14
Xeno Bump Attack toggled on by Default

BobSlim:
## About The Pull Request

Xenos (and everyone else, I suppose) begins with bump attack on by default.

## Why It's Good For The Game

It was on rohesie's simple additions

Don't yell at me, most aliens immediately turn it on anyway.

## Changelog
:cl:
tweak: xeno bump attack is on by default
/:cl:

### DIFF
--- a/code/datums/components/bump_attack.dm
+++ b/code/datums/components/bump_attack.dm
@@ -1,5 +1,5 @@
 /datum/component/bump_attack
-	var/active = FALSE
+	var/active = TRUE
 	var/bump_action_path
 	var/datum/action/bump_attack_toggle/toggle_action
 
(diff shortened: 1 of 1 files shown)


TiviPlus:
based and bumppilled

code/datums/components/bump_attack.dm
```diff
@@ -20,6 +20,7 @@
 	toggle_action.give_action(parent)
 	toggle_action.update_button_icon(active)
 	RegisterSignal(toggle_action, COMSIG_ACTION_TRIGGER, toggle_path)
+	RegisterSignal(parent, COMSIG_MOVABLE_BUMP, bump_action_path)
```
spookydonut, line 23:
```suggestion
	if(active)
		RegisterSignal(parent, COMSIG_MOVABLE_BUMP, bump_action_path)
```

BobSlim: I'm not too sure what problem this solves - that line of code is in /datum/component/bump_attack/Initialize(). I don't actually know when it's called, but I think it's only called once per mob - and active is hard-coded = true at the beginning of the component.

Is this to make it easier to alter bump attack to FALSE in the future? I suppose that would be reasonable.

spookydonut: ideally you should be passing active as a parameter to the component initialize, then xenos adding the component pass active from some config value rather than hard coding it

PR #3862, opened 2020-04-15
Neurotoxin Sting / Sentinel Rework

BobSlim:
## About The Pull Request

Allows sentinels to use the neurotox sting on nested hosts, because that's what the neurotox sting is for. 

Makes neurotox deal reduced stamina damage, but over it's entire duration. Will aid in captures, and keeping hosts down.

Reduces the purge rate of larval accelerant on growth toxin from 10/life() to 0.4/life(), meaning that it's actually worth injecting.

Increases cooldown of Sentinel Neurotoxin Sting

## Why It's Good For The Game

Xenomorphs KNOW that hosts can break out of their nests, they should be able to take reasonable precautions to prevent that. If a sentinel is in the nest, the hosts should not be able to escape readily. Increased cooldown on the sting deincentivizes sitting in the nest and continually stinging people. Sentinels have a choice between using their CD on a new capture, or keeping old captures secure.

Adds more reason to be a sentinel rather than a spitter, and fits their desired purpose better. Allows sentinels more role in capturing, as Neurotoxin immediately begins its (smaller) effect.

Incentivises captures by making them easier. Xeno gameplay should always offer a choice between capturing and killing; and killing is often way too easy. New Neurotoxin means that, while you're dragging a host away, you don't need to have four xenos throttling them, especially if they're wearing armour and you've already worn them down.

## Changelog
:cl:
del: sentinels are no longer prevented from injecting nested hosts with Neurotoxin.
balance: Drastically lowered the rate at which Larval Accelerant purges Neurotoxin (10 -> 0.4).
balance: Stamina damage of neurotoxin is made more flat and front-heavy, rather than a huge spike after 10 cycles.
balance: Neurotoxin sting Cooldown is increased from 12s to 20s.
/:cl:


jroinc:
-1.

One, you have a massive code "error". Neurotox OD's at 36. The logarithm of 30 is ~1.5.
~ooh, a whole ONE POINT FIVE stamina damage! I'm shaking in my boots~

Two, assuming that's fixed being able to neurotox and forget removes any counterplay from captured ungas. There are 2 ways this can go. Either the damage over time will be too low to do anything, and they'll heal up to full real quick while nested, or the damage will be high enough that the unga will stay in staminacrit forever. CM permanesting's wildly unpopular there, and I'd rather not see it become a thing here.

Three, current neurotox is balanced for severe damage over a decent period of time. 30u will drop someone into staminacrit for a decent amount of time, but you have to wait about 2 minutes. Old neuro was a meme in that you get over X amount (20u), you hardcrit and nothing can stop it.

Four, I'd much rather see additional sources of neurotoxin added to assist in capturing, I'm planning on returning neuroclaws to the defiler. Sentinel MIGHT get neurospit, if it's needed, and I currently don't see it needing that.

BobSlim:
reworking it into smaller chunks. I realised that there are two separate things in this particular commit which aren't bonded to each other.

PR #3864, opened 2020-04-15
Neurotoxin improved for capture mechanics, altered overdose

BobSlim:
## About The Pull Request

Alters Neuro Toxin slightly. 

## Why It's Good For The Game

Allows sentinels (and other neuro-toxin applying castes) to play a greater role in captures. Note that any stamina damage will prevent stamina regen. If a sentinel gets the neuro-sting off, the target will no longer regenerate stamina. 

Once the target is downed by tackles, it will stay downed until the neuro-toxin is out of their system. No more of the target getting up mid-pull and punching you if you've already taken them down fair and square. Either that, or the target can be stalked until they fall over due to the neurotoxin being in their system too long.

It also removes the fact that a single 3-part sting will always overdose the target, and adds a small amount of toxin damage to overdoses. Previously, overdosing a healthy host would not do anything significant, as 2 oxy damage is quickly healed via breathing. Now it does something small.

The toxins

## Changelog
:cl:
balance: Xeno Neurotoxin begins its effects immediately, and ramp up over time.
balance: Xeno Neurotoxin deals a small amount of toxin damage if overdosed.
fix: A single Neurotoxin Sting will no longer trigger an overdose. 
/:cl:


jroinc:
Less steep pattern, but immediate damage. Given the 3 neuro counters in existence, I'd say this is fair.

Take a look at the metabolism rates, tho. A single sting is 10u, metabolism 0.4u, 25 ticks. Half-a-point scaling damage means that by tick 8, running's disabled, and by tick 20, they're on their way towards permacrit.

Assuming 12-second recharge time, and a life tick being ~2 seconds, that seems reasonable for people with medics, but very strong against unsupported people. Still, they have 40 seconds to seek help, and that seems fair for a channeled melee ability.

PR #3867, opened 2020-04-16
Cleanup some terms

psykzz:
Im not sure how much this is needed, but just taking the safe route.


code/__DEFINES/atmospherics.dm
```diff
@@ -23,9 +23,9 @@
 #define BODYTEMP_COLD_DAMAGE_LIMIT_TWO		240.15
 #define BODYTEMP_COLD_DAMAGE_LIMIT_THREE	120.15
 
-#define PRESSURE_DAMAGE_COEFFICIENT 4 //The amount of pressure damage someone takes is equal to (pressure / HAZARD_HIGH_PRESSURE)*PRESSURE_DAMAGE_COEFFICIENT, with the maximum of MAX_PRESSURE_DAMAGE
-#define MAX_HIGH_PRESSURE_DAMAGE 4	//This used to be 20... I got this much random rage for some retarded decision by polymorph?! Polymorph now lies in a pool of blood with a katana jammed in his spleen. ~Errorage --PS: The katana did less than 20 damage to him :(
-#define LOW_PRESSURE_DAMAGE 2 	//The amounb of damage someone takes when in a low pressure area (The pressure threshold is so low that it doesn't make sense to do any calculations, so it just applies this flat value).
+#define PRESSURE_DAMAGE_COEFFICIENT 4 ///The amount of pressure damage someone takes is equal to (pressure / HAZARD_HIGH_PRESSURE)*PRESSURE_DAMAGE_COEFFICIENT, with the maximum of MAX_PRESSURE_DAMAGE
+#define MAX_HIGH_PRESSURE_DAMAGE 4	///Maximum amount of damage someone can take
```
TiviPlus, line 27:
```suggestion
#define MAX_HIGH_PRESSURE_DAMAGE 4	///Maximum amount of damage someone can take from being in a too-high pressure area
```

code/modules/mob/living/living_defines.dm
```diff
@@ -6,19 +6,19 @@
...
-	var/brainloss = 0	//'Retardation' damage caused by someone hitting you in the head with a bible or being infected with brainrot.
+	var/bruteloss = 0	///Brutal damage caused by brute force (punching, being clubbed by a toolbox ect... this also accounts for pressure damage)
+	var/fireloss = 0	///Burn damage caused by being way too hot, too cold or burnt.
+	var/oxyloss = 0	///Oxygen depravation damage (no air in lungs)
+	var/toxloss = 0	///Toxic damage caused by being poisoned or radiated
+	var/staminaloss = 0 ///Stamina
+	var/cloneloss = 0	///Damage caused by being cloned or ejected from the cloner early
+	var/brainloss = 0	///Bbrain damage caused by someone hitting you in the head with a bible or being infected with brainrot.
 	var/radiation = 0	//If the mob is irradiated.
 	var/drowsyness = 0
```
TiviPlus, line 23: No comments for these?

code/modules/mob/living/living_defines.dm
```diff
@@ -6,19 +6,19 @@
 	var/resize = RESIZE_DEFAULT_SIZE //Badminnery resize
 
 	//Health and life related vars
-	var/maxHealth = 100 //Maximum health that should be possible.
-	var/health = 100 	//A mob's health
-	var/health_threshold_dead = -100
-	var/health_threshold_crit = 0
+	var/maxHealth = 100 ///Maximum health that should be possible.
+	var/health = 100 	/// Mob's current health
+	var/health_threshold_dead = -100 /// Health at which a mob dies
```
Rohesie, line 11:
```suggestion
	///Maximum health that should be possible.
	var/maxHealth = 100
	/// Mob's current health
	var/health = 100 
	/// Health at which a mob dies
	var/health_threshold_dead = -100
```
This is the format in where the comments will show.

code/modules/mob/living/living_defines.dm
```diff
@@ -6,19 +6,19 @@
...
 	//Damage related vars, NOTE: THESE SHOULD ONLY BE MODIFIED BY PROCS
-	var/bruteloss = 0	//Brutal damage caused by brute force (punching, being clubbed by a toolbox ect... this also accounts for pressure damage)
-	var/fireloss = 0	//Burn damage caused by being way too hot, too cold or burnt.
-	var/oxyloss = 0	//Oxygen depravation damage (no air in lungs)
-	var/toxloss = 0	//Toxic damage caused by being poisoned or radiated
-	var/staminaloss = 0 //Stamina
-	var/cloneloss = 0	//Damage caused by being cloned or ejected from the cloner early
-	var/brainloss = 0	//'Retardation' damage caused by someone hitting you in the head with a bible or being infected with brainrot.
+	var/bruteloss = 0	///Brutal damage caused by brute force (punching, being clubbed by a toolbox ect... this also accounts for pressure damage)
+	var/fireloss = 0	///Burn damage caused by being way too hot, too cold or burnt.
```
TiviPlus, line 16: Does adding /// on the same line work? /tg/ has it always in a
```
///hello
var/hi_there
```
format

psykzz: Both work, i would say it's preference

code/modules/mob/living/living_defines.dm
psykzz: It works in both cases,, and for vars like this i'd prefer it inline, unless the line is already massive.

image

code/modules/mob/living/living_defines.dm
psykzz: I tried to limit the changes because this was mostly around wording but i editing the lines above and below where it made sense.

code/modules/clothing/suits/jobs.dm
```diff
@@ -1,16 +1,22 @@
-/*
-* Job related
+/**
+* Job related clothing
 */
 
-//Botonist
+///Botonist
```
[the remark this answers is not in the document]
psykzz: To be fair it was used more to denote the groups, and prob shouldn't have the triple quotes, i originally added it for each but saw no benefit from it.

The description tells you more than the comment does.

code/modules/mob/living/living_defines.dm
spookydonut: it doesnt work properly if you have
```
/type
    var/foo /// doc comment for var

/**
  * doc comment for proc
 */
/type/proc/bar()
```

psykzz: That sucks, okay ill migrate to the line before

<details>
<summary>For those interested how it turns out</summary>
Code:

```

/**
* Page comment for spooky.dm
*/
/mob/spooky
	var/testing = TRUE  /// This sucks

/**
* This is the proc doc comment
*/
/mob/spooky/proc/test_proc(var/arg_one)
	return testing /// Returns testing

```

Docs

image

</details>

PR #3875, opened 2020-04-18
Refactor hud procs

psykzz:
## About The Pull Request
This moves processing of huds onto the hud instead of the mob

## Why It's Good For The Game
It doesn't affect game

## Changelog
:cl:
refactor: refactors hud code so that the procs live within the hud not the mob
/:cl:


spookydonut:
why move this out of life when the stuff that it works from is all in life?

is there any profiling to support that this needs doing?  surely theres more costly stuff in life that could be done easier

psykzz:
I agree, this was me late night trying a thing.

PR #3880, opened 2020-04-18
Powerloader skill increases to RO, CE, Combat Engineers, and early model synths; RO can now flash marines

XElectricX:
## About The Pull Request

RO, CE, and early gen synths now have level 4 mech skills, on par with STs. Combat engineer gets level 1. RO gains level 1 police skill and a flash spawns in the RO locker.

## Why It's Good For The Game

RO and CE were outclassed by the ST in using the Ripley. Why? I don't know, but now they are just as good so they can do quicker deliveries. Also raised the skill of early synths because synths are one of the primary mech users so leaving it at level 2 is a hindrance and does not make sense.

Combat engineer gets it at level 1 since it seems reasonable for them to be able to use it slightly faster than a regular marine.

RO is an officer so they should have had flash access regardless. They are also the ones who need it the most besides MPs to subdue marines who break into the department and not have to resort to a gun or a fist fight.

## Changelog
:cl:
tweak: RO lockers spawn a flash, which the RO can now use
balance: RO, CE, and early gen synths have level 4 powerloader skill
balance: Combet engis gain level 1 powerloader skill
/:cl:

### DIFF
--- a/code/__DEFINES/skills.dm
+++ b/code/__DEFINES/skills.dm
@@ -115,10 +115,10 @@
 //hidden
 //proficiency with powerloader, changes powerloader speed.
 #define SKILL_POWERLOADER_DEFAULT	0
-#define SKILL_POWERLOADER_DABBLING	1 //Pilot
-#define SKILL_POWERLOADER_TRAINED	2 //Req
-#define SKILL_POWERLOADER_PRO		3 //ST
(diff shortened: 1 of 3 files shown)


TiviPlus:
RO flash, speedmerge 👍

PR #3889, opened 2020-04-19
Add regenerating resin walls for xenos

psykzz:
## About The Pull Request

Resin walls are being replaced with regenerating walls
These walls generally start off with lower health than normal walls but over time increase their max health.

Hivelord can no longer thicken walls / structures, but instead just straight builds the better variants.

Additionally adds the ability for xenos to tear down walls after a do_after instead of dmg.

## Why It's Good For The Game

This benefits the idea of actual hives vs combat walls

## Changelog
:cl:
add: Regenerating resin walls
/:cl:


XElectricX:
Combat walls were fine and weren't tanky at all. Gave teammates breathing room to heal. This also removed the Hivelord's ability to build things in a 3x3 area around it, now it can only build on the tile it stands on.

XElectricX:
Building walls and tearing them down in the Alamo creates space tiles.

psykzz:
Fixed the issues with canterbury walls.

TheBimmer231:
can I get a better view of the new sprites, icon diffs doodoo

psykzz:
Same old icons used, just new features to them

GravitronBlue:
👎 Blaming this PR for the broken lighting caused by resin walls, presumably because resin walls built by hivelords are breaking lighting on the turf they were built.

Side issue too where when a wall thickens it colors the turf purple, this color stays even after the wall is destroyed, resulting in random purple/pink platings in caves.

PR #3891, opened 2020-04-19
Sulaco changes 

Timmeye:
Updates the Sulaco so that it can actually be played.  ****THERE WAS A PR FOR THIS BUT I DELETED IT IN AN ATTEMPT TO FIX ERRORS****
🆑
add: The Sulaco has had massive changes to make it actually playable. NOTE: the sulaco is ment to be a high pop map.
/🆑


Timmeye:
I think somethings wrong with my version of python, I can't mapmerge.

Timmeye:
Thanks to  Hughgent I managed to get this actually working

Timmeye:
You can probably test merge this

KwinOriginal:
What's that empty bulge at the lower deck? Is it supposed to be empty or should something get placed there in the future?

Timmeye:
TRAVIS IS GAY

KwinOriginal:
Excellent, I'm happy to see it in action!

psykzz:
Getting a lot of reports that the vendors have not beeen updated with the current content.
Both ingame and discord
https://discordapp.com/channels/498569646014857217/500801682058379265/703358012647276685

Jergobob:
Some issues. First you can have more then one of the same kit (just had a round with two sniper kits)
Second based of how tgmc calculates spec slots (which is not tied to the map tho I may be wrong) any spec slot after the 4th one can't get a spec kit. Because Sulaco uses spec tokens which only one spawns per vendor.

Timmeye:
the problem with the spec venders was there was 2 types of them. The old ones (the broken ones) are under the parent of vender while the normal versions of the /vending/ are under the parent of /marine_selector/

Pariah919:
DO NOT MERGE THIS THERE'S NO DRONE CONSOLE IN THE HANGAR

Timmeye:
ooh yeah the drone console

PR #3894, opened 2020-04-20
Pillar stuff round something

grinkgo:
## About The Pull Request

big surprise. This time I changed up req, added an autolathe and some other things. Replaced engineering's reactor cells to full ones because it's annoying to have empty reactors and empty fuel cells at roundstart. UH some holopads and cameras too.


KwinOriginal:
I shall wait until Electric has his PR done and then it has to be compared. Jeez, guys!

PR #3895, opened 2020-04-20
[s] updates pillow requirements

Rohesie:
There's vulnerabilities fixed in the newer version.


psykzz:
Has version 7.x.x been tested anywhere?
we only need Pillow>= 6.2.2 to fix to vuln.

Just worried upgrading a major version which could have a breaking change.

Rohesie:
/tg/ uses this version

PR #3904, opened 2020-04-22
tgui v3

spookydonut:
## About The Pull Request
https://github.com/tgstation/tgstation/pull/50497

## Changelog
:cl:
code: tgui updated
/:cl:


tgui/packages/tgui/interfaces/APC.js
```diff
@@ -1,82 +1,50 @@
...
+import { useBackend } from '../backend';
+import { Box, Button, LabeledList, NoticeBox, ProgressBar, Section } from '../components';
+import { Window } from '../layouts';
 import { InterfaceLockNoticeBox } from './common/InterfaceLockNoticeBox';
 
-export const Apc = props => {
-  const { state } = props;
-  const { config, data } = state;
-  const { ref } = config;
+export const Apc = (props, context) => {
```
psykzz, line 7:
```suggestion
export const APC = (props, context) => {
```

spookydonut: I need to rename the file actually

code/modules/power/apc.dm
```diff
@@ -666,7 +666,7 @@
 	ui = SStgui.try_update_ui(user, src, ui_key, ui, force_open)
 
 	if(!ui)
-		ui = new(user, src, ui_key, "apc", name, ui_x, ui_y, master_ui, state)
+		ui = new(user, src, ui_key, "APC", name, ui_x, ui_y, master_ui, state)
```
spookydonut, line 669:
```suggestion
		ui = new(user, src, ui_key, "Apc", name, ui_x, ui_y, master_ui, state)
```

tgui/packages/tgui/interfaces/APC.js
psykzz: either or would've wokred

David-Stormwell:
Stuff like medic vendors and other vendors which use a point system most likely don't work, as in you cannot buy anything because it's simply disappeared.

spookydonut:
I tested all uis and they worked for me can you be more specific

psykzz:
This seems to just be something with the sulaco, and their vendors there.
https://github.com/tgstation/TerraGov-Marine-Corps/pull/3891#issuecomment-619252653

psykzz:
unless you have examples on other maps @David-Stormwell

PR #3908, opened 2020-04-23
Xeno Aging Tweak

InterroLouis:
## About The Pull Request

This basically just slows down Xeno aging progression a bit. 

## Why It's Good For The Game

I suppose we'll just have to see, it's a balance thing and we can only really figure that out...with testing. Hopefully it will make xenos camping FOB less prevalent as they'll have aged a bit less. 

## Changelog
:cl:
balance: Xeno Upgrade progression slowed a bit
/:cl:


spookydonut:
i beg you to use a proper git client

spookydonut:
You could probably just do this with a config value to scale all upgrade thresholds by a set factor and then be able to nag admins to change the value instead of having to merge/update prs to tweak it

terranaut1:
making this a config option is probably not a bad idea

PR #3909, opened 2020-04-23
Revised cloning

psykzz:
## About The Pull Request
This adds a new machine, linked console and reagent.
The machine take the new reagent (biomass) as a resource and will try to create a human every 15mins.

The human will start as an early vat-grown but overtime will natrually change into a full vat-grown species

Early vatgrown take 30% more dmg from all sources, standard vatgrown take 5% more dmg.

## Why It's Good For The Game

Gives marines additional benefits for having staffed non-combat roles.
This is something that a doctor would do, keep it stocked and help geaar up the new spawns.

## Changelog
:cl:
add: marine cloning, gives marines a way to respawn every 15mins that isn't tied to the typical respawn mechanic.
/:cl:

-------------

> The following was apart of the original idea, but was later scraped.
> Keeping it here just for context

# Revised Cloning

120 Total Chemical per full regen
sub par regen requires more reagents

### Core Reagants (required never change)
iron - 10
water - 10
sugar - 10

### Biomass (For req points)
Initial free within the research lab
50 = very cheap

### Secret chems based on blood type
A random 3 part chemical per blood types
A  - Clone XYZ x40
B  - Clone XYZ x40
O  - Clone XYZ x40
AB - Clone XYZ x40

### Todo
 - [x] Create biomass chemical
 - [x] Create new "cloner/regenerator"
 - [x] Map out the botanics and replace with cloner room
 - [ ] Add interaction of bloodtype with clonegoo 
 - [ ] Missing secret chems causes disabilities

#### Disabilities

 - Missing limbs
 - Blind
 - Ear damage
 - high or very high metabolism

#### Notes

 - Randomly spawn ghosts if the mind is ghost.
 

#### Xeno consideration
 - Xeno bodies can be "processed" to double the biomass chemical
 - Drag xeno body to hive to reduce their respawn timer.


Hathkar:
Yeah, some sort of way for xenos to be able to revive or reuse their dead would be a nice addition as well.

jroinc:
Um. Seems like an iffy concept, but it'd be nice for researchers to do stuff.


I'd rather not see marines become even more unkillable, as they can be revived within 5 minutes, and can respawn 25 minutes after that. At a minimum for this to fly and not be hilariously one-sided, aliens recovering dead alien bodies would need to grant larva.

How do minds interact? What skills are given?

What happens if the RSR wises up and starts dragging roundstart dead bodies back to revive for ghosts?

What happens when someone chestbursts, says "fuckit", dies as larva, then gets re-cloned?

spookydonut:
lol cloning

TiviPlus:
lol cloning
image

spookydonut:
Don't we already have a respawn system?

Hughgent:
Lazarus machine.

Classic name for a reviving machine.

psykzz:
> Don't we already have a respawn system? 

We do but its just a timer, which isn't a fun mechanic. The idea here is to allow researchers/doctors to put effort into shortening the respawn timer for marines.

psykzz:
> I'd rather not see marines become even more unkillable, as they can be revived within 5 minutes, and can respawn 25 minutes after that. At a minimum for this to fly and not be hilariously one-sided, aliens recovering dead alien bodies would need to grant larva.

25 mins is still a long time, but of course can still be balanced

> How do minds interact? What skills are given?

If the original body / ghost, this is just a revival. If the ghost isn't available we call out to all ghosts for a player.
If a new ghost appears, while successful the process would say something like "the mind just isn't quite right" and the body would get a new name.


> What happens if the RSR wises up and starts dragging roundstart dead bodies back to revive for ghosts?

Roundstart dead bodies can be except from the system


> What happens when someone chestbursts, says "fuckit", dies as larva, then gets re-cloned?

They would be under the "new mind" group and would be in a queue, if they happened to be next they would be respawned.

As with all of these, there would be a global respawn timer (currently like 5 mins) before you can be eligible to play again.

spookydonut:
I think if you made it 'cloning' in that you can make completely new marines with time and resource investment instead of just reviving the dead that might solve some of the issues brought up.

psykzz:
I'd be happy with that, I wanted to focus on was having researchers improve efficiency via trial and error and having it cost limited via req points.

The idea of recovering bodies was mostly just as a catalyst and incentive for more travel back and forth with the shuttles.

spookydonut:
> 
> 
> I'd be happy with that, I wanted to focus on was having researchers improve efficiency via trial and error and having it cost limited via req points.
> 
> The idea of recovering bodies was mostly just as a catalyst and incentive for more travel back and forth with the shuttles.

the dead could be used as biomass to speed up the process

psykzz:
That'd work well and allow medical to alleviate the cost on req.
Fresher bodies are worth more "biomass"

Hughgent:
Essentially only making new 'squad marines', vat growns, baldies. In fact, force the bald hair style. (they can find a mirror in any of the bathrooms if they truly care)

jroinc:
Still doubtful about this, unless it's a full replacement for marine respawns. 5 min respawns (I can't really see cloning being much longer than that) are short enough that they make running out in the FoB with 2 shotties and riot armor alone attractive.

Sure, you'll die, but you'll just come back, and you lose nothing but some easily-replaceable gear.

Additionally, survivors would just be free latejoin marines that might have killed a few larva.

There's also the "ghost scout" effect, but that's common to any system that allows respawns and can be ignored, I guess.

The biggest issue I see, and mentioned above, is the "Gets nested, chestbursts, rushes unga lines to die as larva/young runner, jumps in that vastly shorter respawn line". With 30 min respawns, it didn't make sense. With 5 minute (Or maybe less, maybe 2 minute~) respawn timers, tactical suicide so you can unga again appears more attractive.

psykzz:
Im down to hash this out further via a doc - https://docs.google.com/document/d/1wa6-mxmeIbUN_0POmmQMSrEhJL8GQNXlrgH_oAqhb1c/edit#

Just a note, this would only take effect after the defib time, there is no reason to "clone" someone that is dead and can be revived, so even 2 mins would turn into 7.

Also all ghosts should have a 5min respawn timer (if they can't reenter their original bodies)

jroinc:
> Im down to hash this out further via a doc - https://docs.google.com/document/d/1wa6-mxmeIbUN_0POmmQMSrEhJL8GQNXlrgH_oAqhb1c/edit#
> Just a note, this would only take effect after the defib time, there is no reason to "clone" someone that is dead and can be revived, so even 2 mins would turn into 7.
> Also all ghosts should have a 5min respawn timer (if they can't reenter their original bodies)

I started commenting on the document, but it was obnoxious as hell to type on, so no.

Cloning should ideally require a body and respawn only the marine who's ghost it is. It should NOT work on chestbursted people for balance reasons (I'm gonna set ayy preference to off and wait for my revive, who cares if there are 5 burrowed). Cloning should have all/some of these disabilities even on a "perfect" clone-

Blind, genetic damage, organ damage, brain damage. These are chosen because they can all be fixed chemically and require little MD/RSR interaction besides hitting the clone with an autoinjector and throwing him under a wrenched shower. 

The assumption is that the cloner timer takes the place of the time it would take a MD/RSR to fix his broken bones/IB. Another assumption is that cloning should be preferable to vatgrown respawns.
Thought-Cloning only works if all broken bones are fixed?

Vatgrown should take ~10-15 minutes to make and require resources as stated. They should monopolize the cloner while being produced and should be storable up to one. They should come out with no damage, but be physically weaker than marines (significantly lower bone-break threshold). This is to disincentivize suicide plays and insta-respawn as a free unga, while still allowing it as an option with tradeoffs.

For both of these, every 3 clones/vatgrown instances should add 1 burrowed larva, just like respawns. If you need either of these, the unga was already dead, fair and square.

psykzz:
I'll figure out later why the line endings on theseus changed. I've only mapmerged.

code/__HELPERS/global_lists.dm
```diff
@@ -137,16 +137,25 @@ GLOBAL_LIST_EMPTY(randomized_pill_icons)
...
 		var/datum/reagent/D = new path()
 		GLOB.chemical_reagents_list[path] = D
+		GLOB.chemical_required_reagents[path] = list()
 
 	for(var/path in subtypesof(/datum/chemical_reaction))
 
 		var/datum/chemical_reaction/D = new path()
 		var/list/reaction_ids = list()
 
 		if(D.required_reagents && D.required_reagents.len)
```
Rohesie, line 147:
```suggestion
		if(length(D.required_reagents))
```

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,237 @@
...
+	opacity = FALSE
+	anchored = TRUE
+	max_integrity = 200
+
+	resistance_flags = UNACIDABLE | INDESTRUCTIBLE // For now, we should work out how we want xenos to counter this
+
+
+/obj/machinery/cloning/relaymove(mob/user)
+	user.visible_message("You hear something bang on the window of \the [src]", "The door won't budge")
+	return FALSE
```
Rohesie, line 19: Does this have a cooldown?

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,237 @@
...
+
+/obj/machinery/cloning/vats/Destroy()
+	deltimer(timerid)
+	timerid = null
+
+	// Force tthe clone out, if they have a client
+	if(occupant?.client)
+		eject_user()
+
+	QDEL_NULL(occupant)
```
Rohesie, line 105: Do you want to qdel the ejected occupant?

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,237 @@
...
+
+/obj/machinery/cloning/vats/attackby(obj/item/I, mob/living/user, params)
+	. = ..()
+	if(.)
+		return
+
+	if(!ispath(I, /obj/item/reagent_containers/glass/beaker))
+		if(beaker)
+			to_chat(user, "<span class='warning'>A beaker is already loaded into the machine.</span>")
+			return
```
Rohesie, line 148: Did you mean `if(ispath)` here?

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,237 @@
...
+
+		update_icon()
+
+		return TRUE
+
+
+/obj/machinery/cloning/vats/examine(mob/user)
+	. = ..()
+	if(!beaker)
+		to_chat(user, "<span class='notice'>It doesn't have a beaker attached</span>")
```
Rohesie, line 168:
```suggestion
		to_chat(user, "<span class='notice'>It doesn't have a beaker attached.</span>")
```

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,237 @@
...
+		return
+	if(timerid)
+		to_chat(user, "<span class='notice'>There is something weird inside</span>")
```
Rohesie, line 171:
```suggestion
		to_chat(user, "<span class='notice'>There is something weird inside.</span>")
```

code/modules/mob/living/blood.dm
```diff
@@ -188,6 +188,22 @@
...
 
+/mob/living/carbon/monkey/inject_blood(obj/item/reagent_containers/container, amount)
+	. = ..()
+
+	// A way to assign a blood type to a monkey for clone research
+	if(blood_type)
+		return
+
+	for(var/datum/reagent/R in container.reagents.reagent_list)
+		// If its blood, lets check its compatible or not and cause some toxins.
```
Rohesie, line 199:
```suggestion
	for(var/r in container.reagents.reagent_list)
		var/datum/reagent/R = r
		// If its blood, lets check its compatible or not and cause some toxins.
```

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -298,7 +298,38 @@
 	heat_level_2 = 480
 	heat_level_3 = 1100
 
+/datum/species/human/clone
+	name = "Human clone"
+	name_plural = "Human clones"
```
Rohesie, line 303: How about Bay vatborns/vat-grown instead, perhaps?

code/game/objects/machinery/cloning/cloning.dm
psykzz: I did but in testing ispath wasn't working at all for me or hugh

Rohesie: Oh, this should be:
```suggestion
	if(istype(I, /obj/item/reagent_containers/glass/beaker))
		if(beaker)
			to_chat(user, "<span class='warning'>A beaker is already loaded into the machine.</span>")
			return
```
`ispath` is to check if a variable is a type path variable

psykzz: Oh sorry no i was looking at the wrong lines and read what you said wrong,
You're right i need istype.

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,237 @@
...
+		return
+
+	if(!ispath(I, /obj/item/reagent_containers/glass/beaker))
+		if(beaker)
+			to_chat(user, "<span class='warning'>A beaker is already loaded into the machine.</span>")
+			return
+
+		// Check if the beaker contains anything other than biomass juice
+		for(var/datum/reagent/R in I.reagents.reagent_list)
+			if(!istype(R, /datum/reagent/medicine/biomass) && !istype(R, /datum/reagent/medicine/biomass/xeno))
```
psykzz, line 152: @Rohesie this is where i wanted to use `ispath` but failed, because from docs i thought it would handle children too

code/game/objects/machinery/cloning/cloning.dm
psykzz: No added one

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,238 @@
...
+	if(.)
+		return
+
+	if(istype(I, /obj/item/reagent_containers/glass/beaker))
+		if(beaker)
+			to_chat(user, "<span class='warning'>A beaker is already loaded into the machine.</span>")
+			return
+
+		// Check if the beaker contains anything other than biomass juice
+		for(var/datum/reagent/R in I.reagents.reagent_list)
```
Rohesie, line 153:
```suggestion
		for(var/r in I.reagents.reagent_list)
			var/datum/reagent/R = r
```

code/game/objects/machinery/cloning/cloning.dm
Rohesie:
```cpp
var/datum/reagent/p = /datum/reagent/medicine/biomass
to_chat(world, "[ispath(p)]") //This shows 1
to_chat(world, "[istype(p)]") //This shows 0, it's not an instanced entity, it's just a path, similar to a string

p = new p() //Now we instance it and create the actual datum.
to_chat(world, "[ispath(p)]") //This shows 0, it's no longer a path, it's an instanced entity now.
to_chat(world, "[istype(p)]") //It has been instanced now with the correct type, so this outputs 1.
```

psykzz: Specifically i was looking to do the subpath matching
`ispath(/my/path/child, /my/path) // TRUE`

Ah i was confused and thought it was a list of paths (even though im using the instances)
I've fixed that by doing the ispath on the actual path, should fix my usage.

code/datums/jobs/job/marines.dm
```diff
@@ -210,3 +210,21 @@ You are also in charge of communicating with command and letting them know about
...
+	display_order = JOB_DISPLAY_ORDER_SQUAD_MARINE
+	outfit = /datum/outfit/job/marine/vatgrown
+	total_positions = 0
+	job_flags = JOB_FLAG_ADDTOMANIFEST|JOB_FLAG_PROVIDES_SQUAD_HUD
+	jobworth = list(/datum/job/xenomorph = LARVA_POINTS_REGULAR)
+
+/datum/outfit/job/marine/vatgrown
+	name = SQUAD_VATGROWN
+	jobtype = /datum/job/terragov/squad/vatgrown
+	id = /obj/item/card/id/dogtag
```
Rohesie, line 230: Redefine the `return_spawn_type` proc for this job.

code/__DEFINES/skills.dm
```diff
@@ -91,10 +91,6 @@
 #define SKILL_SURGERY_MASTER		5 //to be implemented, perhaps instant surgery
 //higher levels means faster surgery.
```
Rohesie, line 93:
```suggestion




```
So the file remains unchanged.

Rohesie:
Why is the Theseus map diff so big?

psykzz:
the CL/CRLF changes

Rohesie:
The map needs to be fixed back to LF, else a ton of things will conflict.

psykzz:
It got fixed https://github.com/tgstation/TerraGov-Marine-Corps/pull/3939, this is another issue now

code/__DEFINES/skills.dm
```diff
@@ -94,7 +94,6 @@
```
Rohesie, line 96:
```suggestion

```
To leave the file untouched by diff.

code/__HELPERS/global_lists.dm
```diff
@@ -137,16 +137,25 @@ GLOBAL_LIST_EMPTY(randomized_pill_icons)
 	for(var/path in subtypesof(/datum/reagent))
 		var/datum/reagent/D = new path()
 		GLOB.chemical_reagents_list[path] = D
+		GLOB.chemical_required_reagents[path] = list()
```
Rohesie, line 140: Is this needed?

code/game/objects/machinery/cloning/cloning.dm
Rohesie: Could you use more descriptive variable names than `I` and `R`?
Single-letter upper-case is widespread in use, but not a really good standard.

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,245 @@
...
+
+/obj/machinery/cloning/vats/Initialize()
+	. = ..()
+	beaker = new /obj/item/reagent_containers/glass/beaker/biomass
+	update_icon()
+
+
+/obj/machinery/cloning/vats/Destroy()
+	deltimer(timerid)
+	timerid = null
```
Rohesie, line 106:
```suggestion
	if(timerid)
		deltimer(timerid)
		timerid = null
```

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,245 @@
...
+	occupant.fully_replace_character_name(occupant.real_name, occupant.species.random_name(occupant.gender))
+	occupant.disabilities |= (BLIND & DEAF)
+	occupant.set_blindness(10) // Temp fix until blindness is fixed.
+	// Blindness doenst't trigger with just the disability, you need to set_blindness
+
+	GLOB.offered_mob_list += occupant
+	notify_ghosts("<span class='boldnotice'>A new clone is available! Name: [name]</span>", enter_link = "claim=[REF(occupant)]", source = src, action = NOTIFY_ORBIT)
+
+	// Cleanup the timers
+	deltimer(timerid)
```
Rohesie, line 244:
```suggestion
```
This is triggered either when no timer exists or when the timer is finished. No need to delete it, just null the reference.

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,245 @@
...
+
+/obj/machinery/cloning/vats/Destroy()
+	deltimer(timerid)
+	timerid = null
+
+	// Force tthe clone out, if they have a client
+	if(occupant?.client)
+		eject_user()
+
+	occupant = null
```
Rohesie, line 112:
```suggestion
	// Force tthe clone out, if they have a client
	if(occupant)
		if(occupant.client)
			eject_user()
		else
			qdel(occupant)
		occupant = null

```

code/game/objects/machinery/cloning/cloning.dm
psykzz: It was mostly incase admemes

code/__HELPERS/global_lists.dm
psykzz: This was originally to improve how easier it was to look up cood and its required reagents, we originally planned for some harder mechanics for cloning around this, but never got around to actually implementing it.

I left it in because it was useful to Hugh at the time.

code/game/objects/machinery/cloning/cloning.dm
psykzz: I'd rather this always kicked out the occupant so that when xenos eventually get to destroy them they get the corpse too

code/__HELPERS/global_lists.dm
Rohesie: But it's currently just creating empty lists?

code/__DEFINES/skills.dm
Rohesie:
```suggestion


```
Same idea as before.

code/__HELPERS/global_lists.dm
psykzz: Originally my thought was, this would set the default for every reagent in the case that it doesn't have a reaction the list would be empty.

However, if they just don't appear we can also assert they don't have reactions

Rohesie:
Let me know when this is ready for another review.

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,244 @@
...
+	if(istype(hit_by, /obj/item/reagent_containers/glass/beaker))
+		if(beaker)
+			to_chat(user, "<span class='warning'>A beaker is already loaded into the machine.</span>")
+			return
+
+		// Check if the beaker contains anything other than biomass juice
+		for(var/instance in hit_by.reagents.reagent_list)
+			var/datum/reagent/regent = instance
+			if(!istype(regent, /datum/reagent/medicine/biomass) && !istype(regent, /datum/reagent/medicine/biomass/xeno))
+				to_chat(user, "<span class='warning'>\The [src] rejects the beaker.</span>")
```
TiviPlus, line 160:
```suggestion
				to_chat(user, "<span class='warning'>\The [src] rejects the beaker due to incompatible contents.</span>")
```

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,244 @@
...
+	occupant.set_blindness(10, TRUE)
+	to_chat(occupant, {"
+<span class='notice'>You are a frestly spawned clone, you appear as a Squad marine, but nothing more.
+You remember nothing of your past life.
+
+You are weak, best rest up and get your strength before fighting.</span>"})
+	occupant.vomit()
+
+	occupant = null
+	update_icon()
```
TiviPlus, line 227: Move this proc to below finish_growing_human() for clarity

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,244 @@
...
+	bound_width = 32
+	bound_height = 64
+	density = TRUE
+	opacity = FALSE
+	anchored = TRUE
+	max_integrity = 200
+
+/obj/machinery/cloning/Initialize(mapload, ...)
+	. = ..()
+	START_PROCESSING(SSmachines, src) // Registered for power usage
```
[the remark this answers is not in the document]
psykzz: Its not actually, this all machinery does
```
/obj/machinery/Initialize()
	. = ..()
	GLOB.machines += src
	component_parts = list()
```

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,244 @@
+/*
+Cloning shit
+
+*/
+/obj/machinery/cloning
```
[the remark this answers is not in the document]
psykzz: This allows you to use the machine as fluff if you want without making it actually work, though i could remove that.

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,244 @@
...
+	list_reagents = list(/datum/reagent/medicine/biomass = 60)
+
+
+/**
+These automatically generate marine bodies baded ona  timer.
+These hold the body until taken by a ghost where they "burst" from the vat.
+
+The vat then needs to be repaired and refilled with biomass.
+*/
+/obj/machinery/cloning_console/vats
```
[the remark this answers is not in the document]
psykzz: Yeah I wanted to keep it simple, though i could add a UI later so people can inspect the timer or so.

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -300,6 +300,45 @@
...
+	brute_mod = 1.3
+	burn_mod = 1.3
+	slowdown = 1.3
+
+	var/timerid
+
+/datum/species/human/vatgrown/early/handle_post_spawn(mob/living/carbon/human/H)
+	. = ..()
+	H.skills = /datum/skills/vatgrown/early
+	timerid = addtimer(CALLBACK(src, .proc/handle_age, H), 15 MINUTES, TIMER_STOPPABLE)
```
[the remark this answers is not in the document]
psykzz: Timers are actually cleaned up alread in `/datum/proc/Destroy`.
Though i could add it here anyway.

code/__DEFINES/cooldowns.dm
```diff
@@ -15,6 +15,7 @@
 #define COOLDOWN_NEST		"nest"
 #define COOLDOWN_BUMP_ATTACK "bump_attack"
 #define COOLDOWN_TASTE		"taste"
+#define COOLDOWN_RELAYMOVE	"relaymove"
 #define COOLDOWN_VENTSOUND	"vendsound"
```
TiviPlus, line 19:
```suggestion
#define COOLDOWN_TASTE		"taste"
#define COOLDOWN_VENTSOUND	"vendsound"
```

TiviPlus (approved):

code/game/objects/machinery/cloning/cloning.dm
```diff
@@ -0,0 +1,246 @@
...
+/obj/item/reagent_containers/glass/beaker/biomass
+	list_reagents = list(/datum/reagent/medicine/biomass = 60)
+
+
+/**
+These automatically generate marine bodies baded on a timer.
+These hold the body until taken by a ghost where they "burst" from the vat.
+
+The vat then needs to be repaired and refilled with biomass.
+*/
```
TiviPlus, line 40:
```suggestion
/**
  *These automatically generate marine bodies based on a timer.
  *These hold the body until taken by a ghost where they "burst" from the vat.
  *
  *The vat then needs to be repaired and refilled with biomass.
  */
```

code/modules/reagents/reactions/medical.dm
```diff
@@ -163,3 +163,9 @@
 	results = list(/datum/reagent/medicine/neuraline = 4, /datum/reagent/toxin/zombiepowder = 1)
 	required_reagents = list(/datum/reagent/medicine/synaptizine = 1, /datum/reagent/medicine/dylovene = 1, /datum/reagent/medicine/tricordrazine = 2, /datum/reagent/consumable/larvajellyprepared = 1)
 	required_catalysts = list(/datum/reagent/consumable/larvajelly = 5)
+
+// Cloning chemicals
+/datum/chemical_reaction/expanded_biomass
+	name = "Biomass"
+	results = list(/datum/reagent/medicine/biomass/xeno = 10)
+	required_reagents = list(/datum/reagent/blood/xeno_blood = 10, /datum/reagent/medicine/biomass = 1)
```
TiviPlus, line 171:
```suggestion
	required_reagents = list(/datum/reagent/blood/xeno_blood = 10, /datum/reagent/medicine/biomass = 1)

```

TiviPlus (approved):

PR #3910, opened 2020-04-23
Binocular buffs

XElectricX:
## About The Pull Request

Binoculars will now be able to target areas with a metal roof as well. Orbital cannons, rail guns, close air support, and obtaining coordinates are no longer blocked by a dinky metal roof. This does not apply to areas under ground. So places like underground bunkers and caves are unaffected by this.

An extra buff was given to mortars this way as well. Range finders no longer give themselves away, giving xenos a huge warning in advance compared to other ordnance. Coordinates are also given instantly.

## Why It's Good For The Game

Fire support is not restricted to 25% of the map (depending on the map) anymore. Not that it would have much use in an open field anyways. Orbital cannons no longer have to live in a corner in fear of a R O O F. Hospital festering with xenos? Call for an OB. Xenos have your FOB under siege but Nexus is conveniently keeping them safe? Not if the mortar can say anything about it. See a building that looked at you funny? Call in air support to reduce it to rubble. Pilots will have an actual job now. You also don't have to keep the webmap open just for coordinates, range finders will be superior since you will be able to see your targets as well.

## Changelog
:cl:
balance: Binoculars can target buildings with metal roofs. Cannot target anything underground or in caves.
balance: Range finders gather coordinates instantly and do not create a green dot for the xenos to laugh at.
/:cl:


David-Stormwell:
massive but needed marine buff.
fuck boilers

XElectricX:
Thank you for your approval stinky.

TiviPlus:
+1, although I think that maybe it should take longer to laze instead to maybe allow more of a transition from glass to caves

code/game/objects/items/binoculars.dm
```diff
@@ -178,15 +178,18 @@
...
-		to_chat(user, "<span class='warning'>INVALID TARGET: target must be visible from high altitude.</span>")
+		to_chat(user, "<span class='warning'>DEPTH WARNING: Target too deep for ordnance.</span>")
 		return
 	if(user.action_busy)
 		return
 	playsound(src, 'sound/effects/nightvision.ogg', 35)
-	to_chat(user, "<span class='notice'>INITIATING LASER TARGETING. Stand still.</span>")
-	if(!do_after(user, max(1.5 SECONDS, target_acquisition_delay - (2.5 SECONDS * user.skills.getRating("leadership"))), TRUE, TU, BUSY_ICON_GENERIC) || world.time < laser_cooldown || laser)
-		return
+	if(!MODE_RANGE_FINDER)
```
spookydonut, line 189:
```suggestion
	if(mode != MODE_RANGE_FINDER)
```

code/game/objects/items/binoculars.dm
```diff
@@ -200,12 +203,8 @@
 		if(MODE_RANGE_FINDER)
 			var/obj/effect/overlay/temp/laser_coordinate/LT = new (TU, laz_name, S)
 			coord = LT
-			to_chat(user, "<span class='notice'>SIMPLIFIED COORDINATES OF TARGET. LONGITUDE [coord.x]. LATITUDE [coord.y].</span>")
+			to_chat(user, "<span class='notice'>COORDINATES: LONGITUDE [coord.x]. LATITUDE [coord.y].</span>")
 			playsound(src, 'sound/effects/binoctarget.ogg', 35)
-			while(coord)
-				if(!do_after(user, 5 SECONDS, TRUE, coord, BUSY_ICON_GENERIC))
-					QDEL_NULL(coord)
-					break
```
psykzz, line 208: why was this removed?

XElectricX: Served no function.

XElectricX:
I'm not inclined to allow the whole map to be a valid target. Underground areas serve as a safe space for xenos. They are usually far enough away from the main LZ of a map to not be a bother. Also still leaves fire support with a downside.

code/game/objects/items/binoculars.dm
psykzz: This served to clean up the laser once you were finished with it.

This along with removing the coord check https://github.com/tgstation/TerraGov-Marine-Corps/pull/3910/files#diff-0601e714d232ec89ad6d88796b9e0ccaL151 will just allow someone to spam to flood with lasers.

I'd rather just keep the check or replace it at least with a `QDEL_NULL` instead.

XElectricX: You are correct. I removed the function to prevent it from making new lasers every 5 seconds but I forget you need QDEL to delete things. Fixed now.

code/game/objects/items/binoculars.dm
```diff
@@ -200,12 +203,9 @@
 		if(MODE_RANGE_FINDER)
 			var/obj/effect/overlay/temp/laser_coordinate/LT = new (TU, laz_name, S)
 			coord = LT
-			to_chat(user, "<span class='notice'>SIMPLIFIED COORDINATES OF TARGET. LONGITUDE [coord.x]. LATITUDE [coord.y].</span>")
+			to_chat(user, "<span class='notice'>COORDINATES: LONGITUDE [coord.x]. LATITUDE [coord.y].</span>")
```
psykzz, line 206: Might as well just not even create the `obj/effect/overlay/temp/laser_coordinate`.
Just get the position of `TU` instead

henriquekill9576:
yes
good

PR #3911, opened 2020-04-24
Refactor xenomorph throw_impact()

spookydonut:
## About The Pull Request
This moves all the special casing into the actual abilities.
There's still some work to be done to remove `/datum/xeno_caste/var/charge_type` since that's now just `var/has_charge` but this pr is already big enough as it is.

## Why It's Good For The Game
Having abilities more self contained makes it easier to work on them.


Rohesie:
Sorry for taking a bit to answer this.
I started refactoring `trow_at` before this but kind of stop in the middle.
There's objective improvements here, but I have an issue with stuff like `COMSIG_XENO_OBJ_THROW_HIT`
Can't we get that signal at the base of the proc and stop redefining it for xenos?
Something like:
```
if(SEND_SIGNAL(src, COMSIG_OBJ_THROW_HIT, hit_atom, speed) & COMPONENT_THROW_HANDLED)
  return
```

spookydonut:
> 
> 
> Sorry for taking a bit to answer this.
> I started refactoring `trow_at` before this but kind of stop in the middle.
> There's objective improvements here, but I have an issue with stuff like `COMSIG_XENO_OBJ_THROW_HIT`
> Can't we get that signal at the base of the proc and stop redefining it for xenos?
> Something like:
> 
> ```
> if(SEND_SIGNAL(src, COMSIG_OBJ_THROW_HIT, hit_atom, speed) & COMPONENT_THROW_HANDLED)
>   return
> ```

That could be done in a future PR, the focus on this was an incremental improvement on abilities being self contained.

Rohesie:
I think this might make that refactor harder, but whatever.
We'll deal with it later.

spookydonut:
> 
> 
> I think this might make that refactor harder, but whatever.
> We'll deal with it later.

I'll be the one doing that refactor so its my cross to bear

Rohesie:
Code looks fine, @psykzz 
If no missed errors are found in testing then it should be ready for merge.

Issue #3728, opened 2020-04-03
Apostrophes cannot be entered into name


CCRWasHere:
Unable to type in an apostrophe into a character's name, thus making the creation of nicknames impossible.

PR #3915, opened 2020-04-25
Fix nicknames

psykzz:
Fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/3728

stripped_input encodes the name in html.
This then got stripped by reject_bad_name.

Opt'd to just let reject_bad_name handle everything


Rohesie:
This is how /tg/ does it.
`reject_bad_name()` is pretty old and could be improved, but it should be done upstream first.

PR #3916, opened 2020-04-25
Runechat port [DO NOT TEST MERGE]

psykzz:
Ports https://github.com/tgstation/tgstation/pull/50608

The PR isn't finished but I wanted to see how this would work with TGMC, with crowded shuttles and xenos, how that interaction will work for us.

### Notes
I reduced the font-size down a little, as i believed the 7px was too large, this does make text a little harder to read.

If this doesn't work out, we may potentially repurpose this to just action on keywords like *medic *hurt *cardinals etc.


Rohesie:
Is this still WIP? It was merged on /tg/ and it seems to be working fine both here and there.

psykzz:
I'd say so, not sure if we should add the pref or leave this as always on now.

Rohesie:
Definitely add the preference. It has a network cost, so people suffering from lag might want to disable it.

psykzz:
I was able to somehow keep this text on screen 
image

code/modules/client/preferences_savefile.dm
```diff
@@ -49,55 +49,12 @@
...
-
-	if(savefile_version < 38)
-		WRITE_FILE(S["menuoptions"], list())
-
 	if(savefile_version < 39)
 		WRITE_FILE(S["toggles_gameplay"], toggles_gameplay)
 
+	if(savefile_version < 41)
+		WRITE_FILE(S["toggles_gameplay"], toggles_gameplay)
+
```
Rohesie, line 57: These two actions are repeated. Why do we have them anyway, why are we increasing the savefile size and why removing the other ones?

psykzz: This was a bad copy-paste from me, it should be the new vars for runechat

I also removed the older settings since they have been there for over 9months and generally only for players that may be returning after that time and we want to progressively improve their prefs, however since then much has changed and they would be better off just having a fresh pref file.

psykzz:
Ready to merge

PR #3924, opened 2020-04-27
Does changes to the fire rate/damage of the BR.

Pariah919:
## About The Pull Request
Heavily changes the stats of the current battle rifle, (fire delay should be 0.25 with the spec gun skill)
Incin now does 20/25
High impact no longer stuns but it does do heavy slowdown stacks and stager, as well as having high pen/sundering.
Normal now does more sundering but less damage and AP
Smart does less AP.
I fixed some naming inconsistencies and the casing bug.

## Why It's Good For The Game
Makes the BR feel better than it's current incarnation.

## Changelog
:cl:
balance: M4RA has been rebalanced
fix: Should have fixed the casing bug for the M4RA, and some of the inconsistent naming
/:cl:


jroinc:
I'd swap the impact/regular sundering stats, impact doing slowdown seems like more of it's niche, while regular ammo doing high sundering gives it a use.

PR #3932, opened 2020-04-27
Fix vendors with points and buy_flags

psykzz:
Fixes vendors not showing a category if it used points instead of buyflags.


psykzz:
@spookydonut i know this will conflict with your changes, but this will at least fix the vendors today for medics and engineers

PR #3937, opened 2020-04-28
Legless trait and limb flags

Rohesie:
* Adds a legless trait so we can check it and react to its changes, instead of checking for number of legs each time.
* Embedding items no longer happens on lost limbs, which would cause it to runtime.
* Fixes lost limbs maintaining a bloody overlay.
* Cleans up limb code. That section is very old code and has a pending refactor.

Blood code also is in dire need of a cleanup and refactor. Lots of small issues.

The icon diff bot seems inactive? The sprite change is just separating the hand blood overlays in two, one for the right hand and one for the left hand, so when a limb is missing there's no ghost blood limb.


code/modules/mob/living/carbon/human/species.dm
```diff
@@ -133,12 +133,12 @@
 	H.limbs += LA
 	var/datum/limb/r_arm/RA = new(C, H)
 	H.limbs += RA
-	var/datum/limb/l_leg/LL = new(G, H)
+	var/datum/limb/leg/l_leg/LL = new(G, H)
```
psykzz, line 136: why was the subtype added?
I dont see it used anywhere

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -127,19 +128,21 @@ GLOBAL_LIST_EMPTY(damage_icon_parts)
 
 	previous_damage_appearance = damage_appearance
 
-	var/icon/standing = new /icon('icons/mob/dam_human.dmi', "00")
+	var/icon/standing = new('icons/mob/dam_human.dmi', "00")
 
-	var/image/standing_image = new /image("icon" = standing, "layer" =-DAMAGE_LAYER)
+	var/image/standing_image = image("icon" = standing, "layer" = DAMAGE_LAYER)
```
psykzz, line 133: Should the `-` have been removed here? Every other standing cache uses the negative values.

code/modules/organs/limbs.dm
```diff
@@ -590,6 +610,29 @@ Note that amputating the affected organ does in fact remove the infection from t
...
+/datum/limb/proc/set_limb_destroyed(new_status = TRUE)
+	if(limb_status & LIMB_DESTROYED)
+		if(new_status != FALSE)
+			return
+		limb_status &= ~LIMB_DESTROYED
+		return new_status
+	if(new_status != TRUE)
+		return
+	limb_status |= LIMB_DESTROYED
+	return new_status
```
psykzz, line 622: Could this not be simplier like
```suggestion
/datum/limb/proc/set_limb_destroyed(new_status = TRUE)
	. = new_status // return the new state of the limb
	if((limb_status & LIMB_DESTROYED) == new_status) // limb is already the state we want
		return
	limb_status ^= LIMB_DESTROYED // toggle the limb destroyed
	return
```

code/modules/mob/living/carbon/human/species.dm
psykzz: I see below nevermind

code/modules/organs/limbs.dm
Rohesie: That would work as well, but the idea is to return `null` if no status change took place, so that procs that call it and update on change can ignore such calls.
It's a pattern I've been using on several places for update-on-demand statuses.

Rohesie:
This should be ready. Maybe give it a round of test or two before, though.

code/modules/organs/limbs.dm
psykzz: Oh in which just remove the default return assignment, and `return new_status` on the last line

Rohesie:
I should change the logic on how embedding works to register on the limb being destroyed, which is now a signal.
The patch I did fixed a runtime when embedding happened on missing limbs, but there's still no handling for what happens when a limb is severed and has embedded stuff in it.
I need to handle it on the embedding code itself but will need to change the logic a bit.
I can do this in a future PR or in this one.

psykzz:
Rather it went into a another PR.
It would just delay merging of this one if its added in.

PR #3947, opened 2020-04-29
FC Buff

Max-023:
## About The Pull Request
Gives the FC NVGs, same as the SG.

## Why It's Good For The Game
The FC, right now, struggles in combat if they choose to stick with the SG. While SGners have no issue pushing with Marines, being able to see the xenos and judge if the situation is good enough for them to stay, for the FC they do not get that leisure. As it currently stands, there is no benefit to going FC over SL other than for the custom M1911. The SL gets their own vendors, the same big text all command roles have, a squad locator so Marines know where to find them, and their own special armor. The FC has the exact same kit as SGs, minus the NVGs. This severely affects their effectiveness and makes it even riskier for FCs to push with Marines, resulting in them being a redundant role to the CO, who sits at the FOB.

## Changelog
:cl:
add: FC has M56 goggles.
/:cl:


Toxici11i:
you have to be joking. The FC already is probably the best marine type right now, hands down. The FC is a synthetic that can use guns, skills included.

Max-023:
Everyone can do what he does and the SL has the better deal. The only things the FC has are no skill checks and a cool pistol. That's it.

TiviPlus:
You should not be a 5th SG as FC.
This conversation has happened plenty of times.
If you want to buff FC then do so in regards of their command ability, giving them NVGs is not the way to do this.

Max-023:
The SG is a massively underpowered weapon without the NVGs, as it is not good in CQC at all and it takes up so much inventory space before you can even fire it. It's either remove the SG and give them something else or complete the SG set so they don't have to constantly abandon their intended kit for something that would actually be useful.

Toxici11i:
> The SG is a massively underpowered weapon without the NVGs, as it is not good in CQC at all and it takes up so much inventory space before you can even fire it. It's either remove the SG and give them something else or complete the SG set so they don't have to constantly abandon their intended kit for something that would actually be useful.

you have to be joking lmao, even when ban evading on 200+ ping and 1fps i can still play FC better than most SG's can play SG.
it is NOT a hard role to play, and it is EXTREMELY unbalanced to buff them even further

Max-023:
Just because David Stormwell can wipe hives as a survivor doesn't mean survivors are OP. Likewise, just because one man can do good as an FC doesn't mean the FC is in a good state.

Toxici11i:
> Just because David Stormwell can wipe hives as a survivor doesn't mean survivors are OP. Likewise, just because one man can do good as an FC doesn't mean the FC is in a good state.

thats not what that means, its an anecdotal example of how easy FC is to play. You could be the most god awful, laggiest or simply most retarded player, and still kill 3-4 xenos even without NVG.

Toxici11i:
Consider this - Tac bino + SG FC (the usual)

If you smartly use your railgun you can push xenos out of cover, force them into your direct line of fire (and 3-4 other peoples) and wipe their hive.
If you play smart, you can act as a synthetic that can use guns - just less tanky.
etc.

David-Stormwell:
I still disagree with this and suggest the FC be expanded upon command wise, even if the FC got a SL vendor but for them would be a good step towards that

psykzz:
Since everyone seems to disagree with this change

PR #3931, opened 2020-04-27
Modular armor

psykzz:
## About The Pull Request

Adds modular armor

See https://docs.google.com/document/d/1xIOOGkCl_DhxMjdKUOp8c5bt84rdltJw05qgFdJNmPg/edit for design docs

This adds 
armor plates (light, medium heavy)
armor modules
armor storage

## Why It's Good For The Game
Armor variations

## Changelog
:cl:
add: Modular armor!
refactor: refactored some code around inventory and suit lights
/:cl:

PR #3948, opened 2020-04-29
cooldowns refactor

Rohesie:
Move the variable to the datum level so it can be used on #3931 and turns it into a lazylist.
The macros should also make the code a little bit more readable, hopefully.

Added some lazy list macros I added on /tg/ before and were found useful, so we have them as well, even if they are not being used currently.


code/__DEFINES/cooldowns.dm
```diff
@@ -0,0 +1,25 @@
...
+#define COOLDOWN_ZOOM		"zoom"
+#define COOLDOWN_BUMP		"bump"
+#define COOLDOWN_ENTANGLE	"entangle"
+#define COOLDOWN_NEST		"nest"
+#define COOLDOWN_TASTE		"taste"
+#define COOLDOWN_VENTSOUND	"vendsound"
+
+#define COOLDOWN_START(cd_source, cd_index, cd_time) LAZYSET(cd_source.cooldowns, cd_index, addtimer(CALLBACK(src, /atom/movable.proc/end_cooldown, cd_index), cd_time))
+
+#define COOLDOWN_CHECK(cd_source, cd_index) (cd_source.cooldowns && cd_source.cooldowns[cd_index])
```
psykzz, line 21:
```suggestion
#define COOLDOWN_CHECK(cd_source, cd_index) LAZYACCESS(cd_source, cd_index)
```

Rohesie: I had not done this initially because `LAZYACCESS` adds a few irrelevant checks, but thinking again that's sort of micro-optimizing and this is more clear and maintainable

PR #3952, opened 2020-04-30
Boiler acid smoke changes

LorenzoTheMarauder:
## About The Pull Request

Boiler acid glob fire delay increased, acid smoke lifetime decreased.

## Why It's Good For The Game

A boiler with nothing but the capacity to avoid being flanked and respond to CAS/OB/Mortar lasers can permanently suppress a narrow area almost alone. A lot of discussion was done and before this PR actually stopping a boiler when you can't push out to flank it is close to impossible and requires a very specific set of circumstances to actually pull off. Those being a competent sniper spec with knowledge of the laser targeter, missing no shots, manage to pop off four of them without the boiler actually moving, and also avoiding acid spit/thrown facehuggers/friendly fire etc. An acid cloud would last ten seconds before completely dissipating and by then you would have fifteen seconds maximum to kill or harm the boiler enough that it retreats before another acid cloud completely blocks your sight. 

## Changelog
:cl:

balance: Made it slightly more difficult for one boiler to completely stonewall marines close to permanently.

/:cl:


psykzz:
This is just a flat nerf to the boiler, decreased smoke time and increased cooldowns between shots.

LorenzoTheMarauder:
That was kind of the point. A single boiler with no backup can keep blasting an entire front so long as nobody with a scope shows up, a boiler with a defender battle buddy is safe from everything but the sniper spec, and if you add a drone or a hivemind you basically have 20-30% holding down the marines while the rest start going for the balls. Too many synergies with boiler smoke that require extremely specific counters or the absolute ineptitude of the boiler itself.

PR #3955, opened 2020-04-30
Fix applying armor / movespeed changes on pickup

psykzz:
Fixes equip to slot to only apply updates if the equipment is actually worn


code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -412,6 +412,9 @@
 		else
 			CRASH("[src] tried to equip [W] to [slot] in equip_to_slot().")
 
+	var/equipped_to_hands = slot == SLOT_R_HAND || slot == SLOT_L_HAND
+	if(equipped_to_hands)
+		return ITEM_EQUIP_CARRIED
 	if(W.flags_armor_protection)
 		add_limb_armor(W)
```
spookydonut, line 419: real fix is using /obj/item/equipped() instead of this.

psykzz: Moving these procs to that call instead?

spookydonut: yeah along with all the other code that triggers some action on equip/unequip

psykzz: I did try to move this but found it caused more issues, dropped doesn't have the slot so it becomes hard to know whether or not to remove the limb armor.

Since what im adding already existing in the inverse proc (doUnEquip) the equip_to_slot also doing the same feels correct here.

(Unless you have another suggestion)

spookydonut: so what you're saying is limb armor is poorly coded

PR #3962, opened 2020-05-01
Immobile trait

Rohesie:
Adds an immobility trait. This should make it easier to respond to events.
One step closer to removing the need for `update_canmove()`


spookydonut:
@MerryweatherLost what are you trying to do?

PR #3965, opened 2020-05-01
Xeno Upgrade Points Alterations

BobSlim:
## TO-DO/Related PRs

- [x] Adjust upgrade thresholds upwards. 
- [x] Alter modifiers to the upgrade uptick: resin silos.
- [x] Add scaling based on the queen's age.

## About The Pull Request

- Adds defines for upgrade_thresholds in mobs.dm, changes numbers in xeno castes to defines.
- Alters the formula for upgrade upticks
- Upgrade_stored is kept between evolution

## Why It's Good For The Game

Gives the game more of a rolling-onwards type feel. Time-pressure is on marines to push the aliens before they're all ancient, aliens are weak until they hit their power-spikes. But this time, it's on a hive-level rather than an alien-by-alien level.

Fixes the issues of aliens being ancient once the gates are lifted, and that spawning as an alien and going to T3 is a lot of waiting while you get to Mature.

## Changelog
:cl:
tweak: Xenos keep their age when they evolve/regress
tweak: Xenos have their ageing tick modified by the number of resin silos active. If you lose silos, you will start to age slower.
tweak: Xenos will age faster if their ruler is older; and slower if their ruler is younger.
balance: Ageing thresholds increased to match
/:cl:


Pariah919:
This is fine if you add it as just being this. But adding this AND making ancient even better seems kind of dumb, as it's currently linear progression. Also this is kind of bad for a TDM-based game mode like Distress and probably shouldn't happen until there's a dedicated objective marines can do.

BobSlim:
> as there's 20 minutes before marines can even land on the planet. So you'd have to take that into account for the game mode.
I'm thinking that upgrades are pretty spread out - like this when the gates open: T1: Elder T2: Mature T3: Young.

Could do a lot of things with the values in this system. But I do think it's healthy for TDM because the round becomes less likely to stalemate out. Marines have an advantage in the first bit of the game, aliens have an advantage in the second. 

Although, I reckon that marines will begin to stomp aliens a lot because they have the advantage. I would go for further defensive buffs to aliens in this case.

> But adding this AND making ancient even better seems kind of dumb, as it's currently linear progression for a reason
Would increasing the power of ancients be palatable if it's meme-level time, like 2 or 3 hours into the game? That's the type of scaling I'm thinking about.

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
```diff
@@ -757,3 +766,22 @@ to_chat will check for valid clients itself already so no need to double check f
 
 /mob/living/carbon/xenomorph/get_xeno_hivenumber()
 	return hivenumber
+
+// ***************************************
+// *********** Processing
+// ***************************************
+
+/datum/hive_status/process() //happens when the hive is processed, (currently) once per 2 secs.
```
spookydonut, line 774: nooooooooooooooooooooooooooo do not process hive datums

code/datums/gamemodes/crash.dm
```diff
@@ -103,8 +103,7 @@
 			X.amount_grown = X.max_grown
 		else // Handles Shrike etc
 			var/mob/living/carbon/xenomorph/X = i
-			X.upgrade_stored = X.xeno_caste.upgrade_threshold
-
+			X.hive?.upgrade_stored = X.xeno_caste.upgrade_threshold
```
psykzz, line 106: This might require some thought because a tier 2 should equal tier 1 + tier 2 requirements, etc for tier 3 since the hive evolution is shared like this now

spookydonut:
So you're flattening all upgrade thresholds?  take them out of caste datums then

BobSlim:
> So you're flattening all upgrade thresholds? take them out of caste datums then

I'm not too sure what you mean, but I think you may have read something wrong? Tier1/Tier2/Tier3/Tier(4) still have different thresholds for upgrading, and each caste still needs to have its upgrade tier defined. Unless you want me to do a switch to determine the caste's tier and then apply its upgrade thresholds?

PR #3967, opened 2020-05-02
Access rework

Rohesie:
* Removed `req_access_txt` and `req_one_access_txt`, we use lists and defines now.

* Refactored the way the lists are stored, so we make use of dictionaries for unique combinations. After the huge var-edit cleanup we'd still have 1014 unique access lists on Pillar + Prison. This system reduces this to just 2 directory lists, plus 25  unique `req_access` and 19 `req_one_access` lists. Not a huge gain, but it's something.

* Since I really don't love myself, removed a ton of var-edits, cleaning up the maps quite a bit.

* Ice Colony and to a lesser degree Prison Station had several access locks. Both were removed, following the standard of other patterns, to avoid having the station be powered be against the marines.

### DIFF
--- a/_maps/map_files/BigRed_v2/BigRed_v2.dmm
+++ b/_maps/map_files/BigRed_v2/BigRed_v2.dmm
@@ -261,10 +261,9 @@
 /turf/open/floor/tile/dark,
 /area/bigredv2/outside/space_port)
 "aaY" = (
-/obj/machinery/door/airlock/mainship/research/glass{
+/obj/machinery/door/airlock/mainship/research/glass/free_access{
 	dir = 4;
-	name = "\improper Eta Lab Secure Storage";
-	req_access_txt = "0"
+	name = "\improper Eta Lab Secure Storage"
 	},
 /turf/open/floor/asteroidfloor,
 /area/bigredv2/outside/nanotrasen_lab/inside)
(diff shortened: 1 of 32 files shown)


spookydonut:
They're for mapping to reduce number of compiled in types, i'm not sure what the trade off is for like memory usage, rsc size etc, might pay to check what the before/after on a bunch of things is with this.

Rohesie:
rsc (on disk) before the PR: `63.0 MB (66,088,960 bytes)`
rsc (on disk) after the PR: `63.0 MB (66,088,960 bytes)`

Memory by Dream Daemon on roundstart before the PR: `330.9 MB`
Memory by Dream Daemon on roundstart after the PR: `331.2 MB`

No significant difference that can be separated from random variance.

spookydonut:
> 
> 
> rsc (on disk) before the PR: `63.0 MB (66,088,960 bytes)`
> rsc (on disk) after the PR: `63.0 MB (66,088,960 bytes)`
> 
> Memory by Dream Daemon on roundstart before the PR: `330.9 MB`
> Memory by Dream Daemon on roundstart after the PR: `331.2 MB`
> 
> No significant difference that can be separated from random variance.
> 
> This fixes a ton of access issues that'd be more clear if we didn't use this var-edit standard, which is quite unmaintainable.

sweet!

Rohesie:
Any issues found?

psykzz:
No complaints so far

Rohesie:
I'd say it's ready for merge. I can fix eventual issues that open up, but I don't think there will be any, and if there are, it will be minor stuff.

PR #3970, opened 2020-05-03
Bump attack also targets foes on the ground when crossed

Rohesie:
Also refactors Crossed a bit, forcing it to call parent like on /tg/ and adding a signal for the thing doing the crossing.

## Changelog
:cl:
balance: Bump attack also targets foes on the ground when crossed.
/:cl:


Rohesie:
Had some issues with github but the conflicts are now solved.

PR #3974, opened 2020-05-03
soft/hard armor

Rohesie:
* Introduces the concept of hard and soft armor and generalizes it for all `living`. Sur wanted to add a soak variable before, and this can play the part. Soft armor is our regular armor, it reduces incoming damage by its value read as a %. Hard armor is what the soak was, it reduces damage by a flat value.

* Removed the 10% of the armor value being soak by default. This balance change can be reverted by adding actual hard armor to specific things as necessary. Personally I'd recommend against using it much, as it promotes spiky balance, making tanky castes tankier instead of the linear scaling soft armor offers. But in the future it might be useful for actual tanks / armored vehicles and whatnot. For now it only works, as before, for projectile damage.

* Since all `living` have armor(s) datums now, and humans already have armor datums on the limbs, added the concept of a human global armor that is made by the sum of all limb armor values scaled to the limb size by approximate body surface (represented by the % `cover_index` limb var, the sum of them reaching 100). This could be used environment damage and the likes.

## Changelog
:cl:
balance: Projectile damage soak zeroed. This means armors scale linearly against bullets, instead of high armors negating exponetially more damage against low values. In other words, high-armor nerf against bullets (and xeno spits).
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -1073,13 +1073,13 @@ Note that amputating the affected organ does in fact remove the infection from t
 
 /datum/limb/proc/add_limb_soft_armor(datum/armor/added_armor)
 	soft_armor = soft_armor.attachArmor(added_armor)
-	var/datum/armor/scaled_armor = added_armor.scaleAllRatings(cover_index * 0.01)
+	var/datum/armor/scaled_armor = added_armor.scaleAllRatings(cover_index * 0.01, 1)
 	owner.soft_armor = owner.soft_armor.attachArmor(scaled_armor)
 
 
 /datum/limb/proc/remove_limb_soft_armor(datum/armor/removed_armor)
 	soft_armor = soft_armor.detachArmor(removed_armor)
-	var/datum/armor/scaled_armor = removed_armor.scaleAllRatings(cover_index * 0.01)
+	var/datum/armor/scaled_armor = removed_armor.scaleAllRatings(cover_index * 0.01, 1)
 	owner.soft_armor = owner.soft_armor.detachArmor(scaled_armor)
 
 
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -14,7 +14,8 @@
 	var/last_dam = -1
 	var/supported = FALSE
 
-	var/datum/armor/armor
+	var/datum/armor/soft_armor
+	var/datum/armor/hard_armor
 
 	var/display_name
 	var/list/wounds = list()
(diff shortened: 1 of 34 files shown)


Pariah919:
nice

psykzz:
Seems the codex is still trying to get old armor values
```
[17:14:11] Runtime in mobs_codex.dm, line 33: Cannot execute null.getRating().
proc name: get mechanics info (/mob/living/carbon/xenomorph/get_mechanics_info)
```

psykzz:
Also `xenomorph/ex_act`
```
[17:19:29] Runtime in damage_procs.dm, line 16: Cannot execute null.getRating().
proc name: ex act (/mob/living/carbon/xenomorph/ex_act)
```

Rohesie:
Good catch, I had forgotten about one variable.
Those issues should be fixed now.

code/modules/organs/limbs.dm
```diff
@@ -55,6 +56,9 @@
 	var/vital //Lose a vital limb, die immediately.
 	var/germ_level = 0		// INTERNAL germs inside the organ, this is BAD if it's greater than INFECTION_LEVEL_ONE
 
+	///What % of the body does this limb cover. Make sure that the sum is always 100.
+	var/cover_index = 0
```
psykzz, line 60: Just to be clear if you didn't sum up to 100, it would just meant that the armor is less effective or more effective than intended?

For example if chest had 100 cover_index it would be 76% more effective, since its only meant to have a 27 coverage?

code/modules/projectiles/projectile.dm
```diff
@@ -743,35 +743,40 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
...
+			else if(soft_armor) //Soft armor/padding, damage reduction.
+				feedback_flags |= BULLET_FEEDBACK_SOAK
+				damage -= damage * soft_armor * 0.01
 
 	if(proj.ammo.flags_ammo_behavior & AMMO_INCENDIARY)
-		living_armor = get_living_armor("fire", proj.def_zone) //This value could potentially be negative, indicating fire weakness.
-		if(living_armor < 100) //If armor is 100 or over the mob is fireproof.
-			adjust_fire_stacks(CEILING(10 - (10 * (living_armor / 100)), 1)) //We could add an ammo fire strength in time, as a variable.
+		soft_armor = get_soft_armor("fire", proj.def_zone) //This value could potentially be negative, indicating fire weakness.
+		if(soft_armor < 100) //If armor is 100 or over the mob is fireproof.
```
psykzz, line 778: in your comment on https://github.com/tgstation/TerraGov-Marine-Corps/pull/3931#discussion_r419178802 you talked about using hard armor for fireproof, but here its soft armor.

Do we want to align that?

code/modules/organs/limbs.dm
Rohesie: If the total value is less than 100 then the sum of all armors, in the mob, will be lower than the median among them.

If the chest had a higher cover index it would contribute more towards the total armor in the mob.

Not sure if this answers your question?

Rohesie:
This should be ready for merge.

PR #3978, opened 2020-05-03
Add respawns into EORD

psykzz:
Add a verb to mobs created during EORD that allow them to respawn when they die.
This will respawn them again as a random xeno or another human.


code/modules/mob/mob_verbs.dm
```diff
@@ -83,6 +83,36 @@
...
+/// This is only available to mobs once they join EORD.
+/mob/proc/eord_respawn()
+	set name = "EORD Respawn"
+	set category = "OOC"
+
+	if(stat != DEAD)
+		to_chat(src, "You can only use this when your dead.")
+		return
+
+	var/list/spawn_types = pick(subtypesof(/mob/living/carbon/xenomorph) + list(/mob/living/carbon/human))
```
Rohesie, line 95: Check what `subtypesof(/mob/living/carbon/xenomorph)`, because I think the list won't be what you expect.

code/modules/mob/mob_verbs.dm
```diff
@@ -83,6 +83,36 @@
 	M.key = key
 
 
+/// This is only available to mobs once they join EORD.
+/mob/proc/eord_respawn()
+	set name = "EORD Respawn"
+	set category = "OOC"
+
+	if(stat != DEAD)
+		to_chat(src, "You can only use this when your dead.")
```
BobSlim, line 92:
```suggestion
		to_chat(src, "You can only use this when you're dead.")
```

code/modules/mob/mob_verbs.dm
psykzz: Yeah I thought about only including fun castes but took the easy way out.
I'll update this to be a list of valid combat xenos instead.

code/modules/mob/mob_verbs.dm
```diff
@@ -83,6 +83,44 @@
...
+	set name = "EORD Respawn"
+	set category = "OOC"
+
+	if(stat != DEAD)
+		to_chat(src, "You can only use this when you're dead.")
+		return
+
+	var/list/spawn_types = pick(list(
+		// Human added a few times to increase the chances without having to set ratios on every mob
+		/mob/living/carbon/human, /mob/living/carbon/human, /mob/living/carbon/human, /mob/living/carbon/human, /mob/living/carbon/human,
```
Rohesie, line 97:
```suggestion
		500 ; /mob/living/carbon/human,
```
For reference: `http://www.byond.com/docs/ref/index.html#/proc/pick`

psykzz:
No this covers everything i wanted out of it.

PR #3979, opened 2020-05-04
Fixes line endings issues forever

spookydonut:
## About The Pull Request
This adds a properly setup .gitattributes (unlike last time i tried this) and also I did `git add --renormalize .` (also unlike last time)

## Why It's Good For The Game
No more dumb line endings issues


psykzz:
Thiis is going to seem like its going to cause problems but there should be easy to fix.
Until we merge this we can just keep doing `git pull upstream master --rebase -X theirs`

spookydonut:
> 
> 
> Thiis is going to seem like its going to cause problems but there should be easy to fix.
> Until we merge this we can just keep doing `git pull upstream master --rebase -X theirs`

i'll rebase this when you want to merge it but won't update it until then

PR #3981, opened 2020-05-04
/tg/ explosions plus our own modifications

Rohesie:
A bit overkill, but am porting the /tg/ explosions framework.

It can be tested, main difference is that explosions will throw things away and shake screen on the epicenter, plus the admins gain the Maxcap bomb, but I'm not done with it yet.

Our explosions code is awfully outdated, and suffered from half-finished backend ports. Hopefully it will be a bit cleaner here.

* Fixed a runtime with autodoc deletion, in where it would call a verb on Destroy(), requiring a `usr`.

* Moved `resistance_flags` back to the `atom` level and merged the concept of `INDESTRUCTIBLE` with the one of `hull` turf. Should probably also merge `GODMODE` and others.

* Added admeme CAS bombs, plus a pretty handy debug verb to check bomb range.

* Minorly adjusted some explosive values. Shouldn't really change much on that department.

## Changelog
:cl:
add: Reworked explosions.
/:cl:

TODO:

* Build a better shock expansion algorithm. (Done)
* Turn `throw_at` into a processing subsystem. (Future PR)
* Integrate `visible_message` into a queue. (Future PR)


psykzz:
I actually tried this a little but found extools to be pretty funky, had to downgrade byond to certain versions to get it to work, otherwise i got IsInContainer exceptions being thrown.

I'd love for that to get more stable before we include it.

If we do end up adding this, i'd like to see a guide on update byond version to ensure that everything is done when we do.
(Update TGS, Dockerfile, + extools dll)

GravitronBlue:
👎 Fun for a single testmerge round, but the broken behavior of a single grenade causing full armor and clothing deletion plus 300 brute and burn damage needs to be debugged and patched. (Preferably before this gets TM'd again)

Rohesie:
> 👎 Fun for a single testmerge round, but the broken behavior of a single grenade causing full armor and clothing deletion plus 300 brute and burn damage needs to be debugged and patched. (Preferably before this gets TM'd again)

Both bugs fixed. The clothing deletion one pre-dated this PR, but should no longer haunt the ungas.

code/modules/shuttle/marine_dropship.dm
```diff
@@ -67,10 +67,10 @@
 	var/turf/left = locate(C.x - leftright, C.y, C.z)
 	var/turf/right = locate(C.x + leftright, C.y, C.z)
 
-	explosion(front, 1, 5, 9, 0)
-	explosion(rear, 3, 6, 10, 0)
-	explosion(left, 3, 6, 10, 0)
-	explosion(right, 3, 6, 10, 0)
```
psykzz: Do you need to also update the canterbury crash explosions?

psykzz: I would've thought the delayed nature of the explosions is what killed the xenos, rather than the size / spread

Rohesie: It was actually the size.
I need to update all explosions because we no longer signalize no-explosion with `-1`, we now do it with `0`, so the first values need to be increased in general.

Rohesie:
This should be ready for merge. The feature on the nuke and the other refactors I can do in a different PR, this has grown big enough as it is.

code/__DEFINES/equipment.dm
```diff
@@ -19,6 +19,8 @@
 #define OVERLAY_QUEUED			(1<<7)
 #define PREVENT_CLICK_UNDER		(1<<8)		//Prevent clicking things below it on the same turf
 #define CRITICAL_ATOM			(1<<9)		//Use when this shouldn't be obscured by large icons.
+/// should not get harmed if this gets caught by an explosion?
```
psykzz, line 22: This comment is a little confusing, how about
```suggestion
/// Does not cascade explosions to its contents.
```

code/game/objects/structures/dropship_ammo.dm
```diff
@@ -287,14 +282,12 @@
...
-			explosion(impact,1,2,3,6,1,0) //relatively weak
-			for(var/turf/T in range(4,impact))
-				T.ignite(60, 30) //cooking for a long time
-			qdel(src)
-
+/obj/structure/ship_ammo/rocket/napalm/detonate_on(turf/impact)
+	impact.ceiling_debris_check(3)
+	explosion(impact, 2, 3, 4, 6) //relatively weak
+	for(var/turf/T in range(4, impact))
+		T.ignite(60, 30) //cooking for a long time
```
psykzz, line 289: could this use flame_turf?

Rohesie: That is a gun proc, though, no?
Fire does need a bit of a refactor...

Rohesie: Used `flame_radius` which does what you requested, I believe.

psykzz: I meant so that you could remove the turf loop here.

Rohesie: oops, I meant to do that

PR #3982, opened 2020-05-04
Doors how do they work

jroinc:
## About The Pull Request

Door HP upped from 500 (default object HP) to 2000
Req door HP upped from 500 to 20000 and made unhackable.
Ideally, button with RO access added to toggle bolts on the req doors to allow him to lock it open if he wants to let people in.

## Why It's Good For The Game

Doors used to break in 30 knife hits, 27 rifle bullets, or 7 shotgun slugs. It's bad enough that people don't bother carrying crowbars for unpowered doors and can literally break down a door faster than you can hack it if you don't know the wires and it's not one of the first 3-4 ones.

By quadrupling this, engineering skill might be a little more valuable and we might see less unga-ing. It's still an option, sure, but stacking 3 ungas screaming with knives on a door is no longer always faster than hacking it, C-4'ing it, plasma-cutting it or JUST CROWBARING UNPOWERED DOORS.

Req doors are made near-indestructible and unhackable. I'm gonna see if Kwin can add a button toggle to them to give RO's the choice to lock them open for ungas, failing that, I'll make them hackable again. This is part of my move to make the only "weak spot" in the Req "defenses" the front desk, so that someone on duty can actually stare down the people trying to break in.

## Changelog
:cl:

balance: Buffed doors.

/:cl:


spookydonut:
i tried this years ago on cm it doesnt really solve the core issue which is that marines need stuff from reqs, which is often unstaffed or incompetent

jroinc:
> i tried this years ago on cm it doesnt really solve the core issue which is that marines need stuff from reqs, which is often unstaffed or incompetent

Agree, it's why I'm leaving one easy path in through the front door if there's NO ONE there at all. But if there's someone actually at the desk, they can have the option to try and do stuff.

InterroLouis:
I'm not quite getting how that's an easy path. Marines can't get in through the front desk because of the shutters. The shutters which aren't open unless req is manned because someone with access needs to open them.

jroinc:
> I'm not quite getting how that's an easy path. Marines can't get in through the front desk because of the shutters. The shutters which aren't open unless req is manned because someone with access needs to open them.

One, that's one reason this is a draft PR, trying to think of issues and I'm not satisfied this is an acceptable change as is.

Two, on distress pop, SOMEONE should have access. PO, MO, FC, MT, MP, CO, we just need one of those from a 30-player pool.

jroinc:
I've noticed the number of RO breakins where people knife the doors have gone down due to me banning people.
Additionally, now that ayys can open unpowered doors faster than ungas can shoot them down, I don't think this is needed.

PR #3990, opened 2020-05-05
Unequip

Rohesie:
* Fixes #3955

* Since `equipped()` reports the event of a slot being filled, added `unequipped()` which responds to the event of a slot being nulled.

There will be more refactors down the line, dropped() is a mess, but this should fix the issue and offer an improvement.


code/modules/mob/inventory.dm
```diff
@@ -188,19 +188,41 @@
...
+  * * Will return TRUE is successfully dropped.
+  * * Will return FALSE if the item can not be dropped due to TRAIT_NODROP via doUnEquip().
+  * * Will return null if there is no item.
+  * If the item can be dropped, it will be forceMove()'d to the ground and the turf's Entered() will be called.
+*/
+/mob/proc/dropItemToGround(obj/item/I, force = FALSE)
+	. = UnEquip(I, force, drop_location())
+	if(.)
+		I.pixel_x = rand(-6,6)
+		I.pixel_y = rand(-6,6)
```
psykzz, line 202: Is this ever reset? Could this lead to +6 every drop until eventually the item is off the turf?

Would `I.pixel_y = initial(I.pixel_y) + rand(-6,6)` or something be better?

Rohesie: Good catch, added a pixel-offset resetting when put on hands. It works properly now.

PR #4004, opened 2020-05-07
Facehuggers no longer instantly leap when dropped.

Pariah919:
## About The Pull Request
Facehuggers go idle when dropped.

## Why It's Good For The Game
This was never made. Seems to be a bug. 

## Changelog
:cl:
fix: Huggers no longer instantly leap when dropped. Instead going idle.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
@@ -152,6 +152,10 @@
 /obj/item/clothing/mask/facehugger/flamer_fire_act()
 	Die()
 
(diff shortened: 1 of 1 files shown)


terranaut1:
This is not a bug, just another nerf in a long list from you, except this one is disguised as a fix.
You don't even need to go further than the top of the `facehuggers.dm` file.
`#define MIN_ACTIVE_TIME 4 SECONDS //Time between being dropped and going idle`

psykzz:
That comment literally states they should go idle but currently they don't

Pariah919:
meh,i don't feel like dealing with this

Issue #3907, opened 2020-04-22
Scope LOS bug


David-Stormwell:
## Testmerges:#
#3877 #3900 #3887 #3888 #3884 #3889 #3873 #3777 #3607 #3830 #3894 #3901

## Reproduction:
When you click on something which you cannot see, such as past a boiler gas cloud you will shoot in a random direction such as to the right of you instead of shooting ahead.

XElectricX:
Confirming this also works on my own repo. Was using binoculars and clicking any dark boxes will make you look in a different direction. Doesn't seem to be a pattern in what direction you turn to.

psykzz:
Fixed in #4006

Issue #3292, opened 2019-11-09
Autofire doesn't work with scopes


Henriquekill9576-taken:
## Testmerges:

Round ID: 1779
BYOND Version: 512.1488
Server revision compiled on: 
Master commit: 7b60722339cd03a814401bf2f051beadb6ec2f43
The following pull requests are currently test merged:
#3283: 'New spacemandmm release' by spookydonut at commit 18cf9ff355
#3284: 'Minor APC cleanup' by Rohesie at commit 53d2fff134
#3287: 'unfuck shutter element' by spookydonut at commit 5bf010099c
#3280: 'Reverts IS to old storage' by Kuro-ai at commit ca9302db5a
#3275: 'Marine Snow Slowdown reduced from +75% move speed penalty to +50% move speed penalty' by BurgerLUA at commit 44cb26c22e
#3251: 'Reworks the Canterbury Cream Egg Shuttle' by BurgerLUA at commit e8cf1b17e9
#3248: 'HUD fix' by Rohesie at commit f60c1696b0
#3272: 'remove br memes' by Kuro-ai at commit d643d21bc7
#3273: 'Buffs machete,nerfs weed nodes' by Kuro-ai at commit 49fd6a8e6b
#3282: 'Respawn Game tab fix' by Rohesie at commit 2fdb1492e4
#3281: 'Riot Armor Nerf' by Isy232 at commit 605c6f613a
#3274: 'T64 DMR' by David-Stormwell-TGCM at commit 9b480e0d0c

Local commit: d5718775b2058d7f1926ac92fef1b56471f52f5a
Server tools version: /tg/station 13 Server v3.2.6.2

## Reproduction:

1. Get a gun with autofire and put a scope in it
2. Try to use the autofire while using the scope

Hughgent:
Current Master.

Turns out you CAN use autofire with a scope, but only if you click within your regular 7 tile view range.

Something to do with 'CLICKCATCHER_PLANE` @Rohesie

Rohesie:
```
/datum/component/automatic_fire/proc/process_shot()
switch(get_dist(shooter, target))
	if(8 to INFINITY) //Can technically only go as far as 127 right now.
		stop_autofiring() //Elvis has left the building.
```
It wasn't designed with scopes in mind. That said, it can be changed to match the user client's `view` value, but it should also take into account tile offsets and changed client eyes. Not trivial, but doable.

davidchan13:
[...]

PR #4006, opened 2020-05-08
Zoom fix

GLACoding:
## About The Pull Request

Fixes 2 bugs

Autofire has its range increased so it can be used while zoomed, keeping it consistent with regular fire #3292
Firing into darkness now finds the correct turf to target #3907

## Why It's Good For The Game

fixes the bug with snipers turning 180° and FFing when firing into darkness

## Changelog
:cl:
fix: fixes some zoom issues
/:cl:


code/datums/components/full_auto_fire.dm
```diff
@@ -281,7 +285,7 @@
 		if(-1 to 0)
 			target = get_step(shooter, shooter.dir) //Shoot in the direction faced if the mouse is on the same tile as we are.
 			target_loc = target
-		if(8 to INFINITY) //Can technically only go as far as 127 right now.
+		if(127 to INFINITY) //Can technically only go as far as 127 right now.
```
Rohesie, line 288: We'll have to remove the `switch()` here to preserve the check.

Store the result of `get_dist(shooter, target)` in a variable, check if smaller or equal to zero, and do the stuff inside the `if(-1 to 0)` then.
And after that `else if(in_view_range(shooter, target))`

code/_onclick/click.dm
```diff
@@ -19,7 +19,7 @@
 /atom/Click(location, control, params)
 	if(flags_atom & INITIALIZED)
 		SEND_SIGNAL(src, COMSIG_CLICK, location, control, params, usr)
-		usr.ClickOn(src, params)
+		usr.ClickOn(src, params, location)
```
Rohesie, line 22:
```suggestion
		usr.ClickOn(src, location, control, params)
```
Propagate this to all the other stances of `ClickOn(` so we preserve the former proc's argument order.

code/datums/components/full_auto_fire.dm
```diff
@@ -148,6 +148,8 @@
 		modifiers["icon-x"] = num2text(ABS_PIXEL_TO_REL(text2num(modifiers["icon-x"])))
 		modifiers["icon-y"] = num2text(ABS_PIXEL_TO_REL(text2num(modifiers["icon-y"])))
 		params = list2params(modifiers)
+		if(source.pixel_x != 0 || source.pixel_y != 0)
+			target = darkness_adjust(target, shooter)
```
Rohesie, line 152: This issue should be happening only with regular clicks, not with autofire, the solution for it being in the code right here over your addition.
Is that not correct?

GLACoding: it does work with no tile offset, but doesnt work if the client is zoomed

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1423,6 +1423,7 @@ Defined in conflicts.dm of the #defines folder.
 	if(bipod_deployed)
 		bipod_deployed = FALSE
 		to_chat(user, "<span class='notice'>You retract [src].</span>")
+		master_gun.aim_slowdown -= 1
```
Rohesie, line 1426: Atomize this change into a different PR so it doesn't get delayed with the review of the other ones.

Try to keep one PR per fix / feature whenever sensible.

code/__HELPERS/view.dm
```diff
@@ -9,13 +9,25 @@
...
+	if (offsetY)
+		viewYpos += (offsetY/32)
+		viewYneg -= (offsetY/32)
+
+	return list(viewXneg, viewXpos, viewYneg, viewYpos)
 
 
 /proc/in_view_range(mob/user, atom/A)
-	var/list/view_range = getviewsize(user.client.view)
+	var/list/view_range = getviewsize(user.client.view, user.client.pixel_x, user.client.pixel_y)
```
Rohesie, line 30:
```suggestion
	var/list/view_range = getviewsize(user.client.view)
```

code/__HELPERS/view.dm
```diff
@@ -9,13 +9,25 @@
...
 	var/turf/source = get_turf(user)
 	var/turf/target = get_turf(A)
-	var/view_x = CEILING(view_range[1]* 0.5, 1)
-	var/view_y = CEILING(view_range[2]* 0.5, 1)
-	return ISINRANGE(target.x, source.x - view_x, source.x + view_x) && ISINRANGE(target.y, source.y - view_y, source.y + view_y)
+	return ISINRANGE(target.x, source.x - view_range[1], source.x + view_range[2]) && ISINRANGE(target.y, source.y - view_range[3], source.y + view_range[4])
```
Rohesie, line 33:
```suggestion
	var/x_offset = user.client.pixel_x / 32
	var/y_offset = user.client.pixel_y / 32
	return ISINRANGE(target.x, source.x - view_x + x_offset, source.x + view_x + x_offset) && ISINRANGE(target.y, source.y - view_y + y_offset, source.y + view_y + y_offset)
```

code/__HELPERS/view.dm
```diff
@@ -1,4 +1,4 @@
-/proc/getviewsize(view)
+/proc/getviewsize(view, offsetX, offsetY)
```
Rohesie, line 1:
```suggestion
/proc/getviewsize(view)
```

code/__HELPERS/view.dm
```diff
@@ -9,13 +9,25 @@
...
+	var/viewYpos = viewY
+	var/viewYneg = viewY
+	if (offsetX)
+		viewXpos += (offsetX/32)
+		viewXneg -= (offsetX/32)
+	if (offsetY)
+		viewYpos += (offsetY/32)
+		viewYneg -= (offsetY/32)
+
+	return list(viewXneg, viewXpos, viewYneg, viewYpos)
```
Rohesie, line 26:
```suggestion
	return list(viewX, viewY)
```

Rohesie:
Modify `params2turf()` from this
```cpp
	tX = CLAMP(origin.x + text2num(tX) - round(actual_view[1] / 2) - 1, 1, world.maxx)
	tY = CLAMP(origin.y + text2num(tY) - round(actual_view[2] / 2) - 1, 1, world.maxy)
```
to this
```cpp
	tX = CLAMP(origin.x + text2num(tX) - round(actual_view[1] / 2) + (C?.pixel_x / 32) - 1, 1, world.maxx)
	tY = CLAMP(origin.y + text2num(tY) - round(actual_view[2] / 2) + (C?.pixel_y / 32) - 1, 1, world.maxy)
```
Let me know if that doesn't fix it.

code/_onclick/click.dm
```diff
@@ -60,6 +61,12 @@
 		return
 
 	var/list/modifiers = params2list(params)
+	if(isnull(location) && isturf(A)) //Checks if the intended target is in deep darkness and adjusts A based on params.
```
Rohesie, line 64: This check is not intended for turfs. It's for the click catcher.

Rohesie:
To clarify on the click catcher, the dark area where clicks are missfiring is a screen object whose path is `/obj/screen/click_catcher`, the click catcher.
It's function is to block unlawful clicks, that is, those outside of your field of vision, besides blocking your view.
That is why you'll find this code here: `/obj/screen/click_catcher/Click(location, control, params)` What it does is transfer the click to the turf under it.

There's a set of variables being transferred with the click, you can see the reference here: http://www.byond.com/docs/ref/index.html#/DM/mouse
In here specifically we are updating `icon-x,` and `icon-y`, which are the pixel positions in where our click landed exactly on the object, because we are transferring the click to the turf, which is 32 pixels wide and long, from the click catcher, which can be a potentially pretty large dark area, and where in it we clicked is different from the pixel coordinates if we had clicked on the turf directly. We use this value to determine exactly where we clicked elsewhere down the line.

Ideally you should handle this at that level, making sure the turf being clicked is the correct one, through `params2turf`, which gets these parameters:
1) `screen-loc`: this is, in pixel coordinates, where you clicked on your screen.
2) The center turf where your character is when you clicked.
3) Your client, from where your `view` is taken from.

With these 3 variables you can triangulate on which exact turf you clicked under normal circumstances. The problem here is that the calculation is not taking offset into account. But that's a set of variables that is also held in the client, in `pixel_x` and `pixel_y`.

Makes sense?

code/__HELPERS/view.dm
```diff
@@ -16,6 +16,8 @@
 	var/list/view_range = getviewsize(user.client.view)
 	var/turf/source = get_turf(user)
 	var/turf/target = get_turf(A)
-	var/view_x = CEILING(view_range[1]* 0.5, 1)
-	var/view_y = CEILING(view_range[2]* 0.5, 1)
-	return ISINRANGE(target.x, source.x - view_x, source.x + view_x) && ISINRANGE(target.y, source.y - view_y, source.y + view_y)
+	var/view_x = FLOOR(view_range[1]* 0.5, 1)
+	var/view_y = FLOOR(view_range[2]* 0.5, 1)
```
Rohesie, line 20:
```suggestion
	var/view_x = round(view_range[1]* 0.5)
	var/view_y = round(view_range[2]* 0.5)
```
`round(x, 1)` will round up or down, but `round(x)` will act like floor. Little BYOND quirk.

code/_onclick/click.dm
```diff
@@ -60,6 +61,12 @@
 		return
 
 	var/list/modifiers = params2list(params)
+	if(isnull(location) && A.plane == CLICKCATCHER_PLANE) //Checks if the intended target is in deep darkness and adjusts A based on params.
+		A = params2turf(modifiers["screen-loc"], get_turf(src.client.eye), src.client)
```
Rohesie, line 65:
```suggestion
		A = params2turf(modifiers["screen-loc"], get_turf(client.eye), client)
```
`src` is implied, it's not necessary. Avoid it unless there's two variables with the same name to require the distinction.

code/datums/components/full_auto_fire.dm
```diff
@@ -277,13 +277,12 @@
...
-		if(-1 to 0)
-			target = get_step(shooter, shooter.dir) //Shoot in the direction faced if the mouse is on the same tile as we are.
-			target_loc = target
-		if(8 to INFINITY) //Can technically only go as far as 127 right now.
-			stop_autofiring() //Elvis has left the building.
-			return FALSE
+	if(get_dist(shooter, target) <= 0)
+		target = get_step(shooter, shooter.dir) //Shoot in the direction faced if the mouse is on the same tile as we are.
+		target_loc = target
+	else if(!in_view_range(shooter, target)) //Can technically only go as far as 127 right now.
```
Rohesie, line 283:
```suggestion
	else if(!in_view_range(shooter, target))
```
Comment is no longer relevant.

code/__DEFINES/dcs/signals.dm
```diff
@@ -198,7 +198,7 @@
 #define COMSIG_MOB_REVIVE "mob_revive"							//from base of mob/on_revive(): ()
 #define COMSIG_MOB_MOUSEDOWN "mob_mousedown"					//from /client/MouseDown(): (atom/object, turf/location, control, params)
 #define COMSIG_MOB_MOUSEUP "mob_mouseup"						//from /client/MouseUp(): (atom/object, turf/location, control, params)
-#define COMSIG_MOB_CLICKON "mob_clickon"						//from base of mob/clickon(): (atom/A, params)
+#define COMSIG_MOB_CLICKON "mob_clickon"						//from base of mob/clickon(): (atom/A, location, control, params)
```
spookydonut, line 201: `if(SEND_SIGNAL(src, COMSIG_MOB_CLICKON, A, params) & COMSIG_MOB_CLICK_CANCELED)`
This comment was correct

code/__DEFINES/dcs/signals.dm
```diff
@@ -219,7 +219,7 @@
 #define COMSIG_MOB_TOGGLEMOVEINTENT "mob_togglemoveintent"		//drom base of mob/toggle_move_intent(): (new_intent)
 
 //mob/dead/observer
-#define COMSIG_OBSERVER_CLICKON "observer_clickon"				//from mob/dead/observer/ClickOn(): (atom/A, params)
+#define COMSIG_OBSERVER_CLICKON "observer_clickon"				//from mob/dead/observer/ClickOn(): (atom/A, location, control, params)
```
spookydonut, line 222: `if(SEND_SIGNAL(src, COMSIG_OBSERVER_CLICKON, A, params) & COMSIG_MOB_CLICK_CANCELED)`

This comment was correct

code/_onclick/click.dm
```diff
@@ -45,7 +45,8 @@
 	* item/afterattack(atom, user, adjacent, params) - used both ranged and adjacent when not handled by attackby
 	* mob/RangedAttack(atom, params) - used only ranged, only used for tk and laser eyes but could be changed
 */
-/mob/proc/ClickOn(atom/A, params)
+/mob/proc/ClickOn(atom/A, location, control, params)
```
spookydonut, line 48: as far as i can tell nothing is using `control`

psykzz:
Runtimes triggering
```
Runtime in view.dm, line 23: Cannot read null.x
proc name: in view range (/proc/in_view_range)

call stack:
in view range(Elric \'Wolf\' Hess (/mob/living/carbon/human), the Medical Airlock (/obj/machinery/door/airlock/multi_tile/mainship/medidoor))
/datum/component/automatic_fir... (/datum/component/automatic_fire): process shot()
...
```

image

code/datums/components/full_auto_fire.dm
```diff
@@ -137,6 +137,8 @@
 		return
 	if(get_dist(source.mob, target) < 2) //Adjacent clicking.
 		return
+	if(QDELETED(target))
+		return
```
Rohesie, line 141: How can you click on a qdeleted target?

GLACoding: the runtime occurs when a target is destroyed just as you begin autofiring on it, the timing is tight but it can happen.

Rohesie: Are you entirely sure the target exists and its `gc_destroyed` has a truthy value instead of `params2turf` having returned `null`? Because that would be pretty unusual.

GLACoding: params2turf doesn't get triggered at all when reproducing the runtime

PR #4007, opened 2020-05-08
Fixes error in SG magazine description

GLACoding:
## About The Pull Request

fixes a small calibre error in the T29 magazine description

### DIFF
--- a/code/game/objects/machinery/vending/vending.dm
+++ b/code/game/objects/machinery/vending/vending.dm
@@ -572,10 +572,11 @@ GLOBAL_LIST_INIT(vending_white_items, typecacheof(list(
 						return
 
 			else if(istype(item_to_stock, /obj/item/ammo_magazine))
-				var/obj/item/ammo_magazine/A = item_to_stock
-				if(A.current_rounds < A.max_rounds)
-					to_chat(user, "<span class='warning'>[A] isn't full. Fill it before you can restock it.</span>")
-					return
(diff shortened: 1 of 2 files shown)


Pariah919:
oh no

Rohesie:
The infinite ammunition through restocking is less a bug we want to turn into a feature (more like we tolerate it), instead of something we want to change through a proper ammo chain of supply.

The name fix is good and should be preserved, but increase the amount of ammo if it's too small in quantity. That is what should have happened with the powerpacks before. Don't make a source of infinite ammo.

GLACoding:
The way i look at it is:
Regular PFC weapons are effectively infinite ammo, due the large amount from the vendors that can be backpacked up and brought down to the surface

Specialist weapons have limited ammo, only available from cargo, to reflect the high power of their weapons

SGs had a middle ground, they couldn't bring a large amount of ammo (2 per SG, not counting the 1 available from the munitions vendor) But this ammo could be refilled without biting into the economy, which is already being drained by spec, turret and other high priority requests.

The SG is a good weapon but not really as powerful as the specs to warrant a constant drain on requisitions points.

Rohesie:
Your concept is not bad, we could make it so the ammo holders themselves are expensive, but recharging is not. Ideally that is done with recharging machines, which is more intuitive than the restocking bug.

That said, this is also about code quality, adding chains of `istype` checks is not very kosher.

Choose between implementing a recharging machine, and I could help you with doing it, or making ammo a bit less scarce if needed.

PR #4009, opened 2020-05-08
Fixes razorwire runtime and cleans up code

Rohesie:
* Turns `entangled_list` into a lazy list.
* Turns the set_frozen into a proper trait for source tracing.

It seems things were getting deleted but not cleared up from `entangled_list`

### DIFF
--- a/code/game/objects/structures/razorwire.dm
+++ b/code/game/objects/structures/razorwire.dm
@@ -65,10 +65,10 @@
 /obj/structure/razorwire/proc/do_razorwire_tangle(mob/living/entangled)
 	ADD_TRAIT(entangled, TRAIT_IMMOBILE, type)
 	ENABLE_BITFIELD(entangled.restrained_flags, RESTRAINED_RAZORWIRE)
-	entangled.entangled_by = src
 	LAZYADD(entangled_list, entangled) //Add the entangled person to the trapped list.
 	RegisterSignal(entangled, COMSIG_LIVING_DO_RESIST, /atom/movable.proc/resisted_against)
 	RegisterSignal(entangled, COMSIG_PARENT_QDELETING, .proc/do_razorwire_untangle)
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/razorwire.dm
```diff
@@ -74,39 +80,40 @@
...
+	var/armor_block = entangled.run_armor_check(def_zone, "melee")
+	entangled.apply_damage(rand(RAZORWIRE_BASE_DAMAGE * 0.8, RAZORWIRE_BASE_DAMAGE * 1.2), BRUTE, def_zone, armor_block, TRUE) //Apply damage as we tear free
+	UPDATEHEALTH(entangled)
 	return TRUE
 
 
+/obj/structure/razorwire/proc/do_razorwire_untangle(mob/living/entangled)
+	UnregisterSignal(entangled, list(COMSIG_PARENT_QDELETING, COMSIG_LIVING_DO_RESIST))
+	LAZYREMOVE(entangled_list, entangled)
+	entangled.entangled_by = null
```
spookydonut, line 98: uh why not just make entangled from itself in its destroy?

spookydonut: actually you can probably just get rid of `/mob/living/var/entangled_by`

spookydonut:
```
/atom/movable/Uncross(atom/movable/AM, atom/newloc)
	. = ..()
	if(SEND_SIGNAL(src, COMSIG_MOVABLE_UNCROSS, AM) & COMPONENT_MOVABLE_BLOCK_UNCROSS)
```
register on that signal instead on things you entangle

Rohesie: Sure. This is Sur code, I didn't stop to improve it because I was fixing other stuff.
I'll see about cleaning it.

spookydonut: its basically the cause of the runtime seemed better to treat the cause.

PR #4010, opened 2020-05-08
Removes set_freeze() occurences in the code in favor of setting the immobile trait

Rohesie:
* Also properly implements the `Immoblize()` status effect procs to be used for set-duration freezes.

~~Related to #4009 , if one gets merged the other should finish removing `set_freeze()` unless if we turn it into an admin tool.~~

* Completely removes the `frozen` variable and `set_freeze()` proc.


code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
```diff
@@ -110,7 +110,7 @@
 	M.Paralyze(paralyze_time)
 
 	step_to(X, M)
-	X.stop_movement()
+	X.Immobilize(X.xeno_caste.charge_type == CHARGE_TYPE_SMALL ? 0.5 SECONDS : 1.5 SECONDS)
```
spookydonut, line 113: noooo please
trying to get rid of charge_type dont add more stuff that uses it!
move it to like `/datum/action/xeno_action/activable/pounce/var/pounce_immobilize_time`

Rohesie:
Ready for merge.

Rohesie:
Conflict fixed.

PR #4025, opened 2020-05-09
Support draw_mode for internal storage

psykzz:
## About The Pull Request

This allows you to set draw_mode on internal storage where, when clicked it will remove the last item from the storage instead of opening the UI.

## Why It's Good For The Game

Adds the required support for the marine boot knife


Rohesie:
I prefer the alt+click option, but this works.

PR #4028, opened 2020-05-10
Fixes explosions happening at the same time as crash

Rohesie:
* Adds a pre-crash proc for both the shuttle and destination to happen one second before crashing.
* It allows for other pre-arrival interactions.
* Does PsyKzz's request of allowing the admins to do instant shuttle crashes without sacrificing the explosions.

### DIFF
--- a/code/__DEFINES/shuttles.dm
+++ b/code/__DEFINES/shuttles.dm
@@ -10,6 +10,11 @@
 #define SHUTTLE_RECHARGING	"recharging"
 #define SHUTTLE_PREARRIVAL	"pre-arrival"
 
+///Whether specific shuttle effects happen on arrivals or not.
+#define NO_ARRIVALS_INTERACTION 0
+#define SHUTTLE_ARRIVING 1
+#define SHUTTLE_ARRIVED 2
+
 #define EMERGENCY_IDLE_OR_RECALLED (SSshuttle.emergency && ((SSshuttle.emergency.mode == SHUTTLE_IDLE) || (SSshuttle.emergency.mode == SHUTTLE_RECALL)))
 #define EMERGENCY_ESCAPED_OR_ENDGAMED (SSshuttle.emergency && ((SSshuttle.emergency.mode == SHUTTLE_ESCAPE) || (SSshuttle.emergency.mode == SHUTTLE_ENDGAME)))
 #define EMERGENCY_AT_LEAST_DOCKED (SSshuttle.emergency && SSshuttle.emergency.mode != SHUTTLE_IDLE && SSshuttle.emergency.mode != SHUTTLE_RECALL && SSshuttle.emergency.mode != SHUTTLE_CALL)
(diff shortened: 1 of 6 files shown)


spookydonut:
one second before is hacky and not a real fix, dnm, test merge is ok

Rohesie:
What would be the non-hacky fix?

spookydonut:
> 
> 
> What would be the non-hacky fix?

what we discussed on discord, explosion accountability

Rohesie:
> > What would be the non-hacky fix?
> 
> what we discussed on discord, explosion accountability

A lot was discussed. Explosions are already indexed by the epicenter, so you could have multiple explosions on the same turf as long from different sources.

What I don't know is how you'd want to fix this. The crash explosions need to happen before the ship crashes. They are not instant anymore, they use a processing subsystem. How do you want for them not to blow the ship if they are not timed a little bit earlier?

spookydonut:
> 
> 
> > > What would be the non-hacky fix?
> > 
> > 
> > what we discussed on discord, explosion accountability
> 
> A lot was discussed. Explosions are already indexed by the epicenter, so you could have multiple explosions on the same turf as long from different sources.
> 
> What I don't know is how you'd want to fix this. The crash explosions need to happen before the ship crashes. They are not instant anymore, they use a processing subsystem. How do you want for them not to blow the ship if they are not timed a little bit earlier?

I want them to allow callbacks when they're finished rather than relying on scheduling/cpu power determine how long a period you need to make the explosions before the shuttle moves

Rohesie:
> 
> 
> > > > What would be the non-hacky fix?
> > > 
> > > 
> > > what we discussed on discord, explosion accountability
> > 
> > 
> > A lot was discussed. Explosions are already indexed by the epicenter, so you could have multiple explosions on the same turf as long from different sources.
> > What I don't know is how you'd want to fix this. The crash explosions need to happen before the ship crashes. They are not instant anymore, they use a processing subsystem. How do you want for them not to blow the ship if they are not timed a little bit earlier?
> 
> I want them to allow callbacks when they're finished rather than relying on scheduling/cpu power determine how long a period you need to make the explosions before the shuttle moves

Ah, feasible. You'll have to store the explosions not only by an epicenter index then, but also by an ID or something like the old datum then. Since you showed concerns on each explosion creating a datum we can have the SS process one explosion ID at a time, and send a signal when each is over.

spookydonut:
> Ah, feasible. You'll have to store the explosions not only by an epicenter index then, but also by an ID or something like the old datum then. Since you showed concerns on each explosion creating a datum we can have the SS process one explosion ID at a time, and send a signal when each is over.

that's the same cost basically as processing per datum, lummox has posted that he's improved slightly the performance of list[index] where index is a number so i can probably do this without having to change how they process.

this pr will do for now as a TM or even a merge and i'll just fix it properly when i get to it, i don't want to rush it

Rohesie:
Close this whenever you stop TMing it.

PR #4030, opened 2020-05-10
auto-fit-viewport pref

Rohesie:
You login, the screen adjusts correctly. Handy as /tg/ now uses a different screen wideness, so you can just log into either and it will auto-adjust without having to hit the button yourself each time.


Pariah919:
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

PR #4037, opened 2020-05-10
removes cooldown from OB

BobSlim:
## About The Pull Request

removes all mention of a cooldown from the orbital cannon. removes the last_fired variable, which tracked when it was last fired in order to calculate cooldown.

## Why It's Good For The Game

OB is whack. 
The time it takes to load the thing is CD enough. 
It's really not powerful enough to justify any cooldown at all.
Limited by number of warheads. 6 shots isn't much, really. Limited by number of active OB beacons, too.

## Changelog
:cl:
del: the orbital cannon has no cooldown!
/:cl:


Rohesie:
There are still references to `last_orbital_firing`

jroinc:
Thanks!

I was gonna do one and lower the C/D to 10 seconds, but this works too. At least it's not 500 anymore.

XElectricX:
Now STs can actually do something other than do cargo stuff.

PR #4039, opened 2020-05-10
Fixes unintentional doubleclicks on the item action ui

Rohesie:
`attack_self` calls for `toggle_item_state`


psykzz:
I dont see where the other attack self is being called here?

Rohesie:
What was happening was that `ui_action_click()` was calling first `toggle_item_state()` and then `attack_self()`, which in turn calls `toggle_item_state()` (again)

PR #4036, opened 2020-05-10
increase head max damage to 150

BobSlim:
## About The Pull Request

1 variable change, head increased hp

## Why It's Good For The Game

better for people without helmets

if you've taken 150 brute damage and you don't have a helmet for some reason you're dead anyway.

this just means that people without helmets won't also have an effective decrease in hp by 2/3.

## Changelog
:cl:
balance: increased the amount of damage (100 -> 150) heads can take before decap. Hurrah for headbands!
/:cl:

PR #4042, opened 2020-05-11
Revert "increase head max damage to 150 (#4036)"

Pariah919:
This reverts commit f32865982b37a431ee5ef5ab94bbcb2faa482675.

## About The Pull Request
Decreases max hp for heads back down to 100

## Why It's Good For The Game
I don't get why this was merged in the first place other than the fact that people believe that non-helmet shouldn't cause death?
Also a large majority of people disagreed with the change anyway, and it was pushed through anyhow.
100 head HP is around 4 slashes from a normal bean if you're wearing one of the berets or caps anyway. I don't see the need for this.

## Changelog
:cl:
balance: people no longer have 150 head hp. Have 100 again.
:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -1204,7 +1204,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	name = "head"
 	icon_name = "head"
 	display_name = "head"
-	max_damage = 150
+	max_damage = 100
 	min_broken_damage = 40
 	body_part = HEAD
 	vital = TRUE


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -5. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

Rohesie:
Non-helmet does cause death. This only affects decapitation threshold.

Pariah919:
The decapitation threshold was fine where it was at before. I don't see the point of raising it any further.

Rohesie:
Since the argument to keep it low is "it's funny to see people decapitated" I believe it better to not to follow that. Schadenfreude isn't a great north when designing the game, IMO. The idea is for the players to have fun, instead, not for the observers to have a laugh at other's grief.

Pariah919:
Doesn't seem like the people in charge want this anyway.
I disagree and think it's dumb as heck but whatever.

PR #4047, opened 2020-05-12
you can refill mags with other mags

BobSlim:
## About The Pull Request

Removes handful requirement from mag refilling.
Adds delay if it is not a handful.

## Why It's Good For The Game

While it is "realistic" that you'd have to juggle around ammo and handfuls and mags, it simply isn't conducive to simulate that in the game environment. This makes refilling mags less of a pain.

## Changelog
:cl:
add: You can refill mags with other mags, now.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -69,14 +69,11 @@ They're all essentially identical when it comes to getting the job done.
 	. = ..()
 
 	if(istype(I, /obj/item/ammo_magazine))
-		var/obj/item/ammo_magazine/MG = I
-		if(!(MG.flags_magazine & AMMUNITION_HANDFUL)) //got a handful of bullets
-			return
 
 		if(!(flags_magazine & AMMUNITION_REFILLABLE)) //and a refillable magazine
 			return
 
-		var/obj/item/ammo_magazine/handful/H = I
+		var/obj/item/ammo_magazine/H = I
 		if(src != user.get_inactive_held_item()) //It has to be held.
 			to_chat(user, "Try holding [src] before you attempt to restock it.")
 			return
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/ammunition.dm
```diff
@@ -98,6 +94,17 @@ They're all essentially identical when it comes to getting the job done.
 		to_chat(user, "The rounds don't match up. Better not mix them up.")
 		return
 
+	if(!source.current_rounds)
+		to_chat(user, "[source] is empty.")
```
psykzz, line 98:
```suggestion
		to_chat(user, "<span class='warning'>\The [source] is empty.</span>")
```

code/modules/projectiles/ammunition.dm
```diff
@@ -98,6 +94,17 @@ They're all essentially identical when it comes to getting the job done.
 		to_chat(user, "The rounds don't match up. Better not mix them up.")
 		return
 
+	if(!source.current_rounds)
+		to_chat(user, "<span class='warning'>\The [source] is empty.</span>")
+		return
+
+	if(!istype(source, /obj/item/ammo_magazine/handful)) //if it isn't a handful, give a delay.
+		to_chat(user, "<span class='notice'>You start refilling [src] with [source].</span>")
+		if(!do_after(user, 15, TRUE, src, BUSY_ICON_GENERIC))
```
psykzz, line 103:
```suggestion
		if(!do_after(user, 1.5 SECONDS, TRUE, src, BUSY_ICON_GENERIC))
```

code/modules/projectiles/ammunition.dm
```diff
@@ -97,8 +97,8 @@ They're all essentially identical when it comes to getting the job done.
 	if(!source.current_rounds)
 		to_chat(user, "<span class='warning'>\The [source] is empty.</span>")
 		return
-
-	if(!istype(source, /obj/item/ammo_magazine/handful)) //if it isn't a handful, give a delay.
+//using handfuls; and filling internal mags has no delay.
+	if(!istype(source, /obj/item/ammo_magazine/handful) && !istype(src, /obj/item/ammo_magazine/internal) )
```
BobSlim, line 101: is this hacky?

psykzz: Have you tested this?
istype is a specific check it wont work for child objects.

You could instead try `ispath(type, /some/type/path`

BobSlim: It seemingly works, haven't heard any complaints, but I'll swap it around anyway

psykzz: It may be you never do checks against other types and this is good enough.

PR #4067, opened 2020-05-15
Stamina regen nerf

Rohesie:
The old design was that walking was a rare thing and big debuf, so you'd be rewarded by quick stamina regen. Currently it's the norm and makes xeno tackles much less effective.


code/modules/mob/living/living.dm
```diff
@@ -41,7 +41,7 @@
 
 
 /mob/living/proc/handle_staminaloss()
-	if(world.time < last_staminaloss_dmg + 3 SECONDS || (m_intent == MOVE_INTENT_RUN && world.time < last_move_intent + 1 SECONDS))
+	if(world.time < last_staminaloss_dmg + 3 SECONDS)
```
psykzz, line 44: Would now be a good time to turn this into a `cooldown`?

jroinc:
Is there a cheap way to tie pain into either max stamina or regen rate that you can think of?

Currently, they aren't connected, so someone at -49 HP is as difficult to tackle down as someone at 100% HP.

code/modules/mob/living/living.dm
Rohesie: This would have to be a stoppable timer, not a regular cooldown, which is more expensive.
I don't see much gain in it, to be fair, this method is by far the cheapest.

Rohesie:
Could you describe a bit better the functionality you want?

psykzz:
I believe its the outcome of the work being done by @Hughgent in https://github.com/tgstation/TerraGov-Marine-Corps/pull/4075

PR #4071, opened 2020-05-16
Fixes green flame bypassing fog

psykzz:
Fixes green flame bypassing fog


psykzz:
This still doesnt capture the tip of the flame if you do it at an angle since the loop is a little off, but will catch more direct fire at least.

PR #4075, opened 2020-05-17
Xeno Stun now based on current health

Hughgent:
## About The Pull Request

In searching how Stamina based stuns were applied, it seems that they were only applied if the amount of stamina damage was greater than their current `maxHealth * 1.5`

This changes it to now be based on current `health`

## Why It's Good For The Game

Xeno stuns are/should be vastly easier on hurt marines.

## Changelog
:cl: Hughgent
balance: Xeno Disarm based stuns are now based on current health rather than max health, should be much easier to stun a marine.
/:cl:


Hughgent:
In testing, a marine wearing the FC armor set was stunned via stamina loss in 17 young drone hits. 

Applying Slashing from that same young drone until the health bar showed red, approximately 1/4 of regular health remaining, the marine in FC armor was stunned via stamina loss in 12 hits.

Hughgent:
This also affects running and other stamina based activities, if you're hurt, you can't run as far.

PR #4089, opened 2020-05-19
Temporary Fog

Hughgent:
## About The Pull Request

This makes temporary opaque fog to block vision and create ambiance for mapping.

## Why It's Good For The Game

No longer shall fog be only for blocking of areas. Now it can be used to make places just dark and scary to walk through as you can't see anything beyond your hand. 

This is true london pea soup fog.

Once anything steps through it, the fog disappears for X(30) seconds. They also disappear with the rest of the regular blocking fog.

## Changelog
:cl: Hughgent
add: Fog that is actually fog like, and not wall like.
/:cl:

### DIFF
--- a/code/game/objects/effects/misc.dm
+++ b/code/game/objects/effects/misc.dm
@@ -147,10 +147,7 @@
 
 /obj/effect/forcefield/fog/passable_fog/update_icon()
 	. = ..()
-	if(invisible)
-		icon_state = NONE
-	else
-		icon_state = "smoke"
+	alpha = invisible ? 0 : initial(alpha)
 
 //used to control opacity of multitiles doors
 /obj/effect/opacifier
--- a/code/game/objects/effects/misc.dm
+++ b/code/game/objects/effects/misc.dm
@@ -124,7 +124,6 @@
 	desc = "It looks dangerous to traverse."
 	icon = 'icons/effects/effects.dmi'
 	icon_state = "smoke"
-	opacity = TRUE
 	density = FALSE
 
 /obj/effect/forcefield/fog/passable_fog/CanPass(atom/movable/mover, turf/target)
@@ -136,7 +135,7 @@
 		return
 	set_opacity(FALSE)
 	alpha = 0
-	addtimer(CALLBACK(src, .proc/reset), 30 SECONDS, TIMER_UNIQUE|TIMER_OVERRIDE)
+	addtimer(CALLBACK(src, .proc/reset), 30 SECONDS)
 
 /obj/effect/forcefield/fog/passable_fog/proc/reset()
 	alpha = initial(alpha)
--- a/code/game/objects/effects/misc.dm
+++ b/code/game/objects/effects/misc.dm
@@ -126,27 +126,18 @@
 	icon_state = "smoke"
 	opacity = TRUE
 	density = FALSE
-	var/invisible = FALSE
 
 /obj/effect/forcefield/fog/passable_fog/CanPass(atom/movable/mover, turf/target)
 	return TRUE
 
 /obj/effect/forcefield/fog/passable_fog/Crossed(atom/movable/mover, oldloc)
 	. = ..()
-	invisible = TRUE
-	opacity = FALSE
-	update_icon()
-	addtimer(CALLBACK(src, .proc/icon_update_check), 30 SECONDS)
+	set_opacity(FALSE)
+	addtimer(CALLBACK(src, .proc/go_invisible), 30 SECONDS, TIMER_OVERRIDE|TIMER_UNIQUE)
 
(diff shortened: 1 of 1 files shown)


Hughgent:
I'm not certain, but this might runtime if a timer is still running (a mob has passed through a fog tile) when the 'fog drops' as that qdel(src) the thing with the timer.

Pariah919:
>more admin abuse tools
close this PR right now. Please. I beg of you.

code/game/objects/effects/misc.dm
```diff
@@ -119,6 +119,38 @@
...
+	opacity = TRUE
+	set_opacity(opacity)
+	update_icon()
+
+/obj/effect/forcefield/fog/passable_fog/update_icon()
+	. = ..()
+	if(invisible)
+		icon_state = NONE
+	else
+		icon_state = "smoke"
```
psykzz, line 153:
```suggestion
	alpha = invisible ? 0 : initial(alpha)
```

code/game/objects/effects/misc.dm
```diff
@@ -119,6 +119,38 @@
...
+	opacity = FALSE
+	set_opacity(opacity)
+	update_icon()
+	addtimer(CALLBACK(src, .proc/icon_update_check), 30 SECONDS)
+
+/obj/effect/forcefield/fog/passable_fog/proc/icon_update_check()
+	invisible = FALSE
+	opacity = TRUE
+	set_opacity(opacity)
+	update_icon()
```
psykzz, line 146: I'd move all this into update_icon save the extra proc call

psykzz:
> I'm not certain, but this might runtime if a timer is still running (a mob has passed through a fog tile) when the 'fog drops' as that qdel(src) the thing with the timer.

timers are cleaned up on qdel

code/game/objects/effects/misc.dm
Hughgent: This works

code/game/objects/effects/misc.dm
```diff
@@ -119,6 +119,34 @@
...
+
+/obj/effect/forcefield/fog/passable_fog/proc/icon_update_check()
+	invisible = FALSE
+	opacity = TRUE
+	update_icon()
+
+/obj/effect/forcefield/fog/passable_fog/update_icon()
+	. = ..()
+	alpha = invisible ? 0 : initial(alpha)
+	set_opacity(opacity)
```
psykzz, line 149:
```suggestion
/obj/effect/forcefield/fog/passable_fog
	name = "fog"
	desc = "It looks dangerous to traverse."
	icon = 'icons/effects/effects.dmi'
	icon_state = "smoke"
	opacity = TRUE
	density = FALSE

/obj/effect/forcefield/fog/passable_fog/CanPass(atom/movable/mover, turf/target)
	return TRUE

/obj/effect/forcefield/fog/passable_fog/Crossed(atom/movable/mover, oldloc)
	. = ..()
	set_opacity(FALSE)
	addtimer(CALLBACK(src, .proc/go_invisible), 30 SECONDS, TIMER_OVERRIDE|TIMER_UNIQUE)

/obj/effect/forcefield/fog/passable_fog/proc/go_invisible()
	alpha = opacity ? initial(alpha) : 0
	set_opacity(TRUE)
```

github suggestions suck

Hughgent:
as per that suggestion, it doesn't work

psykzz:
@Hughgent curious what happened
This worked locally for me
```
/obj/effect/forcefield/fog/passable_fog
	name = "fog"
	desc = "It looks dangerous to traverse."
	icon = 'icons/effects/effects.dmi'
	icon_state = "smoke"
	opacity = TRUE
	density = FALSE

/obj/effect/forcefield/fog/passable_fog/CanPass(atom/movable/mover, turf/target)
	return TRUE

/obj/effect/forcefield/fog/passable_fog/Crossed(atom/movable/mover, oldloc)
	. = ..()
	opacity = FALSE
	alpha = 0
	addtimer(CALLBACK(src, .proc/reset), 30 SECONDS, TIMER_UNIQUE|TIMER_OVERRIDE)

/obj/effect/forcefield/fog/passable_fog/proc/reset()
	alpha = initial(alpha)
	opacity = TRUE
```

Idea just trying to remove the extra proc there

Hughgent:
The old suggestion did update opacity, but didn't make the fog sprite invisible. 

Using the new suggestion, is working according to the 'design' however I'm noticing a bit of slowdown.

Two changes in this case, using `set_opacity()` and adding an early return shows much less lag locally.

psykzz:
yeah definitely better, i forget about set_opacity and the early returns makes sense. Also means you don't have to add the TIMER_UNIQUE|TIMER_OVERRIDE since you can't trigger it twice

PR #4090, opened 2020-05-20
Queen Screech Buff and Heal Addition (Panther's There too I guess)

SeniorScore:
## About The Pull Request

Rolls Panther changes from a previous PR into the same one with a buff to the queen's screech and addition of a slightly better (range and cooldown-wise, it heals the same) psychic cure a la shrike

## Why It's Good For The Game

Screech was *pretty fucking useless* previously. This hopefully helps it find more usable in the face of multiple marines without changing the stun's length by much ( a full half second if you're next to the queen!) so marines can still return fire but multiple xenos supporting the queen can actually take advantage of a screech. Or all the halloss will make them disengage. Either's a win for how squishy the queen is.

Heal was added to help queen preform in the supposed support role she's supposed to feel. I'd argue previously she was bad at both support and combat short of pheros, and you don't even need to do anything for those to make other xenos good.

## Changelog
:cl:
add: Added longer range heal to Queen (5 range as opposed to 3 of shrike's. Big brain = big range)
add: Adding Panther's sting here too I guess.
balance: Panther slowed and HP minorly increased. Pounce cooldown timer reduces with age. Freeze time after a successful pounce reduced to 1 second instead of 2.
balance: Queen Screech increased to a maximum of 130, minimum of 80 halloss when she screeches. Effectiveness through walls reduced to 20% to compensate for this. Stun increased from 1 to 1.5 at maximum, min same.
/:cl:


psykzz:
I dont like that you merged these into 1.
This makes the PR complicated and hard to review / test the changes independantly.

Also there was feedback on the last PR that wasn't addressed.

PR #4092, opened 2020-05-21
erazorwire

spookydonut:
## About The Pull Request
Turns egrill behaviour into an element and adds it to razorwire

## Why It's Good For The Game
zappy zappy

## Changelog
:cl:
add: razorwire can now be wired up the same way as egrills
/:cl:


AbsoluteJerry:
This is a step into the right direction. There is absolutely nothing that can do wrong.

PR #4103, opened 2020-05-22
Fix medbay access on the pillar

psykzz:
Additionally wires up an APC that was missed


psykzz:
It turns it into a lounge, previously it was apart of requisitions and as such the APC was named incorrectly.

PR #4105, opened 2020-05-23
Annual Map Cleanup

spookydonut:
## About The Pull Request
it's been a year since i last did this and it needs doing.


spookydonut:
lookit those deletions

spookydonut:
yep

PR #4108, opened 2020-05-23
Increase crash nuke timer to 3 minutes

psykzz:
Gives xenos more time to react


jroinc:
Is there any way that you can make nuke activation switch the queen/shrike guidance arrow to the nuke instead of the queen/shrike? I see ayys frantically running around at times trying to tell the leader where it is to try and disarm it, and landmarks aren't always the most helpful.

psykzz:
Yeah i could. Perhaps also make any xeno be able to disarm it.

jroinc:
Oh, one more request.

1-2 minute delay on Canterbury launch, BIG announcement for all marines to get on or else (but not ayys), and the ability to cancel the launch at any time in the launch delay. Insta-launch memes are weird. That might not fit with this so much, though.

psykzz:
I think your right it might be for another PR.

Personally i like the miscommunication and stress from not having an automated annoucement

PR #4128, opened 2020-05-28
Xenomorph Resin Jammer

jjpark-kb:
## About The Pull Request
Adds xenomorph resin jammers.
the jammers affect only marines within a radius of 7, and they must be able to see it.
there can only be one jammer placed at a time.
jammers can only be placed by hivelords.

## Why It's Good For The Game
Adding the jammers will add a certain element of getting caught. You are still able to be found on scanners and stuff, but being unable to call out your location will dissuade you from ever straying too far from the group where they may not know where you are at. 
It even adds a little objective somewhat to those captured to try and find the jammer to destroy it so you can call out your location.

## Changelog
:cl:
add: Added xenomorph jammer
code: changed radio code a little to consider the jammer.
/:cl:


jroinc:
Assuming that the jammer works through walls due to ease of use, I'd lower the radius to 5, or something similar. If it requires LOS, 7's a good number, but you mentioned that being harder.

code/game/objects/structures/xeno.dm
```diff
@@ -695,3 +695,70 @@ TUNNEL
...
+
+/obj/structure/xenojammer/attackby(obj/item/I, mob/user, params)
+	if(!isxeno(user))
+		return ..()
+	attack_alien(user)
+
+/obj/structure/xenojammer/attack_alien(mob/living/carbon/xenomorph/M)
+	if(!istype(M) || M.stat || M.lying_angle)
+		return
+
```
spookydonut, line 760: limit it to resin builder classes so young runner 69 doesnt rdm delete jammers

code/modules/mob/living/carbon/xenomorph/castes/hivelord/abilities_hivelord.dm
```diff
@@ -212,3 +212,73 @@ GLOBAL_LIST_INIT(thickenable_resin, typecacheof(list(
...
+			to_chat(owner, "<span class='warning'>We need an empty claw for this!</span>")
+		return FALSE
+
+/datum/action/xeno_action/xenojammer/action_activate()
+	build_jammer(get_turf(owner))
+
+/datum/action/xeno_action/xenojammer/proc/build_jammer(turf/T)
+	var/mob/living/carbon/xenomorph/X = owner
+	var/mob/living/carbon/xenomorph/blocker = locate() in T
+	if(blocker && blocker != X && blocker.stat != DEAD)
```
spookydonut, line 241: move these to can_use_action

jjpark-kb: I believe I did this, can you check again?

jjpark-kb: okay, I had an issue when doing this, what should I do?

code/game/objects/structures/xeno.dm
jjpark-kb: I dont know exactly how I could go about this.

jjpark-kb:
Just had inspiration on how to write this again and make it hopefully look better.

jjpark-kb:
rather than a structure, it uses a similar thing to hugger traps, which is the effect and stuff. I believe its better.

PR #4141, opened 2020-05-30
Fixes pacman generators not working and adjusts them

MacBlaze1:
## About The Pull Request

This PR fixes Pacmans and makes the UI actually work along with the power usage and stuff.
## Why It's Good For The Game

This should have been fixed a long time ago, as it is not intended behavior and its a feature that was ignored. 
## Changelog
:cl:
fix: made the pacman's UI usable
fix: made the pacman actually eat away at the sheets.
balance: adjusted the pacman to be longer lasting and have more power for use
/:cl:

### DIFF
--- a/code/modules/power/port_gen.dm
+++ b/code/modules/power/port_gen.dm
@@ -90,9 +90,9 @@
 	var/sheet_name = ""
 	var/sheet_path = /obj/item/stack/sheet/mineral/phoron
 	var/sheet_left = 0 // How much is left of the sheet
-	var/time_per_sheet = 260
+	var/time_per_sheet = 580
 	var/current_heat = 0
-
+	power_gen = 15000
 /obj/machinery/power/port_gen/pacman/Initialize()
 	. = ..()
 	if(anchored)
--- a/code/modules/power/port_gen.dm
+++ b/code/modules/power/port_gen.dm
@@ -93,6 +93,7 @@
 	var/time_per_sheet = 580
 	var/current_heat = 0
 	power_gen = 15000
+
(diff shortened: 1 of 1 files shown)


jroinc:
You may want to look at the number of available generators on maps, and keep in mind that shock damage scales with available power in the powergrid to avoid making instagib wires.

MacBlaze1:
> You may want to look at the number of available generators on maps, and keep in mind that shock damage scales with available power in the powergrid to avoid making instagib wires.

Is it total power or power split between sources? example, 120 kW available across 8 erazorwires, 10 kW per, 40 excess power, please tell me what is the shock power.

MacBlaze1:
> You may want to look at the number of available generators on maps, and keep in mind that shock damage scales with available power in the powergrid to avoid making instagib wires.

Just calculated it, its 20.6 damage lmao. The minimum damage is 20, so you’re essentially adding 0.6 damage each time.

code/modules/power/port_gen.dm
```diff
@@ -17,7 +17,7 @@
 	var/base_icon = "portgen0"
 	var/datum/looping_sound/generator/soundloop
 
-	//interaction_flags_atom = INTERACT_ATOM_ATTACK_HAND | INTERACT_ATOM_UI_INTERACT | INTERACT_ATOM_REQUIRES_ANCHORED
+	interaction_flags = INTERACT_MACHINE_TGUI
```
spookydonut, line 20: this should only be set on /obj/machinery/power/port_gen/pacman

PR #4142, opened 2020-05-30
[Map] Lava Outpost

jjpark-kb:
## About The Pull Request
LAVA OUTPOST
A new map that features four new turfs ported and made from /tg/ to make. Lava, Basalt, and the Rock wall were from /tg/, but I made a catwalk that had lava under it.
Each disk and area has their own "pipe network" and "power network". Rather than everything being connected, this can help create new possibilities for both xenos and marines to deal with this.

## Why It's Good For The Game
Adds another map to the rotation to stave off repetition. Involves environmental hazards so it will feel a little different compared to others.

## Changelog
:cl:
add: Added "Lava Outpost" as a map
add: Added 5 turfs: Rock, basalt, glowing basalt, lava, and lava catwalks.
/:cl:


psykzz:
First round with some pop
```
How was lavaland
Great: 0
Good: 0
Okay: 5
Bad: 0
Terrible: 2
Never again: 2
Did not vote: 8
```

Overall there needs to be some railings for marines since shrike was very strong.
But i wouldn't add those everywhere, just in areas that can give marines a stable place to fight on catwalks. (south of red disk is an example)

psykzz:
The SMES near the blue disk isn't wired up correctly 
image

psykzz:
I think most of them are like they, power should come into the SMES and out via the terminal

code/game/turfs/open.dm
```diff
@@ -257,3 +257,87 @@
...
+		else if (isliving(thing))
+			. = 1
+			var/mob/living/L = thing
+			if(!L.on_fire)
+				L.update_fire()
+
+			L.adjustFireLoss(20)
+			if(L) //mobs turning into object corpses could get deleted here.
+				L.adjust_fire_stacks(20)
+				L.IgniteMob()
```
psykzz, line 326: You should also check if the mob is alive to not keep burning dead bodies

jjpark-kb: I believe I have done this.

jjpark-kb:
> The SMES near the blue disk isn't wired up correctly
> image
> I think most of them are like they, power should come into the SMES and out via the terminal

Ack, can I blame the autopower cable thing? Im still not used to it and how to use it. I am used to just placing the wires correctly.

terranaut1:
Crash locations I've seen so far are not good. You're always bottlenecked and sometimes can't even access all of the ships entrances, that needs to be addressed.
I would hold off of adding railings; since Psy mentioned it, south of red is an extremely unnecessary catwalk for marines to even be on - if a Shrike is there at the right time and manages to push somebody into the lava that should be rewarded. Lava does light up the area whereever it is, removing the darkness advantage that xenos have on other maps. There were far less lava deaths in the rounds after (probably because I didn't play shrike ;))))))

Max-023:
only issue i have is none of the turf sprites blend. it looks like a lavaland rip from minecraft. thats all i got tho

psykzz:
For doing it manually its a really good job @jjpark-kb 👍

jjpark-kb:
I will be reworking it to keep the idea, but make it more fleshed out.

jjpark-kb:
Actually, ill just this up. Perhaps a smaller lava would be good, and I can always make it better.

code/game/turfs/open.dm
```diff
@@ -257,3 +257,122 @@
...
+        
+		else if (isliving(thing))
+			var/mob/living/L = thing
+			if(L.stat == DEAD)
+				continue
+	
+			if(!L.on_fire || L.getFireLoss() <= 200)
+				L.adjustFireLoss(20)
+				L.adjust_fire_stacks(20)
+				L.IgniteMob()
```
Hughgent, line 325: This section of the code which does the damage to mobs doesn't respect the mobs armor.

My recommendation is to instead use the proc `take_overall_damage`

`take_overall_damage(brute, burn, blocked = 0, sharp = FALSE, edge = FALSE, updating_health = FALSE)`

On humans this affects limbs at random.

So you would set the arguments to `(0, 20, L.getSoftArmor(fire))` however I'm not certain on precisely how to get the targets armor value at this point in time.

jjpark-kb: I think my only issue with this is, is that I dont want marines with high fire protection to have the ability to walk on the lava. That would make it extremely strong.

Hughgent: Could make it half of the soft armor value then.

jjpark-kb: hmm, true.

psykzz: you could use `clamp` here

jjpark-kb: I believe I have done this.

terranaut1:
Can you remove the banhammer from the secure storage vault? It's just an OOC joke item and shouldn't be on the map imo.

jjpark-kb:
> Can you remove the banhammer from the secure storage vault? It's just an OOC joke item and shouldn't be on the map imo.
> The quad launcher that's in the red disk armory needs to be removed too, and any other death squad weapons if there are any. Some things are way too powerful for regular play.

Those are the random tier weapons, which I dont really want to modify in this PR. I can just remove some of the random spawners.

terranaut1:
Yeah please do that then, random weapons are totally fine but not on this level

Hughgent:
On the Tier weapon spawns, perhaps don't include the tier 4 and 5. 

In the JSON file, if you would please include a line similar to the cold maps about temperature for your map, that might be useful in the future.

PR #4147, opened 2020-05-30
Move LZ2 Crash site on Research outpost

psykzz:
This moves the crash site up 4 tiles, this unblocks all the entrances and moves the ship away from the xeno fog wall to limit the free damage xenos get on the cades.


spookydonut:
You could just delete lz2

psykzz:
Maybe, this is more about the crash site though than LZ2

PR #4155, opened 2020-05-31
Allow Ripley to lift vendors, closets, and tanks

XElectricX:
## About The Pull Request

This PR expands what the Ripley can carry with the claws. Anything that is a child of a closet can be carried, including crates. Anything that is a child of reagent_dispensers can also be carried, like water tanks and beer kegs. Vendors can be carried too. However they all still check for anchored status, so nothing that is not supposed to be moved will be affected, like pepper tanks and weapon vendors. Added a new check for anything that is an object to inform the player if something is not liftable. Also fixed the "Can't grab [loaded]" text. Apparently the X = loaded bit came after the text so it would return a blank. Moving loaded to before the anchored check would result in the item being designated as loaded before checking the anchored state so you could move anything since loaded is used by everything else too.

## Why It's Good For The Game

More stuff for mechs to carry.

## Changelog
:cl:
tweak: Ripley can carry vendors, closets, and reagent tanks.
fix: Redid the "Can't grab" error message, works properly now.
/:cl:


tgstation-server:
Your Fix/Feature pull request delta is now above zero (1). Feel free to make Feature/Tweak/Balance PRs.

PR #4167, opened 2020-06-02
Adds some spawn content to whiskey and fixes some ai nodes

psykzz:
## About The Pull Request

Adds some crates of equipment around WO that are normally added by admin
Fixes a missing AI node
Widens some paths and adds a new pathway into medical.
Increases some pathways were AI tend to get caught near the lower part of map.


jroinc:
Can you make the medbay tunnel area smaller? 2 tiles should be the absolute widest, and one in places would be nice. When I made it, I was going for a cramped but quicker way as opposed to the wide open slightly longer ways around.

Also, add in a single metal wall to the tunnel plz. RP purposes.

spookydonut:
@psykzz mergeconflict

PR #4173, opened 2020-06-04
Turn the carbine into a battle rifle

XElectricX:
## About The Pull Request

Carbine is trash tier and MJP insists all it needs is auto fire even though he is adding another AR that does what the carbine was meant to do. Taking it in the opposite direction and making it a battle rifle seems better as every other gun does a better job at mobile and fast RoF than the carbine does. The new stats put it in between the AR and DMR. 45 damage with 15 AP and 2 sunder. Higher accuracy and less scatter. Bullet travels faster. Now has a 30 bullet magazine. Carbine receives a slower fire rate (still faster than DMR), full auto mode, 2 round burst with a slight accuracy penalty, slower wield delay, and a slower movement speed.

## Why It's Good For The Game

Balancing guns to have roles is better than adding more guns that replace them. The carbine is now relevant.

## Changelog
:cl:
balance: Carbine becomes a battle rifle. Adjusted stats accordingly. Has an autofire setting.
/:cl:


RenGusta:
This is just an inferior version of the battle rifle I've planned to make in Arsenal Expansion. Also doesn't use a new ammo type when I do it, you bloat-nerd.

Pariah919:
this is just the old old spec BR, with all the problems and annoyances it causes like the burst fire just basically being double damage.

Kuro-ai:
I'm against this for two reasons:
1-There is already a Battle rifle underway (sprites are finished)
2-the carbine was a easy to fix and giving it back it's role wasn't hard,MJP's pr adresses the problems with gun better than this one

Chaznoodles:
When will we learn that making a midpoint piece of gear that "does ---- better than ----, but ---- worse than ----" is exactly why nobody uses it

See: medium armour

psykzz:
@XElectricX since its looks like an overwhelming majority are against this PR would you like to close it?

psykzz:
Closing since it seems most people are against this

PR #4177, opened 2020-06-04
Moves a crash marker on Sulaco so it isn't in space

spookydonut:
## Changelog
:cl:
fix: fixed a crash spot on Sulaco that exposed the doors to space
/:cl:


psykzz:
When your looking at the keys your looking at groups of objects.
While that could've happened, I assume instead It created a new group, but removed the old since its now used and instead used a pre-existing key.

spookydonut:
> 
> 
> When your looking at the keys your looking at groups of objects.
> While that could've happened, I assume instead It created a new group, but removed the old since its now used and instead used a pre-existing key.

yep those keys were duplicates

PR #4191, opened 2020-06-09
Add the MarineHQ map

psykzz:
Adds the MarineHQ map so it can be selected.


psykzz:
This was originally for Suno to run an event but has conflicts.
Gonna leave this up and just merge it since we can always disable it via server config

PR #4211, opened 2020-06-12
Buffs barricades 

MacBlaze1:
## About The Pull Request
This PR buffs barricades based on suggestions by others and my own inputs. 
## Why It's Good For The Game
Its not exactly balanced to have an ancient spitter or elder praetorian run up to the cades, stun marines and half health and entire column of cades. This will stop spitters from peeking a corner and just firing off a spit on the cades constantly. Stop being cheese. 
## Changelog
:cl:
add: Added new things
balance: gave barricades 128 coverage which at minimum equates to around a 20% chance to spit beyond the barricade. 
balance: gave metal barricades 40 acid armor, it takes 10 acid spits to take them down with barbed wire (all other barricades have 40 acid armor too, but I didnt calculate them)
balance: made acid sprays stop at barricades and not go through them.
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -383,7 +383,7 @@
 	icon_state = "metal_0"
 	max_integrity = 200
 	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
-	coverage = 20
+	coverage = 100
 	crusher_resistant = TRUE
 	barricade_resistance = 10
 	stack_type = /obj/item/stack/sheet/metal
@@ -629,7 +629,7 @@
 	icon_state = "plasteel_closed_0"
 	max_integrity = 600
 	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
-	coverage = 20
+	coverage = 100
 	crusher_resistant = TRUE
 	barricade_resistance = 20
 	stack_type = /obj/item/stack/sheet/plasteel
@@ -882,7 +882,7 @@
 	barricade_resistance = 15
 	max_integrity = 400
 	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
-	coverage = 20
+	coverage = 100
 	stack_type = /obj/item/stack/sandbags
 	hit_sound = "sound/weapons/genhit.ogg"
 	barricade_type = "sandbag"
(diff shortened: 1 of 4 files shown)


jroinc:
Be aware that adding barbed wire changes the health values, and plasteel has triple the default values.

MacBlaze1:
> Be aware that adding barbed wire changes the health values, and plasteel has triple the default values.

Im completely aware of that, I actively tested metal barricades and made sure I accounted for plasteel and sandbags

Chaznoodles:
Might be worth to have coverage drop based on health thresholds, so the more damaged a barricade is, the less chance it has to block spit.

Hughgent:
Would you please provide the math behind why coverage works the way it does? why 128 means 80% chance of hitting the barricade?

MacBlaze1:
> Would you please provide the math behind why coverage works the way it does? why 128 means 80% chance of hitting the barricade?

okay so first generate_bullet() in projectile.dm is called. It gets the stuff but the important thing is the base accuracy of 85 is added to by the ammo.accuracy of 40. Then that accuracy, which is 125 is randomly multiplied by 92% or 108% because its between (95 - ammo.accuracy_var_low (which is 3), 105 + ammo.accuracy_var_high(which is 3)) * 0.01. This comes out between 115 and 135 accuracy.  then we move on to /obj/structure/projectile_hit() in which we take our coverage which is what the barricades coverage is, if we are facing the opposite way of the barricade the barricades coverage is multiplied by 150% and if not it is multiplied by 50%. Then our final calculation for probability to hit happens, in which our coverage, which is presumably 192, is (subtracted by the spit accuracy (minus the spit accuracy) times (the distance traveled out of the max accurate range) squared. This would look something like 192 - (115 - (115 * ( (4/15) * (4/15) ) ) ) at 4 tiles which comes out to 85% chance to hit the barricade at minimum accuracy. At maximum accuracy at 4 tiles its around 66% chance to hit the barricade.

psykzz:
I think we may just want to invert this check 
```
//Bypass chance calculation. Accuracy over 100 increases the chance of squeezing the bullet past the structure's uncovered areas.
	. -= (proj.accuracy - (proj.accuracy * ( (proj.distance_travelled/proj.ammo.accurate_range)*(proj.distance_travelled/proj.ammo.accurate_range) ) ))
	if(!anchored)
```

This assumes you _want_ to hit the target, whereas with barricades we want to skip it and hit past them.

We should increase the coverage the further you are, and lower it as you get closer.

This likely needs to change from `-=` to `+=`

MacBlaze1:
> I think we may just want to invert this check
> 
> ```
> //Bypass chance calculation. Accuracy over 100 increases the chance of squeezing the bullet past the structure's uncovered areas.
> 	. -= (proj.accuracy - (proj.accuracy * ( (proj.distance_travelled/proj.ammo.accurate_range)*(proj.distance_travelled/proj.ammo.accurate_range) ) ))
> 	if(!anchored)
> ```
> 
> This assumes you _want_ to hit the target, whereas with barricades we want to skip it and hit past them.
> 
> We should increase the coverage the further you are, and lower it as you get closer.
> 
> This likely needs to change from `-=` to `+=`

:uhoh: refactor moment?

psykzz:
Can you explain how its already like that?

MacBlaze1:
> Can you explain how its already like that?, so coverage is the chance to hit the barricade, we subtract 
proj.accuracy - (proj.accuracy * ( (proj.distance_travelled/proj.ammo.accurate_range)*(proj.distance_travelled/proj.ammo.accurate_range) ) ) from the coverage 128 ,it comes out to around 80 coverage to hit the barricade, coverage is increased as you are farther away, and decreases as you are closer

PR #4222, opened 2020-06-15
Updates reagents to /tg/ standard

psykzz:
I got annoyed the recording feature wasn't working on the chem dispenser.

This just updates reagents to the TG standard.
Ports docs, structure and small changes.

This updates a few of the machines too.
Generally balance wise nothing has changed.

## Changelog
:cl:
code: improved reagents and chems to be more like /tg/
/:cl:


psykzz:
Some of the machines could still use the TGUIs ported, but i didn't plan to add them in this PR.

code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -1,1049 +0,0 @@
-/obj/machinery/chem_dispenser
```
spookydonut, line 1: this is already tg code you need to git mv

PR #4224, opened 2020-06-16
Add conducts to modular armor

psykzz:
This basically means you gotta remove the suit to get defibbed.


namjay1170:
how about make it take equivalent to the time it would if you where to remove it? QOL.

PR #4258, opened 2020-06-18
Ports some updates to the asset cache

psykzz:
This is just porting the asset cache from TG


psykzz:
Tested this locally, but this should be Tm'd for a few rounds before merge

PR #4259, opened 2020-06-18
Allows medics to do all surgery

MacBlaze1:
## About The Pull Request
this PR allows medics to all surgery and gives them the gear necessary to not cause an infection every time they try.

## Why It's Good For The Game
We have had a massive spike in population and I think that reducing the burden on doctors would be a good thing.
## Changelog
:cl:
add: added a surgical vest to the medical clothes vendor and made it only hold the necessary surgical stuff
balance: reduced fumbling time for medics from 8 to 4 seconds
balance: reduces IB bleeding by 50% from Wound damage / 40 to wound damage / 60
/:cl:


Pariah919:
I'd rather increase the fail chance for if you don't have the skill to perform a surgery.

MacBlaze1:
> I'd rather increase the fail chance for if you don't have the skill to perform a surgery.
I guess I could remove the checks for some surgeries and do your suggestion, which ones?

code/game/objects/items/storage/surgical_tray.dm
```diff
@@ -3,7 +3,7 @@
 	desc = "A small metallic tray covered in sterile tarp. Intended to store surgical tools in a neat and clean fashion."
 	icon_state = "surgical_tray"
 	flags_atom = CONDUCT
-	w_class = WEIGHT_CLASS_BULKY //Should not fit in backpacks
+	w_class = WEIGHT_CLASS_NORMAL //Should not fit in vests or pouches
```
XElectricX, line 6: Leave the tray as is. Instead, add the surgical vest to the medic vendor and add an additional slot. That way it can fit the whole array of surgical tools.

jroinc:
We generally don't balance here against FORBIDDING you to do something.

Currently, medics can do all surgeries, it's just slower. If you wanna buff their surgery speed, go for that, but don't prevent them from doing others.

code/game/objects/items/storage/surgical_tray.dm
MacBlaze1: done

psykzz:
At this point looking at the changes this does nothing a medic can't already do other than giving them free gear.

I'd rather allow then access to the medbay (if they don't already) where they can choose to get this gear themselves.

XElectricX:
Both would be good. Medbays have limited surgical equipment, best for medics to be able to vend some themselves or else doctors would be left without any.

psykzz:
If there are doctors then the need for corpsmen to do surgery isn't there either.

code/modules/surgery/surgery.dm
```diff
@@ -96,8 +96,8 @@ proc/spread_germs_to_organ(datum/limb/E, mob/living/carbon/human/user)
 		E.germ_level += user.germ_level * 0.33
 
 	//Suits
-	if(user.wear_suit)
-		if(istype(user.wear_suit, /obj/item/clothing/suit/surgical))
+	if(user.wear_suit || user.w_uniform)
+		if(istype(user.wear_suit, /obj/item/clothing/suit/surgical) || istype(user.w_uniform, /obj/item/clothing/under/marine/corpsman))
```
spookydonut, line 100: why?  don't need this

code/modules/surgery/surgery.dm
```diff
@@ -107,7 +107,7 @@ proc/spread_germs_to_organ(datum/limb/E, mob/living/carbon/human/user)
 	if(locate(/obj/structure/bed/roller, E.owner.loc))
 		E.germ_level += 75
 	else if(locate(/obj/structure/table/, E.owner.loc))
-		E.germ_level += 100
+		E.germ_level += 75
```
spookydonut, line 110: rollerbeds should always be better than normal tables since you strap them down.

code/modules/surgery/surgery.dm
```diff
@@ -147,7 +147,7 @@ proc/do_surgery(mob/living/carbon/M, mob/living/user, obj/item/tool)
 				if(locate(/obj/structure/bed/roller, M.loc))
 					multipler -= 0.10
 				else if(locate(/obj/structure/table/, M.loc))
-					multipler -= 0.20
+					multipler -= 0.10
```
spookydonut, line 150: same here

code/modules/clothing/under/ties.dm
```diff
@@ -541,13 +541,14 @@
 	hold = /obj/item/storage/internal/tie/white_vest
 
 /obj/item/storage/internal/tie/white_vest
-	storage_slots = 8
+	storage_slots = 1
```
spookydonut, line 544: this is dumb just load the tools into the vest like it is currently

MacBlaze1: I was told to do it like this reeeeee

code/modules/mob/living/blood.dm
```diff
@@ -103,7 +103,7 @@
 			for(var/w in temp.wounds)
 				var/datum/wound/W = w
 				if(W.bleeding())
-					blood_max += (W.damage / 40)
```
psykzz, line 106: Whats the context on 40 & 60 here?
feels like a magic number right now (which we want to avoid)

MacBlaze1: W.damage is the damage of the wound, so if you have 40 damage on your internal bleeding wound, it adds 1 blood loss to blood_max, 60 is a higher threshold and 50% higher

code/modules/surgery/surgery.dm
```diff
@@ -120,7 +120,7 @@ proc/do_surgery(mob/living/carbon/M, mob/living/user, obj/item/tool)
 	if(user.skills.getRating("surgery") < SKILL_SURGERY_PROFESSIONAL)
 		user.visible_message("<span class='notice'>[user] fumbles around figuring out how to operate [M].</span>",
 		"<span class='notice'>You fumble around figuring out how to operate [M].</span>")
-		var/fumbling_time = SKILL_TASK_FORMIDABLE - ( 6 SECONDS * user.skills.getRating("surgery") ) // 20 secs non-trained, 14 amateur, 8 trained, 2 prof
+		var/fumbling_time = max(1,SKILL_TASK_FORMIDABLE - ( 8 SECONDS * user.skills.getRating("surgery") )) // 20 secs non-trained, 14 amateur, 8 trained, 2 prof
```
psykzz, line 123: Max isn't really needed is it?
You could just chaange the check below to `if(fumbling_time && !do_after(user, fumbling_time, TRUE, M, BUSY_ICON_UNSKILLED))`

MacBlaze1: The math I think can go below zero

psykzz: Then perhaps max it to 0 so you can use it as a falsy check. otherwise your doing a 1 decisecond wait, which is wasteful (since you create a progress bar etc, only to remove it the next tick).

code/modules/surgery/surgery.dm
```diff
@@ -120,7 +120,7 @@ proc/do_surgery(mob/living/carbon/M, mob/living/user, obj/item/tool)
 	if(user.skills.getRating("surgery") < SKILL_SURGERY_PROFESSIONAL)
 		user.visible_message("<span class='notice'>[user] fumbles around figuring out how to operate [M].</span>",
 		"<span class='notice'>You fumble around figuring out how to operate [M].</span>")
-		var/fumbling_time = SKILL_TASK_FORMIDABLE - ( 6 SECONDS * user.skills.getRating("surgery") ) // 20 secs non-trained, 14 amateur, 8 trained, 2 prof
+		var/fumbling_time = max(0,SKILL_TASK_FORMIDABLE - ( 8 SECONDS * user.skills.getRating("surgery") )) // 20 secs non-trained, 12 amateur, 4 trained, 0 prof
 		if(!do_after(user, fumbling_time, TRUE, M, BUSY_ICON_UNSKILLED))
```
psykzz, line 124:
```suggestion
		if(fumbling_time && !do_after(user, fumbling_time, TRUE, M, BUSY_ICON_UNSKILLED))
```

PR #4270, opened 2020-06-20
Adds directional slashing 

MacBlaze1:
## About The Pull Request

this PR adds directional attacking in which if you attack in the direction of a mob in melee range, you should damage it with a melee attack
## Why It's Good For The Game
This is a highly requested feature by more active members and it should help make balance better
## Changelog
:cl:
add: Added directional attacking component to xenomorphs
/:cl:


code/datums/actions/directional_attack_toggle.dm
```diff
@@ -0,0 +1,20 @@
...
+
+/datum/action/directional_attack_toggle/update_button_icon(active)
+	if(isnull(active))
+		return
+	if(active)
+		button.vis_contents -= inactive_icon
+		button.vis_contents += active_icon
+	else
+		button.vis_contents -= active_icon
+		button.vis_contents += inactive_icon
```
spookydonut, line 20: fix your editor

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,94 @@
...
+	if(attacker.issamexenohive(L))
+		return //No more nibbling.
+	return living_do_directional_action(L)
+
+
+/datum/component/directional_attack/proc/living_do_directional_action(mob/living/L)
+	var/mob/living/attacker = parent
+	attacker.UnarmedAttack(L, TRUE)
+	COOLDOWN_START(src, COOLDOWN_DIRECTIONAL_ATTACK, CLICK_CD_MELEE)
+
```
spookydonut, line 94: fix your editor

code/datums/actions/directional_attack_toggle.dm
```diff
@@ -0,0 +1,20 @@
+//Toggle directional attacks
+/datum/action/directional_attack_toggle
+	name = "Toggle Directional Attacks"
+	var/static/atom/movable/vis_obj/action/bump_attack_active/active_icon = new
+	var/static/atom/movable/vis_obj/action/bump_attack_inactive/inactive_icon = new
+
```
Hughgent, line 6: I feel like this, being integral to this component, can be in the directional attack component file. If only to have things be slightly easier to port as it's a single file.

Thoughts, other maintainers?

psykzz:
Directional attack doesn't work on lying mobs

MacBlaze1:
fixed the runtimes

code/_onclick/click.dm
```diff
@@ -19,6 +19,7 @@
 /atom/Click(location, control, params)
 	if(flags_atom & INITIALIZED)
 		SEND_SIGNAL(src, COMSIG_CLICK, location, control, params, usr)
+		SEND_SIGNAL(usr, COMSIG_CLICK, location, control, params, usr)
```
psykzz, line 22: There is already `COMSIG_MOB_CLICKON` trigged by the `usr.ClickOn`

psykzz:
With directional attacks on, couldn't destroy windows

MacBlaze1:
> With directional attacks on, I couldn't destroy windows

not designed to, and for your other point,  this pr did not work until I had added that call

code/_onclick/click.dm
MacBlaze1: necessary

code/_onclick/click.dm
```diff
@@ -19,6 +19,7 @@
 /atom/Click(location, control, params)
 	if(flags_atom & INITIALIZED)
 		SEND_SIGNAL(src, COMSIG_CLICK, location, control, params, usr)
+		SEND_SIGNAL(usr, COMSIG_DIRECT_ATTACK, location, control, params, src, usr)
```
psykzz, line 22: Why do you have a new signal? How is this different from `COMSIG_CLICK`

MacBlaze1: I was told to do this by Hughgent

code/_onclick/click.dm
```diff
@@ -137,6 +138,10 @@
 		if(W)
 			W.melee_attack_chain(src, A, params)
 		else
+			GetComponent(/datum/component/directional_attack)
+			var/datum/component/directional_attack/check = src?.GetComponent()
+			if(check && check.active)
+				return
```
psykzz, line 144: Whats going on here? `GetComponent` requires an argument, im pretty sure will never cancel early.

```suggestion
			GetComponent(/datum/component/directional_attack)
			var/datum/component/directional_attack/check = src?.GetComponent()
			if(check && check.active)
				return
```

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,154 @@
...
+	. = living_directional_action_checks(L)
+	if(!isnull(.))
+		return
+	if(QDELETED(L))
+		return
+	switch(attacker.a_intent)
+		if(INTENT_HELP, INTENT_GRAB)
+			return NONE
+	if(attacker.get_active_held_item())
+		return NONE //We have something in our hand.
```
psykzz, line 57: Move this to living check since living has hands as well and you wont need to duplicate later.
```suggestion
```

psykzz:
@MacBlaze1 make sure to actually add the living check, not just accept the suggestion.

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,155 @@
...
+
+/datum/component/directional_attack/Destroy(force, silent)
+    QDEL_NULL(toggle_action)
+    return ..()
+
+/datum/component/directional_attack/proc/select_directional_action(datum/source, atom/A, params)
+	if(ishuman(parent))
+		human_directional_action(A, params)
+	else if(isxeno(parent))
+		xeno_directional_action(A, params)
```
Hughgent, line 26: To cut down one check. please make the `isxeno()` check be the first `if`.

Considering you are only adding it to xenos at this point in time.

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,155 @@
+/datum/component/directional_attack
+	var/active = TRUE
+	var/datum/action/directional_attack_toggle/toggle_action
+
+/datum/component/directional_attack/Initialize()
+	. = ..()
+	if(!isliving(parent))
+		return COMPONENT_INCOMPATIBLE
+	toggle_action = new()
+	var/toggle_path
```
psykzz, line 10: Theres no reason to add this var

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,155 @@
...
+	if(attacker.get_active_held_item())
+		return NONE //We have something in our hand.
+
+/datum/component/directional_attack/proc/carbon_directional_action_checks(mob/living/L)
+	var/mob/living/carbon/attacker = parent
+	. = living_directional_action_checks(L)
+	if(!isnull(.))
+		return
+	if(QDELETED(L))
+		return
```
psykzz, line 55:
```suggestion
		return NONE
```
This should also be in the living check not just for carbons

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,155 @@
...
+	if(presumedPos == target && L == null)
+		return target
+	return L
+
+/datum/component/directional_attack/proc/living_directional_action(atom/A, params)
+	var/atom/T
+	if(isturf(A))
+		T = A
+	else 
+		return
```
psykzz, line 73: Update this to use an early return instead

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,155 @@
...
+	if(isturf(A))
+		T = A
+	else 
+		return 
+	var/atom/hold = figure_out_living_target(T)
+	if (isnull(hold))
+		return
+	if(isturf(hold))
+		hold = A
+	if(isliving(hold))
```
psykzz, line 79: This could be an early return too

code/datums/components/directional_attack.dm
```diff
@@ -0,0 +1,155 @@
...
+
+/datum/action/directional_attack_toggle/update_button_icon(active)
+	if(isnull(active))
+		return
+	if(active)
+		button.vis_contents -= inactive_icon
+		button.vis_contents += active_icon
+	else
+		button.vis_contents -= active_icon
+		button.vis_contents += inactive_icon
```
psykzz, line 155: We should avoid having this as another button.
I'd rather instead this was moved to a verb instead.

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -55,7 +55,7 @@
 
 	RegisterSignal(src, COMSIG_GRAB_SELF_ATTACK, .proc/devour_grabbed) //Devour ability.
 
-	AddComponent(/datum/component/bump_attack)
+	AddComponent(/datum/component/directional_attack)
```
psykzz, line 58: I still think we should keep bump, and just add directional to it.

code/_onclick/click.dm
```diff
@@ -137,6 +137,9 @@
 		if(W)
 			W.melee_attack_chain(src, A, params)
 		else
+			var/datum/component/directional_attack/check = GetComponent(/datum/component/directional_attack)
```
spookydonut, line 140: this needs to be signaled rather than using get component

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
spookydonut: disagree, directional should replace bump

spookydonut: balance is kinda bad with bump still in, directional is less severe a factor compared to bump

psykzz: I don't see bump as a balance concern though, just another way to "click" someone. WASD is easier than clicking for sure, but bump doesn't give them more damage or free hits, it just makes it that when a xeno does get in range to attack they are more likely too because controls are not inhibiting them.

Hughgent:
Do the things about the 'button' and the 'early return'

Hughgent:
TM for like a week. Get feedback about whether or not to keep bump, keep directional, or keep both active.

code/datums/components/directional_attack.dm
MacBlaze1: your suggestion did not work when I tested it

code/datums/components/directional_attack.dm
MacBlaze1: It is now 1 button for switching between

Hathkar:
This does something funky on the new directional slash: You can click things in cardinal directions. For example, a light is north of you, you can't break it unless you're diagonal.

MacBlaze1:
> This does something funky on the new directional slash: You can click things in cardinal directions. For example, a light is north of you, you can't break it unless you're diagonal.

fixed it

MiniMeatwad:
Can't click on people to deal damage normally.

MacBlaze1:
> Can't click on people to deal damage normally.

could you elaborate

MiniMeatwad:
Also, you can slash people through vents or pipes by shift-clicking or something. It seems it replaces normal clicking with a hacky implementation with no sanity checks, and it doesn't play nice with anything except slashing.

Wayland-Smithy:
The setting seems to reset with each evolution/mob (this probably also happened before, but with bump on as the default I would not have noticed) and the UI button always stays green when in both modes. Keeping both attack modes in and not just replacing bump attacks is much appreciated, thanks.

Wayland-Smithy:
There seems to be a problem with diagonal reachability checks as well as the attacking from vents thing. Tested locally with the latest commit at time of recording:
RbllsIQX2d

MacBlaze1:
is there a is_in_vent check or something, lemme look

MacBlaze1:
yeah fixed vent crawling attacking

MacBlaze1:
fixed diagonal attacking on local

Wayland-Smithy:
Clearing snow is behaving oddly when in the new attack mode:
- There is much less delay between grabs at the snow
- Tiles with snow and a body can not be cleared

MacBlaze1:
fixed the dragging, grabbing, disarming and middle clicking, probably other shit

Wayland-Smithy:
When trying to plant an egg on a tile other than the one we are standing on:
image
Clicking your own tile or using the egg in hand still let's you get it planted. Tested with the latest commit.

MacBlaze1:
> When trying to plant an egg on a tile other than the one we are standing on:
> image
> Clicking your own tile or using the egg in hand still gets it planted. Tested with the latest commit.

Ah lemme take a look

MacBlaze1:
@Wayland-Smithy Fixed egg planting

Texicos:
considering this TM changes shit that it wasn't even prescribed to change now, can we not merge it?

MiniMeatwad:
Seems to add the slash delay to abilities. If you use an ability, you cannot slash as if using an ability is slashing. This applies to all abilities, so you can't time abilities while you're slashing or anything.
Still has a problem where you can slash, use an ability, and then spam slash to slash twice as fast.

PR #4273, opened 2020-06-20
Add crash nuke cinematic

psykzz:
News a new crash cinematic


psykzz:
https://streamable.com/v8im6m

PR #4023, opened 2020-05-09
Disables ship explosions during crashes [TEST MERGE]

psykzz:
Explosions are currently killing marines when they land, this disables that for the time being.

PR #4287, opened 2020-06-22
Fixes ERT loading issues

spookydonut:
## About The Pull Request
Closes #4266
Closes #4028 
Closes #4023 

ERT shuttles now load properly, load and spawn directly in the transit level.

Cant now loads direct into transit level

Cant/dropship crash now uses /turf/empty() instead of explosions

removes atmos pipes from cant due to them being unneeded and causing pages of runtimes.

massively shrinks admin/centcomm map as it was only large for loading landmarks/docking ports, with those no longer needed it can be taken back down to the bare essentials.


code/modules/shuttle/gamemodes/crash.dm
```diff
@@ -40,8 +40,13 @@
...
 
 
+	// just wipe the turfs
+	for(var/turf/T in range(3, rear)+range(3, left)+range(3, right)+range(3, front)+range(4, front_right)+range(4, front_left)+range(4, rear_right)+range(4, rear_left))
+		T.empty(/turf/open/floor/plating)
+
 	// Big explosions
-	explosion(front, 3, 4, 7, 0)
+
+/*	explosion(front, 3, 4, 7, 0)
```
psykzz, line 49: Replace the explosions with a smoke effect?
Would be harmless, but still give the same type of effect

spookydonut: sure but it'd need to be after landing not before

psykzz: That'd be fine

spookydonut: outside the scope of this pr :P

PR #4290, opened 2020-06-22
brings the construction skill in line with other time reducing skills

MacBlaze1:
## About The Pull Request
This PR brings the construction skill in line with other time reducing skills by making extra skill reduce construction time
## Why It's Good For The Game

The construction skill was essentially static compared, with it only being used as requirements, like other skills, but not as time reduction, like many many other skills. This changes that so the CE is better than an average engineer, and so are ship technicians. Should provide an extra reason to play them and create an interesting dynamic.

## Changelog
:cl:
tweak: tweaked the construction skill to it so if you have more skill than needed to construct an object, it constructs 10% faster for each extra skill point
:cl:

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -162,6 +162,8 @@
 		var/building_time = R.time
 		if(R.skill_req && usr.skills.getRating("construction") < R.skill_req)
 			building_time += R.time * ( R.skill_req - usr.skills.getRating("construction") ) * 0.5 // +50% time each skill point lacking.
+		if(R.skill_req && usr.skills.getRating("construction") > R.skill_req)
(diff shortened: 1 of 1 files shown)


Hughgent:
Isn't the point of construction the DECREASE in construction time as included in the couple of lines of code above your addition?

InterroLouis:
But don't the of lines of code above their addition relate to the INCREASING of construction time?

Hughgent:
Right, line breaks are funky on the website. The lines above say, if you are not skilled enough, increase the time.

Considering that, please include some logic such that, if you are skilled enough, drop the time, but if you aren't skilled enough do not apply any bonus from your construction skill.

MacBlaze1:
> Right, line breaks are funky on the website. The lines above say, if you are not skilled enough, increase the time.
> 
> Considering that, please include some logic such that, if you are skilled enough, drop the time, but if you aren't skilled enough do not apply any bonus from your construction skill.

thats literally how it is, and if you arent skilled enough time is increased by 50%

PR #4303, opened 2020-06-24
cuts point-to cooldown by 50%

MacBlaze1:
## About The Pull Request

Ive found it hard to demonstrate effectively what I a fellow marine to do or build with such a long timer.
## Why It's Good For The Game

Its so tedious to have to click and wait 5 seconds only for the previous pointer to be gone. I feel I cant effectively illustrate a line with clicks.

## Changelog
:cl:

tweak: tweaked point_to cooldown, its now 2.5 seconds
/:cl:


TiviPlus:
if it was up to me I'd make it even lower but this works

MacBlaze1:
> if it was up to me I'd make it even lower but this works

I would if I dared, but point spam is dumb and Id rather not have that happen. 2.5 I believe is long enough not to spam, yet short enough to get your (NOT INTENDED THE FUNNY) point across.

MacBlaze1:
broke and screwed

PR #4305, opened 2020-06-24
Experimental mob processing smoothing

spookydonut:
## About The Pull Request
This smooths out mob processing load, in theory.  Will need some load testing


spookydonut:
just simple testing of spawning 100 humans on one tile then throwing a fire grenade next to them, 223% tick allocation usage, with the smoothing its < 50% usage

PR #4312, opened 2020-06-25
[Tweak] Necessary Acid Well Changes

jjpark-kb:
## About The Pull Request
adds a range limit to acid wells
adds the cooldown back to them: 30 seconds
makes walking over acid wells deal acid damage (burn)
adds one second delay to gas deployment

## Why It's Good For The Game
tweaking the acid well to be friendlier and not so oppressive.

## Changelog
:cl:
tweak: tweaked cooldown to 30 seconds from nonexistent
tweak: tweaked instant deployment to 1 second gas deployment
tweak: tweaked acid wells to dealing acid damage instead of toxin damage
add: added a range limit to acid wells
/:cl:

### DIFF
--- a/code/game/objects/structures/xeno.dm
+++ b/code/game/objects/structures/xeno.dm
@@ -735,6 +735,7 @@ TUNNEL
 	var/charges = 1
 	var/ccharging = FALSE
 	var/mob/living/carbon/xenomorph/creator = null
+	var/turf/gasturf = null
 
 /obj/effect/alien/resin/acidwell/Initialize()
 	. = ..()
@@ -745,10 +746,13 @@ TUNNEL
 		var/area/A = get_area(src)
 		if(A)
 			to_chat(creator, "<span class='xenoannounce'>You sense your acid well at [A.name] has been destroyed!</span>")
-	var/datum/effect_system/smoke_spread/xeno/acid/A = new(get_turf(src))
-	A.set_up(clamp(charges,0,2),src)
-	A.start()
+	addtimer(CALLBACK(src,.proc/gasdestroy), 1 SECONDS)
 	return ..()
+
+/obj/effect/alien/resin/acidwell/proc/gasdestroy()
+	var/datum/effect_system/smoke_spread/xeno/acid/A = new(gasturf)
+	A.set_up(clamp(charges,0,2),gasturf)
+	A.start()
 
 /obj/effect/alien/resin/acidwell/examine(mob/user)
 	..()
@@ -829,6 +833,7 @@ TUNNEL
 		if(!charges)
 			return
 		C.adjustToxLoss(charges * 15)
+		C.apply_damage(20*charges,BURN,blocked=clamp(C.getarmor("acid"),0,50))
 		charges = 0
 		update_icon()
 		return
--- a/code/game/objects/structures/xeno.dm
+++ b/code/game/objects/structures/xeno.dm
@@ -832,7 +832,7 @@ TUNNEL
 	else
 		if(!charges)
 			return
-		var/armor_blocked = clamp(C.getarmor("acid"), 0, 50)
+		var/armor_blocked = clamp(C.getarmor(null, "acid"), 0, 50)
 		C.apply_damage(20 * charges, BURN, blocked = armor_blocked)
 		charges = 0
 		update_icon()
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/castes/shrike/abilities_shrike.dm
```diff
@@ -401,12 +401,19 @@
 			to_chat(owner, "<span class='warning'>There is a resin node in the way!</span>")
 		return FALSE
 
+	if(locate(/obj/effect/alien/resin/acidwell) in range(1,T))
```
Hughgent, line 404: You may consider using the following instead of 1


"A Dist of 0 includes Center, the contents of Center (normally usr.contents), its location (normally the turf a mob is standing on), and any other contents of that location. A value of 1 extends the region to the neighboring squares on the map and so on. You can also use a rectangular box size using a text string such as "13x11". Both arguments are optional and may be passed in any order"

as the `1` makes a + shape.

Hughgent: See if "3x3" works. if you don't want traps set up diagonally from each other.

jjpark-kb: So yea, I tested the range(1,T) and it includes diagonals.

code/game/objects/structures/xeno.dm
```diff
@@ -745,11 +746,14 @@ TUNNEL
 		var/area/A = get_area(src)
 		if(A)
 			to_chat(creator, "<span class='xenoannounce'>You sense your acid well at [A.name] has been destroyed!</span>")
-	var/datum/effect_system/smoke_spread/xeno/acid/A = new(get_turf(src))
-	A.set_up(clamp(charges,0,2),src)
-	A.start()
+	addtimer(CALLBACK(src,.proc/gasdestroy), 1 SECONDS)
```
psykzz, line 749:
```suggestion
	addtimer(CALLBACK(src,.proc/release_smoke), 1 SECONDS)
```
Try to use `_` more in your naming.
Makes them clearer and easier to read.

code/game/objects/structures/xeno.dm
```diff
@@ -745,11 +746,14 @@ TUNNEL
...
 	return ..()
 
+/obj/effect/alien/resin/acidwell/proc/gasdestroy()
```
psykzz, line 752:
```suggestion
/obj/effect/alien/resin/acidwell/proc/release_smoke()
```

code/game/objects/structures/xeno.dm
```diff
@@ -829,6 +833,7 @@ TUNNEL
 		if(!charges)
 			return
 		C.adjustToxLoss(charges * 15)
+		C.apply_damage(20*charges,BURN,blocked=clamp(C.getarmor("acid"),0,50))
```
psykzz, line 836: Spacing here, try to keep spacees after commas etc
```suggestion
		var/armor_blocked = clamp(C.getarmor("acid"), 0, 50)
		C.apply_damage(20 * charges, BURN, blocked = armor_blocked)
```

code/game/objects/structures/xeno.dm
```diff
@@ -745,11 +746,14 @@ TUNNEL
 		var/area/A = get_area(src)
 		if(A)
 			to_chat(creator, "<span class='xenoannounce'>You sense your acid well at [A.name] has been destroyed!</span>")
-	var/datum/effect_system/smoke_spread/xeno/acid/A = new(get_turf(src))
-	A.set_up(clamp(charges,0,2),src)
-	A.start()
+	addtimer(CALLBACK(src,.proc/release_smoke), 1 SECONDS)
```
spookydonut, line 749: this wont work because the timer will be deleted in the parent destroy

InterroLouis:
This doesn't even remove the toxin damage, it just adds burn damage that's checked against acid armor. That plus, the armor check used doesn't function so this is just...adding 100 burn to the 75 tox of a full acid well.

code/game/objects/structures/xeno.dm
jjpark-kb: I just tested it, and it seems to work.

jjpark-kb:
Im getting no updates and there isnt much I can do. Im closing this because I dont really want to deal with it anymore.
It has do not merge and do not testmerge, what else can I do.

PR #4316, opened 2020-06-25
Defender Bugfix

InterroLouis:
## About The Pull Request

Basically this fixes a bug with defenders where them lowering and then raising their crest sets their bomb defense to 0. The same thing happens with fortify. Normal bomb defense for them is xeno_bomb_resist 2, so I'm setting it to return them to that when they deactivate fortify or crest defense. 

## Why It's Good For The Game

There's a bug. I fix bug. Bug is gone. 

## Changelog
:cl:
fix: Defender Explosion Armor Fixed
/:cl:


psykzz:
If the value doesn't change between toggles why are we even setting it
though good find 👍

tgstation-server:
Your Fix/Feature pull request delta is now above zero (0). Feel free to make Feature/Tweak/Balance PRs.

PR #4321, opened 2020-06-25
Fix mini sentry starting with less health

XElectricX:
## About The Pull Request

Title.

## Why It's Good For The Game

Bug fix.

## Changelog
:cl:
fix: Mini sentry started out with less health due to the folded state's max integrity not being raised. It now matches the deployed health.
/:cl:

### DIFF
--- a/code/game/objects/machinery/sentries.dm
+++ b/code/game/objects/machinery/sentries.dm
@@ -1265,7 +1265,7 @@
 	icon_state = "minisentry_packed"
 	item_state = "minisentry_packed"
 	w_class = WEIGHT_CLASS_BULKY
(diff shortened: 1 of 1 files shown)


code/game/objects/machinery/sentries.dm
```diff
@@ -1265,7 +1265,7 @@
 	icon_state = "minisentry_packed"
 	item_state = "minisentry_packed"
 	w_class = WEIGHT_CLASS_BULKY
-	max_integrity = 155 //We keep track of this when folding up the sentry.
+	max_integrity = 200 //We keep track of this when folding up the sentry.
```
psykzz, line 1268: Instead of just increasing its health you should correctly set its `obj_integrity` to 155.

PR #4323, opened 2020-06-25
reduces point to cooldown by 50%

MacBlaze1:
## About The Pull Request

This PR reduces point to cooldown by 50%
## Why It's Good For The Game

too hard to create understanding with pointing

## Changelog
:cl:

tweak: point to cooldown is now 2.5 seconds instead of 5

/:cl:


spookydonut:
if i remember correctly its this long for balance/exploit reasons to point to things that are hidden, but i cant remember if i made it so you can't do like /point-to "hidden thing"

PR #4331, opened 2020-06-26
Make the resin door faster in opening.

TheLordME:
## About The Pull Request
Decreased the Delay until a resin door opens, allowing Xenos faster movement through their 'hive'.

## Why It's Good For The Game
Xenos rarely use the resin doors because they take to long to open and marines are only merely more slowed down. The code changes should make them faster. (actually @Hughgent came up with the "changes") I will try to provide a better sprite soon as the current sprite is now way to slow.
An decrease of the time to close is probably not needed.

## Changelog
:cl:
balance: Decreased resin-door-open-delay from 10 to 7 frames
/:cl:


TheLordME:
Changes are not doing what was expected

PR #4333, opened 2020-06-26
Reduce resin door opening relay

TheLordME:
Reduces the sleep between initiating the opening and the actual opening from 10 to 7

## About The Pull Request
Reduces the sleep between initiating the opening and the actual opening from 10 to 7.

## Why It's Good For The Game
Resin doors are rarely used as they slow xenos as much as they slow marines.
This should reduce the xeno slowdown a little bit.

## Changelog
:cl:
balance: Decreases resin door opening delay from 10 to 7 frames.
/:cl:

### DIFF
--- a/code/game/objects/structures/xeno.dm
+++ b/code/game/objects/structures/xeno.dm
@@ -290,7 +290,7 @@
 	isSwitchingStates = TRUE
 	playsound(loc, "alien_resin_move", 25)
 	flick("[mineralType]opening",src)
-	sleep(10)
+	sleep(7)
 	density = FALSE
 	opacity = FALSE
 	state = 1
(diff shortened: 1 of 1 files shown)


neohenry:
I will check if the resin doors will be used more when I play queen; concurrently, resin doors are useful for captures _assuming_ you don't spam tackles.

code/game/objects/structures/xeno.dm
```diff
@@ -290,7 +290,7 @@
 	isSwitchingStates = TRUE
 	playsound(loc, "alien_resin_move", 25)
 	flick("[mineralType]opening",src)
-	sleep(10)
+	sleep(7)
```
psykzz, line 293:
```suggestion
	sleep(0.7 SECONDS)
```

PR #4335, opened 2020-06-27
CPR the dead

Rohesie:
## About The Pull Request

* Gives marines the ability to more quickly identify those long gone without tools, and help those still warm.
* Fixes an issue with the death balancing tool, sending the wrong global signal.
* Cleans up death code a bit, and makes the cleanup respond more properly to the event of the change of the `stat` value.

## Why It's Good For The Game

* Making the medical game a bit more interesting. I'd increase the revive timer, personally, from 5 to 10 minutes or so. Didn't do so in this PR because I don't want to include big balance changes.

## Changelog
:cl:
add: Marines can CPR defibable bodies every 7 seconds, buying them some time.
/:cl:

### DIFF
--- a/code/game/objects/machinery/cryopod.dm
+++ b/code/game/objects/machinery/cryopod.dm
@@ -451,7 +451,7 @@
 	user.despawn(src)
 
 
-/obj/machinery/cryopod/proc/go_out()
+/obj/machinery/cryopod/proc/go_out(mob/source, gibbing)
 	if(QDELETED(occupant))
 		return
 
--- a/code/game/objects/machinery/cryopod.dm
+++ b/code/game/objects/machinery/cryopod.dm
@@ -451,7 +451,7 @@
 	user.despawn(src)
 
 
-/obj/machinery/cryopod/proc/go_out(mob/source, gibbing)
+/obj/machinery/cryopod/proc/go_out()
 	if(QDELETED(occupant))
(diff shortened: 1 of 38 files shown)


code/game/objects/machinery/cryopod.dm
```diff
@@ -451,7 +451,7 @@
 	user.despawn(src)
 
 
-/obj/machinery/cryopod/proc/go_out()
+/obj/machinery/cryopod/proc/go_out(mob/source, gibbing)
```
spookydonut, line 454: accidental change? this doesnt do anything far as i can tell

code/modules/mob/death.dm
```diff
@@ -21,10 +20,9 @@
 //Originally created for wizard disintegrate. I've removed the virus code since it's irrelevant here.
 //Dusting robots does not eject the MMI, so it's a bit more powerful than gib() /N
 /mob/proc/dust()
-	death(1)
 	dust_animation()
 	spawn_dust_remains()
-	qdel(src)
+	death(TRUE)
```
spookydonut, line 25: good change!

code/game/objects/machinery/cryopod.dm
Rohesie: Yep, looks like a mistake.

PR #4350, opened 2020-06-27
FOB Drone Construction Speed

InterroLouis:
## About The Pull Request

Basically this makes the FOB Drone build barricades, plasteel barricades, and sentries faster. Does not change material cost of either cade types. 

## Why It's Good For The Game

Less rush for whoever is manning the FOB drone, while the ungas all shout about how they want to deploy. 

## Changelog
:cl:
balance: FOB Drone Builds Faster
/:cl:

### DIFF
--- a/code/modules/mining/remote_fob/remote_fob_actions_misc.dm
+++ b/code/modules/mining/remote_fob/remote_fob_actions_misc.dm
@@ -58,7 +58,7 @@
 			continue
 		to_chat(owner, "<span class='warning'>No space here for a barricade.</span>")
 		return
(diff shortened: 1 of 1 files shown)


psykzz:
Honestly since this is building in vaccum, there is no reason its not instant other than accidental builds.
I'd put everything down to 1sec just so you still have a chance to cancel it.

Alternatively we could rewokr this into bots (think factorio) and have the FOB drone just deploy "blueprints" that get auto built by drones.

This would keep the RP and allow the console to stay instant.

Those drones could be used too as ways to keep an FOB secure or something.

tgstation-server:

PR #4358, opened 2020-06-28
Readds code for setting playercount limits on map voting

spookydonut:
## About The Pull Request
Provisional config values added from player consultation, final values up to keyholders but I feel these suggested numbers are fairly generous.
Admins can still force maps without restriction.

## Why It's Good For The Game
Stops players voting for maps that don't work well with the wrong number of players

## Changelog
:cl:
config: per-map player limits re-added for end of round map voting
/:cl:


XElectricX:
Let players vote for what they want. Setting restrictions is dumb.

spookydonut:
> 
> 
> Let players vote for what they want. Setting restrictions is dumb.

a lot of maps really don't work well with some player counts and all players suffer.

PR #4363, opened 2020-06-29
Fixes turf being deleted with rangefinders

MacBlaze1:
## About The Pull Request

This PR fixes a horrible no good very bad bug that deletes turfs after they are selected by the rangefinder, nothing can be called on them after this

## Why It's Good For The Game

Christ I spent so much time figuring this out and its really bad for gameplay
## Changelog
:cl:
fix: fixed rangefinders literally deleting turfs after selecting a target coordinate 

/:cl:


MacBlaze1:
actually I cant just null the coord let me open a different pr

PR #4374, opened 2020-06-30
Gliding fix/tweak to make it actually work

TiviPlus:
## About The Pull Request

Mobs and so would bob around, especially noticable at higher FPSes
This was because their glide was being multiplied by 0 to always be 0

Might need a round of TMs to see if the number needs to be tweaked
## Why It's Good For The Game

ITS GOOD WHEN MOVEMENT WORKS YES

## Changelog
:cl:
fix: Moving things bobbing/teleporting, especially at higher FPSes, will no longer do so
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -180,7 +180,7 @@ Administrative related.
 	min_val = 0
 
 /datum/config_entry/number/glide_size_mod
-	config_entry_value = 0
+	config_entry_value = 80
 /*
 Voting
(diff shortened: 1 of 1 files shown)


psykzz:
I'll double-check if the server has this set, otherwise, this won't do anything.

psykzz:
setting this on server, seems fine but ghost orbits are still jarring.

Kuro-ai:
this feels very slow and weird,it's terrible for the flow of the game,it needs tweaking

PR #4330, opened 2020-06-26
Tarp Uncovering

jjpark-kb:
## About The Pull Request
Makes tarps get uncovered if a xeno gets too close and stays too close for too long.

## Why It's Good For The Game
adds some counterplay to tarps.

## Changelog
:cl:
add: tarps will uncover if a xeno stands too close for too long
/:cl:

PR #4381, opened 2020-07-01
Reverts tarp uncovering jj

MacBlaze1:
## About The Pull Request

This PR attempts to fix the breaking of tarps that the two tarp changes pr's caused.
I have tested the effects without this PR, and it looks that psykzz's PR is working fine
## Why It's Good For The Game

Tarps are utterly useless due to their broken and glitchy state, I dont believe this was the intent of either of the PRs 

## Changelog
:cl:

del: Reverted #4330 which broke tarps
/:cl:


jjpark-kb:
fix rather than revert.

psykzz:
I also agree with fix don't revert, unless its the wrong direction.
That said a number of fixes have been made about tarps

PR #4388, opened 2020-07-03
Fixes hivemind dying to shuttles

psykzz:
Fixes hivemind dying to shuttles


Hughgent:
Won't it die anyways due to the shuttle removing weeds?

psykzz:
It will be teleported back to the core, not die. It was explicitly being gib'd beforehand.

InterroLouis:
It won't be teleported back to the core either, unless it moves when away from weeds. As is they'll just stand still then poop out a weed node.

PR #4394, opened 2020-07-04
Ports tg gc timeout fix

TiviPlus:
## About The Pull Request

https://github.com/tgstation/tgstation/pull/51890
## Why It's Good For The Game

a

## Changelog
:cl:
code: GC will timeout correctly
/:cl:


Hughgent:
Before, it was always 0 deciseconds, and this changes it to always 2 minutes.

Would you be so kind as to expound upon the benefits of this?

psykzz:
There is more context in the original PR, but the GC has two queues, A check queue and a harddelete queue.
Previously it seemed to have 3, this fixes the timeouts to ensure that the correct timeouts are used, in this case we don't have to wait 2 minutes for something to hardelete, and something isn't marked as a failure if it doesn't qdel on the same tick.

PR #4397, opened 2020-07-04
Fixes a runtime with alien embryo

Hughgent:
## About The Pull Request

Turns out that treating alien embryo's like reagent containers is a bad idea.

## Why It's Good For The Game

no runtimes

## Changelog
:cl: Hughgent
fix: examining alien embryo's won't cause a runtime anymore.
/:cl:


psykzz:
Instead of this you could inject the reagents into the larva at initalization instead of having to have a custom implementation.

Hughgent:
Thinking on this. I still think this is the more 'ideal' implementation.

first, If the stuff gets injected on init, then we are still adding reagents to the reagents var, which was causing the problem in the fist place.

Secondly, this 'hides' the reagents behind a unique variable. which I think is good for future 'xeno parts' being ground into reagents.

PR #4401, opened 2020-07-05
Reduces the volume of the hmg by 10%

psykzz:
## About The Pull Request
Reduces the volume of the hmg by 10%

## Why It's Good For The Game
It was a little loud


David-Stormwell:
Can you do the same for the pistol as well?

PR #4407, opened 2020-07-05
Fixes a bug in where moved ammo would change sprite

Rohesie:
Caused them to go invisible as well.


David-Stormwell:
Blessed owl

PR #4416, opened 2020-07-06
Misc hard del fixes and cleanup

Rohesie:
Tons of issues. The biggest culprits are explosions deleting things inside others that weren't meant to be deleted.

References cleaned:
* Mob HUD on the mob.
* Mobs on themselves through focus.
* Defibrillators, vehicles, plasmacutters and APCs on their powercells.
* Circuit boards on machines.
* Headset cameras on headsets.
* Mobs on machines through operator (this might need more work).
* Bodyscanners on scan consoles.
* Wall-mounted alarms air controls and  on SSradio
* Disposals bin on pipes
* Possibly a thing or two I may have forgotten.

Supply drop structures were made indestructible. There's no real reason to destroy them right now.

Also added a little tg circuitboard to fill the blank. Circuits will need their own port.


code/game/objects/items/defibrillator.dm
```diff
@@ -69,6 +78,21 @@
 	update_icon()
 
 
+///Wrapper to guarantee powercells are properly nulled and avoid hard deletes.
+/obj/item/defibrillator/proc/set_dcell(obj/item/cell/new_cell)
+	if(dcell)
+		UnregisterSignal(dcell, COMSIG_PARENT_QDELETING)
```
spookydonut, line 84: shouldnt you qdel the existing cell rather than just leaving it as a referenceless cell or a cell inside the defib's contents.

Rohesie: That would be assuming a lot. This proc only deals with changing the variable from one value to another and reacting to the event.
Extra handling of the cell needs to be done elsewhere.

PR #4405, opened 2020-07-05
Extools ref tracking

TiviPlus:
## About The Pull Request

NOTE: EXTOOLS REF TRACKING DOES NOT WORK WHEN DEBUGGING IN VSC, USE DREAMMAKER INSTEAD
https://github.com/tgstation/tgstation/pull/51467
Also updated extools dependency yadda yadda

## Why It's Good For The Game

🦀 

## Changelog
:cl:
code: Added Extools ref tracking, which can be enabled in compile_options.dm
/:cl:

PR #4417, opened 2020-07-06
Extools ref tracking 2

Rohesie:
Shamefully stealing Tivi's #4405 PR and adding more stuff. Feel free to close it if they incorporate it in theirs.

Removed the legacy reference finding, which is very, very inferior to this one (5 minutes search giving incomplete results vs instant one where you can move back and forwards between references).

Made things that failed to delete to be stored in a global list that can be accessed by a verb, if `REFERENCE_TRACKING` is enabled, so you can actually check for what is referencing them without having to worry about the GC destroying it.


spookydonut:
PR on to tivi's branch maybe?

TiviPlus:
Nah too lazy ill close mine

What would be the chances of making it possible to enable this on the server then logdumping the failed deletions to make them easier to sort through?

psykzz:
We could enable it for specific rounds easy enough, but would need to inform the server beforehand.

Rohesie:
would need to tweak it some to allow for that
currently it's in test mode only, which is for debugging

that said, this is ready and really makes finding refs easier

Rohesie:
Actually unless our qdel stuff is outdated, there is no need to run this on a regular game, as it will collect data on the things that are being hard deleted
This tool is for finding which those references are, which can be locally

PR #4418, opened 2020-07-06
Drone speed nerf

psykzz:
Slows down drones since they have gotten quite fast recently.


RenGusta:
Personally I'd recommend -0.7, -0.8, -0.9, -1 at most, as they're capable of self pheros to move very fast. Especially as they're not meant to be a mobile mainline combat caste, as even with your nerf they're still among the faster xenos, and as fast/faster than some offensive melee xenos.

PR #4419, opened 2020-07-06
Detpack buff (Maybe)

ZanderDowdall:
Testing this

## About The Pull Request

Literally just changes the det pack layer value from .1 to 2 so it can be hidden below things to allow marines to make traps

## Why It's Good For The Game

Adds balance in the forms of traps for marines, similar to resin holes

## Changelog
:cl:
balance: Marines can now make traps using det packs and creativity! Yay!
code: Changed the detpacks layer value from .1 to 2
/:cl:

### DIFF
--- a/code/game/objects/items/radio/detpack.dm
+++ b/code/game/objects/items/radio/detpack.dm
@@ -7,7 +7,7 @@
 	item_state = "plasticx"
 	flags_item = NOBLUDGEON
 	w_class = WEIGHT_CLASS_SMALL
-	layer = MOB_LAYER - 0.1
+	layer = MOB_LAYER - 2
 	var/frequency = 1457
 	var/on = FALSE
 	var/armed = FALSE


code/game/objects/items/radio/detpack.dm
```diff
@@ -7,7 +7,7 @@
 	item_state = "plasticx"
 	flags_item = NOBLUDGEON
 	w_class = WEIGHT_CLASS_SMALL
-	layer = MOB_LAYER - 0.1
+	layer = MOB_LAYER - 2
```
psykzz, line 10: Would `BELOW_OBJ_LAYER` hit what you intend here instead?
We should try to avoid using such large magic numbers like this.

ZanderDowdall: Ill be honest, I have no fucking idea. Mob layer worked when I tested it with bodies and objects, and the plan was to have the det packs hidden under just about everything, because the only use I can see em having is traps atm.
If that works on hiding the packs under shit, then it works for me

psykzz: So i looked at the layers, and i think the best match is RESIN_STRUCTURE_LAYER
What you have right now is way to low it would go under piping and wiring and basically everything.

`XENO_HIDING_LAYER` is low enough, as it mimics a hiding runner (under tables, items etc)

Hughgent:
@ZanderDowdall you will want to look at a file named `layers.dm` which has defines for every 'layer' in the game right now.

Beyond that, I think @psykzz has the right idea.

psykzz:
Closing since it was a 1 line change and stale

PR #4428, opened 2020-07-06
Experimental Module Prices

Hathkar:
## About The Pull Request

Increased the prices on the experimental jaegar modules to be more in line with  the specialist armor prices.

## Why It's Good For The Game

The fire module now costs almost as much as the specialist kit, as it does nearly the same exact thing.
Acid module also increased in cost to match the fire module, as it provides a similar function.
Valk autodoc also increased, but not as much, since the B18 autodoc outclasses it.
Bulk purchase of all 4 modules increased according to new prices, and given a slight discount afterwards for buying in bulk.

## Changelog
:cl:
balance: rebalanced something
/:cl:


Pariah919:
I think that the modules should be 25 overall, except for tyr which should be 20. 

Spec armor is old and deprecated, shouldn't be balanced around.

PR #4430, opened 2020-07-06
Add a cooldown to the autodoc toggle

psykzz:
Stops ungas spamming the toggle.


Hathkar:
Does this also prevent toggling on/off while in crit/sleep?

psykzz:
It didn't but now it does

PR #4369, opened 2020-06-29
Adds Events and the Intel collection random event

TiviPlus:
## About The Pull Request
Example event: Intel activation, marines must go and activate a terminal which then gives them points

todo: add the computers to maps

babys first tgui

Capture
Capture2
capture3
Capture4
Capture1

## Why It's Good For The Game

More objectives, Randomized round spice-up

## Changelog
:cl:
add: Midround events system
add: Added tgui-next intel computers and a intel event
/:cl:

PR #4431, opened 2020-07-06
Req Passive Income

Hathkar:
## About The Pull Request

Reduces requisitions passive point income to 2 points per minute, down from 4.

## Why It's Good For The Game
Marines now get fewer points for turtling up and doing nothing. This will create more incentive to push out and secure mining drills, xeno corpses, and intel packages (#4369).

## Changelog
:cl:
balance: Passive Point Income
/:cl:


psykzz:
Removing the scaling is a bad idea since it just removes the scaling with pop. What if we get 100 players? 
We already get ~60 players so typically squad points are doubled at round start.

spookydonut:
Don't remove round start point scaling that will massively destroy balance between low/high pop.

Hathkar:
Reverted the scaling change.

RenGusta:
I disagree with the necessity of this PR until we see the impact of the other two Req Impacting PRs you have anyway. 

I'd like if you could refrain from testing this until we see the effect of the other two PRs tanking req impact.

code/datums/gamemodes/distress.dm
```diff
@@ -222,6 +222,7 @@
 
 	var/scale = max(marine_pop_size / MARINE_GEAR_SCALING, 1) //This gives a decimal value representing a scaling multiplier. Cannot go below 1
 
+
```
Rohesie, line 225:
```suggestion
```

code/controllers/subsystem/points.dm
```diff
@@ -77,7 +77,7 @@ SUBSYSTEM_DEF(points)
 
 /datum/controller/subsystem/points/proc/scale_supply_points(scale)
 	supply_points = round(supply_points * scale)
-
+
```
Rohesie, line 80:
```suggestion

```

tgstation-server:

PR #4437, opened 2020-07-07
FIx slow shuttle spawning issues

psykzz:
Delay deleting the transit shuttle if for some reason the shuttle hasn't finished loading by the time the tick fires.
This gives shuttles 10* the shuttle subsystem wait period.


spookydonut:
Give it a shot but I'd rather move shuttle loading out of prestart

PR #4467, opened 2020-07-11
dropped code pixel shifted items will retain their pixel shifts

Hughgent:
## About The Pull Request

If something is pixel shifted in code, (not via mapping) it should retain that pixel shifted state whenever it is picked up.

## Why It's Good For The Game

This look like they are supposed to look like.

## Changelog
:cl: Hughgent
code: randomized drop pixel shifts are now based on their original pixel shifts.
/:cl:

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -202,8 +202,8 @@
 /mob/proc/dropItemToGround(obj/item/I, force = FALSE)
 	. = UnEquip(I, force, drop_location())
 	if(.)
-		I.pixel_x = rand(-6,6)
-		I.pixel_y = rand(-6,6)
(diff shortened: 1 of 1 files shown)


psykzz:
actually if something has modified the x/y i'd rather we just don't add the random pixels. Like if initial(pixel_x) != 0 then dont add random.

Hughgent:
I have to disagree, the original intention was that when you drop 16 pills onto the floor, they 'spill' everywhere. Having them be a neat perfect stack would be weird.

psykzz:
Pills shouldn't have pixel_x offsets though?
My point was like if a large item has a specific offset applied to, there is no reason to modify that.
An example being a 2 tile fuel tank or xenos, they have -16 pixel offset to center them. I'd rather they stack

PR #4481, opened 2020-07-12
POS-FIRING RANGE

Sibg:
## About The Pull Request
Changes the MP space into a firing range for marine use aswell as giving the engineering bay another pacman replacing the useless old one. The MP stuffs are now located in the CIC aswell 

## Why It's Good For The Game
Uses the useless mp space for a firing range marines can use and gives engineering a extra pacman.
## Changelog
:cl:
add: The Firing Range & Extra Pacman
del: Old Military Police Stuffs
/:cl:

PR #4493, opened 2020-07-14
Gives RO a taser

TiviPlus:
## About The Pull Request

title
## Why It's Good For The Game

How hard is it towait 10 seconds for the RO to give you something. For example this screenshot is what happened after a fucking marine couldnt wait 10 seconds
image

## Changelog
:cl:
balance: The RO has a taser
/:cl:

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -375,7 +375,7 @@ engineer, construction, leadership, medical, surgery, pilot, police, powerloader
 	construction = SKILL_CONSTRUCTION_PLASTEEL
 	leadership = SKILL_LEAD_TRAINED
 	powerloader = SKILL_POWERLOADER_MASTER
-	police = SKILL_POLICE_FLASH
+	police = SKILL_POLICE_MP//You broke into the wrong req bucko
 
 /datum/skills/ST
 	name = SHIP_TECH
@@ -639,7 +639,7 @@ engineer, construction, leadership, medical, surgery, pilot, police, powerloader
 
 	medical = SKILL_MEDICAL_EXPERT
 	surgery = SKILL_SURGERY_EXPERT
-  
+
 
 /datum/skills/sectoid
 	name = "Sectoid"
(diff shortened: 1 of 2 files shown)


Ridgyaxe:
The problem with this PR, is that the captain opened fire on two marines first, and there was no RO. The grenades were from a result of the captain shooting.

Hughgent:
The RO already has the FLASH level skill, do they not have access to the flash?

TiviPlus:
It spawns in their locker apparently but god knows

And this isnt a case of "oh it was one time" but marines keep breaking into req over and over and its a pain in the ass for everyone involved

David-Stormwell:
Check if the RO has high enough police skill for it, if not increase it.

Chaznoodles:
> It spawns in their locker apparently but god knows
> 
> And this isnt a case of "oh it was one time" but marines keep breaking into req over and over and its a pain in the ass for everyone involved

The req lockers on PoS aren't the Requisitions Officer-specific one, that has a flash. Change them to the Requisitions Officer locker instead of the Cargo Tech one, and avoid ROs acting like world police in future.

code/datums/skills.dm
```diff
@@ -375,7 +375,7 @@ engineer, construction, leadership, medical, surgery, pilot, police, powerloader
 	construction = SKILL_CONSTRUCTION_PLASTEEL
 	leadership = SKILL_LEAD_TRAINED
 	powerloader = SKILL_POWERLOADER_MASTER
-	police = SKILL_POLICE_FLASH
+	police = SKILL_POLICE_MP//You broke into the wrong req bucko
```
Sibg, line 378: Looks like RO is going to do some hurting.

neohenry:
I would prefer a taser over a flash since I can quick draw a nonlethal weapon against a disrespectful player. Then again, (1) I'm a security player and (2) I'm one of the few players that like to have a taser in the server.

Not having a taser in these rare freak incidents makes me want to make it easier for me to have a taser.

neohenry:
#4481

psykzz:
Consider adding tasers to the CIC armory too, so it can be paied with #4515 and #4513

PR #4498, opened 2020-07-15
Increases bullet and Laser armor on pas-11 and modular exo suit 

Pariah919:
## About The Pull Request
people said ff was pretty bad

## Why It's Good For The Game
makes ff no longer one shot you

## Changelog
:cl:
balance: jaeger and vest now cap out at 75 bullet/laser armor
/:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -242,7 +242,7 @@
 	name = "\improper PAS-11 pattern armored vest"
 	desc = "A somewhat outdated but robust armored vest, still in use despite the rise of exoskeleton armor due to ease of use and manufacturing. Tougher than it looks. Use it to toggle the built-in flashlight."
 	icon_state = "marinevest"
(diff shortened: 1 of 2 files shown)


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -26. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

XSlayer300:
Makes them also bullet sponge-y for people who uses HvH combat though.

RenGusta:
Good tbh, HVH smelly

jroinc:
This is high enough that it removes FF as a concern and allows spraying people down with rifle rounds as a valid tactic with no long-term consequences, given that shrapnel does NOTHING unless you run with it unsplinted.

psykzz:
If this happens, it should be temporary, when aim mode or an alternative comes in that gives you IFF, then this should be reverted.

Pariah919:
fixed in another pr aswell
the one that adds armor variance

PR #4513, opened 2020-07-17
tasers no longer instantly stun now do stamina damage

Pariah919:
## About The Pull Request
title

## Why It's Good For The Game
idk this was planned for a long time

## Changelog
:cl:
balance: taser now do stamina damage
/:cl:

PR #4515, opened 2020-07-17
Shipside Skills Update

David-Stormwell:
## About The Pull Request

Gives most ship staff MP skill instead of just Flashing skill.
Paired with the taser re-balance: 4513 it should be fine.

## Why It's Good For The Game

So that staff can better defend their-self in HvH or if marines are being unlawful.

## Changelog
:cl:
balance: Most shipside staff have MP level skill.
/:cl:


psykzz:
Pending #4513 for merge

PR #4518, opened 2020-07-17
Creates new entrances into self destruct on sulaco

psykzz:
## About The Pull Request

Encircles SD with more maintenance to give xenos a chance to assault from multiple angles

## Why It's Good For The Game

Currently xenos are restricted from assaulting a single narrow path, this adds more narrow paths through maintenance.

## Changelog
:cl:
add: Added new entrances into self destruct on sulaco
/:cl:


psykzz:
@SealCure do you not like it because the number of entrances is too much or do you have another opinion?

SealCure:
I feel like this just gives xenos a way directly into the SD through the east maint that's been added. When it came to an SD siege before, if xenos were competent and worked for it then they could smash marine defences and get through, equally if marines worked together and had a good defence then they'd be rewarded for it. With the change, xenos can now just melt walls directly into the SD room itself and bypass the whole point of an SD siege for both sides.

PR #4523, opened 2020-07-17
Add new Rav ability: Ignore Pain

psykzz:
## About The Pull Request

Allows ravs to ignore crit for 10 seconds, but can still die from too much damage

## Why It's Good For The Game

Give ravs some buttons, and allow them to go "more ham"

## Changelog
:cl:
add: Added new Rav ability: Ignore Pain - Allows ravs to ignore crit for 10 seconds, but can still die from too much damage.
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals.dm
+++ b/code/__DEFINES/dcs/signals.dm
@@ -470,6 +470,7 @@
 #define COMSIG_XENOABILITY_RAVAGER_CHARGE "xenoability_ravager_charge"
 #define COMSIG_XENOABILITY_RAVAGE "xenoability_ravage"
 #define COMSIG_XENOABILITY_SECOND_WIND "xenoability_second_wind"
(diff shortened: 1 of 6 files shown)


jroinc:
This seems like a very dangerous ability to use.

Assuming a "normal" combat situation of ~0-2 other ayys, no weeds near marines, and ~2-8 marines, this will get ayys killed by virtue of either critting off weeds, critting near ungas, or just having ungas rush in cause they know you're gonna probably go down for like 10 seconds nearby.

Could you add in 2 warnings? Move the "plasma draining from our veins" one from when it ends to 5 seconds before it ends, and add in another big, RED warning when you are below crit health? Something like "only our rage sustains us now"?

psykzz:
I added the jitter at 7 seconds, but can move it to 5 seconds.
I also like the red warning i can add that now.

psykzz:
image

psykzz:
Ready to merge

PR #4535, opened 2020-07-19
Removes access restrictions for marinemeds

Mouseofthecake:
## About The Pull Request

Removes the medic/doctor access requirement for marinemeds.

## Why It's Good For The Game

"Haha, yes, Philip, let's put a vending machine filled with meds in the medbay lobby, for all to use. The trick? Ninety-five percent of all marines won't be able to access it without hacking! And the other kicker; let's call it the MarineMed. Funniest shit I've ever come up with."
Really, though, QoL and stuff.

## Changelog
:cl:
del: Removed Marinemed access requirements.
:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -339,7 +339,6 @@
 	icon_state = "marinemed"
 	icon_deny = "marinemed-deny"
 	product_ads = "Go save some lives!;The best stuff for your medbay.;Only the finest tools.;All natural chemicals!;This stuff saves lives.;Don't you want some?;Ping!"
-	req_one_access = list(ACCESS_MARINE_MEDBAY, ACCESS_MARINE_CHEMISTRY, ACCESS_MARINE_MEDPREP) //Medics, doctors and researchers can access
 	wrenchable = FALSE
 	products = list(/obj/item/reagent_containers/hypospray/autoinjector/quickclot = 6,
 					/obj/item/reagent_containers/hypospray/autoinjector/bicaridine = 6,


jroinc:
Is the intention of this PR to give all marines all meds regardless of job?

Seems like that's the intent.

It's called a "marinemed" cause MARINE MEDics can use it.

Chaznoodles:
Those vendors are intended for medic use.

Mouseofthecake:
They're always just hacked at roundstart anyway. Why bother leaving the access requirements on, even if only medics are intended to use it?

TiviPlus:
So make them unhackable?

Mouseofthecake:
That's... why would you do that?

jroinc:
I mean, it's a valid counterpoint.

Your argument is "Everyone uses them and they always hack them, so make them all-access"

His (and my) argument is "They're medic-only per name, so don't. And if it's a big deal, lock'em."

Mouseofthecake:
The reason for this PR is to leave them unlocked, so you don't have to drop what you're doing to find the tools to hack it if you need it. It's simply QoL. I don't know what exactly gave the impression that I think hacking them is bad.

Pariah919:
isn't it called marinemed because its marine medicine

Hughgent:
Get some consensus from the playerbase about this PR please.

Kuro-ai:
I don't think PFCs need access to even more stuff

Dickff:
I'm meh to this PR, but I think 100% we SHOULD NOT start making things unhackable. One of the main reasons I came to this server from CM is the lack of artificial barriers to what marines could and could not do.

spookydonut:
These are for medics to use at round start right?  If PFCs can take stuff from them, will that mean medics won't get the stuff they need?  -> if yes then we shouldnt merge this

Will PFCs getting stuff from these mean they don't need medics to help them or medics don't need to bother taking the stuff that's in these vendors?

After deployment, who uses these vendors apart from latejoining marines?  If no one then just move these into medic prep rooms.

Mouseofthecake:
As a corpsman, the only time I interact with marinemeds is to get imialky and tricord. Which are awfully absent from the corpsman prepper vendors, for some reason. Might look into adding those to remedy marinemeds being public access.
They also have two stasis bags each, which is nice, but no standard PFC is gonna get those just to hoard.

PR #4542, opened 2020-07-20
Buffs the HE shell for orbital cannons

2whatever2:
## About The Pull Request
Buffs the HE warhead shell for orbital beacons

## Why It's Good For The Game
HE is a glorified mortar shot but instead of using a mortar it uses a orbital cannon.

## Changelog
:cl:
tweak: Raises HE explosion size to 6/10/15
/:cl:

### DIFF
--- a/code/game/objects/structures/orbital_cannon.dm
+++ b/code/game/objects/structures/orbital_cannon.dm
@@ -365,7 +365,7 @@
 	icon_state = "ob_warhead_1"
 
 /obj/structure/ob_ammo/warhead/explosive/warhead_impact(turf/target, inaccuracy_amt = 0)
-	explosion(target, 5 - inaccuracy_amt, 7 - inaccuracy_amt, 10 - inaccuracy_amt, 11 - inaccuracy_amt)
+	explosion(target, 6 - inaccuracy_amt, 10 - inaccuracy_amt, 15 - inaccuracy_amt, 11 - inaccuracy_amt)
 
 
 


code/game/objects/structures/orbital_cannon.dm
```diff
@@ -365,7 +365,7 @@
 	icon_state = "ob_warhead_1"
 
 /obj/structure/ob_ammo/warhead/explosive/warhead_impact(turf/target, inaccuracy_amt = 0)
-	explosion(target, 5 - inaccuracy_amt, 7 - inaccuracy_amt, 10 - inaccuracy_amt, 11 - inaccuracy_amt)
```
psykzz, line 368: Just to be clear, that is 6 tiles of insta gib for like 70% of xenos.
Just checking the intention is clear.

The rest im good with 👍

Chaznoodles:
HE and Cluster basically serve the same purpose. I'd be happy with just removing HE and keeping Cluster as-is, for the explosive OB.

code/game/objects/structures/orbital_cannon.dm
2whatever2: If they decide to stand in the OB laser after it left or hear it beep loudly and stay in range for a OB then cluster would've have killed them 10 times over. HE before was basically a HE shell for mortars but the value for heavy 7 is supposed to be 6 i'll go fix that

PR #4578, opened 2020-07-25
Fixes cryo cells

Hughgent:
## About The Pull Request

Cryocells now actually process their mob.

~~I think there was something 'off' with the processing SubSystem. What should happen is that it 'starts processsing' when you turn the machine on. But it wasn't. So I made it restart processing when the mob is inserted into the machine (on the presumption that the cryocell is 'on' beforehand) and the cells start to function again.~~

Turns out, pressing the 'on' button would just set the variable to on, but wouldn't do the actual `turn_on()` proc

Side note, I would really like for this 'understandnessTest' file to be added to the ignore because it's bugging me.

Made the `process_occupant` early return.

## Why It's Good For The Game

Cryocells actually heal now.

## Changelog
:cl: Hughgent
fix: Cryocells now heal again
/:cl:


tgui/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "babel-eslint": "^10.0.3",
     "eslint": "^6.7.2",
-    "eslint-plugin-react": "^7.17.0"
+    "eslint-plugin-react": "^7.17.0",
+    "yarn": "^1.22.4"
```
spookydonut, line 19: revert dis

tgui/yarn.lock
```diff
@@ -6349,3 +6349,8 @@ yargs@13.2.4:
     which-module "^2.0.0"
     y18n "^4.0.0"
     yargs-parser "^13.1.0"
+
+yarn@^1.22.4:
+  version "1.22.4"
```
spookydonut, line 6354: dis too

PR #4582, opened 2020-07-25
Ports runechat subsystem from TG

psykzz:
Original PR https://github.com/tgstation/tgstation/pull/52425

## About The Pull Request

This replaces all use of timers for Runechat with a dedicated subsystem. This subsystem uses a similar approach to the timer subsystem, with an event queue that is broken up into buckets for each tick and those buckets just being a circular doubly-linked list. The result is zero dependence on timers.

This is a tad experimental, comments welcome.

## Why It's Good For The Game

Moves Runechat into a more controllable and monitor-able state, removes any overhead that may be related to timers. Should reduce memory overhead of this system in particular as this was implemented with no additional datums.

## Changelog
:cl: bobbahbrown
add: Runechat now runs under its own subsystem, this should improve performance and allow for finer control over its performance.
/:cl:

### DIFF
--- a/code/__DEFINES/_subsystems.dm
+++ b/code/__DEFINES/_subsystems.dm
@@ -92,6 +92,7 @@
 #define FIRE_PRIORITY_TGUI			110
 #define FIRE_PRIORITY_TICKER		200
 #define FIRE_PRIORITY_CHAT			400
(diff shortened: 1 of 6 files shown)


psykzz:
Was merged for a few rounds no real isssues.

PR #4584, opened 2020-07-25
Fixes view variables

psykzz:
This was broken with ref tracking pr since we seem to have different systems and expectations


psykzz:
Its unusable currently so...

PR #4601, opened 2020-07-27
Nerfs the pacman

psykzz:
## About The Pull Request
Max sheets set to 10
30s max per sheet

## Why It's Good For The Game
Generators are OP.

## Changelog
:cl: 
balance: nerfs the pacman
/:cl:


neohenry:
I also notice that the PAC must be melted before it becomes nonfunctional. Slashing or shooting it won't do any damage to it.

With generators, it takes "infinite" slashes to break them, making it impossible for xenos to break.

davidchan13:
Did the overheat feature of pacmans get removed?

psykzz:
Overheat is there but no one uses "full power" so it never gets hot.

PR #4605, opened 2020-07-28
fix IB cure with bica overdose mechanic

rhomm-dev:
## About The Pull Request

This is a minor code change that makes curing IB using a bicaridine overdose work as intended by removing the wound when it reaches 0 damage. Currently you can heal all the IB wound's damage but still have to wait for a total of 10 minutes before the wound is actually removed -  causing quite a bit of confusion. (note: we should probably refactor this function at some point)

## Why It's Good For The Game

Less confusion good

## Changelog
:cl:
fix: makes curing IB with bica OD work as intended
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -541,7 +541,11 @@ Note that amputating the affected organ does in fact remove the infection from t
 				W.open_wound(0.1 * wound_update_accuracy)
 			if(bicardose >= 30)	//overdose of bicaridine begins healing IB
 				W.damage = max(0, W.damage - 0.2)
-
+			
+			if(W.damage <= 0)
+				wounds -= W // otherwise we are stuck with a 0 damage IB for a while
+				continue
+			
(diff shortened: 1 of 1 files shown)


Hughgent:
This is bad. it's already handled by the 10 minute timer as you say. don't make IB (and by extension all 'internal' wounds) instantly heal when they reach 0. don't make them special.

jroinc:
The issue with that is that all other wounds do nothing at 0, besides look cool.

As far as I'm aware, a 0-damage IB still does bleeding for the full 10 minutes.

factoryman942:
> As far as I'm aware, a 0-damage IB still does bleeding for the full 10 minutes.

it didn't, i spent a while with moths with 0-damage IB wounds, they don't bleed.
https://github.com/tgstation/TerraGov-Marine-Corps/blob/d81a6471871eb2c1c7992e29527264a9149f73f7/code/modules/organs/limbs.dm#L546
the bleeding is proportional to wound damage, so at 0 damage it removes no blood

Pariah919:
@Hughgent if 0 damage IB does literally nothing, then let it cure itself. No reason to lie to the player.

PR #4617, opened 2020-07-30
Escape pod tgui

TiviPlus:
## About The Pull Request

yay
## Why It's Good For The Game

looks good

## Changelog
:cl:
add: New escape pod tgui
/:cl:


tgui/packages/tgui/interfaces/EscapePod.js
```diff
@@ -0,0 +1,27 @@
...
+        <Section title="Escape Pod">
+          Welcome to Nanotransens least luxurious survival pod!
+          Have a pleasant stay!
+          <Box
+            width="100%"
+            textAlign="center">
+            <Button.Confirm
+              m="50"
+              content="Launch evacuation pod"
+              color="red"
```
psykzz, line 20:
```suggestion
              disabled={data.can_launch}
              color="red"
```

psykzz: If not just remove the `ui_data` section

spookydonut:
Do we really want to make it slower to launch an escape pod?

I mean tgui is cool but if i'm rushing into a pod and theres a hostile on me i'd rather it just launch soon as i click the control panel

TiviPlus:
I havent made it slower at all

It just used a html menu before so this should make it faster actually

PR #4651, opened 2020-08-05
New xeno respawns on distress

psykzz:
## About The Pull Request
Adds larva respawns for xenos on distress.
Starting at 15minutes reduced by 2 minutes for every active silo.

This uses the same larvapoints formula, so the number will never get too high.

## Why It's Good For The Game

comebacks and mistakes should be a thing.
Also marines get respawns, why not xenos.
Reduces pressure on getting a capture.

## Changelog
:cl:
add: Added respawns for xenos on distress, timed based on the number of silos available.
/:cl:

### DIFF
--- a/code/datums/gamemodes/distress.dm
+++ b/code/datums/gamemodes/distress.dm
@@ -85,7 +85,7 @@
 	priority_announce(SSmapping.configs[GROUND_MAP].announce_text, SSmapping.configs[SHIP_MAP].map_name)
 
 
-/datum/game_mode/infestation/disstress/spawn_resin_silos()
+/datum/game_mode/infestation/distress/spawn_resin_silos()
 	// Want a max of only 5 silos
 	var/all_silos = GLOB.xeno_resin_silo_turfs.Copy()
 	for(var/i in 1 to 5)
--- a/code/datums/gamemodes/distress.dm
+++ b/code/datums/gamemodes/distress.dm
@@ -5,7 +5,7 @@
 	name = "Distress Signal"
 	config_tag = "Distress Signal"
 	required_players = 2
-	flags_round_type = MODE_INFESTATION|MODE_LZ_SHUTTERS|MODE_XENO_RULER
+	flags_round_type = MODE_INFESTATION|MODE_LZ_SHUTTERS|MODE_XENO_RULER|MODE_XENO_RESPAWN_WAVE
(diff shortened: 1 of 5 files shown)


jroinc:
I would cap this to REQUIRE a silo be intact (if it doesn't already).

Also, there's the issue that ungas can just drop and fire off 5-7 incin shells at memorized coords on many maps.

code/datums/gamemodes/game_mode.dm
```diff
@@ -338,6 +340,7 @@ Sensors indicate [numXenosShip || "no"] unknown lifeform signature[numXenosShip
 
 	for(var/i in GLOB.player_list)
 		var/mob/M = i
+		grant_eord_respawn(null, M)
```
TiviPlus, line 343: isnt this done on COMSIG_GLOB_MOB_LOGIN in line 332

psykzz: Yes but if you don't enter a new mob (ie you were already a ghost) you never got the verb.

code/datums/gamemodes/distress.dm
```diff
@@ -33,13 +33,18 @@
...
 	)
 
+	// Xeno respawns
+	xeno_respawn_wave_timer = 15 MINUTES
+	xeno_respawn_silo_reduction = 2 MINUTES
+
 	var/round_stage = DISTRESS_MARINE_DEPLOYMENT
 
-	var/bioscan_current_interval = 45 MINUTES
+	var/bioscan_current_interval = 30 MINUTES
```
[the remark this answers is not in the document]
psykzz: This is going to be part of a large change were we reduce the time taken to deploy and since we deploy faster, we should start getting bioscans sooner.

I just haven't changed the deployment lock yet.

Hathkar:
Ideally the silos should be located underground so they can't be mortared.

SilencedMP5A5:
so you're gonna bring marine respawns back down to 30 mins too, right?

jroinc:
> so you're gonna bring marine respawns back down to 30 mins too, right?

And make them based on marine/ayy ratio to remove the permarine issue we had earlier, right?~

psykzz:
> Ideally the silos should be located underground so they can't be mortared.

They will be buildable (up to 5), but will have a look at the ceilings too

Hathkar:
Any idea when this will be finished? Assuming it's not finished already.

PR #4658, opened 2020-08-07
Xeno IC names

TiviPlus:
## About The Pull Request

Might want to consider bumping the savefile version to reset xeno names but I wanna hear admins thoughts on that first

## Why It's Good For The Game

image

## Changelog
:cl:
add: Xenos now have IC names
admin: ...yes you have to follow naming rules
/:cl:


jroinc:
yessss

finally IC identification

now we can see hivemind get as toxic as marine comms does~

Hathkar:
It would be very useful if xenos without a name would have random numbers like before instead of being (Undefined). If Young Runner (Undefined) is breaking some rules, and there are multiple Young Runner (Undefined), things can get messy on the admin side of things.

psykzz (approved):

PR #4663, opened 2020-08-08
Use unrelenting force via keybind instead of just selecting

psykzz:
## About The Pull Request
When you use the keybind currently it just selects the ability for use.
Since all you need is a direction if you use via keybind you can just shout in that direction instead.
## Why It's Good For The Game

Playing shrike a little easier

## Changelog
:cl:
tweak: unrelenting force can be used directly from a keybinding
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/shrike/abilities_shrike.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/shrike/abilities_shrike.dm
@@ -139,6 +139,7 @@
 	mechanics_text = "Unleashes our raw psychic power, pushing aside anyone who stands in our path."
 	cooldown_timer = 50 SECONDS
 	plasma_cost = 300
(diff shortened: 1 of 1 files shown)


TiviPlus:
Personally I'd prefer it the way it is but let's see what players say

psykzz:
The only thing is lost having it this way is the face_atom call

TiviPlus (approved):
testmerge seems to say yes

David-Stormwell:
This makes it so much worse tactically.
I want to aim it, not use it where I'm facing.

2whatever2:
I like this so i can keep fling toggled instead of toggling back to unrelenting :>

psykzz:
That was the aim.

David-Stormwell:
@psykzz If we do end up adding this, can we make it a preference option? So I can select it with hotkeys instead of instantly using it?

TiviPlus (approved):
havent seen any more complaints, looks like people got used to it

PR #4665, opened 2020-08-08
Removes FC SG (again(again)) and adds tazers to POS

TiviPlus:
## About The Pull Request

FC should have no SG

Tazers are as reccomended by psy in whatever tazer PR it was again
## Changelog
:cl:
balance: Removes SG in the POS and adds tazers to POS armoury
/:cl:


neohenry:
heh

psykzz:
Would it be better to just have the taser spawned on the command roles that can use them

Chaznoodles:
Can we stop handing out tasers before we end up with people acting like MPs 😢

Sibg:
Command staff are just MP's now when the need arises

PR #4666, opened 2020-08-09
Makes fireman carry easier

psykzz:
## About The Pull Request
Makes doing a firemans carry easier
people with medical skills will be faster.

This makes it inline with what xenos do, keeping it consistent between mobs

Additionally adds a new getPercent for skills that should make it easier to see how much of a skill you have that can be lerp'd against.

## Changelog
:cl:
tweak: you now firemans carry people by dragging and clicking yourself.
/:cl:

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -163,6 +163,9 @@ engineer, construction, leadership, medical, surgery, pilot, police, powerloader
 
 #undef SKILLSID
 
+/datum/skills/proc/getPercent(rating, max_rating)
+	return CLAMP01(vars[rating] * 100 / max_rating * 0.01)
+
 /datum/skills/proc/getRating(rating)
 	return vars[rating]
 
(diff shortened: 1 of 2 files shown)


jroinc:
Is the goal of this to remove stretchers? Cause if, as a medic, I can carry someone faster and quicker than a stretcher with the same downsides for me (and less downsides for others), there's no reason to use a stretcher.

davidchan13:
> Is the goal of this to remove stretchers? Cause if, as a medic, I can carry someone faster and quicker than a stretcher with the same downsides for me (and less downsides for others), there's no reason to use a stretcher.

Stretchers can still mount a stasis bag and are generally faster save for those rare moments the game fucks up and your drag doesn't pull the stretcher along with you.

jroinc:
> Stretchers can still mount a stasis bag and are generally faster save for those rare moments the game fucks up and your drag doesn't pull the stretcher along with you.

Stasis bags aren't generally used outside the Alamo because everyone who isn't a corpse is mobile, and it's hard as hell to convince people to stay in them when they think they can run to the Alamo faster. Same with roller-bedding people who aren't crit/dead, because you can't get people to accept that maybe they'd be faster if they didn't walk.

As far as I'm aware, fireman's carry has no slowdown, while rollerbed has an associated slowdown value.

So-

Rollerbed- Takes a piece of equipment to use, takes about as long to use as a fireman's carry, is slower, and BLOCKS YOUR SHOTS to the rear, so you can't cover a retreat/patient with a one-handed gun.

Fireman's carry- Takes no equipment to use, takes about as long to use as a rollerbed, is faster, allows free-fire in all directions with a one-handed weapon.

Where's the (practical) downside? Removing rollerbeds is fine by me, I'm just pointing it out.

psykzz:
Its a good point, and the intention wasn't to do anything to stretchers just making ffiremans carry standard, perhaps i can add more of a slowdown for firemans carry since you have to carry the weight.

So in order of whats slowest to fastest.

- Dragging 
- Fireman carry
- Strectcher

PR #4675, opened 2020-08-10
Heavy Helmet

Karl-Franz-I:
Thought it was silly the only job in existence that gets the heavy helmet is the smartgunner. It should be an option to take with the PAS-11 if you want. While it could have been added to the "essential set", I don't think it's necessary.

It has been massacred stat wise to appease the masses, due to issues with balance that would result from 75 melee armor. It is now aesthetic, for those true gamers still using PAS-11 vests.

Bing Bing Yahoo!

:cl:
add: Added the heavy helmet variant to surplus vendors.
/:cl:

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -293,11 +293,10 @@
 	flags_item_map_variant = (ITEM_JUNGLE_VARIANT|ITEM_ICE_VARIANT|ITEM_PRISON_VARIANT|ITEM_ICE_PROTECTION)
 
 /obj/item/clothing/head/helmet/marine/heavy
-	name = "\improper M10H pattern marine helmet"
-	desc = "A standard M10H Pattern Helmet. This is a upgraded version of the standard M10 helmet, offering more protection and face cover in expense of accuracy and visibility."
+	name = "\improper M10E pattern marine helmet"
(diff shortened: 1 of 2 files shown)


code/modules/clothing/head/helmet.dm
```diff
@@ -293,11 +293,10 @@
...
 
 /obj/item/clothing/head/helmet/marine/heavy
-	name = "\improper M10H pattern marine helmet"
-	desc = "A standard M10H Pattern Helmet. This is a upgraded version of the standard M10 helmet, offering more protection and face cover in expense of accuracy and visibility."
+	name = "\improper M10E pattern marine helmet"
+	desc = "A standard M10E Pattern Helmet. This is a modified version of the standard M10 helmet, offering an enclosed visor apparatus. It doesn't look to be any more protective damage wise."
 	icon_state = "heavyhelmet"
-	soft_armor = list("melee" = 75, "bullet" = 70, "laser" = 60, "energy" = 40, "bomb" = 60, "bio" = 50, "rad" = 0, "fire" = 50, "acid" = 55)
-	accuracy_mod = -15
+	soft_armor = list("melee" = 50, "bullet" = 60, "laser" = 30, "energy" = 20, "bomb" = 20, "bio" = 50, "rad" = 0, "fire" = 50, "acid" = 50)
```
TiviPlus, line 299:
```suggestion
```
inherited from parent, since you want them to be the same statwise

PR #4680, opened 2020-08-10
Fixes AI interactions

psykzz:
Fixes AI not being able to use pretty much anything.
Previously it was doing an old "range" check instead of a cam vis check.

Also adds some missing flags.


code/_onclick/ai.dm
```diff
@@ -19,12 +19,14 @@
 		A.move_camera_by_click()
 
 
-/mob/living/silicon/ai/ClickOn(atom/A, location, params)
+/mob/living/silicon/ai/ClickOn(src, location, params)
```
psykzz, line 22:
```suggestion
/mob/living/silicon/ai/ClickOn(atom/A, location, params)
```
ops

code/_onclick/ai.dm
```diff
@@ -42,15 +44,23 @@
 	var/turf/pixel_turf = get_turf_pixel(A)
 	if(isnull(pixel_turf))
 		return
-
-	if(!can_see(A) && isturf(A))
+	if(!can_see(A))
+		if(isturf(A)) //On unmodified clients clicking the static overlay clicks the turf underneath
+			return //So there's no point messaging admins
+		message_admins("[ADMIN_LOOKUPFLW(src)] might be running a modified client! (failed can_see on AI click of [A] (Turf Loc: [ADMIN_VERBOSEJMP(pixel_turf)]))")
+		var/message = "[key_name(src)] might be running a modified client! (failed can_see on AI click of [A] (Turf Loc: [AREACOORD(pixel_turf)]))"
```
TiviPlus, line 51: There is currently a bug on both tg and tgmc that spuriously allows ais to see the whole map so this isnt 100% accurate, should elaborate that that might also be the case

psykzz: I dont think there is a need to make the string longer, most of the time when this happens admins either ask each other or ping me about it.

TiviPlus: Okey dokey then

PR #4687, opened 2020-08-12
Revolver Rail light pixel fix

Hughgent:
## About The Pull Request

One pixel lower

## Why It's Good For The Game

aaaahhhhh
scream1

AHHHHHH
scream2

**AAAAAAAHHHHHHHH**
scream3

phew
ahhh

## Changelog
:cl: Hughgent
fix: Standard Revolver rail attachments no longer hover.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/revolvers.dm
+++ b/code/modules/projectiles/guns/revolvers.dm
@@ -370,7 +370,7 @@
 						/obj/item/attachable/scope,
 						/obj/item/attachable/lasersight,
 						/obj/item/attachable/scope/mini)
-	attachable_offset = list("muzzle_x" = 30, "muzzle_y" = 21,"rail_x" = 17, "rail_y" = 23, "under_x" = 22, "under_y" = 17, "stock_x" = 22, "stock_y" = 19)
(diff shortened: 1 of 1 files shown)


psykzz:
I would've put it here (excuse the mspaint job)

image

Hughgent:
Unfortunately, the variable I changed is for all the rail based attachments.

So if I moved it there, then we'd have side mounted red dots and such.

psykzz:
Can each attachment have its own offset, or is it just a mount point on gun?

Hughgent:
So it comes down to visualization of the 'slot'

Each gun has its own 'attachment point' the pixel x and y where that 'slot' is located.

Each attachment has it's own pixel x and y which adjusts how it looks when attached.

So we can adjust the weapons 'slot' but that adjusts all sprites that attach to that 'slot'. We can also adjust the attachment, but that adjusts the attachment for all weapons.

So in this case, the 'rail light' takes the 'rail slot' and we've decided that the 'rail slot' is on top of the weapon where most sights go.

From your picture, you'd probably want to make an 'under barrel light' instead. Which would be it's own thing.

PR #4695, opened 2020-08-13
Rebalances marine armor storage/utility. Baldur lights are no longer free.

Pariah919:
## About The Pull Request
Modular armor storage was a horrifically large powercreep that causes incredible issues longterm and unsurprisingly was a bad idea. It should be cut down to 3 options (Medical/Engineering/General)
What I'm saying is to delete the crap out of ammo/IS. They're overpowered and strictly better than everything else.
Remake general armor storage to fit 2 items, PAS-11 to 2. 
Baldur itself is THE BEST module in the game and it should not be free. It is no longer free. PAS-11 changed to reflect this.

## Why It's Good For The Game
The entire tradeoff of IS was the fact that it was literally medium armor with nearly no benefits other than the storage and marginally better armor. 
Every marine having free rail light is horrifically broken. 
Marine armor storage in general was heavily powercreeped, this had to be done.
Then again this generally shows the fact that modular armor was probably a :mistake:

## Changelog
:cl:
balance: you can no longer acquire IS/Ammo storage modular armor storage, you can no longer acquire baldur lights for free. Baldur price in req significantly increased. PAS-11 light range nerfed in consideration of this, storage nerfed too.
/:cl:


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -38. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

TiviPlus (changes requested):
Split the lamp and storage changes

Dont lazily remove the storages, nerf them instead if they're too good

code/modules/reqs/supplypacks.dm
```diff
@@ -798,12 +796,8 @@ ARMOR
 	name = "Jaeger baldur modules"
 	contains = list(
 		/obj/item/armor_module/attachable/better_shoulder_lamp,
-		/obj/item/armor_module/attachable/better_shoulder_lamp,
-		/obj/item/armor_module/attachable/better_shoulder_lamp,
-		/obj/item/armor_module/attachable/better_shoulder_lamp,
-		/obj/item/armor_module/attachable/better_shoulder_lamp,
 	)
-	cost = 7
+	cost = 12
```
Hathkar, line 800: 12 supply points for a single baldur lamp? Personally I think you're overestimating the value of the lamp. Being (Nearly) fireproof or acid proof is well worth 12 req points.

jroinc: If I have a Baldur module and a raillight on my gun (or a beltgun that's not my primary...), I can see the WHOLE screen, minus one single tile at each corner of the screen.

Might even be able to do that without a raillight, TBH. But free Baldur is functionally equivalent to NVG's for everyone.

jroinc:
> Split the lamp and storage changes
> 
> Dont lazily remove the storages, nerf them instead if they're too good

What about making IS a whole module? I agree it's far too good to be just a "free" satchel, but if you make it take up space that could be an autodoc, fire resist, extra armor, or NVG-level lighting, that seems like an acceptable tradeoff.

Marmio64:
don't really feel like the standard PAS-11 vest needs a storage nerf, but the light nerf would encourage more flare use. I never feel a need to have to use flares due to always having so much light available, but the storage capacity of the vest never seemed overpowered.

Pariah919:
i'm spliltting this

PR #4702, opened 2020-08-16
Revert "Move hardcoded speaking keys into keybind defaults"

### LINKED ISSUE #4694: Cannot use T to talk.
## Reproduction: 
A new player or anyone who has not rebind-ed the hotkey for using "Say" verb will not be able to talk, thus being mute until a staff member or otherwise tells them how to fix this issue.
Not a huge problem as it can be fixed in-game, but because of how often it happens it is noticeable.

Hathkar:
Reverts tgstation/TerraGov-Marine-Corps#4512

Almost every round we have people asking how to talk IC, since this PR broke the default hotkeys.

Fixes #4694

### DIFF
--- a/code/controllers/subsystem/input.dm
+++ b/code/controllers/subsystem/input.dm
@@ -22,6 +22,10 @@ SUBSYSTEM_DEF(input)
 	macro_set = list(
 	"Any" = "\"KeyDown \[\[*\]\]\"",
 	"Any+UP" = "\"KeyUp \[\[*\]\]\"",
+	"O" = "ooc",
+	"L" = "looc",
+	"T" = ".say",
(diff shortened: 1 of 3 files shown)


spookydonut:
It didn't break it, if you reset keybindings it'll work or add t for talk yourself.
I wasn't about to force a keybindings reset for everyone since that would cause more issues, for new players it works fine.

Hathkar:
The issue is we have to constantly tell people almost every single round to reset their keybindings.

Chaznoodles:
I'd rather a forced keybind reset for everyone and not have any further issues, than having to consistently answer mhelps about a function that should work out the box, but doesn't.

PR #4731, opened 2020-08-23
Razorwire can no longer be healed when it is at full/cleanup

TiviPlus:
## About The Pull Request

Title, change stuff to use welder_act and so instead of attackby

## Why It's Good For The Game

yes

## Changelog
:cl:
fix: Razorwire can no longer be repaired when it's at full health
code: cleaned up razorwire code a bit
/:cl:


jroinc:
I thought razorwire wasn't supposed to be fixable at all?

PR #4747, opened 2020-08-30
Newlights

TiviPlus:
## About The Pull Request
Element opacity + overlay component lights

Definitely needs testing

https://github.com/tgstation/tgstation/pull/52881
https://github.com/tgstation/tgstation/pull/52574
https://github.com/tgstation/tgstation/pull/52413
https://github.com/tgstation/tgstation/pull/53317
## Why It's Good For The Game
Removes like  at least 70% of lighting lag, allows for directional lights and FOV cones in future
It has its drawbacks but the performance boost is insane
## Changelog
:cl:
refactor: Lighting lag is dead
add: bullets now glow
balance: most marine armor and rail lights nerfed, flares buffed
/:cl:


spookydonut:
Did you properly port the signal handler stuff?

sunofang:
any optimization is good optimization

Hughgent:
I think that this is `byond` me so just throw it at the server in testing and fix bugs as they come.

XSlayer300:
Flare boxes somehow glow even though you did not pull one out, and the light _literally emits from the on-hand areas_ instead of the person. Also, pulse bolts have the wrong color for their lights.

BOBAMAx:
Flares emit light from inside of containers when not lit, and objects that emit light that are in a player's inventory don't emit light from the player, but where that item is in the players inventory, so flares in your right pocket emit light on the bottom right hand corner of the screen

MiniMeatwad:
Whenever you get armor from a vendor, it appears in your hand first and doesn't exist in the world at any point, and instead of emitting light from your character, it creates light from your armor slot instead.

MiniMeatwad:
Could the light rebalancing be moved to a separate PR? Newlights is awesome, but I think the light balance change affects gameplay enough that it's not a small change.

XSlayer300:
ERT armor light values are unchanged, and because of that, it makes the ERT lights nerfed. Can you make them equal to the exoskeleton/vest?

XSlayer300:
... ERT liiiiiights are not fixeddddd.

PR #4760, opened 2020-09-02
Nerf's ravager charge CD and raises the plasma cost

2whatever2:
## About The Pull Request
Raises the cool down on ravagers charge and raises the plasma cost (6 seconds to 10, 250 plasma to 300 plasma)

## Why It's Good For The Game
Ravager got a health buff of 50+ so their engage/disengage ability should be tweaked

## Changelog
:cl:
tweak: Raises the cool down on ravagers charge and raises the plasma cost (6 seconds to 10, 250 plasma to 300 plasma)
/:cl:

### DIFF
--- a/code/game/objects/structures/orbital_cannon.dm
+++ b/code/game/objects/structures/orbital_cannon.dm
@@ -377,7 +377,7 @@
 
 /obj/structure/ob_ammo/warhead/incendiary/warhead_impact(turf/target, inaccuracy_amt = 0)
 	var/range_num = max(15 - inaccuracy_amt, 12)
-	flame_radius(range_num, target,	burn_intensity = 36, burn_duration = 40, colour = "blue")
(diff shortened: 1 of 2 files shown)


jroinc:
I recommend that in addition to this, you also add a minimal delay (0.1 seconds/1 tick) to both abilities, so it can't be used on the fly at a run.

2whatever2:
> I recommend that in addition to this, you also add a minimal delay (0.1 seconds/1 tick) to both abilities, so it can't be used on the fly at a run.

As a thing i like to say a lot when talking about tier 3's having advantages is, "It's a tier 3 so they get privileges" so i'd rather keep it without a delay (_even though defiler has a delay on his ability but don't tell anyone that_)

TiviPlus (approved):

PR #4783, opened 2020-09-06
Makes regenning walls slightly less bad

TiviPlus:
## About The Pull Request

Puts them not on ssobj and instead on a slower processing ss so they dont clog up ssobj
also checks that timers arent spammed

## Changelog
:cl:
code: regenning walls process better
/:cl:


Hughgent (approved):
I like it

PR #4786, opened 2020-09-07
Burger's Attempt at a Xeno Regeneration Nerf

BurgerLUA:
## About The Pull Request

When you take damage as a xenomorph, you cannot regenerate your health PASSIVELY for 10 seconds. A regen aura bypasses this.
When you can regenerate, your regeneration powers start off at 0. It increases by 5% every 1 second, up to 100%. A regen aura changes this to bypass the initial 10 second cooldown, and increases it by 7.5% every 1 seconds. Note that the 1 health per 2 seconds that you gain is not affected by this.

## Why It's Good For The Game

As a returning player, I noticed that Xenos were buffed like crazy when it comes to passive health regeneration. While this does improve the flow of gameplay a certain amount, it creates a new problem where Xenomorphs can shrug off insane amounts of damage after encounters like it was nothing, resulting in xenos charging in mindlessly and then running away after they take 50% of their max health, only to regenerate it after a few seconds and repeat the process.

This PR aims to make it so that it is no longer as easy to do this. If you take significant damage, you must actually wait some time before returning to perform hit and run attacks instead of shrugging it off 5 seconds later. I've played several Xeno rounds and noticed that it is super easy to make attacks on marines, flee for 4 seconds while resting, and then get back up again. This PR also makes it so that it is faster to regenerate 75% of your health once than regenerating 25% of your health 3 times with the scaling system.

## Changelog
:cl: BurgerBB
balance: Taking damage as a Xenomorph disables PASSIVE regeneration for 10 seconds. When the 10 seconds are up, you regenerate your health at 0% strength, but it increases by 5% every second, up to 100%.
balance: A regeneration aura changes this behavior by disabling the 10 second cooldown, and regenerating your health by  0% strength, but it increases by 7.5% every second, up to 100%.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
@@ -32,6 +32,10 @@
 	// *** Speed *** //
 	var/speed = 1
 
+	// *** Regeneration Delay ***//
+	var/regen_delay = 10 SECONDS //Time after you take damage before you can regen.
+	var/regen_ramp_amount = 0.005 //Regeneration power increases by this amount evey decisecond.
+
 	// *** Plasma *** //
 	var/plasma_max = 10
 	var/plasma_gain = 5
@@ -176,7 +180,10 @@
 	var/warding_aura = 0
 	var/recovery_aura = 0
 
-	var/regen_power = 0 //Increases by 0.1 every time you regen. Resets when you take damage.
+	var/regen_power = 0 //Resets to -xeno_caste.regen_delay when you take damage.
+	//Negative values act as a delay while values greater than 0 act as a multiplier.
+	//Will increase by 10 every decisecond if under 0. Increases by xeno_caste.regen_ramp_amount every decisecond.
+	//If you want to balance this, look at the xeno_caste defines mentioned above.
 
 	var/is_zoomed = 0
 	var/zoom_turf = null
--- a/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
@@ -176,6 +176,8 @@
 	var/warding_aura = 0
 	var/recovery_aura = 0
 
+	var/regen_power = 0 //Increases by 0.1 every time you regen. Resets when you take damage.
+
 	var/is_zoomed = 0
 	var/zoom_turf = null
 	var/attack_delay = 0 //Bonus or pen to time in between attacks. + makes slashes slower.
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -133,7 +133,7 @@
 	if(updating_health)
 		updatehealth()
 
-	regen_power = 0
+	regen_power = -5 //This will take 10 seconds to kick in as I have been told that 1 life tick is 2 seconds.
 
 	return damage
 
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -132,6 +132,9 @@
 
 	if(updating_health)
 		updatehealth()
+
+	regen_power = 0
+
 	return damage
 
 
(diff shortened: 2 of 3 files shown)


tgstation-server:

ilikesaying:
is burger going to race pariah to the -100 pr delta

BurgerLUA:
> is burger going to race pariah to the -100 pr delta

yes

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -176,6 +176,8 @@
 	var/warding_aura = 0
 	var/recovery_aura = 0
 
+	var/regen_power = 0 //Increases by 0.1 every time you regen. Resets when you take damage.
```
psykzz, line 179: Since it can be annoying to find the actual value (-5) can you reverse these? So it starts at -5 and then resets to 0 on damage?

BurgerLUA: It needs to be -5 as that controls the cooldown of it. If you want, I can add a define.

psykzz: I think you misunderstood, im happy with -5 for timing im just suggesting that you invert the logic so that finding or tuning the value would always happen here instead of in the `take_damage` proc.

TiviPlus:
```
code/modules/mob/living/carbon/xenomorph/life.dm:95:/mob/living/carbon/xenomorph/proc/heal_wounds(multiplier = XENO_RESTING_HEAL, var/scaling = FALSE)
changed files contains proc argument starting with 'var'
```
travis angry

BurgerLUA:
yep, don't think this will get immediately merged without more feedback so I'll fix it later

BurgerLUA:
I cut the regeneration power by half after some testing. I also improved the code and the customizability by tying the regeneration cooldown delay as well as regeneration scaling per second to the xeno caste datum.

Denthamos:
I like the idea of this as it will promote more drones/hivelords playing the field and actively healing hurt xenos instead of just sitting back and being relegated to hive building.

DerFlammenwerfer:
> I like the idea of this as it will promote more drones/hivelords playing the field and actively healing hurt xenos instead of just sitting back and being relegated to hive building.

or weed spamming

DerFlammenwerfer:
> A regeneration aura changes this behavior by disabling the 10 second cooldown
Would it be possible to make recovery pheromones reduce it to 5 seconds instead for weak pheros and 2.5 for strong ones rather than bypass it entirely? Also, the source of pheromones probably should not have this bypass applied to itself.

Hughgent (approved):
If you would kindly address some people concerns and suggestions please.

TiviPlus (approved):
Worked well in testing so Ill merge soon

Marmio64:
This would let marines actually make pushes now at least. If there's enough xenos and a queen who has made them leaders with regen pheros, often what would happen is: Xenos charge, you injure most of them to <50% health, and they run away to heal. You charge after a few seconds of preparing, and the xenos you just hurt are already full health and ready to repel you.

Also, secret buff to drones/shrikes.

Marmio64:
This would let marines make pushes far more often now, which is good. If there's enough xenos and a queen who has made them leaders with regen pheros, often what would happen is: Xenos charge, you injure most of them to <50% health, and they run away to heal. You charge after a few seconds of preparing, and the xenos you just hurt are already full health and ready to repel you.

Also, secret buff to drones/shrikes.

BurgerLUA:
> > A regeneration aura changes this behavior by disabling the 10 second cooldown
> 
> Would it be possible to make recovery pheromones reduce it to 5 seconds instead for weak pheros and 2.5 for strong ones rather than bypass it entirely? Also, the source of pheromones probably should not have this bypass applied to itself, though the aura should still benefit its rate of healing otherwise.

I think the way it is is currently fine. If maintainers think otherwise I guess I'll change?

TiviPlus (approved):
Yea this is working decent if you want more tweaks theres always another PR

tgstation-server:
Your Fix/Feature pull request delta is currently below zero (-4). Maintainers may close future Feature/Tweak/Balance PRs. Fixing issues or helping to improve the codebase will raise this score.

PR #4806, opened 2020-09-12
Razorwire HP nerf

jroinc:
## About The Pull Request
Razorwire HP lowered from 200 to 100.

## Why It's Good For The Game

Plasteel is the top tier of fortifications. It has 600 HP +50 with wire, and can be welded to fix it, unless it's severely damaged.
Sandbags are 1 tier below that. They have 400 HP +50 with wire. They cannot be welded to be fixed, but can be repaired with sandbags.
Metal cades are 1 tier below that. They have 200 HP +50 with wire, and can be welded to fix unless severely damaged.

Razorwire is 1 tier below that. It has the same HP as metal (200), cannot be slashed, pounced, or charged without damage/trapping the target, can be repaired from ANY HP, AND can be electrified to make slashing it impossible.

Lowering the HP to 100 keeps the same progression and allows a crusher at full-speed to actually take one down (full speed crusher deals 104 damage to wire) rather than be useless. Note that non-full speed crushers will still be trapped, and electrified wires are still immune to non-crusher melee. 

## Changelog
:cl:
balance: Lowered razorwire HP from 200 to 100.
/:cl:

### DIFF
--- a/code/__DEFINES/objects.dm
+++ b/code/__DEFINES/objects.dm
@@ -147,7 +147,7 @@ GLOBAL_LIST_INIT(restricted_camera_networks, list( //Those networks can only be 
 #define RAZORWIRE_BASE_DAMAGE		40
 #define RAZORWIRE_ENTANGLE_DELAY	5 SECONDS
 #define RAZORWIRE_SOAK				5
-#define RAZORWIRE_MAX_HEALTH		200
+#define RAZORWIRE_MAX_HEALTH		100
 #define RAZORWIRE_SLOWDOWN			10
 #define RAZORWIRE_MIN_DAMAGE_MULT_LOW	0.4 //attacking
 #define RAZORWIRE_MAX_DAMAGE_MULT_LOW	0.6


Chaznoodles:
An alternative or addition to this would be to make it unrepairable.

lovegreenstuff:
Or make it impossible to place behind cades, like by placing it on your current tile. If memory serves correct, the crusher charge not one-shotting razorwire is intended (although not clearly stated).

jroinc:
> An alternative or addition to this would be to make it unrepairable.

It would be an alternative, but at 3 metal a razorwire (2.25 of which is recoverable) means that you're just paying a tiny, sustainable, price to remove any chance for a non-spitting caste to breach ever. And the fact that it entangles crushers (AKA kills if there's anyone on any of the barricades) is an OOF.

jroinc:
> Or make it impossible to place behind cades, like by placing it on your current tile. If memory serves correct, the crusher charge not one-shotting razorwire is intended (although not clearly stated).

That just delays the problem (because you can just place it outside under flamer cover if it's really that good to have).

I know, but I'm changing it so it does. That was before ewire was a thing, and having razorwire that literally cannot be attacked except by spit that takes 3 seconds and 3 metal to replace makes fortifications unbreachable. This just opens up a new way to attack.
It still won't one-shot metal cades, though, even at full momentum.

lovegreenstuff:
*cocks glock*
Always hasn't been.
*bang*

GravitronBlue:
Since I didn't see the PR specifically mention it, will electrified wire still stop a full speed crusher?

Hughgent (approved):
variable change

jroinc:
> Since I didn't see the PR specifically mention it, will electrified wire still stop a full speed crusher?

No. Electrified wire will stop anything BUT a full-speed crusher. I have no intention of making ewire better than it is because ewire's already very, very good (ayys literally cannot melee attack it, it stuns them and does no damage).

SplinterGP:
i mean xenos can attack razorwires even if electrified, they just have to regen after a time as they get hurt from doing so.

jroinc:
> i mean xenos can attack razorwires even if electrified, they just have to regen after a time as they get hurt from doing so. unless that causes no damage to the wire

It causes no damage to the wire. It just stuns and hurts them.

tgstation-server:
Your Fix/Feature pull request delta is currently below zero (-1). Maintainers may close future Feature/Tweak/Balance PRs. Fixing issues or helping to improve the codebase will raise this score.

Issue #4796, opened 2020-09-10
Able to evolve from a larva to tier 3/ Speeds up evo faster than normal


2whatever2:
## Testmerges:
#4783 #4763 #4727 #4740 #4487 #4786

## Reproduction:
1. Evo into any caste
2. Open up the menu for your next evo
3. Regress into a larva
4. Evo from a larva to a tier 2 without waiting 120 seconds
5. Repeat the last 4 steps to become a tier 3
Video Example:
https://discordapp.com/channels/498569646014857217/500801682058379265/753580513524187162

Chaznoodles:
I think this might bypass tier evo restrictions too, unsure.

PR #4829, opened 2020-09-18
Fixes Xeno Larva evolving to Tier 2+ castes by abusing regression. The reboot part IV.

Wayland-Smithy:
## About The Pull Request

Adds a `xeno_caste.evolves_to` sanity check to `do_evolve()` code. Has the derpy negative forced check also so we can run it pre special role (queen, shrike, hivemind) logic as they are also exploitable if we only use the last else block.

Fixes: #4796 

## Why It's Good For The Game

Exploit bad. Sanity checks good.

## Changelog
:cl:
fix: Xenos can no longer evolve to the wrong castes after regressing.
/:cl:


Hughgent (approved):

TiviPlus (approved):
Clean fix

PR #4886, opened 2020-09-25
Fix Xeno Hivemind overriding skill targeting by removing camera mob mouse_opacity override

Wayland-Smithy:
## About The Pull Request

Very simple one line deletion which will allow the byond `mouse_opacity` var to remain its default value of `1` which will both allow right-clicking for admins to easily access PP and XP etc.. while preventing it from having skills like Crusher: Crest Toss wasted on flinging around the Hiveminds AI-Eye because that is what a value of `2` does while a value of `0` makes it not appear in the context menu for the tile or be clickable at all.

@psykzz hopefully this gets your attention now, please see https://discordapp.com/channels/498569646014857217/498576390753615882/758585661523230721

## Why It's Good For The Game

Bug bad. Fix good. Having an ability cast stolen because the Hivemind did not know it was supposed to avoid placing itself on-top of targets is not fun for anyone involved on the side of the xenos.

No more:
OnE5j3YYho

## Changelog
:cl:
fix: The Hivemind camera mob no longer steals left-clicks from mobs/objects on the same tile. 
admin: Hivemind camera mobs can still be accessed from the right-click menu (while a ghost) for ease of administration.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
@@ -20,7 +20,6 @@
 	tier = XENO_TIER_ZERO
 	upgrade = XENO_UPGRADE_ZERO
 
-	mouse_opacity = MOUSE_OPACITY_OPAQUE
 	lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE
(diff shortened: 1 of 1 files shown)


Marmio64:
so when we fixing hivemind being able to place resin without any resin nearby

Hughgent (approved):

PR #4894, opened 2020-09-26
Distress Victory Conditions

Hathkar:
## About The Pull Request

Marine Major: Kill all xenos on the planet
Xeno Major: Kill all marines
Xeno Minor: Hijack the Alamo but die on the main ship
Draw: Self Destruct blows up

## Why It's Good For The Game

Some of the old win conditions were nonsensical. Xenos got a minor victory if every marine died on the planet and none were left on the ship.

Marines can no longer evac planetside to try and cheese a victory on the ship. Losing groundside means you've given up hope of victory, and the best you can hope for is a draw if you blow SD. Evac is now a last resort.

Groundside combat is the main focus of the game, and the win conditions should reflect that.

## Changelog
:cl:
fix: Fixed case where a xeno major was instead being recorded as a xeno minor
/:cl:

### DIFF
--- a/code/datums/gamemodes/distress.dm
+++ b/code/datums/gamemodes/distress.dm
@@ -123,7 +123,7 @@
 			message_admins("Round finished: [MODE_INFESTATION_X_MAJOR]") //xenos wiped our marines, xeno major victory
 			round_finished = MODE_INFESTATION_X_MAJOR
 			return TRUE
-		message_admins("Round finished: [MODE_INFESTATION_X_MINOR]") //xenos wiped out ALL the marines without hijacking, xeno major victory
+		message_admins("Round finished: [MODE_INFESTATION_X_MAJOR]") //xenos wiped out ALL the marines without hijacking, xeno major victory
 		round_finished = MODE_INFESTATION_X_MAJOR
 		return TRUE
 	if(!num_xenos)
--- a/code/datums/gamemodes/distress.dm
+++ b/code/datums/gamemodes/distress.dm
@@ -111,27 +111,27 @@
 	var/num_xenos = living_player_list[2]
 
 	if(SSevacuation.dest_status == NUKE_EXPLOSION_FINISHED)
-		message_admins("Round finished: [MODE_GENERIC_DRAW_NUKE]")
+		message_admins("Round finished: [MODE_GENERIC_DRAW_NUKE]") //ship blows, no one wins
 		round_finished = MODE_GENERIC_DRAW_NUKE
 		return TRUE
 	if(!num_humans)
 		if(!num_xenos)
-			message_admins("Round finished: [MODE_INFESTATION_DRAW_DEATH]")
+			message_admins("Round finished: [MODE_INFESTATION_DRAW_DEATH]") //everyone died at the same time, no one wins
 			round_finished = MODE_INFESTATION_DRAW_DEATH
 			return TRUE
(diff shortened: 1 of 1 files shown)


code/datums/gamemodes/distress.dm
```diff
@@ -111,27 +111,27 @@
...
 			round_finished = MODE_INFESTATION_DRAW_DEATH
 			return TRUE
 		if(round_stage == DISTRESS_DROPSHIP_CRASHED)
-			message_admins("Round finished: [MODE_INFESTATION_X_MAJOR]")
+			message_admins("Round finished: [MODE_INFESTATION_X_MAJOR]") //xenos wiped our marines, xeno major victory
 			round_finished = MODE_INFESTATION_X_MAJOR
 			return TRUE
-		message_admins("Round finished: [MODE_INFESTATION_X_MINOR]")
-		round_finished = MODE_INFESTATION_X_MINOR
+		message_admins("Round finished: [MODE_INFESTATION_X_MINOR]") //xenos wiped out ALL the marines without hijacking, xeno major victory
```
spookydonut, line 126: change this to major also

Marmio64:
if you're going to make distress more of a pain to win on as marines, can you make crash be less xeno sided... Maybe make the disks print faster, since xenos get free respawns? Or, I don't know if it has been changed yet, but I remember any xeno is able to disarm the nuke instantly, perhaps only the queen should be able to do that, and shrike has a slight delay.

Marmio64:
Also, doesn't it seem kind of dumb for xenos to get a minor victory for dying shipside? If marines arm the SD, but then beat back the xenos before it blows up, xenos get a victory for that? Should probably be made a draw instead, since the xenos LOST, they didn't win, they died and failed, marines also failed to protect their ship, so it'd be a DRAW since both sides lost.

XSlayer300:
So wait, there is no marine minor anymore, when the xenos die off in groundside because of hivemind collapse?

TiviPlus:
The only changes here are 
-xenos wiped out ALL the marines without hijacking, xeno major victory    being a major and 
-Xenos dying shipside being a xeno instead of marine minor

This is necessary in order to allow the focus on groundside + encourage focusing groundside and to make the absolutely imbalanced shipside stage not matter as much for the winrate.

if xenos begin to mop the floor with marines marines will recieve a buff for their equipment as crash is heavily xeno favored, though current distress winrate is 40% xenos so who knows if thats even necessary

code/datums/gamemodes/distress.dm
Hathkar: Done, nice catch.

Denthamos:
The winrate being 40% for xenos, is that including when marines when on shipside defense after they get pushed off planetside?  Because currently the only thing that really tends to balance out the scales for planetside vs shipside is xenos should not *want* to just keep marines in the FOB constantly and let them push out, in order to collapse and take them on, no?

With this it would emphasize FOB camping even harder and never letting marines to push out, making them think they are loosing the fight and want to evacuate the planetside fight faster as no matter what at that point, xenos win.  So why stick around and planetside and put up a fight you are going to loose anyways and end the round instead.

spookydonut:
This doesn't change win conditions this just fixes an error in what a rare end-game state counts as.

MiniMeatwad:
I think marine minor for killing all xenos on the ship would make more sense, and a draw for SD.

PR #4899, opened 2020-09-26
Fixes screenwide red xenos

TiviPlus:
## About The Pull Request

No more red beno  swarm

## Changelog
:cl:
fix: Fixes screenwide red xenos
/:cl:


jroinc:
You should probably also fix the fact that the QUENJEL doesn't delete upon use.

So, for instance, if I use it, it's now still in my hand.

It also doesn't delete (or at least not properly) ON EXPIRATION after 12 seconds.

So I can apply NOTHING to other xenos. And it blocks slashing things unless you drop the nothingness from your hands.

TiviPlus:
It does fix that

Hughgent (approved):

PR #4905, opened 2020-09-26
Fixes the big red "You are either AFK.." message showing when you are in fact not AFK and/or disconnected

Wayland-Smithy:
## About The Pull Request

Brings back the `Ping` subsystem with a slight change to just send one way heartbeats to the client.

Now the message will only be displayed if you actually lose connection to the server for some reason.

## Why It's Good For The Game

Bug bad. Fix good.

## Changelog
:cl:
fix: Fixed goonchat showing the red "You are AFK" message while you are connected and playing.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ping.dm
+++ b/code/controllers/subsystem/ping.dm
@@ -1,0 +1,32 @@
+SUBSYSTEM_DEF(ping)
+	name = "Ping"
+	priority = FIRE_PRIORITY_PING
+	wait = 15 SECONDS
+	flags = SS_NO_INIT
+	runlevels = RUNLEVEL_LOBBY|RUNLEVEL_SETUP|RUNLEVEL_GAME|RUNLEVEL_POSTGAME
+
+	var/list/currentrun = list()
+
+/datum/controller/subsystem/ping/stat_entry()
+	..("P:[length(GLOB.clients)]")
+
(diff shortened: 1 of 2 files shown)


code/controllers/subsystem/ping.dm
```diff
@@ -0,0 +1,32 @@
...
+	//cache for sanic speed (lists are references anyways)
+	var/list/currentrun = src.currentrun
+
+	while(length(currentrun))
+		var/client/C = currentrun[length(currentrun)]
+		currentrun.len--
+
+		if(!C?.chatOutput?.loaded)
+			if(MC_TICK_CHECK)
+				return
```
TiviPlus, line 27: Wouldnt it be better to have a global for loaded chats instead of checking whether each client has a loaded chat?

Wayland-Smithy: I am not sure, this is how it did it before.

TiviPlus: on second thought this might be better

TiviPlus (approved):

code/controllers/subsystem/ping.dm
Wayland-Smithy: It just works.

PR #4911, opened 2020-09-27
Makes some weather slightly quieter

TiviPlus:
## About The Pull Request

less loud

Imagine complaining and not just doing a oneline change
## Changelog
:cl:
tweak: Makes some weather slightly quieter
/:cl:


neohenry:
I know that this is a different request:

Is it okay if the HMG's volume decrease? It's quite painful to the ears, especially when there are multiple HMGs.

Hughgent (approved):

PR #4916, opened 2020-09-27
Changes boiler zoom offset to be the same as the railscope.

Eskjjlj:
## About The Pull Request
Changes the offset of the boiler zoom to make it the same as the railscope.

## Why It's Good For The Game
Marines with railscopes could shoot boilers from farther than the boilers could see causing boilers to be useless since there is always a handful of marines with railscopes. Now the playing field is even as boilers and snipers have the same range of vision while zoomed.

## Changelog
:cl:
tweak: Boilers can now see farther zoomed at the cost of not being able to see behind them.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
@@ -20,7 +20,7 @@
 			"<span class='notice'>We start focusing your sight to look off into the distance.</span>", null, 5)
 		if(!do_after(X, 1 SECONDS, FALSE, null, BUSY_ICON_GENERIC) || X.is_zoomed)
 			return
-		X.zoom_in()
+		X.zoom_in(11)
 		..()
 
 // ***************************************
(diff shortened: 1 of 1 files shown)


SplinterGP:
need to see if it doenst causes boilers to be too much strong, seeing as benos have full vision they can spot marine ambushes and all easily with that, other thing is that snipers can't see on dark(except IFF sniper rifle that has NVG), so that should be taken into account.

Eskjjlj:
Boilers emit light though so a sniper can always see them. And it sucks when you cant do your job at all because of one funny marine with a sniper rifle shooting you from offscreen as soon as you peek out of cover

Kuro-ai:
First,one defender or other tanky caste can just stand in front of you and now snipers can't do anything,second,if you have globs stacked and are just standind in the open in front of the FOB you should get shot by snipers,that's their job

Kuro-ai:
Nevermind the fact that snipers can't see in the dark or can shoot past marines (no IFF),so unless you're running around with globs stacked in the open in front of the fob with no defender or other tanky caste to tank shots for you,boiler can attack the fob freely

SplinterGP:
that also without counting once defenders get a fix they'll be even more able to defend boilers.

TiviPlus (approved):

PR #4952, opened 2020-10-04
Fixes many ERT radios getting the ALAMO freq, and gives them the ability to work independent of tcomms

Wayland-Smithy:
## About The Pull Request

Makes `freerange = TRUE` for all `/obj/item/radio/headset/distress` allowing them to bypass the frequency sanity check that was clamping them to 1441.

(Re) Adds `independent` radio frequencies based on encryption keys for ERTs only. 

## Why It's Good For The Game

Better ERT communication.

## Changelog
:cl:
add: Granted most ERT headsets the ability to operate independent of tcomms. (Sectoid, SOM, PMC, etc..)
fix: Fixes ERT headsets getting access to the ALAMO frequency and it set as their main channel.
/:cl:


code/game/objects/items/radio/encryptionkey.dm
```diff
@@ -84,36 +85,43 @@
...
 /obj/item/encryptionkey/som
 	name = "\improper Sons of Mars encryption key"
 	channels = list(RADIO_CHANNEL_SOM = TRUE)
+	independent = TRUE
 
 
 /obj/item/encryptionkey/sectoid
 	name = "\improper alien encryption key"
 	channels = list(RADIO_CHANNEL_SECTOID = TRUE)
+	independent = TRUE
```
Wayland-Smithy, line 124: Maybe some of these should not be independent, or others like Echo Task Force should be?

XSlayer300:
Thank you so much for this PR, I made the one for the common radio, but I wasn't able to fix it.

Hughgent (approved):

TiviPlus (approved):

PR #4959, opened 2020-10-06
buckshot nerf

Pariah919:
## About The Pull Request
nerfs buckshot to 160 total damage if I have correct math by reducing the amount of projectiles to 1+4 and nerfing the bonus projectiles to 30 damage
removes the stagger from buckshot

## Why It's Good For The Game
buckshot is broken right now apparently, though I'd argue if we should even keep it since balancing all 3 shotgun rounds has always been basically an annoyance, i won't remove it yet though 

## Changelog
:cl:
balance: buckshots max damage reduced from 240 to 160, removed stagger
/:cl:


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -51. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

RenGusta:
Is it at least a relatively tight spread in practice without excessive falloff, making it reasonably powerful instead of a joke?

Gooboi:
I think at this rate it would just be appropriate to straight up remove buckshot instead of just nerfing damage since it has 0 AP anyway, this change would effectively make it "flechette but worse in every way".

Pariah919:
buckshot bad

PR #4961, opened 2020-10-06
Removes the chance for EORD Respawn verb to turn dead into xenomorphs, and prevents SSD Xenos searching for a mind pre/post round

Wayland-Smithy:
## About The Pull Request
Inspired by Snow mentioning it, and seeing many many catatonic resting xenos during EORD and finding out why that is.

Removes the random chance (weighted as it was) of becoming one of a few low xeno castes when using the verb.

Adds a check to `xenomorph/on_sdd_grace_period_end()` preventing it from searching for and shoving found candidates into xeno bodies pre and post round.

## Why It's Good For The Game

Xeno mob spam in EORD is bad. Getting shoved into one of those Xenos during EORD, because you had those settings enabled for the round (and it also does not care about your EORD preference to boot), while it is crit because someone got bad RNG and wanted to be a human so they ghosted and re-rolled is just as bad if not worse. Quality of life improvements are good.

Brings the EORD Respawn verb in-line with the EORD gamemode setup logic which makes all dead participants human only.

## Changelog
:cl:
tweak: The EORD Respawn verb now always respawns you as a Human.
fix: Fixes ghosts getting shoved into SSD Xenos before and after the round, like during EORD.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -463,6 +463,8 @@
 		return
 	if(client)
 		return
+	if (SSticker.current_state != GAME_STATE_PLAYING)
+		return
 
 	var/mob/picked = get_alien_candidate()
 	if(!picked)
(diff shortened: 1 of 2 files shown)


Wayland-Smithy:
@jroinc feel free to make a second verb for spawning as a Xeno, or figure out a better way to have both be possible other than pure RNG.

TiviPlus (approved):

PR #4965, opened 2020-10-06
Weeds speed up xenos rather than slowing down marines

TiviPlus:
## About The Pull Request

Title

## Why It's Good For The Game
Should encourage pushing without nerfing xenos?

## Changelog
:cl:
balance: Weeds speed up xenos rather than slowing down marines
/:cl:


code/game/objects/effects/weeds.dm
```diff
@@ -53,9 +53,9 @@
 
 /obj/effect/alien/weeds/Crossed(atom/movable/AM)
 	. = ..()
-	if(ishuman(AM))
-		var/mob/living/carbon/human/H = AM
-		H.next_move_slowdown += 1
+	if(isxeno(AM))
+		var/mob/living/carbon/xenomorph/X = AM
+		X.next_move_slowdown -= 0.4
```
Wayland-Smithy, line 58: I like the idea of the change it reminds me of zerg, but it looks like when both are on weeds the slowdown delta in favor of xenos on weeds is only 40% what is was, am I reading this correctly?

SplinterGP: apparently is, which is not so bad tbh, it will help slower castes and faster castes wont win more much, TM would be very neccessary

RenGusta:
@Wayland-Smithy

It's a -0.4 slowdown mod on Xenos, which is close to an additional max level frenzy ( just a little short ) for when they're on weeds. It's not a % modifier for their speed. 

I think this will be interesting to see in practice.

neohenry:
Marines will definitely chase now that weed don't slow them down. This makes them more susceptible to resin traps, which is the bane of marines. Given that it's a habit of marines to break weed, I think marines will continue to break them until the habit has a reason to decrease in frequency.

Xenos will also now think of using sticky resin to slow down marines.

ilikesaying:
sticky chess resin time

Hughgent (approved):

jroinc:
Only concern I have is this is very strong on crushers and bulls. They're so *fast*

davidchan13:
> Only concern I have is this is very strong on crushers and bulls. They're so _fast_

Boomerlord just became the zoomerlord.

David-Stormwell:
Oh god oh fuck I'm getting PTSD from this

TacticalSpacePilot:
Can we see a video of this in action just to see

Pariah919:
>balance: Weeds speed up xenos rather than slowing down marines

Remove or make the cooldown for hiveminds weed planting ~30-45 seconds rather than 12 again. As hivemind makes map control by weed removal for marines literally impossible.

PR #4982, opened 2020-10-09
razorwire rebalancing (github please stop messing up)

MacBlaze1:
## About The Pull Request

this PR makes razorwire effective again, with crushers at max speed dealing 94.5 damage BEFORE any effects from the razorwire are applied, meaning that one acid spit from a spitter or praetorian hits this then the crusher charges, it will destroy the razorwire without consequences to the crusher. Razorwire cannot be repaired at all, and you can now attack it even if electrified. It deals a bit more damage to compensate for the non electrifying and non repairing.  

## Why It's Good For The Game

this makes razorwire usuable again, but without the inability to do its main function. Marines now have to use metal to maintain a razorwire shield effectively. There is no BS weld repairing infinity. Better hope its in good enough shape and when it breaks repair it fast.

## Changelog
:cl:
del: Removed the ability to repair razorwire with a welder
balance: crushers at full charge now deal 94.5 damage to razorwire, instead of 105. The razorwire still has 100 health.
balance: you can now attack razorwire even when it is electrified.
balance: you take a bit more damage from sharp parts of the razorwire when attacking, now you take 24 to 32 damage instead of 16 to 24 damage.
balance: one spit from a spitter or praetorian or 2 mature boiler acid globs are enough to allow a crusher to charge through it
balance: big xenos do not get stunned when attacking electrified razorwire, other xenos now get stunned for only 4 seconds, down from 8.
/:cl:

### DIFF
--- a/code/game/objects/structures/razorwire.dm
+++ b/code/game/objects/structures/razorwire.dm
@@ -121,19 +121,14 @@
 
 	if(istype(I, /obj/item/stack/sheet/metal))
 		var/obj/item/stack/sheet/metal/metal_sheets = I
-
-		visible_message("<span class='notice'>[user] begins to repair  \the [src].</span>")
-
+		visible_message("<span class='notice'>[user] begins to repair \the [src].</span>")
 		if(!do_after(user, 2 SECONDS, TRUE, src, BUSY_ICON_FRIENDLY) || obj_integrity >= max_integrity)
 			return
-
 		if(!metal_sheets.use(1))
 			return
-
 		repair_damage(max_integrity * 0.30)
 		visible_message("<span class='notice'>[user] repairs \the [src].</span>")
 		update_icon()
-
 		return
 
 	if(!istype(I, /obj/item/grab))
--- a/code/game/objects/structures/razorwire.dm
+++ b/code/game/objects/structures/razorwire.dm
@@ -119,6 +119,22 @@
 /obj/structure/razorwire/attackby(obj/item/I, mob/user, params)
 	. = ..()
 
+	if(istype(I, /obj/item/stack/sheet/metal))
+		var/obj/item/stack/sheet/metal/metal_sheets = I
+
+		visible_message("<span class='notice'>[user] begins to repair  \the [src].</span>")
+
+		if(!do_after(user, 2 SECONDS, TRUE, src, BUSY_ICON_FRIENDLY) || obj_integrity >= max_integrity)
+			return
+
+		if(!metal_sheets.use(1))
+			return
+
+		repair_damage(max_integrity * 0.30)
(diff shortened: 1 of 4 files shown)


Hughgent (approved):
code looks ok

Daftspy:
Razorwire isnt good but this seems worse.

MacBlaze1:
> Razorwire isnt good but this seems worse.

attacking the razorwire when electrified will make you take a shitton of damage 
before this, crushers would instantly destroy razorwire, making it useless
razorwire is incredibly cheap to make and build, so I feel this price for literally hard countering a crusher isnt too bad

jroinc:
Seems good to me. The one thing I'd change is to allow razorwire to be repaired in some way.

Allow welding but double the weld timer, add 2 metal rods, something similar.

Just cause my OCD's gonna bug the shit outta me at the FOB.

code/game/objects/structures/razorwire.dm
```diff
@@ -119,8 +119,26 @@
...
+			return
+
+		if(!metal_sheets.use(1))
+			return
+
+		repair_damage(max_integrity * 0.30)
+		visible_message("<span class='notice'>[user] repairs \the [src].</span>")
+		update_icon()
+
+		return
```
TiviPlus, line 137:
```suggestion
	if(istype(I, /obj/item/stack/sheet/metal))
		var/obj/item/stack/sheet/metal/metal_sheets = I
		visible_message("<span class='notice'>[user] begins to repair \the [src].</span>")
		if(!do_after(user, 2 SECONDS, TRUE, src, BUSY_ICON_FRIENDLY) || obj_integrity >= max_integrity)
			return
		if(!metal_sheets.use(1))
			return
		repair_damage(max_integrity * 0.30)
		visible_message("<span class='notice'>[user] repairs \the [src].</span>")
		update_icon()
		return
```

code/game/objects/structures/razorwire.dm
```diff
@@ -119,8 +119,26 @@
...
+
 	if(!istype(I, /obj/item/grab))
 		return
+
 	if(isxeno(user))//I am very tempted to remove this >:)
```
TiviPlus, line 142:
```suggestion

	if(!istype(I, /obj/item/grab))
		return
	if(isxeno(user))//I am very tempted to remove this >:)
```

TiviPlus (changes requested):
The PR was fine until you made the new changes so undo the repair thing and explain the sunder change or revert it

MacBlaze1:
> The PR was fine until you made the new changes so undo the repair thing and explain the sunder change or revert it

ren told me to do this

TiviPlus (approved):

PR #5016, opened 2020-10-14
Increases Minigun Cost

Chaznoodles:
## About The Pull Request

Increases minigun req cost from 60 to 80, in line with other spec weapons.

## Why It's Good For The Game

It's getting ordered nearly every round, so I'm revisiting.

## Changelog
:cl:
tweak: We appear to be running low on miniguns, so have increased their cost accordingly.
/:cl:


Drevenshek:
xendigos just need to stop resting 5 tiles behind the frontline and they wouldent get brrrrrrd

TiviPlus (approved):

PR #5023, opened 2020-10-15
Removes Eye Protection & Tint from Sunglasses

Chaznoodles:
## About The Pull Request

Removes eye protection and the tint overlay from sunglasses and their related eyewear.

## Why It's Good For The Game

Sunglasses can once more be worn for style, and there's very little that you need eye protection from anyway.

## Changelog
:cl:
tweak: Our sunglasses lense manufacturers cheaped out and they're no longer as dark as they used to be.
/:cl:


TiviPlus (changes requested):
can you see if removing the initialize on SA glasses is possible now

PR #5027, opened 2020-10-15
Distress Nuke Win Condition

Hathkar:
## About The Pull Request

Groundmaps now always have a nuke spawn in.
Win conditions for setting the nuke off.

## Why It's Good For The Game

Marines need a new win condition besides "kill da xenos". This should also encourage more pushes towards objectives (disks and nuke), and provides a way to kill xenos without chasing them down in the caves.

## Changelog
:cl:
add: New win condition for distress: Set off the nuke!
/:cl:


Hathkar:
Definitely needs testing, since I can't do so by myself on a local server.
Some extra conditions for not evaccing the Alamo would also be good, but I'm basically just copypasting crash code.

neohenry:
WILL WE FINALLY GET RID OF CRASH?!

PR #5028, opened 2020-10-15
Add distance check for mobs crawling

### LINKED ISSUE #5006: Xenos can crawl through holes in walls from a distance
Title, Xenos can interact with walls with holes in them from distances and places they shoudn't be able to.

### Reproduction:

- Be a xenomorph and find an unblocked hole in the wall
- Ctrl click the hole
- Get pulled through it like a wormhole even if you're far away from it

Short clip from local testing:
https://i.imgur.com/o8ydDKH.mp4

Also possible to crawl through holes while in a vent and tear open walls with holes in them from a distance.

> Doesn't seem too distant from the powers of time and space folding that hivelord tunnels possess.

Tennessee116:
## About The Pull Request

This distance check _should_ close #5006, where mobs were not checked on their distance from an acid hole, allowing them to practically teleport to the hole, crawling through it. This was tested on a local copy.

## Why It's Good For The Game

Xenos cannot break the laws of physics, and it's exploitable.

## Changelog
:cl:
fix: Fixed issue where mobs can crawl through acid holes at any distance
/:cl:

### DIFF
--- a/code/game/objects/effects/acid_hole.dm
+++ b/code/game/objects/effects/acid_hole.dm
@@ -63,6 +63,10 @@
 
 	if(user.mob_size == MOB_SIZE_BIG || user.incapacitated() || user.lying_angle || user.buckled || user.anchored)
 		return
+	
+	var/dist = get_dist(user, src)
+	if(dist > 1)
+		return
 
 	var/mob_dir = get_dir(user, src)
 	var/crawl_dir = dir & mob_dir
--- a/code/game/objects/effects/acid_hole.dm
+++ b/code/game/objects/effects/acid_hole.dm
@@ -63,9 +63,8 @@
 
 	if(user.mob_size == MOB_SIZE_BIG || user.incapacitated() || user.lying_angle || user.buckled || user.anchored)
 		return
-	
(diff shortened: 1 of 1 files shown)


Hughgent:
Although this 'works' I would prefer if this check was behind `can_interact()` or `CanReach()`

Alternatively, see why these procs won't work in this case.

Tennessee116:
> Although this 'works' I would prefer if this check was behind `can_interact()` or `CanReach()`
> 
> Alternatively, see why these procs won't work in this case.

Ah. I just preferred the built-in BYOND proc. I'll see if there's any reason for it not working.

Hughgent (approved):

TiviPlus (approved):

Wayland-Smithy:
Any chance you can put the check in `specialclick()` instead because `expand_hole()` is also effected and big xenos can destroy walls with holes in them remotely.

Tennessee116:
> Any chance you can put the check in `specialclick()` instead because `expand_hole()` is also effected and big xenos can destroy walls with holes in them remotely.

Will add if the current check has no performance issues during test merge.

PR #5034, opened 2020-10-16
Adds a  distance limit to how close resin can be built near the Landing Zones

MacBlaze1:
## About The Pull Request

This PR adds a 8 (maybe 9) tile distance limit to how close xenos can spam walls near the LZ containment shutters.

## Why It's Good For The Game

This is a request made by Ren, and is designed to stop xeno wall spam or at least alleviate it to a degree.

## Changelog
:cl:
balance: added an area of 8 or 9 tiles away from the LZ containment shutters so that xenos cannot build any resin structures (WALLS) there
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -166,6 +166,11 @@
 		to_chat(X, "<span class='warning'>We can only shape on weeds. We must find some resin before we start building!</span>")
 		return fail_activate()
 
+	var/obj/machinery/door/poddoor/timed_late/containment/shutter = locate(/obj/machinery/door/poddoor/timed_late/containment) in range(8, owner.loc)
+	if(!shutter?.density)
+		to_chat(owner,"<span class='xenonotice'>You cannot build so close to the strange material.</span>")
+		return fail_activate()
+
 	if(!T.check_alien_construction(X, planned_building = X.selected_resin))
 		return fail_activate()
 
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -167,7 +167,7 @@
 		return fail_activate()
 
 	var/obj/machinery/door/poddoor/timed_late/containment/shutter = locate(/obj/machinery/door/poddoor/timed_late/containment) in range(8, owner.loc)
-	if(!shutter?.density)
+	if(shutter?.density)
 		to_chat(owner,"<span class='xenonotice'>You cannot build so close to the strange material.</span>")
 		return fail_activate()
 


zerosexualpartners:
Just place an invisible wall for Hivemind so it couldn't go thru

TacticalSpacePilot:
I think a better way to handle this issue is a hands-on approach. Instead of adding a tile distance for xenos the marines should be given a single use OB that's just a widespread napalm bomb, that can destroy resin walls, to keep it simple. 

Another alternative is to place an invisible wall or impassable fog on a timer. This way marines will be able to push out once their shutters drop but without limiting xeno weeds/building and still stopping wall spam.

I just hate the idea of restricting instead of adding new functions to combat it

SplinterGP:
> Another alternative is to place an invisible wall or impassable fog on the same timer as shutters.. This way marines will be able to push out once their shutters drop but without limiting xeno weeds/building and still stopping wall spam.
> I just hate the idea of restricting instead of adding new functions to combat it

adding a fog means it limits both survivors and xenos placing weeds at that location, aint a good idea.

adding an OB is also not doable as that wastes marine resources into a stupid thing xenos do that admins DO have permission to tell them to stop.

SplinterGP:
blase, add a pref for admins allowing it to happen incase a xeno wants to make something nearby FOB and asks admin but it aint a trump  big wall

MacBlaze1:
> blase, add a pref for admins allowing it to happen incase a xeno wants to make something nearby FOB and asks admin but it aint a trump big wall

how haha

code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -166,6 +166,11 @@
 		to_chat(X, "<span class='warning'>We can only shape on weeds. We must find some resin before we start building!</span>")
 		return fail_activate()
 
+	var/obj/machinery/door/poddoor/timed_late/containment/shutter = locate(/obj/machinery/door/poddoor/timed_late/containment) in range(8, owner.loc)
```
TiviPlus, line 169: This is disgusting
Locate and range are both awful rework this to be sane

TiviPlus (changes requested):
Additionally this does not address the cause of the issue: xeno FOB camping

Spooky was going to do rework shutters but never did I recommend you ask them for their implementation idea

code/modules/mob/living/carbon/xenomorph/abilities.dm
MacBlaze1: I was literally told by fucking splinter to do this, I told him it was bad

TacticalSpacePilot:
> adding a fog means it limits both survivors and xenos placing weeds at that location, aint a good idea.
> 
> adding an OB is also not doable as that wastes marine resources into a stupid thing xenos do that admins DO have permission to tell them to stop.

I've never seen survivors set up near the FOB, so I don't understand what you mean, and restricting xeno weeds would help with FOB camping. 

Sometimes builder castes want to build a forward operating base or strategic corridors and I think creativity is good but removing the ability to do it at all is bad when we can just include it in the rules not to build trumps wall.

Denthamos:
@TacticalSpacePilot Single use means they can just rebuild as xenos Fob camp and keep marines from going out due to the single use/expendable resources of marines versus the non expendable infinite use resource of xenos.

I get the idea of why this is needed, however this would also limit the capability of xenos to place effective walls to stop bullets from hitting them as they try to assault the FOB barricades.  Additionally one of the main issues with building resin walls up against the fob barricades and such is with the hivemind when they can just build a resin wall surrounding/in front of a turret on a flank without any alarm going off from the turret.  The initial great wall of resin can easily be removed with flamers.

PR #5038, opened 2020-10-16
Fixes the /client/keyUp: list index out of bounds runtime when using ESC key

Wayland-Smithy:
## About The Pull Request

Changes a **for** loop conditional to prevent it from throwing `list index out of bounds`

## Why It's Good For The Game

Error bad. Fix Good.

## Changelog
:cl:
fix: Fixed a list index out of bounds runtime related to keybinds.
/:cl:

### DIFF
--- a/code/modules/keybindings/bindings_client.dm
+++ b/code/modules/keybindings/bindings_client.dm
@@ -75,7 +75,7 @@
 	set instant = TRUE
 	set hidden = TRUE
 
-	for(var/i in 1 to 10)
+	for(var/i in 1 to keys_held.len)
 		if(keys_held[i] == _key)
 			keys_held[i] = null
 			break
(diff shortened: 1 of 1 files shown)


code/modules/keybindings/bindings_client.dm
```diff
@@ -75,7 +75,7 @@
 	set instant = TRUE
 	set hidden = TRUE
 
-	for(var/i in 1 to 10)
+	for(var/i in 1 to keys_held.len)
```
TiviPlus, line 78: Can you use length() here since its faster seeing as this is a pretty hot proc

Wayland-Smithy: Sounds good to me. Thanks for the suggestion I was not aware of the perf benefits, will have to remember for the future.

TiviPlus: np

TiviPlus (approved):

PR #5045, opened 2020-10-18
hivemind now has a 30 second cooldown on weeding

Pariah919:
## About The Pull Request
Hivemind makes weed attrition literally impossible because of the 12 second cooldown on weeds overtime. Other than it being annoying and nearly impossible to get rid of as marine then it shouldn't be effectively free weeds all game

## Why It's Good For The Game
Nobody really likes playing against hivemind, its just annoying to play against and is literally an invisible drone you can't do anything against

## Changelog
:cl:

balance: hivemind now has a 30 second cooldown on weeding instead of 12

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -90,7 +90,7 @@
 	return TRUE
 
 /datum/action/xeno_action/plant_weeds/slow
-	cooldown_timer = 12 SECONDS
+	cooldown_timer = 30 SECONDS
 
 // Choose Resin
 /datum/action/xeno_action/choose_resin


Chaznoodles:
This is a LONG time, and with other PR'd changes like barricades blocking weed/limits on weed node placing distance, this may be a tad excessive. It's currently 12-ish, why not make it 15 or 20?

Denthamos:
Can agree with Chaz there.  12 may be way too short considering the how fast weeds grow off weed nodes which leads to the hivemind being able to place down weeds faster than 1-3 marines can destroy them, however 30 does seem a bit long.  Starting off at 20 may be a more reasonable amount.

Hughgent (approved):
Technically correct, but probably overkill

jroinc:
I agree with you in principle.

I wouldn't be opposed to making the hivemind unable to plant weeds, but enabling off-weed roam to the edge of the screen or so. Can still build structures on pre-existing weeds, though.

TiviPlus:
Way overkill the other proposals (which actually fix the issue instead of the extremely lazy bandaiding this is) still need to be measured for impact

PR #5049, opened 2020-10-19
allows engineers to repair three separate barricades at a time, but only 1 repair at a time per barricade

MacBlaze1:
## About The Pull Request

This PR adds an actioned_list which allows for tracking of do_after targets and should lead to future usage for it, while also allowing engineers to repair 3 separate barricades, only one time each (no spam repairing). 

## Why It's Good For The Game

This will make spitter's incredibly undeescalated spit damage less effective. This does not improve repairing on anything else but metal barricades.

## Changelog
:cl:
add: Added new list in mob_defines that is used for storing action targets.
balance: welders can now repair up to 3 metal barricades, but only 1 repair action per barricade.

/:cl:

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -125,7 +125,7 @@
 	var/datum/progressbar/P = prog_bar ? new prog_bar(user, delay, progtarget, user_display, target_display) : null
 
 	user.action_busy++
-	user.actioned_list.Add(target)
+	user.do_afters[target]++
 	var/endtime = world.time + delay
 	var/starttime = world.time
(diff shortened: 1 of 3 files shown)


Hughgent (approved):

TiviPlus (changes requested):
Just turn action list into in assoc list theres no reason to waste memory like this

MacBlaze1:
> Just turn action list into in assoc list theres no reason to waste memory like this

did

TiviPlus (changes requested):
Add a comment to var/action_busy that its deprecated and illl merge

Issue #4999, opened 2020-10-12
Able to Build up Speed by Running Diagonally


Hathkar:
## Testmerges:

#4813: 'Chem rework mk5, the unga strikes back' by jroinc at commit 6fac175bf8
#4954: 'Cas Airplane' by TiviPlus at commit 54b88e84c0
#4679: 'Add buildable resin silos' by psykzz at commit 373a4d7fe0
#4747: 'Newlights' by TiviPlus at commit 17ea1f587e
#4858: 'Fixes defender crest and fortify, warrior on agility mode and warding aura pheromones not affecting xenos armor. and removes deprecated var and defines related.' by SplinterGP at commit d63c943ce5
#4965: 'Weeds speed up xenos rather than slowing down marines' by TiviPlus at commit 93e771fd0d
#4995: 'updates ship technician role to a bit more modern standards and gives some shipside roles paygrades' by MacBlaze1 at commit 909a15ce6b
#4993: 'Replaces the M41A1 with the M412' by Pariah919 at commit 1c80c16aaf

## Reproduction:

Be a hunter (Might work with other castes, didn't test). Run diagonally. Keep running diagonally, and you'll start to build up speed. Whenever you move after that, you'll keep moving super fast.

Hughgent:
old issue purge, make a new issue if this still persists.

PR #5054, opened 2020-10-20
Fixes an early return suppressing keybinding/up() - Shift examine stuck begone

Wayland-Smithy:
## About The Pull Request

Removes an early return suppressing `/datum/keybinding/up()` procs.

Possibly relates to: #4999 

Started as a fix for a runtime error:
```
examine_kb (/datum/keybinding/mob/examine): stack trace("mob_clickon overridden. Use ov...")
...
examine_kb (/datum/keybinding/mob/examine): down(ckey (/client))
```
Which happened when the user would press shift more than once because as it turns out the verb responsible for calling the `/datum/keybinding/mob/examine/up()` function where `UnregisterSignal()` lives for examine on click, or more exactly stop examining on click when shift is released was not ever being called.

Keybinding code is pretty hot and complex, so a testmerge may be in order.

## Why It's Good For The Game

Bug bad. Fix good.

## Changelog
:cl: Wayland/spookydonut
fix: Fixed a bug with keybinds registering signals becoming stuck on.
/:cl:

### DIFF
--- a/code/modules/keybindings/bindings_client.dm
+++ b/code/modules/keybindings/bindings_client.dm
@@ -80,8 +80,7 @@
 			keys_held[i] = null
 			break
 
-	if(keys_held[_key])
-		keys_held -= _key
+	keys_held -= _key
 	var/movement = movement_keys[_key]
 	if(!(next_move_dir_add & movement))
 		next_move_dir_sub |= movement
--- a/code/modules/keybindings/bindings_client.dm
+++ b/code/modules/keybindings/bindings_client.dm
@@ -80,9 +80,8 @@
 			keys_held[i] = null
 			break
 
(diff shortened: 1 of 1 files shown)


Hughgent (approved):

code/modules/keybindings/bindings_client.dm
```diff
@@ -80,9 +80,8 @@
 			keys_held[i] = null
 			break
 
-	if(!keys_held[_key])
-		return
-	keys_held -= _key
+	if(keys_held[_key])
+		keys_held -= _key
```
spookydonut, line 84:
```suggestion
	keys_held -= _key
```

spookydonut (changes requested):

PR #5056, opened 2020-10-20
Fix action yml location

Cyberboss:
Whoops


Hughgent (approved):
My trust was misplaced

PR #5068, opened 2020-10-22
Fixes barricades stopping flamethrowers and other things

MacBlaze1:
## About The Pull Request

This PR will fix an issue in which flamethrowers any anything else that uses LINKBLOCKED in determining if it can be used

## Why It's Good For The Game

Bugs bad fortnite good upvotes to the emoji react button

## Changelog
:cl:
fix: fixed barricades blocking flamers fire and other things
/:cl:


TiviPlus (changes requested):
This is not the way to do it
Properly fix flamer code instead of snowflaking core procs

PR #5076, opened 2020-10-23
Makes carrier into a t3

Pariah919:
## About The Pull Request
Makes carrier into a t3 and does the according changes, makes defiler devo into hivelord instead of carrier.
Carrier is effectively t3 strength as it is and it shouldn't be interchangeable with a t2 slot at all in the current gameplay

## Why It's Good For The Game
having a death stun machine is actually a choice now

## Changelog
:cl:
balance:carriers are t3 now

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/carrier/carrier.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/carrier/carrier.dm
@@ -11,7 +11,7 @@
 	mob_size = MOB_SIZE_BIG
 	var/list/huggers = list()
 	var/eggs_cur = 0
-	tier = XENO_TIER_TWO
+	tier = XENO_TIER_THREE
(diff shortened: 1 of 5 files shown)


davidchan13:
Ancient Carrier has the same base stats as a Young Shrike. 

It's literally not a T3 and comparing it with other T3s (Defiler, Praetorian, Shrike) it's stats lag behind considerably to be considered a T3 when it's only ability is to produce facehuggers. It doesn't even have a secondary stun.

Chaznoodles:
Every time we 'fix' carrier to stop being abused, someone finds a way around it and the cycle starts anew.

This is an alternative to just removing what is essentially an unfixable problem until capture gameplay is removed and I wholly support it.

TiviPlus (changes requested):
Buff carrier stats accordingly

Issue #5065, opened 2020-10-22
Shuffle hugging still exists


SilencedMP5A5:
what it says on the tin

## Reproduction:
1.
play
2.
get shufflehugged and slashed to death or dragged off and camped
2.5 
be forced to cadehug because the second you push, 2 happens
3. 
leave because that shit is cancer as fuck

Wayland-Smithy:
_laughs in xeno_
![4DJ3Wze8sy](https://user-images.githubusercontent.com/64715958/96917109-507fd100-145d-11eb-9a42-c0b7e3d51e37.gif)
Seriously though, I am looking making keeping direct hugger hits not count their victim as stepping on them if it was the result of a shuffle.

Wayland-Smithy:
_laughs in xeno_
![4DJ3Wze8sy](https://user-images.githubusercontent.com/64715958/96917109-507fd100-145d-11eb-9a42-c0b7e3d51e37.gif)
Seriously though, I am looking into making direct hugger hits not count their victim as stepping on them if it was the result of a shuffle.

AtlasRules:
can confirm this, i watched a carrier do it

PR #5102, opened 2020-10-26
Fixes shuffle hugging part 42: the shuffle kerfuffle

Wayland-Smithy:
## About The Pull Request

Adds a call to `go_idle()` on huggers that get successfully `Uncross()` by any xeno preventing them from passing the `Crossed()` check caused by marines that makes stepped on huggers do their thing.

Uses `Uncross()` over `Uncrossed()` because the latter happens after the `Crossed()` caused by the victim. Making sure we only `go_idle()` if conscious prevents multiple calls during the myriad of checks for `Uncross()` that happen for a single move.

Closes: #5065 

## Why It's Good For The Game

Shuffle hugging is bad (even if I love it and it is hilarious), need I say more?

## Changelog
:cl:
fix: Fixed shuffle hugging.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
@@ -226,6 +226,11 @@
 	if(stat == CONSCIOUS)
 		HasProximity(target)
 
+/obj/item/clothing/mask/facehugger/Uncross(atom/movable/AM)
+	. = ..()
+	if(. && stat == CONSCIOUS && isxeno(AM)) //shuffle hug prevention, if a xeno steps off go_idle()
+		go_idle()
+
 /obj/item/clothing/mask/facehugger/on_found(mob/finder)
 	if(stat == CONSCIOUS)
 		HasProximity(finder)


Naaanii:
So, if a marine is hit by a hugger and walks over it, it still instantly hugs but when a Xeno forces marine movement, it doesn't?

Wayland-Smithy:
Conscious "cooked" huggers will still latch on instantly, but after this change you, or any xeno, leaving the same tile as one resets them back to idle.

davidchan13:
Better solution would be to just make huggers only attach to stunned marines and their leap at marines doing stamina damage before going to sleep.

Wayland-Smithy:
I'm sticking with this because for a carrier bullying marines that would not fix the problem of them shuffling people onto as many huggers as needed to get the long stun. Plus that would nerf huggers too much overall imo.

*If* things go according to the plans of the maintainers the mechanic is on its was out anyway. That would make this change in the grand scheme just a band-aid until then to quell some unga salt. Should another PR come along that fixes just the issue in a more clever way and closes this one I welcome it.

TiviPlus (approved):

PR #5143, opened 2020-11-02
Allow xeno attack code to bypass shields

spookydonut:
## About The Pull Request
This doesn't add anything that can bypass shields but its been requested to make it possible to have abilities able to bypass shields.

I've also added the ability to bypass armor as a byproduct.

There was what looks like a bug where hunter sneak attack with disarm intent wasn't applying the pain bonus properly, I've fixed that but may need feedback on if this should just be removed.

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -58,7 +58,23 @@
 		X.visible_message("<span class='danger'>\The [X] shoves at [src], narroly missing!</span>",
 		"<span class='danger'>Our tackle against [src] narroly misses!</span>")
 		return FALSE
-	tackle_pain = check_shields(COMBAT_MELEE_ATTACK, tackle_pain, "melee")
+
+	if(protection_aura)
+		tackle_pain = tackle_pain * (1 - (0.10 + 0.05 * protection_aura))  //Stamina damage decreased by 10% + 5% per rank of protection aura
+
+	var/list/pain_mod = list()
+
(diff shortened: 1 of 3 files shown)


Surrealaser:
You are the saviour this game needs and deserves.

code/modules/mob/living/carbon/xenomorph/attack_alien.dm
```diff
@@ -163,7 +168,30 @@
...
+	var/list/armor_mod = list()
+
+	var/signal_return = SEND_SIGNAL(X, COMSIG_XENOMORPH_ATTACK_LIVING, src, damage, damage_mod, armor_mod)
+
+	// if we don't get any non-stacking bonuses dont apply dam_bonus
+	if(!(signal_return & COMSIG_XENOMORPH_BONUS_APPLIED ))
+		damage_mod += dam_bonus
+
+	if(signal_return & COMPONENT_BYPASS_ARMOR)
+		armor_block = 0
```
TiviPlus, line 184:
```suggestion
	var/armor_block = 0
	
	var/list/damage_mod = list()
	var/list/armor_mod = list()

	var/signal_return = SEND_SIGNAL(X, COMSIG_XENOMORPH_ATTACK_LIVING, src, damage, damage_mod, armor_mod)

	// if we don't get any non-stacking bonuses dont apply dam_bonus
	if(!(signal_return & COMSIG_XENOMORPH_BONUS_APPLIED ))
		damage_mod += dam_bonus

	if(!(signal_return & COMPONENT_BYPASS_ARMOR))
		armor_block = run_armor_check(affecting, "melee")
```

TiviPlus (approved):

PR #5146, opened 2020-11-03
Removes shields from Automated Armaments Vendor

Mouseofthecake:
## About The Pull Request

Removes the defensive shield from all Automated Armaments Vendors that marines have at roundstart. It's still available for purchase at requisitions, for 10 credits.

## Why It's Good For The Game

No-drop shielders have literally no counter besides a humongous swarm of xenos. And marines can get up to 6 of them for free at prep. Who.... who thought this was a good idea? At least give them the valk treatment and make them purchase-only.

## Changelog
:cl:
del: Removed defensive shield from roundstart vendors/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -137,7 +137,6 @@
 		/obj/item/explosive/grenade/cloakbomb = 25,
 		/obj/item/storage/box/m94 = 30,
 		/obj/item/storage/box/recoilless_system = 1,
-		/obj/item/weapon/shield/riot/marine = 3,
 	)
 
 	premium = list()


SplinterGP:
that beats the purpose of which we removed shields,, if it is uncounterable it woulnd't even be on req in first place. we want it be something to be available normally and MORE orderable at req like the RR. The shields are good as they let marines be more offensive, which is not bad as that reduces FOB hugging, results why shield is too strong is called Sledgehammer balancing and we buffed it to hell, now we are slowly nerfing it.

neohenry:
Too hard of a nerf.

MLGTASTICa:
Xenos should be strong in melle combat . I think the shield should just be flat out modified to block all ranged attacks , all acid sprays and any pouncing , but not be a indestructable 50% melle armour modifier.

Mouseofthecake:
What the shield should be and what it is are two very different things. A marine with two shields and B18 takes virtually no damage from melee, even from ancient ravs. Hell, a single marine with a shield and a satchel full of machine-pistols is already capable of pushing back the entire hive on their own. If y'all think this is good balance, I invite you to nerf the shield accordingly. Until then, I don't see why having to pay for it would be an issue.

TiviPlus:
The shield has already been nerfed and this is a kneejerk revert PR so I'm closing it as such

PR #5169, opened 2020-11-04
Makes Carrier stunnable

lovegreenstuff:
## About The Pull Request

Reduces carrier size to MOB_SIZE_XENO instead of MOB_SIZE_BIG. This makes them able to be stunned, because they're apparently immune to stuns, and also keeps them from doing other things. Why the fuck is a T2 unstunnable?

## Why It's Good For The Game

Carrier is for whatever reason completely impossible to stun, and also capable of tearing holes in acid hole walls. Why is this on a T2 and not its T3 evolution is beyond me. 

## Changelog
:cl:
balance: Removed two lines of code that made carrier MOB_SIZE_BIG and slower to drag. That last one is mostly just because every other MOB_SIZE_BIG xeno has that bit, and it's just for precedent's sake.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/carrier/carrier.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/carrier/carrier.dm
@@ -7,8 +7,6 @@
 	health = 200
 	maxHealth = 200
 	plasma_stored = 50
-	drag_delay = 6 //pulling a big dead xeno is hard
-	mob_size = MOB_SIZE_BIG
 	var/list/huggers = list()
 	var/eggs_cur = 0
 	tier = XENO_TIER_TWO


MLGTASTICa:
From a balance perspective , carriers are slow and not that good for frontline combat . Why should they be stunnable if they can't run as fast as other xenos? After all they're a carrier.

TiviPlus (approved):

Surrealaser:
@MLGTASTIC I agree with this; Carrier was given stun immunity for a reason.

KrabSpider:
The way I see it, carriers are immune to stuns because they don't present themselves as a large enough threat to warrant such. They don't have any abilities that would **really** give them the edge in a fight aside from the ability to throw huggers or make resin holes. 

Their tackle damage is fairly average, their slash damage is below average (3 points lower than a Drone at a measly 17), and their base speed is at the default of 0 — with Praetorians being at -0.1 and Hivelords at of 2.4. They have a health of 250 with pretty subpar armor stats across the board.
/shrug

Dolth:
At ancient, carrier has 35 HP more than a hunter but is slow and large.
Do you really want them to be stunlocked by a T39 buckshot?

Hughgent (approved):
Code is fine

Surrealaser:
Bad merge.

PR #5180, opened 2020-11-06
Resin Silo Corpse Fix

Hathkar:
## About The Pull Request

Silos spit out corpses used to create them when destroyed.

## Why It's Good For The Game

Fixes the issue with corpses being used in the silo vanishing into the void.

## Changelog
:cl:
fix: When destroyed, resin silos drop the corpses used to create them
/:cl:

### DIFF
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -69,7 +69,7 @@
 		associated_hive = null
 	for(var/i in contents)
 		var/atom/movable/AM = i
-		AM.forceMove(loc)
+		AM.forceMove(get_step(center_turf, pick(CARDINAL_ALL_DIRS)))
 	playsound(loc,'sound/effects/alien_egg_burst.ogg', 75)
 	return ..()
 
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -67,6 +67,10 @@
 		UnregisterSignal(associated_hive, list(COMSIG_HIVE_XENO_MOTHER_PRE_CHECK, COMSIG_HIVE_XENO_MOTHER_CHECK))
 		associated_hive.xeno_message("<span class='xenoannounce'>A resin silo has been destroyed at [silo_area]!</span>", 2, TRUE)
 		associated_hive = null
+	for(var/i in contents)
+		var/atom/movable/AM = i
+		AM.forceMove(loc)
+	playsound(loc,'sound/effects/alien_egg_burst.ogg', 75)
 	return ..()
 
 


Hughgent (approved):

TiviPlus (approved):

Surrealaser:
Bruh, why on earth would you get the corpses back? They're the raw biomass used to make the things.

Also, given how resin silos work now, they apparently are only valuable for getting rid of corpses, outside of creating backups for the pre-existing ones.

MLGTASTICa:
Isn't this a rebalance ? it was never a thing for resin silos to spit out their corpses.

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -67,6 +67,10 @@
 		UnregisterSignal(associated_hive, list(COMSIG_HIVE_XENO_MOTHER_PRE_CHECK, COMSIG_HIVE_XENO_MOTHER_CHECK))
 		associated_hive.xeno_message("<span class='xenoannounce'>A resin silo has been destroyed at [silo_area]!</span>", 2, TRUE)
 		associated_hive = null
+	for(var/i in contents)
+		var/atom/movable/AM = i
```
Hughgent, line 71: Thinking on this, we know that only humans can make a silo.

Thus I think we should only return bodies that have a chance of being revived. 

So I think instead of `AM` it needs to be a `human` and to `check_tod()` on each of them. `QDEL()` the ones that can't revive.

Hughgent (changes requested):

Hathkar:
The main reason I'm having to eject the corpses is so gear can be recovered.

TiviPlus:
This was always the intent

code/modules/xenomorph/xeno_structures.dm
TiviPlus: Why would you delete unrevivables theres no reason for that

TiviPlus (approved):

Hathkar:
Like Tivi said, I believe this was originally meant to be the intention, as Psy had this block of code for creating the silo. It's likely he just forgot to have them come out when the silo is destroyed.
```
	// Throw the mobs inside the silo
	for(var/iter in valid_mobs)
		var/mob/living/to_move = iter
		to_move.forceMove(hivesilo)
```

code/modules/xenomorph/xeno_structures.dm
Hughgent: The assumption being that since they are fully dead, their body has broken down into sludge and doesn't truly exist anymore.

Hathkar: It probably wouldn't be worth the effort to add a check for if they can be revived, and then having their items be stripped, which can then be expelled on silo death.

Hughgent: New suggestion. make them husked like. to show how much they've dissolved in it.

Hughgent:
```
for(var/thing in contents)
		if(!ishuman(thing))
			continue
		var/mob/living/carbon/human/dead_guy = thing
		dead_guy.icon_state = "husk_s"
		dead_guy.update_icon()
		dead_guy.forceMove(loc)
```

Hughgent (approved):

Issue #5160, opened 2020-11-04
Underslug miniflamer attachment has almost no cooldown.


GravitronBlue:
Title, underslug miniflamers on guns have effectively no cooldown.

## Reproduction:

- Attach a miniflamer to your weapon ~~of choice~~ that has burst/auto (shotgun rate of fire apperantly adds a cooldown)
- Enable miniflamer and fire as much as you'd like until you run out of fuel
- Melt xenos with direct DPS because you can hit them repeatedly with fire (use an autoclicker for best results)

Outclasses actual flamethrowers at the cost of maybe some fuel capacity and special flamer canister types while letting you punish close combat xenos heavilly with fire so you can gun them down with bullets when they flee.

Testing against several ancient castes:
https://i.imgur.com/9IZtJ5f.mp4
Note the damage on the miniflamer alone, now combine that with bullet damage after and you have incredible kill potential.

>Dolth

Update: Miniflamer cooldown is apperantly tied to the rate of fire on the gun or something.

Wayland-Smithy:
It also seems to work through walls directly adjacent too, nice.

PR #5185, opened 2020-11-07
Fixes broken miniflamer and nerfs it too

Wayland-Smithy:
## About The Pull Request

Closes: #5160 

Adds a density and space turf check during `unleash_flame()` similar to the flamer.

Utilizes the attachments existing `attachment_firing_delay` to calculate and respect a firing delay similar to the flamer.

## Why It's Good For The Game

Broken abusable attachment bad. Fix good.

## Changelog
:cl:
fix: Fixed miniflamer flames going through solid walls.
balance: Greatly increased the independent cooldown between miniflamer uses.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1312,7 +1312,7 @@ Defined in conflicts.dm of the #defines folder.
 	fire_sound = 'sound/weapons/guns/fire/flamethrower3.ogg'
 	flags_attach_features = ATTACH_REMOVABLE|ATTACH_ACTIVATION|ATTACH_RELOADABLE|ATTACH_WEAPON
 	attachment_firing_delay = 25
-	var/last_fired = 0		//When it was last fired, related to world.time.
+	COOLDOWN_DECLARE(last_fired)
 
 
 /obj/item/attachable/attached_gun/flamer/unremovable
@@ -1395,13 +1395,7 @@ Defined in conflicts.dm of the #defines folder.
 	if(get_dist(user,target) > max_range+3)
 		to_chat(user, "<span class='warning'>Too far to fire the attachment!</span>")
 		return
-	var/added_delay = attachment_firing_delay
-	if(!user.skills.getRating("firearms")) //no training in any firearms
-		added_delay += 3 //untrained humans fire more slowly.
-	if(world.time <= last_fired + added_delay)
-		return
-	last_fired = world.time
-	if(current_rounds)
+	if(current_rounds && COOLDOWN_CHECK(src, last_fired))
 		unleash_flame(target, user)
 
 
@@ -1411,6 +1405,10 @@ Defined in conflicts.dm of the #defines folder.
 	var/distance = 0
 	var/turf/prev_T
 	playsound(user, 'sound/weapons/guns/fire/flamethrower2.ogg', 50, 1)
+	var/fire_delay = attachment_firing_delay
+	if(!user.skills.getRating("firearms")) //no training in any firearms
+		fire_delay += 3 //untrained humans fire more slowly.
+	COOLDOWN_START(src, last_fired, fire_delay)
 	for(var/turf/T in turfs)
 		if(T == user.loc)
 			prev_T = T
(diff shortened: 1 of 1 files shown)


davidchan13:
I'm not sure if it's been fixed but I think UGLs need attachment_firing_delay fix as well.

jroinc:
wallhack miniflamer bad

Dolth:
UGL ROF was also bound to the rifle ROF.
Although you might wanna look into the masterkey which is still not viable because not using the middle-click functionality.

Hughgent (approved):

jroinc:
Can you drop the miniflamer ROF down to 5 (0.5 seconds)?

The reason is, unlike the normal flamer, the miniflamer doesn't do damage or ignite. It JUST drops fire tiles. It only ignites if an ayy walks onto one, or spends a full tick on it.

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1311,7 +1311,8 @@ Defined in conflicts.dm of the #defines folder.
 	slot = "under"
 	fire_sound = 'sound/weapons/guns/fire/flamethrower3.ogg'
 	flags_attach_features = ATTACH_REMOVABLE|ATTACH_ACTIVATION|ATTACH_RELOADABLE|ATTACH_WEAPON
-	attachment_firing_delay = 35
+	attachment_firing_delay = 25
+	var/last_fired = 0		//When it was last fired, related to world.time.
```
TiviPlus, line 1315: Use the COOLDOWN_DECLARE macro types for these cooldowns, see cooldown.dm

Wayland-Smithy: Sure I will look at that

Wayland-Smithy:
> Can you drop the miniflamer ROF down to 5 (0.5 seconds)?
> 
> The reason is, unlike the normal flamer, the miniflamer doesn't do damage or ignite. It JUST drops fire tiles. 

Doesn't do damage? I thought a major issue with it is: spam click a beno on 1 tile until run out of fuel and they are basically dead?

> It only ignites if an ayy walks onto one, or spends a full tick on it.

A full tick? How do you do stuff in less than a tick? Is that like a half A press?

Sorry I think I will keep the numbers as is and you can make a buff PR later.

jroinc:
> > Can you drop the miniflamer ROF down to 5 (0.5 seconds)?
> > The reason is, unlike the normal flamer, the miniflamer doesn't do damage or ignite. It JUST drops fire tiles.
> 
> Doesn't do damage? I thought a major issue with it is: spam click a beno on 1 tile until run out of fuel and they are basically dead?
> 
> > It only ignites if an ayy walks onto one, or spends a full tick on it.
> 
> A full tick? How do you do stuff in less than a tick? Is that like a half A press?
> 
> Sorry I think I will keep the numbers as is and you can make a buff PR later.

It doesn't do damage, spamclicking an ayy is no different from clicking them once (or not at all, it does no DIRECT damage).
A full LIFE tick, 2 seconds.

Fair, I'll change it later if needed. I think it will be, though.

Wayland-Smithy:
It does do damage on hit.

> Augh! You are roasted by the flames!

bnDHQF8BX4

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1404,12 +1405,18 @@ Defined in conflicts.dm of the #defines folder.
 	var/distance = 0
 	var/turf/prev_T
 	playsound(user, 'sound/weapons/guns/fire/flamethrower2.ogg', 50, 1)
+	var/fire_delay = attachment_firing_delay
+	if(!user.skills.getRating("firearms")) //no training in any firearms
+		fire_delay += 3 //untrained humans fire more slowly.
+	COOLDOWN_START(src, last_fired, fire_delay)
```
TiviPlus, line 1411:
```suggestion
		fire_delay += 0.3 SECONDS //untrained humans fire more slowly.
	COOLDOWN_START(src, last_fired, fire_delay)
```

TiviPlus (approved):

TiviPlus (approved):

Hughgent (approved):

PR #5188, opened 2020-11-07
Nerfs angled grip wield delay removal to 0.1 than 0.3 and gives a scatter mod

Pariah919:
## About The Pull Request
angled grip has 0.1 wield delay bonus than 0.3

## Why It's Good For The Game
Let's look at Pre-PR angled grip.
It gives you -0.3 wield delay, which on paper isn't much but in practice is a lot since wield delays are generally less a second long outside of LMG level weapons.
The angled grip also effectively counters its own downside in the first place because you can wield incredibly fast.
angled grip is literally bar none the best underbarreled attachment in the game because of this.
You can put ARs down to half a second. You can put carbines below that, you can get most guns to sub 0.3~ ms wield delay which makes secondaries irrelevant and just makes wield delay irrelevant, it is bar none the best underbarrel attachment if you don't want a attached gun.

This attachment by far has the most effect compared to the other stat underbarreled attachments-

Gyro gives your gun a small boost to one handed firing and makes you better at firing with slightly less scatter
Vertical grip gives you bonuses to scatters at the cost of wield delay (and is effectively the opposite of this gun.)
Bipod is bruh on 99.9% of guns other than the GPMG.
BFA kinda just exists, as most people don't use burst fire anymore outside of 6 shot carbine gamers.

## Changelog
:cl:
balance: angled grips have been nerfed to -0.1 wield delay, they increase your scatter by 10
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1520,16 +1520,17 @@ Defined in conflicts.dm of the #defines folder.
 
 /obj/item/attachable/angledgrip
 	name = "angled grip"
-	desc = "A custom-built improved foregrip for less recoil, and faster wielding time. \nHowever, it also increases weapon size, and slightly hinders unwielded firing."
+	desc = "A custom-built improved foregrip for less recoil, and faster wielding time. \nHowever, it also increases weapon size, scatter and slightly hinders unwielded firing."
 	icon_state = "angledgrip"
 	attach_icon = "angledgrip_a"
-	wield_delay_mod = -0.3 SECONDS
+	wield_delay_mod = -0.1 SECONDS
 	size_mod = 1
 	slot = "under"
 	pixel_shift_x = 20
 	recoil_mod = -1
 	accuracy_unwielded_mod = -0.1
 	scatter_unwielded_mod = 5
+	scatter_mod = 10
 
 
 


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -61. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

Jeff-jefferson-jeffery:
have you considered the underbarrel weapons and the fact that all the other ones are useless, and not the angled grip being op?

MLGTASTICa:
Maybe nerf it to 0.15 instead of 0.3 , 0.2 change seems too much.

Dolth:
The angled grip downside should be not being able to use an underbarrel attachment, a gyro, or an vertical grip.
You should really start thinking about how to balance the weapons as a whole instead of trying to make each attachments different every 2 months.
As a suggestion, make every guns barely viable without attachments so you would actually customize the way you want to play your rifle.

Pariah919:
>The angled grip downside should be not being able to use an underbarrel attachment, a gyro, or an vertical grip.
You should really start thinking about how to balance the weapons as a whole instead of trying to make each attachments different every 2 months.
As a suggestion, make every gun barely viable without attachments so you would actually customize the way you want to play your rifle.
>
this is the type of balance guns shouldn't be balanced around, attachments shouldn't make your rifle useable. They should make them play slightly differently.
The angled grip is just better than every other underbarrel attachment unless you want to pick an attached gun. There's a reason attachment slots exist. If we don't balance based off comparing them to the other same slot items why even have attachment slots? At that point just remove the slot system entirely.

RenGusta:
He's not wrong when it comes to the point of tweaking a weapon rather than defining how it plays. Attachments should be small but somewhat notable changes to the way your weapon performs, not cutting off almost half or more of the wield delay on the weapon it's applied to. 

I'll be working on a design document tomorrow which goes over weapon stats as well as attachments in relation to that, so you can store up your salt for that.

Dolth:
I think you misunderstood me, what I meant is, if every gun is slightly below average but attachments aren't totally useless then you're totally able to make your gun viable THE WAY YOU WANT.
You can choose to get better accuracy at the expense of not raising damage or lowering scatter or instead you wanna scatter but hit hard and moderately accurately, etc, etc.

I think every attachments should have a purpose at such a point it defines the gun you use.
Because right now you're removing the use of attachments by making them unusable, BC before, now AG?
I might have been a bit salty on the neuro nerf you're working on but on a serious note, this needs more serious insights.

Hughgent (approved):
code is fine

Isy232:
I think this is just poor design, xenos have MASSIVE crowd control and this is buffing that.

A good warrior can CC someone so hard they cant get their gun up for a solid 3 seconds.

I don't think this is good.

Hughgent:
This has been TM'd a few times. Merge or no?

Pariah919:
image
nice

PR #5223, opened 2020-11-10
Ports advanced shuttle landing from tg and adds a dropship using this feature to the Sulaco and Pillar

Eskjjlj:
## About The Pull Request
A small dropship that can land anywhere on the map  has been added to the Sulaco hangar.

Characteristics : 

- It can land on any open turf and on resin walls to prevent xenos cheesing it by putting walls everywhere.
- It has a crew mount and two weapon mounts on the sides.
- It flies slightly faster than the Alamo but has the same cooldown.
- It can't be hijacked but there is no remote control for it on the ship.

There might be issues with icon_states after the shuttle rotates but that's outside the scope of this PR...

## Why It's Good For The Game
Why this new tool marines will be able to hit behind the xenos! No more fobcamping! (Hopefully)

## Changelog
:cl:
add: High Command has outfitted the Sulaco with a new drop shuttle : the Tadpole.
It can land anywhere in the area of operations!
/:cl:


Tennessee116:
Interesting. Will definitely switch shit up a lot.

spookydonut:
rotation already works but we dont have the sprites to support it 100%

allowing marines to land anywhere is a really bad idea.  at the very least it should require someone on the ground to mark or designate an area to land in.

Eskjjlj:
How about not allowing it to land if there are resin walls blocking the way?

Xenos only need to build one wall to block the whole ship and marines will have to clear a landing zone for the ship. This way it wont be able to land unless marines are supporting it on the ground.

jroinc:
> How about not allowing it to land if there are resin walls blocking the way?
> 
> Xenos only need to build one wall to block the whole ship and marines will have to clear a landing zone for the ship. This way it wont be able to land unless marines are supporting it on the ground.

Won't *reliably* be able to land, which seems fair. That lets ayys block off areas, but it's doubtful they'll try (or be able)to block off the whole map. Ungas can keep trying different spots, but probably won't be able to hot-drop.

Pariah919:
i mean hotdropping is just going to result in you dying 99% of the time without support

spookydonut:
Not a fan of hotdropping, having a minishuttle to establish a forward base and shorten supply lines is good though.

Again, I think the marines on the ground should mark the landing zone, especially given the risk of players being gibbed when the shuttle lands.

jroinc:
There's 0 risk players will get gibbed. A sound and the 10-15 second warning of the shuttle outline will prevent that.

spookydonut:
> 
> 
> There's 0 risk players will get gibbed. A sound and the 10-15 second warning of the shuttle outline will prevent that.

In a cramped area its plently possible that players would be unable to leave in time.

jroinc:
> > There's 0 risk players will get gibbed. A sound and the 10-15 second warning of the shuttle outline will prevent that.
> 
> In a cramped area its plently possible that players would be unable to leave in time.

Assuming that it can't land in dense tiles, and it's ~5x9 tiles wide (Size is irrelevant, I just picked a decent smol shuttle size), it would have to land in a perfectly 5x9-shaped hole to stand a chance of trapping people. There will be at least 1 exit from the hypothetical "shuttle-shaped LZ" (otherwise no one could get in to have a chance of being gibbed). With 10-15 seconds with the OBVIOUS hyperspace ripples and sound, any living player will, you know, leave the way they came in.

This leaves two scenarios where someone "could" be gibbed, and I'll treat them the same- Living but tackled down, and dead.

Headbite takes 10 seconds to permakill a marine. The shuttle takes 10-15 seconds to land. Dragging an unga to a silo also deletes his gear.

It's not actually an issue.

Pariah919:
couldn't you just make it not gib you at all

Eskjjlj:
I'm not interested in making it able to land only in areas designated by ground units. 

This is a severe limitation that restricts the gameplay potential of such a feature.

code/_globalvars/lists/objects.dm
```diff
@@ -42,6 +42,7 @@ GLOBAL_LIST_EMPTY(xeno_resin_silo_turfs)
 GLOBAL_LIST_EMPTY(xeno_resin_silos)
 
 GLOBAL_LIST_EMPTY(shuttle_controls_list)
+GLOBAL_LIST_EMPTY(navigation_computers)				//list of all /obj/machinery/computer/camera_advanced/shuttle_docker
```
TiviPlus, line 46: We dont need this list do we

code/game/objects/structures/dropship_equipment.dm
```diff
@@ -362,6 +368,8 @@
 /obj/structure/dropship_equipment/sentry_holder/proc/deploy_sentry()
 	if(!deployed_turret)
 		return
+	setDir(ship_base.dir)
+	deployed_turret.setDir(dir)
```
TiviPlus, line 372: Won this also rotate turrets facing to the side(like on the alamo)?

code/modules/shuttle/navigation_computer.dm
```diff
@@ -0,0 +1,378 @@
...
+	var/shuttlePortName = "Custom Location"
+	var/list/jumpto_ports = list() //hashset of ports to jump to and ignore for collision purposes
+	var/obj/docking_port/stationary/my_port //the custom docking port placed by this console
+	var/obj/docking_port/mobile/shuttle_port //the mobile docking port of the connected shuttle
+	var/list/locked_traits = list(ZTRAIT_RESERVED, ZTRAIT_CENTCOM) //traits forbided for custom docking
+	var/view_range = 0
+	var/x_offset = 0
+	var/y_offset = 0
+	var/list/whitelist_turfs = list(/turf/open/ground, /turf/open/floor, /turf/closed/wall/resin)
+	var/see_hidden = FALSE
```
TiviPlus, line 19:
```suggestion
	var/list/whitelist_turfs = list(/turf/open/ground, /turf/open/floor)
	var/see_hidden = FALSE
```

code/modules/shuttle/shuttle.dm
```diff
@@ -11,19 +11,48 @@
 
 	resistance_flags = RESIST_ALL
 	anchored = TRUE
-//
+
+	/// The identifier of the port or ship.
+	/// This will be used in numerous other places like the console,
+	/// stationary ports and whatnot to tell them your ship's mobile
+	/// port can be used in these places, or the docking port is compatible, etc.
```
TiviPlus, line 18:
```suggestion
	/** 
	  * The identifier of the port or ship.
	  * This will be used in numerous other places like the console,
	  * stationary ports and whatnot to tell them your ship's mobile
	  */ port can be used in these places, or the docking port is compatible, etc.
```

code/_globalvars/lists/objects.dm
Eskjjlj: Nope.

code/game/objects/structures/dropship_equipment.dm
Eskjjlj: ship_base is actually the weapon attach mount which has its own dir thats different from the direction of the shuttle. 

I had to add this because the dir of turrets didn't get properly updated when the ship rotated.
The direction was only calculated when the powerloader attached it on the mount.

So the Alamo turrets will be fine.

TiviPlus (approved):

MemedHams:
Okay, so, since this has been testmerged, multiple marine squads have been rest-piling into it and using it to teleport into the young xenos prep region before the shutters drop.

Walarks:
I think it shouldn't be able to land inside caves. This is due to caves acting like a safe spot for xenos to organise since they are unable to be bombarded and being able to land inside can give too many opportunities to surround xenos without indication for this.

code/modules/shuttle/navigation_computer.dm
```diff
@@ -0,0 +1,379 @@
...
+					. = SHUTTLE_DOCKER_BLOCKED_BY_HIDDEN_PORT
+			else
+				I.icon_state = "red"
+				. = SHUTTLE_DOCKER_BLOCKED
+
+/obj/machinery/computer/camera_advanced/shuttle_docker/proc/checkLandingTurf(turf/T, list/overlappers)
+	// Too close to the map edge is never allowed
+	if(!T || T.x <= 10 || T.y <= 10 || T.x >= world.maxx - 10 || T.y >= world.maxy - 10)
+		return SHUTTLE_DOCKER_BLOCKED
+	// If it's one of our shuttle areas assume it's ok to be there
```
Hughgent, line 250: Around here, check the turfs area, and include a `SHUTTLE_DOCKER_BLOCKED` for when the `ceiling` var of the area is `>= CEILING_DEEP_UNDERGROUND`

Eskjjlj: It has been done... and it works.

Eskjjlj:
Cave landings are now D.

Eskjjlj:
Longer cooldown to make hotdrops more risky.

spookydonut:
I'm still a hard no on freeform shuttle landing.

Eskjjlj:
I understand why someone would dislike this feature if their favourite thing about TGMC was cadehugging at fob for two hours.
This shuttle gives the marines a way to operate outside the fob without being instantly killed.

Hughgent:
```
Commander TGMC BOT
11/24/2020
REPO: Updating origin branch...
REPO: Merging PRs #5254, #5257, #5263, #5267, #5223, #5235...
```
```
DuccBot BOT
Today at 7:29 PM
TGMC Statbus
Winrate (Xenomorph to Marine)
64.52%
View Raw
http://statbus.psykzz.com:8080/api/winrate?delta=3
```
Reasonable balance

Hughgent (approved):

PR #4571, opened 2020-07-23
respawning as marine is now 1 hour instead of 30 minutes

Pariah919:
## About The Pull Request
Title basically.

## Why It's Good For The Game
Before we had the 30 minute respawn as a concession because of our low population. Now that we get average 50~ players we don't really need this to be so short anymore.

## Changelog
:cl:
balance: Marine respawn time doubled.
/:cl:

PR #5232, opened 2020-11-12
Revert "respawning as marine is now 1 hour instead of 30 minutes"

Hathkar:
Reverts tgstation/TerraGov-Marine-Corps#4571

Reverting this because marines are more likely to perma-die now from headbites, and don't want to wait for an hour.


Hughgent:
Yeah please don't do this revert 'in this manner'. Do go ahead and open a new pr with just the variable change.

Seems to contain a bunch of garbage file edits.

PR #5238, opened 2020-11-13
Headbite Changes

Hathkar:
## About The Pull Request

- [x] Can't be used while on fire
- [x] Channel will cancel if damage is taken
- [x] Only usable on the dead

## Why It's Good For The Game

Lots of people want changes, so here they are.

## Changelog
:cl:
balance: Headbite is only usable on the dead
balance: Headbite can't be used while on fire
balance: Headbite will cancel if damage is taken
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -63,7 +63,8 @@
 	if(!X.Adjacent(victim)) //checks if owner next to target
 		return FALSE
 	if(X.on_fire)
-		to_chat(X, "<span class='warning'>We're too busy being on fire to do this!</span>")
+		if(!silent)
+			to_chat(X, "<span class='warning'>We're too busy being on fire to do this!</span>")
 		return FALSE
 	if(victim.stat != DEAD)
 		if(!silent)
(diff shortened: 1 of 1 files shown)


SplinterGP:
did ya find a way to cancel it if takes damage ?

code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -62,6 +62,13 @@
 		return FALSE
 	if(!X.Adjacent(victim)) //checks if owner next to target
 		return FALSE
+	if(X.on_fire)
+		to_chat(X, "<span class='warning'>We're too busy being on fire to do this!</span>")
+		return FALSE
```
TiviPlus, line 67: silent check

Wayland-Smithy:
I suppose adding the requirement they be dead is fine, but headbiting for the kill is great fun, but...
> Channel will cancel if damage is taken

Is there any precedent for this? we already have stagger, do marines who are building/busy doing other stuff get their action canceled on damage cause of any "ouchies the pain causes you to mess up what you are doing" type reasoning?

For non high-speed castes the risk/reward of going for a headbite with marines close by is already good gameplay IMO. Maybe headbiting can "exaust" the xeno for a couple seconds after channel end by capping move speed to that of like queen or crusher? It could have a ui element like purp gas or broken bones.

code/modules/mob/living/carbon/xenomorph/abilities.dm
Hughgent: i.e. put this behind `if(!silent)`

Hughgent (approved):

TiviPlus (approved):

PR #5247, opened 2020-11-14
defender has 90 armor at ancient fortify again

Pariah919:
## About The Pull Request
hugh changed ancient to have 95 armor instead of 90 as was intended because nobody commented to point this out, because it totals up to more armor (As its 40+50 which means 90)
it ends up giving defender way more armor than it should probably have and was intended.

## Why It's Good For The Game
unintended balance change bad comment good

## Changelog
:cl:
fix: defender armor at ancient is once again 90% rather than 95%
/:cl:


Hughgent:
Sunder is a mechanic that exists.

PR #5322, opened 2020-11-24
Tweak new Hive Status health color to be less red all the time

Wayland-Smithy:
## About The Pull Request

Makes the health numbers color change from green, to orange, to red depending on the value.

## Why It's Good For The Game

Less unimportant red text demanding attention.

## Changelog
:cl:
tweak: Hive Status health text is now only red when below a third of max health.
/:cl:


TiviPlus (approved):

Surrealaser:
TBH I just had it as red so it would stand out and be distinct, but I like this

PR #5344, opened 2020-11-26
Fixes xeno_action's add_cooldown() sanity check to prevent 0 length timers

Wayland-Smithy:
## About The Pull Request

Adds a check for 0 length cooldowns to `add_cooldown()`

## Why It's Good For The Game

Fix good.

## Changelog
:cl:
fix: Fixed a timer error with 0 length cooldown xeno actions.
/:cl:

### DIFF
--- a/code/datums/actions/xeno_action.dm
+++ b/code/datums/actions/xeno_action.dm
@@ -138,10 +138,11 @@
 
 /datum/action/xeno_action/proc/add_cooldown()
 	SIGNAL_HANDLER
-	if(cooldown_id) // stop doubling up
+	var/cooldown_length = get_cooldown()
+	if(cooldown_id || !cooldown_length) // stop doubling up or waiting on zero
 		return
 	last_use = world.time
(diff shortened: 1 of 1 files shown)


Hughgent:
why there and not in `xeno_action` ? is it not runtiming for other things like it?

Wayland-Smithy:
I'll check @Hughgent, but I am thinking they are probably not calling `add_cooldown()` if they have not set the `cooldown_timer`

Hughgent:
Well, this would be a kind of 'future proofing' such that if anyone else adds an ability.

Wayland-Smithy:
Can't argue with that, but again it'd only matter if they use `add_cooldown()` by mistake, like from doing some copy pasta which is probably more than likely actually 🤔

Yeah I'll just move it to the base datum.

Hughgent:
Stack trace is still calling 0 length timer with your PR merged locally. please solve

Wayland-Smithy:
With the PR branch checked out or merged into master at the time of writing I am unable to cause a 0 timer stack trace with Plant Weeds. 

Can you please share the log and repro steps?

TiviPlus (approved):

PR #5366, opened 2020-11-27
Making Lasrifle Overcharge Great Again

Surrealaser:
## About The Pull Request

As a xeno player I'm probably going to regret this; don't say I didn't give the marines anything (y'know besides the litany of stuff about a year and a half ago).

Improves the Overcharge mode so it's actually worth using vs heavily armored targets (Ancient Rav or better); Lasrifle Overcharge damage increased from 42 to 60, Penetration increased from 20 to 50, and Sunder increased from 10 to 15.

**Math vs Ancient Rav (50% laser resistance) at range 5 over 1 second of firing; standard bolts have 5x the fire rate of Overcharge:**

**Standard Bolt:**  Base Damage With Falloff (20-0.25 * 5) * Fire Rate Mod (10/2) * Armor Mod (1-1 * (0.5-0.1)) = 56.25
**Overcharge Bolt:** (55-0.25 * 5) * (10/10) * (1-1 * (0.5-0.5)) = 53.75
Overcharge deals +2.5 more Sunder

**vs Ancient Defiler:**
**Standard Bolt:**  (20-0.25 * 5) * (10/2) * (1-1 * (0.4-0.1)) = 65.625
**Overcharge Bolt:** (55-0.25 * 5) * (10/10) * 1 (can't be less than 1) = 53.75
Overcharge deals +2.5 more Sunder

**vs Ancient Hunter:**
**Standard Bolt:**  (20-0.25 * 5) * (10/2) * (1-1 * (0.3-0.1)) = 75
**Overcharge Bolt:** (55-0.25 * 5) * (10/10) * 1 (can't be less than 1) = 53.75
Overcharge deals +2.5 more Sunder

## Why It's Good For The Game

Makes the lasrifle overcharge worth a shit, and situationally good for Sunder, ghetto sniping and vs heavily armored Xenos.

## Changelog
:cl:
balance: Lasrifle Overcharge damage increased from 42 to 55, Penetration increased from 20 to 50, and Sunder increased from 10 to 15.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1228,10 +1228,10 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	name = "overcharged laser bolt"
 	icon_state = "heavylaser"
 	hud_state = "laser_sniper"
-	damage = 42
+	damage = 65
 	max_range = 40
-	penetration = 20
-	sundering = 10
+	penetration = 50
+	sundering = 15
 
 /datum/ammo/energy/lasgun/M43/heat
 	name = "microwave heat bolt"
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1228,7 +1228,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	name = "overcharged laser bolt"
 	icon_state = "heavylaser"
 	hud_state = "laser_sniper"
-	damage = 65
+	damage = 60
 	max_range = 40
 	penetration = 50
 	sundering = 15
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1228,7 +1228,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	name = "overcharged laser bolt"
 	icon_state = "heavylaser"
 	hud_state = "laser_sniper"
-	damage = 60
+	damage = 55
 	max_range = 40
 	penetration = 50
 	sundering = 15


neohenry:
lasgun pb in overcharge time? 👀

Surrealaser:
> lasgun pb in overcharge time? 👀

Spread is better vs most targets; I believe the PB is only better than Spread vs like Crushers, Defenders in defense/fortify mode and Queens

neohenry:
> 
> 
> > lasgun pb in overcharge time? 👀
> 
> Spread is better vs most targets; I believe the PB is only better than Spread vs like Crushers, Defenders in defense/fortify mode and Queens

Yeah, usually I go for lasgun PB in spread assuming that spread PB > overcharge PB.

Chaznoodles:
For reference - this gives it _double_ the sunder of the PFC sniper and 20 more penetration, as well as previously having the same max_range and getting 12 shots instead of 7, while only losing 20 damage (probably less because of the sunder/penetration). There's a reason overcharge is called heavy_sniper in the HUD - because it was meant to function as the worse-than-the-actual-sniper-weapons mode in the jack of all trades weapon. This just seems to make it the preferable option over the actual PFC sniper.

RenGusta:
Pretty fucking meme. Not sure it's the solution, though lasgun modes overall need some love and balance.

Surrealaser:
> For reference - this gives it _double_ the sunder of the PFC sniper and 20 more penetration, as well as previously having the same max_range and getting 12 shots instead of 7, while only losing 20 damage (probably less because of the sunder/penetration). There's a reason overcharge is called heavy_sniper in the HUD - because it was meant to function as the worse-than-the-actual-sniper-weapons mode in the jack of all trades weapon. This just seems to make it the preferable option over the actual PFC sniper.

Sniper has NV and IFF which are absolutely huge differences. Also, 30 pen is basically enough to totally ignore or close to ignore armor vs most castes and upgrade levels, so the damage output of a sniper vs everything but say a Crusher is going to be higher.

Let's consider an Ancient Crusher at say range 10:

Sniper: (80-0.25 * 10) * (1-1 * (0.75-0.3)) = 42.625
Overcharge: (60-0.25 * 10) * (1-1 * (0.75-0.5)) = 43.125

They're almost even with each other and the lasrifle has no NV and no IFF, and this is the absolute best case scenario; every other match up in the game will be worse.

Chaznoodles:
> Sniper has NV and IFF

The fact that you didn't read what was written says it all

Hughgent (approved):
variable changes. 

However I must echo the concern that this becomes equivalent to the more dedicated platform. The lasrifle should not ever be competitive stats wise to a more focused option. Its strength lies in its flexibility.

Please consider a 10 to 20 percent worsening vs the Sniper rifle.

Surrealaser:
> variable changes.
> 
> However I must echo the concern that this becomes equivalent to the more dedicated platform. The lasrifle should not ever be competitive stats wise to a more focused option. Its strength lies in its flexibility.
> 
> Please consider a 10 to 20 percent worsening vs the Sniper rifle.

See above. I'll drop the base damage by 5 but I don't think any further changes are necessary until TM.

Hughgent (approved):

Hughgent:
balance has been 'fine' over the past 3 days.

PR #5372, opened 2020-11-28
Changes Hivemind Collapse timer to be longer in cases where more time is needed, not less

Wayland-Smithy:
## About The Pull Request

Inverses the logic for setting the orphan hivemind collapse timer to either 2.5 or 5 minutes.

## Why It's Good For The Game

I believe it makes more sense to give the hive more time when there is no easy path to a new ruler than to give less.

## Changelog
:cl:
tweak: Hivemind Collapse will now only occur faster if there is a living Drone or Larva when initiated.
/:cl:


Hughgent (approved):
variable changes

PR #5404, opened 2020-12-01
Slug Penetration Reduced, Sunder Increased

Surrealaser:
## About The Pull Request

Slug Penetration reduced from 40 (wtf) to 10.

Slug Sunder increased from 7 to 15.

Slug Damage increased from 80 to 90.

## Why It's Good For The Game

Slug stats now make sense, gain a strong niche as an armor breaker (and reinforces their niche as a support round), and can be used more safely with less FF risk.

I value Damage and Sunder at roughly 1.5 : 1 vs Pen.

## Changelog
:cl:
balance: Slug Damage increased from 80 to 90,
balance: Slug Penetration reduced from 40 to 10.
balance: Slug Sunder increased from 7 to 15.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -559,7 +559,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
 	shell_speed = 3
 	max_range = 15
-	damage = 80
+	damage = 90
 	penetration = 10
 	sundering = 15
 
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -560,8 +560,8 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	shell_speed = 3
 	max_range = 15
 	damage = 80
-	penetration = 40
-	sundering = 7
+	penetration = 10
+	sundering = 15
 
 /datum/ammo/bullet/shotgun/slug/on_hit_mob(mob/M,obj/projectile/P)
 	staggerstun(M, P, weaken = 1, stagger = 2, knockback = 1)


Chaznoodles:
Flechette is already the armor breaker, slug isn't and is intended for damage. Unneeded change.

Surrealaser:
> Flechette is already the armor breaker, slug isn't and is intended for damage, which the penetration on slugs instead of sunder supports. Unneeded change, in combination with your second PR #5405 which would give this the same pen as before to non-xenos target.
> 
> > can be used more safely with less FF risk.
> 
> While at the same time making a PR which would give them exactly the same pen vs your buddies but doing more damage. This isn't a HvH server, we don't want to be taking our buddies out of the round permanently by accident either.

Flechette makes 0 sense as an armor breaker; it fires armor penetrating needles, not armor impacting and destroying impact rounds.

Second, as mentioned on discord, PRs necessarily have to be atomized and thus this PR is subject to change depending on the merge status of others.

Chaznoodles:
> Flechette makes 0 sense as an armor breaker; it fires armor penetrating needles, not armor impacting and destroying impact rounds.

Flechette makes more sense as an armour breaker in terms of the game because buckshot already fills the high damage scatter niche. Flechette is the alternative for breaking xenos armour and forcing them to retreat.
>

> Second, as mentioned on discord, PRs necessarily have to be atomized and thus this PR is subject to change depending on the merge status of others.

This does not change the content of it in relation to others.

Pariah919:
This just makes flechette irrelevant, just delete flechette if you're doing this.

Surrealaser:
@Pariah919 Flechette should be penetrative, not sundering

Pariah919:
> @Pariah919 Flechette should be penetrative, not sundering
>
then delete slugs
ap in general should be lowered, and flechettes having armor penetration would require you to make them hilariously high in general to make them different from buckshot in general.
Flechettes are tiny metal darts. They simply were experimented with then ditched irl because they sucked ass poop garbage at doing the one thing they were supposed to- Which extends the range of buckshot ammo. They should shred through armor, not penetrate it.

You could probably remove flechette and just rework buckshot at this point, shotguns having 3 main ammo types is probably one too many.

Hughgent (approved):
variable changes

PR #5405, opened 2020-12-01
Sunder Now Affects Non-Xeno Targets as Extra Penetration

Surrealaser:
## About The Pull Request

Projectiles with Sunder and the Sunder flag now deal extra pen to non-Xeno targets equal to twice their Sunder value.

This is explained as Sunder weapons imparting armor-penetrating kinetic shock/transfer upon their target.

## Why It's Good For The Game

Eliminates Sunder as being a wholly one-sided specifically anti-xeno metric in violation of the double-edge rule followed by virtually every other primary projectile trait that acts on a target. Sunder is currently a pure profit metric unlike damage, penetration, stagger, stun and incendiary stacks with no FF risk to balance out the reward.

## Changelog
:cl:
rebalance: Sundering projectiles now apply twice their Sunder value as extra penetration to non-Xeno targets.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -779,6 +779,10 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 	var/living_soft_armor = (proj.ammo.flags_ammo_behavior & AMMO_IGNORE_ARMOR) ? 0 : get_soft_armor(proj.armor_type, proj.def_zone, proj.dir)
 	if(living_hard_armor || living_soft_armor)
 		var/penetration = proj.ammo.penetration
+
+		if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING && !isxeno(src)) //Versus non-xenos, Sunder counts as extra penetration; concussive force that transmits through armor
+			penetration += proj.ammo.sundering
+
 		if(penetration > 0)
 			if(proj.shot_from && src == proj.shot_from.sniper_target(src))
 				damage *= SNIPER_LASER_DAMAGE_MULTIPLIER
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -781,7 +781,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		var/penetration = proj.ammo.penetration
 
 		if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING && !isxeno(src)) //Versus non-xenos, Sunder counts as extra penetration; concussive force that transmits through armor
-			penetration += proj.ammo.sundering
+			penetration += (proj.ammo.sundering * 2) //Multiply by 2 or it is pretty much irrelevant for the vast, vast majority of projectiles
 
 		if(penetration > 0)
 			if(proj.shot_from && src == proj.shot_from.sniper_target(src))


neohenry:
All the more the reason to use aim mode.

. . . Shoot, need to talk about that in the wiki.

Surrealaser:
@neohenry The effect of this will be totally negligible for most weapons; very few have any significant Sunder to speak of.

Chaznoodles:
Sunder is supposed to be a xenos-specific thing, all this does is add extra pen to weapons when you FF marines - exactly what we want to avoid. 

I remember discussing friendly fire a while back and numbers were changed to lessen the impact of non-sustained friendly fire, so players can continue to enjoy the game rather than have to sit idle while they get serious wounds caused on accident by other marines fixed by doctors. This change, looks to go back that way, and will not be a move in a positive direction.

Surrealaser:
> Sunder is supposed to be a xenos-specific thing and is balanced around that, all this does is add complications if we want to change sunder regarding xenos down the line and add extra pen to weapons when you FF marines - exactly what we want to avoid.
> 
> I remember discussing friendly fire a while back and numbers were changed to lessen the impact of non-sustained friendly fire, so players can continue to enjoy the game rather than have to sit idle while they get serious wounds caused on accident by other marines fixed by doctors. This change, looks to go back that way, and will not be a move in a positive direction.
> 
> Some numbers:
> 50pen mosins
> 50pen PFC snipers
> 100pen light RR rounds
> O N E F I F T Y P E N heavy RR rounds
> 80 pen lasgun overcharge (with proposed changes, linked below)
> Instacrit shotguns
> 
> Also to note #5404 and #5366, where you increase the sunder of some weapons by a huge amount.

1: Half of those examples you listed are Pen overkill to the point the extra pen either doesn't matter or will barely matter. Anything that was 50 pen or greater will essentially not have a meaningful increment in penetration.
2: Aim mode is workable with all weapons you mentioned except maybe the lasrifle (and the lasrifle doesn't do lasting damage such as fractures). Aim mode is also obviously preferable for the Mosin and Sniper anyways; not weapons you want to snapshot with marines out in front such that FF actually matters.
3: Beyond this, the vast majority of weapons in the game won't be affected by this materially.

Chaznoodles:
> 1: Half of those examples you listed are Pen overkill to the point the extra pen either doesn't matter or will barely matter. Anything that was 50 pen or greater will essentially not have a meaningful increment in penetration.

Then why make this change if it's not going to be meaningful - despite this being false and making heavy hitting weapons do even MORE FF damage?

> 2: Aim mode is workable with all weapons you mentioned except maybe the lasrifle (and the lasrifle doesn't do lasting damage such as fractures). Aim mode is also obviously preferable for the Mosin and Sniper anyways; not weapons you want to snapshot with marines out in front such that FF actually matters.

As stated by the majority of people involved in the discussion on Discord including Tivi, aim should not be taken into account when balancing and things should be balanced by their own merit. Aim is not a reason for this.

> 3: Beyond this, the vast majority of weapons in the game won't be affected by this materially.

Then why make this change if it's not going to affect most weapons, unless that's also false as per the first point?

Surrealaser:
> > 1: Half of those examples you listed are Pen overkill to the point the extra pen either doesn't matter or will barely matter. Anything that was 50 pen or greater will essentially not have a meaningful increment in penetration.
> 
> Then why make this change if it's not going to be meaningful - despite this being false and making heavy hitting weapons do even MORE FF damage?
> 
> > 2: Aim mode is workable with all weapons you mentioned except maybe the lasrifle (and the lasrifle doesn't do lasting damage such as fractures). Aim mode is also obviously preferable for the Mosin and Sniper anyways; not weapons you want to snapshot with marines out in front such that FF actually matters.
> 
> As stated by the majority of people involved in the discussion on Discord including Tivi, aim should not be taken into account when balancing and things should be balanced by their own merit. Aim is not a reason for this.
> 
> > 3: Beyond this, the vast majority of weapons in the game won't be affected by this materially.
> 
> Then why make this change if it's not going to affect most weapons, unless that's also false as per the first point? By your own words you've stated "this is going to affect neither heavy damage weapons or light damage weapons", which is blatantly false.

1. That it doesn't significantly increase the con of some corner case weapons doesn't obviate the reason for making the PR.

2. Aim mode or not, you don't want to fire the Mosin and a non-IFF Sniper with people in front of you. With Aim Mode however, this is a non-issue for these weapons.

3. Because there are enough weapons for which this is relevant, and while individually this may not matter for most weapons (i.e. the impact is not notable for most weapons with one dude FFing at a time), collectively it will matter. Generally it's a good thing to give something as powerful as Sunder, a property more powerful than Penetration at that, a downside, even if that downside is modest overall in the majority of cases.

Hughgent (approved):
simple change

TiviPlus:
Sunder is a xeno only mechanic because it is hellish to balance for both teams (see case in point penetration). Additionally FF is already bad as is so I see no reason for this change

PR #5410, opened 2020-12-02
Removes flechette, changes shotgun ammo in general.

Pariah919:
## About The Pull Request
Removes flechette.
Makes Slugs do 80/30/10 instead of 80/40/7.5, and makes buckshot do 200 from 240 damage but has less spread and projectiles, and so it has better range overall rather than being literally a melee weapon.

## Why It's Good For The Game
3 ammo types is probably too much for the shotgun, as flechette is clear to never be balanced idea wise, as the idea behind it is being either a penetrating ammo weapon (which it would need absurd amounts to actually score higher damage than buckshot in a lot of cases.) and its medium range niche is outdone by slugs, and slugs also deal better damage cqc in a lot of cases anyway.
Also the autoshotgun is just better so uhh
Let it go.

## Changelog
:cl:
balance:   Pump shotgun Flechette is gone. Buckshot deals 200 damage from 240, but has reduced falloff and less projectile for better medium range performance. Slugs changed from 80/40/7.5 to 80/30/10.
/:cl:


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -109. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

RenGusta:
Future changes to shotgun would have attempted two ammos anyway so this is beneficial to that.

Flechett has always been a pain to try and balance and when there's three ammos, it's hard to fill the close and long range niche without one ammo pulling too far ahead. 

I have faith this'll benefit shotgun usage in a positive but not overly powerful way.

XSlayer300:
> balance: **Pump shotgun** Flechette is gone. Buckshot deals 200 damage from 240, but has reduced falloff and less projectile for better medium range performance. Slugs changed from 80/40/7.5 to 80/30/10.

What is the point of that?

SilencedMP5A5:
why take a whole 40 damage off buckshot tho

Guidesu:
Why remove a thing if you can fix it?
Please don't be the Orange and try to balance the flechette.
I like using flechettes in normal shotguns.

Rayfod:
I died to a shotgun. Now I have to make sure that shotguns are terrible guns?

IrkallaEpsilon:
Surprise a weapon designed for close quarter battles is good at close quarter battles, why change buckshot in terms of damage?  Flechette has rather specific used (High AP if I am not mistaken) but has this terrible spread kind of deal to hit. So now everyone just defaults to slugs if this gets through because who the hell uses buckshots, specially after lowering the damage. 

Seems like the other questionable removal of "unused guns" (The Kronos was available in the vendor and had some rather specific uses, but by no means was bloat/behaving more or less exactly like the other pump action). 

Slugs are medium range.. Why buff the far range damage but nerf the medium range damage? That makes no sense.

Pariah919:
> Surprise a weapon designed for close quarter battles is good at close quarter battles, why change buckshot in terms of damage? Flechette has rather specific used (High AP if I am not mistaken) but has this terrible spread kind of deal to hit. So now everyone just defaults to slugs if this gets through because who the hell uses buckshots, specially after lowering the damage.
>
to make it more than a melee stick, buckshot with this pr is actually pretty okay statistically.

>Flechette has rather specific used (High AP if I am not mistaken
>
it has 15 ap and is statistically worse than buckshot/slugs in a lot of engagements, its one use is sundering and the autoshotgun does that better.

> balance: Pump shotgun Flechette is gone. Buckshot deals 200 damage from 240, but has reduced falloff and less projectile for better medium range performance. Slugs changed from 80/40/7.5 to 80/30/10.
>
so people wouldn't think that autoshotgun flechette is gone

Guidesu:
JUST FIX IT

AndroidMK2:
I would say that slugs should be buffed actually. They are less effective at close range then buckshot (obviously) and even though they are supposed to be used at slightly further distances they miss sometimes even with attachments. They do hit mostly but when they do it is not enough damage to for example take down a warrior running away after being shot twice before.

TiviPlus:
Having one type be useless is fine and means you can change what is viable to stop the meta from becomingfrom being a perfectly stale boring thing

PR #5455, opened 2020-12-08
Autoburst no longer more accurate than burst

Hughgent:
## About The Pull Request

Somewhat easier to overlook.

## Why It's Good For The Game

Having your weapon on autoburst was more accurate than just burst mode. So you could just single click on autoburst and it would be better.

## Changelog
:cl: Hughgent
fix: Autoburst firemode now properly applying burst accuracy penalty.
/:cl:


David-Stormwell:
If you could also fix some guns simply not working with auto-burst that'd be great.

Hughgent:
whole separate thing.
⚛️

TiviPlus (approved):

PR #5497, opened 2020-12-14
Makes hostile AI ever so slightly faster.

BeloneX:
This is a very minor change, just makes it so that ListTargets(), a proc which finds a first batch of possible targets, only calls oview once, and only on /atom/movables.

### Changelog
🆑 Altoids BeloneX
fix: Hostile mobs are now like 2% faster at finding a target.
/🆑

Ports:
https://github.com/BeeStation/BeeStation-Hornet/pull/3130 
Which is a port of 
https://github.com/yogstation13/Yogstation/pull/10581

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -131,10 +131,8 @@
 			if(can_see(targets_from, HM, vision_range))
 				. += HM
 	else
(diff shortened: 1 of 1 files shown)


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -7. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

TiviPlus (approved):

PR #5504, opened 2020-12-16
fixes xenos ctrl click starting to grab humans on distance.

### LINKED ISSUE #5474: Xenos can ctrl + click dead mobs from a distance (which causes more issues)
Testmerges irrelevant.

**Reproduction:** As a xenomorph, try to ctrl+click a mob _(this was tested with a human, specifically a dead roundstart survivor NPC)_ from a distance. The animation FX for grabbing will successfully appear but no grab will happen.

Besides this; after repeated instances of the above, when the xenomorph tries to grab a mob as per normal methods _(and not the above)_, the grab seems to lock them in place and make them unable to move.

SplinterGP:
## About The Pull Request
Fixes #5474 

## Why It's Good For The Game
bugfix

## Changelog
:cl:
fix: fixes xeno grabbing starting even when not adjacent when grabbing via ctrl click.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xenoattacks.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xenoattacks.dm
@@ -114,10 +114,7 @@
 			if(INTENT_GRAB)
 				if(M == src || anchored)
 					return 0
-
-				if(Adjacent(M)) //Logic!
-					M.start_pulling(src)
-
+				if(M.start_pulling(src))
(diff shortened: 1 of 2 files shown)


Hughgent (approved):
I bet this was a hold over from some kind of telekinesis in old baycode.

TiviPlus (approved):

SplinterGP:
> I bet this was a hold over from some kind of telekinesis in old baycode.

Theres actually a check for adjacent on the grab code but it only happens after the new grab animation thing, so i had to add one before

PR #5507, opened 2020-12-17
Stamina projectiles no longer trigger FF limit

Hughgent:
## About The Pull Request

Tazers are a hell of a drug.

## Why It's Good For The Game

Tazers do stamina damage, as opposed to having a paralyze effect on them. Thus they trigger the FF limit because you usually shoot allies with it.

It seems that admins wish this to not be the case. A side effect is that stamina damage projectiles no longer trigger FF logs as well.

Is this desirable?

## Changelog
:cl: Hughgent
tweak: Stamina damage projectiles no longer count as FF.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -1041,7 +1041,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 	if(. != BULLET_MESSAGE_HUMAN_SHOOTER)
 		return
 	var/mob/living/carbon/human/firingMob = proj.firer
-	if(!firingMob.mind?.bypass_ff && !mind?.bypass_ff && firingMob.faction == faction)
+	if(!firingMob.mind?.bypass_ff && !mind?.bypass_ff && firingMob.faction == faction && proj.ammo.damage_type != STAMINA)
 		var/turf/T = get_turf(firingMob)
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

terranaut1:
I'm the one who asked for this, after some logs where a player was kicked twice because he was goofing around with somebody else shooting each other with tazers. I'm not defending the fucking around with tazers but I don't think it should cause alarm bells to go off and cause heavy-handed FF/grief bans

Hughgent:
the edge case is this.

player 1 and player 2

player 1 has a lasgun set on disable mode, and wishes to grief player 2
There are no admins currently on.
player 1 disables player 2, reloads, keeps disabling.
This can continue indefinitely until player 1 is out of ammo.
player 2 does not ahelp, and instead shoots back with lethal weaponry.

There are no FF logs from Player 1. Player 2 has FF logs.
Salt ensues.

terranaut1:
Correct me if I'm right but the tasing will still show up in the attack logs, right? All it takes for the player defending themselves is to say as much and for an admin to confirm it.

Hughgent:
Tazers shoot tazer ammo, this ammo does Stamina damage. it will not show up as a FF log.

terranaut1:
I specifically meant attack logs, not FF logs. As long as there's some way to confirm it and tazers aren't completely invisible it should be fine.

Issue #5359, opened 2020-11-27
resin.dm runtime


Hughgent:
## Testmerges:

#5285 #5241 #5159 #5217 #5242 #5253

## Runtime:

Runtime Viewer

<<<Refresh

[02:21:07] Runtime in resin.dm, line 96: undefined variable /turf/open/floor/tile/dark2/var/max_integrity
proc name: attack alien (/turf/closed/wall/resin/attack_alien)
usr: MiniMeatwad/(Mature Shrike (Guardian))
usr.loc: (Aerodrome Hangar Hallway (148, 110, 2))
src: the floor (148,109,2) (/turf/open/floor/tile/dark2)
call stack:
the floor (148,109,2) (/turf/open/floor/tile/dark2): attack alien(Mature Shrike (Guardian) (/mob/living/carbon/xenomorph/shrike))
Mature Shrike (Guardian) (/mob/living/carbon/xenomorph/shrike): UnarmedAttack(the floor (148,109,2) (/turf/open/floor/tile/dark2), 1)
Mature Shrike (Guardian) (/mob/living/carbon/xenomorph/shrike): ClickOn(the floor (148,109,2) (/turf/open/floor/tile/dark2), the floor (148,109,2) (/turf/open/floor/tile/dark2), "icon-x=27;icon-y=24;left=1;scr...")
the floor (148,109,2) (/turf/open/floor/tile/dark2): Click(the floor (148,109,2) (/turf/open/floor/tile/dark2), "mapwindow.map", "icon-x=27;icon-y=24;left=1;scr...")

usr: VV PP FLW
usr.loc: VV JMP

Issue #5358, opened 2020-11-27
carbon.dm runtime


Hughgent:
## Testmerges:

#5285 #5241 #5159 #5217 #5242 #5253

## Runtime:

Runtime Viewer

<<<Refresh

[02:25:06] Runtime in carbon.dm, line 455: Cannot read null.title
proc name: on sdd grace period end (/mob/living/carbon/human/on_sdd_grace_period_end)
usr: *no key*/(Johnny 'Unga' Johnson)
src: Johnny \'Unga\' Johnson (/mob/living/carbon/human)
src.loc: null
call stack:
Johnny \'Unga\' Johnson (/mob/living/carbon/human): on sdd grace period end()
/datum/callback (/datum/callback): Invoke()
world: PushUsr(Johnny \'Unga\' Johnson (/mob/living/carbon/human), /datum/callback (/datum/callback))
/datum/callback (/datum/callback): InvokeAsync()
Timer (/datum/controller/subsystem/timer): fire(0)
Timer (/datum/controller/subsystem/timer): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop()
Master (/datum/controller/master): StartProcessing(0)

usr: VV PP FLW

Issue #5357, opened 2020-11-27
chatmessage.dm Runtime


Hughgent:
## Testmerges:

#5285 #5241 #5159 #5217 #5242 #5253

## Runtime:

Runtime Viewer

<<<Refresh

[02:30:31] Runtime in chatmessage.dm, line 149: Cannot read null.seen_messages
proc name: generate image (/datum/chatmessage/proc/generate_image)
usr: JackTheJackhammer/(Jon Spok)
usr.loc: (Colony Southern Clearing (96, 102, 2))
src: /datum/chatmessage (/datum/chatmessage)
call stack:
/datum/chatmessage (/datum/chatmessage): generate image("<IMG CLASS=icon SRC=\[0xc0000c...", Jon Spok (/mob/living/carbon/human), Bill Torres (/mob/living/carbon/human), /list (/list), 50)
world: ImmediateInvokeAsync(/datum/chatmessage (/datum/chatmessage), /datum/chatmessage/proc/genera... (/datum/chatmessage/proc/generate_image), "Wait", Jon Spok (/mob/living/carbon/human), Bill Torres (/mob/living/carbon/human), /list (/list), 50)

usr: VV PP FLW
usr.loc: VV JMP

Issue #5356, opened 2020-11-27
radial.dm Runtime


Hughgent:
## Testmerges:

#5285 #5241 #5159 #5217 #5242 #5253

## Runtime:

Runtime Viewer

<<<Refresh

[02:30:53] Runtime in radial.dm, line 31: Cannot read null.current_user
proc name: Click (/obj/screen/radial/slice/Click)
usr: Grabbler/(Cooper Morgan)
usr.loc: (Underground Hallway NW (71, 50, 2))
src: Rarm (/obj/screen/radial/slice)
src.loc: null
call stack:
Rarm (/obj/screen/radial/slice): Click(the floor (72,50,2) (/turf/open/floor/tile/dark2), "mapwindow.map", "icon-x=18;icon-y=19;left=1;scr...")
Grabbler (/client): Click(Rarm (/obj/screen/radial/slice), the floor (72,50,2) (/turf/open/floor/tile/dark2), "mapwindow.map", "icon-x=18;icon-y=19;left=1;scr...")

usr: VV PP FLW
usr.loc: VV JMP

PR #5513, opened 2020-12-18
Fixes some logged issues

Wayland-Smithy:
## About The Pull Request

Fixes: #5359
Fixes: #5358 
Fixes: #5357
Fixes: #5356
Fixes: #5355 

## Why It's Good For The Game

Less errors.

## Changelog
:cl:
fix: Fixed several null and out of bounds issues.
/:cl:


code/modules/mob/living/say.dm
```diff
@@ -250,8 +250,8 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 		return lowertext(message[1 + length(key)])
 
 /mob/living/proc/get_message_language(message)
-	if(message[1] == ",")
-		var/key = message[1 + length(message[1])]
+	if(length(message) >= 2 && message[1] == ",")
+		var/key = message[2]
```
Hughgent, line 254: I think this was done this way because there could be two or more character radio/language codes.

Would you be so kind as to double check whether there are any in use?

Wayland-Smithy: I did look at all the language keys and all are one character.

Would `length(message[1])` not always be **1** anyway though? If multiple char keys were intended then I think `length(LD.key)` should be used in the for loop to check.

code/_onclick/hud/radial.dm
```diff
@@ -28,6 +28,8 @@ GLOBAL_LIST_EMPTY(radial_menus)
 		closeToolTip(usr)
 
 /obj/screen/radial/slice/Click(location, control, params)
+	if(!parent) //Orphaned somehow
+		return
 	if(usr.client == parent.current_user)
 		if(next_page)
 			parent.next_page()
```
TiviPlus, line 35: Why not just use ?. here
Also how would this get orphaned?

code/game/turfs/walls/resin.dm
```diff
@@ -89,6 +89,8 @@
 	"<span class='xenonotice'>We start to tear down \the [src].</span>")
 	if(!do_after(M, 4 SECONDS, TRUE, M, BUSY_ICON_GENERIC))
 		return
+	if(!istype(src))
+		return
```
TiviPlus, line 93: Annotate this to explain it

code/modules/mob/living/say.dm
TiviPlus: this is fine imo

TiviPlus (changes requested):

code/_onclick/hud/radial.dm
Wayland-Smithy: Not sure where ?. would go. Which line?

No idea how, but it did: `Runtime in radial.dm, line 31: Cannot read null.current_user`

TiviPlus: I'd reccomend reading up on ?. but yes it basically does this

TiviPlus (approved):

code/_onclick/hud/radial.dm
Wayland-Smithy: The operation is self evident, if the var to the left of the `?.` is null return null otherwise continue on to the member field, but since you felt compelled to point it out I'll go ahead and just assume you want it changed and do that.

IMO a separate early return is easier to read.

code/game/turfs/walls/resin.dm
Wayland-Smithy: "this has been resolved"

TiviPlus (approved):

PR #5553, opened 2020-12-24
Adds the Lagcheck command

TiviPlus:
## About The Pull Request

So people can see how laggy serv is from discord

## Changelog
:cl:
add: New !tgs lagcheck command to check server time dilation
code: Moved some code to the correct file
/:cl:


Hughgent (approved):
I like this, however should it not just be part of `!tgs check` instead?

TiviPlus:
tgs check already has a bunch of info and this is a bit more of a niche info and I dont want it bloating too much

Hughgent (approved):

PR #5559, opened 2020-12-24
You cannot headbite someone without a head

### LINKED ISSUE #5301: You can still headbite a decapped character
## Testmerges:
#4707

## Reproduction:

- Decapitate a person

- Headbite, you will be able to do it even without the head

Hughgent:
## About The Pull Request

This one has either a 1 line fix, or a more involved variation. As it currently stands this is the more involved solution.

The 1 line fix is to just add `owner.headbitten = TRUE` to `dropLimb()` for non-synthetics.

The current fix is to remove the brain organ from the mob, and put it into the contents of the new head object. This is not strictly needed as it would only be used for a kind of disembodied head surgery it seems. A holdover from baystation I suspect.

I had considered an additional check in the headbite `can_use()` for head limb status, but I don't think the effort equals the reward in this situation (a slightly more appropriate `to_chat()`).

## Why It's Good For The Game

You cannot headbite someone without a head.
fixes #5301 

## Changelog
:cl: Hughgent
fix: You cannot headbite someone without a head.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -786,6 +786,12 @@ Note that amputating the affected organ does in fact remove the infection from t
 				organ = new /obj/item/limb/head/synth(owner.loc, owner)
 			else
 				organ = new /obj/item/limb/head(owner.loc, owner)
+				var/datum/internal_organ/O
+				O = owner.internal_organs_by_name["brain"] //This removes (and later garbage collects) the organ. No brain means instant death.
+				owner.internal_organs_by_name -= "brain"
+				owner.internal_organs -= O
+				organ.contents += O //stuff it into the head object. Who knows, someone might care in the future.
+				owner.headbitten = TRUE //if the mob doesn't have a head, they can't be headbitten.
 			owner.dropItemToGround(owner.glasses, force = TRUE)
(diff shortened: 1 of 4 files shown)


code/modules/organs/limbs.dm
```diff
@@ -786,6 +786,12 @@ Note that amputating the affected organ does in fact remove the infection from t
 				organ = new /obj/item/limb/head/synth(owner.loc, owner)
 			else
 				organ = new /obj/item/limb/head(owner.loc, owner)
+				var/datum/internal_organ/O
+				O = owner.internal_organs_by_name["brain"] //This removes (and later garbage collects) the organ. No brain means instant death.
```
TiviPlus, line 790: cast as you assign

TiviPlus (changes requested):
Tested?

Hughgent:
Yeah it didn't work out. going with the 'simple' implementation.

TiviPlus (changes requested):
The correct fix would be to have the brain correctly be removed and added when beheaded instead of faking the mob being headbit

Hughgent:
Effort isn't worth the reward in this case.

PR #5571, opened 2020-12-26
moved the Toggle_Middle/Shift Clicking from IC verb to GamePref

BraveMole:
## About The Pull Request

Allows to save your preference of using shift key instead of middle click button to activate abilities as a xenos. 

## Why It's Good For The Game

Too many times i forgot to switch back to shift key, having to do that each time you spawn a new xenos is a pain in the ass. A better solution would be to put that settings in keybinding, but i'm not experienced enough to do that

## Changelog
:cl:
tweak: moved the Toggle_Middle/Shift Clicking from IC to GamePref
/:cl:


BraveMole:
Tested on my own server, i think i did everything right regarding saved files, but a mistake on these is possible. I didn't try loading a savefile from the main branch to try if it works, cause i don't know how

code/modules/client/preferences.dm
```diff
@@ -28,6 +28,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	var/toggles_chat = TOGGLES_CHAT_DEFAULT
 	var/toggles_sound = TOGGLES_SOUND_DEFAULT
 	var/toggles_gameplay = TOGGLES_GAMEPLAY_DEFAULT
+	var/middleshift_button=TRUE
```
Hughgent, line 31: This should not be its own var. it should be part of `toggles_gameplay`

Hughgent (changes requested):

code/modules/client/preferences.dm
BraveMole: Done

code/modules/client/preferences_savefile.dm
```diff
@@ -254,6 +254,7 @@
 	WRITE_FILE(S["auto_fit_viewport"], auto_fit_viewport)
 	WRITE_FILE(S["menuoptions"], menuoptions)
 	WRITE_FILE(S["key_bindings"], key_bindings)
+
```
Hughgent, line 257: not needed

code/modules/client/preferences_savefile.dm
```diff
@@ -266,6 +267,7 @@
 	WRITE_FILE(S["mute_self_combat_messages"], mute_self_combat_messages)
 	WRITE_FILE(S["mute_others_combat_messages"], mute_others_combat_messages)
 
+
```
Hughgent, line 270: Not needed

Hughgent (changes requested):
getting better

code/__DEFINES/preferences.dm
```diff
@@ -65,8 +65,9 @@
 #define TOGGLES_DEADCHAT_DEFAULT (NONE)
 
 #define RADIAL_MEDICAL (1<<0)
+#define MIDDLESHIFTCLICKING (1<<1)
```
Hughgent, line 68: I would appreciate if these were both tabbed out to some common level. Just so long as they are both on the same `(1<<0)` level, not unlike other bitshift defines.

Hughgent (approved):

code/__DEFINES/preferences.dm
BraveMole: tabbed out? I don't understand what you are asking

TiviPlus: Dont

TiviPlus (approved):

PR #5585, opened 2020-12-28
Removes retreat

TiviPlus:
## Why It's Good For The Game

Its only ever used to force an anticlimatic finish and as shitty as shipside is this is even worse, and we'll frankly be better without.
Also means grav levels matter more since people dont just spam the retreat button anymore

## Changelog
:cl:
del: Removes the ability to orbital retreat
/:cl:


BeloneX:
How about you only make it available once a certain % of marines have died? Like there are only 30% of the forces left, it'll spare us of a one-sided fight.

davidchan13:
> How about you only make it available once a certain % of marines have died? Like there are only 30% of the forces left, it'll spare us of a one-sided fight.

Considering it only exists to deprive xenos of a major victory, it should only be useable if active marine pop is sufficiently higher than xeno pop. (2.5:1 or 3:1 humans to xenos)

TiviPlus:
If marines have more than xenos it's a grief tool and if marines are less than xenos it's a cheap copout so it will suck either way

BraveMole:
Why not just making it very long to retreat (and the alert more noticeable for hive)? And maybe give some kind of boost to xenos when the retreat is being called, so instead of anti-climatic end it's frantick hold of FOB

MLGTASTICa:
Good PR , retreat was a dumb idea that stands against the basic principles of TGMC honestly , its a game afterall and letting one side give up instantly ruins the fun for the other , especially when its in the hands of 2 players and not the entire marine team.

jroinc:
What's the point of *not* always being in the lowest level now?

There are only benefits.

davidchan13:
Longer hijack delay I assume?

SkeletalElite:
You get more req points at higher levels and less at lower levels

SkeletalElite:
I was thinking about this and had an idea

Without retreat, if marines lose hte ground, xenos will be able to wait on psychic points until they have a king before they hijack, once that PR is merged.
What if instead retreat was changed to only be doable if there aren't enough marines on the ground to prevent xenos from calling down the dropship.
That way retreat can force xenos to not delay for king but isn't intended to actually be how marines end the round.

Wayland-Smithy:
> xenos will be able to wait on psychic points until they have a king before they hijack

Good. Maybe the marines would then be incentivized to find it in themselves to break out of a fob siege and actually use their impressive arsenal (Rockets, GL, CAS, OB, Mortar, etc..) plus some - _wait for it_ - strategy and tactics instead of deciding that it's better to just totally run away, hope ERT spam sends PMCs, and/or defend cancerous SD hallways for the grand payoff of a DRAW.

Losenis:
> Maybe the marines would then be incentivized to find it in themselves to break out of a fob siege and actually use their impressive arsenal (Rockets, GL, CAS, OB, Mortar, etc..) plus some - _wait for it_ - strategy and tactics instead of deciding that it's better to just totally run away, hope ERT spam sends PMCs, and/or defend cancerous SD hallways for the grand payoff of a DRAW.

If it isn't being made any easier, I don't see how incentivizing it would make it more feasible. Fob sieges already are boring as hell when it drags out to 2 hours and also happen when enough ungas die to the point a push can't be sustained. This is on top of trying to coordinate players who may be in it to shoot benos and have fun rather than be part of a strategy that may turn out boring as hell for them, IE run and hold a miner for points for better armament they may not even be getting due to another squad eventually doing the main push as they protect said miner. You can't court martial players for disobeying orders, even less so when half of them are dead resulting in a fallback to fob.

Retreats may be a symptom of rounds going to the shitter hard enough some feel it's better to just end the round and start a new one than to drag it out for an inevitable result. Keyword being 'some', xenos obviously miss out on playing to the end of the round as well as marines who fortified the ship and then see that their time and effort went down the shitter with the AO retreat, but for others who may have been in it for the planetside part it may stop being fun altogether.

TiviPlus:
I have seen thus far ONE proper usage of retreat and it was during a 2:30 Ice colony round, every single other insance was a cheap copout
And we shouldnt be getting into delayy situations anyway to need this

Losenis:
Due to the King PR, xenos can choose to delay the hijack and wait for King while marines are shipside unable to retreat. Marines can TRY to do something with tadpole/drop pods, but generally if it got to that stage in the first place they generally can't.

SplinterGP:
> Due to the King PR, xenos can choose to delay the hijack and wait for King while marines are shipside unable to retreat. Marines can TRY to do something with tadpole/drop pods, but generally if it got to that stage in the first place they generally can't.
> 
> EDIT: Xenos delayed for over half an hour after destroying fob to get king. Round almost got to 3 hours. Removing evac was a mistake, just bwoink whoever evacs needlessly. You've gotten rid of preventing a round from dragging out forever and an incentive for xenos to not delay to the point marines care more about ending the round than a victory stat.
> 
> "we shouldnt be getting into delayy situations anyway to need this".
> This just created an absurdly big one. At best this was getting rid of a bandaid without actually solving the undelying issues. AT BEST.

don't think thats enough to leave evac as is, since thats a borderline case and xenos can delay even if they dont get king, could even force marines to go back to planetside if they wanted, that mostly turns into an admin matters mostly imo.

SkeletalElite:
> don't think thats enough to leave evac as is, since thats a borderline case and xenos can delay even if they dont get king, could even force marines to go back to planetside if they wanted, that mostly turns into an admin matters mostly imo.

Like I said earlier in the PR retreat just shouldn't be available until xenos are capable of calling down the dropship, just to prevent them from delaying for a king

SplinterGP (approved):

PR #5607, opened 2020-12-31
Flamer change

### LINKED ISSUE #5598: Delayed ammo counters
Ammo counters show old ammo values in some instances:

1) When tactical reloading, the ammo counter does not update until it is refreshed by shooting or unwielding and wielding again. For example, if your gun has 20 bullets left and you tactical reload a 50 magazine in, the ammo counter will still display 20 rounds. Shooting will refresh it to the correct value, but this can get confusing if you're keeping mags that aren't full on you (Happened once when mags for a gun were almost all vended out but other mags with the same ammo were around, so I had to keep them so as to manually load them at fob).

2) TL-84 ammo counter will subtract the ammo taken by the flame before the current one and display said old value. For example, say you have used a short burst that takes only 2 off the fuel tank. The next flame, you take 5, but given the ammo counter being delayed, it will only visibly subtract 2 and display the old value differing from the fuel tank. This gets confusing because examining the flamethrower only shows the hydrocannon ammo.

BraveMole:
## About The Pull Request

Flamer is now a bit easier to use (auto lighting, and a new keybind to use the hydro cannon). 
Added the back fuel tank, 500 unit of fuel on your back that you can directly use as ammo for your flamer if you link it
Fixed #5598, ammo HUD is now up to date

## Why It's Good For The Game

A new option for pyros, it removes the need of changing fuel tanks every so often at the cost of not having any backpack.

Having an hydro cannon keybind just makes thing much easier, and the mechanic of lighting your flamer is just here to kill new comers trying to fire their flame thrower the first time.

## Changelog
:cl:
add: Added back fuel tank, can be attached to flamethrowers to act as a huge fuel tank
tweak: The flamers will now automaticly lit up when they have access to fuel
tweak: You can toggle the hydro cannon of the tl84 using unique action (space bar as defaut)
del: Remove the pyro backpack (replaced by the back fuel tank)
imageadd: Added images of flamers linked to back fuel tank
fix: ammo HUD is now up to date after firing a flamethrower
/:cl:


neohenry:
Yay, keybind for hydro cannon.

mrhugo13:
Slightly worried about the back fuel tank that you can use as an direct ammo feed, mostly because i don't want the flamer to return to it's previous situation where anyone with a flamer & flame-proof armor could just run at the xenos and constantly flame them without having to be worried about being open for attack when they need to reload or such.

BraveMole:
> Slightly worried about the back fuel tank that you can use as an direct ammo feed, mostly because i don't want the flamer to return to it's previous situation where anyone with a flamer & flame-proof armor could just run at the xenos and constantly flame them without having to be worried about being open for attack when they need to reload or such.

Yeah, i guess the pyro backpack was pretty removed in the past. But the large fuel tank already have 75 unit, that's already a lot and doesn't really prevent you from retard pushing. 

Also, since dropping the gun unlink the tank, one CC and you have to run back really quick. That's actually such a drawback of the tank that i'll change the timings a bit, 5 second is really too much for combat sitatuations

BraveMole:
Closing this because i fucked up with me repo and can't access it somehow.

PR #5608, opened 2020-12-31
Tactical reload ammo HUD fix

### LINKED ISSUE #5598: Delayed ammo counters

BraveMole:
## About The Pull Request

Ammo hud is now up to date when tactical reloading, fixes #5598 

## Why It's Good For The Game

No more confusion when reloading, as before you had to fire your gun for the ammo HUD to update

## Changelog
:cl:
fix: Ammo hud is now up to date when tactical reloading
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -348,6 +348,8 @@ User can be passed as null, (a gun reloading itself for instance), so we need to
 			load_into_chamber()
 
 	update_icon(user)
+	var/obj/screen/ammo/A = user.hud_used.ammo
+	A.update_hud(user)
 	return TRUE
 
 /obj/item/weapon/gun/proc/replace_magazine(mob/user, obj/item/ammo_magazine/magazine)


Hughgent:
code looks good. Add more explanation to the comment though. and make the CL more appropriate.

Hughgent:
image

Hughgent (approved):

PR #5623, opened 2021-01-01
Makes gyroscopic stabilizer reduce aim mode movement speed reduction by 50%

Walarks:
## About The Pull Request
Makes gyroscopic stabilizer reduce aim mode movement speed reduction by 50%, but also reduce damage by 10% for every marine passed. 

Damage decrease is multiplicative, so for a bullet dealing 30 damage, it will deal 30 with no marines then 27>24.3>21.87 and so on.

## Why It's Good For The Game
Adds more variety to attachments as sidegrades. Gives the option to be more mobile during aim mode at the cost of decreased damage when firing from the far back of a group, thus possibly rewarding spreading out or moving in smaller groups.

## Changelog
:cl:Slotbot
balance: gyroscopic stabilizer reduces aim mode movement speed reduction by 50%, but also decreases bullet damage by 10% per marine passed
/:cl:

### DIFF
--- a/code/modules/codex/entries/weapons_codex.dm
+++ b/code/modules/codex/entries/weapons_codex.dm
@@ -63,6 +63,8 @@
 		attach_strings += "Wield delay: [wield_delay_mod / 10] seconds."
 	if(attach_shell_speed_mod)
 		attach_strings += "Bullet speed: [attach_shell_speed_mod * 100]%"
+	if(aim_mode_movement_mult)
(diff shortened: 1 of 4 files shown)


SilencedMP5A5:
interesting

TiviPlus (approved):
Good example of attachments niches

tgstation-server:
Your Fix/Feature pull request delta is currently below zero (-10). Maintainers may close future Feature/Tweak/Balance PRs. Fixing issues or helping to improve the codebase will raise this score.

PR #5636, opened 2021-01-02
Weeds nodes get destroyed by plasma cutter

Walarks:
## About The Pull Request
Weeds nodes get destroyed by plasma cutter in 1 hit.

## Why It's Good For The Game
You don't need to take both machete and plasma cutter to clear xeno structures and weeds, since taking both takes a lot of space and carrying them around is cumbersome. A lot of times marines push into areas with xeno walls without clearing them because no one has bothered taking a plasma cutter.

## Changelog
:cl:Slotbot
balance: plasma cutter destroys weeds nodes with 1 hit
/:cl:

### DIFF
--- a/code/game/objects/effects/weeds.dm
+++ b/code/game/objects/effects/weeds.dm
@@ -170,6 +170,14 @@
 	node_turfs = filled_turfs(src, node_range, "square")
 	SSweeds.add_node(src)
 
+/obj/effect/alien/weeds/node/attackby(obj/item/I, mob/user, params)
+	. = ..()
+
+	if(istype(I, /obj/item/tool/pickaxe/plasmacutter))
+		Destroy()
+		playsound(I.loc, 'sound/machines/click.ogg', 25, 1, 5)
+		playsound(src, "alien_resin_break", 25)
+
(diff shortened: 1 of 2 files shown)


davidchan13:
Cutter doesn't need to be the catch all tool that does everywhere.

neohenry:
Time for synthetic to hog the plasma cutter.

Walarks:
> Cutter doesn't need to be the catch all tool that does everywhere.

I think cutter is too niche currently as its only use is to destroy walls, while any other tool that is good at clearing weeds can also be used to fight xenos.

Surrealaser:
> > Cutter doesn't need to be the catch all tool that does everywhere.
> 
> I think cutter is too niche currently as its only use is to destroy walls, while any other tool that is good at clearing weeds can also be used to fight xenos.

Efficiently destroying resin strucutres/walls is already pretty huge, nevermind breaching shit efficiently.

TiviPlus:
No reason for the cutter to intrude on the machetes niche
If you feel it is TOO situational perhaps make it better within its niche

PR #5664, opened 2021-01-05
Detpack minimum detonation time reduced from 5 to 2 seconds

EvilMelonTheSecond:
## About The Pull Request
Changed both detpack default and minimum timer values from 5 to 2

## Why It's Good For The Game
Detonation packs see no use by anyone ever. They are easily seen by xeno and give a loud warning on top of the slow detonation time. Their price is probably still not worth it but hopefully this will give it some use.

## Changelog
:cl:
balance: Detpack detonation timer is buffed.
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -105,7 +105,7 @@
 //=================================================
 
 //Define detpack
-#define DETPACK_TIMER_MIN			5
+#define DETPACK_TIMER_MIN			2
 #define DETPACK_TIMER_MAX			300
 
 //Define sniper laser multipliers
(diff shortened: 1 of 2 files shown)


Chaznoodles:
Turns them into guaranteed explosion traps, no thanks.

MLGTASTICa:
Good change , xenos should be punished for walking into detpacks.

Surrealaser:
Reduce the cost, or increase the number in a pack; they are way too powerful for a 2 second min timer.

TiviPlus:
I would rather see making them cheaper rather than making them activate much faster

PR #5709, opened 2021-01-11
Adds Plasma Drain Grenades to Req

SkeletalElite:
## About The Pull Request
You can now buy more plasma drain grenades from req for 70 points.

## Why It's Good For The Game

For a time we had so many of these things in vendors that it didn't matter, but now that they are more limited in supply they can run out especially if someone is using them a lot. This adds a way to get more of them. Price is 70 because you get more per box than you do with WP, but I also didn't want the price to be so high people just opt for murder weapons like the sadar instead. 

## Changelog
:cl:
add: You can buy more plasma drain grenades at requisitions now
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -275,6 +275,12 @@ WEAPONS
 	contains = list(/obj/item/storage/box/nade_box/phos)
 	cost = 70
 
+/datum/supply_packs/weapons/explosives_plasmadrain
+	name = "M40-T gas grenade box crate"
+	notes = "Contains 25 grenades"
+	contains = list(/obj/item/storage/box/nade_box/plasma_drain_gas)
+	cost = 70
+
 /datum/supply_packs/weapons/plastique
 	name = "C4 plastic explosive"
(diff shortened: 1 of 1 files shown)


BeloneX:
The price is way too high honestly, make it more like 30 - 50.

BraveMole:
Price seems a bit too high for me too, 50 could be a good compromise?

SkeletalElite:
> The price is way too high honestly, make it more like 30 - 50.

30 is way too low, the regular old incendiary and HIDP grenades go for 50 a box which is 2 points per grenade
The m15 frags also cost 50 but you onlu get 15 of them for 3.33 points per grenade
WP gets 15 for 70 which is 4.6 points per nade.

With these you're getting 25 for 70 points so at 2.8 these are cheaper to buy than frag grenades.

I was honestly concerned my price could be too low, although I do think the other grenades are bit on the expensive side and maybe could be toned down.

Chaznoodles:
70 points is a good price. Tanglefoot completely shuts down xenos pushes/defences on the same level as WP does, whilst not stopping marines from pushing through it.

TiviPlus (approved):

PR #5738, opened 2021-01-14
Fix description

TGCMbetcher:
## About The Pull Request

Fixed the description on the Mark 1 Baldur Light Amplification System to match what it does. If you want it to match the rest of the mark ones the slowdown should be set to 0.3 but then literally no one would ever pick it. I just made the description different from the parent to match what it does in game.

## Why It's Good For The Game

Makes the description accurate so people clearly understand what happens when you use it.

## Changelog
:cl:
spellcheck: Removed four words falsely stating a slowdown
/:cl:

### DIFF
--- a/code/modules/modular_armor/attachables.dm
+++ b/code/modules/modular_armor/attachables.dm
@@ -50,7 +50,7 @@
 
 /obj/item/armor_module/attachable/better_shoulder_lamp/mark1
 	name = "\improper Mark 1 Baldur Light Amplification System"
-	desc = "Designed for mounting on the Jaeger Combat Exoskeleton. Substantially increases the power output of the Jaeger Combat Exoskeleton's mounted flashlight. Slows you down minorly."
+	desc = "Designed for mounting on the Jaeger Combat Exoskeleton. Substantially increases the power output of the Jaeger Combat Exoskeleton's mounted flashlight."
 	power_boost = 3 /// The boost to armor shoulder light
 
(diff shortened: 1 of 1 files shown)


TGCMbetcher:
Splynter, from brazil walked be through how to do the gitkraken stuff or someone else would have had to so this so thanks for that.

code/modules/modular_armor/attachables.dm
```diff
@@ -50,7 +50,7 @@
 
 /obj/item/armor_module/attachable/better_shoulder_lamp/mark1
 	name = "\improper Mark 1 Baldur Light Amplification System"
-	desc = "Designed for mounting on the Jaeger Combat Exoskeleton. Substantially increases the power output of the Jaeger Combat Exoskeleton's mounted flashlight. Slows you down minorly."
+	desc = "Designed for mounting on the Jaeger Combat Exoskeleton. Substantially increases the power output of the Jaeger Combat Exoskeleton's mounted flashlight."
```
TiviPlus, line 53: Due to inheritance this change is incorrect as it inherits the slowdown from its parent
However you could instead change it back and note that it increases light radius by less

TiviPlus (changes requested):

code/modules/modular_armor/attachables.dm
TGCMbetcher: Can do

PR #5757, opened 2021-01-17
Lowers Sentry Costs

Chaznoodles:
## About The Pull Request

Lowers sentry costs to 80/40 for big/small.

## Why It's Good For The Game

Encourages marines to not cadehug, as sentries are more easily available to cover unmanned barricade portions in their absence. If it doesn't, revert this. Works in conjunction with #5758 

## Changelog
:cl:
tweak: Lowers Requisitions sentry costs.
/:cl:

PR #5758, opened 2021-01-17
Lowers Sandbag Health

Chaznoodles:
## About The Pull Request
Lowers sandbag health from 400 to 200, in line with metal barricades.

## Why It's Good For The Game

Sandbags are no longer twice as effective as metal barricades, for a tiny portion of the cost, requiring no skill to construct, and are put down far faster. These are supposed to be the 'no skill easy cades', not supreme, also something something deescalation. Works in conjunction with #5757 .

## Changelog
:cl:
tweak: Sandbag integrity halved.
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -1065,7 +1065,7 @@
 	name = "sandbag barricade"
 	desc = "A bunch of bags filled with sand, stacked into a small wall. Surprisingly sturdy, albeit labour intensive to set up. Trusted to do the job since 1914."
 	icon_state = "sandbag_0"
-	max_integrity = 400
+	max_integrity = 200
 	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
 	coverage = 128
 	stack_type = /obj/item/stack/sandbags


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -19. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

TGCMbetcher:
Are you going to cancel the doubling of the req price? Your argument for that was sandbags have twice the health and you can split the labor time before you put it up. Something you can also do with metal but people are too lazy for that, instant cades instead of quick cades if you unwrench it and then rewrench it.

InterroLouis:
Isn't there another downside for sandbags? You know, the whole "any repairs at all will always cost materials" thing. Metal barricades can be fixed with a welder up to a certain amount of damage.

Chaznoodles:
> Are you going to cancel the doubling of the req price? Your argument for that was sandbags have twice the health and you can split the labor time before you put it up. Something you can also do with metal but people are too lazy for that, instant cades instead of quick cades if you unwrench it and then rewrench it.

That was closed, so this is the result.

TGCMbetcher:
> That's not a particular downside, as much like with constructing the sandbags, you need zero skill to do so. You're also getting 50 sandbags for 10 points still, so you should have a metric ton spare unless you're using every single one to make barricades. There's no real downside to sandbags in current state.

It takes zero skill to weld metal barricades, only plasteel require skill checks to be passed. Its why xenos target plasteel before metal because the limited resource that is engineers are required to fix them or they can permanently damage them to the point of requiring materials to fix before a marine can weld it up. I know, Ive tested on big red and lost to a spitter when we started after hits second spit. If marines bothered to keep spare cades around to be moved into place of a broken one they would also be far better then sandbags, but that requires not unging for 2-3 minutes and a rotate macro.

Chaznoodles:
Closing, no desire to contribute due to repeated instances of behaviour like below.

image

PR #5777, opened 2021-01-20
Both Marines and Xenos now die almost instantly in area/space

mrhugo13:
## About The Pull Request
Adds code for sea tiles and `area/space` so that both instantly kills you if you enter either, bad idea spessman, also gives them the reason for their death when they die as to help them not repeat it.

dreamseeker_2021-01-21_13-47-44
Adds tip to inform both marines and xenos that going into pressure-less (a.k.a spess) areas results in almost immediate death.

## Why It's Good For The Game
Under no circumstances should either a xeno or a marine be in a pressure-less (a.k.a spess) area for any reason, be it groundside or shipside, especially on shipside as there is usually a place shipside where there is 0 pressure but marines/xenos cannot die unless they go into a space tile.

## Changelog
:cl: Vondiech
tweak: Both humans and Xenos now die instantly in pressureless areas (a.k.a space) and the sea/outside on prison station now instantly kills you aswell.
/:cl:

### DIFF
--- a/code/game/turfs/open.dm
+++ b/code/game/turfs/open.dm
@@ -171,6 +171,7 @@
 /turf/open/beach/sea/Entered(atom/movable/AM, atom/oldloc)
 	. = ..()
 	if(isliving(AM))
+		to_chat(AM, "<span class='danger'>The cold vacuum instantly freezes you, maybe this was a bad idea?</span>")
 		var/mob/living/spaceman = AM
 		spaceman.adjustFireLoss(666) //Really bad idea, spessman.
 
--- a/code/game/turfs/open.dm
+++ b/code/game/turfs/open.dm
@@ -168,6 +168,11 @@
 	mediumxenofootstep = FOOTSTEP_WATER
 	heavyxenofootstep = FOOTSTEP_WATER
 
+/turf/open/beach/sea/Entered(atom/movable/AM, atom/oldloc)
+	. = ..()
+	if(isliving(AM))
+		var/mob/living/spaceman = AM
+		spaceman.adjustFireLoss(666) //Really bad idea, spessman.
 
 //Nostromo turfs
 
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -30,6 +30,7 @@
 		handle_aura_emiter()
 
 	handle_aura_receiver()
+	handle_environment()
 	handle_living_sunder_updates()
 	handle_living_health_updates()
 	handle_living_plasma_updates()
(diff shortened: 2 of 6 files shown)


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -15. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -30,6 +30,7 @@
 		handle_aura_emiter()
 
 	handle_aura_receiver()
+	handle_environment()
```
TiviPlus, line 33: This shouldnt be on Life() because it causes extra processing

TiviPlus (changes requested):

code/modules/mob/living/carbon/xenomorph/life.dm
mrhugo13: done

SplinterGP:
good idea to add an overlay on screen when you step on those, be like, a 5 seconds timer on screen like battlefield OUT OF MISSION AREA or something

mrhugo13:
> 
> 
> good idea to add an overlay on screen when you step on those, be like, a 5 seconds timer on screen like battlefield OUT OF MISSION AREA or something

~~I don't have the coder prowess to achieve such a feat and~~ The issue is that you wouldn't be on any of these tiles unless you know what you were doing, which most likely would be (atleast for xenos) to get around any issues ever with marines.

SplinterGP:
> > good idea to add an overlay on screen when you step on those, be like, a 5 seconds timer on screen like battlefield OUT OF MISSION AREA or something
> 
> ~I don't have the coder prowess to achieve such a feat and~ The issue is that you wouldn't be on any of these tiles unless you know what you were doing, which most likely would be (atleast for xenos) to get around any issues ever with marines.

actually theres a few times that alamo opens a hole to space when it lands on shipside for example, and other few things, so i'd rather have that than not.

code/game/turfs/open.dm
```diff
@@ -168,6 +168,11 @@
 	mediumxenofootstep = FOOTSTEP_WATER
 	heavyxenofootstep = FOOTSTEP_WATER
 
+/turf/open/beach/sea/Entered(atom/movable/AM, atom/oldloc)
+	. = ..()
+	if(isliving(AM))
+		var/mob/living/spaceman = AM
+		spaceman.adjustFireLoss(666) //Really bad idea, spessman.
```
TiviPlus, line 175: Add a to_chat warning so they know why they died
and or deal the spacemans health as damage so they dont just instantly implode

TiviPlus (changes requested):

code/game/turfs/open.dm
mrhugo13: `to_chat(AM, "<span class='danger'>The cold vacuum instantly freezes you, maybe this was a bad idea?</span>")`

mrhugo13:
> 
> 
> > > good idea to add an overlay on screen when you step on those, be like, a 5 seconds timer on screen like battlefield OUT OF MISSION AREA or something
> > 
> > 
> > ~I don't have the coder prowess to achieve such a feat and~ The issue is that you wouldn't be on any of these tiles unless you know what you were doing, which most likely would be (atleast for xenos) to get around any issues ever with marines.
> 
> actually theres a few times that alamo opens a hole to space when it lands on shipside for example, and other few things, so i'd rather have that than not.

Usually from what i've seen, this allways opens holes to just outright space instead of tiles where there's something under it but it's still an `area/space`

TiviPlus (approved):

tgstation-server:
Your Fix/Feature pull request delta is currently below zero (-17). Maintainers may close future Feature/Tweak/Balance PRs. Fixing issues or helping to improve the codebase will raise this score.

PR #5781, opened 2021-01-20
removes overclocker upgrade for miners

Pariah919:
## About The Pull Request
hmm today I will choose between my miner producing faster, giving me smaller boxes when the asrs pad exists or more health
yeah the choice is pretty obvious, still might be pretty obvious still after this removal, but honestly miners don't need to produce faster

## Why It's Good For The Game
miners shouldn't produce any faster in any circumstance

## Changelog
:cl:
del: Removed overclocker for miners
/:cl:

### DIFF
--- a/code/game/objects/items/misc.dm
+++ b/code/game/objects/items/misc.dm
@@ -164,12 +164,6 @@
 	icon_state = "mining_drill_reinforceddisplay"
 	uptype = "reinforced components"
 
-/obj/item/minerupgrade/overclock
-	name =  "high-efficiency drill"
-	desc = "A box with a few pumps and a big drill, meant to replace the standard drill used in normal mining wells for faster extraction."
-	icon_state = "mining_drill_overclockeddisplay"
-	uptype = "high-efficiency drill"
-
 /obj/item/minerupgrade/compactor
 	name = "crystalizer module"
 	desc = "A bulky module meant to replace the normal crystalizer in mining wells, used to compress boxes for easy carrying."
--- a/code/game/objects/machinery/autolathe_datums.dm
+++ b/code/game/objects/machinery/autolathe_datums.dm
@@ -124,11 +124,6 @@ GLOBAL_LIST_EMPTY(autolathe_categories)
 	path = /obj/item/minerupgrade/compactor
 	category = "Engineering"
 
-/datum/autolathe/recipe/miningwelloverclock
-	name = "mining well overclock upgrade"
-	path = /obj/item/minerupgrade/overclock
-	category = "Engineering"
-
 /datum/autolathe/recipe/airalarm
 	name = "air alarm electronics"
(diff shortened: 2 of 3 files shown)


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -199. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

BraveMole:
Stop it! Req stuff is cool to play with/ play against, let req be rich and everyone has more fun

Losenis:
This honestly feels like it'll lead to req denying all requests to save money on the inevitable things they'll get, as well as fortifying the shit out of miners to prevent their fall given how they'll need them even more.

MLGTASTICa:
Should just make it reduce the time less , currently it reduces the time to mine by 70 seconds , which is kinda too much , you could do like 20-35 secs and not a full removal.

Pariah919:
no point to bothering

Losenis:
I thought I had posted this here too rather than just on discord,

The problem is the other two hardly give a benefit to match. More hp is only useful in a xeno going in and slashing the miner while ignoring the marines shooting at it, otherwise the additional hp wouldn't make a difference if marines are pushed away from it. Compactor is also less useful than it seems because crates are also fast to move if you're pushing them, and you can move 2 at a time by pushing one and pulling another while compact crates already lock both your hands from usage, and you can't push a second crate unless it's from another miner without the module altogether.

Whereas overcharge, even if it gave a pisspoor 10% bonus speed, would have clear benefits that are continuously applied throughout the entire round compared to the other two.  More hp is only used when the miner is getting attacked, gets more damaged than it would normally tank, but less damage than it'd be destroyed by with the upgrade to be useful. Compactor is not even that good due to push+pull, and only saves time on the guys moving it from point A to point B rather than the actual generation of points (IE the bottleneck is the crate generation, not the moving of said crates). At the end of it, it might have little if any impact given crates can just pile up and be moved regardless.
If the miner is packing crates enough to risk the 2 crate cap and stop the generation of more crates, anyone would simply unload them from the miner, put em aside, and start moving them little by little. This'd be faster than compactor too due to the push+pull trick, you move 2 at a time rather than 1 and are able to keep a gun out to emergency wield if a rouny comes around.

So you'd sooner start seeing the metal needed to make the upgrades sent to fob rather than be used in the other two upgrades. Overclock isn't the problem as much as the other two are extremely situational at best.

SkeletalElite:
After balance freeze is up I'm going to make a PR that changes miner upgrades to be bought with req points instead of printed.

The overclock upgrade's price will be quite high so that it's only worth purchasing if you know you'll be able to get long uptime on a miner. Basically buying the upgrade will be a net loss of points until the miner has been running for an amount of time. I haven't thought up how long that time should be, it will depend on the other miner change PRs that are up.

TiviPlus:
Nerf it heavily instead

PR #5784, opened 2021-01-21
Miner Tweaks

SkeletalElite:
## About The Pull Request
This does a few things
1) Increases miner crate capacity to 8 before it stops making more instead of 2
2) Makes the compactor upgrade the default behaviour
3) Nerfs the overclock upgrade which previously reduced production time from 140 seconds to 70 seconds. It now reduces to 100 seconds.

## Why It's Good For The Game
No one wants to be the guy who sits there dragging crates to an ASRS pad all round. It's boring and tedious. This makes selling crates less tedious and makes it need to be done less often. The miner holds enough crates in it you can get away with selling once every ~13 minutes on an overclocked miner or every ~19 minutes on a non overclocked one.
Nerfed the overclock upgrade a little bit because doubling the output of miners is actually insane (no wonder no one used the other upgrades.) It's still probably the best upgrade because resilience just isn't very good. I have some ideas for more stuff to do with upgrades in the future to make them a little more desirable compared to this one, but that's for another day.

## Changelog
:cl:
balance: Made the overclock upgrade a little less effective
tweak: The miner compactor upgrade is now the default behavior. The old upgrade has been removed.
tweak: Miners can hold more ore before being emptied
/:cl:

### DIFF
--- a/code/game/objects/items/misc.dm
+++ b/code/game/objects/items/misc.dm
@@ -169,9 +169,3 @@
 	desc = "A box with a few pumps and a big drill, meant to replace the standard drill used in normal mining wells for faster extraction."
 	icon_state = "mining_drill_overclockeddisplay"
 	uptype = "high-efficiency drill"
-
-/obj/item/minerupgrade/compactor
-	name = "crystalizer module"
-	desc = "A bulky module meant to replace the normal crystalizer in mining wells, used to compress boxes for easy carrying."
-	icon_state = "mining_drill_compactordisplay"
-	uptype = "upgraded crystalizer module"
--- a/code/game/objects/machinery/autolathe_datums.dm
+++ b/code/game/objects/machinery/autolathe_datums.dm
@@ -119,11 +119,6 @@ GLOBAL_LIST_EMPTY(autolathe_categories)
 	path = /obj/item/minerupgrade/reinforcement
 	category = "Engineering"
 
-/datum/autolathe/recipe/miningwellcompactor
-	name = "mining well compactor upgrade"
-	path = /obj/item/minerupgrade/compactor
-	category = "Engineering"
-
 /datum/autolathe/recipe/miningwelloverclock
 	name = "mining well overclock upgrade"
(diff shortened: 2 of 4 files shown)


SkeletalElite:
Also I didn't remove the old crates because I didn't know if they were used anywhere else but if they're not I can just remove them

Pariah919:
Miners don't need more straight buffs right now, the old behavior was fine for what it was.

Swagile:
id say 4 is better, so if your not actively emptying it then its on you

not babysitting miners is nice, but removing the overclock upgrade in compensation is fair id say because otherwise you make it so 1 FOB miner is good enough for the entire round when marine should be capturing more miners instead

SkeletalElite:
> id say 4 crates is better, so if your not actively emptying it then its on you

The point is you don't need someone actively emptying it all the time. You can put up some static defenses and just have someone occasionally come by and sell the crates. I see it the same as cap gameplay. Just how no one wants to spend their round just watching a marine and keeping him stunned for excessive amounts of time, no one wants to sit around at uncontested miners selling crates while fighting is happening elsewhere. 

> not babysitting miners is nice, but removing the overclock upgrade in compensation is fair id say because otherwise you make it so 1 FOB miner is good enough for the entire round when marine should be capturing more miners instead

Yeah that's fair, although this doesn't really affect the amount of money being made by a single miner, since a single miner right now (even with the unnerfed overclock upgrade) doesn't get bottlenecked by a ASRS cooldown since it takes longer to make a crate than it does to cooldown by a pretty significant margin. You only start to get bottle necked with 2 miners and even then they both have to be upgraded for that the crates to start stacking up.

BraveMole:
This with the other miner nerf would be fair

SplinterGP:
i'll think on this, seems reasonable.

SkeletalElite:
I've thought about this and realized this doesn't fully address the issue I have with miners, I'm going to be doing something else after this balance freeze is up.

PR #5663, opened 2021-01-05
Tick allocation refactor

TiviPlus:
TG pr is technically but this was previously broken so I need a round of testing before I PR it to TG

:cl:
refactor: refactored the MC to better compensate for lag caused by byond processes
/:cl:

PR #5787, opened 2021-01-21
Rework some tick defines to better yield to byond's processing needs

MrStonedOne:
This is a copy of an idea I had for /tg/, that doesn't work over there because the mc needed more time then this would allow it to get. 

Looking at tgmc performance on high pop, this will work over here, as most of the cpu time is spent on byond processing (50 to 90%), and most mc usage of cpu time is spikey, meaning it would benefit from automagically spreading cpu across ticks from yielding to byond more aggressively.


MrStonedOne:
oh hi mark #5663

TiviPlus (approved):

PR #5789, opened 2021-01-22
rework/balance on the T90, and a touch in the T19

fernandos13:
## About The Pull Request
this should make the T90 an option again
wielding T19 now is not detrimental (it will HOLD the same stats as un wielded)
TTK test a LOT
and accuratte up TO 5 tiles "to quote from ammo datum itself `accurate_range = 5` "
from 4 tiles of distance its roughly 6 (above the T12 4.5)
and 1 tile its roughly 4 (right under the T12 4.5)

i also balanced around attachies now letting you use the underbarrel for
gyro
or Agrip (two musts to choose which role will the T90 fit for you)

Agrip siding with the 2 hand side of will get the aim time. back to the original T90 aim time. and leaving the barrel free for a EXT or a BC

while the gyro will catter to the one hand side of SMGs "BUT you will need to use a supressor. or a recoild comp too" and aim will be lacking. unless you give up on your mag harness for a red dot

so its better up close than the all rounder T12, still not a shotty tho
and its worst at far away than.. any rifle... but best than shotties too "unless you get flechett then.. you might have some.. preferences to decide which"

aka
nerfs the damage fall of giving it " `damage_falloff_mult = 1.75`  "  rifles have at BEST 1 and shotties got usually 4 (flechette being the exception) 

gives SOME recoil in wielded and changes the undwielded ammoung "attachies WILL and should solve it"

give 10% damage buff. so i can tweak the T90 without touching the T19
BUT also gives the fall off dont forget that

gives aim at the cost of a HUGE walking speed and fire speed "fire speed being one of the most important factor of this gun"

Time To Kill [tested till crit] (times it takes to kill a mature drone)

- at PB
T12
4.4 seconds
T90
3.2 seconds

- at 5 tiles
t12
4.8
T90
4.9

- 7 tiles
T12
5.6
T90
7.2 

## Why It's Good For The Game
gives the ONLY standard issue SMG some well deserved love, making it more viable in the fights for those using it up close

## Changelog
:cl:
balance: improves the damage up close and nerfs it from far. also rebalance its scatter around the use of mods to compensate for it
/:cl:


tgstation-server:
You currently have a negative Fix/Feature pull request delta of -14. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

Kuro-ai:
Right,the damage mult + the bonkers fire rate you gave it will make the SMG outdamage a ton of guns in the game,I suggest removing the damage mult and putting the firerate at 0.150,which is the old T19 stats,that would be more reasonable for a buff

Pariah919:
While the old smg probably did need a minor buff (probably should let mags fit in black webbing again, and or a minor fire rate increase from like 0.165 to like 0.15 or something) it doesn't need whatever this is, which is literally giving it an incredibly fast ttk for whatever reason at cqc and is just massively powercreeping the gun for whatever reason

It doesn't need angled grip or gyro, that'd both look terrible on the sprite and makes no sense since it just adds another forced attachment to make the gun viable in any way


Just do 0.15 fire rate and 0.2 aim mode instead of massively powercreeping the gun by giving it wacky stats

fernandos13:
thing with this 

> Right,the damage mult + the bonkers fire rate you gave it will make the SMG outdamage a ton of guns in the game,I suggest removing the damage mult and putting the firerate at 0.150,which is the old T19 stats,that would be more reasonable for a buff



not really
if you look at 1 tile away YES
but the more you walk of it fails. up to 2 tiles it gets to the same as t12
and 5 tiles it gets BORDERLINE un usable
and 6 tiles. you will need more than one mag even

so yes its better than the T12 by 0.2/0.4 seconds in the time to kill (at the cost of ammo efficience too)
but it drops a LOT the more you walk away

its made to be a DIFERENT weapin than the T12. albeit valid one to use.

Pariah919:
>not really
if you look at 1 tile away YES
but the more you walk of it fails. up to 2 tiles it gets to the same as t12
and 5 tiles it gets BORDERLINE un usable
and 6 tiles. you will need more than one mag even

so yes its better than the T12 by 0.2/0.4 seconds in the time to kill (at the cost of ammo efficience too)
but it drops a LOT the more you walk away

its made to be a DIFERENT weapin than the T12. albeit valid one to use.
>
okay so if your numbers are right why would I run this over a shotgun, which has higher burst damage, stuns, and deals more damage overall (or just use the t-12 with a masterkey kekw)
nevermind that the pr breaks every other normal smg bullet gun due to breaking the ammo datum itself since you didn't bother to make a new one for some reason

fernandos13:
i was at first changing the damage fall of on the T90 ALONE. but i was told to move it to the ammo. cause in the end smgs should be better at close quarters and worst from afar indeed

code/modules/projectiles/ammo_datums.dm
```diff
@@ -422,7 +422,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	accuracy_var_high = 7
 	damage = 20
 	accurate_range = 5
-	damage_falloff = 1
+	damage_falloff = 2.72 //most SHOTGUNS have 4 so a middle ground for , a lot of testing went in THIS value
```
Pariah919:
```suggestion
	damage_falloff = 1
```

code/modules/projectiles/guns/smgs.dm
```diff
@@ -108,19 +108,26 @@
...
-	scatter = 0
-	fire_delay = 0.165 SECONDS
-	scatter_unwielded = 30
+	scatter = 5 //use the ext barrel for 2 hands a least buut i'm not your boss anyway
+	fire_delay = 0.132 SECONDS
+	scatter_unwielded = 25 //so you need the recoil comp and the gyro if shooting one handed. any other way you got scatter.
+	recoil_unwielded = 2 //you either get the recoil comp and gyro to one hand or you dont one hand  if you want to use suuupressor you got EVEN more damage fall of
 	aim_slowdown = 0.25
 	burst_amount = 0
+	damage_mult = 1.3 //should give that humpf it lacks. making the TTK better overall
```
Pariah919, line 130: add a falloff multiplier so not every smg has doubled fallof as a result of this pr

Pariah919 (commented):
this pr doubles the falloff of every smg in the game

Chaznoodles:
The damage and firerate buffs are the things I'm concerned with most.

SplinterGP:
i'll look at this later, need to test the gun ingame and all.

PR #5791, opened 2021-01-22
Fixes Tactical Binocular sight check

SkeletalElite:
## About The Pull Request
Tactical Binoculars line of sight check fuctions intuitively. If you see tile you can lase it. If you can't see the tile, you can't lase it.

## Why It's Good For The Game

Unintuitive and inaccurate sight check bad.
I believe hugh originally didn't use view() because he didn't want you to be able to lase through walls with mesons, but meson vision does not work while looking through binoculars, so this is not a concern.

## Changelog
:cl:
fix: Tactical Binoculars line of sight check is now accurate
/:cl:


SkeletalElite:
Ah shit, I fucked this up, will fix

PR #5792, opened 2021-01-22
Better Sight Check for Binos

SkeletalElite:
## About The Pull Request
Tac Binos sight check for lasing uses view() instead of can_see()
Hugh originally used can_see because he didn't want you to be able to bypass the check with mesons, but meson vision does not work while you're in binoculars anyways so it's not a concern

## Why It's Good For The Game
The sight check for binoculars functions more intuitively. No more mashing tiles trying to figure out which one you can "see"

## Changelog
:cl:
fix: The sight check for lasing with tac binos is more intuitive. 
/:cl:

### DIFF
--- a/code/game/objects/items/binoculars.dm
+++ b/code/game/objects/items/binoculars.dm
@@ -64,7 +64,7 @@
 	if(!pa.Find("ctrl") && pa.Find("shift"))
 		acquire_coordinates(object, user)
 		return TRUE
-
+	
 	if(pa.Find("ctrl") && !pa.Find("shift"))
 		acquire_target(object, user)
 		return TRUE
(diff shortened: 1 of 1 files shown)


David-Stormwell:
I'll kiss you harry

code/game/objects/items/binoculars.dm
```diff
@@ -168,7 +168,7 @@
 		to_chat(user, "<span class='warning'>You can't focus properly through \the [src] while looking through something else.</span>")
 		return
 
-	if(!can_see(user, A, 25))
+	if(!(A in view(user,25)))
```
TiviPlus, line 171: This is terribad and should use a proper view check instead of this hellishness
If I cant think of wany when I have time later I'll be ok with removing it

TiviPlus (changes requested):

TiviPlus:
Ok I cant think of a better way but people can feel free to remove the check

PR #5800, opened 2021-01-23
Fixes attachments for M412L1 and M412 Elite

Sulaboy:
## About The Pull Request
Fixes the wrong stock icon showing on the M412L1, and M412 Elite. Allows M412L1 to attach rail-scope, and mini-scope

## Why It's Good For The Game
Correct stock looks nice. M412L1 looks like it was originally meant to fit a scope, but couldn't because of sub types. Eris asked to have this fixed. 

## Changelog
:cl: Sulaboy
fix: M412L1 and M412 Elite now show correct stock
fix: M412L1 can now attach rail-scope, and mini-scope
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -934,7 +934,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	scatter_mod = -10
 	movement_acc_penalty_mod = 0.1
 
-/obj/item/attachable/stock/irremoveable/rifle
+/obj/item/attachable/stock/rifle/irremoveable
 	name = "\improper M412 solid stock"
 	pixel_shift_x = 32
 	pixel_shift_y = 13
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -938,14 +938,8 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	name = "\improper M412 solid stock"
 	icon_state = "riflestock"
 	attach_icon = "riflestock_a"
-	wield_delay_mod = 0.2 SECONDS
-	size_mod = 1
 	pixel_shift_x = 32
 	pixel_shift_y = 13
-	accuracy_mod = 0.05
-	recoil_mod = -3
-	scatter_mod = -10
-	movement_acc_penalty_mod = 0.1
 
 /obj/item/attachable/stock/rifle/marksman
 	name = "\improper T-45 marksman stock"
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -934,10 +934,18 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	scatter_mod = -10
 	movement_acc_penalty_mod = 0.1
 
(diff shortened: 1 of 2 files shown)


TG-Eris:
I approve of this change :D

code/modules/projectiles/gun_attachables.dm
```diff
@@ -934,7 +934,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	scatter_mod = -10
 	movement_acc_penalty_mod = 0.1
 
-/obj/item/attachable/stock/irremoveable/rifle
+/obj/item/attachable/stock/rifle/irremoveable
 	name = "\improper M412 solid stock"
 	pixel_shift_x = 32
 	pixel_shift_y = 13
```
SplinterGP, line 940: instead of switching the typepaths just add what icon it shall use, irremoveable stocks should start using irremoveable typepath so it's acknowledgeable, it also has no stats from the irremoveable stock which is intended, this was changed by another PR and dind't check for sprite change
```suggestion
	pixel_shift_y = 13
	icon_state = "riflestock"
	attach_icon = "riflestock_a"
```

SplinterGP (changes requested):

code/modules/projectiles/gun_attachables.dm
Sulaboy: Changed the type path, and added the rifle stock stats to /irremoveable/rifle.

code/modules/projectiles/gun_attachables.dm
```diff
@@ -934,10 +934,18 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
...
+	icon_state = "riflestock"
+	attach_icon = "riflestock_a"
+	wield_delay_mod = 0.2 SECONDS
+	size_mod = 1
 	pixel_shift_x = 32
 	pixel_shift_y = 13
+	accuracy_mod = 0.05
+	recoil_mod = -3
+	scatter_mod = -10
+	movement_acc_penalty_mod = 0.1
```
SplinterGP, line 948: you don't need those stats, the initial stock for it dindt use to have those, even if it was a typepath of rifle it just had all those as 0, so as a irremoveable parent it doens't need

Sulaboy: Putting all the stats into the weapon itself when it has a irremovable attachment does make sense. I‘ll remove the stats then, thanks.

Sulaboy: Removed these, the size mod, and the wield delay in commit. Thanks again.

code/modules/projectiles/guns/rifles.dm
```diff
@@ -714,10 +714,11 @@
 		/obj/item/attachable/angledgrip,
 		/obj/item/attachable/flashlight,
 		/obj/item/attachable/bipod,
-		/obj/item/attachable/stock/rifle,
 		/obj/item/attachable/compensator,
 		/obj/item/attachable/magnetic_harness,
 		/obj/item/attachable/scope,
+		/obj/item/attachable/scope/marine,
+		/obj/item/attachable/scope/mini,
```
SplinterGP, line 721: commenting for clarification due to balance freeze, those dont fall into balance wise because theyre meant to fit scopes fine, just whoever set those thought it would also allow in subtypes

PR #5805, opened 2021-01-24
Fixes Stacking Explosions on Same Turfs

Surrealaser:
## About The Pull Request

Explosions can no longer stack on the same turfs; there is now a 1 tick delay between explosions happening on the same turf.

## Why It's Good For The Game

Fixes explosion stacking exploits so insta gib traps can't be easily set up using fire/flame throwers/fuel tanks.

https://github.com/tgstation/TerraGov-Marine-Corps/issues/5806

## Changelog
:cl:
fix: Explosions can no longer stack on the same turfs; there is now a 1 tick delay between explosions happening on the same turf.
/:cl:

### DIFF
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -44,3 +44,5 @@
 #define EXPLODE_DEVASTATE 1
 #define EXPLODE_HEAVY 2
 #define EXPLODE_LIGHT 3
+
+#define TURF_EXPLOSION_STACK_FILTER_DEFAULT 4 //Effectively permits all explosions to go through per /turf/ex_act for the purpose of the explosion anti-stacking mechanism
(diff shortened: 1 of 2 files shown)


TiviPlus (changes requested):
Apart from the code being terrible for something that shouldve been a 3-4 line change

Explosion traps are very obvious 
if there is an issue with hidden items then it is an issue with grenades not layering and not the flat explosion nerf on everyone that this PR is

Surrealaser:
> Apart from the code being terrible for something that shouldve been a 3-4 line change
> 
> Explosion traps are very obvious
> if there is an issue with hidden items then it is an issue with grenades not layering and not the flat explosion nerf on everyone that this PR is

Ridiculous.

#1: Literally the only thing this does is stop SIMULTANEOUS explosion stacking, which really only AoE instant death grenade traps can presently do. You can empty a GL on the same tile and every grenade will have its full effect.

#2: Making the trap more obvious by tweaking layering doesn't actually fix the fundamental problem that Marines have access to an insta kill aoe death zone while Xenos,  the individually more valuable opposition, has no equivalent at all.

PR #5857, opened 2021-02-01
Fixes SL Paygrade only working on late join

SkeletalElite:
## About The Pull Request
Right now your rank only changes as SL if you late join. This fixes that
## Changelog
:cl:
fix: SL rank now correctly changes with playtime if you join at round start
/:cl:


SplinterGP:
i think synth also happens ? check other jobs if you can please

SkeletalElite:
> i think synth also happens ? check other jobs if you can please

SL is the only job in marines.dm or shipside.dm that has a !latejoin in it so I think this bug is isolated to SL

SplinterGP (approved):

PR #5876, opened 2021-02-03
removes barrel charger from req vendor

Pariah919:
## About The Pull Request
it shouldn't be there in the first place, its not a decently designed attachment

## Why It's Good For The Game
bc bad, bc die

## Changelog
:cl:
del: Removed barrel charger
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -731,7 +731,6 @@
 		/obj/item/attachable/bayonet = 25,
 		/obj/item/attachable/compensator = 25,
 		/obj/item/attachable/extended_barrel = 25,
-		/obj/item/attachable/heavy_barrel = 25,
 		/obj/item/attachable/suppressor = 25,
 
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
So why not rebalance it ?

Kuro-ai:
> So why not rebalance it ?

because damage mults should not be a thing on attachments,unless someone adds a AP value for attachments BC should be shadowrealmed

SkeletalElite:
could rework it to increase sunder instead of damage.

SplinterGP:
This is good, BC is awfull and shoulnd't be acquirable, also causes people to question and ask of its existence.

SplinterGP (approved):

Swagile:
> 
> 
> could rework it to increase sunder instead of damage.

too many weapons already have sunder, increasing that number to literally every weapon is not good game design; you might as well remove armor in that case then since people will find ways to game BC with other attachments to absolutely shred armor (see: RDS, LS, BC on a T42, cringe moment)

PR #5880, opened 2021-02-03
Remove Sight Check from Tac Binos

SkeletalElite:
## About The Pull Request

Removes the sight check from tac binos

## Why It's Good For The Game
Trying to find a tile you can "see" on the current sight check is inaccurate and CBT
Alternative sight check options are also bad for one reason or another.
Plus the ability to get a lase in a spot you can't see is fine since most fire support is pretty easy for xenos to avoid if they use their brains anyway

## Changelog
:cl:
del: Tactical Binoculars terrible sight check removed
/:cl:

### DIFF
--- a/code/game/objects/items/binoculars.dm
+++ b/code/game/objects/items/binoculars.dm
@@ -64,7 +64,7 @@
 	if(!pa.Find("ctrl") && pa.Find("shift"))
 		acquire_coordinates(object, user)
 		return TRUE
-	
+
 	if(pa.Find("ctrl") && !pa.Find("shift"))
 		acquire_target(object, user)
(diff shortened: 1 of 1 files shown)


SplinterGP:
i'm okay with this being removed tbh, OB's now announce globally when being fired and you can distinct different types of artillery due to different laser colors now, time to ask tibi

SkeletalElite:
> i'm okay with this being removed tbh, OB's now announce globally when being fired and you can distinct different types of artillery due to different laser colors now, time to ask tibi

Removal was okayed #5792

TiviPlus (approved):

PR #5892, opened 2021-02-03
CAS can target OB Beacons

SkeletalElite:
## About The Pull Request
CAS can now choose OB beacons as a target

## Why It's Good For The Game
OB beacons are kinda useless, the only advantage they have compared to using binos or CAS flares is that they can just be planted and left alone but this is almost never useful for OBs and they're only ever just planted inside the FOB to be used for nuking the FOB after losing it or plasma drain OBing the FOB during siege. This at least makes it to where they can be slapped down to give CAS a way to start their fire missions even if it's a non ideal spot. This basically makes OB beacons a more expensive, bigger, clunkier, but more durable CAS flare.

## Changelog
:cl:
add: CAS can target OB beacons now. 
/:cl:


BraveMole:
This makes cas able to launch whenever they want. I think tivi was against this, though I think it's personally a good change.

SkeletalElite:
CAS shouldn't be balanced around the clunkiness of getting someone else to get you a mark once every few minutes, fuel/ammo availability should be what balances it. If there's a problem with CAS's balance it's not going to be caused by this PR, just made more evident as all it takes to accomplish virtually the same thing is one dude with 2 packs of CAS flares throwing a single flare any time you ask him to.

If there is a problem with balance that is made obvious by this PR.
The amount of free CAS ammo given and the cost of additional ammo in the fabricator should be adjusted instead.

Swagile:
im feeling like everything you just said needs to get added BEFORE this PR gets merged

because the only thing holding CAS back right now is the fact you need someone to laze or throw a flare

since CAS currently works based off of what it can see, as long this is merged, CAS will be able to constantly be up and bombing xenos the moment they dare attack a marine

dealing with good CAS players is already bad enough right now as xenos, encouraging it more before nerfing parts of why its so fucking annoying to deal with (see: gau insta killing you if you don't dodge in time while trying to focus a marine down to death, and many more problems)

SplinterGP:
Hmmmmmm, gotta think alot aboutta this, probably needs other CAS stuff done before this is viable

SkeletalElite:
We'll come back to this after my CAS balance PR gets through the pipeline

PR #5935, opened 2021-02-06
Healing Infusion No Longer Totally Sucks

Surrealaser:
## About The Pull Request

1. Healing Infusion ticks increased from 5 to 10.

2. Healing per tick is now 7 + 0.035% of max HP; Sunder restored per tick is now 2.1

3. Healing Infusion cooldown decreased from 30 seconds back down to 5 seconds.

4. Healing 3.125 HP / sec and 0.0375 Sunder / sec increased to 6 HP / sec and 0.72 Sunder / sec (assuming 300 max HP target)

HP Calc: ((6 + 300 * 0.03) * 10) /  (20 + 5)
Sunder Calc:   (1.8 * 10) /  (20 + 5)

vs Drone 6.66 HP / sec and 0.4 Sunder / sec (including cooldown and wind up); with up to a 13.2 HP / sec if the target has both Burn and Brute damage that's unconditional and is all up front.

5. Healing Infusion no longer restores Sunder unless the target is on weeds.

## Why It's Good For The Game

Makes Healing Infusion not totally suck.

## Changelog
:cl:
balance:  Healing Infusion ticks increased from 5 to 10. Healing per tick is now 6 + 0.03% of max HP Sunder restored per tick is now 1.8. Healing Infusion cooldown decreased from 30 seconds back to 5 seconds. Healing Infusion no longer restores Sunder unless its target is on weeds.
/:cl:


TiviPlus (changes requested):
Seperate bug fix and balance please

Surrealaser:
Absolute PITA

code/modules/mob/living/carbon/xenomorph/castes/hivelord/abilities_hivelord.dm
```diff
@@ -447,9 +449,8 @@ GLOBAL_LIST_INIT(thickenable_resin, typecacheof(list(
 
 	patient.remove_filter("hivelord_healing_infusion_outline") //Remove the aura
 
-	health_ticks_remaining = 0 //Null vars
+	health_ticks_remaining = 0 //Null/zero out vars
```
TiviPlus, line 452:
```suggestion
	health_ticks_remaining = 0 //zero out vars
```

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -93,13 +93,14 @@
...
-	if(locate(/obj/effect/alien/weeds) in loc) //Weeds double sunder recovery
+	var/on_weeds = locate(/obj/effect/alien/weeds) in loc
+	if(on_weeds) //Weeds double sunder recovery
 		sunder_recov *= 2
 
 	if(recovery_aura)
 		sunder_recov *= 1 + recovery_aura * 0.1 //10% bonus per rank of recovery aura
 
-	SEND_SIGNAL(src, COMSIG_XENOMORPH_SUNDER_REGEN, src)
+	SEND_SIGNAL(src, COMSIG_XENOMORPH_SUNDER_REGEN, src, on_weeds)
```
TiviPlus, line 103: why

TiviPlus (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/hivelord/abilities_hivelord.dm
Surrealaser: Really?

Chaznoodles:
So this is essentially a revert PR to it's previous state.

Surrealaser:
> So this is essentially a revert PR to it's previous state. In said state, Infusion was completely broken. It was nerfed for a reason.

Try doing math for once.

Surrealaser:
@TiviPlus 

Okay, having actually played with existing healing infusion, and observed it, and gotten feedback after weeks of implementation, the unassailable conclusion is that it sucks. It in fact sucks so bad that it's generally better to roll Drone over HL because the former can heal that much better and more efficiently. We've run the experiment and the results are intolerable awful. I atomized this like you wanted, and this change frankly should have been TM flagged if not merged awhile ago.

code/modules/mob/living/carbon/xenomorph/castes/hivelord/abilities_hivelord.dm
```diff
@@ -410,7 +410,7 @@ GLOBAL_LIST_INIT(thickenable_resin, typecacheof(list(
 	#endif
 
 ///Called when the target xeno regains Sunder via heal_wounds in life.dm
-/datum/action/xeno_action/activable/healing_infusion/proc/healing_infusion_sunder_regeneration(datum/source, mob/living/carbon/xenomorph/patient)
+/datum/action/xeno_action/activable/healing_infusion/proc/healing_infusion_sunder_regeneration(datum/source, mob/living/carbon/xenomorph/patient, on_weeds = FALSE)
```
spookydonut, line 413: dont add defaults like this for signal handlers it makes no sense.

also you can just check if you are on weeds because the signal source is the xeno you want to know if its on weeds.

code/modules/mob/living/carbon/xenomorph/life.dm
spookydonut: you don't need this param, you also dont need to send src as a param.

code/modules/mob/living/carbon/xenomorph/castes/hivelord/abilities_hivelord.dm
spookydonut: replace datum/source with mob/living/carbon/xenomorph/patient

TiviPlus (approved):

PR #5983, opened 2021-02-09
CAS Balance Tweaks

SkeletalElite:
## About The Pull Request
1) Removes the random inaccuracy from all CAS munitions that had it
2) Makes CAS munitions travel time affected by orbit
3) Makes High velocity ammo the same as 30mm but with slightly less travel time
4) Slightly increases the cost of GAU ammo cost
5) Dramatically reduces the amount of free ammo/weapons CAS starts with:
     CAS starts with 1 banshee, 1 widowmaker, 2 boxes of 30mm, 2 HE rocket stacks, 1 GAU, 1 mini rocket, 1 rocket pod on both the Sulaco and the POS. 
6) Replaces the useless medevac with an operating table for the alamo
7) Slightly buffs mini rocket and fatty fly time (7 to 6 seconds at base orbit)
8) Moves the dropship printer closer to the condor on the sulaco
9) The blinking green dot of CAS stays up for the entire duration a CAS projectile is coming in.
10) Makes railgun ammo it's own subtype instead of using high velocity rounds. Still functions the same as it used to.

## Why It's Good For The Game
1) The random inaccuracy made the already more niche rockets and mini rockets unreliable, especially if they had smaller explosions.
2) Gives a benefit to lower orbit which is almost never used because it sucks while simultaneously making CAS not significantly better at higher orbits (since CAS point generation is affected by orbit already, meaning CAS got quite a bit more ammo at higher orbits)
3) The high velocity ammo going faster is not only more intutive but currently this ammo just isn't worth using, it affects a longer area which makes it more dangerous to use near marines and makes it easier for xenos to dodge. It also costs more points than regular 30mm. At least you're getting something valuable for spending twice as many points on your GAU ammo now. 
4) GAU ammo costs 75 instead of 50 because the GAU is too spamable for how useful it is.
5) Part of the reason CAS can be spammed so much is because they just start with so much ammo that you basically never have to buy rockets or mini rockets (which is partially due to how niche they are) unless you want to use a rocket type you dont start with for free (fatty/napalm) and you start with less 30mm because starting 4 was also part of the reason 30mm was so spammable. 
6) The medevac is useless and this should also make MOs wanting to do Alamo surgery  not at odds with POs who desperately want to get their laser gun on the go ASAP
7) The fatty doesn't need extra fly time, it's already a very expensive rocket. Mini rockets having more fly time than the GAU is laughable considering how much less safe it is to use near marines since it blinds them even if you don't hit them.
8) Makes CAS less of a PITA on sulaco since you don't have to walk all the way across the hangar to get ammo to the condor
9) Makes it harder to get hit by a projectile with sheer luck due to the fact that it has such long travel time you couldn't see it coming, but also makes it a little easier for marines to avoid getting FFed.
10) Railgun doesn't get changed by CAS changes.

## Changelog
:cl:
tweak: Sulaco dropship fab is now near the condor
balance: the free dropship medevac has been replaced with an operating table
balance: Orbit level affects how long it takes for CAS munitions to reach the ground
balance: CAS gets much less free ammo. CAS Munition random inaccuracy removed from all current munitions that had it. Minirockets fly time reduced to 6 seconds. GAU Ammo price increased slightly. Fatty fly time reduced to 6 seconds. High velocity 30mm ammo is now the same as 30mm but with less fly time.
balance: The blinking dot that indicates incoming CAS remains until the projectile hits
/:cl:

### DIFF
--- a/code/game/objects/structures/dropship_ammo.dm
+++ b/code/game/objects/structures/dropship_ammo.dm
@@ -12,7 +12,7 @@
 	climbable = TRUE
 	resistance_flags = XENO_DAMAGEABLE
 	///Time before the ammo impacts
-	var/travelling_time = 10 SECONDS
+	var/travelling_time = 10 * (GLOB.current_orbit/3) SECONDS
 	///type of equipment that accept this type of ammo.
 	var/equipment_type
 	var/ammo_count
@@ -21,7 +21,7 @@
 	var/ammo_id
 	///whether the ammo inside this magazine can be transfered to another magazine.
 	var/transferable_ammo = FALSE
-	var/accuracy_range = 3 //how many tiles the ammo can deviate from the laser target
+	var/accuracy_range = 0 //how many tiles the ammo can deviate from the laser target
 	///sound played mere seconds before impact
 	var/warning_sound = 'sound/machines/hydraulics_2.ogg'
 	var/ammo_used_per_firing = 1
@@ -80,13 +80,12 @@
 	icon_state = "30mm_crate"
 	desc = "A crate full of 30mm bullets used on the dropship heavy guns."
 	equipment_type = /obj/structure/dropship_equipment/weapon/heavygun
-	accuracy_range = 0 //always hits
-	travelling_time = 6 SECONDS
+	travelling_time =  6 * (GLOB.current_orbit/3) SECONDS
 	ammo_count = 200
 	max_ammo_count = 200
 	transferable_ammo = TRUE
 	ammo_used_per_firing = 20
-	point_cost = 50
+	point_cost = 75
 	///Radius of the square that the bullets will strafe
 	var/bullet_spread_range = 2
 	///Width of the square we are attacking, so you can make rectangular attacks later
@@ -140,11 +139,11 @@
 	name = "high-velocity 30mm ammo crate"
 	icon_state = "30mm_crate_hv"
 	desc = "A crate full of 30mm high-velocity bullets used on the dropship heavy guns."
-	travelling_time = 6 SECONDS
-	ammo_count = 400
-	max_ammo_count = 400
-	ammo_used_per_firing = 40
-	bullet_spread_range = 5
+	travelling_time = 5 * (GLOB.current_orbit/3) SECONDS
+	ammo_count = 200
+	max_ammo_count = 200
+	ammo_used_per_firing = 20
+	bullet_spread_range = 3
 	point_cost = 150
 
 
@@ -155,14 +154,13 @@
 	name = "high-capacity laser battery"
 	icon_state = "laser_battery"
 	desc = "A high-capacity laser battery used to power laser beam weapons."
-	travelling_time = 1 SECONDS
+	travelling_time = 1 * (GLOB.current_orbit/3) SECONDS
 	ammo_count = 100
 	max_ammo_count = 100
 	ammo_used_per_firing = 40
 	equipment_type = /obj/structure/dropship_equipment/weapon/laser_beam_gun
 	ammo_name = "charge"
 	transferable_ammo = TRUE
-	accuracy_range = 0 //its a laser
 	ammo_used_per_firing = 10
 	warning_sound = 'sound/effects/nightvision.ogg'
 	point_cost = 150
@@ -224,7 +222,7 @@
 	ammo_id = ""
 	bound_width = 64
 	bound_height = 32
-	travelling_time = 6 SECONDS //faster than 30mm rounds
+	travelling_time = 6 * (GLOB.current_orbit/3) SECONDS
 	point_cost = 0
 	ammo_type = CAS_MISSILE
 
@@ -237,7 +235,7 @@
 	name = "\improper AIM-224 'Widowmaker'"
 	desc = "The AIM-224 is the latest in air to air missile technology. Earning the nickname of 'Widowmaker' from various dropship pilots after improvements to its guidence warhead prevents it from being jammed leading to its high kill rate. Not well suited for ground bombardment, but its high velocity makes it reach its target quickly."
 	icon_state = "single"
-	travelling_time = 4 SECONDS //not powerful, but reaches target fast
+	travelling_time = 4 * (GLOB.current_orbit/3) SECONDS //not powerful, but reaches target fast
 	ammo_id = ""
 	point_cost = 150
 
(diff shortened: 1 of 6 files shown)


SkeletalElite:
shit I gotta fix this real quick will reopen when I do

SkeletalElite:
okay all good now

David-Stormwell:
CAS powercrepe.
Problem is CAS already does it's job as a supressing support fire without any counter other then run the fuck away.

code/game/objects/structures/dropship_equipment.dm
```diff
@@ -659,7 +659,7 @@
 		playsound(loc, firing_sound, 70, 1)
 	var/obj/structure/ship_ammo/SA = ammo_equipped //necessary because we nullify ammo_equipped when firing big rockets
 	var/ammo_accuracy_range = SA.accuracy_range
-	var/ammo_travelling_time = SA.travelling_time //how long the rockets/bullets take to reach the ground target.
+	var/ammo_travelling_time = SA.travelling_time * (GLOB.current_orbit/3) //how long the rockets/bullets take to reach the ground target.
```
BraveMole, line 662: That's a hellish of a buff, reducing by 66% the travel time? This is almost death sentence to resting xenos if you are 1 second late to react

BraveMole (changes requested):

code/game/objects/structures/dropship_equipment.dm
SkeletalElite: Keep in mind dropship points are affected by orbit too (that's already in the game) so you're also only able to generate 1/3 of the ammo at orbit 1, if it's too ridiculous I can half the affect of orbit on travel time and AMMO but I think they should both be affected by an equal amount

SkeletalElite:
The blinking green dot of incoming CAS now stays until the CAS projectile hits

code/game/objects/structures/dropship_ammo.dm
```diff
@@ -21,7 +21,7 @@
 	var/ammo_id
 	///whether the ammo inside this magazine can be transfered to another magazine.
 	var/transferable_ammo = FALSE
-	var/accuracy_range = 3 //how many tiles the ammo can deviate from the laser target
+	var/accuracy_range = 0 //how many tiles the ammo can deviate from the laser target
```
SplinterGP, line 24: this aint used anymore so delete it
```suggestion
```

code/game/objects/structures/dropship_ammo.dm
```diff
@@ -140,12 +139,7 @@
 	name = "high-velocity 30mm ammo crate"
 	icon_state = "30mm_crate_hv"
 	desc = "A crate full of 30mm high-velocity bullets used on the dropship heavy guns."
-	travelling_time = 6 SECONDS
-	ammo_count = 400
-	max_ammo_count = 400
-	ammo_used_per_firing = 40
-	bullet_spread_range = 5
-	point_cost = 150
```
SplinterGP, line 148: hm ?

SplinterGP (changes requested):

SplinterGP:
btw take care that the GAU ammo for some reason is also used in railgun(???????????)

code/game/objects/structures/dropship_ammo.dm
SkeletalElite: Deleting point cost was an accident but it's a subtype of regular 30mm, so it inherits everything else from the 30mm

SkeletalElite:
> btw take care that the GAU ammo for some reason is also used in railgun(???????????)

Do you want the same effect with a different sound? Or do you want it to use a different type of projectile?

SplinterGP:
> > btw take care that the GAU ammo for some reason is also used in railgun(???????????)
> 
> Do you want the same effect or do you want it to use a different type of projectile?
> 
> edit; Just in case it wasn't clear, while the railgun technically uses GAU ammo, it can't be reloaded, it just starts with 8000 bullets and once it runs out it's done for the rest of the round.

take care that changing the GAU changes the railgun, so dont make any big changes to it.

code/game/objects/structures/dropship_ammo.dm
```diff
@@ -414,3 +406,46 @@
...
+	var/attack_width = 3
+	travelling_time = 6 SECONDS
+
+/obj/structure/ship_ammo/railgun/detonate_on(turf/impact, attackdir = NORTH)
+	var/turf/beginning = impact
+	var/revdir = REVERSE_DIR(attackdir)
+	for(var/i=0 to bullet_spread_range)
+		beginning = get_step(beginning, revdir)
+	var/list/strafelist = list(beginning)
+	strafelist += get_step(beginning, turn(attackdir, 90))
```
TiviPlus, line 428: What is this copypaste?

code/game/objects/structures/dropship_equipment.dm
TiviPlus: We'll see in tm

TiviPlus (changes requested):

code/game/objects/structures/dropship_ammo.dm
SkeletalElite: Splinter asked to sperate the railgun from GAU so that GAU changes don't affect the railgun. If preferred I can change back to it being a subtype of GAU ammo but if the GAU were to get any changes the railgun would be affected again

Denthamos:
The travel times depending on orbit is a bit...odd, considering when you launch for a fire mission you are already on orbit at the planet and not in space, so should the travel times for ordinance not be the same regardless?  Would it not just take longer to get planetside from high orbit costing more fuel to get down there when you launch from the hangar?

SkeletalElite:
> The travel times for munitions depending on orbit is a bit...odd, considering when you launch for a fire mission you are already on orbit at the planet and not in space, so should the travel times for ordinance not be the same regardless? Would it not just take longer to get planetside from high orbit costing more fuel to get down there when you launch from the hangar?

Explaining how it works is secondary to gameplay here. right now there is no reason to lower ships orbit, the only benefits you get are so minor they're meaningless, while you get actual meaningful downsides. This gives a meaningful benefit to low orbit and extending CAS fly time would just be adding yet another meaningless benefit to lower orbits. 

Call it better/worse targeting data or the CAS plane having range constraints on how far away it can get from the ship.

code/game/objects/structures/dropship_ammo.dm
TiviPlus: Make it a child for now this copypaste is nasty

SplinterGP (approved):

tgstation-server:
Your Fix/Feature pull request delta is currently below zero (-5). Maintainers may close future Feature/Tweak/Balance PRs. Fixing issues or helping to improve the codebase will raise this score.

PR #4800, opened 2020-09-10
Flamer changes

jroinc:
## About The Pull Request
 Changes flamers. Much higher firerate, lowered welding kit capacity.

## Why It's Good For The Game

Flamers are currently inferior to miniflamers in nearly every niche due to reload speed and firerate. This isn't really acceptable for a primary, so this changes that without significantly affecting the DPS. 

You still do roughly the same damage, but a miss isn't a "well fuck, guess I'm getting beat to shit", and you can reliably coat large areas. 

Also, being able to carry around 0.7 Welding Fuel tanks feels wrong so I lowered it to 0.5 Welding Fuel tanks (a 50% increase over a satchel full of fuel tanks) so that you might actually need to restock occasionally.

Also, given as we have 2 backpack tanks with the same capacity and stats, but different sprites, there's a clear niche to move one (ideally the standard pack) to the belt slot with ~350u (for a 50% increase over the 3 tanks you could normally hold there), but mobsprites are hard so I'll leave that to kuro. 

## Changelog
:cl:
balance: Lowered both flamer fire delays from 2 seconds to 0.4 seconds. 
balance: Lowered welding fuel backpacks capacity from 600 and 700 to 500 for both.
/:cl:

PR #5998, opened 2021-02-10
Flamer Change

InterroLouis:
## About The Pull Request

Flamer fire rate changes as intended in #4800 which did not function as, well, intended. Added a reduction in blue delay just so that it doesn't become...well, a lot worse than regular, although it's still 6 times slower than the red fire in terms of fire delay. 
## Why It's Good For The Game

The fire delay changes were approved in that PR, were intended to be added, didn't work, and didn't end up being fixed. So, this. 
## Changelog
:cl:
balance: Flamer fire delay lowered
/:cl:


Losenis:
#4800 was made when blue flame wasn't in. Taking the differences with default flame, without the change we'd have blue flame with 1.75x the fire delay. Pre and Post #4800 has it worked properly, default fire would've been buffed to 1/5 of its cooldown. Calculating for both of these, blue fire would need to have a value of 7 in delay.

Blue fire being 6 times slower makes it extremely bad in comparison unless it's a single usage for a line/chokepoint where area coverage does not matter. Burn level, red has 24 and blue 26, only 1.5x. Burn time, red has 17 and blue has 40, just 2.35x. This, despite having a 6x cooldown between shots. An argument can be made that blue flame is better for maze clearing since all shots would hit immobile structures that can't dodge, since red fire takes 2 bursts to down resin walls whereas blue fire only needs one, but it'd take 6x as long for the type X to be able to get to the next line of resin walls to torch down. For mazes, red would just be infinitely better, since sometimes you can't afford to take too much time. Shorter burn time on red also helps to keep the fire down afterwards for marines to push, so blue flame's bigger numbers aren't always a good thing.

And let us not forget, this is a 100 req point item that locks you out of wearing a backpack.

BraveMole:
You should def clean up all the shit files you added. Close this PR, make a branch out of master, make your change, make a new clean PR

PR #6013, opened 2021-02-11
Flamer Fire Rate Change 2

InterroLouis:
## About The Pull Request

Basically just does what #4800 tried to do to the flamer's fire rate. Also brings blue more in line, comparison-wise, with the new fire rate. It is a 100 point purchase, and has some noticable downsides to go with it. 

## Why It's Good For The Game

This was intended to be done before, and didn't work. So, I'm doing it again. Again again. 

## Changelog
:cl:
balance: Flamer fires faster. 
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1636,7 +1636,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	var/fire_color = "red"
 	var/burnlevel = 24
 	var/burntime = 17
-	var/fire_delay = 20
+	var/fire_delay = 4
 
 /datum/ammo/flamethrower/on_hit_mob(mob/M,obj/projectile/P)
 	drop_flame(get_turf(P))
@@ -1672,7 +1672,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	fire_color = "blue"
 	burnlevel = 36
 	burntime = 40
-	fire_delay = 35
+	fire_delay = 20
 	bullet_color = COLOR_NAVY
 
 /datum/ammo/water


Losenis:
Adding it here again due to possibly affecting blue flame delay

#4800 was made when blue flame wasn't in. Taking the differences with default flame, without the change we'd have blue flame with 1.75x the fire delay. Pre and Post #4800 had it worked properly, default fire would've been buffed to 1/5 of its cooldown. Calculating for both of these, blue fire would need to have a value of 7 in delay.

Blue fire being 6 times slower makes it extremely bad in comparison unless it's a single usage for a line/chokepoint where area coverage does not matter. Burn level, red has 24 and blue 36, only 1.5x. Burn time, red has 17 and blue has 40, just 2.35x. This, despite having a 6x cooldown between shots. An argument can be made that blue flame is better for maze clearing since all shots would hit immobile structures that can't dodge, since red fire takes 2 bursts to down resin walls whereas blue fire only needs one, but it'd take 6x as long for the type X to be able to get to the next line of resin walls to torch down. For mazes, red would just be infinitely better, since sometimes you can't afford to take too much time. Shorter burn time on red also helps to keep the fire down afterwards for marines to push, so blue flame's bigger numbers aren't always a good thing.

And let us not forget, this is a 100 req point item that locks you out of wearing a backpack.

TiviPlus:
This was wayyy to broken in testing, so I think flamer is ifne as is

Losenis:
Xenos didn't the see atrociously large logistic train of bringing welding kits and fuel tanks to the front. I ended up running like 3 of them dry because the ability to spam flames means the ability of covering an area doesn't depend on the guy minmaxing the flames to cover as much as possible with as little as possible. After a while I had to just dump the flamer because fuel wasn't coming in anymore, and that's fuel that welders would use too.

InterroLouis:
It bothers me that this was just closed rather than unTMd without the opportunity to alter values, like other balance based PRs have gotten.

PR #6046, opened 2021-02-14
Nerfs DB damage by 25%

Kuro-ai:
## About The Pull Request

the marine double barrel now deals 25% less damage per shot

## Why It's Good For The Game

A gun that fires faster than the combat shotgun shouldn't be doing full shotgun damage,rocket tag memes begone

## Changelog
:cl:
balance:the marine double barreled shotgun now deals 25% less damage
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -431,6 +431,7 @@ can cause issues with ammo types getting mixed up during the burst.
 	accuracy_mult = 1.15
 	accuracy_mult_unwielded = 0.85
 	scatter = 10
+	damage_mult = 0.75
 	scatter_unwielded = 40
 	recoil = 2
 	recoil_unwielded = 4


David-Stormwell:
Don't think it's needed. It's already in a bad state and doesn't need nerfs as yes, it fires faster then almost any shotgun, but that's because once you empty your two shells in a skrimish you'll have to dodge xenos while reloading, and competent xenos will punish you for missing.

TGCMbetcher:
Why use it over the paladin then. It would be 100% worse in every way over that and the MK221. It has like a .3 actual delay between shots (I dont mean code I mean if you play you can tell there is a delay) then you take 2+ seconds to reload, with stuns restarting the process. Paladin would do that with the same delay but you can keep firing and I swear its actually delay is 1 second because the delay is split between pump and fire but you can pump instantly after firing giving you a .7 delay.

TiviPlus:
It already sucks

PR #6064, opened 2021-02-17
Bucketed autofire system

BraveMole:
## About The Pull Request

Bucketed autofire system very similar to runechat ss. Timer begone

## Changelog
:cl:
code: Bucketed autofire system, can be used to sanitize full auto firing and in the xeno turrets
/:cl:


code/controllers/autofire.dm
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * Maintains a timer-like system to handle autofiring. Much of this code is modeled
+ * after or adapted from the runechat subsytem.
+ *
+ * Note that this has the same structure for storing and queueing messages as the timer subsystem does
+ * for handling timers: the bucket_list is a list of autofire component, each of which are the head
+ * of a circularly linked list. Any given index in bucket_list could be null, representing an empty bucket.
+ */
+SUBSYSTEM_DEF(autofire)
+	name = "Autofire"
```
TiviPlus: Rename this to reflect that its Automated fire rather than automatic fire

code/controllers/autofire.dm
```diff
@@ -0,0 +1,185 @@
...
+	/// Index of the first non-empty bucket
+	var/practical_offset = 1
+	/// world.tick_lag the bucket was designed for
+	var/bucket_resolution = 0
+	/// How many shooter are in the buckets
+	var/bucket_count = 0
+	/// List of buckets, each bucket holds every shooter that has to shoot this byond tick
+	var/list/bucket_list = list()
+
+/datum/controller/subsystem/autofire/PreInit()
```
TiviPlus, line 33: No secondary queue?

code/controllers/autofire.dm
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * This will also account for a shooter already being registered, and in which case
+ * the position will be updated to remove it from the previous location if necessary
+ *
+ * Arguments:
+ * * new_next_fire Optional, when provided is used to update an existing shooter with the new specified time
+ *
+ */
+/datum/component/autofire/proc/enter_subsystem(new_next_fire = 0)
+	var/list/bucket_list = SSautofire.bucket_list
```
TiviPlus, line 106: This is less so enter and schedule

code/controllers/autofire.dm
```diff
@@ -0,0 +1,185 @@
...
+	var/static/datum/component/autofire/shooter
+	if (!resumed)
+		shooter = null
+
+	// Iterate through each bucket starting from the practical offset
+	while (practical_offset <= BUCKET_LEN && head_offset + ((practical_offset - 1) * world.tick_lag) <= world.time)
+		var/datum/component/autofire/bucket_head = bucket_list[practical_offset]
+		if (!shooter || !bucket_head || shooter == bucket_head)
+			bucket_head = bucket_list[practical_offset]//seems redundant?
+			shooter = bucket_head
```
TiviPlus, line 64: ?
What are you doing here?

TiviPlus (changes requested):
You're updating as I review so Ill finish later

spookydonut (changes requested):
nothing is using this subsystem/component, will review when something is

code/controllers/autofire.dm
```diff
@@ -0,0 +1,264 @@
+/// Controls how many buckets should be kept, each representing a tick. Max is ten seconds, this system does not handle next fire of more than 100 ticks in order to have better perf
+#define BUCKET_LEN (world.fps * 1 * 10)
```
SplinterGP, line 2: multiplying by 1 ?
```suggestion
#define BUCKET_LEN (world.fps * 10)
```

code/controllers/autofire.dm
```diff
@@ -0,0 +1,264 @@
...
+/obj/structure/turret_debug/slow
+	name = "debug turret slow"
+	firerate = 25
+
+/obj/structure/turret_debug/Initialize()
+	. = ..()
+	ammo = GLOB.ammo_list[/datum/ammo/xeno/acid]
+	target = locate(x+5, y, z)
+	AddComponent(/datum/component/automatedfire/automatic_shoot_at, firerate, ammo)
+	SEND_SIGNAL(src, COMSIG_AUTOMATIC_SHOOTER_START_SHOOTING_AT, target)
```
Rohesie, line 217: Signals are not supposed to just replace proc calls. They're generic events.
Since this s only used by `/automatedfire` to start shooting it should probably be an initialization parameter instead.

code/controllers/autofire.dm
```diff
@@ -0,0 +1,264 @@
...
+
+/datum/component/automatedfire
+	///The owner of this component
+	var/atom/shooter
+	/// Contains the scheduled fire time, used for scheduling EOL
+	var/next_fire
+	/// Contains the reference to the next component in the bucket, used by autofire subsystem
+	var/datum/component/automatedfire/next
+	/// Contains the reference to the previous component in the bucket, used by autofire subsystem
+	var/datum/component/automatedfire/prev
```
Rohesie, line 103: This shouldn't be a component. It works well as its own structure, a datum, but it does not fulfill the constraints expected for a component.

For the same reason `GetComponent()` is undesired in general, keeping a reference to components is not kosher. In fact, it's worse.

Here you have a little guide that's helpful (though it doesn't cover everything): https://tgstation13.org/phpBB/viewtopic.php?f=5&t=22674

code/controllers/autofire.dm
BraveMole: Yes of course, but this was just a debug turret. I used the same structure than the xenos turret, and these needs to be able to stop and start shooting. I do believe signals are the way to go to communicate between components and parents?

TiviPlus: As bravemole said this is for the debug turret

code/controllers/autofire.dm
```diff
@@ -0,0 +1,264 @@
...
+	shooting = FALSE
+
+/datum/component/automatedfire/automatic_shoot_at/process_shot()
+	if(!shooting)
+		return AUTOFIRE_STOPPED_SHOOTING
+	var/obj/projectile/newshot = new(shooter.loc)
+	newshot.generate_bullet(ammo)
+	newshot.permutated += shooter
+	newshot.fire_at(target, shooter, null, ammo.max_range, ammo.shell_speed)
+	SEND_SIGNAL(shooter, COMSIG_AUTOMATIC_SHOOTER_SHOT_FIRED)
```
TiviPlus, line 262: This signal is not used anywhere?

code/controllers/autofire.dm
```diff
@@ -0,0 +1,264 @@
...
+			if (shooter == bucket_head)
+				break
+
+		// Empty the bucket
+		bucket_list[practical_offset++] = null
+
+/datum/controller/subsystem/automatedfire/Recover()
+	bucket_list |= SSautomatedfire.bucket_list
+
+/datum/controller/subsystem/automatedfire/proc/reset_buckets()
```
TiviPlus, line 90: Docs

code/controllers/autofire.dm
```diff
@@ -0,0 +1,264 @@
...
+	flags = SS_TICKER | SS_NO_INIT
+	wait = 1
+	priority = FIRE_PRIORITY_AUTOFIRE
+
+	/// world.time of the first entry in the bucket list, effectively the 'start time' of the current buckets
+	var/head_offset = 0
+	/// Index of the first non-empty bucket
+	var/practical_offset = 1
+	/// world.tick_lag the bucket was designed for
+	var/bucket_resolution = 0
```
TiviPlus, line 27: This doc is unclear to it's function, more detail please

TiviPlus (changes requested):

code/controllers/autofire.dm
BraveMole: Yeah, i use it in the xenos turret and the gun refactoring. That's just for example purpose, cause that's how communication between the comp and the gun/turret is done

code/controllers/autofire.dm
BraveMole: I actually have no idea, spooky gave me this part of the code. I'll ask him what it does, cause it's not explained better in runechat

code/controllers/autofire.dm
BraveMole: Same that above, spooky didn't document this.

code/controllers/autofire.dm
```diff
@@ -0,0 +1,260 @@
...
+	bucket_resolution = world.tick_lag
+
+/datum/controller/subsystem/automatedfire/stat_entry(msg)
+	..("ActShooters:[bucket_count]")
+
+/datum/controller/subsystem/automatedfire/fire(resumed = FALSE)
+
+	if (MC_TICK_CHECK)
+		return
+
```
TiviPlus, line 49: Why is the first thing here a tick check?

code/controllers/autofire.dm
```diff
@@ -0,0 +1,225 @@
...
+	SIGNAL_HANDLER
+	shooting = FALSE
+
+/datum/component/automatedfire/automatic_shoot_at/process_shot()
+	if(!shooting)
+		return
+	SEND_SIGNAL(shooter, COMSIG_AUTOMATIC_SHOOTER_SHOT_FIRED)
+	next_fire = world.time + shot_delay
+	if(world.time - last_fire != shot_delay)
+		message_admins("the delay between shots was [world.time - last_fire] ticks instead of [shot_delay]")
```
TiviPlus, line 222: Debug stuff comitted

TiviPlus (approved):

code/controllers/autofire.dm
BraveMole: It's for the debug turrets only, so it's no big deal

TiviPlus (approved):

TiviPlus (approved):

mokulus (changes requested):

code/controllers/subsystem/autofire.dm
```diff
@@ -0,0 +1,217 @@
...
+	flags = SS_TICKER | SS_NO_INIT
+	wait = 1
+	priority = FIRE_PRIORITY_AUTOFIRE
+
+	/// world.time of the first entry in the bucket list, effectively the 'start time' of the current buckets
+	var/head_offset = 0
+	/// Index of the first non-empty bucket
+	var/practical_offset = 1
+	///How many buckets exist for one tick
+	var/bucket_resolution = 0
```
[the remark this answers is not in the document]
BraveMole: This was the former description, and tivi didn't like it. It's how many buckets exist for one deciseconds

code/controllers/subsystem/autofire.dm
```diff
@@ -0,0 +1,217 @@
...
+	/// How many shooter are in the buckets
+	var/bucket_count = 0
```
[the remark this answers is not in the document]
BraveMole: This was the former name in runechat and timer

code/controllers/subsystem/autofire.dm
```diff
@@ -0,0 +1,217 @@
...
+		if(!shooter)
+			shooter =  bucket_list[practical_offset]
+			bucket_list[practical_offset] = null
+
+		while (shooter)
+			next_shooter = shooter.next
+			INVOKE_ASYNC(shooter, /datum/component/automatedfire/proc/process_shot)
+
+			SSautomatedfire.bucket_count--
+			shooter = next_shooter
```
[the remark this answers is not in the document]
BraveMole: no, you need to save a reference. Processing shot will remove next and prev from shooter

code/controllers/subsystem/autofire.dm
```diff
@@ -0,0 +1,217 @@
...
+	if (bucket_head.next)
+		next = bucket_head.next
+		next.prev = src
+
+	bucket_head.next = src
+	prev = bucket_head
+
+	//Something went wrong, probably a lag spike or something. To prevent infinite loops, we reschedule it to a another next fire
+	if(prev == src)
+		next_fire = next_fire += 1
```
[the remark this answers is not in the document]
BraveMole: Like we talked on discord, this can happen due to byondism, or bad code (like if the component is schedule twice for the same tick)

> but because you insert elements in the linked list after bucket_head, which means that they won't be processed in > that tick, because shooter is already after this element most of the time. Runechat adds new elements before head

This will never happen unless you schedule a component for the immediate tick, which is not something you should ever do. Replacing the bucket head each time like runechat does is useless for this system

code/controllers/subsystem/autofire.dm
```diff
@@ -0,0 +1,217 @@
...
+		schedule_shot()
+
+///Handle the firing of the autofire component
+/datum/component/automatedfire/proc/process_shot()
+	return
+
+#undef BUCKET_LEN
+#undef BUCKET_POS
```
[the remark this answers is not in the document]
TiviPlus: It's at the end of the relevant code so thats fine

TiviPlus (approved):

PR #6080, opened 2021-02-18
Gets rid of the random rocket pods on sulaco

SkeletalElite:
I don't know how they even got there but I may be responsible

## Changelog
:cl:
fix: Got rid of the mini rocket pods that were on the public engi vendors of sulaco
/:cl:


TiviPlus (approved):
Nice mapmerge

PR #6032, opened 2021-02-13
Removes illegal src edits and adds a lint for them

TiviPlus:
## Changelog
:cl:
code: Removes illegal src edits and adds a lint for them
/:cl:

PR #6081, opened 2021-02-18
Weapon toggles QoL changes

### LINKED ISSUE #6049: Weapon verbs opens a window
## Testmerges: 
#5811 
#5985 
#5959 
#5995 
#5855 

## Reproduction:
Have multiple guns on you, press 'load from attachment' verb, notice a window opening mid-fight.
Have multiple guns on you, press 'field strip' verb, notice a window opening mid-fight.
Have multiple guns on you, press 'unload current magazine' verb, notice a window opening mid fight.
Etc, etc.

https://i.gyazo.com/1409cc35d0e8e4cad411a7fcf5819ea2.png

mokulus:
## About The Pull Request

Weapon toggles (field strip, toggle fire mode, eject mag etc) were implemented as verbs for `/obj/item/weapon/gun`, which leads to issues when you have more than one gun equipped (see #6049). I've added similar verbs for `/mob/living/carbon/human` which pick the active firearm (the one in your current hand or if it's empty in the other hand). This doesn't affect gun in inventory slot, because you wouldn't be able to do anything with it because of later checks, while the old verbs allow you to select it. I've left the verbs for guns so they are still in right click menu. I had to add "(Weapon)" because otherwise both usage tips show in Weapons category tab. I've also simplified the code a bit.

Fixes #6049

## Why It's Good For The Game

The game no longer asks you to choose which weapon to toggle when you have one in hand and one in suit slot (picks in hand), or when you have a weapon in each hand (picks in active hand).

## Changelog
:cl:
add: weapon toggle verbs that prioritize current gun
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -449,7 +449,6 @@ should be alright.
 
 
 /obj/item/weapon/gun/verb/field_strip()
-	set category = null
 	set name = "Field Strip (Weapon)"
 	set desc = "Remove all attachables from a weapon."
 
@@ -559,7 +558,6 @@ should be alright.
 
 
 /obj/item/weapon/gun/verb/toggle_autofire()
-	set category = null
 	set name = "Toggle Auto Fire (Weapon)"
 	set desc = "Toggle automatic firemode, if the gun has it."
 
@@ -591,7 +589,6 @@ should be alright.
 
 
 /obj/item/weapon/gun/verb/toggle_burstfire()
-	set category = null
 	set name = "Toggle Burst Fire (Weapon)"
 	set desc = "Toggle burst firemode, if the gun has it."
 
@@ -623,7 +620,6 @@ should be alright.
 
 
 /obj/item/weapon/gun/verb/toggle_firemode()
-	set category = null
 	set name = "Toggle Fire Mode (Weapon)"
 	set desc = "Toggle between fire modes, if the gun has more than has one."
 
@@ -738,7 +734,6 @@ should be alright.
 
 
 /obj/item/weapon/gun/verb/empty_mag()
-	set category = null
 	set name = "Unload Weapon (Weapon)"
 	set desc = "Removes the magazine from your current gun and drops it on the ground, or clears the chamber if your gun is already empty."
 
@@ -757,7 +752,6 @@ should be alright.
 
 
 /obj/item/weapon/gun/verb/use_unique_action()
-	set category = null
 	set name = "Unique Action (Weapon)"
 	set desc = "Use anything unique your firearm is capable of. Includes pumping a shotgun or spinning a revolver."
 
@@ -776,7 +770,6 @@ should be alright.
 
 
 /obj/item/weapon/gun/verb/toggle_gun_safety()
-	set category = null
 	set name = "Toggle Gun Safety (Weapon)"
 	set desc = "Toggle the safety of the held gun."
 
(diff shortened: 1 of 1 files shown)


SplinterGP:
i know what PR messed up with it, and it was https://github.com/tgstation/TerraGov-Marine-Corps/pull/6032

i think you don't need to put verb on Mob, just need to replace all the stuff in the gun verb with instead of dothing() with G.dothing() needs you to test it though

mokulus:
I advise against doing that for a few reasons. The `set src = usr.contents` seems to be a bug in byond, because it assigns a list with `=`. [Documentation](http://www.byond.com/docs/guide/chap04.html#4.3.3) doesn't seem to mention it. So it's hacky because it uses `=` for list, which I think picks one item semi-randomly. The other problem is with logic - what would happen in logic is that you'd call the verb on a gun, but because of `set src = usr.contents` the `verb` is only allowed on one weapon (so there is no prompt from #6049), which would be the `src` in that call, but then you'd ignore the `src` and get the real active weapon from `usr`, completely ignoring `src`. That's why I think it should be a `verb` on `/mob/living/carbon/human`, because `src` isn't used. This logic also break right click menu, because when I hold two weapons, and want to eject mag from the weapon in non active hand (using the `/obj/item/weapon/gun` version), then the `src` in that call is ignored and I eject from the weapon in active hand.

The `/obj/item/weapon/gun` **could** be removed now. Currently you would use either macro, weapons verb tab or hud buttons right to hands. If right click menu gets added these verbs won't be used anyway, and if keybindings get added they are 
**worse** than the verbs on `/mob/living/carbon/human`.

The only argument in favor of the verbs for guns is when you want a prompt when holding two guns in hands and you want to toggle the non-selected gun, which I don't think people find useful, because instead you can swap hands, then call a macro, then swap again, which is 3 presses instead of 1 press with menu and ok button.

code/modules/projectiles/gun_helpers.dm
```diff
@@ -565,24 +574,27 @@ should be alright.
...
+	var/obj/item/weapon/gun/G = get_active_firearm(usr)
+	if(!G)
 		return
-	automatic_gun.do_toggle_burst(usr)
+	G.toggle_burstfire()
 
 
-/obj/item/weapon/gun/proc/do_toggle_burst(mob/user)
+/obj/item/weapon/gun/verb/toggle_burstfire()
+	set category = null
```
TiviPlus, line 594: Does category not default to null?
Ditto for the rest of the verbs

code/modules/projectiles/gun_helpers.dm
```diff
@@ -454,7 +460,7 @@ should be alright.
 		to_chat(usr, "<span class='warning'>You cannot conceviably do that while looking down \the [src]'s scope!</span>")
 		return
 
-	if(G.overcharge == TRUE)
+	if(src.overcharge)
```
TiviPlus, line 463:
```suggestion
	if(overcharge)
```

TiviPlus (changes requested):
Verbs in general are kindof shitty (as seen by the set src hack you remove here) but this fix is fine on its own

code/modules/projectiles/gun_helpers.dm
mokulus: Right. I thought it would fix the double usage message if verbs for different objects have the same name.

mokulus: Done.

TiviPlus (approved):
Verbs in general are kindof shitty (as seen by the set src hack you remove here) but this fix is fine on its own

TiviPlus:
Wow ok thanks github

PR #6132, opened 2021-02-21
Vending and computer repair fixes

### LINKED ISSUE #6127: Unable to fix vendors / req console
## Reproduction:

Currently all vendors can be put into a permanent de-active/destroyed state once they are knocked over with no way to fix them IC wise.  As a xeno this promotes ignoring everything else shipside if you manage to sneak on board or burst on board, in order to destroy all the important vendors like gun and ammo vendors in prep, or even the req console as it too cannot be fixed when destroyed.  

If you try to fix the vendor it says you must right it, but you are unable to do such at all.

mokulus:
## About The Pull Request

Allows tipped vendors to be put back up. Makes disabled machinery unusable for AI and humans. Fixes #6127.

The tipped vendors are now powered, which allows interacting with them. Interaction will either be tgui (dispensing) if the vendor isn't tipped or putting it back up if it is.
 

## Why It's Good For The Game

Fixes. Allows fixing vendors. Disallows AI from using broken things.

## Changelog
:cl:
fix: disabled computers are now unusable (repair with blowtorch)
fix: tipped over vendors can be put back up
fix: AI can't use disabled machines now
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/vending.dm
+++ b/code/game/objects/machinery/vending/vending.dm
@@ -392,16 +392,17 @@ GLOBAL_LIST_INIT(vending_white_items, typecacheof(list(
 	. = ..()
 	if(.) // Handled by ui_interact
 		return
-	if(tipped_level == 2)
-		if(!iscarbon(user)) // AI can't heave remotely
-			return
-		user.visible_message("<span class='notice'> [user] begins to heave the vending machine back into place!</span>","<span class='notice'> You start heaving the vending machine back into place..</span>")
-		if(!do_after(user, 80, FALSE, src, BUSY_ICON_FRIENDLY))
-			return FALSE
+	if(tipped_level != 2) // only fix when fully tipped
+		return
 
-		user.visible_message("<span class='notice'> [user] rights the [src]!</span>","<span class='notice'> You right the [src]!</span>")
-		flip_back()
-		return TRUE
+	if(!iscarbon(user)) // AI can't heave remotely
+		return
+	user.visible_message("<span class='notice'> [user] begins to heave the vending machine back into place!</span>","<span class='notice'> You start heaving the vending machine back into place..</span>")
+	if(!do_after(user, 80, FALSE, src, BUSY_ICON_FRIENDLY))
+		return FALSE
+	user.visible_message("<span class='notice'> [user] rights the [src]!</span>","<span class='notice'> You right the [src]!</span>")
+	flip_back()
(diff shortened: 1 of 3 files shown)


code/game/objects/machinery/vending/vending.dm
```diff
@@ -386,7 +384,27 @@ GLOBAL_LIST_INIT(vending_white_items, typecacheof(list(
 
 	return TRUE
 
+/// Used only when vendor is tipped to put it back up
+/// Normal usage is in ui_interact
```
TiviPlus, line 388:
```suggestion
/**
 * Used only when vendor is tipped to put it back up
 * Normal usage is in ui_interact
 */
```

code/game/objects/machinery/vending/vending.dm
```diff
@@ -386,7 +384,27 @@ GLOBAL_LIST_INIT(vending_white_items, typecacheof(list(
...
+	if(tipped_level == 2)
+		if(!iscarbon(user)) // AI can't heave remotely
+			return
+		user.visible_message("<span class='notice'> [user] begins to heave the vending machine back into place!</span>","<span class='notice'> You start heaving the vending machine back into place..</span>")
+		if(!do_after(user, 80, FALSE, src, BUSY_ICON_FRIENDLY))
+			return FALSE
+
+		user.visible_message("<span class='notice'> [user] rights the [src]!</span>","<span class='notice'> You right the [src]!</span>")
+		flip_back()
+		return TRUE
```
TiviPlus, line 402:
```suggestion
	if(tipped_level != 2)
		return
	if(!iscarbon(user)) // AI can't heave remotely
		return
	user.visible_message("<span class='notice'> [user] begins to heave the vending machine back into place!</span>","<span class='notice'> You start heaving the vending machine back into place..</span>")
	if(!do_after(user, 80, FALSE, src, BUSY_ICON_FRIENDLY))
		return FALSE

	user.visible_message("<span class='notice'> [user] rights the [src]!</span>","<span class='notice'> You right the [src]!</span>")
	flip_back()
	return TRUE
```

mokulus: I will remove this check because it's already in `ui_interact` which has higher priority.

TiviPlus (approved):

TiviPlus (approved):

Losenis:
Finally there'll be a reason to push for medbay or engi.

PR #6137, opened 2021-02-21
Silo proximity alert fix, balance

BraveMole:
## About The Pull Request

Silo proximity alert will now work. They now detect hostiles with a range of 10 instead of 2

## Why It's Good For The Game

Will give xenos hives early warning to give them time to come defend pool

## Changelog
:cl:
fix: xenos silo proximity alerts now work
balance: range of proximity alerts is now 10 instead of just 2
/:cl:

### DIFF
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -54,7 +54,7 @@
 	if(!istype(center_turf))
 		center_turf = loc
 
-	for(var/i in RANGE_TURFS(2, src))
+	for(var/i in RANGE_TURFS(XENO_SILO_DETECTION_RANGE, src))
 		RegisterSignal(i, COMSIG_ATOM_ENTERED, .proc/resin_silo_proxy_alert)
 
 	return INITIALIZE_HINT_LATELOAD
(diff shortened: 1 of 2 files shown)


code/modules/xenomorph/xeno_structures.dm
```diff
@@ -127,36 +119,21 @@
...
-
-	if(get_dist(loc, hostile) > 2) //Can only send alerts for those within 2 of us; so we don't have all silos sending alerts when one is proxy tripped
+	var/mob/living/carbon/human/nearby_human
+	for (var/mob/living/carbon/human/human as anything in cheap_get_humans_near(src, XENO_SILO_DETECTION_RANGE))
+		if(human.stat != DEAD)
+			nearby_human = human
+			break
+		
+	if(!nearby_human)
 		return
```
mokulus, line 132: I don't think this check is necessary. From what I've read [here](https://github.com/tgstation/tgstation/blob/master/.github/CONTRIBUTING.md#avoid-unnecessary-type-checks-and-obscuring-nulls-in-lists) it seems like there won't be any nulls in that loop.

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -127,36 +119,21 @@
...
-		return
-
-	var/mob/living/living_triggerer = hostile
-	if(living_triggerer.stat == DEAD) //We don't care about the dead
-		return
-
-	if(isxeno(hostile))
-		var/mob/living/carbon/xenomorph/X = hostile
-		if(X.hive == associated_hive) //Trigger proxy alert only for hostile xenos
-			return
```
mokulus, line 146: New code doesn't detect hostile hives.

mokulus (changes requested):

code/modules/xenomorph/xeno_structures.dm
BraveMole: Yes, but HVH sucks. :( lazy me didn't want to add

code/modules/xenomorph/xeno_structures.dm
BraveMole: it's null if there's nothing in cheap_get_humans_near, which happens 99% of the time

code/modules/xenomorph/xeno_structures.dm
mokulus: I think you can do similar loop using `cheap_get_xenos_near`, adding different hive check in `if` condition.

BraveMole: yes i know, but i'm lazy ahah

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -47,16 +47,13 @@
 	name = "[name] [number]"
 	number_silo = number
 	number++
-	RegisterSignal(SSdcs, list(COMSIG_GLOB_OPEN_TIMED_SHUTTERS_LATE, COMSIG_GLOB_OPEN_TIMED_SHUTTERS_XENO_HIVEMIND, COMSIG_GLOB_OPEN_SHUTTERS_EARLY), .proc/start_maturing)
 	GLOB.xeno_resin_silos += src
 	associated_hive?.handle_silo_death_timer()
 	center_turf = get_step(src, NORTHEAST)
+	START_PROCESSING(SSslowprocess, src)
 	if(!istype(center_turf))
```
TiviPlus, line 54: Yea no use signals

TiviPlus (changes requested):

TiviPlus (approved):

PR #5843, opened 2021-01-30
Changes Marine Respawn Time

joookschad:
## About The Pull Request

This changes marine respawn time from 30 minutes to 45 minutes, simple as.

## Why It's Good For The Game

Marine respawns were made 30 minutes after the addition of headbite, and aim mode has made marine even stronger since then.  Xenos seem to have an issue with killing marines fast enough to reduce their numbers effectively before marines respawn or get new latejoins, especially on highpop
EDIT:I am a monkey and fucked up the changelog
## Changelog
🆑 
balance: marines now take 45 minutes to respawn instead of 30
/🆑

PR #6147, opened 2021-02-22
Revert "Changes Marine Respawn Time"

Naaanii:
Reverts tgstation/TerraGov-Marine-Corps#5843

Pools too strong.

Dying as marine means join Xeno or play on CM.

With how pools work currently, respawning marines by 30 minutes shouldn't be a problem when in a lot of the games, groundside is lost by 13:00.

### DIFF
--- a/code/_globalvars/admin.dm
+++ b/code/_globalvars/admin.dm
@@ -4,7 +4,7 @@ GLOBAL_VAR_INIT(dsay_allowed, TRUE)
 GLOBAL_VAR_INIT(enter_allowed, TRUE)
 GLOBAL_VAR_INIT(respawn_allowed, TRUE)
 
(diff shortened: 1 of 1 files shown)


tgstation-server:

David-Stormwell:
It's about 45-60 minutes when Marines who respawn instantly after the 30 minutes can actually join the fight most of the time anyways.

Losenis:
Prep time, module order, and alamo wait put it at 45 minutes in practice either way. If the idea of the original pr was to make marine deaths have more of an impact, would it not have been better to reduce the time a marine is defibbable rather than straight up forbid them from playing at all for even longer if there's no burrowed/they dont want to play xeno at all? Stopping people from playing altogether should not be part of any decision.

PR #6159, opened 2021-02-22
Trailing return audit

TiviPlus:
## About The Pull Request

Linting would be too much of a pain


code/game/objects/machinery/computer/station_alert.dm
```diff
@@ -102,5 +102,4 @@
 		icon_state = "atmos2"
 	else
 		icon_state = "atmos"
-	..()
```
SplinterGP, line 105: sounds like a woops

code/game/objects/machinery/computer/station_alert.dm
```diff
@@ -102,5 +102,4 @@
...
-	return
+
```
SplinterGP, line 105: looks like a woops

code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -13,8 +13,8 @@
 	if (src == user.wear_mask)
 		return
 	else
-		..()
-	return
+		return ..()
```
SplinterGP, line 16:
```suggestion
	return ..()
```

code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -13,8 +13,8 @@
 	if (src == user.wear_mask)
 		return
 	else
```
SplinterGP, line 15:
```suggestion
```

SplinterGP (changes requested):

code/game/objects/machinery/computer/station_alert.dm
TiviPlus: This is process() so I

code/game/objects/machinery/computer/station_alert.dm
```diff
@@ -103,4 +103,4 @@
 	else
 		icon_state = "atmos"
 	..()
```
SplinterGP, line 105:
```suggestion
	return ..()
```

SplinterGP (changes requested):

code/game/objects/machinery/computer/station_alert.dm
TiviPlus: process shouldnt return values unless its process kill

SplinterGP: should because some parent proc call might call it

SplinterGP (approved):

PR #6168, opened 2021-02-24
Allows IFF Bullets to pass over the TL-102 HSG

hyper2snyper:
## About The Pull Request

Gives the TL-102 HSG the ability to have IFF bullets pass over it the same way it does with Sentries.

## Why It's Good For The Game

From my experience I've found that the main reason the TL-102 dies is because it gets FFd in five seconds. I figure since its function is similar to a Sentry it shouldn't be able to be killed by FF.

## Changelog
:cl:
balance:  Gives the TL-102 HSG the ability to have IFF bullets pass over it the same way it does with Sentries.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -616,14 +616,14 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 	return TRUE
 
 /obj/machinery/marine_turret/projectile_hit(obj/projectile/proj, cardinal_move, uncrossing)
-    for(var/access_tag in proj.projectile_iff)
+	for(var/access_tag in proj.projectile_iff)
 		if(access_tag in iff_signal) //Checks IFF
 			proj.damage += proj.damage*proj.damage_marine_falloff
 			return FALSE
 	return src == proj.original_target
 
 /obj/machinery/standard_hmg/projectile_hit(obj/projectile/proj, cardinal_move, uncrossing)
-    for(var/access_tag in proj.projectile_iff)
+	for(var/access_tag in proj.projectile_iff)
 		if(access_tag in iff_signal) //Checks IFF
 			proj.damage += proj.damage*proj.damage_marine_falloff
 			return FALSE
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -629,6 +629,20 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 
 	return FALSE
 
+/obj/machinery/standard_hmg/projectile_hit(obj/projectile/proj, cardinal_move, uncrossing)
+	if(hmg_verify_iff(proj.projectile_iff))
+		proj.damage += proj.damage*proj.damage_marine_falloff
+		return FALSE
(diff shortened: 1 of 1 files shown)


neohenry:
If you can, add the feature to repair it. As of right now, no one knows how to repair it.

hyper2snyper:
`/obj/item/standard_hmg/welder_act(mob/living/user, obj/item/I)
	if(user.do_actions)
		return FALSE

	var/obj/item/tool/weldingtool/WT = I

	if(!WT.isOn())
		return FALSE

	for(var/obj/effect/xenomorph/acid/A in loc)
		if(A.acid_t == src)
			to_chat(user, "You can't get near that, it's melting!")
			return TRUE

	if(obj_integrity == max_integrity)
		to_chat(user, "<span class='warning'>[src] doesn't need repairs.</span>")
		return TRUE

	if(user.skills.getRating("engineer") < SKILL_ENGINEER_METAL)
		user.visible_message("<span class='notice'>[user] fumbles around figuring out how to repair [src].</span>",
		"<span class='notice'>You fumble around figuring out how to repair [src].</span>")
		var/fumbling_time = 5 SECONDS * ( SKILL_ENGINEER_METAL - user.skills.getRating("engineer") )
		if(!do_after(user, fumbling_time, TRUE, src, BUSY_ICON_BUILD))
			return TRUE

	user.visible_message("<span class='notice'>[user] begins repairing damage to [src].</span>",
	"<span class='notice'>You begin repairing the damage to [src].</span>")
	playsound(loc, 'sound/items/welder2.ogg', 25, TRUE)

	if(!do_after(user, 5 SECONDS, TRUE, src, BUSY_ICON_FRIENDLY))
		return TRUE

	if(!WT.remove_fuel(2, user))
		to_chat(user, "<span class='warning'>Not enough fuel to finish the task.</span>")
		return TRUE

	user.visible_message("<span class='notice'>[user] repairs some damage on [src].</span>",
	"<span class='notice'>You repair [src].</span>")
	repair_damage(120)
	update_icon()
	playsound(loc, 'sound/items/welder2.ogg', 25, TRUE)
	return TRUE
`
Seems its already possible, I could always add something to the description to relay that info.

hyper2snyper:
> If you can, add the feature to repair it. As of right now, no one knows how to repair it.

https://github.com/tgstation/TerraGov-Marine-Corps/pull/6207

babam

code/modules/projectiles/projectile.dm
```diff
@@ -629,6 +629,20 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
...
+		return FALSE
+	return src == proj.original_target
+
+///Checks to see whether IFF matches
+/obj/machinery/standard_hmg/proc/hmg_verify_iff(list/unique_access)
+	for(var/access_tag in unique_access)
+		if(access_tag in iff_signal)
+			return TRUE
+
+	return FALSE
```
TiviPlus, line 644: This could just be inside the previous proc

hyper2snyper: I suppose it could, i just copied this code from the sentry code above it that does the same thing for the sentry. I'll embed it now i guess.

TiviPlus (approved):

PR #6170, opened 2021-02-24
Minor Powerfist Nerf

Cowbin:
## About The Pull Request
Doubles the power cost of using the powerfist as well as decreases the range thrown by one per power level.  

Changes:
Cell charge used per power level: 100 -> 200
Distance thrown per power level:  5 -> 4

Effects of said changes:
On power level three you get the following number of uses out of a cell:
Lasgun: 3 -> 1
Lasgun Highcap: 5 -> 2
Highcap battery: 33 -> 16
On power level two you get the following number of uses out of a cell:
Lasgun: 5 -> 2
Lasgun Highcap: 8 -> 4
Highcap battery: 50 -> 25
On power level one you get the following number of uses out of a cell:
Lasgun: 10 -> 5
Lasgun Highcap: 16 -> 8
Highcap battery: 100 -> 50

## Why It's Good For The Game
Currently the powerfist is a 15 tile throw that can be used at no cooldown and 33 times before changing the cell.  This PR makes it so that it can be used a more reasonable number of times, as well as encourages people to actually use power settings other than max in order to conserve energy.  When you compare the powerfist to something such as the shrike's screech or warrior's fling, it is significantly more powerful than either, with the powerfist being both a longer throw and not having any cooldown.  This seems to make very little sense balance wise.

While this power cost increase effects lasgun cells in a significant manner, lasgun cells are really only a last resort for powerfist users.  Only to be used after all highcap batteries have been used up, and when you do not wish to bother with getting a cell recharger.

## Changelog
:cl:
balance: rebalanced something
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -160,7 +160,7 @@
 	desc = "A dusty sword commonly seen in historical museums. Where you got this is a mystery, for sure. Only a mercenary would be nuts enough to carry one of these. Sharpened to deal massive damage."
 	icon_state = "mercsword"
 	item_state = "machete"
-	force = 39
+	force = 60
 
 
 /obj/item/weapon/claymore/mercsword/captain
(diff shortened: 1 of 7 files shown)


fernandos13:
Distance thrown per power level: 5 -> 4

this means since wasn't in the description how many tiles your now thrown away is

power 1
4tiles   -1 tile from original
power 2
8tiles  -2 tiles from original
power 3
12tiles -3 tiles from original

fernandos13:
i think powerfist is fine as is

and i wouldn't mind the small nerf given to the throwing power honestly but the energy efficiency one i think its too overboard

PR #6197, opened 2021-02-28
Fixes the AI not having access to APC's

lbnesquik:
## About The Pull Request
How did no one fix this before

## Why It's Good For The Game
Just intended functionality
## Changelog
:cl:
fix: Makes it so the AI can access the APC menu.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -11,7 +11,7 @@
 	sight = SEE_TURFS | SEE_MOBS | SEE_OBJS
 	hud_type = /datum/hud/ai
 	buckle_flags = NONE
-	has_unlimited_silicon_privilege = 1
+	has_unlimited_silicon_privilege = TRUE
 
 	var/list/available_networks = list("marinemainship", "marine", "dropship1", "dropship2")
 	var/obj/machinery/camera/current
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -11,6 +11,7 @@
 	sight = SEE_TURFS | SEE_MOBS | SEE_OBJS
 	hud_type = /datum/hud/ai
 	buckle_flags = NONE
+	has_unlimited_silicon_privilege = 1
```
TiviPlus, line 14: This should be TRUE

TiviPlus (changes requested):

code/modules/mob/living/silicon/ai/ai.dm
lbnesquik: TRUE

code/modules/power/apc.dm
```diff
@@ -750,7 +750,7 @@
 
 /obj/machinery/power/apc/ui_act(action, list/params)
 	. = ..()
-	if(. || !can_use(usr, TRUE) || locked)
+	if(. || !can_use(usr, TRUE) || locked & (!usr.has_unlimited_silicon_privilege))
```
TiviPlus, line 753:
```suggestion
	if(. || !can_use(usr, TRUE) || (locked && !usr.has_unlimited_silicon_privilege))
```

lbnesquik: Why is this better ?

BraveMole: & is a binary operation, it's boolean math.
&& is logical operation

& evaluates both sides of the operation.
&& evaluates the left side of the operation, if it's true, it continues and evaluates the right side.

SplinterGP (approved):

PR #6201, opened 2021-02-28
tgui: Typescript and Jest update and a few fixes.

SplinterGP:
port of: https://github.com/tgstation/tgstation/pull/57081

## TypeScript

Adds proper typescript support to tgui.

You don't have to do anything extra to start benefitting from strong static types in normal javascript.

If you need access to all of the TypeScript in your UI, rename the file you're working on to .tsx (or .ts if you're not writing react/inferno components), and you will have access to the whole set of TypeScript features.

## Why is TypeScript good?

Because you, c*ders, will be reprimanded for type violation and bad code automatically and more efficiently.
ESLint

Fixed ESLint integration with VSCode (partially because now we have a Yarn PnP compatible typescript compiler, which eslint needed).

## Jest

You can now write and run unit tests in tgui.

It's quite simple: create a file ending in .test.ts or .spec.ts (usually with the same filename as the file you're testing), and create a test case:

test('something', () => {
  expect('a').toBe('a');
});

To run the tests, type the following into the terminal:

bin/tgui --test

There is an example test in packages/common/react.spec.ts.

You can read more about Jest here: https://jestjs.io/docs/en/getting-started

Note, that there is still no real solution to test UIs for now, even though a lot of the support is here (jest + jsdom). That's going to come later. You can however test simple functions, and are encouraged to do so!

## Fixes
https://github.com/tgstation/tgstation/pull/57154
https://github.com/tgstation/tgstation/pull/57251

comes along due to touching same files and being small fixes, also because building bundle pain.

## Changelog

Not needed


SplinterGP:
also 

image

TiviPlus (approved):

PR #6209, opened 2021-03-01
Siloless collapse takes 20 minutes instead of 10

BraveMole:
## About The Pull Request

Title

## Why It's Good For The Game

Before psydrain, losing all your silos would mean xenos would have lose, and the timer was just here to prevent useless delay from a running shrike.

Now, even if xenos don't have a lot of points when that happen, they still got a chance to get another one if they manage to slaughter a whole bunch of marine and psydrain them.

This also makes losing silos less of a punition that it needs to be

## Changelog
:cl:
balance: Siloless collapse are now 20 minutes instead of 10
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
@@ -1149,4 +1149,4 @@ to_chat will check for valid clients itself already so no need to double check f
 		return
 
 	xeno_message("<span class='xenoannounce'>A sudden tremor ripples through the hive... the last silo was destroyed! The hive will collapse if nothing is done</span>", 3, TRUE)
-	D.siloless_hive_timer = addtimer(CALLBACK(D, /datum/game_mode.proc/siloless_hive_collapse), 10 MINUTES, TIMER_STOPPABLE)
+	D.siloless_hive_timer = addtimer(CALLBACK(D, /datum/game_mode.proc/siloless_hive_collapse), 20 MINUTES, TIMER_STOPPABLE)


tgstation-server:

Walarks:
I think that the 10-minute timer for losing with no silos is alright since their purpose is to close out games where xenos are running too much. 

I feel like a good solution to silos being rushed would be to have the siloless collapse become available past 1h into the round. With this the xenos have a lot of time to decide whether they would like to save for 1 silo in case all get destroyed or not, along with the marines having less incentive to drop in and wrap things up as fast as possible.

Chaznoodles:
Stop trying to shape the rest of the game around pools when they are near-universally viewed as an unengaging and uninteresting addition, so much so that xenos just _aren't placing them any more._ All this does is mean xenos can just make another pool - and another one when that dies, and another one, until marines get bored of playing.

BraveMole:
> Stop trying to shape the rest of the game around pools when they are near-universally viewed as an unengaging and uninteresting addition, so much so that xenos just _aren't placing them any more._ All this does is mean xenos can just make another pool - and another one when that dies, and another one, until marines get bored of playing. This also means, even if you manage to kill the opening silos via the rush you're **forced** to do as marines or inevitably lose over a 2-3 hour round, xenos are still going to come back unless the shrike or queen deliberately suicide and aren't replaced - meaning you're getting that 2-3 hour round anyway.

1 hour, that's more like it chaz

TiviPlus (approved):
Unless YOU are going to propose proper fixes to the mess that is silos and capturing before it we finish fixing xenos until they're finally not a shitty antag

tgstation-server:
Your Fix/Feature pull request delta is currently below zero (-9). Maintainers may close future Feature/Tweak/Balance PRs. Fixing issues or helping to improve the codebase will raise this score.

PR #6220, opened 2021-03-02
Lowers the cost for tesla cells to be 10 instead of 30

2whatever2:
## About The Pull Request
Cut the price from 30 points to 10

## Why It's Good For The Game
Even if tesla cells are rechargeable it's really slow and you can't get more rechargers, and no one uses tesla gun at all :(

## Changelog
:cl:
balance: Tesla cells now cost 10 instead of 30
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -424,7 +424,7 @@ AMMO
 		/obj/item/cell/lasgun/tesla,
 		/obj/item/cell/lasgun/tesla,
 	)
-	cost = 30
+	cost = 15
 
 /datum/supply_packs/ammo/boxslug
 	name = "Slug Ammo Box"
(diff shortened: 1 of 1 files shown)


code/modules/reqs/supplypacks.dm
```diff
@@ -424,7 +424,7 @@ AMMO
 		/obj/item/cell/lasgun/tesla,
 		/obj/item/cell/lasgun/tesla,
 	)
-	cost = 30
+	cost = 15
```
Pariah919, line 427:
```suggestion
	cost = 10
```
just make it ten, the gun isn't very good already, or you should just make it 3 points and only be 1 item like all other power gun things

2whatever2: ok

TiviPlus (approved):

PR #6228, opened 2021-03-03
Xenos upgrade points are kept when evolving 

BraveMole:
## About The Pull Request

When evolving to a higher class, you'll keep the upgrade points you accumulated when aging.

As an exemple, an ancient drone evolving to a hivelord will keep 420 upgrade points, so it will emerge as an elder hivelord, and no longer a young one like previously. You can never get ancient when evolving, only lower upgrades because highier tiers ask for more upgrades points for each level

Regressing also keeps points

Also adjust the amount of points given when recycling a xeno body, because a typical xeno body will have way more upgrade points

## Why It's Good For The Game

Give more incentive to evolve to higher caste, as previously xenos were relunctant to give out their ancient T1/T2 if that meant having to stop playing for some minutes, while waiting to get mature again.

This remove a tedious part of playing xenos, that I frankly don't see the point of. Age should be a reward for not dying as a xenos, not a punition when you decide to evolve.

Also make switching from one caste easier, which allow for more flexibility for xenos

## Changelog
:cl:
balance: Xenos keep their upgrade points when evolving, meaning that an ancient drone will evolve into an elder hivelord, and not a young one. They also keep points when regressing, which then can easily switch from one caste to another
balance: Change how points from recycling a xeno are calculated, overall a T1 will give less and a ancient T3 will give way more
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -555,7 +555,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 
 #define DRONE_SALVAGE_BIOMASS_WINDUP		5 SECONDS //Delay before the target is salvaged
 #define DRONE_SALVAGE_BIOMASS_RANGE			1
-#define DRONE_SALVAGE_BIOMASS_SALVAGE_RATIO	0.2 //Percentile of stored upgrade and evolution salvaged from the target
+#define DRONE_SALVAGE_BIOMASS_SALVAGE_RATIO	0.1 //Percentile of stored upgrade and evolution salvaged from the target
 #define DRONE_SALVAGE_COOLDOWN				60 SECONDS //Can only salvage one corpse per 60 seconds; try not to die *too* quickly.
 #define DRONE_SALVAGE_UPGRADE_FILTER_LIST	list(XENO_UPGRADE_THREE, XENO_UPGRADE_INVALID)
 #define DRONE_SALVAGE_EVOLUTION_FILTER_LIST	list(XENO_TIER_ZERO, XENO_TIER_THREE, XENO_TIER_FOUR)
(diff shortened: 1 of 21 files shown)


SplinterGP:
I'd think losing 1 when devolving/evolvung would be okay, also bcos the aging times across the castes are different so xenos will wait ancient on t1 then go T3, IMO would be nice to just keep per caste tier or when devolving staying with aging to next evolve, since biggest trouble i see is xenos needing to switch castws but losing all age

bobmancer343:
good

code/__DEFINES/mobs.dm
```diff
@@ -555,7 +555,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 
 #define DRONE_SALVAGE_BIOMASS_WINDUP		5 SECONDS //Delay before the target is salvaged
 #define DRONE_SALVAGE_BIOMASS_RANGE			1
-#define DRONE_SALVAGE_BIOMASS_SALVAGE_RATIO	0.2 //Percentile of stored upgrade and evolution salvaged from the target
+#define DRONE_SALVAGE_BIOMASS_SALVAGE_RATIO	0.1 //Percentile of stored upgrade and evolution salvaged from the target
```
TiviPlus, line 558: Not in changelog

code/modules/mob/living/carbon/xenomorph/evolution.dm
```diff
@@ -295,6 +295,10 @@
 			qdel(new_xeno)
 		return
 
+	new_xeno.upgrade_stored = upgrade_stored
+	while(new_xeno.upgrade_possible() && new_xeno.upgrade_stored >= new_xeno.xeno_caste.upgrade_threshold)
+		new_xeno.upgrade_xeno(new_xeno.upgrade_next(), TRUE)
+
```
TiviPlus, line 301: Explain these while loops

TiviPlus (changes requested):

TiviPlus (approved):

PR #6232, opened 2021-03-04
general purpose storage module max weight is small instead of normal

Pariah919:
## About The Pull Request
its supposed to hold generally smaller items and magazines, not up to normal sized.

## Why It's Good For The Game
GPSM shouldn't hold normal sized items.

## Changelog
:cl:
balance: general purpose storage mooudle only holds up to small items, most magazines are unaffected
/:cl:

### DIFF
--- a/code/modules/modular_armor/storage.dm
+++ b/code/modules/modular_armor/storage.dm
@@ -71,7 +71,7 @@
 /** General storage */
 /obj/item/armor_module/storage/general
 	name = "General Purpose Storage module"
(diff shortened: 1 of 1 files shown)


neohenry:
a sad day for jetpack sg

SpaceLoveSs13:
IS was removed to lower storage creep, especially with drums. General storage was left so one can use it for "two drums" only. If you remove that as well, then what would it be used for?

Knuff-les:
this is just reversing QoL when it comes to storage, again.

Swagile:
nooo

t-42 drums 😔

Losenis:
So why would someone take this rather than just take engi/med and put stuff there while putting what'd have gone in here elsewhere?

TiviPlus:
I dont see the use in this

PR #6233, opened 2021-03-04
PoS prep improvements

fernandos13:
## About The Pull Request
this pr aims to make the prep faster and with a more thoughtfull flow of how marines will move about while also giving spec roles more space to fiddle with their equipment

## Why It's Good For The Game
https://gyazo.com/6f7fc4bfb791c2586c27ae8fc9e600e4

heres a foto of it

## Changelog
:cl:
teak: speeds up marines prep in PoS
/:cl:


fernandos13:
map merger is being annoying please help

TiviPlus:
This is not the bottleneck for faster prep but it could be an improvement on the old prep

Noticable issues:
image
There are way too many food vendors (yellow)
There are only two(?) sets of tools though it may be a past issue
The entire red area needs to be redone because
A) VERY cramped
B) Doubled up lights in green
C) it creates a crossflow which slows down marines moving from a to b (blue)

There is one vendor per 2 marines which will get annoying fast(example in black)
Floating light in pink

Looks nice to me otherwise, though you may consider making the preps on the right less long so people at the far end can talk to people at the window, perhaps by removing one or two table/moving the weapons vendors?

TiviPlus:
Also the tweak label does not exist anymore fix your CL please

Chaznoodles:
Reinventing the wheel, this looks like it'll create more flow problems than it solves.

fernandos13:
i'll reopen a NEW map thing. since this branch is byond fuckery now

PR #6241, opened 2021-03-05
disables smartgunner, makes smart machinegun useable by anyone

Pariah919:
## About The Pull Request
smartgunner has been replaced by aim mode effectively and is no longer required for the game to function right, and so it should be disabled and made into a requisition kit.
It costs 40 still, can be decreased or increased at will

## Why It's Good For The Game
Smartgunner doesn't need to exist anymore

## Changelog
:cl:
del: Removed squad smartgunner
balance: smartgunner disabled, smartguns useable by anyone
/:cl:

### DIFF
--- a/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
+++ b/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
@@ -10488,7 +10488,7 @@
 /turf/open/floor/mainship,
 /area/mainship/squads/general)
 "Ix" = (
-/obj/machinery/vending/marineFood,
+/obj/machinery/marine_selector/gear/smartgun,
 /turf/open/floor/mainship,
 /area/mainship/squads/general)
 "Iz" = (
@@ -11107,6 +11107,7 @@
 /area/mainship/shipboard/firing_range)
 "Kl" = (
 /obj/machinery/door/firedoor/mainship,
+/obj/machinery/door/airlock/mainship/marine/general/smart,
 /turf/open/floor/mainship,
 /area/mainship/squads/general)
 "Km" = (
@@ -13897,7 +13898,7 @@
 /turf/open/floor/mainship/mono,
 /area/mainship/hallways/hangar)
 "UD" = (
-/obj/machinery/vending/cola,
+/obj/machinery/marine_selector/clothes/smartgun,
 /turf/open/floor/mainship,
 /area/mainship/squads/general)
 "UE" = (
@@ -48288,10 +48289,10 @@ It
 th
 Wa
 qC
-Ix
+Wa
 It
 th
-Ix
+Wa
 qC
 sK
 FS
--- a/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
+++ b/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
@@ -5391,6 +5391,10 @@
 /obj/effect/landmark/start/job/squadcorpsman,
 /turf/open/floor/mainship,
 /area/mainship/living/cryo_cells)
+"sO" = (
+/obj/effect/landmark/start/job/squadsmartgunner,
+/turf/open/floor/mainship,
+/area/mainship/living/cryo_cells)
 "sP" = (
 /obj/effect/landmark/start/job/squadleader,
 /turf/open/floor/mainship,
@@ -6257,6 +6261,7 @@
 /turf/open/floor/mainship/tcomms,
 /area/mainship/command/telecomms)
 "vz" = (
+/obj/effect/landmark/start/job/squadsmartgunner,
 /obj/machinery/light,
 /turf/open/floor/mainship,
 /area/mainship/living/cryo_cells)
@@ -49573,7 +49578,7 @@ pQ
 YD
 Ne
 KU
-sy
+sO
 sU
 sx
 sx
@@ -51115,7 +51120,7 @@ pQ
 rh
 Ne
 MC
-sy
+sO
 sU
 sx
 sx
--- a/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
+++ b/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
@@ -5391,10 +5391,6 @@
 /obj/effect/landmark/start/job/squadcorpsman,
 /turf/open/floor/mainship,
 /area/mainship/living/cryo_cells)
-"sO" = (
-/obj/effect/landmark/start/job/squadsmartgunner,
-/turf/open/floor/mainship,
-/area/mainship/living/cryo_cells)
 "sP" = (
 /obj/effect/landmark/start/job/squadleader,
 /turf/open/floor/mainship,
@@ -6261,7 +6257,6 @@
 /turf/open/floor/mainship/tcomms,
 /area/mainship/command/telecomms)
 "vz" = (
-/obj/effect/landmark/start/job/squadsmartgunner,
 /obj/machinery/light,
 /turf/open/floor/mainship,
 /area/mainship/living/cryo_cells)
@@ -49578,7 +49573,7 @@ pQ
 YD
 Ne
 KU
-sO
+sy
 sU
 sx
 sx
@@ -51120,7 +51115,7 @@ pQ
 rh
 Ne
 MC
-sO
+sy
 sU
 sx
 sx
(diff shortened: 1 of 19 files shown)


BraveMole:
Can't you like.... Do that but cleanly? Remove SG code, remove sg prep, remove sg points when spawning, etc etc.

SpaceLoveSs13:
Literally why...and even if it's a all use weapon why make it 40 points? Nobody will use a smartgun that costs 40 points and has nothing special like other spec guns. I literally can't wrap my head around how aimmode replaces smartgun.

P.S. Didn't maintainers already close this same topic pr once before?

TiviPlus (changes requested):
I did but spooky says its happening so not like we have much choice

Remove it properly instead of just backspacing

Pariah919:
>Remove it properly instead of just backspacing

spooky told me to just disable it for now instead of deleting it so i'm disabling it only for now

spookydonut:
> 
> 
> > Remove it properly instead of just backspacing
> 
> spooky told me to just disable it for now instead of deleting it so i'm disabling it only for now

this

AyyRobotics:
After hearing the main problems of Smartgunners I'm still unsure that removing them is the right choice. Currently, while they technically have a different role and should have a different mindset than regular marines, a lot of them want to just Unga+. And, while that's honestly fair when you have a smartgun, isn't the way you should be playing it.

Medics have to focus on keeping track of wounded teammates, bodies, and healing/reviving both. Sure, they can Combat Corpsman, but they put their teammates in danger, themselves in danger, and if they die, the entire marine force is weakened.

Engineers have to focus on defenses and maintenance of outposts, miners, or the FOB. Sure, they can unga, and with Point-Defense sentries they can kind of get away with it, but if they try to be like a marine, they put themselves in danger, and the defenses, and by proxy, their teammates, in danger. If they die, the integrity of the FOB/Miner/Outposts are weakened.

Smartgunners _should_ focus on putting themselves in good spots to make use of their suppressing fire. Sure, they can unga, and because they're smartgunners they can get away with it, but like the others, they put themselves and their teammates at risk. Sure, it's not as in depth as the others, but they have something else to be thinking about than just Unga+.

The main reason that SGs are just Unga+ is that they're often not played in their intended role. I've seen countless SGs take point in assaults where they should be behind the main lines. I'm likely not alone in that, hell I've even been that SG and died because of it. If smartgunners just being Unga+ is the main problem, then they should be punished more for trying to be one. Maybe more slowdown, poorer performance in close range? I don't know exactly, but I feel like that would help.

Removing SGs would deprive the marines of a good suppressing fire option. Sure, you have the T-42, but because of aim-mode, it's only good if you're frontlining, and thus, at risk. The GPMG exists, but because of aim-mode and slowdown, it's more suited to defensive use. The Smartgun bridges that gap and gives the marines a suppressing fire option that can stay safe behind the frontliners. Of course, letting it be bought by req will still let it exist, but doing so like it would only result in 3 possible situations, depending on price:
1. Almost nobody buys it and it rarely gets used.
2. People overbuy it and you end up with massive smartgun-deathballs.
3. People buy a few early in the game, weakening req's roundstart pool of income for other things like pads and marine orders.

Hell, I could be wrong about all of this and it gets removed and nothing really changes. The only way to see exactly is to find out, but just in my personal opinion from what I know, I'm wary about this change.

Kuro-ai:
But aim mode already does the SG's job,that's why it's not needed anymore,yeah sure individual guns on aimmode are not as good as a sg but that's because aim mode is "free" and everyone has it,it can't be as good as a SG.
The GPMGs/LMGs/snipers/DMRs of the world already do good enough suppresive fire even if individually it's not as good as SG.
Punishing SGs for ungaing won't make them less of a "unga+" role, it will just make it be closer in performance to the standard guns,and at that point it's even more of a unecessary role.

AyyRobotics:
> But aim mode already does the SG's job,that's why it's not needed anymore,yeah sure individual guns on aimmode are not as good as a sg but that's because aim mode is "free" and everyone has it,it can't be as good as a SG.
> The GPMGs/LMGs/snipers/DMRs of the world already do good enough suppresive fire even if individually it's not as good as SG.
> Punishing SGs for ungaing won't make them less of a "unga+" role, it will just make it be closer in performance to the standard guns,and at that point it's even more of a unecessary role.

I suppose _maybe_ Aim mode can do what it does, but like you said it's hardly as good. And like I said if marines want good suppressive fire without SGs, they need to either put themselves at risk with the T-42 or play defensively with the GPMG. Snipers can do suppressive fire, but unless there are multiple snipers, enough xenos would be enough to overpower it quickly, not to mention they're huge targets for runners and boilers. DMRs don't really have the range or damage to pull off what the sniper does. For a comparison, sure any marine can use medicine or defibs, or repair cades, but only Medics and Engineers do it well (Besides Synth and CMO, etc.) And assuming Unga+ is just "Play like a regular marine but with better equipment", then it would indeed help the problem since it would shift the direction of Smartgunners towards more team-oriented play. CM does this by having SG armor be weaker compared to some of the other armors, making SG's rely on support from their teammates more. I don't know if we could do anything like that with armor, which is why I mentioned the smartgun itself.

Of course if we actually want to make SG more unique we could do what CM did and give it special abilities, which might make it more supportive. Either way my main point is that we might want to tweak around the smartgunner role itself before immediately dropping it

EvilMelonTheSecond:
Maybe roundstart SG kits could be added to vendors to still give them some use. If it is gonna be req only I doubt we will ever see it being used, nobody wants to make builds based on a gun they can't get at roundstart(At 40 points it's practically gone. Even at 20 points I would be surprised to see the req order get approved in a sea of jaeger module orders)

Kuro-ai:
> > But aim mode already does the SG's job,that's why it's not needed anymore,yeah sure individual guns on aimmode are not as good as a sg but that's because aim mode is "free" and everyone has it,it can't be as good as a SG.
> > The GPMGs/LMGs/snipers/DMRs of the world already do good enough suppresive fire even if individually it's not as good as SG.
> > Punishing SGs for ungaing won't make them less of a "unga+" role, it will just make it be closer in performance to the standard guns,and at that point it's even more of a unecessary role.
> 
> I suppose _maybe_ Aim mode can do what it does, but like you said it's hardly as good. And like I said if marines want good suppressive fire without SGs, they need to either put themselves at risk with the T-42 or play defensively with the GPMG. Snipers can do suppressive fire, but unless there are multiple snipers, enough xenos would be enough to overpower it quickly, not to mention they're huge targets for runners and boilers. DMRs don't really have the range or damage to pull off what the sniper does. For a comparison, sure any marine can use medicine or defibs, or repair cades, but only Medics and Engineers do it well (Besides Synth and CMO, etc.) And assuming Unga+ is just "Play like a regular marine but with better equipment", then it would indeed help the problem since it would shift the direction of Smartgunners towards more team-oriented play. CM does this by having SG armor be weaker compared to some of the other armors, making SG's rely on support from their teammates more. I don't know if we could do anything like that with armor, which is why I mentioned the smartgun itself.
> 
> Of course if we actually want to make SG more unique we could do what CM did and give it special abilities, which might make it more supportive. Either way my main point is that we might want to tweak around the smartgunner role itself before immediately dropping it.

But medics and engis do something actually different, SG just shoots better which is neither unique nor deserving of a special role for that (see specs), even if you added a CM like mechanism to the SG it would still just be a unga+ role.
I don't think nerfing the SG would change anything about this, the argument is that SG does nothing aside from just shooting a gun that is slightly better than what PFCs have, aim mode killed the uniqueness of being the only person that was able to shoot trough other marines.

Losenis:
Problem with SG is that the ammo is still effectively req-only despite most guns having aim mode. Vendors only have 2 ammo drums for it, and even then only the munition vendors that are like 2 a map, 3 if you greytide into pilots' bunks. When you play SG you're at the mercy of req.

Tennessee116:
We wiped the panther off the face of the Earth because it was redundant and served only as a pain in the ass. Why is this any different? Besides, half the gun changes here are some "gun tax" removals.

Smudged:
> We wiped the panther off the face of the Earth because it was redundant and served only as a noob-trap. Why is this any different? Besides, half the gun changes here are some "gun tax" removals.

The panther and removing the SG are 2 entirely different problems.

My main concern being, well what was said earlier about SG death balls.

spookydonut:
We are going to remove SG as a playable role for at least a month, with buyable smartgun kits in reqs.
During that time we may adjust the cost of the smartgun kits as balance dictates.
After the month is up we will make the final call on if SG as a distinct squad role will be removed permanently from the codebase.

As a distinct squad role, SG serves little purpose other than a PFC with a different weapon, but requires you to win an RNG roll at round start to be able to play it.  Making it a kit purchased with points from reqs allows potentially any marine to be able to play it without having to be subject to an RNG roll.

The need for smartgunners in general has be made redundant with the introduction of aim mode, the original idea and intention of aim mode was in-fact to render smartgunners redundant as a catalyst to their removal as a role. This idea and intention have succeeded and so this removal is a logical next step.

PR #6265, opened 2021-03-07
[FEEDBACK][DRAFT] Revert "disables smartgunner, makes smart machinegun useable by anyone"

SpaceLoveSs13:
Reverts tgstation/TerraGov-Marine-Corps#6241

**THIS PR IS MAINLY FOR FEEDBACKS ABOUT SGnr REMOVAL TO AVOID CLOGGING feedback CHANNEL IN DISCORD. AND IS A DRAFT**

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -14,6 +14,7 @@
 		/datum/job/terragov/squad/standard = -1,
 		/datum/job/terragov/squad/engineer = 8,
 		/datum/job/terragov/squad/corpsman = 8,
+		/datum/job/terragov/squad/smartgunner = 1,
 		/datum/job/terragov/squad/leader = 1,
 		/datum/job/terragov/medical/professor = 1,
 		/datum/job/terragov/silicon/synthetic = 1,
--- a/code/datums/gamemodes/distress.dm
+++ b/code/datums/gamemodes/distress.dm
@@ -26,6 +26,7 @@
 		/datum/job/terragov/silicon/ai = 1,
 		/datum/job/terragov/squad/engineer = 8,
 		/datum/job/terragov/squad/corpsman = 8,
+		/datum/job/terragov/squad/smartgunner = 1,
 		/datum/job/terragov/squad/leader = 1,
 		/datum/job/terragov/squad/standard = -1,
 		/datum/job/xenomorph = 2,
--- a/code/datums/gamemodes/extended.dm
+++ b/code/datums/gamemodes/extended.dm
@@ -20,6 +20,7 @@
 		/datum/job/terragov/silicon/ai = 1,
 		/datum/job/terragov/squad/engineer = 8,
 		/datum/job/terragov/squad/corpsman = 8,
+		/datum/job/terragov/squad/smartgunner = 4,
 		/datum/job/terragov/squad/leader = 4,
 		/datum/job/terragov/squad/standard = -1
 	)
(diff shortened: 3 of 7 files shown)


SpaceLoveSs13:
My main concern with SGnr removal is that it will turn into either 1. People buying too many SGs and cadehug miners in order to do that and get SG nerfed. 2. Nobody will use it and it will be considered, "Nobody uses it", "Time to remove it".

Walarks:
> 2. Nobody will use it and it will be considered, "Nobody uses it", "Time to remove it".

We have a lot of underused weapons. I doubt SG will completely get removed since it's more unique/powerful than some other underused weapons in requisitions.

EvilMelonTheSecond:
For 40 points you can also buy the IFF NVG sniper which has a much wider niche and still isn't bought roundstart. SG is reliable suppression on the frontlines that allow people to push without the crippling slow of aim mode which is useful but doesn't offer much else. Considering the current meta of roundstart req purchases being all jaeger mods and logistics stuff I assume we won't be seeing the SG unless marines are stomping xeno and got a few uncontested miners which is practically removing it.

I don't know if the meta will adapt with aim mode of MGs or completely shift to some other gun composition but I really don't like losing a safe and reliable gun. At least add some for free at vendors maybe?

JackTheJackhammer:
The reason why the SGnr role removal is so disliked is because people will either buy 20 of em in the span of 60 minutes and win the fight with superior fire power or it will be so expensive that no one can get one. Limiting the SG gun to 3-4 is a straight nerf as it now has a price. SGnr was not broke. Aim does not replace it. Sure, the volume of fire of a group of say 10 T-12 or T-18 players can equate to the volume of fire of a SGnr. However, the role of the SGnr is not to be a bullet machine. It is to counter close range xenos that rely on both CQC and FF for their damage (Hunters, Runners, and maybe ravs). The straight removal of the SGnr is a straight nerf to marines in a place where it was not needed. I am sure everyone has their own views on this role change but the fact that so many people oppose this change (xeno or marine) should show how we should act. SGnr was not broke, why change it?

Swagile:
Due to the lack of combat engis at the front (due to their horrendously unloved status aka lack of mats to make ANY sort of forward outpost except FOB round start due to FOB getting free mats), the SG was THE defender of unga balls. 

Xenos win because of attrition as while premedding is good, somebody has to defib ungas that go too far, or use specialized chems to counter permanent status affects such as internal damage (bull, rav, hunter, slug / flech FF), IB, etc. But corpsmen can't work when they constantly keep getting slapped by good runner / hunter mains, which ends up in the destruction of a unga ball. SG's are there to counter this; good damage and sunder, IFF, and large ammo count are all that is needed to ensure that a no cade push succeeds; the only equivalent of this is 10 + GPMG marines in aim mode.

This also doesn't stop the fact that GPMG's and TL bolts still require some time to aim mode after being stunned, while SG doesn't, and every second counts in a xeno vs marine encounter; its why Queen screech timed well can do decent damage to marines that overly rely on aim mode, because you have to both weld the gun AND aimmode, and by then you've been slashed multiple times. And then stunned again by a competent xeno like a runner pounce and the cycle continues until you are dead.

SG's don't have these demerits and in the hands of a good player are incredibly dangerous to face unless you have a plan as a xeno.

The only way I can see the removal of the SG job is if combat engineer was buffed to compensate, as the SG is basically a mobile 360 base turret in the right hands, and is one of the only major defenses against a coordinated xeno team during a push.

spookydonut:
As stated sg will be disabled for a month, it's been a few days so any feedback you give now is pointless, except for feedback on the price of the kit.

Wait a few weeks and I'll start collecting feedback on the removal.

PR #6277, opened 2021-03-08
Psy drain cause clone damage to victim

BraveMole:
## About The Pull Request

Psy draining a body will give it 10 clone damage, that can only be healed in cryotube. 

## Why It's Good For The Game

So dying and dying again is a bit painfull for the marine, rather than being a 5 minutes wait like currently

## Changelog
:cl:
balance: psy drain gives 10 clone damage
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -1213,6 +1213,8 @@
 	var/psy_points_reward = 60
 	///How much larva points it gives (8 points for one larva in distress)
 	var/larva_point_reward = 1
+	///How much clone damage it does 
+	var/clone_damage = 10
 
 /datum/action/xeno_action/activable/psydrain/can_use_ability(atom/A, silent = FALSE, override_flags)
 	. = ..() //do after checking the below stuff
@@ -1271,6 +1273,7 @@
 	victim.do_jitter_animation(2)
 
 	ADD_TRAIT(victim, TRAIT_PSY_DRAINED, TRAIT_PSY_DRAINED)
+	victim.apply_damage(clone_damage, CLONE)
 
 	SSpoints.xeno_points_by_hive[X.hivenumber] += psy_points_reward
 	var/datum/job/xeno_job = SSjob.GetJobType(/datum/job/xenomorph)


Guidesu:
even if i am an marine main
It is actual better this way.

Swagile:
Maybe if there is a way besides evac that allows you to heal clone damage; a special corpsmen chem with limited amounts in medbay?

Otherwise this is basically a quasi delimb situation, as clone damage really fucks you up.

BraveMole:
> Maybe if there is a way besides evac that allows you to heal clone damage; a special corpsmen chem with limited amounts in medbay?
> 
> Otherwise this is basically a quasi delimb situation, as clone damage really fucks you up.

I mean the marine was fucking dead. 10 clone damage is OK

Swagile:
> 
> 
> > Maybe if there is a way besides evac that allows you to heal clone damage; a special corpsmen chem with limited amounts in medbay?
> > Otherwise this is basically a quasi delimb situation, as clone damage really fucks you up.
> 
> I mean the marine was fucking dead. 10 clone damage is OK

yeah and 10 clone damage is constant pain which causes slow down, forcing an evac.

problem is even if you have MO's who know what they are doing (rare), someone kind enough to cryo you (sometimes), you also run into the problem of getting back planetside, which is either pod in (limited and gets even more limited if marines keep getting drained then defibbed), or use DS (which is incredibly slow even at orbit 3, and going orbit 1 its okish)

its why i call it a "delimb" situation; your forced to evac because your gimped and probably will take 5-10 mins to get back down; thats on top of the 5 minutes it took to get defibbed and patched up enough so you can actually move

so your basically out of the round on half of the rounds

not to mention cryo will run out of chems and unless MO's are on top of it you get to a point where you can't get healed at all (this is possible if marines have good corpsmen and ungas who push to recover corpses)

tl;dr this exposes some of the flaws of evac and is essentially taking someone out of a round near permanently

problem is i don't see what else you could do to make psy drain punish marines, since all other damage types can be healed by a corpsmen

Swagile:
also this assumes that marines are playing high pop

in med to low pop, you don't always get corpsmen or MO's, or vice versa; man power is focused in one role so if you play low pop and medical players (wisely) choose to play corpsmen (since they are more important), and you get psy drained, your kinda screwed

ilikesaying:
Cryoxadone isn't that hard to make, but the main problem might be the lack of people who know why they have cell damage and how they can heal it, as well as the lack of people using the cryo cells.

TiviPlus:
Against this, making very hard to heal damage for when you die isnt that nice for how simple it is to drain

PR #6279, opened 2021-03-08
Removes soul

TiviPlus:
You will not be missed

removes list2text


David-Stormwell:
my heart will live on for youuuuuuu

PR #6292, opened 2021-03-09
Transit shutters has had its ID changed

mrhugo13:
## About The Pull Request
dreammaker_2021-03-22_13-43-35
simply adds the ID part to this

## Why It's Good For The Game
if any of the buttons on the place it landed has it's ID set to null then it would've triggered the transit shutters because the transit shutters were also null, **this is bad.**

~~kindadread owes me a crash token or two for this favor~~
## Changelog
:cl: Vondiech
fix: Transit shutters can no longer be triggered by any buttons on shipside or groundside that has its ID set to null
/:cl:

### DIFF
--- a/_maps/map_files/Ice_Colony_v2/Ice_Colony_v2.dmm
+++ b/_maps/map_files/Ice_Colony_v2/Ice_Colony_v2.dmm
@@ -2442,6 +2442,13 @@
 /obj/effect/turf_overlay/icerock,
 /turf/closed/ice_rock/corners,
 /area/ice_colony/exterior/surface/cliff)
+"ajQ" = (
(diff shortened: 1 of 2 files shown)


SplinterGP:
please change the alamo shutters ID, this is possible to happen again

SplinterGP (approved):

SplinterGP (changes requested):
buttons are fluff and there are other buttons around map which also have null, need to change alamo shutter ID

TiviPlus:
@mrhugo13 ?

mrhugo13:
There, my bad for procrastinating on this for so long.

TiviPlus (approved):

PR #6300, opened 2021-03-10
Gun rattle update

XSlayer300:
## Changelog
:cl:
add: Added exclusive gun rattle sounds to the GPMG, DMR, battle rifle and sniper instead of sound-pitching them.
del: Revolvers do not use rattling sounds as they do not make any mechanical sense, also for the case of some weapons.
/:cl:


SilencedMP5A5:
i like the revolver rattle, ngl :)

TiviPlus (approved):

Issue #6266, opened 2021-03-07
Warriors "Fling" Button has a Permanent "Cooldown" Clock over it if GrappleToss/Fling is used at any point in the warriors life and stays so until it evolves, or devolves.


usnpeepoo:
## Testmerges:
N/A
## Reproduction:

Go warrior, Use Fling/Grapple toss on anyone human or xeno, then wait for the Fling "clock" to disappear. It doesn't.

PR #6301, opened 2021-03-10
Cooldown Fix for Fling and Grapple Toss

Surrealaser:
## About The Pull Request

Fixes the shared cooldown for Grapple Toss and Fling so it works properly.

## Why It's Good For The Game

Fixes the shared cooldown for Grapple Toss and Fling so it works properly.

Fixes: https://github.com/tgstation/TerraGov-Marine-Corps/issues/6266

## Changelog
:cl:
fix: Shared cooldown for Grapple Toss and Fling fixed so that using one doesn't cause the other to have an indefinite cooldown.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
@@ -132,14 +132,6 @@
 	keybind_signal = COMSIG_XENOABILITY_FLING
 	target_flags = XABB_MOB_TARGET
 
-/datum/action/xeno_action/activable/fling/give_action(mob/living/L)
-	. = ..()
-	RegisterSignal(owner, COMSIG_WARRIOR_USED_GRAPPLE_TOSS, .proc/add_cooldown) //Shared cooldown with Grapple Toss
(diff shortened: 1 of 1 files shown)


TiviPlus:
This appears to have very much been intentional justify your changes

SplinterGP:
i remember this was intentional so you could use fling or grapple toss for more range and more controlled ?

Surrealaser:
> This appears to have very much been intentional justify your changes

No.

The way it was supposed to work is that the cooldown is shared, so you can't combo them both in a single go, but that you could use both, just not in tandem; not that you pick one and then you're stuck with it forever.

I thought this is the way it was working; I was shocked to find out that this was not the case.

When I tested prior to merging it worked as intended.

SkeletalElite:
so does this PR get rid of the shared cooldown? Or was there a bug with the shared cooldown that it fixes?

Surrealaser:
> so does this PR get rid of the shared cooldown? Or was there a bug with the shared cooldown that it fixes?

The shared cooldown is an actual shared cooldown as opposed to usage of one perma disabling the other.

TiviPlus (approved):

PR #6306, opened 2021-03-10
Cloak grenades only stealth marines, supplies cut in half

Walarks:
## About The Pull Request
Cloak grenades now only stealth marines. Cut cloak grenade supply in half to account for this.

## Why It's Good For The Game
Makes cloak grenades easier to play around and can help as an option that encourages pushing by allowing marines to advance while being harder to be focused down or ganged up on by xenos due to being stealthed. Xenos can still see the smoke, so marines using these aren't completely unpredictable either. 

Having xenos be stealthed like currently is much better for xenos than marines being stealthed since xenos can generally tank a few stray shots and most of their skills are projectiles, have a good area of effect or bumpslash, meaning they don't need to specifically see their target to be effective or worry about FF.

## Changelog
:cl:
balance: cloak grenades cloak only marines instead of marines and xenos, decreased alpha by ~70%
balance: cut cloak grenade supplies in half
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -232,10 +232,19 @@
 /mob/living/effect_smoke(obj/effect/particle_effect/smoke/S)
 	. = ..()
 	if(!.)
+		if(CHECK_BITFIELD(S.smoke_traits, SMOKE_CAMO))
+			smokecloak_off()
 		return
+	if(smoke_delay)
+		return FALSE
 	if(CHECK_BITFIELD(S.smoke_traits, SMOKE_XENO) && (stat == DEAD || isnestedhost(src)))
 		return FALSE
+	smoke_delay = TRUE
+	addtimer(CALLBACK(src, .proc/remove_smoke_delay), 10)
 	smoke_contact(S)
+
+/mob/living/proc/remove_smoke_delay()
+	smoke_delay = FALSE
 
 /mob/living/proc/smoke_contact(obj/effect/particle_effect/smoke/S)
 	var/protection = max(1 - get_permeability_protection() * S.bio_protection)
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/living_defense.dm
```diff
@@ -232,10 +232,10 @@
 /mob/living/effect_smoke(obj/effect/particle_effect/smoke/S)
 	. = ..()
 	if(!.)
-		if(CHECK_BITFIELD(S.smoke_traits, SMOKE_CAMO))
+		if(CHECK_BITFIELD(S.smoke_traits, SMOKE_CAMO) && ishuman(src))
```
BraveMole, line 235: istype(src) == bad. Use flags or even better, override the proc

BraveMole (changes requested):

code/modules/mob/living/living_defense.dm
```diff
@@ -232,10 +232,10 @@
...
 			smokecloak_off()
 		return
-	if(CHECK_BITFIELD(S.smoke_traits, SMOKE_CAMO))
+	if(CHECK_BITFIELD(S.smoke_traits, SMOKE_CAMO) && ishuman(src))
```
TiviPlus, line 238: Balance wise turned out to be fine, fix this

code/modules/mob/living/living_defense.dm
Walarks: moved to human/effect_smoke

code/modules/mob/living/living_defense.dm
Walarks: moved to human/effect_smoke

TiviPlus (approved):

PR #6308, opened 2021-03-10
2.5 supply points are given everytime a marine spawn

BraveMole:
## About The Pull Request

Ttile. Supply points start at 0. Doesn't work on respawn

## Why It's Good For The Game

Takes into account the serv pop. Overall, give more supply points in high pop, but since SG was removed i see this as ok balance wise

## Changelog
:cl:
balance: Every marine spawning for the first time will provide req 2.5 points
/:cl:


SpaceLoveSs13:
Does this give point when marines respawn too? or just lobby spawns?

BraveMole:
> Does this give point when marines respawn too? or just lobby spawns?

Just lobby spawn

Swagile:
does this work for latespawns?

for example i don't ready up sometimes cause i want to see if someone is taking a critical role or not (for example, i do this during low pop so that i can take FC or SL if no one does)

BraveMole:
> does this work for latespawns?
> 
> for example i don't ready up sometimes cause i want to see if someone is taking a critical role or not (for example, i do this during low pop so that i can take FC or SL if no one does)

yes

SplinterGP:
Giving this a TM, might be too much, maybe not, who knows

code/__DEFINES/mode.dm
```diff
@@ -112,6 +112,8 @@
 #define COOLDOWN_COMM_MESSAGE 	1 MINUTES
 #define COOLDOWN_COMM_CENTRAL 	30 SECONDS
 
+#define SUPPLY_POINT_MARINE_SPAWN 5
```
SplinterGP, line 115:
```suggestion
#define SUPPLY_POINT_MARINE_SPAWN 5
```
Actually i believe this should be gamemode dependant, just add a var to gamemode type for this

SplinterGP (changes requested):

code/__DEFINES/mode.dm
BraveMole: The only gamemode that uses req point is distress, no crash...?

SplinterGP (approved):
code ok

code/controllers/subsystem/points.dm
```diff
@@ -13,7 +13,7 @@ SUBSYSTEM_DEF(points)
 	wait = 10 SECONDS
 
 	var/dropship_points = 0
-	var/supply_points = 120
+	var/supply_points
```
SplinterGP, line 16: actually
```suggestion
	var/supply_points = 0
```

SplinterGP (changes requested):

SplinterGP (approved):

SplinterGP:
Apparently this is alright @TiviPlus @BraveMole ? i'll merge tomorrow if is

BraveMole:
Yeah i think it's fine, especially since free sg kits are back

PR #6319, opened 2021-03-11
Silo price increase

BraveMole:
## About The Pull Request

Silo price from 900 to 1200. 

## Why It's Good For The Game

Cocoon is an excellent way of having psy points, so the price of everything should increase at least a bit. Will ake turrets more expensive latter

## Changelog
:cl:
balance: Silos are now 1200 points to account for cocoon
/:cl:


TiviPlus:
Make turrets more expensive first

PR #6331, opened 2021-03-13
You can see how much ammo the condor minigun has now. Also typo fix.

lbnesquik:
## About The Pull Request
Just a missing examine. Where are my tendies. 

## Changelog
:cl:
fix: You can now see the amount of ammo the condor minigun has by just looking.
/:cl:

### DIFF
--- a/code/game/objects/structures/cas_plane_parts.dm
+++ b/code/game/objects/structures/cas_plane_parts.dm
@@ -145,6 +145,13 @@
 	///ref to the static weapon
 	var/obj/structure/dropship_equipment/weapon/static_weapon
 
+/obj/structure/caspart/minigun/examine(mob/user)
+	. = ..()
+	if(static_weapon.ammo_equipped)
+		static_weapon.ammo_equipped.show_loaded_desc(user)
+	else
+		to_chat(user, "It's empty.")
+
 /obj/structure/caspart/minigun/connect_to_shuttle(obj/docking_port/mobile/port, obj/docking_port/stationary/dock, idnum, override=FALSE)
 	if(!istype(port, /obj/docking_port/mobile/marine_dropship/casplane))
 		return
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/cas_plane_parts.dm
```diff
@@ -145,6 +145,13 @@
 	///ref to the static weapon
 	var/obj/structure/dropship_equipment/weapon/static_weapon
 
+/obj/structure/caspart/minigun/examine(mob/user)
+	. = ..()
+	if(static_weapon.ammo_equipped)
+		static_weapon.ammo_equipped.show_loaded_desc(user)
+	else
+		to_chat(user, "It's empty.")
```
BraveMole, line 153: add a span class, like span = notice

code/game/objects/structures/cas_plane_parts.dm
```diff
@@ -145,6 +145,13 @@
 	///ref to the static weapon
 	var/obj/structure/dropship_equipment/weapon/static_weapon
 
+/obj/structure/caspart/minigun/examine(mob/user)
+	. = ..()
+	if(static_weapon.ammo_equipped)
+		static_weapon.ammo_equipped.show_loaded_desc(user)
```
BraveMole, line 151: use the ?. operator, you don't need the if

BraveMole (changes requested):

code/game/objects/structures/cas_plane_parts.dm
BraveMole, line 151: make an early return so you can remove the else, or even better, inline this

code/game/objects/structures/cas_plane_parts.dm
lbnesquik: Other weapons don't have a notice

code/game/objects/structures/cas_plane_parts.dm
lbnesquik: Not sure what this means, sorry.

TiviPlus: usually not worth the effort to change, but bravemole would be correct here
```
//They mean this
if(!x)
   return
if(!y)
  return
stuff()
//as opposed to 
if(x)
   if(y)
       stuff()

```

TiviPlus (approved):

PR #5639, opened 2021-01-03
Powerfist balancing

Wayland-Smithy:
## About The Pull Request

TLDR It's a nerf 🦀

- Way worse power consumption: 3 max setting fists per ~10000 cell capacity, or 27 low setting fists
- Worse knockback overall, and if you fist a `MOB_SIZE_BIG` Xeno you go flying instead
- Hitting a `MOB_SIZE_BIG` also does double damage though

For the record since some marine mains seem to think a 15 tile mob throw with basically no cooldown is balanced here are the most comparable xeno abilities:
- Shrike - Unrelenting Force: 6 tile throw, 50 seconds cooldown
- Crusher - Crest Toss: 4 tile throw (5 if backwards), 18 seconds cooldown

## Why It's Good For The Game

Balances a new feature. Once toned down other aspects can be looked at for the Powerfist, maybe it could be used like spring boots to launch off walls and cades as well as xenos.

I think the need for changing justifies itself. You can still carry like a dozen of the things and also plenty of backup power.

## Changelog
:cl:
balance: Increased Powerfist power cell depletion significantly.
balance: Increased Powerfist damage vs big Xenos but reversed the knockback, and reduced the fling distance in general.
spellcheck: Fixed a Powerfist hit message typo.
/:cl:

PR #6343, opened 2021-03-13
Powerfist balancing

Walarks:
## About The Pull Request
Revives #5639 
- Power fist can now only use lasgun power cells
- Energy consumption is 50*power level. This is 20 uses at power 1, 10 at power 2 and 6 at power 3.
- Hitting a MOB_SIZE_BIG does double damage, knocks you back and staggers the xeno similarly to buckshot instead of knocking them back.
- Takes 6 hits at charge 3 to crit an elder crusher and 5 hits to crit a young runner. For reference, harvester takes 4 hits to down a runner and doesn't require batteries, but much more to down a high-armored xeno because it doesn't ignore armor, while powerfist does.

## Why It's Good For The Game
It makes powerfist's crowd control more restrained as getting hit by it while being a slow xeno can be a death sentence, while it doesn't matter much for fast xenos. This also opens room for the powerfist to have its damage improved so that it is still good for pushing xenos, but can also be used for damage.

## Changelog
:cl:
balance: powerfist now has less uses, but deals more damage. If you hit a big xeno, you get pushed back instead and deal additional damage
/:cl:


InterroLouis:
The power use increase was excessive before and it's excessive now.

Naaanii:
Based. Maybe lower the max setting changes a bit.

Walarks:
> The power use increase was excessive before and it's excessive now.

> Based. Maybe lower the max setting changes a bit.

Changed stuff.

Swagile:
maybe if it didn't require a screwdriver to remove battery so you can switch on the go this would be more viable

otherwise this breaks a lot of builds, because you needed the power of high cap to be able to punch more than a few times

plus big mob xenos knocking you back when you hit them is really dumb; remove that and keep the stagger instead. stagger without the knockback of powerfist is already a big enough nerf to powerfist, no need to literally punish people for playing the game by punching them AWAY from xenos (when its already hard to get close to xenos in the first place)

Walarks:
> maybe if it didn't require a screwdriver to remove battery so you can switch on the go this would be more viable
> 
> otherwise this breaks a lot of builds, because you needed the power of high cap to be able to punch more than a few times
> 
> plus big mob xenos knocking you back when you hit them is really dumb; remove that and keep the stagger instead. stagger without the knockback of powerfist is already a big enough nerf to powerfist, no need to literally punish people for playing the game by punching them AWAY from xenos (when its already hard to get close to xenos in the first place)

I think we can test both since you being pushed away can also be a positive since you stun the xeno for a bit and are brought to safety. If you don't get pushed away, the repeated stunning could also be overwhelming.

InterroLouis:
I mean, it doesn't actually stun. It applies a single stack each of weaken and stagger, plus a small slowdown.

SilencedMP5A5:
power fist is fine 

just surround the user when they overextend and automatically win

fernandos13:
THE GOOD THING ABOUT THE POWER FIST IS THE USE.. people take machete and vali for damage the POWERFIST thing IS the utility

SilencedMP5A5:
ngl i take power fist because i'm bad at clicking on xenos and it's satisfying to chain multiple hits and send them careening off into the sunset

PR #6353, opened 2021-03-14
Harvester effects small rework

Walarks:
## About The Pull Request
Changes harvester effect slightly and moves them from longer effects to shorter ones with more immediate impact.

Changed harvester effects cost from 10u to 5u. 
- Kelotane effect now ignites all mobs in the cone, dealing additional fire damage, without leaving fire tiles. 
- Tramadol effect deals 60% bonus damage that ignores armor, but the slow lasts 1s instead of 2s. 
- Bicaridine effect heals 12.5 brute instead of 10 brute. If you target yourself, it instantly (without the usual channeling delay) heals 6 brute and restores 30 stamina.
> For reference, when metabolized 10u of bicaridine heals 50 brute and if the volume is above 20, 10u heals 75 brute, while also giving 60% of tramadol's pain reduction.

## Why It's Good For The Game
First, I think these changes should be tested before merge.

The changes allow for harvester to deal more damage without giving it a flat force increase and help it warrant the danger of having to get closer to the xenos to use it. It decreases the duration of the kelotane and tramadol effects in exchange for higher mobility or damage when applying them.

The tramadol effect now slows for less, but with the added armor-penetrating damage, it allows to deal damage to more armored xenos and xenos have higher melee armor than bullet armor, while not being too strong against xenos with lower armor or health.
The kelotane effect changes allow for followup instead of blocking the engage route or allowing the xeno to shuffle the marine into the fire.
Harvester's bicaridine heal offers quicker healing for highly decreased efficiency. At 10 healing for 10u it poses too much of a waste of resources to use, while at 25 it is still inefficient, but could feel more enticing to use in some scenarios. It can also be used in-combat for less than half healing, but with stamina regeneration for increased mobility.

As the effects last less, I think it would be fair to have lower reagent costs.

## Changelog
:cl:
balance: changed harvester effects cost from 10u to 5u. Kelotane effect now ignites all mobs in the cone, dealing additional fire damage, without leaving fire tiles. Tramadol effect deals 60% bonus damage that ignores armor, but the slow lasts 1s instead of 2s. Bicaridine effect heals 12.5 brute instead of 10 brute. If you target yourself, it instantly restores 6 brute and 30 stamina.
qol: failing to fill the harvester no longer makes you lose reagents from the liquid container, this wasn't an intended thing
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -111,7 +111,7 @@
 		to_chat(user, "<span class='rose'>The blade is powered with [loaded_reagent.name]. You can release the effect by stabbing a creature.</span>")
 		return FALSE
 
-	if(beaker.reagents.total_volume < 10)
+	if(beaker.reagents.total_volume < 5)
 		to_chat(user, "<span class='rose'>You don't have enough substance.</span>")
 		return FALSE
 
@@ -124,7 +124,7 @@
 		return FALSE
 
 	loaded_reagent = beaker.reagents.reagent_list[1]
-	beaker.reagents.remove_any(10)
+	beaker.reagents.remove_any(5)
 	return TRUE
 
 /obj/item/weapon/claymore/harvester/attack(mob/living/M, mob/living/user)
@@ -136,22 +136,25 @@
 
 	switch(loaded_reagent.type)
 		if(/datum/reagent/medicine/tramadol)
-			M.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 2 SECONDS)
+			M.apply_damage(force*0.6, BRUTE, user.zone_selected)
+			M.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 1 SECONDS)
 
 		if(/datum/reagent/medicine/kelotane)
-			var/turf/target = get_turf(M)
-			target.ignite()
 			var/list/cone_turfs = generate_cone(user, 2, 1, 91, Get_Angle(user, M.loc))
 			for(var/X in cone_turfs)
 				var/turf/T = X
-				T.ignite()
+				for(var/mob/living/C in T)
+					C.flamer_fire_act(10)
+					C.apply_damage(20, BURN, user.zone_selected, C.get_soft_armor("fire", user.zone_selected))
+					if(C.IgniteMob())
+						C.visible_message("<span class='danger'>[C] bursts into flames!</span>","[isxeno(C)?"<span class='xenodanger'>":"<span class='highdanger'>"]You burst into flames!</span>")
 
 		if(/datum/reagent/medicine/bicaridine)
 			to_chat(user, "<span class='rose'>You prepare to stab [M]!</span>")
 			if(!do_after(user, 2 SECONDS, TRUE, M, BUSY_ICON_DANGER))
 				return FALSE
 			new /obj/effect/temp_visual/telekinesis(get_turf(M))
-			M.heal_overall_damage(25, 0, TRUE)
+			M.heal_overall_damage(12.5, 0, TRUE)
 			loaded_reagent = null
 			return FALSE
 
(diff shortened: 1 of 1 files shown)


SkeletalElite:
Should make bicardine "life steal" or something instead so when you bop a xeno with it you get a 10 or hp heal.

SplinterGP:
> 
> 
> Should make bicardine "life steal" or something instead so when you bop a xeno with it you get a 10 or so hp heal. Could make it like 5 or 7.5 as well but make it double dip burn/brute so you get extra healing if you have both (meaning you getting ganged on by some xenos)

this one sounds good honestly

TiviPlus:
Agreed

Walarks:
> > Should make bicardine "life steal" or something instead so when you bop a xeno with it you get a 10 or so hp heal. Could make it like 5 or 7.5 as well but make it double dip burn/brute so you get extra healing if you have both (meaning you getting ganged on by some xenos)
> 
> this one sounds good honestly

> Agreed

Changed self-targeting to instantly restore 6 brute and 30 stamina, while targeting others remains the same. I don't think the attack cooldown gets triggered when you use it on yourself as well.

code/game/objects/items/weapons/blades.dm
```diff
@@ -136,22 +136,29 @@
...
+			if(isxeno(M))
+				return ..()
+			to_chat(user, "<span class='rose'>You prepare to stab <b>[M != user ? "[M]" : "yourself"]</b>!</span>")
 			new /obj/effect/temp_visual/telekinesis(get_turf(M))
-			M.heal_overall_damage(10, 0, TRUE)
+			if((M != user) && do_after(user, 2 SECONDS, TRUE, M, BUSY_ICON_DANGER))
+				M.heal_overall_damage(12.5, 0, TRUE)
+			else
+				M.adjustStaminaLoss(-30)
+				M.heal_overall_damage(6, 0, TRUE)
```
BraveMole, line 161: 2 sec for 12.5? Come on slot, why would you use this rather than just giving the bicaridine directly

code/game/objects/items/weapons/blades.dm
```diff
@@ -136,22 +136,29 @@
...
-			target.ignite()
 			var/list/cone_turfs = generate_cone(user, 2, 1, 91, Get_Angle(user, M.loc))
 			for(var/X in cone_turfs)
 				var/turf/T = X
-				T.ignite()
+				for(var/mob/living/C in T)
+					C.flamer_fire_act(10)
+					C.apply_damage(20, BURN, user.zone_selected, C.get_soft_armor("fire", user.zone_selected))
+					if(C.IgniteMob())
+						C.visible_message("<span class='danger'>[C] bursts into flames!</span>","[isxeno(C)?"<span class='xenodanger'>":"<span class='highdanger'>"]You burst into flames!</span>")
```
BraveMole, line 150: Don't remember how surt works, does this bypass the module?

BraveMole (changes requested):

code/game/objects/items/weapons/blades.dm
SkeletalElite: Surt just gives 200 hard fire armor, so I don't think it will

code/game/objects/items/weapons/blades.dm
Walarks: It costs 5u, is usable on dead people for revives and heals about 6.25 per tick versus bicaridine's 1/1.5 per tick. I could make it better later if need be, but I feel like if it is better it effectively makes marines heal much faster than xenos.

code/game/objects/items/weapons/blades.dm
Walarks: It shouldn't I think.

TiviPlus, line 150: Why are you using C for a living mob use better var names

TiviPlus (changes requested):

code/game/objects/items/weapons/blades.dm
Walarks: changed

TiviPlus (approved):

PR #6357, opened 2021-03-15
Removes all of CAS guns/ammo at roundstart, give them points instead.

lbnesquik:
## About The Pull Request
This PR deletes all of the CAS ammo & guns from Sulaco & POS, replacing them with a direct point cost equivalent at roundstart.

## Why It's Good For The Game
Currently, PO's are a bit forced to use what you get at roundstart given the few options to get more points.
Either req spends on you, you wait a while or you move to high orbit where you can't hit anything, but you have more points.

This means the best bang for your buck is to take the guns you have at the start and only buy ammo for them at the start.
Smart, but also not very engaging.

This aims to change that, by giving more freedom for things PO's can build on. 
You want four napalm rockets and nothing else? Sure. Only miniguns, why not. 
You'll just have to spend the points on all of those guns (half your budget at least given the cheapest gun costs 400) and then on the ammo.

## Changelog
:cl:
balance: CAS no longer free equipment at roundstart. They get an injection of points which they then need to spend on everything. 
/:cl:

### DIFF
--- a/code/controllers/subsystem/points.dm
+++ b/code/controllers/subsystem/points.dm
@@ -12,7 +12,7 @@ SUBSYSTEM_DEF(points)
 
 	wait = 10 SECONDS
 
-	var/dropship_points = 0
+	var/dropship_points = 2275			// hey imagine a drone except they are sleeping on your couch and ur jacket is kind of under them with the blankets do you wake em up"
 	var/supply_points = 120
 	///Assoc list of xeno points: xeno_points_by_hive["hivenum"]
 	var/list/xeno_points_by_hive = list()
--- a/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
+++ b/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
@@ -32,8 +32,8 @@
 /turf/open/floor/mainship/mono,
 /area/mainship/hallways/hangar)
 "ah" = (
-/obj/structure/ship_ammo/heavygun,
-/turf/open/floor/mainship/cargo,
+/obj/structure/dropship_equipment/mg_holder,
+/turf/open/floor/mainship/sterile/plain,
 /area/mainship/hallways/hangar)
 "ai" = (
 /obj/machinery/vending/nanomed,
@@ -61,14 +61,13 @@
 /turf/open/floor/mainship/mono,
 /area/mainship/hallways/hangar)
 "al" = (
-/obj/structure/ship_ammo/minirocket/illumination,
-/turf/open/floor/mainship/cargo,
+/obj/structure/dropship_equipment/operatingtable,
+/turf/open/floor/mainship/sterile/plain,
 /area/mainship/hallways/hangar)
 "am" = (
 /obj/machinery/camera/autoname/mainship{
 	dir = 1
 	},
-/obj/structure/ship_ammo/minirocket/illumination,
 /turf/open/floor/mainship/cargo,
 /area/mainship/hallways/hangar)
 "an" = (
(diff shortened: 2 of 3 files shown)


SkeletalElite:
The only gun you DONT have at round start is the laser

Which is intentional. The laser is cost gated and not given for free because it's the best CAS weapon by a long a shot.

Chaznoodles:
Make a vendor instead, much like the standard marine ones, that let you vend a selection from different categories instead of "I am going to buy 4 fatties and then have nothing to do all round".

code/controllers/subsystem/points.dm
```diff
@@ -12,7 +12,7 @@ SUBSYSTEM_DEF(points)
 
 	wait = 10 SECONDS
 
-	var/dropship_points = 0
+	var/dropship_points = 2275			// hey imagine a drone except they are sleeping on your couch and ur jacket is kind of under them with the blankets do you wake em up"
```
Chaznoodles, line 15: 👀

lbnesquik:
I feel like splitting ammo or making things vendors based would end putting us in a situation where PO's aren't given as much freedom as they should have.

Leaving us with PO's who are just picking a GAU and a Rocket Pod and sticking to that. 
Not much more different from just picking a laser if you get down to it.  
Or spamming the GAU all the time, as it is the cheapest option and pretty much the only viable option to keep firing during any long round. 

As for the 1 weapon wonder, i don't actually 100% agree either. 
If you only have one gun, you t have as many options during a fire mission and if you don't have constant flares, you don't really want to waste that because your only weapon ran out of juice. 

If the laser is that powerful and none of the other weapons would get picked, then maybe the laser is overdue for a rebalancing of its values.

lbnesquik:
> 
> 
> Make a vendor instead, much like the standard marine ones, that let you vend a selection from different categories instead of "I am going to buy 4 fatties and then have nothing to do all round because I forgot to buy the launcher for them".

I actually want it to be a possibility. Let stupid PO screw themselves over if they want. Just like req can buy rockets without the launcher.

Chaznoodles:
> > Make a vendor instead, much like the standard marine ones, that let you vend a selection from different categories instead of "I am going to buy 4 fatties and then have nothing to do all round because I forgot to buy the launcher for them".
> 
> I actually want it to be a possibility. Let stupid PO screw themselves over if they want. Just like req can buy rockets without the launcher.

Here's the thing - Req buying rockets without the launcher doesn't completely invalidate a role.

TiviPlus:
Skeletalelite sums it up well,
I'd only really be fine with potentially removing the ammo

lbnesquik:
> 
> 
> > > Make a vendor instead, much like the standard marine ones, that let you vend a selection from different categories instead of "I am going to buy 4 fatties and then have nothing to do all round because I forgot to buy the launcher for them".
> > 
> > 
> > I actually want it to be a possibility. Let stupid PO screw themselves over if they want. Just like req can buy rockets without the launcher.
> 
> Here's the thing - Req buying rockets without the launcher doesn't completely invalidate a role.

Fair on the req point, but other roles can still try other things. Marines can pack 7 revolvers if they want. 
Req not giving you a gun but giving you ammo isn't the same as you being silly with your money.

davidchan13:
I'd say leave the launchers and weapon mounts but remove the ammo. A lot of the mounts in the printer are intentionally higher than they should be because you start with free stuff. Starting with no stock loadouts at all would require another pass on weapon prices

lbnesquik:
> 
> 
> I'd say leave the launchers and weapon mounts but remove the ammo. A lot of the mounts in the printer are intentionally higher than they should be because you start with free stuff. Starting with no stock loadouts at all would require another pass on weapon prices

I could settle on that, especially if a price update comes later on.

code/controllers/subsystem/points.dm
JackTheJackhammer: Do I or do I not wake up the drone?

TiviPlus:
Closing as stale

PR #6365, opened 2021-03-16
Neurotox Rework

fernandos13:
## About The Pull Request
neurotox no longer deal EXTRA TOX damage. and OXY damage on its own.

it will only deal as much as the stamina crit already do

the stam damage got changes tho

it starts at 8 now and drops down to 0.1 over 30cycles

if the marine gets in orange stamina "aka" slowdown but still no crit then the toxin will only halt stamina regen

BUT.. if left un atended after 31ticks it will start to pick up again and go up no matter what. "untill its purged"

## Why It's Good For The Game ##
TLDR:

if the marine is alread low at stamina or gets really close to crit before 31 cycles.
he will be kept from taking more stam damage "FROM THE NEUROTOX: if queen screeches. then you still take damage"
if the marine got full stamina it will deplet the starting stamina faster and slowdown for a while when hits orange

after 31 cycles no matter how much stamina the marine got the stam damage will grow over time untill treated or fully metabolized

## Changelog
:cl:
balance: neurotox have a higher stam damage upfront that lowers overtime
balance: neurotox have a "sweet spot" where the marine is slowdown but not lying down in crit yet
balance: neurotox stam will grow back again after 31 cycles if the marine does nothing about it
balance: neurotox extra oxy and Tox damage removed so will only take damage from the stamina itself
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -442,7 +442,7 @@
 	description = "A debilitating nerve toxin. Impedes motor control in high doses. Causes progressive loss of mobility over time."
 	reagent_state = LIQUID
 	color = "#CF3600" // rgb: 207, 54, 0
-	custom_metabolism = REAGENTS_METABOLISM * 2
+	custom_metabolism = REAGENTS_METABOLISM * 4
 	purge_list = list(/datum/reagent/medicine)
 	purge_rate = 0.2 //same as metabolism rate, effictevelly halving the use of drugs while not really purgin it all at once
 	overdose_threshold = 10000 //Overdosing for neuro is a lie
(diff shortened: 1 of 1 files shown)


Tennessee116:
> neurotoxin

> does not have serious toxic effects

I feel like, at this rate, it might deserve a rename just to clear the confusion.

SkeletalElite:
I like the other PR better

fernandos13:
comments above this one are outdated since recent changes and actually getting the Pull request well done

Walarks:
I think #6364 is better simply because it removes purge, which is an annoying and bad mechanic when spammed.

SplinterGP:
> balance: neurotox extra oxy and Tox damage removed so will only take damage from the stamina itself
can you test and see if this was because it woulndt take directly of stam, would probably ask to atomize this one as its one i might be up to pass easier

TiviPlus:
Neurotox is fine now so the redundancy is not needed anymore

PR #6366, opened 2021-03-16
Salvage dead xenos is gone

BraveMole:
## About The Pull Request

Drone and hivelord cannot recycle bodies anymore

## Why It's Good For The Game

No longer really needed with the maturity buff. And it's pretty lame to prevent marines from selling them that easily, they deserve a reward from killing xenos. We could even prevent xenos from dragging dead xenos bodies, maybe?

## Changelog
:cl:
del: Salvage biomass xeno ability removed.
/:cl:


Pariah919:
good pr
don't reward people for dying

Pinkufeck:
> don't reward people for dying

It gives nothing noticable though.

davidchan13:
So whats the difference here between marines dragging off xeno corpses for req points instead of having only the option to fulton them in the field and xenos being able to drag off corpses and cocooning them where as now they can only cocoon them where the body drops?

Glcbitj:
> > don't reward people for dying
> 
> It gives nothing noticable though.

Doesn't that mean removing it isn't noticeable too?

EvilMelonTheSecond:
> 
> 
> > > don't reward people for dying
> > 
> > 
> > It gives nothing noticable though.
> 
> Doesn't that mean removing it isn't noticeable too?

Removing it won't be noticable for xenos. Marines on the other hand miss out on like 30 req points per body at minimum.

SplinterGP (changes requested):
need keybinds removed

SplinterGP (approved):
nvm removing keybinds is uh wellelelelelele

Al-1ce:
> Removing it won't be noticable for xenos. Marines on the other hand miss out on like 30 req points per body at minimum.

That's the entire point. It's noticeable for Xenos because it means fewer req points for marines, so it makes it easier on their part of they try to recover bodies.

PR #6371, opened 2021-03-18
makes some power weapons (m4ra, railgun, anti material rifle) cheaper

Pariah919:
## About The Pull Request
railgun goes from 130 to 75, rounds reduced from 4 to 3 as you only get one. 
anti mat from 120 to 70, ammo normal halved in price
M4RA goes from 80 to 50

## Why It's Good For The Game
These guns are incredibly niche and only useful in certain situations and I don't really see the point of making as expensive anymore. Xenomorphs get infinite respawns now so power weapons don't need to be as limited

## Changelog
:cl:
balance:  power req guns reduced in price. railgun goes from 130 to 75, rounds reduced as you only get one. anti mat from 120 to 70, ammo normal halved in price M4RA goes from 80 to 50
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -205,12 +205,12 @@ WEAPONS
 /datum/supply_packs/weapons/railgun
 	name = "TX-220 Railgun"
 	contains = list(/obj/item/weapon/gun/rifle/railgun)
-	cost = 130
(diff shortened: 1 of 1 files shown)


David-Stormwell:
Seems fine from a balance view, but delete the M4RA when.

TiviPlus (approved):

PR #6377, opened 2021-03-19
Adds bump attack component to humans, adds bump attack compatibility to harvester and unarmed

Walarks:
## About The Pull Request
Adds bump attacks to harvester and fists.

You still can't FF with bump attacks or bump attack with a gun.

Tested bump attack component changes with a runner and queen and they should work the same as before.

## Why It's Good For The Game
Update: The bump attacks with wielded requirement and damage output reduction were tested and it appears that due to marines moving slower than xenos, especially when damaged or wearing heavy armor, bump attacks aren't as reliable as they are on xenos. This is due to them happening when you move and once a marine moves, it is unable to try and hit something by using the movement keys before it can move again. Unless marines are moving nearly as fast as xenos, this doesn't seem to be as big of a buff as it sounds for this weapon.

Targeting with melee is very clunky and this makes attacking more consistent, while also decreasing potential damage. This should give the option to attack in melee with less, but consistent, DPS rather than deal too much damage when your hits connect or none when they don't.

I think this would provide the option for having more consistent performance as a marine with a lower skill floor in exchange for decreased peak damage, which would help with increasing marines' overall ability to hold an advance, rather than rely too much on who's playing at the moment.

## Changelog
:cl:
add: added bump attack component to humans, added bump attack compatibility to wielded harvester and unarmed
/:cl:


code/game/objects/items/weapons/blades.dm
```diff
@@ -28,8 +28,9 @@
 	force = 60
 	attack_speed = 12
 	w_class = WEIGHT_CLASS_BULKY
-	flags_item = DRAINS_XENO
+	flags_item = DRAINS_XENO|TWOHANDED
```
BraveMole, line 31: undocumented? 
Also, wasn't vali only slighlty viable with a shield? I don't think bumpslash will make it sudendly OP

MLGTASTICa:
imagine if this was added onto shotguns..

code/datums/components/bump_attack.dm
```diff
@@ -2,34 +2,38 @@
 	var/active = TRUE
 	var/bump_action_path
 	var/datum/action/bump_attack_toggle/toggle_action
+	var/mob/living/bumper
```
BraveMole, line 5: You're not supposed to hold a casted reference to a component parent, because this removes some genericty.
That being said, are bumped attacked supposed to be used by something not a mob/living? I don't think it will ever be the case, so why no. Not sure Tivi will like this though

code/datums/components/bump_attack.dm
```diff
@@ -69,33 +69,47 @@
 
 /datum/component/bump_attack/proc/human_bump_action(datum/source, atom/target)
 	SIGNAL_HANDLER
-	var/mob/living/carbon/human/bumper = parent
+	INVOKE_ASYNC(src, .proc/human_bump_action_checks, source, target)
```
BraveMole, line 72: I'm not seing any sleep in the next procs, why are you using a wrapper function with an INVOKE ASYNC?

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -41,6 +41,7 @@
 	RegisterSignal(src, COMSIG_KB_RAILATTACHMENT, .proc/do_activate_rail_attachment)
 	RegisterSignal(src, COMSIG_GRAB_SELF_ATTACK, .proc/fireman_carry_grabbed) // Fireman carry
 	AddComponent(/datum/component/footstep, FOOTSTEP_MOB_HUMAN)
+	AddComponent(/datum/component/bump_attack)
```
BraveMole, line 44: I don't like very much that approach. That means that every human will get the toggle action, while less than 2-5% will actually use it. We don't want to have useless actions lying around.

Even performance wise, you are checking if human can bump or not, even if they are using a minigun loadout... Doesn't make sense.

Either you manage to make the giving action and the registering only when you wield the vali harvester, or you make a bump component for the weapon and not for the living. Also, who cares about unarmed attacks, don't tackle that unless it's really easy to add and doesn't lead to shitcode

BraveMole (changes requested):

code/datums/components/bump_attack.dm
```diff
@@ -2,34 +2,38 @@
...
+	toggle_action = new()
+	toggle_action.give_action(bumper)
 	toggle_action.update_button_icon(active)
 	RegisterSignal(toggle_action, COMSIG_ACTION_TRIGGER, toggle_path)
-	if(active)
-		RegisterSignal(parent, COMSIG_MOVABLE_BUMP, bump_action_path)
 
 /datum/component/bump_attack/Destroy(force, silent)
+	if(active)
+		UnregisterSignal(bumper, COMSIG_MOVABLE_BUMP)
```
BraveMole, line 30: unneeded, signals are automaticly cleaned on object destroy

code/datums/components/bump_attack.dm
```diff
@@ -2,34 +2,38 @@
...
 	QDEL_NULL(toggle_action)
+	bumper = null
```
BraveMole, line 32: Don't take my word for it, but i think you also need to clean that ref when bumper is QDEL. Ask tivi about this, I don't think you are protecting this from hard del enough. It's possible that the component is automaticly destroyed on the parent destroy.

BraveMole (changes requested):

code/game/objects/items/weapons/blades.dm
Walarks: This is required to allow harvester to be wielded.

code/datums/components/bump_attack.dm
Walarks: I think the core signals and checks used for the component require it to be on a living being, so if it was used for object, it would need to be rewritten anyways. 

I added this because for some reason "var/mob/living/bumper = parent" on each proc caused runtimes when using the component on humans.

code/game/objects/items/weapons/blades.dm
BraveMole: Why are you requiring it to be wielded for the bump attack? Is it a design intent or because you didn't find a way to do it without?

Walarks: It's a design intent so people don't use shield + vali and such.

BraveMole: Eh, your choice, I don't see why you keep making vali a weak setup

Losenis:
machetes get this too?

Walarks:
> machetes get this too?

No, probably after this has been tested and is ok to be a thing.

code/datums/components/bump_attack.dm
TiviPlus, line 5: this is terribad

code/datums/components/bump_attack.dm
TiviPlus: yes

code/datums/components/bump_attack.dm
```diff
@@ -69,33 +69,47 @@
...
+
+/datum/component/bump_attack/proc/human_bump_action_checks(datum/source, atom/target)
+	var/mob/living/carbon/human/attacker = bumper
 	. = carbon_bump_action_checks(target)
 	if(!isnull(.))
 		return
 	var/mob/living/living_target = target
-	if(bumper.faction == living_target.faction)
+	if(attacker.faction == living_target.faction)
 		return //FF
```
TiviPlus, line 81: ??

code/datums/components/bump_attack.dm
```diff
@@ -69,33 +69,47 @@
...
 	if(bumper.next_move > world.time)
 		return COMPONENT_BUMP_RESOLVED //We don't want to push people while on attack cooldown.
 	bumper.UnarmedAttack(target, TRUE)
 	GLOB.round_statistics.xeno_bump_attacks++
 	SSblackbox.record_feedback("tally", "round_statistics", 1, "xeno_bump_attacks")
 	TIMER_COOLDOWN_START(src, COOLDOWN_BUMP_ATTACK, CLICK_CD_MELEE)
 	return COMPONENT_BUMP_RESOLVED
+
+/datum/component/bump_attack/proc/human_do_bump_action(atom/target)
+	if(bumper.next_move > world.time)
```
TiviPlus, line 105: No autodoc

code/game/objects/items/weapons/blades.dm
TiviPlus, line 31: hello yes unrelated changes

code/modules/mob/living/carbon/human/human.dm
TiviPlus: yes that

TiviPlus (changes requested):

code/game/objects/items/weapons/blades.dm
Walarks: It needs the twohanded flag to be able to be wielded.

bowlofcereal:
wtf no bump to machete but bump to harvester? this is a personal attack

code/datums/components/bump_attack.dm
Walarks: autodocced

code/datums/components/bump_attack.dm
Walarks: removed var, INVOKE_ASYNC is needed for sleeping procs

code/datums/components/bump_attack.dm
Walarks: removed var

code/datums/components/bump_attack.dm
Walarks: removed var

code/datums/components/bump_attack.dm
Walarks: removed section

code/modules/mob/living/carbon/human/human.dm
Walarks: Made it so that the component's checks and such need to be enabled for the component to run.

code/datums/components/bump_attack.dm
Walarks: There are sleeps in procs that are called further down.

code/datums/components/bump_attack.dm
Walarks: removed var

Smudged:
Bump attack to FIST?! 

This is a mistake, of well, just marines walking forward

davidchan13:
Gimme Bayonet and Entrenching Tool bump attacks.

code/datums/components/bump_attack.dm
```diff
@@ -1,42 +1,51 @@
 /datum/component/bump_attack
-	var/active = TRUE
+	///Whether the component is active
+	var/active
```
TiviPlus, line 3: Set the default for bools to be = FALSE

code/datums/components/bump_attack.dm
```diff
@@ -1,42 +1,51 @@
...
-/datum/component/bump_attack/proc/living_activation_toggle(datum/source)
+/datum/component/bump_attack/proc/living_activation_toggle(datum/source, should_enable = !active)
+	if(should_enable == active)
+		return
 	var/mob/living/bumper = parent
-	active = !active
 	to_chat(bumper, "<span class='notice'>You will now [active ? "attack" : "push"] enemies who are in your way.</span>")
-	if(active)
+	if(toggle_action)
+		toggle_action.update_button_icon(should_enable)
```
TiviPlus, line 38: use .?

code/datums/components/bump_attack.dm
```diff
@@ -1,42 +1,51 @@
...
+	if(should_enable)
+		active = TRUE
 		RegisterSignal(bumper, COMSIG_MOVABLE_BUMP, bump_action_path)
-	else
-		UnregisterSignal(bumper, COMSIG_MOVABLE_BUMP)
-	toggle_action.update_button_icon(active)
+		return
+
+	var/obj/item/held_item = bumper.get_inactive_held_item()
+	if(held_item && held_item.flags_item & CAN_BUMP_ATTACK)
```
TiviPlus, line 45: Use .?

code/datums/components/bump_attack.dm
TiviPlus: Then it should only invoke the proc that actually sleeps which I assume is human_do_bump_action

code/game/objects/items/weapons/blades.dm
```diff
@@ -28,8 +28,9 @@
...
 
+	var/force_wielded = 40
```
TiviPlus, line 33: Autodoc

code/game/objects/items.dm
```diff
@@ -355,6 +355,16 @@
...
 
+///Used to enable/disable an item's bump attack. Grouped in a proc to make sure the signal or flags aren't missed
+/obj/item/proc/toggle_item_bump_attack(mob/user, enable_bump_attack)
+	SEND_SIGNAL(user, COMSIG_ITEM_TOGGLE_BUMP_ATTACK, enable_bump_attack)
+	if(flags_item & CAN_BUMP_ATTACK && enable_bump_attack)
+		return
+	if(enable_bump_attack)
+		flags_item |= CAN_BUMP_ATTACK
+		return
+	flags_item &= ~CAN_BUMP_ATTACK
```
TiviPlus, line 366: Why is this proc neeeded

TiviPlus (changes requested):

code/datums/components/bump_attack.dm
Walarks: changed

code/datums/components/bump_attack.dm
Walarks: changed

code/datums/components/bump_attack.dm
Walarks: changed

code/game/objects/items/weapons/blades.dm
Walarks: changed

code/game/objects/items.dm
Walarks: It's needed since both the bitflag and signal are necessary for the component to function properly and this makes the code more ordered.

code/datums/components/bump_attack.dm
Walarks: changed

TiviPlus (approved):
Tested?

Walarks:
> Tested?

Works based on my tests

TiviPlus (approved):

PR #6393, opened 2021-03-21
Psy draining is quicker

BraveMole:
## About The Pull Request

Psy drain will now take 5 seconds instead of 10 seconds

## Why It's Good For The Game

With psy drain being the combat option over cocoon, it seems completly fair that it should take less time than its counterpart. This reinforce the idea for xenos that they should use it on the frontline, while cocoon should be reserved near backline 

## Changelog
:cl:
balance: Psy drain will now take 5 seconds instead of 10 seconds
/:cl:


JackTheJackhammer:
I would suggest to lower it to 7.  5 seems too fast.

BraveMole:
> I would suggest to lower it to 7. 5 seems too fast.

psy drain give way less than cocoon, wich take only 10 seconds to do. Halving the time when it gives 1/3 of the outcome seems ok

TiviPlus (approved):

Issue #6382, opened 2021-03-19
CIC orders invisible through overwatch


Losenis:
You can see the first CIC order you give through overwatch, but later orders are invisible so you have no feedback on whether they're given or not. Tried as both synth and SO.

PR #6406, opened 2021-03-22
Protection against temporary visuals hard del

BraveMole:
## About The Pull Request

Title

## Why It's Good For The Game

Might be useless, might fix https://github.com/tgstation/TerraGov-Marine-Corps/issues/6382

## Changelog
:cl:
code: Protection against temporary visuals hard del
/:cl:

### DIFF
--- a/code/datums/atom_hud.dm
+++ b/code/datums/atom_hud.dm
@@ -56,8 +56,10 @@ GLOBAL_LIST_INIT(huds, list(
 
 
 /datum/atom_hud/proc/remove_from_hud(atom/A)
+	SIGNAL_HANDLER
 	if(!A)
 		return FALSE
+	UnregisterSignal(A, COMSIG_PARENT_QDELETING)
 	for(var/u in hudusers)
 		var/mob/M = u
 		remove_from_single_hud(M, A)
@@ -101,6 +103,8 @@ GLOBAL_LIST_INIT(huds, list(
 	if(!A || (A in hudatoms))
 		return FALSE
 	hudatoms |= A
+	if(istype(A, /obj/effect/temp_visual))
+		RegisterSignal(A, COMSIG_PARENT_QDELETING, .proc/remove_from_hud)
 	for(var/u in hudusers)
 		var/mob/M = u
 		if(!queued_to_see[M])


code/datums/atom_hud.dm
```diff
@@ -56,8 +56,10 @@ GLOBAL_LIST_INIT(huds, list(
 
 
 /datum/atom_hud/proc/remove_from_hud(atom/A)
+	SIGNAL_HANDLER
 	if(!A)
 		return FALSE
+	UnregisterSignal(A, COMSIG_PARENT_QDELETING)
```
BraveMole, line 62: is it better to istype before unregistering? I dunno

code/datums/atom_hud.dm
```diff
@@ -101,6 +103,8 @@ GLOBAL_LIST_INIT(huds, list(
 	if(!A || (A in hudatoms))
 		return FALSE
 	hudatoms |= A
+	if(istype(A, /obj/effect/temp_visual))
+		RegisterSignal(A, COMSIG_PARENT_QDELETING, .proc/remove_from_hud)
```
BraveMole, line 107: i'm afraid this might lead one day to double registering runtime. Nothing in test so far

code/datums/atom_hud.dm
SplinterGP: why would it

code/datums/atom_hud.dm
SplinterGP: can this proc be called twice without the unregistering one ?

BraveMole: i don't know. I don't think so, it shouldn't be, but this code is used by a lot and a lot of things

TiviPlus, line 107: A) Why only temp visuals and not all atoms
B) I'm  think this registers twice in some cases because of protections we added recently

code/datums/atom_hud.dm
TiviPlus: Do images have signal tables

TiviPlus (changes requested):

TiviPlus (approved):

PR #6415, opened 2021-03-22
Eggs can be done only by defiler

BraveMole:
## About The Pull Request

Shrike and queen lose the ability to produce eggs, defiler gains it

## Why It's Good For The Game

Remnant of capture gameplay begone. Defiler get it because it can make egg bomb

## Changelog
:cl:
balance: queen lose the ability to produce eggs, defiler gains it
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defiler/castedatum_defiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defiler/castedatum_defiler.dm
@@ -42,6 +42,7 @@
 		/datum/action/xeno_action/activable/psydrain,
 		/datum/action/xeno_action/plant_weeds,
 		/datum/action/xeno_action/activable/larval_growth_sting/defiler,
+		/datum/action/xeno_action/lay_egg,
 		/datum/action/xeno_action/activable/inject_egg_neurogas,
 		/datum/action/xeno_action/activable/emit_neurogas,
 		/datum/action/xeno_action/select_reagent,
--- a/code/modules/mob/living/carbon/xenomorph/castes/queen/castedatum_queen.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/queen/castedatum_queen.dm
@@ -56,7 +56,6 @@
 		/datum/action/xeno_action/activable/build_silo,
 		/datum/action/xeno_action/activable/build_turret,
 		/datum/action/xeno_action/place_acidwell,
-		/datum/action/xeno_action/lay_egg,
 		/datum/action/xeno_action/activable/larval_growth_sting,
 		/datum/action/xeno_action/call_of_the_burrowed,
 		/datum/action/xeno_action/activable/screech,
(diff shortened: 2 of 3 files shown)


davidchan13:
I feel like carrier should get the ability to lay eggs too, since they have the ability to store them too.

The only major side effect here is that Corrupted hives won't get to make more eggs without admin intervention, but that's not really much of a balance concern since corrupted hived are already adminbus to begin with.

BraveMole:
> I feel like carrier should get the ability to lay eggs too, since they have the ability to store them too.
> 
> The only major side effect here is that Corrupted hives won't get to make more eggs without admin intervention, but that's not really much of a balance concern since corrupted hived are already adminbus to begin with.

Yes carrier needs it too, but i can't touch it because of the rework that is not merged

ThelastArchiver:
Why exactly are you removing the queen/shrike ability to lay eggs? You could also just give defiler egg laying ability if at all then the removal of the queen egg laying removes one of her core canon abilities and that just feels very weird.

BraveMole:
> Why exactly are you removing the queen/shrike ability to lay eggs? You could also just give defiler egg laying ability if at all then the removal of the queen egg laying removes one of her core canon abilities and that just feels very weird.

It's useless. Core canon abilities from capture gameplay, we don't do that here

ThelastArchiver:
It's useless. Core canon abilities from capture gameplay, we don't do that here. And queen has 15000 buttons, better remove some                               It just feels so wrong like we do know the queen from the franchise as the big mean egg laying monster and that is what makes her the core of the hive without her the hive can’t prosper

Smudged:
Maybe adding a clause that corrupted Shrike's can still lay eggs will solve the problem of NEEDING admin intervention whenever a researcher make a corrupted hive.

Smudged:
> It's useless. Core canon abilities from capture gameplay, we don't do that here. And queen has 15000 buttons, better remove some It just feels so wrong like we do know the queen from the franchise as the big mean egg laying monster and that is what makes her the core of the hive without her the hive can’t prosper

Now to be fair, what TGMC has is becoming further and further from what "canon" alien has.

I'd even go as far to say the aliens on TGMC are practically their own creature, although looking like some alien. 

Not many people are gonna miss the core-canon of what CM and TGMC has become, as we move further away, I mean more people can make whatever they want thus opening more creative avenues, and not be weighed down by the alien "canon" (Which has historically changed through different media, like comics movies etc.

davidchan13:
> Maybe adding a clause that corrupted Shrike's can still lay eggs will solve the problem of NEEDING admin intervention whenever a researcher make a corrupted hive.

 It already requires an admin to corrupt the eggs and allow a corrupted hive. Admins can just spawn more corrupted eggs (giving the RSR or C. Shrike somethings to do first if they feel necessary) or just giving them X amount of eggs to keep the corrupted hive limited. This is a non issue not worth adding extra code for something that already requires admin intervention.

TiviPlus (changes requested):
Keep it on shrike

TiviPlus (approved):

PR #6433, opened 2021-03-26
Arith Replacement

TGCMbetcher:
## About The Pull Request

Replaces Arith with Hyro in injectors.

## Why It's Good For The Game

 Arith is more expensive to make, worse, and dangerous. There is literally no reason for it to be in injectors other then to nerf the usage.

## Changelog
:cl:
qol: Swaps arith for hyro in injectors
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -173,7 +173,7 @@
 	icon_state = "autoinjector-1"
 	list_reagents = list(
 		/datum/reagent/medicine/synaptizine = 3,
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

Dolth:
You should make arith different instead of just removing it from injectors.

TGCMbetcher:
> You should make arith different instead of just removing it from injectors.

Theres nothing wrong with arith its just extra, its for a triple detox pill. Hyro is clearly the best anti tox drug and you use it to make arith. Changing arith would be removing something good from the game when there is something that perfectly fits the role but just isnt used right now.

PR #6460, opened 2021-04-01
Brings upgraded metal cades up to par with plasteel cades

Walarks:
## About The Pull Request
Metal cades have their upgraded durability increased to be better comparable to plasteel cades that have 500 durability.

Bomb upgrade - 400 durability
Acid/Melee upgrades - 300 durability

## Why It's Good For The Game
Metal cades are now a more competitive option to sandbags and upgrading them is worth it. This makes the whole fort more uniform durability-wise, rather than have the doors be 150% more durable than the walls, a.k.a have 500 durability, while the normal walls have only 200 durability.

## Changelog
:cl:
balance: upgraded metal cades durability increased from 200. Bomb upgrade is 400 durability, acid/melee upgrades have 300 durability
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -420,7 +420,7 @@
 	name = "metal barricade"
 	desc = "A sturdy and easily assembled barricade made of metal plates, often used for quick fortifications. Use a blowtorch to repair."
 	icon_state = "metal_0"
-	max_integrity = 200
+	max_integrity = 400
 	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
 	coverage = 128
 	stack_type = /obj/item/stack/sheet/metal
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -420,7 +420,7 @@
 	name = "metal barricade"
 	desc = "A sturdy and easily assembled barricade made of metal plates, often used for quick fortifications. Use a blowtorch to repair."
 	icon_state = "metal_0"
-	max_integrity = 400
+	max_integrity = 200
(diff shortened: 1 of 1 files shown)


neohenry:
This with the combat engineer buff is going to make FOB really strong.

Swagile:
id rather the price of mats be lowered than to make siege worse

Glcbitj:
Bruh, you fucking doubled their HP, their main thing metal has over sandbags is that they are repairable using welding fuel, which is an almost endless resource and they can be upgraded (with more resources), more HP is nice, but maybe not double it perhaps. 

With that said, they do seemed too weak compared to sandbags, or at least the sandbags seemed too strong since their durable and quick to build. Plus we rarely see engies building stuff beyond the main FoB (that's probably due to the necessity of rushing pools though)

Swagile:
the issue is that this stacks with upgrading the cade

so you get something that is easily repairable, high HP, and can be upgraded to be resistant to at least 1 other damage type, making this better than plastisteel; the only reason to use plastisteel after that is to just make doors

DerrickChamberlain:
Yes please! Metal cades are so laughably bad.

TiviPlus:
doubling is way too much

TiviPlus:
plus yes cade upgrades exit

usnpeepoo:
engineers already got a starter matt buff and a double hp'ing pr for the metal cades isnt right tbh

SkeletalElite:
Do 300 instead

SplinterGP:
metal cades shoulnd't be in par with plasteel cades as plasteel cades are stronger and pricier cades, no sense on that

Swagile:
mats being weaker but cheaper is a better solution as it stops FOB meta (since marines excel in defending 1 or 2 strong cadelines over 3-4 more cadelines but weaker)

this PR would be better if xenos had good siege but currently, no

BraveMole:
Xenos are still terrible at siege.
I think engie got enough buffs for now, let the meta rest a little on this part.

TiviPlus:
Engi has enough buffs

PR #6465, opened 2021-04-02
Two more synth paygrades

SpaceLoveSs13:
## About The Pull Request

Two new synthetic paygrades added.

## Why It's Good For The Game

Previously I had added three synthetic oaygrades that had low range of time. Now that we have more and more people with unterest to join and play synthetic after job slot PR, i believe we should have only two paygrades that are very high range.

## Changelog
:cl:
add: Two more synth paygrades. 100hrs and 300hrs.

### DIFF
--- a/code/datums/jobs/access.dm
+++ b/code/datums/jobs/access.dm
@@ -349,9 +349,9 @@
 			. = size ? "Mk.II " : "Mark II"
 		if("Mk.I")
 			. = size ? "Mk.I " : "Mark I"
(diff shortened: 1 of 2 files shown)


SpaceLoveSs13:
Created PR in phone. Probably need to test on PC later.

TiviPlus:
3 Ranks max for standard roles, I'm keeping to this

Losenis:
Could stretch out MkII to 50h and MkIII to 100h

PR #6466, opened 2021-04-02
Belt

TGCMbetcher:
## About The Pull Request

Lets medics use the machine pistol or t90 and store ammo in their belt slot. 

## Why It's Good For The Game

Right now the belt is never used unless they pick up a t14 or something and go onehanded medic with it always wrapped around their wrist. Slotting 7 t90 or t19 mags into the belt and then keeping a minimum amount of kits would let them avoid combat backpack.

## Changelog
:cl:
qol: Lets small SMG ammo fit in 276 medical belt
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -173,7 +173,7 @@
 	icon_state = "autoinjector-1"
 	list_reagents = list(
 		/datum/reagent/medicine/synaptizine = 3,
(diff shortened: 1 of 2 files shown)


TGCMbetcher:
I would get it if people are against the other SMGs so this could just be for machine pistol, so we could nerf the size so only a limited amount of the larger SMG ammo could fit. At the moment though its the same size as pistol ammo but since its copy pasted from CM it doesnt let them fit. T19 absolutely should fit and its been nerfed enough that a storage buff would be welcome.

PR #6467, opened 2021-04-02
276 Belt Storage

TGCMbetcher:
## About The Pull Request

Allows small sized SMG mags to fit in 276 pattern medical belt

## Why It's Good For The Game

A belt thats almost never used except by pistol medics that one hand so they always have a weapon even when knocked down. Now allows t90 and t19 to be used in the same fashion. It should have always allowed t19 but it wasnt in CM and it looks like it was copied over.

## Changelog
:cl:
balance: Medical belt(not lifesaver bag) now fits small SMG mags.
/:cl:


TGCMbetcher:
"and light ammunitions" The t19 should have always been added, if you want to not add the other SMGs ammo they shouldnt have been small in the first place but whatever, this is mostly for the t19.

SplinterGP (approved):

PR #6471, opened 2021-04-02
Silo pinpointers have max range

BraveMole:
## About The Pull Request

Silo pinpointer cannot track the location when the silo is farther than 35 tiles.
Standard marine can get a silo pinpointer in his vendor

## Why It's Good For The Game

This will allow xenos to have way more variety in their silo placement. This forces marines to do some scouting, will lower the usefullness of marines ungaball and maybe (maybe) promotes smaller squad play. 

## Changelog
:cl:
balance: Silo pinpointer cannot track silos that are at a distance greater than 35 tiles
balance: Each standard marine get a silo pinpointer in its vendor
/:cl:

### DIFF
--- a/code/game/objects/items/pinpointer.dm
+++ b/code/game/objects/items/pinpointer.dm
@@ -86,9 +86,28 @@
 		to_chat(user, "Extreme danger.  Arming signal detected.   Time remaining: [bomb.timeleft]")
 
 /obj/item/pinpointer/pool
-	name = "pool pinpointer"
-	desc = "A pinpointer able to detect the psychic energy emmaning from spawning pools"
+	name = "silo pinpointer"
+	desc = "A pinpointer able to detect the psychic energy emmaning from silos"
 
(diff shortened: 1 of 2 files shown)


usnpeepoo:
finally now tadpolians can no longer get a general idea of where a silo is from the ship and drop there

SplinterGP:
ain't 35 tiles a bit too low ? 2 screens.5 ?

Glcbitj:
Great, now Big Red is even more of a pain if you're playing marine and silos are even more cancerous to deal with.

bowlofcereal:
to be fair xenos always put the pools in the same areas so you dont even need a pointer to guess where they're gonna be

BraveMole:
> to be fair xenos always put the pools in the same areas so you dont even need a pointer to guess where they're gonna be

That's cause why even try to be creative, when they'll find the silo anyway

BraveMole:
Eh, maybe latter, probably not that great of an idea right now

PR #6481, opened 2021-04-02
Razorwire now buildable from metal menu

Swagile:
## About The Pull Request

Makes razorwire infinitely easier to make. Just use metal in hand and click on razorwire. Old method stays in the game cause I don't want to comment it out lest it be used for other projects (as example code, etc).

## Why It's Good For The Game

Less hassle and more QoL for engis.

## Changelog
:cl:
qol: You no longer have to fumble around with barbed wire and rods to create razorwire; it can now be made straight from the metal menu.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -32,6 +32,7 @@ GLOBAL_LIST_INIT(metal_recipes, list ( \
 	new/datum/stack_recipe("metal rod", /obj/item/stack/rods, 1, 4, 60), \
 	null, \
 	new/datum/stack_recipe("barbed wire", /obj/item/stack/barbed_wire, 2, 1, 20, time = 1 SECONDS, skill_req = SKILL_CONSTRUCTION_METAL), \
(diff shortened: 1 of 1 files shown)


EvilMelonTheSecond:
The old way of building is useful for when razorwire gets destroyed and it drops barbed wire you can use with rods to rebuild, I'm sure leaving it in is a good choice unless you want less QoL for recycling barbed wire for some reason.

Swagile:
> 
> 
> The old way of building is useful for when razorwire gets destroyed and it drops barbed wire you can use with rods to rebuild, I'm sure leaving it in is a good choice unless you want less QoL for recycling barbed wire for some reason.

I thought I forgot something; forgot you could recycle.

DerrickChamberlain:
> > The old way of building is useful for when razorwire gets destroyed and it drops barbed wire you can use with rods to rebuild, I'm sure leaving it in is a good choice unless you want less QoL for recycling barbed wire for some reason.
> 
> I thought I forgot something; forgot you could recycle.

I figured it was the same system of dropping when broken.

SplinterGP (approved):

PR #6483, opened 2021-04-02
Power Armor made useful again

Swagile:
## About The Pull Request

Rebalances B17 and B18 to actually be power armors worth buying. Changelog has the most details.

## Why It's Good For The Game

B17 and B18 are really subpar power armors despite their price tag and their purpose. They are supposed to be vanguards that push in ahead of everyone while tanking, but it does this job poorly. Not only is it only 5 melee above heavy jaeger tyr 2 (which is only 12 points as opposed to 60 or 100), it doesn't have the utility of jaeger, or the speed of PAS despite being a req buyable item (for a hefty price tag). The same is for B18; its even worse as it has the slowest movement debuff in the game even without welding a gun (1.15). This PR serves to solve this issue by giving both medium tags (0.55 debuff instead), upping their values, and hence making this less of a "well I have too much money and nothing to spend on" to "I need to upgrade marines armor to deal with good xenos, lets buy B17 or B18 for one good marine to rally around".

The speed buff is also because, again, its a medium to high priced req buyable item; you shouldn't be punished for wearing gear that is supposed to help you.

## Changelog
:cl:
add: B18 gauntlets to B18 req crate.
balance: B17 and B18 rebalanced; 70 melee for chest, 75 melee for helmet on both armor sets.
balance: B18 and B17 now have 0.55 slowdown, making it only 0.05 slower than PAS.
balance: In return, B18 is now 120 points from 100 points, and B17 is 80 points from 60.
/:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -283,11 +283,11 @@
 	name = "\improper B18 defensive armor"
 	desc = "A heavy, rugged set of armor plates for when you really, really need to not die horribly. Slows you down though.\nHas an automated diagnostics and medical system for keeping its wearer alive."
 	icon_state = "xarmor"
-	soft_armor = list("melee" = 65, "bullet" = 75, "laser" = 75, "energy" = 80, "bomb" = 80, "bio" = 65, "rad" = 20, "fire" = 80, "acid" = 65)
+	soft_armor = list("melee" = 75, "bullet" = 75, "laser" = 75, "energy" = 80, "bomb" = 80, "bio" = 65, "rad" = 20, "fire" = 80, "acid" = 65)
 	flags_armor_protection = CHEST|GROIN|ARMS|LEGS|FEET|HANDS
 	flags_cold_protection = CHEST|GROIN|ARMS|LEGS|FEET|HANDS
 	flags_heat_protection = CHEST|GROIN|ARMS|LEGS|FEET|HANDS
-	slowdown = SLOWDOWN_ARMOR_VERY_HEAVY
+	slowdown = SLOWDOWN_ARMOR_MEDIUM
 	supporting_limbs = CHEST | GROIN | ARM_LEFT | ARM_RIGHT | HAND_LEFT | HAND_RIGHT | LEG_LEFT | LEG_RIGHT | FOOT_LEFT | FOOT_RIGHT | HEAD //B18 effectively stabilizes these.
 	resistance_flags = UNACIDABLE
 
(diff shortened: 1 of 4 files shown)


code/modules/clothing/suits/marine_armor.dm
```diff
@@ -283,11 +283,11 @@
 	name = "\improper B18 defensive armor"
 	desc = "A heavy, rugged set of armor plates for when you really, really need to not die horribly. Slows you down though.\nHas an automated diagnostics and medical system for keeping its wearer alive."
 	icon_state = "xarmor"
-	soft_armor = list("melee" = 65, "bullet" = 75, "laser" = 75, "energy" = 80, "bomb" = 80, "bio" = 65, "rad" = 20, "fire" = 80, "acid" = 65)
+	soft_armor = list("melee" = 75, "bullet" = 75, "laser" = 75, "energy" = 80, "bomb" = 80, "bio" = 65, "rad" = 20, "fire" = 80, "acid" = 65)
```
SplinterGP, line 286: i don't know how good this actually is tbh, this makes 80 armor with the jumpsuit, might want to buff both it and b17 to 70 ?, even more counting both have same slowdown now

code/modules/reqs/supplypacks.dm
```diff
@@ -670,6 +670,7 @@ ARMOR
 	contains = list(
 		/obj/item/clothing/suit/storage/marine/specialist,
 		/obj/item/clothing/head/helmet/marine/specialist,
+		/obj/item/clothing/gloves/marine/specialist,
```
SplinterGP, line 673: b17 doenst has these too ?

Swagile: yeah cause this is B18 gloves (which isn't named B18 for some reason)

code/modules/clothing/suits/marine_armor.dm
Swagile: i buffed it up cause you pay almost 2x more for B18 than B17; but if thats too much i can up the price or lower it back down to 70 on chest instead

which ones a better solution?

code/modules/reqs/supplypacks.dm
SplinterGP: no i mean, doenst b17 have a B17 gauntlets as well ? funky

code/modules/clothing/suits/marine_armor.dm
SplinterGP: you can put down the price, i believe, too much costly also means its less used overall

code/modules/reqs/supplypacks.dm
Swagile: > 
> 
> no i mean, doenst b17 have a B17 gauntlets as well ? funky

not that i can see in the code; its just B18

code/modules/clothing/suits/marine_armor.dm
Swagile: lowered to 75 chest and helmet armor on par with B17, keeping price to 120 instead of 130 to compensate for autodoc of B18

SplinterGP (approved):

PR #6484, opened 2021-04-03
Minigun changes

Naaanii:
No longer makes Minigun a more expensive GPMG that's harder to maintain.

## About The Pull Request

Buffs minigun to deal less shrap, wield slightly faster and reduces slowdown while being wielded and increases the drum size.

## Why It's Good For The Game

With the GPMG on the rise, minigun feels very underwhelming and not worth buying. This should make it worth buying and a better change than just reducing the price and makes it more of a threat and less of a double edged sword and not feel like a worse GPMG.

## Changelog
:cl:
balance: Reduces shrap chance, wield time and wield slowdown on the minigun.
balance: Increases minigun drum capacity.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1008,7 +1008,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	accurate_range = 5
 	damage = 25
 	penetration = 15
-	shrapnel_chance = 25
+	shrapnel_chance = 5
 	sundering = 2.5
 
 /datum/ammo/bullet/railgun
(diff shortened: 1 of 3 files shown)


Dolth:
Minigun stronger than GPMG, sure, but minigun with little slowdown? Frankly, keep the slowdown, raise ROF to GPMG if it isn't already the case, but lower slowdown is too big imo.

Naaanii:
> Minigun stronger than GPMG, sure, but minigun with little slowdown? Frankly, keep the slowdown, raise ROF to GPMG if it isn't already the case, but lower slowdown is too big imo.

Slowdown reduction was very mild but also done to help since its normally paired with heavy armors. 
Will take it into consideration, though.

Naaanii:
> Minigun stronger than GPMG, sure, but minigun with little slowdown? Frankly, keep the slowdown, raise ROF to GPMG if it isn't already the case, but lower slowdown is too big imo.

Removed Wield speed buff after some testing.

TiviPlus:
Nerf GPMG instead

PR #6487, opened 2021-04-03
Queen Screech Update PR. 

InterroLouis:
## About The Pull Request

Essentially, this fixes a bug I found recently that...probably has been here a long time? Probably. Maybe. Basically upon testing screech and examining the vars of the affected human at various ranges, I found that, well, screeching right next to the target was doing 70 stamina damage while screeching at screen edge from the target was doing 130, and the admittedly short stun time was also doing...wrong. So I took the numbers and swapped them and tested it and hey, wouldn't you know, it worked. Last seen numbers said in PRs have stated intended screech values as 130 beside Queen, and 70 at screen edge, so I've kept those numbers. The 140 is for purposes of having that be the actual numbers, since with the max as 130, you would be only getting 120 stamina damage while adjacent to the Queen. Adjusting the lower number to 70 is to keep the scaling(reduction of 10 per tile distance) the same. This will also mean that being on the same tile as the Queen will do 140 stamina damage, but, hey. Marines will just have to deal with taking a little more stamina damage if a Queen shouts while standing on their chest. 

## Why It's Good For The Game

Fixes a bug. It just works. 

## Changelog
:cl:
balance: Same tile screech does a bit more stamina damage. 
fix: Screech stamina damage no longer reversed. 
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -414,8 +414,8 @@ Contains most of the procs that are called when a mob is attacked by something
 	var/reduce_prot_aura = protection_aura * 0.1
 
 	var/reduction = max(min(1, reduce_within_sight - reduce_prot_aura), 0.1) // Capped at 90% reduction
-	var/stamina_damage = LERP(60, 130, dist_pct) * reduction //Max 130 beside Queen, 60 at the edge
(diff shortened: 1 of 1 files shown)


BraveMole:
lel, good catch

DerrickChamberlain:
Nice work!

SplinterGP (approved):

TiviPlus (approved):

PR #6490, opened 2021-04-03
Prae spray behaviour fix

BraveMole:
## About The Pull Request

Rework made acid spray not having the same behaviour as before, this fixes it
Also make fresh acid stun for 1 second rather than 0.5, since apparently 0.5 was not enough to fix the exploit

## Changelog
:cl:
fix: Prae acid spray back to normal
balance: Fresh acid (from acid spray) stun when you walk on it, acid is considered fresh for 1 second rather than 0.5 previously
/:cl:

### DIFF
--- a/code/game/objects/effects/aliens.dm
+++ b/code/game/objects/effects/aliens.dm
@@ -56,7 +56,7 @@
 	acid_damage = damage
 	xeno_owner = _xeno_owner
 	RegisterSignal(xeno_owner, COMSIG_PARENT_QDELETING, .proc/clean_mob_owner)
-	TIMER_COOLDOWN_START(src, COOLDOWN_PARALYSE_ACID, 5)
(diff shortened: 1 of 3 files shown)


code/game/objects/effects/aliens.dm
```diff
@@ -56,7 +56,7 @@
 	acid_damage = damage
 	xeno_owner = _xeno_owner
 	RegisterSignal(xeno_owner, COMSIG_PARENT_QDELETING, .proc/clean_mob_owner)
-	TIMER_COOLDOWN_START(src, COOLDOWN_PARALYSE_ACID, 5)
+	TIMER_COOLDOWN_START(src, COOLDOWN_PARALYSE_ACID, 10)
```
TiviPlus, line 59:
```suggestion
	TIMER_COOLDOWN_START(src, COOLDOWN_PARALYSE_ACID, 1 SECONDS)
```

TiviPlus (approved):

PR #6497, opened 2021-04-04
Simplifies Vali

Walarks:
## About The Pull Request
Removes reagent modification from Vali. Replaces it with brute/burn heal and stamina regen.
- Restores 6 or 12 brute and burn based on effect
- Stamina regen happens if you have a connected weapon and scales inversely with active time, from 7 and becoming 0 past 20 seconds.

Changes the reagent storage function to store and inject only 30u so that people don't accidentally overdose.
Connection a weapon takes 1 second instead of 2.

Now the suit checks for germ level rather than necrotized limb flag when deciding on what limb to increase the germ amount pf past 20s. This is due to the necrotized flag not doing damage or other effects associated with high germ levels and thus allowing a way to circumvent the system in specific cases. It shouldn't change almost anything in practice.

## Why It's Good For The Game
Makes Vali more foolproof so that I could add more mechanics without it feeling too complicated.

## Changelog
:cl:
balance: vali now heals 6/12 brute/burn instead of boosting reagents and provides stamina regen if you have a connected weapon. Reduced reagent storage from 60u to 30u. Weapon takes 1s to connect instead of 2s. 
fix: vali checks for germ level instead of necrotized limb flag since the negatives associated with necrotized limbs are based on germ levels
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -28,12 +28,12 @@
 	var/obj/item/healthanalyzer/integrated/analyzer
 	///Determines whether the suit is on
 	var/boost_on = FALSE
-	///Stores the value with which effect_mult is modified
+	///Stores the current effect strength
 	var/boost_amount = 0
 	///Normal boost
 	var/boost_tier1 = 1
 	///Overcharged boost
-	var/boost_tier2 = 3
+	var/boost_tier2 = 2
 	///Boost icon, image cycles between 2 states
 	var/boost_icon = "cboost_t1"
 	///Item connected to the system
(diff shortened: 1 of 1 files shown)


DerFlammenwerfer:
Isn't making the stamina regen negative defeating the purpose of keeping your weapon connected? If you want to encourage pure Vali instead of Vali+gun, you cannot make it debuff you for having a connected weapon. It being a melee weapon is already enough of an inherent downside.

Walarks:
> Isn't making the stamina regen negative defeating the purpose of keeping your weapon connected? If you want to encourage pure Vali instead of Vali+gun, you cannot make it debuff you for having a connected weapon. It being a melee weapon is already enough of an inherent downside. I'd suggest instead having it build up slowly over time to a hard limit, starting at no boost.

Removed 20s negative stamina regen. Could add a different build-up mechanic at a later time.

Swagile:
vali being good with a ranged weapon is also a symptom of overbalance on pistols / revolers than an issue with pistol / revolver vali setup

Walarks:
> vali being good with a ranged weapon is also a symptom of overbalance on pistols / revolvers than an issue with pistol / revolver vali setup

This is an issue with melee being exceptionally bad to hit with. I've not made any pure balance changes to vali due to not being able to just hit xenos with it reliably enough to get going and see it happen more often than not when someone uses it.

TiviPlus (approved):

Issue #6374, opened 2021-03-18
Admins are unable to select ERTs manually


lKiseki:
## Testmerges:
#5422 #6313 #6306 #6314 #6284

## Reproduction:
When an ERT is called, attempt to select one manually. It throws an error saying "admin tried to click on a href with bad authorization".

PR #6498, opened 2021-04-04
ERT call are dependant on monitor state

BraveMole:
## About The Pull Request

Probability of each ERT will now be dependant on the current state of the game. If xeno are "losing"for example (aka, the ratio of marines to xenos is in their defavor), XRT will be favorised but PMC and freelancer will be defavorised.

Here are the weights for when the ratio is 3 marine for 1 xenos shipside (I'm considering that an ideal ratio is 2:1)

image

Opposite situation, ratio 1.4:1:

image

These values are exemple, i just changed the value of PMC and freelancers so their prob stay more stable

Also allow admins to deny an ERT without annoncing if they want so, this fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/6374 in a strange way, as you can deny without telling and then using the fun verb distress beacon (that do works)

## Why It's Good For The Game

Make ert less RNG dependant

## Changelog
:cl:
balance: probability of each ERT depends on the state of the game. IE, if xenos are losing shipside there is more chance than an XRT is called rather than a PMC one
balance: Each ERT can be called only once (unless admin interaction)
admin: Admin can deny an ERT without annoncing if they want. They can then use distress beacon verb to send another ERT if they wish to
/:cl:


code/datums/emergency_calls/emergency_call.dm
```diff
@@ -21,6 +20,14 @@
 	var/candidate_timer
 	var/cooldown_timer
 	var/spawn_type = /mob/living/carbon/human
+	///The base probability of that ERT spawning, it is changing with monitor state
+	var/base_probability = 0
+	/**How the current_weight change with the monitor state. A big positive number will make the current weight go down drasticly when marines are winning
```
TiviPlus, line 25: Can you start the actual comments on the next line for the sake of formatting please

code/datums/emergency_calls/emergency_call.dm
```diff
@@ -39,23 +46,28 @@
...
-
-	if(!istype(chosen_call))
-		chosen_call = pick(valid_calls)
-
-	return chosen_call
+		calls_weighted[E] = E.get_actualised_weight(normalised_monitor_state)
+
+	return pickweight(calls_weighted)
+
+/** Return a new current_weight using the base probability, the allignement factor of the ERT and the monitor state
```
TiviPlus, line 58: ditto

code/datums/emergency_calls/emergency_call.dm
```diff
@@ -39,23 +46,28 @@
...
+ */
+/datum/emergency_call/proc/get_actualised_weight(monitor_state)
+	var/probability_direction = (monitor_state * allignement_factor)
+	var/actualised_weight = base_probability
+	if(probability_direction >= 0)
+		actualised_weight *= (1+probability_direction)
+	else
+		actualised_weight /= (1-probability_direction)
+	message_admins("[name] has a weight of [actualised_weight]")
+	return actualised_weight
```
TiviPlus, line 70: You can directly return here, debug message admins, actualized weight isnt needed

TiviPlus (changes requested):

code/datums/emergency_calls/emergency_call.dm
BraveMole: yeah i know, i'm still testing things up

TiviPlus:
Also for those wondering this also fixes a bug which caused probability to not work properly

bowlofcereal:
yeah this does make sense to have im genuinely surprised this wasnt even a thing beforehand, should depend on who's got the upper hand

Swagile:
Does this make it so you can send ERT at ANY time as long as your on red alert or only during hijack situations?

BraveMole:
> Does this make it so you can send ERT at ANY time as long as your on red alert or only during hijack situations?

Nop, doesn't touch it

code/game/objects/machinery/computer/communications.dm
```diff
@@ -218,35 +218,27 @@
...
-						continue
+				var/datum/emergency_call/E = SSticker.mode.get_random_call()
 
-					valid_calls += list(E.name = E)
-
-				var/admin_response = admin_approval("<span color='prefix'>DISTRESS:</span> [ADMIN_TPMONTY(usr)] has called a Distress Beacon. Humans: [AllMarines], Xenos: [AllXenos].",
-					options = valid_calls, default_option = "random",
+				var/admin_response = admin_approval("<span color='prefix'>DISTRESS:</span> [ADMIN_TPMONTY(usr)] has called a Distress Beacon that was received by [E.name]. Humans: [AllMarines], Xenos: [AllXenos].",
 					user_message = "<span class='boldnotice'>A distress beacon will launch in 60 seconds unless High Command responds otherwise.</span>",
+					options = list("approve" = "approve", "deny" = "deny", "deny without annoncing" = "deny without annoncing"),
```
TiviPlus, line 225: You should use defines for this but since its past behaviour I wont force you

TiviPlus (changes requested):

TiviPlus (approved):

PR #6504, opened 2021-04-05
Xeno abilities easier to land on targets

BraveMole:
## About The Pull Request

Add a lot of target flags on xeno abilities, so it's easier to hit the desired target. Hitting the turf where the target is will automaticly redirect your ability on the target

Remove unused targeting flags

## Why It's Good For The Game

With lag, byond being bad, high TD, you sometimes have to spam your abilities to manage to hit you target. This makes it easier to hit those, which is nice QOL

## Changelog
:cl:
balance: Hitting the turf where the target is will automaticly redirect your ability on the target. This works for most xeno ability
code: Remove unused targeting flags
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -318,8 +318,6 @@ if(selected_ability.target_flags & flagname){\
 	TARGET_FLAGS_MACRO(XABB_HUMAN_TARGET, /mob/living/carbon/human)
 	TARGET_FLAGS_MACRO(XABB_XENO_TARGET, /mob/living/carbon/xenomorph)
 	TARGET_FLAGS_MACRO(XABB_MOB_TARGET, /mob/living)
-	TARGET_FLAGS_MACRO(XABB_OBJ_TARGET, /obj)
-	TARGET_FLAGS_MACRO(XABB_WALL_TARGET, /turf/closed/wall)
 	if(selected_ability.target_flags & XABB_TURF_TARGET)
 		return get_turf(A)
 	return A
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -315,8 +315,6 @@ if(selected_ability.target_flags & flagname){\
 		return;}}
 
 /mob/living/carbon/xenomorph/proc/ability_target(atom/A)
-	TARGET_FLAGS_MACRO(XABB_HUMAN_TARGET, /mob/living/carbon/human)
-	TARGET_FLAGS_MACRO(XABB_XENO_TARGET, /mob/living/carbon/xenomorph)
 	TARGET_FLAGS_MACRO(XABB_MOB_TARGET, /mob/living)
 	if(selected_ability.target_flags & XABB_TURF_TARGET)
 		return get_turf(A)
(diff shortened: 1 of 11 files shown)


code/modules/mob/living/carbon/xenomorph/castes/bull/abilities_bull.dm
```diff
@@ -8,6 +8,7 @@
 	ability_name = "plow charge"
 	keybind_signal = COMSIG_XENOABILITY_BULLCHARGE
 	var/new_charge_type = CHARGE_BULL
+	target_flags = XABB_HUMAN_TARGET
```
SplinterGP, line 11: ?????????

code/modules/mob/living/carbon/xenomorph/castes/bull/abilities_bull.dm
```diff
@@ -29,6 +30,7 @@
 	ability_name = "headbutt charge"
 	keybind_signal = COMSIG_XENOABILITY_BULLHEADBUTT
 	new_charge_type = CHARGE_BULL_HEADBUTT
+	target_flags = XABB_MOB_TARGET
```
SplinterGP, line 33: ??????????????

SplinterGP (changes requested):

SplinterGP:
these two skills are not targetted

BraveMole:
> these two skills are not targetted

they are not? I didn't play bull in a very long time

code/modules/mob/living/carbon/xenomorph/castes/bull/abilities_bull.dm
BraveMole, line 33:
```suggestion
```

code/modules/mob/living/carbon/xenomorph/castes/bull/abilities_bull.dm
BraveMole, line 11:
```suggestion
```

code/_onclick/click.dm
```diff
@@ -315,6 +315,8 @@ if(selected_ability.target_flags & flagname){\
 		return;}}
 
 /mob/living/carbon/xenomorph/proc/ability_target(atom/A)
+	TARGET_FLAGS_MACRO(XABB_HUMAN_TARGET, /mob/living/carbon/human)
+	TARGET_FLAGS_MACRO(XABB_XENO_TARGET, /mob/living/carbon/xenomorph)
```
TiviPlus, line 319: Why this can just use XABB_MOB_TARGET

BraveMole: It's not impossible than an incoucious human or xeno are on the same turf than a consiouc mob that you really want to target.

This probably happens rarely, but it's possible. Why not preventing that issue at least partly? Some abilities cannot target a human, let's try to find only xenos then

TiviPlus: Because it adds extra processing on every single ability use for an edge case that will basically never happen

TiviPlus (approved):

PR #6512, opened 2021-04-06
Fix ai order runtiming + give feedback when launching orders

BraveMole:
## About The Pull Request

Ai orders could runtime because i'm stupid. 
Also, when givind an order, you have feedback, so even when it runtime (and it will, cause i've never fixed it), you know you launched the order

## Changelog
:cl:
fix: Ai order no longer runtime that much
qol: When sending an order as CIC or AI, you have feedback on what you've done
/:cl:


TiviPlus (approved):

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -100,8 +100,10 @@
 	SIGNAL_HANDLER
 	if(!current_order)
 		to_chat(src, "<span class='warning'>Your have no order selected.</span>")
+		return
 	current_order.send_order(target)
 
+
```
BraveMole, line 106:
```suggestion
```

TiviPlus (approved):

PR #6516, opened 2021-04-06
Ravage and Shrike Screech can be selected with a keybind

BraveMole:
## About The Pull Request

You can now select these abilities with two new keybinds, and then use shift/middle/right click to proc them.
Doesn't change the old keybinds, you can still use them to immediatly proc the abilities

## Why It's Good For The Game

I've always found very strange that these abilities, that can be targeted with a click, can be immediatly used with a keybind like stomp or ignore pain.

I'm not removing that, because some players may prefer it, but i'm adding a new way to use those two abilities. I'll personnaly use it, because i'm missing two many ravage and screech because of the current system

## Changelog
:cl:
qol: Ravage and Shrike Screech can be selected with a new keybind, and then used with shift/middle/right click
/:cl:

### DIFF
--- a/code/datums/actions/xeno_action.dm
+++ b/code/datums/actions/xeno_action.dm
@@ -190,13 +190,12 @@
 	return ..()
 
 /datum/action/xeno_action/activable/action_activate()
-	SIGNAL_HANDLER
 	var/mob/living/carbon/xenomorph/X = owner
 	if(X.selected_ability == src)
 		return
 	if(X.selected_ability)
-		INVOKE_ASYNC(X.selected_ability, .proc/deselect)
-	INVOKE_ASYNC(src, .proc/select)
+		X.selected_ability.deselect()
+	select()
 
 /datum/action/xeno_action/activable/keybind_activation()
 	. = COMSIG_KB_ACTIVATED
--- a/code/datums/actions/xeno_action.dm
+++ b/code/datums/actions/xeno_action.dm
@@ -195,8 +195,8 @@
 	if(X.selected_ability == src)
 		return
 	if(X.selected_ability)
-		X.selected_ability.deselect()
-	select()
+		INVOKE_ASYNC(X.selected_ability, .proc/deselect)
+	INVOKE_ASYNC(src, .proc/select)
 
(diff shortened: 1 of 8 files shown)


code/datums/actions/xeno_action.dm
```diff
@@ -194,6 +212,7 @@
 	if(can_use_action(FALSE, NONE, TRUE)) // just for selecting
 		action_activate()
 
+
```
BraveMole, line 215:
```suggestion
```

code/modules/mob/living/carbon/xenomorph/castes/bull/abilities_bull.dm
```diff
@@ -10,15 +10,6 @@
 	var/new_charge_type = CHARGE_BULL
 
 
-/datum/action/xeno_action/activable/action_activate()
```
BraveMole, line 13: I moved this cause it was at a very random place

Mannybrado:
epic. dont know why it was changed in the first place; they still depend on where you're facing, not where your mouse is

code/datums/actions/xeno_action.dm
```diff
@@ -172,17 +172,35 @@
 
 
 /datum/action/xeno_action/activable
+	///Alternative keybind signal, that will always select the ability, even ignoring keybind flag
+	var/alternate_keybind_signal
```
SplinterGP, line 176: better, make the keybind  signal var a list with keybinds with assoc of w hat is their behaviour, would work better i believe and you could add multiple keybinds on future that way

SplinterGP (changes requested):

code/datums/actions/xeno_action.dm
BraveMole: Completly out of the scope this PR.

I'm adding two alternates keybind, i'm not refactoring the whole shit.

"you could add multiple keybinds on future that way" I won't, you won't, no one will. And if anyone will, he'll do it himself

SplinterGP:
reviiew this pls

TiviPlus (approved):

PR #6518, opened 2021-04-06
Taser

TGCMbetcher:
## About The Pull Request

Gives the taser non leathal damage and instead makes it a knockdown and stagger weapon. Can be used on xenos like a slug shotgun, but I wouldnt recommend it because it does NO damage and only fires for non combat classes other then FC and honestly the t39 slugs are 100% better in every way.

## Why It's Good For The Game

Tasers no longer are a threat that need to be faced with equal force, IE murdering the RO for trying to kill you.

## Changelog
:cl:
fix: Tasers wont murder people
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1177,13 +1177,13 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	damage = 10
 	penetration = 100
 	damage_type = STAMINA
-	flags_ammo_behavior = AMMO_ENERGY
+	flags_ammo_behavior = AMMO_ENERGY|AMMO_SKIPS_ALIENS
 	max_range = 15
 	accurate_range = 10
 	bullet_color = COLOR_VIVID_YELLOW
 	
 /datum/ammo/energy/taser/on_hit_mob(mob/M,obj/projectile/P)
-	staggerstun(M, P, weaken = 1, stagger = 13, slowdown = 1, knockback = 1)
+	staggerstun(M, P, weaken = 10)
 
 /datum/ammo/energy/tesla
 	name = "energy ball"
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1174,13 +1174,16 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	icon_state = "stun"
 	hud_state = "taser"
 	hud_state_empty = "battery_empty"
-	damage = 120
-	penetration = 0
+	damage = 10
+	penetration = 100
 	damage_type = STAMINA
 	flags_ammo_behavior = AMMO_ENERGY
 	max_range = 15
 	accurate_range = 10
 	bullet_color = COLOR_VIVID_YELLOW
+	
+/datum/ammo/energy/taser/on_hit_mob(mob/M,obj/projectile/P)
+	staggerstun(M, P, weaken = 1, stagger = 13, slowdown = 1, knockback = 1)
 
 /datum/ammo/energy/tesla
 	name = "energy ball"
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1183,7 +1183,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	bullet_color = COLOR_VIVID_YELLOW
 	
 /datum/ammo/energy/taser/on_hit_mob(mob/M,obj/projectile/P)
-	staggerstun(M, P, weaken = 10)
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/ammo_datums.dm
```diff
@@ -1174,13 +1174,16 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
...
+	damage = 10
+	penetration = 100
 	damage_type = STAMINA
 	flags_ammo_behavior = AMMO_ENERGY
 	max_range = 15
 	accurate_range = 10
 	bullet_color = COLOR_VIVID_YELLOW
+	
+/datum/ammo/energy/taser/on_hit_mob(mob/M,obj/projectile/P)
+	staggerstun(M, P, weaken = 1, stagger = 13, slowdown = 1, knockback = 1)
```
Chaznoodles, line 1186: Stagger 13?

TGCMbetcher:
Isnt that the state where marines are barely able to do something? I cant test it without someone to shoot me but Ive been told it halves damage and assuming that 13 is the number of seconds then fine, if its the same as stagger 1 then it doesnt matter. I could only test these things on an empty body. I was considering lowering the fire delay and cost per shot but the it could be used against xenos very effectively so I didnt. Again if the xenos thing becomes an issue I can just do the ifxeno thing and set it all to a single stagger stack, these only effect small xenos that are trashed by electricity anyway.

This is only a ranged way to get close enough to flash people anyway. Its far better for that purpose then before and they wont kill people when doing so. Theyll have like 3 seconds to get in close to flash and drag them out.

code/modules/projectiles/ammo_datums.dm
TiviPlus, line 1186: Make it just flat out stun

TGCMbetcher: > Make it just flat out stun

Youre assuming Im competent, let me look at how the flash does things. This is literally a bandaid for something that ever has to be used to stop people from murdering people with it then going its  taser it cant kill people" multiple times to admins so they get away with it.

InterroLouis:
As far as I'm concerned we could just do away with tasers entirely, unless admins want to go to the trouble of just deleting them all for any HVH. We've gotten rid of MPs, there's always an admin on. If someone is doing something like breaking into requisitions while it's manned, in a manner which would involve you tasing them...what if you just ahelp?

TGCMbetcher:
> As far as I'm concerned we could just do away with tasers entirely, unless admins want to go to the trouble of just deleting them all for any HVH. We've gotten rid of MPs, there's always an admin on. If someone is doing something like breaking into requisitions while it's manned, in a manner which would involve you tasing them...what if you just ahelp?

Majority of the time the RO would be breaking rule 5 so of course they wouldnt get admins involved

SplinterGP:
give it stun, stagger doenst stun it only reduces marine damage by 50%

TGCMbetcher:
> give it stun, stagger doenst stun it only reduces marine damage by 50%

Ive tried like a dozen different things none of them translate over and compile. If I could I would, I even tried agony but none of the things I found worked for that. Best I could do is give its melee attacks the stun baton stuff so you shoot them then whack them when youre close.

InterroLouis:
If you want it to be something to use against humans only just give it the tag AMMO_SKIPS_ALIENS
Then it doesn't matter because you won't be able to hit xenos with it anyways?

TGCMbetcher:
> If you want it to be something to use against humans only just give it the tag AMMO_SKIPS_ALIENS
> Then it doesn't matter because you won't be able to hit xenos with it anyways?

I figured I would just use the ll ifxeno thing if it really mattered to remove the knockdown but they want stun and there is no ranged stun in the game, period. Everything uses proximity, tiles or melee attacks.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1174,13 +1174,16 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
...
+	penetration = 100
 	damage_type = STAMINA
-	flags_ammo_behavior = AMMO_ENERGY
+	flags_ammo_behavior = AMMO_ENERGY|AMMO_SKIPS_ALIENS
 	max_range = 15
 	accurate_range = 10
 	bullet_color = COLOR_VIVID_YELLOW
+	
+/datum/ammo/energy/taser/on_hit_mob(mob/M,obj/projectile/P)
+	staggerstun(M, P, weaken = 10)
```
TiviPlus, line 1186:
```suggestion
	staggerstun(M, P, stun = 10)
```

TiviPlus (approved):

DerrickChamberlain:
> As far as I'm concerned we could just do away with tasers entirely, unless admins want to go to the trouble of just deleting them all for any HVH. We've gotten rid of MPs, there's always an admin on. If someone is doing something like breaking into requisitions while it's manned, in a manner which would involve you tasing them...what if you just ahelp?

That is just no fun, love tazing unruly unga's I do not need to waste the time of admins to do it.

PR #6519, opened 2021-04-07
Allows reloading from pistol belt.

hyper2snyper:
## About The Pull Request

Completes one of these approved suggestions.
https://tgstation13.org/phpBB/viewtopic.php?f=65&t=20487

Allows an empty pistol to be reloaded from a pistol belt using right click so long as right click is disabled and there is no ammo within.

## Why It's Good For The Game

https://tgstation13.org/phpBB/viewtopic.php?f=65&t=20487
It was an approved suggestion.
I think it would make reloading these low capacity weapons far easier. 

## Changelog
:cl:
add: Allows an empty pistol to be reloaded from a pistol belt using right click so long as right click is disabled and there is no ammo within.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -635,10 +635,9 @@
 	for(var/obj/item/ammo_magazine/mag in contents) 
 		if(!istype(gun, mag.gun_type))
 			continue
-		else
-			gun.reload(user, mag)
-			orient2hud()
-			return
+		gun.reload(user, mag)
+		orient2hud()
+		return
 
 /obj/item/storage/belt/gun/pistol/examine(mob/user, distance, infix, suffix)
 	. = ..()
@@ -861,4 +860,3 @@
 	var/obj/item/weapon/gun/new_gun = new /obj/item/weapon/gun/shotgun/double/marine(src)
 	new /obj/item/ammo_magazine/shotgun(src)
 	new_gun.on_enter_storage(src)
-
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -557,6 +557,7 @@
 		/obj/item/weapon/gun/pistol,
 		/obj/item/ammo_magazine/pistol,
 	)
+	var/reload_on = FALSE
 
 /obj/item/storage/belt/gun/Destroy()
 	if(gun_underlay)
@@ -618,6 +619,37 @@
 				to_chat(usr, "<span class='warning'>[src] can't hold any more magazines.</span>")
 			return FALSE
 	return TRUE
+
+/obj/item/storage/belt/gun/attackby(obj/item/I, mob/user, params)
+	if(!istype(I, /obj/item/weapon/gun/pistol) || !reload_on)
+		return ..()
+	var/obj/item/weapon/gun/pistol/gun = I
+	if(gun.current_mag)
+		return ..()
+	for(var/obj/item/ammo_magazine/mag in contents) 
+		if(!istype(gun, mag.gun_type))
+			continue
+		else
+			gun.reload(user, mag)
+			orient2hud()
+			return
+
+/obj/item/storage/belt/gun/examine(mob/user, distance, infix, suffix)
+	. = ..()
+	to_chat(user, "<span class='notice'>Control-Click to toggle \'Tactical Reload Mode\'. While on, clicking on the belt with an empty pistol will perform a reload with the ammo inside.</span>")
+	if(!reload_on)
+		to_chat(user, "<span class='notice'>Reload mode is OFF</span>")
+	if(reload_on)
+		to_chat(user, "<span class='notice'>Reload mode is ON</span>")
+
+/obj/item/storage/belt/gun/CtrlClick(mob/user)
+	if(src.loc != user)
+		return ..()
+	reload_on = !reload_on
+	if(!reload_on)
+		to_chat(user, "<span class='notice'>You toggle the reload mode OFF.</span>")
+	if(reload_on)
+		to_chat(user, "<span class='notice'>You toggle the reload mode ON.</span>")
 
 /obj/item/storage/belt/gun/m4a3
 	name = "\improper M4A3 holster rig"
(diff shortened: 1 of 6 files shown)


code/game/objects/items/storage/belt.dm
```diff
@@ -619,6 +620,37 @@
...
+
+/obj/item/storage/belt/gun/examine(mob/user, distance, infix, suffix)
+	. = ..()
+	to_chat(user, "<span class='notice'>Control-Click to toggle \'Tactical Reload Mode\'. While on, clicking on the belt with an empty pistol will perform a reload with the ammo inside.</span>")
+	if(!reload_on)
+		to_chat(user, "<span class='notice'>Reload mode is OFF</span>")
+	if(reload_on)
+		to_chat(user, "<span class='notice'>Reload mode is ON</span>")
+
+/obj/item/storage/belt/gun/CtrlClick(mob/user)
```
BraveMole, line 645: Maybe say on the examine that control click is the wait to change the reload mode?

code/game/objects/items/storage/belt.dm
```diff
@@ -557,6 +557,7 @@
 		/obj/item/weapon/gun/pistol,
 		/obj/item/ammo_magazine/pistol,
 	)
+	var/reload_on = FALSE
```
BraveMole, line 560: autodoc

code/game/objects/items/storage/belt.dm
```diff
@@ -619,6 +620,37 @@
 			return FALSE
 	return TRUE
 
+/obj/item/storage/belt/gun/attackby(obj/item/I, mob/user, params)
+	if(!istype(I, /obj/item/weapon/gun/pistol) || !reload_on)
+		return ..()
+	var/obj/item/weapon/gun/pistol/gun = I
+	if(gun.current_mag)
+		return ..()
```
BraveMole, line 628: probably want to to return
And a to chat to say that it has already a mag would be nice.
Also, if it's tactictal reload, should it not remove the mag already in place?

code/game/objects/items/storage/belt.dm
```diff
@@ -619,6 +620,37 @@
...
+			gun.reload(user, mag)
+			orient2hud()
+			return
+
+/obj/item/storage/belt/gun/examine(mob/user, distance, infix, suffix)
+	. = ..()
+	to_chat(user, "<span class='notice'>Control-Click to toggle \'Tactical Reload Mode\'. While on, clicking on the belt with an empty pistol will perform a reload with the ammo inside.</span>")
+	if(!reload_on)
+		to_chat(user, "<span class='notice'>Reload mode is OFF</span>")
+	if(reload_on)
```
BraveMole, line 642: use an early return and remove the second if

code/game/objects/items/storage/belt.dm
```diff
@@ -619,6 +620,37 @@
...
+		to_chat(user, "<span class='notice'>Reload mode is ON</span>")
+
+/obj/item/storage/belt/gun/CtrlClick(mob/user)
+	if(src.loc != user)
```
BraveMole, line 646: don't src.loc, just do src

code/game/objects/items/storage/belt.dm
```diff
@@ -619,6 +620,37 @@
...
+		return ..()
+	reload_on = !reload_on
+	if(!reload_on)
+		to_chat(user, "<span class='notice'>You toggle the reload mode OFF.</span>")
+	if(reload_on)
```
BraveMole, line 651: early return

code/game/objects/items/storage/belt.dm
TiviPlus, line 642: "reload mode" seems like a large hassle, and should likely be a different type of click, such as right or ctrl click onto the belt with a gun

TiviPlus (changes requested):
Also @BraveMole use batch review pls

code/game/objects/items/storage/belt.dm
hyper2snyper: I suppose so, but the suggestion did say to have a 
"reload mode." But I could always mess around a little to see what I can do.

code/game/objects/items/storage/belt.dm
hyper2snyper: I did, in Examine()

code/game/objects/items/storage/belt.dm
hyper2snyper: ah right, tactical means theres a mag. Ill change that.

code/game/objects/items/storage/belt.dm
hyper2snyper: why? since ```src``` is the belt, and ```src.loc``` is the mob/tile where it is currently located. If I change it wouldn't ```src != user``` always equal true since ```/obj/item/storage/belt/gun``` doesn't ever equal ```/mob/user``` ?

code/game/objects/items/storage/belt.dm
```diff
@@ -619,15 +619,40 @@
...
+	var/obj/item/weapon/gun/pistol/gun = I
+	if(gun.current_mag)
+		return ..()
+	for(var/obj/item/ammo_magazine/mag in contents) 
+		if(!istype(gun, mag.gun_type))
+			continue
+		else
+			gun.reload(user, mag)
+			orient2hud()
+			return
```
TiviPlus, line 641:
```suggestion
		gun.reload(user, mag)
		orient2hud()
		return
```

code/game/objects/items/storage/belt.dm
```diff
@@ -619,15 +619,40 @@
...
+		return ..()
+	var/obj/item/weapon/gun/pistol/gun = I
+	if(gun.current_mag)
+		return ..()
+	for(var/obj/item/ammo_magazine/mag in contents) 
+		if(!istype(gun, mag.gun_type))
+			continue
+		else
+			gun.reload(user, mag)
+			orient2hud()
```
TiviPlus, line 640: Why is this needed?

TiviPlus (changes requested):

code/game/objects/items/storage/belt.dm
hyper2snyper: its not

code/game/objects/items/storage/belt.dm
hyper2snyper: its not, thanks

TiviPlus (approved):

Dolth:
Honestly I'd like to know who came up with this idea...

PR #6523, opened 2021-04-07
Gives light to miners if repaired.

Swagile:
## About The Pull Request

I used my understanding of combat floodlight code to do this, so it might not be THE BEST way to do this, but it works and when I tested it I got no errors.

As for what this does, it makes have no light when broken, small light when light repair, decent light when medium repaired, and 8 tiles of light when fully operational. This also simulates the intense mining that a miner does; your drilling into the earth and such, yada yada.

## Why It's Good For The Game

A buff to properly defended miners (at minimum caded, at max with turrets and cades), while a nerf to stealth mining (the light is 1 tile extra of a floodlight; a single xeno patrol for 30 seconds to 1 min is now enough to spot stealth miners since they give off huge light).

## Changelog
:cl:
add: Miners now give off their own light; both in the repairing process and in the final running phase.
balance: This buffs properly defended miners.
Balance: However, stealth mining is now nerfed, since fully operational miners give off huge light, which is easy to spot on most maps.
/:cl:

### DIFF
--- a/code/game/objects/machinery/miner.dm
+++ b/code/game/objects/machinery/miner.dm
@@ -2,10 +2,6 @@
 #define MINER_SMALL_DAMAGE	1
 #define MINER_MEDIUM_DAMAGE	2
 #define MINER_DESTROYED	3
-#define MINER_LIGHT_RUNNING 8
-#define MINER_LIGHT_SDAMAGE 4
-#define MINER_LIGHT_MDAMAGE 2
-#define MINER_LIGHT_DESTROYED 0
 #define MINER_AUTOMATED "mining computer"
 #define MINER_RESISTANT	"reinforced components"
 #define MINER_OVERCLOCKED "high-efficiency drill"
@@ -36,6 +32,12 @@
 	var/max_miner_integrity = 100
 	///What type of upgrade it has installed , used to change the icon of the miner.
 	var/miner_upgrade_type
+	//Brightness of the miner
+	var/brightness_on = 0
+
+/obj/machinery/miner/Initialize()
+	. = ..()
+	set_light(brightness_on)
 
 /obj/machinery/miner/damaged	//mapping and all that shebang
 	miner_status = MINER_DESTROYED
@@ -54,16 +56,16 @@
 	switch(miner_status)
 		if(MINER_RUNNING)
 			icon_state = "mining_drill_active_[miner_upgrade_type]"
-			set_light(MINER_LIGHT_RUNNING)
+			set_light(8)
 		if(MINER_SMALL_DAMAGE)
 			icon_state = "mining_drill_braced_[miner_upgrade_type]"
-			set_light(MINER_LIGHT_SDAMAGE)
+			set_light(4)
 		if(MINER_MEDIUM_DAMAGE)
 			icon_state = "mining_drill_[miner_upgrade_type]"
-			set_light(MINER_LIGHT_MDAMAGE)
+			set_light(2)
 		if(MINER_DESTROYED)
 			icon_state = "mining_drill_error_[miner_upgrade_type]"
-			set_light(MINER_LIGHT_DESTROYED)
+			set_light(brightness_on)
 /// Called whenever someone attacks the miner with a object which is considered a upgrade.The object needs to have a uptype var.
 /obj/machinery/miner/proc/attempt_upgrade(obj/item/minerupgrade/upgrade, mob/user, params)
 	if(miner_upgrade_type)
--- a/code/game/objects/machinery/miner.dm
+++ b/code/game/objects/machinery/miner.dm
@@ -2,6 +2,10 @@
 #define MINER_SMALL_DAMAGE	1
 #define MINER_MEDIUM_DAMAGE	2
 #define MINER_DESTROYED	3
+#define MINER_LIGHT_RUNNING 8
+#define MINER_LIGHT_SDAMAGE 4
+#define MINER_LIGHT_MDAMAGE 2
+#define MINER_LIGHT_DESTROYED 0
 #define MINER_AUTOMATED "mining computer"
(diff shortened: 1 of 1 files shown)


TGCMbetcher:
Xeno buff, kills stealth mining

Swagile:
> 
> 
> Xeno buff, kills stealth mining BOOOO

yes, it is in the changelog

suffer

code/game/objects/machinery/miner.dm
```diff
@@ -32,6 +32,12 @@
 	var/max_miner_integrity = 100
 	///What type of upgrade it has installed , used to change the icon of the miner.
 	var/miner_upgrade_type
+	//Brightness of the miner
+	var/brightness_on = 0
```
TiviPlus, line 36: this var is unneeded

code/game/objects/machinery/miner.dm
```diff
@@ -32,6 +32,12 @@
...
+
+/obj/machinery/miner/Initialize()
+	. = ..()
+	set_light(brightness_on)
```
TiviPlus, line 40: Dead code

code/game/objects/machinery/miner.dm
```diff
@@ -50,12 +56,16 @@
 	switch(miner_status)
 		if(MINER_RUNNING)
 			icon_state = "mining_drill_active_[miner_upgrade_type]"
+			set_light(8)
 		if(MINER_SMALL_DAMAGE)
 			icon_state = "mining_drill_braced_[miner_upgrade_type]"
+			set_light(4)
 		if(MINER_MEDIUM_DAMAGE)
 			icon_state = "mining_drill_[miner_upgrade_type]"
+			set_light(2)
```
TiviPlus, line 65: These magic numbers should use defines

TiviPlus (changes requested):

Swagile:
my source tree was acting up, hence why the multiple commits 😔

EvilMelonTheSecond:
Can we fix proper mining before nerfing stealth mining? Lights are cool but does nothing for their defense unless marines are already defending it. Getting new miners means trying to fortify further from the base for each one and xeno can just walk around and hit the least defended one anyways, this doesn't fix the core of the problem and shouldn't be implemented until that's fixed in my opinion.

Swagile:
> 
> 
> Can we fix proper mining before nerfing stealth mining? Lights are cool but does nothing for their defense unless marines are already defending it. Getting new miners means trying to fortify further from the base for each one and xeno can just walk around and hit the least defended one anyways, this doesn't fix the core of the problem and shouldn't be implemented until that's fixed in my opinion.

wdym?

mining used to be even harder before; you HAD to cade a miner or sit in tadpole slowly extracting plat; you were a huge target, and then you needed to either bring it back to alamo or buy a ASRS pad (or more if you had multiple miners)

now miners need no ASRS pad and can be automated

ive also put in a PR to lower turret prices, so you can ask req to buy you a turret to help defend a miner (which dissuades 90% of miner breakage attempts)

if mining was made any easier than this, it would be OP as fuck

EvilMelonTheSecond:
> 
> 
> > Can we fix proper mining before nerfing stealth mining? Lights are cool but does nothing for their defense unless marines are already defending it. Getting new miners means trying to fortify further from the base for each one and xeno can just walk around and hit the least defended one anyways, this doesn't fix the core of the problem and shouldn't be implemented until that's fixed in my opinion.
> 
> wdym?
> 
> mining used to be even harder before; you HAD to cade a miner or sit in tadpole slowly extracting plat; you were a huge target, and then you needed to either bring it back to alamo or buy a ASRS pad (or more if you had multiple miners) AND you had to power the pad (which, outside some niches like Big Red, requires an APC nearby and to defend that APC as well)
> 
> now miners need no ASRS pad and can be automated
> 
> ive also put in a PR to lower turret prices, so you can ask req to buy you a turret to help defend a miner (which dissuades 90% of miner breakage attempts)
> 
> if mining was made any easier than this, it would be OP as fuck

I agree with most of your points except the defence of the miner. The reason stealth mining makes so many points is the fact that nobody has to defend the miner and it can still survive. Xeno have full map control everywhere except marine guarded areas and can easily spot and harrass frontline miners. If you need a 60 point turret to deter a T1 drone and get 15 points every two minutes in exchange, only for a wraith to blink in and slash the miner once or a crusher to charge twice to destroy both the turret and the miner or any acid spit caste just takes out a single cade and it falls apart then most engineers will not even bother with setting up miners let alone protect them while the rest of the team is away in some caves shooting xeno.

I rarely see engineers split off and get miners nowadays and this rework might change that very slightly but in the end I don't expect this to fix the core problem and it nerf stealth mining on top of that.

Make it toggelable maybe? I don't think it would be too much of a buff regardless.

SkeletalElite:
> If you need a 60 point turret to deter a T1 drone and get 15 points every two minutes in exchange, only for a wraith to blink in and slash the miner once

As the person behind many of the recent changes to make mining easier, destroying miners should probably be changed to a few second long do_after much like disarming the nuke in crash so wraiths can't blink in and destroy it before a turret is able to do anything to stop them.

code/game/objects/machinery/miner.dm
TiviPlus: Please do not resolve standing requests

code/game/objects/machinery/miner.dm
TiviPlus: Please do not resolve standing requests

TiviPlus (changes requested):

Swagile:
so basically

when i was trying to fix my source tree, i thought i committed the changes you asked, but evidently i didn't

which is why i resolved it

turns out i had a unga moment, should be fixed now

TiviPlus (approved):

PR #6527, opened 2021-04-07
mortar kit and shells are bulky rather than huge, engineer bag is able to hold them

Pariah919:
## About The Pull Request
image

## Why It's Good For The Game
makes frontlnie mortar more viable and adds more interesting options to the game

## Changelog
:cl:
balance: mortar kit and shells are bulky rather than huge, engineer bag is able to hold them
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -369,6 +369,8 @@
 		/obj/item/ammo_magazine/sentry,
 		/obj/item/ammo_magazine/minisentry,
 		/obj/item/marine_turret/mini,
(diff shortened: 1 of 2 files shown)


Chaznoodles:
Maybe make mortar fit on the back slot instead of in the bag? It looks and feels a little dopey that it's the same size as the things it fires.

hyper2snyper:
I would suggest having the ammo one level smaller too, that way you can carry ammo and actual engi gear/ weapons

psykzz:
no negative fix ratio? 👀

TiviPlus (approved):

PR #6539, opened 2021-04-09
bbuffs dmr and br

Pariah919:
## About The Pull Request
br gets 40 round magazines
dmr gets slightly faster wield delay and less aim mode slowdown

## Why It's Good For The Game
battle rifle is bad, it should've had double the mag capacity from the start
dmr is a bit bad, it could use some minor love

## Changelog
:cl: 
balance: dmr has slightly less slowdown, less aim slowdown, and battle rifle has had its capacity upped from 25 to 35
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/magazines/rifles.dm
@@ -104,7 +104,7 @@
 	icon_state = "t64"
 	w_class = WEIGHT_CLASS_NORMAL
 	default_ammo = /datum/ammo/bullet/rifle/standard_dmr
-	max_rounds = 25
+	max_rounds = 40
 	gun_type = /obj/item/weapon/gun/rifle/standard_br
 
 /obj/item/ammo_magazine/rifle/standard_br/incendiary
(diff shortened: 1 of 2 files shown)


Tupinambis:
The BR not having a reasonable magazine size was my main issue with it. If it could have a 3 round burst option that would be even better. Could even have 36 rounds to a mag.

code/modules/projectiles/magazines/rifles.dm
```diff
@@ -104,7 +104,7 @@
 	icon_state = "t64"
 	w_class = WEIGHT_CLASS_NORMAL
 	default_ammo = /datum/ammo/bullet/rifle/standard_dmr
-	max_rounds = 25
```
TiviPlus, line 107: 35

TiviPlus (changes requested):
Fix your CL

lKiseki:
Good stuff. These weapons needed some love. Should be given the test merge candidate flag so we can see it in action.

code/modules/projectiles/magazines/rifles.dm
```diff
@@ -104,7 +104,7 @@
...
+	max_rounds = 40
```
TiviPlus, line 107:
```suggestion
	max_rounds = 35
```

TiviPlus (approved):

PR #6553, opened 2021-04-10
Destroying a silo give points to marine

BraveMole:
## About The Pull Request

Give 150 supply points when killing a regular silo, 75 when it's a small one

## Why It's Good For The Game

Reward the marines for a successfull silo hunt, i know points are generally not exciting but this is better than nothing. This will give a better sentiment of accomplishment to marines

## Changelog
:cl:
balance: Give 150 supply points when killing a regular silo, 75 when it's a small one
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -129,7 +129,7 @@
 #define SILO_PRICE 1200
 #define XENO_TURRET_PRICE 150
 //Amount of points gave to marine when they destroy a silo
-#define SILO_REWARD 200
+#define SILO_REWARD 150
 
 //The minimum round time before siloless timer can start (13:00)
 #define MINIMUM_TIME_SILO_LESS_COLLAPSE 36000 
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -128,6 +128,8 @@
 
 #define SILO_PRICE 1200
 #define XENO_TURRET_PRICE 150
+//Amount of points gave to marine when they destroy a silo
+#define SILO_REWARD 200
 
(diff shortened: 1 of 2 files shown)


David-Stormwell:
Don't think this is needed at all, there's already so much of a reason to kill Silos, this just further rewards marines who blitz silos.
Let's say one single marine solo rushed a silo now, packed a flamer and nades.
A easy 100 or 200 points at the cost of one single marine.

BraveMole:
> Don't think this is needed at all, there's already so much of a reason to kill Silos, this just further rewards marines who blitz silos.
> Let's say one single marine solo rushed a silo now, packed a flamer and nades.
> A easy 100 or 200 points at the cost of one single marine.

If marine were able to solo rush silos consistently, there would not be any pool.
I see successfull silo snipe every 10 rounds now, and each time it costs 4-5 marines

SpaceLoveSs13:
200/100 seems a bit too much to me. Atleast 200 should be made 150 imo.

BraveMole:
> 200/100 seems a bit too much to me. Atleast 200 should be made 150 imo.

Eh, we can try that if you want

code/__DEFINES/mode.dm
```diff
@@ -128,6 +128,8 @@
 
 #define SILO_PRICE 1200
 #define XENO_TURRET_PRICE 150
+//Amount of points gave to marine when they destroy a silo
+#define SILO_REWARD 200
```
BraveMole, line 132:
```suggestion
#define SILO_REWARD 150
```

TiviPlus:
Theres enough reason to kill silos imo, this is a little over the top

PR #6473, opened 2021-04-02
Nerf psy drain

BraveMole:
## About The Pull Request

Psy drain will now give 40 psy points rather than 60. Amount of larva unchanged

## Why It's Good For The Game

Rebalance psy drain a bit. This is to make cocooning still competitive, while making xeno steamrolling less.

## Changelog
:cl:
balance: Psy drain will now give 40 psy points rather than 60. Amount of larva unchanged
/:cl:

PR #6554, opened 2021-04-10
Cocoon rebalance

BraveMole:
## About The Pull Request

Cocoon will give 80 points rather than 140, in 5 minutes
Also give 1.5 larva points if the cocoon reach maturity (5 minutes), or the alamo is hijacked. If a marine break the cocoon before, no larva points will be generated

Should not be merged without this https://github.com/tgstation/TerraGov-Marine-Corps/pull/6473

## Why It's Good For The Game

Right now, if a marine push manage to kill all xeno silo but are dying in the process, it's a net loss, because their bodies will contribute to contribute to create more silos than there were originally.

I beleive this kind of situation should lead to no side clearly leading. Losing a silo should be a bit more significant than it is currently is.

With this PR, marines dying BUT succeeding in their mission means that the game is still very on, rather than being something the xenos can take a net profit

## Changelog
:cl:
balance: Cocoon will give 80 points rather than 140 points in 5 minutes. Also gives 1/8th of a larva at the end of the timer, this won't happen if a marine break open the cocoon first
/:cl:

### DIFF
--- a/code/game/objects/items/cocoon.dm
+++ b/code/game/objects/items/cocoon.dm
@@ -16,7 +16,7 @@
 	///How much time the cocoon takes to deplete the life force of the marine
 	var/cocoon_life_time = 5 MINUTES
 	///How many psych points it is generating every 5 seconds (84 in total)
-	var/psych_points_output = 0.7
+	var/psych_points_output = 1.4
 	///How many larva points it gives when the timer is over
 	var/larva_points_reward = 1.5
 	///Standard busy check
--- a/code/game/objects/items/cocoon.dm
+++ b/code/game/objects/items/cocoon.dm
@@ -14,7 +14,7 @@
 	///What is inside the cocoon
 	var/mob/living/victim
 	///How much time the cocoon takes to deplete the life force of the marine
-	var/cocoon_life_time = 10 MINUTES
+	var/cocoon_life_time = 5 MINUTES
 	///How many psych points it is generating every 5 seconds (84 in total)
 	var/psych_points_output = 0.7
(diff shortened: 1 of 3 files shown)


lKiseki:
Taking into account that this exists alongside #6473, the difference between cocooning and psydraining wouldn't be that big, which means cocooning would lose value since psydraining has the advantage of not freezing the DNR timer. Either cocooning should have a bigger advantage, like taking less time to fully generate points, or psydraining should have a decreased reward so as to maintain cocooning's value.

code/game/objects/items/cocoon.dm
```diff
@@ -15,8 +15,10 @@
 	var/mob/living/victim
 	///How much time the cocoon takes to deplete the life force of the marine
 	var/cocoon_life_time = 10 MINUTES
```
BraveMole, line 17:
```suggestion
	var/cocoon_life_time = 5 MINUTES
```

code/game/objects/items/cocoon.dm
```diff
@@ -14,9 +14,11 @@
 	///What is inside the cocoon
 	var/mob/living/victim
 	///How much time the cocoon takes to deplete the life force of the marine
-	var/cocoon_life_time = 10 MINUTES
-	///How many psych points it is generating every 5 seconds
-	var/psych_points_output = 1.2
+	var/cocoon_life_time = 5 MINUTES
+	///How many psych points it is generating every 5 seconds (84 in total)
+	var/psych_points_output = 0.7
```
BraveMole, line 19:
```suggestion
	var/psych_points_output = 1.4
```

PR #6562, opened 2021-04-10
hunter nerf, speed and slash nerf

Pariah919:
## About The Pull Request
less slash

## Why It's Good For The Game
hunter shouldn't be doing this much damage

## Changelog
:cl:
balance: nerfs hunter slash by 3 points, weed speed from -0.4 to .1 and he's slightly slower off weeds too
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/hunter/castedatum_hunter.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hunter/castedatum_hunter.dm
@@ -11,7 +11,7 @@
 	wound_type = "hunter" //used to match appropriate wound overlays
 
 	// *** Melee Attacks *** //
-	melee_damage = 23
+	melee_damage = 20
 	attack_delay = 7
 
 	// *** Tackle *** //
@@ -91,7 +91,7 @@
 	upgrade = XENO_UPGRADE_TWO
 
 	// *** Melee Attacks *** //
-	melee_damage = 26
+	melee_damage = 23
 
 	// *** Tackle *** //
 	tackle_damage = 32
@@ -122,7 +122,7 @@
 	ancient_message = "We are the epitome of the hunter. Few can stand against us in open combat."
 
 	// *** Melee Attacks *** //
-	melee_damage = 26
+	melee_damage = 23
 
 	// *** Tackle *** //
 	tackle_damage = 32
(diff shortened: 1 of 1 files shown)


usnpeepoo:
even more xeno nerfs

David-Stormwell:
Fineish on Paper.
Still think hunter should be reverted to how it was before, problem is before bump slash didn't exist.

Naaanii:
Maybe nerf one slash but nerf movespeed on weeds as well.

David-Stormwell:
> 
> 
> Maybe nerf one slash but nerf movespeed on weeds as well.

No, if you nerf a caste too much it'll just be a dogshit caste no one uses much anymore.
You still want to have it in gameplay and used, just not as OP as it is.

TGCMbetcher:
I honestly believe hunters are in a good spot right now. They cant even 1v1 marines anymore. I fumbled with my slug shotgun for like 7 seconds, it broke my arm, and I still won a duel simply because it you hit it consistently it doesnt matter what youre using it cant take the damage.

SilencedMP5A5:
hunter fine tbh

Losenis:
I'd sooner take a look at what lets it deliver that damage rather than the damage itself, since it's essentially a rouny+. At least it'll alleviate the free hits in stun breaking your arm a bit. WASDing with -1.5 offweeds speed and bumpslash is ridiculous when a player that knows how to use it properly controls it, let alone two or three at the same time.

bobmancer343:
get good

Losenis:
>posted while getting free hits with stun

lKiseki:
Hunter is most definitely not fine. While nerfing the slash damage could be a solution, I personally believe nerfing its move speed would be a good change as well. It goes too fast, even off weeds, and it's very hard to punish a Hunter due to this. A Hunter can also WASD you to death, and other marines _normally_ won't shoot due to fear of committing friendly fire.
In summary, I believe nerfing its general move speed a little would be a good change as well.

Naaanii:
Decent changes.

Naaanii:
Maybe reduce the slash nerfs to 2 instead of 3.

TiviPlus (approved):

Al-1ce:
>Hunter can also WASD you to death, and other marines _normally_ won't shoot due to fear of committing friendly fire.
This is why aim-mode exists.
> In summary, I believe nerfing its general move speed a little would be a good change as well.
Slowing down what is essentially a slightly better runner in terms of HP makes evolving into Hunter less worth it. I think that reducing the bump-slash damage somehow would be a better fix, since most hunter players rely on that.

PR #6567, opened 2021-04-11
Lace nerf

David-Stormwell:
## About The Pull Request

Gives the Pistol Lace 10 recoil if one handed. disallowing duel Tp-14s via massive recoil.

## Why It's Good For The Game

Some have proven to me that this is needed.
It'll be about the same as if you had a PPSH or Sub-Machine gun, in which you might as well use one after this.

## Changelog
:cl:
balance: Pistol Lace now has 10 recoil if 1 handed.
/:cl:


David-Stormwell:
Need to revert something since this also enables charging once again.

BraveMole:
what about making it a rail attachement, rather than a barrel one?

RenGusta:
I don't know what you plan to solve with this other than making laces redundant in their most common use, in disarmable backup weapon/double pistol primary

David-Stormwell:
> 
> 
> I don't know what you plan to solve with this other than making laces redundant in their most common use, in disarmable backup weapon/double pistol primary

Yes I am making it redundant.
Pistols are a sidearm which you pull out and use for quick damage.
Sure it'll affect you if let's say a Hunter chopped your hand off, but oh well.
It's either that or we nerf the TP-14 and other such guns, and well even without the lace you can still do duel pistol memes, just respect the fact you won't be using a lace.
And as for making it a rail slot, that changes almost nothing, infact it buffs one handing harder since you may now put on a recoil comp or other attachments which decrease scatter, because the RDS only affects accuracy, not scatter or recoil.

David-Stormwell:
Fuck off linters

David-Stormwell:
Well this is annoying, don't you love it when you find out when you updated something to Master.
I'll be remaking this PR in a second in a fresh branch.

PR #6572, opened 2021-04-11
50HP buff to metal cades so they aren't helpless againsts crushers

EvilMelonTheSecond:
## About The Pull Request
Increased metal barricade base health to 250 making it 300 when wired

## Why It's Good For The Game
Metal barricades are the standard barricade option for engineers. Metal is more expensive than sandbags(which have higher health) and metal is a mandatory material since other cades need barbed wire. Despite being upgradable the 200HP makes it always worse than sandbags in every scenario. The main point of metal cades is that they are repairable for free but the low HP makes them drop to "unrepairable" health with a single crusher charge and since frontline cades are mostly a single cadeline that makes the entire cadeline fall.

This PR is made assuming the current state of meta assures that when a marine push fails/wipes the FOB will not hold the xeno wave since they will likely control most of the map and have extra pools up which will make them outnumber marines and win unless a serious comeback happens. As such I'm hoping that FOB "autism" forts will not be a problem and the overall impact of frontline cades increases without making it too strong.

## Changelog
:cl:
balance: Increased metal barricade health points by 50
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -420,7 +420,7 @@
 	name = "metal barricade"
 	desc = "A sturdy and easily assembled barricade made of metal plates, often used for quick fortifications. Use a blowtorch to repair."
 	icon_state = "metal_0"
-	max_integrity = 200
+	max_integrity = 250
 	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
 	coverage = 128
 	stack_type = /obj/item/stack/sheet/metal


BraveMole:
Engie can make razor wire way quicker now, let them get used to that before buffing engie again i think

EvilMelonTheSecond:
> 
> 
> Engie can make razor wire way quicker now, let them get used to that before buffing engie again i think

Razorwire gets instantly destroyed if it has any damage and doesn't even slow down the charge. I just want metal cades to not get fully countered by a single caste. A crusher still destroys a metal cade with two charges but with this an engi can save it if they aren't interrupted between charges unlike now where the cade needs to use 2 metal to repair just from a single charge.

lKiseki:
I do not believe metal cades require this sort of buff. Engineers can already build more cades to cover up patches, and rather quickly at that. There's also multiple ways to punish Crushers if they target one of your cades, namely Razorwire, and given it can be built faster by engineers now... yeah.

EvilMelonTheSecond:
> 
> 
> I do not believe metal cades require this sort of buff. Engineers can already build more cades to cover up patches, and rather quickly at that. There's also multiple ways to punish Crushers if they target one of your cades, namely Razorwire, and given it can be built faster by engineers now... yeah.

This is more about giving metal cades their niche. Metal is fully outmatched by sandbag cades and the saving grace is supposed to be the free repair but you can't do that if a crusher always puts it in an unrepairable state. As an engineer that makes me almost always use sandbags with their lower cost and faster build time and only times I build metal cades is when I'm out of sandbags and desperate to plug holes.

Metal cades feel left out with their pitiful HP and I think even a 100 HP buff to make them on par with sandbags would be reasonable since it has double the price but I want a smaller scale solution to make them viable without making scary buffs that look like I'm trying to get cadehug meta back.

Georgeur:
Let's get baseline about this pr. 
25% in integrity, so that engis have enough time to travel to a metal cade before its fully fucked. 
I actually agree with this. Copypasta level code, but a good decision. 

Additionally, if you think razorwire protects cades, you've never met a spitter than can 1 shot them, and are therefor bald. Have fun.

TiviPlus:
I'd rather the concussive cades be better against crushers as a counter

PR #6581, opened 2021-04-13
No Siloless collapse until 12:33

InterroLouis:
## About The Pull Request

Changes the minimum time for siloless collapse from 13:03 as it is now, to 12:33 instead. 

## Why It's Good For The Game

A fairly popular strategy for hives is to not place pools until absolutely necessary, so as to deny marines the existence of their objective.  There isn't a lot marines can do against this strategy except maybe just go maximum cadehug, and that's just not fun for anyone. There will, however, still be time to actually place them, instead of having to rush to do it. 

## Changelog
:cl:
balance: No more siloless collapse before 12:33! If you didn't put a silo before that time, the timer will show up even if you never lost any silos. 
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -129,5 +129,5 @@
 #define SILO_PRICE 1200
 #define XENO_TURRET_PRICE 150
 
-//The minimum round time before siloless timer can start (13:00)
-#define MINIMUM_TIME_SILO_LESS_COLLAPSE 36000 
+//The minimum round time before siloless timer can start (12:33)
+#define MINIMUM_TIME_SILO_LESS_COLLAPSE 18000 


Georgeur:
Good PR
would prefer 40 minutes to deal with bald queens
but great pr

Naaanii:
Not placing a silo already stops Xenos from getting burrowed. Maybe just nerf psy drain instead.

BraveMole:
Maybe 12:40 or whatever? On some maps you want to be sure that it's not a tadpole OP to chose your silo location.

EvilMelonTheSecond:
12:40 sounds better. 

Surely the ungatide can wait for a couple minutes(They cannot. They will charge face first into a resin maze and never learn)

InterroLouis:
We'll just have to see how it goes. If this leads to an excessive amount of marine wins due to immediate silo rushes, I might just go ahead and change it to 12:40. As is, I'd like to see it TM'd before I alter the numbers.

code/__DEFINES/mode.dm
```diff
@@ -129,5 +129,5 @@
 #define SILO_PRICE 1200
 #define XENO_TURRET_PRICE 150
 
-//The minimum round time before siloless timer can start (13:00)
-#define MINIMUM_TIME_SILO_LESS_COLLAPSE 36000 
+//The minimum round time before siloless timer can start (12:33)
+#define MINIMUM_TIME_SILO_LESS_COLLAPSE 18000
```
BraveMole, line 133:
```suggestion
#define MINIMUM_TIME_SILO_LESS_COLLAPSE 30 MINUTES
```

BraveMole (changes requested):

davidchan13:
If marines didn't get silo pointers doe free this would be a good pr. As it stands xenos can't properly counter a tadop because marines magically know how many and where silos are while xenos get 0 intel on marines they can't see outside of the inconsistent bioscans.

TiviPlus:
This encourages silo/tad rushing even more, as well as limiting the usefulness of no-silo hive attempts so I dont see a particular use in this

PR #6591, opened 2021-04-14
Splints require additional hits to break based on medical skill

Walarks:
## About The Pull Request
Gives a splint health variable to limbs. Splint health gets decreased every time a splinted limb takes damage by the amount of damage and the splint gets removed at 0 health. Applied splint health is 15 per medical skill rank. If you splint yourself, you get 15 less splint health, but the amount can't be below 0.

## Why It's Good For The Game
As it stands right now valk's autosplint feature makes it a bit too convenient over the other modules, this should bring up the bottom line, while valk still has the benefits of instant and infinite splints.

## Changelog
:cl:
balance: splints require taking a certain amount of damage to break, based on medical skill of the one applying it. Anyone with medical skill adds 15 splint health per medical skill rank. If you splint yourself, this is reduced by 15, but can't be lower than 0.
balance: removes small chance for splints to not be damaged
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -278,6 +278,7 @@
 	icon_state = "splint"
 	amount = 5
 	max_amount = 5
+	var/applied_splint_stacks = 1
 
 
 /obj/item/stack/medical/splint/attack(mob/living/carbon/M, mob/user)
@@ -314,5 +315,5 @@
 			user.visible_message("<span class='warning'>[user] starts to apply [src] to their [limb].</span>",
 			"<span class='notice'>You start to apply [src] to your [limb], hold still.</span>")
 
-		if(affecting.apply_splints(src, user, M)) // Referenced in external organ helpers.
+		if(affecting.apply_splints(src, applied_splint_stacks, user, M)) // Referenced in external organ helpers.
 			use(1)
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -315,5 +315,5 @@
 			user.visible_message("<span class='warning'>[user] starts to apply [src] to their [limb].</span>",
 			"<span class='notice'>You start to apply [src] to your [limb], hold still.</span>")
 
-		if(affecting.apply_splints(src, applied_splint_stacks, user, M)) // Referenced in external organ helpers.
+		if(affecting.apply_splints(src, user == M ? (applied_splint_stacks*user.skills.getRating("medical") - 1) : applied_splint_stacks*user.skills.getRating("medical"), user, M)) // Referenced in external organ helpers.
 			use(1)
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -278,7 +278,8 @@
 	icon_state = "splint"
 	amount = 5
 	max_amount = 5
-	var/applied_splint_stacks = 1
+	///How much splint health per medical skill is applied
+	var/applied_splint_health = 15
 
 
 /obj/item/stack/medical/splint/attack(mob/living/carbon/M, mob/user)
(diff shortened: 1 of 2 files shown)


Naaanii:
Good enough. Maybe up it to 3 to take into account FF and the lack of stabilization too.

Swagile:
this is a better PR than mine mechanically and code wise, so if this gets in ill just keep splint stack and scanner speeds on my own PR

Losenis:
Unsure about this one since it'll only lead to marines self-splinting and then later walking up to a corpsman to tell them to unsplint and splint again. Would also mean that people with nothing better to do in the round would start asking corpsmen to do it again to the same limb after a fight to replenish the stacks for the next fight.

code/game/objects/items/stacks/medical.dm
```diff
@@ -278,6 +278,7 @@
 	icon_state = "splint"
 	amount = 5
 	max_amount = 5
+	var/applied_splint_stacks = 1
```
TiviPlus, line 281: Autodoc

code/modules/organs/limbs.dm
```diff
@@ -338,9 +340,12 @@
...
 
-	if(limb_status & LIMB_SPLINTED && damage > 5 && prob(50 + damage * 2.5)) //If they have it splinted, the splint won't hold.
-		remove_limb_flags(LIMB_SPLINTED)
-		to_chat(owner, "<span class='userdanger'>The splint on your [display_name] comes apart!</span>")
+	if(limb_status & LIMB_SPLINTED && damage > 10) //If they have it splinted and no splint stacks, the splint won't hold.
+		if(splint_stacks <= 0)
+			remove_limb_flags(LIMB_SPLINTED)
+			to_chat(owner, "<span class='userdanger'>The splint on your [display_name] comes apart!</span>")
+		else
+			splint_stacks -= 1
```
TiviPlus, line 348:
```suggestion
			splint_stacks--
```

code/modules/organs/limbs.dm
```diff
@@ -1044,7 +1049,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 			new /datum/effect_system/spark_spread(owner, owner, 5, 0, TRUE, 1 SECONDS)
 
 
-/datum/limb/proc/apply_splints(obj/item/stack/medical/splint/S, mob/living/user, mob/living/carbon/human/target)
+/datum/limb/proc/apply_splints(obj/item/stack/medical/splint/S, applied_stacks, mob/living/user, mob/living/carbon/human/target)
```
TiviPlus, line 1052: Why are you adding this as the second arg and not the last

code/modules/organs/limbs.dm
```diff
@@ -1074,6 +1079,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 		"[text1]",
 		"[text2]")
 		add_limb_flags(LIMB_SPLINTED)
+		splint_stacks = max(splint_stacks, applied_stacks)
```
TiviPlus, line 1082: Why the max(), you already check that in line 1056?

TiviPlus (changes requested):

SplinterGP (changes requested):
why does it take amount of times damage is dealt and not damage itself ? i don't see why getting attacked 5 times with 10 damage should take your splint rather than being attacked once with 60? this implementation seems very yikes to me honestly, even more considering you made it also not reduce per hit that below 10, meaning that if a marine has tyr mk2 alot of slashes wont reduce any splint stack, making the marine close to infinite splint.

Losenis:
We presplinting meta now

Hydrinator:
A cool change, but needs to be implemented right. Like others said, it has to take into account the actual damage rather than number of hits.

You're otherwise just mega pissing on certain benos.

code/modules/organs/limbs.dm
Walarks: I think it looks more ordered like this.

code/modules/organs/limbs.dm
Walarks: removed max

code/game/objects/items/stacks/medical.dm
Walarks: changed

code/modules/organs/limbs.dm
Walarks: changed

TiviPlus (approved):

TiviPlus (approved):

DerFlammenwerfer:
> Unsure about this one since it'll only lead to marines self-splinting and then later walking up to a corpsman to tell them to unsplint and splint again. Would also mean that people with nothing better to do in the round would start asking corpsmen to do it again to the same limb after a fight to replenish the stacks for the next fight.

Non-issue.

PR #6594, opened 2021-04-15
leader xenos now lose queen pheros when queen dies.

SplinterGP:
## About The Pull Request
Leader xenos now stop receiving pheros from a dead queen, they keep leader status however.

## Why It's Good For The Game
Doens't make sense for these pheros to stay even after death, which essentially are infinite pheros since queen doenst even need to be alive to give them by staying after a queen dies.

## Changelog
:cl:
balance: leader pheros are now removed when queen dies.
/:cl:


Swagile:
sad xeno hours

TiviPlus (approved):

PR #6627, opened 2021-04-19
Pool pinpointer is gone. Welcome psy tracker

BraveMole:
## About The Pull Request

Psy tracker, some QOL compared to pool pinpointer (message when a silo is placed / destroyed). Cannot get information on silo while shipside.

Most important, able to track the hive ruler if there are no silos on the map

## Why It's Good For The Game

This is meant for two things:

Prevent extra delay end game with the hive ruler running around everywhere with no pool, or placing one at the last moment. This just leads to extra boredom for all marines, and make the game longer for no reason

Make silo less strategy a bit harder to pull off. You are still able to do it, but that means the hive leader will be tracked.
That allows xeno to not be forced to defend a defensive position if they do not wish too, but while still giving marines an objective they can kill.
Silo less strat are often a competition to see who will wander around alone because of boredom, and that's no good

## Changelog
:cl:
add: Psy tracker, to replace the pool pinpointer. Able to track the hive ruler if no silos are on the map
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -72,7 +72,7 @@ OPERATIONS
 	cost = 20
 
 /datum/supply_packs/operations/pinpointer
-	name = "pool tracker crate"
+	name = "psy tracker crate"
 	contains = list(/obj/item/psy_tracker)
 	cost = 20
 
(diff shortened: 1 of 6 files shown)


code/modules/reqs/supplypacks.dm
```diff
@@ -73,7 +73,7 @@ OPERATIONS
 
 /datum/supply_packs/operations/pinpointer
 	name = "pool tracker crate"
```
BraveMole, line 75:
```suggestion
	name = "psy tracker crate"
```

SpaceLoveSs13:
Much much better. Should have this earlier. Atleast now some people won't say pools is a push or die. Which it was indeed but great pr.

TiviPlus:
I see no reason to nerf no silo hives

PR #6632, opened 2021-04-19
spits go over barricades

Pariah919:
## About The Pull Request
Spits go over barricades again.

## Why It's Good For The Game
Makes barricades a bit less effective, but they're faster overall to make now so I don't think this is needed anymore, and lets you make the engineer stop doing stuff for a bit.

## Changelog
:cl:
balance: spits go over barricades once again
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -422,7 +422,6 @@
 	icon_state = "metal_0"
 	max_integrity = 200
 	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
-	coverage = 128
 	stack_type = /obj/item/stack/sheet/metal
 	stack_amount = 4
 	destroyed_stack_amount = 2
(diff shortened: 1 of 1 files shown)


Naaanii:
Not bad, though will this affect the boiler glob or is that not included?

InterroLouis:
This will affect everything, bullets too.

Naaanii:
Might want to make it only for spits and allow nades to be fired over from range maybe.

DerrickChamberlain:
Pop some kalo and keep on trucking :)

SplinterGP:
FF BBEHIND CADES WOOOOO FF UPGRADED

SpaceLoveSs13:
Bad pr doesn't solve fobbitting problems instead create more imbalance for rest of the cadings where mini fob is applicable. Spitter spam is also possible in just one line or turret. Extremely imbalanced pr imo.

Hydrinator:
This seems like a bad idea for several reasons, one of which is encouraging flimsy ranged xenomorphs to run up face-first into barricades in order to try and spit at people instead of melting said barricades.

TiviPlus:
Basically what people said this doesn't fix anything

PR #6634, opened 2021-04-19
metal cades have less acid armor naturally and at peak

Pariah919:
## About The Pull Request
from 40 to 10, caustic armor changed but even then it has 10 less armor for a max of 50

## Why It's Good For The Game
Makes acid and beno siege more effective overall

## Changelog
:cl:
balance: metal cade acid armor reduced to 10, caustic gives you a total of 50 acid armor instead of 60
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -518,7 +518,7 @@
 		if(CADE_TYPE_MELEE)
 			soft_armor = soft_armor.modifyRating(melee = 30, bullet = 30)
 		if(CADE_TYPE_ACID)
-			soft_armor = soft_armor.modifyRating(bio = 0, acid = 20)
+			soft_armor = soft_armor.modifyRating(bio = 0, acid = 40)
 
 	barricade_upgrade_type = choice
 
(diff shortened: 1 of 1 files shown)


SpaceLoveSs13:
This will imbalance rest of the cadings instead of just fob cades. Having lowered HP cades in crash or mini fob like one in engineering is bad. Not good pr imo.

code/game/objects/structures/barricade.dm
```diff
@@ -421,8 +421,8 @@
 	desc = "A sturdy and easily assembled barricade made of metal plates, often used for quick fortifications. Use a blowtorch to repair."
 	icon_state = "metal_0"
 	max_integrity = 200
-	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 40)
 	coverage = 128
+	soft_armor = list("melee" = 0, "bullet" = 0, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 100, "rad" = 0, "fire" = 80, "acid" = 10)
```
Swagile, line 425: 20 instead of 10

Swagile (commented):
you forgot to lower sandbag acid from 40 to 20 as well

and base_acid_damage from 2 to 4 to make boiler more effective

Hydrinator:
I don't see why this change is good for the game or even necessary. Xenomorphs can already melt walls down en-masse with crushers also existing while you can also completely turn off the ability to repair barricades by having as little as weak drone acid on them.

Small fortifications would literally go down in seconds from RANGED with these changes.

TiviPlus:
I see no reason for this change, for the xth amount of time xenos need more siege tools rather than number changes

PR #6645, opened 2021-04-22
Added max_range = 40 to DMR

EvilMelonTheSecond:
## About The Pull Request
DMR given 40 tile range same as snipers.

## Why It's Good For The Game
DMR has standard range despite being a "Designated marksman rifle" meaning it can't even use its 30 tile accuracy for some reason. Now bullets won't go poof before even reaching the accuracy falloff.

## Changelog
:cl:
fix: DMR gets sniper range
/:cl:


EvilMelonTheSecond:
Time to not webedit

PR #6646, opened 2021-04-22
DMR now has sniper range

EvilMelonTheSecond:
## About The Pull Request
DMR given 40 tile range same as snipers.

## Why It's Good For The Game
DMR has standard range despite being a "Designated marksman rifle" meaning it can't even use its 30 tile accuracy for some reason. Now bullets won't go poof before even reaching the accuracy falloff.

## Changelog
:cl:
balance: DMR gets sniper range
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -524,6 +524,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
 	accurate_range_min = 0
 	accurate_range = 30
+	max_range = 40
 	damage = 65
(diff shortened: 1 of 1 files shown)


EvilMelonTheSecond:
Now with 100% less web edits

Naaanii:
Not really needed nor is it a good change. Just use the TL if you want sniper range.

EvilMelonTheSecond:
> 
> 
> Not really needed nor is it a good change. Just use the TL if you want sniper range.

Why use DMR then? The same range as an AR and lower DPS. I don't see why its a bad thing if people want long range options.

SplinterGP:
> Not really needed nor is it a good change. Just use the TL if you want sniper range.

Imagine having a DMR with less range than GPMG and other rifles

usnpeepoo:
pretty sure DMRs were already buffed with ammo capacity this seems a little overkill

SplinterGP (approved):

Issue #6342, opened 2021-03-13
Infinite Floortiles


Wirpy:
## Reproduction:

Make reinforced floortile -> crowbar it -> infinite normal floor tiles 

This leads to infinite metal because floor tiles can be converted into normal metal using the autolathe.

PR #6655, opened 2021-04-22
Fixes infinite metal from reinforced floors exploit

Wayland-Smithy:
## About The Pull Request

Fixes: #6342 

Please give this one a thorough review and suggest any better alternatives as I did not want to waste a ton of time to figure out why rfloors are `/turf/open/floor/engine` in the first place or why `/turf/open/floor/engine/make_plating()` the proc that usually handles tile removals so they can't be duplicated is totally bypassed for this type. My guess is there are some very old legacy reasons and if I were to change it could have unforeseen and dire consequences.

## Why It's Good For The Game

Bug bad. Fix good.

## Changelog
:cl:
fix: Fixed a glitch which could break the metal economy. 
/:cl:

### DIFF
--- a/code/game/turfs/floor_types.dm
+++ b/code/game/turfs/floor_types.dm
@@ -324,6 +324,8 @@
 	return
 
 /turf/open/floor/engine/attackby(obj/item/I, mob/user, params)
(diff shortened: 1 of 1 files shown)


DerrickChamberlain:
:(

TiviPlus (approved):
Looks like bay shitcode but this works

usnpeepoo:
RIP ST black magic

PR #6662, opened 2021-04-22
Cades can be rotated instantly clockwise with rclick

TiviPlus:
## About The Pull Request

titel

## Changelog
:cl:
qol: Cades can be rotated instantly clockwise with right click
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -277,6 +277,14 @@
 	setDir(turn(dir, 270))
 
 
+/obj/structure/barricade/attack_hand_alternate(mob/living/user)
+	if(anchored)
+		to_chat(user, "<span class='warning'>It is fastened to the floor, you can't rotate it!</span>")
+		return FALSE
+
+	setDir(turn(dir, 270))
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/barricade.dm
```diff
@@ -277,6 +277,14 @@
 	setDir(turn(dir, 270))
 
 
+/obj/structure/barricade/attack_hand_alternate(mob/living/user)
+	if(anchored)
+		to_chat(user, "<span class='warning'>It is fastened to the floor, you can't rotate it!</span>")
+		return FALSE
+
+	setDir(turn(dir, 270))
```
BraveMole, line 285: Might want to make a single proc used for all rotate, rather than copypasting

BraveMole (changes requested):

code/game/objects/structures/barricade.dm
SplinterGP: then you're just copypasting that rotate ? thonk

SplinterGP (approved):

PR #6669, opened 2021-04-23
Lower Jetpacks cost from 15 to 8

Lauwma:
## About The Pull Request
ReQ cost change to an item that doesn't really get any use due to its expensive cost.
Jetpack cost: 15 -> 5

## Why It's Good For The Game
Jetpacks aren't used by a lot of marines, not because of the drawbacks it has (Lack of backpack storage/second weapon) but because its price in req is way expensive for what it does.
The item doesn't need to be stronger in my opinion, it has already the ability to escape warriors/shrikes brazil abilities and give a nice leap when you need to keep shooting at a target to finish it. That's why lowering the price would maybe make it more usable. A lot of ROs thinks jetpacks are a waste of points.

At first I wanted to put the jetpack in the vendor to make it consistant but I was afraid it would become a powergaming strategy (we don't want that to happen).

10 points seems like a large cut because with 105 points you can equip 21 marines with a jetpack instead of 7 but valks/surts/hlin are common and makes that the RO is usually left with less than 20 points (who would use what's left on a single jetpack ?).
Still, after the changes, if jetpacks are becomming too common a raise to 10 would be more appropriate.

## Changelog
:cl:
balance: Lowers Jetpacks cost in requisition from 15 to 5.
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -956,7 +956,7 @@ CLOTHING
 	contains = list(
 		/obj/item/jetpack_marine,
 	)
-	cost = 15
+	cost = 5
 
 /*******************************************************************************
 MEDICAL
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -956,7 +956,7 @@ CLOTHING
 	contains = list(
 		/obj/item/jetpack_marine,
 	)
-	cost = 5
+	cost = 15
 
 /*******************************************************************************
 MEDICAL


AgentMA1:
>5 points for Jetpack

What.

RenGusta:
Hot.

Wayland-Smithy:
I think a 2/3 cost reduction is a bit much, also don't forget to make your changelog more descriptive so players in-game know what you did.

Lauwma:
After reviews I'm considering to lowering the jetpack to 8, roughtly halving the price and making that a module + a jetpack would cost 20. (I mean valkiries are supposed to fly no ?)

Lauwma:
Well uh I did an oopsie with my fork, I can't make changes now but well I guess I'll make another PR if it's approved

PR #6689, opened 2021-04-26
Fixes crusher charge damaging resin silos and turrets

Wayland-Smithy:
## About The Pull Request

Prevents `charge_crush`  logic from acting on resin silos and turrets, doors in their slow form and walls will still be crushed as normal.

## Why It's Good For The Game

Crusher harming these structures appears to have been an oversight as I was unable to find a way for any other caste to cause them harm, so it stands to reason it shouldn't be happening.

## Changelog
:cl:
fix: Crusher charge no longer damages resin silos and turrets. No more mystery silo damage alerts.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
+++ b/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
@@ -286,6 +286,8 @@
 
 	if(isobj(crushed))
 		var/obj/crushed_obj = crushed
+		if(istype(crushed_obj, /obj/structure/resin/silo) || istype(crushed_obj, /obj/structure/resin/xeno_turret))
+			return precrush2signal(crushed_obj.post_crush_act(charger, src))
 		playsound(crushed_obj.loc, "punch", 25, 1)
 		var/crushed_behavior = crushed_obj.crushed_special_behavior()
 		crushed_obj.take_damage(precrush, BRUTE, "melee")


code/modules/mob/living/carbon/xenomorph/charge_crush.dm
```diff
@@ -286,6 +286,8 @@
 
 	if(isobj(crushed))
 		var/obj/crushed_obj = crushed
+		if(istype(crushed_obj, /obj/structure/resin/silo) || istype(crushed_obj, /obj/structure/resin/xeno_turret))
```
BraveMole, line 289: maybe a flag would be cleaner? Like CRUSHER_CHARGE_IMMUNE or something

Wayland-Smithy: Perhaps. I am hoping to maybe just apply it to all resin structures in the future if the need to smash resin doors permanently goes away.

TiviPlus:
Why does it crush silos, it shouldnt be crushing nondense objs if you ask me

Wayland-Smithy:
I don't think crusher damaging silos is new, and simply ignoring non-dense may not work for the resin turret. Could that also make crusher no longer able to destroy non dense non passable objects?

TiviPlus:
Agreed in that it wont work for the turret, but it seems the silo is a seperate issue as I'd expect not hitting a silo to not damage it

SplinterGP (approved):

SplinterGP:
> 
> 
> Agreed in that it wont work for the turret, but it seems the silo is a seperate issue as I'd expect not hitting a silo to not damage it

it checks for anchor only i think, it just hits all objects in way

TiviPlus (approved):
If it checks anchored that sprobably a rabbit hole

PR #6697, opened 2021-04-27
Fix maptables and command headset minimaps

TheOneAndMightyRed:
## About The Pull Request

Random bug i encountered while doing other stuff:
If a maptable exists roundstart , it will get an empty command groundside map and get that empty map saved in the subsystem.
Meaning that every map table and command headset with the groundside minimap (both of which share the same minimap) won't be able to open the minimap because an empty one from the maptable's fetch_minimap_object proccall was saved.
Happens most likely due to the atoms subsystems initializing the maptable waay before the minimap subsystem has the chance to generate the minimaps itself, causing fetch_minimap_object to fuck up.

This bug won't happen if the map doesn't start with a map table initially, which is the case for debugdalus. Might be why this bug got through, since spawning a maptable when one doesn't exist initially doesn't get the minimap fucked up and actually works as intended.

## Why It's Good For The Game

Maptable go brrrr

## Changelog
:cl:
fix: Map tables in conjuntion with command headset minimaps should work again.
/:cl:

### DIFF
--- a/code/controllers/subsystem/minimaps.dm
+++ b/code/controllers/subsystem/minimaps.dm
@@ -285,7 +285,7 @@ SUBSYSTEM_DEF(minimaps)
 		return hashed_minimaps[hash]
 	var/obj/screen/minimap/map = new(null, zlevel, flags)
 	if (!map.icon) //Don't wanna save an unusable minimap for a z-level.
-		return
+		CRASH("Empty and unusable minimap generated for '[zlevel]-[flags]'") //Can be caused by atoms calling this proc before minimap subsystem initializing.
 	hashed_minimaps[hash] = map
 	return map
 
--- a/code/controllers/subsystem/minimaps.dm
+++ b/code/controllers/subsystem/minimaps.dm
@@ -284,6 +284,8 @@ SUBSYSTEM_DEF(minimaps)
 	if(hashed_minimaps[hash])
 		return hashed_minimaps[hash]
 	var/obj/screen/minimap/map = new(null, zlevel, flags)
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/cic_maptable.dm
```diff
@@ -28,6 +24,8 @@
 		return
 	if(!user.client)
 		return
+	if (!map)
+		map = SSminimaps.fetch_minimap_object(targetted_zlevel, allowed_flags)
```
TiviPlus, line 28: Better would probably be to put this on lateinitialize, reason it gets a null map is because minimaps initializes AFTER atoms
better wuld be a queue like markers but this will do just as well for now

code/controllers/subsystem/minimaps.dm
```diff
@@ -284,6 +284,8 @@ SUBSYSTEM_DEF(minimaps)
 	if(hashed_minimaps[hash])
 		return hashed_minimaps[hash]
 	var/obj/screen/minimap/map = new(null, zlevel, flags)
+	if (!map.icon) //Don't wanna save an unusable minimap for a z-level.
+		return
```
TiviPlus, line 288: Make this a CRASH

TiviPlus (changes requested):

code/game/objects/machinery/cic_maptable.dm
TheOneAndMightyRed: LateInitialize doesn't work due to A.LateInitialize() getting called right after A.Initialize() inside the Atoms subsystem init itself.
And the queue markers.... i have no idea.

PR #6702, opened 2021-04-27
Xenos keep their targeting zone selection across evolutions

Wayland-Smithy:
## About The Pull Request

Adds a bit of evolution code to retain `zone_selected` when xenos change caste.

Fixed a small bug causing `set_selected_zone()` to pass the wrong mob to an overridden `update_icon()` making it fail to update `zone_selected` in some cases.

## Why It's Good For The Game

QoL good. Xenos needing to rundown a post evolution checklist in order to get back to full readiness is bad. It was worse when it reset the old mmb/shiftclick pref and this is almost as annoying.

## Changelog
:cl:
qol: Xenos now keep their selected targeting zone when evolving/regressing.
fix: Fixed a small bug with the zone selection hud.
/:cl:

### DIFF
--- a/code/_onclick/hud/screen_objects/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects/screen_objects.dm
@@ -453,7 +453,7 @@
 
 	if(choice != selecting)
 		selecting = choice
-		update_icon(usr)
+		update_icon(user)
 	return TRUE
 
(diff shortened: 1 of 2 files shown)


Naaanii:
Nice

TiviPlus (approved):

PR #6708, opened 2021-04-27
Lower Jetpacks cost from 15 to 12

Lauwma:
## About The Pull Request
ReQ cost change to an item that doesn't really get any use due to its expensive cost.
Jetpack cost: 15 -> 12

## Why It's Good For The Game
Jetpacks aren't used by a lot of marines, not because of the drawbacks it has (Lack of backpack storage/second weapon) but because its price in req is way expensive for what it does.
The item doesn't need to be stronger in my opinion, it has already the ability to escape warriors/shrikes brazil abilities and give a nice leap when you need to keep shooting at a target to finish it. That's why lowering the price would maybe make it more usable. A lot of ROs thinks jetpacks are a waste of points.

At first I wanted to put the jetpack in the vendor to make it consistant but I was afraid it would become a powergaming strategy (we don't want that to happen).

Valks/surts/hlin are highly demanded and makes that the RO is usually left with less than 20 points and doesn't want to spend 15 points for a jetpack.

## Changelog
:cl:
balance: Jetpacks cost 15 -> 12
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -956,7 +956,7 @@ CLOTHING
 	contains = list(
 		/obj/item/jetpack_marine,
 	)
-	cost = 15
+	cost = 10
 
 /*******************************************************************************
 MEDICAL
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -956,7 +956,7 @@ CLOTHING
 	contains = list(
 		/obj/item/jetpack_marine,
 	)
-	cost = 10
+	cost = 12
 
 /*******************************************************************************
 MEDICAL


David-Stormwell:
The more you spam it the more it gets nerfed until it'll become shite.
There's a reason why it's 15 and not less, having a item to avoid all xenos is quite good.

TiviPlus (approved):

Naaanii:
15 points isn't even much for what it offers. The drawback for less storage from a satchel isn't that large of a drawback for granting a decent amount of mobility. A jetpack costing 15 points only is already a boon, lowering the cost further is just biting off more than you can chew.

Lauwma:
Why nobody's playing the jetpack thought if it's that good?

february-ops:
> Why nobody's playing the jetpack thought if it's that good?

high risk high reward, if a marine is good with it he will dominate if he is bad 15 poinits to waste just like half the things on req, it is not beccause its "high cost" is becayse its nieche and difficult to use kinda

PR #6715, opened 2021-04-29
Readds Thermo to req at an increased cost.

Naaanii:
## About The Pull Request

Readds Thermobaric to requisitions but it now costs 500 points.
Ammo cost is lowered to 25.

## Why It's Good For The Game

With the changes to mining in general and it being harder to gain and accumulate a large number of points, the thermobaric launcher is a good way to reward marines managing to hold out long enough or help end stalemates giving them that extra firepower they need to push out in pools or hunt.

Cadehugging and holding FOB rarely ever generates enough points to get a thermo but if marines are constantly pushing, killing and gaining enough ground to secure several miners and manage to generate enough spare points for a weapon that can help them push better, then why not?

## Changelog
:cl:
add: The Quad launcher and it's ammo is back in stock.
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -206,7 +206,6 @@ WEAPONS
 	contains = list(/obj/item/storage/box/recoilless_system)
 	cost = 40
 
-
 /datum/supply_packs/weapons/railgun
 	name = "TX-220 Railgun"
 	contains = list(/obj/item/weapon/gun/rifle/railgun)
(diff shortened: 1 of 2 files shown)


David-Stormwell:
We'll have other guns that will be upgrades to buy from req, if they are willing to spend enough points that is later.
Thermo is overkill.

Moggram:
Ya no, as some one how like to play as RO. I would just buy ten smartgun kits and ammo over this.

TiviPlus:
Maybe as a seperate weapon, but seems people dont like that either
Absolutely not as a deathsquad gun though

PR #6743, opened 2021-05-01
Attachment code improvements

TiviPlus:
## About The Pull Request

Fixed stuff not using defines when it should, qdeleting images, bad var names and a bit more

## Changelog
:cl:
code: improved attachment code
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -315,16 +315,16 @@ should be alright.
 	//Checks if they can attach the thing in the first place, like with fixed attachments.
 	var/can_attach = 1
 	switch(attachment.slot)
-		if("rail")
+		if(ATTACHMENT_SLOT_RAIL)
 			if(rail && !(rail.flags_attach_features & ATTACH_REMOVABLE))
 				can_attach = FALSE
-		if("muzzle")
+		if(ATTACHMENT_SLOT_MUZZLE)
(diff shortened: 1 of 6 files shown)


code/modules/projectiles/gun_helpers.dm
```diff
@@ -533,7 +537,7 @@ should be alright.
 
 	usr.visible_message("<span class='notice'>[usr] strips [A] from [src].</span>",
 	"<span class='notice'>You strip [A] from [src].</span>", null, 4)
-	A.Detach(usr)
+	A.detach_from_master_gun(usr)
```
BraveMole, line 540: guns have gun_user since the rework, you could use that rather than usr. (gun user is null only when the gun is not in a hand)

XSlayer300:
hmm, is it possible to define a silenced firing sound for the same surpressor? like, have each gun their own silencer sound?

code/modules/projectiles/gun_helpers.dm
TiviPlus: this is in a verb

SplinterGP:
> 
> 
> hmm, is it possible to define a silenced firing sound for the same surpressor? like, have each gun their own silencer sound?

you could and you could just make the silencer change the frequency by some amount and lower volume, could work as well

SplinterGP (approved):

PR #6751, opened 2021-05-03
Context Menu on Shift Right Click

SkeletalElite:
## About The Pull Request
Puts context menu on shift right click, the right click toggle will change it back to standard for boomers who don't want better controls.

This is ported from TGs combat mode , just the right click menu changes though, no intent changes.

## Why It's Good For The Game
Allows you to have best of both worlds where you get context menu as well as using right click for game play.

~~If moving the context menu to shift right click is unpopular however, I can make it a preference.~~ I think it's better to not have it as a preference though, since any new usage of right click added has to also have a way to be used without right click if it's a preference.

It has now been made a toggle. I still think it'd be better if it wasn't because we really should make it that way eventually so new controls aren't being held back by boomers using old controls.

## Changelog
:cl:
qol: Context/right click/verb menu toggle now changes whether or not you must hold shift to access the menu instead of disabling it completely
/:cl:


BraveMole:
Oh boy, strap yourself the haters are coming

TiviPlus (commented):
shouldnt you also be setting show_popup_menus = false ?

TiviPlus:
Also ghosts should probably have right click menus on by default

SkeletalElite:
> shouldnt you also be setting show_popup_menus = false ?

show_popup_menus = false disables the menu entirely, this PR just makes it to where you need to press shift + right click to make it pop up

code/modules/client/client_procs.dm
```diff
@@ -881,3 +883,13 @@ GLOBAL_VAR_INIT(automute_on, null)
 		SSambience.ambience_listening_clients[src] = world.time + 10 SECONDS //Just wait 10 seconds before the next one aight mate? cheers.
 	else
 		SSambience.ambience_listening_clients -= src
+
+/client/proc/set_right_click_menu_mode(shift_only)
```
TiviPlus, line 887: autodoc

TiviPlus (approved):

Texicos:
this
but make it a preference

GravitronBlue:
Trying to melt things with acid just brings up the context menu still unlike the old toggle preference.

and as Bravemole predicted, thumbs down to the whole combat mode system in general.

Al-1ce:
If you're going to add this at the bare minimum make it a preference. The majority of players don't want anything to do with combat mode.

Dolth:
Please, make it a preference, this is really a hell to use when I've spent 5 years on CM/TGMC with regular right clicking.
I see no reason to do that.

SplinterGP (changes requested):
people dindt like this much, the option is way more favourable from what i've seen people saying and having the option was better, a keybind for the option is honestly enough IMO and keeps the best of 2 worlds

ZahaSanko:
Begone, fiend! or do it as a choice

SwagstylerS:
The Quality Of Life tag is quite funny when it is just simply annoying. Make it an option and please do not put it on screen. Xenos already have shit ton of buttons on it.

freshguacc:
These controls add little, of use, since the middle mouse button already does what the right mouse button does at least for xenos. It's a control scheme with redundant buttons which should be avoided. If it really is added permanently it should be a preference or a toggle at least. For marines, Z is now the same as pressing RMB, which is again redundant controls. Pressing Z or MMB for fast actions is not a bothersome control scheme, however pressing both shift+RMB IS awkward for an action that you do as frequently as opening the context menu.

Dolth:
@freshguacc It's more a matter of letting people keybind the game like they want and not forcing them a single way to play. Especially when they're no visible use to enforce one way or another.

jroinc:
Does anyone actually like this? Serious question.

jroinc:
Hey, so does anyone actually like this? Serious question.

This isn't an off-topic question. This is a legitimate question. At the time of me posting this, it has one like from bravemole, and 11 dislikes. This appears to be change for the sake of change.

Dolth:
> Hey, so does anyone actually like this? Serious question.
> 
> This isn't an off-topic question. This is a legitimate question. At the time of me posting this, it has one like from bravemole, and 11 dislikes. This appears to be change for the sake of change.

https://i.gyazo.com/3c44ac306550a5397bdfc9ac5c33deb6.png

MiniMeatwad:
Make it optional, do not make it a requirement. Please don't merge this until it's optional.

SkeletalElite:
> Make it optional, do not make it a requirement. Please don't merge this until it's optional.

It's been optional for 13 hours now, the testmerge just hasn't been updated

code/modules/client/client_defines.dm
```diff
@@ -17,6 +17,7 @@
 	//Preferences related
 	var/datum/preferences/prefs 	= null
 	var/inprefs = FALSE
+	var/shift_to_open_context_menu = TRUE ///remembers what our context menu setting is currently set to
```
TiviPlus, line 20: Autodoc goes on the above line

TiviPlus (changes requested):
Ok PSA
As of this PR the only supported control scheme is the shift right click menu
Any prs that do not conform to it will be made to 
Explanation can be seen at https://discord.com/channels/498569646014857217/500801682058379265/839516718208385086

TiviPlus:
Also keep off topic posting and ill lock the pr

SkeletalElite:
> As of this PR the only supported control scheme is the shift right click menu

Would you like me to revert to pre-toggle then?

SplinterGP (changes requested):
can you make it a keybind instead of perma shift ? examine does it already, same with dragging that is ctrl but its still a keybind

code/modules/client/preferences_toggles.dm
```diff
@@ -333,13 +333,23 @@ GLOBAL_LIST_INIT(ghost_others_options, list(GHOST_OTHERS_SIMPLE, GHOST_OTHERS_DE
...
 	set category = "Preferences"
 
-	show_popup_menus = !show_popup_menus
-	to_chat(src, "<span class='interface'>Right click menu [show_popup_menus ? "en" : "dis"]abled.</span>")
+	if(shift_to_open_context_menu)
+		winset(src, "mapwindow.map", "right-click=false")
+		winset(src, "default.Shift", "is-disabled=true")
+		winset(src, "default.ShiftUp", "is-disabled=true")
+		shift_to_open_context_menu = FALSE
+		to_chat(usr, "<span class='notice'>You will no longer need to hold shift to access the right click menu</span>")
```
XSlayer300, line 346:
```suggestion
		to_chat(usr, "<span class='notice'>You will no longer need to hold the Shift key to access the right click menu.</span>")
```

code/modules/client/preferences_toggles.dm
```diff
@@ -333,13 +333,23 @@ GLOBAL_LIST_INIT(ghost_others_options, list(GHOST_OTHERS_SIMPLE, GHOST_OTHERS_DE
...
+	else
+		winset(src, "mapwindow.map", "right-click=true")
+		winset(src, "ShiftUp", "is-disabled=false")
+		winset(src, "Shift", "is-disabled=false")
+		shift_to_open_context_menu = TRUE
+		to_chat(usr, "<span class='notice'>You will need to hold shift to access the right click menu</span>")
```
XSlayer300, line 352:
```suggestion
		to_chat(usr, "<span class='notice'>You will need to hold the Shift key to access the right click menu.</span>")
```

XSlayer300:
The pref is not saving, even after a round. Make sure atleast the player can just toggle it on and off once rather than doing it after each round. By default, it should be off.

SkeletalElite:
> The pref is not saving, even after a round. Make sure atleast the player can just toggle it on and off once rather than doing it after each round. By default, it should be off.

Can't test merge PRs that touch the save file

TiviPlus:
> can you make it a keybind instead of perma shift ? examine does it already, same with dragging that is ctrl but its still a keybind

That's not how it works you were in the discord conversation

SplinterGP:
> 
> 
> > can you make it a keybind instead of perma shift ? examine does it already, same with dragging that is ctrl but its still a keybind
> 
> That's not how it works you were in the discord conversation

the what, it can work

PR #6758, opened 2021-05-04
Makes silos unable to be placed near walls as much

Walarks:
## About The Pull Request
Adds a check for whether there is a wall at/behind/left/right of a silo spot before being able to place one.

## Why It's Good For The Game
Less silos abuse. maybe less ahelps.

## Changelog
:cl:
balance: added a check for whether there is a wall at/behind/left/right of a silo spot before being able to place one
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -1064,6 +1064,17 @@
 	if(!.)
 		return FALSE
 
+	var/turf/T = A.loc
+	if(T?.density)
+		to_chat(owner, "<span class='xenowarning'>You need open ground to place that!</span>")
+		return FALSE
+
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/abilities.dm
```diff
@@ -1064,6 +1064,17 @@
 	if(!.)
 		return FALSE
 
+	var/turf/T = A.loc
```
TiviPlus, line 1067: use get_turf

TiviPlus (changes requested):

code/modules/mob/living/carbon/xenomorph/abilities.dm
Walarks: changed

TiviPlus (approved):

PR #6769, opened 2021-05-05
Tac Reload Buff

SkeletalElite:
## About The Pull Request
You can tac reload while moving

## Why It's Good For The Game
I think tac reloads aren't very useful since you have to stop moving to do it. This will make them a little more useful.

## Changelog
:cl:
balance: You can tac reload while moving
/:cl:


SkeletalElite:
If it's too strong I can make it take longer but I think it's better if it doesn't restrict movement

TiviPlus (approved):

usnpeepoo:
> If it's too strong I can make it take longer but I think it's better if it doesn't restrict movement

you should probably make it a lil bit longer if its the exact same time it takes to reload standing still

PR #6771, opened 2021-05-05
Fix xeno structures resistance_flags, update VV bitfields

Wayland-Smithy:
## About The Pull Request

Fixes resin silos losing their UNACIDABLE and xeno turrets lacking their DROPSHIP_IMMUNE.

Also updated the VV bitmasks to include the until now missing DROPSHIP_IMMUNE define.

## Why It's Good For The Game

Fix good. Acidable silos bad. Tad crushable turrets bad, turf/closed/walls/resin automatically block landing but not dense turrets which cost finite psy points? Yeah this is a salt PR 🧂🧂🧂

## Changelog
:cl:
fix: Fixed resistance_flags for important xeno structures, unacidable + dropship immune.
qol: Added DROPSHIP_IMMUNE resistance flag to the VV bitfields list.
/:cl:


BraveMole:
Tad crushing turrets was intended

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -44,6 +43,7 @@
 
 /obj/structure/resin/silo/Initialize()
 	. = ..()
+	resistance_flags |= DROPSHIP_IMMUNE
```
BraveMole, line 46: ew. Just add all the correct resistance flags at compilation, don't do this

TiviPlus: yes just have them declared properly

Wayland-Smithy: > Just add all the correct resistance flags at compilation

As there are flags set for a parent it was that kind of thinking that caused the silos to lose their UNACIDABLE in the first place. Pretty sure my way was a fine practice at some point.

I lack faith in future contributors/reviewers to not make the same mistake by forgetting to check for parent flags and I do not think taking that risk is worth the performance gain of not changing a bitmask at runtime, but have it your way.

TiviPlus (approved):

PR #6087, opened 2021-02-18
Autolath requires raw mats to work

BraveMole:
## About The Pull Request

Autolath now requires sheets of metal and glass. It also starts full

## Why It's Good For The Game

No more GL grinding. This was nothing but bug abuse, and on top of that it was extremly boring. Better have cheaper metal for engie at start or lower the price in req. This is meant to go with https://github.com/tgstation/TerraGov-Marine-Corps/pull/6099

## Changelog
:cl:
tweak: Autolath can now longer grind objects to have free mats
tweak: Autolath has double the capacity for mats, and starts full
/:cl:

PR #6782, opened 2021-05-06
In which two ways to get large quantities of metal for no cost are described fully and completely.

jroinc:
## About The Pull Request

Method one-
Each T81 GL is worth 1.75 metal sheets when autolathed. No one ever takes them as they're single shot, so the 250 in the vendor became 430 metal sheets, or 8 and a half stacks.

They're now worth 10% of that, so yeah. No more shoveling half of the round's grenade launchers for near-unlimited metal.

Method two-
If you have an autolathe and a stack of 20 rglass, you can feed the rglass in.
Print off one sheet of rglass. Print off all the metal that the rglass gave you, 9 or 10 sheets. Print off a stack of 19 glass.
Click on the glass with the rglass. Now you have 9 metal created from nothing, 20 rglass, and an autolathe ready to go again. Elapsed time: 5-10 seconds for around 10 metal.

This brute-force fixes the metal glitch by making it so low value it's not worth attempting. If someone more experienced than me has a better idea of how to fix it, they're welcome to it.

## Why It's Good For The Game

Prevents people from gaining infinite metal.

## Changelog
:cl:
fix: fixed two exploits allowing production of arbitrarily large amounts of metal
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -654,7 +654,7 @@
 	max_shells = 1 //codex
 	caliber = CALIBER_40MM //codex
 	load_method = SINGLE_CASING //codex
-	materials = list(/datum/material/metal = 7000)
+	materials = list(/datum/material/metal = 700)
 	w_class = WEIGHT_CLASS_BULKY
 	flags_equip_slot = ITEM_SLOT_BACK|ITEM_SLOT_BELT
 	throw_speed = 2
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -654,7 +654,6 @@
 	max_shells = 1 //codex
 	caliber = CALIBER_40MM //codex
 	load_method = SINGLE_CASING //codex
-	materials = list()
 	w_class = WEIGHT_CLASS_BULKY
 	flags_equip_slot = ITEM_SLOT_BACK|ITEM_SLOT_BELT
 	throw_speed = 2
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -654,7 +654,7 @@
 	max_shells = 1 //codex
 	caliber = CALIBER_40MM //codex
 	load_method = SINGLE_CASING //codex
-	materials = list(/datum/material/metal = 700)
+	materials = list()
 	w_class = WEIGHT_CLASS_BULKY
 	flags_equip_slot = ITEM_SLOT_BACK|ITEM_SLOT_BELT
 	throw_speed = 2
--- a/code/game/objects/items/stacks/sheets/glass.dm
+++ b/code/game/objects/items/stacks/sheets/glass.dm
@@ -4,7 +4,6 @@
 *		Reinforced glass sheets
 *		Phoron Glass Sheets
 *		Reinforced Phoron Glass Sheets (AKA Holy fuck strong windows)
-*		Glass shards - TODO: Move this into code/game/object/item/weapons
 */
 
 /*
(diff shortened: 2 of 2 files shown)


MiniMeatwad:
Wasn't there a bug that made reinforced metal tiling infinitely wrenchable, where you use metal rods on plating and wrench the tile for infinite metal?

jroinc:
> 
> 
> Wasn't there a bug that made reinforced metal tiling infinitely wrenchable, where you use metal rods on plating and wrench the tile for infinite metal?

Probably, but I don't know about it. These are the only 2 I know.

MiniMeatwad:
You can also wire cut by spam clicking with wirecutters on fences and then repair them for infinite metal too.

DerrickChamberlain:
> Wasn't there a bug that made reinforced metal tiling infinitely wrenchable, where you use metal rods on plating and wrench the tile for infinite metal?

It was patched, and it was a crowbar.

DerrickChamberlain:
I mean that is hardly infinite, we can only get so much from it. It also requires someone to put in the effort to make it, engineers to build with it. I really see no problem with the stuff we have now. I have not seen any economy breaking metal generators as of yet.

SkeletalElite:
The GL change is a balance change not a fix. Other one is a fix though

jroinc:
> 
> 
> The GL change is a balance change not a fix. Other one is a fix though

https://github.com/tgstation/TerraGov-Marine-Corps/pull/6087

It's not intended. Almost all other guns had all their metal values removed to prevent this. I just lowered it cause I think it does make sense, just not enough sense to get 9 stacks of metal from.

SplinterGP:
> 
> 
> I mean that is hardly infinite, we can only get so much from it. It also requires someone to put in the effort to make it, engineers to build with it. I really see no problem with the stuff we have now. I have not seen any economy breaking metal generators as of yet.

the what

SplinterGP:
weapons are not meant to be put on autolathe, otherwise there would be no 250 of each

code/modules/projectiles/guns/specialist.dm
```diff
@@ -654,7 +654,7 @@
 	max_shells = 1 //codex
 	caliber = CALIBER_40MM //codex
 	load_method = SINGLE_CASING //codex
-	materials = list(/datum/material/metal = 7000)
+	materials = list(/datum/material/metal = 700)
```
SplinterGP, line 657: or remove, dunno if it has parent without materials
```suggestion
	materials = list()
```

SplinterGP (changes requested):

jroinc:
> 
> 
> I mean that is hardly infinite, we can only get so much from it. It also requires someone to put in the effort to make it, engineers to build with it. I really see no problem with the stuff we have now. I have not seen any economy breaking metal generators as of yet.

I have!

Hell, I described it step-by-step! It literally gives infinite metal.

Georgeur:
Two other issues:
In engineering, the pipe dispenser can be used for infinite metal if you put pipes into the autolathe, its really slow, but it is infinite.

Also in the sulaco, in the briefing room, in prep, and in one or two other places, you can deconstruct all the chairs for about 160 metal. This is because unlike the chairs in the pillar of spring prep, that use, obj/strcture/bed/chair/wood, the saluco uses obj/strcture/bed/chair.

SplinterGP:
> 
> 
> Two other issues:
> In engineering, the pipe dispenser can be used for infinite metal if you put pipes into the autolathe, its really slow, but it is infinite.
> 
> Also in the sulaco, in the briefing room, in prep, and in one or two other places, you can deconstruct all the chairs for about 160 metal. This is because unlike the chairs in the pillar of spring prep, that use, obj/strcture/bed/chair/wood, the saluco uses obj/strcture/bed/chair.

the chairs are known and since its limited and all we've never cared much about it as it always has been deconstructed anyway, so we take it into account

code/modules/projectiles/guns/specialist.dm
```diff
@@ -654,7 +654,7 @@
 	max_shells = 1 //codex
 	caliber = CALIBER_40MM //codex
 	load_method = SINGLE_CASING //codex
-	materials = list(/datum/material/metal = 7000)
+	materials = list()
```
SplinterGP, line 657: actually just delete this
```suggestion
```

SplinterGP (changes requested):

SplinterGP (approved):

PR #6785, opened 2021-05-06
Fix psy drained trait being applied too often

Wayland-Smithy:
## About The Pull Request

Don't apply `TRAIT_PSY_DRAINED` during `set_undefibbable()` anymore.

Do apply it on successful headbite, so we can still get the use out of the hud icon for non distress modes.

## Why It's Good For The Game

Eating a bullet, going cold, or calling undefibbable any other way prevents psy draining for some reason. This is the fix.

## Changelog
:cl:
fix: Fixed psy drain not working on some corpses which it should.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -46,7 +46,6 @@
 
 /mob/living/carbon/human/proc/set_undefibbable()
 	ADD_TRAIT(src, TRAIT_UNDEFIBBABLE , TRAIT_UNDEFIBBABLE)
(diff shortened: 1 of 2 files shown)


BraveMole:
Not a big fan, how would you be able to drain something if that thing is permadead?

Also, don't add psy drain on headbite, the hud should stop showing when perma dead or psydrained

Wayland-Smithy:
Handwave it how ever you must to not break your emersion. 

I'm not a big fan of how you set it up so that if someone eats a bullet they not cuck xenos of a host but also their psy points.

If you insist I can make the psy HUD only work during modes with Distress Signal xeno abilities enabled.

TiviPlus (approved):

PR #6787, opened 2021-05-07
tx-11 gets higher force

Pariah919:
## About The Pull Request
title, 30

## Why It's Good For The Game
Getting hit by a brick hurts

## Changelog
:cl:
balance:tx-11 has higher melee damage
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1083,6 +1083,7 @@
 	caliber = CALIBER_492X34_CASELESS //codex
 	max_shells = 70 //codex
 	wield_delay = 0.65 SECONDS
+	force = 30
 	fire_sound = "sound/weapons/guns/fire/m412.ogg"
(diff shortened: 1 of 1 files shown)


Al-1ce:
The TX-11 which is based off the G-11 is not a 'brick' and only weighs about 10bs loaded, the average rifle weight. I don't really understand why boosting the melee damage is reasonable considering its already unwieldy to fire let alone use as a club.

TiviPlus (changes requested):
A) Why
B) fix cl

davidchan13:
On the one hand, bash attacks mean _nothing_ and all the guns could have 30-50 melee damage without major change.

On the other hand, this gives it more bash damage than the T-12, T-37 and T-64, let alone stuff like the AK. This puts it on par with the T-42, T-29 and just shy of the T-60 in terms of bash damage, as a carbine parallel weapon that just doesn't make sense.

TiviPlus:
Stale

PR #6814, opened 2021-05-10
Make accidentally opening the admin dsay dialog less annoying

Wayland-Smithy:
## About The Pull Request

Simply adds a check for a message body to early return for short circuiting further dsay code if there is no message to send in the first place.

## Why It's Good For The Game

Less pointless warnings good.

## Changelog
:cl:
admin: DSAY no longer pesters mentors who cancel or submit an empty message with a notice about needing to be dead to do the thing they canceled doing.
/:cl:

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -534,7 +534,7 @@
 	set name = "dsay"
 	set hidden = TRUE
 
-	if(!check_rights(R_ADMIN|R_MENTOR))
+	if(!msg || !check_rights(R_ADMIN|R_MENTOR))
 		return
 
 	if(is_mentor(src) && mob.stat != DEAD)
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -534,7 +534,12 @@
 	set name = "dsay"
 	set hidden = TRUE
 
-	if(!msg || !check_rights(R_ADMIN|R_MENTOR))
+	if(!check_rights(R_ADMIN|R_MENTOR))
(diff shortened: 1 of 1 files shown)


code/modules/admin/admin_verbs.dm
```diff
@@ -534,7 +534,7 @@
 	set name = "dsay"
 	set hidden = TRUE
 
-	if(!check_rights(R_ADMIN|R_MENTOR))
+	if(!msg || !check_rights(R_ADMIN|R_MENTOR))
 		return
```
TiviPlus, line 538: Why not just move the below checks to under the !msg check?

Wayland-Smithy: !msg check happens much later after a bunch of other branching stuff and a spam check plus a sanitize that just don't ever need to run if we don't provide a dirty msg to begin with

TiviPlus: yes but the checks don't matter when they are runs since it all goes to the same anyway difference is that it fails silently instead

Wayland-Smithy: the question is do we care about trying to sanitize the empty string that we would get from a blank input or cancel button? if the perf hit is nothing than we can just do it first or maybe right after the `check_rights()` if we want to skip the sanitize for blank msg we need to check its dirty contents somewhere

TiviPlus: i mean if you're considering perf you check message twice, and this is not exactly the hardest hitting part of speech

TiviPlus (approved):

PR #6832, opened 2021-05-11
Banish Now Toggles On/Off; Recall Removed v2

Surrealaser:
Banish Now Toggles On/Off; Recall Removed

Declutters interface; better QoL

:cl:
qol: Banish now toggles off; Recall removed. To end Banish early while self-banished use Banish on self or use a Banish keybind.
/:cl:


code/datums/keybinding/xeno.dm
```diff
@@ -516,10 +516,4 @@
 	description = "Banish a creature or object a short distance away within line of sight to null space. Can target oneself and allies. Can be manually cancelled with Recall."
 	keybind_signal = COMSIG_XENOABILITY_BANISH
 
-/datum/keybinding/xeno/recall
-	name = "recall"
-	full_name = "Wraith: Recall"
-	description = "Recall a target from netherspace, ending Banish's effect."
-	keybind_signal = COMSIG_XENOABILITY_RECALL
```
BraveMole, line 523: Pretty sure we can't remove keybinds. 
If this keybind is set and then removed from code, there will be a runtime everytime you user that key, with no possiblity of fixing it. 
Just let it there until we have a method to sanitize keybinds savefiles

SplinterGP: i remember there was a PR that allowed you to do so, we justnever tried to

TiviPlus (approved):

PR #6834, opened 2021-05-12
Reorder end of round automatic votes, groundmap last

Wayland-Smithy:
## About The Pull Request

Title.

## Why It's Good For The Game

More people tend to vote in the later polls after they have died once or twice in EORD. The most important of the three should get the last slot especially because respawning makes you lose the vote UI button and must be found in the OOC tab.

Gamemode polls can always be started by players the following round if the turnout was subpar, but we are stuck with that map 1 or 2 people total voted for (Prison or Barrenquilla) without admin intervention.

## Changelog
:cl:
qol: Groundmap Vote is now the last of the post round automatic polls instead of the first.
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -288,9 +288,9 @@ SUBSYSTEM_DEF(vote)
 
 ///Starts the automatic map vote at the end of each round
 /datum/controller/subsystem/vote/proc/automatic_vote()
-	initiate_vote("groundmap", null, TRUE)
+	initiate_vote("gamemode", null, TRUE)
 	addtimer(CALLBACK(src, .proc/initiate_vote, "shipmap", null, TRUE), CONFIG_GET(number/vote_period) + 3 SECONDS)
-	addtimer(CALLBACK(src, .proc/initiate_vote, "gamemode", null, TRUE), CONFIG_GET(number/vote_period) * 2 + 6 SECONDS)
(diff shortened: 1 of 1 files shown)


BraveMole:
Ok but why game mode last then?

I'd rather make the first vote still be groundmap, but 1 or 2 minute after round end. And make EORD stops like 10 seconds after the last vote

Wayland-Smithy:
> Ok but why game mode last then?
Because..
> Gamemode polls can always be started by players the following round if the turnout was subpar
Read my PR before commenting please

TiviPlus:
i'd say shipmap and gamemode swap since gamemode is more impactful?

Wayland-Smithy:
Gamemode is the only one the can be re-voted on after reboot before the next round starts so that is why I gave it lowest priority, again any player can start the poll over if they feel it was not a complete result.

BraveMole:
> Gamemode is the only one the can be re-voted on after reboot before the next round starts so that is why I gave it lowest priority, again any player can start the poll over if they feel it was not a complete result.

This will very rarely happen. People don't tend to re vote very often, because it feels like disrespecting the previous vote

Wayland-Smithy:
> This will very rarely happen. People don't tend to re vote very often, because it feels like disrespecting the previous vote

That's just like your opinion man. Personally I don't think anyone is going to complain about that poll with all of 2 votes cast, or needing to click on Crash again.

TiviPlus (approved):

PR #6839, opened 2021-05-12
Fix me

BraveMole:
## About The Pull Request

Me is broke, this fix me. This is a me problem cause by me, so me fixed it. 

More seriously, me is an emote with the key ="me", and it was a duplicate key with the keybind me. This broke me, because only the latter keybind (the emote, that shouldn't be called directly) was showed on screen

## Changelog
:cl:
fix: Me is fixed by me cause it was broken by me.
/:cl:


usnpeepoo:
u need fixing?

TiviPlus:
can you explain how it fixes it thnx

SplinterGP:
i wont fix you

code/modules/mob/emote.dm
```diff
@@ -51,9 +51,10 @@
 
 
 /datum/emote/custom
-	key = "me"
+	key = "emote_me"
```
TiviPlus, line 54: Explain this change?

TiviPlus (changes requested):

TiviPlus (approved):

PR #6848, opened 2021-05-13
FF is gone[DNM]

BraveMole:
## About The Pull Request

Marines have a bullet armor of 100. That means no bullet damage, but they still block the bullets so it's not full IFF.

It's just a funny test, no merge.

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -196,7 +196,7 @@
 		slot_l_hand_str = 'icons/mob/items_lefthand_1.dmi',
 		slot_r_hand_str = 'icons/mob/items_righthand_1.dmi',)
 	icon_state = "helmet"
-	soft_armor = list("melee" = 50, "bullet" = 100, "laser" = 50, "energy" = 50, "bomb" = 50, "bio" = 50, "rad" = 50, "fire" = 50, "acid" = 50)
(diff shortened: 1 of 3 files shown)


Tupinambis:
Good thing I still have my trusty high explosives to FF with!

Losenis:
Isn't the 100 affected by ammo penetration?

SpaceLoveSs13:
> Isn't the 100 affected by ammo penetration?

Yes he talked about it, AP still has same affects.

BraveMole:
> Isn't the 100 affected by ammo penetration?

It is. It's not perfect anti-ff system, it's just a test to see if it has potential.

usnpeepoo:
biggest marine buff pr ever do not ever merge this

InTroubleWh:
no, even the TMR is painful *scream

Losenis:
Should probably up the helmet armor too, given nothing modifies it so it stays at 0 and gets damage through by anything that has more than zero pen, which leads to shrap too

lKiseki:
Very cool PR, but this practically invalidates the balance on the new terra laser guns. Those may have to be looked at if this goes through; right now, with this TM and the new guns, I'd say marines are a little bit overpowered, but I'd like to reiterate it's not the TM's fault, but rather a combination of two things.

TiviPlus:
I saw enough

PR #6849, opened 2021-05-13
Fix hair choice on new character

BraveMole:
## About The Pull Request

The species was wrongly set on create_random_character, and it prevented hair choices. This is the fix

## Why It's Good For The Game

We missed grad hair

## Changelog
:cl:
fix: You can chose your hair now
/:cl:

### DIFF
--- a/code/modules/client/preferences_ui.dm
+++ b/code/modules/client/preferences_ui.dm
@@ -30,7 +30,7 @@
 		"age" = age,
 		"gender" = gender,
 		"ethnicity" = ethnicity,
-		"species" = species || "Human", //Sometimes species is null on creation of random char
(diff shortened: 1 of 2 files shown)


code/modules/client/preferences_ui.dm
```diff
@@ -30,7 +30,7 @@
 		"age" = age,
 		"gender" = gender,
 		"ethnicity" = ethnicity,
-		"species" = species || "Human", //Sometimes species is null on creation of random char
+		"species" = species,
```
TiviPlus, line 33: Explain this change? Why's it not just a single |

TiviPlus (approved):

PR #6852, opened 2021-05-13
Runtime new perm

BraveMole:
## About The Pull Request

Runtimes have their own perm

## Changelog
:cl:
admin: Runtimes have their own perm
/:cl:


code/modules/admin/admin_ranks.dm
```diff
@@ -66,6 +66,8 @@
 				flag = R_SERVER
 			if("DEBUG")
 				flag = R_DEBUG
+			if("CONTRIBUTOR")
```
korpow, line 69: Just call it RUNTIMES, all the others match

code/modules/admin/debug_verbs.dm
```diff
@@ -438,7 +438,7 @@ GLOBAL_PROTECT(AdminProcCallSpamPrevention)
 	set name = "View Runtimes"
 	set desc = "Open the runtime Viewer"
 
-	if(!check_rights(R_DEBUG))
+	if(!check_rights(R_RUNTIME|R_DEBUG))
 		return
```
TiviPlus, line 442: Does this not check both ie shouldn't this be two check rights checks

Wayland-Smithy: I had the same concern, but it does indeed check for any like:

https://github.com/tgstation/TerraGov-Marine-Corps/blob/9fe6280ef69d7144135b281f26d8563ba454aaaa/code/modules/admin/admin_verbs.dm#L496

TiviPlus (approved):

PR #6853, opened 2021-05-13
Fix R_COLOR for Contributor rank w/o OOC color (like the disc role)

Wayland-Smithy:
## About The Pull Request

Fixed the COLOR perm so it can be omitted from ranks without defaulting them to brown and spamming permission notices. Makes it possible to toggle OOC color for Contributors so David will not ask for this later if/when we actually start using the rank.

Related to #6852 to aid in the quick deployment of the new perm on an old rank.

## Why It's Good For The Game

Less headache for leads I guess?

## Changelog
:cl:
admin: Fixed admin COLOR perm.
/:cl:


DerrickChamberlain:
Yikes, poor contributors.

TiviPlus (changes requested):
I need to check config for this

TiviPlus (approved):

PR #6860, opened 2021-05-14
Laser gun rebalance and edits

David-Stormwell:
## About The Pull Request

Title

## Why It's Good For The Game

Due to player complaints and the awful timing of other game-changing balancing PRs, nerfs the rifle via giving it a 0.5 increase in fire rate, overcharge as well nerfed to compensate. Additionally, the carbine has a 0.1 increase in burst fire rate and a 0.5 increase in automatic.
Also decreases the max range to 30, and gives all projectiles other than the sniper 0.2-3 damage falloff multiplier.
Sniper gets a slight buff with aim mode and heat mode being the same fire rate as normal but has a 1.5- reduction in speed.
Also, has some additional edits to descriptions and a small bit of white gone from the sniper's barrel that was missed in the first PR.

## Changelog
:cl:
balance: rebalanced lasers to have falloff damage and increase/decrease in fire rate accordingly, as well as sniper now having aimmode.
fix: fixes some white on the sniper's barrel as well as few other small fixes to code that I missed in my first laser PR.
/:cl:


code/modules/projectiles/guns/energy.dm
```diff
@@ -742,13 +742,13 @@
 	wield_delay = 0.3 SECONDS
 	scatter = 0
 	scatter_unwielded = 15
-	fire_delay = 0.15 SECONDS
+	fire_delay = 0.17 SECONDS
```
BraveMole, line 745: doesn't exist

code/modules/projectiles/guns/energy.dm
```diff
@@ -759,7 +759,7 @@
 /datum/lasrifle/base/energy_carbine_mode/auto_burst_standard ///I know this seems tacky, but if I make auto burst a standard firemode it somehow buffs spread's fire delay.
 	charge_cost = 15
 	ammo = /datum/ammo/energy/lasgun/marine
-	fire_delay = 0.15 SECONDS
+	fire_delay = 0.17 SECONDS
```
BraveMole, line 762: doesn't exist

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1358,8 +1358,8 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	damage = 20
 	penetration = 10
 	sundering = 1
-	max_range = 40
-	damage_falloff = 0
+	max_range = 30
+	damage_falloff = 0.2 ///Doesn't seem to even apply when fired out of the laser gun.
```
BraveMole, line 1362: then don't change it? You probably have to change it for each fire mode datum of the lasrifle

BraveMole:
I'd say the worst offender of lasers is the machine gun one. The FF pr make that worste but even without, it's a pretty broken gun :
The ammo that can fit everywhere is a big plus compared to GPMG
The reload time is instant, compared to GPMG
The stats looks better than GPMG one.

GPMG is already kinda broke, and this one is worst. I just don't see the point of taking the standard lasrifle over the machinegun one.

Kuro-ai:
david rof only works in increments of 0.05, 0.17 rof just goes to 0.2.
Anyways I think you should just make the rof of the guns 0.2 all around and decrease the MG bullet count to 150

code/modules/projectiles/ammo_datums.dm
David-Stormwell: No, just the base is enough which I already did in the PR.

David-Stormwell:
> 
> 
> david rof only works in increments of 0.05, 0.17 rof just goes to 0.2.
> Anyways I think you should just make the rof of the guns 0.2 all around and decrease the MG bullet count to 150

Done.

David-Stormwell:
Ready to TM if maintainers would approve it.

TiviPlus:
If you want it dnm then put [dnm] in the title

Pixelknightly:
After a bit of testing and playing around with it. It kinda seems like the carbine's burst is far weaker than it was previously. It's actually faster to just use auto with it now.

And looking at the files changed... I can see why that is.
I get slowing down the burst. But when auto is faster than the burst. It might have been nerfed a bit too much. (Though I guess this is obvious in hindsight since you made the burst firing speed match the auto.)
The normal rifle isn't much better either, since it's basically a worse version of the MG with less ammo now. (Haven't tested the overcharge on it, but considering that got nerfed too. It might also be bad.)

I see this as less of a reason to use the laser rifles, which already had less ammo than the normal rifles and less damage on top of that. Scale back the changes on the rifles a bit? Because even with the nerfs to the laser MG. It's still the better option, especially when the rifles had their ROF nerfed. (Which was the only thing making them better than the MG in some ways.)

David-Stormwell:
Applied some MG nerfs for things I saw while playing the other day.

David-Stormwell:
So far, it's looking fine.
If it goes on being fine enough feel free to merge this

SplinterGP (approved):

BraveMole:
Possible to fix the sounds of lasers?
Sniper one is a deafening bass, and the rest is far too loud

David-Stormwell:
Turn your volume down.
We've had a few other loud guns and they're fine, I can see overcharge being a problem but not much I can do with audacity.

TiviPlus:
> Turn your volume down.
> We've had a few other loud guns and they're fine, I can see overcharge being a problem but not much I can do with audacity.

You should be able to change that with code

TiviPlus (approved):
Did not read

PR #6873, opened 2021-05-16
Fix req

BraveMole:
## About The Pull Request

Cleaning up my mess. 
Also remove dropship faction based points, as it is useless since i won't be using it in HvH mode

## Changelog
:cl:
fix: Req fixed again
/:cl:

### DIFF
--- a/code/controllers/subsystem/points.dm
+++ b/code/controllers/subsystem/points.dm
@@ -11,7 +11,6 @@ SUBSYSTEM_DEF(points)
 	flags = SS_KEEP_TIMING
 
 	wait = 10 SECONDS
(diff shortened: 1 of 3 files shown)


code/controllers/subsystem/points.dm
```diff
@@ -11,7 +11,6 @@ SUBSYSTEM_DEF(points)
 	flags = SS_KEEP_TIMING
 
 	wait = 10 SECONDS
-	///Assoc list of dropship points
 	var/dropship_points = list()
```
spookydonut, line 14:
```suggestion
	var/dropship_points = 0
```

PR #6882, opened 2021-05-16
Prevent restocking of infinite supplies

BraveMole:
## About The Pull Request

Restocking an item with an infinite supply will no longer make the amount of item increment.

This does not fix the stock showing even for negative stocks, as it is working fine on local serb?

## Changelog
:cl:
fix: No longer can break infinite supplies by restocking
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/vending.dm
+++ b/code/game/objects/machinery/vending/vending.dm
@@ -669,7 +669,8 @@
 		if(!recharge)
 			user.visible_message("<span class='notice'>[user] stocks [src] with \a [R.product_name].</span>",
 			"<span class='notice'>You stock [src] with \a [R.product_name].</span>")
-		R.amount++
+		if(R >= 0) //R negative means infinite item, no need to restock
+			R.amount++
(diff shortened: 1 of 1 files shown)


jroinc:
Does this mean I can vend a billion things into a pile that crashes someone if they left-click, and can't be easily fixed cause it's on a vendor tile?

TiviPlus (approved):

Losenis:
That'd probably require admeme roomba regardless given restocking a billion things one by one would be nuts in the first place

BraveMole:
> Does this mean I can vend a billion things into a pile that crashes someone if they left-click, and can't be easily fixed cause it's on a vendor tile?

You could always do that? 
There was 10k ammo in the vendor, if you manage to buy 2 per seconds, it's 1h20 of clicking lel

jroinc:
This appears to prevent all restocking.

Losenis:
This kills medvendor restocking, which is kind of an immense blow to medical and healing

PR #6886, opened 2021-05-17
Minitadilicious

Swagile:
## About The Pull Request

image

Makes it as small as possible while being as functional as possible.

Also adds (1) and (2) extra sentry gun emplacements (the ones you can put on DS / tadpole) to PoS and Sulaco respectively. Fixes the issue of not being able to properly equip tadpole (old tad only had 2 sentry emplacements so this was never an issue, 2 is now mandatory and 4 is optimal).

## Why It's Good For The Game

The goal of tadpole was to be a rapid response combat bus / medevac but it was not possible to serve this purpose. After playing as PO, SL, and FC and using the current tadpole extensively, the issue is... the tadpole is too big and janky. The front of tadpole is hard to defend due to windows being wankily placed so its easy for the turrets to hit the windows themselves while trying to defend tad, its hard to manuever around front of tad to protect it due to the way the windows are positioned, and its hard to place it around cover which makes it easily crusher / acided from afar without much resistance put up.

This fixes the issue of tad being too big and bad window placement at front which now allows it to be a rapid response aircraft. Heal bus is still possible with this setup (though a bit cramped), and tadops will be a true riot to try and wrangle in this beastie.

Last but not least, even if it looks bad, I can tell you at least 3 different ways to properly compensate for the weaknesses of this new design; you just need to be a smart marine (hint: t-gas is one method).

Free method (won't tell you how to do it except that someone else taught me how to do this):

## Changelog
:cl:
balance: Tadpole has been made into a true Clowncar, Boxpol, minitad.
balance: Extra turrets given to the TGMC for free to properly outfit either the DS on all sides or tadpole on all sides.
/:cl:

### DIFF
--- a/_maps/shuttles/minidropship.dmm
+++ b/_maps/shuttles/minidropship.dmm
@@ -1,271 +1,229 @@
 //MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
 "a" = (
-/obj/structure/shuttle/engine/propulsion/burst/right,
-/turf/open/shuttle/dropship/three,
-/area/shuttle/minidropship)
-"b" = (
 /obj/structure/window/framed/mainship/spaceworthy{
 	icon_state = "col_window10"
 	},
 /turf/open/floor/mainship_hull,
 /area/shuttle/minidropship)
-"e" = (
-/obj/structure/barricade/plasteel{
-	dir = 4
-	},
+"b" = (
+/obj/effect/attach_point/weapon/minidropship,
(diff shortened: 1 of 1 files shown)


AnCopper:
smoll

TiviPlus (approved):

Losenis:
Should update the picture with the current iteration that has that lower cadeline lowered

but otherwise **YES**

Swagile:
> 
> 
> Should update the picture with the current iteration that has that lower cadeline lowered
> 
> but otherwise **YES**

Its in Map Diff Renders at the very bottom; shows old and new as well.

lovegreenstuff:
No nanomed.

Downvote.

PR #6916, opened 2021-05-21
Friendly fire damage is reduced

BraveMole:
## About The Pull Request

Added a new var on projectile, the friendly_fire_multiplier var. This will reduce the damage from FF

## Why It's Good For The Game

FF is a terrible mechanic because it punishes the victim of the bad player that just sprayed around.
This is an easy solution that has the benefits of not touching to HvH, and not affecting the HvX balance too much.

I've chose a var on the projectile rather than a global var cause MJP asked to be able to balance for each gun

## Changelog
:cl:
balance: Damage from friendly fire is reduced a tiny bit
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -90,7 +90,7 @@
 
 	var/proj_max_range = 30
 	///A damage multiplier applied when a mob from the same faction as the projectile firer is hit
-	var/friendly_fire_multiplier = 0.3
+	var/friendly_fire_multiplier = 0.5
 
 
 /obj/projectile/Destroy()
@@ -747,7 +747,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		var/mob/living/carbon/human/shooter_human = proj.firer
 		if(shooter_human.faction == faction || m_intent == MOVE_INTENT_WALK)
 			. -= 15
-		///Friendly fire does less damage
+		//Friendly fire does less damage
 		if(shooter_human.faction == faction)
 			proj.damage *= proj.friendly_fire_multiplier
 	return ..()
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -89,6 +89,8 @@
 	var/y_offset
 
 	var/proj_max_range = 30
+	///A damage multiplier applied when a mob from the same faction as the projectil is hit
```
BraveMole, line 92:
```suggestion
	///A damage multiplier applied when a mob from the same faction as the projectile firer is hit
```

Losenis:
SPEEDMERGE WHEN

TiviPlus (commented):
Dont put the number in the pr to keep marines from shooting each other
Also 70 is a lot lets try 50 or 40 first

usnpeepoo:
nooo how am i going to shuffle marines into ff and expect them to die moments later??

jroinc:
This is probably the wrong way to do this, but it's *a* way to do this. Let's see what it does.

EgorDinamit:
FF is literally one of the only available ways for xenos(mainly runners) to kill marines. Marine powercreep PR.

/s

Isy232:
> FF is literally one of the only available ways for xenos(mainly runners) to kill marines. Marine powercreep PR.
> 
> /s

Then buff xenos in other ways or something.

Even when the no FF PR was merged the marines only won a few times, mostly due to xenos making massive positioning mistakes and getting trapped.

BraveMole:
> This is probably the wrong way to do this, but it's _a_ way to do this. Let's see what it does.

Yeah i would prefer to just remove frac and IB from bullets but it's much harder to do cleanly

SplinterGP:
what about instead making the bullets not apply AP on human mobs ? that way the armor plays a bigger part of it, since the biggest trouble, specially with snipers is that they have a high AP which negates most of marine armor

PR #6920, opened 2021-05-21
Increase T-12 Magazine Capacity From 50 to 60

Tupinambis:
## About The Pull Request

Increases T-12 Magazine Capacity from 50 -> 60
Additionally increases the 10x24 ammo box capacity from 150 -> 180 to fit the capacity increase.

## Why It's Good For The Game

Lets the T-12 last longer in the field and be a slightly better "generalist" weapon. Intended to make the T-12 more appealing when compared to other weapons without increasing its moment to moment power.

Additionally this makes the 3-round burst fire setting actually fully drain the weapon instead of having an awkward 2-shot burst at the end.

## Changelog
:cl: Tupina
balance: T-12 magazine capacity increased from 50 -> 60
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -114,7 +114,7 @@
 	unload_sound = 'sound/weapons/guns/interact/t18_unload.ogg'
 	reload_sound = 'sound/weapons/guns/interact/t18_reload.ogg'
 	caliber = CALIBER_10X24_CASELESS //codex
-	max_shells = 50 //codex
+	max_shells = 60 //codex
 	force = 20
 	current_mag = /obj/item/ammo_magazine/rifle/standard_assaultrifle
 	attachable_allowed = list(
--- a/code/modules/projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/magazines/rifles.dm
@@ -63,19 +63,19 @@
 	icon_state = "t12"
 	w_class = WEIGHT_CLASS_NORMAL
 	default_ammo = /datum/ammo/bullet/rifle
-	max_rounds = 50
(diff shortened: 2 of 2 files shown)


Losenis:
I just noticed how it'd also fit nicely with the BFA given 60 is a multiple of both 3 and 5. Huh.

Walarks:
Balance-wise, does the T-12 need this? I personally think it is in an ok spot. 

Buffs would indeed make a weapon more appealing, but this needs to be considering both the marine and xeno side of things. Over time small buffs can stack and cause problems later.

Naaanii:
> Balance-wise, does the T-12 need this? I personally think it is in an ok spot.
> 
> Buffs would indeed make a weapon more appealing, but this needs to be considering both the marine and xeno side of things. Over time small buffs can stack and cause problems later.

Just nerf it harder later :^)

Losenis:
> Balance-wise, does the T-12 need this? I personally think it is in an ok spot.
> 
> Buffs would indeed make a weapon more appealing, but this needs to be considering both the marine and xeno side of things. Over time small buffs can stack and cause problems later.

Given it competes with the T42 which you can slap a backpack full of drums + G8 + jaeger general storage that will have you effectively never run out of ammo, and how T12 mags are as big as drums in terms of backpack space, personally it'd give me more confidence in not running out of them unless I'm straight up using webbing 2 magpouches and even the satchel as a second ammo rig and carrying around an ammo box of the damn mags.

Although that's more of an issue with mags being so stupidly big.

David-Stormwell:
T-12 is already fine, it's in a spot where it isn't bad, isn't great and can be used to chase or otherwise.

TiviPlus:
I dont feel this is needed

PR #6927, opened 2021-05-22
T19 buff

RenGusta:
makes it gooder one handed because gohunt says so, ups rof to T90 standard for SMGs

## About The Pull Request

buffs T19 ROF and reduces scatter to improve one handed use

## Why It's Good For The Game

Gohunt thinks it's good for the game, he's incapable of doing this, so I'll do it for him

## Changelog
:cl:
balance: buffs T19 ROF and scatter
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smgs.dm
+++ b/code/modules/projectiles/guns/smgs.dm
@@ -77,7 +77,7 @@
 	recoil_unwielded = 0
 	scatter = 0
 	fire_delay = 0.15 SECONDS
-	scatter_unwielded = 0 //Made to be better used one handed.
+	scatter_unwielded = 5 //Made to be better used one handed.
 	aim_slowdown = 0.15
 	burst_amount = 5
 	movement_acc_penalty_mult = 0
--- a/code/modules/projectiles/guns/smgs.dm
+++ b/code/modules/projectiles/guns/smgs.dm
@@ -76,8 +76,8 @@
 	accuracy_mult_unwielded = 0.85
 	recoil_unwielded = 0
 	scatter = 0
-	fire_delay = 0.2 SECONDS
-	scatter_unwielded = 10 //Made to be used one handed.
+	fire_delay = 0.15 SECONDS
+	scatter_unwielded = 0 //Made to be better used one handed.
 	aim_slowdown = 0.15
 	burst_amount = 5
 	movement_acc_penalty_mult = 0


David-Stormwell:
This will buff akimbo T19s considerably, of course I expect you understood that?

gohunt001:
> This will buff akimbo T19s considerably, of course I expect you understood that?

Of course, although preferably the amount of scatter on akimbo probably should be increased, it shouldn't be too large as to make akimbo use of the weapon completely inviable. Considering that, with this fire rate buff, the flat, non-akimbo DPS will be the same as the T90(2x that for akimbo use), and on akimbo use the per-mag total possible damage would be similar to the T90 due to holding 2x30 rounds, adding some scatter would help to cut down the hit chance and make it viable, but not overpowered to the point it surpasses the T90 in efficacy.

Problem being, of course, i have no clue what var to edit to increase akimbo scatter only, as in testing, even changing the unwielded scatter modifier to 0.1% causes the weapon to scatter as if it had 100% scatter chance, and i fear that whatever is causing that might also affect the akimbo scatter var, giving the weapon 100% scatter and making akimbo use inviable, which we don't want.

Tupinambis:
T-19 is currently useless by most metrics so I'm glad to see it get a bit of love.

This shouldn't be much more powerful than the TP-14 dual wielded. TP-14 has a fire delay of 0.125 Seconds click spammed versus the T-19s new 0.15 seconds automatic, but T-19 has 0.5 sunder versus the TP-14s considerably higher 1.65.

So from a DPS standpoint the TP-14 remains king with the T-19s selling point being a less tired finger and 9 more bullets to shoot xenos with per magazine. Scatter would also now be identical between the two pistols.

So realistically dual wielded T-19s might become viable but it's unlikely to become the new meta when the TP-14 wins in raw DPS and sunder otherwise.

code/modules/projectiles/guns/smgs.dm
```diff
@@ -76,8 +76,8 @@
 	accuracy_mult_unwielded = 0.85
 	recoil_unwielded = 0
 	scatter = 0
-	fire_delay = 0.2 SECONDS
-	scatter_unwielded = 10 //Made to be used one handed.
+	fire_delay = 0.15 SECONDS
+	scatter_unwielded = 0 //Made to be better used one handed.
```
SplinterGP, line 80: i think 5 is a fair number, you can still use suppressor/recoil compensator with it so its bearable
```suggestion
	scatter_unwielded = 5 //Made to be better used one handed.
```

SplinterGP (changes requested):

code/modules/projectiles/guns/smgs.dm
gohunt001: A problem encountered during testing using simple ingame varediting was that any unwielded scatter above 0 would cause the scatter to act like it was 100% scatter chance for some heretofore unknown reason, perhaps there was a var that made that so, but if there is, we don't know which one it is. Or perhaps it was a problem connected with ingame varediting?

SplinterGP (approved):

PR #6937, opened 2021-05-23
Fixed Movement Bug With Evasion

Surrealaser:
## About The Pull Request

Evasion now fails as it should if the runner is targeted by flamethrower projectiles, or hasn't moved in the past 0.5 seconds.

## Why It's Good For The Game

Evasion now fails as it should if the runner is targeted by flamethrower projectiles, or hasn't moved in the past 0.5 seconds.

## Changelog
:cl:
fix: Evasion now fails as it should if the runner is targeted by flamethrower projectiles, or hasn't moved in the past 0.5 seconds.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
@@ -322,23 +322,23 @@
 
 	var/mob/living/carbon/xenomorph/runner/R = owner
 	if(!evade_active) //If evasion is not active we don't dodge
-		return TRUE
+		return FALSE
 
-	if((R.last_move_time  < (world.time - RUNNER_EVASION_RUN_DELAY))) //Gotta keep moving to benefit from evasion!
-		return TRUE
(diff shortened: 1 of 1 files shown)


SplinterGP (approved):

Wayland-Smithy:
> Evasion now fails as it should if the runner is targeted by flamethrower projectiles
flamethrower projectiles
`p r o j e c t i l e s`

As you point out the flames act as a slow projectile when travelling, so why should they not be dodged unlike a speedy boolet?

Surrealaser:
> > Evasion now fails as it should if the runner is targeted by flamethrower projectiles
> > flamethrower projectiles
> > `p r o j e c t i l e s`
> 
> As you point out the flames act as a slow projectile when travelling, so why should they not be dodged unlike a speedy boolet?

Because it's an abstraction of a large undodgeable wall of fire.

Wayland-Smithy:
IMO they should he able to juke the "wall of fire" which is actually a slow moving stream, and upon being ignited like from crossing a tile with fire their evasion should end.

SplinterGP:
> 
> 
> IMO they should he able to juke the "wall of fire" which is actually a slow moving stream, and upon being ignited like from crossing a tile with fire their evasion should end.

the bullet from the flamethrower signifies a direct hit, so i feel it should hurt

PR #6944, opened 2021-05-25
marines have less buffer, moved to overall staminaloss resist

Pariah919:
## About The Pull Request
-20 buffer +20 overall pooled resist
use only with #6872 

## Why It's Good For The Game
asked for, makes sense

## Changelog
:cl:
balance: marines have less buffer before being made to walk
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -45,7 +45,7 @@
 	var/exhale_type = "carbon_dioxide"      // Exhaled gas type.
 
 	var/total_health = 100  //new maxHealth
-	var/max_stamina_buffer = 50
+	var/max_stamina_buffer = 30
 
 	var/cold_level_1 = BODYTEMP_COLD_DAMAGE_LIMIT_ONE  	// Cold damage level 1 below this point.
 	var/cold_level_2 = BODYTEMP_COLD_DAMAGE_LIMIT_TWO  	// Cold damage level 2 below this point.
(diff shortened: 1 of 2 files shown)


JackTheJackhammer:
Whats a buffer?

lKiseki:
Good PR. With this and the PR that removes the need for stamina, at least for the purposes of running, there should be some semblance of balance what with tools like neurotoxin spit and tail sweep gaining added importance due to being able to stop marines from running. Slowdowns, which already existed and can be perceived in various other abilities, will also gain increased importance, which is quite okay in my opinion.

Bring back #6872 alongside this for some extensive testing, see how it fares, is what I'd suggest.

Wayland-Smithy:
Stamcode has me a bit confused. Since status effects seem to decrease the buffer by a percentage this means tox stam "DPS" is unaffected by this change? 3 sec debounce is kek also.
```dm
/mob/living/proc/handle_staminaloss()
	if(world.time < last_staminaloss_dmg + 3 SECONDS)
		return
	if(staminaloss > 0)
		adjustStaminaLoss(-maxHealth * 0.2, TRUE, FALSE)
	else if(staminaloss > -max_stamina_buffer)
		adjustStaminaLoss(-max_stamina_buffer * 0.08, TRUE, FALSE)
```

TiviPlus:
if your pr cant stand on its own dont split it use this pr

Pariah919:
@TiviPlus splinter told me to split it into another pr so i split it

David-Stormwell:
> if your pr cant stand on its own dont split it use this pr

Either reopen the other PR or this one mixed in, though I'd suggest it be seperated for better atomization and balancing.

BraveMole:
I don't even know if this is really needed, the stamina run PR is fine on its own. 
Playing against it as a xeno doesn't feel too oppressive, xeno already have a lot of tools to escape anyway.
And playing as a marine, it just feels better

TiviPlus:
Put it in the other Pr why you shouldn't split it was told you by myself and spooky on discord

PR #6953, opened 2021-05-26
Miner points is linked to pop + overall miner buff

BraveMole:
## About The Pull Request

The amount of supply points a miner give will be linear with pop.
With this calibration, one phoron shipment gives 0.5 points for each active player.

Overall, it's a miner buff

## Why It's Good For The Game

Always good to link balance things with pop.
I also want to buff miner, because marines really have issues to keep up with xeno in midgame, even when they manage to have a ton of miners.

IMO, if marines manage to keep 5+ miners, they should definitly win most of the time. And this is not the case, in part because marines level is bad (FF with sadar...), but also because it's impossible to equip properly marines with better gear than at start.

Marines need a better force scale factor, and this is far from being enough

## Changelog
:cl:
balance: Miners points output is linked with active player pop
balance: Overall, miners are producing more in mid pop (40 players)
/:cl:

### DIFF
--- a/code/game/objects/machinery/miner.dm
+++ b/code/game/objects/machinery/miner.dm
@@ -10,8 +10,8 @@
 #define MINER_RESISTANT	"reinforced components"
 #define MINER_OVERCLOCKED "high-efficiency drill"
 
-#define PHORON_CRATE_SELL_AMOUNT 15
-#define PLATINUM_CRATE_SELL_AMOUNT 30
+#define PHORON_CRATE_SELL_AMOUNT 0.5
+#define PLATINUM_CRATE_SELL_AMOUNT 1
 
 ///Resource generator that produces a certain material that can be repaired by marines and attacked by xenos, Intended as an objective for marines to play towards to get more req gear
 /obj/machinery/miner
@@ -93,7 +93,7 @@
 			required_ticks = 60
 		if(MINER_AUTOMATED)
 			if(stored_mineral)
-				SSpoints.supply_points[faction] += mineral_value * stored_mineral
+				SSpoints.supply_points[faction] += mineral_value * stored_mineral * get_active_player_count()
 				do_sparks(5, TRUE, src)
 				playsound(loc,'sound/effects/phasein.ogg', 50, FALSE)
 				say("Ore shipment has been sold for [mineral_value * stored_mineral] points.")
@@ -245,7 +245,7 @@
 		to_chat(user, "<span class='warning'>[src] is not ready to produce a shipment yet!</span>")
 		return
 
-	SSpoints.supply_points[faction] += mineral_value * stored_mineral
+	SSpoints.supply_points[faction] += mineral_value * stored_mineral * get_active_player_count()
 	do_sparks(5, TRUE, src)
 	playsound(loc,'sound/effects/phasein.ogg', 50, FALSE)
 	say("Ore shipment has been sold for [mineral_value * stored_mineral] points.")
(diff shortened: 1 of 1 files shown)


EvilMelonTheSecond:
Besides the low point income limiting marines to not get new equipment/upgrades (except a few SADARs) most of the time marines suffer from shitty supply lines. Even if marines stay in the FOB, RO can't receive orders from groundside marines unless shouted in radio (Which has a 99% chance of not being seen and the remaining 1% gets ignored due to not being a SADAR order). The supply drop is kinda slow and FOB tends to get cluttered real fast causing new drops to be buried and lost. Most of the veteran marines never bother with req once deploying unless they get medevaced.

I'm happy that some sort change is being made to req points but I've seen many rounds where 500+ points never get used due to a lack of demand and the fact that it's simply impossible to send new gear to frontline forces since they cannot change their storage options to fit the new gear.

This is easily the weakest part of marine gameplay

SplinterGP:
i don't generally like this because i feel like your points are supposed to scale by pop by getting more miners, and supposedly having more marines = go get more miners.

Wayland-Smithy:
Wow mole, another sneaky `get_active_player_count()` implementation that does not pass the "alive only" argument, so as xeno and marine numbers start to dwindle the income stays high. Just like how you made small silos stay locked out forever from dead pop. Nice meme.

ThatGuyFromThatPlace:
> Wow mole, another sneaky `get_active_player_count()` implementation that does not pass the "alive only" argument, so as xeno and marine numbers start to dwindle the income stays high. Just like how you made small silos stay locked out forever from dead pop. Nice meme.

...Why should it actively lower as you kill xenos and lose marines? That would be kinda retarded

Wayland-Smithy:
> ...Why should it actively lower as you kill xenos and lose marines? That would be kinda retarded

You are kidding right? So you want to reward heavy marines losses in the first push by letting them go "well we didn't need them anyway our miners still produce extra because they were ever alive."?

BraveMole:
> You are kidding right? So you want to reward heavy marines losses in the first push by letting them go "well we didn't need them anyway our miners still produce extra because they were ever alive."?

Rewarding =/ not punishing them?

BraveMole:
> i don't generally like this because i feel like your points are supposed to scale by pop by getting more miners, and supposedly having more marines = go get more miners.

I don't think this holds true in practice. In low pop, you just need one dedicated engineer to fix all miners, and xeno will not have the map control to deny them.

In high pop, there will probably be a lot of runners/ wraiths patrolling and killing miners.

Most of the time, the only miners marines manage to have are the first few miners near FOB (and most of the time, only one), and this is true for all pop size

Wayland-Smithy:
> Rewarding =/ not punishing them?

Your own CL says "active player pop" are observers active players to you? If there are 10 marines, 4 xenos, and 30 observers they should be earning 44 pop worth of miner points? kek?

BraveMole:
> Your own CL says "active player pop" are observers active players to you? If there are 10 marines, 4 xenos, and 30 observers they should be earning 44 pop worth of miner points? Passive psy points should work the same way.

The proc is named active_player_count(), if you don't consider dead players active, even though they can be ressed or respawn the minute after, that's kinda your opinion

Wayland-Smithy:
> if you don't consider dead players active, even though they can be ressed or respawn the minute after, that's kinda your opinion

No. They will be counted the next time once they are revived.

..and the full proc is `/proc/get_active_player_count(alive_check = FALSE, afk_check = FALSE, faction_check = FALSE, faction = FACTION_NEUTRAL)`  so you need to specify to ignore dead people.

SplinterGP:
> > i don't generally like this because i feel like your points are supposed to scale by pop by getting more miners, and supposedly having more marines = go get more miners.
> 
> I don't think this holds true in practice. In low pop, you just need one dedicated engineer to fix all miners, and xeno will not have the map control to deny them.
> 
> In high pop, there will probably be a lot of runners/ wraiths patrolling and killing miners.
> 
> Most of the time, the only miners marines manage to have are the first few miners near FOB (and most of the time, only one), and this is true for all pop size

Im pretty sure this is overall issue with other game mechanics, increasing pointer incomingg is not a good way to solve that since just some time ago we reduced it.

SplinterGP:
Plus the problem above

PR #6954, opened 2021-05-26
Miners are more durable

BraveMole:
## About The Pull Request

A 3 second windup is needed to damage miners. This makes them more durable, and way more easy to defend against wraith or fast runners

## Why It's Good For The Game

Miner integrity system was never designed for wraith, so it was stupidly easy to kill miners with it witout taking any risks.

## Changelog
:cl:
balance: Miners now take time to disable
/:cl:

### DIFF
--- a/code/game/objects/machinery/miner.dm
+++ b/code/game/objects/machinery/miner.dm
@@ -280,17 +280,15 @@
 /obj/machinery/miner/attack_alien(mob/living/carbon/xenomorph/X, damage_amount = X.xeno_caste.melee_damage, damage_type = BRUTE, damage_flag = "", effects = TRUE, armor_penetration = 0, isrightclick = FALSE)
 	if(X.status_flags & INCORPOREAL) //Incorporeal xenos cannot attack physically.
 		return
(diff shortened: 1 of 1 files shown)


Wayland-Smithy:
3 seconds PER ATTACK?

How about like 5 seconds to completely reset it instead?

BraveMole:
> 3 seconds PER ATTACK?
> 
> How about like 5 seconds to completely reset it instead?

It's 12 seconds, it's fine

Wayland-Smithy:
Fine. I suppose multiple can do it at once to parallelize the task.

BraveMole:
> Fine. I suppose multiple can do it at once to parallelize the task.

No clue, i don't think the do_after is blocking

SplinterGP:
Why not increase health though?

BraveMole:
> Why not increase health though?

would mean changing the code for repairs, for the same effect

SplinterGP:
> 
> 
> > Why not increase health though?
> 
> would mean changing the code for repairs, for the same effect

ah i see

PR #6979, opened 2021-05-27
Adds acid spires system

Walarks:
## About The Pull Request
Adds a new system to xenos that revolves around acid spires. Acid spires are 2x2 buildings randomly placed groundside at the start of the round. They produce acid that is stored in a collective pool. 

Xenos can place acid nodes in exchange for acid to increase an acid spire's strength and area it affects. To connect a node to an acid spire, it needs to be placed within its effective range. Then the node increases the acid spire's effective range, allowing for nodes to be placed further away.

The amount of production strength each node provides to the acid spire is based on how far it is placed from the spire, up to 50% falloff when outside the effective range of a spire. When a node is destroyed, the spires it is connected to lose the range and strength buffs it provides, however nodes that end up being outside the effective range still remain connected and contribute production at 50% effectiveness. On the other hand, when a spire's range is increased, the strength falloff for each node is calculated again, meaning the strength provided by previously placed nodes is increased as they are proportionally more within the effective range of the spire.

WIP design doc: https://pastebin.com/D1Lj1Att

## Why It's Good For The Game
Adds a new system for xenos that they can play with. Xenos are incentivized to spread the acid nodes to avoid having the marines destroy all at once, meaning it can make the layout of the map progress in an interesting manner. On the other hand, the marines can effectively sabotage the acid production by destroying nodes or going for the spires and mining points from them.

## Changelog
:cl:
add: Added acid spires and acid nodes
/:cl:


ilikesaying:
what is the acid used for?

Walarks:
> what is the acid used for?

I'll add that when I'm ready to undraft the PR

Walarks:
This needs more work than I expected when I opened it initially. Might reopen/make another PR at a later point if I finish more features.

PR #6992, opened 2021-05-28
Improves Crushers abilities a little, removes stam damage

Wayland-Smithy:
## About The Pull Request

Buffs Crusher a little, see changelog. Feel free to suggest different numbers or argue why unga do not deserve this in the comments.

Fixes a missing `updating_health` arg.

Brings Crest Toss friendly fire mechanics in line with that of Warriors displacements.

Removes and/or converts Crest Toss and Stomp stam damage to brute.

Some Stomp discussion: https://discord.com/channels/498569646014857217/504094319075131402/848408671897124865

## Why It's Good For The Game

Brings stomp back from the desolation of overnerfdom. It should not be something you just quickly get up and run away from. It can actually be used for melee range brawls and escapes from chasing harvesters.

Why is crest toss such high cooldown and plasma cost anyway? It is supposed to counter cade huggers last I checked. The stam damage is on its way to useless, so having it available more often gives slow crusher more chances to use it for escape or saving allies.

TL;DR stam buffs will make most of this imperceptible anyway. Any extra throw distance or stun here is easily made up for with a little sprint, but hopefully crushers can have a little more fun dying to the 7 layer FOBurrito instead of a lone PAS flech shotgunner.

## Changelog
:cl:
fix: Fixed Crusher Stomp damage delay.
qol: Crusher Crest Toss now works on big allies with a penalty like Warrior Fling.
balance: Crusher Stomp briefly knocks down anyone adjacent to crusher for a half second. Victims directly under foot stun increased by one second.
balance: Crusher Crest Toss increases distance with age from 3 to 6 (still less than half max powerfist), and has less cooldown.
balance: Crest Toss and Stomp no longer deal stamina damage, only brute.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
@@ -40,9 +40,8 @@
 			step_away(M, X, 1) //Knock away
 			shake_camera(M, 2, 2)
 			to_chat(M, "<span class='highdanger'>You reel from the shockwave of [X]'s stomp!</span>")
+			M.take_overall_damage_armored(damage, BRUTE, "melee", FALSE, FALSE, TRUE) // give half damage brute instead of stamina
 			M.Paralyze(0.5 SECONDS)
-
-		M.apply_damage(damage, STAMINA, updating_health = TRUE) //Armour ignoring Stamina
 
 /datum/action/xeno_action/activable/stomp/ai_should_start_consider()
 	return TRUE
@@ -143,10 +142,9 @@
 
 	//Handle the damage
 	if(!X.issamexenohive(L)) //Friendly xenos don't take damage.
-		var/damage = toss_distance * 5
+		var/damage = toss_distance * 6
 
-		L.take_overall_damage_armored(damage, BRUTE, "melee")
-		L.apply_damage(damage, STAMINA, updating_health = TRUE) //...But decent armour ignoring Stamina
+		L.take_overall_damage_armored(damage, BRUTE, "melee", updating_health = TRUE)
 		shake_camera(L, 2, 2)
 		playsound(L,pick('sound/weapons/alien_claw_block.ogg','sound/weapons/alien_bite2.ogg'), 50, 1)
 
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
@@ -24,7 +24,7 @@
 	"<span class='xenodanger'>We smash into the ground!</span>")
 	X.create_stomp() //Adds the visual effect. Wom wom wom
 
-	for(var/mob/living/M in range(0,X.loc))
+	for(var/mob/living/M in range(1,X.loc))
 		if(isxeno(M) || M.stat == DEAD || isnestedhost(M))
 			continue
 		var/distance = get_dist(M, X)
@@ -32,18 +32,16 @@
 		if(distance == 0) //If we're on top of our victim, give him the full impact
 			GLOB.round_statistics.crusher_stomp_victims++
 			SSblackbox.record_feedback("tally", "round_statistics", 1, "crusher_stomp_victims")
-			M.take_overall_damage_armored(damage, BRUTE, "melee", FALSE, FALSE)
-			M.Paralyze(20)
+			M.take_overall_damage_armored(damage, BRUTE, "melee", FALSE, FALSE, TRUE)
+			M.Paralyze(3 SECONDS)
 			to_chat(M, "<span class='highdanger'>You are stomped on by [X]!</span>")
 			shake_camera(M, 3, 3)
 		else
 			step_away(M, X, 1) //Knock away
 			shake_camera(M, 2, 2)
 			to_chat(M, "<span class='highdanger'>You reel from the shockwave of [X]'s stomp!</span>")
-		if(distance < 2) //If we're beside or adjacent to the Crusher, we get knocked down.
-			M.Paralyze(20)
-		else
-			M.Stun(20) //Otherwise we just get stunned.
+			M.Paralyze(0.5 SECONDS)
+
 		M.apply_damage(damage, STAMINA, updating_health = TRUE) //Armour ignoring Stamina
 
 /datum/action/xeno_action/activable/stomp/ai_should_start_consider()
@@ -66,8 +64,8 @@
 	action_icon_state = "cresttoss"
 	mechanics_text = "Fling an adjacent target over and behind you. Also works over barricades."
 	ability_name = "crest toss"
-	plasma_cost = 100
-	cooldown_timer = 18 SECONDS
+	plasma_cost = 60
+	cooldown_timer = 10 SECONDS
 	keybind_signal = COMSIG_XENOABILITY_CRESTTOSS
 	target_flags = XABB_MOB_TARGET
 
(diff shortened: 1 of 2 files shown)


BraveMole:
Buffing xeno? NOT ON MY TGMC

InterroLouis:
I'm gonna say no, I don't want that last bit. We don't need to buff the overpowered forced movement any more than it already is.

Wayland-Smithy:
Forced movement bad huh? Tell that to powerfist.

What about a weed speed increase for crusher instead?

InterroLouis:
It's far easier for xenos to ruin people with forced movement, especially since they're faster, can't be bodyblocked, and can easily bodyblock anyone by swapping and then switching from help intent.

Wayland-Smithy:
> especially since they're faster

what??

> and can easily bodyblock anyone by swapping and then switching from help intent.

sure, if you are in a 1 wide hallway and don't know other directions exist, this PR has nothing to do with that at all.

SplinterGP:
Marines are faster only when undamaged, once they get damaged they can be way way way slower

Wayland-Smithy:
Does that mean the code looks good?

InterroLouis:
> Marines are faster only when undamaged, once they get damaged they can be way way way slower

Plus taking stamina damage past the buffer makes you unable to run.

Wayland-Smithy:
You know what the buffer does right? Because I am pretty sure it is just that a "buffer" of **extra** stamina that you have just for absorbing damage before it impact your actual stamina bar which you no longer need for running anyway. That's 150 effective auto regening stamina a single crusher is not going to do dip to.

Wayland-Smithy:
I'd be perfectly happy replacing all the stam damage calls with brute too.

ilikesaying:
> It's far easier for xenos to ruin people with forced movement, especially since they're faster

crusher is the fucking slowest caste when not charging

SplinterGP:
> You know what the buffer does right? Because I am pretty sure it is just that a "buffer" of **extra** stamina that you have just for absorbing damage before it impact your actual stamina bar which you no longer need for running anyway. That's 150 effective auto regening stamina a single crusher is not going to do dip to.

No thats only 50, since above 0 stamina damage you already get reduced to walk and the stamina buffer is 50

Wayland-Smithy:
That's fine, I will remove the AOE stam damage from stomp and change toss to brute.

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -66,8 +63,8 @@
 	action_icon_state = "cresttoss"
 	mechanics_text = "Fling an adjacent target over and behind you. Also works over barricades."
 	ability_name = "crest toss"
-	plasma_cost = 100
-	cooldown_timer = 18 SECONDS
+	plasma_cost = 60
+	cooldown_timer = 10 SECONDS
```
SplinterGP, line 67: i just dont like much these, its a big movimentation skill for a tanky beno, don' think it needs its CD reduced to less than 15 seconds

Wayland-Smithy: how about 12? that puts it inline with shrike single target fling

SplinterGP: seems better then, the plasma cost can be big as it was, crusher doenst use that much plasma in charge

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -24,27 +24,24 @@
...
 			continue
 		var/distance = get_dist(M, X)
 		var/damage = X.xeno_caste.stomp_damage/max(1, distance + 1)
 		if(distance == 0) //If we're on top of our victim, give him the full impact
 			GLOB.round_statistics.crusher_stomp_victims++
 			SSblackbox.record_feedback("tally", "round_statistics", 1, "crusher_stomp_victims")
-			M.take_overall_damage_armored(damage, BRUTE, "melee", FALSE, FALSE)
-			M.Paralyze(20)
+			M.take_overall_damage_armored(damage, BRUTE, "melee", FALSE, FALSE, TRUE)
+			M.Paralyze(3 SECONDS)
```
SplinterGP, line 36: btw not sure if this is in ticks actually, those metric for some status effects are weird

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
Wayland-Smithy: Okay making it 12

>rusher doenst use that much plasma in charge

The cost is reduced to match the cooldown so your plasma drains at the same speed

>(infact you can charge basically infinitely on weeds)

Incorrect, please test it yourself in long lanes if you don't believe me.

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
Wayland-Smithy: It boils down to `apply_status_effect()` which everything else passes ticks, like this previously has

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -24,27 +24,24 @@
 	"<span class='xenodanger'>We smash into the ground!</span>")
 	X.create_stomp() //Adds the visual effect. Wom wom wom
 
-	for(var/mob/living/M in range(0,X.loc))
+	for(var/mob/living/M in range(1,X.loc))
 		if(isxeno(M) || M.stat == DEAD || isnestedhost(M))
```
SplinterGP, line 28:
```suggestion
		if(X.issamexenohive(L) || M.stat == DEAD || isnestedhost(M))
```

SplinterGP (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -24,27 +24,24 @@
 	"<span class='xenodanger'>We smash into the ground!</span>")
 	X.create_stomp() //Adds the visual effect. Wom wom wom
 
-	for(var/mob/living/M in range(0,X.loc))
+	for(var/mob/living/M in range(1,X.loc))
```
SplinterGP, line 27:
```suggestion
	for(var/mob/living/M in range(1, get_turf(X)))
```

SplinterGP (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -24,27 +24,24 @@
...
 		else
 			step_away(M, X, 1) //Knock away
 			shake_camera(M, 2, 2)
 			to_chat(M, "<span class='highdanger'>You reel from the shockwave of [X]'s stomp!</span>")
-		if(distance < 2) //If we're beside or adjacent to the Crusher, we get knocked down.
-			M.Paralyze(20)
-		else
-			M.Stun(20) //Otherwise we just get stunned.
-		M.apply_damage(damage, STAMINA, updating_health = TRUE) //Armour ignoring Stamina
+			M.take_overall_damage_armored(damage, BRUTE, "melee", FALSE, FALSE, TRUE) // give half damage brute instead of stamina
```
SplinterGP, line 43: doenst actually do half damage
```suggestion
			M.take_overall_damage_armored(damage, BRUTE, "melee", FALSE, FALSE, TRUE) 
```

SplinterGP (changes requested):

SplinterGP (approved):

PR #7002, opened 2021-05-31
Adds the ability to override the default respawn timers via a config value

ZephyrTFA:
Adds config values to override the respawn timers for both marines and xenomorphs which are read on the Ticker's Initialize.

MARINE_RESPAWN and XENO_RESPAWN are now config values that you can set to override the default respawn timers.
Hard-coded values are bad and this lets the values be changed globally without needing to be manually changed ingame.

## Changelog
:cl:
server: Two new config values for overriding the respawn timers of both marines and xenomorphs.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -41,6 +41,9 @@ SUBSYSTEM_DEF(ticker)
 
 /datum/controller/subsystem/ticker/Initialize(timeofday)
 	load_mode()
+	
+	GLOB.respawntime = CONFIG_GET(number/marine_respawn)
+	GLOB.xenorespawntime = CONFIG_GET(number/xeno_respawn)
 
 	var/all_music = CONFIG_GET(keyed_list/lobby_music)
 	var/key = SAFEPICK(all_music)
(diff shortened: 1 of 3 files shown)


BraveMole:
If the plan is to make it so they can be tweaked depending on the gamemode, then no, don't do that

ZephyrTFA:
> 
> 
> If the plan is to make it so they can be tweaked depending on the gamemode, then no, don't do that

The idea is that you can change them via a config value. End of story.

BraveMole:
oh, i thought you were xenophile, my bad

code/controllers/subsystem/ticker.dm
```diff
@@ -41,6 +41,9 @@ SUBSYSTEM_DEF(ticker)
 
 /datum/controller/subsystem/ticker/Initialize(timeofday)
 	load_mode()
+	
+	GLOB.respawntime = CONFIG_GET(number/marine_respawn)
+	GLOB.xenorespawntime = CONFIG_GET(number/xeno_respawn)
```
SplinterGP, line 46: i think its better to replace all GLOB.respawntime and GLOB.xenorespawntime for the CONFIG_GET

SplinterGP (changes requested):

code/controllers/subsystem/ticker.dm
TiviPlus: No this is fine

TiviPlus (approved):

code/controllers/subsystem/ticker.dm
SplinterGP: Oh ye thats right

SplinterGP (approved):

PR #7010, opened 2021-06-01
Shields slow you down heavily

Pariah919:
## About The Pull Request
shields slow you down heavily instead of being minor

## Why It's Good For The Game
shields are innanely tanky for how effectively unheavy they are, overall this should make them more bearable, that or I could reduce their hard armor to 0 for melee.

## Changelog
:cl:
balance: shields slow you down heavily
/:cl:


JackTheJackhammer:
People only use shields for melee, which makes that particular marine basically null and void.

Walarks:
What if huge slowdown (1+) and durability buff or make shield not qdel at 0 durability?

AnCopper:
Shields are only used for Melee, so this would completely go against that, I'd rather an item be too powerful than useless.

Pariah919:
i have a better form

PR #7011, opened 2021-06-01
Fixes varedit marking

DeltaFire15:
## About The Pull Request
I was testing stuff on local and wanted to proccall something with a marked thing, but mark no work. This annoyed me enough to fix it, so, here we go.
Also moves marking into its own file as a proc so it can be used in the future if someone bothers porting other vv stuff.

## Why It's Good For The Game
Admins being able to mark stuff is pretty neat.

## Changelog
:cl:
fix: Admins can now actually mark things via the varedit dropdown.
code: Moved marking into its own file as a proc for easier access in the future.
/:cl:

### DIFF
--- a/code/modules/admin/view_variables/mark_datum.dm
+++ b/code/modules/admin/view_variables/mark_datum.dm
@@ -1,0 +1,19 @@
+/client/proc/mark_datum(datum/D)
+	if(!holder)
+		return
+	if(holder.marked_datum)
+		holder.UnregisterSignal(holder.marked_datum, COMSIG_PARENT_QDELETING)
+		vv_update_display(holder.marked_datum, "marked", "")
+	holder.marked_datum = D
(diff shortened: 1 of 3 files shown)


code/modules/admin/view_variables/mark_datum.dm
```diff
@@ -0,0 +1,7 @@
+/client/proc/mark_datum(datum/D)
+	if(!holder)
```
TiviPlus, line 2: See contribution guidelines, all procs should be documented using dmdoc
```suggestion
///Sets the datum D as this clients holders marked datum, given they have a holder
/client/proc/mark_datum(datum/D)
	if(!holder)
```

DeltaFire15: Documentation added. This sufficient?

TiviPlus (approved):
Yep perfect

PR #6999, opened 2021-05-31
Revert "Fixed Xeno Pings Being Generally Visible"

BraveMole:
Reverts tgstation/TerraGov-Marine-Corps#6859

Completly breaks all xeno hud the moment you cross a health threshold, probably a bad runtime or something.

Makes the game unplayable for xeno, please TM until fix

PR #7014, opened 2021-06-02
Fixes Xeno Pings

Surrealaser:
## About The Pull Request

Fixes Xeno Pings (hopefully)

## Why It's Good For The Game

Fixes Xeno Pings (hopefully)

## Changelog
:cl:
fix: Fixes Xeno Pings (hopefully)
/:cl:


TiviPlus (changes requested):
This is a strange approach can you explain this and how it will fix it please?

Surrealaser:
> This is a strange approach can you explain this and how it will fix it please?

Basically the effect object is not necessary to do the ping so I'm getting rid of it.

Testing has worked out in singleplayer on test servers, so far so good, with the desired effects (a ping visual that goes away when it should) and no unintended consequences. There may be multiplayer interactions that aren't being tested which is why I recommend TMing before merge.

SplinterGP:
testmerge while untestmerging https://github.com/tgstation/TerraGov-Marine-Corps/pull/6999

PR #7026, opened 2021-06-03
Fixes a minor harvester blade related runtime

DeltaFire15:
## About The Pull Request
The harvester blade didn't handle being clicked with an item that is not a container very well, resulting in runtimes. This fixes that.

## Why It's Good For The Game
runtimes bad.

## Changelog
:cl:
fix: Harvester Blades no longer runtime if clicked with things that aren't reagent containers.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -82,6 +82,10 @@
 
 	var/trans
 	var/obj/item/reagent_containers/container = I
+	if(!container || !container.reagents)
+		to_chat(user, "<span class='rose'>[I] isn't compatible with [src].</span>")
+		return TRUE
+
 	if(!container.reagents.total_volume)
 		trans = beaker.reagents.trans_to(container, 30)
 		to_chat(user, "<span class='rose'>[trans ? "You take [trans]u out of the internal storage. It now contains [beaker.reagents.total_volume]u" : "[src]'s storage is empty."].</span>")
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/blades.dm
```diff
@@ -82,6 +82,10 @@
 
 	var/trans
 	var/obj/item/reagent_containers/container = I
+	if(!container || !container.reagents)
+		to_chat(user, "<span class='rose'>[I] isn't compatible with [src].</span>")
+		return TRUE
```
TiviPlus, line 87:
```suggestion
	if(!container.reagents)
		to_chat(user, "<span class='rose'>[I] isn't compatible with [src].</span>")
		return TRUE
```

TiviPlus, line 87:
```suggestion
	if(!container.reagents)
		to_chat(user, "<span class='rose'>[I] isn't compatible with [src].</span>")
		return TRUE
	var/obj/item/reagent_containers/container = I
```

DeltaFire15: Would still runtime as the typecast would fail on objects that have reagents but aren't reagent containers of which there are a few (e.g. the engineer welding backpack), leading to trying to access null.reagents.total_volume in l.89

TiviPlus: Typecasts can never fail in dm (a weak point actually as var/list/totallyalist = "a string" will always runtime)
It also means this typecast happens incorrectly often as any item that is a non reagent could be used to click on
Thus a better fix would be a 
if(!isreagentontainer(I))
   return check
or even more optimally finding out why I is null, as it should nver be null

TiviPlus (approved):

PR #7028, opened 2021-06-03
Fix say.dm index out of bounds

Wayland-Smithy:
## About The Pull Request

Fixes a runtime caused by saying just "." or ":" trying to access the next non-existent character in the message for a department key.

## Why It's Good For The Game

Fix good.

## Changelog
:cl:
fix: Say verb get_key() no longer runtimes.
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -245,9 +245,9 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 
 
 /mob/living/proc/get_key(message)
-	var/key = message[1]
-	if(key in GLOB.department_radio_prefixes)
-		return lowertext(message[1 + length(key)])
+	var/prefix = message[1]
+	if(length(message) >= 2 && (prefix in GLOB.department_radio_prefixes))
+		return lowertext(message[2])
 
 /mob/living/proc/get_message_language(message)
 	if(length(message) >= 2 && message[1] == ",")


code/modules/mob/living/say.dm
```diff
@@ -245,9 +245,9 @@ GLOBAL_LIST_INIT(department_radio_keys, list(
 
 
 /mob/living/proc/get_key(message)
-	var/key = message[1]
-	if(key in GLOB.department_radio_prefixes)
-		return lowertext(message[1 + length(key)])
+	var/prefix = message[1]
+	if(length(message) >= 2 && (prefix in GLOB.department_radio_prefixes))
+		return lowertext(message[2])
```
TiviPlus, line 250: Couold this possibly break ;

SplinterGP: Actually could maybe break :b and two or more characters i think

Wayland-Smithy: I don't see how? It's doing exactly what it did before?

`message[1 + length(key)]` = `meessage[2]` when key can only ever be a length of 1 since it is a char index too. Think maints!

Wayland-Smithy: There are no 2 letter department keys we checked when I did the same fix to `get_message_language()`

Even if there were the old code would not work with them either.

TiviPlus: > I don't see how? It's doing exactly what it did before?
> 
> `message[1 + length(key)]` = `meessage[2]` when key can only ever be a length of 1 since it is a char index too. Think maints!

; is a one letter key did you test this

Wayland-Smithy: Yes I did, now if you attempt to say just `;` you transmit `.` over the radio. 

Plus `get_key()` does nothing for `;` anyway besides causing a runtime if you send just one character.

TiviPlus (approved):

PR #7030, opened 2021-06-04
nerfs the t-42

Pariah919:
## About The Pull Request
nerfs the wield slowdown, the aim mode and a few other things

## Why It's Good For The Game
I think that we can all agree that the lmg is vastly too overpowered compared to it's rifle family sisters, and that it needs a severe nerf, so I have added way more wield slowdown and aim time.

## Changelog
:cl:
balance: t-42 nerfs
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/magazines/rifles.dm
@@ -206,7 +206,7 @@
 	caliber = CALIBER_10X24_CASELESS
 	default_ammo = /datum/ammo/bullet/rifle
 	w_class = WEIGHT_CLASS_NORMAL
-	max_rounds = 120
+	max_rounds = 100
 	gun_type = /obj/item/weapon/gun/rifle/standard_lmg
 	icon_state_mini = "mag_t42"
 
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -585,10 +585,10 @@
 	icon_state = "t42"
 	item_state = "t42"
 	caliber = CALIBER_10X24_CASELESS //codex
-	max_shells = 120 //codex
+	max_shells = 100 //codex
 	force = 30
-	aim_slowdown = 0.8
-	wield_delay = 1 SECONDS
+	aim_slowdown = 1
+	wield_delay = 1.4 SECONDS
 	fire_sound =  'sound/weapons/guns/fire/t40.ogg'
 	dry_fire_sound = 'sound/weapons/guns/fire/m41a_empty.ogg'
 	unload_sound = 'sound/weapons/guns/interact/T42_unload.ogg'
@@ -623,8 +623,8 @@
 	gun_skill_category = GUN_SKILL_HEAVY_WEAPONS
 	attachable_offset = list("muzzle_x" = 31, "muzzle_y" = 17,"rail_x" = 4, "rail_y" = 20, "under_x" = 16, "under_y" = 13, "stock_x" = 0, "stock_y" = 13)
 	actions_types = list(/datum/action/item_action/aim_mode)
-	aim_fire_delay = 0.18 SECONDS
-	aim_speed_modifier = 5
+	aim_fire_delay = 0.2 SECONDS
+	aim_speed_modifier = 6
 
 	fire_delay = 0.2 SECONDS
 	burst_amount = 1


SpaceLoveSs13:
What type of reasoning is "we can all agree upon". LMG is a weapon of choice for many due to how comfortable and satisfying it is to use it. It also have a relatively good balanced mag capacity. Increasing it's wield delay would just turn it into a GPMG.

Wayland-Smithy:
> Increasing it's wield delay would just turn it into a GPMG.

If wield delay is all they separates the two guns then maybe T42 has too much damage potential. For sure mag size. Good PR, but maybe not far enough, or maybe GPMG is too good all-around in comparison with stam buffs.

I'd still be slightly faster to wield, but have the matching aim-mode stats. Big deal. I would guess most players if not told it was changed would not even notice a difference until they run out of boolet 20 shots earlier.

lovegreenstuff:
We cannot all agree that it's overpowered.

Mind explaining exactly how?

Tristan63:
It already fires slower on aim mode than T12 or T18, and its ammo cannot be held on a standard belt. Not to mention attachment choices are rather poor.

SplinterGP:
OP how?

Wayland-Smithy:
> It already fires slower on aim mode than T12 or T18, and its ammo cannot be held on a standard belt. Not to mention attachment choices are rather poor.

If it is such a bad gun how come it makes up half the armament of unga balls?

DerrickChamberlain:
Slowdown feels fine, weapon feels fine, has half the capacity of a GPMG, so I am not sure what is wrong with it. Guess this gun will be braziled

DerrickChamberlain:
> OP how?

Yeah I am not sure how it is more OP then other options we have. It is a No from me.

DerrickChamberlain:
> > It already fires slower on aim mode than T12 or T18, and its ammo cannot be held on a standard belt. Not to mention attachment choices are rather poor.
> 
> If it is such a bad gun how come it makes up half the armament of unga balls?

Not sure, personally I see far greater use of GPMG with LMGs then I do with T-42. I see less T-42's floating about, but maybe I am just missing something here.

code/modules/projectiles/magazines/rifles.dm
```diff
@@ -206,7 +206,7 @@
 	caliber = CALIBER_10X24_CASELESS
 	default_ammo = /datum/ammo/bullet/rifle
 	w_class = WEIGHT_CLASS_NORMAL
-	max_rounds = 120
+	max_rounds = 100
```
DerrickChamberlain, line 209: Really?! Even less ammo? That is ridiculous, slowly ROF and less ammo.

SplinterGP:
> 
> 
> > It already fires slower on aim mode than T12 or T18, and its ammo cannot be held on a standard belt. Not to mention attachment choices are rather poor.
> 
> If it is such a bad gun how come it makes up half the armament of unga balls?

because people like, last i checked a gun being alot used  doens't mean its bad,

lovegreenstuff:
> If it is such a bad gun how come it makes up half the armament of unga balls?

Because a lot of people prefer comfort to actual effectiveness. T-60 sucks fucking dick but people use it all the time because they barely have to reload.

Losenis:
> If it is such a bad gun how come it makes up half the armament of unga balls?

Because 120 ammo means less reloading, which is a pita when you have to either:
A) Unwield, switch hands, click container and pull magazine out (or right click it), click on the gun with the nonzero chance of clicking the slot rather than the gun accidentally switching hands, switching hands again, wielding again
B) Clickdrag magazine to gun which you won't actually figure out on your own unless you're specifically told about it or RNGesus gives out the tip on roundstart while you happen to be on, needing to pixelhunt both magazine and gun.
And all of the above which gets 100x times worse in the middle of combat which can be interrupted with knockdowns or stuns making you start from even less than scratch due to having to pull out the gun again.

It's why you have GPMGers despite the gun being FF central due to having scatter even while staying still unless you specifically packed bipod that locks you in place. You don't see a lot of people running T12 and T18 due to the aim mode speed because they're awful to reload, even moreso the latter. It doesn't help that the gun covers part of the backpack, so you can accidentally right click the gun instead of the satchel to take a gun out and end up toggling safety on the gun.

Tristan63:
> > It already fires slower on aim mode than T12 or T18, and its ammo cannot be held on a standard belt. Not to mention attachment choices are rather poor.
> 
> If it is such a bad gun how come it makes up half the armament of unga balls?

Because due to its ammo capacity, and certain attachments, it can be a good solo rusher weapon. You barely have to reload in a rush, and you can kill 2-3 xenos then reload when the advance catches up to you. Its fire rate off of aim mode is decent for what it is, and it also isn't as cumbersome as the T60 which isn't a viable rush weapon.

davidchan13:
GPMG is far more common and LMG is used by a niche group of people who decip squad back in the day. It's only on paper that the T-42 is superior to a T-12 or T-60, in practical use the xeno has to cooperate with you (or be absolutely fucking stupid) to get that maximum potential of damage and mobility the T-42 supposedly has. You're too slow to chase, you don't have near as much suppression as a T-60 and the only real use for the T-42 today is to walk fire into a xeno position to make them scatter, something any defiler, boiler or queen can break in a heartbeat.

TiviPlus:
literally why though

TiviPlus:
Unneeded nerf splinter agreed

PR #7040, opened 2021-06-06
Fix nightfall

BraveMole:
## About The Pull Request

Fix Nightfall

Fix Queen not being able to give plasma to King

## Changelog
:cl:
fix: Nightfall now applies on all electrical lights like it should
fix: You can give plasma as a queen to king
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -885,9 +885,11 @@ Proc for attack log creation, because really why not
 	if(cooldown <= 0)
 		cooldown = 1 SECONDS
 	TIMER_COOLDOWN_START(src, COOLDOWN_LIGHT, cooldown)
-	if(forced & !toggle_on) //Is true when turn light is called by nightfall
+	if(toggle_on == light_on)
+		return NO_LIGHT_STATE_CHANGE
+	if(forced & !toggle_on) //Is true when turn light is called by nightfall and the light is already on
 		addtimer(CALLBACK(src, .proc/reset_light), cooldown + 1)
-	if(sparks)
+	if(sparks && light_on)
 		var/datum/effect_system/spark_spread/spark_system = new
 		spark_system.set_up(5, 0, src)
 		spark_system.attach(src)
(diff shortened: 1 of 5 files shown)


RenGusta:
Y U no separate balance and fix PR my dude

KornFlaks:
I fall to see how this in any way makes the King a viable option to waste points on rather than plopping down more silos.

I've tried summoning it once and even with a very good xeno player at its controls, it seemed to me to be a worse combat caste than the Queen herself and that's saying something.

If instead of summoning a King, it summons a Boiler, I would definitely consider saving points every single game just for the eventual FOB siege as it would be far more useful.

Yes I know, back when it first TM'ed, the King was brokenly over powered and could solo entire marine landings but in its current state it's something to be coddled like a boiler. Perhaps more given the investment required to create one. If the previous abilities were so overpowered as to be tuned into mere shadows of themselves, perhaps consider going back to the drawingboard and think up of new abilities rather than leaving the vestigial remnants hanging off the shambling corpse that is post-nerf King.

Dont get me wrong, the sprite is great but that is pretty much the only redeeming factor of the 1800 or 2500 point dump.

Apologies if this is off topic to the PR at hand. It appears that my frustration at the waste of points I invested into summoning the king in a game earlier is influencing my commenting.

I just would like to ask you to please reconsider the game design and abilities aspect of the King as, in it's current state, it will never be used.

BraveMole:
@TiviPlus

code/game/atoms.dm
```diff
@@ -885,9 +885,11 @@ Proc for attack log creation, because really why not
 	if(cooldown <= 0)
 		cooldown = 1 SECONDS
 	TIMER_COOLDOWN_START(src, COOLDOWN_LIGHT, cooldown)
-	if(forced & !toggle_on) //Is true when turn light is called by nightfall
+	if(toggle_on == light_on)
+		return NO_LIGHT_STATE_CHANGE
+	if(forced & !toggle_on) //Is true when turn light is called by nightfall and the light is already on
```
SplinterGP, line 890: probably wanted the AND operator
```suggestion
	if(forced && !toggle_on) //Is true when turn light is called by nightfall and the light is already on
```

SplinterGP (changes requested):

PR #7042, opened 2021-06-06
Getting up as a xeno takes 1 sec

BraveMole:
## About The Pull Request

Title. It is 2 sec currently

## Why It's Good For The Game

You can't do anything against a marine running towards you. You see the light, you hit rest keybind, too bad, you're dead. Completly unfair

## Changelog
:cl:
balance: Getting up as a xeno takes 1 sec rather than 2
/:cl:

### DIFF
--- a/code/modules/mob/living/living_verbs.dm
+++ b/code/modules/mob/living/living_verbs.dm
@@ -22,7 +22,7 @@
 	else if(do_actions)
 		to_chat(src, "<span class='warning'>You are still in the process of standing up.</span>")
 		return
-	else if(do_mob(src, src, 5, ignore_flags = (IGNORE_LOC_CHANGE|IGNORE_HAND)))
+	else if(do_mob(src, src, 1 SECONDS, ignore_flags = (IGNORE_LOC_CHANGE|IGNORE_HAND)))
 		get_up()
 
 /mob/living/proc/get_up()
--- a/code/modules/mob/living/living_verbs.dm
+++ b/code/modules/mob/living/living_verbs.dm
@@ -22,7 +22,7 @@
 	else if(do_actions)
 		to_chat(src, "<span class='warning'>You are still in the process of standing up.</span>")
 		return
-	else if(do_mob(src, src, 2 SECONDS, ignore_flags = (IGNORE_LOC_CHANGE|IGNORE_HAND)))
+	else if(do_mob(src, src, 5, ignore_flags = (IGNORE_LOC_CHANGE|IGNORE_HAND)))
 		get_up()
 
 /mob/living/proc/get_up()


David-Stormwell:
Dam fast.
Still shouldn't encourage frontline resting though, I'd say 1 second would be better

SplinterGP:
as above, 1 to 1,5 seconds is better, dont want people frontline resting and unpunished mistakes

code/modules/mob/living/living_verbs.dm
```diff
@@ -22,7 +22,7 @@
 	else if(do_actions)
 		to_chat(src, "<span class='warning'>You are still in the process of standing up.</span>")
 		return
-	else if(do_mob(src, src, 2 SECONDS, ignore_flags = (IGNORE_LOC_CHANGE|IGNORE_HAND)))
+	else if(do_mob(src, src, 5, ignore_flags = (IGNORE_LOC_CHANGE|IGNORE_HAND)))
```
BraveMole, line 25:
```suggestion
	else if(do_mob(src, src, 1 SECONDS, ignore_flags = (IGNORE_LOC_CHANGE|IGNORE_HAND)))
```

Wayland-Smithy:
Buffing xeno? NOT ON MY TGMC

MiniMeatwad:
If there's a problem with frontline resting, I suggest making recovery weaker whenever xenos take damage, and have it wear off whenever they're no longer fighting. The first time it was implemented it made it so a full recovery took about 5 minutes if you had no pheromones, which I was told is why it was removed.

SplinterGP:
> 
> 
> If there's a problem with frontline resting, I suggest making recovery weaker whenever xenos take damage, and have it wear off whenever they're no longer fighting. The first time it was implemented it made it so a full recovery took about 5 minutes if you had no pheromones, which I was told is why it was removed.

thats already a thing

ZahaSanko:
Frontline resting should bring risk of getting rushed, without longer wind up xenos will just run away, when you enter their screen.

BraveMole:
> Frontline resting should bring risk of getting rushed, without longer wind up xenos will just run away, when you enter their screen.

It's no longer Frontline resting with the stam or, it's resting anywhere at all knowing some Usain bolt can come at any time anywhere and you barely will have time to react

Wayland-Smithy:
> xenos will just run away, when you enter their screen

This is what we do already. I start getting up the second your light touches my eyeballs. Stam buff makes it not matter in some cases then they roll up and stagger you to death. Very fun.

BraveMole:
Yeah so even without stam pr, this is needed. A marine running at you full speed will not let you any chance to react. 

This is not about front line resting, this happens everywhere. You can be resting peacefully in a random place, and if you marine rush at you you are just dead

EgorDinamit:
> thats already a thing, the problem with frontline resting is that its a incompetent thing to do and allows you to come to battle quicker, which then is of course a bad thing, people shoulndt be awarded by doing stupid things.

Isn't it literally the entire point of playing xenos though?
You are weak, deal little damage, and are easy to kill, but you (should) regenerate quickly.
Whereas marines are tanky af and deal shit-ton of damage, but can receive lasting damage(bones, blood loss) requiring them to visit medic from time to time.

RyderFive:
A bit late but oh well,

Punishing marines for using their limited stamina to catch a enemy off guard seems kinda meh to me. Most xenos have a stun they can use even if they get up late allowing them to run away a bit even after being caught off guard, running towards resting xenos also come with risks typically, and rushing marines can be avoided entirely in some cases with sticky resin or cover. Maybe this could be done better i dunno.

Its also kinda funny that you reopened this a few mins to a hour after a round where I did this to you.

Ieteihowmee:
No longer needed. Close the PR, since Stamina is back.

SplinterGP:
don't think this is necessary without the no stamina use on run is needed, i know your get up gets cancelled if hit by certain stuff but that should be changed instead.

PR #7043, opened 2021-06-06
Balance plasma pistol

Wayland-Smithy:
## About The Pull Request

Removes scope attachment slot and reduces ammo max range.

## Why It's Good For The Game

No other *normal* pistol in the game allows you to attach full blown scopes to it. How did this even get past review? I heard it was a shitpost speedmerge.

Gun should be more specialized and situational, not the perfect tool for long range flame area denial. Flamer and miniflamer do exist, and this is only a *sidearm*.

Actually balance this uber utility weapon. Make people happy.

## Changelog
:cl:
balance: Plasma pistol no longer accepts scope attachments.
balance: Plasma pistol now has 7 range, not 40.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1509,7 +1509,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	icon_state = "overchargedlaser"
 	hud_state = "laser_sniper"
 	damage = 40
-	max_range = 40
+	max_range = 7
 	penetration = 5
 	shell_speed = 1.5
 	flags_ammo_behavior = AMMO_ENERGY|AMMO_INCENDIARY|AMMO_EXPLOSIVE
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -98,7 +98,6 @@
 		/obj/item/attachable/lasersight,
 		/obj/item/attachable/gyro,
 		/obj/item/attachable/lace,
-		/obj/item/attachable/scope/marine,
 	)
 
 	muzzleflash_iconstate = "muzzle_flash_laser"
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -85,7 +85,7 @@
 	icon_state = "tx7"
 	item_state = "tx7"
 	caliber = CALIBER_PLASMA
-	max_shots = 10
+	max_shots = 3
 	reload_sound = 'sound/weapons/guns/interact/tp14_reload.ogg'
 	fire_sound = 'sound/weapons/guns/fire/laser3.ogg'
 	current_mag = /obj/item/ammo_magazine/pistol/plasma_pistol
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -85,7 +85,7 @@
 	icon_state = "tx7"
 	item_state = "tx7"
 	caliber = CALIBER_PLASMA
-	max_shots = 3
+	max_shots = 10
 	reload_sound = 'sound/weapons/guns/interact/tp14_reload.ogg'
 	fire_sound = 'sound/weapons/guns/fire/laser3.ogg'
 	current_mag = /obj/item/ammo_magazine/pistol/plasma_pistol
--- a/code/modules/projectiles/magazines/pistols.dm
+++ b/code/modules/projectiles/magazines/pistols.dm
@@ -30,7 +30,7 @@
 	desc = "An energy cell for the TX-7 plasma pistol."
 	caliber = CALIBER_PLASMA
 	icon_state = "tx7"
-	max_rounds = 10
+	max_rounds = 3
 	w_class = WEIGHT_CLASS_SMALL
 	default_ammo = /datum/ammo/energy/plasma_pistol
 	gun_type = /obj/item/weapon/gun/pistol/plasma_pistol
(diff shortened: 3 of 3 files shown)


David-Stormwell:
Never see much of a problem with it as a xeno, slow ass projectile plus it'll only fire on where you click, so can't really fight xenos with it unless they rush you head first.

Wayland-Smithy:
Do you even play the game David?

You must never have been held at a 3 wide choke with plasma pistol spammer 3 screens away.

DerrickChamberlain:
AOD type weapon anyway, not entirely sure about the merits of the projectile speed, I see it used not often enough. Less ammo, and less utility is just going to make people avoid it.

Wayland-Smithy:
> I see it used not often enough.

1 dedicated user per round is too many currently, try fighting against it sometime.

EvilMelonTheSecond:
I don't see a reason to remove it but at the same time I don't see a reason to keep it. It's a utility pistol it shouldn't have a scope if we want to apply common sense. On the other hand it's... a utility pistol that does almost no damage and leaves a single tile fire, the old lasrifle with heatmode was more deadly than this.

Is this gonna change anything either way? I guess its fine if it will make people happy but this is the first time I'm hearing a complaint about it.

Wayland-Smithy:
> leaves a single tile fire

Fire that lasts 12 seconds, so you can easily maintain 6-8 tiles of fire at max range.

davidchan13:
This is smell so idedplznerf, plas pistol is not a game changer and is really bad at fighting and containing xenos when boilers and spitters can just spray the flames away. Comparing to flamers, the pistol is shit cause it fires so slow, affects so little tiles, and has a dedicated ammo unlike flamers that can be refueled from any fuel source. 

A marine deadset on keeping a hallway locked down needs to use a lot of ammo, of which a standard holster will only hold 48 shots, and a single marine walking between him and his intended target is the worst kind of FF. This is a massive amount of resources a  marine is dedicated to a tactic that is countered by xenos literally just finding another route.

Wayland-Smithy:
@davidchan13 and? I am not stopping any marines from continuing to use it to deny area or fight cqb, just doing it from 3 screens away. Are you really going to miss the scopes or try to argue that using them has any real benefit when you yourself said it is countered by going around? Which btw is not always the case and the reason for the PR with stam buffs making cave and choke defense the norm for xenos this allows for really unfun play consisting of wait for the plasma spammer to stop, go a million miles around, or eat the flame cross and have to resist 4 times with no means of recourse unless you are a boiler with matching range. At least flamer users have to be kind of close and at risk of being separated from the pack. There are plenty of guns that can clear walls from a distance or shoot xenos they are called snipers.

DerrickChamberlain:
> > I see it used not often enough.
> 
> 1 dedicated user per round is too many currently, try fighting against it sometime.

Boy howdy I taste salt in the air.

Wayland-Smithy:
> I get FF'd by fire plenty thank you.

At least you would be able to see the source now.

usnpeepoo:
plasma pistol users are literally cucked by warrior because the bring 1 tile flames that can be tossed into from anyone near it gun doesn't seem unbalanced

Wayland-Smithy:
I think you missed the point of my PR. It will change nothing about the situation you bring up for either side. If someone is 3 screens away and shooting a warrior is not touching them anyway. Their biggest complaint should be the acid spray cucking your remote file tiles.

jroinc:
> 
> 
> Do you even play the game a lot lately David? It is not slow.
> 
> You must never have been held at a 3 wide choke with 2 tile deep fire by plasma pistol spammer 3 screens away.

The projectile is very slow. It can take two to three full seconds to hit a target on the edge of scope range.

KornFlaks:
> The projectile is very slow. It can take two to three full seconds to hit a target on the edge of scope range.

Are we reading the same PR? This is not the problem of hitting something at 3 screens away, it's the fact that the ground can be lit on fire 3 screens away.

Literally nothing else is changing about the gun, except for the fact that it can not be used 3 screens away (40 tile range, how did this pass initial review?).

jroinc:
> 
> 
> > The projectile is very slow. It can take two to three full seconds to hit a target on the edge of scope range.
> 
> Are we reading the same PR? This is not the problem of hitting something at 3 screens away, it's the fact that the ground can be lit on fire 3 screens away.
> 
> Literally nothing else is changing about the gun, except for the fact that it can not be used 3 screens away (40 tile range, how did this pass initial review?).

We are. I have no complaint about the PR. I'm just pointing out that it is in fact one of, if not *the* slowest projectiles ingame.

Wayland-Smithy:
For the record: It's 50% faster than xeno spit, 150% faster than tesla, and 25% slower than rockets, a **pistol**.

KornFlaks:
> For the record: It's 50% faster than xeno spit

Incoming PR to buff xeno spit speed. After all, it cant be the slowest projectile in the game. We need to fix that pronto.

Georgeur:
Fucking finally. 
Anyone complaining that its a slow projectile is a retarded brainlet not realising that getting lit on fire 4 screens away whilst resting is the problem. Well that and sniping turrets from 4 screens away. 
The guy who made the gun told me he didn't think this'd get in when it was tm'd. 
If you're trying to defend it with the whining moans and grunts of a autistic caveman please dm me to prove you're not a lt.

https://user-images.githubusercontent.com/68168644/121568815-02cdce00-ca18-11eb-9590-c52f004c6ade.mp4
Funny armor bypass for a small item size magazine.

DerrickChamberlain:
> > I get FF'd by fire plenty thank you.
> 
> At least you would be able to see the source now. Also Derrick I am confused do you see it not often enough or get FF'd plenty? Can't be both.

I said FF'd by fire plenty, never said plasma pistol, However when it is used. I sure do get FF'd by it sometimes.

PR #6842, opened 2021-05-12
Makes the revolver barrel being an attachment, removing them making snub nosed revolvers.

SplinterGP:
## About The Pull Request
This makes the barrel of the revolvers being an attachment that they start with, which increase their size and gives status(nothing changed from what is now) and when removed they can fit on bags and such again, but with lower status

## Why It's Good For The Game
Nerfs infinite revolver meme without removing entirely

## Changelog
:cl:
balance: revolver barrel is now an attachment that comes with the revolver that makes it larget(not fit in bags) removing this attachment makes the revolver weaker in favor of making it fit in bags.
/:cl:

PR #7047, opened 2021-06-07
Taking guns out of bags now has a delay

ZephyrTFA:
## About The Pull Request

If you attempt to pull a gun out of any storage that isn't a belt it takes half a second.
If you move or 'fumble' during this period you simply drop the gun on the ground instead.

## Why It's Good For The Game

What if we nerfed revolver spam without actually nerfing the revolver?

## Changelog
:cl:
balance: Attempting to pull any gun out of a storage item has a delay; excludingbelts.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -416,6 +416,10 @@
 	if(!istype(I))
 		return FALSE
 
+	var/removal_good = TRUE
+	if(isgun(I) && !istype(src, /obj/item/storage/belt)) // If its a belt we dont care; they're supposed to be 'easy access'
+		removal_good = remove_gun_from_storage(I)
+
 	for(var/i in can_see_content())
 		var/mob/M = i
 		if(!M.client)
@@ -445,7 +449,17 @@
 		I.mouse_opacity = initial(I.mouse_opacity)
 
 	update_icon()
-	return TRUE
+	return removal_good
+
+/obj/item/storage/proc/remove_gun_from_storage(obj/item/weapon/gun/gun)
+	var/mob/user = usr
+	if(!istype(user))
+		return TRUE
+	to_chat(user, "<span class='notice'>You begin drawing [gun].</span>")
+	if(do_after(user, 0.5 SECONDS, TRUE, src))
+		return TRUE
+	to_chat(user, "<span class='warning'>You fumble with the [gun] and it drops to the ground!</span>")
+	return FALSE
 
 //This proc is called when you want to place an item into the storage item.
 /obj/item/storage/attackby(obj/item/I, mob/user, params)
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -450,7 +450,7 @@
 	if(isgun(I) && fumble_guns && user)
 		to_chat(user, "<span class='notice'>You begin drawing [I].</span>")
 		if(!do_after(user, 0.5 SECONDS, TRUE, src))
-			to_chat(user, "<span class='warning'>You fumble with [gun] and it drops to the ground!</span>")
+			to_chat(user, "<span class='warning'>You fumble with [I] and it drops to the ground!</span>")
 			return FALSE
 
 	return TRUE
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -417,15 +417,6 @@
 	if(!istype(I))
 		return FALSE
 
-	var/removal_good = TRUE
-	if(isgun(I) && fumble_guns)
-		if(user)
-			var/obj/item/weapon/gun/gun = I
-			to_chat(user, "<span class='notice'>You begin drawing [gun].</span>")
-			if(!do_after(user, 0.5 SECONDS, TRUE, src))
-				removal_good = FALSE
-				to_chat(user, "<span class='warning'>You fumble with [gun] and it drops to the ground!</span>")
-
 	for(var/i in can_see_content())
 		var/mob/M = i
 		if(!M.client)
@@ -455,7 +446,15 @@
 		I.mouse_opacity = initial(I.mouse_opacity)
 
 	update_icon()
-	return removal_good
+
+	if(isgun(I) && fumble_guns && user)
+		var/obj/item/weapon/gun/gun = I
+		to_chat(user, "<span class='notice'>You begin drawing [gun].</span>")
+		if(!do_after(user, 0.5 SECONDS, TRUE, src))
+			to_chat(user, "<span class='warning'>You fumble with [gun] and it drops to the ground!</span>")
+			return FALSE
+
+	return TRUE
 
 //This proc is called when you want to place an item into the storage item.
 /obj/item/storage/attackby(obj/item/I, mob/user, params)
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -448,7 +448,6 @@
 	update_icon()
 
 	if(isgun(I) && fumble_guns && user)
-		var/obj/item/weapon/gun/gun = I
 		to_chat(user, "<span class='notice'>You begin drawing [gun].</span>")
 		if(!do_after(user, 0.5 SECONDS, TRUE, src))
 			to_chat(user, "<span class='warning'>You fumble with [gun] and it drops to the ground!</span>")
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -448,7 +448,7 @@
 	update_icon()
 
 	if(isgun(I) && fumble_guns && user)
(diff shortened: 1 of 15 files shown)


code/game/objects/items/storage/storage.dm
```diff
@@ -416,6 +416,10 @@
 	if(!istype(I))
 		return FALSE
 
+	var/removal_good = TRUE
+	if(isgun(I) && !istype(src, /obj/item/storage/belt)) // If its a belt we dont care; they're supposed to be 'easy access'
```
BraveMole, line 420: Don't use istype(src)

A flag or an override on storage/belt is much better

Walarks:
Having guns be exempt from the typical inventory mechanics by adding a takeout delay sounds like it would make gameplay more clunky and inconsistent than anything else. Why not increase revolver size instead so you can't hold a ton of them in a bag?

code/game/objects/items/storage/storage.dm
```diff
@@ -445,7 +449,17 @@
 		I.mouse_opacity = initial(I.mouse_opacity)
 
 	update_icon()
-	return TRUE
+	return removal_good
+
+/obj/item/storage/proc/remove_gun_from_storage(obj/item/weapon/gun/gun)
```
BraveMole, line 454: don't use usr, you have to get the mob/user from parent proc (and from attack_hand it seems)

and just inline this in remove_from_storage , it doesn't need to be that far up in the proc

code/game/objects/items/storage/storage.dm
```diff
@@ -445,7 +449,17 @@
...
+	var/mob/user = usr
+	if(!istype(user))
+		return TRUE
+	to_chat(user, "<span class='notice'>You begin drawing [gun].</span>")
+	if(do_after(user, 0.5 SECONDS, TRUE, src))
+		return TRUE
+	to_chat(user, "<span class='warning'>You fumble with the [gun] and it drops to the ground!</span>")
```
BraveMole, line 461:
```suggestion
	to_chat(user, "<span class='warning'>You fumble with  [gun] and it drops to the ground!</span>")
```

code/game/objects/items/storage/storage.dm
BraveMole: Another note, new procs always must be autodoced. See pins in dev channel

BraveMole (changes requested):

code/game/objects/items/storage/storage.dm
```diff
@@ -506,7 +516,7 @@
 	var/turf/T = get_turf(src)
 	hide_from(usr)
 	for(var/obj/item/I in contents)
-		remove_from_storage(I, T)
+		remove_from_storage(I, T, usr)
```
ZephyrTFA, line 519: Yes I know using `usr` is bad.
Fixing this proc is outside the scope of this PR however.

code/game/objects/items/storage/storage.dm
```diff
@@ -412,10 +413,19 @@
 	return 1
 
 //Call this proc to handle the removal of an item from the storage item. The item will be moved to the atom sent as new_target
-/obj/item/storage/proc/remove_from_storage(obj/item/I, atom/new_location)
+/obj/item/storage/proc/remove_from_storage(obj/item/I, atom/new_location, mob/user)
 	if(!istype(I))
 		return FALSE
 
+	var/removal_good = TRUE
```
BraveMole, line 420: Unless i'm missing something, you don't need that var. you can put the rest of the code right above the return.

code/game/objects/items/storage/storage.dm
```diff
@@ -412,10 +413,19 @@
...
+	if(isgun(I) && fumble_guns)
+		if(user)
```
BraveMole, line 422: if(isgun(I) && fumble_guns && user)

BraveMole (changes requested):

code/game/objects/items/storage/storage.dm
ZephyrTFA: fair

ZephyrTFA:
I'm not sure if people realize that your exosuit storage slot isnt a storage slot and won't be affected by this.

Pariah919:
why are we nerfing a single particular playstyle which isn't even that overpowered or overbearing and isn't as easily done as it seems in the first place
literally one guy does it to any effect at all and it's a funny gimmick

AnCopper:
> why are we nerfing a single particular playstyle which isn't even that overpowered or overbearing and isn't as easily done as it seems in the first place
> literally one guy does it to any effect at all and it's a funny gimmick

MJP, have you played on TGMC in the last 3 months?

Pariah919:
> MJP, have you played on TGMC in the last 3 months?
>
i have, i've done it, i've seen people do it, it's not that effective compared to just running akimbo tp-14s or just playing the game with a gun

BraveMole:
ocelot builds are more popular on skyratt than on tgmc. Plus xeno are more robust here, so it's less on an issue. 

Still, nerfing that build doesn't seem to crazy, it's very strong

ZephyrTFA:
What's funny here is that you are comparing using revolvers, which are a secondary, to using a primary; and say they are only <i>slightly better</i>
You do understand the issue here?

Pariah919:
> What's funny here is that you are comparing using revolvers, which are a secondary, to using a primary; and say they are only slightly better
You do understand the issue here?
>
primaries and secondaries as a distinction honestly should stop existing in tgmc
if so then we should nerf absolutely every single firearm in the game as they're all available to be secondaries as 2 gun is unarguably the META for the average marine because it let's you cover the weakness of whatever your main gun is, and you can generally carry a decent amount of ammo for both of your guns (webbing+general storage is enough for 5 magazines which is decent enough for a surprising amount of guns, esp if it's not your main)

that or remove guns fitting on the back slot, that'd fix it really

As for revolver spam, I'd generally argue it's slightly more effective if you manage to actually, well pull it off correctly, which is very unlikely in my opinion, and generally it has bad overall use time, I'd argue it's more of a sum of it's part kit than the whole thing being overpowered (hauberk giving you medium level armor with 0.15 slowdown, revolvers having no slowdown doesn't help that when they're unwielded) aswell as people in general being pretty new/bad at skyrat, as they're probably not used to the fast paced gameplay tgmc generally plays around.

i don't want to argue whataboutism here but ocelot revolvers are generally not really as bad as they are, and this honestly isn't the way to fix it in my opinion, I'd rather nerf the other parts which contribute to making it go from a funnyman glass cannon to basically medium armor level stronk

SilencedMP5A5:
ladies and gentlemen

we got em

David-Stormwell:
If Ocelot is anything like me, he's getting a hard on right now due to all these playstyle nerfs.

SplinterGP:
this already exists and is called PR #6842

code/game/objects/items/storage/storage.dm
```diff
@@ -445,6 +446,14 @@
 		I.mouse_opacity = initial(I.mouse_opacity)
 
 	update_icon()
+
+	if(isgun(I) && fumble_guns && user)
+		var/obj/item/weapon/gun/gun = I
```
BraveMole, line 451: why casting? just use I.

ZephyrTFA: this is an artifact from when I was an idiot

ZephyrTFA, line 451:
```suggestion
```

code/game/objects/items/storage/storage.dm
```diff
@@ -445,6 +446,13 @@
 		I.mouse_opacity = initial(I.mouse_opacity)
 
 	update_icon()
+
+	if(isgun(I) && fumble_guns && user)
+		to_chat(user, "<span class='notice'>You begin drawing [gun].</span>")
```
ZephyrTFA, line 451:
```suggestion
		to_chat(user, "<span class='notice'>You begin drawing [I].</span>")
```

code/game/objects/items/storage/storage.dm
```diff
@@ -445,6 +446,13 @@
...
+		if(!do_after(user, 0.5 SECONDS, TRUE, src))
+			to_chat(user, "<span class='warning'>You fumble with [gun] and it drops to the ground!</span>")
```
ZephyrTFA, line 453:
```suggestion
			to_chat(user, "<span class='warning'>You fumble with [I] and it drops to the ground!</span>")
```

Georgeur:
I love this pr, and I love the groundwork it sets, but its broken. 

quick equip doesnt have a check, and doesnt have a delay
because the fumble is always done, and not when firing, it just throws the gun on the floor, and that means you can outrun your own gun and not be able to pick it up

this isnt a proper delay, its a ramp up
please keep making pr's though, I really like this

if you dont know why this is so good, its because macros can't work against ramp ups, they often break

SplinterGP:
None of the mantainers are willing to Merge this as far as i know.

PR #7064, opened 2021-06-11
medium acid spit no longer has AMMO_EXPLOSIVE

Pariah919:
## About The Pull Request
it doesn't need this anymore, it has a max range already

## Why It's Good For The Game
makes medium acid spit castes way less painful to play, dear god

## Changelog
:cl:
qol:acid spit with no puddle no longer stops where you click
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1784,6 +1784,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 /datum/ammo/xeno/acid/medium
 	name = "acid spatter"
 	damage = 30
+	flags_ammo_behavior = AMMO_XENO
 
 /datum/ammo/xeno/acid/heavy
 	name = "acid splash"


KornFlaks:
Note, medium acid spit applies to both Spitter and Queen (both share medium acid spit).
This may actually be a fairly large buff as the spit can now potentially target moving enemies if the click hits the turf and not the icon.

Icon click spit seems to have some sort of curvature that tracks enemies. I cant otherwise explain why I seem to hit so well when I manage to secure clicking on the enemys' icons. Probably just my confirmation bias.

Turf clicks on the other hand usually miss as the spit disappears on that tile without doing any damage. If it continued on for 8 more tiles (max range of spit) which is about half a screen over past the edge of current view, that will significantly increase the chance to hit.

I recommend some rounds of test merging before fully merging. This may be too large of a hit chance buff.

MLGTASTICa:
> Note, medium acid spit applies to both Spitter and Queen (both share medium acid spit).
> This may actually be a fairly large buff as the spit can now potentially target moving enemies if the click hits the turf and not the icon.
> 
> Icon click spit seems to have some sort of curvature that tracks enemies. I cant otherwise explain why I seem to hit so well when I manage to secure clicking on the enemys' icons. Probably just my confirmation bias.
> 
> Turf clicks on the other hand usually miss as the spit disappears on that tile without doing any damage. If it continued on for 8 more tiles (max range of spit) which is about half a screen over past the edge of current view, that will significantly increase the chance to hit.
> 
> I recommend some rounds of test merging before fully merging. This may be too large of a hit chance buff.

it has no curvature , it just doesn't delete itself if it doesn't reach them.

SplinterGP:
> 
> 
> Note, medium acid spit applies to both Spitter and Queen (both share medium acid spit).
> This may actually be a fairly large buff as the spit can now potentially target moving enemies if the click hits the turf and not the icon.
> 
> Icon click spit seems to have some sort of curvature that tracks enemies. I cant otherwise explain why I seem to hit so well when I manage to secure clicking on the enemys' icons. Probably just my confirmation bias.
> 
> Turf clicks on the other hand usually miss as the spit disappears on that tile without doing any damage. If it continued on for 8 more tiles (max range of spit) which is about half a screen over past the edge of current view, that will significantly increase the chance to hit.
> 
> I recommend some rounds of test merging before fully merging. This may be too large of a hit chance buff.

that can also be explained due to lag, as direct clicking youre clicking, input going where you clicked, and shooting at tile you clicked, when you click on sprite you're clicking on the mob itself, so if the mob moves before everything's done and goes shoot, it will go there, rather than where it was when you clicked.

SplinterGP (approved):

PR #7067, opened 2021-06-11
Fixes ravager charge damage bonus not being applied correctly

MLGTASTICa:
## About The Pull Request
Fixes ravager charge not applying bonus damage
## Why It's Good For The Game
Its a bug and coding mistake
## Changelog
:cl:
fix: Ravager charge now applies its damage bonus properly.
/:cl:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,12 +9,7 @@
     	"**/.yarn": true,
     	"**/.pnp.*": true
 	},
-	"workbench.editorAssociations": [
-		{
-			"filenamePattern": "*.dmi",
-			"viewtype": "imagePreview.previewEditor"
-		}
-	],
+	"workbench.editorAssociations": {},
 	"files.eol": "\n",
 	"gitlens.advanced.blame.customArguments": ["-w"],
 	"tgstationTestExplorer.project.resultsType": "json",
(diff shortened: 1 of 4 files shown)


SplinterGP:
The value in attack_alien_harm is NOT the damage it causes, its EXTRA damage for the attack, so it does slash damage + 25%, same for the charge

RenGusta:
Xenos have two tier slash damage for a reason. It's sane and consistent for balance. Your change drops ravager to a (theoretical ) peak time to crit of 4.9 down from the current 6.3 seconds. When it's already several seconds faster than other T3's with the slash it already has. 

Yucky.

MLGTASTICa:
> The value in attack_alien_harm is NOT the damage it causes, its EXTRA damage for the attack, so it does slash damage + 25%, same for the charge, and just buffing its damage is a nono.

right , after i looked deeper into how the proc for damage work , you're right, however i noticed that ravager bump used attack_alien , and not alien_attack_harm , so it would lose any bonus.
Ill revert the balance changes for now and make this a fix , considering that i misunderstood how it worked.

SplinterGP:
NP, thats good to make it harm, it'd might cause the xeno to just swap if not in harm intent

PR #7070, opened 2021-06-11
Fix Colony floodlight switch staying on after power loss

Wayland-Smithy:
## About The Pull Request

Title

## Why It's Good For The Game

Fix

## Changelog
:cl:
fix: Fixed colony floodlight switch staying on after power loss.
/:cl:


DerrickChamberlain:
So a revert? Why even make the PR? Not that the 'Fix' was added nothing changes.

Wayland-Smithy:
Reverts nothing. The other fix will still stop round start button killing which was the point I think.

Wayland-Smithy:
looks dumb but && works I guess

PR #7071, opened 2021-06-12
Fix AI Eye permanent deletion if looking at an escape shuttle

Wayland-Smithy:
## About The Pull Request

One line fix to check for eyeobj existence.

## Why It's Good For The Game

Fix

## Changelog
:cl:
fix: AI can now recover from a camera blackout if it was looking at an escape shuttle, or anything else causes deletion.
/:cl:


DerrickChamberlain:
OH! That would make sense why it would sometimes screw up my vision.

SplinterGP (approved):

PR #7072, opened 2021-06-12
Refactor diag movements 

BraveMole:
## About The Pull Request

Refactor diag movement code to be true diag movement and not qol like it is right now.
You can now shuffle on diagonals
Allow crusher to diag charge

## Why It's Good For The Game

Shuffling and pushing in diag is really neat. Diag speed = 1.5 is more closer to true diag movement (1.42), and i hope it's not that big of a balance breaker (it should still be possible for a xeno to pursue and attack a marine fleeing, but that should be tested). Can still be put back to two times, without any issue

The refactor prevent silly bugs where you can catch fire when diag moving near lava/fire, prevent double stack fire bug, etc etc.

Allow crusher and bull to charge diagonally, because why not? Because of the slower speed, this will rarely be usefull

## Changelog
:cl:
refactor: Refactor diag movements
balance: Crusher and bull can charge diagonally
qol: Allow to shuffle and push in diagonal movements
/:cl:


code/game/atoms_movable.dm
```diff
@@ -196,6 +198,7 @@
 			. = ..()
 		else //Diagonal move, split it into cardinal moves
 			moving_diagonally = FIRST_DIAG_STEP
+			olderloc = loc
```
SplinterGP, line 201: do you actually need this var at all ? there's oldloc

SplinterGP:
only TM's allowed

code/modules/mob/mob_movement.dm
```diff
@@ -141,7 +141,7 @@
 	. = ..()
 
 	if((direct & (direct - 1)) && mob.loc == n) //moved diagonally successfully
-		add_delay *= 2
+		add_delay *= 1.5
```
SplinterGP, line 144: keep this at 2 for now
```suggestion
		add_delay *= 2
```

SplinterGP (changes requested):

EgorDinamit:
Diagonal charge.  :eyes:

SplinterGP:
i'm gonna close this since i'm not going to deal with it and we don't have other maintainers up to reviewing it.

PR #7092, opened 2021-06-14
Update pain hud

BraveMole:
## About The Pull Request

add ratio to pain hud. Stamina and pain count less, so the perception of human health by xeno will be more accurate

## Why It's Good For The Game

The current pain hud is outdated, pain and stamina are no longer that important

## Changelog
:cl:
balance: Update pain hud (what xeno see on humans) to be more reliable. Pain and stamina have way less impact on it.
/:cl:

### DIFF
--- a/code/game/data_huds.dm
+++ b/code/game/data_huds.dm
@@ -281,6 +281,10 @@
 				status_hud.icon_state = "hudhealthy"
 				return TRUE
 
+#define HEALTH_RATIO_PAIN_HUD 1
+#define PAIN_RATIO_PAIN_HUD 0.25
+#define STAMINA_RATIO_PAIN_HUD 0.5
+
 
 /mob/proc/med_pain_set_perceived_health()
 	return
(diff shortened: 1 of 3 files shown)


DerrickChamberlain:
So how is this different then the other PR you just closed? What does it show then? What is more reliable? I think we are just word playing here.

BraveMole:
> So how is this different then the other PR you just closed? What does it show then? What is more reliable? I think we are just word playing here.

again, unless you played xeno, why do you even care?

korokorokke:
Personally, the fact that the HUD reflected pain rather than actual health was news to me at all, and explains a lot of my confusion.

The fact that much of a marine's pain state lies in a vague, blinking red meter is incredibly confusing. It could definitely stand to be made more granular for the sake of readability - whether as a corpsman or a xeno - though as previously mentioned, I am uncertain whether or not that is really the intention here.

EgorDinamit:
So that's why marines "in crit" were nowhere near actual crit, huh.

SplinterGP (approved):

SplinterGP:
give more info in PR desc please

PR #7099, opened 2021-06-15
Hemodile Buff

Surrealaser:
## About The Pull Request

1. Hemodile no longer deals stamina damage.

2. Hemodile base slowdown increased from 1 to 1.5.

3. Hemodile slowdown is multiplied by 2 for each other Xeno chem in the victim's system (Neurotoxin and Transvitox).

## Why It's Good For The Game

Makes Hemodile worth a shit.

Adds interesting synergies with other Xeno chems and castes.

## Changelog
:cl:
balance: Hemodile no longer deals stamina damage.
balance: Hemodile base slowdown increased from 1 to 1.5.
balance: Hemodile slowdown is multiplied by 2 for each other Xeno chem in the victim's system (Neurotoxin and Transvitox).
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -516,9 +516,9 @@
 /datum/reagent/toxin/xeno_growthtoxin/overdose_crit_process(mob/living/L, metabolism)
 	L.Losebreath(2)
 
-/datum/reagent/toxin/xeno_hemodile //slows by 25% and 50% based on base movement speed of marine with shoes on and deals an additional 20% of damage received as stamina damage
+/datum/reagent/toxin/xeno_hemodile //Slows its victim. The slow becomes twice as strong with each other xeno toxin in the victim's system.
 	name = "Hemodile"
-	description = "A stamina draining toxin. Causes increased stamina loss and slower movement."
+	description = "Impedes motor functions and muscle response, causing slower movement."
 	reagent_state = LIQUID
 	color = "#602CFF"
 	custom_metabolism = 0.4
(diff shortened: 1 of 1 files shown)


DerrickChamberlain:
Yikes. I hope marines carry more water.

SplinterGP:
i don't know if i like that, it removes its stamina reducing stuff and then increases its slowdown as if 1 slowdown aint big already, then allows it to be peaked at 4.5 slowdown which i dont think its a good value since thats like 2x at minimum the ammount of slowdown a marine gets ingame.

Surrealaser:
> i don't know if i like that, it removes its stamina reducing stuff and then increases its slowdown as if 1 slowdown aint big already, then allows it to be peaked at 4.5 slowdown which i dont think its a good value since thats like 2x at minimum the ammount of slowdown a marine gets ingame.

This is effectively a nerf until it gets the battle royale triple threat stack.

Anyways, again, if we can TM INFINITE SPRINT an obviously retarded idea, we can TM and test this relatively modest change.

SplinterGP:
> 
> 
> > i don't know if i like that, it removes its stamina reducing stuff and then increases its slowdown as if 1 slowdown aint big already, then allows it to be peaked at 4.5 slowdown which i dont think its a good value since thats like 2x at minimum the ammount of slowdown a marine gets ingame.
> 
> This is effectively a nerf until it gets the battle royale triple threat stack.
> 
> Anyways, again, if we can TM INFINITE SPRINT an obviously retarded idea, we can TM and test this relatively modest change.

please avoid using those type of words you can get banned by github like derick already did

SplinterGP (changes requested):
the other changes are fine, the multiplying with other chems is way too much and doenst seem like a good mechanic to me.

Surrealaser:
> the other changes are fine, the multiplying with other chems is way too much and doenst seem like a good mechanic to me.

It is arguably worse than the old Hemodile (which is not impressive) unless you get the trifecta which is not easy to do.

SplinterGP:
> 
> 
> > the other changes are fine, the multiplying with other chems is way too much and doenst seem like a good mechanic to me.
> 
> It is arguably worse than the old Hemodile (which is not impressive) unless you get the trifecta which is not easy to do.

its better in the slow in the exchange of stamina damage, for me you can just increase the slow and keep the stamina damage, i don't like the idea of scaling up per other chem because the other chem already has their effects, they dont need to double on each other due to that. you inject hemodile and transvitox you get the effect of hemodile and transvitox.

Surrealaser:
> > > the other changes are fine, the multiplying with other chems is way too much and doenst seem like a good mechanic to me.
> > 
> > 
> > It is arguably worse than the old Hemodile (which is not impressive) unless you get the trifecta which is not easy to do.
> 
> its better in the slow in the exchange of stamina damage, for me you can just increase the slow and keep the stamina damage, i don't like the idea of scaling up per other chem because the other chem already has their effects, they dont need to double on each other due to that. you inject hemodile and transvitox you get the effect of hemodile and transvitox.

Synergies make teamwork more compelling, and it is difficult to actually combo reagents with downside free unlimited gas protection from masks. Further, compounding effects make comboing compelling whereas they currently aren't.

Second, stamina damage steps on the toes of Neuro which is the primary reason I don't like it, besides making the focus of Hemodile too broad.

PR #7114, opened 2021-06-17
t-71 requires two hands to use

Pariah919:
## About The Pull Request
it was funny for one ronud, then it stopped being funny
the ability to throw grenades while being effectively invincible is not funny, that or i nerf hlin

## Why It's Good For The Game
shield gl shouldn't be a thing

## Changelog
:cl:
balance:t-71 requires two hands to use
/:cl:


BraveMole:
Ah yes, the famous one handed GL but it's two handed;

Don't mistake it for the two handed GL but it's also two handed, and it can stocks 6 nades, totally not the same

Pariah919:
mjp willl actually fix this
more at 11

PR #7117, opened 2021-06-17
Remove Explosive Vests from RO vendor

Wayland-Smithy:
## About The Pull Request

Removes vests from ROs vendor. 

Adds a Crate type for them so perhaps in the future they can be added to a spawner in the future, like in the Self Destruct areas perhaps? or spawned in bulk by an admin.

## Why It's Good For The Game

Vests are just too powerful especially during high-pop allowing them to trade at up to a 1:4 marine to xeno ratio when each marine is worth just 1/2.5 the xeno. That means at 4 kills (which is on the high end, but not so hard with high population and time dilation) you are taking out 10 marines worth of xenos. Perhaps vests can return with a fresh balance pass to address this.

## Changelog
:cl:
balance: Explosive Vests are no longer in the RO vendor.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -333,7 +333,6 @@
 			/obj/item/ammo_magazine/shotgun/mbx900 = 2,
 			/obj/item/bodybag/tarp = 2,
 			/obj/item/explosive/plastique = 2,
-			/obj/item/clothing/suit/storage/marine/harness/boomvest = 20,
 			/obj/item/radio/headset/mainship/marine/alpha = -1,
 			/obj/item/radio/headset/mainship/marine/bravo = -1,
 			/obj/item/radio/headset/mainship/marine/charlie = -1,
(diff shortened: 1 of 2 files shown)


SplinterGP:
req maybe ?

Wayland-Smithy:
I'd rather not as they can still just buy a bunch round start unless I were to make them silly expensive. Unless there is a way to time lock them.

Like I already said admins can easily spawn a bunch during shipside or they can be reintroduced via a spawner in SD or after some more balance consideration I am not putting effort into at the moment.

DerrickChamberlain:
LAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAME

spookydonut:
@DerrickChamberlain last warning stop shitposting

Surrealaser:
I would just increase the fuse timer to 2 seconds.

Wayland-Smithy:
I'll think about it, fell free to do it too and `closes:` this.

Al-1ce:
I hate to agree here. Vests do an insane amount of damage for what is inexpensive and likely to happen anyway. ( an unga dying to stupidity )

I'm not certain on the exact number of damage a vest does but it is certainly enough to one-shot most ancient T2s at close range in my experience.

Naaanii:
Decent solution if no one plans on fixing them.

Wayland-Smithy:
not needed at the moment

PR #7128, opened 2021-06-18
Fix shooting behind with autofire

BraveMole:
## About The Pull Request

Fixes this : 

https://user-images.githubusercontent.com/25491240/122566660-17c5e500-d048-11eb-9926-84a9b31274c4.mp4

Makes it so you can no longer shoot on your own tile (which can lead to shoot behind you if you are walking)

## Why It's Good For The Game

It's not perfect, but i think it's better to prevent firing behind than what is currently is.

Don't think i can do better because of byond

## Changelog
:cl:
fix: Fix bug that caused people to shoot behind them while running
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -512,8 +512,6 @@ User can be passed as null, (a gun reloading itself for instance), so we need to
 		dual_wield = TRUE
 	if(gun_user.in_throw_mode)
 		return
-	if(gun_user.Adjacent(object)) //Dealt with by attack code
-		return
 	if(QDELETED(object))
 		return
 	var/list/modifiers = params2list(params)
@@ -537,6 +535,8 @@ User can be passed as null, (a gun reloading itself for instance), so we need to
 
 ///Set the target and take care of hard delete
 /obj/item/weapon/gun/proc/set_target(atom/object)
+	if(gun_user.Adjacent(object))
+		return
 	if(object == target)
 		return
 	if(target)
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/gun_system.dm
```diff
@@ -537,6 +535,8 @@ User can be passed as null, (a gun reloading itself for instance), so we need to
 
 ///Set the target and take care of hard delete
 /obj/item/weapon/gun/proc/set_target(atom/object)
+	if(gun_user.Adjacent(object))
```
MLGTASTICa, line 538: it will cancel it if they push foward and it doesn't update however ,i  think if you can't do better than this then to not change it..

BraveMole: it always update? i

code/modules/projectiles/gun_system.dm
```diff
@@ -665,7 +665,7 @@ and you're good to go.
 //----------------------------------------------------------
 
 /obj/item/weapon/gun/proc/Fire()
-	if(QDELETED(gun_user) || !ismob(gun_user) || !able_to_fire(gun_user) || !target)
+	if(QDELETED(gun_user) || !ismob(gun_user) || !able_to_fire(gun_user) || !target || gun_user.Adjacent(target))
```
SplinterGP, line 668: ???? This stops the gun from shooting if dude is on your side

SplinterGP (changes requested):

code/modules/projectiles/gun_system.dm
BraveMole: that's the point. That was already the case before, because you are not using Fire() to PB.

The only way you could fire on adjacent tiles was by moving towards the target, and this was a bug

PR #7132, opened 2021-06-19
Revert "Ports CBT Juke build"

spookydonut:
Reverts tgstation/TerraGov-Marine-Corps#7130

It broke tgs


SplinterGP:
We got more RAM from MSO and it doenst OOM error anymore when building.

PR #7139, opened 2021-06-21
Light combat pack in medic vendor is 25 points rather than 15

Pariah919:
## About The Pull Request
Look at the title. 25 rather than 15 is way more fair, as you actually have to sacrifice a decent chunk of stuff inorder to have more combat effectiveness.

## Why It's Good For The Game
Light combat pack shouldn't be this cheap.

## Changelog
:cl:
balance: light combat pack in the medic vendor is 25 points rather than 15
/:cl:


SpaceLoveSs13:
Maintainers should probably change the lables. @SplinterGP And about the pr itself, I would say 20 would be better @Pariah919. But it's only a guess because sometimes even 5 points can be bit too much. Regardless, good pr.

Pariah919:
broke

PR #7140, opened 2021-06-21
Alien Slashes Now Attack Bodybag Occupants

Surrealaser:
## About The Pull Request

Alien Slashes Now Attack Bodybag Occupants

## Why It's Good For The Game

Alien Slashes Now Attack Bodybag Occupants

## Changelog
:cl:
fix: Alien slashes now hit bodybag occupants.
/:cl:

### DIFF
--- a/code/game/objects/items/bodybag.dm
+++ b/code/game/objects/items/bodybag.dm
@@ -202,6 +202,8 @@
 	if(opened)
 		return FALSE // stop xeno closing things
 	X.do_attack_animation(src, ATTACK_EFFECT_CLAW)
+	if(bodybag_occupant) //if there's an occupant, slash his ass
+		bodybag_occupant.attack_alien(X)
 	open()
 	X.visible_message("<span class='danger'>\The [X] slashes \the [src] open!</span>", \
 		"<span class='danger'>We slash \the [src] open!</span>", null, 5)
--- a/code/game/objects/items/bodybag.dm
+++ b/code/game/objects/items/bodybag.dm
@@ -202,8 +202,7 @@
 	if(opened)
 		return FALSE // stop xeno closing things
 	X.do_attack_animation(src, ATTACK_EFFECT_CLAW)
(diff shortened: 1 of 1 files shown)


lKiseki:
Yes. All my yes.

RenGusta:
Won't this end up hitting hidden tarp users due to the tarp function being similar to bodybags?

BraveMole:
but why? just open then slash?

BraveMole:
Say it's because of tarp, because don't think you need to slash bodies in cry bags

Surrealaser:
> Say it's because of tarp, because don't think you need to slash bodies in cry bags

Well yes, but no bag should provide any slash protection.

RenGusta:
I still don't see whether or not this will impact bump slash, I can only hope it doesn't?

Otherwise this should probably be labelled balance, not fix.

Surrealaser:
> I still don't see whether or not this will impact bump slash, I can only hope it doesn't?
> 
> Otherwise this should probably be labelled balance, not fix.

Bump slashing has literally nothing to do with this.

EgorDinamit:
What about stasis bags? Can xenos open those?
If not, you could just hide in one and be revivable for eternity.

Surrealaser:
> What about stasis bags? Can xenos open those?
> If not, you could just hide in one and be revivable for eternity.

Yes

code/game/objects/items/bodybag.dm
```diff
@@ -202,6 +202,8 @@
 	if(opened)
 		return FALSE // stop xeno closing things
 	X.do_attack_animation(src, ATTACK_EFFECT_CLAW)
+	if(bodybag_occupant) //if there's an occupant, slash his ass
+		bodybag_occupant.attack_alien(X)
```
SplinterGP, line 206:
```suggestion
	bodybag_occupant?.attack_alien(X)
```

SplinterGP (changes requested):

PR #7147, opened 2021-06-22
Fix OB beacons being activatable on dropships like Alamo

Wayland-Smithy:
## About The Pull Request

Adds a check for `/area/shuttle/dropship` when activating because just checking the z level is not enough it seems when that can change by going groundside. An alternate solution is to have them auto de-activate in an `onShuttleMove()` override.

## Why It's Good For The Game

OB beacons are meant to be used "on the planet" and as such the only way to get one planted on the Alamo is to fly it groundside first then attempt to plant a beacon, however that beacon stays activated even after hijack for funny dropship OB memes. This change rectifies that.

## Changelog
:cl:
fix: Fixed OB beacons being activatable on dropships.
/:cl:

### DIFF
--- a/code/game/objects/machinery/squad_supply/supply_beacon.dm
+++ b/code/game/objects/machinery/squad_supply/supply_beacon.dm
@@ -34,6 +34,9 @@
 	if(A && istype(A) && A.ceiling >= CEILING_DEEP_UNDERGROUND)
 		to_chat(H, "<span class='warning'>This won't work if you're standing deep underground.</span>")
 		return FALSE
+	if(istype(A, /area/shuttle/dropship))
+		to_chat(H, "<span class='warning'>You have to be outside the dropship to use this or it won't transmit.</span>")
+		return FALSE
 	var/delay = max(1.5 SECONDS, activation_time - 2 SECONDS * H.skills.getRating("leadership"))
 	H.visible_message("<span class='notice'>[H] starts setting up [src] on the ground.</span>",
 	"<span class='notice'>You start setting up [src] on the ground and inputting all the data it needs.</span>")


SpaceLoveSs13:
But why should the strategy exactly be removed though?

Wayland-Smithy:
It's unintended, and this is just a fix.

How come you can't activate an OB beacon on the Alamo regardless if you just want to target the LZ? Like right before the xeno ruler calls down the dropship you aren't supposed to be preempting anyway.

SpaceLoveSs13:
> It's unintended, and this is just a fix.
> 
> How come you can't activate an OB beacon on the Alamo regardless if you just want to target the LZ? Like right before the xeno ruler calls down the dropship before a hijack you aren't supposed to be preempting anyway.
> 
> Also I do find it a real issue how so many ops end with LZ OB spam after marines evac groundside. OBs that surly could have been used to not lose groundside in the first place, like HE and Incendiary, because this behavior is enabled and encouraged by being able to spam the LZ and dropship with beacons, now they will have to actually place them on the _ground_ and I doubt much will change at all except for that one SL who will be mildly surprised their strat to bypass the beacon z level restrictions no longer works.

Gotchya, the reason makes sense.

Ieteihowmee:
Absolutely not. You do understand the purpose of spamming OBs is a delay tactic to buy time for shipside combat prep time. Just removing more capacity for marines to prep counters. Just CAP if you do not want the round to go on longer.

Wayland-Smithy:
Lmao. You are totally ignoring the entire intent of only being able to activate beacons on the goundside z-level.

If it was intended Alamo would have been excepted from the check.

SplinterGP (approved):

PR #7151, opened 2021-06-22
Fixes Right Click Mode for Xenos

Surrealaser:
## About The Pull Request

Fixes Right Click mode for Xenos so that it doesn't try to auto-attack at the same time it uses an ability (thus triggering bump slash cooldowns).

Also stops turf targeting with melee attacks from triggering bump slash cooldowns.

## Why It's Good For The Game

Fixes Right Click mode for Xenos so that it doesn't try to auto-attack at the same time it uses an ability (thus triggering bump slash cooldowns). This will allow things like the Wraith's Blink to be used much more fluidly with bump attacks and slashes in general.

Also stops turf targeting with melee attacks from triggering bump slash cooldowns.

## Changelog
:cl:
fix: Fixes Right Click mode for Xenos so that it doesn't try to attack at the same time it uses an ability (thus triggering click/bump slash cooldowns).
fix: Turf targeting with melee attacks no longer triggers click/bump slash cooldowns.
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -334,7 +334,7 @@ if(selected_ability.target_flags & flagname){\
 		if(selected_ability.can_use_ability(A))
 			selected_ability.use_ability(A)
 			return TRUE
-	. = ..()
+	return ..()
 
 /*
 	Right click
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -329,12 +329,12 @@ if(selected_ability.target_flags & flagname){\
 		selected_ability.use_ability(A)
 
 /mob/living/carbon/xenomorph/RightClickOn(atom/A)
+	. = ..()
 	if(selected_ability) //If we have a selected ability that we can use, return TRUE
 		A = ability_target(A)
 		if(selected_ability.can_use_ability(A))
 			selected_ability.use_ability(A)
 			return TRUE
-	return ..()
 
 /*
 	Right click
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -329,12 +329,12 @@ if(selected_ability.target_flags & flagname){\
 		selected_ability.use_ability(A)
 
 /mob/living/carbon/xenomorph/RightClickOn(atom/A)
+	if(selected_ability) //If we have a selected ability that we can use, return TRUE
+		A = ability_target(A)
+		if(selected_ability.can_use_ability(A))
+			selected_ability.use_ability(A)
+			return TRUE
 	. = ..()
-	if(!selected_ability)
-		return FALSE
-	A = ability_target(A)
-	if(selected_ability.can_use_ability(A))
-		selected_ability.use_ability(A)
 
 /*
 	Right click
--- a/code/_onclick/xeno.dm
+++ b/code/_onclick/xeno.dm
@@ -3,10 +3,7 @@
 		return FALSE
 
 	if(!isturf(A)) //We don't care about turfs; they don't trigger our melee click cooldown
-		if(xeno_caste)
-			changeNext_move(xeno_caste.attack_delay)
-		else
-			changeNext_move(CLICK_CD_MELEE)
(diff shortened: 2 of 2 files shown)


vilereaver41:
Omg you listened to me for once! Thank you

code/_onclick/click.dm
```diff
@@ -329,12 +329,12 @@ if(selected_ability.target_flags & flagname){\
 		selected_ability.use_ability(A)
 
 /mob/living/carbon/xenomorph/RightClickOn(atom/A)
+	if(selected_ability) //If we have a selected ability that we can use, return TRUE
+		A = ability_target(A)
+		if(selected_ability.can_use_ability(A))
+			selected_ability.use_ability(A)
+			return TRUE
 	. = ..()
```
BraveMole, line 337:
```suggestion
	return ..()
```

BraveMole (changes requested):

BraveMole:
I think the correct fix would be for attack_alien to return true if the attack succeeded / had any effect, and to add the cooldown only if it returned TRUE

Surrealaser:
> I think the correct fix would be for attack_alien to return true if the attack succeeded / had any effect, and to add the cooldown only if it returned TRUE

Yes, but does attack_alien return true?

That was my first thought, but man, I really don't wanna fucking go through all those instances adding this.

BraveMole:
Did you check if non adjacent atom are procing the cooldown?

BraveMole (approved):

Wayland-Smithy (approved):
Tested locally. Works as advertised. I would also note that if the `can_use_ability()` check fails you still do a `attack_alien()`

code/_onclick/click.dm
```diff
@@ -329,12 +329,12 @@ if(selected_ability.target_flags & flagname){\
...
-		return FALSE
-	A = ability_target(A)
-	if(selected_ability.can_use_ability(A))
-		selected_ability.use_ability(A)
+	if(selected_ability) //If we have a selected ability that we can use, return TRUE
+		A = ability_target(A)
+		if(selected_ability.can_use_ability(A))
+			selected_ability.use_ability(A)
+			return TRUE
+	return ..()
```
SplinterGP, line 337: i think you should still call parent in both cases since it calls a signal but not sure

SplinterGP (changes requested):

Surrealaser:
Would like to get maintainer feedback because the unconditional click cooldown is cancer. At least a TM please.

code/_onclick/xeno.dm
```diff
@@ -2,10 +2,8 @@
 	if(lying_angle)
 		return FALSE
 
-	if(xeno_caste)
-		changeNext_move(xeno_caste.attack_delay)
-	else
-		changeNext_move(CLICK_CD_MELEE)
+	if(!isturf(A)) //We don't care about turfs; they don't trigger our melee click cooldown
+		changeNext_move(xeno_caste ? xeno_caste.attack_delay : CLICK_CD_MELEE)
```
SplinterGP, line 6: cant this trigger on walls ? but i think xeno walls arent turfs right?

code/_onclick/click.dm
SplinterGP: above ?

SplinterGP (changes requested):

code/_onclick/xeno.dm
Surrealaser: I can differentiate for xeno wall turfs if deemed necessary

Wayland-Smithy: resin walls are closed turfs that's why they block tad, why would they need to right-clicked anyway tho?

code/_onclick/click.dm
TiviPlus: Correct this must always parentcall as it has a signal
You could instead of doing this make the selected abilities register a signal on right click that returns one of the flags that cancels the click

code/_onclick/xeno.dm
TiviPlus: Changenext move should trigger on turf clicks make this isopenturf maybe

PR #7152, opened 2021-06-22
HvH mode gives human the ability to see in the dark

BraveMole:
## About The Pull Request

Space Magic, they can see in dark

## Why It's Good For The Game

Fighting with all lights off will get boring quickly

## Changelog
:cl:
expansion: HvH mode gives human the ability to see in the dark
/:cl:


AndroidMK2:
I would actually make a change that the power at any map is fixed when it's HvH and maybe (big maybe) make the lights indestructible then too. It would be fun to still be able to sneak up on someone.

SplinterGP:
> 
> 
> I would actually make a change that the power at any map is fixed when it's HvH and maybe (big maybe) make the lights indestructible then too. It would be fun to still be able to sneak up on someone.

good idea for me

PR #7156, opened 2021-06-22
Fixes infinite jelly duration

### LINKED ISSUE #7054: Infinite fireproofing bug
## Reproduction:

Make a resin jelly, leave it on ground , get 2 xenos to use it at once ,the first to complete the action gets the fireproofing effect permanently

QualityVan:
Refactors resin jelly (that red buff you can slap on yourself from hivelords) into a status effect and condenses a bit of fire code for xenos. Also makes jelly heal some burns while active because I wanted to make use of status effect's ticks.

Fixes #7054 

:cl:
fix: Jelly shouldn't last forever any more in certain cases.
code: Refactored jelly to an effect, plus some fire resistance cleanup.
balance: Jelly heals a bit of burns while active.
/:cl:


code/game/objects/structures/xeno.dm
```diff
@@ -1053,12 +1053,10 @@ TUNNEL
...
-	addtimer(CALLBACK(src, .proc/deactivate_jelly, user), immune_time)
-
-/obj/item/resin_jelly/proc/deactivate_jelly(mob/living/carbon/xenomorph/user)
-	user.remove_filter("resin_jelly_outline")
-	user.fire_resist_modifier += 20
-	to_chat(user, "<span class='xenonotice'>We feel more vulnerable again.</span>")
+	addtimer(CALLBACK(user, /mob/living/carbon/xenomorph/proc/deactivate_jelly), immune_time)
 	qdel(src)
+
+/mob/living/carbon/xenomorph/proc/deactivate_jelly()
```
BraveMole, line 1059: You need to autodoc new procs.

so here you do this

///Remove the jelly effect from the xenomorph
/mob/living/carbon/xenomorph/proc/deactivate_jelly()

And yes, three /

BraveMole (changes requested):

SplinterGP:
make resin jelly use status_effects() from code/datums/status_effects instead

code/datums/status_effects/xeno_buffs.dm
```diff
@@ -0,0 +1,25 @@
+/datum/status_effect/resin_jelly_coating
+	id = "resin jelly"
+	duration = 150
```
SplinterGP, line 3:
```suggestion
	duration = 15 SECONDS
```

SplinterGP (changes requested):

SplinterGP (approved):

PR #7013, opened 2021-06-01
Rebalance Snub-nosed Revolvers

Wayland-Smithy:
## About The Pull Request

Makes regular revolver ammo only stagger and knockback if fired via a long barrel. Moves shake to Mateba which it did not have before.

Call it a reactionary kneejerk salt PR all you want, but the ocelot gimmick is getting real aggravating to fight despite the new downsides of backpack revolvers with the stamina buffs.

## Why It's Good For The Game

2 revolver hits = 1 slug of stagger but you can fire 3.5 slugs worth and just keep dropping and pulling out more "slug shooters" making it strictly better than a shotgun for stagger, knockback, slow, and shake effects.

Stagger is a literal meme so much so Queen is snowflake immune:
```dm
var/stagger_immune = FALSE
if(isxeno(carbon_victim))
	var/mob/living/carbon/xenomorph/xeno_victim = victim
	if(isxenoqueen(xeno_victim)) //Stagger too powerful vs the Queen, so she's immune.
		stagger_immune = TRUE
```
..and it silences abilities until a periodic Life() tick comes along and decays your stagger stacks that revolver fire hoses onto you back down to 0.

This is not fun to prep or face:
image

Why are we letting a *sidearm* be so viable as a main weapon that effectively never needs to be reloaded?

Shorter barrel = less energy imparted on fired rounds, and should equal less impact on the target(s) too.

## Changelog
:cl:
balance: Snub-nosed TP-44 Revolvers no longer apply stagger or knockback to targets. Moves shake to Mateba which it did not have before.
/:cl:

PR #7162, opened 2021-06-23
Fix my ammo datum changes for real this time

Wayland-Smithy:
## About The Pull Request

The non-awaited sequel to #7104 and #7013 where I actually remove shake as was previously taken for granted except that arg specifically does not default to 0 and is at the end of the autodoc info.

## Why It's Good For The Game

Make it true:
image

## Changelog
:cl:
fix: Fixed "standard revolver bullet" still having shake.
/:cl:


SplinterGP:
add changelog telling that it actually does it now.

Wayland-Smithy:
if I must <insert RDJ rolling eyes gif here but it somehow does not get flagged as off topic>

PR #7173, opened 2021-06-24
heal beam heals less sunder

Pariah919:
## About The Pull Request
heal beam sunder heal dropped to -.6 from 1.7, as it heals too much has to little cd for so much heal stacking

## Why It's Good For The Game
hivvelord makes sunder a near irrelevant mechanic in a lot of scenarios
i think this is fair

## Changelog
:cl:
balance: hive heal beam heals 66% less sunder
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -334,7 +334,7 @@
 
 	new /obj/effect/temp_visual/telekinesis(get_turf(patient)) //Visual confirmation
 
-	patient.adjust_sunder(-1.8 * (1 + patient.recovery_aura * 0.05)) //5% bonus per rank of our recovery aura
+	patient.adjust_sunder(-0.6 * (1 + patient.recovery_aura * 0.05)) //5% bonus per rank of our recovery aura
 
 
 /obj/screen/alert/status_effect/healing_infusion


Surrealaser:
Not necessary. What the beam has in sunder it loses in healing efficiency vs the Drone.

EvilMelonTheSecond:
Oh dang a sunder change. I thought that xeno healing sunder instantly was the intended state of sunder. Anything below a SADAR hit will heal sunder faster than health and xeno never get any attrition.

If not this likely won't be enough to change that but at least hivelord running around slapping cure-all magic on everthing it can see is less overpowered

Al-1ce:
Brainless Xeno nerf consequentially turning into a marine buff. If you're going to be biased about your PRs just be honest.

Carnage068:
By 66%? What the fuck? I agree that it could definitely use a nerf, but I have no idea why some people just like to pull huge numbers out of nowhere. This is how things go from OP to trash, then trash to OP again. The amount of sunder it heals should only be slightly more than half the number you're changing it to.

Passe2142:
It is op if the hivelord knows what is doing, maybe some tuning instead of a hard nerf.

SpaceLoveSs13:
Good idea, terrible execution. Bad hsrd nerf.

Ieteihowmee:
Nope.

TiviPlus:
I dont see the use in this and the nerf is too extrmeme as is

PR #7177, opened 2021-06-24
Slightly buff the fire rate of sectoids rifle.

lbnesquik:
## About The Pull Request
Simply put, i felt that this ERT was a bit of a pushover, especially given how they do not really have a magnetic harness in case they get knocked down.
Plus, they are freaking aliens with advanced technology! Their stuff should be good!

## Changelog
:cl:
balance: Slightly increase sectoids fire rate.
/:cl:


lbnesquik:
That other PR was really messed up as made a master commit which is messing everything up till it is merged.

PR #7184, opened 2021-06-24
Psy points output depends on the number of players

BraveMole:
## About The Pull Request

Continuation of the last distress rework:

The number of psy points generated by cocoon and psy drain is now linked with the number of players.

The more marines are on the ground, the less points these abilities gives.

On the contrary, generator output will now be independant of the number of marines, as long as some marines are on the ground

## Why It's Good For The Game

Since the output of silo is now proportional of the number of marines on the ground, not compensating on the psy points output means that small hive will have issues having new silos, while big hive will have a ton of silo

This is an issue, because big hive will have way more larva proportionally than small hives.

This solution is not perfect, but it's better than nothing

## Changelog
:cl:
balance: The number of psy points generated by cocoon and psy drain is now linked with the number of players. The more marines are on the ground, the less points these abilities gives.
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -127,12 +127,22 @@
 #define SILO_PRICE 800
 #define XENO_TURRET_PRICE 100
 #define XENO_KING_PRICE 1800
-//How many psych point one gen gives per person on the server
-#define BASE_PSYCH_POINT_OUTPUT 0.008
-//How many psy points are gave for each marine psy drained
-#define PSY_DRAIN_REWARD 60
-//How many psy points are gave every 5 second by a cocoon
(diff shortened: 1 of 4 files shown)


Lefunny:
btw how much it will fuck highpop? what would be a highpop hive? 15 xenos? 20?, also what is the sense of a big hive being....Bigger? with more silos and etc while a small hive being....Small? with less silos, turrets and etc?

BraveMole:
I consider 50 marines as high pop.

David-Stormwell (approved):
This does sound fine in theory, though not having a preset drain number will be slightly painful I'll know later on.
Just to make sure does this count dead marines?

PR #7192, opened 2021-06-26
medical belts no longer hold magazines, at all

Pariah919:
## About The Pull Request
medical belts can no longer hold magazines, at all. So you can't hold a gazillion smg mags in a medical belt. That isn't intended.

## Why It's Good For The Game
important fix

## Changelog
:cl:
fix: Med belts no longer hold smg magazines
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -110,12 +110,6 @@
 		/obj/item/clothing/mask/surgical,
 		/obj/item/clothing/gloves/latex,
 		/obj/item/storage/syringe_case,
-		/obj/item/ammo_magazine/pistol,
-		/obj/item/ammo_magazine/smg/m25,
-		/obj/item/ammo_magazine/smg/standard_smg,
-		/obj/item/ammo_magazine/smg/standard_machinepistol,
-		/obj/item/ammo_magazine/revolver,
-		/obj/item/ammo_magazine/handful,
 		/obj/item/flashlight/flare,
 		/obj/item/explosive/grenade/flare,
 		/obj/item/reagent_containers/hypospray,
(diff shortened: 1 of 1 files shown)


EvilMelonTheSecond:
Where the hell do I store my SMG mags then, the standard mag belt that holds only 5 despite their small size? Gimme a solution before fixing a problem born from another problem.

AnCopper:
> Where the hell do I store my SMG mags then, the standard mag belt that holds only 5 despite their small size? Gimme a solution before fixing a problem born from another problem.

You hold them in your mag belt.

David-Stormwell:
> Where the hell do I store my SMG mags then, the standard mag belt that holds only 5 despite their small size? Gimme a solution before fixing a problem born from another problem.

you hold them in the belt designed for holding magazines.

QualityVan:
How is it not intended when they're explicitly in the typelist and the item desc includes 'designed to transport medical supplies, and light ammunitions'?

SplinterGP:
> 
> 
> How is it not intended when they're explicitly in the typelist and the item desc includes 'designed to transport medical supplies, and light ammunitions'?

yeah but then you can store 19 or whatever slots that t hing has of ammo instead of only 5, gotta limit how many mags it fits

SplinterGP (commented):
limit how many mags you can put in to 5, so you cant remove meds and have a better mag belt

Swagile:
or improve the standard mag belt to 10 then do this nerf?

there's a reason why the LMG is constantly meta no matter what nerfs you put in it; cause storage for all the other viable guns is dogshit

🤷🏻

SplinterGP:
not a proper balance to it, the right way should be limitting how many mags it can put inside.

PR #7209, opened 2021-06-28
Give hud to SO and ST

BraveMole:
## About The Pull Request

Title

## Changelog
:cl:
qol: Give hud role icon to SO and ST
/:cl:


Ieteihowmee:
Probably for the best. Do STs have the same icon as CSE or are they getting something different? Blue background with a white wrench?

Issue #7191, opened 2021-06-26
M44 Holster Not accepting M44 SAA Revolver


Vesp1:
Says the revolver is too long, when a req officer spawns with it in the holster

PR #7212, opened 2021-06-29
Moves per-item storage limits up in inheritance

QualityVan:
## About The Pull Request
Pistol/revolver belts had a hard limit of only one gun at a time. This refactors that into an aspect of /obj/item/storage and generalizes it so any storage object can limit how many of a specific typepath is contained, separate from the normal size/count limits. A backpack could be limited to only have two hats in it at a time, for instance.

This checks via looped istype in order to maintain existing behavior on pistol belts, so if a pouch allowed two obj/item/weapon/gun and one obj/item/weapon/gun/pistol it can hold at most one pistol and one other gun, or two non-pistol guns.

Also moves holster icon updates from handled by inserted guns to handled by the holster itself, and removes the restriction on holsters saving one item slot for a gun rather than being entirely full of ammo.

## Why It's Good For The Game
Enables expanding or restricting storage more piecemeal. It will need to be actually added to individual storage containers to do anything, of course.

## Changelog
:cl:
code: Added the ability for all storage spaces to have item-specific limits
/:cl:


code/game/objects/items/storage/belt.dm
```diff
@@ -663,24 +662,17 @@
 	if(istype(user)) user.update_inv_belt()
 	if(istype(user)) user.update_inv_s_store()
 
+/obj/item/storage/belt/gun/handle_item_insertion(obj/item/item, prevent_warning = 0, mob/user)
+	. = ..()
+	if(. && istype(item, /obj/item/weapon/gun))
```
BraveMole, line 667: isgun exists IIRC

code/game/objects/items/storage/storage.dm
```diff
@@ -376,6 +377,14 @@
 				to_chat(usr, "<span class='notice'>[src] cannot hold [W] as it's a storage item of the same size.</span>")
 			return FALSE //To prevent the stacking of same sized storage items.
 
+	for(var/T in storage_type_limits)
```
BraveMole, line 380: better var name please

BraveMole:
Does this fix https://github.com/tgstation/TerraGov-Marine-Corps/issues/7191 and related issues? Seems that it doesn't

mrhugo13:
> 
> 
> Does this fix #7191 and related issues? Seems that it doesn't

doesn't seem so since it doesn't touch max_w_class, not to worry though i already have a fix on the way.

code/game/objects/items/storage/storage.dm
```diff
@@ -443,6 +452,9 @@
 	if (storage_slots)
 		item.mouse_opacity = 2 //not having to click the item's tiny sprite to take it out of the storage.
 	update_icon()
+	for(var/T in storage_type_limits)
+		if(istype(item, T))
+			storage_type_limits[T] -= 1
```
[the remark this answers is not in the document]
QualityVan: Not breaking there is intentional, since otherwise it leads to weird cases if you have one limited type that inherits from another.

QualityVan:
> Does this fix #7191 and related issues? Seems that it doesn't
Like hugo said it doesn't, that's just a mismatch between the storage's max_w and the stored object's w_class. Not sure what else'd be related.

code/game/objects/items/storage/storage.dm
```diff
@@ -12,6 +12,13 @@
 	var/list/can_hold = list() //List of objects which this item can store (if set, it can't store anything else)
 	var/list/cant_hold = list() //List of objects which this item can't store (in effect only if can_hold isn't set)
 	var/list/bypass_w_limit = list() //a list of objects which this item can store despite not passing the w_class limit
+	/*Associated list of types and their max count, formatted as
+	  	storage_type_limits = list(
+	  		/obj/A = 1,
+	  	)
+	  Any inserted objects will decrement the allowed count of every listed type which matches or is a parent of that object.
+	  With entries for both /obj/A and /obj/A/B, inserting a B requires non-zero allowed count remaining for, and reduces, both. */
+	var/list/storage_type_limits = list()
```
TiviPlus, line 21: A) autodoc
B) Could this not be a lazylist

TiviPlus (changes requested):

code/game/objects/items/storage/storage.dm
QualityVan: Looked over what lazylists are and changed it to one, they're new to me.

code/game/objects/items/storage/storage.dm
```diff
@@ -12,6 +12,13 @@
 	var/list/can_hold = list() //List of objects which this item can store (if set, it can't store anything else)
 	var/list/cant_hold = list() //List of objects which this item can't store (in effect only if can_hold isn't set)
 	var/list/bypass_w_limit = list() //a list of objects which this item can store despite not passing the w_class limit
+	/**Associated list of types and their max count, formatted as
+	  	storage_type_limits = list(
+	  		/obj/A = 1,
+	  	)
+	  Any inserted objects will decrement the allowed count of every listed type which matches or is a parent of that object.
+	  With entries for both /obj/A and /obj/A/B, inserting a B requires non-zero allowed count remaining for, and reduces, both. */
+	var/list/storage_type_limits = null
```
TiviPlus, line 21:
```suggestion
	  With entries for both /obj/A and /obj/A/B, inserting a B requires non-zero allowed count remaining for, and reduces, both. */
	var/list/storage_type_limits
```

code/modules/projectiles/gun_system.dm
```diff
@@ -1124,23 +1124,6 @@ and you're good to go.
...
-			GB.current_gun = src //If there's no active gun, we want to make this our icon.
-			GB.update_gun_icon()
-
-/obj/item/weapon/gun/on_exit_storage(obj/item/I)
-	if(istype(I,/obj/item/storage/belt/gun))
-		var/obj/item/storage/belt/gun/GB = I
-		GB.holds_guns_now--
-		if(GB.current_gun == src)
-			GB.current_gun = null
-			GB.update_gun_icon()
```
TiviPlus, line 1142: Why are yoou deleting these two procs though

code/game/objects/items/storage/belt.dm
```diff
@@ -929,4 +931,5 @@
 	. = ..()
 	var/obj/item/weapon/gun/new_gun = new /obj/item/weapon/gun/shotgun/double/marine(src)
 	new /obj/item/ammo_magazine/shotgun(src)
-	new_gun.on_enter_storage(src)
+	current_gun = new_gun
+	update_gun_icon()
```
TiviPlus, line 935: This should just use the on_enter storage stuff

TiviPlus (changes requested):

code/game/objects/items/storage/belt.dm
QualityVan: I figured it was better to keep the belt's behavior in the belt itself

TiviPlus: No?
Still pending

code/game/objects/items/storage/storage.dm
```diff
@@ -12,6 +12,13 @@
 	var/list/can_hold = list() //List of objects which this item can store (if set, it can't store anything else)
 	var/list/cant_hold = list() //List of objects which this item can't store (in effect only if can_hold isn't set)
 	var/list/bypass_w_limit = list() //a list of objects which this item can store despite not passing the w_class limit
+	/**Associated list of types and their max count, formatted as
+	  	storage_type_limits = list(
+	  		/obj/A = 1,
+	  	)
+	  Any inserted objects will decrement the allowed count of every listed type which matches or is a parent of that object.
+	  With entries for both /obj/A and /obj/A/B, inserting a B requires non-zero allowed count remaining for, and reduces, both. */
```
TiviPlus, line 20: This does not use autodoc properly

See other examples and pins in the dev channels

TiviPlus (commented):
Rest is fine

TiviPlus (approved):

PR #7105, opened 2021-06-15
Revolver ammo rebalance, seasonal M44 bag nerf

Wayland-Smithy:
## About The Pull Request

Fixed the full stagger effects being applied to `/datum/ammo/bullet/revolver/small/` and `/datum/ammo/bullet/revolver/ricochet/`

Increased the size of the M-44 SAA for it shoots full stagger revolver ammo without a snub-nose variant.

## Why It's Good For The Game

QHuzWEfGOB

## Changelog
:cl:
balance: S&W .357 Revolver ricochet rounds no longer stagger or knockback.
balance: M-44 SAA Revolver is now bulky like an extended barrel TP-44.
balance: M-44 SAA Revolver now does 5 more damage.
/:cl:

PR #7216, opened 2021-06-29
4 Fixes in one PR

### LINKED ISSUE #7020: T19 UB attachment sprites are offset and don't actually connect with the gun
uh, Title

## Reproduction:

1. Grab a T19
2. Put a LS or a Vgrip on it
3. look at it
Screenshot 2021-06-03 040735

### LINKED ISSUE #7191: M44 Holster Not accepting M44 SAA Revolver

### LINKED ISSUE #7201: Chemrette Packets disappearing
## Reproduction:
Take any item out of a chemrette packet, and the packet sprite disappears. Alt clicking the tile the packet is on lets you see it where you'd see all the items on that tile. And you can still remove all the cigs from it.

mrhugo13:
## About The Pull Request
Fixes #7201 by changing the icon state name in-DMI and then changed the icon_state and item_state accordingly.
Presumably this one was ironically caused by #6317 in such a way that chempackets base sprite name did not line up with the other sprite names.

Fixes #7020 by correcting the under_y variable which was set incorrectly, also fixed the muzzle_y to the correct offset since the original turned out to be wrong aswell while testing, i do not know what caused them to be set incorrectly because those variables were last touched when they were made 2 years ago.

And finally, fixes #7191 because the M44 had a higher weight class than the belt allowed so i increased the maximum weight class allowed to 4 (which is the bulky class that the M44 is), also increased the maximum storage space on it by 1 so that it could still fit in alongside all the speedloaders.
Caused by #7105 that increased the weight class which also made it not possible to put in the belt due to the maximum weight class being lower than the revolver itself.

Also fixed the mateba revolver belt using the increased maximum storage space thingy like i did with the M44.

## Why It's Good For The Game
These fixes polish gameplay and sprites so that it makes the greater TGMC experience generally better, dissapearing sprites aren't good, attachments that aren't actually attached on the sprites aren't good, and a belt that can't actually fit the gun it was designed for isn't good.

## Changelog
:cl: Vondiech
fix: You can now put the M44 into the belt again.
fix: Chemrette's no longer dissapear when you take something out of it.
fix: The T-19 doesn't have floating barrel and under attachments anymore.
fix: The Mateba revolver belt now can be used at the max capacity.
/:cl:


TiviPlus (approved):
I am going to merge this but if your prs arent atomized in future they'll be ade to be just fyi

PR #7219, opened 2021-06-29
Slight RR buff.

Chevalt:
I'm proposing this change as a way to make recoiless rifles slightly more useful, as right now, one shell of high explosive equals four shots out of a standard rifle, with only the sundering and slight explosive impact as very minor advantages over said rifles. The main thing they lacked was damage, thus I have slightly buffed it, notably the LE shell by 15. And HE shell by 40. It had a base 100 damage. For example, a standard sadar shell AP shell does about 325 damage.

## Why It's Good For The Game

Currently recoiless rifles are very, very weak. Speaking from experience, a direct hit with a HE shell will only barely damage a T3, remove maybe a quarter of a T2, and sometimes half of a T1. This is a minor change, however it will make them a somewhat better weapon overall, that isn't beaten by simply using a rifle over it.

## Changelog
:cl:
balance: rebalances the RR rifle to do slightly more damage. 140 instead of a base 100 with HE shells. LE shells do 90 instead of 75 with no changes to anything else.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1182,7 +1182,7 @@ datum/ammo/bullet/revolver/tp44
 	accuracy = 40
 	accurate_range = 20
 	max_range = 30
-	damage = 100
+	damage = 140
 	penetration = 50
 	sundering = 50
 
@@ -1220,7 +1220,7 @@ datum/ammo/bullet/revolver/tp44
 	accuracy = 40
 	accurate_range = 15
 	max_range = 20
-	damage = 75
+	damage = 90
 	penetration = 50
 	sundering = 25
 


MLGTASTICa:
Do your changelog properly , it still appears ingame.

Chevalt:
> Do your changelog properly , it still appears ingame.

Mind explaining how? It’s been a while since I used git.

MLGTASTICa:
> > Do your changelog properly , it still appears ingame.
> 
> Mind explaining how? It’s been a while since I used git.

Just instead of "Rebalanced something" put Rebalanced RR to deal 40 more damage with X rocket"

Chevalt:
> > > Do your changelog properly , it still appears ingame.
> > 
> > 
> > Mind explaining how? It’s been a while since I used git.
> 
> Just instead of "Rebalanced something" put Rebalanced RR to deal 40 more damage with X rocket"

Alright, fixed.

Al-1ce:
> Currently recoiless rifles are very, very weak. Speaking from experience, a direct hit with a HE shell will only barely damage a T3, remove maybe a quarter of a T2, and sometimes half of a T1.


If we're going off of purely anecdotal experience, I can say with confidence that the RR has obliterated (see: gibbed) the vast majority of T1s, if not all of them should they be directly hit. 

T3s are not immune to the rockets, and you're exaggerating when you say T2s can shrug it off just like a quarter of damage. I've seen plenty of Warriors, Carriers, and sometimes wraiths get most of their HP taken out with a single well-placed shot from a RR.

To put that in short, your reason for this being good for the game is "in my completely subjective experience it doesn't do enough damage nor does it one-shot T3s!"

Chevalt:
> > Currently recoiless rifles are very, very weak. Speaking from experience, a direct hit with a HE shell will only barely damage a T3, remove maybe a quarter of a T2, and sometimes half of a T1.
> 
> If we're going off of purely anecdotal experience, I can say with confidence that the RR has obliterated (see: gibbed) the vast majority of T1s, if not all of them should they be directly hit.
> 
> T3s are not immune to the rockets, and you're exaggerating when you say T2s can shrug it off just like a quarter of damage. I've seen plenty of Warriors, Carriers, and sometimes wraiths get most of their HP taken out with a single well-placed shot from a RR, if not outright killed as well with the surrounding fire, which almost always tends to be the case.
> 
> To put that in short, your reason for this being good for the game is "in my subjective experience it doesn't do enough damage nor does it one-shot T3s!"

140 Damage as suggested by the buff is far from one-shooting T3s.
100 is just enough to remove most of the health of a T1, but that’s about it. You’ll do a third to most T2 due to the armour they have, and it also cannot kill T1s directly since they too have armour.
You can only use the weapon on T3s for sunder, it has no other uses beyond that on them. 
I think the weapon and especially the HE shell is a little too weak for the fact you only get three of them with the kit. And that there’s only two kits available without ordering them at req. 

Speaking only stat wise, and ignoring explosive resistance.
Most T2s have 200HP.
Most T1s have around 100HP.

at the maximum, you will remove half the health of a T2 that has no armour.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1182,7 +1182,7 @@ datum/ammo/bullet/revolver/tp44
 	accuracy = 40
 	accurate_range = 20
 	max_range = 30
-	damage = 100
+	damage = 140
```
SplinterGP, line 1185: Those are on impact hit the rockets also cause explosion on hit.

SplinterGP (changes requested):

code/modules/projectiles/ammo_datums.dm
Chevalt: This is only meant for impact.

lbnesquik:
I would rather have more ammo rather, even at the cost of less effective ammo.

february-ops:
> > > Currently recoiless rifles are very, very weak. Speaking from experience, a direct hit with a HE shell will only barely damage a T3, remove maybe a quarter of a T2, and sometimes half of a T1.
> > 
> > 
> > If we're going off of purely anecdotal experience, I can say with confidence that the RR has obliterated (see: gibbed) the vast majority of T1s, if not all of them should they be directly hit.
> > T3s are not immune to the rockets, and you're exaggerating when you say T2s can shrug it off just like a quarter of damage. I've seen plenty of Warriors, Carriers, and sometimes wraiths get most of their HP taken out with a single well-placed shot from a RR, if not outright killed as well with the surrounding fire, which almost always tends to be the case.
> > To put that in short, your reason for this being good for the game is "in my subjective experience it doesn't do enough damage nor does it one-shot T3s!"
> 
> 140 Damage as suggested by the buff is far from one-shooting T3s.
> 100 is just enough to remove most of the health of a T1, but that’s about it. You’ll do a third to most T2 due to the armour they have, and it also cannot kill T1s directly since they too have armour.
> You can only use the weapon on T3s for sunder, it has no other uses beyond that on them.
> I think the weapon and especially the HE shell is a little too weak for the fact you only get three of them with the kit. And that there’s only two kits available without ordering them at req.
> 
> Speaking only stat wise, and ignoring explosive resistance.
> Most T2s have 200HP.
> Most T1s have around 100HP.
> 
> at the maximum, you will remove half the health of a T2 that has no armour.

There is still other marines that when you lose halr your half can easily kill you when you have almost nothing

PR #7220, opened 2021-06-29
Create Jelly and Place Jelly Pod Cooldown Change

Cowbin:
## About The Pull Request
Changes the cooldown of create jelly and place jelly pod.
Create Jelly: 1 min -> 5 sec
Place jelly Pod: 3 min -> 1 min

## Why It's Good For The Game
I believe this is good for the game as currently the cooldowns are much too long for a hivelord to actually be able to use them in active combat.  A five second jelly cooldown will allow a hivelord to, at max, jelly 3 xenos endlessly.  This will allow a hivelord in practice to keep two high tier xenos near them fireproof, while requiring a hivelord to make a jelly pod to cover low tier castes.  This ability to cover xenos in jelly was never really unbalanced when it was instant to create jelly, so a short cooldown is justified.

## Changelog
:cl:
balance: Changes jelly cooldown to 5 seconds and jelly pod cooldown to 1 min.
/:cl:


SplinterGP:
it being 5 seconds is way too low, thats 3/1 of the cooldown in jelly uptime meaning that you have 15 seconds of fire immunity for 5 seconds cooldown, the jelly pod change is good and 1minute and 30 seconds or 2 minutes would be more reasonable.

the jelly itself could be 45 seconds~ so 1/3 of cooldown uptime.

davidchan13:
Based change. Jelly did not have enough impact on games before the cooldown, the 3/1 minute cooldowns are not needed or representative of the effect Jelly has.

If @SplinterGP believes the cooldowns are where they should be, then jelly needs to be buffed to having 40+ second uptime.

Wayland-Smithy:
300% uptime is fine when it also has a plasma cost otherwise it's just a slightly faster version of the pods that you have to stack up somewhere and hope the other xenos go get when needed.

Wayland-Smithy (approved):
No code changes.

TiviPlus (approved):

PR #7226, opened 2021-06-30
Adds Health Huds to Clothing Vendors

Cowbin:
## About The Pull Request
Adds health huds to clothing vendors in non-limited amounts.

## Why It's Good For The Game
Currently health huds are limited in number, and thus are only there for use as Gamer Gear™.  By adding them in non-limited amounts it allows everyone to have access without having to bumrush them roundstart.  Plus this allows their use with the new automatic equipment machines.  Helps to remove some of prep as well.

## Changelog
:cl:
qol: Added health huds to clothing vendors
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -899,6 +899,7 @@
 			/obj/item/clothing/head/slouch = -1,
 			/obj/item/clothing/glasses/mgoggles = -1,
 			/obj/item/clothing/glasses/mgoggles/prescription = -1,
+			/obj/item/clothing/glasses/hud/health = -1,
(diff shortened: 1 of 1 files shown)


BraveMole:
You're not even talking about the effects that it would have on game balance.

image

image

This is just powercreep to the highest level

Cowbin:
Honestly I forgot to talk about possible effects on game balance.  I don't believe it will have too much of a major affect on game balance.  While seeing health will cause marines to be more cautious, good marine players already check their health regularly. 
 You are choosing to have less armor or no structure vision (mesons) in order to be able to see your health easier.  If mesons weren't bugged with lighting I would suggest adding them as well, so that it could be a tactical choice when eyeware you pick.  Plus it just makes gear that was limited in number available for everyone.

PR #7227, opened 2021-06-30
Hydro Cannon Fix

### LINKED ISSUE #6695: Hydrocannon button doesn't work
## Testmerges:

Don't think its caused by one

## Reproduction:

1. Click the hydrocannon button
2. Glitch will happen after a couple of tries. The chat will say both hydrocannon on and off and the flamer will still be flamin'

InterroLouis:
Fixes the Hydro Cannon

## About The Pull Request

Fixes the hydro cannon hud button. Now you can actually activate the hydro cannon by clicking the hud button. It also removes ATTACH_ACTIVATION from the hydro cannon, so the load from attachment verb will no longer toggle the hydro cannon. Unique action and the hud button will now be the only ways to toggle it, but I'm assuming people defaulted to unique action anyways. 
Fixes #6695 

## Why It's Good For The Game
It just works. That's really it, the hud button didn't work before and now it does. 
## Changelog
:cl:
fix: Fixed Hydro Cannon hud button.
refactor: Hydro Cannon only works via unique action and hud button, not load from attachment.
/:cl:

### DIFF
--- a/code/datums/actions/item_action.dm
+++ b/code/datums/actions/item_action.dm
@@ -117,7 +117,6 @@
 	RegisterSignal(holder_flamer, COMSIG_ITEM_HYDRO_CANNON_TOGGLED, .proc/update_toggle_button_icon)
 
 /datum/action/item_action/toggle_hydro/action_activate()
-	. = ..()
 	holder_flamer.unique_action(owner)
 
 /datum/action/item_action/toggle_hydro/update_button_icon()
--- a/code/datums/actions/item_action.dm
+++ b/code/datums/actions/item_action.dm
@@ -116,8 +116,6 @@
 	holder_flamer = holder_item
 	RegisterSignal(holder_flamer, COMSIG_ITEM_HYDRO_CANNON_TOGGLED, .proc/update_toggle_button_icon)
 
-/datum/action/item_action/toggle_hydro/action_activate()
-	. = ..()
 
 /datum/action/item_action/toggle_hydro/update_button_icon()
 	button.overlays.Cut()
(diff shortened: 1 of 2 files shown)


lbnesquik:
Based

code/datums/actions/item_action.dm
```diff
@@ -117,7 +117,6 @@
 	RegisterSignal(holder_flamer, COMSIG_ITEM_HYDRO_CANNON_TOGGLED, .proc/update_toggle_button_icon)
 
 /datum/action/item_action/toggle_hydro/action_activate()
-	. = ..()
 	holder_flamer.unique_action(owner)
```
TiviPlus, line 120: Explain this

InterroLouis: It was calling the above proc which ended up making the action happen twice.

TiviPlus: What above proc
Why didnt you just delete the unique action call

InterroLouis: I tried that and while it worked for making it not double, it didn't update the hud icon.

TiviPlus: This proc should be calling parent look into why its being called twice instead

SplinterGP: it uses ui_action_click()

code/datums/actions/item_action.dm
```diff
@@ -118,7 +118,6 @@
 
 /datum/action/item_action/toggle_hydro/action_activate()
 	. = ..()
```
TiviPlus, line 120:
```suggestion
```

TiviPlus (approved):

PR #7229, opened 2021-07-01
Fix window weeds not updating after the frame is gone [UTMOST PRIORITY]

Wayland-Smithy:
## About The Pull Request

9ayLZiqPrq

## Why It's Good For The Game

Annoying guessing game of is there a frame there or not goes away. (except if badmins del windows)

## Changelog
:cl:
fix: Fixed window weeds not updating after the frame is gone.
/:cl:

### DIFF
--- a/code/game/objects/structures/window_frame.dm
+++ b/code/game/objects/structures/window_frame.dm
@@ -56,10 +56,10 @@
 /obj/structure/window_frame/Destroy()
 	density = FALSE
 	update_nearby_icons()
-	for(var/obj/effect/alien/weeds/weedwall/WW in loc)
-		qdel(WW)
+	var/obj/effect/alien/weeds/weedwall/window_wall_weeds = locate() in loc
+	if(window_wall_weeds)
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/window_frame.dm
```diff
@@ -56,6 +56,10 @@
 /obj/structure/window_frame/Destroy()
 	density = FALSE
 	update_nearby_icons()
+	for(var/obj/effect/alien/weeds/weedwall/WW in loc)
+		qdel(WW)
+		new /obj/effect/alien/weeds(loc)
```
TiviPlus, line 61: Why is this a loop if a locate suffices

Wayland-Smithy: It could be a `locate() in loc` I suppose, but I just copy pasta from like 15 lines up:
```
		for(var/obj/effect/alien/weeds/weedwall/window/W in loc)
			weed_found = TRUE
			break
```
🤷

TiviPlus (approved):

PR #7242, opened 2021-07-02
Ravager plasma is now fury

SplinterGP:
## About The Pull Request
Ravager plasma now regens when they cause damage to a living mob and their plasma regen is limited up to 50% in young and mature and 60% for elder and ancient.

## Why It's Good For The Game
It's a new way of plasma working for ravager, letting them keep gaining plasma faster if they keep hitting their enemies, allowing them to keep gaining plasma if playing offensively.

## Changelog
:cl:
expansion: Ravager plasma now regens when they cause damage on an enemy.
balance: Ravager plasma regen reduced.
balance: Ravager plasma now regens even off weeds.
balance: Ravager plasma is now limited to regenning up to 50% until elder+ which is 60%
imageadd: Ravager now has a red plasma bar to imply fury.
/:cl:


code/datums/components/plasma_on_attack.dm
```diff
@@ -0,0 +1,18 @@
...
+	if(!isxeno(parent))
+		return COMPONENT_INCOMPATIBLE
+	RegisterSignal(parent, COMSIG_XENOMORPH_ATTACK_LIVING, .proc/damage_dealt)
+
+/datum/component/plasma_on_attack/proc/damage_dealt(datum/source, mob/living/attacked, damage)
+	var/mob/living/carbon/xenomorph/furious = source
+	furious.gain_plasma(damage * damage_plasma_multiplier)
+
+
+/datum/component/plasma_on_attack/ravager
```
spookydonut, line 14: use component args not subtypes

code/datums/elements/plasma_on_attack.dm
```diff
@@ -0,0 +1,17 @@
+/datum/element/plasma_on_attack
+	element_flags = ELEMENT_BESPOKE
+	id_arg_index = 2
+
+	var/damage_plasma_multiplier = 1
```
BraveMole, line 5: autodoc

code/datums/elements/plasma_on_attack.dm
```diff
@@ -0,0 +1,17 @@
...
+
+/datum/element/plasma_on_attack/Attach(datum/target, damage_plasma_multiplier)
+	. = ..()
+	if(!isxeno(target))
+		return ELEMENT_INCOMPATIBLE
+	RegisterSignal(target, COMSIG_XENOMORPH_ATTACK_LIVING, .proc/damage_dealt)
+	src.damage_plasma_multiplier = damage_plasma_multiplier
+
+/datum/element/plasma_on_attack/proc/damage_dealt(datum/source, mob/living/attacked, damage)
```
BraveMole, line 14: autodoc

code/datums/elements/plasma_on_attack.dm
```diff
@@ -0,0 +1,17 @@
...
+	id_arg_index = 2
+
+	var/damage_plasma_multiplier = 1
+
+/datum/element/plasma_on_attack/Attach(datum/target, damage_plasma_multiplier)
+	. = ..()
+	if(!isxeno(target))
+		return ELEMENT_INCOMPATIBLE
+	RegisterSignal(target, COMSIG_XENOMORPH_ATTACK_LIVING, .proc/damage_dealt)
+	src.damage_plasma_multiplier = damage_plasma_multiplier
```
BraveMole, line 12: ? Doesn't this mean that if one changes this number, everyone has it changed?

BraveMole (changes requested):

code/datums/elements/plasma_on_attack.dm
SplinterGP: Bespoke

code/datums/elements/plasma_on_attack.dm
SplinterGP: don't think i need to autodoc this since its a very small proc and what it does is pretty apparent

TiviPlus: The rule is autodoccing everything though

TiviPlus: also signal handler missing

TiviPlus (approved):

Carnage068:
Ravager can't use Eviscerating Charge anymore, because it's incapable of having enough plasma to do so. Why exactly did you do this, again?

Al-1ce:
This _sounds_ like a good idea - don't get me wrong, it does. Tailoring the mechanics to support and encourage a certain type of gameplay is great, but I feel this is heavily reliant on an actual consistent amount of marines getting attacked or being out in the open.

I'll elaborate further on that - since marines are "always out in the open", right?
Well, sorta. I've seen about fifty separate rav players who initiate their attacks **with** the charge ability, not using it after starting with a regular slash or two. This creates an issue that discourges the rav from actually engaging their target, as they cannot use their abilities to initiate the combat to begin with!

That's not to say the abilities are required to fight, but many heavily rely on them to initiate their strikes. This is the inherent flaw I can see with your PR, as well as what Carnage mentioned.

SplinterGP:
> 
> 
> Ravager can't use Eviscerating Charge anymore, because it's incapable of having enough plasma to do so. Why exactly did you do this, again? Like, it literally doesn't have enough plasma to use the ability, unless you run through 5 tiles of fire to get the plasma boost from being on fire. That's the only way you will ever be full Plasma again. That, and when you upgrade to Elder, and Ancient.

attack a marine ? i dind't run on that issue often because i was able to go hit marines atleast once and have the plasma i needed for it already

SplinterGP:
otherwise what i might do is also receive plasma on getting attacked, but in a different ratio of what both give right now

Carnage068:
> otherwise reducing the charge plasma might be enough

> otherwise what i might do is also receive plasma on getting attacked, but in a different ratio of what both give right now


Both of these things, particularly the former, would help. Like Alice said, I don't actually think these are necessarily bad changes in and of themselves. I think they just need some tweaking, so it's great to know that you're listening to feedback and stuff.

As for the Charge, yeah, right now I think it costs.. What, 500 plasma? 450? Whatever the number is, it's pretty absurd, considering that Ravage is only about 150. Eviscerating Charge should only cost 100 more plasma than that at most, like either 200 or 250. MAYBE 300, but right now it costs so much more than that. If you lowered Charge's cost, I think it would heavily alleviate the problem with PR, if not solve it entirely.

Receiving plasma on taking damage would also be nice, and it would function pretty well with Endure, and the currently being test merged Rage ability. Maybe you could make both of the changes you've mentioned.

One more thing- Have you thought about making the plasma regeneration happening instantly? As in, you slash someone, but instead of regenerating the plasma over time, it happens all at once. It would make the Fury mechanic feel more distinct from just regular plasma, and it would probably be more fitting for the playstyle you're trying to enforce.

SplinterGP:
> One more thing- Have you thought about making the plasma regeneration happening instantly? As in, you slash someone, but instead of regenerating the plasma over time, it happens all at once. It would make the Fury mechanic feel more distinct from just regular plasma, and it would probably be more fitting for the playstyle you're trying to enforce.

it does actually happen instantly but the HUD doenst update instantly for performance reasons, but i can test it out.

SplinterGP:
i want both of them to use plasma as theyre active instead of a set time in future

Carnage068:
> it does actually happen instantly but the HUD doenst update instantly for performance reasons, but i can test it out.

Huh. Could have sworn that it doesn't. I'll have to pay closer attention, I guess.

Carnage068:
WHY WAS THIS MERGED WITHOUT CHANGING ANYTHING? THIS IS A BIG NERF

2whatever2:
> WHY WAS THIS MERGED WITHOUT CHANGING ANYTHING? THIS IS A BIG NERF

I don't really understand this, considering it's passive 60% regen while fighting, ontop of that you gain fury back together with the passive gen on hit with abilities and normal slashes. Rav always had a set amount of ravages before they could use endure but this time they can actually hold onto it longer.

PR #7250, opened 2021-07-02
Allow the hivemind to reach further out of weeds.

lbnesquik:
## About The Pull Request
Hivemind can now reach further

## Why It's Good For The Game
It makes the hivemind less painful, hopefully making it less common for you to get stuck on island on weeds, especially when marines are spamming grenades. 

## Changelog
:cl:
balance: The hivemind can now reach up to three tiles rather than one.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivemind/castedatum_hivemind.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivemind/castedatum_hivemind.dm
@@ -76,6 +76,7 @@
 		/datum/action/xeno_action/activable/rally_hive/hivemind,
 	)
 
+	// *** Weeding range *** //
 	weeding_range = 1
 
 
@@ -103,6 +104,7 @@
 		/datum/action/xeno_action/activable/rally_hive/hivemind,
 	)
 
+	// *** Weeding range *** //
 	weeding_range = 2
 
 /datum/xeno_caste/hivemind/ancient
@@ -130,4 +132,6 @@
 		/datum/action/xeno_action/activable/secrete_resin/slow,
 		/datum/action/xeno_action/activable/rally_hive/hivemind,
 	)
+
+	// *** Weeding range *** //
 	weeding_range = 3
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivemind/castedatum_hivemind.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivemind/castedatum_hivemind.dm
@@ -76,6 +76,8 @@
 		/datum/action/xeno_action/activable/rally_hive/hivemind,
 	)
 
+	weeding_range = 1
+
 
 /datum/xeno_caste/hivemind/elder
 	upgrade_name = "Elder"
@@ -100,6 +102,8 @@
 		/datum/action/xeno_action/toggle_pheromones,
 		/datum/action/xeno_action/activable/rally_hive/hivemind,
 	)
+
+	weeding_range = 2
 
 /datum/xeno_caste/hivemind/ancient
 	upgrade_name = "Ancient"
@@ -126,3 +130,4 @@
 		/datum/action/xeno_action/activable/secrete_resin/slow,
 		/datum/action/xeno_action/activable/rally_hive/hivemind,
 	)
+	weeding_range = 3
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
@@ -56,7 +56,7 @@
 	. = TRUE
 	if(locate(/obj/flamer_fire) in T)
 		return FALSE
-	for(var/obj/effect/alien/weeds/W in range(1, T ? T : get_turf(src)))
+	for(var/obj/effect/alien/weeds/W in range(3, T ? T : get_turf(src)))
 		if(QDESTROYING(W))
 			continue
 		return
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
@@ -32,6 +32,7 @@
 	hud_possible = list(PLASMA_HUD, PHEROMONE_HUD, QUEEN_OVERWATCH_HUD)
 
 	var/obj/effect/alien/hivemindcore/core
+	var/weeding_range_holder = 1	//Starts at 1 then reaches 3 at the end.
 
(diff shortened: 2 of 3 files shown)


EvilMelonTheSecond:
Hiveminds spamming weeds to get in FOB is gonna be even more unbearable. Why not just jump to xenos when you are stuck?

jroinc:
There's a reason it's so low. The reason is *unkillable hiveminds phasing through walls*.

also you can clicc on weeds to jump to them at any range

lbnesquik:
> 
> 
> Hiveminds spamming weeds to get in FOB is gonna be even more unbearable. Why not just jump to xenos when you are stuck?

If they can weed the FOB, it means they got weeds literally all over behind, doesn't seem that bad to my eyes. 
Maybe i could try having it so the range extends with age?

jroinc:
> 
> 
> > Hiveminds spamming weeds to get in FOB is gonna be even more unbearable. Why not just jump to xenos when you are stuck?
> 
> If they can weed the FOB, it means they got weeds literally all over behind, doesn't seem that bad to my eyes.
> Maybe i could try having it so the range extends with age?

Because you haven't seen when they could do that. They can weed through a line of rwalls with 2-tile range.

Elli-Skala:
I've been playing quite abit of hive mind, and like jronic said, this would allow a hive mind to weed a fob before marines even land.

and being able to click on weed to jump solves most of the problem of being islanded and even being able to cross caps (though I found this after quite abit, maybe a tip of the round about hive mind being able to do this would be good?)

honestly the bigger problem I found with hivemind is that you basically have to have a second chat of just the xeno hive mind so you have links to jump to other xenos due to that being the only way to and not having follow buttons in the hive status like the queen does

lbnesquik:
I changed the feature. Now, only at the last age can you have three reach. For most of the game, it should be between one and two tiles of range. That should prevent the fob from being weeded early on. 

If the issue persist, then having a check for 'Is this place the FOB' is my next step.

Al-1ce:
> I changed the feature. Now, only at the last age can you have three reach. For most of the game, it should be between one and two tiles of range. That should prevent the fob from being weeded early on.

IIRC Hivemind takes at least an hour to reach ancient, far beyond the time marines actually land at least. This seems to be a pretty good fix - not to mention allowing weeds to grow that close to the FOB, to begin with should actually have a consequence like this.

Elli-Skala:
> > I changed the feature. Now, only at the last age can you have three reach. For most of the game, it should be between one and two tiles of range. That should prevent the fob from being weeded early on.
> 
> IIRC Hivemind takes at least an hour to reach ancient, far beyond the time marines actually land at least. This seems to be a pretty good fix - not to mention allowing weeds to grow that close to the FOB to begin with should actually have a consequence.

im liking it so far, though still abit sure as maps with long behind flanks (such as big red) still tend to always have weeds at the flanks even when xenos fully lost, meaning (atleast on big red) youll see ton of "back cap" of xeno resin

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -32,6 +32,7 @@
 	hud_possible = list(PLASMA_HUD, PHEROMONE_HUD, QUEEN_OVERWATCH_HUD)
 
 	var/obj/effect/alien/hivemindcore/core
+	var/weeding_range_holder = 1	//Starts at 1 then reaches 3 at the end.
```
BraveMole, line 35: useless

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -166,6 +166,12 @@
 	///Amount of leaders allowed
 	var/queen_leader_limit = 0
 
+	// *** Hivemind Abilities *** //
+	///Distance of weeding allowed.
+	//Starts at 1 then reaches 3 at the end.
+	var/weeding_range = 1
```
BraveMole, line 172: autodoc, see pins in dev general

code/modules/mob/living/carbon/xenomorph/castes/hivemind/hivemind.dm
```diff
@@ -56,7 +57,8 @@
 	. = TRUE
 	if(locate(/obj/flamer_fire) in T)
 		return FALSE
-	for(var/obj/effect/alien/weeds/W in range(1, T ? T : get_turf(src)))
+	weeding_range_holder = src.xeno_caste.weeding_range
+	for(var/obj/effect/alien/weeds/W in range(weeding_range_holder, T ? T : get_turf(src)))
```
BraveMole, line 61:
```suggestion
	for(var/obj/effect/alien/weeds/W in range(xeno_caste.weeding_range, T ? T : get_turf(src)))
```

code/modules/mob/living/carbon/xenomorph/castes/hivemind/castedatum_hivemind.dm
```diff
@@ -76,6 +76,8 @@
 		/datum/action/xeno_action/activable/rally_hive/hivemind,
 	)
 
+	weeding_range = 1
```
BraveMole, line 79: put that after the max_health var for example, not at random like this

BraveMole (changes requested):

Al-1ce:
> edit: but it will fix the abit of hell that is hive mind hijacking (outside again like stated above, another chat just for xenomorph comms so you have links to jump to the ship)


The hivemind cannot hijack. Very rarely it can weed into the Alamo but that is entirely the marines fault.

lbnesquik:
> 
> 
> > edit: but it will fix the abit of hell that is hive mind hijacking (outside again like stated above, another chat just for xenomorph comms so you have links to jump to the ship)
> 
> The hivemind cannot hijack. Very rarely it can weed into the Alamo but that is entirely the marines fault.

Even with three tiles of weeding, it would be pretty obvious when that attempt was happening anyway

TiviPlus:
Hivemind is too annoying with weeding in places it shouldnt be fix that for this to be considered

Elli-Skala:
> > > edit: but it will fix the abit of hell that is hive mind hijacking (outside again like stated above, another chat just for xenomorph comms so you have links to jump to the ship)
> > 
> > 
> > The hivemind cannot hijack. Very rarely it can weed into the Alamo but that is entirely the marines fault.
> 
> Even with three tiles of weeding, it would be pretty obvious when that attempt was happening anyway

abit late i know but sorry i wasnt clear

i didnt mean the hive mind trying to hijack the shuttle

i ment playing hivemind during the hijacking phase on the ship because every time your on a weedless tile (often on the unpreped weeds there) youll be kicked back to the planet and need to jump back

PR #7260, opened 2021-07-03
Defaults Auto Fit Viewport pref to off

Wayland-Smithy:
## About The Pull Request

Brings the default pref state inline with /tg/. A port of sorts if you will.

Players with the pref already enabled will not be affected by this at all, nor will devs who have their pref enabled on local when they switch to a guest since their viewport will already be "fit" from their previous session.

## Why It's Good For The Game

Defaults the pref to a state which if unwanted is much less annoying than the opposite.

## Changelog
:cl:
qol: Auto Fit Viewport now defaults to the disabled state. This makes local testing with a guest client much less CBT.
/:cl:


Wayland-Smithy:
I'm not even sure this works. Still auto fitting on local even with never changing ckey.

code/modules/client/preferences.dm
```diff
@@ -139,7 +139,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	var/see_rc_emotes = TRUE
 
 	///Should we automatically fit the viewport?
-	var/auto_fit_viewport = TRUE
+	var/auto_fit_viewport = FALSE
```
TiviPlus, line 142: No

TiviPlus (changes requested):
You're not the only one that does testing with multiple clients

code/modules/client/preferences.dm
Wayland-Smithy: It's a /tg/ port.

TiviPlus: It is not

Wayland-Smithy: I would not even try to make the change if /tg/ was not already that way. They seem to have no problem with the deafult of `FALSE` https://github.com/tgstation/tgstation/blob/master/code/modules/client/preferences.dm#L120

Wayland-Smithy: It would be super cool if you would checkout and run this locally so you can see how if your viewport is already "fit" it'll stay that way when you connect as a guest and I am betting you'd never even notice this change had this PR not triggered you.

SplinterGP:
Tivi doenst seem to like it and im not willing to advocate.

Wayland-Smithy:
Sure. It's not nearly so much of an annoyance now that the repo config defaults to widescreen.

PR #7287, opened 2021-07-06
Aim mode rework

Walarks:
## About The Pull Request
Makes shots fired while in aim mode have 30% damage decrease per marine passed, but aim mode doesn't decrease fire rate, give slowdown or have a delay for activating.

## Why It's Good For The Game
Could make it less clunky and feel better to play with compared to the previous iteration. I could re-add the slowdown part with decreased slowdown amount of it ends up being too good.

## Changelog
:cl:
balance: Makes shots fired while in aim mode have 30% damage decrease per marine passed, stacking multiplicatively, but aim mode doesn't decrease fire rate, give slowdown or have a delay for activating.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -921,9 +921,8 @@ should be alright.
 	if(CHECK_BITFIELD(flags_gun_features, GUN_IS_AIMING))
 		user.overlays -= aim_mode_visual
 		DISABLE_BITFIELD(flags_gun_features, GUN_IS_AIMING)
-		user.remove_movespeed_modifier(MOVESPEED_ID_AIM_MODE_SLOWDOWN)
 		gun_iff_signal = null
-		modify_fire_delay(-aim_fire_delay)
+		iff_marine_damage_falloff += 0.15
 		to_chat(user, "<span class='notice'>You cease aiming.</b></span>")
 		return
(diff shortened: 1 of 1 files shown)


Pariah919:
have two levels of aim mode

Walarks:
> have two levels of aim mode

Too much work to design it/too abstract to implement directly

Elli-Skala:
so unless im missing something this removes all the down sides of aim mode? doesnt this just mean instead of shooting friendlys you have less damage and that's it, making it so there is no reason not to us it?

Walarks:
> so unless im missing something this removes all the down sides of aim mode? doesnt this just mean instead of shooting friendlys you have less damage and that's it, making it so there is no reason not to us it?

the downside is that shots deal reduced damage for every marine passed, meaning it's less damage if more marines vs constant damage from current aim mode

Elli-Skala:
> > so unless im missing something this removes all the down sides of aim mode? doesnt this just mean instead of shooting friendlys you have less damage and that's it, making it so there is no reason not to us it?
> 
> the downside is that shots deal reduced damage for every marine passed, meaning it's less damage if more marines vs constant damage from current aim mode

no i mean in terms of what it will be like post this pr, after this pr will be done they appers to be no reason not to do it over non aim mode

Elli-Skala:
aim mode thing

TiviPlus:
This isnt needed

PR #7294, opened 2021-07-06
Remplace the hivemind heal with a weaker, faster & easier to employ version.

lbnesquik:
## About The Pull Request
This is basically just the shrike heal but retextured. You get more range and it applies faster than the shrike healing, but it is much weaker. (Multiplier is 2 instead of 10)

## Why It's Good For The Game
I believe this is one way to make the hivemind more of a distinct evo choice than other xenos, while also not breaking balance by denying too many kids. This won't save you if people are actively shooting at you. 

I would really be happy to see this test merged to see how players work around it.

## Changelog
:cl:
balance: rebalanced the hivemind heal into a weaker but faster to use version.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivemind/abilities_hivemind.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivemind/abilities_hivemind.dm
@@ -5,3 +5,80 @@
 
 /datum/action/xeno_action/return_to_core/action_activate()
 	SEND_SIGNAL(owner, COMSIG_XENOMORPH_CORE_RETURN)
+
+// ***************************************
+// *********** Hivemind healing
+// ***************************************
+/datum/action/xeno_action/activable/hivemind_heal
+	name = "Hivemind Heal"
+	action_icon_state = "heal_xeno"
+	mechanics_text = "Heal."
+	cooldown_timer = 1 MINUTES
+	plasma_cost = 200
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/carbon/xenomorph/castes/hivemind/abilities_hivemind.dm
```diff
@@ -5,3 +5,80 @@
...
 	SEND_SIGNAL(owner, COMSIG_XENOMORPH_CORE_RETURN)
+
+// ***************************************
+// *********** Hivemind healing
+// ***************************************
+/datum/action/xeno_action/activable/hivemind_heal
+	name = "Hivemind Heal"
+	action_icon_state = "heal_xeno"
+	mechanics_text = "Heal."
+	cooldown_timer = 1 MINUTES
```
Elli-Skala, line 16: wait this wasn't the original cool down?

lbnesquik: Yes. It is the application time which is shorter. I could have the cooldown be quicker, but that might be better separated inside another PR.

SplinterGP (changes requested):
too much duplicated code, just expand on the psychic cure

SplinterGP:
Dont see much need on this for now and too much duplicated code, ill rework hivemind if my pc gets unbricked since people seem to be willing to play it

PR #7295, opened 2021-07-07
Closet based ghost bullying fix

InterroLouis:
## About The Pull Request

Adds checks to stuffing things into closet so ghosts/hiveminds can't do it. 

## Why It's Good For The Game

No more ghost bullying. The hivemind bit is to be thorough since testing indicated they could do it. 

## Changelog
:cl:
fix: The veil has been pushed back closed. Ghosts can no longer interact with the living. 
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -262,6 +262,10 @@
 
 
 /obj/structure/closet/MouseDrop_T(atom/movable/O, mob/user)
+	if(!isliving(user))
+		return
+	if(isxenohivemind(user))
+		return
 	if(!opened)
 		return
 	if(!isturf(O.loc))


code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -262,6 +262,10 @@
 
 
 /obj/structure/closet/MouseDrop_T(atom/movable/O, mob/user)
+	if(!isliving(user))
+		return
+	if(isxenohivemind(user))
+		return
```
TiviPlus, line 268: Line 279 will always be true fix this please

TiviPlus (changes requested):

code/game/objects/structures/crates_lockers/closets.dm
InterroLouis: Do you mean the old 279 or the new 279?

InterroLouis: Assuming you mean the if(ismob(O)) check, that won't be affected by the isliving check because it's checking for O, the target movable, while the if(!isliving(user)) is checking the user instead.

TiviPlus (approved):

PR #7296, opened 2021-07-07
adds a 5 for 3 railgun ammo pack, railgun rounds reduced to 2

Pariah919:
## About The Pull Request
The Railgun, even at 40 points is basically near useless, as you can just keep ordering RR rockets rather than buying railgun rounds for nearly the same effect as the Railgun has way more punishing moments if you miss, and also doesn't fit in webbing or ammo magazine pouches. So I think this is generally fair.

5 points for 3 rounds, 1 shot is 2.

## Why It's Good For The Game
Makes the railgun more useable rather than being really bad, compared to everything else in the 40-60 point range (autosniper/amr/tx-8)

## Changelog
:cl:
balance: you can buy railgun ammo in packs of 3 for 5 points, and one shot is only two from three
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -503,7 +503,17 @@ AMMO
 /datum/supply_packs/ammo/railgun
 	name = "Railgun round"
 	contains = list(/obj/item/ammo_magazine/railgun)
-	cost = 3
+	cost = 2
+
+/datum/supply_packs/ammo/railgun_pack
+	name = "Railgun round"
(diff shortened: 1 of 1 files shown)


Al-1ce:
Railgun shoots _much_ faster than the RR and can have multiple rounds loaded. It's already viable, just not overpowered.

Pariah919:
>Railgun shoots much faster than the RR and can have multiple rounds loaded. It's already viable, just not overpowered.
it holds one bullet and has around the same travel time
does not explode on hit
requires fourty points while the RR is free (Which is fine)

PR #7297, opened 2021-07-07
warrior fling distance on grapple toss reduced to 3 from 5

Pariah919:
## About The Pull Request
warrior can fling you literally across slightly more than half a screen with a malleable grab compared to fling itself which requires much more careful coordination and grabbing for a shittier overall ability (fling itself stuns for less time I think too)
This gives you a reason to fling instead of tossing on a grapple considering warriors bread and butter is lunge anyway, so I don't see the need for it to be better than fling in about every aspect outside of rare situations where you're caught with lunge on cooldown.
Also 5 tiles is way too much when comboed which lets you literally throw marines 10 tiles away.

## Why It's Good For The Game
Warrior grapple toss throws you way too many tiles away. Three is more fair and slightly less frustrating to deal with than five.

## Changelog
:cl:
balance: warrior grapple toss distance reduced to 3 from 5.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
@@ -231,7 +231,7 @@
 /datum/action/xeno_action/activable/toss
 	name = "Grapple Toss"
 	action_icon_state = "grapple_toss"
-	mechanics_text = "Throw a creature you're grappling up to 5 tiles away."
+	mechanics_text = "Throw a creature you're grappling up to 3 tiles away."
 	ability_name = "grapple toss"
 	plasma_cost = 18
 	cooldown_timer = 20 SECONDS //Shared cooldown with Fling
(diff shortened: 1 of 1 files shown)


usnpeepoo:
salt pr.

EvilMelonTheSecond:
On one hand I really like warrior's combat utility and on the other I hate the screenwide marine displacement xeno can do. Would it be possible to have a displacement immunity time(For shrike toss/screech and warrior flings) for a few seconds after a marine gets flung instead?

TiviPlus (approved):

Surrealaser:
Absolute salt PR.

Warrior is one of the few ways Xenos can reliably dismantle over time a cohesive marine murderball.

usnpeepoo:
also, this warrior isn't CM warrior and I doubt anybodys gonna setup lunge and then use fling afterwards so what you mentioned makes 0 sense.

TheRealMedalis:
This is a skill issues xenos don't need more nerfs (changes approved)

Passe2142:
Warrior is a funny cause lunge is a bit bugged too on its diagonals iirc , fling wise , yes its 5 tiles but you have to be right next to them so its more riskier than funny bug.  Fling is situational , lunge is an starter. Warrior overall is a playmaker. 
Say if you wanted to fish a marine from a push like a surt flamer, you wont kill them on three tiles toss and the hive will get pushed. Hives cant counter OBs, and CAS for pushing mazes but they can flamers. Marines can spam tangle push too.
A Shotty/Martini user counter plays lunges too, even after a good one you lose the grab on the memerine. 
You could buff instead the resist option when lunged grab so you could break more easily. There are ways to counter warrior's lunge as well like unwired cades, flech, open spaces, staggering them.

SilencedMP5A5:
death to fishing

RuneQuester:
Salt PR

PR #7298, opened 2021-07-07
Better keybind registering system

BraveMole:
## About The Pull Request

Better system to register keybinds, more intiutive

## Changelog
:cl:
qol: Better keybind registering system
/:cl:

### DIFF
--- a/tgui/packages/tgui/components/Button.js
+++ b/tgui/packages/tgui/components/Button.js
@@ -310,60 +310,62 @@ export class ButtonKeybind extends Component {
       keysDown: {},
     };
   }
-  
+
   preventPassthrough(key) {
     key.event.preventDefault();
   }
-  
+
   doFinish() {
     const { onFinish } = this.props;
     const { keysDown } = this.state;
-  
+
     const listOfKeys
     = Object.keys(keysDown)
       .filter(isTrue => keysDown[isTrue]);
-  
+
     onFinish(listOfKeys);
     document.activeElement.blur();
     clearInterval(this.timer);
   }
-  
+
   handleKeyPress(e) {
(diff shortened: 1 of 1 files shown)


SplinterGP:
explain pls

BraveMole:
> explain pls

It waits for the first push of key, then it will register every key that was ever pressed 0.2 sec later. 

So you don't need to hold, and in practice this feels exactly like a keybind register for most games

SplinterGP:
> 
> 
> > explain pls
> 
> It waits for the first push of key, then it will register every key that was ever pressed 0.2 sec later.
> 
> So you don't need to hold, and in practice this feels exactly like a keybind register for most games

did you try also doing the thing i said about waiting till its pushed up ?

BraveMole:
> > > explain pls
> > 
> > 
> > It waits for the first push of key, then it will register every key that was ever pressed 0.2 sec later.
> > So you don't need to hold, and in practice this feels exactly like a keybind register for most games
> 
> did you try also doing the thing i said about waiting till its pushed up ?

Harder to code, i'll try if this got complaints.

This new system has barely any change, so easy to do

SplinterGP:
> 
> 
> > > > explain pls
> > > 
> > > 
> > > It waits for the first push of key, then it will register every key that was ever pressed 0.2 sec later.
> > > So you don't need to hold, and in practice this feels exactly like a keybind register for most games
> > 
> > 
> > did you try also doing the thing i said about waiting till its pushed up ?
> 
> Harder to code, i'll try if this got complaints.
> 
> This new system has barely any change, so easy to do

ye i'll get it in, just wanted to know

SplinterGP (approved):

PR #7318, opened 2021-07-09
Fixes unreloadable EORD Thermobaric

InterroLouis:
## About The Pull Request

Basically this just removes the thermobaric XMAS launcher from the weapon spawn list in the tier 5 weapon spawner landmark. It removes the 50% chance of a thermobaric that spawns in EORD being impossible to reload. 

## Why It's Good For The Game

Because this way we won't get people mhelping during EORD asking why their thermo won't reload. 

## Changelog
:cl:
fix: EORD thermobarics can be reloaded reliably.
/:cl:


TiviPlus (changes requested):
Your pr title and changelog are inadequate

InterroLouis:
Does it need a changelog if it only affects EORD weapon spawns? I'm pretty sure the tier 6 weapon spawn isn't used anywhere else at this point.

PR #7323, opened 2021-07-09
Reactivates the CLF ERT

lbnesquik:
## About The Pull Request
This simple PR brings back the CLF. It was deactivated, but i do not see any reason why we shouldn't be able to see them.

## Why It's Good For The Game
More ERT's is fun.

## Changelog
:cl:
add: Brought back the CLF ERT to life.
/:cl:

### DIFF
--- a/code/datums/emergency_calls/clf.dm
+++ b/code/datums/emergency_calls/clf.dm
@@ -2,7 +2,7 @@
 /datum/emergency_call/clf
 	name = "CLF Cell"
 	base_probability = 20
-	alignement_factor = -1
+	alignement_factor = 0
 
 
 /datum/emergency_call/clf/print_backstory(mob/living/carbon/human/H)
--- a/code/datums/emergency_calls/clf.dm
+++ b/code/datums/emergency_calls/clf.dm
@@ -1,7 +1,8 @@
 //Colonial Liberation Front
 /datum/emergency_call/clf
 	name = "CLF Cell"
-	base_probability = 0
+	base_probability = 20
+	alignement_factor = -1
 
 
 /datum/emergency_call/clf/print_backstory(mob/living/carbon/human/H)


XSlayer300:
Wasn't already redundant as the CLF fufills the same purpose as the CLF?

Not to mention, the CLF was disabled because we separated ourselves from CM long time back.

lbnesquik:
> 
> 
> Wasn't already redundant as the CLF fufills the same purpose as the CLF?
> 
> Not to mention, the CLF was disabled because we separated ourselves from CM long time back.

No clue, i didn't find a clear comment aobut it.

Also, i don't see what's really stopping us from using those ERT's, lore wise.

SplinterGP:
Arent them past SoM? Pretty sure SOM replaced them.

lbnesquik:
> 
> 
> Arent them past SoM? Pretty sure SOM replaced them.

I do not know the lore well enough to answer that. It would still be more variety regardless.

Actual worst case scenario, they get renamed in a different PR to something that fits better, If needed.

TiviPlus:
They're different allied factions last I checked

TiviPlus (approved):

code/datums/emergency_calls/clf.dm
```diff
@@ -1,7 +1,8 @@
 //Colonial Liberation Front
 /datum/emergency_call/clf
 	name = "CLF Cell"
-	base_probability = 0
+	base_probability = 20
+	alignement_factor = -1
```
TiviPlus, line 6: Actually thi should be 0 since clf hate both sides

TiviPlus (changes requested):
.

code/datums/emergency_calls/clf.dm
lbnesquik: Applied.

TiviPlus (approved):

PR #7326, opened 2021-07-10
Crit Drag Oxyloss Increase

InterroLouis:
## About The Pull Request

Increases the oxyloss from being dragged while in crit. 

Critdrag is extreme at this point in time, with crit marines being dragged excessive lengths to make up for the fact that xenos cannot drag dead bodies in the distress signal game mode. It's quite obvious that the current damage isn't nearly enough to actually limit crit dragging distance, and thus I aim to increase it. 

crit drag

If crit drag really is meant to be no more than 4 or 5 tiles, then this will make that actually true. The above image is actually on the more conservative end of crit dragging, and doesn't take into account xenos that bump push others in order to move them so much faster than the intended drag speed that they can cross long distances before the target's health properly updates and registers them as dead. 

## Why It's Good For The Game

Less excessive amounts of crit dragging. Xenos can't move dead bodies in distress anyways, and in hunt it doesn't matter because they can. 

## Changelog
:cl:
balance: Crit drag deals more oxyloss. 
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -50,7 +50,7 @@
 #define CARBON_RECOVERY_OXYLOSS -5 //the amount of oxyloss recovery per successful breath tick.
 
 #define CARBON_KO_OXYLOSS 50
-#define HUMAN_CRITDRAG_OXYLOSS 3 //the amount of oxyloss taken per tile a human is dragged by a xeno while unconscious
+#define HUMAN_CRITDRAG_OXYLOSS 10 //the amount of oxyloss taken per tile a human is dragged by a xeno while unconscious
 
 #define HEAT_DAMAGE_LEVEL_1 1 //Amount of damage applied when your body temperature just passes the 360.15k safety point
 #define HEAT_DAMAGE_LEVEL_2 2 //Amount of damage applied when your body temperature passes the 400K point


KornFlaks:
10 per tile? I occasionally play marine and I hate getting brazilied like the next marine but damn. 
That's 4 or 5 tiles before the marine dies and no more movement on Distress. That's less than a single screen and good luck breaking up a concentrated Unga Ball if marines can still see their dead comrade's body.

Please dont triple the damage in a single line PR. At most double it to 6 oxy loss per tile or let Larval Growth heal marines so some skilled xeno teamwork can isolate dead marines.

Wayland-Smithy:
If this gets in I'll try to make larva bursts perma kill again, buff the growth time overall, maybe remove the obvious "oh right I forgot I am hugged" shake and shimmy, and nerf stasis bags.

jroinc:
There needs to be a way to permakill marines.

Want it to do 10 oxyloss/tile? Fine with me. Throw in 2.5 heart damage per tile (if you're crit, worst it can do is 12.5 heart damage).

TiviPlus (changes requested):
Your justification, pr escription and title are low effort and unadequate update them please or this pr will be closed

BraveMole:
yeah, crit drag is often short distance.

The images you are showing are people that are premeding, most probably with ino or dex. If you premed you're gonna get crit drag far, but that's probably the only malus you can ever get for that. And you want to remove it?

"Less excessive amounts of crit dragging. Xenos can't move dead bodies in distress anyways". Yeah in distress xeno have basicly zero mean of permaing someone. Reminder, i made dead marines show on minimaps, so they can't even hide bodies. This mean that if marines are pushing, they will definitly save everyone that was crit dragged. 

If you want to remove excessive amount of crit drag, fix wraith, warrior and shrike cheese

Wayland-Smithy:
> The images you are showing

Also those are second hand probably, and pretty old as I saw them first posted at least a month ago now. Gonna need some better proof than some arbitrary red lines.

InterroLouis:
Literally everyone knows how bad critdrag is. Don't even pretend that you're not circumventing mechanics to prevent you from moving dead bodies in order to accomplish the same goal.

Wayland-Smithy:
> Literally everyone knows how bad critdrag is.

The fact we even have "critdrag" in our lexicon is a huge problem to me, and so is making bodies wholly unpullable, but only for certain modes, and needing to patch work in all these dead stat checks because marines can't deal with perma death in a spessman TDM server while xenos still have nothing even close to a defib for necromancing their dead back to life.

If the marine respawn timer is the issue we should be looking at that instead.

Oxyloss on drag while unconscious by a xeno was a dumb idea, I'm sorry whoever PRed that, and had I been active when that was going on I would have argued against it then and there.

I really don't understand why if pushing the silos is considered a valid route to victory xenos aren't free to haul all the dead marines there or anywhere else they want during distress rounds especially with their victims **corpses appearing on the minimap** now.

Pariah919:
i'm seeing marines get boosted across the map constantly in distress, where larva regens naturally so the hard to kill nature of marines only kind of barely matters, but other than that I think 7.5 or 8 is better than 10.

Wayland-Smithy:
> Also nothing can break a xenomorph drag other than critting them or knocking them back

I like how you forget how  new grabs that connect override old ones no matter the source.

TiviPlus:
I went ahead and took your conservative images and decided to do the actual math

The person in the image, from dragging alone would have taken 69 damage, addditionally dragging a person that long as an elderrunner (ie one of the faster castes) takes ~8 seconds which equates to additional damage of 32 totalling to 101 total damage taken in crit.
Taking into account that we assume the person was put exactly into crit and then immediately dragged with 0 delay that means they took 202% of the damage required to die in the picture.

Thus we can safely assume your proof premedded, and very succesfully so 

Thus, dont premed

David-Stormwell:
> 
> 
> I went ahead and took your conservative images and decided to do the actual math
> 
> The person in the image, from dragging alone would have taken 69 damage, addditionally dragging a person that long as an elder runner (ie one of the faster castes) takes ~8 seconds which equates to additional damage of 32 totalling to 101 total damage taken in crit.
> Taking into account that we assume the person was put exactly into crit and then immediately dragged with 0 delay that means they took 202% of the damage required to die in the picture.
> 
> Thus we can safely assume your proof premedded, and very succesfully so
> 
> Thus, dont premed

Or we could use our eyes and see he has valk on.
Either way, it isn't much of the main problem anyways, the main problem would be with displacing stuns such as shrike or shuffling.

jroinc:
> Or we could use our eyes and see he has valk on.
> Either way, it isn't much of the main problem anyways, the main problem would be with displacing stuns such as shrike or shuffling.

Muh... valk... has a DISADVANTAGE?

Shocking!

PR #7339, opened 2021-07-10
turrets have less max health and 100 less health on spawn

Pariah919:
## About The Pull Request
1500->1250
should be one less grenade to kill and 1 t-12 magazine to kill at point blank range at full health rather than one and a half.

## Why It's Good For The Game
Turrets are a bit too resistant to everything right now, in the opinion of me and several other people, so I think this is fair.

## Changelog
:cl:
balance: resin turrets have 1250 max hp from 1500, 500 on spawn rather than 600
/:cl:

### DIFF
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -249,8 +249,8 @@
 	desc = "A menacing looking construct of resin, it seems to be alive. It fires acid against intruders."
 	bound_width = 32
 	bound_height = 32
-	obj_integrity = 600
-	max_integrity = 1500
+	obj_integrity = 500
+	max_integrity = 1250
 	layer =  ABOVE_MOB_LAYER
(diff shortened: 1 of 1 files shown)


TiviPlus:
Make it allow limited amounts on the alamo/crash phase instead

SilencedMP5A5:
reopen this

Al-1ce:
keep it closed

SpaceLoveSs13:
Keep this closed. "Others agree with this opinion" is not a valid reason. Regardless, turrets can still be very easily killed, especially when marines are not there. So turrets should have good HP to protect an area until marines reach there.

SplinterGP:
> 
> 
> Keep this closed. "Others agree with this opinion" is not a valid reason. Regardless, turrets can still be very easily killed, especially when marines are not there. So turrets should have good HP to protect an area until marines reach there.

those are xeno turrets not marines

PR #7349, opened 2021-07-11
Ravager gains fury when damaged and passive regen reduced.

SplinterGP:
## About The Pull Request
Ravager now gets fury when attacked, 50% of the damage received and their passive regen is severely reduced

## Why It's Good For The Game
This makes ravager less do nothing against suppression, making them able to reciprocate while we reduce its plasma regen so they dont regen quickly around without being actively in combat.

## Changelog
:cl:
expansion: Ravager now also gets fury on receiving damage.
balance: Ravager Passive fury regen has been severely reduced.
/:cl:


Carnage068:
This is just going to fuck over Ravager even harder. Oh well.

TiviPlus (approved):

PR #7353, opened 2021-07-11
Fixes limb heal argument passing

QualityVan:
## About The Pull Request
Human's heal_limb_damage proc doesn't correctly pass its named args to the limb's heal_limb_damage proc, this fixes that.

## Why It's Good For The Game
Bug fix.
The only place that currently uses the proc in question is medical cryotubes, which as a result of this won't fix broken bones any more. Previously they did from the tube's passive healing, regardless of the installed cryo mix.

## Changelog
:cl:
fix: Cryotubes no longer fix broken bones due to incorrect argument passing
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -274,7 +274,7 @@
 	if(!parts.len)
 		return
 	var/datum/limb/picked = pick(parts)
-	if(picked.heal_limb_damage(brute, burn, updating_health))
+	if(picked.heal_limb_damage(brute, burn, updating_health = updating_health))
 		UpdateDamageIcon()
 	if(updating_health)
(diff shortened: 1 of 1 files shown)


lbnesquik:
> 
> ## About The Pull Request
> 
> Human's heal_limb_damage proc doesn't correctly pass its named args to the limb's heal_limb_damage proc, this fixes that.
> 
> Bug fix.
> The only place that currently uses the proc in question is medical cryotubes, which as a result of this won't fix broken bones any more. Previously they did from the tube's passive healing, regardless of the installed cryo mix.
> ## Changelog
> 
> 🆑
> fix: Cryotubes no longer fix broken bones due to incorrect argument passing
> /🆑

So that was a bug? It has been in for so long, i am curious to know if it should stick around in some manner. Good job on finding it tho.

PR #5216, opened 2020-11-10
Nerfs sentry turrets damage/health, buffs sunder

Pariah919:
## About The Pull Request
sentries got nerfed to 150|100 HP for big/small sentries
20|10|3 for big sentry and 15|10|2 for small sentry, gauss reduced to 25

## Why It's Good For The Game
Sentries never got deescalated, its finally time to de escalate them. They should deal more sunder than actual damage.

## Changelog
:cl:
balance: Sentry HP and damage reduced across the board, sundering increased
/:cl:

PR #7372, opened 2021-07-12
Reverts old MJP sentries change that literally killed sentries, with less sunder.

SplinterGP:
This reverts commit 23a7c4a6a8bdac6cb02547d60e96e51649b5938d.

## About The Pull Request
Just reverts #5216  with some slight changes so they dont overwhelm too much.

## Why It's Good For The Game
ever since sentries have been completely shit and doens't even do anything other than alert marines, they literally have a SMG bullet damage with a ULTRA LONG burst fire delay, making them LITERALLY DO NOTHING against xenos, as if it was not enough, THEY DIE TO 4 T12 shots from FRIENDLY FIRE.

this brings the old sentry damage back, but removes sunder so theyre not effective as a very long term deterrent, only as momentarily, by not causing sunder, which lasts longer than damage does(somewhat).

also raises their health, which keeps them from exploding to any shit, they however still knockdown to the treshold they'd explode from before, so if left unattended a sentry is just as easy to break as it was before.

This should make sentries better and not nearly useless, so marines don't need to watch a barricade with 1(one) xeno harassing it since sentries cant deal with it at all, yet they can be easily taken down if unattended by some coordination.

## Changelog
:cl:
balance: sentries now have more health so they dont die as easily, they however suffer knockdown at the old health value(which dindt happen because it died at same time)
balance: sentries now have quite more damage so they have a fair DPS, but they don't get sunder.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1019,9 +1019,9 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range = 10
 	accuracy_var_low = 3
 	accuracy_var_high = 3
-	damage = 20
+	damage = 40
 	penetration = 10
-	sundering = 3 //small damage big sunder
+	sundering = 2
 	damage_falloff = 0.5 //forgot to add this
 
 /datum/ammo/bullet/turret/dumb
@@ -1029,7 +1029,7 @@ datum/ammo/bullet/revolver/tp44
 
 /datum/ammo/bullet/turret/gauss
 	name = "heavy gauss turret slug"
-	damage = 25
+	damage = 30
 	penetration = 30
 	accurate_range = 3
 	sundering = 0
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/ammo_datums.dm
```diff
@@ -1019,27 +1019,27 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range = 10
 	accuracy_var_low = 3
 	accuracy_var_high = 3
-	damage = 20
+	damage = 40
 	penetration = 10
-	sundering = 3 //small damage big sunder
+	sundering = 2
```
SplinterGP, line 1024:
```suggestion
	sundering = 0
```

David-Stormwell:
salt pr

TiviPlus:
@SplinterGP Why is this on the tgstation repo and not your own

SplinterGP:
> @SplinterGP Why is this on the tgstation repo and not your own

I blame github :weary: i thought pressing revert would do in my repo

TiviPlus:
Turret refactor merged + conflicts I dont think this is needed anymore

Issue #7336, opened 2021-07-10
Some people's loadout data are set to var TEXT instead of var LIST/()


2whatever2:
### Description

The one who made loadouts is already aware of this, this is just meant to raise awareness for current players that can't access loadout vendor.

### Test Merges

#7134 #7164 #7257 #7194 #7099 #7270

### Reproduction Steps

1. Try to use loadout vendor
2. NTOS

### Screenshots

Example of unoperable Loadout Vendor client
![image](https://user-images.githubusercontent.com/24631139/125154375-4f7a0580-e18c-11eb-9a67-33df3cc5af62.png)
Example of working Loadout Vendor client
![image](https://user-images.githubusercontent.com/24631139/125154380-5274f600-e18c-11eb-8bf2-6bd73e91e92a.png)

PR #7381, opened 2021-07-12
Loadouts list save system change

BraveMole:
## About The Pull Request

Ditch jatum for loadouts_manager in favor of a simpler json encode.

fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/7336

## Why It's Good For The Game

Should prevent loadouts list wipe

## Changelog
:cl:
fix: Loadouts list won't wipe anymore (hopefully).
/:cl:


code/modules/client/preferences_savefile.dm
```diff
@@ -654,12 +654,14 @@
...
-	var/json_loadout_manager = ""
-	READ_FILE(S["loadouts_manager"], json_loadout_manager)
-	if(!json_loadout_manager)
-		return FALSE
-	loadout_manager = jatum_deserialize(json_loadout_manager)
-	return !isnull(loadout_manager)
+	var/loadout_version = 0
+	READ_FILE(S["loadout_version"], loadout_version)
+	if(loadout_version != CURRENT_LOADOUT_VERSION)
+		return list()
```
TiviPlus, line 660: Why dont you do a conversion from v5 to v6 here

BraveMole: yeah i'm just lazy, i'll do that

code/datums/loadout/loadout_helper.dm
```diff
@@ -174,3 +174,12 @@
 		return
 	var/datum/loadout/loadout = jatum_deserialize(loadout_json)
 	return loadout
+
+/proc/convert_loadouts_list(list/loadouts_data)
+	var/list/new_loadouts_data = list()
+	for(var/i = 1; i <= length(loadouts_data); i += 2)
+		var/next_loadout_data = list()
```
TiviPlus, line 181: for(var/i=0 to 88 step 2)

code/datums/loadout/loadout_helper.dm
```diff
@@ -174,3 +174,12 @@
...
+		next_loadout_data += loadouts_data[i]
+		next_loadout_data += loadouts_data[i+1]
```
TiviPlus, line 183: use ++

code/__DEFINES/loadout.dm
```diff
@@ -526,5 +526,7 @@ GLOBAL_LIST_INIT(loadout_job_supported, list(
 ///The maximum number of loadouts one player can have
 #define MAXIMUM_LOADOUT 50
 
-/// The currently accepted loadout version, all other loadouts will be erased from savefiles
-#define CURRENT_LOADOUT_VERSION 5
+/// The current loadout version
+#define CURRENT_LOADOUT_VERSION 6
+
+GLOBAL_LIST_INIT(accepted_loadout_versions, list(5,6))
```
TiviPlus, line 532: Why not just do it how savefiles already does?

code/datums/loadout/loadout_manager.dm
```diff
@@ -4,32 +4,25 @@
...
 ///Remove the data of a loadout from the loadouts list
 /datum/loadout_manager/proc/delete_loadout(mob/user, loadout_name, loadout_job)
-	for(var/i = 1; i <= length(loadouts_data); i += 2)
-		if(loadout_job == loadouts_data[i] && loadout_name == loadouts_data[i+1])
-			loadouts_data -= loadouts_data[i+1]
-			loadouts_data -= loadouts_data[i]
-			user.client.prefs.delete_loadout(loadout_name, loadout_job)
+	for(var/loadout_data in loadouts_data)
+		if(loadout_data[1] == loadout_job && loadout_data[2] == loadout_name)
+			loadouts_data -= loadout_data
```
TiviPlus, line 19: Does this not copy the list meaning you iterate over the same object multiple times

code/modules/client/preferences_savefile.dm
```diff
@@ -80,6 +80,12 @@
 		WRITE_FILE(S["max_chat_length"], max_chat_length)
 		WRITE_FILE(S["see_chat_non_mob"], see_chat_non_mob)
 
+	if(savefile_version == 43)
+		var/datum/loadout_manager/manager = load_loadout_manager()
+		if(istype(manager))
+			loadout_manager.loadouts_data = convert_loadouts_list(manager?.loadouts_data)
```
TiviPlus, line 86: Why is this not on loadouts through the own versioning it has

TiviPlus (changes requested):

code/modules/client/preferences_savefile.dm
BraveMole: until this PR, the loadout_manager datum itself was saved via jatum. 
Now, we avoid jatum and are saving only the the list. Hence we are no long saving the loadout manager version, because we are no longer saving/loading the manager

If the savefile is older than 44, we have to load the manager one more time and extract the list from it

code/__DEFINES/loadout.dm
BraveMole: We accept version 5, because we have a method of transforming from 5 to 6. 
This was not the case prior

code/datums/loadout/loadout_helper.dm
BraveMole: i've tried step before, it's a pain in the ass. This just doesn't work, while thise works perfectly.

And this syntax is perfectly clear, it's the standard for all major language

code/datums/loadout/loadout_manager.dm
BraveMole: I don't get what you mean.

All i can say is that this do delete one and only one loadout, and the right one

code/datums/loadout/loadout_helper.dm
BraveMole: that means step 1 or i+=1 in the loop description, which overall will be less clear IMO

code/modules/client/preferences_savefile.dm
TiviPlus: ok

code/datums/loadout/loadout_helper.dm
TiviPlus: It is but not for byond

code/datums/loadout/loadout_helper.dm
TiviPlus: No i mean the i+1 in brackets

TiviPlus (approved):

PR #7383, opened 2021-07-12
Fixes destroyed limbs also being broken

QualityVan:
## About The Pull Request
Properly clears flags when a limb is removed. This doesn't change behavior at all since nothing a broken limb can do happens if it's destroyed. Also the bleeding flag cleared instantly anyway since a destroyed limb has no wounds.

## Why It's Good For The Game
It annoys me seeing untreatable breaks on med scanners even if they don't do anything. 

## Changelog
:cl:
fix: If your arm gets torn off, it isn't broken any more.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -756,8 +756,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	if(body_part == CHEST)
 		return FALSE
 
-	remove_limb_flags()
-	add_limb_flags(LIMB_BLEEDING | LIMB_DESTROYED)
+	set_limb_flags(LIMB_DESTROYED)
 
 	for(var/i in implants)
 		var/obj/item/embedded_thing = i


TiviPlus (approved):
Tested extensively?

QualityVan:
Not particularly extensively, just a quick comparison of hitting someone until their arm falls off before and after and seeing what happens. Broken bones in a limb cause pain (overridden by destroyed), wound worsening (no wounds in a destroyed limb), and organ damage (no organs in limbs you can lose and still be alive).

code/modules/organs/limbs.dm
```diff
@@ -756,8 +756,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	if(body_part == CHEST)
 		return FALSE
 
-	remove_limb_flags()
-	add_limb_flags(LIMB_BLEEDING | LIMB_DESTROYED)
+	set_limb_flags(LIMB_DESTROYED)
```
SplinterGP, line 759: Cant this override robotic limbs?

SplinterGP (changes requested):

code/modules/organs/limbs.dm
QualityVan: It'll clear out the limb_robotic flag, yeah, but that's fine because none of the behavior specific to robotic limbs matters when it's missing (reduced damage taken, no bleeding, EMP, different drop text/repairs), the only way to fix a missing limb is via robotic replacement anyway, and I think it's basically reasonable to say a limb you don't have isn't robotic.

PR #7388, opened 2021-07-13
Make the holopad layer above catwalks.

lbnesquik:
## About The Pull Request
This changes the layer to 2.52, where catwalks are layered to 2.51

## Why It's Good For The Game
Less ugly and it makes my brain do happy chemicals

image

## Changelog
:cl:
fix: Change the holopad layer so it renders above catwalks
/:cl:

### DIFF
--- a/code/game/objects/machinery/hologram.dm
+++ b/code/game/objects/machinery/hologram.dm
@@ -5,7 +5,7 @@
 	name = "holopad"
 	desc = "It's a floor-mounted device for projecting holographic images."
 	icon_state = "holopad0"
-	layer = LOW_OBJ_LAYER
+	layer = XENO_WEEDS_LAYER
 	plane = FLOOR_PLANE
 	use_power = IDLE_POWER_USE
 	idle_power_usage = 5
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/hologram.dm
```diff
@@ -5,7 +5,7 @@
 	name = "holopad"
 	desc = "It's a floor-mounted device for projecting holographic images."
 	icon_state = "holopad0"
-	layer = LOW_OBJ_LAYER
+	layer = XENO_WEEDS_LAYER
```
TiviPlus, line 8: Are you sure you dont want to make an inbetween so weeds can never appear below a holopad?

TiviPlus (changes requested):

code/game/objects/machinery/hologram.dm
lbnesquik: I don't know. But this comes off somewhat as a suggestion, so i'll do that.

TiviPlus (approved):

PR #7391, opened 2021-07-13
Xeno Turret Minimum Distance

InterroLouis:
## About The Pull Request

Basically this just adds a minimum distance between xeno turrets, to make it so they can't place them too close to one another. It works like the silo minimum distance but with a shorter distance. In this case, the minimum distance between turrets is 6 tiles. 

## Why It's Good For The Game

Prevents the spamming of large amounts of turrets in a small area. 

## Changelog
:cl:
balance: Xeno turrets can't be built too close to each other. 
/:cl:

### DIFF
--- a/code/_globalvars/lists/objects.dm
+++ b/code/_globalvars/lists/objects.dm
@@ -41,6 +41,7 @@ GLOBAL_LIST_EMPTY(alert_consoles)			// Station alert consoles, /obj/machinery/co
 GLOBAL_LIST_EMPTY(xeno_resin_silo_turfs)
 GLOBAL_LIST_EMPTY(xeno_resin_silos)
 GLOBAL_LIST_EMPTY(xeno_turret_turfs)
+GLOBAL_LIST_EMPTY(xeno_turret_tiles)
 GLOBAL_LIST_EMPTY(xeno_weed_node_turfs)
 GLOBAL_LIST_EMPTY(xeno_resin_wall_turfs)
 GLOBAL_LIST_EMPTY(xeno_resin_door_turfs)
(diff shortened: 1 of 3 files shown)


TiviPlus:
This doesnt work because it checks spawning turfs

InterroLouis:
It adds the turret's location to that list when it initializes it so that map spawned turrets are also not able to be built close to. I have tested this, and it does what it's supposed to.

Wayland-Smithy:
Turret buff since they can't be bargain deals with fire AOE nearly as easy.

code/_globalvars/lists/objects.dm
```diff
@@ -41,6 +41,7 @@ GLOBAL_LIST_EMPTY(alert_consoles)			// Station alert consoles, /obj/machinery/co
 GLOBAL_LIST_EMPTY(xeno_resin_silo_turfs)
 GLOBAL_LIST_EMPTY(xeno_resin_silos)
 GLOBAL_LIST_EMPTY(xeno_turret_turfs)
+GLOBAL_LIST_EMPTY(xeno_turret_tiles)
```
TiviPlus, line 44: make a list of the turrets instead of the tiles

TiviPlus (changes requested):

code/_globalvars/lists/objects.dm
InterroLouis: That just happens to be what I named the list. And that's technically what it is-a list of tiles containing turrets, since it adds the location to the list. Are you saying you want me to rename the list?

TiviPlus: No I want you to change it to a list of turrets because a list like that is much more useful in the lng run and has 0 impact on the rest of your code

TiviPlus (approved):

PR #7393, opened 2021-07-13
nerfs jelly to 20 seconds from 5

Pariah919:
## About The Pull Request
Nerfs jelly to 20 seconds of cd from 5
Considering from pods you can already make jelly, and jelly in general has a 15 second duration (and is an item that never goes away) having it have an actual cooldown beyond the amount of plasma it uses makes a decent level of sense.

## Why It's Good For The Game
Jelly should have a cooldown that matters so you can't basically infinitely keep someone with fireproofing for basically ever. Which is pretty dumb in of itself. This only gives it 5 seconds of effective CD.

## Changelog
:cl:
balance: spawn jelly has 20 seconds of cooldown from 5
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivelord/abilities_hivelord.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivelord/abilities_hivelord.dm
@@ -220,7 +220,7 @@
 	action_icon_state = "gut"
 	mechanics_text = "Create a fireproof jelly."
 	plasma_cost = 100
-	cooldown_timer = 5 SECONDS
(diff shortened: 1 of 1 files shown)


SplinterGP:
this seems fine to me since jelly pods also had their cooldown reduced.

TiviPlus (approved):

lbnesquik:
It is a significant change. Maybe it should be TM'd, if that sort of stuff isn't a complete hassle.

EvilMelonTheSecond:
Why not add a decay timer for jelly on the ground similar to dead huggers? No clue how the current jelly cd is doing to be honest.

PR #7403, opened 2021-07-13
Boiler loses ammo count, glow when firing

BraveMole:
## About The Pull Request

Boiler can no longer stock any ammo, he just fires no matter what.

He will glow in the windup of the shot (2 sec)

## Why It's Good For The Game

Ammo count is just adding a tedious task to do to be able to shoot, without it doing anything balance or fun wise.
It is also a noob trap, sine new players are stocking them and glowing. Everyone that has played at least one time boiler know that you can just make one globule between shots

## Changelog
:cl:
balance: Boiler no longer need to have ammo to shoot
balance: Boiler glows when firing
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/abilities_boiler.dm
@@ -15,13 +15,13 @@
 		X.zoom_out()
 		X.visible_message("<span class='notice'>[X] stops looking off into the distance.</span>", \
 		"<span class='notice'>We stop looking off into the distance.</span>", null, 5)
-	else
-		X.visible_message("<span class='notice'>[X] starts looking off into the distance.</span>", \
-			"<span class='notice'>We start focusing your sight to look off into the distance.</span>", null, 5)
-		if(!do_after(X, 1 SECONDS, FALSE, null, BUSY_ICON_GENERIC) || X.is_zoomed)
-			return
-		X.zoom_in(11)
-		..()
+		return
+	X.visible_message("<span class='notice'>[X] starts looking off into the distance.</span>", \
+		"<span class='notice'>We start focusing your sight to look off into the distance.</span>", null, 5)
+	if(!do_after(X, 1 SECONDS, FALSE, null, BUSY_ICON_GENERIC) || X.is_zoomed)
+		return
+	X.zoom_in(11)
+	..()
 
 // ***************************************
 // *********** Gas type toggle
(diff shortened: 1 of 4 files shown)


EvilMelonTheSecond:
With the removal of sniper aim mode this is more of a QoL. Good change.

InterroLouis:
What does this mean for the plasma costs associated with bombarding?

BraveMole:
> What does this mean for the plasma costs associated with bombarding?

good point, let me fix this

SplinterGP:
I think making the ammo storage mechhanic actually a thing that does something is better IMO

BraveMole:
> I think making the ammo storage mechhanic actually a thing that does something is better IMO

i'm open to ideas, but in the same time, i don't think something interesting can be created with this mechanic

joniroq:
> > I think making the ammo storage mechhanic actually a thing that does something is better IMO
> 
> i'm open to ideas, but in the same time, i don't think something interesting can be created with this mechanic

How about if the Boiler takes damage they expel gas in a 3x3 around themselves, consuming one ammo charge and being on a cooldown (So they don't get all there ammo consumed in a single burst). Acts as a close range deterrent if the boiler gets rushed  and gives snipers a bit more of an edge since they can force a boiler to effectively lose a shot by taking pot shots at them. 

Would encourage Boilers to not stand and try to tank incoming fire to get off a bombard and incentivise the classic Defender/Boiler tag team

davidchan13:
> > I think making the ammo storage mechhanic actually a thing that does something is better IMO
> 
> i'm open to ideas, but in the same time, i don't think something interesting can be created with this mechanic

Outside the scope of the PR but I can think of a few:

Gas Spray which would function similar to acid spray but with gas clouds (Like defiler but only in a straight lines and the boiler doesn't need to run in) Could made into tiers so that it uses all the store ammo'd, for various affects: 1 ammo would create a line of weak (no vision obscure, like neuro spit) acid or neuro, 2 ammo creates a solid line, 3 ammo creates a 3x6( or what ever spray distance is) line of weak gas, 4 ammo 3x6 but solid and vision blocking, and 5-6 respectively make cones like the praespit.

Similar progression could be added to bombards with more ammo spent covering a larger area, or the boiler could even have the option to launch weak gas which would cover 50-100% more area but not obscure vision. For every one ammo spent, the bombard is 1+2x radius, with x being the ammo, so a full 5 ammo spent would have an 11 tile radius, or what ever formula playtesting determines is balanced and fair to play as and against.

While I'd love to see and exhaust or smoke screen ability that effectively is the boiler bombarding their feet to make a large stationary gas cloud, this would probably be pushing too much into the Defiler's niche.

TiviPlus:
> I think making the ammo storage mechhanic actually a thing that does something is better IMO

same

Pariah919:
Almost nothing will top glowing as boiler, really.
I'd just say give him half a tile of glow passively, and then build off there, considering boiler *doesn't* need any buffs as a caste, right now. Unless you make the ammo mechanics in mind with nerfing it in it's base state I suppose.

SplinterGP:
maybe make the gas of the boiler shot scale with ammo size, so they need to charge before hand to shoot big amounts of gas, and consumes all of it.

vilereaver41:
yay,

PR #7412, opened 2021-07-13
Critdrag

jroinc:
## About The Pull Request

Critdrag now kills marines faster, but death is more punishing in general.

Being dragged by a xeno deals 2.5 oxyloss and 2.5 brute damage (up from 3.0/0). Dying in any way gives 20 cloneloss.

## Why It's Good For The Game

Marines won't get dragged offscreen as much. and will be easier to retrieve.
Xenos won't have marines keep popping up like demented jack-in-the-boxes if there's a medic around. Kill someone enough times and he's forced to retreat.

## Changelog
:cl:
add: 20 cloneloss on marine death
balance: critdrag damage upped from 3 to 5
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -50,7 +50,8 @@
 #define CARBON_RECOVERY_OXYLOSS -5 //the amount of oxyloss recovery per successful breath tick.
 
 #define CARBON_KO_OXYLOSS 50
-#define HUMAN_CRITDRAG_OXYLOSS 3 //the amount of oxyloss taken per tile a human is dragged by a xeno while unconscious
+#define HUMAN_CRITDRAG_DAMAGE 5 //the base amount of damage taken per tile a human is dragged by a xeno while unconscious
+#define HUMAN_DEATH_PENALTY 20 //the amount of cloneloss taken per death of a human
 
 #define HEAT_DAMAGE_LEVEL_1 1 //Amount of damage applied when your body temperature just passes the 360.15k safety point
 #define HEAT_DAMAGE_LEVEL_2 2 //Amount of damage applied when your body temperature passes the 400K point
--- a/code/modules/mob/living/carbon/human/death.dm
+++ b/code/modules/mob/living/carbon/human/death.dm
@@ -54,6 +54,7 @@
 /mob/living/carbon/human/on_death()
 	if(pulledby)
 		pulledby.stop_pulling()
+	adjustCloneLoss(HUMAN_DEATH_PENALTY)
 
 	//Handle species-specific deaths.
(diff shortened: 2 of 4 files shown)


lKiseki:
20 cloneloss is unnecessary. It is already easy enough to remove marines from any given round with how much displacement xenos possess across all castes, unless you do some mistakes, like killing them too close to FOB or slashing them too much instead of critdragging.

jroinc:
> 
> 
> 20 cloneloss is unnecessary. It is already easy enough to remove marines from any given round with how much displacement xenos possess across all castes, unless you do some mistakes, like killing them too close to FOB or slashing them too much instead of critdragging.

You may be right. But I think some sort of permanent damage/death is needed. I could be wrong, so I'm watching the revive stats for every round. If they're consistently above 50%, we need it.

InterroLouis:
Where on the body does the brute damage target?

jroinc:
> 
> 
> Where on the body does the brute damage target?

Should be randomly distributed. Could snap a bone if you're already heavily damaged in one spot, but RNG.

KornFlaks:
20 cloneloss seems like quite a lot. 10 seems more reasonable.

jroinc:
> 
> 
> 20 cloneloss seems like quite a lot. 10 seems more reasonable.

I considered 10, but it's too low to be noticed. You die three times, you're only at 30 cloneloss, which is ignorable.
Ideally, we'd do this with a system where you got more each death (e.g. 0 first death, 20 second death, 40 third and so on), but I'm only good enough at coding to do it this way.

TheRealMedalis:
There is no reason to make this change. If the players want death to be more punishing they can vote hunt.

jroinc:
> 
> 
> There is no reason to make this change. If the players want death to be more punishing they can vote hunt.

I think there is~
And lucky for me, I coded it!

jroinc:
Converted to draft while waiting for more data.

Losenis:
>Get 20 cloneloss because CAS or OB decided to get some good hits, or someone gave PVT Unga McDunga a SADAR

Thanks

jroinc:
> 
> 
> > Get 20 cloneloss because CAS or OB decided to get some good hits, or someone gave PVT Unga McDunga a SADAR
> 
> Thanks.
> 
> Might as well move this over to causing it on psydrain/finished cocoon if you're adamant on seeing this through.

That's a really good idea. I'mma change it to that instead. Psydrain/*started* coocoon, though.

jroinc:
Upon review of medical stats, I don't think this is needed.

Losenis:
> Upon review of medical stats, I don't think this is needed.

I beg to differ
image

PR #7415, opened 2021-07-14
Lets runners and larva walk through acid holes

QualityVan:
## About The Pull Request
Enables moving through the holes in walls vomited acid leaves for entities with the PASSTABLE flag. That's larva and runners plus thrown objects (grenades could already be fake-thrown through, but it's more freeform now) and humans using jetpacks. The last one is kind of weird so maybe worth adding a new flag, but hey. Projectiles can't pass, and you can only pass through in the hole's direction, east/west or north/south.

## Why It's Good For The Game
FSAT
Also one caste enabling another's mobility is good.

## Changelog
:cl:
add: Runners and larvae can pass through acid holes without pausing.
/:cl:

### DIFF
--- a/code/game/turfs/walls/walls.dm
+++ b/code/game/turfs/walls/walls.dm
@@ -510,7 +510,7 @@
 		return FALSE
 	var/move_dir = mover.dir
 	var/hole_facing = acided_hole.dir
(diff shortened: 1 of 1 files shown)


Al-1ce:
This seems interesting. I can imagine some useful synergies developing eventually and marines actually blocking off holes.

BraveMole:
yeah jetpack is definitly wtf

QualityVan:
not feelin it

PR #7432, opened 2021-07-14
Hard del tracking, disabling and handling changes

BraveMole:
This port some improvement on how to handle hard del from tg

https://github.com/tgstation/tgstation/pull/56724 : this was already ported in part, i'm adding the rest here
https://github.com/tgstation/tgstation/pull/57728 : this is better ref tracking. I'm not adding the unit test from this
https://github.com/tgstation/tgstation/pull/58972 : fixed some bugs of the hard del traker
https://github.com/tgstation/tgstation/pull/59371 : this optimized the ref tracker

This needs config update and probably LAA
https://github.com/tgstation/tgstation/pull/59750 : if the same type hard del too much, hard dels on that specific type will be disabled.

### DIFF
--- a/code/datums/datum.dm
+++ b/code/datums/datum.dm
@@ -45,7 +45,12 @@
 
 	/// A weak reference to another datum
 	var/datum/weakref/weak_reference
-
+    /**
+     * Lazy associative list of currently active cooldowns.
+     *
+     * cooldowns [ COOLDOWN_INDEX ] = add_timer()
+     * add_timer() returns the truthy value of -1 when not stoppable, and else a truthy numeric index
+     */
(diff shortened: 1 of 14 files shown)


code/datums/datum.dm
```diff
@@ -46,13 +46,35 @@
 	/// A weak reference to another datum
 	var/datum/weakref/weak_reference
 
-	///Lazy associative list of currently active cooldowns.
+	/*
+	* Lazy associative list of currently active cooldowns.
+	*
+	* cooldowns [ COOLDOWN_INDEX ] = add_timer()
```
TiviPlus, line 52: incorrect autodoc

PR #7442, opened 2021-07-15
Adds the HMG to marine vendors, HSG rebalance.

Pariah919:
## About The Pull Request
HSG fires slower, (4 rounds a second from 5), 0.25 fire delay. Not the nested one though, it can stay as is for now.
HMG is added to marine vendors, marines only get one per round for now, they can buy more ammo, but not another one for now, so don't lose it?

-HMG is slower to deploy, has no IFF, has much more ammo (500 rounds) is harder overall to use on the move, as carrying extra ammo requires your back or belt slot.
-HSG fires slower than the HMG, but has IFF. Other than that, it's easier to carry around, as it fits in engi bags, the ammo also fits in a lot more places than the HMG.

## Why It's Good For The Game
Expands the roster of mountable machineguns, and balances the TL-102 a bit, as it's a bit too good right now. 
The MG08/495 is incredibly effective at holding a chokepoint, but has no IFF at all, so it requires more tactical thinking. I think it'll improve marine tactics and add more variety to rounds.

## Changelog
:cl:
expansion: Marines get one HMG, it is harder to lug around than a TL-102, has no IFF, but it fires faster, and deploys slower. Better not lose it.
balance: TL-102 HSG fires slower (0.25 fire delay from 0.2), added one cool HMG gun to the marine vendor. Better not lose it.
spriteadd: sprites for the MG08/495 and TL-102 while in your hand.
fix: tl-102 and mg08 can no longer take any attachments.
qol: minor reorganization of specialized tab in marine vendors
/:cl:


LoganKele:
When is this gonna be merged and what vendor is it going to be in?

David-Stormwell (approved):
In theory, this sounds swell enough, of course though, It requires field testing, and once applied and tested and adjusted if required this should be a swell enough addition.
I'll leave the finer parts of the code for maintainer dissection.

code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -162,6 +163,7 @@
 			/obj/item/ammo_magazine/flamer_tank/large = 20,
 			/obj/item/ammo_magazine/flamer_tank = 20,
 			/obj/item/weapon/shield/riot/marine = 6,
+			/obj/item/ammo_magazine/heavymachinegun = 4,
```
SplinterGP, line 166: put ammo below gun, forgot that on pepperball you can change this too

SplinterGP (changes requested):

code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -154,15 +154,17 @@
 		),
 		"Specialized" = list(
 			/obj/item/weapon/gun/launcher/m92/standardmarine = -1,
+			/obj/item/weapon/gun/heavymachinegun = 1,
+			/obj/item/ammo_magazine/heavymachinegun = 4,
 			/obj/item/weapon/gun/launcher/m81 = -1,
```
SplinterGP, line 159:
```suggestion
			/obj/item/weapon/gun/launcher/m81 = -1,
			/obj/item/weapon/gun/heavymachinegun = 1,
			/obj/item/ammo_magazine/heavymachinegun = 4,
```

SplinterGP (changes requested):

PR #7443, opened 2021-07-15
Add an mundane special force ERT to the mix

lbnesquik:
## About The Pull Request
- They are slightly more likely to spawn than PMC, but less than freelancers.
- Added some skillsets for them to use, so they can operate C4
- Modified the PMC MP5 to dark for this

They are meant to be roughly aligned with marines, with their bond being the colony, not salary. At least, not completely.

image

## Why It's Good For The Game
More variation of ERT's, so we don't cycle through the same four. Now, we can cycle through the same five!

## Changelog
:cl:
add: Added a tacticool ERT for you to enjoy.
/:cl:


code/datums/skills.dm
```diff
@@ -436,6 +442,10 @@ engineer, construction, leadership, medical, surgery, pilot, police, powerloader
 	shotguns = SKILL_SHOTGUNS_TRAINED
 	heavy_weapons = SKILL_HEAVY_WEAPONS_TRAINED
 
+/datum/skills/SL/pmc/special_forces
```
TiviPlus, line 445: No uppercase in typepaths

TiviPlus (changes requested):

GradusNL:
Your fluff doesn't really make sense:
_You are part of the local Special Responses unit, based within this isolated system, having received good training and equipment.
Usually, you spend most of your time providing security to the few big shots around here, in between RnR in your base and the occasional hostage situations. However, recently, one of the local settlement recently went dark, broadcasting a distress signal, just as a TGMC vessel, [SSmapping.configs[SHIP_MAP].map_name] showed up in orbit, and the two are probably linked. Investigate their presence, and do your best to help the colony below!_
So, this is basically a SWAT team responding to the xeno outbreak, instead finding the hijack ship and boarding that? How does that make sense? Whose SWAT team is this? The Colonial Marshall's? Why aren't they working together with the TGMC in the first place? Why are they investigating colonies going dark?
If you want to make a special forces unit to respond to distress calls, why not make it part of the TGMC? Make it a Marine Raider squad responding to the distress call.

I'm also not seeing the point of an ERT that's a mix of Freelancer and PMC, but that's just me. I'd rather have a fresh ERT offering something new.

lbnesquik:
> I'm also not seeing the point of an ERT that's a mix of Freelancer and PMC, but that's just me. I'd rather have a fresh ERT offering something new.

More variety. That's literally the point. If you feel like making another, feel free. The lore may be a little jank, but i wanted some SAS inspired ERT.

code/datums/skills.dm
lbnesquik: Applied the changes and some more

TiviPlus (approved):

PR #7454, opened 2021-07-16
Runner has higher melee armor

Pariah919:
## About The Pull Request
Runner has an extreme issue dealing with melee bump enemies such as vali, and I think personally that it should have higher armor protection against these enemies, as it can't evade them at all.

## Why It's Good For The Game
Runner can't really do anything against a vali shield so it should get more protection against a vali shield.

## Changelog
:cl:
balance: runner has higher melee armor now.
/:cl:


EvilMelonTheSecond:
Why would we need runners to fight vali shield users though? A T1 being weak against a build with severe mobility and range limitations seems fair to me.

David-Stormwell:
> 
> 
> Why would we need runners to fight vali shield users though? A T1 being weak against a build with severe mobility and range limitations seems fair to me.

Runners can already be deleted by almost anything else as well, and having a sword with bump attacks, as another counter to pile upon the already large list of things to be scared of as a novice Xeno player is quite troublesome.

If you want to delete a runner, use a shotgun, using a sword with bump attacks to bypass the evasion as well as a shield to block pounces makes it harder for players to catch a break.

SplinterGP:
maybe hunter as well ?

SplinterGP (approved):

TiviPlus (approved):

PR #7459, opened 2021-07-16
Allow the AI to interact with sentries.

lbnesquik:
## About The Pull Request
- This also allow ghosts to check things out.

By the way, how the heck does manual override works? 

## Changelog
:cl:
expansion: Allow the AI to interact with turrets, and ghosts to check them out as well. No they can't use them.
/:cl:

### DIFF
--- a/code/game/objects/machinery/sentries.dm
+++ b/code/game/objects/machinery/sentries.dm
@@ -227,6 +227,7 @@
 	use_power = 0
 	req_one_access = list(ACCESS_MARINE_ENGINEERING, ACCESS_MARINE_ENGPREP, ACCESS_MARINE_LEADER)
 	hud_possible = list(MACHINE_HEALTH_HUD, SENTRY_AMMO_HUD)
+	interaction_flags = INTERACT_MACHINE_TGUI
 	var/turret_flags = TURRET_HAS_CAMERA|TURRET_SAFETY|TURRET_ALERTS
 	/// The iff bitfield used to determine friendlies from hostiles
 	var/iff_signal = NONE
@@ -410,7 +411,7 @@
 		return
 
 	var/mob/living/carbon/human/user = usr
-	if(!istype(user))
+	if(!istype(user) && !issilicon(user))
 		return
 
 	switch(action)


code/game/objects/machinery/sentries.dm
```diff
@@ -403,14 +403,16 @@
 		"burst_size" = burst_size,
 		"mini" = istype(src, /obj/machinery/marine_turret/mini)
 	)
+/obj/machinery/marine_turret/interact(mob/user)
+	ui_interact(user)
```
TiviPlus, line 407: there is a flag for this

TiviPlus (changes requested):
.

code/game/objects/machinery/sentries.dm
lbnesquik: I am not familiar with this flag

lbnesquik: Applied the changes

code/game/objects/machinery/sentries.dm
```diff
@@ -410,7 +411,7 @@
 		return
 
 	var/mob/living/carbon/human/user = usr
-	if(!istype(user))
+	if(!istype(user) && !issilicon(user))
```
BraveMole, line 414: ew, no.

First you check what type is usr, then you cast.

And do you even need to cast anyway? when are you using user as a human or even as a mob?

BraveMole (changes requested):

code/game/objects/machinery/sentries.dm
lbnesquik: This stuff will get overwritten by the turret rework in like, two weeks, i don't think it needs to be as sturdy as regular code.

lbnesquik:
Welllll, given the sentry rework is being test merged, i'll probably just wait until that is done

SplinterGP:
> 
> 
> Welllll, given the sentry rework is being test merged, i'll probably just wait until that is done

do it, the sentries on the sentry rework uses a completely different typepath and this doenst affect them.

PR #7471, opened 2021-07-17
king takes a hundred minutes to possibly get into the game rather than 60

Pariah919:
## About The Pull Request
considering the game usually starts at 12:30 you can literally get king at around 13:05 which is literally 35 minutes into the game
for a stalemate breaker it should take pretty long to get into the game, now it'll take like 60 minutes or so.

## Why It's Good For The Game
king shouldn't be appearing in early mid game at all, will drive for more interesting rounds.

## Changelog
:cl:
balance: king takes 100 minutes to be able to be summoned, rather than 60.
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -135,7 +135,7 @@
 //How many psy points are gave every 5 second by a cocoon
 #define COCOON_PSY_POINTS_REWARD 2
 
-#define INVOKE_KING_TIME_LOCK 1 HOURS
+#define INVOKE_KING_TIME_LOCK 100 MINUTES
 
 /// How each alive marine contributes to burrower larva output per minute. So with one pool, 15 marines are giving 0.375 points per minute, so it's a new xeno every 22 minutes
 #define SILO_BASE_OUTPUT_PER_MARINE 0.03


Wayland-Smithy:
> rather than 60

After you add the lobby time and the 10 minute pod timer its more like 73

KornFlaks:
So you're saying kings can only be summoned by 13:50? Practically 14:00?

Draw votes happen at 14:00 because both sides are pretty much tired as fuck. 
A stalemate by then has been drawn on for over a hour and multiple kings are needed to even remotely break the siege.

No way should the king be delayed until then. This is ridiculous.

lbnesquik:
Not in favor of this. King isn't even that big of a deal if you have a sadar or aggressive marines.

TiviPlus:
unneeded

David-Stormwell:
@TiviPlus Here's a laydown of how a round goes, 12:30~ shutters open, Marines rush out and in these cases will most likely push, get slowly picked off, or if the Xenos were extra competent, slaughter the forward group of Marines, thus causing the rest of them to fall back to close to the FOB, securing the closest miners and hunkering down, in turn, picking Xenos off.
This should have happened by around, 13:00-13:30ish~.
Now at the FOB siege, Marines have a chance to make or break Xenos via picking the stupid ones off and or getting equipment to kill any of the majorly skilled Xenos who forwarded the Xeno's groups. However, due to the fact that the first group of Marines got slaughtered, they can no longer push due to the fact that Turrets encamp around them preventing a good fast blitz, and as well as the fact as a King most likely by :30 max incoming to kill them all.

**Now do note**, no one likes a FOB siege, but every enemy player will majorly dislike a bullet sponge that can't die if properly supported and now is constantly damaging their turrets and cades, thus causing an Evac without any chance of Marines being able to attack back.
Now, I do think 100 minutes may be over-reacting, but a slight increase in minutes should be considered.

Pariah919:
i think it'd be more fair if i reduce it to ninety
i'd atleast like ninety tested

PR #7479, opened 2021-07-17
Fix living panic bunker rejecting too agressively

TiviPlus:
By default will reject while panic bunker is on and ppl only observed but if you set required playtime to 1 and you have a playtime of 1 you will be rejected anyway


chefferz:
This broke the panic bunker and made it non-functional in my test. Anyone was able to join.

TiviPlus:
https://github.com/tgstation/tgstation/pull/60289/files

Issue #7467, opened 2021-07-16
Loadout tool snaps earpieces and ID's


2whatever2:
### Description

Be admin, make loadout, try to mass give, realize they have no aimmode since their id's are gone, and no headsets 

### Test Merges

#7432 #7267 #7259 #7151 #7372 #7181 #7383 #7452

### Reproduction Steps

1. Make loadout
2. Slap it on a guy
3. ID dissapears, same with headset

### Screenshots

Before: 
![image](https://user-images.githubusercontent.com/24631139/125998445-f54b6858-53bf-4ead-8df7-094f84604b82.png)
After: (How do I still have minimap, isn't that a headset thing)
![image](https://user-images.githubusercontent.com/24631139/125998420-2078c225-1445-4a6f-87a2-ea5b3debe9ac.png)

BraveMole:
It saves the id and the headset as well, so it's erasing everything on those slots

PR #7480, opened 2021-07-17
Loadout tools will now save job title, id's access and iff signal, and headset

BraveMole:
## About The Pull Request

Iff signal and access can be customised on the id before saving
Job (so the title, the rank, the skillset) and headset are hardcoded (if the path doesn't exist in the code, it won't save anything). So you can't make custom jobs

fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/7467

## Changelog
:cl:
expansion: Loadout tools will now save job title, id's access and iff signal, and headset
/:cl:

### DIFF
--- a/code/datums/loadout/item_representation/_item_representation.dm
+++ b/code/datums/loadout/item_representation/_item_representation.dm
@@ -118,3 +118,32 @@
 	stack.update_weight()
 	stack.update_icon()
 	return stack
+
+/**
+ * Allow to representate an id card (/obj/item/card/id)
+ */
+/datum/item_representation/id
+	/// the access of the id
+	var/list/access = list()
+	/// the iff signal registered on the id
+	var/iff_signal = NONE
+
+/datum/item_representation/id/New(obj/item/item_to_copy)
+	if(!item_to_copy)
+		return
+	if(!isidcard(item_to_copy))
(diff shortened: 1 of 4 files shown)


code/datums/loadout/item_representation/_item_representation.dm
```diff
@@ -118,3 +118,32 @@
...
 	stack.update_icon()
 	return stack
+
+/**
+ * Allow to representate an id card (/obj/item/card/id)
+ */
+/datum/item_representation/id
+	/// the access of the id
+	var/list/access = list()
+	/// the iff signal registered on the id
```
TiviPlus, line 128: Will this not create free clonable id cards for marines

BraveMole: this is only for admins (through loadout tool). The id slot is not saved by the loadout vendors

BraveMole: And even if you put an id in your bag, you can't find any in vendors. So the loadout vendor would not sell it back to you

TiviPlus: ok

TiviPlus (approved):

PR #7482, opened 2021-07-17
SO rank changes

Hekzder:
## About The Pull Request

SO now starts at O-3, goes to O-4 at 25 hours, and O-5 at 50 hours. Basically just adds another rank to make them a three rank role instead of their awkward and un-needed position as one of the only two rank roles. Also it's a four slot role so starting them off at O-3 makes for less O-4+ all running around the ship.

## Why It's Good For The Game

Better distinction between novice SOs and very seasoned SOs, also more satisfying rank progression for those who do play SO more.

## Changelog
:cl:
add: Added O-3 to SO as starting rank. SO now gets O-4 at 25 hours and O-5 at 50 hours. 
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -188,7 +188,7 @@ Make the TGMC proud!"})
 //Staff Officer
 /datum/job/terragov/command/staffofficer
 	title = STAFF_OFFICER
-	paygrade = "O4"
+	paygrade = "O3"
 	comm_title = "SO"
 	total_positions = 4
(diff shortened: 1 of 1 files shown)


Hekzder:
I honestly have no idea why this is failing tests. It seems completely unrelated and it's the exact same change that I did on my previous PR that passed tests fine.

davidchan13:
Could honestly have them all start as O-1 Ensigns and it'd still make sense from a hierarchy standpoint.

Hekzder:
> Could honestly have them all start as O-1 Ensigns and it'd still make sense from a hierarchy standpoint.

Honestly yes but I didn't want to change them incredibly dramatically. We _could_ make them O-1 through O-4 and it would still make sense, for sure. I'd be down for it.

TiviPlus (approved):

PR #7494, opened 2021-07-18
Turrets don't detect stealthed hunters

QualityVan:
## About The Pull Request
Adds signal call/response for turrets to pass over targets that've hidden themselves and gives it to hunters.

## Why It's Good For The Game
Stealth should hide you from things.

## Changelog
:cl:
add: Hunters in stealth won't get spotted by turrets.
/:cl:


KornFlaks:
Also give this to the Hivemind's eye. Having turrets shoot at thin air because the hivemind is looking there doesnt make sense. Unless that was fixed already.

Havent seen a hivemind in game in a long time.

QualityVan:
Hivemind doesn't provoke turrets already, which I probably should've checked initially before making a quick fix.

BraveMole:
I know that hunter code stealth code is something from outer world, but you don't need to send a signal to the hunter.

Just add a flag or a trait on the hunter when stealth, and check if that flag/trait is on the xeno when firing

lbnesquik:
Please do not let them slash turrets while cloaked! That would be SO annoying.

QualityVan:
Traits is better, closing to open a separate one since that's a completely different system.

PR #7496, opened 2021-07-18
warding doenst make you recover/die less on crit.

SplinterGP:
## About The Pull Request
Xenos don't take less damage/heal better on crit with warding, back then it was way worse and i nerfed it half mostly so it woulndt be that huge of a nerf, still i think it does a way too good job and makes recovering from crit and not dying to it veery veery better, which led me to remove it.

## Why It's Good For The Game
Warding in my view makes xenos recover way too much and die way too slow off weeds when on crit, this removes this interaction totally, i don't think thats that much of a nerf but its considerable, but warding already is very good by giving armor don't think it doing this more is necessary

## Changelog
:cl:
balance: Xenos don't take less damage/heal better on crit with warding anymore.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -98,9 +98,9 @@
 /mob/living/carbon/xenomorph/proc/handle_critical_health_updates()
 	var/turf/T = loc
 	if((istype(T) && locate(/obj/effect/alien/weeds) in T))
-		heal_wounds(XENO_RESTING_HEAL + (warding_aura * 0.5) * 0.5) //Warding pheromones provides 0.125 HP per second per step, up to 1.25 HP per tick.
+		heal_wounds(XENO_RESTING_HEAL * 0.5)
 	else if(!endure) //If we're not Enduring we bleed out
-		adjustBruteLoss(XENO_CRIT_DAMAGE - (warding_aura * 0.5)) //Warding can heavily lower the impact of bleedout. Halved at 5.
+		adjustBruteLoss(XENO_CRIT_DAMAGE)
 
 /mob/living/carbon/xenomorph/proc/heal_wounds(multiplier = XENO_RESTING_HEAL, scaling = FALSE)
 	var/amount = 1 + (maxHealth * 0.0375) // 1 damage + 3.75% max health, with scaling power.
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -98,7 +98,7 @@
 /mob/living/carbon/xenomorph/proc/handle_critical_health_updates()
 	var/turf/T = loc
 	if((istype(T) && locate(/obj/effect/alien/weeds) in T))
-		heal_wounds(XENO_RESTING_HEAL * 0.5)
+		heal_wounds(XENO_RESTING_HEAL)
 	else if(!endure) //If we're not Enduring we bleed out
 		adjustBruteLoss(XENO_CRIT_DAMAGE)
 


lKiseki:
Y e s.

David-Stormwell:
Quite needed.

SilencedMP5A5:
had a ravager on warding and recovery survive getting shot 8 times with TP23, going in and out of crit repeatedly, that was funny.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -98,9 +98,9 @@
 /mob/living/carbon/xenomorph/proc/handle_critical_health_updates()
 	var/turf/T = loc
 	if((istype(T) && locate(/obj/effect/alien/weeds) in T))
-		heal_wounds(XENO_RESTING_HEAL + (warding_aura * 0.5) * 0.5) //Warding pheromones provides 0.125 HP per second per step, up to 1.25 HP per tick.
+		heal_wounds(XENO_RESTING_HEAL * 0.5)
```
QualityVan, line 101: Intentional that you're also halving the base heal value on top of not applying a warding bonus?

BraveMole: splinter just doesn't know how to do math lel

Surrealaser:
Warding is already probably the worst phero, and we want to make it even less relevant?

code/modules/mob/living/carbon/xenomorph/life.dm
SplinterGP: > 
> 
> Intentional that you're also halving the base heal value on top of not applying a warding bonus?

it halved the base heal + warding aura before

SplinterGP:
> 
> 
> Warding is already probably the worst phero, and we want to make it even less relevant? While undue resilience in crit is bad, Warding definitely should have something else if it loses all the crit benefits.

its not the worst phero at all, recovery is extremely useless in combat since youre constantly getting hit so you dont heal at all, warding however can give up to if i remember 12.5 extra armor which is considerably alot.

code/modules/mob/living/carbon/xenomorph/life.dm
QualityVan: Nah, before it had the base resting heal plus a quarter of the warding aura. B + (W * .5) * .5

Surrealaser:
> > Warding is already probably the worst phero, and we want to make it even less relevant? While undue resilience in crit is bad, Warding definitely should have something else if it loses all the crit benefits.
> 
> its not the worst phero at all, recovery is extremely useless in combat since youre constantly getting hit so you dont heal at all, warding however can give up to if i remember 12.5 extra armor which is considerably alot.

It's the worst phero.

Recovery is pretty much obligatory; without access to a source of it, xenos are fucked.

Frenzy is invaluable for both offense and defense, and can and often does directly mean the difference between life and death for both xenos and marines.

SplinterGP:
Recovery still doenst do anything mid combat.

code/modules/mob/living/carbon/xenomorph/life.dm
SplinterGP, line 101:
```suggestion
		heal_wounds(XENO_RESTING_HEAL)
```

Surrealaser:
> Recovery still doenst do anything mid combat, and alternating between both in and out of combat is a easy thing to do.

Who cares?

It's not about whether it does anything combat, it's about the totality of its value.

Warding's totality of value and impact is the worst easily.

Chip11-n:
It's already almost useless and you want to make it a shit even more? Good decision

TiviPlus (approved):

Carnage068:
thx, u did it. u saved tgmc.,,,,

PR #7498, opened 2021-07-18
Removes Transvitox Having Fractional Gas Reagent Transfer Strength for Some Reason

Surrealaser:
## About The Pull Request

Transvitox no longer has fractional gas reagent transfer strength for some reason.

## Why It's Good For The Game

Transvitox's effect does not justify an arbitrary nerf to its reagent transfer strength as probably the weakest of the xeno toxins, while the others do not have compromised reagent transfer strength.

## Changelog
:cl:
fix: Transvitox no longer has fractional gas reagent transfer strength for some reason.
balance: Transvitox no longer has fractional gas reagent transfer strength for some reason.
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -354,7 +354,6 @@ datum/effect_system/smoke_spread/tactical
 
 /datum/effect_system/smoke_spread/xeno/transvitox
 	smoke_type = /obj/effect/particle_effect/smoke/xeno/transvitox
-	strength = 0.75
 
 /////////////////////////////////////////////
 // Chem smoke


SplinterGP (changes requested):
better reasonings on why its good for game, its just a copypasta of PR description

Pariah919:
>fix: Transvitox no longer has fractional gas reagent transfer strength for some reason.
balance: Transvitox no longer has fractional gas reagent transfer strength for some reason.
>
this literally tells nobody anything about what you've changed or fixed at all

Surrealaser:
> better reasonings on why its good for game, its just a copypasta of PR description

Okay.

Pariah919:
>else if(X.selected_reagent == /datum/reagent/toxin/xeno_transvitox)
        N.smoke_type = /obj/effect/particle_effect/smoke/xeno/transvitox
        N.strength = 0.75
        smoke_range = 4
>
was always like this, so this is actually a balance change, not a fix.

Surrealaser:
> > else if(X.selected_reagent == /datum/reagent/toxin/xeno_transvitox)
> > N.smoke_type = /obj/effect/particle_effect/smoke/xeno/transvitox
> > N.strength = 0.75
> > smoke_range = 4
> 
> transvitox strength was always like this, so the guy made it have 0.75 strength. This is actually a balance change, not a fix.
> 
> EDIT: On another note Transvitox has a higher smoke range than other gasses which is probably why it has lower strength to begin with, but I don't really see the need to buff it to full strength at all.

Because Transvitox kinda sucks?

PR #7506, opened 2021-07-18
Facehugger catching

### LINKED ISSUE #7499: Catching huggers dupes them
### Description

Per title. If you try to catch a Carrier's hugger by having throw mode active, assuming you succeed, you'll get a duplicate living hugger in your hand and another living hugger at your feet.

### Test Merges

#7442, #7372, #7432

### Reproduction Steps

1. Approach a Carrier.
2. Activate Throw mode, in preparation to catch a hugger.
3. Have the Carrier throw a hugger at you.
4. Catch the hugger. Notice you'll have one in your hand and another at your feet.

### Screenshots

_No response_

QualityVan:
## About The Pull Request
Fixes #7499 which was an issue with an object being in two places at once, rather than duplication.
You can catch thrown facehuggers(even leaping ones) with an open hand and throw intent active, but they'll keep trying to attack you. Use in hand to kill it, you monster.
Fixes storages breaking if you put a facehugger inside, the hugger will jump out when you open it now.
Repaths non-larval huggers to condense their similarities and make them not fit on your face.

## Why It's Good For The Game
Fixes things mentioned above, a bit of code improvement, improves some ingame feedback on what huggers are doing.

## Changelog
:cl:
add: Humans can use a hugger in-hand to crush it.
fix: Huggers don't break storage containers and will leap out now.
fix: Catching huggers works now instead of causing quantum huggers.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
@@ -137,6 +137,8 @@
 	if(stat == DEAD || (sterile && !combat_hugger))
 		return ..() // Dead or sterile (lamarr) can be picked.
 	else if(stat == CONSCIOUS && user.can_be_facehugged(src, provoked = TRUE)) // If you try to take a healthy one it will try to hug or attack you.
+		user.visible_message("<span class ='warning'>\The [src] skitters up [user]'s arm as [user.p_they()] try to grab it!", \
+		"<span class ='warning'>\The [src] skitters up your arm as you try to grab it!")
 		if(!Attach(user))
 			go_idle()
 	return FALSE // Else you can't pick.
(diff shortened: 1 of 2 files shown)


KrabSpider:
Does crushing the hugger automatically drop it?

BraveMole:
I absolutly don't understand why you need to create a new combat hugger type.

QualityVan:
What's the cost on adding a parent type, what's the benefit to having them separated out, and what's the better way to have them inherit common behavior that isn't shared by their current parent?

code/modules/mob/living/carbon/xenomorph/facehuggers.dm
```diff
@@ -662,18 +683,23 @@
 /obj/item/clothing/mask/facehugger/larval
 	name = "larval hugger"
 
-/obj/item/clothing/mask/facehugger/neuro
+///Parent type for all non-larval huggers: can't be worn, is sterile
+/obj/item/clothing/mask/facehugger/combat
+	sterile = TRUE
+	combat_hugger = TRUE
+	flags_equip_slot = 0
```
TiviPlus, line 690: NONE

TiviPlus (changes requested):

TiviPlus (approved):

PR #7513, opened 2021-07-18
Fix rebel squadies not having all their comm channels

BraveMole:
## About The Pull Request

For exemple, rebel SL did not have their command channel. This is the fix

## Changelog
:cl:
fix: Fix rebel squadies not having all their comm channels
/:cl:


TiviPlus (approved):

SplinterGP:
i think this needs to be changed since now they have their own job stuff ?

PR #7514, opened 2021-07-19
This Rebellion starts today.

InterroLouis:
## About The Pull Request

So what I'm doing here is adding in job defines for the rebel faction jobs to differentiate them from the main faction jobs. This will clarify the admin job slot manipulation screen which previously showed two of most every job because of the addition of the rebel jobs. Oh and I fixed a TYPO while I was at it in a file I was already changing. 

## Why It's Good For The Game

Lets admins adjust job slots on a faction by faction basis in the civil war gamemode. 

## Changelog
:cl:
fix: Fixed doubled job slots, now rebel jobs have their own job slots that can be altered separately from regular job slots. 
spellcheck: Fixed medical researcher job display order. 
/:cl:


TiviPlus (approved):
@BraveMole "I told you so"

BraveMole:
Check the following things:

Does this break overwatch? the code there is a bit silly, and potentially everything will stop working for rebels
Does this break preferred squad? It would make sense to break it
I assume this doesn't break shit like squad radio, but test it anyway

InterroLouis:
Right so it doesn't break squad radio, preferred squad was already broken. I did some testing with overwatch for rebels looks to work fine too.

PR #7523, opened 2021-07-19
Defender Can Charge From Fortify

Surrealaser:
## About The Pull Request

Defender can now charge from its Fortify stance when both Charge and Fortify are off cooldown.

## Why It's Good For The Game

Makes Fortify more viable and improves its ability to be integrated into Defender combos as a preface/launch point, making Defender play more fluid and flexible.

## Changelog
:cl:
balance: Defender can now charge from its Fortify stance when both Charge and Fortify are off cooldown.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defender/abilities_defender.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defender/abilities_defender.dm
@@ -135,15 +135,15 @@
 	if(!do_after(X, 0.5 SECONDS, FALSE, X, BUSY_ICON_GENERIC, extra_checks = CALLBACK(src, .proc/can_use_ability, A, FALSE, XACT_USE_BUSY)))
 		return fail_activate()
 
-	var/mob/living/carbon/xenomorph/defender/D = X
-	if(D.fortify)
-		var/datum/action/xeno_action/fortify/FT = X.actions_by_path[/datum/action/xeno_action/fortify]
-		if(FT.cooldown_id)
+	var/mob/living/carbon/xenomorph/defender/defender = X
+	if(defender.fortify)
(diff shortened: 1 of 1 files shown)


SpaceLoveSs13:
Defenders get extra armor by fortifying and one kf the disadvantages of doing so is delay in movement. Why let this fundamental balance be changed? Plus wouldn't that means they negate the standing up time and directly goes to charging wind.

hyper2snyper:
make standing up after charge from fortify take longer.

Surrealaser:
> Defenders get extra armor by fortifying and one of the disadvantages of doing so is delay in standing up. Why let this fundamental balance be changed? Plus wouldn't that means they negate the standing up time and directly goes to charging wind.

This doesn't get around that delay. The cooldown on fortify has to be over before you can charge out of fortify.

code/modules/mob/living/carbon/xenomorph/castes/defender/abilities_defender.dm
```diff
@@ -135,6 +135,17 @@
 	if(!do_after(X, 0.5 SECONDS, FALSE, X, BUSY_ICON_GENERIC, extra_checks = CALLBACK(src, .proc/can_use_ability, A, FALSE, XACT_USE_BUSY)))
 		return fail_activate()
 
+	var/mob/living/carbon/xenomorph/defender/D = X
```
BraveMole, line 138: image

Please surr, this really helps with readability

BraveMole (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/defender/abilities_defender.dm
BraveMole: EXACTLY

TiviPlus (approved):

SplinterGP:
I think this should have a small tm before merge.

PR #7524, opened 2021-07-19
Synth Rank update. *low priority*

thespy24574:
## About The Pull Request

*Do note I don't expect this pull request to be merged or implemented anytime soon, I even expect it to be closed. This is absolutely not a necessary addition to the game, It is here to show people do look forward to these sort of changes and additions.* 

Synthetic role has been recently one of my favorites and it seems to have gotten some traction (From what I have seen) and more and more people have been playing synthetic. Granted not to the level of PO, but definitely getting there. And as more and more people join this game, it's definitely going to need some expansion in terms of ranking system (if its going to stay that is). 

The Synth Rank update adds 2 more ranks to the synth. MK.IV and MK.V. 

Rather simple addition no? Well it seems it isn't. As said by Apple on discord they said they'd wouldn't merge it because the current 3 ranks are enough. (Why are there 5 ranks for squad marine, 4 for PO and RO then?). anyways, Apple has said "Its out of hand is my point. If people cant be normal about it noone will get it" (Copy and pasted directly from the discord). And I agree. but people are stupid in these ways, eventually abnormal will be normal. People want more and yeah.... here we are. At least some of us are civil about it. But eventually this would come up again. Cant avoid it. 

td/dr: People are eventually gonna want this added. but apple says no.

## Why It's Good For The Game

Gives more incentive to play more synth by adding the possibility to flex about how the person is more experienced than the MK.Is. Also gives the people around the actual person some sort of comfort that they wont screw up often. that they know their stuff. I've seen MK.IIIs screw up on a basic thing multiple times, might been a one time thing but sometimes it isn't.

Along with showing experience it gives the newer synthetics something to work towards alongside playing the game. Main reason why I like the rank system to be fair. Which then feeds into the incentive to play longer as a synth. 
 
## Changelog
:cl:
expansion: Adds 2 more ranks.
code: added said 2 ranks.
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -988,8 +988,12 @@ Use your office fax machine to communicate with corporate headquarters or to acq
 			new_human.wear_id.paygrade = "Mk.I"
 		if(601 to 3000) // 10 to 50 hrs
 			new_human.wear_id.paygrade = "Mk.II"
-		if(3001 to INFINITY) // more than 50 hrs
+		if(3001 to 6000) // 50 to 100 hrs
 			new_human.wear_id.paygrade = "Mk.III"
+		if(6001 to 9000) // 100 to 150 hrs
+			new_human.wear_id.paygrade = "Mk.IV"
+		if(9001 to INFINITY) // 150+ hrs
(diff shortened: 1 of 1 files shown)


Ieteihowmee:
Seems good to me.

PR #7532, opened 2021-07-20
Allow the AI to unbolt their core, added a noise for it too.

lbnesquik:
## About The Pull Request
This is a simple which allows AI's to unbolt. There is little usage and it will probably get you killed, but hey, the option is here.

## Changelog
:cl:
expansion: Allow the AI to unbolt their core
/:cl:


neohenry:
The only time AI will unbolt is when xenomorphs hijack and the AI will be unsafe if remained in the AI core.

code/modules/mob/living/silicon/ai/ai_verbs.dm
```diff
@@ -226,6 +226,21 @@
 	popup.set_content(dat)
 	popup.open(FALSE)
 
+/mob/living/silicon/ai/verb/toggle_anchor()
```
BraveMole, line 229: Don't add new verbs.

Make this an ability with a proper keybind, or something like that

BraveMole (changes requested):

lbnesquik:
There was a discussion in dev general where that verb usage was said to be okay.

Ieteihowmee:
Many times this would have been handy

TiviPlus (approved):

PR #7533, opened 2021-07-20
readds tl-127 aim mode

Pariah919:
## About The Pull Request
addds tl-127 aim mode
2.35 seconds per shot in aimmode, compared to the old like 1.9

## Why It's Good For The Game
undeltes the tl-127, which is literally probably the worst gun in the game (that you can get from a marine vendor, CONSISTENTLY, I'd argue some seasonals are worse than it.), considering the mosin isliterally just better in every way conceiveable and the DMR has aim mode AND fires faster and has better situational usage in most scenarios

## Changelog
:cl:
balance: tl has aim mode
/:cl:


BraveMole:
Based MJP

TiviPlus (approved):

PR #7534, opened 2021-07-20
Transvitox Bonus Damage Now Applies to Converted Burn Damage Too

Surrealaser:
## About The Pull Request

Transvitox bonus damage now also applies to burn damage converted as well as brute damage dealt to a victim under its effects.

## Why It's Good For The Game

Transvitox is easily the weakest xeno chem and has been notably underwhelming in practice thus far; this should help boost it a little, while also making its behaviour more consistent.

## Changelog
:cl:
balance: Transvitox bonus damage now also applies to burn damage converted as well as brute damage dealt to a victim under its effects.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -591,7 +591,7 @@
 		dam = fire_loss
 
 	L.heal_limb_damage(burn = dam, updating_health = TRUE) //Heal damage equal to toxin damage dealt; heal before applying toxin damage so we don't flash kill the target
-	L.adjustToxLoss(dam)
+	L.adjustToxLoss(dam * (1 + 0.1 * tox_cap_multiplier)) //Apply toxin damage. Deal extra toxin damage equal to 10% * the tox cap multiplier
 
 	return ..()
 


code/modules/reagents/reagents/toxin.dm
```diff
@@ -591,7 +591,7 @@
 		dam = fire_loss
 
 	L.heal_limb_damage(burn = dam, updating_health = TRUE) //Heal damage equal to toxin damage dealt; heal before applying toxin damage so we don't flash kill the target
-	L.adjustToxLoss(dam)
+	L.adjustToxLoss(dam * (1 + 0.1 * tox_cap_multiplier)) //Apply toxin damage. Deal extra toxin damage equal to 10% * the tox cap multiplier
```
SplinterGP, line 594: uh, what this has to do with burn or what ?

SplinterGP (changes requested):

code/modules/reagents/reagents/toxin.dm
Surrealaser: it applies the existing brute damage bonus to the burn damage when it's converted to tox

SplinterGP, line 594: But how this has to do with burn damage itt just has something tox cap or something, even the comment states so.

SplinterGP (changes requested):

code/modules/reagents/reagents/toxin.dm
Surrealaser: This is the 'conversion' portion, where, after the burn damage is healed, an equivalent amount of toxin damage, plus the bonus damage, is applied.

SplinterGP:
ah okay i understand it now

Surrealaser:
Okay, well, let's TM it then.

PR #7115, opened 2021-06-17
nerfs hlin to acceptable armor levels

Pariah919:
## About The Pull Request
95 to 25, so it isn't full immunity, more like 75% with all armor pieces 

## Why It's Good For The Game
full immunity was a mistake

## Changelog
:cl:
balance: hlin no longer gives you full bomb immunity, only around 80%
/:cl:

PR #7537, opened 2021-07-20
nerfs hlin to more acceptable armor levels

Hekzder:
## About The Pull Request

This nerfs Hlin down to about 90% bomb armor with a full suit, so you can self talk about ten grenades instead of.... an infinite amount due to having literal explosive immunity. #7115 was closed for being too drastic at nerfing it to 80% bomb armor, so this compromises between full immunity and a heavy nerf to put it at 90%. 

## Why It's Good For The Game

Hlin, GL, shield is an incredibly strong combo, but even just Hlin and constant GL spam is really, really strong and incredibly annoying to go against. Hlin will still be good and have a niche for GL users, but having literal full immunity is, in my opinion and the opinion of quite a few others, way too strong. 90% reduction is still very good and nothing to scoff at. Being able to tank about ten grenades before going into crit is still very useful for someone spamming grenades.

## Changelog
:cl:
balance: Hlin no longer gives explosive immunity, instead results in about 90 bomb armor. 
/:cl:

### DIFF
--- a/code/modules/modular_armor/attachables.dm
+++ b/code/modules/modular_armor/attachables.dm
@@ -174,7 +174,7 @@
 	desc = "Designed for mounting on the Jaeger Combat Exoskeleton. Uses a complex set of armor plating and compensation to lessen the effect of explosions, at the cost of making the user slower."
 	icon_state = "mod_boomimmune_icon"
 	item_state = "mod_boomimmune"
-	soft_armor = list("bomb" = 95)
+	soft_armor = list("bomb" = 40)
 	slowdown = 0.2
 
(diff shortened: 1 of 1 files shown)


BraveMole:
I suggest just capping bomb armor to 90 or something

turnop:
otherwise known as "makes all grenadiers ivanov"

David-Stormwell (approved):
This should be quite acceptable indeed, the other nerf was fine enough, but this should appease the Grenadier mains more.

TiviPlus (approved):

SplinterGP:
Stupid nerf, Hlin does way less than other modules for somethint very niche

PR #7539, opened 2021-07-20
Increases T-25 Smartrifle damage

Losenis:
## About The Pull Request
Smartrifle damage from 15 to 25. For comparison rifle bullets deal 25 while mg bullets deal 20 comparing a t12 to a gpmg. Even comparing the t25 to t29, smartrifle has less damage and sunder per second for less mag capacity and req only mags + reloading mags with the ammo box.
For each time the t25 shoots 5 times (0.2) the t29 shoots 4 times (0.25), totaling 75 vs 80 dmg and 5 vs 8 sunder
Hell if I did the math right a t12 on aim mode can out-dps it while having vendor ammo assuming full auto, burst auto is even better and even moreso bfa burst auto since burst  doesn't take into account aim delay. If one wants to go smartrifle he's better off even going normal marine with T18 BFA with no worries about running out of ammo.

## Why It's Good For The Game

This should give it more of a reason to be considered over the smartgun as choice for more damage in exchange of ammo shenanigans and less sunder.

## Changelog
:cl:
balance: T-25 Smartrifle damage per hit from 15 to 25
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -999,7 +999,7 @@ datum/ammo/bullet/revolver/tp44
 
 /datum/ammo/bullet/smartgun/smartrifle
 	name = "smartrifle bullet"
-	damage = 15
+	damage = 25
 	penetration = 15
 	sundering = 1
 


BraveMole:
Webedit, shame

Losenis:
The future is now, old man.

RenGusta:
2/3 increase in power kinda whack, should try 20 or something for a TM and then up it if it's still falling short?

And if you have a problem with resupply, could always try addressing their base supply and reloading means.

Losenis:
Resupply isn't a problem by itself, it's just that the smartgun not having it counts as a con for the smartrifle to have something else as a pro. And I took the 2/3 increase in power as okayish snice the smartgun proper still would have 2/3 more sunder per second / _double_ the sunder per hit, and without the inherent IFF it's still just a T12 with 0.5 more sunder and no burst capability for a role-locked weapon if this went through.

SplinterGP (approved):

PR #7543, opened 2021-07-20
Increases price of acid turrets to 150

Hekzder:
## About The Pull Request

Basically the title. Increases the price of Xeno acid turrets to 150 psy points.

## Why It's Good For The Game

Xeno turrets as they stand are really strong due to all of the factors surrounding them. A high fire rate, really high health, hive alert when attacking, and only costing 100 psy points. As of now you can psy drain five Marines and turn around and make three turrets with those psy points alone, not counting _any_ other form of psy point generation like generators or active cocoons. 

So, to keep them retaining their current potential and overall power, this is a cost increase to make the cost match the power a little more and force Xenos to use turrets slightly more sparingly or make a bit of a harder decision when it comes to placing a turret vs saving points for silo or king. 

## Changelog
:cl:
balance: Xeno acid turrets now cost 150 psy points
/:cl:


InterroLouis:
It's not your fault. The integration tests are failing randomly lately.

lbnesquik:
I feel like this is test merge worthy

GradusNL:
Turrets are still easily countered despite their high health using nades, CAS or sniping. The damage only kills marines who stand in front of the turret, since they can outrun the turret if not standing right next to it. The projectile speed is too low to hit marines moving at a distance. Please describe how the factors you describe as making them strong are actually an issue for marines in gameplay.

Turrets are ONLY strong when multiple xeno's support them by force moving marines into them and stunning them. Even that is counterable with the current balance, since marines can just force the xeno's back anyways. They don't need a nerf, not in cost or stats.

Hekzder:
The issue for Marines in gameplay generally boils down to the fact that turrets are dirt cheap and very easy to spam. It's frankly not incredibly fun and given how useful they are in delaying pushes, stopping silo hunters, giving intel, and being combo'd with stuns and CC for what is normally a guaranteed kill, I'm suggesting a price increase that brings them from "dirt cheap" to just cheap.

To put it in perspective, as I outlined in the PR, 100 points is equivalent to two Psy drains, you still have 20 points left over. Or a single cocoon, and you still have 20 points left over. Do you really think a turret is equivalent in value to 1.66 dead Marines that you spent about five seconds Psydraining or a single Marine that you cocooned for the full duration? This is completely ignoring any passive Psy point generation by the way, like through generators.

TiviPlus:
As Gradus said turrets are easy enugh to deal with, and currently there is too much rushing xenos, as they have no turrets, and too much focus on king since its a better bang for your buck than a turret will be

PR #7545, opened 2021-07-20
Lower the price of sadar WP and HE by 1 point, so we see something besides AP for once.

lbnesquik:
## About The Pull Request
See title. Decrease is from 7 to 6. Not huge, but it might be enough to change things a little.

## Changelog
:cl:
balance: Very slight reduction in price of HE and WP sadar rockets, with the aim of varying ammo usages.
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -525,7 +525,7 @@ AMMO
 /datum/supply_packs/ammo/rpg_regular
 	name = "T-152 RPG HE rocket"
 	contains = list(/obj/item/ammo_magazine/rocket/sadar)
-	cost = 7
+	cost = 6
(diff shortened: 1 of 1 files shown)


Hekzder:
I almost feel like you could maybe make it 5? A one point reduction is not super substantial, even in a bulk purchase.

Pariah919:
don't reduce HE
it's better than AP but because of complacency gaming syndrome everyone teters to AP because that's just what they think does the most direct hit damage, but he does more direct hit damage because of the explosion.
Reduce WP to 5 so it has some use

lbnesquik:
WP has been reduced to five. He is still 6 after feedback

Ieteihowmee:
Nice

TiviPlus (approved):

PR #7560, opened 2021-07-21
sg can buy t-81 instead of smartrifle or machinegun

Pariah919:
## About The Pull Request
sg can buy t-81
no nvg scope though you gotta buy that seperately for 15 points else it'd be overpowered so i added the crate for it
oh and t81 can use normal marine scopes for the 1 in 50000 rounds that it'll matter

## Why It's Good For The Game
gives sg more variety

## Changelog
:cl:
expansion: Expands sg content with new gun
balance: t-81 can use normal marine scopes, and sgs can buy t-81s without nvg scopes in their thing instead of t-29s or smartrifles
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -377,6 +377,14 @@
 	new /obj/item/ammo_magazine/rifle/standard_smartrifle(src)
 	new /obj/item/ammo_magazine/rifle/standard_smartrifle(src)
 
+/obj/item/storage/belt/marine/t81/Initialize()
+	. = ..()
+	new /obj/item/ammo_magazine/rifle/autosniper(src)
+	new /obj/item/ammo_magazine/rifle/autosniper(src)
+	new /obj/item/ammo_magazine/rifle/autosniper(src)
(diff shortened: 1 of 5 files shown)


BraveMole:
FFS

Losenis:
Is there an actual benefit to the SGnr using it rather than just req buying it for the bog standard unga? Because I can picture people lyching the SGnr for deciding not to pick the IFF MG for helping the push.

Ieteihowmee:
It is a smart gunner, not a fucking sniper class. I will never get it. It should be a rule that we kill SGs who throw out their SG for a sniper.

Ieteihowmee:
> Is there an actual benefit to the SGnr using it rather than just req buying it for the bog standard unga? Because I can picture people lyching the SGnr for deciding not to pick the IFF MG for helping the push.

Makes me so mad when I see SGs waste their role.

David-Stormwell:
Seethe and Cope machine gunners.
Though I'm against this because the sprite of the auto sniper is dogshit and shouldn't be allowed to exist

Pariah919:
i should probably improve the t-81 sprite

Issue #7500, opened 2021-07-18
Pillar of Spring has a misplaced TerraGov Poster in preps


lKiseki:
### Description

Per title. There is a misplaced TerraGov Poster in Smartgunner preps. I'm assuming it's meant to be on the wall.

### Test Merges

Irrelevant.

### Reproduction Steps

Irrelevant.

### Screenshots

![image](https://user-images.githubusercontent.com/59634950/126061242-6c791f1b-b39b-414c-a9be-3c4b122b07e4.png)

BraveMole:
fixed iirc

PR #7570, opened 2021-07-21
Fix the poster bulging downward. As it turns out, there is a weird thing with the direction giving out a pixel offset.

lbnesquik:
## About The Pull Request

image

Fixes this issue : https://github.com/tgstation/TerraGov-Marine-Corps/issues/7500

## Changelog
:cl:
fix: fixed a misplaced poster.
/:cl:

### DIFF
--- a/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
+++ b/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
@@ -8408,8 +8408,10 @@
 /turf/open/floor/mainship/mono,
 /area/mainship/hallways/port_hallway)
 "Bv" = (
-/obj/structure/sign/prop1,
-/turf/closed/wall/mainship,
+/obj/structure/sign/prop1{
+	dir = 1
+	},
+/turf/open/floor/mainship,
 /area/mainship/squads/general)
 "Bx" = (
 /obj/effect/decal/warning_stripes/thin{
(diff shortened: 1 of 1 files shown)


Ieteihowmee:
Oh god, will we have to properly fix this in maps as well? These posters on some ships were placed offset due to this bug, now they will all be thrown out of alignment!

lbnesquik:
> 
> 
> Oh god, will we have to properly fix this in maps as well? These posters on some ships were placed offset due to this bug, now they will all be thrown out of alignment!

I have noticed a couple of posters which weren't sticking to walls. That whole 'dir' thing seems to be the reason.

TiviPlus (approved):

Ieteihowmee:
> > Oh god, will we have to properly fix this in maps as well? These posters on some ships were placed offset due to this bug, now they will all be thrown out of alignment!
> 
> I have noticed a couple of posters which weren't sticking to walls. That whole 'dir' thing seems to be the reason.

So it does not mess with all the maps, there will be no floating posters even after this?

lbnesquik:
> 
> 
> > > Oh god, will we have to properly fix this in maps as well? These posters on some ships were placed offset due to this bug, now they will all be thrown out of alignment!
> > 
> > 
> > I have noticed a couple of posters which weren't sticking to walls. That whole 'dir' thing seems to be the reason.
> 
> So it does not mess with all the maps, there will be no floating posters even after this?

Uhhh, dunno. I fixed this one on the map files, so probably there will be.

PR #7574, opened 2021-07-21
Hunters hide passively like how runners/larva can

QualityVan:
## About The Pull Request
Sets hunter's default layer to XENO_HIDING_LAYER, the same one runners and larva can enter. This puts them below most objects and other mobs.
Adjusts hunter's pounce to not revert their layer the way runner's does, since they can't set it manually.

## Why It's Good For The Game
Stealthy beno, and also I keep getting annoyed at failing to slash people after I pounce them (skill issue).
Just giving them the hide action was an option, but I like this better because hunters already have an action with the same icon (stealth) and naturally staying slightly hidden seemed like a hunter thing.

## Changelog
:cl:
add: Hunters now passively hide behind stuff the way larva and runners can choose to.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/hunter/abilities_hunter.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hunter/abilities_hunter.dm
@@ -213,6 +213,9 @@
 	plasma_cost = 20
 	range = 7
 
+//Don't need to do anything here, just overriding runner's pounce so we maintain layer
+/datum/action/xeno_action/activable/pounce/hunter/prepare_to_pounce()
+
 // ***************************************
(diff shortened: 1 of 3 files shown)


Ieteihowmee:
Nice. Makes hunters a bit more scary.

TiviPlus:
Both myself and splinter agree that this doesnt fit thematically as hunter can already turn invisible
I also dont like that we dont have anything to counter hidepouncing yet

QualityVan:
Hunter stealth is not tremendously strong, and is becoming weaker with the bugfix. You can still shoot someone who pouncehides by clicking behind them, they're stuck in place for a moment.

PR #7586, opened 2021-07-22
Bundles the Mimir Mark 2 modules in req, makes it slightly cheaper overall

Hekzder:
## About The Pull Request

Bundles Mimir Mark 2 helmet and armor module into one req thingy for 15 points. Matching the Mark 1 Module that is bundled when you get it from the closet, also it just doesn't make sense to have one but not the other. This is a slight buff to Mimir since the overall cost is now 15, down from a really high 17. 3 for the helmet mod, 12 for the armor mod. 

## Why It's Good For The Game

There is very little reason for this not to be bundled and it's already bundled for Mark 1 in the closet, also the price chance just makes Mimir 2 not break the early game bank. A Mimir Mark 2 set should not cost the same as a Valk + Fulton pack, for example.

## Changelog
:cl:
add: Adds a bundle to req for both Mimir Mark 2 modules that replaces the previously separate Mimir Mark 2 modules.
balance: Full Mimir Mark 2 set now costs 15 points, down from 17.
/:cl:


Hekzder:
This will probably conflict with #7587 since I also updated the name, but I'll fix either PR depending if and when each is merged.

Hekzder:
I'd like maintainer input before I do it, but maybe a further price reduction to 14 since it can't really be gamed anymore to have a Mimir 2 without also buying the Mimir 1? Plus, the Mimir 2 is pretty heavily gimped without also the helmet mod. It should be slightly more expensive than other modules for also having a head module, but I think 14 is a bit more reasonable.

SkeletalElite:
a lot of people actually just use mimir mark 1 helmet with the mark 2 armor. I believe the only difference between the mark 1 helmet piece and mark 2 helmet piece is a bit of acid armor so this ends up being a nerf over all

Hekzder:
> a lot of people actually just use mimir mark 1 helmet with the mark 2 armor. I believe the only difference between the mark 1 helmet piece and mark 2 helmet piece is a bit of acid armor so this ends up being a nerf over all

Given how prominent head aiming is when it specifically comes to ranged acid, this is not really a smart idea. If you're running Mimir 2 you may as well be running the full set. Also, a contributing factor to that right now is probably the absurd price of the full Mimir Mark 2 set being 17 points, 5 for helmet, 12 for armor mod. This remedies that by both bundling them and making it cheaper overall, therefore more accessible to get early on.

code/modules/reqs/supplypacks.dm
```diff
@@ -824,16 +824,13 @@ ARMOR
 	cost = 12
 
 /datum/supply_packs/armor/modular/attachments/mimir_environment_protection
-	name = "Jaeger mimir module"
+	name = "Jaeger Mark 2 Mimir module bundle"
```
TiviPlus, line 827:
```suggestion
	name = "Jaeger Mark 2 Mimir module set"
```

TiviPlus (changes requested):
Remove excess spaces from your changelog

SkeletalElite:
> > a lot of people actually just use mimir mark 1 helmet with the mark 2 armor. I believe the only difference between the mark 1 helmet piece and mark 2 helmet piece is a bit of acid armor so this ends up being a nerf over all
> 
> Given how prominent head aiming is when it specifically comes to ranged acid, this is not really a smart idea. If you're running Mimir 2 you may as well be running the full set. Also, a contributing factor to that right now is probably the absurd price of the full Mimir Mark 2 set being 17 points, 5 for helmet, 12 for armor mod. This remedies that by both bundling them and making it cheaper overall, therefore more accessible to get early on.

If it's still going to be more expensive than a standard module, you may as well keep the head piece separate but just reduce the cost of the head piece to 3.

Hekzder:
> Remove excess spaces from your changelog

Fixed, my bad. Also did the same for my other PRs up. The merge conflict is also dealt with so this should be good to merge now.

TiviPlus (approved):

PR #7589, opened 2021-07-22
Removes Defunct Queen Buttons / Actions (Larval Accelerant and Psychic Whisper).

KornFlaks:
## About The Pull Request

This removes two largely vestigial actions in the queen button lineup that exist only to confuse new players.

## Why It's Good For The Game

As Tivi noted, the queen has way too many buttons. New players are confused and intimidated by choices when largely 4 buttons are used in normal gameplay. A reduction in choice lowers the skill floor for new Queens to be slightly more robust.

## Changelog
:cl:
del: Removed the following two buttons from queen: Larval Accelerant Injection and Psychic Whisper.
/:cl:


BraveMole:
Psychic is potential rp

KornFlaks:
Potential RP yes but isn't it even more RP that you cant talk to the queen directly but instead through an intermediary translator like the Shrike? Which also has psychic whisper that is not being removed.

TiviPlus (approved):

PR #7595, opened 2021-07-23
Fix two xeno exploits

spookydonut:
## About The Pull Request
Warrior punch didn't check resist flags using bitmath and crusher charge only checked for indestructible.

Can't check for unacidable for crusher charge because theres a lot of things that cant be melted but should be crushable.

## Changelog
:cl:
fix: Warrior punch no longer allows you to destroy many machines that can not otherwise be damaged
fix: Crushers can no longer destroy SMESes
/:cl:

### DIFF
--- a/code/__DEFINES/obj_flags.dm
+++ b/code/__DEFINES/obj_flags.dm
@@ -11,5 +11,6 @@
 #define ON_FIRE (1<<2) //currently on fire
 #define XENO_DAMAGEABLE (1<<3) //xenos can damage this by slashing and spitting
 #define DROPSHIP_IMMUNE (1<<4) //dropship cannot land on it
+#define CRUSHER_IMMUNE (1<<5)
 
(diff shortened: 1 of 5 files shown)


GradusNL:
Why are these an issue? APC's can be rebuild quite easily and you can just wire power generators directly into the grid. I'd prefer to call them unintended mechanics instead of exploits.

AnCopper:
> Why are these an issue? APC's can be rebuild quite easily and you can just wire power generators directly into the grid. I'd prefer to call them unintended mechanics instead of exploits.

I belive that's called an exploit.

Wayland-Smithy:
> fix: Crushers can no longer destroy SMESes

Balance imo. Yet another engi buff. The warrior stuff does make sense as a long overdue fix.

TiviPlus (approved):

PR #7602, opened 2021-07-24
Fultons are now limited use and more expensive, decreases price of ASRS, and adds fultons to SL vendor and operational vendor

Hekzder:
## About The Pull Request

Basically everything in the title, fulton gets a price increase from 5 to 10 and is limited to three uses, ASRS gets a hefty price decrease from 50 to 30, the fulton is added to the SL vendor at 22 points, and the req operational vendor has two fultons in it at round start.

## Why It's Good For The Game

The ASRS is literally, completely useless in it's current state due to being completely overshadowed by fultons. This is meant to make the fultons somewhat more reasonably priced for their versatility and limit them as well, while also making the ASRS more appealing. Do you approve the three fulton orders that the Marines put in or just give them an ASRS to use instead? This should make the ASRS much more competitive with fultons, given the unlimited vs limited aspect and the fulton's increase in price. 

Adding it to the SL vendor makes sense, presents a competitive option to just buying the RB pack every round, and is somewhat of a compromise for fultons being made quite a bit more expensive from req, the same principal applies to the operational vendor change. Especially since mining comps are in there as standard, fultons may as well be in there too. 

## Changelog
:cl:
add: Fultons can now be purchased from the SL vendor for 20 points and two fultons are in the req operational vendor at round start
balance: Fultons now cost 10 points from req and have only three uses
balance: The ASRS now costs 30 points from req
/:cl:

### DIFF
--- a/code/__DEFINES/loadout.dm
+++ b/code/__DEFINES/loadout.dm
@@ -143,6 +143,7 @@ GLOBAL_LIST_INIT(leader_gear_listed_products, list(
 		/obj/structure/closet/bodybag/tarp = list(CAT_LEDSUP, "V1 thermal-dampening tarp", 5, "black"),
 		/obj/item/storage/box/m94/cas = list(CAT_LEDSUP, "CAS flare box", 10, "orange"),
 		/obj/item/deployable_camera = list(CAT_LEDSUP, "Deployable Overwatch Camera", 2, "orange"),
+		/obj/item/fulton_extraction_pack = list(CAT_LEDSUP, "Fulton Extraction Pack", 25, "orange"),
 	))
 
 GLOBAL_LIST_INIT(smartgunner_gear_listed_products, list(
--- a/code/__DEFINES/loadout.dm
+++ b/code/__DEFINES/loadout.dm
@@ -143,7 +143,7 @@ GLOBAL_LIST_INIT(leader_gear_listed_products, list(
 		/obj/structure/closet/bodybag/tarp = list(CAT_LEDSUP, "V1 thermal-dampening tarp", 5, "black"),
 		/obj/item/storage/box/m94/cas = list(CAT_LEDSUP, "CAS flare box", 10, "orange"),
 		/obj/item/deployable_camera = list(CAT_LEDSUP, "Deployable Overwatch Camera", 2, "orange"),
-		/obj/item/fulton_extraction_pack = list(CAT_LEDSUP, "Fulton Extraction Pack", 22, "orange"),
+		/obj/item/fulton_extraction_pack = list(CAT_LEDSUP, "Fulton Extraction Pack", 20, "orange"),
 	))
 
 GLOBAL_LIST_INIT(smartgunner_gear_listed_products, list(
--- a/code/__DEFINES/loadout.dm
+++ b/code/__DEFINES/loadout.dm
@@ -143,7 +143,7 @@ GLOBAL_LIST_INIT(leader_gear_listed_products, list(
 		/obj/structure/closet/bodybag/tarp = list(CAT_LEDSUP, "V1 thermal-dampening tarp", 5, "black"),
 		/obj/item/storage/box/m94/cas = list(CAT_LEDSUP, "CAS flare box", 10, "orange"),
 		/obj/item/deployable_camera = list(CAT_LEDSUP, "Deployable Overwatch Camera", 2, "orange"),
-		/obj/item/fulton_extraction_pack = list(CAT_LEDSUP, "Fulton Extraction Pack", 25, "orange"),
+		/obj/item/fulton_extraction_pack = list(CAT_LEDSUP, "Fulton Extraction Pack", 22, "orange"),
 	))
 
 GLOBAL_LIST_INIT(smartgunner_gear_listed_products, list(
--- a/code/modules/requisitions/fulton.dm
+++ b/code/modules/requisitions/fulton.dm
@@ -38,6 +38,9 @@
 	user.visible_message(span_notice("[user] finishes attaching [src] to [spirited_away] and activates it."),\
 	span_notice("You attach the pack to [spirited_away] and activate it. This looks like it will yield [export_report.points ? export_report.points : "no"] point[export_report.points == 1 ? "" : "s"]."), null, 5)
 	uses --
+	if(uses < 1)
+		usr.temporarilyRemoveItemFromInventory(src)
+		moveToNullspace()
 
 	qdel(spirited_away)
 
@@ -98,6 +101,8 @@
 	holder_obj.pixel_z = initial(pixel_z)
 	holder_obj.vis_contents -= baloon
 	baloon.icon_state = initial(baloon.icon_state)
+	if(uses < 1)
+		qdel(src)
 	active = FALSE
 
 
--- a/code/modules/requisitions/fulton.dm
+++ b/code/modules/requisitions/fulton.dm
@@ -39,7 +39,7 @@
 	span_notice("You attach the pack to [spirited_away] and activate it. This looks like it will yield [export_report.points ? export_report.points : "no"] point[export_report.points == 1 ? "" : "s"]."), null, 5)
 	uses --
 	if(uses < 1)
-		usr.temporarilyRemoveItemFromInventory(src)
+		usr.temporarilyRemoveItemFromInventory(src) //Removes the item without qdeling it, qdeling it this early will break the rest of the procs
 		moveToNullspace()
 
 	qdel(spirited_away)
--- a/code/modules/requisitions/fulton.dm
+++ b/code/modules/requisitions/fulton.dm
@@ -9,7 +9,7 @@
 	///Reference to the balloon vis obj effect
 	var/atom/movable/vis_obj/fulton_baloon/baloon
 	var/obj/effect/fulton_extraction_holder/holder_obj
-	var/uses = 3
+	var/uses = 3 //How many times you can use the fulton before it goes poof
 
 /obj/item/fulton_extraction_pack/examine(mob/user)
 	..()
@@ -37,9 +37,7 @@
 		SSpoints.export_history += export_report
 	user.visible_message(span_notice("[user] finishes attaching [src] to [spirited_away] and activates it."),\
 	span_notice("You attach the pack to [spirited_away] and activate it. This looks like it will yield [export_report.points ? export_report.points : "no"] point[export_report.points == 1 ? "" : "s"]."), null, 5)
-	uses -= 1
-	if(zero_uses())
-		return TRUE
+	uses --
 
 	qdel(spirited_away)
 
@@ -101,12 +99,6 @@
 	holder_obj.vis_contents -= baloon
 	baloon.icon_state = initial(baloon.icon_state)
 	active = FALSE
-
-/obj/item/fulton_extraction_pack/proc/zero_uses()
-	if(uses < 1)
-		qdel(src)
-		return TRUE
-	return FALSE
 
 
 /obj/effect/fulton_extraction_holder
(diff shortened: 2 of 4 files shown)


BraveMole:
Really meh, ARSR should just disapear.

Asking req to invest points to get points is silly

EvilMelonTheSecond:
ASRS pad was for engineers to set up near miners to create a reliable point generator. People mainly used it in FOB due to lazy engineers/hard to secure miners but dragging a xeno corpse to FOB isn't much different than dragging it in to alamo for RO to collect them on cycle. I would say giving out a free ASRS pad at roundstart is a much better idea than fiddling with the price.

And the fulton price change seems unnecessary. High priced fultons will just lead to them being left in the FOB to make sure they aren't lost forever in the satchel of a dead marine.

Hekzder:
Yeah, probably just going to not be lazy and nerf fultons in a more proper way. Going to draft this for now, though I will address this: 



> Really meh, ARSR should just disapear.
> 
> Asking req to invest that much points to get points is silly

"That much points" is a silly notion when the cheapest Xeno sells for 30 and T2 sells for 40. Killing and selling Young Hunter (752) and selling it pays for the ASRS with a bit left over at that price point.

Hekzder:
Update the PR, this now makes fultons limited use but still does most of what was originally intended. A fulton price increase and ASRS price decrease. The code for the limited uses is potentially shit, I mostly just stole it from stack code, but it works. Someone can proof me on it and review.

Fultons also added to the req operational vendor at round start per @BraveMole's suggestion, makes sense since they will be becoming limited use and mining comps are already in there.

Al-1ce:
>I would say giving out a free ASRS pad at roundstart is a much better idea than fiddling with the price.
Yup. 

>And the fulton price change seems unnecessary. High priced fultons will just lead to them being left in the FOB to make sure they aren't lost forever in the satchel of a dead marine.
This would encourage use of the ASRS.

Hekzder:
> > I would say giving out a free ASRS pad at roundstart is a much better idea than fiddling with the price.
> 
> Yup.
> 
> > And the fulton price change seems unnecessary. High priced fultons will just lead to them being left in the FOB to make sure they aren't lost forever in the satchel of a dead marine.
> 
> This would encourage use of the ASRS.

I'd prefer to stick with Marines just getting a couple free fultons at round start by virtue of it requiring no mapping work, but if the maintainers would prefer a free ASRS at round start in lieu of the free fultons or as well as the free fultons I could go with that. I don't hate the idea. 

With two free fultons at round start and potential to get more via SL vendor I can see the ASRS being more of a sound purchase over buying some fultons, since you can have quite a few people running around with fultons on the ground but they are now limited use and there's always the chance they die and no one ever gets the fultons back. I don't think it would be a waste of points to get an ASRS early on (maybe not right at start of op, but early) and not spend any points on fultons, instead just relying on the SL fultons and the free op fultons for field recovery of Xenos.

code/modules/requisitions/fulton.dm
```diff
@@ -94,6 +102,12 @@
 	baloon.icon_state = initial(baloon.icon_state)
 	active = FALSE
 
+/obj/item/fulton_extraction_pack/proc/zero_uses()
```
SplinterGP, line 105: No need for this proc just do it where the proc is called.

SplinterGP (changes requested):

code/modules/requisitions/fulton.dm
```diff
@@ -32,6 +37,9 @@
 		SSpoints.export_history += export_report
 	user.visible_message(span_notice("[user] finishes attaching [src] to [spirited_away] and activates it."),\
 	span_notice("You attach the pack to [spirited_away] and activate it. This looks like it will yield [export_report.points ? export_report.points : "no"] point[export_report.points == 1 ? "" : "s"]."), null, 5)
+	uses -= 1
```
BraveMole, line 40:
```suggestion
	uses --
```

code/modules/requisitions/fulton.dm
```diff
@@ -9,6 +9,11 @@
 	///Reference to the balloon vis obj effect
 	var/atom/movable/vis_obj/fulton_baloon/baloon
 	var/obj/effect/fulton_extraction_holder/holder_obj
+	var/uses = 3
```
BraveMole, line 12: autodoc

code/modules/requisitions/fulton.dm
```diff
@@ -32,6 +37,9 @@
 		SSpoints.export_history += export_report
 	user.visible_message(span_notice("[user] finishes attaching [src] to [spirited_away] and activates it."),\
 	span_notice("You attach the pack to [spirited_away] and activate it. This looks like it will yield [export_report.points ? export_report.points : "no"] point[export_report.points == 1 ? "" : "s"]."), null, 5)
+	uses -= 1
+	if(zero_uses())
+		return TRUE
```
BraveMole, line 42: do as splinter says. And why return TRUE anyway? You should just qdel when uses reach zero, but certainly not returning before qdel(spirited_away)

BraveMole (changes requested):

Wayland-Smithy:
> the cheapest Xeno sells for 30

I'd like to reiterate that for those who still disagree with the price increase and use limit. 1 lowest tier xeno pays for the 10 point fulton plus an extra 200% and you still have 2 more uses! Xeno corpses can not decompose and are a pain for other xenos to drag away and even then you are guaranteed to find them when pushing, sometimes at the silo you are heading toward anyway, for free points that can never be deleted once that xeno dies.

Hekzder:
Got busy and kind of forgot about this for a bit. Will try and address feedback and also hash out some issues soon to get this merge/TM ready. 

One issue I'm running into is that qdeling the fulton too soon, IE in the extract proc like I am now, doesn't play the animation for fultoning the third Xeno fultoned and just qdels them automatically, presumably because the rest of the procs no longer have an item to call from. Someone with more coding knowledge than me can probably help as to where the best position to place the proc is to prevent this. 

Placing it at the very end when it's running clean up for the extract causes the fulton to linger at zero uses for a few seconds, and then disappears once the entire animation and everything is done. Which _works_ but is obviously less than ideal in a pretty big way.

Hekzder:
The deletion proc upon zero uses now actually works thanks to some help from QV in Discord. This should be good for another review pass and/or a merge of some kind.

code/modules/requisitions/fulton.dm
```diff
@@ -9,6 +9,11 @@
 	///Reference to the balloon vis obj effect
 	var/atom/movable/vis_obj/fulton_baloon/baloon
 	var/obj/effect/fulton_extraction_holder/holder_obj
+	var/uses = 3 //How many times you can use the fulton before it goes poof
+
+/obj/item/fulton_extraction_pack/examine(mob/user)
+	..()
```
TiviPlus, line 15: . = ..()

code/modules/requisitions/fulton.dm
```diff
@@ -9,6 +9,11 @@
 	///Reference to the balloon vis obj effect
 	var/atom/movable/vis_obj/fulton_baloon/baloon
 	var/obj/effect/fulton_extraction_holder/holder_obj
+	var/uses = 3 //How many times you can use the fulton before it goes poof
```
TiviPlus, line 12: autodoc

code/__DEFINES/loadout.dm
```diff
@@ -143,6 +143,7 @@ GLOBAL_LIST_INIT(leader_gear_listed_products, list(
 		/obj/structure/closet/bodybag/tarp = list(CAT_LEDSUP, "V1 thermal-dampening tarp", 5, "black"),
 		/obj/item/storage/box/m94/cas = list(CAT_LEDSUP, "CAS flare box", 10, "orange"),
 		/obj/item/deployable_camera = list(CAT_LEDSUP, "Deployable Overwatch Camera", 2, "orange"),
+		/obj/item/fulton_extraction_pack = list(CAT_LEDSUP, "Fulton Extraction Pack", 22, "orange"),
```
TiviPlus, line 146:
```suggestion
		/obj/item/fulton_extraction_pack = list(CAT_LEDSUP, "Fulton Extraction Pack", 20, "orange"),
```

TiviPlus (changes requested):

code/modules/requisitions/fulton.dm
Hekzder: ?

code/modules/requisitions/fulton.dm
Hekzder: ?

code/modules/requisitions/fulton.dm
```diff
@@ -32,6 +37,10 @@
 		SSpoints.export_history += export_report
 	user.visible_message(span_notice("[user] finishes attaching [src] to [spirited_away] and activates it."),\
 	span_notice("You attach the pack to [spirited_away] and activate it. This looks like it will yield [export_report.points ? export_report.points : "no"] point[export_report.points == 1 ? "" : "s"]."), null, 5)
+	uses --
```
SplinterGP, line 41:
```suggestion
	uses--
```

code/modules/requisitions/fulton.dm
Hekzder: Should be fixed.

code/modules/requisitions/fulton.dm
Hekzder: Should be fixed.

SplinterGP (changes requested):

code/modules/requisitions/fulton.dm
```diff
@@ -32,6 +38,10 @@
 		SSpoints.export_history += export_report
 	user.visible_message(span_notice("[user] finishes attaching [src] to [spirited_away] and activates it."),\
 	span_notice("You attach the pack to [spirited_away] and activate it. This looks like it will yield [export_report.points ? export_report.points : "no"] point[export_report.points == 1 ? "" : "s"]."), null, 5)
+	uses--
+	if(uses < 1)
+		usr.temporarilyRemoveItemFromInventory(src) //Removes the item without qdeling it, qdeling it this early will break the rest of the procs
```
TiviPlus, line 43: user not usr

Hekzder: Done.

Pariah919:
IMO I like this other than the fact that i feel like there could be one use fulton packs
Marines have a deficit of utility items that they can bring outside of certain stuff and one use fulton packs could fill that niche very well in exchange for being storage inefficent as heck, and being decently limited (make them unbuyable, as the multi use one would outdo them anyway.)

Hekzder:
> IMO I like this other than the fact that i feel like there could be one use fulton packs
> Marines have a deficit of utility items that they can bring outside of certain stuff and one use fulton packs could fill that niche very well in exchange for being storage inefficent (normal sized or small) as heck, and being decently limited (make them unbuyable, as the multi use one would outdo them anyway, unless you make them like 3 points per).

I don't mind this idea but would prefer to pursue it in a separate PR once we have a finalized sprite for it and to not delay this PR getting potentially merged any longer.

joniroq:
This sucks. Getting marines to carry a fulton pack to actually fulton corpses is hard enough as it is. making it limited use makes it even less worthwhile

Hekzder:
> This sucks. Getting marines to carry a fulton pack to actually fulton corpses is hard enough as it is. making it limited use makes it even less worthwhile

You get two free fultons in operational vendor and SLs can buy them without using req points. With potentially four SLs and the two free fultons from req, with this PR there should overall be _more_ fultons in the hands of Marines, just the fultons that they are using aren't completely broken and unlimited use point generators. If you can't find two Marines out of the entirety of the Marine force that are willing to take and use a fulton it's probably like 10 pop and fultons don't matter because it's Crash.

PR #7429, opened 2021-07-14
Replaced all span class by span macros

BraveMole:
## About The Pull Request

Tell me when you are reviewing this, cause i ain't fixing conflict every afternoon

Title, ported from TG:

https://github.com/tgstation/tgstation/pull/59645 <-- regex were taken from here
https://github.com/tgstation/tgstation/pull/59685

CL and minimum compiler version bumped to 1539

image

## Why It's Good For The Game

Easier to type, less prone to errors

## Changelog
:cl:
code: Replaced all span class by span macros
/:cl:

PR #7603, opened 2021-07-24
[DNM] Revert "Replaced all span class by span macros"

SplinterGP:
Reverts tgstation/TerraGov-Marine-Corps#7429

apparently broke chat and i had to make it into local branch, just TM


SplinterGP:
should have been fixed

PR #7623, opened 2021-07-26
Nerfs T-25 bullet to a middle ground of 20 damage, also reduces melee force to 25 base from 35 base

Hekzder:
## About The Pull Request

Basically the title, #7539 was probably a bit of an overcorrection in terms of smartrifle damage, 20 is a good middle ground. I agree that 15 was too weak but disagree that 25 is more reasonable, it definitely makes it a tad strong.

Also the T-25 has big meme melee force right now while also supporting a bayonet, giving it higher melee force than the T-29 is kind of silly in that regard.

## Why It's Good For The Game

Balance is good, increasing the damage of a gun by 66% that has faster wield delay, fire rate, and probably other factors that I'm missing compared to the T-29 was probably not a good idea. This is why we balance things in increments instead of a huge increase at once. 

## Changelog
:cl:

balance: T-25 now does 20 damage instead of 25, has 25 melee force instead of 35

/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1001,7 +1001,7 @@ datum/ammo/bullet/revolver/tp44
 
 /datum/ammo/bullet/smartgun/smartrifle
 	name = "smartrifle bullet"
-	damage = 25
+	damage = 20
 	penetration = 15
 	sundering = 1
 
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -911,7 +911,7 @@
 	item_state = "t25"
 	caliber = CALIBER_10x26_CASELESS //codex
 	max_shells = 50 //codex
-	force = 35
+	force = 25
 	aim_slowdown = 0.55
 	wield_delay = 0.6 SECONDS
 	fire_sound = "gun_smartgun"


Losenis:
From tivi's testing he didn't even notice the damage increase and wanted to merge a buff he didn't know was already merged. If 25 damage at 0.2 firerate is "too strong" then pretty much the t12 t18 and t42 are too strong too.

If you're pushing with the smartrifle to chase a xeno the iff is redundant and you're only holding a T12 with req-only ammo.

Hekzder:
> From tivi's testing he didn't even notice the damage increase and wanted to merge a buff he didn't know was already merged. If 25 damage at 0.2 firerate is "too strong" then pretty much the t12 t18 and t42 are too strong too.
> 
> If you're pushing with the smartrifle to chase a xeno the iff is redundant and you're only holding a T12 with req-only ammo.

A damage increase of 10 on a gun that does 15 damage is incredibly noticeable, that is a 66% increase in damage. I highly doubt this is the case. 

The issue boils down to having the T-25 do 25% more damage than the T-29, while also shooting faster, having less wield delay, and also having a greater choice of attachments is not really the best idea in terms of making them competitive. Putting them on the same level of damage boils the choice down to "Do I want the IFF gun with more versatility but less ammo and sunder or do I want the IFF gun that is slower but has more ammo and more sunder?" which is generally a choice of playstyle. 

As it stands now, at 25 damage, I think the smart rifle is just better than the T-29 entirely and the choices are not incredibly competitive.

Losenis:
Bruh the moment you don't have a marine in front of you it's just a t12 clone

Hekzder:
> Bruh the moment you don't have a marine in front of you it's just a t12 clone

The T-25 does 10 more pen and double the sunder of the T-12 and also has free IFF, so when you _do_ have Marines in front of you, you are using a T-12 on steroids basically. But even in your world of a hyper aggressive smartgunner that for some reason is at the frontline chasing, you are still using a better T-12. At 25 damage the T-25 is literally just the T-12 on steroids.

Losenis:
For a role-locked weapon tivi had to RNG 3 rounds to get on top of having to ERP RO to have a chance to get ammo for? Yeah, sounds about right.

Good call on the melee nerf though, that is stupid as hell since you can just push and 1shot purple sacs as you go so xenos can't even maul you on weeds. Hell go even further beyond and remove the ability to mount a bayonet, SGnrs have zero business being in melee.

David-Stormwell:
> 
> 
> Hell go even further beyond and remove the ability to mount a bayonet, SGnrs have zero business being in melee.
I disgree heavily and say keep it.
Otherwise, my personal suggestion would be to nerf it to simply 20 force.
The bullet damage nerf is good, as I myself didn't even agree with the buff.

Hekzder:
Not opposed to nerfing melee force further to 20 but I'll leave that up to maintainers. Though yes I do agree that you should still be able to put a bayonet on the SR regardless, the wider variety of attachments is a factor in what differentiates the SG from the SR.

BraveMole:
the buff was an overbuff, this is more fair

TiviPlus:
With the damage it does, the fact that iff doesnt matter in chasing, the fact you need to manually need to collect your magazines and beg the RO for a refill box all while competing for the reloads with other players hoarding them, and the fact that SG slots are limited means that I really think this buff isnt needed so ill be closing this

PR #7624, opened 2021-07-26
minigun buff and resprite

Pariah919:
## About The Pull Request
so the powerpack change basically unagreeably nerfed the minigun even further into the meme level tier it was in
this makes minigun fire at a rate of 0.15, gives it aim mode which gives you near turret level speed even with no armor, and a rpm of 4(5) shots per second in aimed mode

sprite is from ocelot, though i've reclored it 

image

## Why It's Good For The Game
Makes minigun more viable than being a meme fest.

## Changelog
:cl:
balance: minigun has been improved a bit, has 0.15 fire delay from 0.2, and has aim mode with a aim mode fire delay of 0.25, but you move VERY, VERY slowly while aiming.
/:cl:


lbnesquik:
I approve. I would rather have an expensive but strong minigun than a cheap and weak one. That's what the PP bison is for.

TiviPlus (approved):

PR #7631, opened 2021-07-26
Runtimes in gun system fix

BraveMole:
## About The Pull Request

One signal was not unregistered properly, and crashing while firing cause a runtime because the client is null, which can cause the gun to malfunction when reconnecting

## Why It's Good For The Game

No player facing changes

:cl:
fix: fixes some gun runtimes that would render the gun innoperable.
:cl:


SplinterGP (approved):

SplinterGP:
CL, it fixes a bug that causes gun to stop working.

PR #7632, opened 2021-07-26
Fix loadout deletion

BraveMole:
## About The Pull Request

Loadouts can be deleted

## Why It's Good For The Game

Seriously byond, why is it so hard to handle list

## Changelog
:cl:
fix: You can delete loadouts
/:cl:

### DIFF
--- a/code/datums/loadout/loadout_manager.dm
+++ b/code/datums/loadout/loadout_manager.dm
@@ -6,7 +6,7 @@
 	/**
 	 * List of all loadouts. Format is list(list(loadout_job, loadout_name))
 	 */
-	var/loadouts_data = list()
+	var/list/loadouts_data = list()
 	/// The host of the loadout_manager, aka from which loadout vendor are you managing loadouts
 	var/loadout_vendor
 	/// The version of the loadout manager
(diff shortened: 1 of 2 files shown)


code/datums/loadout/loadout_manager.dm
```diff
@@ -6,18 +6,20 @@
...
 /datum/loadout_manager/proc/delete_loadout(mob/user, loadout_name, loadout_job)
+	var/list/new_loadouts_data = list()
 	for(var/loadout_data in loadouts_data)
-		if(loadout_data[1] == loadout_job && loadout_data[2] == loadout_name)
-			loadouts_data -= loadout_data
-			return
+		if(loadout_data[1] != loadout_job || loadout_data[2] != loadout_name)
+			new_loadouts_data += list(loadout_data)
+	loadouts_data = new_loadouts_data
+	user.client?.prefs.save_loadout_list(loadouts_data, CURRENT_LOADOUT_VERSION)
```
TiviPlus, line 22: Just use cut()?

BraveMole: this is doing the same, but better, since it deletes duplicate loadouts rather than just the first one.

PR #7635, opened 2021-07-26
Speeds up diagonal movement 

BraveMole:
## About The Pull Request

Speeds up diagonal movement, from a *2 delay to only *1.6 compared to cardinal movements.

## Why It's Good For The Game

Right now, moving in diagonal is stricly a "qol" thing, because you moving north then west is taking the SAME time than moving north-west. 

This means than in almost all cases, moving diagonally is a bad choice : 
- You don't chose which path you take, since you will cross two tiles, and a player will be incapable to predict which tiles he will cross
- You stay on the same spot twice as long, so you are far easier to shoot. As a xeno, moving diagonally while under fire is a death sentence

In all these cases, moving cardinally is either better or at least equal to moving diagonally ; there are no cases where moving diagonally is a good tactical choice

With this PR, we give the player a new tool to master movement ; yes, you still move slower while moving diagonally and stays on the same tile longer, but on the other hand it's the shortest path to your objectives. This raises the skill ceiling, and give more opportunity to the player

Overall, there will still be plenty of situations where moving diagonally is a bad choice, but some situations will be optimal for it.

This also makes movement overall more pleasant. 

This is higher than sqr(2), because this probably allow to still bumpslash moving mobs without too much trouble. 1.42 is probably too fast and too much of a balance change, 1.6 is more reasonnable, at least for a start

## Changelog
:cl:
balance: Diagonal move delay reduced from 2 to 1.6
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -1,6 +1,9 @@
 //Some mob defines below
 #define AI_CAMERA_LUMINOSITY 6
 
+//Mob movement define
+#define DIAG_MOVEMENT_ADDED_DELAY 1.6
+
 
 //Pain or shock reduction for different reagents
 #define PAIN_REDUCTION_VERY_LIGHT -10 //alkysine
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -2,7 +2,7 @@
 #define AI_CAMERA_LUMINOSITY 6
 
 //Mob movement define
-#define DIAG_MOVEMENT_ADDED_DELAY 1.6
+#define DIAG_MOVEMENT_ADDED_DELAY_MULTIPLIER 1.6
 
 
 //Pain or shock reduction for different reagents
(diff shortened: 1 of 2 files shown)


Losenis:
Oh right I had also forgotten to mention, bullet speed had never been made to abide by the "diagonal is twice as long because you're moving two tiles" shenanigans and nobody ever minded it.

TiviPlus:
Provide more justification in your PR body please and elaborate on why this is better than the sqrt(2) PR please

code/modules/mob/mob_movement.dm
```diff
@@ -141,7 +141,7 @@
 	. = ..()
 
 	if((direct & (direct - 1)) && mob.loc == n) //moved diagonally successfully
-		add_delay *= 2
+		add_delay *= 1.6
```
TiviPlus, line 144: THis is a magic number

code/__DEFINES/mobs.dm
```diff
@@ -1,6 +1,9 @@
 //Some mob defines below
 #define AI_CAMERA_LUMINOSITY 6
 
+//Mob movement define
+#define DIAG_MOVEMENT_ADDED_DELAY 1.6
```
TiviPlus, line 5: This isnt a delay its a multiplier

TiviPlus (approved):

PR #7642, opened 2021-07-27
Returns a missing camera from the alamo, so i do not have to stare at a dirty, fuzzy line of three blocks.

lbnesquik:
## About The Pull Request
Fix good, Bugs bad

image## 

Changelog
:cl:
fix: fixed a missing camera on the alamo
/:cl:


lbnesquik:
I tested it and it works. Not sure why the map diff crashed.

PR #7649, opened 2021-07-28
Changes rifle bucket 2 to a M412 and a HK-11 and new smg bucket

Pariah919:
## About The Pull Request
Swaps rifle bucket 2 from an uzi to a M412 and a HK-11.
SMG bucket with MP27 and MR25

## Why It's Good For The Game
Makes Rifle bucket 2 more unique, there's already a free m412 and it's barely different from the normal AR, and the HK-11 is a pretty odd sidegrade to the LMG effectively with it's stats overall. And is locked to the UGL.

The MP-27 could use some testing as I think it's pretty bad overall. And the MR-25 is a SMG sidegrade, fires slower and has more in the magazine.

## Changelog
:cl:
expansion: seasonal gun bucket expanded-rifle bucket 2 has a M412 and HK-11 rather than just an Uzi. And added a new bucket with an MR-25 and the MP-27
/:cl:

### DIFF
--- a/code/controllers/subsystem/persistence.dm
+++ b/code/controllers/subsystem/persistence.dm
@@ -32,6 +32,7 @@ SUBSYSTEM_DEF(persistence)
 		/datum/season_datum/weapons/guns/rifle_seasonal_two,
 		/datum/season_datum/weapons/guns/pistol_seasonal_three,
 		/datum/season_datum/weapons/guns/copsandrobbers_seasonal,
+		/datum/season_datum/weapons/guns/smg_seasonal,
 		)
 	)
 	///The saved list of custom outfits names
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/persistence.dm
```diff
@@ -223,3 +226,13 @@ SUBSYSTEM_DEF(persistence)
...
 
+/datum/season_datum/weapons/guns/smg_seasonal
+	name = "SMGs"
+	description = "Two different SMGs. A classic and a new guy."
+	item_list = list(
+		/obj/item/weapon/gun/smg/m25 = -1,
+		/obj/item/ammo_magazine/smg/m25 = -1,
+		/obj/item/weapon/gun/smg/mp7 = -1,
+		/obj/item/ammo_magazine/smg/mp7 = -1,
+		)
```
SplinterGP, line 237: why not put a uzi here

SplinterGP (changes requested):

code/controllers/subsystem/persistence.dm
Pariah919: uzi is already in cops and robbers as one of the choices so i didn't see the need to readd it to another smg list
i'd like the mp27 and mr25 tested a bit

TiviPlus (approved):

PR #7661, opened 2021-07-29
Removes AMR NVGs

Pariah919:
## About The Pull Request
Removes AMR NVG.
The price is unchanged, even though the 60->75 was the reason it got NVGs, I think it's worth 75 even without the NVG.

## Why It's Good For The Game
AMR does not need NVGs, the only restricting factor it has ever had is the fact you need to be in light range to even use it, it required support and teamplay to use effectively.
Giving it NVGs removes any team aspect you needed to play with the rifle, M42A spec had NVGs because it was a different time with a much faster TTK overall and marines really badly needed covering fire to stop ravager man with 50 damage from gigafragging them instantly, but now that doesn't exist, so it's fine for the AMR to require teamplay to work at max range.
Price is unchanged.

## Changelog
:cl:
balance: AMR no longer has an NVG scope.
/:cl:


InterroLouis:
How about no.

lbnesquik:
I'd rather increase the cost if it is that broken

TiviPlus:
Makes it useless

PR #7692, opened 2021-07-30
Adds a new rifle type to marines-Skirmish Rifle

Pariah919:
## About The Pull Request
Adds the T-21 Skirmish rifle, filling the void between a battle rifle and an assault rifle.
It does 30/10/1.5, fires Heavy bullets and it's packet is 125 rounds
image

## Why It's Good For The Game
Adds a new class of rifle which bridges a gap we really don't have right now, and hopefully opens up new styles of play. 
It also looks and sounds great.

## Changelog
:cl:
add: Added new Skirmish Rifle (T-21) to marines, it bridges the gap between a Battle rifle and a Assault rifle. It is 0.2- 30/10/1.5. Aim mode stats are +0.15 and 2.5.
/:cl:


Kuro020 (approved):
the blandest gun in the west, but it's ok

David-Stormwell:
> the blandest gun in the west, but it's ok

go harder on him, we both know the tan and middle section which looks like it was ripped from a carbine don't look swell.

Kuro020:
> > the blandest gun in the west, but it's ok
> 
> go harder on him, we both know the tan and middle section which looks like it was ripped from a carbine don't look swell.

but the actual g3 is like that so uhhhhhhhhh

Tristan63:
totally NOT a g3a3

davidchan13:
Interesting idea but why does the step between the T-12 and T-64 have less mag capacity than both of them?

Hekzder:
I do agree with David that 25 rounds is incredibly low and doesn't make a lot of sense. Maybe 30 at least?

 I also feel with these kinds of stats it will just make the battle rifle even more obsolete? Though IMO that's more an issue of the battle rifle not being very good and less an issue of this being too power creep. If this does end up becoming a standard weapon I feel like a BR buff is even more overdue. Battle rifle will effectively fire almost half the speed while also being slower to handle for +2.5 damage, +5 penetration, +0.5 sunder, a somewhat larger magazine, and the bullets travel further. The bonuses you get for using the battle rifle, while there are a few, are incredibly small in nature and probably not worth it.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -500,6 +500,13 @@ datum/ammo/bullet/revolver/tp44
 	penetration = 10
 	sundering = 1
 
+/datum/ammo/bullet/rifle/heavy
+	name = "heavy rifle bullet"
+	hud_state = "hivelo"
+	damage = 30
+	penetration = 10
+	sundering = 1.5
+
```
SplinterGP, line 509: absurd stats for same firerate as the T-12

SplinterGP (changes requested):

Hekzder:
Should probably add a T-21 magazine box to the req operational vendor as well.

Kuro020 (changes requested):
image
mob sprites are awful, fix

Kuro020 (approved):

Hekzder:
Very cool, quite fun, and pretty well balanced as my first impressions. Seems like a pretty good addition.

Al-1ce:
This is good. Very good. I'm a x*no main but I've tested around with this and it seems pretty balanced to me. Fills the niche that the T-37 should've.

Tristan63:
> I do agree with David that 25 rounds is incredibly low and doesn't make a lot of sense. Maybe 30 at least?
> 
> I also feel with these kinds of stats it will just make the battle rifle even more obsolete? Though IMO that's more an issue of the battle rifle not being very good and less an issue of this being too power creep. If this does end up becoming a standard weapon I feel like a BR buff is even more overdue. Battle rifle will effectively fire almost half the speed while also being slower to handle for +2.5 damage, +5 penetration, +0.5 sunder, a somewhat larger magazine, and the bullets travel further. The bonuses you get for using the battle rifle, while there are a few, are incredibly small in nature and probably not worth it at that point.

battle rifle has always been obsolete

code/modules/projectiles/ammo_datums.dm
SplinterGP: i say still reduce this sunder to like 1 or 1.25

SplinterGP (changes requested):

David-Stormwell (approved):
Should be fine as is, feel free to enforce that sunder change though other things deal more sunder, so it really shouldn't matter.

lbnesquik:
I think we are done here. I'm up for a full merge.

AnCopper:
MJP is incapable of creating anything other than old war era guns.

PR #7703, opened 2021-07-31
Fix crusher charging

TiviPlus:
## About The Pull Request

Forgot about dual definitions
## Changelog
:cl:
fix: Fixed crusher charging
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -222,7 +222,6 @@
 	if(oldarea != newarea)
 		newarea.Entered(src, oldarea)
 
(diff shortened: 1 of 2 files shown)


code/game/atoms_movable.dm
```diff
@@ -222,7 +222,6 @@
 	if(oldarea != newarea)
 		newarea.Entered(src, oldarea)
 
-	Moved(oldloc, direction, FALSE, old_locs)
```
SplinterGP, line 225: It should not call moved?

TiviPlus: No its called in the override

PR #7491, opened 2021-07-18
LV624: LZ2 revamp plus minor cave adjustments

RipGrayson:
Revamps and moves LZ2 farther west, swapping its position with the metal cargo box area. Adds a small tunnel above LZ2 allowing for marines to directly enter southwestern caves. Adjusts a sign near hydro tunnel so it doesn't float in mid air.

LZ2 is widely considered to be the worst LZ by general opinion. Most LZ's are built around having 2-3 fixed points for sieging, LZ2 on the other hand can be sieged from almost anywhere in a 180 degree due to all of its walls being breakable with acid. Additionally, due to the setup of LZ2, only one exit of its three exits is viable for pushing, meaning that xenos can freely concentrate their forces and stifle any attempt at breakout. 
It's not uncommon for marine forces to never make out of FOB, let alone make it across the river.

By revamping and moving the LZ we can reduce those problems while adding additional interesting options to the marine forces.
Marines can now venture from LZ2 into southwest caves directly, at the cost of entering unfriendly terrain easily contested by hostile forces, or fight east into the colony to flank medical and hydro from an odd angle.
Multiple exits viable for pushing means that marines are less likely to get into a stalemate against hostile forces, while still offering many ways for xenos to siege LZ2 if that push fails.
Additionally, moving the LZ should reduce the amount of monotony over battling for the same positions multiple rounds in a row, since LZ1 and 2 offer entirely different routes and encourage different attacking directions.

## Changelog
:cl:
balance: LV624: LZ2 has been rebalanced and moved.
balance: LV624: Western cargo storage has been moved to LZ2's old spot.
fix: LV624: Fixed a disconnected engineering sign near hydro bridge.
/:cl:

PR #7704, opened 2021-07-31
Revert "LV624: LZ2 revamp plus minor cave adjustments (#7491)"

BraveMole:
This reverts commit ef5f36c18587ba876c36e4f9f044c16817d40612.

## About The Pull Request

Revert these LV changes

## Why It's Good For The Game

I don't know why this PR was merged, we just went back to square one, i'd say it's even worse now.

Bringing back the old lz placement just means a very VERY stale gameplay : fight on bridge, fight on bridge, fight on bridge, oh, fight in caves a little bit, fight on bridge...

The LZ in south had issues, but at least people chose the east LZ and were fighting a minimum in varied environments : Nexus, hydro, a bridge, engineering, and then caves.

With the current LV, you don't even fight one second in the colony. 80% of the map is ignored.

So i'm aware the south LZ was baddly design, but just fix it rather than doing THIS. 

LV

## Changelog
:cl:
balance: Bring back lv with the south lz
/:cl:

### DIFF
--- a/_maps/map_files/LV624/LV624.dmm
+++ b/_maps/map_files/LV624/LV624.dmm
@@ -2550,10 +2550,6 @@
 "aod" = (
 /turf/open/floor/plating/catwalk,
 /area/lv624/ground/river2)
-"aoe" = (
-/obj/structure/largecrate/goat,
-/turf/open/floor,
-/area/lv624/ground/southcargo)
 "aof" = (
 /obj/structure/window/framed/colony,
 /turf/open/floor/tile/vault{
@@ -2564,6 +2560,10 @@
 /obj/structure/disposalpipe/segment,
 /obj/structure/window/framed/colony,
 /turf/open/floor/plating/warning,
+/area/lv624/ground/river2)
+"aoh" = (
+/obj/structure/jungle/planttop1,
+/turf/closed/wall,
 /area/lv624/ground/river2)
 "aoi" = (
 /turf/open/floor/plating/warning,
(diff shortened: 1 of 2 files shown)


EvilMelonTheSecond:
Instead of reverting make a proper fix. I hated being forced to shitty LZ1 with both flanks being meltable + maze right outside its doors for xeno to constantly ambush reinforcements. I consistently saw marines either fully push out of the maze and then tad/pod to frontlies OR get suppressed to FOB forever. I even tested how bad LZ1 is by melting the flanks as a boiler and taking over FOB solo multiple times, marines don't want to be near nexus as its a death maze with no objective inside. The south LZ would just never hold against a siege and that was the reason LZ1 started getting picked.

And statics wise, if everybody hates bridge-cave combat yet it gets chosen consistently then doesn't that mean the other option is even worse?

RipGrayson:
I really wish this kind of feedback was given during the testmerge period, there wasn't a single peep of LZ2 being overpowered and bypassing the colony during the days it was TM'ed. 

All that being said, if LZ2 _is_ becoming the meta choice it bears investigating as to why. Is it because LZ2 is so superior or LZ1 is just considered so much worse (as EvilMelonTheSecond says)?

I'm not sure an entire revert is needed, forcing marines to use a (potentially) bad LZ1 by making the alternative garbage again doesn't seem like a good idea. Either LZ1 could be improved or LZ2 could be adjusted to make the marines traverse more of the colony before entering caves without going into full-blown revert mode. I'll have a second PR waiting in the wings if there's a consensus that LZ2 really is the problem here, but I would really like to hear more discussion over the current state of LZ2 first.

TiviPlus:
You either need to fix the previous LZ's issues or fix this LZ's issues not kneejerk revert it a day after it gets merged

PR #7719, opened 2021-08-01
Fixes beer

QualityVan:
## About The Pull Request
Reagent dispensers use an associated list of names to reagent datums to determine what to dispense, but this breaks when multiple reagents have the same name. Regular beer and soporific beer, in this case, are both just 'beer'. This changes the name2reagent list to key to a list of reagents with that name, which the dispenser can iterate through to find the right one.
Also fixes an index access error with hacking reagent dispensers. Turns out byond lists don't start at 0.

## Why It's Good For The Game
Fixes

## Changelog
:cl:
fix: Reagent dispensers can produce beer properly now.
fix: Hacking reagent dispensers with a multitool actually works.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents.dm
+++ b/code/modules/reagents/reagents.dm
@@ -5,7 +5,7 @@ GLOBAL_LIST_INIT(name2reagent, build_name2reagent())
 	for (var/t in subtypesof(/datum/reagent))
 		var/datum/reagent/R = t
 		if (length(initial(R.name)))
-			.[ckey(initial(R.name))] = t
+			LAZYADD(.[ckey(initial(R.name))], t)
 
 /// A single reagent
 /datum/reagent
--- a/code/modules/reagents/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/machinery/chem_dispenser.dm
@@ -111,10 +111,10 @@
 	// I dont care about the type of tool, if it triggers multitool act its good enough.
 	hackedcheck = !hackedcheck
 	if(hackedcheck)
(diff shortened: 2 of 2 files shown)


code/modules/reagents/reagents.dm
```diff
@@ -5,7 +5,7 @@ GLOBAL_LIST_INIT(name2reagent, build_name2reagent())
 	for (var/t in subtypesof(/datum/reagent))
 		var/datum/reagent/R = t
 		if (length(initial(R.name)))
-			.[ckey(initial(R.name))] = t
+			LAZYADD(.[ckey(initial(R.name))], t)
```
TiviPlus, line 8: LAZYADD or LAZYSET?

TiviPlus (changes requested):

code/modules/reagents/reagents.dm
QualityVan: LAZYADD
. is an associated list, mapping the reagent name (the ckey bit) to a list of reagent datums. The datum list itself is just indexed.

SplinterGP: its lazyset, lazyadd is adding, before its easy to see it was just setting . not adding anything to it.

QualityVan:
turns out this breaks recipes oops

code/modules/reagents/reagents.dm
TiviPlus: Dont self resolve in future

QualityVan:
Closing until I come up with a better solution.

PR #7747, opened 2021-08-03
Fixes the UGL not having a fire delay by giving it a fire delay of 1 second

Hekzder:
## About The Pull Request

So, the underbarrel attachments have had an `attachment_firing_delay` for the longest time but the underbarrel flamer is the only attachment that actually properly uses it. Both the masterkey and the UGL currently have _no_ fire delay what so ever and can just be spam click rapid fired. 

This PR specifically fixes the UGL not having a fire delay and also balances it by giving it a fire delay of 1 second, the masterkey remains untouched for now because it is frankly shit code that I don't want to touch and don't fully understand, and also no one uses it. 1 second may seem high but is still faster than the rotary GL at 1.2 seconds of fire delay, the balance part of this PR is mainly meant to combat impacts being insanely deadly in a UGL due to being able to insta spam two of them and immediately unload with your primary. It's incredibly unfun and incredibly hard to counter or predict/prepare for. 

Even in a world without impacts, the UGL should have a fire delay regardless, and it's obviously intended for it to, just no one made it work properly. This PR also fixes a small typo that lists the UGL as being a single shot launcher. 

## Why It's Good For The Game

Being able to rapid spam two grenades back to back from an underslung launcher while in the middle of firing and being able to fire right after is insanely strong. This brings the UGL a bit more down to earth with the recent re-addition of impacts while also fixing something it was intended to always have presumably, fire delay. 

Anyone who says this completely guts the UGL fails to realize the sheer versatility and power of having short fuse grenades on standby without having to switch hands or weapons, even if you can't rapid fire spam them to instant kill a Xeno. 

## Changelog
:cl:
balance: The UGL now has a fire delay of 1 second
fix: Fixed the UGL not having a fire delay at all
spellcheck: Fixed the UGL description listing it as a single shot launcher
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1452,9 +1452,10 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	var/distance = 0
 	var/turf/prev_T
 	playsound(user, 'sound/weapons/guns/fire/flamethrower2.ogg', 50, 1)
+	var/fire_delay = attachment_firing_delay
 	if(!user.skills.getRating("firearms")) //no training in any firearms
-		attachment_firing_delay += 0.3 SECONDS //untrained humans fire more slowly.
-	COOLDOWN_START(src, last_fired, attachment_firing_delay)
+		fire_delay += 0.3 SECONDS //untrained humans fire more slowly.
+	COOLDOWN_START(src, last_fired, fire_delay)
 	for(var/turf/T in turfs)
 		if(T == user.loc)
 			prev_T = T
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1263,7 +1263,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 //The requirement for an attachable being alt fire is AMMO CAPACITY > 0.
 /obj/item/attachable/attached_gun/grenade
 	name = "underslung grenade launcher"
-	desc = "A weapon-mounted, reloadable, one-shot grenade launcher."
+	desc = "A weapon-mounted, reloadable, two-shot grenade launcher."
 	icon_state = "grenade"
 	attach_icon = "grenade_a"
 	w_class = WEIGHT_CLASS_BULKY
@@ -1275,7 +1275,8 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	flags_attach_features = ATTACH_REMOVABLE|ATTACH_ACTIVATION|ATTACH_RELOADABLE|ATTACH_WEAPON
 	///list of grenade types loaded in the UGL
 	var/list/loaded_grenades = list()
-	attachment_firing_delay = 21
+	attachment_firing_delay = 10
+	COOLDOWN_DECLARE(last_fired)
 
 /obj/item/attachable/attached_gun/grenade/unremovable
 	flags_attach_features = ATTACH_ACTIVATION|ATTACH_RELOADABLE|ATTACH_WEAPON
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/gun_attachables.dm
```diff
@@ -1322,14 +1323,16 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
...
+	if(current_rounds > 0 && COOLDOWN_CHECK(src, last_fired))
 		prime_grenade(target,gun,user)
 
 
 /obj/item/attachable/attached_gun/grenade/proc/prime_grenade(atom/target, obj/item/weapon/gun/gun, mob/living/user)
 	set waitfor = FALSE
 	var/nade_type = loaded_grenades[1]
 	var/obj/item/explosive/grenade/frag/G = new nade_type (get_turf(gun))
+	var/fire_delay = attachment_firing_delay
+	COOLDOWN_START(src, last_fired, fire_delay)
```
TiviPlus, line 1335: Why is this a var

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1275,7 +1275,8 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	flags_attach_features = ATTACH_REMOVABLE|ATTACH_ACTIVATION|ATTACH_RELOADABLE|ATTACH_WEAPON
 	///list of grenade types loaded in the UGL
 	var/list/loaded_grenades = list()
-	attachment_firing_delay = 21
+	attachment_firing_delay = 10
+	COOLDOWN_DECLARE(last_fired)
```
TiviPlus, line 1279: declaring new vars comes under redefining new vars, move attachment_firing_delay

TiviPlus (changes requested):

code/modules/projectiles/gun_attachables.dm
Hekzder: This is an exact copy + paste of underbarrel flamer code. If you want me to change this then I'll change under barrel flamer code as well.

code/modules/projectiles/gun_attachables.dm
Hekzder: This is an exact copy + paste of underbarrel flamer code. If you want me to change this then I'll change under barrel flamer code as well.

code/modules/projectiles/gun_attachables.dm
Hekzder: Should be fixed now.

code/modules/projectiles/gun_attachables.dm
Hekzder: Should be fixed now.

code/modules/projectiles/gun_attachables.dm
```diff
@@ -1450,10 +1452,9 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	var/distance = 0
 	var/turf/prev_T
 	playsound(user, 'sound/weapons/guns/fire/flamethrower2.ogg', 50, 1)
-	var/fire_delay = attachment_firing_delay
 	if(!user.skills.getRating("firearms")) //no training in any firearms
-		fire_delay += 0.3 SECONDS //untrained humans fire more slowly.
-	COOLDOWN_START(src, last_fired, fire_delay)
```
TiviPlus, line 1456: This breaks firing delay I have no idea why you changed this

TiviPlus (changes requested):

code/modules/projectiles/gun_attachables.dm
Hekzder: Because you told me to not define a fire delay var and just use attachment_firing_delay. It works perfectly fine in testing for both UGL and UBF and I'm just doing what you asked me to do.

Hekzder: Gone, reduced to atoms.

TiviPlus (approved):

BraveMole:
well, at least i was able to abuse for two days

Captiajoe:
Salt PR, smh the under barrel only has two nades capacity

PR #7760, opened 2021-08-04
tp-44 snub damage buff

Pariah919:
## About The Pull Request
snub revolvers are trash
like literal trash
the only thing they have over a tp-23 or tp14 is that you therotically deal more damage in a short period of time, but considering you're going to be using them as a sidearm they're still worse than both of those, due to terrible ammo capacity.
35 damage gives them +5 raw damage over the tp-23, atleast.

## Why It's Good For The Game
Makes TP-44 snubs way less trash, though personally i think we should just split revolver into three guns (snub, revolver and magnum) and get rid of this entire snub thing, it clearly isn't working out.

## Changelog
:cl:
balance: snub revolvers do five more damage
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1790,7 +1790,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	icon_state = "tp44_barrel"
 	attach_icon = "tp44_barrel"
 	slot = ATTACHMENT_BARREL_MOD
-	damage_mod = 0.25
+	damage_mod = 0.125
 	scatter_mod = -2.5
 	recoil_unwielded_mod = 0.25
 	damage_falloff_mod = -0.5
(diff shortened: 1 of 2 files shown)


SilencedMP5A5:
i don't like this

how about instead you simply do the change i suggested originally

0.) revert to old revolver
1.) remove stagger screenshake and knockback
2.) congrats, revolver is now on par with every other sidearm

besides, backpack man no good, so i can't even do my meme unless i pester req for combat pack

KornFlaks:
It's been months since the days of mass screenshake stagger revolvers of doom. I dont even know why a "free" pistol even had stagger and screenshake. The mateba of course is a captain unique and req only weapon with limited ammo. Why the common variant also had the same effect is so strange in retrospect.

I agree with Ocelot, stub nose variant makes no sense. Remove it and return to a single revolver type. Returning to being able to fit in bags with the amount of DPS is a bit iffy, maybe keep the restriction of only fitting in revolver belts or make the size so large (like HSG size) to only fit 5 in a backpack.

Pariah919:
if i revert to old revolver i'm adding the limit of one per bag

PR #7792, opened 2021-08-05
evasion now actually has a real plasma cost, runner plasma reserves at lower tiers increased, and regen increased

Pariah919:
## About The Pull Request
runner literally does not use plasma at all, you could chain a near infinite amount of evades, which is pretty nuts. Now you can only chain a few.

## Why It's Good For The Game
Plasma should be a real thing that exists, and it works well with savages tradeoff of plasma.

## Changelog
:cl:
balance: evade now costs 75 plasma to use, runner has higher plasma reserves at lower tiers of maturity and regen overall, same plasma at ancient.
/:cl:


TiviPlus (approved):

Surrealaser:
Absolutely ridiculous.

What stops Evasion from infinitely chaining is that it becomes about impossibly difficult after the 1st chain and virtually difficult after the second.

Now even one chain totally drains all of your plasma, killing chaining on a defacto basis.

PR #7806, opened 2021-08-05
king takes 90 minutes to enter the round

Pariah919:
## About The Pull Request
60 to 90, effectively considering most drops are at 12:20/30, this is 60/70 mintutes.

## Why It's Good For The Game
King comes in way too fast into the game in general

## Changelog
:cl:
balance: king is locked for 90 minutes rather than an hour
/:cl:

PR #7807, opened 2021-08-05
Grav crush does less damage to mobs

BraveMole:
## About The Pull Request

Grav crush will still be strong against cade and such, less against marine

## Why It's Good For The Game

King should be good at one thing : kill the fob. Xeno should not use it while defending, they should prefer silos and turrets. This makes king a more specialsed caste, rather than what it is currently

Alternative to : https://github.com/tgstation/TerraGov-Marine-Corps/pull/7806 
which just give you an arbitrary length which a round should last. This is an half ass solution, if xeno wiped out marines they should be able to finish the fight now ; not 30 minutes later because of some stupid time limit

## Changelog
:cl:
balance: Grav crush does less damage to mobs
/:cl:


Tristan63:
I think this is actually a good change, since marines already being wounded can be sent straight to crit by a grav crush.

TiviPlus (approved):

PR #7811, opened 2021-08-06
Fixes the mortar flare shell not working 

lbnesquik:
## About The Pull Request
- Also give some smoke and flare shells to the mortar as they are pretty fun
- 
## Why It's Good For The Game
The flare shell was broken, this fixes it. No point having special code when they do basically the same thing.
Also, smoke & flare shells give some different options.

## Changelog
:cl:
expansion: Added some flares and smoke shells to the mortar kit.
fix: fixed the mortar flare shell not working by just using the code for flare rockets
/:cl:


TiviPlus (changes requested):
Using the same object it is bad for if you want to change balance, an example of this is current railgun

lbnesquik:
> 
> 
> Using the same object it is bad for if you want to change balance, an example of this is current railgun

Implemented

TiviPlus (approved):

PR #7814, opened 2021-08-06
Fixes vendors duping reagent containers and stacks

QualityVan:
## About The Pull Request
Prevents restocking vending machines with partial stacks and reagent containers(bottles, autoinjectors, etc) that don't have their initial reagents.

## Why It's Good For The Game
Partial restocking allows infinite duping. Anything that should be infinite should be made infinite through the vendor supply list, rather than through bug abuse.

## Changelog
:cl:
fix: Reagent containers which don't have their starting reagents can't be used to restock vending machines
fix: Stacks need to be at their starting stack count to restock vendors.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/reagent_container.dm
+++ b/code/game/objects/items/reagent_containers/reagent_container.dm
@@ -68,3 +68,10 @@
 			. += "; [R.name]([R.volume]u)"
 	else
 		. = "No reagents"
+
+///Returns TRUE if the current reagents contain at least everything in the list_reagents, FALSE otherwise
+/obj/item/reagent_containers/proc/has_initial_reagents()
+	for(var/reagent_to_check in list_reagents)
+		if(reagents.get_reagent_amount(reagent_to_check) != list_reagents[reagent_to_check])
+			return FALSE
+	return TRUE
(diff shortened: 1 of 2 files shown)


EvilMelonTheSecond:
Correct me if I'm wrong but wasn't this intentional?

Knuff-les:
There have literally been bugfixes in the past to fix this when infinite vendors were first introduced, breaking restocking. How is this a bug?

QualityVan:
> 
> 
> There have literally been bugfixes in the past to fix this when infinite vendors were first introduced, breaking restocking. How is this a bug?

Could you point them out? I checked back and the only one I saw was #6882 , which's unrelated to this.

jroinc:
It WOULD let us add some advanced stuff into the vendors...

Losenis:
The duping made chem and items bearable when you were latejoining to all items gone by 40 marines and the chem dispenser energy was at zero. You can't rely on people not being wasteful with limited stuff, and honestly I'd probably speedrun chem if I was under threat of arriving to things having run out. Pretty sure other people would do the same aggravating the issue.

TiviPlus:
This should be dependent on the items

The current configuration is fine in that it encourages you to save them while also stopping you from being kneecapped when you run out

However JPR does have a point
I would be ok if it was possible to decide which are restockable so feel free to make a pr for that, but I'm not interested in this pr as is

PR #7816, opened 2021-08-06
Adds medical tape, a worse version of splints

QualityVan:
## About The Pull Request
Basically the equivalent of gauze/ointment compared to advanced kits. Easy to use, no penalty for low medical skill, but not as good. Will fall off in 5 minutes.
Sprite provided by kuro, thanks.
Also makes regular splints slightly sturdier when applied without any skill, but take longer to use. Trained use time and sturdiness aren't changed.

## Why It's Good For The Game
Good to have worse but low-skill medical alternatives.
I'd also like to make these infinite from vendors later, along with gauze and ointment.

## Changelog
:cl:
add: Medical vendors now have tape, which can be used as a worse but untrained alternative to splints.
balance: Splints applied by untrained people are sturdier, but take a little longer to use.
/:cl:


code/game/objects/items/stacks/medical.dm
```diff
@@ -298,13 +303,11 @@
 			to_chat(user, span_warning("You can't apply a splint there!"))
 			return
 
-		if(affecting.limb_status & LIMB_DESTROYED)
-			to_chat(user, span_warning("[user == M ? "You don't" : "[M] doesn't"] have \a [limb]!"))
-			return
-
-		if(affecting.limb_status & LIMB_SPLINTED)
-			to_chat(user, span_warning("[user == M ? "Your" : "[M]'s"] [limb] is already splinted!"))
-			return
```
QualityVan, line 307: These checks aren't necessary because the limb itself handles them in its apply_splints

Losenis:
>A worse version of something that already turns a tyr2 into a 1hit kill by organ damage
I'd sooner retexture normal splints with this and make splints only fall off when they're in bonefrac territory, but lock it to corpsmen skill and up for usage.

code/game/objects/items/stacks/medical.dm
BraveMole: does that mean you have the windup then the warning?

BraveMole (changes requested):

code/game/objects/items/stacks/medical.dm
QualityVan: You'll have the fumble delay first, then the cancel out if the limb is destroyed/splinted, and then the splinting do_after.

BraveMole: so keeping the checks would be better no? why would you fumble to check if a splint is still here or destroy?

QualityVan: Gross incompetence. You can also avoid this by looking at the person, which tells you if any limbs are splinted or missing.

Kuro020 (commented):
image
:mistake: this sprite is just no

Kuro020 (changes requested):
see above ^

Kuro020:
tape
use this one

Kuro020 (approved):

lbnesquik:
I don't like the idea of the thing falling off. At least, not without some really clear indication that this thing is on a timer. 
Maybe large font text?
You could claim it is a skill issue if someone doesn't notice, but if someone else patches you up, and you don't check, your fracture comes right back.

QualityVan:
> 
> 
> I don't like the idea of the thing falling off. At least, not without some really clear indication that this thing is on a timer.
> Maybe large font text?
> You could claim it is a skill issue if someone doesn't notice, but if someone else patches you up, and you don't check, your fracture comes right back.
you know what you right, I'll add a warning a couple minutes before it falls off

SplinterGP:
could maybe instead of falling off or something just give a bit more debuffs than the splints do?

QualityVan:
> 
> 
> could maybe instead of falling off or something just give a bit more debuffs than the splints do?

I've thought about this a bit, and splints really don't have any debuffs as is. There's being easier to break than an unbroken limb, but tape is already worse there. Making tape have finite duration also plays into the idea that splinting is inherently a temporary measure, needing either reapplication later or more permanent treatment.

QualityVan:
Probably too strong as is and also just not interesting.

PR #7829, opened 2021-08-06
Bump ci version

BraveMole:
## About The Pull Request

Bump ci to the version we are compiling the server on.

Fix xeno_logical spawning that will no longer work. It should have never worked, initial(a.vars[b]) doesn't work when a is a typepath. It was working because the typepath was in a list, and the compiler failed to understand it was a typepath

image

image

image

image

image

this last image shows the old behaviour. This no longer works, it was ""fixed""

## Why It's Good For The Game

We should check the code on the version we are on

No player facing changes

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
@@ -146,7 +146,7 @@
 	ancient_message = "We are the master of ranged artillery. Let's bring death from above."
 
 	// *** Melee Attacks *** //
-	melee_damage = 12
+	melee_damage = 20
 
 	// *** Speed *** //
 	speed = 0.1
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
```diff
@@ -146,7 +146,7 @@
 	ancient_message = "We are the master of ranged artillery. Let's bring death from above."
 
 	// *** Melee Attacks *** //
-	melee_damage = 20
+	melee_damage = 12
```
BraveMole, line 149:
```suggestion
	melee_damage = 20
```

code/modules/unit_tests/xeno_logical_scaling.dm
```diff
@@ -28,35 +30,16 @@
...
-			current_value = new_value
-
-			for(var/upgrade in list(XENO_UPGRADE_ONE, XENO_UPGRADE_TWO, XENO_UPGRADE_THREE))
-				new_value = initial(mob_data[upgrade].vars[stat])
-				if(new_value > current_value)
-					Fail("Invalid stats on [xenopath]. It's [stat]@[upgrade] has [new_value] compared to previous [current_value] (expected lower)")
-				current_value = new_value
+		for(var/upgradepath in mob_data)
+			var/typepath = mob_data[upgradepath]
+			caste = new typepath
```
TiviPlus, line 35: allocate()

BraveMole: its a datum, not a mob

TiviPlus (approved):

PR #7840, opened 2021-08-07
balances sentinel missing tier upgrades.

vilereaver41:
## About The Pull Request

Sentinel had no upgrades to its speed which stayed at -0.8 from young to ancient & no upgrades in spit delay which stayed at 1.3 from young to ancient, this pr fixes this.

## Why It's Good For The Game

Because each caste needs this balance and deserves a fair upgrade in their values whether they are T1 or T3. this will make sentinel more viable especially when in most situations T2 slots are full.

## Changelog
:cl:
balance: sentinel speed from 0.8 from young to ancient now gives just a -1 for ancient
balance: sentinel spit delay from 1.3 young to ancient now    1.3 young 1.2 mature 1.1 elder and 1 ancient
balance: sentinel shouldn't need 300 hp, now 275 to make it easier to kill due to its annoyance.
balance: changed plasma gain
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/sentinel/castedatum_sentinel.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/sentinel/castedatum_sentinel.dm
@@ -117,7 +117,7 @@
 	melee_damage = 17
 
 	// *** Speed *** //
-	speed = 0.9
+	speed = -0.9
 
 	// *** Plasma *** //
 	plasma_max = 600
--- a/code/modules/mob/living/carbon/xenomorph/castes/sentinel/castedatum_sentinel.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/sentinel/castedatum_sentinel.dm
@@ -88,7 +88,7 @@
 	melee_damage = 17
 
 	// *** Speed *** //
-	speed = -0.9
+	speed = -0.8
 
 	// *** Plasma *** //
 	plasma_max = 550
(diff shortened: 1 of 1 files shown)


lKiseki:
A bad change. Sentinels can already easily run away from most unfavorable engagements, not to mention they have their neurotoxin spit. They're arguably one of the most annoying castes to deal with, and a speed increase would only make it worse.

Hekzder:
> A bad change. Sentinels can already easily run away from most unfavorable engagements, not to mention they have their neurotoxin spit. They're arguably one of the most annoying castes to deal with, and a speed increase would only make it worse.

A bad take. I have never once in my life thought "What is the most annoying caste to deal with?" and had Sentinel in the top five. I do think Sentinel is easily the worst T1 right now, though not by an insane margin. And this PR just adds things that Sentinel should have already had, while it is technically a balance PR it's also like 25% fixing an over sight.

lKiseki:
> A bad take. I have never once in my life thought "What is the most annoying caste to deal with?" and had Sentinel in the top five. I do think Sentinel is easily the worst T1 right now, though not by an insane margin. And this PR just adds things that Sentinel should have already had, while it is technically a balance PR it's also like 25% fixing an over sight.

Right now Sentinels are able to run away from most encounters, and their gameplay revolves around hit and run. As a Marine, if you've ever bothered chasing a Sentinel, no matter how fast you are it can very easily get away -- _off weeds_. It can heal up then continue doing the same. This PR will only end up accentuating the issue. Oversight or not, I'd rather Sentinels don't get increased speed.

TiviPlus:
closing as stale

PR #7847, opened 2021-08-07
Wraith stats no longer weaker then a T1 drone.

vilereaver41:
## About The Pull Request

Wraith's stats were worse than a T1's this fixes that.

## Why It's Good For The Game

Considering that shipside wraith is no longer a thing it means wraith is meant to be more combat-focused, this makes its stat values more meaningful for a T2 it previously had less health than a drone and very bad armor values that weren't bullet.

## Changelog
:cl:
Balance: wraiths health is no longer less than a T1
Balance: Increases Wraiths laser and energy resistance to something that makes more sense instead of underwhelming values
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/wraith/castedatum_wraith.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/wraith/castedatum_wraith.dm
@@ -35,7 +35,7 @@
 	caste_flags = CASTE_CAN_BE_QUEEN_HEALED|CASTE_EVOLUTION_ALLOWED|CASTE_CAN_VENT_CRAWL|CASTE_CAN_BE_GIVEN_PLASMA|CASTE_CAN_BE_LEADER|CAN_BECOME_KING
 
 	// *** Defense *** //
-	soft_armor = list("melee" = 25, "bullet" = 25, "laser" = 5, "energy" = 5, "bomb" = XENO_BOMB_RESIST_0, "bio" = 10, "rad" = 15, "fire" = 15, "acid" = 10)
+	soft_armor = list("melee" = 25, "bullet" = 25, "laser" = 15, "energy" = 15, "bomb" = XENO_BOMB_RESIST_0, "bio" = 10, "rad" = 15, "fire" = 15, "acid" = 10)
 
 
 	// *** Abilities *** //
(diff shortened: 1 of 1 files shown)


EvilMelonTheSecond:
I thought the whole purpose of wraith was to deny any cadeline other than FOB? 

A fast and teleporting caste should be kept fragile, it isn't a combat caste and shipside wraith was just a meme not something anything was balanced around.

vilereaver41:
> I thought the whole purpose of wraith was to deny any cadeline other than FOB?
> 
> A fast and teleporting caste should be kept fragile, it isn't a combat caste and shipside wraith was just a meme not something anything was balanced around.

You thought wrong.

Hekzder:
Not really a crazy buff, maybe Ancient health could be slightly toned down a bit to like 300, but otherwise fairly niche when it comes to the rest of it, mainly being the laser resist.

lbnesquik:
Bit too much at ancient imo

code/modules/mob/living/carbon/xenomorph/castes/wraith/castedatum_wraith.dm
```diff
@@ -123,11 +123,11 @@
 	plasma_gain = 25
 
 	// *** Health *** //
-	max_health = 260
+	max_health = 310
```
Pariah919, line 126:
```suggestion
	max_health = 310
```
a 50 point increase to a caste with decent armor stats and incredible mobility outisde of being stagger seems kinda high, i'd go with capping out at 290

vilereaver41: 50 Point increase is because its a T2 the health is normal for a T2. It's non bullet armor is dogshit so dno where you getting decent armor stats from.

lKiseki:
I'll be honest with you, I'm not quite sure these changes are needed. Wraiths can already easily escape most engagements thanks to Blink, and if they pick their fights well & they're smart about their plasma management and ability usage, it's practically impossible to catch a Wraith. This is a step in the wrong direction.

As a tier 2, Wraith already has plenty in its toolkit.

vilereaver41:
> I'll be honest with you, I'm not quite sure these changes are needed. Wraiths can already easily escape most engagements thanks to Blink, and if they pick their fights well & they're smart about their plasma management and ability usage, it's practically impossible to catch a Wraith. This is a step in the wrong direction.
> 
> As a tier 2, Wraith already has plenty in its toolkit.

it's not about making it strong in 1v1 combat, the intention is to make its health T2 level, in group combat, wraith is used for warp field stacking the marines meaning it has to blink right into the fray, one wrong move and you die. its laser resistance and energy resistance were too low compared to bullet resistance. wraith is my main, I prob play it more than anyone else on the server i know it's ins and outs, its not about picking your battles so you can barely have a chance it is about making the playing field fair to begin with.

TiviPlus:
Wraith is so weak because its annoying to deal with once it cant escape instantly from any fight and cant get into FOB, ship etc areas we can discuss a buff to it

PR #7773, opened 2021-08-04
Reduces ravager slash damage

lbnesquik:
This is a revival of https://github.com/tgstation/TerraGov-Marine-Corps/pull/7764
I think it was closed too fast and the feedback was apparently some guy yelling in the author DM's.
While i don't really have a horse in this race, i think it is worth having it around for discussion.

.
.
.
## About The Pull Request
As reference, here are the current damage tiers of other similar castes:

Caste: Young / Mature damage - Elder / Ancient damage

Old Ravager: 26 - 32
New Ravager: 25 - 28

Defiler: 23 - 26
Hunter: 21 - 24
King / Queen / Shrike / Prae / Warrior: 20 - 23
Boiler / Wraith / Spitter: 17 - 20

Outliers from this damage increase trend:
Defender: 17 - 21
Bull: 19 - 21

## Why It's Good For The Game

Notice how the current "old" implementation of ravager conducts an increase in melee damage from 26 to 32, a total of 6 points compared to the nearly standardized 3 point increase. This drastic jump in damage by two levels propels it from merely high damaging to far and beyond its contemporaries in the hive and thus making it truly extraordinary in marine slaughtering capacity.

This damage is combined with a button that allows ravagers to increase their effective hitpoints by 33% instantly and a ravage skill that allows for an additional slash and short knockdown that opens up the possibility for even more extremely high damage slashes.

I propose a reduction in base melee damage to return the ravager's attack back to earth from the stratospheric levels reminiscent of bygone era of TGMC where Marines and Xenos conducted a race to the bottom in terms of armor and damage.

This does not affect the other tools available to a ravager. Extremely skilled ravagers may still employ the various combinations in skills and abilities provided in the various ravager mechanic additions that allow them to remain terrors of marine forces. This simply prevents a ravager that exists from round start and thus graduate to elder upon shutters up from dominating marines with little to no skill. Ravager players must now utilize health expansion techniques through endure and rage to deliver the same damage output as current base slashing produces.
Changelog

## Changelog
:cl:
balance: Reduces ravager damage tiers from 26 - 32 to 26 - 29.
/:cl:

PR #7856, opened 2021-08-08
Tweaks ravager endure duration and cooldown

lovegreenstuff:
## About The Pull Request

Lowers endure duration, decreases cooldown somewhat.

## Why It's Good For The Game

Ravs can use endure at the start of a fight and can end with enough time left to escape while still doing sizable damage.
Doing this forces ravs to be smarter about their endure timing. Or just die, that is also a possibility.

## Changelog
:cl:
balance: decreased endure duration, with lowered cooldown to somewhat compensate it.
/:cl:


lovegreenstuff:
very much a salt pr, by the way.

Passe2142:
Yeah, we seen this as the invis button , stats wise is fine cause its meant to be a quick frontline slasher. And ignore pain is no more which iirc was better? but had no visual queue and lacked the stagger protection.

Hekzder:
I think this is a step in the right direction but I would suggest either a further decrease in cooldown in the realm of 40-30 seconds _OR_ less of a decrease in duration in the realm of 6-7.5 seconds. As is right now I feel this is far too strong of a nerf, especially potentially factoring in a damage decrease from #7773.

Pariah919:
remember the cd starts instantly, so this is effectively 40 seconds
overall I'd like to see the damage decrease brew for a day or three before we take a look at endure again, but this is a step in the right direction at making it not literally invinciblity for an entire engagement, though rage might need a rework

Surrealaser:
Ah yes, so, we have a big melee damage nerf in the works (which I'm certain will pass seeing as Evasion's plasma cost was ramped to ridiculous levels that de facto kills chaining which was already challenging), Endure is now only good for covering an escape, and tapping into negative HP rage has gone from very difficult to borderline impossible.

Surrealaser:
> remember the cd starts instantly, so this is effectively 40 seconds
> overall I'd like to see the damage decrease brew for a day or three before we take a look at endure again, but this is a step in the right direction at making it not literally invinciblity for an entire engagement, though rage might need a rework

Ridiculous. Literally invincibility? Lol. Absurd hyperbole. It's literally +50 crit HP, and you don't KO on crit.

Passe2142:
> as E

Stats wise ravey is fine, i think this one is the main thingy to do, the other PR is a bit bananas.

Passe2142:
> 
> 
> remember the cd starts instantly, so this is effectively 40 seconds
> overall I'd like to see the damage decrease brew for a day or three before we take a look at endure again, but this is a step in the right direction at making it not literally invinciblity for an entire engagement, though rage might need a rework

Rage is fine it was nice rework and a much needed one. Still niche and high risk high reward. Stats wise Ravs are fine its supposed to be the main slasher.

usnpeepoo:
honestly' i'd rather see an endurenerf than a slash damage nerf.

lovegreenstuff:
#7773 
If this PR gets merged, I'm closin' this one.

lovegreenstuff:
I'm a moron, wrong button.

lovegreenstuff:
that's weird, it worked locally with no problems.

lovegreenstuff:
cheese burger burger town burger
number 15 burger ki-

PR #7861, opened 2021-08-08
water buff

Pariah919:
## About The Pull Request
has infinte cycles after 6 to 10, less stamina loss recvoery (-0.15 from 0.5) but same limb health

## Why It's Good For The Game
makes water have a bit more utility as a budget paracematol that purges chems you don't like

## Changelog
:cl:
balance: after the initital rush of effects are over, water still has an effect on you. Heals minor stamina loss and health damage.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/other.dm
+++ b/code/modules/reagents/reagents/other.dm
@@ -64,6 +64,9 @@
 		if(6 to 10) //sip 2
 			L.adjustStaminaLoss(-0.5*effect_str)
 			L.heal_limb_damage(0.1*effect_str, 0.1*effect_str)
+		if(11 to INFINITY) //anything after
+			L.adjustStaminaLoss(-0.15*effect_str)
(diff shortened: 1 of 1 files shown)


Ieteihowmee:
Good. Can we make it so we can fill up bottles from water tanks? That would be nice.

davidchan13:
> Good. Can we make it so we can fill up bottles from water tanks? That would be nice.

This is default behavior, you can already do this.

TiviPlus (approved):
stay hydrated

Tristan63:
sergeant major says:

stay hydrated you dumbfuck!

davidchan13:
someone code canteen waterbottles that spawn in standard kit.
MARINES WILL DRINK 4-8 CANTEENS OF WATER EVERY OP, NOT TO EXCEED 8 CANTEENS. HYDRATE, FUCKBAGS!

PR #7870, opened 2021-08-09
Makes the bolt action rifle act closer like a bolt action codewise.

SplinterGP:
## About The Pull Request
Removes the snowflake code stuff from the bolt action, using cock() code for chambering ammo, also makes it now not autochamber the bullet when reloaded.

## Why It's Good For The Game
Better code with less vars and snowflake.

## Changelog
:cl:
code: changed the TL-127 bolt action rifle code to behave more like a bolt action, it also now doenst autochamber on loading another magazine.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -444,7 +444,7 @@ User can be passed as null, (a gun reloading itself for instance), so we need to
 	current_mag = magazine
 	replace_ammo(user,magazine)
 	if(!in_chamber)
-		ready_in_chamber(user)
+		load_into_chamber()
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/gun_system.dm
```diff
@@ -444,7 +444,7 @@ User can be passed as null, (a gun reloading itself for instance), so we need to
 	current_mag = magazine
 	replace_ammo(user,magazine)
 	if(!in_chamber)
-		ready_in_chamber(user)
+		load_into_chamber()
```
SplinterGP, line 447: this should probably have a flag check instead.

TiviPlus (approved):

PR #7886, opened 2021-08-09
silo collaspe starts at 35 minutes rather than 45

Pariah919:
## About The Pull Request
the more i've observed and played i've marines need an early game objective to point and give them cohesion so they can go take it
silos rn take too long to be forced to be made, you need miners but you can't hold more than 2, so marines just take the closest miner and whatever is the next closest
xenomorphs make big ass mazes in the same areas and killing it as marine has no payoff since they'll just run off and force you to take 3 or 4 teamfights before you fucking die because by the time they make a silo it'll be mid to late game anyway, as 45 minutes is about mid to late game usually
35 minutes will force silos ten minutes earlier which is probably 15 minutes into the op rather than 25 which allows marines to buddy up and do something in squad based play rather than forcing them into a ungaball to take some random maze full of autism and random ass garbage

## Why It's Good For The Game
moves silos from a late game end game meme to being early game (15 minutes after drop, which is usually at 12:25 anyway)

## Changelog
:cl:
balance: silo collapse starts at 30 minutes rather than 45
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -139,7 +139,7 @@
 /// This is used to ponderate the number of silo, so to reduces the diminishing returns of having more and more silos
 #define SILO_OUTPUT_PONDERATION 1.75
 //Time (after round start) before siloless timer can start
-#define MINIMUM_TIME_SILO_LESS_COLLAPSE 45 MINUTES
+#define MINIMUM_TIME_SILO_LESS_COLLAPSE 35 MINUTES
 
 #define INFESTATION_MARINE_DEPLOYMENT 0
 #define INFESTATION_MARINE_CRASHING 1


Passe2142:
Xenos at early game can only make 1.The silo less timer does not allowed for a comeback on psy be either cocoon or drain, even with 8~12 xenos against rushers, been there where the tad just lands NE caves of Research outpost, we killed 10 marines or so and we could not get the silo up in 5 mins.
Marines see silo on tracker > tad op scout NVG, rush it with plasma cutter and GG. There are Queens that make an early silo , there is some lead way to make them a bit later on so it doesnt feel forced. Silos only make larvas at 12.33 (after shutters) so placing an early silo has more disadvantages than a semi late one, where tad can land when the hive is mature at most late 12.24 , shutters can be opened early too. There are no benefits of an early silo on diestress  that i see. A silo from 12.33 till 12.45 will make around 1 or 2 larvas depending on pop.

KornFlaks:
While out of the scope of this PR, the silo collapse timer should really be map dependant.

Sure, this doesnt really do much on LV, Big Red, Prison, or other usual high pop maps, Barrenquilla and, to a lesser extent, icy caves are both fucked by the rapid timer.

Barrenquilla is just "Find the silo using tadpole global vision then drop pod instant delete. Med-evac to get out." Xenos either have to win by 12:45 currently or just lose. Reducing the timer would just make barrenquilla an automatic loss, well even moreso than it is currently.

Hekzder:
40 would be a better middle ground IMO while still being a nerf. 35 seems a bit too harsh. Assuming shutters open at about an average time of 12:25, that gives about fifteen minutes of no silo Xenos as opposed to the current roughly twenty.

Hekzder:
> While out of the scope of this PR, the silo collapse timer should really be map dependant.
> 
> Sure, this doesnt really do much on LV, Big Red, Prison, or other usual high pop maps, Barrenquilla and, to a lesser extent, icy caves are both fucked by the rapid timer.
> 
> Barrenquilla is just "Find the silo using tadpole global vision then drop pod instant delete. Med-evac to get out." Xenos either have to win by 12:45 currently or just lose. Reducing the timer would just make barrenquilla an automatic loss, well even moreso than it is currently.

Barrenquilla is a horrible map that should not even be in the rotation, it is also almost never played for that reason. Barrenquilla being a genuinely awful map has nothing to do with this PR.

GradusNL:
Just the way this PR is written makes it feel like a salt PR. It's also entirely based on only the marine perspective, not that of the game as a whole. The reason that xeno don't place silos is because they don't offer any reward even slightly comparable to the downsides, as @Passe2142 describes. This is your second silo timer nerf (first down to 45 from a hour). You just keep forcing the timer down instead of fixing the real issue. Give silos instead a real risk/reward balance, instead of just forcing xeno to place them earlier and earlier.

> moves silos from a late game end game meme to being early game

Instead of a "late game" meme silos will be an early game meme. Equally useless, only placed earlier due to the collapse timer.

Sebbe9123:
Could always make it so silo has a 50% production of larvae before shutters open to make this a thing. Also gives marines an incentive to get groundside.

SplinterGP:
> 
> 
> While out of the scope of this PR, the silo collapse timer should really be map dependant.
> 
> Sure, this doesnt really do much on LV, Big Red, Prison, or other usual high pop maps, Barrenquilla and, to a lesser extent, icy caves are both fucked by the rapid timer.
> 
> Barrenquilla is just "Find the silo using tadpole global vision then drop pod instant delete. Med-evac to get out." Xenos either have to win by 12:45 currently or just lose. Reducing the timer would just make barrenquilla an automatic loss, well even moreso than it is currently.

my tip for that is not play distress on barrenquilla, that map is not made for that and doenst take it into account, play hunt.

TiviPlus:
Basically what Passe said

PR #7887, opened 2021-08-09
SADARs cost 120 from 90

Pariah919:
## About The Pull Request
The SADAR is way too cheap at 90 points and i'm seeing way too many of them rushed early game, which can be fair if it's the only thing you're going for but at 120 you'll see less overall of them in a game and more like one or two
I'd like to do a full rework to the RPGs sooner or later (RR, SADAR, and maybe add a midgame one) but this is a good stopgap for now

## Why It's Good For The Game
SADAR is heads and shoulders above everything else so this really represents that.

## Changelog
:cl:
balance: SADAR RPG costs 120 from 90 in req.
/:cl:

### DIFF
--- a/code/__DEFINES/supply.dm
+++ b/code/__DEFINES/supply.dm
@@ -2,4 +2,4 @@
 #define B18_PRICE 120
 #define B17_PRICE 80
 #define MINIGUN_PRICE 80
-#define SADAR_PRICE 90
+#define SADAR_PRICE 120


KornFlaks:
I think a better option would be increasing the price of the rockets, not the tube itself.

Then again, there's nothing wrong with the SADAR's price currently. This will just drive more to disposable spam which, personally, isnt fun for anyone except the for the marine FF'ing other marines in the back.

lbnesquik:
I'd rather have something a little cheaper. 120 is really quite expensive.

TiviPlus:
SADAR is already good when used properly and is already used constantly it doesnt need this

SplinterGP:
> 
> 
> SADAR is already good when used properly and is already used constantly it doesnt need this

its a NERF on cost not a buff, were you sleepy ?

SplinterGP:
red label because balance freze

Lefunny:
Sadars need be more cheaper than just 90 points, they almost dont do damage and are a waste of points, they being early rushed is a punishment for xenos thinking they stand a chanc agaist marines early game

PR #7888, opened 2021-08-10
Endure No Longer Has Bonus Crit HP; Super Rage Enhances Endure

Surrealaser:
## About The Pull Request

1. Endure loses all bonus crit HP; it now only provides slow/stagger immunity, explosive resistance and immunity to KOing while at negative HP.
2. If active, Endure grants bonus crit HP that scales with the Ravager's rage power  (up to 100 HP) when the Ravager uses Rage while at negative HP and its duration increases by 2 seconds each time the Ravager slashes a living mob.

## Why It's Good For The Game

Endure is less effective for general use, but becomes more powerful when used with Rage, especially while at negative HP.

Skillful and judicious use of Endure with Rage is rewarded/encouraged.

Rewards aggression while super raging.

## Changelog
:cl:
balance: Endure loses all bonus crit HP; it now only provides slow/stagger immunity, explosive resistance, and immunity to KOing while at negative HP.
balance: If active, Endure grants bonus crit HP that scales with the Ravager's rage power  (up to 100 HP) when the Ravager uses Rage while at negative HP (Super Rages) and its duration increases by 2 seconds each time the Ravager slashes a living mob.
/:cl:


lKiseki:
Very epic.

lovegreenstuff:
I still think that lowering the duration of endure would be best for balance.
Like this, it just means that ravager can have 20 seconds of endure uptime, with the latter 10 seconds post rage being the endure bit that matters, since you're already low.

Surrealaser:
> I still think that lowering the duration of endure would be best for balance.
> Like this, it just means that ravager can have 20 seconds of endure uptime, with the latter 10 seconds post rage being the endure bit that matters, since you're already low.
> 
> Basically just more endure for your endure, since you heard ravagers like endure.

Lowering the duration means it's only used for escapes as stated previously.

Also it would take essentially perfect timing to get 19 seconds of duration or better; if you manage to pull it off while chancing Rage at negative HP (the only time you're getting a refresh), all the more power to you.

lovegreenstuff:
I'm sayin' it ain't really fixin' the what I think is the core problem with endure.

It's too forgiving and it lasts too long. The refresh is just another 10 seconds of standard endure with 10 more seconds of stagger immunity before it.

Surrealaser:
> I'm sayin' it ain't really fixin' the what I think is the core problem with endure.
> 
> It's too forgiving and it lasts too long. The refresh is just another 10 seconds of standard endure with 10 more seconds of stagger immunity before it.
> 
> Only time will tell if this will really address the problem, but I'm betting on it not doing that.

It just got significantly less forgiving.

Again for typical usage it has just seen a significant nerf.

Surrealaser:
Changed it from a duration reset to Endure's duration increasing by 2 seconds each time the Ravager slashes a living mob while super rage is active.

EvilMelonTheSecond:
I like the idea of this fix to endure being used as a free escape tool. I wanna see what happens with the current stats when its TM'ed

TiviPlus (approved):

PR #7893, opened 2021-08-10
boiler gets NO speed up from weeds

Pariah919:
## About The Pull Request
boiler is too fast for a backliner, it is hardly punished for being caught off guard, it has average health stats and is nearly always a leader, and you can stack it.

## Why It's Good For The Game
this is desperately needed, boiler shouldn't have got this in the first place

## Changelog
:cl:
balance: boiler gets NO speed up from weeds, at all.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
@@ -15,6 +15,7 @@
 
 	// *** Speed *** //
 	speed = 0.3
+	weeds_speed_mod = 0
 
 	// *** Plasma *** //
 	plasma_max = 800


lKiseki:
Not a bad change. Currently, Boilers are able to outrun most marine setups on weeds and it's rather hard to kill them as a result even if they make mistakes. This should help with that.

lbnesquik:
I am not a fan, but note that boiler fire rate was recently buffed. The caste got a fair bit strong from that.

BorranScail:
Marines can currently outrun Boilers on weeds. This would make the problem even worse.

GradusNL:
> 
> Currently, Boilers are able to outrun most marine setups on weeds

> Marines can currently outrun Boilers on weeds.

The duality of man. Which is true?

davidchan13:
This doesn't make any sense, marines can sprint faster than Boilers can move. Boiler is the tied for the slower xeno in the game, only a young hivelord is slower and an ancient tier Boiler, Crusher and Hivelord all have .1 speed. And now we saying the slowest xeno needs to be slower when it already struggles to defend itself against a competent marine?

TiviPlus:
Boiler doesnt need this nerf

PR #7894, opened 2021-08-11
Crusher  has less weed speed

Pariah919:
## About The Pull Request
crusher is a pretty good caste overall, in general i think it's near unpunishable outside of set scenarios or slapping it with a sadar to the fucking face
this should punish crusher far more for being out of position in team fights and make him unable to cycle as effectively

## Why It's Good For The Game
Crusher is absurd as it is, don't think you can argue against it, it got it's on demand stun back. It deserves a slap.

## Changelog
:cl:
balance: crusher is far slower on weeds (-0.15 instead of 0.4)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
@@ -15,6 +15,7 @@
 
 	// *** Speed *** //
 	speed = 0.1
+	weeds_speed_mod = -0.15 // Slightly more than halved weed speed up, if you're of position this bad, bad times.
 
 	// *** Plasma *** //
 	plasma_max = 200
@@ -22,6 +23,7 @@
 
 	// *** Health *** //
 	max_health = 325
+	sunder_recover = 0.1 // Crusher recovers sunder FAR slower than normal castes
 
 	// *** Evolution *** //
 	upgrade_threshold = 250
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
@@ -23,7 +23,6 @@
 
 	// *** Health *** //
 	max_health = 325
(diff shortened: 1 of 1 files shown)


Sebbe9123:
Might be an idea to make crushers take less sunder in general than the average xeno to compensate so they can actually tank or stay in a fight. Kinda weird how the "tank" beno is less tanky than the "damage" beno.

lKiseki:
Bad idea. I'd say nerf the speed only first, then if necessary we can look into more different solutions than 'this caste in specific regens sunder slower because yes'.

code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
```diff
@@ -15,13 +15,15 @@
...
 	speed = 0.1
+	weeds_speed_mod = -0.15 // Slightly more than halved weed speed up, if you're of position this bad, bad times.
 
 	// *** Plasma *** //
 	plasma_max = 200
 	plasma_gain = 10
 
 	// *** Health *** //
 	max_health = 325
+	sunder_recover = 0.15 // Crusher recovers sunder FAR slower than normal castes
```
Pariah919, line 26:
```suggestion
```

lKiseki:
Okay. Now this is better. Very epic.

lbnesquik:
Given the changes, it isn't certain that the thumbs are still up to date.

SplinterGP (approved):
Okay change, the weed speed boost is pretty big considering frenzy with it, the crusher is a slow xeno and feom 0.1 to -0.3 was alot

BorranScail:
Sadar is supposed to stop Crusher. Crusher is not absurd. It can still die if it goes out of line and tries to solo marines for too long.

davidchan13:
> Okay change, the weed speed boost is pretty big considering frenzy with it, the crusher is a slow xeno and feom 0.1 to -0.3 was alot

Crusher is also the only xeno that doesn't get faster as it gets older, a young queen is faster than a crusher and crushers don't have diagonal charging like corp mercs.

PR #7895, opened 2021-08-11
healing infusion lingers for 40 seconds rather than 60, and has 2.5 more seconds of C/D

Pariah919:
## About The Pull Request
Healing infusion lingers for only 30 seconds rather than 60
Beam has 15 second CD, rather than 12.5

## Why It's Good For The Game
60 seconds of lasting is honestly absurd, considering the ability can be used and lingers out of sight of the hivelord, the argument that it's worse at healing than drone heal is also absurd since it heals far more sunder, and the tiles you save not having to run up to the damn idiot to acidic salve, letting you do this MID COMBAT makes it a far better ability than salve, which is barely used these days anyway.
15 seconds of CD makes more sense than 12.5, considering the ability now lingers for only 30 seconds and at best has 10 ticks of working.

## Changelog
:cl:
balance: Healing infusion only lingers for 40 seconds, and has 2.5 more seconds of CD, totalling up to 15.
/:cl:


code/__DEFINES/mobs.dm
```diff
@@ -591,7 +591,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 #define HIVELORD_TUNNEL_DIG_TIME 10 SECONDS
 #define HIVELORD_TUNNEL_SET_LIMIT 8
 #define HIVELORD_HEAL_RANGE 3
-#define HIVELORD_HEALING_INFUSION_DURATION 60 SECONDS
+#define HIVELORD_HEALING_INFUSION_DURATION 30 SECONDS
```
Pariah919, line 594:
```suggestion
#define HIVELORD_HEALING_INFUSION_DURATION 40 SECONDS
```
40 seconds seems more reasonable than 60

Surrealaser:
No. Healing Infusion is fine, and even if it was overtuned, the way to fix it would be to increase the healing but have only its overheal fix Sunder (though obviously not at a 1 : 1 ratio).

TiviPlus:
Infusion is fine

PR #7902, opened 2021-08-11
Shotgun Shell Rebalance - Buck

lovegreenstuff:
## About The Pull Request
Buckshot has vastly reduced scatter and damage falloff, as well as increased stagger. In return, it no longer full stuns and has less knockback.
Basically, I took Ren's and MJP's suggestions and I turned it into a PR. 

I'm fully expecting this to be closed immediately, but it's not really my ideas anyways.
## Why It's Good For The Game
Currently, shotgun isn't really that good, especially buckshot. Buckshot is currently just a melee weapon and even then, it does less damage on PB than flech does in most cases.

Even like this, in-game testing shows that it gets outperformed by flech in terms of damage at any reasonable range and armor, with the sole exceptions of runner and drone at point blank range.
## Changelog
:cl:
balance: buckshot now has vastly improved ranged capabilities, as well as increased stagger in return for its hardstun and some knockback.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -684,18 +684,18 @@ datum/ammo/bullet/revolver/tp44
 	flags_ammo_behavior = AMMO_BALLISTIC
 	bonus_projectiles_type = /datum/ammo/bullet/shotgun/spread
 	bonus_projectiles_amount = 5
-	bonus_projectiles_scatter = 10
+	bonus_projectiles_scatter = 3
 	accuracy_var_low = 9
 	accuracy_var_high = 9
 	accurate_range = 3
 	max_range = 10
 	damage = 40
-	damage_falloff = 4
+	damage_falloff = 2
 	penetration = 0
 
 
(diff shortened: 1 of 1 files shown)


usnpeepoo:
i don't think buck should be more range, if its worser than flech in pb then it should be buffed to counter that not making it more like flechette where you're shooting farther.

RenGusta:
Having seen it comparatively tested reasonably thoroughly, I'm confident this will make Buck at least appear more in general instead of just in PB. Could always balance flech and buck off one another, but that's the topic of another PR after this one.

OrbisAnima:
I like the change, but shouldnt the /datum/ammo/bullet/shotgun/spread also get the modification to the damage falloff to 2 instead of 4?

lovegreenstuff:
> I like the change, but shouldnt the /datum/ammo/bullet/shotgun/spread also get the modification to the damage falloff to 2 instead of 4?

yes. i am stupid.

lovegreenstuff:
closing for balance freeze

lovegreenstuff:
didn't have to close this for balance freeze apparently.
am stoopid

TiviPlus:
Making buckshot flechette lite is not buckshots intended role its supposed to be the close range option

PR #7903, opened 2021-08-11
Shotgun Shell Rebalance - Flechette

lovegreenstuff:
## About The Pull Request
See title.

## Why It's Good For The Game
Currently, shotgun isn't really that good. 
Also, flechette being basically better than slug and buck at both of their jobs in 99% of scenarios is silly.

## Changelog
:cl:
balance: increased flech AP
balance: flech is now a req buy at 10 points.
tweak: flech side projectiles are now a child of base flech.
/:cl:


lbnesquik:
More range is good. I always prefer damage falloff to vanishing bullets anyway.

OrbisAnima:
Fun changes, but is it intended for only the bonus projectiles to have a buff in penetration and only the main shot to have a buff in the range, or did you meant to do it in both?

lovegreenstuff:
> More range is good. I always prefer damage falloff to vanishing bullets anyway.

I wasn't aware that I increased the range. I think you're thinking of the buckshot PR. This just increases Flech AP.

> Fun changes, but is it intended for only the bonus projectiles to have a buff in penetration and only the main shot to have a buff in the range, or did you meant to do it in both?

Frankly, flech stats are so old that I don't think any of them know if one is supposed to have more AP than the other. I think it's silly that the same projectile has different APs from the bonus, since it's the only projectile in the entire game that does that.

Passe2142:
Slugs my beloved.

lovegreenstuff:
closing for balance freeze

lovegreenstuff:
didn't have to close this for balance freeze apparently.
gonna reopen it and see if the changes i made updated.

Pariah919:
even in the state you put flechette in I don't think it's good enough to warrant being a req buy
req only ammo suffers massively from being overpriced and difficult to use unless you make the amount of points it's worth like 2-3 since you'll only get like 25 shells

lovegreenstuff:
I used the incen slugs as a basis for the shell pricing. I could probably make a PR later to reduce the price of both.

TiviPlus:
Make it the big box that you buy

lovegreenstuff:
did

Lefunny:
This is really good to the game as we need some new shotgun shells, xenos can almost outwin any marine in close range and this make the game really unfair, perfect chang if i can say.

PR #7932, opened 2021-08-14
Plasmacutter Now Does Explosive Devastating Rather Than QDel(Object)

KornFlaks:
## About The Pull Request

Plasma cutters now do devastating explosive damage instead of straight qdel object. This fixes issues with hard qdel instead of going through damage pathways such as #7872 Silo Collapse Timer not being called.

## Why It's Good For The Game

Fixes issues with silo collapse timer when plasma cutters are used to destroy silos. Silos however now require multiple charges (5 to be exact) of the plasma cutter in order to destroy as it now inherits plain resin explosive reaction. No more instant delete silos for synths now.

Tadpole furthermore no longer can plasma cut the windows out without the shutters falling, due to actually going through the intended damage pathways.

Otherwise, there are no other practical differences in plasma cutting operation. Plasma cutter remains the instant delete wall tool it always was. However, some objects, e.g. silos and cockpit windows, will require more charges / hits in order to destroy, as their explosive reaction is overridden with reduced values.

## Changelog
:cl:
fix: Fixed plasma cutter and silo damage interaction. Now does explosive devastating damage instead of instant qdel.
fix: Fixed plama cutter and tadpole window interactions. Shutters now fall if the window below it is plasma cut.
/:cl:

### DIFF
--- a/code/game/objects/items/tools/mining_tools.dm
+++ b/code/game/objects/items/tools/mining_tools.dm
@@ -338,5 +338,5 @@
 		return TRUE
 
 	cut_apart(user, O.name, O)
-	qdel(O)
+	O.ex_act(EXPLODE_DEVASTATE)
 	return TRUE


Losenis:
>Tadpole furthermore no longer can plasma cut the windows out without the shutters falling, due to actually going through the intended damage pathways.
Make the shutters PC-able given they're already acidable anyway

KornFlaks:
> Make the shutters PC-able given they're already acidable anyway

Actually, why are they acidable? They're not suppose to be acidable.

image

KornFlaks:
> Make the shutters PC-able given they're already acidable anyway

Nothing is stopping the current shutters from being plasma cutted. 
image

image

Fixing that is beyond the scope of this PR however.

AnCopper:
So what happens if someone is on the same tile as something being plasmacutted?

KornFlaks:
> So what happens if someone is on the same tile as something being plasmacutted?

Nothing. The "explosive" damage is done directly onto the object being interacted with. It's a direct damage application, no actual explosion.

Mobs / Players being interacted with by a plasma cutter are going through a different pathway and nothing changes on that front.

Losenis:
They were PC-able already? Huh, could swear they weren't.

KornFlaks:
> They were PC-able already? Huh, could swear they weren't.

They arent. But nothing on the actual object itself is preventing the plasma cutter from working. It may be come sort unintended interaction between the layers and plasma cutter. No clue.

Losenis:
That's gonna be a pretty big issue then because the front sentries are at an angle where they WILL shoot the windows off sooner or later and make the shutters drop. Other than the meme build of putting sentries at the front, PCing the windows off and replacing them with cades was also a countermeasure to the sentries blinding you by forcing the shutters down. Also allowed making the front be an entrance to not just block an entire hallway if you landed it there, like nexus in LV.

BraveMole:
The "proper" fix would have been for silos to proc the timer on Destroy() rather than obj_destruction. 

This is arguably more of a balance PR than a fix one IMO

PR #7933, opened 2021-08-14
Makes FOB Drone Console Indestructable

KornFlaks:
## About The Pull Request

Adds flag that make the FOB drone console indestructable from Marine or Xeno action.

## Why It's Good For The Game

No more funny plasma cut or RR at start of round to instantly grief the marine side from making a FOB.
Im surprised it hasnt happened yet and only in my testing that I discovered this.

## Changelog
:cl:
fix: Made FOB drone console indestructable by both marines and xenos.
/:cl:

### DIFF
--- a/code/modules/mining/remote_fob/remote_fob_computer.dm
+++ b/code/modules/mining/remote_fob/remote_fob_computer.dm
@@ -7,7 +7,7 @@
 	icon = 'icons/Marine/remotefob.dmi'
 	icon_state = "fobpc"
 	req_one_access = list(ACCESS_MARINE_REMOTEBUILD, ACCESS_MARINE_CE, ACCESS_MARINE_ENGINEERING)
-	resistance_flags = UNACIDABLE|INDESTRUCTIBLE
+	resistance_flags = RESIST_ALL
 	networks = FALSE
 	off_action = new/datum/action/innate/camera_off/remote_fob
 	jump_action = null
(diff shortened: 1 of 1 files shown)


BraveMole:
"No more funny plasma cut or RR at start of round to instantly grief the marine side from making a FOB." We have admins for that, they can respawn in it 10 seconds?

KornFlaks:
Admins could respawn it but why is it even destructable to begin with? Admin interaction should not be manditory in game breaking instances if the code itself prevents it. It shouldnt be possible in the first place.

BraveMole:
I don't really see the point of grief-proofing, because there are 100000 ways to grief. 

Even with your thing, nothings prevent an engineer to just eject the mats and hide them?

KornFlaks:
Yes, there are 1001 ways to grief. That does not mean we should just shrug our shoulders and shove the responsibility of making a round run smoothly onto the admins when we could do something about a clearly unintended interaction.

An engineer rapidly ejecting the mats and hiding them is a ingame RP issue. That is intended and often a tactic used by Captains / FCs / whoever to force tadpole or no-FOB operation. Another engineer can just insert their own materials to build a FOB anyways.

Deleting the FOB drone console however is a drastic overreaction to force a tadpole operation and locks a key feature of marine gameplay due to the un-reversable action (by players) from a single individual.

AnCopper:
> I don't really see the point of grief-proofing, because there are 100000 ways to grief.
> 
> Even with your thing, nothings prevent an engineer to just eject the mats and hide them?

What can be code should be code, what cannot be must be policy.

TiviPlus (approved):

PR #7951, opened 2021-08-16
Lava won't burn you when you walk diagonally near it

BraveMole:
## About The Pull Request

Lava shoud not burn you when you are visually not stepping on it. This also mean that diagonally walking in lava will no longer cause 3 (3!!!) times the damage when walking cardinally, but just 1 time

## Changelog
:cl:
fix: Lava won't burn you when you walk diagonally near it
/:cl:

### DIFF
--- a/code/game/turfs/open.dm
+++ b/code/game/turfs/open.dm
@@ -343,6 +343,9 @@
 			if(L.stat == DEAD)
 				continue
 
+			if(L.moving_diagonally == FIRST_DIAG_STEP)
+				continue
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

SplinterGP (changes requested):
has a bug that lets you ignore lava by trying to move diagonally while a wall is in the way of second step.

BraveMole:
I hate diag code i hate diag code

PR #7963, opened 2021-08-18
orders last longer

Pariah919:
## About The Pull Request
10->25 ticks, effective buff of 30 seconds

let's be real here orders are complete and total trash because they last for nearly no time so you can't even get an accurate idea of how they affect the match overall
i get there's a balance freeze rn but i don't think anyone would mind seeing this tested in the current state of balance

## Why It's Good For The Game
makes orders kinda exist as a game mechanic rather than non existant

## Changelog
:cl:
balance: marine bufforders last for 30 more seconds
/:cl:


TiviPlus:
Balance freeze

PR #7964, opened 2021-08-18
Spitter was to slow. this adjusts it to be more T2 like.

vilereaver41:
## About The Pull Request

Adjusts Spitter speed.

## Why It's Good For The Game

Spitter was as slow as a large clunky Praetorian whilst being a small Xeno, this adjusts it to be more T2-like in speed.

## Changelog
:cl:
Balance: Spitter speed now  0.6-mature  -0.7-Elder & 0.8-ancient, instead of 0.6 at ancient which is as slow as a large Xeno.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/spitter/castedatum_spitter.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/spitter/castedatum_spitter.dm
@@ -11,7 +11,7 @@
 	melee_damage = 17
 
 	// *** Speed *** //
-	speed = -0.3
+	speed = -0.4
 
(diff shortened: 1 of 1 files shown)


David-Stormwell:
There is a small reason why ranged castes aren't all fast as fuck, if you could run circles around marines while spitting at them, slowing them down and just taking out entire squads solo, then it isn't much of a challenge now is there?
This happened once already with the sentinel back in the day, when it was power creeped beyond everything, going in and out spitting down marines before they could even run away.

BraveMole:
David said no and spitter is already good anyway

PR #7979, opened 2021-08-19
Fix some harddels

BraveMole:
Fix some bad harddels (like fix human and xeno harddels) it seems

The queue time to harddel is set to 5 minutes to 2 minutes, like tg is doing:
https://github.com/tgstation/tgstation/pull/60174

Port from tg:
Prevent QDEL_IN to hard deling : https://github.com/tgstation/tgstation/pull/60937
qdel(weakref(thing)) will also qdel the thing : https://github.com/tgstation/tgstation/pull/56206

5 % of resin nodes are still harddel, i'll fix that in weed refactor.

I probably didn't fix harddel from items deleting, this would need a live test.

All corpsespawner were harddeling, because of me

A lot of old code with no harddel protection


code/modules/orbits/spaceship.dm
```diff
@@ -50,7 +52,11 @@ GLOBAL_VAR_INIT(current_orbit,STANDARD_ORBIT)
 /obj/machinery/computer/navigation/Initialize() //need anything special?
 	. = ..()
 	desc = "The navigation console for the [SSmapping.configs[SHIP_MAP].map_name]."
-	addtimer(VARSET_CALLBACK(src, changing_orbit, FALSE), 30 MINUTES) //people running away far too quickly it seems
+	timer_id = addtimer(VARSET_CALLBACK(src, changing_orbit, FALSE), 30 MINUTES, TIMER_STOPPABLE) //people running away far too quickly it seems
+
+/obj/machinery/computer/navigation/Destroy()
+	deltimer(timer_id)
```
BraveMole, line 58: needed because it's a VARSET_CALLBACK, so it was qdeleting

TiviPlus (approved):

PR #7988, opened 2021-08-21
Prevent from dragging while riding a crusher

BraveMole:
## About The Pull Request

The slowdowns were not applied when a runner was dragging a human for exemple. Two solutions to fix that : 
- make the crusher slows down when the rider grabs something. While possible, i don't like this because as a crusher i don't want to be slowed down and die because my rider is stupid
- Prevent runner from dragging. Which makes sense, the runner is too busy riding to drag

## Why It's Good For The Game

No more funny drag

## Changelog
:cl:
fix: Prevent from dragging while riding a crusher
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -7,7 +7,7 @@
 //#define DEBUG_ATTACK_ALIEN
 
 /mob/living/proc/attack_alien_grab(mob/living/carbon/xenomorph/X)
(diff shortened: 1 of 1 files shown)


QualityVan:
I was gonna say this won't do anything for grabbing something a body's in, like bodybags or loose cocoons, but those mostly don't have slowdown anyway.

BraveMole:
> I was gonna say this won't do anything for grabbing something a body's in, like bodybags or loose cocoons, but those mostly don't have slowdown anyway.

xeno cannot even grab those

TiviPlus (approved):

PR #7999, opened 2021-08-21
Queen is Now Slowed Upon Screech

KornFlaks:
## About The Pull Request

Adds an on average 5 second slowdown for Queens upon screech.

## Why It's Good For The Game

Screech has been used far too often as a free get-out-of-jail cards by queens who screech then run away. This will encourage more combat queens to follow through with screeching and slashing marines. This slowdown additionally rewards skilled queens who wait to screech until they're right next to a marine.

## Changelog
:cl:
balance: Adds an on average 5 second slowdown on Queens following a screech.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/queen/abilities_queen.dm
@@ -109,6 +109,10 @@
 	succeed_activate()
 	add_cooldown()
 
+	// Adds on average 5 seconds of slowdown on the screeching xeno.
+	X.add_movespeed_modifier(type, TRUE, 0, NONE, TRUE, 2)
+	addtimer(CALLBACK(src, .proc/reset_speed), rand(4 SECONDS, 6 SECONDS))
+
 	playsound(X.loc, 'sound/voice/alien_queen_screech.ogg', 75, 0)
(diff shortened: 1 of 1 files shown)


Ieteihowmee:
Nice

Losenis:
> Screech to just not die
> Screech is now on cooldown if the opportunity to btfo the entire ungaball appears
> As well as being on cooldown if you need to not die too soon

Was this actually an issue at all? Would that not make queens even more defensive knowing that they have one less escape ability?

> This slowdown additionally rewards skilled queens
> Rewards

That's some wild wording.

bowlofcereal:
i wish that when i played marine it didnt feel like a fucking cakewalk

Lefunny:
This is a real good chang to the game, i mean queens being able to survive after the screech without even being hurted or outspeeded by marines is really a balancment issue, i would like if the plasma cost become something that really make queens think two times before screething

bobmancer343:
As a 'combat queen' This is a change I feel is unneeded. Queen already has enough struggles in combat and a nerf makes playing combat queen even harder.

Eskjjlj:
The purpose of the screech is to allow the queen to run away though..... Queens will have no way to escape sticky situations with this change.
Right now marines just need to keep the pressure after the screech and kill the queen during the LONG cooldown. Which is more than enough for interesting gameplay.

All in all this change will only make people play queen less when it is already a shunned role.

TiviPlus:
Xenos dont need nerfs right now

PR #8001, opened 2021-08-21
Standardizes Shrike Throw Distance to 4

KornFlaks:
## About The Pull Request

The PR proposes the unification of shrike throw ability distances to 4, set via mob define rather than hardcoded in a single embedded line.

It also includes a sanity check in the shrike unrelenting force throw that was strangely missing.

## Why It's Good For The Game

Understanding the various distances the shrike could throw objects and marines was can be considered confusing to new xenos playing shrike. Who knew that shrike fling threw marines 3 tiles and objects 4 tiles. Shrike unrelenting force threw at a hardcoded 6 tiles, which has been argued to be far too much. This standardizes shrike distances, increasing fling distances for objects to 4 and reducing unrelenting force distances by a mere 2 tiles.

This removes the seemingly random decisions done by previous coders with numbers thrown onto a dart board for throw distances. Making shrike lowering the skill floor to play the caste effectively.

## Changelog
:cl:
balance: Shrike throw distances is now a singular 4 tiles for both abilities.
/:cl:


Lefunny:
This is pretty helpfully for new xenos players making the skill chain of shrike a bit more easy to understand.

Eskjjlj:
Extreme Shrike nerf. I disagree. 
The purpose of fling throw is to stun a single marine for a short amount of time, not throw him into the hive and unrelenting force is meant to push overextending marines towards your xeno friends.
With this change Shrike will not be able to throw marines far enough from their allies to do significant damage to a push.

If you made the cooldown of unrelenting force as short as the other ability then it would be more balanced but still defeats the purpose of Shrike.

KornFlaks:
> Extreme Shrike nerf. I disagree.

Extreme? I beg to differ. In numerical terms, this reduces the shrike unrelenting force distance sure but also increases the alternative and more skillful shrike throw ability, fling, of marines by 1 tile.

When taken into account, a skilled shrike combining the usage of both shrike unrelenting force and shrike fling would result in only 1 tile difference from the current combo. And we can all agree that 3 tiles of throw is far too little, as discussed in the currently merged warrior throw distance to three tiles. Shrike fling also has a far lower cooldown of 12 seconds compared to unrelenting force's 50 seconds.

This may in fact be a shrike buff for all they need to do now is be a little more accurate on the placement and selection of both shrike throw abilities.

bobmancer343:
But unrelenting force is really hard to hit properly with a regular instant activation keybind. This is a direct nerf that isn't needed at all right now.

Eskjjlj:
Yeah true. With a keybind it activates toward the direction you are facing so you must use the mouse to select the ability in order to use it effectively.

KornFlaks:
The usage of terms such as "nerfs" and "buffs" have really soured the waters in these PRs. This is not a "nerf" nor is this a "buff". There are objective downsides but also upsides to this change as described in my previous comment. It is in overall balanced and an improvement to shrike gameplay.

bobmancer343:
I disagree. The reason unrelenting force has more tiles is because it's effective at PB range. It's hard to pull off a decent PB without suiciding and it rewards the shrike player for taking risks. While I agree sure this is fine but I personally disagree with this as I think PBs should be rewarded better.

Ieteihowmee:
Nice

turnop:
maybe if marines actually learned to use their fucking brains we wouldn't see these godawful hard nerfs coming out for every caste constantly

bowlofcereal:
> The usage of terms such as "nerfs" and "buffs" have really soured the waters in these PRs. This is not a "nerf" nor is this a "buff". People see the first word "nerf" or "buff" and never bother to look beyond it.
> 
> There are objective downsides but also upsides to this change as described in my previous comment. It is in overall balanced and an improvement to shrike gameplay.

This is quite literally a nerf though as a shrike main you just made it even harder to stop marine pushes how is that not a nerf

EgorDinamit:
Yet another xeno nerf, how unexpected.
Shrike's fling was already nerfed, wasn't it? Now you are nerfing the only other useful ability Shrike has.

BraveMole (changes requested):
The xeno cabale hates this, and i think it's too big of a nerf as well

PR #8003, opened 2021-08-22
Powerfist Now Applies a Stun Scaling With Power Level

KornFlaks:
## About The Pull Request

A simple addition of a stun that scales with power level of the powerfist being used.

When used on marines, the maximum power level results in very short knockdown that exists only to drop weapons and a stun lasting 3 seconds. The stun time precisely depends on the power level of the powerfist being used.

On xenos, the stun is significantly shorter. As they recieve several reductions in stun time, the exact values could not be determined but testing resulted in approximately 0.5 seconds of stun. I observed that a noticable amount of stun time is actually done while the xeno is flying through the air, resulting in a very short practical stun time once the xeno stops flying. If anything, it gives the xeno in question a very nice scare rather than anything actually debilitating, other than the sudden relocation of course.

## Why It's Good For The Game

Currently, marines hit by powerfists fly multiple screens away and land on their feet with their guns in hand ready to return to fighting with only minor damage thanks to armor blunting the brute damage. Which makes absolutely no sense.

This PR introduces a stun aspect to the power fist that results in marines dropping their weapons near their original position if they do not have a mag harn.

Xenos on the other hand might have a very short icon change for half a second but, since they rarely have anything in their hand, are otherwise unaffected by this PR.

## Changelog
:cl:
add: Powerfist now applies a stun scaling with power level
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/misc.dm
+++ b/code/game/objects/items/weapons/misc.dm
@@ -108,13 +108,7 @@
 	var/atom/throw_target = get_edge_target_turf(M, get_dir(src, get_step_away(M, src)))
 	var/throw_distance = setting * LERP(5 , 2, M.mob_size / MOB_SIZE_BIG)
 	M.throw_at(throw_target, throw_distance, 0.5 + (setting / 2))
-
- 	// Minor stun upon being thrown that scales with power level of the fist.
-	var/stun_multi = isxeno(M) ? 5 : 1 // This is required to make the stun actually do something to xenos other than a flicker of icon.
-	 // Actual stun for xenos mainly occurs during the flight-time following fist attack and results in a practical stun once the xeno stops flying of around half a second.
-	 // Marines on the other hand get knocked over and get up immediately with a stun lasting [setting] seconds.
-	M.apply_effects(setting * stun_multi, 0.1 * setting * stun_multi)
-
+	M.apply_effects(setting, 0.1 * setting)
 	cell.charge -= powerused
 	return ..()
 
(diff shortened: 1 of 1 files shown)


Lefunny:
Sounds good to me, marines should be able to punish xenos for getting in their meele, and with this stun most xenos that could easy couter ungaballs or stuff like that, good stuff for the game, oorah!

QualityVan:
What happens when the launched xeno starts next to a wall, so they don't actually move and the human can punch again right away?

KornFlaks:
Depending on the intent of the xeno, help intent will result in the xeno shuffling through the marine powerfist attacking the xeno.

Harm intent instead results in the xeno bouncing off the marine and being thrown in a random perpendicular direction.

There is also a significant delay between powerfist attacks, around a second. There is no way for a marine to perma-stun a xeno using this.

code/game/objects/items/weapons/misc.dm
```diff
@@ -108,6 +108,7 @@
 	var/atom/throw_target = get_edge_target_turf(M, get_dir(src, get_step_away(M, src)))
 	var/throw_distance = setting * LERP(5 , 2, M.mob_size / MOB_SIZE_BIG)
 	M.throw_at(throw_target, throw_distance, 0.5 + (setting / 2))
+	M.apply_effects(setting, 0.1 * setting)
```
BraveMole, line 111: this proc is blurry as hell, use apply_effect(setting, STUN) and apply_effect(0.1 setting, WEAKEN)

BraveMole (changes requested):

code/game/objects/items/weapons/misc.dm
KornFlaks: Done

BraveMole (approved):
Worth a try

PR #8005, opened 2021-08-22
Praetorian Acid Splash tweak.

vilereaver41:
## About The Pull Request

right now praetorians are only using neuro spit over a 2-second acid splash which is little use, this pr attempts to make it more useful, I don't expect the values to stay as they are as further test merging and in-game testing is required to check if the values are too high or too low.

## Why It's Good For The Game

this will make praetorians acid splash useable and helpful in combat and to slow marine pushes and better assist the hive.

## Changelog
:cl:
balance: acid splash delay changed to 1.2 young, 1.1 mature, 1 elder, and 0.9 ancient. 
balance: acid splash now 150 plasma a shot up from 75 a shot
balance: acid splash damage up from 30 to 45 as it has no regent transfer unlike neuro splash
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1836,9 +1836,9 @@ datum/ammo/bullet/revolver/tp44
 
 /datum/ammo/xeno/acid/heavy
 	name = "acid splash"
-	added_spit_delay = 8
-	spit_cost = 75
-	damage = 30
+	added_spit_delay = 0
+	spit_cost = 120
+	damage = 45
 
 /datum/ammo/xeno/acid/heavy/turret
 	damage = 20


Pariah919:
acid damage increases every maturity unless that got removed so this is going to get to absurd levels of damage
I don't like buffing xeno ranged as it makes both factions important bits less defined
I still think prae is fine as a caste and if you're going to buff it give it something more defined and unique that makes it fit into the master of none role rather than moving it into spit machine eksdee 2000

If neurotoxin is so much better than acid splash then equalize neurotoxin, reduce the purge rate or increase the speed that neurotoxin depletes from a marines system. Or add more softcounters to it.

Lefunny:
This sounds pretty good to me, even if you buff praetorian he cant match agaist a ranged marine and surely inst the better choice agaist a meele a one, the caste get between being incapable of fighting in meele or really helping in ungaballs as the acid ust deal a real little damage and a short knockdown, for now praetorian is just a queen but without buttons, armor but good pheros.

PR #8006, opened 2021-08-22
T-25 smartrifle minor rework

Pariah919:
## About The Pull Request
Changes the T-25 to be less of a Rifle equivalent and moves it more towards being a LMG equivalent.

Ammo capacity- 50->80
Wield delay- 0.6->0.9
Aim slowdown- 0.55->0.7

Damage- 25->20
AP-> 15->10
Sunder 1.5->1

## Why It's Good For The Game
Ammo Capacity
50->80
----
The T-25 should lean more towards a supportive role with this change and I believe giving it higher ammo capacity to trade off the damage/pen and gunstats it lost would be good enough combined with the .5 more sunder I gave it to make it lean towards a supportive stance rather than an aggressive pushing stance with the gun.
Compared to the SMMGs 200, it still loses out 120 bullets, for a faster firerate and better utility options (UGL/Miniflamer) and it still has far better gun stats than the T-29. Aswell as a far larger attachment roster (Bayoent, etc..)
----
Wield delay
0.6->0.9
----
The T-25, with being changed to be more of a LMG rather than an AR equivalent, necessitates 300ms more of wield delay to encourage it into that role, and encourages bringing a backup gun to your existent gun already, I believe this is a good change because it makes T-25s have to stay overall in the back more.
You can improve this to old levels overall by slapping an AGRIP on here, but you sacrifice utility like GLs for it. I believe this is a good overall change.
----
Aim slowdown
0.55->0.7
----
This is necessitated by the increase of the amount of ammo, as the magazines still fit in belts and stuff. I feel that it'll move the T-25 to a more supportive overall role rather than FRAG MAN SG MAN that I didn't intend it going into, and the original 15/15/2 design was pretty trash.
----
Damage
25->20
----
Having the same damage as a T-12 and higher pen is absurd, even if the gun is limited, and has the full roster of attachment options. This is necessary to move the gun to a more supportive stance rather than UNGA BETTER T-12!!!
----
Pen
15->10
----
5 pen has been removed and moved to sundering to move the gun to a more supportive stance.
----
Sunder
1->1.5
-----
This moves the gun to a more supportive role as it crunches up enemy armor for your other teammates to attack into, and makes the gun overall more supportive, it will crunch up armor for the rest of your team to deal with.
----
PACKET
200->320
----
The old packet was 50 times 4, the new packet is 80 times 4, Which is a 120 round increase from the old one. This might not be needed but I believe for consistencies sake to accommodate the higher magazine size, that it is a good change.
----

Overall

This moves the T-25 into a more supportive stance and removes the issue of it literally being the best DPS gun in the game, 30 more rounds to support your team with, it is more of an SAW than a beefed up AR, which it really shouldn't be a beefed up AR, this should hopefully move the gun towards being used as a supportive more mobile role for SG rather than being the main dps man of the entire marine team.

## Changelog
:cl:
balance: T-25 reworked, it is more like an LMG than a Assault Rifle. 80 rounds instead of 50, 1.5 sunder instead of 1, 10 pen instead of 15. 0.7 aim slowdown instead of 0.8, 0.9 wield delay instead of 0.6. Packet changed to accomdate the higher amount of ammo, it holds 320 rounds instead of 200.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1005,9 +1005,9 @@ datum/ammo/bullet/revolver/tp44
 
 /datum/ammo/bullet/smartgun/smartrifle
 	name = "smartrifle bullet"
-	damage = 25
-	penetration = 15
-	sundering = 1
+	damage = 20
+	penetration = 10
+	sundering = 1.5
 
 /datum/ammo/bullet/smartgun/lethal
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -916,15 +916,15 @@
 
 /obj/item/weapon/gun/rifle/standard_smartrifle
 	name = "\improper T-25 smartrifle"
-	desc = "The T-25 is the TGMC's current standard IFF-capable  rifle. It's known for its ability to lay down quick fire support very well. Requires special training and it cannot turn off IFF. It uses 10x26mm ammunition."
+	desc = "The T-25 is the TGMC's current standard IFF-capable rifle. It's known for its ability to lay down quick fire support very well. Requires special training and it cannot turn off IFF. It uses 10x26mm ammunition."
 	icon = 'icons/Marine/gun64.dmi'
 	icon_state = "t25"
 	item_state = "t25"
 	caliber = CALIBER_10x26_CASELESS //codex
-	max_shells = 50 //codex
+	max_shells = 80 //codex
 	force = 35
-	aim_slowdown = 0.55
-	wield_delay = 0.6 SECONDS
+	aim_slowdown = 0.7
+	wield_delay = 0.9 SECONDS
 	fire_sound = "gun_smartgun"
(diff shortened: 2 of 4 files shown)


Lefunny:
t-25 is perfect balanced for now, theres no reason for this changs, a weapon that let marines outspeed some xenos and melt their health is supposed to be like this and good to the game.

Hekzder:
> t-25 is perfect balanced for now, theres no reason for this changs, a weapon that let marines outspeed some xenos and melt their health is supposed to be like this and good to the game.

Completely disregarding whether or not the T-25 may or may not be balanced (it's not), a weapon like this for the supportive fire role is completely silly. It's literally just a power creeped T-12 that people use to rush Xenos with the added bonus of IFF, this is in no way a supportive weapon.

bowlofcereal:
> t-25 is perfect balanced for now, theres no reason for this changs, a weapon that let marines outspeed some xenos and melt their health is supposed to be like this and good to the game.

troll account

turnop:
i'm ok with this, this is really more of a t25 nerf than lmg rebalance but t25 is too strong in its current state imo

BraveMole (approved):
T25 is a powercreped T12 now, this is more in phase with what sg is supposed to be

TiviPlus (approved):

PR #8008, opened 2021-08-22
Makes buckshot a bit more effective mid range

SplinterGP:
## About The Pull Request
increases the buckshot accurate range to 5 and reduces falloff by 1

## Why It's Good For The Game
Buckshot is kinda weird in a way that it is very good(only against low armor classes) at PB and trash at everything else, its supposed to be bad as range goes on but, not only it spreads, hitting less pellets due to it, but it also loses a lot of damage and misses alot after 3 tiles, now this maxes it hit more at a higher range but it will still hit 1 or 2 pellets due to spread anyway

## Changelog
:cl:
balance: buckshot has less falloff
balance: buckshot accurate range(not spread) has been increased to 5 tiles.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -694,7 +694,7 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range = 5
 	max_range = 10
 	damage = 40
-	damage_falloff = 3
+	damage_falloff = 1
 	penetration = 0
 
 
@@ -710,7 +710,7 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range = 5
 	max_range = 10
 	damage = 40
-	damage_falloff = 3
+	damage_falloff = 1
 	penetration = 0
 
 //buckshot variant only used by the masterkey shotgun attachment.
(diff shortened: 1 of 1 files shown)


RenGusta:
So it loses 15 instead of 20 damage a tile? Still not great, but hey. We'll see if someone likes it soon.

Lefunny:
Making a buckshot ((weapon that is good used as meele)) being able to be used now as middle range sounds like a amazing idea for me, surely most people with shotguns cant handle meele fight agaist most t1s this should help with letting shotungs being more ranged, good chang!

EvilMelonTheSecond:
Still worse than flechette in 99% of situations. The 1% is somehow PBing a runner which MIGHT kill it.

Pariah919:
it still loses 15 damage per tile which effectively means all this does is make it slightly better when you pb on something with knockback
reduce the spread a bit and reduce the falloff by 3 rather than 1.

lovegreenstuff:
buck is still going to have absolutely garbage damage potential past 1 tile due to scatter
you're gonna go from doing 32 damage at most to 34 every two seconds

Pariah919:
> Making a buckshot ((weapon that is good used as meele)) being able to be used now as middle range sounds like a amazing idea for me, surely most people with shotguns cant handle meele fight agaist most t1s this should help with letting shotungs being more ranged, good chang!
>
i haven't seen a past pb tile buckshot kill since i got a 6 tile on a hivelord.. With a unnerfed ZX.
Which means it was like, a year and a half ago?

lovegreenstuff:
even less falloff wow.
i still have 17.5 dps at 5 tile range. a whopping 7.5 dps bonus woo.

BraveMole (approved):
Ok let's try that

PR #8011, opened 2021-08-22
Adds stasis check to living

QualityVan:
## About The Pull Request
Skips processing of status effects tracked on mob (not datum ones), health updates (shouldn't be changing if they're in stasis), and organs (which then prevents chems metabolizing).

## Why It's Good For The Game
Apparently you still heal through chems while in a stasis bag. This fixes that.
A stasis check could be added to carbon's life too, but it's not necessary because the only thing carbon does is breath and that independently checks stasis.

## Changelog
:cl:
fix: Stasis bags prevent metabolizing chemicals
fix: Some negative status effects will be paused while in stasis, resuming once you leave
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1,5 +1,5 @@
 /mob/living/proc/Life()
-	if(stat == DEAD || notransform) //If we're dead or notransform don't bother processing life
+	if(stat == DEAD || notransform || HAS_TRAIT(src, TRAIT_STASIS)) //If we're dead or notransform don't bother processing life
 		return
 
 	handle_status_effects() //all special effects, stun, knockdown, jitteryness, hallucination, sleeping, etc


BraveMole (approved):

code/modules/mob/living/living.dm
```diff
@@ -1,5 +1,5 @@
 /mob/living/proc/Life()
-	if(stat == DEAD || notransform) //If we're dead or notransform don't bother processing life
+	if(stat == DEAD || notransform || HAS_TRAIT(src, TRAIT_STASIS)) //If we're dead or notransform don't bother processing life
 		return
 
 	handle_status_effects() //all special effects, stun, knockdown, jitteryness, hallucination, sleeping, etc
```
TiviPlus, line 5: This lets you stunlock someone inside a stasis bag forever with no hope of getting out

TiviPlus (changes requested):

code/modules/mob/living/living.dm
QualityVan: All incapacitating status effects are handled by datums, which aren't affected by this.

BraveMole: what about stamina loss?

QualityVan: It won't cause new disables while you're bagged and any existing ones will fall off.

TiviPlus: stamina loss will keep you stunned yes because its handled in status effects

QualityVan: Stamloss's stun is from living's updatestamina. This is called by adjust_stamina, called by handle_staminaloss, called by handle_status_effects. If you're in stasis, this won't happen. Stamloss's paralyze is a status_effect datum, which as I said before will expire even in stasis.

BraveMole (approved):

PR #8018, opened 2021-08-23
Fix fire light lag

BraveMole:
## About The Pull Request

Fix the lag, while keeping fancy

## Why It's Good For The Game

Players can play game

## Changelog
:cl:
fix: Fix fire light lag
/:cl:


Losenis:
Fancy lighting as a whole had a clear fps loss. Fire was the worst of them, not the only one.  A mandatory purely cosmetic change that traded performance you can't even toggle off should've been an insane enough idea to not even get off the drawing board.

PR #8020, opened 2021-08-23
Jetpack are now free

BraveMole:
## About The Pull Request

You can get jetpacks from the utility tab now. 5 of them free, if this is not too much i'll make it widely avaiblable

## Why It's Good For The Game

Jetpacks are rarely used, despite their low price in req. This is because a lot of the time, you need this extra ammo in the bag, and jetpack is not that interesting in your loadout.

Putting the jetpack for free at the start of the game will allow players to experiment with new builds, and i don't think jetpack is that strong against xenos

## Changelog
:cl:
balance: 3 jetpacks free in vendors
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -229,7 +229,7 @@
 		/obj/item/compass = -1,
 		/obj/item/assembly/signaler = -1,
 		/obj/item/weapon/gun/launcher/m81/flare/marine = -1,
-		/obj/item/jetpack_marine = 5,
+		/obj/item/jetpack_marine = 3,
 		),
 	)
 
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -229,6 +229,7 @@
 		/obj/item/compass = -1,
 		/obj/item/assembly/signaler = -1,
 		/obj/item/weapon/gun/launcher/m81/flare/marine = -1,
+		/obj/item/jetpack_marine = -1,
 		),
 	)
 
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -229,7 +229,7 @@
 		/obj/item/compass = -1,
 		/obj/item/assembly/signaler = -1,
 		/obj/item/weapon/gun/launcher/m81/flare/marine = -1,
-		/obj/item/jetpack_marine = -1,
+		/obj/item/jetpack_marine = 5,
 		),
 	)
 
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -938,6 +938,11 @@ CLOTHING
 	)
 	cost = 5
 
+/datum/supply_packs/clothing/jetpack
+	name = "Jetpack"
+	contains = list(/obj/item/jetpack_marine)
+	cost = 12
+
 /*******************************************************************************
 MEDICAL
 *******************************************************************************/
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -938,11 +938,6 @@ CLOTHING
 	)
 	cost = 5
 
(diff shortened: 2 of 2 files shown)


Lefunny:
Giving marines a way to escape from meele castes ((most castes)) and punishing xenos for trying to get close sounds like...A amazing and balanced idea now making it for free? still sounds amazing xenos shouldt be allowned to get close to marines and marines really need more ways to show xenos how getting in their meele inst rewarding.

Hekzder:
12 points to free seems a bit excessive. Why not try reduction to like 5 points or something? It puts it on the same level as imagers for example.

february-ops:
Jetpacks are incredibly powerfull and can save you in most situations it should not be free at all, backpacks at not even that useful with hou much storage you have marines not buying is totally a skill issue and its their loss they dont buy it, you worked hard on jetpacks and not seeing it used mus be frustating but is not a reason to make it free

BraveMole:
> Jetpacks are incredibly powerfull and can save you in most situations it should not be free at all, backpacks at not even that useful with hou much storage you have marines not buying is totally a skill issue and its their loss they dont buy it, you worked hard on jetpacks and not seeing it used mus be frustating but is not a reason to make it free

You're most probably right, i'm biaised. Despite that, when playing with a jetpack i really can feel the second gun or ammo missing, so it's a compromise.

When jetpack was free (back in the TM days), i didn't have that much issues against those as a xeno, on the contrary it was pretty funny to play against

Passe2142:
I dig the idea, but uhh they could be limited like flamers and such if peeps dont like it too much.
Come get your free jetpacks limited edition now in greyscale too. 
Jokes aside i dont see why not , i played during the TM and barely use it. Its niche and it can be fun to see what new builds will be.

EvilMelonTheSecond:
I've been wanting to do some jetpack builds lately and have found my req requests to be largely declined. Even if it isn't made totally free a few should be up for grab since its such a fun item that isn't a straight up armor/damage increase.

Pariah919:
I don't agree with this since it reeks of powercreep but i can see where you're coming from overall
personally i think i'd rather test making jetpacks cheaper than making them free outright

lKiseki:
I agree with MJP here. Doesn't seem like this'll be balanced at all.

SplinterGP:
i  don't think making jetpacks free is good anyhow, they might be underused but thats mostly because of popularity, the combat backpack is a straight upgrade but it still is underused

GUDmood:
People don't use jetpack because they don't see it much or outright don't know it exist.
Adding 2-5 free jpacks to the vendors sounds like a good plan, so marines can try it out without begging RO.

code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -229,6 +229,7 @@
 		/obj/item/compass = -1,
 		/obj/item/assembly/signaler = -1,
 		/obj/item/weapon/gun/launcher/m81/flare/marine = -1,
+		/obj/item/jetpack_marine = -1,
```
BraveMole, line 232:
```suggestion
		/obj/item/jetpack_marine = 5,
```

bowlofcereal:
Reason we dont see any of these things is because they were exiled to requisitions during their most popular days and were basically never heard of again. I don't think adding a few free ones back to requisitions or just making them free in general is a bad thing since a good majority of people dont even seem to realize they exist anymore.

David-Stormwell:
I suggest limiting it to 2-3 max.
New people can try it out, and buy more if needed.

code/modules/reqs/supplypacks.dm
```diff
@@ -938,11 +938,6 @@ CLOTHING
 	)
 	cost = 5
 
-/datum/supply_packs/clothing/jetpack
-	name = "Jetpack"
-	contains = list(/obj/item/jetpack_marine)
-	cost = 12
```
SplinterGP, line 944: ????????????????

SplinterGP (changes requested):
honestly, we have so much free shit now like 2 or 3 of them on marine vendors that might be a good idea to start putting all of them under a SL tab or something so they can choose BETWEEN which ones they want for free.

TiviPlus (approved):

PR #8026, opened 2021-08-23
Fix larva and king harddel

BraveMole:
Fix those harddels, that are coming from the fact those xeno are initiated in null space

Also maybe fix some acid harddel


code/modules/mob/living/carbon/xenomorph/hive_datum.dm
```diff
@@ -160,7 +160,8 @@
 /datum/hive_status/proc/add_to_lists(mob/living/carbon/xenomorph/X)
 	xenos_by_tier[X.tier] += X
 	xenos_by_upgrade[X.upgrade] += X
-	LAZYADD(xenos_by_zlevel["[X.z]"], X)
+	if(X.z != 0)
```
BraveMole, line 163:
```suggestion
	if(X.z)
```

TiviPlus (approved):

PR #8028, opened 2021-08-23
Enable live ref tracking.

BraveMole:
This will kill server, but it's for the greater good.

Enable ref tracking when a harddel is detected. This will slow down server to a crawl, sadly. We are doing this because harddel are the bane of performances, and it will help us greatly to fix those


Ieteihowmee:
Optimization good in the longer term.

code/_compile_options.dm
```diff
@@ -65,7 +65,7 @@
 //#define REFERENCE_TRACKING_DEBUG
 
 ///Run a lookup on things hard deleting by default.
-//#define GC_FAILURE_HARD_LOOKUP
+#define GC_FAILURE_HARD_LOOKUP
 #ifdef GC_FAILURE_HARD_LOOKUP
 #define FIND_REF_NO_CHECK_TICK
```
LemonInTheDark, line 70: You're running this with no check tick? madman

BraveMole: Oups, fixed

code/modules/admin/view_variables/reference_tracking.dm
```diff
@@ -1,6 +1,14 @@
 #ifdef REFERENCE_TRACKING
 
+GLOBAL_LIST_INIT(reftracking_blacklisted_types, list(
+	/obj/effect/alien/weeds, //This is fixed in weeds refactor
+))
```
BraveMole, line 5:
```suggestion
GLOBAL_LIST_EMPTY(reftracking_blacklisted_types)
```

Al-1ce:
>break the game even more
>to fix the game
yeah no

PR #8036, opened 2021-08-24
Removes an extra argument from mob reagent reaction

QualityVan:
## About The Pull Request
Reaction_mob has an argument named metabolize. It's not used in the declaration or in any children, and I don't know what it would do given that it's called when a reagent comes into contact with a mob. Metabolization is handled by on_mob_life, so that's not it. The only thing that calls reaction_mob is the holder's reaction proc, and it's never used it.

## Why It's Good For The Game
Cleanup.
Also fixes acid hiding the message when it melts your mask/helmet, due to incorrect argument order.

## Changelog
:cl:
fix: Acid will now tell you when it melts your helmet or mask (sulphuric/polytrinic, not xeno acid)
/:cl:

### DIFF
--- a/code/modules/reagents/reagents.dm
+++ b/code/modules/reagents/reagents.dm
@@ -70,7 +70,7 @@ GLOBAL_LIST_INIT(name2reagent, build_name2reagent())
 	return ..()
 
 /// Applies this reagent to a [/mob/living]
-/datum/reagent/proc/reaction_mob(mob/living/L, method = TOUCH, volume, metabolism, show_message = TRUE, touch_protection = 0)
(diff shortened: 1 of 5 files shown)


BraveMole (approved):

SplinterGP (changes requested):
did you check on the calls of it if youre not missing something on order ?

QualityVan:
> 
> 
> did you check on the calls of it if youre not missing something on arg order ?

Like I said, it's only called in one place and that one place wasn't accounting for it, which this fixes.

BraveMole (approved):
looks good

PR #8051, opened 2021-08-25
Reagent Slash is now locked when the ability is activated

Surrealaser:
## About The Pull Request

Reagent Claws is now locked to injecting the reagent selected when it is activated.

## Why It's Good For The Game

Prevents Defilers from getting inter-toxin synergies far too easily; doubly important since the amount of U per reagent slash has been significantly increased.

## Changelog
:cl:
balance: Reagent Claws is now locked to injecting the reagent selected when it is activated.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
@@ -267,7 +267,8 @@
 	var/reagent_slash_count = 0
 	///Timer ID for the Reagent Slashes timer; we reference this to delete the timer if the effect lapses before the timer does
 	var/reagent_slash_duration_timer_id
-
+	///Defines the reagent being used for reagent slashes; locks it to the selected reagent on activation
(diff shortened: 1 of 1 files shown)


lKiseki:
good bye my beloved
i will miss you so

QualityVan:
Bad change. If you switch you're choosing sacrificing duration for short term threat since each chem metabolizes independently. Locking each activation means that you need to be in combat for at least twelve seconds before you hit trinity, and by that time either you or them are dead. If new defile is too strong with triple chem that should be addressed independently.

SplinterGP (approved):

BraveMole (approved):
Good change. You can still change chem durring puffing IIRC, or was it changed?

Surrealaser:
> Good change. You can still change chem durring puffing IIRC, or was it changed?

Changed awhile ago

EvilMelonTheSecond:
I loved having rainbow claws. Guess its back to using only neuro.

Passe2142:
It will be missed , we get in exchange the fancy rework so this is kinda needed.

PR #8053, opened 2021-08-25
Smoke affects you when you step into it

QualityVan:
## About The Pull Request
Makes smoke apply its affects when a mob crosses it, in addition to the usual 2 second process tick. Caps the frequency at a value set per smoke type but defaulting at 1 second.
Requested by Zack.
~~Chem smoke has its own internal limit of once per .4s limit per cloud.~~ Not how it works, actually.
Also fixes stacked smoke clouds not both applying effects. Apparently that was broken.

## Why It's Good For The Game
Makes just sprinting through a smoke cloud less free. ~~You still only get about 12 volume of reagents (6 each light and standard neuro) running the full length of a defiler neuro cloud with a gas mask on, but without this change it was about 1.2u.~~

## Changelog
:cl:
balance: Smoke clouds can apply their effects when you walk into one.
fix: Stacked smoke clouds will both apply their effects instead of only one.
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -20,6 +20,8 @@
 	var/datum/effect_system/smoke_spread/cloud // for associated chemical smokes.
 	var/fraction = 0.2
 	var/smoke_can_spread_through = FALSE
+	///Delay in ticks before this smoke can affect a given mob again, applied in living's effect_smoke
+	var/minimum_effect_delay = 1 SECONDS
 
 	//Remove this bit to use the old smoke
 	icon = 'icons/effects/96x96.dmi'
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -77,8 +77,9 @@
 
 /obj/effect/particle_effect/smoke/Crossed(atom/movable/O)
 	. = ..()
-	if(CHECK_BITFIELD(smoke_traits, SMOKE_CAMO) && isliving(O))
+	if(isliving(O))
 		O.effect_smoke(src)
+		return
 	if(CHECK_BITFIELD(smoke_traits, SMOKE_NERF_BEAM) && istype(O, /obj/projectile))
 		O.effect_smoke(src)
 
--- a/code/modules/mob/living/living_defines.dm
+++ b/code/modules/mob/living/living_defines.dm
@@ -97,7 +97,7 @@
 	var/reagent_pain_modifier = 0 //same as above, except can potentially mask damage
 
 	///Lazy assoc list of smoke type mapped to the next world time that smoke can affect this mob
-	var/smoke_delays
+	var/list/smoke_delays
 	var/smokecloaked = FALSE //For the new Smoke Grenade
 
 	var/no_stun = FALSE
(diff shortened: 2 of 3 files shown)


BraveMole:
This looks like a huge boiler and defiler buff

QualityVan:
> 
> "Chem smoke has its own internal limit of once per .4s limit per cloud." yeah, so it's not per mob, so if you quicky cross 5 tiles to get out of the smoke, it's like you stayed 10 secs in the cloud

For comparison, xeno clouds have a lifespan of 6 ticks. The largest xeno clouds are from ancient boiler globs and violently destroyed transvit eggs, conveniently also at six range. In other words, leaving is never worse than staying inside, but crossing from one side to the other at the widest point usually will be. And that's in the worst case scenario: getting hit by an ancient boiler's glob.

You're right about chem smoke though, thought that was unique adds.

QualityVan:
I'd like to make three arguments in favor of this as is, at least for a test merge. Maybe it'd be grossly overpowered and unsuitable despite them, but hey.

First: This change means gassers get guaranteed minimum value out of actually landing a cloud on top of someone. Right now, even if you manage to stick someone in the center they'll just swooce out in less than one gas tick unless they're seriously disabled. 

Second: Rather than negating the counterplay to clouds, I'd argue it actually increases your options. Before it was either stop being in the cloud (basically always right) or stay in the cloud (why). With this change, your options are stand still in the cloud to essentially hold your breath and maintain position but take the downside of standing still, move in the cloud to maintain both position and mobility but take extra smoke ticks as penalty, or leave the cloud to frontload your smoke hits and sacrifice position, but also minimize smoke while keeping mobility.

Third: This makes gas much more effective as a way to deter people just running through. Right now it doesn't do that at all - if you're at high speed you'll take one tick at most and be on your merry way. Contrast with fire, another dedicated temporary area denial option, which heavily punishes passing through with both extra fire stacks and damage per tile.

InterroLouis:
Will this make WP smoke more effective against xenos?

code/modules/mob/living/living_defines.dm
```diff
@@ -96,7 +96,8 @@
 	var/reagent_shock_modifier = 0 //negative values reduce shock/pain
 	var/reagent_pain_modifier = 0 //same as above, except can potentially mask damage
 
-	var/smoke_delay = FALSE
+	///Lazy assoc list of smoke type mapped to the next world time that smoke can affect this mob
+	var/smoke_delays
```
BraveMole, line 100:
```suggestion
	var/list/smoke_delays
```

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -77,8 +79,9 @@
 
 /obj/effect/particle_effect/smoke/Crossed(atom/movable/O)
 	. = ..()
-	if(CHECK_BITFIELD(smoke_traits, SMOKE_CAMO) && isliving(O))
+	if(isliving(O))
 		O.effect_smoke(src)
+		return
 	if(CHECK_BITFIELD(smoke_traits, SMOKE_NERF_BEAM) && istype(O, /obj/projectile))
```
BraveMole, line 85: Is this still needed?

BraveMole (changes requested):

code/game/objects/effects/effect_system/smoke.dm
QualityVan: It's kind of a niche case, but yeah. Shooting energy weapons, including lasers, though certain smoke clouds cuts their damage. M40 HSDPs use that smoke, don't think anything else does.

BraveMole (approved):

PR #8074, opened 2021-08-26
Flamernerf

jroinc:
## About The Pull Request

Removes spec flamers, adds them to req for 30 pts. Replaces them with the OG PFC flamer.

Adds extinguishers and miniextinguishers to req.

## Why It's Good For The Game

Requested by headmin, approved by project lead.

Flamers are currently OP. They can burn down a hive in under a minute with a 500-ammo tank, and there's no consequences for their use because they can just extinguish instantly. People spray fire, extinguish it, then walk forward again without having to wait for the fire to die out, which makes it impossible to build fortifications if they disappear at *walking speed*.

The extinguisher has so high a capacity that I've never seen anyone run out, and I don't believe it *has* a reload function. It's also a one-click ungaspeed button so that if, heaven forbid, you accidentally make a mistake and light a unga up, you can one-click remove it and one-click get back to flaming, so they don't even take burn damage.

This replaces the zoom zoom extinguisher one with the old PFC one. It has one tile less range, 15 less ammo, a faster wield time, and does not mount the extinguisher. The spec one has been added to req, 30 points.

Miniextinguishers and regular extinguishers are added to gun vendors. "But how are these different, they're the same!" Three ways-

1-Reduces response times. You have to *commit* to putting fire out now, it's not a three-click thing that takes half-a-second. Yes, I know, you don't want to be on fire and you want the instant extinguish. Balance is a cruel mistress, and more ungas burning to death cause their teammates ignored them *is acceptable* for this.
2-Requires not being wielded for the instant extinguish. This reduces unga firerates and opens up vulnerabilities.
3-Has an actual ammo limit (around 10 for the regular extinguisher, 6 for the mini).

## Changelog
:cl:
add: adds extinguishers and miniextinguishers to vendors
balance: rebalanced flamers via req
/:cl:


Pariah919:
#8095 fixes most of the issues here
don't see a point anymore for this really

BraveMole:
With the addition of the miniflamer, and MJP's change to walls, i don't think this is needed. 

Flamers were already nerfed against walls, and if you remove the hydro cannon from it there's basically no reason to not take a miniflamer

PR #8076, opened 2021-08-27
Reduces buckshot extra projectile scatter

Pariah919:
## About The Pull Request
Buckshot other projectile scatter from 10->4

## Why It's Good For The Game
Buckshot is fucking trash, it literally loses all of it's effectiveness past tile ONE.
Tile. ONE.
Now I've tested this and this makes buckshot effective until tile 3 rather than just tile ONE.
Buckshot in general isn't that attractive, this should make it generally more actually viable at it's CQC role since rn it's literally useless past using it as a PB stick, and that point just use flechette so you can shoot the xeno as it approaches you.

## Changelog
:cl:
balance: Buckshot extra projectiles have less scatter.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -688,7 +688,7 @@ datum/ammo/bullet/revolver/tp44
 	flags_ammo_behavior = AMMO_BALLISTIC
 	bonus_projectiles_type = /datum/ammo/bullet/shotgun/spread
 	bonus_projectiles_amount = 5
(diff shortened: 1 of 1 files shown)


David-Stormwell:
Should be fine, as range has always been a problem for it, flechette still has the ability to surpass it in longer ranges and all generalness, but this should be able to hold it's ground then.

David-Stormwell (approved):
Last comment

BraveMole (approved):
Seems balanced enough and david said yes

PR #8094, opened 2021-08-28
Fixes runtime with prison

lbnesquik:
## About The Pull Request
Fixes reported runtimes on prison regarding the turrets.

## Changelog
:cl:
fix: fixed Prison Runtimes regarding sentries
/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-8085.yml
+++ b/html/changelogs/AutoChangeLog-pr-8085.yml
@@ -1,0 +1,4 @@
+author: "TiviPlus"
+delete-after: True
+changes: 
+  - bugfix: "Some messages will now use pronouns correctly instead of their/they"


hyper2snyper:
👏 thanks man.

For reference ima explain how you broke it and how map based sentries should be setup in the future.

So sentries are now guns and not machines. The /mounted/sentry machine you placed is a generic type that acquires all its data from the item that sets it up. In this case you were spawning the generic type alone and so it had no way of knowing what it should be or how to function. The reason the unit tests didnt fail were because it was a map change. 

Since sentries are now guns the way you place a map based sentry is by spawning the /gun item where you want the sentry. If it has the DEPLOY_ON_INITIALIZE flag it will deploy itself once spawned. All the sentry subtypes that have premade in the type name have the flag. So spawning the premade/dumb will be placing already deployed turrets.

Currently they dont have a configurable direction you can set with the map though I intend to fix that in the near future.

hyper2snyper (approved):

BraveMole (approved):

BraveMole:
No dir? dumb turrets are on radial by default?

TiviPlus (approved):

hyper2snyper:
> No dir? dumb turrets are on radial by default?

Yes. All premade ones are except the tadpole and fob ones.

PR #8095, opened 2021-08-28
Resin wall rebalance

Pariah919:
## About The Pull Request
This PR reduces the amount of health walls have to 300.
Reduces the tick damage of flamers from 50 to 15.
Gives walls 60% bullet soft armor, so they have the same general health overall.
Walls spawn stronger as well, +50 HP to thick and normal walls.
Changes Ex_Act defines as well, other than devastation.

## Why It's Good For The Game
This makes melee a more effective option of dealing with walls while keeping bullets not quite as effective against walls, reduces the effectiveness of a flamer in killing a wall, gives melee weapons a decent niche beyond being trash memes.

This also makes walls that are recently made, much much better.

## Changelog
:cl:
balance: Wall rework, max HP lowered, they've been given armor against bullets/lasers so their TTK is about the same, they're less effective against fire, high armor piercing weapons, explosions are about the same, but fire is much worse at killing walls quickly, but it still kills a lot of walls at the same time.
/:cl:

### DIFF
--- a/code/game/turfs/walls/resin.dm
+++ b/code/game/turfs/walls/resin.dm
@@ -12,7 +12,7 @@
 	max_integrity = 200
 	layer = RESIN_STRUCTURE_LAYER
 	tiles_with = list(/turf/closed/wall/resin, /turf/closed/wall/resin/membrane, /obj/structure/mineral_door/resin)
-	soft_armor = list("melee" = 0, "bullet" = 60, "laser" = 0, "energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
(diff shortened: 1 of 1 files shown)


David-Stormwell:
Approved, this is quite what we needed.
It solves the problem for those Xeno mains who complain about how flamers are clearing walls left and right while giving Marines the option to use melee on them.

David-Stormwell (approved):
See above comment.

Derrick2Electricboogaloo:
This is fair

BraveMole:
You did some math so it is roughly equivalent or you just did like you felt it ?

BraveMole (approved):
good changes

PR #8100, opened 2021-08-28
you can slash flashlights off

BraveMole:
## About The Pull Request

Make flashlights slashable. 

## Why It's Good For The Game

image
image

No thanks. Also makes sense since you can slash flashlights on gun

## Changelog
:cl:
balance: you can slash flashlights off
/:cl:

### DIFF
--- a/code/game/objects/items/flashlight.dm
+++ b/code/game/objects/items/flashlight.dm
@@ -38,9 +38,8 @@
 	update_icon()
 
 /obj/item/flashlight/attack_alien(mob/living/carbon/xenomorph/X, isrightclick = FALSE)
-	if(!light_on)
-		return ..()
-	turn_light(X, FALSE)
+	if(turn_light(X, FALSE) != CHECKS_PASSED)
+		return
 	playsound(loc, "alien_claw_metal", 25, 1)
 	X.do_attack_animation(src, ATTACK_EFFECT_CLAW)
 	to_chat(X, span_warning("We disable the metal thing's lights.") )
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

code/game/objects/items/flashlight.dm
```diff
@@ -37,6 +37,12 @@
 	update_action_button_icons()
 	update_icon()
 
+/obj/item/flashlight/attack_alien(mob/living/carbon/xenomorph/X, isrightclick = FALSE)
+	turn_light(X, FALSE)
```
QualityVan, line 41:
```suggestion
/obj/item/flashlight/attack_alien(mob/living/carbon/xenomorph/X, isrightclick = FALSE)
	if(!light_on)
		return ..()
	turn_light(X, FALSE)
```

BraveMole: image

QualityVan: yeah but you'll get the attack animation and to_chat regardless

EvilMelonTheSecond:
Aren't they just flares that are harder to carry? I don't see a problem with flashlight spam, melt em like flares.

bowlofcereal:
I'm fine with this but I just gotta say that the hivelord in that image is a total dunce if he's not gonna even try to melt them

Losenis:
> Aren't they just flares that are harder to carry? I don't see a problem with flashlight spam, melt em like flares.

They have the same slot size and are permanent so long as they don't get acided, as well as being infinite throught the vendor. A backpack of them will essentially last you the op since you can also just go and pick up whatever flashlight is lying around not being useful.

TiviPlus (approved):

PR #8121, opened 2021-08-29
Refactor eggs

BraveMole:
## About The Pull Request

Refactor eggs. Might have small impact on balance, but i've tried to make it minimal

## Why It's Good For The Game

I'm making over types of eggs, i need this

## Changelog
:cl:
refactor: Refactor eggs
/:cl:


code/modules/mob/living/carbon/xenomorph/egg.dm
```diff
@@ -0,0 +1,197 @@
...
+	qdel(facehugger)
+	advance_maturity(stage_ready_to_burst)
+	return TRUE
+
+/obj/effect/alien/egg/gas
+	desc = "It looks like a weird egg"
+	name = "Hugger egg"
+	icon_state = "egg_hugger"
+	maturity_time = 15 SECONDS
+	stage_ready_to_burst = 1
```
QualityVan, line 171: Right now gas eggs start out mature, since they already become mature while they're hugger eggs pre-gas. I'd say just set maturity_stage in the define but you're using advance to set the turf triggers. Maybe check whether mat_stage == burst_stage in init?

Kuro020 (changes requested):
image

Also these sprites already exist in the cimex folder on our repo

code/modules/mob/living/carbon/xenomorph/egg.dm
```diff
@@ -0,0 +1,202 @@
+/obj/effect/alien/egg
```
SplinterGP, line 1: i...wonder what's the thing about this being a child of effect rather than a structure or something like that ?

code/modules/mob/living/carbon/xenomorph/egg.dm
```diff
@@ -0,0 +1,202 @@
...
+	if(!should_proc_burst(entered))
+		return
+	burst()
+
+/obj/effect/alien/egg/proc/should_proc_burst(mob/living/carbon/carbon_mover)
+	if(issamexenohive(carbon_mover))
+		return FALSE
+	if(carbon_mover.stat == DEAD)
+		return FALSE
+	return TRUE
```
SplinterGP, line 61: this should probably use the hugger inside or whatever proc to check if it should attack the enemy, this way its simpler for multiple hugger types.

SplinterGP (changes requested):

code/modules/mob/living/carbon/xenomorph/egg.dm
BraveMole: that requires instantiating the hugger and keeping it in content. Which is really meh

code/modules/mob/living/carbon/xenomorph/egg.dm
BraveMole: I dunno, all alien things are like and i see no reason to change it

TiviPlus (approved):

PR #8131, opened 2021-08-31
Sterilizine disinfects wounds

QualityVan:
## About The Pull Request
Makes sterilizine disinfect external wounds when it's sprayed on a human. Wound germs exist distinctly from limb germs, so this will only indirectly help with an existing limb infection. 
Also makes it more likely to be painful, scaling with target's wounds, and replaced two mins in wound germ formula that I'm convinced were mistaken.

## Why It's Good For The Game
Right now sterilizine doesn't do anything space cleaner doesn't do better (except make people scream), but I don't feel like space cleaner needs toning down. This adds a separate use for sterilizine in the same vein without trivializing infected limbs.

## Changelog
:cl:
add: Sprayed sterilizine disinfects existing wounds. It won't help directly with infected limbs though.
balance: Spraying sterilizine on people with existing wounds is more likely to cause pain.
fix: Wounds no longer infect limbs more slowly when you have antibiotics
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -455,12 +455,12 @@ Note that amputating the affected organ does in fact remove the infection from t
 	for(var/datum/wound/W in wounds)
 		//Open wounds can become infected
 		if (owner.germ_level > W.germ_level && W.infection_check() && W.damage)
-			W.germ_level += min(1,round(W.damage/20)) //The bigger the wound, the more germs will enter
+			W.germ_level += max(1,round(W.damage/20)) //The bigger the wound, the more germs will enter
 
 		if (antibiotics < MIN_ANTIBIOTICS)
 			//Infected wounds raise the organ's germ level.
 			if (W.germ_level)
-				germ_level += min(1,round(W.germ_level/35))
+				germ_level += max(1,round(W.germ_level/35))
 		else if (W.germ_level && prob(80)) //Antibiotics wont be very effective it the wound is still open
 			germ_level++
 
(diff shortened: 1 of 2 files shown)


code/modules/reagents/reagents/other.dm
```diff
@@ -641,7 +641,11 @@
 	if(!(method in list(TOUCH, VAPOR, PATCH)))
 		return
 	L.germ_level -= min(volume * 20 * touch_protection, L.germ_level)
-	if((L.getFireLoss() > 30 || L.getBruteLoss() > 30) && prob(10)) // >Spraying space bleach on open wounds
+	if(ishuman(L))
+		var/mob/living/carbon/human/disinfectee = L
+		for(var/datum/limb/nub AS in disinfectee.limbs)
```
BraveMole, line 646: nub?

code/modules/organs/limbs.dm
```diff
@@ -460,7 +460,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 		if (antibiotics < MIN_ANTIBIOTICS)
 			//Infected wounds raise the organ's germ level.
 			if (W.germ_level)
-				germ_level += min(1,round(W.germ_level/35))
+				germ_level += max(1,round(W.germ_level/35))
```
BraveMole, line 463: a max is not consistent with the next code.

If you have no antibiotics (with current code), you can get from 1 to infinity germ level. With antibiotics, it's 1.

With your code, having no antibiotics is better than having some, since it's between 0 and 1 without antibiotic and 1 with

BraveMole (changes requested):

code/modules/organs/limbs.dm
QualityVan: Nah, you got it backwards. Old code: no antibiotics then 0 to 1, yes antibiotics then probably 1. New code: no anti then 1 to infinity, yes anti then probably 1.  min(1, x) isn't gonna give you more than one.

BraveMole: ah you're right

Issue #7926, opened 2021-08-14
Bombing sentries explode the internal item and make it null


2whatever2:
### Description

Read title

### Test Merges

#7735 #7527 #7534 #7725

### Reproduction Steps

1. Sentry gets bombed
2. Xeno walks in front of it
3. Can't interact with the sentry or pick it up at all.

### Screenshots

https://user-images.githubusercontent.com/24631139/129435796-458298c3-b112-4550-8a2d-80a09091fe41.mov

BraveMole:
@hyper2snyper fixed?

hyper2snyper:
> @hyper2snyper fixed?

Not yet tho coming real soon

BraveMole:
Fixed

PR #8133, opened 2021-08-31
Sentry Tweaks

hyper2snyper:
## About The Pull Request

Fixes Issue #7926 
Also fixes the issue where a sentry will continue blasting its last target while knocked down.

Puts deployed machine armor values as based on the thing it is deployed from.
Adds a list of typepaths that will be of turf the sentry ignores.
These are set to the miners, and other deployed items.

## Why It's Good For The Game

Fixes a pair of bugs. Makes setting armor values item based. Allows for ignoring terrain so we can have cool things in the future like laser sentries that shoot through glass.

## Changelog
:cl:
balance: Sentries can now shoot past other deployed items and Miners.
fix: Fixes the issue with null internal guns on sentries.
fix: Fixes Comrade sentry.
code: Updates /deployable and makes armor values dependent on the deploying item.
/:cl:

### DIFF
--- a/code/game/objects/machinery/deployable.dm
+++ b/code/game/objects/machinery/deployable.dm
@@ -4,6 +4,8 @@
 	///Since /obj/machinery/deployable aquires its sprites from an item and are set in New(), initial(icon_state) would return null. This var exists as a substitute.
 	var/default_icon_state
 
+	flags_atom = PREVENT_CONTENTS_EXPLOSION
+
 	hud_possible = list(MACHINE_HEALTH_HUD)
 
 /obj/machinery/deployable/Initialize(mapload, _internal_item, deployer)
@@ -16,6 +18,10 @@
 	icon = initial(internal_item.icon)
 	default_icon_state = initial(internal_item.icon_state) + "_deployed"
 	icon_state = default_icon_state
+
+	var/obj/internal_object = _internal_item
+	soft_armor = internal_object.soft_armor
+	hard_armor = internal_object.hard_armor
 
 	prepare_huds()
 	for(var/datum/atom_hud/squad/sentry_status_hud in GLOB.huds) //Add to the squad HUD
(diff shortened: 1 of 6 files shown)


code/game/objects/machinery/deployable.dm
```diff
@@ -4,6 +4,8 @@
 	///Since /obj/machinery/deployable aquires its sprites from an item and are set in New(), initial(icon_state) would return null. This var exists as a substitute.
 	var/default_icon_state
 
+	flags_atom = PREVENT_CONTENTS_EXPLOSION
+
 	hud_possible = list(MACHINE_HEALTH_HUD)
```
TiviPlus, line 9: define new vars after redifining new ones

code/modules/projectiles/gun_system.dm
```diff
@@ -131,6 +131,8 @@
 	var/deploy_time = 0
 	///If the gun is deployable, the time it takes for the weapon to undeploy.
 	var/undeploy_time = 0
+	///List of turf/objects/structures that will be ignored in the sentries targeting.
+	var/list/ignored_terrains = list()
```
TiviPlus, line 135:
```suggestion
	///List of turf/objects/structures that will be ignored in the sentries targeting.
	var/list/ignored_terrains
```
lazylist

TiviPlus (changes requested):

code/game/objects/machinery/deployable.dm
```diff
@@ -17,6 +19,10 @@
 	default_icon_state = initial(internal_item.icon_state) + "_deployed"
 	icon_state = default_icon_state
 
+	var/obj/internal_object = _internal_item
```
BraveMole, line 22: internal item is already an obj?

BraveMole (changes requested):

PR #8138, opened 2021-09-01
Fixes zoom values.

SplinterGP:
## About The Pull Request
This fixes the issue with the View refactor PR that caused the zoom stuff to zoom out way more than before, because tivi thought that it somehow set the viewsize itself, while the proc that does that actually **_increases_** the viewrange, this changes the value and the proc called and makes it clear.

## Why It's Good For The Game
Fixes a issue

## Changelog
:cl:
fix: stuff that zooms now has their previous zoom out range that was accidentally changed on view datum refactor.
/:cl:


SplinterGP:
yes this will conflict with my other scope PR, i'll fix conflicts  when this happens

code/game/objects/items.dm
```diff
@@ -715,7 +715,7 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 		return
 
 	if(user.client)
-		user.client.view_size.set_view_radius_to(viewsize/2-2)//sets the viewsize to reflect radius changes properly
+		user.client.view_size.set_width_and_height(viewsize, viewsize)
 
 		var/tilesize = 32
```
TiviPlus, line 720: Does not support widescreen properly

code/modules/projectiles/gun_attachables.dm
```diff
@@ -759,15 +759,15 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
...
-	zoom_viewsize = 7
+	zoom_viewsize = 0
 
 
 /obj/item/attachable/scope/unremovable/tl102
 	name = "TL-102 smart sight"
 	desc = "An unremovable smart sight built for use with the tl102, it does nearly all the aiming work for the gun's integrated IFF systems."
 	attach_icon = "none"
 	zoom_offset = 3
-	zoom_viewsize = 7
```
TiviPlus, line 770: Why are you making zoom 0 here everywhere

TiviPlus (changes requested):

code/modules/projectiles/gun_attachables.dm
SplinterGP: because they were 7 because 7 = 7 tiles to each side of view + 1 of center, now this doenst make sense anymore and even less with widescreen if it comes for the future, this just instead tells how much tiles it will increase.

code/game/objects/items.dm
SplinterGP: it does, this var only sets the width and height var which ARENT the view size, they just increase/reduce the viewsize with its values, infact its most times at 0.

code/modules/projectiles/gun_attachables.dm
```diff
@@ -672,15 +672,17 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
...
+	attach_icon = "none"
+	zoom_viewsize = 0
 	zoom_tile_offset = 3
-	zoom_viewsize = 7
 
 
 /obj/item/attachable/scope/unremovable/tl102
 	name = "TL-102 smart sight"
 	desc = "An unremovable smart sight built for use with the tl102, it does nearly all the aiming work for the gun's integrated IFF systems."
+	attach_icon = "none"
```
SplinterGP, line 683:
```suggestion
```

SplinterGP:
```suggestion
```

code/modules/projectiles/gun_attachables.dm
```diff
@@ -672,15 +672,17 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	name = "MG-08/495 long range ironsights"
 	desc = "An unremovable set of long range ironsights for an MG-08/495 machinegun."
 	flags_attach_features = ATTACH_ACTIVATION
+	attach_icon = "none"
```
SplinterGP, line 675:
```suggestion
```

TiviPlus (approved):

PR #8141, opened 2021-09-01
Priority change in shuffling

BraveMole:
## About The Pull Request

As a xeno: if you are on help intent, you will never be blocked by another xeno

## Why It's Good For The Game

No more body blocking. Maybe more shuffling hell, i think it's a fair price to pay

## Changelog
:cl:
qol: As a xeno: if you are on help intent, you will never be blocked by another xeno
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -320,7 +320,7 @@
 	if(isliving(A))
 		var/mob/living/L = A
 
-		if(mob_size < L.mob_size && mob_size != MOB_SIZE_XENO) //Can't go around pushing things larger than us.
+		if(mob_size < L.mob_size) //Can't go around pushing things larger than us.
 			return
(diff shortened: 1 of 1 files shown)


SkeletalElite:
The marine one seems like it will create a lot of FF issues since you can't avoid getting shuffled by ungas who dont know not to be on help intent in the fight.

SplinterGP (changes requested):
the marine one is scuffed and will cause alot of FF

turnop:
Note that this will also remove pushdragging entirely and make it much harder to secure kills as a result.

BraveMole:
> Note that this will also remove pushdragging entirely and make it much harder to secure kills as a result.

No, this just ask the two xenos to be in harm intent rather than only one

TiviPlus (approved):

PR #8158, opened 2021-09-03
Lets you clear a single recipe from dispensers

QualityVan:
## About The Pull Request
As title. Hit the delete recipes button, choose 'only one', and then press the recipe you wanna clear.
Also fixes a bug with dispenser fumble checks.

## Why It's Good For The Game
Being able to clear a specific recipe is better than having to clear all recipes if you want to remove one and then readd all the others.

## Changelog
:cl:
add: You can choose to only remove a single chem recipe instead of all of them when using a dispenser.
fix: You can't start fiddling with a chem dispenser while you're busy doing something else.
/:cl:

### DIFF
--- a/code/modules/reagents/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/machinery/chem_dispenser.dm
@@ -168,6 +168,7 @@
 	.["recipes"] = user.client.prefs.chem_macros
 
 	.["recordingRecipe"] = recording_recipe
+	.["clearingRecipe"] = clearing_recipe
 
 /obj/machinery/chem_dispenser/ui_act(action, list/params)
 	. = ..()
@@ -177,6 +178,8 @@
 	if(needs_medical_training && ishuman(usr))
 		var/mob/living/carbon/human/user = usr
 		if(!user.skills.getRating("medical"))
+			if(user.do_actions)
+				return
 			to_chat(user, span_notice("You start fiddling with \the [src]..."))
 			if(!do_after(user, SKILL_TASK_EASY, TRUE, src, BUSY_ICON_UNSKILLED))
 				return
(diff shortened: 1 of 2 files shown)


code/modules/reagents/machinery/chem_dispenser.dm
```diff
@@ -249,8 +258,11 @@
 		if("clear_recipes")
 			if(!is_operational())
 				return
-			var/yesno = tgui_alert(usr, "Clear all recipes?", null, list("Yes","No"))
-			if(yesno == "Yes")
+			var/confirm_clear = tgui_alert(usr, "Clear all recipes?", null, list("Yes","No", "Only one"))
```
BraveMole, line 261: use a switch

BraveMole (changes requested):
Can't you add a button to directly clear the specific recipe, rather than this two steps thingy?

QualityVan:
> 
> 
> Can't you add a button to directly clear the specific recipe, rather than this two steps thingy?

If I do it this way I don't have to mess with tgui stuff. Also, I'm not sure about single-stage clearing. Would each recipe have its own remove button? You could clear by having the button let you input a recipe name, but that just seems less convenient than a two-button option.

SplinterGP:
I think this delete mode is better because having a delete button on the side of EACH recipe  is a bit annoying

code/modules/reagents/machinery/chem_dispenser.dm
```diff
@@ -222,6 +227,13 @@
 		if("dispense_recipe")
 			if(!is_operational() || QDELETED(cell))
 				return
+			if(clearing_recipe)
+				var/actually_clear = tgui_alert(usr, "Clear recipe [params["recipe"]]?", null, list("Yes","No"))
+				if(actually_clear == "Yes")
```
BraveMole, line 232:
```suggestion
				if(tgui_alert(usr, "Clear recipe [params["recipe"]]?", null, list("Yes","No")  == "Yes"))
```

code/modules/reagents/machinery/chem_dispenser.dm
```diff
@@ -222,6 +227,13 @@
...
+					usr.client.prefs.chem_macros.Remove(params["recipe"])
+					usr.client.prefs.save_preferences()
+				clearing_recipe = FALSE
```
BraveMole, line 235: eh, you have a clear recipe button to remove that already

code/modules/reagents/machinery/chem_dispenser.dm
```diff
@@ -249,14 +261,21 @@
...
 				return
-			var/yesno = tgui_alert(usr, "Clear all recipes?", null, list("Yes","No"))
-			if(yesno == "Yes")
-				usr.client.prefs.chem_macros = list()
-				usr.client.prefs.save_preferences()
+			if(clearing_recipe)
+				clearing_recipe = FALSE
+				return TRUE
+			var/confirm_clear = tgui_alert(usr, "Clear all recipes?", null, list("Yes","No", "Only one"))
+			switch(confirm_clear)
```
BraveMole, line 268: ditto, useless var

BraveMole (changes requested):

BraveMole (approved):

PR #8167, opened 2021-09-03
seasonals change every 24 hours rather than 4 days

Pariah919:
## About The Pull Request
24 hours to refresh from 4 days

## Why It's Good For The Game
we have enough seasonals that we can refresh far more often IMO, i'm gonna add a few more in the future and i'd rather have them swap often (every day or so) so we can get some seen more variety out there.
2 hours might be too short though, 6 might be better. Though I think 2 hours is better

## Changelog
:cl:
tweak: seasonals now refresh every 24 hours rather than 4 days.
/:cl:

### DIFF
--- a/code/controllers/subsystem/persistence.dm
+++ b/code/controllers/subsystem/persistence.dm
@@ -17,7 +17,7 @@ SUBSYSTEM_DEF(persistence)
 
 	///Stores how long each season should last
 	var/list/seasons_durations = list(
-		SEASONAL_GUNS = 4 DAYS,
+		SEASONAL_GUNS = 2 HOURS,
 	)
 	///Stores the current season for each season group
 	var/list/season_progress = list()
--- a/code/controllers/subsystem/persistence.dm
+++ b/code/controllers/subsystem/persistence.dm
@@ -17,7 +17,7 @@ SUBSYSTEM_DEF(persistence)
 
 	///Stores how long each season should last
 	var/list/seasons_durations = list(
-		SEASONAL_GUNS = 2 HOURS,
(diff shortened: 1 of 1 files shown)


Hekzder:
I kind of agree that 2h is a bit fast. 4-6 would probably be better. Rounds going to 2h is not uncommon so that would mean seasonals are potentially cycling every round if you get back to back long rounds.

GradusNL:
> 
> 
> I kind of agree that 2h is a bit fast. 4-6 would probably be better. Rounds going to 2h is not uncommon so that would mean seasonals are potentially cycling every round if you get back to back long rounds.

I agree with 6, probably more like 8. Seasonals should be long enough to get everyone used to the current seasonal and maybe even change their builds around it. 2 to 4 relegates seasonals to meme status, even if you could make a viable build with them. People won't spend enough time with them to learn how to use them.

davidchan13:
I'd prefer to see 24-48 hours but 2-6 is alright I suppose, given that many of the Seasons just are not liked or even that good (Looking at you cops and robbers.)

code/controllers/subsystem/persistence.dm
```diff
@@ -17,7 +17,7 @@ SUBSYSTEM_DEF(persistence)
 
 	///Stores how long each season should last
 	var/list/seasons_durations = list(
-		SEASONAL_GUNS = 4 DAYS,
+		SEASONAL_GUNS = 2 HOURS,
```
Pariah919, line 20:
```suggestion
		SEASONAL_GUNS = 24 HOURS,
```

TiviPlus (approved):

PR #8170, opened 2021-09-04
Machete damage nerfed back to 60

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
75 melee damage, paired up with a shield, on top of bump attacks. This easily allows you to outperform xenomorphs in melee, which is arguably their main thing.

## Changelog
:cl: Lewdcifer
balance: Machete damage reduced from 75 to 60.
/:cl:


Lefunny:
hah, if you think marines arent supposed to outperform xenos in meele then you are wrong.

lKiseki:
You already have bump attacks on a machete, and shields. I don't think you need the damage buff considering how shitty of an experience it is for virtually any xenomorph trying to attack you.

lKiseki:
Actually, on second thought (after discussing it with Hekzder), this might be unnecessary if #8171 goes through. So I'm closing this and crossing my fingers that that one gets merged instead.

PR #8171, opened 2021-09-04
Maximum shield integrity reduced from 400 to 300

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
Currently, shields make you way too tanky. The armor's not necessarily in a bad spot, so I want to tackle reducing its maximum integrity first and see how that goes before I decide on nerfing it some more.

## Changelog
:cl: Lewdcifer
balance: Maximum shield integrity reduced from 400 to 300.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -137,7 +137,7 @@
 	icon = 'icons/obj/items/weapons.dmi'
 	icon_state = "marine_shield"
 	flags_equip_slot = ITEM_SLOT_BACK
-	max_integrity = 400
+	max_integrity = 300
 	integrity_failure = 100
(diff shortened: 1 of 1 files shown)


Lefunny:
xenos have defender marines have imortal shield, sounds balanced to me.

BraveMole (changes requested):
Maybe out of scope, but it's even more important if the shield has less health. A "damaged" icon state (not broken, but at half hp) would be really neat

Oscari-pog:
> 
> 
> xenos have defender marines have imortal shield, sounds balanced to me

You mean the one that cant move when fortified? While a marine can run around with a shield?

Oscari-pog:
Oops, misread that, void my reply.

PR #8174, opened 2021-09-04
 Baldur Mk1 loses slowdown and Mk2 at a new price, 8.

Passe2142:
## About The Pull Request

Changes Baldur Modules, the slowdown for MK1 to 0,  for Mk2 reduces its price to 8.
Now on Req the item is called "Jaeger Baldur Mk2"

## Why It's Good For The Game

Baldur is one of the less used modules and given that pas lost its light for modules it will help to compesate that lightness and for Mk2 on PAS/Jaeger puts that module as a good contender to the other ones. 

## Changelog
:cl:
Balance: Baldur Mk1 loses slowdown, and Mk2 at a value of 8 from 10. Also upon inspection now it doesnt say that it slows down.
The name for the module on the console now is: "Jaeger Baldur Mk2"
/:cl:


Hekzder:
Very based. The Mk1 changes are good because Vali now has a competitor for "I want to run a module but I don't want slowdown", and frankly a moderate light increase really isn't worth much slowdown. Baldur Mk2 is probably now an actual viable module choice instead of a meme. 10 cost makes it the cost of two optical imagers or a fulton pack, which it isn't really worth IMO. Decently increased light + a cost reduction should make people actually buy and use it more. 

Plus honestly 7 light range is just funny and I think we all want to see a squad of Baldur 2 Marines running around.

code/modules/modular_armor/attachables.dm
```diff
@@ -34,11 +34,11 @@
 /** Shoulder lamp strength module */
 /obj/item/armor_module/attachable/better_shoulder_lamp
 	name = "\improper Mark 2 Baldur Light Amplification System"
-	desc = "Designed for mounting on the Jaeger Combat Exoskeleton. Substantially increases the power output of the Jaeger Combat Exoskeleton's mounted flashlight. Slows you down minorly."
+	desc = "Designed for mounting on the Jaeger Combat Exoskeleton. Substantially increases the power output of the Jaeger Combat Exoskeleton's mounted flashlight."
 	icon_state = "mod_lamp_icon"
 	item_state = "mod_lamp"
 	slowdown = 0
-	var/power_boost = 5 /// The boost to armor shoulder light
+	var/power_boost = 7 /// The boost to armor shoulder light
```
BraveMole, line 41: have you tested if this does anything ? with new lights, there's a serious possibility this doesn't do anything

BraveMole (changes requested):

code/modules/modular_armor/attachables.dm
Passe2142: I tested after i changed it, gonna test pre change and after with a measure.

BraveMole (changes requested):
Is there any situation where baldur mk2 gives more light than baldur mk1? And, is this difference anywhere near worth 8 points?

At this point, just delete MK2

Passe2142:
Uhh its map compile that got carried over? I adjusted the list for just /obj/item/armor_module/attachable/better_shoulder_lamp, that one i did not find in VSC. I could use a hand here.

image

Passe2142:
Trying again in new one.

PR #8180, opened 2021-09-05
Rage Power is Now Maximized While at Negative HP

Surrealaser:
## About The Pull Request

Ravager's Rage now has maximum power when used at any negative HP amount instead of -57 HP or less. This is to make things more consistent and offset the loss of extra critical HP that the original rage power scaling was based on.

## Why It's Good For The Game

Rage Power is now properly adapted to the bonus Critical Hit Points that currently exist. Rage Power scaling is made more consistent and simplified.

## Changelog
:cl:
balance: Ravager's Rage now has maximum power when used at any negative HP amount instead of -57 HP or less.
/:cl:


BraveMole (changes requested):
Your description and cl are not understandable.

Surrealaser:
> Your description and cl are not understandable.

I mean I thought they were pretty clear, but I've cleaned them up

BraveMole (approved):

PR #8183, opened 2021-09-05
Revert #7928

EvilMelonTheSecond:
## About The Pull Request
Reverts the PR  Fixes infinite antibiotics #7928.

## Why It's Good For The Game
Antibiotics were indeed fixed but it seems they had a very flawed design to begin with. As an example "wounds" are independent of damage which means healing via meds will leave you with invisible wounds that get infected. This means that whenever you get even slightly damaged you have to run to backlines and treat ALL your brute injuries THEN take pills or if you healed up before treating, treat your every limb to make sure you don't get an infection. Alternatively every marine needs to take antibiotics all the time or they get constant tox damage that escalates to necrosis which needs surgery.

Infections need a proper rework before they are introduced. As I see it, their current state will slow marine pushes or cause people to cadehug.

## Changelog
:cl:
balance: Infections from wounds are removed again.
/:cl:


SplinterGP:
this seems like a bug with needing bicaridine to fully close wounds, same happens to IB sometimes if you use QC+ without it.

TiviPlus:
Splinter read the PR this is referencing

As for this PR a kneejerk revert is not a solution, tweak the infection numbers instead accordingly

SplinterGP:
> 
> 
> Splinter read the PR this is referencing
> 
> As for this PR a kneejerk revert is not a solution, tweak the infection numbers instead accordingly

yes and this seems VERY wack and old bay code, i dunno if this plays any good at all, even more when this was like 1 PR that moved a number that enabled it all, this must have been removed for a good reason.

PR #8184, opened 2021-09-05
Removes Baldur Mk2 and Mk1 slowdown.

Passe2142:
## About The Pull Request
Removes Baldur Mk2 and the slowdown of Mk1. (I just tested locally and no errors).

## Why It's Good For The Game

Baldur is one of the less used modules and given that pas lost its light for modules it will help to compensate that lightness for PAS/Jaeger putting Baldur module as a good contender to the other ones.

## Changelog
:cl:
Balance: Baldur Mk1 loses slowdown, and Mk2 is removed. Also upon inspection now it doesn't say that it slows down.
The name for the Mk1 now is : Baldur Light Amplification System
/:cl:


BraveMole (changes requested):
remove /obj/item/armor_module/attachable/better_shoulder_lamp/mark1µ

make /obj/item/armor_module/attachable/better_shoulder_lamp have the same stat and name as what you made for mk1

BraveMole (approved):

PR #8189, opened 2021-09-06
Remove Sadar Miniscope

MiniMeatwad:
## About The Pull Request
Removes the miniscope as an attachment option for the sadar.

## Why It's Good For The Game
Removes off-screen one shot kills.

## Changelog
:cl:
balance: Sadar no longer has a miniscope attachment option.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -1019,7 +1019,6 @@
 	general_codex_key = "explosive weapons"
 	attachable_allowed = list(
 		/obj/item/attachable/magnetic_harness,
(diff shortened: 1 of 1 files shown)


Hekzder:
I don't think this is a horrible idea but if you do then you should reduce it's price down from what it got nerfed to.

Lefunny:
> I don't think this is a horrible idea but if you do then you should reduce it's price down from what it got nerfed to.

no need for this, sadar are being abçe to insta kill xenos out screen

MiniMeatwad:
Closing in favor of removing scoping while walking first

Issue #8134, opened 2021-08-31
Multiple teleports on Medevac Stretcher


AzonStarfall:
### Description

When using a Medevac stretcher with displacement evac, with two or more operators, both can trigger teleport spooling sequences and allow for multiple simultaneous teleports off on a single cooldown.

A check is done for a single operator, to make sure that they can't activate the bed more than once. This check is not held for each subsequent operator, who are able to subvert the cooldown, so long as their activation is performed within the initial operators spooling cycle.

### Test Merges

This has been an issue since mid-2020 at the earliest, and is not associated with any specific PR.

### Reproduction Steps

1. Deploy stretcher with two or more operators.
2. Operator 1 begins displacement activation, noted by spooling sound effect.
3. Operator 2 begins displacement activation, noted by second overlapping spooling sound effect.
4. Strap subject onto bed, wait for them to be sent by the first activation.
5. Once subject 1 is teleported, strap the second subject onto the bed, while the second spool is proceeding.
6. Subject 2 will be teleported within the same cooldown window.

PR #8190, opened 2021-09-06
Fix being able to send two people with one medevac

BraveMole:
## About The Pull Request

Fixes : https://github.com/tgstation/TerraGov-Marine-Corps/issues/8134

## Changelog
:cl:
fix: Fix being able to send two people with one medevac
/:cl:


QualityVan:
This is gonna break if someone unbuckles during windup. What about working off whether teleport_timer currently exists?

BraveMole:
> This is gonna break if someone unbuckles during windup. What about working off whether teleport_timer currently exists?

Doesn't work, the timer is set when the mob is sent home, the bug is when you are spamming the medevac before this. Doing like this also fixes a runtime, so better

TiviPlus (approved):

Issue #7697, opened 2021-07-31
Ai breaks tadpole console if they use it remotely (might be related to a bug with humans and tadpole console)


2whatever2:
### Description

Pretty sure this works with humans, as I've dealt with players being unable to move, or cancel camera view in tadpole. When another player accesses tadpole, both players see through the camera eye, but only the inital usercan move the camera, both can either land or send up the ship though.

### Test Merges

doesn't matter

### Reproduction Steps

//AI//
1. Remotely use tadpole
2. Can't use it
3. Can't close it.
//Human// (not sure on this one)
1. Use tadpole
2. Some bumbling idiot moves you
3. Stuck using camera view (They can't move it either.)
4. Land it somewhere (Causes camera view to stop)
5. Using the console again lets you move. (Note some players call for help, and get other players dragging them away from the console, making them stuck until someone puts them to re-use tadpole.)
6. Alternatively, if you're in midair, accessing the tadpole while the initial user is stuck viewing the camera you accessing it lets them move the camera, and land it. (not sure if they can move it if the initial user is not in proximity of the console.)

### Screenshots

![image](https://user-images.githubusercontent.com/24631139/127723593-5de64d5d-3677-4040-b09c-b53db48d08d0.png)

I don't have any screenshots for the human bug, I'll do it later, or never.

BraveMole:
I have no clue how ai manage to have access to the console, didn't manage to do it in local server.

And the first bug you are mentionning is fixed

PR #8194, opened 2021-09-06
Fix ai breaking tadpole 

BraveMole:
## About The Pull Request

Fixes: https://github.com/tgstation/TerraGov-Marine-Corps/issues/7697

previous fix didn't fix

## Changelog
:cl:
fix: Fix ai breaking tadpole 
/:cl:

### DIFF
--- a/code/modules/shuttle/mini_dropship.dm
+++ b/code/modules/shuttle/mini_dropship.dm
@@ -200,6 +200,9 @@
 	if(ui_user)
 		return
 
+	if(isAI(user))
+		return
+
 	if(!ui)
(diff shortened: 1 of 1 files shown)


SplinterGP (changes requested):
woulndt a proper fix go by changing the dropship_state that ui_state() uses ?, since some other machinery is also handled like that(just has a check under its ui state if its UI(like the physical state)

TiviPlus (changes requested):
How is it interacting with it with INTERACT_OBJ_UI anyway

Issue #8087, opened 2021-08-27
Defender cant slash marines while fortified


Oscari-pog:
### Description

What the title says.
Apparently you can still slash cades and objects, just not marines specifically.

### Test Merges

#8065: 'Scatterspit V2.0' by vilereaver41 at commit 3e1c07c017
#8042: 'Yet more Sulaco QOL' by Derrick2Electricboogaloo at commit 7be372b018
#8005: 'Praetorian Acid Splash tweak.' by vilereaver41 at commit 5d22715011
#7815: 'Adds movement and direction switch for scopes with ctrl+Wasd' by SplinterGP at commit 7e3f54674a
#8007: 'Fix Server crashing Signallers Exploit' by SpaceLoveSs13 at commit f3fc7d31e6
#8033: 'Defile Ability Rework' by Surrealaser at commit 52b6407ee1
#7990: 'You can use some xeno abilities while riding[DNM]' by BraveMole at commit df96221e60

### Reproduction Steps

1. Play defender.
2. Fortify.
3. Click marine.
4. Cry

lbnesquik:
Is this on purpose?

Oscari-pog:
> Is this on purpose?

Nope, confirmed by maintainer.

BraveMole:
David said it's on purpose

PR #8198, opened 2021-09-07
Fortified defenders can attack marines as well

BraveMole:
## About The Pull Request

Fixes: https://github.com/tgstation/TerraGov-Marine-Corps/issues/8087

Defender can attack structure while fortified, and since you are immobile i think it's fine to be able to slash marines as well

Also clean the code of some deprecated mechanic 

## Changelog
:cl:
balance: Fortified defenders can attack marines as well
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -34,22 +34,6 @@
 	span_warning("We shove [src]!"), null, 5)
 	return TRUE
 
-/mob/living/proc/can_xeno_slash(mob/living/carbon/xenomorph/X)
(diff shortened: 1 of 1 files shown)


Oscari-pog:
Blessed fix

SplinterGP (changes requested):
i'm pretty sure the defender being able to hit structures is what is actually not intended, plus i believe that while it being able to attack marine is okayish but it being able to attack the defenses while fortified is wack, since they get a pretty big resistance when doing so and theyre supposed to be basically immobile(thats the only thing they can do fortified as far as i know, and getting out of fortify can be done pretty quickly so its not like the defender is overcomitting by fortifying somewhere.

David-Stormwell:
Thought it was something which was made into a feature/not slashing while fortified per balance reasons, and with the recent turret nerf this would make it quite easy to breach FOBs with a defender quite easy.

PR #6752, opened 2021-05-03
Rebalances standard issue pistol sunder (TP14 to 1.65, TP23 to 2.15)

lovegreenstuff:
## About The Pull Request

See title.

## Why It's Good For The Game

Large disparity between the sunder per second and sunder per magazine between the TP14 and TP23, which are supposed to be side-grades. Doing this lowers said disparity to more reasonable levels, and doesn't make the TP14 effectively a straight upgrade over the TP23.

Also helps with the giant meme that is dual TP14s shredding ancient queens in the time it takes to empty one magazine (~3 seconds).

Done at the request of Georgeur/Revy

## Changelog
:cl:
balance: lowered TP14 sunder to 1.4, kept most commonly used ammo types at 2.
balance: changes TP14 to 1.65, buffs TP23 bullet sunder to 2.15.
/:cl:

PR #8217, opened 2021-09-08
TP-14 Standard Pistol Nerf

MiniMeatwad:
## About The Pull Request

I can try targeting akimbo more effectively while retaining the pistol's original design if this nerf is too harsh, but this is just a basic nerf on the TP-14 to guarantee that it is no longer melting xenos so easily.

## Why It's Good For The Game

Testing in-game pre-nerf: 
The akimbo pistol can reduce a 500 health Ancient Empress to 80 health in 42 shots. 
The pistol can empty its mag in 3 seconds with a fire rate of 8 shots per second.
It deals 69.3 sunder from 42 shots.
After nerf:
It reduces Ancient Empress to 44 health in 42 shots. Still takes 3 seconds.
It deals 42 sunder from 42 shots.

Yes, I think I tested poorly to get an increase in damage after the nerf, but it's still weaker anyways purely statistically.

Now the sunder is less insane, at 16 sunder per second for akimbo. It can still laser beam right now though, so I will be adding another modifier for akimbo to make it have a miss chance in another PR and apply it to TP-14 first.

## Changelog
:cl:
balance: The TP-14 sunder is reduced to 1 from 1.65, and its akimbo scatter range from (30 to 40) to (40 to 50). This will come out as 5 to 15 with attachments instead of -5 to 5.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -70,7 +70,7 @@
 	scatter_unwielded = 0
 	recoil = -2
 	recoil_unwielded = -2
-	upper_akimbo_accuracy = 6
+	upper_akimbo_accuracy = 5
 	lower_akimbo_accuracy = 4
 
 //-------------------------------------------------------
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -70,8 +70,8 @@
 	scatter_unwielded = 0
 	recoil = -2
 	recoil_unwielded = -2
-	upper_akimbo_accuracy = 10
-	lower_akimbo_accuracy = 9
+	upper_akimbo_accuracy = 7
+	lower_akimbo_accuracy = 6
 
 //-------------------------------------------------------
 //TX-7 Plasma Pistol
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -70,8 +70,8 @@
 	scatter_unwielded = 0
 	recoil = -2
 	recoil_unwielded = -2
-	upper_akimbo_accuracy = 7
-	lower_akimbo_accuracy = 5
+	upper_akimbo_accuracy = 6
+	lower_akimbo_accuracy = 4
 
 //-------------------------------------------------------
 //TX-7 Plasma Pistol
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -70,8 +70,8 @@
 	scatter_unwielded = 0
 	recoil = -2
 	recoil_unwielded = -2
-	upper_akimbo_accuracy = 5
-	lower_akimbo_accuracy = 4
+	upper_akimbo_accuracy = 4
+	lower_akimbo_accuracy = 3
 
 //-------------------------------------------------------
 //TX-7 Plasma Pistol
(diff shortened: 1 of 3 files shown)


InterroLouis:
The sunder reduction is good but the magazine reduction is entire unnecessary and will lead to it just never getting used again due to uselessness.

code/modules/projectiles/guns/pistols.dm
```diff
@@ -56,7 +56,7 @@
 	icon_state = "tp14"
 	item_state = "tp14"
 	caliber = CALIBER_9X19 //codex
-	max_shells = 21 //codex
+	max_shells = 14 //codex
```
BraveMole, line 59: this kills the tp14 niche. All pistols should stay unique

BraveMole (changes requested):

code/modules/projectiles/guns/pistols.dm
MiniMeatwad: Okay. I'll remove it and we'll see what the sunder change does.

BraveMole (approved):

Hekzder:
Akimbo nerf is w/e, but the sunder nerf just memes it to a degree of making the TP-23 better in every way. You've literally nerfed it's sunder by more than 75%. Akimbo was a mistake, honestly, because balancing pistols is hell and you run into situations like this where nerfing one without nerfing the others just makes it obsolete.

MiniMeatwad:
It'd be cool if we didn't have so many variants of the same weapon.

Pariah919:
this doesn't fix any of the problems appreant with akimbos insane meme dps and only further nerfs pistols into the ground
maybe one day we'lla ctually get someone to fix pistols

davidchan13:
Why does every PR that states it's intention to nerf akimbo pistols never actually touch akimbo mechanics and just does some hackjob nerf of actual intent of the weapon?

MiniMeatwad:
> 
> 
> Why does every PR that states it's intention to nerf akimbo pistols never actually touch akimbo mechanics and just does some hackjob nerf of actual intent of the weapon?

The sunder of the TP23 is now 30 per magazine, making it -6 less for being 9mm, a weaker round. Though rifles deal pretty shit sunder per second, the T-12 sunder per magazine is still 25, making this only -4 worse than a rifle per magazine, and it absolutely beats the rifle in sunder anyways. Especially if akimbo.

Also, I found out that attachments (RDS, RC, GS) make the gun into a laser beam, so I have increased the scatter for akimbo even more, so it's only effective in close range. You don't need to carry 10 pistols and 21 magazines while dealing 320 DPS at long range anyways.

MiniMeatwad:
I should add an akimbo accuracy modifier to cause a chance to miss for using akimbo instead of giving it ungodly amounts of scatter so that it can't hit anything though. I'll work on that.

BraveMole (approved):

Tristan63:
pistols are pretty useless as is, add the akimbo nerf and keep it as it was

BraveMole:
Yeah, this is a bandaid fix. The real fix would be an akimbo refactor, so we can properly balance pistols. In the meantime, all the changes are meaningless.

The sunder change is good i think though

Georgeur:
I did #6752 
I mean what's done is done but want it noted the TP-23 was more powerful than the TP-14 before this. 
The TP-14 should want more sunder per second than the TP-23 as its lower base damage means its greater effected by damage falloff. 
But I ain't no tryhard bitch anymore and akimbo in general makes any gun that can be fired 1 handed god awful to balance so good pr. Also unrelated but we nerfed the t19 3 times for being supposedly overpowered akimbo and then reverted all the nerfs and kept one of its slight buffs, 3 months later. Maintainers be widling.

PR #8227, opened 2021-09-09
Replaces ryetalin in combat injectors with dylovene

QualityVan:
## About The Pull Request
The other ingredients in a combat injector (bkt) purge ryetalin, so you'd get either one or zero ticks of it working before it got removed. Replacing rye with dylo makes it more reliable as a low-grade purgative option. It also adds minor toxin healing, which isn't super important but isn't otherwise available in base marine loadout supplies.

## Why It's Good For The Game
An injector having ingredients that purge each other is a little wack and led to cases where one reagent would just never apply.

## Changelog
:cl:
fix: Combat injectors have dylovene instead of ryetalin, so they'll purge toxins more reliably.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -47,9 +47,9 @@
 	icon_state = "autoinjector-4"
 	volume = 30
 	list_reagents = list(
-		/datum/reagent/medicine/ryetalyn = 5,
 		/datum/reagent/medicine/bicaridine = 10,
 		/datum/reagent/medicine/kelotane = 10,
+		/datum/reagent/medicine/dylovene = 5,
 		/datum/reagent/medicine/tramadol = 5,
 	)
 
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -47,10 +47,10 @@
 	icon_state = "autoinjector-4"
 	volume = 30
 	list_reagents = list(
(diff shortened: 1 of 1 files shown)


BraveMole (changes requested):
IMO it's still dumb to have a mix of medicine that are pruging each other.

How about replacing this with dyloven? Since it doesn't cause satm damage if it's lower than 10 u,  that might do the trick

jroinc:
Too *much* purge in that case. Though I suppose swapping out the 2.5u ryetalin for 1u dylo would be the same effect.

QualityVan:
Swapping it to 5u total dylo, 2.5 per injection, is probably fine. It's significantly better at that point against light neuro but significantly worse against standard, and without the whole self-purging thing.

code/game/objects/items/reagent_containers/autoinjectors.dm
```diff
@@ -47,10 +47,10 @@
 	icon_state = "autoinjector-4"
 	volume = 30
 	list_reagents = list(
+		/datum/reagent/medicine/ryetalyn = 5,
 		/datum/reagent/medicine/bicaridine = 10,
 		/datum/reagent/medicine/kelotane = 10,
 		/datum/reagent/medicine/tramadol = 5,
```
QualityVan, line 53:
```suggestion
		/datum/reagent/medicine/bicaridine = 10,
		/datum/reagent/medicine/kelotane = 10,
		/datum/reagent/medicine/dylovene = 5,
		/datum/reagent/medicine/tramadol = 5,
```

BraveMole (approved):

PR #8258, opened 2021-09-11
vehicle multitile adjecency support and minor cleanup

TiviPlus:
## About The Pull Request
Per title, cleaned up with AS es while i was there

## Changelog
:cl:
code: Backend vehicle multitile prep
/:cl:

### DIFF
--- a/code/_onclick/adjacent.dm
+++ b/code/_onclick/adjacent.dm
@@ -88,9 +88,23 @@
 //Multitile special cases.
 /obj/structure/Adjacent(atom/neighbor)
 	if(bound_width > 32 || bound_height > 32)
-		for(var/X in locs)
-			var/turf/T = X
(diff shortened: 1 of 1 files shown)


code/_onclick/adjacent.dm
```diff
@@ -88,9 +88,23 @@
 //Multitile special cases.
 /obj/structure/Adjacent(atom/neighbor)
 	if(bound_width > 32 || bound_height > 32)
```
BraveMole, line 90: how comes we don't have a flag for multi loc objs?

BraveMole (approved):

code/_onclick/adjacent.dm
TiviPlus: bound width and height need to be >32 for multitile and thats a byond builtin

PR #8260, opened 2021-09-11
species innate species trait support

TiviPlus:
Backend for robots


code/modules/mob/living/carbon/human/species.dm
```diff
@@ -227,10 +229,15 @@
 		to_chat(prefs.parent, span_warning("You forgot to set your synthetic name in your preferences. Please do so next time."))
 
 /datum/species/proc/on_species_gain(mob/living/carbon/human/H, /datum/species/old_species)
+	for(var/newtrait in inherent_traits)
+		ADD_TRAIT(H, newtrait, SPECIES_TRAIT)
 	return
```
BraveMole, line 234: usesless return, forgot the should call parent

BraveMole (changes requested):

code/modules/mob/living/carbon/human/species.dm
TiviPlus: Fixe

BraveMole (approved):

PR #8264, opened 2021-09-11
makes the M-44 into a proper single action+minor stat changes

Pariah919:
## About The Pull Request
M44 is actually single action now
fire delay reduction to 0.15 as per tp44
is normal sized now
also adds an unprimied state but that code is broken so ???

## Why It's Good For The Game
M44 should actually be single action
it has slightly better damage/onhit stats but has the annoyance of priming in it

## Changelog
:cl:
balance: m44 now works like a proper single action, use unique action to prime it, it has slightly less fire delay and is normal sized from bulky
/:cl:


Derrick2Electricboogaloo:
Cringe single action revolver in the 25th century

BraveMole (approved):

Kuro020 (approved):

PR #8267, opened 2021-09-12
Infinite gauze, finite brute kits

QualityVan:
## About The Pull Request
Prevents restocking stacks if they don't have the amount the stack initially had.
Gauze and ointment in the marine meds are infinite.

## Why It's Good For The Game
Limiting the better, specialist option while making the general option more available.
If you really need more kits, you can buy 50 of each for 5 req points.

## Changelog
:cl:
balance: Marine meds have infinite gauze and ointment
balance: Stacks can't be restocked in vendors unless they have the count they started with.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/vending.dm
+++ b/code/game/objects/machinery/vending/vending.dm
@@ -689,6 +689,12 @@
 			if(cell.charge < cell.maxcharge)
 				to_chat(user, span_warning("\The [cell] isn't full. You must recharge it before you can restock it."))
 				return
+		else if(istype(item_to_stock, /obj/item/stack))
+			var/obj/item/stack/stack = item_to_stock
+			if(stack.amount != initial(stack.amount))
+				to_chat(user, span_warning("[stack] has been partially used. You must replace the missing amount before you can restock it."))
+				return
+
 		if(item_to_stock.loc == user) //Inside the mob's inventory
 			if(item_to_stock.flags_item & WIELDED)
 				item_to_stock.unwield(user)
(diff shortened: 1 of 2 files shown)


jroinc:
Seems reasonable. Up the marinemed quantity some, to 20 total?

EvilMelonTheSecond:
So now people will run outta kits to revive people and have to beg req before bodies go cold. The "specialist" option is mandatory for revivals and the cheap option already has the upside of not having a low medical skill debuff, this isn't a good idea.

BraveMole (approved):

code/game/objects/machinery/vending/vending.dm
```diff
@@ -689,6 +689,12 @@
 			if(cell.charge < cell.maxcharge)
 				to_chat(user, span_warning("\The [cell] isn't full. You must recharge it before you can restock it."))
 				return
+		else if(istype(item_to_stock, /obj/item/stack))
```
BraveMole, line 692: isitemstack

BraveMole (changes requested):

code/game/objects/machinery/vending/vending.dm
QualityVan, line 692:
```suggestion
		else if(isitemstack(item_to_stock))
```

Losenis:
>Has a delay if you're not a corpsman
>Covers only one single bodypart
>You have to check what bodypart is damaged first, either by scan of self-check
>One stack doesn't even cover for all bodyparts since there's 10 kits for 11 bodyparts if the guy just got fucked over by explosives or god forbid being on the literal last tile of an HE OB
>Bruh let's nerf it

PR #8271, opened 2021-09-12
Update gun_system.dm, revolver.dm. Add accuracy penalty to akimbo. Reduce revolver firerate to 1.4 seconds per magazine.

MiniMeatwad:
## About The Pull Request

Revolver fire rate nerf, and upper_akimbo_accuracy_mult and lower_akimbo_accuracy_mult variables to affect the accuracy multiplier for akimbo by a probabilistic median percentage modifier.

Example A: Accuracy of 100% and accuracy modifier of 0.8 to 0.6. This reduces it by 20% to 40%, giving you 80% or 60%.
Example B: Accuracy of 120% and accuracy modifier of 0.8 to 0.6. This reduces it by 20% to 40%, giving you 96% or 72%.
Example C: Accuracy of 150% and accuracy modifier of 0.8 to 0.6. This reduces it by 20% to 40%, giving you 120% or 90%.

## Why It's Good For The Game

Akimbo is a gamble now, if you're lucky it could deal up to 2x as much damage, but it wont deal much less damage than without akimbo.

## Changelog
:cl:
balance: Revolver firerate reduced to 1.4 seconds per magazine, or 5 shots per second.
balance: Akimbo accuracy multiplier set to lower 0.6 and upper 0.8 by default. Feels like equivalent damage to single handed fire now.
code: upper_akimbo_accuracy_mult and lower_akimbo_accuracy_mult variables to affect the accuracy multiplier for akimbo by median percentage.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/revolvers.dm
+++ b/code/modules/projectiles/guns/revolvers.dm
@@ -353,7 +353,7 @@
 		/obj/item/attachable/lace,
 	)
 	attachable_offset = list("muzzle_x" = 33, "muzzle_y" = 19,"rail_x" = 13, "rail_y" = 23, "under_x" = 22, "under_y" = 14, "stock_x" = 22, "stock_y" = 19)
-	fire_delay = 0.15 SECONDS
+	fire_delay = 0.2 SECONDS
 	accuracy_mult_unwielded = 0.85
 	accuracy_mult = 1
 	scatter_unwielded = 15
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -139,6 +139,12 @@
 	///determines lower accuracy modifier in akimbo
 	var/lower_akimbo_accuracy = 1
 
+	// Akimbo is a gamble against odds. Akimbo should be equivalent to single handed in damage at its worst. Fortune blesses you.
+	///determines upper accuracy multiplier in akimbo 
+	var/upper_akimbo_accuracy_mult = 0.8
+	///determines lower accuracy multiplier in akimbo
+	var/lower_akimbo_accuracy_mult = 0.6
+
 	///If the gun is deployable, the time it takes for the weapon to deploy.
 	var/deploy_time = 0
 	///If the gun is deployable, the time it takes for the weapon to undeploy.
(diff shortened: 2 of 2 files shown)


Pariah919:
disagree with revolver changes, makes revolver useless compared to tp-23 in every way

if akimbo is broken go fix the actual problem instead of sidestepping it constantly

MiniMeatwad:
> 
> 
> disagree with revolver changes, makes revolver useless compared to tp-23 in every way
> 
> if akimbo is broken go fix the actual problem instead of sidestepping it constantly

Make TP-23 fire faster, revolver fire slower? I don't really think the revolver should fire so fast while having the highest damage and sunder. It's also silly since it has the lowest mag size as well. But I could also increase the damage to compensate for the fire rate change, though that'd be increasing the damage per mag as well.

MiniMeatwad:
I'm willing to separate the revolver nerf from this PR if the balance of the fire rate is debatable.

lovegreenstuff:
i dont think you need to start all of your prs with "update [blank].dm

just sayin bro

MiniMeatwad:
> 
> 
> i dont think you need to start all of your prs with "update [blank].dm
> 
> just sayin bro

ok

code/modules/projectiles/guns/revolvers.dm
```diff
@@ -353,7 +353,7 @@
 		/obj/item/attachable/lace,
 	)
 	attachable_offset = list("muzzle_x" = 33, "muzzle_y" = 19,"rail_x" = 13, "rail_y" = 23, "under_x" = 22, "under_y" = 14, "stock_x" = 22, "stock_y" = 19)
-	fire_delay = 0.15 SECONDS
```
Pariah919, line 356: why is this changed with 0 mention now

MiniMeatwad: I'll remove it and it can be worked on in another PR

BraveMole:
David and I agreed that the other solution is better:
"Making duel wielding cycle though each gun respecting the firerate, so the outcome wouldn't be double DPS, but instead it'd all just be a double magazine size instead."

It's harder to code, but that would fix the akimbo balance issue

This is not a right fix IMO, using rng to balance anything doesn't really fit in TGMC.

PR #8299, opened 2021-09-15
Readds the ZX-76

Pariah919:
## About The Pull Request
By David request it readds the ZX76, 120 points from Req, it is a CQC monster, basically useless past that due to scatter, outside of single fire.

Deals big dick damage and doesn't require special ammo to work, it's pretty nice.

Sprite by Joniroquai#8011 on discord.

image

---
STATS
0.9 damage mult, 2 scatter on semi, 12 on burst fire. 0.45 aim slowdown, 0.75 Wield delay. 0.001 burst delay. +1.25 secondso f extra delay on burst fire
---
Changes from old ZX
- Wield delay is higher overall, aim slowdown is lower since it isn't a roundstart gun anymore.
-10% less damage
-More scatter on burst
-RC/EB removed, if you want to lower that scatter, go get a vert grrip, noob
-+1.25 fire delay if you fire with it on burst.

## Changelog
:cl:
add: the ZX-76 returns in it's (nerfed) glory, avaliable for 100 req points.
balance: zx no longer has rc/eb, deals 10% less damage, burst fire has far more scatter, fires .25 faster. 
/:cl:


David-Stormwell:
Make it 120 points instead, otherwise 100 is too cheap for it's skill ceiling, having a big dick damage cqc weapon which can use standard ammo might make xenos cringe a little, but it does give a good option other then people spamming sadar/minigun like the nerds they are.

Pariah919:
>Make it 120 points instead, otherwise 100 is too cheap for it's skill ceiling, having a big dick damage cqc weapon which can use standard ammo might make xenos cringe a little, but it does give a good option other then people spamming sadar/minigun like the nerds they are.
>
it's 120 now, should be ok

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -910,3 +910,42 @@ can cause issues with ammo types getting mixed up during the burst.
...
+						/obj/item/attachable/reddot,
+						/obj/item/attachable/verticalgrip,
+						/obj/item/attachable/angledgrip,
+						/obj/item/attachable/gyro,
+						/obj/item/attachable/flashlight,
+						/obj/item/attachable/magnetic_harness,
+						/obj/item/attachable/lasersight,
+						/obj/item/attachable/attached_gun/flamer,
+						/obj/item/attachable/attached_gun/shotgun,
+						/obj/item/attachable/attached_gun/grenade)
```
BraveMole, line 940: Why would you ever add attached_gun? It's a gimmick gun with only CQC in mind, this doesn't belong there

BraveMole (changes requested):
Some remarks:
- Does this insta delete let's say mature drones? or mature defenders?
- If this is only useable in CQC, make a special ammo for it, reduce the price of the gun a bit. That's for the sake of consistency (T2 guns have T2 ammo). Also make that ammo expensive, shotguns tends to use a low ammount of shells anyway
- This sprites is way to banal. Xenos should be able to spot the gun and play accordingly, not think it's just another stupid shotguner but ending up blasted in half a sec

David-Stormwell:
> 
> 
> Some remarks:
> 
>     * Does this insta delete let's say mature drones? or mature defenders?
> 
>     * If this is only useable in CQC, make a special ammo for it, reduce the price of the gun a bit. That's for the sake of consistency (T2 guns have T2 ammo). Also make that ammo expensive, shotguns tends to use a low ammount of shells anyway
> 
>     * This sprites is way to banal. Xenos should be able to spot the gun and play accordingly, not think it's just another stupid shotguner but ending up blasted in half a sec

1: So does the SADAR, and other meme weapons/builds.

2: Shotguns use more shells then you realize in long engagements, and if you're maining it, which you'd probably be doing so with the ZX you'll run though quite a bit of ammo.

3: I semi agree, but the sound of it is different enough most of the time that you'll know, and xenos already cope with DBs and normal shotguns ok enough.

4: The utility of it seems fine, if anything, at most, making it so flechette can't be used can be a thing **_At most_**.

Daily reminder other spec weapons exist which can end a xeno's life quite easily, from the scout rifle nobody uses for some odd reason, to the sadar and antimat marine rifle, this fills in the role of being another choice while not being off-screen delete button, and do mind CQC is much more dangerous , from warriors to even a simple defender tailsweep ending your rampage quite easily.

Kuro020 (changes requested):
well like bravemole said,this probably should stand out more,paint it red or something

vilereaver41:
That sprite is disgusting for a 120 req wep.

Pariah919:
>
    Does this insta delete let's say mature drones? or mature defenders?
    If this is only useable in CQC, make a special ammo for it, reduce the price of the gun a bit. That's for the sake of consistency (T2 guns have T2 ammo). Also make that ammo expensive, shotguns tends to use a low ammount of shells anyway
    This sprites is way to banal. Xenos should be able to spot the gun and play accordingly, not think it's just another stupid shotguner but ending up blasted in half a sec
>

Doesn't one shot mature drones, if you're shooting a mature drone at 3 tile which is the intended distance for a 120 point weapon, then this seems decent. Considering it's b18 level of price it should have a noticeable affect on the field.

Attached guns are there because it's a normal gun overall, I don't see the need for it to not have them outside of maybe removing masterkey.

Special ammo would mean I'd reduce the price to 90 or so, this uses up ammo quick, ZX is a ammo hog and the loss of most scatter attachments in this rework means that it needs to make it's shots count more.

Sprite is tan, no other gun is tan, and the shape is recognizable.

TGCMbetcher:
Im more worried about how say a flak vest premed SL would do against the queen while running and having the stamina regen command active. You can keep pace with the queen using normal armor sometimes with the command, high cqc burst and can keep up with t3s would be a deletion machine

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -910,3 +910,43 @@ can cause issues with ammo types getting mixed up during the burst.
 /obj/item/weapon/gun/shotgun/pump/t35/nonstandard
 	current_mag = /obj/item/ammo_magazine/internal/shotgun/pump/buckshot
 	starting_attachment_types = list(/obj/item/attachable/stock/t35stock, /obj/item/attachable/angledgrip, /obj/item/attachable/magnetic_harness)
+
+//-------------------------------------------------------
+//THE MYTH, THE GUN, THE LEGEND, THE DEATH, THE ZX
+
+/obj/item/weapon/gun/shotgun/zx76
+	name = "\improper ZX-76 assault shotgun"
+	desc = "The ZX-76 Assault Shotgun, a incredibly rare, double barreled semi-automatic combat shotgun with a twin shot mode. Possibly the unrivaled master of CQC. Has a 9 round internal magazine."
```
XSlayer300, line 919:
```suggestion
	desc = "The ZX-76 assault shotgun, an incredibly rare, double barreled semi-automatic combat shotgun with a twin shot mode. Possibly the unrivaled master of CQC, and feared amongst other militaries for that reason alone. Has a 9 round internal magazine."
```

code/modules/projectiles/guns/shotguns.dm
```diff
@@ -910,3 +910,43 @@ can cause issues with ammo types getting mixed up during the burst.
...
+	icon = 'icons/Marine/gun64.dmi'
+	icon_state = "zx-76"
+	item_state = "zx-76"
+	max_shells = 10 //codex
+	caliber = "12 gauge shotgun shells" //codex
```
XSlayer300, line 924:
```suggestion
	caliber = "12 gauge" //codex
```
weapon literally became incompatible with shells once you pull them out

BraveMole (approved):
well, david like it and it's expensive, so i guess it's okay

PR #8315, opened 2021-09-16
Revert "Revert "Fix infinite respawn, randomise name on ssd take, add admin loging""

TiviPlus:
Reverts tgstation/TerraGov-Marine-Corps#8314

merge when its fixed


psykzz:
How about splitting this up instead of reverts reverting a revert

PR #8331, opened 2021-09-18
Increase buckshot damage falloff to 8 instead of 1.

MiniMeatwad:
## About The Pull Request

At 40 damage per pellet, ineffective past 5 tiles for a damage reduction of 8 per tile.

Average T3 health: 320, two shot kills T3s.
Pellets per shot: 5
Damage per 5 tiles: 40
Damage per 4 tiles: 80
Damage per 3 tiles: 120
Damage per 2 tiles: 160
Damage per 1 tile: 200

## Why It's Good For The Game

No more long range buckshot, making it CQC specialized while flechette is range specialized. Still probably really effective at 2-3 tiles unlike when the scatter was insanely huge.

## Changelog
:cl:
balance: Buckshot damage falloff increased to 8 from 1.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -698,7 +698,7 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range = 5
 	max_range = 10
 	damage = 40
-	damage_falloff = 1
+	damage_falloff = 8
 	penetration = 0
 
 


RenGusta:
This is worse than old buckshot falloff, which was already shithouse. 

Literally why. Do you only want to see flechette?

BraveMole:
While I agree that shotties are a bit too strong at the moment, compared to other weapons, gutting buckshot like that is way overdoing it. 

Right now, shotties have short, medium and medium-far options. This is good, no need to make slugs and buckshot have the same role

MiniMeatwad:
Bruh. I tested it in-game and it worked around 2-3 tiles, and its effective range was still in the same rough engagement range you'd get yourself into, it's just no longer effective past that.

What am I supposed to do, nerf its damage slightly so it doesn't do anything different from other weapons anyways?

MiniMeatwad:
> 
> 
> This is worse than old buckshot falloff, which was already shithouse.
> 
> Literally why. Do you only want to see flechette?
> 
> Similarly, upon re-reading your about, you fail to mention most T3's have 30-50 armor at various maturities, and buckshot, unless I'm mistaken, lacks any AP or Sunder.

Meh, still 2 shots T3s like praetorian and ravager, 3 shots the others, and it still feels like it fucks over a queen over worse than flechette does at close range. It's really good at close range, and it stills hits about as good as flechette at flechette range without the AP or sunder. It is literally just superior to flechette unless you run into a defender or crusher, which is only two castes in the entire game.

PR #8332, opened 2021-09-18
Add percentage based plasma loss for plasmaloss gas and pepperball gun.

MiniMeatwad:
## About The Pull Request

Plasmaloss gas now removes 20% of your plasma instead of 200 plasma, which is 20% of 1000.
Pepperball gun now removes 5% of your plasma instead of 50, which is 5% of 1000.
Hivelord counts as having +100% plasma when it's orange, affected by its plasma_regen_limit.
Young Ravager counts as having +100% when it's full, affected by its plasma_regen_limit.

## Why It's Good For The Game

So castes like runner and warrior don't lose all their plasma instantly because they have 175 plasma, and all castes are affected equally, except hivelord who has super plasma that they don't get normally.

## Changelog
:cl:
balance: Plasmaloss gas now removes 20% of your plasma_max instead of 200 plasma, which is 20% of 1000.
balance: Pepperball gun now removes 5% of your plasma_max instead of 50, which is 5% of 1000.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
@@ -310,13 +310,13 @@
 	plasma_stored = min(plasma_stored + value, xeno_caste.plasma_max)
 	update_action_button_icons()
 
-/mob/living/carbon/xenomorph/proc/use_plasma_percent(value)
-	value = value * xeno_caste.plasma_max / 100 * xeno_caste.plasma_regen_limit
+/mob/living/carbon/xenomorph/proc/use_plasma_percent(value) // 1 = 100%, 0.5 = 50%, example: 0.5 * 1000 * 1 = 500
+	value = value * xeno_caste.plasma_max * xeno_caste.plasma_regen_limit
 	plasma_stored = max(plasma_stored - value, 0)
 	update_action_button_icons()
 
-/mob/living/carbon/xenomorph/proc/gain_plasma_percent(value)
-	value = value * xeno_caste.plasma_max / 100 * xeno_caste.plasma_regen_limit
+/mob/living/carbon/xenomorph/proc/gain_plasma_percent(value) // 1 = 100%, 0.5 = 50%, example: 0.5 * 1000 * 1 = 500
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
```diff
@@ -310,8 +310,15 @@
 	plasma_stored = min(plasma_stored + value, xeno_caste.plasma_max)
 	update_action_button_icons()
 
+/mob/living/carbon/xenomorph/proc/use_plasma_percent(value) // 1 = 100%, 0.5 = 50%, example: 0.5 * 1000 * 1 = 500
+	value = value * xeno_caste.plasma_max * xeno_caste.plasma_regen_limit
+	plasma_stored = max(plasma_stored - value, 0)
+	update_action_button_icons()
 
-
+/mob/living/carbon/xenomorph/proc/gain_plasma_percent(value) // 1 = 100%, 0.5 = 50%, example: 0.5 * 1000 * 1 = 500
```
BraveMole, line 318: usless procs, use_plasma(the value works enough)

BraveMole (changes requested):

code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
MiniMeatwad: ok

MiniMeatwad:
ok it didn't work

BraveMole (approved):

jroinc:
The only thing I can think that *might* make this an issue is if tanglefoot only affects you every 2 seconds. I *think* it got changed to include on move as well tho.

MiniMeatwad:
> 
> 
> The only thing I can think that _might_ make this an issue is if tanglefoot only affects you every 2 seconds. I _think_ it got changed to include on move as well tho.

Pretty sure it affects on move as well since it's a reagent smoke and neuro smoke does the same.

QualityVan:
> 
> 
> The only thing I can think that _might_ make this an issue is if tanglefoot only affects you every 2 seconds. I _think_ it got changed to include on move as well tho.

It's still got a cooldown on affecting you, maximum once per second. So, faster if you're moving, but not every step.

PR #8333, opened 2021-09-18
Fix hunter stealth not working when two hunters were hiding at the same time

BraveMole:
## About The Pull Request

Forgot that castes were singleton

## Changelog
:cl:
fix: Fix hunter stealth not working when two hunters were hiding at the same time
/:cl:


code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -362,5 +360,7 @@
 
 	///The xenos/silo/nuke currently tracked by the xeno_tracker arrow
 	var/atom/tracked
+	///A target marked
+	var/mob/living/marked_target
```
TiviPlus, line 364: put it on the ability

PR #8345, opened 2021-09-19
Fix gun harddel

BraveMole:
Fix harddel from stock when the gun is taken from the loadout vendor


TiviPlus (changes requested):
Do you mind explainign how thesse fix it tnx

TiviPlus (approved):

PR #8356, opened 2021-09-20
More guns to civil war

BraveMole:
## About The Pull Request

Almost add everything back.

It's a meme gamemode, let's make it fun and powergamy.

Slugs are still banned, as well as miniscope, but you can get sniper, mg, shotguns, grenade launchers, etc.

## Why It's Good For The Game

More guns = more fun. I think?

## Changelog
:cl:
balance: Almost all guns are available in civil war
/:cl:


neohenry:
SHOTGUNNERS BE LIKE

BrandonSmith6:
Slugs should be kept banned they stun for so long and pretty much guarantee a kill.

David-Stormwell:
This seems fine enough, but in the future can you pair with PR with some sort of instant full cover cade?
As pushing a sniper-filled FOB out in the open might be painful as all hell I can see.
If you want an idea, something like Halo's instant deployable energy shield, I'm sure @Kuro020 would be interested in spriting something simliar.

BraveMole:
> This seems fine enough, but in the future can you pair with PR with some sort of instant full cover cade?
> As pushing a sniper-filled FOB out in the open might be painful as all hell I can see.
> If you want an idea, something like Halo's instant deployable energy shield, I'm sure @Kuro020 would be interested in spriting something simliar.

Marines can use smoke or camo nades, they can use mortar. This means you have to be more clever, but i don't think it's a bad thing

Sirsnek15:
only fleche should be allowed in hvh honestly, slugs are awful and both buckshot and slugs need some fall off damage nerf overall

David-Stormwell:
👍 Approved

TiviPlus:
allso fails compile on master

PR #8358, opened 2021-09-20
Prevent permanent damage in Civil War

BraveMole:
## About The Pull Request

Shrap, bye bye, IB, bye bye, organ damage, bye bye

## Why It's Good For The Game

Civil war is more casual, it should stay that way

## Changelog
:cl:
balance: Remove shrap, IB, organ damage from Civil War
/:cl:


code/modules/organs/limbs.dm
```diff
@@ -410,7 +410,7 @@
 		update_wounds()
 
 	//Bone fractures
-	if(CONFIG_GET(flag/bones_can_break) && brute_dam > min_broken_damage * CONFIG_GET(number/organ_health_multiplier) && !(limb_status & LIMB_ROBOT))
+	if(CONFIG_GET(flag/bones_can_break) && brute_dam > min_broken_damage * CONFIG_GET(number/organ_health_multiplier) && !(limb_status & LIMB_ROBOT) && !(SSticker.mode?.flags_round_type & MODE_NO_PERMANENT_WOUNDS))
```
QualityVan, line 413: Should put this check inside fracture instead of here, since there's other things that break limbs directly.

code/modules/organs/organ_internal.dm
```diff
@@ -109,6 +109,8 @@
 				take_damage(1, prob(30))
 
 /datum/internal_organ/proc/take_damage(amount, silent= FALSE)
+	if(SSticker.mode?.flags_round_type & MODE_NO_PERMANENT_WOUNDS)
```
QualityVan, line 112: You'll probably end up finding a couple of sources which give organ damage directly instead of through take_damage but those can be fixed as they show up and should be changed anyway.

code/modules/organs/limbs.dm
BraveMole: the other cases are from surgery mostly, so i think it's fine

code/modules/organs/organ_internal.dm
BraveMole: Looks like everything uses take_damage appart from flashes, which is fine by me

code/modules/organs/limbs.dm
QualityVan: BHJ does in its overdose, is the other one I'm familiar with, but it'd be tricky to trigger that without also repairing them.

David-Stormwell:
> 
> 
> So far Civil war hasn't been too casual, removing these would honestly remove the need for a medical roles, and honestly turn CW into more of a run&gun sort of TDM. I ain't sure about this one, chief. Sorry.

**gtfo medic main**

Anyways, removing _all_ perm damage types might be a bit much, how about keeping bone breaks and IB?

Organ damage can fuck itself.

BraveMole:
Added back bone breaking. I hate IB

PR #8365, opened 2021-09-21
Gives Twin Pillars CAS Again

turnop:
Give me one good reason why I shouldn't be able to dogfight the loyalist pilot in the skies above Ice Colony.

Listen up boot if you got napalmed in the hospital that's a fucking skill issue buddy. Maybe you should have thought about the cas flare in the red cross tent before you got your skull fractured and became extra-crispy unga.

[Gives Twin Pillars back a Condor for each team.]

twin1
twin2


BrandonSmith6:
yes

BraveMole:
This won't work without code change

You lazily removed the drop pods section

This has a lot of balance concern anyway, probably need to remove some ordonnance and slow every missile

PR #8367, opened 2021-09-21
Make command aura permanent for SL and above

EvilMelonTheSecond:
## About The Pull Request
Command aura is now permanent if you are a non aSL command staff.

## Why It's Good For The Game
This is a straight up buff and I didn't balance the numbers at all if maintainers want to test this as is its probably fine though. Discussion about balance can take place both here and discord if you want something changed.

This has been talked about for some while and I wanted to see if discussion could be stimulated more by a bit of testing. I've nerfed aSL leadership skill down to 2 to make the check simpler and make this less similar to xeno pheros. Also tested and made sure one person can only give one aura.

## Changelog
:cl:
balance: Command staff order auras are now permanent
/:cl:

### DIFF
--- a/code/datums/jobs/squads.dm
+++ b/code/datums/jobs/squads.dm
@@ -304,7 +304,7 @@ GLOBAL_LIST_EMPTY(helmetmarkings_sl)
 
 	//Handle aSL skill level and radio
 	if(!ismarineleaderjob(squad_leader.job))
-		squad_leader.skills = squad_leader.skills.setRating(leadership = SKILL_LEAD_EXPERT)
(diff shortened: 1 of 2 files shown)


ilikesaying:
Mfw light jaeger SL outruns queen

BraveMole:
This is absolutly not the way to implement permanent orders. 

This code doesn't work anyway, since you're checking in the human receiving the orders has a good skill.

Implement a permanent auras need a proper refactor, probably with a pheromone refactor so they both use the same system, it's not a 3 lines change

PR #8383, opened 2021-09-22
Space no longer instantly kills you

BraveMole:
## About The Pull Request

Still hurt a bunch

## Why It's Good For The Game

Useless instakill creates way more issues than it is supposed to resolve

## Changelog
:cl:
balance: Space no longer instantly kills you
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -371,3 +371,13 @@
 	REMOVE_TRAIT(owner, TRAIT_MUTED, TRAIT_STATUS_EFFECT(id))
 	return ..()
 
+/datum/status_effect/spacefreeze
+	id = "spacefreeze"
+
+/datum/status_effect/spacefreeze/on_creation(mob/living/new_owner)
+	. = ..()
+	to_chat(new_owner, span_danger("The cold vacuum instantly freezes you, maybe this was a bad idea?"))
+
+/datum/status_effect/spacefreeze/tick()
+	owner.adjustFireLoss(5)
+
--- a/code/game/area/general.dm
+++ b/code/game/area/general.dm
@@ -12,13 +12,6 @@
 	pressure = 0
 	flags_area = NO_DROPPOD
 
-/area/space/Entered(atom/movable/arrived, atom/old_loc, list/atom/old_locs)
-	. = ..()
-	if(isliving(arrived))
-		to_chat(arrived, span_danger("The cold vacuum instantly freezes you, maybe this was a bad idea?"))
-		var/mob/living/spaceman = arrived
-		spaceman.adjustFireLoss(600) //Bad idea, spessman.
-
 /area/engine/
 	ambience = list('sound/ambience/ambisin1.ogg','sound/ambience/ambisin2.ogg','sound/ambience/ambisin3.ogg','sound/ambience/ambisin4.ogg')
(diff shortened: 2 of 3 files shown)


EgorDinamit:
Think the insta-kill thing was also added to some other tiles, like sea.
If they are not a subtype of "space", could you do something about them as well?

BraveMole:
> Think the insta-kill thing was also added to some other tiles, like sea.
> If they are not a subtype of "space", could you do something about them as well?

sea is space

code/datums/status_effects/debuffs.dm
```diff
@@ -371,3 +371,13 @@
 	REMOVE_TRAIT(owner, TRAIT_MUTED, TRAIT_STATUS_EFFECT(id))
 	return ..()
 
+/datum/status_effect/spacefreeze
+	id = "spacefreeze"
+
+/datum/status_effect/spacefreeze/on_creation(mob/living/new_owner)
+	. = ..()
+	to_chat(new_owner, span_danger("The cold vacuum instantly freezes you, maybe this was a bad idea?"))
```
AnCopper, line 379: "Cold vacuum" is an oxymoron.

BraveMole: why would it be? space has a temperature, it happens to be cold

MereWalker:
I just want to point out the reason why the tile become INSTA kill was due to the wraith being very broken at the time. I don't support this but id like to see how this pans out regardless.

code/datums/status_effects/debuffs.dm
AnCopper: Space is nothing, it literally can't be cold, because it's nothing.

BraveMole: You can define temperature stricly out of radiations. And if you consider this temperature, space is cold

bowlofcereal:
Didnt we make them instakill because people were just setting up defenses in space to hide out in during a hijack, or the wraith teleporting out into space a lot? I'm not really against this, just hoping previous issues that caused it to become instakill in the first place can either be resolved or prevented

BraveMole:
> Didnt we make them instakill because people were just setting up defenses in space to hide out in during a hijack, or the wraith teleporting out into space a lot? I'm not really against this, just hoping previous issues that caused it to become instakill in the first place can either be resolved or prevented

The issues were people staying in space. Marines build fort in space, wraith teleporting out and marines not being able to do anything.

Everyone still takes 25 fire damage every sec in space, so these strats are dead as well

Alexeivichi:
Pretty sure space tiles insta-killing was to act as a "No-play" area. Doesn't make sense to make them not akin to a kill brushed area, other than turning them into a "hazard" rather than... Well, "No play deathzone".

Time to build a space bar once more I guess.

bowlofcereal:
> > Didnt we make them instakill because people were just setting up defenses in space to hide out in during a hijack, or the wraith teleporting out into space a lot? I'm not really against this, just hoping previous issues that caused it to become instakill in the first place can either be resolved or prevented
> 
> The issues were people staying in space. Marines build fort in space, wraith teleporting out and marines not being able to do anything.
> 
> Everyone still takes 25 fire damage every sec in space, so these strats are dead as well

Okay, that works then. Cheers.

TiviPlus (approved):

PR #8390, opened 2021-09-23
Self-charging power cells

jroinc:
## About The Pull Request

Adds a pair of RTG power cells to engi vendors, 5/15 points each. Recharge rates are 8x/4x slower than the hand charger and charge capacity is smaller than standard 10,000 charge highcaps, but it recharges with no needed interaction, even inside items.

Smol one is a regular power cell with only 2000 capacity, but a recharge rate of 25 per 2 seconds.
Lorg one has 5000 capacity and a recharge rate of 50/2 seconds, but it doesn't fit in storage and has to be either hand-carried or carried inside an item.

The hand charger can charge up the smol cell, but not the lorg one. Too easy to store one inside there.

## Why It's Good For The Game

More engineering features. A way to power APC's long-term if you have the points. A way to buy a longer-term but weaker plasmacutter battery. An alternative to the existing hand charger.

## Changelog
:cl:
add: Added two self-recharging cells to engi vendors. Low capacity, but they charge up to full slowly without needing a hand charger.
/:cl:


BraveMole (approved):
TM because i don't know if this recharge feature was ever used in game or tested

jroinc:
> 
> 
> TM because i don't know if this recharge feature was ever used in game or tested

I've tested it ingame, and the Pulse Laser batteries use it, but I think they're the only thing. Only quirk I've noticed is they *instantly* blow up if destabilised, possibly off of charge increment... which is a feature to me IMO.

XSlayer300:
> > TM because i don't know if this recharge feature was ever used in game or tested
> 
> I've tested it ingame, and the Pulse Laser batteries use it, but I think they're the only thing. Only quirk I've noticed is they _instantly_ blow up if destabilised, possibly off of charge increment... which is a feature to me IMO.

Exactly, I kinda ported cell recharge system from polaris a long long time ago, only for that specific weapon.

BraveMole (approved):

PR #8410, opened 2021-09-24
Kill larva burst, replace it by zombies

BraveMole:
## About The Pull Request

Larval huggers will inject larva, exactly like they are doing now. The difference will be that these larvas will never burst, they will just constently produce zombium.

You'll start to get noticeable effects (20 zombium, so some oxy and tox damage plus being undefibbable turning you into a zombie) after 3 minutes 30. Zombium can be purged using hypervene or ryetalin, but your best bet is to go to surgery. 

Also clean up the code from isnestedhost and canxenoslash, which are relics from cap gameplay. Not a player facing change

## Why It's Good For The Game

Burst is a remnant of cap gameplay that basically never happens (only with afk players).

Zombies are funnier, even though i have no clue if they fit to normal distress.

This is just a test, i'll maybe bury this project if it's not good enough

## Changelog
:cl:
add: xeno embryo (from larvas hugger) will no longer burst, but constently injects zombium until you die. If you have 20 zombium in your blood stream and you get undeffibable, you'll turn into a zombie
/:cl:


Ragecrabbo:
Zombie Gameplay

XSlayer300:
I wonder if like, they would be infested by parasites that are controlled by the hivemind or something, rather than traditional zombies.

Al-1ce:
In the event that zombies are unfit for distress (which, I can nearly guarentee you they will be), _please_ keep the oxy/tox damage from having larvas in your system.

DarkWolfKai:
Xenos newest caste : The corpse

AndroidMK2:
Sounds cool but kinda not xeno styled, it would be sick if the zombies looked more weedy or like being controlled by parasite.

AndroidMK2:
I would change the icon of the larva embryo to something more sinister looking then, like dark purple worm or something like it.

Pill-U:
This seems way to easy to counter with some simple dylovene. Maybe increase the amount of zombium injected over time so you cant fix it with medicine forever?
Alternative suggestion: give hivemind the ability to turn cold marines into ai controlled zombies (maybe with a maximum of about 3 ) and give hivemind the ability to rally them similiar to how blob in normal tg can rally blob spores. This could give marines even more reasons to spend time to recover dead marine bodies. Maybe make hivemind zombies only heal on weeds too.

lKiseki:
Zombium isn't being properly injected by huggers. In a live round, I had an SSD human that was hugged for more than five minutes, it had no zombium when I checked its reagents.

Ahelped it, an admin gave it 200 units of zombium. The guy died but didn't immediately become a zombie, I'm assuming there's a delay but you may want to shorten it since I heard it's five to ten minutes after perma.

David-Stormwell:
Approved by the way, otherwise check for bugs and the normal, but the idea enough is interesting and changes up gameplay enough.

code/modules/projectiles/projectile.dm
```diff
@@ -651,7 +651,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		return FALSE
 	if(lying_angle && src != proj.original_target)
 		return FALSE
-	if((proj.ammo.flags_ammo_behavior & AMMO_XENO) && (isnestedhost(src) || stat == DEAD))
+	if(proj.ammo.flags_ammo_behavior & AMMO_XENO || stat == DEAD)
```
BraveMole, line 654:
```suggestion
	if(stat == DEAD)
```

BraveMole:
Will try something else

PR #8415, opened 2021-09-24
shotguns no longer get miniflamer

Pariah919:
## About The Pull Request
temp fix till hyper get's off his buttocks and makes flamer refactor

## Why It's Good For The Game
this is a dumb combo

## Changelog
:cl:
balance: shotguns don't get miniflamer anymore
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -236,7 +236,6 @@ can cause issues with ammo types getting mixed up during the burst.
 		/obj/item/attachable/magnetic_harness,
 		/obj/item/weapon/gun/pistol/plasma_pistol,
 		/obj/item/weapon/gun/shotgun/combat/masterkey,
-		/obj/item/weapon/gun/flamer/mini_flamer,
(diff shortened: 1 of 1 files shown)


Lefunny:
mini flamers are fine for now, no need for this, cant see nothing wrong with it

Nirvan4:
Good change, too strong to be put into a weapon already strong, if a xeno approaches you to slash you can PB and flame the thing dealing 50 damage ignoring armor, put them on fire with damage over time, area denial and turning the pheros off.

Just the 50 damage is already stupid, its a huge chunk of health to every caste

Al-1ce:
Now that flamethrowers are actually viable attached to guns I can see the issue with this. T-35/flamer is downright bussin

EvilMelonTheSecond:
Shouldn't we just remove the 50 true damage from miniflamer instead?

QualityVan:
Leaving it on the t35?

hyper2snyper:
> Shouldn't we just remove the 50 true damage from miniflamer instead?

shitcode says no.

BraveMole (approved):
This is a bandaid fix that will be reverted when miniflamer will get refactored by bird (soonTM)

PR #8421, opened 2021-09-25
Changes help intent on xenos to more easily clear fire stacks

lKiseki:
## About The Pull Request
Per title. As a Xenomorph, using help intent on another Xenomorph will more easily remove fire stacks.

## Why It's Good For The Game
Fire stacks have become incredibly oppressive as of late. This is my attempt to tackle part of the issue, and promote teamwork amongst xenomorphs; sticking together will be more of an incentive given this change.
Also changes it so you don't have to mash click to clear fire stacks, which sucks.

## Changelog
:cl: Lewdcifer
balance: Fire stacks will now be more easily cleared by help intent.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xenoattacks.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xenoattacks.dm
@@ -83,7 +83,8 @@
 	switch(X.a_intent)
 		if(INTENT_HELP)
 			if(on_fire)
-				fire_stacks = max(fire_stacks - 1, 0)
+				while(fire_stacks && do_after(user, 5, TRUE, src, BUSY_ICON_FRIENDLY))
+    				fire_stacks = max(fire_stacks - 2, 0)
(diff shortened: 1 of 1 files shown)


lKiseki:
Need to test this properly in local, will undraft when I do that.

code/modules/mob/living/carbon/xenomorph/xenoattacks.dm
```diff
@@ -83,7 +83,8 @@
 	switch(X.a_intent)
 		if(INTENT_HELP)
 			if(on_fire)
-				fire_stacks = max(fire_stacks - 1, 0)
+				while(fire_stacks && do_after(X, 5, TRUE, src, BUSY_ICON_FRIENDLY))
```
BraveMole, line 86: it means that the first stack removed will happened 0,5 sec after you first click, rather than being instantanious now.

Is this what you want?

BraveMole (changes requested):

PR #8423, opened 2021-09-25
Halve weed and building cost before shutters open

TiviPlus:
## About The Pull Request

Note i think SHUTTERS_CLOSED is the permanent state of SSmonitor in crash so idk if this would be op there or something
## Why It's Good For The Game

Weeding and building should be allowed to be done more freely before marines arrive to give builder xenos a proper chance to build with less resting needed
## Changelog
:cl:
balance: Halve weed and building cost before shutters open
/:cl:

### DIFF
--- a/code/controllers/subsystem/monitor.dm
+++ b/code/controllers/subsystem/monitor.dm
@@ -44,7 +44,7 @@ SUBSYSTEM_DEF(monitor)
 
 /datum/controller/subsystem/monitor/Initialize(start_timeofday)
 	. = ..()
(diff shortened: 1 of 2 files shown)


David-Stormwell:
Looks good, just wondering how this will affect tadpole and droppod operations?

neohenry:
Make aciding walls, frames, and etc half, too. Would help acid xenos to melt more stuff.

BraveMole (changes requested):
You need to register COMSIG_GLOB_OPEN_SHUTTERS_EARLY as well in monitor, otherwise this will always be active before 12:33 despite shutters being manually opened

BraveMole (approved):

PR #8436, opened 2021-09-26
Fix language holder harddel [DNM]

BraveMole:
This might do nothing, this is an experiment


LoganKele:
what is the point of this

BraveMole:
> what is the point of this

perf increase

PR #8442, opened 2021-09-27
Automatic cycle alamo mode

BraveMole:
## About The Pull Request

Add an automatic cycle mode to alamo. When arriving to its destination, the shuttle will fly back towards its previous destination after X seconds + the time for the engines to be ready.

This is a bounty for elsa.

## Why It's Good For The Game

Easy to configure system, that prevents the whole "please captain cycle ship". You can disable it easily, so it shouldn't be a balance concern

## Changelog
:cl:
qol: Automatic cycle alamo mode
/:cl:

### DIFF
--- a/code/modules/shuttle/marine_dropship.dm
+++ b/code/modules/shuttle/marine_dropship.dm
@@ -609,6 +609,8 @@
 			. = TRUE
 		if("automation_on")
 			M.automatic_cycle_on = params["automation_on"]
+			if(!M.automatic_cycle_on)
+				M.deltimer(M.cycle_timer)
 		if("cycle_time_change")
 			M.time_between_cycle = params["cycle_time_change"]
 
--- a/code/modules/shuttle/marine_dropship.dm
+++ b/code/modules/shuttle/marine_dropship.dm
@@ -221,7 +221,7 @@
 	if(crashing)
 		force = TRUE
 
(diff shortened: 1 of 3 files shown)


code/modules/shuttle/marine_dropship.dm
```diff
@@ -215,8 +221,22 @@
...
+		if(cycle_timer)
+			deltimer(cycle_timer)
+		cycle_timer = addtimer(CALLBACK(src, .proc/prepare_going_to_previous_destination), rechargeTime + time_between_cycle SECONDS - 20 SECONDS, TIMER_STOPPABLE)
+
 	return ..()
 
+///Announce that the dropship will departure soon
+/obj/docking_port/mobile/marine_dropship/proc/prepare_going_to_previous_destination()
+	cycle_timer = addtimer(CALLBACK(src, .proc/go_to_previous_destination), 20 SECONDS, TIMER_STOPPABLE)
+	priority_announce("Dropship taking off in 20 seconds towards [previous.name]", "Dropship Automatic Departure")
```
TiviPlus, line 234: if timer is say 0 seconds would this message not happen very often

BraveMole: Every like three minutes

TiviPlus (approved):

PR #8443, opened 2021-09-27
mild buckshot nerf because it's funny

lovegreenstuff:
## About The Pull Request
buckshot is cracked rn, this makes it slightly less cracked
also for some reason, the pellets shoot out asymmetrically because it had an odd number of pellets shot around the main one.

this is what ren asked for btw
## Why It's Good For The Game
buckshot op, plz nerf

this slight change makes it so that you get hit by 3 pellets at 4 tile range instead of 5 pellets, as well as making it so that bayo+buck+stock pb doesn't one shot elder runners. you have to wait less to actually play the game now

also we should remove flech

## Changelog
:cl:
balance: buckshot total damage lowered from 240 to 220, spread increased from 4 to 5, and individual bullet damage increased slightly from 40 to 44.
/:cl:


BraveMole:
Other nerf is a tad more interesting, we'll see about this one later

PR #8495, opened 2021-10-03
Nerfs DB damage to 85% per shot

lovegreenstuff:
## About The Pull Request
See title

## Why It's Good For The Game

Shotgun is very strong nowadays. I tried to nerf buck, but that got thrown under the bus like an unwanted baby, so hopefully this helps to address to problem. Also, DB was super strong for ages and it's about time it got what was coming to it.

I should probably nerf the parent DB, the colonist one that you can find in Big Red bar and such that shoots faster, but I want to see what the maintainers have to say about it.

Some numbers for you nerds out there.
After testing various things for various reasons, I've found that I can only reload DB fast enough to be on par with a T35 on average. Since I have much more DB experience than the average person, this will likely lead to DB having lower DPS than T35 overall in exchange for higher alpha damage.
## Changelog
:cl:
balance: marine DB damage mod reduced to 85% from 100%.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -441,7 +441,7 @@ can cause issues with ammo types getting mixed up during the burst.
 	accuracy_mult_unwielded = 0.85
 	scatter = 10
 	scatter_unwielded = 40
-	damage_mult = 0.8
+	damage_mult = 0.85
 	recoil = 2
 	recoil_unwielded = 4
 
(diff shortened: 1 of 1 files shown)


Pariah919:
don't personally like this

db trades all capacity for decent burst damage and -15% damage seems like a hefty change to it, considering it still has the 0.7 fire delay

lovegreenstuff:
eh i'll think of somethin else later
this is pretty lazy, gonna change it brb

PR #8506, opened 2021-10-03
Fix zombies getting shot by xeno turrets and stopping game to end

BraveMole:
## About The Pull Request

Title

## Changelog
:cl:
fix: Zombies are no longer shot by xeno turrets
fix: Zombie no longer prevent the game from stopping
/:cl:

### DIFF
--- a/code/datums/gamemodes/game_mode.dm
+++ b/code/datums/gamemodes/game_mode.dm
@@ -463,6 +463,8 @@ GLOBAL_LIST_INIT(bioscan_locations, list(
 				continue
 			if(H.status_flags & XENO_HOST)
 				continue
+			if(H.faction == FACTION_XENO)
+				continue
 			if(isspaceturf(H.loc))
 				continue
 			num_humans++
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -311,6 +311,8 @@
 	. = ..()
 	if(.)
 		return
+	if(ishusk(usr))
+		return
```
TiviPlus, line 315: Why this snowflake check AFTER the UI is opened this should be a trait

code/modules/mob/living/carbon/xenomorph/facehuggers.dm
```diff
@@ -428,6 +428,9 @@
 	if(check_death && stat == DEAD)
 		return FALSE
 
+	if(ishusk(src))
+		return FALSE
```
TiviPlus, line 432: Istype src?
why not check faction

TiviPlus (changes requested):

PR #8511, opened 2021-10-04
Minor wound apply cleanup

TiviPlus:
Per title nothing interesting

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -365,16 +365,18 @@
 	//Creating wound
 	var/wound_type = get_wound_type(type, damage)
 
-	if(wound_type)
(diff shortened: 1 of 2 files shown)


code/modules/organs/limbs.dm
```diff
@@ -365,16 +365,18 @@
...
-				other.merge_wound(W)
-				W = null // to signify that the wound was added
-				break
-		if(W) wounds += W
+	if(!wound_type)
+		return
+	W = new wound_type(damage)
+
+	//Check whether we can add the wound to an existing wound
+	for(var/datum/wound/other in wounds)
```
BraveMole, line 373: AS in ?

BraveMole (changes requested):

PR #8524, opened 2021-10-05
Buff xeno number per marines

BraveMole:
## About The Pull Request

Make each squad marine give more xeno

## Why It's Good For The Game

I suspect https://github.com/tgstation/TerraGov-Marine-Corps/pull/7184 gonna nerf xeno in high pop too much, while it won't help xeno in small pop at all.

Since the former pr is removing some sustain from xeno (less silos), it's important to give them more xeno at the start to compensate for the loss they'll always suffer early game

## Changelog
:cl:
balance: Marine spawning give more xeno points
/:cl:


BrandonSmith6:
This is a good change cus no more 3 v 50 distress games.

AndroidMK2:
I have to admit that there is always not enough xeno slots for usually big marine team which is quite not fun for both sides.

PR #8527, opened 2021-10-05
Fixes energy weapon ammo counter

ivanmixo:
## About The Pull Request

Fixes ammo counter for energy guns because I didn't notice their procs don't call parents.

I forgot this. Wanted to do it as part of a bigger PR coming up but was told to atomise. 

## Why It's Good For The Game

Ammo counter working is neat.

## Changelog
:cl:
fix: ammo counter works for energy weapons
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -352,6 +352,7 @@
 		if(reload_sound)
 			playsound(user, reload_sound, 25, 1, 5)
 		update_icon(user)
+		user.hud_used.update_ammo_hud(user, src)
 	else
 		cell.loc = src
 		update_icon()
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/energy.dm
```diff
@@ -369,14 +370,16 @@
...
-
+	if(user)
+		user.visible_message(span_notice("[user] unloads [cell] from [src]."), span_notice("You unload [cell] from [src]."), null, 4)
+		set_cell(null)
+		playsound(user, unload_sound, 25, 1, 5)
+		update_icon(user)
+		user.hud_used.update_ammo_hud(user, src)
+	else
+		set_cell(null)
+		update_icon()
```
BraveMole, line 382:
```suggestion
	if(user)
		user.visible_message(span_notice("[user] unloads [cell] from [src]."), span_notice("You unload [cell] from [src]."), null, 4)
		playsound(user, unload_sound, 25, 1, 5)
		user.hud_used.update_ammo_hud(user, src)
	set_cell(null)
	update_icon(USER)
```

BraveMole (changes requested):

code/modules/projectiles/guns/energy.dm
ivanmixo: Aha well this won't work correctly. Why? Because the ammo counter will read the cell *before* it's set to null which is inaccurate.

ivanmixo: Can't because 

`user.visible_message(span_notice("[user] unloads [cell] from [src]."), span_notice("You unload [cell] from [src]."), null, 4)`

PR #8537, opened 2021-10-06
Deletes Miniflamer from Laser Carbine

KrabSpider:
## About The Pull Request

Did I stutter? Shotguns lost their ability to have mini-flamethrowers attached to them, ergo the laser carbine also gets the boot. 

## Why It's Good For The Game
Consistency. Shotguns lost their right to administer a baptism by fire, therefore the laser carbine does as well due to the projectiles being hitscan and quite simply dealing a **truckload** of damage to any mobs within ranges where the miniflamer is optimal.

## Changelog
:cl:
balance: Revokes the Laser Carbine's license to provide a fire baptism.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -563,7 +563,6 @@
 		/obj/item/attachable/magnetic_harness,
 		/obj/item/attachable/scope/marine,
 		/obj/item/attachable/scope/mini,
-		/obj/item/weapon/gun/flamer/mini_flamer,
 		/obj/item/attachable/motiondetector,
 		/obj/item/attachable/buildasentry,
 	)


ilikesaying:
mfw laser hitscan TM
also this will affect carbine as a regular auto/burst weapon too

BrandonSmith6:
if miniflamethrows are being removed from this many guns just remove them to make sure there is no confusion.

KrabSpider:
> I'd think separating the carbine's shotgun aspect of it is better than just doing this.
@Pixelknightly 
Not...really? That would ultimately just make it a somewhat inferior version of the laser rifle, and making an entirely new lasergun with uses **only** spread ammo I feel would just create needless bloat.

Pixelknightly:
> > Not...really? That would ultimately just make it a somewhat inferior version of the laser rifle, and making an entirely new lasergun with uses **only** spread ammo I feel would just create needless bloat.

No it doesn't. For the same reason the t-12 doesn't replace the t-18. The t-18's faster to wield and has a fucking killer burst. The laser carbine might not have that burst, but it's definitely faster to wield and still has a niche for people who need it.

The mini-flamer's strong enough that it's essentially a reason to use the carbine. Separate the carbine's shotgun and auto aspects, suddenly, you got people using carbine's auto just to use the mini-flamer with it.

Besides, we literally have a rifle that's the in-between of the t-18 and the t-64 despite the t-64 being an in-between of the t-18 and the DMR.

And that's not even mentioning how the TX-11 is basically just the t-18 with more ammo but no autoburst.

PR #8541, opened 2021-10-06
Lets xenomorphs use guns

jroinc:
## About The Pull Request

Toy guns, also all toys such as plushies when the fun config is set.

Also allows carrying and use of all foods, all utensils, all drink bottles, and drinking glasses when the fun config is set. And I'm about to add coffee use if it's not covered already. 

And maybe vendor use, but I think I'll leave it as-is.

A few of these items have Force 5, one or two have Force 10, and it's possible to knock someone down for one whole second by hitting them on the head with a glass bottle. I think this is probably acceptable

## Why It's Good For The Game

It's not. Please add it anyway.

## Changelog
:cl:
expansion: If the fun config is set, ayys can interact with food, drink, toys, and utensils. ROUNY WITH A PLASTIC FORK WHAT'S HE GONNA DO
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/food/cans.dm
+++ b/code/game/objects/items/reagent_containers/food/cans.dm
@@ -3,6 +3,11 @@
 	init_reagent_flags = NONE
 	var/canopened = FALSE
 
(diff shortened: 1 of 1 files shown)


BrandonSmith6:
funny, yes

PR #8549, opened 2021-10-07
Fix combat robots speed due to not having shoes.

SplinterGP:
## About The Pull Request
This fixes combat robots having 1 whole more slowdown due to not having shoes they can wear(which gives -1 slowdown which speeds you up), tivi thought they had done it with the armor but they confused -1 with -0.1

## Why It's Good For The Game
Fixes a problem with robots speed that was unintended.

## Changelog
:cl:
fix: Combat robots(Species) doens't have slowdown from not wearing boots anymore.
/:cl:


InterroLouis:
Just give them boots instead.

PR #8553, opened 2021-10-08
Pizza factory is a little more profitable

Eskjjlj:
## About The Pull Request
This raises the profits from selling pizza from 1 point to 3 points.

## Why It's Good For The Game
The cost of the machines for the pizza factory is 30 points so with only 1 point of profit you would need to make and sell pizzas 30 times to recover the initial investment. Considering it takes about a little more than a minute to buy, process and sell a batch of pizza it takes way too long to turn in a meaningful profit and expand the factory.

With this change you would get 3 points of profit per pizza batch so only 10 batches are needed to recover the initial costs which is more reasonable without being OP.

## Changelog
:cl:
balance: Selling pizza is a little more profitable now! 
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -1708,7 +1708,7 @@ FACTORY
 /datum/supply_packs/factory/pizzarefill
 	name = "Nanotrasen \"Eat healthy!\" margerita pizza kit refill"
 	contains = list(/obj/item/factory_refill/pizza)
-	cost = 29 //allows a one point profit if all pizzas are processed and sold back to ASRS
+	cost = 27 //allows a three point profit if all pizzas are processed and sold back to ASRS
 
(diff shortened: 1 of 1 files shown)


Eskjjlj:
> 
> 
> > It takes longer if you take into account the time it takes to buy it on the req console, raise the ASRS pad, load the pizza in the machines manually, wait for the pizza to move along the conveyor lines and lowering the ASRS pad.
> 
> I take into account none of that time, because any RO worth his salt will just automate that shit and not worry about it. Set up the conveyor to dump pizzas into the ASRS the moment the railings go down, buy multiple pizzas at a time and use that to save time, and so on.

It's 29 points for a batch of pizza, only for a 1 point profit. Say you want to do five batches at a time it's 145 points in use only for 5 points of profit. Not worth it.

With the current numbers there is no point in making a pizza factory unless you want to grief.

PR #8557, opened 2021-10-08
Makes throwing knives viable.

hyper2snyper:
## About The Pull Request

Throwing knives now come in stacks of 5, the belt holds 6 stacks for a total of 30 knives a belt.
Knife damage is the same as before, however when holding a stack, clicking will throw one of the stack. Stacks can be refilled by clicking on another stack with a stack in hand.

## Why It's Good For The Game

This makes throwing knives a viable weapon. Itll be cool i swear.

## Changelog
:cl:
expansion: Throwing knives are now stackable and hold 5 a stack, clicking a tile with a stack in hand will throw one of the stack.
/:cl:


TiviPlus:
Why isnt this an item/stack child

PR #8561, opened 2021-10-09
Heal speed of Bica/Kelo/Tricord down to x0.75, duration increased to match total over time

Losenis:
## About The Pull Request

Halves healing rate of Bicaridine, Kelotane, and Tricordrazine. Metabolization rate halved to match the per-unit total healing effect.

18/10/2021
- effect_str and custom_metabolism upped to 0.75. Description is from old 0.5 values.
## Why It's Good For The Game

A core issue that has been going for a while is that marines can heal while in combat while xenos need to run away from combat and find weeds to rest on, something they can't do if they're getting chased (which the marine can also heal during). This can make it extremely difficult for xenos to do much if the damage they're able to deal is simply not enough to get past that healing effect, while marines can constantly deal damage that will force xenos to fall back. This also affects things like acid turrets, sometimes rendered a 4-second delay tool at most due to marines healing while shooting at it.

Bonefracs and organ damage most of the time only happen way after combat had ended, at which point the xeno either A) Killed the marine, B) Ran, C) Died, making their existence mostly irrelevant to the issue here.

Halving the healing rate will mitigate this somewhat by making mid-combat healing not as extreme, while also keeping the total healing and OD of said reagents the same so as to keep the math intact on a per-unit basis. Marines would hopefully need a resting period to actually heal, rather than fighting and getting healed even over bonefrac thresholds while xenos have to run and subsequently get chased as the marine now heals to full. It'd also make it possible to actually chip away at their health in stalemates, ideally.

Only touches the main three healing chems marines can all pack without chem intervention, doesn't touch meralyne/dermaline/paracetamol.

Should probably be TMd thoroughly before thinking of merging.

## Changelog
:cl:
balance: Effects of Bicaridine, Kelotane, and Tricordrazine halved. Duration doubled.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -202,8 +202,8 @@
 	purge_rate = 1
 	overdose_threshold = REAGENTS_OVERDOSE
 	overdose_crit_threshold = REAGENTS_OVERDOSE_CRITICAL
-	effect_str = 0.5
-	custom_metabolism = REAGENTS_METABOLISM * 0.5
+	effect_str = 0.75
+	custom_metabolism = REAGENTS_METABOLISM * 0.75
 
 /datum/reagent/medicine/kelotane/on_mob_life(mob/living/L, metabolism)
 	var/target_temp = L.get_standard_bodytemperature()
@@ -305,8 +305,8 @@
 	overdose_threshold = REAGENTS_OVERDOSE
 	overdose_crit_threshold = REAGENTS_OVERDOSE_CRITICAL
 	taste_description = "grossness"
-	effect_str = 0.5
-	custom_metabolism = REAGENTS_METABOLISM * 0.5
+	effect_str = 0.75
+	custom_metabolism = REAGENTS_METABOLISM * 0.75
 
 /datum/reagent/medicine/tricordrazine/on_mob_life(mob/living/L, metabolism)
 
@@ -635,8 +635,8 @@
 	overdose_threshold = REAGENTS_OVERDOSE
 	overdose_crit_threshold = REAGENTS_OVERDOSE_CRITICAL
 	scannable = TRUE
-	effect_str = 0.5
-	custom_metabolism = REAGENTS_METABOLISM * 0.5
+	effect_str = 0.75
+	custom_metabolism = REAGENTS_METABOLISM * 0.75
 
 /datum/reagent/medicine/bicaridine/on_mob_life(mob/living/L, metabolism)
 	L.heal_limb_damage(effect_str, 0)


SplinterGP:
Not to mention combat autoinjectors exist, which most usefullness is that theyre instant apply and come mixed, so the intent of them existing is being used for that, as opposed to pills which have way more unit per storage but is supposedly harder to use, in the long run i believe this won't actually change anything for the reason this PR exists, much for the contrary, its gonna nerf combat autoinjectors while pills albeit healing slower will still do what theyre good for, being on meds all time which is very good on the long run.

Losenis:
>Not to mention combat autoinjectors exist, which most usefullness is that theyre instant apply and come mixed

How does this pr change that? Combat autoinjectors are just bicaridine, kelotane, tramadol, and a tick of ryetalin. Half of those aren't even touched, let alone the mixture. Not to mention they're refillable. If what you're saying is that their healing rate will get nerfed, then yeah, that's kinda the point of the pr.

SplinterGP:
> 
> 
> > Not to mention combat autoinjectors exist, which most usefullness is that theyre instant apply and come mixed
> 
> How does this pr change that? Combat autoinjectors are just bicaridine, kelotane, tramadol, and a tick of ryetalin. Half of those aren't even touched, let alone the mixture. Not to mention they're refillable. If what you're saying is that their healing rate will get nerfed, then yeah, that's kinda the point of the pr.

thats literally what i mean, combat autoinjectors are literally meant to be used in combat, and i dont see why nerfing exactly the use of those chems mid combat when thats not a big issue at all when the actual issue is pills and not the chems itself, the actual effect chems have in combat is not noticeable by healing too fast, instead theyre noticeable by having a CONSTANT supply of chems that pills give you, letting you have heals for basically unlimited time, meanwhile combat injectors dont have that since you cant carry many/dont have many u per injector, if something should be nerfed instead its pill availability/pill on combat usein constant pre-med.

https://user-images.githubusercontent.com/29824735/136674529-315e2fa6-0946-4f69-8250-ed3c51365c73.mp4

Losenis:
Nerfing healing speed is intended. Everything else about combat injectors remains untouched. If they're not competitive over pills, this pr didn't change anything because it affects both in equal measure, you're just finding out that they weren't competitive in the first place. I also wouldn't nerf availability of chems because that means either corpsmen get shafted or latejoiners find out vendors got Black Friday'd by roundstart players, and that's not really fun. Not only that, some xenos are built with anti-chem like boiler and defiler, they can be buffed in that role too if need be.

And what's the video even meant to show?

S1tr3p:
In my personal opinion, this doesn't fix the issue at hand.

If you want to actually target healing in combat, which I don't see happening very often due to the close quarters nature of the current meta, and with how maturity tower has allowed xenomorphs to become older which allows for more slash damage at earlier times this feels like a bottom tier issue.

If you want to target the issue at hand, make autoinjectors bigger and lessen the amount of pill bottles, or even, limit pills to medical staff and keep autoinjectors for other roles. Pills are the issue that you want to target due to them having 16 per bottle with a set amount, bica for example would be 16 x 15 which equals out to 240 Units in one bottle, whereas the average autoinjector has 15-30 units per.

I think this will cause many issues and allow for xenos to completely dominate and target medics (more than they already are, that is.) who are usually carrying advanced trauma and burn kits that lessen the damage and pain from wounds faster and more efficiently than pills or autoinjectors. Overall, very conflicting change to introduce in an already odd time for both sides of the spectrum.

SplinterGP:
> Jesus splinter, learn how to makes sentences.
> 
> There is zero spacing, zero new sentences. This is not readable 
> 
> 

Sorry english is not my main language :(

EvilMelonTheSecond:
As much as I don't agree with the whole "combatmedding OP" sentiment people seem to have I would rather have slower healing as a nerf than having crippling debuffs from meds or turning med into a chore with more perma damage.

If marine healing needs a nerf I vote for this option.

jroinc:
As "the chem guy", I can think of better ways to do this, but I'd still like to see how this plays out.

lbnesquik:
Wouldn't this make premed stronger given the duration is longer?

PR #8564, opened 2021-10-10
Buffs most Xenomorphs' armor

lKiseki:
## About The Pull Request
Per title. Most xenomorphs have been given +5 to all armor values across the board, and some armor values were standardized to follow increments of 5 rather than seemingly random values. Some exceptions were made here and there for certain castes such as Runner and Sentinel, specifics can be found in each commit's description. Hivemind did not receive any changes because funni Hivemind.

Talked it out with David before making the PR; I have his approval for giving +5 armor to all castes, but value standardization was my own doing. If the latter is not okay, I can change the values to not be standardized, as per a Maintainer's criteria.

## Why It's Good For The Game
Helps xenomorphs out a little given how they've been struggling in the current state of balance and standardizes armor values to follow increments of 5 so they're more easily balanced.

## Changelog
:cl: Lewdcifer
balance: All xenomorphs have buffed armor values.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/praetorian/castedatum_praetorian.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/praetorian/castedatum_praetorian.dm
@@ -30,7 +30,7 @@
 	caste_flags = CASTE_CAN_BE_QUEEN_HEALED|CASTE_CAN_BE_GIVEN_PLASMA|CASTE_CAN_BE_LEADER|CAN_BECOME_KING
 
 	// *** Defense *** //
-	soft_armor = list("melee" = 30, "bullet" = 35, "laser" = 35, "energy" = 35, "bomb" = XENO_BOMB_RESIST_0, "bio" = 28, "rad" = 28, "fire" = 35, "acid" = 28)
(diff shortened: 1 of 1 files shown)


David-Stormwell:
me when lewd makes a commit for each caste 🐞

Ragecrabbo:
Better solution than just giving xenos extra health since sunder is a thing and i like the standardization. Something to keep in mind though is that xenos with already high armor will benefit from this alot more though than the ones on the lower end.

lKiseki:
> Better solution than just giving xenos extra health since sunder is a thing and i like the standardization. Something to keep in mind though is that xenos with already high armor will benefit from this alot more than the ones on the lower end (95% from 90% for example means you essentially take half of the damage from before).

Tru, but I mean, sunder is a thing and it's been becoming increasingly common, so shrug.

Alexeivichi:
Good now buff their damage and make marines wear cardboard robo-armor.

SplinterGP:
> -Ancient crusher goes from 75 to 80 bullet armor. This is a 25% damage decrease, just like that
To note thats a 25% damage decrease compared to the damage dealt post armor, compared to the base damage thats just a 5% decrease.

MiniMeatwad:
From projectile.dm
damage = max(0, damage - (damage * living_soft_armor * 0.01))

PR #8571, opened 2021-10-11
Reduces shield integrity by 150

UntoldTactics:
## About The Pull Request
Reduces shield integrity by 150, making it 250 from a previous 400. 

## Why It's Good For The Game
Shield is absurdly tough and when combined with certain one handed weapons makes for a deadly combo. The fact that a shield user cannot die to even a ravager without about a minute and a half worth of just constant slashing (assuming the shield marine has not premedded) is pretty absurd to me. Combined with other marines the shield marine also becomes an insane bastion of just negating xeno attacks, blocking them from getting to the more squishier marines. Reducing shield integrity means that shields are still strong, viable and capable, but need to be maintained more often and also makes it actually possible for xenos to break the shield in combat, as previously this seemed to never happen. I'd balance it another way by reducing its defensive values perhaps, but looking at the shield code made me scared. So no thanks.

Xenos are melee focused, having something that nearly entirely nullifies this is stupid. 

## Changelog
:cl:
balance: Reduced shield integrity to 250 from a previous 400.

/:cl:

### DIFF
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -137,7 +137,7 @@
 	icon = 'icons/obj/items/weapons.dmi'
 	icon_state = "marine_shield"
 	flags_equip_slot = ITEM_SLOT_BACK
-	max_integrity = 400
+	max_integrity = 250
 	integrity_failure = 100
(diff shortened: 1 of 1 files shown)


BraveMole:
I already said i'll block every shield nerf. Turtles are a danger to no-one, you can just shuffle through them and attack the other marines

PR #8576, opened 2021-10-11
New screen orders for TGMC command

TiviPlus:
## About The Pull Request
https://cdn.discordapp.com/attachments/500801682058379265/897214591057162320/Change_Me_-_Debugdalus_2021-10-11_22-02-52.mp4

## Why It's Good For The Game

Srcreen-based animated alerts make it more obvious to a player where their attention should be than the chat
## Changelog
:cl:
add: New screen-based orders for TGMC command
/:cl:

### DIFF
--- a/code/modules/screen_alert/command_alert.dm
+++ b/code/modules/screen_alert/command_alert.dm
@@ -37,6 +37,8 @@
 	if(!can_use_action())
 		return
 	var/text = stripped_input(human_owner, "Maximum message length [MAX_COMMAND_MESSAGE_LEN]", "Send message to squad", max_length = MAX_COMMAND_MESSAGE_LEN)
+	if(!text)
+		return
 	if(CHAT_FILTER_CHECK(text))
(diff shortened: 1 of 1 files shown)


HeyMyNameIsLouie:
Is it possible for PO's to also have this ability? with the addition of auto alamo, PO's job has slowly gone from babysitting it to more active command-related roles by being eyes on the sky via CAS and having a physical mobile outpost in the form of the tadpole. 
Also, I think it would be fitting since PO's are captains of their own ship, the tad and condor to an extent, an extension of the captains reach. 
I know they are still technically considered auxiliary staff but wouldn't it be a good to slowly evolve them towards a lesser form of the FC? Especially since their decisions and chocies with how they both man tad and cas greatly affect the battlefield. They could divert attention to tadpole landing just south of push, CAS being online and announcing what kind of payload it's packing, etc., etc.
or maybe an entirely separate scroll text-ability is given to them where instead of "Orders Updated:" its something along the lines of "PO Message Relay", "SO Message Relay" so other jobs that are important apart from the SL's, FC's and Cpt's have this pretty neat ability. It really makes herding marines a lot easier and effective

_This is totally not me begging you to add this so I can do a bit of trolling with the orders._


my tad

PR #8588, opened 2021-10-15
Drag nerfs

jroinc:
## About The Pull Request

Dragging any mob, by anyone, has a default two-second delay. Fireman carry timer was lowered by two seconds to keep the timer consistent.

## Why It's Good For The Game

It's too easy to insta-rescue dead marines that are in dubious situations. Someone runs in, clicks once, and 0.1 seconds later, the marine is 2 tiles closer to safety and moving closer every half-second, which ayys can't undo.

This changes it so you need to either stand next to the marine for two seconds, OR use some of the many, many tools that ungas get (ideally rollerbeds as you can deploy them beforehand and run up for the instagrab, but I know ungas will go straight to the powerfist and grenades... which ARE valid options).

This enables some counterplay to insta-recovery of ungas, as you can't shoot through a rollerbed if it's buckled, and it can be destroyed in a few hits by xenos (and powerfist/grenades do damage), but it's still very much possible.

And yes, ungas get something out of it, which is the exact same two-second delay on ayys dragging each other OR crit ungas.

## Changelog
:cl:
add: Passive grab on mobs takes two seconds to do. This affects all mobs, but not objects such as rollerbeds.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -774,7 +774,7 @@
 		return
 	visible_message(span_notice("[src] starts lifting [target] onto [p_their()] back..."),
 	span_notice("You start to lift [target] onto your back..."))
-	var/delay = 5 SECONDS - LERP(0 SECONDS, 4 SECONDS, skills.getPercent("medical", SKILL_MEDICAL_MASTER))
+	var/delay = 3 SECONDS - LERP(0 SECONDS, 4 SECONDS, skills.getPercent("medical", SKILL_MEDICAL_MASTER))
 	if(!do_mob(src, target, delay, target_display = BUSY_ICON_HOSTILE))
 		visible_message(span_warning("[src] fails to fireman carry [target]!"))
 		return


jroinc:
> 
> 
> Allright so, just formalising what we discussed on discord:
> 
>     * Xenos dragging xenos should not be affected by this, because dragging crit xeno, dragging resting xeno, etc. is a big part of xeno gameplay.
> 
>     * Marine dragging xeno should not be affected by this. Especially since dragging dead minions is already a pain in the ass
> 
>     * I don't like it being a delay, i'd rather it being a "you're shot, you drop what you are carrying"

Dubious on the first one, fine on the second one, not gonna do the third one (but if you wanna PR it, we can do that instead).
I'll fix this and change it as listed if it gets PL approval.

DeltaFire15:
Note that in its current state, this'd effectively make Warrior's Lunge not useable at all. Lunge uses the same proc to initiate its grab, and good luck having them stand still for two seconds when the stun is shorter aswell as it being quite unlikely to survive two seconds in melee completely still.

jroinc:
Aight. I'll take that as approval and fix this later. I'll make it so it only affects ungas grabbing ungas when I do as that will fix both.

Any thoughts on me dropping the delay down to one second if it only affects ungas? The goal is to stop insta-rescues in a combat zone, not make dragging impossible.

Kesserline:
I like the global idea. Xenos have real difficulties to secure marine kills. Which leads to comic situation like "Eh, I killed this dude 2 minutes ago, and he is already back on the Frontline."

BrandonSmith6:
> if marines have a drag cooldown so should xenos. It will help with the cringe dragging dead marines to brazil mindset.
> 
> yeah but its the only way for xenos to prevent an instant marine rescue and give them a chance to psydrain

Psydrain can be done pretty fast. For dragging Im not talking about dragging a couple tiles around for cover to psydrain. Im talking about dragging multiple dozen tiles to the middle of no where just to make sure no one finds them. Most people that drag marines to Brazil don't even psydrain. The only thing on their mind is making sure the player has 0 chance to come back into the round. That's cringe.

Losenis:
Would this also affect other things like trying to grab someone to put them on an operating table/autodoc/bodyscanner/dead unga onto medevac or rollerbed?

DeltaFire15:
Hmmm.. what if it only applied to dead mobs then, for both sides? That way normally carrying and combat would not be inhibited, but recovering dead of either side is more difficult, be it marines rescuing dead and xenos trying to drag them off during Crash / Hunt or xenos trying to deny points via recovering dead allies.
Obviously just spitballing ideas.

PR #8606, opened 2021-10-17
Premed nerf, combatmed buff

jroinc:
## About The Pull Request

Changes kelotane and bicard to heal based off a curve. The first 5u heals 1/tick as normal, the next 5u heals a decreasing amount from 1.0 to 0.5/tick, and the next 5u and any more heal a flat 0.5/tick. The high-dose bonus (0.5 healrate if above 20u) remains unchanged. 

Also tricord got nerfed to a flat 0.5 healrate (and 0.25 toxheal) rather than 0.8... so it's 100% as good at total damageheal as a specialized med, not 160%.

Autoinjectors will still heal 25 points per inject, pills will now heal 56.25 points per pill (down from 75), and any chained pills will heal 37.5 points per pill.

That's the nerf. The buff is that addition of bicard or kelo now heals 20% of your brute/burn damage (plus five free points) *instantly* upon mob add, which can be substantial if you're hurt. The break-even point for pills to get equal or greater total heal than oldpills is ~68 health, while autoinjectors have the same performance and get the healrate changes as a straight bonus.

To prevent the obvious exploit of "well I'll just drink a buncha water and keep hittin' myself with a 1u bicakelo injector", ungas now get a *five* second stun when bicard and kelo leave your body, reduced by 0.5 seconds per tick it was in you. This can benefit ayys slightly, but in the worst-case scenario of you popping a standard 15u combat autoinjector *while* you're neurotoxined... 

...you still won't get stunned. 

It'll only get you if you have a microdose injector of less than 10u *and* are actively purging, or if you use an injector with less than 2u in it. Now, a *medic* running a microdose setup might be able to heal large quantities of damage rapidly on someone else at the expense of micromanagement and stunning the victim... but that's a different subject and one I'm more-or-less OK with.

## Why It's Good For The Game

Premed and being on nonstop pills is no longer always the best option. You can now make a choice about whether to be on constant but slower healing, or to be able to burst-heal for more total damage, at the cost of having to take actions to do this.

## Changelog
:cl:
balance: Kelo and bicard now heal less damage over time, but heal some up-front. Tricord healing cut to 0.5/tick (0.25 tox).
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -203,9 +203,14 @@
 	overdose_threshold = REAGENTS_OVERDOSE
 	overdose_crit_threshold = REAGENTS_OVERDOSE_CRITICAL
 
+/datum/reagent/medicine/kelotane/on_mob_add(mob/living/L, metabolism)
+	var/fire_loss = L.getFireLoss()
+	L.heal_limb_damage(0, 0.2*effect_str*fire_loss(+5))
+	to_chat(L, span_warning("Your wounds begin to rapidly knit together!"))
+
 /datum/reagent/medicine/kelotane/on_mob_life(mob/living/L, metabolism)
 	var/target_temp = L.get_standard_bodytemperature()
-	L.heal_limb_damage(0, effect_str)
+
 	if(L.bodytemperature > target_temp)
 		L.adjust_bodytemperature(-2.5*TEMPERATURE_DAMAGE_COEFFICIENT*effect_str, target_temp)
 	if(volume > 10)
@@ -213,7 +218,20 @@
 	if(volume > 20)
 		L.reagent_pain_modifier -= PAIN_REDUCTION_LIGHT
 		L.heal_limb_damage(0, 0.5*effect_str)
+	switch(current_cycle)
+		if(1 to 25)
+			L.heal_limb_damage(0, effect_str)
+		if(26 to 50)
+			L.heal_limb_damage(0, (-0.02 * current_cycle + 1.5)*effect_str)
+		if(51 to INFINITY)
+			L.heal_limb_damage(0, 0.5*effect_str)
 	return ..()
+
+/datum/reagent/medicine/kelotane/on_mob_delete(mob/living/L, metabolism)
+	if(current_cycle > 10)
+		return
+	to_chat(L, span_userdanger("You reel as your medicine stops affecting you!"))
+	L.Paralyze((5*current_cycle)-50)
 
 /datum/reagent/medicine/kelotane/overdose_process(mob/living/L, metabolism)
 	L.apply_damages(effect_str, 0, effect_str)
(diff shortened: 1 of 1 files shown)


Captiajoe:
What about combat pops?

Mrrpip:
Cue marines with blowguns on HvH

neohenry:
so sad

Losenis:
This changes way too much for what it's trying to accomplish, and adds weird unnecessary stuff like stuns that'd in practice not even happen, unless you're in niche situations where you're SoL since not everything is guaranteed to work as intended. I fear this might be going the way of normal tg where chems get gradually more and more complicated to the point for some people it starts being a reason to not delve into them, which is an actual concern in TGMC with how few competent corpsmen there are.

Nothing against the creator, but if I were to join a game and see that this changed this much and had to relearn it, I'd probably shelve it for another day and play some other whatever role or just another game. One thing is brainstorming new mechanics from the comfort of one's free time in a calm day, another is wanting to play something and finding out that you have to relearn or either be blind or not play. It's partly why I hardly even play tg anymore other than going janitor and doing little for the round duration then hiatus again.

Kesserline:
I see a few big design flaws, which according to the most recent devs gamechanges are not aiming the same direction at all :
Recently, most devs changes aimed at removing cadehugging for marines, removing stalemates and increasing fun to play for each sides.

If you nerf that hard the medheal part, there will be more than just "premed is nerfed" consequences :
- 1 : You will have an increase in cadehugging, as marines will tend to more and more be immobilized to be treated.
- 2 : Most marines pushes will be even slower, easier to be ambushed and stopped by xenos, encouraging again, consquence 1.
- 3 : You will increase the complexity of meds, which are, when you check the corpsmen playerbase, already too hard for them.
- 4 : I am not even talking about the stun part, which is pure unfun, especially on a videogame.

For those 4 reasons, this PR will affect the roundflow in a negative way. Both for marines and xenos, as it will result in more stalemates and cade/fobhugging. On top of being a pure lack of fun and QoL for marines/corpsmen actually interested in using meds. (And I'm still not talking about the  5 sec nstun part of this pr)

Dolth:
Terrible idea, as Mark said you're gonna encourage marines to cadehug and you'll slow down pushes at such a point the momentum will get crushed even faster. And yes, why the hell a 5 second stun? WORSE CASE SCENARIO: Make them create toxin or negate themselves or purge but why the hell a 5 second stun? Anyway this is just bad for said reasons...

jroinc:
> 
> 
> > 5 seconds stun when bica and kelo leaves your body
> 
> oh boy I can't wait to fall back from this failing push-
> 
> gets stunned and ganged by the impending 7 xeno push.
> 
> There is seriously too much being nerfed and change in this one PR to change premeds. If you want to kill premeds that badly then kill pill bottles for marines, its a seriously easy change that forces marines to be smart with their meds rather than getting owned by stuns.

You will never hit that stun, even with purgchems in you, unless you play around with microdose custom meds.

I can't kill pill bottles for marines because they always find a way to get more. It'd mean killing every pill bottle on the map, locking the chem dispensers, and physically stopping medics/MD's from handing them out. I don't even *want* to do that.

jroinc:
> 
> 
> Terrible idea, as Mark said you're gonna encourage marines to cadehug and you'll slow down pushes at such a point the momentum will get crushed even faster. And yes, why the hell a 5 second stun? WORSE CASE SCENARIO: Make them create toxin or negate themselves or purge but why the hell a 5 second stun? Anyway this is just bad for said reasons...

See above posts.

jroinc:
> 
> 
> > I don't give a shit about fun.
> 
> People playing for fun: image

K bye.

PR #8616, opened 2021-10-20
Xeno door health buff

jroinc:
## About The Pull Request

Xeno door health buffed from 50 to 100. Armor left at 0.

## Why It's Good For The Game

It takes two bullets to kill a door currently, you can shoot them down faster than you can walk with an autofire gun. This gives them the half HP as a wall, but not the armor, so they have less than a quarter the effective HP (walls have an effective 500 HP). Additionally, flamer fire still does a buncha extra damage to doors and should kill them in two ticks.

Now it'll take four bullets to kill a door.

I also do not believe that doors heal.

## Changelog
:cl:

balance: ayy doors aren't paper-machie anymore

/:cl:

### DIFF
--- a/code/game/objects/structures/xeno.dm
+++ b/code/game/objects/structures/xeno.dm
@@ -120,7 +120,7 @@
 	icon = 'icons/Xeno/Effects.dmi'
 	hardness = 1.5
 	layer = RESIN_STRUCTURE_LAYER
-	max_integrity = 200
+	max_integrity = 100
 	var/close_delay = 10 SECONDS
 
 	tiles_with = list(/turf/closed, /obj/structure/mineral_door/resin)


EvilMelonTheSecond:
200 is a bit much for doors don't you think? Xeno doors open instantly and give xenos the element of surprise unless marines destroy it. Sturdy doors will just end with door spamming instead of mazes.

I would prefer if doors were made stronger against bullets but not melee and flames. They give a fair initiative advantage to xenos so I don't think a buff is necessary anyhow.

BraveMole:
That's still multiplying by 4 the current HP

Losenis:
Could also just decrease the time or plasma cost they need to be built. Compared to walls, doors are stupidly expensive even past their utility.

Losenis:
I had also forgotten to mention, doors collapse if there's no structure next to them, so in some instances they're not just weak, they straight up selfdestruct.

PR #8619, opened 2021-10-20
Razorburn

jroinc:
## About The Pull Request

Some changes to nanites. First, the base metabolism rate was upped from 0.1u to 1u so that incidentally touching it won't be 100% fatal.

Next, damage was halved as long as you're below 5u (5 ticks) of total exposure. This should make briefly touching it less bad. Additionally, after 5 ticks, metabolism rate starts increasing so it washes out faster if you managed to get like a million nanites in you.

Finally, a warning message was added so that you KNOW you're stepping in it.

With this in mind, I undid the earlier razorburn nerf. You can now run beakerrigs again, but you'll have to stand in the middle of the foam and take the damage as stated.

~~also if you somehow manage to go over 100u, it gibs you and turns you into a razorwire. Standing inside a razorwire canister for the full duration adds 62.5 nanites, so this should be impossible to reach in gameplay but i still want it cause it's hilarious~~

## Why It's Good For The Game

Razorburn is less instantly lethal if you touch it, while still being nothing to laugh at. ~~This should NOT be merged or TM'd until the foam refactor's in, and I'll update it to fit when that's the case.~~ foam refactor in

https://github.com/tgstation/TerraGov-Marine-Corps/pull/8550

## Changelog
:cl:
balance: Razorburn is less immediately lethal now
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -441,7 +441,7 @@
 	if(current_cycle > 5)
 		L.apply_damages(2.5*effect_str, 1.5*effect_str, 1.5*effect_str)
 		L.blood_volume -= 5
-		holder.remove_reagent(/datum/reagent/toxin/nanites, ((current_cycle * 0.2) - 1)
+		holder.remove_reagent(/datum/reagent/toxin/nanites, (current_cycle * 0.2) - 1)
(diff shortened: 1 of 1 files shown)


jroinc:
aight, the foam one's in and this needs no changes to fit in. Yay

PR #8629, opened 2021-10-21
packets in vendors

jroinc:
## About The Pull Request

kills vendor pill bottles of the big 4
replaces with proportional numbers of pill packets

## Why It's Good For The Game

maybe every unga won't be an infinite walking pharmacy that thinks nothing of purging all their drugs than instantly re-medding

also maybe ungas won't take half the medvendor's pills and die with them all unused

also i guess bravemole asked for it

## Changelog
:cl:
balance: makes ungas break into medbay roundstart again
/:cl:


QualityVan:
Just to be sure, you're intentionally increasing the total chems available? BKT packets are half a bottle each, tricord is 1/3.

PR #8644, opened 2021-10-22
Pain rebalance

jroinc:
## About The Pull Request

This PR should rework all of pain and shock, but before I do that, I need to understand it (and ideally let everyone ELSE understand it too). 

**So, to start, here's the CURRENT status of pain and how pain works before I start tweaking things-**

All ticks are 2-second lifeticks.

Pain is handled in /code/modules/mob/living/carbon/shock.dm

This file takes all of your direct physical damage, adds it up based on a ratio (1.2 points per burn, 1 point per brute and cloneloss, and 0.75 per point of oxyloss or toxloss) and then adds the total amount of chemical pain (handled in reagants, usually negative) together, *then* adds the total amount of limb pain (40 if necrotized or delimbed, then 30 per break, unless splinted or valk'd, then it's 15 per), THEN adds 1.5 pain per point of organ damage (and 0.05 per germ if you have an infection),  THEN subtracts 40-80 pain for Hold orders... and that's pretty much it.

This blob of code generates your traumatic_shock var, which does nothing directly. Let's jump to-

code/modules/mob/living/carbon/human/life/handle_shock.dm

-which is where most of the effects are handled. This takes your traumatic_shock and uses it to increment your shock_stage  based on a list. If your traumatic_shock is below 99, you have no issues, and in fact gradually *lose* shock_stage (reference example- An unsplinted broken chest and 60 damage is 90 traumatic_shock, so you'll never gain *any* shock_stage!).

The magic numbers are 100 to 150 which adds 1 to shock_stage, 150-200 which adds 2 to shock_stage, and 200+, which *immediately* sets you to 150, and adds 5 per tick. That's how you *get* shock_stage, which is the "mechanical" part of pain.

So, shock_stage. Shock_stage is handled in two places. The first one is back in carbon/shock.dm. It adds and removes a movedelay from 1 to 2 to 3 (10 shock_stage-30, 30-50, 50+) as shock_stage changes. The second one is in life/handle_shock.dm. It sends *scary messages* if you're below 30 shock_stage, adds a bit of eye blur at 30-40, then it starts adding gradually escalating stagger and slowdown up until 150, where you get a one second stun every 10 seconds.

Shock_stage is also capped to 200 and resets on death.

That is how pain *currently* works. Here are the big takeaways-

-*Big dick 80 painkiller Tramadol* is too strong and needs to be nerfed lower. Maybe 50 or so, gotta do math and compare the defines.
-The *instant* jump to 150 shock_stage if over 200 pain needs to be removed, or pushed higher. It's the only dammed thing in the system that functions at all right now.
-Shock_stage takes way too long to climb, *and* you have to be in constant unrelenting agony the whole time or it just fades away
-Instead of nerfing painkillers, I can double the effect that damage has on traumatic_shock (proportionately, I'd also double the shock_stage increment thresholds).

oh and i added a hypo of liquid pain for ease of testing exact numbers

**Alright, here are the SUMMARIZED net changes-**

-All jumps in shock_state from just *existing* have been removed. You won't insta-shockcrit just because your pain got too high, but your shock also won't auto-lower unless you're very close to healed (counting painkiller boost, about 70-100 damage to start). However, all of the 4 meds that can lower your shock_state will try to set it to 150 if it's above 150, allowing quicker escape from paincrit.

-The traumatic_shock checkpoints have been altered to assume that everyone has access to Tramadol, coincidentally mirroring the fact that everyone has had access to Tramadol for months now and are generally lower. The breakpoints are 100 or more traumatic_shock for starting to go up in shock_state, and 35 or less traumatic_shock for starting to go down in shock_state. Some of these breakpoints are below 0, because painkillers gonna kill pain.

-Bonebreak traumatic_shock was reduced to 25 from thirty, but stabilization no longer reduces pain on them. You can still splint it while it's stabilized if you want the pain reduction, but you already got the free 24/7 trama injector.

-Pain defines were vastly lowered on the whole, and full_painkiller was removed. Now, the highest define is PAIN_REDUCTION_VERY_HEAVY at 50. A few chemicals have multiples of VERY_HEAVY, but there's no longer an easy way to just remove *everything* in one go.

-Chemical painkillers (and pain CAUSERS!) were reduced across the board to bring them more on the scale of damage, rather than just overriding any possible amount of damage that you'd be alive with.
Highlights are the bicard/kelo/tricord pill pain got lowered, oxycodone and hydrocodone got lowered from 200 to 50 (we know you'll mix it with Tramadol for an effective 100), Neuraline got lowered from 200 to 100, and QC+ now only causes 50, not 200. Plus ayy neurotoxin and anything define-based (like alcohol) lost a lot of punch.

-Dying no longer resets your shock_stage. You need to actually heal the damage rather than just taking a dirtnap.

-4 chems (oxycodone, hydrocodone, neuraline, and RR) directly reduce your shock stage, and will set your shock_stage to 150 if it was higher than that.

## Why It's Good For The Game

Pain's borked and doesn't work as intended, or really at all. This should hopefully bring it more in-line with the current state of gameplay.

## Changelog
:cl:

balance: Rebalanced pain. It now takes slightly longer to affect you in any way, but it takes more than just popping a pill of Tramadol while missing an arm and having 100 damage to make it go away instantly. Yes, this would have worked before.

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_shock.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_shock.dm
@@ -7,11 +7,11 @@
 		return //Godmode or some other pain reducers. //Analgesic avoids all traumatic shock temporarily
 
 	switch(traumatic_shock)
-		if(200 to INFINITY)
-			adjustShock_Stage(5) //All of these adjust shock_stage based on traumatic_shock
+		if(200 to INFINITY) //All of these adjust shock_stage based on traumatic_shock
+			adjustShock_Stage(5+((traumatic_shock-200)*0.2)) //Uncapped max gain of shock_stage, depending on traumatic_shock.
 
 		if(150 to 200)
-			adjustShock_Stage(2)
+			adjustShock_Stage(1+((traumatic_shock-150)*0.08)) //smooth ramp from 1 to 5
 
 		if(100 to 150)
 			adjustShock_Stage(1)
@@ -33,7 +33,6 @@
 		if(30 to 39)
 			if(prob(20))
 				to_chat(src, span_danger("[pick("It hurts so much", "You really need some painkillers", "Dear god, the pain")]!"))
-			blur_eyes(2)
 			stuttering = max(stuttering, 5)
 		if(40 to 59)
 			if(prob(20))
@@ -47,9 +46,6 @@
 				emote("me", 1, " is having trouble standing.")
 			blur_eyes(2)
 			stuttering = max(stuttering, 5)
-			if(prob(2))
-				if(!lying_angle)
-					emote("me", 1, " is having trouble standing.")
 			adjust_stagger(3, FALSE, 3)
 			add_slowdown(3)
 			if(prob(20))
--- a/code/modules/mob/living/carbon/human/life/handle_shock.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_shock.dm
@@ -2,28 +2,28 @@
 
 /mob/living/carbon/human/handle_shock()
 	. = ..()
-	if(status_flags & GODMODE || analgesic || (species && species.species_flags & NO_PAIN) || stat == DEAD)
+	if(status_flags & GODMODE || analgesic || (species && species.species_flags & NO_PAIN))
 		setShock_Stage(0)
 		return //Godmode or some other pain reducers. //Analgesic avoids all traumatic shock temporarily
 
 	switch(traumatic_shock)
 		if(200 to INFINITY)
-			setShock_Stage(max(shock_stage + 5, 150)) //Indescribable pain. At this point, you will immediately be knocked down, with shock stage set to 150.
+			adjustShock_Stage(5) //All of these adjust shock_stage based on traumatic_shock
 
 		if(150 to 200)
-			adjustShock_Stage(2) //If their shock exceeds 150, add more to their shock stage, regardless of health.
+			adjustShock_Stage(2)
 
 		if(100 to 150)
-			adjustShock_Stage(1) //If their shock exceeds 100, add more to their shock stage, regardless of health.
+			adjustShock_Stage(1)
 
-		if(75 to 100)
-			setShock_Stage(min(shock_stage - 1, 120)) //No greater than 120
+		if(0 to 35)
+			adjustShock_Stage(-1)
 
-		if(50 to 75)
-			setShock_Stage(min(shock_stage - 2, 80))
+		if(-40 to 0)
+			adjustShock_Stage(-2)
 
-		if(-INFINITY to 50)
-			setShock_Stage(min(shock_stage - 5, 60)) //When we have almost no pain remaining. No greater than 60, reduced by 10 each time.
+		if(-INFINITY to -40)
+			adjustShock_Stage(-5)
 
 	//This just adds up effects together at each step, with a few small exceptions. Preferable to copy and paste rather than have a billion if statements.
 	switch(shock_stage)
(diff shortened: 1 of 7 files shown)


jroinc:
Testing shows this works fine, except traumatic_shock can't go negative, despite painkillers that could allow it to do so. Easy fix.

jroinc:
will work on this later, almost finished, just need to *actually* do like 2 things

jroinc:
Okay, new changes as follows.
Oxycodone and hydrocodone drop shock_stage by one every tick, neuraline drops it by an amount equal to it's volume/tick, and RR drops it by 5/tick. All of these also drop your shock_stage to 150 if it was higher. This won't get you out of paincrit if you'd still be in paincrit due to damage and shit, but it'll speed the healing process some if you're crit.

Additionally, the movespeed mod that used to affect ungas at low shock levels now affects them at high shock levels. Straight buff :P

Tests show that it's possible to ignore 200 pain with no ill effects at all with access to common unga chems, so that's fine. But if you let shock_stage build up, it'll take a long time to come down, even with meds, unless you actually heal.

code/modules/mob/living/carbon/human/life/handle_shock.dm
```diff
@@ -2,28 +2,28 @@
...
 	. = ..()
-	if(status_flags & GODMODE || analgesic || (species && species.species_flags & NO_PAIN) || stat == DEAD)
+	if(status_flags & GODMODE || analgesic || (species && species.species_flags & NO_PAIN))
 		setShock_Stage(0)
 		return //Godmode or some other pain reducers. //Analgesic avoids all traumatic shock temporarily
 
 	switch(traumatic_shock)
 		if(200 to INFINITY)
-			setShock_Stage(max(shock_stage + 5, 150)) //Indescribable pain. At this point, you will immediately be knocked down, with shock stage set to 150.
+			adjustShock_Stage(5) //All of these adjust shock_stage based on traumatic_shock
```
BraveMole, line 11: I don't like this change, this means you can get untreatable amount of pain

code/modules/mob/living/carbon/shock.dm
```diff
@@ -4,7 +4,7 @@
 /mob/living/carbon/proc/adjustTraumatic_Shock(amount)
 	if(status_flags & GODMODE)
 		return FALSE	//godmode
-	traumatic_shock = clamp(traumatic_shock+amount,0,maxHealth*2)
+	traumatic_shock = clamp(traumatic_shock+amount,-100,maxHealth*2)
```
BraveMole, line 7: why would traumatic shock be below 0?

BraveMole (changes requested):

code/modules/mob/living/carbon/shock.dm
jroinc: Painkillers. If you have 0 damage and you're on oxy and tramadol, your felt pain should be negative.

code/modules/mob/living/carbon/human/life/handle_shock.dm
jroinc: It's capped to a max of 200 shock_stage. Additionally, oxycodone and three other chems set your shock_stage down to 150 if it's above that.

code/modules/reagents/reagents/medical.dm
```diff
@@ -516,6 +519,7 @@
 	L.heal_limb_damage(10*effect_str, 10*effect_str)
 	L.adjustToxLoss(-2.5*effect_str)
 	L.adjustCloneLoss(effect_str)
+	//L.adjustShock_Stage(-5*effect_str)
```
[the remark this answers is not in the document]
jroinc: did this

jroinc:
Tested and observed in action on ~40 people over a few rounds, works as intended

David-Stormwell:
Should be fine as is, considering it's been in for a few rounds, sure a few minor things here and there but quite forgettable.

PR #8684, opened 2021-10-27
Buff recurring injection

BraveMole:
## About The Pull Request

Make it twice as fast to use recurring injection as a xeno. Also, if you fail to inject any toxin, you don't get a cooldown, (plasma is still used).

## Why It's Good For The Game

Very hard to pull off a recurring injectiosn, this makes it easier but still fair for marines (keep moving and you wont be injected)

## Changelog
:cl:
balance: Make it twice as fast to use recurring injection as a xeno. Also, if you fail to inject any toxin, you don't get a cooldown, (plasma is still used).
/:cl:


David-Stormwell (approved):
Besides in heavy combination with stuns, I almost never see these used any longer, and this should bring them into use slightly moreso.

PR #8686, opened 2021-10-27
Spawners more expensive, silo don't produce ais

BraveMole:
## About The Pull Request

Title. From 600 points to 800

## Why It's Good For The Game

Ais will be more and more dangerous with the pathfinding + escorting improvment. They tend to overrun marines quite easily, so let's reduce the number of xeno ais a bit

## Changelog
:cl:
balance: Silo no longer produces minions
balance: Spawners now cost 800 psy points
/:cl:


BraveMole:
Not needed for now

PR #8702, opened 2021-10-29
(Hopefully) Fixes QC+

Gwuhja:
## About The Pull Request

Should fix QC+ by changing two lines of code. JPR said it has something to do with how wounds interact with IB.

## Why It's Good For The Game

Seriously. Using up to three pens on one IB patient is not fun when Peri+ doesn't have this problem.

## Changelog
:cl:
fix: fixes qc not healing internal bleeding immediately
/:cl:


BraveMole:
> Or do you just want proof that this still does the same thing? That's doable.

yes

QualityVan:
Yeah I don't actually trust this as a fix for the reported problem, but I also still have no idea how the reported problem could be happening. At the very least, this won't make it worse.

BraveMole:
Closing as stale + webeddit + not tried

PR #8714, opened 2021-10-30
M1911/TP-23 Redo/Slight rebalance.

David-Stormwell:
## About The Pull Request

In a nutshell, fixes the recoil that someone broke to have -3 unwielded, to 2, so that you can't akimbo even with RCs, but with an RC you can fire 1 handed fine enough.
M1911 Custom has a new sprite, a wood finish mixed with the TP-23 default, additionally, both guns are childs of the TP-23 taking the stats and magazine now.

## Why It's Good For The Game

The TP-23 has always been super good in akimbo, able to kill xenos instantly, this will change that to be a single hand meta, while upping the magazine side of the non-standard types and having slightly more damage then the marine default. _"Please do note the damage was there before, and has been unchanged."_

## Changelog
:cl:
balance: rebalanced the TP-23, M1911 and the Custom to have recoil while one handed, and additional while akimboed. 
imageadd: Added new M1911 Custom sprite.
imagedel: Removed old M1911 Custom sprite.
/:cl:


BraveMole:
Putting this on hold till hyper do his thing and allow mutliple guns to share same mag type

David-Stormwell:
sure, when that happens as well I'll update the PR and add the new sprites that where gifted.

PR #8719, opened 2021-10-31
Silo collapse happens 15 minutes after shutter drop

BraveMole:
## About The Pull Request

Title. If shutters are drop at 12:25, the 5 minutes silo less timer can happen at 12:40

## Why It's Good For The Game

Incentive for marines to early shutter drop, when the incentive is currently on dropping them as late as possible

## Changelog
:cl:
balance: Silo collapse happens 15 minutes after shutter drop
/:cl:


David-Stormwell (approved):
Should be fine, at least in theory considering how marines play the game.

PR #8736, opened 2021-11-02
removes RNG chance for scatter to not happen

SplinterGP:
## About The Pull Request
Scatter now always happens if you have it, instead of having a RNG chance to not happen at all

## Why It's Good For The Game
Low value scatters now matter more, as it wont have a big chance for it to not scatter at all, which is a issue because big scatter values, which have more chance to happen ALSO goes more out of the straight line, i of course plan to adjust gun scatter after this.

## Changelog
:cl:
balance: gun scatter doens't have a RNG chance to not happen at all anymore
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -1177,9 +1177,6 @@ and you're good to go.
 		if(scatter_tweak)
 			. -= scatter_tweak * 15
 
-	if(!prob(.)) //RNG at work.
-		return 0
-
 
 /obj/item/weapon/gun/proc/simulate_recoil(recoil_bonus = 0, firing_angle)
 	if(CHECK_BITFIELD(flags_item, IS_DEPLOYED) || !gun_user)


BraveMole (changes requested):
Ok but you change scatter values now, rather than just making scatter more prevalant

SplinterGP:
WIP then

jroinc:
I'd love to see a mechanic that caps accuracy loss (from unwielding, attachies, ect) to a certain amount, and any additional accuracy loss is added as scatter. Low accuracy doesn't seem to mean much, but scatter scares ungas.

SplinterGP:
> 
> 
> I'd love to see a mechanic that caps accuracy loss (from unwielding, attachies, ect) to a certain amount, and any additional accuracy loss is added as scatter. Low accuracy doesn't seem to mean much, but scatter scares ungas.

intent is pretty much to remove accuracy except on extreme cases like past 15 tiles

PR #8737, opened 2021-11-03
Uniform and webbing/tie refactor

hyper2snyper:
## About The Pull Request
Refactors Uniforms and Webbings

Uniforms will now use the attachment system for webbings.
Webbings are now attachments by the attachment system and subtypes of armor storage.
Adds support for more than one kind of jumpsuit style.

Adds infrastructure to the attachment system for clothing attachments to be easily added.

This will clear webbings from loadouts. 

## Why It's Good For The Game
Better code. Paving the way for easy addition of clothing attachments.

## Changelog
:cl:
refactor: Uniform and webbing/tie refactor
/:cl:


code/modules/clothing/under/jobs/civilian.dm
```diff
@@ -1,13 +1,12 @@
...
 
 /obj/item/clothing/under/rank
-	rollable_sleeves = TRUE
 
 /obj/item/clothing/under/rank/bartender
 	desc = "It looks like it could use some more flair."
 	name = "bartender's uniform"
 	icon_state = "barman"
-	rollable_sleeves = FALSE
+	adjustment_variants = list()
```
TiviPlus, line 9: should this not be null

code/modules/clothing/under/under.dm
```diff
@@ -9,55 +9,84 @@
...
-		qdel(hastie)
-		hastie = null
-	return ..()
-
-
+	var/list/attachments_allowed = list(
+		/obj/item/armor_module/storage/uniform/webbing,
+		/obj/item/armor_module/storage/uniform/black_vest,
+		/obj/item/armor_module/storage/uniform/brown_vest,
+		/obj/item/armor_module/storage/uniform/white_vest,
```
TiviPlus, line 26: can we make this a string list

code/datums/components/attachment_handler.dm
```diff
@@ -100,6 +101,13 @@
 
 	RegisterSignal(attachment, COMSIG_ATOM_UPDATE_ICON, .proc/update_parent_overlay)
 	update_parent_overlay()
+	if(!CHECK_BITFIELD(attachment_data[FLAGS_ATTACH_FEATURES], ATTACH_APPLY_ON_MOB))
+		return
+	var/obj/parent_obj = parent
+	if(!ismob(parent_obj.loc))
+		return
+	var/mob/wearing_mob = parent_obj.loc
+	wearing_mob.regenerate_icons() ///Theres probably a better way to do this.
```
TiviPlus, line 110: what *are* you doing here

code/datums/loadout/loadout_manager.dm
```diff
@@ -156,11 +163,18 @@
...
+					if(loadout.version < 9)
+						uniform.current_variant = null
+						uniform.attachments = list()
 				var/message_to_send = "Please note: The loadout code has been updated and due to that:"
 				if(loadout.version < 7)
 					message_to_send += "<br>any modular helmet/suit has been removed from it due to the transitioning of loadout version 6 to 7."
 				if(loadout.version < 8)
 					message_to_send += "<br>any PAS11 armor/M10x helmet has been removed from it due to the transitioning of loadout version 7 to 8(Xenonauten Addition)."
+				if(loadout.version < 9)
+					message_to_send += "<br>any uniforms have had their webbings/accessory removed due to the transitioning of loadout version 8 to 9."
```
TiviPlus, line 177: @BraveMole how hard would this be to convert

TiviPlus (changes requested):

code/datums/components/attachment_handler.dm
hyper2snyper: When the item somethings attached to is worn, if i dont do regenerate_icons() you have to take the uniform off to get the new attachments to show. This acts like update_icon() as far as i can tell but i dont like it because it updates all of the mobs worn stuff.

code/datums/loadout/loadout_manager.dm
hyper2snyper: Cant convert because the old item rep is gone. And the item rep saved for the uniform has vars that dont exist anymore.

code/modules/clothing/under/jobs/civilian.dm
TiviPlus: i mean that you check if theres a lenth for the lst already so why not have this null so you dont create an empty list

hyper2snyper: oh i suppose. I was just assuming that when you have a list, to make it null you make it empty

code/datums/loadout/loadout_manager.dm
TiviPlus: ok we will soft reset ties then

XSlayer300:
So like, medals, pins, webbings and such can now be fit into one uniform?

PR #8739, opened 2021-11-03
Reduces Price of the ZX Burst Shotgun

AzonStarfall:
## About The Pull Request

Reduces price of the ZX Burst shotgun from 120 points to 80 points.

## Why It's Good For The Game

The ZX is probably a nice weapon. However due to its cost, it's nearly impossible to see it in action; in any circumstances where one can buy a ZX, a set of B18 is almost always worth the price instead. Dropping the cost of the ZX to minigun levels will allow the weapon to actually be used, as well as give a frame of reference for its effectiveness.

## Changelog
:cl:
balance: Reduced price of the ZX Burst shotgun from 120 to 80.
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -216,7 +216,7 @@ WEAPONS
 /datum/supply_packs/weapons/zx76
 	name = "ZX-76 Twin-Barrled Burst Shotgun"
 	contains = list(/obj/item/weapon/gun/shotgun/zx76)
-	cost = 90
+	cost = 100
 
 /datum/supply_packs/weapons/autosniper
 	name = "IFF Auto Sniper kit"
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -216,7 +216,7 @@ WEAPONS
 /datum/supply_packs/weapons/zx76
 	name = "ZX-76 Twin-Barrled Burst Shotgun"
 	contains = list(/obj/item/weapon/gun/shotgun/zx76)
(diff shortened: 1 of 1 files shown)


Pariah919:
do 90 or 100 imo

Lumipharon:
Just because it's more niche doesn't mean it's overcosted. Cutting the price by a 3rd is a lot for a gun that can easily ggnore benos.

Also keep in mind its ammo is free, where as most req weapons need to keep paying for ammo. So thats an advantage in cost but also practicality. You don't have to hassle req, hope they haven't SSD'd or spent all the points etc.

AzonStarfall:
I can agree with 90 over 80, but 120 prevents this weapon from having a niche at all. In my entire experience running cargo, I've seen this weapon ordered twice. Lowering the cost will give the opportunity, and 90 puts it within the weapon+ammo price range of similar weapons like the TX-8.

David-Stormwell:
Almost every longer round I see, I see a ZX or even a few of them, lowest I'd personally go would be 100.

AndroidMK2:
I used it once and it has shitty recoil making it not that much better from regular shotties, screen going ham after each shot is a bit too much.

Lumipharon:
It has 0 recoil with a vertical grip.

EgorDinamit:
ZX is never used except for rounds where req already has meme amounts of points and they would order crayon packs just because they can.
ZX is also really, really weak. It isn't a weapon to kill with, it's a support weapon at most as it has an atrocious cooldown between shots.

BraveMole (changes requested):
what dabid said

PR #8743, opened 2021-11-05
Remove object from get_hearers_in_view

BraveMole:
Perf improvment maybe. It's 99% not used by our code, or maybe by radios so whatever


TiviPlus:
You need to fix intercoms, holopads, recorders, voice assembly, hear implants, and remove typechecking from lists that use get hearers

BraveMole:
Am lazy

PR #8760, opened 2021-11-09
Nerfs A19 Impact Ammo

KrabSpider:
## About The Pull Request

Removes stagger and screenshake from A19/TX-8 Scout Rifle impact ammo. Slowdown and knockback are the same. 

## Why It's Good For The Game
If the gun in question was slower firing, I'd get _why_ the ammo is the way it is. But it ain't. The A19 is a pre-burst T-64 with built in IFF, it ain't any **business** having applying 2 stagger stacks, a full slowdown stack on the rate of fire it has. 

## Changelog
:cl:
balance: Deletes stagger and screenshake from A19 impact. 
balance: Damage increase to 30.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -565,7 +565,7 @@ datum/ammo/bullet/revolver/tp44
 	sundering = 5
 
 /datum/ammo/bullet/rifle/tx8/impact/on_hit_mob(mob/M, obj/projectile/P)
-	staggerstun(M, P, max_range = 20, slowdown = 2, knockback = 1, shake = 0)
+	staggerstun(M, P, max_range = 20, slowdown = 1, knockback = 1, shake = 0)
 
 /datum/ammo/bullet/rifle/ak47
 	name = "crude heavy rifle bullet"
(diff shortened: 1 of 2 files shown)


ilikesaying:
at least buff the damage a little to compensate for the reduction in debuffs

KrabSpider:
> at least buff the damage a little to compensate for the reduction in debuffs

I'll consider it.

Al-1ce:
TX-8 chews through ammo like nothing else when it comes to purchased guns. I think lowering the firerate and nerfing the stagger (while increasing magazine size or slight damage) would make it more effective.

KrabSpider:
> TX-8 chews through ammo like nothing else when it comes to purchased guns. I think lowering the firerate and nerfing the stagger (while increasing magazine size or slight damage) would make it more effective.

I'll increase the slowdown and give it airburst.  Take it or leave it.

Araulius:
should have removed the burst fire on the t-64 instead. funni salt PR

maikilangiolo:
The TX8 is expensive and burns through ammo. A Minigun is more expensive but it actually kills xenos and ammo costs nothing in comparison. You only see the TX8 rarely because of this. 

Meh PR, but I don't get a say in it so it's up to mantainers.

Passe2142:
The issue was never the rifle but the impact rounds, those rounds dont care if its King, an Ancient Queen or a big caste. The marine mains dont know that screen shake. That stagger denies every chance to use an skill, lag you and sunder ya. Iirc the rifle cost 40 points. Marines have been spamming it so much it shows.  And the strong point of the TX8 is that it can use all kinds of ammo. Only using the funny ones says a lot bout them.

Lumipharon:
Using the Tx8 as a marine with high impact, it doesn't FEEL that powerful.
But playing as a beno getting shot by it though, is god awful, as Passe said. You can't do shit.

BraveMole (changes requested):
leave slowdown alone, rest is fine for a test

PR #4570, opened 2020-07-23
Lv rework

Hathkar:
## About The Pull Request

RedesignsLZ2 almost completely. Moved robotics dome, added atmospherics dome. Changed old atmospherics area's layout and removed the APC.
Redesigned the east side of LZ1.
Moved a few phoron drill locations to account for LZ2 changes.
Added a couple new survivor spawn spots.
Removed a large portion of the northern caves that rarely, if ever, see use.
Moved cave areas to fit with the remaining caves.
Added a bridge north of LZ2 across the river.
Redid some pipes and wiring in the colony.
Changes in jungle to the south of LZ1.
Walls in north LZ.
Map resized, removing large swathes of empty space.

## Why It's Good For The Game

Landing Zone 2 in LV624 is widely regarded as a complete joke. This PR aims to fix that.
A new atmospherics dome fills in some of the empty space in the jungle.
The far north caves contributed nothing meaningful to the map.
Landing zones are a bit less siege-able, to discourage early xeno camping.

## Changelog
:cl:
add: A new atmospherics dome for LV624
add: Completely redone LZ2 LV624
add: New survivor spawn points LV624
add: New bridge north of LZ 2 LV624
del: Removed the far north caves LV624
tweak: Cable and pipe placement LV624
tweak: Jungle south of LZ1 LV624 
balance: LZ1 and LZ2 walls LV624
/:cl:

Previews:
LZ1
https://i.imgur.com/KDFAjgX.jpg
LZ2
https://i.imgur.com/pnuZE30.jpeg
North Caves
https://i.imgur.com/uwodQdi.png

PR #8769, opened 2021-11-10
LV624: 2020 Edition

RipGrayson:
## About The Pull Request

Mostly returns LV624 to it's state before https://github.com/tgstation/TerraGov-Marine-Corps/pull/4570. 
This (mostly) isn't a strict revert though, I've ported our current LZs and added features that didn't exist at the time LV was originally reworked, as such it's more of a fusion between our current LV624 and the old one.

## Why It's Good For The Game

Currently LV624 isn't liked very much, it mostly consists of marines charging into a cave and getting slaughtered time after time. The actual colony part is almost entirely ignored. You could honestly remove 90% of the map and put the landing pads directly south of a generic cave system and the gameplay would be unchanged.
 
I asked around and collected feedback about what could be done to fix LV624 and the most common response outside of "remove caves" was bring back the version of LV624 before it got overhauled. 
This pull does that. 
Looking at the old map, it does have more of an emphasis on the colony compared to our current version. This should result in more varied gameplay for both sides, or at least more rounds where marines and xenos battle over control of the domes instead of endless caves.

Even if it's ultimately not merged, I'm very interested in the feedback regarding this map change and am willing to make further changes based off player reactions. I can still use the feedback for improvement of modern LV624.

## Changelog
:cl:
balance: Fused LV624 with it's pre-rework 2020 counterpart.
/:cl:


RipGrayson:
As a bonus, here's BraveMole's assessment of current LV624.
bmcomplaint

Mrrpip:
Bravemole's diagram seems about right.

DarkWolfKai:
too bad that the rest of the colony's only played during civil war, cuz' loyalist spawns on east

Lumipharon:
Fighting outside as a beno has no advantages and a lot disadvantages.

Lv has a large enough cave network that you can comfortably fallback and keep fighting.
Then it has the river right outside, which is a significant reason to not push out into the open.

That's on top of the normal OB, CAS and mortar firepower to worry about.

If the buildings were bomb proof, that would different.

Walarks:
Fighting outside is 200% more fun than caves. I'd suggest removing 80% of caves and keeping outside. Maybe add some dirt mounds in river so it isn't a total death trap for xeno.

vilereaver41:
changing layout means nothing if you dont give both sides something to fight over or go there for

RipGrayson:
> changing layout means nothing if you dont give both sides something to fight over or go there for

This revision has engineering in the main colony, if that isn't sufficient then I'll add other incentives to make the colony the main focus. Caves are not fun, but they are somewhat necessary for the xenos to have a fighting chance, so I don't want to nerf them unless I have no other balancing option.

BraveMole (changes requested):
Old lv was shit as well, that's why it had so many reworks. 
The river need to be reduced in size, add more bridges, reduce caves size, reweed colony

Sebbe9123:
Can you weed on the grass in the caves? Would be really dumb to not be able to like in the colony.

RipGrayson:
> Can you weed on the grass in the caves? Would be really dumb to not be able to like in the colony.

Weeding on the grass probably deserves a PR of its own. We'll see what the map looks like after all the requested changes have been done.

RipGrayson:
This PR has been superseded by https://github.com/tgstation/TerraGov-Marine-Corps/pull/8806.

PR #8771, opened 2021-11-11
Splints

jroinc:
## About The Pull Request

Splints are now *mostly* time-based. Previously, a splint applied by a medic had 45 health, while a splint applied by a PFC/SL had... zero health.

Now, applying a splint gives 5 minutes of time per level of medical, with a base time of 5 minutes (PFC gets 5, SL gets 10, Medic gets 15, and so on). This is reduced by 5 minutes if self-applied, to a minimum of 5 minutes.

If it's been three or more seconds since you've last moved, the timer's paused. If you're concerned about your splint breaking, don't move around unless you need to. Due to the way the tickrate works, the minimum timer decrement is 4 seconds, even if you just move one step.

Additionally, getting hit on a splint now removes 10 seconds from the splint per hit, so a full HP PFC splint can take 30 and a medic splint can take 90, rather than the previous... 1 and 4-5 respectively. This seems like a lot, but it's constantly decreasing as the timer's ticking down.

When the timer gets within 1 minute of breaking, every time it decrements has a 20% chance to warn you of this.

## Why It's Good For The Game

Splints are less RNG insta-break, but also not something you can just slap on and fight with forever. IMO, it's an overall buff.

## Changelog
:cl:
balance: Splints are now time-based. PFC splints last 5 min, medic splints last 15 min. The timer only counts when moving.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -18,7 +18,7 @@
 	var/max_size = 0
 	var/last_dam = -1
 	var/supported = FALSE
-	///How many instances of damage the limb can take before its splints fall off
+	///How long a splint stays on for, in ticks.
 	var/splint_time = 0
 
 	var/datum/armor/soft_armor
@@ -563,7 +563,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 			remove_limb_flags(LIMB_SPLINTED)
 			to_chat(owner, span_userdanger("The splint on your [display_name] comes apart!"))
 		else
-			if(world.time - last_move_time > 30)
+			if(world.time - owner.last_move_time > 30)
 				return
 			splint_time = max(splint_time - 1, 0) //splint health decrements by one per lifetick, unless it's been more than 10 seconds since your last move.
 			if(splint_time <= 30 && prob(20))
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -19,7 +19,7 @@
 	var/last_dam = -1
 	var/supported = FALSE
 	///How many instances of damage the limb can take before its splints fall off
-	var/splint_health = 0
+	var/splint_time = 0
 
 	var/datum/armor/soft_armor
 	var/datum/armor/hard_armor
@@ -336,11 +336,11 @@
 		owner.custom_pain("You feel something rip in your [display_name]!", 1)
 
 	if(limb_status & LIMB_SPLINTED) //If they have it splinted and no splint health, the splint won't hold.
-		if(splint_health <= 0)
+		if(splint_time <= 0)
 			remove_limb_flags(LIMB_SPLINTED)
 			to_chat(owner, span_userdanger("The splint on your [display_name] comes apart!"))
 		else
-			splint_health = max(splint_health - damage, 0)
+			splint_time = max(splint_time - 5, 0) //if you get hit, it still removes 10 seconds of splint time, thirty hits per level of medic.
 
 	// first check whether we can widen an existing wound
 	var/datum/wound/W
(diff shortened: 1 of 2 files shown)


code/modules/organs/limbs.dm
```diff
@@ -19,7 +19,7 @@
 	var/last_dam = -1
 	var/supported = FALSE
 	///How many instances of damage the limb can take before its splints fall off
```
hyper2snyper, line 21: probably wanna change the autodoc.

code/modules/organs/limbs.dm
```diff
@@ -558,6 +558,17 @@ Note that amputating the affected organ does in fact remove the infection from t
 	if((limb_status & LIMB_ROBOT)) //Robotic limbs don't heal or get worse.
 		return
 
+	if(limb_status & LIMB_SPLINTED) //Splints gradually break over time.
+		if(splint_time <= 0)
+			remove_limb_flags(LIMB_SPLINTED)
+			to_chat(owner, span_userdanger("The splint on your [display_name] comes apart!"))
+		else
+			if(world.time - last_move_time > 30)
```
hyper2snyper, line 566: you didnt test this did you?

hyper2snyper (commented):
Looks pretty good. Is this a buff or nerf?

code/modules/organs/limbs.dm
jroinc: I literally just made this. It still needs testing. I'm like 80% sure it doesn't compile.

code/modules/organs/limbs.dm
jroinc: damm, ok

code/modules/organs/limbs.dm
jroinc: fixed

code/modules/organs/limbs.dm
jroinc: fixed

jroinc:
> 
> 
> Looks pretty good. Is this a buff or nerf?

IMO, it's a buff. Ungas go from being able to take 1 hit (2-3 if a medic applies it) on a splint, to 30-90 if in active combat... but that counter drops slowly over time.

jroinc:
Tested, functions as intended. Needs some large-scale testing for general reactions before merg is considered.

DeltaFire15:
xenos (except warrior punch) needing **90** hits to a **single limb** to break fresh medic splints seems very much overkill. While buffing splint survivability in general a bit seems nice (single throw of some ammo box breaking is it kinda yikes), this imo goes way too far in the other direction, making broken bones effectively meaningless (which marine doesn't carry splints and also does not have someone that carries some nearby)

jroinc:
> 
> 
> xenos (except warrior punch) needing **90** hits to a **single limb** to break fresh medic splints seems very much overkill. While buffing splint survivability in general a bit seems nice (single throw of some ammo box breaking is it kinda yikes), this imo goes way too far in the other direction, making broken bones effectively meaningless (which marine doesn't carry splints and also does not have someone that carries some nearby)

One, Warrior punch still hard-kills a splint in one hit. Two, that counter's always dropping.

I considered just *removing* damage doing anything to splints, but I left a 10-second timer reduction as a compromise.

DeltaFire15:
> One, Warrior punch still hard-kills a splint in one hit. Two, that counter's always dropping.
> 
> I considered just _removing_ damage doing anything to splints, but I left a 10-second timer reduction as a compromise.

Yes, I did state warrior's punch as example for the only thing viable against splints after this change for a reason.
The specific issue is "why is the in-field temporary solution infinitely more resistant to combat than the actual limb itself pre-injurity"
With this, it might actually be preferrable to *never* get bone breaks treated since now you effectively have a hyper-effective wound armor on that limb.

The counter automatically slowly decreasing is effectively almost-irrelevant considering one medic splint alone holds for fifteen minutes on its own.

BraveMole (changes requested):
What delta fire said seems logic, is this not encouraging people to pre splint? 
Aren't you better off with a splint than without?

jroinc:
> What delta fire said seems logic, is this not encouraging people to pre splint? Aren't you better off with a splint than without?

Do they not cause slowdown anymore?

code/modules/organs/limbs.dm
```diff
@@ -558,6 +558,17 @@ Note that amputating the affected organ does in fact remove the infection from t
...
 		return
 
+	if(limb_status & LIMB_SPLINTED) //Splints gradually break over time.
+		if(splint_time <= 0)
+			remove_limb_flags(LIMB_SPLINTED)
+			to_chat(owner, span_userdanger("The splint on your [display_name] comes apart!"))
+		else
+			if(world.time - owner.last_move_time > 30)
+				return
+			splint_time = max(splint_time - 1, 0) //splint health decrements by one per lifetick, unless it's been more than 10 seconds since your last move.
```
Walarks, line 568: Could just reduce like `splint_time -= 1` and then add `splint_time = 0` where you remove the flag.

BraveMole:
closing as stale

PR #8772, opened 2021-11-11
Light neurotox now purges 0.2 meds/tick

jroinc:
## About The Pull Request

Light neurotox now purges 0.2 meds per tick.

## Why It's Good For The Game

As far as I can tell, marines just *ignore* it and don't even bother using purge drugs on it because it has such smol effects. I've tried using it as a sent, and it just feels like it's doing nothing, not even slowing ungas. This makes it *sliiightly* better at it's intended role of suppressing and harrassing, while preventing it from being overpowering.

Please discuss below tho.

## Changelog
:cl:
balance: Light neurotox now purges chems at 1/5 the rate of regular neurotox.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -474,9 +474,8 @@
 
 /datum/reagent/toxin/xeno_neurotoxin/light
 	name = "Light Neurotoxin"
-	description = "A debilitating nerve toxin. Impedes motor control in high doses. Causes progressive loss of mobility over time. This one seems to be weaker enough to not remove other chemicals."
-	purge_list = null
-	purge_rate = 0
+	description = "A debilitating nerve toxin. Impedes motor control in high doses. Causes progressive loss of mobility over time. This one seems to be weaker at removing other chemicals."
+	purge_rate = 0.2
 
 
 /datum/reagent/toxin/xeno_neurotoxin/on_mob_life(mob/living/L, metabolism)


Lumipharon:
As a beno player and a medic, can confirm I completely ignore these normally.

BraveMole (approved):

EvilMelonTheSecond:
Instead of purge I would prefer an effect with slow or something similar, with 0,2 purge I will still ignore it cuz the only way to combat it is rye and not many people carry it with them. Hypervene and water also purge meds so I don't see how this will fix the problem of it getting ignored.

jroinc:
> 
> 
> Instead of purge I would prefer an effect with slow or something similar, with 0,2 purge I will still ignore it cuz the only way to combat it is rye and not many people carry it with them. Hypervene and water also purge meds so I don't see how this will fix the problem of it getting ignored.

Antitoxin. It's a common med that you can take.

EvilMelonTheSecond:
> 
> 
> > Instead of purge I would prefer an effect with slow or something similar, with 0,2 purge I will still ignore it cuz the only way to combat it is rye and not many people carry it with them. Hypervene and water also purge meds so I don't see how this will fix the problem of it getting ignored.
> 
> Antitoxin. It's a common med that you can take.

Dylo is a niche med in my opinion, tox heal is easy with tricord and without the downside of stam drain. I really don't want to deal with ranged chem purge again but if nobody else wants a crowd control type effect(Maybe if light neuro ticks 10 times or something?) this is a small enough purge to keep ignoring I guess.

jroinc:
> 
> 
> > > Instead of purge I would prefer an effect with slow or something similar, with 0,2 purge I will still ignore it cuz the only way to combat it is rye and not many people carry it with them. Hypervene and water also purge meds so I don't see how this will fix the problem of it getting ignored.
> > 
> > 
> > Antitoxin. It's a common med that you can take.
> 
> Dylo is a niche med in my opinion, tox heal is easy with tricord and without the downside of stam drain. I really don't want to deal with ranged chem purge again but if nobody else wants a crowd control type effect(Maybe if light neuro ticks 10 times or something?) this is a small enough purge to keep ignoring I guess.

Ah, and there's where it comes in.
*In your opinion, dylo is a niche med* because you don't take it, because you can just ignore light neuro because it does no meaningful damage.

The goal of this is to add a minimum drain. On one hand, I agree, I don't want to see someone get spit and then all their meds vanish. On the other, I don't want it to continue to feel like neurospitting someone as a sentinel is just useless.

David-Stormwell:
Never saw Neuro spit as useless, but that's just me as a solo player who constantly presses alt/runs.
What the problem you're having is, from what it seems, spitting at the unga horde/medium-ish walking marines moving tightly, not running ahead.
I fucking hate how Dylo works, and attempt to avoid it as it completely shuts my stamina regen down while in me, but even so because of the many toxins that exist, force me to use it instead of tricord.
That all being said, I'd consider this in gameplay, as it shouldn't affect normal marines too much anyways.

BraveMole:
I prefer this https://github.com/tgstation/TerraGov-Marine-Corps/pull/8801

PR #8816, opened 2021-11-20
Fixes Ravager vampirism

DeltaFire15:
## About The Pull Request
Title. It's supposed to give them 10% additional healing from any sources per currently leeched target, but didn't work due to two seperate small issues (one in the heal proc itself, one in vampirism code) (Notice, this "any sources" still only is "those that use heal_wounds()", others get skipped but that has been the case initially too)
This fixes those issues and therefore makes it work properly*.

*The Ravager's heal abilities during enrage do not use heal_wounds() so this doesn't apply to them still. Adding the used comsig to them on the other hand would make them also trigger other things that hook into it e.g. hivelord's healing infusion. Unsure which of the two states is preferred so I didn't touch this logic as of now.

## Why It's Good For The Game
Things working instead of breaking seems kinda good.

## Changelog
:cl:
fix: Fixed Ravager primordial vampirism not working.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
@@ -479,11 +479,10 @@
 	SIGNAL_HANDLER
 	if(!leech_count)
 		return
-	var/count = leech_count
 	//heals 10% extra per leeched
-	count /= 10
-	count += 1
-	heal_data += count
+	var/heal_mod = 1 + (leech_count/10)
+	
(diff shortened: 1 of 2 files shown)


TiviPlus (approved):

DeltaFire15:
Mmm, thinking about it, I could add another comsig specifically used by the Ravager ability heal and make that also connect to the vamp handler. That way those abilities can be integrated without side effects, as them not benefitting from it at all seems kinda weird.
If you want that done say so, otherwise it's good to merge.

TiviPlus:
eh good enough

PR #8820, opened 2021-11-21
TGS Minerva

AzonStarfall:
## About The Pull Request

Adds a new low(ish) pop shipmap, the TGS Minerva. Originally designed as a survey vessel, decommissioned and refitted as a science ship, then recalled and once again retrofitted as a marine deployment cruiser, the TGS Minerva has seen a long life; The xenomorphs may be the end of it.

Minerva

The research sector is given additional equipment in service of the Minerva's former science designation, coming equipped with Surplus Armor and Munition vendors for the deploying research staff.

Requisitions uses a similar shutter access system to the TGS Sulaco, in response to recent notices from High Command regarding the unnecessary demolition of shipside airlocks.

As part of the retrofit, a smaller-than-average command bunks were constructed on-site. While still functionally usable, testing officers have complained about their cramped quarters. Nanotrasen was happy to provide a substantial subsidy for the construction of their Liaison office, which nearly mirrors the entire command bunks section in size.

Additionally, certain access requirements were overlooked during the retrofit. Of note, it appears that Chemistry has become a public access area. Testing officers have noted some discomfort about the availability of chemical synthesis to the general marine population, but note that the standard protections of a single reinforced window rarely stopped anyone anyways.

To cut costs, the onboard chemical stores from the research designation were shifted to the rear of the vessel for use as a self-destruct system. Upon activation, the reaction process will begin, and crew are required to lift control rods in order to manually direct the chemical flow. Upon procedural completion, the final reaction stage can be initiated from the self-destruct console. 
The necessary chemicals are stored on the starboard and port sides of the self-destruct room, which has been cited as a potential breach risk if hostiles were to gain control of the relevant chambers.
This is functionally identical to a standard self destruct.

## Why It's Good For The Game

Adding new shipmap options increases gameplay possibilities. Shipmap is balanced around Pillar of Spring resource levels, in terms of CAS and medical equipment.

## Changelog
:cl:
add: Adds the TGS Minerva, a new shipmap.
config: Adds Minerva to the shipmaps config.
/:cl:


Walarks:
I'd be interested in testing this in terms of lowering prep time since most of the resources are very close to each-other and the Alamo.

Also to me it seems to be a spacious enough map for high-pop as well.

AzonStarfall:
> Also to me it seems to be a spacious enough map for high-pop as well.

I'm a bit worried about using it for high pop, since the prep room is limited and kind of cramped. It'd be a large early-round bottleneck.

Licherr:
> > Also to me it seems to be a spacious enough map for high-pop as well.
> 
> I'm a bit worried about using it for high pop, since the prep room is limited and kind of cramped. It'd be a large early-round bottleneck.

their fault for voting it at that point

AzonStarfall:
Updated map image regarding new commits:

Minerva

BraveMole (changes requested):
Supply drop consol should be unique

image

No maint access here, why?

Rest looks solid

AzonStarfall:
> No maint access here, why?

Intention was to have those areas function as SD stalemate blockers, so that Xenos could acid in for alternative attack points. Since you can only access them by taking either CIC or Engineering, I don't feel like it's completely unbalanced in the Xenos favor.

Correcting the Supply Console.

RipGrayson:
I don't see any problems with it, looks fine to me

Luxyboo:
Does the AI have protective glass?

AzonStarfall:
> Does the AI have protective glass?

It doesn't have the internal 1-directional safety layer, but it's kept in a shuttered room, and it's own kevlar-windowed room inside of that.

code/__DEFINES/__game.dm
```diff
@@ -95,6 +96,12 @@
 /// Is something in the IC chat filter? This is config dependent.
 #define CHAT_FILTER_CHECK(text) (config.ic_filter_regex && findtext(text, config.ic_filter_regex))
 
+//for whether AI eyes see static, and whether it is mouse-opaque or not
```
AzonStarfall, line 99: I still don't know why this is counting as an alteration, or if I need to fix it.

Lumipharon:
Posting the feedback we discussed the other day:
The public chem is too much.

On other maps, not many people can or will powergame chem because only one person can use it at a time, and if people can't do it before deployment they won't usually bother.

Having 4 public chem stations means you can simply have four times as many people use it in the same amount of time.

AzonStarfall:
> Having 4 public chem stations means you can simply have four times as many people use it in the same amount of time.

I feel like this is a more general issue with TGMC chemistry, though. Since everyone can theoretically do it, people are still going to despite it being Medbay's "thing". Locking it off with two dispensers clearly doesn't work, as seen with Theseus and Pillar especially, which both moderately resolve the availability problem by adding another chemistry area. This just elaborates on those fixes, and reduces hassle by having the area publicly accessible rather than requiring a window deconstruction.

That said, having 3000 charge on each is definitely a problem, putting production capacity well over what any other ship is capable of. I'm going to try and reduce the charge and recharge rate of each by half (3000 to 1500), so that the total production amount is the same.

BraveMole (changes requested):
The public chem is iffy and should not be a specific map feature.

If you want to make chem a public space, you post a separate pr for all maps, so it can be properly discussed

AzonStarfall:
Gotcha. In the mean time, to avoid break-ins, would changing it to a shutter-access system kind of like Requisitions be acceptable? Only medbay can open it to the public, so it's still their choice.

AzonStarfall:
Finalized version of the Minerva:

Minerva

ThelastArchiver:
here some feedback from playing as Ai the map is missing cameras at several locations resulting in blindspots for the Ai like in atmos the cls office , the western landing pad and the kitchen no cams in maints is ok but maybe ad a few hologram projectors for the ai for some extra ai fun.

BraveMole (approved):
Another round of test and then merge

PR #8827, opened 2021-11-22
Tl-127 Flak Ammo

KrabSpider:
## About The Pull Request
Adds Tl-127 flak rifle rounds. Small targets that are tagged with these are dealt a minor slowdown and stagger (1 stack of each), and are knocked back targets by. It's req only for 5 points a mag. 

## Why It's Good For The Game
It's a sidegrade ammunition that allows dedicated snipers to be a much better asset to their team by being able to _force_ distance between an oncoming target and a (potentially pinned) ally. 

image

## Changelog
:cl:
add: Tl-127 flak rifle rounds. Available through requisitions for a mere 5 points a pop. Causes deals a mighty blow to targets as they're knocked back 4 tiles. 1 stagger, 1 slow, 4 knockback. 
imageadd: Tl-127 flak mags
/:cl:


InterroLouis:
Pffffffthahahahaha nobody will ever use this.

Lumipharon:
Why is the knockback so huge? And why would flak cause that?

KrabSpider:
> And why would flak cause that?

Because I couldn't think of a better name.

>Why is the knockback so huge?

:)

lKiseki:
i was told to comment on this pr
do i get candy yet

code/modules/projectiles/ammo_datums.dm
```diff
@@ -989,6 +989,20 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range_min = 2
 	damage_falloff = 0.25
 
+/datum/ammo/bullet/sniper/pfc/flak
+	name = "high caliber flak rifle bullet"
+	hud_state = "sniper_supersonic"
+	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING|AMMO_SNIPER
+	damage = 70
```
hyper2snyper, line 996: Make it half damage, and half penetration from the normal rounds. Staggerstun is pretty powerful. Halve em and it looks good to me

hyper2snyper (changes requested):

code/modules/projectiles/ammo_datums.dm
KrabSpider: Done!

PR #8828, opened 2021-11-22
Fix some runtime with on advanced pathfinding

BraveMole:
Title


TiviPlus (changes requested):
qdeleted checks for null already

TiviPlus (approved):

PR #8833, opened 2021-11-24
Rebalances the C4 Plastic Explosive

lKiseki:
## About The Pull Request
Per title. The C4 Plastic _(p l a s t i q u e)_ Explosive has been rebalanced in the following manner, as per the Project Lead's requests:
- Engineer skill requirements removed. Whether you have the skill or not no longer delays any procedures regarding C4s.
- All C4 procedures have had their usual delay reduced from 5 seconds to 2 seconds.
- The C4 is now _mostly_ silent.
- The C4 will create smoke around it when detonating.
- The C4 now has a backblast around it. Don't stand near it unless you just want to get thrown around.
- C4 inventory size reduced.

The C4 has also had its costs altered in various areas:
- The C4 is now worth 2 points instead of 3 in SL and engineer vendors.
- The reqs vendor now has 5 C4 instead of 2 by default.
- Reqs cost for C4 reduced from 5 to 3 points.
- SL and engineer vendors will give 2 C4 instead of 1 by default.

## Why It's Good For The Game
Makes the C4 actually worth a damn. The Project Lead also requested most of these changes, and honestly the C4 has been incredibly lackluster for a long time. I have hopes that this PR will amend that.

## Changelog
:cl: Lewdcifer
balance: C4 rebalance: no skill requirement, deploys faster, mostly silent, creates smoke when detonating, now has a backblast around it, reduced inventory size. C4 made cheaper in reqs, vendors now have more C4 by default.
/:cl:


lKiseki:
haha webedit go brrr

SilencedMP5A5:
good change

ThelastArchiver:
ah yes finally c4 wont be just deleting walls out of existence but any marines staring at it too

code/game/objects/items/explosives/plastique.dm
```diff
@@ -11,18 +11,14 @@
 	var/timer = 10
 	var/detonation_pending
 	var/atom/plant_target = null //which atom the plastique explosive is planted on
+	var/datum/effect_system/smoke_spread/smoketype = /datum/effect_system/smoke_spread/bad // smoke type created when the grenade is primed
```
hyper2snyper, line 14: you should probably auto-doc all these.
```suggestion
	/// smoke type created when the grenade is primed
	var/datum/effect_system/smoke_spread/smoketype = /datum/effect_system/smoke_spread/bad
```

PR #8849, opened 2021-11-27
Sharpening an Entrenching Tool reduces the amount of time it takes to dig

Three333Digit:
## About The Pull Request
Title.
Shovel speed from 20 to 10 if the etool is sharpened. Can be 15 as well.

## Why It's Good For The Game
The only time I've seen someone use the sharpened etool is when a squad marine was using it to oneshot weeds.

## Changelog
:cl:
balance: Buffs shovel speed of sharpened etool from 20 to 10
/:cl:

### DIFF
--- a/code/game/objects/items/tools/shovel_tools.dm
+++ b/code/game/objects/items/tools/shovel_tools.dm
@@ -67,9 +67,6 @@
 		return
 
 	if(isturf(target))
-		if (sharp)
-			shovelspeed = 10
-			
 		if(!dirt_amt)
 			var/turf/T = target
 			var/turfdirt = T.get_dirt_type()
@@ -182,6 +179,7 @@
 		return
 	sharp = IS_SHARP_ITEM_SIMPLE
 	name = "sharpened " + name
+	shovelspeed = 10
 	force = 60
 	update_icon()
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/tools/shovel_tools.dm
```diff
@@ -67,6 +67,9 @@
 		return
 
 	if(isturf(target))
+		if (sharp)
```
hyper2snyper, line 70: Put this where the shovel gets sharpened. Since if its here it will set the var to 10 everytime the shovel starts to dig.

hyper2snyper (changes requested):

code/game/objects/items/tools/shovel_tools.dm
Three333Digit: Done. Is this fine?

hyper2snyper: Thats perfect!

hyper2snyper (approved):
Looks good to me

BraveMole (changes requested):
Then why would you not sharpen it? You can't put it back in a pouch maybe?

Three333Digit:
> Then why would you not sharpen it? You can't put it back in a pouch maybe?

Loses most of its portability, yeah. To make the unsharpened state a considerable option, I can reduce its weight class when folded from normal to small.

BraveMole (approved):

PR #8855, opened 2021-11-27
Removes roundstart Plasma Cutters

lKiseki:
## About The Pull Request
Per title, and a different approach towards nerfing the Plasma Cutter. Leaves it as-is, but makes it so that you have to purchase it from reqs.

## Why It's Good For The Game
The Plasma Cutter has always been a tool that has been too good at its job. It's more easily justifiable if it's locked behind reqs rather than being an item you can just get at the start of a round.

## Changelog
:cl: Lewdcifer
del: Plasma Cutters no longer spawn roundstart.
/:cl:

### DIFF
--- a/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
+++ b/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
@@ -14795,11 +14795,9 @@
 /area/mainship/command/self_destruct)
 "TO" = (
 /obj/structure/rack,
-/obj/item/tool/pickaxe/plasmacutter,
 /obj/machinery/light{
 	dir = 8
 	},
-/obj/item/tool/pickaxe/plasmacutter,
 /obj/item/storage/belt/utility/full,
 /obj/item/storage/belt/utility/full,
 /obj/item/tool/weldpack,
--- a/_maps/map_files/Sulaco/Sulaco.dmm
+++ b/_maps/map_files/Sulaco/Sulaco.dmm
@@ -1845,7 +1845,6 @@
 /area/sulaco/medbay/cmo)
 "ahS" = (
 /obj/structure/table/mainship,
-/obj/item/tool/pickaxe/plasmacutter,
 /obj/item/cell/high,
 /obj/item/cell/high,
 /obj/item/clothing/glasses/welding,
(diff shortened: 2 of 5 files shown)


vilereaver41:
true true

Losenis:
I feel targeted

BraveMole (changes requested):
PC is too good at what it does, ok, but marines don't have a good alternatives.

Before doing that,i think you should give marines another way of cleaning mazes, something less powerfull than current PC. Maybe make C4 kill walls at a range?

Losenis:
Also this may indirectly be a synth/tadpole nerf, since synths' only other option for wallbreaking is meleeing it to death and the roundstart PC allows the front of the tadpole to be PC'd off without having to pray some engi spawns and goes PC to ask for it borrowed. Realistically, nobody's going to buy a PC roundstart either.

Walarks:
PC is very good at clearing mazes and xenos are very good at making them quickly. I'm not sure about replacing PC with a more limited tool, but at the same time I think engi can get it from their prep vendor.

vilereaver41:
> PC is too good at what it does, ok, but marines don't have a good alternatives.
> 
> Before doing that,i think you should give marines another way of cleaning mazes, something less powerfull than current PC. Maybe make C4 kill walls at a range?

flamers and melee?

David-Stormwell:
> PC is too good at what it does, ok, but marines don't have a good alternatives.
> 
> Before doing that,i think you should give marines another way of cleaning mazes, something less powerfull than current PC. Maybe make C4 kill walls at a range?

C4 has already been buffed.
If you want a large explosion then Demo charges exist.
In fact, might as well buff them accordingly as well, decrease the timer, increase the damage to be able to blow up walls, and also decrease range to comp.
Then in the end make them a little more accessible to marines as well.

lKiseki:
> C4 has already been buffed. If you want a large explosion then Demo charges exist. In fact, might as well buff them accordingly as well, decrease the timer, increase the damage to be able to blow up walls, and also decrease range to comp. Then in the end make them a little more accessible to marines as well.

Can do this on another PR. Sir, yes sir, oorah.

lKiseki:
mole said he preferred nerfing the PC, david wants the PC nerfed either way and the detpack buffed, i don't mind either way

PR #8877, opened 2021-12-02
add some vsc recommended extensions

BraveMole:
Git lens is needed for new contrib guide and colorised bracket is cool


.vscode/extensions.json
```diff
@@ -3,10 +3,12 @@
 		"gbasood.byond-dm-language-support",
 		"arcanis.vscode-zipfs",
```
BraveMole, line 4: why did you add that tivi?

Issue #748, opened 2019-02-18
Powerpacks are deleting powercells


Knightfall5:
## Reproduction:

 Not entirely sure on this one. The original powerpack I had worked just fine, and when I was running low on ammo I went back to the FOB, where I'd left the other. I assumed someone stole the powercell from it since it was gone, so I went and inserted another, reloaded, and went on my way. The powercell inside was gone when I needed to reload again a few minutes later, so I found another powercell and tossed it in. This one was gone shortly after as well, and after inserting a third powercell things completely bugged out.

The cell was 'inside' the powerpack, but remained in my hand, and I could toss it on the ground and drag it around, but not pick it up. Removing the cell from the powerpack teleported it to my hand. After a minute of doing this, the cell vanished from my hand, leaving my sprite holding a cell that no longer existed, and leaving both powerpacks unusable, as any attempt to interact with them gave a "There is no cell in the the M56 powerpack." message, leaving one stuck on my back.

Update: I did some experimenting in a new round, simply removing the cell and trying to stick it back in causes the bug to happen after a couple swaps. The cell either vanished while inside the powerpack, or becomes bugged and can't be interacted with on the ground. In addition, this bugs out every powerpack you are holding, the one on my back had it's cell vanish too and I never touched it after putting it on.

DominikPanic:
Thanks Ghommie for fixing my fuckup :+1:

PR #8904, opened 2021-12-04
Makes powerloader faster during prep

Walarks:
## About The Pull Request
Makes loader very fast during prep to make prep take less time. When the shutters drop, its speed returns to normal. This might seem immersion-breaking on one hand, but I think that this could save around 10m of prep and getting to the actual round faster makes up for it.

I'm not sure how harmed xenos would be by marines finishing prep faster, but if it is noticeable I think this can be offset by increasing pre-shutters evolution/aging gains.

## Why It's Good For The Game
It makes the round start and get to the core of the gameplay faster. Currently prep takes a large amount of time and a lot of it is spent performing the same routine every round.

## Changelog
:cl:
qol: makes loader much faster during prep
/:cl:

### DIFF
--- a/code/datums/components/riding/riding_vehicle.dm
+++ b/code/datums/components/riding/riding_vehicle.dm
@@ -96,6 +96,19 @@
 	ride_check_flags = RIDER_NEEDS_LEGS | RIDER_NEEDS_ARMS
 	vehicle_move_delay = 4
 
+/datum/component/riding/vehicle/powerloader/Initialize(mob/living/riding_mob, force, ride_check_flags, potion_boost)
+	. = ..()
+	if(SSmonitor.gamestate != SHUTTERS_CLOSED)
+		return
+	vehicle_move_delay = 0.5
+	RegisterSignal(src, list(COMSIG_GLOB_OPEN_TIMED_SHUTTERS_LATE, COMSIG_GLOB_OPEN_TIMED_SHUTTERS_XENO_HIVEMIND, COMSIG_GLOB_OPEN_SHUTTERS_EARLY, COMSIG_GLOB_TADPOLE_LAUNCHED), .proc/regain_move_delay)
+
+///Returns the loader's movement speed to normal after prep is over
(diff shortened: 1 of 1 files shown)


BraveMole (changes requested):
You said it yourself, completely breaks immersion for a a benefit that i suspect is very small, if not inexistant.

If you want to buff the powerloader, (let's say a speed of 3, 2) throughout the whole game, why not, but this is not credible in the slightest

Walarks:
> You said it yourself, completely breaks immersion for a a benefit that i suspect is very small, if not inexistant.
> 
> If you want to buff the powerloader, (let's say a speed of 3, 2) throughout the whole game, why not, but this is not credible in the slightest

I think that most people are here for the roleplay and things that come after prep for prep immersion to matter that much. Considering the amount of stuff that is moved around, I think that this may decrease the time it takes to prep by a noticeable amount.

Losenis:
I don't think people care for immersion pre-shutters. Either they're focused in their prep to care about ripleys, or they're alt tabbed having already prepped. A faster ripley is less of an immersion breaker than a moshpit at req or chem break-in #748.

Misha600:
> > You said it yourself, completely breaks immersion for a a benefit that i suspect is very small, if not inexistant.
> > If you want to buff the powerloader, (let's say a speed of 3, 2) throughout the whole game, why not, but this is not credible in the slightest
> 
> I think that most people are here for the roleplay and things that come after prep for prep immersion to matter that much. Considering the amount of stuff that is moved around, I think that this may decrease the time it takes to prep by a noticeable amount.

no i'm here to shoot bug and FF marines

TiviPlus (changes requested):
I think that this could save around 10m of prep and getting to the actual round faster makes up for it.

Ok but people aren't waiting fir power loader they're usually waiting for fob drone so this doesn't fix anything

Losenis:
Fucking

Bruh

PR #8917, opened 2021-12-07
TX-11 Buffs

KrabSpider:
## About The Pull Request
TX-11 extra delay reduced to 0.25 from 0.5. 
TX-11 scope zoom offset increased to 7 from 5.

## Why It's Good For The Game
Bad gun is now stronger. 
In other news. The zoom offset is to atleast provide a reason to keep it on the gun rather than ditch it. May or may not be necessary with burst changes...
The burst delay on the other hand is...necessary. Currently its **genuinely** awkward and 60 true damage or not, a gun like this is not strong enough to justify a 0.5 second delay. 

## Changelog
:cl:
balance: TX-11 extra delay decreased to 0.25 from 0.5.
balance: TX-11 scope zoom increased to 7 from 5. Stop littering. 
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -60,7 +60,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	return
 
 ///Special effects for leaving a turf. Only called if the projectile has AMMO_LEAVE_TURF enabled
-/datum/ammo/proc/on_leave_turf(turf/T, atom/firer)	
+/datum/ammo/proc/on_leave_turf(turf/T, atom/firer)
 	return
 
 /datum/ammo/proc/knockback(mob/victim, obj/projectile/proj, max_range = 2)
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/ammo_datums.dm
```diff
@@ -519,6 +519,7 @@ datum/ammo/bullet/revolver/tp44
 	damage = 20
 	penetration = 20
 	sundering = 1.25
+	shell_speed = 3
```
BraveMole, line 522: undocumented

BraveMole (changes requested):

code/modules/projectiles/ammo_datums.dm
KrabSpider: I...actually forgot I even made that change. Anyway, it's been addressed.

BraveMole: This seems too much of a buff no? The extra delay itself is really good

code/modules/projectiles/ammo_datums.dm
```diff
@@ -519,7 +519,6 @@ datum/ammo/bullet/revolver/tp44
 	damage = 20
 	penetration = 20
 	sundering = 1.25
-	shell_speed = 3
```
KrabSpider: Done

BraveMole (approved):

PR #8920, opened 2021-12-07
Give T-25 Smartrifle PB and burst fire

novaepee:
## About The Pull Request

Read title

## Why It's Good For The Game

Rifles have burst fire and PB, therefore it makes sense to give T-25 the same. I swear, this is not a powercreep and not just double downing that T-25 is T-12+. It can be balance since T-25 hits hard already, but T-25 is already an unga weapon.

T-25 has 80 total rounds, and burst fire shoots 3 bullets, so one mag will have 26 burst fire with the remainer of 2 bullets UNTIL you include the round in the chamber, which in the SECOND mag ASSUMING you save the one round in the chamber is 27 burst fire.

## Changelog
:cl:
expansion: Gives T-25 burst fire and PB
balance: T-25 has burst capacity, therefore will punch harder if aimed well.
/:cl:


Lumipharon:
Pretty random that it couldn't PB, but burst fire seems like a pretty decent DPS buff on an already good weapon. Not sure how balanced that will be.

BraveMole:
T25 is a capable gun and needs no buff

PR #8929, opened 2021-12-08
Gives Crusher the Regenerate Skin ability

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
Hoo boy, here we go. I'm going to offer a good enough explanation for the general public.
The intention here is not to powercreep, moreso resolve one of the major glaring issues in Crusher gameplay. You could call it powercreep anyways, though, I guess.

Crusher is a T3 caste that is intended to be a tank. But, quite honestly, it has issues soaking up damage in crowds. So why does this happen, then? The issue is tied to the Sunder mechanic. It has like 80 armor at best, so it's not a matter of armor. It has 370 health at best, which is pretty good, so it's also not a matter of health. It's simply Sunder -- a slow-moving xenomorph is unable to respond properly to damage coming its way, which means it gets sundered very easily, which in turn translates to big damage.
You can tell that increasing armor or health aren't really proper solutions here.

I feel that this might just be the solution we're looking for. This gives Crusher an appropriate response against Sunder, which is an incredibly punishing mechanic towards this caste specifically, to the point that it cripples it big time in terms of gameplay.

If need be, I am willing to compromise by providing an exchange in other aspects, like reducing Crusher's armor or health, or maybe giving Regenerate Skin a wind-up specifically for Crusher, which would make it less of a combat option and moreso something it'd use in its downtime; but I feel that Crusher has needed this sort of help for the longest while, a proper response towards Sunder besides sitting there and crying for several minutes. It'd be more enjoyable in terms of gameplay this way, I feel.

Thank you for coming to my TED Talk.

## Changelog
:cl: Lewdcifer
balance: Crusher now has Regenerate Skin.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
@@ -45,6 +45,7 @@
 		/datum/action/xeno_action/activable/stomp,
 		/datum/action/xeno_action/ready_charge,
 		/datum/action/xeno_action/activable/cresttoss,
+		/datum/action/xeno_action/activable/regenerate_skin,
 	)
 
 /datum/xeno_caste/crusher/on_caste_applied(mob/xenomorph)


BraveMole:
wow wow wow powercreeepppppe

EvilMelonTheSecond:
Can crusher even afford the plasma cost after charging?

lKiseki:
> Can crusher even afford the plasma cost after charging?

Maybe not if it's too young, since Young has 200 plasma total and Mature has like 250. Regenerate Skin _does_ have a decent plasma cost (160), but for Elder and Ancient it's relatively affordable.

Pariah919:
this is hilariously massive powercreep that ignores the entire point of sunder supposed to slowly whittle things down, though that point has been lost long ago I suppose.

i could undertand if you nerf the passive sunder regen but there's a reason defender is the only caste that got this, mostly because it is far slower in any form of combat stance, while crusher doesn't really have this

lKiseki:
> i could undertand if you nerf the passive sunder regen for this but there's a reason defender is the only caste that got this, mostly because it is far slower in any form of combat stance, while crusher doesn't really have this problem as it can just croosh and continue to croosh.

Changing passive sunder regen for just one caste sounds pretty silly. Besides, your Defender point is moot because Defender isn't as slow, and isn't charging into death balls. Crusher meets both of these criteria.

> If you're having BIG sunder problems go ask a drone or hivelord for help, and if there is none, I suppose you could ask the queen for pheromones.

That's how it is currently and it's not an enjoyable experience, hence the PR itself.

> EDIT: Also make it a seperate ability, these two shouldn't share the same C/D or Plasma cost codewise and balance wise, that's just absurd.

Feel free to tell me why, but I don't see an issue in Regenerate Skin as it is currently for Crusher.

Araulius:
Wait... So your issue is the marine counter to high armor high health tanks... and your solution to this, is to give the ability to nullify it? :hampter:

Kuro020:
I do think crusher needs an ability to counter sunder a bit, but not the defender's one since that one immediately heals all sunder and it doenst even have that long of a cooldown

lKiseki:
There may be other solutions, then. I'll explore some more.

PR #8949, opened 2021-12-11
Engineering storage module tweak

Three333Digit:
## About The Pull Request
Increases the max storage space of engineering storage module from 15 to 30, same as the medical storage module. 

## Why It's Good For The Game
With the expanded list and added slot, there are situations where you aren't able to make full use of all 5 slots because of the max storage space limitation making for awkward resource management. This fixes that so you are able to make use of all 5 slots no matter what.

## Changelog
:cl:
balance: tweaks engineering storage module 
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/storage.dm
+++ b/code/modules/clothing/modular_armor/attachments/storage.dm
@@ -138,7 +138,7 @@
 	storage =  /obj/item/storage/internal/modular/engineering
 
 /obj/item/storage/internal/modular/engineering
-	max_storage_space = 15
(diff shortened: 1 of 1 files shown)


chesse20:
seems like this could encourage robots to roundstart take 3 or 4 full stacks of wire from the enginnering vendors that have 10 of them at round-start to put into their engineering storage module. Don't know if thats a problem that needs to be solved right now though

BraveMole:
This is just storage power creep with little to no reasonning. Storage managment is part of the game, deal with it

PR #8952, opened 2021-12-11
You can no longer plasmacutter silos

ivanmixo:
## About The Pull Request

Adds a `PLASMACUTTER_IMMUNE` flag. It does what it says. Currently it's only applied to silos.
Adds defines to `bitfields.dm` which someone forgot. Also adds comments to the defines to explain them.

## Why It's Good For The Game

Being able to destroy a silo in a matter of seconds may not be ideal. Also bug fix good.

## Changelog
:cl:
balance: The resin silo can no longer be quickly cut apart by a resin silo
fix: The silo being destroyed should trigger the silo collapse timer every time now
/:cl:


BraveMole (changes requested):
There should not be side effects from Destroy. 

A correct fix would be to have the hivedatum listen to the destroy signal

BraveMole (changes requested):
Just make the silo PC immune and call it a day

ivanmixo:
This is stupid but okay.

PR #8993, opened 2021-12-16
Martini rebalance

Pariah919:
## About The Pull Request
Rebalances martini
Base damage reduced to 90
Damage falloff is reversed for the Martini, and now it gains 5 damage per tile instead of losing.
stagger stun tile range increased to 6.
--
aim mode added
--

## Why It's Good For The Game
The Martini is a useless meme gun that literally nobody uses and doesn't even have a funny gimmick considering I could just use slug ts-34 for the same mechanic.
This gives it some uniqueness in how it works, functions and plays and makes it not terrible in CQC.
The max damage btw is 195 at 21 tiles, if you somehow get that far, however this overall deals less damage overall.

## Changelog
:cl:
balance: Martini has gotten a rework to make it feel more unique, it's damage has been reduced to 90, it now has reversed falloff (it gains 5 damage per tile it flies through, maxing out at around 195.). It has gained aim mode (but you can barely move while in aim mode) and it's status effects on hitting a mob has been increased in range by one tile.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -980,12 +980,13 @@ datum/ammo/bullet/revolver/tp44
 	hud_state = "sniper_crude"
 	handful_amount = 5
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
-	damage = 120
+	damage = 90
 	penetration = 20
 	sundering = 10
+	damage_falloff = -5
 
 /datum/ammo/bullet/sniper/martini/on_hit_mob(mob/M, obj/projectile/P)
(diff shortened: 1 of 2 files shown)


hyper2snyper:
Can we leave the damage just to see?

BraveMole (approved):

SevenHaven:
coder moment

David-Stormwell:
Have mercy on boilers when?
The anti mat and other such weapons exist, why should this have it and have huge amounts of damage when other spec weapons exist which are now outclassed by a proficient user?

BraveMole (changes requested):
negative falloff is neat but this is way overdone.

Tune this effect down

PR #9008, opened 2021-12-19
Fixes patching burns with wire and burn overlays not updating

ivanmixo:
## About The Pull Request

- Fixes the message so it doesn't look wrong
- Makes the burn overlays update after you repair em

## Why It's Good For The Game

Bug bad

## Changelog
:cl:
fix: Burn overlays update when you patch yourself with cable
fix: Fixes patching burnt wires with cable giving the wrong message
/:cl:

### DIFF
--- a/code/modules/power/cable.dm
+++ b/code/modules/power/cable.dm
@@ -508,8 +508,8 @@ GLOBAL_LIST(cable_radial_layer_list)
 	while(affecting.burn_dam && do_after(user, repair_time, TRUE, src, BUSY_ICON_BUILD) && use(1))
 		user.visible_message(span_warning("\The [user] fixes some wires in [H == user ? "[H.p_their()]" : "[H]'s"] [affecting.display_name] with \the [src]."), \
 			span_warning("You patch some wires in [H == user ? "your" : "[H]'s"] [affecting.display_name]."))
-		affecting.heal_limb_damage(0, 15, robo_repair = TRUE, updating_health = TRUE)
-		H.UpdateDamageIcon()
+		if(affecting.heal_limb_damage(0, 15, robo_repair = TRUE, updating_health = TRUE))
+			H.UpdateDamageIcon()
 		if(!amount)
 			return TRUE
 	return TRUE
--- a/code/modules/power/cable.dm
+++ b/code/modules/power/cable.dm
@@ -502,13 +502,14 @@ GLOBAL_LIST(cable_radial_layer_list)
 	if(H == user)
 		repair_time *= 3
 
-	user.visible_message(span_notice("[user] starts to fix some of the wires in [H]'s [affecting.display_name]."),\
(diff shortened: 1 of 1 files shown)


code/modules/power/cable.dm
```diff
@@ -502,13 +502,14 @@ GLOBAL_LIST(cable_radial_layer_list)
...
-	user.visible_message(span_notice("[user] starts to fix some of the wires in [H]'s [affecting.display_name]."),\
+	user.visible_message(span_notice("[user] starts to fix some of the wires in [H == user ? "[H.p_their()]" : "[H]'s"] [affecting.display_name]."),\
 		span_notice("You start fixing some of the wires in [H == user ? "your" : "[H]'s"] [affecting.display_name]."))
 
 	while(affecting.burn_dam && do_after(user, repair_time, TRUE, src, BUSY_ICON_BUILD) && use(1))
-		user.visible_message(span_warning("\The [user] fixes some wires in [H == user ? "your" : "[H]'s"] [affecting.display_name] with \the [src]."), \
+		user.visible_message(span_warning("\The [user] fixes some wires in [H == user ? "[H.p_their()]" : "[H]'s"] [affecting.display_name] with \the [src]."), \
 			span_warning("You patch some wires in [H == user ? "your" : "[H]'s"] [affecting.display_name]."))
 		affecting.heal_limb_damage(0, 15, robo_repair = TRUE, updating_health = TRUE)
+		H.UpdateDamageIcon()
```
psykzz, line 512:
```suggestion
		if(affecting.heal_limb_damage(0, 15, robo_repair = TRUE, updating_health = TRUE))
			H.UpdateDamageIcon()
```
This just keeps the code like it is elsewhere otherwise 👍

psykzz (changes requested):

code/modules/power/cable.dm
ivanmixo: Pointless check.

ivanmixo: This code is only ran if the do_after succeeds.

psykzz: Honestly, i dont like the idea we manually call `UpdateDamageIcon` at all, and would rather the heal_limb proc handled it, but if we are, we should be consistent with the rest of the codebase.

This isn't a hot loop and so the check isn't game breaking (not like we have 50 synths running around.

If you do feel strongly about this, lets explore how we can remove the call, or move it to a signal where the icon is updated next tick if the signal is present?

ivanmixo: I agree manually calling `UpdateDamageIcon` is stinky, I'll see what I can do.

ivanmixo: Nevermind! TGMC organs and limb code sucks and I can't be bothered. I'll just do what you said, pfft.

TiviPlus (approved):

PR #9012, opened 2021-12-19
Razorwire grenades can no longer be opened

Verlurk:
## About The Pull Request

Razorwire grenades and canister can no longer be opened with wirecutters and the containers inside removed.

## Why It's Good For The Game

Previous fixes have not worked to stop this.

Certain individuals have been abusing the razorwire recipe and the fact that the razorwire grenades can be opened to coat maps in razorwire, in combination with plasmacutting everything in sight. This has made rounds with those individuals online generally unfun for xenos, in that they're extremely restricted due to the sheer volume of it.

This solves one of those problems by making purchaseable razorwire grenades, in line with regular grenades, unopenable. The contents of said grenades remain the same so their intended use will be unchanged, and the recipe remains the same for admin shenanigans. Tested, and works as intended with no visible issues.

(Yes it's yanderedev-tier if statements I know)

## Changelog
:cl:

balance: Razorwire grenades and canisters can no longer have their contents removed.
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/chem_grenade.dm
+++ b/code/game/objects/items/explosives/grenades/chem_grenade.dm
@@ -102,10 +102,10 @@
 			return
 
 	else if(stage == CG_READY && I.tool_behaviour == TOOL_WIRECUTTER && !active)
-		if(name == "Razorburn Grenade")
+		if(ispath(/obj/item/explosive/grenade/chem_grenade/razorburn_smol))
 			to_chat(user, span_notice("The [initial(name)] does not open."))
 			return
-		if(name == "Razorburn Canister")
+		if(ispath(/obj/item/explosive/grenade/chem_grenade/razorburn_large))
 			to_chat(user, span_notice("The [initial(name)] does not open."))
 			return
 		stage_change(CG_WIRED)
--- a/code/game/objects/items/explosives/grenades/chem_grenade.dm
+++ b/code/game/objects/items/explosives/grenades/chem_grenade.dm
@@ -102,6 +102,9 @@
 			return
 
 	else if(stage == CG_READY && I.tool_behaviour == TOOL_WIRECUTTER && !active)
+		if(name == "Razorburn Grenade")
+			to_chat(user, span_notice("The [initial(name)] does not open."))
+			return
 		stage_change(CG_WIRED)
 		to_chat(user, span_notice("You unlock the [initial(name)] assembly."))
 
--- a/code/game/objects/items/explosives/grenades/chem_grenade.dm
+++ b/code/game/objects/items/explosives/grenades/chem_grenade.dm
@@ -105,6 +105,9 @@
 		if(name == "Razorburn Grenade")
 			to_chat(user, span_notice("The [initial(name)] does not open."))
 			return
+		if(name == "Razorburn Canister")
+			to_chat(user, span_notice("The [initial(name)] does not open."))
+			return
 		stage_change(CG_WIRED)
 		to_chat(user, span_notice("You unlock the [initial(name)] assembly."))
 
--- a/code/game/objects/items/explosives/grenades/chem_grenade.dm
+++ b/code/game/objects/items/explosives/grenades/chem_grenade.dm
@@ -40,6 +40,13 @@
 		else
 			return ..()
 
+/obj/item/explosive/grenade/chem_grenade/razorburn_smol/attackby(obj/item/I, mob/user, params)
+	to_chat(user, span_notice("The [initial(name)] is hermetically sealed, and does not open."))
+	return
+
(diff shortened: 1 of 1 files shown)


ivanmixo:
This isn't the way to do this at all. They could use a hand labeler to change it's name and then open it. Pfft.

Verlurk:
> This isn't the way to do this at all. They could use a hand labeler to change it's name and then open it. Pfft.

Got a solution?

Pill-U:
Just tested, can confirm hand labeler circumvents this. Perhaps use
if(I == /obj/item/explosive/grenade/chem_grenade/razorburn_large) and
if(I == /obj/item/explosive/grenade/chem_grenade/razorburn_smol) instead?

Pill-U:
give it an item_flag called UNDECONSTRUCTABLE or sthn and check for that flag?

ivanmixo:
http://www.byond.com/docs/ref/#/proc/istype

Verlurk:
> http://www.byond.com/docs/ref/#/proc/istype

Thanks, ispath works.

BraveMole (changes requested):
nono, the proper fix is to overwrite attackby on /obj/item/explosive/grenade/chem_grenade/razorburn_large and smol

Verlurk (commented):
Requested changes made.

Pill-U:
don't think you need obj/item/I, params since it's not used?

Verlurk:
> don't think you need obj/item/I, params since it's not used?

Without them, the line in chat doesn't seem to trigger as it doesn't recognise it being slapped by a tool.

BraveMole (approved):

Mrrpip:
And is there measures in place to allow medical nanites to be made with it unopenable?
(Even if they are shit)

Verlurk:
> And is there measures in place to allow medical nanites to be made with it unopenable? (Even if they are shit)

@Mrrpip Not the purpose of this PR. Honestly, by the looks of the recipe, you could just remove nanites as a requirement for the reaction, as lemoline's enough of a barrier!

Mechswordsman:
Well, guess the convenience of it's use is no completely nuked.  And now medical nanites are completely impossible to get. Good work.

Verlurk:
> Well, guess the convenience of it's use is no completely nuked. And now medical nanites are completely impossible to get. Good work.

The grenades still work as intended, it's just that now their contents can't be misused. As stated above, medical nanites are not the purpose of this PR.

Mechswordsman:
> 

The only place to get the nanites, to make medical nanites, is from razorburn  gernades. They are from the same thing. That is why you can open them,

Verlurk:
> > 
> 
> The only place to get the nanites, to make medical nanites, is from razorburn gernades. They are from the same thing. That is why you can open them,

The reason you can open them is because they're a type of chemical grenade, not because you're supposed to use the goodies inside for other reasons. There not being a way to get nanites without appropriating razorburn grenades seems like an oversight, and something that would fit perfectly into new researcher stuff.

Mechswordsman:
> > > 
> > 
> > 
> > The only place to get the nanites, to make medical nanites, is from razorburn gernades. They are from the same thing. That is why you can open them,
> 
> The reason you can open them is because they're a type of chemical grenade, not because you're supposed to use the goodies inside for other reasons. There not being a way to get nanites without appropriating razorburn grenades seems like an oversight, and something that would fit perfectly into new researcher stuff.

It isn't a oversight. It was intentional on their creators part

Also, if you're going to use the other excuse  that it 'makes the round unfun for xenos', then why not also make it where taking any damage interrupts acid drops on structures, or being able to do that while inside of a freshly built resin wall which has tons of armor and protects from damage too? A acid drop that completely destroys something that takes way, way longer than a xeno wall to build.

OR maybe how marines have to slowly hack through the entire map worth of resin walls to ever push without dieing. That isn't fun either. But you don't see people removing the Xeno's ability to maze the entire map before the shutters open. Or For the hivemind to build solid walls many, many, MANY tiles thick behind the marine pushes.

InterroLouis:
This shouldn't go through at all without addressing the problems it creates with medical nanites.

ivanmixo:
Do walls hurt the marines to hit? Can xenos spit acid at the marines trying to break the walls thru the walls themselves?

jroinc:
This is against the spirit of the initial PR, and *another* step in narrowing the gameplay loop to "click vendor, drop, 5 minutes, repeat."

I really, really don't know why you think this is a good idea at this point, but it's apparently a theme.

BraveMole:
> This is against the spirit of the initial PR, and _another_ step in narrowing the gameplay loop to "click vendor, drop, 5 minutes, repeat."
> 
> I really, really don't know why you think this is a good idea at this point, but it's apparently a theme.

The spirit of the pr was to allow sea of wires. The sooner this is gone the better

PR #9017, opened 2021-12-19
Adds Nanites to Research Rewards

Verlurk:
## About The Pull Request

Adds 10u bottles of nanites to researcher uncommon rewards tier, coupled with #9012 .

## Why It's Good For The Game

Stops people insulting me on Discord for removing the unintended source of nanites in the form of razorwire grenades (more commonly used to spam dropper wire and not for medical nanites). Means medical nanites and razorwire can be produced in limited quantities with an active researcher.

## Changelog
:cl:
add: Nanite bottles added to Research Rewards.

/:cl:

PR #9020, opened 2021-12-19
(Medical) Nanomachines: For all!

Mrrpip:
Okay, so, we know nanomachines are crummy, they drain your blood and usually end up doing more harm than good, sure, we could buff them.

Or we could give them to everyone, JPR mentioned them being a fun toy for researchers. Let's make it a fun toy for everyone!

## About The Pull Request

Adds Nanomachine injectors and Bottles to the Marinemed and Chemistry vendors respectively, for all to use.

Due to the current questions about razorwire nades and such, i'm hoping to throw my hat in for a solution to the new Medical Nanomachine problem, partly due to Nanomachines not being obtainable in such a case (Reffer to PR #9012).

## Why It's Good For The Game

I believe, that while (Medical) Nanomachines are practically useless, several players have mentioned they're fun to use.
Now, I think Mera/Dermaline already fits the bill for Lemoline healing, with the added bonus of not sucking your blood dry or giving you toxin damage.

Along with that, balance wise, I don't think it'll affect much, partly due to being a gimmick chem and partly because other Lemoline based medicines have alternative obtainment methods (For example Ambrosia Deus for Somolent, Blood tomatos for QC+, some other plant for Quietus.)

## Changelog
:cl:
add: Added Medical Nanites to the common marine vendors, son!
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -261,6 +261,14 @@
 		/datum/reagent/medicine/rezadone = 15,
 	)
 
+/obj/item/reagent_containers/hypospray/autoinjector/medicalnanites
+	name = "nanomachines autoinjector"
+	desc = "An auto-injector loaded with medical nanites. A potent new method of healing that that reproduces using a subject's blood and has a brief but potentially dangerous activation period! Beware of Neurotoxin!
+	icon_state = "autoinjector-6"
+	amount_per_transfer_from_this = 1
+	volume = 1
+	list_reagents = list(/datum/reagent/medicine/research/medicalnanites = 1)
+
 /obj/item/reagent_containers/hypospray/autoinjector/pain //made for debugging
 	name = "liquid pain autoinjector"
 	desc = "An auto-injector loaded with liquid pain. Ow."
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -263,7 +263,7 @@
 
 /obj/item/reagent_containers/hypospray/autoinjector/medicalnanites
 	name = "nanomachines autoinjector"
-	desc = "An auto-injector loaded with medical nanites. A potent new method of healing that that reproduces using a subject's blood and has a brief but potentially dangerous activation period! Beware of Neurotoxin!
+	desc = "An auto-injector loaded with medical nanites. A potent new method of healing that that reproduces using a subject's blood and has a brief but potentially dangerous activation period! Beware of Neurotoxin!"
 	icon_state = "autoinjector-6"
 	amount_per_transfer_from_this = 1
 	volume = 1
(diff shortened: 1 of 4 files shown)


InterroLouis:
Nanomachines, son.

ThelastArchiver:
they harden in response to physical trauma. ( also i agree due to them being sometimes even more deadly then a xeno it probably wont affetc balance to much except maybe more marines dying to lack of blood or medics start bringing bloodbags down.)

Mrrpip:
You can't hurt me, Jack. (We'd also get to see Nanoblood used for once! _(Probably)_)

Franky2025:
CANCER, PUTTING THAT INTO A VENDOR IS FULL ON CANCER AND DOES NOT FIX THEIR USSELESSNES

Mrrpip:
> CANCER, PUTTING THAT INTO A VENDOR IS FULL ON CANCER AND DOES NOT FIX THEIR USSELESSNES

-Well. It does fix the fact that they're unobtainable without a convoluted recipe that relies on research that is now broken.

-Does not give you the downsides of having to swap your entire build when you get it from Xeno research

-Lemoline was a mistake for the gimmick chems, Somolent, the only usable one of the gimmick chems can be obtained trough botany by anyone and that's 10x as useful, 

The general consensus seems to be they're useless but fun. We should let people have their fun.


> It must be free under the vendors of marine medics and medical staff, I dont think that nanites should be that easy to get. I have seen sick fucks inject 0.1u of nanites into food and force feeding them to people. This just shouldn't be able to flow around without the knowledge or supervision of the medical staff.

In this case. I agree, some player COULD inject nanites into everything and forcefeed it. But it would most likely be a staff issue if brought up to that degree. (That and we barely see it with Russian Red)

If we were to relegate it to medic vendors, who would use it? The medics certainly wouldn't unless it was the above case. Or you'd have a PFC begging medics each round.

If it was NTmed+ only, well. It gets hacked each round.

Along with that, hypervene injectors are kept in the dropships and usually by medics. In the supposed case they did forcefeed everyone nanites and got away with it rule-wise. It is much easier to fix than Russian Red.

I believe this method is a happy medium that will allow for gimmick builds without much change balance wise.

Franky2025:
the meta will be that everyone in prep time just fill themselves with nanites, sharing between everyone and you have a whole marine force that is every time on nanites, it will be a meta, i do think nanites should have an another way of obtaining them but not that

Mrrpip:
Okay, so.

Our current choices seem to be:
-We let everyone use nanites, which has the downsides of what you've highlighted above, but would lead to the least problems getting it
-We let medics buy them: Which will lead to people going medic just for nanomachines OR people constantly begging medics to use some of their points for nanomachines
-We make it obtainable trough plants: It works for Somolent and QC+, map dependant, some people will find a way to powergame that.
-We go with the other PR and make it research exclusive via the Analyzer thingy. Which is eh, because it's RNG for something so gimmicky, the shoulder gun/mantis blade is plug and play, Nanites need 10 tons of other chems to not kill yourself with it.
-We think up a new chem recipe: It'll either be bum easy to obtain or lemoline based which makes it one of the rarest things for something that once you make it, is infinite anyway.
-We suddenly think up something else.

Obviously, i'd go with the first one, but i think #3 or #5 could work (Provided the recipe doesn't make you want to bang your head on a wall)

BraveMole:
I much prefer https://github.com/tgstation/TerraGov-Marine-Corps/pull/9017 and i don't want two solutions for this

BraveMole (approved):

PR #9024, opened 2021-12-20
Medical Nanites buff

KAAITTT:
## About The Pull Request
Reduces the time it takes to adopt to 100 cycles.
Slightly reduces the amount of damage nanites will do in the adopting period.
Adds pain reduction to nanites based on the amount of them in the blood in the adopted state. Less nanites = more being used = smart nanites reduce pain.

## Why It's Good For The Game
Nanites are not being used much, the only way to get them anyways is to bug an engineer, minor heal based on your constant blood consumption and toxin damage after almost 5 minutes of close to dying crit even with meds. They deserve at least painkilling abilities, its the nanomachines, son, after all, it only makes sense for them to do that.

## Changelog
:cl: Kait
balance: Nanites buffed: Pain reduction added, time to adopt shortened. 
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1362,7 +1362,7 @@
 			if(volume < 35) //allows 10 ticks of healing for 20 points of free heal to lower scratch damage bloodloss amounts.
 				L.reagents.add_reagent(/datum/reagent/medicine/research/medicalnanites, 0.1)
 				
-				if(volume < 30) // Pain reduction matters on the amount of nanites in the blood, less nanites = more being used to heal = a much severe condition = probably more pain
+				if(volume < 31) // Pain reduction matters on the amount of nanites in the blood, less nanites = more being used to heal = a much severe condition = probably more pain
 				L.reagent_pain_modifier += PAIN_REDUCTION_LIGHT
 			
 				if(volume < 25) //pain reduction medium, 20 ticks of continues healing
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1330,6 +1330,7 @@
 /datum/reagent/medicine/research/somolent/overdose_process(mob/living/L, metabolism)
 	holder.remove_reagent(/datum/reagent/medicine/research/somolent, 1)
 
+
 /datum/reagent/medicine/research/medicalnanites
 	name = "Medical nanites"
 	description = "These are a batch of construction nanites altered for in-vivo replication. They can heal wounds using the iron present in the bloodstream. Medical care is recommended during injection."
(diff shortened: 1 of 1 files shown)


code/modules/reagents/reagents/medical.dm
```diff
@@ -1344,30 +1345,39 @@
...
-
+				
+				if(volume < 30) // Pain reduction matters on the amount of nanites in the blood, less nanites = more being used to heal = a much severe condition = probably more pain
+				L.reagent_pain_modifier += PAIN_REDUCTION_LIGHT
+			
+				if(volume < 25) //pain reduction medium, 20 ticks of continues healing
+				L.reagent_pain_modifier += PAIN_REDUCTION_HEAVY
+				
+				if(volume < 20) // pain reduction a lot, 30 ticks of continues healing
+				L.reagent_pain_modifier += PAIN_REDUCTION_VERY_HEAVY
```
psykzz, line 1372: Check your indentation these are currently not valid.

psykzz (changes requested):

Franky2025:
i have an idea now to make it based on the damage you have at the moment, gonna try that better cause i dont see where did i make a mistake

Franky2025:
it actually worked, so stupid to me, after lua, that it actually does matter.

BraveMole:
Allright, after thinking about this, such a "useless" item does not belong with shoulder gun and armblade. So we'll go with the other solution instead

BraveMole:
Sorry wrong pr lel

BraveMole (approved):
Ok we can try that.

KAAITTT:
rethinking this and discussing this, concerns of this being too good will be reviewed after some playtesting, i dont want to contribute a part of the powercreep issue.

ThelastArchiver:
doc turn off my pain inhibitors ! ( sounds like a good pr)

Licherr:
I personally don't like any changes to the medical nanites as I perceive that the reason it was made alongside some other "useless" chems for experimental funsies for the researcher pre-gacha, and not to be a viable medical chem. When the nanites became public everyone now thinks it should be less destructive and more viable. I just don't like having new chems that already has the same functions as the present chems we already have.

KAAITTT:
> I personally don't like any changes to the medical nanites as I perceive that the reason it was made alongside some other "useless" chems was for experimental funsies for the researcher pre-gacha, and not to be a viable medical chem. When the nanites became public everyone now thinks it should be less destructive and more viable. I just don't like having new chems that already has the same functions as the present chems we already have.

Stop mentioning the idiotic merge of the "nanites for all", its irrelevant to the state of the chem independently, and this pull request was made before it.
This is a basic misunderstanding of the researcher.
Researcher things exist not only for the researchers to just DO THEM and thats it, they have effects for a reason. There is a reason researchers get REWARDS out of xenos, like implants and so on, that can be used. This whole game relays on team work, and your job being: doing something not for the sake of it but because it does something to the whole state of your workplace.

So the researchers gimmicks of "nanites" and them not needing to be useful makes no sense. This is an item that requires quite a bit of chem resources and work to get, for a single portion, for a single person to use. These changes were made looking at the whole state of it, work needing to be done and the balance of it in mind. This is an improvement to of it, because i think nanites need to have better performance, a better pay off for their process of creation.

BraveMole:
since no one likes nanites being there

PR #9025, opened 2021-12-21
Somolent crit heal

Franky2025:
## About The Pull Request
Adds inaprov like crit heal to somolent.
Condition: Patient in crit, more or equal 10 units administered, and unconscious.
Heals 15% all burn&brute
Organ damage to 25.
Does 50 stamina damage.
Has 300 seconds cooldown like everything else.

## Why It's Good For The Game
Somolent is described as "highly potent regenerative drug, designed to heal critically injured personnel", now its true :).
Being research gimmick, annoying to craft and its niche usage only unconscious, it needs more then just healing. Crit stabilization at a cost of hard stamina damage to me seems fair trade to make it little bit more useful then it is now.

## Changelog
:cl: Kait
balance: Somolent can crit heal
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1308,6 +1308,23 @@
 	scannable = TRUE
 	overdose_threshold = REAGENTS_OVERDOSE
 	taste_description = "naptime"
+	
+/datum/reagent/medicine/somolent/on_mob_add(mob/living/L, metabolism)
+	var/mob/living/carbon/human/H = L
+	if(TIMER_COOLDOWN_CHECK(L, name))
+		return
+	if(L.health < H.health_threshold_crit && volume > 9 && L.stat == UNCONSCIOUS) //If you are in crit, unconscious, and someone injects at least 10u into you at once, you will heal 15% of your physical damage instantly and organ below 25.
+		to_chat(L, span_userdanger("you feel much better but very exhausted"))
+		L.adjustStaminaLoss(50*effect_str)
+		L.adjustBruteLoss(-L.getBruteLoss() * 0.15)
+		L.adjustFireLoss(-L.getFireLoss() * 0.15)
+		L.jitter(5)
+		for(var/datum/internal_organ/I AS in H.internal_organs)
+			if(I.damage)
+				if(I.damage < 25)
+					return
+				I.heal_organ_damage((I.damage-25) *effect_str)
+		TIMER_COOLDOWN_START(L, name, 300 SECONDS)
 
 /datum/reagent/medicine/research/somolent/on_mob_life(mob/living/L, metabolism)
 	switch(current_cycle)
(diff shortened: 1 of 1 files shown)


Mrrpip:
While the theory is good, Frank, It'd be easy to use with the Vali, which is already infamous for stacking Inaprovaline and Russian Red.

While you can't use all 3 at once, due to 10 + 15 + 10 = 35. I can see players using 2 chems, then reloading with the third.

While it's up to you, I think Somolent having both a strong pre-med(The healing ramp up) and instant (Crit heal) effects would not be great balance-wise.

Franky2025:
nobody is using somolent, the fact you need to be unconscious for everything to work and the amount of stamina damage it does makes it unusable for the robust vali gang.

Mrrpip:
That, I agree with. I suppose making the ramp-up quicker wouldn't help much, would it?

Franky2025:
> That, I agree with. I suppose making the ramp-up quicker wouldn't help much, would it?

i already did from 50 to 25, i feel like its fine at 25.

Mrrpip:
Must've missed it. My bad.

BraveMole (changes requested):
Don't see how this drug fits in the current drug panel and the why it's good for the game doesn't help me understanding that

Franky2025:
niche drug with niche healing abilities for niche scenarios that rarely anyone bother to create. Other hard to create and potent healing drugs such as neuraline, and an uncraftable RR have the same effect, because regular damage isnt the only type of damage. Its not a better version of ina, you can say its a variant for a niche usage of stabilizing patients with big number of brute/burn.
This changes somolent from being "Useful in 1 out of 20 scenarios" to having actual utility

Franky2025:
i keep focusing on the fact its niche because its important, its a research gimmick that should pay off after bothering with it.

BraveMole:
You're not giving it a niche, you're just taking inapro niche and making it a worse version. I see no reason to use this rather than inapro, and using inapro + somnolent seems like overkill

Franky2025:
i hope you understand that this is not unique to inaprov lol, russian red has the same effect, neuraline aswell, i used russian red+inaprov+neuraline crit heals already in game, this proves nothing.

BraveMole:
So even more bloat?

Franky2025:
Same level of logic as to complain that i can have my full inventory of grenades and spam grenades just because they are there and i can, no matter that it leaves me with no gun and other supplies. Russian Red is not craftable, neuraline is insanely difficult to craft due to lack of larva, inaprov is the only crit healing drug that actully is being used, i have not seen anyone besides me do what i just described, because is spent an entire round on it, why did i said it in the first place? Because you seem to be oblivious to the existence of the effect on other reagents, like im making this an inaprov+, despite it being worse in crit healing then ina. Ina heals more organ and overral crit damage then somolent crit heal, there is no point to use this both besides extra 15% damage heal, i am not seeing the reason to phrase it like some sort of unbalanced abuse of crit heal. Why this was made i already explained, somolent is annoying to make, requires lemoline as catalyst at least, and the damage healing properties is not the main concern in game, its usefulness  is hindered by needing to be asleep, the aim of this was to expand its range of usage a little bit, to justify its quirks.

Franky2025:
it kills your stamina and besides needing to be in crit you have to be also unconscious during crit, nothing like putting your critical patients to sleep speaks "overkill"

PR #9026, opened 2021-12-21
Moving harvester code into component

Waselon:
## About The Pull Request

Moving harvester code into component, changing affected items to work with that change.

## Why It's Good For The Game

Allows to add harvester component to items without inheriting harvester blade

## Changelog
:cl:
fix: adjusted flame cone code to work with the spear
code: moved harvester code into component
code: made needed changes to blade and spear
/:cl:


code/game/objects/items.dm
```diff
@@ -374,6 +374,7 @@
 
 ///Anything unique the item can do, like pumping a shotgun, spin or whatever.
 /obj/item/proc/unique_action(mob/user)
+	SEND_SIGNAL(src, COMSIG_ITEM_UNIQUE_ACTION, user)
```
BraveMole, line 377: in general, if you are sending a signal at a parent proc, every child proc must call the parent proc. This signal was removed because it's impossible with the current gun architecture to call the unique action parent. 

Having this signal here would only cause confusion; a future coder might see this and use it, and not discover that guns actually don't use it.

An alternative i can suggest is using right click for the unique action stuff on vali, it might be way more intuitive that way; and you wouldn't have this issue with the signal

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+			M.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 1 SECONDS)
+
+		if(/datum/reagent/medicine/kelotane)
+			src.apply_flame_cone(M, user)
+			var/list/cone_turfs = generate_cone(M, 1, 0, 181, Get_Angle(user, M.loc))
+			for(var/X in cone_turfs)
+				var/turf/T = X
+				for(var/mob/living/victim in T)
+					src.apply_flame_cone(victim, user)
+					//TODO BRAVEMOLE
```
BraveMole, line 126:
```suggestion
					apply_flame_cone(victim, user)
```
no src.proc, just proc

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+	if(M.status_flags & INCORPOREAL || user.status_flags & INCORPOREAL) //Incorporeal beings cannot attack or be attacked
+		return FALSE
+
+	switch(loaded_reagent.type)
+		if(/datum/reagent/medicine/tramadol)
+			M.apply_damage(W.force*0.6, BRUTE, user.zone_selected)
+			M.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 1 SECONDS)
+
+		if(/datum/reagent/medicine/kelotane)
+			src.apply_flame_cone(M, user)
```
BraveMole, line 120:
```suggestion
			apply_flame_cone(M, user)
```

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+	to_chat(user, span_rose("You start filling up the small chambers along the blade's edge."))
+	if(!do_after(user, 2 SECONDS, TRUE, source, BUSY_ICON_BAR, ignore_turf_checks = TRUE))
+		to_chat(user, span_rose("Due to the sudden movement, the safety machanism drains out the reagent back into the main storage."))
+		return FALSE
+
+	loaded_reagent = beaker.reagents.reagent_list[1]
+	beaker.reagents.remove_any(5)
+	return TRUE
+
+/datum/component/harvester/proc/apply_flame_cone(mob/living/target, mob/living/user)
```
BraveMole, line 102: this is a useless proc. In byond we don't split procs in small bit unless it's really necessary, because the proc overhead cost is huge.

If you don't want to repeat code, make it a macro

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+
+	to_chat(user, span_notice("You begin filling up the [source] with [container.reagents.reagent_list[1]]."))
+	if(!do_after(user, 1 SECONDS, TRUE, source, BUSY_ICON_BAR, null, PROGRESS_BRASS))
+		return TRUE
+
+	trans = container.reagents.trans_to(beaker, container.amount_per_transfer_from_this)
+	to_chat(user, span_rose("You load [trans]u into the internal system. It now holds [beaker.reagents.total_volume]u."))
+	return TRUE
+
+/datum/component/harvester/proc/unique_action(datum/source, mob/user)
```
BraveMole, line 81: every /proc/ must be autodoced, aka :
///Signal handler calling the unique action of the harvester

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+
+/datum/component/harvester/proc/examine(datum/source, mob/user)
+	SIGNAL_HANDLER
+	to_chat(user, span_rose("[length(beaker.reagents.reagent_list) ? "It currently holds [beaker.reagents.total_volume]u of [beaker.reagents.reagent_list[1].name]" : "The internal storage is empty"].\n<b>Compatible chemicals:</b>"))
+	for(var/R in loadable_reagents)
+		var/atom/L = R
+		to_chat(user, "[initial(L.name)]")
+
+/datum/component/harvester/proc/attackby(datum/source, obj/item/I, mob/user)
+	if(user.do_actions)
```
BraveMole, line 38:
```suggestion
/datum/component/harvester/proc/attackby(datum/source, obj/item/I, mob/user)
	SIGNAL_HANDLER
	if(user.do_actions)
```

Every proc that receive a signal must be declared as a signal handler

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
+/datum/component/harvester
+	var/force
```
BraveMole, line 2: variables must be autodoc (same thing as new procs)

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+		/datum/reagent/medicine/bicaridine,
+		/datum/reagent/medicine/tramadol,
+		/datum/reagent/medicine/kelotane,
+	)
+
+/datum/component/harvester/Initialize()
+	if(!isitem(parent))
+		return COMPONENT_INCOMPATIBLE
+
+	beaker = new /obj/item/reagent_containers/glass/beaker/vial
```
BraveMole, line 15: this is prone to harddel, ping me

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+		COMSIG_PARENT_EXAMINE,
+		COMSIG_ITEM_UNIQUE_ACTION,
+		COMSIG_ITEM_ATTACK,
+		COMSIG_PARENT_ATTACKBY))
+
+/datum/component/harvester/proc/examine(datum/source, mob/user)
+	SIGNAL_HANDLER
+	to_chat(user, span_rose("[length(beaker.reagents.reagent_list) ? "It currently holds [beaker.reagents.total_volume]u of [beaker.reagents.reagent_list[1].name]" : "The internal storage is empty"].\n<b>Compatible chemicals:</b>"))
+	for(var/R in loadable_reagents)
+		var/atom/L = R
```
BraveMole, line 34:
```suggestion
	for(var/atom/L AS in loadable_reagents)
```
And change the variable name to a word, no alphabet variable please

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,142 @@
...
+	switch(loaded_reagent.type)
+		if(/datum/reagent/medicine/tramadol)
+			M.apply_damage(W.force*0.6, BRUTE, user.zone_selected)
+			M.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 1 SECONDS)
+
+		if(/datum/reagent/medicine/kelotane)
+			src.apply_flame_cone(M, user)
+			var/list/cone_turfs = generate_cone(M, 1, 0, 181, Get_Angle(user, M.loc))
+			for(var/X in cone_turfs)
+				var/turf/T = X
```
BraveMole, line 123:
```suggestion
			for(var/turf/T AS in cone_turfs)
```

BraveMole (changes requested):
good first attempt at making a component, just small remarks but the rest is good

code/datums/components/harvester.dm
Waselon: if i add SIGNAL_HANDLER, i get this error 
изображение

code/modules/mob/mob_helpers.dm
```diff
@@ -516,7 +516,11 @@ mob/proc/get_standard_bodytemperature()
 
 	var/obj/item/active_item = get_active_held_item()
 	if((istype(active_item) && active_item.unique_action(src) != COMSIG_KB_NOT_ACTIVATED) || client?.prefs.unique_action_use_active_hand)
+		if(active_item != null)
+			SEND_SIGNAL(active_item, COMSIG_ITEM_UNIQUE_ACTION, src)
```
BraveMole, line 520: nop, unique_action can be called by different means (verb and click). 

Since you prefer to keep unique action as the actionner, we can use the old solution for the signal sent.

every time something is calling unique_action, have a wrapper proc that sends the signal and then call unique action/

This is awfull of course, but the proper fix would require a heavy refactor and gun and that's shit

BraveMole (changes requested):
almost there

code/modules/mob/mob_helpers.dm
Waselon: I wanted to keep it as close to original as possible, or if not possible, at least convenient to use. Right-clicking the weapon sprite wasn't convenient IMO. Other solution I can propose is to make reagents autoload into the blade.

code/modules/mob/mob_helpers.dm
```diff
@@ -515,8 +515,8 @@ mob/proc/get_standard_bodytemperature()
 		return
 
 	var/obj/item/active_item = get_active_held_item()
-	if((istype(active_item) && active_item.unique_action(src) != COMSIG_KB_NOT_ACTIVATED) || client?.prefs.unique_action_use_active_hand)
+	if((istype(active_item) && active_item.do_unique_action(src) != COMSIG_KB_NOT_ACTIVATED) || client?.prefs.unique_action_use_active_hand)
 		return
 	var/obj/item/inactive_item = get_inactive_held_item()
 	if(istype(inactive_item))
-		inactive_item.unique_action(src)
+		inactive_item.do_unique_action(src)
```
BraveMole, line 522: you need to replace every unique_action call by do_unique_action

code/game/objects/items.dm
```diff
@@ -372,6 +372,12 @@
 /obj/item/proc/item_action_slot_check(mob/user, slot)
 	return TRUE
 
+///Signal sender for unique_action
+/obj/item/proc/do_unique_action(mob/user)
+	SEND_SIGNAL(src, COMSIG_ITEM_UNIQUE_ACTION, user)
+	unique_action(user)
+	return
```
BraveMole, line 379:
```suggestion
	return unique_action(user)
```

BraveMole (changes requested):

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,164 @@
...
+/datum/component/harvester/proc/examine(datum/source, mob/user)
+	SIGNAL_HANDLER
+	to_chat(user, span_rose("[length(beaker.reagents.reagent_list) ? "It currently holds [beaker.reagents.total_volume]u of [beaker.reagents.reagent_list[1].name]" : "The internal storage is empty"].\n<b>Compatible chemicals:</b>"))
+	for(var/atom/reagent AS in loadable_reagents)
+		to_chat(user, "[initial(reagent.name)]")
+
+///Handles behavior for when item is clicked on
+/datum/component/harvester/proc/attackby_async(datum/source, obj/item/cont, mob/user)
+	if(user.do_actions)
+		return TRUE
```
BraveMole, line 48: everything async cannot return value. So just replace everything with return rather than return TRUE

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,164 @@
...
+			if((target != user) && do_after(user, 2 SECONDS, TRUE, target, BUSY_ICON_DANGER))
+				target.heal_overall_damage(12.5, 0, TRUE)
+			else
+				target.adjustStaminaLoss(-30)
+				target.heal_overall_damage(6, 0, TRUE)
+			loaded_reagent = null
+			return FALSE
+
+	loaded_reagent = null
+	return
```
BraveMole, line 149: no need for a return last line

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,164 @@
...
+				return
+			to_chat(user, span_rose("You prepare to stab <b>[target != user ? "[target]" : "yourself"]</b>!"))
+			new /obj/effect/temp_visual/telekinesis(get_turf(target))
+			if((target != user) && do_after(user, 2 SECONDS, TRUE, target, BUSY_ICON_DANGER))
+				target.heal_overall_damage(12.5, 0, TRUE)
+			else
+				target.adjustStaminaLoss(-30)
+				target.heal_overall_damage(6, 0, TRUE)
+			loaded_reagent = null
+			return FALSE
```
BraveMole, line 146: useless return

code/datums/components/harvester.dm
```diff
@@ -0,0 +1,164 @@
...
+		return FALSE
+
+	to_chat(user, span_rose("You start filling up the small chambers along the blade's edge."))
+	if(!do_after(user, 2 SECONDS, TRUE, source, BUSY_ICON_BAR, ignore_turf_checks = TRUE))
+		to_chat(user, span_rose("Due to the sudden movement, the safety machanism drains out the reagent back into the main storage."))
+		return FALSE
+
+	loaded_reagent = beaker.reagents.reagent_list[1]
+	beaker.reagents.remove_any(5)
+	return TRUE
```
BraveMole, line 110: usless return

BraveMole (changes requested):

BraveMole (approved):

PR #9027, opened 2021-12-21
Static landing zone supply beacon

EvilMelonTheSecond:
## About The Pull Request
Adds an unremovable and inactive supply beacon to every LZ.

## Why It's Good For The Game
Most rounds the squad supply beacon is used inside the LZ to make sure drops can be made at any time. This PR gives a reliable beacon for every LZ so now req has an easier time with sending supplies and squad supply beacons can be used in the field from the freed up FOB beacon.

The beacon starts inactive and activates via clicking and waiting a few seconds from then on marines cannot pickup/deactivate the beacon. The beacon is acidable though.
BeaconPreview

## Changelog
:cl:
qol: Added an unremovable beacon to landing zones
/:cl:


Luxyboo:
wow, marines forget beacons so much that this PR now exists, incredible

Lumipharon:
REQ was given a round start beacon for free, for this reason already.

BraveMole (changes requested):
there's already 3 different way of getting a supply beacon, how is this needed?

EvilMelonTheSecond:
> there's already 3 different way of getting a supply beacon, how is this needed?

I had assumed that supply beacons were limited to 1 active beacon per squad before making this PR. So this PR solves a non-existent problem.

But it still is a small QoL I guess? An unremovable beacon is easily activated by any marine so req has less anxiety on whether or not somebody ran off without activating their squad beacon.

Also a lantern looking beacon looks lovely no matter what kuro says

BraveMole:
Honestly this seems like bloat to me, and kuro doesn't like the sprote anyway

PR #9038, opened 2021-12-23
Rebalance minions

BraveMole:
## About The Pull Request

Mantis and beetle will no longer flee.
Since they stick in the fight, nerf their hp so their overall usefulness stay the same.
Plus see CL for other changes

## Why It's Good For The Game

Close combat minions manage to flee only 5% of the time, because they are stupid. This feature only meant that marines had to shoot it longer, even when they constituate no danger; i think that's lame.

Minions will with this update generate as much damage as previously, but will eat less bullets

## Changelog
:cl:
balance: Mantis and beetle no longer flee from fights when low health
balance: Beetle hp from 400 to 260
balance: Mantis hp from 200 to 150
balance: Beetle forward charge only stuns for 2 seconds, 30 seconds of cooldown (it was 4 seconds, 10 seconds before)
balance: Mantis ravage cooldown upped from 6 seconds to 30 seconds
balance: Scorpion spit deal less damage, and the firerate is reduced
balance: Minions spawn rate divided by two. Minions maxium mob by spawner divided by two
/:cl:

### DIFF
--- a/code/__DEFINES/monitor.dm
+++ b/code/__DEFINES/monitor.dm
@@ -55,5 +55,5 @@ GLOBAL_VAR_INIT(xeno_stat_multiplicator_buff, 1)
 ///50% is the maximum buff that xeno can receive
 #define  MAXIMUM_XENO_BUFF_POSSIBLE 1.5
 
-#define MAX_SPAWNABLE_MOB_PER_PLAYER 0.15 //So for 50 players, each spawner can generate 7 mobs
-#define SPAWN_RATE_PER_PLAYER 36 //For each player, the time between two consecutive spawns is reduced by 36 ticks. So for 35 players, it's one mob every minute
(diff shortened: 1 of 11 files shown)


Lumipharon:
RIP chad beetle.

Should they still try flee if there are no marines around? Like if they get set on fire, it's a bit silly they will still charge off 50 meters away and burn to death.

ThelastArchiver:
140 is a bit much thats almost half its health maybe make 100 less for now?

Tristan63:
Perhaps make more spawn

Luxyboo:
remove AI its literally just a crutch

novaepee:
Scorpon and mantis are faster than runners.

Marines are seething about minions tanking, but I'm going to count that as skill issue. We could make minions less of a bullet sponge if possible.

PR #9043, opened 2021-12-24
Fixes CAS issues with LV and Chigusa

RipGrayson:
## About The Pull Request

The latest LV rework broke CAS without flares due to an oversight, Chigusa never had the appropriate landmark to actually start CAS without flares so I fixed that while I was at it.

## Why It's Good For The Game

Bugs bad, also PO's would probably like to fly their planes without waiting for a flare.

## Changelog
:cl:
fix: Fixed a bug preventing CAS without flares on LV624.
/:cl:


Araulius:
and there was much rejoicing.

PR #9065, opened 2021-12-26
Masterkey balance changes

Pariah919:
## About The Pull Request
Moves damage reduction from 55% to 40%, and doubles falloff.
Sets fire delay to 2 seconds, as normal shotguns.
Higher base scatter.

## Why It's Good For The Game
The masterkey now behaves like a pump action shotgun with t-39 damage malus style, but has high falloff and scatter making it hard to use beyond 1 tile, as it should be.

## Changelog
:cl:
balance: Masterkey has a lower damage malus, higher scatter and lower RPM.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -114,7 +114,7 @@
 
 /obj/item/weapon/gun/shotgun/combat/masterkey
 	name = "masterkey shotgun"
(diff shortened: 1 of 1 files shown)


Lumipharon:
Keep in mind buckshot from a masterkey already does a fraction of the damage (something like 150 vs 240?), due to having its own ammotype unlike slugs/fletch.

So with the PR you already put through, masterkey buck does 67.5 damage with no AP, or 90 if this PR goes through.
Considering buck already has a completely different damage value for masterkeys, would it just make more sense to have custom ammo for all of them (and no damage modifier on the masterkey itself)?

PR #9072, opened 2021-12-27
Praetorian has "SELF-DEFENSE" tail sweep unironically

novaepee:
## About The Pull Request

Prae acquires tail sweep, but it can only knock back marines by one tile. That is it. No stun, no paralyze, no nothing.

It's advised NOT to open with this ability against marines as they are still wielding their weapon EVEN after the tail sweep! Since it is not a stun and instead a displacement, they will continue to shoot you.

I sense prae will use acid spray then tail sweep to displace marine, but even then, it's only one tile with no stun.

Can't wait to see prae's tail sweep with primo prae's acid dash.

## Why It's Good For The Game

Xenomorph cabal, here we go.

I thought about giving this to queen since tail sweep without the stun, like back in the old days, is a good idea but have bad balance. However, queen has too much abilities, though I might do something about how much abilities she has in another PR, and prae makes sense since when cornered, acid spray wind up is not going to do a good job to ward off marines and permit it to escape.

## Changelog
:cl:
add: Prae's tail sweep! IT DOES NOT STUN! ONLY KNOCK BACK BY ONE TILE!
/:cl:


novaepee:
https://user-images.githubusercontent.com/6610922/147438947-994c55ed-b46e-4e31-88c4-c4b6438a0e89.mp4

Pariah919:
Don't really see the point of this if it doesn't stun or do anything, I guess it stops people from pointblanking you, but other than that it kinda seems like noobies will use it and get gigafragged and die, and there'll be niche uses for it otherwise.

So I guess it's okay.

BraveMole (changes requested):
Why the fuck are you copying the code for defender

novaepee:
Xenomorph cabal says no.

PR #9084, opened 2021-12-28
Alamo Expanded!

novaepee:
## About The Pull Request

Alamo has been expanded! Recent developement found that the Alamo can increase load while maintaining its aerodynamic structure. This means that the Alamo has TWO MORE CHAIRS, TWO MORE INTERIOR TILES, AND TWO MORE NANOMEDS!

Also, code in tiles that we have but never use which then I use. See code/game/turfs/open.dm

## Why It's Good For The Game

There were dead space with the Alamo that we could use for chairs and room while maintaining its aesthetics. This could also be seamlessly updated to slotbot's grates. All I did was moved practically half of Alamo down and added the space needed.

## Changelog
:cl:
expansion: Alamo expands! It has two additional chairs, interior tiles, and NanoMeds!
/:cl:

### DIFF
--- a/_maps/shuttles/alamo.dmm
+++ b/_maps/shuttles/alamo.dmm
@@ -292,8 +292,8 @@
 /obj/structure/bed/chair/dropship/passenger{
 	dir = 8
 	},
-/obj/machinery/light{
-	dir = 4
+/obj/machinery/vending/nanomed{
+	dir = 8
 	},
 /turf/open/shuttle/dropship/floor,
 /area/shuttle/dropship/alamo)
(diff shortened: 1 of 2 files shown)


novaepee:
Before:
image

After:
image

novaepee:
I'm thinking about making the side doors as one door. Thoughts on this?

SilencedMP5A5:
i would say you should make them triple doors like the rear cargo door is

like how helicopters have big ol side cargo doors

Lumipharon:
I think the thruster things on the side look a little better on the original as they don't stick out, but I doubt people would even notice.
Also speaking of the grates idea, give alamo overhead lockers by making some of the interior walls basically like morgue trays to hold shit. Y/Y?

RipGrayson:
I like it, as long as it doesn't break landing pad (and it shouldn't) I don't see why not.
I'm less sure about the side doors being one thing, but I haven't seen the sprite you would use or the delay on the new door so I can't really judge.

novaepee:
Will code in side doors later. PR did its job, but thank you all for the ideas on side doors!

PR #9089, opened 2021-12-28
T-29 Balance: Ammo Buff, Burst, and aim_slowdown

novaepee:
## About The Pull Request

Buff T-29 drum from 200 to 300. Also, remove the 4 initial T-29 drums from surplus since this buff is intended to replace that.

Also, give PMC smartgunner a mag harness. Weird that they didn't have it.

## Why It's Good For The Game

T-29 is falling behind the T-25. Nostaglic smartgunners want more ammo for T-29, yet their prayers remain unanswered.

This PR is to help T-29 smartgunners.

Currently, a T-29 smartgunner spawns with 4 SG drums, which is a total of 800 ammo (not including the other 4 SG drums that surplus gives).  A T-60 can carry more ammo than a T-29 even if both max out all the storage they can hold, yet the T-29 is supposed to be the IFF weapon system to be the superior IFF machine gun.

Over at CM SS13, a single drum holds 500 ammo, and a single CM smartgunner spawns at least 7 drums, totaling up to 3,500. While CM xenomorphs are literally bullet sponges and a tough CM round for a CM smartgunner can use more than 10,000 ammo (I have done my share and still am), a T-29 smartgunner will never see that number because (1) TGMC xenomorphs aren't bullet sponges and (2) T-29 drums don't permit that. Despite that minions and acid / resin turrets have become bullet sponges, nothing has changed for the T-29 that have to use its ammo against the increased number of xenomorph targets.

With this PR, a T-29 smartgunner will spawn with 4 SG drums with a total number of 1,200 ammo. I also delete the 4 SG drums in surplus since those 4 SG drums help early T-29 smartgunners more than late T-29 smartgunners. For context, if a T-29 smartgunner that hog all the SG drums in early game, they will have a total of 1,600 ammo, leaving the rest of T-29 smartgunners to have 800 ammo assuming they don't buy more drums. The 1,200 ammo is enough for T-29 smartgunners to get req points and continue to engage while the order is being process and deliever.

However, I know that smartgunners will abuse the new profound 1,200 ammo by attempting to run through xenomorphs, so I increase aim_slowdown. This forces T-29 smartgunners to really choose to either pointman the marine push and can barely move of the way OR snug themselves between marines in light of their aim_slowdown. Note that aim_slowdown is close to being on par with T-60 not in aim mode.

The burst fire is to synergize with being a turtle. If T-25 is not going to get it, let T-29 have it. It does have a long delay between each burst fire, but as a smartgunner you should have conservation of resources in mind all the time even with this ammo buff.

## TL;DR
T-29 smartgunners spawn with 1,200 ammo
T-29 smartgunners have increase aim_slowdown, meaning they are slower when using T-29
T-29 smartgunners have burst fire

MJP, you better be okay with this in light of the better choice that is T-25. Let T-29 have more ammo as it should.

## Changelog
:cl:
balance: T-29 smartgunners spawn with 1,200 ammo, but T-29 smartgunners have increase aim_slowdown, meaning they are slower when using T-29, so T-29 smartgunners have burst fire
del: Surplus doesn't have 4 SG drums
/:cl:

### DIFF
--- a/_maps/map_files/Theseus/TGS_Theseus.dmm
+++ b/_maps/map_files/Theseus/TGS_Theseus.dmm
@@ -2669,11 +2669,29 @@
 /area/mainship/living/cafeteria_starboard)
 "bbj" = (
 /obj/machinery/vending/marineFood,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
 /turf/open/floor/mainship/mono,
 /area/mainship/living/cafeteria_starboard)
 "bbm" = (
 /obj/machinery/firealarm,
 /obj/machinery/vending/marineFood,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
+/obj/item/reagent_containers/food/snacks/protein_pack,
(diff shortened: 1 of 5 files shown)


Waselon:
burst_delay = 0.2? Thats useless  and worse than automatic fire

novaepee:
I made this in the wrong branch, and MJP wants something else.

PR #9095, opened 2021-12-28
T-29 Balance: The Extended 350 Ammo Drum

novaepee:
## About The Pull Request

Give T-29 smartgunners the Extended 350 Ammo drum. Remove the 4 initial T-29 drums from surplus since this buff is intended to replace that.

## Why It's Good For The Game

T-29 is falling behind the T-25. Nostaglic smartgunners want more ammo for T-29, yet their prayers remain unanswered.

This PR is to help T-29 smartgunners.

Currently, a T-29 smartgunner spawns with 4 SG drums, which is a total of 800 ammo (not including the other 4 SG drums that surplus gives). A T-60 can carry more ammo than a T-29 even if both max out all the storage they can hold, yet the T-29 is supposed to be the IFF weapon system to be the superior IFF machine gun.

Over at CM SS13, a single drum holds 500 ammo, and a single CM smartgunner spawns at least 7 drums, totaling up to 3,500. While CM xenomorphs are literally bullet sponges and a tough CM round for a CM smartgunner can use more than 10,000 ammo (I have done my share and still am), a T-29 smartgunner will never see that number because (1) TGMC xenomorphs aren't bullet sponges and (2) T-29 drums don't permit that. Despite that minions and acid / resin turrets have become bullet sponges, nothing has changed for the T-29 that have to use its ammo against the increased number of xenomorph targets.

With this PR, a T-29 smartgunner can choose to get MORE ammo from getting the extended drum. If a T-29 smartgunner uses all their SG points to purchase three extended drums, they will get a total of 1250 since 350 * 3 is 1050 and T-29 defaults with 200 in weapon. However, it takes a long time to put extended mag into T-29. For comparsion, to reload a T-29 with extended drum takes 2.5 SECONDS whereby to reload a T-29 with standard drum takes 1.75 SECONDS. I wouldn't be surprised if T-29 smartgunners would op for the extended drums.

Also, you cannot put extended drum in G8 pouch nor the Jaegar general mod. You can only put extended drum in backpack, satchel, and combat backpack. MJP said so, though I can put it as large if my arm gets twisted enough.

I also delete the 4 SG drums in surplus since those 4 SG drums help early T-29 smartgunners more than late T-29 smartgunners. For context, if a T-29 smartgunner that hog all the SG drums in early game, they will have a total of 1,600 ammo, leaving the rest of T-29 smartgunners to have 800 ammo assuming they don't buy more drums. The 1,250 ammo is enough for T-29 smartgunners to get req points and continue to engage while the order is being process and deliever.

## TL;DR

T-29 extended drums that hold 350 ammo but takes a long time to reload compared to the standard drums that hold 200 ammo but take a smaller time to reload

Loadout for T-29 will look like this

T-29 with 3 standard drums = 800
T-29 with 2 standard drums and 1 extended drum = 950
T-29 with 1 standard drum and 2 extended drums = 1100
T-29 with 3 extended drums = 1250

## Changelog
:cl:
add: T-29 extended drums that hold 350 ammo but takes a long time to reload compared to the standard drums that hold 200 ammo but take a smaller time to reload
balance: T-29 gets more ammo if deck out with extended drums
/:cl:


BraveMole:
SG is not and should not be marine+. Ammo being important for the sg is what makes it an interesting role; t25 being t12+ is no reason to powercreep t29 to be t60+

PR #9124, opened 2022-01-01
Big Red LZ 2 Fix

novaepee:
## About The Pull Request

I made a mistake when mapping. Alamo now clips through wall without fix.

## Why It's Good For The Game

My bad. Attention to detail be like

## Changelog
:cl:
fix: Big Red LZ2 NOW doesn't let Alamo clip through wall.
/:cl:


novaepee:
The mistake:

image

The result:

image

PR #9134, opened 2022-01-02
Soft armor, including Jaegar storage and helmet modules, don't give 100 bio armor

### LINKED ISSUE #9130: Storage module has 100 bio armor making sentinel even more useless
### Description

no one noticed since smoke neuro would give them a bit of neuro

### Test Merges

doesn't matter

### Reproduction Steps

1. Get storage module
2. Look at armor value for bio before attaching it
3. Look at value afterwards

### Screenshots

https://user-images.githubusercontent.com/24631139/147861557-2da9a6e4-cf50-41de-a700-592e19cf4068.mp4

image
image

novaepee:
## About The Pull Request

Title. Fixes #9130. Adding ONE LINE CODE gets the fix.

~~I'm still not sure how modules give more bio armor. That is a mystery for another day, so this is just a fix.~~ The mystery is solved. #4096 points to such reason. Also, thanks InterroLouis for the feedback.

## Why It's Good For The Game

How storage module and helmet modules give 100 bio armor is insane, not to mention on top of M.II Mimir. No more tanking acid shots or neurotoxin. You take it like everyone else.

Cope and seeth, M.II Mimir users.

I have a strong sense that this will make sent pretty oppressive.

## Changelog
:cl:
fix: storage modules don't give more armor as intended. Now you can't have 200+ or possibly 300+ bio armor to spawn
balance: marines don't stack bio armor like no tomorrow, so expect the sentinel castes to give you a harder time 
/:cl:

### DIFF
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -38,7 +38,7 @@
 		soft_armor = getArmor(arglist(soft_armor))
 	else if (!soft_armor)
 		// Default bio armor 100 to avoid sentinels getting free damage on sent
-		soft_armor = getArmor(bio = 100)
+		soft_armor = getArmor(bio = 0) // if bio = 100, marines can have at least 150 bio armor, and with M.II Mimir, 190 bio armor. This permits them to tank acid shots and neurotoxin, which was not part of the game when this was made.
 	else if (!istype(soft_armor, /datum/armor))
 		stack_trace("Invalid type [soft_armor.type] found in .soft_armor during /obj Initialize()")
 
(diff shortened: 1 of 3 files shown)


novaepee:
image

As the Queen Mother intended. I was not paid by xenomorph players to make this PR.

InterroLouis:
The issue lies in code/game/objects/objs.dm actually, and is caused by the bit of code starting at line 35. You'll need to do the same solution for every single object that attaches to uniforms, armor, or helmets that doesn't have its own defined armor values. So webbings basically. Hell, you even have to do this for the jaeger VISOR which gets 100 bio armor which it gives to the helmet from that line of code.

BraveMole (approved):

PR #9159, opened 2022-01-04
Removes neuro sting from queen, shrike and drone

Pariah919:
## About The Pull Request
Shrike, Queen and Drone don't need the powercreep of getting neurosting, and probably shouldn't have had larva sting in the first place because it was gradual powercreep from defilers kit getting spread around.

## Alternatives
Rework neuros purging properties into a different chemical entirely as neurotoxin is incredibly powercreeped.

## Why It's Good For The Game
Queen shouldn't have direct combat abilities in the first place as it's **literally** part of the chess king design that queen now has, and giving them a direct combat ability literally overrides that.
Shrike doesn't need neurotoxin stun fun button, it has two instant stuns already, who even thought this was a good idea? It didn't need compensation.
Drone didn't need compensation for losing larval sting and encourging and buffing combat drone is against it's caste design in the first place.

## Changelog
:cl:
balance: Removes neurosting from queen, shrike and drone
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/drone/castedatum_drone.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/drone/castedatum_drone.dm
@@ -56,7 +56,6 @@
 		/datum/action/xeno_action/activable/psychic_cure/acidic_salve,
 		/datum/action/xeno_action/activable/transfer_plasma/drone,
 		/datum/action/xeno_action/activable/corrosive_acid/drone,
-		/datum/action/xeno_action/activable/neurotox_sting,
 		/datum/action/xeno_action/create_jelly/slow,
 		/datum/action/xeno_action/toggle_pheromones,
 	)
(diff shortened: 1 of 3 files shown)


kirkiness:
Neuro sting injects really small amount of neuro, only 15u that's first stage of poisoning. I think these castes need it to give them chance against premedded and b18/vali users. There is not always trusty sentinel to deliver Neuro and these 15u only stop stamina regen and purge some meds, they aren't downing unga like sent's 50+ units and barely give any pain

BraveMole (changes requested):
Reducing the amount of neuro injected per sting seems a better solution IMO. Maybe only 2-3 instead of 5 like currently?

But keep it 5 for sentinel, since it's the caste which has the hardest time stinging anything.

Sting is really important to counter B18/vali if you don't have the proper castes

Losenis:
Yeeeeeep called it, fix it until it's broken and then remove it.

Jeff-jefferson-jeffery:
I have risen from the grave to inform you that neurotoxin is literal cancer and every time it's buffed, a baby seal is clubbed to death by a shitcoder.

BraveMole:
We are going to take your other idea instead

PR #9177, opened 2022-01-05
Organ damage toxin threshold

Sebbe9123:
## About The Pull Request

This PR changes the minimum required toxin present in marine body for organ damage to start happen from 60 down to 40.

## Why It's Good For The Game

Toxin damage is somewhat hard for Xenos to deal to marines, while marines have plenty of ways to heal self inflicted toxin damage. 40 is still high but any lower I feel like will be too much.

## Changelog
:cl: Fuyu
balance: Changes minimum toxin damage for organ damage from 60 to 40
/:cl:


SpaceLoveSs13:
I believe this issue is not true to the extent you presented here. A more fair amount would be 50 imo.

Sebbe9123:
> I believe this issue is not true to the extent you presented here. A more fair amount would be 50 imo.

I understand your concern, but tox damage is healed by a lot naturally unless the organ is actually damaged. Puking and just life ticks heal tox by like 2-3-4 damage ish per heal proc. Getting a marine halfway to crit with tox only damage is super hard on it's own. I feel like this doesn't really do anything on it's own, it needs something else as well but right now this change will allow at least defilers to MAYBE do a tiny bit of organ damage if they team up with spitter castes (and not kill the marine first)

BraveMole:
I don't see how this improve the game

PR #9178, opened 2022-01-05
Protect ai nodes and pipe from tadpole crushing

BraveMole:
## About The Pull Request

They should not be qdeled

Also add tadpole to debuglabus, because no one cares about normandy

## Changelog
:cl:
fix: Protect ai nodes and pipe from tadpole crushing
/:cl:


code/_compile_options.dm
```diff
@@ -1,4 +1,4 @@
-//#define TESTING //By using the testing("message") proc you can create debug-feedback for people with this
+#define TESTING //By using the testing("message") proc you can create debug-feedback for people with this
```
BraveMole, line 1:
```suggestion
//#define TESTING //By using the testing("message") proc you can create debug-feedback for people with this
```

PR #9180, opened 2022-01-05
Removes bumpattack from vali related tools

Pariah919:
## About The Pull Request
Vali related tools lose bump, spear and machete.

## Why It's Good For The Game
Vali bump is honestly not that great of an idea, spear atleast, shouldn't have it. I might just only remove spear bump. But I feel like marine bump other than machete is a bad idea, it gives machete some level of reason to be used over the spear or machete or vali spear etc.. 
I feel like spear should lose it, but I think Vali melee in general shouldn't have bump. I like the idea of a high risk high difficulty playstyle that vali should have.

## Changelog
:cl:
balance: vali tools like spear and sword no longer have bump attack.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -67,6 +67,14 @@
 	. = ..()
 	AddComponent(/datum/component/harvester)
 
+/obj/item/weapon/claymore/harvester/equipped(mob/user, slot)
+	. = ..()
+	toggle_item_bump_attack(user, TRUE)
+
+/obj/item/weapon/claymore/harvester/dropped(mob/user)
+	. = ..()
+	toggle_item_bump_attack(user, FALSE)
+
 /obj/item/weapon/claymore/harvester/get_mechanics_info()
 	. = ..()
 	. += jointext(codex_info, "<br>")
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -67,14 +67,6 @@
 	. = ..()
 	AddComponent(/datum/component/harvester)
 
-/obj/item/weapon/claymore/harvester/equipped(mob/user, slot)
-	. = ..()
-	toggle_item_bump_attack(user, TRUE)
-
-/obj/item/weapon/claymore/harvester/dropped(mob/user)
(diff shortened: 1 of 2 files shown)


Waselon:
- Spear has same 60 damage while wielded
- While wielding spear you cannot use the shield, leaving you more vulnerable
- If using spear and shield, spear would deal 32 damage onehanded

Rayfod:
I'm pretty sure that the harvester spear having a bump attack is a bug that comes from borrowing some code or something from the harvester blade.

The Vali sword already has a decrease in damage. While the harvester spear also has the same nerf to counter the fact that they are part of the harvest series. 
So I do not see why the harvester blades should be nerfed with removing the bump attack as well.

These are the stats for reference.
75 damage on a normal spear(Which do not have a bump attack from what I know) vs. the 60 damage for the harvest spear
The same stat goes for the two swords.

Verlurk:
> I'm pretty sure that the harvester spear having a bump attack is a bug that comes from borrowing some code or something from the harvester blade.

If that's a bug, shouldn't it have been fixed prior to it being merged?
 
> The Vali sword already has a decrease in damage. While the harvester spear also has the same nerf to counter the fact that they are part of the harvest series. So I do not see why the harvester blades should be nerfed with removing the bump attack as well.

Decreases in damage don't matter when you cause more damage overall via instantly applying fire stacks which disable pheremones and cause DOT, or cause a massive slow which allows you to land more consecutive hits which also slow.

> These are the stats for reference. 75 damage on a normal spear(Which do not have a bump attack from what I know) vs. the 60 damage for the harvest spear The same stat goes for the two swords.

See above.

Waselon:
> I'm pretty sure that the harvester spear having a bump attack is a bug that comes from borrowing some code or something from the harvester blade.

its not a bug

bluezorua:
bump attacks just make combat less of an arthritis inducing clickfest, removing them would just make playing melee a pain in the ass and finger joints

BraveMole:
Bump slash is mostly QOL and makes melee enjoyable

BraveMole (changes requested):
like said in discord you can remove from spear

BraveMole (approved):

PR #9191, opened 2022-01-06
Buffs Overcharge Lasers

KrabSpider:
## About The Pull Request
Buffs Overcharge lasers to 50 from 40 damage.
Decreases fire delay to 0.40 from 0.45

## Why It's Good For The Game
40 damage every 0.45 seconds with no status effects isn't very useful, especially when you can get the job done by simply staying on default because the **heavy duty mode** is quite literally the equivalent of two shots fired from said default.
 

## Changelog
:cl:
balance: Laser Rifle Overcharge RoF increased from 0.45 to 0.40. 
balance: Increases damage of overcharge lasers from 40 to 50.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1513,7 +1513,7 @@ datum/ammo/bullet/revolver/tp44
 	name = "overcharged laser bolt"
 	icon_state = "overchargedlaser"
 	hud_state = "laser_sniper"
-	damage = 40
+	damage = 50
 	penetration = 20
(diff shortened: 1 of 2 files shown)


Lumipharon:
It double AP and sunder, not just damage. With your PR standard would be simply inferior.

KrabSpider:
I am fine with -

> It double AP and sunder, not just damage. With your PR standard would be simply inferior.

Fixed.

BraveMole (changes requested):
same, a comparison of dps with other type of guns would be interesting

PR #9207, opened 2022-01-08
Add teleporters pad to req

BraveMole:
## About The Pull Request

They can be bought in req, in pairs. They are prelinked, and if one is destroyed, the other explode. 
You can move crates, machinery and marines from one teleporter to the other, if it's on the same z level.

Needs to be anchored and the area needs to be powered OR it can have a cell for internal power.

## Why It's Good For The Game

It's quite hard to setup so this shouldn't be too op. This makes forward bases stronger since you can more easily refurnish/retreat from those places

## Changelog
:cl:
add: Add teleporters pad to req. You buy them in pairs, and they can teleport marines/equipment on the same z level. Requires to be anchored, and have power (either area power OR internal cell power).
/:cl:


EvilMelonTheSecond:
Now this is something that can connect shipside roles to groundside. You could already do this with medevacs but now its properly in the game. I expect cool results from this.

Losenis:
>Now this is something that can connect shipside roles to groundside

He did say same z-level so this is mostly going to be linking FOB and forward bases rather than the actual ship.

BraveMole:
I thought of allowing z level transfers but i think this would be to game breaking

EvilMelonTheSecond:
Oh, then this is almost obsolete. Marine gameplay rarely allows frontline outposts let alone a powered one. Even if marines established one they would still have to keep the push and leave it undefended. Carrying ammo with a crate is more useful than this.

Losenis:
Counterpoint, tadpole

EvilMelonTheSecond:
> Counterpoint, tadpole

Make it a tadpole upgrade like the HMG and operating table then. 50 points is just meh for a tadpole only upgrade

Losenis:
>Make it a tadpole upgrade

Why?

EvilMelonTheSecond:
> > Make it a tadpole upgrade
> 
> Why?

It needs power, marines almost never stay in places with power (most of the time because there is no power in caves/open areas). Previously floodlights were changed to use a battery instead of area power by Bravemole for that exact reason. The game wants marines to push to kill silo/spawners which are 100% in caves, meaning setting these up either needs unwilling manpower or even more points and time for sentries and a big cadeline to keep it defended.

Losenis:
That still is no reason to make it a module that competes with two literally free modules found at roundstart. Its already limited use would be limited even more down to uselessness if you can't even use groundside power to have it be a teleporter from LZ1 to gens in big red for example, and stopping engis from using it unless someone else who'd use tadpole also comes along.

Tristan63:
make it 60 points and turn it into portal gun

Lumipharon:
Can this be powered directly from the grid (i.e. a generator on the tad etc), or apc power only? Either way looks FUN.

EvilMelonTheSecond:
The added cell powered function changes a lot of things. This can now connect FOB to forward bases properly in really far pushes, which gives marines a reason to bunker down near frontlines. A dedicated engineer can now alleviate the attrition of giant maps like Big Red or Ice Colony. Finally an engi tool to help pushes.

code/game/objects/structures/teleporter.dm
```diff
@@ -0,0 +1,146 @@
...
+		/obj/machinery,
+	))
+	///The linked teleporter
+	var/obj/machinery/teleporter/linked_teleporter
+	///The optional cell to power the teleporter if off the grid
+	var/obj/item/cell/cell
+
+/obj/machinery/teleporter/proc/set_linked_teleporter(obj/machinery/teleporter/linked_teleporter)
+	if(src.linked_teleporter)
+		CRASH("A teleporter was linked with another teleporter eve though it already has a twin!")
```
hyper2snyper, line 33:
```suggestion
		CRASH("A teleporter was linked with another teleporter even though it already has a twin!")
```

code/game/objects/structures/teleporter.dm
```diff
@@ -0,0 +1,146 @@
...
+	update_icon()
+
+/obj/machinery/teleporter/attackby(obj/item/I, mob/user, params)
+	if(!ishuman(user))
+		return FALSE
+	if(!istype(I, /obj/item/cell))
+		return FALSE
+	if(cell)
+		to_chat(user , span_warning("There is already a cell inside, use a crowbar to remove it."))
+		return
```
hyper2snyper, line 133: 'return FALSE' ?

hyper2snyper (changes requested):
looks good. TM after this

code/game/objects/structures/teleporter.dm
```diff
@@ -0,0 +1,146 @@
...
```
BraveMole, line 133:
```suggestion
		return FALSE
```

code/game/objects/structures/teleporter.dm
```diff
@@ -0,0 +1,146 @@
...
+		to_chat(user, span_warning("A red light flashes on the [src]. It seems it doesn't have enough power."))
+		playsound(loc,'sound/machines/buzz-two.ogg', 25, FALSE)
+		return
+
+	if(!COOLDOWN_CHECK(src, teleport_cooldown))
+		to_chat(user, span_warning("The [src] is still recharging! It will be ready in [round(COOLDOWN_TIMELEFT(src, teleport_cooldown) / 10)] seconds."))
+		return
+	
+	if(!linked_teleporter)
+		to_chat(user, span_warning("The [src] is not linked to any other teleporter"))
```
XSlayer300, line 64:
```suggestion
		to_chat(user, span_warning("The [src] is not linked to any other teleporter."))
```

code/game/objects/structures/teleporter.dm
```diff
@@ -0,0 +1,146 @@
...
+		return
+
+	if(linked_teleporter.z != z)
+		to_chat(user, span_warning("The [src] and the [linked_teleporter] are too far appart"))
```
XSlayer300, line 68:
```suggestion
		to_chat(user, span_warning("The [src] and the [linked_teleporter] are too far apart!"))
```

FosterGrime-Heymynameslouie:
Teleporter pads aren't linked if one is using room power and the other is using cell power.
Teleporter linker has no sprite.
Teleporter Linker cant be picked up.

hyper2snyper:
this ready?

BraveMole:
Yes

PR #9208, opened 2022-01-08
Up the maximum recommend version

BraveMole:
1571 works just fine


BraveMole:
Self merging cause hyper is busy and this doesn't impact anything

PR #9266, opened 2022-01-13
Pain update 

BraveMole:
## About The Pull Request

Change the pain system a bit. Rather than using a switch and ramping up pain, traumatic_shock is converted into shock_stage at a rate linear with the difference between shock stage and traumatic shock.

This keeps overall the same idea that jpr had with pain, while being easier to balance out (and to understand)

I also try to make this process more reactive (aka you gain shock stage fast, but you can heal from it fast), but numbers are hard and this will need further balancing

## Why It's Good For The Game

Make pain system easier to understand. Shock is stacking faster but also disappearing faster

## Changelog
:cl:
balance: Shock is stacking faster but also disappearing faster
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -12,6 +12,8 @@
 #define PAIN_REDUCTION_HEAVY -35 //paracetamol
 #define PAIN_REDUCTION_VERY_HEAVY -50 //tramadol
 
+#define PAIN_REACTIVITY 0.15
+
 
 //Nutrition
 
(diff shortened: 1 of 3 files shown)


jroinc:
This appears to effectively remove any possibility of hitting paincrit, and more-or-less remove the pain system even more than it already is, unless you're sitting near 150 damage or so for an excessively long time.

As shock_stage can no longer increase above the amount of traumatic_shock (and can only increase by 15% of the difference between traumatic_shock and shock_state per tick, according to my read of the code), shock rise will be much slower, while fall will be automatic as pain goes down, even if it's a gradual decrease.

I *may* have an incorrect read of this, and I know it'll up the amount of people at lower shock_states rather than 0, but on the whole, this appears to be a significant nerf of pain. TM it and do some VV checks in combat, I guess.

PR #9279, opened 2022-01-14
MR-25 changes- Scatter reduced massively and UGL added.

Pariah919:
## About The Pull Request
MR-25 has old scatter assumptions, rather than new scatter assumptions, it has 20 scatter, higher than any other weapon in the game, which is kinda absurd if you think about it though, really other than that I think it's overall fine because it has rifle level fire rate and the ability to mount a UGL can give you some utility.

## Why It's Good For The Game
Makes the MR-25 not awful garbage, and makes it useable.

## Changelog
:cl:
balance: MR-25 can mount a UGL, and massive scatter reduction.
/:cl:


BraveMole:
We're in the middle of a scatter rebalance, wait for it to be merged/denied

PR #9281, opened 2022-01-14
T90 rebalance

Lumipharon:
Buffs T90 wielded performance
Nerfs T90 akimbo performance and slightly increase one hand scatter

## About The Pull Request
Buffs the T90's wield time to 0.3 from 0.5 and wield speed penalty to 0.2 from 0.25. These are the same values as the laser carbine.
Reduces the effectiveness of the T90 when used one handed by increased scatter to 30 from 20, and adding 0.2 akimbo additionally delay, on MJP's advice.

This is PR 2/2, with the MR25 readded in a seperate PR to be a less DPS heavy one handed option.

## Why It's Good For The Game

T90 akimbo spam is pretty gross, as it has extremely low scatter and high DPS. Makes it a more mobile and useful weapon when wielded, comparable to the laser carbine, where it trades off higher DPS and capacity compared to the carbine's ability to use scatter shot and have underbarrel attachments.

## Changelog
:cl:
balance: T90 is better wielded, worse one handed and akimbo.

/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smgs.dm
+++ b/code/modules/projectiles/guns/smgs.dm
@@ -83,7 +83,7 @@
 	caliber = CALIBER_10X20_CASELESS //codex
 	max_shells = 50 //codex
 	flags_equip_slot = ITEM_SLOT_BACK
-	wield_delay = 0.5 SECONDS
+	wield_delay = 0.3 SECONDS
 	force = 20
 	type_of_casings = null
 	default_ammo_type = /obj/item/ammo_magazine/smg/standard_smg
@@ -114,9 +114,11 @@
 	accuracy_mult_unwielded = 0.9
 	scatter = -2
 	fire_delay = 0.15 SECONDS
-	scatter_unwielded = 20
-	aim_slowdown = 0.25
+	scatter_unwielded = 30
+	aim_slowdown = 0.2
 	burst_amount = 0
+	upper_akimbo_accuracy = 5
+	lower_akimbo_accuracy = 3
 
 	placed_overlay_iconstate = "t90"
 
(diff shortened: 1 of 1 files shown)


turnop:
A gun that fires such an insane amount of bullets and lets you move as fast as the T90 getting slightly higher scatter when wielded akimbo in exchange for becoming an even more meta weapon wielded with insanely fast speed is not a nerf. Akimbo will still be viable and extremely good with these changes, all this PR will acomplish is making the t90 a meta weapon for onehanded use as well

Lumipharon:
> A gun that fires such an insane amount of bullets and lets you move as fast as the T90 getting slightly higher scatter when wielded akimbo in exchange for becoming an even more meta weapon when wielded twohanded is neither nerf nor rebalance. T90 Akimbo will still be not just viable, but still extremely good with these changes. I don't see the T90's akimbo capabilitiy to take a hunter from full health to critical in under a second in any sort of tight passage being remotely affected by this.
> 
> all this PR will acomplish is making the t90 a meta weapon for twohanded use as well. it will, however, make the less optimal riot shield and t90 gameplay impossible. so it doesn't really attack the actual problem but the changes will make non-optimal playstyles with the weapon unviable

Have you tried akimbo weapons with the higher akimbo scatter min and max? They have a huge amount of extra scatter. So yes, T90 akimbo will still meme a xeno that is point blank range, but 4+ tiles away a significant % of your bullets will miss wildly. It's a fairly large nerf.
Using the T90 one handed on it's own is barely affected. You'll have 15 scatter instead of 5, so you'll still hit almost everything at 7 tiles, but my other PR to bring back MR25(may be RIP) is intended to be a more sensible one handed (non akimbo) option. Gyro makes running one handed shooting a lot better, and the recent fix by Splinter has unfucked it's scatter.

For 2 handed use, I would still personally use the laser carbine over the T90 every day. You exchange 10 bullets for your choice of attachment weapon, and hitscan. This PR just makes the T90 not 100% worse than the carbine.

Lumipharon:
Closing, will maybe add a new smg at some point with wield use in mind, instead of changing around the T90.

PR #9293, opened 2022-01-15
Primos can be bought by tiers

BraveMole:
## About The Pull Request

Rather than selecting which primo you want to buy, you have to buy the whole tier. You also have to buy the first tier, then the second, etc...

## Why It's Good For The Game

Choosing the primos would lead certain castes to never have their primo, and maybe even to fall into usellness. This makes primo a coherent ramp up for the hives

## Changelog
:cl:
add: Primos can be bought by tiers
/:cl:


novaepee:
Way better. Now queen can't just buy funny primo queen upgrade in the first seconds of the round.

Mechanelize:
hand over the primos

novaepee:
Some players are concerned that it will take too long to get primo T3 in light of this PR.

Naaanii:
I think it'd be better to keep it caste only but make it so lower tiers are cheaper and higher tiers are more expensive since they're more likely to be more impactful. It'll probably also lead to a gradual increase in power amongst Xenos depending on what castes they currently have rather than waves of increasing strength with Queen and King (if King has any primo) being last.

I'd also restrict primordial purchasing to Queen and Shrike only imo since it's easy to decide and purchase and so no random hivelord will just buy it themselves.

Pariah919:
Runner is very funny, but shouldn't probably be in the game, I can't see a case for it where it isn't literally awful to play against once the hilarity of it runs off, probably better to run with runner having no primodial until someone comes up with something.

Also feel like getting T1s is far too easy, I see them consistently within 12:30, seems strange. They should start coming into play by 12:45 or so.

PR #9295, opened 2022-01-15
Carrier primo

BraveMole:
Carrier primo, you can put up a hugger turret every 5 minutes. Fire rate of the turret is 10 sec, range is 5 and it will throw whatever huggers the carrier was selecting when building


vilereaver41:
Does this have a distance mechanic so you dont got 5 close together? 5 minutes also seems a bit to long

Surrealaser:
Personally I think you should be able to toss this thing into position, and it shouldn't require weeds.

jroinc:
This seems very weak to me. One hugger every ten seconds, with a 5-tile range is a very low rate of fire, and given as they don't stun, will do minimal damage. Also, huggers intersect with bullets, so will be shot out of the air.

I see either a GPMG unga showing up and deleting it while getting hit by like... one hugger corpse if he even bothers to go within range, or an unga showing up, getting hugged, then deleting it because he's now immune before anyone else shows up since huggerstun is so short.

I guess test it and find out.

PR #9301, opened 2022-01-16
Nerf sandbag integrity

novaepee:
## About The Pull Request

From 300 to 200.

## Why It's Good For The Game

#5758. Metal barricade has 200 health initially, then with barbed wires 250 health. Apparently, even with MJP's nerf, sandbag is still stronger than metal barricade. 

I should also make a PR to improve upgraded metal barricade so that marines actually use it.

## Changelog
:cl:
balance: nerf sandbag health from 300 to 200; now sandbag is not stronger than barred metal barricade
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -1083,7 +1083,7 @@
 	name = "sandbag barricade"
 	desc = "A bunch of bags filled with sand, stacked into a small wall. Surprisingly sturdy, albeit labour intensive to set up. Trusted to do the job since 1914."
 	icon_state = "sandbag_0"
-	max_integrity = 300
+	max_integrity = 200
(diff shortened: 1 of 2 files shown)


InterroLouis:
Metal barricades also have the ability to be repaired without having to use additional materials, if the damage doesn't get to critical levels. Sandbag barricades ALWAYS need more sandbags to be repaired.

Lumipharon:
As above. Sandbags are better because they require more resources to repair, and deploy faster because the have prep time to actually fill them before hand.

EvilMelonTheSecond:
Sandbags HP nerf from 400 to 300 instead of 200 was intentional. If sandbags get nerfed down to metal cade levels engineers won't ever use them because of reasons stated above. 

If you want to make them into a non-engi cade tool you gotta make them cheaper and easier to carry as well as do something for the fact that most non-engis won't have metal for barbed wire.

BraveMole:
Nah, sandbags are fine the way they are imo.

Derrick2Electricboogaloo:
Yeah no. Please god no. You have to fill them, marines already barely do that, this will just encourage them to stop.

PR #9322, opened 2022-01-18
Adds population dependent modular map support

RipGrayson:
## About The Pull Request

I've been needing to do this for a bit, essentially allows for mappers to define min and max player values needed for a modular mapping insert to be in "rotation" for a given map.
E.g I could define a LV dome to only show up with a population of between 50 and 100. 

## Why It's Good For The Game

This sort of thing has been needed for quite a while and the potential is fairly limitless. We could theoretically define places in big maps like Chigusa and Magmoor that shrink on lower pop and grow at higher pop, allowing for much more flexible maps.

## Changelog
:cl:
code: Added population dependent mapping insert support.
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -32,6 +32,8 @@ SUBSYSTEM_DEF(mapping)
 	var/groundmap_voted = FALSE
 	///If true, non-admin players will not be able to initiate a vote to change shipmap
 	var/shipmap_voted = FALSE
+	//The number of connected clients for the previous round
+	var/last_round_player_count
 
 //dlete dis once #39770 is resolved
 /datum/controller/subsystem/mapping/proc/HACK_LoadMapConfig()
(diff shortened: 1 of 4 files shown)


code/modules/mapping/modular_mapping.dm
```diff
@@ -9,6 +9,9 @@
 	var/template_width = 0
 	///Bool for whether we want to to be spawning from the middle or to the topright of the spawner (true is centered)
 	var/keepcentered = FALSE
+	var/min_player_num //minimum player number for a modular to be added to the pool of potential modular map spawns.
```
hyper2snyper, line 12: autodoc these two like the rest.

hyper2snyper (approved):
this is pretty cool

Lumipharon:
Not directly related to the pop based mechanic, but does this functionality mean you could in theory go completely nuts and have a map with a (relative) degree of randomisation? Like shifting buildings around and that sort of thing?

RipGrayson:
Had a talk with Mole, this isn't going to work in any real world situation. Please don't merge it while I work on a better solution.

code/controllers/subsystem/mapping.dm
```diff
@@ -32,6 +32,8 @@ SUBSYSTEM_DEF(mapping)
 	var/groundmap_voted = FALSE
 	///If true, non-admin players will not be able to initiate a vote to change shipmap
 	var/shipmap_voted = FALSE
+	//The number of connected clients for the previous round
```
BraveMole, line 35:
```suggestion
	///The number of connected clients for the previous round
```

BraveMole (changes requested):
How did you test this?

RipGrayson:
> How did you test this?

The first iteration was tested with a modified version of Big Red that has 5 debug modular inserts that I would alter the min and max population for, along with a debug message that told me when an insert had entered the map pool and which insert it was.

The current version largely used the same system, except I could alter the .json file to report a larger number of players allowing for more extensive testing of higher population. It wasn't as extensively tested as the first version, because the core of the map selection process is the same, it's just pulling the player number from a stored .json file from the last round instead of checking for currently connected clients.

PR #9326, opened 2022-01-18
Defiler Hemodile and Transvitox gas is no longer transparent.

vilereaver41:
## About The Pull Request

Changes Transvitox and hemodile to non transparent.

## Why It's Good For The Game

Makes both now viable for use rather then heavily risk death in a group combat scenario because of its opacity

## Changelog
:cl:
balance: opacity for transvitox and hemodile no longer 60 but 255
/:cl:

1
image
2
image


EvilMelonTheSecond:
I agree that they are in need of a buff but how about an effect amplification inside gas cloud instead of just another smoke screen?

It's fine if this is the chosen solution but you could easily give these chem on_mob_add effects so clouds become REALLY dangerous but transparent.

jroinc:
It's still transparent in that you can see through a cloud to the other side. It's just dense. You need to alter another thing than just the cloud opacity, unless that was the intent..

novaepee:
Make defiler see marines, other xenomorphs, and itself. Gas master eyes see all.

BraveMole (approved):

PR #9328, opened 2022-01-18
Miner Upgrades in Operational Supply And Supply Order, Not Autolathe

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

Autolathe was a placeholder for miner upgrades that function should have been just that, a placeholder. In today's meta, autolathe doesn't make sense.

Now it's simply a skill issue for marines to NOT get miner upgrades. Also, limited to like 1 upgrades to make marines think on which miners to upgrade.

Marines can buy more upgrades in req. Spend more money, get more money.

## Changelog
:cl:
add: All miner upgrades are now in Operational Supply
add: You can get more miner upgrades in supply order
del: No more being able to produce infinite miner upgrades in autolathe
qol: miner upgrades in Operational Supply
/:cl:


Mrrpip:
B-but where will i be able to get mining modules on crash from?

EvilMelonTheSecond:
Making overclock and reinforcement 15 points is practically removing them. Nothing they offer competes with the utility of auto upgrade.

Three333Digit:
This is bad. The use of these autolathe miner modules are very situational. Engineers not printing miner modules is a resource issue, not a skill issue - they have to think about their starting vendor points when considering miner modules. These autolathe miner upgrades will never see use again.

BraveMole (changes requested):
too expensive

BraveMole (approved):

PR #9330, opened 2022-01-18
CIC roles get free HealthMate HUD spawn in backpack

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

HealthMate HUD for CIC is actually pretty important to see how marines are doing. All good players in CIC have a HealthMate HUD, and when they spawn late, they don't get HealthMate HUD since marines have taken the free HealthMate HUD. In CM, CIC gets free HealthMate HUD, so we should do the same since CIC really needs it.

Also, I'm not spawning it in map since I know marines will raid CIC for them when there are no CIC, so to deal with that, only CIC roles spawn with HealthMate HUD.

I know the powergamers will either give this away for free, that's a given. What is worrisome is deployable SO with free HealthMate HUD, so admins, keep an eye out.

## Changelog
:cl:
add: CIC roles get free HealthMate HUD spawn in backpack
qol: CIC roles get free HealthMate HUD spawn in backpack for better Overwatch
admins: policy regarding planetside SO should get check over in light of SO getting free HealthMate HUD; I forsee gamers using deployable SO more frequent with free HealthMate HUD
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -78,6 +78,9 @@ Godspeed, captain! And remember, you are not above the law."})
 	jobtype = /datum/job/terragov/command/captain/rebel
 	ears = /obj/item/radio/headset/mainship/mcom/rebel
 
+/datum/outfit/job/command/captain/post_equip(mob/living/carbon/human/H, visualsOnly = FALSE)
+	. = ..()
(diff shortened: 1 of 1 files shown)


BraveMole (changes requested):
why in backpack?
and anyway, they don't need those. If you want cic to be able to see health through overwatch, then you do that directly in  overwatch code

novaepee:
>  If you want cic to be able to see health through overwatch, then you do that directly in overwatch code

time to check overwatch code

PR #9336, opened 2022-01-19
Removes old/no longer needed boiler fire windup.

vilereaver41:
## About The Pull Request

Boiler does not need so many restrictions for an ability that has a long cooldown + 2 seconds to deploy, this makes firing instant with a click.

## Why It's Good For The Game

This is no longer needed due to current metas, it also gives boilers quicker play and more use rather then having to cancel 90% of the time right when you are about to finish to flee.

## Changelog
:cl:
balance: bombard windup from 2 to 0
/:cl:

Notes: current winrates as of making this PR 
image
 Xenos either need a bit of-power creep or marine weaponry needs to be nerfed.


SpaceLoveSs13:
Don't see how this is not needed just because the winrate has gone down. The winrate also shows 14 days of data but a easy search for 30day winrate will tell you it's perfectly balanced which is 44% xeno to Marines.

PR #9353, opened 2022-01-21
Fix tools dependencies

BraveMole:
pillow ver 9 is needed, need to bump python version


BraveMole:
Need this and it's a tool

PR #9372, opened 2022-01-22
Tgui minimap

BraveMole:
## About The Pull Request

Replace the subsytem minimap by a tgui one. The idea is allowing it to be open at all time, at the right of the screen (probably above chat)


Mechanelize:
laggy, doesnt show full map, minimap button is on the far right not far left now,

PR #9387, opened 2022-01-24
Swaps Prae and Defiler Pheromone levels, gives primodal defiler strength 5 pheromones.

Pariah919:
## About The Pull Request
Title. Aswell has primodial defiler has strength 5 pheromones, so it isn't the exact same as ancient, which I think is fine once priomidals come out (if ever)
Elder prae gets 2.4 instead of 2.6, though.

## Why It's Good For The Game
Prae's toolbox is too wide and it doesn't sacrifice in any of the places it has a toolkit, it's aurastrength being the best of the T3s is a longterm CM holdover, it could stand to lose something in comparison to other T3s, and defiler needs some other utility past neurogassing people constantly.

## Changelog
:cl:
balance: Prae and Defiler have swapped pheromone levels, Prae caps at 2.6, Defiler at 4.5.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defiler/castedatum_defiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defiler/castedatum_defiler.dm
@@ -58,7 +58,7 @@
 	)
 
 	// *** Pheromones *** //
-	aura_strength = 1.7 //Defilers aura begins at 1.7 and ends at 2.6. It's .1 better than a carrier at ancient.
+	aura_strength = 3 //Defilers aura begins at 3 and ends at 4.5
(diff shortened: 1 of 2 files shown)


Texicos:
And yet it was considered the worst T3 xeno caste for the past few months, and only coming back because a few statics used it. It gets hit by a grenade or fucks up once it dies.

Pariah919:
people really don't like this so I probably gotta think about this more, even if I personally like the idea

ivanmixo:
> And yet it was considered the worst T3 xeno caste for the past few months, and only coming back because a few statics used it. It gets hit by a grenade or fucks up once it dies.

Prae is easily the best T3 what are you on about

PR #9391, opened 2022-01-24
Gorger can now emit pheromones

lKiseki:
## About The Pull Request
Per title. The pheromone strength is quite literally the same as Defiler's.

## Why It's Good For The Game
Being a support caste and currently being a little bit lackluster, I felt this'd help out Gorger a little while fitting its supportive role. It's a slightly outlandish idea that I figured I'd put out there, see how it'd be received.

## Changelog
:cl: Lewdcifer
add: Gorger can now emit pheros, they're as good as Defiler's.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/gorger/castedatum_gorger.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/gorger/castedatum_gorger.dm
@@ -38,6 +38,10 @@
 
 	// *** Defense *** //
 	soft_armor = list("melee" = 5, "bullet" = 5, "laser" = 5, "energy" = 5, "bomb" = XENO_BOMB_RESIST_1, "bio" = 5, "rad" = 5, "fire" = 5, "acid" = 5)
+	
+	// *** Pheromones *** //
+	aura_strength = 1.7 //Starts at 1.7, ends at 2.1, which is just a little bit below Carrier's.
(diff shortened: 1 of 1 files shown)


Walarks:
I think it's better to let it heal moving xenomorphs and maybe increase the range by 1, than to have pheros. Reasoning for this is that I think having easy access to recov/warding pheros can make it better at soloing, while healing improvement itches it more towards better teamwork.

SpaceLoveSs13:
Powercreep. Don't think gorger needs more abilities now. It's getting tankier more than normal xenoes now. Pheroes already have enough buff as it is. adding it to a xeno who ca outheal anything on weeds is going to result in marines never flanking or pushing.

DeltaFire15:
Gorger has some of its heal mechanics interact with regen phero, making them in specific being able to emit phero on the fly quite powerful. It'd effectively allow them to act on their own even more effectively more than boosting their team, and the Queen can already assign them lead if they want them to act as beacon.

Nonetheless this is probably relatively balanced since pheros drain plasma and Gorgy has zero passive regen, making phero a at best temporary use of it.

bowlofcereal:
This is less mechanic and more balance since you're giving a tank the ability to heal even more on it's own and work completely independently of others

lKiseki:
I have a better idea, closing this and discussing it proper with the relevant individuals.

PR #9408, opened 2022-01-25
Xeno upgrade changes

BraveMole:
## About The Pull Request

Some changes to the upgrade system, to make it more intuitive. See CL for list

## Why It's Good For The Game

There was still some bug with the upgrade system, this should squash them. Also, allowing xenos to upgrade even when hurt is full QOL imo

## Changelog
:cl:
balance: Xeno upgrade threshold is now standardised (gorger, wraith and king were completly broken)
balance: Queen get mature quicker, no other changes
qol: Xeno can upgrade even when hurt
balance: It's easier to get primo as a t2 and especially as a t3
balance: When evolving up, you will lose one upgrade (for exemple, from elder T1 to mature T2). When regressing, the opposite happens. This should fix some exploits
balance: You no longer get back full plasma when upgrading
balance: ssd xenos can mature as well
/:cl:


Luxyboo:
i dont think xenos should be able to heal when damage because this prevents them from evoing and devoing shipside if they get tagged even once.

Pariah919:
Why do xeno upgrades need to be massively buffed across the board? Ancient T2/3s should be big buffs for all involved, not something you get within the game starting, which is happening now. It's bothered me for a while.

BraveMole:
> Why do xeno upgrade timers need to be massively buffed across the board? Ancient T2/3s should be big buffs for all involved, not something you get within the game starting, which is happening now. Its bothered me for a while.

they are not at all

bowlofcereal:
This is a good step into allowing xenos to actually play the game and not be held back by a status of maturity that determines whether or not the marine carrying the shotgun instantly kills them or they can survive 2 hits

PR #9411, opened 2022-01-26
Acid turret nerf- Durablity and Starting health, Explosion resistance

Pariah919:
## About The Pull Request
Acid turrets have less Max durablity, Starting health and Explosion resistance. Reduced by 200 hp overall.

Fire damage reduced in kind. If at max durablity, turrets take longer to die from fire, but other sources will kill them.

Heavy explosions one shot them now.

## Why It's Good For The Game
I feel like acid turrets hold chokepoints far too well against ballistic weapons, but nerfing them without nerfing their hardcounter of fire seems too much, so I nerfed fire damage if they're at max health.

They also feel too strong at spawn. I feel like you should place turrets in anticipation, not because the marine horde is nearby.

## Changelog
:cl:
balance: Acid turret health reduced by 200 overall, dies to heavy explosions easier, but dies to fire slower.
/:cl:

### DIFF
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -845,8 +845,8 @@ TUNNEL
 	desc = "A menacing looking construct of resin, it seems to be alive. It fires acid against intruders."
 	bound_width = 32
 	bound_height = 32
-	obj_integrity = 600
-	max_integrity = 1500
+	obj_integrity = 400
+	max_integrity = 1200
 	layer =  ABOVE_MOB_LAYER
 	density = TRUE
 	resistance_flags = UNACIDABLE | DROPSHIP_IMMUNE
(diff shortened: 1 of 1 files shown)


Lumipharon:
Don't nerf turrets man, if ungas are too dumb to deal with something that doesn't die immediately when they click on it then RIP them.

Turrets are a genuine investment due to the points, and are purely a defensive tool.

novaepee:
If reducing health, reduce psychic points cost.

Losenis:
Inability to deal with turrets is literally a skill issue. You can stand by a corner and shoot a tile or two behind the turret to the side to have the trajectory of the bullet still hit the turret while the turret, targetting your tile, will keep hitting a wall to zero effect. Alternatively you can just see the turret at the edge of the screen, step one tile away to have it offscreen and out of range, and just shoot with impunity in that direction till you get confirmation with the acid cloud that it was destroyed. If you don't even want to shoot the turret, just pop in for a single second and throw an incendiary nade and all the fire damage is guaranteed since xenos have zero way to put the fire out, even with acid spray that the turret blocks from its own tile.

Not only that, psy points will be in high demands with the primordials change given that you'll need tons of maturity towers to have T3s even get close to primordial (1750 evo points from zero to ancient, additional 1000 points from ancient to primordial for a total of **_2750_** points).

BraveMole (changes requested):
Yeah i wanna call skill issue for this

PR #9429, opened 2022-01-28
DNRs no longer show up as defibbable on the minimap

Lumipharon:
## About The Pull Request
When ungas go DNR, they currently still show as defibbable on the minimap until their defib timer runs out.

I am a brainlet coder, but this seems to work and it's been an annoying bug forever.

## Why It's Good For The Game
Marines seethe when they risk their lives to try save a dead guy, only to find out when they reach them that they're actually DNR.

## Changelog
:cl:
fix: DNR's no longer show as revivable on the minimap
/:cl:

### DIFF
--- a/code/game/data_huds.dm
+++ b/code/game/data_huds.dm
@@ -275,6 +275,9 @@
 				var/mob/dead/observer/ghost = get_ghost()
 				if(!ghost?.can_reenter_corpse)
 					status_hud.icon_state = "huddead"
+					if(istype(wear_ear, /obj/item/radio/headset/mainship))
+						var/obj/item/radio/headset/mainship/headset = wear_ear
+						headset.set_undefibbable_on_minimap()
 					return TRUE
 			var/stage
 			switch(dead_ticks)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -35,8 +35,13 @@
...
 			if(dead_ticks > TIME_BEFORE_DNR)
 				set_undefibbable()
-			else
-				med_hud_set_status()
+				
+			if(!client)
+				var/mob/dead/observer/ghost = get_ghost()
+				if(!ghost?.can_reenter_corpse)
+					set_undefibbable()
+
```
TiviPlus, line 43: this will set people that log out and back in while defibbable as undefibbable:
ghosts with no client will return null to get ghost

can_reenter_corpse is also already checked in get ghost and is not necessary

TiviPlus (changes requested):

Losenis:
https://www.youtube.com/watch?v=oXJvOJNUlCs

ThelastArchiver:
best pr 2022 up until now

Lumipharon:
Ok so my original method of doing this works, but apparently it may not be the best way of going about it. New commits for a draft alternate way to do the same thing.

Lumipharon:
Now actually works edition. Thanks to Brave for holding my hand.

BraveMole (approved):

TiviPlus (approved):

PR #9439, opened 2022-01-29
Gives Tx11 autoburst mode

Losenis:
## About The Pull Request
Title

## Why It's Good For The Game
It's a gun that has a quick burst that essentially devolves into spamclicking compared to guns like T12 or T18 with a lengthier burst that still have BA. Turns into arthritis simulator quickly enough. Gun can't attach a bfa anyway so it shouldn't be an issue compared to the aforementioned ones. Also """standardizes""" burst performance for everyone by just having everyone able to hold click rather than whoever clicks faster being stronger.

## Changelog
:cl:
qol: Tx11 now has burst auto, less spamclicking needed for the same thing burst offers
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1226,7 +1226,7 @@
 	)
 
 	flags_gun_features = GUN_AMMO_COUNTER
-	gun_firemode_list = list(GUN_FIREMODE_SEMIAUTO, GUN_FIREMODE_BURSTFIRE, GUN_FIREMODE_AUTOMATIC)
+	gun_firemode_list = list(GUN_FIREMODE_SEMIAUTO, GUN_FIREMODE_BURSTFIRE, GUN_FIREMODE_AUTOMATIC, GUN_FIREMODE_AUTOBURST)
 	starting_attachment_types = list(/obj/item/attachable/stock/irremoveable/tx11, /obj/item/attachable/scope/mini/tx11)
 	attachable_offset = list("muzzle_x" = 31, "muzzle_y" = 17,"rail_x" = 8, "rail_y" = 20, "under_x" = 16, "under_y" = 13, "stock_x" = 19, "stock_y" = 23)
 	actions_types = list(/datum/action/item_action/aim_mode)


Texicos:
no

Luxyboo:
Yes.

KrabSpider:
Very good.

BraveMole (changes requested):
Autoburst has specific delays. How does that impact fps

BraveMole (changes requested):
You didn't give the comparison between normal burst spam click and autoburst

Losenis:
Autoburst has an arbitrary 3x fire delay compared to normal firerate as I found out after making the pr so it'd be like 0.75 delay between bursts in autoburst. Given it's an autoburst thing overall rather than the gun itself, I'm no longer interested in this pr as it means spamclicking is still better for all guns capable of burst due to lower firerate on autoburst in comparison. Autoburst is kind of a noobtrap for people that dont codedive.

image

You can test it in valhalla with stuff like t12 and t18 that get a lower firerate on autoburst than spamclicking normal burst.

Losenis:
Just for the sake of clarity:

Tx11 has 0.25 fire delay with 0.05 burst delay. On full auto it's 1 bullet every 0.25 seconds, so a rate of fire of 4 bullets per second.

On burst auto, is has a 0.15 second long burst of 3 bullets with an additional 0.25*3 delay. 0.15+0.75 is 0.9 seconds for 3 bullets, which would be around 3.3333etc rounds per second. Compared to full auto that's bad already because tx11 is the best case scenario (very low burst delay for relatively ok burst count, T18 is worse and T12 even worse than that), but compared to normal burst you only have to see the numbers.

Normal burst doesn't have the *3 penalty. Math ends up as 0.25+0.15 for burst duration + delay before next shot is taken, so 0.4 for 3 rounds. 3/0.4 is 7.5 rounds per second. You end up with a potential 7.5 rounds per second for normal burst mode vs 3.3 rounds per second for autoburst, which is a penalty of _over half_.

This can be interpreted in a lot of ways, most obvious is that normal burst may be a bit broken, but in the current state using burst auto is an actual detriment that only people that either deliberatedly tried normal burst (Like Tx11 that doesn't have burst auto so you HAVE to spam it, and everyone who goes tx11 does so for the burst) or went codediving would know about.

TL;DR tx11 having autoburst is redundant at best, or a nerf to people who don't know about this to use normal burst over autoburst. Found out about this way after making this pr.

BraveMole (approved):

PR #9440, opened 2022-01-29
Wraith laser armor/energyarmor balance changes

vilereaver41:
## About The Pull Request

overdue adjustment of armor for wraith that has been the same since prehitscan, incredibly unfair and imbalanced.

## Why It's Good For The Game

Wraith ancient has a laser and energy armor of just 20, it was fine before hit scan lasers as you had some time to move, but now that lasers are instant you take more damage and have less time to blink away. 

at least now wraiths have a fair armor value that DOESN'T make it less then a T1 drone. Ancient runner has a laser/energy armor of just 19 whilst wraith is 20???? how does this make sense that its only an increase of 1 lol.

just because it can blink and hyper-position does not mean every circumstance will allow it so. as being hit by bullets slows your blink and hyper position requires you to stand still for a moment.

## Changelog
:cl:
balance: wraith laser/energy armor adjusted for maturitys, now 15/20/30 instead of 10/15/20
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/wraith/castedatum_wraith.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/wraith/castedatum_wraith.dm
@@ -76,7 +76,7 @@
 	upgrade_threshold = 240
 
 	// *** Defense *** //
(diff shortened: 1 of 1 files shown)


bobmancer343:
d

TooFewSecrets:
Ancient should be 25 instead of 30 for consistency reasons.
Wraith doesn't need this buff anyway.

vilereaver41:
closing cause people are stinky

PR #9473, opened 2022-02-01
Infinite packets and ammo boxes

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

You can get infinite weapons, ammo, and mags to the heart of UNGA BUNGA but you cannot get infinite packets and ammo boxes. This might be storage creep, but take this into consideration:

1) If round lasts long enough, late-spawn marines won't get ammo box, which is weird considering that late-spawn marines shouldn't get this treatment for being late-spawn, which result in just getting backpacks full of ammo
2) No matter how much ammo marines have (with the exception of spec weapons and gears), if xenomorphs robust marines, there isn't enough ammo in the multiverse to save incompetent marines from xeno minor; you can't cramp all the ammo from the ammo box into your gun faster than the xenomorph swamp pit that is slashing you until the universe's heat death
3) QoL; 'nough said
4) As long we don't magically make ammo boxes fit into marines' internal storage, or essentially make a backpack within a backpack, this PR will only just help all marines get as much ammo to their heart's desire; a single marine will have to carry one ammo box, two ammo boxes, use motocycle to carry a crate full of ammo boxes, or have req to send ammo boxes for a single marine to ruck with all that sweet sweet ammo, making them susceptible to xenomorphs
5) You can't mass produce ammo boxes with mags in them. Not even loadout can save you.

## Changelog
:cl:
qol: Infinite packets and ammo boxes
balance: It will affect marine's storage since they can bring a buttload of ammo boxes, but the effect size will be small
/:cl:


Naaanii:
If it does end up being storage creep, you can just add it to req later.

BraveMole (approved):

PR #9474, opened 2022-02-01
Scatter balance pass

Waselon:
## About The Pull Request

Recent scatter changes made some guns unviable or even unusable. This PR attempts to make those guns usable again.

## Why It's Good For The Game

balance is good

## Changelog
:cl:
balance: T-60, T-18, TX-11, Minigun, Laser Machinegun recieved new scatter values
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -729,7 +729,7 @@
 
 	aim_slowdown = 1
 	wield_delay = 1.5 SECONDS
-	scatter = 4
+	scatter = 1
 	fire_delay = 0.2 SECONDS
 	accuracy_mult = 0.95
 	accuracy_mult_unwielded = 0.3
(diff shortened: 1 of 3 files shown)


Lumipharon:
T18 and laser machine gun are both fine as is. The scatter is entirely negligible for both.
Tx11, similar to the t18 has low enough scatter that you shouldn't miss anything you're shooting on screen with burst fire.
T60 is pretty bad without a bipod though, so I agree there.
Minigun is whatever, it probably does need better scatter considering it's expensive.

BraveMole:
@Pariah919 @SplinterGP 
Hello opinions please

Pariah919:
```
T18 and laser machine gun are both fine as is. The scatter is entirely negligible for both.
Tx11, similar to the t18 has low enough scatter that you shouldn't miss anything you're shooting on screen with burst fire.
T60 is pretty bad without a bipod though, so I agree there.
Minigun is whatever, it probably does need better scatter considering it's expensive.```
I pretty much agree with this, T60 needs less, T18/LASMG could be fine having a little less scatter base
TX-11 could use a tiny bit less in my experience playing with it.

BraveMole (approved):
mjp said yes

PR #9486, opened 2022-02-02
Warrior lunge range qol

BraveMole:
## About The Pull Request

Prevent from lunging if the target is too far

## Changelog
:cl:
qol: Prevent warrior from lunging if the target is too far
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
@@ -73,6 +73,11 @@
 	if(!.)
 		return FALSE
 
+	if(get_dist_euclide_square(A, owner) > 36)
+		if(!silent)
+			to_chat(owner, span_xenonotice("You are too far!"))
+		return FALSE
+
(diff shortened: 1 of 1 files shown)


Lefunny:
pov: instead of fixing how horrible warrior lunge is they nerf it even more instead

code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
```diff
@@ -73,6 +73,11 @@
 	if(!.)
 		return FALSE
 
+	if(get_dist_euclide_square(A, owner) > 36)
+		if(!silent)
+			to_chat(owner, span_xenonotice("You are too far!"))
```
TiviPlus, line 78: shouldnt this 36 be a define

TiviPlus (changes requested):

Naaanii:
Will it also prevent the lunge into corners but missing the target?

BraveMole:
> Will it also prevent the lunge into corners but missing the target?

no

TiviPlus (approved):

PR #9496, opened 2022-02-03
Drone get ranged building ability ; Xenos can no longer build on their turf

BraveMole:
## About The Pull Request

Title

## Why It's Good For The Game

It's meant to prevent the cheese of placing a wall on top of you and then melting/attacking the cades

## Changelog
:cl:
balance: Drone get ranged building ability ; Xenos can no longer build on their turf
fix: Hivemind construction are no longer strong as hivelord's
/:cl:


Mechswordsman:
Mother of god please. End my pain

Lumipharon:
Isn't building a wall on top of the cade pretty much just as bad?

Losenis:
>Isn't building a wall on top of the cade pretty much just as bad?

I personally haven't tried to do that as xeno but I don't think you can build a wall on the cade tile due to it being considered as a structure. You can't put sentries on cade tiles for that reason, which I remember well from tadpole shenanigans.

Naaanii:
The drone can wall next to the cade tile whilst still standing on the tile and bullets hit the wall first while the drone is still able to channel acid onto the cade.

BraveMole:
> The drone can wall next to the cade tile whilst still standing on the tile and bullets hit the wall first while the drone is still able to channel acid onto the cade.

True, but it's less cheesy. And at least the turret is shooting at something

Frampox:
Building on a drone will be extremely inconvenient.
The reason is that now you need to select an adjacent tile in order to build anything.
You can build without selecting the desired tile, which is much faster and more convenient. 

So I really don't like this request. I want to use only WASD + right-click on any space

Al-1ce:
> So I really don't like this request. I want to use only WASD + right-click on any space


people actually use rightclick?

PR #9509, opened 2022-02-04
Makes autoburst extra (extra) delay a var

Lumipharon:
## About The Pull Request
Currently autoburst serves no purpose other than to lower your dps in exchange for less carpal tunnel, as it adds a significant extra delay multiplier between bursts.

This PR adds a var that can be configured on guns to change this extra delay to a set amount, so guns can get this tweaked to make it actually useful.
Changed the default delay to x2 fire_delay instead of x3. It's still worse than manually bursting, but less so.

## Why It's Good For The Game
Autoburst might have a reason to exist.

## Changelog
:cl:
add: new var to control autoburst extra (extra) delay
/:cl:

### DIFF
--- a/code/datums/components/autofire.dm
+++ b/code/datums/components/autofire.dm
@@ -29,7 +29,6 @@
 
 	RegisterSignal(parent, COMSIG_GUN_FIRE_MODE_TOGGLE, .proc/modify_fire_mode)
 	RegisterSignal(parent, list(COMSIG_GUN_AUTOFIREDELAY_MODIFIED, COMSIG_XENO_AUTOFIREDELAY_MODIFIED), .proc/modify_fire_shot_delay)
-	RegisterSignal(parent, COMSIG_GUN_AUTOBURSTDELAY_MODIFIED, .proc/modify_autoburst_shot_delay)
 	RegisterSignal(parent, COMSIG_GUN_BURST_SHOTS_TO_FIRE_MODIFIED, .proc/modify_burst_shots_to_fire)
 	RegisterSignal(parent, COMSIG_GUN_BURST_SHOT_DELAY_MODIFIED, .proc/modify_burstfire_shot_delay)
 	RegisterSignal(parent, list(COMSIG_GUN_FIRE, COMSIG_XENO_FIRE), .proc/initiate_shot)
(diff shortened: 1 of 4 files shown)


Lumipharon:
@BraveMole what are you thoughts on changing the default value here? x3 delay is bizarre, x1 makes it a better autofire in most cases, but you've said before 'clicking faster shouldn't give you an advantage'?

code/datums/components/autofire.dm
```diff
@@ -37,7 +38,7 @@
 	auto_fire_shot_delay = _auto_fire_shot_delay
 	burstfire_shot_delay = _burstfire_shot_delay
 	burst_shots_to_fire = _burst_shots_to_fire
-	auto_burst_fire_shot_delay = 3 * auto_fire_shot_delay
+	auto_burst_fire_shot_delay = _auto_burst_fire_shot_delay * auto_fire_shot_delay
```
BraveMole, line 41: not a fan of this solution, this is less modular than having a value independant of auto_fire_shot_delay.

It's not that bad though, but in that case you need to rename the argument _auto_burst_fire_shot_delay into like auto_burst_fire_shot_multiplicator or something

code/datums/components/autofire.dm
```diff
@@ -59,6 +60,11 @@
 	SIGNAL_HANDLER
 	auto_fire_shot_delay = _auto_fire_shot_delay
 
+///Setter for autoburst shot delay
+/datum/component/automatedfire/autofire/proc/modify_autoburst_shot_delay(datum/source, _auto_burst_fire_shot_delay)
+	SIGNAL_HANDLER
+	auto_burst_fire_shot_delay = _auto_burst_fire_shot_delay
```
BraveMole, line 66: if you don't allow for a direct set in the constructor, it seems weird to allow it here

code/modules/projectiles/gun_system.dm
```diff
@@ -223,6 +223,8 @@
 	var/burst_delay = 0.1 SECONDS
 	///When burst-firing, this number is extra time before the weapon can fire again. Depends on number of rounds fired.
 	var/extra_delay	= 0
+	///when autobursting, this is the amount of extra (extra) time before the weapon fires again. Multiplier of fire_delay
+	var/autoburst_delay = 3
```
BraveMole, line 227: if it's a multiplier, change the name accordingly

BraveMole (changes requested):

Losenis:
>x3 delay is bizarre, x1 makes it a better autofire in most cases, but you've said before 'clicking faster shouldn't give you an advantage'?

If I knew code I'd have rather changed the default value to 2 while also creating a var that changes normal burst delay also to 2. This'd probably make tx11 autoburst better than full auto, but that's kind of the point of the gun so it'd be reasonable for it. Yet to do the math on the T12 and T18 dps on either mode.

Lumipharon:
Made the var a fixed amount instead of a multiplier.
Made the default x2 instead of x3 firedelay.

code/modules/projectiles/gun_system.dm
```diff
@@ -223,6 +223,8 @@
 	var/burst_delay = 0.1 SECONDS
 	///When burst-firing, this number is extra time before the weapon can fire again. Depends on number of rounds fired.
 	var/extra_delay	= 0
+	///when autobursting, this is the amount of extra (extra) time before the weapon fires again. Multiplier of fire_delay
+	var/autoburst_delay = null
```
BraveMole, line 227:
```suggestion
	var/autoburst_delay = 0
```

BraveMole (changes requested):

Lumipharon:
@BraveMole should be set now.

TiviPlus (approved):

BraveMole (approved):

PR #9510, opened 2022-02-04
Light Marksman bullet Buff and Battle Rifle auto-burst Buff

Redstonize:
## About The Pull Request

As per title says, but it is balanced.
Now requires https://github.com/tgstation/TerraGov-Marine-Corps/pull/9509 .

## Why It's Good For The Game

1-hand: Makes the Battle Rifle bullets moderately less underpowered. This will slightly reward marines using the T-64 BR for it's higher magazine capacity over the T-37 DMR but has less damage, range and accuracy. THIS DOES NOT MAKE BURST SUPERIOR TO AUTOFIRE.
2nd-hand: You don't need to click as much. That's for the benefit of just burst users.

## Changelog
:cl:
balance: Added 9.99 damage, 2 AP, 0.2 sundering to the Light Marksman Bullet for the T-64 BR (from 30 damage, 10 penetration, 1 sunder) and made it's auto burst delay shorter.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -608,9 +608,9 @@ datum/ammo/bullet/revolver/tp44
 	hud_state = "hivelo"
 	hud_state_empty = "hivelo_empty"
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
-	penetration = 10
-	damage = 30
-	sundering = 1
+	penetration = 12
+	damage = 37.5
+	sundering = 1.2
 
 /datum/ammo/bullet/rifle/standard_br/incendiary
 	name = "incendiary light marksman bullet"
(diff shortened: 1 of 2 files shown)


Redstonize (commented):
In case nobody's convinced..
I've ran one test which is using the Ancient Defiler as an ubiquitous xenomorph supposed to represent **the average xenomorph player while not moving and not using any wards/pheromones.** The test is supposed to show how fast you can kill the xeno with autofire, since that's often the better fire mode for most weapons. I've used my personnal phone as chronometer for the test, and used both the T-64 and T-37 as side comparison.
_Test results show that it takes 7.00 seconds to crit an Ancient Defiler 5 tiles away using the T-64._
_It also showed that it takes 5.40 seconds to crit the same Ancient Defiler (rejuvenated) using the T-37._ 
_What are we gonna do ? Nerf the DMR ?_
With this PR in mind, the TTC for an Ancient Defiler nearby (not moving and not warded or pheromoned') comes from 7.00 whole seconds to 5.38 seconds on Auto fire. Therefore the gun is now slightly better or equivalent to the T-37 using auto fire or single fire. 
Burst is no better with the results being similar; the TTC comes lower at around 5.40 seconds.
Remember, a real xeno player is likely to be moving and warded, so this test only shows that on static targets, the guns are equivalent now, and before that, the T-64 was very much on the worse side of DPS.

PR #9553, opened 2022-02-07
Lowers ZX76's cost to 40

Losenis:
## About The Pull Request
Changes the cost of ZX76 from 100 to 40.

## Why It's Good For The Game
The ZX is a sidegrade shotgun bordering downgrade in more cases than not. Other than PBing xenos or just funny double slug snipe on big xenos that don't get knocked back (more on that below), it doesn't do much better than the other shotguns or is actively worse at it. Not only that, its benefits also become a massive detriment in case of FF. Burst damage of two flechettes that can be achieved with DB is not worth more req points than a Tx8 with a full ammo belt.

In normal usage it also ends up accidentally worse than it should be on paper. This is because using it with buckshot or slugs causes a knockback on most xenos, which has them actually miss the second shot due to being on the ground. End effect of this is a gun that does what a t35 does at the cost of twice the ammo usage + way higher fire delay, using individual shells unlike the Tx16 which uses mags with simpler reload. Double barrel becomes better in this instance because you can time the shots so that they get up and get hit by the second volley for a gun that is available from roundstart on vendors.

Exhibit A: Slug burst, one hits xeno, other misses due to xeno being on the ground knocked down
image
image

## Changelog
:cl:
add: Added more mechanics or gameplay changes
expansion: Expands content of an existing feature
del: Removed old things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -317,7 +317,7 @@ WEAPONS
 /datum/supply_packs/weapons/zx76
 	name = "ZX-76 Twin-Barrled Burst Shotgun"
 	contains = list(/obj/item/weapon/gun/shotgun/zx76)
-	cost = 100
+	cost = 40
 
 /datum/supply_packs/weapons/shotguntracker
 	name = "12 Gauge Tracker Shells"


Pariah919 (changes requested):
No. 80, maybe. But no.

David-Stormwell:
have thy attempted to use anything but slugs on a db?

Waselon:
> No. 80, maybe. But no.

40 is reasonable, even more reasonable if masterkey is removed

Losenis:
> have thy attempted to use anything but slugs on a db?

DB doesn't cost 100 req points, so yeah, I've had the chance to use the DB.

ilikesaying:
the ZX's advantage over DB is the internal magazine meaning less reloading required. This means it has a faster fire-rate overall and 2 flech can really fuck up xenos regardless of armor due to the sheer volume of fire

Losenis:
Still not worth 10 more points than a Tx8 with full ammo belt though

Lumipharon:
Not having to buy ammo via req is a huge advantage in of itself, as is the ability to have underbarrel weapons.
All the underbarrels massively boost it's ability to burst fuckers to death that no other gun can match.
Also certain ammo not working with ZX mechanics doesn't mean there is an issue with the gun, the issue is you should be using fletch.

Losenis:
For 100 req points? No.

BraveMole (changes requested):
It's not a sidegrade, and honestly masterkey makes it more unique

BraveMole:
> Wait so now masterkey on a shotgun is ok?

On a 100 points req shotgun that basically has only this for it, yes?

Waselon:
if masterkey stays, it should cost 60-75 maybe

BraveMole (changes requested):
Like MJP said, maybe 70-80, and it keeps masterkey

OpsCoretango:
Better to hive him angled grip and remove master

TiviPlus (changes requested):
What brave said

Araulius:
> the ZX's advantage over DB is the internal magazine meaning less reloading required. This means it has a faster fire-rate overall and 2 flech can really fuck up xenos regardless of armor due to the sheer volume of fire

do you know what a magazine is? the zx has be reloaded shell by individual shell. this is not a magazine fed gun.

Losenis:
He did say _internal_ magazine.

OpsCoretango:
okay, 84 points is not-bad just drop someone his cost or i call DP-12 a "burn in oblivion"

OpsCoretango:
okay, burn in oblivion new cost 😐

OpsCoretango:
lets stop on 70+ points cost and close it

Al-1ce:
> the ZX's advantage over DB is the internal magazine meaning less reloading required. This means it has a faster fire-rate overall and 2 flech can really fuck up xenos regardless of armor due to the sheer volume of fire


The internal magazine separates reloading, but the delay makes it as equally slow in usage as the DB is. Not having to worry about buying unique ammunition when they should be standardized isn't really a "bonus" -- 100 points for what is really little more than a DB with a few more spaces for shells is absurd.

TiviPlus:
Closing as stale

PR #9565, opened 2022-02-07
Akimbo delay applies to autofire

Lumipharon:
## About The Pull Request
When firing guns akimbo, there is an extra fire delay associated with it that means you shoot a bit slower. Currently this only applies to semi auto firing, not to full auto.
Makes autofire actually respect the akimbo_additional_delay var.

Also reduced the extra delay value for all SMGs to 0.2 from 0.5. This means guns that fire at 0.15 will fire at 0.2 akimbo.
Rifles reduced to 0 as they're terrible at akimbo, and don't need to be even worse.

## Why It's Good For The Game
Akimbo additional delay now actually works with autofiring. This has balance implications for guns that are used akimbo with autofire :^)

## Changelog
:cl:
fix: Autofiring while akimbo now actually shoots slower like semi auto
balance: Reduced akimbo additional delay for SMG's and rifles to compensate
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -600,6 +600,7 @@
 			return
 		if(gun_user.hand && isgun(gun_user.r_hand) || !gun_user.hand && isgun(gun_user.l_hand)) // If we have a gun in our inactive hand too, both guns get innacuracy maluses
 			dual_wield = TRUE
+			modify_fire_delay(akimbo_additional_delay) // Adds the additional delay to auto_fire
 			if(gun_user.get_inactive_held_item() == src && (gun_firemode == GUN_FIREMODE_SEMIAUTO || gun_firemode == GUN_FIREMODE_BURSTFIRE))
 				return
 		if(gun_user.in_throw_mode)
@@ -650,7 +651,9 @@
 	shots_fired = 0//Let's clean everything
 	set_target(null)
 	windup_checked = WEAPON_WINDUP_NOT_CHECKED
-	dual_wield = FALSE
+	if dual_wield == TRUE
+		modify_fire_delay(-akimbo_additional_delay) // Removes the additional delay from auto_fire
+		dual_wield = FALSE
 	gun_user?.client?.mouse_pointer_icon = initial(gun_user.client.mouse_pointer_icon)
 
 ///Inform the gun if he is currently bursting, to prevent reloading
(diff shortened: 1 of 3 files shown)


BraveMole (changes requested):
akimbo_additional_delay is a multiplier

Lumipharon:
> akimbo_additional_delay is a multiplier

I realised this yesterday, but testing is weird.
If it was adding 0.5, delay it should be 0.65 (like a dmr) but it definitely isn't that slow. It does however seem to shoot slower than one handed... code, how do you work.

code/modules/projectiles/gun_system.dm
```diff
@@ -652,7 +652,7 @@
 	set_target(null)
 	windup_checked = WEAPON_WINDUP_NOT_CHECKED
 	if (dual_wield == TRUE)
-		modify_fire_delay(-akimbo_additional_delay) // Removes the additional delay from auto_fire
+		modify_fire_delay(-(fire_delay - fire_delay/(1 + akimbo_additional_delay))) // Removes the additional delay from auto_fire
```
BraveMole, line 655:
```suggestion
		modify_fire_delay(-fire_delay + fire_delay/(1 + akimbo_additional_delay)) // Removes the additional delay from auto_fire
```

BraveMole (changes requested):

BraveMole (changes requested):
ok for 0,2 for t90, but what about the other rifles/mg? they should stay at zero, it's not like they are any good

Lumipharon:
> ok for 0,2 for t90, but what about the other rifles/mg? they should stay at zero, it's not like they are any good

Default is still 0.5, but does anyone actually use any rifles akimbo? You need RC + Gyro to fire straight, and you still have terrible accuracy.
I'm happy to add lower values in case people do though.

code/modules/projectiles/gun_system.dm
```diff
@@ -650,7 +651,9 @@
 	shots_fired = 0//Let's clean everything
 	set_target(null)
 	windup_checked = WEAPON_WINDUP_NOT_CHECKED
-	dual_wield = FALSE
+	if (dual_wield == TRUE)
+		modify_fire_delay(-fire_delay + fire_delay/(1 + akimbo_additional_delay)) // Removes the additional delay from auto_fire
```
TiviPlus, line 655:
```suggestion
	if(dual_wield)
		modify_fire_delay(-fire_delay + fire_delay/(1 + akimbo_additional_delay)) // Removes the additional delay from auto_fire
```

Lumipharon: Neat

BraveMole (approved):

PR #9572, opened 2022-02-08
Handheld charger rightclick action

EvilMelonTheSecond:
## About The Pull Request
Makes rightclicking handheld charger take out the battery

## Why It's Good For The Game
Small QoL for taking out the battery without having to take out the whole charger from their bag/container.

## Changelog
:cl:
add: Rightclicking handheld charger now takes out the battery instead of leftclick
/:cl:


BraveMole (changes requested):
then remove the code from attack hand

PR #9587, opened 2022-02-09
Attempt to make Water useful

Three333Digit:
## About The Pull Request
Water no longer purges toxin, increases its purge rate from 1 to 2, but in exchange you gain double the stamina regen of synaptizine. Personally, I think trading your premed in exchange for stamina regen is interesting. You're still fighting against pain slowdown and with enough neurotoxin, the benefits water provides can become a liability instead.

## Why It's Good For The Game
Attempts to make water useful, now those Nanotrasen-sponsored bottled spring water sitting in FOB actually has a use now. 

## Changelog
:cl:
balance: water no longer purges toxin, purge rate increased from 1 to 2, in exchange you get double the stamina regen of synaptizine
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/other.dm
+++ b/code/modules/reagents/reagents/other.dm
@@ -34,7 +34,7 @@
 	color = "#0064C8" // rgb: 0, 100, 200
 	overdose_threshold = REAGENTS_OVERDOSE * 2
 	custom_metabolism = REAGENTS_METABOLISM * 5 //1.0/tick
-	purge_list = list(/datum/reagent/toxin, /datum/reagent/medicine, /datum/reagent/consumable)
+	purge_list = list(/datum/reagent/medicine, /datum/reagent/consumable)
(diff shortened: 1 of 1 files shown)


Naaanii:
This is a pretty big buff for something that's freely provided and has a minor use.

jroinc:
It's *water*. It's not intended to be some zoomer drug that restores tens of points per second forever or be better at purging toxins and chemicals than specialized drugs.

The rate is as it is (complex and trailing off to zero) to keep someone from just chugging a water bottle then entering *PERMASPRINT MODE*, which this will allow.

YaBoiFrey:
Mmmm meth water

Derrick2Electricboogaloo:
But water cleanses your organs it should purge toxin. Also it purges toxin? Damn I new it can purge meds but now more reason for me to carry it.

novaepee:
H Y D R A T E ! ! ! !

PR #9614, opened 2022-02-11
Miner upgrades in engineer point vendor

Three333Digit:
## About The Pull Request
Title. For 4 points since that's the same points you would spend if you were to convert bought metal into miner modules back when they were in the autolathe. 

## Why It's Good For The Game
Removing them from the autolathe is good since it prevents printing infinite miner modules with enough patience, but removing an engineer's ability to acquire them outside of req stomps on any mining plans an engineer might have for the round. This returns miner control to the role that is responsible for repairing them.

## Changelog
:cl:
balance: miner upgrades in engineer point vendor for 4 points (does not include autominer)
/:cl:


Luxyboo:
no longer waiting 20 minutes for a 5 point upgrade. :)

David-Stormwell:
Sounds swell enough, hardly see them used anymore anyways.

PR #9604, opened 2022-02-10
Fix larva regress exploit

BraveMole:
## About The Pull Request

Fix larva being ancient when you regress

## Changelog
:cl:
fix: Fix larva regress exploit
/:cl:

PR #9627, opened 2022-02-12
feat: Reworked xeno_upgrade section

KuzyXD:
## About The Pull Request
you can read why in comments #9604 

## Changelog
:cl:
fix: issue when you de-evolve and don't get +1 maturity rank as should
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/evolution.dm
@@ -303,10 +303,15 @@
 		if(new_xeno)
 			qdel(new_xeno)
 		return
-	if(tier != XENO_TIER_ONE || !regression )
-		new_xeno.upgrade_xeno(upgrade, TRUE)
-	if(!regression && upgrade != XENO_UPGRADE_INVALID)
-		new_xeno.upgrade_xeno(new_xeno.upgrade_prev(), TRUE)
+
+	if(!regression)
+		if(upgrade != XENO_UPGRADE_INVALID)
+			new_xeno.upgrade_xeno(upgrade_prev(), TRUE)
+	else
+		if(tier == XENO_TIER_ONE)
+			new_xeno.upgrade_xeno(XENO_UPGRADE_INVALID, TRUE)
+		else
+			new_xeno.upgrade_xeno(upgrade_next(), TRUE)
 
 	SEND_SIGNAL(src, COMSIG_XENOMORPH_EVOLVED, new_xeno)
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/evolution.dm
```diff
@@ -303,10 +303,15 @@
...
-		new_xeno.upgrade_xeno(upgrade, TRUE)
-	if(!regression && upgrade != XENO_UPGRADE_INVALID)
-		new_xeno.upgrade_xeno(new_xeno.upgrade_prev(), TRUE)
+
+	if(!regression)
+		if(upgrade != XENO_UPGRADE_INVALID)
+			new_xeno.upgrade_xeno(upgrade_prev(), TRUE)
+	else
+		if(tier == XENO_TIER_ONE)
+			new_xeno.upgrade_xeno(XENO_UPGRADE_INVALID, TRUE)
```
TiviPlus, line 312: Why are you setting an invalid upgrade here

TiviPlus (changes requested):

code/modules/mob/living/carbon/xenomorph/evolution.dm
KuzyXD: > Why are you setting an invalid upgrade here

Because of XENO_TIER_ZERO (hivemind, larva) has not upgrade points.

code/modules/mob/living/carbon/xenomorph/evolution.dm
```diff
@@ -303,10 +303,15 @@
...
+		else
+			new_xeno.upgrade_xeno(upgrade_next(), TRUE)
```
BraveMole, line 314: this was removed intentionally, because that creates exploit

KuzyXD: > this was removed intentionally, because that creates exploit

What kind of exploit?

KuzyXD: I tested it in many ways. And btw it's commit on RUTGMC

BraveMole: You can get primo iirc. There was another one with queen devo, cant remember.

Also this meant that going from one t3 to another had no cost, which wasnt really balanced imo

hyper2snyper:
What does this actually do?

You probably shouldn't be resolving maintainers comments

PR #9227, opened 2022-01-10
Evacuation can prevent xeno major

BraveMole:
## About The Pull Request

Change the evacuation system:
- Add more evacuation shuttles on most maps
- The pods launch 10 minutes after the evacuation order, rather than 3 currently
- If more than 50% of marines manage to evacuate, xeno cannot get a major (they still can get minor if they kill the rest of the marines)

## Why It's Good For The Game

This gives marines another option in hijack situation :
- They can try to guard SD for 20 minutes, and the round is draw
- They can try to wipe out xenos and get a minor
- Or now, they can try to evacuate and deny xenos a major. Usefull if marines are really not enough to win or even to hold SD

## Changelog
:cl:
add: Evacuation can prevent xeno major
/:cl:

PR #9631, opened 2022-02-12
Fixes Evac

Naaanii:
## About The Pull Request

Removes max capacity requirement. Any number of marines can now evac.
@Walarks 

## Why It's Good For The Game

With #9227 Marines can now evac after 10 minutes to try and prevent a Xeno major. More than 50% of marines
alive have to evac to attain this round end conclusion but it's harder to do so on some shipmaps like Minerva or on higher pops. This removes the evac capacity limit and fixes a few checks (Tested and seems to work so far).

## Changelog
:cl:
del: Removes evac max capacity.
fix: Fixed an apparent infinite loop check.
/:cl:

### DIFF
--- a/code/modules/shuttle/escape_pod.dm
+++ b/code/modules/shuttle/escape_pod.dm
@@ -10,9 +10,6 @@
 
 	var/list/doors = list()
 	var/list/cryopods = list()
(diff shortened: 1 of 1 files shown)


Naaanii:
> Dont see any infinite loops but ok

@Walarks

PR #9654, opened 2022-02-17
Xeno's turret deploy time reduced to 10s.

Passe2142:
## About The Pull Request

Halves the build time for xeno turrets. If its balance or QoL I dont mind either. 15s to build something is too much.

## Why It's Good For The Game

When prepping as a hive leader be either low pop or normal pop the 15s timer for a turret that will start with 300 hp feels like a punishment. Even if your being pushed hard by marines, 15s to make a defense structure is too much when you will be mostly needed at the front.

## Changelog
:cl:
balance: Build time from 15s to 10s.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
@@ -151,7 +151,7 @@ GLOBAL_LIST_INIT(tier_to_primo_upgrade, list(
 	psypoint_cost = XENO_TURRET_PRICE
 	flags_gamemode = ABILITY_DISTRESS
 	///How long to build one turret
-	var/build_time = 7 SECONDS
+	var/build_time = 10 SECONDS
 	///What type of turret is built
 	var/turret_type = /obj/structure/xeno/xeno_turret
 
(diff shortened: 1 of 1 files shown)


Pill-U:
I think this makes turretspamming to hold a choke too viable. Perhaps lower the build time during pre-shutters?

Passe2142:
> 

There is still a limit distance from each other of 6 tiles,  and the most turret spam thing you could achieve would be 4 if you can  said choke. They would start with 300hp, hence marines will shoot it down easily if you want to slow them up it wont do. And for prep its intended cause you also need to maze. Not just make funny turret go brr. Marines can PC cut them, shoot them from off screen, flamers, UBF, plasma pistol and less than a magazine to kill it. And when you happen to be the only one that can make them its pain. The hive has more things to expend the points on, primos, spawners, smart minions, towers. If they can spare a few turrets it would be on prep.

Lumipharon:
The long build time is to stop combat building, which would be a real problem otherwise. Honestly just making them start on 1hp or whatever but with a short build timewould be good. It would be dumb to combat build but you could build actual defenses in a reasonable amount of time.

Passe2142:
They could start on a 100HP, I dont wanna waste time making them or see hive leaders struggle to make a defense on hard push.
Marines got tools. Lately the turret lacks impact or feels like a guardian for cocoons. Also I dont an issue with combat building its a tool for xenos to use. We do with what we have.

Naaanii:
Seems fine since new turrets don't really last long against a marine push though I'd honestly make it 10 seconds instead of straight up halfing it. Turret spam shouldn't be too viable because of easy it is to dismantle newly formed turrets. It can delay marine pushes but also cost points so it's done at a cost and not much like wall spam which is technically more infinite in terms of resources.

Passe2142:
!0s is fine too.

David-Stormwell:
Should be fine in theory, as you can't place turrets next to each other any longer which makes assaulting 1 easy enough.

PR #9661, opened 2022-02-18
Reduces Acid well Cost.

vilereaver41:
## About The Pull Request

Attempts to find a good price for Acid well so it doesn't cost an arm and a leg to make or fill up whilst also making sure its not super spammed

## Why It's Good For The Game

Acid well's are a defensive & utility structurefor xenomorphs, but they use up a LOT of plasma 500 to make and 200 to build, when creating a maze or base xenos wont be able to spare the plasma, Lowered Acid Well cost will allow xenos to better hold defensive lines rather then falling victim to marine pushes hat cant be stopped.

When retreating Marines will often follow xenos into weeded areas, but I believe with the reduced cost for acid wells they will think twice.

## Changelog
:cl:
balance: Acidwell now 400 down from 500
balance: Filling an acidwell is now 150 down from 200
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -443,7 +443,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 #define XENO_DEFAULT_VENT_EXIT_TIME 2 SECONDS //Standard time for a xeno to exit a vent.
 #define XENO_DEFAULT_ACID_PUDDLE_DAMAGE 14 //Standard damage dealt by acid puddles
 #define XENO_ACID_WELL_FILL_TIME 2 SECONDS //How long it takes to add a charge to an acid pool
-#define XENO_ACID_WELL_FILL_COST 200 //Cost in plasma to apply a charge to an acid pool
+#define XENO_ACID_WELL_FILL_COST 100 //Cost in plasma to apply a charge to an acid pool
 #define XENO_ACID_WELL_MAX_CHARGES 5 //Maximum number of charges for the acid well
 
 #define HIVE_CAN_HIJACK (1<<0)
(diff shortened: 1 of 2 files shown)


Mrrpip:
Good thing the players don't know player scorpions can fill acid wells with their 1500 plasma pool!

vilereaver41:
> > Good thing the players don't know player scorpions can fill acid wells with their 1500 plasma pool!
> 
> 700 for the spawner and 300 for control minions. and that has no bearing on placing acid wells and setting up a base before shutters.

Naaanii:
Nerf the damage as well if you're going to nerf the cost so acid well spam won't be an issue.

vilereaver41:
> Acid wells are fine, their only issue is castes that can't make them never fucking bother to fill the damn things. Like if you empty one out, have the damn courtesy to refill it.

because they take 500 plasma and 200 plasma to fill????  t1-t2 castes who fill it will waste their entire plasma reserve and have to rest anyway.

David-Stormwell:
Seeing quite a bit of room for spamming after this PR. 
If you want this to be accepted make it 400/150 then we can go from there.

vilereaver41:
400 imo still seems to high. no marine with half a brain cell will willingly walk into the big glowing pool of acid.

InterroLouis:
Right so I'm just gonna go ahead and clock in here with my my ill informed knowledge gained of being an idiot. 
The cost to make them doesn't matter that much for spamming them because of the 2 minute cooldown timer. No caste isn't going to regenerate 500 plasma in 2 minutes that also has the ability to make acid wells. The main effect of the lowered plasma cost will mean that when the cooldown IS up, the player will be more likely to make one because it eats into their plasma store less. 
As for the refill, well, yeah this won't make it so a runner is going to put plasma in, but it softens the blow on a xeno that happens to need to rest to heal but has plasma spare. 
Personally I don't think this will lead to excessive spamming of plasma wells, or even much of an increase from what gets made now,  but what it will lead to will be more of the plasma wells that do exist being filled. 
Also I've played as a scorpion before and just went around filling every acid well I saw, good god scorpions regen their plasma slow as hell for having such a big pool of it.

PR #9662, opened 2022-02-18
Replaces the AI vox voice again

RipGrayson:
## About The Pull Request

Replaces the old AI vox with a new set. Again. 

## Why It's Good For The Game

My long term plan is to introduce vox for commanders to coordinate their troops better, but the current voice is unpopular and certain players would probably riot if they had to listen to ~~Bonzi Buddy~~ our current vox voice all game long without a chance of escaping it.
Hopefully the new voice will be less grating.

## Changelog
:cl:
soundadd: Replaced the old TTS vox voice with a more modern version.
sounddel: Removed all old TTS vox in favor of a new version.
/:cl:


novaepee:
Text to Speech, baby

AnCopper:
Why not just use the Half Life 1 military announcer?

RipGrayson:
> Why not just use the Half Life 1 military announcer?

I've been over this a fair number of times now, I really should just post it at the start of all vox related PRs.

1. We cannot use the Half Life vox, it's the property and copyright of Valve Inc, the only reason we had it briefly was a stopgap measure while I generated  a new vox announcer.
2. HL military announcer has 200 lines, the regular vox announcer has 500 something. There's no vocabulary for things like "shrike", "marine" "self destruct" etc.
3. Even if Gabe Newell descended from above and allowed exclusive usage of the original vox for TGMC, we don't know how the vox was made beyond a vague clue, so there's no addressing the above mentioned lack of useful words.
4. Before anyone suggests using the TG vox I would like to point out that about half its vocabulary are irrelevant words and the person who made it has cut all ties from the project, so there's no making new words for it either.

PR #9664, opened 2022-02-18
Removes plasma pistol from the T35 and T39 as an attachment option

bowlofcereal:
## About The Pull Request

Removes Plasma Pistol from the T35 and T39 as an attachment

## Why It's Good For The Game

Under barrel flamers were removed for the exact same reason I am removing this. Plasmapistol under barrel, for those unaware, allows a marine to add a weapon that sets xenos on fire to their firearm. This is allowed on shotguns, leading to people being able to solo most castes in the game more so then they already can. When combined with a group of marines helping, this almost always results in the xenos death if caught in the blast. Removing this is a good step in removing powercreep features.

Examples of why this is powercreep:
https://www.youtube.com/watch?v=6KGZCEojNSA

This does not even take into consideration the fact that there would be more than likely more than one marine shooting at these xenos, since this was done in a controlled environment. I have no plans on removing it from other guns, but they should not be on shotguns.

## Changelog
:cl:
balance: Removed plasma pistol as an option from T35 and T39 attachments
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1557,7 +1557,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	pixel_shift_y = 17
 	stream_type = FLAMER_STREAM_CONE
 
-///Funny red wide nozzle that can fill entire screens with flames. Admeme only. 
+///Funny red wide nozzle that can fill entire screens with flames. Admeme only.
 /obj/item/attachable/flamer_nozzle/wide/red
 	name = "red spray flamer nozzle"
 	desc = "It is red, therefore its obviously more effective."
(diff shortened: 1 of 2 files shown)


EvilMelonTheSecond:
The video also doesn't compare it to normal shotgun kill times OR the difference between shotty+plasgun and other guns+plasgun. I'm not against the PR but this seems like another shotgun saltPR

bowlofcereal:
> The video also doesn't compare it to normal shotgun kill times OR the difference between shotty+plasgun and other guns+plasgun. I'm not against the PR but this seems like another shotgun saltPR

I will happily post a video with normal shotguns' kill times as well in this PR if you give me a day or two to get one up. I'm not removing it from other guns because I have yet to see them used on any other gun besides the shotgun. The shotgun having its stun, combined with the xenos being lit on fire, can result in some instantaneous kills. I am sure nobody enjoys fighting against that. The shotgun should stay as a strong close quarter deterrent, yes, but the plasma pistol makes it far more than it should be.

Naaanii:
Seems like a good alternative to straight up nerfing the plasma pistol.

OpsCoretango:
next time get friend player, for "dodging target"
Its balanced, what matter? Xeno skill issue.

PR #9679, opened 2022-02-21
(WIP) Refactor on_hit code for ammo_datums.dm

Lumipharon:
## About The Pull Request
Putting up this draft so maintainers can tear it apart because it may be bad.

This is my poor attempt at cleaning up the code and making it easier/neater to add on hit effects to ammo that apply regardless of what is hit, i.e. rockets etc)

Instead of having 4 pretty much copy pasta procs for every single ammo calling up the exact same effect proc, the 4 main on hit procs now call a single stub proc (instead of being stubs themselves), meaning you just need to add the specific effect to the on_hit_all proc for each ammo type.

Still more work to do on other effects, just drop_nade so far (and some things I think I can improve for the beno nade_drop stuff), but it compiles and initial testing shows it seems to work, but I didn't test everything yet.

## Why It's Good For The Game
Less shitcode (or more, ayylmao), easier to add shit, less copy paste stuff.

## Changelog
:cl:
refactor: refactored on_hit effects
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -49,18 +49,25 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	var/on_pierce_multiplier = 1
 
 /datum/ammo/proc/do_at_max_range(obj/projectile/proj)
+	on_hit_all(get_turf(proj), obj/projectile/proj)
 	return
 
 /datum/ammo/proc/on_shield_block(mob/M, obj/projectile/proj) //Does it do something special when shield blocked? Ie. a flare or grenade that still blows up.
 	return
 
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/ammo_datums.dm
```diff
@@ -2331,20 +2307,20 @@ datum/ammo/bullet/revolver/tp44
 	max_range = 10
 
 /datum/ammo/grenade_container/on_hit_mob(mob/M,obj/projectile/P)
-	drop_nade(get_turf(P))
+	on_hit_all(get_turf(P), P)
 
 /datum/ammo/grenade_container/on_hit_obj(obj/O,obj/projectile/P)
-	drop_nade(get_turf(P))
+	on_hit_all(get_turf(P), P)
```
Walarks, line 2313: Why not call parent?

Lumipharon:
Closed for now, I know how to do this differently but we'll see if it's actually worth while.

PR #9696, opened 2022-02-24
Boiler cloud radius no longer scales past Ancient

Lumipharon:
## About The Pull Request

Primo isn't supposed to scale stats, however boiler gas clouds scale with maturity, leading to monstrously huge clouds when Primo. Fixes this, seriously, ancient clouds are already massive.

## Why It's Good For The Game
Big stinky cloud go home.

## Changelog
:cl:
fix: Boiler gas cloud no longer scales in size past Ancient
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -2091,7 +2091,7 @@ datum/ammo/bullet/revolver/tp44
 		var/mob/living/carbon/xenomorph/X = firer
 		smoke_system.strength = X.xeno_caste.bomb_strength
 		if(fixed_spread_range == -1)
-			range = max(2, range + min(X.upgrade_as_number(), 4))
+			range = max(2, range + min(X.upgrade_as_number(), 3))
 		else
 			range = fixed_spread_range
 	smoke_system.set_up(range, T)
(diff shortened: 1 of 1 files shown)


ilikesaying:
it scales?

vilereaver41:
Cope

DeltaFire15:
This is an intentional feature which was outlined by me in the original PR.
Boiler's primordial is getting better bombard, which materializes as "slightly better normal globs + get lance globs" whereas the somewhat bigger normal size compensates for situations where lances are simply not useful and or one prefers base globs, which can be a pretty common occurance.
Upgrading an ability as opposed to stats, which are not changed, is not unique to boiler either, Warrior's empower mechanic for example drastically buffs base abilities, bull's whole primordial is a upgrade to its base ability, etc.

Lefunny:
boiler is already hard enough to survive as, also it takes a real long time to evolve in t3, cope

Andrewhedge:
cope

Lumipharon:
Fair enough if it was intended, but I don't think it's great gameplay to obscure almost an entire screen with acid smoke.


> boiler is already hard enough to survive as, also it takes a real long time to evolve in t3, cope
If you just mature as T2 to primo first, it's super fast.

Lumipharon:
@TiviPlus Thoughts? Primo wasn't supposed to scale stats, boiler does considerably.

TiviPlus:
> @TiviPlus Thoughts? Primo wasn't supposed to scale stats, boiler does considerably.

Im still thinking about it

PR #9700, opened 2022-02-25
Let Sugar regenerate blood

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

Did you know that sugar never give blood despite ironsugar pills exist? Yeah, it was a weird arcane knowledge that only a few people know.

Anyway, sugar now gives blood.

## Changelog
:cl:
fix: sugar can now regenerate blood for ironsugar to work
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/food.dm
+++ b/code/modules/reagents/reagents/food.dm
@@ -84,6 +84,7 @@
 	taste_multi = 1.5 // stop sugar drowning out other flavours
 	nutriment_factor = 10
 	taste_description = "sweetness"
+	var/blood_gain = 0.2
 
 /datum/reagent/consumable/virus_food
 	name = "Virus Food"


Lumipharon:
Afaik sugar helps blood loss by providing nutrients, which helps blood regen.

Andrewhedge:
yeah I always thought the process was [ sugar -> nutrient -> blood ] ? am I wrong?

code/modules/reagents/reagents/food.dm
```diff
@@ -84,6 +84,7 @@
 	taste_multi = 1.5 // stop sugar drowning out other flavours
 	nutriment_factor = 10
 	taste_description = "sweetness"
+	var/blood_gain = 0.2
```
DeltaFire15, line 87: besides this change being questionable in itself,
this doesn't even work.. the var existing does nothing as blood_gain handling is a ../reagent/consumable/nutriment/.. thing whereas sugar is not a type of that (which is why you had to put the var/ before blood_gain so it compiled, and should also be getting a "var not used" warning

TiviPlus:
> Afaik sugar helps blood loss by providing nutrients, which helps blood regen.

correct

PR #9704, opened 2022-02-26
Fixes the bouncy revolver

Lumipharon:
## About The Pull Request
Currently the bouncy revolver 'bounce' shots don't come from the turf they hit, they come from the firer themselves. I'm not entirely sure how long this has been the case...
Fixes this so that bonus projectiles are fired from the location of the original projectile when the proc is called.

This also makes the proc more flexible for other uses, without breaking buckshot and the like. Not sure if I should make the origin point an arg instead, to allow for even more options. Maintainers let me know what you think.

Also if anyone can explain how pixel location magic works, it would be great to use the exact pixel location of the origin projectile instead of just the turf location.

## Why It's Good For The Game
Bouncing bullets now actually work.

## Changelog
:cl:
fix: Bouncy revolver bullets now actually bounce from what they hit
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -188,7 +188,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 			new_angle += 360
 		else if(new_angle > 360)
 			new_angle -= 360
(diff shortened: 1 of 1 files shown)


Lumipharon:
Wild unga shoots himself with bouncing bullet.
https://user-images.githubusercontent.com/7869430/155830091-42784308-a3e7-476e-9727-61f6a15a5fb4.mp4

novaepee:
The number one reason why things don't get fixed:

People don't use it.

PR #9708, opened 2022-02-27
Arrowhead Energy Shield

hyper2snyper:
## About The Pull Request
Adds a new Jaeger/Xenonaut module: The Arrowhead Energy Shield
The way it works:
Once attached it recharges at a set rate until it reaches maximum health.
The shield absorbs all damage until its health pool is expended.
Once all shield health is expended, the shield must undergo a reboot, afterwords it will recharge again.
Shield is dark red when its health percentage is at 0 - 33.
from 33 - 66 its light red.
66 to 100 its light blue.

Damage does not apply mobs armor until shield is dropped

Current stats:
Max Shield Health : 40
Recharge Rate per 2 seconds: 8
Broken Shield Reboot: 5 Seconds

The shield is available roundstart.

image
image
image
image

## Why It's Good For The Game

David wanted halo eshields.
This provides ability for light skirmishers to take almost zero damage without being overpowered in extended combat.

From David:
For a few reasons, as this solves one of the main Marine problems that being splints, as you can now finally splint yourself then fight a skirmish or 2 without having to constantly splint yourself.
Additionally, as what most people are forgetting, this takes place of either valk or tyr which are passive mods, and this is a pure health shield, unmodified by body armor, thus allowing Xenos to attack it directly, and in most cases, this shield goes out in either 2-3 hits, thus not stopping you from being hurt while stunned or otherwise.
All this does is stop that one annoying spitter or runner from deleting you before you get a chance to react, but not stopping any damage from when you get stunned, in which case tyr or even valk may be better for the long run.

## Changelog
:cl:
add: New Armor Module: The Arrowhead Energy Shield
/:cl:


zee11zee:
That sounds cool cant wait

Lumipharon:
So it removes all armour fullstop, or just from the actual armour it's attached to? Does that mean you'll get decapped when it runs out? And it does nothing against gas? Cool either way though.

hyper2snyper:
> So it removes all armour fullstop, or just from the actual armour it's attached to? Does that mean you'll get decapped when it runs out? And it does nothing against gas? Cool either way though.

only removes armor of the thing its attached too
helmet will keep its armor and so will uniforms/gloves and others

Mechanelize:
USE THE COOL EVASION BLUR SPRITE THIS LOOKS LAME

jroinc:
Eight recharge rate per two seconds is similar to being on Russian Red at all times (that heals 9 points per two seconds, as we subtract the cloneloss).

I have a feeling that this will be strong on light armor and OP on heavy armor.

ilikesaying:
we really going the route of Halo

Notamaniac:
No slowdown? Doesn't this make it strictly better than the baldur and tyr for a roundstart module?

OpsCoretango:
roundstart???
wow but for balance roundstart avalaible only for FC, SL, Corpsman

zee11zee:
> roundstart??? wow but for balance roundstart avalaible only for FC, SL, Corpsman

Also CPT don't leave out CPT

Kuro020 (approved):

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,142 @@
...
+	///Spark system used to generate sparks when the armor takes damage
+	var/datum/effect_system/spark_spread/spark_system
+
+	///Shield color when the shield is 0 - 33% full
+	var/shield_color_low = COLOR_MAROON
+	///Shield color when the shield is 33 - 66% full
+	var/shield_color_mid = COLOR_MOSTLY_PURE_RED
+	///Shield color when the shield is 66% to full
+	var/shield_color_full = COLOR_BLUE_LIGHT
+
```
TiviPlus, line 233: these dont need to be vars

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,142 @@
...
+	///Delay it takes to start recharging again once the shield is completely broken.
+	var/broken_shield_charge_delay = 5 SECONDS
+	///Cooldown used to determin when the shieild should start charging once more.
+	COOLDOWN_DECLARE(broken)
+
```
TiviPlus, line 238: ideally shield_broken_cooldown or such

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,142 @@
...
+		return
+	if(slot != SLOT_WEAR_SUIT)
+		UnregisterSignal(equipper, COMSIG_LIVING_SHIELDCALL)
+		STOP_PROCESSING(SSobj, src)
+		equipper.remove_filter("eshield")
+		shield_health = 0
+	else
+		if(COOLDOWN_CHECK(src, broken))
+			START_PROCESSING(SSobj, src)
+			playsound(equipper, 'sound/items/eshield_recharge.ogg', 40)
```
TiviPlus, line 277: early returns

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,142 @@
...
+			if(0.33 to 0.66)
+				affected.add_filter("eshield", 1, outline_filter(1, shield_color_mid))
+			if(0.66 to 1)
+				affected.add_filter("eshield", 1, outline_filter(1, shield_color_full))
+		spark_system.start()
+	else
+		shield_health = 0
+		STOP_PROCESSING(SSobj, src)
+		addtimer(CALLBACK(src, .proc/begin_recharge), broken_shield_charge_delay)
+		COOLDOWN_START(src, broken, broken_shield_charge_delay)
```
TiviPlus, line 309: early return

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,142 @@
...
+		parent.say("Warning: Shield is down! Rebooting in [broken_shield_charge_delay/10] seconds!")
+		return shield_left * -1
```
TiviPlus, line 311: Why is this a say

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,142 @@
...
+	playsound(wearer, 'sound/items/eshield_recharge.ogg', 40)
+	START_PROCESSING(SSobj, src)
+
+
+/obj/item/armor_module/module/eshield/process()
+	var/new_shield_health = shield_health + recharge_rate
+	if(new_shield_health > max_shield_health)
+		new_shield_health = max_shield_health
+	shield_health = new_shield_health
+	if(shield_health == max_shield_health)
```
TiviPlus, line 328: min exists

TiviPlus (changes requested):

TiviPlus:
Also as a side note isnt there a way to make outlines more faded and less poppy like this one is

code/modules/clothing/modular_armor/attachments/modules.dm
hyper2snyper: david wants to add different versions later on, this is so we can have different colors depending on shield type

code/modules/clothing/modular_armor/attachments/modules.dm
hyper2snyper: cuz i like machinery doing talking

hyper2snyper:
> Also as a side note isnt there a way to make outlines more faded and less poppy like this one is

i like poppy

code/modules/clothing/modular_armor/attachments/modules.dm
hyper2snyper: I like machinery talking

FosterGrime-Heymynameslouie:
speed mehrg

Mannybrado:
wow! a secondary health pool that recharges automatically forever, has no effect on your character, and you get your normal armor when its depleted! surely this will not crack balance like an egg... surely. it is COOL though, that's what matters right?

ThelastArchiver:
i have to ask this does it recharge the shield constantly or only out of combat?

hyper2snyper:
> i have to ask this does it recharge the shield constantly or only out of combat?

so the way it works: it will recharge constantly. if the shield takes damage to the point of depletion, there will be a 5 second delay before the shield is allowed to start charging again. If it takes damage to depletion when it starts charging again, the 5 second delay will be applied once more

hyper2snyper:
> wow! a secondary health pool that recharges automatically forever, has no effect on your character, and you get your normal armor when its depleted! surely this will not crack balance like an egg... surely. it is COOL though, that's what matters right?

cool is what matters after all

bluezorua:
I feel this is too powerful roundstart, maybe req only? hell maybe even research locked?

Lumipharon:
If it had some sort of negative effect when the shield breaks, it would have a genuine trade off. I.e. a very short stagger/slow when it breaks, or something.

David-Stormwell:
@TiviPlus 
### Why It's Good For The Game

For a few reasons, as this solves one of the main Marine problems that being splints, as you can now finally splint yourself then fight a skirmish or 2 without having to constantly splint yourself.
Additionally, as what most people are forgetting, this takes place of either valk or tyr which are passive mods, and this is a pure health shield, unmodified by body armor, thus allowing Xenos to attack it directly, and in most cases, this shield goes out in either 2-3 hits, thus not stopping you from being hurt while stunned or otherwise.
All this does is stop that one annoying spitter or runner from deleting you before you get a chance to react, but not stopping any damage from when you get stunned, in which case tyr or even valk may be better for the long run.

Kuro020 (changes requested):
in retrospect, the xenonauten sprite for this module is not up to standards, it needs a redo

ThelastArchiver:
does the shield protect from gas like boiler acid gas and does the sentinel spit still slow when hitting the shield?

David-Stormwell:
> 

Instead of saying it needs a redo comment what about it needs a redo, and provide more useful information to Hyper.

SevenHaven:
Hi! I just played a round of this.

The shield negates most xeno openers and allows a marine to have more health than a xenomorph.

Marines are already faster than xenomorphs. (Depends on setup.)

I would suggest limiting the shield to blocking a single attack or making heavy armor a requirement for it to be used.

With all due respect;

BRUH.

Otterwatch:
Having seen it in action. This allows marines to completely invalidate hit and runs so that only gas will do anything to a group larger than maybe 6. First round in use, xenos had to xenoball and were lucky at the very end that marines just tried to walk into 7 xenos by themselves 15 times just trickling in. Xeno and marine players were in agreement that shield is way too strong as a free item and should at least be a pricey req item (some said 80 for it's near b18 qualities for giving free health). I believe this module should definitely have a downside if it continues. It's downside shouldn't be that it takes place of other modules because it is straight up better than those modules as valk is only good for autosplinting which this fixes the issue for and tyr gives armor but this GIVES MORE STRAIGHT UP HEALTH. This module should definitely give slowdown and reduce your armor by 25-50%. The alternative is to keep it as is and make it an upper priced req item (not normal module price). Literally every marine will take this over any other module because it's free and it's better than the rest. 


Also, for some bugs. 
1. This module blocks pounce in the same way as a normal shield. 
2. Warrior Punch seems to go through the shield. (I can get a video of it if you want)


I actually think this module would be much cooler if it didn't give a health pool and just had one free stun block instead. That way you can still have the gimicky halo sound effects and visual stuff and it's a decent module but it isn't over the top powercreep. Just a thought.

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,153 @@
 	chemsystem.RemoveComponent()
 	return ..()
 
+
+/obj/item/armor_module/module/eshield
+	name = "Arrowhead Energy Shield System"
+	desc = "A brand new innovation in armor systems, this device creates a shield around the user that is capable of negating all damage. If it sustains too much it will deactivate, and, due to the install requirements, leave the user completely vunerable."
```
XSlayer300, line 210:
```suggestion
	desc = "A brand new innovation in armor systems, this module creates a shield around the user that is capable of negating all damage. If it sustains too much it will deactivate, and, due to the install requirements, leave the user completely vulnerable."
```

jroinc:
After testing this, I can confirm that it's quite strong on heavy armor as well.

Firefox13:
Maybe add a little cooldown not only when depleted, but also damaged (but should be very short, much less than a second)

jroinc:
apparently if you remove the armor via removing your undersuit, the 40-point shield stays. It won't re-appear when broken, not sure if it recharges.

ThelastArchiver:
id recommend adding a delay to recharging after taking damage like 5 seconds without taking damage

ilikesaying:
>id recommend adding a delay to recharging after taking damage like 5 seconds without taking damage

Just like the Real John Halo

ThelastArchiver:
i mean there is a reason in videogames why your shield only recharges when not taking damage and its called balance.

FosterGrime-Heymynameslouie:
Is the cooldown time supposed to be 15 minutes or 15 seconds

ilikesaying:
~~give deathsquad energy shields to make them more death!!~~

Al-1ce:
is busted but at least it'll prevent people from using other types

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -204,6 +204,153 @@
 	chemsystem.RemoveComponent()
 	return ..()
 
+
+/obj/item/armor_module/module/eshield
+	name = "Arrowhead Energy Shield System"
```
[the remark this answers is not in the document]
XSlayer300: >:(

Victor239:
This invalidates the type of hit-and-run attacks that xenos are designed for, particularly wraiths and runners. Ravagers additionally don't get rage from hitting the shield, and it blocks a number of other engage-type abilities.

Kuro020 (approved):

ilikesaying:
doesn't seem to recharge after depletion, only when taken off an re-equipped

FosterGrime-Heymynameslouie:
The shield dosent seem to regenerate.
Getting damaged wont regen shield health and once it gets depleted, the cooldown timer counts down but dosent reset the shield. Taking it out and putting it back in seems to be the only wy for it to reset.

MiniMeatwad:
Merge conflict

Round ID: 14308
BYOND Version: 514.1575
Server revision compiled on: 
Master commit: [4117fe048576249241b0bacb55c379730f75f75e](https://github.com/tgstation/TerraGov-Marine-Corps//commit/4117fe048576249241b0bacb55c379730f75f75e)
The following pull requests are currently test merged:
[#10017: 'Adds the T-100Y howitzer and rebalances the mortar, renamed mortar.' by Pariah919 at commit b667312bfa](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10017)
[#10032: 'Plasma Cutter Change' by Arctrooper8 at commit ee845aa771](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10032)

Local commit: 38722fd15375815160b652ba0bd5cd3ecdd81ac0
Server tools version: /tg/station 13 Server v3.2.6.2

BraveMole (approved):

PR #9709, opened 2022-02-27
Nerfs Sentinel's neurotoxin spit

lKiseki:
## About The Pull Request
Per title. List of changes:
- Reagent transfer reduced from 4 > 5 > 6 > 6.5 to 2.5 > 3 > 3.5 > 4.
- On-Hit stamina damage reduced from 40 > 30.
- Slowdown removed from direct hits.

## Why It's Good For The Game
Sentinel's neurospit has been really overwhelming due to its powercreeped stats. This PR aims to amend that.

## Changelog
:cl: Lewdcifer
balance: Sentinel neurospit nerfs. Reagent transfer 2.5 > 3 > 3.5 > 4; on-hit stamina damage 40 > 30; slowdown removed.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1722,12 +1722,11 @@ datum/ammo/bullet/revolver/tp44
 	max_range = 10
 	accuracy_var_low = 3
 	accuracy_var_high = 3
-	damage = 40
+	damage = 30
 	stagger_stacks = 1.1
-	slowdown_stacks = 1.5
 	smoke_strength = 0.5
 	smoke_range = 0
-	reagent_transfer_amount = 1.5
+	reagent_transfer_amount = 2.5
 
 ///Set up the list of reagents the spit transfers upon impact
 /datum/ammo/xeno/toxin/proc/set_reagents()
@@ -1791,15 +1790,15 @@ datum/ammo/bullet/revolver/tp44
 
 /datum/ammo/xeno/toxin/upgrade1
 	smoke_strength = 0.6
-	reagent_transfer_amount = 2
+	reagent_transfer_amount = 3
 
 /datum/ammo/xeno/toxin/upgrade2
 	smoke_strength = 0.7
-	reagent_transfer_amount = 2.5
+	reagent_transfer_amount = 3.5
 
 /datum/ammo/xeno/toxin/upgrade3
 	smoke_strength = 0.75
-	reagent_transfer_amount = 3
+	reagent_transfer_amount = 4
 
 
 /datum/ammo/xeno/toxin/heavy //Praetorian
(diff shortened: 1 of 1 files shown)


Lumipharon:
Pretty sure he meant direct (stam) damage on hit, not the transfer, I unno.

lovegreenstuff:
stagger spam, direct stamina damage that scales with maturity, slowdown...

yeah lets nerf the reagent transfer amount, that'll really fix the problem

EvilMelonTheSecond:
Like lovegreenstuff said I believe the real problem is the combination of stagger+slowdown+stamloss every hit. I would recommend removing stagger fully and maybe turn down some slowdown or stamloss

Lefunny:
lfmao thats a salty pr for sure, guy just saw numbers and nerfed anything in his way

jroinc:
This appears to remove any sort of meaningful damage from sentinel spit.

Also this *is* a reflex-driven salt PR.

Derrick2Electricboogaloo:
B-B-B-Based? Stagger is annoying.

Naaanii:
Too many nerfs. I'd suggest starting out small. Powercreep bad but gutting a T1 so harshly is just as bad.

Maybe just remove stagger for now I'd say.

Redstonize:
Check out Ravager, I'd say it's pretty OP right now, even wearing Tyr-2. I suggest comparing this to what you're trying to do.

MiniMeatwad:
Marines have a large variety of different ways to deal stagger or some other debuff preventing fighting at full strength, don't mind the Xenos having a debuff ability which only targets one person.

Especially with weapons like the Tesla rifle which deals AoE stagger in a large radius (unless they changed that recently), and the plasma loss gas grenade spam. If a nerf is justified, try removing toxin damage and oxy damage from the spit, since that actually does something aside from making the target less powerful.

Ex: You could try giving it a maximum stamina damage so it can't cause toxin damage from stamina > tox conversion.

Redstonize:
Nevermind what I said about prae, after a few days I've experience full scale of what Sent stun entails, and let me tell you about it; It isn't fun for anyone except if you like delaying rounds and playing cbt with marines.
Sentinel is the only target for this PR because it is the only xeno that's difficult to fight with this feature.
Prae has it, but it's counter-balanced by being large and slow, and having other abilities to use which means you won't necessarily play cbt with marines by force.
Sentinel should be removed any kind of stamina damage, because it is like a slow and steady, insidious stun spamming.
There is no way to counter the stamina damage taken, not even a sort of module or pills can do it. Hypervene won't save you.
That should be saying something: don't leave the stamina damage AS it is or make it any worse than it already is, because nothing can be used to counter it, not even Synaptizine which does nothing, and Hypervene which is supposedly the worst powerful chem you could use.

PR #9388, opened 2022-01-24
TX-8 Scout Rifle impact ammo rebalance

lKiseki:
## About The Pull Request
Per title.

Stats have been changed as follows;
- Penetration reduced from 45 > 30. Seriously.
- Stagger and knockback removed.
- Shots will now pierce through targets.

## Why It's Good For The Game
TX-8 impact ammo has been really dumb for too long. Stagger, slowdown, and knockback on a gun with 25 damage, 45 penetration, and 5 sunder per bullet.

## Changelog
:cl: Lewdcifer
balance: TX-8 impact ammo; 45 > 30 pen, stagger and knockback removed, shots pierce through targets.
/:cl:

PR #9713, opened 2022-02-28
Tx8 Impact Ammo Re-Rebalance

EvilMelonTheSecond:
## About The Pull Request
Adds back stagger to TX8 impact ammo, increases damage by 15 and removes penetration trait

## Why It's Good For The Game
TX8 rebalance PR fully butchered the niche of the higher cost impact ammo by removing both its higher damage and the crowd control capabilities of it. The penetration added in exchange is frankly obsolete on a gun with full auto with lower end DPS.

This PR adds back the stagger and increases damage so it isn't out DPS'd by the cheaper ammo giving it the crowd control niche back partially. This probably isn't enough to make it viable but I'm no numbers expert, feel free to recommend stats.

## Changelog
:cl:
balance: TX8 impact ammo stagger readded, damage increased by 15, penetration removed.
/:cl:


Naaanii:
Stagger is a bad mechanic and shouldn't be viable or readded. Just increase the damage but the weapon still kind of feels fine. Just keep the increased damage if needed.

Luxyboo:
stagger is a great mechanic and should be viable for every weapon that lost it

lKiseki:
This is a very silly PR. Not only are you readding back what was arguably the most broken part prior to its nerf in #9388, you're also increasing its damage by a whopping 15. This is just plain powercreep.

KrabSpider:
This is literally just the regular ammo with stagger. Stop.

EvilMelonTheSecond:
TX8 is a req gun with DMR levels of firerate and mag capacity, the damage being so low makes it a terrible choice and the stripped stagger made it fully useless. You lads are freaking out as if I'm buffing T12 ammo and not a 5 req points per 20 ammo magazine gun.

Araulius:
> TX8 is a req gun with DMR levels of firerate and mag capacity, the damage being so low makes it a terrible choice and the stripped stagger made it fully useless. You lads are freaking out as if I'm buffing T12 ammo and not a 5 req points per 20 ammo magazine

it's 7 points per magazine, actually.

lKiseki:
> TX8 is a req gun with DMR levels of firerate and mag capacity, the damage being so low makes it a terrible choice and the stripped stagger made it fully useless. You lads are freaking out as if I'm buffing T12 ammo and not a 5 req points per 20 ammo magazine gun.

For good reason, my dude. The TX-8 being a requisitions gun does _not_ justify it being as broken as you intend it to be. The damage isn't even that low, it has a lot of armor penetration and its stats are good enough. To compare stats to the T-12, which is practically the baseline for all guns;
image

It easily deletes armor and ignores a good bunch of it already. Stagger denies the usage of abilities and having it be applied on every bullet is unhealthy for balance.
If you want to buff it, look into other alternatives instead of basically reverting changes. There's a reason why those changes were made.

EvilMelonTheSecond:
It has been made clear this is not the correct rebalance needed. I may be defeated now but I will return stronger and with a much dumber PR.

PR #9718, opened 2022-03-01
Removes ship cameras from Big Red underground complex

RipGrayson:
## About The Pull Request

Apparently command can look through cameras located in the underground complex and see what the xenos are up to. 
Since the xenos never base anywhere but SE complex, command can essentially tell hive composition, resin silo location etc from the safety of their ship without even dropping shutters.

## Why It's Good For The Game

This isn't intended behavior, marines shouldn't be able to peep on xenos half a map away in an underground silo within 30 seconds of mission start.

## Changelog
:cl:
balance: Marines can no longer spy on xenomorph activity in Big Red's underground complex.
/:cl:


vilereaver41:
oopsie

PR #9726, opened 2022-03-05
Improves Lil-Canterbury ship to be less siegeable

vilereaver41:
## About The Pull Request

Makes Cantebury more defendable without wanting to change to much. added defenses instead of marines required to make them. and changes a few sections

## Why It's Good For The Game

Because marines are boneheads in crash, this will give them more prep time instead of worrying about defenses especially on low pop. the way i see it sieging the Canterbury should be the last thing on the xeno bucket list to do and should only be siege able when the entire hive is there.

often times there are blind spots where you cant get good views of xenos but I added some areas that are good for defensive points via windows + lasers.

This should atleast give marines a chance to fight back instead of lose hope.

## Changelog
:cl:
balance: redid sections of Canterbury
/:cl:

image

### DIFF
--- a/_maps/shuttles/tgs_canterbury.dmm
+++ b/_maps/shuttles/tgs_canterbury.dmm
@@ -10,45 +10,49 @@
 /turf/open/floor/plating,
 /area/shuttle/canterbury/cic)
 "ad" = (
-/obj/structure/table/mainship,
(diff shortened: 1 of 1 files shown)


vilereaver41:
closed because i dont want to remap every single god damn crash point to make this work, FUCK

PR #9727, opened 2022-03-05
Removes Big Red underground complex

RipGrayson:
## About The Pull Request
Title.

## Why It's Good For The Game

Big Red has become stale, the things marines and xenos do each round is almost exactly the same across multiple rounds. 
This is largely because there's no reason for xenos to base anywhere but SE caves and no reason for marines to land anywhere but LZ1, so all combat follows the same flow from round to round.

By removing the underground complex the xenos will have to base in labs or other parts of caves once more, like they did before the underground complex was a thing. 
Before the xeno mains bring out the pitchforks I would like to state that I'm not against making labs or other areas of the map more xeno friendly to compensate here, I'm just trying to avoid a situation where the hive must effectively defend the same routes over and over again for eternity.

Lastly before it's brought up "Why don't you fix LZ2 so marines don't have to use LZ1 all the time if you want more variation?" I would like to respond, I'm working on that idea, but it's a different PR.

EDIT: I have been given go ahead by the original maker of the SE complex to remove it, they don't like it either.

## Changelog
:cl:
balance: The underground complex in Big Red no longer exists.
/:cl:

### DIFF
--- a/_maps/map_files/BigRed_v2/BigRed_v2.dmm
+++ b/_maps/map_files/BigRed_v2/BigRed_v2.dmm
@@ -11694,11 +11694,8 @@
 /turf/open/floor/tile/yellow/full,
 /area/bigredv2/outside/general_store)
 "aSW" = (
-/obj/machinery/atmospherics/pipe/simple/green/hidden{
-	dir = 6
-	},
-/obj/structure/cable,
-/turf/open/floor/mainship/sterile/dark,
+/obj/docking_port/stationary/crashmode,
+/turf/closed/mineral/bigred,
 /area/bigredv2/caves/southeast)
 "aTa" = (
 /obj/effect/landmark/weed_node,
@@ -19463,8 +19460,7 @@
 /turf/open/floor/tile/dark,
 /area/bigredv2/outside/office_complex)
 "bHg" = (
-/obj/item/paper/clockresearch1,
-/turf/closed/wall/r_wall,
+/turf/closed/mineral/bigred,
 /area/bigredv2/caves/southeast)
 "bHH" = (
 /obj/effect/attach_point/electronics/dropship1,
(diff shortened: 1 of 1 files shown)


RipGrayson:
Here's a rough layout of how combat in Big Red currently goes
bigred

usnpeepoo:
CAS buff.
👎

RipGrayson:
> CAS buff. -1

There's just as much cave to hide in as before, just no complex. For that matter, all the labs count as underground so CAS isn't (more of) an issue when it comes to basing.

Luxyboo:
dont get rid of gen store, cargo, and bar 
also you are making big red into small red. maybe just make this version of big red a low pop variation for stuff like crash

Lumipharon:
Players are going to always do what they perceive to be the 'best' course of action. Removing the 'best' spot just means they will just find a new 'best' spot, and marines will determine a new 'best' route to it. And the cycle will continue.

RipGrayson:
> dont get rid of gen store, cargo, and bar also you are making big red into small red. maybe just make this version of big red a low pop variation for stuff like crash

Wait what? I didn't touch any of that stuff.
The only thing gone is the south east underground complex, and even then the total amount of cave is still the same.

RipGrayson:
> Players are going to always do what they perceive to be the 'best' course of action. Removing the 'best' spot just means they will just find a new 'best' spot, and marines will determine a new 'best' route to it. And the cycle will continue.

This sort of thing is to be expected, what we have going on in Big Red is that one area is so obviously superior to the rest in terms of defense that nobody ever does anything different.
Before this complex was added xenos might have had their favored spots, but they didn't monopolize one area exclusively, at least not to point of every operation being the same.

EvilMelonTheSecond:
I fully agree with the reasoning of the removal. Silos being in map corners is expected but having an extra fortified corner just removes any chance of other spots being used. There are two maze like labs in the map that xeno can utilize and giant cave systems to fall back to but they're all obsolete compared to the steel hellbox in the furthest corner the map extends to.

ThelastArchiver:
i also agree with this pr but another reason for this spot being favored is that it is the furthest away from the lzs with lets say lz 2 was moved to the top right that would probably lead to the chossing of other spots for the silo.

RipGrayson:
> i also agree with this pr but another reason for this spot being favored is that it is the furthest away from the lzs with lets say lz 2 was moved to the top right that would probably lead to the chossing of other spots for the silo.
LZ2 is another PR entirely, but I am looking into either moving it or fixing it so marines aren't consigned to using a single LZ over and over again.

Lefunny:
Xenos are going to always get the better place for silo placing, no changes at all this is straight up a nerf

RipGrayson:
New orders from the top, Tivi would prefer if I modularized the caves instead of an outright deletion. So this PR is going on ice, thank you for the feedback everyone.

PR #9684, opened 2022-02-22
Updating Squad Leader Vendor

novaepee:
## About The Pull Request

List of things that I have done to SL's vendor:
- Remove wack smokebomb that does nothing but blind marines for bad concealment.
- Remove choice to purchase M40 HEDP grenade and CAS flare box since why would they need to buy this when they can go to armament vendor
- Remove Station bounded radio and ziptie box since SL never use this
- Change placement of certain stuff like whistle, fulton, deployable overwatch camera, signaler, and tarp to make sense of list
- Update flamethrower on SL; I might just remove this since this is an old and forgotten relic

## Why It's Good For The Game

Clean up SL's vendor. Simple enough.

## Changelog
:cl:
del: Remove choice to get smoke grenade from SL vendor
del: Remove choice to get M40 HEDP grenade from SL vendor
del: Remove choice to get CAS flare box from SL vendor
del: Remove Station bounded radio and ziptie box since SL never use this
add: Change placement of certain stuff like whistle, fulton, deployable overwatch camera, signaler, and tarp to make sense of SL vendor
balance: Update flamethrower on SL; I might just remove this since this is an old and forgotten relic
/:cl:

PR #9729, opened 2022-03-06
Gives the SL the compass

Arctrooper8:
## About The Pull Request

https://github.com/tgstation/TerraGov-Marine-Corps/pull/9684 removes squad leader radio from vendor but not essentials. Elsa's reasoning is that no one uses it, therefore I trust that removing it from the essential is also wise (I have also not seen anyone use it, if you have please tell me). However the main point of this PR is giving the SL a compass as part of their essentials. Yes compasses are free but by making it part of the essentials it will be brought more to their attention.

## Why It's Good For The Game

The compass gives coordinates (the type you use for a drop pod) and is a mostly unused item. Putting it directly in the hands of  SLs makes it a more known tool and allows SLs to easily give coordinates to those wishing to drop pod right into combat.

## Changelog
:cl:
balance: replaces SL essentials kit radio with a compass
/:cl:

### DIFF
--- a/code/__DEFINES/loadout.dm
+++ b/code/__DEFINES/loadout.dm
@@ -710,7 +710,7 @@ GLOBAL_LIST_INIT(loadout_role_essential_set, list(
 		/obj/item/explosive/plastique = 1,
 		/obj/item/beacon/supply_beacon = 2,
 		/obj/item/whistle = 1,
-		/obj/item/radio = 1,
+		/obj/item/compass = 1,
 		/obj/item/binoculars/tactical = 1,
 		/obj/item/pinpointer/pool = 1,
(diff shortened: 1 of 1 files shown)


Squigger:
Just use an antenna on your helmet, or the webmap?

Arctrooper8:
> Just use an antenna on your helmet, or the webmap?
Fair point.
However, the antenna on your helmet takes up a useful module slot that is better taken by Mimir, Tyr, or something else. And the webmap requires you to have your browser open, which if it isn't you then have to spend time navigating.
The overall goal of the PR is to hopefully make a little used tool more visible to those who can best utilize it. It is much faster to use a compass than to check a webmap.

PR #9739, opened 2022-03-07
Resin Turrets Same Price as Acid

Otterwatch:
## About The Pull Request
Title explains it

## Why It's Good For The Game
Resin Turrets can currently be spammed out and are especially egregious at lower population levels when it takes an entire marine push many minutes just to push past 1-2 of these turrets. They slow you down to ridiculous speeds and allow you to get warrior/shrike/crusher displaced extremely easily for a cheap cost. You can also double up or triple up on them in the same area which will outright kill a marine if he gets stunned into them. This is about the same effect as an acid turret, but these actually deplete stamina. Therefore, I believe that these turrets should be the same price as acid turrets which is 100. Currently double resin turrets current price, but they are extremely effective. They can help xenos take out jetpack marines more effectively than acid can and they are very useful for chokepoints. For example, here are three images where lowpop xenos just spammed them to great effect. Marines got all the way to silo and crumbled due to 5 in the same area with shrike and warrior using them to kill marines.

resin turrets

resin

res

## Changelog
:cl:
balance: Made resin turrets 100 psypoints just like acid turrets.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
@@ -206,7 +206,7 @@ GLOBAL_LIST_INIT(tier_to_primo_upgrade, list(
 	name = "Sticky resin turret"
 	desc = "Places a sticky spit spitting resin turret under you. Must be at least 6 tiles away from other turrets, not near fog and on a weeded area."
 	icon = "resinturret"
-	psypoint_cost = 50
+	psypoint_cost = XENO_TURRET_PRICE
 	turret_type = /obj/structure/xeno/xeno_turret/sticky
 
 /datum/hive_upgrade/xenos


Naaanii:
75 points would be a better compromise so they can't be spammed as much and won't be completely unused because acid turrets are always the better alternative if they're priced the same.

Lumipharon:
Resin turrets are hugely underrated, the stamage damage is huge, especially with how strong neuro spit is currently stacking with it. Double is quite a lot though, go to 75 and it can always be revisited.

Otterwatch:
I think xenos being able to create a silo and 10 turrets of their choosing from the get go is fine. Resin turrets are currently fairly strong and can be used to combo with xeno abilities very nicely. This is still a lethal turret and it slows to the point that the only effective way of taking it out is standing from a far and sniping at it or nading and hoping shrike doesn’t do you in with either of those scenarios. Either way, 75 or 100 is still avg 10 less turrets that can be placed roundstart. 50 is just extremely cheap for the utility. I don’t think making it 100 points will make it any worse than acid still. Both have their uses and I find resin works much more effectively. It definitely doesn’t care about armor or modules or chems for that matter. Resin is currently underrated and should be used more tbh.

Lefunny:
Resin turrets are fair easy to be taken down

BraveMole:
Making turret price scale in reverse with population seems a better idea imo

Ragecrabbo:
> Making turret price scale in reverse with population seems a better idea imo

having certain psi point purchases scale with pop or something similar would honestly work really well. Stuff like primordials is rarely worth it in lowpop while in highpop its almost mandatory. Same goes for turrets where they become less useful the higher the population goes as few turrets are easy targets for a big marine murderball while on a lower pop all it takes is a smart shrike and a few other xenos to completely wipe a push by tossing them into turret hell.

PR #9745, opened 2022-03-08
Surt slowdown reduction

Arctrooper8:
## About The Pull Request

Changes surt slowdown from .4 to .3 (PR was originally .4 to .2) the light change was moved to a different PR.

## Why It's Good For The Game

Surt is a 12 point module that only gives you almost (you still take minor burns standing in it) fire immunity but slows you down to a crawl when using it. The only other module with .4 slowdown is the free Tyr mk1 which gives 10 armor to every category. Whereas Surt costs 12 points for armor only against fire. Those 12 points could also be used for 15 armor and .3 slowdown, autosplinting and automedication, or .2 slowdown and immunity to explosives. I am aware these changes are controversial but I feel they're worth a TM at most.

Of course community/maintainer opinion is greatly appreciated 

## Changelog
:cl:
balance: Surt is now faster having .3 slowdown instead of .4
/:cl:


Otterwatch:
Looks like power creep for a module that has downsides for a reason. Fire is one of the most effective tools against xenos and being able to rush them with flame retardant gear and zone the area without being really slowed or anything else is very powerful. Light change makes sense but slowdown change is too much.

Lumipharon:
> Looks like power creep for a module that has downsides for a reason. Fire is one of the most effective tools against xenos and being able to rush them with flame retardant gear and zone the area without being really slowed or anything else is very powerful. Light change makes sense but slowdown change is too much.

For reference, the slowdown of a T12 is 0.4.
The slowdown of a flamer is 1.75, the same as as the rocket launcher.

So while I agree with what you're saying, you're realling not rushing anywhere with a flamer. And if you're using surt without a flamer, good for you I guess, but you're not getting the most out of it.

Otterwatch:
> So while I agree with what you're saying, you're really not rushing anywhere with a flamer. And if you're using surt without a flamer, good for you I guess, but you're not getting the most out of it.

Except for the fact that using a flamer is second rate to using a miniflamer or plasma pistol in the current state of the game. Miniflamers can flame multiple wall layers and repulse xenos at the same, so normal flamers are effectively a thing of the past or for xfuel if you can ever get some. Iirc, you can even connect a miniflamer to a back fueltank unless that got removed. There is no reason not to use a surt with current meta if this pr goes through. The only other competitor is valk for autosplints, but underbarrels that cause flames are used to great effect every single round and are meta. Buffing that to this degree is frankly silly. I still stick by that light buff should stay since it doesn't really make a difference (you can take a pocket pistol or railflashlight and belt harness to the same effect) and slowdown buff is too strong. Slowing them down allows xenos to actually counter and be able to avoid being unga rushed by a single man with a ubf or plasma pistol in a xeno maze.

Lumipharon:
Mimir is the best upgrade right now due to the strength of neuro, objectively so.

Miniflamer does no on hit damage as that was fixed, so is utterly garbage at wall clearing and significantly worse at actually hurting xenos. It's only good for area denial. And flamers in general were hugely nerfed (indirectly) by the diagonal movement fix, as now you can just side step fire tiles in most situations.
Plasma gun is a direct hit thing, it's valuable because of the 40 damage on hit and having a big capacity compared to an undershotty, and as above, you can just dodge the fire tiles as marine or beno so being fire immune there isn't really a big deal.

Otterwatch:
> Mimir is the best upgrade right now due to the strength of neuro, objectively so.
> 
> Miniflamer does no on hit damage as that was fixed, so is utterly garbage at wall clearing and significantly worse at actually hurting xenos. It's only good for area denial. 

So, first of all. Mimir is mediocre because neuro currently can easily be countered by a hypervene autoinjector or dylo premed. Mimir is better just for the general gas protection. Even then, you can just do what I stated above to run through gas with near zero downsides. Dylo under 10 units doesn't prevent stamina regen. 
Secondly, Miniflamer still does damage on hit like every other flamer? He only removed the bonus damage 2x modifier for it to double damage.  Also, what's the downside to using a miniflamer along with your primary weapon at the same time compared to using a normal flamer? The problem I have with this pr Lumi is that you can chase down benos as if you were using any other module and go through fire with ease while doing it. There is zero reason to get any other module besides surt and valk with this change. The current meta is regarding flames on underbarrel which shred xenos in every context besides resin jelly. UBF can still shoot a stream through walls and be connected to a backpack fuel tank. Also, you can kill an ancient crusher by flaming it repeatedly for the hit damage in 12 seconds with the ubf (no other damage source) while just flaming it once and it standing on said flames takes 35 seconds. You can kill the same ancient crusher in 6 seconds with t12 burst and ubf hit damage while it takes ~7-8 seconds to kill it with one flame hit and t12 burst. Speaking of which, using a plasma pistol on a t35 with bayo and stock with flechette on the same Ancient Crusher allows you to kill it in t12 assault rifle record time of around 8 seconds including crit finish. All of this above points to flame type weapons being extremely powerful and being able to trivialize the effect flames have on marines (which is devastating at times) is nonsensical. They already have extinguishers and meds and all sorts of other ways to deal with fire and keep their speedy moves. There's absolutely no reason to allow them to just get a 12 point module that let's them ignore everything about fire besides the hit damage which they won't be subject to anyway. They just have to pop a kelo, tram, and maybe tric if you're feeling spicy and you're fine to chase beno into maze with nearly zero reprecussions  compared to setting yourself on fire normally when you flame the xeno ahead of you. Diagonals have made flames a little less effective but mazes are still usually 2x1s that are unaffected by this change and still force xenos to run through fire and take damage that you can run through just as fast as them if you med it right with the same amount of light you normally would have. I have spent many hours of my life testing to see if this could work as a good idea and I think the pr currently needs an adjustment on the slowdown. Fire is just too powerful currently to buff users. Surt users are still common and I see many "robust" players using it daily despite its shortcomings.

Naaanii:
2 buffs at once isn't really good for a module that still has a use and is still alright despite being underused. I'd say either buff the light range or slowdown.

Light range would probably have less of an impact and slowdown would be better to tm and see how powergamers abuse said buffs.

Lumipharon:
I agree with what you're saying, although plasmapistols are more popular now than ubf for a reason.

The slowdown definitely shouldn't LESS than valk though.

OpsCoretango:
i want to see surt in engineer vendor on crash mode. Like on wiki with Xenonauten-m with engineer storage module with surt and TL-84.
Mmua,perfecto!

Arctrooper8:
> Light range would probably have less of an impact and slowdown would be better to tm and see how powergamers abuse said buffs.

A good idea, because the light thing was a strange and bizarre nerf from the past that is more annoying then nerf. The slowdown buff is where the real contention and buff lies. I'll probably do it after maintainers look it over.

Also, thanks for all the feedback!

Arctrooper8:
> i want to see surt in engineer vendor on crash mode. Like on wiki with Xenonauten-m with engineer storage module with surt and TL-84. Mmua,perfecto!

Very tempting, I'll ask balance heads in the discord if it's a good separate PR idea.

BraveMole:
Surt is less used because flamer is less meta. It is still a very good module nonetheless

Squigger:
"Oh but X exists so we should buff Y" isn't a good reason. Powercreep PR.

TiviPlus:
Light is better option

PR #9755, opened 2022-03-10
Essentials kit Quality of life

Arctrooper8:
## About The Pull Request

Stops the vendor from giving FC unneeded gloves and a useless radio.
And essentials kits no longer come with an excess coif when the map is cold.

## Why It's Good For The Game

Right now essentials kit code gives a radio and gloves to all marines, however the FC already has these which just results in extra clutter for them. And there's no reason to give a coif to marines, this isn't CM any mask works to stop the cold. And the vendors already have a mask tab, alongside fobs having extra mask crates for those who do forget.

## Changelog
:cl:
qol: Removed the extra coif that is given with essentials on cold maps.
qol: FC is not given an extra pair of gloves and a worse radio anymore.
/:cl:

The compass removal is because that's the loadout vendor, which should only be giving items that aren't free and infinite (like the compass) just another coding mistake of mine I fixed.


Lumipharon:
Wait, why is this headband literally as good as an actual helmet? And why are you giving a 'field commander' item to SL's?

Arctrooper8:
> Wait, why is this headband literally as good as an actual helmet? And why are you giving a 'field commander' item to SL's?

I'm not giving anything to SLs unless I made a coding error (I will go check). And the headband is the same stats as beret (which is also helmet value, and of course FC has access to the better leader helmet type), if the FC can have a beret with helmet stats I see no harm in giving them a headband with helmet stats too.

Arctrooper8:
> Wait, why is this headband literally as good as an actual helmet? And why are you giving a 'field commander' item to SL's?

I checked the code and did very minor testing, I don't believe I am giving anything to SLs but if you could point out which item I would be grateful.

Lumipharon:
Nah you just referred to the SL getting it in your text or I imagined you did, code is all good.

TiviPlus (changes requested):
these changes need to be atomized

Arctrooper8:
> these changes need to be atomized

Doing it!

PR #9785, opened 2022-03-14
at gun changes

Pariah919:
## About The Pull Request
APHE now passes trhough turfs and not mobs, and has a slower shell speed

## Why It's Good For The Game
Gives people more time to react, and the pass through mob thing was kinda dumb, honestly.

## Changelog
:cl:
balance: APHE on AT Gun now flies slower, and now passes through turfs/walls and not mobs.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1328,8 +1328,8 @@ datum/ammo/bullet/revolver/tp44
 	icon_state = "atgun"
 	hud_state = "shell_heat"
 	hud_state_empty = "shell_empty"
-	flags_ammo_behavior = AMMO_EXPLOSIVE|AMMO_ROCKET|AMMO_SUNDERING|AMMO_PASS_THROUGH_MOVABLE
-	shell_speed = 3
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

Victor239:
So it can one-shot xenos hiding behind walls? This sounds like a worse change tbh. It seems more problematic that it was available for free, rather than costing points.

novaepee:
> So it can one-shot xenos hiding behind walls?

If you can see a wounded xenomorph behind walls, sure.

> This sounds like a worse change tbh.

False statement. The APHE has a bug where it multiplies damage with AMMO_PASS_THROUGH_MOVABLE, making it cause more damage than needed.

> It seems more problematic that it was available for free, rather than costing points.

Change it if you want.

Lumipharon:
I should add btw, I think I no what the actual bug is.
Rockets explode on hit with everything, but once something is hit, there's some code to stop them from hitting them again. BUT they also explode on hitting their target turf.

So if you shoot a beno and click directly on them/their turf, the projectile:
1) Hits the mob causing direct damage,
2) causes an explosion on the mob, causing more damage,
3) then hits the turf, exploding a 2nd time on the same spot

PR #9787, opened 2022-03-14
Give req a single free sidecar upgrade for motorbike

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

Tivi makes good content, but marines never buy this. They are too busy getting SADAR, B18, autosniper, AT guns, specialized ammo, and weapon systems instead of transport systems. Heck, if I see marines buying more motorbikes, I would be downright surprised. I beomce even more surprised when I see a motorbike sidecar since someone REALLY want this.

It is true that marines can buy the sidecar upgrade since they are given that option, but given the fact that marines have better options, the sidecar upgrade will not see the day of light. The frequency of choosing this item is further decreased since the only upgrade is having two people riding the motorbike. The driver must use two hands, and the rider one hand, meaning that one person is there for the ride.

## Changelog
:cl:
add: Give req a single free sidecar upgrade for motorbike
balance: Give req a single free sidecar upgrade for motorbike
/:cl:

### DIFF
--- a/_maps/map_files/Minerva/TGS_Minerva.dmm
+++ b/_maps/map_files/Minerva/TGS_Minerva.dmm
@@ -6900,6 +6900,10 @@
 	dir = 1
 	},
 /area/mainship/living/evacuation)
+"xb" = (
+/obj/item/sidecar,
+/turf/open/floor/mainship/mono,
+/area/mainship/squads/req)
 "xd" = (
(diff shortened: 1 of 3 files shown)


novaepee:
welp, map conflict is back

Luxyboo:
iirc sidecar makes to where you can only have one hand, maybe remove that to make it better.

PR #9790, opened 2022-03-15
Updates leader essentials kits, FC has specific Jaeger kit.

Arctrooper8:
## About The Pull Request

Gives the FC an essentials kit specifically for jaeger and removes the useless motion detector from FC and SL essentials kits.
000FC
Also renamed the jaeger essentials paths since jaeger is no longer the only modular suit.

## Why It's Good For The Game

Less clutter in the FC's office if they don't want to use Jaeger, and considering that motion detectors are a free and infinite gun attachment that people barely ever use in hand anymore (let alone actually call out the xeno about to ambush everyone) so having it in an essentials kit is somewhat silly.
And having the typepath be named basicmodular is an outdated way since the majority of marine armor is modular these days (also smushing words together makes me sad).

## Changelog
:cl:
add: FC now has a specific essentials kit for jaeger
del: SL and FC no longer start with an extra motion detector
code: renamed jaeger essentials kits pathtype to better convey what they are to coders
/:cl:

### DIFF
--- a/code/__DEFINES/loadout.dm
+++ b/code/__DEFINES/loadout.dm
@@ -229,7 +229,7 @@ GLOBAL_LIST_INIT(loadout_linked_vendor, list(
 
 GLOBAL_LIST_INIT(marine_clothes_listed_products, list(
 		/obj/effect/essentials_set/basic = list(CAT_STD, "Standard Kit", 0, "white"),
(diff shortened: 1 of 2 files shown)


Arctrooper8:
Child of: https://github.com/tgstation/TerraGov-Marine-Corps/pull/9755

PR #9794, opened 2022-03-16
Make coverage code actually work

Lumipharon:
## About The Pull Request
Coverage code now actually works.

What is coverage code? Why it's the code that allows structures such as window frames, tables, barracades and the like to possible block bullets.

Currently it is utterly messed up, cades only work due to counteracting jank code with more jank code, while everything else either never blocks bullets, or blocks bullets 100% of the time.

This change means coverage is actually accurate to what it does.
**The vast majority of machinery and structures can now be shot past with varying degrees of reliability**

Window frames have a 50% base chance to block bullets for example, while a table is only going to block 10% of shots.

If your accuracy is over 100, coverage will be decreased by the amount your accuracy exceeds 100. i.e. a bullet with 120 accuracy only has a 30% chance to be blocked by a window frame.
This means you can now shoot past cades facing you if you have extremely accurate weapons (the odds may be very low however).

Snipers, rockets and IFF weapons have also lost their cover ignoring magic, however they do still retain a bonus against cover (and tend to all have very high accuracy on top of this).

Also makes flipped tables actually do something. Currently they... do absolutely nothing in terms of blocking bullets, they don't block any better than a non flipped table. Now they will be decent barricades, of course with such low HP, they'll only block 1 or 2 bullets before exploding.

Finally increased window frame hp. A while back they were changed to be hittable with melee, and combined with this change they're actually super fragile considering they're a block of metal or whatever. In particular, 'reinforced' windowframes had no extra hp.

## Why It's Good For The Game
Code that works is good/
Makeshift barricades are cool, and it makes defending buildings etc actually do something. This is probably a slight buff to benos since, you now, marines all use guns.
You can now recreate shootouts as the ungapolice try attack you in the bar.

## Changelog
:cl:
refactor: Coverage now works - window frames and other obstacles can actually block bullets, certain objects no longer always block bullets
balance: Snipers, rockets and IFF weapons no longer completely ignore cover
balance: Slightly buffed window frame HP, in particular reinforced window frames
balance: Flipped tables now actually block bullets better than non flipped tables
/:cl:


novaepee:
I'm going to hate all windows now.

Lumipharon:
> I'm going to love all windows now.

Fixed that for you :^)
Also SG's ignore cover magically, and I haven't nerfed that yet so you're fine.

Ragecrabbo:
stealth roony buff

MiniMeatwad:
oh

Lumipharon:
Fixed metal foam allow bullets through it, and gave target stakes a 90% coverage change so you can actually hit them properly.

Shout out if you find any more wonky behaviour please.

Lumipharon:
Fixed for turrets/deployables, reduced cover for miners since they're indestructable.

TiviPlus (approved):

PR #9796, opened 2022-03-16
Double Acid Turret

chesse20:
## About The Pull Request

dreamseeker_EFAfqqJiq7
Double acid turret, it shoots twice as fast but has less range due to being clunky. Costs 150 points instead of 100 to compensate. Since it has less range you can shoot it from the edge of the screen without a scope and it can't shoot back 

## Why It's Good For The Game

Xeno winrate is down the drain and players keep asking for xeno content so i made some

## Changelog
:cl:
add: Double headed acid turret that spits twice as fast with less range, watch out!
/:cl:


Naaanii:
Kind of overkill since Xenos already tend use regular turrets as a killer comboing with displacement abilities like Shrike and Warrior.

Lumipharon:
You can setup killboxes with multiple turrets all shooting at the sam ungas, so it's not really needed.

ThelastArchiver:
this highly depends on the hive having capabale warriors and shrikes its also probably really easy to take out because of its limited range this seems reasonably balanced. i think the phrase high risk high reward first very well here.

Pill-U:
tgmc turning into PvZ

AnCopper:
> tgmc turning into PvZ
marinez

Kuro020 (changes requested):
side sprite is way too similar to the normal acid turret

hyper2snyper:
ima tmc this

Kuro020:
> > side sprite is way too similar to the normal acid turret
> 
> new double turret image vs single image better?

nope, the shape is too similar

chesse20:
any ideas on how to make it look different the perspective is hard

Pill-U:
perhaps make it a different color make it easier to recognize like the resin turret? maybe red to emphasize on how dangerous it is or green because of all the acid inside?

vilereaver41:
Sorry gona down vote. If xeno winrates are down this feels like a forced buff that is randomly stitched together we dont need double turrets we need a stat rework

TiviPlus:
@Kuro020 review sprite

Kuro020:
> any ideas on how to make it look different the perspective is hard

try to separate both uh turret mouths more, make one go on top and the other one below it on the side sprite

chesse20:
image

PR #9801, opened 2022-03-17
Keybinds supported for all actions

Walarks:
## About The Pull Request
Makes keybinds assignable to all actions by moving the associated code from xeno actions to base actions. Previously they were assignable only to xenos or by making custom code.

Makes actions not need an object as a target. The target is currently only used for copying its icon to act as an overlay for the button. 

Improves(?) vali action code. Makes it reuse the valk scan action instead of its own unique action since they are identical in function.

## Why It's Good For The Game
Makes it easier to assign keybinds to actions.

## Changelog
:cl:
code: keybinds are easily assignable to all actions now
qol: adds keybind to vali configure action, moves connecting a weapon from the configure menu to its own action
/:cl:


private-tristan:
yes yes yes yes I can do SO MUCH WITH THIS YES YES YES

TiviPlus (approved):

PR #9818, opened 2022-03-20
Removes ballistic module slowdown

x64SLUMLORD:
## About The Pull Request

Changes the ballistic module from 0.2 slowdown to 0.

## Why It's Good For The Game

Ballistic is a very straightforward module that's supposed to dampen damage from FF. It's still affected by armor penetration, shrapnel and fractures, and takes up your armor's module slot. It's rendered less useful due to the fact that it has a significant slowdown for 40 more bullet and laser armor. It should be more on par with Baldur's simplicity and something newer players can use without significant downside for not that impactful of an upside. It also has a cool sprite.

## Changelog
:cl: Slum Lord
balance: ballistic armor no longer slows you down
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -175,7 +175,7 @@
 */
 /obj/item/armor_module/module/ballistic_armor
 	name = "\improper Hod Accident Prevention Plating"
-	desc = "Designed for mounting on modular armor. A substantial amount of additional reflective ballistic armor plating designed to reduce the impact of friendly fire incidents, will lessen the affects of bullets and lasers. Will impact mobility."
+	desc = "Designed for mounting on modular armor. A substantial amount of additional reflective ballistic armor plating designed to reduce the impact of friendly fire incidents, will lessen the affects of bullets and lasers. Will not impact mobility."
 	icon = 'icons/mob/modular/modular_armor_modules.dmi'
 	icon_state = "mod_ff"
 	item_state = "mod_ff_a"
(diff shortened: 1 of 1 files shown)


Naaanii:
No module should have 0 slowdown and with FF already deescalated this buff seems rather unnecessary.

Losenis:
Ballistic module offers absolutely no benefit whatsoever against xenos, it sholdn't have a slowdown. The ability to magdump xenos while a shotgun marine goes in without fear of FF is in theory ok but in practice either never happens due to coordination needed, the slowdown itself making it hard if not impossible, or just has all the bullets blocked by the guy anyway. The additional slowdown may also have you lag behind the others making you miss opportunities to fight xenos due to carrying a module that doesn't help you vs xenos in the first place.

EvilMelonTheSecond:
> No module should have 0 slowdown and with FF already deescalated this buff seems rather unnecessary.

This would've been correct if both vali and baldur didn't get powercrept to have 0 slowdown. While I do agree that modules shouldn't be "free picks" but instead trades, having already two good free picks makes that point null. Unless game direction turns back to no free modules, this PR is the correct way to make ballistic module have some competing chance against others.

x64SLUMLORD:
Ballistic is ran once in a blue moon by either a new marine that doesn't grasp that slowdown can and will get you killed, someone who is REALLY tired of FF, or me (I run it solely for fashion). I don't understand the logic behind it's fairly significant slowdown for an upside that doesn't quite match. Baldur already has 0 slowdown for a fairly eh benefit and Ballistic is pretty much at the same tier.

While I do understand the concept of 'no module should be free', I feel it's best to group up the weak modules into easy choices for people that don't want to run the more useful ones.

It also has a really cool sprite.

TiviPlus:
Powercreep

Losenis:
>Not getting killed by your own teammates is powercreep

PR #9824, opened 2022-03-21
Tesla gun rebalance

goose4429:
## About The Pull Request

About The Pull Request
Tesla gun is a rarely used weapon, highly underestimated by marines. But in fact when it is used, it can have unexpectedly high impact. Current effects are:
- Skill block for xenos for LONG time
- Strong plasma drain with effect lasting for 10 seconds, draining 60% of your max plasma
- Slowdown, which usually makes difference between whether you are faster than marine or not
- All this with abysmally large effect range
- But also slow speed and firerate

This PR is trying to make tesla gun less painful for xenos and more fun for marines.

## Why It's Good For The Game

Before, one pew pew from marine could break Queen Screech assault by blocking skills of every xenos in large radius. Even more painful, it made possible to hunt xeno by having few marines by your side when you shoot the tesla to hamper xeno retreat who are slowed and cannot use skills to escape tesla's large effect radius.

Now the tesla has less one-time impact, but steadier and more reliable performance. It's not a xeno-bully weapon, but a disable weapon making beno retreat faster. Ammo efficiency improved to make it more viable. Added strong on-hit plasma drain to reward hits (before you'd be better off by NOT hitting the beno).

## Changelog
:cl:

balance: Decreased plasma drain effect time from 10 to 2
balance: Decreased plasma regen stop time from 10 to 2
balance: Plasma drain now drains 10% per tick (from 6%)
balance: Removed stagger
balance: Increased ammo per cell from 4 to 6
added: on-hit plasma drain (30% of max)

/:cl:

### DIFF
--- a/code/datums/beam.dm
+++ b/code/datums/beam.dm
@@ -174,7 +174,7 @@
 			continue
 		source.beam(living, icon_state="lightning[rand(1,12)]", time = 3, maxdistance = zap_range + 2)
 		if(living.xeno_caste.caste_flags & CASTE_CAN_BE_GIVEN_PLASMA) //need 1 second more than the actual effect time
-			living.apply_status_effect(/datum/status_effect/noplasmaregen, 5 SECONDS/length(.))
-			living.apply_status_effect(/datum/status_effect/plasmadrain, 5 SECONDS/length(.))
+			living.apply_status_effect(/datum/status_effect/noplasmaregen, 3 SECONDS)
+			living.apply_status_effect(/datum/status_effect/plasmadrain, 3 SECONDS)
 		living.add_slowdown(2)
 		log_attack("[living] was zapped by [source]")
--- a/code/datums/beam.dm
+++ b/code/datums/beam.dm
@@ -176,5 +176,5 @@
 		if(living.xeno_caste.caste_flags & CASTE_CAN_BE_GIVEN_PLASMA) //need 1 second more than the actual effect time
 			living.apply_status_effect(/datum/status_effect/noplasmaregen, 5 SECONDS/length(.))
 			living.apply_status_effect(/datum/status_effect/plasmadrain, 5 SECONDS/length(.))
-		living.adjust_stagger(0.75) //stacks for every zap
+		living.add_slowdown(2)
 		log_attack("[living] was zapped by [source]")
--- a/code/datums/beam.dm
+++ b/code/datums/beam.dm
@@ -173,9 +173,8 @@
 		if(living in blacklistmobs)
 			continue
 		source.beam(living, icon_state="lightning[rand(1,12)]", time = 3, maxdistance = zap_range + 2)
-		if(living.xeno_caste.caste_flags & CASTE_CAN_BE_GIVEN_PLASMA)
-			living.apply_status_effect(/datum/status_effect/noplasmaregen, 10 SECONDS/length(.))
-			living.apply_status_effect(/datum/status_effect/plasmadrain, 10 SECONDS/length(.))
-		living.adjust_stagger(1)
-		living.add_slowdown(2)
+		if(living.xeno_caste.caste_flags & CASTE_CAN_BE_GIVEN_PLASMA) //need 1 second more than the actual effect time
+			living.apply_status_effect(/datum/status_effect/noplasmaregen, 5 SECONDS/length(.))
+			living.apply_status_effect(/datum/status_effect/plasmadrain, 5 SECONDS/length(.))
+		living.adjust_stagger(0.75) //stacks for every zap
 		log_attack("[living] was zapped by [source]")
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -202,7 +202,7 @@
 /datum/status_effect/plasmadrain/tick()
 	var/mob/living/carbon/xenomorph/xenoowner = owner
 	if(xenoowner.plasma_stored >= 0)
-		var/remove_plasma_amount = xenoowner.xeno_caste.plasma_max / 17
+		var/remove_plasma_amount = xenoowner.xeno_caste.plasma_max / 10
 		xenoowner.plasma_stored -= remove_plasma_amount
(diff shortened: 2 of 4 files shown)


Naaanii:
Would be better to honestly keep it back as req only now that the strength is realised and let marines stick to T-gas as the budget Tesla.

goose4429:
> Would be better to honestly keep it back as req only now that the strength is realised and let marines stick to T-gas as the budget Tesla.

Probably a good idea as well. Make it reasonably expensive, and stick it there. Still needs some reworks though.

Mrrpip:
> Would be better to honestly keep it back as req only now that the strength is realised and let marines stick to T-gas as the budget Tesla.

Meh, might as-well mention while we're at it, that the Tesla can randomly spawn at T1 weapon sites on some maps(Most frequently seen on Icy Caves).

So whatever you guys do, keep that in mind, yeah?

EvilMelonTheSecond:
Nerfing EVERY stat is a bit much, the gun barely has a niche even as a freebie. It is only effective is open areas with nothing to stop the projectile, can escaped by moving a few tiles away, deals no damage and is limited by line of sight. If nobody can deal with a gun in that narrow of a niche just send it to req hell where nobody will ever see it again.

goose4429:
> Nerfing EVERY stat is a bit much, the gun barely has a niche even as a freebie. It is only effective in open areas with nothing to stop the projectile, can escaped by moving a few tiles away, deals no damage and is limited by line of sight. If nobody can deal with a gun in that narrow of a niche just send it to req hell where nobody will ever see it again.

It is indeed much less effective in tighter spaces. What do you think should be changed instead? This is my first take on this stuff, and I'd be glad to hear what you'd do

EvilMelonTheSecond:
> > Nerfing EVERY stat is a bit much, the gun barely has a niche even as a freebie. It is only effective in open areas with nothing to stop the projectile, can escaped by moving a few tiles away, deals no damage and is limited by line of sight. If nobody can deal with a gun in that narrow of a niche just send it to req hell where nobody will ever see it again.
> 
> It is indeed much less effective in tighter spaces. What do you think should be changed instead? This is my first take on this stuff, and I'd be glad to hear what you'd do

Another idea for balance might be to give it IFF in exchange for stat nerfs to broaden the niche while taking down the effect. I'm not a good at number balancing though so I would recommend balance testing at either side of extremes and adjusting as time shows.

goose4429:
> > > Nerfing EVERY stat is a bit much, the gun barely has a niche even as a freebie. It is only effective in open areas with nothing to stop the projectile, can escaped by moving a few tiles away, deals no damage and is limited by line of sight. If nobody can deal with a gun in that narrow of a niche just send it to req hell where nobody will ever see it again.
> > 
> > 
> > It is indeed much less effective in tighter spaces. What do you think should be changed instead? This is my first take on this stuff, and I'd be glad to hear what you'd do
> 
> Another idea for balance might be to give it IFF in exchange for stat nerfs to broaden the niche while taking down the effect. I'm not a good at number balancing though so I would recommend balance testing at either side of extremes and adjusting as time shows.

That sounds like a very good idea - it both makes it less cumbersome to use, and more viable as a support weapon.

goose4429:
I've received some feedback. Increased range to 3. Gave the gun built-in IFF because why the hell not, it's a support weapon. Added on-hit stagger and slowdown.

OpsCoretango:
oh no, balance moment

TiviPlus (changes requested):
Your changelog is not player legible

I have some concerns:
Why make it do more damage and effects on hit, I neither see the point or benefot
Why add IFF
Why make the shell move faster (especially since you say you want consistent performance yet a faster shell makes it less consistent)

goose4429:
> Your changelog is not player legible
> 
> I have some concerns: Why make it do more damage and effects on hit, I neither see the point or benefit Why add IFF Why make the shell move faster (especially since you say you want consistent performance yet a faster shell makes it less consistent)

1. More damage and effects just to encourage xenos to dodge the ball. In current state, you're better off by taking the ball on, because you'll be affected by less zap procs
2. I can't see how you can properly use the weapon without some kind of aim. To use it you need to be on the frontline, or have small number of marines nearby, like for shotguns of laser weapons. Laser weapons are used much less often than ballistic weapons. Shotguns are used because they have strong damage, and stagger. Tesla has stagger, but doesn't have the damage to back it up, so you must rely on your allies. And as described above, it conflicts with having no aim mode. I think having IFF or aim mode is necessary.
3.  This is the very reason I have started to try to rework tesla. Slow move speed allows for many many procs of zap(), which gives stagger and slow. Slow doesn't stack (but is still pain). However stagger _does_ stack, and personal experience + testing showed that you can get 10 seconds of stagger quite easily. And if it affects a sleeping (or otherwise standing still xeno), 15 seconds. Both of these are absurd, given it's 4 sec of reload.

All in all, I tried to conserve the idea of the weapon but making it more balanced and fun. Of these three changes, I think that:
Projectile speed - very important
IFF/aim - important, but not required
damage/on-hit effects - not important at all

I'd be glad to hear what you think

FosterGrime-Heymynameslouie:
speedmehrg

FosterGrime-Heymynameslouie:
> Would be better to honestly keep it back as req only now that the strength is realised and let marines stick to T-gas as the budget Tesla.

if this was to happen, CAS Tangle MIssle should be buffed back to 25 or 50 because the reason why it was nerfed in the first place was because Tesla gun was going to be free for all marines to use and it needed a counter balance, thus cas tangle got murked.

also speedmehgr based pr

AGasCanister:
image
Terra Experimental Tesla Shock Rifle, you were a great weapon, almost never used because no one acknowledged your existence, I used you several times and got some good assists, I will remember you. I drink water in your honor

Pariah919:
Concur with Tivi on this one.

FosterGrime-Heymynameslouie:
the IFF tesla gun was pretty fun
can it be split
like the standard tesla gun is the current one without IFF but higher plasma drain
but an "exprimental" version of the tesla gun that does have IFF with lower plasma drain be added to the vendors or req maybe

PR #9820, opened 2022-03-20
Remove razor wire from the game

BraveMole:
## About The Pull Request

Remove razorburn and related from the game

## Why It's Good For The Game

It's been abused for months by one player (thea), no one wants to fix it, admins chose to ignore it (this is a bug abuse and was stated as such for months).

This is slowing down the game to a crawl for the amusment of one player. If admins do not want to inteverne, I will.

If you are not happy with this, feel free to finally fix microdosing (Or maybe, finally warn thea to stop doing bug abusing?)

## Changelog
:cl:
del: Remove razor wire from the game
/:cl:

PR #9825, opened 2022-03-21
Fixes razorwire duplication exploit, adds ability to blacklist medical chems

RipGrayson:
## About The Pull Request
Acts as an alternative to https://github.com/tgstation/TerraGov-Marine-Corps/pull/9820

You had to know this was coming.
Right now the game suffers from a small group of players who have figured out how to duplicate razorburn and then microdose it. 
This PR patches the current methods being used and adds some future protections against this sort of thing.

To start off, reagents no longer transfer to dead mobs, this was being used to gas large groups of dead monkeys with nanites.
Sleepers no longer perform dialysis on monkeys at all, this was being used to transfer the nanite reagent to beakers.

I'm going to answer some questions here to save time.
Q. But what if they just start gassing live humans to extract nanite reagent?
A. Sleepers now purge any chemicals instead of removing them to a beaker.
Q. I've heard you can use cryo to duplicate chems, won't they just use that?
A. Blacklisting a chemical means cryo won't accept it either.
Q. If I _did_ get it, could I use it in a custom mix?
A. Medical machinery will refuse beakers with blacklisted reagents, no turning your pilfered 60u bottle of nanites into 10 6u beakers so groundside can be filled with razorwire. 
Q. Can I work with a blacklisted reagent at all?
A. Not if I have anything to say about it.

## Why It's Good For The Game

Razorwire spam has become an issue again lately, along with chemical duplication in general. Removing it will probably help our flagging xeno winrates and prevent more abuse. 
If this system has no holes in it, it can be rolled out to any other reagent that we don't want players touching.

## Changelog
:cl:
add: Added a system to blacklist chemicals from being used in medbay.
balance: Razorburn nanites are now a blacklisted chemical.
balance: You can no longer use the sleeper to give monkeys dialysis.
balance: Dead mobs no longer accumulate smoke or foam reagents.
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -237,7 +237,7 @@
 		if(istype(I, /obj/item/reagent_containers/glass/bucket))
 			to_chat(user, span_warning("That's too big to fit!"))
 			return
-		
+
 		beaker =  I
 		
 
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -224,6 +224,12 @@
 	. = ..()
 
 	if(istype(I, /obj/item/reagent_containers/glass))
+		
+		for(var/datum/reagent/X in I.reagents.reagent_list)
+			if(X.medbayblacklist)
+				to_chat(user, span_warning("The cryo cell's automatic safety features beep softly, they must have detected a harmful substance in the beaker."))
+				return
+
 		if(beaker)
 			to_chat(user, span_warning("A beaker is already loaded into the machine."))
 			return
@@ -231,10 +237,12 @@
 		if(istype(I, /obj/item/reagent_containers/glass/bucket))
 			to_chat(user, span_warning("That's too big to fit!"))
 			return
-
+		
 		beaker =  I
+		
 
 		var/reagentnames = ""
+
 		for(var/datum/reagent/R in beaker.reagents.reagent_list)
 			reagentnames += ", [R.name]"
 
--- a/code/modules/reagents/holder.dm
+++ b/code/modules/reagents/holder.dm
@@ -504,11 +504,9 @@
 /datum/reagents/proc/reaction(atom/A, method=TOUCH, volume_modifier=1, show_message = 1)
 	var/react_type
 	if(isliving(A))
-		var/mob/living/L = A
-		if (L.stat == DEAD)
-			return
 		react_type = "LIVING"
 		if(method == INGEST)
+			var/mob/living/L = A
 			L.taste(src)
 	else if (isturf(A))
 		react_type = "TURF"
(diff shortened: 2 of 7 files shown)


QualityVan:
Why not just kill dialysis storing the chems and have it purge them directly instead?

RipGrayson:
> Why not just kill dialysis storing the chems and have it purge them directly instead?

Not a bad idea, I was trying to avoid rocking the boat by making too many unneeded changes to functionality, but I can easily implement purging if the maintainers think it's the better solution.

Naaanii:
> Why not just kill dialysis storing the chems and have it purge them directly instead?

This seems like a much better method.

RipGrayson:
> > Why not just kill dialysis storing the chems and have it purge them directly instead?
> 
> This seems like a much better method.

Following the current consensus, I have adjusted the sleeper so it only purges chemicals instead of storing them in a beaker.

Ragecrabbo:
please merge this, you can dupe pretty much any chem with the monkey method mentioned above, including high value lemoline only chems (and lemoline itself)

TooFewSecrets:
> > > Why not just kill dialysis storing the chems and have it purge them directly instead?
> > 
> > 
> > This seems like a much better method.
> 
> Following the current consensus, I have adjusted the sleeper so it purges chemicals instead of storing them in a beaker.

Are you going to remove the medical nanites recipe because raw nanites are no longer obtainable anyway? (Not that there was ever a reason to make medical nanites when the vendors have an infinite quantity.)

RipGrayson:
> Are you going to remove the medical nanites recipe because raw nanites are no longer obtainable anyway? (Not that there was ever a reason to make medical nanites when the vendors have an infinite quantity.)

That's somewhat out of scope for this PR, maybe once it's merged I'll adjust medical nanites to make them attainable if maintainers want it.

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -224,17 +224,25 @@
...
 		if(beaker)
 			to_chat(user, span_warning("A beaker is already loaded into the machine."))
 			return
 
 		if(istype(I, /obj/item/reagent_containers/glass/bucket))
 			to_chat(user, span_warning("That's too big to fit!"))
 			return
-
+		
 		beaker =  I
```
TiviPlus, line 241:
```suggestion

		beaker =  I
```

code/modules/reagents/holder.dm
```diff
@@ -504,9 +504,11 @@
 /datum/reagents/proc/reaction(atom/A, method=TOUCH, volume_modifier=1, show_message = 1)
 	var/react_type
 	if(isliving(A))
+		var/mob/living/L = A
+		if (L.stat == DEAD)
+			return
 		react_type = "LIVING"
 		if(method == INGEST)
-			var/mob/living/L = A
 			L.taste(src)
```
TiviPlus, line 512: why?

TiviPlus (changes requested):

code/modules/reagents/holder.dm
RipGrayson: The check itself is needed to avoid corpses soaking up reagents (which was being used to dupe chems).
You would think the (isliving) check would prevent that but it doesn't, hence the L.stat == DEAD check.

RipGrayson:
Razorburn specifically will not accumulate in dead corpses, I considered removing razor nanite spread entirely but if I did that, marines would be able to open a Razorburn grenade and drop it at their feet with no consequences. This would lead to a marine defensive strategy of sitting inside an instantly deployed field of 7-8 squares of razor with complete freedom to shoot incoming xenos.

TiviPlus (approved):

PR #9826, opened 2022-03-22
Xeno ex act refactor

goose4429:
## About The Pull Request
 
Code for xeno handle of explosions was strange, and unconcise at times. This PR makes it more standardized, and very slightly nerfs explosions in one particular case.

## Why It's Good For The Game

It is now easier to change explosion behavior. 
It is also easier to understand how explosions work.
Finally, slightly nerfs light explosions on benos with no bomb resistance (most benos actually) so that they are in line with other cases in terms of stagger and slowdown.

## Changelog
:cl:
refactor: Refactored xeno explosion damage procs
balance: Slowdown from light explosions on weak benos reduced from 4 to 3 to be in line with other cases
balance: Stagger is now always 2 less than slowdown with xeno explosions
/:cl:

### DIFF
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -41,6 +41,6 @@
 #define SHIELD_PARENT_INTEGRITY (1<<2) //Transfers damage to parent's integrity.
 
 #define EXPLODE_NONE 0
-#define EXPLODE_LIGHT 3
-#define EXPLODE_HEAVY 2
 #define EXPLODE_DEVASTATE 1
+#define EXPLODE_HEAVY 2
+#define EXPLODE_LIGHT 3
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -41,6 +41,6 @@
 #define SHIELD_PARENT_INTEGRITY (1<<2) //Transfers damage to parent's integrity.
 
 #define EXPLODE_NONE 0
-#define EXPLODE_DEVASTATE 1
+#define EXPLODE_LIGHT 1
 #define EXPLODE_HEAVY 2
-#define EXPLODE_LIGHT 3
+#define EXPLODE_DEVASTATE 3
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -41,6 +41,6 @@
 #define SHIELD_PARENT_INTEGRITY (1<<2) //Transfers damage to parent's integrity.
 
 #define EXPLODE_NONE 0
-#define EXPLODE_LIGHT 1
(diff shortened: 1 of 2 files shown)


Lumipharon:
Won't this require all instances of explosions being called in the code to change to the new format (if the args haven't been explicitly named)?

goose4429:
> Won't this require all instances of explosions being called in the code to change to the new format (if the args haven't been explicitly named)?

1. I haven't change the args format, so I don't see how it would matter
2. This is only handler for explosions for xeno. If, for example, a wall is hit by TAT, wall's own explosion handler will be called, which is also untouched. Only xeno.
3. Only thing I've changed which has global effect is actual numbering of defines. But the only place that mattered is in the code I've refactored here. In all other places, only comparisons with global defines are called.

code/__DEFINES/combat.dm
```diff
@@ -41,6 +41,6 @@
 #define SHIELD_PARENT_INTEGRITY (1<<2) //Transfers damage to parent's integrity.
 
 #define EXPLODE_NONE 0
-#define EXPLODE_DEVASTATE 1
+#define EXPLODE_LIGHT 1
 #define EXPLODE_HEAVY 2
-#define EXPLODE_LIGHT 3
+#define EXPLODE_DEVASTATE 3
```
TiviPlus, line 46: This is tg side so for us to do this it should be done on tg first

TiviPlus (changes requested):

code/__DEFINES/combat.dm
goose4429: Done

code/__DEFINES/combat.dm
```diff
@@ -41,6 +41,6 @@
 #define SHIELD_PARENT_INTEGRITY (1<<2) //Transfers damage to parent's integrity.
 
 #define EXPLODE_NONE 0
-#define EXPLODE_DEVASTATE 1
-#define EXPLODE_HEAVY 2
 #define EXPLODE_LIGHT 3
```
TiviPlus, line 44: no reason changes

TiviPlus (changes requested):
Can you ping me on discord I have issues with the code

code/__DEFINES/combat.dm
```diff
@@ -41,6 +41,6 @@
...
+#define EXPLODE_HEAVY 2
+#define EXPLODE_DEVASTATE 1
```
TiviPlus, line 46:
```suggestion
#define EXPLODE_DEVASTATE 1
#define EXPLODE_HEAVY 2
#define EXPLODE_LIGHT 3
```

TiviPlus (approved):

PR #9830, opened 2022-03-23
King is now slightly less of a meme

Victor239:
## About The Pull Request

King is now cheaper (750 points) and marines are no longer notified about it's summoning (but can still locate the pod).

## Why It's Good For The Game

King is a meme at 1800 points thanks to there being lots of other good options (primordials, silos, spawners, turret spam) so it's been made cheaper. 

Also there is no longer a priority announcement for King summoning because it's not that dangerous a caste and it will usually be randomly given to a xeno who dies with it in 2 mins because they don't know it sucks in combat. 

## Changelog
:cl:
balance: King is now cheaper (750 points) and marines are no longer notified about it's summoning (but can still locate the pod).
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
@@ -217,7 +217,7 @@ GLOBAL_LIST_INIT(tier_to_primo_upgrade, list(
 	desc = "Places a Psychic Echo chamber that tallhosts can detect, then after a summon time selects a random sister to take over the mind of the gravity manipulating King."
 	icon = "king"
 	flags_gamemode = ABILITY_DISTRESS
-	psypoint_cost = 1800
+	psypoint_cost = 750
 
 /datum/hive_upgrade/xenos/king/can_buy(mob/living/carbon/xenomorph/buyer, silent = TRUE)
 	. = ..()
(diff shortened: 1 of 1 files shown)


Lefunny:
too spammable, make it stronger instead of cheaper

Mrrpip:
> Also there is no longer a priority announcement for King summoning because it's not that dangerous a caste and it will usually be randomly given to a xeno who dies with it in 2 mins because they don't know it sucks in combat.

I don't think I like the motive for the announcement priority removal. It seems too 'eh' IMO, even new queens/shrikes have tells (Hivemind not collapsing being one, but also the massive noise upon queen evolve)

Victor239:
> too spammable, make it stronger instead of cheaper

1800 points when there's so many other options would mean it would have to be ridiculously powerful considering it's given to a random xeno. Got any ideas for that then?

> I don't think I like the motive for the announcement priority removal. It seems too 'eh' IMO, even new queens/shrikes have tells (Hivemind not collapsing being one, but also the massive noise upon queen evolve)

Can marines tell that hivemind is collapsed? Also a noise for queen evolution is not the same as being given a massive universal priority announcement that the summoning is happening in order to prevent it. The King pod can still be located as a structure to destroy, it just won't be a notification.

Luxyboo:
> too spammable, make it stronger instead of cheaper

you cant make more than 1 king at a time 
also 750 is too much for instant cade break the least it should be is 1000 points

KrabSpider:
Maybe Kings should have a lot more abilities that fit into it being designed as the ultimate siege caste?

Naaanii:
This doesn't make King any more viable or even worth getting in any way honestly.

Pariah919:
What?
No.

vilereaver41:
Prob not a good idea, this will make early game king + silo and tower possible which would make early game very hard for marines

novaepee:
This is not the buff we want for king.

BraveMole:
Not a good idea fellah.

PR #9835, opened 2022-03-24
Nerf silos in nuclear war 

BraveMole:
## About The Pull Request

Small nerf (from 3x to 2.5) for silos.
And set spawner output to be the same as in distress

## Why It's Good For The Game

3 times is a bit too strong, 2.5 should be more balanced while still being a challenge for marines. Minions spam is boring to play, and far too strong

## Changelog
:cl:
balance; Silos in nuclear war only have a 2.5 spawn multiplicator, rather than 3
balance; Minions output in Nuclear War is now the same thing as in Distress

### DIFF
--- a/code/controllers/subsystem/monitor.dm
+++ b/code/controllers/subsystem/monitor.dm
@@ -72,10 +72,10 @@ SUBSYSTEM_DEF(monitor)
 	if(SSticker.mode?.flags_round_type & MODE_SPAWNING_MINIONS)
 		//Balance spawners output
 		for(var/silo in GLOB.xeno_resin_silos)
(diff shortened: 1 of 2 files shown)


Lumipharon:
Spawner rate is also pretty high, especially when the pop is a bit lower.

Lefunny:
lets go, time to make xenos weak in nuclear war just like they are in distress

Redstonize:
You seriously think 2.5 is enough ? I'd go for straight up 2.

BraveMole:
> You seriously think 2.5 is enough ? I'd go for straight up 2.

The bigger issue is prob spawners. I am fine with the game being harder with marines, they can just organise better for a change

PR #9846, opened 2022-03-28
TL-127 Renewal

Redstonize:
## About The Pull Request

TL-127 bolt action rifle has 5 more rounds per magazine. It's flak rounds still exist despite nobody using them. But we added hollow points, which basically deal a lot of sunder and has higher damage but AP is lowered, so each shot does incremental, time-bound damage.

The base accuracy gets dropped.

Aimed shots have a delay of 2 seconds in-between, while bipod would reduce it to 1.70 sec so we nuke the bipod (sundering works to create more damage the quicker you are.) You can now attach either a gyroscope stabilizer or an angled grip.

When hitting marines with hollow point, there's a 50% chance of generating shrapnel on hit.

Ammo magazines cost 7 points, an ammo belt costs ~~30~~ 40.

You will still most likely miss any unwielded shots 100% of the time.

## Why It's Good For The Game

Normally the TL-127 is detrimental to a marine's arsenal in the sense that it requires other weapons involved, not being a useful tool in good hands but rather a standard NPC weapon, so it's better to just forget sniping altogether if you are skilled (sniping is a weak niche for you) and go with another weapon.

This PR partially addresses that, by adding one alternate ammo type. It deals a minimum damage that's respectable for it's price. As it turns out, it still won't two-shot runners since the number of shots to crit remain the same for runners and the hunter, but since there's incremental shot damage (that is also time-bound) introduced by sunder, the more successive shots you hit in a small time frame, the more damage you deal with a minimum zone of 3-4 shots (so damage is higher vs healthier xenos for instance, on the 4th and 5th shots).

Also we are elongating the fire delay by 0.05 and shortening things like wield delay via angled grip or alternatively aimed movement speed with the gyro (that's not very significant on it's own).

You still won't be able to PB with it.

As a round goes on, assuming that it's being used, (and that you can avoid **friendly fire**) this will become a game-changer (meaning stalemate ending), as long as you're sniping. And even then there is margin of error, that's why it's balanced. 

## Changelog
:cl:
balance: TL-127 bolt action rifle magazines have 5 more rounds.
balance: The TL-127 flak (they're impact..) magazines exist and cost 5 points.
balance: You can purchase special TL magazines that do more damage, but costs 7 points.
/:cl:


Redstonize:
**In case you're wondering about before-and-afters: Here's how it affects gameplay:**
1. In any scenario where you fit some attachements to the gun, **you'll likely miss any unwielded shots**, which means if you fire the weapon before it is fully wielded, you waste an entire 1.4s - 2.0s.
2. If you try to PB with the weapon, it will not allow you to shoot, meaning **you cannot PB with this weapon,** only dealing melee damage which is around 45 brute with a bayonet.
3. Most weapons will take at least 3 to 5s to crit any xeno. Before, it would take you 7 shots to crit an ancient shrike with the TL-127, which because of the fire delay (1.35) **usually took about 9s. In the same amount of time, you could theoretically kill a shrike with only the bayonet**. Now, it takes 4 shots to crit with the weapon, but keep in mind it's fire dealy is 1.4s so **it takes 4,2s, and the shrike won't die yet, only crit.**

Lumipharon:
Uh, you can deploy a bipod anywhere btw.

Redstonize:
> Uh, you can deploy a bipod anywhere btw.

Well I certainly don't want to balance it around a single attachement if there are other attachments available.

I think angled grip is a good one, it makes you less prone to the typical T-60 with Vertical grip problem where you get molested for having been stunned once.

Naaanii:
But why?

Redstonize:
To be fair, I'm not really satisfied with how the TL-127 is.

Andrewhedge:
hmm..4 unga sniper team is really scary as shit, but ok. are you sure about sunder tho?

Redstonize:
> hmm..4 unga sniper team is really scary as shit, but ok. are you sure about sunder tho?

In practice, sunder makes a xeno's armor weak to projectiles.
While in reality a shell as large as a sniper caliber would definitely break up some armor, it would definitely feel wrong to give unecessary sunder to someone who just got his health cut by 25%. Sunder is a tool of punishment just like damage with good AP. Besides, it is only possible to give a bullet full damage (100%) after applying sunder, not 120% or 150%, so it's not gonna cut it unless we reduce the AP and make the damage even higher (this would theoretically two-shot rounies.)

InterroLouis:
This feels a bit like powercreep to me. We just buffed the Antimat, and now you want to make the TL-127 outperform it again?

Redstonize:
> This feels a bit like powercreep to me. We just buffed the Antimat, and now you want to make the TL-127 outperform it again?

The TL-127 won't outperform the antimat even with this.

KrabSpider:
> > This feels a bit like powercreep to me. We just buffed the Antimat, and now you want to make the TL-127 outperform it again?
> 
> If the TL with special mags outperforms the AMR it's an AMR problem. Perhaps the AMR is truly not a great sniper for the cost.

**No because that is a textbook case of powercreep. A roundstart sniper should not have an special ammo that puts it at competition with a 40 point req weapon, especially when anyone can  pick it up and sniper-ball. What the hell is wrong with you?**
🗞️

Redstonize:
> This feels a bit like powercreep to me. We just buffed the Antimat, and now you want to make the TL-127 outperform it again?

A "damage, cost and icons" commit is underway. I just need to get back on my PC.

TiviPlus:
Sniper is already a solid choice in coordination, and the offscreen plinkiness is intentional as offscreening is often unfun for the recieving end

PR #9857, opened 2022-03-30
Dragunov Special Ammo

Redstonize:
## About The Pull Request

Made to compete with https://github.com/tgstation/TerraGov-Marine-Corps/pull/9846 .

Adds a new magazine with special tipped ammo (soft-point ammo) that does more damage, costs 5 req points. Normal magazines are cheaper, and the SVD Dragunov is 20 points instead of 15.
Also it adds angled grip because the longer ranged scope adds more delay than the medium ranged scope which adds delay, and it's only fair with the req price of the weapon.
## Why It's Good For The Game

( People malded over #9846 so I made this, for the same reasons. )
Sniper has always been a worse niche when it comes to a good player's arsenal. There is no good sniper weapon at the moment.
A sniper should be playable other than by NPCs.
This should allow for the meta to change over from shotgun to sniper more nicely.
It's an alternate ammo thing that is req-locked, and since the weapon can be found in starter crates, I might change the randomization, but it definitely won't two-shot rounies even with special ammo.
It doesn't make the dragunov good in close quarters since you can't attach bayonet nor pb with it, but it certainly has an effect on frontlining, so even NPCs can kill an ancient shrike in 6 or 7 shots (7 or 8 shots with regular ammo), which was something of a skill issue before.

## Changelog
:cl:
balance: Dragunov now costs 20 points, it's magazines now cost 2 points. 
balance: New ammo that uses special tipped rounds for the Dragunov can be ordered from req for 5 points, they provide higher damage but no sunder and more falloff.
/:cl:

### DIFF
--- a/code/__DEFINES/calibers.dm
+++ b/code/__DEFINES/calibers.dm
@@ -51,6 +51,7 @@
 #define CALIBER_86X70 "8.6x70mm"
 #define CALIBER_10X99 "10x99mm"
 #define CALIBER_762X54 "7.62x54mm Rimmed" //SVD and mosin
+#define CALIBER_762X54_plus "7.62.54mm Rimmed" //SVD only
 #define CALIBER_557 ".557/440" //Martini Henry
 #define CALIBER_RAILGUN "rail projectile"
 
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -971,6 +971,16 @@ datum/ammo/bullet/revolver/tp44
 	penetration = 35
 	sundering = 15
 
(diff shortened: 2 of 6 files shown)


InterroLouis:
Maybe document that you're adding angled grip to the gun?

lovegreenstuff:
>made to compete with tl127 powercreep pr
yeah as if either of these two prs are ever gettin merged
c06

Redstonize:
> > made to compete with tl127 powercreep pr
> > yeah as if either of these two prs are ever gettin merged
> > c06

I have managed T-64 buffs to get merged before, don't know why either of these PRs wouldn't be possible unless the gun balance maintainer was against it.

Redstonize:
Focusing efforts on TL-127 instead.

PR #9862, opened 2022-03-30
Crusher unique explosion handler removal

goose4429:
## About The Pull Request

Crusher was the only xenos to have unique explosion handler, which overrides default behavior with 0 damage from light or heavy explosions, but doubled damage from devastating ones. This PR removes this handler. This code is likely a remainder from CM, since it wasn't ever changed by anybody.

## Why It's Good For The Game

Makes game more intuitive - game plays as you expect it to play. Crusher now can be damaged by nades, but is very resistant to them, as he still has tier 3 armor. He can also be damaged by all CAS weapons. For info - CAS machinegun and nades will deal 30 damage on average. 

## Changelog
:cl:
code: Removed ex act from crushher.dm
balance: crusher is no longer immune to some explosions
balance: crusher gets less damage from biggest explosions
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/crusher.dm
@@ -18,13 +18,6 @@
 	old_x = -16
 	old_y = -3
 
-/mob/living/carbon/xenomorph/crusher/ex_act(severity)
-
-	flash_act()
-
-	if(severity == EXPLODE_DEVASTATE)
-		adjustBruteLoss(rand(200, 300), updating_health = TRUE)
-
 
 /mob/living/carbon/xenomorph/crusher/handle_special_state()
 	if(is_charging >= CHARGE_ON)


Lefunny:
its missing the balancment cl

Wayland-Smithy:
1. Lefunny is correct this is a balance change and needs to be CLed as such.
2. Please do not remove crushers ex act immunity because they are slow and suffer enough from the displacement effects as is, and spinning under CAS railgun is of the few joys in my life.

goose4429:
Thanks to the guys above, fixed cl with balance

Naaanii:
Nerfing the most currently underwhelming T3 guh...

Crusher feels like it should be immune to explosions with the way it plays and even with this trait, it hasn't ever felt too oppressive. Maybe Gorger should have something similar actually.

Naaanii:
> > Nerfing the most currently underwhelming T3 guh...
> > Crusher feels like it should be immune to explosions with the way it plays and even with this trait, it hasn't ever felt too oppressive. Maybe Gorger should have something similar actually.
> 
> Doing this just for the sake of consistency. We can give Crusher ex immunity no problem. For info, usual nade will deal 30 damage on average, which is very small. No slowdown, no sunder.

Fair enough

TiviPlus (approved):
Merging for maintenance purposes if theres balance issues people can feel free to buff it with numbers

PR #9877, opened 2022-04-01
Lower the Cap needed for T1's to Mature.

vilereaver41:
## About The Pull Request

Lower's the cap a little so they mature a tad bit faster.

## Why It's Good For The Game

Often times a leader wont buy maturity tower's and even then as a new xeno you spend time doing nothing after evolving to a T1 until you mature instead of risk death, this should speed the process up a little bit. also there is no such thing as "Young" so yee.

## Changelog
:cl:
balance: lower maturity costs for t1's
/:cl:


lKiseki:
Should honestly just remove Young as a tier. It serves no purpose.

Naaanii:
Better solution might be to pre spawn in a maturity and evo tower per map so they still have a function and this won't feel like a double buff should Xenos decide to use them and would be a better buff sandstone nce T2s and T3s can benefit from them too.

PR #9894, opened 2022-04-02
Scope vision + offset now resets properly when ghosting/on death

breadhunt:
## About The Pull Request

Offset is now reset on reset_perspective(), and ghosting triggers reset_perspective() like it should.
Previously dying or ghosting or whatever while scoped/zoomed/whatever just straight up wouldn't reset it.

## Why It's Good For The Game

Not only is there no way to remove this (annoying) if you go into another body you keep the zoom and can move and shoot around with it. Sniper scope SADAR bad!!!!

## Changelog
:cl:
fix: Zoomed vision + offset now resets properly when ghosting/on death
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -634,7 +634,7 @@ below 100 is not dizzy
 		return
 
 	update_sight()
-	animate(client, time = 0, pixel_x = 0, pixel_y = 0)
+	animate(client, pixel_x = 0, pixel_y = 0)
 	if(client.eye && client.eye != src)
 		var/atom/AT = client.eye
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/living.dm
```diff
@@ -634,6 +634,7 @@ below 100 is not dizzy
 		return
 
 	update_sight()
+	animate(client, time = 0, pixel_x = 0, pixel_y = 0)
```
TiviPlus, line 637: Actually shouldnt this be an animate() with no args

breadhunt: it doesn't work without client, pixel+x, and pixel_y but i've removed time = 0. thanks

PR #9901, opened 2022-04-03
Adds balloon_alerts to lots of actions

RipGrayson:
## About The Pull Request

This is the QOL update that TGMC needs but didn't know it needed. Most construction/surgery/xeno actions now inform their user with overhead text.
No more missing a cooldown notification for your xeno ability because it got lost in the textbox during a fight.
No more missing a rookie doctor botching a surgery because the notification for the step he used got lost during a command announcement.
No more missing the notification for running out of ammo in a fire fight because there's 20 other marines firing and the text alert gets shuffled off the screen.

## Why It's Good For The Game

We're moving away from the philosophy of needing to stare at chat in order to see what's going on, runechat did it for speech, balloon_alerts do it for actions. It's just so much easier to see what's going on by looking at your character than it is to dig around in a bunch of text.
On the aesthetic side it actually looks really cool to see overhead notifications like "Franklin Ramis fixes the internal wiring of platinum miner" or "Queen (236) starts emitting pheremones", it makes us look very smooth.
So by implementing this not only do we look fancy, but it also enhances team cooperation because xenos/marines can see what their teammates are doing while being focused on combat. (Not that they couldn't before, they would just have to scroll through chat, which made it impractical).

## Changelog
:cl:
qol: Added balloon_alerts to many many different actions.
spellcheck: Fixed a few typos while adding balloon_alerts
/:cl:


TiviPlus (changes requested):
Very good pr

However you cant just copy paste the to chat, you need to remove the to chat (In some cases though you might want to keep it in case the log is useful)
Ex. conversions:
balloon_alert(usr, "cannot reach")
to_chat(src, span_warning("We cannot reach this from here!"))

X.balloon_alert(X, "Cannot do this here")
to_chat(owner, span_warning("We can't do this here!"))

buckling_mob.balloon_alert(buckling_mob, "Unable to buckle")
to_chat(buckling_mob, span_warning("You are unable to buckle yourself to [src]!"))

IE note: no punctuation, short as possible, generally avoid name inserts unless needed because they make text longer

Otherwise very good PR hope you finish this, ping me if you need any help

code/game/objects/items/reagent_containers/robodropper.dm
```diff
@@ -19,7 +19,7 @@
 				return
 
 			if(!target.is_injectable() && !ismob(target)) //You can inject humans and food but you cant remove the shit.
-				to_chat(user, span_warning("You cannot directly fill this object."))
+				user.balloon_alert(user, "You can't fill this object")
```
TiviPlus, line 22:
```suggestion
				balloon_alert(user, "You can't fill this object")
```

code/game/objects/items/reagent_containers/robodropper.dm
```diff
@@ -49,7 +49,7 @@
 						addtimer(CALLBACK(reagents, /datum/reagents.proc/reaction, safe_thing, TOUCH), 5)
 
 
-						to_chat(user, span_notice("You transfer [trans] units of the solution."))
+						user.balloon_alert(user, "You transfer [trans] units of the solution.")
```
TiviPlus, line 52:
```suggestion
						balloon_alert(user, "Transferred [trans] units")
```

code/game/objects/items/reagent_containers/spray.dm
```diff
@@ -27,23 +27,23 @@
 
 	if((A.is_drainable() && !A.is_refillable()) && get_dist(src,A) <= 1)
 		if(!A.reagents.total_volume)
-			to_chat(user, span_warning("[A] is empty."))
+			user.balloon_alert(user, "[A] is empty")
```
TiviPlus, line 30:
```suggestion
			A.balloon_alert(user, "It is empty")
```

code/game/objects/items/reagent_containers/spray.dm
```diff
@@ -27,23 +27,23 @@
...
 			return
 
 		if(reagents.holder_full())
-			to_chat(user, span_warning("[src] is full."))
+			user.balloon_alert(user, "[src] is full")
```
TiviPlus, line 34:
```suggestion
			balloon_alert(user, "It is full")
```

TiviPlus (changes requested):

RipGrayson:
I don't have time to manually check all things right now for review compliance, will reopen when I have more free time

PR #9905, opened 2022-04-04
Change xeno bomb armor to usual soft armor

goose4429:
## About The Pull Request

Xeno bomb armor is now a usual armor. Before it was essentially hard armor. No magic and noob coders who waste time on something like explosions. 

Raw numbers for bigger ex adjusted so that total damage in line with old damage. Slowdown, stagger and sunder have same effects as before. gib range is same.

**Before:**
Common case, Nade or CAS machinegun (light ex): 
 100dmg on armor 0, 70 dmg on armor 3
Rare case, Mortar HE direct hit or CAS rocket direct hit (devastating ex): 
 gib armor 0, 170 dmg on armor 1, 150 dmg on armor 3

**After:**
Common case, Nade or CAS machinegun (light ex): 
 100dmg on armor 0, 70 dmg on armor 3
Rare case, Mortar HE direct hit or CAS rocket direct hit (devastating ex): 
 gib armor 0, 180 dmg on armor 1, 140 dmg on armor 3

## Why It's Good For The Game

- Correct armor values in game codex.
- People don't need to check the code to see how explosions hit beno.
- People don't need to know it's actually hard and not soft armor.
- Allows further changes without disturbing the balance too much.
- No changes except for described above. Abilities give same resistances as before. Benos have same resistances as before.

## Changelog
:cl:
add: Soft bomb armor
del: Old bomb armor (in levels)
del: Old devour code (that functions only in CM)
balance: Bomb armor is now affected by sunder
balance: Changed bomb armors all over the board. Totally needs game testing.
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -174,14 +174,6 @@
 #define RESTRAINED_RAZORWIRE (1<<3)
 #define RESTRAINED_PSYCHICGRAB (1<<4)
 
-
-//Explosion resistance
-#define XENO_BOMB_RESIST_4 100
-#define XENO_BOMB_RESIST_3 80
-#define XENO_BOMB_RESIST_2 60
-#define XENO_BOMB_RESIST_1 40
-#define XENO_BOMB_RESIST_0 0
-
 #define SINGLE_CASING (1 << 0)
 #define SPEEDLOADER (1 << 1)
 #define MAGAZINE (1 << 2)
--- a/code/modules/mob/living/carbon/xenomorph/castes/beetle/castedatum_beetle.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/beetle/castedatum_beetle.dm
@@ -27,7 +27,7 @@
 	caste_flags = CASTE_CAN_BE_QUEEN_HEALED|CASTE_DO_NOT_ALERT_LOW_LIFE|CASTE_IS_A_MINION
 
 	// *** Defense *** //
-	soft_armor = list("melee" = 30, "bullet" = 30, "laser" = 25, "energy" = 20, "bomb" = XENO_BOMB_RESIST_2, "bio" = 20, "rad" = 20, "fire" = 30, "acid" = 20)
+	soft_armor = list("melee" = 30, "bullet" = 30, "laser" = 25, "energy" = 20, "bomb" = 30, "bio" = 20, "rad" = 20, "fire" = 30, "acid" = 20)
 
 	// *** Ranged Attack *** //
 	charge_type = CHARGE_TYPE_LARGE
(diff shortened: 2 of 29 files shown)


InterroLouis:
How will this work with the bomb armor for defenders? It seems to me that they'll end up with less protection from explosions when fortified or crest defensed than they would previously.

goose4429:
> How will this work with the bomb armor for defenders? It seems to me that they'll end up with less protection from explosions when fortified or crest defensed than they would previously.

Good question!

It's depends. Actually more armor in most cases. Previously defender had 2 armor without buffs, which decreased damage from explosions by 20. Base damage for nade is 100, resulting in 20% decrease. Now it is 20% at young and 35% at ancient. Stance before gave 3 armor = 30%, now 55% at ancient. Fortify previously gave immunity (armor 4), now 70% at young and 90% at ancient (about in line with other resistances). So more armor by default, less at fortify. Note that now it also stacks with defensive auras, so with Prae pheros you can get immunity as well.

I have changed bomb armor handling of stance and fortify so that bomb armor is not different from other armors. I tried to keep the defender defensive, and these values seems reasonable. They can be changed, though, so tell me if you feel that's ok or needs changes.

DeltaFire15:
This PR is kinda weird and feels like you are just throwing around numbers. Defenders no longer get fortify bomb immunity, anything that had bomb_armor_0 before takes more damage from all explosions, all with bomb_armor_1 tend to take less from most. bomb_armor_2 before take either a bunch more or a bunch less. bomb_armor_3 takes less or 0 now. The only ones unchanged are ones with bomb_armor_4 which still always take zero.
Gorger suddenly has a great nade armor, deva explosions gib a bunch of xenos which previously were ungibbable - uhm?
All in all I am quite unsure why this change is really needed.

goose4429:
> This PR is kinda weird and feels like you are just throwing around numbers. Defenders no longer get fortify bomb immunity, anything that had bomb_armor_0 before takes more damage from all explosions, all with bomb_armor_1 tend to take less from most. bomb_armor_2 before take either a bunch more or a bunch less. bomb_armor_3 takes less or 0 now. The only ones unchanged are ones with bomb_armor_4 which still always take zero. Gorger suddenly has a great nade armor, deva explosions gib a bunch of xenos which previously were ungibbable - uhm? All in all I am quite unsure why this change is really needed.

This PR is focused on making bomb armor same as others, with concise logic and comparable effects as other damage. Also remove sharp edges around the gib range and diff between armor 0 and 1 beno.

Well, yes, hard to make big change without having a lot of small issues.
Gorger is a hard one - he should be ungibbable 100%, but at the same time has poor defense stats. I decided that first one is more important, especially given that he's big slab of meat. But that one is the only one I'm not sure about.

Def fortify - he can still be immune. And this change was made so that the defences are in line - I don't see why must he have full immunity from bombs (which CAN damage armored stuff well) but not from bullets (which are NOT able to penetrate any armor unless special). Given he doesn't have full bullet immunity, he should have bomb one.

Damage numbers - there is marginal increase in some places, and significant decrease in others, yes. That was the main point of PR, actually. Because for example DEVA explosion will gib 300+HP elder prae. But when prae is ancient, suddenly it is not gibbed, but dealt ~150 dmg instead. And Queen, heavy-armored against explosions would be damaged by 20% less than lightly-armored Prae. 

All numbers are subject to change, so if you have propositions - please. Just saying "this number is bad" doesn't mean any other is better.

goose4429:
Disregard previous comments. The numbers WERE bad. Instead of doing my own numbers, I've done everything to be 99% same as was before, but with armor being soft. Abilities work 100% same as before.

TiviPlus (approved):
lgtm, if the define is unused remove it pls otherwise ill merge

PR #9911, opened 2022-04-04
Give hivemind resin spit

BraveMole:
## About The Pull Request

Pretty low fire rate

## Why It's Good For The Game

Make hivemind less passive, but it is still not very opressive for marine since it's only resin spit + 2 sec firerate

## Changelog
:cl:
balance: Give hivemind resin spit
/:cl:


Pariah919 (approved):
Hivemind is already annoying enough, but I'd rather see it changed in another aspects so this is a okay change from my PoV.

PR #9916, opened 2022-04-04
Fix BR runtimes and map load loops

TiviPlus:
## About The Pull Request

I have no idea what caused it but splitting it into an ss is a good idea and fixed it

## Changelog
:cl:
fix: fixed a batch of modular mapping bugs
/:cl:

### DIFF
--- a/code/__DEFINES/_subsystems.dm
+++ b/code/__DEFINES/_subsystems.dm
@@ -69,7 +69,8 @@
 #define INIT_ORDER_MAPPING 9
 #define INIT_ORDER_PERSISTENCE 7 //before assets because some assets take data from SSPersistence, such as vendor items
 #define INIT_ORDER_ATOMS 6
-#define INIT_ORDER_MACHINES 5
+#define INIT_ORDER_MODULARMAPPING 5
+#define INIT_ORDER_MACHINES 4
 #define INIT_ORDER_AI_NODES 3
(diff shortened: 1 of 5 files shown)


BraveMole (approved):

fira:
I might say something stupid, but... It looks less worse but I still don't think that's correct 

I'm assuming what lead to wanting to do this is issues from parallel interactions around loading resulting from partial MC locking + CHECK_TICK in the maploader, and the INVOKE_ASYNC in that LateInitialize of the spawners

Now if you queue it to a SS for doing that synchronously in its Initialize, that means several map inserts don't race each other, but because the MC doesn't wait for SS (the MC Initialize is waitfor=0), I believe this still means that map loadings by that SS will race with the rest of init

TiviPlus:
> I might say something stupid, but... It looks less worse but I still don't think that's correct
> I'm assuming what lead to wanting to do this is issues from parallel interactions around loading resulting from partial MC locking + CHECK_TICK in the maploader, and the INVOKE_ASYNC in that LateInitialize of the spawners

no

> Now if you queue it to a SS for doing that synchronously in its Initialize, that means several map inserts don't race each other, but because the MC doesn't wait for SS (the MC Initialize is waitfor=0), I believe this still means that map loadings by that SS will race with the rest of init

read the code again

PR #9943, opened 2022-04-06
Revert "Triples damage from guns/acid spit and xeno melee"

TiviPlus:
pranked


Ragecrabbo:
Finally some runner nerfs

PR #9942, opened 2022-04-06
Triples damage from guns/acid spit and xeno melee

Pariah919:
## About The Pull Request
3x damage for guns and xeno melee.

## Why It's Good For The Game
I think and believe our population is too high on average on a basic god given TGMC round, and we need a PR like this to reduce our overall player count by removing people from the round faster.

PR #9945, opened 2022-04-06
Revert "Triples damage from guns/acid spit and xeno melee"

Lefunny:
Reverts tgstation/TerraGov-Marine-Corps#9942


BraveMole:
Ah yes, good pr, zero changes

PR #9951, opened 2022-04-07
Give Hivemind Transfer Plasma

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

The hivemind is one of the premiere support roles in the hive. It provides healing, building, and weeding against all odds from marines' superior firepower. It's the drone that keeps on supporting.

Weirdly, even though it can heal, it cannot give plasma to other xenomorphs. This can be an alternative to #9911 , but it depends on the maintainers. I'm giving it the drone plasma transfer since ~~hivelord has a lot more compared to drone, I'm giving hivemind a stronger plasma transfer, between drone and hivelord, just to make it more viable for a drone caste player to go hivemind, and it makes sense that~~ hivemind is drone that saw the weakness of fresh and embrace the resin.

## Changelog
:cl:
add: Give Hivemind Transfer Plasma
balance: Provide Hivemind plasma transfer so it can support xenos; it is the same as drone

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/hivemind/abilities_hivemind.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/hivemind/abilities_hivemind.dm
@@ -46,6 +46,12 @@
 /datum/action/xeno_action/activable/psychic_cure/hivemind/should_show()
 	return !(owner.status_flags & INCORPOREAL)
 
+/datum/action/xeno_action/activable/transfer_plasma/hivemind
+	plasma_transfer_amount = PLASMA_TRANSFER_AMOUNT * 3
+
+/datum/action/xeno_action/activable/transfer_plasma/hivemind/should_show()
+	return !(owner.status_flags & INCORPOREAL)
+
 /datum/action/xeno_action/toggle_pheromones/hivemind/should_show()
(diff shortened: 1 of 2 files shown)


Naaanii:
Could give it it's own version imo.

More plasma transfer than drone, slightly higher plasma cost, much longer CD, almost on par with heal so it's not spammable like drones and hivelord.

BraveMole (changes requested):
Pretty sure you can use it even when unmanifested, so no

code/modules/mob/living/carbon/xenomorph/castes/hivemind/abilities_hivemind.dm
```diff
@@ -46,6 +46,12 @@
 /datum/action/xeno_action/activable/psychic_cure/hivemind/should_show()
 	return !(owner.status_flags & INCORPOREAL)
 
+/datum/action/xeno_action/activable/transfer_plasma/hivemind
+	plasma_transfer_amount = PLASMA_TRANSFER_AMOUNT * 3
```
BraveMole, line 50: that's just way too strong, no

BraveMole (changes requested):

BraveMole (approved):

PR #9952, opened 2022-04-07
Wraith stats buff

Lefunny:
## About The Pull Request

Wraith is in a bad time right now, its supportive skills are fair lower than the usual after recent nerfs and the caste itself have a very low survivability as the caste got quite much a low amount of plasma, speed and damage for a t2, the point of this PR inst changing the damage or health itself but letting the wraith perform more gameplay with a slight stat buff on its speed and plasma by a 50 up in the board by each age stage ((for plasma ofcourse)) and a 0.1 buff in the board for speed, making in the end an ancient wraith with 1.3 speed and 450 plasma 

## Why It's Good For The Game

Making wraith more of a scout/quick combat caste while also opening ways for new uses of its skill as the caste itself got more plasma, i hope it gives coders a better idea of a good direction to take the caste in.

## Changelog
:cl:

balance: Wraith plasma now 450 in ancient ((following a +50 plasma for each age))
balance: Wraith speed now 1.3 in ancient ((folowwing a +0.1 speed for each age))

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/wraith/castedatum_wraith.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/wraith/castedatum_wraith.dm
@@ -65,7 +65,7 @@
 	speed = -1.1
 
 	// *** Plasma *** //
-	plasma_max = 300
+	plasma_max = 350
 	plasma_gain = 20
 
 	// *** Health *** //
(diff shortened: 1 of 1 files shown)


2whatever2:
Wraith doesn't even need a buff especially looking at the nerfs that's used as a reason. Since the nerfs literally just target an unfair tactic bordering on abuse, applying a constant blur, stagger, and screenshake on every blink, and this isn't even mentioning the fact that no wraith player does this besides the every now and then static (even then it's not much). The nerfs didn't even affect wraith as a whole since most if not all players on wraith don't even spam blink while shifted to do this tactic. 😩

TiviPlus:
Zack said it

PR #9953, opened 2022-04-07
T-25/29 changes- Same bullet between both of them, stat changes for T-25

Pariah919:
## About The Pull Request
Renames smartmachinegun bullet to smartgun bullet, considering the t-25/29 share it now.

T-25 changed to .25 bracket, it now uses the Smartmachinegun bullet.
T-25 has +20 ammo, 100 bullets now.
T-25 Packet now useable in T-29 magazines.
Smartgun magazines now say '10x26 High Pressure' to denote the difference.
10x26 HP packet reduced to 300 from 320.
Wield delay lowered.
image

## Why It's Good For The Game
T-25 and T-29 should be consistent in the damage they deal between eachother, and should generally perform the same but feel a bit different.

## Changelog
:cl:
balance: T-25/29 share the same bullet, caliber and damage/sunder. The T-25 has a bit less wield delay and +20 rounds. As it does less sunder and damage overall from this change.
spriteadd: t25 box changed to say SG instead of T-25.
/:cl:


code/modules/projectiles/magazines/misc.dm
```diff
@@ -50,12 +50,12 @@
 	max_rounds = 210
 
 /obj/item/ammo_magazine/packet/t25
-	name = "box of 10x26mm (T-25)"
-	desc = "A box containing 320 rounds of 10x26mm caseless tuned for a T-25 smartrifle. It cannot be used for T-29 drums!"
+	name = "box of 10x26mm high-pressure"
+	desc = "A box containing 4004 rounds of 10x26mm 'HP' caseless tuned for a smartgun."
```
novaepee, line 54: I'm assuming it doesn't contain 4004 rounds but instead 300.

PR #9956, opened 2022-04-08
Marines doing defibs or manning mounted weapons cannot be pushed by other humans.

hyper2snyper:
## About The Pull Request

Title

## Why It's Good For The Game

SERIOUSLY STOP FUCKING PUSHING ME!

## Changelog
:cl:
qol: Marines doing defibs or manning mounted weapons cannot be pushed by other humans.
/:cl:


Losenis:
PLEASE add surgery steps to this also

Lumipharon:
Muh immersions.

Honestly I don't think this is warranted. Just stand next to a wall out of the way.
I don't want to have to stop wielding my gun to shove a medic that's decided to block a corridor while frontline defibbing.

novaepee:
> Muh immersions.
> 
> Honestly I don't think this is warranted. Just stand next to a wall out of the way. I don't want to have to stop wielding my gun to shove a medic that's decided to block a corridor while frontline defibbing.

If there a few things in CM that I want in TGMC, it's marines not pushing corpsmen using defib on a red marine.

code/modules/mob/living/living.dm
```diff
@@ -382,6 +382,9 @@
 		if(!(L.status_flags & CANPUSH))
 			return
 
+		if(HAS_TRAIT(L, TRAIT_NO_PUSHING_BY_HUMAN) && istype(src, /mob/living/carbon/human))
```
BraveMole, line 385: is type src. Beeeeh.

And putting something that will almost never happen in bump, that happens all the time, is so meh as well.

Find another solution that does not require to meddle with a proc like that

BraveMole (changes requested):

code/modules/mob/living/living.dm
TiviPlus: i dont see why setting anchored wouldnt work

PR #9960, opened 2022-04-08
Refactors energy weapon magazines to use overlays

Lumipharon:
## About The Pull Request
Makes energy weapon magazines (mainly TE cells since those are the only ones we normally see in game) use overlays instead of different icon states, and makes them actually update as they're currently broken.

Cell items use overlays for their cell charge but for whatever reason TE cells uses icons. This is dumb and it wasn't currently even working, so I've changed them to use overlays also.

image

Pulse rifle ammo recharges so I can't be bothered getting an actual image of the empty sprite.

Also made the pulse rifle actually use pulse rifle ammo again. I assume this was an oversight at some point.
edit: Most critical change, potato battery is now actually centered on the charge level, rather than it just kinda floating on the side.

## Why It's Good For The Game
Cells will no longer prank you by appearing full when they're actually empty.
Children of cell using the same system makes sense

## Changelog
:cl:
refactor: Energy weapons now use overlays to display their charge level.
fix: TE power cell sprites will no longer visually appear to always be at full power.
fix: Pulse rifles now use actual pulse rifle cells
/:cl:


TiviPlus (changes requested):
Copy pasting update icon root is not a fix, it should parentcall

BraveMole (changes requested):
Why is update_icon override for cell lasgun, and why are you copy pasting the update_icon code of atom or whatever ?

deleting line 62 to 63 should be enough, right ?

Lumipharon:
> Copy pasting update icon root is not a fix, it should parentcall

@TiviPlus @BraveMole 
Lasgun has a blanked out update_icon, and the parent /obj/item/ has an entirely different update_icon also. see https://github.com/tgstation/TerraGov-Marine-Corps/blob/d9d1c27290855a76c9552ba77cd11f4d87a931d7/code/modules/power/cell.dm
Is there a way to call parent butskip like,  3 levels?

Lumipharon:
damn shitcode, ok I need to make more changes to all cells so I can actually call to parent.

code/modules/power/cell.dm
```diff
@@ -22,14 +22,12 @@
...
+/obj/item/cell/update_overlays()
+	. = ..()
 	if(charge < 0.01)
 		return
-	else if(charge / maxcharge >= 0.995)
-		add_overlay("cell-o2")
-	else
-		add_overlay("cell-o1")
+	var/remaining = CEILING((charge / max(maxcharge, 1)) * 100, 25)
+	return "[charge_overlay]_[remaining]"
```
TiviPlus, line 30: . is a list that returns overlays to add
. += "Icon_state" would work but making an image or mutable appearance is better

TiviPlus (changes requested):
Rest lgtm

code/modules/power/cell.dm
Lumipharon: done.

TiviPlus (approved):

PR #9969, opened 2022-04-08
RANDOM. DISK GENERATOR. LOCATIONS.

MrHorizons:
## About The Pull Request
Thanks to some code created by SandPoot#5908 in coderbus, there is now a system that randomizes the locations of the nuke disk generators. Currently, they can only occur at the former locations of intel computers plus the original locations of the generators, but if desired, a different sort of variance can be made.

## Why It's Good For The Game
It's better to not have objectives be in the same locations every round.

## Changelog
:cl:
add: Added a system that randomizes the location of the nuclear disk generators with those of intelligence computers.
/:cl:


Arctrooper8:
Is there a system to prevent all the disks from being too close?

lKiseki:
this is okay but
is there a system in place to prevent disk generators from spawning on canterbury locations? cause the canterbury is particularly known for landing on many things, including intelligence computers

MrHorizons:
> Is there a system to prevent all the disks from being too close?

Not specifically, but I believe most intel computers are sufficiently far enough from each other that the chances are low. I'd like to see how it does in a playtest.

> this is okay but is there a system in place to prevent disk generators from spawning on canterbury locations? cause the canterbury is particularly known for landing on many things, including intelligence computers

Now that's something I didn't consider. I'll see what I can do to remedy that.

MrHorizons:
Turns out this wasn't as close to complete as I had thought.

PR #9971, opened 2022-04-09
Fixes/improves Quick Equip functionality

Lumipharon:
## About The Pull Request
Currently quick equip is... janky as hell. This PR makes it work the way it seems to have been intended, and refines it a bit to be more likely to get the item you're actually wanting (normally a weapon).

Currently the way it works, it cycles through your inventory starting with your preferred slot, but after that it does so in a way that mixes different slot defines with each other. This means depositing to, or (more significantly) drawing from certain spots functionally impossible with normal loadouts, or even better yet, in some cases you'll simply take off an item like your boots or uniform instead of something equipped to it like a boot knife.

For example, it will currently ALWAYS try take something from your belt before taking something from your back, like your second gun.

For using quick equip to store items, I've also changed it so you're more likely to put things where you want them to go.
Preferred slot is still the first choice, but after that it will choose the less common slots first (like boot slot, shoulder/pouch holsters etc), so you should actually store pistols and what not in their specific storage slots instead of your bag all the time.

This also fixes certain slots simply not being checked at all for both storing and drawing, such as shoulder holsters. Wew.

Behold, the power of quick equip and zero video editing!

https://user-images.githubusercontent.com/7869430/162551515-185fc69c-da34-4641-bd1b-7e9bce677794.mp4

I have made the draw and store order as logical as possible but it might not be perfect. If you have any feedback on how it could be improved, or if I've missed anything please shout out.

This should also play nice with Tivi's 2nd quick equip addition, as that just adds a new hotkey + new preferred slot, it didn't touch how the procs actually ran.

Edit:
Here is the new draw order. For reference things like B_holster are still checking the back slot, but are checking for actual back holsters/guns on the back, where as slot_back will check anything on the back like a bag.
image

Please provide feedback on how this works for you, as I'm happy to make changes.

## Why It's Good For The Game
Main hotkey in the game actually working properly is good.

## Changelog
:cl:
refactor: quick equip now actually draws from or stores into everything you'd expect it to, and does so in a more sane order

/:cl:


Lumipharon:
I have no idea what this lint error is, send help.

Lumipharon:
That linter error message is still ???? but the issue is sorted at least and Git seems to be happy.

Lumipharon:
Merge conflict is gone but the tag is still there?

code/modules/mob/mob.dm
```diff
@@ -331,15 +331,66 @@
...
+	if (slot == SLOT_IN_B_HOLSTER && ( !( istype(I, /obj/item/storage/holster) || istype(I, /obj/item/weapon) ) ) )
+		return FALSE
+	if (slot == SLOT_IN_ACCESSORY && (!istype(I, /obj/item/clothing/under ) ) ) //note to self, draw code has a further istype for the attachment, does this also need an early return if there is no attachment, or it's fine? probs fine
+		return FALSE
+	if (slot == SLOT_IN_L_POUCH && ( !( istype(I, /obj/item/storage/holster) || istype(I, /obj/item/weapon) || istype(I, /obj/item/storage/pouch/pistol) ) ) ) //if pistol pouches get merged into holsters, this will need to be updated
+		return FALSE
+	if (slot == SLOT_IN_R_POUCH && ( !( istype(I, /obj/item/storage/holster) || istype(I, /obj/item/weapon) || istype(I, /obj/item/storage/pouch/pistol) ) ) )
+		return FALSE
+
+	//belt holsters
```
TiviPlus, line 348: this massive if chaining is ugly
also please dont do (  ( ( ( ) ) ) ) do (((())) instead

code/modules/mob/mob.dm
```diff
@@ -331,15 +331,66 @@
...
+		return TRUE
+		//general storage module
+	if (istype(I, /obj/item/clothing/suit))
+		var/obj/item/clothing/suit/B = I
+		if (!istype(B.attachments_by_slot[ATTACHMENT_SLOT_STORAGE], /obj/item/armor_module/storage) )
+			return FALSE
+		var/obj/item/armor_module/storage/U = B.attachments_by_slot[ATTACHMENT_SLOT_STORAGE]
+		var/obj/item/storage/S = U.storage
+		if(!length(S.contents))
+			return FALSE
```
TiviPlus, line 387: my eyes what is this

TiviPlus (changes requested):

code/modules/mob/mob.dm
Lumipharon: Will do.

code/modules/mob/mob.dm
Lumipharon: Perfection :^)

Lumipharon:
@TiviPlus Brave never responded, send help. Or merges.

TiviPlus (changes requested):
As explained in discord, fix the istype chains

code/game/objects/items.dm
```diff
@@ -317,6 +317,15 @@
...
 
+///Called to return an item to equip using the quick equip hotkey. Will try return a stored item, otherwise returns itself to equip.
+/obj/item/proc/do_quick_equip()
+	var/obj/item/found = locate(/obj/item/storage) in contents
+	if(!found)
+		found = locate(/obj/item/armor_module/storage) in contents
+	if(found)
+		return found.do_quick_equip()
+	else
+		return src
```
TiviPlus, line 328:
```suggestion
	return src
```

code/modules/clothing/under/under.dm
```diff
@@ -140,6 +140,16 @@
...
 
+//we only want to quick equip from actual 'holster' type webbings
+/obj/item/clothing/under/do_quick_equip()
+	var/obj/item/found = locate(/obj/item/armor_module/storage/uniform/holster) in contents
+	if(!found)
+		found = locate(/obj/item/armor_module/storage/uniform/knifeharness) in contents
+	if(found)
+		return found.do_quick_equip()
+	else
+		return src
```
TiviPlus, line 151:
```suggestion
	return src
```

TiviPlus (changes requested):

2kajowsmol:
in my opinion the back slot should be a higher priority than the pouches

Lumipharon:
Updated draw order to following based off some initial feedback.
image

cazuccmuss:
I frequently use a belt harness with ammo in my back slot and now im pulling medicine from pockets instead of ammo. I'd say either bump back priority higher or let you manually place priority on slots because its killing my reload.

Lumipharon:
> I frequently use a belt harness with ammo in my back slot and now im pulling medicine from pockets instead of ammo. I'd say either have an option to bump back priority higher or let you manually place priority on slots because its killing my reload.

Then you get the flipside where someone with ammo in their mag pouch is drawing items (potentially with a draw delay if it's a backpack) from thei back slot instead of their pocket.

Having full control over the order would be ideal but is out of scope for this PR at least. Are you using both your quick equip keys? You can key one to pull from your backpack.

cazuccmuss:
> Having full control over the order would be ideal but is out of scope for this PR at least. Are you using both your quick equip keys? You can key one to pull from your backpack.

Yeah that's understandable, just something that would be nice to keep in mind for the future I guess. The only problem I have with alt quick equip is that it messes up muscle memory, but it'll be fine in time.

PR #9977, opened 2022-04-09
Health kits chain apply

QualityVan:
## About The Pull Request
After being applied to a treatable limb, health kits (gauze/ointment/advanced) will then cycle through all the other patient's limbs with a delay on each treatable one. They'll only use one unit of the stack for this entire cycle.
Alternative to #9968 

## Why It's Good For The Game
Helps with a primary complaint of basic kits, which is that if you apply them regularly you'll burn through them super fast.
Might be excessive with regard to advanced kits, in which case their max stack could be reduced to five or so.

## Changelog
:cl:
add: After applying a health kit to one limb, you'll start applying it to all other damaged ones without using additional stacks.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -97,7 +97,7 @@
 	for(affecting AS in patient.limbs)
 		if(!can_affect_limb(affecting))
 			continue
-		 //Always delay on the first try, otherwise only delay if you patched the last iterated limb.
+		//Always delay on the first try, otherwise only delay if you patched the last iterated limb.
 		if(affected && (user.do_actions || !do_mob(user, patient, SKILL_TASK_VERY_EASY / (unskilled_penalty ** 2), BUSY_ICON_FRIENDLY, BUSY_ICON_MEDICAL)))
 			to_chat(user, span_notice("You stop tending to [patient]'s wounds."))
 			return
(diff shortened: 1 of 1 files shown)


BraveMole:
I prefer MJP solution.

BraveMole (changes requested):

Durbles:
Kits seem to apply chain healing to limbs that have no damage. I've even seen it apply the kit to robotic limbs.

QualityVan:
> Kits seem to apply chain healing to limbs that have no damage. I've even seen it apply the kit to robotic limbs.

That'll happen if there's untreated wounds on the limb, even if they have zero damage. The robolimbs thing was an oversight, fixed now.

Durbles:
> > Kits seem to apply chain healing to limbs that have no damage. I've even seen it apply the kit to robotic limbs.
> 
> That'll happen if there's untreated wounds on the limb, even if they have zero damage. The robolimbs thing was an oversight, fixed now.

How can a limb have a wound if it has zero damage?

XSlayer300:
Make it so that the brute healing supplies (bandages and brute trauma kits) chain applies will prioritize bleeding first then the rest of the damaged parts. In addition, should the chain healing stuff should have a minimum damage limit of sorts? Maybe around 10 so you won't waste healing kits?

Losenis:
Any chance to open a new pr and make bandages and ointments (IE NOT Trauma and burn kits) apply instantly to all bodyparts at a high enough medical skill level? Like MO or Synth high. It'd make them have a use compared to trauma kits and burn kits, and they don't heal enough to make midcombat spamclicking relevant afaik.

PR #9979, opened 2022-04-09
Add wraith portal, remove hyperposition, remove phasing

BraveMole:
## About The Pull Request

Scrap hyperposition and shadow abilities.

Add a new ability, portal. Wraith can place pairs of portal, which allow instantanious deplacement without any cooldown. The wraith can freely set them where he wants (max range between two portals is 20 tiles), but keep in mind that marines can cross it too, as well as bullets, grenades, etc.

This is probably busted, and will need some tweaks. But this looks very very fun. 

## Why It's Good For The Game

Hyperposition is meh, this allows for way more emergent gameplay. We will see about balance later.
Phasing is meh removing as well. Got plan to replace later

## Changelog
:cl:
add: Add wraith portals
delete: Remove hyperposition
delete: Remove wraith phasing
/:cl:


novaepee:
Insert Wheel of Time's traveling reference
image

KrabSpider:
Good.

InterroLouis:
Haha this is going to take Braziling to the next level.

Lumipharon:
So can can a wraith just portal benos into fob with this?

BraveMole:
> So can can a wraith just portal benos into fob with this?

yes

Pill-U:
Just please forbid them from being placed on canterbury

Pill-U:
Wraith can put one portal on tadpole/alamo and one outside and they won't unlink when they go shipside. Since marines can shoot back through it it might be fine to leave it that way, but your choice.

Naaanii:
Maybe make it so a portal can't be placed in LZ or alamo as it's too easy for wraith too get onto LZ and forces marines to forever stay in LZ where both game modes require them to push.

code/datums/round_statistics.dm
```diff
@@ -65,7 +65,6 @@ GLOBAL_DATUM_INIT(round_statistics, /datum/round_statistics, new)
 	var/hivelord_healing_infusions = 0
 	var/spitter_acid_sprays = 0
 	var/spitter_scatter_spits = 0
-	var/wraith_hyperpositions = 0
```
TiviPlus, line 68: should add a tracker for teleported objects maybe idk

code/modules/mob/living/carbon/xenomorph/castes/wraith/abilities_wraith.dm
```diff
@@ -800,3 +613,154 @@ GLOBAL_LIST_INIT(wraith_banish_very_short_duration_list, typecacheof(list(
...
+	if(locate(/obj/effect/wraith_portal) in get_turf(owner))
+		if(!silent)
+			to_chat(owner, span_xenowarning("There is already a portal here!"))
+		return FALSE
+	return ..()
+
+/datum/action/xeno_action/portal/action_activate()
+	. = ..()
+	qdel(portal_one)
+	portal_one = new(get_turf(owner))
```
TiviPlus, line 652: consider making these qdel null in case someone decides to clown with this proc

code/modules/projectiles/projectile.dm
```diff
@@ -289,7 +289,7 @@
 	GLOB.round_statistics.total_projectiles_fired++
 	SSblackbox.record_feedback("tally", "round_statistics", 1, "total_projectiles_fired")
 
-	if(ammo.flags_ammo_behavior & AMMO_BALLISTIC)
+	if(ammo.flags_ammo_behavior & AMMO_BALLISTIC && !recursivity)
 		GLOB.round_statistics.total_bullets_fired++
```
TiviPlus, line 293: ?

code/modules/projectiles/projectile.dm
```diff
@@ -328,6 +328,9 @@
 		if(PROJECTILE_FROZEN)
 			return
 
+	if(QDELETED(src))
+		return
```
TiviPlus, line 332: how would it be qdeleted here

code/modules/projectiles/projectile.dm
```diff
@@ -350,6 +353,9 @@
 		if(PROJECTILE_FROZEN)
 			return PROCESS_KILL
 
+	if(QDELETED(src))
+		return PROCESS_KILL
```
TiviPlus, line 357: how would it be qdelted here, its checked above

code/modules/projectiles/projectile.dm
```diff
@@ -475,10 +481,12 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
...
 				break
-			if(HAS_TRAIT(turf_crossed_by, TRAIT_TURF_FREEZE_BULLET))
+			if(HAS_TRAIT(turf_crossed_by, TRAIT_TURF_BULLET_MANIPULATION))
+				SEND_SIGNAL(turf_crossed_by, COMSIG_TURF_PROJECTILE_MANIPULATED, src)
+				if(HAS_TRAIT_FROM(turf_crossed_by, TRAIT_TURF_BULLET_MANIPULATION, PORTAL_TRAIT))
+					return
 				RegisterSignal(turf_crossed_by, COMSIG_TURF_RESUME_PROJECTILE_MOVE, .proc/resume_move)
-				permutated.Cut()
-				STOP_PROCESSING(SSprojectiles, src)
+				permutated?.Cut()
```
TiviPlus, line 489: why a ?.

code/modules/projectiles/projectile.dm
```diff
@@ -1131,8 +1146,11 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 			ammo.do_at_max_range(src)
 			end_of_movement = TRUE
 			break
-		if(HAS_TRAIT(next_turf, TRAIT_TURF_FREEZE_BULLET))
+		if(HAS_TRAIT(next_turf, TRAIT_TURF_BULLET_MANIPULATION))
+			SEND_SIGNAL(next_turf, COMSIG_TURF_PROJECTILE_MANIPULATED, src)
 			QDEL_LIST_IN(laser_effects, 1)
+			if(HAS_TRAIT_FROM(next_turf, TRAIT_TURF_BULLET_MANIPULATION, PORTAL_TRAIT))
+				return
```
TiviPlus, line 1153: whats with these snowflake seeming trait checks

code/modules/projectiles/projectile.dm
```diff
@@ -985,7 +997,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 
 	x_offset = round(sin(dir_angle), 0.01)
 	y_offset = round(cos(dir_angle), 0.01)
-	if(projectile_batch_move() == PROJECTILE_FROZEN)
+	if(projectile_batch_move(!recursivity) == PROJECTILE_FROZEN)
 		permutated.Cut()
```
TiviPlus, line 1001: ?

code/modules/projectiles/projectile.dm
```diff
@@ -614,7 +625,8 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 ///Signal handler to tell the projectile to move again
 /obj/projectile/proc/resume_move(datum/source)
 	SIGNAL_HANDLER
-	UnregisterSignal(source, COMSIG_TURF_RESUME_PROJECTILE_MOVE)
+	if(source)
+		UnregisterSignal(source, COMSIG_TURF_RESUME_PROJECTILE_MOVE)
```
TiviPlus, line 629: ???
source should never be null here

TiviPlus (changes requested):

code/modules/projectiles/projectile.dm
BraveMole, line 489:
```suggestion
				permutated.Cut()
```

code/modules/projectiles/projectile.dm
BraveMole: So we don't count twice bullets that passed through a portal

code/modules/projectiles/projectile.dm
BraveMole: in batch move

code/modules/projectiles/projectile.dm
BraveMole: Less performance cost than having two traits to check.

We check if something weird is happening to the bullet, and if yes, we are looking for the source so we know what we shoudl do with it

code/modules/projectiles/projectile.dm
BraveMole: it prevents the first part of the visual of the laser to appear (the one that comes directly from the gun). 

See where first projectile is used.

code/modules/projectiles/projectile.dm
```diff
@@ -537,9 +545,12 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
...
 			end_of_movement = i
 			break
-		if(HAS_TRAIT(next_turf, TRAIT_TURF_FREEZE_BULLET))
+		if(HAS_TRAIT(next_turf, TRAIT_TURF_BULLET_MANIPULATION))
+			SEND_SIGNAL(next_turf, COMSIG_TURF_PROJECTILE_MANIPULATED, src)
+			if(HAS_TRAIT_FROM(next_turf, TRAIT_TURF_BULLET_MANIPULATION, PORTAL_TRAIT))
+				return
 			RegisterSignal(next_turf, COMSIG_TURF_RESUME_PROJECTILE_MOVE, .proc/resume_move)
-			permutated.Cut()
+			permutated?.Cut()
```
BraveMole, line 553:
```suggestion
			permutated.Cut()
```

ThelastArchiver:
what happens if an ob hits the tile with a portal? does it just ignore it does the ob fly through the portal or do we get a 1 tile explosion on the other side?

code/modules/projectiles/projectile.dm
BraveMole: ditto, in batch move

tyeagg:
needs a way to destroy it since xenos can just get in and out instantly with it and if a marine tries to chase their gonna end up 20 tiles away from friendlies and probably surrounded by xenos

FosterGrime-Heymynameslouie:
This is a pretty funny idea from the start but thinking about it critically, it might need some severe nerfs if it dosent get outright scrapped.

- This is a WAYY to powerful to be an ability that wraith gets right off the bat, it should be a primordial at best or an ancient ability at worse.

- This completely invalidates one of the Kings ability, the mass teleport one. It should go to show that mass transportation, being an endgame ability being such a game changer, should not be given to wraith with the current accesability it has.

- The portal being able to redirect BOILER GLOBS directly into places like Crash Canter, Tadpole, FOB, and virtually any kind of marine hold is pretty busted.

- Plasma cost and cooldown needs to be jaked way up, in its current state, the wraith can make 2 way in and out portals whenever it feels like granted it rests for 5 seconds to regain its plasma. Should drain all of its reserve plasma and have atleast a 2 minute cooldown so marines can see where the other end will be placed and anticipate it, granted the other end gets placed first.

TLDR 2 way Portal too busted, need to be primordial ability at worse or entirely revamped, scrapped, or severely nerfed at best.

BraveMole:
> Apparently looping a bullet through the 2 portals will eventually crash the server, dunno if you're aware of that/already fixed it so just wanted to make sure

Fixed !

Kuro020 (approved):

Autisem:
I think replacing this with phasing is relativly balanced since that means you cant nearly as easily just go into a restricted/secured space and setup shop. Not sure about the bullets going trough but would make for some cool gameplay, also makes pressure from more then 1 direction bit easier so thats fun

Wayland-Smithy:
Resting, passable, xenos seem to block travel and charge_crush momentum gets reset on crossing.

Al-1ce:
>places portal on lava

It's gamer time.

Al-1ce:
It's a nice idea but is actually dogshit in play. This would be fine if only the Wraith was allowed to enter the portal.

Lumipharon:
This seems super abusable to ggnore marines to perma kill them by portalling them to their death and either blocking the return portal (which considering humans have no way of shuffling xenos isn't hard), or simply closing the portal behind them.

You can also do zanny shit like making deathtraps they can't escape from (portal onto lava then wall it off), or death chambers with acid pools and turrets etc.

Pariah919:
Watching this in play in magmoor was the funniest 1 hour of my life of watching marines basically die instantly with no chance to play against it.

This shouldn't be a thing outside of the wraith themselves, though. Gonna be real.

Al-1ce:
The second portal is only able to be spawned by the default right-click toggle which a fair amount of players (including myself) do not use. Bit of an oversight.

BraveMole:
> The second portal is only able to be spawned by the default right-click toggle which a fair amount of players (including myself) do not use. Bit of an oversight.

Not an oversight, left click is deprecated for a long time now.

Al-1ce:
> Also you can just use the alternate keybind


Maybe I'm old fashioned. I couldn't find an alternate keybind for right-click specific abilities, like the second portal or loading underslungs.

novaepee:
You can shoot it to destroy portal, but you can't melee it to destroy portal. Maybe a bad idea to melee it, but so it goes.

PR #9980, opened 2022-04-09
Revert "bandages heal more, but take longer to use"

BraveMole:
Reverts tgstation/TerraGov-Marine-Corps#9968


BraveMole:
Mjp coding doesn't work

PR #9985, opened 2022-04-10
Revert "Makes missing limbs bleed"

Pariah919:
This makes bleeding literally nonstop without any way of stopping it until you replace the limb entirely, which is pretty absurd.


Derrick2Electricboogaloo:
If you make it so salving/bandaging/cauterizing it during surgery stops the crazy blood loss. This would be fine.

BraveMole:
You're not reverting a fix because the balance displease you.

You make a proper balance change if you do not like it, either by deleting the feature, or by adding a new thing like derrick said

PR #9986, opened 2022-04-10
Drastically reduces blood loss from missing limbs

lKiseki:
## About The Pull Request
Per title. Missing limb blood loss has been reduced from 5 > 0.8.

## Why It's Good For The Game
image
This happened in a minute or so. This is very much _**not**_ okay.

## Changelog
:cl: Lewdcifer
balance: Drastically reduced blood loss from missing limbs.
/:cl:

### DIFF
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -87,7 +87,7 @@
 		for(var/l in limbs)
 			var/datum/limb/temp = l
 			if(temp.limb_status & LIMB_DESTROYED && !(temp.limb_status & LIMB_AMPUTATED))
-				blood_max += 5 //Yer missing a fucking limb.
+				blood_max += 0.8 //Yer missing a fucking limb.
 				continue
 			if(!(temp.limb_status & LIMB_BLEEDING) || temp.limb_status & LIMB_ROBOT)
 				continue


Lumipharon:
This needs to happen, but a better long term solution would be for a delimbing to causing bleeding to the torso/groin, as that's where the blood would be coming from, and that would mean it could be bandaged.

I'll try look into unless someone smarter can do it first.

Arctrooper8:
> This needs to happen, but a better long term solution would be for a delimbing to causing bleeding to the torso/groin, as that's where the blood would be coming from, and that would mean it could be bandaged.
> 
> I'll try look into unless someone smarter can do it first.
> 
> Edit: Oh wait, that's literally what happens now/and did before the fix. :^)

So the real problem would be informing players about the fact that bandaging their torso will stop their lots limb bleeding?

Lumipharon:
> > This needs to happen, but a better long term solution would be for a delimbing to causing bleeding to the torso/groin, as that's where the blood would be coming from, and that would mean it could be bandaged.
> > I'll try look into unless someone smarter can do it first.
> > Edit: Oh wait, that's literally what happens now/and did before the fix. :^)
> 
> So the real problem would be informing players about the fact that bandaging their torso will stop their lots limb bleeding?

Nah, when you get delimbed it does X damage to the attached part (which causes bleeding).
But the bug fix was for a specific missing limb bleed thing, which is supposed to happen in addition to that. Which didn't work due to shitcode.

TiviPlus (changes requested):
Dont put numbers players dont understand in your changelog: 
bad:
Changed gun damage from 5-> 10

good: 
Doubled guns damage
Drastically increased guns damage

lKiseki:
> Dont put numbers players dont understand in your changelog: bad: Changed gun damage from 5-> 10
> 
> good: Doubled guns damage Drastically increased guns damage

fixed

PR #9989, opened 2022-04-10
Splints work as tourniquets

QualityVan:
## About The Pull Request
As title. Slap a splint on a missing limb to stop bleeding and pain. Won't expire on their own.
Also reduced the bleeding from missing limbs so that an arm and its hand being gone only bleed as much as a hand alone did before.
Alternative to #9986 
Also fixes a minor bug where if you amputated a limb it wouldn't properly count as amputated.

## Why It's Good For The Game
Field treatment option along with qc, slow to apply and still leaves you missing a limb but prevents the other penalties.
Also, funny as it was for someone to hit 0 blood in 100 seconds, that's a little excessive.

## Changelog
:cl:
add: Use splints on missing limbs to apply a tourniquet, stopping the bleeding and pain
balance: Missing limbs bleed half as much
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -260,7 +260,20 @@
 			return
 
 		if(affecting.limb_status & LIMB_DESTROYED)
-			to_chat(user, span_warning("[user == M ? "You don't" : "[M] doesn't"] have \a [limb]!"))
+			if(affecting.limb_status & LIMB_AMPUTATED)
+				to_chat(user, span_warning("[user == M ? "Your" : "[M]'s"] missing [limb] is already sealed up!"))
+				return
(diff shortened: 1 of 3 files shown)


EvilMelonTheSecond:
Does it still break from damage or neurotoxin spit? Not much changes whether it does or not since fighting with a missing limb is a dumb idea anyways but would still be annoying to lose the tourniquet while running due to IB damage or something

QualityVan:
> Does it still break from damage or neurotoxin spit? Not much changes whether it does or not since fighting with a missing limb is a dumb idea anyways but would still be annoying to lose the tourniquet while running due to IB damage or something

Nope, just stays forever until the limb gets replaced.

Pill-U:
imo missing an arm/leg should still be worse than loosing a hand/foot since arms/legs are a lot harder to remove

BraveMole (approved):

Issue #9898, opened 2022-04-02
Flamers can shoot through solid windows


Lumipharon:
### Description

Flamethrowers are not stopped by any type of glass.

### Test Merges

[#9824: 'Tesla gun rebalance' by goose4429 at commit 38c1559b91](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9824)
[#9772: 'TGUI updates' by TiviPlus at commit 0984f0187b](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9772)
[#9776: 'update atomns init' by TiviPlus at commit 8f37f788e7](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9776)

### Reproduction Steps

shoot flame through windows

### Screenshots

![image](https://user-images.githubusercontent.com/7869430/161405406-7a8e3831-89af-470b-b73d-15f4dd047363.png)

mrhugo13:
the relevant code
![Code_2022-04-11_12-29-23](https://user-images.githubusercontent.com/17747087/162722050-315f7964-a15a-4f54-912b-a908b4ef7371.png)

PR #10006, opened 2022-04-12
Flamethrower fixes

Lumipharon:
## About The Pull Request
Stops flamethrowers from going through windows, and allows them to go through vehicles and deployables.

Turns out flamers use throwpass for these checks, so the coverage change effected these objects.

Closes https://github.com/tgstation/TerraGov-Marine-Corps/issues/9898

## Changelog
:cl:
fix: Flamers will no longer go through windows
fix: Flamers can go through vehicles and deployables again
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -21,6 +21,7 @@
 	var/junction = 0 //Because everything is terrible, I'm making this a window-level var
 	var/damageable = TRUE
 	var/deconstructable = TRUE
+	throwpass = FALSE
 
 //I hate this as much as you do
 /obj/structure/window/full
--- a/code/modules/projectiles/guns/flamer.dm
+++ b/code/modules/projectiles/guns/flamer.dm
@@ -222,7 +222,7 @@
 			turfs_to_burn -= turf_to_check
 			continue
 		for(var/obj/object in turf_to_check)
-			if(!object.density || object.throwpass || istype(object, /obj/structure/mineral_door/resin) || istype(object, /obj/structure/xeno))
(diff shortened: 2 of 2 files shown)


code/game/objects/structures/window.dm
```diff
@@ -21,6 +21,7 @@
 	var/junction = 0 //Because everything is terrible, I'm making this a window-level var
 	var/damageable = TRUE
 	var/deconstructable = TRUE
+	throwpass = FALSE
```
hyper2snyper, line 24: no throwing things through windows?

hyper2snyper (changes requested):

code/game/objects/structures/window.dm
Lumipharon: @hyper2snyper Throwpass does not actually have anything to do with throwing, it applies to projectiles.
...But windows have their own hit code which ignores this and just checks if it's a laser.
...and flamethrowers, which checks it to see if fire can go through it or not.

Yes it's dumb, I'm pretty sure it's been like that forever, maybe that used to share a var?

TiviPlus: you can just rename it in anothe r pr

TiviPlus:
Apart from the windows are these really bugs

Lumipharon:
> Apart from the windows are these really bugs

Stuff like motorbikes or sentries only have throwpass = false because having to click on them to hit them is really annoying, and means you can't do so from off screen. But they're not really blocking a tile, and you can shoot flamethrowers past stuff like vending machines or other dense but not opaque objects.

So it's really inconsistant that you can't flame someone because they're on or behind a bike, yet they can happily shoot you.

PR #10011, opened 2022-04-13
Fixes Tyr module so it doesn't "stick" on sprites

MrHorizons:
## About The Pull Request
The Tyr had an `attachment_layer` var that made it stay on a marine's sprite once equipped, even after taking it off. No other armor module seems to have this var, and in my testing, it didn't seem like the lack of it made a difference.

## Changelog
:cl:
fix: Fixed the Tyr attachment remaining on sprites even after being taken off.
/:cl:


hyper2snyper:
This isn't a fix since that var gets it to display in the correct layer. If you want to fix it see about fixing it in attachmenthandler

PR #10012, opened 2022-04-14
Reworks TE cell sprite and overlay.

Crystal9156:
## About The Pull Request

Changes TE energy cell's charge meter from a 2x1 to a 2x5 for enhanced visibility, making it easier on the eyes and to tell at a glance. This was originally going to be before Lumi fixed the energy cells but they added overlays so I changed the overlays instead.

## Why It's Good For The Game

Makes it easier on the eyes instead of a tiny 2 pixel, it's now a 10 pixel in total color so that you can tell from even the barest of glances.

## Changelog
:cl:
qol: Makes Energy cell icon easier to check at a glance.
/:cl:
Old:
TECELLold1

New:
TECellNew1


Lumipharon:
@Crystal9156 post the screenshot again. TBH though, now that the states all have different colours I think visibility is fine as it has good contrast.

TiviPlus (approved):

TiviPlus:
@Kuro020

PR #10014, opened 2022-04-14
Removes external wounds

QualityVan:
## About The Pull Request
Wounds exist as a whole thing on top of limbs, tracking their individual damage and description and treatedness. The only gain from this is that if you examine someone you get to see what their wounds look like. This pr kills wounds as a datum with the exception of internal bleeding, which is refactored to process itself instead of being handled by the limb. Everything should work the same, with the sole exception of internal bleeding not contributing directly to the limb's brute loss.

## Why It's Good For The Game
Removes a level of abstraction which generally isn't thought about ingame and only really offers flavor (individual wound descs) as a benefit.

## Changelog
:cl:
refactor: Limbs have damage themselves instead of separating it into wounds
refactor: Internal bleeding doesn't count directly as damage on a limb.
/:cl:

### DIFF
--- a/code/modules/surgery/internal_bleeding.dm
+++ b/code/modules/surgery/internal_bleeding.dm
@@ -17,9 +17,8 @@
 
 /datum/surgery_step/fix_vein/can_use(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/limb/affected, checks_only)
 	if(affected.surgery_open_stage >= 2)
-		for(var/datum/wound/W in affected.wounds)
-			if(W.internal)
-				return 1
+		if(affected.wounds.len)
+			return 1
 
 /datum/surgery_step/fix_vein/begin_step(mob/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/limb/affected)
 	user.visible_message(span_notice("[user] starts patching the damaged vein in [target]'s [affected.display_name] with \the [tool].") , \
@@ -31,10 +30,9 @@
 	user.visible_message(span_notice("[user] has patched the damaged vein in [target]'s [affected.display_name] with \the [tool]."), \
 		span_notice("You have patched the damaged vein in [target]'s [affected.display_name] with \the [tool]."))
 
-	for(var/datum/wound/W in affected.wounds)
-		if(W.internal)
-			affected.wounds -= W
-			affected.update_damages()
+	for(var/datum/wound/W AS in affected.wounds)
+		affected.wounds -= W
+		affected.update_damages()
 	if(ishuman(user) && prob(40))
 		user:bloody_hands(target, 0)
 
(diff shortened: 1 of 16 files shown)


Lumipharon:
What's the benefit of this?

code/modules/surgery/internal_bleeding.dm
```diff
@@ -31,10 +30,8 @@
 	user.visible_message(span_notice("[user] has patched the damaged vein in [target]'s [affected.display_name] with \the [tool]."), \
 		span_notice("You have patched the damaged vein in [target]'s [affected.display_name] with \the [tool]."))
 
-	for(var/datum/wound/W in affected.wounds)
-		if(W.internal)
-			affected.wounds -= W
-			affected.update_damages()
+	for(var/datum/wound/internal_bleeding/W in affected.wounds)
+		qdel(W)
```
TiviPlus, line 34: Could you use QDEL_LIST() here

TiviPlus (approved):

Al-1ce:
There is literally no point or benefit in this game. Why?

QualityVan:
> There is literally no point or benefit in this change. Why?

Is there a reason you'd want wounds to keep existing as is?

BraveMole (approved):

Al-1ce:
> Is there a reason you'd want wounds to keep existing as is?


It's harmless fluff. Additionally, it previously served as a quick "heads up" to any general marines who might need to tend an ally should a medic not be present. (As most marines do not carry scanners/or have valky)

There is literally no reason for this change. You listed removing "abstraction/flavor" as if we haven't seen tens if not hundreds of PRs similar in nature shut down for remvoing just that - harmless, if not somewhat helpful fluff.

You're the one under the burdon of reason, **not me**. You didn't respond or give a reason to the other person in this PR asking 'why', either. You should have a reason for why you'd want something merged, rather than demanding one from me. Needless to say, I'm not sure how this slipped by Mole who has denied similar PRs just like this. Fluff is exactly that - there for flavor. Might as well remove gun descriptions and char-bio's while you're at it.

QualityVan:
> Additionally, it previously served as a quick "heads up" to any general marines who might need to tend an ally

So for this specifically, I actually kept that behavior of damage being visible on examine. It's still not usable most of the time because they only display if there's no clothing covering that limb, but it's there.

In terms of reason, that's in the pr body. It's an entire level of mechanics which add obscurity, being almost entirely invisible in game and then causing confusion when they actually matter for once (do you know how many people asked me why the bandage chain pr would target healed limbs? it was several). It's not that I want to remove it because it's flavor, but because it's detrimental to understanding and upkeep with the only benefits being flavor text that, again, is invisible 99% of the time because it only shows up on examine for undressed mobs and, if you consider it a benefit, another step in the limb infection chain.

PR #10028, opened 2022-04-18
Robotic Limbs no longer get shrap

Arctrooper8:
## About The Pull Request
Robotic limbs no longer get shrapnel.

## Why It's Good For The Game
Reminder
Others besides MJP have also voiced the suggestion that robots should not have shrap (Zack for example)

Overall though, the entire point of combat robots were to avoid the silly problems that humans face, at the cost of no one caring about you and the complete inability to use a majority of armor designed for humans.
With them losing the ability to scan when the new med scanner rework comes out (Soon TM) there is even more reason to remove their inability to get shrapnel which is difficult to self diagnose the location of without harming yourself via jostling the shrapnel.

## Changelog
:cl:
balance: The TGMC have stopped using recycled soda cans to create their robotic personnel
/:cl:


TiviPlus (approved):
>using mjp and zack as "approval"

Shoulve closed this just for that

Arctrooper8:
> > using mjp and zack as "approval"
> 
> Shoulve closed this just for that

image

Lumipharon:
Machinery gets fucked when rubbing against it's own parts without oil. What do you think happens when literal chunks of shrapnel get in there?

PR #10033, opened 2022-04-18
Removes body types (Ecto, Meso, Endo)

Arctrooper8:
## About The Pull Request

Removes the ability to choose your body type and simplifies the DMI folder as a result, it has been tested.

## Why It's Good For The Game

Currently each ethnicity has 6 torso sprites (3 for each gender). With a total of 23 ethnicities this leads to a whopping total of 138 separate torso sprites. This makes the human sprites difficult to maintain or improve due to the fact each sprite must be individually changed. 
So many sprites and an entire system dedicated to choosing a single torso sprite that will be covered by: Underwear, Uniform, Webbing, Armour, Modules, and maybe even a cape.
Instead of having 3 sprites of varying quality per gender (that all share the same side sprite by the way, so the Ecto, Meso, Endo stuff is really only a couple of abdomen pixels considering underwear) We can have one good torso sprite that we can focus on.

TLDR: So many sprites is difficult to maintain or update, and simply having so many sprites and an entire system to make sure it works is silly when most people are indifferent and will only see it at the beginning of the round when they wake from cryosleep.

## Changelog
:cl:
del: It turns out that in a universe where everyone is the same height and shape, body types do not exist.
/:cl:


Lefunny:
> ## About The Pull Request
> Removes the body types (Ecto, Meso, Endo) in preparation for the sprite change. Yes, I tested it.
> 
> Removes more CM rubbish, I mean seriously what the hell even is Ecto, Meso, Endo? Everyone is quite literally the same size and the torso is the one part of the body that's already quite covered with things such as underwear and jumpsuits. 00 A profound hatred of CM has been shared this day Not the specific quote but Kuro said I could remove the Ecto, Meso, Endo stuff somewhere in that conversation, I'm just doing it now.
> 
> ## Changelog
> 🆑 del: It turns out that in a universe where everyone is the same height and shape, body types do not exist. /🆑

Just answering the question they are genetic types, like if you are big, smaller and etc most gym ppl talk about it

Kuro020 (approved):

Arctrooper8:
> Just answering the question they are genetic types, like if you are big, smaller and etc most gym ppl talk about it

Thanks for answering, it still seems weird to have that in something like SS13.

Arctrooper8:
The plan has changed and will now revolve around fixing the pixel errors on the human sprites.

Arctrooper8:
Reopening because I am confused and Kuro and Tivi need to discuss whether or not body types are actually going.

Arctrooper8:
Also, look at these beautiful lines. So nice and organized:
image
Much better than what we have currently.

PR #10038, opened 2022-04-20
Adjust layer order for mob item rendering

Lumipharon:
## About The Pull Request
The order in which equipped items are rendered on mobs is a little bit funky. Suit slot items and back items are currently rendered underneath everything head related. So if you have a large gun or backpack on, it appears underneath your hair or large helmet.

This PR means you get this:
image
Instead of this:
image

The system isn't perfect either way, as you still get issues with large guns (like the above RR) combined with large helmets like the jaeger causing clipping when facing downwards but in my opinion this is the lesser of two evils, as it results in significantly less clipping over all.
Also Jaeger helmets are bizarrely huge.

The ideal solution would be to have render order based on direction faced, or to have over/underlays that can apply to other item layers, but that's not something I would know how to manage.

## Why It's Good For The Game
Guns ramming into the back of your skull is jank.

## Changelog
:cl:
code: Back and suitslot items render over head layers - guns will no longer grow into the back of your skull when stored.
/:cl:


TiviPlus:
@Kuro020

TiviPlus:
Ok kuro ignored if it looks bad his problem

PR #9434, opened 2022-01-28
Implements Rohesie's smoothing system, creates/ports a number of new wall sprites.

RipGrayson:
## About The Pull Request

First of all massive disclaimer, I didn't write 99% of this code, Rohesie created the smoothing system here https://github.com/tgstation/TerraGov-Marine-Corps/pull/4462 but left before it could be implemented.
I'm bringing it back mainly because our current smoothing system is utterly ancient and got in the way of respriting cave walls.

This PR implements a new smoothing system and uses it to support an original re-sprite of mineral walls on LV, Chigusa, and Magmoor. While I was at it, I also ported cult walls and jungle walls from CM and one generic set of walls from Eris because the current one dimensional walls look horrible next to cave walls with 3/4ths perspective on them (once again, Rohesie had already converted the CM walls to the new system, I'm responsible only for bug testing them for missing icon states).

## Why It's Good For The Game

The current smoothing system is ancient and gets in the way of better sprites for our maps, it's time it was replaced. New sprites might also be good, players won't have to pretend that a one dimensional plane represents a cave wall anymore.

## Changelog
:cl:
add: Implemented Rohesie's smoothing system.
imageadd: Created new cave wall sprites.
imageadd: Ported jungle wall sprites from CM.
imageadd: Ported cult wall sprites from CM.
imageadd: Ported a set of wall sprites from Eris to replace old Sulaco walls.
imagedel: Deleted old Sulaco wall graphics.
/:cl:

PR #10039, opened 2022-04-20
Fixes invisible windows

RipGrayson:
## About The Pull Request

When I implemented the smoothing system by Rohesie here https://github.com/tgstation/TerraGov-Marine-Corps/pull/9434 there was a problem that I didn't fix. Essentially the system is smart enough to find a logical connector between structures/turf but if it can't find a valid icon then the structure just becomes invisible due to incorrect icon state.

Essentially if you do something like 
spacetiles3
then you'll get an invisible window because it doesn't have a correct icon for being surrounded on three sides. The smoothing system sets the icon that it *should* have, but it doesn't exist so we get invisibility.

This PR fixes this issue for most windows by adding new custom icons for the missing directional variants.

## Why It's Good For The Game

When I originally did the smoothing PR I went over the maps to eliminate invisible window configs, but it's popping up again due to an increase in mapping activity. It's time for a permanent fix so people don't have to wonder why their windows are invisible in game.

## Changelog
:cl:
fix: Windows should no longer go invisible when surrounded by 3 or more walls.
imageadd: Created a number of windows directional variants to fix bug causing invisible windows.
/:cl:


Derrick2Electricboogaloo:
Ah yes! Finally, this has been bugging me for so long.

PR #10050, opened 2022-04-24
Injector buff

goose4429:
## About The Pull Request

Doubles the volume of basic injectors (Bica, Kelo, Trico, Trama, Dylo), to 30, as well as default transfer rate, to 10. (also makes QC and Peri injectors identical)

## Why It's Good For The Game

Nobody uses basic injectors, since for same space you can have 1 pill bottle (240u), 1 pill packet (120u), 1 bottle (60u) or 2 injectors (30u). Injector being identical to _single_ pill is meh. Now they'll be used more, and allow some more med meta instead of every sane marine running around with pill bottles.

## Changelog
balance: basic injectors now have double the meds.
/:cl:


BraveMole (changes requested):
Buffing something because "it is not used" is absolutly not enough a reaso,

goose4429:
> Buffing something because "it is not used" is absolutly not enough a reason

This will disperse the meta a little.
I see that that new (RU) players tend to use the injectors as well, and this will help them play their own way instead of following "make pills" meta. I feel that it's a right thing.

Sounds like a good reason?

InterroLouis:
This looks good to me because injectors are a pretty bad option compared to pill bottles.

Mrrpip:
I don't think I'm convinced, they can all be refilled at the tadpole right now, while tricord, tramadol (And combat) can be refilled at the many nanomeds around the map, provided engineers can be bothered (They usually are not) or the alamo.

Used well, at the moment, they're great, reusable tools. I don't think they need buffing IMO. Just convince others to stop throwing their used injectors away.

TiviPlus:
This does not address the core issue which is that pills self use, hypo and injector are all instant

This really feels like powercreep

If you want to look more at autoinjectors feel free to tag me in disc

PR #10051, opened 2022-04-24
Chem dispenser regen buff

goose4429:
## About The Pull Request

Title. x2 regen speed.

## Why It's Good For The Game

Currently it takes about 2 hours to fully recharge chem dispenser. In longer round, this means no chemistry, unless you change the battery. And if you can change the battery anyway, why not give a little more energy in the first place. QoL.

## Changelog
:cl:
balance: Increased chem dispenser energy regen
/:cl:

### DIFF
--- a/code/modules/reagents/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/machinery/chem_dispenser.dm
@@ -16,7 +16,7 @@
 	var/obj/item/cell/cell
 	var/powerefficiency = 0.1
 	var/amount = 30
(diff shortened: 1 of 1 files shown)


Naaanii:
Buff labeled as qol moment.

usnpeepoo:
If the battery can be changed why buff it’s regeneration so the main way of recharging it isn’t used?

BraveMole (changes requested):
We don't like chem here, so no

PR #10045, opened 2022-04-22
makes the AMRs scope unuseable on the move

Pariah919:
## About The Pull Request

https://user-images.githubusercontent.com/46353991/164780296-59647659-facc-42d3-987c-e57d724f0a79.mp4

https://streamable.com/lm0fyv

## Why It's Good For The Game
you should really, not be able to move while using this thing

## Changelog
:cl:
balance: T-26 antimat's scope can no longer be used on the move
/:cl:

PR #10052, opened 2022-04-24
T-26 Slowdown remove

goose4429:
## About The Pull Request

Title.

## Why It's Good For The Game

**TL;DR** slowdown with IFF and aim is super bad.

1. Slowdown (3) on hit means the beno will NOT be able to escape, unless it was already running away. It is a very strong debuff when benos is trying to engage multiple marines, as it both makes harder to avoid bullets and  increases retreat time. And the T-26 holder can shoot from any point with firelane and without regard to friendlies (IFF), as opposed to most other sources of slowdown (shotguns, explosions).
2. Slowdown with added aim means you will land 100% of your shots.

Rather than increasing it's price to the skies, I think it's still a fun weapon everyone should be able to get, so overall _small_ nerf had been done. IMO it's still an extremely OP weapon in scenarios like FOB defense, but now not as good in others. 

## Changelog
:cl:
balance: T-26, Removed on-hit slowdown
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -128,9 +128,8 @@
 
 //Define sniper laser multipliers
 
-#define SNIPER_LASER_DAMAGE_MULTIPLIER 1.7 //+70% damage vs the aimed target
-#define SNIPER_LASER_ARMOR_MULTIPLIER 1.7 //+70% penetration vs the aimed target
-#define SNIPER_LASER_SLOWDOWN_STACKS 3
+#define SNIPER_LASER_DAMAGE_MULTIPLIER 1.5 //+50% damage vs the aimed target
+#define SNIPER_LASER_ARMOR_MULTIPLIER 1.5 //+50% penetration vs the aimed target
 
 //Define lasrifle
 #define ENERGY_STANDARD_AMMO_COST 20
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -128,8 +128,8 @@
 
 //Define sniper laser multipliers
 
-#define SNIPER_LASER_DAMAGE_MULTIPLIER 1.5 //+50% damage vs the aimed target
-#define SNIPER_LASER_ARMOR_MULTIPLIER 1.5 //+50% penetration vs the aimed target
+#define SNIPER_LASER_DAMAGE_MULTIPLIER 1.7 //+70% damage vs the aimed target
+#define SNIPER_LASER_ARMOR_MULTIPLIER 1.7 //+70% penetration vs the aimed target
 
 //Define lasrifle
 #define ENERGY_STANDARD_AMMO_COST 20
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -61,7 +61,7 @@ Note that this means that snipers will have a slowdown of 3, due to the scope
 	flags_gun_features = GUN_WIELDED_FIRING_ONLY|GUN_AMMO_COUNTER|GUN_IFF
 	starting_attachment_types = list(/obj/item/attachable/scope/antimaterial, /obj/item/attachable/sniperbarrel)
 
-	fire_delay = 2.5 SECONDS
+	fire_delay = 2.35 SECONDS
 	burst_amount = 1
 	accuracy_mult = 1.50
 	recoil = 2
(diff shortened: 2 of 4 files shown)


Naaanii:
4 nerfs in one PR is a bit much to be fair. While AMR is up there towards the more overpowered tiers of weapons, this might completely make it worthless and make people default back to the T-81 instead.

goose4429:
> 4 nerfs in one PR is a bit much to be fair. While AMR is up there towards the more overpowered tiers of weapons, this might completely make it worthless and make people default back to the T-81 instead.

It's 3 nerfs and 1 buff. But anyway, thanks for feedback - in it's current state, do you really think it's overboard?

Naaanii:
> > 4 nerfs in one PR is a bit much to be fair. While AMR is up there towards the more overpowered tiers of weapons, this might completely make it worthless and make people default back to the T-81 instead.
> 
> It's 3 nerfs and 1 buff. But anyway, thanks for feedback - in it's current state, do you really think it's overboard?

Even for an absurdly strong weapon like the AMR, it'd be better to remove its power from one aspect rather than nerf a lot of things at once.

As it stands the AMR has a lase that removes the need to actually aim and hit Xenos once a target is marked, NVG, long range, slowdown on hit and the ability to move and shoot while scoped while also being only a 65 point req weapon.

It can remain strong if it were harder to acquire and less spammable via making it more expensive which is fair. Or one of its strong aspects could be nerfed like what #10045 tries to do. If the weapon is still oppressive, you can always nerf it again later rather than completely gut it to the point where it goes unused and gets buffed again.

DeltaFire15:
> > > 4 nerfs in one PR is a bit much to be fair. While AMR is up there towards the more overpowered tiers of weapons, this might completely make it worthless and make people default back to the T-81 instead.
> > 
> > 
> > It's 3 nerfs and 1 buff. But anyway, thanks for feedback - in it's current state, do you really think it's overboard?
> 
> Even for an absurdly strong weapon like the AMR, it'd be better to remove its power from one aspect rather than nerf a lot of things at once.
> 
> As it stands the AMR has a lase that removes the need to actually aim and hit Xenos once a target is marked, NVG, long range, slowdown on hit and the ability to move and shoot while scoped while also being only a 65 point req weapon.
> 
> It can remain strong if it were harder to acquire and less spammable via making it more expensive which is fair. Or one of its strong aspects could be nerfed like what #10045 tries to do. If the weapon is still oppressive, you can always nerf it again later rather than completely gut it to the point where it goes unused and gets buffed again.

Ah yes, because "weapon gets used less than other weapon" obviously means that it needs to be buffed and not that it is simply a matter of preference.
It is still effective, it's still highly damaging. It still has its niche of scoping targets, high single shot damage as opposed to the autosniper's more suppressive style, dotting which makes any attempts to catch bullets worthless (see: boiler + defender against normal sniper rifles). Just because it got buffed thrice because "it got used less than the other req sniper" doesn't mean that returning it to a state still somewhat better than before is bad. Base damage buff is still there, 80(!!) AP is still a thing, it's still cheaper than before to buy the gun itself.

InterroLouis:
> > Even for an absurdly strong weapon like the AMR, it'd be better to remove its power from one aspect rather than nerf a lot of things at once.
> > As it stands the AMR has a lase that removes the need to actually aim and hit Xenos once a target is marked, NVG, long range, slowdown on hit and the ability to move and shoot while scoped while also being only a 65 point req weapon.
> > It can remain strong if it were harder to acquire and less spammable via making it more expensive which is fair. Or one of its strong aspects could be nerfed like what #10045 tries to do. If the weapon is still oppressive, you can always nerf it again later rather than completely gut it to the point where it goes unused and gets buffed again.

The AMR laser doesn't 'remove the need to aim and hit xenos', it removes the ability to lead your target by locking you into having to click directly on the lased target in order to fire. As has been pointed out many times, this means that if you are moving in anything but a straight line directly at or directly away from the sniper, there's a good chance they'll just wiff their shot. Also 10045 just got merged. As far as I care the damage multiplier reduction should stay, it seemed an unnecessary buff to me, but the slowdown stacks(which only happen while lased on a target) ought to stay simply because that's what makes it a good support weapon for preventing xenos from escaping groups of marines by the skin of their teeth. 
Also it should state in your excel sheet that the TL-127 stats as displayed are with aim mode on, and maybe add the stats for it while aim mode is off.

goose4429:
In spite of #10045 actually getting merged, I think it is fair to only remove slowdown. Slowdown is super bad, other stuff is not that important, as mentioned. Thanks for the feedback to the orators above.

Pariah919:
Because of how the laser works you'll miss 75% of your shots inorder to land a single one, this isn't the major problem with the AMR in my opinion, the problem is that you can move with the scope.

This is just unnecessary, and doesn't really fix the main issue with the gun.

BraveMole (approved):

PR #10068, opened 2022-04-26
TAT tweaks

Lumipharon:
## About The Pull Request
Changes the behavior of TAT ammo.

APCR no longer causes an explosion, just a 1 tile flash (same as AP rockets). This was causing it to do around 240-260 damage instead of 160. Instead, it will do a short (0.5) stagger, slow and 2 tile knockback. **Direct damage has been buffed to 200 as well.** So a net 40-60 (depending on armour) damage nerf, as well as no AOE.

Additionally, for both APCR and APHE, their piercing behavior has been changed so that every time they pierce something, their max range is reduced. This stops the silliness of projectiles blasting into the distance and possibly killing benos through like 20 tiles of solid rock by fluke.
For APHE this is 10 tiles per pierce,
For APCR this is 5 tiles per pierce.

So for example, you shoot through a 2 tile deep resin wall that is 7 tiles away from you with APHE. This means you've effectively travelled 27 tiles, so you'll reach your max 3 tiles later.

I have also reverted #9843 as APHE no longer trigger explosions on turfs at all (and APCR doesn't matter anymore), so this extra code isn't needed.

If you think this nerfs APCR too much, let me know.

Edit: Also forgot the most important part. This also fixes a bug where apcr can explode 3 (possibly more) times per shot.

## Why It's Good For The Game
Makes APCR less oppressive.
Makes piercing less janky.
Multiple explosions per shot is bad.

## Changelog
:cl:
balance: APCR TAT rounds no longer trigger light explosions, this reduces it's damage from around 260 to 160.
add: APCR and APHE TAT are now slowed by anything they pierce, reducing the total distance they can travel.
fix: Fix a bug where APCR TAT rounds can explode multiple times.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1490,7 +1490,7 @@ datum/ammo/bullet/revolver/tp44
 	hud_state = "shell_he"
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING|AMMO_PASS_THROUGH_TURF|AMMO_PASS_THROUGH_MOVABLE
 	shell_speed = 4
-	damage = 160
+	damage = 200
 	penetration = 40
 	sundering = 65
 
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1482,6 +1482,9 @@ datum/ammo/bullet/revolver/tp44
 /datum/ammo/rocket/atgun_shell/drop_nade(turf/T)
 	explosion(T, 0, 2, 3, 2)
 
+/datum/ammo/rocket/atgun_shell/on_hit_turf(turf/T, obj/projectile/P)
+	P.proj_max_range -= 10
+
 /datum/ammo/rocket/atgun_shell/apcr
 	name = "tungsten penetrator"
 	hud_state = "shell_he"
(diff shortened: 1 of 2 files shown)


Frampox:
wow, it's actually good

goose4429:
10/10 would fix TAT again. Good PR.

novaepee:
> Still offscreen thoug

It's going to be offscreen no matter what.

Al-1ce:
> Still offscreen though


Yes. Artillery will do that.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1492,7 +1495,18 @@ datum/ammo/bullet/revolver/tp44
 	sundering = 65
 
 /datum/ammo/rocket/atgun_shell/apcr/drop_nade(turf/T)
-	explosion(T, 0, 0, 1, 0)
+	explosion(T, 0, 0, 0, 1)
```
BraveMole, line 1498: Not a fan, remove that for now and nerf again if needed

BraveMole (changes requested):

code/modules/projectiles/ammo_datums.dm
Lumipharon: damage buffed to 200 per discord

BraveMole (approved):

PR #10074, opened 2022-04-27
fix defib fumble time

Lumipharon:
## About The Pull Request
Fixed defib fumble time for unskilled mobs.

So it turns out that fumble time for defibbing someone was written... poorly.

If you have medical skill of 2 or above (corpman) you ignore fumble time entirely. If you're below 2 however, you have a fumble time of three seconds for medical skill 1 (SL's), or ONE SECOND for medical skill 0 (ungas).

I have no idea why it was like this, it's the only fumble check written in this way and it's written correctly in the same file by the same PR... but it's been like this for 2 years.

## Why It's Good For The Game
Literal greyshirts knowing how to use defibs better than people with basic medical training is bad.

## Changelog
:cl:
fix: Mobs with 0 medical skill will no longer know how to operate defibs better than those with basic medical training.
/:cl:

### DIFF
--- a/code/game/objects/items/defibrillator.dm
+++ b/code/game/objects/items/defibrillator.dm
@@ -157,7 +157,7 @@
 	if(skill < SKILL_MEDICAL_PRACTICED)
 		user.visible_message(span_notice("[user] fumbles around figuring out how to use [src]."),
 		span_notice("You fumble around figuring out how to use [src]."))
-		var/fumbling_time = SKILL_TASK_AVERAGE - ( SKILL_TASK_VERY_EASY * ( SKILL_MEDICAL_PRACTICED - skill ) ) // 3 seconds with medical skill, 5 without
+		var/fumbling_time = SKILL_TASK_AVERAGE - (SKILL_TASK_VERY_EASY * skill) // 3 seconds with medical skill, 5 without
 		if(!do_after(user, fumbling_time, TRUE, H, BUSY_ICON_UNSKILLED))
 			return
 	else


code/game/objects/items/defibrillator.dm
```diff
@@ -157,7 +157,7 @@
 	if(skill < SKILL_MEDICAL_PRACTICED)
 		user.visible_message(span_notice("[user] fumbles around figuring out how to use [src]."),
 		span_notice("You fumble around figuring out how to use [src]."))
-		var/fumbling_time = SKILL_TASK_AVERAGE - ( SKILL_TASK_VERY_EASY * ( SKILL_MEDICAL_PRACTICED - skill ) ) // 3 seconds with medical skill, 5 without
+		var/fumbling_time = SKILL_TASK_AVERAGE - (SKILL_TASK_VERY_EASY * skill) // 3 seconds with medical skill, 5 without
```
QualityVan, line 160:
```suggestion
		var/fumbling_time = SKILL_TASK_AVERAGE - (2 SECONDS * skill) // 3 seconds with medical skill, 5 without
```
Same thing, but easier to glance at.

Lumipharon: Yeah I thought so as well, but I decided to keep it consistant with the one earlier in the file and elsewhere. Happy to change it if maintainers think it's better though.

BraveMole: No magic numbers

TiviPlus: it doesnt matter imo

PR #10079, opened 2022-04-28
Shuffle Big Red disk locations

RipGrayson:
## About The Pull Request
Apparently Big Red disks were either too easy or too hard, this PR adjusts them to the following locations
Red has been moved from Marshal's to Atmos. 
Blue is a room farther north in ETA. 
Green has been moved from Viro into Admin.

## Why It's Good For The Game

People have been complaining for a bit about Big Red disk locations, this should make things better balanced.

## Changelog
:cl:
balance: Shuffled Big Red nuclear generator computers.
balance: Moved a single platinum miner.
/:cl:

PR #10086, opened 2022-04-28
Adds a tank prop for mappers, adds 4 modulars to Big Red using it.

RipGrayson:
## About The Pull Request

Our broken version of the tank is increasingly being used by mappers, this shouldn't be happening so I've created a modular props for both tanks and APCs for use in mapping. To show it off I also created a couple modulars for Big Red that use the new props.

This PR shouldn't be merged before https://github.com/tgstation/TerraGov-Marine-Corps/pull/10079 due to changing of disk locations.

## Why It's Good For The Game

This has needed done for a while now, there's already one map in our codebase being actively exploited due to bad placement of a buggy tank, and some more WIP maps using the broken CM tank as prop. It's time we have a prop version so mappers can use the tank without worrying about marines interacting with it.

## Changelog
:cl:
add: Added a prop version of the tank for mappers.
add: Created a handful of new modular maps for Big Red.
add: Added a prop version of CM's APC for mappers.
imageadd: Ported CM APC icons for mapper use.
/:cl:


_maps/modularmaps/big_red/bigredatmosvar5.dmm
```diff
@@ -0,0 +1,1713 @@
...
+"aO" = (
+/obj/structure/closet/firecloset/full,
+/turf/open/floor,
+/area/bigredv2/outside/filtration_plant)
+"aP" = (
+/obj/structure/reagent_dispensers/fueltank,
+/turf/open/floor{
+	dir = 1;
+	icon_state = "bot"
+	},
```
TiviPlus, line 26: Why is the floor varedited?

TiviPlus (changes requested):
rest lgtm

_maps/modularmaps/big_red/bigredatmosvar5.dmm
RipGrayson: The base modular in this case was probably ported from CM and then adjusted and remapped as necessary. CM absolutely adores their icon_state varedits and this one slipped past me.

TiviPlus (approved):

PR #10088, opened 2022-04-28
Replaces DETACHABLE_HEAD with UNDETACHABLE_HEAD

Arctrooper8:
## About The Pull Request

Removes head replacement surgery and changes the trait DETACHABLE_HEAD to instead prevent decapitation entirely.

## Why It's Good For The Game

The head reattachment surgery still doesn't work for combat robots (When you don't test code) and realistically it's almost never going to happen since combat robots have no fashion reason to not wear helmets like humans (It's all about that visor style). 

This also affects synthetics however due to their limited combat prowess and already reduced incoming damage they will be fine, it's not as if synthetics should be seeking combat anyway.

Head reattachment is a fun idea but just doesn't work in execution, robot heads don't have a visible sprite when it goes flying off, the surgery still doesn't work, the scenario in which a robot actually gets decapitated is so rare that it will never actually be seen in game, the surgery/mechanic itself isn't well known which can lead synthetics who have been decapitated to be written off as perma.

Overall, the synthetic head reattachment is an interesting idea that doesn't work well in execution. Or just doesn't work at all in the case of combat robots.

## Changelog
:cl:
add: Combat robots and Synthetics will no longer lose their heads.
del: Head reattachment surgery is no longer a thing.
/:cl:


Derrick2Electricboogaloo:
Head reattachment did not work for synths? Since when I have had my head reattached before.

Arctrooper8:
> Head reattachment did not work for synths? Since when I have had my head reattached before.

It doesn't work for combat robots (even after Tivi's fix), excuse me if I made a mistake in my PR description.

Arctrooper8:
Rattle me bones

PR #10094, opened 2022-04-29
Combat injectors for Corpsman

goose4429:
## About The Pull Request

Add syringe case containing 3 combat injectors to the Corpsman vendor. Costs 3 points.
Empty case for 2pts.

## Why It's Good For The Game

Combat injectors are a nice no-brainer, and we have tricord in the vendor already. Second variant of lifesaver bag has these too. Hopefully it will add more variety to loadouts, and avoid powercreep (you can get these for free anyway).

## Changelog
🆑 
add: Added combat injector case to Corpsman vendor, and empty one.
/:cl:


EvilMelonTheSecond:
Seems kinda useless? Corpsman can already mix chems inside hyposprays from pill bottles and many times they discard the injectors they get from lifesaver belt. I suppose there is no harm in more options though.

goose4429:
> Seems kinda useless? Corpsman can already mix chems inside hyposprays from pill bottles and many times they discard the injectors they get from lifesaver belt. I suppose there is no harm in more options though.

Yeah, purely for options and QoL. IMO synaptizine, dexa+ and AC injectors, as well as tricord are useless as well.

goose4429:
> while you're at it, code in a reagent bottle case that is empty for one point per.

Done. Although 2pts, just to be sure no big abuse would be profitable.

goose4429:
Cases are actually widely available, so this PR is just "combat injectors" without the cases. Which is dumb.

PR #10095, opened 2022-04-29
Minor Rework of Research Outpost

tyeagg:
## About The Pull Request
minor rework of research outpost so it can get back into rotation

## Why It's Good For The Game
to return research outpost back into rotation

## Changelog
:cl:
balance: expanded caves slightly
balance: made lz2 more open
balance: adds another flank to both lz's
balance: moved fog locations
balance: more gens
fix: tadpole being able to land in northern caves
/:cl:


novaepee:
Red line means remove, blue line means add, and yellow means not connected.

image
Blocked path, don't let walls be between AI nodes.

image
Useless node and blocked path.

image
Add new nodes and useless nodes abound.

image

image

image

image

image

image

image

image

RipGrayson (commented):
If these map changes are just balance in case it gets added back the changes look solid enough to me. If the consideration here is to restore it to active play I would prefer a TM first to get feedback.

BraveMole (approved):
I'm merging directly, if that's bad I'll just put back RO in not playable maps pool

PR #10103, opened 2022-04-29
Give researcher, MD, and CMO req radio access

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

Researcher and Req are deeply in love with each other, therefore it makes sense that researcher gets radio access. I was surprised that researcher doesn't have this, so this is simply a one line change.

Consequently, MD and CMO will get access to req radio.

## Changelog
:cl:
add: Give researcher, MD, and CMO req radio access
/:cl:


realroyalderg:
Yeah its better to let researchers yell at RO over req radio than general so general can stay consistent for high command to tell all squads and not researcher rambling.

PR #10107, opened 2022-05-01
Removes unmanned vehicles from Valhalla

Arctrooper8:
## About The Pull Request

Removed the two unmanned vehicles from Valhalla.

## Why It's Good For The Game

Valhalla has two unmanned vehicles in its cargo section as decoration, the problem is that these are on the AI's connection list. Now the AI can't actually do or see anything while using them, but they are still visible to the AI and make the list longer and more annoying to deal with.

## Changelog
:cl:
qol: Removed the unmanned vehicles from Valhalla.
/:cl:

### DIFF
--- a/_maps/map_files/generic/Admin_Level.dmm
+++ b/_maps/map_files/generic/Admin_Level.dmm
@@ -4690,7 +4690,6 @@
 /obj/machinery/firealarm{
 	dir = 1
 	},
-/obj/vehicle/unmanned/medium,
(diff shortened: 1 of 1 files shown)


RipGrayson (approved):
Yeah this shouldn't be in Valhalla.

TiviPlus (approved):
better fix would be to ignore droids on admin z level but this is good enough

PR #10110, opened 2022-05-01
Give robotic limbs in medbay spawn

novaepee:
## About The Pull Request

Shipside's robotic limb printer spawns with a robotic left and right arm and left and right leg.

Shipside's medbay storage spawns with a locker with 3 of each limb.

Only give to Theseus and Minerva since PoS PR might map conflict and I hate that and let Derrick work on Sulaco.

## Why It's Good For The Game

As I played CM's medical doctor (shocking, right?) and prepared for field surgery, I forgot to print out robotic limbs. I then thought to myself, "Why do I need to print out robotic limbs every round when I can just make a freaking locker AND make robotic limbs a spawn item?"

It reduces prep time for surgery and negligence of duty for people like me. It also makes sense from a gameplay and lore perspective: if people know that marines are going to lose limbs, why not have them ready before deployment starts? Robotic limbs aren't going to expire any time soon.

## Changelog
:cl:
add: robotic limbs spawning in round start; you can find them on robotic limb printer and medbay storage's surplus cyborg limb locker. Only for Theseus and Minerva for mapping reasons
qol: round start robotic limbs
balance: makes it easier to do field surgery with backpacks full of robotic limbs without breaking a sweat, but that's already expected
/:cl:

PR #10114, opened 2022-05-01
Bioprinter now prints limbs almost instantly

Arctrooper8:
## About The Pull Request

An alternative to #10110 
Makes the bio printer print limbs almost instantly.

## Why It's Good For The Game

As of now limbs take quite a while to print, and patients will often find themselves tossed into the autodoc for limb replacement surgery instead of being tended to manually.
This is a problem because manual surgery is supposed to be faster than the autodoc, and waiting so long for limbs to print is boring for the marine and doctor.

compared to #10110  It is much more sensible to speed up the bio printer itself for a variety of reasons: 
* People are already aware that the bioprinter makes limbs, 
* It doesn't further clutter medbay with another locker, 
* The bioprinter buff is given to all ships at the same time instead of being beholden to map conflicts or being forgotten in medbay reworks
* Print time is much more easier to balance/change than the locker solution.

## Changelog
:cl:
balance: The bioprinter has been overclocked with an NVIDIA TITAN V, and now prints limbs at SUPER SPEED in HIGH FIDELITY!
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -299,7 +299,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 #define SUTURE_MIN_DURATION 80
 #define SUTURE_MAX_DURATION 90
 
-#define LIMB_PRINTING_TIME 550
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

Derrick2Electricboogaloo:
I agree I think this is a better alternative then spawning them.

PR #10120, opened 2022-05-02
Updates logic that determines whether xeno is on weeds

Walarks:
## About The Pull Request

Updates the logic that is used to checked whether a xeno is on weeds to be hopefully more performant.

## Why It's Good For The Game

We can potentially have stuff that checks more frequently whether something is on weeds without being a performance hog. Also makes checking whether a xeno is on weeds more consistent.

## Changelog
:cl:
qol: xeno auto-weeding no longer places weeds when standing on weeds
refactor: changes how a xeno being on weeds is determined, shouldn't affect gameplay
/:cl:


code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -304,6 +304,10 @@
 
 	var/is_zoomed = 0
 	var/zoom_turf = null
+
+	///Type of weeds the xeno is standing on, null when not on weeds
+	var/obj/effect/alien/weeds/loc_weeds_type = null
```
TiviPlus, line 309:
```suggestion
	///Type of weeds the xeno is standing on, null when not on weeds
	var/obj/effect/alien/weeds/loc_weeds_type
```
Also is this a typepath or a reference, and if its a path why is it cast

tgmc_unit_test_focus_file.dm
```diff
@@ -0,0 +1 @@
+TEST_FOCUS(/datum/unit_test/weed_ability)
```
TiviPlus, line 1: ??

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -386,3 +389,17 @@
 	GLOB.offered_mob_list -= src
 	AddComponent(/datum/component/ai_controller, /datum/ai_behavior/xeno)
 	a_intent = INTENT_HARM
+
+/// Handles logic for weeds nearby the xeno getting removed
+/mob/living/carbon/xenomorph/proc/handle_weeds_adjacent_removed(datum/source)
+	SIGNAL_HANDLER
+	var/obj/effect/alien/weeds/found_weed = locate(/obj/effect/alien/weeds) in loc
+	if(!QDELING(found_weed))
+		return
```
TiviPlus, line 398: Why this check here and why isnt it qdeleted

TiviPlus (changes requested):

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
Walarks: From what I understand, the `COMSIG_LIVING_WEEDS_ADJACENT_REMOVED` signal is sent before the thing is deleted. Although, now that I checked again, hivemind code uses "QDESTROYING" instead so that may be better.

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
Walarks: The typepath is needed when doing `initial(loc_weeds_type.<weed_property>)`

tgmc_unit_test_focus_file.dm
Walarks: Removed, I don't know where it came from :mothdonk:

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
Walarks: Replaced `QDELING `with `QDESTROYING`

KornFlaks:
One step closer to a minion that automatically weeds...

Wayland-Smithy:
The plasma regen fails to restart when you evo and then do not move. I imagine this means the game thinks you are no longer on weeds since your mob was changed until you move. Please fix.

PR #10128, opened 2022-05-03
OD fix

goose4429:
## About The Pull Request

You will now get OD when _over_ 30u, not equal. Same for COD. 

## Why It's Good For The Game

Basically removes that one OD tick.

When you eat 2 tricord pills quickly, you get 1 tick of OD and lose 10 IQ. In case of brain damage, amt of dmg doesn't matter, only that you have some, so you will start dropping items. Similar for Peri+, COD of QC+, etc.

If you give more pills than OD, you will still get OD properly.

Before:
image

After:
image

## Changelog
:cl:
balance: OD and COD start when _over_ threshold
/:cl:

### DIFF
--- a/code/modules/reagents/holder.dm
+++ b/code/modules/reagents/holder.dm
@@ -282,21 +282,21 @@
 		if(L.reagent_check(R) != TRUE)
 			if(can_overdose)
 				if(R.overdose_threshold)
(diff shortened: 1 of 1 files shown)


goose4429:
I call it a fix, but some ppl will probably come to say that it's not, so balance flair.

PR #10130, opened 2022-05-03
Random weapon set spawners for mapping

Lumipharon:
## About The Pull Request
For mapper use only.

Adds some spawners that randomly spawns a gun from a list, along with it's associated ammo. This should allow mappers to have random spawns without the issue of the gun not matching the ammo due to being two separate spawners.

## Why It's Good For The Game
mismatched items is pain.

## Changelog
:cl:
add: Add some spawners for mapper use for random guns to spawn with ammo they actually use.

/:cl:

### DIFF
--- a/code/game/objects/effects/spawners/random.dm
+++ b/code/game/objects/effects/spawners/random.dm
@@ -568,7 +568,7 @@
 					/obj/item/ammo_magazine/pistol/standard_pocketpistol)
 
 
-////new test
+//Random spawners for multiple grouped items such as a gun and it's associated ammo
 
 /obj/effect/spawner/random_set
 	name = "Random Object"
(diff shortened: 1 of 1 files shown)


code/game/objects/effects/spawners/random.dm
```diff
@@ -566,3 +566,113 @@
...
+	. = ..()
+	if(!prob(spawn_nothing_percentage))
+		var/choice = rand(1, length(option_list)) //chooses an item on the list
+		spawned_gear_list = option_list[choice] //sets it as the thing(s) to spawn
+		for(var/typepath in spawned_gear_list)
+			if(spawned_gear_list[typepath])
+				new typepath(loc, spawned_gear_list[typepath])
+			else
+				new typepath(loc)
+	qdel(src)
```
TiviPlus, line 596: return INITIALIZE_QDEL or whatever its called instead

code/game/objects/effects/spawners/random.dm
```diff
@@ -566,3 +566,113 @@
...
+/obj/effect/spawner/random_set
+	name = "Random Object"
+	desc = "This item type is used to spawn random objects at round-start"
+	icon = 'icons/effects/landmarks_static.dmi'
+	icon_state = "random_loot"
+	/// this variable determines the likelyhood that this random object will not spawn anything
+	var/spawn_nothing_percentage = 0
+	///the list of what actually gets spawned
+	var/list/spawned_gear_list
+	///the list of lists, providing the options that can be spawned
```
TiviPlus, line 582: you should probably add an extra comment on how exactly a mapper should format and what the nested list does

TiviPlus (changes requested):

code/game/objects/effects/spawners/random.dm
Lumipharon: done

code/game/objects/effects/spawners/random.dm
Lumipharon: done

TiviPlus (approved):

PR #10132, opened 2022-05-04
Remove rocket sledge from req

Wayland-Smithy:
## About The Pull Request

See title.

## Why It's Good For The Game

MJP said so.

## Changelog
:cl:
del: Removed rocket sledge from req.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -475,7 +475,7 @@
 		else
 			stun = 1
 
-	if(!M.IsStun() && !M.IsParalyzed() && M.mob_size != MOB_SIZE_BIG) //Prevent chain stunning. Big mobs are protected.
+	if(!M.IsStun() && !M.IsParalyzed() && !isxenoqueen(M)) //Prevent chain stunning. Queen is protected.
 		M.apply_effects(stun,weaken)
 
 	return ..()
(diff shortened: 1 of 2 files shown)


Pariah919:
Silly, considering the entire gimmick of the sledge in the first place.

Wayland-Smithy:
Then why does it not effect queen? I think because the creator knew it would be silly so why is it okay for other big slow mobs?

private-tristan:
Gorgers' existences no longer hard countered by a 60 point req item? cool

Pariah919:
>Then why does it not effect queen? I think because the creator knew it would be silly so why is it okay for other big slow mobs?
>
IIRC, people got mad at it stunlocking queen, since she's pretty slow compared to all the other T3s and "Large" xenos.
I don't really see a reason for it to exist in the first place if your'e gonna extend the immunity, anyway, just delete it.

Wayland-Smithy:
If full removal will be accepted more readily then sure. I prefer this change since removing content is no fun and this way how it effects targets is also more predictable. The majority of any healthy hive is not even "large" so the overall *impact* of this is being blown out of proportion I feel.

BraveMole (changes requested):
IDK, as MJP said that's the whole hammer gimmick. 

Queen got a pass because people mald at the hammer, but I don't think she deserved that special treatment either

Wayland-Smithy:
As per MJP suggestion I have removed it from req altogether, but left it in code elsewhere for Valhalla and special events.

BraveMole:
Hammer is fine and doesn't need to be killed

Wayland-Smithy:
Okay Mole what ever you say. Just don't be surprised when this and the almost as cheap AMR come out then xenos SSD from the BS.

PR #10140, opened 2022-05-05
Harvester can switch reagents

Walarks:
## About The Pull Request

Makes harvester able to switch reagents with a new action it gives on equip. This also means it can be loaded with all of its reagents at the same time (but you still can have only 1 active/primed).

## Why It's Good For The Game

Makes harvester less clunky to use and opens up more stuff to do with the reagent switching.
Also makes the code better hopefully.

## Changelog
:cl:
expansion: makes harvester able to switch reagents, this also means it can store all of its compatible reagents at the same time
/:cl:


Kuro020:
nice

Kuro020 (approved):

code/datums/components/harvester.dm
```diff
@@ -132,31 +159,75 @@
...
+	var/datum/reagent/selected_reagent = null
+	for(var/datum/reagent/reagent_entry AS in loaded_reagents)
+		if(initial(reagent_entry.name) == selected_option)
+			selected_reagent = reagent_entry
+
+	update_selected_reagent(selected_reagent)
+
+
+/datum/component/harvester/proc/update_selected_reagent(datum/reagent/new_reagent)
+	src.selected_reagent = new_reagent
```
TiviPlus, line 213:
```suggestion
	selected_reagent = new_reagent
```

TiviPlus (approved):
rest lgtm

PR #10147, opened 2022-05-07
Mostly reverts AMR state to pre nerf/buffs, with minor changes.

Pariah919:
## About The Pull Request
Reduces damage from 153 to around 120 when lased, massively reduces pen overall.
Damge stays at 90, pen reduced to 50 compared to pre change AMR, so armor actually matters in some cases.
Reduces damage/pen to old values, before david changes.
Adds back the 15 sunder and slowdown on hit.

## Why It's Good For The Game
Moves the AMR back to being a long range support tool rather than a big dick killing machine, overall.

Changes compared to pre nerf/buff AMR is because I personally believe scope movement should stay gone, and I think it needs minor compensation with slightly more damage overall, and the pen nerf compared to old one is nearly non-noticeable overall IMO, I think defoonders and crushers should feel the difference. Otherwise, I think this PR is better than alternatives.

## Changelog
:cl:
balance: AMR mostly reverted to pre david touch state, has 90 damage and 135 when lased, pen down to 50, it slows down on hit again. Lase reverted to 1.5x. Sunder returned.
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -131,6 +131,7 @@
 
 #define SNIPER_LASER_DAMAGE_MULTIPLIER 1.5 //+55% damage vs the aimed target
 #define SNIPER_LASER_ARMOR_MULTIPLIER 1.5 //+50% penetration vs the aimed target
+#define SNIPER_LASER_SLOWDOWN_STACKS 3
 
 //Define lasrifle
 #define ENERGY_STANDARD_AMMO_COST 20
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -129,7 +129,7 @@
 
 //Define sniper laser multipliers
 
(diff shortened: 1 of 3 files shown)


code/__DEFINES/conflict.dm
```diff
@@ -129,8 +129,9 @@
 
 //Define sniper laser multipliers
 
-#define SNIPER_LASER_DAMAGE_MULTIPLIER 1.7 //+70% damage vs the aimed target
-#define SNIPER_LASER_ARMOR_MULTIPLIER 1.7 //+70% penetration vs the aimed target
+#define SNIPER_LASER_DAMAGE_MULTIPLIER 1.5 //+55% damage vs the aimed target
```
magatsuchi, line 132: ```+55%```
huh

magatsuchi (changes requested):

code/__DEFINES/conflict.dm
BraveMole, line 132:
```suggestion
#define SNIPER_LASER_DAMAGE_MULTIPLIER 1.5 //+50% damage vs the aimed target
```

PR #10152, opened 2022-05-07
Deployables cannot be deployed on a tile with any structures or machinery, as opposed to only dense ones

DeltaFire15:
## About The Pull Request
Title. We're seeing an immense rise of people trying to do wacky stuff with deployables, mainly sentries, via using them in conjunction with objects that can vary their density and as such do not initially block placement, but can be used as cover afterwards via using whichever toggle said objects have. This removes this vector of wackyness and disallows deployables from being placed on tiles with any other structure or machinery present.
Additionally, improves the check during deployment to properly reflect the state after the do_after aswell, due to the possibility of people moving things onto the tile after starting deployment and the previous iteration only checking before the do_after.

## Why It's Good For The Game
Less cheese tends to be good for the game. Except if you are a mouse.

## Changelog
:cl:
balance: Any machinery / structure on a tile now blocks deployable deployment, as opposed to only dense ones.
code: Deployables checking for blocking objects are now less prone to being bypassed during their do_after.
/:cl:

### DIFF
--- a/code/datums/elements/deployable_item.dm
+++ b/code/datums/elements/deployable_item.dm
@@ -63,7 +63,7 @@
 		if(!ishuman(user) || CHECK_BITFIELD(item_to_deploy.flags_item, NODROP))
 			return
 
-		if(item_to_deploy.check_blocked_turf(location))
+		if(item_to_deploy.check_blocked_turf(location, dense_only = FALSE))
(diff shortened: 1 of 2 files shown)


TiviPlus (approved):

Naaanii:
Uhm, I believe this also makes deployables not able to be placed on button tiles, extinguisher locker tiles and shutter tiles too.

PR #10153, opened 2022-05-07
Add statistic for acquired credits through research

Walarks:
## About The Pull Request

Add statistic for acquired credits through research in the research console and round-end stats.
## Why It's Good For The Game

Makes it possible for people to check how much req points researcher is producing, which is an interesting stat and can possibly help other players determine whether to provide more support to good researchers.
## Changelog
:cl:
qol: adds a stat for produced req points through research to researcher console and round-end, updates console UI
/:cl:

### DIFF
--- a/code/game/objects/machinery/research.dm
+++ b/code/game/objects/machinery/research.dm
@@ -107,6 +107,8 @@
 /obj/machinery/researchcomp/ui_static_data(mob/user)
 	. = ..()
 
+	var/points = GLOB.round_statistics.points_from_research
+	.["acquired_points"] = points
 	.["anchored"] = anchored
 	.["researching"] = researching
 	if(!init_resource)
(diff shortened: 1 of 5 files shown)


code/game/objects/machinery/research.dm
```diff
@@ -107,6 +107,8 @@
 /obj/machinery/researchcomp/ui_static_data(mob/user)
 	. = ..()
 
+	var/points = GLOB.round_statistics.points_from_research
+	.["acquired_points"] = points
```
TiviPlus, line 111:
```suggestion
	.["acquired_points"] = GLOB.round_statistics.points_from_research
```

TiviPlus (changes requested):

code/game/objects/machinery/research.dm
Walarks: For some reason this didn't work when I was testing it, but works now...

PR #10155, opened 2022-05-07
Times the cost of all weapons by 1.5x and increases supply point increase and exports by 1.5c.

Pariah919:
## About The Pull Request
Increases the amount you get from corpses, miners and passive points by 1.5x.
Increases all gun costs by 1.5x.

## Why It's Good For The Game
We NEED to decompress req weapon costs, and generally I think the price of everything else can be done on case by case after this PR.

## Changelog
:cl:
costs, all point gains are increased by 1.5x, all req weapons are increased by 1.5x and to be decreased/increased on a case by case basis.
/:cl:

### DIFF
--- a/code/__DEFINES/supply.dm
+++ b/code/__DEFINES/supply.dm
@@ -1,5 +1,5 @@
 //Prices of items that are accounted for by the monitor
-#define B18_PRICE 120
-#define B17_PRICE 80
-#define MINIGUN_PRICE 80
-#define SADAR_PRICE 120
+#define B18_PRICE 180
+#define B17_PRICE 120
+#define MINIGUN_PRICE 120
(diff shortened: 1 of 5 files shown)


Walarks:
> all point gains are increased by 1.5x

Update researcher credits export rewards

novaepee:
inflation in tgmc is here

AGasCanister:
THIS IS WHAT HAPPENS WHEN YOU SPAM THE SANCTION BUTTON

goose4429:
I'm not sure you consider each of these a "weapon", so also consider changing:
- Stationary weaponry (mortars)
- Turrets
- CAS voucher (which gives CAS points)

Pariah919:
I got my feedback Ill reopen this when its ready

PR #10162, opened 2022-05-08
Engineer kits

Lumipharon:
## About The Pull Request
Adds some bundle sets to the engineer points vendor, and the functionality to do the same with other vendors if desired.

The idea with this is to push engineers a little bit more into specific directions, as well as add cost to certain things like teleporters that need it, without making them either rediculously expensive, or allowing you to buy multiple of them.

This is relatively barebones, so I imagine I or someone else can expand on this in the future easily.

**How they work**
You can only choose one kit (and they are optional), but they offer equipment that might not be available outside of the kits, or at a discount at the trade off for the loss of flexibility. This also means certain items that would be OP if you could buy multiple of them can be reasonably priced for their actual value (like teleporters).
image

Any spare points you have you can spent as normal.

**Available kits**

**Builder kit: 75 points**
Just materials, and lots of them. You completely lose out on gear customisation but you get something like 100 points worth of materials in exchange.
x300 metal
x100 plasteel

**Teleporter kit: 30 points**
Teleporters are no longer free. They're extremely powerful, and never really should have been free in the first place.
x1 pair of teleporters

**Minelayer kit: 45 points**
If you really like mines.
x1 mine layer
x2 large mine boxes

**Mortar kit: 30 points**
Mortar is no longer available roundstart (the howitzer is still there though), and can be bought. Comes in a special bag that can carry the mortar and a large number of shells.
x1 mortar bag
x1 mortar
x13 assorted shells
x1 range finder
x3 fire support encryption keys

**Plasmacutter kit: 20 points**
Plasma cutter is a kit now. Same cost but there's an opportunity cost it choosing it instead of another kit.
x1 plasma cutter
x1 superior welding module

**Flamethrower kit: 30 points**
Do you like fire? Do really, REALLY like fire? Maybe you'll like this kit. Maybe your fellow marines will not. Comes with a pyro bag with an internal fuel tank like a welder bag, and can mount a specific flamethrower that only be fueled via any type of back tank. Does NOT have a built in extinguisher.
x1 flamethrower bag
x1 engineer flamethrower
x2 WP nades
x3 C4

Sprites for the mortar and pyro bags
image

Also made mineboxes slightly cheaper as they're absurdly expensive for what they do.

## Why It's Good For The Game
A bit more control over gear for balance purposes, encourages engies to focus on a specific thing a little bit more.

## Changelog
:cl:
add: Engineering kits added to the engineering equipment vendor. Optional, but you can only choose one.
balance: Teleporters are no longer free round start, instead they are available via engineering kit.
balance: Mortar is no longer free round start, instead they are available via engineering kit.
balance: Make mine boxes slightly cheaper
/:cl:


novaepee:
It's finally here, MJP's engineer rework.

Arctrooper8:
Make the Plasma cutter kit give a welding helmet module since most PC users will be using mesons.

Lumipharon:
> Make the Plasma cutter kit give a welding helmet module since most PC users will be using mesons.

Will do

Kuro020 (approved):

KornFlaks:
Every man a mortar

Arctrooper8:
> > Every man a mortar
> 
> Yeah in theory you could have ALL THE MORTARS, but uh, then you need to get ammo for your 12 mortars. Speaking of which, mortar ammo is super cheap, but that's a separate issue.

The more mobile mortar idea is good since there is now the howie for FOB artillery, but without any mechanic/balance changes it's never going to be an effective/used weapon compared to the value of the other kits.

EvilMelonTheSecond:
Finally specs are back baby. Just pick engi for a free flamethrower bag or a mortar!

In all seriousness I hope people will actually use these for support instead of ya know... treating engi slots like a limited supply extra gun slot.

BraveMole (changes requested):
Mine sweeper kit looks expensive and weak.

I do not get the flamer kit, what is the point

Why not removing the deployable machine gun and give it in a kit ?

There should be a turrets kit

Lumipharon:
> Mine sweeper kit looks expensive and weak.
> 
> I do not get the flamer kit, what is the point
> 
> Why not removing the deployable machine gun and give it in a kit ?
> 
> There should be a turrets kit

Turrets are good enough as they are, I don't think any sort of kit to make them better/cheaper is a good idea.
flamer kit is really just so you carry more shit while still being a flamer man, but I'm fine with it not being super strong. Getting a free gun slot is strong enough as it is.
Mine layer is just points savings if you go mines, which people do. On top of the decrease to mine boxes. Happy to decrease them further if you think they're shit.

MG kit, I mean I can, but 9/10 times the people using MG's aren't engies, I wouldn't really want to take that away from basic ungas.

Pariah919 (commented):
I think most of the kits are fine, personally I would make an HMG thats more specialized for engineers but giving them a nerfed 102 might work.

Lumipharon:
> I think most of the kits are fine, personally I would make an HMG thats more specialized for engineers but giving them a nerfed 102 might work.

MG kit I feel would either be shit (I.e. not worth it over whatever the existing basic MG is now called), or OP and we'll see 8 engineers spamming it every round.

It's different from the mortar which requires coordination and req to be effective. MG is pretty stand alone until you need more ammo.

code/game/objects/items/storage/holsters.dm
```diff
@@ -126,6 +125,90 @@
...
+	w_class = WEIGHT_CLASS_HUGE
+	storage_slots = null
+	max_storage_space = 16
+	max_w_class = 4
+	access_delay = 0
+	holsterable_allowed = list(/obj/item/weapon/gun/flamer/big_flamer/engineer,)
+	bypass_w_limit = list(/obj/item/weapon/gun/flamer/big_flamer/engineer,)
+	storage_type_limits = list(/obj/item/weapon/gun/flamer/big_flamer/engineer = 1,)
+	///The internal fuel tank
+	var/obj/item/ammo_magazine/flamer_tank/internal/tank = null
```
TiviPlus, line 183:
```suggestion
	var/obj/item/ammo_magazine/flamer_tank/internal/tank
```

code/game/objects/items/storage/holsters.dm
```diff
@@ -115,7 +115,6 @@
 		/obj/item/ammo_magazine/rocket,
 		/obj/item/weapon/gun/launcher/rocket/recoillessrifle,
 	)
-	sprite_sheets = list("Combat Robot" = 'icons/mob/species/robot/backpack.dmi') //robots have their own snowflake back sprites
```
TiviPlus, line 118: ?

TiviPlus (changes requested):
I also dont know if flamer is a good idea it should be discussed

code/game/objects/items/storage/holsters.dm
Lumipharon: Ah balls, I left that out because species sprites were broken at the time I started this PR. I'll sort this out. Bloody robots.

Losenis:
Isn't the flamethrower a complete downgrade from normal flamethrower? No extinguisher, can only be fueled in one way when normal ones can already run with welderpack and backtank, I imagine the benefit of carrying inside the backpack is lost when engis can never have enough storage given they're the ones carrying all the mats and equipment.

Lumipharon:
> Isn't the flamethrower a complete downgrade from normal flamethrower? No extinguisher, can only be fueled in one way when normal ones can already run with welderpack and backtank, I imagine the benefit of carrying inside the backpack is lost when engis can never have enough storage given they're the ones carrying all the mats and equipment.

Yes, complete downgrade. The ability to have another gun slot is quite strong though, so I felt it needed to be balanced accordingly. It's still basically a satchel worth of space, and in my experience carrying engie shit is fine because of webbing + engie module + construction and tool pouches. But again, tradeoffs.

Starfirejordan:
Good PR

PR #10163, opened 2022-05-08
Include SSD mobs with grace timer in count_humans_and_xenos()

Wayland-Smithy:
## About The Pull Request

Title.

Possibly makes the explicit client checks in the conditional redundant.

### Pros:
- Momentary network lapses that boot all clients will no longer cause premature round ends.

### Cons:
- Could drag out normal round ends when the last members of a team decide to give up and go SSD without explicitly ghosting, requiring additional admin intervention to avoid the delay.

## Changelog
:cl:
code: SSD mobs with a grace timer are now included in count_humans_and_xenos(count_flags = COUNT_IGNORE_ALIVE_SSD) results.
/:cl:

### DIFF
--- a/code/datums/gamemodes/game_mode.dm
+++ b/code/datums/gamemodes/game_mode.dm
@@ -474,7 +474,7 @@ GLOBAL_LIST_INIT(bioscan_locations, list(
 			var/mob/living/carbon/human/H = i
 			if(!istype(H)) // Small fix?
 				continue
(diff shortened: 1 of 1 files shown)


Arctrooper8:
We could just TM for when we need it if the con is too great long term during the current DDoS for example.

BraveMole (approved):
Yeah good solution

PR #10164, opened 2022-05-09
Banish mass crashes to the shadowrealm

Lumipharon:
## About The Pull Request

Have you ever crashed when loads of bullets are flying around?
Have you ever wanted to fire guns at absurdly high rates of fire?

This PR fixes an extremely old bug that has plagued the server and it seems our predecessors, crashing everyone on screen.
Shout out to @2whatever2 for discovering the source of the bug: When a turf is hit by a bullet, a short bullet bounce animation is played. When a large number of these bounces are done on a single turf, the act of actually REMOVING the image from the turf just randomly mass crashes everyone on screen.

This is a byond bug so who knows why it actually happens, but it seems they way these were being handled was kinda shit code anyway, so I changed it to overlays which has fixed the problem.

This now opens the doors to guns with all the dakka, like so.
https://user-images.githubusercontent.com/7869430/167321668-d3ea4fe4-bbb7-4519-ac08-504f03c81b78.mp4

## Why It's Good For The Game
Mass crashes for everyone on screen because 2 people hit the same wall is bad.
BRRRT is good.

## Changelog
:cl:
fix: Fixed an old bug that would cause screen wide client crashes due to large amounts of shooting
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -1305,13 +1305,8 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		rotate.Turn(angle)
 		I.transform = rotate
 		add_overlay(I)
-		update_icon()
-		addtimer(CALLBACK(src, .proc/remove_bullet_ping, I), 3, TIMER_CLIENT_TIME)
+		addtimer(CALLBACK(src, .proc/cut_overlay, I), 3, TIMER_CLIENT_TIME)
 
-//removes the bullet ping without mass crashing everyone on screen
-/atom/proc/remove_bullet_ping(image/I)
-	cut_overlay(I)
-	update_icon()
 
 
 #define BULLET_MESSAGE_NO_SHOOTER 0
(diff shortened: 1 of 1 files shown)


Arctrooper8:
Super SG minigun is now possible
image

lKiseki:
madlads
they did it
they fucking fixed it

code/modules/projectiles/projectile.dm
```diff
@@ -1304,7 +1304,14 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
...
 		I.transform = rotate
-		flick_overlay_view(I, src, 3)
+		add_overlay(I)
+		update_icon()
+		addtimer(CALLBACK(src, .proc/remove_bullet_ping, I), 3, TIMER_CLIENT_TIME)
+
+//removes the bullet ping without mass crashing everyone on screen
+/atom/proc/remove_bullet_ping(image/I)
+	cut_overlay(I)
+	update_icon()
```
TiviPlus, line 1314:
```suggestion
		add_overlay(I)
		addtimer(CALLBACK(src, .proc/cut_overlay, I), 3, TIMER_CLIENT_TIME)

```

TiviPlus (changes requested):

code/modules/projectiles/projectile.dm
Lumipharon: Oh, update_icon isn't needed, honk.

BraveMole (commented):
timer abuse e_e

TiviPlus:
Flick overlay already used a timer so its an improvement perf wise

TiviPlus:
misclick

Issue #10165, opened 2022-05-09
Deployables cannot be placed on a huge percentage of tiles


Lumipharon:
### Description

Caused by #10152

The above PR restricted deployables from being deployed on a turf with ANY structures or machinery, dense or not.
This uh, includes this such as: 
- Power cables
- disposal or air pipes
- buttons
- extinguisher cabinets
- light fixtures
- patches of grass
etc etc

This probably needs to be reverted, or a large white list/black list of things need to be setup for this to work properly.

### Test Merges

Not relevant

### Reproduction Steps

Place deployable on a tile with a power cable under the floor
Cry

### Screenshots

![image](https://user-images.githubusercontent.com/7869430/167323338-62422974-884a-4eae-809d-89b56f0c3e64.png)

FosterGrime-Heymynameslouie:
speed revent pr, cant place the literally all engie deployables in 90% of tiles this sucks

Lumipharon:
Closed by #10166

PR #10166, opened 2022-05-09
Revert "Deployables cannot be deployed on a tile with any structures or machinery, as opposed to only dense ones"

Lumipharon:
Reverts tgstation/TerraGov-Marine-Corps#10152

Causes a whole bunch of issues. #10165

Makes a significant percentage of turfs invalid for deployables, due to things like patches of grass, pipes, wires, light fixtures etc.

This needs to be done differently to work.


Derrick2Electricboogaloo:
Xenos rejoice for the short time of engineers having fucked placement equipment.

Issue #10063, opened 2022-04-25
"Join As" button for call burrowed kicks you out of larva queue


Shirbu:
### Situation: 
Was in larva candidate queue in 2nd position. Saw button to Join As a burrowed larva. Clicked button.

### What I expected to happen:
I expected to either join as a larva, or more likely, to not make it in, but keep my position in the larva candidate queue

### What happened instead:
I got kicked to larva candidate queue position 5, without having spawned as a larva.

### Test Merges

Round ID: [14362](https://scrubby.melonmesa.com/round/14362)
BYOND Version: 514.1583
Server revision compiled on: 
Master commit: [ce9618a3933b1aba0563f1b9d96f72b22f60aa09](https://github.com/tgstation/TerraGov-Marine-Corps//commit/ce9618a3933b1aba0563f1b9d96f72b22f60aa09)
The following pull requests are currently test merged:
[#10031: 'Refactors gun icon/item code' by Lumipharon at commit 622e614bac](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10031)
[#10008: 'Boxing+' by breadhunt at commit beb71c8b35](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10008)
[#9977: 'Health kits chain apply' by QualityVan at commit 5afedc795b](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9977)
[#9772: 'TGUI updates' by TiviPlus at commit 16b87e0607](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9772)

### Reproduction Steps

1. Join larba queue
2. See shiny button that says "Join As" a burrowed larva
3. Click it
4. Get kicked out of larva queue

### Screenshots

![image](https://user-images.githubusercontent.com/23252965/165042611-b6aae729-60c9-42a6-9229-3bee9a65eb2c.png)

Shirbu:
![image](https://user-images.githubusercontent.com/23252965/165044353-4e328446-bf2f-4c47-8913-c4a045dee2a7.png)

PR #10173, opened 2022-05-09
Fix call of the burrowed unintentional larva queue leaves

Wayland-Smithy:
## About The Pull Request

Since `attempt_to_join_as_larva()` acts like a toggle now if you are already in the queue when you click the link in chat to Join as a Larva the result will be you get booted as if you clicked the regular button.

This adds a check to prevent that from happening.

Fixes: #10063

## Why It's Good For The Game

Fix good.

## Changelog
:cl:
fix: Fixed call of the burrowed unintentional larva queue leaves.
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -202,7 +202,7 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 
 		var/datum/hive_status/normal/HS = GLOB.hive_datums[XENO_HIVE_NORMAL]
 		if(LAZYFIND(HS.candidate, ghost))
-			to_chat(usr, span_warning("You are already in the queue to become a Xenomorph."))
+			to_chat(ghost, span_warning("You are already in the queue to become a Xenomorph."))
 			return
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/dead/observer/observer.dm
```diff
@@ -200,6 +200,11 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 			return
 		var/mob/dead/observer/ghost = usr
 
+		var/datum/hive_status/normal/HS = GLOB.hive_datums[XENO_HIVE_NORMAL]
+		if(LAZYFIND(HS.candidate, ghost))
+			to_chat(usr, span_warning("You are already in the queue to become a Xenomorph."))
+			return
```
TiviPlus, line 206:
```suggestion
			to_chat(ghost, span_warning("You are already in the queue to become a Xenomorph."))
			return
```

TiviPlus (approved):
rest lgtm

BraveMole (changes requested):
Does this verb needs to exist ?

Wayland-Smithy:
Idk, and I don't think that is a good reason to block the PR. It can always be deleted wholesale later.

TiviPlus:
Fix now remove later

PR #10175, opened 2022-05-10
Remove sentry kit near FOB drone and additional map stuff

novaepee:
## About The Pull Request

All sentries shall spawn inside TerraGovTech Engineer System Vendor; to do that, I remove the sentry kit near FOB drone and just add an additional sentry kit in the TerraGovTech Engineer System Vendor.

image

Made some PoS' prep QoL by adding green facepaint near vendors since people would want to use it for coloring.
image

Remove 5 stacks of metal and 3 stacks of  plasteel near Theseus' FOB drone.
image

Remove extra crowbars and screwdrivers in all prep if needed since we don't need it to dissemable Jaegar.
image

## Why It's Good For The Game

The sentry kit next to FOB drone is a relic of an old time, so it makes senes to consolidate all the sentry kits into one place.

PoS' prep got an additional change since I saw Minerva has green facepaint for marines to use and PoS has none, so I added that. Also, deleted the vendors that lurk outside of prep to fit the ship's aesthetics.

The Theseus' metal and plasteel surplus is a surprise. Since no one delete the metal and plasteel when contributor let metal and plasteel spawn inside FOB drone, this makes Theseus spawn with more supplies. Most interesting.

## Changelog
:cl:
add: additional sentry kit in the TerraGovTech Engineer System Vendor; now engineers can acquire all four sentry kits in the vendor
del: sentry kit near FOB drone
add: green facepaint in PoS' prep
del: Theseus' metal and plasteel surplus near FOB drone
del: extra crowbars and screwdrivers in all prep if needed since we don't need it to dissemable Jaegar
/:cl:

### DIFF
--- a/_maps/map_files/Minerva/TGS_Minerva.dmm
+++ b/_maps/map_files/Minerva/TGS_Minerva.dmm
@@ -10308,13 +10308,6 @@
 	dir = 4
 	},
 /area/mainship/medical/lower_medical)
-"He" = (
-/obj/structure/rack,
-/obj/item/storage/box/sentry,
-/turf/open/floor/mainship/red{
-	dir = 8
-	},
-/area/mainship/living/briefing)
 "Hf" = (
 /obj/machinery/atmospherics/pipe/simple/supply/hidden/layer1{
 	dir = 4
@@ -24328,7 +24321,7 @@ Dn
 Cg
 Np
 Gn
-He
+Gn
 HU
 tt
 Jx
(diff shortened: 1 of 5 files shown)


Losenis:
You can get facepaint from the armor vendors though.

novaepee:
> You can get facepaint from the armor vendors though.

Yes, but Minerva has them already laid out on the table, which is actually pretty convenient. More convenient time.

BraveMole (approved):

PR #10183, opened 2022-05-11
Maturity tower provides 25% less bonus than before.

Pariah919:
## About The Pull Request
25% less bonus, .2 to .15.

## Why It's Good For The Game
Maturity tower currently does far too much in its current state. Considering you can basically just place it in a corner of the map and never lose it.

## Changelog
:cl:
balance: Maturity tower nerfed by 25%.
/:cl:

### DIFF
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -1142,7 +1142,7 @@ TUNNEL
 	///hivenumber of this tower
 	var/hivenumber
 	///boost amt to be added per tower per cycle
-	var/boost_amount = 0.2
+	var/boost_amount = 0.125
 
 /obj/structure/xeno/maturitytower/Initialize(mapload, hivenum)
 	. = ..()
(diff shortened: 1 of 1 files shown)


Al-1ce:
> Maturity tower currently does far too much in its current state.


I'm fully convinced you've never actually seen a maturity tower in action before. Even several do little more than make a **mild at best** impact on the hives overall maturity, which already feels slow enough as it is, especially for latejoined xenomorphs who have to deal with an awkward "afk" phase if they'd rather not die at mature. 


>Considering you can basically just place it in a corner of the map and never lose it.


This is ..intended. It's not meant to be a high-value target like a silo, because it isn't. And yet, despite taking a fairly large amount of points to make, a single maturity tower will barely make a dent in the overall gameplay. It acts as little more than QOL to ease the waiting by a minute or two.

Salt PR. Simple as.

Walarks:
I think maturity tower should do much more and be restricted to more vulnerable spots.

private-tristan:
sitting in a corner until you hit mature/elder is good gameplay :)

BraveMole:
Ok for changes to maturity tower, but a simple number balance won't cut it here.

PR #10187, opened 2022-05-11
Harvester fixes and qol

Walarks:
## About The Pull Request

Fixes harvester storing 60 instead of 30 units of a reagent, along with some text prompts.
Updates text prompts from chat messages to runechat.
## Why It's Good For The Game

More clarity and fixes unintentional buff.
## Changelog
:cl:
balance: fixes harvester storing 60 instead of 30 of each reagent
qol: updates harvester messages to runechat instead of chat
fix: select reagent button icon no longer disappears sometimes when re-equipping the weapon
/:cl:

### DIFF
--- a/code/datums/components/harvester.dm
+++ b/code/datums/components/harvester.dm
@@ -1,9 +1,9 @@
-#define MAX_LOADABLE_REAGENT_AMOUNT 60
+#define MAX_LOADABLE_REAGENT_AMOUNT 30
 #define NO_REAGENT_COLOR "#FFFFFF"
 
 /datum/component/harvester
 	///reagent selected for actions
-	var/selected_reagent
+	var/datum/reagent/selected_reagent
 	///Loaded reagent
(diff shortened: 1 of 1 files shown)


code/datums/components/harvester.dm
```diff
@@ -69,32 +69,32 @@
 		return
 
 	if(!isreagentcontainer(cont) || istype(cont, /obj/item/reagent_containers/pill))
-		to_chat(user, span_rose("[cont] isn't compatible with [source]."))
+		user.balloon_alert(user, "[cont]: not compatible with [source].")
```
TiviPlus, line 72: ballon alerts should use inserts extremely sparingly

The message in balloon alerts should always be as short as possible and not make reference to who or what since the viewer already sees what is being worked with

TiviPlus (changes requested):
Other than that code ok

TiviPlus (approved):

PR #10191, opened 2022-05-11
Prevents kitting undamaged limbs

QualityVan:
## About The Pull Request
Since wounds don't exist any more we can be more discerning about whether a limb needs kit treatment, beyond just whether it's bandaged or not. Don't bother bandaging if there's no brute damage, don't bother salving if there's no burns, and don't bother disinfecting if there's no chance to gain germs from current damage.
Also fixes movement germs being applied in a couple incorrect cases.

## Why It's Good For The Game
Helps avoid confusion and slowdown from kitting healthy limbs, both from direct apply and from cycling.

## Changelog
:cl:
fix: You can't kit undamaged limbs
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_organs.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_organs.dm
@@ -43,7 +43,7 @@
 				I.take_damage(rand(3,5))
 
 			//Moving makes open wounds get infected much faster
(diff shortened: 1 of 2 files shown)


FosterGrime-Heymynameslouie:
based pr speedmehrge

BraveMole (approved):

PR #10193, opened 2022-05-12
Psychic link is manually cancellable, Buffs range

Walarks:
## About The Pull Request

Psychic link process is now able to be cancelled manually by activating the skill again.
Also increases psychic link range.
## Why It's Good For The Game

Psychic link seems underused in most cases, so I'm interested in buffing its range to make it easier to utilize. I think this would still keep it somewhat proactive by requiring some setup as opposed to making it more reactive by reducing the channeling time.
## Changelog
:cl:
qol: psychic link's activation can be manually cancelled by activating the skill again
balance: psychic link has 15 instead of 10 range
/:cl:


Pariah919:
Cancellable seems okay, range seems like a creep.

Walarks:
> 

The thing is that it almost isn't used currently. Also when you are linked as gorger, you don't bring anything else to a fight.

goose4429:
It is 100% used in RUTGMC, and has good performance. Cancellation is nice, buff range is too much. Although +1 or +2 tiles seems like what has to be done.

Walarks:
> It is 100% used in RUTGMC, and has good performance. Cancellation is nice, buff range is too much. Although +1 or +2 tiles seems like what has to be done.

I made this change based on my experience on TGMC instead of RUTGMC. I don't think RUTGMC should be a concern since it's a different server.

code/datums/status_effects/xeno_buffs.dm
```diff
@@ -101,7 +101,7 @@
 	///Minimum health threshold before the effect is deactivated
 	var/minimum_health
 	///If the target xeno was within range
-	var/was_within_range = TRUE
+	var/was_within_range
```
TiviPlus, line 104: is this a bool or not cus itf its a bool make it false explicitly

TiviPlus (approved):
rest lgtm

code/datums/status_effects/xeno_buffs.dm
Walarks: updated

PR #10195, opened 2022-05-12
Removes under barrel stuff from SH-15

Arctrooper8:
## About The Pull Request

Removes the plasma pistol from the SH-15

## Why It's Good For The Game

The SH-15 has a unique niche where it has access to under barrel attachments, it makes up for this with it's specialized ammo types which comes in mags and do less damage (alongside not having buckshot at all).
However the plasma pistol is arguably still too much even for a the more reserved magazine fed cousin of the other shotguns.
What with it's on hit bonus damage and fire stacks, the plasma pistol combined with any low fire rate high damage per shot weapon is a strong and generally unfun to play against combination. (One is left to wonder why it even became an attachment in the first place)

TLDR: SH-15 having under barrels good niche, plasma pistol bad and dumb and stupid and too strong why is it even an attachment.

## Changelog
:cl:
balance: Removed plasma pistol as an attachment option for the SG-15
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -941,7 +941,7 @@
 		/obj/item/attachable/stock/tx15,
 		/obj/item/attachable/compensator,
 		/obj/item/attachable/extended_barrel,
-		/obj/item/attachable/heavy_barrel,
+		/obj/item/attachable/heavy_barrel,efnnleanlfae
 		/obj/item/attachable/motiondetector,
 	)
 
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -943,6 +943,9 @@
 		/obj/item/attachable/extended_barrel,
 		/obj/item/attachable/heavy_barrel,
 		/obj/item/attachable/motiondetector,
+		/obj/item/weapon/gun/flamer/mini_flamer,
+		/obj/item/weapon/gun/grenade_launcher/underslung,
+		/obj/item/weapon/gun/rifle/pepperball/pepperball_mini,
 	)
 
 	flags_gun_features = GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_WIELDED_FIRING_ONLY //Its a shotgun type weapon effectively, most shotgun type weapons shouldn't be able to point blank 1 handed.
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -941,7 +941,7 @@
 		/obj/item/attachable/stock/tx15,
 		/obj/item/attachable/compensator,
 		/obj/item/attachable/extended_barrel,
-		/obj/item/attachable/heavy_barrel,efnnleanlfae
+		/obj/item/attachable/heavy_barrel,
 		/obj/item/attachable/motiondetector,
 	)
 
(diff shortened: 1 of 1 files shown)


Naaanii:
Pretty sure TX-15 could always have underbarrel stuff before more powercreep was introduced. It's also the only shotgun that takes magazines. I think having underbarrel is a good niche for this gun.

Lumipharon:
Only the plasma pistol is an issue, if at all. The rest are entirely non issues.

It was a deliberate choice to keep underbarrels on it.

It uses different (weaker) ammo that normal shotguns, the slugs have notably worse CC. It's coded as a rifle as that's functionally what it is.

Arctrooper8:
> Only the plasma pistol is an issue, if at all. The rest are entirely non issues.
> 
> It was a deliberate choice to keep underbarrels on it.
> 
> It uses different (weaker) ammo that normal shotguns, the slugs have notably worse CC. It's coded as a rifle as that's functionally what it is.

I think both fire attachments are an issue, as while the SG-15 slug may not be as strong as the normal SH slug it still causes knockback which stuns you and makes you much easier to hit with the plasma pistol and flamer.
And the UGL niche is already taken by the italian shotgun (which is seasonal to be fair)
And let's be honest, nobody uses the mini pepperball gun.

Naaanii:
> > Only the plasma pistol is an issue, if at all. The rest are entirely non issues.
> > It was a deliberate choice to keep underbarrels on it.
> > It uses different (weaker) ammo that normal shotguns, the slugs have notably worse CC. It's coded as a rifle as that's functionally what it is.
> 
> I think both fire attachments are an issue, as while the SG-15 slug may not be as strong as the normal SH slug it still causes knockback which stuns you and makes you much easier to hit with the plasma pistol and flamer. And the UGL niche is already taken by the italian shotgun (which is seasonal to be fair) And let's be honest, nobody uses the mini pepperball gun.

It knocks back, but doesn't stun, slowdown is a given as the weapon relies on landing consistent shots. Once again this weapon is more like a rifle rather than an actual shotgun so having underbarrel attachments on it is fine. If anything, deleting plasma pistol would be the only solution if it was an actual issue as the auto shottie has had ubf for a long time and while strong, it has never been an actual issue.

lovegreenstuff:
hey what's up guys it's ya boi

uhhh the original shotgun underbarrel pr that i made for pette intentionally left out tx15 because it was mostly a matter of burst damage, aka buck, being combo'd into shit like another buck or an impact nade and basically oneshotting people

tx15 only does like 85 damage max per shot compared to like 170 for t39 buck, so we just thought it wasnt terribly necessary

since then, impact nades got sent to the shadow realm, flamer damage on hit was set to zero, and masterkey firerate got nerfed, so it's probably less necessary to do so now.

but hey if you want to fuckin uhhhhh keep goin' with this pr, go for it mang

nutsupreme:
salt pr

Pariah919 (changes requested):
50/50 on Plasma pistol, I'm fine with it being gone, but I feel T18 carbine is pertty equivalent anyway.
Hard no on the rest of them.

Arctrooper8:
The people have spoken. Considering that the SH-15 has different ammo which does reduced damage, and the unique niche having under barrels gives this gun, it is obviously better to let them keep the under barrel attachments.
Except the plasma pistol, because it sucks.
That repulsive duct taped DIY puke green fire pistol, it shouldn't have ever been an attachment.

BraveMole:
SH 15 needs it niche, and having attachments is fine

PR #10197, opened 2022-05-12
Lore Preferences Update

Arctrooper8:
## About The Pull Request

Removes the citizenship type.
Removes some unused lore variables
Removes Paganism (Wicca)
Replaces Mormon with Anglican
Adds a new religion choice

## Why It's Good For The Game

The citizenship choice is confusing when compared to what little lore that we have. Unlike our progenitor CM where the setting still has independent earth nations, TGMC's setting has only the authoritarian solar system encompassing Terra Government aka TG(I'm sure whoever wrote it felt very clever). This means that the foreign choice is completely absurd, what with the only foreign entities being the ICC and SoM who are hiding out in other systems and actively hostile to the Terran Government. Green card colony doesn't make much sense either since the colonies are under the direct jurisdiction of the Terran Government (albeit it a lot of colonies are owned by megacorps such as NT, they still fall under TG sovereignty).

Another factor is the Vatborn and combat robots. These two groups are mostly created on terra, either in birthing facilities or factories.
Overall citizenship just doesn't fit the setting. 

The unused lore variables were unused, you can't set them in character creation and they suffer from a lot of the same problems as citizenship even if you could set them.

We currently have Paganism(Wicca) and Wiccanism, these are the same thing and having both is silly.

Unlike CM where they are United States marines, TGMC has marines coming from all over Earth and the solar system. Therefore it is completely absurd to have a minority Christian denomination that is almost entirely based in the US state of Utah be its own choice instead of lumped in with Christianity other (truly a Mormon moment). A good replacement for this is Christianity (Anglican) due to the fact there are many many more Anglicans, and they're more prestigious.

Finally, the new religion is a fun (and copyright free, I made sure to word it properly) new choice.

## Changelog
:cl:
add: Added a lot more Christian Denominations
add: Adherents of the Machine God added 01010000 01110010 01100001 01101001 01110011 01100101 00100000 01001111 01101101 01101110 01101001 01110011 01110011 01101001 01100001 01101000
add: Paganism
del: removed Wiccanism and Pagan(Wicca)
del: Removed unused lore preference variables alongside broken Corporates relations
fix: Synthetic type button now actually works instead of making you Western
/:cl:


TiviPlus (approved):
Code ok whats up with your changelog

Al-1ce:
As excerpt, from the wiki: 


>Territory: Sol and surrounding systems
>Terragov controls almost everything in and around the solar system



It should be emphasized here that, while TerraGov retains control of the entire Solar System, this is restricted to exactly that - _**our**_ solar system. Meaning our planets, like Earth, Venus, Mars, etcetera. It does not say that TerraGov is an all-encompassing entity of the entire universe, like you've implied here. Equally-so, life is not restricted within Sol. That's just silly.

It's clear that other solar-systems which are not controlled by TerraGov are  certainly inhabited by independents, such as explorers, corporations, and the like. These are, as expected, out of the bounds of TerraGov's borders and are thus not really expected to behave like a vassal or something of the like.


>Green card colony doesn't make much sense either since the colonies are under the direct jurisdiction of the Terran Government


This would only really make sense if TerraGov had a wide-encompassing grip on the entire universe, which it does not. It controls Sol and a few surrounding systems - leaving much, much more of the inhabitable galaxy open to terraforming and independent settling. 

Citizenship does fit the setting if you look at it in a more contextual and broader context, rather than jumping the gun and assuming that TerraGov has absolute control of every single planet, colony, and system, which is.. just silly; especially considering the all-out war against the xenomorph threat. There are much, much bigger fish to fry.



>Wicca / Paganism


I'll give a simple run-down on the differences between these two, although I understand and agree with your point, somewhat.
1: Pagans tend to believe in one, or several gods, whereas Wiccans tend to believe in only two. (These vary WILDLY)
2: Pagans do not preform witchcraft, whereas Wiccans do.
3: Paganism does not have a founder, whereas Wiccanism was invented.

It'd be much simpler to erase these two and replace it with a general "spiritism", or just "Pagan".
 

>Mormonism


Are you seriously implying that Mormons wouldn't acquire FTL _just_ to ring your doorbell?

Arctrooper8:
Good feedback, I shall implement it

Arctrooper8:
Updated the CL, too lazy to update desc

Lumipharon:
There is actually some useful fluff in the codex and the like if you look for it.
image

image

PR #10200, opened 2022-05-13
Swapping attachments keeps the old attachment in-hand

breadhunt:
## About The Pull Request

Per title. Previously, the old attachment would be dropped on the floor.

## Why It's Good For The Game

Mainly just QOL; having to pick up some very small icons from under you each time you swap an attachment can get really tedious when trying out builds or whatever. Also sucks for flamers. I've wanted this for quite a while, and it's a oneline change, so

## Changelog
:cl:
qol: When swapping an attachment, the old attachment now goes into your hand instead of on the floor
/:cl:

### DIFF
--- a/code/datums/components/attachment_handler.dm
+++ b/code/datums/components/attachment_handler.dm
@@ -74,20 +74,26 @@
 		QDEL_NULL(attachment)
 		return
 
+	var/obj/item/old_attachment = slots[slot]
+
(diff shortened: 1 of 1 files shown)


Losenis:
Finally, good change. Would make carrying attachments like swapping from RDS to scope on the run more viable.

TiviPlus (approved):

PR #10208, opened 2022-05-14
Disable phoron miners on hijack

Wayland-Smithy:
## About The Pull Request

Permanently disables phoron miners on global hijack signal.

## Why It's Good For The Game

Fix good. Seems like a bug when there is no possible way for xenos to prevent tad from fixing miners even if they were to go around and disable them all before hijacking (that is if the dropship locking window even allows).

## Changelog
:cl:
fix: Fixed phoron miners continuing to function after hijack.
/:cl:


Naaanii:
Would be nice to also disable tad being allowed to take off from the ship post hijack.

TiviPlus (approved):

PR #10209, opened 2022-05-14
SG minigun fixes

Arctrooper8:
## About The Pull Request

Addresses some of the issues of the SG minigun:
It now uses the normal minigun sound (it doesn't cause tinnitus anymore)
And fixed a misdesignation error in the SG vendor.

## Why It's Good For The Game

People say the sound is a bit much, I have to agree.
Names matching their gun is good.

## Changelog
:cl:
fix: SG vendor now displays the proper name for the SG minigun
soundadd: SG minigun now uses default minigun sound, ears of the world rejoice!
sounddel: Removed minigun_sg since it is now unused and hurts everyone's ears
/:cl:


TiviPlus:
Why didnt you just turn it down

Arctrooper8:
Because the minigun sound is better, even MJP said so

PR #10210, opened 2022-05-14
SG Minigun balance

Arctrooper8:
## About The Pull Request

1. SG minigun firerate from .1 -> .15
2. SG minigun windup delay from .75 -> .4
3. SG minigun damage multiplier from .8 -> 1
4. SG minigun damage per bullet from 10 -> 15
5. Removed smartgun type bullet and the tank weapon that used it (Both were unused)

## Why It's Good For The Game

1. Right now the minigun is a bit too fast
2. Windup delay was a copy paste from MJP's original minigun I forget to change (shouldn't affect too much)
3. The SG minigun has its own bullet type and doesn't need a damage multiplier (Another copy paste forget)
4. Less firerate means more damager per bullet to maintain a DPS similar to the SG-29
5. Unused, removed to reduce confusion between actual SG weapon ammo

## Changelog
:cl:
balance: SG minigun fires slower and does more damage
fix: SG minigun is now consumer safe and no longer bullies low end devices
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1105,19 +1105,8 @@ datum/ammo/bullet/revolver/tp44
 //================================================
 */
 
-/datum/ammo/bullet/smartgun
-	name = "smartgun bullet"
-	icon_state = "redbullet" //Red bullets to indicate friendly fire restriction
-	hud_state = "smartgun"
-	hud_state_empty = "smartgun_empty"
-	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
-	accurate_range = 15
-	damage = 20
-	penetration = 20
-	sundering = 1
-
 /datum/ammo/bullet/smartmachinegun
-	name = "smartgun bullet"
+	name = "smartmachinegun bullet"
 	icon_state = "redbullet" //Red bullets to indicate friendly fire restriction
 	hud_state = "smartgun"
 	hud_state_empty = "smartgun_empty"
@@ -1128,7 +1117,7 @@ datum/ammo/bullet/revolver/tp44
 	sundering = 2
 
 /datum/ammo/bullet/smartminigun
-	name = "smartgun bullet"
+	name = "smartminigun bullet"
 	icon_state = "redbullet" //Red bullets to indicate friendly fire restriction
 	hud_state = "smartgun"
 	hud_state_empty = "smartgun_empty"
@@ -1136,7 +1125,7 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range = 15
 	damage = 10
 	penetration = 15
-	sundering = 2
+	sundering = 1
 
 /datum/ammo/bullet/smartgun/lethal
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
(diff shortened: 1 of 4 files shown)


Losenis:
It may be beyond the scope of this pr but is it possible to have ammo boxes made to refill the powerpack with that could be put in the G8 belt storage? Absolutely loathing powerpacks when you're stuck to one single slot for ammo and don't even have half the ammo a T60 dunga could have.

Arctrooper8:
> It may be beyond the scope of this pr but is it possible to have ammo boxes made to refill the powerpack with that could be put in the G8 belt storage? Absolutely loathing powerpacks when you're stuck to one single slot for ammo and don't even have half the ammo a T60 dunga could have.

Brilliant idea, that'll be my next PR

KornFlaks:
I am quite concerned about this IFF Minigun buff.

Reducing windup and increasing damage by 87.5% (damage mult increased by 1.25x times damage value increased by 1.5x = 1.875x) while only reducing fire rate by 50%? 

This needs some testing on time to kill before merging. Please.

Arctrooper8:
> I am quite concerned about this IFF Minigun buff.
> 
> Reducing windup and increasing damage by 87.5% (damage mult increased by 1.25x times damage value increased by 1.5x = 1.875x) while only reducing fire rate by 50%? I would expect only the damage value to be changed to remain the same as the current minigun.
> 
> Nor does this touch it's sunder value which is the main concern of this gun. Now we increase the damage as well? Please, we're not blind.
> 
> This needs some testing on time to kill before merging. Please.
> 
> I know it's not gonna happen and instead speedmerged but still, I would like to think that this might get tested before we see it permanently on the battlefield. Unlike the... IFF minigun itself.


I just used the values that MJP suggested. And yeah, the SG minigun should've been TM'd but oh well, that's why I'm making these Knee-jerk reaction PRs.
I'll do a bit more mathematics and testing though, thanks for the feedback.

Lumipharon:
> I am quite concerned about this IFF Minigun buff.
> 
> Reducing windup and increasing damage by 87.5% (damage mult increased by 1.25x times damage value increased by 1.5x = 1.875x) while only reducing fire rate by 50%? I would expect only the damage value to be changed to remain the same as the current minigun.
> 
> Nor does this touch it's sunder value which is the main concern of this gun. Now we increase the damage as well? Please, we're not blind.
> 
> This needs some testing on time to kill before merging. Please.
> 
> I know it's not gonna happen and instead speedmerged but still, I would like to think that this might get tested before we see it permanently on the battlefield. Unlike the... IFF minigun itself.

Nerfing firing rate from 0.1 to 0.15 cuts the sunder/second by a third. Still fairly high though.

goose4429:
You should decrease sunder further, it's still pretty high at 13/sec. For each second under fire, beno needs to rest on white weeds under rec pheros for about 15 secs. Not very cool.

turnop:
> You should decrease sunder further, it's still pretty high at 13/sec. For each second under fire, beno needs to rest on white weeds under rec pheros for about 15 secs. Not very cool.

Or they just need to use armor fruit that literally nobody ever uses, or fight at less than 100% armor (after a full second of concentrated minigun fire mind)? This looks good to me, the 85 definitely needs a buff though I am sad that it will brrt less

Losenis:
[Laughs in Gorger]

Arctrooper8:
> Or they just need to use armor fruit that literally nobody ever uses, or fight at less than 100% armor (after a full second of concentrated minigun fire mind)? This looks good to me, the 85 definitely needs a buff though I am sad that it will brrt less

Sadly too much Brrrrt kills Russian computers, perhaps one day when we evolve from the limitations of Byond we shall reach enuff Dakka

Arctrooper8:
A lot of people don't seem to be considering the fact that most Xenos dodge, hide behind cover, or the fact that SG weapons have always had the niche of high sunder. You're not gonna hit every single shot, and you're certainly not gonna hit any shots if the Xenos get close to you.

PR #10212, opened 2022-05-14
Removes Xeno Ability to Instantly Extinguish On Fire Humans With Help Intent

KornFlaks:
## About The Pull Request

Removes ancient ability of xenos that allowed them to instantly extinguish human marines with a single click of the help intent.

## Why It's Good For The Game

Xeno extinguishing marines is one of the last remaining relics of the capture days. Back when marines would instead flame themselves and thus kill themselves to avoid capture.

Now, it's just a bald (e.g., me) xeno misclick if accidently on help intent and giving marines a free instant fire extinguish.

Xenos should not be helping marines by reducing damage on them. If a marine is on fire, let them burn.

This will not effect xenos helping other xenos to extinguish their flames. Only between xenos and animals.

## Changelog
:cl:
balance: Xenos on help intent no longer extinguish their sworn enemies and instead watches them burn.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
+++ b/code/modules/mob/living/carbon/xenomorph/attack_alien.dm
@@ -208,6 +208,9 @@
 
 	switch(X.a_intent)
 		if(INTENT_HELP)
+			if(on_fire)
+				X.visible_message(span_danger("[X] stares at [src]."), span_notice("We stare at the roasting [src], toasty."), null, 5)
+				return FALSE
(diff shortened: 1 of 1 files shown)


DeltaFire15:
This still has a purpose from its original capture days though considering marines like setting themselves on fire to get huggers off, besides other situations where an instant extinguish may be useful. I don't think this needs removing just because they got clicked on help instead of harm accidentally.

Naaanii:
Honestly, skill issue. Just use the right intent.

BraveMole (approved):

PR #10213, opened 2022-05-14
Fire armor pass

DeltaFire15:
## About The Pull Request
So basically, the entire structure of xeno fire resistance was very cursed, going of a base resistance value that never got used anywhere, not accounting for fire soft armor, not accounting for all cases, and so on.
This aims to make fire armor more consistant in that fire damage sources (primary concerns: DoT, initial ignition conditions, fire stacks, initial / impact damage) respect fire soft armor. Did you know that it didn't matter at all for crossing fire or being on fire before? Yeah. Cool ain't it? The only thing that mattered was jello, or hard armor being very high which only surt has ever used (although it does get to keep it)
This PR removes some completely unused or deprecated variables related to fire from the code aswell as making cases also care about soft armor. As usual with soft armor, it gets handled as percentile reduction, with 100 being equivalent to full immunity. For xenomorphs, the fire resistance additionally gets modified by jello which effectively causes full immunity to fire-based effects while active.
Damage now gets handled as is usual with soft armor so -soft_armor%
Fire stack increases now are affected simillarely, although they are CEIL'd (this means if you would get 0.3 stacks you still get 1 stack, but if you get 0 it's 0). The use of ceiling here is meant to not cause soft armor to reduce fire stack increases disproportionately as they do not support floats well and the alternative would be to just round it.
Ignition attempts are changed in that a fire resistance of 100% will negate any attempts of igniting a target.

Additionally, due to fire stacks potentially lingering for a long time even without ignition, xenomorphs now lose one stack per life tick when *not* on fire. Think of this akin to them getting rid of the flammable material on them doing whatever they are doing - this change is not required for the functional stuff and is merely meant to reduce inconsistancies regarding fire stack persistance, although nonetheless a balance change. If wanted, I could change this to be exclusive to resting instead, which would be fine by me. This affects stuff like, say, splashing xenos with alcohol or fire weaponry failing to ignite targets but increasing their stacks anyways, leading to a stronger burn effect in the future.

Please testmerge this PR before merging as there is never a 100% change for all cases to be covered. Should be fine but nonetheless better to do. Need to do some testing myself to make sure this works as it is intended.
Disclaimer, as this makes armor actually matter against burns, this may end up weakening the effects of igniting weapons such as flamers, plasma pistols, CAS lasers and everything else that applies DoT fire. However, maybe this can allow a buff for flamers at a later time without all castes having it have the same effects on them due to soft armor not mattering.

I also found a random one line hivelord fix for jello throwing sometimes not working as intended so I threw it in here because I'm lazy. Inbefore "atomize it".

## Why It's Good For The Game
Fire damage as of now is in a weird state where sometimes, armor matters, sometimes, only this weird modifier only affected by jello does, sometimes it only matters for humans (lava) and in other cases yet again, it just doesn't care about anything (DoT from being on fire, fire stacks, simillar)
This aims to bring fire to a more consistant state which should allow for better future updates to be made on the base that fire armor actually matters and say, castes without high armor may suffer more from fire than those with good fire armor, as opposed to all suffering the same as it is right now.

## Changelog
:cl:
code: Refactored a bunch of fire-related things (DoT, fire stack additions, some weapon special effects, Ignition) to care about percentile fire armor (aka soft armor) instead of it not mattering in many cases.
balance: Xenomorphs, when NOT on fire lose one fire stack per life tick, counteracting the prospect of non-igniting events adding firestacks having too much of an impact on subsequent ignitions, provided sufficient time inbetween.
fix: Throwing resin jello as hivelord should be less likely to not have an impact now, due to a previous possible race condition.
balance: Flamethrower fire intensity has been increased by 30% to counteract armor being taken into account. Other weaponry making use of incendiary effects unaffected.
/:cl:


code/modules/mob/living/living_defense.dm
```diff
@@ -194,6 +194,8 @@
 /mob/living/proc/adjust_fire_stacks(add_fire_stacks) //Adjusting the amount of fire_stacks we have on person
 	if(status_flags & GODMODE) //Invulnerable mobs don't get fire stacks
 		return
+	if(add_fire_stacks)	//Fire stack increases are affected by armor, end result rounded up.
+		add_fire_stacks = CEILING(fire_stacks * get_fire_resist(), 1)
```
QualityVan, line 198: did you want this to be ceiling(add_fire_stacks ?

code/modules/mob/living/carbon/xenomorph/xenomorph_defense.dm
```diff
@@ -31,5 +31,5 @@ Contains most of the procs that are called when a xeno is attacked by something
 	return ..()
 
 ///Calculates fire resistance given caste and coatings, acts as a multiplier to damage taken
-/mob/living/carbon/xenomorph/proc/get_fire_resist()
-	return clamp(xeno_caste.fire_resist + fire_resist_modifier, 0, 1)
+/mob/living/carbon/xenomorph/get_fire_resist()
+	return clamp((100 - get_soft_armor("burn", BODY_ZONE_CHEST)) * 0.01 + fire_resist_modifier, 0, 1)
```
KornFlaks, line 35:
```suggestion
	return clamp((100 - get_soft_armor("fire", BODY_ZONE_CHEST)) * 0.01 + fire_resist_modifier, 0, 1)
```

"burn" armor doesnt exist. It's "fire". BURN remaps to "fire".

I'm gonna spin up a PR refactor replacing all these direct quotes with macros.

code/modules/mob/living/living_defense.dm
```diff
@@ -134,6 +134,8 @@
 		return FALSE
 	if(!CHECK_BITFIELD(datum_flags, DF_ISPROCESSING))
 		return FALSE
+	if(get_fire_resist() <= 0 || get_hard_armor("fire", BODY_ZONE_CHEST) >= 100)	//having high fire resist makes you immune
```
KornFlaks, line 137: Did you change how resin jelly works? It's still magic-numbering fire_resist_modifier and removing the magic number check here breaks resin jelly.

code/modules/mob/living/living_defense.dm
```diff
@@ -279,3 +281,6 @@
 		. = shield_check.Invoke(attack_type, ., damage_type, silent)
 		if(!.)
 			break
+
+/mob/living/proc/get_fire_resist()
+	return clamp((100 - get_soft_armor("burn", BODY_ZONE_CHEST)) * 0.01, 0, 1)
```
KornFlaks, line 286: The chest body specifier is not needed. It's not used in the generic mob living soft armor function.

KornFlaks (changes requested):

code/modules/mob/living/living_defense.dm
DeltaFire15: oops

code/modules/mob/living/carbon/xenomorph/xenomorph_defense.dm
DeltaFire15: missed that one when I accidentally typed them as burn then changed them to fire.

code/modules/mob/living/living_defense.dm
DeltaFire15: not clamping the resist value should do it since all instances of handling get_fire_resist catch multipliers <= 0

code/modules/mob/living/living_defense.dm
DeltaFire15: humans do not override get_fire_resist but do overide get_soft_armor. Therefore, it is still needed or it will runtime.

KornFlaks: Oh right. I changed that in my own armor refactoring. Woops.

code/modules/mob/living/living_defense.dm
KornFlaks: It might be better to instead add Tivi's robotic trait TRAIT_NON_FLAMMABLE and remove once the jelly runs out.

I dont know if that will then nullify damage when moving across flames though.

DeltaFire15: I'll see about it when I get around to test, if it doesn't work, maybe.

KornFlaks: Just crunched the math. -20 is plenty but eww, magic number.

Feel free to resolve this comment though. This will work.

Lumipharon:
Nice fixes, I knew fire armour was jank, but not this bad.

I would recommend not making fire armour effect fire stacks, as that represents something flammable being on the mob. Being entirely fire proof doesn't stop someone pouring fuel over your head, and fire armour reducing damage received is enough there.

Similarly, I would keep the xeno passive stack removal to resting only, otherwise applying stacks without ignition is going to be functionally pointless.

DeltaFire15:
> Nice fixes, I knew fire armour was jank, but not this bad.
> 
> I would recommend not making fire armour effect fire stacks, as that represents something flammable being on the mob. Being entirely fire proof doesn't stop someone pouring fuel over your head, and fire armour reducing damage received is enough there.
> 
> Similarly, I would keep the xeno passive stack removal to resting only, otherwise applying stacks without ignition is going to be functionally pointless.

The issue here is that fire stacks are used both as "how much material is on you" aswell as "how on fire are you" / "whats the intensity of your fire". A variety of things that don't put material on you change fire stacks which in their case is just setting you on fire, e.g. lasers and so on. Not having it affected by armor would lead to jelly being a trap in that it would cause the next ignition after using it to be far stronger due to still getting firestacks despite not being flammable at all. I suppose it would be an option to only block fire stack additions if at full resistance instead as this is the main concern of counterintuitive behavior with them.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -44,6 +44,8 @@
 /mob/living/carbon/xenomorph/handle_fire()
 	. = ..()
 	if(.)
+		if(resting && fire_stacks > 0)
+			adjust_fire_stacks(-1)	//Passively lose firestacks when not on fire while resting and having firestacks built up.
```
KornFlaks, line 48: Resting and maybe resisting as well? This PR made me realize that xeno firestack icon isnt actually bugged, it displays the real remaining firestacks on a xeno and the default method to remove fire is resisting.

Resting runs counter to the standard xeno method of removing fire and will need to be displayed in chat or in a balloon_alert maybe?

```suggestion
			adjust_fire_stacks(-1)	//Passively lose firestacks when not on fire while resting and having firestacks built up.
			balloon_alert(src, "Firestack reduced")
```

I dont know if src is right though.

DeltaFire15: Well, this is for the passive decay of "status buildup" that the next "status trigger" aka ignition would use to have stronger effects. I don't think it needs to be displayed more than the icon decreasing as it'd just spam you with messages while resting. Besides, it *should* be pretty rare to occur now that jelly isn't comedically backfiring like right now (noone ever noticed that though)

DeltaFire15:
Haha queens have 100 fire armor which never did anything because of fire not caring but now makes them fully fireproof. I'm gonna have to ask maintainers for a good replacement value.

DeltaFire15:
> I would recommend not making fire armour effect fire stacks, as that represents something flammable being on the mob. Being entirely fire proof doesn't stop someone pouring fuel over your head, and fire armour reducing damage received is enough there.

After some testing, fire stack buildup from many effects tends to be significant enough for fire stack armor to not null them too much besides for very fire resistant xenos, although it does make continued buildup more of a strategy instead of many things directly capping stacks.  I think I'll be keeping its effect on fire stacks for now. (walking through 3 tiles of fire still caps fire stacks of 30 armor castes, while 2 have it at about ~75%)

code/modules/mob/living/living_defense.dm
DeltaFire15: does in fact seem to work pretty well, in fact better than before I believe due to also affecting some other cases.

code/modules/mob/living/carbon/xenomorph/xenomorph_defense.dm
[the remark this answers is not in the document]
DeltaFire15: should be very easy to fix, we'll just see whose gets merged first.

code/modules/mob/living/carbon/xenomorph/life.dm
KornFlaks:
```suggestion
			adjust_fire_stacks(-1)	//Passively lose firestacks when not on fire while resting and having firestacks built up.
			to_chat(src, span_bold("Firestack reduced due to resting."))
```

Possibly instead of a balloon alert then just to_chat? Passive actions even with the flame hud icon changes seem iffy. I know sunder has no chat response but at least it has a game tab value that increases to signify something is happening.

Fancy chat will collapse the to_chat spam to just (x3) and so on so there's no flooding of the chat box.

DeltaFire15: Nothing else that has passive effects gives you messages for what it does and this is a super-edge-case thing anyways since most times if you get stacks you are getting set on fire and being extinguished tends to happen at 0 stacks.

QualityVan:
I might suggest buffing flamer ammo burn level by a third to roughly equal the average 25% or so xeno armor.

KornFlaks:
Make sure to check for xeno flag `CASTE_FIRE_IMMUNE` as well.

DeltaFire15:
> Make sure to check for xeno flag `CASTE_FIRE_IMMUNE` as well.
A bunch of things were already accounted for by it although some weren't (fire stacks / ignition). Those are now, although impact damage from being hit by something that can cause fire still applies.

> I might suggest buffing flamer ammo burn level by a third to roughly equal the average 25% or so xeno armor.
Flamer ammo's burn level (aka damage / fire stack) has been increased by 30%, other things have been left untouched. I'd rather avoid modifying values too much in this initial PR especially before it is seen on the server.
Do note however that xenomorph fire armor is affected by sunder, so this the armor average tends to be somewhat lower than the initial armor values due to many marine weapons causing sundering.

code/modules/mob/living/carbon/xenomorph/life.dm
```diff
@@ -44,10 +44,12 @@
...
```
Wayland-Smithy, line 48: IMO this should be sped up a little. -1 is as fast as being on fire and you have to be resting, plus you can't resist to shake it off any faster because that signal is woven into actual ignition. Coded it to happen via Life() all the time for xenos when not on fire then decided to check this PR just in case.
```suggestion
			adjust_fire_stacks(-2)	//Passively lose firestacks when not on fire while resting and having firestacks built up.
```

DeltaFire15: Having fire stacks while not on fire, with this PR at least, should be a very rare case that needs some special circumstances to occur. Currently, jelly prevents igniting but not gaining fire stacks (which e.g. walking on fire causes), which is how this is seen so commonly. With this PR, if you can't be ignited / are fully resistant (like jelly does), you shouldn't get fire stacks. This means that the only way for this to happen is to a) not be fully resistant to fire b) not be on fire and c) be affected by something that increases fire stacks without igniting you. Said combination of events (should) only happen with e.g. splashing ethanol or simillar on a xenos (which, who does that?) and is very very rare. This passively slowly decaying when resting is effectively just a catch so you don't have it for forever if something like that tags you and doesn't follow up with fire anytime soon, if you want to get rid of it quicker you can always set yourself on fire in a controlled way which immediately allows you to burn away the stacks.

Wayland-Smithy: > With this PR, if you can't be ignited / are fully resistant (like jelly does), you shouldn't get fire stacks.

Okay that was my main concern and in-game it appears to be working that way so this is fair enough then. Consider it resolved. Thanks for the detailed response.

KornFlaks:
People are reporting that underbarrel flamer is doing an extreme amount of damage on hit but that's probably because my armor PR was un-TM'ed.

DeltaFire15:
> People are reporting that underbarrel flamer is doing an extreme amount of damage on hit but that's probably because my armor PR was un-TM'ed.

Kinda peculiar because it checks for the fire resistance multiplier when "hitting" something with the fire cone, but also uses take_overall_damage_armored so it should effectively work the same in either situation (your PR fixes the armored check while the fire mod does nothing, my PR fixes the damage being changed by the fire mod while the armor does nothing. I'll test around with it sometime.

PR #10228, opened 2022-05-16
Laser MG buffed in normal shot, efficency reworked.

Pariah919:
## About The Pull Request
Someone suggested this for the Laser MG to not straight up be complete trash, and so it might be interesting to try.
Normal shot has 1.5 sunder over 1 sunder.
Efficency reduced to 5 damage, upped to 20 pen so it actually still does SOME damage to things.
Efficency now has 3 sunder.

## Why It's Good For The Game
Makes efficent lasgun fire actually do something, and makes sense considering it'll have some level of gimmick now.
Laser MG is in a terrible spot, so this could be interesting to try.

## Changelog
:cl:
balance: Laser MG normal fire does .5 more sunder, efficency does 5 damage over 8.5, and 20 pen over 15. Does 3 sunder on hit over 1.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1823,11 +1823,14 @@ datum/ammo/bullet/revolver/tp44
 	name = "machine laser bolt"
 	damage = 15
 	penetration = 15
+	sundering = 1.5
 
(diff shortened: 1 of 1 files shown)


Pariah919:
guh, back to the drawing board

PR #10229, opened 2022-05-16
Give energy sniper rifle aim mode, increase fire delay, & balance its heat mode

novaepee:
## About The Pull Request

Title.

https://cdn.discordapp.com/attachments/504094319075131402/975634899266580500/2022-05-15_22-43-55.mp4

## Why It's Good For The Game

Barely anyone use energy sniper rifle. You have to frontline with it, which defeats the purpose, or you are content with accidently FFing marines (with the latter more plausable).

Many people agree that the energy sniper rifle needs more, but they fear that a hitscan sniper will wreck balance. In an attempt to calm such doubts, I ensure that in heat mode it shall take all the energy of the energy sniper rifle, meaning that any gamers that want to lit xenomorphs up from afar shall have to reload every time they use head mode energy sniper rifle.

I also put an additional second fire delay on standard ammo. Pray that I don't increase fire delay.

## Changelog
:cl:
add: aim mode on energy sniper rifle
balance: aim mode on energy sniper rifle
balance: increase fire delay to 2 seconds on energy sniper rifle
balance: heat mode, which lit xenomorphs on fire, on energy sniper rifle uses all of 600 energy in mag, therefore one needs a fresh mag for every heat mode shot
/:cl:

PR #10230, opened 2022-05-16
Laser sniper changes (IFF mode)

Arctrooper8:
## About The Pull Request

Before anyone says anything about Elsa's PR #10229
image
I've had this idea and been working on it for longer so this isn't some kneejerk reaction.

Changes the sniper rifles main fire mode to have the IFF tag (Does not affect heat shot)
This IFF mode loses damage logarithmically for each friendly target it passes through.
Changes its sunder to 4 -> 1 
Increases the damage from 60 -> 85

Laser sniper can no longer take magnetic harness (it has an unremovable scope so you shouldn't be able to put it on in the first place but oh well)
Laser sniper can no longer take under barrels
Laser sniper can now take the bipod 

##I would like to stress that none of these variables are permanent and are all open to discussion

## Why It's Good For The Game

Everyone knows that the laser sniper rifle is bad in its current form, it has quite honestly always been bad.
Compare the laser sniper rifle to the SR-127 often considered the quintessential round start sniper rifle:
image
SR-127 bullet:
image

That's pretty good. Sure you're not gonna be deleting any xenos with it but you can provide support. In comparison the laser rifle is awful.

Now, the major issue with the laser sniper is the fact that is doesn't have aim mode and cannot actually snipe. However, given that the main contention with giving laser sniper aim mode is the heat mode (due to lasers being hitscan, and instant long ranged ignition application is frowned upon) why not give the laser sniper something unique without putting heat mode into an odd spot. 

In #10229 the heat mode is essentially soft removed (unless you have a handcrank or are sitting at fob with the battery recharger, both of which are reliant on Engineering). There is also the fact that even with the cost now being the entire magazine it will still be incredibly unfun if just infrequent (Nobody wants to be instantly set on fire from off screen, and there are plenty of ways for it to be spammed).

In comparison, by not giving the heat mode the ability to bypass marines it becomes a risk reward situation. Where the sniper must get closer and in front of their allies to get a heat shot, it also gives the sniper a closer range defensive tool in case of ambush opening up an interesting role for the laser sniper user.

Overall giving the laser sniper rifle an IFF mode gives it a unique gimmick whilst avoiding the unfun mechanic that hitscan aim mode heat instant burn heat shots would be.

TLDR: Laser sniper aim mode is a cool and unique gimmick, and having aim mode hitscan heat shots is unfun for Xenos no matter the ammo cost.

## Changelog
:cl:
balance: Laser sniper rifle now has an IFF mode that decreases in damage logarithmically for every marine it passes through (try to get a clear shot!)
fix: You can no longer force a mag harness unto the laser sniper rifle and bug out the scope sprite
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1810,22 +1810,20 @@ datum/ammo/bullet/revolver/tp44
 /datum/ammo/energy/lasgun/marine/sniper
 	name = "sniper laser bolt"
 	hud_state = "laser_sniper"
-	damage = 60
+	damage = 70
 	penetration = 30
-	sundering = 4
+	flags_ammo_behavior = AMMO_ENERGY|AMMO_IFF|AMMO_SUNDERING|AMMO_HITSCAN|AMMO_SNIPER
 	max_range = 40
 	damage_falloff = 0
 	hitscan_effect_icon = "beam_heavy"
 
 /datum/ammo/energy/lasgun/marine/sniper_heat
-	name = "sniper IFF bolt"
+	name = "sniper heat bolt"
 	icon_state = "microwavelaser"
 	hud_state = "laser_heat"
-	shell_speed = 2.5
 	damage = 40
 	penetration = 0
-	flags_ammo_behavior = AMMO_ENERGY|AMMO_SUNDERING|AMMO_HITSCAN|AMMO_IFF
-	sundering = 1
+	flags_ammo_behavior = AMMO_ENERGY|AMMO_INCENDIARY|AMMO_SUNDERING|AMMO_HITSCAN
 	hitscan_effect_icon = "u_laser_beam"
 
 /datum/ammo/energy/lasgun/marine/pistol
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1810,7 +1810,7 @@ datum/ammo/bullet/revolver/tp44
 /datum/ammo/energy/lasgun/marine/sniper
 	name = "sniper laser bolt"
 	hud_state = "laser_sniper"
-	damage = 85
+	damage = 70
 	penetration = 30
 	flags_ammo_behavior = AMMO_ENERGY|AMMO_IFF|AMMO_SUNDERING|AMMO_HITSCAN|AMMO_SNIPER
 	max_range = 40
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1818,13 +1818,13 @@ datum/ammo/bullet/revolver/tp44
 	hitscan_effect_icon = "beam_heavy"
 
 /datum/ammo/energy/lasgun/marine/sniper_heat
-	name = "sniper heat bolt"
+	name = "sniper IFF bolt"
 	icon_state = "microwavelaser"
 	hud_state = "laser_heat"
 	shell_speed = 2.5
 	damage = 40
 	penetration = 0
-	flags_ammo_behavior = AMMO_ENERGY|AMMO_INCENDIARY|AMMO_SUNDERING|AMMO_HITSCAN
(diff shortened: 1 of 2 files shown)


DeltaFire15:
Hmm, what if instead of a low magsize / damage buff, it would instead lose damage / effect strength for sunder for each friendly it passes? Reasoning being something about the IFF calculating the beam past friendlies having to defocus it leading to a weaker ray or whatever else one can come up with.

Arctrooper8:
> Hmm, what if instead of a low magsize / damage buff, it would instead lose damage / effect strength for sunder for each friendly it passes (probably on a logarithmic scale in that the difference between 0 and 1 friendly penned is highish, 1 ->2 is lower, 2 -> 3 lower still and so on)? Reasoning being something about the IFF calculating the beam past friendlies having to defocus it leading to a weaker ray or whatever else one can come up with. This would allow using it in the backline but with less damage depending on how many friendlies you get inbetween, which still incentivizes getting a free (or at least not _as_ occupied) line of fire.

Now that is a really good idea that I'll look into, I might need some help with the math though.

Lumipharon:
> Hmm, what if instead of a low magsize / damage buff, it would instead lose damage / effect strength for sunder for each friendly it passes (probably on a logarithmic scale in that the difference between 0 and 1 friendly penned is highish, 1 ->2 is lower, 2 -> 3 lower still and so on)? Reasoning being something about the IFF calculating the beam past friendlies having to defocus it leading to a weaker ray or whatever else one can come up with. This would allow using it in the backline but with less damage depending on how many friendlies you get inbetween, which still incentivizes getting a free (or at least not _as_ occupied) line of fire.

This functionality already exists, although I don't believe any guns currently use it. See var/damage_marine_falloff

Arctrooper8:
Logarithmic stuff added and tested, ready for TM

Shirbu:
The fact that this is now the highest damage sniper weapon with the largest magazine capacity by 5 more than the next highest and fastest firerate and built in IFF (meaning max firerate even when friendlies are around) is critically unbalanced, and causes this weapon to functionally upseat every other sniper weapon available both backline and frontline.

Arctrooper8:
> This would become the highest damage per shot sniper weapon, with the largest magazine capacity by 5 more than the next highest, and fastest firerate, and built in IFF (meaning max firerate even when friendlies are around).
> 
> Unless I am misunderstanding some part of this.

It loses logarithmically loses damage for every marine it passes through, and only the normal shot is IFF. For example: Pass through one marine lose 8.5 damage (it is currently set to 10%) pass through two marines lose 8.5+6.25 ish damage etc etc.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1810,9 +1810,9 @@ datum/ammo/bullet/revolver/tp44
 /datum/ammo/energy/lasgun/marine/sniper
 	name = "sniper laser bolt"
 	hud_state = "laser_sniper"
-	damage = 60
+	damage = 85
```
BraveMole, line 1813: this seems completly uneeded

BraveMole (changes requested):

Tailiik:
If you want the heat shot to be a quick defense swap perhaps it could be an alt fire rather than a mode switch, doing unique action and clicking a transparent button disrupts aim long enough that you might as well drop the sniper and swap to a secondary.

code/modules/projectiles/ammo_datums.dm
Arctrooper8: I wanted to give it a damage boost because it now loses damage for each marine it passes through, however you're right that +25 damage is unneeded. How about 70 instead? If that is still too much I'll just leave it at 60.

Arctrooper8:
> If you want the heat shot to be a quick defense swap perhaps it could be an alt fire rather than a mode switch, doing unique action and clicking a transparent button disrupts aim long enough that you might as well drop the sniper and swap to a secondary.

Good idea, I'll add that to my list

BraveMole (approved):

PR #10227, opened 2022-05-16
Sandbags are unmeltable.

Pariah919:
## About The Pull Request
Makes sandbags unmeltable.

## Why It's Good For The Game
Gives sandbags a bit more of a gimmick, and mainly buffs foward FOBs, so you can infact not look at that single tile of drones and not get your entire fioward fob instantly melted by a single drone.

## Changelog
:cl:
balance: you can no longer melt sandbag barricades
/:cl:

PR #10232, opened 2022-05-17
Caustic upgrade cades cannot be melted.

Pariah919:
## About The Pull Request
image
probably more balacned and enjoyable than #10227, but if people want 227 they can reopen it.

## Why It's Good For The Game
More balanced then sandbags, overall. And I feel like foward FOBs could use this. You need quite a bit of metal to make this work.

## Changelog
:cl:
balance: Caustic armor cades are no longer meltable by vomit acid.
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -534,6 +534,7 @@
 			soft_armor = soft_armor.modifyRating(melee = 30, bullet = 30)
 		if(CADE_TYPE_ACID)
 			soft_armor = soft_armor.modifyRating(bio = 0, acid = 20)
+			resistance_flags = XENO_DAMAGEABLE|UNACIDABLE
 
 	barricade_upgrade_type = choice
 


Kuro020:
better balanced than the sandbag pr

Lumipharon:
Is this really a good idea?

BraveMole (changes requested):
You can remove a upgrade.
Also you need enable the xeno meltable flag, not set the whole bitfield

Naaanii:
This makes more sense as opposed to the sandbag PR as they can still be slashable and spat at and require engineers to set it up. I'd be down to increase the mat cost per upgrade too and maybe buff the other cade upgrades as they seem very underwhelming.

Ragecrabbo:
> This makes more sense as opposed to the sandbag PR as they can still be slashable and spat at and require engineers to set it up. I'd be down to increase the mat cost per upgrade too and maybe buff the other cade upgrades as they seem very underwhelming.

Honestly this, razorwire seems to be the only cade upgrade you see regularly used in rounds (and actually have a noticable effect) unless engies get bored and are sitting on unlimited req points.

Lumipharon:
Is this actually a problem though, it's not like anyone thinks fobbing is too weak right? I mean I hope it won't happen, but if you have acid proof cades and keep repairing them, you will HAVE to melee your way through them, which is difficult if ungas have enough firepower.

FosterGrime-Heymynameslouie:
spedd merhg

Naaanii:
> Is this actually a problem though, it's not like anyone thinks fobbing is too weak right? I mean I hope it won't happen, but if you have acid proof cades and keep repairing them, you will HAVE to melee your way through them, which is difficult if ungas have enough firepower.

True, but they also can be spat at as they are not immune to spitting. All this really does is stops one drone from freely melting a whole line of unattended cades, or walling and corner melting cades but a runner can still slash them all down or a spitter can still spit them from range.

Caustic cades already feel rather underwhelming with how strong spit castes are, so this change seems alright and doesn't seem too oppressing either.

breadhunt:
To me this seems like just more of a buff to cading off flanks/unattended cades. Melting isn't that useful in sieges (with competent engis) usually. 

But having to slash an entire cadeline with barbed wire _is_ annoying as FUCK

If it's unmeltable, it shouldn't have barbed wire. Grayson said something similar in sandbag PR, it's just shitty to deal with.
Other than that good PR

BraveMole:
I'm not sure my comment was clear enough, so:

You can remove upgrades, yet you are not resetting the flags there, leading to a bug

CursedBirb (commented):
Line 773 add this:

resistance_flags = XENO_DAMAGEABLE

Im not a coder but worked for me on local and when removed upgrade acid worked again

Pariah919:
#10306 is just this but not hit by smallscale burnout

PR #10233, opened 2022-05-18
Rework accuracy mechanics

Lumipharon:
## About The Pull Request
This PR is a refactor and rework of accuracy to,
A: Fix a whole bunch of bugs in accuracy code
B: Make accuracy actually do something
C: By extension of above, make attachment choices actually mean something beyond magharness, bayo and underbarrel weapons

I'm very much open to feedback on how to tweak this or whether this is salt inducing trash, and this will definitely need test merging to see how it actually plays out in reality.

**Accuracy mechanic changes**
The main actual changes (that aren't just bugfixes) are as follows:

1. The penalty to firing when moving now applies to wielded fire rather than just unwielded fire, but at half the rate. 
To accompany this, all guns have had their movement penalties tweaked, so guns you'd expect to perform well on the move like shotguns/SMG's/carbines etc have lower modifiers, while heavy weapons like snipers and machineguns perform worse when moving.

2. The size of the mob now changes the chance to hit. Monkeys and the like get a negative bonu to hit for being manlet size, while xenos get a bonus to hit for being big (with a larger bonus for big xenos like queen etc).

3. The target mob is harder to hit when they are moving. The faster you are moving (i.e. lower slowdowns, run vs walking, sticky resin, high pain, frenzy, orders, etc all effect this), the higher your evasion bonus.
Combined with the size changes, smaller xenos will tend to be a bit harder to hit, while large xenos will probably be a little easier.

4. You no longer get a 2nd "roll" to hit another bodypart when you miss by 20% or less. Instead, if you actually hit but only just, you'll likely hit a bodypart you weren't aimming for.
This was a HUGE factor for why accuracy seems to do basically nothing right now. As an example, an accuracy of 80% meant you actually hit around **94%** of the time. 

**Then there a variety of smaller changes/fixes:**

1. Most guns have had some tweaks to their accuracy or related stats where appropriate. In particular, snipers no longer have accuracies that went up into the 200's
2. The 'focus' order no longer DECREASES your accuracy if your leadership isn't high enough
3. The  'focus' order has been buffed somewhat to actually be maybe useful
4. You aren't magically easier to hit when unconscious vs just lying down
5. Accurate minimum range now actually works, and works for any gun that has it, not just ones specifically designated as snipers
6. Accurate max range now causes a drop off in accuracy per tile **exceeding** that range, rather than the **total** range travelled
7. Pain decreases accuracy a little bit more than it did previously (yes this is a thing)
8. Projectile damage now only gets the 50% FF reduction if it actually **hits**
9. Rolled the stat bonuses from unremovable stocks into the guns themselves because it was confusing as shit (and removed some unused stocks from the code)
10. Rolled the accuracy bonus **ALL** projectiles got into the base projectile accuracy because it was confusing and served no purpose
11. Reduced the flat accuracy bonus lasers got because it's meme

And maybe some other stuff.

**What does this mean for the average unga?**
The first thing you'll notice is that shooting one handed on the move tends to be a bad idea, despite me actually buffing many guns that often get used this way. This is because they had dogshit accuracy, but the jank code meant they just hit most of the time anyway. Attachments will of course offset this to a degree, but even 'one hand weapons' will still be better wielded where possible.

Attachment choices now actually have some relevance.

Currently attachment choices has two very simple flows:
1: Does my gun have high recoil/scatter I need to sort out? Add a Recoil comp and/or gyro
2: Add a mag harness, bayonet and underbarrel weapon where possible

Going forwards, you can still do this and be entirely fine, your DPS might just be marginally lower since you'll actually miss every now and then. Significantly though, adding just one accuracy improving mod (like an extended barrel instead of a bayo) pushes your accuracy up an actually noticable chunk.

And lastly, with some guns it will actually pay to stand still while shooting, instead of sprinting after beno, and gyro will actually do something for wielded guns.

For xenos, I have kept their high accuracy bonus since they're designed to be running around all the time, and frankly they don't need a nerf. So they probably won't notice much change when it comes to shooting marines.

## Why It's Good For The Game
Lots of bug fixes good.
Accuracy actually works, creating some small level of decision making for weapon and attachment choice.

## Changelog
:cl:
add: Projectile accuracy is impacted by the shooter moving or standing still (even when wielding)
add: Projectile accuracy is impacted by how fast the target is moving, and how large they are
balance: Projectiles no longer have two chances to hit a target instead of one. Low accuracy shots will now actually miss more in line with what you would expect
balance: A balance pass has been made of guns and some attachments relating to their accuracy
fix: Fixed a variety of accuracy related bugs
fix: Removed stat modifiers from non removable weapon stocks and added them to their base gun stats

/:cl:


lKiseki:
Raising initial concerns here, based entirely off of what I've read in this PR;
1. Is it safe to assume the evasion bonus is not going to be stupid good for castes like Runner?
2. How much accuracy do snipers get? I think they ought to get decently good accuracy given they're, well, snipers.
3. Do these changes imply that one-handed firing got nerfed in a way? If so, I do think one-handed should remain viable, at least on the side of accuracy.

Lumipharon:
> Raising initial concerns here, based entirely off of what I've read in this PR;
> 
>     1. Is it safe to assume the evasion bonus is not going to be stupid good for castes like Runner?
> 
>     2. How much accuracy do snipers get? I think they ought to get decently good accuracy given they're, well, snipers.
> 
>     3. Do these changes imply that one-handed firing got nerfed in a way? If so, I do think one-handed should remain viable, at least on the side of accuracy.

Evasion is capped at 30%. Some modifiers are additive, where as evasion is a multiplier, so the lower your accuracy, the less (in actually percentage chance to hit terms) effect it has.

Snipers are still extremely accurate. Base they're like 104%, before attachments. And scoping in adds 40% ontop of this. So realistically they're still going to be like, 150%+ before evasion etc.
So it just means they don't have ludicrously high accuracy so that they can never miss no matter what, now negative factors mean they CAN miss, especially if running and gunning.

One handed didn't get directly nerfed, I actually buffed basically every gun you might currently one hand. It's just the mechanics didn't really work well for them previously. And once you add on attachments, (standard stuff like RC and laser sight on a T90 etc) you've got respectable accuracy. But it means if you do have the chance to wield your gun, it WILL actually be a noticable difference.

code/modules/projectiles/projectile.dm
```diff
@@ -729,64 +729,76 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
...
+	if(hit_chance <= 0) //If by now the sum is zero or negative, we won't be hitting at all.
 		return FALSE
 
 	var/hit_roll = rand(0, 99) //Our randomly generated roll
 
-	if(. > hit_roll) //Hit
+	if(hit_chance > hit_roll) //Hit
+		//friendly fire reduces the damage of the projectile, so only applies the multiplier if a hit is confirmed
+		if(shooter_living)
+			if(shooter_living.faction == faction)
```
BraveMole, line 796: shooter_living?.faction

BraveMole (approved):

Lumipharon:
PR updated, xenos generally easier to hit, gun wielded scatter tweaked down slightly. Wielded fire is either equal or slightly better scatter, while moving wielded fire is equal or slightly worse.

code/modules/projectiles/projectile.dm
Lumipharon: Done

FosterGrime-Heymynameslouie:
image


my take away from this is that being a manlet rocks

Kuro020:
but why does accuracy need to be a thing? if you want a gun to be bad long range you have falloff, if you want a gun to miss you have scatter, accuracy seems unecessary here, not only that but missing shots that actually connected because rngesus decided so is a major feelsbad for the player

Lumipharon:
> but why does accuracy need to be a thing? if you want a gun to be bad long range you have falloff, if you want a gun to miss you have scatter, accuracy seems unecessary here, not only that but missing shots that actually connected because rngesus decided so is a major feelsbad for the player

Scatter is flawed, it is irrelevant at low amounts, and entirely irrelevant at close range.
Long range shooting isn't really effected by this, as all long range weapons are super accurate, and most of them have low or no damage fall off, which is fine.

Missing isn't RNG lose game like pokemon or something, it's just a reduction in dps as shooting in this game is just spam loads of bullets.

The only guns where it is more significant to miss one shot are shotguns and snipers. Both of which have great accuracy. If shotguns are an issue I can increase their accuracy, that's just a balance issue. But PB (2 tile range) is still mega accurate, and now bayo isn't just the autopick attachment regardless of other considerations.

Kuro020:
> > but why does accuracy need to be a thing? if you want a gun to be bad long range you have falloff, if you want a gun to miss you have scatter, accuracy seems unecessary here, not only that but missing shots that actually connected because rngesus decided so is a major feelsbad for the player
> 
> Scatter is flawed, it is irrelevant at low amounts, and entirely irrelevant at close range. Long range shooting isn't really effected by this, as all long range weapons are super accurate, and most of them have low or no damage fall off, which is fine.
> 
> Missing isn't RNG lose game like pokemon or something, it's just a reduction in dps as shooting in this game is just spam loads of bullets.
> 
> The only guns where it is more significant to miss one shot are shotguns and snipers. Both of which have great accuracy. If shotguns are an issue I can increase their accuracy, that's just a balance issue. But PB (2 tile range) is still mega accurate, and now bayo isn't just the autopick attachment regardless of other considerations.

Except you're already wrong since scatter got reworked, even low ammmounts of scatter matter, and scatter/falloff arent even the only stats that can make guns balanced at close/long range. The sniper for example was always sucky at close range even before this pr. 
The problem here is that bullets that physically hit xenos should never rng miss, because this is a top down shooter and that feels like shit.
So again what does accuracy accomplish here that other gun stats dont? because from what I can feel from playing with it, it just slaps a RNG miss chance on a game that really doesnt need one

PR #10240, opened 2022-05-19
Xeno Pheromones Selection Now Made in Radial Menu

KornFlaks:
## About The Pull Request

Xeno pheromones (queen and standard) are now unified into a single radial menu. The buttons previously placed along the actions list have now been relocated to a radial menu activated by selecting the emit pheromones action.

Furthermore, refactoring of the pheromone code has been done in order to bring this section of logic up to current day standards.

image

**Sprite changes are simply renaming to match defines. No spite artwork changed.**

## Why It's Good For The Game

The actions bar of xenos are extremely cluttered. The queen most of all. It has always been a goal to reduce the number of action buttons and pheromone selection was one of the greatest offenders.

Well no longer. The buttons are now a modern radial selection menu akin to weed type selection and drone primordial plant selection.

Require #10250 to be merged in order for keybindings to work.

## Changelog
:cl:
qol: Xeno pheromones selection are now done in a radial menu.
fix: Xeno phermones keybindings now function at all times, even if the phero selection is not active.
/:cl:

PR #10250, opened 2022-05-19
Refactors Hivemind Ability Checks to Not Use should_show()

KornFlaks:
## About The Pull Request

Due to a previous exploit with hiveminds, should show xeno action proc was used as a easy override check to ensure a hivemind that is not corporal could not use specific abilities.

This however broke hidden keybind only buttons I intended on implementing with my pheromone wheel of fortune #10240.

Instead, the could use check has been properly relocated to can_use_action.

## Why It's Good For The Game

Required fixes to allow for hidden keybind only buttons to be functional.

image

## Changelog
:cl:
qol: The actions that hiveminds can use while manifested no longer disappear but instead are disabled.
/:cl:

### DIFF
--- a/code/datums/actions/xeno_action.dm
+++ b/code/datums/actions/xeno_action.dm
@@ -104,9 +104,6 @@
 			to_chat(owner, span_warning("We can't do this while in a solid object!"))
 		return FALSE
 
-	if(!should_show())
-		return FALSE
-
 	return TRUE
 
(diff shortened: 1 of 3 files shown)


FosterGrime-Heymynameslouie:
speed mehrge so that the other pr that needs this also gets spped mehrg

TiviPlus (approved):

PR #10258, opened 2022-05-20
CSE Update

Naaanii:
## About The Pull Request
Gives CSE starting mats

## Why It's Good For The Game

CSE is a glorified engineering support role but it they start with no materials and an empty electronics pouch for some reason. This should update the role a bit further.

## Changelog
:cl:
balance: CSE now starts with one stack of metal, plasteel, sandbags and barbed wire
balance: CSE now starts with a welderpack instead of a satchel
/:cl:


Derrick2Electricboogaloo:
We get this based PR just to get an equally unbased PR removing metal scrapable items from ships when it really was not a big deal and hurts shipside roles more.

PR #10259, opened 2022-05-21
Underbarrel weapon rebalance

Lumipharon:
## About The Pull Request
This PR adds wield time and wielded movement speed maluses to underbarrel weapon attachments.

Underbarrel weapons provide marines huge damage and/or utility buffs with literally zero downside.
This is in comparison with most other attachments that have some sort of downsides for the benefits they provide.

The changes are pretty mild. All 5 underbarrels now add 0.2 second of wield time.

## Why It's Good For The Game
Make underbarrel weapons slightly more reasonable

## Changelog
:cl:
balance: Underbarrel weapons now slightly impact wield times for their parent guns.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -303,6 +303,10 @@
 	var/attach_delay = 0 SECONDS
 	///Time it takes to detach src to a master gun.
 	var/detach_delay = 0 SECONDS
+	///Changes the slowdown amount when wielding a weapon by this value.
+	var/aim_speed_mod	= 0
+	///How long ADS takes (time before firing)
+	var/wield_delay_mod	= 0
 
 
 /*
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -303,8 +303,6 @@
 	var/attach_delay = 0 SECONDS
 	///Time it takes to detach src to a master gun.
 	var/detach_delay = 0 SECONDS
-	///Changes the slowdown amount when wielding a weapon by this value.
-	var/aim_speed_mod	= 0
 	///How long ADS takes (time before firing)
 	var/wield_delay_mod	= 0
 
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1704,7 +1704,6 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	if(!istype(attached_to, /obj/item/weapon/gun))
 		return
 	master_gun = attached_to
-	master_gun.aim_slowdown					+= aim_speed_mod
 	master_gun.wield_delay					+= wield_delay_mod
 	if(gun_user)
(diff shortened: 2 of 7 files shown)


Naaanii:
Wield time is good, wield speed seems kind of lame.

Pariah919:
I don't really see why you'd nerf all the underbarreled attachments, and honestly other than PP and UBF they're kinda mediocre in general.

Lumipharon:
> I don't really see why you'd nerf all the underbarreled attachments, and honestly other than PP and UBF they're kinda mediocre in general.

It's 0.1, which is negligible. No one's going to stop using them because of it, and a whole bunch of attachments that provide very tame benefits have genuine drawbacks, compared to underbarrel weapons which have absolutely none, which is super inconsistant.

> Miniflamer is pretty much as good as the plasma pistol, I don't see the need for snowflake nerfs particularly. Perhaps dealing with the attachments themselves would be a better option rather than nerfing underbarrel as a whole.

PP is objectively the most powerful underbarrel since it has the 40 damage flat + fire with no real range limit and 10 shots.
UGL can do more damage but has only 2 shots and no more impact grenades so are less reliable, and you can't use it up close.
MK has a 0.6 damage modifer and only 3 shots. It's only real strength is it gives you stuns, which is OK I guess.
Miniflamer is good, but it got the on hit damage fixed + only 4 tile range, so it's good for area denial but not for killing things.

Naaanii:
> > I don't really see why you'd nerf all the underbarreled attachments, and honestly other than PP and UBF they're kinda mediocre in general.
> 
> It's 0.1, which is negligible. No one's going to stop using them because of it, and a whole bunch of attachments that provide very tame benefits have genuine drawbacks, compared to underbarrel weapons which have absolutely none, which is super inconsistant.
>
0.1 alone is negligible but it still stacks up with other slowdown values. Hitting underbarrels with a double whammy nerf is also wack. One drawback seems good enough and nerfing underbarrel as a whole seems kind of unnecessary.

goose4429:
Smol changes, seem reasonable. But larger underslugs should have more penalty, and plasma pistol is the smallest of all. I'd make them all the same.

Al-1ce:
> Smol changes, seem reasonable. But larger underslugs should have more penalty, and plasma pistol is the smallest of all. I'd make them all the same.


Agreed. It's overall much more tactile to add a standardized penalty rather than selectively changing slightly different variables just because one is bigger and what-not.

Lumipharon:
Equalised and added pepperball because I forgot it exists.

Al-1ce:
> Equalised and added pepperball because I forgot it exists.

Looks good. I've also noticed how almost every other attachment came with a downside, whereas the only downside underbarrels had was that you could pick one. B-18 underslungs are no more!

Lauwma:
Please do not add slowdown, marines are already quite slowed. Just increase wield time it makes sense.

BraveMole (changes requested):
I don't like aim slowdown added. wield time is fine

Lumipharon:
Changed to wield time only

code/modules/projectiles/gun_system.dm
```diff
@@ -303,6 +303,10 @@
 	var/attach_delay = 0 SECONDS
 	///Time it takes to detach src to a master gun.
 	var/detach_delay = 0 SECONDS
+	///Changes the slowdown amount when wielding a weapon by this value.
+	var/aim_speed_mod	= 0
```
BraveMole, line 307: unused

BraveMole (changes requested):

BraveMole (approved):

PR #10264, opened 2022-05-21
Adds a projectile nozzle for the flamer

QualityVan:
## About The Pull Request
Adds a nozzle for the marine standard flamethrower that causes it to shoot a standard projectile instead of flaming a line of turfs. It's in the usual spot of the attachment vendor. 

## Why It's Good For The Game
More flamer nozzle options! Sacrifices area coverage for fuel efficiency and a faster fire rate. Possibly just a strictly worse plasma pistol, but hey you can still extinguish with it.

## Changelog
:cl:
add: Extended nozzle for the standard flamer in the weapon vendor. Shoot fireballs to ignite things from a distance! Up front damage is pretty lacking though.
/:cl:


SilencedMP5A5:
flamer like the one ds13 has when

Kuro020 (approved):

Al-1ce:
Seems more suited to hallway area denial than anything else. That's going to be a pretty useless niche though. I could imagine it coming in somewhat useful if flamers had a pouch or kit with different nozzles inside of it that they could switch from on the go. Dunno.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2576,24 +2576,28 @@ datum/ammo/bullet/revolver/tp44
 	hud_state = "flame"
 	hud_state_empty = "flame_empty"
 	damage_type = BURN
-	flags_ammo_behavior = AMMO_INCENDIARY|AMMO_IGNORE_ARMOR|AMMO_FLAME
+	flags_ammo_behavior = AMMO_INCENDIARY|AMMO_IGNORE_ARMOR|AMMO_FLAME|AMMO_EXPLOSIVE
```
TiviPlus, line 2579: What?

TiviPlus (changes requested):

code/modules/projectiles/ammo_datums.dm
QualityVan: Explosive ammo tag basically means 'can be aimed at an open turf and will land there instead of continuing past', which seemed like a reasonable and desired thing for it to have. Plasma pistol's the same.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2564,24 +2564,28 @@ datum/ammo/bullet/revolver/tp44
...
+	max_range = 7
+	damage = 0
 	bullet_color = LIGHT_COLOR_FIRE
 	var/fire_color = "red"
 	var/burnlevel = 31
 	var/burntime = 17
-	var/fire_delay = 20
+
+/datum/ammo/flamethrower/drop_flame(turf/T)
+	if(!istype(T))
```
BraveMole, line 2577: What is that ? Why is it needed

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2594,11 +2598,10 @@ datum/ammo/bullet/revolver/tp44
 /datum/ammo/flamethrower/blue
 	name = "blue flame"
 	hud_state = "flame_blue"
-	max_range = 6
+	max_range = 7
```
BraveMole, line 2601: Uh, why is this in this pr ?

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2594,11 +2598,10 @@ datum/ammo/bullet/revolver/tp44
...
 	fire_color = "blue"
 	burnlevel = 46
 	burntime = 40
-	fire_delay = 20
```
BraveMole, line 2601: Ditto

BraveMole (changes requested):
Why all those changes to normal flamer? Does it have any effect ?

code/modules/projectiles/ammo_datums.dm
QualityVan: This change exclusively affects use of the ammo datum as a projectile (same with the damage change). The other nozzles operate off the gun's max_range for determining how long of a line to ignite and won't be affected by any of the ammo datum changes unless you touch burn_time and burn_level, since they don't actually fire projectiles.

code/modules/projectiles/ammo_datums.dm
QualityVan: This is just cleanup. Flamer ammo has a this var uniquely (line 2575 above) which is applied to the flamer on reload, but all flamer ammo has an identical delay to it and it's applied in a weird way (on reload, the gun's fire delay is set to the ammo datum's initial, then set back to the gun's initial on unload). I adjusted the flamer's default fire_delay to the value applied by the ammo, since that's what it was always.

QualityVan: The exception to this is the miniflamer - it has a coded fire_delay of 2.5s, but since loading ammo sets it anyway this has never done anything before.

code/modules/projectiles/ammo_datums.dm
QualityVan: This is actually just me cargo coding from the tank flamer code. I think it's for firing past the edge of the map.

FosterGrime-Heymynameslouie:
I cast, fireball!! 

does it penetrate resin walls though? cuz it might be a bit op if it does

QualityVan:
> does it penetrate resin walls though? cuz it might be a bit op if it does

The base flamer already does that but no, the ball from this will hit a wall and ignite that tile.

BraveMole (approved):

PR #10272, opened 2022-05-22
Maturity and evolution towers only work after shutters opening

BraveMole:
## About The Pull Request

Tittle

## Why It's Good For The Game

Will discourage fob hugging and xenos immediatly crushing marines. This is more in line with other xeno building, like silos which are not producing anything before the shutters open

This is made with nuclear war in mind, I hope this won't create too much trouble in distress

## Changelog
:cl:
balance: Maturity and evolution towers only work after shutters opening
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
@@ -16,10 +16,6 @@
 	var/list/list/xenos_by_upgrade = list()
 	var/list/dead_xenos = list() // xenos that are still assigned to this hive but are dead.
 	var/list/list/xenos_by_zlevel = list()
-	///list of evo towers
-	var/list/obj/structure/xeno/evotower/evotowers = list()
-	///list of upgrade towers
-	var/list/obj/structure/xeno/maturitytower/maturitytowers = list()
 	var/tier3_xeno_limit
 	var/tier2_xeno_limit
 	///Queue of all observer wanting to join xeno side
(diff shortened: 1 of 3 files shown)


Naaanii:
I feel like this only addresses the early game and doesn't do much to curb the powerspike Xenos generally get with their respawns. A way to rework how fast Xenos mature or where towers can be placed would be neat, considering all towers are usually placed in caves by the silo, but forcing them out in the open leaves them open to OBs and other artillery.

Al-1ce:
This might _seem_ like a sensible change - however as mentioned above would only be marginally effective until the shutters are down, ultimately giving marines much more time to ungaball a weaker and less prepared hive. 

I don't see this harming Nuclear War too much, although Distress is definitely going to be a marine-major most of the time with a change like this in store.

With that said, Marines at round-start have access to a _plethora_ of deadly and effective tools for completely free. Xeno's shouldn't be penalized for a lack of utilization by the opposing team, vice-versa.

Dzhimax:
the author of this shit is the greatest clown in the history of the world

goose4429:
From my experience (300h as xeno, 100h as queen, if you don't think that enough then disregard comment) - evolution towers are literal trash, because you are only constrained by maturity. And maturity towers are mostly only useful after all primos are bought. Roundstart you are growing faster by buying primos. Mid-game primos are again more useful because they have combat impact. Not saying a bad PR, but they definitely need a buff.

TiviPlus (changes requested):
Only for maturity towrs cus evo is already bad enough

PR #10282, opened 2022-05-23
Removes Largely Unused Overwatch Keybind (Ctrl-Middle Click) and OW Changes

KornFlaks:
## About The Pull Request

Deletes control middle click overwatch selection for overwatch.

Overwatch now cancels if the watcher is damaged.

## Why It's Good For The Game

Removes code bloat. Also prevents xenos from being surprise killed when they're watching someone else.

## Changelog
:cl:
add: Overwatch is now canceled if the watcher is damaged.
refactor: Overwatch from chat now uses target reference instead of trying to match nicknames. This should mean it should no longer fail to work if xeno names are identical.
del: Queen can no longer select xenos to overwatch using control middle click.
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -73,9 +73,6 @@
 	if(modifiers["shift"] && modifiers["ctrl"])
 		CtrlShiftClickOn(A)
 		return
-	if(modifiers["ctrl"] && modifiers["middle"])
-		CtrlMiddleClickOn(A)
-		return
 	if(modifiers["middle"] && MiddleClickOn(A))
 		return
 	if(modifiers["shift"] && modifiers["right"])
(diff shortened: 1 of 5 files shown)


Arctrooper8:
Can you make it so that you can use overwatch while resting please? Considering that as a Queen using the ability can be done while resting it would make sense you can watch others while resting via the F key too.

Arctrooper8:
I am blind, I see the fix now. Many thanks!

KornFlaks:
PAIN

code/_onclick/click.dm
```diff
@@ -499,13 +496,6 @@ if(selected_ability.target_flags & flagname && !istype(A, typepath)){\
 	SEND_SIGNAL(src, COMSIG_CLICK_CTRL_SHIFT)
 
 
-/*
-	Ctrl+Middle click
-*/
-/atom/proc/CtrlMiddleClickOn(atom/A)
-	SEND_SIGNAL(src, COMSIG_CLICK_CTRL_MIDDLE, A)
-
```
TiviPlus, line 507: ???
why are you removing this??

TiviPlus (changes requested):

code/_onclick/click.dm
KornFlaks: The only action using control middle mouse click is the queen overwatch selection.

Leaving it in would be dead code. Doing nothing.

If someone wants to add it back in for another feature or action, they are welcome to do so. Same with any other combination.

KornFlaks: Oh, I didnt delete the comsig declaration. I'll need to do that tomorrow

PR #10284, opened 2022-05-24
IB/QCP rebalances

QualityVan:
## About The Pull Request
IB wounds don't get worse over time. Bloodloss is slightly scaled up to keep it consistent with old values.
Bic OD can't cure IB. Get QCP or use surgery or cryo.
Instead of healing IB's damage over time QCP now removes one wound entirely after every ~20s. The cloneloss gain is tied to this removal, as is a final chunk of brute damage to the parent limb.
QC/low damage no longer prevents IB from damaging the limb it's on. Bic+ina still does.

## Why It's Good For The Game
Puts to rest any issues with QCP seeming to not work.
Makes it easier to tell which limb specifically has internal bleeding, since you can pause the bleeding with QC but let the minor limb damage gain continue (0.1 per 2s, so any brute healing chem is more than enough to keep it from mattering if you want).
Makes it easier for corpsmen to tell when one IB wound is cured, since the patient will get hit with a bit of brute (this can't dismember or cause organ damage or another internal wound, but it could in theory push you past the threshold for a limb break. It probably won't unless you're specifically trying to.)

## Changelog
:cl:
balance: IB won't naturally get worse over time.
balance: Quickclot won't stop IB from slowly increasing bruteloss on the limb it's on. Bic+ina still will.
balance: Bic OD won't slowly cure IB.
fix: QCP will always remove internal bleeding so long as you have at least 5u injected and no purgatives. Every 5u injection will handle one instance of IB.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -779,7 +779,7 @@
 	if(!ishuman(L))
 		return
 	var/mob/living/carbon/human/body = L
-	for(var/datum/limb/possible_limb in body.limbs)
+	for(var/datum/limb/possible_limb AS in body.limbs)
 		for(var/datum/wound/internal_bleeding/possible_IB in possible_limb.wounds)
 			target_IB = possible_IB
 			RegisterSignal(target_IB, COMSIG_PARENT_QDELETING, .proc/clear_wound)
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -774,8 +774,41 @@
 	custom_metabolism = REAGENTS_METABOLISM * 2.5
 	///The IB wound this dose of QCP will cure, if it lasts long enough
 	var/datum/wound/internal_bleeding/target_IB
+	///Ticks remaining before the target_IB is cured
+	var/ticks_left
+	///Ticks needed to cure an IB
+	var/ticks_to_cure_IB = 10
 
 /datum/reagent/medicine/quickclotplus/on_mob_add(mob/living/L, metabolism)
+	select_wound(L)
+
+/datum/reagent/medicine/quickclotplus/on_mob_delete(mob/living/L, metabolism)
+	if(target_IB)
+		to_chat(L, span_warning("The searing pain in your [target_IB.parent_limb.display_name] returns to a dull ache..."))
+		UnregisterSignal(target_IB, COMSIG_PARENT_QDELETING)
+		target_IB = null
+
+/datum/reagent/medicine/quickclotplus/on_mob_life(mob/living/L, metabolism)
+	L.reagents.add_reagent(/datum/reagent/toxin,5)
+	L.reagent_shock_modifier -= PAIN_REDUCTION_VERY_HEAVY
+	L.adjustStaminaLoss(15*effect_str)
+	if(!target_IB)
+		select_wound(L)
+		ticks_left-- //Keep it at the total ticks_to_cure, including the tick used to select a wound
+		return ..()
+	ticks_left--
+	if(!ticks_left)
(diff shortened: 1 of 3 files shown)


novaepee:
NOOOOOOOOOOOOOOO MY BICA OD

code/modules/reagents/reagents/medical.dm
```diff
@@ -774,39 +768,56 @@
...
+	///The IB wound this dose of QCP will cure, if it lasts long enough
+	var/datum/wound/internal_bleeding/target_IB
 
 /datum/reagent/medicine/quickclotplus/on_mob_add(mob/living/L, metabolism)
-	if(TIMER_COOLDOWN_CHECK(L, name))
+	if(!ishuman(L))
 		return
-	L.adjustCloneLoss(5*effect_str)
+	var/mob/living/carbon/human/body = L
+	for(var/datum/limb/possible_limb in body.limbs)
```
BraveMole, line 782: As in for both loops

code/modules/reagents/reagents/medical.dm
```diff
@@ -774,39 +768,56 @@
...
 
 /datum/reagent/medicine/quickclotplus/on_mob_life(mob/living/L, metabolism)
-	var/mob/living/carbon/human/H = L
-	for(var/datum/limb/X in H.limbs)
-		for(var/datum/wound/internal_bleeding/W in X.wounds)
-			W.damage = max(0, W.damage - (2.5*effect_str))
 	L.reagents.add_reagent(/datum/reagent/toxin,5)
 	L.reagent_shock_modifier -= PAIN_REDUCTION_VERY_HEAVY
 	L.adjustStaminaLoss(15*effect_str)
+	if(current_cycle == 9 && target_IB) //Minimum 5u per dose, and make sure nothing else has cleared it up in the meantime.
```
BraveMole, line 802: 9 is a magic number

code/modules/reagents/reagents/medical.dm
```diff
@@ -774,39 +768,56 @@
...
+		to_chat(L, span_alert("The searing pain in your [target_IB.parent_limb.display_name] peaks, then slowly fades away entirely."))
+		target_IB.parent_limb.createwound(CUT, target_IB.damage / 2)
```
BraveMole, line 804: Why that create wound ?

BraveMole (changes requested):
Like the idea

code/modules/reagents/reagents/medical.dm
QualityVan: Second one needs to be that way because it's looking specifically for IB and not just any wounds, but you're right on the first

code/modules/reagents/reagents/medical.dm
QualityVan: A lot of reagent code has hardcoded cycle checks, where would you prefer it?

code/modules/reagents/reagents/medical.dm
QualityVan: Like the pr desc says, it gives visible feedback on scanner of one IB getting cured for a medic if the patient has multiple. IB damage is usually 15, so this is fairly minor in terms of impact.

BraveMole: Ok

Lumipharon:
I'm not necessarily opposed to this or anything, but "QCP doesn't work" is 99% due to people not know they should double dose and OD people with it for significantly improved effect.

Losenis:
>Quickclot won't stop IB from slowly increasing bruteloss on the limb it's on
>Bic OD won't slowly cure IB
>You need a minimum of 5u of cloneloss juice for it to do anything

Isn't this just making "Kill the guy, surgery his IB away on a roller bed, remove necro, revive him" more viable

PR #10288, opened 2022-05-25
Fix Turf Targeted Balloon Message Spam

KornFlaks:
## About The Pull Request

Balloon alerts acting on turf objects are replicated for all identical turfs.

That is bad. Therefore, no message for turfs.

## Why It's Good For The Game

image

## Changelog
:cl:
fix: Balloon alerts resulting from an action targeting the ground no longer spams balloon alerts.
/:cl:

### DIFF
--- a/code/modules/balloon_alert/balloon_alert.dm
+++ b/code/modules/balloon_alert/balloon_alert.dm
@@ -33,6 +33,10 @@
 // I would've made the maptext_height update on its own, but I don't know
 // if this would look bad on laggy clients.
 /atom/proc/balloon_alert_perform(mob/viewer, text)
+	if(isturf(src))
+		// Balloon alerts occuring on turf objects result in mass spam of alerts.
+		return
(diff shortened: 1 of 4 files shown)


code/modules/balloon_alert/balloon_alert.dm
```diff
@@ -33,6 +33,10 @@
 // I would've made the maptext_height update on its own, but I don't know
 // if this would look bad on laggy clients.
 /atom/proc/balloon_alert_perform(mob/viewer, text)
+	if(isturf(src))
```
BraveMole, line 36: No istype src. Just override the proc

BraveMole (changes requested):

code/modules/balloon_alert/balloon_alert.dm
KornFlaks: Done.

PR #10275, opened 2022-05-23
xeno defines redo to bypass bitfield limit

Mechanelize:
## About The Pull Request

redoing of xeno defines to bypass bitfield 21 line limit

## Why It's Good For The Game

allows for more xeno defines

## Changelog
:cl:
code: Split xeno related flags containing "CAN" into a new category called xeno_can_flags to bypass the 21 flag bitfield limit
/:cl:

PR #10293, opened 2022-05-26
Larva Can Now Evolve Without Hive Ruler

KornFlaks:
## About The Pull Request

Larva, in their conversion to using standard evolution logic, can now no longer evolve if there is no shrike or queen. Such as round start.

This fixes that.

Once #10275 is merged, this will instead be converted to a xeno caste flag check so it's not larva snowflake code.

## Why It's Good For The Game

Larva can now evolve round start without having to ask admins.

## Changelog
:cl:
fix: Larva now no longer need a hive ruler to evolve.
/:cl:


TiviPlus (changes requested):
Talk to me about this istype src check, you should know they're not allowed in general

KornFlaks:
This should work better then.

PR #10294, opened 2022-05-26
Larvae get two evo points per tick if on weed, again

DeltaFire15:
## About The Pull Request
This PR returns larva to 2 (base) evo / tick if on weed as it's just time spent waiting for the evo to tick up and young xenos are still very squishy regardless. As before, if not on weed this bonus is lost.
~~It was also fixing the leader stuff regarding larva but *someone* opened their PR two seconds before I hit open on mine so I reverted that part out of it and changed this description~~

## Why It's Good For The Game
Larva being able to reach an actual xeno quicker as before the evo rework is a good thing, given that larva is just a sitting around state even more than young xenomorphs and young xenomorphs are still quite vulnerable. Besides, this also promotes larva remaining on weed as opposed to being overly daring.

## Changelog
:cl:
balance: Larva gets 2 points evo / tick if on weeds, like before the larva evo point rework.
/:cl:


KornFlaks:
Remove the comment on the larva ruler check.

I agree but it's gonna merge conflict once mine goes through as well.

Walarks:
I think larva should get the bonus regardless of weeds or not get a bonus at all. I personally like the consistency of getting the same evo points regardless of what you stand on. 

Also I don't think this helps make larvas less daring since weeds are often right outsside fob, where xenos can often die from getting chased/bombed while on weeds.

A change I'd be more ok with is making larvas gain double evo on weeds, but not any points while off weeds since this is a much clearer interaction and can enable more interesting scenarios than just increasing the time you have to hide.

Arctrooper8:
Why not just remove larva evo points at all, make it so they can evolve right as they come out of the ground/silo?
Waiting as a larva is even worse than waiting as a young tier one, because as a larva you can do practically nothing. The overall balance effects of this would be minor, and the improvement to the xeno gameplay experience well worth it.

MiniMeatwad:
How about just making evolution take 5 seconds or something? Can't evolve immediately, but it might as well be.

DeltaFire15:
I dislike having zero larva evo off weed as there's times where this can happen quite possibly, e.g. a shipside burst or someone doing experiments again.
Nor do I like having it just doubled per default (or zero at all), due to the fact that larba should be pretty vulnerable at least for some time (whereas somewhat less so on weed). The pre-evo-rework amount worked fine at that time and I don't see much reason to change it.
If I had to choose between the two extremes, I'd choose just halving required points, however I'd prefer not to choose either and use the middle way.

Issue #10298, opened 2022-05-27
Timers can be deleted without their callback proc being triggered


Lumipharon:
### Description

When the server is under a bit of load (i.e. the round is in full swing and lots of people are shooting and doing stuff), the timer SS or something connected to it can remove timers without actually triggering their callback as expected.

This is responsible for bullet_ping overlays getting stuck sometimes, and is almost certainly causing #9867 (and the duplicate #9596), where guns never get switched out of their 'bursting' state.

Bullet pings are the most obvious victim of this issue due to the sheer quantity of them in a game, but I suspect anything that user a timer could potentially be impacted in the same way.

### Test Merges

NA

### Reproduction Steps

Shoot a wall in the middle of a high pop game
Watch it ping forever

KornFlaks:
Cannot replicate

https://cdn.discordapp.com/attachments/500801682058379265/979588190241427546/unknown.png

![image](https://user-images.githubusercontent.com/29745705/170624529-34d92dcc-cb46-46f1-8243-e4d6aa4fc168.png)

BraveMole:
This is also causing shield module recharge to fail sometimes

BraveMole:
This issue doesn't exist actually

PR #10300, opened 2022-05-27
Converts Bullet Ping Overlay Duration to Subsystem

KornFlaks:
## About The Pull Request

* Currently on the live server, there are issues with bullet pings targeting walls and objects persisting beyond their assigned 3 tick duration.

* Normally, the removal of this animation is done by a timer callback to cut_overlays() on the object that the bullet ping is rendering on top of. However, for some inexplicable reason, this callback either fails or does not properly work.

* Me and several other testers have spent hours attempting to reproduce this on a local server using the master (current) code to no avail, as documented in issue #10298

## Why It's Good For The Game

This is a potential fix to that issue by removing the requirement for a timer callback using a miniature subsystem that records each bullet overlay occurring in the tick and cycles through a rotating buffer to clear previous cycle bullet overlays.

Again, in testing this functions well to remove all bullet pings but note, in testing we were also **unable to replicate the persistent bulling pings** despite some rather extreme circumstances.

## Changelog
:cl:
code: Bullet ping overlay duration is now handled by a subsystem and not timers.
/:cl:


lKiseki:
can we give this man a medal already for doing so many good prs

Lumipharon:
Just a note that while this may fix bullet pings,there is an underlying timer issue that's causing other problems (broken burst firing guns, arrowhead shield, etc), which will ultimately need to be fixed.

PR #10307, opened 2022-05-27
Removes mini scope from RL-152

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
Instakill from offscreen is quite unfun.

## Changelog
:cl: Lewdcifer
balance: RL-152 no longer accepts mini scopes.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -611,7 +611,6 @@ Note that this means that snipers will have a slowdown of 3, due to the scope
 
 	flags_gun_features = GUN_WIELDED_FIRING_ONLY|GUN_AMMO_COUNTER
 	gun_skill_category = GUN_SKILL_FIREARMS
-	starting_attachment_types = list(/obj/item/attachable/scope/mini)
 	dry_fire_sound = 'sound/weapons/guns/fire/launcher_empty.ogg'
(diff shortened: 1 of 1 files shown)


Pariah919:
Eh, SADARs cost 120 points, but I guess I can understand people's frustration with this one, the RR one is just what, though.

Naaanii:
Seems fine but removing it from both is better consistency.

Lumipharon:
Is this really that much an issue? It gets a tiny increase in view distance, but needs a clear LOS unlike TAT with aim mode and much longer range.

Araulius:
if this goes through, adjustments to the HE shell need to be made. such as, less radius, but more damage within it.

Lauwma:
it's a rocket launcher, not a close quarter combat weapon

Wayland-Smithy:
> Eh, SADARs cost 120 points,

The "req items do not need to be wallenced" argument is getting really old when points are all but free if you put BAS and normal sentries on 1 miner next to FOB or there is a researcher making 100s or any effort is put to selling never decaying corpses.

Xenos can not be revived and dying offscreen no matter the myriad of requirements, no IFF, needs LOS, etc.. is not fun gameplay.

TiviPlus:
If its too weak people can lower the price

PR #10308, opened 2022-05-27
Removes mini scope from T-160 RR

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
Full sunder from offscreen is quite unfun.

## Changelog
:cl: Lewdcifer
balance: T-160 RR no longer accepts mini scopes.
/:cl:


Pariah919:
I don't like this because the RR is supposed to be a support weapon, and the explosion range is supposed to make it risky to use onscreen without a Miniscope, and LE is made for long range sniping, don't like this change.

lKiseki:
Probably unneeded for the RR, closing it.

PR #10309, opened 2022-05-28
Drops nuclear war silo scaling from 2.5 to 2.

Pariah919:
## About The Pull Request
Probably a good idea as a banaid fix while people rack their heads around how to better design this mode, in my opinion.

## Why It's Good For The Game
Might be better to make this map based to be honest, but in most maps xenos don't really need the incredibly high scaling, and death should atleast have some meaning in nuclear war.

## Changelog
:cl:
balance: silo scaling is 2 from 2.5
/:cl:

### DIFF
--- a/code/datums/gamemodes/nuclear_war.dm
+++ b/code/datums/gamemodes/nuclear_war.dm
@@ -1,7 +1,7 @@
 /datum/game_mode/infestation/distress/nuclear_war
 	name = "Nuclear War"
 	config_tag = "Nuclear War"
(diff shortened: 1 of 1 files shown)


FosterGrime-Heymynameslouie:
this is a long time coming honeslty,
but im sad to say that those all hope is lost, a swarm of xenos is coming and no backup will arrive moments will be missed (probably not by the majority of people)

Al-1ce:
X*no main here. Seems reasonable.

PR #10311, opened 2022-05-28
Flamers are stronger against resin walls

BraveMole:
## About The Pull Request

Revert flamer nerf from a while ago.

## Why It's Good For The Game

Engineers should not be the only pratical way for marines to clear out mazes. Mazes should not be as sturdy as they are especially in nuclear war

## Changelog
:cl:
balance: Resin walls take twice as much damage from flamer
/:cl:

### DIFF
--- a/code/game/turfs/walls/resin.dm
+++ b/code/game/turfs/walls/resin.dm
@@ -22,7 +22,7 @@
 
 
 /turf/closed/wall/resin/flamer_fire_act()
-	take_damage(25, BURN, "fire")
+	take_damage(50, BURN, "fire")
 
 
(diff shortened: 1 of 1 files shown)


matt81093:
Surprised this is a full revert instead of a partial increase. Although I suppose it's harder to push on some maps/areas than others.

Losenis:
I'd go far enough to say have them start at a lower health too, hivemind can drop like 5 of them in quick succession from range and it takes several hits for them to go down.

novaepee:
FLAMERS ARE COMING BACKKKKKKKKKK

PR #10313, opened 2022-05-28
Mobs no longer breathe

BraveMole:
## About The Pull Request

Should not affect balance

## Why It's Good For The Game

That's a relic from atmoscode, it's jsut wasting processing time

## Changelog
:cl:
code: Mobs no longer breath
/:cl:


code/modules/mob/living/carbon/life.dm
```diff
@@ -174,70 +171,6 @@
...
 
-/mob/living/carbon/proc/breathe()
-	if(!need_breathe())
-		return
-
-	if(pulledby && pulledby.grab_state >= GRAB_KILL)
-		Losebreath(3)
-
-	if(health < get_crit_threshold() && !reagents.has_reagent(/datum/reagent/medicine/inaprovaline))
-		Losebreath(1, TRUE)
```
QualityVan, line 185: This and the lines following is the code that makes it so mobs in crit suffocate and die. I'd recommend leaving in breathe(), need_breathe() to handle stasis stuff, and handle_breathe changed to just run off a bool as to whether you should heal or take damage that breath. It's also going to kill needing a mask for ice maps and anesthetic tanks working on surgical beds.

goose4429:
What about  Boiler/Defiler clouds?

BraveMole:
> What about Boiler/Defiler clouds?
not affected

XANAv2:
I don't know if this is intended but it seems that anesthetic tanks don't work or is this unrelated, affecting only Xenos.

QualityVan:
> I don't know if this is intended but it seems that anesthetic tanks don't work or is this unrelated, affecting only Xenos.

Are you seeing this with operating tables or with anesthetic tanks on people's backs?

QualityVan:
Having thought about it some more, this probably also needs removal of the ability to turn on internals since it doesn't do anything any more.

TiviPlus:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/10331 should be fine

PR #10314, opened 2022-05-28
Adds an argument to flamer_fire_act, buffs flamer against resin

QualityVan:
## About The Pull Request
All of the flamer_fire_act calls already used this argument, it just was ignored by most of the proc defs. Also adds one missing span call.
Buffs flamers against resin walls. One flame duration will always destroy a resin wall, taking ~15 seconds to do so. Miniflamers are nerfed by roughly an equivalent amount, so they're the same against mazes and weaker against mobs. They won't take a fully matured wall down in one shot, but will kill fresh ones.
Alternative to #10311 

## Why It's Good For The Game
Lets different flamers/ammo more uniformly burn stuff at different rates.
Improves flamers at one of their dedicated roles, area clearing.
Nerfs miniflamers, which're apparently still overperforming.

## Changelog
:cl:
code: flamer_fire_act gets a new argument that, among other things, means x-fuel is better than default at burning non-mob stuff.
balance: Flamers chew through resin faster.
balance: Miniflamers are weaker against mobs, but the same as before against resin.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -417,8 +417,8 @@ directive is properly returned.
 	return //For handling the effects of explosions on contents that would not normally be effected
 
 
-//Generalized Fire Proc.
-/atom/proc/flamer_fire_act()
+///Generalized Fire Proc. Burn level is the base fire damage being received.
+/atom/proc/flamer_fire_act(burnlevel)
 	return
 
 
--- a/code/game/objects/effects/decals/Cleanable/fuel.dm
+++ b/code/game/objects/effects/decals/Cleanable/fuel.dm
@@ -87,7 +87,7 @@
 		ignite_fuel(I)
 		log_attack("[key_name(user)] ignites [src] in fuel in [AREACOORD(user)]")
 
-/obj/effect/decal/cleanable/liquid_fuel/flamer_fire_act()
+/obj/effect/decal/cleanable/liquid_fuel/flamer_fire_act(burnlevel)
 	. = ..()
 	ignite_fuel()
 
--- a/code/game/objects/items/bodybag.dm
+++ b/code/game/objects/items/bodybag.dm
@@ -218,11 +218,11 @@
 		to_chat(bodybag_occupant, span_danger("You jolt out of [name] upon being hit!"))
 		open()
 
-/obj/structure/closet/bodybag/flamer_fire_act()
+/obj/structure/closet/bodybag/flamer_fire_act(burnlevel)
 	if(!opened && bodybag_occupant)
 		to_chat(bodybag_occupant, span_danger("The intense heat forces you out of [name]!"))
 		open()
-		bodybag_occupant.flamer_fire_act()
+		bodybag_occupant.flamer_fire_act(burnlevel)
 
 /obj/structure/closet/bodybag/ex_act(severity)
 	if(!opened && bodybag_occupant)
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -83,7 +83,7 @@
 	explosion(loc, light_impact_range = src.light_impact_range, small_animation = TRUE)
 	qdel(src)
 
-/obj/item/explosive/grenade/flamer_fire_act()
+/obj/item/explosive/grenade/flamer_fire_act(burnlevel)
 	activate()
 
 /obj/item/explosive/grenade/attack_hand(mob/living/user)
--- a/code/game/objects/items/explosives/grenades/marines.dm
+++ b/code/game/objects/items/explosives/grenades/marines.dm
@@ -16,7 +16,7 @@
 	throw_range = initial(throw_range)
 
 
-/obj/item/explosive/grenade/training/flamer_fire_act()
+/obj/item/explosive/grenade/training/flamer_fire_act(burnlevel)
 	return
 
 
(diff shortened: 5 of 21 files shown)


Losenis:
LOOKS LIKE FLAMERS ARE BACK ON THE MENU BOYS

Arctrooper8:
This is an absolutely absurd PR, we've been here before in the game's history. It makes it near impossible to have a maze stand for any considerable amount of time.

Arctrooper8:
Also making miniflamer worse yet leaving plasma pistol untouched is very silly when they perform a similar role and nerfing one will just result in the other being used more.

Losenis:
>It makes it near impossible to have a maze stand for any considerable amount of time.

Because there's no middle ground between "Maze is clearable in an actual tolerable amount of time, let's clear it" and "We'll get arthritis by the end of it and nobody signed up to play cookie clicker, let's push the maze". Conversely there's no middle ground between "Maze is CBT so it's worth building" and "Marines can clear the maze without getting bored out their ass and pushing it, so it's not worth building".

>nerfing one will just result in the other being used more

Wasn't plasma pistol going to be removed from underbarrel attachables?

Arctrooper8:
> Because there's no middle ground between "Maze is clearable in an actual tolerable amount of time, let's clear it" and "We'll get arthritis by the end of it and nobody signed up to play cookie clicker, let's push the maze". Conversely there's no middle ground between "Maze is CBT so it's worth building" and "Marines can clear the maze without getting bored out their ass and pushing it, so it's not worth building".

The thing is that we have actually been here before in the game's history. Flamers used to completely demolish mazes leaving Xenos with little to no cover, this is essentially a revert back to those times. It's a cycle for god's sake and we're going round and round. The fact alone that mazes used to be nothing to fire and we decided that it was better for them to be awful to deal with should speak tons about why this PR is a bad idea.

### STUCK IN A LOOP STUCK IN LOOP AAAAAAAAAAAAAAA!!

> Wasn't plasma pistol going to be removed from underbarrel attachables?

It was only removed from shotguns, it is still an attachment for the majority of weapons. Honestly we should just delete the Plasma pistol entirely, we have the projectile flamer to fulfill that niche now.

Losenis:
>and we decided that it was better for them to be awful to deal with should speak tons about why this PR is a bad idea.

Speak for yourself instead of trying to astroturf this "we" that only includes people agreeing with you, because I sure as fuck am not included in that "we" and nor is every marine pushing the 1x1 instead of playing cookie clicker. No, it was an awful idea because making a game aspect deliberatedly unfun to play to balance it out despite being a critical part that will possibly decide whether you win or lose is so absolutely cancerous some people just stopped caring about playing the game and effectivelly suicided out for 2 mins because it was more fun. It wasn't inconsequential to xenos getting like 80% winrate on nuclear one day when they had enough numbers to almost guarantee someone would start mazing.

Not to mention it's not back to the same thing. Back then hivemind didn't have corporeal long ranged wall spam, PC wasn't nerfed, walls started with way less health that you could knife (ie not T60 bayo) them faster than a drone would build them, nuclear didn't exist, etc. Sacrificing fun for balance is how you get people to ignore it altogether and make balance inconsequential whent they avoid interacting with it even if it literally kills them.

Dueu:
So here are my 2 cents: The plasma cutter nerf was needed and badly, I and other 3 engineers regulary not only cut through every maze but also entire buildings and caves as well because we could just load a new battery whenever we run out of energy and keep cutting everything. Now its only useful when we need to quickly cut down a choke point that can't be destroy with fire.

While the fire buff certainly makes flamers marines much more powerful at slow pushes, you also need to remember that mazing is risk free and drones/hivelords can give a temporal immunity to fire IF they actually do remember to keep producing and dropping the resin jelly and resin jelly pods. There was this nuclear war round that I just hold the green disk terminal building in Big Red as drone and by the way is the closest to both LZs and the marines simply couldnt get in and push me out because of the maze in the building and me constantly walling off every single push, then the other xenos just flank from outside and I just repaired the maze. If one xeno player using only a T1 caste to the full potential can literally hold against marines because there wasn't a good handheld tool for marines to destroy mazes then resin building is absolutly overpowered.

BraveMole:
> This is an absolutely absurd PR, we've been here before in the game's history. It makes it near impossible to have a maze stand for any considerable amount of time.

sure but it was in distress time. Mazes are less important for xenos now, since they defend less

MiniMeatwad:
Bad

Losenis:
Either dishonest exaggeration or mad salting. It takes a caste as much time to build a wall as it takes a flamer on extended nozzle to cooldown, not this weird 10 minute number you pulled from somewhere. This is only accurate when you're completely 100% walling off an entire room like in nuclear or crash, in which case yes, this is working as intended for the counter it is in those circumstances. This also applies to tightly packed mazes that have zero open space if you're also making doors on the corridors like a couple days ago in LV caves, in which case yeah that's also meant to be a counter much like how I can't complain if I spend 10 minutes premedding myself on every chem and then a defiler instadeletes them in one click. A milder instance is a flamer being able to torch down an entire section of maze due to it being conveniently built in such a way that it can be neatly flamed in a line or so tightly built that it can be covered with the spray nozzle, in which case, I'd sooner blame the builder for building it in the perfect way to counter it.

Or just tell the builder to build more. Plasma is infinite, the backtank can only hold so much fuel.

This is also discounting the ton of FF that inevitably happens when marine neurons activate on seeing a xeno past the 1x1 choke on fire so they rush it not paying attention to. At one point I just stopped wasting water on idiots that deliberatedly ran in FRONT of flamers clearing walls and blocking shots on top of wasting time and reducing their own usefulness as meatshields.

And the one factor that is intensifying this to an extreme: Every unga and their mother will go for whatever got updatedfor the novelty, flamer in this case. Xenos will face up to 4 flamers because that's the limit on vendors. Yes, mazing will inevitably be extremely difficult when the marine force intentionally or not geared up to counter mazes to the absolute limit their team-wide equipment pool allows. If marines still die to mazes despite them exhausting all the flamers, there is a very big problem.

PR #10316, opened 2022-05-28
Heart and lung damage affect stamina, health and speed

BraveMole:
## About The Pull Request

Heart damage will decrease your maximum stamina and your maximum health
Lungs damage will decrease your stamina regen and add a slowdown

All other effects have been removed.

Remove peridaxon from the game

## Why It's Good For The Game

I think dying from an invisible wound suck. Losing blood from lungs and heart damage sucks. Fainting because of oxyloss because your heart is damaged from one warrior punch sucks.

The game should be fun for the players, not pain ; on the other hand, xenos should be able to permanently cripples marines, else the unga ball would not be counterable. So this is a compromise, as I hope that this is less cancer for marine players, and still balanced for xenos.

## Changelog
:cl:
balance: Heart damage will decrease your maximum stamina and maximum health. Nothing else
balance :Lungs damage will decrease your stamina regen and your speed. Nothing else
delete: Peridaxon is no longer in the game, as effects from organ damage are less severe
/:cl:


code/modules/organs/organ_internal.dm
```diff
@@ -122,9 +118,25 @@
...
 	if (!silent)
 		owner.custom_pain("Something inside your [parent.display_name] hurts a lot.", 1)
+	set_organ_status()
+
+/// Set the correct organ state
+/datum/internal_organ/proc/set_organ_status()
+	if(damage > min_broken_damage)
+		if(organ_status != ORGAN_BROKEN)
+			organ_status = ORGAN_BROKEN
+		return TRUE
```
QualityVan, line 128: are the TRUE returns meaningful here? seems like it happens in all cases except where the organ was healthy both previously and now

code/modules/organs/organ_internal.dm
```diff
@@ -122,9 +118,25 @@
...
+		if(organ_status != ORGAN_BRUISED)
+			organ_status = ORGAN_BRUISED
+		return TRUE
+	if(organ_status != ORGAN_HEALTHY)
+		organ_status = ORGAN_HEALTHY
+		return TRUE
 
 /datum/internal_organ/proc/heal_organ_damage(amount)
 	damage = max(damage - amount, 0)
+	set_organ_status()
```
QualityVan, line 139: Do this when taking damage too

code/modules/organs/organ_internal.dm
```diff
@@ -204,20 +224,16 @@
...
-			owner.emote("me", 1, "gasps for air!")
-			owner.Losebreath(4)
-	else if(is_broken())
-		if(prob(30))
-			owner.emote("me", 1, "coughs up blood!")
-			owner.drip(10)
-		if(prob(50))
-			owner.emote("me", 1, "gasps for air!")
-			owner.Losebreath(10)
+	if((organ_status == ORGAN_BRUISED && prob(15)) || (organ_status == ORGAN_BROKEN && prob(50)))
```
QualityVan, line 227: prob(50) on life is gonna be real spammy when broken lungs don't actually kill you, maybe 5 and 20?

code/modules/organs/organ_internal.dm
```diff
@@ -122,9 +118,25 @@
...
+		if(organ_status != ORGAN_BROKEN)
+			organ_status = ORGAN_BROKEN
+		return TRUE
+	if(damage > min_bruised_damage)
+		if(organ_status != ORGAN_BRUISED)
+			organ_status = ORGAN_BRUISED
+		return TRUE
+	if(organ_status != ORGAN_HEALTHY)
+		organ_status = ORGAN_HEALTHY
+		return TRUE
```
QualityVan, line 135:
```suggestion
	if(damage > min_broken_damage)
		if(organ_status != ORGAN_BROKEN)
			organ_status = ORGAN_BROKEN
			return TRUE
		return FALSE
	if(damage > min_bruised_damage)
		if(organ_status != ORGAN_BRUISED)
			organ_status = ORGAN_BRUISED
			return TRUE
		return FALSE
	if(organ_status != ORGAN_HEALTHY)
		organ_status = ORGAN_HEALTHY
		return TRUE
```
Probably cleaner, having actually read how you're using it.

code/modules/organs/organ_internal.dm
BraveMole: oh i forgot

BraveMole: uh, but i was doing it haha, i didn't forget

Lumipharon:
It's so easy to fix though? Why are we essentially removing internal damage when delimbing is far worse and not field fixable?

breadhunt:
Combat debuffs like in this PR might actually be a better way to go. The effects of organ damage atm aren't really ever seen by xenos, but they're painful as fuck to marines. This is pretty cool but some more effects might be good, because this is making organ damage almost invalid

Arctrooper8:
I can't wait to take one peri pill and some synaptizine and completely ignore the supposedly "debilitating permanent" damage.

Arctrooper8:
Idea:
Remove organs giving you damage (liver tox, lung oxyloss, etc etc)
Remove normal peri (Since organ damage no longer kills you having a magic chem that allows you to ignore organ damage is silly, you should have to use Peri+ which is limited or go shipside for surgery)

BraveMole:
> Idea: Remove organs giving you damage (liver tox, lung oxyloss, etc etc) Remove normal peri (Since organ damage no longer kills you having a magic chem that allows you to ignore organ damage is silly, you should have to use Peri+ which is limited or go shipside for surgery)

Yeah that makes sense

goose4429:
The penalties should be harsh, because I can't see why removing a thing which was significantly harming 3-20 marines every round is not powercreep. Marine having half as much max stamina vs marine literally dying from hypoxia is not the same. Stam is not even a problem, while for organs you _have_ to go to doc after some time.

Yes, peri can counter it, but let's not assume every marine out there gets his peri.

ilikesaying:
Liver damage could reduce chemical effects, increase metabolism without the benefits or reduce overdose thresholds

Al-1ce:
This is a pretty harsh nerf to organ damage. Getting stomped on by crusher should do much more than make you a little tired. I digress.

GrumgusMungus:
> This is a pretty harsh nerf to organ damage. Getting stomped on by crusher should do much more than make you a little tired. I digress.

I agree, there should be a greater penalty to getting your heart smacked than getting a little tired . This change seems to heavily favor marines without giving a decent compromise to Xenos and further simplifies the unga gameplay.

TiviPlus (approved):
Conflicts

Al-1ce:
> I agree, there should be a greater penalty to getting your heart smacked than getting a little tired . This change seems to heavily favor marines without giving a decent compromise to Xenos and further simplifies the unga gameplay.

Through testing, I oughta' agree.

goose4429:
If this passes, I'm giving benos +5 attack so that marines have no breathing problems, and no hands.

breadhunt:
I got organ damage and completely ignored it for 4 rounds straight today. FYI, synap still means infinite stam with this PR

Heart/lung damage should probably be a -10 max HP or a slowdown or something in addition to stam damage. Nonlethal but actually has an effect.
 As-is, organ damage doesn't do anything. Which is really really bad for xenos

Al-1ce:
>As-is in this PR, organ damage doesn't do anything. Which is really really bad for xenos

That definitely seems to be the intention.

breadhunt:
Afaik intention is to move away from "invisible" organ damage/etc. deaths (confusing for new players, not really fun) and shift towards combat debuffs (stam damage, skills lowered..)

If you have any combat debuff ideas to make organ damage more disabling without making it lethal, it'd probably be good to dump them

Dolth:
"I think dying from an invisible wound suck. Losing blood from lungs and heart damage sucks. Fainting because of oxyloss because your heart is damaged from one warrior punch sucks."

If you think the issue is dying from an invisible wound, then make it more visible when you have organ damage...
No need to buff marines further when you already know where the problem is.

BraveMole:
> I got organ damage and completely ignored it for 4 rounds straight today. FYI, synap still means infinite stam with this PR
> 
> Heart/lung damage should probably be a -10 max HP or a slowdown or something in addition to stam damage. Nonlethal but actually has an effect. As-is in this PR, organ damage doesn't do anything. Which is really really bad for xenos

Took your idea

code/modules/organs/organ_internal.dm
```diff
@@ -264,11 +281,10 @@
...
 		var/filter_effect = 3
-		if(!owner.reagents.get_reagent_amount(/datum/reagent/medicine/peridaxon) >= 0.05)
-			if(is_bruised())
-				filter_effect -= 1
-			if(is_broken())
-				filter_effect -= 2
+		if(organ_status == ORGAN_BRUISED)
+			filter_effect -= 1
+		if(organ_status == ORGAN_BROKEN)
+			filter_effect -= 2
```
QualityVan, line 287:
```suggestion
		filter_effect -= organ_status
```

Al-1ce:
>Lungs damage will decrease your stamina regen and add a slowdown


This makes sense. 

>Heart damage will decrease your maximum stamina and your maximum health


I _want_ to agree with this, but it feels lackluster a little. Maximum health is obviously a big deal - Russian Red makes that plainly clear, but it feels almost miniscule when compared to the oxyloss from before. Which I never experienced passing out from, seldom.

While this PR pretty much did a complete backflip, this still feels like organ damage being overall lesser than what it previously was. I know that is obviously the intent, to avoid cancer - but still it feels like this wouldn't affect ungaball all that much.

If the variables and slowdown are harsh enough I may be completely wrong. I'm looking forwards to a TM to see.

crayonLover09:
The slowdown from lung damage is VERY strong right now. Anecdotally it feels like wearing tyr heavy armor when I have light armor

MiniMeatwad:
Doesn't seem to update organ status when organs are healed? Can't tell, someone had 0 organ damage but a status of "BRUISED" and permanent effects.

MiniMeatwad:
Aside from that, my main complaint with organ damage is it being annoying to find out organ damage. I say just make health scanners tell you organ damage.

BraveMole:
> Aside from that, my main complaint with organ damage is it being annoying to find out organ damage. I say just make health scanners tell you organ damage.

PR is up, need this one to be merged though

goose4429:
So we're finally moved to stage "organ damage is just a damage type". Told you it's the solution.
Speed penalty is a good idea.

code/modules/mob/living/blood.dm
```diff
@@ -29,11 +29,11 @@
 			if(!heart)
```
breadhunt, line 29: FYI you need to remove this part of the code or heart effects aren't removed. How heart damage works is by reducing "effective" blood level. For example, if you've got 100% blood and a bruised heart, you'll get the effects of 70% blood. If you've got 90% blood and a broken heart you'll get the effects of 45% blood. This also drains blood

I don't know if you've axed it in another part of the code or not, but this was happening to me in-game during the TM so I'm pretty sure it's not removed

BraveMole:
> Doesn't seem to update organ status when organs are healed? Can't tell, someone had 0 organ damage but a status of "BRUISED" and permanent effects.

Do you know how was this organ healed ? In local testing it worked fine

QualityVan:
> Do you know how was this organ healed ? In local testing it worked fine

Surgery sets damage directly to 0 rather than routing through heal_organ_damage, and autodocs use organ.rejuvenate which has the same issue. You could also call set_organ_status in their process if you want more of a catchall, but that's more expensive.

PR #10318, opened 2022-05-29
platform cover fix

Lumipharon:
## About The Pull Request
Makes platform bullet coverage behave correctly. They still have real low coverage though.

Platforms are the structures that split turfs into 'high' and 'low' ground. They stop movement like barricades, but for blocking bullets, the high ground is supposed to provide the advantage.
i.e. shooting at the low ground doesn't stop bullets, while the unga below you has a chance to hit the platform.

Unfortunately, platforms are actually on the 'low ground' turf for spriting reasons, so the code that barricades and other similar objects use needs to be reversed.

A little bit ugly to copy paste the base code with a minor change, but I thought this was better than adding an if check specifically for this one structure, and there isn't really a way to pull that direction hit code out and just call to parent, as there is a significant number of objects that use the relevant ON_BORDER flag.

Please let me know if I've been smooth brain and there is a neater way to do this, or if just adding the if check in the parent proc is actually preferrable.

## Why It's Good For The Game
Platforms block bullets in the correct direction.

## Changelog
:cl:
fix: Having the high ground now correctly gives you the advantage in a firefight.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -678,6 +678,34 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		return
 	bullet_act(proj)
 
+//platforms
+/obj/structure/platform/projectile_hit(obj/projectile/proj, cardinal_move, uncrossing)
+	if(!density && !(obj_flags & PROJ_IGNORE_DENSITY)) //structure is passable
+		return FALSE
+	if(src == proj.original_target) //clicking on the structure itself hits the structure
+		return TRUE
+	if(!throwpass)
+		return TRUE
+	if(proj.distance_travelled <= proj.ammo.barricade_clear_distance)
+		return FALSE
+	. = coverage //Hitchance.
+	if(!(cardinal_move & dir)) //The bullet will only hit if the platform and its movement are facing the same direction, as platforms are on the 'lower' turf, so act in the opposite direction to barricades.
+		if(!uncrossing)
+			proj.uncross_scheduled += src
+		return FALSE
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -678,6 +678,34 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		return
 	bullet_act(proj)
 
+//platforms
+/obj/structure/platform/projectile_hit(obj/projectile/proj, cardinal_move, uncrossing)
```
BraveMole, line 682: that's a straight copy paste of obj projectile hit. Why do we even need that

BraveMole (changes requested):

code/modules/projectiles/projectile.dm
Lumipharon: The difference is in the border direction check. As mentioned in the PR, I can just add the check into the parent proc, but I wasn't certain what was best practice.

Otherwise I didn't see a way to call to parent as the parent explicitly has the inverse check, and too many objects use it to justify splitting that out into their own child procs.

BraveMole: Is the dir inversed on platform ?

BraveMole (changes requested):

code/modules/projectiles/projectile.dm
Lumipharon: What? There's nothing wrong with the dir, it's working as intended. The platform structure represents the separation between a lower and higher level, on the lower level turf.

PR #10322, opened 2022-05-29
Defibrillator get a new mode that can instantly heal a marine to 195 damage. Cost more energy

BraveMole:
## About The Pull Request

Defib gets a new mode, super charge. If you right click instead of left click, you can instantly defib the target.

You need to have medical skill in order to do that, and this will use up 5000 energy.
A normal defib will use 500, and you have a battery of 10k.

You can use your backpack to recharge your defib, but only once since it has a reserve of 10k as well.

## Why It's Good For The Game

Getting more than 200 damage comes down mostly to luck. Corp have no better way of dealing with that than spamming the defib, which can perma you if you are unlucky.

I think being revived or not should only depend on how quickly marines get you, and not if you were unlucky enough to die on a flamed tile or if you took a nade while dead.

If marines managed to get back your corpse, you should be saveable imo.

With this new mode, we keep the normal complexity of reviving, but if someone is too roasted to be resed, the corp can use that.

## Changelog
:cl:
balance: Defibrillator will instantly heal you to 195 damage if you use super charge mode (right click)
balance: Super charge mode consummes 5k energy (out of 10k battery), normal defib uses 500
balance: You can only recharge your defib to full once with your corp backpack (it was 10 times before)
delete: Defib can always be used, no need to open it 
/:cl:

### DIFF
--- a/code/game/objects/items/defibrillator.dm
+++ b/code/game/objects/items/defibrillator.dm
@@ -152,7 +152,7 @@
 	if(!ishuman(H))
 		to_chat(user, span_warning("You can't defibrilate [H]. You don't even know where to put the paddles!"))
 		return
-	if(dcell.charge <= super_charge ? charge_cost * SUPER_CHARGE_MULTIPLIER_COST : charge_cost)
+	if(dcell.charge <= (super_charge ? charge_cost * SUPER_CHARGE_MULTIPLIER_COST : charge_cost))
 		user.visible_message(span_warning("[icon2html(src, viewers(user))] \The [src] buzzes: Internal battery too low. Cannot analyze nor administer shock."))
 		to_chat(user, maybe_message_recharge_hint())
 		return
--- a/code/game/objects/items/defibrillator.dm
+++ b/code/game/objects/items/defibrillator.dm
@@ -1,3 +1,5 @@
+#define SUPER_CHARGE_MULTIPLIER_COST 10
+
 /obj/item/defibrillator
 	name = "emergency defibrillator"
 	desc = "A handheld emergency defibrillator, used to restore fibrillating patients. Can optionally bring people back from the dead."
@@ -10,12 +12,13 @@
 	throwforce = 6
 	w_class = WEIGHT_CLASS_NORMAL
 
-	var/ready = FALSE
 	var/damage_threshold = 8 //This is the maximum non-oxy damage the defibrillator will heal to get a patient above -100, in all categories
-	var/charge_cost = 66 //How much energy is used.
+	var/charge_cost = 500 //How much energy is used.
 	var/obj/item/cell/dcell = null
 	var/datum/effect_system/spark_spread/sparks
 	var/defib_cooldown = 0 //Cooldown for toggling the defib
+	/// If we are attempting a super charged defibbrilation
+	var/super_charge = FALSE
 
 
 /obj/item/defibrillator/suicide_act(mob/user)
@@ -28,7 +31,7 @@
 	sparks = new
 	sparks.set_up(5, 0, src)
 	sparks.attach(src)
-	set_dcell(new /obj/item/cell())
+	set_dcell(new /obj/item/cell/high())
 	update_icon()
 
 
@@ -41,9 +44,7 @@
 
 
 /obj/item/defibrillator/update_icon()
-	icon_state = "defib"
-	if(ready)
-		icon_state += "_out"
+	icon_state = "defib_out"
 	if(dcell?.charge)
 		switch(round(dcell.charge * 100 / dcell.maxcharge))
 			if(67 to INFINITY)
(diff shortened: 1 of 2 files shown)


Lumipharon:
Corps can inaprov and kit a corpse. You can't perma a corpse via defib, but if the heart gets damaged enough they will actually need surgery.

The only way to get someone completely unrevivable is to max out their tox or clone damage.

BraveMole:
> Corps can inaprov and kit a corpse. You can't perma a corpse via defib, but if the heart gets damaged enough they will actually need surgery.
> 
> The only way to get someone completely unrevivable is to max out their tox or clone damage.

they cannot inaprov a corpse, and they can kit only once per limb. So if you have 400, you need to spam defib, which can leads to perma death (unless surgery)

goose4429:
> they cannot inaprov a corpse

They are indeed able to ina the corpse. And RR it, and neura it as well.
Imagine giving work to MO, and not solve every problem via cryo or defib spam.

SzkieletorNutric:
> they cannot inaprov a corpse
Can confirm, Inaprovaline instantly healing a portion of damage somehow works even when the body is dead

BraveMole:
> > they cannot inaprov a corpse
> 
> They are indeed able to ina the corpse. And RR it, and neura it as well. Imagine giving work to MO, and not solve every problem via cryo or defib spam.

then that's a pretty big exploit

goose4429:
I think that if you do both defib and Ina fix, would be cool

BraveMole:
> I think that if you do both defib and Ina fix, would be cool

I mean sure, but ina wouldn't matter with this new defib

Losenis:
I feel that medical has enough mechanical problems to make this understandable. There's some very weird stuff going on that make marines get more damage after death even if it's clearly not intentional, like the time acid sprays would damage corpses or how you can actually damage corpses by punching them normally getting them way past 600 brute. Sometimes corpses on fire would take damage and other times I'd see corpses that were on fire for centuries just have like 20 dmg or so, and random spacetiles would skyrocket the corpse damage to its literal limit. I'd consider this a failsafe to the inevitable bugs that may appear than an actual buff or nerf, since a marine getting 600 dmg after dying at 200 is clearly not intentional and even shipside will struggle to even do anything that isn't deepfrying its heart with defibspam of multiple people.

If marines get over 200 dmg to the point that some call out this affects balance, then something is going extremely wrong. There's already enough things to send a marine shipside that isn't just normal damage numbers.

BraveMole:
> Sometimes corpses on fire would take damage and other times I'd see corpses that were on fire for centuries just have like 20 dmg or so, 
corpse on fire don't take damage, they take damage when on tiles with fire.

goose4429:
> I feel that medical has enough mechanical problems to make this understandable. There's some very weird stuff going on that make marines get more damage after death even if it's clearly not intentional, like the time acid sprays would damage corpses or how you can actually damage corpses by punching them normally getting them way past 600 brute. Sometimes corpses on fire would take damage and other times I'd see corpses that were on fire for centuries just have like 20 dmg or so, and random spacetiles would skyrocket the corpse damage to its literal limit. I'd consider this a failsafe to the inevitable bugs that may appear than an actual buff or nerf, since a marine getting 600 dmg after dying at 200 is clearly not intentional and even shipside will struggle to even do anything that isn't deepfrying its heart with defibspam of multiple people.
> 
> If marines get over 200 dmg to the point that some call out this affects balance, then something is going extremely wrong. There's already enough things to send a marine shipside that isn't just normal damage numbers.

True words, I agree that failsafe value is higher than some possible balance issues in this case

DeltaFire15:
Massive corpsmen buff considering this takes away any pre-defibbing considerations even if command decided to OB a corpse pile.
Most situations with really high damage can be resolved via just kitting + defibbing a few times, whereas if there's massive damage due to some reason (like aforementioned OBs) it can still be fixed by surgery. Having dead bodies in the proximity of combat should influence if you have to be more or less careful around them, besides the fact that otherwise there'd also be zero tradeoff for 'grenade fishing' dead bodies with something like a scoped GL or w/e.
If you are really that much on the fence about people having enough damage to give shipside med a job to do, cap brute / burn damage at something like 400 each instead of just sidestepping revival neccessities.

Losenis:
There's already little to no tradeoff for grenade fishing because the grenade damage is all spread out evenly throughout the body with little threat of fracs over concentrated damage, even if the guy receives 100 brute damage off grenades.

And for every grenade fishing attempt there's another that punched the body further into the maze. This isn't a damage issue, this is a grenade issue.

Losenis:
Not to mention, this wouldn't change much about grenade fishing since due to the aforementioned damage spread that is also both burn and brute, you'd be looking at trauma and health kits reducing the damage of every bodypart twice. Defib damage wouldn't come into play unless something else gave the guy absurd amounts of damage, at which point it also stops being an issue about grenades and instead something else that is most likely not supposed to be doing that to an already dead marine.

Ragecrabbo:
> Getting more than 200 damage comes down mostly to luck. Corp have no better way of dealing with that than spamming the defib, which can perma you if you are unlucky.

Dont sutures exist to deal with this problem? Why not just buff them instead?

goose4429:
New mode is at cool at least

ilikesaying:
> There's already little to no tradeoff for grenade fishing because the grenade damage is all spread out evenly throughout the body with little threat of fracs over concentrated damage, even if the guy receives 100 brute damage off grenades.

pretty sure grenade explosions deal damage to one limb

Lumipharon:
I'm not opposed to this, but this makes it harder to attrition marines due to them being faster to revive, and making it pointless to medivac unless they're missing limbs.

code/game/objects/items/defibrillator.dm
```diff
@@ -220,10 +202,6 @@
 		user.visible_message(span_warning("[icon2html(src, viewers(user))] \The [src] buzzes: Defibrillation failed: Paddles registering >100,000 ohms, Possible cause: Suit or Armor interferring."))
 		return
 
-	var/datum/internal_organ/heart/heart = H.internal_organs_by_name["heart"]
-	if(!issynth(H) && !isrobot(H) && heart && prob(25))
-		heart.take_damage(5) //Allow the defibrilator to possibly worsen heart damage. Still rare enough to just be the "clone damage" of the defib
-
```
QualityVan, line 226: Not in the PR desc or changelog at all. Also don't think it's a needed change given what the pr is adding.

BraveMole: ah, forgot about this one.

Yeah, I shoudl probably remove

goose4429: BM subtly buffing marines

bowlofcereal:
MO and CMO gonna be alt tabbing out to watch youtube videos as their main gameplay here soon

Autisem:
Ye this does make the already pretty boring loop of playing MO/CMO that much worse. kinda silly that you can just heal them like that. rechargers exist and I believe the corpsmanbackpack includes are powercell that automaticly recharges defibs so you can staple that onto it aswel(this powercell also is swappable for extra comedic effect)

PR #10323, opened 2022-05-29
Brain damage reduces your skills rating, nothing else

BraveMole:
## About The Pull Request

If you have a bruised brain, you lose 1 rating to all your skills

## Why It's Good For The Game

Current effects of brain damage are random and cancer. This is still an impediment but might be more supportable by players

## Changelog
:cl:
balance:Brain damage reduces your skills rating, nothing else
/:cl:


goose4429:
Can you maybe give an idea of what's going to happen now?
**Before** - drop item randomly, each tick have 3% to drop or smth.

**Now** what? And what skills does this affect?
+2 recoil on weapons and cpsmen suddenly forget how to use hypo?
What happens if marine goes into negative medical, they forget to breathe?
SGs can't fire the SG?

BraveMole:
You cannot go below 0, and yeah, for the most part, it will affect recoil, corps, sg, etc...

code/datums/skills.dm
```diff
@@ -62,6 +62,8 @@ engineer = 0, construction = 0, leadership = 0, medical = 0, surgery = 0, pilot
 	var/police = SKILL_POLICE_DEFAULT
 	var/powerloader = SKILL_POWERLOADER_DEFAULT
 	var/large_vehicle = SKILL_LARGE_VEHICLE_DEFAULT
+	/// Malus or bonuses to all skills
+	var/global_skill_modificator = 0
```
QualityVan, line 66: Humans with the same set of skills share the same skillset datum, this'll cause problems. My plan on this was to make a helper like modifyRating below that took a single value and applied it to all skills, then call that through the brain's update.

goose4429:
> You cannot go below 0, and yeah, for the most part, it will affect recoil, corps, sg, etc...

So SG will literally forget how to use SG gun? Very based, I like it

Pariah919:
> > You cannot go below 0, and yeah, for the most part, it will affect recoil, corps, sg, etc...
> 
> So SG will literally forget how to use SG gun? Very based, I like it
> 
> Edit: skills cant go past 0, and you need 0 to use SG. Marines have -4 by default.

SGs have 0 skill already

Al-1ce:
I could see this implimented _alongside_ the current debuffs, but as a standalone effect it seems to be much less dehibilitating in a general sense. Sure, you might not be able to shoot as well - but most other functions aside will be functioning if you're not disabled/injured in some other sense. Brain damage is supposed to be severe, not a mild annoyance you can shrug off.

FosterGrime-Heymynameslouie:
being brain damaged actually sounds and act like being brain damaged?
based

TiviPlus:
finish whene

code/datums/skills.dm
```diff
@@ -158,10 +178,10 @@ engineer, construction, leadership, medical, surgery, pilot, police, powerloader
 #undef SKILLSID
 
 /datum/skills/proc/getPercent(rating, max_rating)
-	return CLAMP01(vars[rating] * 100 / max_rating * 0.01)
+	return CLAMP01(max(vars[rating], 0) * 100 / max_rating * 0.01)
 
 /datum/skills/proc/getRating(rating)
-	return vars[rating]
+	return max(vars[rating], 0)
```
QualityVan, line 184:
```suggestion
	return vars[rating]
```
Turns out negative skill ratings exist and are relied on for a couple things already, smartgun penalties being the most prominent.

Losenis:
> > > You cannot go below 0, and yeah, for the most part, it will affect recoil, corps, sg, etc...
> > 
> > 
> > So SG will literally forget how to use SG gun? Very based, I like it
> > Edit: skills cant go past 0, and you need 0 to use SG. Marines have -4 by default.
> 
> SGs have 0 skill already

image

Pariah919 (approved):
Brain damage was awful cancer before and this is cool idea.

QualityVan:
If you want stronger feedback about how the new penalties feel, kill alky or make it a suppressant instead of a cure.
Is there any direct feedback for someone with brain damage, aside from everything being slightly slower?

PR #10332, opened 2022-05-29
Boiler globs open resin doors

Wayland-Smithy:
## About The Pull Request

Title.

## Why It's Good For The Game

Reduces chances for builder castes to inadvertently troll boilers.
Makes approaching closed resin doors riskier.

## Changelog
:cl:
balance: Boiler globs now open resin doors in their way.
/:cl:

### DIFF
--- a/code/game/objects/structures/xeno.dm
+++ b/code/game/objects/structures/xeno.dm
@@ -169,6 +169,10 @@
 		Open()
 		return TRUE
 
+/obj/structure/mineral_door/resin/projectile_hit(obj/projectile/proj, cardinal_move, uncrossing)
+	if(!state && ispath(proj.ammo.type, /datum/ammo/xeno/boiler_gas))
+		Open()
+	return ..()
 
 /obj/structure/mineral_door/resin/attack_larva(mob/living/carbon/xenomorph/larva/M)
 	var/turf/cur_loc = M.loc


Lumipharon:
This is very silly, it lets you fire unseen, which goes against boiler design. It makes boilers unstoppable unless you're sniping every resin door in a 30 tile range or whatever.

Wayland-Smithy:
> it makes boilers unstoppable unless you're sniping every resin door

Idk doors are pretty weak and if boilers are so close as to only be behind a few they run the risk of being run down. If they are far away behind many then that limits the number of possible angles to way below "every resin door in a 30 tile range".

That said your concerns are valid, but I believe it is worth a testmerge at least. This is far less impactful than infinite stam for instance and we had that for like a week.

Lumipharon:
> > it makes boilers unstoppable unless you're sniping every resin door
> 
> Idk doors are pretty weak and if boilers are so close as to only be behind a few they run the risk of being run down. If they are far away behind many then that limits the number of possible angles to way below "every resin door in a 30 tile range".
> 
> That said your concerns are valid, but I believe it is worth a testmerge at least. This is far less impactful than infinite stam for instance and we had that for like a week.

You only need one door, and how strong they are is irrelevant, as it's about LOS. You have a wall and a door, you can visually see the path your projectile will take so you just need to stand in the right spot for it not to block you. You don't need to worry about having 5 doors deep because marines don't know you're there, so any sniper isn't able to know that they need to shoot one specific door to get to you.

This makes being a boiler 100% safe, and is pretty cheese.

ilikesaying:
With the merging of the Laser Sniper iff pr, Boilers are under constant threat of instantly getting tagged by a sniper, due to the gun's iff meaning it passes through friendlies and the hitscan meaning the boiler will always get hit the moment they get into the field of fire.
Source : my experience with the Laser Sniper post iff merge

Wayland-Smithy:
> You don't need to worry about having 5 doors deep because marines don't know you're there

Won't they find out pretty quickly once you shoot at them?

One machinegun can clear entire lanes of doors in a small fraction of the time it takes to build them. I just don't think it is that cheesy when we are seeing mazes being cleared at all time record speeds.

Lumipharon:
> > You don't need to worry about having 5 doors deep because marines don't know you're there
> 
> Won't they find out pretty quickly once you shoot at them?
> 
> One machinegun can clear entire lanes of doors in a small fraction of the time it takes to build them. I just don't think it is that cheesy when we are seeing mazes being cleared at all time record speeds.

Then you just reposition, which boilers normally do anyway. You can also do this from out of normal mg LOS, so they'd have to blind fire.

> > This makes being a boiler almost 100% safe, and is pretty cheese.
> 
> That is just straight up false. As the fighting moves around there is a 100% chance boilers will not be able to avoid being out in the open to be effective at some point.

Yes, this is really only applicable for seiges, or fairly static fights, but that's a pretty significant part of what boilers do.

Naaanii:
> > > You don't need to worry about having 5 doors deep because marines don't know you're there
> > 
> > 
> > Won't they find out pretty quickly once you shoot at them?
> > One machinegun can clear entire lanes of doors in a small fraction of the time it takes to build them. I just don't think it is that cheesy when we are seeing mazes being cleared at all time record speeds.
> 
> Then you just reposition, which boilers normally do anyway. You can also do this from out of normal mg LOS, so they'd have to blind fire.
> 
> > > This makes being a boiler almost 100% safe, and is pretty cheese.
> > 
> > 
> > That is just straight up false. As the fighting moves around there is a 100% chance boilers will not be able to avoid being out in the open to be effective at some point.
> 
> Yes, this is really only applicable for seiges, or fairly static fights, but that's a pretty significant part of what boilers do.

Hard to reposition when fighting is in one part of caves and Xenos are holding a singular choke, not to mention doors are weak and don't last long enough to do much. Door spam is unviable to do on the go because of the higher build time and door spam means a flamer or even and LMG will just annihilate the lane, giving vision of boilers which pretty much almost nullifies this change either way.

Autisem:
anti martial rifles can also bypass defenders defending the boiler. I welcome this change with only 1-2 snipers it becomes near impossible to entrench and fire without dying. Makes it less of a non-choice that building casts should not build because 1-2 boilers become near useless in choke due to bunch of walls that are easily cleared anyhow

BraveMole:
image

PR #10333, opened 2022-05-29
Remove chance for boiler globs to be inaccurate

Wayland-Smithy:
## About The Pull Request

Title.

## Why It's Good For The Game

Sniper projectiles not going where you click is bad. `accuracy_var_high = 10` is the same as shotgun ammo.

## Changelog
:cl:
balance: Removed chance for boiler globs to be inaccurate.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -2291,7 +2291,6 @@ datum/ammo/bullet/revolver/tp44
 	flags_ammo_behavior = AMMO_XENO|AMMO_SKIPS_ALIENS|AMMO_EXPLOSIVE
 	var/danger_message = span_danger("A glob of acid lands with a splat and explodes into noxious fumes!")
 	armor_type = "bio"
-	accuracy_var_high = 10
 	max_range = 30
 	damage = 50
 	damage_type = STAMINA


BraveMole (approved):

Lumipharon:
This change makes boiler globs LESS accurate, you do realise?

The base accuracy of a projectile is calculated as:
accuracy   *= rand(95 - ammo.accuracy_var_low, 105 + ammo.accuracy_var_high) * 0.01

So by changing it's high variance from 10 to the xeno default of 3, the accuracy range of globs decreases.

And are you actually trying to change accuracy, or scatter? Your PR wording sounds like you're talking about scatter, which is 0.

Wayland-Smithy:
Good point. I looked at scatter, but like you said it appears to be zero. In that case for accuracy math they should have 10 min then which makes sense with the val for shotty so you don't ever miss. Will have to debug some more to be sure if scatter is actually 0 *all the time*.

Lumipharon:
> Good point. I looked at scatter, but like you said it appears to be zero. In that case for accuracy math they should have 10 min then which makes sense with the val for shotty so you don't ever miss. 

What do you mean by this? There's a hardcoded accuracy variance of +-5%, thevar high/var low is just to change this variance

>Will have to debug some more to be sure if scatter is actually 0 _all the time_.
Afaik nothing should alter scatter for xenos, they don't use guns so don't have movement penalties, and boilers have to hold still anyway.

What is actually happening that shouldn't?

Wayland-Smithy:
> What do you mean by this? There's a hardcoded accuracy variance of +-5%, thevar high/var low is just to change this variance

So really it just needs swapping from high to low to negate the variance.

> What is actually happening that shouldn't?

I thought it was the old scatter RNG rearing its head, but that *shouldn't* be the case. Will probably automate testing and update here if it can be demonstrated. For now it just looks like globs have a chance to miss (fly past) a target mob in their path.

Lumipharon:
> > What do you mean by this? There's a hardcoded accuracy variance of +-5%, thevar high/var low is just to change this variance
> 
> So really it just needs swapping from high to low to negate the variance.
> 
> > What is actually happening that shouldn't?
> 
> I thought it was the old scatter RNG rearing its head, but that _shouldn't_ be the case. Will probably automate testing and update here if it can be demonstrated. For now it just looks like globs have a chance to miss (fly past) a target mob in their path.

Yes, scatter would mean it literally flies off in a different direction. If you're talking about flying past a mob, then that's accuracy.

If you want to remove variance, make the values -5. But that's not the actual issue. Xeno projectiles have an accurate range of 15. Outside of accurate range your accuracy starts dropping off.
Xenos have a base accuracy of like 125% hit chance, and each tile past 15 drops it by 5. 
So at 20 tiles you have 100% accuracy,
25 is 75%
And at max range of 30, hit chance is 50%.

So what you need to change is to increase the accurate range to like 20 or something.

Optionally, you could increase the accuracy var on the ammo type to increase the base accuracy. Anything over 100 acc decreases the probability of cover blocking the projectile. Cades have a coverage of 128, so pretty much any increase to glob accuracy would give it a (small) chance to bypass cades.
I only suggest this because this was supposedly always a thing, but due to how the code works, wasn't actually possible.

Issue #10060, opened 2022-04-24
CAS broken


Nirvan4:
### Description

 Strafe direction changing on its own, flare minirockets and minirockets not working. 
 
 Strafe direction changes on its own when you move around with WASD.
Flare minirockets and fatty doesnt work at all, when you fire them they dont have any hit prediction or effects, they just disappear.

### Test Merges

[#9772: 'TGUI updates' by TiviPlus at commit 16b87e0607](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9772)
[#9977: 'Health kits chain apply' by QualityVan at commit 5afedc795b](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9977)
[#10008: 'Boxing+' by breadhunt at commit beb71c8b35](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10008)
[#10031: 'Refactors gun icon/item code' by Lumipharon at commit 622e614bac](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10031)
[#9979: 'Add wraith portal, remove hyperposition, remove phasing' by BraveMole at commit 0b2b974fee](https://github.com/tgstation/TerraGov-Marine-Corps//pull/9979)

### Reproduction Steps

1. Get on CAS
2. Move around WASD
3. Watch your strafe direction changing on its own
4. Try to use flare minirockets
5. Try to use fatty
6. Fail 4 and 5.

novaepee:
CAS changing direction due to WASD is due to this PR: #9851

Nirvan4:
> CAS changing direction due to WASD is due to this PR: #9851

you are right e

PR #10345, opened 2022-05-30
More hotkeys for CAS

breadhunt:
## About The Pull Request

Keys 1, 2, 3, 4, 5 select weapons
6 deselects weapon
South and East keybindings were the wrong way around, swaps them round to be correct
(so before down would = east, right would = south - that's swapped)

Closes https://github.com/tgstation/TerraGov-Marine-Corps/issues/10060 along with the lighting fix

You still need to focus on the window for keybindings (good), but firing a weapon un-focuses the weapon. I don't know how to fix that, so you'll still have to click on the window every time you fire a weapon. If there's a way to fix that please ping

## Why It's Good For The Game

More QoL good
Bugfixes good
This has been wanted for a while now

## Changelog
:cl:
qol: 123456 keybind weapon swapping for CAS
fix: Swaps south/east keybinds for CAS to where they should be
/:cl:


Nirvan4:
Now that youre on it: flare minirockets and FATTY doesnt work currently, they both vanish when fired

breadhunt:
> Now that youre on it: flare minirockets and FATTY doesnt work currently, they both vanish when fired.
> 
> If ya willing check that

I have a PR up for flares. Fatty works completely fine, it just clears walls and I'm going to buff it in the near future

PR #4436, opened 2020-07-07
Restats the slash damage of nearly every caste in the game.

Pariah919:
## About The Pull Request
Redoes almost every castes damage to fit the homogenized armor in #4434. 
Every caste has 2 tiers of damage they have, 1 at young and 1 at elder.
Most of this is done assuming 50% armor.
tm with #4434 

## Why It's Good For The Game
Better balance.

## Changelog
:cl:
balance: Every caste has their slash damage redone to fit the new homogenized armor. Slash damage now has two tiers, one at young and one at elder.
/:cl:

PR #10351, opened 2022-05-31
Allow many non-combat essential objects to be slashed at all or faster by xenos

Wayland-Smithy:
## About The Pull Request

See changelog.

Part 1 of X.

## Why It's Good For The Game

Closets (including coffins and toggleable crates) probably should not be immune from slash destruction, and opening one should not trigger attack cooldown.
Wooden crates, barrels, cases, plastic flaps, beer kegs, ore boxes hydro trays etc.. take a long time to destroy especially with acid which is even slower than slashing which the majority of overlooked movement blocking objects lack the resistance flag for.

## Changelog
:cl:
balance: Reduced the max_integrity of: coffins, large crates (the kind that must be broken or opened with a crowbar), cases, barrels, ore boxes, plastic flaps, hydro trays, mop buckets, janitorial carts, filing cabinets, and reagent dispensers.
fix: Added the xeno damageable flag to: janitorial carts, filing cabinets and reagent dispensers (beer keg, boozeomat, etc..)
fix: Removed useless train vehicle attack_alien override preventing damage. 
fix: Fixed xenos not being able to open wrapped large parcels i.e. wrapped lockers.
fix: Opening gracefully, not smashing, closets does not put xeno attack on cooldown.
fix: Fixed closets not taking damage from slashes even though the animation played and informed the xeno they "smashed" the object.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets.dm
@@ -200,8 +200,10 @@
 		else
 			X.visible_message(span_danger("\The [X] smashes \the [src]!"), \
 			span_danger("We smash \the [src]!"), null, 5)
+			take_damage(damage_amount, damage_type, damage_flag, effects, null, armor_penetration)
 		SEND_SIGNAL(X, COMSIG_XENOMORPH_ATTACK_CLOSET)
 	else if(!opened)
+		X.changeNext_move(0) // opening an unlocked closet does not trigger attack cooldown
 		return attack_hand(X)
 
 /obj/structure/closet/attackby(obj/item/I, mob/user, params)
--- a/code/game/objects/structures/crates_lockers/closets/coffin.dm
+++ b/code/game/objects/structures/crates_lockers/closets/coffin.dm
@@ -4,6 +4,7 @@
 	icon_state = "coffin"
 	icon_closed = "coffin"
 	icon_opened = "coffin_open"
+	max_integrity = 40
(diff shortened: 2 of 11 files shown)


Lumipharon:
Don't nerf integrity into the ground, it's especially relevant for shooting. 40 integ is like 2 bullets/slashes/spits.

Wayland-Smithy:
~2 slashes is the goal. Keep in mind these are mostly for fluff and have not been touched for a long time while slash damage has only been going down.

Wayland-Smithy:
> When has slash damage changed?

#4436 #6562 I'm sure there are others, and you are free to dive the blames or just ask MJP I bet, these are just the ones I remember. 

I still think there is some confusion here about what exactly is being reduced to 2/3 slashes to break. Practically nothing shipside as the useful "toggleable crates" are children of closets whose integrity remains unchanged.

Hydro trays in particular give the same effect as unwired cades and should have 50 more integrity than metal cades, and as a side effect take forever to clear out. Same goes for the other objects effected. Please pick something else to take issue with and stop shitting on what is really a QOL change.

Naaanii:
Is janitor cart and mop bucket included in this too? Sometimes run into them on Prison Station.

Wayland-Smithy:
mop bucket could already be slashed, but was 500 integ, janitor cart needed both treatments

that is the last addition to this PR
more will have to wait for part 2

PR #10354, opened 2022-05-31
Nerf energy sniper's damage

novaepee:
## About The Pull Request

Nerf energy sniper from 70 damage to 60 damage.

## Why It's Good For The Game

Energy sniper is a bit too good. It has a faster firing rate than the SR-127 since the SR-127 has a slower firing rate on top of aim mode delay. Xenomorphs players are complaining, so to appease to balance before energy sniper gets destroyed further by nerf, I will nerf it.

Use SR-127 for damage, use energy sniper for faster firing rate and hitscan.

## Changelog
:cl:
balance: energy sniper damage from 70 to 60
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1793,7 +1793,7 @@ datum/ammo/bullet/revolver/tp44
 /datum/ammo/energy/lasgun/marine/sniper
 	name = "sniper laser bolt"
 	hud_state = "laser_sniper"
-	damage = 70
+	damage = 60
 	penetration = 30
 	flags_ammo_behavior = AMMO_ENERGY|AMMO_IFF|AMMO_SUNDERING|AMMO_HITSCAN|AMMO_SNIPER
 	max_range = 40


Wayland-Smithy:
How does the cabal feel about giving it `accurate_range_min`? Been seeing this frontlined a lot lately despite the recent changes that upped the damage and gave IFF for the purpose of staying in the back. What actually happened just like when aim mode was added is now there is zero risk to marine groups brawling with it, missing, and hitting friendlies that is if the lock-on cheese is not fixed/used removing the need for click accuracy altogether.

BraveMole:
> How does the cabal feel about giving it `accurate_range_min`? Been seeing this frontlined a lot lately despite the recent changes that upped the damage and gave IFF for the purpose of staying in the back. What actually happened just like when aim mode was added is now there is zero risk to marine groups brawling with it, missing, and hitting friendlies that is if the lock-on cheese is not fixed/used removing the need for click accuracy altogether.

So you can use it only as a sniper rifle, and not a cqc weapon ? Seems cool

Lumipharon:
Whoops, does it not already have a min accurate range? If so that's an oversight.

Arctrooper8:
> Whoops, does it not already have a min accurate range? If so that's an oversight.

I thought it had that what with it being a sniper rifle and all. If not we should add it.

Lumipharon:
> > Whoops, does it not already have a min accurate range? If so that's an oversight.
> 
> I thought it had that what with it being a sniper rifle and all. If not we should add it.

Yeah it doesn't I'll PR it in a bit. Damage and or fire rate still needs a serious nerf though, IFF hitscan roundstart weapon with big DPS is not particularly balanced.

MiniMeatwad:
Make the damage over time equivalent to the other snipers.

lKiseki:
> Make the damage over time equivalent to the other snipers. Also make it absolutely terrible in close range, and even worse akimbo.

60 damage feels like too little, 50 might be a bit better so that it doesn't outdo every other sniper in the game.

MiniMeatwad:
> 60 damage feels like too little, 50 might be a bit better so that it doesn't outdo every other sniper in the game.

Makes sense, though this gun works really well with SG minigun just because SG minigun is also really strong. Somehow in balance we added a weapon that gives xenos literally 100 sunder and a weapon that deletes xenos in 2 shots if they have no armor.

Autisem:
Yes please, getting rushed by a sniper with histcan is not that fun

novaepee:
Too much energy sniper PRs trying to nerf it.

At least this PR got the ball rolling.

PR #10356, opened 2022-06-01
Disables SADAR from firing until wielded

lKiseki:
## About The Pull Request
Per title. SADAR can no longer be fired until wielded.

## Why It's Good For The Game
People are insta PBing out of knockdown/stuns with SADAR AP. Yet another example of rocket tag gameplay with no viable or enjoyable response.

## Changelog
:cl: Lewdcifer
balance: SADAR is unable to fire until wielded.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -601,7 +601,7 @@ Note that this means that snipers will have a slowdown of 3, due to the scope
 	flags_equip_slot = NONE
 	w_class = WEIGHT_CLASS_HUGE
 	force = 15
-	wield_delay = 20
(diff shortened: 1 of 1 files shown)


Lumipharon:
I would probably make it full wield fire only but leave the wield time. You can always touch that later but this is a double whammy nerf on top of the loss of miniscope

lKiseki:
> I would probably make it full wield fire only but leave the wield time. You can always touch that later but this is a double whammy nerf on top of the loss of miniscope

I checked the wield delay and it didn't amount to much, hence the increased delay. I'll wait and see what maintainers say, regardless.

TiviPlus:
dont change wield time

PR #10357, opened 2022-06-01
Nerfs SG minigun sunder

lKiseki:
## About The Pull Request
Per title. Sundering value reduced from 2 > 1.

## Why It's Good For The Game
SG Miniguns are sundering really quickly, faster than Hivelords can treat entire hives worth of sunder.

## Changelog
:cl: Lewdcifer
balance: SG minigun sunder 2 > 1.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1120,7 +1120,7 @@ datum/ammo/bullet/revolver/tp44
 	accurate_range = 12
 	damage = 10
 	penetration = 15
-	sundering = 2
+	sundering = 1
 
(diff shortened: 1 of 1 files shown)


lKiseki:
If maintainers prefer another alternative, we could also increase the sunder healed by xeno healing abilities, since that's less of a bandaid than this.

Pariah919:
> If maintainers prefer another alternative, we could also increase the sunder healed by xeno healing abilities, since that's less of a bandaid than this.

Does not really fix anything considering it just makes hivelord more required to deal with sunder as a mechanic, just up the rate sunder is healed at when resting drastically and up sunder damage.

PR #10366, opened 2022-06-01
Shifts projectile flamer to deal damage on hit

QualityVan:
## About The Pull Request
Previously all of the damage was dealt through igniting the mob's turf, which led to certain weirdness like not actually damaging moving mobs and AMMO_INCENDIARY code not triggering. This shouldn't change the effective damage in any way, just clean out niche bugs.
Adds an incendiary_strength var to ammo datums to adjust how hard AMMO_INCENDIARY actually ignites stuff.
Also fixes projectile flamer not being able to clear out vines or snow.

## Why It's Good For The Game
Bugfixes

## Changelog
:cl:
balance: Flamer's extended nozzle can remove vines and snow, along with other sources of ground fire.
fix: Flamer projectiles should more reliably ignite on hit.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -59,6 +59,8 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	var/deflagrate_multiplier = 1
 	///Flat damage caused if fire_burst is triggered by deflagrate
 	var/fire_burst_damage = 10
+	///Base fire stacks added on hit if the projectile has AMMO_INCENDIARY
+	var/incendiary_strength = 10
 
 /datum/ammo/proc/do_at_max_range(obj/projectile/proj)
 	return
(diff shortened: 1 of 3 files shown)


code/modules/projectiles/guns/flamer.dm
```diff
@@ -447,6 +437,15 @@
...
 
+	for(var/obj/structure/jungle/vines/vines in src)
+		QDEL_NULL(vines)
+
+/turf/open/floor/plating/ground/snow/ignite(fire_lvl, burn_lvl, f_color, fire_stacks = 0, fire_damage = 0)
+	if(slayer > 0)
+		slayer -= 1
+		update_icon(1, 0)
+	return ..()
+
```
DeltaFire15, line 448: This does not affect only flamers, but anything that uses ignite(), which is a bunch of other things which are not flamers. So, less a fix and more a balance change given how much this touches.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2545,19 +2547,21 @@ datum/ammo/bullet/revolver/tp44
...
 	max_range = 7
-	damage = 0
+	damage = 31
+	damage_falloff = 0
+	incendiary_strength = 30 //Firestacks cap at 20, but that's after armor.
 	bullet_color = LIGHT_COLOR_FIRE
 	var/fire_color = "red"
-	var/burnlevel = 31
 	var/burntime = 17
+	var/burnlevel = 31
```
DeltaFire15, line 2559: You are doubling impact damage of projectile flamers, aswell as fire stack application. Flamer ammo's fire being created on a tile due to e.g. a projectile flamer's flame hitting a target already iterates through any mobs on it and triggers flamer_fire_act on them, which deals damage and ignites targets. So, there is no such thing as it "relying on the target to be moving", if you actually hit it.

code/modules/projectiles/guns/flamer.dm
QualityVan: You're right and I overlooked that, but I'm also completely fine with fire sources generally killing vines and snow. I'll toss it in the changelog.

code/modules/projectiles/ammo_datums.dm
Lumipharon: Turfs igniting mobs seem to be processed every tick or something. This was something I noticed when I made the tx54. I could ignite the turf I was standing on, and if I walked off it quickly enough, I wouldn't ignite.

That said, people also see this sort of issue due to the fact that movement is instant, but the visual effect of your sprite moving from one turf to another is not. So it can look like you're hitting a turf with a mob on it, but they've actually already moved.

PR #10367, opened 2022-06-01
Ballistic goggles protect against flashes

QualityVan:
## About The Pull Request
As title. They give a three tile tint around the edges of the screen, too. You'll still hurt your eyes if you try welding with them on. Also removed equipping them to your mask slot.

## Why It's Good For The Game
An option for people dedicated to explosives. I'd say 'at the cost of your eye slot', but that isn't much of a cost.

## Changelog
:cl:
add: Ballistic goggles protect against flashing lights, but you can't wear them as masks any more (why could you?). Not strong enough to weld with.
/:cl:

### DIFF
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -137,17 +137,21 @@
 
 /obj/item/clothing/glasses/mgoggles
 	name = "marine ballistic goggles"
-	desc = "Standard issue TGMC goggles. Mostly used to decorate one's helmet."
+	desc = "Standard issue TGMC goggles. Mostly used to decorate one's helmet, but the tinting help with bright flashes too."
 	icon_state = "mgoggles"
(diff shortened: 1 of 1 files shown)


breadhunt:
If these are really needed you just go loot PO's sunglasses. But that requires tiding in and destroying lockers.

Good PR imo, flash protection isn't insanely strong at all (it's kinda shit compared to healthhud/imagers lol) and  the green welding goggles still exist as a straight-up better version of these.

Pariah919:
i use these for style please no

Lumipharon:
Flash in tgmc is only really to make close range nade use worse. I don't think nade spam needs to be buffed though.

Naaanii:
Nades flash?

bluezorua:
some people use these for style, maybe  add a variant with this effect?

TiviPlus:
Its used by people for style make a new variant for it

PR #10368, opened 2022-06-02
Laser sniper rework (IFF mode seperated)

Pariah919:
## About The Pull Request
Seperates IFF mode from normal mode, redoes description.

IFF MODE STATS
-
50/30 damage and 30 pen.
Short windup. (0.5)

## Why It's Good For The Game
Makes laser sniper not trash poopoo and gives it some level of gimmick.

## Changelog
:cl:
balance: Laser sniper IFF mode detangled from normal fire, it has a windup when firing, normal fire readded. IFF mode does lower damage, but is otherwise the same aside from damage and windup
spellcheck: Redid the description for laser sniper to be a bit more fluff due to the rework.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1795,10 +1795,17 @@ datum/ammo/bullet/revolver/tp44
 	hud_state = "laser_sniper"
 	damage = 70
 	penetration = 30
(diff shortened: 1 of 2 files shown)


novaepee:
Add `accurate_range_min` while you're at it. If energy sniper gets IFF, then it must be double down that it's not another carbine.

PR #10375, opened 2022-06-03
Make automated miners require open turfs on all 8 sides to sell ore (excluding map spawn turfs and xeno resin walls)

Wayland-Smithy:
## About The Pull Request

Guess who's back?
Can later be extended to include dense objects (like closed airlocks 😉)

This change is only a stop gap measure and will not prevent all wall cheese as in places where mapped in walls are you will be able to get away with leaving space around the rest and connecting to the existing closed turfs. I much prefer removing the upgrade altogether.

## Why It's Good For The Game
See mathematical explanation for why this is bad here: https://github.com/tgstation/TerraGov-Marine-Corps/pull/10375#issuecomment-1146151053
image
image
image

## Changelog
:cl:
balance: Automated miners require open turfs on all 8 sides to sell ore with the exception of xeno resin walls and turfs present when the map loads.
/:cl:

### DIFF
--- a/code/game/objects/machinery/miner.dm
+++ b/code/game/objects/machinery/miner.dm
@@ -41,6 +41,8 @@
 	var/miner_upgrade_type
 	///What faction secured that miner
 	var/faction = FACTION_TERRAGOV
+	///What directions had closed turfs on init, to skip during later checks
+	var/list/ignore_turf_dirs = list()
 
 /obj/machinery/miner/damaged	//mapping and all that shebang
 	miner_status = MINER_DESTROYED
@@ -56,6 +58,10 @@
 	SSminimaps.add_marker(src, z, hud_flags = MINIMAP_FLAG_ALL, iconstate = "miner_[mineral_value >= PLATINUM_CRATE_SELL_AMOUNT ? "platinum" : "phoron"]_off")
 	start_processing()
 	RegisterSignal(SSdcs, COMSIG_GLOB_DROPSHIP_HIJACKED, .proc/disable_on_hijack)
+	for(var/direction in GLOB.alldirs)
+		var/turf/selected_turf = get_step(loc, direction)
+		if(!isopenturf(selected_turf) && !istype(selected_turf, /turf/closed/wall/resin)) //Must be open on all sides to operate
+			ignore_turf_dirs += direction
 
 /obj/machinery/miner/update_icon()
 	switch(miner_status)
@@ -267,6 +273,8 @@
 		if(miner_upgrade_type == MINER_AUTOMATED)
 			var/obstructions = list()
 			for(var/direction in GLOB.alldirs)
+				if(ignore_turf_dirs.Find(direction))
+					continue
 				var/turf/selected_turf = get_step(loc, direction)
 				if(!isopenturf(selected_turf) && !istype(selected_turf, /turf/closed/wall/resin)) //Must be open on all sides to operate
 					obstructions += selected_turf.name
--- a/code/game/objects/machinery/miner.dm
+++ b/code/game/objects/machinery/miner.dm
@@ -265,18 +265,25 @@
 		return
 	if(add_tick >= required_ticks)
 		if(miner_upgrade_type == MINER_AUTOMATED)
-			for(var/direction in GLOB.cardinals)
-				if(!isopenturf(get_step(loc, direction))) //Must be open on one side to operate
-					continue
-				SSpoints.supply_points[faction] += mineral_value
-				do_sparks(5, TRUE, src)
-				playsound(loc,'sound/effects/phasein.ogg', 50, FALSE)
-				say("Ore shipment has been sold for [mineral_value] points.")
+			var/obstructions = list()
+			for(var/direction in GLOB.alldirs)
+				var/turf/selected_turf = get_step(loc, direction)
+				if(!isopenturf(selected_turf) && !istype(selected_turf, /turf/closed/wall/resin)) //Must be open on all sides to operate
+					obstructions += selected_turf.name
+
+			if(length(obstructions) > 0)
+				playsound(loc,'sound/machines/buzz-two.ogg', 35, FALSE)
+				say("Ore shipment export failed. Obstructed by: [jointext(obstructions, ", ")].")
 				add_tick = 0
 				return
-			playsound(loc,'sound/machines/buzz-two.ogg', 35, FALSE)
+
+			SSpoints.supply_points[faction] += mineral_value
+			do_sparks(5, TRUE, src)
+			playsound(loc,'sound/effects/phasein.ogg', 50, FALSE)
+			say("Ore shipment has been sold for [mineral_value] points.")
 			add_tick = 0
 			return
+
 		stored_mineral += 1
 		add_tick = 0
 	if(stored_mineral >= 8)	//Stores 8 boxes worth of minerals


Derrick2Electricboogaloo:
Terrible. Easy for beans to crush/acid the walls in a minute

Lumipharon:
This seems excessive considering normal walls can be melted by any acid caste  and really require an absurd amount of plasteel and time to make.

FosterGrime-Heymynameslouie:
the amount of risk-reward ,

(the sheer amount of metal to make walls, how easy it is to destroy walls in the first place, and the output of the miner itself) 

makes straight up walling the miner hurt for marines/engies in the short term but helpful in the long term if xenos just dont decide to take out the walls that takes a minute to melt, or 30 seconds to crush, and thats not mentioning the fact that once a wall has acid it will forever be unfixable and makes a giant unrepairable hole on it.

Naaanii:
Neat. Marines can still wall off miners, but will now require even more mats.

Wayland-Smithy:
I think it's funny to call this little bit of code excessive yet the engineers building ~8 walls because they know automated miners do not require access to gather the phoron crates get a pass. Will it still be okay when they start building additional defenses around the little wall forts to make it extra cancer to disable 1 plat miner? At some point the engi powercreep needs to be held at bay.

Let's dissect this old and tired "just melt it" argument with a basic walls + weak acid example (which also requires a large xeno to open the acid holes). I won't even consider crushing as it requires a specific T3 with a lane of at least 14 tiles and even then is not guaranteed to break a wall with any speed because of RNG. The fact crushing was even brought up IMHO shows at worst your bad faith position and at best how little thought you put into the argument. If xenos had some sort of special defense that only expensive req gear could break then perhaps I could understand at least drawing a parallel.

First we need some code for reference: 
https://github.com/tgstation/TerraGov-Marine-Corps/blob/8279c2d9406356e7eda639bdefda1954d8ecf3d0/code/game/objects/effects/aliens.dm#L141-L145
https://github.com/tgstation/TerraGov-Marine-Corps/blob/8279c2d9406356e7eda639bdefda1954d8ecf3d0/code/game/objects/effects/aliens.dm#L154-L158
https://github.com/tgstation/TerraGov-Marine-Corps/blob/8279c2d9406356e7eda639bdefda1954d8ecf3d0/code/game/objects/effects/aliens.dm#L171-L173
https://github.com/tgstation/TerraGov-Marine-Corps/blob/8279c2d9406356e7eda639bdefda1954d8ecf3d0/code/game/objects/effects/acid_hole.dm#L59-L62

@Lumipharon or anyone please feel free to correct my math below if I am mistaken.

```
8 strength_t / 0.0016 acid_strength = 5000 real ticks requiring specific castes that can use acid
5000 real ticks = 500 seconds (plus up to 5 more seconds for SSslowprocesisng) = weak acid time needed to make a hole
60 real ticks = 6 seconds = 1 acid hole expanding requiring specific castes that can expand holes
30 real ticks = 3 seconds = 1 miner slash doAfter

1400 real ticks = 140 seconds = 1 miner ore export
1 miner ore export = 15 to 30 req points
8 + 2 + 2 metal = 12 metal = 1 wall from scratch
50 metal sheets = 20 req points = 4 walls from scratch (for reference right now the cheese requires 6 walls)

~512 seconds to disable a walled miner with perfect coordination assuming available castes
divide by 140 seconds per export
equals 3 exports at minimum with more than half progress to a 4th!

if regular:
3 * 15 = 45 points = 100 metal sheets + 5 points leftover = 8.3 new walls from scratch
= 138% return on investment minimum!
if plat:
3 * 30 = 90 points = 200 metal sheets + 10 points leftover = 16.6 new walls from scratch
= 277% return on investment minimum!
```
How is this not cheesy?
Especially compared to a miner protected with cades that can be slashed by any caste, and by multiple xenos at once to speed up the process.

I do not like the solution of blocking automatic miner operations if closed turfs are touching either, but it is the easiest thing to change without having even harder conversations about acid speed to dissolve turfs, nerfing automated miners some other way, or increasing the cost of walls to require a ton of metal and raising the cost of rwalls even higher. However if you have any constructive points on the matter I am all ears.

Losenis:
>Needing all turfs not occupied around the miner

Did you actually go and check every single miner in every single map to see if they're actually free in that manner? And that they're free enough to land the tadpole close without unintentionally shutting them off with the tadpole?

And does this mean xenos can just drop a single wall close to them and completely bypass the point behind the minimap showing if a miner is operational or not?

Wayland-Smithy:
> Did you actually go and check every single miner in every single map to see if they're actually free in that manner? And that they're free enough to land the tadpole close without unintentionally shutting them off with the tadpole?

Engineers have access to this neat tool called a Plasma Cutter than can remove like any turf in the game.

> And does this mean xenos can just drop a single wall close to them and completely bypass the point behind the minimap showing if a miner is operational or not?

If you read the changelog it mentions how resin walls are do not count. The logic there is so hivemind can't just cuck miners with little risk.

Derrick2Electricboogaloo:
> t

Notice how it is mostly on prison? Seems one map is the issue.

Lumipharon:
It's excessive because it takes time to setup walls around a miner. Either I set the walls first so the miner isn't running that whole time so it's not visible on the map),orI set it up first and risk getting attacked while building 8 walls...

Or better yet, I don't waste time on walls, and instead spend the time setting up 3 other miners, and save all those mats.

I dunno what Plat miners you can do with this, aren't they all in caves now, except maybe prison?

So I'm not opposed to this PR, I'm saying it's excessive because I think walling is a waste of time and resources in most cases.

Losenis:
I've also noticed that the 1 empty space free requirement also already guarantees a small caste can still get inside to slash it after another caste acids the wall.

Wayland-Smithy:
> It's excessive because it takes time to setup walls around a miner. Either I set the walls first so the miner isn't running that whole time so it's not visible on the map),orI set it up first and risk getting attacked while building 8 walls...

It only takes 6 walls, so if you build 5 then fix the miner, you only need to finish the last wall by positioning the girder and adding the 4 metal. Regardless of the risk/reward spread it is too good on the most minimal risk/minimal reward side based on my math which assumes the mining starts after all 6 walls are in place which means in practice you might be able to get 4 exports before it is taken out.

You, and the others here, also seem to be assuming xenos are hyper vigilant and will notice a new miner online the very same tick it is activated. In fact the reality of the situation could not be further from the truth if my anecdotes are anything to go by of pleading in hivemind chat for anyone to help take out even non-cheesed miner setups with a few cades and a single guard. That though is a skill issue, so I think you can easily start it mining in a far corner from the action and expect to get 6 walls up in that time. If you are worried about some random runner showing up put 1 or 2 BAS near you. Sentries are so over tuned that if you clear surrounding weeds that will keep you safe for a long time.

> Or better yet, I don't waste time on walls, and instead spend the time setting up 3 other miners, and save all those mats.

As soon as you leave one undefended it can be taken out. As it should be. The issue is walls are stronger than cades and require specific tools (i.e. xeno castes) to deal with and even with those the time needed guarantees you will make back the cost of metal for the walls and then some. 

> I dunno what Plat miners you can do with this, aren't they all in caves now, except maybe prison?

Not sure how the miner being in caves changes the validity of the cheese. I have seen it in Big Red and LV caves before.

> So I'm not opposed to this PR, I'm saying it's excessive because I think walling is a waste of time and resources in most cases.

Cool, so we agree on the fact that walling miners is excessive and unfun.

Wayland-Smithy:
> I've also noticed that the 1 empty space free requirement also already guarantees a small caste can still get inside to slash it after another caste acids the wall.

Sure yes, and that 1 gap is only needed because that was anti-cheese part 1 and as is apparent does not stop the cancer.
A small caste can bypass the 6 additional seconds that big/strong castes can not. It does not change the math much if at all.

TiviPlus:
Discussed this with bravemole we prefer a fix like https://github.com/tgstation/TerraGov-Marine-Corps/pull/10408 as we consider this a mapping issue as well as requiring all 8 sides being excessive

PR #10384, opened 2022-06-04
Martini has reverse falloff

Pariah919:
## About The Pull Request
Martini gains 3 damage per tile, rather than having no falloff.

## Why It's Good For The Game
Gives martini some level of gimmick beyond being a funny meme gun, honestly, and in general distances you won't see the damage buff TOO too much, unless you're 40 tiling people somehow with this thing.

## Changelog
:cl:
balance: Martini now gains damage per tile traveled, 3 per tile.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1050,6 +1050,7 @@ datum/ammo/bullet/revolver/tp44
 	damage = 120
 	penetration = 20
 	sundering = 10
+	damage_falloff = -3
 
 /datum/ammo/bullet/sniper/martini/on_hit_mob(mob/M, obj/projectile/P)
 	staggerstun(M, P, weaken = 1, stagger = 1, knockback = 2, slowdown = 0.5, max_range = 5)


goose4429:
bruh

AnCopper:
So like, can you explain why?

Kuro020:
> So like, can you explain why?

read the why its good for the game

Al-1ce:
> read the why its good for the game

Usually making a change like this for shits and giggles (A gimmick) isn't good reasoning but this does sound interesting.

Lumipharon:
I don't want to say "much realism", but this makes no sense and isn't intuitive for players. It also probably make no noticable difference.

Does it need a magic gimmick when it already has a gimmick of huge damage, one shot gun?

goose4429:
This is unintuitive, and will not be known by players. Except if they check Git or code. 

Also, unbalanced - the only change from this is that xeno players are suddenly more affected by people further from them. What.

If you want to buff a weapon, just give it damage. "Explodes in hands" is also a gimmick, but it is a bad idea to add it to any gun. There is no reason to add this.

Dueu:
I'm pretty sure I remember we TM the reverse falloff martini a long time ago and wasn't merged.

Al-1ce:
> I'm pretty sure I remember we TM the reverse falloff martini a long time ago and wasn't merged. Ah found it #8993

That PR was way overdone compared to this. I think at least testing a smaller, minimalized version would be interesting. The martini probably should have some sort of gimmick outside of being hundreds of years old.

I wonder though, would removing the scope attachment for the martini balance this? I could definitely see it being used as an improved sniper should negative falloff be implimented.

theta060:
> Gives martini some level of gimmick beyond being a funny meme gun, honestly, and in general distances you won't see the damage buff TOO too much, unless you're 40 tiling people somehow with this thing.

you underestimate my unga

BraveMole:
I remember tming that and it was shit

PR #10390, opened 2022-06-05
Adds the SMG-38 Personal Defense Rifle to the game and marine vendors.

Pariah919:
## About The Pull Request
Adds the SMG38, a new SMG-Rifle hybrid for marines to use, also adds Penetration modifier as a modifier to guns for better control of gun modification.

Sprites by Kuro
-
Other than the T38 box, which I made in 10 seconds..

Stats of the Gun
-
Damage is 15.
Pen is 15.
Magazine size is 70.
Slowdown is 0.35.
Extra delay between bursts is M16 level, could be subject to change.
Burst delay is 0.2, burst amount is 3 bullets, SMG standard.
Unwielded stats are rifle level, not much akimboing here.
No Underbarreled attachments, limited attachments overall.

## Why It's Good For The Game
Adds a gimmicky gun to the marine loadout that trades off certain things to be allowed to rock with rifle stats and such, it should outperform the P90 on armored castes but still be worse against unarmored ones.

## Changelog
Kuro, Pariah919
:cl:
add: Added new SMG-38 personal defense rifle to the marine vendors, sprites by Kuro, gun balance mostly by me other than the base idea.
code: Added penetration modifier to guns, additive, so negative penetration modifiers make you lose pen, positive adds to pen.
/:cl:


Kuro020 (approved):
Kuro sprites are always an immediate approval, he's pretty good

TooFewSecrets:
I see a few issues with the early stats:
For one, I don't think the extra pen is going to ever make the gun noticeably deadlier than the P90. Swapping 5->10 pen with 20->15 damage breaks even at 90 base armor, so, only ancient defenders who are mid-fortify (ancient king warding pheros bring ancient crushers exactly up to the break-even point), and either way you're dealing ~3 damage per bullet so sunder might actually let the P90 pull ahead? Either way, this gun flatly does less damage against any other target along with worse handling than the P90 (so you can't chase targets as far and end up doing even less damage).
In exchange for this you get to shoot 3 bullets every 0.4 seconds instead of every 0.45, and you get 20 more rounds which are not an advantage because you need more bullets to kill things when your damage has been reduced by 25%. The only other benefit I see is having 0 scatter instead of 1, which is probably going to be fixed on the P90 by a barrel mod because you can't put bayonets on SMGs and they can only take laser sights as an underbarrel anyway - and damage falloff, but if you care about that you can just take the AR-21 for similar ergonomics and better DPS in most cases.
On that note, this thing is a rifle-sized PDW, it should have higher force - most rifles get 20, right now it has 8.
Also, why does it have access to the BFA? Doesn't that just make it worse? Shots within bursts are fired at the normal speed, you only pull ahead at the end of bursts because of the lowered delay there, so you want bursts as short as possible.

goose4429:
What's the point of giving guns pen and dmg modifiers instead of just giving it its own ammo? That's what you're doing essentially, just with more steps. And that's what you for one told me not to do. Bruh.

But pen modifier is probably nice, for example to use on ext barrel attach. 

70 round mag on smol SMG is questionable.

Other than that, in current state seems pretty useless. Why would I want SMG with just some higher mobility over TX-11 with quite literally 50% more DPS?

Maybe give the SMG more pen? We don't have SMG with AP rounds, that would be nice. For example MP7 is probably good prototype for that. https://en.wikipedia.org/wiki/Heckler_%26_Koch_MP7

Kuro020:
> I see a few issues with the early stats: For one, I don't think the extra pen is going to ever make the gun noticeably deadlier than the P90. Swapping 5->10 pen with 20->15 damage breaks even at 90 base armor, so, only ancient defenders who are mid-fortify (ancient king warding pheros bring ancient crushers exactly up to the break-even point), and either way you're dealing ~3 damage per bullet so sunder might actually let the P90 pull ahead? Either way, this gun flatly does less damage against any other target along with worse handling than the P90 (so you can't chase targets as far and end up doing even less damage). In exchange for this you get to shoot 3 bullets every 0.4 seconds instead of every 0.45, and you get 20 more rounds which are not an advantage because you need more bullets to kill things when your damage has been reduced by 25%. The only other benefit I see is having 0 scatter instead of 1, which is probably going to be fixed on the P90 by a barrel mod because you can't put bayonets on SMGs and they can only take laser sights as an underbarrel anyway - and damage falloff, but if you care about that you can just take the AR-21 for similar ergonomics and better DPS in most cases. On that note, this thing is a rifle-sized PDW, it should have higher force - most rifles get 20, right now it has 8. Also, why does it have access to the BFA? Doesn't that just make it worse? Shots within bursts are fired at the normal speed, you only pull ahead at the end of bursts because of the lowered delay there, so you want bursts as short as possible.

at what ap levels would it start outdamaging the p90 on armored castes but still losing to it on unarmored castes? if its too weak we can buff the ap to that level so it outpeforms the p90 on said castes

Kuro020:
> What's the point of giving guns pen and dmg modifiers instead of just giving it its own ammo? That's what you're doing essentially, just with more steps. And that's what you for one told me not to do. Bruh.
> 
> But pen modifier is probably nice, for example to use on ext barrel attach.
> 
> 70 round mag on smol SMG is questionable.
> 
> Other than that, in current state seems pretty useless. Why would I want SMG with just some higher mobility over TX-11 with quite literally 50% more DPS?
> 
> Maybe give the SMG more pen? We don't have SMG with AP rounds, that would be nice. For example MP7 is probably good prototype for that. https://en.wikipedia.org/wiki/Heckler_%26_Koch_MP7

its a smg-rifle hybrid thingy, having 70 rounds is fine (its not even a small gun).
tx11 is a burst gun so comparing an automatic gun to it is pretty inacurrate I would say, aside from that like I said on the coment above, if its too weak buffing AP would be a good idea yeah

Pariah919:
Upped to 15/15 over 15/10.

Al-1ce:
Looks interesting. As mentioned above, seperate AP ammunition would fix most of the concerns in terms of damag3 here.

TiviPlus:
Wheres the gimmick

Kuro020:
> Wheres the gimmick

Read why its good for the game nerd

Dueu:
I find this gun very similar to the AR-11 in a lot of ways:

- Both have the same damage and penetration at the same number (AR-11 has 20 and the SMG-36 has 15)
- Both have a positive accuracy modifier when wield (AR-11 is lightly better)
- Both have burst of 3 bullets
- Both have very limited options when it comes to attachments
- Both cant be akimbo

The few differents are that the SGM-36 is better when it comes to provide cover fire as it has less aim slowdown, a bit less fire delay and probably better quick draw as the AR-11 requires 0.65 seconds to wield it and submachineguns tends to have less but changes a bit of raw damage output and extra falloff for for less fire delay.

Kinda niche but when someone care about SMG content? How long the T90 and T19 was the only 2 SMG centered ballistic wepaons in vendors normally? Not including seasonal weapons.

PR #10394, opened 2022-06-05
Removes germs from internal organs

QualityVan:
## About The Pull Request
Bet you didn't know this even happened. Usually completely irrelevant, and absolutely uncommunicated except via body scanners/autodocs. And coughing, sometimes.

## Why It's Good For The Game
Streamlining medical. Germs hang out on limbs and do stuff there, plus if they went necrotic they already have a route to hurt organs.

## Changelog
:cl:
del: Internal organs can't get infections. They mostly didn't already.
/:cl:

### DIFF
--- a/code/game/objects/machinery.dm
+++ b/code/game/objects/machinery.dm
@@ -421,29 +421,8 @@
 		if(i.robotic == ORGAN_ROBOT)
 			mech = "Mechanical:<br>"
 
-		var/infection = "None"
-		switch (i.germ_level)
-			if (1 to INFECTION_LEVEL_ONE + 200)
-				infection = "Mild Infection:<br>"
-			if (INFECTION_LEVEL_ONE + 200 to INFECTION_LEVEL_ONE + 300)
-				infection = "Mild Infection+:<br>"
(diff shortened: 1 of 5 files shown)


breadhunt:
Why remove it though? This feels like it should be expanded on, not removed, at least to me

QualityVan:
Any expansion on infections I'd want to do would be on limbs specifically. They've got multiple ways to diagnose and treat, while organ infections really only have spaceacillin. Organs have their own health system as is, and infection can already interact with that in ways that aren't the organ's own germ level. That being said, if you have ideas go nuts.

breadhunt:
> Any expansion on infections I'd want to do would be on limbs specifically. They've got multiple ways to diagnose and treat, while organ infections really only have spaceacillin. Organs have their own health system as is, and infection can already interact with that in ways that aren't the organ's own germ level. That being said, if you have ideas go nuts.

Because I forgot to respond and just realised this got merged: Yeah that's fine, I was just wondering the reason behind it. If they ever do get expanded on it's probably better to bundle it as external germs

PR #10396, opened 2022-06-05
Revert "Health kits chain apply (#9977)"

yyzsong:
## About The Pull Request
This reverts commit 80cd10017ea02765796f3d21e49aa66f1b4d99a0.

## Why It's Good For The Game
No clue how this got merged

#9977 makes it so that advanced kits can chain apply to all limbs, healing them without a need for input on the user's part. This *sounds* fine, except where it might waste kits on like, 3 damage limbs

except this PR made the chained applications *not consume kits*

one brute kit can heal up to 132 damage with a single kit, as long as you stay still!!

it's a needless corpsman buff. If you run out of kits, bring more. 
Btw this PR i'm reverting was an 'alternative' to one that literally just gave kits +3 healing

## Changelog
:cl:
del: Reverts PR#9977 (Chain medkit application PR)
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -42,24 +42,19 @@
 			to_chat(user, span_warning("You can't apply [src] through [H.wear_suit]!"))
 			return TRUE
 
-	if(!can_affect_limb(affecting))
-		to_chat(user, span_warning("This isn't useful at all on [affecting.limb_status & LIMB_ROBOT ? "a robotic": "an organic"] limb."))
+	if(affecting.limb_status & LIMB_ROBOT)
+		to_chat(user, span_warning("This isn't useful at all on a robotic limb."))
 		return TRUE
 
 	H.UpdateDamageIcon()
 
 	if(user.skills.getRating("medical") < skill_level_needed)
-		if(user.do_actions || !do_mob(user, M, unskilled_delay, BUSY_ICON_UNSKILLED, BUSY_ICON_MEDICAL))
+		if(user.do_actions)
 			to_chat(user, span_warning("You're busy with something else right now!"))
+		if(!do_mob(user, M, unskilled_delay, BUSY_ICON_UNSKILLED, BUSY_ICON_MEDICAL))
 			return TRUE
 
 	return affecting
-
-///Checks for whether the limb is appropriately organic/robotic
-/obj/item/stack/medical/proc/can_affect_limb(datum/limb/affecting)
-	if(affecting.limb_status & LIMB_ROBOT)
-		return FALSE
-	return TRUE
 
 /obj/item/stack/medical/heal_pack
 	name = "platonic gauze"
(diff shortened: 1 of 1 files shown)


FosterGrime-Heymynameslouie:
image
honestly this should only apply to advance kits, ointment and gauze is fine as is.
should either revert the chain apply on advance kits or reducde the amount of kits itself to 5 or something as state on the original pr.

XSlayer300:
I did mention this in the PR, but it was merged before complaints were actually being considered.
image

yyzsong:
>honestly this should only apply to advance kits, ointment and gauze is fine as is. should either revert the chain apply on advance kits or reducde the amount of kits itself to 5 or something as state on the original pr.

I'm no coder, and am incapable of such a feat. Revert will do in the meantime.

Reducing the amount of kits means reducing the amount of *people* you can treat, and each kit still treats 132 damage each. It'd still be, like, 5x more healing than a manual stack of kits

even if you restricted it down to *one*, it'd still heal 132 damage.
a full stack of old kits would heal 10x12, for 120 healing

and if you tried to restrict sizes down to one or something, you'd just get back at the same problem the original PR was trying to fix (Running out of kits) (although I think this really isn't a problem, just give reqtorio kit packs or whatever)

related to lowering stack size: It'd detract from the amount of time a corpsman could spend at a frontline. Plus less amounts of more powerful heals doesn't work if the numbers go up.

Lumipharon:
This is kinda overblown.

It has a high theoretical heal value, but how often do you have serious damage on like 5 body parts?

If you're burnt maybe, but brute is usually 1-2 body parts, and pill bottles are 16 pills of 75 healing without having to stand around for a while.

DeltaFire15:
> This is kinda overblown.
> 
> It has a high theoretical heal value, but how often do you have serious damage on like 5 body parts?
> 
> If you're burnt maybe, but brute is usually 1-2 body parts, and pill bottles are 16 pills of 75 healing without having to stand around for a while.

People who got exploded tend to have damage on every single limb, same with victims of flame or environmental effects. When using chain apply kits that was the case where the efficiency difference was very notable. Not taking either side on this, merely stating my experience.

Mechanelize:
haven't seen this ever abused, overused, or complained about, seems like a non-issue
advanced kits really only used by non-medics besides trying to get damage below 200 for defib
while the amount of healing potential is high, I don't think it gets near there in practice, compared to pills which work automatically and are easier to store.
before the PR making them chain i found myself pretty much never using them outside of getting people below 200 as a medic, now its somewhat viable but puts you out of the fight for a bit to heal

FosterGrime-Heymynameslouie:
and also the apply time for advance kits, if you dont have medical skill, is abysmal. 
so this really only applies to medics and docs.

as a compromise I think the amount of kits should just be lowered to 5 or something

TooFewSecrets:
I don't see any significant reason to use trauma kits without this functionality. Even WITH this they're still worse than pills at their best, and assuming the maximum possible heal for every single charge is hilariously overblown - for normal marines, they're worse than even pill packets. You don't present any reason this *should* be reverted in your PR. I haven't seen a single player complaining about it, either.

TiviPlus:
If you want every use to use them then make a pr fixing it instead of reverting?
You fix issues rather then removing the entire feature

PR #10401, opened 2022-06-06
Fixes projectile speed modifiers

### LINKED ISSUE #10364: Bullet speed modifiers do nothing
### Description

Gun variable shell_speed_mod(and by extension attachment attach_shell_speed_mod) does not affect actual projectile speed at all.
Practically it means that supressor, barrel charger and extended barrel does not affect bullet speed. It also means barrel charger is effectively a direct downgrade.

### Test Merges

None

### Reproduction Steps

1. Get a gun
2. Attach barrel charger/supressor
3. Witness no difference

Bonus steps
1. Get a gun
2. Directly modify it's shell_speed_mod to ridiculous values
3. Witness even more of no difference

### Screenshots

https://user-images.githubusercontent.com/78968085/171409953-107e8046-c114-4ec9-9c82-a4dc35fbab27.mp4

Lumipharon:
## About The Pull Request
Fixed #10364

It turns out modifiers to projectile speed does fucking nothing, because the proc that adds it just gets, gets overridden by the proc that calls it in the first place.

Barrel charger projectiles are a bit funny because you can barely see them they move so fast, but eh, bugfix.

## Why It's Good For The Game
Bugfix good.

## Changelog
:cl:
fix: Projectile speed modifiers now actually work
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -871,7 +871,7 @@
 
 	simulate_recoil(dual_wield, firing_angle)
 
-	projectile_to_fire.fire_at(target, master_gun ? gun_user : loc, src, projectile_to_fire.ammo.max_range, projectile_to_fire.ammo.shell_speed, firing_angle, suppress_light = HAS_TRAIT(src, TRAIT_GUN_SILENCED))
+	projectile_to_fire.fire_at(target, master_gun ? gun_user : loc, src, projectile_to_fire.ammo.max_range, projectile_to_fire.projectile_speed, firing_angle, suppress_light = HAS_TRAIT(src, TRAIT_GUN_SILENCED))
 
 	shots_fired++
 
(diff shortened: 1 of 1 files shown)


Lumipharon:
> Can wait to see marine winrate skyrocketing because of ridiculous overall accuracy buff and the fact that a lot of balance tweaks were merged without taking the bug into account

I would be entirely surprised if it made any difference whatsoever.

Jordanstolethepreciousslam:
I mean, i can see how with that much speed ballistic weapons would compete with lasguns for hitscan niche. Also sniper rifles hitting anything that is not still target outside cqc range? ridiculous, 10/10, speedmerge

Jordanstolethepreciousslam:
Well, "compete with lasguns for hiscan niche" is a bit of overstatement, but the difference notable

Jordanstolethepreciousslam:
Btw this is how the changes looks on practice. 

https://user-images.githubusercontent.com/78968085/172384552-c85e8ad3-2be0-457d-95e9-c8365952f071.mp4

crayonLover09:
Suppressor looks like its moving through butter. The attachies affected by this might need a rework since this is the first time they work as intended

TiviPlus:
If theres balance changes you fix that instead of leaving things broken on purpose

PR #10404, opened 2022-06-06
Fixes failing legs code

QualityVan:
## About The Pull Request
Any two legs/feet that were missing or broken were supposed to cause an occasional knockout but this usually didn't happen because missing limbs don't process. Fixes that, and also reduces the knockout from a 20 second unconscious to a 3 second knockdown. It was a little overkill.
Also fixes amputated limbs not being amputated.

## Why It's Good For The Game
Bugfixes, and a random 20 second unconscious is not necessary.

## Changelog
:cl:
fix: When you have two broken or missing legs or feet, you'll now fall down once or twice a minute.
fix: Amputated limbs are amputated.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_organs.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_organs.dm
@@ -22,7 +22,7 @@
 			if(overeatduration > 1)
 				overeatduration -= 2	//Doubled the unfat rate
 
-	var/leg_tally = 2
+	var/leg_tally = 0
 
 	last_dam = getBruteLoss() + getFireLoss() + getToxLoss()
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/life/handle_organs.dm
```diff
@@ -31,6 +31,11 @@
 
 	for(var/i in limbs)
 		var/datum/limb/E = i
+
+		if((E.name in list("l_leg", "l_foot", "r_leg", "r_foot")) && !lying_angle) //Need to do this before checking need_process in order to catch missing limbs
```
TiviPlus, line 35: O know this is just moved code but my eyes

code/modules/organs/limbs.dm
```diff
@@ -666,7 +666,10 @@ Note that amputating the affected organ does in fact remove the infection from t
 	if(body_part == CHEST)
 		return FALSE
 
-	set_limb_flags(LIMB_DESTROYED)
+	if(amputation)
+		set_limb_flags(LIMB_AMPUTATED & LIMB_DESTROYED)
```
TiviPlus, line 670: this is the same as
set_limb_flags(0) do you mean |

TiviPlus (approved):

TiviPlus (changes requested):
.

code/modules/organs/limbs.dm
QualityVan: i sure do whoops

PR #10409, opened 2022-06-07
Optimize projectiles moving (by changing how lit turfs calculates with low light range)

Vallat:
## About The Pull Request

Each projectile lights the turfs around it and while moving updates their luminosity in the range of 2 tiles. However, in fact, projectile lights turfs only in range of 1 tile because of the lighting mask sprite. Therefore, we can safely replace the expensive `view()` with a cheaper `block()` for calculating lit turfs for small lights.

Profling before changes
image
Profiling after
image

The `get_new_turfs()` function is about 20% of the load of `projectile_batch_move()`, so in overall we reduced the projectile movement load by about 10%

I also changed some logical operators to get performance increase in 0.00000001%, main logic hasn't been changed

## Why It's Good For The Game

Less lags - more bullets will fly

## Changelog
:cl:
refactor: refactored luminosity calculation for small movable lights therefore increased projectiles movement performance by ~10%
/:cl:

### DIFF
--- a/code/datums/components/overlay_lighting.dm
+++ b/code/datums/components/overlay_lighting.dm
@@ -127,15 +127,16 @@
 /datum/component/overlay_lighting/proc/get_new_turfs()
 	if(!current_holder)
 		return
+	LAZYINITLIST(affected_turfs)
 	if(range <= 2)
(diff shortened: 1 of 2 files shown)


Lumipharon:
Neat

PR #10413, opened 2022-06-08
Akimbo nerfs

Lumipharon:
## About The Pull Request
Someone ded, plsnerf.

**shotguns**
For some reason shotguns had the default akimbo scatter values (so an extra 2-4 scatter for going akimbo).
Changed this to the rifle values of 3 and 5 (so between 6 and 10 extra scatter for akimbo), as it is presumably quite hard to akimbo two shotguns.

This should make akimbo combat shotguns still deadly at pb range, but medium/longer range slug spam will be considerably less reliable.

The only other shotgun you can actually do this with is the sawn off.

**Rifles**
Also changed rifles to have the standard akimbo firedelay debuff. It was removed when akimbo was made to actually effect autofire and I thought it was a bad idea at the time. Most notably it makes akimbo AR11 really gross at close range.

**SMG90**
Increased akimbo scatter for the SMG90 from the default 2-4 scatter to 4-8. This is still lower than the MP19/rifles, and as above should serve to make it less effective at longer ranges, while still being deadly at closer range.

## Why It's Good For The Game
Certain akimbo weapons are a bit too strong, outside of close range. This is bad because they have no slowdown, so can be a bit oppressive.

## Changelog
:cl:
balance: shotguns and the SMG90 no longer have extremely low akimbo scatter penalties
balance: rifles no longer have zero akimbo firedelay penalties
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -19,7 +19,6 @@
 	scatter_unwielded = 13
 	recoil_unwielded = 4
 	damage_falloff_mult = 0.5
-	akimbo_additional_delay = 0
 	upper_akimbo_accuracy = 5
 	lower_akimbo_accuracy = 3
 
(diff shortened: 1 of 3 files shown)


Pariah919 (approved):
Seems decent to me, though it needs a better title.

Lumipharon:
> Seems decent to me, though it needs a better title.

Pluralised direct kinetic projectile thrower game entertainment adjustment

PR #10415, opened 2022-06-08
ravager plasmadrain immunity

Mechanelize:
## About The Pull Request

Gives the Ravager caste plasmadrain immunity, like the Gorger.

## Why It's Good For The Game

Ravagers and Gorgers do not use plasma, instead using Rage and Blood respectively, so should not be affected by the plasmadrain status effect in the first place. For that reason, and because both castes need to obtain a significant portion of their plasma substitute from marines instead of the standard over time regen, Ravager should be given immunity to plasmadrain.

## Changelog
:cl:
balance: Ravager caste now has plasmadrain immunity.
/:cl:


Mechanelize:
iirc tesla zap will still drain some but won't apply the over time status effect

FosterGrime-Heymynameslouie:
> iirc tesla zap will still drain some but won't apply the over time status effect

 this is all I ever wanted in life

based pr speede merhg

PR #10428, opened 2022-06-10
Makes resin structures cheaper faster to build, start weaker

Losenis:
## About The Pull Request
Changes wall stats so that they start at a far lower health (40 from 150, 100 from 250 for HL). Resin door plasma and time cost equalized to wall (it was triple as expensive and twice as time consuming despite getting shot out in a single buckshot). Sticky and wall build time halved. Plasma cost down to 60 from 75, sticky to 20 from 25. Health upgrade per tick on walls increased from 3 to 6 so that they mature to max health at roughly the same time (bit faster due to rounding).

## Why It's Good For The Game
Making mazes is CBT, and with the recent flamer changes it's becoming more of a pain to maze given how easily it can be destroyed with the right tools. Even without said counters, to build a maze is an exercise in constant clicking, waiting, and labor that requires your constant focus. Inevitably as people get tired of building them (For example, a hivelord player that mazed half the map only to find out that the marines got fobsieged and massacred), they'll inevitably start building only the necessary. This can result in instances where only the fighting areas and the silo are mazed with hardly any other areas.

The issue becomes apparent in situations where xenos get one or two killed and the rest pushed back out of their frontline maze, only to find that marines will chase them through open ground due to the pain of building mazes that could've allowed for a safe retreat path. Castes that can build said maze would find it more fun and more impactful to just ignore that and head to the front to heal combat castes and so on, leading to said wipes.

This PR aims to make mazing a little less of a pain while still making sure to not cause an overboard frontline spam of quick walls by having them be weak at the moment of building them, so walling in a push of marines would not be feasible without a lot of teamwork. Ideally this'll allow sporadic building of walls with every break of combat or moment of inactivity without requiring so much time and focus. Due to the starting health being lowered that much, the maturity rate has been increased so as to make them mature fully at roughly the same time, so it's still feasibly to strengthen a maze close to the front, just not at a location where they're actively getting shot at.

## Changelog
:cl:
balance: Build time of resin structures halved. Resin door equalized in time and plasma cost to wall. Cost of walls and doors down to 60, sticky down to 20. Walls now start at 40 health, 100 for hivelord, but increase in max health faster so that they still fully mature at the same time.
/:cl:

### DIFF
--- a/code/game/turfs/walls/resin.dm
+++ b/code/game/turfs/walls/resin.dm
@@ -150,14 +150,14 @@
  * Regenerating walls that start with lower health, but grow to a much higher hp over time
  */
 /turf/closed/wall/resin/regenerating
-	max_integrity = 40
+	max_integrity = 150
 
 	/// Total health possible for a wall after regenerating at max health
 	var/max_upgradable_health = 300
 	/// How much the walls integrity heals per tick (5 seconds)
 	var/heal_per_tick = 25
 	/// How much the walls max_integrity increases per tick (5 seconds)
-	var/max_upgrade_per_tick = 6
+	var/max_upgrade_per_tick = 3
 	/// How long should the wall stop healing for when taking dmg
 	var/cooldown_on_taking_dmg = 30 SECONDS
 	///Whether we have a timer already to stop from clogging up the timer ss
@@ -200,4 +200,4 @@
 
 /* Hivelord walls, they start off stronger */
 /turf/closed/wall/resin/regenerating/thick
-	max_integrity = 100
+	max_integrity = 250
--- a/code/game/turfs/walls/resin.dm
+++ b/code/game/turfs/walls/resin.dm
@@ -150,14 +150,14 @@
  * Regenerating walls that start with lower health, but grow to a much higher hp over time
  */
 /turf/closed/wall/resin/regenerating
-	max_integrity = 150
+	max_integrity = 40
 
 	/// Total health possible for a wall after regenerating at max health
 	var/max_upgradable_health = 300
 	/// How much the walls integrity heals per tick (5 seconds)
 	var/heal_per_tick = 25
 	/// How much the walls max_integrity increases per tick (5 seconds)
-	var/max_upgrade_per_tick = 3
+	var/max_upgrade_per_tick = 6
 	/// How long should the wall stop healing for when taking dmg
 	var/cooldown_on_taking_dmg = 30 SECONDS
 	///Whether we have a timer already to stop from clogging up the timer ss
@@ -200,4 +200,4 @@
 
 /* Hivelord walls, they start off stronger */
 /turf/closed/wall/resin/regenerating/thick
-	max_integrity = 250
+	max_integrity = 100
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -166,10 +166,10 @@
 	action_icon_state = RESIN_WALL
 	mechanics_text = "Builds whatever resin you selected"
 	ability_name = "secrete resin"
-	plasma_cost = 75
+	plasma_cost = 60
 	keybind_signal = COMSIG_XENOABILITY_SECRETE_RESIN
 	///Minimum time to build a resin structure
-	var/base_wait = 1 SECONDS
+	var/base_wait = 0.5 SECONDS
 	///Multiplicator factor to add to the building time, depends on the health of the structure built
 	var/scaling_wait = 1 SECONDS
 	///List of buildable structures. Order corresponds with resin_images_list.
@@ -229,7 +229,7 @@
 		if(/obj/effect/alien/resin/sticky)
 			build_resin_modifier = 0.5
 		if(/obj/structure/mineral_door/resin)
-			build_resin_modifier = 2
+			build_resin_modifier = 1
 
 	return (base_wait + scaling_wait - max(0, (scaling_wait * X.health / X.maxHealth))) * build_resin_modifier
 
@@ -329,7 +329,7 @@
 		if(/obj/effect/alien/resin/sticky)
 			plasma_cost = initial(plasma_cost) / 3
 		if(/obj/structure/mineral_door/resin)
-			plasma_cost = initial(plasma_cost) * 3
+			plasma_cost = initial(plasma_cost)
 
 	if(new_resin)
 		add_cooldown(SSmonitor.gamestate == SHUTTERS_CLOSED ? get_cooldown()/2 : get_cooldown())


FosterGrime-Heymynameslouie:
basded pr spede mrghe

make building forts fun again

Lumipharon:
I don't think this is a good idea, purely because flamers are overpowered against resin structures right now, and they really need to be nerfed back.

This is kind of bandaiding that problem, but then it leaves nonflamer methods of clearing as useless, then marines either have to spam flamers, or someone makes some other PR to make other wall clearing shit better/walls weaker or whatever.

Al-1ce:
>(40 from 150, 100 from 250 for HL)

Just so we're clear, this would make drone walls **_barely_** stronger than a larva. Barely.

```
icon_state = "Bloody Larva"
a_intent = INTENT_HELP //Forces help intent for all interactions.
maxHealth = 35
```

Fresh walls already are squishy. Marines can easily delete them in just a spray or two, but this is ridiculously, woefully underpowered. I can only imagine this being a good thing before marines have landed, assuming support castes actually do their job. Expecting all walls to have matured in a combat situation where you are periodically repairing resin structures as the battle pushes back and forth is beyond unrealistic. Again. Almost the same amount of health as a larva. 

I would be in favor of lowering plasma consumption in general as the waiting/rest periods are demoralizing, especially to new drones. Additionally, it would fit the higher pop climate we've seen with Nuclear War.

Otherwise, I'm going to say no on this one. Plenty sure others can agree...

BraveMole (changes requested):
You are not making maze less of a pain, you are doing the opposite.

Now as marine you have more walls to down and as a xeno you have more walls to build...

Autisem:
I think xeno structures as they stand are too weak already. marines can trivialy unga ball trough even wack as mazes using 1 flamer. And this will make the building way more tedious and make them way less effective other then what you build around the start

Lauwma:
The thing that makes mazing CBT is not the build time, it the fact you need to travel every tiles you build on, you can't realistically make mazes built faster because of that. And I don't think there's a solution to make making mazes less tedious. The only real solution to the problem is a flamer revert

Dueu:
So the thing that actually makes mazing CBT is how is done: As drone you go, build a resin wall (click), move to the next postion (key presses) and repeat many, many, many times and is not that much better to do as hivelord as you are limited to the area next to you.

The cost of plasma is already reduced by half before shutters open so maybe only reducing time construction in exchange of starting health could be a solution but I would prefer a way to control how much health the structures begins with, that way I can keep building somewhat strong walls as hivelord during pushes and make mazes faster before combat begins.

Other solution I come to think would be structures schematics that the queen, shrike or leaders can make and any caste that can give plasma to others can construct them, you give all your plasma to one tile and it starts to built itself to all connected structure schematics until it uses all the plasma given, while less efficient in plasma cost, is faster because you are using the plasma in bursts instead of constant usage with periods of resting to gain plasma faster and way less clicks and key presses.

Losenis:
> You are not making maze less of a pain, you are doing the opposite.
> 
> Now as marine you have more walls to down and as a xeno you have more walls to build...

Marines having more walls to down is part of the point given their flamers got buffed already that made it less of a cookie clicker pain. I've noticed in most rounds that xenos get wiped due to getting pushed back through open ground since it's way too labor intensive to preemptively build a small maze on the retreat path if you're not really guaranteed to use it should marines just die instead of pushing the xenos into it. 

>So the thing that actually makes mazing CBT is how is done: As drone you go, build a resin wall (click), move to the next postion (key presses) and repeat many, many, many times and is not that much better to do as hivelord as you are limited to the area next to you.

I agree, but as I said in discord, the problem is both the labor intensive aspect of it and the time aspect of it. I can't fix the labor intensive aspect of WASDing and right clicking, but I can at least change the time aspect. I have absolutely zero idea how to translate your suggestion into code format, for example.

>I think xeno structures as they stand are too weak already. marines can trivialy unga ball trough even wack as mazes using 1 flamer.

Ungaball is an issue of its own outside the scope of the pr. Flamer already busting through mazes is part of the reason I figured it was a good time to buff the build time of said mazes. The starting wall health was only brought down that low because a cheeky drone could wall off an entire marine push withot them being able to do much had it been buffed without the lower health. For instance, you need 3 hits of a machete to bring down a just made wall at 150hp which take 1 second per hit. By the time you brought down the first wall, a drone already made 3 more. If it had been a hivelord, you'd need 5 hits. Maybe one more hit for both if you were unlucky enough to have seen the wall late and it had a bit of time to increase in max health. A buff in time to that would be pretty unfun as pushes would just get walled in.

Losenis:
An idea is allowing rangebuilding walls like the hivelord can so that the whole "have to move 1 tile every single time you want to build something", but that'd rest on the starting health being low so that it can't be abused midcombat. For what QoL it may be it'll still be quite the balance change if it'll enable drones to build a wall behind a cade to start aciding it, unless the wall is too weak to actually shield it from the turret or marine bullets.

Losenis:
>Just so we're clear, this would make drone walls barely stronger than a larva. Barely.

This is intended to avoid the shenanigans of spamming walls at the front to wall marines in or other cases outside of proper mazebuilding, like what led to the pr that disabled aciding cades after building a wall on top of yourself. In exchange the maturity rate was increased so that they still reach 300hp in around 250 seconds much like the old walls. Due to rounding it was made so that they mature in like 240 seconds or so but not that big of a difference. If the wall didn't get shot for 4 minutes, it should have 300hp just like old walls.

Al-1ce:
>Only vs melee, walls have some wild ranged armor.


When it comes down to it, what little range armor walls have is rendered irrelevant when the overall health is miniscule. Even if walls were given more ranged defense, knives would still 2-3 hit them. Nearly all LMGs with a bayonet would instakill fresh walls.


>If the wall didn't get shot for 4 minutes, it should have 300hp just like old walls.


That's an ideal situation. Marines will absolutely hold the advantage once a maze has been overwhelmed. Rebuilding becomes irrelevant when the health is so miniscule most guns will shred whatever you put down in a shot or two. This will lead to a scenario where the overall push cannot be held back as walls aren't given enough time to mature. It's a negative feedback loop and results in more stress for marines and xenos alike.

Losenis:
What would you suggest to have walling in pushes and building next to the cades while actively getting shot not be effective while also avoiding the problems you say?

Al-1ce:
> What would you suggest to have walling in pushes and building next to the cades while actively getting shot not be effective while also avoiding the problems you say?


Honestly? By not changing anything. With some finesse and a little bit of luck, that's already possible. But xenos rarely tend to offensively build. My main concern here is that walls are not going to have enough time to mature and will be too easy to destroy, encouraging marines to _not_ let them mature.

Losenis:
Very well. Hopefully a TM shows if it ends up exploitable or not. If it does, an hp lessening or some other thing would be needed.

Al-1ce:
> Very well. Hopefully a TM shows if it ends up exploitable or not. If it does, an hp lessening or some other thing would be needed.

Regardless, the things I've pointed out are little more than speculation given off from what I can read in the code and what you've described. I still think this would be valuable to TM.

BraveMole:
nop

PR #10431, opened 2022-06-10
Explosives can no longer be attached to unmanned vehicles

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
Plastic explosives have devastation range. These will instakill any mobs that fail to meet the bomb armor check. You can attach one such explosive to unmanned vehicles in general. Instakills in general are _terrible_, hence the PR.

## Changelog
:cl: Lewdcifer
balance: Unmanned vehicles no longer accept explosive attachments.
/:cl:

### DIFF
--- a/code/modules/vehicles/unmanned/unmanned_vehicle.dm
+++ b/code/modules/vehicles/unmanned/unmanned_vehicle.dm
@@ -92,8 +92,6 @@
 			. += image('icons/obj/unmanned_vehicles.dmi', src, "heavy_cannon")
 		if(TURRET_TYPE_LIGHT)
 			. += image('icons/obj/unmanned_vehicles.dmi', src, "light_cannon")
-		if(TURRET_TYPE_EXPLOSIVE)
-			. += image('icons/obj/unmanned_vehicles.dmi', src, "bomb")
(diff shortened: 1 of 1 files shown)


novaepee:
What if we just nerf C4 and let unmanned vehicles jihad to their heart's content?

lKiseki:
> What if we just nerf C4 and let unmanned vehicles jihad to their heart's content?

You effectively kill C4's niche, which is probably not a good idea. It's better to target the problem itself, since C4 is alright.

Derrick2Electricboogaloo:
I smell salt

FosterGrime-Heymynameslouie:
coupled with the fact that the light uav thats standard issue for engie's is, well, light, and does in a few acid hits, and slash hits, emits a near blinding light, and is extremely loud which can make it be heard from offscreen. I think if you die to uav suicide c4, you probably deserve it.

InterroLouis:
C4? Not detpacks?

Pariah919:
mfw no VBIED

TiviPlus (changes requested):
Make it so scout droid cant take explosives instead or change the checks to take expllosive as a turret type

TiviPlus:
Also ided

PR #10436, opened 2022-06-10
Reduce plasma cost of all wraith abilities

BraveMole:
## About The Pull Request

Blink from 50 to 30
Banish from 100 to 50
time stop from 150 to 100
protal from 100 to 50

## Why It's Good For The Game

Wraith is always starving for plasma, which cripples the caste. It relies entirely on abilities since it's very squishy with little damage, it needs more plasma

## Changelog
:cl:
balance: Reduce plasma cost of all wraith abilities
/:cl:


Naaanii:
I think giving wraith a larger plasma pool to work with would be nicer.

PR #10440, opened 2022-06-11
Fix jetpack flying while buckled

Wayland-Smithy:
## About The Pull Request

No more wasted fuel while buckled to big things, and no more taking small buckled things along for a free ride.

## Why It's Good For The Game

Fix good.

## Changelog
:cl:
fix: Fixed jetpack flying while buckled.
/:cl:

### DIFF
--- a/code/game/objects/items/jetpack.dm
+++ b/code/game/objects/items/jetpack.dm
@@ -89,6 +89,9 @@
 
 ///Make the user fly toward the target atom
 /obj/item/jetpack_marine/proc/use_jetpack(atom/A, mob/living/carbon/human/human_user)
+	if(human_user.buckled)
+		balloon_alert(human_user, "Cannot fly while buckled")
+		return
 	if(!do_after(user = human_user, delay = 0.3 SECONDS, needhand = FALSE, target = A, ignore_turf_checks = TRUE))
 		return
 	TIMER_COOLDOWN_START(src, COOLDOWN_JETPACK, JETPACK_COOLDOWN_TIME)


TiviPlus (commented):
Shouldnt this be checking the immobilized and so traits instead

Wayland-Smithy:
Not sure. Is it intended that jetpack is able to lift not only the wearer, but also what they are pulling **and** whatever they happen to be riding? If so maybe a fuel and/or distance penalty should be considered for those "hauling" cases.

I already think it is silly that pulling even works because whatever is behind the jetpack would be caught in the jet blast, so taking with whatever you are strapped to that is presumably, at least partially, blocking the jet nozzles is even more whacky.

BraveMole:
> Not sure. Is it intended that jetpack is able to lift not only the wearer, but also what they are pulling **and** whatever they happen to be riding? If so maybe a fuel and/or distance penalty should be considered for those "hauling" cases.

Don't think that was the question. Isn't there a better way of doing that, ie use immobilize or traits, which would cover more cases

TiviPlus:
^^

Wayland-Smithy:
> Isn't there a better way of doing that, ie use immobilize or traits, which would cover more cases

I'd hoped you all could tell me seeing as I have less codebase knowledge.

After some testing the answer seems to be no that is not an option.

dreamseeker_NPSg8DgEiq
While in the loader traits and status effects looks like this:
image

The effect is the same if you are buckled into the Alamo, but this time have the buckled trait for some reason:
image

For roller beds and surgery tables this is not an issue since you must be upright to use the jetpack otherwise I bet it would be.

BraveMole (approved):

PR #10442, opened 2022-06-11
ZX-76 no longer takes masterkey

Wayland-Smithy:
## About The Pull Request

Xzibit would be disappointed.

## Why It's Good For The Game

When converted to a triple barrel it is capable of 1 tapping tier two xenos (or in a vacuum get 99% of the way there).

I think all other shotguns can't have this for a reason?

Please hold all comments along the lines of "it's a req weapon so walance does not matter" because when firing 3x volleys of free flech ammo the DPS rivals that of rockets at medium range.

## Changelog
:cl:
balance: ZX-76 no longer accepts the masterkey attachment.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -653,7 +653,6 @@
 		/obj/item/attachable/magnetic_harness,
 		/obj/item/attachable/lasersight,
 		/obj/item/weapon/gun/flamer/mini_flamer,
-		/obj/item/weapon/gun/shotgun/combat/masterkey,
 		/obj/item/weapon/gun/grenade_launcher/underslung,
 	)
 


Lumipharon:
Don't you dare remove the based triple barrel, it was specifically left out of the great masterkey purge.

Al-1ce:
> When converted to a triple barrel it is capable of 1 tapping tier two xenos.

I tested this myself a couple hours ago. Not the case.

Dolth:
It doesn't even crit an ancient drone right in front of them and it's a req gun, by definition it's meant to be stronger. ROF is low.
Also the fact you're complaining about burst of damage but just wanna remove masterkey is just a proof of this being a salt PR (UGL is even stronger than masterkey, just more expensive)

Araulius:
> > When converted to a triple barrel it is capable of 1 tapping tier two xenos.
> 
> I tested this myself a couple hours ago. Not the case.
not enough apparently. was at full health as a primo hive lord. was one tapped.

Frinid1:
ZX is unused double shot burst shotgun with elephant size price (100 points), which doesn't work on any t1-t2 xeno except hivelord because they're knocked down from first shot
Also without mastekey it almost oneshots him in perfect melee pb shot with bayonet anyway
изображение

Dolth:
> > > When converted to a triple barrel it is capable of 1 tapping tier two xenos.
> > 
> > 
> > I tested this myself a couple hours ago. Not the case.
> 
> not enough apparently. was at full health as a primo hive lord. was one tapped. by dolthgar specificly.
> 
> edit: instead of removing the underbarrel, it should have a longer fire delay. At the moment, it does this: https://youtu.be/eMeyflHGWrg

I am using flechette with the hivelord and flechette doesn't kill with one PB and one masterkey shot pointblank so no, I did not onetap you (unless you weren't full health).

Anyway you're playing hivelord which isn't really tanky, shotguns kill you in two PBs aswell so it's kind of the same. Except ZX is a requisition gun that costs 100 points, if it wasn't better than already existing guns it would be useless to buy it.

Araulius:
https://youtu.be/G3XcnhRp2rA
https://youtu.be/hq2BO8KM04g

here is two more tier 2 benos being bursted down to no health in less than 3 seconds.  mind you, these are all primo.
this is acceptable for sadar, which has a cost associated with its ammo. this is not acceptable.

Dolth:
Insulting won't make your point any more relevant. Shotguns already kill in less than 3 seconds if you constantly PB just like you have shown in your examples. This is nothing new, just slightly stronger because it's requisition.

Losenis:
>with elephant size price (100 points)

Where were you when I made this same pr and also tried to reduce the cost of the gun?

Dolth:
As comparison with T35; TTK is the same on a primordial carrier and damage are slightly bigger since no crit but do keep in mind ZX is a requisition weapon AND uses a ton of ammo and is longer to reload. 
https://i.gyazo.com/dc22b73d0c3bd809a1d122620b27a494.mp4

Losenis:
TTK on an upfront burst damage weapon is misleading if not dishonest. It's two guaranteed shots on the first click for a PB and actual gameplay isn't a series of 1v1s, it's ungaballs where xenos won't be at full health because bullets will be flying around a lot. If the shotgun PBd something, 99% of the time that something is also getting shot by something else. It's how it can effectively 1hit a HL, because the scenario where it doesn't specifically needs the HL to be at full health and the shotgunner to be completely alone with no other marines shooting.

Dolth:
Without masterkey it's barely any stronger than T35 and it's weaker than DB apparently due to the ROF : https://gyazo.com/74eb1813e556869b03e6595b4bcea569
https://i.gyazo.com/8bdbe5b5dc18b552282476a804175bb6.mp4
I get it, it's frustrating to take a lot of damage in just a single click but the same applies to marines when they get tossed into the hive or stunlocked.

MiniMeatwad:
I play xeno and I haven't died to it. Honestly don't mind this req weapon.

Wayland-Smithy:
> the same applies to marines when they get tossed into the hive or stunlocked

except marines which are already the larger force can easily revive the dead in almost all situations excepting wipes or dumb shot callers sending them into a death choke

even at range it is putting out ~230 post mitigation (armor) damage in a single tick with flech so that's like 2000 DPS cyclical
so it's basically a short/medium range rocket tag weapon with much less expensive ammo
image

Dolth:
> > the same applies to marines when they get tossed into the hive or stunlocked
> 
> except marines which are already the larger force can easily revive the dead in almost all situations excepting wipes or dumb shot callers sending them into a death choke
> 
> even at range it is putting out ~230 post mitigation (armor) damage in a single tick with flech so that's like 2000 DPS cyclical so it's basically a short/medium range rocket tag weapon with free ammo image

Yes that's on an average the same as other requisition guns but it requires to be CQC. Without masterkey, as my videos above proved, it's literally just a DB with more ammo. Also, another proof that another REQ gun is as dangerous with masterkey; https://i.gyazo.com/1ab9aa3f1241444c38975197ef823000.mp4

MiniMeatwad:
If you shoot a xeno with a forced movement weapon like ZX, unless they're against the wall they will get pushed away. That puts them out of CQC, and also there's a quirk with the ZX. If you shoot a xeno and they're against a wall (don't move) they absorb every shot. If they get pushed, they don't absorb every shot.

I tested it against a spitter and it's really not that much more dangerous than normal shotguns, except it's a req weapon so it's easier to use.

Wayland-Smithy:
3 tiles away is "close quarters"? Melee xenos can't CQC from that far away.

> https://i.gyazo.com/1ab9aa3f1241444c38975197ef823000.mp4

That second weapon in the video clearly has less DPS.

> If you shoot a xeno with a forced movement weapon like ZX, unless they're against the wall they will get pushed away.

Flech does not knockback, so there is no change in range or accuracy for all 3 shells.

Dolth:
> > https://i.gyazo.com/1ab9aa3f1241444c38975197ef823000.mp4
> 
> That second weapon in the video clearly has less DPS.

Why shouldn't it have less DPS than a requisition shotgun that lands 3 PB bayonnet in a row? The point is, without using any underbarrel it kills in almost as much time as a shotgun that constantly PBs. Remove the masterkey and it won't be anything more than a DB with more ammo, won't be worth using.

Wayland-Smithy:
Dolth makes some good points. I think under barrels just need to apply the fire delay to the parent gun and only use the attachments value if it is higher.

Dolth:
Well, if you do that some guns will have up to 2 seconds delay for their UB to use, but some rifles just 0.2 seconds. So let's say a laser sniper rifle won't be able to emergency fuck-off with a masterkey, but a rifle will be able to burst damage with a flamer.
Adding a flat time to any UB seems to me a more fair idea imo.

Al-1ce:
> Well, if you do that some guns will have up to 2 seconds delay for their UB to use, but some rifles just 0.2 seconds. So let's say a laser sniper rifle won't be able to emergency fuck-off with a UBF, but a rifle will be able to burst damage with a flamer. Adding a flat time to any UB seems to me a more fair idea imo.

That's a better idea, yes. If all guns are given an equal delay to their respective class (SMGs, rifles, and in this case - shotguns) it would be for the better.

TiviPlus:
ZX really doesnt need this

PR #10446, opened 2022-06-11
May fix bullet ping not being removed

BraveMole:
## About The Pull Request

I'm guessing they staying because two pings are being created on the same tile at the same price. 

Just a guess though, TM first

## Changelog
:cl:
fix: Fix bullet ping not being removed
/:cl:

### DIFF
--- a/code/__DEFINES/equipment.dm
+++ b/code/__DEFINES/equipment.dm
@@ -32,6 +32,8 @@
 #define BUMP_ATTACKABLE (1<<15)
 ///This atom will not be qdeled when a shuttle lands on it; it will just move onto the shuttle tile. It will stay on the ground when the shuttle takes off
 #define SHUTTLE_IMMUNE (1<<16)
+/// If this atom has a bullet ping on it
+#define IS_BULLET_PINGED (1<<17)
 /// Should we use the initial icon for display? Mostly used by overlay only objects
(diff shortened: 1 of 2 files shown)


Lumipharon:
Doesn't this just stop a single turf from having more than one ping on it at a time?

Worth a try either way I suppose, but if the root cause is causing issues like shield recharge timer and burst fire timers getting fucked, I'm a bit doubtful.

BraveMole:
> Doesn't this just stop a single turf from having more than one ping on it at a time?
> 
> Worth a try either way I suppose, but if the root cause is causing issues like shield recharge timer and burst fire timers getting fucked, I'm a bit doubtful.

No timers are not the issue. Arrowhead bug was not a timer bug, and burst doesnt even use timer

Lumipharon:
> > Doesn't this just stop a single turf from having more than one ping on it at a time?
> > Worth a try either way I suppose, but if the root cause is causing issues like shield recharge timer and burst fire timers getting fucked, I'm a bit doubtful.
> 
> No timers are not the issue. Arrowhead bug was not a timer bug, and burst doesnt even use timer

I know it's not a issue specifically with timers, but they are all timed procs/thingos (start recahrging for arrowhead, end burst mode for burst fire, delete overlay for bullet pings). I'd love to be wrong tho

PR #10447, opened 2022-06-11
Under barrels share last_fired with their master_gun

Wayland-Smithy:
## About The Pull Request

Prevents UBs from being fired at the exact same time as their parent gun.

Closes #10442

## Why It's Good For The Game

Being able to shoot 2 different ammo types or an additional volley of the same type as is the case with the ZX in a single tick is pretty cheesy.

## Changelog
:cl:
balance: Under barrels share last_fired with their master_gun.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -744,6 +744,12 @@
 		return
 
 	last_fired = world.time
+	if(master_gun)
+		master_gun.last_fired = world.time
+	if(attachments_by_slot[ATTACHMENT_SLOT_UNDER])
+		var/obj/item/weapon/gun/under = attachments_by_slot[ATTACHMENT_SLOT_UNDER]
+		under.last_fired = world.time
+
 	SEND_SIGNAL(src, COMSIG_MOB_GUN_FIRED, target, src)
 
 	if(!max_chamber_items)


Naaanii:
Muh plasma pistol autoshottie...

InterroLouis:
F in chat for anyone ever using masterkey

Losenis:
Honestly this might actually allow masterkey shotties to be reenabled if one wants to effectively bind two ammo types to fire and alt-fire

breadhunt:
This might be better off as a flag or var defaulting to enabled; we had it on all weapons before and it fucked up a few cool combos

Lumipharon:
This psuedo removes it from any gun with slower rof, like the ZX, as there's actively no benefit from using it except for alt damage effects like incendiary slugs of whatever.

Wayland-Smithy:
> This might be better off as a flag or var defaulting to enabled; we had it on all weapons before and it fucked up a few cool combos. See what Dolth said on the other PR

A var on the master_gun or the attachments?

I'm also considering using the `COMSIG_MOB_GUN_FIRED` signal which would also allow attachments to sync their `last_fired` when the master_gun shoots removing the assumption that shooty attachments will only ever be at `ATTACHMENT_SLOT_UNDER`

Either way that means going through each gun & attachment combo which I do not want to do for anti-cheese purposes, but I am willing to implement if someone can provide a list.

Dolth:
Basically if your share fire_delay with the parent gun you're gonna have to wait 2 seconds to shoot the underbarrel where a rifle will have to wait 0.2 seconds. I suggested having the underbarrel own fire_delay bound to the parent gun. (If UGL is 1 second delay, then you need to wait 1 second after you shot with your gun, regardless of what it is. Hope that's clear :(

Pariah919 (commented):
Why not just make this its own stat for underbarreled guns? I.E delay when master gun fired or something.

SplinterGP:
> Why not just make this its own stat for underbarreled guns? I.E delay when master gun fired or something.

above would be nice, or also make a stat like "Extra delay before parent weapon can fire again", i dunno how this goes for some guns but i was adding some QOL stuff like right click shooting stuff not stopping autofire

BraveMole (changes requested):
Muh plasma pistol autoshottie...

More seriously, the idea makes sense, not a fan of the execution. I'd rather have a fixed delay between master and under, or something related to a var on the attachment

code/modules/projectiles/gun_system.dm
```diff
@@ -744,6 +744,12 @@
 		return
 
 	last_fired = world.time
+	if(master_gun)
+		master_gun.last_fired = world.time
```
BraveMole, line 748:
```suggestion
		master_gun.last_fired = max(world.time - master_gun.fire_delay + a_avariabl_on_attachment, master_gun.last_fired)  ?
```
Not tested, but here's an idea

Wayland-Smithy:
Offsetting `last_fired` on its own could work, and Mole's suggestions makes sense, but it could also introduce problems down the line with logic that might assume it is `=< world.time` as has always been the case. A more comprehensive approach would probably be for the best including some attach signals for maximum compatibility.

PR #10448, opened 2022-06-11
Forward charge numbers change but i labelled it as a rework which means tivi was dissapointed when opening the pr to review it

Losenis:
## About The Pull Request
Reduces Fordward Charge's stun duration to be equal to tail sweep's. Increases range to 6 (1 tile before edge of screen). Reduces windup to 0. Increases cooldown to 15.

## Why It's Good For The Game
As rewarding as hitting a 4 second long stun may be, stuns aren't fun to play against, and both sides should have fun fighting each other. Reworks the defender charge from a stun that is predominantly used midcombat into an initiation/escape tool. This'll let the defender close in a large gap into combat or be able to escape combat safely, given that it's a slow boy and its ability that grants more armor just makes it even slower resulting in the chase being either "Get killed quicker with less armor" or "Stay in the gunline for longer". Given the ability to escape combat, it shold allow defenders to play it a little bit riskier and remain in combat a bit longer, since even though 1v1 the stun did the same, marines tend to be in groups. Also lets it survive surprise freak scenarios of walking into marines already running towards its general direction from offscreen.

Cooldown is extended to 15 seconds so that it can't actually both initiate AND escape an engagement unless it's one that has dragged on for a while.

## Changelog
:cl:
balance: Defender's Forward Charge stun reduced on par with tail sweep's. Windup down to 0, range to 6 from 4, cooldown to 15 from 10.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defender/abilities_defender.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defender/abilities_defender.dm
@@ -87,16 +87,16 @@
 /datum/action/xeno_action/activable/forward_charge
 	name = "Forward Charge"
 	action_icon_state = "charge"
-	mechanics_text = "Charge up to 4 tiles and knockdown any targets in our way."
+	mechanics_text = "Charge up to 6 tiles and knockdown any targets in our way."
 	ability_name = "charge"
-	cooldown_timer = 10 SECONDS
+	cooldown_timer = 15 SECONDS
 	plasma_cost = 80
 	use_state_flags = XACT_USE_CRESTED|XACT_USE_FORTIFIED
 	keybind_signal = COMSIG_XENOABILITY_FORWARD_CHARGE
 	///How far can we charge
-	var/range = 4
+	var/range = 6
 	///How long is the windup before charging
-	var/windup_time = 0.5 SECONDS
+	var/windup_time = 0 SECONDS
 
 /datum/action/xeno_action/activable/forward_charge/proc/charge_complete()
 	SIGNAL_HANDLER
--- a/code/modules/mob/living/carbon/xenomorph/castes/defender/defender.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defender/defender.dm
@@ -58,7 +58,7 @@
 	var/target_turf = get_step_away(src, H, rand(1, 2)) //This is where we blast our target
 	target_turf = get_step_rand(target_turf) //Scatter
 	H.throw_at(get_turf(target_turf), 4, 70, H)
-	H.Paralyze(40)
+	H.Paralyze(5)
 
 
 /mob/living/carbon/xenomorph/defender/lay_down()


Al-1ce:
The tail sweep was intended as an escape tool, whereas forwards-charge was not. I wouldn't be against lowering the stun by a second or two but making it a cookie-cutter variation of another move isn't good for gameplay as well.

It may also be of note that charge had a warning symbol which was put there for a reason. Reducing the windup to null negates that entirely.

If you're intending on removing especially cancerous stuns, I'd be looking into bull before the defender.

Losenis:
First time I hear of the tail sweep being the escape tool, and if it is, it's pretty bad at it. Most it does for escaping is just unwield the gun for the gun that'll immediately start shooting anyway.

>cookie-cutter variation of another move

The knockdown is literally the only similarity it'd have, though. One is a ranged charge and the other is an AoE knockdown.

>It may also be of note that charge had a warning symbol which was put there for a reason. Reducing the windup to null negates that entirely.

Yeah I talked about that in the discord with another guy that thought the same. The windup and hazard icon were put there because it was still a 4 second long stun, having it occur with zero windup or warning from a xeno that only needed to click you is as gruellingly unfun as it gets. The stun is effectively removed though, and keeping a windup would actually risk messing up with how it's suppposed to go because I envison it being used by defenders moving and spamclicking to get away. A windup even if small has the chance of just getting BYOND'd in that scneario.

>If you're intending on removing especially cancerous stuns, I'd be looking into bull before the defender.

I'll think about it, but bull already has a pretty strict setup and can't really be chained in the middle of a fight where other xenos are swarming a guy so easily, on top of the other charges already having what the stun would give. If you have any ideas, I'd like to hear them.

Lumipharon:
6 is really long range for the slow tank caste.

Al-1ce:
> The stun is effectively removed though

All castes utilize stuns to some capacity.  Stuns in the current meta are one of the only ways xenomorphs can deliver any fatal amounts of damage without themselves dying. Xenomorph health pools aren't regularly large enough for sustained combat. Even a ravager can only stick in the fray for a minute or two at most.

>The knockdown is literally the only similarity it'd have, though. One is a ranged charge and the other is an AoE knockdown.

They are nearly identical abilities, minding the _very_ short amount of range (which is not AoE) FWC is given.

You also intend make them linear in stun-time as well. 

>First time I hear of the tail sweep being the escape tool, and if it is, it's pretty bad at it. Most it does for escaping is just unwield the gun for the guy that'll immediately start shooting anyway.

I hope you're aware that making Forwards Charge have a similarly useless stun-effect will only repeat this problem.

Losenis:
>They are nearly identical abilities
image

>minding the very short amount of range

Almost the screen's edge in one tick is short range? What? We're talking about what this PR would turn the ability into rather than what it already is, right? If it was any more range than that it'd be mechanically impossible to use without giving defender a farther zoom for default view.

And if we're talking about how it currently is, I don't see the point given it's being changed into an actual farther range that's more reliable to use due to no windup to get interrupted by anything.

Al-1ce:
>They are nearly identical abilities

Currently, ..yes. The charge does not have immediate movement. The tail is AoE, I thought you were referring to the charge as having AoE, I'll clarify that.

>Almost the screen's edge in one tick is short range? What?

I play on widescreen. Your edge of the screen is.. well, not what mine is. 

>If it was any more range than that it'd be mechanically impossible to use without giving defender a farther zoom for default view.

I didn't mention anything about extending the range, only that it wasn't that much. That's fine. It shouldn't be excessive. I'm not asking you to make a PR to extend the range, I'd do that myself if I felt it necessary. The distance the Defender's abilities can reach is finite, yes, but the vision cone of the player can be extended beyond that.
https://youtu.be/DLacLDv-L2A

> And if we're talking about how it currently is, I don't see the point given it's being changed into an actual farther range that's more reliable to use due to no windup to get interrupted by anything.

The windup and the stun are counterbalancing one-another. Foward Charge is difficult to land on a moving opponent, the range and distance isn't that big and the slow windup usually allows for marines to escape. That's a _**good**_ thing. It's something the player has to account for, and not something which should be more easy "reliable" to use. It's a high risk, high reward offensive tool. Removing the stun makes it little more than a ranged tail-sweep. Which isn't saying much.


```
Most it does for escaping is just unwield the gun for the guy that'll immediately start shooting anyway.

```

You mentioned this earlier when referring to tail sweep. I feel the need to ask again, have you considered that this would also apply if the charge had no stun?

Losenis:
>have you considered that this would also apply if the charge had no stun

Considering its usage as escape tool would be charging away form the marine rather than hitting him, with a 6 tile in one tick travel distance into what is not really going to be 100% open flat ground but rather a corner of a shotblocker or edge of maze, yeah it'd be a pretty big change. Hitting the marine is no longer the point of the ability, it's the instant travel time to a point be it into a vulnerable marine or away from a threat.

>Having to time it properly, and as a marine being able to dodge it is fun.

Only in a 1v1. When xenos fight in groups, you're going to get hit by it while the marine is either bodyblocked or already stunned, which isn't fun when there's potentially 2 or 3 defenders since this is still a 4 second stun on a limitless T1 caste. That's the whole point behind reducing the number of stuns around, they're all fine on paper but with so many xenos around, it quickly turns into "pile more stuns the already stunned marine" while any shortcomings the xeno may have get bandaided with "just stun the marine, bro" like what you might see in deadpop crash when there aren't enough xenos to stunlock marines.

Dolth:
> The windup is crucial to that ability. Having to time it properly, and as a marine being able to dodge it is fun.

No windup doesn't mean you can't dodge. People still dodge pounces and ravager charges and crushers and boiler globs and bulls.

Tanya you're removing the stun in exchange of windup why do you feel like raising the CC ontop of it? I understand *having* to stun is aids for xeno gameplay and to be stunned 5 seconds as a marine aswell but I think the beno still need his tools to play a bit.

Mechanelize:
The stuns are what make defender capable of combat, the charge is easy to dodge and sweep for following up after that. Stun combat is how the game works, giving it bootleg pounce without the pounce is a straight up nerf.

Losenis:
"The stuns are what make defender capable of combat"

IE the defender is not capable of combat and needs marines to be locked out of moving, shooting, breathing, coping, and seething for it to be able to? Yeah that's kind of the issue I've seen with lots of stun-centric castes.

Losenis:
I don't see how a windup will have you dodge an ability that may not even be aimed at you or you may be physically incapable of dodging at all with bodyblock/other stuns, but the reason I removed the windup was also because it might conflict with using it while moving or getting shuffled around. BYOND isn't the most reliable of engines.

Losenis:
I had also talked it out with Zack time ago, eventually he said it was fun because even rav has a no windup charge of sorts, so there already is some precedent there.

Mechanelize:
> Yeah that's kind of the issue I've seen with lots of stun-centric castes.

Relying on stuns is an issue for castes that have stun abilities? Relying on your abilities is just how the game works.

Yeah stuns aren't fun, it's called PUNISHMENT, stuns require you to get somewhat close to initiate and stay close to keep dealing damage, removing stuns hurts the caste and let's marines get away with not maintaining distance.

SplinterGP:
> I had also talked it out with Zack time ago, eventually he said it was fine because even rav has a no windup charge of sorts, so there already is some precedent there.

Ravager is a relatively fast T3 xeno, Defender is a slow tanky boi that is a T1, their skills are slow because of that and the windup plays part of that on the charge.

EvilMelonTheSecond:
> > The windup plays part of that on the charge
> 
> The windup is there because otherwise it'd be a monstrosity that can instaclick you into a 4s stun that'd be stupid unfun to play against that can be chained into a second defender or whatever other group of xenos is also around. If you have ideas for a change that'd be thematic to the slowboy, I'm open to suggestions, I just want the stuns to not be so excessive because even though on paper it's fine it just leads to stunlock central when there's more than 1 xeno with stuns around. I don't aim to remove all stuns from the game, and ideally the xenos get something in return too.

Add back the windup and add either slowdown or damage to the effects. Adding damage makes it feel like a mini crusher which is funni but major slowdown is probably a better effect. Don't ever add stagger though.

Another idea is lowering cooldown massively and having the only limit be the plasma cost, making defender the T1 CC caste that focuses on disrupting enemy firepower(one might even say a "defender" caste... whoa)

Losenis:
>Don't ever add stagger though.

Huh? How come? For xenos it's understandable since a staggered xeno straight up can't use abilities at all but for marines it's only a 50% damage reduction for a duration, and I don't think it affects sunder.

PR #10449, opened 2022-06-11
Fix organ damage not being treated by surgery

BraveMole:
## About The Pull Request

Title

## Changelog
:cl:
fix: Fix organ damage not being treated by surgery
/:cl:


Al-1ce:
At the time, when this isn't p merged, the autodoc is the only thing which can solve lung or heart damage. I'd imagine that has more to do with the rejuvination flag for its method of healing, which seems to be fine.

PR #10452, opened 2022-06-12
Removes Hacking

Arctrooper8:
## About The Pull Request

Removes hacking from all vendors.

## Why It's Good For The Game

Hacking allows marines to access items they shouldn't have or steal limited items from jobs which use or even need them. 
This isn't a full removal of the actual hacking code, but it does prevent hacking from happening in game. Obviously if hacking is removed some items should be moved around and some vendor access rights changed. 
Discussion is welcome and appreciated.

## Changelog
:cl:
del: You can no longer hack vendors
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/vending.dm
+++ b/code/game/objects/machinery/vending/vending.dm
@@ -151,7 +151,7 @@
 	///How much tipped we are.
 	var/tipped_level = 0
 	///Stops the machine from being hacked to shoot inventory or allow all access
-	var/hacking_safety = FALSE
+	var/hacking_safety = TRUE
 
 	var/scan_id = TRUE
 


BraveMole (changes requested):
remove it properly please, don't leave dead code

Losenis:
>Play a role that has access to said vendor
>Get the entire server hounding me to dispense them harder than they hound req for modules and imagers
>New meta is invalidating the vendors by just dispensing every single thing they have onto the floor

Thanks

Derrick2Electricboogaloo:
Sad, I will miss hacking.

novaepee:
People are going to break into doors, which will lead to admins stepping in and telling people they have to ask AI or people with access for said door.

Mrrpip:
This is dumb, it removes all remaining incentive for non medics to get the ground medbay online.

Cryotube can be used trough showers that are close by on all maps, vendors can't be used by non-medics, and surgery can be done on roller beds/tad.

The limb printer -could be- useful groundside, though you can just preprint shipside before the drop if you wanted to do healbus or something.

Secondly, I do believe you can grab gear from the shipside variant even when unhacked trough the use of loadouts, thus it's redundant at best.

The conclusion I see, it kneecaps the creativity of normal players groundside, while the shipside powergamers/hooligans/etc. Won't be affected much
(They'll just steal the table belts anyhow/use loadouts as mentioned previously)

Pariah919:
This is daft and breaks lowpop crash and nuke war by making it impossible to open APCs without a engi.

Mrrpip:
> This is daft and breaks lowpop crash and nuke war by making it impossible to open groundside Vendors without a engi.

This one only affects vending machines.

Arctrooper8:
> remove it properly please, don't leave dead code

Will do

Al-1ce:
Wouldn't it be nice if these items could be accessed without hacking the machine? I find it nuts that nobody has considered this a valid solution to the problem. And the removal of hacking would definitely encourage marines to blow down whatever doors/walls they need to get what they want. That's bad.

Autisem:
Much better solution already would be to only have this restriction on shipside vendors, groundside vendors dont need to be unhackable.

Though in general low pop times is an issue that this wont solve

goose4429:
As the guy who moved some of the stuff to restricted vendors, my position is that you can always just ask the respective ship crew for them. If that fails, the crew person is either asshole or there is none. Any way this is the only time when you will want to hack, and it seems like hacking is a good thing in this case.

TiviPlus:
Hacking is needed for things like low pop

If you want to do it properly use skeleton crew that checks for jobs present or so

See https://discord.com/channels/498569646014857217/500801682058379265/985870392981880872

PR #10456, opened 2022-06-13
Hyro, rye and synap now have lemoline in their recipes

BraveMole:
## About The Pull Request

Title.

## Why It's Good For The Game

Hyro, better dylo with no downside
Rye, better hyper with no downside
Synap, almost-infinite stamina with no downside

They should either not exist, or have some cost added to them

## Changelog
:cl:
balance: Hyro, rye and synap now have lemoline in their recipes
/:cl:

### DIFF
--- a/code/modules/reagents/reactions/medical.dm
+++ b/code/modules/reagents/reactions/medical.dm
@@ -38,7 +38,7 @@
 /datum/chemical_reaction/ryetalyn
 	name = "Ryetalyn"
 	results = list(/datum/reagent/medicine/ryetalyn = 2)
-	required_reagents = list(/datum/reagent/medicine/arithrazine = 1, /datum/reagent/carbon = 1)
+	required_reagents = list(/datum/reagent/medicine/arithrazine = 1, /datum/reagent/carbon = 1, /datum/reagent/lemoline = 1)
 
 /datum/chemical_reaction/cryoxadone
 	name = "Cryoxadone"
@@ -116,7 +116,7 @@
 /datum/chemical_reaction/synaptizine
 	name = "Synaptizine"
 	results = list(/datum/reagent/medicine/synaptizine = 3)
-	required_reagents = list(/datum/reagent/consumable/sugar = 1, /datum/reagent/lithium = 1, /datum/reagent/water = 1)
+	required_reagents = list(/datum/reagent/consumable/sugar = 1, /datum/reagent/lithium = 1, /datum/reagent/water = 1, /datum/reagent/lemoline = 1)
 
 /datum/chemical_reaction/leporazine
 	name = "Leporazine"
@@ -127,7 +127,7 @@
 /datum/chemical_reaction/hyronalin
 	name = "Hyronalin"
 	results = list(/datum/reagent/medicine/hyronalin = 2)
-	required_reagents = list(/datum/reagent/radium = 1, /datum/reagent/medicine/dylovene = 1)
+	required_reagents = list(/datum/reagent/radium = 1, /datum/reagent/medicine/dylovene = 1, /datum/reagent/lemoline = 1)
 
 /datum/chemical_reaction/arithrazine
 	name = "Arithrazine"


Frampox:
delete lemoline instead

goose4429:
First of all, add lemoline to recipe = will never see it again in the game, except for two random injectors by CMO which are lying around aimlessly. Players will just consider this a removal.
These chems are mainly required by corpsmen, which do not have insta access to lemo, and will never have it because reqs work for weapons in priority.

That said, if you do, consider making arithrazine a med which is made from dylo+radium, as it's pretty bad by itself. I think it would be somewhat used instead of hyro.

Other remarsk:
Hyro: doesn't purge toxins. Dylo does. So not super-better, but yes.
Rye: Is good, I agree. Though you need a lot of it if you have B/K/T in blood. 
Synap: true. But should probably lower corpsman vendor cost for each, because 4 points for 30 sec of funny stam is just lol. 1 point.

Lumipharon:
Reytalyn need hyro to make, so is double the lemoline. Also rip ever making neuroline, as it needs synap and hyro.

Dunno why you added lemoline to hyro at all actually, it's literally worse than dylo because it doesn't purge. Its only upside is no Stam loss over 10u, but if you're making hyro you can make lower dose dylo.

Losenis:
Hyro and rye don't need lemoline, one is only tox heal without reagent purge of dylov and the other gets purged by the most common drugs in existence that every single marine has.

BraveMole:
> First of all, add lemoline to recipe = will never see it again in the game, except for two random injectors by CMO which are lying around aimlessly. Players will just consider this a removal.

That was that or a removal

> These chems are mainly required by corpsmen, which do not have insta access to lemo, and will never have it because reqs work for weapons in priority.

I see plenty of non-corp using it, and it's not required by corp they can play without. Corp can buy synap for a few points.


> That said, if you do, consider making arithrazine a med which is made from dylo+radium, as it's pretty bad by itself. I think it would be somewhat used instead of hyro

Removing arithrazine in other pr

BraveMole:
And I'm not against rebalancing the corpsman vendor to give access to more chems

goose4429:
> Hyro and rye don't need lemoline, one is only tox heal without reagent purge of dylov and the other gets purged by the most common drugs in existence that every single marine has.

To be fair, rye IS a better version of hyper. It purges 5u and dies if you have B/K/T. Hyper on the other hand purges all your meds in addition to toxins. And if not purged, rye is long-lasting.

But it's important to make it assessible to cpsmen and stuff, otherwise it's basically a removal and unfun.

QualityVan:
It's probably better to just cut hyro entirely than to make the current version use lemoline, although we don't at present have a tox equivalent of mera/derm. Maybe bump the tox heal rate on it?
For rye, even if you're making it lemo-limited I'd suggest dropping the purge rate down to at most 1. Stuff's nuts.

EvilMelonTheSecond:
Hyro is not better dylo. Marines do not need tox healing most of the time and if they do it is likely related to high end effects of neurotox which can't be purged by hyro. Only times I would prefer hyro over dylo is in a chem mix with neuraline or synap in it, in which case the dosage probably wouldn't have been over 10u anyways. Delete the chem instead of lemoline blocking it since it feels like a noobtrap over making meraderm

Rye is not better hyper. Hyper purges ALL your meds which is used by corpsmen to either flush excessive toxins or remove OD effects. A 10u dylo pill is the superior rye as it can be used alongside bkt. As it stands rye is just a utility chem that people think is good from the days neurotox purged meds. Again, delete instead of making it a noobtrap.

I got nothing to say for synap, it is a powergame chem and the effects should probably get put on a cooldown or something. Or as we've seen time and time again, powergamers will find a way to hoard the limited stock to powergame as if nothing changed.

pyronie:
ryetaln- maybe reduce it effects but i don't think it should be nerfed to lemoline requirement as it fills a niche that other chems don't have - that is rewarding for prior preparation, and it sometimes can be useless if the xenos have a lineup with less chem spitters. maybe make it need phoron but not lemoline.
hyrolanin - its pretty good but its weaker dylovene with less drawbacks. definitely shouldn't need lemoline in its use as its basically just bicard or kelotane but for tox.

pyronie:
> ryetaln- maybe reduce it effects but i don't think it should be nerfed to lemoline requirement as it fills a niche that other chems don't have - that is rewarding for prior preparation, and it sometimes can be useless if the xenos have a lineup with less chem spitters. maybe make it need phoron but not lemoline. hyrolanin - its pretty good but its weaker dylovene with less drawbacks. definitely shouldn't need lemoline in its use as its basically just bicard or kelotane but for tox.
for reference the effects of bicard, kelo, and hyronalin next to eachother.
`Hyronalin is a medicinal drug used to counter the effect of poisoning.
Heals 1 TOX.`
`Bicaridine is an analgesic medication and can be used to treat physical trauma.
Heals 1 BRUTE. Causes increasing pain at 10u (+5) and 20u (+15), as well as healing an addition 0.5 BRUTE above 20u.`
`Kelotane is a drug used to treat traumatic burns.
Heals 1 BURN and lowers temperatures to normal. Causes pain (+5) at 10u and heals an additional 0.5 BURN at the cost of more pain (+15) above 20u.`

pyronie:
as for synaptizine, it is a risk vs reward sort of chem. you are not mentioning how it has the obvious downside of only limited stam regen before it starts hurting stam. and due to the chem metabolization rework or whatever its called it cannot be microdosed as effectively anymore. at most you can just have increased stam for a bit, and at worst you can have no stam in the middle of running around enemies.

Weedmaster77:
So Hyronalin needing Lemoline to make is such a joke I don't even think most medics know its a thing and Medical Staff use it with Dylovene for counteracting Polyhexanide when doing ground surgery so taking it and making it hard to make for not being used much and on the lest common damage type is so senseless. 

Then Ryetaln don't see many people use it and its key on zombie round so making it need Lemoline also make no sense 

Synaptizine on the other hand is really easy to make but I don't see it use but by a hand full of people could it take more chems and more time to make sure but make everything need Lemoline is a bad solution to fix this problem.

breadhunt:
> Hyro is not better dylo.  
-dylo blocks stam regen
-the purge rate is pitiful and you're better off just using rye/hyper
-they can just be stacked, that's what makes it better

> if they do it is likely related to high end effects of neurotox which can't be purged by hyro.
taking rye/hyper means you never have to worry about neuro

> Rye is not better hyper.
-standard dosage pill of rye, even with BKT premed, purges the same amount as a standard hyper pill
-it does this without purging normal meds, and potentially lasting a long time if no BKT
-this is powerful because you can keep taking rye to completely shrug off all toxins without having to worry about your premed

> it is a risk vs reward sort of chem.
it's not, at all. you can just take 2u of synap, 4u of hyronalin (or tric+dd, etc.) and have 0 downsides, at all. that's 40 seconds of stun resist, about 20-30 seconds of infinite stam, literally 0 downsides and you can take another pill straight after. i personally take 1u synap/2u hyro pills so i can take them repeatedly and get infinite stam

> at worst you can have no stam in the middle of running around enemies.
...so the same as if you weren't taking synap????

> Medical Staff use it with Dylovene for counteracting Polyhexanide
99.9% sure dylo+tric does the job just fine, and if it doesn't, polyhex can be adjusted

> its key on zombie round
zombie round is an event mode + hyper works fine + admins can spawn in rye if needed

> but I don't see it use but by a hand full of people
You're blind. It is also still powerful as fuck

Weedmaster77:
> 99.9% sure dylo+tric does the job just fine, and if it doesn't, polyhex can be adjusted

The problem with Polyhexanide isn't its flat damage, its the after 50 ticks damage so if your doing surgery that's not just one limb or one presaged is where Hyronalin comes in. Changing  Polyhexanide is a good idea assuming it not going to make it harder to use and do more damage.

goose4429:
Based comment above, insta upvote
(Jokes aside, agree with 99% of the points)

pyronie:
@breadhunt im not saying they are perfectly balanced or anything, i was saying they shouldn't be nerfed into requiring lemoline in their recipe- making them far harder to obtain when they dont have game changing benefits(except for maybe synap, but on that case i just wanted to point out mole didnt mention the downsides). im not saying they shouldnt be changed but rather shouldn't need lemoline. the only one i think that should be touched minimally is hyro, because while you say the purge is minimal, its still something that it has over hyro and is more beneficial, along with the fact few people dose people with over 10u of dylo in combat.

breadhunt:
lemoline isn't that bad. MDs get a 30u bottle, researchers get a 10u bottle. for synap, I'd say that's enough - a 1u dosage is still 20 secs of infinite stam, meaning a doctor can produce about 2 pill bottles of synap still. 

hyro+rye are straight upgrades and _shouldn't be free_. you can stack dylo+hyro at 0 cost for double tox heal. ryetalyn is still a better hyper for neuropurge. this makes them completely unsuitable to be free; rewarding people for wasting their life in the chem room at the start of every op is bad

they ARE shit for lemoline stuff atm though so they'll have to be tweaked to fit it, for example making hyro the dermaline equivalent for tox, etc.

goose4429:
> Give medic vendors these meds for cheap.
> 
> Synaptizine injector with 3 uses should probably be 1 point, or you could make the injector have 10 uses and still cost 4 points (10u synaptizine, 20u hyronalin).
> 
> Hyronalin? Pill bottle of something like 5u pills, or just a syringe case. Maybe 8 points, not too good, not too bad. Very nice if someone gets 150 toxin damage. It does make toxin damage not very special, though all it did was make you take significantly longer to heal, because nothing healed toxin damage except dylovene.
> 
> Ryetalyn? Make it cost either as much as meraderm bottles, or something like 8 to 5 points. 3u Ryetalyn pill bottles for medics, or even just a single hypospray for 10 points would be a very nice option to have. Good for purging when someone gets 150u of neurotoxin, or you just need to get rid of their xeno chems while they have a ton of chems healing them already. 3u will be purged nearly immediately if they have premed, so it can't be abused in combat for very long unless they splurge points for it.

I am planning to do this in a week or so. Medic will have a separate section with specialized injectors, which will contain cool stuff. Including this.

SzkieletorNutric:
> lemoline isn't that bad. MDs get a 30u bottle, researchers get a 10u bottle.
>
No longer the case as RSRs gain as much Lemoline as MDs nowadays at 30u.

Issue #6864, opened 2021-05-15
Right click & Middle Mouse Click allows aimbot.


2whatever2:
## Testmerges:
#6748 #6804 #6848

## Reproduction:
--For Single Fire--
1. Hold right click on a person.
2. Drag it off.
3. Any further shots with single fire will always aim at that person, even if you're offscreen.

https://user-images.githubusercontent.com/24631139/118352439-02b2ed80-b594-11eb-8724-39490b5f464d.mp4

--For Auto Fire--
1. Have your target on the edge of the screen.
2. Hold right click on a person.
3. Without moving the cursor hold down left click.
4. Quickly move your mouse to the chatbox.
5. If done right it will automatically shoot towards the person you targeted. If done wrong you will shoot on their tile instead.

https://user-images.githubusercontent.com/24631139/118352676-393d3800-b595-11eb-80f1-644cd3c4fe6f.mov

SkeletalElite:
finally a use for right click

/s

This is a byond issue I believe it also happens with middle click even before right click was a thing

JackTheJackhammer:
Dear god

2whatever2:
> finally a use for right click
> 
> /s
> 
> This is a byond issue I believe it also happens with middle click even before right click was a thing

After some testing yes, mmb does the same thing too.

PR #10460, opened 2022-06-14
Minigun pre barrel spinning

SplinterGP:
## About The Pull Request
This code is pretty wack and it's not any final, i doubt this will be finished either, drafting so i can show code to maintainers and see if we can find solutions.

I think most of the code of course is organizable and doable, BUT there's an annoying bug which i believe is related to #6864
meaning it's a byond issue and sucks, once you start spinning up(right click) the mouse apparently "locks" into that place for byond, meaning that if you click the button to fire(M1)  apparently will click there instead of wherever your mouse is, mousedragging still sets target and all but when your mouse is still it tries to fire where you first aimed when spun your minigun :(
Only way i find to actually fix that is by making it a toggle, since it gives you slowdown having that woulnd't be a problem i guess but it wont face where youre spinning to or whatever, maybe that'd be fine for unique action.

https://user-images.githubusercontent.com/29824735/173480595-2eecd64b-5d54-4624-b0d9-75bc03bbb8b7.mp4

maybe we could talk to lummox about this ?

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -78,6 +78,7 @@
 #define GUN_AMMO_COUNT_BY_SHOTS_REMAINING (1<<12)
 #define GUN_NO_PITCH_SHIFT_NEAR_EMPTY (1<<13)
 #define GUN_SHOWS_AMMO_REMAINING (1<<14)
+#define GUN_HAS_ALT_FIRE (1<<15)
 
 //reciever_flags. Used to determin how the gun cycles, what kind of ammo it uses, etc.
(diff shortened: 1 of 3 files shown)


SplinterGP:
accidentally made it undreafted oops

Arctrooper8:
It's a cool idea but it might cause a lot of balance issues considering wind up is one of the major ways to control minigun power.

SplinterGP:
> It's a cool idea but it might cause a lot of balance issues considering wind up is one of the major ways to control minigun power.

yep thats why i came with idea to make it pretty slow moving speed when using it.

PR #10464, opened 2022-06-14
Increases Damage to Organs

goose4429:
## About The Pull Request

Recently effects of organ damage for heart, lungs were changed. Now instead of killing you over time, they give debuffs. This PR aims to restore long-term game balance.

When limb _is significantly damaged_, all brute attacks can cause organ damage with low chance.
Before, only strong attacks could do this. For example, in **light** armor, marine could get direct organ damage only if the attack had more than 33 raw damage. In **heavy+tyr**, 50 damage. Usual xeno attacks are 23 or 26, so not enough.

Overall, marines only got organ damage if they ran around with broken bones. Now beating them to a pulp actually does something.

## Why It's Good For The Game

Balance good. Before you could die from damage, now you don't.
We can appreciate the nerfs to organ damage if more organs are damaged.

_Note_: Yes, peri use was allowing you to ignore organ damage. But if I had a penny for every marine who used it, I wouldn't get enough money for a cup of coffee.

## Changelog
:cl:
balance: once limb can be broken, hitting them can damage organs.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_organs.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_organs.dm
@@ -45,7 +45,7 @@
 			if(E.is_broken() && E.internal_organs && prob(15))
 				var/datum/internal_organ/I = pick(E.internal_organs)
 				custom_pain("You feel broken bones moving in your [E.display_name]!", 1)
-				I.take_damage(rand(3,5))
+				I.take_damage(rand(4,6))
 
 			//Moving makes open wounds get infected much faster
 			if(!(E.limb_wound_status & LIMB_WOUND_DISINFECTED) && E.brute_dam >= 20)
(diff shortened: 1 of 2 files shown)


breadhunt:
> This PR aims to restore long-term game balance.
> chance.

yes let's remove one set of bullshit and replace it with another set of bullshit

pretty sure this was removed previously, for VERY good reason (RNG is shit for both sides lol)

> We can appreciate the nerfs to organ damage if more organs are damaged.

peri removal _immediately_ made that a buff, because now anyone competent can't just premed on peri 24/7 and ignore organ damage

goose4429:
>pretty sure this was removed previously
Removed what?

> peri removal immediately made that a buff, because now anyone competent can't just premed on peri 24/7 and ignore organ damage, but okay

I agree that removing peri was an impactful thing for handling organ damage. The thing is, you can still run around with the damage and ignore it. Without peri. Lung slowdown is equal to wearing another set of heavy+tyr armor (which is not game-changing). Heart damage is just... there. That's it.

Previously you were dying from organs. Now you get funny debuffs, which are pretty weak.
As for RNG - yeah, that's not the best way probably. But whole organ damage thing is currently based on chance already. Quite literally any source of organ damage is random.

goose4429:
Had a good discussion in #balance.
Key points:
-Peri wasn't used by noobs, noobs need not to suffer anyway.
-Main concern is lack of effects from organs, not noob abuse

Derrick2Electricboogaloo:
Just revert it at this point if you are going to open this PR

PR #10468, opened 2022-06-14
optimize /obj/item for initialization.

SplinterGP:
## About The Pull Request
this removes two var/list/whatever = list() from /obj/item, they start as null now(it was already made to be handled as null on actions code btw).

Why you may ask:
https://github.com/tgstation/tgstation/blob/master/.github/guides/STANDARDS.md
Under optimization.
tl;dr : starting a list on it's var definition causes byond to call a hidden init proc when the object is created.
http://www.byond.com/forum/?post=2086980&page=2#comment19776775

## Changelog
:cl:
code: removed unneeded list initialization on all items in order to improve game initialization
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -249,6 +249,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	for(var/datum/action/action_to_update AS in master_gun.actions)
 		if(action_to_update.target != src)
 			continue
+		LAZYREMOVE(master_gun.actions, action_to_update)
 		qdel(action_to_update)
 		break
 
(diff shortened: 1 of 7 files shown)


SplinterGP:
tried to profile this but i have other stuff open on my pc so initialization times vary alot + everytime i run independent of which branch it always initializes faster

code/modules/projectiles/gun_attachables.dm
```diff
@@ -249,6 +249,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	for(var/datum/action/action_to_update AS in master_gun.actions)
 		if(action_to_update.target != src)
 			continue
+		LAZYREMOVE(master_gun.actions, action_to_update)
 		qdel(action_to_update)
```
TiviPlus, line 253: why is this here of all places

SplinterGP: you mean the qdel ? No idea

TiviPlus: No why did you add the lazyremove

TiviPlus: Why would it not remove it? even if it doesnt why wouldnt it be handled in the Destroy()

code/modules/projectiles/gun_attachables.dm
```diff
@@ -249,6 +249,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	for(var/datum/action/action_to_update AS in master_gun.actions)
 		if(action_to_update.target != src)
 			continue
+		LAZYREMOVE(master_gun.actions, action_to_update)
```
SplinterGP, line 252:
```suggestion
```

PR #10470, opened 2022-06-15
Nerfs fire strength against resin walls

lKiseki:
## About The Pull Request
The strength of fire against resin walls specifically has been reduced from 1.5 to 1.25, as per Tivi's request.

## Why It's Good For The Game
Fire is too good at clearing walls. This should make it a tad bit more bearable.

## Changelog
:cl: Lewdcifer
balance: Fire strength against resin walls reduced from 1.5 to 1.25.
/:cl:

### DIFF
--- a/code/game/turfs/walls/resin.dm
+++ b/code/game/turfs/walls/resin.dm
@@ -22,7 +22,7 @@
 
 
 /turf/closed/wall/resin/flamer_fire_act(burnlevel)
-	take_damage(burnlevel * 1.5, BURN, "fire")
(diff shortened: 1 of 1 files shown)


Lumipharon:
does this actually reduce it enough to require an extra dose of fire?

lKiseki:
> does this actually reduce it enough to require an extra dose of fire?

As pointed out by @QualityVan in the Discord, no. It just makes it take a few more seconds, apparently.

BraveMole (approved):
Let's start with that, even if it's low impact, then nerf even more if needed

PR #10472, opened 2022-06-15
Necrotic limbs applies organ damage over time

QualityVan:
## About The Pull Request
I was completely convinced this code already existed, which's why I opened #10394 . It's even part of the why section. But, uh, it didn't actually exist. Adds it.

## Why It's Good For The Game
Turns out germs were in fact the only way necrosis hurt organs. Oops. Without this, necrosis on head/chest/gut doesn't really do anything.
Also cleans some carryover medscanner stuff from wound removal.

## Changelog
:cl:
add: Necrosis causes organ damage over time. Accidentally removed this when I killed organ germs.
/:cl:

### DIFF
--- a/code/game/objects/items/scanners.dm
+++ b/code/game/objects/items/scanners.dm
@@ -163,7 +163,7 @@ REAGENT SCANNER
 	if(ishuman(patient))
 		var/mob/living/carbon/human/human_patient = patient
 		var/infection_message
-		var/infected
+		var/infected = 0
 		var/internal_bleeding
 
 		var/unknown_implants = 0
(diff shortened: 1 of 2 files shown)


Losenis:
Wouldn't this lead to slight bouts of organ damage that happen if a limb gets necro'd for a bit in surgery instances? I'm not entirely sure if organ damage naturally heals if it's little enough or gradually increases in severity to know if this'll be an issue or not.

QualityVan:
> Wouldn't this lead to slight bouts of organ damage that happen if a limb gets necro'd for a bit in surgery instances? 

Yes but a: you need to leave necro for two minutes before the organ damage will do anything and b: if you're in surgery you can just organ damage anyway.

novaepee:
Septic shock incoming.

Weedmaster77:
@QualityVan How would this show up on hand scanners or would it at all? Would hope that there would be some way to know if its happening since some organs that are damaged cause toxin damage. Also to fix this would it just be a necro surgery when you have someone's bones bent open or some other method?

TiviPlus (approved):

PR #10473, opened 2022-06-15
Injector pouch buff

goose4429:
## About The Pull Request

Increases pouches for injectors so they are not as trash. Again.

Med pouches for pills hold 120+120+120+90+90=540u of reagents by default. Injector pouches hold at most 8x30=240u. 7 slots for pill bottles vs 6/8 slots for injectors is a joke. 

Buffed to 14 slots (2 injectors per 1 pill bottle slot). Injector pouch can hold 14x30=420u max, still not ideal but coping.

Figure 1: cpsman pouch
image

Figure 2: marine/SL pouch (they are same after last storage PR)
image

## Why It's Good For The Game

Consistency with med pouches, and less meta builds.
A lot of colorful stuff to let marines eye rest.

## Changelog
:cl:
balance: Injector pouches have 14 slots
/:cl:


Arctrooper8:
Couple of issues
14 seems like a bit much, the injector armor module only has ten. Pouches are generally around the same size as armor modules.

Injectors are supposed to be rapid application but smaller size. Perhaps we should look into reconsidering a pill bottle draw time for those without medical skill.

Also, if we're just giving marines more of the same injectors why not simply make the injectors bigger? It's easier to use that way and accomplishes the same goal.

Also squad marines aren't supposed to be getting pill bottles anymore, and there'll be more work on limiting them to pill packets in the future. Try comparing the injector pouch to the new first aid pouch with it's pill packets (120u bicard, 120u kelo, 120u trama, 90u tricord, and 90u dylo) 

If the issue of the injector pouch is that it cannot hold as many chems as the medkit pouch, perhaps we should look into increasing the storage of basic chem injectors.

goose4429:
Will rebalance injectors first, then return to this.

PR #10445, opened 2022-06-11
Fix flak jacket accepting modules

novaepee:
## About The Pull Request

Title. Sorry, synthetic players. Maybe we can make a dedicated synthetic armor that accept modules but give barely any armor. Food for thought.

## Why It's Good For The Game

Bug fix. One line change with null list.

The 6 months of bug abuse on flak jacket that synthetic players have used is no more. I recommend making a new armor for synthetic.

## Changelog
:cl:
fix: flak jacket accepting modules
/:cl:

PR #10479, opened 2022-06-17
adds modules back to the flak vest

Mechanelize:
## About The Pull Request

adds armor and storage modules back to the flak vest

## Why It's Good For The Game

fixes #10445
only modular vest that synths can use, inferior to normal jaeger in every other way

## Changelog
:cl:
balance: Readded armor and storage modules to the Flak Vest.
/:cl:


Mechanelize:
better solution incoming

PR #10480, opened 2022-06-17
Revert "Fix flak jacket accepting modules"

Mechanelize:
## About The Pull Request

Adds armor and storage modules back to the flak vest.

## Why It's Good For The Game

fixes #10445

The line added by 10445 is meant to be missing because as the flak vest is a variation of Xeno-M armor, it inherits the module list from its parent, Xeno-M, which allows for all modules. This is the same with both Xeno-H and Xeno-L, which also 'miss' that line of code because they inherit the accepted module list from Xeno-M, so don't need a list for themselves.

## Changelog
:cl:
fix: Readded armor and storage modules to the Flak Vest.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/modular.dm
+++ b/code/modules/clothing/modular_armor/modular.dm
@@ -262,6 +262,7 @@
 
 	allowed_uniform_type = /obj/item/clothing/under
 
+// Thank Jeff for providing sprites to color flak jacket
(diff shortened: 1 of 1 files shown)


BraveMole (changes requested):
So you are making a fix....

By reverting a fix that was asked for by a maintainer.... ok.

BraveMole:
And even readding the useless comment. And it is a duplicate

PR #10481, opened 2022-06-17
Reenables modules on flak vest

Losenis:
## About The Pull Request
Allows the flak vest to accept the modules Xenonauten armor has (IE, no jaeger amor pieces).

## Why It's Good For The Game
It had been disabled over a bugfix which had been mostly called a bug due to unintended behavior rather than breaking anything. During the time it had been enabled there wasn't any game breaking bugs and nor exploits of equal order since there's no reason to use flak over normal armors that can offer the same or more.

The bugfix however ended up as a crippling nerf to synth in terms of storage because they can't wear armor, therefore can't use said modules, other than what the flak vest used to be. This isn't about the armor values, as they can't even use the M3 harness that has literally no armor, although it can't accept modules either. The flak armor itself is a downgrade to other armors regardless, it's just that it's the only option it has.

Due to the unintended long testing of how it'd be without the "bugfix", and how little people seemed to complain about it and how vital it turned out to be for some loadouts, we could now call it an intentional change.

## Changelog
:cl:
balance: Flak vest can now accept modules. You still need a good reason to pick it over the other armors, though.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/modular.dm
+++ b/code/modules/clothing/modular_armor/modular.dm
@@ -272,8 +272,6 @@
 	soft_armor = list("melee" = 40, "bullet" = 50, "laser" = 50, "energy" = 25, "bomb" = 30, "bio" = 5, "rad" = 5, "fire" = 25, "acid" = 30)
 	slowdown = 0.25
 
-	attachments_allowed = list()
-
 	allowed = list(
 		/obj/item/weapon/gun,
 		/obj/item/tank/emergency_oxygen,


Arctrooper8:
The labcoats have the same storage as a medical jaeger armor module. Would it not be best to maybe give something such as hazard vests or some other engineering clothing the same storage as a Jaeger engineering pouch instead?
This would leave Synthetics with the decision of either storage or armor instead of having both at the same time.

FosterGrime-Heymynameslouie:
its not a bug, its a feature! 

based pr spede merhg

BraveMole (changes requested):
I do not see any justification for that storage creep.

This bug was here for a while is not an accepted justification on this codebase.

TiviPlus:
I said no that means no

Losenis:
>I do not see any justification for that storage creep.

Bravemole, synth has the absolute lowest storage quantity of anyone on the marine side because it can't use storage module, despite being expected to pack for both engineer, corpsman, and surgery given its skillset. How is that powercreep?

PR #10487, opened 2022-06-17
Delete healing portion of Doctor's Delight

novaepee:
## About The Pull Request

You cannot heal from Doctor's Delight anymore.

## Why It's Good For The Game

The time of anti-powergamer PRs is now.

Doctor's delight is a superior tricord. You can't OD from doctor's delight, so your bloodstream can be filled with doctor's delight. I'm not even sure if this med counts against a marine if they get defiler purge.

It's true that doctor's delight makes you hungry, but you can just counter this by producing ironsugar or nutrition pills (yes, nutrition pills exist) and consume both doctor's delight and either the two pills mentioned above to not be hungry.

Lolipop is a good premed tool, but doctor's delight is superior.

This ends now.

## Changelog
:cl:
balance: doctor's delight doesn't heal anymore
del: doctor's delight doesn't heal anymore
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/drink.dm
+++ b/code/modules/reagents/reagents/drink.dm
@@ -473,10 +473,6 @@
 	adj_dizzy = - 10
 
 /datum/reagent/consumable/drink/doctor_delight/on_mob_life(mob/living/L, metabolism)
-	L.adjustBruteLoss(-0.5, 0)
-	L.adjustFireLoss(-0.5, 0)
-	L.adjustToxLoss(-0.5, 0)
-	L.adjustOxyLoss(-0.5, 0)
 	L.AdjustConfused(-10 SECONDS)
 	return ..()
 


goose4429:
11/10 speed meregr

Dolth:
Why not tweaking it instead of entirely removing it? Giving it an OD threshold and a drawback more severe than just food or just weakening it to 0.25 or something else?

Like, I am not even gonna say the chem is strong but why just removing things?

KrabSpider:
Just give it booze power motherfucker. Its a cocktail, you can literally balance it around that.

InterroLouis:
> Just give it booze power motherfucker. Its a cocktail, you should balancing it around **that** instead of as a medicine. Marines can't constantly chug alcohol and fuck, if they want to try then let them kill their livers and die of liver damage

So...change the recipe too? There's no alcohol in the current Doctor's Delight recipe at all.

KrabSpider:
> > Just give it booze power motherfucker. Its a cocktail, you should balancing it around **that** instead of as a medicine. Marines can't constantly chug alcohol and fuck, if they want to try then let them kill their livers and die of liver damage
> 
> So...change the recipe too? There's no alcohol in the current Doctor's Delight recipe at all.

Yes

QualityVan:
I agree with the intent but I'm not wild about the method. Maybe make it so instead of draining nutrition it adds it so you get fat if you use it too much? Or add an alcohol level to it, yeah.

FosterGrime-Heymynameslouie:
doctors delight making you infinitely fat instead of infinitely hungry is infinitely funnier and is actually an infinitely better deterrent than just straight up removing the healing "Doctor"  part of the drug.

even though I never used it myself, lmao.

Derrick2Electricboogaloo:
Make it drain more nutrition, since that slows marines down. Also. If pops are better, then why remove it seems in-consequential? Not everyone uses to gamer increase their HP gain at crazy rates. And I hardly see it mattering in an OP. I know I for one use it to supplement healing with other better chems so marines recover quicker after being in crit/dead.

breadhunt:
should probably just add some sort of actual drawbacks, like confusion/stagger

Dolth:
Careful frens there is lipozine to remove fattness, it's super effective.

Jordanstolethepreciousslam:
A  superior trico that have less healing and requires user to actually eat? Sugar/nutri pills do work but if you constantly eat them you can aswell just eat regular trico pills. 
DD is not a better trico, it's just another trico for those who use builds incompatible with regular trico or just think that one trico is not enough.
If there is a problem with that, i suggest to just make DD purge regular trico so you couldn't use basically the same drug in two forms simultaneously

novaepee:
Balance is not my forte.

PR #10498, opened 2022-06-20
Remove action button in FOB drone to produce turret voucher since such voucher is dead

novaepee:
## About The Pull Request

Title. Git will remember.

## Why It's Good For The Game

Remove old code that no one can possibly use. If ya want it back, git will remember and make a better design.

## Changelog
:cl:
del: Remove action button in FOB drone to produce turret voucher since such voucher is dead
/:cl:


Arctrooper8:
I've got a PR placing sentries in the FOB drone round start and I'm also trying to figure out how to get sentries to eject.

PR #10516, opened 2022-06-24
give AR-11 PB

novaepee:
## About The Pull Request

Title

## Why It's Good For The Game

This rifle doesn't have PB because MJP forgot. I always thought this was a feature that MJP intended, but it's not the case.

## Changelog
:cl:
add: give AR-11 PB
balance: marines have more reasons to use it since it's a pretty strong weapon
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1201,7 +1201,7 @@
 		/obj/item/attachable/scope/marine,
 	)
 
-	flags_gun_features = GUN_AMMO_COUNTER
+	flags_gun_features = GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER
 	gun_firemode_list = list(GUN_FIREMODE_SEMIAUTO, GUN_FIREMODE_BURSTFIRE, GUN_FIREMODE_AUTOMATIC, GUN_FIREMODE_AUTOBURST)
 	starting_attachment_types = list(/obj/item/attachable/stock/irremoveable/tx11, /obj/item/attachable/scope/mini/tx11)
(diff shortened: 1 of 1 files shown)


Al-1ce:
> It's already 2 stronk, would be insane with 40 melee damage + a burst as PB imo.

I'm not sure if you know this, but point-blanks fire a single shot in addition to the melee damage of the gun. You cannot burst with a PB. It fires a single bullet.

Dolth:
> > It's already 2 stronk, would be insane with 40 melee damage + a burst as PB imo.
> 
> I'm not sure if you know this, but point-blanks fire a single shot in addition to the melee damage of the gun. You cannot burst with a PB. It fires a single bullet.

No? When you PB with a T18 it shoots 4 times + melee. Unless you meant something else?

Lumipharon:
Yeah, pb burst definitely fires the whole burst. Honestly I think AR11 is fine without PB, it already has silly dps.

Naaanii:
Would be fine to give it PB for consistency and Nerf the actual rifle itself which seems to be the real issue.

Al-1ce:
> No? When you PB with a T18 it shoots 4 times + melee. Unless you meant something else?

I just learned that PBing with autoburst shoots only one time! I have no idea why this happens. For the three years I've been on this server, I've always used autoburst, so I didn't even know burst-pbing was possible. 

https://youtu.be/wCTf0mriBmg

Lumipharon:
autoburst uses autofire code, is probably why.

TiviPlus:
Going to go with the safe side and agree with dolth here + you even mention its already strong in the CL
Alternatively if you insist and others agree you could explore nanis suggestion

PR #10530, opened 2022-06-25
Adds a pinpointer to op vendor

Losenis:
## About The Pull Request
Operational supplies vendor in req now has a free pinpointer.

## Why It's Good For The Game
Lack of people playing command roles for whatever reason can have a tremendous impact. This should alleviate the hit and encourage people to actively ask for aSL and lead other players should they find out there's no SLs, as well as not punish req by forcing a 20 point purchase for 1 single pinpointer over people not picking a role.

~~Should be good for two squads, but not for 3 or 4.~~ After some more thinking, I realized that if CIC is staffed to make someone aSL, it's also staffed to merge all squads into 1 if there's no SLs at all. That nobody actually brought this up on discord and had to figure it out on my own a day later only tells me nobody even plays command that often lmao.

TL;DR encouraging teamwork is good.

## Changelog
:cl:
balance: Operational supplies vendor now has a free pinpointer.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -570,6 +570,7 @@
 	req_one_access = list(ACCESS_MARINE_CARGO, ACCESS_MARINE_LOGISTICS)
 	products = list(
 		"Surplus Special Equipment" = list(
+			/obj/item/pinpointer = 2
 			/obj/item/beacon/supply_beacon = 1,
 			/obj/item/ammo_magazine/rifle/autosniper = 3,
 			/obj/item/ammo_magazine/rifle/tx8 = 3,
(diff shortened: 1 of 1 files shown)


Naaanii:
1 seems more than enough if this concerns low pop.

Losenis:
You'd be surprised at how few people play command even in 20+ marine rounds. I went with 2 because I'm imagining the scenario where two of the same squad get it and it doesn't really do much more than 1 because they're gonna take the same path to the silo anyway, but if it's 2 squads that aren't actively merged that's like half the marines losing compass, much like how it'd be if only one single SL joined and would need the other squad to have a aSL.

Araulius:
maybe if you want access to the tool, you should pick the role that is clearly open... consequences.

Naaanii:
Marines should be punished in a way if the push dies. Either way, a new SL could spawn or another could be purchased for 20 points if it's really needed.

Most low pop maps are kind of small enough for marines to eventually find silos and the respawn rates aren't too fast for Xenos to recuperate and like Araulius mentioned, people should go command and this just gives even less incentive to go command.

Losenis:
If the push dies the marines are kind of already punished by losing the round if xenos don't suicide in a fobsiege. You can't really push after a cocooned wipe if you couldn't push with marines all starting relatively healthy.

Pinpointer count still reduced to 1, reason in description.

PR #10535, opened 2022-06-26
Moves Arrowhead to Req

Pariah919:
## About The Pull Request
Moves arrowhead over to req from normal play.

## Why It's Good For The Game
Arrowhead is pretty inanne for a normal kit module from my time playing it, personally. It fits better in req considering it lets you shrug off minor attacks like its nothing, considering marines mainly die from chip damage this is a pretty big thing.

## Changelog
:cl:
balance: Arrowhead is now a req only module.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -1052,7 +1052,6 @@
 			/obj/item/armor_module/module/ballistic_armor = -1,
 			/obj/item/armor_module/module/better_shoulder_lamp = -1,
 			/obj/item/armor_module/module/chemsystem = -1,
-			/obj/item/armor_module/module/eshield = -1,
 		),
 	)
 
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -744,6 +744,13 @@ ARMOR
 	contains = list(/obj/item/armor_module/module/hlin_explosive_armor)
 	cost = 12
 
+/datum/supply_packs/armor/modular/attachments/eshield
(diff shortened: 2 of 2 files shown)


XxBarotraumaBugReporterxX:
I'd be for buffing it if it was to be moved to req, though

vilereaver41:
> I'd be for buffing it if it was to be moved to req, though

no

FosterGrime-Heymynameslouie:
is good if massively buffed or has low point count, other wise it becomes something like the explosion module, banished in req hell for being expensive and has doo doo practicality for its price (I forgot its name) 

maybe has a price range near the optical scanner?

Dolth:
Arrowhead is already MASSIVE. You people tend to forget xeno fights with ATTRITION which is heavily negated by arrowheads.
And even though 12 points for something that protects you from 2 to sometimes 3 slashes with such a low cooldown? Nah this nerf is on point.

Weedmaster77:
Unless it made relatively cheap in REQ I agree that is will probably not be bought much. 

In my opinon it should be taking from the armor vendor and left in the Medic, Engineer, Smartgunner and Leader Prep vendor so its not something that every marine grunt can use only marines that have a more important role. 
Therefore making it less common on regular marines the ATTRITION affect would be more present but give roles that stand still preforming there role offend and get targeted the usefulness of the mod while not just dooming it to REQ.

Lumipharon:
You could make it only fit on light armour, that way it leans into the skirmisher niche, but ensures you'll be taking solid damage when down.

BraveMole:
Arrow head doesnt deserve to be sent to req (aka never used). 

If you think it is too strong, you can eventually increase slowdown ?

PR #10536, opened 2022-06-26
Removes reload time on the ALF machinecarbine

Losenis:
## About The Pull Request
ALF-51B Kauser machinecarbine no longer has reload time

## Why It's Good For The Game
Channeled reload like T60 on a cqc gun that is only useful at close range is wack

## Changelog
:cl:
balance: ALF-51B Kauser machinecarbine no longer has reload time
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/magazines/rifles.dm
@@ -319,7 +319,6 @@
 	default_ammo = /datum/ammo/bullet/rifle/heavy
 	w_class = WEIGHT_CLASS_NORMAL
 	max_rounds = 80
-	reload_delay = 1 SECONDS
 	icon_state_mini = "mag_t42"
 
 //-------------------------------------------------------


Lumipharon:
The gun is exceptionally powerful, with almost no slowdown, very high damage and capacity. The reload time is the only thing that makes it balanced.

Losenis:
Lad it has a 3 damage falloff per tile, it's already worse than a t12 in damage past 2 tiles on top of having no aim mode. It already has its pros and cons regardless of the reload which conceptually is ass if you're gonna make a cqc weapon not be able to be reloaded while moving and take time. Not even the T42 has that.

Lumipharon:
> Lad it has a 3 damage falloff per tile, it's already worse than a t12 in damage past 2 tiles on top of having no aim mode. It already has its pros and cons regardless of the reload which conceptually is ass if you're gonna make a cqc weapon not be able to be reloaded while moving and take time. Not even the T42 has that.

You are completely misunderstanding the weapon if you're comparing it to a t12 or t42.

It is a chasing weapon with very low slowdown, it is comparable to the t18 or even p90. T12 and above are too slow to be used to chase, and fundamentally are completely different in purpose.

Against a t18, it has the same burst delay but much longer extra delay, in exchange for 50% more damage, 100% more AP, over 100% more sunder, and over 100% more capacity. Damage fall off does not matter to a gun designed specifically to chase people and shoot in the pointblank in the face. 

The reload stops it from just being impossible to melee fight an ALF user, and the fall off (plus relatively high scatter) stops it from just being powerful at all ranges. Why would you even bring up aim mode?

Losenis:
Sounds like a best case scenario of fighting xenos that stand still in front of marines and slash without using any abilties and really like going solo around open areas. The sunder and pen don't seem to be an issue for the tx11 that doesn't have the damage falloff either on top of having the tacreload and only 10 rounds less. And if you want to chase you can always just 1hand a t90 or whatever, they're always free so you don't have to juggle ammo boxes to refill mags for seasonals out of season.

Lumipharon:
> The sunder and pen don't seem to be an issue for the tx11 that doesn't have the damage falloff either on top of having the tacreload and only 10 rounds less. And if you want to chase you can always just 1hand a t90 or whatever, they're always free so you don't have to juggle ammo boxes to refill mags for seasonals out of season.

This may be a surprise, but movement speed is quite important in this game. So having a low slowdown is a huge advantage over other guns that needs to be balanced appropriately, because as you put it, the xeno does not in fact need to stand still in front of a marine when the marine can run as fast or faster than them. Damage application is what's important, not just actual DPS.

This is why guns that are very strong, do not have good movement speed, or like the ALF, have significant drawbacks. And guns that are very mobile, like the P90 have much lower stats.

Losenis:
If you're not running heavy armor, you're gonna get slowed down hard by damage on the first or second slash. That's assuming you're not getting spit on by a ranged caste the machinecarbine will be garbage against anyway. And if you're running heavy armor you're not chasing.

Again, best case scenario of some dude actually having the speed to chase a lone xeno that needs to flee through chaseable terrain. Usually either the marine gets too damaged to chase or the xeno got skill issued hard enough in the fight that he couldn't land a slash on the marine. And no I'm not counting a ALF marine running in after the xeno fought another because that'd include any other weapon like a scoped RR.

Pariah919 (commented):
Disagree, thought about this myself, Lumi convinced me its a bad idea in general considernig how the gun is.

Naaanii:
Remove reload time and Nerf damage also works if straight buffs are bad.

Autisem:
This does sound like it would be very unfun for most xeno casts to fight. I know its already extremely frustrating to fight a marine thats dog fighting you because you dont have a spit(or they simply dont care) dont think they need to also be able to reload immidiatly, offers SOME change at escape

Losenis:
Fair enough, though from my experience with the gun the only thing it does is just make it an annoyance to reload outside of a fight by forcing you to stay still, since that 80 round means the fight would already be over by the time you need to reload 9 out of 10 times.

PR #10537, opened 2022-06-26
Adds the machinecarbine and its ammo to req

Losenis:
## About The Pull Request
Adds ALF-51B machinecarbine and its ammo to req for the same price as most other seasonals (5 for gun, 1 for ammo).

## Why It's Good For The Game
It's a fun gun with pros and cons that should be free to order from req if it's not in seasonal, same as other seasonals.

## Changelog
:cl:
add: ALF-51B machinecarbine and its ammo is now orderable from req
/:cl:


TiviPlus:
Brave and I are against adding seasonals to req

PR #10540, opened 2022-06-27
Eye damage reduces accuracy, imi suppresses

QualityVan:
## About The Pull Request
Bruised eyes cause an accuracy reduction. I got no idea how much is reasonable, so the amount is a little random. You'll still get a bit of blur occasionally, but it's just there to remind you you got injured eyes. Broken eyes still blind you.
Imidazoline changed to suppress eye damage, similar to how QC works, as long as it's in the mob. Applies to both the accuracy penalty and the blur/blind. Causes 1 kidney damage per pill, each lasting just over 3 minutes.

## Why It's Good For The Game
Blur sucks and is annoying.
Making imi a suppressant keeps it useful as a temp fix rather than just being a situational peri+ without the cloneloss. Kidney damage is small, but means you can't just stay on imi all round. Bruise around half an hour, broken around triple that assuming no other damage. Metabolism is reduced from default but not tremendously, so you'll still chug through pills at a decent clip.

## Changelog
:cl:
balance: Eye damage reduces your accuracy with guns
balance: Imidazoline suppresses eye damage instead of curing it but metabolises slower.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -160,7 +160,7 @@
 
 /obj/item/reagent_containers/hypospray/autoinjector/imidazoline
 	name = "imidazoline autoinjector"
-	desc = "An auto-injector loaded with 3 doses of imidazoline, medicine for fixing eyesight."
+	desc = "An auto-injector loaded with 3 doses of imidazoline, medicine for temporarily ignoring eye damage. Excessive use advised against."
 	icon_state = "autoinjector-5"
 	list_reagents = list(/datum/reagent/medicine/imidazoline = 30)
 
--- a/code/game/objects/items/reagent_containers/pill.dm
+++ b/code/game/objects/items/reagent_containers/pill.dm
@@ -193,7 +193,7 @@
 	pill_id = 14
 
 /obj/item/reagent_containers/pill/imidazoline
-	pill_desc = "An imidazoline pill. Heals eye damage."
+	pill_desc = "An imidazoline pill. Suppresses eye damage as long as it's in the patient, but slowly harms kidneys."
 	list_reagents = list(/datum/reagent/medicine/imidazoline = 10)
 	pill_id = 17
 
--- a/code/game/objects/items/storage/firstaid.dm
+++ b/code/game/objects/items/storage/firstaid.dm
@@ -477,7 +477,7 @@
 
 /obj/item/storage/pill_bottle/imidazoline
 	name = "imidazoline pill bottle"
-	desc = "Contains pills that heal eye damage."
+	desc = "Contains pills that suppress eye damage. Long term use advised against."
 	greyscale_config = /datum/greyscale_config/pillbottleround
 	pill_type_to_fill = /obj/item/reagent_containers/pill/imidazoline
 	greyscale_colors = "#F7A151#ffffff" //orange like carrots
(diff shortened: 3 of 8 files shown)


Derrick2Electricboogaloo:
So how do you fix eye damage then? If imi no longer fixes it and just suppresses it?

Losenis:
Please don't. Making it temporary is bad enough, adding another organ damage on top is just CBT tgchem shenanigans, but the worst part is that eye healing surgery is a complete separate surgery to brain healing despite both being in the head and usually both happening on the same bone frac. The other one is lung and heart that are both fixed in the same surgery step. It's such a pain in the ass that even when I run surgery I'd rather spend the already field-limited imialky mix rather than do those surgeries. And imialky mixes are a luxury, you usually need 2 slots just for imi and for alky in separate if you don't want to go to chem, which was already separate from peridaxon back in the day.

Not to mention old peridaxon before it got canned didn't even give organ damage despite being effectively the same. This'd be a massive nerf to imi as well because it'd be not only less effective in the imialky fix, it'd also explode in demand by being just temporary rather than a full fix to something that either was a freak incident (welding without goggles/module) or something that will likely keep on happening (Headfrac, splint torn off on damage).

Mrrpip:
...So if one of the best gun types is the shotgun category. And eye damage reduces accuracy, why won't you use the CQC guns?

I mean, I know shotgun meta and all that is boring, but well....

novaepee:
This will make targeting eyes more viable than aiming for limbs.

QualityVan:
> So how do you fix eye damage then? If imi no longer fixes it and just suppresses it?

Surgery, autodoc, or peri+, same as every other organ.

> This will make targeting eyes more viable than aiming for limbs and even brain.

That's not something that exists.

Pariah919:
blur makes my irl head hurt so if you're going to make eye damage in any way good I'd say that's an autoclose

breadhunt:
First off, I'm pretty sure that A LOT of things change bluriness (blood, etc.) and eye bluriness doesn't really.. heal quickly at all. You mentioned that the main objective here was to remove bluriness; but this just seems to make the issue worse, especially considering imidozaline as-is can prevent bluriness (a good thing).
Since bluriness wears off pretty slowly, you might also want to adjust the probability for it to get worse

Main balance concerns.

What is the problem with imi as-is? It is effective at healing eye damage but considering how - gonna be honest here, ass - eye damage is, that's fine. With accuracy debuffs, making imi useless is a very large nerf to anything ranged and means bluriness cannot be healed quickly (as I said earlier) which makes blur even more ass than it already is.

Imidozaline also contributes to defile chems, liver damage chems.. etc. If the issue is that imadozaline is too strong, you could alternatively nerf the rate of healing or increase the metab rate/lower OD threshold so it can't be premedded with as effectively (because premed = eye immunity)

Eye damage for liver damage is also a bad idea. It'd be a bad idea even if it healed eye damage. 
Exchanging damage types for other damage types (on common meds) can make the med system overly confusing; cobbychem had this issue, I'm pretty sure.
Pretty sure carrot juice can also be used as a better version of imi if this PR goes through

As Mrrpip said, this makes eye damage ass for most weapons, and heaven for PB guns. Meaning there's no point to healing it or caring about it if you've got a shotgun. Sounds worse for balance in general, especially if it's as hard to heal as other organs.


Removing blur is good, but different damage effects are needed (accuracy isn't used by all weapon types), surgery needs streamlining because eye surgery sucks and changing imi just sounds like a bad idea in general.

QualityVan:
Splitting this into two prs, have a better route for the accuracy mod.

PR #10538, opened 2022-06-27
Dylovene reduces stamina regen with any volume

QualityVan:
## About The Pull Request
Right now dylovene has the drawback of dealing stamina damage (and thus stopping stam regen entirely) above 10u. This changes that to a 50% regen reduction that activates at any volume.
Changes stamina regen multiplier to be dynamically calculated from a modifier list, similar to movespeed. Can't go below zero for now.

## Why It's Good For The Game
Removes the necessity to use a split dose for a basic healing chem in order to avoid drawbacks. You'll still want to mix it with other tox chems if you want a higher heal rate.

## Changelog
:cl:
balance: Dylovene reduces stamina regen by half instead of preventing it entirely, but does so at any volume.
/:cl:

PR #10542, opened 2022-06-27
Dylovene doesn't affect stamina

QualityVan:
## About The Pull Request
Alternative to #10538 
I made that one because I was annoyed at using supplies as provided being pretty strictly the wrong choice, but on rumination dylo's a core chem and probably doesn't strictly need a drawback. So, this. Keeps the stamina regen modifier stuff codeside because I'd like to use it in the future.

## Why It's Good For The Game
Basic supplies should be usable as is, and dylo might not need this drawback.

## Changelog
:cl:
balance: Dylovene doesn't impact stamina at all.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -78,6 +78,7 @@
 
 	set_armor_datum()
 	AddElement(/datum/element/gesture)
+	stamina_regen_modifiers = list()
 
 /mob/living/Destroy()
 	for(var/i in embedded_objects)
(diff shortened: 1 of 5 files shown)


goose4429:
Friendly reminder stam drain is there to keep premeds off, because having dylo in the blood already means neuro just doesn't affect you. Its hard for xeno to put reagents inside you faster than 1.4 per sec.

Lumipharon:
What Goose said. Premed dylo is really shit for gameplay, I don't think it's a good idea to remove it.

BraveMole:
Prefered the alternative

PR #10551, opened 2022-06-29
Adds code to enable preventing reagent container refills

QualityVan:
## About The Pull Request
The frame from #7814 , but purely as an option. No gameplay changes, the var enabling it isn't active on anything.

## Why It's Good For The Game
Lets rarer chems be available in vendors in a limited amount, rather than forcing them to either be absent or infinite.

## Changelog
:cl:
code: Reagent containers can be set to not allow restocking after use.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/reagent_container.dm
+++ b/code/game/objects/items/reagent_containers/reagent_container.dm
@@ -13,6 +13,8 @@
 	var/volume = 30
 	var/liquifier = FALSE //Can liquify/grind pills without needing fluid to dissolve.
 	var/list/list_reagents
+	///Whether we can restock this in a vendor without it having its starting reagents
+	var/free_refills = TRUE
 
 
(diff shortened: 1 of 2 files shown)


Weedmaster77:
like what chems?

code/game/objects/items/reagent_containers/reagent_container.dm
```diff
@@ -68,3 +70,9 @@
 			. += "; [R.name]([R.volume]u)"
 	else
 		. = "No reagents"
+
+/obj/item/reagent_containers/proc/has_initial_reagents()
```
TiviPlus, line 74: autodoc

TiviPlus (changes requested):

Autisem:
This should prop be done per reagent bottle from the vendor and not on vendor level, does this mean you cannot refil bica bottles anymore?

PR #10556, opened 2022-07-01
DMI editor actually works now + remove outdated extension

SplinterGP:
## About The Pull Request
the removed extension ins't needed anymore because spaceman's extension already does that better https://github.com/tgstation/tgstation/pull/66862


SplinterGP:
btw the dmi editor extension wanst working because of the .dmi being associated to imagepreview, now with this PR it should work properly

PR #10489, opened 2022-06-17
Stun tweaks

Losenis:
## About The Pull Request
Disclaimer: Values up to change as testing is done so that equivalent buffs are applied where stun removal acts as a harsh nerf. This is meant to distance gameplay away from chaining stuns, not net nerf or buff to xeno compositions.

Rebalances some xenos to be competitive while toning down their stuns in exchange for buffs in other areas, or other xeno aspects as a whole. Draft while working it out.

## Why It's Good For The Game
The game oughta be fun for both sides, to play as and to play against. Stuns lead to a bit of fun on the one using them, while the ones on the receiving end end up in stunlock city. However, due to stuns playing a large role in the power of a caste, they end up weak in other areas that turn into a mess when they're no longer in a situation they can use their stuns, extreme example being ungaballs of 20+ marines if they're not rushing the 1x1 chokes so that they're fighting only a few at a time.

A smaller scale PR had been TM'd for the shrike and after talking with some people, it seems that even though initially it was tedious and difficult to get used to the new gameplay, they found it not only fun to play the new shrike, but also fun to play against as well. This PR would be the next step in toning down stuns for xenos as a whole but also giving them something in return while thinking of them as a team rather than castes in a vacuum.

This does NOT aim to remove ALL stuns from xenos.

Changes:

Shrike:
HP increased ~~by 25%~~ to 350/390/420/450 Young/Mature/Elder/Ancient (Health was initially put there before speed was considered, both together ended up a bit overboard), speed equalized to ravager. Stun effectively removes from fling and unrelenting force, knockdown remains. Cooldown and cost of abilities reduced by half.

Boiler, Spitter, Praetorian:
Acid spray turf duration increased by 2x. Buffs them as an area denial tool since inevitably fewer stuns means xenos won't be able to keep marines from just bumrushing forward even harder.

Defender:
Forward charge stun equalized to tail sweep. Gives around 5s of stagger (for marines, it means around a 50% damage debuff), increased the bonus damage from 0.5x to 1x, in practice meaning if you hit a marine you're effectively landing 2 slashes. Range increased to 6. Plasma cost reduced to 60 from 80, even on ancient a defender would've only been able to use it twice; it's considerably weaker, so the old high cost is no longer justified. Purpose is making it less of a midcombat stun and more of a mobility ability that'll still tank your plasma, be it as initiation or retreat tool, while still retaining a benefit for using it against a target.

[...]

PR #10563, opened 2022-07-02
Removes stagger from most guns that have it

Losenis:
## About The Pull Request
Originally part of #10489, split in order to keep the former as xeno-only changes.

## Why It's Good For The Game
Copypasted from the other PR:

[Guns:
Removed stagger from most of them. Even though staggered marines had a 50% damage debuff, staggered xenos were locked out of their abilities altogether, a good few of them needing said abilities for their own survival rather than horizontaling spessmen faster. Same way marines shold have fun playing vs xenos, xenos should also have fun playing vs marines, and getting locked out of your abilities by marines is just as unfun as getting stunlocked by xenos. Note, this only affects stagger; that is to say, slugs and buckshot still knockdown and punch xenos a few tiles, so does martini, and so on. Only the "xeno can't use any abilities" part is changed.]

There's also very little feedback on your abilities being locked from usage from weapons you might not notice some marines having since you're still WASDing around to care about the second marine shooting at you. For people that haven't memorized all the guns that cause stagger, it's difficult to figure out what's happening until you're told. This is also an issue when new guns are introduced for people that don't read github. Let's be honest, we all had to learn the hard way. Ideally other means are used that give way more feedback, like the knockdown that slugs have already that already visually show you're incapacitated in some manner.

## Changelog
:cl:
balance: Stagger removed from most guns.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -357,7 +357,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	sundering = 2
 
 /datum/ammo/bullet/pistol/hollow/on_hit_mob(mob/M,obj/projectile/P)
-	staggerstun(M, P, stagger = 1, slowdown = 0.5, knockback = 1)
+	staggerstun(M, P, stagger = 0, slowdown = 0.5, knockback = 1)
 
 /datum/ammo/bullet/pistol/ap
 	name = "armor-piercing pistol bullet"
@@ -383,7 +383,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	sundering = 3.5
 
 /datum/ammo/bullet/pistol/superheavy/on_hit_mob(mob/M,obj/projectile/P)
-	staggerstun(M, P, stagger = 1, slowdown = 1, shake = 0)
+	staggerstun(M, P, stagger = 0, slowdown = 1, shake = 0)
 
 /datum/ammo/bullet/pistol/superheavy/derringer
 	handful_amount = 2
@@ -440,7 +440,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	sundering = 3
 
 /datum/ammo/bullet/revolver/on_hit_mob(mob/M,obj/projectile/P)
-	staggerstun(M, P, stagger = 1, slowdown = 0.5, knockback = 1)
+	staggerstun(M, P, stagger = 0, slowdown = 0.5, knockback = 1)
 
 datum/ammo/bullet/revolver/tp44
 	name = "standard revolver bullet"
@@ -485,7 +485,7 @@ datum/ammo/bullet/revolver/tp44
 	sundering = 3
 
 /datum/ammo/bullet/revolver/highimpact/on_hit_mob(mob/M,obj/projectile/P)
-	staggerstun(M, P, weaken = 1, stagger = 1, slowdown = 1, knockback = 1, shake = 0.5)
+	staggerstun(M, P, weaken = 1, stagger = 0, slowdown = 1, knockback = 1, shake = 0.5)
 
 /datum/ammo/bullet/revolver/ricochet
 	bonus_projectiles_type = /datum/ammo/bullet/revolver/small
@@ -603,7 +603,7 @@ datum/ammo/bullet/revolver/tp44
 	sundering = 1.25
 
 /datum/ammo/bullet/rifle/repeater/on_hit_mob(mob/M, obj/projectile/P)
-	staggerstun(M, P, max_range = 3, slowdown = 2, stagger = 1, shake = 0.5)
+	staggerstun(M, P, max_range = 3, slowdown = 2, stagger = 0, shake = 0.5)
 
 /datum/ammo/bullet/rifle/incendiary
 	name = "incendiary rifle bullet"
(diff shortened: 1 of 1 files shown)


bluezorua:
I don't really care about the other weapons but Martini and shotguns should be unchanged as it fits in with their slow fire rate but high damage niche

GrayRachnid:
The only gun that should have stagger left is martini IMO. Shotguns aren't that slow of a firerate and they already have knockbacks+stuns.

Losenis:
Martini's stun is the same duration as the stagger. In practice it won't be noticeable save for freak incidents where you hit a xeno that is weaken-immune but not stagger-immune or some bug happens.

GrayRachnid:
> Martini's stun is the same duration as the stagger. In practice it won't be noticeable save for freak incidents where you hit a xeno that is weaken-immune but not stagger-immune or some bug happens.

Most T3s are stun immune

FosterGrime-Heymynameslouie:
one small step for ability based gameplay, one giant leap for completely obliterating stuns.

is a cool pr but what exactly are the guns that have been changed? "Stagger removed from most guns." is probably not very descriptive.

Losenis:
Basically all of them save for the weird ones that have like 0.1, which I can only imagine are there to not fuck up something else about them (I couldn't even remove the line without it fucking up with a permastun, so it had to be left at 0). You can check in Files Changed tab, it highlights the changes.

Araulius:
12 gauge slugs should trade their stagger for level 3 slowdown. currently they have level 2 slowdown.

Al-1ce:
I may be the odd one out here (as a xeno main) but this has never bothered me. If anything you're making the game considerably easier. These are things which, in my opinion, xenos should prepare for and be adapted to. Also, it makes a couple of large explosives more or less only inflict slowdown alongside the damage.

Lumipharon:
This is just removing difficulty from xenos because the opposing team does something I don't like to me. 
Repeater is pretty shitty though.

EvilMelonTheSecond:
You should probably add slowdown or something in exchange for stagger especially for shotguns since they rely on reducing xeno damage potential in exchange for range. Long range guns will probably be fine without anything though.

XxBarotraumaBugReporterxX:
> This is just removing difficulty from xenos because the opposing team does something I don't like to me.
Of course xenos don't like something such cancerous as stagger.
> Repeater is pretty shitty though.
On it's own, it sucks ass, just like slugs, but it allows via stagger to have your friends gun down and chase the staggered xeno

Naaanii:
Stagger seems like a mechanic that doesn't need too much tweaking as whenever it was an issue, it's already been nerfed and death with (eventually).

Reducing stuns and slowdown seems okay to actually favour slowdown as the more 'unfun' part of playing Xeno is hard or soft CC that can be slammed with little or no opportunity to escape or off screens. Not being able to use abilities for a short while is okayish, but combining it with a stun and slow or being able to spam it makes it more annoying.

Pariah919 (commented):
I will repeat my opinion on this once again.

Stagger is an okay mechanic because stuns don't work on large castes and most causes of stagger generally have poor if no DPS, making them primarily support weapons, and generally aren't too horriffic to fight, the main issue is that there's no feedback, as you yourself state.

>There's also very little feedback on your abilities being locked from usage from weapons you might not notice some marines having since you're still WASDing around to care about the second marine shooting at you. 
>

This seems like a player feedback issue, rather than a issue with the mechanic itself outside of certain guns (Which are quickly nerfed for many reasons.)

EDIT

I'm not against making certain abilities immune to stagger as a better comprimise for escape abilities and the like, which is better than removing it from a lot of guns which basically have it as their only gimmick like Martini and Repeater.

Zar000:
Why not just have a " dizzy " icon while you're affected by stagger??

GrayRachnid:
I think stagger is fine if it wasn't always combod with a knockback and a slow, that's the issue with it.

lovegreenstuff:
how about grenades

Losenis:
Were it up to me I'd change it but given grenades tend to end up wiping marines more often than not, it might have unforeseen balance implications if reasons to use them diminished. Would need to ask.

Naaanii:
Rocket sledge could use a rework too.

code/modules/projectiles/ammo_datums.dm
```diff
@@ -1084,7 +1084,7 @@ datum/ammo/bullet/revolver/tp44
 	damage_falloff = 0.25
 
 /datum/ammo/bullet/sniper/pfc/flak/on_hit_mob(mob/M, obj/projectile/P)
-	staggerstun(M, P, knockback = 4, slowdown = 1.5, stagger = 1, max_range = 17)
+	staggerstun(M, P, knockback = 4, slowdown = 1.5, stagger = 0, max_range = 17)
```
KrabSpider, line 1087: This should stay...
It'd be absolutely worthless on large xenos, the damage is piss poor + it takes time to travel on account of being a sniper. You can't reasonably counter that but let's also just face the fact that it's only situationally useful on top of being req-locked (my bad).

The magazine is limited, the firing rate slow enough to the point where you're pretty much bound to miss your target by the time you're ready to fire again, and unlike the Martini (real talk, I strongly disagree with that having it's stagger removed as well; you only get 1 round in the chamber + a lengthy reload) actually requires you to line up your shots or be extraordinarily lucky.

Stagger having a lack of feedback should be dealt with by adding an alert icon to your HUD or by giving a balloon alert when its proc'd.

KrabSpider:
> Leave stagger on shotguns and slow firing weapons, bolt action weapons for example. I do like the fact that the cringe revolver spam won't make your screen shake like it has a epilepsy attack when you're a xeno. If I'm not mistaken stagger is the disable of abilities plus screen shake right?

The screenshake is its own thing.

TiviPlus:
Stagger as of itself is fine if you want stagger immunity for certain abilities be my guest
And feedback issues shouldn't warren removal
We want less hard CC like stuns not CC removal

PR #10566, opened 2022-07-03
Removes Crash

MisterGrimm:
## About The Pull Request
Removes the crash gamemode, since people seem to really hate it compared to other gamemodes like distress and nuclear war.

## Why It's Good For The Game
Crash is a outdated and generally disliked gamemode that only exists due to lowpop. With the inclusion of Nuclear War, Crash now has a legitimate replacement that can do everything Crash can do in both low and high pop games. There is no reason keeping this barely fun gamemode around when Nuclear War can now do the same thing, but better.

## Changelog
:cl:
del: Removed the Crash gamemode
/:cl:


Weedmaster77:
As someone that doesn't mind crash and plays at all pop levels its meant for low pop so that both side have a far chance with fewer players. Its normal only played when server is starting to get pop or when its late and pop is dropping I almost never see it in mid server up time. There is no good reason to get rid for it just because some people don't like it they can either vote for the other modes or wait till new round or not play, which most crash rounds are not that long on starting pop hours maybe 30 minutes to a hour and in late pop hours both distress and NW are normally steamrolls for one or the other side with is not fun when you have no resistant. An trying to say NW is a replacement for crash is the worse reasoning as they are only the same in objective nothing else one has REQ and PSY points the other doesn't one has a ship with medbay on ground the other doesn't also fogged silos which give xenos a place to recover so they can try to not get steamrolled and have some sort of balance to double or triple the marines to xenos there is a reason its a mode.

GrayRachnid:
Crash isn't bad, it's just broken. Fix the Larva generation on it and it'll be good. Maybe even maintain it.

MisterGrimm:
I just figured I would throw this out there, since I had a 50/50 split on whether people would care if it was removed. Ill just let it sit here for a few days before closing it, assuming maints don't want to deal with this.

Al-1ce:
Crash became a redundancy the minute nuclear war was merged. I don't think it should be entirely atomized but making it an event/admin only (like WO) would probably be for the better.

Dzhimax:
it`s bad for the game. Автор долбоёб

Lumipharon:
Crash is fun as a xeno, you just can't afford to be suicidal like you can in nuke war.

The main ship game mode functions poorly at low pop, while providing marines with loads of gear and shit that is not balanced for lowpop, while crash is fundamentally designed around lowpop. 

If crash isn't fun, people won't vote for it. If you have particular problems with it, PR changes to it.

Losenis:
Nuclear is just distress with disks. If crash was replaceable by nuclear, it'd have already been replaced by distress long ago.

Weedmaster77:
Also I like Crash to play new caste that I've never play so that its not hurting the team in NW or distress. I think it good for trying new loadouts and ideas in a field experience and not just on the ship in the range or in a round where dyeing as xenos matters.

Andrewhedge:
you say NW is suitable replacement for crash, but its wrong. On lowpop, Marines are always under-staffed on command and nobody gives a S*it about FOB cading or Reqing which all of you are desperate about. Crash is where command staff is NOT needed and people just can brrrrt away - this applies to the crash hive as well. no shrike or queen necessary. Unless there is alternative to THIS, crash has to be maintained/played. If you disagree you can cope.

Autisem:
crash is required, playing nuclear war with 3 players is anything but fun

DerpiestBruh:
soulless pr

PR #10567, opened 2022-07-04
Dead marines do not reduce number of xenos for crash

Autisem:
## About The Pull Request

Overwrites the job larva worth proc so that dead marines will not reduce total number of xenos to spawn.

Also 1 minut interval between larval checks

Its tad difficult to test this properly, it does compile atleast. Could this be TM'ed atleast once to see if it works properly?

## Why It's Good For The Game

This avoids annoying situation on crash where a few marines die killing xenos, reducing the xeno pop to a point they cannot really fight them(1 of any cast cannot solo all the marines behind cades)

The latter change because same one pinged me about it

## Changelog
:cl:
balance: Dead marines will not cap the xeno pop for crash
balance: Xeno larva check will proc every 1 minut instead of 2 minuts for crash
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -206,7 +206,7 @@
 	for(var/mob/living/carbon/human/H AS in GLOB.human_mob_list)
 		if(!H.job)
 			continue
-		if(!(H.z in z_levels) || isspaceturf(H.loc))
+		if(isspaceturf(H.loc))
 			continue
 		. += H.job.jobworth[/datum/job/xenomorph]
 
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -26,7 +26,7 @@
 	// Round start info
 	var/starting_squad = "Alpha"
 	///How long between two larva check
-	var/larva_check_interval = 2 MINUTES
+	var/larva_check_interval = 1 MINUTES
 	///Last time larva balance was checked
 	var/last_larva_check
 	bioscan_interval = 0
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -26,7 +26,7 @@
 	// Round start info
 	var/starting_squad = "Alpha"
 	///How long between two larva check
-	var/larva_check_interval = 1 MINUTES
+	var/larva_check_interval = 2 MINUTES
 	///Last time larva balance was checked
 	var/last_larva_check
 	bioscan_interval = 0
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -203,13 +203,10 @@
 /datum/game_mode/infestation/crash/get_total_joblarvaworth(list/z_levels, count_flags)
 	. = 0
 
-	for(var/i in GLOB.human_mob_list)
-		var/mob/living/carbon/human/H = i
+	for(var/mob/living/carbon/human/H AS in GLOB.human_mob_list)
 		if(!H.job)
 			continue
-		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && !H.client))
-			continue
-		if(H.status_flags & XENO_HOST)
+		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && H.client))
 			continue
 		if(!(H.z in z_levels) || isspaceturf(H.loc))
 			continue
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -206,7 +206,7 @@
 	for(var/mob/living/carbon/human/H AS in GLOB.human_mob_list)
 		if(!H.job)
 			continue
-		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && H.client))
+		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && !H.client))
 			continue
 		if(!(H.z in z_levels) || isspaceturf(H.loc))
 			continue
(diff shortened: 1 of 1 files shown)


XxBarotraumaBugReporterxX:
Will crash finally be playable ?

code/datums/gamemodes/crash.dm
```diff
@@ -199,3 +199,19 @@
...
 	xeno_hive.update_tier_limits()
+
+/datum/game_mode/infestation/crash/get_total_joblarvaworth(list/z_levels, count_flags)
+	. = 0
+
+	for(var/i in GLOB.human_mob_list)
+		var/mob/living/carbon/human/H = i
+		if(!H.job)
+			continue
+		if(count_flags & COUNT_IGNORE_HUMAN_SSD && !H.client)
```
Autisem, line 210:
```suggestion
		if(count_flags & COUNT_IGNORE_HUMAN_SSD)
```
I believe I may have to do this because if a player leaves there corpse it may not have a client. But this may have other consiquences

goose4429:
Speed merge

vilereaver41:
wait dead marines already did? what a stupid fucking thing lmfao

Pariah919 (commented):
```
larva_check_interval = 1 MINUTES
```
I don't agree with halving larva check, otherwise this seems okay.

Autisem:
> ```
> larva_check_interval = 1 MINUTES
> ```
> 
> I don't agree with halving larva check, otherwise this seems okay.

I got pinged to add it so its up for debate, I can unchange it if we feel thats needed or not

goose4429:
2 minutes seem fine if the checks actually work

code/datums/gamemodes/crash.dm
```diff
@@ -199,3 +199,19 @@
...
 	xeno_hive.update_tier_limits()
+
+/datum/game_mode/infestation/crash/get_total_joblarvaworth(list/z_levels, count_flags)
+	. = 0
+
+	for(var/i in GLOB.human_mob_list)
+		var/mob/living/carbon/human/H = i
+		if(!H.job)
+			continue
+		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.state != DEAD && !H.client))
```
Wayland-Smithy, line 210:
```suggestion
		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && !H.client))
```
smol fix

Naaanii:
Honestly, reverting the delay in which burrowed are granted would also be a good change for crash.

code/datums/gamemodes/crash.dm
```diff
@@ -199,3 +199,19 @@
...
+/datum/game_mode/infestation/crash/get_total_joblarvaworth(list/z_levels, count_flags)
+	. = 0
+
+	for(var/i in GLOB.human_mob_list)
+		var/mob/living/carbon/human/H = i
+		if(!H.job)
+			continue
+		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && !H.client))
+			continue
+		if(H.status_flags & XENO_HOST)
```
BraveMole, line 212: That check is useless, cap is dead

code/datums/gamemodes/crash.dm
```diff
@@ -199,3 +199,19 @@
 		return //Things are balanced, no burrowed needed
 	xeno_job.add_job_positions(1)
 	xeno_hive.update_tier_limits()
+
+/datum/game_mode/infestation/crash/get_total_joblarvaworth(list/z_levels, count_flags)
+	. = 0
+
+	for(var/i in GLOB.human_mob_list)
```
BraveMole, line 206: Var/mob/living/carbon/human/H AS in GLOB.human_mob_list

code/datums/gamemodes/crash.dm
```diff
@@ -199,3 +199,19 @@
...
+		var/mob/living/carbon/human/H = i
+		if(!H.job)
```
BraveMole, line 208: Are you sure that spawned dead mob have no jobs ?

code/datums/gamemodes/crash.dm
```diff
@@ -199,3 +199,19 @@
...
+			continue
+		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && !H.client))
+			continue
```
BraveMole, line 211: Then dead marines do not count ...

BraveMole (changes requested):
Does not work

code/datums/gamemodes/crash.dm
Autisem: Looking around the map and check with VV it seems like it, if its not the case then this would also be broken in the original code since I simply copied it to overwrite it

code/datums/gamemodes/crash.dm
Autisem: I mainly want to avoid that when some one dies and leaves, they would not have a client in there body there for count. So this should only happen if they ARENT dead AND dont have a client. would mean they are SSD. if they are dead then it will leave this guard clause.

code/datums/gamemodes/crash.dm
```diff
@@ -199,3 +199,16 @@
...
 	xeno_job.add_job_positions(1)
 	xeno_hive.update_tier_limits()
+
+/datum/game_mode/infestation/crash/get_total_joblarvaworth(list/z_levels, count_flags)
+	. = 0
+
+	for(var/mob/living/carbon/human/H AS in GLOB.human_mob_list)
+		if(!H.job)
+			continue
+		if(count_flags & COUNT_IGNORE_HUMAN_SSD && (H.stat != DEAD && !H.client))
```
BraveMole, line 209:
```suggestion
		if(count_flags & COUNT_IGNORE_HUMAN_SSD && !H.client && H.stat != DEAD)
```

BraveMole (changes requested):

Lumipharon:
ungas are saying that this pr is fucking up the ratio so that xenos are even worse off. I haven't seen it myself though.

PR #10568, opened 2022-07-04
Resin silos can use pheros(for now crash only)

Autisem:
## About The Pull Request

This PR aims to adds a way for silos to carry aura effects with them. For now just in context of crash.

The silo will carry its own very strong aura and be able to channel leaders aura

## Why It's Good For The Game

May be a neat feature that can be worked on in the future, but for the here and now this lets xenos recover some what quickly at silos. Which is great when a leader could not be spawned due to low pop. 

## Changelog
:cl:
add: Resin silos can now use phero's(currently only in crash)
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -80,7 +80,8 @@
 /datum/game_mode/infestation/crash/post_setup()
 	. = ..()
 	for(var/i in GLOB.xeno_resin_silo_turfs)
-		new /obj/structure/xeno/silo(i)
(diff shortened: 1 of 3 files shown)


Autisem:
I want to test localy properly before anything else, but figurd id post the code bit ahead of time

Autisem:
Maby in future we(I) can use this code to make a phero tower or smth. Should be cool

TiviPlus:
This has been discussed before and it was agreed that pheros need to be genericified before copy pasting won't fly I'm afraid

PR #10574, opened 2022-07-05
Ravagers/queen fully ignore ground fire

QualityVan:
## About The Pull Request
At the moment, ravagers and queens specifically ignore standing in ground fire, but still have the normal xeno results of walking into it. That's weird. This fixes that. Also removes a duplicate bitfield entry.

## Why It's Good For The Game
Consistency. Alternatives are:
Remove the standalone code for queen and ravager, and just have them handle fire normally.
Remove that code, but give one or both CASTE_FIRE_IMMUNE. The difference is that right now both are still vulnerable to other ignition sources (mostly projectile flamer/pp, plus any other incendiary ammo).

## Changelog
:cl:
balance: Ravager and queen more fully ignore ground flames.
/:cl:


Lumipharon:
They're definitely not supposed to be entirely fire immune, so don't do that. Considering fire armor actually works now, I would just have them handle fire normally.

Wayland-Smithy:
> Remove the standalone code for queen and ravager, and just have them handle fire normally.

This may be the best option, and then immunity can be achieved through a little armor tweak e.g. King with no sunder plus warding.

XxBarotraumaBugReporterxX:
I think fire needs a nerf on the asspect of the very easy availability of it, not on making xenos more fire resistant.

TiviPlus:
What lumi and wayland said

PR #10576, opened 2022-07-06
Orders can be used through overwatch

QualityVan:
## About The Pull Request
The buffing ones, specifically. The defend location ones already could, although they're fairly jank and could use cleanup.

## Why It's Good For The Game
Makes shipside SOs slightly more capable.

## Changelog
:cl:
add: You can use buffing orders through overwatch. They'll apply in a radius around your camera viewpoint.
/:cl:

### DIFF
--- a/code/game/objects/machinery/computer/camera_advanced.dm
+++ b/code/game/objects/machinery/computer/camera_advanced.dm
@@ -292,6 +292,9 @@
 	icon_state = "nothing"
 	var/icon_state_on = "cas_camera"
 	hud_possible = list(SQUAD_HUD_TERRAGOV)
+	var/command_aura
+	var/command_aura_ticks
+	var/command_aura_strength
 
 /mob/camera/aiEye/remote/hud/Initialize()
 	. = ..()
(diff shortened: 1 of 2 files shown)


MathiasOrtega:
Actually makes sense.

Naaanii:
Should probably reduce the effectiveness or keep some kind of limiter as the Captain's orders are a very strong buff to give, especially through just cams and without the need of actually being there since they can just cycle through cams.

Same for SO as it's SL level buffs iirc but again, kind of free and don't require them to physically be there.

Lumipharon:
"All effort incinerators"
Cool to make ship officers do a bit more, but this is kinda jank for a few reasons.
A: Does this work when comms are down?
B: Does this work with any camera? i.e. can PO's give order while in CAS, or even tad cam
C: You'll be able to jump all over the map, carrying your orders with you, which is kinda weird.

FosterGrime-Heymynameslouie:
based pr spede mehrg

Naaanii:
> "All effort incinerators" Cool to make ship officers do a bit more, but this is kinda jank for a few reasons. A: Does this work when comms are down? B: Does this work with any camera? i.e. can PO's give order while in CAS, or even tad cam C: You'll be able to jump all over the map, carrying your orders with you, which is kinda weird.

POs giving orders through CAS or tad cam seems okayish, if orders through cams didn't give the full effect as CAS can also be used as a temp overwatch. Comms shouldn't really have an effect on orders but with comms down, it'd make more sense as to make it so orders don't work but I still think it'd be okay if it wasn't the full effect.

Duping orders should be looked into however like giving orders through tad cams and then landing, giving the benefit around the tad as well, not sure if cam orders stay where the cam is even when the cam is inactive or disabled.

QualityVan:
> "All effort incinerators" Cool to make ship officers do a bit more, but this is kinda jank for a few reasons. A: Does this work when comms are down? B: Does this work with any camera? i.e. can PO's give order while in CAS, or even tad cam C: You'll be able to jump all over the map, carrying your orders with you, which is kinda weird.

What? A: Yes. Command orders do too. B: Any camera that gives you a freefloating viewpoint. Yes cas, no for tad. C: Seems fine to me. If you couldn't move with them then people would just run out of your move order.

> Duping orders should be looked into however like giving orders through tad cams and then landing, giving the benefit around the tad as well, not sure if cam orders stay where the cam is even when the cam is inactive or disabled.

With this, if you use a buff order through a camera you won't emit it yourself. The camera eye (from what I could tell) stays where it is when you leave overwatch.

> Same for SO as it's SL level buffs iirc but again, kind of free and don't require them to physically be there.

Free aside from requiring someone to be SO instead of being on the ground with a gun which can be used more than a third of the time.

QualityVan:
Implementing this with less jank. Or more jank, but in different ways.

PR #10592, opened 2022-07-10
Accuracy tweaks

Lumipharon:
## About The Pull Request
A few tweaks to the projectile accuracy system.

1. Slightly lower multiplier to hit chance for large mobs (20 instead of 25 for xeno, 40 instead of 50 for large xeno)
2. Lowered PB accuracy bonus from 50 to 30 (reminder PB range in projectile terms is normally 2 tiles)
3. Made the 'In view range" 9 tiles instead of 7 due to widescreen (currently you'd lose accuracy when shooting things at the edge of your screen horizontally but not vertically)
4. Akimbo now makes your PB range 0. This means you cannot get the PB accuracy bonus while akimbo, unless you're on top of a crit beno or something.

## Why It's Good For The Game
Improves accuracy system slightly, fixes a funny bug, makes akimbo less poo.

## Changelog
:cl:
balance: accuracy bonus for large mob size reduced slightly
balance: accuracy bonus for PB range reduced slightly
balance: Akimbo makes PB range 0 instead of 2
fix: Fixed an accuracy malus triggering for a target being out of view when it was still in widescreen view.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -1562,6 +1562,11 @@
 			iff_signal = sentry.iff_signal
 		projectile_to_fire.iff_signal = iff_signal
 	projectile_to_fire.damage_marine_falloff = iff_marine_damage_falloff
+	//no point blank bonus when akimbo
+	if(dual_wield == TRUE)
+		projectile_to_fire.point_blank_range = 0
+	else
+		projectile_to_fire.point_blank_range = projectile_to_fire.ammo.point_blank_range
 
 
 /obj/item/weapon/gun/proc/setup_bullet_accuracy(obj/projectile/projectile_to_fire, mob/user, bullets_fired = 1)
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -770,6 +770,14 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	user.sync_lighting_plane_alpha()
 	return TRUE
 
+/obj/item/attachable/scope/zoom(mob/living/user, tileoffset, viewsize)
+	. = ..()
+	//Makes the gun zoom align with the attachment, used for projectile procs
+	if(zoom)
+		master_gun.zoom = TRUE
+	else
+		master_gun.zoom = FALSE
+
 /obj/item/attachable/scope/optical/update_remote_sight(mob/living/user)
 	. = ..()
 	user.see_in_dark = 2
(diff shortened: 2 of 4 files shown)


code/modules/projectiles/gun_system.dm
```diff
@@ -1562,6 +1562,11 @@
 			iff_signal = sentry.iff_signal
 		projectile_to_fire.iff_signal = iff_signal
 	projectile_to_fire.damage_marine_falloff = iff_marine_damage_falloff
+	//no point blank bonus when akimbo
+	if(dual_wield == TRUE)
```
TiviPlus, line 1566: if(dual_wield)?

code/modules/projectiles/gun_system.dm
```diff
@@ -1562,6 +1562,11 @@
...
+		projectile_to_fire.point_blank_range = 0
+	else
+		projectile_to_fire.point_blank_range = projectile_to_fire.ammo.point_blank_range
```
TiviPlus, line 1569: I dont think I like how this is outside the gun code 
Stuff like UAVS and etc that use gun code will not respect it anymore

TiviPlus (changes requested):

code/modules/projectiles/gun_system.dm
Lumipharon: I smrt

code/modules/projectiles/gun_system.dm
Lumipharon: updated

MathiasOrtega:
Why remove PB for akimbo? If this is because of the combat shotgun funny akimbo it’s understandable. But weapons like pistols and SMGs does not make sense considering their gimmick. I feel like it would harm akimbo sensible guns more than balance the problem.

Lumipharon:
> Why remove PB for akimbo? If this is because of the combat shotgun funny akimbo it’s understandable. But weapons like pistols and SMGs does not make sense considering their gimmick. I feel like it would harm akimbo sensible guns more than balance the problem. There has already been the scatter reworks and accuracy mod having effects. Akimbo also slows fire rate already.

Akimbo just combines a whole bunch of traits together to make a balance shitsandwich, and being at close range is A: easy to do with no slowdown, and B: negates all the downsides of being akimbo in the first place. 
A change like this is targetting at the akimbo cheese without just generally punishing akimbo, because literally no issues exist with akimbo when you're at a semi reasonably distance from the target. But PB range with akimbo MP19 or AR11? You just delete xenos.

MathiasOrtega:
> 

That’s the variety of akimbo. Not to mention the downsides akimbo trades off. Not everyone wants to be shotgun or Melee Unga for CQC. Just make certain guns unable to be double’d like the AR11 and similar. And really MP19? That’s it’s whole niche isn’t it?

PR #10594, opened 2022-07-10
C4 smoke reduction

Lumipharon:
## About The Pull Request
The previous supplier of C4 has been given the boot due to their excessively smokey explosions.

C4 now has a 1 tile smoke cloud on detonation, and only very a very short duration.

Currently it lasts for a very long time, making it impossible to use with any sort of aggression, as you'd have to push through the smoke blind or wait forever for it to clear.

## Why It's Good For The Game
Makes C4 a tiny bit more useful, mainly for HvH

## Changelog
:cl:
balance: C4 causes less smoke for a much shorter duration on detonation
/:cl:


Naaanii:
I'd say keep the smoke radius the same but lower duration is neat.

Pariah919 (approved):
Yeah seems good to me, makes the smoke attacker rather than defender sided.

PR #10595, opened 2022-07-10
Converts auras to a subsystem

QualityVan:
## About The Pull Request
Changes auras to work through a subsystem, rather than individual mobs each emitting them. Use SSaura's add_emitter proc to start a new one. An aura finishes when the atom it's centered on is destroyed, when it runs out of duration, or when the aura itself is manually qdeld.
Lets humans send buff orders through overwatch. You'll be able to see active auras on the overwatch eye.

## Why It's Good For The Game
Code improvement.
Enables arbitrary aura emitters, like overwatch camera eyes or xeno towers.
Makes SO slightly less do nothing.

## Changelog
:cl:
refactor: Anything can emit auras. both buff orders and pheromones use the same system, although in slightly different ways.
add: Humans using overwatch computers can transmit buff orders through them.
/:cl:


Lumipharon:
I'm gonna PR the faction order thing, since that's gonna be real relevant to HvH, honk.

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	flags = SS_NO_INIT
+	///Stores currently active aura_bearer datums
+	var/list/active_auras = list() //We can't use a normal subsystem processing list because as soon as an aura_bearer leaves the list it needs to die
+
+/datum/controller/subsystem/aura/fire(resumed)
+	for(var/datum/aura_bearer/bearer AS in active_auras)
+		bearer.process()
+
+///Use this to start a new emitter with the specified stats. Returns the emitter in question, use stop_emitting() to end early.
+/datum/controller/subsystem/aura/proc/add_emitter(atom/center, type, range, strength, duration)
```
BraveMole, line 14: you should be able to emit multiple auras at once without doing the range finding twice

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	///String indicating aura variety
+	var/aura_type
+	///How far from our center we apply
+	var/range
+	///Multiplier to aura effectiveness
+	var/strength
+	///How many ticks we have left, 15 by default.
+	var/duration
+	///If our aura_type is in this, we'll pulse to humans
+	var/list/human_auras = list(AURA_HUMAN_MOVE, AURA_HUMAN_HOLD, AURA_HUMAN_FOCUS)
```
BraveMole, line 34: uh, shouldn't this be a const or at the very least a static?

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	var/list/xeno_auras = list(FRENZY, WARDING, RECOVERY)
+	///Whether we care about xenos
+	var/affects_xenos = FALSE
+
+/datum/aura_bearer/New(atom/aura_center, aura_name, aura_range, aura_strength, aura_duration)
+	..()
+	center = aura_center
+	range = aura_range
+	strength = aura_strength
+	duration = aura_duration || 15
```
BraveMole, line 47: why 15

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+		affects_humans = TRUE
+	if(xeno_auras.Find(aura_type))
+		affects_xenos = TRUE
+	RegisterSignal(center, COMSIG_PARENT_QDELETING, .proc/stop_emitting)
+	SSaura.active_auras += src
+
+///Center gets destroyed, we run out of duration, or any other reason to finish. Perish immediately.
+/datum/aura_bearer/proc/stop_emitting()
+	SIGNAL_HANDLER
+	SEND_SIGNAL(src, COMSIG_AURA_FINISHED) //For visuals, since mob huds need to track what and when they're personally emitting.
```
BraveMole, line 59: why isn't this on Destroy ?

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	qdel(src)
+
+/datum/aura_bearer/Destroy()
+	if(center)
+		UnregisterSignal(center, COMSIG_PARENT_QDELETING)
```
BraveMole, line 64: pretty sure signals are cleaned on Destroy by themselves

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	return ..()
+
+/datum/aura_bearer/process()
+	if(affects_humans)
+		pulse_humans()
+	if(affects_xenos)
+		pulse_xenos()
+	if(duration < 0) //Negative duration means infinite. Usually for pheromones.
+		return
+	duration--
```
BraveMole, line 76: it fires every 2 secs (which is the right number). So duration is not in ticks like said in the description

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	if(!istype(center))
+		return
+	if(!type || !range || !strength)
+		return
+	return new /datum/aura_bearer(center, type, range, strength, duration)
+
+///The thing that actually pushes out auras to nearby mobs.
+/datum/aura_bearer
+	///What we emit from
+	var/atom/center
```
BraveMole, line 24: emitter, not center

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	///String indicating aura variety
+	var/aura_type
```
BraveMole, line 26: consider making that a string as you could be emitting multiple type at once.

code/controllers/subsystem/aura.dm
```diff
@@ -0,0 +1,98 @@
...
+	for(var/datum/aura_bearer/bearer AS in active_auras)
+		bearer.process()
+
+///Use this to start a new emitter with the specified stats. Returns the emitter in question, use stop_emitting() to end early.
+/datum/controller/subsystem/aura/proc/add_emitter(atom/center, type, range, strength, duration)
+	if(!istype(center))
+		return
+	if(!type || !range || !strength)
+		return
+	return new /datum/aura_bearer(center, type, range, strength, duration)
```
BraveMole, line 19: do you plan to use that datum elsewhere than the SS ? return the datum is probably a good idea, but you need to add it to the active emiter list here

BraveMole (changes requested):

code/controllers/subsystem/aura.dm
QualityVan: the bearer datum adds itself during its new, I'm returning it because xeno pheros work off plasma upkeep instead of a flat duration so they need to be able to end their own pheromones

code/controllers/subsystem/aura.dm
QualityVan: there's no instance right now where you can have multiple at once and they'd have to have the same strength, but that's probably fine and wouldn't hurt to have anyway

code/controllers/subsystem/aura.dm
QualityVan: can't be a const since it's an instantiated object, but static yeah

code/controllers/subsystem/aura.dm
QualityVan: i was coming from marine orders which use that as the default, but they can handle that themselves instead.

code/controllers/subsystem/aura.dm
Lumipharon: Isn't a phero emitter that is also a leader emitting 2?

QualityVan: that's more like the queen emitting once from the leader and the leader emitting once. the two pheros don't share strength, range, or whether they're active or not

code/game/objects/machinery/overwatch.dm
```diff
@@ -795,8 +797,30 @@ GLOBAL_LIST_EMPTY(active_cas_targets)
...
+	if(aura_target != src)
+		command_aura = null //We aren't emitting personally, so we don't want the on-mob buff icon
+	else
+		RegisterSignal(src, COMSIG_AURA_FINISHED, .proc/end_command_aura)
+	command_aura_cooldown = addtimer(CALLBACK(src, .proc/end_command_aura_cooldown), 45 SECONDS)
+
 	update_action_buttons()
 
+///Choose what we're sending a buff order through
+/mob/living/carbon/human/proc/pick_order_target()
```
BraveMole, line 810: not a fan of that.
I don't think this should be oop

BraveMole (changes requested):

PR #10596, opened 2022-07-10
You can no longer give orders to other factions

Lumipharon:
## About The Pull Request
As QV pointed out, currently orders just apply to all humans in range... even hostile ones.
Makes it so that orders only apply to those within your own faction. So you won't buff SOM by accident, but also can't buff free lancers for example.

If people like being able to order friendly factions, maybe someone can code that in the future, but currently factions are only designated friendly/neutral/hostile, but this is fixed, and not relative to other specific factions.

## Why It's Good For The Game
HvH will be a bit silly when you're buffing your enemies.

## Changelog
:cl:
fix: Orders no longer apply to humans in factions other than your own
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_status_effects.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_status_effects.dm
@@ -24,12 +24,13 @@
 			command_aura_strength = skills.getRating("leadership") - 1
 			var/command_aura_range = round(4 + command_aura_strength * 1)
 			for(var/mob/living/carbon/human/H in range(command_aura_range, src))
-				if(command_aura == "move" && command_aura_strength > H.mobility_new)
-					H.mobility_new = command_aura_strength
-				if(command_aura == "hold" && command_aura_strength > H.protection_new)
-					H.protection_new = command_aura_strength
-				if(command_aura == "focus" && command_aura_strength > H.marksman_new)
-					H.marksman_new = command_aura_strength
+				if(H.faction == faction) //You can only give orders to people in your own faction
+					if(command_aura == "move" && command_aura_strength > H.mobility_new)
+						H.mobility_new = command_aura_strength
+					if(command_aura == "hold" && command_aura_strength > H.protection_new)
+						H.protection_new = command_aura_strength
+					if(command_aura == "focus" && command_aura_strength > H.marksman_new)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/life/handle_status_effects.dm
```diff
@@ -24,12 +24,13 @@
...
-				if(command_aura == "hold" && command_aura_strength > H.protection_new)
-					H.protection_new = command_aura_strength
-				if(command_aura == "focus" && command_aura_strength > H.marksman_new)
-					H.marksman_new = command_aura_strength
+				if(H.faction == faction) //You can only give orders to people in your own faction
+					if(command_aura == "move" && command_aura_strength > H.mobility_new)
+						H.mobility_new = command_aura_strength
+					if(command_aura == "hold" && command_aura_strength > H.protection_new)
+						H.protection_new = command_aura_strength
+					if(command_aura == "focus" && command_aura_strength > H.marksman_new)
```
TiviPlus, line 32: Side note this could be an else if

code/modules/mob/living/carbon/human/life/handle_status_effects.dm
```diff
@@ -24,12 +24,13 @@
...
 			var/command_aura_range = round(4 + command_aura_strength * 1)
 			for(var/mob/living/carbon/human/H in range(command_aura_range, src))
-				if(command_aura == "move" && command_aura_strength > H.mobility_new)
-					H.mobility_new = command_aura_strength
-				if(command_aura == "hold" && command_aura_strength > H.protection_new)
-					H.protection_new = command_aura_strength
-				if(command_aura == "focus" && command_aura_strength > H.marksman_new)
-					H.marksman_new = command_aura_strength
+				if(H.faction == faction) //You can only give orders to people in your own faction
+					if(command_aura == "move" && command_aura_strength > H.mobility_new)
```
TiviPlus, line 28: early continue?

TiviPlus (changes requested):

code/modules/mob/living/carbon/human/life/handle_status_effects.dm
Lumipharon: done

TiviPlus (approved):

PR #10601, opened 2022-07-11
Volkite powerpack fix

Lumipharon:
## About The Pull Request
Currently you can only tactical reload a culverin by dragging the powerpack to the gun (which is kinda funny since if you get brain damage now, you can no longer operate it).

The same also applies to the caliver if used with a powerpack of course.

Added attackby code to the culverin so you can now normal reload it (click on the powerpack with the gun).
I also moved cell charging from click drag to power pack to just clicking the powerpack with the cell.

## Why It's Good For The Game
Braindamaged ungas should be able to use their guns
bugfix good

## Changelog
:cl:
fix: Volkite culverins and calivers can now be normally reloaded by clicking the powerpack with the gun, instead of only being tac reloadable
add: Volkite powerpacks now recharge cells by clicking the powerpack with the cell instead of click dragging
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/energy.dm
+++ b/code/modules/projectiles/magazines/energy.dm
@@ -123,15 +123,16 @@
 		gun.reload(src, user)
 		return
 
-	if(istype(I, /obj/item/cell))
-		if(I != user.r_hand && I != user.l_hand)
-			to_chat(user, span_warning("[I] must be in your hand to do that."))
-			return
-		var/obj/item/cell/D = I
-		var/charge_difference = D.maxcharge - D.charge
-		if(charge_difference) //If the cell has less than max charge, recharge it.
-			var/charge_used = use_charge(user, charge_difference) //consume an appropriate amount of charge
-			D.charge += charge_used //Recharge the cell battery with the lower of the difference between its present and max cap, or the remaining charge
-			D.update_icon()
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/magazines/energy.dm
```diff
@@ -114,20 +114,24 @@
...
-	var/obj/item/cell/D = W
-	var/charge_difference = D.maxcharge - D.charge
-	if(charge_difference) //If the cell has less than max charge, recharge it.
-		var/charge_used = use_charge(user, charge_difference) //consume an appropriate amount of charge
-		D.charge += charge_used //Recharge the cell battery with the lower of the difference between its present and max cap, or the remaining charge
-		D.update_icon()
-	else
-		to_chat(user, span_warning("This cell is already at maximum charge!"))
+	if(istype(I, /obj/item/cell))
+		if(I != user.r_hand && I != user.l_hand)
```
TiviPlus, line 127: early return

TiviPlus (changes requested):

code/modules/projectiles/magazines/energy.dm
Lumipharon: done

TiviPlus (approved):

PR #10603, opened 2022-07-12
Splint break sound range limit

Lumipharon:
## About The Pull Request
Being able to hear your splint break is a super useful addition to the game, but currently the sound has a theoretical max range of 100 tiles, which effectively means you'll hear everyone's splints breaking which is uh, not ideal since you can't tell if it's yours or not.

Makes the sound effect have a range of 1, so you'll only hear other people's splints break if you're right next to them (So xenos still get to hear that juicy pop when they melee someone's splints off).

## Why It's Good For The Game
Hearing everyones splints breaking across the map is funny but not useful.

## Changelog
:cl:
fix: Splint break sounds no longer carry over large distances
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -195,7 +195,7 @@
 		if(splint_health <= 0)
 			remove_limb_flags(LIMB_SPLINTED)
 			to_chat(owner, span_userdanger("The splint on your [display_name] comes apart!"))
-			playsound(owner, 'sound/items/splint_break.ogg', 100, falloff=5)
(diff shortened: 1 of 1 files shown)


FosterGrime-Heymynameslouie:
I hear every bone you break

Losenis:
I felt a great disturbance in the corps. As if millions of splints suddenly cried out in terror and were suddenly silenced.

TiviPlus:
> I felt a great disturbance in the corps. As if millions of splints suddenly cried out in terror and were suddenly silenced.

Master corpsman there's too many of them what should we do?

TiviPlus (approved):

PR #10607, opened 2022-07-17
Adds a little Tox power to Sent neuro spit.

vilereaver41:
## About The Pull Request

Makes sentinel neuro a bit better.

## Why It's Good For The Game

Currently Sentinel neuro spit is very very slow acting on marines, the only marines that die from neuro are either decapped with no way back or afk. its also very easy to purge, it also has no use against robots.

This gives Sentinel neuro spit some Tox power so the sentinel can be rewarded for good hits rather then feeling they do nothing. it is currently set at 0.3 for now to see how it affects gameplay.

Marines currently are having easy times killing xenos whilst xenos have no stopping power because of the new gear marines get, this aims to give xenos a bit more punch for their risk.

## Changelog
:cl:
balance: Neuro spit for sentinel = 0.3 Tox power.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -2353,7 +2353,7 @@ datum/ammo/bullet/revolver/tp44
 	bullet_color = BOILER_LUMINOSITY_AMMO_NEUROTOXIN_COLOR
 	reagent_transfer_amount = 30
 	///On a direct hit, how long is the target paralyzed?
-	var/hit_paralyze_time = 1 SECONDS
+	var/hit_paralyze_time = 2.5 SECONDS
 	///On a direct hit, how much do the victim's eyes get blurred?
 	var/hit_eye_blur = 11
 	///On a direct hit, how much drowsyness gets added to the target?
@@ -2465,7 +2465,7 @@ datum/ammo/bullet/revolver/tp44
 	damage_type = BURN
 	penetration = 40
 	bullet_color = BOILER_LUMINOSITY_AMMO_CORROSIVE_COLOR
-	hit_paralyze_time = 1 SECONDS
+	hit_paralyze_time = 2 SECONDS
 	hit_eye_blur = 1
 	hit_drowsyness = 1
 	reagent_transfer_amount = 0
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -470,7 +470,7 @@
 	custom_metabolism = REAGENTS_METABOLISM * 2
 	overdose_threshold = 10000 //Overdosing for neuro is what happens when you run out of stamina to avoid its oxy and toxin damage
 	scannable = TRUE
-	toxpwr = 0
+	toxpwr = 0.3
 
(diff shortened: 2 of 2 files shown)


QualityVan:
This doesn't change sent spit, it changes neuro as a whole

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2353,7 +2353,7 @@ datum/ammo/bullet/revolver/tp44
 	bullet_color = BOILER_LUMINOSITY_AMMO_NEUROTOXIN_COLOR
 	reagent_transfer_amount = 30
 	///On a direct hit, how long is the target paralyzed?
-	var/hit_paralyze_time = 1 SECONDS
+	var/hit_paralyze_time = 2.5 SECONDS
```
QualityVan, line 2356: Leftovers from whatever you branched this off?

vilereaver41: woops

Naaanii:
Neuro is arguably better for sent, but more of how bad its acid spit is. If anything, I'd buff its acid spit maybe but this seems okay. I'd even be down for upping the tox damage as like you said, it's only useful against humans, and remove the remnant cloud effect for neuro so it's meant for direct hits only.

Al-1ce:
> Currently Sentinel neuro spit is very very slow acting on marines


This is intended. Neurotoxin across all castes is the same, mind you. If you were to adjust the variable for paralyze time, rather than the effects of the chemical itself you're giving a buff to neuro as a whole. Need I not mention that this is a T1 afterall, with access to the same neurotoxin that the Praetorian has? (Light and regular Neuro are identical in stats)


>the only marines that die from neuro are either decapped with no way back or afk


Neurotoxin isn't supposed to be lethal by itself. That's not to say it isn't lethal, -it is, but it takes a ridiculous amount of neurotoxin alone to kill a marine. 


>It's also very easy to purge


Hypervene by itself incurs heavy pain, slowdown, and potential vomiting. In an ideal situation, that's bad enough on it's own. Not to mention most marines don't actually carry hyper. 


>It also has no use against robots.


Intended feature. The exchange for this immunity is that acid does _much_ more damage to robots than regular marines.


----

>If anything, I'd buff its acid spit maybe but this seems okay. I'd even be down for upping the tox damage as like you said, it's only useful against humans, and remove the remnant cloud effect for neuro so it's meant for direct hits only.


I don't think the gap between spitter and sentinel is large enough to warrant an acid buff. It doesn't make much sense to increase the toxicity damage of this T1 just because robots are immune to neurotoxin. The damage multiplier for acid -- even weak acid is very dangerous for any robots looking to get close to a sentinel.

As for the remnant cloud, it hardly makes a difference. You'd need to stand completely still in it for a couple solid seconds to even notice it was dangerous.

QualityVan:
> Intended feature. The exchange for this immunity is that acid does _much_ more damage to robots than regular marines.

If you're talking about the burn mod on robots, that doesn't exist and hasn't since they were merged.

Naaanii:
Yeah, rebalance to robots might just be a better solution

BraveMole:
Things were said by Al1ce

PR #10612, opened 2022-07-19
Allow multiple choice voting

BraveMole:
## About The Pull Request

Example, you can vote for 2 out of 4 maps during map voting.

## Why It's Good For The Game

This type of vote is superior in all points to single choice votes.

See https://en.wikipedia.org/wiki/Approval_voting 

For example, let's say 60% of the players want to play HVX, but 35% vote distress and 25% vote nuclear war. This let combat patrol win despite only having 40% in support of HVH. To avoid that HVX lover would strategically vote for either Distress or Nuclear war, which can lead to weird effect like everyone voting nuclear war despite it being less popular than distress in our example.

Same can be said with maps, etc. Multiple voting system allow people to just vote for their favorites (or least hated) choice without having to strategize

## Changelog
:cl:
add: Allow multiple choice voting
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -190,6 +190,7 @@ SUBSYSTEM_DEF(vote)
 	if(multiple_vote)
 		choices_by_ckey[usr.ckey] += vote
 	else
+		choices[choices[choices_by_ckey[usr.ckey][1]]]--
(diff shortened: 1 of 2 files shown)


Lumipharon:
First past the post voting: How minority parties have clung to power for centuries.

TiviPlus (approved):

PR #10626, opened 2022-07-24
change unnecessary /obj/ typepaths to /atom/movable.

SplinterGP:
## About The Pull Request
Some stuff like effects used /obj/ for no reason at all other than they just did, this PR basically makes them be child of /atom/movable instead which makes more sense, some stuff like xeno weeds which use /obj/ stuff like integrity have remained /obj/'s


SplinterGP:
i used some replace all's so there's probably some stuff that maybe should not be changed

code/game/objects/effects/decals/misc.dm
```diff
@@ -1,24 +1,24 @@
 
 // Used for spray that you spray at walls, tables, hydrovats etc
```
SplinterGP, line 2:
```suggestion
// Used for spray that you spray at walls, tables, hydrovats etc.
```

PR #10634, opened 2022-07-26
Adds normal smoke grenades

Lauwma:
## About The Pull Request

This Pull request add normal smoke grenades that produces thick smoke that hides terrain, xenos and marines.

## Why It's Good For The Game

They definitely have some uses because they allow to hide objects such as other grenades, tarps, mines, dropped guns. 

## Changelog
:cl:
add: Adds back good old smoke grenades
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -112,6 +112,7 @@
 			/obj/item/explosive/grenade/m15 = 30,
 			/obj/item/explosive/grenade/sticky = 125,
 			/obj/item/explosive/grenade/incendiary = 50,
+			/obj/item/explosive/grenade/smokebomb = 25,
 			/obj/item/explosive/grenade/smokebomb/cloak = 25,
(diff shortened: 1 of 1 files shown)


Losenis:
They hide the screams of the people you're FFing with an MG08.

Speedmerge.

Al-1ce:
>They definitely have some uses because they allow to hide objects such as other grenades, tarps, mines, dropped guns.



That's intended to be the only downside to regular smoke grenades. After seeing a couple variations of this tested already before I can assure you they look (and preform) like ass.

Lauwma:
They're just grenades that are worse than cloak in most situations but that can allow interesting plays in others. They're also grenades we used to have before the vendor rework, grenades that can be looted. 

During 2 rounds of the testmerge I've seen 2 good uses like hiding a turret in smoke during a siege, throwing the smoke in the maze so xenos don't see where they are heading and are forced to pull back , tarping inside of it to eventually survive if you are on your own, combat mine layering (very funny).

PR #10644, opened 2022-07-28
Fix off-set of spitter sprites

novaepee:
## About The Pull Request

image

## Why It's Good For The Game

Code was adjusted for when spitter had 48x48 sprites, not 2x2 sprites. This changes that.

## Changelog
:cl:
fix: Fix off-set of spitter sprites
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/spitter/spitter.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/spitter/spitter.dm
@@ -7,8 +7,8 @@
 	health = 180
 	maxHealth = 180
 	plasma_stored = 150
-	pixel_x = -12
-	old_x = -12
+	pixel_x = -16
+	old_x = -16
 	tier = XENO_TIER_TWO
 	upgrade = XENO_UPGRADE_ZERO
 	inherent_verbs = list(


MiniMeatwad:
The spitter is as fragile as cardboard, so the new beefy and warrior-like variant is misleading. The weak and slender variant was more suitable to convey how weak the spitter is.

Inanichts:
> The spitter is as fragile as cardboard, so the new beefy and warrior-like variant is misleading. The weak and slender variant was more suitable to convey how weak the spitter is. Visual balance is broken, I am going to sue you

You're just too used to the older spitter, the hivelord is quite fragile and it's sprite is basically warrior, carrier its pretty big, yet fragile too, and they're all T2. If you were complaining about the sentinel sprite it would make more sense, since its a T1.

Al-1ce:
>You're just too used to the older spitter, the hivelord is quite fragile and it's sprite is basically warrior, carrier its pretty big, yet fragile too, and they're all T2. If you were complaining about the sentinel sprite it would make more sense, since its a T1.

Carrier looks more spindly than anything else. The danger of huggers makes that neglible. Have you ever even seen a hivelord? They are the squishiest looking xeno we have by far. All that size is for resin, not armor.

DerpiestBruh:
TBH my only complaint about the sprite itself is that there's a notable difference with their "lighting", which looks off sometimes when in a group of xenos. Though I am not sure if I can really say much about that, I'm not a spriter.

Al-1ce:
> oh fuck i replied to a merged pr


This will likely be changed/removed when an overhaul for visual balance shows up. It's not set in stone you retard.

PR #10648, opened 2022-07-29
Give attack, defend, and retreat order to FC, SL, and aSL (also CO, SO, CSE, RO, and CMO)

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

Coordinating marines is important. Sergeants are the backbone of the military, so it is the mission of the SL to coordinate marines so that they can eliminate the xenomorphs. Sometimes, it's like herding cats, so SLs need help with this. In extension of the SL, the FC needs this as well.

## Changelog
:cl:
add: Give attack, defend, and retreat order to FC, SL, and aSL (also CO, SO, CSE, RO, and CMO)
balance: better coordination with marines, hopefully
/:cl:


crayonLover09:
Thats gonna be a lot of buttons at the top of the hud since you're going to have the 3 orders and also the 3 buffs. If you could consolidate them into the menu xenos have for picking weed type, I think that would be better.

Lumipharon:
> Thats gonna be a lot of buttons at the top of the hud since you're going to have the 3 orders and also the 3 buffs. If you could consolidate them into the menu xenos have for picking weed type, I think that would be better.

This. Also honestly not sure how useful it is anyway - what difference does an SL get from these specific orders as opposed to the existing rally order? Overwatch is different since you can place it where ever, but the dude in the field can only place it where they physically are.

code/datums/actions/order_action.dm
```diff
@@ -97,29 +101,52 @@
...
+		owner.say(message)
+
 /datum/action/innate/order/rally_order
 	name = "Send Rally Order"
 	action_icon_state = "rally"
 	verb_name = "rally to"
 	arrow_type = /obj/screen/arrow/rally_order_arrow
-	///What skill is needed to have this action
 	keybind_signal = COMSIG_KB_RALLYORDER
-	var/skill_name = "leadership"
```
novaepee, line 120: I remove this here so that the other orders can inherit.

PR #10651, opened 2022-07-30
More volkite tweaks

Lumipharon:
## About The Pull Request
Some further tweaks to volkite weaponry, mainly aimmed at making them a bit less oppressive at longer range.

- Volkite projectiles are less accurate (5 bonus accuracy instead of 15 like lasers)
- Volkite projectiles do less deflagrate damage (25% less for the charger, 20% less for the others)
- Increased the damage falloff on the Serpenta, Caliver and Culverin (as well as less damage at range, this also reduces the chance of deflagrate triggering).
- Reduced the caliver magazine capacity to 40 shots from 50 (backpack is unchanged)
- Reduced the accurate range of the Caliver and Culverin

## Changelog
:cl:
balance: Slightly reduced the accuracy of volkite projectiles, and decreased the damage of deflagration
balance: Volkite weapons have higher damage falloff
balance: Reduced the accurate range of the VX-33 Caliver and VX-42 Culverin
balance: Reduced the magazine capacity of the VX-33 Caliver
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -2190,11 +2190,12 @@ datum/ammo/bullet/revolver/tp44
 	shell_speed = 4
 	accuracy_var_low = 5
 	accuracy_var_high = 5
+	accuracy = 5
 	point_blank_range = 2
 	damage = 20
 	penetration = 10
 	sundering = 2
-	fire_burst_damage = 20
+	fire_burst_damage = 15
 
 	//inherited, could use some changes
 	ping = "ping_s"
@@ -2207,17 +2208,17 @@ datum/ammo/bullet/revolver/tp44
 
 /datum/ammo/energy/volkite/medium
 	max_range = 25
-	accurate_range = 15
+	accurate_range = 12
 	damage = 30
 	accuracy_var_low = 3
 	accuracy_var_high = 3
-	fire_burst_damage = 25
+	fire_burst_damage = 20
 
 /datum/ammo/energy/volkite/heavy
 	max_range = 35
-	accurate_range = 18
+	accurate_range = 12
 	damage = 25
-	fire_burst_damage = 25
+	fire_burst_damage = 20
 
 /datum/ammo/energy/volkite/light
 	max_range = 25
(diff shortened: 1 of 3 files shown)


DerFlammenwerfer:
I feel obligated to address the elephant in the room. While it should reduce the effectiveness of long range spraying, this PR does everything but address the DPS disparity between volkite weapons and ballistic weapons. SOM's volkite weapons, quite simply, have double, or even triple the effectiveness against Marines than ballistic weapons (the Majority of Terragov's loadouts) against SOM. This is because SOM armors have equal or greater ballistic protection than Marine armor does, and their weapons check against energy armor.

Lumipharon:
> I feel obligated to address the elephant in the room. While it should reduce the effectiveness of long range spraying, this PR does everything but address the DPS disparity between volkite weapons and ballistic weapons. SOM's volkite weapons, quite simply, have double, or even triple the effectiveness against Marines than ballistic weapons (the Majority of Terragov's loadouts) against SOM. This is because SOM armors have equal or greater ballistic protection than Marine armor does, and their weapons check against energy armor.

They likely will require further tweaking, but the difference is by design, and is not solely based around DPS. If you're only looking at DPS, then ballistics shouldn't even be in the discussion, as lasers are entirely superior in that regard due to som armor values. (laser carbine is a 3 shot crit in scatter mode even against vets, for example)

Ballistics however cause crippling long term effects. Getting shot by volkite means you heal with some chems. Getting shot by ballistics means you need a medic to remove the shrap out of your head or you're gonna constantly take damage and pain, and go blind/get brain damage. The ability to inflict shrap is not reduced over distance - you can happily shoot 20 tiles away and cause just as much shrap as you would have at pb range, where as deflagrate scales with range (and will do so more with this pr).

Ballistics (also lasers) have underbarrel weapons compared to volkites as well. A stickynade/stunnade to the face is a significant factor.

Fundamentally SOM are incentivised to be as aggressive as possible, while marines are rewarded for being cautious and defensive. I.e. deflag doesn't trigger on objects, so against cades or other cover are notably worse than ballistics, marines have mg27 with miniscope, they can aim mode past each so can apply more damage in enclosed spaces etc.

So yes, 100% volkite is strong in simple damage, but it's a fundamental mistake to try and challenge that in a situation that literally gives them the most advantage. There's a big mindset shift that is needed, that hurts marines more than SOM, as SOM can basically play the same as marines are used to in HvX (i.e. ungarush), while marines have to adjust their playstyle considerably.

BraveMole (approved):
As a general note I won't give my opinion on balance tweaks from Lumi (for HvH) and just merge

PR #10654, opened 2022-07-31
Squish shitcode and a future bug

Lumipharon:
## About The Pull Request
Projectile code still had a funny bit of code where it did a return prob(.). This works fine in 514.1584, but in 1585 this breaks entirely, meaning it would be impossible to shoot past objects.

Fixes this bug before it becomes an actual problem, and clean up the shitcode.

## Changelog
:cl:
fix: fixes a projectile bug for a future version of byond
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -656,7 +656,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		return TRUE
 	if(proj.distance_travelled <= proj.ammo.barricade_clear_distance)
 		return FALSE
-	. = coverage //Hitchance.
+	var/hit_chance = coverage //base chance for the projectile to hit the object instead of bypassing it
 	if(flags_atom & ON_BORDER)
 		if(!(cardinal_move & REVERSE_DIR(dir))) //The bullet will only hit if the barricade and its movement are facing opposite directions.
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -656,7 +656,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 		return TRUE
 	if(proj.distance_travelled <= proj.ammo.barricade_clear_distance)
 		return FALSE
-	. = coverage //Hitchance.
+	var/hit_chance = coverage //base chance for the projectile to hit the object instead of bypassing it
```
BraveMole, line 659: why do you even need a new var when coverage does the trick

BraveMole: Also there should be no difference between . and a "normal" var.

If there is, you really need to report that to lummox

BraveMole (changes requested):

code/modules/projectiles/projectile.dm
Lumipharon: I bugreported it.

The var is modified before being returned. If I use coverage directly that would be permanently changing the coverage var of the object.

Lumipharon: Oh he actually responded. 
image

PR #10663, opened 2022-08-01
Revert /obj/effect change to atom/movable [MDB ignore]

SplinterGP:
## About The Pull Request
Realizes /atom/movable childs can't be mapped in mapping tools, dind't know that since i dind't use mapping much, lot of stuff under /obj/effect needs to be mappeable and if there's anything that could be /atom/movable i'll do later.

MDB ignore because too many maps MDB wont even finish that task


SplinterGP:
dind't revert mecha_wreckage because no need to

PR #10668, opened 2022-08-02
DB recoil adjustment

Lumipharon:
## About The Pull Request
This PR attempts to nerf DB juggling without nerfing the DB itself, or slugs.

DB juggling has been a bit problematic for a while now, as being able to apply 800 damage with very high ap, sunder, stagger and slow in a few seconds (enough to kill an ancient crusher confortably), is frankly not ideal.

Neither the DB or the slugs that are used are problematic in of itself however, so this PR increases the DURATION of recoil, the the horizontal screenshake.

What this means is that a single DB is perfectly fine to fire on it's own, but if you rapidly juggle them, the recoil essentially builds up, making it harder to actually land follow up shots.

I'm fucking terrible, but here's a video to show what it looks like.

https://user-images.githubusercontent.com/7869430/182292671-ef5fdfb3-0b55-492a-843a-cc74459e90d6.mp4

## Changelog
:cl:
balance: Degraded the ability to juggle double barrel shotguns by increasing the duration and horizontal shake of its recoil
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -221,6 +221,8 @@
 	scatter_unwielded = 10
 	recoil = 2
 	recoil_unwielded = 4
+	recoil_backtime_multiplier = 3
+	recoil_deviation = 35
 
 
 //-------------------------------------------------------


QualityVan:
This doesn't actually make chain DBs less able to permadisable, right? It just makes it harder to use?

Lauwma:
Make it not able to shoot if not wielded

helper-2:
Probably, finest way to nerf DB

Al-1ce:
You're firing while unwielded, or in the process of wielding the DB in the video. That alone is having an effect on the recoil.

That's fine for a portrail of how this might look in practice (most jugglers don't wait for wield timers), but in this instance that might be the only thing to do to avoid the added recoil here. ..Which would mean the problem is only slightly mitigated.

Pariah919:
This doesn't really solve the problem, honestly.

Naaanii:
This does nothing to nerf the actual issue with the db and just looks like a placebo nerf.

Lumipharon:
> This does nothing to nerf the actual issue with the db and just looks like a placebo nerf.

What would you call the actual issue with the DB?
A single DB is not problematic, slugs in isolation are not problematic, the issue is being able to fire lots of slugs very quickly. This PR makes it harder to rapid fire slugs in a way that doesn't touch weapon/projectile stats, as there's no way of doing that without punishing non juggling use. If someone can reliably hit a moving target with escalating randomised recoil, they are hell of a lot better than me, and I wouldn't consider it a problem at that point.

Losenis:
I said it in discord but I don't believe I've said it in github, so I'll throw in my two cents.

In practice the issue is the firerate. For reference, T35 firerate is 20, T39 is 14, DB is 5. The lengthy manual reload of the DB benefits just skipping the reloading part altogether and packing multiple of them to take advantage of that 5 firerate. A change in firerate to make it 10 for example, randomly picked, wouldn't perfectly solve juggling but it'd still increase considerably the time before the second shot of the db comes out. Given half the shots of a juggling dber are the second of any given single db, it'd have considerable effect and alleviate it somewhat. I don't believe anyone really has an ideal solution that cuts down on juggling for good without sacrificing too much on other aspects.

I've been told it'd also nerf normal usage of db but I believe the nerf is only harsh on paper. In practice, you generally don't want to immediately shoot the second shell since if it misses you're going to have to spend a lengthy process reloading a gun that is mechanically complex (comparatively) with no tacreload options. Actually shooting the second shell as soon as the firerate allows isn't as viable. Even in the worst case scenario, it'd still make it exceptionally faster than a T35 which'd still have the same per-shell damage due to no damage malus, unlike the T39.

Naaanii:
> > This does nothing to nerf the actual issue with the db and just looks like a placebo nerf.
> 
> What would you call the actual issue with the DB? A single DB is not problematic, slugs in isolation are not problematic, the issue is being able to fire lots of slugs very quickly. This PR makes it harder to rapid fire slugs in a way that doesn't touch weapon/projectile stats, as there's no way of doing that without punishing non juggling use. If someone can reliably hit a moving target with escalating randomised recoil, they are hell of a lot better than me, and I wouldn't consider it a problem at that point.

The issue is being able to carry 3 DBs and fire them all in quick succession. A decent nerf would be to remove the ability to carry 3 at once. A nerf that actually works is removing the ability to fire while wielding since there's more downtime, but I've left my thoughts on the other PR as well.

Another take could be to reduce the damage of subsequent shots from slugs and/or flech, in quick succession or in such a short period of time but I'm unsure how much the code would allow this. This could also potentially nerf/tackle the issue with akimbo combat shotties as they too are fired in quick succession and while it will nerf the base DB slightly, it'll nerf juggling more where using just one DB would be more beneficial. This would probably need a lot of testing but restricting the ability to carry 3 DBs would be a good start.

Pariah919 (changes requested):
I don't personally care for DB juggling but this just makes it more annoying to do than any actual nerf tbh.
Reduce the amount of DBs carried first.

Lumipharon:
> I don't personally care for DB juggling but this just makes it more annoying to do than any actual nerf tbh. Reduce the amount of DBs carried first.

Then people will just run it with a different weapon in that slot as the finisher.

Naaanii:
> > I don't personally care for DB juggling but this just makes it more annoying to do than any actual nerf tbh. Reduce the amount of DBs carried first.
> 
> Then people will just run it with a different weapon in that slot as the finisher.

That is how it was originally used and kind of meant to be used tbh.

Lumipharon:
> > > I don't personally care for DB juggling but this just makes it more annoying to do than any actual nerf tbh. Reduce the amount of DBs carried first.
> > 
> > 
> > Then people will just run it with a different weapon in that slot as the finisher.
> 
> That is how it was originally used and kind of meant to be used tbh.

I mean like, 3 db's and a t35, instead of 4 db's etc.

Naaanii:
> > > > I don't personally care for DB juggling but this just makes it more annoying to do than any actual nerf tbh. Reduce the amount of DBs carried first.
> > > 
> > > 
> > > Then people will just run it with a different weapon in that slot as the finisher.
> > 
> > 
> > That is how it was originally used and kind of meant to be used tbh.
> 
> I mean like, 3 db's and a t35, instead of 4 db's etc.

Should then make it so it's only possible to carry 2 DBs. 4 Weapons either way isn't really the norm and only because DB can be carried so easily in multiple slots.

Losenis:
Removing the magharn might actually be a good idea that leaves everything else untouched. An ultrasweaty guy might go triple DBs with belt harness but that will already absolutely kill the option of the shotgun shell belt/DB belt or lock out the "drop, equip, shoot" withot leaving all of it on the ground. Or getting knocked around.

BraveMole:
MJP proposed something else

PR #10670, opened 2022-08-02
DB juggling nerf -> Cannot fire DB unless stable wielding

Lauwma:
## About The Pull Request

DB cannot be shot unless the full wielding bar is completed. 

## Why It's Good For The Game

This prevents people from juggling 3 DBs in order to be able to shoot 6 slugs at a crazy firerate. Triple DBs are currently a concern and are in the rise since the T39 akimbo nerf. The issue is the same, a crazy firerate of slugs that staggers. 
This nerf keeps DB as a nice burst weapon that can quickly shoot out two shotgun ammo. The wield delay is 0.6 seconds, it's not that long but I believe it's enough to prevent juggling. I've been told this would impact single DBs greatly but I'm not sure of that since they rarely shoot at the same rate as jugglers that's why this PR should be testmerged for a few days if the maintainers are willing to try it.

**DO NOT MERGE THIS WITH #10668 THAT WOULD BE TOO MUCH OF A NERF**

## Changelog
:cl:
balance: Double Barrels now requires you to stable wield them to be able to shoot. 
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -212,7 +212,7 @@
 		/obj/item/attachable/scope/mini,
 	)
 
-	flags_gun_features = GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_WIELDED_FIRING_ONLY
+	flags_gun_features = GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_WIELDED_FIRING_ONLY|GUN_WIELDED_STABLE_FIRING_ONLY
 	attachable_offset = list("muzzle_x" = 33, "muzzle_y" = 17,"rail_x" = 15, "rail_y" = 19, "under_x" = 21, "under_y" = 13, "stock_x" = 13, "stock_y" = 16)
 
 	fire_delay = 5


Lumipharon:
This just nerfs db is general, not specifically juggling.

Lauwma:
I don't think it affects other ways of playing db much because the wield delay is only 0.6 sec (can be reduced to 0.4 with attachments)

Lumipharon:
It's a fact though, it's been talked about before and not implemented because the benefit of a shotgun is you can pull it out and immediately use it effectively. A warrior can combo you and leave before you could wield a db. And god help anyone that actually wants to reload a db in combat.

Lauwma:
Regardless of wield delay a warrior can lunge + punch you then leave without you having you having the opportunity to shoot your DB anyway so I think that's not really relevant. I understand it undermines the easy to pull out and shoot but I really do believe 0.6-0.4 seconds really aren't nothing, in fact when most people PB they usually do it in that frame of time. 
I really think it's only a theory nerf because in practice the stars don't align to make effective use of that "equip and shoot". Juggling solely exist to setup a situation where you can use it

Losenis:
Pretty harsh nerf for something that necessitates you unwield and wield again every 2 shots at minimum. It'd make the db deliberatedly worse than any other shotgun in that respect on top of already having a 2 round max capacity that needs a reload with more steps than normal.

I can only imagine this will lock it further into a juggling gun as it'd make it worse than it already is than the other shotguns for normal usage, since despite the full wielding with juggling you still skip the reloading and unless it takes 2.5 seconds to wield, it'll still have a higher firerate than t35 for damage purposes.

Naaanii:
This nerfs DB as a whole and not just juggling.

Al-1ce:
> This nerfs DB as a whole and not just juggling.

You can't nerf juggling without nerfing the DB itself. That's just the harsh reality of the situation.

I think a much tamer and simpler way to fix this without adjusting too much would be removing the ability to carry a DB on your backpack slot. You can't juggle if you only have one DB.

Naaanii:
> > This nerfs DB as a whole and not just juggling.
> 
> You can't nerf juggling without nerfing the DB itself. That's just the harsh reality of the situation.
> 
> I think a much tamer and simpler way to fix this without adjusting too much would be removing the ability to carry a DB on your backpack slot. You can't juggle if you only have one DB.

Yeah, a better way to nerf it would be to just nerf the number of DBs you can actually carry. 2 DBs should be fine where 3 is the real problem. DBs can't be akimbo'd but the ability to carry 2 is more fair imo as apposed to one.

Pariah919:
Doesnt seem to fix the problem to me considering DB is inaccurate with slugs until wield bar is finished, remove DB belt perhaps as people can run 4 DBs with it.

RikuTheKiller:
I've got a better idea. Make the magnetic harness attachment incompatible with DBs. This means that if you actually manage to lunge someone as a melee caste you'll have a good chance of getting away with at least some damage before they get ready to juggle again.

Saying this from the perspective of a xeno player, died to a marine that was using DB juggling 3 times in a single round while they got around 5 - 10 kills total. (Almost one third of the entire hive. By a single marine.)

Later on I even became the shrike and the entire hive ended mainly because of the DB juggler. They were the only one scoring any decent amount of kills by themselves.

The issue with such a potent close range weapon is that it forces you to play a ranged caste. The issue with this, however, is that weapons like the sniper or laser guns can completely deny ranged castes, so if both strategies are combined there's literally nothing you can do. Ranged? Get sniped. Close range? Get stunlocked and killed in less than a second.

Not to mention the time when I got killed almost instantly as an ancient defender with my crest down and full health by shotguns. Like, the most defensive caste in the entire game gets killed in like, a second.

Lauwma:
The mag harness doesn't really help, jugglers drop their db on the ground already. A solution that I read and could help is removing back slot capability but that still leaves with 2 DB and another shotgun like a T35. To this day I still believe my solution is not really a nerf, I think people are overreacting on the wield time of the DB which is the shortest of the game with T35

BraveMole:
MJP said no

PR #10679, opened 2022-08-03
Fix large object transparency opacity issue

XElectricX:
## About The Pull Request

Was wondering why my skinny street lamp was taking up 3 tiles worth of clicking space. Turns out when the transparency wears off, the mouse opacity becomes "opaque", which means any sprite will take up the entire tiles of space it occupies when you hover over the tiles with your cursor. Nothing in the game currently uses the large object transparency component except trees but this one line change will make trees and any future large objects work correctly.

## Why It's Good For The Game

Fixes large objects with transparency. If you had trouble with trees, not anymore.

## Changelog
:cl:
fix: Large object transparency no longer makes all tiles a multitile object occupies the only thing clickable.
/:cl:

### DIFF
--- a/code/datums/components/largeobjecttransparency.dm
+++ b/code/datums/components/largeobjecttransparency.dm
@@ -108,4 +108,4 @@
 	var/atom/par_atom = parent
 	par_atom.alpha = initial_alpha
 	if(toggle_click)
-		par_atom.mouse_opacity = MOUSE_OPACITY_OPAQUE
+		par_atom.mouse_opacity = MOUSE_OPACITY_ICON


Lumipharon:
Fucking trees man, good fix.

tyeagg:
great service to mappers we can abuse trees as hard as we can now

XElectricX:
> great service to mappers we can abuse trees as hard as we can now

Not just trees. Any multitile item that you can walk behind! Very good mechanic that should be used more often, just need to declare it on Initialize () and set the boundaries.

PR #10734, opened 2022-08-09
No hair no longer has TWO brackets on early synth.

Pariah919:
## About The Pull Request
A change so powerful its allowed to have TWO brackets.
No longer.

## Why It's Good For The Game
Code looks nicer.


Pariah919:
oh he direct commit reverted it one second

PR #10739, opened 2022-08-09
Revert "Fix bonus projectile not having IFF"

lKiseki:
Reverts tgstation/TerraGov-Marine-Corps#10702.

May be the culprit of ghost bullets, worth a shot.


lKiseki:
Not it.

PR #10740, opened 2022-08-09
Adjacency check for external magazines on turrets

XElectricX:
## About The Pull Request

How it works: when a deployable's reload() is called, or on Initialize(), it looks for if the magazine is external (WORN). Then it starts process() which checks if the magazine is adjacent to the deployable. If you walk off with the magazine, it disconnects from the turret.

## Why It's Good For The Game

Was told flamethrower turrets can use external magazines which lead to teleportation of flamer fuel to wherever the turret is if you walk away with the backpack. This fixes that and any future deployables with external magazines.

## Changelog
:cl:
fix: Worn magazines disconnect from turrets if you walk away with it.
/:cl:

### DIFF
--- a/code/modules/projectiles/mounted.dm
+++ b/code/modules/projectiles/mounted.dm
@@ -25,7 +25,9 @@
 		CRASH("[internal_item] was attempted to be deployed within the type /obj/machinery/deployable/mounted without being a gun]")
 
 	var/obj/item/weapon/gun/new_gun = internal_item
-
+	for(var/obj/item/ammo_magazine/ammo_magazine in new_gun.chamber_items)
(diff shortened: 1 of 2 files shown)


BraveMole (changes requested):
No that's just a bad solution. You shouldn't need to use process.

Why even allowing mag bag to feed sentries in the first place

XElectricX:
> No that's just a bad solution. You shouldn't need to use process.

What is wrong with process()? It's what IV stands use.

> Why even allowing mag bag to feed sentries in the first place

Emergent gameplay entering the chat.

Lumipharon:
This bug is not specific to turrets, it occurs with guns in hand as well.

PR #10672, opened 2022-08-02
End of round stat update

Lumipharon:
## About The Pull Request
Some minor rework to end of round stats.

Changes various stats to be lists organised and recorded by faction, and bullets fired now captures all projectiles (i.e. lasers etc).

Created a new set of stats specifically for Combat Patrol since it's HvH.

## Why It's Good For The Game
Slightly more accurate stats.
Useful stats for HvH.

## Changelog
:cl:
add: Added some HvH specific end of round stats for Combat Patrol, projectiles are counted instead of just bullets
code: tidied up end of round stats
/:cl:

PR #10742, opened 2022-08-09
Revert "End of round stat update"

lKiseki:
Reverts tgstation/TerraGov-Marine-Corps#10672. Possible culprit of ghost bullets.


lKiseki:
Confirmed; the reverted PR is the culprit of ghost bullets.

PR #10745, opened 2022-08-09
Changes to the gun64 T18 and Tp-14 sprite so it doesn't move

Xsdew:
## About The Pull Request

The AR-18 is slightly different when loading and unloading. This fixes the sprite to be correct in both states. As the time of making this PR the P-14 also has movement between loaded and unloaded states.

## Why It's Good For The Game

This is a bug fix for a recently implemented sprite.

## Changelog
:cl:

fix: The gun64 sprite of the AR18 & P-14 shouldn't shrink anymore.

/:cl:


Xsdew:
Going to update the sprite momentarily for the P-14.

Al-1ce:
This will conflict with #10746

Xsdew:
Kuro is going to do more modifications to the sprite. I will open another PR with all the changes in the future.

Al-1ce:
Fair enough.

PR #10748, opened 2022-08-10
Support for underlayers on worn item rendering

SplinterGP:
## About The Pull Request
Adds support for rendering item underlayers on human mobs, this way you dont need to cut parts of the sprite so it doenst clip over armor or other part, you can just put it actually behind, especially directions, example, when looking south you want the suit storage weapons or some backpack to render BEHIND the character instead of rendering above and having to make cuts on the sprite, so you just go and put that state that would be cut on the overlay south state, and just put it on an underlay south state.

left some comments on human update_icon() because poofy is gonna need it for suit store and back(since i dindt put it to actually look for underlays on any slot since none use it yet. 

Also removed icon override(useless, use the other ways to change icon) and changed some stuff that was using images to use MA's, 

## Why It's Good For The Game
Spriters wont need to make stuff cut for different armor shapes or anything when it is intended to render behind everything in a certain direction(like south) anymore, making them more free to sprite it how they want.

## Changelog
:cl:
code: added support for items to use underlayers so its easier for them to be made to look behind the character without needing to be cut or clip over other items worn.
/:cl:


SplinterGP:
changing it to multiple layers support

PR #10756, opened 2022-08-12
Motorbikes will not act as free armor

XElectricX:
## About The Pull Request

Try to shoot someone on a bike. Only some shots will hit them and the bike will absorb most of it. From testing it seems you can only hit them if you hover over their sprite and even then not all shots connect. The tiny change makes it so that if the vehicle has riders, projectile will ignore the bike when it calculates a hit.

Throwpass has been enabled and coverage has been adjusted as well so that if there is no one on a bike, you can still shoot over it instead of acting as a wall.

## Why It's Good For The Game

Motorbikes gave not only speed but a free health barrier. Pretty weird to let it do this.

## Changelog
:cl:
balance: Motorbikes no longer act as barriers and can be shot over.
fix: Hitting riders on motorbikes is now consistent.
/:cl:


Lumipharon:
Pretty sure a corpse strapped to a bikewould be an unbreakable cade (from shooting damage)

XElectricX:
> Pretty sure a corpse strapped to a bikewould be an unbreakable cade (from shooting damage)

In my many hours of testing I have found you can't strap corpses to bikes so not actually a problem! And even if you could, they would eventually fall over if you damage them to the point they lose their arms.

Naaanii:
Based PR if it works right and doesn't cause any other issues.

code/modules/vehicles/motorbike.dm
```diff
@@ -195,8 +197,7 @@
 /obj/vehicle/ridden/motorbike/projectile_hit(obj/projectile/P)
```
BraveMole, line 197: could you please move that proc to projectile.dm ? 

It's where projectile_hit are supposed to be.

code/modules/vehicles/motorbike.dm
```diff
@@ -195,8 +197,7 @@
...
 	if(!buckled_mobs)
 		return ..()
-	var/mob/buckled_mob = pick(buckled_mobs)
-	return buckled_mob.projectile_hit(P)
+	return FALSE
```
BraveMole, line 200:
```suggestion
	return src == proj.original_target
```
How about that? Just a suggestion though

BraveMole (changes requested):

code/modules/vehicles/motorbike.dm
TiviPlus: I like it

code/modules/vehicles/motorbike.dm
XElectricX: Moved.

code/modules/vehicles/motorbike.dm
XElectricX: Set whoever is buckled as the src.

PR #10758, opened 2022-08-13
Spears Have Normal Reach While Unwielded

RikuTheKiller:
## About The Pull Request

Spears can no longer reach 2 tiles while unwielded. It's as simple as that.

I was originally going to raise the unwielded damage a little bit, but after looking at how much damage knives deal I decided to revert the change. (Using them as a reference for how a spear held at the middle with one hand would function as a weapon.)

## Why It's Good For The Game

Right now you can use shields and a spear together with only lesser damage as a downside (And still more damage than a knife.), this fixes that and gives swords an actual purpose. (Also affects the harvester spear.) Two handed weapons shouldn't be more effective while held in one hand compared to one handed weapons. Also, have you ever tried using a spear with one hand by holding it at the end? You really don't have that much force. And your accuracy is so low that hitting a mobile target is almost impossible anyways.

## Changelog
:cl:
balance: spears now have shorter reach while unwielded
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -237,7 +237,7 @@
 	desc = "A haphazardly-constructed yet still deadly weapon of ancient design."
 	icon_state = "spearglass"
 	item_state = "spearglass"
-	force = 40
+	force = 50
 	w_class = WEIGHT_CLASS_BULKY
 	flags_equip_slot = ITEM_SLOT_BACK
 	force_wielded = 75
@@ -264,7 +264,7 @@
 	desc = "TerraGov Marine Corps' experimental High Point-Singularity 'Harvester' spear. An advanced weapon that trades sheer force for the ability to apply a variety of debilitating effects when loaded with certain reagents. Activate after loading to prime a single use of an effect. It also harvests substances from alien lifeforms it strikes when connected to the Vali system."
 	icon_state = "vali_spear"
 	item_state = "vali_spear"
-	force = 32
+	force = 40
 	force_wielded = 60
 	throwforce = 60
 	flags_item = DRAINS_XENO | TWOHANDED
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/twohanded.dm
```diff
@@ -260,7 +264,7 @@
 	desc = "TerraGov Marine Corps' experimental High Point-Singularity 'Harvester' spear. An advanced weapon that trades sheer force for the ability to apply a variety of debilitating effects when loaded with certain reagents. Activate after loading to prime a single use of an effect. It also harvests substances from alien lifeforms it strikes when connected to the Vali system."
 	icon_state = "vali_spear"
 	item_state = "vali_spear"
-	force = 32
+	force = 30
```
QualityVan, line 267: Did you mean to reduce this from live?

FosterGrime-Heymynameslouie:
image

RIP larping as a 18th century spearsman fighintg against spooky aliens

code/game/objects/items/weapons/twohanded.dm
RikuTheKiller: Oh, woops, gonna fix that.

RikuTheKiller:
> image
> 
> RIP larping as a 18th century spearsman fighintg against spooky aliens

Honestly? They're holding it from the middle. It's not that effective anymore at that point. A little bit longer range? Yeah, sure, but honestly that's not much longer than a sword.

TiviPlus (changes requested):
Cool argument but it already does almost half damage this is a salt pr isnt it

RikuTheKiller:
> Cool argument but it already does almost half damage this is a salt pr isnt it

I don't even play against spear users? I sometimes use spears myself.

Anyways, it's because half damage isn't even close to enough of a debuff for essentially having a ranged melee weapon. (Reminder that in the case of the harvester you can't even have it dropped out of your hand.)

Naaanii:
Looks fine. If there's wield slowdown, would be better to remove it otherwise it's good to make wielding actually have benefits vs one handing.

Losenis:
>it's because half damage isn't even close to enough of a debuff for essentially having a ranged melee weapon

You say that until some dunga fracs your groin with a misclicked spear to the rouny on top of you because it's pixelhunting without aim mode

bowlofcereal:
Using spear one handed massively lowers its damage to pointlessness if you see anyone using shield and spear its because they are new and don't understand there is a damage debuff. 

Unnecessary PR considering the above statement I have made and what others have already stated.

Al-1ce:
Xeno main here. Spear already has enough drawbacks on it's own. outside of being low damage on it's own. This just isn't needed.

RikuTheKiller:
After considering all the points made by everyone in the comments, I'll be closing this PR.

Issue #6055, opened 2021-02-16
Removing Storage Module from Armor destroys anything inside of it


SkeletalElite:
## Reproduction:

Remove storage module from your exoskeleton (general storage in case it doesn't apply to the other types, but it probably does).
The stuff inside of it is nowhere to be found. Im guessing it gets deleted when the module is removed because the module item itself isn't a storage object.

PR #10759, opened 2022-08-13
Fixes storage module destroying all their contents

JohnFulpWillard:
## About The Pull Request

Destroying or unequipping the storage module will now cause it to drop all its contents to the floor.

## Why It's Good For The Game

Closes https://github.com/tgstation/TerraGov-Marine-Corps/issues/6055

## Changelog
:cl:
fix: Storage modules will now drop their contents when unequipped/destroyed.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/storage.dm
+++ b/code/modules/clothing/modular_armor/attachments/storage.dm
@@ -38,8 +38,6 @@
 	storage.master_item = parent
 
 /obj/item/armor_module/storage/on_detach(obj/item/detaching_from, mob/user)
-	for(var/obj/item/all_items AS in storage.contents)
-		storage.remove_from_storage(all_items, get_turf(src))
 	time_to_equip = initial(time_to_equip)
 	time_to_unequip = initial(time_to_unequip)
 	UnregisterSignal(parent, list(COMSIG_ATOM_ATTACK_HAND, COMSIG_PARENT_ATTACKBY))
--- a/code/modules/clothing/modular_armor/attachments/storage.dm
+++ b/code/modules/clothing/modular_armor/attachments/storage.dm
@@ -23,8 +23,11 @@
 	storage = new storage(src)
 
 /obj/item/armor_module/storage/Destroy()
-	. = ..()
-	QDEL_NULL(storage)
+	if(storage)
+		for(var/obj/item/all_items as anything in storage.contents)
+			storage.remove_from_storage(all_items, get_turf(src))
+		QDEL_NULL(storage)
+	return ..()
 
 /obj/item/armor_module/storage/on_attach(obj/item/attaching_to, mob/user)
 	. = ..()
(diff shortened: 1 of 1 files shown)


code/modules/clothing/modular_armor/attachments/storage.dm
```diff
@@ -23,8 +23,11 @@
 	storage = new storage(src)
 
 /obj/item/armor_module/storage/Destroy()
-	. = ..()
-	QDEL_NULL(storage)
+	if(storage)
+		for(var/obj/item/all_items as anything in storage.contents)
```
BraveMole, line 27:
```suggestion
		for(var/obj/item/all_items AS in storage.contents)
```
We have an alias for as anything in TGMC

code/modules/clothing/modular_armor/attachments/storage.dm
```diff
@@ -35,6 +38,8 @@
 	storage.master_item = parent
 
 /obj/item/armor_module/storage/on_detach(obj/item/detaching_from, mob/user)
+	for(var/obj/item/all_items as anything in storage.contents)
```
BraveMole, line 41: ditto

BraveMole (approved):

Lumipharon:
What is this fixing? Module storage is not deleted when you remove it.
image

BraveMole:
> What is this fixing? Module storage is not deleted when you remove it.

Yeah it just empties it ?

Losenis:
Is it possible that this is only on jaeger exoskeleton? I remember very well this actually working when I need to move contents from a loadout with a xenonaut of one weight to a xenonaut of another weight. I was even able to interact with the module outside the armor to put or retrieve items from it.

Losenis:
Tried to replicate bug report with both xenonaut and exoskeleton, it works as intended. Something else might've caused the bug to happen, this fix isn't necessary and might not actually fix what could've really happened.

image

JohnFulpWillard:
> Tried to replicate bug report with both xenonaut and exoskeleton, it works as intended. Something else might've caused the bug to happen, this fix isn't necessary and might not actually fix what could've really happened.

If the contents carry over to the physical item (which seems odd, but im more used to working with modsuits) then I'll keep it on the Destroy in case the suit is somehow qdel'ed

Lumipharon:
> > Tried to replicate bug report with both xenonaut and exoskeleton, it works as intended. Something else might've caused the bug to happen, this fix isn't necessary and might not actually fix what could've really happened.
> 
> If the contents carry over to the physical item (which seems odd, but im more used to working with modsuits) then I'll keep it on the Destroy in case the suit is somehow qdel'ed

Almost all storage items in the codebase have storage in the same way, so the bugreport is incorrect (it's over a year old to be fair), the storage is working entirely as intended.

I wouldn't recommend even leaving this for qdel, because if you're quelling the storage container you typically want to delete the contents also (I.e. gibbing).

BraveMole:
I remember that happening though, and don't remember having seen a fix though. 

I guess that's good

BraveMole (changes requested):
image

Storage do not hard dels though

breadhunt:
> I remember that happening though, and don't remember having seen a fix though.
> 

This was an issue, but hyper fixed it with his attachie reworks (since storage modules are now technically armour attachments)
Bless for going to the effort to fix the issue anyway, though

SkeletalElite:
Back when I opened the issue report it was definitely the case, I remember testing it. Removing the storage module would 100% delete anything inside of it.

It's possible the issue was fixed but the issue was never closed though, I haven't played in ages.

PR #10763, opened 2022-08-14
Other surgery fixes

QualityVan:
## About The Pull Request
The do_surgery improvements were a little incomplete. This follows up by making all surgery steps compliant with the new defines plus clearing up some logic.
Also removed the surgeries for organ implant/removal, since those were already commented out of use, and removes a couple dead vars.

## Why It's Good For The Game
bugfix

## Changelog
:cl:
fix: Surgeries work
/:cl:


BraveMole (approved):
Amazing surgery cleanup, you get a cookie

Weedmaster77:
Not sure if this is the cause of it but today after this was merged the Radial Medical Wheel when doing surgery seem to be some what broken. Was trying to do surgery with it and selecting targeted limbs on almost all steps seem to make it want you to reselect the part again and again and again and takes like four or five time I had to select that limb to get it to start the step only steps that seem to not have this problem where cutting necro flesh and applying membrane. But incision, clamping and retracting all had the same problem.

PR #10767, opened 2022-08-15
Fixes detpacks

breadhunt:
## About The Pull Request

Detpacks now check for the Z level of whoever's hand they're in, instead of only their own Z level. This broke detpacks because, when an object is on a person or inside of a storage unit, the object's z level is always set to 0. This caused issues with https://github.com/tgstation/TerraGov-Marine-Corps/pull/10703

## Why It's Good For The Game

Fixes good, I like using my funny detpacks :)

## Changelog
:cl:
fix: Detpacks now work when a signaler is in your hand again, instead of only working on the floor
/:cl:

### DIFF
--- a/code/game/objects/items/radio/detpack.dm
+++ b/code/game/objects/items/radio/detpack.dm
@@ -132,7 +132,8 @@
 	if(!signal || !on)
 		return
 
-	if(!(signal.source.loc.z == z || signal.source.z == z))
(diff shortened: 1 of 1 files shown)


TiviPlus:
Wouldn't get turf be better here

TiviPlus (approved):

PR #10773, opened 2022-08-16
Add mortar & howitzer shots to end game stats

lbnesquik:
## About The Pull Request
Mortar & Howitzers go boom. It would be nice to know if there has been like, 5 shots or 50 during a particular round.

## Why It's Good For The Game
Stats good. Not a huge performance impact given we track bullets fired.

## Changelog
:cl:
add: Added mortar & howitzers shots to end round.
/:cl:

### DIFF
--- a/code/game/objects/machinery/mortar.dm
+++ b/code/game/objects/machinery/mortar.dm
@@ -1,6 +1,9 @@
 //The Marine mortar, the T-50S Mortar
 //Works like a contemporary crew weapon mortar
 
+#define TALLY_MORTAR  1
+#define TALLY_HOWITZER 2
+
 /obj/machinery/deployable/mortar
 
 	anchored = TRUE
@@ -39,8 +42,10 @@
 		/obj/item/mortal_shell/plasmaloss,
 	)
 
-
 	use_power = NO_POWER_USE
+
+	///Used for round stats
+	var/tally_type = TALLY_MORTAR 
 
 /obj/machinery/deployable/mortar/attack_hand(mob/living/user)
 	. = ..()
@@ -210,6 +215,13 @@
 		flick(icon_state + "_fire", src)
 		mortar_shell.forceMove(src)
 
+		if(tally_type == TALLY_MORTAR)
+			GLOB.round_statistics.howitzer_shells_fired++
+			SSblackbox.record_feedback("tally", "round_statistics", 1, "howitzer_shells_fired")		
+		else if(tally_type == TALLY_HOWITZER)
+			GLOB.round_statistics.mortar_shells_fired++
+			SSblackbox.record_feedback("tally", "round_statistics", 1, "mortar_shells_fired")		
+
 		var/turf/G = get_turf(src)
 		G.ceiling_debris_check(2)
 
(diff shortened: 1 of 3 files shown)


code/game/objects/machinery/mortar.dm
```diff
@@ -210,6 +210,14 @@
 		flick(icon_state + "_fire", src)
 		mortar_shell.forceMove(src)
 
+		if(istype(I, /obj/item/mortal_shell/howitzer))
+			GLOB.round_statistics.howitzer_shells_fired++
+			SSblackbox.record_feedback("tally", "round_statistics", 1, "howitzer_shells_fired")		
+
+		else if(istype(I, /obj/item/mortal_shell))
```
QualityVan, line 217: This block already knows that I is a shell from 157, you can just have this line as an else.

code/game/objects/machinery/mortar.dm
```diff
@@ -210,6 +210,13 @@
 		flick(icon_state + "_fire", src)
 		mortar_shell.forceMove(src)
 
+		if(istype(src, /obj/machinery/deployable/mortar/howitzer))
```
TiviPlus, line 213: Istype src is bad

TiviPlus (changes requested):

code/game/objects/machinery/mortar.dm
BraveMole: Two solutions : 
Either you override the proc for howitzer

Else, you add a var on the projectile or mortar to specify on which tally it should be added

code/game/objects/machinery/mortar.dm
```diff
@@ -39,6 +39,8 @@
 		/obj/item/mortal_shell/plasmaloss,
 	)
 
+	var/tally_type = 1	//Used for round stats. 1 is mortar, 2 is howitzer
```
BraveMole, line 42: you need autodoc, and you need to use define
///Used for round stats
var/tally_type = MORTAR_ROUND_TALLY

code/game/objects/machinery/mortar.dm
```diff
@@ -210,6 +212,13 @@
 		flick(icon_state + "_fire", src)
 		mortar_shell.forceMove(src)
 
+		if(tally_type == 1)
+			GLOB.round_statistics.howitzer_shells_fired++
+			SSblackbox.record_feedback("tally", "round_statistics", 1, "howitzer_shells_fired")		
+		if(tally_type == 2)
```
BraveMole, line 218: should be elseif or even else

BraveMole (changes requested):

code/game/objects/machinery/mortar.dm
```diff
@@ -1,6 +1,9 @@
 //The Marine mortar, the T-50S Mortar
 //Works like a contemporary crew weapon mortar
 
+#define MORTAR 1
```
BraveMole, line 4: In general, you want your vars /defines to be as verbose as possible.

Here, TALLY_MORTAR would be a better name

code/game/objects/machinery/mortar.dm
```diff
@@ -39,6 +42,9 @@
 		/obj/item/mortal_shell/plasmaloss,
 	)
 
+	///Used for round stats
+	var/tally_type = MORTAR
+
 
 	use_power = NO_POWER_USE
```
BraveMole, line 49:
```suggestion
	use_power = NO_POWER_USE
	///Used for round stats
	var/tally_type = MORTAR
```

New vars after existing ones

BraveMole (changes requested):
You also need to undefine the defines at the end of the file, since they are only used there

BraveMole (approved):

PR #7021, opened 2021-06-03
removes hard melee from shields and they make you slower

Pariah919:
## About The Pull Request
you no longer have hard melee armor with a shield and they make you a decent bit slower

## Why It's Good For The Game
shields are innanely good in modern tgmc and -5 damage off every slash in the game and then -50% is way too much considering losing one hand is no longer as much of a downside as it used to be

## Changelog
:cl:
balance: shields no longer have hard melee armor and slow you down a bit more
/:cl:

PR #10775, opened 2022-08-16
Rework how shields reduce incoming damage

XElectricX:
## About The Pull Request

I was wondering why shields were not working when giving them high armor values. Discovered that if hard armor was higher than the damage being dealt, it just did an early return.

This PR gives shields a new formula for damage reduction, changes the integrity damage on the shield to be the amount that was absorbed overall minus the soft armor on itself, and adds new messages for when most or all of the damage is negated.

## Why It's Good For The Game

The TGMC shields have very low hard armor so I do not think this problem was ever faced yet, but in the future if any sturdier shields are added they won't bug out from trying to reduce damage to 0.

## Changelog
:cl:
fix: Hard armor on shields were not reducing damage at all if it was greater. Not anymore.
refactor: Rework how shields reduce incoming damage.
/:cl:

### DIFF
--- a/code/datums/components/shield.dm
+++ b/code/datums/components/shield.dm
@@ -207,14 +207,19 @@
 			return prob(50 - round(incoming_damage / 3))
 		if(COMBAT_MELEE_ATTACK, COMBAT_PROJ_ATTACK)
 			var/absorbing_damage = incoming_damage * cover.getRating(damage_type) * 0.01 * status_cover_modifier  //Determine cover ratio; this is the % of damage we actually intercept.
-			absorbing_damage = max(0, absorbing_damage - hard_armor.getRating(damage_type)) //We apply hard armor *first* _not_ *after* soft armor
-			if(!absorbing_damage)
+			if(!absorbing_damage)	//This should be checked before we get to damage calculations!
 				return incoming_damage //We are transparent to this kind of damage.
-			. = incoming_damage - absorbing_damage
+			var/calculated_damage = max(0, absorbing_damage - hard_armor.getRating(damage_type)) //We apply hard armor *first* _not_ *after* soft armor
+			if(incoming_damage != absorbing_damage)	//If only part of the damage is being reduced, use the formula below
+				. = abs((calculated_damage ? calculated_damage : absorbing_damage) - incoming_damage)	//If calculated_damage is 0, then use what got absorbed for the math
+			else
+				. = calculated_damage
 			absorbing_damage *= (100 - soft_armor.getRating(damage_type)) * 0.01 //Now apply soft armor
-			if(absorbing_damage <= 0)
-				if(!silent)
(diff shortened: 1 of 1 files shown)


Dzhimax:
автор долбоёб, it`s bad for the game. пиздуй на подстанцию, электрик

helper-2:
> автор долбоёб, it`s bad for the game. пиздуй на подстанцию, электрик

Джимакс, ты чё, вообще забыл на каком языке тут говорят?

Naaanii:
I believe hard armour was removed for a reason.
https://github.com/tgstation/TerraGov-Marine-Corps/pull/7021

Don't really want to go back to God mode shields and having to slash someone 50+ times just to kill.

Lumipharon:
What are you doing with the code here exactly? The code was wack before but I don't think this exactly resolves it.
Absorbing_damage is what's actually applied to the shield, the damage that's reduced against the mob is separate to the shield's actual armour values, but yeah hard armour was wack.

BraveMole:
@helper-2 @Dzhimax only english there please

XElectricX:
> I believe hard armour was removed for a reason.
> https://github.com/tgstation/TerraGov-Marine-Corps/pull/7021
> 
> Don't really want to go back to God mode shields and having to slash someone 50+ times just to kill.


This does not add hard armor to the shield, please reread the PR. Also if melee was 0 that means shields were not actually doing anything against melee this whole time.

XElectricX:
> What are you doing with the code here exactly? The code was wack before but I don't think this exactly resolves it.
> Absorbing_damage is what's actually applied to the shield, the damage that's reduced against the mob is separate to the shield's actual armour values, but yeah hard armour was wack.


It says it in the PR. It just fixes the calculation for when the hard armor value is higher than the damage. If a bullet did 5 damage but the hard armor was 10, the bullet still did 5 damage. Now it would do 0.

I know what absorbing_damage does, it has no effect on the damage reduced beyond telling the shield the portion of damage it can block, then it calculates hard armor. The change I made to it is that instead of the total damage reduced being what the shield takes in terms of integrity damage, it will now be what was absorbed overall. For example, a 10 damage bullet with the shield having 5 hard armor. 5 damage is the damage taken by the user, but 10 damage is what the shield takes because it was trying to block all 10 points of damage. Now let's say the shield has a different cover value and it only calculates 50% of all damage taken. The absorbing_damage would equal 5, and because that is what the shield was attempting to block, it takes the 5 integrity damage.

Pariah919:
Shields work and are balanced off a two step system of cover that redirects the damage to the shield rather than the mob. And then that damage to the shield is reduced by hard armor (which shields have hard bullet armor to reduce low damage rounds for) then soft armor reduces that leftover damage, whatever is left over reduces the shields HP.

XElectricX:
Misinterpreted how shields were meant to function as.

PR #10776, opened 2022-08-16
Shield hard armor works, bullet pen applies

QualityVan:
## About The Pull Request
Hard armor on shields had two issues: if it was more than the damage taken the shield didn't absorb anything at all, and if it wasn't then it still increased the damage the mob took.
Bullet pen didn't apply to shields at all. With this it'll increase the damage the shield itself takes (mob damage is unaffected) for absorption shields like riot and reduce block chance for binary shields like eswords. 

## Why It's Good For The Game
Bugfixes

## Changelog
:cl:
add: High-pen bullets deal more damage to shields
fix: Shield hard armor reduces damage instead of increasing it
/:cl:

### DIFF
--- a/code/datums/components/shield.dm
+++ b/code/datums/components/shield.dm
@@ -180,7 +180,7 @@
 		return
 	affecting_shields[intercept_damage_cb] = layer
 
-/datum/component/shield/proc/item_intercept_attack(attack_type, incoming_damage, damage_type, silent)
+/datum/component/shield/proc/item_intercept_attack(attack_type, incoming_damage, damage_type, silent, penetration)
 	var/obj/item/parent_item = parent
 	var/status_cover_modifier = 1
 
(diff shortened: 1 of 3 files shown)


BraveMole (changes requested):
Fix good, but question: do we still want hard armor to be in the code anyway ?

QualityVan:
Shields actually have some bullet hard armor, just to tone down FF further (it makes it worse right now because bugs but y'know). Beyond that though I think it does have uses in terms of a way to intuitively restrict weaker guns/castes from damaging specific emplacements, like how miner reinforcement locks destruction to only large xenos.
Also it'd be a pain to remove.

BraveMole (approved):

Serxule:
yeah shields block bullets from FF, but only a few

PR #10779, opened 2022-08-16
Correct bit of wrong text with supply drop console

lbnesquik:
## About The Pull Request
Text was wrong. Said 1 minute, but it is 30s in game.

## Why It's Good For The Game
Fixes are great.

## Changelog
:cl:
fix: fixed a wrong line about time regarding drops.
/:cl:


TiviPlus (approved):
Should use a define tbh

BraveMole (changes requested):
Yeah please use var/define

PR #10780, opened 2022-08-16
Double Shot Martini

Cowbin:
## About The Pull Request
Makes the Martini able to hold two rounds at a time.

## Why It's Good For The Game
The Martini is currently a particularly bad weapon, its slow manual fire rate makes it difficult to actually secure xeno kills with it, as well as its lack of aim mode forcing you to be in the front line unlike other snipers.  You would almost always be better off using a DB with slugs rather than a martini.  Plus, buffing this will also be a nerf to marines as FF would go up.  Just a test PR btw.

## Changelog
:cl:
balance: Makes the martini able to hold two rounds.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -371,15 +371,16 @@
 
 /obj/item/weapon/gun/shotgun/double/martini
 	name = "\improper Martini Henry lever action rifle"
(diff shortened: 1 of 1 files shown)


Pariah919 (changes requested):
Would be funny as an event meme but not balanced at all.

Naaanii:
Funny meme. You're better off making a separate gun.

PR #10797, opened 2022-08-20
remove r windows in prep to free roomba

novaepee:
## About The Pull Request

title

## Why It's Good For The Game

Let roomba do its job. The r windows in prep block roomba since its AI is not that good.

## Changelog
:cl:
del: remove reinforced windows in prep to free roomba
/:cl:


RipGrayson (approved):
I would prefer a solution that allows roombas to pass under glass, but this will work.

Issue #10791, opened 2022-08-19
Piercing projectiles can hit multiple times (AT Gun oneshots queens)


collinlunn:
### Description

1) If a mob is running in the same direction as a AMMO_PASS_THROUGH_MOVABLE projectile (Railgun, TX8 impact magazine, AT APCR, a few others) that hits them, they can be hit twice. Looks like they are hit once, they move in front of the projectile, then the projectile updates and hits them again. I'm not sure if this happens frequently in practice, but some guns like Railgun can one-hit crit a xeno if they manage to double-hit.
2) Relatedly, the AT gun's APCR ammo has knockback that pushes the hit target further into the projectiles path, so you don't need good timing for it to multi-hit. In fact, you can easily get it to triple hit, for ~600 damage with 40 armor pen, which instantly kills (not crits, kills) most xenos (Including Ancient Queen, Ancient Ravager, Elder Crusher). Since this is a roundstart weapon, I think this is a big issue.

### Test Merges

Tested on master 8/19 (d5b49d2e49bcaae9840024416953181426e166a8)

### Reproduction Steps

For AT gun:
1) Spawn in AT gun (/obj/machinery/deployable/mounted/moveable/atgun)
2) Spawn in APCR ammo (/obj/item/ammo_magazine/standard_atgun/apcr)
3) Spawn a Primordial Queen directly up from the gun
4) Fire the gun well past the xeno
5) See in the logs how it was hit 3 times, instantly killing it

For other guns:
1) Put a xeno on a conveyor belt
2) Fire the piercing gun around the time it the xeno away from you (TX8 impact magazine is easiest since it has a lot of ammo)
3) Note that sometimes chat says you hit the xeno twice (Can't consistently time the shot to make it do it, but I definitely got it to happen multiple times)

Lumipharon:
Pretty sure caused by #10629 
@BraveMole

PR #10799, opened 2022-08-21
Partial fix to APCR ammo instakilling

collinlunn:
## About The Pull Request
Removes the knockback from the AT gun's APCR ammo.

## Why It's Good For The Game
Currently piecing projectiles can hit a target multiple times under certain cases (#10791). For the AT gun's APCR ammo, the knockback pushes the target away, so the mob-piercing projectile can then move forward and hit them again. This consistently triple-hits xenos for ~600 damage, which instantly kills nearly all xenos from full health. Removing knockback prevents this from happening.

I don't know if a general fix for projectiles hitting multiple times is wanted/doable, so I'm making this PR as a bandaid fix, since this is regularly affecting rounds from what I've watched.

## Changelog
:cl:
balance: The AT gun's APCR ammo no longer has knockback.
fix: The AT gun's APCR ammo shouldn't hit its target multiple times anymore.
/:cl:


BraveMole:
That's a bandaid fix and we don't do that here. Especially since this is perfectly fixeable

PR #10818, opened 2022-08-23
Printable Tadpole Guns

novaepee:
## About The Pull Request

To enable tadpole guns into existence, the TerraGov Marine Corps' Logistic Division decided that allowing POs to print out tadpole guns is a go.

## Why It's Good For The Game

Marines never see the existence of different tadpole guns, but this PR is more complex than it seems.

Tadpole and Condor will have to share fabricator points. This means that if tadpole wants to use other weapons beside the machine gun, condor will have to wait. This makes tadpole have to help with miners so that they can use req points to purchase fabricator points, and they are quite expensive.

image

Also, you have to buy ammo from req.

## Changelog
:cl:
add: Enable PO to print out more guns for tadpole
balance: combat tadpole is getting reinforced a lot
/:cl:

### DIFF
--- a/code/game/objects/structures/dropship_equipment.dm
+++ b/code/game/objects/structures/dropship_equipment.dm
@@ -483,7 +483,7 @@
 	desc = "A box that deploys a modified M56D crewserved machine gun. Fits on the crewserved weapon attach points of dropships. You need a powerloader to lift it."
 	equip_category = DROPSHIP_CREW_WEAPON
 	icon_state = "mg_system"
-	point_cost = 300
+	point_cost = 190
 	///machine type for the internal gun and for checking if the gun is deployed
 	var/obj/machinery/deployable/mounted/deployed_mg
 
@@ -524,7 +524,7 @@
 	desc = "A box that deploys a modified MG-2005 crewserved minigun. Fits on the crewserved weapon attach points of dropships. You need a powerloader to lift it."
 	equip_category = DROPSHIP_CREW_WEAPON
 	icon_state = "minigun_system"
-	point_cost = 0 //this removes it from the fabricator
+	point_cost = 300
 	///machine type for the internal gun and for checking if the gun is deployed
 	var/obj/machinery/deployable/mounted/deployed_minigun
 
@@ -565,7 +565,7 @@
 	desc = "A box that deploys a modified ATR-22 crewserved dualcannon. Fits on the crewserved weapon attach points of dropships. You need a powerloader to lift it."
 	equip_category = DROPSHIP_CREW_WEAPON
 	icon_state = "ac_system"
-	point_cost = 0 //this removes it from the fabricator
+	point_cost = 225
 	///machine type for the internal gun and for checking if the gun is deployed
 	var/obj/machinery/deployable/mounted/deployed_dualcannon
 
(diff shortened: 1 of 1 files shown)


novaepee:
Not going to lie, this will make tadpole more oppressive, but need to see people's thoughts on this, see how much they want this in, and what needs to be changed if we let PO do this.

FosterGrime-Heymynameslouie:
the cabal of po mains (like 3 people and a potted plant) approves of this

also in terms of how oppresive it can be, how expensive it should be should ought to mitigate it, although only testing would give the right information on trying to find the sweet spot.
really like them being potentially seen outside of req basement hell.

Lumipharon:
The fact that this shares CAS points offsets a lot of the potential power of this, since it becomes impractical to be running effective CAS while also splashing out on nice tad upgrades.

Weedmaster77:
One question I have is will all the mounts that are currently onboard like the 4 sentry and 2 flare launchers are they still going to be there round start for the tad and alamo to use. Because if you take away the sentries that are spawn round start you'll never see the tad used again in ,my opinion but if the 500 points for a additional sentry mount on top of those that you spawn normally with everything seems fine. Plus making the other mountable guns printable seems like a nice idea making them easier to get earlier since tad seems to be used more then CAS and also seems to be used more effectively more often.

Losenis:
Finally, tadpole and condor POs killing each other while I nab the tadpole myself

novaepee:
> are they still going to be there round start for the tad and alamo to use

Yes.

Al-1ce:
The fabricator point production rate should be increased to balance things beteeen condor/cas. Otherwise this might actually become a nerf.

Starfirejordan:
I suggest leaving tadpole weapons systems as buyable because point between the two will become restrictive too much
unless you buff point genes or straight up make "tadpole printer" to balance it so the PO dont fight over point and most likely FF each other because "oh no i cant buy my missiles because the stupid other PO bought a minigun emplacement for there stupid Tad they have done nothing with!!"

Lumipharon:
IC issues are completely fine, it's literally just adding more options to the vendor. And points can still be bought with req points.

MereWalker:
I suggest making a tadpole fabrication bench. So condor and tadpole don't inadvertently kill each other over what gear the points are used on. Also can we get tadpole version 4 make it bigger and give it more guns. With love.

Losenis:
Bigger tadpole would be a devastating nerf given the spaces it can JUST fit into right now and how fewer walls it'd need to block it from landing anywhere

MereWalker:
Revert it to the chad cave pole. I loved that design. It was so good too. Maybe somewhat unfair but letting it go into caves a little bit would be nice.

Losenis:
Wasn't it forbidden from cavepoling over the front being indestructible and blocking caves with it? Now that the front is very much destructible, cavepole might be interesting

MereWalker:
> 

You bout to make me act up. Please don't make me excited about cavepole unless you for real for real.

TiviPlus (changes requested):
Why add 2 aquisition methods one will always be superior
Why force tad and cas pos to fight over the points

novaepee:
> Why add 2 aquisition methods one will always be superior 

As of right now, no one buys tadpole guns from req. It's barely in use.

> Why force tad and cas pos to fight over the points

They can get more points by having tadpole acquire miners.

TiviPlus:
That doesnt answer either of my questions

TiviPlus:
Misclick

Naaanii:
Dropship fabricator pass buff or changes would make this more interesting. Should probably delete it from req though if it's going to be in the fabricator.

Al-1ce:
The tadpole covering miners isn't something that'll happen every round. If anything, it's pretty infrequent. I think relying on that would hardly be a band-aid to this problem. 

Increasing the overall point generation would cover both the Condor and the Tadpole. IC issues that spawn from this are a direct consequence from not doing exactly that. By allowing the tadpole to spend these points, it's adding an additional drain that wasn't ever counted for in the past.

novaepee:
I will let tye figure this out.

PR #10823, opened 2022-08-24
Leg breaks adjustment

Lumipharon:
## About The Pull Request
Currently, having a broken leg does precisely fuck all.
You cannot fall over unless you have 2 broken limbs (including legs and feet) or more, while having one shatter leg is totally fine.

Changes this so having a single broken leg/foot can cause the stun. The calculation for probability is unchanged, the more broken bones you have the more likely you are to fall over, it just couldn't happen at all with one (so having 2 broken legs will have the same probability as previously).

Also reduced the stun time from a meaty 3 seconds to 1 second, since you're just falling over, and a hard stun is a really big debuff (and is far more likely to actually occur in game now).

## Why It's Good For The Game
Running on a broken leg has actual consequences
Falling over is no longer a death sentence

## Changelog
:cl:
balance: Having a single broken leg or foot can now trigger falling over instead of needing 2 or more breaks
balance: Falling over due to a broken leg or foot now only stuns for 1 second instead of 3
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_organs.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_organs.dm
@@ -53,9 +53,9 @@
 					E.germ_level++
 
 	//Hard to stay upright
-	if(leg_tally > 1 && prob(2.5 * leg_tally))
+	if(leg_tally > 0 && prob(2.5 * leg_tally))
 		if(!(species.species_flags & NO_PAIN))
 			emote("pain")
 		visible_message(span_warning("[src] collapses to the ground!"),	\
 			span_danger("Your legs give out from under you!"))
-		Knockdown(3 SECONDS)
+		Knockdown(1 SECONDS)


novaepee:
the leg slashing meta begins
started in cm
now in tgmc

Lumipharon:
On average it is gonna only trigger every like, 3 minutes+, so it's not that bad honestly. Delimbing a leg is more effective since 2 missing limbs (leg and foot) doubles the odds.

QualityVan:
Yeah this isn't actually going to make breaking legs any good. Also fun fact it was a 20 second knockdown before.

Al-1ce:
The stun being basically nothing isn't that great. Maybe stagger as well?

GravitronBlue:
The bigger question now is if having a broken leg is worse than entirely missing a leg.

Lumipharon:
missing leg is as bad as having a broken leg and foot.

Naaanii:
Slowdown would be better for broken legs and stun for delimbs imo. Delimbs remove the ability to sprint but broken legs do pretty much nothing but pain as of now.

Random stuns aren't too great as old brain damage used to do that and was removed because of the randomness iirc.

Lumipharon:
> Slowdown would be better for broken legs and stun for delimbs imo. Delimbs remove the ability to sprint but broken legs do pretty much nothing but pain as of now.
> 
> Random stuns aren't too great as old brain damage used to do that and was removed because of the randomness iirc.

Organ damage is different in that it can't be self treated, only peri+ or being sent shipside can treat it, and you don't even get feedback that you've got it until it starts fucking you.

Broken bones can be self splinted to entirely mitigate the negative effects, and having a permanent slowdown on top of the pain and likely brute damage you've received would honestly be more terrible to deal with than the a 1 second knockdown every few minutes on an unsplinted leg.

zzurge:
good thing you had the foresight to remove the ridonculous falldown time for broke legs otherwise this would have been a yike

PR #10825, opened 2022-08-24
Add an auxiliary fuel tank for the condor

lbnesquik:
## About The Pull Request
This PR adds a fuel tank to the condor. It takes a weapon slot, but allows for a longer stay on target.

I didn't have a sprite for it.
image

Things to do : 
- [ ] Get yelled at for ugly code
- [ ] Get a semi decent sprite for it
- [x] Get feedback for the price
- [ ] Optional : Make it un-selectable in the condor screen

Actually, we don't care. If someone wants to have 4 extra fuel tanks and only the minigun, they are free to do that choice.~~- [ ] Make it so you can only have one fuel tank on the condor~~

## Why It's Good For The Game
A little more choice in your kit besides which flavor of xeno killing you like.

## Changelog
:cl:
add: Added an auxiliary fuel tank for the condor
/:cl:

### DIFF
--- a/code/game/objects/structures/dropship_equipment.dm
+++ b/code/game/objects/structures/dropship_equipment.dm
@@ -49,6 +49,9 @@
 		return TRUE
 	if(installed_equipment || clamp.loaded != loaded_equipment)
 		return TRUE
+	if(loaded_equipment.attempt_attach(ship_tag))
+		to_chat(user, span_warning("Error, baby."))
+		return TRUE
 	to_chat(user, span_notice("You install [loaded_equipment] on [src]."))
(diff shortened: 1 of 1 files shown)


Al-1ce:
If you're open to adjusting things, would it be feasable to add some sort of weapon for CAS to launch it's own flares? There's already a regular flare launcher, but CAS can't use the light from that for aiming. I've seen entire rounds go by without a single green flare being thrown and relying on marines solely really sucks.

RikuTheKiller:
personally? fuel for a weapon slot seems fair enough.

also more flight time is always great.

PR #10834, opened 2022-08-25
Add an item which allows the AI to remote target for artillery pieces. 

lbnesquik:
## About The Pull Request
The device is a little module which you can attach to the mortar or howitzer. Doing so will link it with the current active AI, if there is one. 

From there, the AI can use shift click on a turf to transmit coordinates to the linked artillery piece. It can also click on the piece itself to unlink itself from it, at which point the device falls to the ground. 

It can be found in req and engineers lockers at a fairly high price. 

## Why It's Good For The Game
I always felt like the AI could use more ways to do some sort of overwatch+, which currently, it currently can't really do. 
This PR would give it a boon for helping to coordinate marines, and bring it a little more towards a command role.

I think this is a buff, but it isn't quite as strong as it may sound. For one thing, the AI vision remains limited to the view afforded to it by cameras, and it lacks dark vision. This is a **lot** less range than the rangefinders allow, so no surprise sniping. It also costs a fair bit of investment, something which could be lowered if the device doesn't actually perform.

## Changelog
:cl:
add: Added a device which allows the AI to remotely target artillery pieces. 
/:cl:


Naaanii:
Seems unnecessary as Mortar and Howies already come with range finders and generally a spotter fires shells themselves.

AI being able to get coordinates past the range of range finders is a bit powercreepy as howies don't have accuracy maluses with range. AI having stuff to do isn't bad but this seems a bit unnecessary.

Then again, AIs can also access consoles to quick fire OBs instead of FCs or SLs having to wait for the lase.

Weedmaster77:
Like the idea since I have seen the AI give cords manual to Howy gunner so this make it easier for the AI and gunner and also makes it a item. Would just hope its not more then like 25-30 points in REQ since it seems like such a uncommon sight to have a AI working with Howy crew.

Starfirejordan:
good luck both your pr are idea I had that tivi already denied soooo good luck with that

Al-1ce:
> the lack of accuracy maluses for range makes this a bit powercreepy.

There are no accuracy maluses for any piece of artillery in the game, outside of operator error. Why would it matter now, all of the sudden?

Naaanii:
> > the lack of accuracy maluses for range makes this a bit powercreepy.
> 
> There are no accuracy maluses for any piece of artillery in the game, outside of operator error. Why would it matter now, all of the sudden?

There used to be and still is with Mortar iirc, Howie doesn't have any. This just allows deeper range of artillery without having to actually move the howitzer closer to the frontline.

Then again, manual coordinates could be given but also, less downtime, hence powercreepy.

BraveMole (changes requested):
I don't really see a lot of situation where this will be usefull. 

Ai has limited vision due to camera, you are way better off with a marine spotter.

code/game/objects/machinery/mortar.dm
```diff
@@ -231,6 +238,18 @@
 		addtimer(CALLBACK(GLOBAL_PROC, .proc/playsound, T, fall_sound, 50, 1), travel_time)
 		addtimer(CALLBACK(src, .proc/detonate_shell, T, mortar_shell), travel_time + 45)//This should always be 45 ticks!
 
+	if(istype(I, /obj/item/ai_target_beacon))
+		for(var/mob/living/silicon/ai/AI AS in GLOB.ai_list)
```
BraveMole, line 242: Should not be a for, because that code doesn't work if there are multiple ais.

Also, does this work if there are no ais ?

code/game/objects/machinery/mortar.dm
```diff
@@ -253,6 +272,20 @@
...
+	if(user.linked_artillery == src)
+		switch(tgui_alert(usr, "This artillery piece is linked to you. Do you want to unlink yourself from it?", "Artillery Targeting", list("Yes", "No")))
+			if("Yes")
+				user.clean_artillery_refs()
+			if("No")
+				return
+
+/obj/machinery/deployable/mortar/proc/unset_targeter()
+	say("Linked AI spotter has relinquished targeting privileges. Ejecting targeting device.")
+	ai_targeter.loc = src.loc
```
BraveMole, line 285: A direct loc set ? Also should use get turf

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -387,6 +390,22 @@
 		reset_perspective()
 	remote_control = controlled
 
+/mob/living/silicon/ai/proc/associate_artillery(mortar)
+	if(linked_artillery)
+		UnregisterSignal(linked_artillery, COMSIG_PARENT_QDELETING)
+	if(linked_artillery == mortar)
```
BraveMole, line 396: Only usefull if linked artillery is not null, so put that in the other if scope

BraveMole (changes requested):

code/game/objects/machinery/mortar.dm
```diff
@@ -47,6 +47,13 @@
 	///Used for round stats
 	var/tally_type = TALLY_MORTAR 
 
+	var/obj/item/ai_target_beacon/ai_targeter
```
BraveMole, line 50: Autodoc

code/game/objects/machinery/mortar.dm
```diff
@@ -253,6 +276,20 @@
...
+	if(user.linked_artillery == src)
+		switch(tgui_alert(usr, "This artillery piece is linked to you. Do you want to unlink yourself from it?", "Artillery Targeting", list("Yes", "No")))
+			if("Yes")
+				user.clean_artillery_refs()
+			if("No")
+				return
+
+/obj/machinery/deployable/mortar/proc/unset_targeter()
+	say("Linked AI spotter has relinquished targeting privileges. Ejecting targeting device.")
+	ai_targeter.loc = get_turf(src)
```
BraveMole, line 289: You need to use move or forcemove

BraveMole (changes requested):

code/modules/reqs/supplypacks.dm
```diff
@@ -642,6 +642,12 @@ EXPLOSIVES
 	cost = 2
 	available_against_xeno_only = TRUE
 
+/datum/supply_packs/explosives/ai_target_module
+	name = "AI targeting module"
+	contains = list(/obj/item/ai_target_beacon)
+	cost = 50
```
BraveMole, line 648: that's hella expensive.

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -387,6 +390,21 @@
 		reset_perspective()
 	remote_control = controlled
 
+/mob/living/silicon/ai/proc/associate_artillery(mortar)
```
BraveMole, line 393: Autodoc

code/game/objects/machinery/mortar.dm
```diff
@@ -253,6 +277,20 @@
...
 
+/obj/machinery/deployable/mortar/attack_ai(mob/living/silicon/ai/user)
+	if(user.linked_artillery == src)
+		switch(tgui_alert(usr, "This artillery piece is linked to you. Do you want to unlink yourself from it?", "Artillery Targeting", list("Yes", "No")))
+			if("Yes")
+				user.clean_artillery_refs()
+			if("No")
+				return
+
+/obj/machinery/deployable/mortar/proc/unset_targeter()
```
BraveMole, line 288: Doc

code/game/objects/machinery/mortar.dm
```diff
@@ -253,6 +277,20 @@
 	qdel(mortar_shell)
 	firing = FALSE
 
+/obj/machinery/deployable/mortar/attack_ai(mob/living/silicon/ai/user)
+	if(user.linked_artillery == src)
+		switch(tgui_alert(usr, "This artillery piece is linked to you. Do you want to unlink yourself from it?", "Artillery Targeting", list("Yes", "No")))
```
BraveMole, line 282: You can do the same thing with a single if. Including the user.linked artillery == src

code/game/objects/machinery/mortar.dm
```diff
@@ -231,6 +239,22 @@
...
+		AI = tgui_input_list(usr, "Which AI would you like to associate this gun with?", null, GLOB.ai_list)
+		if(!AI)
+			return
+		to_chat(user, span_notice("You attach the [I], allowing for remote targeting."))
+		to_chat(AI, span_notice("NOTICE - [src] has been linked to your systems, allowing for remote targeting. Use shift click to set a target."))
+		user.transferItemToLoc(I, src)
+		AI.associate_artillery(src)
+		playsound(loc, 'sound/items/ratchet.ogg', 25, 1)
+		ai_targeter = I
+		return
```
BraveMole, line 257: No need for a return at the end of the proc + add an empty line anyway

code/game/objects/machinery/mortar.dm
```diff
@@ -231,6 +239,22 @@
 		addtimer(CALLBACK(GLOBAL_PROC, .proc/playsound, T, fall_sound, 50, 1), travel_time)
 		addtimer(CALLBACK(src, .proc/detonate_shell, T, mortar_shell), travel_time + 45)//This should always be 45 ticks!
 
+	if(istype(I, /obj/item/ai_target_beacon))
+		if(!GLOB.ai_list)
+			to_chat(user, span_notice("There is no AI to associate with."))
+			return
+		var/mob/living/silicon/ai/AI
```
BraveMole, line 246: Why are you defining then declaring in two different lines

BraveMole (changes requested):

code/game/objects/machinery/mortar.dm
```diff
@@ -231,6 +239,21 @@
 		addtimer(CALLBACK(GLOBAL_PROC, .proc/playsound, T, fall_sound, 50, 1), travel_time)
 		addtimer(CALLBACK(src, .proc/detonate_shell, T, mortar_shell), travel_time + 45)//This should always be 45 ticks!
 
+	if(istype(I, /obj/item/ai_target_beacon))
+		if(!GLOB.ai_list)
```
BraveMole, line 243: you need to check the len, that's cleaner.

BraveMole (changes requested):

Derrick2Electricboogaloo:
FINALLY!

Derrick2Electricboogaloo:
> I don't really see a lot of situation where this will be usefull.
> 
> Ai has limited vision due to camera, you are way better off with a marine spotter.

Not to mention it makes buying and building camera's a lot more useful other than vision. And hopefully encourages people to utilize that type of equipment.

PR #10849, opened 2022-08-28
Adds a buffer between a limb's max damage and the brute required to sever

QualityVan:
## About The Pull Request
Just about the same brute damage as before on all limbs, but limbs can hold more damage than before so you need 20-30 burn damage to become immune to amputations instead of any at all.

## Why It's Good For The Game
Solves tactical 1 burn damage with minimal potential for increasing burn damage potency on the whole.

## Changelog
:cl:
balance: You need 20-30 burn damage on a limb in order for it to be immune to getting chopped off, instead of any at all.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -1,3 +1,5 @@
+#define LIMB_MAX_DAMAGE_SEVER_RATIO 0.8
+
 /****************************************************
 				EXTERNAL ORGANS
 ****************************************************/
@@ -13,7 +15,7 @@
 	var/brute_dam = 0
 	///burn damage this limb has taken as a part
 	var/burn_dam = 0
-	///Max damage the limb can take before being destroyed
+	///Max damage the limb can take. Extremities sever when they have at least LIMB_MAX_DAMAGE_SEVER_RATIO as a fraction of this in brute damage.
 	var/max_damage = 0
 	var/max_size = 0
 	var/last_dam = -1
@@ -264,7 +266,7 @@
 		if(updating_health)
 			owner.updatehealth()
 		return update_icon()
-	if(CONFIG_GET(flag/limbs_can_break) && brute && brute_dam + burn_dam >= max_damage * CONFIG_GET(number/organ_health_multiplier))
+	if(CONFIG_GET(flag/limbs_can_break) && brute_dam >= max_damage * LIMB_MAX_DAMAGE_SEVER_RATIO && CONFIG_GET(number/organ_health_multiplier))
 		droplimb() //If the limb's at its max damage and takes any brute, it severs
 		if(!(owner.species && (owner.species.species_flags & NO_PAIN)))
 			owner.emote("scream")
@@ -1010,7 +1012,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	name = "l_arm"
 	display_name = "left arm"
 	icon_name = "l_arm"
-	max_damage = 125
+	max_damage = 150
 	min_broken_damage = 50
 	body_part = ARM_LEFT
 	cover_index = 7
@@ -1023,7 +1025,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	name = "l_leg"
 	display_name = "left leg"
 	icon_name = "l_leg"
-	max_damage = 100
+	max_damage = 125
 	min_broken_damage = 50
 	body_part = LEG_LEFT
 	cover_index = 14
@@ -1033,7 +1035,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 	name = "r_arm"
 	display_name = "right arm"
 	icon_name = "r_arm"
-	max_damage = 125
+	max_damage = 150
 	min_broken_damage = 50
 	body_part = ARM_RIGHT
 	cover_index = 7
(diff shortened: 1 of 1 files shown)


Lumipharon:
As mentioned in discord, this completely breaks HvH balance.

Losenis:
You can get burn damage even if you're dead by being on a fire tile, this'll absolutely skyrocket the number of delimbs and make magmoor essentially impossible to play

Serxule:
>  make magmoor essentially impossible to play

who takes magmoor in account for balance? its literally never chosen

blackdav123:
> > make magmoor essentially impossible to play
> 
> who takes magmoor in account for balance? its literally never chosen

it is relevant because this would mean falling into lava and dying would result in all your limbs falling off in the goofiest way possible

better solution would be to allow burn damage to reach up to the damage cap and make any subsequent brute damage cause the delimb

novaepee:
> > make magmoor essentially impossible to play
> 
> who takes magmoor in account for balance? its literally never chosen

Magmoor is gone.

Losenis:
It only takes a map getting another "get burn damage if you step on it" hazard before the argument is back in play. Flamers already do burn damage to corpses by the fire tile anyway.

BraveMole (changes requested):
Feels like the solution has vastly more impact than just fixing the bug.

Serxule:
probably just make it so delimbing requires brute damage when damage is maxed and that fixes the problem (unless its already like that)

Weedmaster77:
Don't like this idea at all it's just going make more marines get sent shipside for new limbs when it happens between 3 and 10 times a round on a pretty far non-steamroller round. Delimbing is powerful enough with brute damage burn doesn't need to make it even more punishing then it already is. An when ravs and warriors can just pop your arms, hands and legs off pretty easy with enough targeting which skilled xeno plays do pretty often. Also with light armor being the most commonly used armor delimbing is frequent enough . An with my many hours of un-suiting dead marines to rev I can confidently say that.

Al-1ce:
Generally balancing your PR off of somebody falling into lava is a bad idea.

Losenis:
It's not just lava, it's that tiles that do burn damage continue to do so even when the person is dead. See also: space, blue fire, WP smoke (probably), both combined in incen OB, some rare bug floating around that had corpses constnatly get tiny bits of damage that I never could pinpoint enough to bug report, etc. Lava just happens to be the thing most people already are familiar with, which is why it's the center of the argument. Magmoor wasn't the only one with map hazards either, there was that other map with the green toxin rivers that did the same thing without setting you on fire, as well as barrenquilla when it was in rotation. With brute damage it also happens but there aren't that many things out there to actually make it as pronounced, though one thing you can do in HvH is take a corpse's helmet off, help intent click it with knife, and you'd keep doing brute damage despite dead till you decap it.

The change clashes with too many things that may be bugs, oversights, intended mechanics, and so on for its own good.

QualityVan:
Took the suggestion of requiring a brute hit in addition to reaching damage cap to cause a delimb. Should prevent the concerns raised about hvh and lava/fire.

Losenis:
inb4 delimbed by a cigarette
image

BraveMole (changes requested):
Like we talked in discord, I think I'd prefer having the delimb threshold be a percentage of max_damage. Let's say 80-90 %

RikuTheKiller:
QV this overall makes it easier to delimb people.

I raised this point when the idea was originally brought up as well.

RikuTheKiller:
Right now this makes arms easier to delimb. I'd recommend making it a flat 25 instead of a ratio so you can always just use that as a reference. (Take for example the arms, 150 isn't a 20% increase to 125, get your math right.)

Losenis:
Is this actually necessary at all? Even a gamer that burns himself to avoid delimb just trades that to get killed faster assuming he actually burns the correct limb. Delimb is already extremely punishing because it means you get a robot limb that can't get healed by chems nor bandages/ointment and needs you to carry two more items they may not really have space for without a loadout change midround. At that point I personally just DNR and wait for the next round because even if it's possible to play like that it's not terribly fun.

RikuTheKiller:
> Is actually necessary at all?

Yes, since you don't have to do it intentionally to benefit from it, just take some acid or flame damage and you're fully immune to delimbs, no intention to abuse this required.

Losenis:
"benefit from it"

Bro you're dead

Texicos:
Death is preferable to a fracture or delimb

Losenis:
Not playing is preferable tbh

code/modules/organs/limbs.dm
```diff
@@ -264,8 +266,8 @@
 		if(updating_health)
 			owner.updatehealth()
 		return update_icon()
-	if(CONFIG_GET(flag/limbs_can_break) && brute_dam >= max_damage * CONFIG_GET(number/organ_health_multiplier))
-		droplimb() //Reached max damage threshold through brute damage, that limb is going bye bye
+	if(CONFIG_GET(flag/limbs_can_break) && brute_dam >= max_damage * LIMB_MAX_DAMAGE_SEVER_RATIO * CONFIG_GET(number/organ_health_multiplier))
+		droplimb() //If the limb's at its max damage and takes any brute, it severs
```
QualityVan, line 270:
```suggestion
		droplimb()
```

PR #10851, opened 2022-08-28
Improves the  wide flamer nozzle and related refactor

Lumipharon:
## About The Pull Request
Edit: I have no reduced the flame duration of the wide nozzle by 20%, which means the wide nozzle will no longer be able to kill walls in a single click. You can still just click twice in a row to kill them, but it means you'll absolutely blast through fuel like nothing else.
I think making any flamer be able to kill walls in one click was a questionable change, but I'll reserve changing overall flamer behavior to another pr as that's out of scope.

So this was a rabbit hole of code to rework, guh.

Changed the way the wide nozzle calculates the turfs to burn, giving it more of a natural icecream cone sort of shape instead of a wide pyramid.

https://user-images.githubusercontent.com/7869430/187071890-a785afd5-f6ee-426f-9cbc-9da550b455fa.mp4

This should hopefully be actually useful, compared to the current iteration which is frankly niche if not outright bad.

Also refactors a bunch of flamer code, and some global procs related to CONES (literally the worst), and calculating valid pathing between turfs.
Notably there is now a cone proc that actually works (so for an obvious example prae acid cone could be changed as well, among other things), and flamers will now actually burn turfs that have border objects on them like windoors or barricades where it makes sense for them to do so.

I've tested this pretty thoroughly, but probs TM it anyway. Also maybe people will think that the range or angle of the cone should be tweaked.

## Why It's Good For The Game
Wurf flammen betterer
Bug fix good
I have finally conquered the vicious cone

## Changelog
:cl:
balance: Changed the AOE calculation for the wide flamer nozzle to be more useful
fix: fixed flamers not igniting tiles with border objects like barricades or windoors in suitable cases
refactor: refactored some flamer and cone code
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -161,53 +161,41 @@ GLOBAL_REAL_VAR(list/stack_trace_storage)
 
 
 ///Returns true if the path from A to B is blocked. Checks both paths where the direction is diagonal
-/proc/LinkBlocked(turf/A, turf/B, bypass_window = FALSE, projectile = FALSE)
+/proc/LinkBlocked(turf/A, turf/B, bypass_window = FALSE, projectile = FALSE, bypass_xeno = FALSE)
 	if(isnull(A) || isnull(B))
 		return TRUE
 	var/adir = get_dir(A, B)
 	var/rdir = get_dir(B, A)
 	if(adir & (adir - 1))//is diagonal direction
 		var/turf/iStep = get_step(A, adir & (NORTH|SOUTH))
-		if(!iStep.density && !LinkBlocked(A, iStep) && !LinkBlocked(iStep, B))
+		if((!iStep.density || (istype(iStep, /turf/closed/wall/resin) && bypass_xeno)) && !LinkBlocked(A, iStep, bypass_window, projectile, bypass_xeno) && !LinkBlocked(iStep, B, bypass_window, projectile, bypass_xeno))
 			return FALSE
 
 		var/turf/pStep = get_step(A,adir & (EAST|WEST))
-		if(!pStep.density && !LinkBlocked(A, pStep) && !LinkBlocked(pStep, B))
+		if((!pStep.density || (istype(pStep, /turf/closed/wall/resin) && bypass_xeno)) && !LinkBlocked(A, pStep, bypass_window, projectile, bypass_xeno) && !LinkBlocked(pStep, B, bypass_window, projectile, bypass_xeno))
 			return FALSE
 		return TRUE
 
-	if(DirBlocked(A, adir, bypass_window, projectile))
+	if(DirBlocked(A, adir, bypass_window, projectile, bypass_xeno))
 		return TRUE
-	if(DirBlocked(B, rdir, bypass_window, projectile))
+	if(DirBlocked(B, rdir, bypass_window, projectile, bypass_xeno))
 		return TRUE
 	return FALSE
 
 ///Checks if moving in a direction is blocked
-/proc/DirBlocked(turf/loc, direction, bypass_window = FALSE, projectile = FALSE)
-	for(var/obj/structure/window/D in loc)
-		if(!D.density || bypass_window) //if the window is not dense, or we want to ignore windows, for example lasers
+/proc/DirBlocked(turf/loc, direction, bypass_window = FALSE, projectile = FALSE, bypass_xeno = FALSE)
+	for(var/obj/object in loc)
+		if(!object.density)
 			continue
-		if(D.is_full_window())
-			return TRUE
-		if(D.dir == direction) //border edge windows only block movement in that direction
-			return TRUE
-
-	for(var/obj/machinery/door/D in loc)
-		if(!D.density) //the door is open
+		if(object.throwpass && projectile) //projectiles can bypass dense objects with throwpass
 			continue
-		if(istype(D, /obj/machinery/door/window))
-			if(D.dir == direction && !bypass_window) //border edge windoors only block movement in that direction
-				return TRUE
-		else
-			return TRUE	// it's a real, air blocking door
-	for(var/obj/structure/mineral_door/D in loc)
-		if(D.density)
-			return TRUE
-	for(var/obj/structure/barricade/B in loc)
-		if(!B.density || projectile) //anything that behaves like a projectile can travel over a barricade
+		if((istype(object, /obj/structure/mineral_door/resin) || istype(object, /obj/structure/xeno)) && bypass_xeno) //xeno objects are bypassed by flamers
 			continue
-		if(B.dir == direction)
-			return TRUE
+		if((istype(object, /obj/machinery/door/window) || istype(object, /obj/structure/window)) && bypass_window) //windows are bypassed energy weapons
+			continue
+		if(object.flags_atom & ON_BORDER && object.dir != direction)
+			continue
+		return TRUE
 	return FALSE
 
 ///Returns true if any object on a turf is both dense and not a window
@@ -1232,7 +1220,7 @@ GLOBAL_LIST_INIT(survivor_outfits, typecacheof(/datum/outfit/job/survivor))
  *	cone_direction - at what angle should the cone be made, relative to the game board's orientation
  *	blocked - whether the cone should take into consideration solid walls
  */
-/proc/generate_true_cone(atom/center, max_row_count = 10, starting_row = 1, cone_width = 60, cone_direction = 0, blocked = TRUE)
+/proc/generate_true_cone(atom/center, max_row_count = 10, starting_row = 1, cone_width = 60, cone_direction = 0, blocked = TRUE, bypass_window = FALSE, projectile = FALSE, bypass_xeno = FALSE)
 	var/right_angle = cone_direction + cone_width/2
 	var/left_angle = cone_direction - cone_width/2
 
@@ -1268,7 +1256,7 @@ GLOBAL_LIST_INIT(survivor_outfits, typecacheof(/datum/outfit/job/survivor))
 				if(turf_angle > right_angle && turf_angle < left_angle) //in the specified angle
 					continue
 				if(blocked)
-					if(T.density || LinkBlocked(trf, T) || turf_blocked_projectile(T, REVERSE_DIR(direction))) //if turf is dense || XXX || if any object on the turf is dense and not a window
+					if(LinkBlocked(trf, T, bypass_window, projectile, bypass_xeno) || turf_blocked_projectile(T, REVERSE_DIR(direction))) //if turf is dense || XXX || if any object on the turf is dense and not a window
 						continue
 				cone_turfs += T
 				turfs_to_check += T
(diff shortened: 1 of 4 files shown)


code/__HELPERS/unsorted.dm
```diff
@@ -160,64 +160,59 @@ GLOBAL_REAL_VAR(list/stack_trace_storage)
...
-
-/proc/LinkBlocked(turf/A, turf/B)
+/**
+ *	Returns true if the path from A to B is blocked. Checks both paths where the direction is diagonal
+ *	Variables:
+ *	bypass_window - whether it will go through transparent windows like lasers
+ *	projectile - whether throwpass will be checked to ignore dense objects like projectiles
+ *	bypass_xeno - whether to bypass dense xeno structures like flamers
+ */
+/proc/LinkBlocked(turf/A, turf/B, bypass_window = FALSE, projectile = FALSE, bypass_xeno = FALSE)
```
QualityVan, line 170: Seems like you could just have a PASSX bitflag here instead of running a separate var for each thing to check.

code/modules/projectiles/guns/flamer.dm
```diff
@@ -69,6 +69,8 @@
 	var/lit_overlay_offset_y = 0
 	///Damage multiplier for mobs caught in the initial stream of fire.
 	var/mob_flame_damage_mod = 2
+	///how wide of a cone the flamethrower produces on wide mode.
+	var/cone_angle = 55
```
QualityVan, line 73: Should this be on the nozzle itself rather than the gun?

QualityVan:
Around 9-10 seconds in the demo vid, why do some flames appear directly under you when you aren't firing?

Lumipharon:
> Around 9-10 seconds in the demo vid, why do some flames appear directly under you when you aren't firing?

Recursive flames is checking for tiles already flamed, relative to where you are. I assumed it's because you're holding down the trigger for a moment. It's not actually the tile I'm on, it's the tile in front of me as soon as I move.
This is intended, but if people think it's fucky it would be easy enough to use the location of src at the time of firing.

code/modules/projectiles/guns/flamer.dm
Lumipharon: There's no other nozzles that change it, and if there was then it would make sense for it to be a modifier to the base gun's angle.

code/__HELPERS/unsorted.dm
Lumipharon: mesmoothbrain, what's that?

QualityVan: Just a single var to use any combination of https://github.com/tgstation/TerraGov-Marine-Corps/blob/master/code/__DEFINES/equipment.dm#L1
You'd need to add a define for xeno walls, not 100% sure if any apply perfectly to what you want with bullets.

blackdav123:
cone flamer being a niche tool that is generally bad was what it is designed for

if xenos are walling off entire sections of cave then you use cone flamer to stop that, this would make it the best maze clearing option in all scenarios

BraveMole:
> > Around 9-10 seconds in the demo vid, why do some flames appear directly under you when you aren't firing?
> 
> Recursive flames is checking for tiles already flamed, relative to where you are. I assumed it's because you're holding down the trigger for a moment. It's not actually the tile I'm on, it's the tile in front of me as soon as I move. This is intended, but if people think it's fucky it would be easy enough to use the location of src at the time of firing.

Look funky as hell yes

Lumipharon:
Updated to only run the code proc once and instead burn the tile in them layer by layer, so it should be a bit more efficient without changing it's behavior.

Also changed it to use get_turf(src) when first fired, so the cone won't shift if you move/get teleported etc.

code/__HELPERS/unsorted.dm
```diff
@@ -1210,4 +1205,57 @@ will handle it, but:
...
+	var/list/turfs_to_check = list(get_turf(center))
+	var/list/cone_turfs = list()
+
+	for(var/r in 1 to max_row_count)
+		if(r > 2)
+			cardinals = GLOB.cardinals
+		for(var/X in turfs_to_check) //checks the inital turf, then afterwards checks every turf that is added to cone_turfs
+			var/turf/trf = X
+			for(var/direction in cardinals)
+				var/turf/T = get_step(trf, direction) //checks all turfs around X
```
TiviPlus, line 1242: use AS, and please use proper var names this reads like algebra

TiviPlus (changes requested):
Rest lgtm I spose

code/__HELPERS/unsorted.dm
```diff
@@ -1210,4 +1205,57 @@ will handle it, but:
...
+	var/left_angle = cone_direction - cone_width/2
+
+	//These are needed because degrees need to be from 0 to 359 for the checks to function
+	if(right_angle >= 360)
+		right_angle -= 360
+
+	if(left_angle < 0)
+		left_angle += 360
+
+	var/list/cardinals = GLOB.alldirs
```
BraveMole, line 1232: is that a copy ?

code/__HELPERS/unsorted.dm
```diff
@@ -1210,4 +1205,57 @@ will handle it, but:
...
+				var/turf/T = get_step(trf, direction) //checks all turfs around X
+				if(cone_turfs.Find(T))
+					continue
+				var/turf_angle = Get_Angle(center, T)
+				if(right_angle > left_angle && (turf_angle > right_angle || turf_angle < left_angle))
+					continue
+				if(turf_angle > right_angle && turf_angle < left_angle)
+					continue
+				if(blocked)
+					if(LinkBlocked(trf, T, bypass_window, projectile, bypass_xeno))
```
BraveMole, line 1251: only one if

code/__HELPERS/unsorted.dm
```diff
@@ -1210,4 +1205,57 @@ will handle it, but:
...
+						continue
+				cone_turfs += T
+				turfs_to_check += T
+			turfs_to_check -= trf
+		for(var/turf/checked_turf in cone_turfs)
```
BraveMole, line 1256: AS

code/modules/projectiles/guns/flamer.dm
BraveMole: I kinda agree with QV here

code/modules/projectiles/guns/flamer.dm
```diff
@@ -150,80 +152,59 @@
...
-		return
+	burn_list(turfs_by_iteration)
 	iteration++
-	addtimer(CALLBACK(src, .proc/recursive_flame_cone, iteration, turfs_to_ignite, dir_to_target, range, current_target), flame_spread_time)
-
-#undef RECURSIVE_CHECK
+	addtimer(CALLBACK(src, .proc/recursive_flame_cone, iteration, turfs_to_ignite, dir_to_target, range, current_target, flame_source), flame_spread_time)
 
 ///Checks and lights the turfs in turfs_to_burn
 /obj/item/weapon/gun/flamer/proc/burn_list(list/turf/turfs_to_burn)
```
BraveMole, line 207: ? looks useless

code/modules/projectiles/guns/flamer.dm
```diff
@@ -150,80 +152,59 @@
...
 		light_pilot(FALSE)
 		return
-
-	if(RECURSIVE_CHECK(old_turfs, range, current_target, iteration))
+	//recursive checks
+	if(iteration > range || !current_target)
 		return
 
+	var/list/turf/turfs_by_iteration = list()
+	for(var/turf/turf in turfs_to_ignite)
```
BraveMole, line 198: AS

BraveMole (changes requested):

code/modules/projectiles/guns/flamer.dm
Lumipharon: The reason why I want to have it on the gun is so that I can add one or more new flamers in the future that different angles, while still being compatible with the sam nozzles.

code/__HELPERS/unsorted.dm
Lumipharon: done

code/__HELPERS/unsorted.dm
Lumipharon: done

code/modules/projectiles/guns/flamer.dm
Lumipharon: done

code/__HELPERS/unsorted.dm
Lumipharon: whoops, used to be more there.

Lumipharon:
ahhh sorry, Was clicking the one button and it was just your names changing place.

BraveMole (approved):

Araulius:
this makes the standard nozzle obsolete, I hope you know that

Lumipharon:
> this makes the standard nozzle obsolete, I hope you know that

It absolutely chews through ammo like nothing. I burned through an entire backtank in like 20 minutes and I wasn't even spamming it as much as I would have liked.

Sure a single blast covers more area than the narrow nozzle, but that's not any more damage on anything you hit, and you can't choose to cover a shorter distance when that's not necessary. It's full blast or nothing.

Araulius:
> > this makes the standard nozzle obsolete, I hope you know that
> 
> It absolutely chews through ammo like nothing. I burned through an entire backtank in like 20 minutes and I wasn't even spamming it as much as I would have liked.
> 
> Sure a single blast covers more area than the narrow nozzle, but that's not any more damage on anything you hit, and you can't choose to cover a shorter distance when that's not necessary. It's full blast or nothing.

yes, but generally if you're a pyro nerd, you're taking point, and you know to bring welding tanks down to the planet. x fuel is gonna become a bajillion times more oppressive. hivelords are going to mald so hard.

Lumipharon:
> > > this makes the standard nozzle obsolete, I hope you know that
> > 
> > 
> > It absolutely chews through ammo like nothing. I burned through an entire backtank in like 20 minutes and I wasn't even spamming it as much as I would have liked.
> > Sure a single blast covers more area than the narrow nozzle, but that's not any more damage on anything you hit, and you can't choose to cover a shorter distance when that's not necessary. It's full blast or nothing.
> 
> yes, but generally if you're a pyro nerd, you're taking point, and you know to bring welding tanks down to the planet. x fuel is gonna become a bajillion times more oppressive. hivelords are going to mald so hard.

you can fire 25 times before a backtank is empty, you're gonna burn through that xfuel real quick

lKiseki:
As per this PR, the wide nozzle is a bit too oppressive. I have attached a screenshot of its range, which is worse than it's ever been.
image

For further reference, this is the standard nozzle flamer.
image

The other nozzle is just a fireball and one tile of fire, so I've decided to omit that.

This is objectively just a direct upgrade (and buff) to the flamer, allowing it to occupy a lot more range than what is healthy for balance. It is likewise complicated to deal with, as it chews through resin structures relatively quickly and leaves little space for xenos to maneuver. Gets worse with X fuel.
Fuel is not exactly an issue when you can bring plenty, and effectively invalidate the amount of fuel that is used up for every flamer shot.

QualityVan:
A: Why'd you nerf line flamer, B: why isn't it in the changelogs

PR #7142, opened 2021-06-22
combat shotgun can be fired one handed

Pariah919:
## About The Pull Request
combat shotgun can be fired one handed 

## Why It's Good For The Game
a much needed buff to a near useless gun

## Changelog
:cl:
balance: you can fire the combat shotgun one handed
/:cl:

PR #10852, opened 2022-08-28
Two Handing added to SH39

Xsdew:
## About The Pull Request

This PR re-adds the requirement to start wielding the SH39 before being able to fire it.
It was removed as apart of [https://github.com/tgstation/TerraGov-Marine-Corps/pull/7142](7142)

## Why It's Good For The Game

Currently one handing the SH39 allows you to juggle up to three shotguns effectively. More so when combined with a shield.
This change is to mitigate some of the oppressiveness of three SH39 being rapid-fire on xenos.

## Changelog
:cl:
add: You are required to start wielding the SH39 before being able to fire it again.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -80,7 +80,7 @@
 	icon_state = "t39"
 	item_state = "t39"
 	fire_sound = 'sound/weapons/guns/fire/shotgun_automatic.ogg'
-	flags_gun_features = GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER
+	flags_gun_features = GUN_CAN_POINTBLANK|GUN_AMMO_COUNTER|GUN_WIELDED_FIRING_ONLY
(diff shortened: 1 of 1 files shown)


kirkiness:
I like unga's arms torn off more

Lumipharon:
If you want to do this because you generally think one handed shotgun is bad, go for it. But this 100% does not address juggling. It's literally just one extra key press, no extra delay, exact same result.

We already see this with db juggling.

It does stop the shield combo though

TiviPlus:
We've chosen this fix because it fixes the issue without adding mob vars like all the other fixes
We realize this means that stuff like shield/shotty will no longer work and people are welcome to PR something that restores it at a later point in time without bringing back juggling

ivanmixo:
You do realise you can juggle guns besides the T39, right? And that it's even more busted?

RikuTheKiller:
This is the worst choice I could've expected.

It doesn't solve the problem, fucks over other shit and the gun itself AND can be bypassed even on the guns that are affected, all for the sake of avoiding 1 - 2 variables.

I'm disappointed, honestly.

PR #10853, opened 2022-08-29
Removes juggling

ivanmixo:
## About The Pull Request

- Removes juggling by using a var on the mob who fired which tracks when the next shot of the gun they fired should be possible (current world time + firing delay of the weapon). If the time isn't up we arent allowed to fire any new guns. 

## Why It's Good For The Game

I should probably write a tirade about how terrible juggling is for the game, but I'll spare you. 

Basically being able to bypass the balancing of slugs (and shotguns in general) of a slow fire rate is not good.

https://user-images.githubusercontent.com/22431091/187104113-70a4b6f0-e539-4706-a436-3a79583272f9.mp4

## Changelog
:cl:
balance: Juggling weapons is no longer possible, see #10848 for details
/:cl:

PR #10856, opened 2022-08-29
Removes Juggling (Alternate)

RikuTheKiller:
## About The Pull Request

Basically #10853 with a different solution to akimbo and no possibility of juggling attachments or anything of the sort.

Prevents things on the same gun from affecting eachother entirely. Instead of ignoring attachments, it saves the item you're holding which is always going to be the guns parent even if it's an attachment and ignores it if it's the same gun. To detect akimbo it uses the same variable and checks if you're holding it in your other hand. This makes sure it's being alternated as you literally cannot bypass the akimbo check without shooting the other gun.

This is kind of a pick & choose which one you want to merge, I guess.

Credit to Ivan for the original PR I based this one off of.

## Why It's Good For The Game

Juggling is shit, read more about it in #10853 

## Changelog
:cl:
del: Removed juggling
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defines.dm
+++ b/code/modules/mob/living/living_defines.dm
@@ -139,3 +139,6 @@
 
 	/// This is the cooldown on suffering additional effects for when we exhaust all stamina
 	COOLDOWN_DECLARE(last_stamina_exhaustion)
+
+	var/lastgundelay //This is the time after which the next gun can be fired in world time.
+	var/lastgun //This is a reference to the last gun that the mob fired.
--- a/code/modules/mob/living/living_defines.dm
+++ b/code/modules/mob/living/living_defines.dm
@@ -140,5 +140,7 @@
 	/// This is the cooldown on suffering additional effects for when we exhaust all stamina
 	COOLDOWN_DECLARE(last_stamina_exhaustion)
 
-	var/lastgundelay //This is the time after which the next gun can be fired in world time.
-	var/lastgun //This is a reference to the last gun that the mob fired.
+	///This is the time after which the next gun can be fired in world time.
+	var/last_gun_delay
+	///This is a reference to the last gun that the mob fired.
+	var/last_gun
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/living_defines.dm
```diff
@@ -139,3 +139,6 @@
 
 	/// This is the cooldown on suffering additional effects for when we exhaust all stamina
 	COOLDOWN_DECLARE(last_stamina_exhaustion)
+
+	var/lastgundelay //This is the time after which the next gun can be fired in world time.
+	var/lastgun //This is a reference to the last gun that the mob fired.
```
BraveMole, line 144: This is extremly messy. I assume you need that to deal with akimbo, but you could probably just use dual_weild var

code/modules/mob/living/living_defines.dm
```diff
@@ -139,3 +139,6 @@
 
 	/// This is the cooldown on suffering additional effects for when we exhaust all stamina
 	COOLDOWN_DECLARE(last_stamina_exhaustion)
+
+	var/lastgundelay //This is the time after which the next gun can be fired in world time.
```
BraveMole, line 143: Convention is to use _ between words.
Also that is not how autodocing is, it is three / and one line before

BraveMole (changes requested):

code/modules/mob/living/living_defines.dm
RikuTheKiller: Its for preventing attachment juggling AND prevents you from just holding an ammoless gun to bypass dual_wield.

It also makes it so that low fire rate guns don't affect the fire rate of their attachments.

The other PR uses a different solution.

code/modules/mob/living/living_defines.dm
RikuTheKiller: Sigh I copied documentation from other vars but sure. I'll fix it once I'm home.

RikuTheKiller: Fixed.

TiviPlus:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/10852#issuecomment-1231960123

PR #10858, opened 2022-08-30
You are now your character when spawned into vallhalla

ivanmixo:
## About The Pull Request

- This PR makes it so vallhalla copies your prefs when you spawn in, in other words you are your character in valhalla

## Why It's Good For The Game

I think it's nicer to be your character rather than some random gen with pink hair and a cyan beard. 

## Changelog
:cl:
qol: You are now your character when spawned into vallhalla
/:cl:

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -910,10 +910,9 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 		new_fallen.regenerate_icons()
 
 	log_game("[key_name(usr)] has joined Valhalla.")
+	client.prefs.copy_to(new_fallen)
 	mind.transfer_to(new_fallen, TRUE)
 	valhalla_job.after_spawn(new_fallen)
-
-
 
 /mob/dead/observer/reset_perspective(atom/A)
 	clean_observetarget()


Serxule:
wont this prevent you from spawning in normally as your character though? im pretty sure its why you get randomized

FosterGrime-Heymynameslouie:
blessed pr, spede merhg



> wont this prevent you from spawning in normally as your character though? im pretty sure its why you get randomized

I believe this only happens if you go to valhalla first in the start of the round and go to respawn 30 minutes later resulting in the "there is already a character with this name!" error. 
I think

Losenis:
Would need to be tested because in valhalla if you die or SSD your character is fully deleted, there's no corpse left or anything to actually hold the same name unlike normal respawn where you'd be able to find your old corpse if you went looking for it.

ivanmixo:
Depends on how the system looks for the same name... which I can't test because my PC bricked itself this morning for no reason.

BraveMole (changes requested):
This will conflict if you already spawned normally

ivanmixo:
> This will conflict if you already spawned normally

Okay, I've tested this and no, it won't. The mobs spawned in valhalla never get their name added to the list of players joined (only the people who join the game through the lobby do), so they never conflict.

The call stack is basically `LateSpawn()` (on `/datum/game_mode`), `create_character()`  (on `/mob/new_player`), `on_spawn()` (on `/mob`, or `/mob/living/carbon/human`, in this case) which finally calls  `update_names_joined_list()` (on `/mob`), which adds the name to `GLOB.real_names_joined`, which is the list that gets checked for if the character name was already used in the round. 

TL;DR: It doesn't conflict, merge pleas.

PR #10861, opened 2022-08-30
Increases the survivability of certain objects and structures

Lumipharon:
## About The Pull Request
Adds soft armour for bullet, laser and energy to certain objects and structures.

A previous PR heavily reduced the HP of these objects and structures, to make it easier for xenos to clear junk out.
This was a sensible QOL change, but in HvH guns simply output far more dps, so these structures are functionally useless for cover, which isn't ideal.

This PR means they are still easily destroyed by melee or acid, but are a bit more survivable when shot by guns
Also slightly increases the coverage for hydrotrays, since I made them lower back when they had a load of HP.

## Why It's Good For The Game
Cover not disintegrating after 2 bullets is good.

## Changelog
:cl:
balance: Increased the durability of some map objects and structures against guns
balance: Hydro trays provide slightly better cover
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/largecrate.dm
+++ b/code/game/objects/structures/crates_lockers/largecrate.dm
@@ -8,6 +8,7 @@
 	var/dropmetal = TRUE
 	resistance_flags = XENO_DAMAGEABLE
 	max_integrity = 40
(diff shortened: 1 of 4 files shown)


TiviPlus (changes requested):
Use the defines for armor pls

Lumipharon:
> Use the defines for armor pls

Guh, I forgot like half the code doesn't use the defines.

TiviPlus (approved):

PR #10862, opened 2022-08-30
Accuracy system tweaks

Lumipharon:
## About The Pull Request
Some minor tweaks to how accuracy and hits are calculated.

Moved the pain and focus order checks to when the bullet is fired instead of at hit. Because this was previously checked only when hitting mobs, this PR means pain and focus orders now actually impact your ability to shoot past cover.

Pain now correctly uses shock_stage instead of traumatic_shock.

Objects being unanchored no longer reduces their coverage by half.

Tweaked the order different affects are calculated when checking to hit. You're unlikely to notice much difference in game other than large xenos won't be so easily hit way off screen anymore.

## Why It's Good For The Game
Improve and clean up accuracy/hit code slightly, certain buffs and maluses working better is good.
bug fix good.

## Changelog
:cl:
code: Some tweaks to projectile accuracy and hit code
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -1638,12 +1638,14 @@
 				var/datum/status_effect/stacking/gun_skill/debuff = living_user.has_status_effect(STATUS_EFFECT_GUN_SKILL_SCATTER_DEBUFF)
 				gun_scatter += debuff.stacks
 
+		if(ishuman(user))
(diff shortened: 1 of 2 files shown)


FosterGrime-Heymynameslouie:
order command actually being useful long term?
blessed

MereWalker:
This is where the fun begins. Orders like Xeno musk but marine... so body musk

BraveMole (approved):

PR #10865, opened 2022-08-31
Add a camera to the tadpole

lbnesquik:
## About The Pull Request
This add a camera to the tadpole. It should be decently out of the way, looking like a dashcam

## Why It's Good For The Game
Fewer moments of wondering WHERE THE HELL IT IS.

## Changelog
:cl:
qol: Added a camera to the tadpole, like alamo. 
/:cl:

### DIFF
--- a/_maps/shuttles/minidropship.dmm
+++ b/_maps/shuttles/minidropship.dmm
@@ -65,6 +65,7 @@
 /area/shuttle/minidropship)
 "u" = (
 /obj/machinery/computer/camera_advanced/shuttle_docker/minidropship,
+/obj/machinery/camera/autoname/mainship,
 /turf/open/floor/mainship/orange,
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
Liveleak footage of tadpole getting bombarded by boilers , defilers and praetorians ? just what i need as command!

Losenis:
More like russian dashcam footage of tadpole hitting marines on landing

Lumipharon:
Make sure the PO can view it on their tablet.

Derrick2Electricboogaloo:
I noticed this a long time ago, especially when we kept getting reports of the tadpole being lost after AI launched it. Maybe we can bring back AI interaction with the tad since the camera is back. That is what really was wrong. Should probably put one of those invisible ones if that is the case.

PR #10871, opened 2022-09-02
Adds Capronine, A New Chemical

RikuTheKiller:
## About The Pull Request

Code improvements are appreciated, some things may be a little hacky such as the damage spreading.

Adds capronine, a new chemical to the game. It heals brute and burn damage by consuming itself and then causes an equal amount of burn damage when leaving the body.

The chemical consumes itself faster and faster the more total damage it has absorbed. This is to the point where absorbing over 100 damage becomes unreasonably difficult, dare I say impossible for any period of time over 15 seconds. The effect only increases if you try going past that. (It practically purges itself instantly at around 150 damage.)

This essentially stores damage for later. It avoids fractures and delimbs, however, it causes immense pain when the burn damage is dealt.

The chemical prioritizes healing brute damage and whatever remains of the total 20 healing per cycle is used on burn damage.

Capronine has a low OD of 15u that causes 10 toxin damage per cycle. (Infinite damage soaking wouldn't be very balanced.)

Capronine has a recipe of 1 part quick-clot, 2 parts paracetamol, 1 part chlorine and 1 part lithium in exchange for 5 parts capronine.

Capronine's main purpose is to be used as a more advanced option for easily running to get medical care/survive long enough to get medical care OR to use it before/in a fight to absorb incoming damage so you don't get advanced injuries.

Math references for nerds:
Inefficiency = max(1, damage_stored * 0.02)
Capronine Metabolization = (0.1 + Amount Healed / 10) * Inefficiency
Total Brute OR Burn Damage < 1 Is Ignored
Total Potential Heal Per Cycle = min(volume * 10, 20 * effect_str)
Which means that it's basically 20 damage stored per cycle with brute being prioritized first. If there's still some burn damage left and the remaining_heal value is above 0, it'll use the rest of it on that. This means that brute damage is healed first and burn damage is healed afterwards.

## Why It's Good For The Game

We don't really have anything that can offset damage like this. (Unless you count painkillers, but that's different for a number of reasons.) Also, why are new chemicals good for the game? Because they're cool and fun to use. This is also specifically for brute and burn damage, not for overall pain. You can't offset organ damage, cloneloss, toxloss, oxyloss, fractures or anything else for that matter.

## Changelog
:cl:
add: Added capronine, a healing chemical
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1345,7 +1345,7 @@
 	name = "Capronine"
 	description = "A chemical that rapidly consumes itself to heal wounds. Causes immense burning when leaving the body."
 	color = "#dada00"
-	custom_metabolism = REAGENTS_METABOLISM * 0.5
+	custom_metabolism = 0 //Handles metabolism on it's own since it's based off of the inefficiency variable. Actual metabolism is 0.2u/cycle times inefficiency.
 	scannable = TRUE
 	taste_description = "sweetness, with a metallic aftertaste"
 	overdose_threshold = REAGENTS_OVERDOSE * 0.5
@@ -1357,7 +1357,7 @@
 /datum/reagent/medicine/capronine/on_mob_life(mob/living/L, metabolism)
 	var/remaining_heal = min(volume*10, 20*effect_str)
 	var/amount = 0
-	var/inefficiency = min(1, damage_stored*0.02)
+	var/inefficiency = min(1, damage_stored*0.05)
 	if(L.getBruteLoss(TRUE) >= 1 && remaining_heal)
 		amount = min(remaining_heal, L.getBruteLoss())
 		damage_stored += amount
@@ -1371,8 +1371,8 @@
 		L.reagents.remove_reagent(/datum/reagent/medicine/capronine, amount*0.1*inefficiency)
 		L.heal_overall_damage(0, amount)
 		to_chat(L, span_notice("Your burns are disappearing!"))
-	if (volume)
-		return ..()
+	L.reagents.remove_reagent(/datum/reagent/medicine/capronine, REAGENTS_METABOLISM*0.5*inefficiency*effect_str)
+	return ..()
 
 /datum/reagent/medicine/capronine/on_mob_delete(mob/living/L, metabolism)
 	if (damage_stored)
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1345,7 +1345,7 @@
 	name = "Capronine"
 	description = "A chemical that rapidly consumes itself to heal wounds. Causes immense burning when leaving the body."
 	color = "#dada00"
-	custom_metabolism = 0 //Handles metabolism on it's own since it's based off of the inefficiency variable. Actual metabolism is 0.2u/cycle times inefficiency.
+	custom_metabolism = REAGENTS_METABOLISM*0.5 //Changes constantly since it's based off of the inefficiency variable. Actual metabolism is 0.1u/cycle times inefficiency.
 	scannable = TRUE
 	taste_description = "sweetness, with a metallic aftertaste"
 	overdose_threshold = REAGENTS_OVERDOSE * 0.5
@@ -1372,7 +1372,7 @@
 		L.heal_overall_damage(0, amount)
 		to_chat(L, span_notice("Your burns are disappearing!"))
 	if (volume)
-		L.reagents.remove_reagent(/datum/reagent/medicine/capronine, REAGENTS_METABOLISM*0.5*inefficiency*effect_str)
+		custom_metabolism = REAGENTS_METABOLISM*0.5*inefficiency
 		return ..()
 
 /datum/reagent/medicine/capronine/on_mob_delete(mob/living/L, metabolism)
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1379,9 +1379,9 @@
 	if (damage_stored)
 		var/message
 		switch(damage_stored)
-			if(0 to 2)
+			if(0 to 20)
 				message = span_danger("You feel something burning under your skin.")
-			if(2 to 5)
+			if(20 to 50)
 				message = span_userdanger("You feel intense burning all over your body!")
 			else
 				message = span_userdanger("Your entire body is burning up from the inside!")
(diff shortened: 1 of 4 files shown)


Losenis:
What's the actual intended purpose of this chem? I honestly only see it being useful as funny premed for quite literally killing yourself to stop critdrag, which shouldn't really happen unless you're going PFC+ with the corpsman slot if I guess this right in that the average marine isn't going to get this roundstart.

I don't see it being used to heal others outside of combat because you're better off giving bicaridine and kelotane after kitting. I wouldn't use it on myself because of the above on top of meraderm, and add other funny mixes like oxyc and neuraline that you can make yourself.

I'd probably PB with buck whoever gave it to me midcombat because in practice it'd only cause me to drop dead since I'd not be able to actually tell in the chatflood and screen combat that I was given that specific chem and that I'm about to get a shitton of damage to crit or kill me in a couple seconds, like what happened with the old century-long stun russian red had. The lemoline requirement on top of this doesn't really help either, and not least needing actual dermaline to make it.

And it doesn't really have any use in healing people that got revived, so that's another thing it's not really useful for.

RikuTheKiller:
It's mainly intended for shrugging off damage until later. I could change the recipe, though, since it's far less powerful than it was originally intended to be. (With the recipe taken into account.)

RikuTheKiller:
A few, yeah.
1. I didn't see a single chemical in the entire medicine list with any variables.
2. It serves as a way of telling when the chemical should be dosed again if you want to use it out of combat. And no, there really isn't any other reason for it than scanning, consuming itself to make it and, well, yeah.
3. The chemical has a 12u OD that deals 10 toxin damage every tick. (Marines shouldn't try using it more than once.)
4. It converts all brute and burn damage into spread-out burn damage once it's done, which causes immense pain and takes longer to heal using bandages or kits. It's meant to serve as a temporary method of removing the damage and avoiding any advanced injuries.

Anyways I see your point. Time to convert the chem into a damage_stored var. (Easier to explain the chem anyways.)

Jordanstolethepreciousslam:
Oh no. A cursed lovechild of Red Russian and Nanites.

code/modules/reagents/reagents/medical.dm
```diff
@@ -1340,3 +1340,68 @@
...
+	to_chat(L, span_userdanger("Your wounds start to seemingly disappear!"))
+
+/datum/reagent/medicine/curine/on_mob_life(mob/living/L, metabolism)
+	var/remaining_heal = min(volume*10, 20*effect_str)
+	var/amount = 0
+	if(L.getBruteLoss(TRUE) >= 1 && remaining_heal)
+		amount = min(remaining_heal, L.getBruteLoss())
+		L.reagents.remove_reagent(/datum/reagent/medicine/curine, amount*0.1)
+		L.reagents.add_reagent(/datum/reagent/solidcurine, amount*0.1)
+		L.heal_overall_damage(amount, 0)
```
Jordanstolethepreciousslam, line 1363: So this can heal up to 20 burn/brute per tick

code/modules/reagents/reagents/medical.dm
```diff
@@ -1340,3 +1340,68 @@
...
+
+/datum/reagent/medicine/curine/on_mob_add(mob/living/L, metabolism)
+	to_chat(L, span_userdanger("Your wounds start to seemingly disappear!"))
+
+/datum/reagent/medicine/curine/on_mob_life(mob/living/L, metabolism)
+	var/remaining_heal = min(volume*10, 20*effect_str)
+	var/amount = 0
+	if(L.getBruteLoss(TRUE) >= 1 && remaining_heal)
+		amount = min(remaining_heal, L.getBruteLoss())
+		L.reagents.remove_reagent(/datum/reagent/medicine/curine, amount*0.1)
```
Jordanstolethepreciousslam, line 1361: And metabolisation can reach a max of 4.1

Jordanstolethepreciousslam:
Am i seeing things or this heals up to 20 brute and 20 burn per tick?
This is just powercrept Red Russian with virtually no meaningful drawback, other than having to walk to the nearest corpsman, die near them and get revived instantly without fractures and IB's because xeno just weren't able to hurt you with your 57-114 EHP regen per tick

Jordanstolethepreciousslam:
Aside from that i don't think marines need yet another way to die without consequences. this is already an issue

RikuTheKiller:
> Aside from that i don't think marines need yet another way to die without consequences. this is already an issue

It's almost like an earlier comment just said I was majorly changing the chem.

RikuTheKiller:
After extensive testing and trying to figure out a single bug I can say this is fairly balanced.

1. I hit myself twice with a harvester. (71 brute damage.)
2. I popped a 10u pill of capronine.
3. Capronine drained to around 2u rather quickly.
4. Capronine started to slowly drain away.
5. Popped another 10u pill of capronine.
6. Hit myself once with a harvester.
7. Capronine drained to around 2u almost immediately.

Suffice to say this means it's far less efficient the more damage you try to absorb with it. (And the passive metabolization rate is also increased!)

RikuTheKiller:
Recipe is now much easier since you can't bullshit over 100 damage with it. I also added 5 autoinjectors of this with 2 doses each into the MarineMed so people actually get to use it. (It's for marines to use, not really for doctors.)

You can't really combine this with any other premeds, since it'll just store any damage your other premeds would heal. (And it makes it even worse for bicaridine premedding.)

Al-1ce:
It looks like the only downside of Capronine is pain. If that's the case, then it should drain painkilling drugs, like oxycodone or tramadol. Otherwise it's only downside is easily negated, and capro+tric+oxyc will become the premed meta.

RikuTheKiller:
> It looks like the only downside of Capronine is pain. If that's the case, then it should drain painkilling drugs, like oxycodone or tramadol. Otherwise it's only downside is easily negated, and capro+tric+oxyc will become the premed meta.
> 
> edit: Storing the damage isn't that important for the short-term, which is the only instance I can imagine this drug being used.

It IS important for the short term though. Since the healing is intentionally rapid (20 total health per tick, in other words, 10 health/s.) it means you don't get fractures or delimbs and such as often. Also, the aftereffect isn't direct pain, it's burn damage. Burn damage causes extreme pain and at around 100 damage (When a single dose of capronine will usually purge itself.) it'll cause enough pain to the point where no painkilling meds will have any effect. Tricordazine also won't have an effect when using this since any brute or burn it would've healed is stored until the capronine is purged, meaning it can't heal the damage during your fight. (Lessens the effectiveness of other premeds.)

It's also useful for long term applications such as actually getting to medical help. It won't help you if you're about to go critical from shrapnel, broken bones, bloodloss, oxygen damage, toxin damage, cloneloss and the list goes on.

kiVts:
You sir invented a thing called painkiller, maybe plus

RikuTheKiller:
> You sir invented a thing called painkiller, maybe plus

Yes, but also no.

TiviPlus:
Needed changes were outlined on discord convo a while back

RikuTheKiller:
Just procrastinating. Maybe copying Tivi a bit.

PR #10873, opened 2022-09-02
Rockets use bullet armor for direct impact damage

Lumipharon:
## About The Pull Request
What it says on the tin.
Currently, all rockets (except thermobaric) use bomb armor when calculating the damage done from the actual impact. This means they are doing far more damage than they should, since almost everything has lower bomb armour than bullet armour.

Many rockets already have high pen, up to 100, so no change will be seen in game, but for rockets that have lower pen like RR and some TAT shells, they'll do less damage to most xenos (and to marines you manage to hit in the back of the head).

Note: This is a bug fix for TAT, this was always supposed to use bullet armor.

## Why It's Good For The Game
Non explosion damage uses non explosion armor

## Changelog
:cl:
balance: All rocket type ammo now uses bullet armor when calculating direct impact damage
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1554,7 +1554,7 @@ datum/ammo/bullet/revolver/tp44
 	ping = null //no bounce off.
 	sound_bounce	= "rocket_bounce"
 	flags_ammo_behavior = AMMO_EXPLOSIVE|AMMO_ROCKET|AMMO_SUNDERING
-	armor_type = "bomb"
+	armor_type = BULLET
 	damage_falloff = 0
 	shell_speed = 2
 	accuracy = 40
@@ -1680,7 +1680,6 @@ datum/ammo/bullet/revolver/tp44
 	hud_state = "shell_he"
 	hud_state_empty = "shell_empty"
 	flags_ammo_behavior = AMMO_EXPLOSIVE|AMMO_ROCKET|AMMO_SUNDERING
-	armor_type = "bomb"
 	damage_falloff = 0
 	shell_speed = 2
 	accurate_range = 20


BraveMole (changes requested):
"Currently, all rockets (except thermobaric) use bomb armor when calculating the damage done from the actual impact. This means they are doing far more damage than they should"

Considering they are rockets, I do not understand why you think this is unintended.

This is in no way a fix and is just a balance change, and as such lacks proper "why is it good with the game"

Arctrooper8:
This is the problem with rockets in HvH in the first place. They're a weapon intended to kill tanky xenos who have X-ray and thus will know about it before they properly engage. It's not a weapon that fits well into HvH where you aren't able to see what your enemy has until they use it on you the second you walk into their vision range.

Worst of all, wouldn't this affect how rocket weaponry works in HvX? I thought the idea was to not let HvH balance affect HvX balance.

I mean why is a rocket launcher being used in an anti-personnel situation anyway? It makes sense against monstrous aliens, but sounds ridiculous in a battlefield with not a single large target.

Overall the rocket launcher should just be removed from HvH. Only the marines have it, the high burst damage (even with new rounds) isn't fun, and you keep trying to change it so as to force it to fit into HvH when it should just stay as is in HvX.

Lumipharon:
> This is the problem with rockets in HvH in the first place. They're a weapon intended to kill tanky xenos who have X-ray and thus will know about it before they properly engage. It's not a weapon that fits well into HvH where you aren't able to see what your enemy has until they use it on you the second you walk into their vision range.
> 
> Worst of all, wouldn't this affect how rocket weaponry works in HvX? I thought the idea was to not let HvH balance affect HvX balance.
> 
> I mean why is a rocket launcher being used in an anti-personnel situation anyway? It makes sense against monstrous aliens, but sounds ridiculous in a battlefield with not a single large target.
> 
> Overall the rocket launcher should just be removed from HvH. Only the marines have it, the high burst damage (even with new rounds) isn't fun, and you keep trying to change it so as to force it to fit into HvH when it should just stay as is in HvX.

This is a HvX PR, nothing to do with combat patrol, I already made a new ammo type for that.

I noticed this issue when I was testing tat pen bug, and saw that queen was taking the ful 200 damage from a single shot even with their high armour. MJP said it had always meant to be bullet armour, and I assume the same logically applies to other types of rockets.

Arctrooper8:
> This is a HvX PR, nothing to do with combat patrol, I already made a new ammo type for that.
> 
> I noticed this issue when I was testing tat pen bug, and saw that queen was taking the ful 200 damage from a single shot even with their high armour. MJP said it had always meant to be bullet armour, and I assume the same logically applies to other types of rockets.

Then I feel bravemole's comment really sums it up. Changing rocket to use bullet armor doesn't make sense. If rockets are doing too much damage they should be tweaked in some way, not changed to an entirely different armor type.

Bullet armor should stay for more standard marine weaponry, adding direct rocket hits to bullet armor would likely causes balance issues due to the disparity between a rifle and a rocket launcher.

BraveMole:
No why it is good for the game, closing

PR #10874, opened 2022-09-03
V-21 tweak

Lumipharon:
## About The Pull Request
Increases burst scatter (so only effects the higher rate of fire mode), and decreases wielded accuracy.

The V-21 has worse accuracy and scatter on fuller auto, but it was still a bit too effective, especially at longer range. It will still be extremely effective, you just won't be able to stand near the edge of the screen and still land almost everything.

## Why It's Good For The Game
Balance the bullet hose

## Changelog
:cl:
balance: V-21 wielded accuracy reduced slightly, and fuller auto scatter increased
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smgs.dm
+++ b/code/modules/projectiles/guns/smgs.dm
@@ -419,7 +419,7 @@
 	aim_slowdown = 0.15
 	wield_delay = 0.4 SECONDS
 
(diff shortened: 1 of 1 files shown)


MereWalker:
How dare you. Not my space gun. How dare you nerf the vector. Should make it not a insta win pick.

PR #10885, opened 2022-09-04
Ravager ravage can no longer be triggered if there is no one to ravage.

MLGTASTICa:
## About The Pull Request
Ravager's ravage ability now won't trigger if there is nobody to ravage infront

## Why It's Good For The Game
Considering ravage's high speed,along with shotgun knockback and other variables , this will increase their lethality in general and make playing ravager less of a pain for new players. Most xeno melle abilities cannot miss in general , so this should be no exception
(It also brings players more in line with the lethality of AI ravagers , which wreck because.. they cannot miss)

## Changelog
:cl:
qol: Ravager's ravage cannot be used if there is no one to ravage
code: Added the get_facing_direction proc for atoms , returns the DIR that would be set on the src if face_atom was called.
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -975,6 +975,35 @@
 		else
 			setDir(WEST)
 
+// This is face_atom, but instead of changing the dir it just returns the dir the original proc would
+// used for checking ravager's ability to ravage without actually forcing them to face the atom.
+/atom/movable/proc/get_facing_direction(atom/A)
+	if(!A || !x || !y || !A.x || !A.y)
+		return
+	var/dx = A.x - x
+	var/dy = A.y - y
+	if(!dx && !dy) // Wall items are graphically shifted but on the floor
+		if(A.pixel_y > 16)
+			return NORTH
+		else if(A.pixel_y < -16)
+			return SOUTH
+		else if(A.pixel_x > 16)
+			return EAST
+		else if(A.pixel_x < -16)
+			return WEST
+		return
+
+	if(abs(dx) < abs(dy))
+		if(dy > 0)
+			return NORTH
+		else
+			return SOUTH
+	else
+		if(dx > 0)
+			return EAST
+		else
+			return WEST
+
 /mob/face_atom(atom/A)
 	if(buckled || stat != CONSCIOUS)
 		return
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
@@ -101,14 +101,18 @@
 	if(!..())
 		return FALSE
 	var/mob/living/carbon/xenomorph/ravager/X = owner
+	var/facing_dir = X.get_facing_direction(A) // this prevents false-positives , since the actual ravage forces you to face the clicked atom(which can miss if you click behind you while having people infront)
 	var/sweep_range = 1
 	var/list/L = orange(sweep_range, X) // Not actually the fruit
 	var/target_facing
 	for(var/mob/living/carbon/human/H in L)
 		target_facing = get_dir(X, H)
-		if(target_facing != X.dir && target_facing != turn(X.dir,45) && target_facing != turn(X.dir,-45) ) //Have to be actually facing the target
-			return TRUE // just need ONE to be facing
-	to_chat(X, span_notice("There is nobody to ravage infront of you!"))
+		if(H.stat == DEAD || isnestedhost(H))
+			continue
+		if(target_facing != facing_dir && target_facing != turn(facing_dir,45) && target_facing != turn(facing_dir,-45) ) //Have to be actually facing the target
+			continue
+		return TRUE
+	X.balloon_alert(X, "There is nobody to ravage infront of you!")
 	return FALSE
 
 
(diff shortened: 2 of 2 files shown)


Naaanii:
There should honestly be a degree of skill required to use this ability and shouldn't work only if a player is in range which makes it spammable without any chance of missing. Though I'm unsure if warrior's ghost lunge is a bug or intended for it to go on cd.

Should honestly be fine if it were a single target ability.

MLGTASTICa:
> There should honestly be a degree of skill required to use this ability and shouldn't work only if a player is in range which makes it spammable without any chance of missing. Though I'm unsure if warrior's ghost lunge is a bug or intended for it to go on cd.
> 
> Should honestly be fine if it were a single target ability.

It just artificially inflates the skill cap for  ravager.

EvilMelonTheSecond:
> > There should honestly be a degree of skill required to use this ability and shouldn't work only if a player is in range which makes it spammable without any chance of missing. Though I'm unsure if warrior's ghost lunge is a bug or intended for it to go on cd.
> > Should honestly be fine if it were a single target ability.
> 
> It just artificially inflates the skill cap for ravager.

I don't see anything artificially difficult about an ability that requires positioning and timing much like many other abilities in the game

MLGTASTICa:
> > > There should honestly be a degree of skill required to use this ability and shouldn't work only if a player is in range which makes it spammable without any chance of missing. Though I'm unsure if warrior's ghost lunge is a bug or intended for it to go on cd.
> > > Should honestly be fine if it were a single target ability.
> > 
> > 
> > It just artificially inflates the skill cap for ravager.
> 
> I don't see anything artificially difficult about an ability that requires positioning and timing much like many other abilities in the game

it depends on mob facing , which is very easy to miss , especially when a marine gets pushed by said ravager.

Nirvan4:
Sounds good to try, makes easier to not die like a potato and make Rav better at fighting highly mobile marines.

As someone who plays with 250 300 and higher ping this PR makes rav way more appealing too

usnpeepoo:
Could at least give it a penalty for missing, instead of allowing you to spam use the ability until it will hit.

MiniMeatwad:
There is no need to punish players for having shit ping, they already struggle with even moving to the specific position they need to be, the 3 tiles they have to align in a minuscule time window is already hard enough when you're not lagging.

KrabSpider:
> Could at least give it a penalty for missing, instead of allowing you to spam use the ability until it will hit.

I'm going to play Devil's Advocate and argue that losing 200 plasma as a caste that is required to deal damage to regen it is a decent enough penalty. 
I can see it being oppressive against solo marines, but honestly that isn't necessarily much of a point as just about every assault other caste is, too.

usnpeepoo:
> > Could at least give it a penalty for missing, instead of allowing you to spam use the ability until it will hit.
> 
> I'm going to play Devil's Advocate and argue that losing 200 plasma as a caste that is required to deal damage to regen it is a decent enough penalty. I can see it being oppressive against solo marines, but honestly that isn't necessarily much of a point as just about every assault other caste is, too.

Ok but you won’t be losing 200 plasma if it’s not activating ravage for no one being there. You’re just gonna be spam clicking right click/MC/SC until you are in range of a marine and hit it regardless. Also I’m assuming it would bypass whatever screen shake you achieved because it will still hit/not miss. I believe this is a good change but too safe especially for ravager.

Naaanii:
Making the targetting for ravage better would also be a good idea, but I'm really not sure how that can be done.

Losenis:
"an ability that requires positioning and timing" bro you're talking about timing on something that has less of a window of opportunity than people's average ping. Half the reason pit boss is so lethal is because he plays at like 0 ping while everyone else is fucking around at 300, 80% of actual skill involved in gameplay is living close to the server since it's not like you can pick other servers with better ping.

MLGTASTICa:
I believe its fair for ravager to not miss every ravage ? considering they are a T3 caste that's meant to RIP and TEAR, but that usually needs to retreat the very first second of engagement due to missing their stun on a shotgun user...

Losenis:
Considering I'm one of the sods playing at 200+ ping, yeah I'd absolutely love for that ability to actually have some manner of guarantee that it'll hit after I have to actually move up to the marine. Being a melee range ability, you have to have to time the click so that you're next to the marine by the time the ping registers, which is much harder to do in the middle of combat vs a moving opponent. It's one of the reasons I don't play ravager as often, it's very ping-dependant.

novaepee:
That feeling whem you bump marine then ravage where they were a few millisecond ago.

Glcbitj:
Why are you marking an obviously balance change as "Quality of Life". This also makes playign Rav way easier as you can just spam the funny button with no repercussions at all.
High ping is no excuse to make something a no-brainer, why don't you remove everything that requires any aiming or positioning if you can just say muh high pigh as an excuse?

Losenis:
"High ping is no excuse"

My brother in Ra you're playing a BYOND game with one single server for the whole planet, do you see the average person deflecting GL nades with defender tail swipe

TiviPlus:
We're not interested in what more or less amounts to autoaim to abilities

MLGTASTICa:
Oh the misery.

MiniMeatwad:
Even if you hate quality of life, the biggest thing would just be making it cost rage, a limited resource, for missing. So you can still get punished for missing, but you will still be able to hit someone and not die for having 300 ping.

PR #10887, opened 2022-09-04
You can now respawn into EORD no matter your pref for it

ivanmixo:
## About The Pull Request

- Basically the title, you couldn't respawn into EORD if you didn't have the pref for it due to only being attributed the verb if you changed mob *after* it started. This resulted in you not being able to join if you were dead and had the pref off when it begun.

- Also sanitises so we don't get duplicate verbs, just in case

## Why It's Good For The Game

Sometimes you're in the mood to go into EORG, sometimes you aren't. It's nice to have the option to join in if you feel like it. This essentially makes the pref just a "force pull me into EORG roundend" pref.

## Changelog
:cl:
qol: You can now respawn into EORD no matter your pref for it
/:cl:

### DIFF
--- a/code/datums/gamemodes/game_mode.dm
+++ b/code/datums/gamemodes/game_mode.dm
@@ -271,7 +271,7 @@ GLOBAL_LIST_INIT(bioscan_locations, list(
 
 /datum/game_mode/proc/grant_eord_respawn(datum/dcs, mob/source)
 	SIGNAL_HANDLER
-	source.verbs += /mob/proc/eord_respawn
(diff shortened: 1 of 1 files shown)


RikuTheKiller:
Good job, I often did this anyway if I somehow managed to get a hold of the verb, but I really don't like being pulled out into EORD automatically, especially as a xeno. Xenomorph EORD sucks ass since usually over half the hive doesn't go into it so you have no weed and just get killed. Then you're off to the default loadouts again.

TiviPlus (approved):

PR #10891, opened 2022-09-06
tadpole swagger buff [MDB ignore]

tyeagg:
## About The Pull Request
tadpole looks slightly less like a flying box, doesn't affect size. the cockpit windows are now objects and easier to remove

tadpoledoner

## Why It's Good For The Game
i think it looks nicer

## Changelog
:cl:
add: tadpole remap
balance: you can now fire over tadpole window and easier to remove
/:cl:


FosterGrime-Heymynameslouie:
tye I love you

DerpiestBruh:
The exaggerated swagger of a PO main

mc-oofert:
this just makes it harder to fire fixed guns and it looks worse than the chadpole we already have

AzonStarfall:
You may want to update the sprites at the front of the cockpit that look like they should be angled, it somewhat clashes against the rest of the wall design.

Pariah919 (commented):
360 firing of the team weapon mounted on the tad seems kinda strong?

DerpiestBruh:
no its not

tyeagg:
> 360 firing of the team weapon mounted on the tad seems kinda strong?

every single person literally removes the windows on tadpole this is basically a reason to sort of keep the windows

Lumipharon:
Making the windows throwpass is kinda weird because there is no reason to think they are  from their appearance. They also basically the same as alamo cockpit windows which aren't throwpass.

BraveMole (changes requested):
I agree with lumi, it really doesnt look like you can fire over these windows.

Either update the sprite to something believable (if that is even possible) or do not change the behaviour

Weedmaster77:
Just make the cockpit window like the alamo where you cant throwpass unless you cut them out and replace them like what normally happens with tad. An maybe just make the cockpit part have the same health as the windows with shutter down in the cockpit so the only real buff its sight but with shutter health.

Al-1ce:
>windows

Ever heard of an embrasure? Do that instead. Much more plausable to shoot thru. #

zzurge:
chadpole man does it again,zased

Lumipharon:
It would be nice to get the tiles on either side of the console to be corner tiles, at the moment there's no connection to the outer corner tile and it just kinda stops. Looking great over all though.

GravitronBlue:
> If this is just an artistic change, it's based.

Sadly it's not.

- Back engines that you could once shoot over are gone
- Window shutters don't exist
- Windows are now gunports

Though I like the appearance change this PR provides, my preference is still towards the original mechanics of the old tadpole.

FosterGrime-Heymynameslouie:
apparently this causes this 
https://github.com/tgstation/TerraGov-Marine-Corps/issues/11015#issue-1380194572

PR #10893, opened 2022-09-06
Adds a buffer to delimbs (Alternate)

RikuTheKiller:
## About The Pull Request

Simply adds a flat 25 (Value may change.) to the max damage of a limb when it's created and then uses it when determining whether or not to sever the limb. The severing threshold is max damage minus the buffer.

Also removes the organ_health_multiplier config option. (Bravemole told me to, not my idea.)

This prevents burn damage from being abused to avoid delimbs, but uses a different method that isn't hardcoded into the max damage values themselves and is a flat value instead of a percentage so it doesn't cause any differences in the sever resistance of limbs, which the other PR does.

The amount can be changed later since it's not deeply hardcoded. (DELIMB_BUFFER in limbs.dm)

## Why It's Good For The Game

Bug fix good? Is this a feature or a bugfix? Eh, whatever, not like it matters.

closes #10849

## Changelog
:cl:
add: Added a threshold to delimbs
del: Removed the organ health multiplier config option
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -269,7 +269,7 @@
 			owner.updatehealth()
 		return update_icon()
 	if(CONFIG_GET(flag/limbs_can_break) && brute_dam >= (max_damage - DELIMB_BUFFER) * CONFIG_GET(number/organ_health_multiplier))
-		droplimb() //Reached max damage threshold through brute damage, that limb is going bye bye
+		droplimb() //Reached limb severing threshold, that limb is going bye bye
 		if(!(owner.species && (owner.species.species_flags & NO_PAIN)))
 			owner.emote("scream")
 		return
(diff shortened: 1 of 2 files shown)


code/modules/organs/limbs.dm
```diff
@@ -83,6 +86,7 @@
 		RegisterSignal(owner, COMSIG_PARENT_QDELETING, .proc/clean_owner)
 	soft_armor = getArmor()
 	hard_armor = getArmor()
+	max_damage += DELIMB_BUFFER
```
BraveMole, line 89: no do it in code manually. 

There is no reason to do something at runtime if you can do it in compilation

BraveMole (changes requested):

code/modules/organs/limbs.dm
RikuTheKiller: Did exactly that.

RikuTheKiller:
Closed due to the other one being merged.

This is an alternate PR, after all.

PR #10903, opened 2022-09-08
Wraith can't banish weeds

QualityVan:
## About The Pull Request
So for a while, wraith couldn't banish weeds/resin because the ability checked if something was an obj/effect to see if it was immune to banish. #10626 repathed a bunch of stuff out of obj/effect, which made weeds viable target. This fixes that with a check for /obj/alien.
Included in this are weeds+nodes, resin, eggs, and hivemind cores. 

## Why It's Good For The Game
Fixes burning your ability on the ground by accident.

## Changelog
:cl:
fix: Wraith can't banish weeds/resin
/:cl:


QualityVan:
This does not disable banishing walls.

PR #10905, opened 2022-09-08
Condensing seasonals

maikilangiolo:
## About The Pull Request
Very simply, condenses the seasonals to that they have 4 guns each, in order to both increase player choice during a single seasonal, since they can play with four toys, and make seasonals cycle a bit faster, since you cant always play TGMC all-day every day and missing that one seasonal you like sucks. Then again, the SOM seasonals have 4 guns which gave me the idea.
Not just that, but this PR adds the SO seasonals to the Valhalla seasonal vendor, and the old V10 shotgun to the shotgun seasonal. huzzah

## Why It's Good For The Game
Just a simple change to grant more choice per seasonal and faster rotation. Descriptions changed to match the new contents.

## Changelog
:cl:
qol: SOM seasonals are now in the seasonal vendor
qol: seasonals condensed into 4 guns for more choice and faster rotation
add: the old V10 shotgun made his way into the seasonals
/:cl:


Weedmaster77:
Like the idea hope it gets pushed!

Al-1ce:
good PR

PR #10910, opened 2022-09-09
Add Carrier ability: Drop all facehugger panic button

Wayland-Smithy:
### New carrier ability: Drop All Facehuggers
**Description**: Drops all stored facehuggers just like when a carrier dies
**Cooldown**: 50 seconds (same as shrike screech)
**Plasma Cost**: All stored plasma, which also halts self emitted pheros

Action button sprite provided by @novaepee

## Why It's Good For The Game

Xeno content

## Changelog
:cl: Wayland-Smithy & novaepee 
add: Added new carrier ability: Drop all facehuggers
fix: Fixed useless loop condition in hugger drop on death code and moved it into the action
/:cl:


mc-oofert:
carriers on their way to rush and drop 6 larval huggers into the ungaball instead of using it as a panic ability

Wayland-Smithy:
Whether they rush or are rushed I think both situations call for a bit of panic, and of course the cooldown can always be adjusted if the risk of using it recklessly is not harsh enough.

novaepee:
> carriers on their way to rush and drop 6 larval huggers into the ungaball instead of using it as a panic ability

Then they pray they don't die from the burst that is marines.

code/modules/mob/living/carbon/xenomorph/castes/carrier/carrier.dm
```diff
@@ -17,24 +17,23 @@
 		/mob/living/carbon/xenomorph/proc/vent_crawl,
 	)
 
-// ***************************************
-// *********** Death
-// ***************************************
-/mob/living/carbon/xenomorph/carrier/on_death()
-	. = ..()
+/mob/living/carbon/xenomorph/carrier/proc/try_drop_all_huggers()
```
BraveMole, line 20: Code should be on an action IMO. And you can call it on death with a signal

BraveMole (changes requested):

Losenis:
Me and the boys about to get instawiped by 4 carriers that dropped all clawed huggers after the defiler vented in

Strayed-Raven:
It sounds abusable in defiler smoke. But Carriers are squishy as heck anyways. Still this would make SWAT masks more useful IIRC, since I am pretty sure it's one on hugger going after an individual marine yes? If it is too strong maybe make huggers die faster just sitting out in the open but not from an egg or resin hole? Maybe marines will actually start using smoke grenades more.

code/datums/keybinding/xeno.dm
```diff
@@ -191,6 +191,12 @@
 	description = "Prompt a wheel to choose which hugger you will deploy with the Throw Hugger ability."
 	keybind_signal = COMSIG_XENOABILITY_CHOOSE_HUGGER
 
+/datum/keybinding/xeno/drop_all_hugger
+	name = "drop_all_hugger"
+	full_name = "Carrier: Drop All Facehugger"
```
XSlayer300, line 196:
```suggestion
	full_name = "Carrier: Drop All Facehuggers"
```

code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
```diff
@@ -177,6 +177,34 @@ GLOBAL_LIST_INIT(hugger_images_list,  list(
 	succeed_activate()
 	add_cooldown()
 
+// ***************************************
+// *********** Drop all hugger, panic button
+// ***************************************
+/datum/action/xeno_action/carrier_panic
+	name = "Drop All Facehugger"
```
XSlayer300, line 184:
```suggestion
	name = "Drop All Facehuggers"
```

code/datums/keybinding/xeno.dm
Wayland-Smithy: uhh sure I guess since it's the only ability that can do multiple at once

Wayland-Smithy:
> Me and the boys about to get instawiped by 4 carriers that dropped all clawed huggers after the defiler vented in

Reading this reminded be of the Swarm Host unit from SC2 where it's very scary once in a while especially in groups, but the other 70% of their life is just taking up supply while they reload that could have been spent on a better rounded less situational unit capable of sustained fighting. I like it.

code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
```diff
@@ -189,6 +189,15 @@ GLOBAL_LIST_INIT(hugger_images_list,  list(
 	keybind_signal = COMSIG_XENOABILITY_DROP_ALL_HUGGER
 	use_state_flags = XACT_USE_LYING
 
+/datum/action/xeno_action/carrier_panic/give_action(mob/living/L)
+	. = ..()
+	RegisterSignal(owner, COMSIG_MOB_DEATH, .proc/action_activate)
```
BraveMole, line 194: if it's a signal then it needs to have SIGNAL_HANDLER, and I really doubt it has it or should have it

BraveMole (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
Wayland-Smithy: Can't because you are not able to `SIGNAL_HANDLER` an override which `action_activate` is.. so should I add it to the parent proc? Keep in mind this was your idea.

code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
```diff
@@ -177,6 +177,51 @@ GLOBAL_LIST_INIT(hugger_images_list,  list(
...
+/datum/action/xeno_action/carrier_panic/action_activate()
+	var/mob/living/carbon/xenomorph/carrier/X = owner
+
+	if(!X.huggers)
+		return
+
+	X.visible_message(span_xenowarning("A chittering mass of tiny aliens is trying to escape [X]!"))
+	while(X.huggers > 0)
+		var/obj/item/clothing/mask/facehugger/F = new X.selected_hugger_type(get_turf(X))
+		step_away(F,X,1)
```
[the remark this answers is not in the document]
Wayland-Smithy: It is random yes

BraveMole (approved):

PR #10911, opened 2022-09-09
Expand Pillar of Spring's AI core slightly

lbnesquik:
## About The Pull Request
image

## Why It's Good For The Game
The AI core on POS is very small, and being against the wall to the right weakens in a lot. 
This makes the AI not look as shoved into a storage closet, and it makes the core decently stronger. 

## Changelog
:cl:
add: Expanded the AI core on Pillar of Spring
/:cl:

### DIFF
--- a/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
+++ b/_maps/map_files/Pillar_of_Spring/TGS_Pillar_of_Spring.dmm
@@ -3164,8 +3164,6 @@
 /turf/open/floor/wood,
 /area/mainship/command/corporateliaison)
 "iX" = (
-/obj/structure/window/reinforced/extratoughened,
-/obj/machinery/camera/autoname/mainship,
 /obj/vehicle/unmanned/droid,
 /turf/open/floor/mainship/tcomms,
 /area/mainship/command/airoom)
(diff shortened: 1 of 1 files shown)


RipGrayson (changes requested):
While I'm not against giving the AI more space, I don't like the one tile corridor here.
miles

As a general rule we don't do one tile corridors except in maintenance areas, they tend towards very binary positions where one side effortlessly controls the area by virtue of limited space to move and dodge.

RipGrayson:
> Any idea on how to change this? Besides encroaching on the armory, or moving it somewhere else?

There's very little excess space in PoS CIC, I don't see how you could enlarge the AI area without cramping something else. My gut instinct is that the AI should have its own little area near the CIC, probably made out of repurposed flavor rooms that nobody uses anyway.

MereWalker:
Place it in the staff quarters. Wait.. that got taken out. Uhh maybe put it in atmos?

Losenis:
This is actually a good change because in hijack you could just acid the wall and kill the AI very fast, since there was no glass on the wall side

QualityVan:
What if you included the left hall in that image as part of the core and cut the center wall/door?

MereWalker:
You could put it in engineering.

PR #10919, opened 2022-09-11
Removes BAS from rocket launchers

QualityVan:
## About The Pull Request
Removes BAS being attachable to SADAR, RR, and quad themobaric. Also the RL-5 but I don't know if that's actually acquirable ingame. Random weapon spawns, maybe.

## Why It's Good For The Game
autoaim oneshot bad

## Changelog
:cl:
balance: Rocket launchers can't take build-a-sentry as an attachment.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -547,7 +547,6 @@ Note that this means that snipers will have a slowdown of 3, due to the scope
 	attachable_allowed = list(
 		/obj/item/attachable/magnetic_harness,
 		/obj/item/attachable/scope/mini,
-		/obj/item/attachable/buildasentry,
 		/obj/item/attachable/shoulder_mount,
 	)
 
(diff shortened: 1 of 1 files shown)


BraveMole (changes requested):
Not a fan.

If you really want to nerf those, you can add inaccuracy (see laser turrets)

QualityVan:
I just don't think it's reasonable to say most xenos should never move around the map below 80% health because they might walk into turret range and die immediately. A human at least usually has a light source and requires a player to wait to ambush.
This was opened at MJP's suggestion, whatever that's worth.
An alternative I've considered is a target lock time equal to the weapon's base wield time, but that always seemed like it'd cause too much knock-on stuff.
I'm not wild about just making it a diceroll, since that seems fundamentally frustrating for both whoever set the turret and whoever gets shot.

BraveMole (changes requested):
Just as a reminder, we discussed about it and I think we agreed additional target locking/windup was a good solution

QualityVan:
ye

PR #10931, opened 2022-09-12
Reduces the flash range of rockets and RR rockets.

MLGTASTICa:
## About The Pull Request
Reduces the flash range of the following rockets
Default rocket subtype , from 5 to 2
recoilless rockets default subtype , from 5 to 2 
light recoiless rocket , from 5 to 1
recoilless low impact from 3 to 2

## Why It's Good For The Game
The RR launcher has been seeing no use, we have already added a lot of telegraphing whenever a marine fires a rocket , so i believe it is fair to let marines actually see the xeno they want to kill.
This makes it so getting hit by an RR  while fighting close-up marines is far more deadly, and far more of a risk.
(And actually lets marines score a kill whenever they do manage to hit a xeno)
## Changelog
:cl:
balance: Reduced the default rocket subtype flash range from 5 to 2
balance: Reduced RR rocket flash range from 5 to 2
balance: Reduced the LRR rocket flash range from 5 to 1
balance:Reduced the low impact RR rocket flash range from 3 to 2
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1569,7 +1569,7 @@ datum/ammo/bullet/revolver/tp44
 	bullet_color = LIGHT_COLOR_FIRE
 
 /datum/ammo/rocket/drop_nade(turf/T)
-	explosion(T, 0, 4, 6, 5)
+	explosion(T, 0, 4, 6, 2)
 
 /datum/ammo/rocket/on_hit_mob(mob/M, obj/projectile/P)
 	drop_nade(get_turf(M))
@@ -1693,7 +1693,7 @@ datum/ammo/bullet/revolver/tp44
 	sundering = 50
 
 /datum/ammo/rocket/recoilless/drop_nade(turf/T)
-	explosion(T, 0, 3, 4, 5)
+	explosion(T, 0, 3, 4, 2)
 
 /datum/ammo/rocket/recoilless/heat
 	name = "HEAT shell"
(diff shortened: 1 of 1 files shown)


Lumipharon:
This is a straight nerf to rockets though? Blind your victims is really strong when combo'd with the random direction knockback the explosion causes. If you're getting blinded, you're already blowing yourself up.

BraveMole (approved):
Seems like a reasonable idea.

@Pariah919 what do you think ?

MLGTASTICa:
> This is a straight nerf to rockets though? Blind your victims is really strong when combo'd with the random direction knockback the explosion causes. If you're getting blinded, you're already blowing yourself up.

this is a change meant for XvH , where xenos do not get affected by blindness, and every marine in a 5 tile radius would.

MereWalker:
Personally I like this. With sadar being with out a scope that means it would flash and normally hurt you. Now it won't flash but it will still hurt you half the time. The RR gets a nice buff. Now I can shoot and not get blinded as much. Look forward to it! If I might add into this when a rocket or sadar is shouldered could be we get emote or auto say "rocket up ""back blast clear" to get marines to stay the hell out of the line of fire. Or make it a emote. Just a thought.

RyderFive:
> This is a straight nerf to rockets though? Blind your victims is really strong when combo'd with the random direction knockback the explosion causes. If you're getting blinded, you're already blowing yourself up.

 RPG flash is a poor CM holdover (if I remember correctly) that was designed to punish marines for being too close with a RPG. In HvX it mostly just punishes marines, and its user with blindness for being within view distance of a RPG. This usually just lets the xeno you shot get away since your blinded. The flash is only going to be used to actual effect on hostile humanoid targets from a ERT which is already not really common every round. The flash should just be removed or only happen maybe within 1 tile of a direct hit.

PR #10940, opened 2022-09-13
Fix naming issue with unboxer

lbnesquik:
## About The Pull Request
The pamphlet says outputter but the current name is Unboxer. 

## Why It's Good For The Game
Less new RO confusion

## Changelog
:cl:
fix: fixed the req factory saying Outputter instead of Unboxer
/:cl:


AzonStarfall:
Realistically it'd probably be better to standardize this entirely by removing Outputter from the typepaths as well, but this works too.

TiviPlus (approved):

TiviPlus (approved):

PR #10948, opened 2022-09-13
Standardizes lasgun underbarrels

Jalopkomit18:
## About The Pull Request
As per title, standardizes the lasgun underbarrel weapons. Before the lasguns were inconsistent on which underbarrel weapons were able to be attached. For example, the lasrifle could only attach the miniflamer and pepperball, but the carbine could only attach the grenade launcher and pepperball. I don't understand why only some can be attached, and with seemingly arbitrary prerequisites as to which lasgun is able to attach which underbarrel. I understand that this may be a balance concern, so please feel free to discuss. However, it seemed that the standing precedent was that weapons that were able to attach underbarrel weapons were able to attach all of them, not just some of them. This would amend that. 

## Why It's Good For The Game
I feel that above explains it best.

## Changelog
:cl:
balance: lasguns get access to all underbarrel weapons instead of just some.
/:cl:


Starfirejordan:
time to put a underbarrel shotgun on my underbarrel lasgun

lKiseki:
Bad idea. Certain weapons cannot equip certain attachments due to balance decisions.

Jalopkomit18:
> Bad idea. Certain weapons cannot equip certain attachments due to balance decisions.

Is this due to carbine having a shotgun mode? Given that lasguns are sidegrades to their ballistic counterparts, I feel that the other weapons shouldn't be an issue. Las-sniper also may be an issue if it got all of the underbarrels, but I wanted to see the consensus on that first.

Lumipharon:
carbine plus plaspistol is a lot of burst damage. No reason why laser rifle doesn't have a ugl though

BraveMole:
As already discussed on discord, guns need to be unique; consistency is exatcly what we do not want, so changing the balance with this argument doesn't work

PR #10956, opened 2022-09-14
why does ths thing have opacity in the first place remove opacity from explosion

tyeagg:
## About The Pull Request
remove opacity from explosion particles

## Why It's Good For The Game
no more random particles that block your vision when an explosion happens

## Changelog
:cl:
del: removed opacity from explosion particles
/:cl:


Losenis:
Most of the time it's noticeable it's when a debris bugs out and stays floating in the air blocking sight rather than disappearing in the like 2 seconds at most that it flies through the screen. This wouldn't really change smoke from explosions and explosions themselves aren't opaque anyway.

PR #10957, opened 2022-09-14
(Re?)adds move_force

TiviPlus:
## About The Pull Request
We had no way to give push priority (Which mechs need) because for some reason we only have a half-implemented move_resist

Idk why the safe throw was commented out it's only used for shuttles so whats the worst

## Changelog
:cl:
code: Added push priorities
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -794,9 +794,9 @@
 		AM.onTransitZ(old_z,new_z)
 
 
-/atom/movable/proc/safe_throw_at(atom/target, range, speed, mob/thrower, spin = TRUE)
-	//if((force < (move_resist * MOVE_FORCE_THROW_RATIO)) || (move_resist == INFINITY))
-	//	return
+/atom/movable/proc/safe_throw_at(atom/target, range, speed, mob/thrower, spin = TRUE, force = MOVE_FORCE_STRONG)
+	if((force < (move_resist * MOVE_FORCE_THROW_RATIO)) || (move_resist == INFINITY))
+		return
 	return throw_at(target, range, speed, thrower, spin)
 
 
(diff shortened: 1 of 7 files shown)


BraveMole (changes requested):
I do not want to have two push/no push systems, make that consistent with xeno/marine push please

code/game/atoms_movable.dm
```diff
@@ -1040,6 +1042,25 @@
 		return
 	DISABLE_BITFIELD(flags_pass, HOVERING)
 
+///returns bool for if we want to get forcepushed
+/atom/movable/proc/force_pushed(atom/movable/pusher, force = MOVE_FORCE_DEFAULT, direction)
```
BraveMole, line 1046: Why is this a proc

code/game/atoms_movable.dm
```diff
@@ -1040,6 +1042,25 @@
...
+		visible_message(span_warning("[src] forcefully pushes against [pushed_atom]!"), span_warning("You forcefully push against [pushed_atom]!"))
+
+///returns bool for if we want to get handle move crushing, return is bool if we can move an anchored obj
+/atom/movable/proc/move_crush(atom/movable/crushed_atom, force = move_force, direction, silent = FALSE)
+	. = crushed_atom.move_crushed(src, force, direction)
+	if(!silent && .)
+		visible_message(span_danger("[src] crushes past [crushed_atom]!"), span_danger("You crush [crushed_atom]!"))
+
+///returns bool for if we want to get crushed
+/atom/movable/proc/move_crushed(atom/movable/pusher, force = MOVE_FORCE_DEFAULT, direction)
```
BraveMole, line 1062: Ditto

BraveMole (changes requested):

code/game/atoms_movable.dm
TiviPlus: TG port its so damage and etc can be handled as needed

PR #10962, opened 2022-09-14
Anti-personnel landmines no longer trigger on dead mobs

MLGTASTICa:
## About The Pull Request
Any mob thats dead can no longer trigger landmines

## Why It's Good For The Game
The amount of times i've seen people blow themselves trying to sell dead xenos and walking through a mind-field is unreal.
I don't see why we should keep this either , it only serves to make landmines even worse(like they aren't already quite niche and weak)

## Changelog
:cl:
balance: Anti-personnel landmines no longer get triggered by dead mobs
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/mine.dm
+++ b/code/game/objects/items/explosives/mine.dm
@@ -203,8 +203,8 @@ Stepping directly on the mine will also blow it up
 
 	if(linked_mine && isliving(A))
 		var/mob/living/unlucky_person = A
-		// Don't trigger for dead xenos
-		if(isxeno(unlucky_person) && unlucky_person.stat == DEAD)
+		// Don't trigger for dead people
+		if(unlucky_person.stat == DEAD)
 			return
 		linked_mine.trip_mine(A)
 
--- a/code/game/objects/items/explosives/mine.dm
+++ b/code/game/objects/items/explosives/mine.dm
@@ -204,7 +204,7 @@ Stepping directly on the mine will also blow it up
 	if(linked_mine && isliving(A))
 		var/mob/living/unlucky_person = A
 		// Don't trigger for dead xenos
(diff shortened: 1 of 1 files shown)


QualityVan:
Better to just not trigger on dead things in general, no?

Naaanii:
No longer will people dragging corpses destroy the FOB.

code/game/objects/items/explosives/mine.dm
```diff
@@ -202,6 +202,10 @@ Stepping directly on the mine will also blow it up
 		return
 
 	if(linked_mine && isliving(A))
+		var/mob/living/unlucky_person = A
+		// Don't trigger for dead xenos
+		if(isxeno(unlucky_person) && unlucky_person.stat == DEAD)
```
BraveMole, line 207: why the isxeno check ?

BraveMole (changes requested):

code/game/objects/items/explosives/mine.dm
MLGTASTICa: i don't wannna affect HvH balance where you can throw people at landmines to detonate them ? or should it also not trigger for anything dead in general?

MLGTASTICa: made it not trigger for anything dead.

BraveMole (approved):

QualityVan:
If I was being finicky I'd move the check up to inside trip_mine since that's where the rest of the triggering mob's conditions are checked, but that's not super important.

Lumipharon:
Boo, I will  no longer be able to throw dead marines into minefields to blow them up in combat patrol.

Probs good change though.

Al-1ce:
hint for all you aspiring robustos out there: you can disable these with a multitool instead of being a murderous psychopath

PR #10964, opened 2022-09-14
Ais now user timer like ss

BraveMole:
## About The Pull Request

Two effects: 
- should vastly reduce the lag from minions. Especially when you are at 50+ minions, it will be noticeable. I had 200 minions running around with zero td
- The speed of the minion is now exactly what is said on their castedatum; before it was always a bit lower due to the granularity of minions. They also move more naturally (rather than all at the same time)

## Changelog
:cl:
balance: Minions moves faster
code: Minions movement are way less laggy
/:cl:


Lumipharon:
timer defs man. Neat

PR #10972, opened 2022-09-15
Xenos no longer have a respawn delay in crash.

Naaanii:
## About The Pull Request

There is no longer a respawn delay when dying as Xeno in crash.

## Why It's Good For The Game

Playing Xeno in crash is suffering and dying is already punishable enough as you lose maturity and have to re age and re evolve. This buffs T1 spam slightly but marines can deal with most T1s easily with the wider Xeno:Marine ratio.

## Changelog
:cl:
balance: Larva refund no longer has a delay in crash.
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -26,7 +26,7 @@
 	// Round start info
 	var/starting_squad = "Alpha"
 	///How long between two larva check
-	var/larva_check_interval = 2 MINUTES
+	var/larva_check_interval = 0
 	///Last time larva balance was checked
 	var/last_larva_check
 	bioscan_interval = 0


lKiseki:
Speaking from experience, playing as a Xenomorph in Crash is certainly an exercise in frustration. The one bonus they've got going for them is how fast they can replenish their numbers, though even that does not make enough of a difference. I'd say this is a good change, as it allows xenos to recover faster and have more holding power.

MathiasOrtega:
I can see why this would be done. But Crash should be more fleshed out either way considering it’s never picked besides lowpop.

Serxule:
dont think this would be balanced for actual lowpop crash but itd definitely help on the "max 10 xenos vs 30 marines" crash rounds

BraveMole (changes requested):
That' s ultra shitcode. Come on, at least try to understand where is this larva check interval is used.

BraveMole:
Actually this is absolutly not doing what it says it does

PR #10984, opened 2022-09-16
Fix copy paste woopsie

TiviPlus:
lol


MLGTASTICa:
> lol

why not , why can't we have rebel commands ? why can't we question authority at all ?!?!? we got MP's to track down the rebel commander anyway! the TGMC rebel opression ends now.

PR #11003, opened 2022-09-19
Projectile hit proc fixes

Lumipharon:
## About The Pull Request
Changed the explosive ammo flag to trigger do_at_max_range() (and added a turf arg) instead of on_hit_turf().
This specifically fixes TAT APHE ammo exploding correctly on the clicked tile, while avoiding funny multiple explosion bugs again.

Then updated both  do_at_max_range() and on_hit_turf() for all relevant projectiles to check for turf density. If the turf is dense it instead triggers on the projectile's turf (i.e. the turf before the wall or whatever).
This means projectiles that reach their max range will actually trigger on the max range tile, not the tile before it.
Some projectiles specifically want to hit dense turfs, so of course these are excluded.

Currently some projectiles check this, but did the same thing in multiple different ways, while others didn't check at all, meaning you could trigger certain effects like explosions and fireballs through walls, and other funny things.

**This fixes a lot of bugs:**
Shooting a rocket at a wall creates no actual explosion currently as it gets stuck in the solid turf, the same happens for incendiary types
Several acid/neuro spits would similarly not cause their effect as it was triggering on a dense turf.
and various other related bugs

Edit: Refactored how plasma pistol apply their effects to be less sucky.

## Why It's Good For The Game
Bug fixes good
Consistant code good

## Changelog
:cl:
refactor: refactored some projectile on hit code
fix: Certain projectile hit effects will no longer incorrectly trigger on the wrong or dense turfs
/:cl:


code/modules/projectiles/ammo_datums.dm
```diff
@@ -1355,11 +1355,12 @@ datum/ammo/bullet/revolver/tp44
 	drop_nade(get_turf(O))
 
 /datum/ammo/tx54/he/on_hit_turf(turf/T, obj/projectile/P)
-	drop_nade(T)
-
-/datum/ammo/tx54/he/do_at_max_range(obj/projectile/P)
-	drop_nade(get_turf(P))
+	var/target = T.density ? P.loc : T
```
BraveMole, line 1358: target is a useless var

BraveMole (changes requested):

code/modules/projectiles/ammo_datums.dm
Lumipharon: Cut the var out/

code/modules/projectiles/ammo_datums.dm
```diff
@@ -2202,16 +2202,27 @@ datum/ammo/bullet/revolver/tp44
...
-	T.ignite(heat, burn_damage, fire_color)
+	var/turf/target = get_turf(O)
+	if(!target)
+		target = get_turf(proj)
+	target.ignite(heat, burn_damage, fire_color)
+
+/datum/ammo/energy/plasma_pistol/do_at_max_range(turf/T, obj/projectile/proj)
+	var/burn_mod = 1
+	if(istype(T, /turf/closed/wall))
+		burn_mod = 3
```
QualityVan, line 2219: I don't like this duplicating code from pp's turf hit. Better off using a drop_flame

PR #11020, opened 2022-09-21
Shield rebalance

Lumipharon:
## About The Pull Request
So I'm a dumb dumb and didn't realise that the default coverage of shield items against all projectiles is... 100%
i.e. until your shield gets degraded, it absorbs all incoming projectile damage.

Changed it to instead block 60% of bullet, laser and energy damage, and reduced fire block from 80 to 70.
This should have very little impact on HvX but is rather needed for HvH.

## Why It's Good For The Game
Full damage invulnerability is not ideal.

## Changelog
:cl:
balance: Reduced the percentage of damage absorbed by shields for bullet, lasers, energy and fire
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -6,7 +6,7 @@
 	set_shield()
 
 /obj/item/weapon/shield/proc/set_shield()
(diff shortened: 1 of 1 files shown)


BraveMole (approved):

TiviPlus (changes requested):
use defines pls

Lumipharon:
> use defines pls

One day I'll actually remember without you reminding me... maybe.

RikuTheKiller:
A bit too hard a nerf, 70% - 75% at full durability would be more ideal in my opinion.

PR #11028, opened 2022-09-22
Carrier panic drop change

Lumipharon:
## About The Pull Request
Carrier panic drop can now only be used when at 25% of health or under.

Currently panic drop is without question the most powerful ability in the game, and is just used as an insta kill ability, not an 'oh shit' panic ability as intended.

This PR retains its potency, but means you can't be as hilariously aggressive with it.

## Why It's Good For The Game
ggnore button is bad

## Changelog
:cl:
balance: Carrier panic drop can only be used at 25% health or less
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/carrier/abilities_carrier.dm
@@ -207,6 +207,10 @@ GLOBAL_LIST_INIT(hugger_images_list,  list(
 	if(!.)
 		return FALSE
 	var/mob/living/carbon/xenomorph/carrier/X = owner
+	if(X.health > (X.maxHealth * 0.25))
+		if(!silent)
+			to_chat(X, span_xenowarning("We are not injured enough to panic yet!"))
+		return FALSE
 	if(X.huggers < 1)
 		if(!silent)
 			to_chat(X, span_xenowarning("We do not have any young ones to drop!"))


novaepee:
very sad, now i can't do cqc carrier anymore

MathiasOrtega:
No more tripping on an empty resin hole and suddenly a carrier drops all slasher huggers on top of you.

Naaanii:
Lame way of forcing an ability to only be used when near death considering it's not a very durable caste either.

Naaanii:
> Eh, keep it as it is. Huggers are a 1 hit kill anyways. The ability is useless in sieges and mainly only useful for lone marines or healing marines.
> 
> Reminder than using all huggers is quite costly for a carrier, as they lose a crap ton of plasma and time getting them back.
> 
> Oh and also, if you're being chased at 25% health you're 99% likely dead anyways and death does the same thing as panic drop, so like, this practically invalidates it.

If it's really an issue. I'd say up the cooldown, maybe even make it on par with Queen Screech as a panic ability shouldn't be 'spammable', even though it takes time to setup again but keeping eggs ready for a second panic drop isn't exactly hard or rocket science, just takes a bit more prep and carriers are deadliest when they prep right.

You can panic while at full HP too. Gating it behind low health mechanics isn't exactly fun or necessary.

private-tristan:
why not have the number of huggers be based around the % of hp the carrier has left?
75-100 = 2 huggers dropped
50-75 = 5 huggers 
0-50 = all

BraveMole (approved):
Cool idea

RikuTheKiller:
that... was not great

RikuTheKiller:
im still personally on the side of just giving the huggers extra delay depending on how many are dropped so marines have a little bit of extra time

it still validates it as a panic ability since stepping on a hugger makes its delay almost nonexistent and most of the time when you're in panic you don't even have that many huggers, especially at like 25%, so the delay wouldn't be much

PR #11030, opened 2022-09-22
you can no longer be moved when using mounted guns

tyeagg:
## About The Pull Request
you can no longer be moved while using mounted guns, you can also reload without having to get off it now removed clicking to get off gun since that just puts you inside of the tile of the gun with this use wasd to get off also unable to push gun if its manned also removes being able to deploy deployables diagonally because it causes this (you cannot get on the gun even if your behind it you have to click it diagonally and it breaks turning)
image

https://user-images.githubusercontent.com/98577688/191786171-e3c6806f-a6b5-4962-99b4-a7025d48b7fe.mp4

## Why It's Good For The Game
pvt pisser shitter can no longer stop you from getting the epic frag by pushing you or pushing the gun and you can reload without getting off the gun

## Changelog
:cl:
qol: can no longer be moved while using mounted gun 
qol: reload without getting off a mounted gun
qol: movable mounted guns can't be moved if manned even if its unanchored
/:cl:


FosterGrime-Heymynameslouie:
s pede merhg immdeantly

maikilangiolo:
Extremely based change. It's a buff not because of runner pounce, but because MARINES KEEP PUSHING ME AROUND STOP PUSHING ME GO AROUND

Losenis:
I'm calling the mods, it's against the rules for a pr to be this based. This also fixes the dumb af problem of mounted guns having 1 less view range than actual guns because you're standing a tile behind the actual shooting gun.

tyeagg:
> I'm calling the mods, it's against the rules for a pr to be this based. This also fixes the dumb af problem of mounted guns having 1 less view range than actual guns because you're standing a tile behind the actual shooting gun.
> 
> Does this function with MG27 and MG08? If it works with the MG08 you should probably make it so that it's unusable unless it's anchored, since I can imagine a guy using it and another guy pushing behind/dragging with bike it to detroit-driveby xenos.

functions with all mounted guns you cant really use it with mg 08 or tat since it just unsets you way it works is just that it puts you in the same tile as the mounted gun with pixel offset for visual

Losenis:
SPEEDMERGE

XSlayer300:
noooo my hivemind shuffle abuse!!

UntoldTactics:
speedmergnow

BraveMole (changes requested):
The buff is huge. Doesnt that make you mostly invulnerable to xeno spits and a ton of xeno abilities ?

Even if it doesnt make you immune, this should just prevent marine from pushing/moving you, not xenos imo

tyeagg:
> The buff is huge. Doesnt that make you mostly invulnerable to xeno spits and a ton of xeno abilities ?
> 
> Even if it doesnt make you immune, this should just prevent marine from pushing/moving you, not xenos imo

first is solved second i think this should prevent xeno from pushing you because hivemind shuffle and stuns exist already

Losenis:
>Doesnt that make you mostly invulnerable to xeno spits

Why would a xeno spit at a revivable marine that's probably healing on chems rather than spit at the mounted object that'll be destroyed for good?

MereWalker:
Been asking for this for a while. Glad to see someone finally worked on it.

tyeagg:
before you ask, no the gun having no density doesn't prevent projectiles from hitting it

BraveMole:
Way too extrem of an action. It buffs deployed gun way too much (when they already are a pain to deal with), is a potential nest to bug with xeno abilities and such (you should not be in the same turf as a dense object...).

If you want to remove shuffling, then remove shuffling, no more

Starfirejordan:
We need this so Tadpole gun can be used with ease instead of pain

Losenis:
If Unga McDunga the 3h PVT stands next to you you can't turn the gun and you're punished for someone else's actions, nor is there a way to forcibly remove them other than PBing them in the face. Shuffling was one of the issues but not the only one. This'd have fixed both, honestly.

PR #11033, opened 2022-09-22
Fix the AI not being able to target weeds

lbnesquik:
## About The Pull Request
The AI recently got the ability to find the range of turfs, and aim any linked artillery to it. 
Unfortunately, it did this only when clicking on turfs, and as it turns out, weed is not considered one. 
This is unintended behavior, and this PR fixes that.
I've made a little proc for it, in case we have to add other things in the future. 

## Why It's Good For The Game
Fixing unintended behavior is good.

## Changelog
:cl:
fix: Fix the AI not being able to target weeds. 
/:cl:


code/_onclick/ai.dm
```diff
@@ -190,24 +190,37 @@
...
+	var/area/A = get_area(target)
 	if(istype(A) && A.ceiling >= CEILING_UNDERGROUND)
 		to_chat(user, span_warning("You cannot hit the target. It is probably underground."))
 		return
 	to_chat(user, span_notice("Sending targeting information to [user.linked_artillery]. COORDINATES: X:[x] Y:[y]"))
-	user.linked_artillery.recieve_target(src,user)
+	user.linked_artillery.recieve_target(target, user)
+
+/turf/AIShiftClick(mob/living/silicon/ai/user)
+	ai_artillery_aim(user, src)
```
TiviPlus, line 208: Just make a signal on the AI for this instead of puttin gai altillery aim on like everythin in the game

TiviPlus (changes requested):

code/_onclick/ai.dm
TiviPlus: I doubt I will merge it in this state then but I might talk about it with brave

BraveMole: I already said no to small about that on discord.

PR #11039, opened 2022-09-22
actually prevents insane rate of fire

Zar000:
## About The Pull Request

Adds autoburst_delay = 5 to mp19 without making it useless and current " fix " doesnt seem to solve it

## Why It's Good For The Game

Melting not fun

## Changelog
:cl:
balance: nerfing mp19 burst delay to remove incentive from abusing it in akimbo
/:cl:


Lumipharon:
You uh, realise people are just using normal burst fire right? until my latest pr, autoburst is slower than normal burst on almost every single gun.

PR #11035, opened 2022-09-22
Adds the Pheromone Tower

lKiseki:
## About The Pull Request
This is a web commit. Fear me, maintainers.

Adds a new **xenomorph structure** to the game, called the **Pheromone Tower**.
This structure can be purchased using the Mother's Blessings menu, for **150 psychic points**.

Untitled

This structure **can be interacted with**, allowing you to **select the type of pheromone** that it will emit.
It can emit **recovery**, **warding**, and **frenzy**, and the strength of these pheromones is the maximum possible strength (**Very strong**, comparable to a mature queen's).
Its range is **22 tiles**, on par with a mature Queen's pheromones range.
Its **color**, and the **faint light** it emits, **change** according to the selected pheromone type.

## Why It's Good For The Game
- Adds more options for xenomorphs to the Mother's Blessings menu.
- Allows certain castes like the Queen/Empress to emit other types of pheromones, rather than being forced to emit Recovery at all times if they want to assist their team (the same applies to a Shrike in the absence of a Queen).

## Changelog
:cl: Lewdcifer
add: Adds the Pheromone Tower to the game, alongside its related sprites.
/:cl:

PR #11041, opened 2022-09-23
Adds the Pheromone Tower (VSC edition)

lKiseki:
## About The Pull Request
This is NOT a web commit. Damned maintainers...
Note that I opened a PR for this before, but was rejected because I was doing web commits.
The sprites have already been approved by Kuro in this other PR, check #11035 for that.

Adds a new **xenomorph structure** to the game, called the **Pheromone Tower**.
This structure can be **purchased** using the **Mother's Blessings** menu, for **100 psychic points**.

Untitled

This structure **can be interacted with**, allowing you to **select the type of pheromone** that it will emit.
It can emit **recovery**, **warding**, and **frenzy**, and the strength of these pheromones is the maximum possible strength (**Very strong**, comparable to a mature queen's).
Its range is **22 tiles**, on par with a mature Queen's pheromones range.
Its **color**, and the **faint light** it emits, **change** according to the selected pheromone type.

## Why It's Good For The Game
- Adds more options for xenomorphs in the Mother's Blessings menu.
- Allows certain castes like the Queen/Empress to emit other types of pheromones, rather than being forced to emit Recovery at all times if they want to assist their team (the same applies to a Shrike in the absence of a Queen).

## Changelog
:cl: Lewdcifer
add: Adds the Pheromone Tower to the game, alongside its related sprites.
/:cl:


Kuro020 (approved):

FosterGrime-Heymynameslouie:
speede merhg instantly

also I think it should be like, 50 psy points or be buildable normally just with an abnormally long ass cooldown because getting psy points to begin with is a bit of a hassle already. I think.

private-tristan:
150 psypoints seems crazy expensive for this, as foster said, maybe it should instead emit a weak Z-level wide pheromone (scaling with number of towers)?

also who can change the pheremones of the tower? it'd be quite troublesome if someone kept switching it around, maybe only queen/shrike+leaders+hivelord

lKiseki:
In lieu of the above posts, I've decided to drop the cost from **200** (my mistake here, was supposed to be 150 but I overlooked that) to **100**. I will not drop it any further, however, since tier 5 pheromones are no joke.

lKiseki:
> also who can change the pheremones of the tower? it'd be quite troublesome if someone kept switching it around, maybe only queen/shrike+leaders+hivelord

Anyone can. If maintainers believe it pertinent, it can be done, but my personal feelings on the matter are against such a restriction.

Pariah919 (changes requested):
> In lieu of the above posts, I've decided to drop the cost from **200** (my mistake here, was supposed to be 150 but I overlooked that) to **100**. I will not drop it any further, however, since tier 5 pheromones are no joke.
>
400 PP minimum at tier 5 pheros considering I know the main reason anyone would use this is to stack resting weeds, and this frees up queen pheros to be far more flexible.

lKiseki:
> 400 PP minimum at tier 5 pheros considering I know the main reason anyone would use this is to stack resting weeds, and this frees up queen pheros to be far more flexible.

I beg your pardon? Pheromone towers, worth 400 psychic points _(100 more than maturity and evolution tower)_, because it can be stacked with resting weeds and gives queen flexibility?

I think I'd rather wait and see what Tivi or Mole say. I'm open to the idea of increasing its cost back to **200** if need be, but 400 is just outrageous. Don't forget the fact that this is a purchase, not something they're given for free. It has to be worth the cost.

UntoldTactics:
Is anyone not noting that its still limited in range? It has to be within close proximity (relative) to marines. That means it's likely to be destroyed if marines push- its good for hive defense but otherwise its very...tactically odd, and might only be viable in certain mazes or in FOB sieges. The cost should accordingly probably be on the lower side or it just won't be used as much if ever.

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -1163,6 +1163,64 @@ TUNNEL
 		if(EXPLODE_LIGHT)
 			take_damage(300)
 
+/obj/structure/xeno/pherotower
+	name = "Pheromone Tower"
```
XSlayer300, line 1167:
```suggestion
	name = "Pheromone tower"
```
consistancy for the maturity tower above.

KrabSpider:
I can see the recovery tower being used for setting up a healing bay in out of the way locations and such.

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -1163,6 +1163,64 @@ TUNNEL
...
+	bound_height = 64
+	obj_integrity = 400
+	max_integrity = 400
+	xeno_structure_flags = CRITICAL_STRUCTURE
+	///The type of pheromone currently being emitted.
+	var/datum/aura_bearer/current_aura
+	///Strength of pheromones given by this tower.
+	var/aura_strength = 5
+	///Range (in tiles) of the pheromones given by this tower.
+	var/aura_range = 22
```
TiviPlus, line 1181: diameter or radius?

TiviPlus (changes requested):
Rest LGTM

MereWalker:
Should make it emit low light for red, slightly brighter light for yellow and harsh light for blue.You can also choose to ignore me. Nonetheless I like this for xenos.

Weedmaster77:
Anything above T3 pheros from a tower seems way to much from a structure for only 100 PP it need to be like 25 PP-30 PP for T1, 30 PP- 50 PP for T2,50 PP-100 PP for T3,120 PP -150 PP for T4 and 200 PP-250 PP forT5 pheros. Then give the tower the ability to upgrade it to the next Tier. An thinking T5 for 100 is a good idea is so broken T5 needs to be at least 200-250 if not more or it just takes away the usefulness of pheros caste so much which people aright don't want to play so taking away a large reason of play them for a place-able 100 PP T5 tower is a bad idea.

PR #11060, opened 2022-09-25
Remove nuclear shaking for people on other Z level

lbnesquik:
## About The Pull Request
When the nuke goes off, it shakes everyone, no matter the Z-level. 
This PR changes that to only shake if the player is on the same Z-level as the nuke. 

## Why It's Good For The Game
Makes a little more sense, and saves the drama for people who are about to die.

## Changelog
:cl:
del: Removed shaking for people on different Z-level when nuke goes off.
/:cl:


Lumipharon:
shaking is cool when you escape in the alamo/tad in the nick of time tho.

Lumipharon:
You could probs just make it shake the transition z-level, not a big deal either way though.

PR #11073, opened 2022-09-26
Finally fix the window issue on POS medical

lbnesquik:
## About The Pull Request
Before : 
image
After :
image

Grayson and I fixed it in #mapping. The issue was that the widow was there, but it was looking for a sprite which did not exist.
We fixed it by duplicate the likely correct sprite and putting it in the missing place (white_rwindow7 cloned into white_rwindow6)

I also added another sprite that looks more 'logical', in the event we ever have our windows make a distinction between vendors & windows. 

## Why It's Good For The Game
It looked silly, and now it doesn't. 

## Changelog
:cl:
fix: fix POS Medical invisible window issue
/:cl:


RipGrayson (approved):
Approved but since this change technically only touches a sprite it's a Kuro PR

TiviPlus (approved):

FosterGrime-Heymynameslouie:
I always thought it was supposed to be a public door and the shutter is a replacement for the window due to the sheer amount of people who use the chem dispensers, and the fact that they already destroy it 100% of the time to get in

PR #11078, opened 2022-09-26
MP now has more fitting ranks.

Pariah919:
## About The Pull Request
Mech pilot ranks make absolutely no sense from the draft I saw Tivi input from a lore perspective, they're above every marine in the Division, if anything they should probably be warrant officers, but WO ranks aren't as well known as enlisted ranks, they now cap out to being corporals.

## Why It's Good For The Game
MP is important becuase of their silly mech, but they don't have 15+ years of experience on average, that's just silly. They'd be the highest enlisted on board on average.

## Changelog
:cl:
tweak: MPs have lower rank.
/:cl:


TiviPlus:
preapproval applies to you too

PR #11079, opened 2022-09-27
Panic Hugger Rebalance

Naaanii:
## About The Pull Request
Increases threshold for Carrier's panic drop to 40%.
Increases the cooldown to 3 minutes so it's not as spammable.

## Why It's Good For The Game
In it's initial state panic drop was used as more of an offensive tool rather than in the case of panic.

In it's current state, it's barely useable as carriers aren't the most durable castes and most forms of damage will just skip out the 25% health threshold resulting in crit with no chance to use the ability at all.

## Changelog
:cl:
balance: Increased panic drop activation threshold to 40% health.
balance: Increased panic drop's cooldown to 180 seconds.
/:cl:


BraveMole:
Lumi is right this should not be easy to use it is down right op

PR #11080, opened 2022-09-27
Refactors SSauras to be more consistant and to respect MC_CHECK_TICK

DeltaFire15:
## About The Pull Request
Basically, the problem with the aura subsystem at the moment is that it "pulses" mobs on SSaura, but "finishes" cycles (clears phero received while applying effects for what pulsed it, etc.) on Life which is on SSmobs. This means that when SSaura and SSmobs get sufficiently delayed or desynced relative to eachother (high TD tends to cause this), mobs will begin to have cycles they did not receive a pulse from SSauras, leading to mobs not gaining aura effects despite being supposed to.
This addresses said issue by also moving the call of the finishing proc into SSauras, iterating through all humans and xenos to do so. It seemed like the easiest solution to achieve consistancy with only having to iterate through those mobs once per cycle and other possible options to combat the issue that I could think of tended to require more inefficient means than just iterating through those two mob lists once.
If a maintainer has a better idea feel free to suggest so, but if not I think this one should be acceptable.

After some pointers from tivi, this also takes fire() in general and makes it work with MC_CHECK_TICK which should prevent it from taking too much of the MC's time. Also changes phero durations to ticks instead of subsystem ticks and fixes pheromone affecting xenos on fire.

## Why It's Good For The Game
All auras breaking at moderate TD by cause of subsystems ticking at different times and creating a blind timeframe due to that is a bad thing. This fixes that issue.

## Changelog
:cl:
fix: Pheromones (And orders) should no longer "flicker" at high TD, remaining consistant instead.
fix: Xenos can no longer receive pheromones when on fire.
refactor: SSauras now includes MC_CHECK_TICK to prevent it from eating too much of the MC's time.
/:cl:

### DIFF
--- a/code/controllers/subsystem/aura.dm
+++ b/code/controllers/subsystem/aura.dm
@@ -5,11 +5,48 @@ SUBSYSTEM_DEF(aura)
 	flags = SS_NO_INIT
 	///Stores currently active aura_bearer datums
 	var/list/active_auras = list() //We can't use a normal subsystem processing list because as soon as an aura_bearer leaves the list it needs to die
+	///Stores what the current aura processing stage was working on for if it is paused.
+	var/list/current_cache = list()
+	///Auras go through three stages. Pulse auras -> finish xeno cycles -> finish human cycles. Depending on where it was paused, this tells it where to resume fire()
+	var/stage = 1
 
 /datum/controller/subsystem/aura/fire(resumed)
-	finish_aura_cycles()
-	for(var/datum/aura_bearer/bearer AS in active_auras)
-		bearer.process()
+	if(stage == 1)
+		if(!resumed)
+			current_cache = active_auras.Copy()
+		while(current_cache.len)
+			var/datum/aura_bearer/bearer = current_cache[current_cache.len]
+			current_cache.len--
+			if(QDELETED(bearer))
+				continue
+			bearer.process()
+			if(MC_TICK_CHECK)
+				return
+		stage = 2
+	if(stage == 2)
+		if(!resumed)
+			current_cache = GLOB.xeno_mob_list.Copy()
+		while(current_cache.len)
+			var/mob/living/carbon/xenomorph/xeno = current_cache[current_cache.len]
+			current_cache.len--
+			if(QDELETED(xeno))
+				continue
+			xeno.finish_aura_cycle()
+			if(MC_TICK_CHECK)
+				return
+		stage = 3
+	if(stage == 3)
+		if(!resumed)
+			current_cache = GLOB.human_mob_list.Copy()
+		while(current_cache.len)
+			var/mob/living/carbon/human/human = current_cache[current_cache.len]
+			current_cache.len--
+			if(QDELETED(human))
+				continue
+			human.finish_aura_cycle()
+			if(MC_TICK_CHECK)
+				return
+		stage = 1
 
 ///Use this to start a new emitter with the specified stats. Returns the emitter in question, just qdel it to end early.
 /datum/controller/subsystem/aura/proc/add_emitter(atom/center, type, range, strength, duration, faction)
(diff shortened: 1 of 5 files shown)


QualityVan:
Looks good to me. I just assumed that as tidi increased, both systems' cycle time would stay more or less the same anyway, and this works nicely around that being wrong. My only concern at a glance is that this'll cause cycling through dead mobs, which didn't tick before. Maybe swap to alive_human/xeno_list instead? But that could cause auras to stick on corpses, if that matters at all.

DeltaFire15:
> Looks good to me. I just assumed that as tidi increased, both systems' cycle time would stay more or less the same anyway, and this works nicely around that being wrong. My only concern at a glance is that this'll cause cycling through dead mobs, which didn't tick before. Maybe swap to alive_human/xeno_list instead? But that could cause auras to stick on corpses, if that matters at all.

I'm checking through all instead of the limited lists due to the significant possibility of conditions changing between them receiving phero and it checking again.

code/controllers/subsystem/aura.dm
```diff
@@ -5,11 +5,48 @@ SUBSYSTEM_DEF(aura)
...
+			current_cache.len--
+			if(QDELETED(bearer))
+				continue
+			bearer.process()
+			if(MC_TICK_CHECK)
+				return
+		stage = 2
+	if(stage == 2)
+		if(!resumed)
+			current_cache = GLOB.xeno_mob_list.Copy()
```
QualityVan, line 28: If you resume during stage one this'll never happen.

code/controllers/subsystem/aura.dm
```diff
@@ -90,15 +119,17 @@ SUBSYSTEM_DEF(aura)
...
 		pulse_humans()
 	if(affects_xenos)
 		pulse_xenos()
 	if(duration < 0) //Negative duration means infinite. Usually for pheromones.
 		return
-	duration--
-	if(!duration)
+	duration -= (world.time - last_tick)
+	last_tick = world.time
+	if(duration <= 0)
```
QualityVan, line 132: Is there a benefit to this over just calculating desired ending world.time at aura start and checking against it?

code/controllers/subsystem/aura.dm
```diff
@@ -124,5 +155,7 @@ SUBSYSTEM_DEF(aura)
 			continue
 		if(potential_hearer.faction != faction)
 			continue
+		if(!(potential_hearer.xeno_caste.caste_flags & CASTE_FIRE_IMMUNE) && potential_hearer.on_fire) //Xenos on fire cannot receive pheros.
```
QualityVan, line 158: Won't a fire immune caste never be on fire?

code/controllers/subsystem/aura.dm
DeltaFire15: suppression is supposed to freeze duration, no? That would be impossible to handle with that method which is why I chose this.

code/controllers/subsystem/aura.dm
DeltaFire15: oh, yeah makes sense.

code/controllers/subsystem/aura.dm
DeltaFire15: I took the check handling from handle_fire's phero suppression. I suppose it could happen if something modified caste flags while on fire, but that usually shouldn't happen, yeah

code/controllers/subsystem/aura.dm
QualityVan: Suppression interacting with duration wasn't actually something I put thought into when setting this up, since only xenos use suppression and their duration is always infinite. The purpose I had in mind was just to make the aura in question not emit for a tick.

code/controllers/subsystem/aura.dm
DeltaFire15: Should be resolved.

code/controllers/subsystem/aura.dm
DeltaFire15: "Whether we should skip the next tick. Set to false after skipping once. Won't pulse to targets **or reduce duration**.
I simply preserved the functionality in the var's autodoc comment.

TiviPlus (approved):

comfyorange:
#### Test Merge Deployed

##### Server Instance
TGMC

##### Merged By
GAME\tiviplus

##### Revision
Origin: fd13cccb034c3ed8b1e7f742ecb93ac14c7e70a6
Pull Request: 1b2fc9a3680bdea4d492d7ef7c4e5855fc2a2cd0
Server: 932d5fc994524e8e809075a694fc03e13400a193

##### Comment
No comment set - Fast Add
[GitHub Deployments](https://github.com/tgstation/TerraGov-Marine-Corps/deployments/activity_log?environment=TGS%3A%20TGMC)

comfyorange:
#### Test Merge Updated

##### Server Instance
TGMC

##### Merged By
GAME\tiviplus

##### Revision
Origin: fd13cccb034c3ed8b1e7f742ecb93ac14c7e70a6
Pull Request: 1b2fc9a3680bdea4d492d7ef7c4e5855fc2a2cd0
Server: 03bfcac3bdfdd9dab131bb3b9be127e451515037

##### Comment
No comment set - Fast Add
[GitHub Deployments](https://github.com/tgstation/TerraGov-Marine-Corps/deployments/activity_log?environment=TGS%3A%20TGMC)

comfyorange:
#### Test Merge Updated

##### Server Instance
TGMC

##### Merged By
GAME\naaanii

##### Revision
Origin: d9f05eb44f958814783b33f5e725a88de2ce56cb
Pull Request: 1b2fc9a3680bdea4d492d7ef7c4e5855fc2a2cd0
Server: a1f8f0c0a902ff188116e56b2f95f1572f222a74

##### Comment
No comment set - Fast Add
[GitHub Deployments](https://github.com/tgstation/TerraGov-Marine-Corps/deployments/activity_log?environment=TGS%3A%20TGMC)

comfyorange:
#### Test Merge Removed

PR #11082, opened 2022-09-27
Makes mechs unable to immediately flip around

TiviPlus:
Kuro pointed out this mitigated a lot of the direction based mechanics

Now instead they will randomly pick right or left and then face the direction you were moving towards


TiviPlus:
Going to maybe pr this to tg as well so you can wait on merge if you dont feel like it brave

DeltaFire15:
This should check for strafing, otherwise you'll just randomly strafe left and right without ever moving backwards.

Starfirejordan:
Wraith and runner are gonna have a field day against mech that cant turn for shit

PR #11087, opened 2022-09-28
Makes engineering storage module fit filled sandbags

maikilangiolo:
## About The Pull Request
Engie storage for armor can now take filled sandbags

## Why It's Good For The Game
With all the other great storage modules, it was an annoying handicap not being able to carry filled sandbags in it. Grenade and injector modules are great choices, and engie module does the same thing as a webbing (which also fits empty sandbags and tools), so I feel like this is a fair compromise for what are seldom use barricades.

## Changelog
:cl:
balance: engineering storage module can fit filled sandbags
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/storage.dm
+++ b/code/modules/clothing/modular_armor/attachments/storage.dm
@@ -169,6 +169,7 @@
 		/obj/item/stack/rods,
 		/obj/item/stack/cable_coil,
 		/obj/item/stack/sandbags_empty,
+		/obj/item/stack/sandbags,
 		/obj/item/stack/razorwire,
 		/obj/item/tool/shovel/etool,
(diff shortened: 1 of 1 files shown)


FosterGrime-Heymynameslouie:
finally, something that does something to how absurdly cumbersome filled sandbags are

spded mergh now

MLGTASTICa:
I find this a balance change and one which is bad.
Sandbags are meant to be cheaper than metal, but extremely cumbersome and require shoveling to actually fill.

maikilangiolo:
They're already balanced for that. The max stack is 50, and when filled it's 25 (twice the storage for the same amount). You can also only repair them with more sandbags instead of (infinite) welder.

TiviPlus (approved):

PR #11090, opened 2022-09-28
Mp holster in clothing vensor, smg25 holster in seasonal

maikilangiolo:
## About The Pull Request
Adds the two holsters for the mp and seasonal smg

## Why It's Good For The Game
Drip, i guess, and in the case of the smg25 something to carry it and make it worth it over the p90, which is generally better all around.

## Changelog
:cl:
add: added holsters for the mp19 and smg-25
/:cl:

### DIFF
--- a/code/controllers/subsystem/persistence.dm
+++ b/code/controllers/subsystem/persistence.dm
@@ -255,6 +255,7 @@ SUBSYSTEM_DEF(persistence)
 		/obj/item/ammo_magazine/smg/uzi = -1,
 		/obj/item/weapon/gun/smg/m25 = -1,
 		/obj/item/ammo_magazine/smg/m25 = -1,
+		/obj/item/storage/holster/m25 = -1,
 		/obj/item/weapon/gun/smg/mp7 = -1,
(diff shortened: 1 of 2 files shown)


FosterGrime-Heymynameslouie:
unsure as to how this would go down with people who get dunked on constantly by the machine peestol people but having unusued guns finally be used, and having the SMG section finally not only being comprised by p90 (the machine pistol dosent count) is cool

maikilangiolo:
> unsure as to how this would go down with people who get dunked on constantly by the machine peestol people

actually changes nothing. the MP is only good when dual wielded, and this would only let you carry one without any ammo to boot. If someone choose to carry a single one, and likely with a stock since they wont dual it, this holster is a valid choice, if only for the drip.

Lumipharon:
MP19 fits in the belt slot already, the holster is purely decorative, and for the whoosh sound.

TiviPlus (changes requested):
No SMG-25/holster for it we have seasonals for it already

TiviPlus (approved):

PR #11101, opened 2022-09-29
check path proc

Lumipharon:
## About The Pull Request
Adding a generic proc to draw a path between two atoms and check to see if anything is blocking that path, with a few args to tweak it's behavior.
This is useful because procs like get_hear() ignore transparent objects like windows or machinery, and is blocked by opaque atoms like smoke.

I made this for widow, but there's other things it can be used for, so PR'ing it on it's own for now, but let me know if that's not OK.

## Why It's Good For The Game
Useful proc good

## Changelog
:cl:
code: Added a check_path proc to see if a path between atoms is blocked
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1270,7 +1270,7 @@ GLOBAL_LIST_INIT(survivor_outfits, typecacheof(/datum/outfit/job/survivor))
 /proc/check_path(atom/start, atom/end, bypass_window = FALSE, projectile = FALSE, bypass_xeno = FALSE)
 	var/list/path_to_target = getline(start, end)
 	var/line_count = 1
-	while(line_count < length.path_to_target)
+	while(line_count < length(path_to_target))
 		if(LinkBlocked(path_to_target[line_count], path_to_target[line_count + 1], bypass_window, projectile, bypass_xeno))
 			return FALSE
 		line_count ++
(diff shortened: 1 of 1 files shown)


code/__HELPERS/unsorted.dm
```diff
@@ -1257,3 +1257,23 @@ will handle it, but:
...
+ */
+/proc/check_path(atom/start, atom/end, bypass_window = FALSE, projectile = FALSE, bypass_xeno = FALSE)
+	var/list/path_to_target = getline(start, end)
+	var/line_count = 1
+	for(var/path_turf in path_to_target)
+		if(line_count == length(path_to_target))
+			break
+		if(LinkBlocked(path_to_target[line_count], path_to_target[line_count + 1], bypass_window, projectile, bypass_xeno))
+			return FALSE
+		line_count ++
```
TiviPlus, line 1278: why does line_count exist if this isnt a while loop?

TiviPlus (changes requested):

code/__HELPERS/unsorted.dm
Lumipharon: made it a while since that basically what I was going for, unless you think it's better to do otherwise.

Lumipharon: I'll fix tomorrow.

TiviPlus (approved):

PR #11127, opened 2022-10-03
Runechat shows on top of whatever you're in

TiviPlus:
## About The Pull Request
ported from tg cant find the pr anymore

## Why It's Good For The Game

Runechat works even in mechs, droppods or closets!!

## Changelog
:cl:
add: runechat now works when you are inside something
/:cl:

### DIFF
--- a/code/datums/chatmessage.dm
+++ b/code/datums/chatmessage.dm
@@ -140,7 +140,7 @@
 	approx_lines = max(1, mheight / CHAT_MESSAGE_APPROX_LHEIGHT)
 
 	// Translate any existing messages upwards, apply exponential decay factors to timers
-	message_loc = target
+	message_loc = isturf(target) ? target : get_atom_on_turf(target)
 	if (owned_by.seen_messages)
 		var/idx = 1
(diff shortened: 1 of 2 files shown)


code/datums/chatmessage.dm
```diff
@@ -140,7 +140,7 @@
 	approx_lines = max(1, mheight / CHAT_MESSAGE_APPROX_LHEIGHT)
 
 	// Translate any existing messages upwards, apply exponential decay factors to timers
-	message_loc = target
+	message_loc = isturf(target) ? target : get_atom_on_turf(target)
```
BraveMole, line 143: I am a bit scared by that while loop on every message tbh

BraveMole (changes requested):

code/datums/chatmessage.dm
TiviPlus: We use it in every balloon message already /shrug

TiviPlus: And well it does do what it says

PR #11142, opened 2022-10-05
fix synthesized instruments

Dikozda:
## About The Pull Request

currently synthesized instruments tend to bug out if the sound they try to play is too long.

turns out this is because the arguments in a proc call were putting the channel number into is_global and setting the actual channel to null resulting in the sounds getting assigned a random channel, meaning that the instrument will try to update the wrong channels

this just sets is_global to false and replaces null with channel

also would love if anyone could double check this, I am confident this should work but second opinions are nice.

## Why It's Good For The Game

not having synthesized instruments bug out is nice

## Changelog
:cl:
fix: fixed synthesized instruments
/:cl:


TiviPlus (approved):
Oh my god it was this all along 😭

PR #11148, opened 2022-10-06
M3 Harness and Leather Jackets now accept Jaeger modules

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
Just because they look different or have 0 armor and no slowdown doesn't mean they shouldn't get access to a crucial part of marine equipment.

## Changelog
:cl: Lewdcifer
balance: M3 Harness and Leather jackets now accept Jaeger modules.
/:cl:

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -129,7 +129,6 @@
 	soft_armor = list(MELEE = 0, BULLET = 0, LASER = 0, ENERGY = 0, BOMB = 0, BIO = 0, "rad" = 0, FIRE = 0, ACID = 0)
 	slowdown = 0
 	flags_atom = NONE
-	flags_item = SYNTH_RESTRICTED
 
 	attachments_by_slot = list(
 		ATTACHMENT_SLOT_MODULE,
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -130,6 +130,26 @@
 	slowdown = 0
 	flags_atom = NONE
 
+	attachments_by_slot = list(
+		ATTACHMENT_SLOT_MODULE,
+		ATTACHMENT_SLOT_BADGE,
+	)
+	attachments_allowed = list(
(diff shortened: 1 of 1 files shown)


Lumipharon:
How many times has this been pr'd or suggested?

lKiseki:
> How many times has this been pr'd or suggested?

this was once a thing but it got pr'd out because synths had access to jaeger modules
the alternative is reducing synth health to account for modules, but i don't really remember why it got reverted

Kuro020 (approved):

Al-1ce:
I think leaving things as they are is the better alternative. Different types of armor should have downsides. It should be up to the player to account for that.

Synths having access to modules is a bad thing regardless.

Al-1ce:
Different types of armor should have their respective downsides. Modules can negate that, which is generally bad. This would make it easy to have the mobility of an M3 harness with the armor quality of scout armor.

I suppose I can see it being added to leather jackets but both of these items are basically meme gear that aren't actually intended to be used. This overall just doesn't seem that needed or necessary.

I should also throw in that restricting synths clothing options is simply uncool.

Oscari-pog:
Taking drip away from synths and adding modules to a armor who nobody literally ever uses and will never use even with this weird change.

Losenis:
M3 harness wasn't usable by synth in the first place though. I tried back in the day when I wanted a source of light on the suit slot. The flag is up there in the /obj/item/clothing/suit/storage/marine parent that the M3 inherits.

Losenis:
image

ivanmixo:
> M3 harness wasn't usable by synth in the first place though. I tried back in the day when I wanted a source of light on the suit slot. The flag is up there in the /obj/item/clothing/suit/storage/marine parent that the M3 inherits.
> 
> Also what the actual fuck Kuro, when I suggested this you adamantly opposed it because it looked like shit and wanted better sprites for it.

This is a blatant lie, I use a M3 harness to hold a machette scabbard every time I play synth.

Pariah919 (approved):
Makes sense to me.

Al-1ce:
I could've absolutely sworn I've worn the M3 as a synth. Huh.

PR #11149, opened 2022-10-06
Halve nanite initial adjustment period

Jalopkomit18:
## About The Pull Request
Lowers nanite adjustment period from 150 ticks to 75 ticks. Nanite adjustment before took about 5 minutes or upwards. As such, this should reduce nanite adjustment period to 2.5 minutes. 

## Why It's Good For The Game
Marine setup has gotten faster with loadout vendors, expecting faster dropship times. As such, I think the adjustment period should be tuned down because you can't accomplish much when you're adjusting due to pain/damage. 5min of having to babysit yourself is a little much imo. Should not be a balance concern since it just affects initial adjustment period; just qol in light of increased speed of marine setup. I honestly don't think the nanite adjustment period is necessary at all; it just feels like a noob trab to those unfamiliar and annoying to those in the know. However I think that this is good middle ground solution. 

## Changelog
:cl:
qol: Nanite adjustment period halved to 75 ticks
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1269,16 +1269,16 @@
 
 /datum/reagent/medicine/research/medicalnanites/on_mob_life(mob/living/L, metabolism)
 	switch(current_cycle)
-		if(1 to 150)
+		if(1 to 75)
 			L.take_limb_damage(0.015*current_cycle*effect_str, 0.015*current_cycle*effect_str)
 			L.adjustToxLoss(1*effect_str)
 			L.adjustStaminaLoss((1.5)*effect_str)
-			L.reagents.add_reagent(/datum/reagent/medicine/research/medicalnanites, 0.20)
+			L.reagents.add_reagent(/datum/reagent/medicine/research/medicalnanites, 0.40)
 			if(prob(5))
 				to_chat(L, span_notice("You feel intense itching!"))
-		if(151)
+		if(76)
 			to_chat(L, span_warning("The pain rapidly subsides. Looks like they've adapted to you."))
-		if(152 to INFINITY)
+		if(77 to INFINITY)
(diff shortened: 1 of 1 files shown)


Al-1ce:
Nanites are typically used by goobers, so decreasing the lethality is a good thing from my perspective.

ivanmixo:
This isn't exactly qol, it's balance. But whatever, who cares about the tags amirite

private-tristan:
> This isn't exactly qol, it's balance. But whatever, who cares about the tags amirite

technically yes, but i don't ever see a scenario where someone would inject nanites outside of shipside, so this wouldn't really effect groundside at all. the only time this could possibly be a balance change is when xenos have hijacked and a marine really, really wants his nanites before the alamo crashes, which seems like such a niche and rare scenario that i'd never happen.

TL:DR yes, but actually no

also, due to how nanites will adjust faster, will the marine lose blood from taking them after the adjustment period is over? because they'd only have 15u of nanites in them when adjustment is done.

Jalopkomit18:
> > This isn't exactly qol, it's balance. But whatever, who cares about the tags amirite
> 
> technically yes, but i don't ever see a scenario where someone would inject nanites outside of shipside, so this wouldn't really effect groundside at all. the only time this could possibly be a balance change is when xenos have hijacked and a marine really, really wants his nanites before the alamo crashes, which seems like such a niche and rare scenario that i'd never happen.
> 
> TL:DR yes, but actually no
> 
> also, due to how nanites will adjust faster, will the marine lose blood from taking them after the adjustment period is over? because they'd only have 15u of nanites in them when adjustment is done.

This pretty much summarizes my thoughts on whether it's a balance concern. I highly doubt that this situation or something like it would ever come up.
As for the second question, I doubled the nanite growth rate in the adjustment phase from 0.2 to 0.4. After adjustment ends, it goes back down to the normal nanite growth rate.

PR #11150, opened 2022-10-06
Vali Necro Warning Adjustment

Jalopkomit18:
## About The Pull Request
Currently Vali alerts to impending after 12 secs, giving necro hud and periodic text warnings. Changes to two step warning, with text warning only at 10 secs onwards, and necro hud appearing at <s>17</s> 15 secs onwards. This translates to text warning appearing at 10 seconds left and necro hud appearing at 3 seconds left. Values still undecided until tested.

## Why It's Good For The Game
Current vali warning is imprecise and misleading imo. The margin of 8 seconds on the warning is too wide and defeats the purpose of the warning. Also the text alerts are inconsistent and don't happen at consistent intervals, so you can't rely on them that well. Splitting the warning into text and hud at different intervals would be better I think. Text warning will be there to let you know that necro is approaching and allow you to look over to see that exactish time remaining. However, the hud warning will trigger close before necro hits, giving an actually useful warning to let you know that you have to turn off vali NOW or you will get necro. 
Feedback and thoughts from vali users would be greatly appreciated as this is only my experience with it.

## Changelog
:cl:
qol: Vali necro warning divided into two warnings at separate intervals. Will first give warning with how much time remaining starting at 10secs onwards. Second warning will begin 15secs onwards and give the necro hud and an additional warning.
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -69,7 +69,7 @@
 	///Vali movement speed buff is this value
 	var/movement_boost = 0
 	///How much time left on vali heal till necrosis occurs
-	var/vali_necro_timer = world.time - processing_start
+	var/vali_necro_timer
 
 	/**
 	 * This list contains the vali stat increases that correspond to each reagent
@@ -187,6 +187,7 @@
 
 	wearer.adjustToxLoss(-tox_heal*boost_amount)
 	wearer.heal_limb_damage(6*boost_amount*brute_heal_amp, 6*boost_amount*burn_heal_amp)
+	vali_necro_timer = world.time - processing_start
 	if(connected_weapon && vali_necro_timer < 20 SECONDS)
 		wearer.adjustStaminaLoss(-7*stamina_regen_amp*((20 - (vali_necro_timer)/10)/20)) //stamina gain scales inversely with passed time, up to 20 seconds
 	if(vali_necro_timer > 10 SECONDS && vali_necro_timer < 20 SECONDS)
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -68,6 +68,8 @@
 	var/stamina_regen_amp = 1
 	///Vali movement speed buff is this value
 	var/movement_boost = 0
+	///How much time left on vali heal till necrosis occurs
+	var/vali_necro_timer = world.time - processing_start
 
 	/**
 	 * This list contains the vali stat increases that correspond to each reagent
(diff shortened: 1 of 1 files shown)


Jalopkomit18:
Adjusted values to have text only warning occur starting at 10sec onwards. Necro hud warning occurs at 15sec onwards, and also has its own text warning distinct from the regular warning. Ready to merge.

TiviPlus:
Did you test this?

Jalopkomit18:
> Did you test this?

Yes. The only thing I'm worried about is that with lag and network delay, the timers may be too short. I added a buffer already even though I wanted the final timer to be shorter, but if it's too short still I can change the numbers around again.

code/datums/components/chem_booster.dm
```diff
@@ -187,9 +187,11 @@
 	wearer.heal_limb_damage(6*boost_amount*brute_heal_amp, 6*boost_amount*burn_heal_amp)
 	if(connected_weapon && world.time - processing_start < 20 SECONDS)
 		wearer.adjustStaminaLoss(-7*stamina_regen_amp*((20 - (world.time - processing_start)/10)/20)) //stamina gain scales inversely with passed time, up to 20 seconds
-	if(world.time - processing_start > 12 SECONDS && world.time - processing_start < 15 SECONDS)
-		wearer.overlay_fullscreen("degeneration", /obj/screen/fullscreen/infection, 1)
-		to_chat(wearer, span_highdanger("WARNING: You have [(200 - (world.time - processing_start))/10] seconds before necrotic tissue forms on your limbs."))
+	if(world.time - processing_start > 10 SECONDS && world.time - processing_start < 20 SECONDS)
+		to_chat(wearer, span_bold("WARNING: You have [(200 - (world.time - processing_start))/10] seconds before necrotic tissue forms on your limbs."))
+		if(world.time - processing_start > 15 SECONDS && world.time - processing_start < 20 SECONDS)
```
BraveMole, line 192: You are checking the same thing twice. And also you can cache world.time - processing start in a var

BraveMole (changes requested):

code/datums/components/chem_booster.dm
```diff
@@ -232,7 +237,7 @@
 	if(boost_on)
 		STOP_PROCESSING(SSobj, src)
 		wearer.clear_fullscreen("degeneration")
-		var/necrotized_counter = FLOOR(min(world.time-processing_start, 20 SECONDS)/200 + (world.time-processing_start-20 SECONDS)/100, 1)
+		var/necrotized_counter = FLOOR(min(vali_necro_timer, 20 SECONDS)/200 + (vali_necro_timer-20 SECONDS)/100, 1)
```
Lumipharon, line 240:
```suggestion
		vali_necro_timer = world.time - processing_start
		var/necrotized_counter = FLOOR(min(vali_necro_timer, 20 SECONDS)/200 + (vali_necro_timer-20 SECONDS)/100, 1)
```
We want to make sure  the value is accurate

Lumipharon (changes requested):

code/datums/components/chem_booster.dm
```diff
@@ -185,11 +187,14 @@
...
-		wearer.adjustStaminaLoss(-7*stamina_regen_amp*((20 - (world.time - processing_start)/10)/20)) //stamina gain scales inversely with passed time, up to 20 seconds
-	if(world.time - processing_start > 12 SECONDS && world.time - processing_start < 15 SECONDS)
-		wearer.overlay_fullscreen("degeneration", /obj/screen/fullscreen/infection, 1)
-		to_chat(wearer, span_highdanger("WARNING: You have [(200 - (world.time - processing_start))/10] seconds before necrotic tissue forms on your limbs."))
+	vali_necro_timer = world.time - processing_start
+	if(connected_weapon && vali_necro_timer < 20 SECONDS)
+		wearer.adjustStaminaLoss(-7*stamina_regen_amp*((20 - (vali_necro_timer)/10)/20)) //stamina gain scales inversely with passed time, up to 20 seconds
+	if(vali_necro_timer > 10 SECONDS && vali_necro_timer < 20 SECONDS)
+		to_chat(wearer, span_bold("WARNING: You have [(200 - (vali_necro_timer))/10] seconds before necrotic tissue forms on your limbs."))
+		if(vali_necro_timer > 15 SECONDS)
```
BraveMole, line 195: You do not have to indent that, it being in main scope is fine

code/datums/components/chem_booster.dm
```diff
@@ -185,11 +187,14 @@
 
 	wearer.adjustToxLoss(-tox_heal*boost_amount)
 	wearer.heal_limb_damage(6*boost_amount*brute_heal_amp, 6*boost_amount*burn_heal_amp)
-	if(connected_weapon && world.time - processing_start < 20 SECONDS)
-		wearer.adjustStaminaLoss(-7*stamina_regen_amp*((20 - (world.time - processing_start)/10)/20)) //stamina gain scales inversely with passed time, up to 20 seconds
-	if(world.time - processing_start > 12 SECONDS && world.time - processing_start < 15 SECONDS)
-		wearer.overlay_fullscreen("degeneration", /obj/screen/fullscreen/infection, 1)
-		to_chat(wearer, span_highdanger("WARNING: You have [(200 - (world.time - processing_start))/10] seconds before necrotic tissue forms on your limbs."))
+	vali_necro_timer = world.time - processing_start
+	if(connected_weapon && vali_necro_timer < 20 SECONDS)
```
BraveMole, line 191: Rather than checking timer < 20 two times, you can just check if once and early return it is above.

Ex: 
If(timer > 20)
      Return

BraveMole (changes requested):

PR #11154, opened 2022-10-07
Moves engineer teleporters to engineer gear vendor.

Aquilar:
## About The Pull Request
This PR moves the teleporter option from the engineer clothes vendor to the engineer gear vendor, making it cost 25 points, 25 points less than the Requisition option 

## Why It's Good For The Game
Such a powerful and easy to set up mobility, reinforcement, escape and resupply option should have never been an item that was given out freely especially since the other backpack options the engineer role gets are incredibly easy to obtain making the teleporters an obvious and only backpack pick. Req or an SL can chuck you a radiopack if you really want one, welderpack can be found anywhere.
 
The 35 point cost was picked so the engineer doesn't get the option to bring a normal sentry or an iguana and still allows them to buy a build-a-sentry or a combination of barricade materials, plasma cutter and other options.  Either way engineers will still gladly bring teleporters because they already got a taste and appreciation of it's strengths despite it's new cost of no longer being free.

edit: changed pricing from 35 to 25. 

## Changelog
:cl:
balance: Teleporters moved to engineer gear vendor and no longer free. costing 25 points.
/:cl:


Autisem:
my only gripe is that 35 points still is alot, I know TP is powerfull but kinda fucks with my builds

Naaanii:
25 points would be a nice compromise.

Pariah919 (approved):

PR #11159, opened 2022-10-08
Ammo hud refactor and gives ammo counter to mechs

TiviPlus:
## About The Pull Request

1) No more user passing
2)autodoc
3)Can be given to any datum
4) no longer creates and deletes ammo objects constantly
5) no more spawn
6) other misc cleanups

I folded in mech ammo but I can split it out if you want

## Changelog
:cl:
fix: Ammo HUDs will now always blink when you are out of ammo instead of only the bottommost one
/:cl:

### DIFF
--- a/code/_onclick/hud/screen_objects/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects/screen_objects.dm
@@ -671,75 +671,65 @@
 
 	usr.hud_used.hidden_inventory_update()
 
-
+/**
+ * HUD ammo indicator
+ *
+ * Displays a number and an icon representing the ammo for up to 4 at a time
+ */
 /obj/screen/ammo
 	name = "ammo"
 	icon = 'icons/mob/ammoHUD.dmi'
 	icon_state = "ammo"
 	screen_loc = ui_ammo1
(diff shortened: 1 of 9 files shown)


code/_onclick/hud/screen_objects/screen_objects.dm
```diff
@@ -671,75 +671,65 @@
...
+	overlays.Cut()
 
-/obj/screen/ammo/proc/update_hud(mob/living/user, obj/item/weapon/gun/G)
-	if(!user?.client?.screen.Find(src))
-		return
-
-	if(!G || !(G.flags_gun_features & GUN_AMMO_COUNTER))
-		remove_hud(user)
+	if(rounds <= 0)
+		var/hud_state_empty = ammo_type[2]
```
BraveMole, line 720: Why is that a var ?

BraveMole (changes requested):

code/_onclick/hud/screen_objects/screen_objects.dm
TiviPlus: because "2" isnt very clear, i can make it a define if you prefer

PR #11171, opened 2022-10-09
Removes damage reduction from robotic limbs

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
imagen

## Changelog
:cl: Lewdcifer
balance: Robotic limbs no longer have a damage reduction.
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -169,10 +169,6 @@
 	if(limb_status & LIMB_DESTROYED)
 		return 0
 
-	if(limb_status & LIMB_ROBOT && !(owner.species.species_flags & ROBOTIC_LIMBS))
-		brute *= 0.50 // half damage for ROBOLIMBS if you weren't born with them
-		burn *= 0.50
-
 
 	//High brute damage or sharp objects may damage internal organs
 	if(internal_organs && ((sharp && brute >= 10) || brute >= 20) && prob(5))
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -169,6 +169,10 @@
 	if(limb_status & LIMB_DESTROYED)
 		return 0
 
+	if(limb_status & LIMB_ROBOT && !(owner.species.species_flags & ROBOTIC_LIMBS))
+		brute *= 0.50 // half damage for ROBOLIMBS if you weren't born with them
+		burn *= 0.50
+
 
 	//High brute damage or sharp objects may damage internal organs
 	if(internal_organs && ((sharp && brute >= 10) || brute >= 20) && prob(5))


Al-1ce:
this is good since the 30%+ burn penalty for robots never actually got implemented.

helper-2:
> this is good since the 30%+ burn penalty for robots never actually got implemented.

They don't have any buff or debuff to a damage resistance. Also, they don't have damage redaction from a replaced limb.

QualityVan:
This is funny to me because there were years on end where people have complained endlessly that robotic limb replacements are awful and never worth it and you might as well leave the round if you lose an arm.

Al-1ce:
> This is funny to me because there were years on end where people have complained endlessly that robotic limb replacements are awful and never worth it and you might as well leave the round if you lose an arm.
> Also, yeah, this is only for replacement limbs. No change for robots.


Oh. Well that changes things. Not really a fan of this then.

Naaanii:
> This is funny to me because there were years on end where people have complained endlessly that robotic limb replacements are awful and never worth it and you might as well leave the round if you lose an arm.
> 
> Also, yeah, this is only for replacement limbs. No change for robots.

They were buffed along with arms when delimbing and aiming arms were the meta iirc.

QualityVan:
> They were buffed along with arms when delimbing and aiming arms were the meta iirc.

Two years ago, yes. Prior to that it was a .66 multiplier to damage taken instead of .5.

TiviPlus (approved):

Losenis:
Tested it in valhalla, the picture seems misleading because for me it reduced the final damage by half while in that picture it seems to have full immunity. Unless I'm mistaken in the math, for it to have damage immunity it should already have 100+ effective armor before the dmg multiplier of the limb. Not only that, in examination it shows the affected limb as bruised rather than dented as well as showing no difference in healthscanner to indicate it's a robot limb. Unless I'm missing something, having robot limbs is already a pain in the ass even with the dmg multiplier given the additional loadout to accomodate a quadrillion 20-capacity blowtorches and wires to manually heal out of combat due to not chem.

image

srxtreme:
now make a screenshot of b18, PMC uniform (which is getting removed btw), FC hold order, premed + nanites.

Nickkoranda:
Skewed image, that is with PMC outfit + b18 + robo limbs + double shield + hold order. While 1/2 damage is pretty crazy the root cause is armor stacking more than damage reduction from limbs. You still experience the full damage if xenos aim chest, and you're still heavily affected by neuro (b18 so you can't use mimir). You're still susceptible to facehuggers, unless you're swapping your SWAT mask constantly. As mentioned before you also need to heal the limbs with separate mats than just chems, but you still need to have chems because of your chest/head.

lbnesquik:
The reasons for this PR aren't very readable. What are we looking at ?

PR #11172, opened 2022-10-09
Removes most uniform and mask armor values

Lumipharon:
## About The Pull Request
Edit: This PR now removes the armour values from hopefully all non admeme uniforms and powergamable masks.

Removes the armour from PMC uniforms and balaclavas and put it into their helmets and armour.
No change for PMC, but prevent marine cheese when they get their hands on it on certain maps.

## Why It's Good For The Game
Super high melee resist cheese bad

## Changelog
:cl:
balance: Removed armour from PMC masks and uniforms, and increased their helmet and suit armour accordingly
balance: Removed armour from non admin uniforms
/:cl:

### DIFF
--- a/code/modules/clothing/head/helmet.dm
+++ b/code/modules/clothing/head/helmet.dm
@@ -439,7 +439,7 @@ obj/item/clothing/head/helmet/marine/pilot/green
 	name = "\improper PMC tactical helmet"
 	desc = "A protective helmet made from flexible aramid materials. Standard issue protection for a lot of security firms."
 	icon_state = "pmc_helmet"
-	soft_armor = list(MELEE = 55, BULLET = 55, LASER = 55, ENERGY = 50, BOMB = 50, BIO = 50, "rad" = 50, FIRE = 50, ACID = 50)
+	soft_armor = list(MELEE = 65, BULLET = 65, LASER = 60, ENERGY = 55, BOMB = 60, BIO = 50, "rad" = 50, FIRE = 55, ACID = 55)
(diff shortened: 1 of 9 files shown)


Jordanstolethepreciousslam:
And what about sec uniform? 5 armor is not a free tyr but is a big armor buff nontheless.

TiviPlus (approved):

DerpiestBruh:
The horrors!

Nickkoranda:
> PMC are the special response forces so they should have some more armored gear if you ask me.

The chest armor gains the armor from the uniform, no change to PMCs just marines getting the uniform.

lbnesquik:
Perhaps we should replace the PMC gear on the maps instead.

GravitronBlue:
Considering the goal of this PR consider changing the armor values of tactical jumpsuits, swat helmets, and possibly swat shoes found on bodies around the crashed ship on prison station.

Lumipharon:
Done. Shoes/helmets and the like don't matter since you have to replace your equipment for that so it's not just free armour stacking.

Jordanstolethepreciousslam:
Now this is based. But i would argue that the real problem is the way armor stacks. It shouldn't stack additively at all. Armor should stack multiplicatively. That would fix bullshit hand/feet armor and make targeting them viable and not a noob trap

FIXTHERADBOXPLEASE:
Senior SS13 server running expert here. This will kill TGMC in 3 months if merged.

Strayed-Raven:
If you do this. Add the uniforms to clothing vendor. Tis only fair. Otherwise I can see why reduction of armor.

PR #11175, opened 2022-10-09
LV 624 minor balance-related changes

Mannybrado:
## About The Pull Request

1. The weed node placement in most places in the northern caves is now significantly more uniform. Xenos no longer have to spend 5 minutes patching holes in the weeds in the hive at the beginning of the round, in an area that should already be completely covered.
2. The Blue Disk temple(NE one) is now less overgrown. There is more area for weeds to spread, and part of the hive is now extended into it. This should, hopefully, make it at least slightly desirable to hold for xenos... while it is over the river, the previous design was heavily favoring a marine push into it, as the grass made it impossible for xenos to effectively fortify.

image

## Why It's Good For The Game
LV sucks pretty bad right now, hopefully this should make it a little better, as xenos have a pretty hard time on this map. It is effectively THE "hit and run ambush map", but now that marines are able to keep the pace with xeno's speed and can take more damage, that is effectively nullified, leaving it in a spot where 40% of the map is uninhabitable for xenos with little benefit for trying to use the flanking routes. This is meant to make at least one of the disks realistically defendable for xenos, though this isn't really a massive change. Also, whoever did the weed placement originally had parkinsons or something.

## Changelog
:cl:
tweak: LV624 has better cave weed placement
tweak: LV624 blue disk temple is no longer ridiculously overgrown, as the xenos have done some groundkeeping.
/:cl:

### DIFF
--- a/_maps/map_files/LV624/LV624.dmm
+++ b/_maps/map_files/LV624/LV624.dmm
@@ -1803,11 +1803,12 @@
 /turf/open/ground/river,
 /area/lv624/ground/sand2)
 "aQc" = (
-/obj/effect/landmark/weed_node,
-/turf/open/floor/engine/cult{
-	dir = 2
+/obj/structure/bed/chair/comfy/black{
+	dir = 8
 	},
-/area/lv624/ground/caves/east1)
+/obj/effect/landmark/weed_node,
+/turf/open/floor/plating/ground/dirt,
+/area/lv624/lazarus/sandtemple)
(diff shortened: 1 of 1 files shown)


RipGrayson (changes requested):
I like the concept, I don't like the aesthetics. Please make a new type of weedable grass and use that instead of removing the jungle aesthetic inside the sand temple.
Also tweak isn't a proper CL change, please use balance instead.

PR #11182, opened 2022-10-09
Adds a new event, Queen Mothers blessing

RipGrayson:
## About The Pull Request

It's a new event, Tivi wants more of them and I sat down and made some.

The queen mother's blessing is a random boost to a single xeno, it heals, increases evolution points, and boosts a xeno to ancient (unless Primordial). You also get a shoutout to all of your sisters so they know how ~~lucky~~ skilled you are. 

## Why It's Good For The Game

More events good.

## Changelog
:cl:
add: Added a new event, queen mother's blessing.
/:cl:

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -103,6 +103,11 @@
 #define isxenohivemind(A) (istype(A, /mob/living/carbon/xenomorph/hivemind))
 #define isxenowraith(A) (istype(A, /mob/living/carbon/xenomorph/wraith))
 #define isxenowidow(A) (istype(A, /mob/living/carbon/xenomorph/widow))
+#define isbeetle(A) (istype(A, /mob/living/carbon/xenomorph/beetle))
+#define ismantis(A) (istype(A, /mob/living/carbon/xenomorph/mantis))
+#define isscorpion(A) (istype(A, /mob/living/carbon/xenomorph/scorpion))
+#define isspiderling(A) (istype(A, /mob/living/carbon/xenomorph/spiderling))
+#define isminion(A) (istype(A, /mob/living/carbon/xenomorph/spiderling)) || isscorpion(A) || ismantis(A) || isbeetle(A)
 
 //Silicon mobs
 #define issilicon(A) (istype(A, /mob/living/silicon))
@@ -255,7 +260,6 @@
 #define iscrashgamemode(O) (istype(O, /datum/game_mode/infestation/crash))
 #define isinfestationgamemode(O) (istype(O, /datum/game_mode/infestation))
 #define iscombatpatrolgamemode(O) (istype(O, /datum/game_mode/combat_patrol))
-#define issensorcapturegamemode(O) (istype(O, /datum/game_mode/combat_patrol/sensor_capture))
 
 #define isxenoresearcharea(A) (istype(A, /area/mainship/medical/medical_science))
 
--- a/code/modules/events/queen_mothers_blessing.dm
+++ b/code/modules/events/queen_mothers_blessing.dm
@@ -1,0 +1,36 @@
+#define QM_HEAL_AMOUNT 250
+
+/datum/round_event_control/queen_mothers_blessing
+	name = "Queen Mother's Blessing"
+	typepath = /datum/round_event/queen_mothers_blessing
+	weight = 15
+	earliest_start = 5 MINUTES
+
+	gamemode_blacklist = list("Combat Patrol","Civil War")
+
+/datum/round_event_control/queen_mothers_blessing/can_spawn_event(players_amt, gamemode)
+	if(length(GLOB.alive_xeno_list) <= 0)
+		return FALSE
+	return ..()
+
+/datum/round_event/queen_mothers_blessing/start()
+	for(var/mob/living/carbon/xenomorph/X in shuffle(GLOB.alive_xeno_list))
+		if(isminion(X))
+			continue
+		bless_xeno(X)
+		break
+
+/datum/round_event/queen_mothers_blessing/proc/bless_xeno(mob/living/carbon/xenomorph/X)
+	X.evolution_stored = X.evolution_stored + (X.evolution_stored * 1.75)
+	if(!isxenolarva(X) && (X.tier != XENO_UPGRADE_FOUR)) //larva do not have proper caste datums, trying to force one results in a runtime
+		X.upgrade_xeno(XENO_UPGRADE_THREE)
+	X.adjustBruteLoss(-QM_HEAL_AMOUNT)
+	X.adjustFireLoss(-QM_HEAL_AMOUNT, updating_health = TRUE)
+	X.adjust_sunder(-QM_HEAL_AMOUNT/20)
+	xeno_message("The Queen Mother has blessed [X], may they do great things for the hive.")
+	priority_announce("Sensors have detected a malevolent psychic force hovering over the battlefield, surely this doesn't bode well for us...")
+	var/sound/queen_sound = sound(get_sfx("queen"), channel = CHANNEL_ANNOUNCEMENTS)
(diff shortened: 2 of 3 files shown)


code/__DEFINES/is_helpers.dm
```diff
@@ -255,7 +260,6 @@
 #define iscrashgamemode(O) (istype(O, /datum/game_mode/infestation/crash))
 #define isinfestationgamemode(O) (istype(O, /datum/game_mode/infestation))
 #define iscombatpatrolgamemode(O) (istype(O, /datum/game_mode/combat_patrol))
-#define issensorcapturegamemode(O) (istype(O, /datum/game_mode/combat_patrol/sensor_capture))
```
ivanmixo, line 258: Probably didn't mean to do this.

code/modules/events/queen_mothers_blessing.dm
```diff
@@ -0,0 +1,36 @@
...
+/datum/round_event/queen_mothers_blessing/start()
+	for(var/mob/living/carbon/xenomorph/X in shuffle(GLOB.alive_xeno_list))
+		if(isminion(X))
+			continue
+		bless_xeno(X)
+		break
+
+/datum/round_event/queen_mothers_blessing/proc/bless_xeno(mob/living/carbon/xenomorph/X)
+	X.evolution_stored = X.evolution_stored + (X.evolution_stored * 1.75)
+	if(!isxenolarva(X) && (X.tier != XENO_UPGRADE_FOUR)) //larva do not have proper caste datums, trying to force one results in a runtime
```
ivanmixo, line 25: Shouldn't this be an or?

code/modules/events/queen_mothers_blessing.dm
```diff
@@ -0,0 +1,36 @@
+#define QM_HEAL_AMOUNT 250
+
+/datum/round_event_control/queen_mothers_blessing
+	name = "Queen Mother's Blessing"
+	typepath = /datum/round_event/queen_mothers_blessing
+	weight = 15
+	earliest_start = 5 MINUTES
+
+	gamemode_blacklist = list("Combat Patrol","Civil War")
```
ivanmixo, line 9: Should add the new gamemode to this list as well.

code/modules/events/queen_mothers_blessing.dm
```diff
@@ -0,0 +1,36 @@
...
+
+/datum/round_event/queen_mothers_blessing/start()
+	for(var/mob/living/carbon/xenomorph/X in shuffle(GLOB.alive_xeno_list))
+		if(isminion(X))
+			continue
+		bless_xeno(X)
+		break
+
+/datum/round_event/queen_mothers_blessing/proc/bless_xeno(mob/living/carbon/xenomorph/X)
+	X.evolution_stored = X.evolution_stored + (X.evolution_stored * 1.75)
```
ivanmixo, line 24:
```suggestion
	X.evolution_stored += X.evolution_stored * 1.75
```
Also this is 175% of their evolution stored and I'm 99% sure it doesn't go over 100%.

ivanmixo (changes requested):

code/modules/events/queen_mothers_blessing.dm
RipGrayson: Weirdly enough the evolution can go over 100% if forced, during testing I saw it hit 1000% percent after multiple forced consecutive uses of Queen Mother's blessing on a single xeno. It still resets to 0 after an evolution though, no matter how large the number.

code/modules/events/queen_mothers_blessing.dm
```diff
@@ -0,0 +1,36 @@
...
+	if(!isxenolarva(X) || (X.tier != XENO_UPGRADE_FOUR)) //larva do not have proper caste datums, trying to force one results in a runtime
```
DeltaFire15, line 25:
```suggestion
	if(!isxenolarva(X) && (X.tier != XENO_UPGRADE_FOUR)) //larva do not have proper caste datums, trying to force one results in a runtime
```
|| would result in any xeno that is not a larva being forced to upgrade three, including primordials, aswell as larvae counting as viable target.

Oscari-pog:
Does evolving take the blessing away?

code/modules/events/queen_mothers_blessing.dm
ivanmixo: Hmm true, should make this two if statements because if it's `&&` it'll never be true because you can't be primo and a larva at the same time if I'm not understanding it wrong.

DeltaFire15: && would mean "do thing if [not a larva] and also [not upgrade four]"
Aka
is larva + not upgrade four => 0
is larva + is upgrade four (somehow) => 0
not larva + not upgrade four => 1
not larva + is upgrade four => 0

code/modules/events/queen_mothers_blessing.dm
```diff
@@ -0,0 +1,37 @@
...
+	var/sound/queen_sound = sound(get_sfx("queen"), channel = CHANNEL_ANNOUNCEMENTS)
+	for(var/mob/living/carbon/xenomorph/receiving_xeno in GLOB.alive_xeno_list)
+		SEND_SOUND(receiving_xeno, queen_sound)
+	for(var/mob/living/carbon/xenomorph/X in shuffle(GLOB.alive_xeno_list))
+		if(isxenolarva(X))
+			continue
+		else if(isminion(X))
+			continue
+		bless_xeno(X)
+		break
```
Lumipharon, line 26: It probably doesn't matter, but I would personally have this as a return, as it's more consistant/clear that there's nothing left in the proc to do.

code/modules/events/queen_mothers_blessing.dm
```diff
@@ -0,0 +1,37 @@
...
+	gamemode_blacklist = list("Combat Patrol","Civil War","Sensor Capture")
+
+/datum/round_event_control/queen_mothers_blessing/can_spawn_event(players_amt, gamemode)
+	if(length(GLOB.alive_xeno_list) <= 4)
+		return FALSE
+	return ..()
+
+/datum/round_event/queen_mothers_blessing/start()
+	var/sound/queen_sound = sound(get_sfx("queen"), channel = CHANNEL_ANNOUNCEMENTS)
+	for(var/mob/living/carbon/xenomorph/receiving_xeno in GLOB.alive_xeno_list)
```
Lumipharon, line 18: It's extremely unlikely, but it's possible this event could trigger and there's no valid xenos (i.e. shrike hijacks then dies, and it's just a bunch of larva. It may be worth having this sound part trigger in the for loop, before the bless_xeno proc gets called, to ensure it only triggers the sound if the event successfully hits a beno.

Lumipharon (changes requested):

code/modules/events/queen_mothers_blessing.dm
```diff
@@ -0,0 +1,37 @@
...
+
+	gamemode_blacklist = list("Combat Patrol","Civil War","Sensor Capture")
+
+/datum/round_event_control/queen_mothers_blessing/can_spawn_event(players_amt, gamemode)
+	if(length(GLOB.alive_xeno_list) <= 4)
+		return FALSE
+	return ..()
+
+/datum/round_event/queen_mothers_blessing/start()
+	var/sound/queen_sound = sound(get_sfx("queen"), channel = CHANNEL_ANNOUNCEMENTS)
```
Lumipharon, line 17:
```suggestion
	var/sound/queen_sound = sound(get_sfx("queen"), channel = CHANNEL_ANNOUNCEMENTS, volume = 50)
```
100 is default, 50 is QM announcements.

Lumipharon (changes requested):

TiviPlus:
Lumi and mole are against as they believe numberschange is boring and something like hive threat is a more interactive version of these kind of buffs

Lumi also says they're making a few more events

PR #11188, opened 2022-10-10
you can no longer be moved when manning mounted guns by humans and a fix

tyeagg:
## About The Pull Request
you can no longer be moved while using mounted guns (only applies to marines xenos can still push and shuffle you) also removes being able to deploy deployables diagonally because you cannot get on the gun even if your behind it you have to click it diagonally, breaks aiming, and it breaks turning

## Why It's Good For The Game
pvt pisser shitter can no longer stop you from getting the epic frag by pushing you

## Changelog
:cl:
qol: can no longer be moved while manning mounted gun
fix: can no longer deploy mounted gun diagonally since it breaks stuff
/:cl:


FosterGrime-Heymynameslouie:
speed merhg instantly

PR #11194, opened 2022-10-10
Adds a new event, CAS supply increase

RipGrayson:
## About The Pull Request

It's a new event, Tivi wants more of them and I sat down and made some.

CAS supply increase looks at the total number of POs and adds between 50 and 175 points to the ammo fabricator per pilot (for reference, a single missile cost is around 200 points). The total number is capped at 300.

## Why It's Good For The Game

More events good.

## Changelog
:cl:
add: Added a new event, CAS supply increase
/:cl:

### DIFF
--- a/code/modules/events/cas_supply_increase.dm
+++ b/code/modules/events/cas_supply_increase.dm
@@ -20,7 +20,6 @@
 	for(var/mob/living/carbon/human/H in GLOB.alive_human_list_faction[FACTION_TERRAGOV])
 		if(ismarinepilotjob(H))
 			cas_points_to_be_added += rand(50,175)
-	if(cas_points_to_be_added > 300) //cap the added points to 300 in case rng really favors pilots
-		cas_points_to_be_added = 300
+	cas_points_to_be_added = min(cas_points_to_be_added , 300)
 	SSpoints.dropship_points += cas_points_to_be_added
 	priority_announce("Due to recycling of discarded ammo casings, our dropship fabricator has absorbed enough material to increase capacity by [cas_points_to_be_added] points.")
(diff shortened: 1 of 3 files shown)


code/modules/events/cas_supply_increase.dm
```diff
@@ -0,0 +1,26 @@
...
+/datum/round_event_control/cas_supply_increase/can_spawn_event(players_amt, gamemode)
+	if(length(GLOB.alive_human_list_faction[FACTION_TERRAGOV]) < 25)
+		return FALSE
+	if(SSpoints.dropship_points >= 250)
+		return FALSE
+	return ..()
+
+/datum/round_event/cas_supply_increase/start()
+	var/cas_points_to_be_added //var to keep track of how many point we're adding to dropship
+	for(var/mob/living/carbon/human/H in GLOB.alive_human_list_faction[FACTION_TERRAGOV])
```
Lumipharon, line 20: Is it intentional that a PO being temporarily dead means you receive less CAS points?

code/modules/events/cas_supply_increase.dm
```diff
@@ -0,0 +1,26 @@
...
+		if(ismarinepilotjob(H))
+			cas_points_to_be_added += rand(50,175)
+	if(cas_points_to_be_added > 300) //cap the added points to 300 in case rng really favors pilots
+		cas_points_to_be_added = 300
```
Lumipharon, line 24:
```suggestion
	cas_points_to_be_added = min(cas_points_to_be_added , 300)
```

Lumipharon (changes requested):

code/modules/events/cas_supply_increase.dm
RipGrayson: Yes, this is intentional. The marines shouldn't be rewarded for Tad suicide missions (which is really the only thing short of a LZ breach that would kill a PO).

PR #11195, opened 2022-10-10
Lets SR-127 fit barrel charger

Losenis:
## About The Pull Request
Title

## Why It's Good For The Game
Trade extended barrel's accuracy bonus for more bullet speed and an accuracy malus. DMR already can fit it while having the same range, nobody seems to complain.

## Changelog
:cl:
balance: SR-127 can now fit barrel charger
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1184,6 +1184,7 @@
 		/obj/item/attachable/scope/unremovable/tl127,
 		/obj/item/attachable/stock/tl127stock,
 		/obj/item/attachable/extended_barrel,
+		/obj/item/attachable/heavy_barrel,
 		/obj/item/attachable/suppressor,
 		/obj/item/attachable/bayonet,
 		/obj/item/attachable/bayonetknife,


KrabSpider:
> DMR already can fit it while having the same range, nobody seems to complain.
Because it isn't a sniper and it doesn't have a jacked shell speed. It actually **loses** accuracy and the tradeoff is only barely worth it if you're trying to actually snipe long range targets. 

> Trade extended barrel's accuracy bonus for more bullet speed and an accuracy malus.
Did you check the stats when you attached the BC to the sniper? The reduction still leaves with 50% accuracy. This is barely a downside.

Al-1ce:
> Because it isn't a sniper and it doesn't have a jacked shell speed. It actually **loses** accuracy and the tradeoff is only barely worth it if you're trying to actually snipe long range targets


You've clearly never used the DNR with the intent to snipe. BC+Bipod makes it better than the TL-127 by several bounds. It's only fair.

KrabSpider:
> You've clearly never used the DMR with the intent to snipe. 

Correction; DMR bullets don't travel nearly as fast as sniper bullets and the accuracy malus is actually present when a BC is attached to it. 

> BC+Bipod makes it better than the TL-127 by several bounds.

Your bullets are still slower than sniper bullets, have far less AP, and don't have nearly as sunder as one. No the fuck it isn't.

Lumipharon:
Snipers have projectile speed of 4. DMR and other rifles have 3.
EB increases this by 1, BC by 2.
So BC DMR is the same projectile speed as an EB 127.

Snipers also don't have any damage falloff where as the DMR does (only 0.25 per turf, but at long range that'll make a bit of a difference.

Bother guns are extremely accurate, with high accurate range, and a large acc boost from scoping, so the accuracy penalty from a BC means precisely fuck all.

TiviPlus (approved):

Al-1ce:
> Correction; DMR bullets don't travel nearly as fast as sniper bullets and the accuracy malus is actually present when a BC is attached to it. 

As mentioned by Lumi, the malus is pretty much non-existent. 

>Your bullets are still slower than sniper bullets, have far less AP, and don't have nearly as sunder as one. No the fuck it isn't.

The difference between the speed doesn't matter when you're able to shoot 3x-2x faster. Even if it was as slow as you think it is, it wouldn't matter. The difference isn't night and day. So, I think giving the SR-127 the BC would allow it to compete to a fairer extent.

PR #11198, opened 2022-10-11
stop drop and roll no longer stuns you for 10 seconds

tyeagg:
## About The Pull Request
stop drop and rolling now makes you get stunned for 5 seconds from 10 seconds

## Why It's Good For The Game
fire damage does allot of pain and stop drop and rolling doesn't remove all the fire stacks so 10 seconds of being stunned and the fire still being there because you didn't bring an extinguisher is not fun

## Changelog
:cl:
balance: stop drop and roll only takes 5 seconds now
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -200,7 +200,7 @@
 /mob/living/proc/resist_fire(datum/source)
 	SIGNAL_HANDLER
 	fire_stacks = max(fire_stacks - rand(3, 6), 0)
-	Paralyze(80)
+	Paralyze(10)
 
 	var/turf/T = get_turf(src)
 	if(istype(T, /turf/open/floor/plating/ground/snow))
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -200,7 +200,7 @@
 /mob/living/proc/resist_fire(datum/source)
 	SIGNAL_HANDLER
 	fire_stacks = max(fire_stacks - rand(3, 6), 0)
-	Paralyze(10)
+	Paralyze(30)
 
 	var/turf/T = get_turf(src)
 	if(istype(T, /turf/open/floor/plating/ground/snow))


MLGTASTICa:
But fire is super supressive for xenos too ,i think its fair game whenever a marine gets thrown into their own fire / pushed and now has to deal with the fire damage.
It also lets people optimize loadouts further to only be for killing, which i believe is a bad thing. People should trade space for flamethrowers if they want to not get burned.In order to keep the gameplay more "diverse" and tactic.

Losenis:
There's a good few xenos with their own fire extinguishers (Acid wells), they have resin jelly, and their resist on fire only stuns then for a second. They aren't affected by pain either. The stun for dropping and rolling is so long that in some cases you're actually better off not extinguishing at all and just taking kelo tricord while backing off at the relative pace of a leisure walk.

Losenis:
I actually found the video

https://user-images.githubusercontent.com/72712909/195152370-26114c7c-47f6-4c7c-a06b-c2c93d1c4997.mp4

MLGTASTICa:
> There's a good few xenos with their own fire extinguishers (Acid wells), they have resin jelly, and their resist on fire only stuns then for a second. They aren't affected by pain either. The stun for dropping and rolling is so long that in some cases you're actually better off not extinguishing at all and just taking kelo tricord while backing off at the relative pace of a leisure walk.
> 
> Either you're alone, in which case you ABSOLUTELY DONT WANT TO GET STUNNED, or you're in a group, in which case you can just have the others put you out with their own hands.

It doesn't change the fact that acid wells provide marines with free lightning , resin jelly barely lasts for more than 1 minute , and that fire is terribly effective against xeno mazes. I think it is in a good spot as it is right now , with flamers being able to extinguish themselves if they fuck up (or other marines who do not have situational awareness).

Jordanstolethepreciousslam:
Literally a crutch for bald marines to free their only busy braincell from using extinguisher

Losenis:
"Get over it fire doesn't even hurt that much" -Flamer wiping the entire push #7298

KrabSpider:
10 seconds is a stupidly long time, especially for an action thats meant to keep you from dying. Kind of defeats the point.

Araulius:
learn to extinguisher

Starfirejordan:
make it 5-6 seconds and call it a day

lbnesquik:
Middle ground seems to be the better option imo.

tyeagg:
adjusted to 5 seconds

code/modules/mob/living/living_defense.dm
```diff
@@ -200,7 +200,7 @@
 /mob/living/proc/resist_fire(datum/source)
 	SIGNAL_HANDLER
 	fire_stacks = max(fire_stacks - rand(3, 6), 0)
-	Paralyze(80)
+	Paralyze(30)
```
BraveMole, line 203:
```suggestion
	Paralyze(3 SECONDS)
```

BraveMole (changes requested):

code/modules/mob/living/living_defense.dm
tyeagg: apparently this is the amount of paralyze whatever the fuck that means not the actual length of paralyze but its basically +20/10 length of the set number

PR #11208, opened 2022-10-11
Adds new AI core icons

Lumipharon:
## About The Pull Request
Adds some new AI core icons, thanks to GregorMN.
Also used his frame on the existing core icons, since the old black and white ones look like ass.

Note, there are a metric butt load of AI states that have a gorillion frames that I need to manually adjust since they're all 1 pixel too low. So rather than doing them all now, I'm PRing this and asking AI players/maintainers which ones to just trash.

Shodan_1

## Why It's Good For The Game
Better sprites are better

## Changelog
:cl:
imageadd: added some new AI core sprites, and updated some old ones
/:cl:


Lumipharon:
This is technically WIP until I update/remove some of the old ones, I forgot to set the PR to WIP though and uh, forgot how to change it.

Kuro020 (approved):

Derrick2Electricboogaloo:
I was actually wondering if you would be amenable to adding a sheriffs badge icon for the AI? 
image
Something similar to this or exactly that.

TiviPlus (approved):

lbnesquik:
Actually, here's a better solution, rather than having us settle for one 'core' design. Why don't we have all frames as something variable? We rip the surrounding screen bits, and then we have the animation under it. That would make it easy for us to add more frames in the future, and give a bit of customization to the role with the least ATM.

Lumipharon:
They're not consistant with the area they use for the screen (a couple of them even have different sized frames. But I reverted the ones where there was any cropping. Someone can figure out what they want to do with those, I ain't coding in multiple frame options, especially considering the screen locations don't even align currently.

PR #11218, opened 2022-10-12
Surt Buff

Naaanii:
## About The Pull Request
Buffs Surt in a meaningful way.

## Why It's Good For The Game
Surt is a very underwhelming module and comparable to just running in through fire and extinguishing yourself only getting the prevention of ignition and a lot of slowdown.
This should help Surt users take a lot more fire damage but users are also still not completely immune to the effects of fire.

## Changelog
:cl:
balance: Adds 200 soft fire armour to Surt.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -58,7 +58,7 @@
 	desc = "Designed for mounting on modular armor. Providing a near immunity to being bathed in flames, and amazing flame retardant qualities, this is every pyromaniacs' first stop to survival. Will impact mobility."
 	icon_state = "mod_fire"
 	item_state = "mod_fire_a"
-	soft_armor = list(FIRE = 200)
+	soft_armor = list(FIRE = 50)
 	hard_armor = list(FIRE = 200)
 	slowdown = 0.4
 	slot = ATTACHMENT_SLOT_MODULE
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -58,8 +58,8 @@
 	desc = "Designed for mounting on modular armor. Providing a near immunity to being bathed in flames, and amazing flame retardant qualities, this is every pyromaniacs' first stop to survival. Will impact mobility."
 	icon_state = "mod_fire"
 	item_state = "mod_fire_a"
-	soft_armor = list("fire" = 200)
-	hard_armor = list("fire" = 200)
+	soft_armor = list("FIRE" = 200)
+	hard_armor = list("FIRE" = 200)
 	slowdown = 0.4
 	slot = ATTACHMENT_SLOT_MODULE
 	variants_by_parent_type = list(/obj/item/clothing/suit/modular/xenonauten = "mod_fire_xn", /obj/item/clothing/suit/modular/xenonauten/light = "mod_fire_xn", /obj/item/clothing/suit/modular/xenonauten/heavy = "mod_fire_xn")
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -58,6 +58,7 @@
 	desc = "Designed for mounting on modular armor. Providing a near immunity to being bathed in flames, and amazing flame retardant qualities, this is every pyromaniacs' first stop to survival. Will impact mobility."
 	icon_state = "mod_fire"
 	item_state = "mod_fire_a"
+	soft_armor = list("fire" = 200)
 	hard_armor = list("fire" = 200)
 	slowdown = 0.4
 	slot = ATTACHMENT_SLOT_MODULE
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -58,8 +58,8 @@
 	desc = "Designed for mounting on modular armor. Providing a near immunity to being bathed in flames, and amazing flame retardant qualities, this is every pyromaniacs' first stop to survival. Will impact mobility."
 	icon_state = "mod_fire"
 	item_state = "mod_fire_a"
-	soft_armor = list("FIRE" = 200)
-	hard_armor = list("FIRE" = 200)
+	soft_armor = list(FIRE = 200)
+	hard_armor = list(FIRE = 200)
 	slowdown = 0.4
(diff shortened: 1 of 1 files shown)


MereWalker:
What about the suit light? There is practically no light save from your flamer and whatever optics you can find.

Naaanii:
> What about the suit light? There is practically no light save from your flamer and whatever optics you can find.

Suit light was already buffed previously and does not need a further buff imo.
https://github.com/tgstation/TerraGov-Marine-Corps/pull/9753

Naaanii:
> > This should help Surt users take a lot more fire damage but users are also still not completely immune to the effects of fire.
> 
> They still get damage?

Yes but its greatly reduced. They are however not completely immune.

DeltaFire15:
Did you test this PR? If I remember fire code right this just does nothing because hard fire armor's handing of "take a bit of fire damage but don't do anything else" bypasses any normal armor and you left the hard armor in..
And if it did work 200 soft fire armor would indeed mean they'd completely ignore fire because thats how soft armor works.

Dgeen:
> Yes but its greatly reduced. They are however not completely immune.

Then i think i can help. One person from rutgmc conducted a study on fire damage, then I will give his text (Google translate, I hope it will be clear)

> Melancholy helped me to learn that no fireproof boots play a role in protecting against fire damage through the suit.
After adding gloves with 300 firepruff armor to the surt, the damage from ordinary fire ceased to pass at all, throwing only pain. The damage from X fuel still penetrates the suit with gloves.

Naaanii:
> Did you test this PR? If I remember fire code right this just does nothing because hard fire armor's handing of "take a bit of fire damage but don't do anything else" bypasses any normal armor and you left the hard armor in.. And if it did work 200 soft fire armor would indeed mean they'd completely ignore fire because thats how soft armor works.

I did. You are not immune to it and take a greatly reduced amount of fire damage. Didn't account for meds but the damage is standard across fire and blue fire too.

Naaanii:
> > Yes but its greatly reduced. They are however not completely immune.
> 
> Then i think i can help. One person from rutgmc conducted a study on fire damage, then I will give his text (Google translate, I hope it will be clear)
> 
> > Melancholy helped me to learn that no fireproof boots play a role in protecting against fire damage through the suit.
> > After adding gloves with 300 firepruff armor to the surt, the damage from ordinary fire ceased to pass at all, throwing only pain. The damage from X fuel still penetrates the suit with gloves.
> 
> So, if you want full immune to fire, you need add surt gloves to a complete.

I will ask Maints and see if they agree that full immunity is okay.

Dgeen:
You still have full immune if you eat kelotram pill. Even with current surt

DeltaFire15:
> I did. You are not immune to it and take a greatly reduced amount of fire damage. Didn't account for meds but the damage is standard across fire and blue fire too.

effective overall soft armor with surt at 200 soft is 194 which means standing on fire does nothing since that uses overall armor.
Stepping into fire however applies 0.2 x base damage to each individual limb which is modified by armor. This in return means it's still immunity for all limbs except head, whereas the head "only" gets ~50% additional reduction from the helmet. So, crossing fire deals 0.009x base damage, as damage exclusively to the head.

Naaanii:
> > I did. You are not immune to it and take a greatly reduced amount of fire damage. Didn't account for meds but the damage is standard across fire and blue fire too.
> 
> effective overall soft armor with surt at 200 soft is >=192 which means standing on fire does nothing since that uses overall armor. Stepping into fire however applies 0.2 x base damage split over each individual limb which is modified by armor. This in return means it's still immunity for all limbs except head, whereas the head "only" gets ~50% additional reduction from the helmet (assuming you don't find one with better than that). So, crossing fire deals 0.009x base damage, as damage exclusively to the head.

Yeah, only head was taking burn damage while testing. Could make a head module for full immunity but unsure if I should or not yet.

Al-1ce:
I feel like surt taking damage from fire to begin with has always been contradictory and not really intuitive for gameplay. There's the argument about flamers being "able to push nonstop", but having to eventually refuel already compensates for that.

I won't even mention the pain of having a broken bone with Surt.

Autisem:
I think its fine, I dont think infinte push is a problem because the fire still hurts everyone else. so if a pyro is going to walk in a sea of flames he will propably get sniped by spitters or just mauled by fire resistant jelly xenos while marines cant really get to them to save him(especialy relevant for blueflames that last alot longer)

code/modules/clothing/modular_armor/attachments/modules.dm
```diff
@@ -58,6 +58,7 @@
 	desc = "Designed for mounting on modular armor. Providing a near immunity to being bathed in flames, and amazing flame retardant qualities, this is every pyromaniacs' first stop to survival. Will impact mobility."
 	icon_state = "mod_fire"
 	item_state = "mod_fire_a"
+	soft_armor = list("fire" = 200)
```
Lumipharon, line 61:
```suggestion
	soft_armor = list(FIRE = 200)
```
use defines. Hard armour should be the same while you're at it.

Lumipharon (changes requested):

code/modules/clothing/modular_armor/attachments/modules.dm
Naaanii: Done

lbnesquik:
I don't think this Pr requires a test merge

Derrick2Electricboogaloo:
Give the people what we want. We want a surt that works! We are the 99%

Derrick2Electricboogaloo:
> I feel like surt taking damage from fire to begin with has always been contradictory and not really intuitive for gameplay. There's the argument about flamers being "able to push nonstop", but having to eventually refuel already compensates for that.
> 
> I won't even mention the pain of having a broken bone with Surt.

And not to mention the mobility penalty, and being way out of position and getting killed is a big factor of not pushing indefinitely. No one without surt will follow you into the fire. No one reasonable that is.

Mannybrado:
> There's the argument about flamers being "able to push nonstop", but having to eventually refuel already compensates for that.

No. No, it doesn't. If you use a flamer pack you will be able to push and push and push until the xenos have no more hive left. This will probably be heavily abused as it was before surt was nerfed, only for it to be nerfed again.

This is going to end poorly. Fire's use is area denial but the downside is that it also denies the area for marines; completely removing that downside, as shown a very long time ago, leads to some incredibly stupid gameplay.

TiviPlus:
No from me until https://github.com/tgstation/TerraGov-Marine-Corps/pull/11276, which should be changing the numbers anyway

PR #10921, opened 2022-09-11
X fuel exploit fix

Lumipharon:
## About The Pull Request
You can no longer refill X fuel using regular fuel tanks, for infinite X fuel.
Also if you try fill a full tank, it now says it's full, instead of saying it was refilled and making the refuel sound, because that was annoying.

## Why It's Good For The Game
Exploit bad.

## Changelog
:cl:
fix: X fuel tanks no longer convert ordinary fuel to X fuel
/:cl:

PR #11220, opened 2022-10-12
Adds X fuel tank to req. Replaces incinerator tanks in req with large tanks.

Naaanii:
## About The Pull Request
Adds X fuel large tanks to req for 30 points.
Removes incinerator tanks from req and replaces them with the regular large tanks for double the cost (3>6).
Amount of large flamethrower tanks in vendors reduced from 30 to 16.

## Why It's Good For The Game

X fuel is only available in the form of a back tank. This adds the regular X fuel tanks to req for half the price of a back fuel tank.

There's never a reason to order incinerator tank with the lesser fuel it provides and tanks being refillable to an infinite amount and its greater versions being available roundstart.

The amount of initial large tanks available have been reduced now that they're orderable in req.

## Changelog
:cl:
add: Added X tank to requisitions for 30 points.
balance: Replaced incinerator tanks with large flame tank for higher cost (6 points).
balance: Reduced starting availability of large flame tanks from 30 to 16.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -118,7 +118,7 @@
 			/obj/item/weapon/gun/rifle/pepperball = 4,
 			/obj/item/ammo_magazine/rifle/pepperball = 40,
 			/obj/item/weapon/gun/flamer/big_flamer/marinestandard = 4,
-			/obj/item/ammo_magazine/flamer_tank/large = 30,
+			/obj/item/ammo_magazine/flamer_tank/large = 16,
 			/obj/item/ammo_magazine/flamer_tank/backtank = 4,
 			/obj/item/jetpack_marine = 3,
 			/obj/item/bodybag/tarp = 10,
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -471,6 +471,11 @@ WEAPONS
 	contains = list(/obj/item/ammo_magazine/flamer_tank)
 	cost = 3
 
+/datum/supply_packs/weapons/napalm_X
+	name = "FL-84 X fuel tank"
+	contains = list(/obj/item/ammo_magazine/flamer_tank/large/X)
+	cost = 30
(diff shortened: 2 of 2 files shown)


maikilangiolo:
30 points for a single use 76 ammo tank, when the 500 fuel one is 60. Because it's single use right? If it can be refilled, then it's an exploit.

Al-1ce:
> 30 points for a single use 76 ammo tank, when the 500 fuel one is 60. Because it's single use right? If it can be refilled, then it's an exploit.

It being refiled is intentional. Same with regular X-FUEL. That's why this PR sort of confuses me.

Al-1ce:
A single container of x-fuel which can be infinitely refilled for a fraction of the price is blatant powercreep. If marines want x-fuel, they should have to purchase the real thing.

I seriously can't see this being used outside of an easy way to steamroll hives with something you normally wouldn't have that early into the match.

Autisem:
I do feel like x fuel is better balanced when you also add the commitment if having to sacrifice back slot. but maby im wrong I dont see it used much anyway

Naaanii:
> A single container of x-fuel which can be infinitely refilled for a fraction of the price is blatant powercreep. If marines want x-fuel, they should have to purchase the real thing.
> 
> I seriously can't see this being used outside of an easy way to steamroll hives with something you normally wouldn't have that early into the match.

Could just remove the refuel and add a seperate req purchasable welder pack for X-fuel tbh.

lbnesquik:
Didn't #10921 fix the refueling issues already?

Naaanii:
> Didn't #10921 fix the refueling issues already? If you can't make X-fuel appear out of thin air, the smaller tanks should really be cheaper.

Only fixes issues with the back tank.

Naaanii:
I could make it unrefillable, but I feel this would be a nicer refillable option if needed. Would be willing to increase the price too.
image

Naaanii:
Refuel fix has been PRd.
https://github.com/tgstation/TerraGov-Marine-Corps/pull/11226

PR #11221, opened 2022-10-12
Fix ravage

TiviPlus:
## About The Pull Request
very ez fix
## Changelog
:cl:
fix: fixed ravage not working on humans
/:cl:


Naaanii:
But why is it good for the game? kek

DerpiestBruh:
rav mains go home

BraveMole (changes requested):
What about xenos from other hives

TiviPlus:
> What about xenos from other hives

Can we have someone else do this I dont feel like doing it and it didnt work before either

PR #11222, opened 2022-10-13
Volkite resprite and minor balance changes

Lumipharon:
## About The Pull Request
Resprites the Charger, Caliver and Culverin. I'm, lazy and don't feel like doing the serpenta right now.

All the other SOM weapons are made by competent spriters, so the volkite sprites I made were prety shoddy in comparison. They're still probably not that great, but it should be less jarring now.

image

Mob sprites now have loaded/unloaded states, and volkite and high cap volkite cells have been consolidated into a single cell type (with a new sprite). i.e. both the charger and caliver now use the same cell, similar to how all marine lasers use the same cell.

One balance change is that the charger now has 45 shots instead of 40, so it actually gets more shots than the more powerful caliver.

Also a couple of minor graphics bugs.
RPG now correctly shows it's sprite when stored in your suit
Guns now properly update their icon at init if they start unloaded.

## Why It's Good For The Game
Better sprites is better
Single cell makes logistics/scavaging less of a headache
Charger having more shots because it's weaker makes sense and is better for balance.

## Changelog
:cl:
balance: Volkite chargers and calivers now use the same volkite power cell
balance: Volkite chargers have 45 shots per mag
fix: fixed missing RPG sprites if stored in suit storage
fix: fixed guns not updating their icon state correctly on init if they don't start loaded
imageadd: Resprited the volkite charger, caliver and culverin, as well as volkite power cell
/:cl:


Al-1ce:
Pretty sure the ```high capacity volkite energy cell``` is only intended, or able to be used for the Serpenta. 

As for the ammo change, the Culverin was intended to have a higher capacity than the charger. I don't see why increasing one or either is all that neccesary.

Lumipharon:
serpenta uses a smaller cell specific to it. the charger ammo increase I was already going to do, it just made sense to do it in this pr.

Kuro020 (approved):

PR #11224, opened 2022-10-13
Removes some bolted/welded varedits on Gelida

RipGrayson:
## About The Pull Request

Apparently some varedits slipped past me while porting Gelida, and people didn't like the bolts/welds. This PR removes them.

## Why It's Good For The Game

If you ask people what they hate about Gelida, usually they'll bring up bolted doors. Removing them should be a positive.

## Changelog
:cl:
balance: Gelida no longer has bolted or welded doors at roundstart.
/:cl:


ivanmixo:
Think just removing the whole lines will do the same instead of setting them to 0

PR #11226, opened 2022-10-13
Fixed welderpack/kit still refueling xfuel tanks

Lumipharon:
## About The Pull Request
What it says on the tin.

So refuel code is pretty spaghetti, welder kits, welder packs, and welding tanks all run their refuel code in different places, adds the missing checks to the two bags, although long term they could probably use a refactor to just be generally less jank.

Also added a missing check welder kits so they don't appear to refuel full tanks.

## Why It's Good For The Game
Exploits bad.

## Changelog
:cl:
fix: fixed both kinds of welder bag still refilling xfuel tanks
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -707,6 +707,9 @@
 
 	else if(istype(I, /obj/item/ammo_magazine/flamer_tank))
 		var/obj/item/ammo_magazine/flamer_tank/FT = I
+		if(FT.default_ammo != /datum/ammo/flamethrower)
+			to_chat(user, span_warning("Not the right kind of fuel!"))
+			return ..()
 		if(FT.current_rounds == FT.max_rounds || !reagents.total_volume)
 			return ..()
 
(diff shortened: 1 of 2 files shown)


Losenis:
I thought they couldn't refuel backtanks at all regardless of xfuel or not.

FosterGrime-Heymynameslouie:
> I thought they couldn't refuel backtanks at all regardless of xfuel or not.

they can refuel normal flamer fuel but supposedly not xfuel

wondering how xfuel will be refueled now since its a 60 point a pop tank

Losenis:
I said backtank, I know they can refuel normal flamer tank but not the one that goes on backpack slot which is also the xfuel's type.

Al-1ce:
Is a tank of x-fuel not intended to be refueled? That doesn't make sense at all.

maikilangiolo:
> I said backtank, I know they can refuel normal flamer tank but not the one that goes on backpack slot

Are you saying you can't refill normal flamer back tanks now, too?

Lumipharon:
It's a bug, it was fixed, shitcode just meant it wasn't fixed for all items.

PR #11247, opened 2022-10-14
Coverage change - HvH mainly

Lumipharon:
## About The Pull Request
Set the barricade clear distance for bullets, energy projectiles and rockets (i.e. no xeno projectiles) to 2 instead of 1.
This means you can safely shoot past any object 2 tiles away instead of having to be right next to it.

This should have minimal if any impact on HvX, and is mainly aimed at HvH modes.

## Why It's Good For The Game
This should make using cover a lot easier in practice when you're active moving around, or shooting at diagonals (like shooting out of a 2-3 tile wide window). 
It also means you can take advantage of layers of cover, for example many maps have 2x2 areas of crates and other objects, which previously would always atleast partially block your bullets.

## Changelog
:cl:
balance: guns can now shoot past obstacles up to 2 tiles away instead of 1
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -304,6 +304,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	damage = 10
 	shrapnel_chance = 10
 	bullet_color = COLOR_VERY_SOFT_YELLOW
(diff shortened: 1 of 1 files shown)


Mannybrado:
should be increased for xeno spit as well. you literally said the scenario yourself-- there are many places on the map with 2xW/E sized obstructions that marines would now be able to shoot over and xenos couldn't.

Issue #11244, opened 2022-10-14
There is no beer 


Edelor68:
### Description

You can not dispense beer from the booze dispenser. Tried with multiple people and multiple different dispensers.
Every other beverage works as normal, just beer wont work for some reason

### Test Merges

[logs/sybil/server_side_modifications.dm](https://tgstation13.org/parsed-logs/sybil/server_side_modifications.dm)

Round ID: [17191](https://scrubby.melonmesa.com/round/17191)
BYOND Version: 514.1585
Server revision compiled on: 
Master commit: [3bb051ba922305abb667b07f59e39b5e379de56f](https://github.com/tgstation/TerraGov-Marine-Corps//commit/3bb051ba922305abb667b07f59e39b5e379de56f)
The following pull requests are currently test merged:
[#11056: 'Intro Sequence Or Something Idk What To Call It' by tyeagg at commit b06ecc9b13](https://github.com/tgstation/TerraGov-Marine-Corps//pull/11056)
[#11117: 'Investigation for the crashes. DO NOT MERGE!' by hyper2snyper at commit c069b98506](https://github.com/tgstation/TerraGov-Marine-Corps//pull/11117)

Local commit: 82be73307205f3692e8657c7c0651ae6bc1f3476
Server tools version: 5.2.2
[Common (Mar)] LCPL Kanata

### Reproduction Steps

1. Go to booze dispenser
2. Click on beer
3. Nothing happens

helper-2:
Literally unplayable.

PR #11251, opened 2022-10-15
Fixes beer not being dispensable

RipGrayson:
## About The Pull Request
Title.
Closes https://github.com/tgstation/TerraGov-Marine-Corps/issues/11244.
There's a mixup in the beer datum with beer2, causing a check to fail and the beer to not be dispensed. Because beer2 is a toxic knockout drink that has no place in TGMC I removed it entirely.

## Why It's Good For The Game

Bugs bad fixes good.

## Changelog
:cl:
fix: The booze dispenser can dispense beer again.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -328,22 +328,6 @@
 	L.reagent_pain_modifier = volume
 	return ..()
 
-/datum/reagent/toxin/beer2	//disguised as normal beer
-	name = "Beer"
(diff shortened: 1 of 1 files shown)


QualityVan:
This is a lot simpler than the list management fix I tried out a year ago and should be fine. It doesn't fix the underlying "problem", but honestly the conditions to trigger it (two reagents with the same ID) should be avoided anyway.

Derrick2Electricboogaloo:
That seems odd because the list does not show the toxin reagent. I am seeing the ethanol/beer.

Derrick2Electricboogaloo:
Well I will be damned, it worked. Not sure how that makes sense, but it works.

PR #11261, opened 2022-10-15
Removes the MP5 scatter

lbnesquik:
## About The Pull Request
~~The MP5 has a pretty bad middle ground right now. Slightly buffing the fire rate should make it a bit more competitive. ~~
I replaced the fire rate increase with scatter removal. 

## Why It's Good For The Game
The gun will be slightly more fun to use, and less of a dead weight. 

## Changelog
:cl:
balance: Removed basic MP5 scatter
/:cl:


Derrick2Electricboogaloo:
Firerate for which fire mode, or just in general?

MLGTASTICa:
its not a multiple of 0.05 so its actually 0.15 , which from 0.20 is a 25% decrease.

Lumipharon:
Pretty sure it rounds up, so no change at all.

Also T90 is 3 shots per 0.45, MR25 is 3 shots per 0.5.
MR25 has better burst damage and higher capacity. It also has some better attachment options.

It's not exactly in a bad place.

Derrick2Electricboogaloo:
> Pretty sure it rounds up, so no change at all.
> 
> Also T90 is 3 shots per 0.45, MR25 is 3 shots per 0.5. MR25 has better burst damage and higher capacity. It also has some better attachment options.
> 
> It's not exactly in a bad place.

Definitely agree. I am curious if this will make the full auto better or not, since full auto is not exactly that good. Atleast in my experience. Burst on it is perfect.

lbnesquik:
> its not a multiple of 0.05 so its actually 0.15 , which from 0.20 is a 25% decrease.

Oh, that's not quite what I wanted. But should I change it to another number?

lbnesquik:
> > Pretty sure it rounds up, so no change at all.
> > Also T90 is 3 shots per 0.45, MR25 is 3 shots per 0.5. MR25 has better burst damage and higher capacity. It also has some better attachment options.
> > It's not exactly in a bad place.
> 
> Definitely agree. I am curious if this will make the full auto better or not, since full auto is not exactly that good. Atleast in my experience. Burst on it is perfect.

I have to admit, I don't exactly know how fire rates work in this code base. I wanted to make the full auto better, mostly. I'll see about that and come back.

code/modules/projectiles/guns/smgs.dm
```diff
@@ -190,7 +190,7 @@
 	accuracy_mult = 1.1
 	accuracy_mult_unwielded = 0.95
 	scatter = 4
-	fire_delay = 0.2 SECONDS
+	fire_delay = 0.18 SECONDS
```
Pariah919, line 193:
```suggestion
	fire_delay = 0.18 SECONDS
```
This is rounded up to my knowledge, even if you were making it fire fast I'd rather recommend you remove the scatter entirely and give it more attachment options, the entire gimmick of the MP5 is SMG stats with rifle fire rate and attachment options.

lbnesquik: I didn't know that was its gimmick. I didn't even know it had one, I thought someone thought it was fun to add it, and there we got.  I'll see about making some changes this week.

Lumipharon:
This uh, literally does nothing. The base scatter for all guns is 4, so you're not actually changing it.

I don't think this gun needs any changes, but if you're trying to reduce scatter, you should reduce the burst scatter multiplier by 1.

TiviPlus:
Seasonal gun micro-tweaking I dont see this as needed

PR #11269, opened 2022-10-16
Runners and Spiderlings will now bump UAVs instead of going over them

### LINKED ISSUE #11211: Runner cannot bump slash UV-L Iguana
### Description

Other castes can bump slash UV-L Iguana, but not runner. Runner just phase through.

### Test Merges

NA

### Reproduction Steps

1. Get UV-L Iguana
2. Get runner
3. Runner phrases through
4. ????
5. Get drone
6. Drone bump slashes UV-L Iguana
7. okay

### Screenshots

https://www.youtube.com/watch?v=mZu5ZlhcXuY

lKiseki:
## About The Pull Request
Per title. Fixes #11211, even though it's not strictly a fix and also a balance change.

## Why It's Good For The Game
Allows these specific castes to contest UAVs instead of being screwed over.
 
## Changelog
:cl: Lewdcifer
balance: Xenos such as the Runner and the Spiderling will now bump into UAVs instead of going over them.
/:cl:

### DIFF
--- a/code/modules/vehicles/unmanned/unmanned_vehicle.dm
+++ b/code/modules/vehicles/unmanned/unmanned_vehicle.dm
@@ -41,7 +41,7 @@
 	/// If something is already controlling the vehicle
 	var/controlled = FALSE
 	/// Flags for unmanned vehicules
-	var/unmanned_flags = OVERLAY_TURRET|HAS_LIGHTS
+	var/unmanned_flags = OVERLAY_TURRET|HAS_LIGHTS|UNDERCARRIAGE
 	/// Iff flags, to prevent friendly fire from sg and aiming marines
 	var/iff_signal = TGMC_LOYALIST_IFF
 	/// muzzleflash stuff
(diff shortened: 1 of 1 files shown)


TiviPlus:
Liguana

code/modules/vehicles/unmanned/unmanned_vehicle.dm
```diff
@@ -41,7 +41,7 @@
 	/// If something is already controlling the vehicle
 	var/controlled = FALSE
 	/// Flags for unmanned vehicules
-	var/unmanned_flags = OVERLAY_TURRET|HAS_LIGHTS|UNDERCARRIAGE
+	var/unmanned_flags = OVERLAY_TURRET|HAS_LIGHTS
```
Lumipharon, line 44: This isn't used anywhere else, so if you're going to remove it you might as well remove the define and proc.
That said, removing this does have the knock on effects of a bunch of other shit no longer going over RC's, so I don't think it's the best fix.

I would just add an early return to CanAllowThrough() to disallow mobs.

Pariah919 (commented):
Idea makes sense to me, not sure about the implementation.

BraveMole (changes requested):
In case you forgot, you need to respond to Lumi

code/modules/vehicles/unmanned/unmanned_vehicle.dm
lKiseki: I believe that UAVs aren't _that_ small to go over everything.
My logic for keeping the flag around was in case we wanted to do smaller vehicles. Feels like a waste to remove code that might be used later.

lKiseki:
> In case you forgot, you need to respond to Lumi

Donezo.

code/modules/vehicles/unmanned/unmanned_vehicle.dm
Lumipharon: fair enough, however my second point still stands, I would just make a check specifically so mobs can't just walk onto it, so runners can bump slash it but you can still throw shit over it/etc.

lKiseki: Eh, sure.

PR #11270, opened 2022-10-16
Corrosive Spit.

vilereaver41:
## About The Pull Request

Adds a low skill entry CC spit for spitter

## Why It's Good For The Game

Marine pushes are unstoppable as of late, and for xenos a lot needs to be taken into consideration. boiler globs are high cooldown and not punishing enough unless the individual is directly hit and that is 1 person out of 40 marines who will get back into battle in less then 10 seconds the radius also more often then not wont cover all routes towards the xenos unless ancient as well as requires good coordination with the hive which lets be real isn't something xenos are known for as we have a large influx of new xeno players.

Queen screech suffers from risk vs reward, again requiring coordination and high cooldown,  the only time screech is good is when you are fighting against marines that cbf playing well. in a marine static vs xeno static round marines will always win because they have more flexibility in tools and understand the overpowered weapons that haven't been fine tuned.

praetorian acid cone requires an open space, is risky as you need to go into the open with a use time, depends on if the player saves it for cc situations or random fuckery. the only thing i would say is of use for low skill CC situations is neuro cloud but lets be real for sent and prae neuro cloud is easily mended, is useless against robots. 

Right now xenos are in a tight spot where balance isn't good. marine weapons are tuned towards killing xenos very quickly regardless of circumstance which i believe is bad gameplay. things have moved to the point where marines can just rush into mazes and outclass xenos in weaponry and utility which causes a lot of ground to be lost after effort was put into building mazes.

No one is going to feel motivated to build anymore if your work doesn't even help in stopping pushes anymore, frustrating gameplay is going to show if it isnt already that one side is favored.

Xenos very rarely get new stuff, yeah we got widow but lets be real its useless (sorry zaro) xenos have limiting kits without flexibility like marines, but at the same time nothing is approved because there is a fear that a new addition will curb stomp marines and "we cant have that now" 

This leads me to corrosive spit, a low damage on impact ability that rewards in two ways.

**1) Cowd Control** - a spitter can better openly control pushes with gas clouds, as the risk in this is the reagent transfer and tick damage, no marine is dumb enough to push into an area littered with Acid clouds, or at least that's the theory?

**2) Flank + Ambush**. Marines become to comfortable in open ground without fortifying around them. if a spitter flanks and spams corrosive spit clouds eventually this is going to do one of two things if not both, (1) piss off the medics, (2) weaken your backline further leaving it open for more xenos and more harassment.

I don't believe stat buffs should be needed anymore and the best route for TGMC is to make xenomorph kits as flexible as marines to allow better combat adaption to suit the needs of the battle. 

Personel testing of the new spit type)

1) its useless as shit against barricades, 20 clouds stacked on a cade does minor damage and only a true unga would choose it over the main spit type.

2) anyone with mimir will find normal difficulty and that's if they stand in 5 clouds, the only annoyance is the reagent.

## Changelog
:cl:
add: corrosive spit.
/:cl:

Clip one with medium Marine gear with mimir 1 helm and body.

EDIT: The actual spit delay is much slower then in this clip.

https://user-images.githubusercontent.com/64131993/196018721-18579ba8-70ec-474c-954b-eb9b8391b04c.mp4


Lumipharon:
Winrates are almost perfectly 50/50 on both crash and nuke war, so saying xenos are in a bad spot isn't really true.
Mimir makes you immune to all gas clouds, including acid, so unless you've made some snowflake code around it, the gas is doing nothing, and it's only the direct hit that effects them. You tested it with a gas mask, right?

Also you should totally just replace spitter primo with this, by simply making their normal spit leave a gas cloud behind. Spitter primo is meme.

vilereaver41:
> Winrates are almost perfectly 50/50 on both crash and nuke war, so saying xenos are in a bad spot isn't really true. Mimir makes you immune to all gas clouds, including acid, so unless you've made some snowflake code around it, the gas is doing nothing, and it's only the direct hit that effects them. You tested it with a gas mask, right?
> 
> Also you should totally just replace spitter primo with this, by simply making their normal spit leave a gas cloud behind. Spitter primo is meme.

i made a toxin for it that transfers an acid reagent which makes sense and imo is pretty fair. as for winrates, during peak hours its the marines that stomp xenos due to there not being any way to stop 30+ marines charging forward. putting this behind a primo wall is pretty iffy considering this ability isnt that powerful.

Pariah919:
Honestly spitter in my experiences plays pretty okay right now and it shouldn't really need a straight ability that scales with the more spitters there are, xeno ranged should be pretty mediocre as everything should be and all I'm reading is that its mediocre.

If anything make this the primo.

Autisem:
>Mimir makes you immune to all gas clouds, including acid,

this would not invalidate it. this means they cannot carry arrowshield or valkery modules or what ever.

Boilers especialy aged are anything but weak the gas does not do ALOT of damage but its very opressive since running trough it will propably kill you. Its area denial and sieging.

I somewhat agree marines can unga ball very hard but I dont think this should be for the standard spit. I'd be cool if this was a more expensive alt fire for the spitter

vilereaver41:
closed for adjustments

Zar000:
How rude to call out widow :(

Naaanii:
This would be neat and more fitting for sentinel ngl.

PR #11276, opened 2022-10-17
Fire and damage calculation refactor

Lumipharon:
## About The Pull Request
WIP, up here for feedback.

Refactors fire to be slightly less insane. Armour now actually behaves like armour does literally everywhere else in the game, while combusting is handled by traits.
i.e. Fire armour is only used for calculating damage actually inflicted by fire, instead of 4 different calculations around with you can actually catch fire or not.

Refactors how damage is calculated in general as it is HORRIBLY inconsistant and wew.
Added a new proc modify_by_armor() that modifies a number by both hard and soft armour, and spits out the result.
Currently, this has to be done manually, and is pretty much only seen in projectile.dm. Hard armour is rarely used, but it exists and should be dealt with consistantly.

I've used this proc for all fire related stuff (probs missed some things, still WIP), projectiles, and ammo calculations, but should basically be (and will be) used every where unless there's a specific reason you would want to ignore a type of armour.

todo:
1. Rework how surt works to ensure best behavior (up in the air on what that is, but being ignitable but taking little/no damage is how it currently is, which is FUN)
2. Make sure jelly is consistant with current behavior (I think it is I need to double check I haven't missed anything)
3. ~~Replace get fire_resist~~ done
4. use modify_by_armor everywhere instead of soft armour checks
5. Check limb damage procs and replace them as appropriate with apply_damage
6. Find the other various guh damage related shitcode and try unshit it

## Why It's Good For The Game
Less shitty code is good.
~~Hopefully my shitcode is slightly less bad~~

## Changelog
:cl:
refactor: refactored armour and fire code
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
@@ -421,16 +421,15 @@
 		GLOB.round_statistics.praetorian_spray_direct_hits++
 		SSblackbox.record_feedback("tally", "round_statistics", 1, "praetorian_spray_direct_hits")
 
-	var/armor_block = get_soft_armor("acid", BODY_ZONE_CHEST)
 	var/damage = X.xeno_caste.acid_spray_damage_on_hit
-	INVOKE_ASYNC(src, .proc/apply_acid_spray_damage, damage, armor_block)
+	INVOKE_ASYNC(src, .proc/apply_acid_spray_damage, damage)
 	to_chat(src, span_xenodanger("\The [X] showers you in corrosive acid!"))
 
-/mob/living/carbon/proc/apply_acid_spray_damage(damage, armor_block)
-	apply_damage(damage, BURN, null, armor_block, updating_health = TRUE)
+/mob/living/carbon/proc/apply_acid_spray_damage(damage)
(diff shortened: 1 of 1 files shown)


DeltaFire15:
>Fire armour is only used for calculating damage actually inflicted by fire, instead of 4 different calculations around with you can actually catch fire or not.

Fire armor should still have an impact on ignition and possibly stacks as for xenomorphs especially being on fire is a significant debuff. Fire armor should mitigate or in case of king's 100+ armor prevent suffering that debuff in its entirety, not just null the DoT. The fire_immune trait here wouldn't work as that one isn't affected by sundering as supposed to.

Lumipharon:
> > Fire armour is only used for calculating damage actually inflicted by fire, instead of 4 different calculations around with you can actually catch fire or not.
> 
> Fire armor should still have an impact on ignition and possibly stacks as for xenomorphs especially being on fire is a significant debuff. Fire armor should mitigate or in case of king's 100+ armor prevent suffering that debuff in its entirety, not just null the DoT. The fire_immune trait here wouldn't work as that one isn't affected by sundering as supposed to.

losing your fire immunity and therefore pheros because you took 1 more sunder is dumb though, imo

Arctrooper8:
How will this affect Plasma Cutter Melee attacks since it uses the fire flag? Also why don't we have some sort of true damage flag type?

PR #11317, opened 2022-10-21
Makes punch and jab share a cooldown

ivanmixo:
## About The Pull Request

- Makes jab and punch share a cooldown, meaning if you use punch and it goes on cooldown, so does jab and vice versa.
- Also fixes punch giving a double alert when it's cooldown is up

## Why It's Good For The Game

Currently warrior primo just has absurd burst damage, in addition to base warrior's already excellent burst damage. I really don't think this should be possible since it makes guaranteed fractures way too reliable with no counterplay. This change should preserve it's usefulness since the cooldown is only 10 seconds, while dealing with the busted burst damage jab + punch allowed.

Also you now have to think a bit harder if you want to use jab or punch in your grapple combo, both having different varying up and downsides.

## Changelog
:cl:
balance: Primo warrior's jab and punch share a cooldown
fix: Fixes punch giving a double alert when it's cooldown is up
/:cl:


Al-1ce:
For those who don't know, every third ability used by the warrior activates the `"combo-process"`, which can inflict a variety of de-buffs depending on which ability you use. If the third move is a punch, it will do **_50% more damage._** That's what is causing the fractures.

This PR subverts warrior's primo to a nasty extent. It'd probably be better to lower the scale to 25% instead of outright nerfing the caste like this.

Al-1ce:
For those who don't know, every third ability used by the warrior activates the ``"combo-process"``, which can inflict a variety of debuffs depending on which ability you use. If the third move is a punch, it will do 50% more damage. That's what is causing the fractures here.

This PR subverts warrior's primo to a nasty extent. Lower the modifier to 25% instead of outright nerfing the caste like this.

Naaanii:
Warrior is strong but once again the nerf to it is lame and doesn't solve much. This just nerfs primo being a thing and reverts it back to pre primo Warrior, that's about it.

Nerf CD or the damage, or give lunge a windup and warning indicator like defender charge.
(Don't also nerf everything at once like MJP did)

Al-1ce:
> Nerf CD or the damage, or give lunge a windup and warning indicator like defender charge.


Agree with the first two, but lunge has absolutely nothing to do with the actual issue of fractures. That'd be a horrible nerf and it isn't even related to the fractures.

Al-1ce:
> Nerf CD or the damage, or give lunge a windup and warning indicator like defender charge.
> (Don't also nerf everything at once like MJP did)


Nerfing lunge would not change anything. I don't see how that's relevant here. It also would just generally destory the combat potential of the caste.

Al-1ce:
> Nerf CD or the damage, or give lunge a windup and warning indicator like defender charge.
> (Don't also nerf everything at once like MJP did)

Nerfing lunge will not change anything about this. Damage is the best variable here to edit, probably from 50% to 10% or 15%

usnpeepoo:
This is a bad fix for warrior primordial burst damage issue.

ivanmixo:
Then make your own instead of backseating

usnpeepoo:
> Then make your own instead of backseating

Ever thought that people leave feedback on open PRs?

S1tr3p:
> Then make your own instead of backseating

shut up retard

ivanmixo:
"This is bad" is not feedback

Goblinu:
> "This is bad" is not feedback

This is terrible

Al-1ce:
> "This is bad" is not feedback 

We have thumbs up/down on community PRs for a reason. You need to gauge the reaction of the playerbase before shamelessly nerfing something that could otherwise be done in a better way.

People don't like this change, - not in it's current state. That itself is feedback. It shows you that there is something here which isn't being well recieved. I went out of my way to (briefly) explain why this isn't good, but.. whatever, I guess?

DerpiestBruh:
This PR is the best thing for tgmc we should merge it immediately

Starfirejordan:
> > "This is bad" is not feedback
> 
> This is terrible

Horrible even

Zar000:
if you introduce them sharing the same cooldown, make each effect more pronounced.
Examples:
- Jab slow increase
- Punch stun increase
- Replace primo punch with uppercut
- Replace jab with double-leg kick

Idk, but I think changing it and making the choice matter more is better

Lumipharon:
While I agree that warrior has some pretty gross burst damage, I don't think this is the right solution, warrior relies on comboing abilities together (well, all xenos do for that matter), and as stated above this directly goes against warrior's primo gimmick.

ivanmixo:
There is no smart comboing, you just use one ability then mash the other out in 0.1 seconds.

PR #11324, opened 2022-10-22
Arrowhead no longer has slowdown.

Pariah919:
## About The Pull Request
Removes the slowdown off Arrowhead, the energy shield module.

## Why It's Good For The Game
Its currently pretty bad, it drops you down 2 armor levels and has slowdown making it difficult to use at all other than tanking spits and mimir is better at that.

## Changelog
:cl:
balance: Arrowhead no longer has slowdown.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -256,7 +256,6 @@
 	icon_state = "mod_eshield"
 	item_state = "mod_eshield_a"
 	slot = ATTACHMENT_SLOT_MODULE
-	slowdown = 0.2
 	soft_armor = list(MELEE = -10, BULLET = -5, LASER = 0, ENERGY = 0, BOMB = 0, BIO = -5, FIRE = 0, ACID = -5)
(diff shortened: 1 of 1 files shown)


Lumipharon:
get rid of the funny pointless negative bullet armour.

Losenis:
The funny pointless negative bullet armor was put in place due to it having free health regen that everyone and their mother would pick back in the day for a free module to invalidate any caste that couldn't just land more than 2 or 3 hits on your without having to run.

Which is every caste if you're surrounded by ungas on aim mode.

Lumipharon:
Xenos don't shoot bullets

Al-1ce:
> Xenos don't shoot bullets

But SOM does.

Lumipharon:
> > Xenos don't shoot bullets
> 
> But SOM does.

Yes, and I added it to combat patrol prenerf, so it was a guh change and wasn't intended from my perspective.

Losenis:
> Xenos don't shoot bullets

lmao I missed the bullet part, my bad

~~inb4 preemptive gun xeno buff~~

FosterGrime-Heymynameslouie:
I think this would make shield better at its intended role (back line support)

with its negative armor on literally everything encouraging you to do backline activities instead of sticking around too long at front for an extended period of time making you not as slow as a snail kinda makes sense for scout/backline support roles it's now intended to have.

spedeb merhred

Al-1ce:
Agree ^

PR #11333, opened 2022-10-23
Make crash great again

Zar000:
## Why It's Good For The Game
Crash is pain and this will hopefully make it less pain

## Changelog
:cl:
balance: 2->4 xenos at start on crash | T2 start as elder | T3 start as mature
/:cl:


srxtreme:
speedmerghe

FosterGrime-Heymynameslouie:
I beg 

spedmed rhg

Dgeen:
As much as I would like to fix the crash. This is the wrong way to fix it. 
Look, let's start with maturity, now when you evolve into t1, you see that you have been thrown age, it seems logical, you need to disperse with regular deaths, but when you evolve into t2, you suddenly notice that 2 stages have been taken away from you, not 1.
After your pr, the questions will only increase, why when you evolve into t2 my age wasn't taken away from me? And now they've taken it away. And during the evolution in t3, two stages were selected at once. Although it looks consistent by code, in the game it will look very inconsistent and misleading
In a good way, the evolution in t2 and t3 does not need to be touched at all, but to leave the normal path.
Now for the starting larvae. Changing the initial quantity will not fix the broken mechanism for issuing new ones. Look, now the larva-Marines ratio is in the range between 2-1 to 1-10. Having given two larvae, first we will come to a ratio of 4-1 to 1-10, do you notice the problem? In addition to the fact that 4 larva breaks the balance with <8 players, after killing the first 4, there is no guarantee that the game will not take away 1 larva, or 2, or maybe all 3.

BraveMole:
"Crash is pain and this will hopefully make it less pain"

I am closing just because you are not giving any arguments in the why it is good for the game

PR #11335, opened 2022-10-23
[Discussion Needed] Lower cost of x fuel in req

maikilangiolo:
## About The Pull Request
The idea was thrown around a bunch of times without anyone committing to it, so I did it. Back tank cost 60 -> 30, 72-round large tank 30 (lmao) -> 10, and normal large tank 6->5, both to be in line with most other mags in req and to have a large point spread between that and the x fuel tank (half as much).

## Why It's Good For The Game
X fuel cannot be refueled anymore. while the 60 point cost was fine when you could, right now its an excessive cost. Not just that, but the 30-point tanks @Naaanii added were also single use, and I am not sure why you would spend 60 points for 140 x fuel vs 60 points for the back tank and 500 fuel. Now priced at 10, if you spend 60 points you get almos 100 less fuel than a back tank's worth.

_**This PR was marked with "discussion needed" because I wanted to indicate the price nerf is negotiable and I set them to something I personally viewed as fair**_. Maybe the back tank could be priced at 40, so both the single magazines and the back tank are double the cost of the normal equivalent (5 and 20 points vs 10 and 40, in this case). Though 40 is a lot for being disposable.

## Changelog
:cl:
add: Added more mechanics or gameplay changes
expansion: Expands content of an existing feature
del: Removed old things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -480,7 +480,7 @@ WEAPONS
 /datum/supply_packs/weapons/napalm_X
 	name = "FL-84 X fuel tank"
 	contains = list(/obj/item/ammo_magazine/flamer_tank/large/X)
-	cost = 8
+	cost = 10
 
 /datum/supply_packs/weapons/back_fuel_tank
 	name = "Standard back fuel tank"
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -475,12 +475,12 @@ WEAPONS
 /datum/supply_packs/weapons/napalm
 	name = "FL-84 normal fuel tank"
 	contains = list(/obj/item/ammo_magazine/flamer_tank/large)
-	cost = 6
+	cost = 5
 
 /datum/supply_packs/weapons/napalm_X
 	name = "FL-84 X fuel tank"
 	contains = list(/obj/item/ammo_magazine/flamer_tank/large/X)
-	cost = 30
+	cost = 8
 
 /datum/supply_packs/weapons/back_fuel_tank
 	name = "Standard back fuel tank"
@@ -490,7 +490,7 @@ WEAPONS
 /datum/supply_packs/weapons/back_fuel_tank_x
 	name = "Type X back fuel tank"
 	contains = list(/obj/item/ammo_magazine/flamer_tank/backtank/X)
-	cost = 60
+	cost = 30
 
 /datum/supply_packs/weapons/rpgoneuse
 	name = "RL-72 Disposable RPG"


Araulius:
pyro mains rejoice.

Losenis:
Once you get xfuel xenos basically have no mazes anymore, maybe dont spamclick on spray nozzle if you run out of it that fast.

maikilangiolo:
the same thing happens with just normal flamer., which is free and infinite, and thus ten times more spammable thatn x fuel. especially because it still deletes walls in one burn, but it burns out faster so marines can push the maze sooner without wasting water. and jelly makes you 100% to x fuel too, and it is easily spammed as well.

Al-1ce:
The way I see it, this is just powercreep. Things are fine as they are currently.

Naaanii:
Point reduction is okay because regular fire has the same effect. My use for X fuel is only for damage but jelly does negate it, putting it in a very weird and shitty spot.

I intend to make regular tanks refillable but a separate welderpack would be required, but I do agree as of now it does not feel worth the cost at all considering jelly exists and Surt itself is in a terrible spot.

> Once you get xfuel xenos basically have no mazes anymore, maybe dont spamclick on spray nozzle if you run out of it that fast.

Xfuel does not speed up the deletion of mazes at all and is only deadlier to marines even more rather than Xenos as it takes the same amount of fire to kill (with the exception of spray nozzle) and is sub optimal. This relies more on the issue of spray nozzle itself, but it isn't an issue and most Xenos cope because of the lack of jelly or hivelord which while it's necessary to have, a flamer is also necessary as mazes are CBT for marines and PC is in an even worse spot now more than ever leaving flamer as the only viable option.

Flamers could be balanced in a way where some nozzles excel in maze clearing while others can be used more for damage and let PC do its thing with maze clearing but with the limited amount, rouny snatch and Xeno drag being a thing, flamers are the better option no matter what.

Rather than nerfing maze damage, Xeno jelly could use changes or a buff regarding availability so Xenos can actually combat through fire and slow down fire pushes instead of just relying on off screen castes as it still takes time for the walls and cover to actually die.

Starfirejordan:
No add a X-Fuel tank to req in instead so you can refill it, make it light blue or something, boom easy fix

Lumipharon:
> No add a X-Fuel tank to req in instead so you can refill it, make it light blue or something, boom easy fix

I actually prefer this. It keeps a higher initial cost, but then a lower maintenance cost via cheaper refills.

maikilangiolo:
Thought the same. Just needs spriting and coding. Does your code block all refueling from tanks, or only those with normal/welder fuel inside?

Lumipharon:
> Thought the same. Just needs spriting and coding. Does your code block all refueling from tanks, or only those with normal/welder fuel inside?

I added a check for the fuel type. So when you make a xfuel tank, you'll just need to make a child version of the proc with basically inverted criteria.

PR #11336, opened 2022-10-23
Increase Synth Nade Permissions

SiegeB0t:
## About The Pull Request
Lets synth use mirage grenades.

## Why It's Good For The Game
Mirage has little combat value and should be considered safe.

## Changelog
:cl:
balance: Synth can use mirage grenades.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/mirage.dm
+++ b/code/game/objects/items/devices/mirage.dm
@@ -3,6 +3,7 @@
 	desc = "A special device that, when activated, produces a pair of holographic copies of the user."
 	icon_state = "delivery"
 	item_state = "delivery"
+	dangerous = FALSE
 	///the parent to be copied
 	var/mob/living/current_user
 	///How long the illusory fakes last
--- a/code/game/objects/items/explosives/grenades/chem_grenade.dm
+++ b/code/game/objects/items/explosives/grenades/chem_grenade.dm
@@ -243,6 +243,7 @@
 	desc = "Contains construction nanites ready to turn a small area into razorwire after a few seconds. DO NOT ENTER AREA WHILE ACTIVE."
 	icon_state = "grenade_razorburn"
 	item_state = "grenade_razorburn"
+	dangerous = FALSE
 	stage = CG_READY
 	icon_state_mini = "grenade_chem_yellow"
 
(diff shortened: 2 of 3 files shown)


BraveMole (changes requested):
Your why it is good for the game is a lore argument. That is not something we want to consider

Losenis:
>Lore

Wasn't limiting synth to non-combat stuff an actual gameplay concern and not lore? Considering synth can already make razorwire, razorwire nades not working seems like an oversight. Mirage has zero combat potential other than confusing xenos. Tangle might be overboard though since you could actually end up with a synth as a full dedicated tangle thrower with a backpack full of them.

Goblinu:
> Tangle might be overboard though since you could actually end up with a synth as a full dedicated tangle thrower with a backpack full of them.

Tangle is incredibly powerful against xenos, the last thing i would want is a option for marines to deploy a powerful deterent or push stopper with little to no risk to themselves

SiegeB0t:
Synth does not have a reliable deterrent when caught out of position. If the synth is part of a push there are significant downsides to a synth supporting with only tangle as opposed to a demoman with a grenade box. Its weird that synths are being singled out given that they have to sacrifice engineering/medical storage space for grenades that any other unga could spam.

Tangle is a scarce resource given that marines invariably run out of the roundstart supply during a midpop game. A synth hogging this resource would starve the rest of the force. Additionally, a synth with a box of chem nades is way less reliable than CAS which can spam larger smoke bombs 4.32 times per minute. Even then its a rarely used strategy.

Losenis:
As much as I agree that singling out synth when marines already have combat robots is dull, the argument isn't about who is using it, it's about the ability of a synth to ALSO do that with the push on top of everything else it should be doing. Takes a couple points to have RO order one, drop on your beacon, and start throwing nades on your downtime. If said box is brought down everyone can use it too, and given the synth can't use guns, throwing them is all it'll focus on without distractions.

Then again the argument is invalidated if there's a flamer around, just throw it into the fire to activate it. Everyone packs fire nowadays to clear maze and remove huggers, so I don't know if it'll be as bad as one thinks.

Starfirejordan:
Wait.. they couldn't use theses nade types before?

Lumipharon:
I'd say yes to mirage but no to the other two.

TiviPlus (changes requested):
Agree with lumi

Lumipharon (approved):

PR #11342, opened 2022-10-25
Adds dispenser pick to Synthetics

Losenis:
## About The Pull Request
Title

## Why It's Good For The Game
Brings in line with what engineer can pick for backpacks since it already could choose technician backpack, welderpack, and radiopack.

## Changelog
:cl:
add: Synthetics now have Dispenser as backpack choice in their vendor
/:cl:


Aquilar:
I suggest moving the dispenser from engineer clothes vendor to engineer gear vendor before doing this, as this is storage creep.

TiviPlus:
I dont see it at the same level as welderpacks and would prefer it stays engi only

Losenis:
Fair enough.

PR #11344, opened 2022-10-25
HMG08 Small Mag

Ben10083:
## About The Pull Request

Adds a Smaller 250-round box magazine for the HMG-08.

## Why It's Good For The Game

HMG-08 users have alot of trouble when it comes to storing their ammo, as unlike other MGs, the magazines for the HMG-08 cannot be stored in something like a belt. This smaller variant allows for the user to carry more magazines, albeit a smaller clip size. 

## Changelog
:cl:
add: Smaller Magazine for HMG-08 added
/:cl:


Pariah919 (approved):
Seems good to me overall, other than maybe you should double the amount of small magazines so the round count is still effectively the same.

FlakBait00:
However there is a issue with balance, MG08 magazines are meant to be carried either in the back or the belt of the uh gunner, so uh someone carrying like four times the ammo in a IMP backpack would be unfunny i think.

revelation-8468:
> However there is a issue with balance, MG08 magazines are meant to be carried either in the back or the belt of the uh gunner, so uh someone carrying like four times the ammo in a IMP backpack would be unfunny i think.

Not necessarily - Do you know how LONG it takes for the MG08 to be reloaded?

PR #11358, opened 2022-10-26
Cleans the server

Lumipharon:
## About The Pull Request
Soap REFACTORED
Showers REFACTORED
Rivers FRESH
Server CLEANER THAN EVER

Adds a clean_mob() and clean_turf() proc to mobs and turfs respectively. Maybe I should have just added this at the atom level but eh.

Made shower code a little bit less guh, and means RIP's rain pr can just reference this proc instead of copy pasting shower code, like I've done with river code.
Soap can now actually target a turf to clean all the crap off it instead of targeting one decal at a time.

Fixes a pretty funny bug where showers can be used to qdel ANY effect/overlay, for example OB/CAS laser indicators, certain effects like timestop and gib animations. Also palmtrees.
Also stops showers from runtiming constantly as they clean decals in the most bizarre way possible.

## Why It's Good For The Game
Cleaner cleaning code
bugfix good.

## Changelog
:cl:
balance: Buffed soap. You can now directly clean turfs
fix: fixed some shower exploits/bugs
refactor: refactored soap, showers and rivers slightly
/:cl:

### DIFF
--- a/code/game/blood.dm
+++ b/code/game/blood.dm
@@ -135,11 +135,11 @@
 
 /mob/living/carbon/human/clean_mob()
 	. = ..()
-	var/washgloves = 1
-	var/washshoes = 1
-	var/washmask = 1
-	var/washears = 1
-	var/washglasses = 1
+	var/washgloves = TRUE
+	var/washshoes = TRUE
+	var/washmask = TRUE
+	var/washears = TRUE
+	var/washglasses = TRUE
 	if(wear_suit)
 		washgloves = !(wear_suit.flags_inv_hide & HIDEGLOVES)
 		washshoes = !(wear_suit.flags_inv_hide & HIDESHOES)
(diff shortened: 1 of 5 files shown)


code/game/blood.dm
```diff
@@ -115,6 +115,65 @@
...
+			if(wear_mask.clean_blood())
+				update_inv_wear_mask()
+
+/mob/living/carbon/human/clean_mob()
+	. = ..()
+	var/washgloves = 1
+	var/washshoes = 1
+	var/washmask = 1
+	var/washears = 1
+	var/washglasses = 1
```
TiviPlus, line 142: true/false`?

TiviPlus (changes requested):
rest lgtm

code/game/blood.dm
Lumipharon: whoops

TiviPlus (approved):

PR #11363, opened 2022-10-27
Pepperball magazine, aim mode buff, infinite magazines, added to req and rail light addition

Pariah919:
## About The Pull Request
Fixes Pepperball gun not being buyable in req, first and foremost.

Makes magazines infinite in the marine vendor instead of there only being 40 of them, no real reason to force you to shred threw these things, and if I did add a req only variant I'd readd the old Tanglefoot SAN balls.

Aim mode has no slowdown now and I reduced aim mode fire delay to .2 from .25 total, and basically serves as a timed IFF to stop people from having .1 fire rate aim mode and instead have rifle firerate.

Upped the guns magazine size to 100, and added rail light as an attachable if you want it over mag harness.

## Why It's Good For The Game
The Forgotten weapon, the Pepperball gun.

It was outdated on release as I never gave it a followup and this should atleast give it an identity now as a good support tool, if its still poopy I'll give it underbarreled attachments I guess.

## Changelog
:cl:
fix: You can actually buy Pepperball guns now.
balance: Pepperball gun magazines are now infinite, aim mode is buffed to have no slowdown when aiming and it now has rail light. It now has 100 rounds in a magazine over 70.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -116,7 +116,7 @@
 		),
 		"Specialized" = list(
 			/obj/item/weapon/gun/rifle/pepperball = 4,
-			/obj/item/ammo_magazine/rifle/pepperball = 40,
+			/obj/item/ammo_magazine/rifle/pepperball = -1,
 			/obj/item/weapon/gun/flamer/big_flamer/marinestandard = 4,
 			/obj/item/ammo_magazine/flamer_tank/large = 16,
 			/obj/item/ammo_magazine/flamer_tank/backtank = 4,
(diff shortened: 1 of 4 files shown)


Naaanii:
Slowdown removal with aim mode, good. RoF reduction is powercreep.

Don't think it needs underbarrel attachments at all. Rail flash is fine.

FosterGrime-Heymynameslouie:
paintball actually making sense? baller
never really understood why it's ammo was limited and why it was never in req

a req version with old San tfoot gas balls might be cool as a true crowd control tool but would probably be mauled by xeno players everywhere 

spdee meghfed

PR #11364, opened 2022-10-27
Plasma cutter now has the ability to nodrop toggle.

Pariah919:
## About The Pull Request
Works like shield and powerfist
image

## Why It's Good For The Game
Makes plasma cutter more reliable in the age of xeno drag and makes sense anyway considering its a high value item.

## Changelog
:cl:
balance: Plasma cutter now has togggleable nodrop.
/:cl:

### DIFF
--- a/code/game/objects/items/tools/mining_tools.dm
+++ b/code/game/objects/items/tools/mining_tools.dm
@@ -309,3 +309,18 @@
 	cut_apart(user, O.name, O)
 	qdel(O)
 	return TRUE
+
+/obj/item/tool/pickaxe/plasmacutter/AltClick(mob/user)
+	if(!can_interact(user))
+		return ..()
+	if(!ishuman(user))
+		return ..()
+	if(!(user.l_hand == src || user.r_hand == src))
+		return ..()
+	TOGGLE_BITFIELD(flags_item, NODROP)
+	if(CHECK_BITFIELD(flags_item, NODROP))
+		to_chat(user, span_warning("You feel the [src] clamp shut around your hand!"))
+		playsound(user, 'sound/weapons/fistclamp.ogg', 25, 1, 7)
+	else
+		to_chat(user, span_notice("You feel the [src] loosen around your hand!"))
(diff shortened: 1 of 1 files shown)


TiviPlus:
just give it higher drag delay

Aquilar:
wouldn't this be an indirect buff to synths?

Losenis:
SPEEDMERGE HOLY SHIT LETS FUCKING GOOOOOOOOOOOOOOOOOOOOOOOO

image

MereWalker:
If your lose your fun stick that is on you imo.

Losenis:
Posted while wraith banishes and runs into the maze with it

Losenis:
Xenos have enough ways to force knockdown and displacement that can only be countered with 「Future Vision」 given while you're stunned or tiles away another xeno can drag the PC. In practice a PC will be used to aid a push, which is going to be defended by multiple xenos rather than one.

Examples are wraith banishing you, warrior tossing you elsewhere and god forbid it's a maze, shrike doing the same, crusher displacement + tanking, defiler gas right after a knockdown, so on and so forth in any combination and any number given that there's no reason why there can't be even 10+ xenos in one given place. For the sake of the argument we'll also ignore the hundred million ways the marines themselves will help you lose your PC, like getting slugged in the back in the head since you need to be in melee range to use the PC anyway.  The ability to pull items of which plasma cutter is not going to be blacklisted on the list of ended up a probably unintended nerf on top of all the other nerfs like map removal, synth removal, inability to change cells, damage reduction, and so on. I'd like to bring emphasis to the most recent damage reduction that already preemptively dealt with memelords going shield + PC going overboard.

FIXTHERADBOXPLEASE:
For context I always see Losenis go shield + PC engineer and kill/cryo himself if his cutter ever gets melted. Not slander just a relevant observation.

Losenis:
>shield + PC engineer

The only times I go shield is when I go t90 tyr2, and in those times I go PFC+ engi so I spend 60 points on 3 PCs so I don't kill/cryo myself if I lose one. Are you confusing me for someone else?

PR #11365, opened 2022-10-27
MP19 is now bulky.

Pariah919:
## About The Pull Request
~~TP-19~~ MP-19 is now bulky size rather than normal size once more.

## Why It's Good For The Game
I don't think this thing should've ever been normal size in the first place, honestly I swore it was bulky for the longest time but apparently not.

## Changelog
:cl:
balance: MP-19 is now bulky sized.
/:cl:


TiviPlus:
makes it a shitter smg90, no

PR #11366, opened 2022-10-27
Explosion light fix

Lumipharon:
## About The Pull Request
The explosion visual and lighting effect are now both correctly centered on where the explosion occured.

The cool new explosion animation and lighting effect had an offset problem, where the explosion was centered but the light was not.
This is because the light source is based off where the icon is, which is offset due to being 64x64.

Making the explosion an overlay fixed this, as that can be offset separately. I also gave it a rotate, so while barely noticable, it means when you have 10 explosion go off on screen at once, they won't all be perfectly identical, which I think is visually nice.

## Why It's Good For The Game
bug fix good.

## Changelog
:cl:
fix: fixed explosion lights not being correctly centered
/:cl:

### DIFF
--- a/code/game/objects/effects/temporary_visuals/miscellaneous.dm
+++ b/code/game/objects/effects/temporary_visuals/miscellaneous.dm
@@ -17,9 +17,8 @@
 	set_light(radius, radius, color)
 
 	var/image/I = image('icons/effects/explosion.dmi',src,"explosion",10 , pixel_x = -16, pixel_y = -16)
-	var/angle = rand(0, 359)
 	var/matrix/rotate = matrix()
-	rotate.Turn(angle)
+	rotate.Turn(rand(0, 359))
 	I.transform = rotate
 	add_overlay(I) //we use an overlay so the explosion and light source are both in the correct location
 
--- a/code/game/objects/effects/temporary_visuals/miscellaneous.dm
+++ b/code/game/objects/effects/temporary_visuals/miscellaneous.dm
@@ -10,13 +10,13 @@
 	duration = 8
 	light_system = STATIC_LIGHT
 	///The icon for the explosion itself
-	var/explosion_icon
+	var/explosion_icon = "explosion"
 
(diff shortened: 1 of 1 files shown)


code/game/objects/effects/temporary_visuals/miscellaneous.dm
```diff
@@ -6,16 +6,22 @@
...
 
 /obj/effect/temp_visual/explosion/Initialize(mapload, radius, color)
 	. = ..()
 	set_light(radius, radius, color)
 
+	var/image/I = image(icon, src, explosion_icon, 10, pixel_x = -16, pixel_y = -16)
+	var/matrix/rotate = matrix()
+	rotate.Turn(rand(0, 359))
+	I.transform = rotate
+	add_overlay(I) //we use an overlay so the explosion and light source are both in the correct location
```
TiviPlus, line 23: can we just overlays += I here to microopt

code/game/objects/effects/temporary_visuals/miscellaneous.dm
```diff
@@ -6,16 +6,22 @@
...
 	name = "explosion"
 	icon = 'icons/effects/explosion.dmi'
-	icon_state = "explosion"
+	icon_state = ""
 	duration = 8
-	pixel_x = -16
-	pixel_y = -16
 	light_system = STATIC_LIGHT
+	///The icon for the explosion itself
+	var/explosion_icon = "explosion"
```
TiviPlus, line 13: couldnt you just set icon_state then null this in initialize

TiviPlus (changes requested):

code/game/objects/effects/temporary_visuals/miscellaneous.dm
Lumipharon: done

code/game/objects/effects/temporary_visuals/miscellaneous.dm
Lumipharon: done

TiviPlus (approved):

PR #11370, opened 2022-10-28
Decreases jab blur

Losenis:
## About The Pull Request

Warrior primo jab will no longer blur you for so long it can stack with itself or others

## Why It's Good For The Game
Turns out that it lasts for long enough that it can actually stack with itself/other warrior punches and effectively blind you harder than having actual proper eye damage which is pretty absurd. Lowers it to a third of its duration so that while it still impacts the fight the warrior is in it wont last for centuries after the warrior runs. It was purely cosmetic anyway since you can tell that the big purple thing is a xeno to shoot at.

## Changelog
:cl:
balance: Primordial warrior's blind duration lowered.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
@@ -584,8 +584,8 @@
 	if(!target.punch_act(X, damage, target_zone, push = FALSE, punch_description = "precise", stagger_stacks = 3, slowdown_stacks = 6))
 		return fail_activate()
 	if(X.empower())
-		target.blind_eyes(3)
-		target.blur_eyes(6)
+		target.blind_eyes(1)
+		target.blur_eyes(2)
 		to_chat(target, span_highdanger("The concussion from the [X]'s blow blinds us!"))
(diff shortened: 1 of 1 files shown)


FIXTHERADBOXPLEASE:
Salt PR?

Lumipharon:
How many warriors are chain jabbing you to actually make blur matter? And how are you not dead before it matters except for like, heavy tyr + shield?

Al-1ce:
Your reasoning seems heavily situational and reliant on exceptions.

DerpiestBruh:
sus

Dgeen:
You wrote that you reduce the blur. But in fact, you also reduce the blinding. Misleading is not very good, do you know that? Although I don't know why to describe something in a typical salt pr

Losenis:
I lowered the visual effects overall by a third. Instead of being condescending and accusing of malice you could've just said "hey maybe keep the blinding aspect" and I'd have done so without problem.

ivanmixo:
People have a shitfit whenever you try to touch anything about warrior, so.

TiviPlus (changes requested):
If your problem is the stacking then why not use a max?

Naaanii:
You can also reduce the duration within Imi which is funny.

PR #11378, opened 2022-10-29
Fixes ravage being able to move dead bodies

ivanmixo:
## About The Pull Request

Bit copy-pasty but it'll do the trick.

## Why It's Good For The Game

This is unintended.

## Changelog
:cl:
fix: Fixes ravage being able to move dead bodies
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
@@ -123,7 +123,7 @@
 		var/mob/living/carbon/human/attacking = ravaged
 		if(attacking.stat == DEAD)
 			continue
-		step_away(ravaged, X, 1, 2)
+		step_away(attacking, X, 1, 2)
 		attacking.attack_alien_harm(X, X.xeno_caste.melee_damage * X.xeno_melee_damage_modifier * 0.25, FALSE, TRUE, FALSE, TRUE)
 		shake_camera(attacking, 2, 1)
 		attacking.Paralyze(1 SECONDS)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
```diff
@@ -116,14 +116,14 @@
 	for(var/atom/movable/ravaged AS in atoms_to_ravage)
 		if(!(ravaged.resistance_flags & XENO_DAMAGEABLE))
 			continue
-		if(!ravaged.anchored)
-			step_away(ravaged, X, 1, 2)
-		if(!ishuman(ravaged))
+		if(!ishuman(ravaged) && !ravaged.anchored)
```
Lumipharon, line 119: You've removed the is human check before typecasting ravaged to human below. This means for example, you'll runtime when you ravage a wall because it's not human but is anchored, so will be typecast as a human.

Lumipharon (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
ivanmixo: Walls are not `/atom/movable`, so this shouldn't come up. It also slashes non human atoms just fine.

Lumipharon: Whoops bad example, but it should still apply to anchored atoms, unless I'm illiterate which is possible.

Lumipharon (approved):

PR #11379, opened 2022-10-30
Fixes smoothing when mining turfs

RipGrayson:
## About The Pull Request

Title.
The new smoothing system doesn't really understand the concept of ChangeTurf() that well and so you get ugly black voids or weird missing walls if you plasma cut a wall.
This PR fixes that by integrating a smoothing check into ChangeTurf itself so the walls behave properly.

## Why It's Good For The Game

I have really needed to do this for a while, it's a small consistency fix that makes mining more bearable.

## Changelog
:cl:
fix: Walls will now properly smooth when destroyed via plasma cutter.
/:cl:

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -316,7 +316,7 @@
 	if(thisarea.lighting_effect)
 		W.add_overlay(thisarea.lighting_effect)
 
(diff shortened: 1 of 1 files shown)


Lumipharon (approved):
I thought there might be another way to do this but then I forgot, so seems good.

PR #11380, opened 2022-10-30
Neuraline Buff

Araulius:
## The Current State of Neuraline

Neuraline at the moment sees next to no use. It is extremely costly to make, and only is craftable in batches of 1 (one) dose. 4u per batch. It requires 2 ingredients that already consume lemoline to make, on top of the lemoline catalyst required for the final chemical reaction to form neuraline. On top of that, you require prepared larval jelly. One dead larvae can create one, (1), unit of prepared larval jelly. At the moment, one dose of neuraline, (4u), lasts 10 ticks, and deals a total of 38 toxin damage.

## The Changes This Would Implement

1. Neuraline would go from 10 ticks to metabolize per dose (4u), to 20 ticks per dose (4u). This would increase its effectiveness as a combat stimulant.
2. Toxin damage dealt over time reduced from 3.75 per tick, to 2.75 per tick. This would result in a net toxin damage total of 55, which is roughly 20 more damage total than before.

## Changelog
:cl:
balance: Increased how long it takes for Neuraline to metabolize. 
balance: decreased neuraline toxin damage per tick from 3.75 to 2.75
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -457,7 +457,7 @@
 	name = "Neuraline"
 	description = "A chemical cocktail tailored to enhance or dampen specific neural processes."
 	color = "#C8A5DC" // rgb: 200, 165, 220
-	custom_metabolism = REAGENTS_METABOLISM * 2
+	custom_metabolism = REAGENTS_METABOLISM
 	overdose_threshold = 5
 	overdose_crit_threshold = 6
 	scannable = FALSE
@@ -494,7 +494,7 @@
 	L.AdjustSleeping(-40)
 	L.adjustStaminaLoss(-30*effect_str)
 	L.heal_limb_damage(7.5*effect_str, 7.5*effect_str)
-	L.adjustToxLoss(3.75*effect_str)
+	L.adjustToxLoss(2.75*effect_str)
 	if(iscarbon(L))
 		var/mob/living/carbon/C = L
 		C.setShock_Stage(min(C.shock_stage - volume*effect_str, 150)) //will pull a target out of deep paincrit instantly, if he's in it


Lumipharon:
I agree it's unreasonably difficult to make due to the lemoline changes, but the chem itself isn't weak and imo doesn't need a buff.

Araulius:
> I agree it's unreasonably difficult to make due to the lemoline changes, but the chem itself isn't weak and imo doesn't need a buff.

The chem isn't weak, that is correct. However, given its extreme scarcity, I feel it should do more than it does. In general, I would rather do this, than make neuraline easier to acquire. If you make neuraline more easy to acquire, suddenly instead of one marine becoming a wall for 20 second, 3 times per round; You get multiple marines becoming a juggernaut for 10 seconds, several times per round. I feel like having far fewer instances of a stronger buff, would be preferable to having a larger amount instances of the current buff it provides.

Al-1ce:
> I agree it's unreasonably difficult to make due to the lemoline changes, but the chem itself isn't weak and imo doesn't need a buff.

Neuraline is shit. I've been using it consistiently as corpsman and the price just isn't worth it. 14 points for a single injector is bull.

 The healing factor is _decent_ but it doesn't really compensate for the absurdly cost.

Al-1ce:
> I agree it's unreasonably difficult to make due to the lemoline changes, but the chem itself isn't weak and imo doesn't need a buff.

Neuraline is shit for what it costs. 14 points for a 3 use injector is probably the least consistient item in the entire game. The amount of healing it provides isn't enough for it to be practical with the cost in mind.

You rarely see Neuraline used in the field for a reason. I feel like buffing it would give the expense some ground to stand on.

Naaanii:
Reducing the cost and giving a pill or bottle option sounds like a better idea.

Pariah919 (commented):
I don't like this since Neuraline is supposed to be a super chem with downsides, the issue with it is that its super incredibly stingy for no real reason and Inaprov eclipsed it ages ago.

Make it more available first as the actual T2 equivalent to both ina and synap by making it more avaliable (not in pill bottle form standard) in bottled form.

Al-1ce:
> Make it more available first as the actual T2 equivalent to both ina and synap by making it more avaliable (not in pill bottle form standard) 
>in bottled form.

This sounds stupid as hell but I'd prefer just about anything to the current situation.

Al-1ce:
> Make it more available first as the actual T2 equivalent to both ina and synap by making it more avaliable (not in pill bottle form standard) in bottled form.
>Bottle

Maybe? I dunno, I don't see why all three of these things shouldn't be tried. Can't seriously see medics using bottle/syringe in the modern climate. Why not a hypospray?

TiviPlus:
What mjp said

PR #11384, opened 2022-10-30
Rebalances engineer gear vendor

Aquilar:
## About The Pull Request

This PR rebalances engineer gear by putting a heavy restriction on the number of power items an engineer can take to the field on their own.
had to revert the ability to have separate point totals per category to make this work as the alternative to keeping it got rather complicated. 

In essence this limits each engineer to a single power item while still costing points for that item.
you purchase one power item, the other power items become greyed out.
In the future new strong engineer items should be added to this category, this category is also available to the other vendors.  
unknown

## Why It's Good For The Game

I think this a good compromise from straight up taking toys away from engineers while reigning in their ability as a force multiplier by bringing multiple strong items to the field. 

## Changelog
:cl:
add: New equipment category added to marine vendors called: Specialist Equipment.
balance: Dispenser gets move to new specialist equipment category and costs 15 points.
balance: Liguana, ST-580 mini sentry and teleporters moved to new specialist equipment category.
code: reverted some coding changes to marine vendors for bringing back category restrictions on points costing items.
/:cl:


FosterGrime-Heymynameslouie:
my yee haw loadout, gone, obliterated, reduce to ashes.

Losenis:
This feels like a nerf out of everyone suddenly going engi due to new engi content. Has this even been a problem in the past even when dispenser was last TMd?

>Dispenser costs 15 points and is now mutually exclusive with teleporter

Dispensers already get shot to death due to catching bullets even on aim mode, and marines barely ever defend teleporters that will get blown up by nades. On top of another pr nerfing dispenser health. And that's on top of needing 2 recharger cells for teleporters to not meme you with a battery used up needing you to walk half the map to the other teleporter to swap cells, assuming you packed a ton of cells and didn't spend 3 additional points for the cell recharger that you'll eventually need if the round drags on for too long. This is way too heavyhanded, and I'd also question dispenser being moved there allowing engis to grab a backpack slot on top of dispenser, since radiopack would be a very strong choice given it's essentially a free supply beacon + request console.

When novelty wears off and less people go engi this'll affect the remaining engis pretty hard.

Aquilar:
@Losenis 
it has 2.5 times the storage space of a normal backpack, can fit anything including weapons, mortar, HSG, HMG, shells, welding kits and fuel tank backpacks but not storage items, on top of that is heals combat robots up to 2/3 tiles away at a steady rate of 2 brute, 2 burn points every tick with no limit on how many are affected. Issues with it's durability can be dealt with in other PRs.

because of the changes in the PR I did to the vendors, you make stuff in the cloth vendor cost points,  it'll affect the other vendors you can interact but I wanted to keep all points based gear in the gear vendor and not the clothing vendor.

Losenis:
>it has 2.5 times the storage space of a normal backpack

Dragging a bike with a crate is an option. Using tele to keep your storage needs in fob and deploying it as necessary is also an option. Dispenser is great, no contest, but it's not the only option out there. Also account for the inevitable nobrainer of having the radiopack for storage on top of that since it'll be a free pick should this pass.

Healing robots means it has to be deployed which means you **WILL** have your shit stolen and dispenser shot up, assuming someone doesn't outright steal it because the only thing stopping that is the player's own ethical standards and cirumstances like it getting attacked and the engi not being close enough so another undeployed it. And if robots can stand around doing nothing to get healed, they can heal themselves with their own wires and blowtorch just fine as well, or they're in the middle of a fight so close to the dispenser that it should 100% be undeployed and run away with due to the risk of getting caught in the crossfire and destroyed.. Honestly I gave so little a damn about robot healing when it was introduced that I ignore it altogether, as engi you're already VERY well equipped to heal robots already if a corpsman drags a dead one to you. I'd not miss that aspect one bit if it got removed from dispenser.

>in fact this could allow it to be a bit stronger while not being free

It doesn't need to be any stronger in what it does. No offense, but that idea feels like promising buffs purely to accomodate a nerf with no guarantee that it will ever get said buff anyway. People were already screaming powercreep on the storage alone.

Aquilar:
@Losenis  then what would be your compromise then. with the vendor changes I can easily make either the radio pack or the dispenser while sitting in the clothing vendor also cost points?

Losenis:
Honestly? None. I don't see why they'd need to cost points given 10 mins after shutters open you'd have them coming from req anyway. It's just a chore for the engineer to stick close to fob camping the first miner till they actually get the tele or whatever else, assuming they're not suddenly at the mercy of a bald or nonexistent RO that is more pain than an actual balance concern.

Aquilar:
well this is one option on functionally reducing the force multiplier of engineers while not nerfing the items in question, because the items in question are not an issue.
They become an issue because engineers can take multiple power items at the same time. Without in some way limiting how many power items an engi can take to the field, outside of mucking with points totals, this issue will keep boiling up as if engineers get yet another power item masquerading as a free gear in the clothes vendor

Arctrooper8:
Why the reverts on loadout code? It feels unnecessary and more like a loss for potential future loadout vendor mechanics than anything. Would it not be possible to add instead of removing?

Losenis:
>reducing the force multiplier of engineers

That's kind of the point of support roles, to be force multipliers.

FosterGrime-Heymynameslouie:
> Why the reverts on loadout code? It feels unnecessary and more like a loss for potential future loadout vendor mechanics than anything. Would it not be possible to add instead of removing?

yeah this seems a bit out of place considering the pr only touches on dispensor, or maybe im missing something. 

I can see the reasoning as to why dispenser could be selection limited but the revert on loadout vendor seems a bit strange, maybe! 

not touching on the dispenser part though, there's good arguments between keeping it as it is and going with the solution this pr provides.

Aquilar:
@FosterGrime-Heymynameslouie 
I didn't touch any of the Loadout vendor just the gear and cloth vendor so I move points because to being a general pool rather than a list of pools that basically no one uses or has even a reason to use. or the other circumstance, separated pools of points on multiple categories which again no one uses and I can not think of a reason for anyone to use.

FosterGrime-Heymynameslouie:
> @FosterGrime-Heymynameslouie I didn't touch any of the Loadout vendor just the gear and cloth vendor so I move points to being a general pool rather than a list of pools that basically no one uses or has even a reason to use. or the other circumstance, separated pools of points on multiple categories which again no one uses and I can not think of a reason for anyone to use.
> 
> edit: I actually need to fix that because it lets people cheese the system

oh, alright! my bad, cool code clean up is cool

Al-1ce:
Making this big of a change based solely on dispensers is not healthy.

I will admit that Engi needs an overhaul, but not like this.

Al-1ce:
The way I see it, this will encourage engineers to actually bring metal/sandbags instead of focusing on whatever retarded meme build.

I've seriously had encounters with engineers who never bring turrets, tools, or don't evem have metal. This is for the overall good of the caste as far as I'm concerned.

Aquilar:
could get the bugs with loadout to go away, well this is a failure.

EvilMelonTheSecond:
> "I think this a good compromise from straight up taking toys away from engineers while reigning in their ability as a force multiplier by bringing multiple strong items to the field."

This is implying that there is an issue of engineers having too many strong items they can carry around in the field while most of those options occupy the backpack slot and the point costs make them already mutually exclusive. Remember that every point spent on these power items is a point not spent on cheap metal and plasteel. These items are only powerful at roundstart where req can't afford to buy them with all the requests piling on and not much metal is needed to fortify outposts.

Autisem:
other then style there is not much reason to bring dispencer compared to teleporter because you can leave all heavy stuff near the FOB side of the teleporter and go back and fourth on whim to get stuff. them needing to be mutuealy exclusive would not help. This would also make it so you cant buy new one IF they get destroyed(which invetivably happens)

I dont see the point of this

Aquilar:
I'm closing this so I can start it over on a fresh branch, I screwed up the coding and it's better if I started over so it's useable by all roles and not just kicking it down the road for someone else to do.

main problems are: 
-getting both progress bars in the vendor UI to appear at the same time
-finding away to retain points pools per category as lumi requested
-getting points pools and choice pools to work together so items may be limited but also cost the user points.

PR #11395, opened 2022-10-31
Arty range fix

breadhunt:
## About The Pull Request

Replaces several instances of get_dist() with get_dist_euclide() in artillery code.
Get_dist() has a cap of 127 tiles (thanks BYOND) which meant artillery shots had an effective range of 127 tiles (~63 on diagonals). Lest we forget the many pushes decimated by BYOND's cruel gaze.

Previously, the max_range on the shell datum didn't actually affect the range of the shot. It now does. The default for all shells has been set to 1000 so this doesn't actually change gameplay currently, but the code's ready if you want to.

Also alters the screenshake check to use cheap_get_humans_near/cheap_get_xenos_near instead of checking if every single mob is next to the mortar or not, which SHOULD be slightly more efficient.

## Why It's Good For The Game

Fixes good

## Changelog
:cl:
fix: Artillery shots no longer have a pitiful range of 127, exploding way before their intended target. Spread will be significantly more consistent now
fix: The max_range on shell datums is actually functional now
balance: Default shell max_range 250 -> 1000, because it's intended to be infinite. PR a change for this if you want
/:cl:

### DIFF
--- a/code/game/objects/machinery/mortar.dm
+++ b/code/game/objects/machinery/mortar.dm
@@ -225,9 +225,9 @@
 ///Start firing the gun on target and increase tally
 /obj/machinery/deployable/mortar/proc/begin_fire(atom/target, obj/item/mortal_shell/arty_shell)
 	firing = TRUE
-	for(var/mob/M in (cheap_get_humans_near(src, 7) + cheap_get_xenos_near(src,7)))
-		shake_camera(M, 1, 1)
-
+	for(var/mob/M in GLOB.player_list)
+		if(get_dist(M , src) <= 7)
+			shake_camera(M, 1, 1)
 	switch(tally_type)
 		if(TALLY_MORTAR)
 			GLOB.round_statistics.mortar_shells_fired++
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/mortar.dm
```diff
@@ -223,11 +223,11 @@
...
 ///Start firing the gun on target and increase tally
-/obj/machinery/deployable/mortar/proc/begin_fire(target, obj/item/mortal_shell/arty_shell)
+/obj/machinery/deployable/mortar/proc/begin_fire(atom/target, obj/item/mortal_shell/arty_shell)
 	firing = TRUE
-	for(var/mob/M in GLOB.player_list)
-		if(get_dist(M , src) <= 7)
-			shake_camera(M, 1, 1)
+	for(var/mob/M in (cheap_get_humans_near(src, 7) + cheap_get_xenos_near(src,7)))
+		shake_camera(M, 1, 1)
+
```
TiviPlus, line 230: I dont think this is cheaper as cheap_get_humans_near & cheap_get_xenos_near just looper over the entire human/xeno lists for that zlevel

TiviPlus (changes requested):
rest lgtm I guess

Lumipharon:
This is what was causing the most severe arty offset shittery, honk.

PR #11396, opened 2022-10-31
Adds a limiter for CAS hotkeys

breadhunt:
## About The Pull Request

Currently for CAS, while the window is selected, WASD and arrow keys will change the directions of CAS guns firing (N/E/S/W)
These are the same keys for movement. When CAS moves with the window selected, this spams tgui with inputs constantly, which kicks the player for topic spam.

## Why It's Good For The Game

Crashes bad

## Changelog
:cl:
fix: CAS no longer spams directional changes. This should prevent people being kicked for topic spam
/:cl:


tgui/packages/tgui/interfaces/MarineCasship.tsx
```diff
@@ -60,7 +61,7 @@ export const MarineCasship = (props, context) => {
           if (keyCode === KEY_6) {
             act('deselect');
           }
-          if (data.plane_state !== 0) {
+          if ((data.plane_state !== 0) && (limiter === false)) {
```
TiviPlus, line 64: Wouldnt just checking if the directions are the same be better? 
You wouldnt miss rotations randomly and it wouldnt topic spam

breadhunt: Yeah I'm dumb, will change it tomorrow

TiviPlus:
good enough

PR #11405, opened 2022-11-01
Adds radial menu support to medical tweezers, chain-tweezing.

RubberRats:
## About The Pull Request

This PR adds support to the corpsman's medical tweezers for the same radial menu that medkits and surgery use. This PR also makes it so you will continue attempting to remove shrapnel as long as your patient has any in them.

## Why It's Good For The Game

Makes using the tweezers slightly less inconvenient. Also it was weird that virtually every other limb-based medical thing supported the radial menu setting but the tweezers didn't.

## Changelog
:cl:
qol: The corpsman's medical tweezers now support radial menus and will continue attempting to remove shrapnel if the patient still has any.
/:cl:

### DIFF
--- a/code/datums/elements/shrapnel_removal.dm
+++ b/code/datums/elements/shrapnel_removal.dm
@@ -24,7 +24,7 @@
 		M.balloon_alert(user, "You only know how to remove shrapnel from humans!")
 		return
 	var/mob/living/carbon/human/target = M
-	var/datum/limb/targetlimb = target.get_limb(user.zone_selected)
+	var/datum/limb/targetlimb = user.client.prefs.toggles_gameplay & RADIAL_MEDICAL ? radial_medical(target, user) : target.get_limb(user.zone_selected)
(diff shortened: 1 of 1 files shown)


Lumipharon:
In the case of tweezers, I think this is actually not a good thing. You can repeatedly click to spam multiple uses, so having to navigate a radial menu will make pulling large amounts of shrap hell.

Of course if you wanted to code tweezers to use similar logic to kits, then it would be sweet.

RubberRats:
> In the case of tweezers, I think this is actually not a good thing. You can repeatedly click to spam multiple uses, so having to navigate a radial menu will make pulling large amounts of shrap hell.
> 
> Of course if you wanted to code tweezers to use similar logic to kits, then it would be sweet.

I could probably write some code for chain-tweezing without too much trouble. Should be able to add it sometime tomorrow assuming nothing comes up.

TiviPlus (approved):

Lumipharon (approved):

PR #11408, opened 2022-11-01
Infections are Slightly Faster

Ben10083:
## About The Pull Request
counter now increases by 3.3 instead of 2.5. Doing a small change to avoid having it go TOO fast.

## Why It's Good For The Game
Huggers are currently in the unfortunate position of being a relic of the capture loop xenos used to have, unfortunately, maintainers have stated that increases to burrowed larva or psy points when infected is a no-go, so making the incubation faster since there are a myriad of tools that can be used to slow down/remove the infection, not to mention that a marine can usually reach a medic or doctor to help deal with the infection in the event of a hug.

## Changelog
:cl:
balance: Infection from facehuggers now progress faster.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/embryo.dm
+++ b/code/modules/mob/living/carbon/xenomorph/embryo.dm
@@ -68,7 +68,7 @@
 /obj/item/alien_embryo/proc/process_growth()
 
 	if(stage <= 4)
-		counter += 2.5 //Free burst time in ~7/8 min.
+		counter += 3.3 //Free burst time in ~6/7 min.
 
 	if(affected_mob.reagents.get_reagent_amount(/datum/reagent/consumable/larvajelly))
 		counter += 10 //Accelerates larval growth massively. Voluntarily drinking larval jelly while infected is straight-up suicide. Larva hits Stage 5 in exactly ONE minute.


revelation-8468:
carrier powercreep?

cringe

Xsdew:
Another change you could do is associated with being set on fire preventing getting hugged. Help intent clicking is implemented meaning its much simpler to remove a hugger on the mask slot.

Lumipharon:
Hugger time was already severely reduced.

Ben10083:
Hugger infections are not too often really, this just makes the marine more incentivized to deal with the infection sooner.

Pariah919 (commented):
6 to 7 minutes seems okay to me, could be okay to test.

BraveMole (changes requested):
That is not a proper what is good for the game.

Ben10083:
> That is not a proper why it is good for the game.

I discussed it with others and overall infections are very low priority when it comes to treatment. Yeah if you don't treat it you get dead for a bit before you get a revive, but the gestation period if slow enough where you have alot of leeway in terms of when you actually need to deal with the infection. I was hesitant on doing any significant increase, but having the infection go faster makes it the bigger threat its supposed to be. (not to mention how hard it can be to actually get the infection in the first place).

Starfirejordan:
> > That is not a proper why it is good for the game.
> 
> I discussed it with others and overall infections are very low priority when it comes to treatment. Yeah if you don't treat it you get dead for a bit before you get a revive, but the gestation period if slow enough where you have alot of leeway in terms of when you actually need to deal with the infection. I was hesitant on doing any significant increase, but having the infection go faster makes it the bigger threat its supposed to be. (not to mention how hard it can be to actually get the infection in the first place).

as someone who has played marine this is a good change, as hugger are still laughable, what it takes to hug someone and then have them even get pregged THEN for the CHANCE for them to even burst

EX: Fire or help intent click to remove hugger to prevent pregging, shooting the hugger upon seeing it (easy) yata yata.. too much leeway for marines to keep fighting, will obviously increases cope as "oh no i cant keep fighting! and have to go shipside or ask medic groundside? this sucks!!"

Ben10083:
Going to hold off on this PR for now as I wait to see how carrier ability and all that affects things

PR #11413, opened 2022-11-01
Mech health nerfed

Kuro020:
## About The Pull Request

Nerfs overall mech health, buffs heavy mech chest slowdown slightly (so slightly it probably doesn't even make a difference but the numbers not being symetrical were bothering me)

Stats (taking into account a full set instead of mixed parts)
Light: 1500 hp >> 1300hp
Medium: 2750 hp >> 2200hp
Heavy: 4000hp >> 3200hp

## Why It's Good For The Game

Less tanky mechs, considering how easy they are to repair they shouldn't be this tanky

## Changelog
:cl:
balance: Mech overall health reduced
/:cl:


Losenis:
Good idea if the control module breaking removal is merged

PR #11423, opened 2022-11-01
Paracetamol no longer purges Tricordrazine

lKiseki:
## About The Pull Request
Paracetamol will no longer purge Tricordrazine.

The implications of this change are as follows;
- Allows players to have a viable alternative to the usual bica/kelo/tram trinity, in the form of paracetamol / oxycodone / tricordrazine, since the latter chem can counteract oxycodone's tox loss. Dylovene cannot do this since it purges paracetamol.
- Actually enables very niche builds such as Vali paracetamol builds, by allowing them to counteract oxycodone's consequence via tricordrazine.

## Why It's Good For The Game
- Adds more variety to chem gameplay by giving players viable alternatives to the usual everyday chems.
- Enables niche Vali builds based on paracetamol, which in turn means more variety for Vali players.
- Paracetamol gains more relevance, instead of being a forgotten and unused chem that nobody cares about.

## Changelog
:cl: Lewdcifer
balance: Paracetamol no longer purges Tricordrazine.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -86,7 +86,7 @@
 	color = "#cac5c5"
 	scannable = TRUE
 	custom_metabolism = REAGENTS_METABOLISM * 0.125
-	purge_list = list(/datum/reagent/medicine/kelotane, /datum/reagent/medicine/tricordrazine, /datum/reagent/medicine/bicaridine)
+	purge_list = list(/datum/reagent/medicine/kelotane, /datum/reagent/medicine/bicaridine)
 	purge_rate = 1
 	overdose_threshold = REAGENTS_OVERDOSE*2
 	overdose_crit_threshold = REAGENTS_OVERDOSE_CRITICAL*2


QualityVan:
In terms of making para work with oxy, you could just buff its tox healing to 0.2. That'd match it to the brute/burn healing anyway.

lKiseki:
> In terms of making para work with oxy, you could just buff its tox healing to 0.2. That'd match it to the brute/burn healing anyway.

True, but it feels better from a balance standpoint to force players to use Tricordrazine, since that's 3 chems (which do count towards the debuffs granted when you go over the chem cap). Can always change it to work like this if a maintainer disagrees, though.

Derrick2Electricboogaloo:
I think making para heal .3 or .4 would make it more viable as well. I do not mind if trico purges it personally. It pairs well with nanites if you can handle the blood loss.

maikilangiolo:
Seems like the solution is either this PR, which is good, or improving paracetamol as a single use, all-purpose chem. Right now it's very lacking, so either it needs synergy with a second chem (tricord) or has to be better on its own.

Pariah919 (commented):
Give this a try, I'd say. Would give more choices than just the gamer cocktail of BKTT.

PR #11424, opened 2022-11-01
oxycodone now gives 100 pain resistance over 50.

Pariah919:
## About The Pull Request
Oxy has 100 pain resist over 50.

## Why It's Good For The Game
Oxycodone is super mediocre, and barely worth carrying in its current state, it doesn't give you enough pain resist to wobble away from whatever just beat the crap out of you unless you double dose on tramadol as well.

## Changelog
:cl:
balance: Oxycodone gives 100 pain resistance instead of 50 now.
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -13,6 +13,8 @@
 #define PAIN_REDUCTION_MEDIUM -25 //synaptizine
 #define PAIN_REDUCTION_HEAVY -35 //paracetamol
 #define PAIN_REDUCTION_VERY_HEAVY -50 //tramadol
+#define PAIN_REDUCTION_SUPER_HEAVY -100 //oxycodone
+
 
 #define PAIN_REACTIVITY 0.15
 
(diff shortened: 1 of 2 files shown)


Naaanii:
200 please.

Al-1ce:
> 200 please.

Not even going to explain why that's a stupid idea. No. Just, no.

Lumipharon:
In what situation are you not going to also have tram though?

Al-1ce:
> In what situation are you not going to also have tram though?

Tram gives 50. Neuraline gives 100. I think stacking painkillers should give diminished returns somewhat to begin with but I think somewhere around 85-75 would be better.

Jalopkomit18:
I thought oxycodone always had higher pain resistance than tramadol. If this wasn't the case, then what was the point of using oxycodone over tramadol besides stam on inject and long pain?

Al-1ce:
> I thought oxycodone always had higher pain resistance than tramadol. If this wasn't the case, then what was the point of using oxycodone over tramadol besides stam on inject and long pain?

Those were exactly the reasons. Also because you can mix it with tramadol for any "oh shit" moments. It seems to be good at 100.

adamkad1:
Fun.

PR #11426, opened 2022-11-01
Rebalances Paracetamol bonuses granted by Vali boost

lKiseki:
## About The Pull Request
The following changes aim to rebalance Paracetamol to be a more viable alternative to bica/kelo/tram, and should be considered an overall buff, since it was quite underwhelming.
These changes are as follows:
- Movement speed bonus removed.
- Brute and burn healing are now increased by 10%.
- Stamina regen bonus unchanged.

## Why It's Good For The Game
- Paracetamol now fulfills the role of Bicaridine and Kelotane in Vali boosts, by increasing brute and burn healing. This further enables a certain variety in Vali builds.
- Paracetamol's speed bonus amounted to nothing, practically speaking. Felt better to remove it in favor of the above, especially since we don't really need faster ungas.

## Changelog
:cl: Lewdcifer
balance: Paracetamol bonuses granted by Vali boosts changed. Movement speed bonus removed, but brute and burn healing are now increased by 10%.
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -79,7 +79,7 @@
 	var/static/list/reagent_stats = list(
 		/datum/reagent/medicine/bicaridine = list(NAME = "Bicaridine", REQ = 5, BRUTE_AMP = 0.1, BURN_AMP = 0, TOX_HEAL = 0, STAM_REG_AMP = 0, SPEED_BOOST = 0),
 		/datum/reagent/medicine/kelotane = list(NAME = "Kelotane", REQ = 5, BRUTE_AMP = 0, BURN_AMP = 0.1, TOX_HEAL = 0, STAM_REG_AMP = 0, SPEED_BOOST = 0),
-		/datum/reagent/medicine/paracetamol = list(NAME = "Paracetamol", REQ = 5, BRUTE_AMP = 0, BURN_AMP = 0, TOX_HEAL = 0, STAM_REG_AMP = 0.2, SPEED_BOOST = -0.1),
+		/datum/reagent/medicine/paracetamol = list(NAME = "Paracetamol", REQ = 5, BRUTE_AMP = 0.1, BURN_AMP = 0.1, TOX_HEAL = 0, STAM_REG_AMP = 0.2, SPEED_BOOST = 0),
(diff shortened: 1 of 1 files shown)


Al-1ce:
I've never used paracetamol vali for any reason except the movement speed + stamina. This just feels like a nerf.

Coastfront:
Doubling down on what 1ce said, paracetamol without nanites was the choice between go fast with stamina or have more healing ontop of greenblood heal for vali, making it just another generic "vali get health back" chem for "vali get health back" builds doesent really "Opens more build varieties for vali" so much as it "Enables more generic healing builds for vali"

Al-1ce:
@Coastfront bascially hit the nail on the head here. More healing chemicals does the exact opposite of giving vali variation. The playerbase wants to see something beyond a % buff to recovery for once.

Al-1ce:
> making it just another generic "vali get health back" chem for "vali get health back" builds doesent really "Open more build varieties for vali" so much as it "Enables more generic healing builds for vali"


[Coastfront](https://github.com/Coastfront) really hit the nail on the head here. What the playerbase wants with variation is something beyond a % bonus to recovery rates.

lKiseki:
0.1 slowdown amounts to basically nothing, and the +20% stamina regen is already doing funny things like compensating against neuro and extending your default run. I don't think you'll be needing that slowdown reduction at all. Additionally, do remember that paracetamol prevents the usage of chems such as Bicaridine and Kelotane, so you need to account for that. It's not like I'm just increasing the regen and removing the other things it has, just doing a tradeoff between a speed bonus nobody's ever felt before or some form of regeneration so it can compete with the holy trinity of bica/kelo/tram.

Al-1ce:
> 0.1 slowdown amounts to basically nothing, and the +20% stamina regen is already doing funny things like compensating against neuro and extending your default run. I don't think you'll be missing that slowdown reduction at all. 

Most armor slowdown ranges from 0.2 to 0.3. It is noticable, you're just not paying much attention/too heavy for it to make a difference.

>Additionally, do remember that paracetamol prevents the usage of chems such as Bicaridine and Kelotane, so you need to account for that.

That's the downside, yes. It would be better if the slowdown removal was higher, but it isn't. Even still, it is better than what you propose, which instills the same monotanous valishit we've seen for ages.

>It's not like I'm just increasing the regen and removing the other things it has, just doing a tradeoff between a speed bonus nobody's ever felt before or some form of regeneration so it can compete with the holy trinity of bica/kelo/tram.

I'm aware of what your doing. And it isn't adding any variation what-so-ever. It's stale and just follows the cookie-cutter properties of every other drug on the menu. No good.

PR #11430, opened 2022-11-02
kill the shield

tyeagg:
## About The Pull Request
shields removed from vending machines and req still available in hvh

image

## Why It's Good For The Game
makes marines way too tanky, nerfing will make it useless

## Changelog
:cl:
del: remove shields from vending and req
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -79,7 +79,6 @@
 			/obj/item/weapon/twohanded/spear/tactical = -1,
 			/obj/item/weapon/twohanded/spear/tactical/harvester = -1,
 			/obj/item/weapon/powerfist = -1,
-			/obj/item/weapon/shield/riot/marine = 6,
 			/obj/item/weapon/shield/riot/marine/deployable = 6,
 		),
 		"Sidearm" = list(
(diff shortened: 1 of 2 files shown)


FosterGrime-Heymynameslouie:
thank you for taking the time to censor his name and image
idtentity theft is bad and we must do everything we can to prevent it

CabbageOnTable:
Salt pr?

Lumipharon:
> thank you for taking the time to censor his name and image idtentity theft is bad and we must do everything we can to prevent it

Who is this mystery individual? We must unravel their secrets.

Mrrpip:
Well. That's a yikes from me, chief.

But on a less jokey note, does this really make marines too tanky? Unless you're going AAA tier powergaming (Or robot, which has it's own set of drawbacks) it's countered pretty well be defiler, carrier and sentinel to some extent.

Two shields are probably over the top for tanking, that I agree with, but with the innate slowdown, you're not really chasing much, let alone the fact you can't carry many weapons besides a sword due to a lack of storage space balances it out.

Synth and Shield is also strong. But that's a symptom of synth having increased melee skill and no proper marine armor bar the flak jacket, so of-course they'd benefit a ton from it. I'd reckon that mess would still be a pain, even with the lesser deployable shield.

I mean, yeah, not much will be lost if shields are removed, but, well. It's a fun little thing to use, I guess.

DerpiestBruh:
:(

tyeagg:
thought of a rework

Zar000:
Speedmerge rn

567Turtle:
Shields slow you down a bunch and take up a whole hand, which you, y'know, kind of need to be able to shoot something, this just seems like you got killed by a shield marine and then frantically began backspacing code through the tears in your eyes. There's only 6 of them anyway so yknow it's not like they're being given to every marine.

Derrick2Electricboogaloo:
Yeah this is some omega cope, only get 5, lots of slowdown. who plays hvh?! Basically deleting it from the came

TiviPlus:
Shields have downsides I dont see them as something crazy

PR #11436, opened 2022-11-03
Facehugger Impregnates Faster

Ben10083:
## About The Pull Request
Impregnate timer shortened to 10 seconds from 12.

## Why It's Good For The Game
As it stands, with the facehugger being easier to remove with just help intent, marines can reliably have the hugger removed as long as they are in a group. This change should not change that dynamic, just adjusts the window of time the marine has to take into account the ease of removal. This does not affect gestation period, only how long until an infection begins.

## Changelog
:cl:
balance: Hugger impregnate timer shortened from 12 seconds to 10.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
@@ -1,7 +1,7 @@
 ///After how much time of being active we die
 #define FACEHUGGER_DEATH 10 SECONDS
 ///Time it takes to impregnate someone
-#define IMPREGNATION_TIME 12 SECONDS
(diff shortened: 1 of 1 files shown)


srxtreme:
You should probably slightly reduce the amount of time it takes to drag a hugger off, considering how much better is the plasma pistol at hugger removal. Since this will probably result in even more people using PP instaremoval.

Ben10083:
> You should probably slightly reduce the amount of time it takes to drag a hugger off, considering how much better is the plasma pistol at hugger removal. Since this will probably result in even more people using PP instaremoval.

eh, sacrificing weapon strength for the utility of the PP sounds fine. As for time for removal, it is already quite fast already.

PR #11439, opened 2022-11-03
Lowers the inventory size of thermal dampening tarps

lKiseki:
## About The Pull Request
Per title. Tarps have had their inventory size reduced, allowing more storage options to carry them.

## Why It's Good For The Game
This thing is so underwhelming nowadays, since xenos have a lot of tools to catch people in tarps.
Tarps would see more use if more inventory storages could hold them, and would also put them in line with other similar options, such as rollerbeds.

## Changelog
:cl: Lewdcifer
balance: Thermal dampening tarp inventory size reduced. More inventory storage options can hold them now.
/:cl:


Pariah919 (changes requested):
Eh. Indifferent. I think tarp is still strong, just underused.

PR #11440, opened 2022-11-03
Explosive pouches maximum item size increased

lKiseki:
## About The Pull Request
Per title. I find it weird that explosive pouches cannot hold certain items such as recoilless rockets, so here we are.

## Why It's Good For The Game
Explosive pouches can hold recoilless rockets, giving those builds an alternative in storage options.
Also, explosive pouches are grossly underused. This'll give them an edge.

## Changelog
:cl: Lewdcifer
balance: Explosive pouches can now hold larger items (such as recoilless rockets).
/:cl:

### DIFF
--- a/code/game/objects/items/storage/pouch.dm
+++ b/code/game/objects/items/storage/pouch.dm
@@ -312,7 +312,7 @@
 	desc = "It can contain grenades, plastiques, mine boxes, and other explosives."
 	icon_state = "large_explosive"
 	storage_slots = 4
-	max_w_class = 3
+	max_w_class = 4
 	can_hold = list(
 		/obj/item/explosive/plastique,
 		/obj/item/explosive/mine,


Lumipharon:
RR is free round start, them not fitting in the explosive pouch is a deliberate balancing factor.

lKiseki:
> RR is free round start, them not fitting in the explosive pouch is a deliberate balancing factor.

Yes, I'm aware, hence why I made this PR. I find it weird that you can't just use one of your pouch slots in an explosive pouch to carry three rockets. Feels like a decent tradeoff, and I don't really understand why it'd be denied.

Pariah919 (commented):
I don't personally like RR rockets fitting in pouches as its a deliberate design change and this is mainly targeted to that.

RyderFive:
RR being balanced around forcing people to use a snowflake satchel that can only hold the rockets besides crates, and storage creeped to hold the RR without really fixing the issues it has/had is pretty meh. 

Ignoring how bad it was before the recent buffs, and the rework it kinda needs. Giving more storage options, and undoing the weird snowflake storage creep satchel dependency would be a better direction for the RR imo.

Not to mention have fun trying to get it during highpop times since there's only 2

BraveMole (changes requested):
I think the consensus among maintainers was:

Ok for rockets in pouch, but 4 rockets probably too much.

How about a new pouch holding rockets that can hold 2-3 ?

PR #11445, opened 2022-11-03
Defiler now consumes plasma after noxius gas has channeled

Goblinu:
## About The Pull Request
Defiler now consumes plasma after noxius gas has been channeled instead of before/during
Defiler still gets its plasma consumed if stunned during channel

## Why It's Good For The Game
Noxius gas most of the time gets cancelled due to server ticks, ping or other lag issues, having it get canceled on the final tick and consuming plasma is very obnoxious and just makes the defiler more annoying and harder to play than it already is. 

## Changelog
:cl:

balance: defiler now consumes plasma after noxius gas has channeled

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
@@ -207,7 +207,7 @@
 
 	GLOB.round_statistics.defiler_neurogas_uses++
 	SSblackbox.record_feedback("tally", "round_statistics", 1, "defiler_neurogas_uses")
-	
+
 	X.visible_message(span_xenodanger("[X] emits a noxious gas!"), \
 	span_xenodanger("We emit noxious gas!"))
 	dispense_gas()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
```diff
@@ -206,7 +206,8 @@
 
 	GLOB.round_statistics.defiler_neurogas_uses++
 	SSblackbox.record_feedback("tally", "round_statistics", 1, "defiler_neurogas_uses")
-
+	
+    succeed_activate()
 	X.visible_message(span_xenodanger("[X] emits a noxious gas!"), \
```
Penwin0, line 211:
```suggestion
	succeed_activate()
	X.visible_message(span_xenodanger("[X] emits a noxious gas!"), \
```

Naaanii:
Would need to make sure that it consumes plasma if Defiler gets stunned mid emission as that's how it was made to be if I remember correctly.

Goblinu:
> Would need to make sure that it consumes plasma if Defiler gets stunned mid emission as that's how it was made to be if I remember correctly.

That should fix it i think

TiviPlus (approved):

PR #11446, opened 2022-11-03
Raises Queen Leader Slots

Sunsidec:
## About The Pull Request

Raises Queen leader slots from 2/3/3/4 at young/mature/elder/ancient to 2/3/4/5.

## Why It's Good For The Game

Makes more sense, spreads xenos pheremones a bit more, rewards maturing for Queen.

## Changelog
:cl:
balance: Queen leader slots adjusted.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/queen/castedatum_queen.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/queen/castedatum_queen.dm
@@ -140,7 +140,7 @@
 	aura_strength = 4.5
 
 	// *** Queen Abilities *** //
-	queen_leader_limit = 3
+	queen_leader_limit = 4
 
 /datum/xeno_caste/queen/ancient
 	caste_desc = "The most perfect Xeno form imaginable."
@@ -173,7 +173,7 @@
 	aura_strength = 5
 
 	// *** Queen Abilities *** //
-	queen_leader_limit = 4
+	queen_leader_limit = 5
 
 //same stats as ancient
 /datum/xeno_caste/queen/primordial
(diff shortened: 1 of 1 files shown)


Pariah919 (changes requested):
Why though? This mostly affects lategame Queen. Pheromone towers exist now, there's no reason to increase the amount of leaders on the map on lategame to me at present time, honestly.

Sunsidec:
> Why though? This mostly affects lategame Queen. Pheromone towers exist now, there's no reason to increase the amount of leaders on the map on lategame to me at present time, honestly.

For the sake of neatness! 2/3/3/4 looks strange, when everything else increases with maturity. It's either that or turn it to 1/2/3/4, which I don't think is a very nice thing to do with the current hell xenos are being put through. It's also a rare occasion you see a pheremone tower set to recovery over warding or frenzy due to queen pheremones generally being present at whatever fight has stalemated long enough for pheremone towers to be safely placed, and would enable an extra leader to be assigned to make flanks more of a threat.

Naaanii:
With phero towers, don't think Queen needs more leader slots honestly.

Though Queen may have one of the best Primo abilities in the game, I wouldn't be opposed to adding an extra leadership slot for Primo Queen since it's rather hard to obtain and all the more punishing should she die.

Losenis:
More leader slots are good due to the rally too, a xeno shouting "POD OPS IN THE FUNNY MINER UHHHHHHHHH NORF UHHH BY THE RED DOORS" can only help so much.

DerpiestBruh:
sus pr

TiviPlus (changes requested):
How about make it 4 accross the board

Sunsidec:
> How about make it 4 accross the board

Seems silly as generally everything scales across maturation, this should be no exception, also see @Losenis' comment for additional justification re. rallies.

Sunsidec (commented):
Updated to 4/4/4/4 across the board which is the preferable alternative to a closed PR.

PR #11447, opened 2022-11-03
Marine masks no longer protect against ash/sandstorms

Sunsidec:
## About The Pull Request

Masks no longer protect against inclement weather, meaning marines outside are at risk.

## Why It's Good For The Game

Anyone with more than two minutes of playtime will wear a mask as there is no downside to doing so, and therefore this mechanic only negatively impacts one side whilst letting the other roam free. This will hopefully force both sides into more close encounters within buildings while taking shelter from storms, whilst at the same time not removing the ability for both to quickly dart in and out of the weather outside or shore up defences with haste.

## Changelog
:cl:
balance: Masks no longer make marines immune to inclement weather.
/:cl:

### DIFF
--- a/code/datums/weather/weather_types/ash_storm.dm
+++ b/code/datums/weather/weather_types/ash_storm.dm
@@ -77,8 +77,6 @@
 	while (L && !isturf(L))
 		if(iscarbon(L))// if we're a non immune mob inside an immune mob we have to reconsider if that mob is immune to protect ourselves
 			var/mob/living/carbon/the_mob = L
-			if(the_mob.wear_mask)	//safety first
-				return TRUE
 			if(the_mob.status_flags & INCORPOREAL)
 				return TRUE
 		L = L.loc //Check parent items immunities (recurses up to the turf)


Losenis:
Is there a possibility to make damage from sandstorms be only nonbleeding brute? At the moment they cause bleeding so you're gonna end up drained of your blood by 10 bleeding bodyparts. Also due to it being brute rather than burns like LV's acid rain, someone caught outside runs the risk of getting fractured bones in freak instances, which is pretty stupid.

Sandstorms are stupidly brutal for marines compared to acid rain. Another idea is changing it to burn damage and pretend it's superheated particles or friction burns or something.

Lumipharon:
I'm not personally opposed to this, but sand kills unga rather quickly, so this'll be entertaining.

Strayed-Raven:
You would have to rework sand itself to ever get this reasonable.

Losenis:
image

TiviPlus (changes requested):
This code has not been tested and does not work

TiviPlus:
On a note of what you want to be happening please respond to criticism in the comments

Sunsidec:
> This code has not been tested and does not work

Hubris, will fix this tomorrow. 


> On a note of what you want to be happening please respond to criticism in the comments

Above

Mechswordsman:
> Is there a possibility to make damage from sandstorms be only nonbleeding brute? At the moment they cause bleeding so you're gonna end up drained of your blood by 10 bleeding bodyparts. Also due to it being brute rather than burns like LV's acid rain, someone caught outside runs the risk of getting fractured bones in freak instances, which is pretty stupid.
> 
> Sandstorms are stupidly brutal for marines compared to acid rain. Another idea is changing it to burn damage and pretend it's superheated particles or friction burns or something.

Not to mention, the sandstorms also cause lung damage when you're not wearing a mask, and if masks don't protect from them, you'll see people more or less instantly get crit from the lung damage and die from that before the bleed even gets bad.

Sunsidec:
> This code has not been tested and does not work

image

Seems to work just fine in execution on my end. Sandstorm deals damage through gas masks (takes 34 times to crit, not blocked or lessened by armour), unless I'm missing something!

> Is there a possibility to make damage from sandstorms be only nonbleeding brute? At the moment they cause bleeding so you're gonna end up drained of your blood by 10 bleeding bodyparts. Also due to it being brute rather than burns like LV's acid rain, someone caught outside runs the risk of getting fractured bones in freak instances, which is pretty stupid.
> 
> Sandstorms are stupidly brutal for marines compared to acid rain. Another idea is changing it to burn damage and pretend it's superheated particles or friction burns or something.

Sandstorms ARE stupidly brutal, taking somewhere around 30 seconds to crit through regular armour but at the same time a lot of Big Red is buildings connected to each other which you can use for short distances. The broken bones are somewhat extreme, but the occurrence of which seems to be massively rare and serve to further discourage travel - I'd like to see the impact of leaving the damage & type as-is and see the impact of such, before making changes to it. Need to keep in mind that any changes to the damage amount would also impact xenos, so changing it to burn might be the preferable option, but that comes with it's own downsides regarding making weather have very little variation / the massive amount of pain burn damage causes to marines.

> You would have to rework sand itself to ever get this reasonable.

Above. As far as I'm concerned this was the original intent, but happy to reconsider and tweak after seeing it in live practice.

> Not to mention, the sandstorms also cause lung damage when you're not wearing a mask, and if masks don't protect from them, you'll see people more or less instantly get crit from the lung damage and die from that before the bleed even gets bad.

If this is a thing, it's not clearly signposted, as the proc referenced is just brute damage.

MereWalker:
Just a pitch in the dark. Maybe adding an additional inconvenience to marines. Filter swapping/changing of oxygen tanks. Granted they should take not much space whatsoever like tiny so as to not screw over space management. Have a slot added to masks that hold the item to allow for "safe breathing". Just a thought. A very inconvenient thought.

Sunsidec:
> Just a pitch in the dark. Maybe adding an additional inconvenience to marines. Filter swapping/changing of oxygen tanks. Granted they should take not much space whatsoever like tiny so as to not screw over space management. Have a slot added to masks that hold the item to allow for "safe breathing". Just a thought. A very inconvenient thought.

That just sounds like adding an extra step to making storms not doing anything again, and the only people falling for it are the aforementioned people with less than two minutes of playtime.

MereWalker:
No no. Still not immune to storms and the such just smoke gas the such.

Sunsidec:
> If you wanted to tone it down just a little you could make sandstorms respect armor instead of being flat damage.

It seemed to be fine enough during the TMs. Forced marines to stay inside, whilst small jaunts still being possible to move between structures. I'm happy enough with the damage as it currently is.

PR #11448, opened 2022-11-03
dispenser fixes

Lumipharon:
## About The Pull Request
Adds the missing throwpass var, so they can be fired past/used as cover.
Corrected the integ to 250, I did actually mean this in the old pr, whoops.

## Why It's Good For The Game
Fix good.

## Changelog
:cl:
fix: dispensers can be fired past like other objects, and integrity is correctly set to 250
/:cl:


Losenis:
They can be shot past?

UNGAS WILL NOT KEEP DELETING THEM ALL THE TIME?

PRAISED BE

PR #11453, opened 2022-11-04
remove semi fire mode from guns with autofire [MDB Ignore]

tyeagg:
## About The Pull Request
removes semi auto fire from guns that have automatic fire

## Why It's Good For The Game
less fire modes to cycle through, less confusing for new players. There are 0 reasons why you want to use semi auto when theres autofire mode

## Changelog
:cl:
del: remove semi auto from automatic fire guns
/:cl:


tyeagg:
i got a really goofy map diff render error

PR #11391, opened 2022-10-31
Bull remastered: The rebullening: The bull within

Lumipharon:
## About The Pull Request
Removing Bull is lame.

Bull can now change direction while charging without primo.
Changing direction mid charge now has a cooldown, so you can change direction, but you can't do so repeatedly in a short period.
 Primo reduces this cool down to make you more agile, so is now far less overwhelming.
Note: You cannot double back on yourself without losing charge, so no more going back and forth to endlessly charge a poor unga.

Primo bull now also gets the stomp ability, since it's going to be notably weaker than before. It also means it can do something other that literally charge and nothing else.

## Why It's Good For The Game
Primobull is bullshit, non primo is bullbad
Addresses both of these points somewhat

## Changelog
:cl:
add: Bulls can now natively change direction when charging except for 180 degrees, on a cooldown
add: Primo bull reduces charge direction change cooldown and has the stomp ability
/:cl:

PR #11459, opened 2022-11-05
Bull Headbutt Adjustments (Primo Nerf Non Nuclear Edition)

Wayland-Smithy:
For the record I personally don't see the need for any changes, but if it helps the overall health of the game I am for it.

## About The Pull Request

Removes the *still* far too long stun from bull headbutt charge, but keeps it as a viable tool both with and without primo.

## Why It's Good For The Game

The current state of headbutt especially with primo is unfun to face as a marine and boring to play as a xeno.

Better than the alternative of nerfing the primo gimmick into the ground when the real issue is the perma stun from headbutt, and maybe keeps primo bull a powerhouse of 1v1s.

## Changelog
:cl:
balance: Bull headbutt charge now staggers and slows instead of stunning.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
+++ b/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
@@ -547,7 +547,9 @@
 		if(CHARGE_CRUSH)
 			Paralyze(CHARGE_SPEED(charge_datum) * 20)
 		if(CHARGE_BULL_HEADBUTT)
-			Paralyze(CHARGE_SPEED(charge_datum) * 25)
+			Paralyze(1) //Token paralyze to drop stuff etc...
+			adjust_stagger(CHARGE_SPEED(charge_datum) * 25)
+			add_slowdown(CHARGE_SPEED(charge_datum) * 10)
 
 	if(anchored)
 		charge_datum.do_stop_momentum(FALSE)


DemonOfHatred:
I mean not really though? you have to charge in and risk instant PB or death via rockets i think its fine.

Wayland-Smithy:
Not every fight is like that. Nobody wants to get perma stunned in a 1v1 which is where I think all the salt towards primo bull comes from. Stun mechanics just feel bad, so any alternative I hope will be welcome.

My goal is to find a compromise that everyone can be happy with that is not as fun killing as #11391

DemonOfHatred:
But the whole strength of being a marine is to be with your team, if you're by yourself you're going to die anyway.

Losenis:
"But the whole strength of being a marine is to be with your team"

image

KrabSpider:
Thank you. I am terribly sorry for keeping the stun that long.

KrabSpider:
At that point, just turn headbutt into a subtype of forward charge without the stun. Stagger and slow are only useful if you can actually pull it off immediately on a target you *want* to weaken in that moment. Otherwise you're just putting yourself at unnecessary risk, since youre going to either hit the wrong target, not hit anything at all, or just not end up getting much of anything done at all in the grand scheme of things.

ivanmixo:
This is a buff to primo while making regular bull even more useless. A second is nothing when primo bull can build up speed to stun you again in that time.

FIXTHERADBOXPLEASE:
Seems kinda pointless when there's already a bull rework being made. If something's gotta be done between then I'd honestly prefer just turning off the primo until the other PR's done.

Naaanii:
Headbutt is already in an okayish state and this doesn't solve the plow charge and gore charge spam on primo bull.

Imo the issue is the charge having no cd or increased plasma cost etc. that makes headbutt an issue.

Nerfing all charges when using the primo charge as a less damaging but spammable option while making the primo charge as a toggleable option might be a step in the right direction though.

Wayland-Smithy:
> This is a buff to primo

Right... making something strictly less effective is a buff. I'll have to remember that. Did you even look at the code?

> Imo the issue is the charge having no cd or increased plasma cost etc.

Gore actually has a short CD per victim to fix previous spam funnies. I'm considering extending that and applying to all charge modes and/or adding an additional plasma cost when turning with primo. Additionally having a "agile" toggle while primo and reducing charge damage across the board to compensate for the spam-ability could work.

ivanmixo:
It's a buff to primo because of the other 2 effects you added. You didn't just lower the stun time.

Wayland-Smithy:
Okay I see your point. Sure you will be easier to hit, but only by virtue of not being horizontal on the ground and 100% vulnerable to any and all other xeno attacks as you are now. Would you not rather be able to slowly make your way to reinforcements than be totally immobilized with a slim chance (none if vs me) to pop off a round?

Wayland-Smithy:
Will try a different approach if that other PR seems at all likely of being merged, otherwise it'll just be a sign to no longer play this caste and master the others until they are also nerfed.

PR #11489, opened 2022-11-08
Scatter increase per shot mechanics.

SplinterGP:
## About The Pull Request
What is this ? so, this is basically a scatter increase per shot mechanic, guns now have a min scatter, max scatter, scatter increase and scatter decay values

minimum scatter and maximum scatter is pretty obvious.

scatter increase is how much scatter increases per shot of a gun.

scatter decay is how much scatter decreases per 1/10 of a second not shooting.

attachments are also able to modify those so that should also take account into balance, certain attachments instead of modifying only the static scatter guns had, they can now modify increase, decay, maximum and minimum.

Currently i dind't actually make any gun use this system other than testing it, balancing and numbers need to be done, also, the default values means no gun is actually affected by this system(i also need to check if everything is working like akimbo and all with it)

## Why It's Good For The Game
Adds more depth to scatter mechanics in the game, hopefully opening also more ways to balance and make guns more unique, maybe some scatter/spread pattern stuff could be added as well, who knows.

## Changelog
:cl:
add: Adds more depth to scatter mechanics, now gun scatter increases the more you shoot up to a maximum and 
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -60,9 +60,9 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	///Increases or decreases scatter chance but for onehanded firing.
 	var/scatter_unwielded_mod = 0
 	///Maximum scatter
-	var/max_scatter_mod = 360
+	var/max_scatter_mod = 0
 	///Maximum scatter when wielded
-	var/max_scatter_unwielded_mod = 360
+	var/max_scatter_unwielded_mod = 0
 	///How much scatter decays every X seconds
 	var/scatter_decay_mod = 0
 	///How much scatter decays every X seconds when wielded
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -61,7 +61,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	var/scatter_unwielded_mod = 0
 	///Maximum scatter
 	var/max_scatter_mod = 0
-	///Maximum scatter when wielded
+	///Maximum scatter when unwielded
 	var/max_scatter_unwielded_mod = 0
 	///How much scatter decays every X seconds
 	var/scatter_decay_mod = 0
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -59,6 +59,22 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	var/scatter_mod 	= 0
 	///Increases or decreases scatter chance but for onehanded firing.
 	var/scatter_unwielded_mod = 0
+	///Maximum scatter
+	var/max_scatter_mod = 360
+	///Maximum scatter when wielded
+	var/max_scatter_unwielded_mod = 360
+	///How much scatter decays every X seconds
+	var/scatter_decay_mod = 0
+	///How much scatter decays every X seconds when wielded
+	var/scatter_decay_unwielded_mod = 0
+	///How much scatter increases per shot
+	var/scatter_increase_mod = 0
+	///How much scatter increases per shot when wielded
+	var/scatter_increase_unwielded_mod = 0
+	///Minimum scatter
+	var/min_scatter_mod = 0
+	///Minimum scatter when unwielded
+	var/min_scatter_unwielded_mod = 0
 	///If positive, adds recoil, if negative, lowers it. Recoil can't go below 0.
 	var/recoil_mod 		= 0
 	///If positive, adds recoil, if negative, lowers it. but for onehanded firing. Recoil can't go below 0.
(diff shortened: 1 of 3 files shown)


Strayed-Raven:
Coderman thats all a good bit of text but what does this mean?

SplinterGP:
> Coderman thats all a good bit of text but what does this mean?

i'm gonna post a vid from how it happens on ss14 since it's pretty much similar, i dindt do debug visuals to show it in here so its easier to show from there, but its the same thing

https://user-images.githubusercontent.com/29824735/200673425-ccd0193d-85b4-4fee-93f0-539ff1ef345a.mp4

Lumipharon:
Cool system, although it does raise questions about burst fire more generally. Currently burstfire is higher ROF at the cost of worse scatter and acc. This goes against reality and this scatter system, where it should be lower ROF but better acc/scatter.

Losenis:
If you also add multipliers for 1handing for both increase and decay this could be the perfect system for balancing akimbo usage

Lumipharon:
> If you also add multipliers for 1handing for both increase and decay this could be the perfect system for balancing akimbo usage

one hand already has worse scatter values, and akimbo adds a multiplier. This doesn't really fix akimbo since akimbo is for going fast and shooting people pointblank, where scatter doesn't matter anyway. Shoot someone at like 4 tiles with akimbo smg's on the move and you'll see how much shit misses.

Strayed-Raven:
> > Coderman thats all a good bit of text but what does this mean?
> 
> i'm gonna post a vid from how it happens on ss14 since it's pretty much similar, i dindt do debug visuals to show it in here so its easier to show from there, but its the same thing
> 
>  2022-07-09_04-18-55.mp4

this video is amazing and exactly what I needed. So go ahead and make akimbo great again

SplinterGP:
> If you also add multipliers for 1handing for both increase and decay this could be the perfect system for balancing akimbo usage

theres different values for 1 handed already

FosterGrime-Heymynameslouie:
will this effect bipodee guns though

SplinterGP:
> will this effect bipodee guns though

yes

SplinterGP:
btw there shoulnd't be much point to testmerge this PR now since i dindt add values to any gun, i'll do some after my tests on 13 and 20 of this month

code/modules/projectiles/gun_system.dm
```diff
@@ -1619,8 +1635,12 @@
 
 	if(((flags_item & WIELDED) && wielded_stable()) || CHECK_BITFIELD(flags_item, IS_DEPLOYED) || (master_gun && CHECK_BITFIELD(master_gun.flags_item, WIELDED) && master_gun.wielded_stable()))
 		gun_accuracy_mult = accuracy_mult
+		scatter = max(max(min_scatter, 0), min((scatter + scatter_increase) - ((world.time - last_fired - 1) * scatter_decay), max_scatter))
 		gun_scatter = scatter
 		wielded_fire = TRUE
+	else
+		scatter_unwielded = max(max(min_scatter_unwielded, 0), min((scatter + scatter_increase_unwielded) - ((world.time - last_fired - 1) * scatter_decay_unwielded), max_scatter_unwielded))
+		gun_scatter = scatter_unwielded
```
Lumipharon, line 1643: This is adding a floor to scatter of 0. As well as causing bugs like that bipod one, it also means that anything that effects gun_scatter (instead of scatter) will ensure a positive value, where as currently many guns have negative scatter for the purpose of offsetting those penalties.
i.e, movement, burst, staggered, etc.

Would it be better to have a separate var for 'current scatter' etc, instead of modifying scatter var directly?

Lumipharon (changes requested):
edit: double posted somehow. See below.

code/modules/projectiles/gun_system.dm
SplinterGP: yeah that makes sense, the floor of 0 should be further down which already exists as well, also no need since if you dont setup the increase/decay or a min/max it shoulnd't change the scatter at all, it uses scatter because when the system is in work having scatter var as it was doenst make sense anymore

TiviPlus: pending?

code/modules/projectiles/gun_attachables.dm
```diff
@@ -59,6 +59,22 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	var/scatter_mod 	= 0
 	///Increases or decreases scatter chance but for onehanded firing.
 	var/scatter_unwielded_mod = 0
+	///Maximum scatter
+	var/max_scatter_mod = 0
+	///Maximum scatter when wielded
```
Lumipharon, line 64:
```suggestion
	///Maximum scatter when unwielded
```
assuming thats a typo

code/modules/projectiles/gun_system.dm
```diff
@@ -198,10 +198,26 @@
...
 	var/recoil_deviation = 22.5
-	///How much the bullet scatters when fired while wielded.
+	///How much the bullet currently scattered when last fired.
 	var/scatter	= 4
 	///How much the bullet scatters when fired while unwielded.
 	var/scatter_unwielded = 12
+	///Maximum scatter
+	var/max_scatter = 360
+	///Maximum scatter when wielded
+	var/max_scatter_unwielded = 360
```
Lumipharon, line 208: I'm assuming this is basically just placeholder, but the default should still be something somewhat reasonable (i.e. less than 180)

code/modules/projectiles/gun_system.dm
```diff
@@ -1619,8 +1637,12 @@
 
 	if(((flags_item & WIELDED) && wielded_stable()) || CHECK_BITFIELD(flags_item, IS_DEPLOYED) || (master_gun && CHECK_BITFIELD(master_gun.flags_item, WIELDED) && master_gun.wielded_stable()))
 		gun_accuracy_mult = accuracy_mult
+		scatter = max(max(min_scatter, 0), min((scatter + scatter_increase) - ((world.time - last_fired - 1) * scatter_decay), max_scatter))
```
Lumipharon, line 1640: what's the -1 for? Doesn't that mean decay will increase your scatter if you've fired in the last second?

Lumipharon (changes requested):

code/modules/projectiles/gun_system.dm
SplinterGP: doenst matter much, it is that high just so it doenst care or mess with any gun that doenst have the system implemented

code/modules/projectiles/gun_system.dm
SplinterGP: thats game ticks, it is so it counts 0 on next tick and doenst care about it, think it should be according to firerate though or just not at all and care all ticks

Lumipharon (approved):

GravitronBlue:
Blaming this PR for turning akimbo MP SMGs into zero scatter burst rifles, thank you for breaking gun balance again.

PR #11496, opened 2022-11-09
Weed node variation

Lumipharon:
## About The Pull Request
Adds some variation to weed node animations, and slows them down a bit.
No more field of syncronised pulsing.
## Why It's Good For The Game
Looks a lot better.
## Changelog
:cl:
imageadd: randomised the weed node animation timing
/:cl:


Lumipharon:
Note to self: Lummox is adding a delay arg to animate() in 515 that will be a better replacement for this in the future.

PR #11500, opened 2022-11-09
Tiny perf improvement for guncode

Lumipharon:
## About The Pull Request
Guns won't update their icon unless the icon state is actually changing. There is a slight performance benefit for doing it this way, and since gun are used by 50 people spamming update_icon multiple times per second, it can't hurt. May also further help with returning hyperscale guns.

Ideally byond would automatically do this but uh, byond.
## Why It's Good For The Game
Micro improvement good
## Changelog
:cl:
code: slightly improved gun icon updates
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -470,8 +470,8 @@
 		new_state = base_gun_icon + "_l"
 
 	if(icon_state == new_state)
-		return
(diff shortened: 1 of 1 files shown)


Lumipharon:
Apparently this actually makes it slightly worse, lmao. Well into the trash until I can figure out a way to also account for overlays, which might make it worth it.

PR #11523, opened 2022-11-10
Some mech weapon balance changes

Kuro020:
## About The Pull Request

Nerfs light cannon and minigun, buff pistol and  rebalance shotgun

## Why It's Good For The Game

Lightcannon and minigun are way too good, this should bring them into line, the pistol isnt bad but it's not that desirable compared to the other light weapons so I'm buffing it to make it a better choice, and the shotgun is overall whack, now it has buckshot like CC and less spread, but does less damage as well, also higher fire delay to account for the new CC it got

## Changelog

:cl:
balance: Mech lightcannon and minigun nerfed, pistol buffed, shotgun now stuns but does less damage and fires slower
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -391,8 +391,8 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 
 /datum/ammo/bullet/pistol/mech
 	name = "super-heavy pistol bullet"
-	damage = 40
-	penetration = 15
+	damage = 45
+	penetration = 20
 	sundering = 1
 
 /datum/ammo/bullet/pistol/mech/burst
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/ammo_datums.dm
```diff
@@ -1044,21 +1044,24 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
...
 /datum/ammo/bullet/shotgun/mech/spread
 	name = "super-heavy additional buckshot"
 	icon_state = "buckshot"
 	max_range = 10
-	damage = 50
+	damage = 75
 	damage_falloff = 4
 
+/datum/ammo/bullet/shotgun/mech/on_hit_mob(mob/M, obj/projectile/proj)
+	staggerstun(M, P, weaken = 1, stagger = 1, knockback = 2, slowdown = 0.5, max_range = 3)
```
Lumipharon, line 1063:
```suggestion
	staggerstun(M, proj, weaken = 1, stagger = 1, knockback = 2, slowdown = 0.5, max_range = 3)
```

Lumipharon (changes requested):

PR #11531, opened 2022-11-12
Infections Give Larva Points

Ben10083:
## About The Pull Request
When an embryo reaches stage 2, subsequent stages give 0.2 job points towards the next larva. (This means that a infection that reaches stage 5 gives 0.6 points, stage 6 doesnt give anymore since that is burst)
## Why It's Good For The Game
As it stands, when an infection occurs via huggers, it only serves as a timer for marines to get to surgery, and as such has encouraged the practice of simply ignoring the infection for awhile until around five minutes past, then medevac to get it dealt with. This results in no boon for the xeno exlcuding the marine is now not on the ground (which doesn't do too much considering how fast said marine can come back, esp with drop pod).

This change gives the xenos a small but tangeable benefit to getting the hard to acquire infection, and moves the infection dynamic from a relic of the capture loop to something more fitting of the modern TGMC.

TLDR: Current infection system accounted for the infected to be captures by xenos, current state means infections are not really worth acquiring outside of the occasional burst from the bald marine, this will make xenos get a new larva quicker if the infection gets to the later stages

It is important to note that infections are very hard to acquire in the current state of the game, with infections only occuring via xeno intervention, marine being isolated, or dumb luck.
## Changelog
:cl:
add: Embryos now give 0.2 larva points to xenos when they reach stage 3-5.
/:cl:

### DIFF
--- a/code/datums/gamemodes/game_mode.dm
+++ b/code/datums/gamemodes/game_mode.dm
@@ -467,6 +467,8 @@ GLOBAL_LIST_INIT(bioscan_locations, list(
 		dat += "[GLOB.round_statistics.larva_from_silo] larvas came from silos."
 	if(GLOB.round_statistics.larva_from_cocoon)
 		dat += "[GLOB.round_statistics.larva_from_cocoon] larvas came from cocoons."
+	if(GLOB.round_statistics.larva_from_embryo)
+		dat += "[GLOB.round_statistics.larva_from_embryo] larvas came from infections."
 	if(GLOB.round_statistics.larva_from_marine_spawning)
 		dat += "[GLOB.round_statistics.larva_from_marine_spawning] larvas came from marine spawning."
 	if(GLOB.round_statistics.larva_from_siloing_body)
--- a/code/datums/round_statistics.dm
+++ b/code/datums/round_statistics.dm
@@ -85,6 +85,7 @@ GLOBAL_DATUM_INIT(round_statistics, /datum/round_statistics, new)
 	var/larva_from_marine_spawning = 0
 	var/larva_from_silo = 0
 	var/larva_from_cocoon = 0
+	var/larva_from_embryo = 0
 	var/larva_from_psydrain = 0
 	var/larva_from_siloing_body = 0
 	var/req_items_produced = list()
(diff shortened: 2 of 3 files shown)


Ben10083:
Willing to take feedback over raising or lowering psy point per late stage.

KerbalCitizen:
I just keep on fighting when infected until late in infection, when I just get medevac'd. It can actually be quite useful to fight when infected, since larval huggers won't do anything to you. In one round, I set off at least 3 larval hugger traps that were covered by items while infected, which means I prevented 3 non infected marines from getting infected. 

I had some concerns that marines will just carry more anti hugger stuff, plasma pistol being most popular example, or keeping infected marines in stasis bags. After talking ben10083 on discord though, cocoon gives 1.5 larva points and psychic drain gives 1 larva point, so 0.6 from infection is really not that bad in comparison. 

I'm still unsure how useful this will be to xenos since its a pretty small amount of larva points and whether marines will care about infections more since it doesn't affect them personally that much. It is a good change imo to have a consequence to waiting later in infection, even if it's a bit on the small side.

zzurge:
incentive to get that hugger out asap,works for me. and further increases the value of swat masks

Losenis:
So you're telling me I'll actually be mechanically justified in indulging in that fit of rage after the guy needlessly moves 3 times in the middle of taking his hugger off to PB him with a t35?

Based, speedmerge

EvilMelonTheSecond:
> So you're telling me I'll actually be mechanically justified in indulging in that fit of rage after the guy needlessly moves 3 times in the middle of taking his hugger off to PB him with a t35?

This might start to become a problem if huggers are buffed much further, marines will start to actively damage their team which might push people to execute marines they deem incompetent.

Ben10083:
> This might start to become a problem if huggers are buffed much further, marines will start to actively damage their team when hugged which might push people to execute marines they deem incompetent.

Pretty sure this is against OOC rules, and would be unlikely to happen even if it doesn't. People will just push that marine to deal with it sooner.

TiviPlus:
I already explained to you in discord why infections giving larva points isn't a good idea

Losenis:
> Pretty sure this is against OOC rules

A mechanic that encourages you to break the rules, imagine

PR #11536, opened 2022-11-12
Emergency chem tweaks

QualityVan:
## About The Pull Request
Removes the organ healing from emergency chems (rr, neura, inaprov)
Reduces inaprov pill from 30u to 15u

## Why It's Good For The Game
The organ damage changes meant that e.g. having broken lungs is no longer hugely crippling, and the raw healing provided by the three is strong as is. Organ damage is supposed to be hard to remove, too.
Inaprov has fairly limited benefit after the initial application's heal, just minor pain relief and denying certain forms of oxloss. Reduces the dosage a: keeps it consistent with most other mainstay pills and b: lets it leave the body earlier to open chem slots back up.

## Changelog

:cl:
balance: Inaprov pills changed to 15u
balance: Inaprov, RR, and neuraline don't unbreak organs
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/pill.dm
+++ b/code/game/objects/items/reagent_containers/pill.dm
@@ -165,7 +165,7 @@
 
 /obj/item/reagent_containers/pill/inaprovaline
 	pill_desc = "An inaprovaline pill. Used to stabilize patients."
-	list_reagents = list(/datum/reagent/medicine/inaprovaline = 30)
+	list_reagents = list(/datum/reagent/medicine/inaprovaline = 15)
 	pill_id = 10
 
 /obj/item/reagent_containers/pill/dexalin
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -26,11 +26,6 @@
 		L.adjustBruteLoss(-L.getBruteLoss(TRUE) * 0.30)
 		L.adjustFireLoss(-L.getFireLoss(TRUE) * 0.30)
 		L.jitter(5)
-		for(var/datum/internal_organ/I AS in H.internal_organs)
-			if(I.damage)
-				if(I.damage < 29)
-					return
-				I.heal_organ_damage((I.damage-29) *effect_str)
 		TIMER_COOLDOWN_START(L, name, 300 SECONDS)
 
 /datum/reagent/medicine/inaprovaline/on_mob_delete(mob/living/L, metabolism)
@@ -473,11 +468,6 @@
 		L.adjustBruteLoss(-L.getBruteLoss(TRUE) * 0.20)
 		L.adjustFireLoss(-L.getFireLoss(TRUE) * 0.20)
 		L.jitter(10)
-		for(var/datum/internal_organ/I AS in H.internal_organs)
-			if(I.damage)
-				if(I.damage < 29)
-					return
-				I.heal_organ_damage((I.damage-29) *effect_str)
 		TIMER_COOLDOWN_START(L, name, 300 SECONDS)
 
 /datum/reagent/medicine/neuraline/on_mob_life(mob/living/L)
@@ -566,11 +556,6 @@
 		L.adjustBruteLoss(-L.getBruteLoss(TRUE) * 0.20)
 		L.adjustFireLoss(-L.getFireLoss(TRUE) * 0.20)
 		L.jitter(10)
-		for(var/datum/internal_organ/I AS in H.internal_organs)
-			if(I.damage)
-				if(I.damage < 29)
-					return
-				I.heal_organ_damage((I.damage-29) *effect_str)
 		TIMER_COOLDOWN_START(L, name, 300 SECONDS)
 
 /datum/reagent/medicine/russian_red/on_mob_life(mob/living/L, metabolism)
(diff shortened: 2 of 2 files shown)


Losenis:
AFAIK it only capped the organ damage to a set amount, not healing it over it. It's still extremely crippling but it still had the niche of covering for freak scenarios of getting over 30 organ damage, which honestly I've only seen when inexperienced people do field surgery and don't remove the necro. By the time the person has over 30 organ damage either he's already out to get it treated due to the more than obvious symptoms, or he's walking around without a chest splint in which case the sudden revert to 29 organ damage isn't going to do a whole lot as he accrues more and more afterwards with a 300 second cooldown on the chem. In fact, in most cases where the guy gets over 30 organ damage, it's very likely he'll keep getting more organ damage after the heal.

I'd argue that the organ damage heal should stay purely to cover freak cases of things one can't really foresee. Warrior 1clicking 10+ lung damage is just one of possibly many, along with getting doped by a bald MO that gives you necro so you're desperatedly trying to survive just long enough to die on the alamo on the way shipside.

Strayed-Raven:
So is this just getting rid of the healing from these chems when you are crit and injected? Because if so lmao. If these had healing outside of the crit condition. also lmao. More med changes for the sake of it I guess.

Lumipharon:
Is the removing the organ heal that neccesary? It only reduces it down to a cap, so it's only helping people with very severe organ damage.

Having 30 organ damage on what's likely multiple organs is still pretty bad, but if some unga managed to give themselves 200 heart damage from being a smoothbrain, they're not entirely fucked.

QualityVan:
Surgery is a full heal anyway, and if you have 200 organ damage you should probably have to do something more about it than take a basic supplied hypo. Plus, it's not like going past broken causes further issues on organs aside from the pain.

If you do want an easy, readily available way to unbreak organs (which I'm against, of course), this isn't even a particularly good implementation of that. Tying it to crit health is awkward, and then it's locked to the same cooldown regardless of whether it does anything with no particular indicator either way.

Losenis:
Guys that 200 organ damage was 100% deliberate and was showing that in normal circumstances you're fucked or out of the fight before that point. You're combat ineffective before even hitting the heal cap of 29

Al-1ce:
>RR
Partly defeats the purpose.
>Inaprov
Intentionally designed to stabilize organs. Again, defeating the purpose of the chem.
>Neuraline
Neuraline costs 14 points at the corpsman vendor for a two-use injector. That's equal to about 14 QC/Peri+ syringes. Touching Neuraline and it's organ healing would be a massive nerf to a chem which is already overshadowed by Inaprov in sheer practicality.

This overall just rubs me wrong. All three of these require a rework, but this is absolutely not the way to go about it.

Al-1ce:
>RR

There's an arguement to be made here about how this makes RR even less useful in the field but whatever I guess.

>Inaprov
Intentionally designed to stabilize organs. Massively defeats the purpose of the chem.

>Neuraline
Neuraline costs 14 points at the corpsman vendor for a two-use injector. That's equal to about 14 QC/Peri+ syringes. Touching Neuraline and it's organ healing would be a massive nerf to a chem which is already overshadowed by Inaprov in sheer practicality alone.

This PR overall just rubs me wrong. All three of these require a rework, but this is absolutely not the way to go about it.

Pariah919 (commented):
Not sure about this one, don't think neuraline needs a nerf. Inaprov might need a rework and RR is fine honestly.

QualityVan:
Neuraline is and will continue to be the strongest healing chem in the game. Ina's good at what it is (basic emergency stabilization chem) and doesn't need more than that, although if you wanted to really be generous you could move the emergency activation from on inject to during metabolism so you could leave it in a mob brainlessly.

RR is the only one I could see as sensible, but a: we already have a chem that does organ healing for cloneloss, and b: the organ heal on it is restricted by both crit application (I'd say RR is mostly applied to low health but still upright mobs these days) and dosage (which you want to restrict as much as possible with RR due to cloneloss). I don't think it should touch organs at all aside from possibly harming the liver, but if you really want it to this is a terrible way to do it.

MathiasOrtega:
Barely anyone is using Innap for this. Nobody is going to bother sharing RR either. And only Corps will have neura anyway. This just seems like it’s being made to match the twin PR about organ damage to fit a scheme or theme. Since this makes the organ damage PR and useless. Corps don’t carry Innap injector pretty sure. But maybe hypospray?

KerbalCitizen:
When I play corpsman, inap pill bottle with the 30u pills and hypospray is a godsend. Refill hypospray with the 30u inap pills, each pill getting you 2 15u injects. 16 pills x 2 injects per pill + 4 injects loaded in hypospray, 36 funny instant crit heals. So much better to instantly inject inap into a crit marine than for them to die, which forces you to strip armor and defib, then they deaf and need to put their armor on again. Guess I'll just carry two inap pill bottles then...

Strayed-Raven:
Thinking about it. you could just balance emergency chems to instantly use up their units when used as an emergency chem. Add brain damage to innap used this way and cloneloss instantly applies to RR. So if you want more RR or innap in the marine after the effect you must use more than the emergency threshold

Losenis:
Changing things for the sake of changing things

PR #9786, opened 2022-03-14
CAS prediction, fatty rework, lase changes, minor CAS smoke/flare buff

breadhunt:
## About The Pull Request
A few changes to do with CAS. The aim of this PR is to make CAS less bullshit for xenos death-wise, and in particular, more supportive to marines. Because it's cool as fuck to be able to work together with the cool as fuck PO and their cool as fuck jet:

---

### CAS prediction:
-CAS now displays where it will hit, _for marines._ This is a blue display. If you're in it, you're about to be exploded or set on fire or suffer other harm. 
-If you're not in it, you're safe. Keep in mind this may not be 100% accurate if, say, an airlock closes while the ammo's still traveling.

### Lase changes:
-All ammo types (except HV GAU) have their own unique CAS lase (green blinking dot). _Both xenos and marines can see this._
This means, xenos can know what's being fired at them (and direction for GAU/laser). They don't know the AoE or time until it hits.

### CAS flare buff:
Flare actually hits where you want it to. Previously it would hit anywhere within a 4 tile radius.
Flare doesn't murder marines standing near it.
Effect of the actual flare itself is unchanged. But hopefully this means flare isn't a % chance for you to fuck over your own team anymore, for using a support weapon.
Also removed explosion from smoke minirockets (not tfoot), because you can't even use them for bluffing anymore due to lase changes

### Fatty rework:
Okay, so this might be the change everybody screams at me for the most.
Instead of its old function (initial small explosion, 8 delayed explosions nearby), it now deletes resin walls nearby. _This does not harm marines._
I'll explain why this is good further down

### CAS prediction:
This allows CAS to support marines easier, and lets marines know before they are about to be absolutely obliterated by their own teammate; encouraging teamwork on both sides.

### Lase changes:
I know this removes a lot of bluffing potential for CAS, and as such a _really fucking big part of CAS_ is just gone. It's still a good change. This isn't fun at all for xenos, and can completely ruin the game for them. It needs to be removed, and to compensate there's other stuff to help marines on the ground and ways to shoot xenos more often. Bullshit deaths just aren't fun.

### CAS flare buff:
-Flare and smoke are already regarded as garbage
-You can no longer make xenos think they will be instakilled by them.
-They are now completely useless
-By making them more reliable and usable within large clusters of marines, they should actually be fucking viable for the first time in a thousand years

### Fatty changes:
[...]

PR #11546, opened 2022-11-13
Fatty Rework 2.0 (Cluster strikes back!)

KacperRaid:
## About The Pull Request

Changes fatty back to its original state from before #9786

Adds prediction to the first Explosion for Fatty

## Why It's Good For The Game
_Approved from original author!_
Fatty back

### But why?: 

- HE OB is just better version of you, with more range and damage to every xeno structure and xenos for less hussle.
- PC does your job faster, and its less expensive in long run, it can also remove weeds, other types of walls and hurt xenos.
- Flamers are practicly dominating wall removal game, by the time you manage to get a shot out, they will remove that maze twice.
- Marines with boot knife can effectly remove walls _if they suck it up and actually do it_, and they cost less then your rockets!

**Overall fatty is in pretty bad place right now.**
so to change that, i'm making it back to cluster explosive!

## Changelog
:cl:
add: Adds prediction to first explosion for Fatty
balance: Instead of walls, Fatty now remove lives (also brings back cluster explosion)
balance: Changes price back to 250 points
/:cl:


Araulius:
bring back the x fuel napalm bomb while you're at it

TiviPlus (changes requested):

Starfirejordan:
marine keep getting FF by me because they see the area of effect warning for fatty and think its small thus walks into the area of effect for the split of 8 explosion's, it quite funny watching a marine think there safe near it only to be hit by 2-3 explosion's and instantly crit or die

KacperRaid:
> marine keep getting FF by me because they see the area of effect warning for fatty and think its small thus walks into the area of effect for the split of 8 explosion's, it quite funny watching a marine think there safe near it only to be hit by 2-3 explosion's and instantly crit or die

i've actually intented to make prediction for that cluster effect as well, but scarpped it due to preformence issues it caused.

**but on topic of FF, that is actually by original design!**

"Fatties" were supposed to give false sense of security after first explosion (tricking xenos or _**ungas who don't pay attention to comms**_ into thinking its minirocket) into pushing right after it gone off, just to be blasted by 8 other explosions few seconds later. 

**However**
back then we only had green dot for every CAS weapon, making it really unpredictable what CAS is going to drop on everyone heads. 
**But** now we have custom dots for every single one of them, so I think its just gonna be matter of time before everyone starts recognizing 4 deep blue dots with cluster death from above. _and get hint to start running away from that "safe zone"._

TiviPlus:
>but scarpped it due to preformence issues it caused.

This sounds like you were doing something very wrong

KacperRaid:
> This sounds like you were doing something very wrong

Yea, i was thinking same thing, so i asked original author about it.
and he had same problems.

> Also code-wise calculating prediction for 9 different explosions would be painful so we're not doing that. Fatties already freeze the game for a couple seconds on their own

from #9786

Lumipharon (approved):

PR #11572, opened 2022-11-19
Runtime fix when linking telecomms machinery

GravitronBlue:
Copypaste fix from TG that should fix linking telecomms machinery resulting in UI errors and runtimes.
Tested locally and didn't have issues with it, no clue why it works though, perhaps something about removing laziness.

Should have someone who can understand it take a look and ensure this doesn't harm anything though.

Credit to KacperRaid for digging up this code from TG.
### Why It's Good For The Game
Because telecomms shouldn't become a game of "How long does it take multiple admins and Pit Boss to fix" or "How one AI stole the airwaves."

### Changelog

:cl:
Linking telecomms machinery together should now function properly
/:cl:

PR #11574, opened 2022-11-19
Adds extra grief protection to tcomms

RipGrayson:
## About The Pull Request

Title.
Adds extra logging to tcomms, a number of options like removing frequencies or changing networks are locked to admins only.

## Why It's Good For The Game

There's a long running griefer who's ban evading and griefing telecomms, while https://github.com/tgstation/TerraGov-Marine-Corps/pull/11572 will fix one element of it it won't be enough. There's a myriad of other ways to break comms that I can think of as an ex-TG engineer main. 

The griefer will just swap to another method and the pain will continue, this PR makes it significantly harder by blacklisting a number of options that have no use (when was the last time a regular player had to adjust frequencies or networks??) to a normal player but could be used to grief.

## Changelog
:cl:
add: Adjusted tcomms, added more logging.
/:cl:


Arctrooper8:
Is there any reason at all to have the equipment itself actually do anything? You might as well just make it all decorative except for APC room power to still allow for Xeno sabotage.

RipGrayson:
> Is there any reason at all to have the equipment itself actually do anything? You might as well just make it all decorative except for APC room power to still allow for Xeno sabotage.

We run TG comms, which depends on the equipment to function, the machines themselves cannot become props.
A different question would be "Is there any reason that people should be allowed to configure telecomms?" and the answer to that would be probably no.

The only reason I haven't removed all telecomm config settings is in case an admin needs to adjust telecomms for an event.

Arctrooper8:
>We run TG comms, which depends on the equipment to function, the machines themselves cannot become props.
A different question would be "Is there any reason that people should be allowed to configure telecomms?" and the answer to that would be probably no.

I more meant why not just move the functionality of the comms itself off the comms machines, and only leaving that cutting APC power would turn off comms (for xeno sabotage), allowing the machines to be made props. However thinking about it more that'd probably be a lot of work when just adding grief protection should make them foolproof.

BraveMole (changes requested):
Why are you even keeping these features. Admin do not need and shouldnt touch that

BraveMole:
> The only reason I haven't removed all telecomm config settings is in case an admin needs to adjust telecomms for something (like an event).

No admin ever touched that or ever will

PR #11581, opened 2022-11-20
Adds a way for radios to bypass tcomms

Autisem:
## About The Pull Request

adds a var for radios to bypass tcomms

## Why It's Good For The Game

either we make tcomms uninteractable or we add a way for admins to choose to void tcomms. because grief is starting to get old

## Changelog
:cl:
code: There is a new bool for radios to bypass tcomms system
admin: you can use SDQL to update all radios to have this var enabled, letting them bypass tcomms
/:cl:


BraveMole:
We are not adding hacky solution to the codebase

PR #11591, opened 2022-11-21
Xeno Drag Changes

Naaanii:
## About The Pull Request
<s>Increases windup timer for dragging items.
Reduces the speed malus while dragging items so pre thought out setups are slightly easier to make.</s>
Taking damage now stops the windup for dragging items and stops the drag itself.

Dragging Xenos or mobs SHOULD be unaffected as far as I've tested.
## Why It's Good For The Game
Combat dragging is a bit of an issue. Xenos can acid mid combat to try and prevent nuanced items (like PC) but will have to work harder to drag it mid combat or combat drag flares or weapons with no retaliation and helps make a less forced mag harness meta.
It also helps Rouny keep it's primo niche of combat dragging.
## Changelog
:cl:
add: Xenos now stop drag channel when taking damage.
add: Xenos now stop dragging objects when taking damage.
/:cl:


srxtreme:
excellent PR.
يدخل Sentinel في وضع القاذف ، ويكتسب بالإضافة إلى خمسينTier 1 caste بالمائة سرعة نحو خزانات وقود الwelding fuel tankلحام. يمكن للحارس سحب خزانات الوقود. دمرت FOB الغرب. + 20٪ ضرر مقابل الحواجز ، انفجار خفيف 4 fire tiles + light aoe النارdamage vs barricades. سلبي ، حارس يمكن أن يموت أو يمكن أن يقتل لحام خزان الوقود

maikilangiolo:
Faster but interruptible drag seems fine. If marines fluoride stare the xenos dragging the item, it's on them

Losenis:
I'd have gone with decreased windup but taking damage interrupts it at any stage of dragging tbh

Starfirejordan:
decreased time but having it intrudable would be better tbh

Naaanii:
> I'd have gone with decreased windup but taking damage interrupts it at any stage of dragging tbh

Its now the same but damage interrupts dragging.

srxtreme:
might as well remove dragging completely and instead reduce acid time...

Goblinu:
> might as well remove dragging completely and instead reduce acid time...

Yeah, just buff acid. I still find it appaling that it takes a whole 8 mins for weak acid to burn through a flare

Losenis:
Bluenames were stubborn on keeping dragging in despite bringing nothing but pain and unintended behaviors

DemonOfHatred:
Having it be any damage seems unreasonable, if it had a damage cap of say 100 i could see it im going to give it a thumbs down.

Pariah919 (approved):
Makes sense to me.

Oscari-pog:
This is dumb. Whats the point of trying to steal items during combat now?

TiviPlus:
I can get the taking damage while windup but why the rest? also does nothing about the "flares dragging"issue you mentioned

Skatuk:
Now make the same for marines

Beepinex:
> Now make the same for marines
arm fractures?

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -86,6 +86,10 @@
 	if(!damage) //If we've actually taken damage, check whether we alert the hive
 		return
 
+	if(isobj(pulling))
+		stop_pulling()
+
```
psykzz, line 91:
```suggestion
	// additional chance when pulling to stop pulling the mob.
	if(isobj(pulling) && prob(5))
		stop_pulling()
		
```

Personally i would add this a chance.
With the amount of bullets being hit at you from a marine firestorm its likely going to be triggered anyway, but a single marine isn't going to stop anything.

Specific value can be tweaked.

Naaanii: Is this a 5% or 50% chance to drop?

The idea is that any source of damage stops dragging (FIRE, bullets, laser guns etc.)

psykzz: This would be 5%.

Persoanally don't agree that every instance of damage should cause you to stop dragging, but some threshold (or in this case, random chance).

psykzz: Note that this isn't really a flat 5% since there are a number of things that apply multiple instances of damage (shotguns, fire etc).

Naaanii: While multiple instances of damage could make it cumbersome to use, I think the percentage should be slightly higher as there are already castes that can quite easily transport items that can be dragged i.e wraith. This just gives a bit of counterplay from the marine's side outside of needing to have a stun weapon. Would 30 or 25% be better?

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
```diff
@@ -73,6 +73,13 @@
 
 	regen_power = -xeno_caste.regen_delay //Remember, this is in deciseconds.
 
+	if(!damage) //If we've actually taken damage, check whether we alert the hive
+		return
```
Lumipharon, line 77: This check already exists, row 54.

Lumipharon (changes requested):

code/modules/mob/living/carbon/xenomorph/damage_procs.dm
Naaanii: Removed dupe

Lumipharon (approved):

PR #11613, opened 2022-11-23
T4 primo cost reduction

Naaanii:
## About The Pull Request
Haves the cost for T4 Primo.
## Why It's Good For The Game
T4 primo is almost always the last bought primo in all games and only buffs 2 castes compared to the others buffing double, triple or even more.
Reducing the cost helps seeing it bought more often and not having wait much longer for it or even help in lowish pop scenarios.
## Changelog
:cl:
balance: Reduced T4 primo cost from 600 to 300.
/:cl:


EvilMelonTheSecond:
Psypoints scale with pop, no need to try to adjust costs for it.

The problem with T4 primo seems to be its function failing to aid the whole team as opposed to the price being too high to buy. I'd say changing the abilities so that the team as a whole would prefer it to be bought instead of primos with wider coverage would be the proper solution. This change is just going to encourage hiveleaders to sneak in the upgrade for themselves before getting everybody else's.

Also I hate queen charge, having to chase the highest hp xeno that shakes the screen and run off into the sunset after taking 3 bullets unless you've pushed them into a corner in the caves is the worst experience for me in the game so far.

Naaanii:
> Psypoints scale with pop, no need to try to adjust costs for it.
> 
That doesn't change that you're spending 600 to only enhance 2 Xenos. Psy drain and cocoon scaling only affects how soon or later you can attain these upgrades.
>The problem with T4 primo seems to be its function failing to aid the whole team as opposed to the price being too high to buy. I'd say changing the abilities so that the team as a whole would prefer it to be bought instead of primos with wider coverage would be the proper solution. This change is just going to encourage hiveleaders to sneak in the upgrade for themselves before getting everybody else's.
>
Doubt this will be the case as in 99% of all rounds, it is the last upgrade purchased. That could also be hardcoded to make it so it's the last upgrade allowed to be purchased but having a spare amount of or the upgrade costing 300 points would be a neat insurance for hijack or other such scenarios.

As it stands, it has the least ROI (barring King but its getting reworked) and this should keep it on par with other upgrades.

EvilMelonTheSecond:
> > Psypoints scale with pop, no need to try to adjust costs for it.
> 
> That doesn't change that you're spending 600 to only enhance 2 Xenos. Psy drain and cocoon scaling only affects how soon or later you can attain these upgrades.

My point was, despite effecting only 2 castes those two are almost 100% present in the round. Their effects should be related to their leadership role and not just a skill that buffs them as an individual. Keeping current primo skills simply makes them scale backwards with xeno pop no matter how cheap they are. 

This change MIGHT be helpful but the problem will remain that picking T4 primo before others will still be the selfish option since leader castes, who gotta keep themselves safe to better lead the hive, don't get any help to leading the hive with it, just a combat buff for themselves.

PR #11615, opened 2022-11-23
Railing smoothing removal

Lumipharon:
## About The Pull Request
In some req areas, we get ugly smoothing like this:
image

Removes this.
## Why It's Good For The Game
Uggo smoothing be gone.
## Changelog
:cl:
fix: Req railing will not longer smooth with walls nearby
/:cl:

### DIFF
--- a/code/game/objects/machinery/doors/railing.dm
+++ b/code/game/objects/machinery/doors/railing.dm
@@ -7,6 +7,8 @@
 	opacity = FALSE
 	open_layer = CATWALK_LAYER
 	closed_layer = WINDOW_LAYER
+	smoothing_behavior = null
+	smoothing_groups = null
 
 	var/obj/docking_port/mobile/supply/linked_pad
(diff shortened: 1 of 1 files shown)


code/game/objects/machinery/doors/railing.dm
```diff
@@ -7,6 +7,8 @@
 	opacity = FALSE
 	open_layer = CATWALK_LAYER
 	closed_layer = WINDOW_LAYER
+	smoothing_behavior = null
+	smoothing_groups = null
```
RipGrayson, line 12: Shouldn't this be 

smoothing_behavior = NO_SMOOTHING
smoothing_groups = NONE

instead of null

Lumipharon: yep, should use the define.

TiviPlus (approved):

PR #11631, opened 2022-11-27
Combat canteen has more possible transfer amounts

novaepee:
## About The Pull Request

Combat canteen has more possible transfer amounts from inheriting /obj/item/reagent_containers/food/drinks .

## Why It's Good For The Game

Sometimes, marines want to have more measured transfer amount when drinking out of their canteen. possible_transfer_amounts is the same as /obj/item/reagent_containers/glass, which is what beaker inherits from. I would not be surprised if marines make some cocktail of a canteen with this PR.

## Changelog

:cl:
add: Combat canteen has more possible transfer amounts
balance: More options on transfer amounts for combat canteen will inevitably change how marine consume from combat canteen
/:cl:


Losenis:
Microdosing water meta lets GOOOO

Lumipharon:
This sounds like it's purely for chem shenanigans

PR #11638, opened 2022-11-28
Vali bugfix

Zar000:
## About The Pull Request
Bugfix good

## Why It's Good For The Game
instant necro not fun :(
## Changelog
:cl: Credit TiviPlus & Slotbot
fix: fixed vali instant necro bug
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -258,24 +258,24 @@
 		COOLDOWN_START(src, chemboost_activation_cooldown, 10 SECONDS)
 		setup_bonus_effects()
 		return
+	else
+		if(!COOLDOWN_CHECK(src, chemboost_activation_cooldown))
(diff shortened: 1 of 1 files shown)


Lumipharon:
Just to be clear here, you're changing the behavior of vali from how it was originally. The checks and early returns were below the 'turn vali off' code - i.e. you could turn it off without going through those checks, which is now no longer possible.

I can't say if that's good or not, but it is changing.

code/datums/components/chem_booster.dm
```diff
@@ -277,6 +275,7 @@
 		meds_beaker.reagents.trans_to(wearer, 30)
 	setup_bonus_effects()
 
+
```
Lumipharon, line 278:
```suggestion
```

Lumipharon (approved):

PR #11642, opened 2022-11-28
balance: increase cooldown on mech computer

psykzz:
## About The Pull Request

Increase cooldown on crafting mechs to 1 minute each.
Additionally updates the display to show when its busy.

## Why It's Good For The Game

There is currently no limit to how many you can spam.
No external resources used or anything.

This will slow the spam of mechs, until a proper solution can be implemented.

## Changelog

:cl:
balance: mechs now take a minute to build
/:cl:

### DIFF
--- a/code/modules/vehicles/mecha/combat/greyscale/greyscale_constructor.dm
+++ b/code/modules/vehicles/mecha/combat/greyscale/greyscale_constructor.dm
@@ -181,8 +181,6 @@ GLOBAL_LIST_INIT(greyscale_weapons_data, generate_greyscale_weapons_data())
 		return FALSE
 
 /obj/machinery/computer/mech_builder/ui_interact(mob/user, datum/tgui/ui)
-	if(currently_assembling)
-		return
 	ui = SStgui.try_update_ui(user, src, ui)
 	if(ui)
 		return
@@ -219,6 +217,7 @@ GLOBAL_LIST_INIT(greyscale_weapons_data, generate_greyscale_weapons_data())
 	data["selected_name"] = selected_name
 	data["selected_equipment"] = selected_equipment
 	data["current_stats"] = current_stats
+	data["currently_assembling"] = currently_assembling
 	return data
 
 /obj/machinery/computer/mech_builder/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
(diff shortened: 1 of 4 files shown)


Al-1ce:
> 1 minute each.

This is still very easily spammable.

blackdav123:
it takes like two or so minutes on most maps to drive the mech onto the alamo then run back to print another mech this wont change anything

blackdav123:
just found out about conveyor belts, I think I'm gonna throw up.

Serxule:
> > 1 minute each.
> 
> This is still very easily spammable.
> 
> EDIT: I'd say 15 minutes at a bare minimum.

i'd say 5-10, 15 is a bit ridiculous

code/modules/vehicles/mecha/combat/greyscale/greyscale_constructor.dm
```diff
@@ -292,10 +291,9 @@ GLOBAL_LIST_INIT(greyscale_weapons_data, generate_greyscale_weapons_data())
 			if(isnull(selected_visor))
 				return FALSE
 			currently_assembling = TRUE
-			addtimer(CALLBACK(src, .proc/deploy_mech), 1 SECONDS)
+			addtimer(CALLBACK(src, .proc/deploy_mech), 60 SECONDS)
 			playsound(get_step(src, dir), 'sound/machines/elevator_move.ogg', 50, 0)
```
TiviPlus, line 295: This was intended as a visual delay to sync it up with the sound effects when it deployed
Maybe make it use TIMER_COOLDOWN instead so you just cant make successive ones instead of needing to wait

TiviPlus (changes requested):

Al-1ce:
> i'd say 5-10, 15 is a bit ridiculous 


Spamming up to fifty mechs on a fucking assembly line for the net cost of **_literally_** nothing is broken. There is no arguement to be had there.

And **IF**, by some ungodly miracle, the MP is stupid enough to have his nigh-invulnerable machine destroyed, it would serve right that they have to wait at least a couple minutes until they can produce another.

I find it very, very hard to imagine MPs getting their mechs destroyed in a span of _15 minutes or less._ That's ridiculous, and nothing more than a skill issue. Mechs are not intended to be mass manufactured. The slots are limited to two (in most instances) for exactly that reason.

There should never be more mechs on the map than there are pilots, period. You are defending an absolute cheesewheel of an oversight.

ThelastArchiver:
i think a 20 minute cooldown or maybe  a bit longer would be good itn alows the mp to have atleast one reserve mech before heading down while not making it super spamable

MereWalker:
Just put a limit of three max spawned mechs for a pilot. Simple fix imo. Easy code? Not a clue.

Loldaddy5676:
how about make the mech cost resources, req can buy supplies to make one mech for a high cost. this forces the marines to not just use the mech as a damage sponge.

psykzz:
> ow about make the mech cost resources, req can buy supplies to make one mech for a high cost. this forces the marines to not just use the mech as a damage sponge.

Long term this is really the approach i think we should take, but its not what i want to tackle in this PR.

zzurge:
If you guys are going to make mech take up to fifteen minutes to be replaced then you better make the arms re attachable because thats just fucking absurd

TiviPlus:
Rest looks good

TiviPlus (approved):

PR #11643, opened 2022-11-28
Mech minigun buff

Kuro020:
## About The Pull Request

Buff mech minigun damage, nerfs AP

## Why It's Good For The Game

I might have made the minigun just straight up worse than the LMG in the last nerf pr, oops

## Changelog

:cl:
balance: Mech minigun damage buffed to 30, AP nerfed to 10
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1256,8 +1256,8 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 
 /datum/ammo/bullet/minigun/mech
 	name = "vulcan bullet"
-	damage = 20
-	penetration = 20
+	damage = 30
+	penetration = 10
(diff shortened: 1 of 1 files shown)


maikilangiolo:
Why nerf the AP a second time?

Kuro020:
> Why nerf the AP a second time?

It wasnt nerfed on the first pr, this is basically reverting the damage nerf and nerfing ap instead to differentiate the minigun from the lmg a bit more

maikilangiolo:
Makes sense. Lmg still triumph with same rof, more ammo, accuracy, less slowdown...
I think giving the minigun a higher sunder would be good. It becomes a harder hitting gun where you can chip away, like I assume it's supposed to be.

Serxule:
> I think giving the minigun a higher sunder would be good. It becomes a harder hitting gun where you can chip away, like I assume it's supposed to be.

with something like this, i dont think you could do noticeably higher sunder without it being OP

PR #11647, opened 2022-11-29
Fixes tier 2 xeno calculation

Lumipharon:
## About The Pull Request
Tier 2 slots will now be calculated correctly.

It was adding the existing t2 xenos to the calculation, so you'd never run out.
## Why It's Good For The Game
Bug fix good.
## Changelog
:cl:
fix: T2 xeno slots being calculated incorrectly
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
@@ -1060,7 +1060,7 @@ to_chat will check for valid clients itself already so no need to double check f
 	var/fours = length(xenos_by_tier[XENO_TIER_FOUR])
 
 	tier3_xeno_limit = max(threes, FLOOR(((zeros + ones + twos + fours) * (evotowers.len * 0.2 + 1)) / 3 + 1, 1))
-	tier2_xeno_limit = max((twos + zeros + ones + fours) * (evotowers.len * 0.2 + 1) + 1 - threes)
+	tier2_xeno_limit = max(twos, (zeros + ones + fours) * (evotowers.len * 0.2 + 1) + 1 - threes)
 
 // ***************************************
 // *********** Corrupted Xenos


FIXTHERADBOXPLEASE:
An unnecessary xeno nerf in a gamestate that already heavily favors marines.

FIXTHERADBOXPLEASE:
Also, the balance tag should be applied to this.

Wayland-Smithy:
Bad move. Turns out more slots is more fun and should just be called a feature. Winrates worshippers where are you now?

Goblinu:
image

ScheveningenZonn:
It's not a balance change, it's literally not working as intended.

Goblinu:
> It's not a balance change, it's literally not working as intended.

Shotguns and stickies also "works as intended"

Losenis:
Everyone complaining about it doesn't realize that the moment you put one single evolution tower down you suddenly have more T2 slots than you have actual xenos of all tiers around

Wayland-Smithy:
@Losenis That is not how the math would appear to work. Each tower grants +20% more slots so x1.2 for 1, x1.4 for 2, and so on.. curb your exaggerations.

Losenis:
We had more T2 slots than we had xeno players.

Wayland-Smithy:
At this moment yes, but that is because of the bugged math.

Losenis:
Isn't that what I said originally?

Wayland-Smithy:
Yes, so are you not implying that after this fix all xenos need to do is slap down a couple towers to have the same effect?

Since you were being aggro toward the xeno biased comments it seemed that way. Sorry if I misread.

Losenis:
I'm saying that with the bugged math, putting down an evo tower gives xenos more T2 slots than they have players, effectively intensifying the bugged math to the point it makes the cap irrelevant

Wayland-Smithy:
That is true the tower boost is outsized from the bugged math, but even without any built I have seen the t2 slots outnumber alive xenos, so they're not really a factor here imho which is why I find mentioning them odd unless to point to for coping.

❤️

PR #11651, opened 2022-11-29
Marine Vendor QOL

collinlunn:
## About The Pull Request
Makes the Automated Closet autoequip certain items (other vendors unchanged). Clothing, pouches, belts, and such autoequip. For all other items, or if equipping fails, the items attempt to insert into the player's hands.

https://user-images.githubusercontent.com/60152240/204478318-abf4c63b-aec2-42c6-8cd7-55b5e6e9b11d.mp4

## Why It's Good For The Game
Allows the player to do the same thing in less clicks and menus. Most items vended from the closet have a clear place they should be automatically equipped. By automatically equipping them, it saves the player from needing to click a bunch.

## Changelog
:cl:
qol: The Automated Closet now auto-equips certain items when vended.
/:cl:

## Code Stuff
1. obj/item has on_vend method that different item types override for custom auto-equipping functionality
2. Previously, /obj/effect/essentials_set and /obj/effect/modular_set were used to spawn multiple items from a single path, with the latter running an effect on each spawned item. These were merged into /obj/effect/vendor_bundle, with the effects being run on each item moved to item/on_vend().


novaepee:
beautiful, thank you for your volunteering

Losenis:
image

MereWalker:
Now that is a good change. I like this.

Serxule:
latejoin before xeno hijack powercreep





(joke)

code/game/objects/items/marine_gear.dm
```diff
@@ -246,6 +246,10 @@
 	time_to_unequip = 1 SECONDS
 	flags_inventory = NOQUICKEQUIP
 
+/obj/item/belt_harness/on_vend(faction)
+	if (!usr.equip_to_appropriate_slot(src))
+		..()
```
TiviPlus, line 251: do not use usr
preserve parent return, use return ..() pls
applies for all other instances of this

code/game/objects/machinery/vending/new_marine_vendors.dm
```diff
@@ -571,102 +575,99 @@
 	resistance_flags = INDESTRUCTIBLE
 	lock_flags = JOB_LOCK
 
+/obj/effect/vendor_bundle
+	var/list/gear_to_spawn = list()
+	var/list/spawned_gear = list()
```
TiviPlus, line 580: autodoc is missing everywhere

TiviPlus: also use lazylists here please

TiviPlus (changes requested):
Please dont bundle changes, this is difficult to review

code/datums/components/attachment_handler.dm
```diff
@@ -32,254 +32,285 @@
...
 	RegisterSignal(parent, COMSIG_ITEM_UNEQUIPPED, .proc/remove_overlay)
 
-///Starts processing the attack, and whether or not the attachable can attack.
-/datum/component/attachment_handler/proc/start_handle_attachment(datum/source, obj/attacking, mob/attacker)
-	SIGNAL_HANDLER
-	INVOKE_ASYNC(src, .proc/handle_attachment, attacking, attacker)
+///Gets data from the /datum/element/attachment on the specified item. Returns null if target isn't really an attachment.
+/datum/component/attachment_handler/proc/get_attachment_data(obj/attachment)
+	var/list/attachment_data = list()
+	SEND_SIGNAL(attachment, COMSIG_ITEM_IS_ATTACHING, attachment_data) //attachment element adds data to list if present on item
```
BraveMole, line 53: Signal have a rule : they need to be sent at only one place in code.

code/game/objects/items.dm
```diff
@@ -1355,3 +1355,7 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 	user.swap_hand()
 	user.update_inv_l_hand(0)
 	user.update_inv_r_hand()
+
+/obj/item/proc/on_vend(faction)
+	if(usr.put_in_any_hand_if_possible(src, warning = FALSE))
```
BraveMole, line 1360: Never use usr, we do not allow it on this codebase for new code :)

code/modules/clothing/modular_armor/attachments.dm
```diff
@@ -253,3 +278,8 @@
 /obj/item/armor_module/armor/proc/extra_examine(datum/source, mob/user, list/examine_list)
 	SIGNAL_HANDLER
 	examine_list += "Right click the [parent] with paint to color the [src]"
+
+/obj/item/armor_module/armor/on_vend(faction)
+	..()
```
BraveMole, line 283: Always keep the return value of ..() in .

BraveMole (changes requested):

hyper2snyper (changes requested):
Ok so first rules of Components/Elements. You should never have direct calls to them. For all purposes their functions should be private to themselves.
Signals should only be sent from one location, and they should be considered events, not calls.

You've got a lot of places where you divide something into two functions, with one being an 'is' check. We dont need these. You should remove all of them and just check the thing you want to check directly.

I don't appreciate you piecemealing, rearranging and renaming all the code for the sake of 'readability' when that readability is only more readable to yourself. Keep the names the same please.

As for the logic, signals should be done of course. Have an event that is the vendor doing its thing, then have the component register for the event.

start_handle_attachment > handle_attachment > finish_handle_attachment
the first is the call from the signal, the next handles all checks, and the third is the thing that actually does the attach. You should keep the names the same. There also shouldnt be separate call stacks for mobs vs non-mobs. There should just be a mob check to determine whether the mob checks are done.

The same is true for detaches,
start_detach > finish_detach.


Basically, dont reorganize everything. Leave names the same. dont abuse signals. stop splitting checks into functions and just do the check. Dont abuse components.

collinlunn:
The attachment changes ended up being more difficult than I had anticipated, so I am cutting all features/changes relating to it.

code/datums/components/attachment_handler.dm
collinlunn: Resolved by me removing the attachment changes.

code/datums/components/attachment_handler.dm
```diff
@@ -32,254 +32,285 @@
...
+	return length(attachment_data) == 0 ? null : attachment_data //if the list is empty there wasn't an attachment element to add data to it
 
-/datum/component/attachment_handler/proc/handle_attachment(obj/attachment, mob/attacher, bypass_checks = FALSE)
+///Get the attachment item in a slot
+/datum/component/attachment_handler/proc/attachment_in_slot(slot)
```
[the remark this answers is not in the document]
collinlunn: Resolved by me removing the attachment changes.

code/game/objects/items/marine_gear.dm
collinlunn: Removed all uses of usr in on_vend, and modified all on_vend overrides to preserve parent return

code/game/objects/items.dm
collinlunn: Removed all uses of usr in on_vend. I am still using usr in the vendor code (new_marine_vendors.dm) because the ui_act already makes extensive use of it, and I don't think I can remove any of it without refactors outside the scope of this PR.

code/modules/clothing/modular_armor/attachments.dm
collinlunn: Preserved return values with . = ..() for all calls that did not do so.

code/game/objects/machinery/vending/new_marine_vendors.dm
collinlunn: Replaced with lazylists, and added autodoc to the new classes, methods, variables in PR

code/game/objects/machinery/vending/new_marine_vendors.dm
```diff
@@ -571,102 +575,103 @@
...
 	. = ..()
-	for(var/typepath in spawned_gear_list)
-		if(spawned_gear_list[typepath])
-			new typepath(loc, spawned_gear_list[typepath])
-		else
-			new typepath(loc)
-	qdel(src)
-
+	for(var/typepath in gear_to_spawn)
+		LAZYADD(spawned_gear, new typepath(loc))
```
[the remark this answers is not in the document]
collinlunn: LAZYINITLIST'd gear_to_spawn before accessing

code/game/objects/items.dm
```diff
@@ -1355,3 +1355,8 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 	user.swap_hand()
 	user.update_inv_l_hand(0)
 	user.update_inv_r_hand()
+
+///Used by vendors to attempt to automatically equip a vended item. Attempts to put the item in the user's hand.
+/obj/item/proc/on_vend(mob/user, faction)
+	if(user.put_in_any_hand_if_possible(src, warning = FALSE))
+		pickup(usr)
```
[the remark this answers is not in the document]
collinlunn: Removed usr call I missed on first pass

XSlayer300:
paaaain.

PR #11654, opened 2022-11-30
Remove t2 xeno cap

Wayland-Smithy:
## About The Pull Request

Formally removes the cap on t2 xenos as has been the case gameplay wise since September due to a bug, and with less rounys around having seemingly negligible impacts on balance.

## Why It's Good For The Game

More options more often for fun.

## Changelog
:cl:
del: Deleted the cap on t2 xenos.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/evolution.dm
@@ -143,7 +143,6 @@
 		return
 
 	// used below
-	var/no_room_tier_two = length(hive.xenos_by_tier[XENO_TIER_TWO]) >= hive.tier2_xeno_limit
 	var/no_room_tier_three = length(hive.xenos_by_tier[XENO_TIER_THREE]) >= hive.tier3_xeno_limit
 
 	if(new_caste_type == /mob/living/carbon/xenomorph/queen) //Special case for dealing with queenae
@@ -217,9 +216,6 @@
 
 		if(regression)
 			//Nothing, go on as normal.
-		else if(tier == XENO_TIER_ONE && no_room_tier_two)
-			to_chat(src, span_warning("The hive cannot support another Tier 2, wait for either more aliens to be born or someone to die."))
-			return
 		else if(tier == XENO_TIER_TWO && no_room_tier_three)
 			to_chat(src, span_warning("The hive cannot support another Tier 3, wait for either more aliens to be born or someone to die."))
 			return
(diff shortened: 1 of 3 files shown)


567Turtle:
What do you mean "as it has been for awhile now", not only does a t2 xeno cap continue to be in game right now, but there is another PR right now that makes it better

Wayland-Smithy:
> What do you mean "as it has been for awhile now"

Because of the code merged back in September the calculated limit is often above even the total amount of xenos making the cap in practice non existent.

567Turtle:
My previous comment was misinformed, forgive me

Naaanii:
Would this make T2s and T3s separate? Iirc, they're counted as the same.

Think it would be better to at least have a T3 cap and make evo towers be of some use (if not much at all).

Wayland-Smithy:
This has no impact on T3s as all lower castes are of equal value for their math.

BraveMole (commented):
What was your opinion on making evo time longer between t1 to t2 ? I like this but I am slightly afraid that it makes t1 globally unused, which is maybe untrue

Lumipharon:
I don't mind more T2 slots or whatever, but without a dramatic increase in T1 evo time, then T1 other than probably drone will essentially stop existing, since by the time you're mature/elder you're ready to evo to T2.

Wayland-Smithy:
> What was your opinion on making evo time longer between t1 to t2 ?

Am on board, but my only concern is that it won't change too much for highly risk averse xenos that will just wait longer away from the battle to evo.

> then T1 other than probably drone will essentially stop existing

Some players will always like staying rouny, but really it sounds like there just needs to be more incentive to play t1 castes rather than just because of arbitrary limits. All I can think of is the abstract concept of denying req points currently.

We could bring back non zero xeno respawn timers which increase with evo, make primo blessings cheaper for t1, and find other ways to encourage tier diversity.

Al-1ce:
> other than probably drone will essentially stop existing

There has been <sub>(basically)</sub> unlimited T2 slots for the past few months. In that time, T1s vanishing has never been a thing. It hasn't happened already, and I don't see any reason at all why it would later.

Goblinu:
T1's will always be in demand as they will a role T2's and T3's doesnt, runners are the fastest cast with the best hit and run kit, drones are good healers and caretakers, sentinels are . . . buff sentinel.

Wayland-Smithy:
Since the t2 slot math fix has gone live you can usually find t2 full and a handful t1s gone SSD. I propose this is because they'd rather go do something else than fill a role they find unfun.

The net result on balance is bad for xenos for 2 main reasons:

1. If the SSD xenos had just stayed larva that would grant a boost to the whole hive.
2. SSD xenos are hidden behind the "take SSD mob" button, so a xeno candidate could be in the queue for a long time not contributing to the fight when they might be happy to take over. I think some text in the join queue message could help here, but if it's Queen or a t1 with no upward mobility it'll be a hard sell regardless.

TiviPlus:
Not interested I would rather incorporate capacity into blessings

PR #11667, opened 2022-12-02
Disables the scatter increase per shot code from PR#11489

DeltaFire15:
## About The Pull Request
Effectively disables what #11489 did by commenting all integration of the new variables into the actual gun code.
## Why It's Good For The Game
According to that PR, "Currently i dind't actually make any gun use this system other than testing it[...]"

Except, that is false, and every single gun uses that system due to how that code is integrated. There is no check for any flag or something that determines whether to actually use the newly added system.

Due to this, every gun currently has its scatter permanently zerod, regardless of its situation, and is therefore laser accurate.
This is obviously not intended behavior, which is why this PR disables it.
## Changelog
:cl:
fix: Gun scatter is no longer permanently 0.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -1627,11 +1627,12 @@
 		*/
 		gun_scatter = scatter
 		wielded_fire = TRUE
+	/* Scatter increase per shot code.
 	else
-		/* Scatter increase per shot code.
+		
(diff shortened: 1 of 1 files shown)


Losenis:
It's weird, some guns do have scatter though noticeably very little, like T18 on BFA without any scatter-altering mods.

PR #11672, opened 2022-12-03
Actual lever action guns, assorted gun improvements and fixes

XElectricX:
## About The Pull Request

"Lever action" gun code implemented. No changes made to existing guns.
- Unique action opens and closes the receiver.
- Close receiver to fire, open it after firing, repeat
- Opening the receiver will eject the current bullet in the chamber, or a casing if fired
- Gun can only be un/loaded when the receiver is open.
- Gun can be cycled as fast as the user wants to. Still limited by the fire delay though.

Bug fixes & improvements:
- You can eject the last bullet in a gun through unique action
- gun/rifle/ had an incorrectly set `flags_gun_features`, removed duplications in children
- Fixed bug where shell ejections stopped occurring when there were less than 3 rounds left
- Cycling code changes to account for different scenarios of how a gun is or is not loaded

## Why It's Good For The Game

Proper yee-haw guns can be made.

## Changelog
:cl:
qol: Last bullet in a chamber can be ejected.
fix: Casing ejection fixes.
refactor: True lever action guns. Unique action key opens and closes receiver. Must be cycled after every shot.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -454,6 +454,8 @@
 	RegisterSignal(gun_user, COMSIG_KB_UNLOADGUN, .proc/unload_gun)
 	RegisterSignal(gun_user, COMSIG_KB_FIREMODE, .proc/do_toggle_firemode)
 	RegisterSignal(gun_user, COMSIG_KB_GUN_SAFETY, .proc/toggle_gun_safety_keybind)
+	scatter = min_scatter
+	scatter_unwielded = min_scatter_unwielded
 
 
 ///Null out gun user to prevent hard del
@@ -477,9 +479,9 @@
 	. = ..()
 	if(CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_TOGGLES_OPEN) && !CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_CLOSED))
 		icon_state = base_gun_icon + "_o"
-	else if(CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_REQUIRES_UNIQUE_ACTION) && !in_chamber && length(chamber_items))
+	else if(!CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_TOGGLES_OPEN) && CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_REQUIRES_UNIQUE_ACTION) && !in_chamber && length(chamber_items))
 		icon_state = base_gun_icon + "_u"
-	else if((!length(chamber_items) && max_chamber_items) || (!rounds && !max_chamber_items))
+	else if((!length(chamber_items) && max_chamber_items) || !rounds)
 		icon_state = base_gun_icon + "_e"
 	else if(current_chamber_position <= length(chamber_items) && chamber_items[current_chamber_position] && chamber_items[current_chamber_position].loc != src)
 		icon_state = base_gun_icon + "_l"
@@ -1017,6 +1019,26 @@
 	if(!length(chamber_items) && in_chamber && !CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_TOGGLES_OPEN) && !CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_REQUIRES_UNIQUE_ACTION))
 		unload(user)
 		return
+	if(CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_REQUIRES_UNIQUE_ACTION) && CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_TOGGLES_OPEN) && !CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_ROTATES_CHAMBER))	//Typically lever action guns
+		if(CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_CLOSED))	//Reciever is currently closed, let's open it
+			DISABLE_BITFIELD(reciever_flags, AMMO_RECIEVER_CLOSED)
+			playsound(src, opened_sound, 25, 1)
+			cycle(user, FALSE)
+			if(casings_to_eject)	//Eject the current shell/cartridge when opening the reciever
+				make_casing()
+				casings_to_eject = 0
+			to_chat(user, span_notice(chamber_opened_message))
+			if(in_chamber)
+				chamber_items.Insert(current_chamber_position, in_chamber)
+				in_chamber = null
+		else	//Reciever is currently open, let's close it
+			ENABLE_BITFIELD(reciever_flags, AMMO_RECIEVER_CLOSED)
+			playsound(src, cocked_sound, 25, 1)
+			to_chat(user, span_notice(chamber_closed_message))
+			cycle(user, FALSE)
+		update_ammo_count()
+		update_icon()
+		return
 	if(CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_REQUIRES_UNIQUE_ACTION) && !special_treatment)
 		if(last_cocked + cock_delay > world.time)
 			return
@@ -1325,7 +1347,7 @@
 
 ///Cycles the gun, handles ammunition draw
 /obj/item/weapon/gun/proc/cycle(mob/living/user, after_fire = TRUE)
-	if(!length(chamber_items) || (CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_MAGAZINES) && !get_current_rounds(chamber_items[current_chamber_position])))
+	if(!length(chamber_items) && (CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_HANDFULS) && !in_chamber) || (CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_MAGAZINES) && (!length(chamber_items) || !get_current_rounds(chamber_items[current_chamber_position])) && after_fire))	//Skip cycling if: manual operated has no bullets left, pulled trigger on empty magazine-fed gun
 		update_ammo_count()
 		return
 	if((CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_ROTATES_CHAMBER) && !CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_MAGAZINES)) || (CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_ROTATES_CHAMBER) && CHECK_BITFIELD(reciever_flags, AMMO_RECIEVER_MAGAZINES) && get_current_rounds(chamber_items[current_chamber_position]) <= 0))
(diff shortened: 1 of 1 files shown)


KerbalCitizen:
Seems like you have to use unique action twice after firing a lever action (leicaster repeater probably) in order to shoot again, at least thats how I interpret your words. Pump action shotguns need to have the pump pulled back then pushed forward again, but its one unique action. People shooting lever actions irl can cycle the lever back and forth pretty quickly too. I don't get why it should take two unique actions to cycle the lever. 

"Gun can only be un/loaded when the receiver is open." What. A decent amount of lever action guns have a loading gate on the side that lets you reload at any point, For martini or whatever, that makes sense. But for leicaster repeater, you should be able to reload at any time. Also, who unloads guns? The only time I unload non mag fed guns is when a shotgun on my loadout comes with buck and I swap it to slugs or fletch. 

About it making using lever actions harder to use, it seems like you have to cycle exactly twice every shot. Cycle once and all you did was eject the casing. Cycle three times and you just ejected a live round. 

Currently, you just cycle once and if you try more times, it just says pump locked and refuses to cycle. So you can just try to cycle it before you enter combat to make sure a live round is chambered. 

I honestly don't see the point of this pr on tgmc. On a server where people care a lot about realism and RP, I guess it might make sense. But we just play tgmc like a TDM. Once again, who unloads guns, who wants lever actions to be harder to use, why can you only load lever actions when receiver open but side loading gates exist.

maikilangiolo:
I'm used to double-cycle mechanics from another server, so I don't mind this (I like it because despite making the guns harder to use, it's also more fun and cool).

Probably, the stats could use an improvement in exchange for the higher skill ceiling (does this also prevent lever action akimbo?), and maybe QoL (clicking with a packet on an open lever gun reloads with a delay like a tac reload for magazine guns)

Losenis:
I feel this is only increasing the chances of getting BYOND'd by adding additional inputs to do essentially the same thing as racking a bolt back and forth, made worse by how accidentally hitting spacebar thrice will effectively waste your shot and lock you out of shooting until you hit spacebar again. I can't consider that a skill issue because you won't accidentally rack the lever open on an actual proper gun and leave it as such to attempt shooting, whereas ingame it's hitting the exact same key for both actions.

Isn't doubling the amount of inputs just anti-QoL? It's not exactly realistic if you consider opening and closing separate actions since that'd need two separate hotkeys to avoid the aforementioned "I accidentally ejected my cartridge and left the lever entirely open and didn't realize until I wanted to actually shoot" issue, and needing two keys to operate it would be as anti-QoL as it gets for no discernible compensation on the gun to make it worth it in the first place.

Pumping is considered one operation. Racking a bolt is considered one operation. The only exception to the rule is the martini, which has a magazine capacity of 1 so actioning it counts as reloading it much like a revolver needing opening before reloading and closing afterwards. Even normal firearms that'd require racking a bolt if reloading without a round in the chamber to put one there are abstracted into just the insertion of the magazine. Why is actioning a lever considered two actions? How does it benefit it other than making it a pain in the ass to operate compared to basically every other type of gun?

XElectricX:
> Seems like you have to use unique action twice after firing a lever action (leicaster repeater probably) in order to shoot again, at least thats how I interpret your words. Pump action shotguns need to have the pump pulled back then pushed forward again, but its one unique action. People shooting lever actions irl can cycle the lever back and forth pretty quickly too. I don't get why it should take two unique actions to cycle the lever.

The aim isn't entirely out of realism, it's to make a unique class of guns. There's not a lot of mechanical variety. I'd do the same for pumps and bolts except that they already play fine the way they are.

> "Gun can only be un/loaded when the receiver is open." What. A decent amount of lever action guns have a loading gate on the side that lets you reload at any point, For martini or whatever, that makes sense. But for leicaster repeater, you should be able to reload at any time.

I looked at some more videos of lever action reloads and you are right, the hammer doesn't have to be down to insert shells. At the moment though, I think it is fine the way it reloads. Could always be changed to allow reloads independent of receiver status very easily if it's annoying.

> Also, who unloads guns? The only time I unload non mag fed guns is when a shotgun on my loadout comes with buck and I swap it to slugs or fletch.

It's just a bug fix. You could unload single bullets from any firearm using unique action except for the last one in the chamber.

> About it making using lever actions harder to use, it seems like you have to cycle exactly twice every shot. Cycle once and all you did was eject the casing. Cycle three times and you just ejected a live round.

Precisely. Very few skill issue guns.

> 
> Currently, you just cycle once and if you try more times, it just says pump locked and refuses to cycle. So you can just try to cycle it before you enter combat to make sure a live round is chambered.

I agree. This could be solved with chambered sprite states for lever actions or something else. Maybe an icon on the bullet count HUD to show if it's chambered.

> I honestly don't see the point of this pr on tgmc. On a server where people care a lot about realism and RP, I guess it might make sense. But we just play tgmc like a TDM. Once again, who unloads guns, who wants lever actions to be harder to use, why can you only load lever actions when receiver open but side loading gates exist.

Just adding variety to the gun mechanics.

XElectricX:
> I feel this is only increasing the chances of getting BYOND'd by adding additional inputs to do essentially the same thing as racking a bolt back and forth, made worse by how accidentally hitting spacebar thrice will effectively waste your shot and lock you out of shooting until you hit spacebar again. I can't consider that a skill issue because you won't accidentally rack the lever open on an actual proper gun and leave it as such to attempt shooting, whereas ingame it's hitting the exact same key for both actions.
> 
> Isn't doubling the amount of inputs just anti-QoL? It's not exactly realistic if you consider opening and closing separate actions since that'd need two separate hotkeys to avoid the aforementioned "I accidentally ejected my cartridge and left the lever entirely open and didn't realize until I wanted to actually shoot" issue, and needing two keys to operate it would be as anti-QoL as it gets for no discernible compensation on the gun to make it worth it in the first place.
> 
> Pumping is considered one operation. Racking a bolt is considered one operation. The only exception to the rule is the martini, which has a magazine capacity of 1 so actioning it counts as reloading it much like a revolver needing opening before reloading and closing afterwards. Even normal firearms that'd require racking a bolt if reloading without a round in the chamber to put one there are abstracted into just the insertion of the magazine. Why is actioning a lever considered two actions? How does it benefit it other than making it a pain in the ass to operate compared to basically every other type of gun?

I agree with a lot of there not being enough benefit. Was told it was best to leave the existing guns alone. The code will exist but can be used by someone else to make unique lever actions that take these mechanics into account. Personally, I dislike how blandly the guns operate and I wanted to make more ways to use one. I don't think there are any BYOND problems from the many hours I've tested lever actions.

Losenis:
BYOND plays nicely when it's a private server or deadpop hours, but that's the best case scenario. Nowadays it's highpop more often than not and TD tends to jump to over 30% if marines don't wipe or get wiped 5 minutes after shutters opening, making the chances of something breaking exponentially higher as the TD increases. BFA T18 already breaks by getting its burstfire interrupted and locking it into a state where its attachments no longer work, reverts to fullauto, and can't be stored in suit slot which can be absolutely devastating if you're knocked down and magharn decides not to work. Those problems in specific are irrelevant to this pr, but remember that this is BYOND we're playing on, it's a very unwieldy and temperamental platform that always has the chance of breaking things due to the engine itself shitting the bed rather than any code-specific bug. Adding more inputs than necessary is playing with fire that way, since sometimes you can't even tell if the game didn't recognize your input or it's just lagging out by time dilation and accidentally hit spacebar one time more or less than necessary, or some other weird otherwise unrelated cooldown is affecting it and cancelling your input. Things tend to be bland when stability and fuckupproofing is the priority yeah, but that's a necessary evil for this engine until we migrate to another in the next 50 years or a second server goes up to split the playerbase down to manageable levels.

XElectricX:
I think worst case scenario, you might cycle too fast and lose a bullet maybe? It's the equivalent of a light switch. Cycling code isn't that long either, just loads in the next bullet. This would be like being concerned about automatic guns existing.

Losenis:
Worst case scenario you try to shoot your gun and find out it does nothing while the xeno is mauling you. Maybe your mind is on auto or you're playing with 300ms ping that you have to chain your next actions before you even shoot and close the lever to then open it again and find out it doesn't shoot since it seems to be a thing you'll be absolutely spamming if its firerate is friendly to how fast you can do it. Actual worst case scenario you let the primordial queen run instead of critting it and potentially lose the round, since, well, worst case scenario has to be the actual worst case scenario. There'll also be the time you get knocked down and can't tell if the input actually went through and did it or not before losing control of the gun, because usually the chat is flooded by combat and it might be too noisy to tell as I found out the hard way the last time I ran martini. There's a lot of things that can go wrong with each input, and adding inputs only adds more links to the chain that each are a potential failure point. There's multiple reasons why people decided to go full Ocelot and just pack 50 revolvers instead of reloading the one they had, and this is one of them: Hit z to unwield, hit unique action key, swap hands, grab speedloader, click on revolver, put empty speedloader back in container, swap hands, use unique action, hit z to wield again. Hyperbole given that's 9 inputs at worst compared to 2, but you get the point.

Actually that might not even be hyperbole given that'd still be 9 inputs total for like 7 rounds that can be done outside of combat while this'd be 2 inputs for every single round not counting reloading, and all in the middle of combat to boot that you won't be doing in a calm manner in a controlled environment. That's a shitton of potential failure points for a gun that'll100% be viewed at as "more clunky to handle than other guns".

And remember, we're talking about high TD rounds in practice, the kind of things that have bullet freeze in midair and boiler globs hit things they should absolutely not while releasing the gas cloud a screen away from impact. _Everything_ is affected no matter how cheap in performance. As of right now there's 86 people in the server, and this number seems to be the norm nowadays.

image

maikilangiolo:
Mega copium due to skill issues in these comments. Just spend 5 minutes adjusting, it's that simple

XElectricX:
> Mega copium due to skill issues in these comments. Just spend 5 minutes adjusting, it's that simple 

Lot's have valid criticisms but some really are just this. Majority of the guns really lack a skill factor.

Losenis:
Adding skill factor to a gun will lead to most people not using it over something more reliable and less prone to TD shenanigans, and the few that know how to use ~~macros~~ it will mop the floor with them given they'll need to be buffed to account for the unreliable operation. If they're not buffed, they'll just be gimmick guns that will be largely ignore by most people to the point they might as well be seasonal. This already happened with DB juggling and Ocelot's revolvers.

~~The absolute state of xenos that we're talking about adding skill issues to guns rather than the combat with them~~

KrabSpider:
Can you post a video of this in action?

FIXTHERADBOXPLEASE:
Plenty of other servers use this for their bolt-actions and it's fun, reliable, and more engaging than the boring one-press. I've never had any problems using these sorts of guns in BYOND. It's just more tactile feedback.
Anyone mad about this PR just hates change. If you're trying to play TGMC with 300 ping you should probably reconsider how you're spending your time.

XElectricX:
Had to make a new branch because git fucked over the previous beyond repair. For the record (since the commit history is gone), this PR changed some of the current lever actions to use the new system but someone recommended it be removed and only the code stay so that the guns can be balanced separately.

Lumipharon (changes requested):
I'm somewhat dubious as to why we would actually want a two step lever action in the codebase.
Also there's still a bug with casing not dropping for the last round in the gun, if it's a mag fed gun. I think its to do with not cycling if there's no mag.

maikilangiolo:
> I'm somewhat dubious as to why we would actually want a two step lever action in the codebase.

It's fun, it's cool and it's absolutely not clunky to use. I say this personal experience (a lot) on another server eh two step bolt and lever guns. Even with my abysmal ping and lag, the complaints about high ping fucking you over are largely unfounded. Because it's pure muscle memory, after a few minutes you get the hang of it.
Fire and press whatever macro you set for unique action twice, fire and macro twice. It will also get easier by timing it with the sound of the action cycling.

Lumipharon:
> > I'm somewhat dubious as to why we would actually want a two step lever action in the codebase.
> 
> It's fun, it's cool and it's absolutely not clunky to use. I say this personal experience (a lot) on another server eh two step bolt and lever guns. Even with my abysmal ping and lag, the complaints about high ping fucking you over are largely unfounded. Because it's pure muscle memory, after a few minutes you get the hang of it. Fire and press whatever macro you set for unique action twice, fire and macro twice. It will also get easier by timing it with the sound of the action cycling.

Yes but why? We could make shotguns require two keys to pump but it adds literally nothing to the game.

PR #11674, opened 2022-12-04
Mech Pilots are now worth STRONG Larva points, and have HIGH smartgunner point value

Pariah919:
## About The Pull Request
Title, literaly. They went from shipside larva point cost (1) to STRONG (6) and up a level in smartgunner point value (1 to 2)
## Why It's Good For The Game
Mech pilot is clearly a role with a decent amount of power in the game, so it should be worth quite a bit of larva and smartgunner points.
## Changelog
:cl:
balance: Mech pilot is now worth 6 larva points from 1, and 1 more smartgunner spawn point.
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -375,8 +375,8 @@ Though you are a warrant officer, your authority is limited to the dropship and 
 	job_points_needed = 40
 	job_points = 15
 	jobworth = list(
-		/datum/job/xenomorph = LARVA_POINTS_SHIPSIDE,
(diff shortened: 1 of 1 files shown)


Lumipharon:
Why the SG points? Not that it really matters since there's only gonna be like, 1-2 mp, but doesn't sg points just mean more SG slots open?

KerbalCitizen:
MP must be balanced out by smooth brain SGers that shoot mech in the back since they don't think about FF.

Naaanii:
Specs aren't real. They can't hurt you.

PR #11677, opened 2022-12-04
Spears act like actual spears when thrown

XElectricX:
## About The Pull Request

Spears will not do a silly spinning animation when thrown. Instead, they are rotated towards their destination. Spears will also not "bounce" when hitting a turf.

## Why It's Good For The Game

Spears are cool, until you throw them. It's a spear, not a boomerang.

## Changelog
:cl:
qol: Spears look like actual spears when thrown and no longer spin. No longer bounce when hitting turfs.
/:cl:


code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
...
+		if(SOUTHEAST)
+			rotation_value = 135 - current_angle
+		if(SOUTH)
+			rotation_value = 180 - current_angle
+		if(SOUTHWEST)
+			rotation_value = 225 - current_angle
+		if(WEST)
+			rotation_value = 270 - current_angle
+		if(NORTHWEST)
+			rotation_value = 315 - current_angle
```
Lumipharon, line 273: Why don't you just use get_angle?

Lumipharon (changes requested):

code/game/objects/items/weapons/twohanded.dm
XElectricX: Did not know that existed. Is it more efficient? Comment says it's for beams, where I'd assume you want precision. Just aiming to get the general direction of the spear's throw.

Lumipharon: It gives you the exact angle, but is there a reason you wouldn't want it to be exactly aiming at your target?

XElectricX: Simply a performance concern. Exact angle's not necessary for this process, which is why I went with get_dir. Using anything other than 45 degree angles would result in a funky sprite as well.

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
...
+		var/mob/living/M = hit_atom
+		M.hitby(src, speed)
+	else if(isobj(hit_atom)) // Thrown object hits another object and moves it
+		var/obj/O = hit_atom
+		if(!O.anchored)
+			step(O, dir)
+		O.hitby(src, speed)
+	else if(isturf(hit_atom))
+		set_throwing(FALSE)
+	SEND_SIGNAL(src, COMSIG_MOVABLE_IMPACT, hit_atom)
```
TiviPlus, line 292: Signal duplication

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
...
+
+/obj/item/weapon/twohanded/spear/pickup(mob/user)
+	. = ..()
+	//Reset the angle of the spear when picked up off the ground so it doesn't stay lopsided
+	if(initial(current_angle) != current_angle)
+		var/matrix/rotate_me = matrix()
```
TiviPlus, line 298: Early return

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
...
+		rotate_me.Turn(initial(current_angle) - current_angle)
+		//Rotate the object in the opposite direction because for some unfathomable reason, the above Turn() is applied twice; it just works
+		rotate_me.Turn(-(initial(current_angle) - current_angle))
+		src.transform = rotate_me
```
TiviPlus, line 302: Src usage

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
...
+		if(202.5 to 247.5)
+			rotation_value = 225 - current_angle
+		if(247.5 to 292.5)
+			rotation_value = 270 - current_angle
+		if(292.5 to 337.5)
+			rotation_value = 315 - current_angle
+	current_angle += rotation_value
+	var/matrix/rotate_me = matrix()
+	rotate_me.Turn(rotation_value)
+	src.transform = rotate_me
```
TiviPlus, line 277: Src usage

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
...
+	. = ..()
```
TiviPlus, line 278: Use return ..()

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
 	sharp = IS_SHARP_ITEM_SIMPLE
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	attack_verb = list("attacked", "stabbed", "jabbed", "torn", "gored")
+	var/current_angle = 45	//Based on what direction the tip of the spear is pointed at in the sprite; maybe someone makes a spear that points northwest
+
+/obj/item/weapon/twohanded/spear/throw_at(atom/target, range, speed, thrower, spin, flying)
+	spin = FALSE
+	//Find the angle the spear is to be thrown at, then rotate it based on that angle at 45 degree increments
```
TiviPlus, line 255: This seems hacky but ok

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
 	sharp = IS_SHARP_ITEM_SIMPLE
 	hitsound = 'sound/weapons/bladeslice.ogg'
 	attack_verb = list("attacked", "stabbed", "jabbed", "torn", "gored")
+	var/current_angle = 45	//Based on what direction the tip of the spear is pointed at in the sprite; maybe someone makes a spear that points northwest
```
TiviPlus, line 251: Autodoc

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -248,6 +248,59 @@
...
+			rotation_value = 270 - current_angle
+		if(292.5 to 337.5)
+			rotation_value = 315 - current_angle
+	current_angle += rotation_value
+	var/matrix/rotate_me = matrix()
+	rotate_me.Turn(rotation_value)
+	src.transform = rotate_me
+	. = ..()
+
+//Same thing as atom/movable but removing the bouncing from hitting a turf
```
TiviPlus, line 280: Generally copy pasting code but missing 1 line is a bad thing

TiviPlus (changes requested):

code/game/objects/items/weapons/twohanded.dm
XElectricX: Because it's only meant to reference itself? Parent proc does it as well, and so do the BYOND docs for doing transformations. Proc happens on the source.

code/game/objects/items/weapons/twohanded.dm
XElectricX: Lumi recommended finding the precise angle as it has no impact performance. Sprite can look bad at some odd angles so it's better to find the precise angle and then turn it in any of the 8 directions. Could probably add another 8 directions so that it turns at every 22.5 degrees but that may be as far as it goes before the player gets confused at what they are looking at (in my opinion).

code/game/objects/items/weapons/twohanded.dm
XElectricX: Using the spin var on the throw_at() proc to check if the object should bounce or not then. Generally if it doesn't spin, probably isn't something that would bounce off.

code/game/objects/items/weapons/twohanded.dm
Lumipharon: I meant just use the precise angle full stop. There's not really a point in getting the precise angle but then just changing it into a cardinal angle anyway.

code/game/objects/items/weapons/twohanded.dm
Lumipharon: I haven't used transform much, but I don't see why src would be needed. transform is used on its own in various places in the code as is.

code/game/objects/items/weapons/twohanded.dm
XElectricX: Problem is it looks bad when it is not a cardinal angle. Sprites are important for conveying info. Imagine a player looking for their spear in the heat of battle and missing it because it was turned at a 68 degree angle and didn't look like a spear anymore.

code/game/objects/items/weapons/twohanded.dm
TiviPlus: using src.var unless var and an argument share a name is bad practice and we dont allow it

code/game/objects/items/weapons/twohanded.dm
```diff
@@ -253,25 +253,8 @@
...
-		if(157.5 to 202.5)
-			rotation_value = 180 - current_angle
-		if(202.5 to 247.5)
-			rotation_value = 225 - current_angle
-		if(247.5 to 292.5)
-			rotation_value = 270 - current_angle
-		if(292.5 to 337.5)
-			rotation_value = 315 - current_angle
+	//Find the angle the spear is to be thrown at, then rotate it based on that angle
+	var/rotation_value = Get_Angle(thrower, get_turf(target)) - current_angle
```
Lumipharon, line 257: you don't even need this. just current angle = get_angle

XElectricX: It's needed because spears are at a 45 degree angle already. If for example a spear was thrown north, the Get_Angle would be 0, and thus the rotation doesn't do anything, leaving the spear still pointing northeast.

code/game/objects/items/weapons/twohanded.dm
XElectricX: It was what was shown in the BYOND reference for transformations, thought it was needed but makes sense now.

code/game/objects/items/weapons/twohanded.dm
Lumipharon: Oh right, the actual rotation part, honk

Lumipharon (approved):

PR #11680, opened 2022-12-05
Scout/Sniper Cloak Balance :)

567Turtle:
## About The Pull Request

Lowers the price of both the scout cloak and the sniper cloak to 35.

## Why It's Good For The Game

Let me ask you, when's the last time you saw somebody run a scout cloak or sniper cloak? That's right. You've never seen somebody run a scout cloak or a sniper cloak. They're way too expensive to be viable, by the time you can afford 50 points for one item for one marine, the push is either dead as fuck or way too far in for a stealth unga to be viable. Now, this doesn't make them quite so viable that every marines gonna have one, but maybe you're gonna see one or two of them every few rounds.

## Changelog
:cl:
balance: Scout/Sniper cloak now 35 points from req
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -782,12 +782,12 @@ ARMOR
 /datum/supply_packs/armor/scout_cloak
 	name = "Scout Cloak"
 	contains = list(/obj/item/storage/backpack/marine/satchel/scout_cloak/scout)
-	cost = 50
+	cost = 35
 
 /datum/supply_packs/armor/sniper_cloak
 	name = "Sniper Cloak"
 	contains = list(/obj/item/storage/backpack/marine/satchel/scout_cloak/sniper)
-	cost = 50
+	cost = 35
 
 /datum/supply_packs/armor/grenade_belt
 	name = "High Capacity Grenade Belt"
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -787,7 +787,7 @@ ARMOR
 /datum/supply_packs/armor/sniper_cloak
 	name = "Sniper Cloak"
 	contains = list(/obj/item/storage/backpack/marine/satchel/scout_cloak/sniper)
-	cost = 35
+	cost = 30
 
 /datum/supply_packs/armor/grenade_belt
 	name = "High Capacity Grenade Belt"


code/modules/reqs/supplypacks.dm
```diff
@@ -782,12 +782,12 @@ ARMOR
...
 	name = "Scout Cloak"
 	contains = list(/obj/item/storage/backpack/marine/satchel/scout_cloak/scout)
-	cost = 50
+	cost = 35
 
 /datum/supply_packs/armor/sniper_cloak
 	name = "Sniper Cloak"
 	contains = list(/obj/item/storage/backpack/marine/satchel/scout_cloak/sniper)
-	cost = 50
+	cost = 35
```
Pariah919, line 790:
```suggestion
	cost = 30
```
30 points makes more sense considering this thing is a straight downgrade to scout cloak. You can't use it while moving, and so its straight up worse in every way..

code/modules/reqs/supplypacks.dm
```diff
@@ -782,12 +782,12 @@ ARMOR
 /datum/supply_packs/armor/scout_cloak
 	name = "Scout Cloak"
 	contains = list(/obj/item/storage/backpack/marine/satchel/scout_cloak/scout)
-	cost = 50
+	cost = 35
```
Pariah919, line 785:
```suggestion
	cost = 40
```
Scout cloak is stronger than you think, I think 40 points is fair honestly. 50 is too high a bracket for it though, yeah.

Pariah919 (commented):
I don't think they should be equal in price, sniper cloak is straight up worse. These are also bags so..

code/modules/reqs/supplypacks.dm
567Turtle: I know it's strong, I don't underestimate it, it's worth 50 points, its just too expensive to be viable when the scout cloak is best. By the time you can afford 50 points, its too late in the op for you to stealth insurge more most of the time. The 35 cost point allowed it to well, be used sometimes.

Luxyboo:
B-B-B-B-BAA-BASED????

FIXTHERADBOXPLEASE:
The Sniper cloak ain't "worse", but it's definitely more niche. It's essentially no more than a tarp on your back. It being cheaper than the 35-point Scout cloak would be fine I think.

Starfirejordan:
suggestion: 40 for scout cloak, 30 for sniper cloak
reason: scout cloak is far more useful then sniper as both are basically exactly the same besides one making you unable to move awhile the other lets you move

iLikeMemes:
Finally, i will be able to do my sadar + sniper cloak shenanigans for 15 points less.

567Turtle:
> i feel like if the cloaks were made more effective (look at CM for the example) the price would be worth it, but they're kinda useless as you become visible after only a couple of steps

This, honestly. I tried out the scout cloak for a few rounds and it's kind of shit, you can barely move a tile without practically becoming fully visible again. If it were more effective like in CM it'd be worth a higher price but honestly in the scout cloaks current state I struggle to come up with why it isn't like 20 points.

KerbalCitizen:
I think cloaks are perfectly fine at their current price if used well. Pretty sure I saw turtle chasing with a cloak activated, worst way to use cloak imo because it turns visible so fast. Also, some people trying to larp as Snake, often trying to stealthily destroy silo in this day and age. 

Wear mimir, carry some razorwire, have cloak smoke nades. Throw cloak smoke near a 1 tile wide choke point. Run next to the chokepoint, activate cloak, rest to avoid FF and bumpslash. When xeno runs past, stand up, grab razorwire (this is actually the most visible part because of item grabbing animation), place it to block the chokepoint and therefore the xeno's retreat path. 

Or just stay stationary with stuff like rocket sledge or RR, weapons that can hit hard and their main counter is just seeing it and avoiding it.

Araulius:
can you even lase targets without being detected with these cloaks?

KerbalCitizen:
> can you even lase targets without being detected with these cloaks?

I believe lasing will still make the green circle over your head, so any xeno paying attention could see you fairly easily. Rangefinding for mortar or howi doesn't make any green circle though since it's instant. 

Unsure about this, but you might be able to lase while inside of a tarp. If you can, then that would be way better than a cloak.

RipGrayson:
Keeping this open as an alternative to https://github.com/tgstation/TerraGov-Marine-Corps/pull/11745

PR #11684, opened 2022-12-06
Ravager Vampirism Rework

Zar000:
## About The Pull Request
Makes vampirism actually do something.

Ravager will heal for 25% of missing health per hit with a 2 second cooldown.

Yes it calculates even if you are under 0 health in endure

## Why It's Good For The Game

Primo is currently useless, reward for continued prolonged fights is rather low and if anything discouraged, this will change that.
## Changelog
:cl:
add: New primo to ravager
balance: The primo for ravager
add: Cool blood dripping effect when used
/:cl:


Lumipharon:
uwot? This seems hilariously high considering the low cooldown and it stacks with superrage healing.
Superrage had to get nerfed due to its absurd healing factor, keep in mind.

Luxyboo:
i feel like 10% would be much better than 25% though it would need some testing.

FIXTHERADBOXPLEASE:
Yeah, Ravager's primo is both underwhelming and counter-intuitive. This is a good step away from that. Even if bumped down to something small like 10 or 15%, it'd still be preferable.

Al-1ce:
> uwot? This seems hilariously high considering the low cooldown and it stacks with superrage healing. Superrage had to get nerfed due to its absurd healing factor, keep in mind.

The health cap to use superrage is too low to be efficiently used. I can say this with confidence as someone who usually goes ravager.

code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
```diff
@@ -491,38 +499,25 @@
...
+	x.adjustFireLoss(-((x.xeno_caste.max_health - x.health) / 4))
 	update_button_icon()
+	COOLDOWN_START(src, last_healed, heal_delay)
+	play_particles(x)
 
-///Called when the leech effect is supposed to end
-/datum/action/xeno_action/vampirism/proc/end_leech()
-	leech_count--
-	update_button_icon()
+/datum/action/xeno_action/vampirism/proc/play_particles(attacker)
```
TiviPlus, line 518: inline

code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
```diff
@@ -491,38 +499,25 @@
...
+	particle_holder = new(attacker, /particles/xeno_slash/vampirism)
+	particle_holder.pixel_y = 18
+	particle_holder.pixel_x = 18
+	QDEL_IN(particle_holder, 2 SECONDS)
```
TiviPlus, line 523: this can bug out and due to tidi make it so the particles get deleted preemtively
also leaves a hanging ref so this will hard del

TiviPlus (changes requested):

PR #11692, opened 2022-12-07
Queen screech now gives stagger

MLGTASTICa:
## About The Pull Request
Queen screech gives stagger

## Why It's Good For The Game
The ability is better used against larger groups of marines , by providing them with the damage debuff , instead of it currently being used on solo marines to secure easy kills.

## Changelog
:cl:
balance: Queen screech now gives a maximum of 15 stagger , adjusted based on distance , with the minimum being 3.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -394,7 +394,7 @@ Contains most of the procs that are called when a mob is attacked by something
 	Stun(stun_duration)
 	Paralyze(stun_duration)
 	//15 Next to queen , 3 at max distance.
-	adjust_stagger(LERP(15, 3, dist_pct) * reduction)
+	adjust_stagger(LERP(7, 3, dist_pct) * reduction)
 	//Max 140 under Queen, 130 beside Queen, 70 at the edge. Reduction of 10 per tile distance from Queen.
 	apply_damage(LERP(140, 70, dist_pct) * reduction, STAMINA, updating_health = TRUE)
 	if(!ear_deaf)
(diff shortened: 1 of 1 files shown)


EvilMelonTheSecond:
As long as the stun remains it will be used against solo marines. This buff won't change how its used.

MLGTASTICa:
> Contributor

it will always be used against solo marines , but now this makes it not be useless against larger groups , and perhaps better for initiating pushes against marines

UntoldTactics:
Pretty good thing. Queen screech as a whole being used against solo marines is purely because securing a kill is vital. Plus, if your alone and encounter a queen that's a you problem.

The stagger will definitely make it actually more useful against large groups, as it often doesn't do too much in that specific regard, since marines just get right back up, usually only works well when screeching a dying group, but thats just a finisher.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -390,10 +390,12 @@ Contains most of the procs that are called when a mob is attacked by something
 	var/reduction = max(min(1, reduce_within_sight - reduce_prot_aura), 0.1) // Capped at 90% reduction
 	var/stamina_damage = LERP(140, 70, dist_pct) * reduction //Max 140 under Queen, 130 beside Queen, 70 at the edge. Reduction of 10 per tile distance from Queen.
 	var/stun_duration = (LERP(1, 0.4, dist_pct) * reduction) * 20 //Max 1.5 beside Queen, 0.4 at the edge.
+	var/stagger_amount = LERP(15, 3, dist_pct) * reduction // 15 Next to queen , 3 at max distance.
```
BraveMole, line 393: All of those vars are complely uneeded, just put that line in the adjust_stagger

Same for stam damage, etc

BraveMole (changes requested):

Lumipharon:
I'm not opposed to stagger on screech, but uh, I fail to see how applying 30(?) seconds of 50% damage on an unga would make me want to solo screech someone any less.

RipGrayson (approved):

RipGrayson:
I've got to move things along, I'm merging this in 24 hours if we don't have any objections.

Lumipharon:
> I've got to move things along, I'm merging this in 24 hours if we don't have any objections.

for reference buckshot does 1 stack of stagger, warrior punch does 3 (6 if grappled). Up to 30 seconds of 50% damage on everyone screen wide is very strong.

I would have it much lower like LERP(7, 3, dist_pct).

RipGrayson (changes requested):
I defer to Lumi's superior knowledge here, please reduce the stagger, 30 seconds of 50% damage is way too strong.

Lumipharon (approved):

PR #11709, opened 2022-12-12
GUI Fix for medical analyzer's display of infections/necrosis

collinlunn:
## About The Pull Request
This PR makes changes to the GUI for healthanalyzer, to correct the following usability issues.
1. If any limb was infected, every limb was labelled as infected. Now, only the specific limb is is labelled.
2. Limbs who where infected or necrotizing were not marked as such unless they had another condition. Now, they are labelled.
3. If a limb has necrosis, it suppressed the limb being labelled with infected. Now, they are separate (This is important because limbs can have necrosis while not being infected).

## Why It's Good For The Game
Due to the above oddities with the GUI, infections seemed to exhibit strange behavior. This should make it more clear to the user what is happening. 

## Changelog
:cl:
fix: Heath analyzers are now more clear about what limbs have infections and necrosis.
/:cl:

### DIFF
--- a/code/game/objects/items/scanners.dm
+++ b/code/game/objects/items/scanners.dm
@@ -176,12 +176,12 @@ REAGENT SCANNER
 						continue
 					internal_bleeding = TRUE
 					break
-			if(limb.limb_status & LIMB_NECROTIZED)
-				infection_message = "Subject's [limb.display_name] has necrotized. Surgery required."
-				necrotized = TRUE
 			if(limb.germ_level > INFECTION_LEVEL_ONE)
 				infection_message = "Infection detected in subject's [limb.display_name]. Antibiotics recommended."
 				infected = TRUE
+			if(limb.limb_status & LIMB_NECROTIZED)
+				infection_message = "Subject's [limb.display_name] has necrotized. Surgery required."
+				necrotized = TRUE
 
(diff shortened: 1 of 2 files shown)


collinlunn:
TqsL0TIzjt

collinlunn:
Regarding infection_message, I have adjusted it so the necrosis message supersedes the infection message.

For the issue of necrosis implying infection, the problem is that certain treatment tools remove infection without removing necrosis, and there's not a clear indication that they do so - because the UI makes necrosis look like a subtype of infection, which doesn't go away when removing infection.

Working on making tests pass.

Losenis:
Imperfect information on health analyzers as a concept sounds nice to people who don't have to do a quadrillion scans per round and have all the benefit of sitting in fob forever to take their time handling it all tbh. Medical difficulty should be in the procedures, not in questioning whether the info you have is even right when the other guy is the one who's going to be punished for misdiagnosing and gets med rookies roasted into not playing the role again rather than learn through mistakes he might not even realize were mistakes.

XElectricX:
Speed merge this PR right this moment. If you're on autopilot and forgot which limb you were working on or got someone who already had necro/infection from another doctor, it was a game of roulette trying to find which limb to de-necrotize. All these fixes are vital.

PR #11711, opened 2022-12-12
Sunder adjustments

Waselon:
## About The Pull Request

Xenos get a sunder resistance based on missing sunder. 0% at 100 sunder, 30% at 70 sunder, 50% at 50 sunder, etc.

Drone, Queen and Shrike healing abilities got a sunder healing buff.

## Why It's Good For The Game

Tanky xenos can tank regular firearms a bit more succesfully. Marines have to be a bit more careful in their weapon choice, and consider bringing something to deal with armored castes.

Buffed Healing Salve allows to spread sunder healing among other xenos, so Hivelord isnt the only one doing that.

## Changelog
:cl:
balance: added sunder resistance based on missing sunder, buffed sunder healing for drone, queen and shrike healing abilities.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -1409,6 +1409,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 /mob/living/proc/get_sunder()
 	return 0
 
+///returns sunder cap value for xenos, in range from 1 to 0
 /mob/living/proc/get_sunder_cap()
 	return 0
 
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -909,7 +909,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 
 	if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING)
 		if(get_sunder() > 0.5)
-			adjust_sunder(proj.sundering)
+			adjust_sunder(min(proj.sundering, get_sunder() * 100 - 50))
 		else if(proj.ammo.flags_ammo_behavior & AMMO_HARD_SUNDERING)
 			adjust_sunder(proj.sundering)
 
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -908,10 +908,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 			feedback_flags |= (BULLET_FEEDBACK_FIRE)
 
 	if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING)
-		if(proj.ammo.flags_ammo_behavior & AMMO_HARD_SUNDERING)
-			adjust_sunder(proj.sundering)
-		else if(get_sunder() > get_sunder_cap())
-			adjust_sunder(min(proj.sundering, (get_sunder() - get_sunder_cap()) * 100))
+		adjust_sunder(proj.sundering)
 
 	if(damage)
 		var/shrapnel_roll = do_shrapnel_roll(proj, damage)
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -908,7 +908,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 			feedback_flags |= (BULLET_FEEDBACK_FIRE)
 
 	if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING)
-		adjust_sunder(proj.sundering)
+		adjust_sunder(proj.sundering * get_sunder())
 
 	if(damage)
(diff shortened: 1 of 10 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -908,7 +908,10 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 			feedback_flags |= (BULLET_FEEDBACK_FIRE)
 
 	if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING)
-		adjust_sunder(proj.sundering)
+		if(get_sunder() > 0.5)
+			adjust_sunder(proj.sundering)
```
DeltaFire15, line 912: This has some issues with hardcapping sunder, if you want to hardcap it you have to make sure you limit the sunder the shot can apply or the very last sundering impact pre 50% sunder will "overflow" and lead to more sunder than should be possible (assuming the 50% "hard" limit for base projectiles)

DeltaFire15:
On another note, something to think about with this would be to have this cap a var on xeno castes so that more tanky xenos might be more resistant to non-true sundering shots, or squishy ones less so, aswell as maybe having sunder resist be lower at younger maturities.

code/modules/projectiles/projectile.dm
Waselon: I know. If thats a big issue, ill fix it

Waselon:
> On another note, something to think about with this would be to have this cap a var on xeno castes so that more tanky xenos might be more resistant to non-true sundering shots, or squishy ones less so, aswell as maybe having sunder resist be lower at younger maturities.

done

RipGrayson (approved):

Pariah919:
Could give some laser guns hard sunder to give them more of a gimmick. Laser Rifle overcharge is one I can think of.

Waselon:
> Could give some laser guns hard sunder to give them more of a gimmick. Laser Rifle overcharge is one I can think of.

Might do that in different PR

code/modules/projectiles/projectile.dm
```diff
@@ -908,7 +908,10 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 			feedback_flags |= (BULLET_FEEDBACK_FIRE)
 
 	if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING)
-		adjust_sunder(proj.sundering)
+		if(proj.ammo.flags_ammo_behavior & AMMO_HARD_SUNDERING)
```
Lumipharon, line 911: I would put this bit of code in adjust sunder itself, imo. Sunder is only projectiles right now, but might not always be. You can make hard sunder an arg.

Lumipharon (changes requested):
Hard limits are kinda funky, and mean you get weird interactions depending on the order in which things are hitting the xeno.

Have you considered making sunder past the sunder cap apply at a significantly reduced rate instead of just not at all?

Waselon:
> Hard limits are kinda funky, and mean you get weird interactions depending on the order in which things are hitting the xeno.
> 
> Have you considered making sunder past the sunder cap apply at a significantly reduced rate instead of just not at all?

Having sunder resistance at ceratin breakpoints would be less predictable. Another option i considered was naking sunder resistance (damage muliplyer) linear, 1 at 0 sunder, 0.7 at 30 sunder, 0.5 at 50 sunder etc.

Another reason for hardcap is i wanted to make stripping xeno armor to 0 and maintaining it take some effort.

code/modules/projectiles/projectile.dm
Waselon: Explosions sunder is applied in explosions code, projectile sunder is applied in projectile code, idk why would you want to split it

Lumipharon: ?
They both use the same proc do they not? I'm saying put the code in that proc so it's not split like you currently have it.

Waselon: they both use apply_sunder proc, im just changing the value applied. Why would i want to bring projectile code into the xenoproc?

Lumipharon: It doesn't matter, someone else can do it in the future, but I meant adding an arg to apply_sunder on whether to respect the sunder limit or not. Then in projectiles, the arg is true/false based on the flag. But it would mean other sunder sources could also use it in the future more readily.

Lumipharon (approved):
Lets see how it goes, although this may be a bit funny with the high sunder stuff since it's only calculated once.

Losenis:
This shold've been TMd rather than merged. Xenos are hardly ever getting past the second stage of the broken shield icon. With strong ward to get even more armor + frenzy to run away faster and not get additional sunder on retreat, this made xenos extremely tanky across the board.

You can't really chip away at their armor engagement after engagement so it's become very difficult to deal with them if they just run and come back fully fresh in seconds rather than either returning with broken armor or taking more time to rest and recover leaving for some respite to marines.

Also how does this interact with decimals? Does it round up, down, or are the decimals kept on the armor sundering when 0.5 sunder gets math'd down to like 0.35?

PR #11715, opened 2022-12-14
Howitzer removed from roundstart, reworked some shells, added mustard shell.

Pariah919:
## About The Pull Request
Howitzer HE now does much more Heavy on location hit.
Howitzer shells have a generally slower speed.
Howitzer WP has slightly more flame radius.
Howitzer is no longer roundstart, and has been replaced with a spare mortar.
Howitzer has a 3 second reload time, and now 25 second cooloff.
Added new howitzer shell that releases a burning gas (basically WP without the fire and much longer lasting) that damages anything in the area. Lasts 40 seconds. Costs 4 points for 2 shells.
## Why It's Good For The Game
Reworks howitzer into true lategame artillery, and now works better at breaking apart mazes with devastate. The higher cooloff will make artillery less spammy and require marines to invest more into it, and the new shelltype lets marines deny area to both sides for a period of time, but allows you to pass through it still, at the expense of damage.
## Changelog
:cl:
balance: Howitzer reworked, HE does much more heavy on hit, incend has a bit more range, new chemical gas shell type. Has cooloff of 25 seconds, it is no longer roundstart. It now costs 75 points.
/:cl:


Naaanii:
THE ARTILLERY IS REAL

Edelor68:
Good change except for the cooldown. 
If you want to counteract spam, make the shells more expensive so players have to ration them better or add a heat-up mechanic, but this just feels lazy and like you are brute forcing the problem.

maikilangiolo:
I feel like loading taking time is a balancing factor that doesn't require a cooldown. Increment the load time to 5 seconds, but remove the silly cooldown. It won't be spammable without xenos being able to run away.
It's the same thing as the MLRS, with the stupid cooldown despite taking a long time to reload already.

Araulius:
did you also apply the cooldown to the mortar?

Starfirejordan:
> did you also apply the cooldown to the mortar?

pretty sure they removed it after my feedback and if im given permission will be making a PR for mortar

Araulius:
> > did you also apply the cooldown to the mortar?
> 
> pretty sure they removed it after my feedback and if im given permission will be making a PR for mortar

if you ended up putting a cooldown on the mortar, you just end up destroying the quad barrel mortar's whole gimmick

MereWalker:
I like this change. I personally think a ten second cooldown is better. What will will the howi cost? Will it come with ammo once bought as well? As for the second mortar. Does mean a second mortar piece or the entire box. I order a second box since it would make sense.

Pariah919:
>I like this change. I personally think a ten second cooldown is better. What will will the howi cost? Will it come with ammo once bought as well? As for the second mortar. Does mean a second mortar piece or the entire box. I order a second box since it would make sense./ edit / I just read it. I see that it's two mortar crates. I like that. The cost for the howie seems a bit high for my tastes. Will have to see if it's worth it.
>
Its a second box for the mortar, and 10 seconds is too short considering the howitzer now has devastate on near-direct contact for most xeno castes, rather than just on direct contact. I prefer the noticeable cooloff to incentivize buying more howitzers (or quantity saturation artillery like MLRS) if you want to put a lot of fire downrange.

>if you ended up putting a cooldown on the mortar, you just end up destroying the quad barrel mortar's whole gimmick
>
Quad barrel could use some cooloff. You're firing two rounds so you should have some higher level of requirement, however its slaved to the TAT so I can care less.

>Good change except for the cooldown.
If you want to counteract spam, make the shells more expensive so players have to ration them better or add a heat-up mechanic, but this just feels lazy and like you are brute forcing the problem.
>
I want to incentivise people to buy more artillery if they want more shells down range, its more of a personal design thing.
I don't see how this is lazy? It has a cooldown, its barely half a minute. Nerfing shell prices seems to extreme as in my opinion the entry cost should be the main thing you're paying for.

Kuro020 (approved):

Lumipharon:
3 range dev means this can gib xenos in a 49 tile box, which seems uh, strong.

Pariah919:
Reduced devastation, and upped heavy explosion by another 1 instead. @Lumipharon

code/modules/reqs/supplypacks.dm
```diff
@@ -709,6 +709,11 @@ EXPLOSIVES
 	contains = list(/obj/item/mortal_shell/howitzer/incendiary)
 	cost = 40
 
+/datum/supply_packs/explosives/mortar_ammo_mustard
+	name = "MG-100Y howitzer ;Mustard' chemical weapon shell (x2)"
+	contains = list(/obj/item/mortal_shell/howitzer/mustard, /obj/item/mortal_shell/howitzer/mustard)
+	cost = 4
```
Pariah919, line 715:
```suggestion
	cost = 40
```

Starfirejordan:
Please dont merge this

code/modules/reqs/supplypacks.dm
```diff
@@ -709,6 +709,11 @@ EXPLOSIVES
 	contains = list(/obj/item/mortal_shell/howitzer/incendiary)
 	cost = 40
 
+/datum/supply_packs/explosives/mortar_ammo_mustard
+	name = "MG-100Y howitzer ;Mustard' chemical weapon shell (x2)"
```
XSlayer300, line 713:
```suggestion
	name = "MG-100Y howitzer 'Mustard' chemical weapon shell (x2)"
```

code/game/objects/machinery/mortar.dm
```diff
@@ -601,6 +603,13 @@
 	icon_state = "howitzer_ammo_wp"
 	ammo_type = /datum/ammo/mortar/smoke/howi/wp
 
+/obj/item/mortal_shell/howitzer/mustard
+	name = "\improper 150mm chemical X-50 'Mustard' artillery shell"
+	desc = "An 150mm artillery shell, loaded with an experimental gas known as X-50 which chokes, burns and lasts a long time. Known better as 'Mustard Gas' due to the bright yellow colorant it uses to denote its location to troops, whatever is hit by this will have their day, skin and future ruined, with a demand for a warcrime tribunal."
```
XSlayer300, line 608:
```suggestion
	desc = "A 150mm artillery shell, loaded with an experimental X-50 toxic chemical gas which chokes, burns and lasts a long time. Known better as 'Mustard Gas' due to the bright yellow colorant it uses to denote its location to troops, whatever is hit by this will have their day, skin and future ruined, with a demand for a warcrime tribunal."
```

Starfirejordan:
> I don't know how long this has been test merged, so take my opinion with a grain of salt but ive only seen the howitzer once, and that was today, if that speaks anything for how much more uncommon its going to be that or im a schizophrenic one or the other.

I love banishing things to req to die, I love banishing things to req to die

chemistrymain2:
This isn't fun.

The 25 second delay is obscene.

Increase the loading time instead.

Coastfront:
Ahem, further time has passed, along with further rounds etching their way unto my slate of history. And further still do I fail to encounter the howitzer anymore.

To the genuine point that I don't even see people consider buying it now. I haven't even seen somebody ask for it, nor req even mention "hey i can buy howie :)" so if the overall goal of this is to sweep away the dirt beneath howie into a boundless abyss into which it shall fall, it seems sufficient in every way for such a task. Perfect as a scythes edge to reap a harvest.

Pariah919:
after like 4 months of testing, its clear that this PR failed. I'll throw it back into the drawing board and try again another year, I guess.

PR #11427, opened 2022-11-02
Limit the availability of M-22 welder kits

Lumipharon:
## About The Pull Request
Welding kits are no longer unlimited, but are available through req for 5 points.

## Why It's Good For The Game
It's a little too easy to spam flamers forever. Now there's either a cost involved, or you have to deal with constantly reloading the small tanks.
I could be a real grinch and limit the small tanks as well, but I'll leave that for now.

## Changelog
:cl:
balance: M-22 welder kits are no longer unlimited, but can now be purchased from req for 5 points
/:cl:

PR #11719, opened 2022-12-15
Removes Infinite flamerfuel source by setting avaliable welderpacks to 20

ClosetedSkeletonGH:
## About The Pull Request
This PR sets the amount of welderpacks available within in clothing vendor from infinite to 20. 

This PR does not change the avaliable amount of welderpacks from other vendors such as engi base equipment vendor.

## Why It's Good For The Game
Right now, being a xeno, mazing is extremely unenjoyable but it is absolutely vital to winning. The main factor of it being unenjoyable is the fact that flamer fuel is functionally infinite, meaning people can spray extended nozzle as much as they like clearing huge swaths of mazes, refuel from their backpack until it is empty, then go get another backpack/have a backpack sent down to them for absolutely free and no drawbacks, instead of having to use a gas tank which can be found around the map/bought by RO. 

This PR seeks to limit infinite flamer fuel by cutting the amount of available welder packs from infinite to 20, while this doesn't make it a viable choice for engi's to take from their equipment vendor, it does solve the infinite flamer fuel issue.

## Changelog
:cl:
balance: Sets the amount of welderpacks in the clothing vendor to 20
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -1338,7 +1338,7 @@
 		"Backpacks" = list(
 			/obj/item/storage/backpack/marine/standard = -1,
 			/obj/item/storage/backpack/marine/satchel = -1,
-			/obj/item/tool/weldpack/marinestandard = -1,
+			/obj/item/tool/weldpack/marinestandard = 20,
 		),
 		"Instruments" = list(
 			/obj/item/instrument/violin = -1,


FosterGrime-Heymynameslouie:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/11427

I believ this was already discussed here

do agree that something has to be done about the flamer on maze thing but I dont think outright chopping flamer supply is the best way to go about it, albeit it is one of ways to go about it.

Flamer rework by tye I think is on the works and talk about fireproof resin structures is going around so I think we could wait for those to pop in before considering nuking fuel supply? Beucase this also affetcts droid healing and welding cades

FIXTHERADBOXPLEASE:
Yeah there's already a flamer rework being made.

Al-1ce:
My disapointment is imeasurable and my flaming is ruinedl

Naaanii:
Nerfing flamers would be a better solution either way.

Losenis:
It was already discussed why this'd impact everything but flamers already

Beepinex:
what if flamethrowers got their own unique fuel reagent and _couldn't_ use welding fuel?

KerbalCitizen:
I prefer extended nozzle for maze clearing because of the convenience of low fuel consumption, I stay safe because 6 tile range, and ungas can safely push/ chase because my flames don't block them off. This PR is meaningless if marines are using extended nozzle. On a really long round, I can use up like 3 welding kits max. 1500 resin walls burned with 3 welding kits. 

Making welding fuel technically but not practically limited. Not only did you fail to nerf flamers (which I honestly think are a bit too good against mazes), you just pissed off a bunch of people.

Coastfront:
Nerf flamethrowers not robots, some other guy already tried this, didnt work, people gave their arguements there.

RipGrayson (commented):
I agree with the reasoning of this PR and do think flamers should be more restricted, but this change hurts robots more than it would hurt flamers.

Tomorrow I'll try to get in touch with tye and see how the flamer rework is moving along. If it's dead I'll do some thinking about how to address the flamer spam.

Lumipharon:
This isn't quite the right solution.

PR #11720, opened 2022-12-15
Makes limb replacement surgery respect radial choice

collinlunn:
## About The Pull Request
Currently, the limb replacement surgery does not work with radial - you need to both select the right part on the radial AND select the right part on the targeting doll with the prosthetic. This corrects that, you now only need to select the correct radial part if using radial.

## Changelog

:cl:
fix: Limb replacement surgery's attach prosthetic step now works with the radial menu.
/:cl:

### DIFF
--- a/code/game/objects/items/robot/robot_parts.dm
+++ b/code/game/objects/items/robot/robot_parts.dm
@@ -11,25 +11,25 @@
 	name = "robot left arm"
 	desc = "A skeletal limb wrapped in pseudomuscles, with a low-conductivity case."
 	icon_state = "l_arm"
-	part = list(ARM_LEFT, HAND_LEFT)
+	part = list(BODY_ZONE_L_ARM, BODY_ZONE_PRECISE_L_HAND)
 
 /obj/item/robot_parts/r_arm
 	name = "robot right arm"
 	desc = "A skeletal limb wrapped in pseudomuscles, with a low-conductivity case."
 	icon_state = "r_arm"
-	part = list(ARM_RIGHT, HAND_RIGHT)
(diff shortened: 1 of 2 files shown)


Al-1ce:
wicked

code/game/objects/items/robot/robot_parts.dm
```diff
@@ -11,25 +11,25 @@
 	name = "robot left arm"
 	desc = "A skeletal limb wrapped in pseudomuscles, with a low-conductivity case."
 	icon_state = "l_arm"
-	part = list("l_arm", "l_hand")
+	part = list(ARM_LEFT, HAND_LEFT)
```
QualityVan, line 14: These are clothing protection bitflags. Check defines/mobs.dm for BODY_ZONE ones instead.

QualityVan: Wait, limbs actually use these ones for surgery/body_part? That's weird. Disregard me then.

code/modules/surgery/robolimbs.dm
```diff
@@ -132,7 +132,7 @@
 	if(..())
 		var/obj/item/robot_parts/p = tool
 		if(p.part)
-			if(!(target_zone in p.part))
+			if(!(affected.body_part in p.part))
```
QualityVan, line 135: Oh, it's because you changed it here. You can use affected.name instead and use the proper BODY_ZONE defines.

code/game/objects/items/robot/robot_parts.dm
collinlunn: I've modified the part list to use the correct BODY_ZONE defines, and for the surgery code to check the  limb name instead.

PR #11729, opened 2022-12-16
Req Console Upgrade

SiegeB0t:
## About The Pull Request
Currently, req consoles throw a tgui error when opened by ghosts. Now ghosts can view the consoles.
## Why It's Good For The Game
I want to rage in dchat whenever a bald RO buys 500 points worth of plasteel.

## Changelog
:cl:
qol: Ghosts can view supply console information.
/:cl:

### DIFF
--- a/code/modules/reqs/supply.dm
+++ b/code/modules/reqs/supply.dm
@@ -263,7 +263,7 @@ GLOBAL_LIST_INIT(blacklisted_cargo_types, typecacheof(list(
 	icon = 'icons/obj/machines/computer.dmi'
 	icon_state = SHUTTLE_SUPPLY
 	req_access = list(ACCESS_MARINE_CARGO)
-	circuit = /obj/item/circuitboard/computer/supplycomp
+	circuit = null
 	var/datum/supply_ui/SU
 	///Id of the shuttle controlled
 	var/shuttle_id = SHUTTLE_SUPPLY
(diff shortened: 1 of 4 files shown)


FIXTHERADBOXPLEASE:
Secretly a salt PR against shipside xenos sabotaging req.

Lumipharon:
Yeah atomise the pr. The circuit board is a balance change unrelated to ghost qol.

SiegeB0t:
This is not a balance change. Currently, when a xeno destroys req consoles, they can be repaired with a blowtorch. However the consoles are not repairable in the case of ff. This does not make much sense. Thus, this pr is still a fix. I'll split the pr anyway.

FIXTHERADBOXPLEASE:
Fuck I actually didn't know you could fix them with a blowtorch and I main RO. That's kind of stupid you can't fix FFed consoled then. My bad.

code/modules/reqs/supply.dm
```diff
@@ -282,7 +282,7 @@ GLOBAL_LIST_INIT(blacklisted_cargo_types, typecacheof(list(
 	. = ..()
 	if(.)
 		return
-	if(!allowed(user))
+	if(isliving(user) && !allowed(user))
 		return
```
TiviPlus, line 286: is this supposed to be a negation?

TiviPlus (changes requested):

code/modules/reqs/supply.dm
SiegeB0t: This is correct. The function should not fail when the user is dead. allowed() works on id cards so only needs to be called if the user is living.

I could write this but it seems unnecessary with the nesting:

if(isliving(user))
	if(!allowed(user))
		return

Lumipharon (approved):

PR #11744, opened 2022-12-19
Removes recoil from masterkey

Naaanii:
## About The Pull Request
Title
## Why It's Good For The Game
Normal shotguns don't have recoil when wielded and masterkey can only be fired when wielded.
Makes sense to remove the recoil.
## Changelog
:cl:
balance: removes recoil from the masterkey.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -131,7 +131,7 @@
 	fire_delay = 20 // Base shotgun fire delay.
 	pixel_shift_x = 14
 	pixel_shift_y = 18
-
+	recoil = 0
 	wield_delay_mod	= 0.2 SECONDS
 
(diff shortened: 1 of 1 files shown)


FIXTHERADBOXPLEASE:
Isn't the recoil there to make landing follow-up shots with the gun it's attached to more difficult?

Naaanii:
> Isn't the recoil there to make landing follow-up shots with the gun it's attached to more difficult?

Probably, but this was also when masterkey dealt significantly more damage.

Lumipharon:
I'm pretty ambivalent on this change, although I lean towards leaving recoil since it's a stun button on your gun, but shotguns do have recoil wielded. It's specifically the t39 that doesn't, and the LORE is due to the inbuilt stock, vs the t35 that doesn't have one.

PR #11758, opened 2022-12-21
Throwpass refactor

Lumipharon:
## About The Pull Request
Removes the throwpass var because it was bad and used for several related but not quite identical situations.
Replaced with flags for specific kinds of behavior. 

There's probably a bunch of stuff that could have some of these correctly assigned to fix slightly odd current behavior, but I'll leave that to some one else.
This also makes it easier to add some more specific object/turf behaviors. For example a wall/object that can let gas through but not thrown object/projectiles, etc.

The only in game changes are that warlock crush and flamers will bypass mechs and I think acid turrets like you'd expect them to.

NOTE: throwpass doesn't actually work for throwing, in most/all cases it's actually passtable, but again that's a pr for another day.
## Why It's Good For The Game
crush mech good.
## Changelog
:cl:
fix: Warlock crush and flamers will no longer get blocked by mechs
refactor: refactored throwpass
/:cl:

### DIFF
--- a/code/__DEFINES/equipment.dm
+++ b/code/__DEFINES/equipment.dm
@@ -7,6 +7,11 @@
 #define PASSSMALLSTRUCT (1<<5)
 #define PASSFIRE (1<<6)
 #define PASSXENO (1<<7)
+#define PASSTHROW (1<<8) //you can throw past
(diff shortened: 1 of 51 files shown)


TiviPlus (changes requested):
use NONE not null for empty bitflags

PR #11769, opened 2022-12-23
Annihilate 6 chokepoints near Ice Colony's LZ2

novaepee:
## About The Pull Request

image

image

## Why It's Good For The Game

Aesthetically, the ships are pretty garbage.

Tactically, you cannot engage them safely since you cannot blow them up.

Mapwise, the empty space looks awful, so I will have map maintainer's help to see what we can do to fill the void. I'm thinking of ice cover.

## Changelog

:cl:
del: two """""ships""""" near LZ2 that are indestructable and make 6 chokepoints.
/:cl:

### DIFF
--- a/_maps/map_files/Ice_Colony_v2/Ice_Colony_v2.dmm
+++ b/_maps/map_files/Ice_Colony_v2/Ice_Colony_v2.dmm
@@ -9247,10 +9247,6 @@
 /obj/machinery/atmospherics/pipe/simple/green/hidden,
 /turf/open/floor/tile/dark2,
 /area/ice_colony/surface/hangar/alpha)
-"aRu" = (
-/obj/structure/window/framed/mainship/escapeshuttle,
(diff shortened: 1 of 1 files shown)


FIXTHERADBOXPLEASE:
Salt PR. The (intentional) downside of taking this LZ is that it gives the Xenos a favorable approach to assault the FoB.
If you can't cope with Xenos using the terrain against you, then take the north LZ.

Derrick2Electricboogaloo:
I think the shuttles are fine. Hardly choke points when xenos use them 99.9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999% of the time and marines rarely touch that place as if it has AIDS

Al-1ce:
I think the entire area (including the dead-end deathtrap for bald xenos) should be atomized/redesigned.

novaepee:
need to think omre on this

PR #11777, opened 2022-12-24
Primo spitter tweaks

Lumipharon:
## About The Pull Request
Minor buff to primo spitter to give it a bit more area control/flavour.

Primo spitter's autofire spit now leaves acid pools on hit (like prae already does), and primo spitter now has acid blood, which is more flavour than anything, but reinforces the 'acid guy' theme.
Autospit now has the explosive flag like most spits (i.e. lands where you click).
Reduced damage fall off from 1 to 0.3 since it has almost the worst falloff in the game considering its base damage.

Also fixed a bug where acid pools on hit will bypass obstacles like cades, or entirely solid ones like glass panes etc.

And fixed a bug and reworked acid blood. Currently there is a splatter chance that... does nothing. And the splatter itself didn't respect acid armour either.
## Why It's Good For The Game
More interesting spitter primo than "shoots fast". More useful in large fight or in tight spaces.
Generally less terrible primo as it's actually incredibly bad.
## Changelog
:cl:
add: Primo spitter now leaves acid pools on hit with autospit, and has acid blood
balance: Reduced primo spitter damage falloff so it's less garbage
fix: acid pools from projectiles will no longer bypass barricades or other solid objects
fix: acid blood ignoring acid armour and splatter chance not working

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/spitter/castedatum_spitter.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/spitter/castedatum_spitter.dm
@@ -156,6 +156,7 @@
 	caste_desc = "Master of ranged combat, this xeno knows no equal."
 	upgrade = XENO_UPGRADE_FOUR
 	primordial_message = "Our suppression is unmatched! Let nothing show its head!"
+	caste_flags = CASTE_EVOLUTION_ALLOWED|CASTE_ACID_BLOOD
 
 	// *** Melee Attacks *** //
 	melee_damage = 20
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -160,19 +160,12 @@
 
 		if(!(xeno_caste.caste_flags & CASTE_ACID_BLOOD))
 			return
-		var/splash_chance = 40 //Base chance of getting splashed. Decreases with # of victims.
-		var/distance = 0 //Distance, decreases splash chance.
-		var/i = 0 //Tally up our victims.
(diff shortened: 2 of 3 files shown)


UsernameIs-UnAvailable:
I thought the acid splashing behind cades was feature. I do not think it should be removed.

Al-1ce:
Primo spitter is insanely strong as it is. Giving it area denial by acid splatter is a bad idea imo.

Lumipharon:
> I thought the acid splashing behind cades was feature. I do not think it should be removed.

It will go through entirely solid objects, like windows, so it's very buggy behavior.

>Primo spitter is insanely strong as it is. Giving it area denial by acid splatter is a bad idea imo.

Autospit is actual garbage.
It does 10 damage vs 30 for normal spit, for 33.3 dps vs 37 or something for normal spit.
Additionally, it has the same damage falloff. So at 5 tiles autospit is doing 50% damage while normal spit is doing 83% damage, so 16.7 dps vs 31.1 dps.

Autospit is only better at plasma efficiency, not damage.

Lumipharon:
Reduced that falloff actually, since it was actually extremely bad, worse than anything in the game except for masterkey buckshot.

FIXTHERADBOXPLEASE:
Spitter is another Xeno who had a lame primo.
The only two uses it had were for destroying cades, and maybe making up for bad aim due to the faster firing rate

This actually gives it some personality and notable gameplay changes, in line with the other primos.

srxtreme:
also, isn't autospit less plasma efficient than normal spit? 25 plasma for 10 damage as opposed to 50 plasma for 30 damage.

PR #11779, opened 2022-12-24
Skill check fixes

Lumipharon:
## About The Pull Request
Fixes the ability to bypass certain skill fumble times by giving yourself enough brain damage. Yes you can currently be some dumb you become able to do some things faster.

A few skill checks were written to only check if a skill value was present, which was probably fine before the brain damage rework.
After the rework, brain damage can cause your skills to go into the negative, which means you have a value for that skill, so the fumble is not performed.
## Why It's Good For The Game
Brain damage exploit bad.
## Changelog
:cl:
fix: fixed brain damage making you better at certain things
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -117,7 +117,7 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 		to_chat(user, span_warning("Can't do tactical reloads with [src]."))
 		return
 	//no tactical reload for the untrained.
-	if(!user.skills.getRating("firearms"))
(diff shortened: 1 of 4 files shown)


KerbalCitizen:
Local marine too retarded to second guess themself.

Losenis:
Ogryn meta

psykzz:
what about instead?

```dm
/datum/skills/proc/getRating(rating)
	return max(vars[rating], 0) // Negative ratings not allowed
```

Lumipharon:
> what about instead?
> 
> ```dm
> /datum/skills/proc/getRating(rating)
> 	return max(vars[rating], 0) // Negative ratings not allowed
> ```

Negative ratings are actually used though.

PR #11782, opened 2022-12-24
Mech weapon rebalances.

MLGTASTICa:
## About The Pull Request
Rebalances mech weaponary
Missile pod now gets 5 rockets instead of just 2
Flamethrower ammo increased , slowdown increase from 0.4 to 0.7
Leto autocannon ammo per mag decreased to 16 , total extra ammo decreased to 64
Briareus LMG firerate decreased from 6.6 rounds per second to 5, and slowdown increase from 0.3 to 0.6 
Moonlight cutter integrity buffed to 800, and delay decreased to 2 seconds instead of 3
Phoebe shotgun ammo mag increased to 20 , with cache increased too.
Laser rifle slowdown reduced from 0.4 to 0.3

## Why It's Good For The Game
The missile pod has only 75 damage + no devastation , meaning its not that great and its also hard to hit since mechs can barely turn, thus not beign used at all.
The flamethrower had too little ammo to be usefull at melting , and fire is not that usefull for a mech that has to go constantly refill, slowdown was increased to justify worries about constant rushing
The letoautocannon has a huge mag capacity and as thus is seen constantly spammed to break off walls or resin structures instead of the flamethrower , along with the huge damage potential of its grenades , the reduction aims to make MP's more conservative of it.
The briarus LMG had too little slowdown for being quite almost the same as the vulcan , thus its slowdown was doubled and its fire-rate reduced , its main advantage being that its more accurate 
Moonlight cutter has rarely seen use due to the super slow slash rate , and along with the fact that xenos kept spinning around it and taking off the arms , the doubled integrity and faster hit-rate should help.
Phoebe shotgun had too little ammo for it to be used on the front constantly , along with the horrendous accuracy and spread.
Laser rifle just had too much slowdown for just being a infinite-ammo hitscan rifle.
## Changelog
:cl:
balance: Missile pod now has 5 missiles 
balance: Flamethrower ammo capacity buffed , slowdown increased from 0.4 to 0.7
balance: Leto autocannon ammo mag capacity decreased from 30 to 16, and mag cache decreased from 300 to 64
balance: Briarus LMG slowdown increased from 0.3 to 0.6 , firerate decreased from 6.6 bullets per second to 5.
balance: Moonlight cutter slash rate increased from 3.3 per second to 5 per second , integrity doubled from 400 to 800
balance: Phoebe shotgun ammo capacity increased from 10 per mag to 20 per mag , and from 120 in the cache to 200
balance: Laser rifle slowdown reduced from 0.4 to 0.3
/:cl:

### DIFF
--- a/code/modules/vehicles/mecha/equipment/weapons/greyscale_weapons.dm
+++ b/code/modules/vehicles/mecha/equipment/weapons/greyscale_weapons.dm
@@ -151,12 +151,12 @@
 	)
 	ammotype = /datum/ammo/bullet/shotgun/mech
 	max_integrity = 350
-	projectiles = 10
-	projectiles_cache = 120
-	projectiles_cache_max = 120
+	projectiles = 20
+	projectiles_cache = 200
+	projectiles_cache_max = 200
 	variance = 6
 	projectile_delay = 2.0 SECONDS
-	slowdown = 0.3
+	slowdown = 0.5
 	harmful = TRUE
 	ammo_type = MECHA_AMMO_SHOTGUN
 	hud_icons = list("shotgun_buckshot", "shotgun_empty")
@@ -179,8 +179,8 @@
 	projectiles_cache = 1200
 	projectiles_cache_max = 1200
 	variance = 25
-	projectile_delay = 0.15 SECONDS
-	slowdown = 0.3
+	projectile_delay = 0.2 SECONDS
+	slowdown = 0.6
 	harmful = TRUE
 	ammo_type = MECHA_AMMO_GREY_LMG
 	hud_icons = list("rifle_heavy", "rifle_empty")
@@ -199,9 +199,9 @@
 	)
 	ammotype = /datum/ammo/tx54/mech
 	max_integrity = 400
-	projectiles = 30
-	projectiles_cache = 300
-	projectiles_cache_max = 300
+	projectiles = 16
+	projectiles_cache = 64
+	projectiles_cache_max = 64
 	variance = 20
 	projectile_delay = 0.7 SECONDS
 	slowdown = 0.4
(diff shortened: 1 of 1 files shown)


maikilangiolo:
Missile magazine change literally doesn't matter. While welcome, it's not used because you can't reload it.

Starfirejordan:
based balance PR

Mechswordsman:
Literally the only thing the autocannon is good for is breaking walls though, since it can't hit prone/crit xenos and does less damage to them per shot than their crit regen rate.  It's been nerfed enough and having a option to clear xeno walls when no one else will is nessecary. Unless you want to make the flamer do that, which it currently does not. It does basically no burn damage with the AOE, all of it's damage is on the direct hit. So it can't really burn down xeno walls like the marine flamers can using it.

Mechswordsman:
Also, the LMG is already kind of shit because it's scatter rate when your target is further then 2-3 tiles is massive. almost all of the shots fly off the mark. IF you can't both unload constantly and stay close to the target, it does no damage.

Starfirejordan:
I would nerf LMG ammo count and buff vulcan ammo count in turn

Lumipharon:
laser doesn't need buffs, LMG as far as I'm aware is already not good.
Why are you buffing various mag sizes? The need to reload is meant to make them less spam happy.

MLGTASTICa:
> laser doesn't need buffs, LMG as far as I'm aware is already not good. Why are you buffing various mag sizes? The need to reload is meant to make them less spam happy.

Because some weapons suffer from absymally small mag sizes compared to others ? a cross-flamethrower not being able to burn a lot makes it very unattractive,whilst the leto gets 240 rounds that can easily take out walls.
LMG is good as it has a super burst potential at close range
Laser rifle is not much better than the normal rifle, hence i don't see why it should be making mechs way slower.
Mechs are quite slow , and if they do have to return to FOB,  they should have some time of fighting if they bother to do a full reload.

PR #11784, opened 2022-12-25
You now spin while stop drop rolling.

carlarctg:
## About The Pull Request

You now spin while stop drop rolling. 3 seconds of duration, same as paralyze, 1.5 abstract time value for time between direction spins.

## Why It's Good For The Game

Presumably this should pop an error out considering it's a signal handler and there's sleeps in spin(), but there weren't any so who knows. I'll invoke_async if told to.

## Changelog

:cl:
add: You now spin while stop drop rolling.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -201,7 +201,7 @@
 	SIGNAL_HANDLER
 	fire_stacks = max(fire_stacks - rand(3, 6), 0)
 	Paralyze(30)
-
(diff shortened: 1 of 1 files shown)


Lumipharon:
If you get up early, won't you keep spinning?

FosterGrime-Heymynameslouie:
> If you get up early, won't you keep spinning?

this should definitely be kept it if it does because it's very funny

carlarctg:
> 

yes and it's awesome

PR #11785, opened 2022-12-25
Primo rav fix and adjustment

Lumipharon:
## About The Pull Request
Fixed a bug where vampirism would separetely heal both brute and burn for 25% of total health lost, effectively meaning the heal could be 200+ health.

Reduced the heal amount to 12.5% from 25%.

25% heal every 2 seconds is simply not balanced, even without the bug, especially taking into account endure and (super) rage.

I'm not entirely happy with the vampirism mechanic because it's essentially passive, but the previous one was trash.
## Why It's Good For The Game
Bug fix good.
Having a queen healing you every 2 seconds bad.
## Changelog
:cl:
balance: Reduced Ravager vampirism healing by 50%.
fix: fixed Ravager vampirism healing more than it should
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
@@ -511,8 +511,8 @@
 	if(timeleft(timer_ref) > 0)
 		return
 	var/mob/living/carbon/xenomorph/x = owner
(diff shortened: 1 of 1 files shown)


maikilangiolo:
Should still make the cooldown higher. It's not like rav isn't good and primo rav doesn't have other abilities. Vampirism should be a minor help, not a major mechanic.

PR #11800, opened 2022-12-27
Grenade launcher tweak

Lumipharon:
## About The Pull Request
Adjusted sticky nade det time when fired from a GL from 1 second to 3-5 seconds.

Currently, all nades fired from a GL have their det time hard set to 1 second, giving a (very) small window to try avoid the blast.
Stickies are a bit difficult because they, well, stick, making them a guaranteed wombo combo tool.

This makes more of a trade off between normal grenades with a larger radius and shorter det time for that immediate staggerslow, and stickies with their guaranteed hit (if you hit initially) but longer det time.

This can also be used to individually tweak det times for specific grenades, if required.
## Why It's Good For The Game
Make stickies a bit less unpleasant to deal with.
## Changelog
:cl:
balance: Sticky grenades detonate after four seconds instead of one second, when fired from a grenade launcher
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -93,6 +93,9 @@
 	walk(src, null, null)
 	return
 
+///Returns an adjusted det time, used for grenade launchers
+/obj/item/explosive/grenade/proc/launched_det_time()
+	return min(10, det_time)
 
 ////RAD GRENADE - TOTALLY RAD MAN
 
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -93,9 +93,9 @@
 	walk(src, null, null)
 	return
 
-///Returns an adjusted det time, used for grenade launchers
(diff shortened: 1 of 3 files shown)


novaepee:
It was something that I observed and neglected after witnessing it in combat, to be honest. the GL bypasses the initial design that sticky nades should take a few seconds to denonate rather than a second. I didn't expect sticky nades to be meta until as of lately.

my bad

vilereaver41:
eat S$#$ nade spammers who abuse easy methods.

Naaanii:
Isn't the point of the nade launcher to reduce det timer?

Would be better to increase the rof imo so the first nade is more impactful but it's harder to stunlock/spam subsequent nades.

Strayed-Raven:
you might as well just bring back impact nades lol

FosterGrime-Heymynameslouie:
I dont really know about the balance implication of this but I find it highly funny now that runners who know who are dead anyway to have an option within those 3 seconds of instead of running away, charging straight at the marine horde and blowing them up with their own stick nades 10/10 while jihad anytime of the year

PR #11811, opened 2022-12-28
HvH E-shields

Lumipharon:
## About The Pull Request
Bit of an experiment to see how it changes the game.

Adds eshields to all default loadouts (i.e. loadouts that don't already have another module), excluding engies and medics.

This should add a level of survivability, and especially help against plinking hits/surprise attacks.

Also removes the weird -5 bullet armour. There was no negative for any other human specific armour types so it never really worked, and 'you get FF'd worse' is a pretty meme idea anyway.
## Why It's Good For The Game
More survivability in HvH
## Changelog
:cl:
balance: eshields no longer have -5 bullet armour
balance: HvH: eshields added as default modules for roles other than engie and medic, where another module is not present

/:cl:

### DIFF
--- a/code/datums/quick_load_outfits.dm
+++ b/code/datums/quick_load_outfits.dm
@@ -103,7 +103,7 @@
 	ears = /obj/item/radio/headset/mainship/marine
 	w_uniform = /obj/item/clothing/under/marine/black_vest
 	shoes = /obj/item/clothing/shoes/marine/full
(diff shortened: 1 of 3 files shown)


FIXTHERADBOXPLEASE:
Stealth salt PR to buff the halo shield against friendly fire. Nice try but it's already one of the strongest choices in the game.

PR #11815, opened 2022-12-29
EVERYTHING in req now costs ten times more, everything that gives POINTS now gives ten times more.

Pariah919:
## About The Pull Request
Everything that gives req points and costs req points now costs an extra zero, so about ten times more for both results, ideally this results in no noticeable change for now.
## Why It's Good For The Game
Point brackets are fucked. An AMR is literally 15 points more than a goddamn Autosniper, something needs to change, and this is a good start, because we really do need more granularity between point brackets, especially once we get to the lower level of points for things like ammunition and the 10-40 point bracket.
## Changelog
:cl:
tweak: Everything in Req costs ten times more, this changes nothing. As everything gives ten times more the Req point gain.
/:cl:


Losenis:
Bossman I don't think 10 times 2.5 is 20.5
image

Pariah919:
> Bossman I don't think 10 times 2.5 is 20.5 image
> 
> Also where's the changes to research's credits?

i had a c in math

novaepee:
image

truly stuff costs more with cash

Losenis:
Might also want to rename the actual item itself for consistency.
image

Pariah919:
> Might also want to rename the actual item itself for consistency. image

i mean its still 5 credits
tgmc prices are just inflated lets be real here

Losenis:
Dude I'm from argentina I had enough of inflation for ~~a~~ ten lifetimes.

Lumipharon (approved):

Lumipharon:
PR needs a more specific reason though, your example about amr and autosniper isn't really relevant since you do actually have points to play with between them.

However the very cheap stuff is where this granularity is useful (i.e. ammo)

FosterGrime-Heymynameslouie:
I cant wait to do math

lbnesquik:
image

Derrick2Electricboogaloo:
Holy hell inflation hit TGMC too? Only in Joe Biden's America

Derrick2Electricboogaloo:
Wait does that means B18 is now 1200 points?!

bowlofcereal:
https://www.youtube.com/watch?v=61Q6wWu5ziY&ab_channel=Bizonacci

567Turtle:
But... But why?

Alexeivichi:
Have you seen my economy? He is in shambles.

Tupinambis:
10x the cost seems a bit overkill and might lead to kind of ridiculous values (1200 B18 like, wew)

5x might be a bit better and keep things out of the quadruple digits.

Pariah919:
>10x the cost seems a bit overkill and might lead to kind of ridiculous values (1200 B18 like, wew)

5x might be a bit better and keep things out of the quadruple digits.
>
10x is literally just a 0
x5 will lead to decimal values in some of these

567Turtle:
Wait what? You can't keep decimals from multiplying numbers without decimals. So unless there's some decimal valued req item I don't know about x5ing it would be fine.

Pariah919:
> Wait what? You can't get decimals from multiplying numbers without decimals. So unless there's some decimal valued req item I don't know about x5ing it would be fine.

you know i posted that and then realized it was dumb
I'm not doing five times anyway as it only results in different (if smaller) numbers when the entire point of this is to make the same numbers but bigger to allow for more brackets, thus being pointless

MiniMeatwad:
Cool, but now you should display it as decimal because this is how decimals are actually added to integer numbers in coding. But I don't know what Byond does, so maybe there's a specific solution.

PR #11822, opened 2022-12-29
SL demotion/promotion fix

Lumipharon:
## About The Pull Request
SL's no longer lose leadership skill from the trauma of demotion.
## Why It's Good For The Game
Bug fix good
## Changelog
:cl:
fix: SL's no longer lose leadership skill from the trauma of demotion
/:cl:


QualityVan:
Demotion should actually remove the entire skill datum of the mob in question in order to be more immersive.

PR #11826, opened 2022-12-30
Re-adds Laying eggs and psychic whisper to the Queen.

Tupinambis:
## About The Pull Request

Simply gives the queen to ability to lay eggs and use psychic whisper again.

## Why It's Good For The Game

Laying eggs was removed in early 2021 for being useless which I don't agree with. They work fine as traps and as a way of drawing marine fire, and they can be very helpful for the carriers and defilers in your life. Also they are thematically appropriate and give queen more to do.

Queen also used to have psychic whisper but it was removed at some point. It seems pretty silly for the leader role to not have it when Warlock, Shrike, and Hivemind do. It's a little bit of extra flavor for queen and it can be fun to mock the ungas.

Queen at the moment is arguably the weakest caste in terms of gameplay, so hopefully this will make her slightly less painful to play. She definitely deserves a rework at some point, but those take time and skill and this takes two lines of code.

## Changelog

:cl: Tupina
add: Queen can once again lay eggs.
add: Queen can once again use psychic whisper to annoy marines.
/:cl:


FIXTHERADBOXPLEASE:
The only argument against having these was "toolbar bloat," but any sane xenomorph uses hotkeys anyway. Especially a Queen.

MiniMeatwad:
Yes. Now I can play Queen, and have a good time being distracted by watching marines walk into eggs.

PR #11830, opened 2022-12-30
Makes Dylovene's stamina debuff only trigger at >5u

Losenis:
## About The Pull Request
Dylovene's stamina debuff now only triggers at >5u rather than at any volume of the reagent in the body.

## Why It's Good For The Game
The stamina debuff change didn't account for combat injectors having trace amounts of dylovene, so using a combat injector cucked you off your stamina despite being something to use in combat. This will make combat injectors safe to use while still making dylovene pills cause stamina debuff.

## Changelog

:cl:
balance: Dylovene's stamina debuff now only happens at >5u of reagents in the body.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -352,11 +352,6 @@
 	overdose_crit_threshold = REAGENTS_OVERDOSE_CRITICAL
 	taste_description = "a roll of gauze"
 
-/datum/reagent/medicine/dylovene/on_mob_add(mob/living/L, metabolism)
-	if(volume > 5)
-		L.add_stamina_regen_modifier(name, -0.5)
-	return ..()
-
 /datum/reagent/medicine/dylovene/on_mob_delete(mob/living/L, metabolism)
 	L.remove_stamina_regen_modifier(name)
 	return ..()
@@ -364,6 +359,8 @@
 /datum/reagent/medicine/dylovene/on_mob_life(mob/living/L,metabolism)
 	L.hallucination = max(0, L.hallucination -  2.5*effect_str)
 	L.adjustToxLoss(-effect_str)
+	if(volume > 5)
+		L.add_stamina_regen_modifier(name, -0.5)
 	return ..()
 
 /datum/reagent/medicine/dylovene/overdose_process(mob/living/L, metabolism)
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -353,7 +353,8 @@
 	taste_description = "a roll of gauze"
 
 /datum/reagent/medicine/dylovene/on_mob_add(mob/living/L, metabolism)
-	L.add_stamina_regen_modifier(name, -0.5)
+	if(volume > 5)
+			L.add_stamina_regen_modifier(name, -0.5)
 	return ..()
 
 /datum/reagent/medicine/dylovene/on_mob_delete(mob/living/L, metabolism)
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -354,7 +354,7 @@
 
 /datum/reagent/medicine/dylovene/on_mob_add(mob/living/L, metabolism)
 	if(volume > 5)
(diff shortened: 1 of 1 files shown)


srxtreme:
Powercreep/powergame PR. This effectively kills any purpose to dylovene's stamina penalty (as dylovene metabolizes extremely slowly), and allows marines to ridiculously easily purge xeno toxins with no downsides. All other methods (ryetalyn, water, hypervene) have downsides. 5u of dylovene can fix up to 35 units of neurotoxin - which is more than what boiler neurotoxin lance injects. This will significantly unsettle the chemistry meta and may lead to unforseen consequences in the future. Furthermore - this renders hyronalin obsolete and encourages chemistry powergaming (chem man bad!). Overall a drastically negative contribution to the codebase and a step back instead of step forward. Also relatively undoes what #10538 does.

Redrover1760:
Have you considered changing what is in combat autoinjectors instead of changing Dylo so that its downside can be easily avoided?

Strayed-Raven:
Xenos thinking this is a power buff PR: 😭 
Me, who never brings anything for neuro and has never died to neuro unless its a screech neuro combo: 🗿

QualityVan:
Given dylo has no threshold benefits and a low metab rate, and the only purgative xenos still have has a high purge rate, why would you ever dose dylo above 5u if this change was in?

code/modules/reagents/reagents/medical.dm
```diff
@@ -353,7 +353,8 @@
 	taste_description = "a roll of gauze"
 
 /datum/reagent/medicine/dylovene/on_mob_add(mob/living/L, metabolism)
-	L.add_stamina_regen_modifier(name, -0.5)
+	if(volume > 5)
```
QualityVan, line 356: This won't trigger if you take 3u and then take another 3u.

Losenis:
How come? I'm taking it from the other reagents so wouldn't taking a total of 6 start the debuff till it's under 5?

image

QualityVan:
Because dylo triggers the modification on add, which is only called when a new reagent arrives in the mob. The bic in your image checks it on metabolize instead.

Losenis:
>Given dylo has no threshold benefits and a low metab rate, and the only purgative xenos still have has a high purge rate, why would you ever dose dylo above 5u if this change was in?

Dylov pill bottles still have 15u, and they're more space efficient than injectors on a usage basis since it's still 16 uses per bottle. We can argue about someone gaming it by going to chem and making a pill bottle of 5u dylov pills but that'd probably fall more on the chem mixing aspect than the reagent itself.

Dylov's debuff puts it in a weird spot in that it gives combat injectors a debuff but at the same time due to having the debuff at all volumes you're just better off packing a bottle than carrying injectors or doing chem mixing.

Losenis:
Updated it, is that how it should go then?

code/modules/reagents/reagents/medical.dm
```diff
@@ -352,17 +352,15 @@
...
-
 /datum/reagent/medicine/dylovene/on_mob_delete(mob/living/L, metabolism)
 	L.remove_stamina_regen_modifier(name)
 	return ..()
 
 /datum/reagent/medicine/dylovene/on_mob_life(mob/living/L,metabolism)
 	L.hallucination = max(0, L.hallucination -  2.5*effect_str)
 	L.adjustToxLoss(-effect_str)
+	if(volume > 5)
+		L.add_stamina_regen_modifier(name, -0.5)
```
QualityVan, line 363: You're now adding a modifier every single tick it's over 5u and only removing it after the reagent leaves the body entirely.

Losenis:
fuck

QualityVan:
> Dylov pill bottles still have 15u, and they're more space efficient than injectors on a usage basis since it's still 16 uses per bottle. We can argue about someone gaming it by going to chem and making a pill bottle of 5u dylov pills but that'd probably fall more on the chem mixing aspect than the reagent itself.

So you bring a pill bottle and a hypospray set to 5u(or just a flask), and it's still strictly a mistake to ever use the supplies you're issued the way they're issued.

Losenis:
Fuck it this is beyond my ability

PR #11845, opened 2023-01-01
feat: use vis flag to determine direction

psykzz:
## About The Pull Request

Instead of using a signal to set the dir of the wounds overlay, we can just use the vis_flags of inherit dir from parent.
Additionally using the vis_hide flag for wounds when they have nothing to show.

This helps remove them from the right click menu (but they will return when xenos are hurt).

## Why It's Good For The Game
Just a cleanup

## Changelog
:cl:
code: removed an extraneous signal
refactor: hide xeno wounds when they have nothing to show
/:cl:


TiviPlus (approved):

TiviPlus:
You might also be able to remove the wound_owner var I dont think I used it anywhere else

psykzz (commented):
Remove wound_owner

PR #11849, opened 2023-01-02
Reflavors brute packs and ointments to be TG-based

carlarctg:
## About The Pull Request

Brute packs are now rolls of medical gauze.

Ointment is now regenerative mesh. (Needs to be ripped open to use)

Ported modern CM splint sprites.

Everything is identical stat-wise.

Repathed ointment.

## Why It's Good For The Game

Brute packs and ointments feel like ancient baystation12 leftover while medical gauze and mesh feels cool and tg and modern.

In my opinion the splint sprites are a bit more out of place looking than the current CM ones.

## Changelog

:cl:
imageadd: Brute packs are now rolls of medical gauze.
imageadd: Ointment is now regenerative mesh. (Needs to be ripped open to use)
imageadd: Ported modern CM splint sprites.
add: Everything is identical stat-wise.
code: Repathed ointment.
/:cl:


novaepee:
For people that want to see the sprites

image

Losenis:
>Ointment is now regenerative mesh. (Needs to be ripped open to use)

Doesn't this just add busywork or ripping open every mesh you have as soon as you grab it off the vendor or loadout?

Al-1ce:
I believe they were resprited to look like the healing kits from Metal Gear. In either case I'm really not a fan of this change. I think the sprites from tg look worse by a very large margin.

Xander3359:
Not a big fan, but also, if you are to replace burn with regen mesh, then brute should be replaced with sutures, not gauze.

Derrick2Electricboogaloo:
I would rather not. I much prefer the ointment and gauze we have over TG-base. And their sprites were fine.

Xsdew:
Given that Ointment and Roll of Gauze function differently to Regenerative Mesh and Sutures. It may be best to keep the sprites separate. This is purely to avoid players from assuming they function like TGMed.

Kuro020:
Cool and tg dont fit on the same comment, also these sprites are worse than the current ones and the gauze looks like the bandages which is confusing

Derrick2Electricboogaloo:
> Cool and tg dont fit on the same comment, also these sprites are worse than the current ones and the gauze looks like the bandages which is confusing

That is how the gauze functions, should just rename it bandages.

Starfirejordan:
why was this killed

PR #11856, opened 2023-01-03
Refactor gun modifier calculations and some related stuff

Lumipharon:
## About The Pull Request
- Refactors gun accuracy and scatter calculations. There's a large amount of calculation done currently every time a projectile is fired. A large chunk of this has now been pushed out to the gun itself. When relevant factors change, such as the gun user changing their pain threshold, or wielding the gun etc, this calculation is reran. Spooky says this should be better performance wise, if I haven't made a mess of it.
- Acc and scatter status effects now directly effect mob acc and scatter modifiers, instead of snowflake checks in gun code.
- Fixed a couple of minor gun bugs/removed some unneeded code/procs.
- Burst accuracy multipler works the way it should
- A gun being fully wielded is now set by a flag, instead of a bizarre proc check everywhere.
- Fixed several effects such as stagger/shock not properly taking godmode into account (often god mode would stop any change, so if you went into god mode staggered, you'd never recover).

As part of the status effect changes, Offguard now applies for 10 seconds, but the strength of the effect does not decrease over time.
## Why It's Good For The Game
Better performance (probably)
Fixes good
## Changelog
:cl:
fix: fixed various minor gun and status bugs
refactor: refactored accuracy and scatter calculations for projectiles
balance: Offguard duration changed to 10 seconds, effect no longer decreases over time
/:cl:


FosterGrime-Heymynameslouie:
stimulate your sense

code/game/objects/items/stimulants.dm
```diff
@@ -0,0 +1,46 @@
...
+/obj/item/stimulant/drop
+	name = "drop booster"
+	desc = "Drop pushes the user into a heightened state of neural activity, greatly improving hand eye co-ordination but making them more sensitive to pain. Also known to cause severe paranoia and hallucinations, at higher doses."
+	icon_state = "drop"
+	stim_type = STATUS_EFFECT_STIMULANT_DROP
+	stim_message = "You can suddenly feel everything!"
+
+/obj/item/stimulant/exile
+	name = "exile booster"
+	desc = "Exile inhibits several key receptors in the brain, triggering a state of extreme aggression and dumbness to pain, allowing the user to continue operating with the most greivous of injuries. Exile does not actually prevent any damage however, and can gradually lead to neural degeneration."
```
carlarctg, line 36:
```suggestion
	desc = "Exile inhibits several key receptors in the brain, triggering a state of extreme aggression and numbness to pain, allowing the user to continue operating with the most grievous of injuries. Exile does not actually prevent any damage however, and can gradually lead to neural degeneration."
```

zee11zee:
TGMC Tarkov?

Serxule:
drop sounds pretty much useless, exile sounds extremely strong, and crash is eh

Al-1ce:
The names of these stims seem kind of scuffed but overall +1 from me.

MereWalker:
A can't wait to become a German soldier high on meth for this pr let's go!

Coastfront:
Drop seems neigh worthless, Exile seems absurd, but crash just looks like it has no downside.
If its downside only triggers from sprinting too often you can just codedive to see the heart damage threshold and 99% it all the time. Depending on by how much speed is increased you might still cover the same if not similar distances as you would a non enhanced full stamina bar, and then just be left with higher stam regen.

code/datums/status_effects/gun_skill.dm
```diff
@@ -1,46 +1,75 @@
-/datum/status_effect/stacking/gun_skill
```
TiviPlus, line 1: Why change it from stacking ??

code/datums/status_effects/gun_skill.dm
```diff
@@ -1,46 +1,75 @@
...
-
-/datum/status_effect/stacking/gun_skill/Destroy()
+/datum/status_effect/gun_skill/Destroy()
 	if(particles)
 		QDEL_NULL(particles)
 	return ..()
 
+//Base accuracy effect
+/datum/status_effect/gun_skill/accuracy
+	var/accuracy_modifier = 0
```
TiviPlus, line 19: doc

code/datums/status_effects/gun_skill.dm
```diff
@@ -1,46 +1,75 @@
...
-/datum/status_effect/stacking/gun_skill/accuracy/debuff
+/datum/status_effect/gun_skill/accuracy/debuff
 	id = "gun_skill_accuracy_debuff"
+	accuracy_modifier = -25
 	particles = /particles/gun_skill/accuracy/debuff
 
-/datum/status_effect/stacking/gun_skill/scatter/buff
+//Base scatter effect
+/datum/status_effect/gun_skill/scatter
+	var/scatter_modifier = 0
```
TiviPlus, line 46: more doc

code/modules/mob/living/carbon/human/life/handle_regular_hud_updates.dm
```diff
@@ -7,6 +7,26 @@
...
+			hud_used.nutrition_icon.icon_state = "nutrition1"
+		if(hud_used.toxin_icon)
+			hud_used.toxin_icon.icon_state = "tox0"
+		if(hud_used.oxygen_icon)
+			hud_used.oxygen_icon.icon_state = "oxy0"
+		if(hud_used.fire_icon)
+			hud_used.fire_icon.icon_state = "fire0"
+		if(hud_used.bodytemp_icon)
+			hud_used.bodytemp_icon.icon_state = "temp0"
+		return
```
TiviPlus, line 28: ?

code/modules/mob/living/living.dm
```diff
@@ -535,9 +535,15 @@
...
 
+///Modifies the mobs inherent accuracy modifier
 /mob/living/proc/adjust_mob_accuracy(accuracy_mod)
 	ranged_accuracy_mod += accuracy_mod
+	SEND_SIGNAL(src, COMSIG_RANGED_ACCURACY_MOD_CHANGED, ranged_accuracy_mod)
 
+///Modifies the mobs inherent scatter modifier
+/mob/living/proc/adjust_mob_scatter(scatter_mod)
+	ranged_scatter_mod += scatter_mod
+	SEND_SIGNAL(src, COMSIG_RANGED_SCATTER_MOD_CHANGED, ranged_scatter_mod)
```
TiviPlus, line 546: pass amount changed not the var people can fetch that off the source ref

code/modules/mob/living/status_procs.dm
```diff
@@ -637,9 +637,17 @@
 		stagger = clamp(stagger + amount, 0, capped)
 		return stagger
 
-	stagger = max(stagger + amount,0)
+	set_stagger(max(stagger + amount,0))
 	return stagger
 
+/mob/living/proc/set_stagger(amount)
```
TiviPlus, line 643: doc

code/modules/mob/mob_helpers.dm
```diff
@@ -471,3 +471,8 @@ GLOBAL_LIST_INIT(organ_rel_size, list(
 	var/obj/item/inactive_item = get_inactive_held_item()
 	if(istype(inactive_item))
 		inactive_item.do_unique_action(src)
+
+///Handles setting or changing a mob's skills
+/mob/proc/set_skills(datum/skills/new_skillset)
+	skills = new_skillset
+	SEND_SIGNAL(src, COMSIG_MOB_SKILLS_CHANGED, skills)
```
TiviPlus, line 478: skimming this looks like it has too much boilerplate for being a setter

code/modules/projectiles/gun_system.dm
```diff
@@ -616,13 +650,22 @@
...
 	return TRUE
 
+/obj/item/weapon/gun/toggle_wielded(user, new_value)
+	switch(new_value)
+		if(null)
+			flags_item ^= WIELDED
+		if(FALSE)
+			flags_item &= ~(WIELDED|FULLY_WIELDED)
+		if(TRUE)
+			flags_item |= WIELDED
```
TiviPlus, line 668: why is this using bools and null while not treating null as 0

TiviPlus (changes requested):
This shouldve been split ill need to do multiple passes

code/datums/status_effects/gun_skill.dm
Lumipharon: done

code/datums/status_effects/gun_skill.dm
Lumipharon: done

code/datums/status_effects/gun_skill.dm
Lumipharon: Stacking is cringe because the effect is based on the number of stacks remaining, meaning it does almost nothing at low stacks. It also makes it harder to balance as high stacks make it impossible to hit a barn door.

code/modules/mob/living/carbon/human/life/handle_regular_hud_updates.dm
Lumipharon: what the comment says. Being fully immune to pain means you literally can't tell that you're hurt.

code/modules/mob/living/living.dm
Lumipharon: done.

code/modules/mob/living/status_procs.dm
Lumipharon: done

code/modules/projectiles/gun_system.dm
Lumipharon: Mentioned in discord, the null if from the base proc. As far as I understand it is unused and doesn't do anything, but I didn't want to delete it if I was missing something.

code/modules/mob/mob_helpers.dm
Lumipharon: There's 3 different skill procs to consider, so I'm not sure how I would put their functionality into set_skills without making it guh.

code/datums/status_effects/gun_skill.dm
```diff
@@ -1,46 +1,77 @@
-/datum/status_effect/stacking/gun_skill
+/datum/status_effect/gun_skill
 	id = "gun_skill"
```
TiviPlus, line 2: why remove the stacking btw this was used to make kings scatter debuff decay

code/datums/status_effects/stimulants.dm
```diff
@@ -0,0 +1,180 @@
...
+	human_owner.adjust_mob_scatter(5)
+
+	human_owner.clear_fullscreen("drop")
+	owner.balloon_alert(owner, "You feel the world turn grey")
+
+/datum/status_effect/stimulant/drop/tick()
+	if(!ishuman(owner))
+		return FALSE
+	if(owner.stat == DEAD)
+		return FALSE
```
TiviPlus, line 41: are you sure you dont just want this as a chem

code/datums/status_effects/stimulants.dm
```diff
@@ -0,0 +1,180 @@
...
+
+	human_owner.overlay_fullscreen("exile", /atom/movable/screen/fullscreen/bloodlust)
+
+	return TRUE
+
+/datum/status_effect/stimulant/exile/on_remove()
+	if(!ishuman(owner))
+		return FALSE
+	var/mob/living/carbon/human/human_owner = owner
+	REMOVE_TRAIT(human_owner, TRAIT_PAIN_IMMUNE, "exile stimulant")
```
TiviPlus, line 83: missing defines

TiviPlus (changes requested):

code/datums/status_effects/gun_skill.dm
Lumipharon: Yeah I didn't like the decay. It's visually very unclear how badly debuffed you are, and once it gets under a certain number of stacks it functionally does nothing. I was probs going to reduce the total effect from offguard in a followup pr though.

Lumipharon:
Gonna probs atomise this while I think about whether this should be a chem or not.

PR #11866, opened 2023-01-04
HvH spawn protection

Lumipharon:
## About The Pull Request
Adds spawn protection when you deploy to the ground in Combat Patrol and Sensor Capture. For 10 seconds you have godmode. This may need to be tweaked if you can sprint to anything important in that time.

I got tired of watching ungas mindlessly deploying 1 by 1 into a Denholm spawn camp they know about, rather than using their brain (and map table) and just deploying to the 2nd spawn point.

As I can't cure ungabrain, I'll have to settle with this.
## Why It's Good For The Game
Removes some salt from people being dumb.
Also spawn camping is cringe.
## Changelog
:cl:
balance: HvH: You have 10 seconds of spawn protection when deploying groundside
/:cl:

### DIFF
--- a/code/game/objects/structures/patrol_points.dm
+++ b/code/game/objects/structures/patrol_points.dm
@@ -48,6 +48,7 @@
 	user.visible_message(span_notice("[user] goes through the [src]."),
 	span_notice("You walk through the [src]."))
 	user.trainteleport(linked_point.loc)
+	add_spawn_protection(user)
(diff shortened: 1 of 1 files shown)


Joe13413:
late as hell but i think it'd be sick as hell to add a sort of turret that you can access shipside to shoot at people spawncamping you instead of godmode.
just a shot in the dark here but sounds like a cool idea

PR #11911, opened 2023-01-11
SMG-25 added to marine vendors, now fits UGL

Pariah919:
## About The Pull Request
SMG-25 added to marine vendors, it now fits UGL
## Why It's Good For The Game
~~Gun bloat~~
Gives a interesting gun option, and is already used by marines in HvH, I don't think it shakes up balance too hard, as it isn't the B2 version..
## Changelog
:cl:
add: Added SMG-25 to marine vendors.
balance: SMG-25 now fits UGL.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -35,6 +35,8 @@
 			/obj/item/ammo_magazine/smg/standard_smg = -1,
 			/obj/item/weapon/gun/smg/standard_machinepistol = -1,
 			/obj/item/ammo_magazine/smg/standard_machinepistol = -1,
+			/obj/item/weapon/gun/smg/m25 = -1,
+			/obj/item/ammo_magazine/smg/m25 = -1,
 		),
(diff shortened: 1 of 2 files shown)


Derrick2Electricboogaloo:
So we are going the half life route?

Derrick2Electricboogaloo:
Also this is already in seasonals and able to be req bought

Derrick2Electricboogaloo:
Make sure to add SMG-25 belt/holster

Pariah919:
> Also this is already in seasonals and able to be req bought

wait what
well guess then this was pointless

Derrick2Electricboogaloo:
> 

Yep. only 5 points... Err.. 50 now probably. But still cheap and easy to get.

FIXTHERADBOXPLEASE:
Could still be added IMO. There's like five different rifles and only two marine SMGs, one of which is actually a machine pistol.

FIXTHERADBOXPLEASE:
More like the MP19 had its niche changed into an actual machine pistol, thus creating a missing slot for SMGs.

PR #11919, opened 2023-01-12
sunsets mp lobby mexican standoffs

Aporhtonoma:
## About The Pull Request
Makes mech accesible roundstart
## Why It's Good For The Game
mech was locked behind 25 pop because balance was out of wack and people really did not want to face it at low pop
mech has recieved significant balance changes since then and currently a number of players are essentially spawncamping the slot leading to them spamming join to grab the slot .Which is both not fun and discourages others from trying it out in the first place.This aims to make mech available roundstart and changes new slots to be available every 65 players

if this gets denied my only hope is that someone adresses it better than me 
## Changelog
:cl:
balance: changed mp slots to be available roundstart ,second at 65, from 25/65 to 0/65
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -363,7 +363,7 @@ Though you are a warrant officer, your authority is limited to the dropship and 
 	req_admin_notify = TRUE
 	paygrade = "E3"
 	comm_title = "MCH"
-	total_positions = 0
+	total_positions = 1
 	skills_type = /datum/skills/mech_pilot
 	access = list(ACCESS_MARINE_WO, ACCESS_MARINE_PREP, ACCESS_MARINE_MECH, ACCESS_CIVILIAN_PUBLIC)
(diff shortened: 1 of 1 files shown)


Starfirejordan:
Good PR, now all we need it mech weapon changes PR to be revived to better balance the weapon between themselves and where golden

Lumipharon (changes requested):
Mech is exceptionally unfun to play against at lowpop, this hasn't changed. The pop requirement was added for a reason.

I don't know if the solution is something meme like having an MP but they can't actually create a mech until the pop threshold is reached, but this is 100% not the right solution.

SilencedMP5A5:
perhaps with this change i will finally be able to try mech

Aporhtonoma:
> Mech is exceptionally unfun to play against at lowpop, this hasn't changed. The pop requirement was added for a reason.
> 
> I don't know if the solution is something meme like having an MP but they can't actually create a mech until the pop threshold is reached, but this is 100% not the right solution.

yea you are right ill sunset this pr for now and reopen it if I get a better solution (better than this half assed one at least) going unless someone else fixes it first  (or i forget)

PR #11920, opened 2023-01-12
Barrel charger improvement

Lumipharon:
## About The Pull Request
Barrel chargers now slightly reduce projectile damage falloff.

Barrel chargers are kinda useless currently. They increase projectile speed but reduce accuracy, where as an EB increases projectile speed, accuracy and scatter. This should give it some niche benefit for longer range shooting, especially with lower damage projectiles.
## Why It's Good For The Game
BC has a very small niche instead being basically trash.
## Changelog
:cl:
balance: Barrel chargers reduce projectile damage falloff slightly
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -422,11 +422,12 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 
 /obj/item/attachable/heavy_barrel
 	name = "barrel charger"
-	desc = "A fitted barrel extender that goes on the muzzle, with a small shaped charge that propels a bullet much faster.\nGreatly increases projectile speed."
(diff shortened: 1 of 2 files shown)


Strayed-Raven:
isn't damage falloff already negligible?

Losenis:
Not really, some guns that can fit barrel charger have higher than average damage falloff. T18 already has like 0.8 and T90 has 1, to say nothing of seasonals like the ALF.

Lumipharon:
Damage falloff is also linear. Falloff of 1 means 1 damage lost per tile travelled. So for a bullet that does 20 damage instead of 40, the same damage falloff is x2 the effect percentage wise.

QualityVan:
blasma cutter

TiviPlus (approved):

lbnesquik:
But, doesn't that still kinda suck? Damage falloff reduction is nice, but you still are less accurate than with an extended.

PR #11927, opened 2023-01-13
CAS can no longer be launched before shutters open, or before Tadpole launches

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
I am getting so tired of the classic PO griefing the round for whatever reason.

## Changelog
:cl: Lewdcifer
fix: CAS is no longer usable before shutters open, or before Tadpole launches.
/:cl:

### DIFF
--- a/code/modules/shuttle/cas_plane.dm
+++ b/code/modules/shuttle/cas_plane.dm
@@ -50,6 +50,7 @@
 	cas_usable++
 	if(occupant)
 		to_chat(occupant, span_notice("Combat initiated, CAS now available."))
+	UnregisterSignal(SSdcs, list(COMSIG_GLOB_OPEN_TIMED_SHUTTERS_LATE, COMSIG_GLOB_OPEN_TIMED_SHUTTERS_XENO_HIVEMIND, COMSIG_GLOB_OPEN_TIMED_SHUTTERS_CRASH, COMSIG_GLOB_OPEN_SHUTTERS_EARLY))
 
 ///Handles updating the cockpit overlay
 /obj/structure/caspart/caschair/proc/set_cockpit_overlay(new_state)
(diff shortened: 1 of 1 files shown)


Naaanii:
Speedmerge please

maikilangiolo:
You forget POs and CMOs that grief by launching tadpole before shutters, too

lKiseki:
> You forget POs and CMOs that grief by launching tadpole before shutters, too

I am told Tadpole already has a time lock.

novaepee:
after 3 years of dealing with this
finally, the pr is here

code/modules/shuttle/cas_plane.dm
```diff
@@ -42,6 +45,13 @@
 	if(occupant)
 		to_chat(occupant, span_notice("CAS laser detected. Target: [AREACOORD_NO_Z(incoming_laser)]"))
 
+/obj/structure/caspart/caschair/proc/cas_usable(datum/source)
+	SIGNAL_HANDLER
+	cas_usable++
```
Lumipharon, line 50: ? Just make it true.
Also are you sure all these signals are needed? The one labeled crash for example, is that not actually related to the crash gamemode?

code/modules/shuttle/cas_plane.dm
```diff
@@ -21,11 +21,14 @@
 	var/mob/living/carbon/human/occupant
 	///Animated cockpit /image overlay, 96x96
 	var/image/cockpit
+	/// Whether CAS is usable or not.
+	var/cas_usable = 0
```
Lumipharon, line 25: This doesn't need to be defined, since that makes it null anyway. Otherwise it can be FALSE

Lumipharon (changes requested):

code/modules/shuttle/cas_plane.dm
lKiseki: Gotcha.

code/modules/shuttle/cas_plane.dm
lKiseki: Made it true.
Also, you'd be right on the crash signal. Every other signal is necessary, however.

PR #11887, opened 2023-01-06
Removes Optical Imager Goggles from the game

lKiseki:
## About The Pull Request
Per title. Removes Optical Imager Goggles entirely from the code, removing them from the game by extension.

## Why It's Good For The Game
My argument for this one will be a little weak since it is maintainers that have come to dislike imagers, but I will go ahead and say that imagers are in a difficult spot at the moment. They are incredibly sought after in requisitions, and prove to be one of the most common purchases; there's certainly an issue when any given marine refuses to deploy without their funny vision enhancer.

This PR is open at the same time that #11888 is; both were meant to be in the same PR, but I found it a more sensible choice to separate these into two PRs, so as to give maintainers options to choose from.

## Changelog
:cl: Lewdcifer
del: Removed Optical Imager Goggles from the code and the game.
/:cl:

PR #11949, opened 2023-01-16
Partially revert #11887 - 58fc27e25cf5fdb66ecc8808ddb8fea512f5c17b

lbnesquik:
## About The Pull Request
This PR partially reverts #11887.
It returns the code for the imager googles, but it does not add them back in for marines to use.

## Why It's Good For The Game
I think the googles are an interesting middle ground between mesons and full night vision, and there's a real chance that they could be used in the future. Keeping the code doesn't cost us anything, and still keeps the previous PR goals fulfilled. I could see myself giving them to an ERT, or as rare loot in future map making endeavors. 

Actually, it does cost us something. 2,2 KB of data. I think we can handle that much in our codebase. 

## Changelog
:cl:
code: Bring back imagers in the code, but not in the vendors or supply. 
/:cl:

### DIFF
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -83,6 +83,12 @@
 		qdel(I)
 		qdel(src)
 		user.put_in_hands(P)
+	else if(istype(I, /obj/item/clothing/glasses/night/imager_goggles))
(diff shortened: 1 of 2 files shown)


TiviPlus:
We do not keep dead code

FIXTHERADBOXPLEASE:
Erm weren't imagers used for some veteran roles in the HvH gamemodes? Not really dead code...

PR #11965, opened 2023-01-18
Announces xeno autobalance to both sides.

Jalopkomit18:
## About The Pull Request
Per title. Will send a flavored announcement unique to each side when xeno autobalance is implemented. As an example, a 150% buff will multiply stats by 1.5. In this instance, it will be announced that xenos are 50% stronger.
## Why It's Good For The Game
Recently there was a round where many players on both sides did not know autobalance was on. This resulted in a lot of anger. Although some of it was misguided, I do think that transparency is important in a mechanic that affects gameplay so greatly. This is not a one-off occurrence, and occurs in crash rounds and lowpop frequently. It is never known to the player when this is implemented, and causes confusion to both marines and xenos as to why it feels gameplay has shifted so suddenly. 
## Changelog
:cl:
qol: Announces xeno autobalance to both sides when activated.
/:cl:

### DIFF
--- a/code/controllers/subsystem/monitor.dm
+++ b/code/controllers/subsystem/monitor.dm
@@ -213,6 +213,6 @@ SUBSYSTEM_DEF(monitor)
 		for(var/upgrade in GLOB.xeno_caste_datums[xeno_caste_typepath])
 			var/datum/xeno_caste/caste = GLOB.xeno_caste_datums[xeno_caste_typepath][upgrade]
 			caste.melee_damage = initial(caste.melee_damage) * GLOB.xeno_stat_multiplicator_buff
-	priority_announce("Our sensors have picked up a spike in psionic energy coming from the hive. Our experts estimate that the xenomorphs will be [GLOB.xeno_stat_multiplicator_buff*100]% stronger.", "Priority Alert")
-	xeno_message("A rabid fervor of energy sweeps across the hive; we feel [GLOB.xeno_stat_multiplicator_buff*100]% more fierce. The Queen Mother has sent us her strength!", force = TRUE)
+	priority_announce("Our sensors have picked up a spike in psionic energy coming from the hive. Resident experts estimate that the xenomorphs will be [(GLOB.xeno_stat_multiplicator_buff*100)-100]% stronger.", "Priority Alert") //if xeno buff is set to 150%, will announce that xenos are 50% stronger
+	xeno_message("A rabid fervor of energy sweeps across the hive; we feel [(GLOB.xeno_stat_multiplicator_buff*100)-100]% more fierce. The Queen Mother has sent us her strength!", force = TRUE) //same as above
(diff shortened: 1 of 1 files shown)


ZahaSanko:
counter argument, get rid of autobalance as it only brings issues

FosterGrime-Heymynameslouie:
> counter argument, get rid of autobalance as it only brings issues

agree with this

but hey knowing when you can get delimed in 1 (one) slash is I guess thats also neat to know

Luxyboo:
remove AUTYOBALANCE!!!!!!!!!!!C :D

Jalopkomit18:
> counter argument, get rid of autobalance as it only brings issues

Honestly blanket stat buffs are pretty wew, especially when changing a single stat value of 20 to 21 causes so much vitriol. However at least now we will know when it is implemented, and be able to gauge its effect more soundly.

Tupinambis:
Wait there were buffs to xenos during autobalance? That's actually horrendous remove that

ThelastArchiver:
Those that say autobalance should be removed are free to give an alternative that makes the game fun for aliens when half the hive consists of burrowed larva and the aliens have to fight with an extreme disadvantage.iam open for suggestions

MereWalker:
Darn. I was hoping it would at least be silent on if it was denied by an admin. I did that a lot.

Jalopkomit18:
> Darn. I was hoping it would at least be silent on if it was denied by an admin. I did that a lot.

What do you mean by this? Does it still announce even if it the autobalance request is denied?

lbnesquik:
Autobalance removal is a discussion for another PR, babes.

Jalopkomit18:
> Darn. I was hoping it would at least be silent on if it was denied by an admin. I did that a lot.
Regarding this, I don't see any reason why it would be announced if it denied. It should only be done when the balance values are actually applied.

On another note, there are also systems where respawn times are changed if the system decides it's necessary. Have any admins seen this occur? It should be logged, but I don't ever recall noticing any changes. If it does, I think that would also warrant announcement.

TiviPlus:
it does not announce to players because players cannot directly influence the result and this will only result in immersion breaks and OOC rage ( see https://discord.com/channels/498569646014857217/500801682058379265/1067192959830868079)

The argument can be made for improved logging to better tune it but this isn't it I'm afraid

chizzy376:
bring it back

Victor239:
> it does not announce to players because players cannot directly influence the result and this will only result in immersion breaks and OOC rage ( see https://discord.com/channels/498569646014857217/500801682058379265/1067192959830868079)

Agreed as someone who has played as an autobalanced xeno a fair amount this past week.

There are ways to prevent immersion-breaking by tailoring the announcement message text. And OOC rage isn't well reasoned as the linked message isn't even "rage", they're just pointing out why it would be beneficial for both sides if this mechanic existed. It in fact causes OOC rage when marines are suddenly surprised by autobalance, and why the xeno they were just duelling managed to completely overpower them.

PR #11967, opened 2023-01-18
Decouples CAS from Orbit and effectively buffs CAS

Tupinambis:
## About The Pull Request

CAS now uses (roughly) minimum orbit for delay between firing and impact. CAS points are now generated at max orbit levels.

Buffs lasers, rockets, minirockets, and the banshee because rounding and I figured it'd probably be fine, can adjust if it's not fine.

## Why It's Good For The Game

CAS is currently impotent, rarely being much use to the ground force and rarely doing even a little bit of damage to the xenos due to how much time xenos have to just walk away from the very clear warning dots. 

Part of this is the inconsistency of orbit level, something POs don't even have control over, and which there may be nobody available to or willing to change. The fact orbit level effects POs at all also doesn't really make sense when it is a jet that is entirely separate from the actual ship.

Hopefully this will make CAS more rewarding to play and more impactful on the round without being awful to deal with on the xeno side.

I'd like to see how CAS performs when it consistently has the best travel times and point generations it can have simultaneously before considering any other buffs.

## Changelog
:cl: Tupina
balance: Dropship point generation is no longer reliant on orbit level, instead generating as though it were maximum orbit.
balance: Condor/CAS projectile travel time is no longer reliant on orbit level, instead traveling as though it were minimum orbit.
/:cl:

### DIFF
--- a/code/controllers/subsystem/points.dm
+++ b/code/controllers/subsystem/points.dm
@@ -1,5 +1,5 @@
 // points per minute
-#define DROPSHIP_POINT_RATE 18 * ((GLOB.current_orbit+3)/6)
+#define DROPSHIP_POINT_RATE 24
 #define SUPPLY_POINT_RATE 20 * (GLOB.current_orbit/3)
 
 SUBSYSTEM_DEF(points)
--- a/code/game/objects/structures/dropship_ammo.dm
+++ b/code/game/objects/structures/dropship_ammo.dm
@@ -205,7 +205,7 @@
 	icon_state = "30mm_crate"
 	desc = "A crate full of 30mm bullets used on the dropship heavy guns. Moving this will require some sort of lifter."
 	equipment_type = /obj/structure/dropship_equipment/weapon/heavygun
-	travelling_time =  6 SECONDS
+	travelling_time = 4 SECONDS
 	ammo_count = 200
 	max_ammo_count = 200
 	transferable_ammo = TRUE
@@ -266,7 +266,7 @@
 	name = "high-velocity 30mm ammo crate"
 	icon_state = "30mm_crate_hv"
 	desc = "A crate full of 30mm high-velocity bullets used on the dropship heavy guns. Moving this will require some sort of lifter."
-	travelling_time = 3 SECONDS
+	travelling_time = 2 SECONDS
 	point_cost = 150
 
 
@@ -314,7 +314,7 @@
 	name = "high-capacity laser battery"
 	icon_state = "laser_battery"
 	desc = "A high-capacity laser battery used to power laser beam weapons. Moving this will require some sort of lifter."
-	travelling_time = 1 SECONDS
+	travelling_time = 0.5 SECONDS
 	ammo_count = 100
 	max_ammo_count = 100
 	ammo_used_per_firing = 40
@@ -382,7 +382,7 @@
 	ammo_id = ""
 	bound_width = 64
 	bound_height = 32
-	travelling_time = 4 SECONDS
+	travelling_time = 2.5 SECONDS
 	point_cost = 0
 	ammo_type = CAS_MISSILE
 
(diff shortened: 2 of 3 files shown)


Jalopkomit18:
Would this mean that staying max orbit would be the best and only option, for req point gen and buying time during hijack? I think that wouldn't be a bad idea as having a static value to balance is easier, but this pr should address this aspect.

GrayRachnid:
This would effectively remove any reason for lower orbit

maikilangiolo:
> This would effectively remove any reason for lower orbit

Yes. It already has zero reasons because absolutely no one goes lowest level for cas, considering just how good the higher levels are for point gen (which also apply to cash points btw).

Naaanii:
This removes the point of orbit completely ngl.

wezsel11:
the point for low orbit is that it makes me kill xenos quicker, i hate waiting for hijack. 
besides, what if we play with the OB/Railgun settings. Low orbit makes it possible to use OB/Railgun quicker by lowering the laser-time / impact time (/offshoot?)

Instead of only changing CAS we could also take in account OB/Railgun. Maybe we could take it all together.
In short:

**Low Orbit:**  
Cons: Lower ReQ points 
Pros: Railgun / OB more precise/lower laser time. 

**High Orbit:**
Cons: Railgun and OB takes more time to land it's shot (to further nerf/balance it create an offshoot, missing more often, more merciful against xenos)
Pros: Req points higher.

And CAS is decoupled. 

It adds more thought in decision. How you wanna roll this mission. Better ReQ or better OB/Railgun possibilities. 
just food for 🧠

Tupinambis:
If CAS is the only reason for orbit than I'm happy for orbit to be useless. That said making railgun and ob more orbit based isn't a bad idea, but outside the scope of this pr

lbnesquik:
I understand weapons travel times are changed, but was the trip back and forth for the condor changed too? 
It might not be needed to change both.

Tupinambis:
> I understand weapons travel times are changed, but was the trip back and forth for the condor changed too? It might not be needed to change both.

it was not.

Al-1ce:
The main issue I have with CAS is the very short deployment window. I hope this will alleviate that somewhat.

chemistrymain2:
> If CAS is the only reason for orbit than I'm happy for orbit to be useless. That said making railgun and ob more orbit based isn't a bad idea, but outside the scope of this pr

Intentionally invalidating a game mechanic in the process of buffing a feature isn't just lazy, it's outright bad. 

There's a feature being obsoleted, sure, but this is just going 'I want X to be stronger, I don't care if Y is useless in the process' and then going 'It's outside of the scope of the PR' in response.

Don't just be lazy and make messes for other people to clean up after you. If you want to buff CAS, go ahead. But don't make orbit useless in the process-either find an alternative solution or make railgun and OB orbit based as compensation as you described, since it's definitively not outside of the scope.

If you just want to half ass the PR, why do it at all?

Tupinambis:
> > If CAS is the only reason for orbit than I'm happy for orbit to be useless. That said making railgun and ob more orbit based isn't a bad idea, but outside the scope of this pr
> 
> Intentionally invalidating a game mechanic in the process of buffing a feature isn't just lazy, it's outright bad.
> 
> There's a feature being obsoleted, sure, but this is just going 'I want X to be stronger, I don't care if Y is useless in the process' and then going 'It's outside of the scope of the PR' in response.
> 
> Don't just be lazy and make messes for other people to clean up after you. If you want to buff CAS, go ahead. But don't make orbit useless in the process-either find an alternative solution or make railgun and OB orbit based as compensation as you described, since it's definitively not outside of the scope.
> 
> If you just want to half ass the PR, why do it at all?

When you get yelled at enough times for not atomizing PRs you keep shit simple, orbit still effects the Alamo and req generation so it's not completely useless 

And the point wasn't just to buff CAS, it was to make CAS more reliable without needing to basically fight the cap or SO to lower orbit.

Naaanii:
> And the point wasn't just to buff CAS, it was to make CAS more reliable without needing to basically fight the cap or SO to lower orbit.

CAS is already reliable. Free buffs with little to no downside, not even compensation is powercreepy, despite CAS already being in an okay state.

Tupinambis:
Not to cave to pressure but there's probably a better way of doing this. Closed for re-evaluation.

Al-1ce:
what shite :(

Redrover1760:
If invalidating orbit is an issue, would simply just flat decreasing the time it takes for CAS to hit would help? Or alternatively, buffing dropship point growth for each orbit flat. And maybe increasing the amount of points buying it with Req gets you.

PR #11050, opened 2022-09-24
Hivemind core is trackable by the xeno structure pinpointer

ivanmixo:
## About The Pull Request

- Hivemind core is trackable by the xeno structure pinpointer
- This required me making it a `/obj/structure/xeno` instead of `/obj/alien` to avoid shitcode

## Why It's Good For The Game

It makes sense, right? Hiveminds often hide themselves similar to maturity/evolution towers so there should be some way to track them down.

## Changelog
:cl:
balance: Hivemind core is trackable by the xeno structure pinpointer
/:cl:

PR #11971, opened 2023-01-18
Hivemind core is NO LONGER trackable by the xeno structure pinpointer

chizzy376:
## About The Pull Request
Reverts #11050 , you may now no longer locate the hivemind core using the pinpointer.
## Why It's Good For The Game
The change was not only disliked (Considering the positive to negative ratings on the PR) but it has severely restricted where hivemind can place their core. Right now you can very easily check and triangulate the position and order a marine to drop pod on the secluded hivemind core, provided the foolish T0 hasn't put it near the silo and every other structure. Why is this a bad thing, shouldn't hivemind stick to others? Due to the nature of the caste, you are practically placing an anchor at the spot you evolve, and having this anchor get destroyed kills you instantly. No way to defend it other than to wall spam (Which is still nothing considering flamers hit over 21 tiles at once in a single spray using the wide cone; or even a single PC user). I would really prefer this change to be merged as to bring back sneaky hivemind spots which in turn rewards the player from finding a very good spot by not dying instantly if the silo room is breached. The glow is still there, so putting it out of sight is not the only thing you should look out for when finding a good spot. But to the main question.
**HOW, and WHY** will this improve the current game? The main reason for this PR is to reward a skilled hivemind and to prevent the one and only core spot it has been forced into by the #11050 change.
## Changelog
:cl:
balance: Hivemind core is no longer trackable by the xeno structure pinpointer
/:cl:


Derrick2Electricboogaloo:
How about................. No!

chizzy376:
You're encouraged to rework hivemind into a fun caste, because right now it's objectively "The throw pick", as many put it.

chizzy376:
oh my god I don't know how to github
uhh uhhhh quick

PR #11982, opened 2023-01-19
Fixes Reqtorio AMR incendiary ammo printing violins

chizzy376:
## About The Pull Request
Someone capsed out the amr which is a non-existing item.

## Why It's Good For The Game
amr_magazine_incend is an item in the game.
AMR_magazine_incend does not exist and defaults to creating violins.

## Changelog

:cl:
fix: No more violins from Reqtorio AMR incendiary rounds
/:cl:

### DIFF
--- a/code/modules/factory/parts.dm
+++ b/code/modules/factory/parts.dm
@@ -205,7 +205,7 @@ GLOBAL_LIST_INIT(IFF_ammo, list(
 	. = ..()
 	recipe = GLOB.IFF_ammo
 
-/obj/item/factory_part/AMR_magazine_incend
+/obj/item/factory_part/amr_magazine_incend
 	name = "IFF antimaterial Incendiary bullet box"
 	desc = "A box with unfinished antimaterial Incendiary rifle rounds inside"
 	result = /obj/item/ammo_magazine/sniper/incendiary


shelteredfreak:
how the f-

HOW DOES SOMEONE MAKE IT MAKE VIOLINS

Lumipharon (approved):
honk.

chizzy376:
> HOW DOES SOMEONE MAKE IT MAKE VIOLINS

Quite simple. Since an AMR object doesn't exist, it defaults to the parent set object, which in this case is a violin. You can check this yourself at the 17th line here. https://github.com/tgstation/TerraGov-Marine-Corps/blob/d9116eb3bf54b6c9642a6e35b9ad71b26c39fe06/code/modules/factory/parts.dm

Strayed-Raven:
ah yes violins are perfect when you get disappointed. But sure merge this

chizzy376:
It's a behavior that's not supposed to happen outside of bug testing. What do you expect, a minigun? Violins are obvious that there's something very wrong.

FosterGrime-Heymynameslouie:
> > HOW DOES SOMEONE MAKE IT MAKE VIOLINS
> 
> Quite simple. Since an AMR object doesn't exist, it defaults to the parent set object, which in this case is a violin. You can check this yourself at the 17th line here. https://github.com/tgstation/TerraGov-Marine-Corps/blob/d9116eb3bf54b6c9642a6e35b9ad71b26c39fe06/code/modules/factory/parts.dm

this is unequivocal proof that the perfect weapon that all weapons stem from is the violin

we must reject shooting lead and return to smacking xenos with instruments

TiviPlus (changes requested):
Please consider bad commit messages are newly subject to pr closure I will correct it manually only once

PR #11984, opened 2023-01-20
Repairing a combat robot works like kits

Three333Digit:
## About The Pull Request
Specifically, you only consume 2 units of fuel (if it's a blowtorch) or 2 cables (if it's a cable coil) to repair a combat robot entire's body. Provided you don't move or get shoved out of repairing yourself like with kits. 

I increased the stack consumed from cable coil to repair from 1 to 2 for balance purposes. Also, indirect medic buff.

## Why It's Good For The Game
Makes combat robots less ass or painful to play. You have to bring an entire YouTool just to be able to "viably" play a combat robot or even yoink an Engineer's soldering tool.

## Changelog
:cl:
balance: combat robot repairing works like kits
balance: cable consumed when repairing a robot increased from 1 to 2
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -458,9 +458,10 @@ Contains most of the procs that are called when a mob is attacked by something
 
 	user.visible_message(span_notice("[user] starts to fix some of the dents on [src]'s [affecting.display_name]."),\
 		span_notice("You start fixing some of the dents on [src == user ? "your" : "[src]'s"] [affecting.display_name]."))
+	I.use_tool(volume = 50, amount = 2)
 
 	add_overlay(GLOB.welding_sparks)
-	while(do_after(user, repair_time, TRUE, src, BUSY_ICON_BUILD) && I.use_tool(volume = 50, amount = 2))
+	while(do_after(user, repair_time, TRUE, src, BUSY_ICON_BUILD))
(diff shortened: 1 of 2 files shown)


Three333Digit:
> Absolutely based for this PR, this works for soldering right? Because if not, then... It already has worked this way for blow torches and cables. If not, this is the opposite of a based PR.

The way it works right now is repairing a combat robot consumes 2 fuel units or 1 cable coil for each repair. This PR changes that so you only consume 2 fuel units or 2 cable coil ONCE to continuously repair a combat robot in one session... like how kits work right now.

ilikesaying:
still gonna steal welderpacks lol

QualityVan:
One difference is that kits apply once per limb. Robo repair can be done as many times as you want.

Xander3359:
> this works for soldering right?

Solder is already infinite

Lumipharon:
This is a huge buff, kits are not the same as welder/wire for robots.

Al-1ce:
> This is a huge buff, kits are not the same as welder/wire for robots.


..How, exactly? They are linear counterparts.

Coastfront:
> > This is a huge buff, kits are not the same as welder/wire for robots.
> 
> ..How, exactly? They are linear counterparts.

iirc kits only work for one limb, forcing you to re-use them (and lose another charge) per limb, welding/wire you click once and they start repairing every limb. Which would then mean that with this pr they'd use their fuel/wire cost once to fully heal every single limb.

PR #11989, opened 2023-01-20
King summon fix

Lumipharon:
## About The Pull Request
Fixed a bug with summon where dead/qdel'd xenos could get summoned.
## Why It's Good For The Game
Bug fix good.
## Changelog
:cl:
fix: fixed a bug with King's summon
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/king/abilities_king.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/king/abilities_king.dm
@@ -493,6 +493,7 @@
 			sister.remove_filter("summonoutline")
 		return fail_activate()
 
+	allxenos = X.hive.get_all_xenos() //refresh the list to account for any changes during the channel
 	for(var/mob/living/carbon/xenomorph/sister AS in allxenos)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/king/abilities_king.dm
```diff
@@ -493,6 +493,7 @@
 			sister.remove_filter("summonoutline")
 		return fail_activate()
 
+	allxenos = X.hive.get_all_xenos() //refresh the list to account for any changes during the channel
```
psykzz, line 496: Will this really resolve the issue? There is no check for dead, or whatever within `/proc/get_all_xenos`

Lumipharon: that procs uses xenos_by_typepath which is updated when xenos die etc, otherwise shrike/queen evo checks would always fail as they use the same list.

psykzz (approved):

novaepee:
When King is a necromancer

PR #12011, opened 2023-01-25
Deagle rework

Pariah919:
## About The Pull Request
Deagle damage mult goes from 1->2 (45x2=90)
Deagle fire rate goes from 0.7 to .25.
Deagle Recoil is now 0.
Loses all barrel attachies other than barrel charger due to the scatter changes in this PR.
Scatter is now 0, the minimum is 0, the max is 360.
It now has increasing scatter on shots, 15 per shot. It decays at a rate of 3 per second.

Golden Deagle is now a true admeme gun, it does x4 damage, but has insanely high scatter on shot. Better hit.

## Why It's Good For The Game
Makes the Deagle unique via the use of scatter increasing per shots, and is a good testbed to try this out for other guns in general in my opinion.

## Changelog

:cl:
balance: Deagle reworked, it now has .25 fire rate, no more barrel attachments, doubled the damage to 90. It has 0 scatter and recoil. It now builds scatter per shot, at a rate of 15 scatter every time you fire. This scatter decays at a rate of 3 per second.
balance: Golden Deagle is now gamebreakingly overpowered. It has x4 damage mult, enabling it to two shots runners.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -297,7 +297,7 @@
 
 /obj/item/weapon/gun/pistol/heavy
 	name = "\improper Desert Eagle pistol"
-	desc = "A magnum chambered in .50AE that comes with a serious kick. This one is engraved, <i>\"Peace through superior firepower\"</i>."
+	desc = "A magnum chambered in .50AE, the kick on followup shots will be hell, shoot with care to your shoulder. This one is engraved, <i>\"Peace through superior firepower\"</i>."
 	icon_state = "deagle"
 	item_state = "deagle"
 	caliber = CALIBER_50AE //codex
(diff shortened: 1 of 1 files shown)


Al-1ce:
> Add it to EORD or we will rarely ever see it in action lad 

It's already found in most EORD maps, and can be bought from req for cheap.

maikilangiolo:
Now make headshots reset the scatter, just like the hit game Synthetik

Derrick2Electricboogaloo:
> Add it to EORD or we will rarely ever see it in action lad

It is in seasonals, and req, and not expensive.

MereWalker:
Time to use it as the off hand main weapon for FC.

Tupinambis:
The Counter Strike is real

PR #12015, opened 2023-01-25
King slash scaling normalization and nerf

Pariah919:
## About The Pull Request
King slashes are normalized to the two bracket system every other xeno uses.
King slash changed from from 20->23->26->30 to 24->26.
## Why It's Good For The Game
King slash now works like every other xenos slash damage values, you gain slash at elder.
King slash is no longer ravager levels, it is now overall above average for its weight class.
## Changelog
:cl:
balance: King now scales like every other xenos, instead of gaining damage every maturity, it now gains slash at elder, new stats are 24 for young and mature, and 26 at Elder+. Old values were 20->23->26->30, so a net nerf at ancient maturity.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/king/castedatum_king.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/king/castedatum_king.dm
@@ -9,7 +9,7 @@
 	wound_type = "king" //used to match appropriate wound overlays
 
 	// *** Melee Attacks *** //
-	melee_damage = 20
+	melee_damage = 24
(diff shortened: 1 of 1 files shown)


Naaanii:
Basado

Strayed-Raven:
inb4 Xenos say no but mech gets the same tweaks either way

Shootnik01:
https://user-images.githubusercontent.com/75479005/214973204-7a2cde05-5b49-4476-8531-329b2370672d.mp4

PR #12023, opened 2023-01-27
Reduces Medium and Heavy armor slowdown. 

Tupinambis:
## About The Pull Request

Increases medium and heavy armor speeds from 0.5 and 0.7 to 0.4 and 0.5 respectively

## Why It's Good For The Game

Each tier of armor generally only lets you take an additional hit and sometimes not get fract when you otherwise would, which in a game where movespeed is more important than health most of the time, is kinda goofy. Light ends up being a pretty obvious choice if you are competent, unless you are using very specific loadouts that rely on modules with their own slowdown to balance things out.

This should hopefully make medium and heavy armor more appealing as choices. Small armor buffs should warrant smaller slowdown penalties. 

## Changelog

:cl: Tupina
balance: Reduces medium and heavy armor slowdown penalties.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/chestplates.dm
+++ b/code/modules/clothing/modular_armor/attachments/chestplates.dm
@@ -13,7 +13,7 @@
 	desc = "Designed for use with the Jaeger Combat Exoskeleton. It provides moderate protection and encumbrance when attached and is fairly easy to attach and remove from armor. Click on the armor frame to attach it. This armor appears to be marked as a Infantry armor piece."
 	icon_state = "infantry_chest"
 	soft_armor = list(MELEE = 25, BULLET = 45, LASER = 45, ENERGY = 35, BOMB = 30, BIO = 30, FIRE = 30, ACID = 35)
-	slowdown = 0.3
+	slowdown = 0.2
 	greyscale_config = /datum/greyscale_config/modularchest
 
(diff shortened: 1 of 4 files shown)


Lumipharon:
Armour has an exponential benefit. 5 extra armour at 50, vs 80 is not equivilent.

If anything I would say that light armour has too little slowdown, not that heavy armour has too much.

Coastfront:
The statement most armor classes allow you only to survive an extra hit isnt entirely true.

Light from medium is actually about 2 extra hits, where-as medium to heavy (unless if tyr or arrowhead are involved) is only actually extra hits from ravagers, and anything above 25-26 damage.  Heavy armor only gains value from tyr or arrowhead, where the extra armor then becomes relevent, and reduces most scources of damage by more than a single extra point.

Though, even with such defficiencies inmind this change seems unwarranted as a beartrap to crush a mouse. And ive neither witnessed light armor swelling with populariy nor medium & heavy being neglected, even in awe of the fact that tyr-less heavy is no tankier than medium in 9/10 cases. 
Light armor is popular for people who play offensively due to its speed, whereas heavy-medium is generally favored by people who are either more cautious, or care less about soloing a xeno (or are just playing melee.)

Armor speeds dont particularly feel as though they're crying out for change.

Tupinambis:
This PR is dumb lol.

PR #12031, opened 2023-01-28
Xeno Autobalance Rework

Tupinambis:
## About The Pull Request

Xenos now get the equivalent of phero buffs to speed and armor simultaneously at all times when autobalance is in effect. It can stack with actual pheros and any other modifications to those values. Recovery buffs are unchanged

Xenos can get up to 12.5 extra armor and 0.3 movespeed by default.

## Why It's Good For The Game

1.5x health and damage was goofy and led to roided up xenos that could kill you extremely quickly. With these changes xenos get more mobility and armor instead when underpopulated. This should compensate for the inherent lack of pheros when underpopped and give xenos an easier time engaging and disengaging from fights while retaining the recovery buffs that allowed them to quickly rejoin fights. 

## Changelog

:cl: Tupina
add: Xenomorph underpop autobalance now gives all xenos scaling armor and movement speed.
del: Xenomorph underpop autobalance no longer gives xenos scaling damage and health.
/:cl:


Araulius:
okay, but in practice, how does this help when there's 4 sisters vs about 30 marines?

Lumipharon:
This doesn't work for several reasons. 

Autobalance is there because there aren't as many bodies on the field as there should be, which means you aregoing to take more damage, do less damage, have less stuns and other abilities as a hive then you should otherwise. Damage and health buffs are the best thing to substitute for this.

Secondly the armour buff since it stacks means you WILL have situations where xenos have over 100 armour, making them unkillable without sunder or high ap weapons which is giga meme. I am also 99% sure this will cause issues when combined with other armour modifiers like that sentinel ability.

PR #12043, opened 2023-01-30
Mecha and pilot minimap fix

tanker4390:
## About The Pull Request
Mechas and their pilots will now properly display their minimap icons, with icons for occupied and unoccupied mechas (previously the pilot's icon would remain at the location where they entered the mecha until they exited it).

Screenshot_20230129_084227

Might still be buggy.
## Why It's Good For The Game
Bug fix.
## Changelog
:cl: Tanker
fix: Mechas and their pilots will now properly display their minimap icons
/:cl:


Starfirejordan:
will the In mech icon move with mech, or will mech have an icon for itself?

tanker4390:
> will the In mech icon move with mech, or will mech have an icon for itself?

The mech will have an icon, which changes depending on whether it's occupied.

Lumipharon:
Does this remove the minimap button for the MP when in a mech?

tanker4390:
> Does this remove the minimap button for the MP when in a mech?

Hmm. I didn't think of that being a side effect of calling `remove_minimap`. I'll see if I can fix that.

TiviPlus (changes requested):
This really feel like a hack instead of fixing the core issue which is that minimap registers to Move() of the user and not the mech

PR #12049, opened 2023-01-30
cheap barbed wire

tyeagg:
## About The Pull Request
barbed wire now costs 1 metal from 2
## Why It's Good For The Game
metal is expensive and limited and barbed wire is essential for a cade line. if your planning to make a good amount of cades its gonna take a good amount of metal just for the barbed wire so this solves that problem
## Changelog
:cl:
balance: barbed wire now costs 1 metal
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -12,7 +12,7 @@
 */
 GLOBAL_LIST_INIT(metal_recipes, list ( \
 	new/datum/stack_recipe("metal barricade", /obj/structure/barricade/metal, 4, time = 8 SECONDS, max_per_turf = STACK_RECIPE_ONE_DIRECTIONAL_PER_TILE, on_floor = TRUE, skill_req = SKILL_CONSTRUCTION_METAL), \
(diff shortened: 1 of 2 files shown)


Shootnik01:
> metal is expensive and limited

And that is why it cost 2 not 1. Not only it makes xenos lose their hp,trying to break it, it also adds additional hp to cads.

Losenis:
This'd also reduce the metal cost of razorwire from 3 to 2 since you can craft it with metal rods and barbed wire. This change affects way more than just barbed wire.

Derrick2Electricboogaloo:
Blessed PR, make it happen!

chizzy376:
Creepin'

Tailiik:
Sittin and shittin on cades marines love it. Having a massive fob should be costly

Naaanii:
Just order more metal or make it easier to get/salvage.

chizzy376:
If you'd want more metal, I'd suggest instead PRing a reversal of decon mats. Like now, you only get 1 metal sheet and 1 metal rod from manually deconning an rwall, a time costly endeavor. Pry up some floor tiles instead and in the same time you get over 20-30 sheets using an autolathe. The balance is all over the place, yeah.

Lumipharon:
Something being very good isn't a reason to make it cheap.

PR #12061, opened 2023-02-01
Fix vent crawl 

Dgeen:
## About The Pull Request

A year and a half ago, the mechanics of climbing into the ventilation were reworked. Larvae and hunters had to get into the ventilation quickly and silently. And it is. BUT ONLY IF YOU USE THE ALIEN TAB.
If you use alt-click speedup doesn't work. This pr fixes this bug.

## Why It's Good For The Game

Fix bugs good. Larva buff double good

## Changelog
:cl:
fix: Larva and hunter now crawl into the ventilation faster (as it should have been for like a year and a half)
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/vent_pump.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/vent_pump.dm
@@ -164,6 +164,10 @@
 /obj/machinery/atmospherics/components/unary/vent_pump/AltClick(mob/user)
 	if(!isliving(user))
 		return
+	if(isxeno(user))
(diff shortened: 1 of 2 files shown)


Ragecrabbo:
THANK YOU

TiviPlus (commented):
Shouldnt you just be updating handle_ventcrawl to just pass args properly

Dgeen:
> Shouldnt you just be updating handle_ventcrawl to just pass args properly

"vent_enter_speed" and "silent_vent_crawl" is a xeno var, so you can't use them with living_mob. And I can't rule out that it will be used not only by xenos

TiviPlus:
ehwhatever

PR #12063, opened 2023-02-01
spiderlings now get coated in resin jelly too :)

Zar000:
## About The Pull Request

When widow gets resin jellied her spiderlings do too.
## Why It's Good For The Game
Loosing all of your precious spiderlings or having to resist away 6 firestacks can be pain.
Also 1 flame tile denying you any gameplay isnt nice either
## Changelog
:cl:
qol: spiderlings get coated in jelly too :)
/:cl:


TiviPlus (changes requested):
Doc and signal handler

Xander3359:
Doc means you need to /// before the proc explaining what the proc does such as like this
image

PR #12065, opened 2023-02-01
Widow movespeed buff 

Zar000:
## About The Pull Request

Makes widow move slightly faster
## Why It's Good For The Game

Currently widow feels a little too sluggish and this should hopefully make her feel better.
## Changelog
:cl:
balance: Widow movement speed -0.2 -> -0.4
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/widow/castedatum_widow.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/widow/castedatum_widow.dm
@@ -13,7 +13,7 @@
 	melee_damage = 16
 
 	// *** Speed *** //
-	speed = -0.2
+	speed = -0.4
 
 	// *** Plasma *** //
 	plasma_max = 500


Oscari-pog:
Have you seen how a decent widow can utterly fuck up marines with its dualstun and frac
Powercreep PR

Dawson1917:
As someone who played widow for 2 weeks straight I don't think this is needed at all. I summed up widow's strengths and weaknesses compared to ravager (a caste that's a mix of ambush and frontline) on the discord so I'll just repeat it here: what widow has is higher DPS than a ravager with a longer stun while having less health, less speed, and more hard counters (any fire weapon, any explosive, and shotguns). 
Widow thrives on getting the jump on a single marine and doing massive damage to them - they should be punished if they fail their ambush or pick a bad target. Buffing their movement speed adds a lot more leniency to their mistakes, which I don't think is the attitude we should take on ambush castes.

chizzy376:
As a widow main/static I'd say the speed is not needed. You can already reach ridiculous speeds in 1 wide tunnels (Which grants you an insane movespeed buff, makes you the fastest in game if I recall correctly). What it actually does need is for burrow to not be useless, as right now it's a stalling tool that leaves you wide open if spotted. It would be cool to see the ability act as a resin hole in the form of tripping marines who step over you. Also having a different primo ability would be great, as the one we've got right now isn't really useful in any scenario.

Lumipharon:
Yeah I don't think they need a speedbuff, considering they're supposed to go fast in tight corridors/mazes

PR #12066, opened 2023-02-01
MG-60 reduction

Lumipharon:
## About The Pull Request
Reduced MG-60 bullet damage to 25 from 27.5. AP and sunder unchanged.

Also reduced capacity to 200 because the damage per mag is still hugely better than what it was previously.

The MG-60 buff pushed the gun to a level of DPS exceeding anything else in the game (except the two bigger MG's) which is not particularly ideal considering it is usable in hand, and viable on the move.

It will still be better DPS that any other portable gun by a fair margin, but slightly less guh to fight.
## Why It's Good For The Game
Giga MG-60 bad.
## Changelog
:cl:
balance: Reduced MG-60 damage to 25 from 27.5, and reduced magazine capacity to 200 from 250
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -633,7 +633,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 /datum/ammo/bullet/rifle/machinegun
 	name = "machinegun bullet"
 	hud_state = "rifle_heavy"
-	damage = 27.5
+	damage = 25
 	penetration = 10
 	sundering = 0.75
 
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -775,7 +775,7 @@
 	inhand_y_dimension = 32
 
 	caliber = CALIBER_10x26_CASELESS //codex
-	max_shells = 250 //codex
+	max_shells = 150 //codex
 	force = 35
 	aim_slowdown = 1.2
 	wield_delay = 1.5 SECONDS
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -775,7 +775,7 @@
 	inhand_y_dimension = 32
 
 	caliber = CALIBER_10x26_CASELESS //codex
-	max_shells = 150 //codex
+	max_shells = 200 //codex
 	force = 35
 	aim_slowdown = 1.2
 	wield_delay = 1.5 SECONDS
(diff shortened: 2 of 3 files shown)


FosterGrime-Heymynameslouie:
viable on the move

implies that the t60 can be moved around alot with bipod deployed and aimmode move speed reduction enabled?

srxtreme:
RIP NPC

XElectricX:
Leave the mag size alone. It's the most important/useful part of the gun. Just revert the damage buff it was recently given.

Tupinambis:
Disagree with mag size change. Damage nerf is good though

This1sPanda:
In which timeline do I live, where this gun is worthy of any kind of nerf. It's already a miracle we still have that peashooter in weapons vendor.

Al-1ce:
the T-160 is already a noob trap. this is really not needed.

Starfirejordan:
I support because currently a belt that holds 3 T60 mags plus 1 one loaded in the gun = 1000 rounds of ammo, not to mention if they use general storage moduel to carry 2 more which make its 1500, this also doesn't include satchel and backpacks, which make it ammo econ god tier. not to mention ammo packets which have more ammo and save better on room

So even tho I dislike this PR out of personal reasons, I support this balance change. because yea know maybe having more ammo then the SG85 with god tier suppressing fire isn't such a good thing

Lumipharon:
For reference, T60 is currently just a better, free minigun.
After the change it will be the same as a minigun but slightly lower ap.

Starfirejordan:
> For reference, T60 is currently just a better, free minigun. After the change it will be the same as a minigun but slightly lower ap.

Personally already been routing for Minigun to get a buff since it kinda just sucks, only thing good about it is ammo capacity, that the only reason I get it

Al-1ce:
> For reference, T60 is currently just a better, free minigun


 I don't really get your arguement. Minigun powerpacks have like... a thousand rounds. The T-60 didn't come up last time the minigun was buffed, so why now? What changed? I feel you got hosed down with it and died:  -- is this is a salt PR?

I don't really get the 'issue' here, but if you really want ...maybe consider making the minigun more dangerous? Higher RPM, sunder, raw damage, or a _MUCH_ larger ammo capacity would work.

Lumipharon:
> > For reference, T60 is currently just a better, free minigun
> 
> I don't really get your arguement. Minigun powerpacks have like... a thousand rounds. The T-60 didn't come up last time the minigun was buffed, so why now? What changed? I feel you got hosed down with it and died: -- is this is a salt PR?
> 
> IMO there is no 'issue' here, but if you really want ...maybe consider making the minigun more dangerous? Higher RPM, sunder, raw damage, or a _MUCH_ larger ammo capacity would work.

What changed is the t60 was buffed less than two weeks ago from 20 damage to 27.5. It was overbuffed, and I am tweaking it down. It is still significantly better than it was a fortnight ago.

Durbles:
If you're nerfing the damage and AP then nerfing the magazine seems unnecessary. 90% of the time that I see this gun in the field it's by marines who just wanna dakka.

PR #12069, opened 2023-02-02
Synths no longer enter crit

Neocloudy:
## About The Pull Request
This makes synths never enter crit and continue to exist until -125 health.
## Why It's Good For The Game
Synthetic critdragging being able to go on forever, or just being in crit permanently until you ghost, disconnect or rarely a marine realizes you're not dead and takes you away are the most annoying things ever. This annihilates both of those problems.

*Drafted probably until early 2/02 (tomorrow) for me to fix some stuff and be solid sure about the fixes*
## Changelog
:cl:
balance: Synthetics no longer have a critical state and will continue with a massive slowdown until -125 health.
/:cl:

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -52,6 +52,7 @@
 //Human sub-species
 #define isrobot(H) (is_species(H, /datum/species/robot))
 #define issynth(H) (is_species(H, /datum/species/synthetic) || is_species(H, /datum/species/early_synthetic))
+#define isartificial(H) (is_species(H, /datum/species/synthetic) || is_species(H, /datum/species/early_synthetic) || is_species(H, /datum/species/robot))
 #define isspeciessynthetic(H) (H.species.species_flags & IS_SYNTHETIC)
 #define ismoth(H) (is_species(H, /datum/species/moth))
 #define issectoid(H) (is_species(H, /datum/species/sectoid))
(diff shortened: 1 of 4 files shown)


Coastfront:
This also gives them a cold timer, right?

Naaanii:
Why not just make it like combat robots?

Coastfront:
> Why not just make it like combat robots?

This im fairly certain is "like combat robots", can continue on during negative HP until they drop dead, damage slowdown cripples em. The only way this seems different is whether or not synths still have no cold timer

Neocloudy:
My current plan is to make them continue until death, and never go perma, and just see if maintainers are fine. Anyway, my computer crashed so I’ll probably edit this later tonight or early tomorrow

TiviPlus:
Pr intends to fix synth permanent crit but you're instead just giving them robot race mechanics?
For what this PR is aiming to achieve this is a very intrusive solution, something like a button you can push to die while in crit would be better.

Lumi was also against

Losenis:
"robot race mechanics" That's kind of what they already are given the soldering repair and lack of sprint. What they have in storage options they lack in the ability to use guns to actually take advantage of said storage options. I don't know what reasoning there is that'd make it ok for a robot that can use sadar and autosnipers to not have crit to shoot xenos even more while the robot that can't is out of the question.

Derrick2Electricboogaloo:
God no do not make synth worse. STOP TOUCHING SYNTH.

chizzy376:
Me when the PR title lies about what it does

PR #12080, opened 2023-02-03
Marine light medical belt

tanker4390:
## About The Pull Request
The light medical belt, Available infinitely in clothing vendors. Has ~~13~~ 15 storage slots for medical items.

Screenshot_20230203_022014
## Why It's Good For The Game
Gives marines the option to fumble around with medical supplies, similarly to the tool:belt/pouch... This doesn't invite marines to hoard (non-negligibly) more meds, and is actually about the same as 2 medkit pouches. Medical belts are also slightly difficult to come by for some jobs that may have better use for them, such as SL and the hypospray-toting CO.
## Changelog
:cl: Tanker
add: Added a lighter medical belt for marines to use.
/:cl:


Losenis:
If it's the same as 2 pouches, increase the size by 1 or 2 slots so as to make a dedicated belt slot better, same rationale that made ammo belt better than 2 ammo pouches back when it actually held the same amount of stuff. This is because if you're running ammo belt you have more space overall by still skipping this item for 2 medkit pouches rather than medbelt and 2 ammo pouches, even if you'd be ready to sacrifice a few ammo slots for more meds.

Wait, if medkit pouch holds 7, isn't this actually worse than two?

code/game/objects/items/storage/belt.dm
```diff
@@ -81,6 +81,29 @@
...
+		/obj/item/reagent_containers/pill,
+		/obj/item/reagent_containers/syringe,
+		/obj/item/storage/syringe_case,
+		/obj/item/storage/pill_bottle,
+		/obj/item/reagent_containers/hypospray/autoinjector,
+		/obj/item/stack/medical,
+		/obj/item/storage/pill_bottle/packet,
+		/obj/item/reagent_containers/hypospray,
+		/obj/item/tool/surgery/solderingtool,
+		/obj/item/reagent_containers/dropper
```
novaepee, line 104: This line is calling the linter check to fail.

novaepee: fuck, this wasn't intended to be a review, was only a comment

Lumipharon:
Isn't this basically the medical rig belt that no one ever uses?

Losenis:
Medrig can fit roller bed, defib, and stasis bed, and nobody uses it because if you're looking for a belt slot medical storage item you'll loot a lifesaver belt anyway. Medrig does sound like the marine equivalent of lifesaver though, now that I think about it.

Xsdew:
> Isn't this basically the medical rig belt that no one ever uses?

It would be a smaller lifesave belt, the medrig can carry some things to avoid having a backpack/satchel for defibs.

PR #12082, opened 2023-02-03
GL 70 nerf

Naaanii:
## About The Pull Request
Increases the fire delay of the GL 70 to 2
## Why It's Good For The Game
Nade spamming shouldn't be as easy as it is, this should tone it down a bit and AGLS will retain it's top spot for nade spam.
## Changelog
:cl:
balance: Increases GL 70 fire delay to 2
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/grenade_launchers.dm
+++ b/code/modules/projectiles/guns/grenade_launchers.dm
@@ -115,7 +115,7 @@ The Grenade Launchers
 	)
 	starting_attachment_types = list(/obj/item/attachable/stock/t70stock)
 	attachable_offset = list("muzzle_x" = 33, "muzzle_y" = 18,"rail_x" = 14, "rail_y" = 22, "under_x" = 19, "under_y" = 14, "stock_x" = 11, "stock_y" = 12)
-	fire_delay = 1.2 SECONDS
+	fire_delay = 2 SECONDS
 	max_chamber_items = 5
 
 


Al-1ce:
Nades are limited. Wait them out you salty goober.
-1

Tupinambis:
Incredibly severe nerf tbh. If they have to be nerfed it shouldn't go all the way from 50 RPM to 30 RPM

GreatKingWombat:
> Nades are limited. Wait them out you salty goober. -1

"Limited"

Naaanii:
> Incredibly severe nerf tbh. If they have to be nerfed it shouldn't go all the way from 50 RPM to 30 RPM

This also gives single handed launchers a niche along with the AGLS which both have a lower rate of fire but one requires skill to use and the other points.

Tailiik:
> Nades are limited. Wait them out you salty goober. -1

The only way marines are using up the 800+ nades they get is if they have a whole squad of nade spammers in a 3 hour hellround or if 1 griefer detonates them all with fire

srxtreme:
In my opinion a nerf to GL70 is long overdue, but the delay between shots should be around 1.6 or 1.7 instead of 2.0

Lumipharon:
This seems like a pretty severe nerf.

Naaanii:
> This seems like a pretty severe nerf.

Severe but just. It puts it on par with the 35 and again, the one handed launchers can still be used and have more reason to do so with the right amount of skill.

Lumipharon (changes requested):
nerf it to like 1.6 if you want, but 2 is a lot.

Naaanii:
I'll let someone else nerf nade spam.

PR #12083, opened 2023-02-03
AGS Improvements and WP ammo

maikilangiolo:
## About The Pull Request
Man, I just wanted to fix the frag ammo but ended up making big QOL changes to AGLS.
1. HE nade explosion radius 4 --> 3. It was too big, which ended up being too cancerous for xenos to fight against because large explosion area from a normal grenade, but automatic. At the same time, it was awful to use due to FF, self hits, bounces etc. Rather than a nerf, it is an improvement
2. Fire dedlay 1 --> 1.1. Mostly done to avoid grenades being bouncing all over the place (you could cheese it to send grenades flying way past the max range of the gun). Bonus is that they wont bounce back into your face as much. Rejoice!
3. Frag ammo changes: gained the explosive tag so you could turf aim, and the shrapnel is now piercing. Shrapnel amount 20 --> 15, Spread 25 --> 20, damage/AP/sunder 25/20/1 --> 30/20/3, range 4 --> 2. These series of changes both make frag not be stupidly weak (22 shells to kill an ancient warrior vs 6 HE ones), as well as improve usability by changing the AOE from 9x9 to 5x5. While this was intended to avoid massive FF and self hitting because you shot a xeno that wasnt 6 tiles away from anything else, it also means the damage area is smaller, so a benefit for xenos too. The mob piercing is to compensate from the large AOE loss and to make the ammo worth the cost.
4. New WP ammo. It only ignites mobs without setting tiles on fire. It does not lose that much damage compared to normal frag, but it loses the mob piercin in exchange. Testing felt like an accetable tradeoff: fire stacks (with no flames) in a more concentrated area. Remember jelly protects against it
5. finally, recolored the ammo since they were nearly undistinguishable.
## Why It's Good For The Game
The new ammo is something I thought was needed for the AGLS to be more versatile. For a 700 point gun, it was a bit limited when a free GL70 or the AMR offered more options (the latter which has shrap and incendiary ammo too). 
All the other changes are rebalances and improvements, arguably for both xenos (smaller aoes, slightly slower firing) and marines (less (self)FF, actual ammo choice).

video showing the damage comparison. note: both frag and HE have the old AOEs (4 each), and they were changed to 2 and 3 respectively in this PR.

https://user-images.githubusercontent.com/40152611/216758775-86213f90-2b9d-42b5-b51c-56660b9a5bb6.mp4

## Changelog
:cl:
add: added WP shrapnel AGLS ammo
balance: rebalanced AGLS frag to be better
balance: rebalanced AGLS AOEs to be less oppressive for both sides
qol: recolored AGLS ammo to be easier to distinguish
/:cl:


TiviPlus (approved):
Code OK not commenting on balance

PR #12087, opened 2023-02-03
Fixes vatborn having way too much slowdown

QualityVan:
## About The Pull Request
Aged vatborn have 1.05 species slowdown. This would be pretty minor, were it not for the default species slowdown being 0 and not 1. Oops.
## Why It's Good For The Game
I went slow once.
It was honestly alright but this is excessive.
## Changelog
:cl:
fix: Vatborn have been given permission to take off their training weights.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -349,7 +349,7 @@
 	icobase = 'icons/mob/human_races/r_vatgrown.dmi'
 	brute_mod = 1.05
 	burn_mod = 1.05
-	slowdown = 1.05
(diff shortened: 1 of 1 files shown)


Al-1ce:
Personally, I've always thought the health reduction was bad enough. Vatborns take forever to grow; we rarely get more than one or two a round. I don't see why they need slowdown to begin with.

I know for a fact this _is_ legacy code from when this server was an infant, and I don't think vatborns being quite this shart fits anymore.

SilencedMP5A5:
honestly vatties should probably not have any disadvantages considering take SSD mob exists

PR #12091, opened 2023-02-04
Unjanks rav fire interation

Lumipharon:
## About The Pull Request
Ravagers now gain 50 rage every tick while on fire.

Currently, they completely refill their rage AND reset their charge cooldown every time a fire turf ticks while they are standing on it.
This is extremely powerful when abused, but also super unintuitive as you can move on and off fireturfs and never get the buff, or get it for touching a turf for 0.1 second.
## Why It's Good For The Game
Fire is no longer giga for rav, and provides a more modest, consistant bonus.
## Changelog
:cl:
balance: Ravagers now gain rage every tick when on fire, instead of refilling rage entirely when on a flame turf
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/ravager.dm
@@ -30,6 +30,14 @@
 	H.throw_at(get_turf(target_turf), RAV_CHARGEDISTANCE, RAV_CHARGESPEED, H)
 	H.Paralyze(2 SECONDS)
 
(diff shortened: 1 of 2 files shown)


QualityVan:
One potentially significant balance hit from this that you don't mention is that previously rav didn't take any damage from standing in fire and now it will.

Lumipharon:
Good point, although rav has fairly high fire resist, and it still takes damage from walking through fire tiles (just not when it ticks, because fire code is awful).

I'm going to address that in a separate pr actually.

TiviPlus (approved):

PR #12092, opened 2023-02-04
Thermobarics have 5 second reload time instead of 20 seocnds.

Pariah919:
## About The Pull Request
Title, 5 second reload instead of 20 second.

## Why It's Good For The Game
Honestly, DS is so rare that if you get lucky enough to get DSL you should be able to use your overpowered instafrag gun pretty often, and the marine thermo really, really needs this.

## Changelog

:cl:
balance: Both Marine and DS Thermobarics have a 5 second reload time.
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/specialist.dm
+++ b/code/modules/projectiles/magazines/specialist.dm
@@ -235,7 +235,7 @@
 	icon_state = "quad_rocket"
 	max_rounds = 4
 	default_ammo = /datum/ammo/rocket/wp/quad
-	reload_delay = 200
+	reload_delay = 5 SECONDS
 
 /obj/item/ammo_magazine/rocket/m57a4/ds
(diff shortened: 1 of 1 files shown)


Lumipharon:
A sadar takes 6 seconds to reload a single rocket. Thermo 20 seconds is actually faster than this already.

Naaanii:
> A sadar takes 6 seconds to reload a single rocket. Thermo 20 seconds is actually faster than this already.

It's one 4 rocket mag, essentially making it more than 3 times longer.

maikilangiolo:
> A sadar takes 6 seconds to reload a single rocket. Thermo 20 seconds is actually faster than this already.

A sadar doesn't have 0 direct hit damage and sunder. And the DS are OP on purpose

srxtreme:
Great change. This is probably enough to make the thermobaric launcher playable, if a rather niche and fun thing to use.

PR #12093, opened 2023-02-04
Reworks Queen Healing and Queen Plasma to hold charges.

Tupinambis:
## About The Pull Request

Queen Heal and Queen Give Plasma can hold up to 3 charges at a time that can be used in quick succession, they recharge at the same rate as the old cooldowns for these abilities. Also updates some old flavor text and re-adds psychic whisper.

## Why It's Good For The Game

Queen Heal and Give Plasma have very low cooldowns (5 and 8 seconds respectively), and unless you are hardcore minmaxxing or are very patient you tend to waste a ton of uptime on those abilities. It's pretty annoying, especially with give plasma, so the goal of this PR is to let queen's focus on other things for a bit longer without being penalized. Additionally it encourages teamwork and decision making a bit more in how you can either dump plasma/healing on one xeno or distribute them among multiple.

Flavor text is just flavor, if you don't like the flavor we can change the dish.

## Changelog

:cl: Tupina
balance: Queen heal and give plasma now hold up to 3 charges (uses) that can be used in quick succession. No other stat changes.
/:cl:


QualityVan:
You're adding the same code in two places. How effort would it be to put cooldown stacks as an option on the base xeno action?

Lumipharon:
This is hugely buffing both abilities, I wouldn't merge on that alone.

Tupinambis:
> This is hugely buffing both abilities, I wouldn't merge on that alone.

chizzy376:
I like it, makes the queen have to think ahead instead of mindlessly spending her charges.

Lumipharon:
> So how would you want me to rebalance it?

Why are you changing it exactly? Adding charges just means queen can heal 3 times as often, or more likely, just insta fully heal hurt xenos.

QualityVan:
It's not three times as often. It's exactly as often, except you can pool them. Instead of being able to do one use every eight seconds, you can do three every 24.

RaidarV:
> It's not three times as often. It's exactly as often, except you can pool them. Instead of being able to do one use every eight seconds, you can do three every 24

If using the skill doesn't reset the ongoing stack's cooldown... Yeah, this buffs Queen so her healing is MUCH faster recharging.

Lumipharon:
Not ok with this if the CD is unchanged. AS above, you're getting far more effective use out of an already fairly spammable skill.

PR #11479, opened 2022-11-07
Implements new floors and window sprites

RipGrayson:
## About The Pull Request
Title.

**I did not have hand in making any of these sprites all credit goes to https://github.com/aGhostCoder, and Gregormack** Go thank them instead.

## Why It's Good For The Game

This is just part of a widespread respriting, it needs to be done to set the foundation for more resprites.

## Changelog
:cl:
imageadd: Added new window and floor icons, check the PR for proper attribution.
/:cl:

PR #12095, opened 2023-02-04
turf and other stuff resprite project 

tyeagg:
## About The Pull Request
an attempt to continue this pr
https://github.com/tgstation/TerraGov-Marine-Corps/pull/11479
image
image
image

## Why It's Good For The Game
more cohesive artstyle with the newer stuff??? general idea is just less is more
## Changelog
:cl:
imageadd: new stuff
/:cl:


Al-1ce:
The standard hallway floors could be a bit darker. Rest looks fine
+1 from me.

Blundir:
Do I really need to tell that minimalist floors without shading or anything, don't really fit the artstyle well...

ClosetedSkeletonGH:
I don't want to sound rude but I agree, the old tiles had slight perspective that helped with the fake directional perspective. I feel like the previous floors did not have any overwhelming issues to need replacing with something like this that is a graphical downgrade

RoyalEntomologist:
i have no idea if you have any interest in atomizing this or doing edits to the sprites in the PR. some thoughts if you do, though:

- any file cleaning is good, removing dupes and streamlining names all looks okay, no errors there
- i like the scorched effects on the new tiles more
- my problem is the actual tiles themselves, as others pointed out; reducing color palette --> higher contrast on the tiles which results in them moving from the background of the graphics into the midground and drawing more visual attention when looking at everything on the whole. this is great for important machinery and objects meant to layer on top of others, less good for the "backdrop" stuff.
if you have any interest in comboing the scorched decals with the old tiles, i think those would look nice

Kuro020 (changes requested):
It seems to me that these tilesets look weird because they're saturated to oblivion, some of them are straight up neon signs, that probably needs toning down

tyeagg:
Later

PR #12108, opened 2023-02-06
Eviscerating charge now cost less, deals stagger but has much less stun.

Pariah919:
## About The Pull Request
1/10ths Eviscerating charges stun duration, it now deals 2 stagger.
Decreases cooldown from 20 to 15 seconds.
Charge plasma cost reduced by 40%.
## Why It's Good For The Game
The fire changes weren't really enough to move Ravager back down to a reasonable level of balance, so I believe these changes are fine since the caste will still be very viable, .2 stun is enough to force marines to drop their weapon and combo'ed with ravage is still a high amount of stun.

Ravager has around 3 seconds of stun in current state, this is very long for an ability that isn't very hard to land, and on the highest damage caste in the game with a large duration of EHP is quite strong, these changes should generally make Ravager easier to deal with, while keeping charge useful as an ability by letting you use it more often in an engagement and team fight.

I considered nerfing Ravage but that would be a far harsher nerf in my opinion as its generally used more often than charge, and my issue with the caste right now is the 3 seconds overall of stun the combo has.
## Changelog
:cl:
balance: Eviscerating charge stuns for .2 seconds rather than 2 seconds, charge cooldown reduced to 15 seconds from 20. Charge cost reduced by 40%.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
@@ -7,7 +7,7 @@
 	desc = "Charge up to 4 tiles and viciously attack your target."
 	ability_name = "charge"
 	cooldown_timer = 20 SECONDS
-	plasma_cost = 500 //Can't ignore pain/Charge and ravage in the same timeframe, but you can combo one of them.
+	plasma_cost = 300
 	keybinding_signals = list(
 		KEYBINDING_NORMAL = COMSIG_XENOABILITY_RAVAGER_CHARGE,
 	)
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/ravager.dm
@@ -29,6 +29,7 @@
 	target_turf =  get_step_rand(target_turf) //Scatter
 	H.throw_at(get_turf(target_turf), RAV_CHARGEDISTANCE, RAV_CHARGESPEED, H)
 	H.Paralyze(0.2 SECONDS)
+	H.adjust_stagger(2)
 
 /mob/living/carbon/xenomorph/ravager/fire_act()
 	. = ..()
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/ravager.dm
@@ -18,7 +18,7 @@
 // ***************************************
 // *********** Mob overrides
 // ***************************************
-/mob/living/carbon/xenomorph/ravager/Bump(atom/A)
+/mob/living/carbon/xenomorph/ravager/Bump(atom/A) // This is mostly triggered by eviscerating charge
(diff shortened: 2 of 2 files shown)


Barnet2:
Good change. After playing ravager for awhile, def deserving of a nerf.

Naaanii:
Doesn't .2 mean there's no stun at all? You might as well remove it and keep it a knockdown only.

Al-1ce:
Not a fan.

KillerOrcaCora:
Don't like this change either.

srxtreme:
In my opinion reducing the stun to 1 second but also reducing the cooldown to 15 seconds would be better.

Dawson1917:
The cooldown doesn't matter much since the limiting factor of Eviscerating Charge is how much Rage it takes to cast. Ravagers only regenerate enough Rage on their own to use Ravage, so to even use Eviscerating Charge they need to go find some fire or stand infront of bullets for a few seconds, run to weeds and heal whatever damage/sunder they took in that time, *then* they can go use the Eviscerating Charge + Ravage combo on a single marine and most likely still not even get that marine into crit before the Ravager needs to run off and start the cycle over again.
0.2 seconds is such a microscopic stun it may as well not exist. Ravagers definitely don't need to be made even weaker than they are now.

Losenis:
Reducing stuns and buffing other areas? Where have I seen this before?

Goblinu:
The 5 seconds of reduced cooldown is negligible and only there to sugarcoat this PR

Nicolai904:
you should make the cost way less if your taking the whole stun away p much

QualityVan:
Since you're reducing the stun by 90%, I think it's only fair that you also reduce the cd by 90%, down to two seconds.

ivanmixo:
Completely missing the point, the limiting factor of using the charge is not the cooldown, but the rage.

Araulius:
you're out of touch, I'm out of time

Pariah919:
I guess I'll just lower the plasma cost and make it stagger on hit

Araulius:
if you're gutting the stun, I expect a higher delimb chance

Naaanii:
> I guess I'll just lower the plasma cost and make it stagger on hit

Funny idea. Give it 0 cooldown, make it slightly harder for Rav to generate rage passively (fire, taking damage etc.) Reward it more for landing hits and kill t-foot/pepperball removing rage.

Goblinu:
> l t-foot/pepperball

It doesnt matter what game it is, but introducing mechanics or weapons that hardcounters a specific type(in this case xenos) never ends well

Naaanii:
> > l t-foot/pepperball
> 
> It doesnt matter what game it is, but introducing mechanics or weapons that hardcounters a specific type(in this case xenos) never ends well

The mechanic already exists.

Goblinu:
> The mechanic already exists.

image

zee11zee:
The new idea seems a lot fairer for both.

KillerOrcaCora:
Your not making it to primo anymore...

Lumipharon:
I'm not particularly keen on this. Charge is basically there for the stun, and uses a giga amount of rage to use. And if you fuck it up that's a large loss.

PR #12109, opened 2023-02-06
Makes acid melt flares faster

Zar000:
## About The Pull Request
Exactly what the title says

## Why It's Good For The Game

Marines have a lot of tools to light up areas, some of them, such as howie flares & CAS flares have almost no counter-play other than waiting them out.

The thrown flares however, we can melt but it takes forever unless queen / boiler does it.

This still makes it take a while with drone level acid, but 2x faster.
## Changelog
:cl:
balance: All acids melt flares 2x faster
/:cl:


docRoboRobert:
after the flare boxes massive buff i'd say this is needed.

code/game/objects/effects/aliens.dm
```diff
@@ -168,7 +170,10 @@
 		return
 	if(loc != acid_t.loc && !isturf(acid_t))
 		loc = acid_t.loc
-	ticks += delta_time * acid_strength
+	if(istype(acid_t, /obj/item/flashlight/flare))
+		ticks += delta_time * min((acid_strength * flare_strength_multiplier), flare_strength_max_melt_rate) // we dont wanna make it possible to melt flares too fast
```
Lumipharon, line 174: Having the typecheck here is snowflake. There should be a proc on the target side to return an acid time or whatever.

Lumipharon (changes requested):

Tailiik:
It would be very funne if a primo prae could trigger flare melt with ranged spit

Lumipharon (changes requested):
Make this a proc instead of an object var as discussed.

Mannybrado:
maintainers help

PR #11625, opened 2022-11-25
Flamethrower Rework

tyeagg:
## About The Pull Request
reworks the flamethrower to be like a proper flamethrower, fire burn time significantly decreased so if you want area denial your gonna need to constantly expend ammo deletes nozzles and flamethrower code also reworks the mech flamethrower to also be like this deletes x fuel and replaced with new flamethrower that fires x fuel but uses normal flamethrower tanks

todo: make bounce projectiles actually work, get a new flamethrower sound and finish my overheat mechanics pr to apply it here

https://user-images.githubusercontent.com/98577688/203898713-7ae7d7a3-64ba-427e-b57e-4ac300425652.mp4

https://user-images.githubusercontent.com/98577688/203898734-d2a8fe60-b3b7-4d20-b912-165c8493d140.mp4

## Why It's Good For The Game
more fun flamethrower gameplay, flamethrower will require more ammo management

## Changelog

:cl:
add: Added new mechanics or gameplay changes
add: Added more things
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

PR #12113, opened 2023-02-06
Removes wide flamer nozzle from marine vendor

ClosetedSkeletonGH:
## About The Pull Request

Only removes the wide nozzle attachment from the marine vendor. 

This does not remove it from the game or anywhere else.

## Why It's Good For The Game

With SURT and wide nozzle, you can functionally stand in the middle of your flame while slowly walking forward and flaming with infinite ammo, while an ungaball walks behind you ready to fire on an open and bright field from flames. 

Wide absolutely destroys mazes, the downside to wide is that it goes through ammo fast, but flamer fuel is infinite so it has 0 tangible downsides to ever using it, and not using it right now means you are going at a severe disadvantage.

Its clear that:
1) the flamer rework is never going to get done (#11625)
2) flamer fuel is going to remain infinite
3) buffing jelly by an extreme amount just to counter 1 attachment means that the attachment is the issue

Flamers suck in so many ways right now and are clearly not going to get better so this is what it has come to. You get to choose between being absolutely useless (extended) or being an absolutely monster that if you have atleast 3 braincells and some cooperative marines, you can burn down a whole map's worth of maze.

Before you whine and sob, you can PR a fix yourself if you don't like this one.

image
image
image

## Changelog

:cl:
del: Removed wide flamer nozzle from marine vendor
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -169,7 +169,6 @@
 			/obj/item/weapon/gun/rifle/pepperball/pepperball_mini = -1,
 			/obj/item/ammo_magazine/rifle/pepperball/pepperball_mini = -1,
 			/obj/item/attachable/flamer_nozzle = -1,
-			/obj/item/attachable/flamer_nozzle/wide = -1,
 			/obj/item/attachable/flamer_nozzle/long = -1,
 		),
 		"Boxes" = list(


FIXTHERADBOXPLEASE:
At the time this wasn't needed since there was a flamer rework ongoing... it's clear by now that it's never going to be completed though. 
And if the complete rework is ever merged, this PR is easily undone.

Strayed-Raven:
This will change nothing. It only makes it more tedious

FIXTHERADBOXPLEASE:
"More tedious" = takes more time = gives Xenos both more time to do anything, and more abilities to punish flamers. Especially now that it'll take multiple sprays to seal off a path.

Al-1ce:
Add it to req at a bare minimum.

novaepee:
Likely this PR won't get merged since Lumi made wide flamer nozzle.

FosterGrime-Heymynameslouie:
I think it would be better to half the distance, maybe even more, the flames can reach instead of out right removing the attachment? would be more sensible I think

still, this pr was a long time coming

Strayed-Raven:
> "More tedious" = takes more time = gives Xenos both more time to do anything, and more ability to punish and manuver around flamers. Especially now that it'll take multiple sprays to seal off a path.

You actually believe that? Amazing

Araulius:
> > "More tedious" = takes more time = gives Xenos both more time to do anything, and more ability to punish and manuver around flamers. Especially now that it'll take multiple sprays to seal off a path.
> 
> You actually believe that? Amazing

maybe actually bother to refute it instead of making yourself out to be a clown

ClosetedSkeletonGH:
> Add it to req at a bare minimum.

Adding it to req just means it now is a problem 5-15 minutes into the Op instead of at the start of the Op

lbnesquik:
> > Add it to req at a bare minimum.
> 
> Adding it to req just means it now broken at 5-15 minutes into the Op instead of at the start of the Op

Req hell is a thing.

Kuro020:
Dealing with wide flamer is very guh, the only counter is a hivelord spamming jelly and thats not always a thing, moreover fuel is infinite, surt is pretty cheap and kelotane is basically infinite as well, there is literally not a limit to how much time a wide flamer can keep going and playing against it is very unfun, kill it

Joe13413:
first quickbuild being way overpowered... now flamer wide nozzle is being removed.... truly we are in the dark age of mazes..
(sidenote: most good flamers just use standard anyways because of the fact that spray can't destroy a wall in 1 blast)

Al-1ce:
I guess the alternative I'd prefer here would be making it _very_ limited - ie, restricted to one in the vendors, as a specialized thing, in the specialized tab. Similar to some of the HMGs and what-not. 

..As for jelly being "unobtainable", it isn't. T1 drones are able to dispense it with ease. Removing the nozzle entirely is stupid because there is a guaranteed 100% effective counter to flames every round.

I should throw in that dedicated flamers barely carry more than a pistol, so the amount of defense they throw away for picking the class is quite severe. They're basically an AoE glass cannon. Their vulnerability (being far from the group or in the front to clear mazes) is what makes them balanced. At least that's the way I've always seen it.

I've fought against flamers, even as a drone. It is not difficult. Jelly is cheap as shit and makes you completely immune to fire. I think this is overall a massive skill issue on the xeno of things.

ClosetedSkeletonGH:
> 

Jelly is extremely difficult to coordinate with more than 1-2 xenos, the issue with wide is that it clears wide fields and leaves a huge amount of flames which provides light. If the marines have any braincells, they can have many marines follow the flamer, so that way any xeno attacking the flamer, jelly or not, just gets melted from multiple lines of bullets in an open field. If jelly could be applied to 3+ xenos easily then sure, but thats entirely unfeasible rn

I don't know which flamers you've seen but every flamer of any non-suicidal quality has a backup gun, usually shotguns or AR's. Taking only a flamer without having a very good marine team behind you is suicide and an awful player. 

Again, fighting a flamer =/= fighting a wide flamer that has a marine team behind them, which is why this is an issue. Sure a lot of flamers suck, but you've clearly never gone up against one that actually understands how the flamer works, how to stick with a group, and what its niche is.

Lumipharon:
I'm open to removing it, but if it's proving problematic I would just dramatically reduce the flame duration.

That makes it horribly impractical for wall clearing, and the area denial becomes extremely short term.

Tailiik:
Anyone who isn't a 0 xeno hours marine main can see that wide flamer is cheese

Kuro020:
> I'm open to removing it, but if it's proving problematic I would just dramatically reduce the flame duration.
> 
> That makes it horribly impractical for wall clearing, and the area denial becomes extremely short term.

make it fire green flames, dew it

Lumipharon:
#12146 we're going to nerf duration first.

PR #12119, opened 2023-02-07
Buffs power fist a little

TiviPlus:
## About The Pull Request

Power fist uses 33% less power and pushes "large" mobs ie most T2+ 1*setting more tiles away
## Why It's Good For The Game

Power fist is kinda in a meh spit this should hopefully make it a little more usable

## Changelog
:cl:
balance: power fist uses a little less power and pushes large xenos a little further away
/:cl:


Losenis:
Make it able to be tacreloaded like the rest of energy weapons too, please. Powerfist is still ancient enough from the days it used normal cells than actual energy weapon cells so it never got tacreload.

PR #12122, opened 2023-02-07
Adds maintenance activation to robots

QualityVan:
## About The Pull Request
Robots get an action button they can press to stun and blind themselves for 5 seconds while healing up to 20 total brute and burn damage.

## Why It's Good For The Game
Essentially a very light, robotic version of surgery/sleep. Commit to disabling yourself for a period and you can heal more/for free.

## Changelog

:cl:
add: Maintenance mode for robots: disable yourself for five seconds to heal up to 20 damage.
fix: Fixed blindness not clearing properly if you tried to remove it a certain way.
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -153,6 +153,44 @@
 		if(prob(10) && owner.health > owner.health_threshold_crit)
 			owner.emote("snore")
 
+///Basically a temporary self-inflicted shutdown for maintenance
+/datum/status_effect/incapacitating/repair_mode
+	id = "repairing"
+	tick_interval = 1 SECONDS
+	///How much brute or burn per second
+	var/healing_per_tick = 4
+	///Whether the last tick made a sound effect or not
+	var/last_sound
+
+/datum/status_effect/incapacitating/repair_mode/on_apply()
+	. = ..()
+	if(!.)
+		return
+	owner.disabilities |= BLIND
+	owner.blind_eyes(1)
+	ADD_TRAIT(owner, TRAIT_INCAPACITATED, TRAIT_STATUS_EFFECT(id))
+	ADD_TRAIT(owner, TRAIT_IMMOBILE, TRAIT_STATUS_EFFECT(id))
+
+/datum/status_effect/incapacitating/repair_mode/on_remove()
+	owner.disabilities &= ~BLIND
+	REMOVE_TRAIT(owner, TRAIT_INCAPACITATED, TRAIT_STATUS_EFFECT(id))
+	REMOVE_TRAIT(owner, TRAIT_IMMOBILE, TRAIT_STATUS_EFFECT(id))
+	return ..()
+
+/datum/status_effect/incapacitating/repair_mode/tick()
+	var/sound_to_play
+	if(owner.getBruteLoss())
+		owner.heal_limb_damage(healing_per_tick, 0, TRUE, TRUE)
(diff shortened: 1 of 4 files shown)


Jalopkomit18:
I'm just assuming, but wouldn't regular tool healing be faster than this? Is this just supposed to be a last ditch heal or an analogue to regular tool healing.

FosterGrime-Heymynameslouie:
image

code/datums/status_effects/debuffs.dm
```diff
@@ -153,6 +153,44 @@
...
+		owner.heal_limb_damage(healing_per_tick, 0, TRUE, TRUE)
+		sound_to_play = pick('sound/items/welder.ogg', 'sound/items/welder2.ogg')
+	else if(owner.getFireLoss())
+		owner.heal_limb_damage(0, healing_per_tick, TRUE, TRUE)
+		sound_to_play = 'sound/items/deconstruct.ogg'
+	if(!sound_to_play || last_sound)
+		last_sound = FALSE
+		return
+	last_sound = TRUE
+	playsound(owner, sound_to_play, 50)
```
Lumipharon, line 192: This is quite loud, you'll be able to hear robros powernapping from like 4 screens away unless you add a max range.

code/datums/status_effects/debuffs.dm
```diff
@@ -153,6 +153,44 @@
...
+	if(owner.getBruteLoss())
+		owner.heal_limb_damage(healing_per_tick, 0, TRUE, TRUE)
+		sound_to_play = pick('sound/items/welder.ogg', 'sound/items/welder2.ogg')
+	else if(owner.getFireLoss())
+		owner.heal_limb_damage(0, healing_per_tick, TRUE, TRUE)
+		sound_to_play = 'sound/items/deconstruct.ogg'
+	if(!sound_to_play || last_sound)
+		last_sound = FALSE
+		return
+	last_sound = TRUE
```
Lumipharon, line 191: This is basically just making it only generate a sound every two ticks right? Would it be easier to just have 2 second ticks and double the heal?

Lumipharon (approved):

code/datums/status_effects/debuffs.dm
QualityVan: Trouble with that is each tick is purely either brute or burn and purely on a specific limb, so larger but slower ticks makes it less value on spread damage (which is a target situation for use). Also, hard to have 5 second duration then.

FosterGrime-Heymynameslouie:
after playing with it awhile

this is actually pretty good, not as good as repairing yourself manually but is a good alternative to repairing yourself manually

code/datums/status_effects/debuffs.dm
Lumipharon: Would heal_overall_damage not work here, or do you specifically want it to be one damage/one limb?

chizzy376:
tm?

code/datums/status_effects/debuffs.dm
QualityVan: It would, but HOD applies as much healing as possible to one limb before moving to the next anyway and the per-tick healing from this is small enough that I'm not too concerned about wasted overflow.

chizzy376:
Please add this to synth too

TiviPlus (approved):

Shootnik01:
Bobots was OP before this pr, now every bobot can hide in closet for half a minute to exit with full health

PR #12135, opened 2023-02-09
Fix Plasma OB detonating in OB room

Dikozda:
## About The Pull Request

Fixs Plasma OB going off in the OB room

this solution works when i test it on a local server, but i have no idea why it works, if someone who know how this stuff works could check it that would be great

also can someone explain to me why this works?
## Why It's Good For The Game

being able to use the plasma OB and not having it blow up in the OB room and basically waste fuel is nice.
## Changelog
:cl:
fix: Plasma OB will now hit the lased position instead of going off in the OB room
/:cl:

### DIFF
--- a/code/game/objects/structures/orbital_cannon.dm
+++ b/code/game/objects/structures/orbital_cannon.dm
@@ -414,11 +414,11 @@
 	name = "\improper Plasma draining orbital warhead"
 	warhead_kind = "plasma"
 	icon_state = "ob_warhead_4"
-	var/datum/effect_system/smoke_spread/plasmaloss/smoke
+
 
 /obj/structure/ob_ammo/warhead/plasmaloss/warhead_impact(turf/target, inaccuracy_amt = 0)
 	. = ..()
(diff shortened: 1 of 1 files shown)


Strayed-Raven:
Probably because its a gas instead of an effect. Like explosion or fire or cluster. Should probably check how CAS does T-gas missiles

Dikozda:
its more the fact that moving a var from outside a proc into inside the proc would somehow fix it that confuses me

TiviPlus (approved):
its cus something about attach() is broken

DeltaFire15:
Seems more to me that it is related to how the holder was the warhead (in the OB room) before, which if an holder exists it short circuits the location var to the holder's turf instead. Now that there isn't an arg given in new (and therefore no holder it is attached to), the location var is used as it is supposed to work.

Lumipharon:
> its cus something about attach() is broken

You made it always default to the holder's location unless holder is qdel'd. This completely invalidates having the location set prior to that check.

PR #12136, opened 2023-02-10
Rownin and flak vest fix

Lumipharon:
## About The Pull Request
Fix an incorrect state name for flak vests which made the jungle variant invisible.

Moved Rownin out of the xenonauten type path as it inherited a bunch of stuff it shouldn't including map colours which is doesn't, also making them invisible.

Also removed a number of unneeded vars from both.
## Why It's Good For The Game
Invisible armour is bad.
## Changelog
:cl:
fix: fixed Rownin and flak vests being invisible in some cases
/:cl:


novaepee:
my bad

TiviPlus (approved):

PR #12141, opened 2023-02-11
Adds the Carlford Armory Union C-1 Garand.

Pariah919:
## About The Pull Request
Bought to you by the Carlford Armory Union, (CAU) the majestic C-1 'Garand' self loader is a fresh and authentic update of an ancient classic from before the Great Nuclear strikes! Get it now for cheap! Limited time only!

image

...What do you mean the most recent shipment was stolen by rebels?! What in tarnation?!
image

(swapping over to non-larp)
Stats are a slightly slower firing DMR, meant to be used by CLF so it has no minimum range.
.8 FD, Aim mode totals up to 2ish seconds. I would give it no aim mode but that might make it useless for marine use.
1 second wield delay and .75 aim slowdown.
75/25/1.25 (Damage/Pen/Sunder)
8 round clips. Works like a magazine, since its an enbloc.
## Why It's Good For The Game
CLF needed some more love, makes sense they'd get super old rifles. I'd keep it CLF only but the entire server would rip my throat out for keeping a true classic out of there hands. (Sprites by Crow, by the way. Other than the magazine, that's me.)
This is a seasonal, so not really gun bloat to the max, I could make it CLF only if it feels like it treads too much on other seasonals that already exist.
## Changelog
:cl:
add: Added new C-1 'Garand' SLR. Brought to you by Crowford Armory Union! Mostly for CLF, its in the WW2 seasonal for marines. And req. 
/:cl:


GreatKingWombat:
There aren't really any good seasonal DMRs, are there. This should be fun.

Kuro020 (approved):

Lumipharon:
Crowford is very much a player reference.

PR #12146, opened 2023-02-11
Spray nozzle nerf

Lumipharon:
## About The Pull Request
Reduced the flame duration for the spray nozzle to 30%, down from 80%.

This makes the spray nozzle significantly less practical for burning down walls, and reduces its area denial ability.

Also made the funny admeme nozzle actually do something more than the normal one.
## Why It's Good For The Game
Less oppressive spray nozzle.
## Changelog
:cl:
balance: Reduces the duration of fire produced by spray nozzles to 30% of normal
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1549,17 +1549,16 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	name = "spray flamer nozzle"
 	desc = "This specialized nozzle sprays the flames of an attached flamer in a much more broad way than the standard nozzle. It serves for wide area denial as opposed to offensive directional flaming."
 	icon_state = "flame_wide"
-	range_modifier = 0
 	pixel_shift_y = 17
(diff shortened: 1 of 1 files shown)


Zar000:
If wall destruction is the issue you could always lower the wall penetration here.
bild
flamer.dm

FosterGrime-Heymynameslouie:
could also just lower the range at which the flames reach
would make sense and give it more of a niche

normal spray = long range, tight wall removal
spray = short range, wide wall removal

Lumipharon:
> could also just lower the range at which the flames reach would make sense and give it more of a niche
> 
> normal spray = long range, tight wall removal spray = short range, wide wall removal

That was the old wide nozzle, which was basically never used.

FosterGrime-Heymynameslouie:
> > could also just lower the range at which the flames reach would make sense and give it more of a niche
> > normal spray = long range, tight wall removal spray = short range, wide wall removal
> 
> That was the old wide nozzle, which was basically never used.

ah

PR #12147, opened 2023-02-11
Stacking fire

Lumipharon:
## About The Pull Request
Makes different applications of fire stack on a turf.

Currently, applying fire to a tile simply deletes the old one, even if it was strong than the new one.

Now it will combine the firelevel and burn level of the two before deleting the old one (capped at double the new fire's strength to avoid stacking 500 incend grenades to make insta death fire).
## Why It's Good For The Game
Applying flame sometimes making fire weaker is bad.
More fire gud.
## Changelog
:cl:
balance: Applying fire to an already burning turf now increases the strength of the fire instead of removing the old fire
/:cl:


Oscari-pog:
Does this do anything visually such as making the flames bigger

chizzy376:
Does this mean that having 2 x-fuel fire stacks on a fire melts benos turbo fast due to applying 2 ticks of fire stack dmg at once?

Lumipharon:
Flame sprites already change based on firelevel, so yes.

This will still delete the old fire object, so you're not going to walk onto a turf and get burnt 5 times or anything.

iforgotmeword:
well time to make hell smoke everyday now

ThelastArchiver:
does the same apply to acid now too?

code/modules/projectiles/guns/flamer.dm
```diff
@@ -402,11 +402,12 @@
...
 	//extinguish any flame present
-	var/obj/flamer_fire/F = locate(/obj/flamer_fire) in src
-	if(F)
-		qdel(F)
-
-	new /obj/flamer_fire(src, fire_lvl, burn_lvl, f_color, fire_stacks, fire_damage)
+	var/obj/flamer_fire/old_fire = locate(/obj/flamer_fire) in src
+	var/obj/flamer_fire/new_fire = new(src, fire_lvl, burn_lvl, f_color, fire_stacks, fire_damage)
+	if(old_fire) //we combine their stats, but limited to x2 so you can't stack it to absurd levels
+		new_fire.firelevel = min(new_fire.firelevel + old_fire.firelevel, new_fire.firelevel * 2)
```
TiviPlus, line 408: Why are you creating/deleting fire instead of just adjusting the existing fire

TiviPlus (changes requested):
Probably a terrible idea but w/e

code/modules/projectiles/guns/flamer.dm
Lumipharon: done

code/modules/mob/living/living_defense.dm
```diff
@@ -120,6 +120,7 @@
 		on_fire = TRUE
 		RegisterSignal(src, COMSIG_LIVING_DO_RESIST, .proc/resist_fire)
 		to_chat(src, span_danger("You are on fire! Use Resist to put yourself out!"))
+		visible_message(span_danger("[src] bursts into flames!"), isxeno(src) ? span_xenodanger("You burst into flames!") : span_highdanger("You burst into flames!"))
```
TiviPlus, line 123: Stinky issrc in useless ternary

TiviPlus (approved):

PR #12155, opened 2023-02-12
More fire cleanup and rav fire act actually working edition

Lumipharon:
## About The Pull Request
Makes the rav fire proc the right thing. I am smoothbrain, but also firecode sucks.
Also added a 1 second CD so you can't just move back and forth mega fast for quick plasma.

Organised some fire procs a bit more so they're not all just stuffed in the flamerthrower gun dm for whatever reason.
## Why It's Good For The Game
Fix my own mistakes
Cooldown is good
Slightly better organised code
## Changelog
:cl:
fix: Fixed ravager fire interaction and added a one second cooldown
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -201,7 +201,7 @@
 	if(status_flags & (INCORPOREAL|GODMODE)) //Ignore incorporeal/invul targets
 		return
 	if(hard_armor.getRating(FIRE) >= 100)
-		to_chat(src, span_warning("Your suit protects you from the flames."))
+		to_chat(src, span_warning("You are untouched by the flames."))
 		return
 
 	take_overall_damage(rand(10, burnlevel), BURN, FIRE, updating_health = TRUE)
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -195,6 +195,23 @@
 	adjust_fire_stacks(rand(1,2))
 	IgniteMob()
 
+/mob/living/flamer_fire_act(burnlevel)
+	if(!burnlevel)
+		return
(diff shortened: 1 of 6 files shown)


QualityVan:
I thought you meant to put this on the life on fire stuff but that's handle_fire

Lumipharon:
> I thought you meant to put this on the life on fire stuff but that's handle_fire

Truly fire code is the worst.

DeltaFire15:
The previous PR did indeed make this seem like it was meant to be when "on fire" not when "*currently in the process of being set on* fire". But yeah handle_fire does that (and also runs on a two second interval because it's handled by life which means no cooldowns needed)

Lumipharon:
I genuinely don't remember which one I intended. I did another pr right after to unfug the procs of fire turfs itself.

The issue with tying this to being on fire instead of standing in fire is you can just get lit then run and hide in a corner for free full plasma anyway.

code/modules/mob/living/living_defense.dm
```diff
@@ -195,6 +195,23 @@
 	adjust_fire_stacks(rand(1,2))
 	IgniteMob()
 
+/mob/living/flamer_fire_act(burnlevel)
+	if(!burnlevel)
+		return
+	if(status_flags & (INCORPOREAL|GODMODE)) //Ignore incorporeal/invul targets
+		return
+	if(hard_armor.getRating(FIRE) >= 100)
+		to_chat(src, span_warning("Your suit protects you from the flames."))
```
TiviPlus, line 204: on living
"suit"

TiviPlus (changes requested):
rest ok I guess

code/modules/mob/living/living_defense.dm
Lumipharon: "your skin suit protects you from the flames"

PR #12156, opened 2023-02-12
Reduces bonus time for capturing a sensor tower

Lumipharon:
## About The Pull Request
Reduced the extra time put on the clock for capturing a tower to 3 minutes instead of 5.

The time is a little too generous now that the clock pauses during activation.
## Why It's Good For The Game
Bit less tilted towards marines.
## Changelog
:cl:
balance: HvH: Capturing a sensor tower now adds 3 minutes to the clock instead of 5
/:cl:


Tailiik:
SC is cool but it would be nice if SOM had a win condition too...neutral king of the hill instead of defense only king of the hill might be better

PR #12160, opened 2023-02-12
Psychic Whisper and Egg Laying added to Queen

MiniMeatwad:
## About The Pull Request

Same as #11826 : 
Simply gives the Queen the ability to lay eggs and use psychic whisper again.
## Why It's Good For The Game

Same as #11826 : 
" Laying eggs was removed in early 2021 for being useless which I don't agree with. They work fine as traps and as a way of drawing marine fire, and they can be very helpful for the carriers and defilers in your life. Also they are thematically appropriate and give queen more to do.

Queen also used to have psychic whisper but it was removed at some point. It seems pretty silly for the leader role to not have it when Warlock, Shrike, and Hivemind do. It's a little bit of extra flavor for queen and it can be fun to mock the ungas.

Queen at the moment is arguably the weakest caste in terms of gameplay, so hopefully this will make her slightly less painful to play. She definitely deserves a rework at some point, but those take time and skill and this takes two lines of code. "

## Why I Think It's Good For The Game

Anecdote from players that play Queen gathered from in-game: Queen is a boring caste, and it's not so much that the caste is weak, but that it's hard to enjoy and usually results in the player playing too risky and dying.

My Anecdote: Adding Egg laying and Psychic whisper gives Queen some silly shenanigans to pull off while keeping restrained, perhaps reducing the likelihood that the player will do something risky and make Xenos lose their precious Queen pheromones, instead they can sit back and do goofy shit.

## Changelog

:cl:
balance: Egg Laying and Psychic Whisper added back to Queen.
/:cl:


TiviPlus:
A) this doesnt make the caste interesting
B) does not adress the original reason for removal

PR #12161, opened 2023-02-12
TL buff

Naaanii:
## About The Pull Request
Buffs the TL mag sizes to 10 rounds instead of 7. (TL Flak mags inherit this.)
## Why It's Good For The Game
TL is quite underwhelming especially compared to the DMR or even a GPMG with rail scope.
This should make it slightly better to use, without needing an ammo box constantly and without
buffing it's actual damage in combat.
## Changelog
:cl:
balance: Increases TL mag size from 7-10
/:cl:

### DIFF
--- a/code/modules/projectiles/magazines/rifles.dm
+++ b/code/modules/projectiles/magazines/rifles.dm
@@ -253,7 +253,7 @@
 	caliber = CALIBER_86X70
 	icon_state = "tl127"
 	default_ammo = /datum/ammo/bullet/sniper/pfc
-	max_rounds = 7
+	max_rounds = 11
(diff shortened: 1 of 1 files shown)


Oscari-pog:
salt pr?

Pariah919:
TL mag size should probably be ten, or increase ammo box to 55 to reflect it being 11 rounds.

Overall 7 is very steep for your ammo eco, right now you basically are far better off running an ammo box than bringing extra mags with you if you're sniping consistently.

Lumipharon (changes requested):
@Naaanii make it 10

PR #12162, opened 2023-02-13
Adds the V-34 carbine for SOM and some MPI changes

Lumipharon:
## About The Pull Request
Adds the V-34 carbine for SOM use in HvH by standards, engineers and medics. Thanks to @SilencedMP5A5 for the gun sprites.

Added black versions of both the V-34 and MPI, as more modern production versions, with a little more attachment options but mainly for fluff.

Magazines are all cross compatible as they share calibre and bullets. Larger mags have gun handling penalties however, so slapping a large mag into the V-34 for example is often not ideal for how the gun performs, but it does make scavanging for ammo in HvH much easier.

Additionally resprited the MPI mob sprites (and of course added V-34 mob sprites), so they should be a little less uglo.

image

The VX-32 charger was created to fill the carbine niche, but significant supply constraints meant a classic weapon from the era of the Martian uprising was modernised and reintroduced to fill the demand. Also the average SOM fanatic is probably inspired to greater feats of suicidal bravery at the thought holding the legacy of their heroic forebearers from the rebellion.

The V-34 is lightweight with good mobility stats, and excellent DPS as it shares the same rounds as the MPI but with a much better ROF. However its shortened size means it suffers significant falloff, and lacks any kind of underbarrel weapon options as well as a smaller magazine.

The V-21 still offers superior mobility (and against light armoured targets comparable damage), but the V-34 has superior mobility and raw DPS compared to the V-31, at the cost of less flexibility, and being very ammo hungry.

Also fixed a incorrect var autodoc, and fixes a bunch of capitalisations and minor stuff with loadout names.
Veterans and SL's now have the legacy MPI's/V-34's that fluff wise them bringing out the family heirloom to keep up the tradition. While the other classes have the modern production versions, that are functionally slightly better due to a few more attachment options. You get less cool Mars points though.

## Why It's Good For The Game
More variety for SOM is good.
## Changelog
:cl:
add: SOM forces now have access to the V-34 carbine
add: Cross compatible MPI and V-34 styles and magazines
/:cl:


code/datums/quick_load_outfits.dm
```diff
@@ -1141,6 +1141,29 @@
 	H.equip_to_slot_or_del(new /obj/item/explosive/grenade/som, SLOT_IN_ACCESSORY)
 	H.equip_to_slot_or_del(new /obj/item/explosive/grenade/incendiary/som, SLOT_IN_ACCESSORY)
 
+/datum/outfit/quick/som/marine/light_carbine
+	name = "V-34 Light infantryman"
```
XSlayer300, line 1145:
```suggestion
	name = "V-34 Light Infantryman"
```

Lumipharon: I probs have formatted this way across the whole file. I should probably change it though.

Oscari-pog:
Add this to valhalla if you havent please.

Kuro020 (approved):

PR #12171, opened 2023-02-13
Increases passive regeneration on kitted limbs

QualityVan:
## About The Pull Request
A kitted limb will regen for 2 total damage split appropriately between burn and brute, up from 0.5 of each with a 75% chance.
This amount reduces with multiple damaged limbs, but you'll still heal faster with multiple damaged and treated limbs in total.
Specifically, per limb regen is divided by the square root of the damaged limb count. Four total will halve the per-limit amount and so on.
## Why It's Good For The Game
Makes kitting more valuable even when the limb stays treated for a shorter period.
## Changelog
:cl:
balance: Kitted limbs passively heal faster
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life/handle_organs.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_organs.dm
@@ -22,6 +22,8 @@
 	for(var/datum/internal_organ/I in internal_organs)
 		I.process()
 
+	var/multi_limb_regen_penalty = 1 / (max(1, length(get_damaged_limbs(TRUE, TRUE))) ** 0.5) //Per-limb regen decreases with multiple damaged limbs, but slower than linear
+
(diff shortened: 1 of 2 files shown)


Al-1ce:
I like the idea of this, but I don't really feel that it's necessary in the current climate. Bandages and ointment are practically unheard of on the field, people just use kits instead. It doesn't feel right to buff them, at least not to me.

QualityVan:
I'm using kits as a catchall for both advanced kits and the basic supplies, limbs don't differentiate once they're treated.

EvilMelonTheSecond:
A buff of this scale will actually make me take kits over pills. Absolutely worth a test merge at the very least.

chizzy376:
Give this to robots (And synth) too

Lumipharon (approved):

PR #12177, opened 2023-02-15
Minisentry cost buff and Comrade sentry return

Pariah919:
## About The Pull Request
Comrade sentry returns. 
He now uses his proper gauss bullet. So he does 60 damage per bullet rather than a measly 25, I also reverted the stats to the stats of yore., which is .15 burst delay, 10 round bursts and 1 second inbetween bursts.

Minisentry costs 50 points less, the ammo is also much, much cheaper at 25 rather than 100 points.

## Why It's Good For The Game
Dropship turrets are fixed, if you're fighting infront of the dropship expect some pain, but otherwise just bait it.
## Changelog
:cl:
balance: Mini-Sentry now costs 50 points less, ammo costs 75 points less.
balance: DS turrets have 10 round bursts, and .15 burst delay. However there is a ten second delay between every burst.
fix: DS turrets use their proper bullet. They now use gauss bullets, rather than normal autocannon bullets.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1169,8 +1169,8 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 
 /datum/ammo/bullet/turret/mini
 	name = "small caliber autocannon bullet"
-	damage = 15
-	penetration = 30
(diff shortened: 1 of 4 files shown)


Kuro020:
BASED RETURN OF COMRADE SENTRY

Lumipharon:
What does this pr actually do now? Update your desc etc

PR #12179, opened 2023-02-15
Reduces cable coil material cost

Barnet2:
## About The Pull Request
Changes cable autolathe material cost from 50 to 5. 

## Why It's Good For The Game
Makes it more tedious to try and abuse autolathe for materials. I would like to remove the ability to make metal from thin air from autolathe all together, but that is for later. With this material cost, it only takes two thirty stacks to make, for example, mass produced crates; so it is not totally inconvenient for the lesser items.

## Changelog
:cl:
balance: cable mat cost changed to 5
/:cl:

### DIFF
--- a/code/modules/power/cable.dm
+++ b/code/modules/power/cable.dm
@@ -400,7 +400,7 @@ GLOBAL_LIST_INIT(cable_coil_recipes, list(new/datum/stack_recipe("cable restrain
 	var/cable_color = "yellow"
 	var/obj/structure/cable/target_type = /obj/structure/cable
 	var/target_layer = CABLE_LAYER_2
-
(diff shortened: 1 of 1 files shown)


chizzy376:
I disagree with this.
>  I would like to remove the ability to make metal from thin air from autolathe all together, but that is for later

Autolathe is for lathing metal, not for being useless. It's not powercreep - it's been here since the start. It's not unbalanced - it takes time and effort to gather material.

Lumipharon:
Standing in a spit clicking the infinite item then clicking into a lathe is not good or desirable gameplay. I would really prefer the lathe just not be able to accept infinite items all together, as these aren't really ideal fixes.

PR #12210, opened 2023-02-20
Fixes Caste death timers not working

Arturlang:
## About The Pull Request
Um, I forgot to re-implement the actual checking for the death timers, which is like, the whole point of it. Anyway, here you go, It's there, now.
## Changelog
:cl:
fix: Caste death timers now work, whoops.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/evolution.dm
@@ -164,6 +164,12 @@
 	if(CHECK_BITFIELD(new_caste_flags, CASTE_CANNOT_EVOLVE_IN_CAPTIVITY) && isxenoresearcharea(get_area(src)))
 		to_chat(src, span_warning("Something in this place is isolating us from Queen Mother's psychic presence. We should leave before it's too late!"))
 		return
+	// Check if there is a death timer for this caste
+	if(new_caste_type.death_evolution_delay)
+		var/death_timer = hive.caste_death_timers[new_caste_type.caste_type_path]
+		if(death_timer && timeleft(death_timer) > 0)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/evolution.dm
```diff
@@ -164,6 +164,12 @@
 	if(CHECK_BITFIELD(new_caste_flags, CASTE_CANNOT_EVOLVE_IN_CAPTIVITY) && isxenoresearcharea(get_area(src)))
 		to_chat(src, span_warning("Something in this place is isolating us from Queen Mother's psychic presence. We should leave before it's too late!"))
 		return
+	// Check if there is a death timer for this caste
+	if(new_caste_type.death_evolution_delay)
+		var/death_timer = hive.caste_death_timers[new_caste_type.caste_type_path]
+		if(death_timer && timeleft(death_timer) > 0)
```
Lumipharon, line 170: Shouldn't the timer be null anyway if it's reached zero? I don't think you need the timeleft check here.

Lumipharon (changes requested):

code/modules/mob/living/carbon/xenomorph/evolution.dm
Arturlang: Was worried about timeleft runtiming if death timer was null but I can check it if you want

Arturlang: Ah, looks like I misunderstood the question

Arturlang: Tested your suggestion and it looks like it works

Lumipharon (approved):

PR #12214, opened 2023-02-21
Grenades no longer require throw mode to throw

Xander3359:
## About The Pull Request
WE HAVE THE TECHNOLOGY, WE CAN MAKE HIM BETTER.
This makes it so that grenades throw on /afterattack , instead of just enabling throw mode.
## Why It's Good For The Game
Throw mode is janky, being able to avoid it (while still supporting it) is a better way.
## Changelog
:cl:
add: Primed grenades throw on click regardless of intent.
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -11,10 +11,13 @@
 	flags_equip_slot = ITEM_SLOT_BELT
 	hitsound = 'sound/weapons/smash.ogg'
 	icon_state_mini = "grenade_red"
-	var/launched = FALSE //if launched from a UGL/grenade launcher
-	var/launchforce = 10 //bonus impact damage if launched from a UGL/grenade launcher
+	///if launched from a UGL/grenade launcher
+	var/launched = FALSE
+	///bonus impact damage if launched from a UGL/grenade launcher
+	var/launchforce = 10
 	var/det_time =  4 SECONDS
-	var/dangerous = TRUE 	//Does it make a danger overlay for humans? Can synths use it?
+	///Does it make a danger overlay for humans? Can synths use it?
+	var/dangerous = TRUE
 	var/arm_sound = 'sound/weapons/armbomb.ogg'
 	var/hud_state = "grenade_he"
 	var/hud_state_empty = "grenade_empty"
@@ -51,10 +54,12 @@
 		var/image/grenade = image('icons/mob/talk.dmi', user, "grenade")
 		user.add_emote_overlay(grenade)
 
-	if(iscarbon(user))
-		var/mob/living/carbon/C = user
-		C.throw_mode_on()
-
+/obj/item/explosive/grenade/afterattack(atom/target, mob/user, has_proximity, click_parameters)
+	. = ..()
+	user.temporarilyRemoveItemFromInventory(src)
+	forceMove(get_turf(src))
+	src.throw_at(target, throw_range, throw_speed, user, TRUE)
+	playsound(src, 'sound/effects/throw.ogg', 30, 1)
 
 /obj/item/explosive/grenade/proc/activate(mob/user)
 	if(active)
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -56,6 +56,8 @@
 
 /obj/item/explosive/grenade/afterattack(atom/target, mob/user, has_proximity, click_parameters)
 	. = ..()
+	if(!active)
+		return
 	user.temporarilyRemoveItemFromInventory(src)
 	forceMove(get_turf(src))
 	src.throw_at(target, throw_range, throw_speed, user, TRUE)
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -58,10 +58,7 @@
 	. = ..()
 	if(!active)
 		return
-	user.temporarilyRemoveItemFromInventory(src)
-	forceMove(get_turf(src))
-	src.throw_at(target, throw_range, throw_speed, user, TRUE)
-	playsound(src, 'sound/effects/throw.ogg', 30, 1)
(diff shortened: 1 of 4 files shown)


ivanmixo:
They automatically toggle throw mode for you when you activate them, though? So you never have to toggle it yourself and can just click immediately after activation. This is begging for misclicks lol

Xander3359:
> They automatically toggle throw mode for you when you activate them, though? So you never have to toggle it yourself and can just click immediately after activation. This is begging for misclicks lol

They don't turn off throw mode if you ever unequip the grenade. Thats my main annoyance

Oscari-pog:
This is actually terrible ngl

Jalopkomit18:
What happens if you click when the grenade is not primed? Will it still throw, or not do anything like it should? If it doesn't throw when not primed, I see no reason why this shouldn't be merged. Having it toggle throw mode adds a lot of jank. However, if you do toggle throw mode manually with this pr, what happens?

Xander3359:
> What happens if you click when the grenade is not primed? 

It doesnt throw if it's not primed. If you want to throw un-primed grenades, you have to turn on throw mode and toss it yourself.

You can turn on throw mode and throw a primed grenade as well (If it makes you happy)

Lumipharon:
If your issue is with dropping nades not turning off throw mode... why don't you make drop turn off throw mode?

Xander3359:
I would rather divorce throwing things from throw mode to be honest. Things like throwing knives/nades, objects which are intended to be thrown, should just function without needing to fuck around with throw mode. 

Throw mode should be used consciously when you intent on throwing something for any personal reason. The only times you should enter throwing intent is when you intend on throwing something, not because you are forced into throw intent to pick throw a nade.

Primed grenades are always intended on being thrown, if you pick up a live grenade to re-throw it you have to manually re-enable throw mode, but on this build you can simply pick it up and toss it without having to fumble your keys.

Al-1ce:
I am convinced you have never thrown a grenade in this codebase before.

XElectricX:
Great PR. Those that hate it should ask themselves if they think throwing knives should use throw mode as well.

TChalkman:
When you prime a grenade, it toggles your throw mode. It doesn't turn it on, only toggles it. Which causes things like:

-If you prime flares and want to just drop them instead of throw, throw mode will still be on and you'll have to press throw key or click twice.
-If you get tackled, stunned, or dropped... before throwing, throwing mode literally REFUSES doing anything except throwing what's on hand (or do nothing, if the hand is empty). This means you need to click twice on a grenade to grab it. You might also accidentally throw your weapon if you try to use it instead, etc.
-If for some reason you had throw mode already on, priming a grenade will toggle your throwing mode OFF. 

I agree with this PR. There are so many things that could go wrong when using any kind of grenade, and its not like they're so different from throwing knives. Why would you ever want to NOT throw a primed grenade? You aren't going to MELEE someone with it or something. 
And since you need to prime the grenade before throwing, there shouldn't be any issue with accidentally throwing an unprimed nade or anything like that
Half the reason an experienced player wants to use a GL is not because of just the grenade delay, but to avoid getting trolled by the throwing system

Strayed-Raven:
please merge this. im tired of throwing guns in the middle of battle just because i get stunned after priming a nade and then taking my rifle out.

Naaanii:
Disguised nerf for skill issues. So real.

Tailiik:
Ignore the haters, eliminating jank is good.

Notamaniac:
How much jank would it involve to turn this into a preference?

Xander3359:
> How much jank would it involve to turn this into a preference?

Is there any time you have a primed grenade in hand that you wouldnt want to throw?

Al-1ce:
> Is there any time you have a primed grenade in hand that you wouldnt want to throw?

That's not the point. The point is forcing new mechanics that don't actually add anything to the game. Any SS13 veteran will use throw-mode by instinct. 

Don't fix what isn't broken. This PR is retarded.

Xander3359:
> Any SS13 veteran will use throw-mode by instinct.

Any TGMC veteran will just prime and throw, as currently priming a grenade puts you in throw mode.
Gameplay wise, this change should be unnoticed by most of the marines in most cases except the few that are positive.

Losenis:
This is a good change because anyone who hasn't gotten stockholm syndrome'd into the current throwing mechanic would intuitively think "I must throw the grenade after priming it, therefore I must prime the grenade and enable throw mode", which can actually result in you disabling throw mode with the primed nade in your hand. The current mechanic has the possibility of fucking up on people, this pr makes it more reliable by having it throw regardless of intent. There's no real reason to reject this PR other than people who are used to what the old way was given it has zero benefits over this. I don't think even muscle memory would conflict with the PR getting merged since it'd still throw regardless.

Xander3359:
> I don't think even muscle memory would conflict with the PR getting merged since it'd still throw regardless.

This exactly, wether you are on throw mode or not, you will throw the grenade no matter what.

code/game/objects/items/explosives/grenades/grenade.dm
```diff
@@ -51,10 +54,14 @@
...
-		C.throw_mode_on()
-
+/obj/item/explosive/grenade/afterattack(atom/target, mob/user, has_proximity, click_parameters)
+	. = ..()
+	if(!active)
+		return
+	user.temporarilyRemoveItemFromInventory(src)
+	forceMove(get_turf(src))
+	src.throw_at(target, throw_range, throw_speed, user, TRUE)
+	playsound(src, 'sound/effects/throw.ogg', 30, 1)
```
TiviPlus, line 64: replace with /mob/living/carbon/throw_item ?

TiviPlus (changes requested):
Rest lgtm

code/game/objects/items/explosives/grenades/grenade.dm
Xander3359: yeh

TiviPlus (approved):

Al-1ce:
Another bad PR merged. Oh the misery.

Losenis:
>bad pr

absolute state of ungabrains

PR #12216, opened 2023-02-21
PMC updates

Lumipharon:
## About The Pull Request
Updates the PMC ERT loadouts slightly, and changes all the PMC typepaths to pmc.

Across the board PMC have more ammo and grenades, as currently they're actually very limited in supplies.
Non snipers now have their pistol in a pistol holster, and all pistols actually have attachments now.
A slight tweak in medical supplies, they all have an advanced combat injector in their hat, and standards/leader have gauze and ointment for what its worth.

Standards have a magazine belt, nade pouch and an underbarrel flashlight on their guns, which is surprisingly useful.
SG's now have a tac sensor and belt harness (as well as far more ammo) and a mighty two (2) claymores instead of just one.
Sniper have a lot more ammo and some other minor tweaks.
Leaders have more ammo and grenades.

Walance changes:
A couple of very minor changes to some PMC guns.

Sniper: The gun no longer have scatter of 3 for whatever reason, and removing some CM era snowflake recoil stun for non PMC users.
VP78: The burst now has a delay of 0.2. Currently the burst is the same rate of fire as semi auto, so it's just worse since you get scatter penalties.
Removed the burstfire mode from the PMC rifle. Same with the VP78, it was the same rate of fire as automatic, but with burst scatter penalties. I removed it because the gun already has a 50% damage bonus so probably didn't need a buff. Although I did change the burst delay to 0.15 like the base rifle, so if you get a burstfire assembly, you can swap out the GL for slightly better dps.
## Why It's Good For The Game
Refreshes PMC a bit, makes them less terrible after running out of ammo from 5 minutes of fighting.
## Changelog
:cl:
balance: PMC ERT updates slightly with better loadouts, mainly for more ammunition and grenades
balance: VP78 burst fire is no longer just a worse version of semi auto
balance:  PR-412E no longer has a pointless burstfire mode
/:cl:


Al-1ce:
> removing some CM era snowflake recoil stun for non PMC users.

This alone makes this a great PR. +1 from me.

PR #12219, opened 2023-02-22
Specifies that time shift doesnt save you from death or from crit

SuperSlayer0:
## About The Pull Request

Time Shift description now says that it will not save you from being dead or in crit when it's effect ends.

## Why It's Good For The Game

Prevents stupid deaths of new wraith players

## Changelog

:cl:
qol: Time Shift description now says that it wouldn't save the target from death or from being in crit
/:cl:


KrabSpider:
It should also be observed that the ability won't save you from crit, either.

SuperSlayer0:
> It should also be observed that the ability won't save you from crit, either.

Yeah changed the description to say that it also doesn't save you from crit

TiviPlus (approved):

PR #12225, opened 2023-02-23
Fixes a dispenser bug

Lumipharon:
## About The Pull Request
Dispensers will no longer teleport across time and space into your hand when they finish undeploying.

This applies to all deployables, in case you somehow get similar situations in the future.
## Why It's Good For The Game
Teleporting bag of holding bad.
## Changelog
:cl:
fix: fixed dispensers teleporting into your hand when undeploying regardless of distance
/:cl:

### DIFF
--- a/code/datums/elements/deployable_item.dm
+++ b/code/datums/elements/deployable_item.dm
@@ -137,7 +137,11 @@
 	attached_item.toggle_deployment_flag()
 
 	user.unset_interaction()
-	user.put_in_hands(attached_item)
+
+	if((get_dist(deployed_machine, user) > 1) || deployed_machine.z != user.z)
+		attached_item.forceMove(get_turf(deployed_machine))
+	else
+		user.put_in_hands(attached_item)
 
 	attached_item.max_integrity = deployed_machine.max_integrity
 	attached_item.obj_integrity = deployed_machine.obj_integrity


code/datums/elements/deployable_item.dm
```diff
@@ -137,7 +137,11 @@
 	attached_item.toggle_deployment_flag()
 
 	user.unset_interaction()
-	user.put_in_hands(attached_item)
+
+	if((get_dist(deployed_machine, user) > 1) || deployed_machine.z != user.z)
+		attached_item.forceMove(get_turf(deployed_machine))
```
TiviPlus, line 142: Why can you still redeploy even if you're not on the same z/not next to it? should it just be disallowed in the first place

TiviPlus (changes requested):

code/datums/elements/deployable_item.dm
Lumipharon: Because the dispenser does this snowflake 4 second delay disassemble. So you have to be next to it to trigger it, but you can just walk off while it's transforming back into a bag.

TiviPlus: Why not fix the delay not canceling instead

Lumipharon: There's nothing wrong with the delay though, and it's not cancelling, its just making it drop on the floor.
Basically with a dispenser you're pressing a button, and then it disassembles itself (hence the animation). deployable code wasn't made with that in mind though.

TiviPlus: uh ok

PR #12227, opened 2023-02-23
Xeno evolution fixes

Lumipharon:
## About The Pull Request
Probs TM first.
Fixed a few checks that were getting missed when evolving which can lead to multiple of the same leader caste, and other issues like that.

Also reorganised evo code a bit more, as it was still a huge ass single proc, now split into 4 procs.

Added balloon alerts where appropriate, and removed an unneeded child proc.
## Why It's Good For The Game
Bug fixes good
Neater code good
## Changelog
:cl:
fix: fixed some evolution bugs
/:cl:


TiviPlus:
But y make it 4 procs

PR #12229, opened 2023-02-23
Adds a prompt for evolving to hivemind

ivanmixo:
## About The Pull Request

- Adds a prompt upon selecting to evolve to hivemind,

`You are about to evolve into a hivemind, which places it's core on the tile you're on when evolving. This core cannot be moved and you cannot regress. Are you sure you would like to place your core here?`
## Why It's Good For The Game

Should hopefully lessen the funny new player thing of putting your hivemind core right outside the LZ shutters or an otherwise inconvinient/exposed place by giving more info when evolving.
## Changelog
:cl:
qol: Added a prompt for evolving to hivemind which should hopefully help with newer players putting their cores in unintended places
/:cl:


code/modules/mob/living/carbon/xenomorph/evolution.dm
```diff
@@ -99,6 +99,9 @@
 	if(eaten_mob)
 		to_chat(src, span_warning("We cannot evolve with a belly full."))
 		return
+	
+	if(caste_type == /mob/living/carbon/xenomorph/hivemind && tgui_alert(src, "You are about to evolve into a hivemind, which places it's core on the tile you're on when evolving. This core cannot be moved and you cannot regress. Are you sure you would like to place your core here?", "Evolving to hivemind", list("Yes", "No"), FALSE) == "No")
```
QualityVan, line 103:
```suggestion
	if(caste_type == /mob/living/carbon/xenomorph/hivemind && tgui_alert(src, "You are about to evolve into a hivemind, which places its core on the tile you're on when evolving. This core cannot be moved and you cannot regress. Are you sure you would like to place your core here?", "Evolving to hivemind", list("Yes", "No"), FALSE) == "No")
```
'Its' is the possessive, 'it's' is the contraction.

ivanmixo: i hate english

Al-1ce: I'd suggest replacing "evolve into **_a_** hivemind" with "the" hivemind, since it is a singular entity.

ivanmixo: Egh its an instance of a manifestation of the hivemind the way I see it

Al-1ce: > Egh its an instance of a manifestation of the hivemind the way I see it

I thought this too for a while, but it would mean that each time the hivemind dies and respawns, it's a different "entity" of sorts, which doesn't really make sense. It's the same Hivemind, just being manifested across various places, various times.

It also gives  the impression of there being "multiple" hiveminds, whereas there is only one within the hive itself. And for the role to be singular, as well.

FosterGrime-Heymynameslouie:
also stops you from accidentally evoing to hivemind when you wanted to go hivelord

very cool pr spdeemergh instantly

Lumipharon:
This code was only just refactored to not have snowflake type code. Also I'm not sure why this isn't showing as a merge conflict, as this code was reorganised so this would get spammed twice.

ivanmixo:
All you did was split up the proc into 3 for whatever reason but sure I'll pull in master

TiviPlus (approved):

TiviPlus:
Caste evolution checks is checked twice

TiviPlus:
It was fine before since it was still in do evolve

PR #12246, opened 2023-02-26
Acid now melts everything in the container that has gotten melted minus mobs

ivanmixo:
## About The Pull Request

- Makes it so that if you melt a container the things inside the container melt too, with the exception of mobs
- So, if you melt a medibelt you don't have to melt all the pill bottles inside then all the pills inside said bottles (guh)

(this is tehnically a partial revert of #11334)
## Why It's Good For The Game

Xenos having to babysit containers til they're full melted then babysit the contents to melt them as well (while having to spend absurd amounts of plasma to melt all the items inside as well, which makes it take even longer) is not fun!
## Changelog
:cl:
balance: If you melt a container the things inside the container melt too, with the exception of mobs
/:cl:

### DIFF
--- a/code/game/objects/effects/aliens.dm
+++ b/code/game/objects/effects/aliens.dm
@@ -188,20 +188,24 @@
 			WF.deconstruct(FALSE)
 
 		else
-			if(acid_t.contents.len) //Hopefully won't auto-delete things inside melted stuff..
-				for(var/atom/movable/M in acid_t.contents)
-					if(acid_t.loc) M.forceMove(acid_t.loc)
-			qdel(acid_t)
-			acid_t = null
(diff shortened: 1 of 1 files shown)


Shootnik01:
Why are mobs feel ok after their box was melted? Make mobs taking damage from acid, with 3 different stages (10 burns, 30 and 50 for example), depends on acid type. Or one total damage

Jalopkomit18:
Very much needed. As much as I loved how my ammo box being acided was just an inconvenience most of the time, it is really lame.

Autisem:
> Why are mobs feel ok after their box was melted? Make mobs taking damage from acid, with 3 different stages (10 burns, 30 and 50 for example), depends on acid type. Or one total damage

the only moment this would come into play is aciding a already dead marine in stasis bag to further make revival difficult. I dont see that as a good idea

PR #12271, opened 2023-02-27
Baldur buff

Al-1ce:
## About The Pull Request

Increases the ``light_mod`` of Baldur by two.

## Why It's Good For The Game

Baldur has been outclassed by every other module for a while now. Some people don't even know it exists. Nobody is willing to sacrifice additional armor or a shield for something that can be supplemented with a single flare or two. This PR aims to change that. 

_Ideally_, I'd like this to be TM'd to see how it would work in actual gameplay, and see if things need adjusting further from there.
## Changelog

:cl:
balance: Baldur provides more light.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -16,7 +16,7 @@
 	icon_state = "mod_lamp"
 	item_state = "mod_lamp_a"
 	slowdown = 0
-	light_mod = 4 /// The boost to armor shoulder light
+	light_mod = 6 /// The boost to armor shoulder light
 	slot = ATTACHMENT_SLOT_MODULE
 	variants_by_parent_type = list(/obj/item/clothing/suit/modular/xenonauten = "mod_lamp_xn", /obj/item/clothing/suit/modular/xenonauten/light = "mod_lamp_xn", /obj/item/clothing/suit/modular/xenonauten/heavy = "mod_lamp_xn")
 


maikilangiolo:
There is a hard limit on light radius of 6 tiles max, since light isn't produced by objects not rendered in. Isn't baldur at the max already? What does this accomplish

Al-1ce:
> There is a hard limit on light radius of 6 tiles max, since light isn't produced by objects not rendered in. Isn't baldur at the max already? What does this accomplish 

If you took 5 seconds to look at the code, you'd know Baldur was at four. It will be at six with this PR.
As for the hard limit, I think I know a way to skirt around it, if need be.

Lumipharon:
> > There is a hard limit on light radius of 6 tiles max, since light isn't produced by objects not rendered in. Isn't baldur at the max already? What does this accomplish
> 
> If you took 5 seconds to look at the code, you'd know Baldur was at four. It will be at six with this PR. As for the hard limit, I think I know a way to skirt around it, if need be.

Light range is capped. This increases the light range of armour, which Baldor already causes to exceed the cap. So uh, this does nothing.

Lumipharon:
this is just a modifier, its changing the light range of the armor, which is capped at 6, but is already 4 or something.

Al-1ce:
> this is just a modifier, its changing the light range of the armor, which is capped at 6, but is already 4 or something.

..Okay. So, does buffing it to 6, from 4, actually do anything? I'm looking for simple clarification here. No need to be esoteric and vague about it.

PR #12279, opened 2023-02-28
Lowers flaregun minimum range. Rearranges flare pouch

Jalopkomit18:
## About The Pull Request
Lowers flaregun minimum range to 2. I wanted to remove its min range entirely, but I think there is a global check to see if it is at least two tiles away when you are clicking with any gun. If someone knows how to workaround this, please tell.
//inprog
## Why It's Good For The Game
The min range check is for grenade launchers since there's chance of self harm. Flare gun doesn't need that, and its not conducive for its purpose of flaring the screen. If you want to drop a flare right below you, it wouldn't let it. Even though that's a perfectly reasonable request. //inprog flare pocuh
## Changelog
:cl:
qol: Flare gun minimum range lowered to 2
qol: Flare pouch has flare gun in last slot, and will now automatically draw from pouch upon clicking.
/:cl:


Naaanii:
Based af.

Barnet2:
Basado change.

Jalopkomit18:
Two problems right now:

I removed the minimum range check that all other grenade launchers have for the flare gun. However, there's still the gun-wide check that doesn't let you shoot under you or in the immediate 6 tiles around you. Xander suggested making it skip adjacent check but I don't know what that entails.

Can't figure out how to reference flaregun weight class in my current rendition of code. As a result im using a hard coded value as placeholder. Also I don't know why even with hard code, the order of the items in the pouch is off. I want the flare gun to be the last item in the pouch so you can draw it instantly upon clicking the pouch. But its not the case

Xander3359:
> Is there a way you can make it attempt to draw flaregun first, regardless of item order, then draw flares? I thought holsters had this, like db belt.

I mentioned a solution on discord but I don't think you saw it. Will link it here just to be sure.
https://discord.com/channels/498569646014857217/498576390753615882/1080041033141854228

Jalopkomit18:
Closed till I fix my VSC via the nuclear option. I'll probably come back to this though since it's close to done

PR #12287, opened 2023-03-01
TX15 slug buff

Naaanii:
## About The Pull Request
Adds stagger to the autoshottie slug.
## Why It's Good For The Game
Flech superiorly outperforms slugs with the autoshottie.
This buff should help keep it in line with flech without being too oppressive and fulfill the soft cc role of slugs vs flech with cc vs raw damage.
## Changelog
:cl:
balance: Autoshottie slugs now have stagger.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1865,6 +1865,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	hud_state = "rpg_fire"
 	icon_state = "rpg_incendiary"
 	flags_ammo_behavior = AMMO_ROCKET
+	effect_radius = 5
 
 /datum/ammo/rocket/wp/quad/ds
 	name = "super thermobaric rocket"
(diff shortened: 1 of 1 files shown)


maikilangiolo:
The slug was also out performed by the lever action (which also has stagger). Seems like a good PR, time to see it not get merged

Lumipharon:
This makes it perma stagger which is pretty cringe.

EvilMelonTheSecond:
> This makes it perma stagger which is pretty cringe.

I believe T39 with slug already has that but with better stats.

Naaanii:
> This makes it perma stagger which is pretty cringe.

Issue addressed.

Ragecrabbo:
> The slug was also out performed by the lever action (which also has stagger). Seems like a good PR, time to see it not get merged

> I believe T39 with slug already has that but with better stats.

Unlike both of these guns autoshotty lets you use underbarrel attachments like miniflamer or grenade launcher (which already makes it better than either of these imo) along with many other attachments. Additionally T39 has a -30% damage and twice the drop-off of autoshotgun while lever action is lever action. Giving autoshotgun slugs stagger would make it easily the best option of these three in every situation i can think of.

Naaanii:
> > The slug was also out performed by the lever action (which also has stagger). Seems like a good PR, time to see it not get merged
> 
> > I believe T39 with slug already has that but with better stats.
> 
> Unlike both of these guns autoshotty lets you use underbarrel attachments like miniflamer or grenade launcher (which already makes it better than either of these imo) along with many other attachments. Additionally T39 has a -30% damage and twice the drop-off of autoshotgun while lever action is lever action. Giving autoshotgun slugs stagger would make it easily the best option of these three in every situation i can think of.

Autoshottie slugs are half the damage.

Al-1ce:
> The slug was also out performed by the lever action (which also has stagger). Seems like a good PR, time to see it not get merged 

This gun was added _years_ before the lever action was. Slugs still preform better.

PR #12291, opened 2023-03-02
Mateba akimbo nerf (Additional delay)

Pariah919:
## About The Pull Request
Mateba now has 1 additional delay rather than .5, making akimbo basically just give you extra capacity.
## Why It's Good For The Game
Mateba getting any DPS increase from akimbo is insane, mostly because it stuns, staggers and does high enough damage that is mostly balanced by the format it comes out of which is well.. A six round revolver. Akimbo solves the capacity issue with a hefty DPS increase, so remove the DPS increase and it should be fine.
## Changelog
:cl:
balance: Mateba has much more fire delay in akimbo.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/revolvers.dm
+++ b/code/modules/projectiles/guns/revolvers.dm
@@ -187,7 +187,7 @@
 	accuracy_mult = 1.15
 	scatter = 0
 	accuracy_mult_unwielded = 0.8
-	akimbo_additional_delay = 1 // Akimbo only gives more shots.
+	akimbo_additional_delay = 0.9 // Akimbo only gives more shots.
 	scatter_unwielded = 7
 
 /obj/item/weapon/gun/revolver/mateba/notmarine
--- a/code/modules/projectiles/guns/revolvers.dm
+++ b/code/modules/projectiles/guns/revolvers.dm
@@ -187,6 +187,7 @@
 	accuracy_mult = 1.15
 	scatter = 0
 	accuracy_mult_unwielded = 0.8
+	akimbo_additional_delay = 1 // Akimbo only gives more shots.
 	scatter_unwielded = 7
 
 /obj/item/weapon/gun/revolver/mateba/notmarine


Naaanii:
This is kinda weird, considering its a req gun and can't fit mag harnesses so it needs to be spammed out.

Not saying that req guns shouldn't be balanced but for it's price and niche, Mateba is in a good spot and one of the last req weapons that need walancing.

KillerOrcaCora:
"i think it's gotta be said that marines got a hell of a lot more combos and things to abuse than xenos do. Xenos just have one set of abilities per caste, as opposed to the many item and weapon combinations that marines have, so by nature xenos are going to complain a lot more about OP shit, because there is a lot of unintended combos that can occur with marines and their items.
anyways i gotta say that what ofnir is pulling off is generally unfun on account of the xeno side. Most weapons that can deal damage like this are rocket launchers, PB shotguns (only massive damage if unarmored), but then you got shit like revolver man to start, then you got the DB juggling, then you got the machinepistol dual wield, now we got mateba dual wield.

I can say that most of this shit is/was extremely unfun to play against as xeno due to the insane damage you got. Ofnir was/still is a consistent high priority target because of the weapons he wields and the insane damage he constantly outputs because he uses weapon combos that most others don't use. Getting killed in 2 seconds by a weapon that you don't expect to do much damage (IE: not rocket launcher or shotgun) isn't fun, there isn't much warning to that happening other than the fact you got killed by it before and now gotta look out for those things. Damage like this makes sense with RR, sadar, or shotgun against T1s, not against the most heavily armored T3 in the game."

Ragecrabbo:
remove akimbo altogether

Oscari-pog:
> 

Can you really call it a req exclusive gun when Ofnir does this dual mateba on Crash of all gamemodes

Naaanii:
> > 
> 
> Can you really call it a req exclusive gun when Ofnir does this dual mateba on Crash of all gamemodes

Can just remove it from the locker/FC loadout which used to have the 1911 instead of a req sidearm.

Seems like a crash issue more than a balance one because req weapons aren't meant to be in crash to begin with, for the most part at least.

chizzy376:
> > 
> 
> Can you really call it a req exclusive gun when Ofnir does this dual mateba on Crash of all gamemodes

Fc spawns with mateba. There is a free mateba in canter.

Lumipharon:
1 akimbo delay means you're better off just wielding one at a time for less scatter, since you can just store them whereever. Aren't pistols like 0.9 for the same reason?

Strayed-Raven:
> 1 akimbo delay means you're better off just wielding one at a time for less scatter, since you can just store them whereever. Aren't pistols like 0.9 for the same reason?

So go ahead and remove the tip about akimbo mechanic.

SilencedMP5A5:
we allowed a power weapon to be powerful?????

thats so dumb nerf that shit right now this instant

makes the game unplayable smh smh

Lumipharon (approved):

PR #12295, opened 2023-03-02
Adds MBX to standard marine vendors, rebalances MBX SABOT

Pariah919:
## About The Pull Request
1:1 adds MBX to marine vendor, buckshot and gun handling stats unchanged as I do not believe they require any changes.

SABOT has been reworked to deal slowdown on hit, it has lost damage in exchange for that, making it a unique chasing gun.
SABOT damage stats are 30/40/1 and 1 slowdown on hit. Old stats are 50/40/3 and no slowodwn.

The nomeculture change is that it is now a "Light Shotgun" rather than just shotgun to better represent its role.
## Why It's Good For The Game
The MBX fills the role and niche of a light chasing shotgun, it has good CQC statistics, low slowdown, buckshot for CQC, and SABOT for midrange combat even if it is rather poor at anything but pen, this gun is focused around a chasing and mobile playstyle compared to something like the T-35, which serves a more devastating CQC role.

Though I'd say the closest cousin would be the autoshotty which is more focused around the midrange/support shotgun style with flechette/slugs which deal solid midrange damage and support respectively.
## Changelog
:cl:
add: Added the MBX lever action to marine vendors.
balance: MBX sabot reworked, it now slows down on hit but has lost damage and sunder for that. 
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -63,7 +63,6 @@
 			/obj/item/weapon/gun/shotgun/pump/lever/mbx900 = -1,
 			/obj/item/ammo_magazine/shotgun/mbx900 = -1,
 			/obj/item/ammo_magazine/shotgun/mbx900/buckshot = -1,
-
 		),
 		"Machinegun" = list(
 			/obj/item/weapon/gun/rifle/standard_lmg = -1,
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -265,6 +265,9 @@
 			/obj/item/weapon/gun/rifle/standard_autoshotgun = -1,
 			/obj/item/ammo_magazine/rifle/tx15_flechette = -1,
 			/obj/item/ammo_magazine/rifle/tx15_slug = -1,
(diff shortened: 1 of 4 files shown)


Naaanii:
Where niche?

Al-1ce:
"Where niche?" The bald, glassy-eyed maintainer says, somehow incapable of understanding the reason this gun (or any other ) was added to req in the first place.

Jokes aside, I like this. The MBX is cheap as shit, but came with the drawback of sabots being unpurchasable/admin only outside of the 2 that spawn in the vendor. 

Given the unique role the MBX preforms compared to many other guns, especially with the change to Sabot, I think it might be a good idea to put it (and the ammo, respectively) in the specialized tab, with a limited quantity. The ability to chase and inflict hard slowdown shouldn't be something _every_ marine ought to get for free roundstart.

FIXTHERADBOXPLEASE:
It definitely has a niche, the only concern is that it might be a powerful one. Chasing is how the top-tier Marines tend to secure kills.
Still, like I said in the last MBX PR, it's not going to be much more oppressive than an SMG in practice. Which is this shotgun's equivalent for rifles.

Mrrpip:
I do want to point out that in certain cases, while it's still restricted by pump action, it can be used akimbo with 0 recoil.

...On the other hand, MW2 memes.

PR #12297, opened 2023-03-02
Removes Lemoline from some recipes. Doubles Lemoline bottle content.

Naaanii:
## About The Pull Request
Removes lemoline from the recipes of:
Hyronolin
Ryetalyn
Neuraline (As a catalyst)

Removes Synap from all recipes requiring synap to make
## Why It's Good For The Game
Allows doctors and corpsmen to work together and make certain medication for the marine force giving marines that much needed teamwork like a well oiled machine while also letting doctor's make a few better chems and a bit more roundstart (if they even bother that is). Also neuraline requiring jelly AND lemoline is wack and lame despite the potential it has.
Recipes requiring synap and lemoline to make is also wack, so removing synap from said recipes seems fine.

Players doing things to help their team mates is good.
Sweaty powergamer making their own mixes so they can avoid having to rely on team members as much as physically possible is bad.
## Changelog
:cl:
del: Removes lemoline from Hyronolin and Ryetalyn recipes.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/glass/bottle.dm
+++ b/code/game/objects/items/reagent_containers/glass/bottle.dm
@@ -225,7 +225,7 @@
 	list_reagents = list(/datum/reagent/medicine/lemoline = 10)
 
 /obj/item/reagent_containers/glass/bottle/lemoline/doctor
-	desc = "A small bottle. Contains 30 units of lemoline, a reagent used in the creation of advanced medicine."
+	desc = "A small bottle. Contains 60 units of lemoline, a reagent used in the creation of advanced medicine."
 	list_reagents = list(/datum/reagent/medicine/lemoline = 60)
 
 /obj/item/reagent_containers/glass/bottle/doctor_delight
--- a/code/game/objects/items/reagent_containers/glass/bottle.dm
+++ b/code/game/objects/items/reagent_containers/glass/bottle.dm
@@ -226,7 +226,7 @@
 
 /obj/item/reagent_containers/glass/bottle/lemoline/doctor
 	desc = "A small bottle. Contains 30 units of lemoline, a reagent used in the creation of advanced medicine."
-	list_reagents = list(/datum/reagent/medicine/lemoline = 30)
+	list_reagents = list(/datum/reagent/medicine/lemoline = 60)
 
 /obj/item/reagent_containers/glass/bottle/doctor_delight
 	name = "\improper Doctor's Delight bottle"
--- a/code/modules/reagents/reactions/medical.dm
+++ b/code/modules/reagents/reactions/medical.dm
@@ -38,7 +38,7 @@
 /datum/chemical_reaction/ryetalyn
 	name = "Ryetalyn"
 	results = list(/datum/reagent/medicine/ryetalyn = 2)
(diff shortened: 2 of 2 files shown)


Lumipharon (changes requested):
Hyro shouldn't be lemo locked, synap needs to be however.

chizzy376:
Since skill lock chemlab is merged you can just fully remove lemolime while you're at it I think

Xander3359:
Good PR, I might actually bother with chem now that it's not fucked by lemoline.

RaidarV:
> Hyro shouldn't be lemo locked, synap needs to be however.

Can't you get infinite synap through injecting it to beakers through marine med injectors and adding them back to the vendor ?

Naaanii:
> Hyro shouldn't be lemo locked, synap needs to be however.

You already get 2 free per marine, I don't see much issue unless you intend on removing the free ones.

KerbalCitizen:
> > Hyro shouldn't be lemo locked, synap needs to be however.
> 
> You already get 2 free per marine, I don't see much issue unless you intend on removing the free ones.

You could also get two construction pouches (25 sandbags and 5 barbed wire each), both pouches making 10 wired sandbag cades. The 2 pouches from vendor at least have opportunity cost between injectors (synap) and construction.

Derrick2Electricboogaloo:
I would make these mixes and med synth to give to marines, then BM removed the ability without waiting on req to be nice. Good work!

Lumipharon:
> > Hyro shouldn't be lemo locked, synap needs to be however.
> 
> You already get 2 free per marine, I don't see much issue unless you intend on removing the free ones.

Because 2 injectors is not the same as being able to make an infinite amount of free synap mix.

Naaanii:
> > > Hyro shouldn't be lemo locked, synap needs to be however.
> > 
> > 
> > You already get 2 free per marine, I don't see much issue unless you intend on removing the free ones.
> 
> Because 2 injectors is not the same as being able to make an infinite amount of free synap mix.

It's hardly a gamebreaking chem and requires a corpsman to produce. Don't really see the issue nor the need for it to be lemolocked to be a never made chem, but sure.

Lumipharon (changes requested):
Keep the bottle at 30. I don't really see why you've removed the catalyst from neuraline, although I agree with removing the synap from the recipies.

Naaanii:
> Keep the bottle at 30. I don't really see why you've removed the catalyst from neuraline, although I agree with removing the synap from the recipies.

Requiring both Lemo and Larval jelly is lame considering the RNG involved with getting larva to begin with. I can remove the larva requirement, but then it would be too easy to make considering how much potential the chem has. Keeping both makes it almost never made despite the potency.

Lumipharon:
Catalyst just means you need to have the reagent, it's not used up.

Naaanii:
> Catalyst just means you need to have the reagent, it's not used up.

Ohh real, will add it back then.

QualityVan:
Oh yeah also neuraline is busted and this both makes it easier to mix (although you still need larbas) and makes two chems that help neuro be extra free to use (hyro and arith, although the second's not as value) much more available by removing lemolock.

Naaanii:
Closing in favour of #12510

PR #12301, opened 2023-03-02
Revert "Grenades no longer require throw mode to throw"

Al-1ce:
## About The Pull Request

There seems to be a clear divide between the community - those who want this, and those who don't. I've spoken to various people in OOC, and the Discord and it seems to be a wider consensus that this is not a well-received change, partially to it being forced on everyone without a preference option available. 
## Why It's Good For The Game

Using throw mode for grenades is something that I find ridiculous to remove. It's one step in the hole of removing or classic mechanics, such as intents, or rendering them mostly useless at a minimum.

The primary reason I could find for this being merged was people accidentally throwing their guns, which is a massive skill issue. Not only for not having a mag-harness, but for doing that to begin with.

## Changelog

:cl:
del: Removed obtuse throwing mechanics.
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -11,13 +11,10 @@
 	flags_equip_slot = ITEM_SLOT_BELT
 	hitsound = 'sound/weapons/smash.ogg'
 	icon_state_mini = "grenade_red"
-	///if launched from a UGL/grenade launcher
-	var/launched = FALSE
-	///bonus impact damage if launched from a UGL/grenade launcher
-	var/launchforce = 10
+	var/launched = FALSE //if launched from a UGL/grenade launcher
+	var/launchforce = 10 //bonus impact damage if launched from a UGL/grenade launcher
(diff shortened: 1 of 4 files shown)


TiviPlus:
Salt revert, you need to discuss this with maintainers beforehand.
As a side note "Soul" is not accepted as an argument as for why its good for the game

Al-1ce:
> Salt revert, you need to discuss this with maintainers beforehand. As a side note "Soul" is not accepted as an argument as for why its good for the game

It was already discussed in the original PR as to why this was not a good thing, but.. muh maintainer

Xander3359:
Lmao

Joe13413:
> complains about grenades not needing throw mode to throw being forced on everyone, as there was no preference option
> tries to force needing throw mode to toss nades on everyone, as there is no preference option
bro.

Al-1ce:
>tries to force needing throw mode to toss nades on everyone, as there is no preference option


I don't see how returning a mechanic to it's original state is "forcing" it on someone. With that logic, every mechanic we have is forced onto the player. 

It matters which came first, for something to be "forced"

PR #12313, opened 2023-03-03
Surgically removed limbs are re-attachable.

Xander3359:
## About The Pull Request
This makes arms and legs surgically re-attachable.
## Why It's Good For The Game
Once, as a field surgeon, I had a marine needing a left leg. I had the idea to remove the left leg of a DNR marine to be able to attach it to the living one. This did not work due to not being able to attach biotic limbs.
So to resolve this, if you surgically remove a limb, it gives you a re-attachable limb. 
IF YOU GET DELIMBED YOU CANNOT REATTACH IT, this PR doesnt touch that.
## Changelog
:cl:
add: surgically detached limbs are now reattachable.
/:cl:


Edelor68:
Makes logical and in game sense, but I am worried that this only makes shipside med bay more obsolete and just encourages more groundside surgery. Medical officer can already be a boring job with long stretches of just nothing happening. I fear this could take a lot of enjoyment out of one job for a small qol for another one.

Xander3359:
This change is more for Medical doctors that choose to deploy to fob surgery. If the only doctor goes to fob to perform surgery, there's usually a bunch of DNR marines piling up, but nobody shipside available to print limbs to send them down. Granted, a synth/researcher can also utilize this for field surgery, but corpsmen are not likely to be performing surgery at all, so they would likely just medevac as usual.

QualityVan:
This renders all other surgeries useless for arms and legs, because it's both faster and safer to amputate then reattach and it fixes bones/IB just as well.

Losenis:
"This renders all other surgeries useless for arms and legs"

Are we picturing the same thing? It's infinitely faster to do the 4 steps for bonefrac or even increase it to 5 for IB too rather than grab a random body IF THERE EVEN IS ONE YET, put him on a roller bed, amputate, remove roller bed, put the unga on it, amputate his perfectly healable limb, do the surgery steps that will also not be muscle memory'd due to the rarity of that surgery before this PR, and give him a new limb.

And this is withot knowing if said limb is also damaged to shit, because I'm not sure if the mob will inherit the damage the limb had beforehand, at which point you'd either need to do the healing and surgeries again which beat the point of that, or preemptively heal the corpse's limbs which, again, would still need as much effort if not more because you'd be battling infections with suture due to a corpse not processing chems. I imagine that if it doesn't, there'll inevitably be a PR down the line to avoid the whole "remove the guy's limb and attach it again" cheesing anyway.

QualityVan:
I'm saying this specifically because you don't need to grab a random body. Take the guy who came in with a broken arm and IB, amp the limb, reattach it, and you've fixed both in two surgery steps (plus healed half the brute/burn on the arm, but that's pretty insignificant).

Losenis:
It doesn't actually inherit the damage that the limb previously had? Sounds like something that should be changed.

I still feel the benefit would be marginal at best given the amputation has a particularly lengthy timer, on top of having to go and pick up the limb wherever it flies off as well as dropping the guy's gloves that'd make it pretty annoying to be on the receiving end of. I don't imagine it being usable in clutter central like tadpole or close to the frontline either, so you just saved... 2 seconds of surgery, at the theoretical maximum speed achievable? 3? It's not like you can carry only the tools to do that surgery alone given you'd also need to be prepped for chest and head issues.

Derrick2Electricboogaloo:
I really do not see why you are not allowed to reattach delimbed limbs. Just have it reapply the damage when reattached so they have to treat that.

Lumipharon:
You can already attach biotic limbs from medical, this just leads to exploits and ignoring medbay.

Derrick2Electricboogaloo:
> You can already attach biotic limbs from medical, this just leads to exploits and ignoring medbay.

Exploits? What exploits? Also you can still ignore medbay by just printing limbs and keeping them on hand or close by.

Losenis:
>Calls exploits
>Closes the PR
>Refuses to elaborate further
>Leaves

Sasuga maintainers

PR #12314, opened 2023-03-03
Polyhexanide bottle has 60 units

Xander3359:
## About The Pull Request
Updates polyhexanide bottle to have 60 units of reagent.
## Why It's Good For The Game
Every other bottle in the Nanotrasen Medplus has 60 units, except polyhex. This looks like an oversight so I'm updating it to the match the other reagent bottles. This is hardly a balance change anyways since you can always just restock the bottles, this just aims at making it less tedious when transferring to pills.
## Changelog
:cl:
balance: Polyhexanide reagent bottles in the medplus contain 60 units like the other bottles in the vendor.
/:cl:


Lumipharon (changes requested):
The description makes it clear its not an oversight, not that it matters. Just update that as well.

Lumipharon (approved):

PR #12317, opened 2023-03-03
Recolor some of the human resprites

ClosetedSkeletonGH:
## About The Pull Request

In the most recent human PR, black people got removed, this PR seeks to fix that with some new palettes to make them match the new style. 

I mean look at this (first is old, second one is the new ones, third through sixth are the new ones for coastal african, central african, and australasian)

image
(Theyre not actually smaller, these pics are just at different scales)

## Changelog

:cl:
imageadd: Increases darkness of several race choices
/:cl:


Kuro020 (approved):
I was gonna make a joke but then I would get canceled

Ragecrabbo:
Can you make asian and causcasian people look less corpse colored aswell?

ClosetedSkeletonGH:
> Can you make asian and causcasian people look less corpse colored aswell?

Tbh the people coloring scheme has always been an issue, its like, either very color burnt or like you said corpse colored.

Id rather just have this PR be its own purpose rn because a PR named "rise of the ubermensch" removing black people is kinda :/

There is also the whole issue of like, 1500 different european races as options that are all a barely noticeable tone shift and then theres "Central African" which with the way some of them are named it feels a bit race science-y

I think a whole replacement of the ethnicity system would be better

PR #12330, opened 2023-03-05
Tweaks the Garand's reciever to eject clips after the final shot is fired.

Mrrpip:
## About The Pull Request

Slightly tweaks the Garand's reciever code, removing a flag that didn't do anything, as it's intertwined with another define that the Garand lacks and replacing it with a flag that makes it's clip eject after the final round is fired.

This does, however mean you can't +1 the rifle.

## Why It's Good For The Game

I mean, it is a tad bit awkward to get the Garand Ping before it's actually empty. Though the PR isn't perfect, as you can't +1 it any more. But hey, if it doesn't get merged, at the very least attention will be brought to it!

## Changelog
:cl: Skye
fix: Removed a flag that doesn't do anything from the C1 Garand
fix: The Garand's clip now pings out after it's empty!
balance: ...And by extension the Garand cannot be +1ed. Sorry!
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1757,7 +1757,7 @@
 	)
 
 	flags_gun_features = GUN_WIELDED_FIRING_ONLY|GUN_AMMO_COUNTER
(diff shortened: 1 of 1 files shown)


maikilangiolo:
The garand (irl) isn't 8+1. It's just 8 due to how the en bloc and reloading works, so if anything, the PR is perfect.

RaidarV:
xenocabal stealth nerfing garand (real)

PR #12341, opened 2023-03-07
Fixed king stagger interaction

Lumipharon:
## About The Pull Request
Fixed stagger not applying against king in some instances where it should be.
King is vulnerable to projectile based stagger when channeling abilities, but the way it was actually checking for stagger was inconsistant.

Also increases the penalty cooldowns for getting interrupted, to make it a bit more beneficial outside of cancel itself.
## Why It's Good For The Game
Bug fix good.
Interrupt more rewarding
## Changelog
:cl:
fix: King is now consistantly vulnerable to stagger when channeling abilities
balance: Interrupting a king while channeling now put the ability on cooldown for 10 seconds instead of 5
/:cl:


Naaanii:
So it WAS bugged!
image

PR #12345, opened 2023-03-07
Valhalla Req Vendor now vends imports.

Xander3359:
## About The Pull Request
Much copypaste, very cool.
Req Vendor has all the imports now under "Imports"
Removes Mosin and it's ammo from imports (Because it's in the weapons vendor)
## Why It's Good For The Game
Testing imports in valhalla.
Cleans up the vendor a bit.
Mosin is a waste of points when it's infinitely vendable.
## Changelog
:cl:
add: all import weapons are in valhalla req vendor now.
del: You can no longer buy mosin from req, because it's already unlimited in the weapons vendor.
/:cl:


Lumipharon:
Shouldn't these be in the req vendor?

Xander3359:
Maybe? But since there's a lot of overlap between seasonal/imports, I thought it made more sense in the seasonal vendor.
If you want me to, I can put them in req.

Lumipharon:
If you only get them from req, then yes they should be in the req vendor. Also you've doubled up a whole bunch of items that are already in the vendor.

Naaanii:
If all seasonals are in the req vendor, would make more sense to put them there in their own tab and make the new seasonal vendor a 'ERT' weapon vendor or something like that.

PR #12348, opened 2023-03-07
Boiler globs do significantly more damage to mechs.

Pariah919:
## About The Pull Request
Boiler globs/lances have a seven times multiplier against mecha targets on direct hit like SoM HEAT (which is 3x), which causes them to do much more damage against mechs on direct hits, however this doesn't account for armor so the actual damage is much lower.

## Why It's Good For The Game
Boilers doing measly damage on direct hits to mechs makes no sense considering the difficulty of landing a shot on a mech directly, but this also adds a soft counter.

## Changelog

:cl:
balance: Mech takes much more damage from direct hits by boiler acid globs and lances.
/:cl:


Kuro020 (approved):
10 times feels like too much but I guess we'll see

Lumipharon:
Local man makes boiler do 750 damage with 60 pen vs mechs

srxtreme:
Based

PR #12357, opened 2023-03-08
Mech sniper changes, slashes damage in half, adds reverse falloff, ups pen a bit.

Pariah919:
## About The Pull Request
Mech sniper stats are 50/45 and with -2.5 falloff, meaning it gains damage per tile.
## Why It's Good For The Game
Mech sniper rework that ideally aims to make the gun an actual sniper rather than a IFF DMR you use in combo with other high damage weapons to just one shot people randomly, the reverse falloff isn't actually as insane as it seems as at 21 tiles you only deal like, 100 damage.
## Changelog
:cl:
balance; Mech sniper reworked, slashed base damage in half, upped penetration by a bit. It now uses reverse falloff meaning it gains damage per tile rather than lose any.
/:cl:


Lumipharon (changes requested):
No to negative falloff, also its literally hard coded to not work.

If you want it to be ineffective at close range, play with min accurate range and its base accuracy.

Pariah919:
back to draw board

PR #12358, opened 2023-03-08
Mech pilot skill and larva strength changes

Pariah919:
## About The Pull Request
Drops MP engi skill to 1 from 3, gives them construction 1, making them fumble and take longer to repair their ride.
Larva strength upped from shipside strength to regular strength.
## Why It's Good For The Game
Mech pilots are far too self sufficent, and should look for engineers to repair them in the frontlines rather than doing it themselves, the construction part is because it'd be weird for them to know how to repair cades but not make them as well as encouraging more emergent use of your skillset outside the mech.

Larva strength is just an oversight fix..
## Changelog
:cl:
balance; Mech pilots are equal larva strength to a Squad Marine, previously they were worth the same as a Shipside role, so less.
balance; Mech pilot engineering skill nerfed to 1 from 3, added 1 construction skill.
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -375,7 +375,7 @@ Though you are a warrant officer, your authority is limited to the dropship and 
 	job_points_needed = 40
 	job_points = 15
 	jobworth = list(
-		/datum/job/xenomorph = LARVA_POINTS_SHIPSIDE,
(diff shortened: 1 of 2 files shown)


Kuro020 (approved):

Autisem:
I feel like MP's should be able to somewhat fix there own mechs, same train of thought that marines have pill packets now. maby 2 instead of 3?

PR #12363, opened 2023-03-09
Increased the chance for robotic arms to drop stuff if their arm is malfunctioning

ivanmixo:
## About The Pull Request

Increases the chance per tick to drop the thing youre holding from 5% to 20% if you have a robotic limb.

(For reference the chance for a normal arm to drop is 15%)
## Why It's Good For The Game

Robot arms cannot fracture, which is a big upside, so they shouldn't get less impact from their arm being hurt on top of that. This also makes robotic limbs have some sort of tradeoff for their unbreakability.
## Changelog
:cl:
balance: Increased the chance for robotic arms to drop stuff if their arm is malfunctioning
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -928,7 +928,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 			owner.emote("me", 1, "[(owner.species && owner.species.species_flags & NO_PAIN) ? "" : emote_scream ] drops what [owner.p_they()] [owner.p_were()] holding in their [hand_name]!")
 			return
 	if(is_malfunctioning())
-		if(prob(5))
+		if(prob(20))
 			owner.dropItemToGround(c_hand)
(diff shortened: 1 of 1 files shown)


novaepee:
This applies to combat robot as well.

ClosetedSkeletonGH:
I don't exactly understand this nerf; robo's aren't posing huge issues and aren't better than humans in many ways (they get better limbs, not being set on fire, and the new 1 health revive). This is nerfing one of their three advantages while adjusting nothing else for combat robo's

Lumipharon (approved):

ivanmixo:
I'm just not a fan of how much suicide rush potential they have and the complete lack of punishment for it. You don't get cloneloss, you don't get fracs/IB, you don't get staggered from pain, you can't get dragged far due to how their (lack of) crit works. 

Aim of this is to deal with the robots being able to dish out full DPS to the moment they die. 20% chance might be a bit much, but it should at least be human levels, I'll see what maints think.

PR #12374, opened 2023-03-10
Repairing shipside gens and removing fuel cells now scales with engi skill

Naaanii:
## About The Pull Request
Title
## Why It's Good For The Game
CSE and STs are the masters of engineering and the masters of ship maintenance. Makes sense for them to do the job more efficiently. Synth too but that's if they're shipside.
## Changelog
:cl:
qol: Makes repairing shipside gens and removing fuel cells scale with engineering skill.
/:cl:

### DIFF
--- a/code/modules/power/fusion_engine.dm
+++ b/code/modules/power/fusion_engine.dm
@@ -247,14 +247,24 @@
 		playsound(loc, 'sound/items/crowbar.ogg', 25, 1)
 		user.visible_message(span_notice("[user] starts prying [src]'s fuel receptacle open."),
 		span_notice("You start prying [src]'s fuel receptacle open."))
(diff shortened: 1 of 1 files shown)


Lumipharon:
Probs just make it scale with skill level instead of just a flat bonus for master level.

Naaanii:
> Probs just make it scale with skill level instead of just a flat bonus for master level.

Done

Derrick2Electricboogaloo:
Nevermind this is not for repairing.

zzurge:
BASED

Derrick2Electricboogaloo:
To be fair, should just be all gen types. There is only 1 ship that has busted ship gens right off the bat, I have never seen them breakdown anymore.

Naaanii:
> To be fair, should just be all gen types. There is only 1 ship that has busted ship gens right off the bat, I have never seen them breakdown anymore.

I'll address groundside in another PR.

PR #12282, opened 2023-02-28
Underbarrel attachments no longer fire on middle click.

Xander3359:
## About The Pull Request
This makes it so that the underbarrel attachment no longer fires when you middle click.
## Why It's Good For The Game
Closes https://github.com/tgstation/TerraGov-Marine-Corps/issues/8182

https://user-images.githubusercontent.com/66163761/221994838-c41cd79e-af35-4ec4-9e7a-8b4fffc3f83c.mp4

This ^^ was stupid.
## Changelog
:cl:
del: You can no longer fire underbarrel attachments with middle click.
/:cl:

PR #12388, opened 2023-03-12
Brings back MMB for special abilities

Al-1ce:
## About The Pull Request

In #12282, middle-click was removed as a means to use underbarrel weapons, or for abilities in general. This PR returns things to the way they previously were. 
<sub>(OF NOTE:) 12282 could not be reverted due to recent changes, this PR was made as a result. </sub>

## Why It's Good For The Game

Middle-mouse is used by a large chunk of the community for abilities. A function of this caliber being completely removed is disruptive and does more harm than good. For those who use right click primarily; imagine if it were removed at the drop of a hat. This is no different. 

In a **_very_** specific context, MMB would cause a player wearing a jetpack to fly into the fire produced from their underbarrel flamer. I find this to be more of a skill issue than a mechanical bug above all else. At a minimum, I don't think removing the ability to use MMB in it's _fucking entirety_ is the right way to go about things.

## Changelog

:cl:
code: MMB returns as an ability key.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -676,11 +676,11 @@
 	var/list/modifiers = params2list(params)
 	if(modifiers["shift"])
 		return
-		
+
 if(modifiers["right"] || modifiers["middle"])
 		modifiers -= "right"
 		modifiers -= "middle"
-		
+
 		params = list2params(modifiers)
 		active_attachable?.start_fire(source, object, location, control, params, bypass_checks)
 		return
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -677,10 +677,9 @@
 	if(modifiers["shift"])
 		return
 
-if(modifiers["right"] || modifiers["middle"])
+	if(modifiers["right"] || modifiers["middle"])
 		modifiers -= "right"
 		modifiers -= "middle"
-
 		params = list2params(modifiers)
 		active_attachable?.start_fire(source, object, location, control, params, bypass_checks)
 		return
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -676,11 +676,11 @@
 	var/list/modifiers = params2list(params)
 	if(modifiers["shift"])
 		return
-
-	if(modifiers["middle"])
-		return
-	if(modifiers["right"])
+		
+if(modifiers["right"] || modifiers["middle"])
 		modifiers -= "right"
+		modifiers -= "middle"
+		
 		params = list2params(modifiers)
 		active_attachable?.start_fire(source, object, location, control, params, bypass_checks)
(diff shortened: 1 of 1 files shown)


Al-1ce:
## Context

Due to the nature of this PR being functionally a revert, I did try and speak with Xander about why removing MMB was bad. 
Their response was... not exactly productive...

``"MMB sucks.
You should not be using MMB for shooting.
Main clicks."``

I don't think this is exactly a healthy base for MMB to be removed. _You_ may not use MMB, or you may not even like it, but disallowing those who do for that reason (or a very, VERY specific incident that can hardly be coined as a 'bug') is unfair and forceful.

Lumipharon:
What exactly stopped working MMB? The PR in question only altered gun code, it shouldn't impact abilities at all, and hasn't as far as I can tell.

cazuccmuss:
Would it be possible to make the option toggleable? I personally hate using RMB for abilities and attachies but there are people who prefer using it apparently and that way both sides get what they want.

ClosetedSkeletonGH:
Id much rather LMB/RMB/MMB be able to be binded on specfic actions. I dont use right click and having flamer extinguisher being moved from a customizable underbarrel bind to an uncustomizable RMB. Same thing with the orbital spotlight, since I dont have an MMB it is very difficult for me to use something that is impossible to change the bind for AND no toggle to change it to shift + click like you can with xeno abilities

Al-1ce:
> What exactly stopped working MMB? The PR in question only altered gun code, it shouldn't impact abilities at all, and hasn't as far as I can tell.


It deleted the interaction between "middle", which disables the ability to use mmb when selecting an underbarrel weapon and firing it. This is toggled with the ''shift/mmb''


>Would it be possible to make the option toggleable? I personally hate using RMB for abilities and attachies but there are people who prefer using it apparently and that way both sides get what they want.


It is already an option in the preferances tab

>Id much rather LMB/RMB/MMB be able to be binded on specfic actions. I dont use right click and having flamer extinguisher being moved from a customizable underbarrel bind to an uncustomizable RMB. Same thing with the orbital spotlight, since I dont have an MMB it is very difficult for me to use something that is impossible to change the bind for AND no toggle to change it to shift + click like you can with xeno abilities

Binding LMB/MMB/RMB on specific actions is functionally no different. I also do not use right click; I abhor it. You can find my comments in the first few right-click PRs showing dismay. The keybind does not need to be customizable since you can already toggle it in prefs.

You say that you don't have a MMB, but that changes nothing in regards to this  PR. With actual shift. It's literally called that. 

Nothing will actually change from what they've been like for 2 years. This PR simply reverts a very poorly thought-out bandaid to a much larger problem; i.e, sets things the way they were.

TiviPlus:
This only affects gun code so idk what you're talking about

Al-1ce:
> This only affects gun code so idk what you're talking about

It prevents using MMB for guns entirely. That is the problem.

PR #12415, opened 2023-03-14
Buffs and QoL improvements to large water and fuel tanks

XElectricX:
## About The Pull Request

- Double integrity so they aren't made of paper
- 10x more reagents inside each tank, that way they are actually useful to keep around
- 0 drag delay so you can move them around faster (THEY LITERALLY HAVE WHEELS)
- The default dispensing amount is 500 so you can fill most things in one click

## Why It's Good For The Game

Fuel tanks are pretty much useless because of how little fuel they have. The other issues like low durability and slowing you down when dragging just add to how awful of an item it is. Water is marginally more useful only because water is depleted slower. These changes should make bringing these tanks into play worth _something_.

## Changelog
:cl:
qol: Fuel and water tanks dispense reagents at a rate of 500 units by default.
balance: Fuel and water tanks: 100 -> 200 HP, 1K -> 10K reagent amount, 0 drag delay
/:cl:

### DIFF
--- a/code/game/objects/structures/reagent_dispensers.dm
+++ b/code/game/objects/structures/reagent_dispensers.dm
@@ -62,7 +62,7 @@
 	desc = "A watertank"
 	icon = 'icons/obj/objects.dmi'
 	icon_state = "watertank"
-	obj_integrity = 200
+	max_integrity = 200
 	amount_per_transfer_from_this = 500
 	possible_transfer_amounts = list(10, 25, 50, 100, 500)
 	tank_volume = 10000
@@ -74,7 +74,7 @@
 	desc = "A fueltank"
 	icon = 'icons/obj/objects.dmi'
 	icon_state = "weldtank"
-	obj_integrity = 200
+	max_integrity = 200
 	amount_per_transfer_from_this = 500
 	possible_transfer_amounts = list(10, 25, 50, 100, 500)
 	tank_volume = 10000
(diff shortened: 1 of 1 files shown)


Xander3359:
Does this make the explosion bigger? Seems to me that this won't fix xenos just stacking up all fuel tanks to fob for shutters to open to nuke all the marines.

XElectricX:
It should not affect explosions, that was a concern I had that I checked. Code just checks how filled it is to determine whether to do a large, medium, or small explosion. What it would change though is fuel leaks. You could probably cover half the map or more with one fuel tank being dragged around leaking since it is based on transfer amount, so you could leak about 5 units of fuel per tile over loooong distances.

Naaanii:
10 times the amount of fuel is quite the massive buff ngl.

Drag speed and integrity makes sense.

Al-1ce:
>You could probably cover half the map or more with one fuel tank being dragged around leaking since it is based on transfer amount, so you could leak about 5 units of fuel per tile over loooong distances.


Brave to imply it wouldn't get blown up immediately.

Lumipharon:
integrity and drag changes are all good, but yeah x10 the capacity on fuel is a very large buff

KerbalCitizen:
Makes CAS PO pretty much never have to worry about fuel. I mean fuel tanks are pretty cheap, so it's not a big deal, but something to keep in mind.

XElectricX:
> Makes CAS PO pretty much never have to worry about fuel. I mean fuel tanks are pretty cheap, so it's not a big deal, but something to keep in mind.

PO gets a ton of fuel tanks because of how little they hold. I foresee marines yoinking some of them now that they aren't useless, but if they're not used as much I can just remove some tanks.

PR #12420, opened 2023-03-14
Battle against req hell: import edition

maikilangiolo:
## About The Pull Request
Makes ammo for import guns both less costly and also more user friendly. Long guns get 3 magazines per order and pistols 5 per order. Instead of lowering the individual cost, I thought it was better if you bought multiple mags to save on clicking, especially considering how ammo intensive small arms are.

Included are a couple price changes:
- scorpion 150 -> 50 (lol, it was 150 points), ammo 50 -> 10
- sawn off 150 --> 100 (50 is too cheap and spammable but 150 too expensive for what it is)
- FAMAS ammo 50 -> 10 (gun stays expensive, but the mags are very small and shouldn't cost as must as HPR ones)
- MBX 150 -> 50 and ammo 50 -> 10 (it's really not that good, it's like a worse repeater)
- lever action 150 -> 50, and removed ammo because .44 packets are free (it's legit a worse repeater)
- moist nugget removed. Funny oversight that's been there for years
## Why It's Good For The Game
No more req hell due to exorbitant ammo price for guns that are in seasonals anyway or not even good (lever action, MBX) or have infinite ammo anyway due to sharing calibers with marine guns (almost every single import gun)
## Changelog
:cl:
balance: rebalanced prices of scorpion, sawn off, mbx and repeater (both guns and ammo)
balance: rebalanced famas ammo cost
balance: all import gun magazines come in 3x (long arms) or 5x (handguns)
fix: fixed the mosin still being in imports
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -1374,7 +1374,7 @@ Imports
 	contains = list(
 		/obj/item/ammo_magazine/smg/mp7,
 		/obj/item/ammo_magazine/smg/mp7,
-		/obj/item/ammo_magazine/smg/mp7
+		/obj/item/ammo_magazine/smg/mp7,
 	)
 	cost = 10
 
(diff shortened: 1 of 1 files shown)


Al-1ce:
>MBX price buff

nuclear level based

Joe13413:
god, im so sad that this is closed.

PR #12422, opened 2023-03-14
Cautery no longer disinfects

QualityVan:
## About The Pull Request
Finishing a surgery no longer automatically sets the limb to entirely germless.
Reduces the germs from using a roller bed by a third.
## Why It's Good For The Game
Automatic disinfection from any surgery means that surgery infection becomes incredibly binary: either you do enough to generate necrosis, or it's irrelevant - and shorter surgeries with proper care shouldn't causes necrosis, groundside or not.
Minor infections are mostly annoying, just messages and a bit of toxloss, and can be ignored entirely by using a chem slot on spaceacillin (one pill of which will fully clear up any existing issue), but still present a potential threat and take time to clear up.
Roller germ reduction brings down the unavoidable germs to make it less likely to cause serious infection if you wash yourself first but still stay high enough to ensure you'll at least have something lingering.
## Changelog
:cl:
balance: Cautery to finish surgery no longer clears germs
balance: Roller bed surgery germs reduced (separate from germ transfer by surgeon, you still need to keep yourself clean)
/:cl:

PR #12423, opened 2023-03-14
Removing necrosis hurts

QualityVan:
## About The Pull Request
Removing necrosis from a limb breaks it and applies half its fracture threshold in brute damage (can't cause delimbs/IB)

## Why It's Good For The Game
Followup to #12422 
With that in place, the option exists to simply ignore any threat of infection for all the surgery you need to do, do necrosis at the end, and be completely fine. This is the best way to prevent that I could come up with.

Fixing a broken limb right after necrosis is easy - you're mid surgery, the incision is already open, and the patient isn't moving around so there's no threat of organ damage. Still, the extra steps are enough to ensure an infection remains active (with a rollerbed - with a surgery table this can be avoided), thus preventing any chance of tactical necrosis for germ treatment.

The brute damage is mostly there for flavor.

## Changelog

:cl:
balance: Surgically removing necrosis breaks the limb. No, it's not the surgeon doing something wrong.
/:cl:

### DIFF
--- a/code/modules/surgery/generic.dm
+++ b/code/modules/surgery/generic.dm
@@ -265,7 +265,6 @@
 	span_notice("You cauterize the incision on [target]'s [affected.display_name] with \the [tool]."))
 	target.balloon_alert_to_viewers("Success")
 	affected.surgery_open_stage = 0
-	affected.germ_level = 0
 	affected.remove_limb_flags(LIMB_BLEEDING)
 	DISABLE_BITFIELD(affected.limb_wound_status, LIMB_WOUND_CLAMPED) //Once the incision is closed, any clamping we did doesn't matter
 
--- a/code/modules/surgery/necrosis.dm
+++ b/code/modules/surgery/necrosis.dm
@@ -38,16 +38,17 @@
 
 /datum/surgery_step/necro/fix_dead_tissue/end_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/limb/affected)
 	user.visible_message(span_notice("[user] has cut away necrotic tissue in [target]'s [affected.display_name] with \the [tool]."), \
-		span_notice("You have cut away necrotic tissue in [target]'s [affected.display_name] with \the [tool]."))
+		span_notice("You cut away the necrosis in [target]'s [affected.display_name], but can't avoid some damage to the surrounding tissue."))
 	target.balloon_alert_to_viewers("Success")
 	affected.necro_surgery_stage = 1
+	affected.createwound(CUT, affected.min_broken_damage / 2)
+	affected.fracture()
 
 /datum/surgery_step/necro/fix_dead_tissue/fail_step(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/limb/affected)
 	user.visible_message(span_warning("[user]'s hand slips, slicing an artery inside [target]'s [affected.display_name] with \the [tool]!"), \
 	span_warning("Your hand slips, slicing an artery inside [target]'s [affected.display_name] with \the [tool]!"))
 	target.balloon_alert_to_viewers("Slipped!")
-	affected.createwound(CUT, 20, 1)
-	affected.update_wounds()
+	affected.createwound(CUT, 20)
 
 /datum/surgery_step/necro/treat_necrosis
 
@@ -63,8 +64,8 @@
 	necro_step = 1
 
 /datum/surgery_step/necro/treat_necrosis/begin_step(mob/user, mob/living/carbon/human/target, target_zone, obj/item/tool, datum/limb/affected)
-	user.visible_message(span_notice("[user] starts applying \the [tool] on the affected tissue in [target]'s [affected.display_name].") , \
-	span_notice("You start applying \the [tool] on the affected tissue in [target]'s [affected.display_name]."))
+	user.visible_message(span_notice("[user] starts applying \the [tool] on the severed tissue in [target]'s [affected.display_name].") , \
+	span_notice("You start applying \the [tool] on the severed tissue in [target]'s [affected.display_name]."))
 	target.custom_pain("Something in your [affected.display_name] is causing you a lot of pain!", 1)
 	target.balloon_alert_to_viewers("Applying...")
 	..()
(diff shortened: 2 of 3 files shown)


Losenis:
>Removing flesh breaks bone

Ok listen I know this is a videogame and sometimes realism needs to be sacrificed but what the fuck man

Also this is only going to get surgeons lynched, I don't expect people to realize what's going on given corpsmen to this day are still unaware you can just defib robots without repairing them despite being a far more common occurence.

Strayed-Raven:
First tweezers can break bones and now just removing dead skin can too. Just admit the medical theme isn't it. And make it not crap.

Al-1ce:
Ask yourself, _"If this was implemented by someone who was not me, would I still like it?"_ 
This is shit. I need not more to elaborate.
-1 from me.

XElectricX:
I thought this was going to just give brute damage and thought "why though", but this breaks field surgery. Organ repair surgery takes long enough to guarantee necrosis when on a roller bed unless you have polyhexanide. Necrosis itself is a pretty strong illness that results from mistakes on the doctor's or medic's part. The punishment already is having to add another 2 steps to surgery (with their own chances of messing up) or chugging down dylovene until you can find a surgeon. I get you want some flavor but you already have that in terms of pain inflicted during surgery, whether it be getting dead skin flayed or mending a broken bone.

Losenis:
>The punishment already is having to add another 2 steps to surgery (with their own chances of messing up) or chugging down dylovene until you can find a surgeon.

A punishment that, as anti-fun as it gets, is applied to a completely different guy rather than the one suffering from skill issues. TGMC already has enough of those monstrosities and getting field surgery, if you can find anyone at all, is running the dice on whether the guy even knows how to do it properly given the stupidly high learning curve it has for people starting out on it given shipside has hardly any opportunities for practicing and people will absolutely refuse to get treated by someone who they haven't memorized the name of as robust in surgery due to the aforementioned issues.

There is a reason why a lot of people who need surgery just rush to the autodoc, it has little (read: still nonzero due to nobody doing gens) risk of _you_ being punished for the skill issues of _another person_.

Xander3359:
Won't this cause the necrosis to come back by the time you finish fixing the bone again? 
That would cause an endless loop of fix bone -> fix necro -> fix bone -> fix necro until the patient gets delimbed or dies.

SzkieletorNutric:
A bit iffy but I like the direction this is going

Pariah919:
tactical application of 1 spaceacillin

chizzy376:
> refuse to get treated by someone who they haven't memorized the name of as robust in surgery due to the aforementioned issues.

This is unfortunately a genuine thing, I made a new static and offered to remove IB and 3 fracs from a guy who started fucking punching me when I wanted to grab him, and he kept saying to just toob him even though I'm way faster >:(

Coastfront:
And this, is as ideal a change as rotted flesh is a pleasent scent.

This hard kills basically any form of field surgery outside of fixing fractures if you don't have poly, and in some cases iirc still would outright kill doing surgery.

This also, makes necroing on vali even more crippling.
One of the prime issues of surgery, is a lack of expended resources for its benefits. It is a harvest with no crops to tend. 
The caveat to this, is the fact that there still has to be a potential fucking benefit. Your other change, causing cauteries to no longer instantly clear infections, is good for adding a resource cost to surgery.
It makes it so that general surgeries, will either have to expend resources to clear an infection, OR they will have to bite the bullet and expend more time to let it necro and clear it. Creating a trade-off: Lessened resource impact, but takes longer if its not urgent, vs can clear surgery more urgently, however it expends a carried resource.

This, instead makes it a "fuck you, if you don't have chems fuck off". There is neither weight nor even another scale to weigh. Surgery becomes a flat resource expense, which, given that spaceacillin can already at times feel scarce if you engage in a more liberal use of it (sutures cause 8 billion infections if you spam them enough), will mean that you either need to sacrifice yet already limited inventory space further, or B. Accept that there is now a far harder limit on how long you can go before medevaccing yourself to refill and podding down, praying that nobody stole your stretcher.

This is a bad change with next to no nuance, and when combined with your previous (decent) change, acts soley as a constricting chain along its desperate to swim heel.


This also makes abso-fucking-lutely zero sense basic logic wise. Generally, stretches of realism are only nice when its BENEFICIAL to gameplay, not when it makes the game objectively less fun.
Stretch of realism in an okay way: we have magical chemicals that can close a fucking bullet wound in 3 minutes
Stretch of realism in an irritating way: Using tweezers on my ribs shattered every single one instantly.

Stretch of realism in a, somehow worse way: Removing dead flesh from my chest ripped my ribs out with it.

If you're dead set on removing necrosis having a cost, either make the surgery step:
A. Longer so that its a greater time investment
B. just deal fucking brute damage, so that there is a time investment OR alternate chem investment spent fixing that.

QualityVan:
There's a lot of comments about bone breaks feeling unfitting. The original idea was just to apply brute, but the problem was that brute doesn't matter at all unless it causes permanent damage somehow - especially when you're already in surgery. So since the purpose of the brute damage was enough to cause a break to occur passively, I just opted to break directly instead and include less brute. As a reminder, if you hit someone in the arm with a spoon twelve times that arm will fracture. Spacemans simply be that way.

I realized, though, that I could avoid any issue of people taking advantage of necrosis removal clearing germs entirely by just making it not do that. With this it only removes most of them, and you'll need to stay on antibiotics for the rest. Funnily enough that's already how autodoc necrosis removal works. There's still some brute damage for flavor since you are cutting out their meats, but it shouldn't be a threat and the second step fixes it.

Losenis:
You could just make necrosis take a large chunk of blood, like 20% or even 25% so that one necrosis removal is already noticeable and has you either out of the fight for a bit to replenish blood or be premedded on inap forever, yet multiple necrosis removals will have you completely ineffective for a while even if you're doped on isotonic + MREs + premed. That way you have your ideal resource cost of blood replenishment that comes in the shape of either MREs or pills.

QualityVan:
Blood's an option I hadn't thought much about, but offhand I'm not wild about tying it in. At the moment blood's very binary about whether you care about it or not since above 60 you have just a bit of oxloss and below it you fall unconscious and suffocate. That means if you're fixing one or two, it's not a real drawback, and if it's more than that it's crippling with no real inbetween. Plus, I do want some nerd who's got all eleven limbs rotted from absurd misuse to be fixable with moderate effort, and 220% blood removed is a lot.

Low blood also strictly forces downtime, which I don't like as much as letting you return to play more quickly post-treatment but temporarily weakened. Admittedly, 'weakened' in the case of this iteration is pretty minimal.

Losenis:
Make it so that bloodloss caps your minimum oxyloss to a certain number? So that instead of being a binary "ignorable/fucking dead" it's cloneloss-lite in that if you fight with low blood you're gonna get fucked up way earlier than someone that isn't a spent ketchup package? That way even sub-90 blood levels can be noticeable but progressively worse from "man I should have a bite" to draining the IV bloodbag machine thrice in a row.

Granted that'd be a pr of its own but hey, the idea is there.

PR #12425, opened 2023-03-14
Spaceacillin doesn't protect against necrosis

QualityVan:
## About The Pull Request
As title, only polyhex works once you're in the right germ threshold. Spaceacillin still helps avoid getting there, of course.
## Why It's Good For The Game
Necrosis is a major kind of long term damage which either arises from long term negligence or a major misplay. Spaceacillin is a basic chem with high availability and no real drawbacks for ODing. The later granting immunity to the former isn't great. Alternatives and why I don't like them:
Make spaceacillin less available - basic chems should be available, and making chems unavailable never actually lasts anyway.
Make ODing on spacea less absolutely nothing - it's a basic chem, akin to bica/kelo, and misusing it shouldn't be any more than a gentle reminder not to do that.
## Changelog
:cl:
balance: Spaceacillin OD doesn't prevent necrosis from forming
/:cl:

### DIFF
--- a/code/modules/organs/limbs.dm
+++ b/code/modules/organs/limbs.dm
@@ -491,7 +491,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 				if (parent.germ_level < INFECTION_LEVEL_ONE*2 || prob(30))
 					parent.germ_level++
 //LEVEL III
-	if(germ_level >= INFECTION_LEVEL_THREE && spaceacillin < 25 && polyhexanide <2)	//overdosing is necessary to stop severe infections, or a doc-only chem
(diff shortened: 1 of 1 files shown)


Losenis:
Remember when the change to reagents in the body was implemented that having more than a number of chems would make you randomly change directions that at worst could have you walk into lava? Yeah that's the reason people tend not to take a lot of chems, spaceacillin being one of them, given infections happen when you're fucked up enough to take a cereal bowl of other chems before that point.

XElectricX:
I was about to call this fake news since I give all my patients spaceacillin and they still get necrosis, but apparently ODing is what stops it. What a weird interaction.

Lumipharon (approved):

PR #12435, opened 2023-03-15
Slight Xeno crash buffs (fix)

Naaanii:
## About The Pull Request
T1 Xenos get the maturity points for evolving in crash.
## Why It's Good For The Game
Since maturity points carry on when evolving, crash has the issue of giving T1s Elder but starting off with 0 maturity points and makes T2s mature by default, also not giving them the points, but T3s take a massive amount of time to reach ancient since maturity changes.

This should rectify that issue and give the missing points when the first evolve is done.

This should help reintroduce the previous interaction without making Xenos too powerful whilst also being a bit of a slight buff when Xenos evolve instantly.
## Changelog
:cl:
balance: T1 Xenos get the maturity points for evolving.
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -178,7 +178,9 @@
 		if(XENO_TIER_ONE)
 			new_xeno.upgrade_xeno(XENO_UPGRADE_TWO)
 		if(XENO_TIER_TWO)
-			new_xeno.upgrade_xeno(XENO_UPGRADE_ONE)
+			new_xeno.upgrade_xeno(XENO_UPGRADE_TWO)
+		if(XENO_TIER_THREE)
+			new_xeno.upgrade_xeno(XENO_UPGRADE_TWO)
 
 /datum/game_mode/infestation/crash/can_summon_dropship(mob/user)
 	to_chat(src, span_warning("This power doesn't work in this gamemode."))
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -176,7 +176,7 @@
 	SIGNAL_HANDLER
 	switch(new_xeno.tier)
 		if(XENO_TIER_ONE)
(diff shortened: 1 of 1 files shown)


MiniMeatwad:
Because you're setting the upgrade value, you're also potentially removing upgrade value. This means crash Xenos cannot benefit from keeping their upgrade progress when they stay as a lower tier.

Naaanii:
> Because you're setting the upgrade value, you're also potentially removing upgrade value. This means crash Xenos cannot benefit from keeping their upgrade progress when they stay as a lower tier.
> 
> Also the guy who did this (Whoever did crash) was highly insistent that it's intentional game design and that it's not a bug, which is funny.

What upgrade value is being removed exactly?

Lumipharon (changes requested):
As QV said, this means reverting back to T1 will REDUCE your maturity points back to TIER_ONE_MATURE_THRESHOLD which is severely nerfing the xeno.

This also removes the bonus from t2 evos so this seems like a funny xeno nerf pr.

Naaanii:
> As QV said, this means reverting back to T1 will REDUCE your maturity points back to TIER_ONE_MATURE_THRESHOLD which is severely nerfing the xeno.
> 
> This also removes the bonus from t2 evos so this seems like a funny xeno nerf pr.

I literally tested this and it doesn't

It doesn't remove the T2 bonus because it grants the maturity points required for T1 so when Xenos evo into T2, they're mature so it changes nothing.

Even regressing to larva and evoing back into a T1 does not reset your maturity points.

Lumipharon:
Oh cool, turns out finish_evolve has a hard check to stop maturity from decreasing. Its still bad to just rely on that though, as regardless you're just hard setting the maturity points which can mean decreasing it. Better to have a check here as well. i.e. = max(current amount, TIER_ONE_MATURE_THRESHOLD )

Lumipharon (approved):

PR #12436, opened 2023-03-15
Xenos always start at ancient and some rebalancing. (Nukewar only)

Naaanii:
## About The Pull Request
Xenos always start at ancient.
Silo cost increased.
Primordial cost increased.
Mat towers removed.
Evo timer for T1>T2 doubled.
Evo timer for T2>T3 increased by 50%.
Queen death timer doubled.
Hivemind collapse timer lowered to 2 minutes.
## Why It's Good For The Game
Xenos often have to wait to fight at full strength and risk fighting not at full power with the risk of being squishier and dealing less damage. Young-Mature T3s are nearly worthless but evo towers help in that regard, however it doesn't make it any better or more intuitive.

This aims to make all Xenos start at full power while making it slightly harder to obtain higher tiered castes and punish Xeno deaths slightly more but not keep it as bad but also makes Xenos heavily rely on caste specialization making them commit slightly harder.

I'll actually look into not giving this to Shrike to avoid instant ancient delay shenanigans.

I should probably cap evo towers too but this will be interesting to TM and I can clean out maturity code if it's a success.
## Changelog
:cl:
balance: Xenos always start at Ancient when evoing.
balance: Evo points required for T2 doubled from 80 to 160.
balance: Evo points required for T3 increased from 180 to 270.
balance: Queen death timer doubled from 5 to 10 minutes.
balance: Orphan hivemind timer lowered from 5 to 2 minutes.
balance: Maturity towers can no longer be bought in nukewar (distress).
balance: Silo cost increased from 800 to 1200 psy points.
balance: Primordial cost increased from 600 to 800 psy points.
/:cl:


KillerOrcaCora:
image

Pariah919:
Wait isn't this just removing maturity? We don't play anything other than nuke war, unless Distress is suddenly voteable for.

Naaanii:
> Wait isn't this just removing maturity? We don't play anything other than nuke war, unless Distress is suddenly voteable for.

Yes.

Crash exists but I can do a full removal if this is a success, inshallah.

chizzy376:
> I hope primordial will still exist

Read the pr

TiviPlus:
Maturity exists because
A) reward for not dying 
B) Xeno version of Power scaling

A is debatable B you are not replacing
Examples of replacement for B are mechanics similar to primo of which we have one (blessings) which isnt very fleshed out
Removing maturity can be discussed once it actually has a decent replacement

PR #12437, opened 2023-03-15
Fixes non-normal hive pheros not working

ivanmixo:
## About The Pull Request

Fixes non-normal hive pheros not working
## Why It's Good For The Game

Pheros being hardcoded to only work for one hive is stinky.
## Changelog
:cl:
fix: Fixed non-normal hive pheros not working
/:cl:


QualityVan:
I'm a little concerned about the field being null for all the aura gens you aren't touching in this pr (marine shouts, basically), but it's probably fine.
Given that xenos all share xeno faction anyway I bet you could put a hive ID in the current aura faction var with basically zero effort, then just check against faction for humans and hivenumber for xenos.

PR #12446, opened 2023-03-16
Fix valhalla playtime exploit

TiviPlus:
Didnt test btw


Lumipharon (approved):
I think this only adds to living playtime, not to actual jobs because they're completely separate but eh.

PR #12450, opened 2023-03-16
Makes xeno structures care about which hive made it

ivanmixo:
## About The Pull Request

Makes the structures hive specific - this means that if a silo is placed by a X hive xeno then the silo isn't going to generate larva for other hives. (TODO: make this work for hives besides the normal hive, currently only the normal hive gets considered. note: this requires xeno job refactoring)

Basically made every list of silos/spawners etc into an assoc list that is divided by hive numbers. 

Also removed acid turret marker since it's unused.

## Why It's Good For The Game

Everything being coded with assuming there's just one hive is bad.

## Changelog
:cl:
fix: Makes structures exist per hive, which prevents different hives from sharing larva gen rates
/:cl:


Lumipharon (approved):
Seems fine at a quick review, and TMing because of funny exploits.
Need a more thorough look over though.

Lumipharon (approved):

PR #12451, opened 2023-03-17
Kills suit pockets

Lumipharon:
## About The Pull Request
TM before merging.
Kills the evil that is var/pockets on suits.

Now we are in the modern age of attachments, the ancient pocket method of internal storage was super out of date and snowflake and created more code debt elsewhere which bugs me whenever I come across it.

Now all suit/storage type armour has an irremovable general storage module. This is a very slight buff for PAS/B18 etc since general storage can hold a few things the old pocket couldn't (like pistols).

Also did some various related code clean up.

Pockets still exist for boots and hats, so I'll take them out the back soon as well.
There's some further work I'll do to move the recolouring stuff down to the suit level as well, since there's really no reason for that to specifically be at the modular level.
## Why It's Good For The Game
Old code bad. New code consistant and good, with less bloat.
## Changelog
:cl:
balance: legacy armour such as B18 now uses irremovable general storage modules, so can fit a few extra item types
code: Updated legacy armour code to use storage attachments instead of the pockets var
/:cl:


chizzy376:
Does this mean that some armor that already has pockets loses it? Like B18, rsr coat

Oscari-pog:
Read PR.

Mrrpip:
Heretic cabal nerfing local Imperial Guards by nerfing their pockets, smh, smh.

chizzy376:
> Read PR.

I can't read

PR #12466, opened 2023-03-19
Throwing knifes rebalance

Xander3359:
## About The Pull Request
Throw force 25 -> 50
Throw Range 7 -> 4
Throw Delay 0.2 -> 0.4
## Why It's Good For The Game
Throwing knives are in a sad state. This should make them more rewarding to use without making them stronger.
## Changelog
:cl:
balance: Throwing knifes throw force 25 -> 50
balance: Throwing knifes throw Range 7 -> 4
balance: Throwing knifes throw Delay 0.2 -> 0.4
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -318,9 +318,9 @@
 	sharp = IS_SHARP_ITEM_ACCURATE
 	force = 20
 	w_class = WEIGHT_CLASS_TINY
-	throwforce = 25
+	throwforce = 50
 	throw_speed = 5
-	throw_range = 7
+	throw_range = 4
 	hitsound = 'sound/weapons/slash.ogg'
 	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut")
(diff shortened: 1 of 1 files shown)


Al-1ce:
> This should make them more rewarding to use without making them stronger

..By affecting the throw force you've done exactly that. But I get where you're coming from.
+1 from me.

Xander3359:
The idea is that by giving it more throw force, this makes it more appealing to those who choose throwing knives as a tech option. bootknife/holster knife users should get some more bang out of their storage space. 
But the overall DPS of throwing knives builds should be about the same.
Decreasing the range as well should make it harder for knives users to crit xenos from across the screen, as they should be strongest at slightly-further-than-melee range

isotope436:
this is gonna get abused with marcos very fast and probably fall into the same state the revolvers did when you could just carry 30 of them i especially see someone trying to pull the "but the range is shorter" card which doesnt matter since most of the combat you will have with any xeno is within 4 tiles since they are infact mostly melee. 

also that .2 increase to the throw delay is funny cause its not even noticeable and honestly will probably just get drowned out by TD

PR #12468, opened 2023-03-19
Mech pilot job worth required doubled to 80 from 40, job points start at 0

Pariah919:
## About The Pull Request
You now need 80 MP job points to an MP rather than 40, this makes mech pilots only really appear on super high pop, as everyone in the marine team is worth 1 point.
Job points also now start at 0.
## Why It's Good For The Game
Makes mech pilots more rare rather than something that occurs basically every midpop round.
## Changelog
:cl:
tweak: MPs now require double the amount of marine spawns to be acquired, as well as starting out at 0 job points rather than 15.
/:cl:


TiviPlus (approved):

Barnet2:
We are so back.

maikilangiolo:
Now make king pop requirement 20 xenos, then

PR #12486, opened 2023-03-21
Re-adds MMB for guns.

Al-1ce:
## About The Pull Request

In, #12282, MMB was changed so that it could not be used to fire selected objects, such as an underbarrel shotgun, or grenade launcher. This was done in the guise of being a "bugfix" - but as someone who uses MMB for underbarrels AND jetpacks, I feel like completely removing this was stupid and not the right way to go about things.

## Why It's Good For The Game

Allowing players to use the control scheme they've been used to for years is a good thing. Suddenly changing something a large portion of the server uses without informing the community isn't.

..And yes. I know there is the *VERY* specific scenario where a jetpack user could fly into their own flames. And that IS an issue. But removing MMB entirely isn't a fix. It's hardly even a band-aid. We wouldn't remove right-click because there was an issue caused by an oversight involved with it, now would we?

## Changelog
:cl:
add: MMB returns for usage in underbarrel attachments.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -678,9 +678,8 @@
 		return
 
 	if(modifiers["right"] || modifiers["middle"])
-	modifiers -= "right"
-	modifiers -= "middle"
-	modifiers -= "right"
+		modifiers -= "right"
+		modifiers -= "middle"
 		params = list2params(modifiers)
 		active_attachable?.start_fire(source, object, location, control, params, bypass_checks)
 		return
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -677,10 +677,10 @@
 	if(modifiers["shift"])
 		return
 
-	if(modifiers["middle"])
-		return
-	if(modifiers["right"])
-		modifiers -= "right"
+	if(modifiers["right"] || modifiers["middle"])
+	modifiers -= "right"
+	modifiers -= "middle"
+	modifiers -= "right"
 		params = list2params(modifiers)
 		active_attachable?.start_fire(source, object, location, control, params, bypass_checks)
 		return


Joe13413:
bro. you cannot use jetpacks well with underbarrels because of the weird MMB interaction. If you aren't gonna leave it as an option you can toggle on/off, then don't include this change. I'd say that a good portion of people put underbarrels on their primary weapons, and if those people run jetpacks then they'd shoot whenever they try to jetpack. Not very good. this isn't some super niche scenario you're describing, people wont run underbarrels with jetpacks because they cannot be used effectively.

ivanmixo:
> Suddenly changing something a large portion of the server

(me)

Xander3359:
I didnt touch MMB Jetpack. In fact, I specifically left that in, as MMB is supposed to be the jetpack button.

Al-1ce:
> I didnt touch MMB Jetpack. In fact, I specifically left that in, as MMB is supposed to be the jetpack button.

Guh. Well, I meant the gun code mostly anyways. Removing the ability to use MMB for guns is shitty.


> bro. you cannot use jetpacks well with underbarrels because of the weird MMB interaction. If you aren't gonna leave it as an option you can toggle on/off, then don't include this change. 

I'd love to, but being primarily a player and not a coder I've no idea how. I understand the conflict between under-barrels and jetpacks, yes, -- but removing MMB for guns completely is horseshit and not the way to go about things. Until someone else can make it a toggle, I don't feel like it's as big as an issue as you make it out to be.

>I'd say that a good portion of people put underbarrels on their primary weapons, and if those people run jetpacks then they'd shoot whenever they try to jetpack. Not very good. this isn't some super niche scenario you're describing, people wont run underbarrels with jetpacks because they cannot be used effectively


I'd agree if more than a couple of these spawned in vendors roundstart. Not only that, but very few people will order one. Jetpacks are infrequent at best and rare at worst when it comes to see them on the battlefield.

Additionally, most marines _do not_ actually use underbarrel attachments like the flamer, pepperball and shotgun - the angled grip is much, much more popular. So I don't feel this issue is as widespread as you're making it out to be.

TiviPlus:
This reintroduces the issue of conflicting keybindings, so no

PR #12496, opened 2023-03-22
CIC-Overwatch and CAS missions get minimaps, improves AIs minimap

DeltaFire15:
## About The Pull Request
This PR adds a minimap to CIC's consoles while in the overwatch camera view, which updates its z to the one currently viewed (or none if the z is minimapless).
Similarly, the AI minimap is now improved to track whichever z the AI's eye is on currently.
CAS also gets a minimap while running strike missions, and this might allow it to give it laze locations on the minimap at a later time (would be out of scope for this PR and I'm already doing enough here)

The locator also tracks the eyes and not the user in CIC / wherever, for all of these added / modified minimaps.
Minimap code may contain a little bit of signal sorcery to achieve this but it should work fine. Please do testmerge this before fullmerging, as there may nonetheless be things I missed or oopsied.

Also fixes two minor minimap related things,
one is that *default_overwatch_level* for minimaps didn't work properly due to a missing return call (though *default_overwatch_level*s aren't used at the moment now that I made AIs able to view different z's maps)
second was that miners used to runtime due to doing some minimap related things twice
## Why It's Good For The Game
CIC: Overwatch is supposed to be able to coordinate marines and be their eyes, which is somewhat difficult without easy access to the minimaps every single groundside marine has. This should make keeping an eye on the force as a whole easier, while also helping navigate the map.

AI: AIs minimap at the moment is effectively useless, only ever showing the ship's map which only matters when xenos board. This should allow it to be a more useful "eye" to the marines, akin to CIC.

CAS: As CAS pilot, it can sometimes be difficult to ascertain where marines are, aswell as where major combat is happening, especially ever since lazes were made unnecessary to deploy into strike mode. This should make it easier to locate marines and engagements, while also being able to coordinate somewhat better.

Bugfixes: Fix man good.
## Changelog
:cl:
add: CIC Overwatch now has a minimap while using the camera system.
add: CAS now has a minimap while on a groundside strike mission.
balance: AI's minimap now tracks the z the AI's eye is on, instead of only shipside
fix: Miners no longer cause minimap-related runtimes every init.
fix: default_overwatch_level for minimaps should now work properly.
code: Minimaps now support tracking atoms which aren't the owner of the action.
fix: Fixes a very minor, easy to trigger, droppod runtime.
/:cl:

### DIFF
--- a/code/controllers/subsystem/minimaps.dm
+++ b/code/controllers/subsystem/minimaps.dm
@@ -381,6 +381,8 @@ SUBSYSTEM_DEF(minimaps)
 	var/minimap_displayed = FALSE
 	///Minimap object we'll be displaying
 	var/atom/movable/screen/minimap/map
+	///Overrides what the locator tracks aswell what z the map displays as opposed to always tracking the minimap's owner. Default behavior when null.
+	var/atom/movable/locator_override
 	///Minimap "You are here" indicator for when it's up
 	var/atom/movable/screen/minimap_locator/locator
 	///This is mostly for the AI & other things which do not move groundside.
@@ -392,6 +394,7 @@ SUBSYSTEM_DEF(minimaps)
 
 /datum/action/minimap/Destroy()
 	map = null
+	locator_override = null
 	QDEL_NULL(locator)
 	return ..()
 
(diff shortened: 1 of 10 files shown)


ClosetedSkeletonGH:
Huge

Lumipharon:
Nice

code/game/objects/machinery/miner.dm
```diff
@@ -47,17 +47,23 @@
...
 	mineral_value = PLATINUM_CRATE_SELL_AMOUNT
 
 /obj/machinery/miner/Initialize()
 	. = ..()
-	SSminimaps.add_marker(src, z, hud_flags = MINIMAP_FLAG_ALL, iconstate = "miner_[mineral_value >= PLATINUM_CRATE_SELL_AMOUNT ? "platinum" : "phoron"]_off")
+	init_marker()
 	start_processing()
 	RegisterSignal(SSdcs, COMSIG_GLOB_DROPSHIP_HIJACKED, PROC_REF(disable_on_hijack))
 
+/obj/machinery/miner/proc/init_marker()
```
TiviPlus, line 64: autodoc

TiviPlus (changes requested):
rest ltgm

code/game/objects/machinery/miner.dm
DeltaFire15: forgor

DeltaFire15:
> Make it for the tadpole too? I do wish sometimes I could actually see where marines and bodies are without having to land then relaunch and work on memory.

Shouldn't be much of a hassle to do, but I'd prefer getting this merged first without increasing the things I touch even further. I'll code and PR that seperately once this is merged.

zzurge:
Someone remember AI exists in a PR,i kneel

code/controllers/subsystem/minimaps.dm
```diff
@@ -381,69 +381,141 @@ SUBSYSTEM_DEF(minimaps)
...
 	var/atom/movable/screen/minimap/map
+	///Overrides what the locator tracks aswell what z the map displays as opposed to always tracking the minimap's owner. Default behavior when null.
+	var/atom/movable/locator_override
 	///Minimap "You are here" indicator for when it's up
 	var/atom/movable/screen/minimap_locator/locator
 	///This is mostly for the AI & other things which do not move groundside.
 	var/default_overwatch_level = 0
+	keybinding_signals = list(
+		KEYBINDING_NORMAL = COMSIG_KB_TOGGLE_MINIMAP,
+	)
```
TiviPlus, line 392: move this to above new var definitions

TiviPlus (approved):

code/controllers/subsystem/minimaps.dm
DeltaFire15: done.

PR #12500, opened 2023-03-23
Moves green disk to engineering on LV

ivanmixo:
## About The Pull Request

Moves green disk across the river to engineering on LV.
## Why It's Good For The Game

I really hate how LV is being played right now. 

Every round is the exact same pointless marine push to engi followed by them just sitting there until the xenos attrition them enough. This should make that push a bit less pointless and give marines a timer to gtfo out of engi once the disk is done instead of sitting there eternally.

On the other hand this should also help with how marine sided the disks (and LV as a whole) are if the marines actually go for disks because it's harder to get compared to green being in hydro which is essentially free as it stands right now.
## Changelog
:cl:
balance: Moved green disk to engineering on LV
/:cl:

### DIFF
--- a/_maps/map_files/LV624/LV624.dmm
+++ b/_maps/map_files/LV624/LV624.dmm
@@ -7437,6 +7437,10 @@
 /obj/effect/turf_decal/warning_stripes,
 /turf/open/floor/mech_bay_recharge_floor,
 /area/shuttle/drop2/lz2)
+"hIs" = (
+/obj/machinery/computer/nuke_disk_generator/green,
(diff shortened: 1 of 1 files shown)


Lumipharon:
this would heavily cluster objectives more than they already are. I'd rather see red on the other side of the river, but eh.

Losenis:
It's not only free, it's right next to medbay so whoever dies there 100% gets revived and healed.

Naaanii:
Moving red to engineering might be better for combat flow but this works.

PR #12510, opened 2023-03-25
Kills lemoline recipes

Naaanii:
## About The Pull Request
Title
Keeps it as a catalyst for the remaining lemo gated chems.
## Why It's Good For The Game
Lemogate isn't really required since marines and even SLs can't make their chem cocktails. Gives shipside MR and CMO and Doctors something to do when they're not handling patients

Players doing things to help their team mates is good.
Sweaty powergamer making their own mixes so they can avoid having to rely on team members as much as physically possible is bad.
## Changelog
:cl:
del: Removes lemoline from all recipes requiring it.
/:cl:

### DIFF
--- a/code/modules/reagents/reactions/medical.dm
+++ b/code/modules/reagents/reactions/medical.dm
@@ -23,7 +23,7 @@
 /datum/chemical_reaction/dermalime
 	name = "Dermaline"
 	results = list(/datum/reagent/medicine/dermaline = 3)
-	required_reagents = list(/datum/reagent/oxygen = 1, /datum/reagent/phosphorus = 1, /datum/reagent/medicine/kelotane = 1, /datum/reagent/medicine/lemoline = 1)
+	required_reagents = list(/datum/reagent/oxygen = 1, /datum/reagent/phosphorus = 1, /datum/reagent/medicine/kelotane = 1)
 
 /datum/chemical_reaction/dexalinplus
 	name = "Dexalin Plus"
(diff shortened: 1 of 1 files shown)


Mrrpip:
Eh? But why would you need a catalyst then? If doctors (And medics) are the only people that can use it to begin with.

If they don't have lemoline, they buy an 8 point bottle and they're set, assuming they recycle.

Al-1ce:
> Edit: Point is, is there any particular reason we don't know about not to go the whole 9 miles?

If you'd read the PR you'd realize it doesn't do that. All this does is remove lemolame, which has been a boring and meaningless gatekeep chem for years

Lumipharon:
I'm not sure why you've made this when I already had issues with the last PR which did less than this one.

SzkieletorNutric:
image

Mrrpip:
> > Edit: Point is, is there any particular reason we don't know about not to go the whole 9 miles?
> 
> If you'd read the PR you'd realize it doesn't do that. All this does is remove lemolame, which has been a boring and meaningless gatekeep chem for years

Exactly! If you read my comment, you'd know I knew that! Catalyst and Removing the lock entirely would've been the same thing, so you may aswell have just done the latter.

Al-1ce:
> Exactly! If you read my comment, you'd know I knew that! Catalyst and Removing the lock entirely would've been the same thing, so you may aswell have just done the latter.

Because we remove enough shit full stop as it is. Mama mia...

PR #12528, opened 2023-03-27
Smartpistol restatting

Pariah919:
## About The Pull Request
Does a pass over smartpistol.

tl;dr makes it go from normal pistol ammo to AP, adds full auto. Makes it more light SMG-esque than full pistol.
Detailed changes below
<details>
Left(old)|Right(New)

Damage:20|20
AP:5|12.5
Sunder:1|2 [Swapped over from /pistol to /pistol/ap. in order to make it simmilar to other SGs with their higher penetration, I don't think a damage malus is necessary considering it fires at rifle fire rate.]

Mag Capacity: 25|30 [Heightens ammo count a tiny bit to make it more viable on its own.
Added Full auto [Makes sense considering this is dollar store fire support pistol, and 88mod has it which is its closest comparison statwise.]

Gun handling is unchanged, I do not believe this thing needs any nerfs in that regard, considering you need to buy it or have a Captain spawn in and hand it to you.

Gun description got a change as well.
</details>
## Why It's Good For The Game
Gun kind of poopy right now, needs some changes to make it worth picking up over a 88mod or TP-14, hopefully this is enough.

## Changelog
:cl:
balance: Smartpistol now uses pistol AP bullets rather than normal pistol bullets, and has full auto and a bit more magazine size (25->30).
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -728,12 +728,12 @@ It is a modified Beretta 93R, and can fire three round burst or single fire. Whe
 // Smart pistol, based on Calico M-950
 /obj/item/weapon/gun/pistol/smart_pistol
 	name = "\improper SP-13 smart pistol"
-	desc = "The SP-13 is a latest solution for personal officer defence produced by Terran Armories. A cutting-edge miniaturization technology allows mounting of smartgun IFF system on the pistol, albeit at high manufactoring cost. Unique design feature high-capacity mag on top of the barrel, with integrated sight. As with all smartgun systems, requires special training."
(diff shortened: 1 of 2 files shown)


Naaanii:
SGs should have the option to get this by default and not be a Captain only niche.

FosterGrime-Heymynameslouie:
making it so that the gun takes in normal pistol mags or make the smart pistol mag gun alot more accessible is pretty cool, mags are practically the only thign keeping da gun for being cool 

cool pr regardless, finally the smartpeestol being somewhat useful

Lumipharon (approved):
this makes it actually a very good gun but eh, captain only.

PR #12529, opened 2023-03-27
Autocannon flak now does 2.5x damage to all walls.

Pariah919:
## About The Pull Request
Autocannon flak does x2.5 damage to all walls, desc changed to reflect this.
## Why It's Good For The Game
Gives flak an actual gimmick, it annhilates walls now, not much else to add.
## Changelog
:cl:
balance: Autocannon flak ammo does x2.5 damage to walls such as resin or metal walls.
/:cl:


maikilangiolo:
Hmm achktually it won't pierce through people because it doesn't have the right PASS_THROUGH flag 🤓🤓

Al-1ce:
I don't think the variable is high enough to be useful compared to, say, a plasma cutter. Maybe round it to just 3?

Lumipharon (changes requested):
As mentioned, this increases damage EVERY time it goes through a wall, and applies to mobs hit. DNM until fixed.

Pariah919:
going to think on this a bit more

PR #12542, opened 2023-03-29
Hive threat reward alt fix

QualityVan:
## About The Pull Request
Alternative to #12540 
Turns out that comparing upgrade defines against xeno tier isn't actually what you want to do
## Why It's Good For The Game
T3/4 xenos actually get the bump to ancient if they're low maturity, as they should
## Changelog
:cl:
fix: Hive threat event rewards instant ancient to any kind of xeno instead of sometimes de-aging a primordial t2
/:cl:

### DIFF
--- a/code/modules/events/hive_threat.dm
+++ b/code/modules/events/hive_threat.dm
@@ -51,7 +51,7 @@
 		receiving_xeno.salve_healing()
 		if(receiving_xeno == drainer)
 			receiving_xeno.evolution_stored = receiving_xeno.xeno_caste.evolution_threshold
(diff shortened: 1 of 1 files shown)


QualityVan:
Having thought about it a bit more, I kind of like Lumi's change more just because it's a lot more agnostic regarding current xeno tier - you can get the drain as t1 and later evolve to a fully aged t3, whereas this version almost punishes you for taking the absorb as a low tier xeno. I'll still leave this up in case of dissent.

Lumipharon:
Closing this because the original reward could be gamed somewhat in that the xeno claiming the reward had a direct impact on the actual reward given. I prefer my change as its just a flat bonus.

PR #12543, opened 2023-03-29
SSsilo priority tweak

Lumipharon:
## About The Pull Request
Increases the priority of the Silo subsystem.

I saw an odd occurance today where SSsilo got stuck in SS_SLEEPING. This completely stops larva silo larva generation without admin intervention which is not ideal. It also made me wonder how often this has happened, as it's not directly noticable unless you actually watch to see if larva gen is ticking every minute.

It only fires once per minute and is very basic, so I don't think a very high priority should be an issue, considering its relative importance.
## Why It's Good For The Game
Larva gen working reliably is good.
## Changelog
:cl:
code: Increased SSsilo priority to reduce the chance of it getting stuck
/:cl:


code/__DEFINES/_subsystems.dm
```diff
@@ -122,6 +122,7 @@
 #define FIRE_PRIORITY_AUTOFIRE 450
 #define FIRE_PRIORITY_OVERLAYS 500
 #define FIRE_PRIORITY_EXPLOSIONS 666
+#define FIRE_PRIORITY_SILO 667
 #define FIRE_PRIORITY_TIMER 700
```
TiviPlus, line 126: ok but not higher than explosions wtf
make it like points tier-ish

TiviPlus (changes requested):

code/__DEFINES/_subsystems.dm
Lumipharon: I've never touched this before. What are the consequences of having it at this level? My logic is its very important gameplay wise, but uses up very little processing time (it doesn't do much and only fires once a minute).

TiviPlus: It determines how "important" it is when master cant run everything
IE here it decides if it only has enough time for 1 it will run silos instead of explosions which will wait until next tick

PR #12545, opened 2023-03-29
Remove the xeno tier limit on tier 2 and 3 casts

Autisem:
## About The Pull Request

Removes the limit, making PR to also debate the topic

## Why It's Good For The Game

I argue for because while the tier 3 xenos are some what better, they are not THAT much better that they warent being hard locked. this is even more true for tier 2. I'd rather see these xenos being closer together in terms of power and making it more akin to picking a loadout for marines then.  You ultimitaly still want a mix of xenos and stacking a bunch of 1 cast may work only if it catches the enemy completely off guard.

This also prevents you being locked at tier 1, which sometimes it not fun if you want to say play carrier.

Something to consider could also be only removing tier limit on tier 2, if tier 3's are really too strong. but most of the xenos are in tier 3 currently so letting players play as those would be more fun

## Changelog
:cl:
balance: There is no limit on xeno's per tier anymore
/:cl:

### DIFF
--- a/code/controllers/subsystem/silo.dm
+++ b/code/controllers/subsystem/silo.dm
@@ -26,8 +26,6 @@ SUBSYSTEM_DEF(silo)
 	current_larva_spawn_rate += larva_spawn_rate_temporary_buff
 	GLOB.round_statistics.larva_from_silo += current_larva_spawn_rate / xeno_job.job_points_needed
 	xeno_job.add_job_points(current_larva_spawn_rate)
-	var/datum/hive_status/normal_hive = GLOB.hive_datums[XENO_HIVE_NORMAL]
-	normal_hive.update_tier_limits()
 
 ///Activate the subsystem when shutters open and remove the free spawning when marines are joining
 /datum/controller/subsystem/silo/proc/start_spawning()
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -196,7 +196,6 @@
 	if(larva_surplus < 1)
 		return //Things are balanced, no burrowed needed
 	xeno_job.add_job_positions(1)
-	xeno_hive.update_tier_limits()
 
 /datum/game_mode/infestation/crash/get_total_joblarvaworth(list/z_levels, count_flags)
 	. = 0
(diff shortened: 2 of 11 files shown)


ClosetedSkeletonGH:
So imagine this team comp: 5 boilers, 4 ravs, 1 king, 1 queen, 1 shrike, full pheros.

You can do this with any combo. Perhaps T2 I can understand unlocking but any amount of duplicate T3’s can be an absolute game ender for marines if they don’t suicide (see 3 boilers/3 ravs/3 gorgers/etc). It’s very difficult to get that right now though, but this would make it a very common occurrence.

I’d love to see more T2’s, but I think we see an appropriate amount of T3’s

Autisem:
> So imagine this team comp: 5 boilers, 4 ravs, 1 king, 1 queen, 1 shrike, full pheros.
> 
> That’s a hive with just 12 xeno’s
> 
> You can do this with any combo. Perhaps T2 I can understand unlocking but any amount of duplicate T3’s can be an absolute game ender for marines if they don’t suicide (see 3 boilers/3 ravs/3 gorgers/etc). It’s very difficult to get that right now though, but this would make it a very common occurrence.
> 
> I’d love to see more T2’s, but I think we see an appropriate amount of T3’s

stacking boilers will only work until they decide to buy mimir and upgrade to acid proof cades, the ravs are strong but mainly in ambushes.

stacking widows would honestly be something takes makes me scared but I even find that unlikely that it will be an actuel issue other then occasional funny round

chizzy376:
15 rav time

Losenis:
>So imagine this team comp: 5 boilers, 4 ravs, 1 king, 1 queen, 1 shrike, full pheros.

Req spams autosnipers, entire composition hardcountered. Alternatively there's so many ungas that those 4 ravs king shrike and queen can't realistically stop the chasers from just BTFOing the boilers because for 12 xenos there's going to be between 24 to 36 marines.

ivanmixo:
Yeah good luck getting the points for autosnipers when you can't even get out of fob.

Losenis:
That already happens when marines bald out

maikilangiolo:
Surely this is a joke PR, right?

BaconBro1357:
Hey, I know this PR is pretty unpopular but I would like to say, as a person who plays xeno semi-regularly, I can say that the tier 3 slots rarely if ever get full, or even approach getting full. I am not a coder but I do want to give my personal experiences since it seems that removing the tier 3 slot limit would not impact the game in any real way.

Naaanii:
This nullifies the Evo tower buff as well.

Lumipharon:
This is not a good idea for balance.

Joe13413:
dude pro tip for getting your shit merged: dont do fentanyl before coding and making a PR

PR #12546, opened 2023-03-29
claymore balances

567Turtle:
## About The Pull Request

sword hits slower :)

## Why It's Good For The Game

whats the point

## Changelog
:cl:
balance: Claymore now hits slower 
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -359,7 +359,7 @@
 	throwforce = 65
 	throw_speed = 3
 	edge = 1
-	attack_speed = 20
+	attack_speed = 24
 	sharp = IS_SHARP_ITEM_BIG
 	w_class = WEIGHT_CLASS_BULKY
 	flags_item = DRAINS_XENO | TWOHANDED
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -359,7 +359,7 @@
 	throwforce = 65
 	throw_speed = 3
 	edge = 1
-	attack_speed = 20
+	attack_speed = 22
 	sharp = IS_SHARP_ITEM_BIG
 	w_class = WEIGHT_CLASS_BULKY
 	flags_item = DRAINS_XENO | TWOHANDED
@@ -378,14 +378,6 @@
 /obj/item/weapon/twohanded/glaive/harvester/Initialize()
 	. = ..()
 	AddComponent(/datum/component/harvester, 60)
-
-/obj/item/weapon/twohanded/glaive/harvester/equipped(mob/user, slot)
-	. = ..()
-	toggle_item_bump_attack(user, TRUE)
-
-/obj/item/weapon/twohanded/glaive/harvester/dropped(mob/user)
-	. = ..()
-	toggle_item_bump_attack(user, FALSE)
 
 /obj/item/weapon/twohanded/glaive/harvester/get_mechanics_info()
 	. = ..()
--- a/code/game/objects/items/weapons/twohanded.dm
+++ b/code/game/objects/items/weapons/twohanded.dm
@@ -359,7 +359,7 @@
 	throwforce = 65
 	throw_speed = 3
 	edge = 1
-	attack_speed = 22
+	attack_speed = 20
 	sharp = IS_SHARP_ITEM_BIG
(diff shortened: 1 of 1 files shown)


Dawson1917:
Without bumpslash it's just worse than the spear.

Dawson1917:
>With bumpslash its better then the sword, what do you propose I change instead?

Sword is 1h bumpslash so you can use it with shield or as a secondary, and doesn't need to be wielded so you don't have to wait and hit wield after every stun/knockdown for your damage to not be dogshit.

The real issue with claymore is that tramadol is and has always been OP on vali weapons and the other two chems are completely useless.

XharlemiX:
Good luck on your way making a weapon useless to take

maikilangiolo:
Remove bumpslash, don't nerf the speed. Bumpslash is for skill issued marines, and it's already a two handed, one tile range weapon. Sword can be used with shield/a gun, spear has two tiles range. Doesn't need to be slower

Dawson1917:
Bumpslash is for skill issue and anyone with high ping. If you click a xeno with spear it'll still hit if they move a tile away, sword and claymore need them to stay directly next to you when you click for the time it takes your inputs to reach the server. It's why I now take it over spear (200-240ms ping).
Same reason we have bumpslash for all xenos.

KillerOrcaCora:
Wifi issue

Lumipharon (changes requested):
Claymore should have some slowdown when wielded or whatever. Tiny reductions in hit rate don't matter when you're not hitting perfectly anyway.

Joe13413:
damn man can you at least describe what you changed, i dont wanna go digging through code to look for changes you made. gimme some good numbers, spend more than10 seconds writing your PR

Dawson1917:
Claymore does already make your movement slowdown matter more for chases and bumpslashes, since the one-tile range makes landing hits (and applying tram slowdown) more difficult compared to spear, and bumpslash triggers on you moving into a tile so having high speed makes WASDing more effective.

Trying to tactically wield/unwield for chases isn't an option either since melee attacks lock all input (wield/unwielding, grabs, clicks, etc) for like two full seconds so any attack on either your target xeno, a bumpslash, or weeds is going to eat your inputs for the next few seconds. I don't know why the input eating is a thing and would fix it if I could.

chizzy376:
Weapon gets merged for 5 miliseconds, nerf PR cabal is already here. Mfw xenos can't beat a 1 tile ranged marine when most their kit is range based.

Lumipharon:
> Claymore does already make your movement slowdown matter more for chases and bumpslashes, since the one-tile range makes landing hits (and applying tram slowdown) more difficult compared to spear, and bumpslash triggers on you moving into a tile so having high speed makes WASDing more effective.
> 
> Trying to tactically wield/unwield for chases isn't an option either since melee attacks lock all input (wield/unwielding, grabs, clicks, etc) for like two full seconds so any attack on either your target xeno, a bumpslash, or weeds is going to eat your inputs for the next few seconds. I don't know why the input eating is a thing and would fix it if I could.

Wielding an item doesn't increase slowdown unless actually specified. Wielding a claymore is just as fast as unwielded.

FIXTHERADBOXPLEASE:
Honestly I'd have liked bumpslash removed so it's a more skill-based weapon compared to the other Vali, but ping difficulties just make that untenable... So yeah, I guess less speed is the only way.

Dawson1917:
>Wielding an item doesn't increase slowdown unless actually specified. Wielding a claymore is just as fast as unwielded.

I meant my comment in response to yours about adding slowdown to the claymore. ie, how the claymore works (compared to spear) makes the slowdown you already have from your armor type matter more both for chases and bumpslashing, so adding more would be compounding the issue. As it stands spear is the better pick for chases and defensive fighting (hitting xenos while staying out of their melee range) while claymore is better when xenos are the ones coming after *you*.

Given the difference between claymore and spear is 30 damage, a slower attack speed and half the attack range, I'm going to posit that the complaints about claymore come down to two things: the usual suspect of something being new and everyone using it, so people dying to it call it OP, and the fact that tramadol on vali weapons is busted and nobody cares enough to rebalance vali chems. I guarantee there would be zero complaints if people were running kelo on their claymores, because tram is the problem, not the weapon.

Lumipharon:
The weapon is probably still problematic in some ways, but I'm not opposed to this I guess.

Xander3359:
🦀🦀🦀Claymore is dead🦀🦀🦀

PR #12553, opened 2023-03-30
Railgun changes

Pariah919:
## About The Pull Request
Railgun now holds 3 in a canister, price in req upped from 3(0)->5(0) to compensate
Railgun, since it now has an actual magazine holding more than 1 round now has 3 fire delay.
## Why It's Good For The Game
Railgun currently has terrible ammo economy, this should assist it in acquiring some real ammo economy so it isn't awful to use. I don't personally think this is too OP outside of maybe HVAP should stay at 1 and 3 points.
## Changelog
:cl:
balance: Railgun now holds 3 shots in a canister rather than 1, has 3 seconds of fire delay inbetween shots. Railgun ammo now costs 50 points rather than 30 to accommodate for the larger magazine size.
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -399,6 +399,7 @@
 		/obj/item/ammo_magazine/revolver,
 		/obj/item/ammo_magazine/sniper,
 		/obj/item/ammo_magazine/handful,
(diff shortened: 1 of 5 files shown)


srxtreme:
1 by 1 forgotten req items are made great again

PR #12563, opened 2023-03-31
Adds cryo to small Cant, adds cryo beakers to both variants

Mrrpip:
## About The Pull Request

It does exactly as it says on the tin. To be more specific, these changes have been made, indicated by sharpie.

Canterbury1

Canterbury2

## Why It's Good For The Game

Well...It kinda sucks that one version of the Cant has cryo while the other doesn't, especially with certain clone loss causing meds. 

In an ideal world, I'd like both versions to have chemistry too, but uh...I can't be bothered to figure out how to fit chem into SmallBury

## Changelog
:cl: Skye
qol: Adds cryogenics to the TGS Smallbury, used in the Crash game mode. (Also moves a table.)
balance: Adds a cryomix beaker to both Canterbury variants.
/:cl:

### DIFF
--- a/_maps/shuttles/tgs_bigbury.dmm
+++ b/_maps/shuttles/tgs_bigbury.dmm
@@ -161,6 +161,7 @@
 /obj/machinery/light/small{
 	dir = 4
 	},
+/obj/item/reagent_containers/glass/beaker/cryomix,
 /turf/open/floor/mainship/sterile,
 /area/shuttle/canterbury/medical)
 "aO" = (
--- a/_maps/shuttles/tgs_canterbury.dmm
+++ b/_maps/shuttles/tgs_canterbury.dmm
@@ -381,10 +381,15 @@
 /turf/open/floor/mainship/mono,
 /area/shuttle/canterbury)
 "bt" = (
-/obj/machinery/door/airlock/mainship/marine/canterbury{
-	dir = 8
-	},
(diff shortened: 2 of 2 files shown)


Mrrpip:
Yes, I know, could've done better than sharpies, but still. It works.

RipGrayson (approved):
Mapping is fine, other people can debate the balance implications

Naaanii:
Or remove it from the bigbury for the opposite consistency. Would probably be a better alternative than giving Xenos psydrain or something in crash.

Extra beaker is kinda dumb because almost all maps have an easily useable and lootable cryo cell next to a medbay disk so it'll always be reachable.

Mrrpip:
> Or remove it from the bigbury for the opposite consistency. Would probably be a better alternative than giving Xenos psydrain or something in crash.

That is a distinct possibility, though this PR seeks to make both equal for the time being. Merged or not, I'll consider that idea later, as it has much bigger implications than 'One ship has, one hasn't'.

> Extra beaker is kinda dumb because almost all maps have an easily useable and lootable cryo cell next to a medbay disk so it'll always be reachable.

I'd like to trust the judgement of the person who suggested it. Regardless, while it seems redundant, with the Bigbury existing with it's own chemistry and dispensable beakers, the mix could be easily recreated on that ship, except for the fact that you need a CMO, medic or Synth to do it.

I am curious as to what the maintainers think that point, though.

PR #12569, opened 2023-03-31
retvrn to colonial marines [DNM] [April Fools]

tyeagg:
## About The Pull Request
feel free to review even if its draft
image
terragov and nanotrasen sucks screw this scifi jaeger mech bullshit we out here using kitbashed vietnam equipment
happy apirl 1

credits to crow for smartgun sprite

todo:
maybe a few more guns
quick load loadouts for uscm
## Why It's Good For The Game
1 of april
## Changelog
:cl:
add: Added new mechanics or gameplay changes
add: Added more things
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

PR #12573, opened 2023-03-31
[DNM] [April Fools] Complete and utter disregard for sound effects & your eardrums

mrhugo13:
## About The Pull Request
randomizes a fuckton of sound effects by combining them all into one big sound effect list, some sound effects are still always the same but stuff like *warcry and opening up your backpack picks a random sound effect from this list.

probably not very efficient to run on the server however it does compile and it functions in-game.
probably won't work with #12569 in which case give priority to TM'ing that PR instead.

## Why It's Good For The Game
april fools is a wonderful time to have fun, this is one way you can have fun.

## Changelog
:cl: Me, the main character.
add fun
/:cl:


Oscari-pog:
Slash sound replaced with a lobby song

chizzy376:
> Slash sound replaced with a lobby song

good

PR #12574, opened 2023-03-31
Sunder multiplier variable added

MiniMeatwad:
## About The Pull Request

Adds multiplier define for application of sundering.

## Why It's Good For The Game

So MJP can play with numbers.

## Changelog

:cl:
add: Added sunder multiplication define.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
@@ -509,7 +509,7 @@
 	. = ..()
 	if(.)
 		return
-	sunder = clamp(sunder + adjustment, 0, xeno_caste.sunder_max)
+	sunder = clamp(sunder + adjustment * xeno_caste.sunder_multiplier, 0, xeno_caste.sunder_max)
 
 /mob/living/carbon/xenomorph/set_sunder(new_sunder)
 	. = ..()
--- a/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
@@ -523,12 +523,6 @@
 		return
 	return (sunder * -0.01) + 1 
 
-/mob/living/carbon/xenomorph/get_sunder_multiplier()
-	. = ..()
-	if(.)
-		return
-	return xeno_caste.sunder_multiplier
-
 /mob/living/carbon/xenomorph/adjust_stagger(amount)
 	if(is_charging >= CHARGE_ON) //If we're charging we don't accumulate more stagger stacks.
 		return FALSE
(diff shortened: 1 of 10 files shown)


code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
```diff
@@ -521,7 +521,13 @@
 	. = ..()
 	if(.)
 		return
-	return (sunder * -0.01) + 1
+	return (sunder * -0.01) + 1 
+
+/mob/living/carbon/xenomorph/get_sunder_multiplier()
```
QualityVan, line 526: What's the use case where you'd want more than just the last line here?

Lumipharon (changes requested):
If you have a mult like this, it should apply in the adjust_sunder proc itself, not have to put it in every set of args for the proc.

MiniMeatwad:
> If you have a mult like this, it should apply in the adjust_sunder proc itself, not have to put it in every set of args for the proc.

What's funny is bravemole told me not to do this on a different code piece before, and to leave it as optional, but I agree with you so I will do so

code/modules/mob/living/carbon/xenomorph/xenoprocs.dm
MiniMeatwad: What do you mean?

Lumipharon:
This is fine to merge, but I'd rather not add it if its not actually being utilised. The vague "mjp might use it in the future' is the same as mjp causing req inflation and then not using it for months.

TiviPlus (changes requested):
my complaint is why tf do we need this it's just pointless micro tweaking that adds complexity for ???

MiniMeatwad:
> it's just pointless micro tweaking that adds complexity for ???

This is just a proposition for sunder changes. The PR exists solely to bring up discussion about sunder changes, if MJP is willing to do numbers, and maintainers are fine with changing sunder. I'll draft it to accompany this.

> why tf do we need this

MJP considered reducing the sunder cap on crusher so you'd keep 40% of your armor at max sunder, and he wanted to reduce sunder adjustment to keep sunder values consistent with healing and dealing, which is what this PR currently does.

As for what I would like to do with sunder:

I don't it being concrete with touching sunder_max itself, I think it could work with an armor class which just does the equivalent in pseudo code of subtracting from 100 by clamp(20 * weapon_class - armor_class, 0, 100) at 1-5 per class, and I could also set it up to do diminishing returns with that in mind, so as to not make it a hard cap.

I like the concept of using some number to limit certain weapons to a level against certain quality castes, so as an example SMGs can still have really high sunder and shred against T1s which already have low armor, and barely affect high armor castes. Obviously this requires someone willing to put attention to detail for this, so it's only an idea.

MiniMeatwad:
Closing because of lack of approval.

PR #12592, opened 2023-04-02
Bicaridine, Kelotane, and Tramadol reagent bottles are infinite in medical vendors

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
I have no idea what kind of silly war is being waged against chemicals, power gaming, and whatever other reason is being thrown around -- Vali and its players have been getting the short end of the stick, and recent changes have essentially disabled any other role besides Corpsman from being able to use it properly.

You probably shouldn't be disabling an entire playstyle regardless of excuses.

## Changelog
:cl: Lewdcifer
balance: Bicaridine, Kelotane, and Tramadol bottles are infinite in medical vendors.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/vending_types.dm
+++ b/code/game/objects/machinery/vending/vending_types.dm
@@ -173,9 +173,9 @@
 			/obj/item/reagent_containers/hypospray/advanced/dylovene = 5,
 		),
 		"Reagent Bottle" = list(
-			/obj/item/reagent_containers/glass/bottle/bicaridine = 6,
-			/obj/item/reagent_containers/glass/bottle/kelotane = 6,
-			/obj/item/reagent_containers/glass/bottle/tramadol = 6,
+			/obj/item/reagent_containers/glass/bottle/bicaridine = -1,
+			/obj/item/reagent_containers/glass/bottle/kelotane = -1,
+			/obj/item/reagent_containers/glass/bottle/tramadol = -1,
 			/obj/item/reagent_containers/glass/bottle/dylovene = 6,
 			/obj/item/reagent_containers/glass/bottle/inaprovaline = 6,
 			/obj/item/reagent_containers/glass/bottle/dexalin = 6,


maikilangiolo:
What

QualityVan:
You're talking about general vali use and a vendor which (by default) is locked to md/corps anyway. Just give vali the ability to load from pills.

Lumipharon:
> You're talking about general vali use and a vendor which (by default) is locked to md/corps anyway. Just give vali the ability to load from pills.

This. Hypos can already liquify pills so just use the same functionality.

Naaanii:
Empty hypos in regular vendors would be nice. Should probably include tricord somewhere too.

Arctrooper8:
> > You're talking about general vali use and a vendor which (by default) is locked to md/corps anyway. Just give vali the ability to load from pills.
> 
> This. Hypos can already liquify pills so just use the same functionality.

Then the vali "economy" should be changed alongside it because it's focused around not just intaking chems but also duplicating them from green blood you get from aliens. You'd want to change it to maybe print pills or have some sort of internal beaker where the liquefied pills go allowing you to duplicate basic chems as if it was a reagent bottle. 
Either way even if you could put pills into the vali it'd still be sub par to bottles.

Autisem:
Maby add vials just for vali? a bottle that can only be inserted into vali suit and nothing else. this would also let them bypass the acces issue by putting it in a vendor they have acces too

chizzy376:
> because it's focused around not just intaking chems but also duplicating them from green blood you get from aliens

vali can duplicate chems with green blood? How? I never heard of this before

Arctrooper8:
> > because it's focused around not just intaking chems but also duplicating them from green blood you get from aliens
> 
> vali can duplicate chems with green blood? How? I never heard of this before

If you have any amount of the basic chems (Bicard, Kelo, Trama) you can hold out the container and then via the vali settings menu use the extract green blood to put it into the bottle. When the green blood mixes with any of those three chems it becomes that chem, essentially duplicating it.

PR #12603, opened 2023-04-03
Adds large 10x26mm caseless box

irRegularGuy646:
## About The Pull Request
Adds a large ammo box for loose 10x26mm caseless that holds 2400 rounds and can be infinitely vended from the req supplies vendor like all other large ammo boxes.

## Why It's Good For The Game
I thought it was kind of silly that you can sling large boxes of rifle and SMG ammo over your back, but no such thing existed for the quintessential machinegun caliber. You know, the caliber designed to be spammed in large volumes in the general direction of the enemy. The one that would be most useful in a large, bulk format such as a big box of loose rounds. it holds the same amount as the rifle ammo box because a backpack full of 10x26 packets is about 1200 rounds, half of 2400.
## Changelog
:cl:
add: Added big 10x26mm caseless ammo box to req operational vendor
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -904,6 +904,7 @@
 		"Ammo Boxes" = list(
 			/obj/item/big_ammo_box = -1,
 			/obj/item/big_ammo_box/smg = -1,
+			/obj/item/big_ammo_box/mg = -1,
 			/obj/item/shotgunbox = -1,
 			/obj/item/shotgunbox/buckshot = -1,
 			/obj/item/shotgunbox/flechette = -1,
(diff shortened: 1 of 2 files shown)


Losenis:
Assuming an ammo packet for the GPMG would be 3x the quantity of the drum since ammo packet for T12 is 3x its mag, you'd have a packet being 600 rounds. With the math used for the values for the other big ammo boxes, it'd have to be backpack full of packets x2, so 600x8x2 for 9600 rounds for it to actually be worth not only losing the backpack slot to carry drums in but also locking the belt slot to a G8 pouch for drums, on top of having to manually refill each drum instead of just carrying a gigantic ammo box of it to grab from.

PR #12605, opened 2023-04-03
Adds self germ check

QualityVan:
## About The Pull Request
Shows up on self click under the wounds check.
Also included an absolutely critical grammar fix.
## Why It's Good For The Game
Makes it more apparent that germs actually exist. Didn't add a way to check them on someone else because that should never matter.
## Changelog
:cl:
add: When you check yourself for wounds, you can see the dirt you've accumulated. Mostly only matters for surgeons.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -39,7 +39,7 @@
 				return FALSE
 
 			if(species?.species_flags & ROBOTIC_LIMBS)
-				to_chat(H, span_boldnotice("You cant help this one, [p_they()] have no lungs!"))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -39,7 +39,7 @@
 				return FALSE
 
 			if(species?.species_flags & ROBOTIC_LIMBS)
-				to_chat(H, span_boldnotice("You cant help this one, [p_they()] have no lungs!"))
+				to_chat(H, span_boldnotice("You can't help this one, [p_they()] have no lungs!"))
```
ivanmixo, line 42: Also change the have to the macro too while you're at it. It often appears as `He have no lungs!`

Issue #12616, opened 2023-04-04
Marines can get abducted into fog


chizzy376:
### Description

Any caste with a blink or a special throw move can do this.

### Test Merges

Server-Side Code Modifications: https://tgstation13.org/parsed-logs/sybil/server_side_modifications.dm

Round ID: [20503](https://scrubby.melonmesa.com/round/20503)
BYOND Version: 515.1604
Server revision compiled on: 
Master commit: [ac56f43309c2ba24213b24fd70b7a72ce7246bda](https://github.com/tgstation/TerraGov-Marine-Corps//commit/ac56f43309c2ba24213b24fd70b7a72ce7246bda)
The following pull requests are currently test merged:
[#12584: 'Adds minimap drawing' by TiviPlus at commit 957675a705](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12584)

Local commit: 925e61c433d25affcf115c0798d9156308ba1ae6
Server tools version: 5.7.3

### Reproduction Steps

1. Play crusher (Or wraith, woyer also works iirc)
2. Stand on fog with help intent
3. Click marine with crest toss adjacent to you
4. They get tossed into the fog and can't ever get back out

Xander3359:
Duplicate Issue https://github.com/tgstation/TerraGov-Marine-Corps/issues/11857

chizzy376:
> Duplicate Issue #11857

Just goes to show that nobody has fixed it yet

JackTheJackhammer:
https://user-images.githubusercontent.com/77555989/233492504-5dbd9277-0102-4aff-84ee-d8cbcd491444.mp4

Video of it

Issue #12615, opened 2023-04-04
Wraith can teleport nuke into fog


chizzy376:
### Description

The griefer caste known as Wraith can utilize portals to push one-of-a-kind objects into UNREACHABLE areas for the enemy team.

### Test Merges

Server-Side Code Modifications: https://tgstation13.org/parsed-logs/sybil/server_side_modifications.dm

Round ID: [20503](https://scrubby.melonmesa.com/round/20503)
BYOND Version: 515.1604
Server revision compiled on: 
Master commit: [ac56f43309c2ba24213b24fd70b7a72ce7246bda](https://github.com/tgstation/TerraGov-Marine-Corps//commit/ac56f43309c2ba24213b24fd70b7a72ce7246bda)
The following pull requests are currently test merged:
[#12584: 'Adds minimap drawing' by TiviPlus at commit 957675a705](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12584)

Local commit: 925e61c433d25affcf115c0798d9156308ba1ae6
Server tools version: 5.7.3

### Reproduction Steps

1. Steal unanchored nuke from marines (Or even have marines do it for you by putting the portal in front of nuke as they're moving it lol!!!)
2. Put one portal in fog, one outside of fog
3. Push nuke through
4. Make the game unwinnable for marines

### Screenshots

This shouldn't be a thing
![bullshit](https://user-images.githubusercontent.com/80391698/229752754-38e7cf73-18cd-44b1-9089-9e5229319216.png)

PR #12618, opened 2023-04-04
New element to restrict things from entering places they should not

Autisem:
## About The Pull Request

This element will deny the attached object entry on any turf where the "spawn_protection" effects are placed.  This will prevent various bugs like wraiths being able to teleport the nuke into the xeno spawn or, if attached to marines. prevent marines from being moved into the turf.

The ultimate "fuck off your not allowed here"

I can add a version that checks for specific conditions(say restrict xenos instead of marine equipment) if desired.

Also maby rename since the name kinda suck

nukefix

## Why It's Good For The Game

A lightweight(I suppose) way of preventing importent objects from entering unintended area's may be usefull. Solves the bug that wraiths can teleport nuke into spawn

## Changelog

:cl:
add: New way to prevent things going where they are not supposed to
/:cl:


Autisem:
resolves #12615 and can be applied to resolve #12616

FosterGrime-Heymynameslouie:
very cool thank yo for your service

Autisem:
We could also atomize adding it to the actuel map to dif PR's to make it a bit more managable

PR #11387, opened 2022-10-30
Nerfs Warrior Lunge range from 6 to 4 tiles

lKiseki:
## About The Pull Request
Per title. This PR is meant to go together with the Lunge Punch damage modifier nerf PR here: #11386.
It is published separately since maintainers don't like unatomized nerfs.
**If a choice must be made between either, prioritize #11386 before this one.**

### Old Warrior Lunge range

https://user-images.githubusercontent.com/59634950/198903302-02540db4-a6d8-4822-9a36-aea107d0c060.mp4

### New Warrior Lunge range

https://user-images.githubusercontent.com/59634950/198903315-16079228-2fd4-4c1b-972b-5eee9843eeaa.mp4

Effective range:
image

## Why It's Good For The Game
Lunge having 6 tiles of range has contributed to the current issue of Warriors being too good at what they do, not to mention being able to easily punish from considerable distances. This PR aims to amend that.

## Changelog
:cl: Lewdcifer
balance: Warrior Lunge range reduced from 6 to 4 tiles.
/:cl:

PR #12623, opened 2023-04-04
Reverts #11387 . Warrior can lunge in 6 tiles again.

DecemberIsHere:
## About The Pull Request

This PR reverts #11387 by allowing warriors to lunge within a 6 tile range.

## Why It's Good For The Game

Lunge caught a stray nerf, to an entirely damage related issue with warriors kit. Its ability was within marine screen view range, and was countered by using flares/lights. It also allowed for warriors to engage without being forced to be within shotgun/slug wield range, due in part to how warriors lunge has a long cooldown and requires precise clicks. Now, it's been nerfed to 4 tiles which means you are at complete risk in engagements for an ability that can potentially kill you, and you will almost always be seen before being lunged IE there is no factor to pushing blindly into darkness that a warrior might teeter the edge and lunge you.  In the PR even, it states that 6 tile lunges allowed for warriors to be "too good at what they do" which was true, warriors were too good with their **flings** and their **damage dealing** abilities, and **NOT** their main ability, lunge which was only a setup for said abilities. 

Also something to consider in mind before pitch forks are raised about concerns with grapple toss, that ability has already been nerfed and requires a competent hive and warrior to setup kills as so. A 6 tile lunge did not change this, and nerfing lunge to 4 tiles was not good for the game.

## Changelog

:cl:
balance: Warrior lunge is now 6 tiles, up from 4
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/warrior/abilities_warrior.dm
@@ -77,7 +77,7 @@
 	if(!.)
 		return FALSE
 
-	if(get_dist_euclide_square(A, owner) > 20)
+	if(get_dist_euclide_square(A, owner) > 36)
 		if(!silent)
 			to_chat(owner, span_xenonotice("You are too far!"))
 		return FALSE


ivanmixo:
Better to fix the bug that stuns you when you lunge and toss instead of this. Lunge is way too oppresive at 6 tiles and it's in a good spot now where it feels both fair to play against and use.

Al-1ce:
Hi. Hello. I'm one of the people who tried to explain why this nerf wasn't a good thing way back when. I'll chip in here. <sub> (See #11317) </sub>

For those who don't know, every third ability used by the warrior activates the `"combo-process"`, which inflicts a variety of debuffs depending on the ability you used. If the third ability used is a punch, for example, it did 50% more damage. This was what was causing insane amount of fracture and has-since been nerfed severely to ~20-15%, which removed the fracture problem. 

Warrior's lunge, which had **_nothing at all_** to do with the actual damage from the combo-process was nerfed as well purely out of salt for the caste. This was stupid for various reasons.

* Lunge had no affect on the damage variable or multiplier
* Lunge's reach had been the same for the entirety of the server's history, with no serious complaints besides salt
* Lunge has one of the longest cooldowns for any combat caste, and _the_ longest cooldown for an active combat ability.
* Lunge stuns the warrior for a duration when used correctly. This was intended to punish cheap tactics and poor planning.
* Lunge often requires the warrior to expose themself to heavy fire as their own ability slows them down.

Out of all of the things you could fairly nerf about warrior, I'd say lunge isn't one of them. It is a very high risk/high reward tool and paints a massive target on your back due to the self-stun. The ability has balanced itself out for a very long time now and the reduction of fishing range from 6-4 was absolutely nothing more than a salt-pr.

Losenis:
Based. Never figured why lunge was the thing nerfed in specific rather than what seemed to be the actual problems.

Pariah919:
Lunge being a cross screen thing again doesn't sit well with me, esp with widescreen being a thing now and xenos having perfect information. I don't think this is really justified in the case of "Well, this wasn't an issue back then!" because it was. 

Warrior is still a pretty good caste as it is right now anyway, buffing lunge back to 6 tiles also makes the caste near unzoneable, slugs are pretty uncommon and citing them as a reason to buff something makes me think that you're specifically having issues with the soft counter to your kit, which is good? I don't see the reasoning behind buffing this back to 6 tiles.

Losenis:
>I don't think this is really justified in the case of "Well, this wasn't an issue back then!" because it was.

Partly, it is justified, because back then widescreen wasn't a thing as you mentioned. Horizontally, it'll still be somewhat worse than back then due to warriors getting spotted and shot up on approach unless marines are pushing directly north/south.

It also makes the ability more reliable because try landing a lunge on a moving object while you yourself are moving due to the short range, it's exceptionally difficult made worse by ping that varies for everyone. Lunge on its own also does literally nothing other than a 1s knockdown while putting you in front of the entire ungaball; if there's an issue, it shold be dealt with whatever ability follows said lunge.

lKiseki:
As the former creator of the PR that is being reverted here, I'd just like to say that I agree with that @Pariah919 is saying in this case. Warrior's Lunge being _6_ tiles of range was quite oppressive and meant it was something you couldn't properly react to. A Warrior can simply reach out from the darkness where it cannot be seen coming, and score some free damage.

I do not believe that reverting back to that sort of exchange is healthy for the game's balance, and would rather keep Lunge as it is currently. Warrior is not particularly in a bad spot in terms of balance, being able to punish marines well enough with its current kit and dealing some decent damage at the same time. There is no need to change that.

Losenis:
>From the darkness

After the massive buff flares got in terms of storage, that kind of falls into skill issue nowadays

Lumipharon:
Warrior is still the best T2 currently, hands down. I don't see what it needs to be buffed back up.

Naaanii:
Missed out on this but,

Salt PR?

Al-1ce:
Bias.

PR #12624, opened 2023-04-05
Cade repair improvements

Lumipharon:
## About The Pull Request
Two changes. Cade repair time and fuel use cut in half (2.5 seconds from 5 seconds, and 1 fuel from 2).
Repair amount reduced to slightly above 50% (85hp from 150).

This should make it easier to do minor repairs, especially if it's difficult to stand there for a prolonged period of time due to boiler acid etc.

There's more ways of damaging cades, in particular damaging multiple cades for xenos, and no more multirepair, so I buffed the repair/second very slightly to make it less guh to try hold cades.
## Why It's Good For The Game
Repair QOL, make cade repairing slightly more effective against modern xeno cade murdering options.
## Changelog
:cl:
balance: Barricade repair time and fuel use cut in half. Repair amount reduced from 150 to 85
/:cl:

### DIFF
--- a/code/game/objects/structures/barricade.dm
+++ b/code/game/objects/structures/barricade.dm
@@ -551,7 +551,7 @@
 	. += span_info("It is [barricade_upgrade_type ? "upgraded with [barricade_upgrade_type]" : "not upgraded"].")
 
 /obj/structure/barricade/metal/welder_act(mob/living/user, obj/item/I)
-	. = welder_repair_act(user, I, repair_threshold = 0.3, skill_required = SKILL_ENGINEER_METAL)
+	. = welder_repair_act(user, I, 85, 2.5 SECONDS, 0.3, SKILL_ENGINEER_METAL, 1)
 	if(. == BELOW_INTEGRITY_THRESHOLD)
 		balloon_alert(user, "Too damaged. Use metal sheets.")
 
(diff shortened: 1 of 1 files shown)


QualityVan:
What ways do xenos have to damage multiple cades at once? Offhand the only ones I can think of are acid globs, which as I understand are pretty zero damage, and ravage.

Autisem:
> What ways do xenos have to damage multiple cades at once? Offhand the only ones I can think of are acid globs, which as I understand are pretty zero damage, and ravage.

globes, ravager ravage, warlock, king screech, prae acid spray, grav grenade

ontop of the various CC abilitys to cancel marine from welding like pychic fling etcv

Lumipharon:
Yeah the most notable changes are warlock psyblast to hit multiple cades. Ravage was also fixed to hit objects, then stuff like king freeze to melt/smack cades.

I'm not really too strong on the slight repair buffs, but just from my own experience it is definitely harder to hold cades then it used to be.

Autisem:
> Yeah the most notable changes are warlock psyblast to hit multiple cades. Ravage was also fixed to hit objects, then stuff like king freeze to melt/smack cades.
> 
> I'm not really too strong on the slight repair buffs, but just from my own experience it is definitely harder to hold cades then it used to be.

IMO other then the buff reducing repair time intervals is not bad to avoid getting constnatly stunned out of it

Losenis:
Holding cades got a bit harder because multirepair is gone even for plasteel cades. Engis can only repair cades one at a time whereas before they could repair every cade they were in range of.

Lumipharon:
> Holding cades got a bit harder because multirepair is gone even for plasteel cades. Engis can only repair cades one at a time whereas before they could repair every cade they were in range of.

Oh yes, this was the main reason I did the PR, then forgot to mention it lol.

PR #12626, opened 2023-04-05
Ointments disinfect

Autisem:
## About The Pull Request

The description says they do, yet they do not. curious. makes them do that thing they said they should

## Why It's Good For The Game

immersion or smth

## Changelog
:cl:
fix: Ointments disinfect now
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/medical.dm
+++ b/code/game/objects/items/stacks/medical.dm
@@ -155,7 +155,7 @@
 	singular_name = "ointment"
 	icon_state = "ointment"
 	heal_burn = 3
-	heal_flags = SALVE
+	heal_flags = SALVE | DISINFECT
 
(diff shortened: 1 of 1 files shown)


chizzy376:
iirc this is leftover flavor text from /tg/, where for wounds you use ointments to treat and disinfect the wounds faster.
Since we lack the wounds system, then /shrug

QualityVan:
No strong opinion about this. On one hand, ointment already helps prevent infections because the burn damage counts as treated afterwards and lore/flavor aren't real reasons for balance changes. On the other, germs from limb damage are tuned in such a way that this wouldn't actually change anything and thus doesn't matter either way.

Lumipharon:
If QV says it literally won't actually do anything, I'd rather just leave it tbh.

PR #12632, opened 2023-04-06
Adjusts some things on drain sting, fixes do_after stacking

RikuTheKiller:
## About The Pull Request

You may have noticed that you can spam resist to immediately rid yourself of the intoxication debuff. I talked about it with Lewd and Ivan and we came to the conclusion that making the resist itself stronger but removing the stacking would be the best way to go.

This PR ups the resist time from 3 seconds to 5 seconds and ups the amount removed from 3 stacks to 8 stacks. This makes the resist 1.6 times stronger than previously. (3 / 3 = 1 and 8 / 5 = 1.6)

The resist timer now also repeats itself on it's own, similarly to repairing things.

This PR also adds messages to the sting since I noticed it lacks them unlike most other abilities, making you unable to see the sting in chat log unless you got buffed by it, which has it's own message.
## Why It's Good For The Game

Resist spam is unintuitive and can be affected by lag, this should even the mechanic out for all players. It's also an unintended "mechanic" that was embraced only because of it's balance implications. (Which pissed off my OCD for whatever reason, leading me to making my first TGMC PR in months, lol.)

This prevents you from ridding yourself of a 30 stack instantly simply because the sentinel missed one shot, that's just plain dumb and completely negates the punishment factor of the debuff. 3 seconds is also short enough for this to be done in the middle of combat, which it shouldn't since it's meant to force you to retreat and stop you from rushing.
## Changelog
:cl:
balance: Resisting intoxication takes 5 seconds instead of 3 seconds and removes 8 stacks instead of 3 stacks. This makes it 1.6 times more effective in general, but harder to use in the middle of combat.
qol: Resisting intoxication now automatically repeats itself, similarly to repairing things.
fix: You can't stack intoxication resists anymore to remove it almost instantly.
/:cl:

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -506,7 +506,7 @@
 		return
 	if(HAS_TRAIT(debuff_owner, TRAIT_INTOXICATION_RESISTANT) || (debuff_owner.get_soft_armor(BIO) > 65))
 		stack_decay = 2
-	var/debuff_damage = SENTINEL_INTOXICATED_BASE_DAMAGE + round(stacks / 10) * is_resisting ? 0.7 : 1 // If they're resisting, reduce the damage taken by 30%.
+	var/debuff_damage = SENTINEL_INTOXICATED_BASE_DAMAGE + round(stacks / 10) * is_resisting ? 0.7 : 1 // If they're resisting, reduce the extra damage taken by 30%.
 	debuff_owner.adjustFireLoss(debuff_damage)
 	playsound(debuff_owner.loc, "sound/bullets/acid_impact1.ogg", 4)
 	particle_holder.particles.spawning = 1 + round(stacks / 2)
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -506,7 +506,7 @@
 		return
 	if(HAS_TRAIT(debuff_owner, TRAIT_INTOXICATION_RESISTANT) || (debuff_owner.get_soft_armor(BIO) > 65))
 		stack_decay = 2
-	var/debuff_damage = SENTINEL_INTOXICATED_BASE_DAMAGE + round(stacks / 10)
+	var/debuff_damage = SENTINEL_INTOXICATED_BASE_DAMAGE + round(stacks / 10) * is_resisting ? 0.7 : 1 // If they're resisting, reduce the damage by 30%
 	debuff_owner.adjustFireLoss(debuff_damage)
 	playsound(debuff_owner.loc, "sound/bullets/acid_impact1.ogg", 4)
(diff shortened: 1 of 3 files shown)


code/datums/status_effects/debuffs.dm
```diff
@@ -519,9 +521,13 @@
 
 /// Resisting the debuff will allow the debuff's owner to remove some stacks from themselves.
 /datum/status_effect/stacking/intoxicated/proc/resist_debuff()
-	if(!do_after(debuff_owner, 3 SECONDS, TRUE, debuff_owner, BUSY_ICON_GENERIC))
+	is_resisting = TRUE
```
ivanmixo, line 524: Never gets set to false. Also use the `do_actions` list on the owner mob instead of a var. (`length(owner.do_actions)`, etc)

RikuTheKiller: i somehow also managed to forget to add a check for returning at all, good job me

RikuTheKiller:
I had a bit of a crisis on whether to keep some of the things in the PR, as you can see in the commit logs.

code/modules/mob/living/carbon/xenomorph/castes/sentinel/abilities_sentinel.dm
```diff
@@ -184,6 +184,8 @@
 	xeno_owner.gain_plasma(drain_potency * 3.5)
 	xeno_owner.do_attack_animation(xeno_target, ATTACK_EFFECT_DRAIN_STING)
 	playsound(owner.loc, 'sound/effects/alien_tail_swipe1.ogg', 30)
+	xeno_owner.visible_message(message = span_xenowarning("\A [xeno_owner] stings [xeno_target]!"), self_message = span_xenowarning("We sting [xeno_target]!"), ignored_mob = xeno_target)
+	to_chat(xeno_target, span_userdanger("You're stung by [xeno_owner]!")) // We proudly tell our target they're fucked.
```
TiviPlus, line 188: Why, the player sees that they get stung and this doesnt give them any necessary long term info

TiviPlus (changes requested):
Rest lgtm ask lewd for his feedback

RikuTheKiller:
> Rest lgtm ask lewd for his feedback

lewd already looked at this

Lumipharon:
The long channel time makes it somewhat guh to use. If that's the intention then all good, otherwise I would say just cut the channel time and stack reduction both in half for qol.

RikuTheKiller:
> The long channel time makes it somewhat guh to use. If that's the intention then all good, otherwise I would say just cut the channel time and stack reduction both in half for qol.

thats the intention, to avoid it being used in combat as much
its meant to make you back away, if you can get rid of it easily while pushing then it's sort of pointless

Lumipharon (approved):

PR #12634, opened 2023-04-06
BAS attachment increases gun size

Losenis:
## About The Pull Request
BAS attachment now has a size increase of 1.

## Why It's Good For The Game
Sholdn't do anything right now after the dispenser nerf, but should make it more difficult to carry 50 BAS T60s in the future if something that fits normal guns comes along.

## Changelog

:cl:
balance: Buildasentry attachment increases object (gun) size by 1.
/:cl:


ivanmixo:
Won't solve the issue since you can just not attach it til you need it

Losenis:
Baby steps. Could make the attachment bulky in another pr. If a storage comes along that can fit it it'd turn into a problem but it'd at least chop the storage of BAS guns down to half due to having 2 bulky objects rather than just 1.

Naaanii:
Should be fine but like ivanmixo said it won't do much unless the attachment itself is addressed.

It should still be able to fit in tech satchels and backpacks for consistency since other sentries fit in it as well.

Lumipharon:
Why's the size mod 3? Just make it 1 like the other size mod attachments.

PR #12636, opened 2023-04-07
Fixes psi point gen

ivanmixo:
## About The Pull Request

Fix the number of gens being universal, sigh. Also only counts marines for the calculations not all humans in existance.  Also fixes being able to recorrupt an already corrupted gen.
## Why It's Good For The Game

Psi gen being correct is good.
## Changelog
:cl:
fix: Fixed psi point gen calculations being wrong. Also you cant corrupt an already corrupted gen anymore.
/:cl:

### DIFF
--- a/code/modules/power/groundmap_geothermal.dm
+++ b/code/modules/power/groundmap_geothermal.dm
@@ -34,8 +34,6 @@
 
 	if(!GLOB.xeno_generators_by_hive)
 		GLOB.xeno_generators_by_hive = list()
-	
-	GLOB.xeno_generators_by_hive[corrupted] += 1
 
 /obj/machinery/power/geothermal/examine(mob/user, distance, infix, suffix)
 	. = ..()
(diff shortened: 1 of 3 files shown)


Naaanii:
Personally think it should include other factions like SOM, Sectoids etc. as they are all hostile to Xenos.

Valhalla mobs should definitely not count (if they're even included).-

code/modules/power/groundmap_geothermal.dm
```diff
@@ -89,8 +93,8 @@
 
 /obj/machinery/power/geothermal/process()
 	if(corrupted && corruption_on)
-		if(length(GLOB.humans_by_zlevel["2"]) > 0.2 * length(GLOB.alive_human_list))
-			SSpoints.add_psy_points("[corrupted]", GENERATOR_PSYCH_POINT_OUTPUT / GLOB.geothermal_generator_ammount)
+		if(length(GLOB.humans_by_zlevel["2"]) > 0.2 * length(GLOB.alive_human_list_faction[FACTION_TERRAGOV]))
+			SSpoints.add_psy_points(corrupted, GENERATOR_PSYCH_POINT_OUTPUT / GLOB.xeno_generators_by_hive[corrupted])
```
DeltaFire15, line 97: This is vulnerable to issues if another hive corrupts the gen after it is decorrupted (e.g. corrupted hive vs. normal hive), as the list of generators per hive does not ever change and thus can cause div-by-0 errors in this case.

Lumipharon (approved):

PR #12637, opened 2023-04-07
Adds new FC Gear Selector vendor. Adds FC vendors to crash ships.

Naaanii:
## About The Pull Request
Title.

Copies all the gear SL can choose from.
Moves the FC Essentials from the FC Equipment vendor to the FC Gear Selector Vendor.
Adds C4 Medihud to FC Essentials.

Currently uses SL vendor sprites but will get unique ones or add them eventually if needed.
## Why It's Good For The Game
FC should have their own vendor of gear to choose from and while this PR doesn't add any powercrept gear or options, it lays down the framework for people to add gear or equipment for FCs to choose from in the future.

FCs should also have medihuds and C4 just like SLs and access to most gear SLs have access to.
Since the role is unique and SL+ people are free to add more in the forseeable future. 
## Changelog
:cl:
add: Added FC Gear Selector Vendor.
add: Adds FC Vendors to all crash ships.
add: Added Medihud and C4 to FC Gear Vendor.
del: Removed Commander Essentials and Jaeger Commander Essentials and replaced it with Standard and Jaeger clothing essentials.
qol: Moved the actual FC Essentials (Medevac, supply beacon etc.) to FC Gear Selector Vendor
/:cl:


code/game/objects/items/cards_ids.dm
```diff
@@ -271,12 +271,18 @@
 		CAT_MEDSUP = MEDIC_TOTAL_BUY_POINTS,
 	)
 
+/obj/item/card/id/dogtag
```
QualityVan, line 274: Should be its own type like the rest instead of redefining dogtag.

code/datums/loadout/loadout_helper.dm
```diff
@@ -56,8 +56,8 @@
 /proc/buy_stack(obj/item/stack/stack_to_buy_type, datum/loadout_seller/seller, mob/living/user, amount)
 	//Hardcode to check the category. Why is this function even here? But it doesn't work, and here I am doing hardcode to make it work because it's hardcoded anyway.
 	var/item_cat = ""
-	if(user.job.title == SQUAD_LEADER)
-		item_cat = CAT_LEDSUP
+	if(user.job.title == SQUAD_LEADER || FIELD_COMMANDER)
+		item_cat = CAT_LEDSUP || CAT_FCSUP
```
QualityVan, line 60: This isn't going to do what you want it to.

QualityVan (changes requested):
No opinions on balance effects, I'm clueless about fc setups.

code/game/objects/items/cards_ids.dm
Naaanii: FC's id uses this

code/datums/loadout/loadout_helper.dm
Naaanii: Reverted

code/game/objects/items/cards_ids.dm
QualityVan: So do PFCs and MPs.

code/datums/loadout/loadout_helper.dm
QualityVan: Did you test vending stacks after changing?

Naaanii: Yes, works fine when I tested.

QualityVan: Tried it myself. Spawned in vendor, bought a stack of sandbags, put it in backpack, saved loadout. Restarted server, vended loadout, no sandbags. Initial purchasing is fine, loadouts aren't.

Naaanii:
> Also I only just noticed that the new vendor isn't mapped in anywhere yet.

Skye will be opening a PR once this is merged as I can't map it for the moment.

code/datums/loadout/loadout_helper.dm
Naaanii: Does it spend the points and not give sandbags?

Naaanii: Nevermind, fixed and tested. Works.

code/game/objects/items/cards_ids.dm
Naaanii: They don't have access but made it its own type anyways.

Mrrpip:
Shockingly, this did not backfire. -Skye

code/__DEFINES/loadout.dm
```diff
@@ -460,8 +487,8 @@ GLOBAL_LIST_INIT(smartgunner_clothes_listed_products, list(
 	))
 
 GLOBAL_LIST_INIT(leader_clothes_listed_products, list(
-		/obj/effect/vendor_bundle/basic_squadleader = list(CAT_STD, "Standard kit", 0, "white"),
-		/obj/effect/vendor_bundle/basic_jaeger_squadleader = list(CAT_STD, "Essential Jaeger Kit", 0, "white"),
+		/obj/effect/vendor_bundle/basic_commander = list(CAT_STD, "Standard kit", 0, "white"),
+		/obj/effect/vendor_bundle/basic_jaeger_commander = list(CAT_STD, "Essential Jaeger Kit", 0, "white"),
```
QualityVan, line 491: Isn't this changing what SLs have available?

QualityVan (commented):
Re: having to maintain parity in future, it only now occurs to me: Could you just give FCs access and points for the SL vendor directly instead of making a new one?

code/__DEFINES/loadout.dm
Naaanii: No, SL defines for their vendor are in a separate file.

Naaanii:
> Re: having to maintain parity in future, it only now occurs to me: Could you just give FCs access and points for the SL vendor directly instead of making a new one?

No because it would require them having the same id tags and access. It would also make changing what both roles receive differently much more cumbersome to use, i.e give FC the choice of Mateba or 1911 etc.

code/__DEFINES/loadout.dm
Naaanii: Fixed

code/__DEFINES/loadout.dm
```diff
@@ -460,7 +487,7 @@ GLOBAL_LIST_INIT(smartgunner_clothes_listed_products, list(
 	))
 
 GLOBAL_LIST_INIT(leader_clothes_listed_products, list(
-		/obj/effect/vendor_bundle/basic_squadleader = list(CAT_STD, "Standard kit", 0, "white"),
+		/obj/effect/vendor_bundle_squadleader = list(CAT_STD, "Standard kit", 0, "white"),
```
QualityVan, line 490: Still there.

QualityVan:
Access is easy, points are just a var on the ID so no trouble either. If you don't plan on keeping the lists the same though that's fine then.

Naaanii:
> Access is easy, points are just a var on the ID so no trouble either. If you don't plan on keeping the lists the same though that's fine then.

Yeah, I'm hoping both roles get their own unique gear. This is just the stepping stone PR so to speak.

RipGrayson (changes requested):
rotated
No.
Otherwise mapping is fine

RipGrayson (approved):

code/__DEFINES/loadout.dm
Naaanii: It shouldn't be wut

PR #12646, opened 2023-04-07
Sentry ammo is now the same weight as the sentry.

Xander3359:
## About The Pull Request
Weight clast = 3 -> WEIGHT_CLASS_NORMAL
Sentry ammo set to WEIGHT_CLASS_NORMAL
## Why It's Good For The Game
It makes no sense for the ammo to be heavier than the sentry.
## Changelog
:cl:
balance: Sentry ammo is now sentry sized (Normal)
/:cl:


Lumipharon (approved):
Honestly the big sentry should be w_class 4 and only fit in tech bags/their crate, so the PD sentry has an actual purpose. But eh, out of scope.

PR #12658, opened 2023-04-09
Adds basic squishing to both the codebase, and the rocket hammer.

GreatKingWombat:
## About The Pull Request

Added the squish element, and applied it to the rocket sledge.
squish

## Why It's Good For The Game

It's funny, and adds something to the codebase that can be used in future projects. Works on marines and xenos.

## Changelog

:cl: QuarianCommando
add: Rocket Sledge now reduces the stature of targets, both literally and metaphorically.
code: Squish element added to datums.
/:cl:

### DIFF
--- a/code/datums/elements/squish.dm
+++ b/code/datums/elements/squish.dm
@@ -1,0 +1,25 @@
+#define SHORT 5/7
+#define TALL 7/5
+
+/**
+ * squish.dm - (Mostly) ported from /tg/. All credit to the original author.
+*/
+
+/datum/element/squish
+	element_flags = ELEMENT_DETACH
+
+/datum/element/squish/Attach(datum/target, duration=20 SECONDS)
+	. = ..()
+	if(!iscarbon(target))
+		return ELEMENT_INCOMPATIBLE
+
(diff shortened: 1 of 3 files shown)


novaepee:
https://user-images.githubusercontent.com/6610922/230753438-1fca9c79-48cf-428b-958f-6e5c96e918bf.mp4

blackdav123:
if you're going to give marines such a huge buff you might as well make the crusher's stomp do this too

GreatKingWombat:
I'm already pushing it enough as is with the rocket sledge. I'd add it to crusher in a heartbeat if I thought I could get away with it.

QualityVan:
Ten seconds seems like a long time, but eh. How much of a nuisance would it be to align to bottom instead of center on the squish?

GreatKingWombat:
It would be an incredible pain in the ass, and outside the scope of the PR. Aside from slapping the effect onto the rocket sledge, the actual squish effect is a /tg/ port. I just had to rename one or two things and remove about half of the functionality to get it to work on baycode.

PR #12662, opened 2023-04-09
Claymore loads up reagent on afterattack.

Xander3359:
Added this var to claymore harvester.
It means that you will automatically load up the weapon on afterattack
## About The Pull Request
Adds a new var loadup_on_attack, which allows the harvester to immediately start the loadup process after a swing.
## Why It's Good For The Game
Claymore is currently the slowest Vali in the game. If you miss a load up due to it's weird timing, you are input locked for quite a long time and you lose out on a chance to apply whatever reagent you wanted to load up.

This syncs up more the swing + load up so it's a lot more seemless, when the loadup is done, the melee is ready to swing again. Just gives it a much nicer flow to it.
## Changelog
:cl:
qol: Claymore will load up reagent on afterattack.
/:cl:

### DIFF
--- a/code/datums/components/harvester.dm
+++ b/code/datums/components/harvester.dm
@@ -17,8 +17,9 @@
 	var/datum/action/harvester/reagent_select/reagent_select_action
 	///The maximum amount that one chemical can be loaded
 	var/max_loadable_reagent_amount = 30
+	var/loadup_on_attack = FALSE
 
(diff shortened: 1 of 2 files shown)


DeltaFire15:
Isn't this actually a disadvantage because you can attack every 1.2 seconds, but the vali chargeup locks you out of actions for 2 full seconds? This seems like something better left to individuals to decide whether to hit the button every single time or not to.

PR #12663, opened 2023-04-09
Green blood harvested from vali scales off of attack speed.

Xander3359:
## About The Pull Request
Makes the amount of green blood you gain per swing dependent on your swing speed.
## Why It's Good For The Game
Seeing how claymore has more than double the attack speed as the other Vali weapons (24/11 = 2.18), it really struggles at maintaining green blood.
Compared to spear, you sacrifice 1 tile of range and lose a lot of DPS for a 1-use 90 damage strike.
This brings it more on par with making the green blood gain somewhat more comparable.
## Changelog
:cl:
balance: Vali weapons now gain green blood dependent on their attack speed. Claymore is the most affected by this.
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -367,7 +367,7 @@
 	return TRUE
 
 ///Handles resource collection and is ativated when attacking with a weapon.
-/datum/component/chem_booster/proc/drain_resource(obj/item/weapon/source, mob/living/M, mob/living/user)
+/datum/component/chem_booster/proc/drain_resource(datum/source, mob/living/M, mob/living/user)
 	SIGNAL_HANDLER
 	if(!isxeno(M))
 		return
@@ -375,7 +375,7 @@
 		return
 	if(resource_storage_current >= resource_storage_max)
 		return
-	update_resource(source.green_blood_drained)
+	update_resource(connected_weapon.green_blood_drained)
 
 ///Adds or removes resource from the suit. Signal gets sent at every 25% of stored resource
 /datum/component/chem_booster/proc/update_resource(amount)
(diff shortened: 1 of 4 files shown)


maikilangiolo:
Do maints even give a single shit about comments on PRs? In the small chance they do, this is a good PR. Double the gain, because you attack twice as slow (yes, you also deal less DPS because it's a single 90 vs two at...54?)

code/datums/components/chem_booster.dm
```diff
@@ -367,15 +367,15 @@
 	return TRUE
 
 ///Handles resource collection and is ativated when attacking with a weapon.
-/datum/component/chem_booster/proc/drain_resource(datum/source, mob/living/M, mob/living/user)
+/datum/component/chem_booster/proc/drain_resource(obj/item/weapon/source, mob/living/M, mob/living/user)
```
TiviPlus, line 370: incorrect casting

code/game/objects/items/weapons/weapon.dm
```diff
@@ -14,6 +14,7 @@
 	var/max_shots = 0 //codex, energy weapons
 	var/scope_zoom = FALSE//codex
 	var/self_recharge = FALSE //codex
+	var/green_blood_drained = 20
```
TiviPlus, line 17: Missing autodoc + if you are changing this due to attack speed why is the resource calculation not just updated to incorporate attackspeed

TiviPlus (changes requested):

code/game/objects/items/weapons/weapon.dm
Xander3359: Sure

code/datums/components/chem_booster.dm
Xander3359: Isn't the source always a weapon?

Xander3359:
Walance change.

PR #12667, opened 2023-04-10
Projectile stun resistance

Lumipharon:
## About The Pull Request
Adds a shitty component to to mitigate hard stuns.

Currently this only applies against projectile based stuns, so is really only relevant to HvH.

What this does is allow some objects (items that currently have damage shielding, and b18 for now) to have a % chance to convert incoming 'hard' stuns (stun and weaken) to 'soft' CC (stagger and slow).

Being staggered or already stunned also reduced your odds of mitigating a stun, so buckshotting the dude multiple times is far more likely to actually put them on their ass.

**Example**

You buckshot some nerd with a shield in the face. The shield has a 100% base chance to mitigate the stun, and they're not staggered, so you mitigate the stun
So instead of **1 weaken, 1 stagger, 0.5 slowdown**, it becomes **0 weaken, 2 stagger, 1.5 slowdown**.

You fire again and they're still staggered, so now they have only a 40% chance to mitigate it. They fail their roll, and take a tumble.

**Walance**
This makes surprise stuns more difficult (at least against opponents with the right countermeasures), but rewards repeated hits or focusing on weakened targets (i.e. pain staggered, flashbanging them first, etc)

I don't have breakpoints so deving in 515 is currently cursed as fuck, but I am 95% sure this doesn't runtime or otherwise fuck out. Trying to debug old issues was utter suffering though. Also I feel like this can probably be done in a nicer way but eh.

Additionally:
Fixes an issue with toggling item active state not actually working.
Fixes toggleable shield items ALSO not working for unrelated reasons
I think there's another issues related to toggling with the existing shield component that I've fixed for this component, but I'll do it in another pr.
## Why It's Good For The Game
Adds some counter measure to hard stuns in HvH makes for more interesting gameplay.
Fixes: Various toggleable items and behavior not working correctly
## Changelog
:cl:
add: HvH: some items such as B18 or shields now have the chance to convert projectile based hard stuns into staggerslow
fix: Fixed item active toggling not working correctly in many cases

/:cl:


Joe13413:
spegmrgeg...... speedmrgege..s....

Al-1ce:
This will still affect HvX, no?
I don't feel like B18 needs any more buffs.

Lumipharon:
> This will still affect HvX, no? I don't feel like B18 needs any more buffs.

Not unless rouny learns to fire a shotgun.

code/datums/components/stun_mitigation.dm
```diff
@@ -0,0 +1,186 @@
...
+		stack_trace("Invalid type found in cover during /datum/component/stun_mitigation Initialize()")
+
+
+/datum/component/stun_mitigation/Destroy()
+	shield_detatch_from_user()
+	cover = null
+	return ..()
+
+///A horrible proc for if the shield is initialised in a slot
+/datum/component/stun_mitigation/proc/shield_init(mob/holder_mob)
```
QualityVan, line 49: I kind of hate this. When are shield comps getting created after an item is already equipped?

code/datums/components/stun_mitigation.dm
```diff
@@ -0,0 +1,186 @@
...
+	if(islist(shield_cover))
+		cover = getArmor(arglist(shield_cover))
+	else if(istype(shield_cover, /datum/armor))
+		cover = shield_cover
+	else
+		cover = getArmor()
+		stack_trace("Invalid type found in cover during /datum/component/stun_mitigation Initialize()")
+
+
+/datum/component/stun_mitigation/Destroy()
```
QualityVan, line 43: Do you not need to clear signals? I remember something about it specific to comps but not what.

code/datums/components/stun_mitigation.dm
```diff
@@ -0,0 +1,186 @@
...
+///Actually activates the mitigation effect
+/datum/component/stun_mitigation/proc/activate_with_user()
+	RegisterSignal(affected, COMSIG_LIVING_PROJECTILE_STUN, PROC_REF(on_attack_stun_mitigation))
+
+///Actually deactivates the mitigation effect
+/datum/component/stun_mitigation/proc/deactivate_with_user()
+	UnregisterSignal(affected, COMSIG_LIVING_PROJECTILE_STUN)
+
+///Handles removing the mitigation from a user
+/datum/component/stun_mitigation/proc/shield_detatch_from_user()
```
QualityVan, line 139: detach

code/datums/components/stun_mitigation.dm
```diff
@@ -0,0 +1,186 @@
...
+
+	if(affected.IsStun() || affected.IsKnockdown() || affected.IsParalyzed())
+		status_cover_modifier *= 0.5
+
+	if(iscarbon(affected))
+		var/mob/living/carbon/C = affected
+		if(C.stagger)
+			status_cover_modifier *= 0.4
+
+	mitigation_prob *= status_cover_modifier
```
QualityVan, line 175: Why set up status_cover_mod at all instead of just multiplying mit prob?

code/game/objects/items.dm
```diff
@@ -1120,7 +1120,7 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 	if(!isnull(new_state))
 		if(new_state == active)
 			return
-		new_state = active
+		active = new_state
```
QualityVan, line 1123: funni

QualityVan (changes requested):

code/datums/components/stun_mitigation.dm
Lumipharon: I did wonder this myself, but the shield component has it (and basically everything that has this component also has the shield one) so I assume it can happen somehow. I didn't actually check though, so maybe not?

code/datums/components/stun_mitigation.dm
Lumipharon: The parent call clears sigs, unless I'm mistaken

code/datums/components/stun_mitigation.dm
Lumipharon: Honk.

code/datums/components/stun_mitigation.dm
Lumipharon: Good catch, I had some other interactions but decided to just keep it simple and it was left over.

code/game/objects/items.dm
Lumipharon: henk

code/datums/components/stun_mitigation.dm
```diff
@@ -0,0 +1,183 @@
...
+
+///Actually deactivates the mitigation effect
+/datum/component/stun_mitigation/proc/deactivate_with_user()
+	UnregisterSignal(affected, COMSIG_LIVING_PROJECTILE_STUN)
+
+///Handles removing the mitigation from a user
+/datum/component/stun_mitigation/proc/shield_detach_from_user()
+	if(!affected)
+		return
+	SEND_SIGNAL(affected, COMSIG_MOB_SHIELD_DETATCH)
```
QualityVan, line 142: de2atch
I know it's an existing signal but it should be easy enough to find/replace.

QualityVan (approved):
Just do a quick testrun to make sure everything still works as it should if you haven't already and should be good.

Lumipharon:
Yeah its good to go.

PR #12672, opened 2023-04-10
Makes large fuel tanks slightly cheaper.

Xander3359:
## About The Pull Request
Large Flamer tank 60 -> 40
Large xfuel tank 300 -> 120
## Why It's Good For The Game
I personally think these are too expensive.
The large tanks contain 15% of the capacity of the back tank, yet are 50% of the cost.
I am reducing the cost to 20% the price of a back tank.
## Changelog
:cl:
balance: Large Flamer tank 60 points -> 40 points
balance: Large xfuel tank 300 points -> 120 points
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -565,12 +565,12 @@ WEAPONS
 /datum/supply_packs/weapons/napalm
 	name = "FL-84 normal fuel tank"
 	contains = list(/obj/item/ammo_magazine/flamer_tank/large)
-	cost = 60
+	cost = 40
 
 /datum/supply_packs/weapons/napalm_X
 	name = "FL-84 X fuel tank"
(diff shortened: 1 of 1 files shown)


Naaanii:
Powercreep.

QualityVan:
Basic fuel is fine, though it's pretty cheap and plentiful as is. For xfuel there's real benefits to using cans instead of a back tank (storage slot stays open, no need to reattach), so I'm not sure about that cut.

Xander3359:
I'm fine with cutting less for xfuel, but I do think it needs *some* kind of cut no matter what. 300 is insane.

Naaanii:
300 isn't really much considering points are easy to get and the value x-fuel provides.

Al-1ce:
Regular fuel is fine. X-Fuel is expensive for a reason. No bueno on that.

Xander3359:
Powercreep is right.

PR #12671, opened 2023-04-10
Adds the x-fuel tank to req

Xander3359:
## About The Pull Request
Adds a 1000u xfuel tank to req
image

It's 300 points for 1000u of xfuel. I made the price cheaper than the tanks because this is ONLY able to refuel xfuel tanks.
## Why It's Good For The Game
Gives you an option to refill your xfuel, rather than just throwing it away.
## Changelog
:cl: @KillerOrcaCora 
add: You can buy an x-fuel tank from req. It's capable of refuelling xfuel tanks
imageadd: x-fuel tank sprites.
/:cl:

PR #12676, opened 2023-04-10
adds the X-fuel fuel tank

Lumipharon:
## About The Pull Request
Remake of #12671

Adds a large fuel tank of X-fuel to req, so you can refill your tanks to save some points.

Added an X-fuel specific explosion for the tank. You uh, probably don't want to be next to it if it explodes.

## Why It's Good For The Game
Recycling good.
Hilarious FF better.

## Changelog

:cl:
add: Adds an X-fuel fuel tank to req. No smoking within 15 meters
/:cl:


KillerOrcaCora:
yaay

code/modules/reqs/supplypacks.dm
```diff
@@ -582,6 +582,12 @@ WEAPONS
 	contains = list(/obj/item/ammo_magazine/flamer_tank/backtank/X)
 	cost = 600
 
+/datum/supply_packs/weapons/fueltank
+	name = "fuel tank"
```
Xander3359, line 586: Granted this is in the weapons tab and not the engineering tab, but it might be worthwhile to distinguish the xfuel tank from a normal fuel tank.

code/modules/reagents/reagents/other.dm
```diff
@@ -401,11 +401,12 @@
 	overdose_threshold = REAGENTS_OVERDOSE
 	overdose_crit_threshold = REAGENTS_OVERDOSE_CRITICAL
 	taste_description = "gross metal"
+	var/obj/effect/effect_type = /obj/effect/decal/cleanable/liquid_fuel
```
TiviPlus, line 404: Autodoc, also dont get why this is cast if its just a normal typepath without initial()

TiviPlus (changes requested):

code/modules/reagents/reagents/other.dm
Lumipharon: I was going to do something different, forgot to change it after I just did it this way.

PR #12678, opened 2023-04-10
razorwire construction option changed to a ratio of 2 metal for 1 razorwire, with 2 razorwire being made with 4 metal

MacBlaze1:
## About The Pull Request

1 units of razorwire now costs 2 metal. This will change how much stuff engineers can build and reinforce. This does not affect razorBURN grenades. Only handbuild razorwire.

## Why It's Good For The Game

Compared to razorburn, razorwire is very expensive. I believe this should make it so that razorwire is good for prepared positions, and razorburn remains good for quick plans. This does not change the actual damage, effects, or the fact crushers can delete razorwire at 50 integrity. Only the cost. 

## Changelog

:cl:
balance: razorwire construction options now makes 2 razorwire instead of 1 for 4 metal.
fix: Changed a min to a max to allow deconstructing razorwire
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/barbed_wire.dm
+++ b/code/game/objects/items/stacks/barbed_wire.dm
@@ -27,10 +27,6 @@
 /obj/item/stack/barbed_wire/full
 	amount = 20
 
-/obj/item/stack/barbed_wire/attackby(obj/item/I, mob/user, params)
-	. = ..()
-
-
 /obj/item/stack/razorwire
 	name = "razor wire assembly"
 	desc = "A bundle of barbed wire supported by metal rods. Used to deny access to areas under pain of entanglement and injury. A classic fortification since the 1900s."
--- a/code/game/objects/items/stacks/barbed_wire.dm
+++ b/code/game/objects/items/stacks/barbed_wire.dm
@@ -27,6 +27,23 @@
 /obj/item/stack/barbed_wire/full
 	amount = 20
 
+/obj/item/stack/barbed_wire/attackby(obj/item/I, mob/user, params)
+	. = ..()
+
+	if(!istype(I,/obj/item/stack/rods))
+		return
+
+	var/obj/item/stack/rods/R = I
+	if(R.amount < 4)
+		to_chat(user, span_warning("You need [4 - R.amount] more [R] to make a razor wire obstacle!"))
+		return
+
+	R.use(4)
+	use(1)
+
+	var/obj/structure/razorwire/M = new /obj/item/stack/razorwire(user.loc, 3)
+	to_chat(user, span_notice("You combine the rods and barbed wire into [M]!"))
+
 /obj/item/stack/razorwire
 	name = "razor wire assembly"
 	desc = "A bundle of barbed wire supported by metal rods. Used to deny access to areas under pain of entanglement and injury. A classic fortification since the 1900s."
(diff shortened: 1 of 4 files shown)


QualityVan:
I guess my question is, even with this change why would you ever use metal for creation over razorfoam grenades?

MacBlaze1:
> I guess my question is, even with this change why would you ever use metal for creation over razorfoam grenades?

pre prepared positions are incredibly powerful, and even though razorburn grenades are large, the ability to place razorwire where you want leads to tactical advantages. For example, placing razorwire behind barricades to prevent crushers from full speed charging or getting stuck. Or to prevent melee of cades. If the cades CANT be damaged, then thats additional EHP.

Mrrpip:
Might as-well ask, as barbed wire affects cade HP and what not, will this have any effects on deconstruction refunds?

MacBlaze1:
> Might as-well ask, as barbed wire affects cade HP and what not, will this have any effects on deconstruction refunds?

Yes. Deconstruction refunds are based off of % of HP * stack amount of mat to construct

QualityVan:
> pre prepared positions are incredibly powerful, and even though razorburn grenades are large, the ability to place razorwire where you want leads to tactical advantages.

Sure, but you can preplace razor wire where you want with grenades too. Not like the stuff decays.

>deconstruction

Oh, yeah, you'll want to adjust razorwire's deconstruction since otherwise I think it's an infinite metal dupe with this change. And the manual razor wire creation via combining barbed wire and rods, since otherwise that'll still use two metal worth of rods.

MacBlaze1:
> > pre prepared positions are incredibly powerful, and even though razorburn grenades are large, the ability to place razorwire where you want leads to tactical advantages.
> 
> Sure, but you can preplace razor wire where you want with grenades too. Not like the stuff decays.
> 
> > deconstruction
> 
> Oh, yeah, you'll want to adjust razorwire's deconstruction since otherwise I think it's an infinite metal dupe with this change. And the manual razor wire creation via combining barbed wire and rods, since otherwise that'll still use two metal worth of rods.

Nah its not an infinite metal dupe, deconstruction is based off of stack amount and obj integrity. Its about 8.33 points per razorburn razorwire, and 8.0 per metal razorwire, which is pretty nice, except currently you CANT deconstruct razorburn razorwire for a full return.

QualityVan (approved):
Turns out my memory's rubbish and you get four rods per metal sheet, not two. Should be good to go unless someone's opposed for balance reasons.

Lumipharon:
I don't really like it because it makes wired cades cheaper. Cades aren't a problem and that change specifically has been rejected before.

QualityVan:
further iteration is required

code/game/objects/items/stacks/barbed_wire.dm
```diff
@@ -30,19 +30,6 @@
 /obj/item/stack/barbed_wire/attackby(obj/item/I, mob/user, params)
 	. = ..()
 
-	if(!istype(I, /obj/item/stack/rods))
```
QualityVan, line 33: You can just cut the rest of this proc since razorwire making was the only thing it did.

QualityVan (changes requested):
Adjust structure/razorwire deconstruction, since it shouldn't be producing barbed wire now. Back into a razorwire stack, or metal/rods if you want to make partial materials more granular. Up to you.

QualityVan (approved):

TiviPlus (approved):

Lumipharon:
Why'd you remove the other construction method btw?

MacBlaze1:
> Why'd you remove the other construction method btw?

@Lumipharon I cant make the cost of razorwire 2 without reducing barbed wires cost, since barbed wire can be used to make razorwire

PR #12679, opened 2023-04-10
Integrates tables with the new smoothing system

RipGrayson:
## About The Pull Request

Title.
Currently tables use a snowflake smoothing system that can occasionally break. 
brokentable
This PR makes them use the current smoothing system instead.

## Why It's Good For The Game
This should fix the problems with table smoothing, even if it doesn't, the newer system is easier to debug than 8 year old spaghetti code we got from CM.

## Changelog
:cl:
fix: Tables should smooth correctly in odd circumstances
code: Integrated tables with the new smoothing system
/:cl:


code/game/objects/structures/tables_racks.dm
```diff
@@ -39,13 +43,18 @@
...
 	for(var/direction in CARDINAL_ALL_DIRS)
 		var/obj/structure/table/table = locate(/obj/structure/table, get_step(location,direction))
 		if(!table)
 			continue
 		INVOKE_NEXT_TICK(table, /atom/proc.update_icon)
+	QUEUE_SMOOTH(src)
 
+/obj/structure/table/Destroy()
+	var/tableloc = loc
+	update_adjacent(tableloc) //so neighbouring tables get updated correctly
```
TiviPlus, line 56:
```suggestion
	update_adjacent(loc) //so neighbouring tables get updated correctly
```

code/game/objects/structures/tables_racks.dm
```diff
@@ -590,11 +486,29 @@
...
+	icon = 'icons/obj/smooth_objects/table_fabric.dmi'
+	icon_state = "table_fabric-0"
+	base_icon_state = "table_fabric"
+	table_prefix = "fabric"
+	parts = /obj/item/frame/table
+	reinforced = TRUE
+
+/obj/structure/table/reinforced/fabric/wrench_act(mob/living/user, obj/item/I)
+	. = ..()
+	return FALSE
```
TiviPlus, line 506: ?

code/game/objects/structures/tables_racks.dm
```diff
@@ -39,13 +43,18 @@
 			new sheet_type(src)
 	return ..()
 
-/obj/structure/table/proc/update_adjacent(location = loc)
+/obj/structure/table/proc/update_adjacent(location = loc, unflipping = FALSE)
```
TiviPlus, line 46: This proc doesnt make sense to me doesnt A) QUEUE_SMOOTH handle everything this proc does B) random new unused arg

TiviPlus (changes requested):

code/game/objects/structures/tables_racks.dm
RipGrayson: update_adjacent is still needed, tables can be flipped on their sides and when one section is destroyed it requires special handling so that the icons around do not break

code/game/objects/structures/tables_racks.dm
RipGrayson: lame attempt to keep the table from being disassembled, removed now

PR #12682, opened 2023-04-10
Fixes resin doors not being damagable in melee

ivanmixo:
## About The Pull Request

Apparently this never called parent but used the stupid hardness system!

Gave doors 50 soft armor against melee so it feels the same as it was pre refactor.

## Why It's Good For The Game

Bug fix.

## Changelog

:cl:
fix: Fixed resin doors not being damagable in melee
/:cl:

### DIFF
--- a/code/game/objects/structures/mineral_doors.dm
+++ b/code/game/objects/structures/mineral_doors.dm
@@ -76,11 +76,10 @@
 
 
 /obj/structure/mineral_door/attackby(obj/item/W, mob/living/user)
-	if((W.flags_item & NOBLUDGEON) && !W.force)
-		attack_hand(user)
+	. = ..()
+	if(QDELETED(src))
 		return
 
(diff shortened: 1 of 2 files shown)


QualityVan:
Not sure what you mean with regard to the melee armor pre-refactor. If you're talking about the bonus damage from burn, that's still in the attackby.

QualityVan:
Oh, got it, moving from 1.5/2 hardness to 100/160 integrity. Any particular reason you didn't push them up to 150/200 integrity in the refactor? I guess because that would apply to projectiles/fire too.

ivanmixo:
The goal is to keep them the exact same. Making doors more tanky to projectiles wouldn't be good since it'd be a buff to doors (which they don't need).

Me not doing this in the refactor was an oversight in the first place. Didn't notice it didn't call parent at all.

PR #12688, opened 2023-04-11
Vali green blood gained scales off the weapon attack speed.

Xander3359:
## About The Pull Request
Vali Green blood gains reagent based on the attack speed of the weapon used.
## Why It's Good For The Game
This way you no longer get punished with less green blood for choosing a slower weapon.
## Changelog
:cl:
balance: Vali green blood gain scales off weapon attack speed.
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -367,7 +367,7 @@
 	return TRUE
 
 ///Handles resource collection and is ativated when attacking with a weapon.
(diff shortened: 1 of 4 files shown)


Naaanii:
Doesn't this negate the point of lower attack speed?

Xander3359:
> Doesn't this negate the point of lower attack speed?

I don't think starving claymore users of green blood was the intent when it was slowed down to 24 attack speed.
If it was, maintainers can simply close this PR.

PR #12695, opened 2023-04-11
Swaps dexaline for isotonic in lifesaver belt

Losenis:
## About The Pull Request
Swaps the dexaline pill bottle in lifesaver belt for an isotonic solution pill bottle

## Why It's Good For The Game
Dexaline's use is redundant in most cases when you have inaprovaline. As it is, the default loadout has no real blood replenishing options so a corpsman has to go out of their way to get one. Shouldn't be too big a change given both are available from vendors anyway, so more of a qol change than balance.

## Changelog

:cl:
balance: Dexaline pill bottle in lifesaver belts replaced with Isotonic Solution pill bottle
/:cl:


QualityVan (approved):
Was going to bring up lack of an available treatment if someone gets stuck suffocating at 50+ ox but that's not actually a thing that happens.

TiviPlus (approved):

PR #12708, opened 2023-04-12
Debloats random spawner icons [MDB Ignore] 

RipGrayson:
## About The Pull Request

Title. 
Currently all our random spawner icons live in the massive and unsorted landmarks_static.dmi file. I have split all the spawners into multiple parts, both in code and icon.
Additionally I have removed a number of ancient icons that we have no use for and never will. (We still have landmark icons hanging around for Clock Cult and Blob spawns to name just two)

## Why It's Good For The Game

By splitting the files up it makes it both easier to add new random spawners and eliminates the performance downside from using a giant single .dmi file for several hundred icons.

## Changelog
:cl:
imagedel: Deleted some ancient unused landmark icons.
code: Reorganized random spawners into smaller individual sections.
/:cl:


Autisem:
right in the soul

Lumipharon (approved):

PR #12713, opened 2023-04-13
Thirds non-laser CAS weapon costs.

dopamiin0:
## About The Pull Request

Thirds all non-laser CAS weapon costs in the printer.

## Why It's Good For The Game

All shipmaps start with two of each non-laser weapon already, and using more is never efficient, as it means you are not using some of the starting PO ammo. This enables fringe gimmick strategies (FIVE 30MMS BAYBEE) without really powercreeping the PO, as they still have to actually pay for all the ammo they want to use.

## Changelog
:cl:
balance: Divided the cost of non-laser CAS weapons by three.
/:cl:

### DIFF
--- a/code/game/objects/structures/dropship_ammo.dm
+++ b/code/game/objects/structures/dropship_ammo.dm
@@ -329,6 +329,13 @@
 	ammo_type = CAS_LASER_BATTERY
 	cas_effect = /obj/effect/overlay/blinking_laser/laser
 
+/obj/structure/ship_ammo/laser_battery/focused
+	name = "focused laser battery"
+	desc = "A specialized laser battery designed for extreme precision. Moving this will require some sort of lifter."
+	ammo_used_per_firing = 10
+	laze_radius = 1
+	point_cost = 100
+
 /obj/structure/ship_ammo/laser_battery/examine(mob/user)
 	. = ..()
 	. += "It's at [round(100*ammo_count/max_ammo_count)]% charge."
--- a/code/game/objects/structures/dropship_ammo.dm
+++ b/code/game/objects/structures/dropship_ammo.dm
@@ -329,13 +329,6 @@
 	ammo_type = CAS_LASER_BATTERY
 	cas_effect = /obj/effect/overlay/blinking_laser/laser
 
-/obj/structure/ship_ammo/laser_battery/focused
-	name = "focused laser battery"
-	desc = "A specialized laser battery designed for extreme precision. Moving this will require some sort of lifter."
-	ammo_used_per_firing = 10
-	laze_radius = 1
-	point_cost = 100
-
 /obj/structure/ship_ammo/laser_battery/examine(mob/user)
 	. = ..()
 	. += "It's at [round(100*ammo_count/max_ammo_count)]% charge."
--- a/code/game/objects/structures/dropship_equipment.dm
+++ b/code/game/objects/structures/dropship_equipment.dm
@@ -859,7 +859,7 @@
 	desc = "A dismounted GAU-21 'Rattler' 30mm rotary cannon. It seems to be missing its feed links and has exposed connection wires. Capable of firing 5200 rounds a minute, feared by many for its power. Earned the nickname 'Rattler' from the vibrations it would cause on dropships in its inital production run. Moving this will require some sort of lifter."
 	icon_state = "30mm_cannon"
 	firing_sound = 'sound/weapons/gunship_chaingun.ogg'
-	point_cost = 100
+	point_cost = 150
 	dropship_equipment_flags = USES_AMMO|IS_WEAPON|IS_INTERACTABLE
 	ammo_type_used = CAS_30MM
 
@@ -887,7 +887,7 @@
 	desc = "A rocket pod weapon system capable of launching a single laser-guided rocket. Moving this will require some sort of lifter."
 	firing_sound = 'sound/weapons/gunship_rocket.ogg'
 	firing_delay = 5
-	point_cost = 100
+	point_cost = 200
 	ammo_type_used = CAS_MISSILE
 
 /obj/structure/dropship_equipment/weapon/rocket_pod/deplete_ammo()
@@ -911,7 +911,7 @@
 	icon = 'icons/Marine/mainship_props64.dmi'
 	firing_sound = 'sound/weapons/gunship_rocketpod.ogg'
 	firing_delay = 10 //1 seconds
-	point_cost = 100
+	point_cost = 200
 	ammo_type_used = CAS_MINI_ROCKET
 
 /obj/structure/dropship_equipment/weapon/minirocket_pod/update_icon()
@@ -935,7 +935,7 @@
 	icon = 'icons/Marine/mainship_props64.dmi'
 	firing_sound = 'sound/weapons/gunship_laser.ogg'
 	firing_delay = 50 //5 seconds
-	point_cost = 450
+	point_cost = 600
 	dropship_equipment_flags = USES_AMMO|IS_WEAPON|IS_INTERACTABLE
 	ammo_type_used = CAS_LASER_BATTERY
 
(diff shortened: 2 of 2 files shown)


capsaicinz:
based cas economy buff

Al-1ce:
>The laser is in an interesting spot. It does decent damage and fills an interesting role as a precision alternative for CAS, but rarely finds itself used by more experienced POs, as it requires a whopping 685 points to get online, and simply doesn't offer enough OOMPH to make this hefty cost worth it. 


You said yourself that it does "decent damage", but also say that it doesn't do "enough" to make it viable.

Let me be clear here; most CAS systems aren't intended to be crazy dangerous. The Condor is a support tool. What makes the laser viable is the accuracy and the very minimal strike time.

I don't find making the price of the laser 1-1 with other weapons fair, because it outperforms them both in every way. 

>For this, I have changed two things - I made the base laser gun 150 points cheaper, to make its use more realistic especially in medium and high orbit

After someone revived my PR that allowed CAS to get points from miners, orbit-level barely matter outside of fuel economy.

I can't imagine you're talking about the time it takes for the laser to hit, since at any orbit it is still near instantaneous.


>and I have added a second battery type for when even the normal laser's AOE is too large. This creates a weapon that, with skill and precision, can realistically be used in areas where marines are close by

CAS having _pinpoint_ accuracy is something that has been discussed and tried before. The AOE of all CAS systems are intended hinderances, because it is a support tool - not something meant to wipe xenos clean.

Things like this are antithetical to every other principle CAS has followed and has since been following.

dopamiin0:
> You said yourself that it does "decent damage", but also say that it doesn't do "enough" to make it viable. What makes the laser viable is the accuracy and the very minimal strike time, which nigh-all other weapons lack.
> 
You say "what makes the laser viable" but in my experience it simply isn't. POs rarely use it because it has an incredibly high entry barrier, and those same points could be better spent using cheaper, more reliable options. I explicitly avoided buffing damage in this PR because that's not the laser's niche. The laser is not unique in its minimal strike time - low orbit highvel and widowmakers are similarly rapid and much easier to hit.
> I don't find making the price of the laser 1-1 with other weapons fair, because it outperforms them both in every way.
> 
It's not 1-1. It still has a high barrier to entry. For reference, the new cost, not including ammo, can purchase 3 Banshees, or 6 crates of 30mm, or can otherwise be spent in any other number of ammunition options. I lowered it by 150 points. This is a buff, don't get me wrong, but the laser isn't used very commonly to begin with.
> 
> After someone revived my PR that allowed CAS to get points from miners, orbit-level barely matter outside of fuel economy.
>
This is simply false. Passive point gen is still a significant part of total point gen, especially in games where marines are unable to control more than one or two miners, which are the same games in which your point gen is that much more important! (It doesn't matter if you have a billion points if marines are wiping xenos hard, as they'll be in caves and with a jillion req points anyways - it's just a win more.) The mining point buff was a nice enhancement to CAS, but the economy and orbit level are still very important things to consider. 
> I can't imagine you're talking about the time it takes for the laser to hit, since at any orbit it is still near instantaneous. In-fact, at higher orbits, I'd expect pilots to be using the laser over all else. Which has since become plausible, now that miners add to the printer.
> 
You hold a weird argument here that I don't entirely get? You're talking about people always being on low orbit now because of the miner, but now pilots are also comfortably at high orbit laser spamming? Regardless, I think it's fine for high orbit pilots to be able to afford the fancy toys - that's the point of high orbit. 
> 
> CAS having _pinpoint_ accuracy is something that has been discussed and tried before. The AOE of all CAS systems are intended hinderances, because it is a support tool - not something meant to wipe xenos clean.
> 
> Things like this are antithetical to every other principle CAS has followed and has since been following.

Here I'll concede that the precision battery may have been a mistake. I was unaware similar things had been tried before. I will remove this from the PR to keep it to just the other aspects.

Naaanii:
Powercreep lol.

iforgotmeword:
You guys are going to you own cas you know the dragon T4 xeno

dopamiin0:
> Powercreep lol.

I assume you're talking about the laser here - I rarely see it used, due to the astronomical entry cost. This is simply a buff to make a less competitive option more viable. Laser simply wasn't "meta" for CAS, when you have to pay an absurd sum to even use it. 450 is still a high cost (535 if you include the cost of the first battery) but it's more fitting of the laser, IMO - it's still an expensive tool you have to consider, but it no longer chews up your entire budget for the round. The laser isn't a magic "I win" button.

Naaanii:
> > Powercreep lol.
> 
> I assume you're talking about the laser here - I rarely see it used, due to the astronomical entry cost. This is simply a buff to make a less competitive option more viable. Laser simply wasn't "meta" for CAS, when you have to pay an absurd sum to even use it. 450 is still a high cost (535 if you include the cost of the first battery) but it's more fitting of the laser, IMO - it's still an expensive tool you have to consider, but it no longer chews up your entire budget for the round. The laser isn't a magic "I win" button.

The cost was increased because of it's instant fire and high damage. The instant fire was made to not be instant, the damage has remained the same iirc but its still a great kill and zoning tool in the right hands. It's still a powerful weapon in the right hands and the notorious 4 laser build is still viable, granted you sink points into it. You're still guaranteed one laser and battery by 12:45 either way along with having a decent arsenal of free roundstart stuff. If anything, you'd probably have a better chance of adding a free laser and battery roundstart on all shipmaps, which would still be a significant buff.

Reducing the cost of guns for more than 5 times the amount is also wack and I'll admit it's not even powercreep but a straight up buff. It makes POs not require to really invest much at all and freely spam weapons, granted our POs aren't the brightest of bulbs but this seems unnecessary for an already good role.

dopamiin0:
Meh. I personally think this would hardly be problematic and would enable some fun options to CAS pilots that are currently less than ideal, but popular consensus seems to disagree and arguing is tiring.

dopamiin0:
i just had a round where i decided to try an all one weapon meme strategy and i spent over an hour on high orbit with several miners for literally nothing because CAS weapons are ridiculously overpriced. running memes is already economically difficult due to the simple fact that you start with an even spread of ammo for different weapons.

i removed the laser change because apparently everyone and their mother has been instantkilled by 80 lasers at once or something.

FosterGrime-Heymynameslouie:
blessed be, cas not giving me a heart attack from thinking about its price

Naaanii:
This is still powercrept.

Al-1ce:
I'll reconsider. This is worth a TM at the very least

dopamiin0:
i dont know why this is saying it isnt successful this changes 3 numbers. im going to assume that its fine unless told otherwise. github is my mortal enemy

Lumipharon:
PO's were given a bunch of extra weapons and ammo across all maps, they used to get less (except for one map, I forget which), and now get funny miner points, so they're in a better position than they've ever been.

dopamiin0:
> PO's were given a bunch of extra weapons and ammo across all maps, they used to get less (except for one map, I forget which), and now get funny miner points, so they're in a better position than they've ever been.

Respectfully, I ask that you reconsider this.

As-is, two out of the three weapons cost the same as printing the laser, with the third being close. There is no logical reason why you would ever want to spend the same amount of points as it costs to print the laser to print a weapon that you already have two of. This doesn't cheapen ammo or increase the amount of CAS a PO can do in the round. There are only two reasons why a CAS PO would ever be printing a non-laser weapon: either a) they were griefed by a marine, in which case they shouldn't have their job invalidated and should be able to easily access replacements; or b) they want to use _more than two_ of the same weapon, which I don't see a strong or compelling reason why this should be soft-banned by making the weapon cost as expensive as it is. They don't get any more ammo to work with, they still have to pay for ammo (which is where CAS costs should really be going anyways.)

The fact that the laser costs the **_same_** as a third minirocket pod is ridiculous. Letting POs run weird builds won't own xenos harder than the PO could have owned them anyways.

Issue #12371, opened 2023-03-10
When you start entering the hypersleep chamber in the evac pod, you can get ghosted but still cancel entering the pod


CCorvet:
### Description

I'd expect for the forced ghost to happen only once you actually enter the pod, not after a delay when you start entering it, as that can be cancelled.

### Test Merges

Server-Side Code Modifications: https://tgstation13.org/parsed-logs/sybil/server_side_modifications.dm

Round ID: [20094](https://scrubby.melonmesa.com/round/20094)
BYOND Version: 514.1589
Server revision compiled on: 
Master commit: [ff86174466cb798a34b9ca0274dfe4a1e137cd15](https://github.com/tgstation/TerraGov-Marine-Corps//commit/ff86174466cb798a34b9ca0274dfe4a1e137cd15)
The following pull requests are currently test merged:
[#12200: 'Adds more loadouts to EORD' by carlarctg at commit 527d1bf109](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12200)

Local commit: 004ac78c47ed97fd8bd1597d73753ab2dee9e4c4
Server tools version: 5.2.3

### Reproduction Steps

1. Go to an evac shuttle (Dont recall it working for other cryo pods)
2. Start entering a hypersleep pod
3. Move away from it before you enter.
4. Still get ghosted, as your SSD body is standing there like a dummy.

PR #12715, opened 2023-04-13
Fixes Evac cryo pod ghosting you immediately if you start getting into it.

Xander3359:
## About The Pull Request
Closes https://github.com/tgstation/TerraGov-Marine-Corps/issues/12371
## Why It's Good For The Game
bugfix
## Changelog
:cl:
fix: Cancelling the channel to enter a cryo pod on the evac shuttle will no longer ghost you.
/:cl:

### DIFF
--- a/code/game/objects/machinery/cryopod.dm
+++ b/code/game/objects/machinery/cryopod.dm
@@ -262,7 +262,7 @@
 	if(helper && user != helper)
 		if(user.stat == DEAD)
 			to_chat(helper, span_notice("[user] is dead!"))
-			return
+			return FALSE
 
 		helper.visible_message(span_notice("[helper] starts putting [user] into [src]."),
 		span_notice("You start putting [user] into [src]."))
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/cryopod.dm
```diff
@@ -279,10 +279,12 @@
 		return
 
 	user.forceMove(src)
-
 	occupant = user
 	update_icon()
 
+	if(istype(src, /obj/machinery/cryopod/evacuation)) //Will ghost you if it's an evac pod
+		user.ghostize(FALSE)
+
```
TiviPlus, line 287: Why are you istype srcing instead of just return TRUE/FALSE on success/fail and checking that?

TiviPlus (changes requested):
.

code/game/objects/machinery/cryopod.dm
Xander3359: Where should I be checking it from?

code/modules/shuttle/escape_pod.dm
```diff
@@ -159,7 +159,8 @@
 
 /obj/machinery/cryopod/evacuation/climb_in(mob/living/carbon/user, mob/helper)
 	. = ..()
-	user.ghostize(FALSE)
+	if(. == TRUE)
```
TiviPlus, line 162:
```suggestion
	if(.)
```

TiviPlus (changes requested):

PR #12716, opened 2023-04-13
Tac-reloadable guns are opaque (so when in hands)

Xander3359:
## About The Pull Request
Guns that are in-hand are opaque. 
## Why It's Good For The Game
This makes it easier to click on them when you are trying to insert/remove a magazine.
This makes it easier to click drag a magazine onto the sprite when attempting to do a tactical reload.
## Changelog
:cl:
qol: Equipped guns are easier to click on
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -430,6 +430,10 @@
 		return
 	drop_connected_mag(chamber_items[current_chamber_position], user)
 
+/obj/item/weapon/gun/unequipped(mob/user, slot)
+	. = ..()
+	mouse_opacity = MOUSE_OPACITY_ICON //So that it doesnt remain opaque when you drop or discard the gun
+
 ///Set the user in argument as gun_user
 /obj/item/weapon/gun/proc/set_gun_user(mob/user)
 	active_attachable?.set_gun_user(user)
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -413,12 +413,12 @@
 		O.emp_act(severity)
 
 /obj/item/weapon/gun/equipped(mob/user, slot)
-	unwield(user)
 	if(ishandslot(slot))
 		set_gun_user(user)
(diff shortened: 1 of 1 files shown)


XElectricX:
Won't this be a problem if a gun is multitile?

QualityVan:
When is this being unset? Unless they'll just stay opaque forever after being equipped once.

code/modules/projectiles/gun_system.dm
```diff
@@ -413,6 +413,8 @@
 		O.emp_act(severity)
 
 /obj/item/weapon/gun/equipped(mob/user, slot)
+	if(ishandslot(slot) || slot == SLOT_BACK || slot == SLOT_BELT || slot == SLOT_S_STORE)	//add more if needed
```
QualityVan, line 416: Are there slots you'd equip it to without wanting it to be opaque? This just seems like the list of places guns fit.

Xander3359: I've set it to the places where you would equip guns because these are where guns are tac-reloadable.
In testing, if you click drag a magazine on a gun, but you miss the gun sprite, you will fail your tac reload. 
You can tac reload a gun on back/belt/suit storage or hands, so I'm making it opaque to match those slots.

It also just makes it easier to click on the gun sprite to draw out of those slots.

code/modules/projectiles/gun_system.dm
```diff
@@ -413,11 +413,12 @@
...
 /obj/item/weapon/gun/equipped(mob/user, slot)
-	unwield(user)
 	if(ishandslot(slot))
 		set_gun_user(user)
-		return ..()
-	set_gun_user(null)
+		mouse_opacity = MOUSE_OPACITY_OPAQUE
+	else
+		set_gun_user(null)
+	unwield(user)
```
Lumipharon, line 421: Why'd you move this? Check procs before moving them around, this effects the user, so if you're unsetting gun user before calling this you're going to break a bunch of stuff.

Lumipharon (changes requested):

Lumipharon (approved):

PR #12717, opened 2023-04-13
Fixes out of hand tac reloads

QualityVan:
## About The Pull Request
If both one hand and parent gun are empty, you could tac reload even when you aren't actually holding the gun.
## Why It's Good For The Game
bugfix
## Changelog
:cl:
fix: You need to be holding a gun to tac reload it
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_helpers.dm
+++ b/code/modules/projectiles/gun_helpers.dm
@@ -104,7 +104,7 @@ As sniper rifles have both and weapon mods can change them as well. ..() deals w
 			return
 		to_chat(user, span_warning("[new_magazine] cannot fit into [src]!"))
 		return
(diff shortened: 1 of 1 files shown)


Strayed-Raven:
This is a marine nerf considering if you get delimb you can not longer reload. Pretty sure it was intended and not a bug.

QualityVan:
You can still do it with one hand missing so long as the gun's in the hand you still have, and if you're missing both hands then not being able to reload is not a major concern.

Strayed-Raven:
> You can still do it with one hand missing so long as the gun's in the hand you still have, and if you're missing both hands then not being able to reload is not a major concern.

From a gameplay aspect removing an option that has no effect on balance is odd, but alright

PR #12718, opened 2023-04-13
Makes bursts correspond to the hivenumber of the hugger

ivanmixo:
## About The Pull Request

So the larva that pops out is the same hive as the hugger that did the hug.

## Why It's Good For The Game

Other hive carriers hugging people shouldn't make larvas of the normal hive pop out.

## Changelog

:cl:
fix: The burst larvas have the same hive as the hive of the carrier who did the hug.
/:cl:


QualityVan:
Might as well include source while you're at it, though looking at it now the only thing that matters for is currently bugged.

QualityVan (approved):

Autisem:
Does this work properly when hugger is hatched from an egg from another hive?

ivanmixo:
yes

Naaanii:
Hugger turrets fixed too?

PR #12730, opened 2023-04-15
Makes transfusion less annoying to use

RikuTheKiller:
## About The Pull Request

This PR makes transfusion not cancel if the target moves and cancel if the target takes damage. It does not get cancelled if the target takes damage while immobilized. It's also cancelled if the target is knocked into crit during it, even if the target is immobilized.

On a more technical note, I've also split IGNORE_LOC_CHANGE into IGNORE_TARGET_LOC_CHANGE and IGNORE_USER_LOC_CHANGE and turned IGNORE_LOC_CHANGE into a combination of the two. This shouldn't have any effects in-game.
## Why It's Good For The Game

Okay, so transfusion is a great ability and all, but it's just really frustrating to use. Why? Because xenos don't stay still. Like, at all. When they're building, they move 1 tile exactly when your ability is about to finish. Even when you'd think they'd be idle, they keep moving anyways since it seems to be engrained in every xeno player's brain that standing still is equal to death.

Now, you may say "just tell them to stop moving" and to that I say "getting them to listen takes 5 times longer than the ability itself, not to mention constantly asking people to stop moving is obnoxious" which is why I made this PR.

The change to make it cancel when the target takes damage is so that the change can't be abused, which really isn't needed at the moment. It's already a strong ability that is owned by a strong caste, it does not need buffs.

This PR also solidifies transfusion as an ability to be used outside of combat, except in the specific case where your target is immobilized due to fortify, which is a combo I don't want to remove.
## Changelog
:cl:
balance: Gorger's transfusion ability is no longer cancelled by the target moving. In exchange, it's cancelled if the target takes damage. (Unless the target is immobilized.)
/:cl:


RikuTheKiller:
Also this was tested in-game with jimmyl and most, if not all edge cases have been taken care of.

mc-oofert:
> Also this was tested in-game with jimmyl and most, if not all edge cases have been taken care of.

me

Arctrooper8:
Gorger enthusiasts rejoice!

code/modules/mob/living/carbon/xenomorph/castes/gorger/abilities_gorger.dm
```diff
@@ -180,6 +180,8 @@
 		KEYBINDING_NORMAL = COMSIG_XENOABILITY_TRANSFUSION,
 	)
 
+	var/target_health //Used to keep track of the target's previous health for extra_health_check()
```
Lumipharon, line 183: autodoc

code/modules/mob/living/carbon/xenomorph/castes/gorger/abilities_gorger.dm
```diff
@@ -202,8 +204,15 @@
 		if(!silent)
 			to_chat(owner, span_notice("We can only help living sisters."))
 		return FALSE
-	if(!do_mob(owner, target_xeno, 1 SECONDS, BUSY_ICON_FRIENDLY, BUSY_ICON_MEDICAL))
+	target_health = target_xeno.health
+	if(!do_mob(owner, target_xeno, 1 SECONDS, BUSY_ICON_FRIENDLY, BUSY_ICON_MEDICAL, ignore_flags = IGNORE_TARGET_LOC_CHANGE, extra_checks = CALLBACK(src, PROC_REF(extra_health_check), target_xeno)))
+		return FALSE
+	return TRUE
+
+/datum/action/xeno_action/activable/transfusion/proc/extra_health_check(mob/living/target)
```
Lumipharon, line 212: autodoc

Lumipharon (changes requested):

Lumipharon (approved):

code/__DEFINES/mobs.dm
```diff
@@ -817,8 +817,10 @@ GLOBAL_LIST_INIT(human_body_parts, list(BODY_ZONE_HEAD,
 #define SCREWYHUD_HEALTHY 3
 
 //do_mob() flags
-#define IGNORE_LOC_CHANGE (1<<0)
-#define IGNORE_HAND (1<<1)
+#define IGNORE_LOC_CHANGE IGNORE_USER_LOC_CHANGE | IGNORE_TARGET_LOC_CHANGE
```
Lumipharon, line 820:
```suggestion
#define IGNORE_LOC_CHANGE (IGNORE_USER_LOC_CHANGE|IGNORE_TARGET_LOC_CHANGE)
```
honk. This break uh, a lot of stuff.

Lumipharon (changes requested):

Lumipharon (approved):

PR #12731, opened 2023-04-15
Changes FAMAS stats to be more in line with other rifles.

GreatKingWombat:
## About The Pull Request

Removes the damage multiplier from the FAMAS, adds 50% more scatter, and increases the movement penalty on accuracy.

## Why It's Good For The Game

FAMAS is currently just using up LOC since its' removal from seasonals, since nobody spends req points on import weapons. If the weapon being dogshit is what it takes for it to be used again, then so be it.

## Changelog

:cl: QuarianCommando
balance: FAMAS stats adjusted to be more in-line with other rifles.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -782,8 +782,9 @@
 	burst_delay = 0.15 SECONDS
 	accuracy_mult = 1.15
 	wield_delay = 0.5 SECONDS
-	scatter = 1.5
-	movement_acc_penalty_mult = 5
+	damage_mult = 1.1
+	scatter = 1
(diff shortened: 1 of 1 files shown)


Pariah919:
Why are you nerfing this to make it fit into seasonals? This was always supposed to be a ERT gun first, the import high-expense part is because its an ERT gun so it has superior stats, this just makes the rifle into a shitty carbine clone rather than a high mobility CQC gun with a lot of punch.

GreatKingWombat:
I don't even want it nerfed. I just want it to be allowed into seasonals so it can be used.

Lumipharon:
No to nerfing ERT guns just so you can add them to marines.

PR #12733, opened 2023-04-16
Fixes the psi gen fix fix by making the math actually not stupid

ivanmixo:
## About The Pull Request

Makes the psi output based off the ratio of currently corrupted generators and generators on the ground. Why wasn't it like this? I have no idea. 

If you manage to corrupt the gens shipside you get extra psi, wow.
## Why It's Good For The Game

Psi gen math being sane is neat.
## Changelog
:cl:
fix: Fixes the psi gen fix fix by making the math actually not stupid. Psi is based off ratio of generators corrupted to uncorrupted.
/:cl:

### DIFF
--- a/code/modules/power/groundmap_geothermal.dm
+++ b/code/modules/power/groundmap_geothermal.dm
@@ -3,6 +3,8 @@
 #define GEOTHERMAL_MEDIUM_DAMAGE 2
 #define GEOTHERMAL_HEAVY_DAMAGE 3
 
+GLOBAL_VAR_INIT(generators_on_ground, 0)
+
 /obj/machinery/power/geothermal
 	name = "\improper G-11 geothermal generator"
 	icon = 'icons/turf/geothermal.dmi'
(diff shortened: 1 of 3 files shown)


Lumipharon:
psypoints scaling with gen number is useful for larger vs smaller maps imo.

ivanmixo:
(Just ignore how research outpost and big red have the same number of gens)

Lumipharon:
that's a mapping issue though, not a code one.

code/modules/power/groundmap_geothermal.dm
```diff
@@ -95,7 +105,8 @@
 /obj/machinery/power/geothermal/process()
 	if(corrupted && corruption_on)
 		if((length(GLOB.humans_by_zlevel["2"]) > 0.2 * length(GLOB.alive_human_list_faction[FACTION_TERRAGOV])) && GLOB.xeno_generators_by_hive[corrupted] > 0)
-			SSpoints.add_psy_points(corrupted, GENERATOR_PSYCH_POINT_OUTPUT / GLOB.xeno_generators_by_hive[corrupted])
+			//You get points proportional to the % of generators corrupted (for example, if 66% of generators are corrupted the hive gets 0.66 points per second)
+			SSpoints.add_psy_points(corrupted, GENERATOR_PSYCH_POINT_OUTPUT * (GLOB.xeno_generators_by_hive[corrupted] / GLOB.generators_on_ground) / GLOB.xeno_generators_by_hive[corrupted])
```
QualityVan, line 109: You're multiplying a value by a hive's generator count and then dividing it by same. That does a grand total of nothing.

QualityVan:
Agree that point gen scaling with the ratio of owned objectives rather than total count feels against precedent. It's still an improvement over the current flat value if you have any at all, admittedly, because that's very weird.

Also, this seems like something that should happen in hive datums maybe? They don't actually process though.

code/modules/power/groundmap_geothermal.dm
ivanmixo: Uh no? 

Okay let's go step by step.

`GENERATOR_PSYCH_POINT_OUTPUT` is defined as 1.

We multiply `GENERATOR_PSYCH_POINT_OUTPUT` by the number of gens the hive has divided by the number of groundside generators. This gives us a % of the number of generators groundside that are corrupted. Nice. Now we have to divide that by the number of corrupted generators so each generator gives it's part of the psi instead of every generator giving the full amount (remember, this is per generator). We don't want to give 10 psi per second if there's 10/10 generators corrupted. 

So let's say there's 10 generators groundside and 5 are corrupted.  1 *  (5/10) / 5. We get 0,1 psi per generator. 
Added up for 5 generators it comes out to 0,5. It checks out.

QualityVan: Yeah, because doing nothing by definition doesn't change the output. 
N*X*(1/Y)*(1/X)=N*(X/X)*(1/Y)=N/Y
Cut both generator_by_hive values and see if the result changes at all.

PR #12735, opened 2023-04-17
Smokening

Lumipharon:
## About The Pull Request
1) All smoke can now flow through any type of smoke unimpeded.

This was previously limited to just defiler smoke. Gone will be the days of one stray smoke tile blocking an entire boiler glob through a doorway.

2) Smoke that crosses paths with smoke of the same type will merge with the existing smoke instead of stacking more smoke on top of it

~~This behaves like fire. It will stack in duration up to a maximum of x2 the original duration~~, which means you don't get cucked by trying to resmoke 0.5 seconds too early and having the old smoke completely block it.
Edit: Now just goes up to the existing duration or duration of the incoming smoke, so its a little less oppressive to have 5 acid globs thrown on top of you.

This is a fairly significant balance change that positively and negatively effects both sides, so I'd want to TM this to see how it actually plays out, although I suspect it will be Xenos that gain the most from this.
## Why It's Good For The Game
Less fucky smoke interactions, behaves more like you would expect.
## Changelog
:cl:
balance: Smoke can now flow past other existing smoke clouds
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -170,7 +170,7 @@
 	amount = parent.amount-1
 	origin = parent.origin
 	if(merge)
-		lifetime = min(parent.lifetime, lifetime)
+		lifetime = max(parent.lifetime, lifetime)
 		strength = max(parent.strength, strength)
 	else
 		lifetime = parent.lifetime
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -133,7 +133,7 @@
 			if(!moveable.CanPass(src, T))
 				can_spread = FALSE
 				break
-			if(!istype(moveable, src.type))
+			if(!(moveable.type == src.type))
 				continue
 			new_smoke = moveable
 			if(origin == new_smoke.origin) //part of the same smoke bloom
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -19,17 +19,20 @@
 	var/bio_protection = 1 // how unefficient its effects are against protected target from 0 to 1.
 	var/datum/effect_system/smoke_spread/cloud // for associated chemical smokes.
 	var/fraction = 0.2
-	var/smoke_can_spread_through = FALSE
 	///Delay in ticks before this smoke can affect a given mob again, applied in living's effect_smoke
 	var/minimum_effect_delay = 1 SECONDS
+
+	var/origin
 
 	//Remove this bit to use the old smoke
 	icon = 'icons/effects/96x96.dmi'
 	pixel_x = -32
 	pixel_y = -32
 
-/obj/effect/particle_effect/smoke/Initialize(mapload, range, smoketime, smokecloud)
+/obj/effect/particle_effect/smoke/Initialize(mapload, range, smoketime, smokecloud, parent)
 	. = ..()
+	if(parent)
+		origin = parent
 	if(smokecloud)
 		cloud = smokecloud
 		LAZYADD(cloud.smokes, src)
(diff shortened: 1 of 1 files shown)


DeltaFire15:
I don't know what you mean with advantages for both sides, there isn't really any offensive smoke that marines use where stacking would be useful. On the other hand, for xenos, this completely unchains all smoke castes and turns something that could be called powerful, and sometimes pretty daunting, into definitely oppressive, aswell as removes the only counterplay marines had for heavy use of smoke deploying castes.

Lumipharon:
> I don't know what you mean with advantages for both sides, there isn't really any offensive smoke that marines use where stacking would be useful. On the other hand, for xenos, this completely unchains all smoke castes and turns something that could be called powerful, and sometimes pretty daunting, into definitely oppressive, aswell as removing the only counterplay marines had for heavy use of smoke deploying castes.

You can now chain smokes together, to extend duration of plasma smoke, or plasma smoke while in other smoke etc. Previously you would effectively waste your smoke if it was blocked by any other smoke.

Losenis:
>Quad defiler stacking all the chems in one go

image

Lumipharon:
Defilers already can do that.

Pariah919:
honestly this is pretty cringe and kills a ton of emergent gameplay, I dislike this

Araulius:
welp, standard smoke nades are kill

Autisem:
This does change quiet alot of balance around not just boiler but stacking as in general.

Would still rather have this be an toggolable option but its would be interseting

novaepee:
If this gets merge, just make a marine tool that allow to block xenomorph gas. Simple.

CCorvet:
> This just makes more sense and should've been default behaviour all along, esp considering smoke is literally a requirement for splitting up ungaballs; though, it could be cool to have some kind of specialised smoke dispersal tool/machinery on the marine side as a counter beyond "take mimir lol"

big fan, boom

Ragecrabbo:
### **_The fog is coming_**

Losenis:
>Defilers already can do that.

Can they? I thought if 4 defilers vented gas in the same area they didn't overlap.

shelteredfreak:
All xeno gasses can expand inside eachother, acid, neuro, and defiler gasses. This PR would make stacking gasses more effective due to increasing the duration, as well as allowing Xeno gasses to pass through marine gasses like cloak, smoke, and tangle. 

P much a straight buff to Defiler and Boiler. Which is a good or bad thing depending on how you view balance. I'm biased and would like to see this go through, or at the very least be TM'd.

Lumipharon:
> All xeno gasses can expand inside eachother, acid, neuro, and defiler gasses. This PR would make stacking gasses more effective due to increasing the duration, as well as allowing Xeno gasses to pass through marine gasses like cloak, smoke, and tangle.
> 
> P much a straight buff to Defiler and Boiler. Which is a good or bad thing depending on how you view balance. I'm biased and would like to see this go through, or at the very least be TM'd.

Specifically, defiler gases, and defiler gases only, allow OTHER gases to flow through them. So defilers can stack gas, and a boiler can shoot into defiler gas etc, but a defiler can gas into a boiler cloud.

QualityVan (approved):
Agree that this kills the only use for smoke grenades in current hvx, but it's cleaner as a system because the way smoke determined passthrough was weird and this is just more intuitive.

QualityVan:
It's a little out of scope for this particular pr, but what if generic smoke grenade smoke got a flag to destroy any other clouds in its process?

Lumipharon:
> It's a little out of scope for this particular pr, but what if generic smoke grenade smoke got a flag to destroy any other clouds in its process?

Easy enough, but yeah I'd rather do any other changes in a future pr.

Also normal smoke isn't even useful for that now, since you can just use plasma or cloak smoke for the same effect.

JackTheJackhammer:
gonna be boiler hell

Strayed-Raven:
This gives absolutely no counter to mass spam of Xeno gas. Boiler and Defiller stacking is just ggnore now because of this PR. I would say give some kind of weight to each gas so it does not fully bleed into every other smoke so marines smoke can at least still be a shield. otherwise Mimir should never be nerfed after this.

SiegeB0t:
I am a fan of the smoke denial interactions and this reduces boiler counter play options to just tanking boiler globs with your face. Perhaps we can fix this by triggering boiler globs prematurely on explosion or bullet hit.

JackTheJackhammer:
Just played a round with this and the gas spam is iNSANE. Its 1914 out there!

code/game/objects/effects/effect_system/smoke.dm
```diff
@@ -170,7 +170,7 @@
 	amount = parent.amount-1
 	origin = parent.origin
 	if(merge)
-		lifetime = min((parent.lifetime + lifetime), initial(lifetime) * 2)
+		lifetime = min(parent.lifetime, lifetime)
```
QualityVan, line 173: Shouldn't this be max?

Lumipharon: Yes, honk

Derrick2Electricboogaloo:
Damn smoke nades are useless now.

PR #12754, opened 2023-04-20
Shoulder holster can no longer carry mp19

Xander3359:
## About The Pull Request
This removes MP19 from fitting inside the shoulder holster webbing.
## Why It's Good For The Game
MP19 is not a pistol, it doesn't fit in pistol belt, it should not fit in pistol holster webbing.
## Changelog
:cl:
balance: MP19 no longer fits in pistol webbing.
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/uniform.dm
+++ b/code/modules/clothing/modular_armor/attachments/uniform.dm
@@ -146,8 +146,6 @@
 		/obj/item/ammo_magazine/pistol,
 		/obj/item/weapon/gun/revolver,
 		/obj/item/ammo_magazine/revolver,
-		/obj/item/weapon/gun/smg/standard_machinepistol,
-		/obj/item/ammo_magazine/smg/standard_machinepistol,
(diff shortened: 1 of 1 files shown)


Lumipharon (approved):

Al-1ce:
If you can stuff six of them into a standard backpack I don't see why they shouldn't be able to fit into the shoulder-holster. 

They are small and intended as a TEC-9 ripoff.. let it be.
-1

Xander3359:
MP19 is not a pistol. If it were a pistol, it would fit in pistol belt and be in the sidearm category.
Shoulder holster is supposed to carry pistols.

Al-1ce:
> MP19 is not a pistol. 
>If it were a pistol, it would fit in pistol belt.

o_o
there is zero point to be made here
of course it isn't a pistol

> Shoulder holster is supposed to carry pistols.

It's of equal size. I don't see the problem. You can sling a little machinegun in a shoulder just as you can a pistol.

PR #12757, opened 2023-04-20
Nerfs Defile strength scaling

lKiseki:
## About The Pull Request
Nerfs the scaling for Defiler's strength based on existing chems. Instead of multiplying its strength by 2 for every chem present, it will now do so by 1.5.
## Why It's Good For The Game
Defile being such an instakill sentence with just two chems is probably not okay.

## Changelog
:cl: Lewdcifer
balance: Defile strength scaling per chem reduced from 2 to 1.5.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defiler/abilities_defiler.dm
@@ -132,7 +132,7 @@
 					continue
 				else
 					neuro_applied = TRUE
-			defile_strength_multiplier *= 2
+			defile_strength_multiplier *= 1.5
 	if(living_target.has_status_effect(STATUS_EFFECT_INTOXICATED))
 		var/datum/status_effect/stacking/intoxicated/debuff = living_target.has_status_effect(STATUS_EFFECT_INTOXICATED)
 		defile_reagent_amount += debuff.stacks


Al-1ce:
Ehh. I don't know how I feel about this. Defiler IS powerful but squishy. The risky nature of their caste compared to every other T3 but rav has always defile fair in my eyes.

I think the issue here is the primordial ability _guaranteeing_ free defiles where otherwise not possible, which has made much of the castes stratagem redundant, and up'd the lethality 200%.

Al-1ce:
I think the issue here is the primordial ability guaranteeing free defiles where otherwise not possible, which has made much of the castes stratagem redundant, and up'd the lethality 200%.

Passe2142:
Its a finisher, good vs valis, b18s, and pre medders. It does punish triage too done right. Maybe increase the CDs for missing?

Lumipharon:
How much does defile actually do, with the aforementioned 2 chems?

I generally wouldn't want to nerf it considering its a channeled melee ability.

shelteredfreak:
We did it boys, Defiler got recognized enough that it got a salt PR. Mission accomplished.

Jokes aside. I hope you mean chem chems and not toxin chems. Because Defiler can only give 2 toxins at a single time via Gassing and Reagent Slashing both. 21u of slash maximum and a variable amount from gas. In general, if you are the one who somehow gets hit with all of this, and **also** getting Defile'd? Safe to say that you deserve it. Gas is the only chem that applies to multiple people at once, Defiler otherwise has to focus on one person at a time or just pepper small amounts among the crowd with reagent slashes. 

TL:DR If you really want to nerf Defiler, there are better ways to do it. Touching gas or reagent slash would likely be less finnicky than Defile, because of it being a whole combo ender.

Araulius:
the squishiest tier 3 that has to melee to do anything does not need a nerf. If you hate dealing with chemicals, simply borg yourself.

PR #12764, opened 2023-04-20
Fixes synth erroneously being marked as "cold" on both minimap and examine

ivanmixo:
## About The Pull Request

Because synths do not go cold!
## Why It's Good For The Game

Bug bad. 
## Changelog
:cl:
fix: Fixes synth erroneously being marked as "cold" on both minimap and examine
/:cl:


Lumipharon:
Isn't this going to cause problems if a synth does actually dnr?

PR #12767, opened 2023-04-21
Removes the medscanner gloves from the Marinemed

PestoVerde322:
## About The Pull Request

This PR removes the new Medscanner gloves from the normal MarineMed, as really it shouldn't make sense that a common marine can get their hands on it.

[updating the pic, please stand by]

## Why It's Good For The Game

It's more of a balancing stuff, since it's pretty much a flat upgrade from the default marine gloves as it is right now.

## Changelog

:cl:
removed: Removed the Medscanner gloves from the Marinemed
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -1113,7 +1113,6 @@
 		),
 		"Misc" = list(
 			/obj/item/defibrillator = 8,
-			/obj/item/healthanalyzer/gloves = 8,
 			/obj/item/healthanalyzer = 16,
 			/obj/item/bodybag/cryobag = 24,
 		),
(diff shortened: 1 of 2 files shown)


Lumipharon:
Just kill it in vendors entirely, it should have never been put in them.
Medics can pay a paltry amount of points to buy it, and non combat healing roles like doctors don't need it because they don't have any issues with storage.

Ragecrabbo:
> Just kill it in vendors entirely, it should have never been put in them. Medics can pay a paltry amount of points to buy it, and non combat healing roles like doctors don't need it because they don't have any issues with storage.

This honestly, if they are just a better version of regular gloves with no downsides then giving them a point cost and putting them in corpsman vendor would be the best solution.

Lumipharon (approved):

PR #12771, opened 2023-04-22
Deletes duplicate proc def and unused var of nonexistant type on the antagonist datum

ivanmixo:
## About The Pull Request

`/datum/team` isn't a type that exists. The `Topic()` proc was defined twice. The var was unusued so I commented it out because I assume we'll want teams.
## Why It's Good For The Game

Unused vars of non existant type are bad.
## Changelog
:cl:
fix: Deleted duplicate proc def and unused var of nonexistant type on the antagonist datum
/:cl:

### DIFF
--- a/code/modules/antagonists/_common/antag_datum.dm
+++ b/code/modules/antagonists/_common/antag_datum.dm
@@ -197,7 +197,6 @@ GLOBAL_LIST_EMPTY(antagonists)
 /datum/antagonist/custom
 	antagpanel_category = "Custom"
 	show_name_in_check_antagonists = TRUE //They're all different
-	///var/datum/team/custom_team - TODO: actually implement teams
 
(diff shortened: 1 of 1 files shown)


code/modules/antagonists/_common/antag_datum.dm
```diff
@@ -197,35 +197,15 @@ GLOBAL_LIST_EMPTY(antagonists)
 /datum/antagonist/custom
 	antagpanel_category = "Custom"
 	show_name_in_check_antagonists = TRUE //They're all different
-	var/datum/team/custom_team
+	///var/datum/team/custom_team - TODO: actually implement teams
```
QualityVan, line 200: Just cut it entirely if it isn't in use now, it can be added back later easily enough.

ivanmixo: sure

PR #12773, opened 2023-04-22
Fixes two args having the same name in the squeak component

ivanmixo:
## About The Pull Request

Uh oh, how will the proc know which we're referring to!

Found via opendream:tm:
## Why It's Good For The Game

Arg ambiguity isn't.
## Changelog
:cl:
fix: Fixes two args having the same name in the squeak component
/:cl:


Lumipharon (commented):
Does the arg even need to be there?

ivanmixo:
The signal sends 2 args so that's why it has 2.

TiviPlus (approved):

PR #12786, opened 2023-04-23
Medical Rig belt has isotonic instead of dexalin

Xander3359:
## About The Pull Request
Medical Rig belt has isotonic instead of dexalin
## Why It's Good For The Game
Isotonic > Dex
Also, same reasoning as https://github.com/tgstation/TerraGov-Marine-Corps/pull/12695
"Dexaline's use is redundant in most cases when you have inaprovaline. As it is, the default loadout has no real blood replenishing options so a corpsman has to go out of their way to get one. Shouldn't be too big a change given both are available from vendors anyway, so more of a qol change than balance."
I feel like this also applies to doctors (or corpsman who opt to take the medrig belt instead of lifesaver)
## Changelog
:cl:
balance: Medical rig belt has isotonic instead of dexalin.
/:cl:


TiviPlus:
Where why its good for the game

QualityVan (approved):

PR #12812, opened 2023-04-24
Synth vendor no longer has the base satchel and IMP pack

Xander3359:
## About The Pull Request
Removes 2 noob trap options from the synth vendor, IMP and satchel.
## Why It's Good For The Game
You spawn with a satchel as a synth so there is no reason to have it in the vendor.
The IMP backpack is a noob trap when it's infinitely vendable in the clothes vendor.
I only left the industrial backpack because I dont think it's vendable elsewhere despite being only a slight reskin.
## Changelog
:cl:
del: Synth vendor no longer has satchel or IMP backpack.
/:cl:


QualityVan:
There's a ton of clothes available in starting vendors which are infinitely available elsewhere, I don't really see it as an issue.

Xander3359:
It's a noob trap that only serves to fuck over players that are new to Synth.
Unlike clothing options, there's no gameplay problem if you pick a different hat or jumpsuit.
But having a normal backpack when you can vend a combat backpack is getting trolled.

PR #12817, opened 2023-04-24
Remove three miners close to landing zone in Icy Cave redux

novaepee:
## About The Pull Request

image

## Why It's Good For The Game

Icy Cave is under maintained, therefore it produces rounds where marines benefit more from taking a walk to miners from landing zone. We've done this with other maps for the same reason.

Xander notes that xenomorphs cannot weed these area until marines open shutters, making these miners easy to capture and gain req points faster than any planetside maps.

## Changelog

:cl:
del: Remove three miners close to landing zone in Icy Cave
balance: Remove three miners close to landing zone in Icy Cave
/:cl:

### DIFF
--- a/_maps/map_files/icy_caves/icy_caves.dmm
+++ b/_maps/map_files/icy_caves/icy_caves.dmm
@@ -2794,10 +2794,6 @@
 /obj/effect/landmark/start/job/survivor,
 /turf/open/floor/plating/ground/snow/layer0,
 /area/icy_caves/outpost/outside/center)
-"pD" = (
-/obj/machinery/miner/damaged,
-/turf/open/floor/plating/ground/snow/layer2,
-/area/icy_caves/outpost/outside)
 "pE" = (
 /obj/machinery/atmospherics/pipe/simple/green/hidden,
 /obj/structure/cable,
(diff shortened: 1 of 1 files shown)


RipGrayson (approved):

Al-1ce:
north miner is like 13 steps away from the fob
probably should just remove/relocate it

Al-1ce:
The two are the right are close enough to share turret range.

What silly goober made this?

Oscari-pog:
Always found those miners silly

EvilMelonTheSecond:
Wouldn't this reverse the problem and make the colony grounds fully useless for marines? The medbay is the only thing worth holding and the rest is in the caves or too far apart to be reasonable. South of the colony has always been ignored by marines and I believe removing miners instead of spreading them around will just lead to the ENTIRE colony being ignored

Al-1ce:
> Wouldn't this reverse the problem and make the colony grounds fully useless for marines? The medbay is the only thing worth holding and the rest is in the caves or too far apart to be reasonable. South of the colony has always been ignored by marines and I believe removing miners instead of spreading them around will just lead to the ENTIRE colony being ignored

3 miners being removed, 2 of which are _fourteen_ steps away from the FOB does not make the colony grounds "useless."

PR #12825, opened 2023-04-26
Big Red LZ1 Changes

novaepee:
## About The Pull Request

All Big Red LZ1 shall have 4 or 3 flanks as par standard of landing zones NOT to exceed or equal to 5 flanks. 

Some of Big Red LZ1 map modules have either 4 or 3 flanks, reduced from 5 flanks.

Annihilate this potential breach to FOB from all Big Red LZ1 map modules.

image

Change part of northern cave tunneling to LZ1 in light of reducing flank.

image

Allow xenomorphs to have an easier time making a funny hole from marshal to caves linking to LZ1.

image

## Why It's Good For The Game

No LZ with the exception of Big Red have more than 4 flanks. RipGrayson noted that Shiryu added an additional flank for some reason.

Due to the reduced flank, I ensure that xenomorphs can still contest against marines' FOB.

## Changelog

:cl:
balance: Some of Big Red LZ1 map modules have either 4 or 3 flanks, reduced from 5 flanks. This means that there are less fronts for xenomorphs to engage LZ1 but still maintaining the standard 3 or 4 flanks that engineers must consider when building FOB.
balance: Allow xenomorphs to have an easier time making a funny hole from marshal to caves linking to LZ1.
/:cl:

### DIFF
--- a/_maps/modularmaps/big_red/bigredlzvar1.dmm
+++ b/_maps/modularmaps/big_red/bigredlzvar1.dmm
@@ -733,6 +733,9 @@
 /area/bigredv2/outside/space_port)
 "tg" = (
 /obj/structure/closet/secure_closet/injection,
+/obj/structure/sign/safety/hazard{
+	dir = 1
+	},
 /turf/open/floor,
 /area/bigredv2/outside/marshal_office)
 "to" = (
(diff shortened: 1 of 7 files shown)


chizzy376:
At least add some metal rods to where the window used to be, replacing the entire thing with glass shards makes it look like the steel frame just disappeared

RipGrayson (approved):

PR #12846, opened 2023-04-27
Makes the AI able to fire and aim railguns

RipGrayson:
## About The Pull Request

Title.
This PR allows AI to aim and fire the railgun at the ground via control + shift click. 

In order to keep it balanced, I've given it some downsides compared to a SL with binos doing it. Namely the AI can only fire a single volley of 10 shots (subject to adjustments, a human can fire the railgun until disturbed or the ammo runs out) before cooldown, has to wait 30 seconds after setting the laser down before it fires and cannot fire within 10 tiles of a marine. It also cannot play mind games by placing a laser down and then not firing on it. Once a laser is placed the volley will be fired unless the AI is damaged during the process. 

EDIT: After some thought I have changed the railgun volley the AI releases from 5 shots to 10. It's a 6 minute cooldown before they can fire again, I don't think unreasonable that it gets a few extra rounds.

## Why It's Good For The Game

It's just a cool concept that allows the AI to be slightly more useful during a match. Sure 5 shots won't do much, but the AI can in rare circumstances provide a little covering fire for a retreat. Besides it makes sense thematically that the AI would be able to aim and fire the railguns onboard, it already controls all shipside electronics and can get coords from the battlefield, thematically it should be able to fire the guns at a point on the ground.

## Changelog
:cl:
qol: Made the AI capable of firing shipside railguns at the ground.
/:cl:


Gregormack:
Yes!! More abilities for the mighty AI!

RipGrayson:
> Yes!! More abilities for the mighty AI!

Someday the QoL changes will add up and the AI will become a viable role like any other, instead of an eye in the sky only able to yell at marines not to enter caves.

Guidesu:
Can you make AI able to use orbital spotlight?

DeltaFire15:
How exactly is it supposed to use this if it cannot laze anything within ten tiles of marines when those are its source of vision? I suppose it might work while piloting drones, but seems fairly questionable nonetheless.

RipGrayson:
> How exactly is it supposed to use this if it cannot laze anything within ten tiles of marines when those are its source of vision? I suppose it might work while piloting drones, but seems fairly questionable nonetheless.

I lowered the range down to 5, should make things somewhat easier.

code/_onclick/ai.dm
```diff
@@ -218,3 +225,51 @@
...
+/turf/AICtrlShiftClick(mob/living/silicon/ai/user)
+	///var to keep us from queuing a million railgun shots and breaking the system, do_afters don't work well with AI actions
+	var/static/busy = null
+	var/laz_name = user.name
+	var/obj/effect/overlay/temp/laser_target/laser
+	var/area/A = get_area(user.eyeobj.loc)
+	if(busy)
+		to_chat(user, span_warning("The rail guns are already targeting a location, wait for them to finish."))
+		return
+	if(locate(/mob/living/carbon/human) in range(AI_RAILGUN_HUMAN_EXCLUSION_RANGE, get_turf(user.eyeobj.loc)))
```
TiviPlus, line 237: get_humans_cheap?

code/_onclick/ai.dm
```diff
@@ -218,3 +225,51 @@
...
+		return
+	if(A.ceiling > CEILING_METAL)
+		to_chat(user, span_warning("DEPTH WARNING: Target too deep for ordnance."))
+		return
+	if((GLOB.marine_main_ship?.rail_gun?.last_firing + 300 SECONDS) > world.time)
+		to_chat(user, "[icon2html(src, user)] [span_warning("The rail gun hasn't cooled down yet!")]")
+	else if(!A)
+		to_chat(user, "[icon2html(src, user)] [span_warning("No target detected!")]")
+	else
+		to_chat(user, span_notice("Firing orbital railguns at [loc], COORDINATES: X:[x] Y:[y]"))
```
TiviPlus, line 251: early returns

code/_onclick/ai.dm
```diff
@@ -218,3 +225,51 @@
 		message += " It is underground."
 	to_chat(user, span_notice(message))
 
+/turf/AICtrlShiftClick(mob/living/silicon/ai/user)
+	///var to keep us from queuing a million railgun shots and breaking the system, do_afters don't work well with AI actions
+	var/static/busy = null
```
TiviPlus, line 230: This is very gross

code/_onclick/ai.dm
```diff
@@ -218,3 +225,51 @@
...
+	var/laz_name = user.name
```
TiviPlus, line 231: you can just use user.name directly no

code/_onclick/ai.dm
```diff
@@ -218,3 +225,51 @@
...
+		to_chat(user, span_warning("The rail guns are already targeting a location, wait for them to finish."))
+		return
+	if(locate(/mob/living/carbon/human) in range(AI_RAILGUN_HUMAN_EXCLUSION_RANGE, get_turf(user.eyeobj.loc)))
+		to_chat(user, span_warning("Friendly biologicals detected near [loc] target location, aborting."))
+		return
+	if(!is_ground_level(user.eyeobj.z) || isdropshiparea(A)) //can't fire the railgun off the ground level, or at the DS
+		to_chat(user, span_warning("Incompatible target location."))
+		return
+	if(A.ceiling > CEILING_METAL)
+		to_chat(user, span_warning("DEPTH WARNING: Target too deep for ordnance."))
```
TiviPlus, line 244: some of these tochats could be balloons

TiviPlus (changes requested):
Rest good enough I guess

code/_onclick/ai.dm
RipGrayson: I tried a bunch of things, including balloon_alert_to_viewers and couldn't do this one. The alert won't follow the eyeobj, even if you directly specify that as where the alert should spawn. I had brief luck with using balloon_alert_to_viewers, but there's a cutoff a certain distance away from the AI core.

code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -16,6 +16,8 @@
 	hud_type = /datum/hud/ai
 	buckle_flags = NONE
 	has_unlimited_silicon_privilege = TRUE
+	///are we already firing a railgun laser
+	var/is_laser_firing = FALSE
```
TiviPlus, line 20: Could make this a trait

RipGrayson: Done, checks are failing because CHAT_FILTER_CHECK is missing. I have no idea why it's missing, I haven't touched it, but it's failing on multiple PRs other than this one.

ivanmixo:
Could we get a countdown til next available fire in game tab or such?

RipGrayson:
> Could we get a countdown til next available fire in game tab or such? Also its deceptive with the way it fires - it fires on the tile its on when it requires shiftctrl click to use? Why? Oh also the shift click requires you to click directly on the turf which is rather difficult with weeds in the way

All issues addressed in the latest patch, thanks for the feedback.

Lumipharon:
When I saw this in action, the AI ROF on the railgun seemed to be notable faster than than when humans were using it, which is a very weird inconsistancy.

RipGrayson:
> When I saw this in action, the AI ROF on the railgun seemed to be notable faster than than when humans were using it, which is a very weird inconsistancy.

Rate of fire is one second faster for two reasons.
1. The AI railgun has a fixed volley size, it shouldn't take 20 seconds to land 10 shots. While the underlying idea for the railgun is support, giving the AI an uninterruptible ability to deny an area for 20 seconds is too long.
2. More importantly, due to the way lasers work they automatically delete themselves after a time. This means you cannot do a volley above a size of (8ish? I think that's the limit when tested) because it takes nearly 16 seconds to finish and the laser deletes itself if you go above that time. The solution is to cut the time between shots down from 2 seconds to 1 so the volley always finishes.

Lumipharon:
My view is that it should be more or less exactly the same as marine fired, only for a shorter duration/max shots. (as xenos have zero counterplay, unlike a marine which they can attack/interrupt)

Its still the same actual railgun firing, so having a different fire rate/bypassing the normal CD etc is really weird.

The laser is a temp effect so the lifetime is specified by a var on the effect itself.

RipGrayson:
> My view is that it should be more or less exactly the same as marine fired, only for a shorter duration/max shots. (as xenos have zero counterplay, unlike a marine which they can attack/interrupt)

I get what you're saying, and originally we started with that philosophy, but for gameplay practicalities keeping the AI and human railgun behavior 100% the same doesn't play well.

> 
> Its still the same actual railgun firing, so having a different fire rate/bypassing the normal CD etc is really weird.

This is an example of what I'm talking about above, logically it makes sense for them both to share the same cooldown, it's the same gun firing. 

In practice though, the AI and the field marines fight over who can land a railgun volley. If a marine lases for a railgun volley so much as once, the AI cannot fire for 5 minutes. Likewise the AI can flatly deny groundside the ability to fire the railgun for 5 minutes by firing as soon as its cooldown is done. (And in my experience the AI usually will, given that firing the railgun is its one ability that makes a direct impact on the battlefield.)

It's nothing but frustration and usually leads to lots of yelling and threats between the AI and whoever is trying to use the railgun groundside.
> 
> The laser is a temp effect so the lifetime is specified by a var on the effect itself.

I'm aware of the var, but during testing with a much higher number it didn't seem to matter for some reason. Regardless, having the AI be able to deny an area for 20 seconds or more seems a little much to me. 

I think the solution to all of this, if we had the sprites and code, would be to separate the AI firing action from the railgun and give it its own piece of equipment (like a shipside laser cannon). Then it wouldn't have to be linked to the railgun and the logical inconsistencies you mentioned wouldn't be a problem.

1kajowsmol:
pls add

Pockets-byte:
> I think the solution to all of this, if we had the sprites and code, would be to separate the AI firing action from the railgun and give it its own piece of equipment (like a shipside laser cannon). Then it wouldn't have to be linked to the railgun and the logical inconsistencies you mentioned wouldn't be a problem.
I definitely like this idea. giving the AI some less slow tangible way to affect the battlefield that isn't drones would be a massive boon

PR #12864, opened 2023-04-30
More realistic water and lava

Lumipharon:
## About The Pull Request
Mobs no longer float over liquid turfs like lava and water.
image
image

https://user-images.githubusercontent.com/7869430/235412314-85235d7a-2b88-45d9-a4a0-a1e0ea64317f.mp4

The transition time is tied to your slowdown, so should hopefully look fairly good regardless of speed.
It may need some tweaking around border tiles to ensure best visual appearance however (i.e. you won't go under water on a shallow tile facing south, east or west, which may or may not be the visually best option).

There is functionality to alter the 'depth' a mob sits under a liquid, although I'm not using it right now.

Also repathed both lava and watermore generally, so that all liquid turfs now actually are children of a single liquid turf parent, instead of several different copies under different type paths.

## Why It's Good For The Game
No more Jesus larp
## Changelog
:cl:
add: Mobs no longer float above water or lava
code: repathed water and lavato remove duplicate unrelated variations
/:cl:


Gregormack:
Holy smokes, that's great.

Al-1ce:
Can rouny hide under water?

Lumipharon:
> Can rouny hide under water?

It works for all mobs, but no, rouny won't go entirely under.

XSlayer300:
How about fire and smoke, etc?

ilikesaying:
when do we get to sink into the abyss

Losenis:
This only applies to mobs, right? Reducing the sprite of mobs already sounds wack when you need to click them for certain things but I'd dread to see things like claymores be invisible.

Derrick2Electricboogaloo:
This is amazing.

Lumipharon:
> How about fire and smoke, etc?

Only mobs at this stage.

BlueHNT:
> How about fire and smoke, etc?

Not being able to set water on fire with basic fire sounds about right
Smoke-based stuff can still be atop liquids since it's not in the water... But if some certain damaging smokes won't be able to interact with feet sound like a funny detail

Al-1ce:
> It works for all mobs, but no, rouny won't go entirely under. 

I think they SHOULD be able to.

KerbalCitizen:
> > How about fire and smoke, etc?
> 
> Not being able to set water on fire with basic fire sounds about right Smoke-based stuff can still be atop liquids since it's not in the water... But if some certain damaging smokes won't be able to interact with feet sound like a funny detail

A lot of flammable liquids float on top of water and continue to burn. Gasoline and napalm (effectively thickened gasoline) should both do this afaik. No particular reason to expect that welding fuel would either sink in water or extinguish for other reasons.

Aporhtonoma:
if something sinks that deep into lava like you know MOLTEN ROCKS KIND OF LAVA you aint getting out my dude

Lumipharon:
> if something sinks that deep into lava like you know MOLTEN ROCKS KIND OF LAVA you aint getting out my dude

You can take an armour piercing rocket to the skull and recover with a bandage and some aspirin. I uh, don't think we're doing damage realism here.

Lumipharon:
I never looked into it further because it didn't really bother me (a guy floating in water is more likely to have their legs submerged then randomly to have their side under the water).

If you want to do it though, you'll need to do something a bit different, as reset transform/keep apart work on appearances but apparently not on filters, for whatever reason.
You can have an overlay image that is itself transformed/filtered to avoid rotations of the parent, but you'd have to mess around to see how you could work that.

Lumipharon:
Neat, got a screenshot/clip of how it looks?

Lumipharon:
I think the lying down water alpha looks kinda funky, but thats honestly more an issue of SS13 human sprite perspectives I guess.

You can check my code for how I've done it, I don't get the move on and off thing you're describing.

PR #12887, opened 2023-05-03
Water slowdown refactor

Lumipharon:
## About The Pull Request
Some minor rework of how water slowdown is applied.

Slowdown is now define based via a proc.
All liquid turfs apply slowdown, and have a multiplier var to adjust slowdown.

This means you can no longer swim through lava faster than ankle deep water.

Lava and deep water have  a x1.5 modifier.

Killed the water slowdown xeno caste var.
Removed the weird negative slowdown boilers have (yes, they currently go through water FASTER than dry land), and make it just 0. I guess they float or whatever.

Make liquid turf effects apply to all mobs instead of just carbons.
## Why It's Good For The Game
Better code good.
More adjustable stuff good too.
## Changelog
:cl:
balance: Boilers no longer travel on water faster than land (just equal speed)
balance: Lava now applies slowdown similar to water
refactor: Tiny refactor of how water slowdown works
/:cl:

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -464,3 +464,7 @@ GLOBAL_LIST_INIT(organ_rel_size, list(
 /mob/proc/set_skills(datum/skills/new_skillset)
 	skills = new_skillset
 	SEND_SIGNAL(src, COMSIG_MOB_SKILLS_CHANGED, skills)
+
+///Returns the slowdown applied to the mob when moving through liquids like water
+/mob/proc/get_liquid_slowdown()
+	return MOB_WATER_SLOWDOWN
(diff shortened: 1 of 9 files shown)


Al-1ce:
Boiler going faster on water is such a specific niche I feel like you should keep it for the lulz sake.

code/modules/mob/mob_helpers.dm
```diff
@@ -464,3 +464,7 @@ GLOBAL_LIST_INIT(organ_rel_size, list(
 /mob/proc/set_skills(datum/skills/new_skillset)
 	skills = new_skillset
 	SEND_SIGNAL(src, COMSIG_MOB_SKILLS_CHANGED, skills)
+
+///Returns the slowdown applied to the mob when moving through liquids like water
```
QualityVan, line 468: Do these need to be a whole proc instead of a var on mob?

QualityVan (commented):
Agree that boiler swimming lore is critical to keep.

code/modules/mob/mob_helpers.dm
Lumipharon: If you can convince tivi that adding a mob var is fine, then sure.

Lumipharon:
Also to clarify, boiler gets no slowdown in water, so its still faster than anything else (along with warlock, which floats over the top)

TiviPlus (approved):

zzurge:
Ngl Boiler swimming was pretty funny,i think we should keep it for the culture,its not like boilers gonna apex predator LV river or something lol

PR #12891, opened 2023-05-03
Experimental Hugger changes

Sebbe9123:
## About The Pull Request
This PR aims to make changes to the Facehugger activation time when a carrier throws and directly hits a marine.

## Why It's Good For The Game
Currently playing carrier doesn't feel as fun, nor impactful as it used to, even after a long break.  Having talked to a fellow Carrier enjoyer we thought it would at least be interesting to see how carrier changes when rewarding risk more with direct hits and less safe gameplay.
Makes Carrier slightly less Utility focused as it is now and more viable in combat since of those 6/10 huggers you throw that don't get shot down by stray bullets instantly, 2 of them might directly hit and actually do something instead of sitting there, activating with BYOND lag an dying before they can act.
Doesn't change Larva hugger since it's by far the strongest hugger if it actually hits a marine, up for change though.

## Changelog

:cl:
balance: 'Rebalances' Huggers to be instant if directly hitting a marine
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
+++ b/code/modules/mob/living/carbon/xenomorph/facehuggers.dm
@@ -684,7 +684,7 @@
 	name = "neuro hugger"
 	desc = "This strange creature has a single prominent sharp proboscis."
 	color = COLOR_DARK_ORANGE
-	impact_time = 1 SECONDS
+	impact_time = 0 SECONDS
 	activate_time = 1.5 SECONDS
 	jump_cooldown = 1.5 SECONDS
 	proximity_time = 0.5 SECONDS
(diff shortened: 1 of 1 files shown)


KillerOrcaCora:
Not gonna lie, they had us at the first half. This seems really lazy though.

Sebbe9123:
> Not gonna lie, they had us at the first half. This seems really lazy though.

I'm open to better suggestions, Carrier is very good at utility right now, goal is just to make them better in combat, especially for newer players and laggy ass players, Doesn't need to be completely on theme with what carrier currently is either.

Lumipharon:
We're not turning huggers into impact grenades.

PR #12892, opened 2023-05-03
Adds a delay to throwing

ivanmixo:
## About The Pull Request

Same as attack delay.

Bonus deletion of unused define.
## Why It's Good For The Game

Should help with unfunny adhesive strats + added protection from griefers using external tools to throw something high damage at mach 5 to instakill people.
## Changelog
:cl:
balance: Added a delay to throwing
/:cl:


Davidththird:
salt nerf

Strayed-Raven:
> The Ofnir PR

and many more to come.

blackdav123:
tg had to do this a while ago too because people were using macros to lob 50 deathnettle at someone instantly

QualityVan:
Think it'd be worth having a var on item instead of a flat define? Same as how different items have different attack delays, so stuff ~~like throwing knives~~ don't become completely unusable. Scratch that bit, apparently they basically bypass throwcode anyway and would ignore this change. Still though.

zuneshine:
Endorsing.

zuneshine (approved):
LGTM

ivanmixo:
> Think it'd be worth having a var on item instead of a flat define? Same as how different items have different attack delays, so stuff ~like throwing knives~ don't become completely unusable. Scratch that bit, apparently they basically bypass throwcode anyway and would ignore this change. Still though.

No, I don't. I can't think of any item that'd make the game better if it was able to be thrown faster than every 0.8s, and most of the stuff that does like throwing knives completely bypass this anyhow.

Lumipharon:
The global input delay is pretty cringe tbh. Throwing random items has no balance implication for throwdelay, this is an issue that is specific to two items).

ivanmixo:
I'm not adding a var to `/obj/item` so it's either this or PR death and it goes unaddressed unless someone else wants to do it

bowlofcereal:
> I'm not adding a var to `/obj/item` so it's either this or PR death and it goes unaddressed unless someone else wants to do it.

"I'm not going to make it so only specific items have the throw delay which would alleviate concerns and instead keep going with my bad idea"

Input delays are bad in any regard when it comes to physically interacting with something in this way. Especially when its a global change to everything when you are only coping about two things here.

> Want to add tg has this exact same delay and it's perfectly fine.

We're not /tg/ lol

Losenis:
I can picture the nade FF skyrocketing as byond decides to eat your input and slap in a magical delay on throwing even when TD is low.

ivanmixo:
Its the exact same input lock time as clicking on any item around you so I think you're overestimating the impact quite a bit.

chizzy376:
LBTM, the game already feels clunky enough with input delays as is, stop making lag a feature lmao 😂😂

Losenis:
>I think you're overestimating the impact quite a bit

You never had your inputs eaten in the middle of a fight? It can very well happen with pretty much anything that needs you to click things fast like trying to reload a shotgun asap despite it not even having a delay. Add delay on top of that and start saying adios to your actions. I can already see this fucking up flares because there really isn't a reason to spam them when you're throwing them.

JackTheJackhammer:
as the top comment says, this probably will fuck up throw actions. delay due to lag is already quite bad. With this, the action would be very annoying

TiviPlus (changes requested):
Probably should at least make it 4

code/__DEFINES/conflict.dm
```diff
@@ -4,11 +4,11 @@
 #define CLICK_CD_RANGE 4
 #define CLICK_CD_CLICK_ABILITY 6
 #define CLICK_CD_MELEE 8
+#define CLICK_CD_THROWING 8
```
ivanmixo, line 7:
```suggestion
#define CLICK_CD_THROWING 4
```

Pariah919:
Universal delays are pretty cringe, even if its 0.4 now. Should be an item level var.

Mrrpip:
Mrm. Might as-well complain as it'll affect my playstyle.

But if I catch a hugger, throw it, grab another and fail to throw it because of the delay. That's no bueno, because now I've usually got a second hugger that I can't catch.

We should probably do it on an item per item basis as a few others had mentioned. Because that way we actually have to think about what it affects.

1kajowsmol:
plz dont add

TiviPlus (approved):

XANAv2:
This is a bad addition for me, as I have shitty internet, and my clicks tend to get ignored occasionally. I am afraid this will affect my gameplay.

zuneshine:
> This is a bad addition for me, as I have shitty internet, and my clicks tend to get ignored occasionally. I am afraid this will affect my gameplay.

Bear in mind that throw-delay is set IN-BETWEEN THROWS to prevent throwspamming. Additionally, your latency does nothing but delay inputs at their baseline depending on how long it takes to do a round trip.

BlueHNT:
>20 copers complaining about not being able to lob 20 grenades in 1 second.

XANAv2:
> 

Thanks, I didn't know that.

Naaanii:
With new acid wells this doesn't really seem needed as most cope was addressed.

QualityVan (approved):
I was reminded that you can throw things which aren't items (mobs, in particular) so an item var wouldn't be enough for a system fix, and I'm not super thirsty to put it on atoms.

AGasCanister:
Man if you go to any dictionary of any language and search for the word "Xope" the definition is literally this PR link. If the problem is the grenade spam (a guy that only uses nades is completely useless after throwing all his nades btw) just give a throw delay to nades no need to fuck the entire marine playerbase with a hellish 10 hour delay!!

pilotmans:
i've been unable to throw grenades multiple times now due to this and have ended up biting the dust to it, if i had to guess the delay gets fuckied by time dilation

BlueHNT:
Man, I love not being pelted by 20 stickies the moment I poke my head out. God merge fr

Losenis:
>Someone decides to add adhesive nades that is Impact nades but worse because it doesn't throw the xeno around for you to throw more stickies at
>"Clearly, the act of throwing is the problem"

AGasCanister:
"tgstation deleted a comment from AGasCanister"
lol the cope, is as easy as limiting throw delay to only nades but I guess we gotta remove the comment that mentions it!!

bowlofcereal:
it really is a github moment to ignore all people saying its causing problems and a bad idea in general, delete comments suggesting alternatives, and then have it tm'd anyways and it as expected causing bugs.

i wont understand having an issue with one thing and then completely murdering the entire system its built on rather then trying to alleviate the main problem itself, this doesnt stop nadespam at all

chizzy376:
Nicht gud, makes throwing items for the sake of organization a pain in the arse. Vend 2 ammo boxes from OP vendor, toss them into crate, do it too fast and everything feels so messy because half your throws are eaten. This makes req even more painful than it was before. Please make this PR only affect grenades.

pbfied:
With a ping of about 80ms, the game has an effective click delay of about 300-400 ms. You can test this for yourself very easily by setting up a macro for throwing knives or any other repetitive in-game action. 

Incidentally, grenades are far from being the most abusable mechanic allowed by the game's de facto click delays.

Lumipharon (approved):

Lumipharon:
I cast Monster Reborn, and special summon this nerf into play in attack mode.

chizzy376:
Please please please please please please please remove this dumb thing from Req entirely, give ro a skill to bypass this or just have this nerf apply to grenades I am going to go skizo mode having to wait an arbitrary amount of time when I'm throwing 30 things into a crate. I can't even wrench the vendors closer anymore because yall removed that too.

PR #12897, opened 2023-05-04
fixed acid spray dealing double damage to all objects

MacBlaze1:
## About The Pull Request

If an obj has already been hit by acid spray once in a tile check, it will not be hit again by the atom/moveable acid spray check below.

## Why It's Good For The Game

This bug has made praetorian acid spray 2x as effective on all defensive structures and objs that are also moveable, which is unintended. 

## Changelog

:cl:
fix: prae acid spray no longer deals double damage to all objects that are also atom/moveables
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/praetorian/abilities_praetorian.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/praetorian/abilities_praetorian.dm
@@ -85,9 +85,11 @@ GLOBAL_LIST_INIT(acid_spray_hit, typecacheof(list(/obj/structure/barricade, /obj
 	if(T.density)
 		return
 	var/is_blocked = FALSE
+	var/list/acided_objs = list()
 	for (var/obj/O in T)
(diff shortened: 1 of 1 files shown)


Losenis:
>carrier acid spray

did they chainsmoke lucky strikes and evolve or something

MacBlaze1:
> > carrier acid spray
> 
> did they chainsmoke lucky strikes and evolve or something

oops, typo

TiviPlus:
The fix with the list will be rather unperformant for what it is i dont like it

MacBlaze1:
uhhh this is intended apparently

MacBlaze1:
and we're back

Lumipharon:
To clarify here in the pr: This is old funny CM code. Doing extra damage to razor is intended, but should be correctly handled by just editing the acid act proc for it.
multitile vehicles don't exist so we don't care about them.
Cades are the only other than that can get hit twice, and this ONLY happens if you spray them sideways (as if you spray them front on, it blocks the acid and the second hit doesn't occur).

PR #12900, opened 2023-05-04
acid well changes and minor grenade code

Scarlet-Oduko:
## About The Pull Request

Acid wells now clear stickies from both marines and xenos. Minor changes to grenade code
## Why It's Good For The Game

Gives an active way for xenos to avoid death if prepared or near acid wells. Takes two charges to clear stickies.
## Changelog
:cl:
add: Acid wells remove sticky grenades
code: Changes to grenade code, specifically clear_refs

/:cl:

### DIFF
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -600,9 +600,15 @@ TUNNEL
 
 	var/datum/effect_system/smoke_spread/xeno/acid/acid_smoke
 
-	for(var/obj/item/explosive/grenade/sticky/sticky_bomb in stepper.contents)
-		sticky_bomb.clean_refs()
-		qdel(sticky_bomb)
+	if(charges >= 2)
+		for(var/obj/item/explosive/grenade/sticky/sticky_bomb in stepper.contents)
+			sticky_bomb.clean_refs()
+			qdel(sticky_bomb)
+			if(isxeno(stepper))
+				charges--
+				charges--
+				update_icon()
+				return
 	if(isxeno(stepper))
 		if(!(stepper.on_fire))
 			return
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -600,9 +600,6 @@ TUNNEL
 
 	var/datum/effect_system/smoke_spread/xeno/acid/acid_smoke
 
-	for(var/obj/item/explosive/grenade/sticky/trailblazer/fire_bomb in stepper.contents)
-		fire_bomb.clean_refs()
-		qdel(fire_bomb)
 	for(var/obj/item/explosive/grenade/sticky/sticky_bomb in stepper.contents)
 		sticky_bomb.clean_refs()
 		qdel(sticky_bomb)
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -600,6 +600,12 @@ TUNNEL
 
 	var/datum/effect_system/smoke_spread/xeno/acid/acid_smoke
 
+	for(var/obj/item/explosive/grenade/sticky/trailblazer/fire_bomb in stepper.contents)
+		fire_bomb.clean_refs()
+		qdel(fire_bomb)
+	for(var/obj/item/explosive/grenade/sticky/sticky_bomb in stepper.contents)
+		sticky_bomb.clean_refs()
+		qdel(sticky_bomb)
 	if(isxeno(stepper))
 		if(!(stepper.on_fire))
 			return
--- a/code/modules/xenomorph/xeno_structures.dm
+++ b/code/modules/xenomorph/xeno_structures.dm
@@ -605,10 +605,11 @@ TUNNEL
 			sticky_bomb.clean_refs()
 			qdel(sticky_bomb)
 			if(isxeno(stepper))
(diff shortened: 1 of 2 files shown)


code/modules/xenomorph/xeno_structures.dm
```diff
@@ -600,6 +600,12 @@ TUNNEL
 
 	var/datum/effect_system/smoke_spread/xeno/acid/acid_smoke
 
+	for(var/obj/item/explosive/grenade/sticky/trailblazer/fire_bomb in stepper.contents)
+		fire_bomb.clean_refs()
+		qdel(fire_bomb)
+	for(var/obj/item/explosive/grenade/sticky/sticky_bomb in stepper.contents)
+		sticky_bomb.clean_refs()
+		qdel(sticky_bomb)
```
TiviPlus, line 608:
```suggestion
	for(var/obj/item/explosive/grenade/sticky/sticky_bomb in stepper.contents)
		sticky_bomb.clean_refs()
		qdel(sticky_bomb)
```

TiviPlus (changes requested):

Pariah919:
This is pretty wacky considering acid wells can be instantly placed.

Scarlet-Oduko:
> This is pretty wacky considering acid wells can be instantly placed.
only by certain castes at least, but it provides some form of active counter to being stuck. which unlike regular grenades cannot be flung or ran from. any suggestions are always at least sometimes helpful.

MacBlaze1:
code improvement and feature my ass, label it a balance pr moron

Naaanii:
It should really drop the nade and not qdel it.

iLikeMemes:
I see other comments adressing the fact that acid wells can be instantly placed and this being combined with removing stickies being a bit unfair. I agree with that. I'd like to suggest that stickies are only removed if acid wells have at least 2 or 3 charges in them. In that case you do have a counter for the stickies but you'd have to have set it up beforehand. (or you could maybe do something like 1 charge only removes x amount of stickies)

AlexCrimson:
If they are just removed, that would blow up the acid well and anything around it, no? Seems annoying. Just make it cost 2 well charges and take a few seconds to place the well. Hivelords mostly just spam 1-charge wells right now, because there is little reason to have more. Making it cost 2 charges would be a big balance change.

Scarlet-Oduko:
two charges is a nice compromise between them falling off and one charge deleting them. I'll do some testing and then work on it.

Lumipharon:
why is stepping on acid insta deleting the nades?

Scarlet-Oduko:
> why is stepping on acid insta deleting the nades?

active way of getting rid of what is essentially a status effect like fire, takes two charges though so one charge acid wells can't be spammed to clear them. its like taking a big ole acid bath, part of an over arching set of PR's I'm putting together around stickies/fire as a whole.

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -600,6 +600,16 @@ TUNNEL
...
+	if(charges >= 2)
+		for(var/obj/item/explosive/grenade/sticky/sticky_bomb in stepper.contents)
+			sticky_bomb.clean_refs()
+			qdel(sticky_bomb)
+			if(isxeno(stepper))
+				charges-= 2
+				acid_smoke = new(get_turf(stepper))
+				acid_smoke.set_up(1, src)
+				acid_smoke.start()
+				update_icon()
```
TiviPlus, line 612: Dont get this isxeno check, why is it here?
```suggestion
			if(isxeno(stepper))
				charges -= 2
				acid_smoke = new(get_turf(stepper))
				acid_smoke.set_up(1, src)
				acid_smoke.start()
				update_icon()
```

Scarlet-Oduko: its so that when the acid well is triggered to clear stickies it leaves acid smoke like it does fire. so I copied how its done for fire as I didn't know a simpler way to do it.

Lumipharon:
> > why is stepping on acid insta deleting the nades?
> 
> active way of getting rid of what is essentially a status effect like fire, takes two charges though so one charge acid wells can't be spammed to clear them. its like taking a big ole acid bath, part of an over arching set of PR's I'm putting together around stickies/fire as a whole.

Why not just drop it on the floor though? Acid doesn't insta melt anything in any other situation.

Scarlet-Oduko:
> > > why is stepping on acid insta deleting the nades?
> > 
> > 
> > active way of getting rid of what is essentially a status effect like fire, takes two charges though so one charge acid wells can't be spammed to clear them. its like taking a big ole acid bath, part of an over arching set of PR's I'm putting together around stickies/fire as a whole.
> 
> Why not just drop it on the floor though? Acid doesn't insta melt anything in any other situation.

Following how the wells work, clearly fire completely, I made it two charges to clear stickies. As currently that is a lot of effort, and if it just dropped, I'd bump the charges down to one to make them so. they clear right now because I think that makes the most sense with how acid wells work and balance wise. rewarding you for preparing acid wells, it may be possible to get a 2 charge well made after you get stuck, but unless you're immediately making a well or have someone else immediately doing it. the current game environment tends to not value pre-prepared acid wells. it *is* different then clearing fire, but getting a clutch acid well off wouldn't be as near rewarding if all that did was make them drop and still just kill you regardless.

iLikeMemes:
youre removing charges and triggering smoke regardless of someone having a sticky or not

PR #12908, opened 2023-05-06
Platform rework

Lumipharon:
## About The Pull Request
Do you hate having to click drag to climb over platforms?
Do you hate that some platforms are entirely cosmetic but you can't tell which is which?

Reworks all platforms so that they apply slowdown when crossed, instead of requiring climbing.
Replaced all cosmetic platforms with real platforms because of this.

95% of the work in this pr was actually changing how platforms are laid out. They now are on the 'high' turf, while still being visually on the lower turf. This fixes some minor bugs such as coverage working in the wrong direction.

As they are no longer actual dense, they are now indestructable per some old TODO.

Added functionality so platforms can be made to only apply slowdown on entry or exit specifically (i.e. we could make it so there is no slowdown when jumping down a platform, and only slowdown when climbing up).

Also fixed a few minor map issues I noticed while going through and changing every platform in the game.

NOTE FOR MAPPERS:
Platforms will now auto generate outer corner pieces if you have the right directions set. This, and the actual slowdown mechanics are both controled by the platform direction. THIS IS NOT LIMITED TO 8 DIRECTIONS.

Any combination of direction flags can be manually set. i.e. you have have dir = 3, and it will be a platform facing both north and south. You can have 3, and in some cases all 4 directions on a single structure. SDMM does not visually show this, so you have to manually set the dir.

## Why It's Good For The Game
Less click fuckery
Smoother gameplay
bug fixes.

## Changelog

:cl:
add: Platforms now apply slowdown when crossed, but do not require climbing at all. All platforms are now functional instead of some being cosmetic
balance: rebalanced something
fix: fixed coverage for platforms working in the wrong direction

/:cl:


Gregormack:
Holy shit

code/game/objects/structures/platforms.dm
```diff
@@ -1,156 +1,175 @@
...
+		return
+	if(!isturf(old_loc))
+		return
+	if(get_dist(src, old_loc) != 1) //teleporter shenanigans
+		return
+
+	var/mob/arriving_mob = arrived
+	var/old_loc_dir = get_dir(src, old_loc)
+	if(old_loc_dir & dir)
+		arriving_mob.client.move_delay += climb_slowdown //applied directly to client movedelay as mob delay would only apply after an additional movement
```
QualityVan, line 112: Doesn't this cause slowdown when moving from one platform to another? Same with exit.

code/game/objects/structures/platforms.dm
```diff
@@ -1,156 +1,175 @@
...
-/obj/structure/platform/proc/on_try_exit(datum/source, atom/movable/O, direction, list/knownblockers)
+	var/static/list/entry_only = list(COMSIG_ATOM_ENTERED = PROC_REF(on_enter))
+
+	var/static/list/exit_only = list(COMSIG_ATOM_EXITED = PROC_REF(on_exit))
+
+
+	AddElement(/datum/element/connect_loc, (platform_flags & ENTRY_SLOWDOWN) ? (platform_flags & EXIT_SLOWDOWN) ? entry_and_exit : entry_only : exit_only)
+
+///clears and sets overlays based on current dir
+/obj/structure/platform/proc/apply_overlays()
```
QualityVan, line 44: Why not handle this through update_overlays?

code/game/objects/structures/platforms.dm
```diff
@@ -1,156 +1,175 @@
...
-	name = "rock cliff"
-	desc = "A collection of stones and rocks that form a steep cliff, it looks climbable."
+		if(NORTH)
+			layer = ABOVE_MOB_PLATFORM_LAYER
+		if(SOUTH)
+			layer = ABOVE_MOB_PLATFORM_LAYER
+		if(SOUTHEAST)
+			layer = ABOVE_MOB_PLATFORM_LAYER
+		if(SOUTHWEST)
+			layer = ABOVE_MOB_PLATFORM_LAYER
```
QualityVan, line 164: Shouldn't north draw under the mob?

Lumipharon: They are the two north corner bits, so the draw above the mob like the north facing platform they'd link to.

code/game/objects/structures/platforms.dm
Lumipharon: ? If two platforms were facing each other I guess. If they're facing the same way they're not going to proc.

code/game/objects/structures/platforms.dm
Lumipharon: true

code/game/objects/structures/platforms.dm
```diff
@@ -1,156 +1,174 @@
...
-			I.pixel_x = 16
-		if(WEST)
-			I.pixel_x = -16
-	overlays += I
-	var/static/list/connections = list(
-		COMSIG_ATOM_EXIT = PROC_REF(on_try_exit)
-	)
-	AddElement(/datum/element/connect_loc, connections)
+	update_icon()
+	icon_state = ""
```
TiviPlus, line 26: ???

code/game/objects/structures/platforms.dm
```diff
@@ -1,156 +1,174 @@
...
+		return
+	if(!isturf(old_loc))
+		return
+	if(get_dist(src, old_loc) != 1) //teleporter shenanigans
+		return
+
+	var/mob/arriving_mob = arrived
+	var/old_loc_dir = get_dir(src, old_loc)
+	if(old_loc_dir & dir)
+		arriving_mob.client.move_delay += climb_slowdown //applied directly to client movedelay as mob delay would only apply after an additional movement
```
TiviPlus, line 111: I dont get why you backload the slowdown instead of a quick doafter while crossing which is logically more consistant

TiviPlus (changes requested):

code/game/objects/structures/platforms.dm
Lumipharon: Remember the autoclimb PR? Its a good concept but in practice it was really annoying.

code/game/objects/structures/platforms.dm
Lumipharon: made null.

code/game/objects/structures/platforms.dm
QualityVan: I was thinking about platforms as full tile things rather than edges.

PR #12913, opened 2023-05-06
hotfix for incorrect var name

Lumipharon:
## About The Pull Request
Quick hotfix. Two prs were merged that both touched the same file causing a conflict with var names getting changed.
## Why It's Good For The Game
Compiling is pretty useful.


Lumipharon:
self merging because master not compiling is probably bad.

PR #12920, opened 2023-05-07
Makes the weather sound preference actually work

ivanmixo:
## About The Pull Request

Combines the ash/sandstorm looping sounds into two (active/weak). This means they won't exactly always be appropriate, but not like there's much difference between the outside/inside variations at all.

Accomplishes this by making the sounds play directly to mobs instead of to areas based on preferences.
## Why It's Good For The Game

The weather sound preference actually working is nice. Not everyone wants the weather sfx to make it hard to hear anything.
## Changelog
:cl:
fix: Made the weather sound preference actually work
/:cl:


blackdav123:
holy shit pr of the year

PR #12921, opened 2023-05-07
Increases firerate of laser rifle

Losenis:
## About The Pull Request
Laser rifle firerate change from to:
Standard, 0.2 > 0.15
Overcharged, 0.45 > 0.35

## Why It's Good For The Game
Comparing it to the AR12 as a sidegrade choice, the laser rifle sacrifices aim mode, attachments, and space efficiency for a hitscan projectile that as it turns out has less dps vs <85 armor mobs and less damage per mag to boot. Extended/Barrel charger bullet speed mods made hitscan have less of a difference, and sunder rework also made the sunder change less impactful given xenos get less sunder the more they have. Ballistics also have the option of grabbing ammo packets for more efficient ammo storage (150 rounds for the space of a mag in the backpack, or the gigantic ammo pack for like 2400 rounds), which lasguns do not have other than an engilocked item that isn't even available in req or the big dense recharger made obsolete by just bringing more ammo boxes much like a normal gun. A faster firerate will allow the lasrifle standard shot to have mildly more DPS while still sacrificing attachments and aim mode than the AR12

The closest equivalent to overcharge mode in relation to standard shot is BR64's additional stats at the cost of firerate to the AR12. BR64 in relation to AR12 gets 50% more fire delay (0.2 to 0.3), 30% more damage, a whooping 200% more pen, and 150% more sunder for ~70% of the magcount. Overcharge in relation to standard has _over 100%_ the fire delay (0.2 vs 0.45), 100% more damage, 100% more pen, and 100% more sunder for 40% of the magcount. Fair stats on their own, though reduced to 0.35 to keep it equivalent to standard. End result is overcharge mode will still be proportional to the stats of standard shot while keeping its drawbacks and mag inefficiency.

Vanilla stats:
image

Proposed stats:
image

## Changelog

:cl:
balance: Lasrifle firerate: Standard shot, 0.2 > 0.15. Overcharge: 0.45 > 0.35.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -427,7 +427,7 @@
 	wield_delay = 0.5 SECONDS
 	scatter = 0
 	scatter_unwielded = 10
-	fire_delay = 0.2 SECONDS
+	fire_delay = 0.15 SECONDS
 	accuracy_mult_unwielded = 0.55
 	damage_falloff_mult = 0.2
 	mode_list = list(
@@ -444,7 +444,7 @@
 /datum/lasrifle/base/energy_rifle_mode/standard
 	rounds_per_shot = 12
 	ammo_datum_type = /datum/ammo/energy/lasgun/marine
-	fire_delay = 0.2 SECONDS
+	fire_delay = 0.15 SECONDS
 	fire_sound = 'sound/weapons/guns/fire/Laser Rifle Standard.ogg'
 	message_to_user = "You set the laser rifle's charge mode to standard fire."
 	fire_mode = GUN_FIREMODE_AUTOMATIC
@@ -454,7 +454,7 @@
 /datum/lasrifle/base/energy_rifle_mode/overcharge
 	rounds_per_shot = 30
 	ammo_datum_type = /datum/ammo/energy/lasgun/marine/overcharge
-	fire_delay = 0.45 SECONDS
+	fire_delay = 0.35 SECONDS
 	fire_sound = 'sound/weapons/guns/fire/Laser overcharge standard.ogg'
 	message_to_user = "You set the laser rifle's charge mode to overcharge."
 	fire_mode = GUN_FIREMODE_AUTOMATIC


ClosetedSkeletonGH:
I main overcharge and I can say without a doubt it doesn’t need a fire rate increase. 

Overcharge is THE best laser weapon mode, and THE best sniper in the basic loadout (50% because it’s good 50% every other laser weapon underperforms)

it’s already on the upper end of very good, any buff would make it pretty busted

The normal rifle hit rate being buffed I support

Losenis:
Fair. I'll wait on a maint to see if it's needed and I'll change it.

Lumipharon:
I see issues with your logic here.

Firstly lasers have very low falloff, so the dps comparison isn't accurate.

Saying sunder is bad because it scales also doesn't make sense. It just means it's hard to get extreme amounts of sunder, and that effects Mr t12 just as much as a laser rifle.

Saying eb/bc negates hitscan ignores a: that they can still miss, while hitscan is 100% guaranteed to hit if you click in the right spot. And b: taking those mods means no bayonet.

Losenis:
>Firstly lasers have very low falloff

If you're using it as a sniper for the low falloff to have a noticeable impact you're either using the no aim mode gun to its strengths in having no marines in front, which is a reward in itself, or you're risking FFing other people due to using a sniper with no falloff. It has to have good strengths to make up for its weaknesses in other areas, otherwise it's not a sidegrade, it's a flat upgrade or a flat downgrade. There's a reason people go for the T127 which has aim mode which penalty can even be lessened by bipod rather than using the laser sniper. We can call up stats but I'm sure even if they were increased to match it'd not really be as popular, much less because 127 can fit a barrel charger to be almost-hitscan anyway. Again, it doesn't negate, but it does make the differences less pronounced.

I also didn't say sunder is bad. Never did, actually. I meant that it's less pronounced because of the sunder rework making it slower to accumulate sunder. Back then there'd have been an argument, and I have used the old old lasgun to delete crushers with its atrociously big sunder that made this one pale in comparison, but even if old lasgun was brought back it'd not really be as bonkers as back then. You also have way less DPS so even if you're sundering xenos, you're not going to be doing as much damage until you have landed a good number of hits to compensate. I would run the math but fuck if I know what formula to use anymore after that pr, all the spreadsheets I know of are very outdated.

I also didn't say neither EB nor EC negated hitscan. I said it makes the difference less pronounced, not nonexistent. It doesn't let you put a bayonet on the gun, of course, but people can bring a bayonet on their webbing to put it there. That or use the knife in their boot that will also 2hit weeds as would a bayoneted gun, or clear walls just as good since even though it has less damage it hits walls way faster for whatever reason. Not only that, you're also able to use EB/BC on a gun that has aim mode, which you can also double up with an RDS or bipod to negate the penalty completely for something like a T60. We can even return to the T12 in the shape of the T42, I carried a bipod and rds on webbing while using a gyro and miniscope and just swapped to those to simply sit tight and clear maze at a distance by simply shooting at it with marines in front while not having ammo issues since I had a 2400 round backpack box to refill from as the mags conveniently dropped to the box next to it. Those guns have the versatility of switching attachments on the fly, lasrifle has few if any let alone those relevant to aim mode that it lacks.

Losenis:
DPS at 9 tiles which is pretty much screen edge where a xeno can just walk a tile and be completely out of sight assuming no maze even earlier, and assuming flare city since even baldur can't let you see that far. Still only breaks even at 65 armor. Overcharge has it beat since armor at 20, in fact.

image

Taking the extreme case of 20 tiles of range since that's where T12 bullets 100% disappear, some even earlier, lasrfile starts beating T12 even at 0 armor by... 5 points of dps. I applaud the guy who'll pick a T12 rather than a 127 to use as sniper but power to him. Overcharge is actually equal to standard funnily enough because the same falloff doesn't hit it as much, and gets more DPS the more armor the xeno has. Even runners have more than 0 armor so, settling on OC beating standard. Would pick 25 tiles since that's the max a T47 scope can see, or even 25 if you can spot the glow of a boiler a tile behind it, but T12 bullets don't reach that far and it's only going to make standard shot fall behind even more. Again, all assuming flare city where everything that far is illuminated and pure open ground, which hardly ever happens.

image

If Overcharge's stats are fine, I got no problem with that. Again, I just buffed it to keep it proportional to standard, so I can revert it if it's a problem.

ClosetedSkeletonGH:
While I do like looking at numbers, I’m a big proponent of feel > numbers

I think this is a good change personally for the following reasons:

1) It never feels like there’s a reason to switch out of overcharge even when a rouny or other speedy critter is attacking you because overcharge is very good and standard mode barely holds a candle to it

2) switching between firing modes is a cool gimmick of the laser guns

3) I think switching modes in combat to changes niches is fun and cool but it’s hard to justify it with TE-R standard being the way it is

——
I think videos of the firing speed would be the best tho, like versus an ancient rouny versus an ancient queen. That way feel can be displayed some

Losenis:
> —— I think videos of the firing speed would be the best tho, like versus an ancient rouny versus an ancient queen. That way feel can be displayed some

https://files.catbox.moe/xygjhu.mp4

BlueHNT:
I do feel like Lasguns should be stronger dps contenders than normal rifles, for the simple reason of the fact they ain't got aim mode. This pretty much eliminates the majority of their usage by default and if they are gonna be niche similar to revolvers and such need to be equally as good.

Sunder is a strong weapon against Xenos, but one hiver completely gets rid of that benefit and unless you direct them with a rocket is not gonna do shit.

Losenis:
So I just found out that barrel charger can reduce damage falloff by 0.2, so assuming the attachment there reducing both AR12 and BR64 to 0.3 damage falloff and using the extreme example of 20 tiles afar pretending it's all lit and there's no maze or even basic map terrain cover... Yeah they still get beat at anything under 70 armor. Sheesh, lasrifle can't take a break.

image

Lumipharon:
It really doesn't need this. You're just comparing raw dps.
Overcharge is already strong enough as it is. Normal mode isn't particularly great, but buffing the dps by 33% is cringe and makes it just straight up superior to most projectile weapons.

Losenis:
>You're just comparing raw dps.

No I'm not? I brought up the ammo economy, I brought up the space aspect, I brought up the lack of attachments and aim mode, I brought up the lack of packets, I brought up the past change to sunder that made it less impactful across the board, I brought up the barrel charger bullet speed change that lessened the impact of hitscan and then also brought up the barrel charger falloff damage that made this "low falloff" benefit amount to a 0.1 damage difference per tile which wouldn't even amount to 1 point of damage when shooting at anything visible withot a scope. And even in discord the saving grace of hand recharger got brought up that would probably be removed from the game already.

>and makes it just straight up superior to most projectile weapons
It doesn't do anything the others already can't given even the video I put up up there against the absolute best case scenario of a primo crusher with tower ward due to the higher pen benefit of lasguns. BR64 can even crit a crusher with a full mag when OC lasrifle can't.

Losenis:
Did the AR11, even with the proposed buff this idea that it'd make laser guns, and I quote, "straight up superior to most projectile weapons" is either bullshit or actual dishonest arguing.

image
image

PR #12922, opened 2023-05-07
Gets rid of gyroscopic stabilizer's damage debuff

Losenis:
## About The Pull Request
So time ago someone decided to give gyroscopic stabilizer a 10% damage debuff per marine the bullet went over in aim mode. Problem is it's fucking impossible to tell unless you explicitly get told about it by a codediver or were around for the PR. Gyroscopic stabilizer's aim mode speed buff already competes with vgrip's speed buff that even applies outside of aim mode so it's not a braindead choice to need that debuff anyway.

## Why It's Good For The Game
Hard to convey mechanics bad

## Changelog

:cl:
balance: Gyroscopic stabilizer no longer has a 10% damage penalty per marine the bullet on aim mode goes through.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -1149,7 +1149,6 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	scatter_unwielded_mod = -2
 	recoil_unwielded_mod = -1
 	aim_mode_movement_mult = -0.5
-	shot_marine_damage_falloff = -0.1
 
 /obj/item/attachable/lasersight
 	name = "laser sight"


Joe13413:
for what reason was this even added i dont get it

Naaanii:
Why not just add it to the codex then?

ClosetedSkeletonGH:
Because a 10% damage debuff for a gyro stabilizer but ONLY while in sim mode is weird and an extremely specific mechanic. 

Aimmode shouldn’t change damage, just DPS imo

Lumipharon:
> for what reason was this even added i dont get it

The pr added 2 things to the gyro together.
A: Lower aim mode movement penalty
B: aim mode damage penalty thingo

The two were added together as a tradeoff rather than just a straight upgrade.

Gyro in general is just a reall good upgrade. It benefits single hand and wielded stats, as well as aim mode, and has no other downsides compared to say, vgrip, that punishes one hand stats/wield time in exchange for various wielded benefits.

That said, this is pretty weird and yeah not properly explained. I thought this stat was actually nonfunctional lol.

PR #12938, opened 2023-05-10
Fixes robot crit stagger

DeltaFire15:
## About The Pull Request
Title.
Currently, due for some reason happening in updatehealth() (???), robots will only get staggerslowed anytime their health is updated, means, damage, repairs, a bunch of other random procs called by a variety of things.
This is very inconsistant and very much not a stable "you are staggerslowed below -25 health", instead varying between "no effect" and "staggerstacked into next year"

This PR improves that behavior by moving the staggerslow adjustment into Life(), which is called consistantly, aswell as establishes a cap on stagger stack buildup (slow overlaps instead of stacking) to avoid being permanently staggered after going out of crit.
## Why It's Good For The Game
Fix man good?
## Changelog
:cl:
fix: Combat robot crit staggerslow now behaves consistently.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -466,12 +466,11 @@
 		clear_fullscreen("robothalf")
 		clear_fullscreen("robotlow")
 
-/mob/living/carbon/human/species/robot/Life()
-	. = ..()
-	if(health > -25)
+/datum/species/robot/handle_unique_behavior(mob/living/carbon/human/H)
+	if(H.health > -25) //Staggerslowed if below crit threshold.
 		return
-	adjust_stagger(2, capped = 10)
-	adjust_slowdown(1)
+	H.adjust_stagger(2, capped = 10)
+	H.adjust_slowdown(1)
 
 ///Lets a robot repair itself over time at the cost of being stunned and blind
 /datum/action/repair_self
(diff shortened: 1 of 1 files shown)


Lumipharon (approved):
Whoops, not sure why I didn't put that there to begin with.

Just a note with the stagger, you naturally regen 1 stagger every life tick, so you shouldn't be able to stack it infinitely anyway. This method is fine though.

DeltaFire15:
> Whoops, not sure why I didn't put that there to begin with.
> 
> Just a note with the stagger, you naturally regen 1 stagger every life tick, so you shouldn't be able to stack it infinitely anyway. This method is fine though.

Yes, the 1 natural stagger regen is why this adds 2 in life, as to be 1 effective stagger increase. If not capped, it would potentially increase as long as the robot is in crit, which may lead to a very long time of being staggered post crit-exit. Not infinitely of course, just.. a long while (potentially)

Lumipharon:
> Yes, the 1 natural stagger regen is why this adds 2 in life, as to be 1 effective stagger increase. If not capped, it would potentially increase as long as the robot is in crit, which may lead to a very long time of being staggered post crit-exit. Not infinitely of course, just.. a long while (potentially)

You don't need 2 stagger though. It happens in order.
Stagger goes down by 1, to a minimum of 0.
Stagger is then increased by 1.

So even with just 1 stagger, you'll never break out of stagger without healing, but it won't creep up either. Again, either way is fine though.

QualityVan (changes requested):
This should be on the species datum instead, handle_unique_behavior should work fine.

PR #12945, opened 2023-05-11
Boiler Smoke is Transparent

tyeagg:
## About The Pull Request

https://github.com/tgstation/TerraGov-Marine-Corps/assets/98577688/cc92a818-f0c9-479e-bb10-3e06bf3f53ac

ignore the irobot clip playing
## Why It's Good For The Game
ever since smokening pr made smokes actually effective its been hard to counter a good boiler main reason is you cant see it to even counter it via long range weapons
## Changelog
:cl:
balance: boiler smoke is now transparent
/:cl:

PR #12951, opened 2023-05-11
Tfoot smoke eats xeno smoke

Lumipharon:
## About The Pull Request
All forms of Tanglefoot smoke now reduces the lifetime of any xeno type smoke found on the same tile.

This means xeno smoke can still spread through tangle, but will rapidly get removed.

Also technically applies to WP smoke since that has the same anti plasma properties.
## Why It's Good For The Game
Adds a limited counter measure to smoke spam.
## Changelog
:cl:
balance: Tanglefoot gas now quickly removes xeno smoke occupying the same space
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/smoke.dm
+++ b/code/game/objects/effects/effect_system/smoke.dm
@@ -296,6 +296,13 @@
 	expansion_speed = 3
 	smoke_traits = SMOKE_XENO
 
+/obj/effect/particle_effect/smoke/xeno/effect_smoke(obj/effect/particle_effect/smoke/S)
+	. = ..()
+	if(!.)
+		return
+	if(S.smoke_traits & SMOKE_PLASMALOSS)
+		lifetime -= 2
+
 //Xeno acid smoke.
 /obj/effect/particle_effect/smoke/xeno/burn
 	lifetime = 6


BlueHNT:
Purge the xeno smoke on the tile where the tangle is also present, but make it both ways. So tangle will "eat" the xeno smoke at the cost of itself. Otherwise, marines will forever tangle and get completely free pushes.

shelteredfreak:
I would prefer https://github.com/tgstation/TerraGov-Marine-Corps/pull/12945 over this to be honest.

I'm very biased, so yes I do hate it because of loving Defiler.

But I would more say that you should know that if this gets TM'd or merged fully, the first reaction for marines is gonna be starting to mass-buy TFoot nades to prevent smoke. And I don't think that's very enjoyable as xenos. Having a direct counter to what makes a xeno good in the first place can very quickly spoil the experience. I love the Smokening PR, but this is VERY harsh. 

I think Defiler being able to deploy gas inside marine smoke is entirely fair, because it has to do it in person. A Boiler can fire from offscreen without being stopped (not a lot of instant kills to stop a smart boiler from fleeing after taking hits).

It's a bit annoying that it circles back around to marine smokes countering xeno smokes, just now with only one type, which means marines are gonna start using the crap out of that one type.

Naaanii:
Would be nicer if it just prevented the spread of boiler or defiler gas so foresight is rewarded instead of just spamming it as a harder counter vs Xeno cover. Still better than the alternative ig.

isotope436:
this is better 100% i need more reasons to abuse tangle foot

Wirpy:
Will the tanglefoot stay after removing the xeno smoke? 
I could foresee that causing issues by creating a new meta where marines just unga their way through everything by just everyone spamming tanglefoots which would make for a unfun experience overall due to essentially being handed free pushes without any drawback. 

I'd like to believe It would be better as other people have mentioned that both smokes essentially cancel eachother out (they both remove eachother) 
This way marines can counter xeno smoke and xenos can counter tanglefoot creating a much more interesting dynamic between the two sides and could lead to more stragety behind fighting and what not. 

And on the other hand this could lead to a new boring meta where marines just spam tanglefoot to no effect because xenos spam defiler/boiler for no effect effectively rendering idea. In my own belief I think It would probably equal less fun for xenos It could lead to two castes essentially being pretty boring to play / useless. 

Anyways overall I believe with this PR and the idea behind it is a way more suitable solution than the boiler smoke pr If implemented into the game in a ideal manner. Hard to tell what will happen really but It'll sure be interesting.

GreatKingWombat:
None of these apocalyptic scenarios happened before the smokening, so there shouldn't be a problem now.

QualityVan:
You'd prefer to put this on tfoot/phos? Part of the complaints previously were that smokening killed any particular use for tac smoke, while plasma drain already sends plasma to zero over ten seconds.

Lumipharon:
> You'd prefer to put this on tfoot/phos? Part of the complaints previously were that smokening killed any particular use for tac smoke, while plasma drain already sends plasma to zero over ten seconds.

I chose tfoot because it's much more limited than tac smoke, so you can't just spam your way to victory

Aporhtonoma:
good proper way to deal with smokes.I agree with naaanii that it would be better if it blocked the spread all together but honestly eating it is good too.

QualityVan (approved):
Not convinced this is either necessary or the correct solution to the apparent issue, but no complaints codewise.

RikuTheKiller:
I agree with making them eat each other instead of tfoot coming on top. I don't want more reasons for marines to tfoot spam, they already do it enough.

isotope436:
after having used literally only tsmoke for an entire op i will state that this works really well and is a very good step forward in countering the gas meta well i do hate that its not an instant deletion of the gas unless you specificly build to counter it theres no chance of this becoming a problem of spam and even then at the current winrate marines could use a way to counter boiler and defiler spam

maikilangiolo:
> Not convinced this is either necessary or the correct solution
Well if the gas stacking PR didn't get merged, this wouldn't have been necessary.

PR #12960, opened 2023-05-12
New sunder no longer scales on existing sunder

Losenis:
## About The Pull Request
image

## Why It's Good For The Game
Sunder scaling on applied sunder excessively hit weapons that have high firerate and low sunder. As an example:

A projectile that does 100 sunder will bring a xeno's armor% down from 100% to 0%.
A projectile that does 50 sunder and is shot twice will first bring down the armor% from 100% to 50%, and then for the second shot have the sunder value multiplied by 0.5 to do 25 and leave the xeno with 25% armor remaining rather than 0.
This pattern keeps going for guns that have lower sunder per shot.

## Changelog

:cl:
balance: Projectile sunder is no longer multiplied by the % of remaining armor of the struck xeno

/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
+++ b/code/modules/mob/living/carbon/xenomorph/damage_procs.dm
@@ -40,7 +40,7 @@
 	adjust_stagger(max(0, ex_slowdown - 2)) //Stagger 2 less than slowdown
 
 	//Sunder
-	adjust_sunder(max(0, 50 * (3 - severity) * bomb_sunder_multiplier * get_sunder()))
+	adjust_sunder(max(0, 50 * (3 - severity) * bomb_sunder_multiplier))
 
 	//Damage
 	var/ex_damage = 40 + rand(0, 20) + 50*(4 - severity)  //changed so overall damage stays similar
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -915,7 +915,7 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 			feedback_flags |= (BULLET_FEEDBACK_FIRE)
 
 	if(proj.ammo.flags_ammo_behavior & AMMO_SUNDERING)
-		adjust_sunder(proj.sundering * get_sunder())
+		adjust_sunder(proj.sundering)
 
 	if(damage)
 		var/shrapnel_roll = do_shrapnel_roll(proj, damage)


TiviPlus (approved):

Lumipharon:
Just stating it here, that scaling sunder is good because it means it's easy to degrade xeno armour (making them weaker and rewarding the unga) but difficult to utterly break it and make the xeno completely ineffective for prolonged periods of time.
 Having to rest for long periods of time because your health is full but sunder is shot, and you don't have a hive around isn't good gameplay.

Scaling sunder means there is an equilibrium between sunder application and sunder recovery, where non scaling sunder and the ability for marines to far outstrip sunder healing means there is an inevitable creep towards full sunder, and leads to the endless sunder walance stuff as both sides cope about the other side having it better.

Barnet2:
Given that 
1) Backlining is already viable with near no downsides an
2) If you have the hive around sunder is non existent due to basically every heal healing it
I absolutely love this change. Sunder might finally matter again outside of 100 - 0 sunder.

ClosetedSkeletonGH:
The issues with the change is that we were told more changes would follow that would make it feel better to go against, but instead we have a sunder that is just entirely negated by being near a xeno with a healing ability.

Don’t get me wrong, I like that these castes can heal sunder, it makes them feel even more useful, but the issue is sunder is never applied in high enough numbers to make it where it’s not just 15 seconds of drone spamming their ability and then the xeno is perfectly fine after bullet sponging

Naaanii:
> The issues with the change is that we were told more changes would follow that would make it feel better to go against, but instead we have a sunder that is just entirely negated by being near a xeno with a healing ability.
> 
> Don’t get me wrong, I like that these castes can heal sunder, it makes them feel even more useful, but the issue is sunder is never applied in high enough numbers to make it where it’s not just 15 seconds of drone spamming their ability and then the xeno is perfectly fine after bullet sponging

Railgun go brrrt.

maikilangiolo:
> scaling sunder is good

It isn't, because as explained, all sunder guns that weren't gigantic sunder sticks (i.e. 100 sunder or close to it) got absolutely shafted. You say that non-scaling sunder causes marines to apply more sunder than it was healed. Well, scaling sunder caused xenos to heal more sunder than it was applied, particularly because the same PR buffed sunder heal as well as nerfed sunder. Scaling sunder absolutely does not mean there's an equilibrium, because as another commenter said, promised further PRs to tune this better were never made.

Lumipharon:
> > scaling sunder is good
> 
> It isn't, because as explained, all sunder guns that weren't gigantic sunder sticks (i.e. 100 sunder or close to it) got absolutely shafted. You say that non-scaling sunder causes marines to apply more sunder than it was healed. Well, scaling sunder caused xenos to heal more sunder than it was applied, particularly because the same PR buffed sunder heal as well as nerfed sunder. Scaling sunder absolutely does not mean there's an equilibrium, because as another commenter said, promised further PRs to tune this better were never made.

Sunder healing was nerfed after scaling sunder was added.

Also when people talk about scaling sunder shafting sunder application, I don't think they realise how little difference it actually makes in most circumstances, and only becomes quite noticable when sunder is already high.

For example, if you empty 50 t12 shots into a full armour xeno, the difference in sunder between scaled and linear sunder is 2.8 sunder.

Losenis:
>that scaling sunder is good because it means it's easy to degrade xeno armour (making them weaker and rewarding the unga) but difficult to utterly break it

Now that's just a flat out lie because nowhere was it ever changed afterwards to make sunder "easier" to apply while everyting but railgun was nerfed across the board, and i doubt the sunder healing was ever nerfed to pre-rework levels.

>For example, if you empty 50 t12 shots into a full armour xeno, the difference in sunder between scaled and linear sunder is 2.8 sunder.

Find me a non-ssd xeno in the wild that'll stick around completely still to get shot up by a full t12 mag from a single guy while offweeds. In actual practice a xeno will get shot up by 3 guys in a second, go back to the maze, heal up the damage, return, and repeat a couple hundred times until he gets killed off by something that doesn't even take into account sunder like an AMR or br8 or zune going br64 etc. This low difference repeated that often was the difference between a xeno resting and only minding health due to the sunder healing up or a xeno healing up to full and having trace amounts of sunder that'll start building up throughout engagements forcing him to rest longer if he wanted to actually keep his armor topped off too. And it's most noticeable on high armor xenos like crushers and the sort because low armor rounies will sooner get killed off before they even get 95% sunder as if a 0.95 multiplier on 19 armor actually mattered other than the freak case scenario of a buckshot pb.

You know, like I observed since day 1 of that pr.

BlueHNT:
This is a good change, but as a xeno, I mald for we already suffer the insane artillery and AMR lately.

Pariah919:
I feel like you could just make this use a bullet level multiplier to make it more reasonable than outright removal.

PR #12964, opened 2023-05-12
TGUI-ify the strip menu

ivanmixo:
## About The Pull Request

Makes the strip menu tgui. Revises the inventory equip flags somewhat, but that'll need a whole other PR, which I pinky promise I'll do after this is in. Also fixes cuffs being bad.

Partial port of https://github.com/BeeStation/BeeStation-Hornet/pull/5705/ with stuff added/removed.

2kcaXtPLKp

dreamseeker_Kyj2w3QUIU

TM this, seriously. It has potential to break shit.

## Why It's Good For The Game

Modern menus are neat.
## Changelog
:cl:
add: Made the strip menu into a TGUI iteration
add: You can also untighten/tighten shield straps from the menu now!
fix: Handcuff overlays should no longer be buggy
/:cl:


Losenis:
>Loosen strap
Finally, more more chopping the mf's arm off if I wanted to shove him in cryo

code/modules/mob/living/simple_animal/friendly/parrot.dm
```diff
@@ -610,3 +530,86 @@
...
+/datum/strippable_item/parrot_headset/get_item(atom/source)
+	var/mob/living/simple_animal/parrot/parrot_source = source
+	return istype(parrot_source) ? parrot_source.ears : null
+
+/datum/strippable_item/parrot_headset/try_equip(atom/source, obj/item/equipping, mob/user)
+	. = ..()
+	if(!.)
+		return FALSE
+	if(!istype(equipping, /obj/item/radio/headset))
+		to_chat(user, "<span class='warning'>[equipping] won't fit!</span>")
```
TiviPlus, line 546: Missing span class/balloon

code/modules/mob/living/simple_animal/friendly/parrot.dm
```diff
@@ -610,3 +530,86 @@
...
+	var/mob/living/simple_animal/parrot/parrot_source = source
+	if(!istype(parrot_source))
+		return
+
+	if(!user.transferItemToLoc(radio, source))
+		return
+
+	parrot_source.ears = radio
+
+	to_chat(user, "<span class='notice'>You fit [radio] onto [source].</span>")
```
TiviPlus, line 569: ditto

code/modules/mob/living/simple_animal/friendly/parrot.dm
ivanmixo: Mm guh should really implement a grep for this

chizzy376:
Why not just use the icons we get for automated vendor instead of a list? You can check how skyrat does it. Also currently this PR breaks several things. 
>Xenos can strip, I don't need to elaborate why this is bad


>Pressing E with an item in your right hand puts it in your left hand instead of actually being put where it's usually put


>Putting cells into a handheld cell charger makes them disappear

Lumipharon:
@ivanmixo you apparently can't set the back slot for quick equip anymore.

Pixelknightly:
I was able to take off a fellow marine's ID with this. You're supposed to take a copy of the tag. Not the actual ID itself.

ivanmixo:
I hate how snowflake that was, if someone wants to add it back after this is added they're free to

Losenis:
Can't seem to fit guns into suit slots, IE having a recently defibbed person and putting his armor back on and then his gun, but the gun won't fit

chizzy376:
Fix this. 
image

chizzy376:
> I hate how snowflake that was, if someone wants to add it back after this is added they're free to

Ah yes, removing functionality because you hate it. You're the shining image of a good coder.

ivanmixo:
The functionality is still there, it's just not exactly the same weird old

PR #12969, opened 2023-05-13
Updates Xeno Toxin Hud Desc

shelteredfreak:
## About The Pull Request

Hey all, Dagger the Defiler main here. 

I've been on a mission to make sure all the info that you can find on Defiler is correct, to avoid misinformation. Since Defiler can tend to be a pretty complex caste for some. I love Defiler, and you should too. (But not enough to, yknow, get it nerfed 😅 )

Just a small quick change to a feature that has been left deprecated since Afflictor became a part of Defiler, and the Reagent Sight feature was never updated. Hopefully shouldn't break anything.

image

All this does is update the descriptions displayed when examining a human with any of the toxins present in them. Certainly not a massive gamechanger but very convenient to have.

image

Also updates the ability description for Select Reagent, which seemingly also had deprecated info. (Yes I know that it is HUMOROUSLY long, but it was sorta like that before too. All that info had to be readily available from the start SOMEHOW)

## Why It's Good For The Game

I think we all can agree that updated info is very nice to have.

## Changelog
:cl:
qol: Updates Xeno Toxin hud examine description info.
qol: Updates Defiler Select Reagent ability description info.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -20,10 +20,16 @@
 			msg += "<span style='font-weight: bold; color: purple;'>You sense this creature is dead.</span>\n"
 		else if(stat || !client)
 			msg += "[span_xenowarning("It doesn't seem responsive.")]\n"
-		if(reagents.get_reagent_amount(/datum/reagent/toxin/xeno_transvitox))
-			msg += "Transvitox: 40% brute/burn injuries received are converted to toxin\n"
+		if(reagents.get_reagent_amount(/datum/reagent/toxin/xeno_neurotoxin))
+			msg += "Neurotoxin: Causes increasingly intense pain and stamina damage over time, incrementing at the 20th and 45th cycle of metabolism\n"
 		if(reagents.get_reagent_amount(/datum/reagent/toxin/xeno_hemodile))
-			msg += "Hemodile: 20% stamina damage received, when damaged, and slowed by 25% (inject neurotoxin for 50% slow)\n"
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -20,10 +20,16 @@
 			msg += "<span style='font-weight: bold; color: purple;'>You sense this creature is dead.</span>\n"
 		else if(stat || !client)
 			msg += "[span_xenowarning("It doesn't seem responsive.")]\n"
-		if(reagents.get_reagent_amount(/datum/reagent/toxin/xeno_transvitox))
-			msg += "Transvitox: 40% brute/burn injuries received are converted to toxin\n"
+		if(reagents.get_reagent_amount(/datum/reagent/toxin/xeno_neurotoxin))
+			msg += "Neurotoxin: Causes increasingly intense pain and stamina damage over time, incrementing at the 20th and 45th cycle of metabolism\n"
```
QualityVan, line 24: Talking about metabolism cycles, while accurate, isn't really helpful to someone who's trying to figure out what their abilities do. Could talk about total seconds instead, make it generally more vague, or reword some other way.

shelteredfreak: I'm not sure what the exact seconds are. And aren't ticks usually server dependent? Unsure either way.

DeltaFire15: Life() (which calls chem processing, which in turn increments cycles and metabs) ticks every 2 seconds by default.

shelteredfreak: Got it.

QualityVan (approved):

PR #12970, opened 2023-05-13
Lets Valhalla work in HvH gamemodes and deletes Civil War

bluezorua:
## About The Pull Request
Does what it says on the tin

this might have the unintended effects of having armour vendors work during HvH gamemodes but I don't think there are any armour vendors in HvH maps
## Why It's Good For The Game
a non-round effecting area should not be effected by the round

also I don't think I've seen civil war run once in around the year I spent playing this server
## Changelog
:cl:
fix: Valhalla armour vendor and loadout machines will work during CP and SC
del:Civil War and most things related to it
/:cl:


ivanmixo:
Note that civil war is still in the code so it'll be affected.

Lumipharon:
Yeah this either needs to be changed to
A: A civil war check
B: Valhalla vendor specific
Or preferably C: remove civil war entirely

bluezorua:
I don't have the patience to sift trough every mention of "rebel" check if it's used decoratively or has now defunct use

PR #12974, opened 2023-05-13
Dropship Fabricator is Indestructible

tyeagg:
## About The Pull Request
be playing cas po
someone lets a xeno through and it melts the fab
you cannot fix it or replace it
now your entire job of playing cas is useless
## Why It's Good For The Game
cas po should be able to play the entire purpose of their job 
## Changelog
:cl:
balance: dropship fabricator is indestructible
/:cl:

### DIFF
--- a/code/game/objects/machinery/dropship_part_fabricator.dm
+++ b/code/game/objects/machinery/dropship_part_fabricator.dm
@@ -10,6 +10,7 @@
 	idle_power_usage = 20
 	icon = 'icons/obj/machines/drone_fab.dmi'
 	icon_state = "drone_fab_idle"
+	resistance_flags = RESIST_ALL
 	/// List of everything in queue
 	var/list/queue = list()
 	/// Whether the fabricator is currently printing something or not


ClosetedSkeletonGH:
Should've set up claymores

novaepee:
Allow fab to get slash so that PO has to fix it via blowtorch like other computers.

FIXTHERADBOXPLEASE:
This is just removing the only counterplay to a Marine job with no viable counters(CAS)?

blackdav123:
> This is just removing the only counterplay to a Marine job with no viable counters(CAS)?

you can melt all of the ammo lying around and the cas loader to force them to waste time making more or buying a new loader from req

you can also just jump the PO when they land

Al-1ce:
> This is just removing the only counterplay to a Marine job with no viable counters(CAS)?

Not everything needs a counter-play. Some things are designed that way. Crazy, right?

novaepee:
> you can also just jump the PO when they land

don't need to when PO just crush all their fuel sources and have to wait for a hot minute to get more fuel from req

FIXTHERADBOXPLEASE:
> Not everything needs a counter-play. Some things are designed that way. Crazy, right?

This is called unfun design. Why should the game be slowed to a slog so one guy can shoot from a plane for five minutes?

isotope436:
> > Not everything needs a counter-play. Some things are designed that way. Crazy, right?
> 
> This is called unfun design. Why should the game be slowed to a slog so one guy can shoot from a plane for five minutes?

cas doesnt need a counter play (espeically one thats unfixable) and even then a single xeno sneaking on the ship (let alone one thats able to use acid) is extremely rare and as stated there are a billion other ways to cuck cas plus cas isnt even a threat to xenos unless said xeno is both not paying attention and already injured among being completely useless if orbit isnt low. hell cas isnt even ment to get kills at the moment its ment to harrass xenos and be annoying.

PR #12982, opened 2023-05-14
Fixes nades erroniously appearing primed when they aren't

ivanmixo:
## About The Pull Request

Yes. Tested by putting a nade into a GL and taking it out, usually this makes it appear like it's about to go off (with the flashing !)
## Why It's Good For The Game

Bug bad.
## Changelog
:cl:
fix: Fixed nades erroniously appearing primed when they aren't
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/grenades/grenade.dm
+++ b/code/game/objects/items/explosives/grenades/grenade.dm
@@ -84,7 +84,7 @@
 /obj/item/explosive/grenade/update_overlays()
 	. = ..()
 	if(active && dangerous)
-		overlays += new /obj/effect/overlay/danger
+		. += new /obj/effect/overlay/danger
 
 
 /obj/item/explosive/grenade/proc/prime()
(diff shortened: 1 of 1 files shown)


Lumipharon:
the ! is specifically for dangerous nades, not all nades, afaik, so it check for both dangerous and active.

Lumipharon (approved):

code/game/objects/items/explosives/grenades/grenade.dm
```diff
@@ -83,7 +83,7 @@
 
 /obj/item/explosive/grenade/update_overlays()
 	. = ..()
-	if(active)
+	if(active && dangerous)
 		overlays += new /obj/effect/overlay/danger
```
Lumipharon, line 87: Oh, this is also wrong, although I know its old code. update overlays should just return a list of overlays it should currently have. So:
. += new /obj/effect/overlay/danger

Lumipharon (changes requested):
comment

Lumipharon (approved):

Derrick2Electricboogaloo:
Finally!

PR #12984, opened 2023-05-14
Runner can't begin evasion if on fire

ivanmixo:
## About The Pull Request

Title. With the way it's currently done the evasion only gets interrupted on the first life tick, which can be a bit (relatively). Seems like an oversight.
## Why It's Good For The Game

Bug bad.
## Changelog
:cl:
fix: Fixed runner being able to start evasion while on fire
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
@@ -227,10 +227,14 @@
 
 /datum/action/xeno_action/evasion/can_use_action(silent = FALSE, override_flags)
 	. = ..()
+	var/mob/living/carbon/xenomorph/xeno_owner = owner
 
 	if(evade_active) //Can't evade while we're already evading.
 		if(!silent)
-			owner.balloon_alert(owner, "Already evading")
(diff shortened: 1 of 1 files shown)


Strayed-Raven:
inb4 bug good for gameplay

code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
```diff
@@ -227,10 +227,14 @@
 
 /datum/action/xeno_action/evasion/can_use_action(silent = FALSE, override_flags)
 	. = ..()
+	var/mob/living/carbon/xenomorph/xeno_owner = owner
```
Lumipharon, line 230: very minor, but its better to put the var after the evade active check.
p.s. why the fuck doesn't the parent xeno_action just set a xeno_owner? Its used by fucking everything.

Lumipharon (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
ivanmixo: BLeh owner is just a flat `mob`

PR #12595, opened 2023-04-02
Mounted AC rework

Pariah919:
## About The Pull Request
Mounted autocannon rework. It now spawns as a moveable gun like the TAT, it is not roundstart if you're wondering..
It is now dedicated wall destroyer, dealing bonus true damage to walls, mostly effective against softer wall types like resin walls.

Overall, it fires faster, has a FULLER AUTO MODE, and the gun is balanced for heavy rifle firerate alongside the fuller auto.

Further in detail belowl
<details>

Left is new|Right is old.

Gun
-

Has burst fire, 12 round. .25 firerate, has some level of scatter.
Fire delay 0.4|0.5
Scope upgraded to TAT level.
Armor, coverage upgraded to TAT level.
Still has IFF.

Ammo
-

In general now has a minimum accurate range of 5 tiles, and does true damage to walls.

Normal
Damage profile of 30/50/1 from 40/100/8. Does 15 true damage to walls.

Flak
Damage profile of 40/20/5 from 40/20/12. Does 5 true damage to walls.

PASS 1
Autoburst reworked into burst fire, should make it shred things less.

</details>

Sprites
image

by crow, so credit to him.
## Why It's Good For The Game
Makes the autocannon unique and have a niche at the same time in both wall destruction and very long range fire support.
## Changelog
:cl:
balance: Autocannon reworked, it now has a new sprite by Crow, it now does bonus damage to walls, it does much less sunder and damage, now fires much faster and with an autoburst.
/:cl:

PR #12990, opened 2023-05-16
Autocannon rework MK2

Pariah919:
## About The Pull Request
Reopening of #12595 since I have time again now, main changes compared to the last tested version is flak no longer pens cover but does more damage, and burst is a bit more forgiving

Mounted autocannon rework. It now spawns as a moveable gun like the TAT, it is not roundstart if you're wondering..
It is now dedicated wall destroyer, dealing bonus true damage to walls, mostly effective against softer wall types like resin walls.

Overall, it fires faster, has a burst fire mode, and the gun is balanced for heavy rifle firerate alongside the burst fire.

Further in detail belowl
<details>

Left is new|Right is old.

Gun
-

Has burst fire, 12 round. .25 firerate, has some level of scatter.
Fire delay 0.4|0.5
Scope upgraded to TAT level.
Armor, coverage upgraded to TAT level.
Still has IFF.

Ammo
-

In general now has a minimum accurate range of 5 tiles, and does true damage to walls.

Normal
Damage profile of 30/50/1 from 40/100/8. Does 15 true damage to walls.

Flak
Damage profile of 50/30/5 from 40/20/12. Does 5 true damage to walls, does NOT penetrate through walls.

PASS 1
Autoburst reworked into burst fire, should make it shred things less.

</details>

Sprites
image

by crow, so credit to him.
## Why It's Good For The Game
Makes the autocannon unique and have a niche at the same time in both wall destruction and very long range fire support.
## Changelog
:cl:
balance: Autocannon reworked, it now has a new sprite by Crow, it now does bonus damage to walls, it does much less sunder and damage, now fires much faster and with an autoburst.
/:cl:


FosterGrime-Heymynameslouie:
kinda suprised this wasnt merghed before tbh

maikilangiolo:
> kinda suprised this wasnt merghed before tbh

I'm glad it didn't get merged before, because the AC was heavily nerfed. From 80 DPS 100 AP and 16 sunder to 75 DPS, 50 AP and 2 sunder, and the burst had such a scatter malus you missed 75% of shots at 3 tiles range

isotope436:
the flak ammo doesnt work (outside of being a direct impact), 

theres HEAVY lag when trying to use the scope (looks to be more of a delay), 

theres a grammar error with the smart detonation mags saying they hold 80 rounds but actually hold 100, 

the burst mode doesnt have the ability to change where you are shooting which is terrible since its shoots 12 rounds per click and the auto burst doesnt work 

also i sent you a better firing sound on discord :>

maikilangiolo:
This is better than the last. What could use improving is the bullet speed. Too slow for the ranges it's supposed to be shot at. Should get a boot considering other weapons can fit EB or BC for sniping.

The gun could also use an internal storage like the TAT for like 4 mags, maybe 5. 
Zero feedback on the airburst ammo, but that's ancient flak code for you. Does it even do damage? And how much? I know it says it has a 1x mult, but does it also carry over the AP?

Kuro020 (approved):

PR #13007, opened 2023-05-18
Variable nuke disk generator locations

Lumipharon:
## About The Pull Request
Random nuke disk locations.

Maps can now have more than three possible nuke disk gen locations. The game will randomly choose three locations to actually spawn the disk gens.

This occurs round start for Nuke war, or when the ship crashes during Crash (thus avoiding disk gens getting crushed by the Cant).

For this initial PR I've just added one or two new spots for the larger maps (and RO). I'll leave it to Grayson or someone else to take a proper look in the future, as I'm sure more spots can be added, but the existing spots may need to be moved around as part of that.
## Why It's Good For The Game
Players are very static with how the play the game. They find optimal paths and there's little reason to deviate from them. This means gameplay tends to happen in the exact same areas every single game.

This change should help spread gameplay out across the map more, and add some variety with different sets of disk locations per game.
## Changelog
:cl:
add: Nuke disk generators are now randomised in location, based on a list of possible spawn points
/:cl:


Mrrpip:
Oooh, fancy.

That said, how does it work exactly?

Can it work with nightmare inserts for example? Or is it restricted to 'guaranteed' map areas?

Lumipharon:
> Oooh, fancy.
> 
> That said, how does it work exactly?
> 
> Can it work with nightmare inserts for example? Or is it restricted to 'guaranteed' map areas?

It works with modulars, you can see the map diff where I'm changing them

RipGrayson (approved):
mapping is fine

Arturlang:
Honestly at this point it'd be better to just have generic computers around that CAN either have intel or nuke disk programs

Lumipharon:
> Honestly at this point it'd be better to just have generic computers around that CAN either have intel or nuke disk programs

It was considered, but intel computers can be close to each other without problem, where as its very undesireable for disk computers to be too close to each other.

code/game/objects/machinery/computer/nuke_disk_generator.dm
```diff
@@ -200,3 +200,20 @@
 	icon_state = "nuke_blue"
 	disk_type = /obj/item/disk/nuclear/blue
 	disk_color = "blue"
+
+//list of disk gens we want to spawn for nuke related gamemodes. Could just use subtypes, but using a list in case more gen related stuff is added in the future
+GLOBAL_LIST_INIT(nuke_disk_generator_types, list(/obj/machinery/computer/nuke_disk_generator/red, /obj/machinery/computer/nuke_disk_generator/green, /obj/machinery/computer/nuke_disk_generator/blue))
+
+/obj/structure/computer3frame/nuke_disk_candidate
```
TiviPlus, line 207: why is this a computer3frame child

TiviPlus (changes requested):
rest fine

code/game/objects/machinery/computer/nuke_disk_generator.dm
Lumipharon: Copy pasted for stats/icon, forgot to change it. Updated now.

PR #13050, opened 2023-05-22
Engineer and corpsman slots scale on squad marines

Losenis:
## About The Pull Request
They didn't scale, the number was flat from the start. Need 5 marines to unlock 1 corpsman/engi, seems like a good ratio. Unable to test because no clue how to fake population and also check remaining available slots.

## Why It's Good For The Game
You ever played deadpop and everyone was either engi, SGnr, or corpsman without a single squad marine? Yeah
You ever played highpop and had 100 marines and 1 or 2 corpsmen and engis that survived the 2 earlier hours? Yeah

image
image

## Changelog

:cl:
balance: Engi and corpsman slot now scales on pop of marines
/:cl:

### DIFF
--- a/code/datums/jobs/job/marines.dm
+++ b/code/datums/jobs/job/marines.dm
@@ -33,6 +33,8 @@ Make your way to the cafeteria for some post-cryosleep chow, and then get equipp
 	jobworth = list(
 		/datum/job/xenomorph = LARVA_POINTS_REGULAR,
 		/datum/job/terragov/squad/smartgunner = SMARTIE_POINTS_REGULAR,
+		/datum/job/terragov/squad/corpsman = SMARTIE_POINTS_REGULAR,
+		/datum/job/terragov/squad/engineer = SMARTIE_POINTS_REGULAR,
 		/datum/job/terragov/silicon/synthetic = SYNTH_POINTS_REGULAR,
 		/datum/job/terragov/command/mech_pilot = MECH_POINTS_REGULAR,
 	)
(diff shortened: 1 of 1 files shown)


MacBlaze1:
speedmerge, there have been multiple rounds I have been unable to join as an engie because its late game 100 pop

QualityVan:
Good in principle, but I'm concerned that the way job assignment cycles will cause this to not quite work at roundstart. Will be fine for any latejoin though.

Losenis:
It should work on the number of open positions by default, would it not?

image

Al-1ce:
* Medicine is free & infinite. Corpsmen do not have to worry about actually running out of chems.

* Engineers _DO_ have to worry about running out of metal. 

* It is also rather expensive, and salvaging chairs/tables etc is hardly worth it.

Basically, the more Engineers, the less and less fair metal distribution becomes. Even at our 8 current engi slots it becomes nearly impossible to be well-stocked half the time.

This would exacerbate that.

Losenis:
If you have any ideas on how to alleviate that, I'm all ears, though it may be out of scope of this PR and need a separate one. My primary concern is the round dragging out enough that there's 10 engis for 100 marines that were full 30 minutes in, 7 of which may well be dead due to the average lifespan of an unga if they didn't ssd.

image

Al-1ce:
> If you have any ideas on how to alleviate that, I'm all ears, though it may be out of scope of this PR and need a separate one. My primary concern is the round dragging out enough that there's 10 engis for 100 marines that were full 30 minutes in, 7 of which may well be dead due to the average lifespan of an unga if they didn't ssd.

Well, I guess there'd probably be a way to scale the price of metal with the amount of engineers running around. Probably outside of the scope but it is still big hit on balance.

...Not to mention rarely in any round are the total engineer slots taken.

bowlofcereal:
fuckin' based

TiviPlus (approved):
Huh forgot it didnt already

zzurge:
We gettin outta big red with this one boys

QualityVan:
> It should work on the number of open positions by default, would it not?

Yeah to be clear I'm just quibbling about interior system stuff, it shouldn't actually matter almost any of the time. This'll be fine.

PR #13059, opened 2023-05-24
Gelida changes 2023 edition

RipGrayson:
## About The Pull Request

Makes a bunch of Gelida changes in hopes of making it less awful.

Here's the new version of Gelida in screenshot form
oldbr

The changes are as follows;
-Revamps LZ2 to be less of a straight shot into caves while also making the area around it slightly more open
-Revamps LZ1 to actually be somewhat viable compared to LZ2
-Changes the cave power generators to be in the center of caves instead of the side
-Extends no build zones near the LZs somewhat
-Gives underground power generator room some lights
-Fixes mismatched LZ zones

## Why It's Good For The Game

Gelida isn't in a good place currently, it's always the same push from LZ2 into caves (which is usually ruined by a combination of overextension and the larva generation giving the xenos 15+ new bodies by the time marines make it to power gen). This is unfun and most of the time marines don't even make it to any of the disks, let alone the silo. Hopefully the changes in this PR at least alter the flow so it's not the same push every time. 

## Changelog
:cl:
balance: Altered LZ1 on Gelida to be more defensible
balance: Altered the area round LZ2 on Gelida to be slightly more open
balance: Changed the location of underground power generation on Gelida to be in the center of caves.
balance: Changed a couple random disk generation locations on Gelida.
balance: The underground power generation plant in Gelida now has lights.
balance: Extended the preshutter no build zone on Gelida some
balance: Gelida landing zones now start with flares
fix: Fixed Gelida LZ areas being reversed
/:cl:

### DIFF
--- a/_maps/map_files/gelida_iv/gelida_iv.dmm
+++ b/_maps/map_files/gelida_iv/gelida_iv.dmm
@@ -218,7 +218,7 @@
 	pixel_y = 3
 	},
 /turf/open/floor/prison/whitegreenfull2,
-/area/gelida/caves/west_caves)
+/area/gelida/cavestructuretwo)
 "afm" = (
 /obj/structure/stairs/seamless,
 /turf/open/floor/plating/ground/snow/layer0,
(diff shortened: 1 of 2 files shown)


Lumipharon (approved):

AlexCrimson:
No idea if this is the right PR, but did this change remove a bunch of weeds from around the LZs? Its quite annoying to re-weed such a massive area. Probably needs a bunch added to the map at some point.

PR #13063, opened 2023-05-24
Hypobelt modernization

BlueHNT:
## About The Pull Request
The current hypospray belt has the majority of the stuff but it hasn't really been updated in a long time, still leaving it with dexalin from the old neurotoxin days. This pull aims to give it a suitable update without making it a lifesaver 2.0.

## Why It's Good For The Game
It still has a niche and some people use it as an actual belt. A few minor changes on top of the tagger change will make it more viable while not comparing itself to other belts due to its uniqueness.

From my view it's a sidegrade to make hypospray belt viable instead of being a source of several free hypos to put into a pouch after filling them with some obscene revive or combat mix.

## Changelog

:cl:
add: Added big combatmix hypo
add: Added big isotonic hypo [replaced]
add: Added capability to hold syringe cases in the belt [limited by the amount of cases on ship]
del: Removed old dexalin hypo [replaced]
qol: Moved the bottles up to the first slots
balance: Made imialky small due to the combat mix being added
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/hypospray.dm
+++ b/code/game/objects/items/reagent_containers/hypospray.dm
@@ -562,6 +562,18 @@
 	)
 	description_overlay = "Ti"
 
+obj/item/reagent_containers/hypospray/advanced/big/combatmix
+	name = "big combat mix hypospray"
+	desc = "A hypospray loaded with combat mix. There's a tag that reads BKTT 40:40:20:20."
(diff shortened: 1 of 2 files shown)


QualityVan (approved):
I'm fine with this as is. Syringe cases are always sort of awkward but that can be worried about more later.

PR #13068, opened 2023-05-24
Xray rifle std firerate increase

Losenis:
## About The Pull Request
Xray rifle standard shot firerate is now 0.25. Piercing mode unchanged.

## Why It's Good For The Game
Gives it an actual reason to be used over piercing given it had AR12 stats for OVER TWICE the firedelay, giving it the same reasons to not be used as heat mode from the laspistol and lassniper. The firestacks don't even do enough damage to make up for it, funnily enough, and the low firerate makes it unwieldy against fast xenos like rouny that on paper it'd be supposed to be ok against. Assuming you didn't just set the 5 marines around you on fire with ff.

Final stats still deliberatedly worse than AR12.

## Changelog

:cl:
balance: xray rifle standard shot firerate now 0.25 (Comparison: AR12 has 0.2)
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -801,7 +801,7 @@
 	wield_delay = 0.5 SECONDS
 	scatter = 0
 	scatter_unwielded = 10
-	fire_delay = 0.5 SECONDS
+	fire_delay = 0.25 SECONDS
 	accuracy_mult_unwielded = 0.55
 	damage_falloff_mult = 0.3
 	mode_list = list(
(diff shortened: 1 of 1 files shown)


Lumipharon:
Not saying this is good, but I would compare it to tx8 incend ammo, which has a cost while xray ammo is free.

Losenis:
Unless you intentionally take off your ID and become bwoinkbait, you can't set half the marines on the screen on fire even if you try. Tx8 has inherent IFF, lasguns in general don't even have aim mode.

And tx8 incen is still way better since it has 15 more pen and 1.5 more sunder per shot for a gun that still shoots faster than a base xray, assuming people don't just skip the incen for high impact that straight up demolishes armor with 10 sunder and 40 damage with the same pen.

FosterGrime-Heymynameslouie:
I think increasing damage instead of the rate in which you can dispense damage is alot better, I think

cool pr nonetheless

TiviPlus:
Eh whatever could go wrong

PR #13076, opened 2023-05-25
Allow MRE to hold its food

novaepee:
## About The Pull Request

When you take packaged or unpackaged MRE meal out of MRE, you can put it back in.

## Why It's Good For The Game

muy immersion

It never made sense to me why you can't put the MRE food back into the MRE. I have eaten MRE before (source: it was revealed to me in my dream about FX, for which I have never went to), and the MRE can definitely hold its stuff back.

Can also be balance since marines were forced to put the unpackaged MRE meal inside a backpack | satchel, so this PR is also a storage balance.

## Changelog

:cl:
add: Allow marines to put packaged or unpackaged MRE meal back inside of MRE
balance: Permit marines to have more storage by not forcing marines to use backpack | satchel space for MRE food
/:cl:


blackdav123:
just put the uneaten spaghetti in your pockets

BlueHNT:
make exception for spaghetti, it needs to be in the pocket.

PR #13106, opened 2023-05-29
Silo scaling

Lumipharon:
## About The Pull Request
Makes silo larva gen scale based off the ratio of active marines to xenos.

In what has become much more apparent due to our more regular very high pop numbers, xenos can often get into a situation where they have somewhat absurd larva regen, making it extremely difficult for marines to actually reduce their numbers in any meaningful way.

This is caused by a few factors, but most important to understand is that:

1. Larva gen scales off the number of active marines on the z-level, but has **no scaling for the number of xenos**
2. Xenos directly get more larva for killing and draining/cocooning marines which **does not scale with pop**

These two points means growing xenos numbers snowballs much easier at high pop compared to lowpop.

This PR uses an 'ideal' ratio which is based on the number of larva points received from human spawns (which is about 3.077 humans per xeno). This the ratio that you should see before shutters drop.

So with this PR, if the ratio is say 2 ungas for every xeno, the larva gen would be about 65% of what it would currently be (in addition to all the extra larva points for murdering enough marines to get to that ratio).

**Note**: This is specifically to make it scale **down** to a minimum of 50% of normal gen. It won't scale above normal generation as there are already various factors to offset that.

## Why It's Good For The Game
Less overwhelming snowballing in xeno numbers at highpop.

Ultimately, this might not be needed, or may not be the ideal solution, but the issue of super high larva gen at high pop crops up more often these days, so I thought this may be an interesting solution to at least test out.

## Changelog

:cl:
balance: Silo larva gen now scales down with the ratio of marines to xenos, where xeno numbers are high
/:cl:

### DIFF
--- a/code/controllers/subsystem/silo.dm
+++ b/code/controllers/subsystem/silo.dm
@@ -15,18 +15,26 @@ SUBSYSTEM_DEF(silo)
 
 /datum/controller/subsystem/silo/fire(resumed = 0)
 	var/datum/job/xeno_job = SSjob.GetJobType(/datum/job/xenomorph)
+	var/active_humans = length(GLOB.humans_by_zlevel[SSmonitor.gamestate == SHIPSIDE ? "3" : "2"])
+	var/active_xenos = length(GLOB.alive_xeno_list_hive[XENO_HIVE_NORMAL]) + (xeno_job.total_positions - xeno_job.current_positions)
 	//The larval spawn is based on the amount of silo, ponderated with a define. Larval follow a f(x) = (x + a)/(1 + a) * something law, which is smoother that f(x) = x * something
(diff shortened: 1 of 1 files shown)


novaepee:
Good changes. I remembered asking Bravemole whether larva rates and psychic points scale with pop size, and he said no. This was in the context that 50-60 pop was the norm, and we had no reason to suspect that TGMC will increase pop. I will be curious as to what other features need to be scaled with pop as this won't be the last PR to balance TGMC to a new change of pop (that is summer break).

TiviPlus:
> Xenos directly get more larva for killing and draining/cocooning marines which does not scale with pop

This intrinsically scales with pop as more players = more dead players

ChocolateBickie:
-1

Grinding through xenos should be hard. Marines already have a 20 minutes I win button in the form of disks. Nerfing larva regen is extremely punishing.

It won't help on low pop because the issue isn't having enough larva as xeno but having enough players.

PR #13107, opened 2023-05-29
Synthetic skillset changes + defib changes

Neocloudy:
## About The Pull Request
Defibrillators will take 3.5 seconds to shock a person if the user has Medical 5 [normally 7 seconds]. Synthetics now have Medical 5. 
*[does not make defibrillators bulky anymore]*
## Why It's Good For The Game 
Medical 5 is pointless without any big difference aside from the defib healing 20% of a dead person's health. Splint health is pretty useless as synthetics or other units such as the CMO/doctors are typically doing field surgery for broken bones anyway [or a marine who just rushes straight back into combat will usually just have the splint break again], and faster syringe use is also very pointless because of hyposprays injecting things instantly and marines just being able to sleep on their own if they need to be unconscious for surgery. There's probably other stuff Medical 5 does that is so useless you don't notice it.

On the skillset change, what a synth usually does is heal and revive marines, and they're also a machine that's probably supposed to be superior to a normal Doctor at healing and reviving. Giving them a boost in medskill doesn't seem that bad alongside the defibrillator changes.
## Changelog
:cl:
add: Defibrillators will take half as long to shock a person when being used by someone with Medical 5 skill.
balance: Standard Synthetics now have Medical 5 and the defibrillator benefits above.
/:cl:


Losenis:
>Defibs also become bulky sized when you take out the cover so you can't cheese this as much.

This is only adding busywork and inputs for byond to eat up, all it does is force you to hit Z every time you want to take it out or into storage. This also disproportionately fucks up anyone that isn't a corpsman and up that'd want to multiclass a bit since they'd be fumbling the opening and closing of the defib every time on top of already getting barely if any healing from the actual defib.

chizzy376:
Iirc synth was made with 2 types in mind, one that has more medical and one that has more construction since people didn't like it when the all-rounder role had all-rounder skills (Lol)
I'm not sure if the maints will like you changing skills of synth, but I'm all for it.

Neocloudy:
Doesn't make defibs bulky anymore as losenis made a good point with occasionally having to use a defibrillator on a non-medic role

Losenis:
>Iirc synth was made with 2 types in mind, one that has more medical and one that has more construction

It wasn't. Those two synths were purely aesthetical changes much like robot models. Some random guy decided to try split them but then gave them buffs that didn't actually do anything due to the skillchecks only checking for fumbling rather than giving a bonus on higher values while nerfing them under the untested assumption those buffs did anything at all, and maints didn't spot this so the pr got merged regardless.

CursedBirb:
> Doesn't make defibs bulky anymore as losenis made a good point with occasionally having to use a defibrillator on a non-medic role

I also agree with that. Whenever i play as engineer i carry one in back for situations where getting medic is impossible or they are dead. Having to fumble with it everytime on top of spending time taking out of backpack will make it even worse

Naaanii:
Doesn't med 5 increase healing from kits and splint HP?

Synth getting med 5 is okay but the defib change is powercreepy.

Coastfront:
As no grass blade longs for embers, nothing stable yearns for change.

I, genuinely, see absolutely no reason to buff synths healing capabilities. at all.
I don't even see a reason to change defibs in this way.

I've seen all of one person ever complain about medskill being useless, and that was when somebody tried to justify gimping synths other skills to shadownerf them. Otherwise, i've genuinely never seen a single person say "We should update medical skill!" 

It's just, these things are already fine. Why are you attempting to change them? Why do we seek more stable footing when we are already on smooth tiled floors?
Synth, in terms of healing speed, is "fine". Their main draw is that they can do everything extremely well, and they already do that. They are fast to build, repair, and heal. And as you said, the vast majority of things medical skill touches doesen't matter for the roles with it, as they can just surgery instead. Which if somebody is good at it, is both faster and more effective.

This is just an arbitrary buff to synth, that will be used to justify fucking synth over in some other way by somebody else in the future.

This isn't something that has any reason to happen, nor is there a purpose for it to exist that leads anywhere worthwhile.

TiviPlus:
This is pretty powercreepy, synth isnt just medical like MO they also have engi, etc that they can do

PR #13109, opened 2023-05-29
Removes AMR bullet slowdown and laser penetration bonus

lKiseki:
## About The Pull Request
- Removes the slowdown inflicted when an AMR bullet hits a xenomorph.
- Removes the bullet penetration bonus associated with an active laser.

## Why It's Good For The Game
The AMR is a bit too oppressive as it stands right now, and has been for the longest time.
About damn time someone gave it the heavy hand.

## Changelog
:cl: Lewdcifer
balance: AMR slowdown removed, bullet penetration bonus from laser removed.
/:cl:

### DIFF
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -137,8 +137,6 @@
 //Define sniper laser multipliers
 
 #define SNIPER_LASER_DAMAGE_MULTIPLIER 1.5 //+50% damage vs the aimed target
-#define SNIPER_LASER_ARMOR_MULTIPLIER 1.5 //+50% penetration vs the aimed target
-#define SNIPER_LASER_SLOWDOWN_STACKS 3 // Slowdown applied on hit vs the aimed target.
 
 //Define lasrifle
 #define ENERGY_STANDARD_AMMO_COST 20
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -878,8 +878,6 @@ So if we are on the 32th absolute pixel coordinate we are on tile 1, but if we a
 
 	if(proj.shot_from && src == proj.shot_from.sniper_target(src))
 		damage *= SNIPER_LASER_DAMAGE_MULTIPLIER
-		proj.penetration *= SNIPER_LASER_ARMOR_MULTIPLIER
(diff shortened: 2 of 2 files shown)


BonkaiB:
In all honest opinion having played against it and with it as well as used it. The oppressiveness is much of a variability and of course this is obvious as with every req gun it depends on the player. The gun at the moment i believe is in a good spot balance wise after it got rebalanced #10147 of which the major thing added back was the slowdown on dotted shot since all things considering that's the only thing that gets it chosen over the auto with exception of the bypass over xenos when dotting

The removal of the damage pen from the dot as well as the removal of the slowdown will infact make this sniper not worth it's price in req at any price since a DMR or two would outpace it since you can scope-walk with DMR or any other sniper, This doesn't mean that these snipers are strong or oppressive it just means that could do the job better since you don't need to buy ammo from req or have to constantly scope and un-scope. Yes the damage may still be high however your ability to hit these shots at range depended extremely on many factors, The slowdown limiting these factors to let you hit your shots. The damage doesn't matter if you cannot hit your shots when dotted since you cannot manually aim the dot to predict xeno movement. It auto aims for you

TDLR: It got the heavy hand before and when it did it became a fairly weak sniper even with just the slowdown removed in which at the end of the day all you really need to defeat it is some walls and good game sense to back off when the big fat red dot is on you. Damage pen removal is honestly fine however the removal of the slowdown will immediately make it a significantly worse gun than the majority of snipers given for free from vendors or the auto

FosterGrime-Heymynameslouie:
cope pr got killed by jammers with amr lol

Naaanii:
Double whammy nerf. No compensation. Ye olde 'I ded' PR.

Would be nicer to give it a focus as a slowdown spam weapon with sunder or a high damage dealing weapon, or split the two into two rounds idk. Shots can ironically be dodged by moving parallel or even diag and any cloud cover removes the lase so there are clear counters. Both nerfs remove any threat the AMR has honestly and while off screen combat is frustrating to play against, this would effectively kill the weapon.

Losenis:
One nerf at a time, jesus

Araulius:
remove its ability to shoot past defenders while you're at it, so there's actually a way to siege fob against it.

zzurge:
AMR haters when they see a single resin wall

PR #13111, opened 2023-05-29
SG can give points for SG, synth, and MP slots

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

Smartgunners giving points for smartgunner slots is debatable, but all marines count points to smartgunner slots, so it makes sense that smartgunners that already spawned do the same.

Smartgunners giving points for synth and MP slots make sense, and I was surprised that smartgunners didn't.

## Changelog

:cl:
balance: SG can give points for SG, synth, and MP slots
fix: SG can give points for SG, synth, and MP slots
/:cl:


Lumipharon:
This is deliberate as far as I understand, unless anyone has any compelling reason to change it.

PR #13135, opened 2023-06-01
Fix variants

Lumipharon:
## About The Pull Request
Fixes variant outfits.

I missed a couple of dumb things because 515 has no break points so I didn't see any runtimes. I have no idea how it magically worked without initial to begin with though.
## Why It's Good For The Game
Bug fix good, Lumi bad.
## Changelog
:cl:
fix: fixed an issue with variant outfits
/:cl:


Lumipharon:
TM because simple fix and it breaks some guh stuff

PR #13136, opened 2023-06-02
Minimap improvement

Lumipharon:
## About The Pull Request
The map background will now be black instead of transparent.

image
## Why It's Good For The Game
Better readability, especially when there's lots of stuff going on onscreen.
## Changelog
:cl:
qol: Improved minimap readability with a black background
/:cl:


RipGrayson (changes requested):
Odd request, tcomms shouldn't be embedded in a wall. 
oldbr3
I have no idea if it impacts functionality, but I don't like the way it looks.

RipGrayson (approved):
Apparently it makes the minimap ugly if the space is retained, also tcomms work fine while being in a wall/

GreatKingWombat:
Can you drop the alpha down to about 70% and see how that looks? Without some transparency, this is a flat nerf.

maikilangiolo:
Please don't. Not only it looks uglier, it's the opposite of an improvement having a giant black square blocking your vision

ivanmixo:
Better readability but also makes it harder to see what's going on behind the map..

chizzy376:
Why not have this as a togglable preference?

Lumipharon:
Reduced opacity

TiviPlus:
> Why not have this as a togglable preference?

not feasible

PR #13137, opened 2023-06-02
Ravager endure rework

imsxz:
## About The Pull Request
Changes Ravagers endure threshold from -100 to negative max HP(-290, -310, -330, -350). The ravager will die if it's below -100HP at the time endure ends.
## Why It's Good For The Game
Half of ravs power is basically inaccessible because going for super rage is way too risky in most circumstances. You can go from 0HP to -100 faster than you can click the rage button. With this PR, my hope is that rav will feel better to take risks since there's a margin for error when trying to time your Rage activation. I am going to restate that Ravager will die instantly if it's not above the death threshold by the time Endure is over, so it's still incredibly risky and likely to result in many deaths. The idea is that Ravager will try to heal out of the death threshold with the healing granted from super rage.
## Changelog
:cl:
balance: Ravager's Endure death threshold is much higher, but you will die if you're below the normal death threshold when Endure is over.
/:cl:


Lumipharon:
This is going to be mega denied btw. Super rage is already extremely good, this makes it effectively godmode, due to the combination of huge health pool, ability to refresh endure repeatedly, drain slash and vampirism.
Traditional counters to super rage like burst damage from db etc will be ineffective as it has so much health now it can just eat is and keep healing.

imsxz:
> ability to refresh endure repeatedly, 

i couldnt find endure refreshing in the code

> Traditional counters to super rage like burst damage from db etc will be ineffective as it has so much health now it can just eat is and keep healing.

i feel like ravs are still going to die a lot unless they kill the whole room

Barnet2:
Ravager is in no need of a buff, in fact it was nerfed a month or two ago. There is a reason super rage is dead, keep it that way. You can ball on ravager without rage regardless.

Lumipharon:
> > ability to refresh endure repeatedly,
> 
> i couldnt find endure refreshing in the code
> 
> > Traditional counters to super rage like burst damage from db etc will be ineffective as it has so much health now it can just eat is and keep healing.
> 
> i feel like ravs are still going to die a lot unless they kill the whole room

Its part of drain slash.

Derrick2Electricboogaloo:
Holy hell, current one is already a get out of jail free card, you would never kill a rav unless it just suicided on you, and even then.

Pariah919:
Back in the day (like 1+ years ago good god) endure used to be like -200 HP and ravager basically never died, -100 is already a huge ehp boost in a game dominated by health differences as little as 100 between tiers.

Araulius:
yeah, endure doesn't refresh when you activate rage. it isn't a thing.

zzurge:
Ravager is supposed to be a glass cannon for how much raw DPS it has,this feels like coating a skill issue especially since ravagers can often put marines out of the fight much faster than their peers can

zzurge:
Dont even get me started on vampirism...

Lumipharon:
> yeah, endure doesn't refresh when you activate rage. it isn't a thing.
> 
> activate rage, endure is on cool down, and it's not coming off cooldown.

It's triggered by drain slash. Every time you attack you get +2 seconds.

Closing as I see no compelling reason for this.

PR #13138, opened 2023-06-02
Egg laying for Queen

MiniMeatwad:
## About The Pull Request

Egg laying now that there is less Queen button bloat. Egg laying was removed for bloat.

## Why It's Good For The Game

So Queen has something to do while building, also it's fun. Anyone who brings a plasma pistol will be immune to huggers anyways.

## Changelog

:cl:
balance: Queen egg laying added back.
/:cl:


TiviPlus:
The point was to remove marginal utility like this and replace it with actual abilities not just to add back marginal utility

MiniMeatwad:
> The point was to remove marginal utility like this and replace it with actual abilities not just to add back marginal utility

I was told it was for removing bloat while having conversation with you about it.

But whatever, I don't think anyone's adding any ability any time soon.

Joe13413:
why the fuck do coders hate egg laying for some reason i actually cannot understand
many people i've seen have expressed that they want to see this on queen and for some reason they always just remove it smh

PR #13141, opened 2023-06-02
Fixes hive threat reward

Lumipharon:
## About The Pull Request
The reward from hive threat was failing sometimes, it turns out due to the order in which stuff happens during xeno init was leaving null entries in the normal hive list due to admeme/valhalla xenos.

## Why It's Good For The Game
Bug fix good.
## Changelog
:cl:
fix: fixed hive threat rewards not working some times
/:cl:


TiviPlus:
Changed a couple of lazyremoves to just -=, since there wasn't really any reason for the lazyremoves.

Can you explain please

PR #13145, opened 2023-06-02
new crosshair 

tyeagg:
## About The Pull Request
new mouse pointer stuff

https://github.com/tgstation/TerraGov-Marine-Corps/assets/98577688/cbd610ac-ac78-49a8-9616-2ba0484709e7
## Why It's Good For The Game
simpler and easier to read
## Changelog
:cl:
qol: different gun types have different crosshairs
fix: semi auto guns have crosshair when clicked now
imageadd: new crosshairs
code: use update mouse pointer proc instead of initial
/:cl:


theta891:
can you make a setting that allow us to change its color? i fine green chrosshairs better then white

Losenis:
I'd rather have it be a triangle like the normal cursor on normal usage because there's a good chunk of things you have to pixelhunt, like tactreloading or just grabbing something out of a pile of 20 objects in the same tile.

Losenis:
Wait is it even CENTERED AT ALL?
image

Strayed-Raven:
personally I do not mind the crosshairs when you are shooting a gun on full auto or something. That old one needed to be done better than what we currently have. But for regular mouse clicking and dragging I am just gonna override that.

iLikeMemes:
This should probably be something that can be enabled/disabled in the options.

chizzy376:
Why not let us color it? This is extremely bad for visibility

Pinkufeck:
make it be a preference option and i don't care.

novaepee:
> make it be a preference option and i don't care.

already is

image

Kapu1178:
I really dislike this and I think it actively worsens the new player experience.

Burger1998:
make this off by default

MisterBomb7:
I like the crosshair appearing while using guns, but the dot should not replace the cursor outside of lobby.

Lumipharon:
Even when I turn off the new cursor, it causes constant cursor flicking for whatever reason.

novaepee:
Now people can't disable custom cursor.

Strayed-Raven:
After trying this out. It is superior to completely cut it out to make it back to normal. Will continue to do this until this is made a preference because a users OWN CUSTOM cursors are far superior to this.

Kuro020 (changes requested):
when you're shooting the gun its fine, but the white dot instad of a normal cursor is really not, its hard to see when theres a lot going on and its too small, just put the normal cursor back in in non shooty mode

tyeagg:
> when you're shooting the gun its fine, but the white dot instad of a normal cursor is really not, its hard to see when theres a lot going on and its too small, just put the normal cursor back in in non shooty mode

thats been removed decades ago

chizzy376:
Even when you disable the custom cursor (Like it should be), the new aiming cursor remains which is a bug

Kuro020 (approved):

Lumipharon:
@tyeagg what's happening with this?

PR #13156, opened 2023-06-04
AMR special ammo buffs

Pariah919:
## About The Pull Request
Makes the sidegrade ammo types for AMR more distinct and useful.

Flak
+ 2 slowdown on direct hit.
+ Airburst multiplier is now .5 from a middling .25.
- It now does 15 sunder rather than 25.

This moves flak to being the crowd control round, with its rather middling pen it doesn't do much to highly armored castes unless they let you keep shooting at them, but it has some solid crowd control on it, rather than its current state where its just a crusher trolling round.

Incend
+ Damage increased to 70 from 50
+ Penetration increased to 30 (This is just the Pen it had with lase.)

The loss of Pen bonus makes this round even more terrible, it has awful sunder, all it does is deny pheros which is a semi-useful tool but there is little reason to pick this over normal rounds right now so this should ideally make it a better pick.

## Why It's Good For The Game
AMRs special ammo types got hit indirectly with #13109 a lot when they lost slowdown and the pen bonus, I don't think the normal rounds should have slowdown (as your mainly using them to kill whatever your target is) but this should ideally shift the trio of rounds into specialized focuses. 

## Changelog

:cl:
add: Added more things
del: Removed old things
balance: AMR Flak now does slowdown and has a higher airburst multiplier, but has reduced sundering. AMR Incend does a bit more damage even if it is still less than normal ammo as well as slightly more pen.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1119,9 +1119,9 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	name = "flak sniper bullet"
 	hud_state = "sniper_flak"
 	damage = 90
-	penetration = 10
+	penetration = 0
 	sundering = 15
-	airburst_multiplier = 0.25
+	airburst_multiplier = 0.5
 
 /datum/ammo/bullet/sniper/flak/on_hit_mob(mob/victim, obj/projectile/proj)
 	staggerstun(victim, proj,  slowdown = 2)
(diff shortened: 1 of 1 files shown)


isotope436:
this gun does not need a buff its already reguarded as top 3 best marine req weapons

Pariah919:
> this gun does not need a buff its already reguarded as top 3 best marine req weapons

Nobody touches special ammo with the AMR, they're all statistically very bad. Flak's airburst damage is basically chip tier, incend does such bad damage that you might as well just shoot people with normal rounds considering that they'll retreat anyway since the loss of whatever pheromones is basically irrelevant in the face of the overwhelming damage you do.

I have not touched normal ammo in this PR. I am attempting to make the specialized ammo have a reason to exist beyond being a noobtrap noob RO's buy by accident and never buy again.

Tonklaatech:
> this gun does not need a buff its already reguarded as top 3 best marine req weapons

Bro they just nerf flank like 3 days ago calm down

Lumipharon:
The slowdown is functionally useless, since it has a default max range of 5 that you haven't overridden

Pariah919:
> The slowdown is functionally useless, since it has a default max range of 5 that you haven't overridden

oh heck i forgor

Issue #13112, opened 2023-05-29
Smartgunners cannot spawn with backpack and instead spawn with satchel


novaepee:
### Description

I want to spawn with backpack, but I instead get satchel.

### Test Merges

[#12892: 'Adds a delay to throwing' by ivanmixo at commit a99898c494](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12892)
[#12964: 'TGUI-ify the strip menu' by ivanmixo at commit c8e5c1c9a1](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12964)
[#12995: 'Laser rework' by Lumipharon at commit 27c0719c18](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12995)
[#12893: 'Acid refactor + Acid buff against flares' by Zar000 at commit c67872f686](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12893)
[#13078: 'Command TTS volume is increased' by TiviPlus at commit 19b300e1ce](https://github.com/tgstation/TerraGov-Marine-Corps//pull/13078)

### Reproduction Steps

1. Go to game preference
2. Go to gear customization
3. Select backpack
4. Spawn in as SG
5. Get satchel instead
6. Very sad

PR #13160, opened 2023-06-04
Makes the back slot pref actually work

ivanmixo:
## About The Pull Request

image
This actually determines what you spawn with now.

Removed the default bags from most roles so you don't get spawned with a satchel/backpack anyway.

The gear you pick now gets equipped into the appropriate slots on spawn if possible, otherwise it spawns them in your bag.

Also removed a pointless human var

Closes https://github.com/tgstation/TerraGov-Marine-Corps/issues/13112

Also kinda unrelated thing but I renamed the backpacks from the job vendors like med/engi to indicate it's not just a normal IMP backpack you get, but the specialized one.
## Why It's Good For The Game

A pref not being useless is good.
## Changelog
:cl:
qol: The specialized job vendors now actually indicated the backpacks are not the normal IMP
qol: The gear you equip now gets spawned on you instead of your bag, or failing that it'll spawn in your bag
fix: The back slot pref actually works now
/:cl:


Arctrooper8:
What happens if you pick nothing but still have stuff like sunglasses or a hat. Where do those spawn then?

ivanmixo:
You don't get them, but that's for another PR since it's a bit out of scope and more of an issue with how gear is given out

PR #13178, opened 2023-06-07
Smart minigun now does true damage to walls.

Pariah919:
## About The Pull Request
Smart machinegun now does true damage to walls, this is mostly aimed at resin walls so the value is low, being 6.

This combined with the normal damage makes it do 10 damage to walls.

Mathematically this makes it probably the best wall clearing gun behind the autocannon, and should give it some level of niche beyond being a noobtrap when in a maze fight.

## Why It's Good For The Game
Gives the Smart Minigun a reason to exist as it can now help clear the maze much better than before when laying down fire.

## Changelog

:cl:
balance: Smart minigun now deals bonus damage to walls to improve its ability to destroy softer forms of walls like resin or wooden walls with ease.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1225,6 +1225,11 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	sundering = 1
 	damage_falloff = 0.1
 
+/datum/ammo/bullet/smart_minigun/on_hit_turf(turf/T, obj/projectile/P)
+	if(istype(T, /turf/closed/wall))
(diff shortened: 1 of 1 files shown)


Joe13413:
honestly based, smartminigun didnt really have much going for it before.
You traded a backpack slot for SLIGHTLY higher dps, and ammo economy
sg-29 has way more versatility that allows it to not be shit when you get stunned

XSlayer300:
hmm, do you think you can do it for the pulse energy rifle too?

Lumipharon:
probably wouldn't merge this until we're finished messing around with quick build.

Also I've got a pr up fixing all projectiles not using their AP when hitting turfs.

Pariah919:
will revive after quikbld ™️ is all ogre

PR #13184, opened 2023-06-08
Improves caste limits slightly

DeltaFire15:
## About The Pull Request
This PR improves the per-caste maximum code to better facilitate numbers that aren't 1 or INFINITY for the limit of a xeno caste.
Additionally, makes the number display in the hive status if limited.

Backendwise, "no limit" now also is symbolized by -1 instead of INFINITY, as I'd not like to pass infinity to tgui out of fear of different infinities.
## Why It's Good For The Game
While this feature is not currently in use except for "unique" castes. fixing it up should support those who may want to use it.
## Changelog
:cl:
code: Per-caste xeno limits should now be slightly more convenient to use and also display in the hive status for non-unique (but still limited) castes.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/evolution.dm
@@ -276,9 +276,15 @@
 		if(death_timer)
 			to_chat(src, span_warning("The hivemind is still recovering from the last [initial(new_caste_type.display_name)]'s death. We must wait [DisplayTimeText(timeleft(death_timer))] before we can evolve."))
 			return FALSE
-	var/maximum_active_caste = new_caste_type.maximum_active_caste
-	if(maximum_active_caste != INFINITY && maximum_active_caste <= length(hive.xenos_by_typepath[new_mob_type]))
-		to_chat(src, span_warning("There is already a [initial(new_caste_type.display_name)] in the hive. We must wait for it to die."))
+	var/maximum_active = new_caste_type.maximum_active
+	if(maximum_active != -1 && maximum_active <= length(hive.xenos_by_typepath[new_mob_type]))
+		switch(maximum_active)
+			if(0)
+				to_chat(src, span_warning("Evolving into a [initial(new_caste_type.display_name)] is not possible."))
+			if(1)
+				to_chat(src, span_warning("There is already a [initial(new_caste_type.display_name)] in the hive. We must wait for it to die."))
+			else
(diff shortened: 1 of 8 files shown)


code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -231,10 +231,10 @@
 	var/vent_exit_speed = XENO_DEFAULT_VENT_EXIT_TIME
 	///Whether the caste enters and crawls through vents silently
 	var/silent_vent_crawl = FALSE
-	// The amount of xenos that must be alive in the hive for this caste to be able to evolve
+	///The amount of xenos that must be alive in the hive for this caste to be able to evolve
 	var/evolve_min_xenos = 0
-	// How many of this caste may be alive at once
-	var/maximum_active_caste = INFINITY
+	///How many of this caste may be alive at once. -1 = unlimited.
+	var/maximum_active = -1
```
BraveMole, line 237: -1 = unlimited is not an improvement IMO

code/modules/mob/living/carbon/xenomorph/evolution.dm
```diff
@@ -276,9 +276,15 @@
 		if(death_timer)
 			to_chat(src, span_warning("The hivemind is still recovering from the last [initial(new_caste_type.display_name)]'s death. We must wait [DisplayTimeText(timeleft(death_timer))] before we can evolve."))
 			return FALSE
-	var/maximum_active_caste = new_caste_type.maximum_active_caste
-	if(maximum_active_caste != INFINITY && maximum_active_caste <= length(hive.xenos_by_typepath[new_mob_type]))
-		to_chat(src, span_warning("There is already a [initial(new_caste_type.display_name)] in the hive. We must wait for it to die."))
+	var/maximum_active = new_caste_type.maximum_active
+	if(maximum_active != -1 && maximum_active <= length(hive.xenos_by_typepath[new_mob_type]))
```
BraveMole, line 280: if it's INFINITY you don't need to check maxium_active != -1 or maxium_active != INFINITY

BraveMole (changes requested):

code/modules/mob/living/carbon/xenomorph/evolution.dm
DeltaFire15: If you want me to instead have to pass another value to the UI because trusting defines to stay the same is bad, be my guest.

DeltaFire15:
mmm no.

PR #13185, opened 2023-06-09
Genghis incendiary charge

QualityVan:
## About The Pull Request
Attach it to a resin wall or door and it'll burn through all connected ones in straight lines. It won't reach around corners. 
Available for engineer points (15) and in req (150). Both are pretty off the cuff and might need adjusting.
Features codersprites, but it's just a palette swap. I was gonna give it its own fire color too but that's way too many sprites for me.
Also fixes not being able to attach plastiques to resin walls, and makes them actually update sprite properly when attached to stuff.
## Why It's Good For The Game
Mister Gorbachev, tear down this wall!
## Changelog
:cl:
add: EX-62 incendiary charge, available to engineers and through req.
fix: You can attach plastic explosives to resin walls
fix: Plastic explosives update their sprite properly
/:cl:


Joe13413:
actually seems like a based idea, but will it instantly put fire on all walls/doors? could cause issues with xenos instantly getting hit by fire without being able to see it coming

Joe13413:
now that i've said something, the fire hitting xenos is probably never gonna do anything other than maybe deal like 20 damage to a xeno once in a while

blackdav123:
seems like something that would instantly melt down most of caves

if this ends up being a problem this could probably be changed into a mortar round or cas missile or something

QualityVan:
There's a short propagation delay between tiles, so between that, the ten second delay on the charge going off after being set, and xeno vision there's no particular reason you'd get caught in it aside from getting stuck in a dead end.

It only going in straight lines from the origin and not turning corners is largely because of stuff like cave walls, yeah. Obviously it'll depend on layout but hopefully it's restricted enough there.

maikilangiolo:
> instantly melt down most of caves

It's 150 points for, essentially, a worse flamer. Single use, only straight line of walls, only works on resin walls. A flamer might be slower, but it can demolish an entire maze

KerbalCitizen:
Xenos can just leave some holes in their chinawalls to make the thermal charges end. Marginally less effective against marines not using thermal charges, but against thermal charges, you just made them spend 15(0) req points or 15 engi points on like 8 resin walls.

FosterGrime-Heymynameslouie:
heres how to check what sprites have been changed or added in a pr so anyone else passing by reading this can check for it themselves

image

image

heres the sprites by the way

image

zzurge:
Buss nut

Lumipharon:
I don't really like hard counter concepts, and this exists for the sole and exclusive purpose of shutting down spammed resin structures.

QualityVan:
I think in this case a hard counter is justified because any soft counter that actually handles massed walls comfortably is going to also be even better at destroying normally-used walls than existing options, and that's not particularly needed at the moment.

Lumipharon:
I don't think massed walls is specifically an issue though. It is(was) an issue due to high pop quickbuild allowing you to spam literally thousands of walls at no delay. With the quickbuild rework this shouldn't be possible. Unless someone spends their entire game doing it, in which case uh, good for them, but they ain't helping their team.

Either way, I'd prefer to let the quickbuild rework pan out first to see how things land, before doing something like this pr, or even mjp's smartgun pr

TiviPlus (approved):

TiviPlus:
@Kuro020

PR #13192, opened 2023-06-09
Combat patrol is no longer votable

imsxz:
## About The Pull Request
makes combat patrol not votable anymore
## Why It's Good For The Game
it's a straight downgrade from SC. i believe that CP detracts from public opinion about HvH because it's just so ass comparatively.
## Changelog
:cl:
del: Combat patrol is no longer a votable gamemode
/:cl:

### DIFF
--- a/code/datums/gamemodes/combat_patrol.dm
+++ b/code/datums/gamemodes/combat_patrol.dm
@@ -1,6 +1,7 @@
 /datum/game_mode/combat_patrol
 	name = "Combat Patrol"
 	config_tag = "Combat Patrol"
+	votable = FALSE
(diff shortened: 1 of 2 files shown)


Lumipharon:
If people don't want to play it, they won't vote for it. I tend to find SC better over all, but people have been voting patrol more recently so.

imsxz:
> If people don't want to play it, they won't vote for it.

if that were true i can guarantee some maps would never get played at all simply because the majority of the players would prefer a different map. half the time people just vote for the first option, or whatever already has the most votes. 

If there was a gamemode option to just watch 5 roombas go in a straight line for an hour, with no fighting, people would still vote for it

PR #13193, opened 2023-06-09
Makes the strap on certain items an element

ivanmixo:
## About The Pull Request

Yes. The random sleep deletions are because they could've potentially made the signal sleep according to spacemandmm.

## Why It's Good For The Game

Element is better than each item separately defining it's own strap functionality.

## Changelog

:cl:
code: Made item straps an element
/:cl:


Lumipharon (approved):
This really needed to be done.

Isn't there a bunch of other stuff that should use this though?

ivanmixo:
If you can list them I'll do it, couldn't think of any others that toggle nodrop on alt click

Lumipharon:
Oh yeah vali does a different thing.

Powerfist has the same functionality, but the fluff message is different.

PR #13203, opened 2023-06-10
Fix King roar reliability

Wayland-Smithy:
## About The Pull Request

Applies the roar effects in a 2 tile wide band instead of 1 while tracking victims to make sure they are not hit more than once, so that simply moving towards the King does not negate what should be one of its most effective abilities.

## Why It's Good For The Game

Fix good.

## Changelog
:cl:
fix: King's roar no longer misses things moving towards the King,
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/king/abilities_king.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/king/abilities_king.dm
@@ -195,6 +195,8 @@
 	keybinding_signals = list(
 		KEYBINDING_NORMAL = COMSIG_XENOABILITY_SHATTERING_ROAR,
 	)
+	/// Tracks victims to make sure we only hit them once
+	var/list/victims_hit = list()
 
 /datum/action/xeno_action/activable/shattering_roar/use_ability(atom/target)
 	if(!target)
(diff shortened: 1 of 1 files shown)


ivanmixo (changes requested):
Harddel potential here, cut the list after the roar is done rather than before

ivanmixo:
Actually how would you ever even get hit twice in the first place?

Lumipharon:
Being able to avoid the roar was intentional btw. Its also possible with prae/boiler/spitter abilities.

Wayland-Smithy:
> Harddel potential here, cut the list after the roar is done rather than before

Done. I had tested both ways, but didn't consider the harddel.

> Actually how would you ever even get hit twice in the first place?

I'm sure there are contrived ways like something pushes you (xeno, explosion, right after you are hit. I think it's worth preventing the potential salt.

> Being able to avoid the roar was intentional btw. Its also possible with prae/boiler/spitter abilities.

I get that, but you still have to step in acid and take damage. Timing your step towards the king is easy with any time dealation and avoids 100% of the damage and effects. Is that fair for a T4?

ivanmixo:
> I'm sure there are contrived ways like something pushes you (xeno, explosion, right after you are hit. I think it's worth preventing the potential salt.

I don't think the stun is severe enough to warrant it and it's not exactly implausible to get thrown back into a shockwave. But whatever.

ivanmixo (approved):

PR #13173, opened 2023-06-06
Quickbuild Rework

RikuTheKiller:
## About The Pull Request

Replacement for #13095

Combats chinawalls by making the quickbuild point pool shared across the hive.
Essentially, if someone makes a chinawall, the hive loses out on the rest of the maze.
## Why It's Good For The Game

Tries to make quickbuild balanced without removing it from the game.
The wonders of having quickbuild without all the chinawalls.
## Changelog
:cl:
balance: Quickbuild now uses a shared pool of points.
/:cl:

PR #13209, opened 2023-06-11
Makes deconstructing xeno structures as a xeno faster

RikuTheKiller:
## About The Pull Request

Makes the delay 1s instead of 4s.
To avoid people accidentally tearing things down during combat, I couldn't make it instant, so 1s it is.
## Why It's Good For The Game

A bug was found in #13173 that made deconstruction instant.
As it turns out, this improved xeno game flow, at least from my perspective.
Less stupid sabotage, faster gameplay, very good.
## Changelog
:cl:
balance: Deconstructing xeno structures as a xeno now takes 1s instead of 4s.
/:cl:


iforgotmeword:
meh fair i it's guess lessen the effect of bad mazes

PR #13245, opened 2023-06-14
bayoknife two-shots weeds

cheapcents:
## About The Pull Request

Bayoknives now deal as much damage as the combat knife, enough to destroy weed nodes in two hits

Combat knives can no longer be made into attachments using cable coil, but now fit inside your helmet. Because it is a small (not tiny) item, placing it in your helmet will occupy all storage slots in it and prevent you from putting anything else inside

## Why It's Good For The Game
image
This post was made before the sprites were changed, but the point still stands that some people just prefer some sprites over others, especially now that they look so different

This PR justifies both knives existing by giving them a specific characteristic over the other without taking away the core advantage that makes any knife even worth using, **killing the damned weed nodes**

Will you sacrifice space for potentially life saving hyper/dylo shots so that you can fit an MRE in your boot, will you use the bayoknife so you can change attachments on the fly, or are you just using one because you think it looks better? Doesn't matter, because both can be just as efficient as the other

## Changelog
:cl:
balance: combat knive now fits into your helmet, and can no longer be made into an attachment
balance: bayoknife now deals enough dmg to kill weed nodes in two hits
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/blades.dm
+++ b/code/game/objects/items/weapons/blades.dm
@@ -189,23 +189,6 @@
 	hitsound = 'sound/weapons/slash.ogg'
 	attack_verb = list("slashed", "stabbed", "sliced", "torn", "ripped", "diced", "cut")
 
-
-/obj/item/weapon/combat_knife/attackby(obj/item/I, mob/user)
-	if(!istype(I,/obj/item/stack/cable_coil))
-		return ..()
-	var/obj/item/stack/cable_coil/CC = I
-	if(!CC.use(5))
-		to_chat(user, span_notice("You don't have enough cable for that."))
-		return
-	to_chat(user, "You wrap some cable around the bayonet. It can now be attached to a gun.")
-	if(loc == user)
-		user.temporarilyRemoveItemFromInventory(src)
-	var/obj/item/attachable/bayonet/F = new(src.loc)
-	user.put_in_hands(F) //This proc tries right, left, then drops it all-in-one.
-	if(F.loc != user) //It ended up on the floor, put it whereever the old flashlight is.
-		F.loc = get_turf(src)
-	qdel(src) //Delete da old knife
-
 /obj/item/weapon/combat_knife/Initialize(mapload)
 	. = ..()
 	AddElement(/datum/element/scalping)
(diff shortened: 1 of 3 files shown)


ivanmixo:
Uhh no, we aren't making it a straight upgrade

cheapcents:
> Uhh no, we aren't making it a straight upgrade

Currently, the combat knife is superior to any other knife in the game. It's the only one capable of two-shotting weeds, and its only disadvantage is that you need to use 5 cable coil pieces to turn it into an attachment. There is absolutely no reason to use any other knife if you intend to keep it in your boot.

This straight upgrade certainly isn't making it any better than what the combat knife already is.

Lumipharon:
If you put cable on the combat knife, the damage goes down.
If you're carrying spare cable in case you want to turn a combat knife into a bayo, I uh, have no idea why you don't just have a bayo.
If you have a bayo on a gun, you two shot weeds already.

I'm not sure what the point of this is. Combat knife is basically just to 2 shot weeds where you can't/don't want to have a bayo on your gun.

cheapcents:
> If you put cable on the combat knife, the damage goes down.
If you're carrying spare cable in case you want to turn a combat knife into a bayo, I uh, have no idea why you don't just have a bayo.
If you have a bayo on a gun, you two shot weeds already.

I agree. Why even have the bayo attachie and the bayoknife be different items, then? Why make an item specifically only to be held in your boots when you could just detach the bayoknife?

It's pointless, but this PR makes it a little less so. And even then, it barely affects balance while giving people the chance to choose their knives' "skin".

> I'm not sure what the point of this is.

What about the "Why It's Good For The Game" doesn't make what I think clear?

Strayed-Raven:
Ahh once again a simple change causes so much controversy. Reminds me of the shoulder holster webbing being changed. Otherwise maybe agree that having the CM leftovers is bad and should just homogenize the melee options instead of grasping straws over keeping a difference between. Upvote from me

Lumipharon:
Yeah I still don't see the point of this.

One is a item specifically meant to be used as a weapon, the other is meant to be an attachment that can also be used as a weapon. This is just buffing the bayo knife because ??

PR #13256, opened 2023-06-15
Adds Russian Red autoinjectors to Marinemed Vendor

KerbalCitizen:
## About The Pull Request
Puts 30 RR autoinjectors into the marinemed vendor. 

## Why It's Good For The Game
RR is available in relatively large quantities because of the 6 RR pill bottles in Medplus vendor and 1 RR pill bottle in tadpole medical vendor. The squad marine only vendor also lets you choose injector pouch with your 2 options of pouches, which contains an RR injector. It contains 10u of RR and 5u of ryet, 15u dose setting by default. 

As per QV's request, I have made the RR injectors unable to be restocked to prevent people from farming RR and ryet. 

## Changelog
:cl:
add: Added 30 RR autoinjectors to Marinemed vendor, which cannot be restocked. 
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -265,6 +265,7 @@
 		/datum/reagent/medicine/ryetalyn = 5,
 	)
 	description_overlay = "Rr"
+	free_refills = FALSE
 
 /obj/item/reagent_containers/hypospray/autoinjector/polyhexanide
 	name = "polyhexanide autoinjector"
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -1111,6 +1111,7 @@
 			/obj/item/reagent_containers/hypospray/autoinjector/imidazoline = 20,
 			/obj/item/reagent_containers/hypospray/autoinjector/quickclot = 10,
 			/obj/item/reagent_containers/hypospray/autoinjector/medicalnanites = 20,
+			/obj/item/reagent_containers/hypospray/autoinjector/russian_red = 30,
 		),
 		"Heal Pack" = list(
 			/obj/item/stack/medical/heal_pack/gauze = -1,


FosterGrime-Heymynameslouie:
me omw to inject 30 rr injectors onto the RO (he denied my order)

ivanmixo (approved):

QualityVan (changes requested):
I'm not fond of this, especially with them being refillable. Something being available in initial loadouts doesn't mean it should be infinite from the general vendors.

KerbalCitizen:
> I'm not fond of this, especially with them being refillable. Something being available in initial loadouts doesn't mean it should be infinite from the general vendors.

I talked with marilyn about the ryet in there, first of all why is there, then secondly some concerns of farming ryet with it. The russian red itself is practically unlimited because I've never seen all of the starting RR get used up before, especially if one groundside medbay vendor is operational to give more RR pill bottles. There should be no real balance change by having unlimited RR since people never use up the currently large but limited stock. For other stuff like the synap injector in injector pouch or the sandbags and barbed wire in construction pouch, there is a big balance change if someone just made them unlimited because they're very useful and it's easy to use up the entire free supply of them. 

I was planning on changing the mix in the injectors to something more reasonable like 20u RR and 10u oxy with 15u default dose setting, which also stops ryet farming. I was under the assumption that I should make a separate PR for that because I hear atomize a lot.

ivanmixo:
> I'm not fond of this, especially with them being refillable. Something being available in initial loadouts doesn't mean it should be infinite from the general vendors.

Massive non issue because of the aformentioned pill bottles. Why would RR of all things be an issue if it were refillable?

QualityVan:
6 pill bottles is significantly less than infinite, and those don't contain rye. RR heals 20 damage per life tick for a not particularly significant downside (especially in situations where the alternative is dying and getting psidrained). 

>The russian red itself is practically unlimited because I've never seen all of the starting RR get used up before

'People don't make full use of it' isn't actually an argument in favor of buffing things.

QualityVan (approved):

PR #13259, opened 2023-06-16
Adds cloneloss to marines over time while they are cocooned

Sebbe9123:
## About The Pull Request
This PR adds cloneloss while you stay inside the cocoon. For the duration of the cocoon you get a total of 40, Double of psydrain.
## Why It's Good For The Game
Finally adds cloneloss to cocoon which I was originally gonna do last year or so when I added it to psydrain but I couldn't wrap my head around process()
Now cocoon isn't just straight up a worse option than psydrain unless the marine is garuanteed to never be rescued, now it _might_ be worth it to cocoon someone relatively close to the front lines. A freshly rescued marine might not have any impactful cloneloss, While a fully cocooned but still rescued marine (cocoon resets defib timer if you didn't know) has 40 cloneloss, or double that of psydrain.
## Changelog
:cl:
add: Adds cloneloss to cocoon.
/:cl:

### DIFF
--- a/code/game/objects/items/cocoon.dm
+++ b/code/game/objects/items/cocoon.dm
@@ -42,8 +42,8 @@
 	var/psych_points_output = COCOON_PSY_POINTS_REWARD_MIN + ((HIGH_PLAYER_POP - SSmonitor.maximum_connected_players_count) / HIGH_PLAYER_POP * (COCOON_PSY_POINTS_REWARD_MAX - COCOON_PSY_POINTS_REWARD_MIN))
 	psych_points_output = clamp(psych_points_output, COCOON_PSY_POINTS_REWARD_MIN, COCOON_PSY_POINTS_REWARD_MAX)
 	SSpoints.add_psy_points(hivenumber, psych_points_output)
-	//Gives marine cloneloss for a total of 30.
-	victim.adjustCloneLoss(0.1)
+	//Gives marine cloneloss for a total of 40.
+	victim.adjustCloneLoss(0.13)
 
 /obj/structure/cocoon/take_damage(damage_amount, damage_type, damage_flag, effects, attack_dir, armour_penetration)
 	. = ..()
--- a/code/game/objects/items/cocoon.dm
+++ b/code/game/objects/items/cocoon.dm
@@ -42,6 +42,8 @@
 	var/psych_points_output = COCOON_PSY_POINTS_REWARD_MIN + ((HIGH_PLAYER_POP - SSmonitor.maximum_connected_players_count) / HIGH_PLAYER_POP * (COCOON_PSY_POINTS_REWARD_MAX - COCOON_PSY_POINTS_REWARD_MIN))
 	psych_points_output = clamp(psych_points_output, COCOON_PSY_POINTS_REWARD_MIN, COCOON_PSY_POINTS_REWARD_MAX)
 	SSpoints.add_psy_points(hivenumber, psych_points_output)
+	//Gives marine cloneloss for a total of 30.
+	victim.adjustCloneLoss(0.1)
 
 /obj/structure/cocoon/take_damage(damage_amount, damage_type, damage_flag, effects, attack_dir, armour_penetration)
 	. = ..()
--- a/code/game/objects/items/cocoon.dm
+++ b/code/game/objects/items/cocoon.dm
@@ -42,7 +42,7 @@
 	var/psych_points_output = COCOON_PSY_POINTS_REWARD_MIN + ((HIGH_PLAYER_POP - SSmonitor.maximum_connected_players_count) / HIGH_PLAYER_POP * (COCOON_PSY_POINTS_REWARD_MAX - COCOON_PSY_POINTS_REWARD_MIN))
 	psych_points_output = clamp(psych_points_output, COCOON_PSY_POINTS_REWARD_MIN, COCOON_PSY_POINTS_REWARD_MAX)
 	SSpoints.add_psy_points(hivenumber, psych_points_output)
-	//Gives marine cloneloss for a total of 40.
+	//Gives marine cloneloss for a total of 30.
 	victim.adjustCloneLoss(0.5)
 
 /obj/structure/cocoon/take_damage(damage_amount, damage_type, damage_flag, effects, attack_dir, armour_penetration)
--- a/code/game/objects/items/cocoon.dm
+++ b/code/game/objects/items/cocoon.dm
@@ -43,7 +43,7 @@
 	psych_points_output = clamp(psych_points_output, COCOON_PSY_POINTS_REWARD_MIN, COCOON_PSY_POINTS_REWARD_MAX)
 	SSpoints.add_psy_points(hivenumber, psych_points_output)
 	//Gives marine cloneloss for a total of 40.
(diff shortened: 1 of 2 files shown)


Naaanii:
This reduces the risk reward aspect of cocooning and draining on the frontlines.

The whole point of cocooning is that the marine is unobtainable for 10 minutes or if a Xeno wants to risk it they can go ahead and cocoon not to mention that if the cocoon isn't opened up on the spot, it can be permadragged to hell which some Xenos actually do.

This ironically makes cocooning straight up better than drains and invalidates the risk/reward aspect to cocooning in the first place. I really don't see why buff the cloneloss aspect of all things when it already exists to give more psypoints or in some cases, exists to permabrazil that one annoying marine.

Sebbe9123:
> This reduces the risk reward aspect of cocooning and draining on the frontlines.
> 
> The whole point of cocooning is that the marine is unobtainable for 10 minutes or if a Xeno wants to risk it they can go ahead and cocoon not to mention that if the cocoon isn't opened up on the spot, it can be permadragged to hell which some Xenos actually do.
> 
> This ironically makes cocooning straight up better than drains and invalidates the risk/reward aspect to cocooning in the first place. I really don't see why buff the cloneloss aspect of all things when it already exists to give more psypoints or in some cases, exists to permabrazil that one annoying marine.

The cocoon is supposed to be anchored and unmoveable, you should report that bug

QualityVan:
Why the upfront 10 instead of all 40 over time?

Sebbe9123:
> Why the upfront 10 instead of all 40 over time?

I figured that if we gave a little bit of cloneloss to cocoon when it's first used, xenos would be more willing to try to cocoon marines on the front lines and fight to keep the marine cocooned instead of just sucking them and not really caring about the body after unless it's a high value target like Elsa or Oats.

Naaanii:
> > Why the upfront 10 instead of all 40 over time?
> 
> I figured that if we gave a little bit of cloneloss to cocoon when it's first used, xenos would be more willing to try to cocoon marines on the front lines and fight to keep the marine cocooned instead of just sucking them and not really caring about the body after unless it's a high value target like Elsa or Oats.

But the whole point of cocooning is to not do it so close to the frontlines where the risk of losing the body is a concern. Reducing that risk or adding a win win situation is kind of odd.

Naaanii:
> > This reduces the risk reward aspect of cocooning and draining on the frontlines.
> > The whole point of cocooning is that the marine is unobtainable for 10 minutes or if a Xeno wants to risk it they can go ahead and cocoon not to mention that if the cocoon isn't opened up on the spot, it can be permadragged to hell which some Xenos actually do.
> > This ironically makes cocooning straight up better than drains and invalidates the risk/reward aspect to cocooning in the first place. I really don't see why buff the cloneloss aspect of all things when it already exists to give more psypoints or in some cases, exists to permabrazil that one annoying marine.
> 
> The cocoon is supposed to be anchored and unmoveable, you should report that bug

It's not a bug when the nest is broken.

Lumipharon:
I agree with Naaanii. Cocooning is a greed move by xenos, banking on being able to hold the body for a far longer time for a significant boost in reward, in exchange for the risk of missing out on the reward entirely and have a greatly extended revive timer.

Any cloneloss should only be at the end, and I don't see why it should be higher than draining.

QualityVan:
I'm personally fine with cocoon having more cloneloss, in the scope of higher reward for xenos, but yeah not frontloaded. Either fully periodic or split periodic and final burst.

Tangential, but what about instead setting unrevivable if a cocoon finishes fully? That's the same as a normal death timer, but loses all the time stocked if freed. Pretty significant buff from current cocooning though.

chizzy376:
All this does is buff cocoon clone loss while not acknowledging the actual issue?

Sebbe9123:
> I'm personally fine with cocoon having more cloneloss, in the scope of higher reward for xenos, but yeah not frontloaded. Either fully periodic or split periodic and final burst.
> 
> Tangential, but what about instead setting unrevivable if a cocoon finishes fully? That's the same as a normal death timer, but loses all the time stocked if freed. Pretty significant buff from current cocooning though.

Yeah I'll make it all over time, feels like a good solution.

ivanmixo:
Yeaaah not a fan. Coccon/drain is meant to be risk v reward and this just makes cocoon better in basically every way.

zzurge:
Look man you cant have increased psi points and clone damage too,the whole point of the succ over cocoon thing is the tradeoff,instant points,instant clone,no pausing of DNR timer. This has worked

Sebbe9123:
> Yeaaah not a fan. Coccon/drain is meant to be risk v reward and this just makes cocoon better in basically every way.

Yes, cocoon WAS better in every way until i added cloneloss to suck, then it became better in every way because the increase in points and larva was never worth loosing out on the cloneloss. Cocoon is SUPPOSED TO be better than suck in every way, its how it was after headbite was removed and its how it really should be now.


> Look man you cant have increased psi points and clone damage too,the whole point of the succ over cocoon thing is the tradeoff,instant points,instant clone,no pausing of DNR timer. This has worked

We very much can, what you are spouting is something i sadly have to inform you is something you have made up without even realizing it. Cocoon is supposed to be succ but better, the tradeoff IS the pausing of the dnr timer. The only reason cocoon hasnt had cloneloss so far is that i didnt kniw how to add it when i added it to suck in the first place and i took a long ass break from the game to focus on school.

Your points kinda grow moot when you talk about risk and reward without actually looking at the game as it is now and see that there is no risk vs reward because there is no reason to ever cocoon unless the fob has already fallen.
With this addition xenos will have a reason to use cocoon when marines are still groundside, there will be a reason to cocoon marines on the front lines and fight to defend the cocoon. The thing about cocoons being moveable is a bug that is getting exploited by some xenos and should be fixed in a different pr.

shelteredfreak:
Still probably gonna use drain more than cocoon simply for getting perma marines, but this definitely makes it a lot more valuable instead of just centering around points with no real upside otherwise.

code/game/objects/items/cocoon.dm
```diff
@@ -42,6 +42,8 @@
 	var/psych_points_output = COCOON_PSY_POINTS_REWARD_MIN + ((HIGH_PLAYER_POP - SSmonitor.maximum_connected_players_count) / HIGH_PLAYER_POP * (COCOON_PSY_POINTS_REWARD_MAX - COCOON_PSY_POINTS_REWARD_MIN))
 	psych_points_output = clamp(psych_points_output, COCOON_PSY_POINTS_REWARD_MIN, COCOON_PSY_POINTS_REWARD_MAX)
 	SSpoints.add_psy_points(hivenumber, psych_points_output)
+	//Gives marine cloneloss for a total of 40.
```
QualityVan, line 45:
```suggestion
	//Gives marine cloneloss for a total of 30.
```

Lumipharon (approved):

QualityVan:
30 is a decent chunk of cloneloss, but it's about three and a half minutes of not being rescued before it outpaces just draining and in any case where the cocoon does finish, a bit of cryo time is still a lot less than the perma that would have applied if drained.

Beyond that, I'm just generally in favor of keeping drain and cocoon largely parallel in functionality.

PR #13260, opened 2023-06-16
Resin doors are destroyed in 1 second

RikuTheKiller:
## About The Pull Request

Title, as it turns out they use a different attack_alien proc.
## Why It's Good For The Game

Consistency and also this is a bug lol.
## Changelog
:cl:
fix: Doors are now destroyed in 1 second just like everything else.
/:cl:


QualityVan:
Use better commit names in the future.

RikuTheKiller:
> Use better commit names in the future.

I do that for commits that aren't a single value change or are part of a PR that also isn't a single value change.

PR #13261, opened 2023-06-16
AGLS scope swapped to TAT scope, added support grenade types.

Pariah919:
## About The Pull Request
Swaps the AGLS over to TAT scope.
Adds Cloak/Flare/Tanglefoot grenade types. Flare/Cloak are cheaper than damaging ammo at 30 points each, tangle foot is 55 points.
## Why It's Good For The Game
AGLS could use a range increase considering its unmoving, and unlike TL it isn't roundstart nor does it have insane damage to really warrant TL level scope.
The support grenades ideally should let it move more into a supporting niche compared to the other req deployables in its point range.
## Changelog
:cl:
add: Added new grenade types to the AGLS, it now has Flare/Tangle/Cloak grenade types.
balance: AGLS now has longer zoom when scoping in.
/:cl:


maikilangiolo:
Good stuff. Considerations:
1. The fact the magazines are 30 rounds. There's absolutely zero occasions you need to rapid fire 30 flares or 30 cloak nades in a row. While it makes kinda sense the AGLS gets support nades, it really would be better if those were mixed in inside an offensive magazine. Shame the code doesn't support ammo mixing in a magazine (like why????), so you would have to constantly load and unload magazines and clear the chamber (this wastes a grenade because it spawns a bugged grenade you can't put back into the magazine).
2. You need to increase the grenade explosion radius. I lowered it to 2 in a previous PR because the short scope was unusable with the large AOE (you couldn't use the grenade in most of your range because you would blow yourself up). 
3. Just make the grenade timer 1 second to blow, so the grenades blow up before you fire a second one (fire delay was increased to 1.1). Otherwise, the AGLS is unusable with grenades because they blow up and scatter, resulting in unpredictable and frankly all but effective results.

ivanmixo (approved):

Pariah919:
> Good stuff. Considerations:
> 
>     1. The fact the magazines are 30 rounds. There's absolutely zero occasions you need to rapid fire 30 flares or 30 cloak nades in a row. While it makes kinda sense the AGLS gets support nades, it really would be better if those were mixed in inside an offensive magazine. Shame the code doesn't support ammo mixing in a magazine (like why????), so you would have to constantly load and unload magazines and clear the chamber (this wastes a grenade because it spawns a bugged grenade you can't put back into the magazine).
> 
>     2. You need to increase the grenade explosion radius. I lowered it to 2 in a previous PR because the short scope was unusable with the large AOE (you couldn't use the grenade in most of your range because you would blow yourself up).
> 
>     3. Just make the grenade timer 1 second to blow, so the grenades blow up before you fire a second one (fire delay was increased to 1.1). Otherwise, the AGLS is unusable with grenades because they blow up and scatter, resulting in unpredictable and frankly all but effective results.

1. Don't see the point of reducing it, I might give them lower load times in the future though, for the rest that's just a limitation of how we do stuff sadly.
2. I think the current one is actually fine for what you get, if its still bad I'd rather make the fuze time lower like 3 suggests.

maikilangiolo:
> 1. Don't see the point of reducing it

Not what I suggested. Really, faster load time is the only thing that works. 


> 2. I'd rather make the fuze time lower

Please do.

PR #13264, opened 2023-06-16
Less wacky footstep overrides

Lumipharon:
## About The Pull Request
Stops mobs from doing a locate() every step they take, and makes it generic instead of snowflake as physically possible.

Also makes catwalks actually override footstep sounds like they were supposed to. There's still some other wacky catwalk shitcode that needs to be fixed though.
## Why It's Good For The Game
mega shitcode bad.
## Changelog
:cl:
code: Fixed some wacky footstep code
/:cl:


code/game/turfs/open.dm
```diff
@@ -48,6 +48,21 @@
 	. = ..()
 	. += ceiling_desc()
 
+///Checks if anything should override the turf's normal footstep sounds
+/turf/open/proc/get_footstep_override(footstep_type)
+	var/list/footstep_overrides = list()
+	SEND_SIGNAL(src, COMSIG_FIND_FOOTSTEP_SOUND, footstep_overrides)
+	if(!length(footstep_overrides))
+		return
```
Kapu1178, line 56: shouldnt this be
```
if(length(footstep_overrides))
	return
```

ivanmixo (approved):

PR #13270, opened 2023-06-16
Fixes debris runtime

ivanmixo:
## About The Pull Request

image

AddElement gets called in the parents of all closed turfs, then on the children which overrides it. The element itself is highlander, the signal might as well be too.
## Why It's Good For The Game

Runtime bad.
## Changelog
:cl:
fix: Fixed a debris runtime
/:cl:

### DIFF
--- a/code/datums/elements/debris.dm
+++ b/code/datums/elements/debris.dm
@@ -47,7 +47,7 @@
 	debris_velocity = _debris_velocity
 	debris_amount = _debris_amount
 	debris_scale = _debris_scale
-	RegisterSignal(target, COMSIG_ATOM_BULLET_ACT, PROC_REF(register_for_impact))
+	RegisterSignal(target, COMSIG_ATOM_BULLET_ACT, PROC_REF(register_for_impact), TRUE)
 
 /datum/element/debris/Detach(datum/source, force)
 	. = ..()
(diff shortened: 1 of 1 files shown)


Lumipharon:
It's not overridden by child types. They all just override the same proc. The issue is with how turfs get replaced by other turfs.

The turf is never deleted, it just changes type, so the previous element is not removed from the turf, but it tries to attach the element again regardless.

I don't really think this is fixing the issue

ivanmixo:
> The turf is never deleted, it just changes type, so the previous element is not removed from the turf, but it tries to attach the element again regardless.

And that's fine since by default elements are highlander, which means they'll override eachother, so the signal should override the previous one too.

code/datums/elements/debris.dm
```diff
@@ -47,7 +47,7 @@
 	debris_velocity = _debris_velocity
 	debris_amount = _debris_amount
 	debris_scale = _debris_scale
-	RegisterSignal(target, COMSIG_ATOM_BULLET_ACT, PROC_REF(register_for_impact))
+	RegisterSignal(target, COMSIG_ATOM_BULLET_ACT, PROC_REF(register_for_impact), TRUE)
```
ivanmixo, line 50:
```suggestion
	RegisterSignal(target, COMSIG_ATOM_BULLET_ACT, PROC_REF(register_for_impact), TRUE) //override because the element gets overriden
```

PR #13273, opened 2023-06-16
SMG-25 changes

Pariah919:
## About The Pull Request
Changes -
- Has gained all the unbarreled gun attachments (masterkey/PP/masterkey/mini/UGL)
- Has rifle level akimbo maluses.
- Has carbine-level falloff damage multiplier. (0.9)
- Has .4 akimbo fire delay multiplier.
## Why It's Good For The Game
Makes the SMG-25 more unique in its class, its basically a rifle with the damage profile and handling of an SMG now.
## Changelog
:cl:
balance: SMG-25 can now fit underbarreled guns.
balance: SMG-25 and its elite version now have higher akimbo maluses, and have less damage falloff over distance.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smgs.dm
+++ b/code/modules/projectiles/guns/smgs.dm
@@ -185,8 +185,12 @@
 		/obj/item/attachable/extended_barrel,
 		/obj/item/attachable/heavy_barrel,
 		/obj/item/attachable/scope/mini,
-		/obj/item/attachable/burstfire_assembly,
 		/obj/item/attachable/magnetic_harness,
+		/obj/item/attachable/burstfire_assembly,
+		/obj/item/weapon/gun/pistol/plasma_pistol,
+		/obj/item/weapon/gun/shotgun/combat/masterkey,
+		/obj/item/weapon/gun/flamer/mini_flamer,
+		/obj/item/weapon/gun/grenade_launcher/underslung,
 		/obj/item/attachable/gyro,
 	)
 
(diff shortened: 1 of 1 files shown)


Wayland-Smithy:
Overall a buff unless akimbo. Do SMGs really need less falloff?

Pariah919:
> Overall a buff unless akimbo. Do SMGs really need less falloff?

SMG-25 is basically only good because of akimbo, this nerfs the accuracy (if I remember my numbers right) but gives it rifle level falloff to complement the rifle level fire delay.

Lumipharon:
I'll stab you with a spork.

maikilangiolo:
> Overall a buff unless akimbo. 

Which is good. The akimbo smg25 is crazy

ivanmixo (approved):
An actual cross of smg and rifle, wow.

Lumipharon:
I don't think it should have the improved falloff, it's superior to actual rifles like the t18.

Lumipharon (changes requested):
I don't think it should have the improved falloff, as that's better than many rifles, like the T18.

Also the akimbo vars you've tweaked really only effect it at long range. If you want to actually reduce akimbo effectiveness, hit the akimbo firedelay penalty.

Pariah919:
> I don't think it should have the improved falloff, as that's better than many rifles, like the T18.

T-18 has a better burstfire than this, could just set it to t-18 falloff which is .9 I guess.

> Also the akimbo vars you've tweaked really only effect it at long range. If you want to actually reduce akimbo effectiveness, hit the akimbo firedelay penalty.

fair enough

PR #13276, opened 2023-06-17
Martini Rework MK.IV - Revenge of the Tea

Pariah919:
## About The Pull Request
Reworks Martini from a meme with no purpose beyond FF and triple juggle memes to something resembling a gun with a gimmick
- Ammo no longer has minimum range
- Sundering and staggerstun are entirely removed from the ammo.
- It now does 10 seconds of Shatter on hit.
- It now has aim mode, with an incredibly fast aiming time.
## Why It's Good For The Game
Martini now has some level of niche in the marine arsenal, it breaks through a large chunk of armor to allow teammates to follow up.
I removed the minimum range since it doesn't really feel like it fits here.
## Changelog
:cl:
balance: Martini rework, no longer has staggerstun nor minimum range. It now deals the shattering effect on hit instead.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1155,7 +1155,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	penetration = 20
 	accurate_range_min = 0
 	///shatter effection duration when hitting mobs
-	var/shatter_duration = 15 SECONDS
+	var/shatter_duration = 10 SECONDS
 
 /datum/ammo/bullet/sniper/martini/on_hit_mob(mob/M, obj/projectile/proj)
 	if(!isliving(M))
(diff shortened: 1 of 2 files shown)


Mrrpip:
God save the Queen.

Also, do we get red coats or not...?

Lumipharon:
Light railgun

Barnet2:
Aim mode on this should have insane slowdown for firing line LARP

Al-1ce:
Hard agree above. Slowdown should compensate.

ClosetedSkeletonGH:
it being able to be fired with one hand also should probably be removed

ivanmixo (approved):
Probably fine tbh considering martini has barely any niche right now over slugs. Duration might be a bit long but it is just martini.

ivanmixo (changes requested):
Actually a bit concerned about how fast the aiming time is. Considering you can take off 20 armor for 10 seconds it should definitely not be that fast to IFF.

PR #13281, opened 2023-06-17
remove smoke effect from mini rocket impacts

tyeagg:
## About The Pull Request
this also fixes illum flares not lighting up
## Why It's Good For The Game
this is the only explosion that does it and its breaking illum flares
## Changelog
:cl:
del: smoke effect from minirockets
/:cl:


iforgotmeword:
nope gunpowder also causes smoke

PR #13286, opened 2023-06-18
allows for more than 2 squads when Pillar of Spring is the ship map

coldironwarrior:
## About The Pull Request

removes the limitation on pillar of spring generating only two squads

## Why It's Good For The Game

allows for four teams. in the rare event that someone actually assigns teams to do things, there'll now be more teams to cover various parts of the operation.

mostly this is to allow for selecting four Squad Leader roles on join/roundstart like sulaco, or to allow for assigning up to four Squad Leader roles. Pillar of Spring gets played decently often enough on high pop that i think having four SLs is justified.

## Changelog

:cl:
balance: Pillar of Spring can now support up to 4 squads roundstart
/:cl:

### DIFF
--- a/_maps/pillar_of_spring.json
+++ b/_maps/pillar_of_spring.json
@@ -2,5 +2,5 @@
     "map_name": "Pillar of Spring",
     "map_path": "map_files/Pillar_of_Spring",
     "map_file": "TGS_Pillar_of_Spring.dmm",
-    "traits": [{"Marine Main Ship": true}],
+    "traits": [{"Marine Main Ship": true}]
 }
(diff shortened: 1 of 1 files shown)


Strayed-Raven:
Pillar of Spring...cannot support more than two squads? Guh? What?

ClosetedSkeletonGH:
xoxo could you have Theseus allow 4 squads too?

coldironwarrior:
> xoxo could you have Theseus allow 4 squads too?

Different PR but yeh. not sure if there'll be compelling reasons one way or the other on this map or that for there to only be two squads

ivanmixo (approved):
Never understood why it was only 2 squads anyway. We aren't that lowpop anymore.

PR #13294, opened 2023-06-18
Suppressor Tweak

KerbalCitizen:
## About The Pull Request
Removes the bullet speed debuff from the suppressor. Gives it aim mode time reduction.

## Why It's Good For The Game
Even before suppressor's bullet speed nerf, it was just not useful at all because either you went for recoil compensator for stats, or now for barrel charger for the bullet speed + damage bonus due to the falloff reduction. Alternatively, bayonet to wack weeds. After the bullet speed nerf, the only usage it got was placing BAS T60 in smoke and hoping xenos didn't notice they were getting shot at, which already got it removed from the gun. Now gives it a niche as well as having a barrel attachment focused on aim mode playstyles given to complete the rail attachment and under attachment ones.

## Changelog
:cl:
balance: Removed bullet speed debuff for suppressor, gives it aim mode time reduction.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -111,7 +111,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	///the delay between shots, for attachments that fire stuff
 	var/attachment_firing_delay = 0
 	///modifier for the time it takes to enter aim mode
-	var/aim_mode_mod = 0
+	var/aim_time_mod = 0
 
 	///The specific sound played when activating this attachment.
 	var/activation_sound = 'sound/machines/click.ogg'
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/gun_attachables.dm
```diff
@@ -332,13 +336,13 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
...
 	silence_mod = TRUE
 	pixel_shift_y = 16
-	attach_shell_speed_mod = -1
 	accuracy_mod = 0.1
 	recoil_mod = -2
 	scatter_mod = -2
 	recoil_unwielded_mod = -3
 	scatter_unwielded_mod = -2
 	damage_falloff_mod = 0.1
+	aim_time_mod = -10                   ///this is in deciseconds
```
mc-oofert, line 345:
```suggestion
	aim_time_mod = -1 SECONDS
```

probably should use defines

Lumipharon (changes requested):
Its had the speed reduction for 4 years, this isn't a new thing.

This change makes the suppressor almost entirely better than a recoil comp as well.

Lastly the actual code uh, doesn't actually change anything, and is also unsafe as aim time could go negative.

Naaanii:
Creep.

PR #13299, opened 2023-06-19
Jump fix

Lumipharon:
## About The Pull Request
Fixes some mobs losing passflags they naturall have after jumping (i.e. runner and warlock).
## Why It's Good For The Game
Bug fix good.
## Changelog
:cl:
fix: fixed some mobs like runner losing passtable after jumping
/:cl:


ivanmixo (approved):

Lumipharon:
Changed this back to 'revert to initial' as after thinking about it, just returning to a saved state means you could exploit it by comboing abilities.

In the future pass_flags need to be modified via traits or something though, so this won't be an issue.

PR #13304, opened 2023-06-19
Refactor flags_pass

Lumipharon:
## About The Pull Request
Tries to tackle the guh that is flags_pass.

We now have two vars.
Allow_pass_flags, which says 'things with this flag can pass us'
pass_flags, which says 'we can pass things with this flag'.

removed a load of copy paste code and brough it to the object level.
Renames some flags to be clearer, and purges unneeded ones.

Cleaned up Allow_pass_flags in a few places that I noticed, so fixes a bunch of minor behavior bugs.

You can now throw items past far more objects. Basically anything that you think you should be able to throw past, you now can. No more nading yourself in the face because you hit a crate or something.

Fixed many dropship window structures either being opaque and/or not letting lasers through as expected.
Various other code cleanup.

I'm almost certain there will be at least some issues with this PR because it touches a gorillion things. TM before merging.
## Why It's Good For The Game
Better, more usable code.
## Changelog
:cl:
refactor: refactored how pass flags work
fix: All dropship windows are now correctly transparent and allow lasers through
qol: You can throw thing past far more objects as you would realistically expect
/:cl:


Lumipharon:
Ready for review, still definitely will need TM'ing though.

ivanmixo (approved):
Don't see anything glaring on a skim. TMs will decide.

chizzy376:
Can't throw items into req pit anymore through Retractable Railings. Pls fix.

PR #13306, opened 2023-06-19
Resting via bind as a xeno now respects the use flags of the rest action

ivanmixo:
## About The Pull Request

Addresses inconsistencies between the bind and the action, clicking the action would get blocked in cases where pressing the bind wouldn't.
## Why It's Good For The Game

Prevents unintended behaviour like resting while rooted and other cases.
## Changelog
:cl:
fix: Resting via bind as a xeno now respects the use flags of the rest action
/:cl:


mrhugo13:
ROOT STUCK
ROOT STUCK
PLEASE
I BEG YOU
GRENADE!!

PR #13311, opened 2023-06-19
Fix space dealing inconsistent damage

Blundir:
## About The Pull Request

Space now always does damage by applying debuff on every step, no longer can you do weird damage evasion in space, which allowed some xenos to do abuses outside the ship. (wraith doing portals to space and xenos building a silo there)

## Why It's Good For The Game

Fix = good

## Changelog

:cl:
fix: fixed cases when damage from space turfs could've been negated
/:cl:


zzurge:
Eord spacewalk is kill,this is so sad

PR #13321, opened 2023-06-20
Smart minigun now has reduced windup/slowdown

Pariah919:
## About The Pull Request
Smart minigun now has superior handing statistics than before.

The windup has been made the same as the req minigun at .4 rather than .7.
It now has had its slowdown reduced by like, 20%~ from 1.5 to 1.2. at 1.5 it was incredibly awful. 1.2 is the same as the GPMG.
## Why It's Good For The Game
SG-85 damage on paper is fine, but the handling stats for it are all incredibly bad for it, this ideally should make it a more viable choice as a support weapon.
## Changelog
:cl:
balance: SG-85 now has 20%~ less slowdown when wielded, and has identical windup to req minigun.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -454,22 +454,13 @@ Note that this means that snipers will have a slowdown of 3, due to the scope
 	wield_delay = 1.5 SECONDS
 	flags_gun_features = GUN_AMMO_COUNTER|GUN_WIELDED_FIRING_ONLY|GUN_IFF
 	gun_skill_category = SKILL_SMARTGUN
-	gun_firemode_list = list(GUN_FIREMODE_AUTOMATIC, GUN_FIREMODE_AUTOBURST) //full auto, fuller auto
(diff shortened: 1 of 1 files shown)


Lumipharon:
plsno.

Guidesu:
plsyes

Mrrpip:
Does it come with Sandwich?

Arctrooper8:
What happened to giving it super wall damage? Also aren't the entire point of smart weapons supposed to be their accuracy and IFF? Does it fire two bullets for the price of on ammo in the backpack or are you just going to delete your ammo twice as fast while doing even less than before thanks to the new scatter?

Arctrooper8:
Don't forget to change the patch notes, this looks a lot better without the peculiar double bullets and spread. Are we still gonna get the funny wall changes by the way? I thought those were actually really interesting.

Pariah919:
> Don't forget to change the patch notes, this looks a lot better without the peculiar double bullets and spread. Are we still gonna get the funny wall changes by the way? I thought those were actually really interesting.

Delayed since quikbld extravaganza ™️ so until that I'll just let it simmer for now.

PR #13335, opened 2023-06-22
SOM adjustments

Lumipharon:
## About The Pull Request
Couple of misc minor changes for SOM.

Adds the missing coeffient bonuses to the biohazard helmet. I forgot to transfer them over when I made it a stand alone hat instead of an attachment.

Tweaks the V-34 slightly to be slightly less of a cannon at close range. 10% less base damage (27 instead of 30 damage) but reduced the falloff from 2 to 1.4.
This means at long range (above 5 tiles) its DPS will be slightly higher, while under 5 tiles it's DPS will be slightly lower.

Also reduced the folding stock accuracy bonus from 0.3 to 0.2. I put in the wrong value originally, so it actually had better accuracy than the MPI.

## Why It's Good For The Game
Minor HvH tweaks.

## Changelog

:cl:
balance: V-34 does slightly lower damage and close range, and slightly better damage at longer range
fix: fixed SOM biohazard helmet not being gas proof

### DIFF
--- a/code/modules/clothing/modular_armor/som.dm
+++ b/code/modules/clothing/modular_armor/som.dm
@@ -192,6 +192,9 @@
 	icon_state = "som_helmet_bio"
 	item_state = "som_helmet_bio"
 	soft_armor = list(MELEE = 45, BULLET = 70, LASER = 60, ENERGY = 60, BOMB = 50, BIO = 75, FIRE = 50, ACID = 70)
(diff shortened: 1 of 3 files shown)


Alexeivichi:
Love muh Kalash and SoM guns.

ivanmixo (approved):

PR #13356, opened 2023-06-25
Ravager endure crit effect

imsxz:
## About The Pull Request
This adds a screen effect for endure if you fall below 0 HP while it's active. The effect is copied from gorger's feast ability.
## Why It's Good For The Game
The HP bar has very poor clarity and you can't spare the time to look at chat to see if you're in crit yet.
## Changelog
:cl:
add: Ravager's Endure now has a screen effect when you have below critical HP in endure.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
@@ -228,7 +228,8 @@
 	ADD_TRAIT(X, TRAIT_STAGGERIMMUNE, ENDURE_TRAIT) //Can now endure impacts/damages that would make lesser xenos flinch
 	ADD_TRAIT(X, TRAIT_SLOWDOWNIMMUNE, ENDURE_TRAIT) //Can now endure slowdown
 
-	RegisterSignal(X, COMSIG_XENOMORPH_TAKING_DAMAGE, PROC_REF(damage_taken)) //Warns us if our health is critically low
+	RegisterSignal(X, COMSIG_XENOMORPH_BRUTE_DAMAGE, PROC_REF(damage_taken)) //Warns us if our health is critically low
+	RegisterSignal(X, COMSIG_XENOMORPH_BURN_DAMAGE, PROC_REF(damage_taken))
 
 	succeed_activate()
 	add_cooldown()
@@ -245,10 +246,12 @@
 /datum/action/xeno_action/endure/proc/endure_deactivate()
 	var/mob/living/carbon/xenomorph/X = owner
 
-	UnregisterSignal(X, COMSIG_XENOMORPH_TAKING_DAMAGE)
+	UnregisterSignal(X, COMSIG_XENOMORPH_BRUTE_DAMAGE)
+	UnregisterSignal(X, COMSIG_XENOMORPH_BURN_DAMAGE)
 
 	X.do_jitter_animation(1000)
 	X.endure = FALSE
+	X.clear_fullscreen("endure", 0.7 SECONDS)
 	X.remove_filter("ravager_endure_outline")
 	if(X.health < X.get_crit_threshold()) //If we have less health than our death threshold, but more than our Endure death threshold, set our HP to just a hair above insta dying
 		var/total_damage = X.getFireLoss() + X.getBruteLoss()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
```diff
@@ -228,7 +228,8 @@
 	ADD_TRAIT(X, TRAIT_STAGGERIMMUNE, ENDURE_TRAIT) //Can now endure impacts/damages that would make lesser xenos flinch
 	ADD_TRAIT(X, TRAIT_SLOWDOWNIMMUNE, ENDURE_TRAIT) //Can now endure slowdown
 
-	RegisterSignal(X, COMSIG_XENOMORPH_TAKING_DAMAGE, PROC_REF(damage_taken)) //Warns us if our health is critically low
+	RegisterSignal(X, COMSIG_XENOMORPH_BRUTE_DAMAGE, PROC_REF(damage_taken)) //Warns us if our health is critically low
```
ivanmixo, line 231: Why split into two signals?

ivanmixo (changes requested):

Lumipharon:
This applies every time you take damage under crit health. Its also quite an annoying visual effect (I made the current version of this sprite for combat stims but never used it), and takes up a lot of screen space.

A visual (on the mob) or audio cue would be nicer, but is this even needed?

code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
imsxz: For checking if you heal out of crit hp, it clears the visual. for instance if you're constantly going in and out of crit due to primo heal/super rage vamp. There is no single signal that I could find that could do this.

imsxz:
> This applies every time you take damage under crit health. Its also quite an annoying visual effect (I made the current version of this sprite for combat stims but never used it), and takes up a lot of screen space.
> 
> A visual (on the mob) or audio cue would be nicer, but is this even needed?

I'm not entirely familiar with overlays, does applying it that often cause overlap issues? I didnt notice anything wonky during testing.

Ive been playing rav a lot recently and something ive noticed is that you can go from 10HP to -90 HP from a single bullet, but your HP bar looks the exact same, which is very poor clarity. The only indication that you're about to die is chat messages. There's times where i'll be sitting right ABOVE crit hp healing from primordial, but still receiving gunfire, and I'm not sure what health im at.

In my opinion the visual being "annoying" isn't a big issue, the point is to be immediately visible regardless of where you're looking. Gorger's feast ability uses the same visual, and I didnt have any issues fighting with it being on my screen for prolonged periods.

I considered a roar for going into crit with endure, but I wanted the ravager to also know when they got OUT of crit, hence the visual that clears if you heal the damage. A visual on the rav itself would be nice, however I dont have any ideas for what it would look like, the overlay was already here so I decided to use it.

code/modules/mob/living/carbon/xenomorph/castes/ravager/abilities_ravager.dm
ivanmixo: What? Look at where the signals are sent. The general taking damage signal already encompasses both cases

imsxz: the xeno heals dont use the apply damage proc and only directly adjust burn and brute, so the xeno taking damage signal isnt used by them

ivanmixo: What a nightmare

Lumipharon:
> > This applies every time you take damage under crit health. Its also quite an annoying visual effect (I made the current version of this sprite for combat stims but never used it), and takes up a lot of screen space.
> > A visual (on the mob) or audio cue would be nicer, but is this even needed?
> 
> I'm not entirely familiar with overlays, does applying it that often cause overlap issues? I didnt notice anything wonky during testing.
> 
> Ive been playing rav a lot recently and something ive noticed is that you can go from 10HP to -90 HP from a single bullet, but your HP bar looks the exact same, which is very poor clarity. The only indication that you're about to die is chat messages. There's times where i'll be sitting right ABOVE crit hp healing from primordial, but still receiving gunfire, and I'm not sure what health im at.
> 
> In my opinion the visual being "annoying" isn't a big issue, the point is to be immediately visible regardless of where you're looking. Gorger's feast ability uses the same visual, and I didnt have any issues fighting with it being on my screen for prolonged periods.
> 
> I considered a roar for going into crit with endure, but I wanted the ravager to also know when they got OUT of crit, hence the visual that clears if you heal the damage. A visual on the rav itself would be nice, however I dont have any ideas for what it would look like, the overlay was already here so I decided to use it.

they don't stack, but its still removing and adding an overlay everytime you're shot, which is bad.

imsxz:
I cant find any examples of checking for overlays before applying them, so it seems like thats normal functionality, or am I mistaken? Though im guessing it would be an issue in this case because it's a gif overlay?

ivanmixo (approved):
I think this is fine honestly. From the code it seems that this is how most damage overlays do it (see where `flash_weak_pain` is called). (This could probably be rectified to not reload the overlay every time but that's out of scope for this.)

ivanmixo:
If it lags can just revert but I'm willing to give this a try

PR #13363, opened 2023-06-25
give king the ability to place acid well

novaepee:
## About The Pull Request

Title.

## Why It's Good For The Game

Both shrike and queen can place acid well when on fire, and they are T4. I think it will help king players enjoy being king by warding off being set on fire.

## Changelog

:cl:
balance: give king the ability to place acid well
/:cl:


Lumipharon:
No, its already literally fire immune when not sundered, and its supposed to be the more offensive oriented t4 compared to the other two. It doesn't need more advantages that make it even harder to counter.

PR #13383, opened 2023-06-28
Fixes BAS (and probably other sentries) not losing target when undeployed

ivanmixo:
## About The Pull Request

Funny bug. Gun would keep shooting even when in your hands and while undeploying.
## Why It's Good For The Game

Bug bad.
## Changelog
:cl:
fix: Fixed BAS (and probably other sentries) not losing target when undeployed
/:cl:


Lumipharon:
Shouldn't we just call stop_fire when a gun is undeployed?

ivanmixo:
> Shouldn't we just call stop_fire when a gun is undeployed?

That is not how sentries are done, no

PR #13388, opened 2023-06-29
Refactors moving between low objects

Lumipharon:
## About The Pull Request
**Walking between objects**
Added a new PASS_WALKOVER flag to indicate an object can be walked over without issue, if you're already on an object with the flag. i.e. moving from one table to another, or to a windowframe etc. Adds it to a various objects that were missing it.

Moving between low objects (like tables/windowframes etc) had some snowflake code specified on every object with this behavior (so it missed some).

Instead makes it generic. This should also be better perf I guess, since it's not looping through everything on your turf every time you try move into one of these objects.

**Climbing**
Additionally refactors climbing objects the old fashion way. This was also awful, as it would continuously cycle through the contents of both turfs for the entire duration of the do_after, then do it some more afterwards for good measure.

Also aligns it with the PASS_WALKOVER changes above, for consistant behavior.

Also changes a funny TEN SECOND stun for having a table flipped under you, to 2 seconds.
## Why It's Good For The Game
Much nicer code, less snowflake, more consistant behavior
## Changelog
:cl:
qol: Objects are more consistant in how they can be climbed or walked over
balance: Having a table flipped, or crate opened under you no longer stuns you for 10 seconds
refactor: refactored some code for moving between or climbing certain objects
/:cl:


code/game/objects/objs.dm
```diff
@@ -109,15 +109,18 @@
...
 	if((allow_pass_flags & PASS_GLASS) && (mover.pass_flags & PASS_GLASS))
 		return TRUE
 
 	if(mover?.throwing && (allow_pass_flags & PASS_THROW))
 		return TRUE
 
+	if((allow_pass_flags & PASS_LOW_STRUCTURE) && (mover.pass_flags & PASS_LOW_STRUCTURE))
+		return TRUE
+
+	if((allow_pass_flags & PASS_WALKOVER) && ismob(mover) && SEND_SIGNAL(target, COMSIG_OBJ_TRY_ALLOW_THROUGH))
```
Kapu1178, line 121: This is generally bad design, signals are not replacements for procs, they are to be used inside of existing generic behaviors.

Lumipharon: I'm all ears if you have another suggestion, I'm just trying to clear out snowflake code, and loops through turf contents.

Kapu1178: atleast follow the convention of positive return values (COMPONENT_DENY_ENTRY) interrupting action instead of the other way around

PR #13391, opened 2023-06-29
Doubles the cost of Runner's Pounce ability

chizzy376:
## About The Pull Request
Right now it's way too easy for this T1 to secure kills by pouncing marines with a 2 second stun that it can use a lot more times during combat. This PR aims to reduce the combat effectiveness of runner off weeds while not hindering his usefulness on weeds (or when under the effects of plasma regen by a nearby teammate).
Pros:
-Nothing changes when fighting on weeds
Cons:
-Can't cycle through its kit more times during a single fight off weeds
Notes:
-Makes the runner rely more on weeds
## Why It's Good For The Game
Being extremely strong for a T1, I don't want to make the runner too weak, but his offweeds combat capability is just too strong as of now.
## Changelog
:cl:
balance: Doubled the cost of runner pounce
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
@@ -72,7 +72,7 @@
 	action_icon_state = "pounce"
 	desc = "Leap at your target, tackling and disarming them."
 	ability_name = "pounce"
-	plasma_cost = 10
+	plasma_cost = 20
 	keybinding_signals = list(
 		KEYBINDING_NORMAL = COMSIG_XENOABILITY_POUNCE,
 	)


TiviPlus:
Runner doesnt need this

chizzy376:
> Runner doesnt need this

Propose idea on discord

00ze-cyclone:
Wasn’t runner entire purpose to be be the one harassing backlines tho? Where’s there’s no weeds.

chizzy376:
> Wasn’t runner entire purpose to be be the one harassing backlines tho? Where’s there’s no weeds.

There's harassing backline and then there's removing backline with 4 seconds of hardstun from the dark

Al-1ce:
> There's harassing backline and then there's removing backline with 4 seconds of hardstun from the dark


The stun is the only thing that allows the runner to inflict any meaningful damage before having to run away and heal.

I agree that Runner is a relic of CM and needs a tune-up, but nerfing the core ability is not the right way to go about it. Not like this.

Joe13413:
> > Wasn’t runner entire purpose to be be the one harassing backlines tho? Where’s there’s no weeds.
> 
> There's harassing backline and then there's removing backline with 4 seconds of hardstun from the dark

being alone should be a dangerous thing to do. Runner is REALLY good at backlining due to incredible speed & 1v1 potential. But doubling plas cost from pounce would probably interfere with evasion, or maybe snatch.
I dont think 10 extra plas is that big of a deal, but i dont think the nerf is nessacary (nor will it affect backlining)

PR #13416, opened 2023-07-03
GL fix

Lumipharon:
## About The Pull Request
Fixes the grenade loadout exploit.

A better long term solution would be to have loadouts actually vend ammo, but there's some complication to that, and WYCI.
## Why It's Good For The Game
exploit bad.
## Changelog
:cl:
fix: Grenade launchers will no longer vend free grenades through the loadout vendor
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -104,7 +104,7 @@
 		),
 		"Grenades" = list(
 			/obj/item/weapon/gun/grenade_launcher/single_shot = -1,
-			/obj/item/weapon/gun/grenade_launcher/multinade_launcher = -1,
+			/obj/item/weapon/gun/grenade_launcher/multinade_launcher/unloaded = -1,
 			/obj/item/weapon/gun/rifle/tx54 = 2,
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -520,7 +520,7 @@
 		),
 		"Grenades" = list(
 			/obj/item/weapon/gun/grenade_launcher/single_shot = -1,
-			/obj/item/weapon/gun/grenade_launcher/multinade_launcher = -1,
+			/obj/item/weapon/gun/grenade_launcher/multinade_launcher/unloaded = -1,
```
QualityVan, line 523: It's kind of funny to include valhalla in this change but it's not like it particularly matters.

QualityVan (approved):

code/game/objects/machinery/vending/marine_vending.dm
Lumipharon: If I didn't, it would mess people up trying to save loadouts in valhalla

ivanmixo (approved):

PR #13419, opened 2023-07-03
Fix division by 0 in process() of corrupted thermal generators

XElectricX:
## About The Pull Request

Xeno valhalla has corrupted generators which cause thousands of runtimes if a groundside map has no thermal generators. This adds a check where if the groundside generators list is empty, just stop the processing of corrupted generators. Also moves the addition of the generator to the list above the corruption proc in initialization otherwise corrupted generators at roundstart would shut themselves down before adding themselves to the list.

## Why It's Good For The Game

Annoying runtime spam bad, bug fix good.

## Changelog
:cl:
fix: Corrupted thermal generators check if there actually are any groundside for psy point generation.
/:cl:

### DIFF
--- a/code/modules/power/groundmap_geothermal.dm
+++ b/code/modules/power/groundmap_geothermal.dm
@@ -34,11 +34,11 @@ GLOBAL_VAR_INIT(generators_on_ground, 0)
 	update_icon()
 	SSminimaps.add_marker(src, MINIMAP_FLAG_ALL, image('icons/UI_icons/map_blips.dmi', null, "generator"))
 
-	if(corrupted)
-		corrupt(corrupted)
-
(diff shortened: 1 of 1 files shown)


QualityVan:
Do corruptable generators actually need to tick at all off groundside?

XElectricX:
I asked the same thing and Elsa informed me it gives psy points to the valhalla hive. Probably a better way to meet that goal than 30 corrupted generators. Pretty sure you only need 1 anyways since it is a flat value with the percentage of generators corrupted being what affects it.

ivanmixo:
> Do corruptable generators actually need to tick at all off groundside?

They used to when I had my math wrong, now it's just pointless.

PR #13439, opened 2023-07-06
[should probably TM] Macro optimizes get_hearers_in_view()

Kapu1178:
## About The Pull Request
Ports:
- https://github.com/DaedalusDock/daedalusdock/pull/421

### Explanation
One of the hottest procs in the game. The way it works is it uses the spatial grid to grab all hearers in a given range. Then, it assigns mobs called ears to every turf that contains a hearer. Then, it uses view() to determine who should actually hear the message based on sight. 

There's a pretty simple optimization here, given to us by Dantom back in the dark ages. `hearers()`. `hearers()` is a *very* fast proc that returns all mobs in view, ignoring darkness. We can't replace Ears with this entirely, because SS13/TGMC have objects that need to be aware of hearing, like Radios. But this is still a massive improvement.

### Benchmark
Benchmarked using Lohikar's closed-source benchmarking tools. Left-to-right:  Average Iterations, Harmonic Average Iterations, Standard Deviation, Standard Deviation As Percent.

image
The Code:
```
WORLD(5, 5, 1)
/mob/foo

/proc/_hearers()
  for(var/mob/foo/M in hearers(locate(3, 3, 1)))
    return

/proc/_view()
  for(var/mob/M in view(locate(3, 3, 1)))
    return

MAIN
  for(var/turf/T in world)
    new /mob/foo(T)
    new /obj(T)
    new /obj(T)
    new /obj(T)
  
  BEGIN_BENCH(2)
    BENCH_PHASE("view", _view())
    BENCH_PHASE("hearers", _hearers())
  END_BENCH
  ```
## Why It's Good For The Game
Speed is good.
## Changelog
:cl:
code: Saved a large amount of tick time in sound.
/:cl:

### DIFF
--- a/code/__HELPERS/spatial_info.dm
+++ b/code/__HELPERS/spatial_info.dm
@@ -102,9 +102,6 @@
 
 	var/list/assigned_oranges_ears = SSspatial_grid.assign_oranges_ears(hearables_from_grid)
 
-	var/old_luminosity = center_turf.luminosity
-	center_turf.luminosity = 6 //man if only we had an inbuilt dview()
-
 	//this is the ENTIRE reason all this shit is worth it due to how view() and the contents list works and can be optimized
 	//internally, the contents list is secretly two linked lists, one for /obj's and one for /mob's (/atom/movable counts as /obj here)
 	//by default, for(var/atom/name in view()) iterates through both the /obj linked list then the /mob linked list of each turf
@@ -118,7 +115,6 @@
 	for(var/mob/oranges_ear/remaining_ear as anything in assigned_oranges_ears)//we need to clean up our mess
 		remaining_ear.unassign()
 
-	center_turf.luminosity = old_luminosity
 	return .
 
 /**
--- a/code/__HELPERS/spatial_info.dm
+++ b/code/__HELPERS/spatial_info.dm
@@ -110,11 +110,9 @@
 	//by default, for(var/atom/name in view()) iterates through both the /obj linked list then the /mob linked list of each turf
 	//but because what we want are only a tiny proportion of all movables, most of the things in the /obj contents list are not what we're looking for
 	//while every mob can hear. for this case view() has an optimization to only look through 1 of these lists if it can (eg youre only looking for mobs)
-	//so by representing every hearing contents on a turf with a single /mob/oranges_ear containing references to all of them, we are:
-	//1. making view() only go through the smallest of the two linked lists per turf, which contains the type we're looking for at the end
-	//2. typechecking all mobs in the output to only actually return mobs of type /mob/oranges_ear
-	//on a whole this can outperform iterating through all movables in view() by ~2x especially when hearables are a tiny percentage of movables in view
-	for(var/mob/oranges_ear/ear in view(view_radius, center_turf))
+	//so by representing every hearing contents on a turf with a single /mob/oranges_ear containing references to all of them, we are
+	//able to use hearers() over view(), which is a huge speed increase.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: code/__HELPERS/spatial_info.dm [hunk 1]
```diff
@@ -110,11 +110,9 @@
...
 	//but because what we want are only a tiny proportion of all movables, most of the things in the /obj contents list are not what we're looking for
 	//while every mob can hear. for this case view() has an optimization to only look through 1 of these lists if it can (eg youre only looking for mobs)
-	//so by representing every hearing contents on a turf with a single /mob/oranges_ear containing references to all of them, we are:
-	//1. making view() only go through the smallest of the two linked lists per turf, which contains the type we're looking for at the end
-	//2. typechecking all mobs in the output to only actually return mobs of type /mob/oranges_ear
-	//on a whole this can outperform iterating through all movables in view() by ~2x especially when hearables are a tiny percentage of movables in view
-	for(var/mob/oranges_ear/ear in view(view_radius, center_turf))
+	//so by representing every hearing contents on a turf with a single /mob/oranges_ear containing references to all of them, we are
+	//able to use hearers() over view(), which is a huge speed increase.
+	for(var/mob/oranges_ear/ear in hearers(view_radius, center_turf))
```
Kylerace, line 115: You should remove setting the center turfs luminosity too

Kylerace (changes requested):

code/__HELPERS/spatial_info.dm
Kapu1178: I thought I did, forgot to push

Lumipharon:
I'm fairly skeptical about this. The entire spatialgrid system basically exists to optimise this one proc, by making view only use the mob list instead of object list internally.
So the benchmarking isn't applicable, since thats the unoptimised view.

Kapu1178:
> I'm fairly skeptical about this. The entire spatialgrid system basically exists to optimise this one proc, by making view only use the mob list instead of object list internally. So the benchmarking isn't applicable, since thats the unoptimised view.

the way view is optimized is at compile time, it checks what the var thats being used as the iterator is typed to.

Lumipharon:
> the way view is optimized is at compile time, it checks what the var thats being used as the iterator is typed to.

Yes, and it dramatically improved the performance of this proc. If you think there are gains here, you should pr it on TG and we'll port the optimisations, but this doesn't seem right to me.

Kapu1178:
> > the way view is optimized is at compile time, it checks what the var thats being used as the iterator is typed to.
> 
> Yes, and it dramatically improved the performance of this proc. If you think there are gains here, you should pr it on TG and we'll port the optimisations, but this doesn't seem right to me.

Bee optimized this a year ago. KylerAce, who is in the comments of this very PR, confirmed this is correct. I don't pr to tg anymore because i dont play tg.

Lumipharon:
benchmark the this proc specifically, if its good, its good.
I'm just surprised if this is the case it slipped past tg for 2 years because as you said, this is a hungry proc regardless.

Kapu1178:
> 

TG misses alot of incredibly obvious optimizations, they are some godlike supercoders that do no wrong. I asked Kyler and he said, "I probably just forgot". I already benchmarked it, the principles in the bench code are exactly as they should be.

Lumipharon:
Yeah you're probably right, also avoiding the funny dview style workaround is also good.
You probably still need to clean up the comments more though, since the references to view optimisations aren't relevant.

Lumipharon (approved):

TiviPlus:
you should probably tg it as well :shrug:

PR #13461, opened 2023-07-10
Drop Pod Reqtorio Addition

AusarAtkin:
## About The Pull Request

Adds Drop Pod to Req Factory Assembly

## Why It's Good For The Game

At high pop or long Ops, running out of drop pods are almost guaranteed , this should at least help with that

I'm not sure about the price and total refill amount , was thinking since this could lessen the distance to refill the drop pods in its room, it should have little to no discount (Normally 50 each, with this you pay 250 for 6) and decently long recipe (5 machine or about 250 point)

image

And first PR so i dont know what im doing
## Changelog

:cl:
add: Added Drop Pods to Reqtorio, and the refills for them to the supply console.
/:cl:


FosterGrime-Heymynameslouie:
setting up drop pod factorio inside of the pod bad itself also makes it so that you dont have to haul ass 2 dropods each with a slow powerloader from req to pod bay

speedermgh instantly

TiviPlus (approved):

ivanmixo (approved):

PR #13470, opened 2023-07-12
Gravity

Lumipharon:
## About The Pull Request
Experimental PR to add variable gravity to the server.

Gravity is modified by z-level trait (it could be modified by other things in the future like tg grav gens), and effects a few different things:

### Low gravity

- Throw things further (this combo's with jump throw...)
- Jump higher and for longer, for less stamina
- Run for longer

### High gravity

- Can't throw as far
- Jump takes more stamina, and is shorter
- Runnning eats more stamina

Note that this is **everything** that uses 'throw' which includes jetpacks, all type of xeno lunges etc.

Because this is a z-level trait thing, its defined config side so isn't on by default.
Is this a good idea? Is it FUN? Walanced?

Who knows, but it can definitely change things up, and I think it could be interesting to have maps designed with altered gravity in mind.
## Why It's Good For The Game
Interesting thing maybe.
## Changelog
:cl:
add: Added variable gravity as a z-level based function to effect various things such as running, jumping and throwing
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -36,8 +36,7 @@ SUBSYSTEM_DEF(mapping)
 	///The number of connected clients for the previous round
 	var/last_round_player_count
 
-	///shows the default gravity value for each z level. recalculated when gravity generators change.
-	///List in the form: list(z level num = max generator gravity in that z level OR the gravity level trait)
+	///shows the gravity value for each z level
 	var/list/gravity_by_z_level = list()
 
 //dlete dis once #39770 is resolved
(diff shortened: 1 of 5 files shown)


Lumipharon:
Now that I made this, I realise I should probably make grav effect jumping inside the actual jump component, so I can still easily have other jump modifying things apply.

bluezorua:
do any maps have different gravity or is this just an admin fuckery option?

Lumipharon:
this is for maps, not admins. Not applying to any maps by default however

blackdav123:
could be interesting if this was a randomized thing like 10% of rounds will be low grav and 10% of rounds will be high grav

shipside could also have a way to alter their gravity so they can decide what kind they want

ju45he:
On the subject of gravity, we need high-gravity grenades that slow Beno's and stay working for like 10 seconds until they crush themselves and become the size of a pebble.

Lumipharon:
> > this is for maps, not admins. Not applying to any maps by default however
> 
> Why add dead code?

because its a config change to turn it on, not code

bowlofcereal:
marines landing on a cool planet for the first time

the extremely high gravity in question:

FosterGrime-Heymynameslouie:
> could be interesting if this was a randomized thing like 10% of rounds will be low grav and 10% of rounds will be high grav
> 
> shipside could also have a way to alter their gravity so they can decide what kind they want

maybe like an event simmilar to ionspherics/acid rain?
something like an unstable magnetic shift event/ one of the moons passing too close by causing high/low gravity for a moment

TiviPlus:
Grayson is working on a moon map iirc so you might want to show him how to use it

PR #13475, opened 2023-07-12
Replaces the Tad Heavy RR with the Heavy Infantry Support Gun

Pariah919:
## About The Pull Request
Replaces the basically entirely unused heavy RR with the ISG-15 Heavy Infantry Support Gun ™️ 
Caliber is 15cm (155mm)
Costs 1600 (Double of a SADAR for reference, 200 less than the Wombat for reference) in req, shells are 175 points. Generally most of the cost is meant to be upfront rather than upkeep cost.

Has an 8 second windup to fire. 15 second fire delay and 8 second reload.
Shells function like normal bullets.
It fires literal howitzer shells (as in literally slightly more explosive radius, same travel speed of .75.)
It is IFF-Capable because else the entire marine force would randomly die to getting hit by it.
Has TAT level scope.

Has TAT level Armor/Mobilty but slightly more integrity.
Basically only needs sprites for this to be finished.
## Why It's Good For The Game
The Heavy RR is basically pointless, you can't really make it good because double mortar kind of accomplishes the same task at a much more manageable cost and the req cost is insane for something that'll be forever stuck in Tad.
This basically accomplishes the idea of heavy direct-fire support weapon that heavy RR was attempting to accomplish in a slightly more mobile package, if unwieldy in windup and such.
## Changelog
:cl:
add: ISG-15 HISG, a deployable direct fire support weapon that fires 15cm high explosive shells, replaces the RR-15.
del: Tad heavy RR has been replaced by the ISG-15 HISG.
/:cl:


Texicos:
this will still be useless also, sadar should have been the peak of one hit damage but you keep adding shit that tries to one up it but can't because of the balance

Pariah919:
> this will still be useless also, sadar should have been the peak of one hit damage but you keep adding shit that tries to one up it but can't because of the balance

I don't even know how this is related to sadar outside of both firing high yield HE
Though if you can't dodge a projectile moving at the rate of .75 tiles from a gun with 5 second wind up to fire I have no idea what else to say than unironic skill issue

Mrrpip:
I'd probably say more luck issue, as even if tracked by a direct click, it doesn't sound like you'd be hit by it, but anyhow.

That said, I don't quite get why you wouldn't use the normal howitzer as they share the bullet speed.

I think you can PB with the HISG, while the howi has it's minimum distance from what I can gather but uh...That doesn't sound likely nor ideal.

Anyhow, point is, uhhhhh. Do you have any more of your own thoughts on it, MJP?

Pariah919:
> I'd probably say more luck issue, as even if tracked by a direct click, it doesn't sound like you'd be hit by it, but anyhow.
> 
> That said, I don't quite get why you wouldn't use the normal howitzer as they share the bullet speed.
> 
> I think you can PB with the HISG, while the howi has it's minimum distance from what I can gather but uh...That doesn't sound likely nor ideal.
> 
> Anyhow, point is, uhhhhh. Do you have any more of your own thoughts on it, MJP?
> 
> Edit: In the case of a direct hit, I guess the inherited 100 sunder from the rocket type could be neat.

Its the dichotomy of direct versus indirect fire, mostly. After you pay the upfront cost you can fire this thing directly and have a bit more explosive damage, also the sunder *should* be 0 since I removed the tag, but I'll put 0 sundering just in case. The explosion is plenty enough.

LoganKele:
Holy shit... ISG?!?!!?!

COLONIALS!!! FOXHOLE!!! I LOVE FOXHOLE!!!


no but seriously, a 155mm HIFG? That seems pretty good! Tally hoe lads! Portside Bow!

XANAv2:
may need to make a true broadside tad with these installed.

Pariah919:
still waiting on the spriite, so will reopen this later

PR #13489, opened 2023-07-15
Adds tunnel landmarks to LV624

RipGrayson:
## About The Pull Request

This PR re-implements tunnel landmarks which we had previously but were deleted and adds them to LV624. I've also added a landmark for xeno resin jelly in case we want it in the future. I can remove atomize if needed.

## Why It's Good For The Game

It's been suggested several times that we could use placeable tunnels, I think Nova attempted it once before giving up. In this case I need the tunnels because I want to see if it relieves any pressure on xeno defense of caves before I make larger changes.

## Changelog
:cl:
add: Re-added tunnel landmarks for mapper use
add: Created a mapper landmark for resin jelly pods
balance: Added preplaced tunnel landmarks to LV624
/:cl:


ivanmixo (approved):

code/datums/gamemodes/infestation.dm
```diff
@@ -21,6 +21,12 @@
 		T.ChangeTurf(/turf/closed/wall/resin, T.type)
 	for(var/i in GLOB.xeno_resin_door_turfs)
 		new /obj/structure/mineral_door/resin(i)
+	for(var/i in GLOB.xeno_tunnel_spawn_turfs)
+		var/obj/structure/xeno/tunnel/new_tunnel = new /obj/structure/xeno/tunnel(i)
```
ivanmixo, line 25: Add a hivenumber here and on jelly pods in new, it'll not add to the proper list otherwise.
```suggestion
		var/obj/structure/xeno/tunnel/new_tunnel = new /obj/structure/xeno/tunnel(i, XENO_HIVE_NORMAL)
```

ivanmixo (changes requested):

lbnesquik:
Are those default, pre-built tunnels?

PR #13492, opened 2023-07-15
Gives PO the same timelock as Captain

Barnet2:
## About The Pull Request
Read title.
## Why It's Good For The Game
In the original PR for adding timelock, it was stated:
"These hours are not much too ask if you compare to the griefing power these roles have."
I believe that Bravemole severely underestimated, in this case, the griefing power of PO. This is a grievance as admin due to the amount of ahelps I get about under 10 hour living PO's, throwing entire op's due to the low timelock and not understanding the game properly. This will not solve all of the issue with new PO, it is a game after all, but hopefully new PO's can have more experience.
## Changelog
:cl:
admin:  Gives PO the same timelock as Captain
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -273,7 +273,7 @@ You are in charge of logistics and the overwatch system. You are also in line to
 	skills_type = /datum/skills/pilot
 	display_order = JOB_DISPLAY_ORDER_PILOT_OFFICER
 	outfit = /datum/outfit/job/command/pilot
-	exp_requirements = XP_REQ_INTERMEDIATE
+	exp_requirements = XP_REQ_EXPERT
 	job_flags = JOB_FLAG_LATEJOINABLE|JOB_FLAG_ROUNDSTARTJOINABLE|JOB_FLAG_ALLOWS_PREFS_GEAR|JOB_FLAG_PROVIDES_BANK_ACCOUNT|JOB_FLAG_ADDTOMANIFEST|JOB_FLAG_PROVIDES_SQUAD_HUD
 	jobworth = list(
 		/datum/job/xenomorph = LARVA_POINTS_SHIPSIDE_STRONG,


chizzy376:
2 years in the making, I kneel
Make it even longer, I beg

TooFewSecrets:
I don't think Tad even existed before timelocks were added. CAS PO needed to avoid FF obviously, but that's not too hard - being effective is another thing entirely but even the really good POs struggle with that sometimes. Tad PO is probably the most important marine unit outside of the mech pilot. Even FC can only suggest movements, and can't actually facilitate them the same way the Tadpole can. And while CAS FF can be punished by admins, being an idiot as Tad pilot isn't covered by the rules, even if it causes just as many deaths.
Definitely needs to be a higher expectation placed on POs nowadays. Borderline don't think 15 straight hours is enough - I would like to see an extra time requirement linked to command roles like SL or SO, but TGMC does not do that unfortunately.

bowlofcereal:
> I don't think Tad even existed before timelocks were added. CAS PO needed to avoid FF obviously, but that's not too hard - being effective is another thing entirely but even the really good POs struggle with that sometimes. Tad PO is probably the most important marine unit outside of the mech pilot. Even FC can only suggest movements, and can't actually facilitate them the same way the Tadpole can. And while CAS FF can be punished by admins, being an idiot as Tad pilot isn't covered by the rules, even if it causes just as many deaths. Definitely needs to be a higher expectation placed on POs nowadays. Borderline don't think 15 straight hours is enough - I would like to see an extra time requirement linked to command roles like SL or SO, but TGMC does not do that unfortunately.

I don't believe it did, no. And also, yeah, you're pretty on the nose. CAS FF can be easily dealt with, however an unga PO landing the tadpole in the middle of the xenos or something like this is annoying because its not exactly something you can fix without interfering in a round directly (ie giving the marines back the tadpole which is a big grey area).

Increasing the timelock won't STOP this from happening, it's still gonna happen, but it will definitely make it less likely as hopefully a PO will now have a bit more playtime and know what is expected of them, as unfortunately nobody really reads the wikis on any server these days.

RipGrayson:
Probably needed, I've seen multiple incidents over the last week or so where a PO ruined the marines chances of winning through incompetence.

Naaanii:
Kill the bloat slot so FCs and SLs can take tad as God intended.

TiviPlus:
Captain seems a little excessive shouldnt it be like RO level or so

TiviPlus:
So update, its confirmed the locks dont work

Mannybrado:
This is a great change, but then you still have meme tad options that are are, in effect, officially supported griefing. After this PR you should remove those, too.

XSlayer300:
> This is a great change, but then you still have meme tad options that are are, in effect, officially supported griefing. After this PR you should remove those, too.

Perhaps either they can be regulated in a higher timelock setting, giving the easiest and most recommended tad options the earliest, then the semi-recommended ones, then the memey ones MUCH later.

AND/OR, it can be regulated as a "fun" setting, or a rare seasonal, or automatically enabled during holidays.

honorconner:
please for the love of god merge this, bald POs are a leading cause of malt and mope. Synthetic and FC should be added as well

ivanmixo (approved):
imo (tad) PO has way more round impact than captain ever will

Lumipharon:
timelocks were apparently fixed. If anyone has an issue with the exact time lock, that'll be for another pr.

PR #13499, opened 2023-07-17
Adds the dead mob list to the server maint SS

ivanmixo:
## About The Pull Request

\+ deletes an unused helper and makes `listclearnulls` actually return TRUE/FALSE so nulls in global lists actually get logged.
## Why It's Good For The Game

Nulls in this list cause issues like completely breaking emotes. It's not really feasible to track down why there are nulls since they are most likely the results of harddels.
## Changelog
:cl:
fix: Fixed emotes breaking at random
/:cl:

### DIFF
--- a/code/__HELPERS/_lists.dm
+++ b/code/__HELPERS/_lists.dm
@@ -153,6 +153,8 @@
 
 #if DM_VERSION < 515
 /proc/listclearnulls(list/list_to_clear)
+	if(!list_to_clear)
+		return
 	var/start_len = length(list_to_clear)
 	var/list/new_list[start_len]
 	list_to_clear -= new_list
--- a/code/__HELPERS/_lists.dm
+++ b/code/__HELPERS/_lists.dm
@@ -146,23 +146,15 @@
 			return TRUE
 	return FALSE
 
-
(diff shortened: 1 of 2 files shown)


code/__HELPERS/_lists.dm
```diff
@@ -146,23 +146,15 @@
...
-
+/**
+ * Removes any null entries from the list
+ * Returns TRUE if the list had nulls, FALSE otherwise
+**/
+/proc/listclearnulls(list/list_to_clear)
+	var/start_len = length(list_to_clear)
+	var/list/new_list[start_len]
+	list_to_clear -= new_list
+	return length(list_to_clear) < start_len
```
Kapu1178, line 157: 515 adds `list.RemoveAll()` which removes all instances of a given value in a list. I benchmarked this a while ago and it's much faster than listclearnulls.
```suggestion
#if DM_VERSION < 515
/proc/listclearnulls(list/list_to_clear)
	var/start_len = length(list_to_clear)
	var/list/new_list[start_len]
	list_to_clear -= new_list
	return length(list_to_clear) < start_len
#else
#define listclearnulls(list) list.RemoveAll(null)
#warn You can remove the old listclearnulls now
#endif
```

Lumipharon (approved):

PR #13506, opened 2023-07-17
Fixes minimap not tracking mobs inside objects properly

ivanmixo:
## About The Pull Request

Basically going inside lockers, pipes, etc was fucky because your xyz vars would all get set to 0 which would result on you disappearing on the map. (it also caused runtimes)

Solved this by just `get_turf`ing the target, so the turf is tracked instead.

## Why It's Good For The Game

Runtime bad, more accurate minimap good.
## Changelog
:cl:
fix: You don't disappear on minimap anymore if you go inside an object like a locker or ventcrawl
/:cl:

### DIFF
--- a/code/controllers/subsystem/minimaps.dm
+++ b/code/controllers/subsystem/minimaps.dm
@@ -212,15 +212,17 @@ SUBSYSTEM_DEF(minimaps)
 		earlyadds += CALLBACK(src, PROC_REF(add_marker), target, hud_flags, blip)
 		return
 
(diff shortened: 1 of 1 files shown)


ClosetedSkeletonGH:
does this fix people who are in cocoons/stasis bags being shown at their location of cocooning/bagging instead of where they are actually at if they are moved?

Lumipharon (approved):

ivanmixo:
> does this fix people who are in cocoons/stasis bags being shown at their location of cocooning/bagging instead of where they are actually at if they are moved?

Yeah it should

PR #13508, opened 2023-07-17
4th explosion tier

Lumipharon:
## About The Pull Request
Adds a 4th explosion tier below light, "weak".
Weak explosions generally do half or less of light explosions, and do not natively cause throw (explosions can and sometimes do still manually override throw range).

Fairly limited use of it right now, but I'll probably expand it before the pr is merged.
20mm HE nades do 1 light instead of 2, but 3 weak range (they are now also cheaper)
Sticky nades have 2 light range instead of 3, but 3 weak range (so they're not literally just normal nades if you don't hit a beno)
AGS 2 light instead of 3, 4 weak.
Rad rockets do 4 weak instead of 3 light.

I sort of want to make grenades in general have a larger radius but more weak range at the cost of light, as currently its kinda wew to have 1 tile difference = take 100 damage or no damage at all, but I may leave that to a separate walance pr.

Also cleaned up some ex_act procs. Notably I made human and xeno ex_act consistant in format, but no balance changes other than adding a gib threshold to humans (if you get OB'd naked, you will gib but uh, pretty sure that's skill issue unless a rouny can literally rob the clothes off your back).

## Why It's Good For The Game
More granular explosions is better for balance.

## Changelog

:cl:
add: Added a new explosion tier of weak, below light explosions
balance: Rad rockets, AGS, sticky and 20mm grenades all utilise weak explosions to some extent
balance: Humans can be gibbed by devestating explosions if they lack bomb armor, just like xenos
/:cl:


XSlayer300:
Make ammo explosions due to fire use weak explosions too.

Lumipharon:
> Make ammo explosions due to fire use weak explosions too.

good idea

ivanmixo (approved):

PR #13511, opened 2023-07-18
Mounted guns are dense again

### LINKED ISSUE #12762: Crushers crush past manned deployables without affecting them
### Description

When you crush something, usually you'd expect it to affect the solid things on the tile. However, in the case of manned deployables, you will crush the person manning it, but completely bypass the actual gun, passing through it.
This will likely end up landing you somewhere you really do not want to be, with the exit blocked off, since you just crushed the person off the deployable and it is now solid again.

This is because #11189 makes deployables not dense when they are manned, which means crushing will ignore them instead of damaging them, and they were likely handled before the person was crushed off the deployable making it solid again.
Fixing it will likely require some amount of change in how density is handled for manned deployables.

### Test Merges

Irrelevant

### Reproduction Steps

1. Have somebody sit on a deployable, direction irrelevant
2. Be a crusher, crush towards the deployable and expect to crush the person and the deployable
3. Crush the person, but pass the deployable, ending up on the other side without any effect on it
4. (Bonus) Be trapped because due to not being manned it is now solid again, which can be a little problem if it was, say, in a cadeline.

### Screenshots

Probably not needed, but I'm sure I can get some if requested.

Lumipharon:
## About The Pull Request
What is says on the tin.

Mounted guns are no longer magically nondense when mounted.
This had the incredibly negative outcome of making crushers (previously the strongest counter to a mounted gun) not only entirely ineffective against mounted weapons, but it was actually almost guaranteed suicide to charge one as they'd act as one way doors.

Guns are normal objects to they have coverage like everything else. So being dense still means you can shoot the gunner (although TAT/AA gun have good coverage due to their shields), but now you can once again more easily destroy the guns 
themselves.

Closes #12762
## Why It's Good For The Game
Crushers can actually crush mounted guns again instead of just suiciding.
Easier to actually shoot a mounted gun.
## Changelog
:cl:
balance: Mounted guns are dense when mounted again. Crushers rejoice
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/mounted.dm
+++ b/code/modules/projectiles/guns/mounted.dm
@@ -192,7 +192,7 @@
 
 /obj/machinery/deployable/mounted/moveable/auto_cannon
 	resistance_flags = XENO_DAMAGEABLE|UNACIDABLE
-	coverage = 75 //has a shield
+	coverage = 85 //has a shield
 
 //-------------------------------------------------------
 //TE-9001 mounted heavy laser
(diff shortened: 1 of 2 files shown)


DeltaFire15:
This fixes #12762

ivanmixo (approved):

PR #13524, opened 2023-07-20
Revert "Purges some dead throw code"

Lumipharon:
Reverts tgstation/TerraGov-Marine-Corps#13510

So turns out it does actually do something... but specifically only for mobs, due to the way canallowpass is setup for mobs.

TM a revert until #13525 is TM'd/merged


Lumipharon:
fix merged.

PR #13526, opened 2023-07-20
Allows AI to trigger a bioscan on its own

RipGrayson:
## About The Pull Request

Yet another PR in my quest to make the AI the master of shipside systems and information. This PR allows an AI to manually trigger the ARES bioscan (with some minor changes I'll get to) every 10 minutes.

The AI bioscan is somewhat different than the regular in the following ways;

1. Its accuracy is based on how close to the planet the ship is, if you're at the lowest level it's slightly more accurate than the regular bioscan, at midlevel it's less accurate and the highest orbital level is extremely inaccurate.

2. It doesn't give any specific area the xenos are in, unlike the regular bioscan.

3. It can tell if a silo has been placed (But only if the silo exists, not where).

## Why It's Good For The Game

It's thematically appropriate and gives the AI yet another supporting ability without being overpowered. Now the AI can give the marines a rough estimate of xeno forces slightly more often than normal, or at critical junctures like when the ship is being overran.

## Changelog
:cl:
add: Added the ability for the AI to conduct a bioscan on its own
/:cl:


DeltaFire15:
Seems like a bit quick to get information for stealth situations, should have a few minutes of delay from starting a scan to actually finishing it.

FosterGrime-Heymynameslouie:
cool pr spedneghermth

RipGrayson:
> Seems like a bit quick to get information for stealth situations, should have a few minutes of delay from starting a scan to actually finishing it.

I just gave it a 45 second delay in performing the scan, I can increase that if you want.

code/_onclick/hud/ai.dm
```diff
@@ -46,6 +46,17 @@
 	var/mob/living/silicon/ai/AI = usr
 	AI.announcement_help()
 
+/atom/movable/screen/ai/bioscan
+	name = "Issue bioscan"
```
XSlayer300, line 50:
```suggestion
	name = "Issue Manual Bioscan"
```
imma stop you right there for consistency

Al-1ce:
>should have a few minutes of delay from starting a scan to actually finishing it.


The entire point of this ability is to gather information about the _current_ situation, not two or three minutes _after_ the scan button has been pressed.

A long cooldown is one thing. A timer until you receive the information (the entire purpose of the ability) is another.

lbnesquik:
> Yet another PR in my question to make the AI the master of shipside systems and information.

*quest

Speaking of cooldowns, I think it would be better to have something within 10–15 seconds of you pressing the button. It doesn't actually change anything to have it done later, but having it be done faster makes it feel more responsive, and less like the AI twiddling its thumbs. 

Also, how does it compare to the other bioscan frequency? Is it much faster?

mc-oofert:
use the actual cooldown defines

Strayed-Raven:
If we merge this can we eventually remove Nuke then?

dopamiin0:
any plans to give something similar to shrike or queen? This seems like pretty nice QOL and it'd be cool if xenos got to benefit as well

code/modules/mob/living/silicon/ai/say.dm
```diff
@@ -119,6 +119,78 @@
...
+
+///Annonce to everyone the number of xeno and marines on ship and ground
+/mob/living/silicon/ai/proc/ai_bioscan(delta = 1)
+
+	to_chat(usr, span_warning("Scanning for hostile lifeforms..."))
+	if(!do_after(usr, AI_SCAN_DELAY , TRUE, usr, BUSY_ICON_GENERIC)) //initial windup time until firing begins
+		return
+
+	#ifndef TESTING
+	if((last_ai_bioscan + COOLDOWN_AI_BIOSCAN) > world.time)
```
TiviPlus, line 135: err is this just a copypaste
why not make a generic proc in that case

TiviPlus (changes requested):
rest lgtm

ivanmixo:
address reviews please this is neat

code/modules/mob/living/silicon/ai/say.dm
RipGrayson: I did some experimenting, making bioscan entirely generic runs into a number of issues. Instead I adjusted the AI bioscan action to use the round bioscan instead of copypasting a the proc.

ivanmixo (changes requested):
Why reveal the number of silos. And why is it not just the same as a normal scan?

RipGrayson:
> Why reveal the number of silos. 

Because AI doesn't get to see the locations of xenos like a normal scan. Why? Because if they could they could invalidate shipside xenos with a click of a button.

The silos thing is an attempt to have an extra feature to compensate , but I can remove it if needed.

The other change is that the scan has differing accuracy based on ship level. Honestly this is something I can integrate into regular scan if you want, it's thematically appropriate and rewards the marines for managing ship level.

ivanmixo:
Yes do both of those

ivanmixo (approved):

PR #13529, opened 2023-07-20
Adds the Harvester Knife to Valhalla and Crash

Barnet2:
## About The Pull Request
Read title.
## Why It's Good For The Game
Seems like an oversight given all the other harvester weapons are in Valhalla and Crash. If I need to remove them from crash for whatever reason (I really don't see why I would need to), then I'll make the necessary edits.
## Changelog
:cl:
add: Harvester Knives are now in your local Valhalla and Crash vendors.
/:cl:


ivanmixo (approved):
Don't know why it's limited but hey

PR #13549, opened 2023-07-23
Get rid of AS macro

vvvv-vvvv:
## About The Pull Request

This macro stopped being relevant a long time ago. MIN_COMPILER_BUILD is 1584 but AS checks for <1540
Used `git grep -l 'AS' code | xargs sed -i 's/\bAS\b/as anything/g`

## Changelog

:cl:
code: AS macro is removed
/:cl:


ivanmixo:
Doesn't hurt to have it, plus its less typing than typing out `as anything`

Lumipharon:
We'll keep AS, it's convenient and frequently used shorthand tbh

ivanmixo:
> It causes inconsistencies. If you want to keep it it should be moved out of byond_version_compat.dm

How come? It's still technically compatibility and it doesn't cost us a thing to keep it so.

TiviPlus:
> We'll keep AS, it's convenient and frequently used shorthand tbh

Agree

Joe13413:
heehee ass macro

PR #13557, opened 2023-07-24
Fixes readying up bypassing exp requirements + adds the missing exp flags to the default config

ivanmixo:
## About The Pull Request

Still need to enable these in the server config for it to actually work.
## Why It's Good For The Game

Exp requirements working is good.
## Changelog
:cl:
fix: Fixed readying up bypassing exp requirements 
config: Added the missing exp flags to the default config
/:cl:

### DIFF
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -215,9 +215,6 @@ SUBSYSTEM_DEF(job)
 			// If the player wants that job on this level, then try give it to him.
 			if(player.client.prefs.job_preferences[job.title] != level)
 				continue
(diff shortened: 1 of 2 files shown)


Lumipharon:
Will this mess with admin setting roles or whatever? Or this creates various duplicate checks, I feel like its easier to just add this to assign_players_to_occupations as it already tries to do some of the checks in IsJobAvailable

ivanmixo:
No and no. I want this to cover all the cases where assignRole is used, there is a grand total of one duplicate check though which I'll remove ig. Admin stuff uses something else.

TiviPlus (approved):

PR #13567, opened 2023-07-25
Fixes stagger and slowdown recovery

Lumipharon:
## About The Pull Request
Fixes stagger and slowdown.
So for the last couple of years, humans have recovered from stagger twice as fast as they should.
Xenos on the other hand have recovered from stagger THREE times as fast as they should, on top of recovering from slowdown twice as fast as they should.

Funny copy proc calls.

Yes this means stagger and slow in general will be a lot better, but if they need adjustments thats out of scope of this pr.
## Why It's Good For The Game
Copy paste code when it calls to parent is bad.
soft stuns not working correctly is also not ideal.
## Changelog
:cl:
fix: fixed humans recovering from stagger twice as fast as they should
fix: fixed xenos recovering from stagger three times as fast as they should, and recovering from slowdown twice as fast as they should
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -161,7 +161,6 @@
 		if(81 to INFINITY)
 			reagent_shock_modifier += PAIN_REDUCTION_HEAVY
 
-	handle_stagger()
 	handle_disabilities()
 
 /mob/living/carbon/proc/handle_impaired_vision()
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -34,11 +34,6 @@
 	update_action_button_icons()
 	update_icons(FALSE)
 
-/mob/living/carbon/xenomorph/handle_status_effects()
-	. = ..()
-	handle_stagger() // 1 each time
-	handle_slowdown() // 0.4 each time
-
 /mob/living/carbon/xenomorph/handle_fire()
 	. = ..()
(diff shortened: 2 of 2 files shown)


MiniMeatwad:
The PR tells you that it's maintainer approved to go and re-adjust stagger timers to fit with the normal ratio. Go do it if you care.

MiniMeatwad:
Just make a PR to return the balance to normal. This is an actual bug, and I think you get permission from @Lumipharon to do said balance change yourself.

That being said, please confirm @Lumipharon

lbnesquik:
Yeah, this is not a balance change, it is a fix. Good one too.

Lumipharon:
For people that aren't aware, this actually make zero difference for MOST applications of stagger and slow.

You lose 1 stagger per life tick, and most things that apply stagger only apply 1 stack or less (buckshot etc). So the fact beno is removing 3 per tick makes no difference. The nature of the tick system means 1 stack lasts 2 seconds AT MOST (it could drop off 0.1 second later, depending on timing.

TiviPlus (approved):

srxtreme:
> For people that aren't aware, this actually make zero difference for MOST applications of stagger and slow.
> 
> You lose 1 stagger per life tick, and most things that apply stagger only apply 1 stack or less (buckshot etc). So the fact beno is removing 3 per tick makes no difference. The nature of the tick system means 1 stack lasts 2 seconds AT MOST (it could drop off 0.1 second later, depending on timing.

Aged like milk in a summer day

PR #13587, opened 2023-07-26
Kills scope on MG27 and MG60

Lumipharon:
## About The Pull Request
What it says on the tin. They're now both relegated to miniscope.

I added to the MG27 purely because the MG60 had it as well and they're directly comparable. However MG's having sniper tier scopes in general is kinda not ideal.
A: They become better than the higher tier MG's at long range due to the simple fact they can see
B: They kinda invalidate actual sniper rifles (which are already pretty bad).

Also added some missing bayonet options for certain guns.
## Why It's Good For The Game
MG's are more consistant in how they can be used, and don't encroach entirely on sniper rifles.
## Changelog
:cl:
balance: MG27 and MG60 can no longer use rail scopes
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/mounted.dm
+++ b/code/modules/projectiles/guns/mounted.dm
@@ -373,7 +373,6 @@
 		/obj/item/attachable/bayonet,
 		/obj/item/attachable/bayonetknife,
 		/obj/item/attachable/bayonetknife/som,
-		/obj/item/attachable/scope/marine,
 		/obj/item/attachable/scope/mini,
 		/obj/item/attachable/scope/unremovable/mmg,
 		/obj/item/attachable/stock/t27,
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -529,6 +529,8 @@
 	aim_slowdown = 0.5
 	attachable_allowed = list(
 		/obj/item/attachable/bayonet,
+		/obj/item/attachable/bayonetknife,
+		/obj/item/attachable/bayonetknife/som,
 		/obj/item/attachable/flashlight,
 		/obj/item/attachable/magnetic_harness,
 		/obj/item/attachable/buildasentry,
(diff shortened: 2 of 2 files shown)


Pariah919:
Maybe could reduce all rail scopes to TAT level and just leave snipers with current range.

chizzy376:
If snipers are so bad, why not buff snipers instead?

Lumipharon:
> If snipers are so bad, why not buff snipers instead?

Because this has nothing to do with buffing snipers.

ClosetedSkeletonGH:
T-47 scope on MG-60 was already a meme since if you were using it at T-47 range you arent going to be effective. its just not the weapons niche.

The normal scope I cant comment on

Lumipharon:
the railscope is the 'normal scope'. the base scope doesn't actually exist, its just miniscope or rail scope, other than gun specific types.

FosterGrime-Heymynameslouie:
kinda kills the effectiveness of both the machine guns significantly 
I think just removing the t47 scope on the mg27 and reverting back to old days would be a pretty good compromise? becuase if this goes through, this kills the mgs and would result them in being "invalidated" even more compared to snipers because sustained fire effectiveness weapons and being able to see as far as a bat does not equate well

plus, mg27 and mg60 have fall off over long distances, making them not really ideal in the role of a sniper, only reason why xenos die to mgs is because most of the time, running on a straight line in a corridor dosent really spell life longevity (an argument can be made for off screen weapons, but thats an entirely different can of worms, also side not mgs cant reach boilers with scopes because of afformentioned boolet dropoff)

plus, snipers are pretty good, you just gotta aim, if you think thats too mcuh of a requirment to use a sniper, an aim centric gun, skill issue

also ALSO, mgs scoped in are mega fuckin tunnel visioned,  youd be suprised as to how many of them fold when you just come at them from behind (they suddenly no longer have vision, their gun has been dropped/unweilded, they have to wait 1-2 seconds to wield again, more if they want to aim mode, cqc against a xeno with a gun that required you to sustain fire and aim while being ganked is synonymous with dying a violent death)

tldr: skil l issue lamoe

pilotmans:
Well this will surely effect my loadouts

ivanmixo (approved):

PR #13588, opened 2023-07-26
SH34 (DB) Nerf

Barnet2:
## About The Pull Request
Slightly increases fire delay on the DB, by 0.15 seconds.
Removes gyro as an attachment.
Reduces recoil to 1 instead of 2 while wielded.
## Why It's Good For The Game
DB as it currently stands is the best vendor weapon due to it's frankly insane DPS with no downside due to it not being hit with the juggle nerf, and the ability to be reloaded very quickly with good keybinds. This keeps it as a burst weapon, but with less accuracy on the farther ranges if you are not running vgrip (which has a tradeoff of wield time) and overall less DPS.
## Changelog
:cl:
balance: Reduces the Double Barrel's firerate, removes gyro as an attachment, reduce base recoil.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/shotguns.dm
+++ b/code/modules/projectiles/guns/shotguns.dm
@@ -204,7 +204,6 @@
 		/obj/item/attachable/bayonetknife/som,
 		/obj/item/attachable/verticalgrip,
 		/obj/item/attachable/reddot,
-		/obj/item/attachable/gyro,
 		/obj/item/attachable/flashlight,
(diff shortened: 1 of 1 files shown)


maikilangiolo:
> with good keybinds

You mean just the default Z and X?

Barnet2:
> > with good keybinds
> 
> You mean just the default Z and X?

Bind em together, you can also bind unloading the gun and unwielding on the same key. Good keybinds does not equal hard to setup.

PR #13591, opened 2023-07-26
Custom squads

MacBlaze1:
Simply for TiviPlus to take over the code


LoganKele:
what is this if you can answer? Does this mean that the current squads, (Alpha, Bravo, Charlie & Delta) are being removed potentially?

ivanmixo:
What even is this? Just give them the branch the code is on

PR #13594, opened 2023-07-26
Normalises stagger/slowdown recovery back to previous values

ivanmixo:
## About The Pull Request

Going through each instance of where stagger happens and adjusting it is very unreasonable, so I opted to just make them `mob/living` vars that hold our recovery rates. Adjusted for old values with the multiple proc being called multiple times.
## Why It's Good For The Game

Staggers and slows being multiple seconds on guns that previously dealt less than a second of either isn't ideal.
## Changelog
:cl:
balance: Normalised stagger/slowdown recovery back to previous values
/:cl:


zee11zee:
will this effect the R-76 relying on stagger to keep the user alive? thus making them a bad choice?

Lumipharon:
I can't balance numbers if you make regen broken again

code/modules/mob/living/status_procs.dm
```diff
@@ -681,12 +681,12 @@
 /mob/living/proc/add_slowdown(amount, capped = 0)
 	if(HAS_TRAIT(src, TRAIT_SLOWDOWNIMMUNE))
 		return
-	adjust_slowdown(amount * STANDARD_SLOWDOWN_REGEN)
+	adjust_slowdown(amount * slowdown_recovery_rate)
```
ivanmixo, line 684:
```suggestion
	adjust_slowdown(amount * -slowdown_recovery_rate)
```

PR #13598, opened 2023-07-26
Make it so inspecting your wounds show bleeding, like examine. 

lbnesquik:
## About The Pull Request
It was weird how you could see some wounds only by using the external examine on yourself, instead of the one you have by default, with helping intent on yourself. This PR fixes it, at least for bleeding.

## Why It's Good For The Game
More consistent. Helps people not walk around bleeding like idiots. 

## Changelog

:cl:
qol: made bleeding show up on help intent self inspect
/:cl:


QualityVan (changes requested):
I'd rather not have a full line for each bleeding limb when each limb already has their own line. Either tie it into the main limb message or have bleeding be one line, like 'You are bleeding from your left arm, head, and chest'

lbnesquik:
> I'd rather not have a full line for each bleeding limb when each limb already has their own line. Either tie it into the main limb message or have bleeding be one line, like 'You are bleeding from your left arm, head, and chest'

Here you go boss. 

image

PR #13602, opened 2023-07-26
Fix pinpointer not turning itself off after target is destroyed

forestbbbbbbb:
## About The Pull Request

title

## Changelog

:cl:
fix: fixed pinpointer not turning itself off after target is destroyed
/:cl:

### DIFF
--- a/code/game/objects/items/pinpointer.dm
+++ b/code/game/objects/items/pinpointer.dm
@@ -69,7 +69,7 @@
 
 /obj/item/pinpointer/proc/activate(mob/living/user)
 	set_target(user)
-	if(!target)
+	if(QDELETED(target))
 		return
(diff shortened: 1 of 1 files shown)


code/game/objects/items/pinpointer.dm
```diff
@@ -85,9 +85,9 @@
 
 
 /obj/item/pinpointer/process()
-	if(!target)
-		icon_state = "pinonnull"
+	if(QDELETED(target))
 		active = FALSE
+		icon_state = "pinonnull"
 		return
```
TiviPlus, line 91: Should this also PROCESS KILL

Lumipharon:
Icon state should probs be handed in update_icon_state but eh.

PR #13603, opened 2023-07-26
quick_equip prioritizes prefs

Xander3359:
## About The Pull Request
Makes your client preferences matter in the case of storing an item using quick equip.
## Why It's Good For The Game
If you have a keybind to put an item in a specific storage, it will now prioritize your keybind rather than dump anything it can into a valid open storage.
## Changelog
:cl:
qol: Quick equipping into your belt now puts things into your belt and not your backpack.
/:cl:


Lumipharon:
No. This makes it so there is no way to quick equip to open storage, which is one thing that you do explicitly have control over.

PR #13604, opened 2023-07-27
Stagger and stun improvement and refactor

Lumipharon:
## About The Pull Request
Stagger is now a status effect instead of a snowflake var/life tick thing.

What does this mean?
Currently stagger is reduced every life tick (approx 2 seconds), which means 0.1 stagger and 1 stagger (or up to 3 before the fix) was functionally identical in duration AND that duration could vary between 0.1 to 2 seconds, depending on the timing of when you applied it.

As a status effect, it will now actually have an accurate duration (SSfastprocessing, so accurate to 0.2 seconds). so stagger duration should now actually be accurate instead of RNG.

Also unfucked various stun application stun. No more funny *20 multipliers, now everything uses the actual duration (i.e. SECONDS define) for consistancy and clarity.

Some various misc stagger walance changes I've already forgotten to list, but including repeater, slugs, railgun, flashbangs, huggers and wraith blink, all being reduced somewhat.
## Why It's Good For The Game
Less snowflakey code, vastly more accurate and sensible stagger performance.
## Changelog
:cl:
fix: Stagger will now have an accurate duration instead of relying on life ticks
code: all staggerstuns now consistantly use SECONDS defines
refactor: refactored stagger code
balance: Reduced the stagger or some guns and effects due to stagger working correctly
/:cl:


Lumipharon:
Updated to make all the stun procs consistant and relatively sensible.

code/datums/status_effects/debuffs.dm
```diff
@@ -16,6 +16,21 @@
...
+//STAGGERED
+/datum/status_effect/incapacitating/stagger
+	id = "paralyzed"
+
+/datum/status_effect/incapacitating/stagger/on_apply()
+	. = ..()
+	if(!.)
+		return
+	ADD_TRAIT(owner, TRAIT_STAGGERED, TRAIT_STATUS_EFFECT(id))
+	SEND_SIGNAL(owner, COMSIG_LIVING_STAGGER_APPLIED, duration)
```
TiviPlus, line 28: Cant you use the signal from the trait being applied/removed instead?

TiviPlus (commented):
rest lgtm

code/datums/status_effects/debuffs.dm
Lumipharon: Replaced it with just modifying the scatter mod var, so cleaner code/less signals.

Devinil:
I just saw a carrier in eternal stun from a shotgun user. Real improvement there huh.

Lumipharon:
Working as intended

TiviPlus (approved):

PR #13622, opened 2023-07-28
Self harm toggled off by default, default keybind for self harm toggle

Barnet2:
## About The Pull Request
See title.
## Why It's Good For The Game
More often than not you want self harm to be off rather than on (looking at melee users), adds an easy keybind if you want it on though. If you already have keybinds set, this won't change it.
## Changelog
:cl:
qol: Adds default keybind for toggling self harms, turns self harm off by default.
/:cl:


lbnesquik:
Keybind is 0, to save someone a click.

Lumipharon:
just TM'ing in case there is some valid reason(s) to self harm proc yourself that has been overlooked.

PR #13628, opened 2023-07-29
Allows the AI to aim linked artillery via minimap

lbnesquik:
## About The Pull Request
You know how the droppods have a nifty little minimap thing to fire shots? Well, this PR (badly) ports this for the AI linked guns.

Essentially, the AI, once someone has purchased and applied the linking module, can click an action to open a minimap of the ground. There, they just click on the place they want to aim the gun, and it sets it to aim there. It does not replace the current aiming method, and it does not bypass the accuracy targets.

I also increased the cost of the targeting modules heavily, as this ought to be a powerful option, even if the price is now quite heavy (RO won't like them). 

If there persists balance concerns, I could see it inducing more random spread than normal. Something like a small <s>mark-</s>multiplier to the srpead

I might try and add a little cursor shape when it is used, change the notice given by the artillery piece to inform it is blind firing, and add railgun control too. Maybe even a shell request selection, like how it looks like when you choose what type of resin weed you want to plant.

I could see it being extended to CIC too, which would give them more influence, and more 'Good Hits' ability. As long as we keep it firing blind. 

## Why It's Good For The Game
It makes the AI aimed guns work slightly differently, which is more interesting and fun in my opinion. It also just makes sense that a computer could designate coordinates without having to see it within blurry, jolting headset cameras. 

## Changelog

:cl:
add: Gave the AI the ability to aim artillery via minimap
balance: Increased price of the AI artillery targeting module

/:cl:


AlexCrimson:
Its real. The AI will take all our jobs. Marines will need to go on strike at this rate.

KM-Catman:
So it's a small QOL change that's balanced out by making the module it's for effectively inaccessible? When the fuck is an engi going to spend 50 points on a module that doesn't help them? 800 Req points is equivalent to almost 7 Valks.

lbnesquik:
> So it's a small QOL change that's balanced out by making the module it's for effectively inaccessible? When the fuck is an engi going to spend 50 points on a module that doesn't help them? 800 Req points is equivalent to almost 7 Valks.

The price could be lower, but if you buy that module, you are going to be pretty glued up to the howitzer. Also, there's a fair few times where req gets +2500 points, so I can see them spending that amount of dosh easily.

KM-Catman:
> > So it's a small QOL change that's balanced out by making the module it's for effectively inaccessible? When the fuck is an engi going to spend 50 points on a module that doesn't help them? 800 Req points is equivalent to almost 7 Valks.
> 
> The price could be lower, but if you buy that module, you are going to be pretty glued up to the howitzer. Also, there's a fair few times where req gets +2500 points, so I can see them spending that amount of dosh easily.

Doesn't matter if they CAN buy it, it's that you're forcing an absurd nerf for what should just simply be a QOL change. It's like if they cut the number of drop pods down to 4 when they finally let you target them with the minimap.

lbnesquik:
> > > So it's a small QOL change that's balanced out by making the module it's for effectively inaccessible? When the fuck is an engi going to spend 50 points on a module that doesn't help them? 800 Req points is equivalent to almost 7 Valks.
> > 
> > 
> > The price could be lower, but if you buy that module, you are going to be pretty glued up to the howitzer. Also, there's a fair few times where req gets +2500 points, so I can see them spending that amount of dosh easily.
> 
> Doesn't matter if they CAN buy it, it's that you're forcing an absurd nerf for what should just simply be a QOL change. It's like if they cut the number of drop pods down to 4 when they finally let you target them with the minimap.

I reduced the price, but it is still overall higher, given this absolutely does improve the AI ability to aim artillery massively.

TiviPlus:
This seems like creep

lbnesquik:
> This seems like creep

Sure makes that module more useful, and the AI if it is used. That's why I increased the price quite a bit. It's still not as strong as the old droids were, though.

chizzy376:
> 800 Req points is equivalent to almost 7 Valks.

10 valks with 50 pts leftover.

lbnesquik:
> > 800 Req points is equivalent to almost 7 Valks.
> 
> 10 valks with 50 pts leftover.

No longer the price.

lbnesquik:
Git bot don't stale me no more

lbnesquik:
I'll come back to this later in another shape.

PR #13632, opened 2023-07-29
Allow xenos to slow build once they run out of quickbuild points

lbnesquik:
## About The Pull Request
The title is pretty clear in explaining what this does.

Slight note regarding the code, I was not able to test it, because I can't manage to launch a game with a proper hive alone, without it ending immediately thanks to no marines. However, given it is just 1 argument, I think this will be fine. 

## Why It's Good For The Game
This allows xenos to continue to fortify once they've spent their quick build points. 
This only happens with a lot of builders, or when the marines open shutters late. In the first case, the xenos invested in the drone caste, and I think they ought to get to build some more, instead of going for more combat. In the second, marines shouldn't have waited until 12:25 to land and get out, and that's on them. 

There is the case of some hivelord making some giant puddle of walls somewhere, but this still takes all of the QB points away from the **ENTIRE** hive, so I think the abuse potential is minor. 

## Changelog

:cl:
balance: Gave xenos the ability to slow build once they run out of QB points
/:cl:


Barnet2:
1. You can test in any round by going to the server tab at the top right and clicking start round once, waiting like 5 seconds, then clicking it again and accepting to start the round disabling mode check.

2. The shared quickbuild PR was accepted specifically because it didn't have manual build preshutters, while Grayson's version did. I highly doubt this will get a merge but hey, I would love to get proven wrong because I too hate QB running out (though I am biased in this, in all reality balance would be better without this)

lbnesquik:
> 1. You can test in any round by going to the server tab at the top right and clicking start round once, waiting like 5 seconds, then clicking it again and accepting to start the round disabling mode check.

I know how testing work, I just did not manage to not have it end early while having a hive, as written. 

>     2. The shared quickbuild PR was accepted specifically because it didn't have manual build preshutters, while Grayson's version did. I highly doubt this will get a merge but hey, I would love to get proven wrong because I too hate QB running out (though I am biased in this, in all reality balance would be better without this)

I think there was more to it, Grayson's version also reduced the zone for QB, which meant it wasn't all that useful for building outside caves, which is where a good chunk of the work had to be done. Didn't quite reduce the workload for the poor drone chumps. It was also more complicated and harder to understand, I think.

AlexCrimson:
Id love this, but i doubt its needed at this point. Right now most maps have enough points to maze important, high-traffic areas. That should be fine. At most, some maps maybe could use some more points. Or it would be nice to at least have Sticky Resin removed as a quickbuild option, so we can instead slow build with it(and the added benefit of xenos not wasting QB points spamming it).

If you have nothing to do as a Drone, go spam sticky resin nodes outside the Marine LZ to make it annoying for them to advance without manually clearing every single one. If you are a Hivelord, go plant Acid Wells throughout your mazes.

Lumipharon:
No, quick build combined with slow build is not going to happen, this has been discussed multiple times.

Having slow build after means going back to the optimal strategy being to build nonstop for the entire preshutter period.

Quick build was ONLY added to change that.

PR #13637, opened 2023-07-30
Mounted gun scope range increase 

maikilangiolo:
## About The Pull Request
Brings the mounted guns' "miniscope" to actual miniscope levels (3 -> 5) and increases by +1 the tadpole mounted/tat scope to keep a significant difference between the two (7 -> 8)

## Why It's Good For The Game

Lumi approved (at least the miniscope one). The issue is, the miniscope and the mounted gun scopes were the same and intended to be: 3 tile zoom range. However, the miniscope got buffed to 5, but the mounted variant got left behind cause no one cares. This means that the MG27, a "lower tier" mounted gun, can outrange the higher tier mounted guns (HSG and MG08 notably) because it can mount the 5-range miniscope instead of using a 3-range legacy scope. Nevermind the other handheld guns that don't have the disadvantage of being fixed in place and being able to be destroyed.

The +1 range to the long scope is so that the guns using it still have an more pronounced advantage over those with the now-buffed mini (from 3 and 7, to 5 and 8). It is still shorter than the boiler and rail scope, as intended (which are 10 tiles).
## Changelog
:cl:
balance: mounted gun miniscope has actual miniscope zoom now
balance: increased mounted gun rail scope zoom by 1
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -748,7 +748,7 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 	desc = "An unremovable set of long range ironsights for an HMG-08 machinegun."
 	icon_state = "sniperscope_invisible"
 	zoom_viewsize = 0
-	zoom_tile_offset = 3
+	zoom_tile_offset = 5
 
 /obj/item/attachable/scope/unremovable/mmg
 	name = "MG-27 rail scope"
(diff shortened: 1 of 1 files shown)


Lumipharon (changes requested):
Why are you changing mounted guns?

maikilangiolo:
> Why are you changing tad mounted guns?

The reason is listed twice in both description and the "why" part. Is the change to tad guns the only one that has to be undone? Since they're locked to tadpole, having a bit more scope compared to the buffed mini would be beneficial, but you're the boss anyway

Al-1ce:
Tad being able to offscreen is normally bad, but I think the stuck angle it has on the Tadpole is a fair compensation all things considered.

Al-1ce:
Tad being able to offscreen is normally bad, but I think the stuck angle it has on the Tadpole is a fair compensation.

Al-1ce:
Being able to offscreen is normally bad, _but_ I think the stuck angle on the Tadpole is a more than fair compensation/tradeoff.

Lumipharon:
Don't change the other ones please.

Lumipharon (approved):

PR #13638, opened 2023-07-30
Gives SL extra access

breadhunt:
## About The Pull Request

<s>Gives SL access to engi, medic and other squad comms. Gives SL ID access to bridge and prep so they can coordinate in person with commanders and their own squad.</s>
Gives SL bridge access.

## Why It's Good For The Game

Lets SL meet with their superiors to discuss plans and complain about marines. Lets the command team coordinate things better in general, with TTS.

## Changelog
:cl:
add: Squad leaders now have access to CIC
/:cl:


novaepee:
Giving SL req access is too strong of a buff. Good FC players already show a lot of constraint to not buy stuff from req for themselves.

We will see what maintainers have to say about all access.

breadhunt:
COMPLETELY forgot about req access, yeah i think you're right there

coldironwarrior:
ability to swipe for red alert as SL will be nice

DeltaFire15:
This seems overkill. SLs exist to lead their squad, not to be FC with less responsibility.

Besides the fact this is completely unneeded of a buff (as opposed to ""qol"" the changelog is set to), the access changes are also applied on full crew, which is even more unneccessary than this PR already was.

Lumipharon:
I don't agree with this tbh.

QualityVan:
Bridge access seems sensible, I'm leery about all radio, and prep is entirely unnecessary.

breadhunt:
Changed to just be bridge access.

PR #13646, opened 2023-07-30
Fixes some motion sensor (and probably other) bugs

ivanmixo:
## About The Pull Request

These lists can have nulls in them (due to harddels or otherwise) and it's not wise to feed nulls to this because it causes runtimes. This might fix motion sensor bricking at random and it fixes motion sensor not working if you get inside a locker.
## Why It's Good For The Game

Bug bad.
## Changelog
:cl:
fix: Motion sensor should keep working when inside a locker now
code: cheap_get_human/xenos_near shouldn't return nulls anymore
/:cl:

### DIFF
--- a/code/__HELPERS/ai.dm
+++ b/code/__HELPERS/ai.dm
@@ -5,6 +5,8 @@
 /proc/cheap_get_humans_near(atom/movable/source, distance)
 	. = list()
 	var/turf/source_turf = get_turf(source)
+	if(!source_turf)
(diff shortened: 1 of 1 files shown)


ivanmixo:
love runtimes i cant reproduce on local just great

TiviPlus:
how the hell is it harddelling with a ref in this list?

Lumipharon (approved):

PR #13655, opened 2023-07-31
removes loose xeno structure pinpointers on canterbury and bigbury

coldironwarrior:
## About The Pull Request

removes the loose xeno structure pinpointers on the canterbury and bigbury

## Why It's Good For The Game

i think they're pretty much a total noob trap. silos and all can't be destroyed on crash. 
whether spawncamping silos is behavior maints want to allow or not is a part of the considerations to be made for this and if it's decided that's intentional, then i get that.

only possibly used otherwise for the hivemind on crash that places their core not in the zones blocked off by fog.

## Changelog

:cl:
del: removed loose pinpointers on the crash ships
balance: rebalanced something
/:cl:

### DIFF
--- a/_maps/shuttles/tgs_bigbury.dmm
+++ b/_maps/shuttles/tgs_bigbury.dmm
@@ -179,7 +179,6 @@
 /area/shuttle/canterbury)
 "aR" = (
 /obj/structure/table/reinforced,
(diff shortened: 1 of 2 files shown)


TiviPlus (approved):
these used to be how you found the disk

PR #13675, opened 2023-08-02
Adds a implant that dusts the user on death.

Xander3359:
## About The Pull Request
I was told the elite moth ERT was meant to have this but it was removed as deprecated code.
So I'm bringing this back and giving it to the admin-spawn ERT, with plans to allow admins to give any implant using VV edits later.
## Why It's Good For The Game
So that admins don't have to gib smite corpses in order to prevent marines from looting things they shouldn't have.
## Changelog
:cl:
add: Self-dusting implant that triggers on death for the elite moth ERT.
/:cl:


isotope436:
now that i think about it we should give this to the deathsquad ERT aswell

chizzy376:
> now that i think about it we should give this to the deathsquad ERT aswell

That would be cool, but the deathsquad desperately needs a loadout update as well. Rn all they have is their armor, everything else is kinda weak

yyzsong:
> > now that i think about it we should give this to the deathsquad ERT aswell
> 
> That would be cool, but the deathsquad desperately needs a loadout update as well. Rn all they have is their armor, everything else is kinda weak

dont dsquaddies still have the pulse rifles?

Derrick2Electricboogaloo:
> 

If someone can kill the DS, they should get that sweet reward, besides DS is super duper exceptionally rare.

isotope436:
> > 
> 
> If someone can kill the DS, they should get that sweet reward, besides DS is super duper exceptionally rare.

its admin spawn only its not ment to be killed and with the recent buff the only way you can kill it is via abusing the fact auto injector peri+ od's can kill them in 5 seconds flat.

lbnesquik:
Make them emote a scream

code/game/objects/items/implants/implant.dm
```diff
@@ -36,9 +36,7 @@
 
 
 /obj/item/implant/Destroy(force)
-	unimplant()
 	QDEL_NULL(activation_action)
-	part?.implants -= src
```
Lumipharon, line 41: what are you doing

FosterGrime-Heymynameslouie:
> Make them emote a scream

honestly, if you can pull that off against a Dsquad hats off to you you probably deserve the gear more than him

code/game/objects/items/implants/implant.dm
Xander3359: So it turns out, unembed_ourselves() will run when the mob gets deleted. So I'm just gonna let that proc handle the unimplanting and removing the implant.

code/game/objects/items/implants/implant.dm
```diff
@@ -87,6 +89,8 @@
 /obj/item/implant/proc/unimplant()
 	SHOULD_CALL_PARENT(TRUE)
 	forceMove(get_turf(implant_owner))
+	if(!implanted)
```
Lumipharon, line 92: why is this after the forcemove?

Lumipharon (changes requested):

code/game/objects/items/implants/implant.dm
Xander3359: Because if I put it before the forcemove it runtimes.

Lumipharon (approved):
This should presumably give a gib action button or something, since if the intention is to deny loot... you jsut crit them, loot them, then kill them.

PR #13685, opened 2023-08-03
Reqtorio screen background color change

novaepee:
## About The Pull Request

Before: 
image

After:
image

## Why It's Good For The Game

Reqtorio has some neat animation, but you cannot see the animation in play due to the dark background with dark objects, so I change the background.

## Changelog

:cl:
imageadd: Reqtorio screen background color change
/:cl:


Arctrooper8:
How does it actually look in game?

chizzy376:
Goon station computer coloring
This will make it look really depressing like an industrial setting from the 80's
Don't make the Req mains even more depressed, please

Also fix the screwdriver direction change to be clockwise why the hell is it counter-clockwise--

PR #13692, opened 2023-08-04
Xenos can no longer spam click miners to break them faster

Xander3359:
## About The Pull Request
Xenos can no longer left click spam a miner to break it in < half the required time.
## Why It's Good For The Game
I'd call this a bugfix, if miners are intended to be destroyed at the faster rate (of left click spam), then the amount of time it takes to slash it should be reduced.
## Changelog
:cl:
fix: Xenos can no longer spam left click miners to break them faster than normally.
/:cl:

### DIFF
--- a/code/game/objects/machinery/miner.dm
+++ b/code/game/objects/machinery/miner.dm
@@ -318,6 +318,8 @@
 		span_notice("We can't slash through [src]'s reinforced plating!"))
 		return
 	while(miner_status != MINER_DESTROYED)
+		if(X.do_actions)
+			return balloon_alert(X, "busy")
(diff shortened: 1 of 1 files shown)


Lumipharon:
The way the code is written heavily implies this is in fact a bug. Otherwise what is the point in requiring multiple channels if you can just do them all at the same time?

TooFewSecrets:
I honestly don't care if whoever coded this considers it to be intended behavior. Either make it one channel if current balance is fine, or force multiple separate channels. There's no reason to let the player do something objectively worse for no benefit.

Lumipharon:
if someone has a problem with how long this takes, they can walance it. But this is a fix. Also does this hapen with tad console still?

PR #13694, opened 2023-08-05
Progressive sprites for empty belts and pouches

Sanuki357:
## About The Pull Request

This adds a new sprites for belts and pouches when they become more empty.
The sprites progressively changes as they become emptier, the lids of each individual pouches gets flipped up.

The math behind the sprite changes supports both storage_slots and max_storage_space.
All you have to do when you are adding a new pouch, is making sure each sprites ends with "__numbers_", _0 being empty and increment until the full version with no underscore at the end of the name.

This also changes base sprite of some belts and pouches to improve a very bad sprite that would not look good with the new sprites, including the grenade pouch.

## Why It's Good For The Game

This gives the marines a visual cue of how empty their belts or pouches are.
By this, marines would no longer have to drag and drop the sprite of belts or pouches to marine's sprite only to check if they have something or not.
This would reduce the stress of having to look each container individually and make inventory management easier.
This would especially be helpful in the situation like looting a corpse of fellow marine. without having to check each container, but with just an glance at the sprite of belts and pouches.

## Changelog

:cl:
imageadd: New progressive sprites for empty belts and pouches which responses to how empty they are
imageadd: Changed the grenade pouch sprite
/:cl:


code/game/objects/items/storage/pouch.dm
```diff
@@ -10,12 +10,15 @@
 	allow_drawing_method = TRUE
 	var/fill_type
 	var/fill_number = 0
+	///Defines how many versions of the empty sprite there are
+	var/sprite_slots = null
```
Lumipharon, line 14: This should probably be at the storage level, not pouch, especially since you already have said you plan to do this for other storage types.

code/game/objects/items/storage/pouch.dm
```diff
@@ -10,12 +10,15 @@
...
 
 /obj/item/storage/pouch/Initialize(mapload)
 	. = ..()
 	if(fill_number && fill_type)
 		for(var/i in 1 to fill_number)
 			new fill_type(src)
+	update_icon_state()
```
Lumipharon, line 21: you should never be calling update_icon_state directly, just update_icon

code/game/objects/items/storage/pouch.dm
```diff
@@ -35,6 +38,23 @@
 	..()
 	return user.equip_to_appropriate_slot(src)
 
+/obj/item/storage/pouch/update_icon_state()
+	if(!sprite_slots)
+		return
```
Lumipharon, line 43: this should be setting it to initial probs

code/game/objects/items/storage/pouch.dm
```diff
@@ -35,6 +38,23 @@
...
+
+	var/total_weight = 0
+	for(var/obj/item/i in contents)
+		total_weight += i.w_class
```
Lumipharon, line 47: this is only used if the storage doesn't use slots, so it should be in the if statement for that situation, not before. Otherwise its just getting calculated for no reason.

Lumipharon (changes requested):

code/game/objects/items/storage/pouch.dm
Sanuki357: Gotcha. It's update_icon() now.

code/game/objects/items/storage/pouch.dm
Sanuki357: Gotcha. There seems to be no existing update_icon_state() so it is in there now.
I was not sure where to put it and all of the codes in storage.dm are all procs and verbs so I pasted them in the last line. Line 869
image

code/game/objects/items/storage/pouch.dm
Sanuki357: I thought this would avoid needing to call initial() every time it is opened, but I'll test stuff with initial().

code/game/objects/items/storage/pouch.dm
Sanuki357: Damn. This should've been obvious to me when I wrote it...
`total_weight += i.w_class` now lives where it would get called when w_class is relevant.

code/game/objects/items/storage/pouch.dm
Sanuki357: I added a check for if a storage is completely empty, and set the icon_state to have "_e" instead of "_0", since _0 representing a bag being empty is a too odd of a convention and I want to avoid confusion from whoever touches pouches and belts.

Lumipharon (approved):

PR #13708, opened 2023-08-06
Fixes AGLS Incendiary ammo not making fire.

Xander3359:
## About The Pull Request
This ammo has a drop_flame() proc which seems to want to drop fire on tiles it hits, but it is never called anywhere.
I'm just gonna copy the flamer code for this so that it actually flames things as intended.
## Why It's Good For The Game
Bugfix
## Changelog
:cl:
fix: AGLS Incendiary ammo drops fire now.
/:cl:


Lumipharon:
lel

ivanmixo (approved):

PR #13721, opened 2023-08-08
change broken computer warning to be a bit less misleading.

Xander3359:
## About The Pull Request
"It is broken" instead of "It is broken and needs to be rebuilt."
## Why It's Good For The Game
You can't rebuild tad computer, this is misleading.
## Changelog
:cl:
spellcheck: change broken computer warning to be a bit less misleading.
/:cl:

### DIFF
--- a/code/game/objects/machinery/computer/computer.dm
+++ b/code/game/objects/machinery/computer/computer.dm
@@ -35,7 +35,7 @@
 		. += span_warning("It is currently disabled, and can be fixed with a welder.")
 
 	if(machine_stat & BROKEN)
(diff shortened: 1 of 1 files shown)


EvilMelonTheSecond:
Why not clarify further and specify it cannot get repaired with something like "It is broken for good" or "It is broken beyond repair"

Xander3359:
> Why not clarify further and specify it cannot get repaired with something like "It is broken for good" or "It is broken beyond repair"

That was my initial idea but seeing how tad is literally the only un-repairable computer I was told to do it like this.

ivanmixo (approved):

PR #13740, opened 2023-08-09
Dynamic sprite revert for SOM and ICC pouches

Sanuki357:
## About The Pull Request

This will fix the SOM and ICC pouches going invisible when they are empty.
## Why It's Good For The Game

SOM operatives would appreciate a stealth camouflage but not if it is only applied to their pouches.
## Changelog
:cl:
fix: Fixed SOM and ICC pouches going invisible when they are empty
/:cl:


ivanmixo (approved):
Make a PR addressing them appearing empty even when full as well (when you dispense from loadout vendor for example)

PR #13741, opened 2023-08-09
Hunter illusions now work as intended

Zar000:
## About The Pull Request

Bugfix good and very much overdue.

Mirage illusions now always follow hunter until they find a marine within range ( 5 tiles as of now ).
When that happens, they will stick to them and attack until they die.
## Why It's Good For The Game
bugfix good
## Changelog
:cl:
fix: hunter mirage now works as intended
/:cl:


shelteredfreak:
FINALLY. Now all we need is to get Silence back and all that's good in the world will have returned.

ivanmixo (approved):

PR #13147, opened 2023-06-03
Marine light medical belt

tanker4390:
## About The Pull Request
Light medical belt, available infinitely in clothing vendors. Can hold 15 slots of various medical items.

216548462-38bdd814-f702-40fb-81dd-ea7a593d8691
## Why It's Good For The Game
Gives marines the option to fumble around with more medical supplies. I thought the lack of a specialized medical belt among its pouch and other counterpart was odd.
## Changelog
:cl: Tanker
add: Added a lightweight medical belt for marines to use.

PR #13744, opened 2023-08-09
Replaces pill packets with pill bottles and limits oxycodone availability

ivanmixo:
## About The Pull Request

- Replaced pill packets with pill bottles in regular vendor, removed from nanomed
- Also replaced the packets in the default first aid pouch from the quick equip vendor
- Removed oxycodone injectors/bottles from nanomed and marinemed
- Corpsmen can get oxycodone pill bottles for 8 points each, each pill is 7.5u
## Why It's Good For The Game

As it stands right now packets are just inferior to pill bottles in every way. This results in players who are less experienced and don't know you can get pill bottles relatively easily being at a disadvantage for simply following what the game recommends they use, which is not ideal in my opinion. Having to break into medical every round to get these better bottles is also not ideal, so everyone having equal access should resolve both of these issues.

Oxycodone is a tier 2 chem, in other words just better tramadol. I have no idea why it's in the vendors and it only encourages people to ditch tram. It being limited like this could probably also be used as justification for removing meraderm purging oxy. 
## Changelog
:cl:
balance: Replaced pill packets with pill bottles in regular vendor, removed from nanomed
balance: Also replaced the packets in the default first aid pouch from the quick equip vendor
balance: Removed oxycodone injectors/bottles from nanomed and marinemed
balance: Corpsmen can get oxycodone pill bottles for 8 points each, each pill is 7.5u
/:cl:

### DIFF
--- a/code/__DEFINES/loadout.dm
+++ b/code/__DEFINES/loadout.dm
@@ -99,6 +99,7 @@ GLOBAL_LIST_INIT(medic_gear_listed_products, list(
 		/obj/effect/vendor_bundle/medic = list(CAT_ESS, "Essential Medic Set", 0, "white"),
 		/obj/item/storage/pill_bottle/meralyne = list(CAT_MEDSUP, "Meralyne pills", 16, "orange"),
 		/obj/item/storage/pill_bottle/dermaline = list(CAT_MEDSUP, "Dermaline pills", 16, "orange"),
+		/obj/item/storage/pill_bottle/oxycodone = list(CAT_MEDSUP, "Oxycodone pills", 8, "orange"),
 		/obj/item/storage/syringe_case/meralyne = list(CAT_MEDSUP, "syringe Case (120u Meralyne)", 16, "black"),
 		/obj/item/reagent_containers/hypospray/advanced/meralyne = list(CAT_MEDSUP, "hypospray (60u Meralyne)", 8, "black"), //half the units of the mera case half the price
 		/obj/item/storage/syringe_case/dermaline = list(CAT_MEDSUP, "syringe Case (120u Dermaline)", 16, "black"),
--- a/code/game/objects/items/reagent_containers/pill.dm
+++ b/code/game/objects/items/reagent_containers/pill.dm
@@ -159,6 +159,11 @@
 	list_reagents = list(/datum/reagent/medicine/tramadol = 15)
 	pill_id = 7
 
+/obj/item/reagent_containers/pill/oxycodone
+	pill_desc = "A oxycodone pill. Acts like a high grade painkiller."
+	list_reagents = list(/datum/reagent/medicine/oxycodone = 7.5)
+	pill_id = 7
+
 /obj/item/reagent_containers/pill/isotonic
 	pill_desc = "A pill with an isotonic solution inside. Used to stimulate blood regeneration."
 	list_reagents = list(/datum/reagent/medicine/saline_glucose = 15)
--- a/code/game/objects/items/storage/firstaid.dm
+++ b/code/game/objects/items/storage/firstaid.dm
@@ -418,6 +418,15 @@
 	greyscale_colors = "#FD5964#ffffff"
 	description_overlay = "Me"
 
+/obj/item/storage/pill_bottle/oxycodone
+	name = "oxycodone pill bottle"
+	desc = "Contains pills that act as a high grade painkiller."
+	icon_state = "pill_canistercomplete"
+	pill_type_to_fill = /obj/item/reagent_containers/pill/oxycodone
+	greyscale_config = /datum/greyscale_config/pillbottleround
+	greyscale_colors = "#8a8686#ffffff"
+	description_overlay = "Ox"
+
 /obj/item/storage/pill_bottle/dexalin
 	name = "dexalin pill bottle"
 	desc = "Contains pills that heal oxygen damage. They can suppress bloodloss symptoms as well."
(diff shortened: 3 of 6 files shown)


chizzy376:
> in other words just better tramadol


Oxycodone: Gives patients very heavy pain **(-50)** reduction and reduces Long Pain whilst it is in their system at the cost of 0.2 TOX. **Is purged by Meralyne and Dermaline.**
Additionally, on injection heals 20 stamina damage. However, once the oxycodone runs out, further instant stamina heals from injection or consumption are locked out for 60 seconds.


Tramadol: A simple, yet effective painkiller **(-50).** Very strong and second to Oxycodone in effectiveness. Additionally, above 20u inside the body, **gives slightly higher pain resistance (-15) at the cost of 0.25 TOX**. **Isn't purged by Meralyne and Dermaline.**

ivanmixo:
> > in other words just better tramadol
> 
> Oxycodone: Gives patients very heavy pain **(-50)** reduction and reduces Long Pain whilst it is in their system at the cost of 0.2 TOX. **Is purged by Meralyne and Dermaline.** Additionally, on injection heals 20 stamina damage. However, once the oxycodone runs out, further instant stamina heals from injection or consumption are locked out for 60 seconds.
> 
> Tramadol: A simple, yet effective painkiller **(-50).** Very strong and second to Oxycodone in effectiveness. Additionally, above 20u inside the body, **gives slightly higher pain resistance (-15) at the cost of 0.25 TOX**. **Isn't purged by Meralyne and Dermaline.**

https://github.com/tgstation/TerraGov-Marine-Corps/pull/11424

chizzy376:
> #11424

Updated numbers from code.
### OXYCODONE:
PROS:
* -20 stamina loss on inject (Has cooldown etc.)
* super heavy pain reduction
* helps from crit

CONS:
* overdose turns you into a vegetable
* quickly metabolizes
* is purged by meraderm
* toxin

### TRAMADOL:
PROS:
* very heavy pain reduction + light pain reduction on 20u+ (but gives toxin)
* od is harmless
* better to dose on since slower metabolization

CONS:
* can't help from crit (But just carry synap for that)
### RESULTS
The pros of oxycodone are heavily outweighed by its downsides. If you seek the upsides with no downsides, just carry inaprovaline/synaptizine/have valk.

Mrrpip:
Hm. If you're removing pill packets from most sources, this item should be taken into account. As it can hold packets, but not bottles.

https://github.com/tgstation/TerraGov-Marine-Corps/pull/13147/files

Barnet2:
It's good to commit to something either way, if this doesn't go through I'll make pill bottles be dispensed by medical staff. Current rush into chem to hack vendor is wack.

Strayed-Raven:
I am sorry but I cannot see anyone actually ditching tramadol for straight Oxycodone. So is this the start of the fabled medical chem rework finally?

imsxz:
you dont have to ditch tramadol, you're supposed to stack them lol.

Lumipharon:
Tram metabolises at half the speed of most chems, while oxy metabolises faster than most chems.

So basically tram is good for premedding since its easy to perma have tram, while oxy you stack on top as needed, but you can't perma run it unless you're packing a fuckload.

Sirsnek15:
As someone who is a frequent, experienced player, but rarely ever uses limited equipment, pill packets are pretty suitable and don't need a storage buff, and limiting oxy is ermmm not very cool or thought out. I only see that making sense if the default injector pouch for squaddies still has the oxycodone pen, but even then its a bit rough.

As someone who sticks to packets I really only run out at around the 45 minute mark, and if I find a perma with the same packets that have more pills left, then I can resupply that way too. That brings me to my recommendation: rather than doing any of these things, why not allow pills to be put back into packets? I know thats not really how packets work but that'd be an overall better solution to your problem with them, though I personally don't think there is one.

QualityVan:
A single BKTD packet lasts 20 minutes of constant use, tram twice that, and para two and a half hours. I'm not at all convinced that's insufficient. If the issue is that there's no reason to not just spend prep to upgrade gear from what you're issued to something strictly better, I'd rather add a reason than double base medical storage. Putting a skilllock or access delay on bottles is the most obvious option offhand.

Oxy changes seem sensible, although it's reminding me that I sort of want to add a chem point buy option for PFCs rather than the current first come first serve marinemeds.

ivanmixo:
20 minutes is the best case scenario, assuming you only take one pill at a time and miss out on the extra healing and your meds absolutely never get purged.

QualityVan:
Simultaneous pills doesn't matter because it ticks down at the same rate regardless. You could take the whole packet at once and it'd have the same total duration. Purging does, of course, but that's the point of purging.

ivanmixo:
It does. You get extra healing above 20u

Lumipharon:
I unno why you're saying having less pills stops you from eating 2 pills... it still lasts the same duration as eating one at a time, and you still get the healing bonus.

Joe13413:
i think that oxycodone is in a relatively good place right now, most players just carry a few autoinjectors of it in case of emergency. It's a good "OH SHIT" autoinjector for emergencies and i don't think it needs to be limited, seeing as most people can only carry autoinjectors of it without using vali, or begging people to do chem for you.

Joe13413:
and also, i never really tend to run out of pills in my packets. The amount of medicine you get is fine as-is.

Jackal-boop:
Marines don't need more Healy chems on their person

ivanmixo:
luv me status quo

PR #13745, opened 2023-08-09
Engineer and corpsman slots actually scale properly now

ivanmixo:
## About The Pull Request

Makes them actually need 5 marines per, so there won't be 14 corpsmen and engi slots on 30 marines.
## Why It's Good For The Game

This is the intended way I'm pretty sure when https://github.com/tgstation/TerraGov-Marine-Corps/pull/13050 was made.
## Changelog
:cl:
fix: Engineer and corpsman slots actually scale properly now
/:cl:

### DIFF
--- a/code/datums/gamemodes/_game_mode.dm
+++ b/code/datums/gamemodes/_game_mode.dm
@@ -14,6 +14,7 @@ GLOBAL_VAR(common_report) //Contains common part of roundend report
 	var/round_finished
 	var/list/round_end_states = list()
 	var/list/valid_job_types = list()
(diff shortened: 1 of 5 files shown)


Yttriums:
Gave this thought elsewhere but I'll reiterate it here ~ given the criticality of these roles I feel there should be a floor of 3-5 slots, with scaling from there as the change suggests. I feel like the scaling will negatively affect lower population rounds, especially when newer players occupy one of a few contested spots. It will also pressure engineer players into very narrow, meta builds and away from enjoying the fun variety in the role.

Lumipharon (approved):

PR #13748, opened 2023-08-10
Adds chemistry request lines to (most) shipmaps

Barnet2:
## About The Pull Request
Read title.
Here is what chem looks like now on the shipmaps:
image
image
image
image
## Why It's Good For The Game
Since there are still req lines (on PoS, for example), if we ever need chemical request lines they are there; smartfridge only is sort of sad.
## Changelog
:cl:
add: Added chemistry request lines to most shipmaps.
/:cl:

### DIFF
--- a/_maps/map_files/Minerva/TGS_Minerva.dmm
+++ b/_maps/map_files/Minerva/TGS_Minerva.dmm
@@ -5093,6 +5093,9 @@
 /area/mainship/command/corporateliaison)
 "oM" = (
 /obj/machinery/light/mainship,
+/obj/structure/bed/chair/office/dark{
+	dir = 4
+	},
 /turf/open/floor/mainship/sterile/purple/corner{
 	dir = 4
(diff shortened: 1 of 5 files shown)


Derrick2Electricboogaloo:
I can not wait for these to not be used, since MDs still do not make chems for people.

iforgotmeword:
Yum yum more metal to the fob gods

Strayed-Raven:
Ah I see. instead of breaking windows we just have to break windoors now.

ThelastArchiver:
the reason usualy mds dont make chems for marines is because they already always break into chemistry on their own

RipGrayson (approved):

Derrick2Electricboogaloo:
> the reason usualy mds dont make chems for marines is because they already always break into chemistry on their own

Literally not even the sole reason. Most MDs do not want to, or do not know how to.

PR #13749, opened 2023-08-10
Reduces MD slots to four

Barnet2:
## About The Pull Request
Read title.

## Why It's Good For The Game
Having 6 MD slots is a bit silly considering 3 is the **maximum** amount of doctors you would want in the Medbay (including CMO); anymore and it gets crowded and you have to fight over patients more than you already would have to.

## Changelog

:cl:
balance: Medical Doctor slots reduced to 2 from 6.
/:cl:

### DIFF
--- a/code/datums/gamemodes/nuclear_war.dm
+++ b/code/datums/gamemodes/nuclear_war.dm
@@ -13,7 +13,7 @@
 		/datum/job/terragov/engineering/tech = 2,
 		/datum/job/terragov/requisitions/officer = 1,
 		/datum/job/terragov/medical/professor = 1,
-		/datum/job/terragov/medical/medicalofficer = 6,
+		/datum/job/terragov/medical/medicalofficer = 4,
 		/datum/job/terragov/medical/researcher = 2,
 		/datum/job/terragov/civilian/liaison = 1,
 		/datum/job/terragov/silicon/synthetic = 1,
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -724,7 +724,7 @@ Make sure that the doctors and nurses are doing their jobs and keeping the marin
 	title = MEDICAL_DOCTOR
 	comm_title = "MD"
 	paygrade = "RES"
-	total_positions = 6
+	total_positions = 2
 	supervisors = "the chief medical officer"
 	access = list(ACCESS_MARINE_MEDBAY, ACCESS_MARINE_CHEMISTRY)
 	minimal_access = list(ACCESS_MARINE_MEDBAY, ACCESS_MARINE_CHEMISTRY, ACCESS_MARINE_CARGO, ACCESS_MARINE_DROPSHIP)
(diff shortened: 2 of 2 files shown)


InterroLouis:
Isn't the point of those slots to allow for groundside doctor? Planetside surgery? Non-shipbound medical triage? Perhaps even medpole tadpole surgerydoctor. This would mean any groundside doctor would most likely end up leaving shipside medical understaffed.

Lumipharon:
I don't see the point of this. As above, doctors could be doing other things outside of medical, and there is no particular issue with having more doctors. If you join as an MD and you can see there is already 4 doctors, you know what you're getting into.

yyzsong:
> Isn't the point of those slots to allow for groundside doctor? Planetside surgery? Non-shipbound medical triage? Perhaps even medpole tadpole surgerydoctor. This would mean any groundside doctor would most likely end up leaving shipside medical understaffed.

rsr

ClosetedSkeletonGH:
> I don't see the point of this. As above, doctors could be doing other things outside of medical, and there is no particular issue with having more doctors. If you join as an MD and you can see there is already 4 doctors, you know what you're getting into.

as someone who loves to MD, these people dont, and it often ends up with too many cooks in the kitchen. Having two docs is a great time with enough stuff for both to do while never being overly overwhelmed, then a CMO joins, then another doc joins, and none of them want to go groundside. it makes medbay insufferable and at some points when im medbaying with another doc, Im just in fear that a CMO/another MD is gonna join and ruin the round experience by just snatching away patients that im already working on

lbnesquik:
3 slots IMO

Glcbitj:
Please, make it so it can increase based on the pop number or increase it to 3. Ideally there should be a separate role (field medic/surgeon) for people who want to do grounside surgery, but as it stands, MDs are the one who are most apt to do that.

Barnet2:
> I don't see the point of this. As above, doctors could be doing other things outside of medical, and there is no particular issue with having more doctors. If you join as an MD and you can see there is already 4 doctors, you know what you're getting into.

It's less of "you know what you are getting into if you join at 4" and more of people will do it regardless, as said above ruining the round for those in medbay forcing them to go groundside, which they did not want to do in the first place.

Naaanii:
Reducing to 4 and having 2 default available then scaling with pop size seems better.

ClosetedSkeletonGH:
> Reducing to 4 and having 2 default available then scaling with pop size seems better.

this does the opposite of fixing the problem unless its pop locked to like 55-70

Naaanii:
> > Reducing to 4 and having 2 default available then scaling with pop size seems better.
> 
> this does the opposite of fixing the problem unless its pop locked to like 55-70 marines like how mech has a ridiculously high pop

How? You get 2 slots roundstart plus CMO and as pop increases, you get 2 more potential slots scaling with pop.

ivanmixo:
Don't think there's much point to this at all.

ClosetedSkeletonGH:
Shipside roles should be joining to play shipside, not to be sent groundside/stand around in medbay til you either cryo or make someone else cryo out of boredom. there is rarely a need for anything more than two MD's, two MD's and a CMO is already overkill and unfun for the MD's, three MD's and a CMO has never been needed in the entirety of my time MD'ing.

ivanmixo (approved):

PR #13754, opened 2023-08-10
Prevents shutter open button from working before the shuttles can even take off

ivanmixo:
## About The Pull Request

Title. There were some AIs opening shutters before 12:10 even apparently.
## Why It's Good For The Game

Limiting (mainly) AI grief capacity good.
## Changelog
:cl:
fix: You can't open the shutters before the alamo can even take off anymore
/:cl:

### DIFF
--- a/code/game/objects/machinery/buttons.dm
+++ b/code/game/objects/machinery/buttons.dm
@@ -128,6 +128,11 @@
 /obj/machinery/button/door/open_only/landing_zone/attack_hand(mob/living/user)
 	if((machine_stat & (NOPOWER|BROKEN)))
 		return
+	#ifndef TESTING
+	if(world.time < SSticker.round_start_time + SSticker.mode.deploy_time_lock)
+		to_chat(user, span_notice("The containment shutters can't open yet!"))
+		return
+	#endif
 	if(!allowed(user))
 		to_chat(user, span_danger("Access Denied"))
 		flick("[initial(icon_state)]-denied", src)


Lumipharon:
Better question is why can the AI interact with the button at all? Aren't they supposed to not be able to interfer directly with groundside stuff?

ivanmixo:
I think it's fine for moments when there isn't really any command. It can interact with alamo yknow

Lumipharon:
The alamo is ship related, so thats definitely intended. But I'm pretty sure AI can't bolt lock groundside doors for example.

lbnesquik:
The AI is supposed to be able to interact with groundside stuff, as long as it isn't too impactful. You can interact with turrets, computers, cryo, intel computers and so on. It's just that there's very little _worth_ using, generally, and you only see stuff marines see, so they can use it themselves better. 
Doors were removed from the AI control because there was no counter to their usage. Bit like some versions of the hivemind. Same reason I neutered its control of the alamo doors. 

I think it's better to have this button be disabled before the shuttle gets there, as it prevents any actor from doing silly things, rather than just the AI. And also, being able to open shutters really helps when no one has access groundside, which happens not never!

Lumipharon:
True, I'm just mindful in the future if we have different gamemodes or whatever/admin events etc, where that timer might be bypassed this causes guh. As opposed to just not letting the AI touch it.

QualityVan (approved):
I was going to say AI doesn't use attack_hand, but attack_ai just routes into it for buttons anyway. This seems sensible as a current solution, and is easy enough to revisit if it ends up being a problem in some future gamemode. Seems unlikely though since that's explicitly what deploy_time_lock is for.

PR #13756, opened 2023-08-10
Adds dummy predator landmarks for ruTGMC

RipGrayson:
## About The Pull Request

Title.
The landmarks have no code behind them.

## Why It's Good For The Game

So ruTGMC wants to rebase on us, but the lack of predator landmarks means that any change we do to our maps will conflict and probably require ruTGMC to replace their predator landmarks whenever we update any map.

To avoid this I've added dummy predator landmarks to our maps. I don't see it as a huge sacrifice to make for our downstream, it's a single type (with no code attached) and less than a dozen landmarks on our maps.

If somebody knows a better way to solve this problem I'm all ears. 

## Changelog

:cl:
add: Added dummy predator landmarks for ruTGMC
/:cl:


RipGrayson:
This ain't it chief

PR #13757, opened 2023-08-10
Fixes freezing on hit

breadhunt:
## About The Pull Request

Fixes a significant bug which was the source of a _lot_ of freezing, particularly for xenos, upon hit. The cause was vignettes/etc. not being loaded in by BYOND, so whenever you'd be hit by any kind of projectile, BYOND's Very Funny Brilliant Code™️ would suddenly have to load every single screen icon ever, from pain to oxyloss to Gorger abilities.

This is probably not the correct solution, so please yell at me in discord to do whatever I need to do to get this PR to a workable state. It does work though. I tested it in local by smashing my head in with a machete and holy shit, it's so smooth. Usually I get a 10 second freeze on my awful laptop, when testing there was no freeze whatsoever.
## Why It's Good For The Game

Freezing very bad, especially when it can be in the double digit seconds.
## Changelog
:cl:
fix: Fixes a significant bug which caused freezing when hit.
/:cl:

### DIFF
--- a/code/_onclick/hud/fullscreen.dm
+++ b/code/_onclick/hud/fullscreen.dm
@@ -50,13 +50,14 @@
 /mob/proc/clear_fullscreens()
 	for(var/category in fullscreens)
 		clear_fullscreen(category)
+	overlay_fullscreen("empty", /atom/movable/screen/fullscreen/empty)
 
 
 /mob/proc/hide_fullscreens()
 	if(client)
 		for(var/category in fullscreens)
 			client.screen -= fullscreens[category]
-
+		client.screen += fullscreens["empty"]
 
 /mob/proc/reload_fullscreens()
 	if(client)
@@ -68,6 +69,7 @@
 				client.screen |= screen
 			else
 				client.screen -= screen
+		client.screen += fullscreens["empty"]
 
 
 /atom/movable/screen/fullscreen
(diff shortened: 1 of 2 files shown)


breadhunt:
> Seems like kinda a wack solution.

As said in the PR, this is probably the correct cause, wack solution. I am no coder, please advise a better way

> Lag from icon load is a one off. Once the icon has been loaded the issue shouldn't persist. So if this is causing issues for people its probs peanut internet speed or at least it should only happen the first time they're shot or whatever.

The issue is that if you do freeze, sometimes you're dead, so it doesn't really matter that it doesn't persist.
Also this is an extremely common issue, it's not just peanut internet speed, we've had it for years

TiviPlus:
I'm on mobile so I'm just gonna lock this so the plebs don't spam this pr

This pr isn't based on any knowledge of byond internal structure and is a completely random guess but I'll give it the benefit of the doubt and give it dnm until I can examine it

TiviPlus:
I'm on mobile so I'm just gonna lock this so the plebs don't spam this pr

This pr isn't based on any knowledge of byond internal structure and is a completely random guess but I'll give it the benefit of the doubt and give it dnm until I can examine it

Lumipharon:
AFAIK this is now redundant, regardless of whether it works or not, due to that fade in pr.

PR #13764, opened 2023-08-12
Compress icons with oxipng

vvvv-vvvv:
## About The Pull Request

`oxipng -o max --keep zTXt -arZ icons` was used. -12M difference

## Changelog

:cl:
imageadd: compressed icons
/:cl:


Lumipharon:
Wouldn't this need to be rerun every time a dmi is modified or added to maintain the benefits?

lbnesquik:
This could be a monthly thing, maybe?

spookydonut:
Instead of applying this to dmis and committing them, add it to the build script similar to how tgui is built where you compress them before the rsc is created.

PR #13765, opened 2023-08-12
Changes SG-85 sundering amount from 1 -> 1.75

ClosetedSkeletonGH:
## About The Pull Request

Changes SG-85 Sundering amount from 1 -> 1.75

for those of you who are math inclined, this is a 75% increase of sundering capability

Against: ancient queen (65 bullet armor, SG-85 has 25 AP)

BEFORE: 1 SUNDER

https://github.com/tgstation/TerraGov-Marine-Corps/assets/120620754/fcf53cc3-0bdd-4b3f-b20c-49d9b7255466

AFTER: 1.75 SUNDER

https://github.com/tgstation/TerraGov-Marine-Corps/assets/120620754/afb21211-3a7a-4f4a-bfb6-1371e6fffea8

## Why It's Good For The Game

This further seperates the SG-85 from SG-29 by making sure their sundering amounts aren't equivalent, its no secret that the SG-85 sucks (a lot).

It occupies a niche in the game of a Smartgun that actually functions as a suppression weapon, but due to many many issues of it always losing its ammo, kinda sucking, not being able to chase at all, I feel like this change is very appropriate since it encourages xenos getting shot by the SG-85 to probably try and not get shot by the SG-85.

SG-85 will always be the worst choice for nearly every situation but this hopefully makes its niche a bit stronger and stop SG's from shouting at new players when they take the SG-85. 

## Changelog

:cl:
balance: changes SG-85 Sundering from 1 -> 1.75
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1220,7 +1220,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	accurate_range = 12
 	damage = 10
 	penetration = 25
-	sundering = 1
+	sundering = 1.75
 	damage_falloff = 0.1
 
 /datum/ammo/bullet/smarttargetrifle


coldironwarrior:
i mean, yea. if we're going to sit and call it a noob trap, let's rebalance it to try and make it not a newb trap or less of; or straight-up remove it.

currently the only thing that really makes it "good" is that the bullets hitting xenos just lags them out, and im not sure i want this gun to be known for having the niche of...? capitalizing on spaghetti code?

lordexvar:
i was always in favor of the idea to transform the s-85 into a sort of req smartgun comparable to the vindicator, with buffs to fill it out for that role. nearly doubling its sunder in comparison to the sg-29 while keeping it as a round start weapon seems iffy to me, especially when it can already ably suppress.

ClosetedSkeletonGH:
> i was always in favor of the idea to transform the sg-85 into a sort of req smartgun comparable to the vindicator, with buffs to fill it out for that role. nearly doubling its sunder in comparison to the sg-29 while keeping it as a round start weapon seems iffy to me, especially when it can already ably suppress.

I mean its not the SG-29, nor does it fill the same role. Its movement is nearly static and so is its playstyle, but right now it hits pitifully weak. SG-29 works because it can chase down xenos decently well since you just dont need to stop firing ever, SG-85 fails because it struggles to deter xenos even with constant fire, allowing them to get very close which means the SG-85 user loses all capability to fire

Lumipharon:
SG-85 has better dps than the sg-29 in all situations, and signficantly more at medium to long range. If you try chase with it/frontline then yeah course its drawbacks make it worse.

If you want to be at the front/chase, you use the sg-29.
If you want to sit behind the lads and brrt, then the sg-85 is simply better.

Joe13413:
ye uhhh 75% more sunder is fucked
glad this got closed because changing 10 sunder/sec to 17.5 sunder/sec would make the sg-85 overpowered bro

PR #13767, opened 2023-08-12
Scout ammo changes

Lumipharon:
## About The Pull Request
Probably TM this first.

Scout ammo no longer has a minimum accurate range. It's had this for 6 years but it didn't even work until I fixed the mechanic, and I'm not entirely sure why the scout has it when its explicitly described as a gun that 'can be used at any range'.

Incendiary ammo now pierces mobs. Currently is completely over shadowed by base ammo in general usefulness, and completely overshadowed in terms of fire application by funny rapid fire incend xray laser.

High impact rounds no longer pierce mobs. They have slightly higher damage, lower AP and better sunder, but most significantly do knockback on top of the current slowdown. This makes it quite a strong soft CC tool against smaller xenos, but it'll need a TM to see if it's too much.

Also removed the 0% shrap chance. No real change in HvX, but this is another ?? stat.

And some important bullet color changes.
## Why It's Good For The Game
Acc minimum range was ?? for this req gun.
Incend ammo and high impact were both pretty meh compared to basic ammo. Should now have a more distinct niches of (more effective) incendiary application, and soft CC/sunder.
## Changelog
:cl:
balance: Scout no longer has an accurate minimum range
balance: Scout incendiary ammo pierces mobs
Balance: Scout high impact ammo now inflicts knockback and slowdown on smaller xenos, but no longer pierces. Also has better sunder but lower AP, and slightly higher damage
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -720,11 +720,9 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	name = "A19 high velocity bullet"
 	hud_state = "hivelo"
 	hud_state_empty = "hivelo_empty"
-	shrapnel_chance = 0
 	damage_falloff = 0
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
 	accurate_range = 15
-	accurate_range_min = 6
 	damage = 40
 	penetration = 20
 	sundering = 10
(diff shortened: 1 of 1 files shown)


Tailiik:
But can we put any kind of BR8 ammo into any BR8 magazine yet? :hampter: 

Not being able to refill your mag because it has a red stripe on it and the ammo box has a blue stripe is infinite agony

maikilangiolo:
Incend ammo is still the weakest choice. Half damage,  and 1/4 the sunder for easy to remove fire stacks (they're decent but overrated). Since both normal and impact have huge sunder, incend should get much higher AP so the very low damage is at least consistent, even against high armor.

ivanmixo (approved):

Joe13413:
> Incend ammo is still the weakest choice. Half damage, and 1/4 the sunder for easy to remove fire stacks (they're decent but overrated). Since both normal and impact have huge sunder, incend should get much higher AP so the very low damage is at least consistent, even against high armor.

honestly just think that standard ammo is the norm because everything else is so much expensive. I'd be much more willing to use impact/incend more if it was just more available

PR #13772, opened 2023-08-12
Increased Mortar Factory Viability 

lbnesquik:
## About The Pull Request
This PR did the following, and yes, it is my commit message : 
- Added smoke shells to the factory
- Added some flavor text to make things easier to handle
- Reduced the price of refills from 150 to 120 for HE and INC
- Reduced the price of refills for flare from 150 to 100
- Fixed the flare refills being called HE in the supply computer

Here is the math for HE, flares, and incendiary shells : 

For 10 pts, you get 2 in the req elevator. That's 5 pts per shell. 
For 150 pts, you can get a refill that becomes 30 shells. 150/30 = 5pts per shell. 

The new math for refills is that you start with 120 for HE and INC, meaning you get shells for 4 points each, roughly 20% of discount, without counting possible factory machines, and the bother to actually set it up. Given that tanglefoot shells currently enjoy an effective 40% discount via the factory, I think this is more than fair. 

The howitzer has actual discounts, so I didn't have to touch it. 

## Why It's Good For The Game
Factories are good for 2 things, making things you can't get otherwise, or making things in bulk for cheaper. 
Currently, the mortar factory was straight up worse for everything, save tanglefoot shells. This PR makes it something besides a noob trap. Especially if you factor in that some RO's likely bought machines to make said factories, worsening the ratio. 

## Changelog

:cl:
add: Added mortar smoke shells to the factory
qol: Improved refills flavor text, so you know what's actually in your hands
fix: fixed the Flares refills being wrongly called High Explosives
balance: Actually implemented a discount for HE, INC, Flare and Smoke mortar shells via factory
/:cl:

### DIFF
--- a/code/modules/factory/parts.dm
+++ b/code/modules/factory/parts.dm
@@ -491,6 +491,15 @@ GLOBAL_LIST_INIT(mortar_shell, list(
 	. = ..()
 	recipe = GLOB.mortar_shell
 
+/obj/item/factory_part/mortar_shell_tfoot
+	name = "mortar shell"
+	desc = "An unfinished flare mortar shell."
+	result = /obj/item/mortal_shell/plasmaloss
+
+/obj/item/factory_part/mortar_shell_tfoot/Initialize(mapload)
+	. = ..()
+	recipe = GLOB.mortar_shell
+
 /obj/item/factory_part/mortar_shell_flare
 	name = "mortar shell"
 	desc = "An unfinished flare mortar shell."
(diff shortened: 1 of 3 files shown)


Naaanii:
Would probably be better to remove the x2 for the base price of shells.

lbnesquik:
> Would probably be better to remove the x2 for the base price of shells.

I feel like that's outside the PR scope.

ivanmixo (approved):

PR #13784, opened 2023-08-13
Gives CMO the same timelock as Captain

chizzy376:
## About The Pull Request

Read title.
## Why It's Good For The Game
Higher bar of entry for an integral role that can make or break a round.

cmomoment
## Changelog
:cl:
admin: Gives CMO the same timelock as Captain
/:cl:

### DIFF
--- a/code/datums/jobs/job/shipside.dm
+++ b/code/datums/jobs/job/shipside.dm
@@ -650,7 +650,7 @@ A happy ship is a well-functioning ship."})
 		/datum/outfit/job/medical/professor,
 		/datum/outfit/job/medical/professor/robot,
 	)
-	exp_requirements = XP_REQ_EXPERIENCED
+	exp_requirements = XP_REQ_EXPERT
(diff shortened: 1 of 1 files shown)


ClosetedSkeletonGH:
"integral role", CMO is the least important role to TGMC and could be removed with 0 consequence (hint)

Just accept that a CMO cut your head off

chizzy376:
> "integral role", CMO is the least important role to TGMC

All the ops lost because of a bald CMO letting a larva shipside burst..

There is a reason why most people hate getting surgery and instead opt to autodoc which in turn griefs corpsmen because now their beacons have no power.

> could be removed with 0 consequence (hint)

So in your words, you wouldn't mind if this PR goes through? After all, he can be "removed with no consequences".

> Just accept that a CMO cut your head off

This is not the only time this has happened.

ivanmixo:
Personally don't think CMO has enough round impact to justify such a high bar.

PR #13794, opened 2023-08-14
Gives pistols full auto

ivanmixo:
## About The Pull Request

Also adjusts the unwielded scatter of 1911 and tp23 since they could be potentially problematic.
## Why It's Good For The Game

Pistols are kinda neglected right now, you see basically none carried. I think this is because of the harsh click gate they're locked behind, meaning you have to click for every shot, which isn't enjoyable to use. I tested out the feel and most pistols are still not more powerful than rifles even with full auto so the balance shouldn't be questionable. This doesn't affect revolvers. 
## Changelog
:cl:
balance: Pistols now have full auto! No more spam clicking
balance: Increased the unwielded scatter of the P23 and 1911
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -37,6 +37,7 @@
 	scatter = -2
 	scatter_unwielded = 4
 	akimbo_additional_delay = 0.9
+	gun_firemode_list = list(GUN_FIREMODE_AUTOMATIC)
 
 	placed_overlay_iconstate = "pistol"
 
@@ -195,6 +196,7 @@
 	fire_delay = 0.2 SECONDS
 	accuracy_mult = 1.20 //Has a forced laser sight.
 	accuracy_mult_unwielded = 0.95
+	scatter_unwielded = 4
 	recoil = -2
 	recoil_unwielded = -2
 
@@ -229,7 +231,7 @@
 	accuracy_mult_unwielded = 0.85
 	damage_mult = 1.15
 	recoil = -2
-	recoil_unwielded = -1
+	scatter_unwielded = 2
 
 /obj/item/weapon/gun/pistol/m1911/custom
 	name = "\improper P-1911A1 custom pistol"


Barnet2:
I suspect there will be some funny walance but overall a good chance with (possible?) tweaking, spam clicking is guh.

Lumipharon:
Pistols are sidearms, so they're always going to be neglected as they'd be unbalanced if they were actually useable over main weapons. Most people don't need sidearms because they're magharnessed up.

So its more of a case of they're not used because they're not needed. I don't particularly see why normal pistols should be full auto. Although if you're looking purely in terms of whats efficient, burst pistols (i.e. mod4) is still notably better than anything other than like, the FC pistol. Although that's probs because it wasn't meant to be a marine pistol at all.

Also you've actually buffed the 1911 scatter, and haven't changed the t23's at all.

ivanmixo:
> So its more of a case of they're not used because they're not needed. I don't particularly see why normal pistols should be full auto as it makes some very strong, when they shouldn't be by design.

They are still worse than rifles, but are actually usable.

> Most people don't need sidearms because they're magharnessed up. Pistols are sidearms, so they're always going to be neglected as they'd be unbalanced if they were actually useable over main weapons (t23 does better DPS than a t12, but does more than 4 times as much sunder, mod4 does more dps than a t18 against half the xenos, etc).

So why is them being used an issue?

> Although if you're looking purely in terms of whats efficient, burst pistols (i.e. mod4) is still notably better than anything other than like, the FC pistol. Although that's probs because it wasn't meant to be a marine pistol at all. God knows why they get it.

So the others shouldn't be usable because the mod is better?

Lumipharon:
Nah just mentioning the mod4 because someone inevitably will, and pointing out its not even balanced relative to other pistols because its not supposed to be a marine pistol.

Pariah919:
Pistols suck because you can't mag harness them and they have low capacity compared to rifles, I think mking them full auto then giving them some sort of magh barrel attachie would do good to make them much more appealing to run.

Mrrpip:
Aside from the lace and belt harness (Which is a bit overkill for a pistol.), there isn't much choice for the above.

That would be interesting to see.

Anyhow, onto what I was pondering, do you think it'd be simple enough to add Burst Auto for the pistols that can take the BFA but don't start with BA? (Namely the TP14 9mm.)

RyderFive:
Pistols are neglected and bad for a few main reasons.

Most have to be click spammed.

Damage is kinda bad on most and falls off.

Terrible storage aside from backpack slot. Shoulder holster, and pistol pouch are kinda just laughable to use given how useful most pistols are. Same with the belt slot carry rig.

Less reason to carry these if you can't even reap the main benefit of most pistols being quickdraw/wield aside from better storage, or if the damage can't make up for just wielding say a rifle when knocked down or stunned.

 Juggling pr also punishes any high firing delay setups that being mostly shotgun setups trying to use pistols because there's not exactly much reason to use a pistol when your gonna have to wait out a delay anyways to even fire it, may as well wield a better gun.

Why use a pistol over say the ar-18 for example? Simply no reason to use pistols if there's no real benefit over other weapons.

Pistols have needed this for a while to be more competitive along with a few other changes though.

Mod88 argument is quite funny to see.

PR #13810, opened 2023-08-16
ASL nerf

Lumipharon:
## About The Pull Request
ASL's now have a leadership of 2 instead of 3 (which is the same as actual squad leaders.

Also fixed an oversight in leaderpods, as they're not supposed to allow nonleaders in.

## Why It's Good For The Game
ASL being just as good as an actual SL is ??? considering it gives them direct buffs. Its still a very large advantage to have ASL, just not as good as actually being an SL anymore.

## Changelog

:cl:
balance: Acting squad leaders now have 1 level lower in leadership compared to actual squad leaders
fix: Fixed squaddies being able to use leader droppods
/:cl:

### DIFF
--- a/code/datums/jobs/squads.dm
+++ b/code/datums/jobs/squads.dm
@@ -301,7 +301,7 @@
 
 	//Handle aSL skill level and radio
 	if(!ismarineleaderjob(squad_leader.job) && !issommarineleaderjob(squad_leader.job))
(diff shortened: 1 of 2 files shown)


lbnesquik:
ngl, it was a long time coming, especially with the 8 SL tactics people were pulling off lately.

imsxz:
Why can you even assign ASL’s with living squad leaders lol

spookydonut (approved):

PR #13815, opened 2023-08-17
Removes the MK88 Mod 4 Combat Pistol from standard vendors and map spawns

lKiseki:
## About The Pull Request
Per title. ERTs and HvH loadouts will still get it, of course.

## Why It's Good For The Game
This is option 1 of two PRs I intend to put out for the 88M4 (see: #13816). This thing is allegedly meant to be an ERT-only sidearm that somehow became common access. It currently is hands down one of the best weapons in the game, and xenos aren't having a fun time fighting it in its current state... if they even get the chance to, considering how fast this kills.

## Changelog
:cl: Lewdcifer
del: MK88 Mod 4 Combat Pistol is no longer obtainable outside of ERTs and HvH loadouts.
/:cl:

PR #13816, opened 2023-08-17
Nerfs the MK88 Mod 4 Combat Pistol

lKiseki:
## About The Pull Request
Per title.
Sunder per bullet reduced from 2 to 0.5.
Burst fire now has an additional delay of 0.3 seconds between bursts.

## Why It's Good For The Game
This is option 2 of two PRs I intend to put out for the 88M4 (see: #13815). This thing is allegedly meant to be an ERT-only sidearm that somehow became common access. It currently is hands down one of the best weapons in the game, and xenos aren't having a fun time fighting it in its current state... if they even get the chance to, considering how fast this kills.

## Changelog
:cl: Lewdcifer
balance: MK88 Mod 4 Combat Pistol nerfed. Sunder reduced from 2 > 0.5, burst fire now has a 0.3s delay between bursts.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -407,7 +407,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	damage = 20
 	penetration = 12.5
 	shrapnel_chance = 15
-	sundering = 2
+	sundering = 0.5
 
 /datum/ammo/bullet/pistol/heavy
 	name = "heavy pistol bullet"
--- a/code/modules/projectiles/guns/pistols.dm
+++ b/code/modules/projectiles/guns/pistols.dm
@@ -516,6 +516,7 @@
 
 	fire_delay = 0.2 SECONDS
 	burst_delay = 0.1 SECONDS
+	extra_delay = 0.3 SECONDS
 	burst_amount = 3
 	accuracy_mult = 1.2
 	accuracy_mult_unwielded = 0.95


ivanmixo:
> This thing is allegedly meant to be an ERT-only sidearm that somehow became common access.

It was always in the weapon vendor, if "contraband". So no.

Mrrpip:
It should be noted, due to sharing the exact same ammo, this affects the Smartpistol by proxy, from the sunder change.

FIXTHERADBOXPLEASE:
I wouldn't worry so much about the Smartpistol, that thing still needs to be reworked too.

Mrrpip:
> I wouldn't worry so much about the Smartpistol, that thing still needs to be reworked too.

I'll believe it when I see the rework, haha. 

It's no big deal right now, though, don't think I've seen anyone use the SmartPistol as a gimmick primary.

Pariah919:
>This thing is allegedly meant to be an ERT-only sidearm that somehow became common access. 
>
The fact you're saying this just tells me you have absolutely no idea what you're actually talking about when it comes to 88mod history so I'll just clairify-

88mod has always been avaliable to marines, it was mostly just a PO gun that ended up in some of the contraband/req sections and could be acquired from there and when req vendor and stuff was removed it ended up shifted to marine vendors and became unlimited.

Though, nerfing the *entirety of pistol/ap* and then nerfing the burst fire on it is pretty harsh, I'd do the latter over the former since this just indirectly punishes anything that uses pistol/ap or will use pistol/ap, even if very little other than the mod4 uses it. Nerfing the sunder to be below normal pistol bullets is just silly.

ivanmixo (approved):
It still has higher more than double the AP compared to normal pistol bullets so probably okay.

Coastfront:
> It still has higher more than double the AP compared to normal pistol bullets so probably okay.

But why does this need to happen.
A hill, ripped away to nothing more than a stump from an endless river of rain yet stands taller than a field of grass, but it didn't need to be bulldozed.

This is a fine change, but why does it need to happen? akimbo 88 mod4 is strong but its not literally uncounterable.
If we went "X SIDE DOESEN'T LIKE PLAYING AGAINST THIS... IT KILLS TOO FAST...." then by that fucking logic runners should heal marines when they hit them, because marines don't like getting stunned into effectively-guaranteed-fractures when they go off alone.
The 88 mod4's weakness is both ammo economy and the fact that after you orgasm out all of your ammo (which happens fairly quickly due to its fire-rate) its absurdly easy to punish them, attack them while they scramble to tactical load one, stun them as they try, they both now have to pick them back up if they didn't use a pistol lace and bare minimum their reload is stalled to hit them more, and maybe break a leg/arm.
And even still is part of their logic made from a formless concept: "ert gun made available to everyone for some reason" is iirc just outright wrong, or bare minimum, is absurdly misguided given exactly how long this has been a thing, seemingly without issue.

There is such thing as a tree that knows its height, a mountain that needn't shave its peek, and change that doesen't need to happen.

Joe13413:
uhhh seems like a bit of a shit pr honestly

the sunder is the only problem i have with this gun, maybe just reduce it to 1-1.5 sunder? It's just a bit annoying to deal with the high sunder damage when you're a xeno.

PR #13817, opened 2023-08-17
railgun overhaul

Scarlet-Oduko:
## About The Pull Request
Gave a visual overhaul of railgun projectile and muzzle flash, along with upping its ammo some and giving it a burst setting, compensating with lower damage and sunder per shot, but higher overall if all shots land per burst.  

railgun-AP
	max_range = 14
	damage = 60
	penetration = 100
	sundering = 10
	on_pierce_multiplier = 0.85

railgun-hvap
	shell_speed = 5
	max_range = 21
	damage = 50
	penetration = 30
	sundering = 20

railgun-smart
	damage = 50
	penetration = 20
	sundering = 10
## Why It's Good For The Game
Sometimes appeal can be the difference between if a weapons used or not, updating its sprite, and in my controversial opinion giving it more rounds in a burst would help it preform in a more fun way, while the average damage will be the same, marines should be much less scared of missing a few shots as you can easily compensate by hitting more in the future to make up loss value.
## Changelog
:cl:
balance: All railgun canister have 9 rounds in them, lower base stats, higher total output if all shots connect
balance: Railgun has a 3 round burst now
add: new visuals for railgun rounds and muzzleflash visual changes 
/:cl:


Blundir:
The problem remains, high impact weapon with that idea originally in mind, turns into yet another rifle for no reason, railgun never needed an overhaul of it's core mechanic, it's good as it is.

Mrrpip:
...So uh....The Smart shell's superceded by the mosin of all things.

The HVAP will barely be rewarding for landing a scoped longshot.

And the normal one's a clumsier slug/superseded by the AMR.

I uhh....The railgun's not great as is, but uh...I don't think this helps it a lot.

Scarlet-Oduko:
> ...So uh....The Smart shell's superceded by the mosin of all things.
> 
> The HVAP will barely be rewarding for landing a scoped longshot.
> 
> And the normal one's a clumsier slug/superseded by the AMR.
> 
> I uhh....The railgun's not great as is, but uh...I don't think this helps it a lot.

suggests would be great, but do keep in mind that the balance of this is of hitting the burst, while individually the rounds may not be as strong as other, you'll be rapid firing three of them.

Mrrpip:
I don't know, railguns are typically hard hitters, I'll just throw out shitty ideas, maybe one will stick. I guess.

Uhhhhhh. let it burst fire a good chunk of it's mag at once? Make the ammo work like an energy gun (So you can choose between all 3)? Give it an underbarrel tactical sensor so wallbangs can be viable even with a scope?

Give it the AMR scope for easy aiming/extra damage or something.

----

Aside from that though, I pondered after waking up from a nap, maybe it could be alright with it's CC elements spammed? I don't know how fun that'd be to play against though.

PR #13820, opened 2023-08-18
Mentors no longer get automuted by using private message

DeltaFire15:
## About The Pull Request
This PR makes the private messages of mentors bypass the spam filter. This applies to both adminpms aswell as mentorpms, as both use the same framework after what tg did to them and I have not seen any way to clearly distinguish the two.
Done via private messages sending another arg to the spam filter check, which checks for it. Also done for admins, however admins generally bypass the filter anyways. I did do it nonetheless to ensure better futureproofing if someone for some reason removes general admin spam filter bypassing.
## Why It's Good For The Game
Any mentor will tell you how annoying it is to be writing up a long response, only for "Automute triggered, muted from adminpms / adminhelp". Please actually let people write long responses for things that warrant one instead of having to artificially truncate out important parts.
If someone writes malicious messages that would correctly trigger this filter, they shouldn't be staff in the first place, and it should be handled via that line instead.
## Changelog
:cl:
admin: Mentors now bypass spam filters when using private messages.
/:cl:

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -932,10 +932,10 @@ GLOBAL_VAR_INIT(automute_on, null)
 	if(isnull(GLOB.automute_on))
 		GLOB.automute_on = CONFIG_GET(flag/automute_on)
 
-	if((special_message_flag & MESSAGE_FLAG_MENTOR) && check_rights(R_MENTOR))
+	if((special_message_flag & MESSAGE_FLAG_MENTOR) && check_rights(R_MENTOR, FALSE))
 		return //Please stop muting me in my own responses.
 
-	if((special_message_flag & MESSAGE_FLAG_ADMIN) && check_rights(R_ADMIN))
+	if((special_message_flag & MESSAGE_FLAG_ADMIN) && check_rights(R_ADMIN, FALSE))
 		return //Technically not needed due to admin bypasses later in this proc, but I'll throw it in if for some reason someone changes their immunity.
 
 	total_message_count += 1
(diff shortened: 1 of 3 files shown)


code/modules/client/client_procs.dm
```diff
@@ -927,11 +927,17 @@
...
 	//Performance
 	if(isnull(GLOB.automute_on))
 		GLOB.automute_on = CONFIG_GET(flag/automute_on)
 
+	if((special_message_flag & MESSAGE_FLAG_MENTOR) && check_rights(R_MENTOR))
+		return //Please stop muting me in my own responses.
+
+	if((special_message_flag & MESSAGE_FLAG_ADMIN) && check_rights(R_ADMIN))
+		return //Technically not needed due to admin bypasses later in this proc, but I'll throw it in if for some reason someone changes their immunity.
+
```
psykzz, line 940:
```suggestion
	if((special_message_flag & MESSAGE_FLAG_MENTOR) && check_rights(R_MENTOR, FALSE))
		return //Please stop muting me in my own responses.

	if((special_message_flag & MESSAGE_FLAG_ADMIN) && check_rights(R_ADMIN, FALSE))
		return //Technically not needed due to admin bypasses later in this proc, but I'll throw it in if for some reason someone changes their immunity.

```

psykzz: otherwise you'll be showing a message "You don't have permission to do that"

DeltaFire15: You are right, good catch.

ivanmixo (approved):

PR #13828, opened 2023-08-19
Fix a swapped on/off message in valk UI

vvvv-vvvv:
## About The Pull Request

Nobody fixed it in 5 years

## Changelog

:cl:
fix: Fix a swapped on/off message in valk UI
/:cl:


Lumipharon:
? The message is correct though.
image

Lumipharon:
toggle or turn, "off" is still the correct word when it is currently on lol.

If I turn off the computer while its on, I wouldn't say turn it on,

If I have something thats shows its current state like say, injectors, I would say it as such.
image

PR #13840, opened 2023-08-20
Revert "Nerfs the MK88 Mod 4 Combat Pistol"

Lumipharon:
Reverts tgstation/TerraGov-Marine-Corps#13816

This PR wasn't thought through at all.
It makes the mod4 literally worse than every other normal marine pistol by a significant margin, both in terms of DPS and sunder. It also additionally made the burst fire on the burst pistol worse in every way then semi auto...

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -407,7 +407,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	damage = 20
 	penetration = 12.5
 	shrapnel_chance = 15
-	sundering = 0.5
+	sundering = 2
 
 /datum/ammo/bullet/pistol/heavy
 	name = "heavy pistol bullet"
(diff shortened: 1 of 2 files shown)


technicianman:
i feel that the sunder nerf was warranted but not the delay

KrabSpider:
> i feel that the sunder nerf was warranted but not the delay

Seconded. I'm testing it with the delay changes only, and it feels relatively fine-ish.

ivanmixo:
Not letting a flat revert through. It needs its sunder brought down at least. It has the highest AP and highest mag cap. You're just looking at the bullet numbers an not considering anything else, like how burst auto akimbo is stupidly strong and melts just about every caste in 3 seconds or less. Make a proper PR instead of a kneejerk revert.

Tailiik:
The p14 has the highest mag capacity doe honk

ivanmixo:
It's really bad, even with full auto and all

Lumipharon:
T14 does like 30-40% more dps than the mod4 now on top of double sunder per bullet and probably 3-4 times sunder/second, its incredibly funny

PR #13846, opened 2023-08-20
Resprites miners, deletes old miners

Xander3359:
## About The Pull Request
Resprites the miners.

https://github.com/tgstation/TerraGov-Marine-Corps/assets/66163761/c8616f5a-713c-4fbe-86bb-c66bdfafeee3
## Why It's Good For The Game
Old CM Sprites must die.
## Changelog
:cl: @KillerOrcaCora @Xander3359 
imageadd: Miners have new sprites
/:cl:


Barnet2:
Objectively better sprites, but personally do like the old ones better as stated above, too busy.

Al-1ce:
These would feel much more fitting on a super dark map.

technicianman:
these are very hard on the eyes but yeah they look cool af

novaepee:
decrease overclock animation time for speed

code/game/objects/machinery/miner.dm
```diff
@@ -42,14 +42,19 @@
 	var/miner_integrity = 100
 	///Max health of the miner
 	var/max_miner_integrity = 100
-	///What type of upgrade it has installed , used to change the icon of the miner.
+	///What type of upgrade it has installed , used to change the behaviour of the miner.
 	var/miner_upgrade_type
+	///What type of miner this is. Options: default_miner, autominer, armor_miner, overclock_miner, platinum_miner
+	var/miner_type = "default_miner"
+	///If the miner is a platinum miner
+	var/isplatinum = FALSE
```
ivanmixo, line 50: This is extremely snowflakey

ivanmixo (changes requested):
A bit too detailed if that's even possible, see what kuro says ig. Besides that can you seriously not think of a better way of doing this?

lbnesquik:
I will not lie, I do not think this fit our style.

chizzy376:
Love these miners, I'll find a use for them as a resprite, not as a complete replacement.
Could you make this its own separate entity while keeping the og miners untouched?

code/game/objects/machinery/miner.dm
Lumipharon: yeah setup some flags or something instead of both of these vars, this is mega wew.

Kuro020 (approved):
there is no such a thing as "too good to fit here", if its too good then the other sprites need to catch up, not the other way around

lbnesquik:
> there is no such a thing as "too good to fit here", if its too good then the other sprites need to catch up, not the other way around

I think most people mean that they don't fit with the rest of the current game.

Tailiik:
Yeah it's like...well sprited but the color palette is very...harsh I guess? Before I read any of the comments my first thought was also that it looked too "busy" compared to the rest of the sprite style. 

I dunno if it's just the colors, the amount of stripes, or both combined. It doesn't look _bad_ by any means, just out of place

henriquekill9576:
the miners we currently got are smooth, simple, have a single flashing light and generally stick to the same color for the entire bulkhead, from first glance these just look like an atomic bomb that is about to explode, colors are too messy, too many animations happening at once and etc, but yeah, as described above, they're just too busy

PR #13855, opened 2023-08-21
The AI icon is now random at roundstart

lbnesquik:
## About The Pull Request
When the AI spawns, it gets a random core icon. It helps people see the less common ones. 
I'll probably add a set preference for the icon in the future, but not tonight. 

## Why It's Good For The Game
More icon variety good. 

Ai-gif

## Changelog

:cl:
add: AI core display is now random at spawn
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/ai/ai.dm
+++ b/code/modules/mob/living/silicon/ai/ai.dm
@@ -82,6 +82,8 @@
 	laws += "Protect: Protect the personnel of your assigned vessel, and all other TerraGov personnel to the best of your abilities, with priority as according to their rank and role."
 	laws += "Preserve: Do not allow unauthorized personnel to tamper with your equipment."
 
+	set_random_ai_screen()
+
 	mini = new
 	mini.give_action(src)
 	create_eye()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/ai/ai.dm
```diff
@@ -440,6 +442,9 @@
 	linked_artillery = null
 	to_chat(src, span_notice("NOTICE: Connection closed with linked mortar."))
 
+/mob/living/silicon/ai/proc/set_random_ai_screen()
+	var/list/iconstates = GLOB.ai_core_display_screens
+	icon_state = resolve_ai_icon(pick(iconstates))
```
Lumipharon, line 447: why is this a separate proc? Its only used once.

Lumipharon (changes requested):

code/modules/mob/living/silicon/ai/ai.dm
lbnesquik: I figured it could be used a bunch of times in the future.

Lumipharon: We tend to want to avoid 'just in case' code. Unless you have specified use cases in mind you'd be using for a future pr after this is merged, its better to keep it inlined. If someone wants the functionality in the future, its like 5 seconds to make it into a proc.

lbnesquik: Ok I'll fix

Lumipharon (approved):

PR #13865, opened 2023-08-24
G8 from storage_slots=3 to max_storage_space=9

Looking4bros:
## About The Pull Request

Changes the G8 General utility pouches storage behavior from 3 slots to 9 storage space, same functional space, just divisible into smaller pieces.

## Why It's Good For The Game

Makes the G8 General utility pouch storage behave like the normal general pouches, just bigger and can hold normal weight class items. Good because one injector shouldn't fill the same amount of space as a GPMG ammo box.

## Changelog

G8 General utility pouch storage

:cl:
balance: G8 belt is no longer slot based, so can hold more smaller items
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -964,7 +964,8 @@
 /obj/item/storage/belt/sparepouch
 	name = "\improper G8 general utility pouch"
 	desc = "A small, lightweight pouch that can be clipped onto Armat Systems M3 Pattern armor or your belt to provide additional storage for miscellaneous gear or box and drum magazines."
-	storage_slots = 3
+	storage_slots = null
+	max_storage_space = 9
 	w_class = WEIGHT_CLASS_BULKY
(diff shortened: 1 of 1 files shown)


yyzsong:
> same functional space, just divisible into smaller pieces.

That means greater functional space, stinky
it can now hold nine of what it could only hold three of before.

Looking4bros:
> That means greater functional space, stinky
I eat crayons, Semantics isn't my strong suit.
---
> Makes the G8 General utility pouch storage behave like the normal general pouches,
Is my argument for this change. There are no general purpose belt slots, this belt is looks like it's supposed to be.

9 max_storage_space is half of the normal ammo belts, the only potential issue i can see with this setup is that the throwing knife rig holds 6 tiny items in comparison. The next closest belt is the aforementioned ammo belt with 6 storage_slots and 18 max_storage_space.

Lumipharon:
I feel like it was slots for some specific reason... But I can't think of what it is, so guess we'll wait for any xope.

PR #13866, opened 2023-08-24
Adds a recipe to medical nanites

henriquekill9576:
## About The Pull Request

changes the medical nanites recipe from: 
10 parts nanites (not medical, razorburn for whatever reason), 100 parts iron, 5 parts radium and 5 parts lemoline = 1u of nanites
to: 
1 part medical nanites catalyst, 10 part iron and 1 part lemoline = 9u of nanites

## Why It's Good For The Game

nanites are already free in the vendor, this recipe is a just a QOL feature incase you want bigger quantities of nanites but don't want to be sitting at the vendor injecting beaker, restocking, vending and repeat while also keeping the small cost of lemo

## Changelog

:cl:
add: Added an actual recipe for medical nanites
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/autoinjectors.dm
+++ b/code/game/objects/items/reagent_containers/autoinjectors.dm
@@ -309,6 +309,7 @@
 	amount_per_transfer_from_this = 1
 	volume = 1
 	list_reagents = list(/datum/reagent/medicine/research/medicalnanites = 1)
+	free_refills = FALSE
 
 /obj/item/reagent_containers/hypospray/autoinjector/pain //made for debugging
 	name = "liquid pain autoinjector"
--- a/code/modules/reagents/reactions/medical.dm
+++ b/code/modules/reagents/reactions/medical.dm
@@ -215,8 +215,9 @@
 
 /datum/chemical_reaction/medicalnanites
 	name = "Medical Nanites"
-	results = list(/datum/reagent/medicine/research/medicalnanites = 10)
-	required_reagents = list(/datum/reagent/medicine/research/medicalnanites = 1, /datum/reagent/iron = 5, /datum/reagent/medicine/lemoline = 1)
+	results = list(/datum/reagent/medicine/research/medicalnanites = 9)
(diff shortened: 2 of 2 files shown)


Xander3359:
so 1u of lemoline for 4u of nanites? This seems kinda bad chief idk.

InterroLouis:
That seems really bad actually. Especially since the lemoline is consumed when making them and nanites are generally considered...well, not particularly good in terms of medical usefulness.

henriquekill9576:
> That seems really bad actually. Especially since the lemoline is consumed when making them and nanites are generally considered...well, not particularly good in terms of medical usefulness.

downside of nanites is that you lose blood when they go below a certain unit threshold, with this in mind, this allows you to produce massive quantities of nanites to inject yourself with incase you don't want to have to deal with the blood thing with the added cost of some lemo, it also doesn't completely negate all downsides of nanites since you'll get some tox damage from the OD, which can be countered by dylo but will need continuous management to ensure you don't die from the tox damage

although i do agree with xander that the cost might be a little bit too low, probably gonna raise it to 1 lemo for 10 nanites

overall i think its just a nice change for me and the other 3 people that do nanite builds so we can have a little bit more variety instead of just always packing 10 pill packs of iso every time you're doing nanites

InterroLouis:
Imagine using lemoline to make nanites when you could use it for T2 healing chems.

henriquekill9576:
a small price to pay for para tri oxy nanites combo

Mrrpip:
Long Live the paracetamol, tricordazine bifecta.

ivanmixo (approved):

henriquekill9576:
> 1 to 9 is a fairly high ratio, but I'm fine with it if you kill nano injector restocking. See #13256 for how that'd work.

imo killing restocking would do more bad than good, would going back to 1 to 5 be okay if i were to keep restocking in that case?

QualityVan (approved):

ivanmixo (approved):

PR #13883, opened 2023-08-25
Makes the sulaco drop pod bay look better

lbnesquik:
## About The Pull Request
The current sulaco drop pod bay is not lined up well with the alamo landing south of it, which is most displeasing. This PR fixes that, and makes it more pleasing.  It also increased the number of drop pods by about 10%

## Why It's Good For The Game
It seriously looks better.

## Changelog

:cl:
qol: Made the sulaco drop pod bay line up with the nose of the alamo :)
balance: Increased the number of drop pods in the sulaco by 11%

/:cl:


Derrick2Electricboogaloo:
Fair enough, I should have done this way back when, but figured I would get smacked down for extra pods.

lbnesquik:
There are no standard drop pods amounts.

Issue #10652, opened 2022-07-30
recycler does not work on the pillar of spring


restedwaves:
### Description

the recycler should remove things that are belted into it, it does not do so

### Test Merges

Server-Side Code Modifications: https://tgstation13.org/parsed-logs/sybil/server_side_modifications.dm

Round ID: [15761](https://scrubby.melonmesa.com/round/15761)
BYOND Version: 514.1584
Server revision compiled on: 
Master commit: [a3363bfdf0467375ecebe88a7153c9d8248dc7b7](https://github.com/tgstation/TerraGov-Marine-Corps//commit/a3363bfdf0467375ecebe88a7153c9d8248dc7b7)
The following pull requests are currently test merged:
[#10634: 'Adds normal smoke grenades' by Lauwma at commit 7ef930e0e1](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10634)
[#10567: 'Dead marines do not reduce number of xenos for crash' by Autisem at commit d316869fa7](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10567)
[#10558: 'Vendors improve' by goose4429 at commit c857ce75eb](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10558)
[#10629: 'Remove permutated from projectile code' by BraveMole at commit e9e3026a99](https://github.com/tgstation/TerraGov-Marine-Corps//pull/10629)

Local commit: 5443d5fc88b483b92b5637889f2a1bbb4cd85fda
Server tools version: /tg/station 13 Server v3.2.6.2

### Reproduction Steps

turn on the recycler's belt, watch as the garbage is all merely forced to one tile in front of said recycler

Derrick2Electricboogaloo:
Easy fix, copy minerva, or Sulaco recycler and replace it. It will then work.

Lumipharon:
not sure if this is still occuring, but easy fix.

PR #13887, opened 2023-08-26
Makes recyclers point in their relevant direction. 

lbnesquik:
## About The Pull Request
Fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/10652

The issue was that recyclers only have sprites for 1 direction, and they point north by default, instead of west. I made the default point west and fixed minerva's one by forcing it to point north. Avoids us having to redo it. 

## Why It's Good For The Game
Working gear is good!

## Changelog

:cl:
fix: fixed some recyclers not working properly.
/:cl:

### DIFF
--- a/code/modules/recycling/recycler.dm
+++ b/code/modules/recycling/recycler.dm
@@ -6,6 +6,8 @@
 	layer = ABOVE_MOB_LAYER
 	anchored = TRUE
 	density = TRUE
+	//Pointing west because that's the only sprite we got
+	dir = 8
 
 /obj/machinery/recycler/update_icon()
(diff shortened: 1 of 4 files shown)


code/modules/recycling/recycler.dm
```diff
@@ -6,6 +6,8 @@
 	layer = ABOVE_MOB_LAYER
 	anchored = TRUE
 	density = TRUE
+	//Pointing west because that's the only sprite we got
+	dir = 8
```
TiviPlus, line 10: Use the define

TiviPlus (changes requested):

code/modules/recycling/recycler.dm
lbnesquik: done

PR #13888, opened 2023-08-26
Adds xenomorph IFF tags

DeltaFire15:
## About The Pull Request
This PR adds a new component to the game which when attached to a xenomorph carries an IFF signal for them.
Additionally, adds integration for said component into things that work with IFF (if there are more and I forgor them, tell me)

Using this component, adds an item that when used on a xenomorph applies this component to them, also also adds a terragov version of this IFF tag to requisitions for 130 points.
Intended for corrupted, or for admins that want to do some eventing, without their allied xenos getting blasted by smartgunners, or accidentally'd by a Vali with bump enabled.

Note: Does not have a visual appearance on the xeno, however existence and allegiance are reflected on examining the xeno; Use the special xeno hives for easier recognition if you are bussing. Also is preserved throughout maturing and evolutions.

Supports different IFFs, aswell as having multiple components interact with the new signal at once. (However xenos can only have one tag attached as a precaution for some silly bean trying to stack ten tags of the same kind on a xeno for some reason)
Different IFFs demonstrated via some subtypes of the tag, for ease of use (though it's very easy to modify anyways, thank you varedit)

Is comsig-based as mentioned because that seemed useful here and also having the iff check signal may allow for some interesting subterfuge-related xeno stuff in the future.

## Why It's Good For The Game
There are quite a few times where xenos exist that are allied to one or another human faction due to events or corrupted being on the field, which however can be more of a nuisance to the human faction than they are of help, due to them getting hit by IFF weapons, aim mode fire, and especially anything with an AI brain that only cares for IFF.
With this change, it should be a lot less problematic for xenos to be used to support a primarily-human faction, without humans needing even higher levels of care to avoid accidents.

## Changelog

:cl:
add: Xenomorph IFF tags, which allow xenomorphs to be recognized as friendly by things with a compatible IFF signal.
/:cl:


ivanmixo (changes requested):
Super weird to have xeno IFF specifically be a component, but the rest of how IFF gets handled is through flag vars. Honestly not sure how to feel about it.

DeltaFire15:
>Super weird to have xeno IFF specifically be a component,

I've had two reasons for this particular implementation.
1. If I used a var I'd have had to decide whether to *always* pass it on evolve, or to *never* pass it on evolve, both of which is suboptimal. Here, it is on a case-by-case basis as the component decrees.

2. This implementation supports seperate "sources" of the same IFF existing and / or being removed on one xenomorph. With a variable, you could not do this as you can only either turn the bitfield on or off. Here, seperate sources are aggregated instead, futureproofing interactions between potentially seperate IFF-carrsying sources.

It may be a bit odd compared to humans, but it fits well for xenomorphs with their lack of equipment to carry it. and their high amount of mob changes.

QualityVan:
When would you want to not inherit IFF flags on evolve?

DeltaFire15:
> When would you want to not inherit IFF flags on evolve?

Many things. Just about any addition that would be caused by something ability based, for example.
Hell, even for the tags here I contemplated whether to preserve it or not (and decided for the former in this case because it'd artificially inhibit xeno flexibility with a req pricetag)

ivanmixo (approved):

PR #13923, opened 2023-09-02
Fix TTS preview voice being always robotic

forestbbbbbbb:
## About The Pull Request

The "This is my voice" preview in prefs was always robotic because in python `bool('0') == True`. This could be fixed on the tts-api.py side too but I'm not sure anyone will update that in a meaningful amount of time

## Changelog

:cl:
fix: Fix TTS preview voice being always robotic
/:cl:


TiviPlus (approved):
lol

PR #13615, opened 2023-07-27
Maturity Removal (Part of the Xeno Rework)

Barnet2:
## About The Pull Request
Starts the xeno rework as per https://hackmd.io/@uckgLKizQP6QNGj1w6mHQg/r14h9qS9h by Tivi, Barnet (me) and Lewdcifer
Removes maturity for xenos, sets starting stats as ancient
Increases evo timer by 25%
Maturity system stays in for primos; you have to live long enough to become primo
Maturity towers become Psychic Relays, increase slots like current evo towers. Evo towers increase evo points and maturity points (leftover for primo) by 0.2 opposed to 0.25
Psy points lowered to 1400 in anticipation for the tactical point rework; enough for a silo and two towers
**THIS IS SUPPOSED TO BE COMBINED WITH OTHER PR'S, VIEW IT AS SUCH**
## Why It's Good For The Game
As per the xeno rework with reducing overall xeno spawns, we are able to remove a horrible mechanic: Maturity.
Why do I call Maturity so bad?
1. Firstly, maturity is horrible for new players. Call it skill issue, sure, but a new player cannot experience the full power of a Xenomorph because they will die before they get there. Even if they will be able to play less (will be remedied with playing minions in queue), they will have a more consistent experience.
2. Changes how you play the game negatively. After talking to multiple players that (for the record) play a good amount of xenomorph and marine, resting until you are at X age is not good gameplay for either sides. A consistent xenomorph is a better system. Killing a xenomorph will have more reward, too, in reducing their numbers by a good bit for awhile as per the silo rework.
For the evo towers, needed to split them up into psychic relays with the removal of Maturity.
## Changelog
:cl:
add: Adds Psychic Relays, increase xenos slots. Removes slot increase from evo towers. Starting psy down to 1400
del: Removes maturity towers
balance: Removes the maturity system, keeps it for primos. Xenos now start at ancient stats.
/:cl:

PR #13925, opened 2023-09-03
Adds a xeno respawn timer, nerfs silo scaling

Barnet2:
## About The Pull Request
ONLY supposed to be paired up with https://github.com/tgstation/TerraGov-Marine-Corps/pull/13615
Adds a xeno respawn timer as an alternative to the silo rework and nerfs Nuclear War larva generation, to keep a punishment for dying with the maturity rework.
Current time is 5 minutes, 3 for crash. Silo scaling down from 2x to 1.5x.
It applies if you die as a xeno or ghost as a xeno, but not if you are a minion to not punish players for taking minions while waiting
## Why It's Good For The Game
Keeps a punishment for dying as Xeno while keeping it user friendly (i.e larva generation being x number per minute)
## Changelog
:cl:
balance: Added a xeno respawn timer on xeno death and ghosting, does not apply if you took a minion
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -106,7 +106,10 @@
 #define TIME_BEFORE_TAKING_BODY 1 MINUTES
 
 #define DEATHTIME_CHECK(M) ((world.time - GLOB.key_to_time_of_role_death[M.key]) < SSticker.mode?.respawn_time)
-#define DEATHTIME_MESSAGE(M) to_chat(M, span_warning("You have been dead for [(world.time - GLOB.key_to_time_of_role_death[M.key]) * 0.1] second\s.</span><br><span class='warning'>You must wait [SSticker.mode?.respawn_time * 0.1] seconds before rejoining the game!"))
+#define DEATHTIME_MESSAGE(M) to_chat(M, span_warning("You have been dead for [(world.time - GLOB.key_to_time_of_xeno_death[M.key]) * 0.1] second\s.</span><br><span class='warning'>You must wait [SSticker.mode?.respawn_time * 0.1] seconds before rejoining the game!"))
+
+#define XENODEATHTIME_CHECK(M) ((world.time - (GLOB.key_to_time_of_xeno_death[M.key] ? GLOB.key_to_time_of_xeno_death[M.key] : -INFINITY) < SSticker.mode?.xenorespawn_time))
+#define XENODEATHTIME_MESSAGE(M) to_chat(M, span_warning("You have been dead for [(world.time - GLOB.key_to_time_of_xeno_death[M.key]) * 0.1] second\s.</span><br><span class ='warning'>You must wait [SSticker.mode?.xenorespawn_time * 0.1] seconds before rejoining the game as a Xenomorph!"))
 
 #define COUNT_IGNORE_HUMAN_SSD (1<<0)
 #define COUNT_IGNORE_XENO_SSD (1<<1)
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -106,7 +106,7 @@
 #define TIME_BEFORE_TAKING_BODY 1 MINUTES
 
 #define DEATHTIME_CHECK(M) ((world.time - GLOB.key_to_time_of_role_death[M.key]) < SSticker.mode?.respawn_time)
-#define DEATHTIME_MESSAGE(M) to_chat(M, span_warning("You have been dead for [(world.time - GLOB.key_to_time_of_xeno_death[M.key]) * 0.1] second\s.</span><br><span class='warning'>You must wait [SSticker.mode?.respawn_time * 0.1] seconds before rejoining the game!"))
+#define DEATHTIME_MESSAGE(M) to_chat(M, span_warning("You have been dead for [(world.time - GLOB.key_to_time_of_role_death[M.key]) * 0.1] second\s.</span><br><span class='warning'>You must wait [SSticker.mode?.respawn_time * 0.1] seconds before rejoining the game!"))
 
 #define XENODEATHTIME_CHECK(M) ((world.time - (GLOB.key_to_time_of_xeno_death[M.key] ? GLOB.key_to_time_of_xeno_death[M.key] : -INFINITY) < SSticker.mode?.xenorespawn_time))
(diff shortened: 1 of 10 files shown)


Arturlang:
Does it reset if you go minion, or get completely removed then? Would be best if you could spend your time as a minion and it would notify you when you could respawn and take a larva

Barnet2:
> Does it reset if you go minion, or get completely removed then? Would be best if you could spend your time as a minion and it would notify you when you could respawn and take a larva

It keeps ticking while you are as minion, if I figure it out I could apply the status timer in the infobox while you are playing minion

code/modules/mob/living/carbon/xenomorph/hive_datum.dm
```diff
@@ -939,6 +939,15 @@ to_chat will check for valid clients itself already so no need to double check f
 		if(length(possible_silos))
 			return attempt_to_spawn_larva_in_silo(xeno_candidate, possible_silos, larva_already_reserved)
 		if(SSticker.mode?.flags_round_type & MODE_SILO_RESPAWN && !SSsilo.can_fire) // Distress mode & prior to shutters opening, so let the queue bypass silos if needed
+			if(XENODEATHTIME_CHECK(xeno_candidate))
+				if(check_other_rights(xeno_candidate.client, R_ADMIN, FALSE))
+					if(tgui_alert(xeno_candidate, "You wouldn't normally qualify for this respawn. Are you sure you want to bypass it with your admin powers?", "Bypass Respawn", list("Yes", "No")) != "Yes")
+						log_admin("[key_name(xeno_candidate)] used his admin power to bypass respawn before his timer was over")
```
ivanmixo, line 945: Oh actually I noticed this is wrong. This message will get sent if they press anything but yes, which is... not ideal because then it'd just be a lie. They declined to skip the queue, why is it being sent?

ivanmixo (changes requested):

code/modules/mob/dead/observer/observer.dm
```diff
@@ -338,6 +347,12 @@ GLOBAL_VAR_INIT(observer_default_invisibility, INVISIBILITY_OBSERVER)
 			. += "Respawn timer: READY"
 		else
 			. += "Respawn timer: [(status_value / 60) % 60]:[add_leading(num2text(status_value % 60), 2, "0")]"
+	if(GLOB.respawn_allowed && MODE_INFESTATION)
```
TiviPlus, line 350:
```suggestion
	if(GLOB.respawn_allowed)
```
MODE_INFESTATION is a number so always true

code/datums/gamemodes/_game_mode.dm
```diff
@@ -28,6 +28,8 @@ GLOBAL_VAR(common_report) //Contains common part of roundend report
 	var/deploy_time_lock = 15 MINUTES
 	///The respawn time for marines
 	var/respawn_time = 30 MINUTES
+	//The respawn time for Xenomorphs
+	var/xenorespawn_time = 5 MINUTES
```
TiviPlus, line 32: 0 or 1 on extended?

TiviPlus (commented):
rest lgtm

code/__DEFINES/mode.dm
```diff
@@ -106,7 +106,10 @@
 #define TIME_BEFORE_TAKING_BODY 1 MINUTES
 
 #define DEATHTIME_CHECK(M) ((world.time - GLOB.key_to_time_of_role_death[M.key]) < SSticker.mode?.respawn_time)
-#define DEATHTIME_MESSAGE(M) to_chat(M, span_warning("You have been dead for [(world.time - GLOB.key_to_time_of_role_death[M.key]) * 0.1] second\s.</span><br><span class='warning'>You must wait [SSticker.mode?.respawn_time * 0.1] seconds before rejoining the game!"))
+#define DEATHTIME_MESSAGE(M) to_chat(M, span_warning("You have been dead for [(world.time - GLOB.key_to_time_of_xeno_death[M.key]) * 0.1] second\s.</span><br><span class='warning'>You must wait [SSticker.mode?.respawn_time * 0.1] seconds before rejoining the game!"))
```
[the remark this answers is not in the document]
Barnet2: Oops, fixed

Arturlang:
Alternative: If a xeno dies within 15 mins of a respawn, apply this timer only then. And disable it on crash. *please*

Joe13413:
not really a big fan of this, especially considering that there is a 1 minute timer before you can take a minion

Mannybrado:
this sucks so bad. its just not fun. at the very least, remove the minion and SSD mob timers so you can still play the game. 
its just 5 minutes, sure, but previously, as long as the server has existed, there has only ever been a break in xeno playtime if there weren't any larva. this just kills any momentum i have playing the game if i end up dying and i just want to stop. i think the rest of the xeno rework changes are decent, but this is easily the only straight up bad one.

Victor239:
The respawn timer doesn't adjust in any way to the context - time you spent alive, how many larva there are, minion count, etc. 

If you stay alive for 30 mins for example there shouldn't be a respawn timer - you've already proven careful to avoid the risk of death. Likewise if there are multiple burrowed larva a respawn timer doesn't make sense, your team is already being punished by being underpopulated. Minion count (e.g. 3+) should also remove the minion respawn timer, since no-one really cares to play them if they have to wait a whole minute to do so.

ivanmixo:
You guys are completely ignoring the fact that xenos are more powerful when they are alive now due to no mat. The game shouldn't accommodate you dying by just letting you back in instantly.

Victor239:
Yes, which is why the majority of the time there would be this respawn timer. But there are a few exceptions as mentioned where it makes less sense to have one, or it being reduced in length. 

Like yesterday there was a round with 8 burrowed and the few dead xenos in deadchat were complaining about this respawn timer existing then. The team is already at a massive disadvantage - if say 6 of those slots would allow instant respawn that'd go a long way to balancing out the teams when there are already insufficient xeno players online. 

I believe your argument then when something similar was pointed out was "but then some 5 hour runner is gonna waste all those slots", but really:
- An extra xeno to be a distraction for a while is better than a permanent burrowed slot open
- It's just a game, let's not be elitist towards newbies please
- Newbies would learn castes faster by having some quick respawns available and then be much less likely to die quickly in the future

Mannybrado:
image

Funny that this can happen. Funny.

Burger1998:
its over bro's...

Victor239:
Or just put them in last place in the queue, better QoL.

PR #13926, opened 2023-09-04
Changes chem related brute/burn healing behavior

Looking4bros:
## About The Pull Request

Changes chem related brute and burn healing to carry over additional health to another damaged limb when the randomly targeted limb has less damage than the chem is supposed to heal. This is most notable with big ticket healing chems, such as Russian red, Neuraline, Somolent, Clonexedone, and Cyroxadone (not that it matters much with cyropods)

## Why It's Good For The Game

Good because it makes all chems heal their specified amount, regardless of randomly selected limb.

## Changelog
:cl:
balance: Chem healing on some chemicals is now overall instead of per limb
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
+++ b/code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
@@ -59,7 +59,7 @@
 		occupant.adjustOxyLoss(-1)
 	if (occupant.getToxLoss())
 		occupant.adjustToxLoss(-1)
-	occupant.heal_limb_damage(1, 1, updating_health = TRUE)
+	occupant.heal_overall_damage(1, 1, updating_health = TRUE) // changed to use heal_overall_damage as cyropods wash over the entire body, not one limb at a time
 	var/has_cryo = occupant.reagents.get_reagent_amount(/datum/reagent/medicine/cryoxadone) >= 1
 	var/has_clonexa = occupant.reagents.get_reagent_amount(/datum/reagent/medicine/clonexadone) >= 1
 	var/has_cryo_medicine = has_cryo || has_clonexa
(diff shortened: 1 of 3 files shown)


Looking4bros:
i tested the chems on a local server to make sure there wasn't any wacky added healing. They only heal what people thought they heal now. Russian red will heal 100 Brute and Burn, Neuraline will heal 75 brute and burn, Vali will heal it's prescribed (6 * boost_amount * brute/burn_boost_amount) If things turn out to be too strong, i suggest lowering the total values, instead of reverting the code, as this removes the randomness.

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
```diff
@@ -59,7 +59,7 @@
 		occupant.adjustOxyLoss(-1)
 	if (occupant.getToxLoss())
 		occupant.adjustToxLoss(-1)
-	occupant.heal_limb_damage(1, 1, updating_health = TRUE)
+	occupant.heal_overall_damage(1, 1, updating_health = TRUE) // changed to use heal_overall_damage as cyropods wash over the entire body, not one limb at a time
```
ivanmixo, line 62: Pointless comment but lgtm

ivanmixo (approved):

code/modules/atmospherics/machinery/components/unary_devices/cryo.dm
Looking4bros: i'm at the babbling point of learning code. i'll keep my babbling to github comments in the future.

ivanmixo: Yea get rid of it please

QualityVan (approved):
I'm intuitively a little leery of this on nanites specifically, same with vali, but I can't elucidate why even to myself so it's probably fine. Sensible change in general, although with the future caveats you mention in the desc.

PR #13931, opened 2023-09-06
Tarp Invisibility bug

Looking4bros:
## About The Pull Request

Stops people from being able to turn their character invisible for EORD

## Why It's Good For The Game

No more staying invisible when you are forceMove'd from a tarp. Only really happens for EORD, but yeah.

## Changelog

No more EORD invisibility.

:cl:
fix: Thermal Tarp no longer keeps you invisible when you are not inside it.
/:cl:


Looking4bros:
ope, this just brings back the issue of AI being able to see mobs inside invisible structures. Is no good.

PR #13939, opened 2023-09-07
Sends all of the Tadpole variants to the adminrealm

Pariah919:
## About The Pull Request
Makes all tadpole variants currently ingame admin only.
## Why It's Good For The Game
Tadpole variants that aren't basically slightly changed standard tad are bad for balance because it can let tad do things it shouldn't like landing in halls and stuff, aswell as dumb stuff such as 3 weapon hardpoints this mainly aims to makes tadpole consistent again in what it can and cannot do.
## Changelog
:cl:
del: All tadpole variants other than standard are admin only now.
/:cl:

### DIFF
--- a/code/datums/shuttles.dm
+++ b/code/datums/shuttles.dm
@@ -130,6 +130,7 @@
 	suffix = "_big"
 	description = "Tadpole-01, the old model barely in service for TGMC, replaced by the newer Tadpole-03. Much like an APC, is pretty armored. Very lacking in firing angle."
 	display_name = "Tadpole Carrier Model"
+	admin_enable = FALSE
 
 /datum/map_template/shuttle/minidropship/food
 	suffix = "_food"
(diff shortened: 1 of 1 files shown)


FosterGrime-Heymynameslouie:
I kinda like fatpole and suspole it feels like normal tadpole but fatter

lbnesquik:
At least keep the fat pole!

FIXTHERADBOXPLEASE:
Like 90% of them are questionable at best, but the sussy wussypole has genuine trade-offs.

Lumipharon (changes requested):
I don't have a problem with big tad or the sustad, or classic tad. They all have tradeoffs due to their large/unusual size.

Mannybrado:
just remove the meme tads please

dopamiin0:
idk i like all the serious tads fine even though i basically always just use the standard one anyways. i do think removing ones like kitchen tad is good though because PO already has a reputation as a grief role and giving them literal grief options does not help

Lumipharon:
@Pariah919 the large tad is fine as well imo, but I'm open if other people think it's unreasonable in some way.

PR #13954, opened 2023-09-13
Fix vendor having no icon while vending

Blundir:
## About The Pull Request

When pr about removing vendor delay was created, author didn't take into account that vendors needs at least some delay for the icon code to trigger. I added 0.1 delay which is basically nothing.

## Why It's Good For The Game

Bug - bad, fix - good.

## Changelog
:cl:
fix: fixed vendor not using special icon while vending an item
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/vending.dm
+++ b/code/game/objects/machinery/vending/vending.dm
@@ -76,7 +76,7 @@
 	///If the vendor is ready to vend.
 	var/vend_ready = TRUE
 	///How long it takes to vend an item, vend_ready is false during that.
-	var/vend_delay = 0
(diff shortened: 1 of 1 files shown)


Lumipharon:
?? Why not just change the vend code to not rely on a vend delay? (seriously, why the fuck do machines only draw power if they have a vend delay, what even is this code)

Blundir:
> ?? Why not just change the vend code to not rely on a vend delay? (seriously, why the fuck do machines only draw power if they have a vend delay, what even is this code)

I'd personally want to use delay code later, also why the person who removed delay, didn't it themselves, I just want to fix the issue, not to cut functionality

PR #13957, opened 2023-09-15
Superheavy pistol changes

Pariah919:
## About The Pull Request
Super heavy pistol ammo now deals knockback and has slightly less sunder/slowdown. It's effectively nearly the same as non TP-44 revolver ammo, also reduced falloff slightly.

Deagle
Fire delay 0.7->0.3
Better base handling overall.

Hipower
Fire delay 1->0.35
## Why It's Good For The Game
These guns suck complete doodoo, they don't work as pocket stunners because their damage and handling is trash, M-44s generally just handle better despite being stuck as single action, and their ammo economy is equally as bad. If the fire delay is too low I'd probably just make it .35 and .4 respectively and they'd be fine forever, since getting rid of the stagger would just make them worse heavy pistols.
## Changelog
:cl:
balance: Superheavy pistols like the Deagle and Hipower have significantly less fire delay, deal a bit less sunder, deal halved slowdown and have better handling.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -422,10 +422,10 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	hud_state = "pistol_superheavy"
 	damage = 45
 	penetration = 15
-	sundering = 3.5
+	sundering = 3
 
 /datum/ammo/bullet/pistol/superheavy/on_hit_mob(mob/M,obj/projectile/P)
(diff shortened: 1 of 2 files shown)


Mannybrado:
MJP name a buff pr correctly as a buff challenge(impossible)

Joe13413:
ok now i dont use these pistols much but it's worth mentioning that on the hipower, some people love to use that one for the stagger. These guns are generally shit IMO but worth mentioning in case you overlooked that

ivanmixo (approved):
Has potential to be cancer but we'll see

PR #13959, opened 2023-09-15
Decreases time between autobalance checks

Barnet2:
## About The Pull Request
Read title. Currently at 5 minutes between each check; moving down to 3 minutes.

## Why It's Good For The Game
A common occurrence when autobalance happens is when xenos join in at the start of the check, which *would* disable autobalance, but because they joined after the check is doesn't change. This leads to 3-4 minutes of autobalanced xenos when they really shouldn't be. This hopes to reduce that. If it is fine to move down to two minutes, please tell me; this is just being on the cautious side.

## Changelog

:cl:
balance: Autobalance check now occurs every 3 minutes instead of every 5
/:cl:

### DIFF
--- a/code/controllers/subsystem/monitor.dm
+++ b/code/controllers/subsystem/monitor.dm
@@ -2,7 +2,7 @@ SUBSYSTEM_DEF(monitor)
 	name = "Monitor"
 	init_order = INIT_ORDER_MONITOR
 	runlevels = RUNLEVEL_GAME
-	wait = 5 MINUTES
+	wait = 3 MINUTES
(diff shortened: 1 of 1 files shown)


Pariah919:
Why is autobalance even enabled past the first 30 minutes
Its very silly to watch it trigger into a *two hour match*.

TiviPlus (approved):

DeltaFire15:
Uh I'm a bit late on this PR, but I'm pretty sure the "stalemate" checks check for x *monitor cycles* before it declares things a stalemate, so maaay want to touch those.

PR #13968, opened 2023-09-17
Gau hits twice again

ivanmixo:
## About The Pull Request

Yes. Revert of https://github.com/tgstation/TerraGov-Marine-Corps/pull/13735, but I add a qdeleted check instead.
## Why It's Good For The Game

Eh this wasn't really intented, plus gau is already kinda bad so halving its damage makes it sad.
## Changelog
:cl:
balance: Gau hits twice again
/:cl:

### DIFF
--- a/code/game/objects/structures/dropship_ammo.dm
+++ b/code/game/objects/structures/dropship_ammo.dm
@@ -254,6 +254,10 @@
 		strafelist -= strafed
 		strafed.ex_act(EXPLODE_LIGHT)
 		new /obj/effect/temp_visual/heavyimpact(strafed)
+		for(var/atom/movable/AM AS in strafed)
+			if(QDELETED(AM))
+				continue
+			AM.ex_act(EXPLODE_LIGHT)
 
 	if(length(strafelist))
 		addtimer(CALLBACK(src, PROC_REF(strafe_turfs), strafelist), 2)
--- a/code/game/objects/structures/dropship_ammo.dm
+++ b/code/game/objects/structures/dropship_ammo.dm
@@ -257,7 +257,8 @@
 		for(var/atom/movable/AM AS in strafed)
 			if(QDELETED(AM))
 				continue
-			AM.ex_act(EXPLODE_LIGHT)
+			//This may seem a bit wacky as we're exploding the turf's content twice, but doing it another way would be even more wacky because of how hard it is to modify explosion damage without adding a whole other explosion type
+			AM.ex_act(EXPLODE_LIGHT) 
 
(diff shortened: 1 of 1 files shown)


dopamiin0:
YEAH BABY CAS IS PLAYABLE AGAIN WOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO THATS WHAT IVE BEEN WAITING FOR THATS WHAT ITS ALL ABOUT. IM GONNA BE TELLING MY SPOUSE AND KIDS ABOUT THIS ONE IN THE FUTURE, THE DAY THAT TGMC WAS RESTORED TO ITS FORMER GLORY

Lumipharon (changes requested):
Readding a funny shitcode version of ex_acting the turf twice instead of just buffing its effects seems a bit uh, weird.

i.e. Heavy explosion instead of light is 50% more damage, double sunder and whatever more stagger/slow compared to light. 
Then it also suffers from extremely high travel time (6 seconds base, wtf), which makes it much harder to use, regardless of its damage

I feel like touching those would be better than this.

ivanmixo:
I'm not giving gau a chance to instadelete walls, lumi

Lumipharon:
Good point. At least add a comment or something saying this is intended though, as its a really funny method.

code/game/objects/structures/dropship_ammo.dm
```diff
@@ -254,6 +254,10 @@
 		strafelist -= strafed
 		strafed.ex_act(EXPLODE_LIGHT)
 		new /obj/effect/temp_visual/heavyimpact(strafed)
+		for(var/atom/movable/AM AS in strafed)
+			if(QDELETED(AM))
+				continue
+			AM.ex_act(EXPLODE_LIGHT)
```
ivanmixo, line 260:
```suggestion
			//This may seem a bit wacky as we're exploding the turf's content twice, but doing it another way would be even more wacky because of how hard it is to modify explosion damage without adding a whole other explosion type
			AM.ex_act(EXPLODE_LIGHT) 
```

iforgotmeword:
would it hurt if cas was given heavy explosion for HV ammo?

dopamiin0:
> would it hurt if cas was given heavy explosion for HV ammo?

highvel is already one of the best purchases available to a cas pilot, i don't really see why it needs a buff

PR #13969, opened 2023-09-17
Adds roundstart heavy weapon rotation.

Pariah919:
## About The Pull Request
Adds heavy weapon rotation to roundstart kit.
Effectively, the TAT and RR can get replaced by other stuff, everything totals up to an imaginary power level of like 12, with the TAT being 6 and an RR being 3.

Current buckets are listed down below, name of the tab is "Operational Weapons". Rerolls every 24.

<details>

- Default - Which is the current one of just TAT and 2 RRs.
- Fire and Forget - TAT, 8 disposables, and 2 Thermobarics.
- Autorail - Flak gun and Railgun (With no HVAP!)
- Shock - 2 RRs and an MLRS

</details>

## Why It's Good For The Game
Adds more variety to roundstart marine rollouts and thoughts, gives 40-60 point range weapons that are underused some more time in the spotlight and lets us test new roundstart marine powerkit rollouts to see how it affects the game, which is the main thing I'm interested in with this PR.

Boring design process/thoughts blarb if you care at all below.

<details>
I also briefly considered rolling other stuff like flamers, HMG/HSG into this "Operational rotation" but I eventually realized that marines with 0 flamers roundstart would probably not be a very good idea, and there's not enough variety in possible artillery pieces to have any more interesting loadouts beyond the mortar+howi we already have, so this mainly aims at TAT and RR, maybe in the future other stuff could be rolled into this.

Another train of thought I maybe thought of was making things be "Equivalent" I.E Railgun and RR replace eachother in loadouts, and generally structuring everything but I eventually realized that would be probably a bad idea, I also considered adding a BR8/Autosniper bin but I don't have enough political power to convince the playerbase that those are fine roundstart inexchange for no TAT or RR. (Even though I think it would!)

~~I also meant for this to be my birthday joke PR but it became serious so uhhhhhhh.~~

Funny chart of power below, power in general is entirely subjective. Ideally most bins should equal 12 or so. Everyone will disagree here.
image

</details>

## Changelog

:cl:
add: Rotational heavy weapons system, the tab itself ingame is called 'Operational Weapons' which rotates every 24 hours between bins like seasonals, TAT and RR have been moved to here and the 3 other bins contain objects like Flak, Thermo, and DRPGs.
/:cl:

### DIFF
--- a/code/controllers/subsystem/persistence.dm
+++ b/code/controllers/subsystem/persistence.dm
@@ -18,7 +18,7 @@ SUBSYSTEM_DEF(persistence)
 	///Stores how long each season should last
 	var/list/seasons_durations = list(
 		SEASONAL_GUNS = 24 HOURS,
-		SEASONAL_HEAVY = 24 HOURS,
+		SEASONAL_HEAVY = 8 HOURS,
 	)
 	///Stores the current season for each season group
 	var/list/season_progress = list()
@@ -40,6 +40,7 @@ SUBSYSTEM_DEF(persistence)
 		/datum/season_datum/weapons/guns/heavy_defualt,
 		/datum/season_datum/weapons/guns/heavy_ff,
 		/datum/season_datum/weapons/guns/heavy_autorail,
+		/datum/season_datum/weapons/guns/heavy_shock,
 		),
 	)
 	///The saved list of custom outfits names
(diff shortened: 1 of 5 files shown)


Mrrpip:
Hm...I am a bit skeptical about the MLRS/Hwatcha being a possible pool option, as it's an indirect fire weapon.

iforgotmeword:
should captains and/or ground commanders be able to force a loadout on a round cooldown?
like if he has a plan of some kind that NEEDS TAT he/she can force it?

for how many times i say 1 per 8 hour cycle

Pariah919:
Swappe it over to 24 hour cycles after discord suggestions.

yyzsong:
bad pr. Muh RR.

PR #13971, opened 2023-09-17
King no longer has additional special autobalance weight

Victor239:
## About The Pull Request

King had special autobalance weight since the caste used to work mechanically as a "game ender" that cost a lot of psy points, now instead it's just another T4 that is pop-locked so it has the same weight as other T4s.

## Why It's Good For The Game

King used to weigh up to 350% of a Queen for autobalance calculations, this hasn't been updated since King got reworked into a much weaker pop-locked T4 caste.

## Changelog

:cl:
del: King no longer has additional special autobalance weight.
/:cl:

### DIFF
--- a/code/__DEFINES/monitor.dm
+++ b/code/__DEFINES/monitor.dm
@@ -35,7 +35,6 @@
 #define PRIMO_T3_WEIGHT -100
 #define NORMAL_T4_WEIGHT -80
 #define PRIMO_T4_WEIGHT -120
-#define KING_WEIGHT -200 //king is also counted as a T4
 #define SPAWNING_POOL_WEIGHT -100
 
 //The weight of each statistics in the state calculator before shutters drop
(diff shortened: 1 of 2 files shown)


ivanmixo:
Autobalance already triggers way too often when it shouldn't and this will only make it worse

Victor239:
> Autobalance already triggers way too often when it shouldn't and this will only make it worse

That would be a separate issue, this is addressing specifically King being a snowflake caste in regards to autobalance. I'm not sure exactly what you mean by triggering "when it shouldn't" though, but if that's the case I'd suggest tweaking the algorithm it uses.

Arturlang:
Kings don't win games on their own, and they're slow enough to be easily focused down, nowhere near as dangerous as a mech.

Lumipharon:
Yeah autobal being wack is its own problem

PR #13973, opened 2023-09-18
Fix xeno spawns not counting towards autobalance

forestbbbbbbb:
## About The Pull Request

These were being decremented but never incremented so xeno deaths were actually counting *towards* turning autobalance on

## Changelog

:cl:
fix: Fix xeno spawns not counting towards autobalance
/:cl:


Barnet2:
MFW I was messaged about this 10 minutes ago, I start testing, and you already fix it. Actual mad man

PR #10055, opened 2022-04-24
Change most guns nomenclature

BraveMole:
## About The Pull Request

Rather than T-something, or having something random, the letters follow a regular nomenclature :

AR : assault rifle
SR : sniper rifle
R : revolver
FL : flamer
P : pistols
SH : shotguns
RL : rocket launcher
MG : machine gun
HMG : heavy machine gun
SG : smartgun
HSG : heavy smartgun
SMG : submachine gun

## Why It's Good For The Game

The names right now are confusing as fuck, even for old players. So imagine for new ones ...
There will be an adapation time, but I think on the long run this is much better

## Changelog
:cl:
qol: Guns' name letter now follow a nomenclature.
/:cl:

PR #13979, opened 2023-09-19
Reverts Gun Nomeculture (Mostly) entirely.

Pariah919:
## About The Pull Request
This reverts gun nomeculture changes from #10055 for marine and non-marine guns, outside of Sentries which stay ST.
## Why It's Good For The Game
This change has not worked for the last year, if anything it has made everything more confusing because it changed *non marine weapons for no discernable reason beyond "I think its less confusing" and at absolute random.* 
Which causes you be to be unable to tell which gun is which on a per faction basis because the gun class was assigned at seemingly random, now that we have gun manufacturers on most gun names now which should make most guns easier to identify even with the T-## for marines so this confusing and utterly worthless gun class system is even more irrelevant than ever, it changed nothing but cause confusion for veteran and newer players who use different naming schemes (or for the vast of them I talk to, get consolidated into the former.)

I kept the changes on a few things like sentry guns since I think those are better with gun class than with T-###, outside of that the only other thing I kept was the SVD rename from the top of my head.
## Changelog
:cl:
spellcheck: Reverted gun nomeculture for marine guns to T-###, reverted gun names for non-marine guns to pre gun class change names.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -183,7 +183,7 @@
 	starting_attachment_types = list(/obj/item/attachable/stock/t12stock, /obj/item/attachable/magnetic_harness, /obj/item/attachable/extended_barrel, /obj/item/weapon/gun/grenade_launcher/underslung)
 
 //-------------------------------------------------------
-//DMR-37 DMR
+//T-37 DMR
 
 /obj/item/weapon/gun/rifle/standard_dmr
 	name = "\improper T-37 SCA designated marksman rifle"
@@ -265,7 +265,7 @@
 
 /obj/item/weapon/gun/rifle/standard_br
 	name = "\improper T-64 SCA battle rifle"
-	desc = "The San Cristo Arms BR-64 is the TerraGov Marine Corps main battle rifle. It is known for its consistent ability to perform well at most ranges, and medium range stopping power with bursts. It is mostly used by people who prefer a bigger round than the average. Uses 10x26.5smm caseless caliber."
+	desc = "The San Cristo Arms T-64 is the TerraGov Marine Corps main battle rifle. It is known for its consistent ability to perform well at most ranges, and medium range stopping power with bursts. It is mostly used by people who prefer a bigger round than the average. Uses 10x26.5smm caseless caliber."
 	icon_state = "t64"
 	item_state = "t64"
 	icon = 'icons/Marine/gun64.dmi'
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -23,11 +23,11 @@
 	lower_akimbo_accuracy = 3
 
 //-------------------------------------------------------
-//AR-18 Carbine
+//T-18 Carbine
 
 /obj/item/weapon/gun/rifle/standard_carbine
-	name = "\improper AR-18 Kauser carbine"
-	desc = "The Keckler and Hoch AR-18 carbine is one of the standard rifles used by the TerraGov Marine Corps. It's commonly used by people who prefer greater mobility in combat, like scouts and other light infantry. Uses 10x24mm caseless ammunition."
+	name = "\improper T-18 Kauser carbine"
+	desc = "The Keckler and Hoch T-18 carbine is one of the standard rifles used by the TerraGov Marine Corps. It's commonly used by people who prefer greater mobility in combat, like scouts and other light infantry. Uses 10x24mm caseless ammunition."
 	icon_state = "t18"
 	item_state = "t18"
 	fire_sound = 'sound/weapons/guns/fire/tgmc/kinetic/gun_ar18.ogg'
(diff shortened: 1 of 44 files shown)


00ze-cyclone:
The guns of the marines are the focus, all the other factions barely exist, why do we even need to distinguish guns on a faction basis?

Texicos:
T series is back

breadhunt:
> The guns of the marines are the focus, all the other factions barely exist, why do we even need to distinguish guns on a faction basis?

^
Also I hear the nomecultures a ton more than T- but that might just be me.

code/modules/projectiles/guns/rifles.dm
```diff
@@ -453,10 +453,10 @@
 
 
 //-------------------------------------------------------
-//PR-11
+//HK-11
 
 /obj/item/weapon/gun/rifle/m41a
-	name = "\improper PR-11 pulse rifle"
+	name = "\improper HK-11 pulse rifle"
```
XSlayer300, line 459: You know, we already have the TX-11...

code/modules/projectiles/guns/rifles.dm
```diff
@@ -1642,7 +1642,7 @@
 
 /obj/item/weapon/gun/rifle/alf_machinecarbine
 	name = "\improper ALF-51B Kauser machinecarbine"
-	desc = "The Kauser ALF-51B is an unoffical modification of a ALF-51, or better known as the AR-18 carbine, modified to SMG length of barrel, rechambered for a stronger round, and belt based. Truly the peak of CQC. Useless past that. Aiming is impossible. Uses 10x25mm caseless ammunition."
+	desc = "The Kauser ALF-51B is an unoffical modification of a ALF-51, or better known as the T18 carbine, modified to SMG length of barrel, rechambered for a stronger round, and belt based. Truly the peak of CQC. Useless past that. Aiming is impossible. Uses 10x25mm caseless ammunition."
```
XSlayer300, line 1645:
```suggestion
	desc = "The Kauser ALF-51B is an unoffical modification of a ALF-51, or better known as the T-18 carbine, modified to SMG length of barrel, rechambered for a stronger round, and belt based. Truly the peak of CQC. Useless past that. Aiming is impossible. Uses 10x25mm caseless ammunition."
```

code/modules/projectiles/guns/smgs.dm
```diff
@@ -27,11 +27,11 @@
 	movement_acc_penalty_mult = 3
 
 //-------------------------------------------------------
-// MP-19 Machinepistol. It fits here more.
+// TP-19 Machinepistol. It fits here more.
 
 /obj/item/weapon/gun/smg/standard_machinepistol
-	name = "\improper MP-19 machinepistol"
-	desc = "The MP-19 is the TerraGov Marine Corps standard-issue machine pistol. It's known for it's low recoil and scatter when used one handed. It's usually carried by specialized troops who do not have the space to carry a much larger gun like medics and engineers. It uses 10x20mm caseless rounds."
+	name = "\improper T-19 machinepistol"
```
XSlayer300, line 33: TP-19 or T-19?

code/modules/projectiles/guns/rifles.dm
Pariah919: Its was always HK11.

code/modules/projectiles/guns/smgs.dm
Pariah919: Was always T-19, from my memory.

Otterwatch: It left the code as T-19 to MP-19.

Wisemonster:
tbf, weapons have what gun they are at the end of their description (for example, AR-55 would appear as TX-55 assault rifle on exampine), though I assume most people will just say the weapon pre-fix and number rather than the whole name ("I need some AR-55 ammo" vs. "I need TX-55 ammo") due to being faster to type.

chesse20:
unneeded change imo

QualityVan:
I think having class names consistent across marine weapons is basically a good thing, particularly when you're communicating stuff about a gun to someone who doesn't know it already. It's a lot easier for eg the RO to find ammo to send if he's got a weapon class implicit from the name. If they're inconsistent right now that should be addressed, although I do agree that it's a lot less important for non-terragov stuff since most of that is fixed loadouts.

AzonStarfall:
This has been said, but changing the naming scheme to only differentiate weapons by a number, and not category+number, definitely hinders the identification process. With the current schema, it's easy to determine the weapon class and use case of any given item, without having its specific nametag memorized. As a new player looking at T-18 and T-19 without any frame of reference, there's no way to reasonably guess what either weapon would be used for, without testing each personally. In the current nomenclature, AR-18 and MP-19, having basic knowledge of what the prefix means would lead players to a general assumption of the weapons capability, with minimal prior experience with either weapon class and without knowing the specific weapon number designation within said class.

As it stands, reverting to the T-### naming system demands that a player is intimately aware with each weapon's specific number for even the most general identification, and can easily hinder the experience of players unaware with this system.

TiviPlus:
No

PR #11388, opened 2022-10-31
Drowsiness sleep chance changed to only activate after a certain stack threshold

lKiseki:
## About The Pull Request
Hey! You! Yes, you. Come on over, let me show you something really funny...

So I noticed that Boiler globs have a var called `hit_drowsyness`, which is the amount of drowsiness stacks applied on a direct hit. A normal acid glob applied one stack of drowsiness. A lance acid glob, applied 18. A neuro lance glob applied 2.

So what does drowsiness do, you wonder? I am so glad you asked...
Every tick, if you have a stack of drowsiness, it'll blur your eyes and you'll lose a stack. However, _**it also has a 5% chance to make you fall asleep and unconscious**_. This was taken from `/mob/living/carbon/handle_status_effects()`:
image

What does this mean for a poor, poor marine? After being hit directly by a Boiler glob, there's a 5% chance they'll just fall asleep and unconscious as long as they have drowsiness stacks. Fun and emergent gameplay.

This also applied to other sources of drowsiness, like beer.

## Why It's Good For The Game
Why in god's name can a Boiler glob make you fall asleep and unconscious at a 5% chance? That's just unfun.
This also extended to other sources of drowsiness, so I've changed that. One sip of beer is probably not gonna make you fall asleep instantly.

## Changelog
:cl: Lewdcifer
balance: Drowsiness sleep chance will only activate after a certain stack threshold.
/:cl:

PR #13995, opened 2023-09-21
Chemerette tweaks

Looking4bros:
## About The Pull Request

Tweaks Chemerettes, Neurokiller now has Serotrotium to better fit the description that it may tire the smoker, lowered its chem volume and transfer rate so the chems are only in the smokers system while they're being smoked. Bicaridine and Kelotane now do the same instead of dumping 5u into the smoker for 6 ticks. Smoke time has been bumped up one for each as they were not using all the chems beforehand. (use the RR cig and see only 9 cloneloss)
## Why It's Good For The Game

Good because it makes existing thing work more like what the descriptions say they do.

## Changelog

:cl:
balance: Chemerette tweaks
/:cl:

### DIFF
--- a/code/game/objects/items/tools/flame_tools.dm
+++ b/code/game/objects/items/tools/flame_tools.dm
@@ -379,10 +379,10 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	icon_state = "anticigoff"
 	item_state = "anticigoff"
 	icon_on = "anticigon"
-	smoketime = 30
-	chem_volume = 60
-	transquantity = 2 // one of each for the whole duration
-	list_reagents = list(/datum/reagent/medicine/ryetalyn = 30, /datum/reagent/water = 30)  //some water so it purges the rye too
+	smoketime = 31
+	chem_volume = 6
+	transquantity = 0.2
+	list_reagents = list(/datum/reagent/medicine/ryetalyn = 3, /datum/reagent/serotrotium = 3)
 
 /obj/item/clothing/mask/cigarette/emergency
 	name = "Red Comrade"
@@ -390,7 +390,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	icon_state = "rrcigoff"
 	item_state = "rrcigoff"
 	icon_on = "rrcigon"
-	smoketime = 10
+	smoketime = 11
 	transquantity = 1
 	list_reagents = list(/datum/reagent/medicine/russian_red = 10)  //same ammount as a pill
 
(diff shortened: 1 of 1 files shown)


Looking4bros:
Tested on local.

iforgotmeword:
uhhh why? isn't dylo better?

Looking4bros:
Dylo is ASS!!!!!!!!!!!!!!!!1!!!!!! It purges Paracetamol, penalizes stamina recovery, and combines with Synaptizine to make toxin.

Looking4bros:
As well, Rye purges at the rate of 5u, where Dylo purges at 1u. big difference, and why i think adding Serotrotium to the cig is better than water, as it does have a chance to apply slowdown via Drowsyness. Rye is a T2 chem.

Al-1ce:
Serotrotium has a low-chance proc that puts the imbiber to sleep, which is very bad if you're in the field.
I think water is fine. It does what it needs to.

Looking4bros:
Serotrotium on Critical OD (50u) does have a chance to put the user to sleep. While it is in your system it has a 2% chance to apply 5 drowsy, which gives slowdown. Smoking it as this is coded you will only have .1u in your system.
Water purges indiscriminately, which is also bad in the field, it was only implemented to purge the Rye as it has a metabolism rate of .025, which would mean you would have an infinite amount of free T2 chem. The ALTERNATE way of doing it is setting the transfer rate low enough that it gets purged out of the system each tick, which .1 is that transfer rate. 
Ryetalyn purges toxins at a rate of 5u, the same as Hypervene, you don't need the added 1u that water provides.

carlarctg:
> Serotrotium on Critical OD (50u) does have a chance(10%) to put the user to sleep. Smoking it as this is coded you will only have .1u in your system. While it is in your system it has a 2% chance to apply 5 drowsy, which gives slowdown, you need >18 for a (5%)sleep/unconscious chance, Drowsy is reduced by 3 each tick, with an additional 12 if you are resting. you would have to get the 2% chance proc to happen 8 times in a row for it to have a chance by itself to get the opportunity to roll the 5% chance of Sleep/unconscious
> 
> Water purges indiscriminately, which is also bad in the field, it was only implemented to purge the Rye as it has a metabolism rate of .025, which would mean you would have an infinite amount of free T2 chem. The ALTERNATE way of doing it is setting the transfer rate low enough that it gets purged out of the system each tick, which .1 is that transfer rate. Ryetalyn purges toxins at a rate of 5u, the same as Hypervene, you don't need the added 1u that water provides.

Uh just make a ryetalyn subtype that has a faster metabolization rate then???? Like what.

TiviPlus:
@QualityVan

QualityVan (changes requested):
Sero can't actually sleep you any more due to changes with how drowsiness stacks work (see #11388 ) so that's not an issue. The chem volume change make a lot sense thematically, although it's a hefty nerf to total effect per cig. The smoketime changes feel intuitively like the order of operations somewhere is off, and that should be fixed itself rather than adjusting times to compensate.

PR #14002, opened 2023-09-22
Superheavy pistol balanc pass

Pariah919:
## About The Pull Request
Nerfs stagger from 2 to 0.5. Also nerfs the firedelay by 0.1 on all of them.
Hipower does 1.1 rather than 1.2, scatters a bit more and has natural recoil.
## Why It's Good For The Game
Makes super heavy pistols more in line with stagger on hit and general damage.
## Changelog
:cl:
balance: Another balance pass on superheavy pistols, their stagger has been significantly reduced, alongside their fire rate being minorly made worse (0.05 more seconds, so .35 for Deagle and .4 for automag.) Automag does a bit less damage and has a bit more scatter.
/:cl:


Mrrpip:
Hmm...

I'm thinking, because it copies the default pistol attachie set. Perhaps the Automag should have it's own curated set?

As it's competing with the BFA (Which is particularly meh on the automag) and Laser sight, the Gyro would likely be a no-brainer in most cases. Maybe.

ivanmixo (approved):

PR #14005, opened 2023-09-23
Fix backpack prefs overriding job backpacks

Blundir:
## About The Pull Request

Fixes backpack prefs overriding the actual specified backpacks in job datums, because of which ST and CE never spawn with a their back fuel tanks for example. Also made it possible to specify having no ID slot item and no backpack slot item at all, right now it's impossible, if you set ID to null it'd still spawn, if you set backpack to null it'd just spawn the one from prefs anyway.

## Why It's Good For The Game

Bugs - bad, fix - good, flexible datums - even better.

## Changelog
:cl:
fix: id card having no mob icon when spawning on mob
fix: fixed backpack prefs overriding spawn of job assigned backpacks
code: made it possible to specify to spawn without backpack or id in job datums
/:cl:

### DIFF
--- a/code/datums/jobs/job/job.dm
+++ b/code/datums/jobs/job/job.dm
@@ -310,6 +310,8 @@ GLOBAL_PROTECT(exp_specialmap)
 
 		equip_role_outfit(job)
 
+		regenerate_icons() //some items lack icons unless updated
+
 	if((job.job_flags & JOB_FLAG_ALLOWS_PREFS_GEAR) && player)
 		equip_preference_gear(player)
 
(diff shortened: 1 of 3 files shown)


code/datums/jobs/job/job.dm
```diff
@@ -289,17 +289,29 @@ GLOBAL_PROTECT(exp_specialmap)
...
+				if(BACK_SATCHEL)
+					new_backpack = new /obj/item/storage/backpack/marine/satchel(src)
+			equip_to_slot_or_del(new_backpack, SLOT_BACK)
+
+		job.outfit.handle_id(src, player)
 
 		equip_role_outfit(job)
 
+		regenerate_icons() //some items lack icons unless updated
+
```
TiviPlus, line 314: Should not equip_to_slot_or_del be handling this

Blundir: it should, but for some reason some icons just refuse to update

Lumipharon: should probably uh, debug why that happens then

TiviPlus: yea just remove this line since its a bandaid rest is fine

Lumipharon (approved):

PR #14013, opened 2023-09-24
Reduces medivac bed price in corpsmen vendor

lbnesquik:
## About The Pull Request
Price goes from 45 to 20, which is about half of a corpsman budget. Still expensive, but less demanding now.

## Why It's Good For The Game
The medivac beds are way overpriced, actually. It's not worth skimping on gloves, mera, derma, meladerm or the rest. This PR makes it possible to squeeze the extra without sacrificing everything.

## Changelog
:cl:
balance: Made medivac beds ~50% cheaper for corpsmen
/:cl:


novaepee:
With this buff, a few corpsmen can teleport people faster than droppods.

lbnesquik:
> With this buff, a few corpsmen can teleport people en masse traveling between shipside and groundside faster than droppods.

50% more than before, which hasn't happened yet as far as I can tell.

Lumipharon:
This was only added because it used up all your points.

PR #14016, opened 2023-09-24
Makes Gelida IV a HvH only map

Victor239:
## About The Pull Request

Removed Gelida IV ground map from XvH - now votable on HvH only.

## Why It's Good For The Game

Gelida has a number of flaws that mean it doesn't reach the same standards of other maps in rotation for XvH:
- Heavily lacks preweed, to the point that if you don't have 5+ builders you're very unlikely to be able to sufficiently preweed and use up your quickbuild points. 
- Platforms are unfun since the perspective makes it difficult to distinguish them, and on this map they're often placed infront of windows which makes it them act like invulnerable window frames. Additionally for xenos to hit-and-run attack from many areas you have to jump across at least 2 platforms, and due to the jump cooldown this means xenos can't do that effectively in these areas. Some solutions are make them meltable, add way more stairs to buildings, make them not dense (cosmetic-only), reduce amount of platforms, etc.
- There's at least one dead-end enclosed area that is "full wipe bait" unlike other maps which have better flow. 
- Caves are small and there are only a few pathways.
- FOBs have few angles to siege, and no way to flank assault them. 

If someone wants to rework this map then it can be easily readded to XvH, but until then it should only be votable in HvH. 

## Changelog

:cl:
del: Removed Gelida IV ground map from XvH - now votable on HvH only.
/:cl:


Barnet2:
These aren't even valid reasons to remove a map from rotation. The reasons you have listed are easily fixable. I hate Gelida too, but if you wager for a removal you should argue on how it is fundamentally designed if you hope for it to be accepted. -1 from me.

Victor239:
Agreed it's fixable, but the map has been around for a long time and still hasn't been fixed. Since no-one wants to maintain it then it should be removed until that situation changes.

RipGrayson (commented):
You know I'm sympathetic to your position, Gelida isn't a good map and it does need adjusting. I think going for a full removal is overkill though.
I'm not going to outright veto this PR though, I want to hear what other players have to say first.

lbnesquik:
Honestly 2/3 of those issues are very easily fixed, leaving one decent reason which is not enough to can a whole map in my opinion. And removing platforms is also a clear, easy option that doesn't chuck an entire working map into the trash.

Don't merge this

Lumipharon:
This seems like extreme cope. Also you can just jump platforms now, and they're a distinct hazard stripe so they're not exactly hard to miss.

Victor239:
@RipGrayson fair enough. And I appreciate that you did rework Gelida earlier this year, but I think for now it's still not at the stage where keeping it in is advantageous compared to not until it's reworked some more.

@lbnesquik I've already agreed it's potentially fixable, but if nobody is willing to or has time to do so right now then the end result is the same. And just because it's out of the codebase for now doesn't mean it has to be gone permanently, it could be very well be in the near future that a maptainer fixes these issues.

@Lumipharon I've heard plenty of complaints from both xenos and marines about the aforementioned map issues making the game unfun, you can ask them yourself if you don't believe me. I just personally don't see the same advantage of keeping in a disliked map just to increase the pool of available maps.

lbnesquik:
> @lbnesquik I've already agreed it's potentially fixable, but if nobody is willing to or has time to do so right now then the end result is the same. And just because it's out of the codebase for now doesn't mean it has to be gone permanently, it could be very well be in the near future that a maptainer fixes these issues.
> 

I don't think that's the same thing. It's throwing away a good house because you don't like the furniture, it needs some painting work and a wooden floor redone. We did spend a lot of collective effort making it better. Chucking it into the trash because it lacks weeds, you don't like railings and hate 1 area is wasting all of it. 

If my desktop computer PSU wasn't fried, I could honestly tackle all 3 of those issues in 2 hours tops. And so could anyone who took a stab at it. But it's much less likely to happen if the thing's already done and gone. 

Also, none of those things are current issues, so it's not like it was properly signaled to be a problem.

TiviPlus:
Regardless of merit fails CI

RipGrayson:
Still thinking this one over, less against it that I was when the PR was first opened

RipGrayson (changes requested):
Alright, I've come to a decision. You can remove it from XvH, but it should be kept for HvH since other maintainers need it.

Victor239:
@RipGrayson done.

PR #14017, opened 2023-09-24
Somewhat cursed lateload z-level minimap hard del fix

Lumipharon:
## About The Pull Request
Fucking guh. Fixes the funny hard del.

Please someone suggest a less rediculous way to do this.

### DIFF
--- a/code/controllers/subsystem/minimaps.dm
+++ b/code/controllers/subsystem/minimaps.dm
@@ -226,7 +226,7 @@ SUBSYSTEM_DEF(minimaps)
 		if(!(earlyadds["[target.z]"]))
 			earlyadds["[target.z]"] = list()
 		earlyadds["[target.z]"] += CALLBACK(src, PROC_REF(add_marker), target, hud_flags, blip)
-		RegisterSignal(target, COMSIG_QDELETING, PROC_REF(remove_earlyadd))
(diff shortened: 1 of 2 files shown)


spookydonut:
a bit weird but it works

spookydonut (approved):

code/controllers/subsystem/minimaps.dm
```diff
@@ -226,6 +226,7 @@ SUBSYSTEM_DEF(minimaps)
 		if(!(earlyadds["[target.z]"]))
 			earlyadds["[target.z]"] = list()
 		earlyadds["[target.z]"] += CALLBACK(src, PROC_REF(add_marker), target, hud_flags, blip)
+		RegisterSignal(target, COMSIG_QDELETING, PROC_REF(remove_earlyadd), override = TRUE) //if it was loaded before the z-level
```
TiviPlus, line 229: This never gets unregistered if its processed normally is that intentional

TiviPlus: Also you need to write a detailed explanation for using override = true

Lumipharon: when the zlevel finished loading, the signal gets overridden

TiviPlus (approved):

PR #14028, opened 2023-09-26
Reduces number of widow's spiderlings to 4 from 5

Barnet2:
## About The Pull Request
Read title.
## Why It's Good For The Game
Widow's damage is a bit silly, especially considering the maturity rework making her always have the ancient amount of spiderlings. This caste's design is already very questionable, having even more insane burst damage on average is not great.
## Changelog
:cl:
balance: Widow's max spiderlings is now 4 instead of 5.
/:cl:


ivanmixo:
Very boring way to nerf this which doesn't even address the actual issues with it.

PR #14029, opened 2023-09-26
Slightly nerfs king pheros

Barnet2:
## About The Pull Request
Takes king phero strength from 6 down to 5.5
## Why It's Good For The Game
Another issue with the maturity rework was perma ancient king pheros, which were fine with maturity on the board, but now are a bit overwhelming considering that they are always on warding, giving a good amount of armor to benos.
## Changelog
:cl:
balance: King pheros strength reduced to 5.5 from 6
/:cl:


ivanmixo:
What's even the point of such a small change? Cmon now I really don't think this is an issue.

DeltaFire15:
> What's even the point of such a small change? Cmon now I really don't think this is an issue.

For most castes, magnitude 6 warding is a more powerful DR than a permanently active FC hold order (post nerf), and for some, even pre-nerf.

PR #14054, opened 2023-09-29
Funny minimap fix

Lumipharon:
## About The Pull Request
Fixes the minimap not tracking you while inside objects (i.e. mechs/bodybags).

I put this in the campaign pr because this bug became VERY apparent during mech war. This should be a separate pr though.

Also this solution is kinda scuffed, there's probably a better solution but eh.

Added the line in mech code because it would forcemove you into the mech twice.
## Why It's Good For The Game
Minimap work good.
## Changelog
:cl:
fix: Minimap properly tracks you inside objects like mechs or bodybags
/:cl:

### DIFF
--- a/code/controllers/subsystem/minimaps.dm
+++ b/code/controllers/subsystem/minimaps.dm
@@ -299,9 +299,34 @@ SUBSYSTEM_DEF(minimaps)
  */
 /image/proc/minimap_on_move(atom/movable/source, oldloc)
 	SIGNAL_HANDLER
-	var/turf/source_turf = get_turf(source)
-	pixel_x = MINIMAP_PIXEL_FROM_WORLD(source_turf.x) + SSminimaps.minimaps_by_z["[source_turf.z]"].x_offset
-	pixel_y = MINIMAP_PIXEL_FROM_WORLD(source_turf.y) + SSminimaps.minimaps_by_z["[source_turf.z]"].y_offset
+	if(isturf(source.loc))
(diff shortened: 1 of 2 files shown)


code/controllers/subsystem/minimaps.dm
```diff
@@ -299,9 +299,34 @@ SUBSYSTEM_DEF(minimaps)
...
-	pixel_y = MINIMAP_PIXEL_FROM_WORLD(source_turf.y) + SSminimaps.minimaps_by_z["[source_turf.z]"].y_offset
+	if(isturf(source.loc))
+		pixel_x = MINIMAP_PIXEL_FROM_WORLD(source.x) + SSminimaps.minimaps_by_z["[source.z]"].x_offset
+		pixel_y = MINIMAP_PIXEL_FROM_WORLD(source.y) + SSminimaps.minimaps_by_z["[source.z]"].y_offset
+		return
+
+	if(!ismovableatom(source.loc))
+		return
+
+	var/atom/movable/movable_loc = source.loc
```
ivanmixo, line 310: pointless cast, source is already an atom/movable here

ivanmixo (changes requested):
Otherwise lgtm

code/controllers/subsystem/minimaps.dm
Lumipharon: we're typecasting loc though, not source

Lumipharon:
Closing this as more up to date code is in the campaign pr

PR #14056, opened 2023-09-30
Fix a Big Red modular with incorrect height

vvvv-vvvv:
## About The Pull Request

Don't know why it was 29x25 when it's actually 29x24 in reality, seems to work fine ingame though

## Changelog

no player-facing changes


RipGrayson (approved):
I don't know how you figured out the dimensions were off by one but you're correct.

PR #14059, opened 2023-10-01
Adds Succumb for Mobs in hardcrit

efzapa:
## About The Pull Request

Allows for mobs to succumb when in hardcrit threshold
(also DNM for now until i get crit thresholds and code actually working, opening right now purely for feedback (and help))

## Why It's Good For The Game

This was honestly made as a way to balance the plague for marines that is critdragging. As it stands (not even counting warrior/shrike fling stunlock combos), you can get dragged anywhere from 5-20 tiles along depending on how damaged you were, meds, etc.
This should not be a thing to abuse so that xenos can effectively just domino marines that try and save other marines like 20-25 tiles in hive, we already have stunlocks and shuffling which is bad enough.

## Changelog

:cl:
add: Added Succumb, lets you die on command in crit.
/:cl:


Xander3359:
honestly I would prefer just having soft/hardcrit thresholds and allow crawling than this tbh.

efzapa:
> honestly I would prefer just having soft/hardcrit thresholds and allow crawling than this tbh.

that would be fine but one issue I forsee is marine dying, going into harder it, then just crawling to cades or otherwise an open space

ivanmixo:
You are not removing crit drag, it is very important for allowing xenos to actually secure kills.

PR #14060, opened 2023-10-01
Removes Vali's chemical boost multiplier option

lKiseki:
## About The Pull Request
This PR removes the option to change the Vali's chemical boost to x2, meaning it will always be on the default (x1). This is, in fact, a nerf to the armor module.
I have taken the liberty of clearing away unnecessary code to account for this change.
This also changes the resource drain value, from 4 at x1 to 8. The new value is what the drain on x2 normally was.

## Why It's Good For The Game
Vali allows marines to get away with far much more than they should for the sake of balance health. Throwing yourself into hives, surviving multiple xenos like nobody's business, and being able to overdo xenos in melee is a bit _too_ much.

This is my attempt to nerf this package slightly by reducing the regeneration it offers, as opposed to tackling the Vali weaponry or anything else.

## Changelog
:cl: Lewdcifer
del: Removed the option to change Vali's chemical boost, effectively removing the x2 boost.
balance: As per the above, Vali regeneration was essentially nerfed to x1 values. Resource drain changed from 4 in x1 to 8 (this is the rate at which x2 drained).
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -37,14 +37,6 @@
 	var/obj/item/healthanalyzer/integrated/analyzer
 	///Determines whether the suit is on
 	var/boost_on = FALSE
-	///Stores the current effect strength
-	var/boost_amount = 0
-	///Normal boost
-	var/boost_tier1 = 1
-	///Overcharged boost
-	var/boost_tier2 = 2
-	///Boost icon, image cycles between 2 states
-	var/boost_icon = "cboost_t1"
 	///Item connected to the system
 	var/obj/item/connected_weapon
 	///When was the effect activated. Used to activate negative effects after a certain amount of use
@@ -90,7 +82,6 @@
 /datum/component/chem_booster/Initialize()
 	if(!isitem(parent))
 		return COMPONENT_INCOMPATIBLE
-	update_boost(boost_tier1)
 	analyzer = new
 	meds_beaker = new
 	setup_reagent_info()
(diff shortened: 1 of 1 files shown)


Xander3359:
TLDR Heavily nerfs Vali healing.

ivanmixo:
Seems a bit much? What the hell is the point of vali without good healing?

lKiseki:
> Seems a bit much? What the hell is the point of vali without good healing?

I wouldn't say so, even at x1 the healing is still very decent. I also believe Vali has set a bad precedent when it comes to what a **free** armor module should be capable of doing or allowing.

sa1tman:
removal pr next

Victor239:
Vali healing is absurd for roundstart equipment, not even a Gorger can frontline like this:


https://github.com/tgstation/TerraGov-Marine-Corps/assets/12621257/3c23c6fd-6c79-4688-92fe-60da1082ae44

Yttriums:
Lexia does pop 4 rr pills and deal about 0 damage for the duration of that clip, I don't think it's a great example of Vali being OP in a common use case and doesn't really justify the nerfs in the PR.

Removing 2x is enough of a nerf, greenblood usage on 1x should stay the same imo.

Breadeagle:
This doesn't only half the rate at which vali can heal, it also guts the healing efficiency, being 2x worse. This makes a spear hit reward you with like 15 health. I don't know how this is "nerf this package slightly by reducing the regeneration it offers" since you're both halving its throughput and still halving its blood to healing efficiency. I have a hard time seeing why someone would bother now.

lKiseki:
After thinking it better, I'll close this PR and commit to other options. Vali by itself isn't really _that_ bad, I suppose. I'll certainly tackle other things like Russian Red in #14061, and probably do something about held defensive shields.

PR #14061, opened 2023-10-01
Increases Russian Red's clone loss

lKiseki:
## About The Pull Request
Per title. Increased from 1 cloneloss per 1u to 2.5 cloneloss per 1u

## Why It's Good For The Game
Russian Red is one of the major offenders that allows marines to tank obscene amounts of damage for a drawback you can easily overlook. A pill will restore at least 100 damage for just 10 cloneloss, which seems pretty busted to me. Now, it's 100 damage in exchange for 25 cloneloss.

## Changelog
:cl: Lewdcifer
balance: Russian Red cloneloss damage increased from 1 per 1u to 2.5 per 1u.
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -37,14 +37,6 @@
 	var/obj/item/healthanalyzer/integrated/analyzer
 	///Determines whether the suit is on
 	var/boost_on = FALSE
-	///Stores the current effect strength
-	var/boost_amount = 0
-	///Normal boost
-	var/boost_tier1 = 1
-	///Overcharged boost
-	var/boost_tier2 = 2
-	///Boost icon, image cycles between 2 states
-	var/boost_icon = "cboost_t1"
 	///Item connected to the system
 	var/obj/item/connected_weapon
 	///When was the effect activated. Used to activate negative effects after a certain amount of use
@@ -90,7 +82,6 @@
 /datum/component/chem_booster/Initialize()
 	if(!isitem(parent))
 		return COMPONENT_INCOMPATIBLE
-	update_boost(boost_tier1)
 	analyzer = new
 	meds_beaker = new
 	setup_reagent_info()
@@ -145,7 +136,7 @@
 ///Adds additional text for the component when examining the item it is attached to
 /datum/component/chem_booster/proc/examine(datum/source, mob/user, list/examine_text)
 	SIGNAL_HANDLER
-	examine_text += span_notice("The chemical system currently holds [resource_storage_current]u of green blood. Its' enhancement level is set to [boost_amount].")
+	examine_text += span_notice("The chemical system currently holds [resource_storage_current]u of green blood.")
 	examine_text += get_meds_beaker_contents()
 
 ///Disables active functions and cleans up actions when the suit is unequipped
@@ -188,8 +179,8 @@
 		return
 	update_resource(-resource_drain_amount)
 
-	wearer.adjustToxLoss(-tox_heal*boost_amount)
-	wearer.heal_overall_damage(6*boost_amount*brute_heal_amp, 6*boost_amount*burn_heal_amp)
+	wearer.adjustToxLoss(-tox_heal)
+	wearer.heal_overall_damage(6*brute_heal_amp, 6*burn_heal_amp)
 	vali_necro_timer = world.time - processing_start
 	if(vali_necro_timer > 20 SECONDS)
 		return
@@ -211,7 +202,6 @@
 ///Shows the radial menu with suit options. It is separate from configure() due to linters
 /datum/component/chem_booster/proc/show_radial()
 	var/list/radial_options = list(
-		BOOST_CONFIG = image(icon = 'icons/mob/radial.dmi', icon_state = "[boost_icon]"),
 		EXTRACT = image(icon = 'icons/mob/radial.dmi', icon_state = "cboost_extract"),
 		LOAD = image(icon = 'icons/mob/radial.dmi', icon_state = "cboost_load"),
 		VALI_INFO = image(icon = 'icons/mob/radial.dmi', icon_state = "cboost_info"),
@@ -219,14 +209,6 @@
 
 	var/choice = show_radial_menu(wearer, wearer, radial_options, null, 48, null, TRUE, TRUE)
 	switch(choice)
-		if(BOOST_CONFIG)
-			if(boost_amount == boost_tier2)
-				update_boost(boost_tier1)
-				boost_icon = "cboost_t1"
-				return
-			update_boost(boost_tier2)
-			boost_icon = "cboost_t2"
-
 		if(EXTRACT)
 			extract(10)
 
@@ -279,12 +261,6 @@
 		to_chat(wearer, get_meds_beaker_contents())
 		meds_beaker.reagents.trans_to(wearer, 30)
 	setup_bonus_effects()
-
-///Updates the boost amount of the suit and effect_str of reagents if component is on. "amount" is the final level you want to set the boost to.
-/datum/component/chem_booster/proc/update_boost(amount)
-	boost_amount = amount
-	wearer?.balloon_alert(wearer, "Enhancement level set to [boost_amount]")
-	resource_drain_amount = boost_amount*(3 + boost_amount)
 
 ///Handles Vali stat boosts and any other potential buffs on activation/deactivation
 /datum/component/chem_booster/proc/setup_bonus_effects()
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -37,6 +37,14 @@
 	var/obj/item/healthanalyzer/integrated/analyzer
 	///Determines whether the suit is on
 	var/boost_on = FALSE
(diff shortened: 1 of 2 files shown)


DeltaFire15:
This is more CL than you'd get from dying and getting drained instead....

You say the drawback from RR is barely noticable, however suffering a chunk of health that you are now permanently missing is not minor at all, as you are much easier to be brought to slowdown thresholds, and spiral into situations needing you to RR that you wouldn't have had to before.
If anything, remove the CL heal from sleeping so you actually have to go shipside (or to ground med if fixed) and can't just nap for 30-60 seconds (which is what makes "minor" CL such as that caused by RR counterable).

lKiseki:
> You say the drawback from RR is barely noticable, however suffering a chunk of health that you are now permanently missing is not minor at all, as you are much easier to be brought to slowdown thresholds, and spiral into situations needing you to RR that you wouldn't have had to before.

10 cloneloss is very minor in my honest opinion. Not like you can't just go shipside real quick, spend a few minutes on the cryo pod, then drop pod again like nothing happened. I'm willing to compromise to 20 cloneloss is maintainers deem it appropriate, but I generally believe an increased drawback is what's needed to make Russian Red less egregious.

> If anything, remove the CL heal from sleeping so you actually have to go shipside (or to ground med if fixed) and can't just nap for 30-60 seconds (which is what makes "minor" CL such as that caused by RR counterable).

I'm not sure this is a good idea. I do agree that sleeping healing damage is... hilarious, to say the least, but it's hardly the preferred choice to clear damage. As I said above, just a few minutes in the cryo pod and you're good. Much less time wasted compared to sleeping away 10 to 20 cloneloss, takes much longer by comparison.

Xander3359:
> Not like you can't just go shipside real quick,
> then drop pod again like nothing happened,

This is an awful line of thinking, because we no longer play the same game we did back in 2021. We actively reach pop levels of 80+, and drop pods run out insanely quickly. We've come to the point that medics no longer medevac people because it depletes the ungaball incredibly fast without it being replenished in any meaningful way.

ivanmixo (changes requested):
This makes RR completely pointless. Why would I take it when I would get punished the exact same (arguably even WORSE) than if I just died? 

And no, even 20 cloneloss makes it completely pointless. I might as well just die and avoid potential future fracs.

bluezorua:
if your going to do this, make it impossible to inject someone else with a RR autoinjector because with this change using RR on someone else would become soft-grief

Victor239:
Isn't the whole point of RR to be an emergency big heal, so there needs to be a sizeable drawback?

verywhiterabit:
I don't have an issue with it being raised, 10cl for a 10u injection really isn't much and I pretty much ignore it at that point. 25u for a dose makes it pretty much worthless though as just dying would be less of a nuisance. 

15u would be fair enough as 2 doses adding up to 30cl is a much more noticeable debuff than the current 20.

coldironwarrior:
this in concert with the vali PR really just seem to be a matter of being beaten by certain players in certain scenarios. RR is fine as is.

Pariah919:
This makes Russian Red incredibly punishing unless you hyper microdose it, 25% of your hp is hueg.

lKiseki:
> This makes RR completely pointless. Why would I take it when I would get punished the exact same (arguably even WORSE) than if I just died?
> 
> And no, even 20 cloneloss makes it completely pointless. I might as well just die and avoid potential fracs and organ damage.

Answering to @ivanmixo: That's an overreaction. 10 cloneloss isn't punishing enough for a chemical that heals 10 overall damage per tick, with normal pill dosages healing up to 100.
Referencing @00ze-cyclone, since I see ivanmixo upvoted it, I'd say 20 is a much better compromise than 15 to begin with, and doesn't make it "completely pointless" as you so strongly worded; simply makes it less of a no-brainer option.

I'm open to feedback and suggestions, which I would appreciate _especially_ from a maintainer, instead of just complaints.

verywhiterabit:
Something that was mentioned in the discord was potentially giving it a stun at the end or something so it acts more as an emergency heal to get you back to safety rather than extending your engagement time to a point that it's obnoxious for xenos. Might consider that as a trade-off to more CL.

lKiseki:
> Something that was mentioned in the discord was potentially giving it a stun at the end or something so it acts more as an emergency heal to get you back to safety rather than extending your engagement time to a point that it's obnoxious for xenos. Might consider that as a trade-off to more CL.

Not a bad idea! See, there's other approaches I could probably take like this one, or maybe having it drain stamina (only saying the latter as an example), or what have you. I just need to get Ivanmixo's thoughts and suggestions so I can see how best to proceed.

Sirsnek15:
> Answering to ivanmixo: That's an overreaction. 10 cloneloss isn't punishing enough for a chemical that heals 10 overall damage per tick, with normal pill dosages healing up to 100 in just 10 ticks. Referencing @00ze-cyclone, since I see you upvoted their comment, I'd say 20 is a much better compromise than 15 to begin with, and doesn't make it "completely pointless" as you so strongly worded; simply makes it less of a no-brainer option, and something you probably shouldn't spam to begin with.
> I'm open to feedback and suggestions, which I would appreciate _especially_ from a maintainer, instead of just complaints. Perhaps there's other ways to nerf this besides simply increasing the cloneloss, after all.

That is absolutely not an over reaction. 10 cloneloss *per injection* is admittedly not very punishing. Its true that the first time I use it I'm left off the hook, and in some fringe cases perhaps with a fracture if I used it in melee and healed my burns. Second time I use it WILL be felt. As others have said, 20 cloneloss per injection would literally be psysucking myself. That is not an "over reaction", that is what I would be doing, TWICE. PER. PEN. I would rather my character die than waste the tiny item space. Actually, I would rather the entirety of marines save a tiny item space rather than have RR as well in that circumstance. 20 cloneloss, is completely pointless, as I too have so "strongly worded". Its not strong wording, its simply what it does.

When I was arguing in the discord against this PR, I honestly thought the PR was to double the cloneloss, not make it TWENTY. FIVE. PER, INJECTION. Again, not strongly worded, just entirely literal. The problems with this is so obvious that its kind of hard to say anything more to it, but in conclusion: "25 cloneloss" no.  "I'd say 20 is a much better-" no. "15 cloneloss" NOW we are talking.

### **TLDR**
15 cloneloss is a perfectly sensible amount, and not because its just a nice number. 2 dosages amounts to 30 cloneloss, which I should mention is more than the current 20 in case you are still viewing this as trying to defend 40 or even 50, and would be more punishing without making it useless. 30 cloneloss is the tipping point for getting shipside treatment or staying in the fight a little longer In normal gameplay, if I had 30 cloneloss (either from 3 injections or 1 injection and a psysuck), going to 40 cloneloss with another injection would be the point where I know I have to go up shipside no matter what. 45 is slightly more scary and that would be only from 3 injections. Taking a few acid spits at that amount would instantly put me under. And 15 is not the start point of negotiating against the nerf, its the middle point. 16-17 cloneloss would be tolerable but still cruel, and 11-12 cloneloss would be just barely too lenient if you want to make RR 'interesting'. If you want to make a reasonable approachable RR-cloneloss nerf discussion, the range is set. 12-18 cloneloss is argueable and anything else is, frankly, a waste of time and **a hell of a headache trying to get people to understand when they'd rather ignore anything that goes into detail.**

> Something that was mentioned in the discord was potentially giving it a stun at the end or something so it acts more as an emergency heal to get you back to safety rather than extending your engagement time to a point that it's obnoxious for xenos. Might consider that as a trade-off to more CL.

P.S. Stop being compelled with the power of suggestion by an agressive PR into thinking any creatively evil nerf to RR is a good idea. This is a really bad habit for TGMC nerfs/buffs and I see it Every. Time. when one of these comes up. Just because they're making an aggressive demand doesn't mean we have to all the sudden think "oh maybe we can do this instead thats secretly just as bad" or all of the sudden be forced to compromise, when the opposition was set up with bad faith or a lack of understanding.

SzkieletorNutric:
Watch this get merged

Lumipharon:
Making RR freely available in pill form was a mistake. Agree that this is too severe of a nerf though.

Joe13413:
feel like RR is only aggegious when the people who abuse it, use it. Only really been an issue once the robustos started spamming it, i think it'd be better if you thought of a better fix. Maybe make RR rarer?

efzapa:
tbh I'm a bit against this

if you really want to merge this AND make a compromise, just make it like QK+ where if you take an extra dose within 60 secs or so it does extra CL

otherwise base CL is fine

Lumipharon:
Having a timer like inaprov would be good actually.

zzurge:
This aint it chief

iforgotmeword:
Doesn’t this also make RR an effective 7.5 heal per tick? Down from 9 heal per tick?

efzapa:
> Doesn’t this also make RR an effective 7.5 heal per tick? Down from 9 heal per tick?

pretty sure that makes RR better than a BKTT/MD mix by like 1-1.5 heal per tick lmao
considering BKTT is free and MD is (basically) free, this might as well kill RR if your math adds up

Lumipharon:
Killing this. If you want to nerf RR in someway, this isn't it.

PR #14080, opened 2023-10-04
Mild adjustment to fire stack logic (aka fire buff)

DeltaFire15:
## About The Pull Request
As some of you may remember, a long long while ago I investigated fire armor not mattering at all and rewrote some fire code related to it.
During that time, I made the (for myself) questionable decision to have fire armor effectively doubledip in one particular scenario.

Usually, for fire damage like walking onto fire / getting shot with fire damage type, it's simple: Take damage, apply armor, apply modified damage.

However, this is not the case for "fire stacks", the DoT component of fire.
Here, I (back then) made the decision to have armor apply to the *stacks* being applied (aka how long / how intense the fire will be), but *also* applied it to the actual DoT from life when the fire stacks are being resolved.
It made sense for me back then, but here I am today revisiting it since I'm feeling it gimped the DoT component from fire a bit too much.

So, backstory aside, what *does* this PR actually do?
Fire armor now **only** changes the fire stacks being *applied*.
The DoT from fire stacks already *on* you now is true damage, not modified by any armor

It also adjusts the damage applied to humans by firestack DoT to be 0.5 * fire stacks (firestacks max at 20), instead of a flat 10, which means you will take less damage on average, but if you actually get flamed for full stacks, it will be more lethal.
It also means your DoT will slowly decrease over time instead of being the same regardless of if you have 1 or 20 stacks.

tl;dr:
Armor only affects fire stacks being applied.
The actual DoT from already gained stacks is now true damage.
Damage to humans now scales with stacks, instead of being constant.
## Why It's Good For The Game
My past decisions on fire stack handling may have gimped the effectiveness of fire damage a bit.
This revisits that and attempts to improve the DoT component of fire, since having armor apply twice made it a bit too weak.
## Changelog
:cl:
balance: Fire armor now doesn't apply to the damage from already present firestacks, which is now true damage, only affecting firestacks being gained instead. (This has no effect on anything except for firestacks, aka "being set on fire")
balance: Firestack DoT now scales for humans (similar to xenos), instead of always being 10 damage / life tick.
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -188,7 +188,7 @@
 #define TRAIT_INTOXICATION_IMMUNE "intoxication_immune" // Immune to the Intoxication debuff.
 #define TRAIT_INTOXICATION_RESISTANT "intoxication_resistant" // Resistant to the Intoxication debuff. Maximum amount of stacks limited.
 #define TRAIT_PAIN_IMMUNE "pain_immune" // We don't feel pain.
-///Prevent mob from being ignited due to IgniteMob()
+///Prevent mob from being ignited due to IgniteMob(), aswell as the DoT from already being on fire if the case.
 #define TRAIT_NON_FLAMMABLE "non-flammable"
 /// Prevents mob from riding mobs when buckled onto something
 #define TRAIT_CANT_RIDE "cant_ride"
--- a/code/modules/mob/living/carbon/human/life/handle_fire.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_fire.dm
@@ -5,7 +5,7 @@
 	if(.)
 		return
 	var/thermal_protection = get_flags_heat_protection(30000) //If you don't have fire suit level protection, you get a temperature increase and burns
-	if((1 - thermal_protection) > 0.0001)
+	if((1 - thermal_protection) > 0.0001 && !HAS_TRAIT(src, TRAIT_NON_FLAMMABLE))
 		adjust_bodytemperature(BODYTEMP_HEATING_MAX)
 		adjustFireLoss(0.5 * fire_stacks)
 	species?.handle_fire(src)
--- a/code/modules/mob/living/carbon/human/life/handle_fire.dm
+++ b/code/modules/mob/living/carbon/human/life/handle_fire.dm
@@ -7,5 +7,5 @@
 	var/thermal_protection = get_flags_heat_protection(30000) //If you don't have fire suit level protection, you get a temperature increase and burns
 	if((1 - thermal_protection) > 0.0001)
 		adjust_bodytemperature(BODYTEMP_HEATING_MAX)
-		apply_damage(10, BURN, blocked = FIRE)
+		adjustFireLoss(0.5 * fire_stacks)
 	species?.handle_fire(src)
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -40,7 +40,7 @@
 		if(resting && fire_stacks > 0)
 			adjust_fire_stacks(-1)	//Passively lose firestacks when not on fire while resting and having firestacks built up.
 		return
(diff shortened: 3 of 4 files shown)


QualityVan:
At a glance, doesn't this make jellying a xeno currently on fire do nothing to prevent the damage?

DeltaFire15:
> At a glance, doesn't this make jellying a xeno currently on fire do nothing to prevent the damage?

Guh, you are currently right since resin jello got adjusted at some later date to no longer just short circuit out of the handle fire.
I'll just make TRAIT_NON_FLAMMABLE also prevent the DoT from being on fire.
(Seems a bit weird to not be able to be set on fire with the trait, or gain stacks at all, yet still be able to take damage if you gain it while on fire)

Lumipharon:
I don't really like this. Walking through fire with high fire resist vs being naked will now do the same amount of damage if you're getting extinguished after.

For xenos, they extinguish so fast that reducing stacks is far less significant than the actual damage.

DeltaFire15:
> I don't really like this. Walking through fire with high fire resist vs being naked will now do the same amount of damage if you're getting extinguished after.

 No, it will not, as fire armor applies directly for contact damage, which is what matters the most if you are extinguished right after.
 If you are immediately extinguished, it does in fact behave exactly the same as before, since contact armor is unchanged.
 
 What it does mean is that you partially resisting off firestacks helps against your DoT, aswell as that you need to take care of your fire stacks if you do decide to run through enough fire to cap out of them despite stack reduction.
 
 > For xenos, they extinguish so fast that reducing stacks is far less significant than the actual damage.
 This is false. Adjusting fire stacks instead of fire damage affects burn duration, while allowing for a higher DoT maximum at high stacks than if it were damage itself.
 Extinguishing also is *not* fast unless you use one of the specialized xeno methods, in which case it *would not matter* whether damage or stacks are protected against.
 If speedy methods are not present, it requires resisting, repeatedly, potentially with other xenos, which takes a bunch of time, and you cannot do while retreating since it stuns you.

Lumipharon:
Ok cool, but looking at the pr, fire damage scales with stacks, which means A: fire resist is still providing protection against the DOT, and B: scaling damage is extremely lame since low stacks is just laughably ignorable damage, while high stacks (which were not previously possible, but now very much are due to fire stacking changes) doing comically high damage...
If you think fire doesn't do enough DOT, why don't you just increase fire DOT?

Also xenos resist fire in like 0.2 seconds or something absurdly short, but that's probably its own issue...

DeltaFire15:
> Ok cool, but looking at the pr, fire damage scales with stacks, which means A: fire resist is still providing protection against the DOT

Yes, that is the entire point of this PR.
Armor now protects against the DoT *once*, instead of in two seperate places (stack add *and* the actual damage from the DoT)
>B: scaling damage is extremely lame since low stacks is just laughably ignorable damage, **while high stacks (which were not previously possible, but now very much are due to fire stacking changes)**  doing comically high damage...

Pardon? Please explain the highlighted part of the sentence, since how many stacks you get from x / what stacks you cap out at hasn't been changed at all by this. What has changed is humans scaling damage instead of effectively always assuming highest even if almost extinguished.

>If you think fire doesn't do enough DOT, why don't you just increase fire DOT?

As I explained in the PR, the problem (which I created long ago) is fire armor double dipping on the DoT element, applying to both the DoT stack increases aswell as armor, effectively doubling the what effect armor has.


I may have to touch up on ground-fire fire stacks in general however, since those are currently very inflexible (being linked to both contact damage and the DoT fire stacks.

Lumipharon:
No, literally just increase the damage fire dot does, instead of all these other changes.

Fire has been reworked since you touched it, fire itself was buffed heavily and now stacks, which increases the intensity of fire it applies to mobs.

DeltaFire15:
Disagree;
Anyways, this needs deeper reaching changes since this is but a symptom of how binary fire stacks tend to be for anything but king, aswell as firestacks and contact damage being directly linked instead of variable.
Something for future me to look at.

PR #14082, opened 2023-10-04
Clarifies Smart Pistol Vendor Purchase

Wisemonster:
## About The Pull Request
This PR clarifies that the smart pistol bundle includes smart goggles in the purchase
Change Me - Pillar of Spring 10_4_2023 3_01_00 PM
## Why It's Good For The Game
It is very unclear that the bundle includes the smart goggles as well, so this clarifies that.
## Changelog
:cl:
spellcheck: Changed the smart pistol bundle's name in the smart gunner vendor to clarify that it comes with smart goggles.
/:cl:


Wisemonster:
To clarify, this pr does NOT touch what is included in the bundle, simply renames the bundle to state that it comes with the goggles.

yyzsong:
Why even have two bundles?

FosterGrime-Heymynameslouie:
should probably add the smart pistol itself into the normal bundle instead

but then again I dont think sg gunners would go out of their way to use their smart pistol given how  much more effective your main gun is and if your main sg gun has no ammo your probably fucked either way

so maybe we get to see smart pistol uses more and more with normal marines instead if sg gunners instead just throw their smart pistols out of their prep room?

Wisemonster:
If the smart pistol is considered a smart gun, then you're gonna need smartgun skill in order to use it effectively, meaning marines might not be able to use them well.

PR #14083, opened 2023-10-05
Bandana no longer hides facial hair

Blundir:
## About The Pull Request

Updated some old code, also made it so it longer hides facial hair, it is already rendered over it, there is no point in hiding it, it only causes funny moment when your whiskers vanish because of it.

Also made it so switching between bandana modes actually makes logical sense in terms of stats.

## Why It's Good For The Game

More visual clarity - good.

## Changelog

:cl:
del: bandana no longer hides facial hair
/:cl:

### DIFF
--- a/code/modules/clothing/masks/breath.dm
+++ b/code/modules/clothing/masks/breath.dm
@@ -71,10 +71,9 @@
 	desc = "A colored, resilient, and insulating cloth to cover your face from the elements. This one is Desert Tan"
 	icon_state = "bandanna"
 	item_state = "bandanna"
-	flags_armor_protection = NONE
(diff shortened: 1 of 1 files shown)


Blundir:
> this causes hair to clip out the sides, doesn't it?

It causes more cases of hair vanishing than hair clipping, tested it on like 15 facial hair styles, they are sprited in a way, that clipping is unlikely in general

ivanmixo (approved):
So it used to disable the gas block effect flag forever if you deactivated it eh? Funny.

ivanmixo:
The review is good, you should listen

PR #14108, opened 2023-10-08
TTS update

vvvv-vvvv:
## About The Pull Request

Updated the TTS stuff, mostly code improvements but also there is a lowpass filter on gas masks now

## Changelog

:cl:
qol: Gas masks now muffle your voice with TTS.
/:cl:


code/modules/clothing/masks/gasmask.dm
```diff
@@ -19,6 +19,7 @@
 	var/breathy = TRUE
 	///This covers most of the screen
 	var/hearing_range = 5
+	voice_filter = "lowpass=f=750,volume=2"
```
TiviPlus, line 22: new vars after old ones

TiviPlus (commented):
rest ok i guess

TiviPlus (approved):

Lumipharon:
Do we actually want gas mask effects? Because that'll be applied to everyone, basically.

FosterGrime-Heymynameslouie:
I do think we want gasmasks muffling your voice it sounds cool

carlarctg:
> Do we actually want gas mask effects? Because that'll be applied to everyone, basically.

Let people disable it if they want?

Sirsnek15:
can they be disabled by default they're like super buzzy in the example provided

Sirsnek15:
Also, what qualifies as gasmask? Ski mask has all the benefits but should not have the sound change

PR #14127, opened 2023-10-10
Maptable fixes

Lumipharon:
## About The Pull Request
A (probably) proper fix for maptables when their target z-level changes.

I have no idea why we have two version of the map table that are almost identical, but resolving that can wait for another pr.
## Why It's Good For The Game
Having a map stuck on your screen for the entire game is not ideal.
## Changelog
:cl:
fix: fixed an issue with map tables when their target z-level changes
/:cl:

### DIFF
--- a/code/game/objects/machinery/cic_maptable.dm
+++ b/code/game/objects/machinery/cic_maptable.dm
@@ -13,6 +13,8 @@
 	var/targetted_zlevel = 2
 	///minimap obj ref that we will display to users
 	var/atom/movable/screen/minimap/map
+	///List of currently interacting mobs
+	var/list/mob/interactees = list()
 
 /obj/machinery/cic_maptable/Initialize(mapload)
 	. = ..()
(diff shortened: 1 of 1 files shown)


code/game/objects/machinery/cic_maptable.dm
```diff
@@ -31,6 +33,7 @@
 	if(!map)
 		map = SSminimaps.fetch_minimap_object(targetted_zlevel, allowed_flags)
 	user.client.screen += map
+	interactees += user
```
TiviPlus, line 36: harddel protection

TiviPlus (changes requested):

code/game/objects/machinery/cic_maptable.dm
Lumipharon: deletion triggers on_unset_interaction, so it shouldn't harddel

PR #14135, opened 2023-10-11
giga FC buff

Lumipharon:
## About The Pull Request
Small buff to the FC's sword.
FC sword is literally just a worse machete currently, which is kinda sad.

- Has a strap and bump slash like machete
- Has a small amount of AP
- Has very slightly better attack speed

It's still objectively worse than using a spear, but at least it shouldn't be worse than a common machete anymore.
## Why It's Good For The Game
FC sord be gud.
## Changelog
:cl:
balance: FC's sword has a strap and bumpslash
balance: FC's sword has slightly improved attack speed and AP
/:cl:


ivanmixo (approved):
Yea but one of a kind

PR #14139, opened 2023-10-11
Removes an evil operator & improves some neuro sanity.

DeltaFire15:
## About The Pull Request
Kills an evil operator (:) used in toxin code.
Also improves xeno neurotoxin sanity, as the current logic could cause damage higher than it should do under circumstances where close to the stamina damage cap (or especially if somehow above it)
## Why It's Good For The Game
: is a very evil operator and I will exorcise it wherever possible.
Neurotoxin not being able to be somewhat wacky with its damage under very specific circumstances is neat too.
## Changelog
:cl:
code: Removes an evil operator
fix: Xeno neurotoxin now has one less edge case. 
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -465,12 +465,12 @@
 
 	//Apply stamina damage, then apply any 'excess' stamina damage beyond our maximum as tox and oxy damage
 	var/stamina_loss_limit = L.maxHealth * 2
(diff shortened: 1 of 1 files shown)


Lumipharon:
Neuro does this by design, why are you nerfing it?

DeltaFire15:
> Neuro does this by design, why are you nerfing it?

code comments state neuro is supposed to "If we exceed maxHealth * 2 stamina damage, apply any excess as toxloss and oxyloss"
Note this states *excess*, and is likely the intended behavior.
However, if, say, you have 15 damage from max force neuro, and 14 gets soaked by stam (means 1 overflow), the current code will apply the full 15 as overflow damage despite the correct amount being one.

Lumipharon (approved):
Oh I see, funny.

TiviPlus (approved):

PR #14140, opened 2023-10-11
Fix an exploit allowing you to make C4 instant.

Xander3359:
## Changelog
:cl:
fix: you can no longer make C4 instant
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/plastique.dm
+++ b/code/game/objects/items/explosives/plastique.dm
@@ -32,7 +32,10 @@
 		icon_state = "[icon_state][alarm_sounded ? "_armed" : "_on"]"
 
 /obj/item/explosive/plastique/attack_self(mob/user)
+	. = ..()
 	var/newtime = tgui_input_number(usr, "Please set the timer.", "Timer", 10, 60, 10)
+	if(!newtime)
+		return
 	timer = newtime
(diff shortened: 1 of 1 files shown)


code/game/objects/items/explosives/plastique.dm
```diff
@@ -32,7 +32,10 @@
 		icon_state = "[icon_state][alarm_sounded ? "_armed" : "_on"]"
 
 /obj/item/explosive/plastique/attack_self(mob/user)
+	. = ..()
 	var/newtime = tgui_input_number(usr, "Please set the timer.", "Timer", 10, 60, 10)
+	if(!newtime)
+		return
```
TiviPlus, line 38:
```suggestion
	if(newtime < 1)
		return
```

Xander3359: Should I not make it where it checks the minimum and maximum amount of the timer instead?

Lumipharon: the tgui_input_number clamps it between 10 and 60, so !newtime is fine to me, since this only comes up if nothing is entered at all.

Xander3359: That was my reasoning for it as well

Lumipharon (approved):

PR #14144, opened 2023-10-12
Nerfs Synth CQC skill

Xander3359:
## About The Pull Request
Nerfs Synth CQC to SKILL_CQC_TRAINED (Same as marines)
I would reduce it to lower, but then Synth can no longer easily deweed.
## Why It's Good For The Game
Synth is a non-combat oriented role. People should not be incentivized to pick Synth due to them having an extra 4 melee damage from skills compared to a normal marine.
## Changelog
:cl:
balance: Reduces Synth CQC level.
/:cl:

### DIFF
--- a/code/datums/skills.dm
+++ b/code/datums/skills.dm
@@ -330,7 +330,7 @@ engineer, construction, leadership, medical, surgery, pilot, police, powerloader
 	construction = SKILL_CONSTRUCTION_MASTER
 	firearms = SKILL_FIREARMS_UNTRAINED
 	medical = SKILL_MEDICAL_EXPERT
-	cqc = SKILL_CQC_MASTER
+	cqc = SKILL_CQC_TRAINED
 	surgery = SKILL_SURGERY_EXPERT
 	pilot = SKILL_PILOT_TRAINED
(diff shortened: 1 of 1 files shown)


Alexeivichi:
I approve this nerf. No sarcasm, please nerf more. I am not kidding, do it. No balls.

efzapa (changes requested):
if you are looking to nerf their melee damage, you should be looking at changing SKILL_MELEE_DEFAULT instead to something like weak, or ideally make a new VERYWEAK skill level

CQC affects disarm chances on humans iirc

vvvv-vvvv:
> if you are looking to nerf their melee damage, you should be looking at changing SKILL_MELEE_DEFAULT instead to something like weak, or ideally make a new VERYWEAK skill level
> 
> CQC affects disarm chances on humans iirc

no it adds to melee damage

Lumipharon:
this modifies UNARMED melee damage... I'm pretty sure only against humans as well. What even is this pr?

tanker4390:
Perhaps if you want to preserve maze-clearing ability, the `SKILL_MELEE_WEAPONS` could be divided into "damage vs. obj" and "damage vs. mob?"

PR #14149, opened 2023-10-12
Xeno fade in on spawn and evolution

Waselon:
## About The Pull Request

Adds fade in type fullscreen overlay on xeno spawn and evolution

## Why It's Good For The Game

Adds immersion

## Changelog
:cl:
add: Added fade in type fullscreen overlay on xeno spawn and evolution
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/evolution.dm
+++ b/code/modules/mob/living/carbon/xenomorph/evolution.dm
@@ -183,6 +183,8 @@
 	qdel(src)
 	INVOKE_ASYNC(new_xeno, TYPE_PROC_REF(/mob/living, do_jitter_animation), 1000)
 
+	overlay_fullscreen_timer(2 SECONDS, 20, "roundstart2", /atom/movable/screen/fullscreen/spawning_in)
+
 ///Check if the xeno is currently able to evolve
 /mob/living/carbon/xenomorph/proc/generic_evolution_checks()
 	if(do_actions)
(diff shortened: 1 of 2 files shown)


dopamiin0:
FIRST HIT LAG FIX REAL!????????

TiviPlus:
> yes this is confirmed that loading this overlay file causes lag

Where citation

yyzsong:
post an animation of it

Waselon:
> > yes this is confirmed that loading this overlay file causes lag
> 
> Where citation

I tried adding an overlay on spawn on RUTGMC and xenos reported less lags. When fix was removed, game started to freeze-on-first-hit once again. Adding and removing the fix showed strong correlation with lags presence.

Waselon:
> post an animation of it

same animation a marine would see when waking up from the cryosleep

TiviPlus:
Because it doesnt bloody work, and people gaslight themselves that it does thats why

If this was a working fix then marines would not have hit lag (they already have this same visual as even indicated by the author), I asked multiple people who serially complain about it and they said it happens on marine. The PRs "fix" is just confirmation bias which is why I do NOT TM any of these "fixes" unless they have a technically sound background.

Im removing all references to fixing anything but the actual effect is fine

TiviPlus (approved):

breadhunt:
> If this was a working fix then marines would not have hit lag (they already have this same visual as even indicated by the author)
Not necessarily. The xeno hit lag (although infamous foe first hit deaths) does return after a while of not being attacked, possibly BYOND unloads it? If this fixes it as a byproduct then awesome though.
..come to think of it, I don't think I've ever had hit lag pre drop.

> The PRs "fix" is just confirmation bias

A 20 second freeze --> instant death is pretty noticeable but I do get your point. I could provide recordings if you want to see the difference

Lumipharon:
fade in when you evo is kinda weird but eh.

PR #14171, opened 2023-10-14
Removes marine jetpacks from standard vendors

lKiseki:
## About The Pull Request
Per title. This of course excludes Valhalla vendors and HvH loadouts.

## Why It's Good For The Game
These are free get out of jail cards, that allow you to go very deep and get out of situations you really shouldn't be able to get out of. Rather than nerfing it to oblivion, I believe it best to keep this locked behind reqs; it is worth 120 points there.

## Changelog
:cl: Lewdcifer
balance: Removes marine jetpacks from standard vendors.
/:cl:


GrayRachnid:
Out of all the nerf PRs post maginot line, this is the only one I disagree with. Let marines keep their niche quirky jetpack that comes in low quantity and is only really good at one thing.

Marines already have so much crap behind req that I feel like they never know what they want. This might not even ever get picked up unless req is rich just because it's only really good at one thing.

Derrick2Electricboogaloo:
The amount of crying about get of jail free cards for marines while ignoring the plathora of get out of jail free shit beans have.

imsxz:
As a jetpack abuser I have no clue why they get any for free, idk why it’s even as powerful as it is. IMO req locking isn’t the solution, the type of people to abuse jetpack to its full potential will order it roundstart anyway(myself included)

Needing a back slot doesn’t matter I’ll if you’re not playing a support or double gun build, carrying stuff around on tad is good enough most of the time

IMO jetpack is only extremely obnoxious with vali because it buys 2-3 vali healing ticks. Jetpack is annoying but still very manageable vs non vali. Honestly makes me think that the best change would be making jetpack into a suit module so it can’t be used with vali, b18, tyr 2, etc.

imsxz:
P.s. I don’t sacrifice an inventory slot to refuel jetpack I just put mini incend tanks on tad OR refuel shipside when I go up for fracs/CL or find groundside fuel tank

lKiseki:
> IMO jetpack is only extremely obnoxious with vali because it buys 2-3 vali healing ticks. Jetpack is annoying but still very manageable vs non vali. Honestly makes me think that the best change would be making jetpack into a suit module so it can’t be used with vali, b18, tyr 2, etc.

I talked with ivanmixo about this in Discord, I'll make it a module that reduces armor. Very gud idea, thanks.

efzapa:
improve don't remove my man

TiviPlus:
I dont like this, it seems fine to me and feel the getoutofjail aspects are pretty exaggerated

TiviPlus:
If you're making a module should probably be a second pr instead

PR #14177, opened 2023-10-14
Ravager endure vali slowdown protection

Waselon:
## About The Pull Request

Protects Ravagers in Endure from getting slowed down by vali tramadol attack

## Why It's Good For The Game

Fixing an oversight/inconsistency between vali and bullets slowdown for ravager

## Changelog
:cl:
balance: Ravagers in Endure are no longer slowed down by vali tramadol attack
/:cl:

### DIFF
--- a/code/datums/components/harvester.dm
+++ b/code/datums/components/harvester.dm
@@ -180,6 +180,10 @@
 	switch(loaded_reagent)
 		if(/datum/reagent/medicine/tramadol)
 			target.apply_damage(weapon.force*0.6, BRUTE, user.zone_selected)
+			if(isxeno(target))
+				var/mob/living/carbon/xenomorph/xeno_target = target
+				if(xeno_target.endure)
+					return
 			target.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 1 SECONDS)
 
 		if(/datum/reagent/medicine/kelotane)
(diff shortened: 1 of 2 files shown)


yyzsong:
Should be a feature of rage instead, keeps you from being staggered and boosts your movespeed.

code/datums/components/harvester.dm
```diff
@@ -180,6 +180,10 @@
 	switch(loaded_reagent)
 		if(/datum/reagent/medicine/tramadol)
 			target.apply_damage(weapon.force*0.6, BRUTE, user.zone_selected)
+			if(isxeno(target))
+				var/mob/living/carbon/xenomorph/xeno_target = target
+				if(xeno_target.endure)
+					return
```
DeltaFire15, line 186: Using return here will prevent this proc from reaching certain sections after the switch case, such as informing the user if the reagent storage is empty. I recommend just throwing the status application into the inverted if conditions instead.

DeltaFire15 (changes requested):

code/datums/components/harvester.dm
Waselon: return is used later in the code in bicardine section, does it cause issues?

DeltaFire15: For bicardine it is correct since you do not want the reagent effect to fire at all when attacking xenos since it heals.
Here however, you still apply the bonus AP damage, which basically means on enduring Ravagers you'd get free bonus damage on every stab (since it'd not consume the charge)

QualityVan (changes requested):
Rather than coding the specific ability into harvester code, it'd be better to make the status effect properly respect TRAIT_SLOWDOWNIMMUNE

QualityVan (approved):

PR #14191, opened 2023-10-15
Nerfs the scout rifle's sundering stats

lKiseki:
## About The Pull Request
Normal magazine sundering reduced from 10 > 5.
Impact magazine sundering reduced from 12.5 > 6.5

## Why It's Good For The Game
The scout rifle deletes armor really, _really_ quickly. This should be a tad bit more fair.

## Changelog
:cl: Lewdcifer
balance: Scout rifle: normal mag sunder reduced from 10 > 5, impact mag sunder reduced from 12.5 > 6.5.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -753,7 +753,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	hud_state = "hivelo_impact"
 	damage = 30
 	penetration = 10
-	sundering = 6.5
+	sundering = 7
 
 /datum/ammo/bullet/rifle/tx8/impact/on_hit_mob(mob/M, obj/projectile/P)
 	staggerstun(M, P, max_range = 14, slowdown = 1, knockback = 1)
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -736,7 +736,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	accurate_range = 15
 	damage = 40
 	penetration = 20
-	sundering = 10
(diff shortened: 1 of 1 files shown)


x64SLUMLORD:
As an avid enjoyer of the Scoutie and long time (ab)user, I heavily disagree with the nerf.

Sunder is and has always been **the** purpose of the gun. It's a slow firing precision rifle specifically made for supporting your teammates from behind to make them do more damage. By itself, the Scoutie is good, but it shines best when you're inside the ungaball helping suppress. Without the sunder, it's just an unimpressive gun at best.

The issue with the Scout is that it's a criminally good gun that is cheap. It costs 520 points if I'm not mistaken, which is as easy to get as fultoning a singular Xeno or getting a miner running for a few minutes. The base price should definitely be increased substantially so it can't be spammed and it weighs on Req to get one like a B18 or minigun does. It's okay for Req guns to be good, that's the point of them being Req-locked in the first place.

Lumipharon:
For reference, the scout rifle has lower dps than a t12 except at long range or very high armour castes like crusher.
Hell, aimmode t12 is barely lower than a scout rifle, even without attachments or aim order.

Sunder is the point of the scout, so this is a pretty extreme nerf.

Unlike a ZX or TEX, there is an ammo cost for the scout, I seriously don't think this makes sense.

Joe13413:
ehhh im gonna be real, the scout rifle will still have a nice sunder/sec of 12.5
I think 25sunder/sec is way too much, but 12.5 puts it just above SGs (which isnt really worth it IMO)
maybe something in between the two would keep the scout rifle a competitive choice, but not have extremely high sunder rates?

FosterGrime-Heymynameslouie:
it feels like this is a nerf pr for the sake of a nerf pr 

same goes with all the other prs that was set up in conjunction with eachother

BoltedAirlock:
"should be a tad bit more fair" you are halving the sunder of a weapon that exists TO sunder without a single damage top-up to compensate for it performing around the same level as a non-req gun.

PR #14194, opened 2023-10-16
Instant asset fix

Lumipharon:
## About The Pull Request
Instant effect assets will correctly reproc when obtained.
## Why It's Good For The Game
Asset work proper is gud.
## Changelog
:cl:
fix: Instant effect assets correctly reapply if obtained multiple times
/:cl:

### DIFF
--- a/code/datums/gamemodes/campaign/campaign_rewards.dm
+++ b/code/datums/gamemodes/campaign/campaign_rewards.dm
@@ -62,6 +62,13 @@
 	faction = null
 	return ..()
 
+///Reapplies the asset, refreshing it or restoring uses
+/datum/campaign_reward/proc/reapply()
+	existing_reward.uses += initial(existing_reward.uses)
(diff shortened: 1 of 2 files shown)


code/datums/gamemodes/campaign/campaign_rewards.dm
```diff
@@ -62,6 +62,13 @@
 	faction = null
 	return ..()
 
+///Reapplies the asset, refreshing it or restoring uses
+/datum/campaign_reward/proc/reapply()
+	uses += initial(uses)
```
QualityVan, line 67: Just to be clear, you want behavior here to be carrying over unspent uses rather than strictly refreshing? I'm not particularly familiar with campaign flow.

Lumipharon: yep. You can stack CAS uses in a single mission, for example.

QualityVan (approved):

PR #14197, opened 2023-10-16
Req ARSR and drop console aren't deconstructable anymore

ivanmixo:
## About The Pull Request

Don't move these guh!
## Why It's Good For The Game

These really shouldn't be moved since it screws over req.
## Changelog
:cl:
del: Req ARSR and drop console aren't deconstructable anymore
/:cl:

### DIFF
--- a/code/game/objects/machinery/squad_supply/supply_console.dm
+++ b/code/game/objects/machinery/squad_supply/supply_console.dm
@@ -38,6 +38,9 @@
 	supply_pad = null
 	return ..()
 
+/obj/machinery/computer/supplydrop_console/screwdriver_act(mob/living/user, obj/item/I)
+	return TRUE //no deconstructing this please
+
 /obj/machinery/computer/supplydrop_console/ui_interact(mob/user, datum/tgui/ui)
 	. = ..()
 	ui = SStgui.try_update_ui(user, src, ui)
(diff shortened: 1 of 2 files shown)


lKiseki (changes requested):
Computers use an `attackby` proc and don't call `screwdriver_act` anywhere. You'd have to change that code a little, but the easiest solution seems to be removing the `circuit` var for these two consoles here. It's only needed for deconstruction from what I can see, and its absence prevents it too.

lKiseki:
> Computers use an `attackby` proc and don't call `screwdriver_act` anywhere, so this does nothing. You'd have to change that code a little, but the easiest solution seems to be removing the `circuit` var for these two consoles here. It's only needed for deconstruction from what I can see, and its absence prevents it too.

DeltaFire15:
> but the easiest solution seems to be removing the `circuit` var for these two consoles here
Both the current and suggested approach would make you unable to repair either console if damaged or destroyed. If you are doing this, the console must be indestructible aswell.

Wirpy:
The req console can be killed by like 3 paint ball shots btw.

Lumipharon (changes requested):
Did... you test this?
Also does this stop you from repairing it if it gets damaged?

FosterGrime-Heymynameslouie:
I kinda like moving them to places more convinient within req to make req run more efficiently, specially the drop console one

PR #14205, opened 2023-10-17
Autobalance Fix

Barnet2:
## About The Pull Request
So for some reason Bravemole used a length on a number making it always return 0, meaning the calculation for how much of a stat buff xenos would get would always be 150%.
In normal people terms this means that the autobalance stat buff will care about how many burrowed there actually are
## Why It's Good For The Game
Bug left unfixed for 2 years bad
## Changelog
:cl:
fix: Autobalance will now care about the ratio of alive xenos to burrowed xenos when giving out stat buffs
/:cl:


lbnesquik:
It is a good idea for a fix, but at the same time, doesn't it kinda nerfs things a bit ?

Lumipharon (approved):

PR #14216, opened 2023-10-19
Makes Tramadol in a harvester no longer have extra damage.

Xander3359:
## About The Pull Request
Removes the damage bonus of Tramadol.
## Why It's Good For The Game
image

## Changelog
:cl:
balance: Tramadol in a harvester weapon no longer has extra damage.
/:cl:

### DIFF
--- a/code/datums/components/harvester.dm
+++ b/code/datums/components/harvester.dm
@@ -179,7 +179,7 @@
 
 	switch(loaded_reagent)
 		if(/datum/reagent/medicine/tramadol)
-			target.apply_damage(weapon.force*0.6, BRUTE, user.zone_selected)
+			target.apply_damage(weapon.force*0.6, BRUTE, user.zone_selected, MELEE)
 			target.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 1 SECONDS)
 
 		if(/datum/reagent/medicine/kelotane)
--- a/code/datums/components/harvester.dm
+++ b/code/datums/components/harvester.dm
@@ -179,7 +179,6 @@
 
 	switch(loaded_reagent)
 		if(/datum/reagent/medicine/tramadol)
-			target.apply_damage(weapon.force*0.6, BRUTE, user.zone_selected, MELEE)
 			target.apply_status_effect(/datum/status_effect/incapacitating/harvester_slowdown, 1 SECONDS)
 
 		if(/datum/reagent/medicine/kelotane)


GravitronBlue:
A nice balance PR and good start on balancing vali, but why does tramodol specifically have the extra 60% damage on top of slowdown to begin with?

Ultimately rather see the extra damage removed/moved elsewhere (perhaps kelo) and instead just give spears in general a flat 10-20% AP modifier.

Lumipharon:
I'd rather just nerf tram so its not just the absolute best option due to doing significantly more damage, and slowdown. Just one or the other. Then other chems can actually be balanced against that more reasonable standard/have options expanded, and general balance vali from that point.

ivanmixo:
This makes vali absolutely worthless.

UncleBourbon:
why dont you nerf the dumb stuff like ar-11 instead of the fun stuff like vali

Sirsnek15:
claymore loses all purpose without bonus damage

zzurge:
Imagine if you will having marines ditch all their ranged options in favor of only being able to engage you at your most effective combat range only to then fail at killing them in that range,Your defilers have failed you if tramspears are getting you guys this pressed

Joe13413:
good and bad pr at the same time
i feel theres a large sentiment of "vali is only good with the extra damage" but the extra damage + slowdown is too much i feel
I can almost get behind this PR, maybe buff vali in addition to this?

Xander3359:
> I can almost get behind this PR, maybe buff vali in addition to this?
I would like to buff Vali, but this is a necessary first step before anything meaningful can happen.

ivanmixo:
Make one PR which will do everything you want related to vali instead of doing it incrementally like this. I don't want to risk you losing motivation for future PRs and leaving vali in a gutted state.

Xander3359:
> Make one PR which will do everything you want related to vali instead of doing it incrementally like this. I don't want to risk you losing motivation for future PRs and leaving vali in a gutted state.

Vali needs this sort of nerf no matter how you look at it.
If Vali is crippled because it relies on Tram to exist, then Vali needs to be looked at as a whole.
I'm only 1 player and I can only get so much of an idea balance-wise through my own experience, and I can only guesstimate what it would feel like without the damage bonus.

TiviPlus:
I'd move it at least partially to the base weapon damage honestly if the concern is that it outstrips everything

Xander3359:
> I'd move it at least partially to the base weapon damage honestly if the concern is that it outstrips everything

Make the AP damage innate? Or do you mean just adding a flat damage amount.

TiviPlus:
Whatever the change you removed honestly

Xander3359:
I'm doing more vali stuff

PR #14217, opened 2023-10-19
A harvester weapon with Kelotane applies some sunder.

Xander3359:
## About The Pull Request
A harvester weapon with Kelotane applies some sunder, I plugged in the value of a slug shotgun, but if thats too high/low I can adjust.
## Why It's Good For The Game
Kelotane is currently really bad, and I think this differentiates it from Tramadol.
You would use the Tram for the slowdown, and Kelotane for the armor damage.
I think this would give both chems a fair niche.
## Changelog
:cl:
balance: Harvester weapon with Kelotane causes sundering
/:cl:


Lumipharon:
nerf tram. Balance/add chems based off that. Current tram is aids in that it's in an entirely different class.

TiviPlus (approved):

PR #14228, opened 2023-10-20
Fix statpanel not loading turf icons

### LINKED ISSUE #14209: TG stat panel has horrible alt tile click performance
### Description

Previously, when using alt click, you could QUICKLY and SMOOTHLY see the list of items on a tile, and browse through them. Sure, icons broke sometimes, but it was the one good way for you to find out what a tile held. 

Since https://github.com/tgstation/TerraGov-Marine-Corps/pull/13546 merged however, the performance has been actually awful, and worse than right clicking big stacks! It actually made my game crash multiple times, especially with big crates of stuff. 

### Test Merges

Not relevant

### Reproduction Steps

Go to a big stack of objects
Alt click it
Watch as the game suffocates to death for some reason. 

### Screenshots

_No response_

vvvv-vvvv:
## About The Pull Request

Fixes #14209, which was because it was spamming a bunch of requests to these icons
<details>
<summary>a</summary>
Why is it not this way upstream? I don't know. The reason it wasn't working before is because it tried reloading them but then called draw_listedturf immediately afterwards which just removes the old elements and makes new ones. And yes, the URL will be huge if it fails to load many times but I don't think it's a problem
</details>

## Changelog

:cl:
fix: Fix statpanel not loading turf icons
/:cl:

### DIFF
--- a/html/statbrowser.js
+++ b/html/statbrowser.js
@@ -30,7 +30,6 @@ var permanent_tabs = []; // tabs that won't be cleared by wipes
 var turfcontents = [];
 var turfname = "";
 var imageRetryDelay = 500;
-var imageRetryLimit = 50;
 var menu = document.getElementById('menu');
 var under_menu = document.getElementById('under_menu');
 var statcontentdiv = document.getElementById('statcontent');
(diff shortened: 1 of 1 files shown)


vvvv-vvvv:
I'll refactor this entire thing later when I feel like it, it's really wacky

PR #14233, opened 2023-10-21
Expands on random disks by adding disk sets

DeltaFire15:
## About The Pull Request
This is a purely backend change that will not affect anything until mappers make use of it.

This PR makes makes nuke disk candidates belong to at least one "set" of disks, which are defined in the json of the map.
How this works (For you, mysterious Mapper) is:
You set up the json by having a name-weight list for your disk set's name.
You add the disk candidates to the map, and adjust their "set_associations" list accordingly.
You are done.
If you would like to, you can also add sets to "force_for_sets" on the disk candidate (**aswell as having said sets in the associations of the candidate**), which will make that candidate always appear for the given sets. This is only useful for sets with more than 3 disks (aka "base rng")

Disk sets must be at least 3 candidates. If more than 3, it behaves like the current iteration where it will pick 3 random disk locations from the potential ones in the set.
Disk candidates default to the "basic" set if you do not edit the set_associations var. If your map has a single disk set, just add a info for the basic set into the json (like i have done for all current maps, just copypaste that), and it'll behave.

(Note that json changes require a map change to take effect due to how the mapping subsystem backups the current map's json, so change your map when localtesting json changes)
(Disk candidates can belong to multiple sets, which is why the list of associations is a list and not a var)

When disk generators are next loaded, it will pick one of the disk sets you gave in it the json (going by weighted rng depending on the weight you gave each se),
then take as many candidates as it needs (3) from the set (prioritizing ones forced), and turn them into the actual computers.
If it doesn't manage, it will yell.

This PR adjusts all current jsons of maps that have nuke disks accordingly. If I forgot one (reads: the nuke disks do not load correctly and it starts yelling in runtimes), do tell. Should only happen is cases of badly mapped canter crash sites though. Fixing those is outside the scope of this PR however.

**Scuffed screenies:**
Json:
grafik

Vars to touch in map:
grafik
## Why It's Good For The Game
Adds some more flavor for nuke disk location mapping, allowing mappers to not only clearly define how rng sets can look like (besides color), but also making it possible to force certain computers, and to have some sets be less common than others.
## Changelog
:cl:
code: Nuke disk generator candidates now support sets for mapping, which allows more control over which disk groups appear aswell as how common they are. Look at this PR if you need to know how.
code: Nuke disk generators can also be forced to always appear for specific sets.
/:cl:

### DIFF
--- a/_maps/bigred_v2.json
+++ b/_maps/bigred_v2.json
@@ -2,6 +2,9 @@
     "map_name": "Big Red",
     "map_path": "map_files/BigRed_v2",
     "map_file": "BigRed_v2.dmm",
+	"disk_sets": {
+		"basic": 1
+	},
 	"quickbuilds": 1800,
 	"announce_text": "A second generation colony has had a beacon transmitting the same signal, nonstop. Attempts to hail the colony over comms have proved futile. Because the ship was at a nearby drydock, it has been dispatched to figure out what's wrong. TGMC, prepare to deploy!",
 	"traits":[{
--- a/_maps/deltastation.json
+++ b/_maps/deltastation.json
@@ -2,6 +2,9 @@
     "map_name": "Delta Station",
     "map_path": "map_files/deltastation",
     "map_file": "deltastation.dmm",
+	"disk_sets": {
+		"basic": 1
+	},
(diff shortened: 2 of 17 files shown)


coldironwarrior:
so cool, tysm for this, was thinking about it just today for magmoor and gelida.

imsxz:
🙏I really disliked how much the old disk RNG could heavily favor one team. I look forward to seeing how mappers use this PR

PR #14238, opened 2023-10-21
Tanky xenos get more sundering capacity

Waselon:
## About The Pull Request

Queen 100 > 150 sunder
King 100 > 150 sunder
Crusher 100 > 200 sunder
Defender 100 > 150 sunder

## Why It's Good For The Game

Hopefully this makes slow tanky xenos more comfortable to play since they suffer the most from sundering feature.

## Changelog
:cl:
balance: Queen, King, Crusher and Defender get more sundering capacity
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/castedatum_crusher.dm
@@ -23,6 +23,9 @@
 	// *** Health *** //
 	max_health = 400
 
+	// *** Sunder *** //
+	sunder_max = 200
+
 	// *** Evolution *** //
 	upgrade_threshold = TIER_THREE_THRESHOLD
 
--- a/code/modules/mob/living/carbon/xenomorph/castes/defender/castedatum_defender.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/defender/castedatum_defender.dm
@@ -23,6 +23,9 @@
 	// *** Health *** //
 	max_health = 320
 
+	// *** Sunder *** //
+	sunder_max = 150
+
 	// *** Evolution *** //
 	evolution_threshold = 100
 	upgrade_threshold = TIER_ONE_THRESHOLD
--- a/code/modules/mob/living/carbon/xenomorph/castes/king/castedatum_king.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/king/castedatum_king.dm
@@ -22,6 +22,9 @@
 	// *** Health *** //
 	max_health = 650
 
+	// *** Sunder *** //
+	sunder_max = 150
+
 	// *** Evolution *** //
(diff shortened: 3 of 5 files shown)


Alexeivichi:
Values are too high, will break the balance and cause massive xeno Ws endlessly. Try smaller value increases instead, maybe?

DeltaFire15:
Don't all sunder heals heal flat sunder? 
This just makes beans more hard to crack, while sending them to rest jail for even longer if they get hit by any funny "you are now sundered to hell" weapons. (crusher needs to rest for twice as long to get the same amount of armor% back, for example)

ivanmixo:
These are the castes that are supposed to suffer from sunder though

Pariah919:
I gave crusher lowered sunder in the crusher pr I made ages back because I gave it less HP. All this does is make tanky xenos resistant to armor breaking despite that being supposed to be one of their big downsides, funny powercreep really.

bowlofcereal:
the entire point of sunder is to effect tanky xenos so they actually get punished

Waselon:
I think tanky xenos should recieve updated sunder values since marine sundering capabilities have been inevitably increased over the years.
Values are going to get lowered if a maintainer says so.

tmurphy7:
While I think there should be some sunder tweaks, the values on 95% of xenos is fine as is.  The only caste I think that should have a sunder capacity increase is King because most of their abilities require them to stand still and tank damage for a while, making them more susceptible to sunder.  Even then, King is mostly fine overall, so I'd only argue for 125% on king at most.  The rest of these would just make those xenos unbearably strong.

Pariah919:
> I think tanky xenos should recieve updated sunder values since marine sundering capabilities have been inevitably increased over the years. Values are going to get lowered if a maintainer says so.

The only sunder straight buff have mostly been like... some mounted weapons and GPMG if I remember right? If anything, sunder is LOWER across the board than it was on release, most things have reduced in sunder, and or otherwise stood the same. The biggest changes has been makign railgun magfed and reducing its sunder to 50 and me making disposables more expensive. (which is a net positive for game health imo) 

Either way sundering is basically nearly the same as it was on release, so that's not a good justification.

Waselon:
> > I think tanky xenos should recieve updated sunder values since marine sundering capabilities have been inevitably increased over the years. Values are going to get lowered if a maintainer says so.
> 
> The only sunder straight buff have mostly been like... some mounted weapons and GPMG if I remember right? If anything, sunder is LOWER across the board than it was on release, most things have reduced in sunder, and or otherwise stood the same. The biggest changes has been makign railgun magfed and reducing its sunder to 50 and me making disposables more expensive. (which is a net positive for game health imo)
> 
> Either way sundering is basically nearly the same as it was on release, so that's not a good justification.

_Flechette casually removing half of the xeno armor in just 3 shots_

TiviPlus:
See bowl/ivan/mjp/etc comments

PR #14245, opened 2023-10-22
Hugger psypoints reward

Waselon:
## About The Pull Request

Successfully hugging a marine rewards 25 psypoints.
If lavra is fully grown inside a marine, xenos are rewarded with 75 psypoints.

## Why It's Good For The Game

Makes lavra huggers less useless.

## Changelog
:cl:
balance: Hugger psypoints reward
/:cl:

### DIFF
--- a/code/__DEFINES/mode.dm
+++ b/code/__DEFINES/mode.dm
@@ -130,6 +130,10 @@
 #define COCOON_PSY_POINTS_REWARD_MAX 3
 //How many psy points are gave every 5 second by a cocoon at high pop
 #define COCOON_PSY_POINTS_REWARD_MIN 1
+//How many psy points are given for each hugger impregnation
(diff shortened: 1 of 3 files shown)


AlexCrimson:
Its easy to hug a marine, but impossible to do much else. If they are alone, at most you just get a ministun and send them back to the ship to get surgery. More likely they will have a source of fire to remove the hugger. If they are in a group, which is way more likely, then someone else will throw a flare to remove the hugger.

Fire damage is negligible. They can still fight back.

Larval huggers suck ass and should just be changed to something else at this point. They are so bad that its an active detriment to the caste to use them. 

This PR is definitely not the answer. Giving psy points for a hug is ridiculous. But Larval huggers do need some kind of buff, rework or just outright removal.

coldironwarrior:
thematically not into this

Lumipharon:
Why tho? The benefit is getting a free larva.

PR #14004, opened 2023-09-23
Sow for all

Waselon:
## About The Pull Request

Adds Sow (xeno flowers) ability for every **builder** caste as their **primo** ability.
Hivemind gets it as a base ability since he has no primo rank.

## Why It's Good For The Game

I believe that Sow is very well designed and perfectly balanced ability. Adding it to more castes as a primo ability expands building complexity and adds more gameplay options while not contributing towards powercreep.

## Changelog
:cl:
add: Added Sow as primo for Drone, Queen, Shrike. Added Sow as a base ability for Hivemind.
/:cl:

PR #14247, opened 2023-10-22
New hivelord primo and xeno flowers as primo for all builders

Waselon:
## About The Pull Request

New hivelord primo: Resin Globe.
Sow was added to all builder castes as primo ability. As per maintainer request, Hivelord gets its own primo.
Fixed plasma fruit so it now respects special plasma options (ravager, gorger, puppeteer)

*cooldown was reduced for the showcase

https://github.com/tgstation/TerraGov-Marine-Corps/assets/69199543/bbe1bc6a-258a-499f-a996-69cb80e48878

## Why It's Good For The Game

Explained in https://github.com/tgstation/TerraGov-Marine-Corps/pull/14004

## Changelog
:cl:
add: New hivelord primo: Resin Globe.
balance: Sow was added to all builder castes as primo ability.
fix: Plasma fruit now respects special plasma options (ravager, gorger, puppeteer)
/:cl:


Xander3359:
Praise the almighty sticky hivelord!
Also, sow might need a few nerfs.
It's literally only balanced by the fact that hivelords never get primo, which means it doesn't see use.

Dgeen:
> It's literally only balanced by the fact that hivelords never get primo, which means it doesn't see use.

Initially sow was a drone primo. And drone got it every round (I was that drone, give sow back to the drone).

KrabSpider:
> Power fruits give gorgers 100% blood and I think the same goes for ravs. 

That's a bug...

GrayRachnid:
That is waaay too much sticky for what appears to be negligible plasma cost.

This will halt and fuck over marine pushes more than any quick walling ever can. Even if they're 1 hit to remove, they're so cheap to spam.

coldironwarrior:
> Sow will need some changes if this goes through. Power fruits give gorgers 100% blood and I think the same goes for ravs, and since you can spam them that means they're always topped up. Not a huge issue for ravs but gorgers having 100% blood means they can keep the entire hive overhealed very easily.

i'm not sure if the flag you edited in the code, waselon, is also the one used for puppeteer's biomass amount flags, but ditto to what dawson said, in case it needs to be changed for puppeteer also.

Waselon:
Resin Globe plasma cost has been increased from 50 to 200, cooldown increased from 5 to 8 seconds

code/modules/xenomorph/xeno_structures.dm
```diff
@@ -1534,7 +1534,7 @@ TUNNEL
 		return TRUE
 
 	var/mob/living/carbon/xenomorph/X = user
-	if(isxenoravager(X)) //Ask if this should be made into a trait for xenos with special ressources
+	if(!(X.xeno_caste.can_flags & CASTE_CAN_BE_GIVEN_PLASMA)) //Ask if this should be made into a trait for xenos with special ressources
```
Arctrooper8, line 1537: You're probably good to remove the comment, and you should change the flavor text if you can due to it being more themed around Ravagers and their rage resource.

TiviPlus (approved):

PR #14263, opened 2023-10-24
Icy cave balance changes

vilereaver41:
## About The Pull Request

Attempts to fix LZ2 to make it more appealing to choose for Marines and removes some ice rocks that made cover for xenos to easy. Adds more crash spawns for canterbury and removes the spawns where it could crash right beside the fog which makes life bad for marines, also removes it from crashing right beside disks. Moves the Disc spawn in Monorail area to be in the medbay/morgue area, i believe this to be best for both sides.

extends the fog in north monorail a bit more. also made the nuke hallway of death west spawn for xenos, further away to be more fair for marines.

## Why It's Good For The Game

This should hopefully make crash in Icy caves more bareable, and also make nuclear-war there better with the option of a more open LZ2 

## Changelog

:cl:
balance: Lz2 is more open and less cluttered.
balance: cantabury no longer crashes right besides discs or fog.
balance: north monorail disc will be better to defend and attack.
/:cl:

image
image
image

### DIFF
--- a/_maps/map_files/icy_caves/icy_caves.dmm
+++ b/_maps/map_files/icy_caves/icy_caves.dmm
@@ -501,7 +501,7 @@
 /area/icy_caves/caves/northwestmonorail/medbay)
 "cv" = (
 /obj/effect/landmark/lv624/fog_blocker/xeno_spawn,
-/turf/open/floor/tile/dark,
+/turf/closed/wall/r_wall,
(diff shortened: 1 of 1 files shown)


Luxyboo:
i feel like the new fog expansion in the north west is a little excessive.

vilereaver41:
> i feel like the new fog expansion in the north west is a little excessive.

It isnt. its much much better then the previous fog.

PR #14104, opened 2023-10-07
ghost runechat

Xander3359:
## About The Pull Request
Adds ghost runechat
## Why It's Good For The Game
image
Makes it easier for communication between ghosts without your message getting lost in a sea of dchat and OOC
## Changelog
:cl:
add: ghost runechat
/:cl:

PR #14267, opened 2023-10-24
Revert "ghost runechat (#14104)"

vvvv-vvvv:
It's currently broken (only the speaker can see it) and it's a stupid idea anyway, it takes up a sizable portion of actual gameplay and the only toggle is disabling ghost vision entirely

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -96,7 +96,7 @@
 		displayed_key = null
 
 	deadchat_broadcast(rendered, source, follow_target = src, speaker_key = displayed_key)
-	create_chat_message(src, /datum/language/common, message)
+
 
 /mob/living/proc/get_message_mode(message)
 	var/key = message[1]


Xander3359:
> it takes up a sizable portion of actual gameplay

What. The only size it would take up, would be the size of the text on screen.
I don't know why it doesn't currently work, but it would be cooler if you attempted to fix it rather than just removing it for no good reason.

Luxyboo:
As nice as this is i think a fair bit of people would like to be able to toggle it off.

Xander3359:
> As nice as this is i think a fair bit of people would like to be able to toggle it off.

a toggle can easily be added as well!
Question though, can you disable runechat for living people? Because if you can't, then it wouldn't make sense to be able to disable specifically ghost runechat, when there is already a "toggle ghosts" option.

TiviPlus:
Adding a toggle and making it visible to multiple people isnt that hard

PR #14269, opened 2023-10-24
Crusher toss fix

### LINKED ISSUE #14265: Crusher can double the range of crest toss by using it diagonally
### Description

Crusher crest toss range is mathematically doubled by using it diagonally.
Crest toss has a range of 6? tiles. If you use it diagonally, the marine will travel the full 6 tiles diagonally. 
Both mathematically and displacement-wise, this is double the distance.

### Test Merges

N/A

### Reproduction Steps

1) Be crusher
2) Stand diagonal to a human (North west, north east, south east or south west corner)
3) Crest toss

### Screenshots

https://github.com/tgstation/TerraGov-Marine-Corps/assets/6613514/0309b379-c55b-41b1-a659-1182f3935b31

https://github.com/tgstation/TerraGov-Marine-Corps/assets/6613514/4846bda8-c22f-4a11-9999-02b860093dad

Lumipharon:
## About The Pull Request
Fixes #14265

Crest toss tosses the victim from the CRUSHER'S turf instead of teleporting the victim behind you or tossing from their turf (forwards).
This means some cases where previously you couldn't throw someone because there was a wall behind you will now allow a throw (into the wall, but still).

Crest toss now properly respects the intended throw range, as the victim is actually thrown from the xeno's turf.
The victim now has the hover flag when thrown, so they'll properly fly over cades instead of relying on funny snowflake checks. This also means they won't burst into flame when thrown OVER a flaming turf.

Previously, the throw TARGET was based on the range from the crusher. But the victim being on a different turf meant that max range was actually further than this. This caused no issues when throwing in cardinal directions as it would just reach the target turf, but on diagonal throws this would cause them to be thrown further than expected.

Note that while this is a nerf, the diagonal toss range is still considerably better than it used to be prior to my euclide diagonal throw change.

Also just generally cleaned up the code as it was god awful.
## Why It's Good For The Game
Bug fix good.
## Changelog
:cl:
fix: Crest toss now throws the correct distance diagonally
balance: Crest toss is from the xeno's turf, so some tosses will be possible that were not previously
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
@@ -103,8 +103,6 @@
 	var/turf/target_turf = throw_origin //throw distance is measured from the xeno itself
 	var/big_mob_message
 
-	//if(isclosedturf(throw_origin)) //shouldn't be possible anyway
-	//	return
 	if(!X.issamexenohive(A)) //xenos should be able to fling xenos into xeno passable areas!
 		for(var/obj/effect/forcefield/fog/fog in throw_origin)
 			A.balloon_alert(X, "cannot, fog")
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
@@ -121,12 +121,7 @@
 	if(X.a_intent == INTENT_HARM) //If we use the ability on hurt intent, we throw them in front; otherwise we throw them behind.
 		facing = get_dir(X, A)
 	else
-		for(var/obj/O in throw_origin)
-			if(!istype(O, /obj/structure/barricade) && !O.CanPass(A, throw_origin)) //Ignore barricades because they will once thrown anyway
-				to_chat(X, span_xenowarning("We try to fling [A] behind us, but there's no room!"))
-				return fail_activate()
 		facing = get_dir(A, X)
-	A.forceMove(throw_origin)
 
 	var/turf/temp
 	for(var/x in 1 to toss_distance)
@@ -142,6 +137,7 @@
 
 	succeed_activate()
 
+	A.forceMove(throw_origin)
 	A.throw_at(target_turf, toss_distance, 1, X, TRUE, TRUE)
 
 	//Handle the damage
--- a/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
@@ -99,10 +99,16 @@
 	X.face_atom(A) //Face towards the target so we don't look silly
 	var/facing = get_dir(X, A)
 	var/toss_distance = X.xeno_caste.crest_toss_distance
-	var/turf/T = X.loc
-	var/turf/temp = X.loc
+	var/turf/throw_origin = get_turf(X)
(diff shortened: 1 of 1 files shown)


lKiseki:
Don't really like the fact that this makes mobs thrown by Crest Toss ignore flames on the ground now. Rather uncalled for in my opinion.

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -97,12 +97,18 @@
...
+	var/facing
 	var/toss_distance = X.xeno_caste.crest_toss_distance
-	var/turf/T = X.loc
-	var/turf/temp = X.loc
+	var/turf/throw_origin = get_turf(X)
+	var/turf/target_turf = throw_origin //throw distance is measured from the xeno itself
 	var/big_mob_message
 
+	//if(isclosedturf(throw_origin)) //shouldn't be possible anyway
+	//	return
```
ivanmixo, line 107: Why not just delete?

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -97,12 +97,18 @@
...
+	if(!X.issamexenohive(A)) //xenos should be able to fling xenos into xeno passable areas!
+		for(var/obj/effect/forcefield/fog/fog in throw_origin)
+			A.balloon_alert(X, "cannot, fog")
```
ivanmixo, line 110:
```suggestion
			A.balloon_alert(X, "Cannot, fog")
```

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
```diff
@@ -113,32 +119,16 @@
...
-			to_chat(X, span_xenowarning("We try to fling [A] behind us, but there's no room!"))
-			return fail_activate()
-		if(!X.issamexenohive(A)) //xenos should be able to fling xenos into xeno passable areas!
-			for(var/obj/effect/forcefield/fog/fog in T)
-				A.balloon_alert(X, "cannot, fog")
-				return fail_activate()
-		for(var/obj/O in throw_origin)
-			if(!O.CanPass(A, get_turf(X)) && !istype(O, /obj/structure/barricade)) //Ignore barricades because they will once thrown anyway
-				to_chat(X, span_xenowarning("We try to fling [A] behind us, but there's no room!"))
-				return fail_activate()
```
ivanmixo, line 134: Should've really mentioned you removed this in the PR description, this is pretty huge.

ivanmixo (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
Lumipharon: Forgot

code/modules/mob/living/carbon/xenomorph/castes/crusher/abilities_crusher.dm
Lumipharon: Wdym? This isn't needed anymore, there's no actual in game change. This used to be needed because crest toss would just teleport the mob BEHIND the crusher.

ivanmixo: Also made it so you wouldn't be able to toss with something on the tile behind you, which cucked me personally many times

TiviPlus (approved):

ivanmixo (approved):

PR #14277, opened 2023-10-25
Fixes razorwires from producing infinite barbed wires.

Aporhtonoma:
## About The Pull Request

Currently dismantling 2 razorwire structures returns 8 rods and 2 barbed wires.
The creation of 2 razorwire structures  requires 8 rods and 1 barbed wire.
Infinite barbed wire bad.
changes recipe from 8 rods 1 barbed to 8 rods 2 barbed.

## Why It's Good For The Game

infinite barbed wire bad.

## Changelog
:cl:

balance: razor wires cost 8 iron rods 2 barbed to make
fix: fixes infinite barbed duplication

/:cl:

### DIFF
--- a/code/game/objects/items/stacks/rods.dm
+++ b/code/game/objects/items/stacks/rods.dm
@@ -28,7 +28,7 @@
 		B.use(2)
 		var/obj/structure/razorwire/M = new /obj/item/stack/razorwire(user.loc, 2)
 		to_chat(user, span_notice("You combine the rods and barbed wire into [M]!"))
-
+		return
 	if (iswelder(W))
 		var/obj/item/tool/weldingtool/WT = W
 
--- a/code/game/objects/items/stacks/rods.dm
+++ b/code/game/objects/items/stacks/rods.dm
@@ -21,8 +21,11 @@
 		if(amount < 8)
 			to_chat(user, span_warning("You need at least [8 - amount] more [src] to make razorwire obstacles!"))
 			return
+		if(B.amount < 2)
+			to_chat(user, span_warning("You need [2 - B.amount] more [B] to make a razor wire obstacle!"))
+			return
 		use(8)
-		B.use(1)
+		B.use(2)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/stacks/rods.dm
```diff
@@ -21,8 +21,11 @@
...
 			to_chat(user, span_warning("You need at least [8 - amount] more [src] to make razorwire obstacles!"))
 			return
+		if(B.amount < 2)
+			to_chat(user, span_warning("You need [2 - B.amount] more [B] to make a razor wire obstacle!"))
+			return
 		use(8)
-		B.use(1)
+		B.use(2)
 		var/obj/structure/razorwire/M = new /obj/item/stack/razorwire(user.loc, 2)
 		to_chat(user, span_notice("You combine the rods and barbed wire into [M]!"))
```
QualityVan, line 31:
```suggestion
		to_chat(user, span_notice("You combine the rods and barbed wire into [M]!"))
		return
```

QualityVan (changes requested):
It's kind of ugly having razor wire creation be duplicated in two places like this, but whatever.

code/game/objects/items/stacks/rods.dm
Aporhtonoma: imma be honest you are more code litterate in DM than me so imma trust you on that return

QualityVan: So this is what's called an early return. Since you did everything you wanted to in the attackby proc, you return out as soon as you can so unnecessary stuff you don't want or need to do doesn't happen. If you look down in the iswelder block below it, you'll see the same thing for the same reason.

Aporhtonoma: ah I see should propably do the same to the barbed wire then

QualityVan (approved):
Don't mark stuff resolved on your own in the future, it makes re-reviewing more of a nuisance.

PR #14281, opened 2023-10-26
Attempts to split overwatch cams between factions

Lumipharon:
## About The Pull Request
Requires testing but works as far as I can tell. 
This is somewhat of a guh solution, I'm some something nicer could be done, but honestly camera code in general needs a look at.

This PR splits it out into a 'standard' camera net, and a SOM camera net. Everything from cameras to chunks are now associated with a specific camera net.

Currently the overwatch/AI camera view has visibility on ALL cameras, regardless of network which is uh, very much not ideal in HvH.

## Why It's Good For The Game
Campaign needs this.
## Changelog
:cl:
add: SOM has their own camera net
/:cl:


code/game/objects/machinery/computer/camera_advanced.dm
```diff
@@ -21,12 +22,16 @@
...
-	off_action = new
-	jump_action = new
 	actions = list()
 	for(var/i in networks)
 		networks -= i
 		networks += lowertext(i)
+	if(SOM_CAMERA_NETWORK in networks)
+		parent_cameranet = GLOB.som_cameranet
+	else
+		parent_cameranet = GLOB.cameranet
```
QualityVan, line 32: Leery about this, but given that networks seems to support arbitrary length (why would you want multiple?) it works.

QualityVan (commented):
Camera structure sure is labyrinthine

TiviPlus (approved):

code/game/objects/machinery/computer/camera_advanced.dm
TiviPlus: yea this else is the messiest part

Lumipharon: what would be a better option? I guess I could just have a cameranet var or something, but at that point tying it to the network seems better

PR #14282, opened 2023-10-26
Removes barrel charger as an attatchment option for MG-27

ClosetedSkeletonGH:
## About The Pull Request

removes barrel charger from MG-28

## Why It's Good For The Game

With next to 0 aimmode slowdown encouraging mini-scope usage, this will make the gun a bit worse for being entirely used as an offscreen aimmode weapon

## Changelog

:cl:
balance: Removes barrel charger from MG-27
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/mounted.dm
+++ b/code/modules/projectiles/guns/mounted.dm
@@ -391,7 +391,7 @@
 	deployable_item = /obj/machinery/deployable/mounted
 	gun_firemode_list = list(GUN_FIREMODE_AUTOMATIC)
 	actions_types = list(/datum/action/item_action/aim_mode)
-	aim_fire_delay = 0.15 SECONDS
+	aim_fire_delay = 0.1 SECONDS
 	aim_speed_modifier = 5
 	soft_armor = list(MELEE = 0, BULLET = 50, LASER = 50, ENERGY = 50, BOMB = 50, BIO = 100, FIRE = 0, ACID = 0)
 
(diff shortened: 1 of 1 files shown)


Lumipharon:
This makes it worse than the mg60 when deployed.

Lumipharon:
RDS alone does nothing for the MG27 aim ROF due to byond ticks. The MG60 has the same dps as the MG27 if both are deployed.
MG27 is better when not on aim mode (which this doesn't touch) or when compared when using miniscopes.

This also makes it hilariously worse than prebuff.

Lumipharon:
I removed the dnm, but honestly the more useful nerf would be to remove the BC. As stated, at closer range the T60 is already just as good, so this pr will make it worse. However with miniscopes (and so at further range) the MMG is considerably better than the T60.

iforgotmeword:
"MMG is considerably better than the T60."
isn't that point? you're being forced TO STAY still and have a deployment time 
it's like nerfing the the rocket launchers because they are considerably better than recoilless launchers

Xander3359:
image
Checks are failing because you removed /obj/item/attachable/heavy_barrel from attachable_allowed, however, you did not remove it from the gun that some loadouts have it on at line 413.

image

Lumipharon (approved):

Xander3359 (approved):

antropod:
This PR made the gun useless, xenos are just outrunning bullets now

PR #14297, opened 2023-10-27
Creates CEILING_OBSTRUCTED, a middle ground between caves and not caves, and applies it to Lawanka Outpost's Engineering.

dopamiin0:
## About The Pull Request

Creates a new type of ceiling, CEILING_OBSTRUCTED. An obstructed ceiling acts identically to a metal ceiling, with the exception that it counts as caves for the purposes of tadpole and droppod landing. Changes the ceiling type of Lawanka Outpost's Engineering area from CEILING_UNDERGROUND_METAL to CEILING_OBSTRUCTED.

## Why It's Good For The Game

Over time, marines have gotten more and more flexible in how and where they can deploy. The tadpole and droppods both allow for large amounts of marines to be deployed virtually anywhere on the map, and in the case of the tadpole, instantly have moderate levels of fortification there. Previously, the only tool for map makers to deny an entire ungaball from appearing in an area was to mark it as underground. However, this has the unfortunate side effect of denying or constricting several other features (requisitions, radio, CAS, OB, artillery, etc.) Obstructed areas allow map makers to create areas that marines cannot directly deploy into without affecting other features.

Engineering on Lawanka Outpost is a weird disk, hovering right on the edge of caves (not dissimilar to engineering on 624.) Balance concerns have been raised with marines being able to tadpole directly inside it, making the disk much more difficult for xenomorphs to contest. However, the area surrounding it to one side has a large amount of open sky, and completely treating the area as caves turns the balance in the opposite direction, creating a sudden dramatic bottleneck marines HAVE to either push through or die that they no longer have access to their traditional fire support tools for. By setting the area to obstructed, both problems are solved simultaneously; xenomorphs don't have to worry about a tadpole landing instantly winning the disk, while marines don't have to worry about xenomorphs forcing them through a powerful chokepoint they can't use their traditional tools to clear.

## Changelog
:cl:
add: Added a new ceiling type, obstructed, that blocks tadpole and droppods but not other things caves ordinarily block.
balance: Engineering on Lawanka Outpost now has an obstructed ceiling, instead of being caves.
/:cl:


code/game/objects/items/binoculars.dm
```diff
@@ -221,6 +221,8 @@
 				is_outside = TRUE
 			if(CEILING_METAL)
 				is_outside = TRUE
+			if(CEILING_OBSTRUCTED) //given that our values for these are integers i dont know why we don't just check if it's less than a certain value but i dont know enough about dm to change that safely
+				is_outside = TRUE
```
Lumipharon, line 225: yeah don't follow the old shitcode

Lumipharon (approved):
lgtm other than that. I assume you've checked every use of ceiling so this isn't messing anything up, right?

Barnet2 (approved):
Lewanker creator here, seems good to me.
In the end up to Grayson

dopamiin0:
> lgtm other than that. I assume you've checked every use of ceiling so this isn't messing anything up, right?

i have scanned like four times now. if i have missed something i will feel very stupid

code/game/objects/items/binoculars.dm
```diff
@@ -214,13 +214,8 @@
...
 	if(is_ground_level(TU.z))
-		switch(targ_area.ceiling)
-			if(CEILING_NONE)
-				is_outside = TRUE
-			if(CEILING_GLASS)
-				is_outside = TRUE
-			if(CEILING_METAL)
-				is_outside = TRUE
+		if (targ_area.ceiling <= CEILING_OBSTRUCTED)
+			is_outside = TRUE
```
Lumipharon, line 218: Just combine this into a single line. if ground side and ceiling less than obstructed

dopamiin0: > Just combine this into a single line. if ground side and ceiling less than obstructed

no because im pretty sure shipside maps dont have ceilings and if i ask both at once i think itll be mad and runtime or something which is why i did it this way

Lumipharon: bruh.

dopamiin0: > bruh, its false by default.

i will trust you

PR #14299, opened 2023-10-27
Buffs CAS export percentage point gain from 5% to 10%.

dopamiin0:
## About The Pull Request

As in title.

## Why It's Good For The Game

In my original PR, many of the cost adjustments were made with the idea that marines would be generating roughly 2000 points per 10 minutes, or an average of 200 points a minute, as well as holding 2.5 miners at a time. In practice, it seems this was an overestimation, and marines generate less than this total. This means that the alterations made to the prices of several items, originally only intended to cause items per minute to stay the same (while rewarding exceptional marines and punishing ones not selling bodies) instead just nerfed CAS economy across the board. By bumping it to 10% (in parity with the miner percentage) CAS should ideally have roughly the same effective generation as before after exports are factored in.

## Changelog
:cl:
balance: The dropship fabricator now recieves 10% of the value of exports, rather than 5%.
/:cl:

### DIFF
--- a/code/modules/requisitions/supply_export.dm
+++ b/code/modules/requisitions/supply_export.dm
@@ -5,7 +5,7 @@
 		return FALSE
 
 	SSpoints.supply_points[faction_selling] += points
-	SSpoints.dropship_points += points * 0.08
+	SSpoints.dropship_points += points * 0.1
 	return new /datum/export_report(points, name, faction_selling)
 
 /mob/living/carbon/human/supply_export(faction_selling)
--- a/code/modules/requisitions/supply_export.dm
+++ b/code/modules/requisitions/supply_export.dm
@@ -5,7 +5,7 @@
 		return FALSE
 
 	SSpoints.supply_points[faction_selling] += points
-	SSpoints.dropship_points += points * 0.05
+	SSpoints.dropship_points += points * 0.08
 	return new /datum/export_report(points, name, faction_selling)
 
 /mob/living/carbon/human/supply_export(faction_selling)


ivanmixo:
Slowly creeping the points cas gets is not cool

Oscari-pog:
Didnt CAS already get one buff to points that makes it more annoying than it has any right to be

dopamiin0:
> Slowly creeping the points cas gets is not cool

this is after already making a change that drastically increased the cost of items. widowmakers literally cost triple what they did before. when i made that change, i was aiming it so that after calculations, they would still roughly be getting the same amount of points. my calculations missed the mark, and as a result, CAS was effectively nerfed by my change. this isn't a buff, it's an un-nerf of an unneeded nerf.

UncleBourbon:
tgmc players try to use their fucking brains challenge

dopamiin0:
By the way, to anyone viewing at home; with how the price increases were calculated, marines are expected to be selling 1000 points per 10 minutes, or 100 points per minute, for CAS to be breaking even after this change. (In practice, this is actually an even higher number, as I tended to round up when I was increasing the cost of items.) With the current 5%, they are currently expected to be selling twice that. Spoiler alert: marines don't sell 200 points per minute on average.

aarrisu:
nah

dopamiin0:
> nah

“I sent plagues among you as I did to Egypt. I killed your young men with the sword, along with your captured horses. I filled your nostrils with the stench of your camps, yet you have not returned to me,” declares the LORD.

dopamiin0:
> cas is already usually pretty ineffective, making them ineffective and also constrained by a tight budget to do _anything_ just means a player slot will get increasingly more useless
I don't disagree, but to clarify, this PR is aiming to fix that budget issue, not increase it.

> probably could make some of the more basic munitions like gau rounds and HE/flare rockets be timer/recharge based instead of bought but that's out of scope
Would be nice, but again, out of scope.

dopamiin0:
> This PR has been inactive for long enough to be automatically marked as stale. This means it is at risk of being auto closed in ~ 3 days, please address any outstanding review items and ensure your PR is finished, if these are all true and you are auto-staled anyway, you need to actively ask maintainers if your PR will be merged. Once you have done any of the previous actions then you should request a maintainer remove the stale label on your PR, to reset the stale timer. If you feel no maintainer will respond in that time, you may wish to close this PR youself, while you seek maintainer comment, as you will then be able to reopen the PR yourself

What if I killed you. What if I fed you to your hounds.

ivanmixo:
This is just trying to justify buffs with marine stupidity, I'm still against this

PR #14301, opened 2023-10-27
Removes limited construction items from GHMME vendor pouches and adds them to engivend + Engivend goofening

ClosetedSkeletonGH:
## About The Pull Request

Yes I know this is an extensive PR but atleast try to read it before commenting about waaa waaa u nerf me favorite meem strat. This is all in one PR because to try and split it up would leave huge holes of balance.

**Note, this PR has gone through significant modifications, look through it again**

All changes:

YouTool:
-killed and combined into the engivend, all map refs of it were of no consequence for this change

Engivend:
-gets shared inventory
-cannot be wrenched now (I dont think it could before either)
-now contains an all tools tab 
-loadout vendors now allow engivend stuff to be saved to them
-entrenching tool has been added to engivend tools tab (marines were having a hard time finding it, figured this cant hurt)

-has a new engineering supplies tab
-has 25 vends of 5 sandbags (125 total, or 25 sandbag cades worth, I am open to change this number)
-has 5 vends of 5 bob wires (25 total, or 25 cades worth, I am open to change this number but kept it the same in order to combo with the sandbags (but may swap to 30 cause of marines using cablecoils on helmets for style))
-unlimits the amount of hardhats avaliable for drip

-adds new misc gear tab that just has all the useless stuff engivend did before

---

GHMME Vendors:
-No longer vends full construction pouches

---

Map changes:
-random spawns between youtools and engivends now can only be engivend (or other alt options)
-all youtools replaced with engivends

Ship map changes:
-All preps but Theseus' had engivends added to them  (theseus was actually good with this)
-_Unrelated QOL: Arachne has a new cryo computer thats actually findable (arachne prep is so annoying, this has been bothering me so much, sorry for no atomization)_

## Why It's Good For The Game

Groundmap changes:
-these are all of no real consequence but makes handling mapping space easier

shipmap changes:
-all QoL for vending in prep and Req

Engivend changes:
-the fact youtool and engivend were seperate before was kinda dumb

-getting 25 sandbag cades here means marines roundstart will want to get their stuff if theyre doing a sandbag relevant build, I chose 25 because 5 MG-27's round start and that means each gets 5 cades to buildthemselves a good nest
-getting 25 wires just combos with the cades

-unlimited helmet for drip

GHMME changes:
[...]

PR #14304, opened 2023-10-28
Prevents the surplus clothing vendor from being moved and allows it to vend full medpouches/limited autoinjector pouches

ClosetedSkeletonGH:
## About The Pull Request

Change 1:
-There are now two types of autoinjector pouches, basic and first aid. Basic is everything but synap and RR, firstaid is how it is now

Change 2:
-BASIC (no rr or synap) autoinjector pouches and filled medpouches are now avaliable in the surplus clothing vendor

Change 3:
-Clothing vendor can no longer be moved

If its not in this list, it wasnt changed (ERT's still have the same loadout)

edit for clarification: GHMME autoinjector literally hasnt changed at all. why do I even have to say this

## Why It's Good For The Game

This was previously in #14301 but I decided to atomize that PR. This is mostly QoL. It vends nothing limited and ingame, the FURTHEST a clothing vendor is from a medvendor is 9 (!!) tiles (ignoring 1 outlier lol). This is mostly for new folk since I've had PVT's be confused at why their GHMME medpouch has meds but not their clothing one and marinemed can be a bit daunting.

also no more infinite flares/infinite fuel groundside for free 

There are 0 (zero) groundside clothing vendors so this effects balance in the way that you get your medpouch 6 seconds slow (2 seconds to walk to medvendor and 4 seconds to click down a line of the meds). God bless

This PR also means loadout vendor gets a lot less fucked up with medpouches cause right now it wont give you one if you've already taken two other pouches from it

## Changelog

:cl:
balance: Clothing vendor can no longer be moved
QoL: autoinjector pouches and full medpouches can now be vended from the clothing vendor
/:cl:

### DIFF
--- a/code/game/objects/items/storage/pouch.dm
+++ b/code/game/objects/items/storage/pouch.dm
@@ -518,7 +518,7 @@
 	desc = "Standard marine first-aid injector pouch. Specialized to store only auto-injectors. Contains basic injectors, a stabilizing injector, stimulant injector, and an emergency injector."
 
 /obj/item/storage/pouch/medical_injectors/basic
-	desc = "Standard marine autoinjector pouch. Specialized to store only auto-injectors. Contains basic injectors and a stabilizing injector,
+	desc = "Standard marine autoinjector pouch. Specialized to store only auto-injectors. Contains basic injectors and a stabilizing injector."
 
 
 /obj/item/storage/pouch/medical_injectors/basic/Initialize(mapload)


FosterGrime-Heymynameslouie:
moving around clothing vendors and armor vendors in req is really useful and having groundside clothing vendor for the occasional guy needing something from it is convinient, I dont see the purpose as to why we have to anchor it

FosterGrime-Heymynameslouie:
> > moving around clothing vendors and armor vendors in req is really useful and having groundside clothing vendor for the occasional guy needing something from it is convinient, I dont see the purpose as to why we have to anchor it
> 
> RO has clothing vendors in all of their requisitions, having this change will also mean loadouts get less fucked up with med pouches as opposed to the way they are currently.
> 
> edit: also when do you ever need new clothing groundside? clothes don't break ever? If you really need something that bad RO can send it? Ive seen clothing vendors get brought groundside a few times but ive never seen them used. Armor vendors on the other hand I see used plenty as they have modules and actual use since people may want to swap for heavier/lighter armor

they get used alot, specifically the flare pouches 
every single tad op has them, every good tad op atleast, m22 welder kits in FOB on demand is also pretty good, makes the RO job alot easier granted they dont have to answer request on flare and fuel refills every 5 minutes

also the clothing vendor and vendors being wrenchable in general makes it so that ROs have their own setups on what is the most optimal vending areas for them, I move my vendors all the time because their always miles apart from where they need to be

all in all I really think the unwrenchable part, here and in the engineering vendor pr is needed because it just kills a huge part of optimization and ease of access to the guys who actually use it, and in the name of what to?

Al-1ce:
Moving lockers closer to the Alamo/groundside/wherever is both very useful and still very common. I don’t think you’ve shown a good enough reason to restrict them like this.

-1

Arctrooper8:
I still think this is bad for all the reasons I listed in https://github.com/tgstation/TerraGov-Marine-Corps/pull/14301

However I have to ask why you're now also removing the first aid not only from marines but from campaign asset rewards and ERTs? These groups literally don't have chemistry or access to marine med.

Arctrooper8:
> > I still think this is bad for all the reasons I listed in #14301
> > However I have to ask why you're now also removing the first aid not only from marines but from campaign asset rewards and ERTs? These groups literally don't have chemistry or access to marine med.
> 
> Please read the PR. This doesn’t do literally any of these things you listed. I really can’t even make it clearer.
> 
> It leaves the first aid one in the GHMME, people still get synap and RR. It also doesn’t even touch ERT’s. It just makes clothing pouch ones give out a non-synap and RR version
> 
> This opens it up to future changes but doesn’t make any of those changes

I misread because you've still got that weird new path and organization changes, I still don't see why you didn't just make /basic it's own new subtype of /injectors instead of forcing it to be the parent of /firstaid?

Lumipharon:
You really need to atomise properly. Not being able to move the surplus vendors has nothing to do with the pouch changes.

Also I'm not sure why we need filled pouch in the surplus vendor in the first place, but even then I have no idea why you've changed the type path of the existing one instead of just creating a new one.

Al-1ce:
> Having access to infinite flares and welding fuel groundside without req cooperation is not a good thing and only further provides why they should be locked

”It’s not a good thing”

* >doesn’t actually say why it isn’t a good thing

Funneling fuel back and forth is repetitive and boring, seldom having existed in our past as a server as a relic from C-M. Even *if* you had actually made a point here, there is nigh-infinite amount of fuels in canisters scattered around every single fob. This is something that has been trivialized years ago because it was boring to keep it as it was.

PR #14311, opened 2023-10-29
CL no longer gives shock

Xander3359:
## About The Pull Request
Removes Cloneloss from giving shock.
## Why It's Good For The Game
Reducing your max HP by a certain amount is already enough punishment. I don't think further punishing by giving you shock as well is needed.
## Changelog
:cl:
balance: Cloneloss no longer gives shock
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -57,12 +57,11 @@
 		traumatic_shock = 0
 		return
 
-	traumatic_shock = 			\
-	0.75	* getOxyLoss() + 		\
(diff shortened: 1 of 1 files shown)


ivanmixo:
Heavy disagree on this one

Zangeiti:
Heavy agree on this one

Lumipharon:
Cloneloss doesn't need to be less impact full.

QualityVan:
I do think it makes sense to decouple cloneloss from pain, but that's more due to shortcomings with the pain/painkiller system than anything else. Certainly if this were to go through I'd want to see some other replacement effect.

PR #14315, opened 2023-10-29
Removes the stalemate system

Barnet2:
## About The Pull Request
Removes the stalemate system that procs rarely to decrease respawn timer by 5 and give xenos a larva buff
## Why It's Good For The Game
image
I don't think it has ever worked, and the game handles "stalemates" better than what the system can do, in req and psy points. How the game determines a stalemate is dubious as well
## Changelog
:cl:
del: Removed stalemate system
/:cl:


yyzsong:
I actually think xenos getting 150 larva out of the blue is a good gameplay mechanic, actually.

Lumipharon:
this is almost certainly a bug caused due to null entries in the observer list or whatever its checking.
Whether or not the system is a good idea, I'm not going to remove it due to a bug.

If its still an issue after the bug is fixed, we can revisit.

Pariah919:
I think the issue is that like, even just removing 5 minutes of respawn just makes you respawn like 20% faster is like nothing compared to just getting say 2 larva out of the blue

TiviPlus:
If you're not gonna merge it then close it

PR #14318, opened 2023-10-29
Adds syringe case to NTMED

Xander3359:
## About The Pull Request
Adds Syringe cases to the "Chemistry Equipment" Section of the NT Med.
## Why It's Good For The Game
These can hold at most 180u if you fill it with 3 bottles, which is still a lower capacity than a pill bottle.
On top of that, you can't directly drink liquid out of a bottle, so you need to either keep the syringe it comes with, or carry an autoinjector that you can refill.
## Changelog
:cl:
add: Syringe cases in the NT Medplus
/:cl:


Derrick2Electricboogaloo:
Yeah something I never noticed until after playing TG, no idea why we can not just drink out of it. Would be nice.

PR #14323, opened 2023-10-30
Makes Vali require 1u of Synap instead of 3u.

Xander3359:
## About The Pull Request
Makes Vali require 1u of Synap instead of 3u to receive the effect of the activation.
## Why It's Good For The Game
This feels more like a fix, because it seems whoever made vali confused the 3u injection of a synap injector to mean that 3u of synap was injected.
A single dosage of Synaptizine is 1u, so vali should be using that value.
## Changelog
:cl:
balance: Vali uses 1u of Synaptizine instead of 3u
/:cl:


ilikesaying:
going over 2u of Synap made you take Stamina damage so IDK why Vali even required such a debilitating amount for the boost

Strayed-Raven:
What are you doing. This is something that actually makes better sense than what we have already. Powercreep alert! *snicker*

PR #14324, opened 2023-10-30
Reduces engineer point cost for st580 point defense sentry

Yttriums:
## About The Pull Request

Reduces the cost of the ST580 point defense sentry in engineer vendor from 50->40
## Why It's Good For The Game

This is currently a weak option in terms of req value (relative to mats and pcs).

Setting this value at 40 will still restrict engineers to 1 turret while allowing them to take some more mats and QOL items.

Should make this a more viable choice.
## Changelog
:cl:
balance: reduced engineer vendor cost for point defense sentry from 50->40
/:cl:


Yttriums:
someone pointed out that you can buy 580 and bas so maybe this is naive, nevermind

PR #14357, opened 2023-11-01
Make hyperscale init slightly faster

vvvv-vvvv:
No more millions of calls to ReadRGB on init. Something like ~3s faster on my machine


TiviPlus:
Good change, is this also done upstream?

vvvv-vvvv:
I don't think hyperscale exists upstream

TiviPlus:
ah wait I thought you were editing base /layer nvm

PR #14363, opened 2023-11-02
Fixes throw impact not respecting speed

DeltaFire15:
## About The Pull Request
Title.
During Lumi's throw changes, the wall impact was changed to have a "default" speed of 5 if no argument was passed.
Surprise, the impact **never** had the speed passed as impacting things happens in bump (which is outside of the throw itself).
Thus, all impacts were handled as if they were executed with a speed of 5 (resulting in five times higher damage in most cases).
This PR fixes that by preserving current thrown speed.
Unfortunately, has to be a movable var due to no viable vectors to pass it existing (tgmc does not have thrownthing datums like base tg, which would handle this).
## Why It's Good For The Game
args actually being passed seems kinda useful.
## Changelog
:cl:
fix: throw_impact() now actually receives the speed arg.
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -15,6 +15,8 @@
 	var/drag_windup = 1.5 SECONDS
 	var/throwing = FALSE
 	var/thrower = null
+	///Speed of the current throw. 0 When not throwing.
+	var/thrown_speed = 0
 	var/turf/throw_source = null
 	var/throw_speed = 2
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

Lumipharon:
Just a note that this may cause... extreme problems in certain niche situations. namely that 5 is the default speed, but defender charge throws mobs at 70 speed, likely causing instant death in some cases.

PR #13317, opened 2023-06-20
Gorger buff of the year (RUTGMC port)

Blundir:
## About The Pull Request

- Feast now works on all xenos in 4 tile radius, cooldown decreased from 180 to 30
- Feast healing now has visual effect when it's active
- Drain and Carnage drain real blood from people (30u)
- Transfusion range increased from 2 tiles to 3
- You can move while casting Transfusion
- Psychic Link cooldown from 50 to 15
- Psychic Link doesn't require to stand still
- Gorger now has passive plasma regen up to 10% of maximum plasma

dreamseeker_41t1jVpGkr

100% port of https://github.com/Cosmic-Overlord/TerraGov-Marine-Corps/pull/122 and https://github.com/Cosmic-Overlord/TerraGov-Marine-Corps/pull/129

## Why It's Good For The Game

Gorger is really uncomfortable to play in general, these are merely QoL considering what state he is in right now, Gorger cannot sustain effectively neither offensive nor support abilities, some abilities are very rarely used in general, this tries to correct that (very successfully, considering RU experience with it). Probably needs a testmerge to estimate how good/bad these changes are. These changes are present on RU for 3 months without any balance issues.

## Changelog
:cl:
add: Drain and Carnage drain real blood from people (30u)
add: Feast now works on all xenos in 4 tile radius
qol: gorger can cast transfusion while moving
qol: gorger regenerates 10% of blood automatically
qol: Psychic Link doesn't require to stand still
balance: Transfusion range increased from 2 tiles to 3
balance: Feast cooldown lowered  from 180 to 30
balance: Psychic Link cooldown lowered from 50 to 15
/:cl:

PR #14384, opened 2023-11-05
Gives gorger passive blood regen

Xander3359:
## About The Pull Request
Atomization: https://github.com/tgstation/TerraGov-Marine-Corps/pull/13317
Gives Gorger passive blood regen up to 10%
## Why It's Good For The Game
Copypaste: "Gorger is really uncomfortable to play in general, these are merely QoL considering what state he is in right now, Gorger cannot sustain effectively neither offensive nor support abilities, some abilities are very rarely used in general, this tries to correct that (very successfully, considering RU experience with it). Probably needs a testmerge to estimate how good/bad these changes are. These changes are present on RU for 3 months without any balance issues."

## Changelog
:cl:
balance: Gorger passively regens blood up to 10% blood
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/gorger/castedatum_gorger.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/gorger/castedatum_gorger.dm
@@ -19,8 +19,8 @@
 
 	// *** Plasma *** //
 	plasma_max = 400
-	plasma_gain = 0
-	plasma_regen_limit = 0
+	plasma_gain = 5
+	plasma_regen_limit = 0.1
 	plasma_icon_state = "fury"
 
 	// *** Health *** //


ivanmixo:
Yea edit that in or this is getting closed. 


Also this directly hurts the caste since you want to be low on blood for carnage to stun but wharever

Cerrik:
You only need to be below 50% for carnage to stun, no?  This PR only goes up to 10% blood.  At least, that's described how it works from the original PR this is atomizing.

AlexCrimson:
Considering i almost never see a Gorger overhealing other Xenos, i do not see the issue. Blood is such a hard resource to gather, but overhealing is supposed to be one of the castes main gimmicks. Passive regen up to 10% isnt even a lot of blood.

Cerrik:
Unironically, that is a skill issue argument.  The only situation I have trouble getting blood is FOB sieges.  If you can melee a marine, you can collect blood.

AlexCrimson:
Maybe it is a skill issue argument. Sadly, the entire Xenos player population isnt as cool as you, so they need shit to be a bit easier. Gorger is probably the most unpopular T3 caste, because it sucks to play. If you are good enough to always have blood, then this change is not going to make you better. If you struggle to farm blood for whatever reason, this is a good little buff to get you started.

Cerrik:
I'm not cool and I'm not that good, but I can get into melee range of marines (thus, I can collect blood), which is the bare minimum requirement to play xeno.

AlexCrimson:
Gorger max blood is 400. Drain gives you 80 if you land both hits. 10% passive blood is such a minor thing. Its enough for 2 Transfusions, which would give one xeno max overheal, i think. This just gives Gorgers a way to use their healing without needing to have blood stored up, and even then its an incredibly small amount of healing.

Considering Gorger is a T3 healer of sorts, its woefully bad at the role.

Cerrik:
I'm just going to point to my first comment: it all depends on the regen rate.  10% is only two transfusions... until it regens back up to 10% again.  It's infinite transfusions, only limited by the regen rate.

AlexCrimson:
>These are bad gorgers that aren't using one of their most powerful tools

Isnt that the point though? Good Gorgers will not benefit from this. They will have more blood because they are attacking more. Bad Gorgers will just actually get to contribute something to the team and at least overheal some Xenos. 

The biggest issue i see with this is the fact that when Xenos hijack, if there is a Gorger, the downtime & big group of Xenos means everyone will be overhealed by the time the ship crashes. Thinking about that, i could see this being too much.

Pariah919:
This just makes blood into plasma

Arturlang:
Instead maybe make draining marines give more blood?

chizzy376:
Isn't this a nerf to his stun which is the main part of his kit?

Xander3359:
Yeah this is probably just not needed at all tbh,

PR #14388, opened 2023-11-05
Gorger transfusion buff

Xander3359:
## About The Pull Request
Atomization: https://github.com/tgstation/TerraGov-Marine-Corps/pull/13317
Increase Transfusion range by 1
Can cast transfusion while moving
## Why It's Good For The Game
Gives Gorger a lot of quality of life, being able to be able to heal through shuffle makes overhealing xenos less painful.

## Changelog
:cl:
balance: Gorger can move during transfusion
balance: Gorger has 1 more tile range for transfusion
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/gorger/abilities_gorger.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/gorger/abilities_gorger.dm
@@ -197,7 +197,7 @@
 
 	if(owner.do_actions)
 		return FALSE
-	if(!line_of_sight(owner, target_xeno, 2) || get_dist(owner, target_xeno) > 2)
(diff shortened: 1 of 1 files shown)


ivanmixo:
Quite powercreepy considering this means neither the target nor the gorger have to stand still

Cerrik:
This would be an incredibly insane buff.  It's already a powerful tool to heal another xeno mid combat to keep them going, but having increased range and only being cancelled by stagger means you could keep a xeno fighting for an unreasonable amount of time.

AlexCrimson:
This is probably the only one of the buffs i would consider to be OP. The range is fine, but whilst moving? That is too much imo. At most, it should only be the target that can move. Even then it might still be too much.

PR #13990, opened 2023-09-20
Campaign gamemode

Lumipharon:
## About The Pull Request
Still heavily WIP. Message me if you want to help with mapping or spriting.

Adds a new gamemode. Might be terrible.

I'll explain in some more detail but here's a brief rundown for now.

### **About the game mode**
Campaign is an attempt to significantly depart from our normal gameplay loop.
Instead of a shipside and groundside, campaign revolves around two factions having multiple battles against each other, across and around a planet in an attempt to wrestle control of them planet for themselves.

### Missions
Gameplay is divided into a series of missions, each with its own set of objects, particular scenario, and rewards.
When a mission is loaded, **a new z-level is loaded specifically for that mission**. So as a campaign progresses, players will fight across a whole range of maps, trying to achieve a variety of goals to ultimately drive the other faction of out of the system.

Deployments is typically via tad rappel like existing HvH modes, but TGMC have access to drop pods (including mech and sentry pods) while SOM have an AOE bluespace teleporter.
image
image

The faction that wins a mission then has the opportunity to select the next mission that is played, choosing from a number of options they have available to them.

image

### Victory Points
Victory Points (VP) are accrued by winning missions, which either allow the faction to play a 'end mission' to conclusively win the campaign, or otherwise grind down the enemy team by scoring a VP (minor) victory.

While VP is the ultimate points of the gamemode, it is not the only thing played for during missions.

### Campaign Assets
Both factions have access to a variety of campaign assets to help them in their fight. These can be won as mission rewards, or purchased with attrition points (AP).

Assets can range from gear upgrades, deployment options, allied support to air support.

image
image

image

Some "assets" can be negative, and are usually assigned to the **opposing** team as a reward for certain mission which can do things like reduce your AP generation rate, to permanently disabling certain assets you would normally have access to.

### Attrition Points
While attrition points can be spent on assets, but their primary purpose is to allow for player respawns.
Every mission, your faction leader determines how many attrition points will be spent on respawns. Unspent points are lost however, so simply spamming points can be wasteful, and a team on the backfoot could even deliberately avoid spending points to save them up for a cheeky counter attack in a later mission, etc.

### Code stuff
Thanks to @TiviPlus for the tgui work, which I am utterly useless at.

Codewise, Campaign is not hardcoded to specific factions like existing modes are, so changing the factions (to sectoids vs CLF, marines vs xenos, etc) is extremely easy.

[...]

PR #14392, opened 2023-11-05
Fix adding slots in job slots panel

vvvv-vvvv:
## About The Pull Request

Job titles are supposed to be unique because of SSjob.name_occupations which expects unique keys. Up until #13990, this was somehow working due to init order

## Changelog

:cl:
fix: Fix adding slots in job slots panel
/:cl:


Lumipharon:
this doesn't mess with their access to role vendors does it?

vvvv-vvvv:
> this doesn't mess with their access to role vendors does it?

nope

Lumipharon (approved):

Lumipharon:
> > this doesn't mess with their access to role vendors does it?
> 
> nope

ok, I know title effects something, but I think it may be map icons which doesn't matter for valhalla anyway so ye.

KM-Catman:
No idea how to contact you (Lumi didn't tell me), but your PR broke Loadout Vendors in Valhalla since you made the Fallen job's titles different. Do you know if there is any way to fix that with a simple solution, or would it require fully changing how the Loadout Vendor dispenses loadouts?

vvvv-vvvv:
I'll look

PR #14403, opened 2023-11-07
Adjusts harvester chemicals

Xander3359:
## About The Pull Request
Changes Vali chemicals as follows:
- All chems now have the damage bonus that used to be strictly on Tramadol
- Bicaridine heal scales off medical skill.
- Bicaridine will self-heal when you attack a xeno
- Kelotane is just a fire stack
- Tramadol only has the additional slow (functionally unchanged)
- Tricordrazine will sunder the same amount as a shotgun slug
- Tricordrazine will cause shatter for 3 seconds

## Why It's Good For The Game
Normalizing the damage bonus from reagent means that Tramadol is no longer the only viable pick. You would decide to switch your chems based on what effect you are looking for instead of just picking the one with the highest damage. This also means that it's easier to add chems in the future now that the support is there.
## Changelog
:cl:
add: Harvesters can use tricordrazine, which applies a shatter effect
balance: All chems in a harvester will do the same damage bonus
balance: Bicaridine in a harvester will have more healing if you finish the channel with higher medical skill
/:cl:


Barnet2:
I said it on Discord so I'll say it here too:
Weird that you are giving the previous Tram damage bonus to all weapons (if I am reading this right); the issue with tram before is that it did two things well at once compared to the other two, and wasn't all fun to fight (since you know, near immortality if no xeno with ozy). Now you uh, gave that to all of the weapons. I would much rather see Vali as a support weapon, and the changes other than the added damage do that quite well.

Pariah919:
Not the vali buff we deserve... Or really need honestly.

bowlofcereal:
This is actually a really good idea and change, kudos

BonkaiB:
>  I would much rather see Vali as a support weapon, and the changes other than the added damage do that quite well.

Melee.... support...

yyzsong:
make tramadol purple

Kuro020 (approved):
it only took 600 years for someone to do this

Oscari-pog:
"rework" this is a sprite change and a direct buff lmao

Lumipharon:
zz, why'd you do the walance change with the visual one

code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -1215,6 +1215,7 @@
 			/obj/item/storage/holster/t19 = -1,
 			/obj/item/storage/holster/blade/machete/full = -1,
 			/obj/item/storage/holster/blade/machete/full_harvester = -1,
+			/obj/item/weapon/twohanded/glaive/harvester = -1,
```
Lumipharon, line 1218: ?

code/datums/components/harvester.dm
```diff
@@ -225,26 +216,23 @@
 	if(target.status_flags & INCORPOREAL || user.status_flags & INCORPOREAL) //Incorporeal beings cannot attack or be attacked
 		return
 
+	if(isxeno(target))
```
Lumipharon, line 219: why's this only work on xenos?

code/datums/components/harvester.dm
```diff
@@ -225,26 +216,23 @@
...
-			if(isxeno(target))
-				return
-			INVOKE_ASYNC(src, PROC_REF(attack_async), source, target, user, weapon)
-			. = COMPONENT_ITEM_NO_ATTACK
+			INVOKE_ASYNC(src, PROC_REF(attack_bicaridine), source, target, user, weapon)
+			if(!(isxeno(target)))
+				. = COMPONENT_ITEM_NO_ATTACK
+
+		if(/datum/reagent/medicine/tricordrazine)
+			target.apply_status_effect(/datum/status_effect/shatter, 3 SECONDS)
```
Lumipharon, line 235: shatter is quite good, but seems kinda wack vs just using kelo for sunder and fire.

Lumipharon (changes requested):

code/game/objects/machinery/vending/marine_vending.dm
Xander3359: Something something I didn't pull master I think.
I'll fix this now.

code/datums/components/harvester.dm
Xander3359: Do you mean this is weak when compared to kelo?
Or that kelo is too strong compared to this.
I don't really know how strong of an effect Shatter is, but anytime I've gotten shattered as a xeno I always took that as a sign to GTFO, so I assumed the effect was quite powerful enough to not need any additional effect to accompany it.

Lumipharon (approved):

code/datums/components/harvester.dm
```diff
@@ -259,6 +248,25 @@
...
+		target.apply_damage(weapon.force*0.6, BRUTE, user.zone_selected)
+		user.adjustStaminaLoss(-30)
+		user.heal_overall_damage(5, 0, updating_health = TRUE)
+		return
+
+	to_chat(user, span_rose("You prepare to stab <b>[target != user ? "[target]" : "yourself"]</b>!"))
+	new /obj/effect/temp_visual/telekinesis(get_turf(target))
+
+	if(do_after(user, 2 SECONDS, TRUE, target, BUSY_ICON_DANGER)) //Channeled heal on human strike
+		var/skill_heal_amt = user.skills.getRating(SKILL_MEDICAL) * 5
```
Lumipharon, line 264: why does this scale with med?

code/datums/components/harvester.dm
```diff
@@ -259,6 +248,25 @@
...
+		target.heal_overall_damage(10 + skill_heal_amt, 0, updating_health = TRUE) //5u of Bica will normally heal 25 damage. Medics get this full amount
+	else
+		target.adjustStaminaLoss(-30)
```
Lumipharon, line 267: I know you didn't do this part, but why the fuck does it heal stam if you SKIP the do_after?

code/datums/components/harvester.dm
```diff
@@ -231,20 +222,18 @@
...
-					victim.flamer_fire_act(10)
-					victim.apply_damage(max(0, 20 - 20*victim.hard_armor.getRating("fire")), BURN, user.zone_selected, FIRE)
 
 		if(/datum/reagent/medicine/bicaridine)
-			if(isxeno(target))
-				return
-			INVOKE_ASYNC(src, PROC_REF(attack_async), source, target, user, weapon)
-			. = COMPONENT_ITEM_NO_ATTACK
+			INVOKE_ASYNC(src, PROC_REF(attack_bicaridine), source, target, user, weapon)
+			if(!(isxeno(target)))
```
Lumipharon, line 231: this sohuld be an intent lock or something, as it means you can't attack hostile erts etc.

Lumipharon (changes requested):

code/datums/components/harvester.dm
Xander3359: The stamina heal is attached to the lesser heal.
The do_after is only for the greater heal.

code/datums/components/harvester.dm
Xander3359: Because it allows you to get the 25 damage heal you would get if you injected your patient with 5u of bicaridine.
I thought it was quite fitting that medics would heal more with the heal vali

code/datums/components/harvester.dm
Lumipharon: yes, so you can just interrupt the do_after for instant stam heal? fucking vali man

Xander3359: Yes

code/datums/components/harvester.dm
```diff
@@ -250,7 +250,7 @@
 
 ///Handles behavior when attacking a mob with bicaridine
 /datum/component/harvester/proc/attack_bicaridine(datum/source, mob/living/target, mob/living/user, obj/item/weapon)
-	if(isxeno(target)) //Self-heal on xeno strike
+	if(target.faction != user.faction) //Self-heal on attacking an enemy
```
Lumipharon, line 253: IFF sord is lame.

Xander3359: What would be the preferred fix? Purely just an intent check?

Lumipharon: yes

Lumipharon:
I still think giving them all the damage bonus is shitty, but still an improvement

PR #14408, opened 2023-11-08
remove miniscope on recoilless rocket

novaepee:
## About The Pull Request

title

## Why It's Good For The Game

This might be harder to shallow compared to removing miniscope on RR, though it had been argued that RR is stronger than SADAR and the fact that an offscreen weapon with high burst damage is not okay on the receiving end of said weapon (xenomorph players).

Would like to see maintainer's thoughts on this. If they say no to this, then this PR is a precedence for having miniscope on RR.

## Changelog

:cl:
del: remove miniscope on recoilless rocket
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -787,7 +787,6 @@ Note that this means that snipers will have a slowdown of 3, due to the scope
 	general_codex_key = "explosive weapons"
 	attachable_allowed = list(
 		/obj/item/attachable/magnetic_harness,
-		/obj/item/attachable/scope/mini,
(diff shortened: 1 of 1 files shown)


Barnet2:
I thought we learned with SADAR why a scope on a rocket launcher is a bad thing. While RR is weaker than SADAR that doesn't mean it needs a compensation for it, considering it (can) be free and has a lower price.

Strayed-Raven:
All this will do is encourage more blind fire.

ivanmixo (approved):
Inevitable tbh

zzurge:
Rip in peace a real one

PR #14412, opened 2023-11-08
Removes Laser snpier, rolls incendiary shot into lasrifle and shatter into carbine.

Pariah919:
## About The Pull Request
Removes lassniper and rolls its shatter and heat firemodes into lasrifle and carbine respectively.

LasRifle Heat has had its damage reduced to 30.
LasCarbine shatter had its damage reduced to 30 and pen to 10.
## Why It's Good For The Game
Laser sniper is forever stuck in this limbo of being unable to be good due to being a sniper and being hitscan and thus being aim modeless which leads it to barely being used on top of minimum ranges if we existed in a TGMC with no aim mode it would likely just make the 127 irrelevant by how powerful hit scan is, but if you want to long-range lasgun the lasrifle overcharge mode is just... there.

Its gimmick shots are better held by something else that can make better use of them, ricochet shot has nowhere to go so I got rid of it, someone can reintroduce it if they want.
## Changelog
:cl:
del: Marine laser sniper.
balance: As marine laser sniper is gone, some of its fire modes are incorporated into other guns. Specifically incendiary shot for the LasRifle and shatter for LasCarbine
/:cl:


Barnet2:
Agree, this really feels like it doesn't need to be in the game. A sniper without aimmode defeats the purpose in TGMC.

yyzsong:
We're just going back to the old multipurpose laser rifle

GrayRachnid:
I dislike this PR as one of 3 las sniper users. There is nothing wrong with the las sniper. It is usable and I have at least 6 videos in clips in the last 2 weeks of me securing kills with it.

Lumipharon:
If you want to add more stuff to any laser gun, go for it. But removing it because you don't use it is lame. Loads of our guns are statistically not optimal either.

PR #14414, opened 2023-11-09
Laser rifle can't use a full scope anymore

Lumipharon:
## About The Pull Request
What it says on the tin.
Big scope makes the laser rifle step on the toes of the laser sniper severely, much more so than projectile weapons.
## Why It's Good For The Game
More clear cut roles
## Changelog
:cl:
balance: Laser rifles can no longer equip full scopes
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -448,7 +448,6 @@
 		/obj/item/attachable/lasersight,
 		/obj/item/attachable/flashlight,
 		/obj/item/attachable/magnetic_harness,
-		/obj/item/attachable/scope/marine,
 		/obj/item/attachable/scope/mini,
 		/obj/item/weapon/gun/flamer/mini_flamer,
 		/obj/item/attachable/motiondetector,


Tailiik:
Guns that shoot half the range of the t47 scope have been allowed to use it for eons, epochs even. My homie laser rifle is innocent!

Tommy-211:
laser sniping is not real don't act like it is

DaTruRiZZoB:
closes #14412, immediately makes pr to remove scope on laser rifle. like pottery. instead of removing the scope from the rifle why not try to make the laser sniper a more viable pick?

Joe13413:
gonna go full ideasguy here, but maybe adding a dedicated sniper scope, and then reducing the sight range of the full scope would be more in order? i have no fucking idea how this'd pan out, it's just some half baked idea i've given no thought to

ivanmixo:
This is really silly lumi

Lumipharon:
> This is really silly lumi

It's the same logic as removing it off t27 and t60, don't see the difference. Non snipers shouldn't be better at sniping then actual snipers

technicianman:
then why not make the laser sniper better at sniping

ivanmixo:
> > This is really silly lumi
> 
> It's the same logic as removing it off t27 and t60, don't see the difference. Non snipers shouldn't be better at sniping then actual snipers

It's more akin to removing it off the T12 if anything. I don't like how this was basically made as a response to "laser sniper sucks and the rifle is better". Maybe the sniper being bad at its job is the issue?

Lumipharon:
T12 isn't good at long range though. I could also give t60/mg27 scopes back and just buff all snipers on the same logic, but I don't think that'd be a good idea.

Issue #13039, opened 2023-05-21
Req beacons on tad misbehave


chizzy376:
### Description

With a beacon placed on tad that's groundside, sending orders there works as normal, but then when the tad goes to the sky and lands elsewhere, sending things to the beacon sends things to the OLD location, not the new one 

### Test Merges

Server-Side Code Modifications: https://tgstation13.org/parsed-logs/sybil/server_side_modifications.dm

Round ID: [21246](https://scrubby.melonmesa.com/round/21246)
BYOND Version: 515.1607
Server revision compiled on: 
Master commit: [037175690c375449937d061d078f3eb9e7606d09](https://github.com/tgstation/TerraGov-Marine-Corps//commit/037175690c375449937d061d078f3eb9e7606d09)
The following pull requests are currently test merged:
[#12892: 'Adds a delay to throwing' by ivanmixo at commit a99898c494](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12892)
[#12990: 'Autocannon rework MK2' by Pariah919 at commit 75d34f1c10](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12990)
[#13004: 'Text to speech' by TiviPlus at commit 03d2a45c79](https://github.com/tgstation/TerraGov-Marine-Corps//pull/13004)
[#13035: 'Fix some pref bugs' by Lumipharon at commit 2aea1551af](https://github.com/tgstation/TerraGov-Marine-Corps//pull/13035)
[#13027: 'Maybe fix radio keys being filtered' by TiviPlus at commit a959345250](https://github.com/tgstation/TerraGov-Marine-Corps//pull/13027)
[#12977: 'Reworks boiler into having to be rooted to fire' by ivanmixo at commit ce93326fe3](https://github.com/tgstation/TerraGov-Marine-Corps//pull/12977)

Local commit: 1aca0dafb7d05c099bb673d6c2a855abbda7d7af
Server tools version: 5.12.1

### Reproduction Steps

1. Go groundside with tad, land anywhere
2. plant beacon on tad that's landed
3. RO sends things to beacon, works as expected
4. Tad moves into the sky, lands elsewhere on ground
5. Sending things to beacon now makes the beacon send to the old location

PR #14421, opened 2023-11-10
Filter Supply Beacon Drop Locations if its in the Air

SiegeB0t:
## About The Pull Request

The req launch console now does not allow sending crates to supply beacons that are not on the ground. Thus when tad launches with beacons onboard, an oblivious RO will not deliver B18's to the hive. Also when tad lands, crates will be sent to the right place.
## Why It's Good For The Game
RO does not know when tad launches and has to ask the pilot if they moved every time a crate needs to be sent. This is silly.

Unlike antennas, beacons are static objects and should update their drop locations when moved.

Fixes issue #13039.
## Changelog
:cl:
qol: req can be confident that crates to a supply beacon on tad will arrive on tad.
fix: A supply beacon on tad will work after tad moves.
/:cl:

### DIFF
--- a/code/game/objects/machinery/squad_supply/supply_beacon.dm
+++ b/code/game/objects/machinery/squad_supply/supply_beacon.dm
@@ -114,6 +114,11 @@
 	SIGNAL_HANDLER
 	beacon_datum = null
 
+/obj/item/beacon/supply_beacon/onTransitZ(old_z,new_z)
+	. = ..()
+	//Assumes doMove sets loc before onTransitZ
+	beacon_datum.drop_location = loc
+
 /obj/item/beacon/supply_beacon/activate(mob/living/carbon/human/H)
 	var/area/A = get_area(H)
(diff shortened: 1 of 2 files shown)


code/game/objects/machinery/squad_supply/supply_beacon.dm
```diff
@@ -114,6 +114,11 @@
 	SIGNAL_HANDLER
 	beacon_datum = null
 
+/obj/item/beacon/supply_beacon/onTransitZ(old_z,new_z)
+	. = ..()
+	//Assumes doMove sets loc before onTransitZ
```
psykzz, line 119: It does 
https://github.com/SiegeB0t/TerraGov-Marine-Corps/blob/d92507c2d393a6dcb0752ecad22754adce2a1867/code/game/atoms_movable.dm#L427

psykzz (approved):

PR #14430, opened 2023-11-12
Attempts to fix an issue with larva queue

lKiseki:
## About The Pull Request
So during some rounds, people realized some weird behavior with the larva queue. Looking into runtimes, @Barnet2 saw this:
image

I've tried to correct this as seen in files changed. I don't know if this is the correct solution, and I'd appreciate it if @ZeWaka could look at this change to make sure it's okay. Otherwise you may slap me in the face.

## Why It's Good For The Game
Fixes a bug with larva stuff.

## Changelog
:cl: Lewdcifer
fix: Fixed some weird behavior in the larva queue.
/:cl:

### DIFF
--- a/code/datums/gamemodes/infestation.dm
+++ b/code/datums/gamemodes/infestation.dm
@@ -295,9 +295,9 @@
 /datum/game_mode/infestation/announce()
 	to_chat(world, span_round_header("The current map is - [SSmapping.configs[GROUND_MAP].map_name]!"))
 
-/datum/game_mode/infestation/attempt_to_join_as_larva(mob/dead/observer/observer)
+/datum/game_mode/infestation/attempt_to_join_as_larva(client/waiter)
 	var/datum/hive_status/normal/HS = GLOB.hive_datums[XENO_HIVE_NORMAL]
-	return HS.add_to_larva_candidate_queue(observer)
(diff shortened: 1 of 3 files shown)


vvvv-vvvv (changes requested):
what you want to do is make this proc take clients instead https://github.com/tgstation/TerraGov-Marine-Corps/blob/2b2872bef868d57dfb065bab40d0bcee16db855a/code/datums/gamemodes/infestation.dm#L298-L300

Lumipharon:
you're only changing the child proc. Either change the args on both, or just make the proc itself pull client from the mob arg.

lKiseki:
> you're only changing the child proc. Either change the args on both, or just make the proc itself pull client from the mob arg.

That's because the other proc already had `client` in it.
https://github.com/tgstation/TerraGov-Marine-Corps/blob/2b2872bef868d57dfb065bab40d0bcee16db855a/code/modules/mob/living/carbon/xenomorph/hive_datum.dm#L1086-L1104

Lumipharon:
gaslighting myself, I even looked at it first. aylmao.

ZeWaka (approved):
thanks - didn't test this usecase i guess, my bad

PR #14436, opened 2023-11-13
this language purports to only use floats but then doesn't (spiderling runtime fix)

ZeWaka:
## About The Pull Request

so... like....
`rand()`
it's a proc right?
 
to give you random numbers?
between two numbers?
 
 
what are numbers in byond?
floats, right?
so, rand() should work with floats, right?
 
 
right?
 
...right?

.........RIGHT?
 
**no.**
 
**_it doesn't._**
 
 
 
 
contribute to opendream today

 

 

 

code ported from goonstation ([pali](https://github.com/pali6)), not really copyrightable but they gave permission

## Why It's Good For The Game

byond just never fails to disappoint doesn't it
fixes spiderlings being shit and runtiming
admins begging yet no coders fix
sad

## Changelog
:cl:
fix: Spiderlings and some assorted damage rolls should now actually work instead of being super weird
/:cl:

### DIFF
--- a/code/__HELPERS/math.dm
+++ b/code/__HELPERS/math.dm
@@ -125,5 +125,5 @@
 	return TRUE
 
 
-/// rand() but for floats, returns a random floating point number between L and H
-#define randfloat(L, H) ((L) + rand() * ((H) - (L)))
+/// rand() but for floats, returns a random floating point number between low and high
+#define randfloat(low, high) ((low) + rand() * ((high) - (low)))
--- a/code/__HELPERS/math.dm
+++ b/code/__HELPERS/math.dm
@@ -125,4 +125,5 @@
 	return TRUE
 
 
-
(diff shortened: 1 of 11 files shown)


code/__HELPERS/math.dm
```diff
@@ -125,4 +125,5 @@
 	return TRUE
 
 
-
+/// rand() but for floats, returns a random floating point number between L and H
+#define randfloat(L, H) ((L) + rand() * ((H) - (L)))
```
vvvv-vvvv, line 129:
```suggestion
/// rand() but for floats, returns a random floating point number between low and high
#define randfloat(low, high) ((low) + rand() * ((high) - (low)))
```

ZeWaka: sure, just following the byond documentation ;)

https://www.byond.com/docs/ref/#/proc/rand

psykzz (approved):

PR #14441, opened 2023-11-13
emissive droppods

Lumipharon:
## About The Pull Request
Emissives everywhere.

image
## Why It's Good For The Game
looks nice
## Changelog
:cl:
imageadd: drop pods are now emissive
/:cl:


psykzz:
Is it not possible to have emissive without light?

Kuro020 (approved):

Lumipharon:
> Is it not possible to have emissive without generating light?

You can use emissives without light, but if you don't have enough light to see the object, you can't see the emissive.
So some light is required, but I stylistically like it as well (compared to cigarettes for example where I have it as weak as possible)

ivanmixo (approved):

PR #14442, opened 2023-11-13
Add area specific sound enviroments and adds reverb to caves

Dikozda:
## About The Pull Request

saw some of this when digging through TG code, makes it so areas can have their own sound environments, like caves and hangers having reverb.

also made it so caves on some maps use the cave sound environment, has some reverb.

## Why It's Good For The Game

i think its cool to walk into an area and for it to sound different, caves having some reverb and what not.

## Changelog

:cl:
soundadd: caves now have reverb
/:cl:

### DIFF
--- a/code/game/area/area.dm
+++ b/code/game/area/area.dm
@@ -60,6 +60,9 @@
 	///Used to decide what the maximum time between ambience is
 	var/max_ambience_cooldown = 120 SECONDS
 
+	///What sound eviroment to use
+	var/sound_environment = SOUND_ENVIRONMENT_ROOM
+
 	///Boolean to limit the areas (subtypes included) that atoms in this area can smooth with. Used for shuttles.
 	var/area_limited_icon_smoothing = FALSE
 
--- a/code/game/area/area.dm
+++ b/code/game/area/area.dm
@@ -61,7 +61,7 @@
 	var/max_ambience_cooldown = 120 SECONDS
 
 	///What sound environment to use
-	var/sound_environment = SOUND_ENVIRONMENT_ROOM
+	var/sound_environment = SOUND_ENVIRONMENT_NONE
 
 	///Boolean to limit the areas (subtypes included) that atoms in this area can smooth with. Used for shuttles.
 	var/area_limited_icon_smoothing = FALSE
--- a/code/game/area/area.dm
+++ b/code/game/area/area.dm
@@ -61,7 +61,7 @@
 	var/max_ambience_cooldown = 120 SECONDS
 
 	///What sound environment to use
-	var/sound_environment = SOUND_ENVIRONMENT_NONE
+	var/sound_environment = SOUND_ENVIRONMENT_ROOM
 
 	///Boolean to limit the areas (subtypes included) that atoms in this area can smooth with. Used for shuttles.
 	var/area_limited_icon_smoothing = FALSE
--- a/code/game/area/area.dm
+++ b/code/game/area/area.dm
@@ -60,8 +60,8 @@
 	///Used to decide what the maximum time between ambience is
 	var/max_ambience_cooldown = 120 SECONDS
 
-	///What sound eviroment to use
-	var/sound_environment = SOUND_ENVIRONMENT_ROOM
+	///What sound environment to use
+	var/sound_environment = SOUND_ENVIRONMENT_NONE
 
 	///Boolean to limit the areas (subtypes included) that atoms in this area can smooth with. Used for shuttles.
 	var/area_limited_icon_smoothing = FALSE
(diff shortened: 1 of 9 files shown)


code/game/area/area.dm
```diff
@@ -60,6 +60,9 @@
 	///Used to decide what the maximum time between ambience is
 	var/max_ambience_cooldown = 120 SECONDS
 
+	///What sound eviroment to use
+	var/sound_environment = SOUND_ENVIRONMENT_ROOM
```
vvvv-vvvv, line 64: ROOM as the default is bad because it has a reverb
```suggestion
	///What sound environment to use
	var/sound_environment = SOUND_ENVIRONMENT_NONE
```

vvvv-vvvv (changes requested):

code/game/area/area.dm
Dikozda: ah ok

Lumipharon:
Aren't sound environments like, really wack? My understanding is when a client is sent a sound environment ALL sounds they hear are effected by that environment until a new environment is sent to them.

Dikozda:
> Aren't sound environments like, really wack? My understanding is when a client is sent a sound environment ALL sounds they hear are effected by that environment until a new environment is sent to them.

not sure, but i think having the base area set to no sound environment should mean it would always go back to normal when they leave an area like caves

Dikozda:
[2023-11-13 22-06-40.webm](https://github.com/tgstation/TerraGov-Marine-Corps/assets/12964904/962f4330-4fc4-4489-be8a-7de99aff4e59)

Dikozda:
also i know its kinda hard to hear the difference in the vid, might try and make another later that is more clear.

Dikozda:
turns out, SOUND_ENVIRONMENT_NONE has more reverb then SOUND_ENVIRONMENT_ROOM so i am going to change that back

vvvv-vvvv:
> SOUND_ENVIRONMENT_NONE has more reverb then SOUND_ENVIRONMENT_ROOM

my bad you're right https://gist.github.com/vvvv-vvvv/ec84d69ae3b1d43f9f6700069e7ae882

vvvv-vvvv (approved):

code/game/sound.dm
```diff
@@ -106,7 +106,12 @@ A good representation is: 'byond applies a volume reduction to the sound every X
 		S.falloff = falloff ? falloff : FALLOFF_SOUNDS * max(round(S.volume * 0.05), 1)
 
 	if(!is_global)
-		S.environment = SOUND_ENVIRONMENT_ROOM
+		//get area sound enviroment and use it for sound
+		var/area/A = get_area(src)
+		S.environment = A.sound_environment
+		//need room and roomHF to be not -10000 for reverb to work
+		S.echo[3] = 0
+		S.echo[4] = 0
```
vvvv-vvvv, line 114: doubting myself one more time this is what tg does https://github.com/tgstation/tgstation/blob/186c9d8f7e4c9e7e1b7e8dd3dd778b9d232a9eb4/code/game/sound.dm#L167-L178

BYOND's SOUND_ENVIRONMENT_NONE does this, so that's why it had more reverb:
> The default value (-1) specifies no change in environment.

To get around this what tg does is keep the last environment used, but turn Room and RoomHF all the way down so nothing applies. So you would want this:
```suggestion
		if(S.environment != SOUND_ENVIRONMENT_NONE)
			//need room and roomHF to be not -10000 for reverb to work
			S.echo[3] = 0
			S.echo[4] = 0
```

code/game/area/area.dm
```diff
@@ -60,6 +60,9 @@
 	///Used to decide what the maximum time between ambience is
 	var/max_ambience_cooldown = 120 SECONDS
 
+	///What sound environment to use
+	var/sound_environment = SOUND_ENVIRONMENT_ROOM
```
vvvv-vvvv, line 64: and this too
```suggestion
	///What sound environment to use
	var/sound_environment = SOUND_ENVIRONMENT_NONE
```

vvvv-vvvv (changes requested):

code/game/area/area.dm
Dikozda: seems good, thanks

Lumipharon:
give it a try

Dikozda:
might need to tune down the reverb a bit

Dikozda:
so a lot of things seem to sound different between testing local and hearing it on the server, changed it so there should hopefully be a lot less reverb, might be a bit too much reduction but can't be sure till i hear it live

Lumipharon:
this seems to have some issues in some cases.
Either some areas have wack sound environments, or I'm picking it up when I shouldn't

Dikozda:
yeah, not sure what is causing it, have not had any luck recreating it on a local server, so it might take me a bit find out whats causing it

Dikozda:
currently finding myself short on time and have had no luck fixing some of the issues with this, think i will close this for now and give it another go when i have more time

PR #14456, opened 2023-11-15
Valk buff

Xander3359:
## About The Pull Request
Adds a check for stabilized limbs when adding pain to a human.
## Why It's Good For The Game
Bugfix
## Changelog
:cl:
balance: Stabilized limbs no longer give un-splinted amounts of pain
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -79,7 +79,7 @@
 			if(((O.limb_status & LIMB_DESTROYED) && !(O.limb_status & LIMB_AMPUTATED)) || O.limb_status & LIMB_NECROTIZED)
 				traumatic_shock += 40
 			else if(O.limb_status & LIMB_BROKEN || O.surgery_open_stage)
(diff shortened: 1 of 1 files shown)


lKiseki:
The one drawback I can find as a xeno against valkyrie is precisely stacking fractures and the pain they give, because otherwise there is no reward for excessive damage on a valkyrie user. This is most certainly an unnecessary buff to the module and I'd rather it stayed as-is.

Lumipharon:
its not a bug fix, its a deliberate thing, and a fairly reasonable/minor tradeoff for not needing splints imo.

Xander3359:
> its not a bug fix, its a deliberate thing, and a fairly reasonable/minor tradeoff for not needing splints imo.

Everything else checks for stabilized limbs except for this here, that's what makes me think this is something that was forgotten.
But I'll go ahead and change the label to balance

novaepee:
I would be genuinely surprised if this gets merged.

ivanmixo:
No thanks

PR #14464, opened 2023-11-16
Heavy laser has additional firemodes

Pariah919:
## About The Pull Request
It also has ammo like a rel lasgun, which is weird it didn't considering it was always one? I guess it means you can recharge it, maybe.
It has burst fire shatter rounds which are just laser sniper ammo with no min range, and coming in 3 round bursts (and 1 sunder instead of 5.)
It has the ricochet shot, which is unchanged compared to laser sniper, it has slightly faster fire rate though.
Also adds clearer unloaded and loaded sprites for the deployed variant in your hand.
## Why It's Good For The Game
Makes the heavy laser gun use the laser gun gimmick of versatility in fire mode selection, rather than being this weird masquarding flamethrower thing, specifically has ricochet shot and shatter shot because it makes the most sense for a "support lasgun", even without IFF. Your main tool of fire is still probably gonna be glob, though.
## Changelog
:cl:
add: Added new Heavy Laser fire modes.
/:cl:


Lumipharon:
Autofire ricochet gooo

Lumipharon (approved):

PR #14468, opened 2023-11-17
Max health code improvements

Lumipharon:
## About The Pull Request
Max health (for humans) is no longer hardcoded to species maxhealth. Added support for modifiers to maxhealth (i.e. a hat or drug could change your max health), but no player facing changes currently.

Damage slowdown is now based on your current max health (so again no changes for players but it means it scales with your maxhp if that ever changes), along with a couple of other random misc things.
## Why It's Good For The Game
Better/less hardcoded mechanics, prep for later features.
## Changelog
:cl:
code: Removed some hardcoded limitations around max health
/:cl:


code/modules/clothing/clothing.dm
```diff
@@ -22,7 +23,8 @@
 
 	/// Used by headgear mostly to affect accuracy
 	var/accuracy_mod = 0
-	flags_inventory = NOQUICKEQUIP
+	///Modifies the wearer's max health
+	var/health_mod = 0
```
QualityVan, line 27: Putting this on clothing now and unused seems odd, easy enough to add it with any relevant gear.

QualityVan (changes requested):
You'll want to adjust the medscanner interface too, along with possibly sleepers. Cryo already handles it properly.

code/modules/clothing/clothing.dm
Lumipharon: oh yeah, forgot to remove this for the pr, will do in a bit.

QualityVan (approved):

PR #14479, opened 2023-11-18
Xenos that die on crash refund 1 larva

Xander3359:
## About The Pull Request
Xenos that die on crash refund 1 larva
## Why It's Good For The Game
Currently Crash suffers from low xeno spawn rate.
Already they are locked behind the 3 minute xeno respawn time, so sitting in larva queue because balance_scales() decided that you don't deserve any larva is not a good experience.
Crash should have either 0 xeno respawn timer, or they should have some sensical larva generation so people aren't permanently stuck in larva queue.
## Changelog
:cl:
balance: Xenos get 1 larva refunded when they die on crash
/:cl:

### DIFF
--- a/code/__DEFINES/dcs/signals.dm
+++ b/code/__DEFINES/dcs/signals.dm
@@ -23,6 +23,7 @@
 #define COMSIG_GLOB_NUKE_EXPLODED "!nuke_exploded"
 #define COMSIG_GLOB_NUKE_DIFFUSED "!nuke_diffused"
 #define COMSIG_GLOB_DISK_GENERATED "!disk_produced"
+#define COMSIG_GLOBAL_XENO_DEATH "!xeno_death"
 
 #define COMSIG_GLOB_SHIP_SELF_DESTRUCT_ACTIVATED "!ship_self_destruct_activated"
 
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -95,6 +95,7 @@
 		computer_to_disable.update_icon()
 
 	SEND_GLOBAL_SIGNAL(COMSIG_GLOB_OPEN_TIMED_SHUTTERS_CRASH)
+	RegisterSignal(SSdcs, COMSIG_GLOBAL_XENO_DEATH, PROC_REF(regain_larva))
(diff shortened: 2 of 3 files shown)


ivanmixo:
Kind of an issue because this means xenos can never lose pop from marines dying, they can only go up

Zangeiti:
this + the maturity removal makes it harder for marines to keep up w xeno numbers

Strayed-Raven:
Crash is bad so no one is gonna care. But I think this is not quite the right way to do it. Unless you want Crash to be a xeno protagonist mode

ZeWaka:
Perhaps 0.3-0.5 larva or something instead?

Xander3359:
> Perhaps 0.3-0.5 larva or something instead?

Me and @psykzz did discuss this a bit in the discord a while back.
We both agree that Crash as a gamemode needs a lot more changes than this, but I personally think that this would be a step in the right direction even disregarding other changes to the mode as a whole.

Tommy-211:
Xenos having infinite respawns?
This surely cant go wrong.

lbnesquik:
Honestly good for the low pop gamemode

DaTruRiZZoB:
if anything this just means that the only thing holding xenos back from having a 100% winrate is playercount. whilst it's good that xenos get to constantly be in the action and play the game, this also means that there's less punishment for dying other than evo. whereas if you were to wear the shoes of a marine you see that there's no point in risk. marines still have to wait an absurd amount of time compared to xenos now when they go perma, and if there's a wipe then there is ZERO reason to even bother fighting and with recent crash rounds marines majority of the time evac when the ship is even lightly pressured.

this doesn't fix GL spam, this doesn't fix how one-sided (either xeno or marine) the gamemode can be, and this doesn't fix the imbalance of respawning in what is supposed to be a lowpop gamemode. 
i suggest that marine respawn (from perma) should be lowered inturn to this change, the supplies (grenades and specialties such as shields and jetpacks) should be lowered to account for the faster respawn for all players, and xenos should have the ability to build special buildings (outside of the fog) or a way to spend generate psy points for miscellaneous upgrades.
crash has too many flaws and only changing 1 thing at a time will result in a constant imbalance until the gamemode gets improved.
or im drunk and this is all silly

Lumipharon:
crash is a pretty cursed mode, but this seems like a terrible idea.

PR #14415, opened 2023-11-09
Deletes Baneling.

Pariah919:
## About The Pull Request
Gets rid of Baneling.
## Why It's Good For The Game
This caste is absolutely awful to play against from a design perspective you cannot play against it in almost any form and if you have beat it you've likely already won or they placed their pod in an absolutely awful spot., if you care about my full 2 paragraph rant look below.

<details>

## The Bane of Lings

To preface why I'm removing Baneling from a design perspective, the caste breaks almost all precedence when it comes to gas, all the gas castes (save sent primo) have to highly risk themselves inorder to gas. Boiler has to root, Defiler has to fart and run up to marines.

Baneling? Baneling can just run in. If you die, you waste 2 minutes. Just run back! It doesn't matter that you die considering most rest cycles take equally if not longer under bad/decent conditions, and your pod is likely in the middle of the hive anyway!

This all leads to baneling being pretty much unbeatable from a marine perspective, if you shoot it it explodes, it comes back, it blows up again. If you shoot it far away it just comes back, you shoot it, it comes back again. It can be stacked infinitely and there's nothing you can do about it!

## Why

Why should it be removed? Why not leave it in for someone to rework it? Now we get into subjectiveness.

You can say this, but I disagree on the principle that other castes arguably in this limbo-like Bull and old!Sent aren't quite as game health breaking, primo bull was utterly busted but once it was fixed the caste faded back into the "waiting for rework" bin, but you can't do the same thing with baneling unless you make the caste literally unplayable like I am because the entire design is rotten from the very core due to being incredibly low risk in a game about risk/reward management due to its inherient nature as a PvP game.
</details>

## Changelog
:cl:
del: Removed the Baneling xeno caste.
/:cl:

PR #14484, opened 2023-11-19
Baneling death rework (only on intentional death)

ZeWaka:
## About The Pull Request

People keep screaming about how baneling is shit and needs to be outright removed, see: #14415 
More context can probably be found on the discord.

image

I didn't make the change where it no longer releases gas on permadeath, since that seems at least a good way to go out.
Marines still accomplish killing a xeno for good. Gas radius is still smaller, ofc.

uhh i tested it locally and it seemed to work lol
did a small bugfix and cleanup while i was here (it was showing the 'cant explode' message every few secs)

## Why It's Good For The Game

I honestly would rather banelings be nerfed less than this PR, but if this is the alternative to them getting removed entirely then I can stomach it.

## Changelog

unsure how to mark this more majorly lol
:cl:
balance: Banelings no longer respawn in their pod if they are killed, except for if they use their explode abilities.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/baneling/abilities_baneling.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/baneling/abilities_baneling.dm
@@ -16,6 +16,7 @@
 	keybinding_signals = list(
 		KEYBINDING_NORMAL = COMSIG_XENOABILITY_BANELING_EXPLODE,
 	)
+	var/died_due_to_ability = FALSE
 
 /datum/action/xeno_action/baneling_explode/handle_button_status_visuals()
 	var/mob/living/carbon/xenomorph/baneling = owner
@@ -72,6 +73,9 @@
 	var/smoke_range = X.plasma_stored/60
 	/// Use up all plasma so that we dont smoke twice because we die.
 	X.use_plasma(X.plasma_stored)
+	// If the smoke was triggered via ability, flag that
+	if (ability)
+		died_due_to_ability = TRUE
 	/// If this proc is triggered by signal, we want to divide range by 4
 	if(!ability)
 		smoke_range = smoke_range*0.25
--- a/code/modules/mob/living/carbon/xenomorph/castes/baneling/abilities_baneling.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/baneling/abilities_baneling.dm
@@ -48,18 +48,26 @@
 
 /datum/action/xeno_action/baneling_explode/can_use_action()
 	. = ..()
-	var/mob/living/carbon/xenomorph/X = owner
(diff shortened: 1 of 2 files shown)


zee11zee:
Ummm I dont think so, maybe on death no gas? Or ima blow up everyone of these guys i see for -1 larvae.

ZeWaka:
> Ummm I dont think so, maybe on death no gas? Or ima blow up everyone of these guys i see for -1 larvae.

make your own PR, I explained my justification

AlexCrimson:
As long as manual exploding is instant i do not see how this nerfs them much. Maybe they will need to be a bit more careful, but overall its still very easy to roll in an explode quickly.

This also does not address the second "issue" with the caste. That being the fact its boring to sit and wait for respawn.

In my irrelevant opinion, the caste should be removed rather than mope nerfed into oblivion because Marine players hate playing against the constantly respawning enemy, for once.

ZeWaka:
> As long as manual exploding is instant i do not see how this nerfs them much. Maybe they will need to be a bit more careful, but overall its still very easy to roll in an explode quickly.

You can't explode if you're staggered. Unags just need to learn how to play against the caste at that point.

Carnage068:
ayyyy i actually had an impact on something for once! I like the change, it probably won't 'fix' Baneling outright, but it would make it a lot more tolerable for Marines and make Baneling more risky to actually play. It's a good change imo, but then again I'm the one who suggested it lol

Carnage068:
> This also does not address the second "issue" with the caste. That being the fact its boring to sit and wait for respawn.


How is it boring for marines to wait for Banelings to respawn? If you're trying to say it's boring for xenos, then nobody would play it, and nobody would care about it being removed. I don't think it's fair to dictate what is and isn't fun for other people and use that to support the idea of it being taken out of the game. The people who don't find it fun can just play other castes. That's kind of like saying me and various other people don't find Spitter fun because of how often you need to wait for more plasma, so it should be removed.

AlexCrimson:
The fact its boring for the xeno player has been mentioned multiple times since its creation. It was also mentioned in the deletion PR. Right now, attempts are being made to nerf the caste, but none to fix the issues on the xeno-side. This is just going to lead to the caste barely being played, which is why i said its cleaner to just remove it, rather than turn it into something even more unfun to play.

And that spitter comparison is dumb as fuck, because Spitters regen plasma very quickly, so its nothing like waiting 120 seconds to respawn.

SilvernKell:
Just fucking remove it at this point. This is utterly horrible of an idea, 1 second of lag whilst being hit by bullets (which constantly causes lag) and you're dead while spamming the explode button. Or a shotgun stagger and you're dead. 
100% worthless caste on this that will only become a joke and cause frustration, this is absolutely no better then a removal and is now a wasted larva.

PR #14488, opened 2023-11-19
Do after fix

Lumipharon:
## About The Pull Request
A bunch of do_after shit. Probably needs a quick tm before merge, but should be merged relatively quickly. There'll be no notification if people add do_afters using the old arg format which could lead to weird behavior.

1. do_after now uses flags like tg instead of a hand and turf_move var.
2. Fixed a bug with ignoring movement not working if your target is yourself.
3. Purged do_mob. As far as I can tell we just ported the same thing twice, at different times in different ways. Absolute state.
4. Updated every instance of do_after and do_mob to use the correct args.

## Why It's Good For The Game
Fixes good, duplicate code bad, flags good

## Changelog

:cl:
code: updated do_after and removed do_mob
/:cl:


lKiseki (approved):
image
Checked all files, didn't find anything wrong.

PR #14494, opened 2023-11-20
Pill bottles for all

Lumipharon:
## About The Pull Request
The maintainers are furiously throwing dice at each other and we rolled double six's. According to the wheel that means you get pill bottles back.

Replaced packets entirely with pill bottles for marine use.
## Why It's Good For The Game
Unga gets to gobble more pills.
## Changelog
:cl:
balance: Replaced packets with pill bottles for all marine use
/:cl:

### DIFF
--- a/code/game/objects/items/storage/pill_packets.dm
+++ b/code/game/objects/items/storage/pill_packets.dm
@@ -65,7 +65,7 @@
 	pip_color = COLOR_PACKET_DYLOVENE
 
 /obj/item/storage/pill_bottle/packet/paracetamol
-	name = "paracematol pill packet"
+	name = "paracetamol pill packet"
 	desc = "This packet contains paracetamol pills, also known as tylenol. A long lasting but minor painkiller. Once you take them out they don't go back in. No more than 4 pills in a long period."
 	pill_type_to_fill = /obj/item/reagent_containers/pill/paracetamol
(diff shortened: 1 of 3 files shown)


Pariah919:
The great bottling

Xander3359 (approved):
death to paracematol

code/game/objects/items/storage/pill_packets.dm
```diff
@@ -65,7 +65,7 @@
 	pip_color = COLOR_PACKET_DYLOVENE
 
 /obj/item/storage/pill_bottle/packet/paracetamol
-	name = "paracematol pill packet"
+	name = "paracetamol pill packet"
```
spookydonut, line 68: unatomic change

Oscari-pog:
I know this seems like gimping self but could you leave packets existing

Like put pill bottles on top if you want to, just make packets still available

Al-1ce:
> I know this seems like gimping self but could you leave packets existing


..Why? The only issue I can see is this might bug up some load-outs, but otherwise the packets are completely moot when compared to any of the pill bottles.

InterroLouis:
Can we keep pill packets around for those marines who want to have quick access to pills instead of a 2.5 second delay for every pill they want to pull?

InterroLouis:
Never mind I just missed the fumble revert.

ivanmixo (approved):

PR #14515, opened 2023-11-22
Halves the cost of HEDP grenade boxes in Req.

chizzy376:
## About The Pull Request
The current price of a box of HEDP grenades which contains 25 grenades has a supply point cost of 500.
For that same cost, you can buy a brand new mortar and 25 HE shells without reqtorio.
If you were to reqtorio shells outright, you'd have to purchase 100 points worth of machinery and then you could purchase 3 resupplies of HE (For the cost of 360) for 90 HE shells, more than triple the amount with points to spare.
What I'm trying to say is, that the cost of 500 is wayy too big for what it gives. Currently you can also just buy 15 WP nades for 700 which are miles better than HEDP.
## Why It's Good For The Game
As good as grenade launcher meta is, convincing Req to buy you an extremely expensive box of only 25 nades (Which go by REALLY fast) is not only wasteful of money, but also extremely hard. This should make the purchase more desirable and less of a drain on req wallet.

### TLDR: Make nades cheaper because the price is too high

(All of this is said by a Req main, I don't really know about the per-point-cost vs Performance of each grenade, but it's definitely more worth it to buy modules or WP nades rather than 25 nades for **20 points each**)
## Changelog
:cl:
balance: halved the cost of HEDP boxes in req. (500 >> 250)
/:cl:

### DIFF
--- a/code/modules/reqs/supplypacks.dm
+++ b/code/modules/reqs/supplypacks.dm
@@ -822,7 +822,7 @@ EXPLOSIVES
 	name = "M40 HEDP high explosive grenade box crate"
 	notes = "Contains 25 grenades"
 	contains = list(/obj/item/storage/box/visual/grenade/frag)
-	cost = 500
+	cost = 250
 
 /datum/supply_packs/explosives/explosives_hidp
 	name = "M40 HIDP incendiary explosive grenade box crate"


chizzy376:
> This is bad reasoning, you get 600 grenades in the vendor.

EvilMelonTheSecond:
> This is bad reasoning, you get 600 grenades in the vendor. Buffing something because other stuff is good is powercreep and bad.
> 
> To have this pass you'd have to lower the amount in the vendor, you can't have it both ways.
I agree that it probably shouldn't be both ways but having a lot of freebies meaning expensive as hell restock price makes the loadout a dice roll to permanently run out of ammo mid round depending on pop and number of people using said loadout. Having RNG for ammo stockpiles will probably push people to take as much as they can so that doesn't happen, making this loadout a first come first serve while also depleting grenades as a utility option.

Which might be justified for really good gear but grenades seem to be the only example of this and I'm assuming by the PR they aren't good enough for marines to want to invest in it, making it a weird version of req hell(considering the amount of stuff in req hell due to shaky balance, it might've been intended).

Lumipharon:
Yeah this is wack reasoning.

PR #14525, opened 2023-11-24
Reduces roundstart grenades from 600->200

Pariah919:
## About The Pull Request
600->200 HEDPs.
## Why It's Good For The Game
600 grenades is enough for the entire game, encourages incredible amounts of spam, reduces logistical need for anyone to actually properly manage grenades RO wise, and generally was just an old overreaction to making grenades much more accessible.
At 200, it should still ideally be highly accessible alongside support grenades, but not quite overbearing to use if the force isn't properly longterm supplied.
## Changelog
:cl:
balance: Amount of free roundstart HEDPs reduced to 200 from 600.
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -110,7 +110,7 @@
 			/obj/item/ammo_magazine/rifle/tx54/incendiary = 4,
 			/obj/item/ammo_magazine/rifle/tx54/smoke = 4,
 			/obj/item/ammo_magazine/rifle/tx54/smoke/tangle = 2,
-			/obj/item/explosive/grenade = 600,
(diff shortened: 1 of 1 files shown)


lbnesquik:
That is a big slash.

lbnesquik:
Well, worth seeing if that mentionned PR gets merged alongside it or not. The one that reduces the price of the boxes in req was refused.

chizzy376:
> This means that ~29 marines can take a full 7 HEDP nades. Which 1) not that many marines do this 2) most marines don’t take exclusively HEDP since there are a significant amount of other varieties that spammers tend to favor first

I don't understand this math. If you're going grenadier, you'd only get 2 marines with full loadout before you run out of HEDP nades.

PR #14527, opened 2023-11-24
Runner can no longer refresh cooldown off evasions.

Pariah919:
## About The Pull Request
Title basically, gets rid of chaining and all the things related to it.

## Why It's Good For The Game
Chaining is bad for the game, it makes a backliner who's main counter is supposed to be traveling around in small groups of 3 to 5 able to dive them with incredible ease, I don't mind extension as much because it makes sense, but being able to clear your C/Ds to escape at no cost as you use it in combat?
That's just plain silly to me and quite bad design, you shouldn't be continually diving into groups as runner, that's a job for other castes, having both it and extension is just plain bad because it allows you to attack a caste that should be weak to marines in small groups.

## Changelog

:cl:
balance: Runner can no longer refresh their evasion cooldown by dodging enough bullets.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
@@ -157,7 +157,7 @@
 	evasion_stacks = max(0, evasion_stacks - proj.damage) // We lose evasion stacks equal to the burn damage.
 	if(evasion_stacks)
 		owner.balloon_alert(owner, "Evasion reduced, damaged")
-		to_chat(owner, span_danger("The searing fire compromises our ability to dodge![RUNNER_EVASION_COOLDOWN_REFRESH_THRESHOLD - evasion_stacks > 0 ? " We must dodge [RUNNER_EVASION_COOLDOWN_REFRESH_THRESHOLD - evasion_stacks] more projectile damage before Evasion's cooldown refreshes." : ""]"))
+		to_chat(owner, span_danger("The searing fire compromises our ability to dodge!"))
 	else // If we have no stacks left, disable Evasion.
 		evasion_deactivate()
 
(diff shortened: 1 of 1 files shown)


yyzsong:
> I don't mind extension as much because it makes sense, but being able to clear your C/Ds to escape at no cost as you use it in combat?

This pr removes both, doesn't it?

Arturlang:
Keep in mind runner evasion has plenty of counters (meele, PB shotgun instakill, fire, tangle) Is this really neccesary? Being able to dance in front of HUGE marine groups is super fun though you can't do that much damage without doing something incredibly risky

lbnesquik:
> Keep in mind runner evasion has plenty of counters (meele, PB shotgun instakill, fire, tangle) Is this really neccesary? Being able to dance in front of HUGE marine groups is super fun though you can't do that much damage without doing something incredibly risky

Melee and PB on the smallest, and generally most agile caste isn't easy. Fire is limited heavily by fuel, and doesn't really hurt xenos. Tangle is even more limited.

Arturlang:
> > Keep in mind runner evasion has plenty of counters (meele, PB shotgun instakill, fire, tangle) Is this really neccesary? Being able to dance in front of HUGE marine groups is super fun though you can't do that much damage without doing something incredibly risky
> 
> Melee and PB on the smallest, and generally most agile caste isn't easy. Fire is limited heavily by fuel, and doesn't really hurt xenos. Tangle is even more limited.

Only limited if you don't have a fuel tank, which you will have shipside, FOB, or on the TAD, TAD can have those welderpacks which might also work. Flamer fuel is essentially unlimited and just straight up cancels runner evasion. Quite many marines tend to have flamers as well

Al-1ce:
> you shouldn't be continually diving into groups as runner, that's a job for other castes

This is a dead relic mentality from CM.

ivanmixo:
I'd prefer runner being unable to slash during evasion over this. It just causes so much trouble.

AlexCrimson:
>You shouldnt be continually diving into groups as a Runner

Why not? Marines are often in groups. You are basically saying Runners should be completely useless unless they encounter someone solo? You can already counter Evasion with point-blank shots, melee, stagger & fire. A runner needs to be in PB range to slash you, so you can always fight back.

I really do not understand why everyone seems to want to nerf Runners. They are near-useless in a siege, have a ton of counters, and one mistake with the caste WILL cost you your life.

PR #14530, opened 2023-11-24
FK-88 economy and ammo buff

Pariah919:
## About The Pull Request
Makes the gun 1000 points.
Ammo is now 50 points for both APFDS and HE.
APFDS
damage 200->300
penetration 70->50
slowdown 0.5->1 
HE
damage 50->100 (matches HE sadar)

## Why It's Good For The Game
FK-88 is insanely cost inefficent, I was too cautious cost wise so making it old sadar cost was a fools errand, 1000 points is a much fairer milestone to aim for, the ammo is also 50 points overall.
The ammo buffs are to make the ammo feel more high impact, you already have a high cost to entry.

## Changelog

:cl:
balance: FK-88 now costs 1000 points from 1200, and ammo is 50 points across the board.
balance: FK-88 APFDS has 50 penetration rather than 70, but 300 damage rather than 200. HE does 100 damage on direct hit rather than 50.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -2128,7 +2128,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	hud_state = "bigshell_he"
 	hud_state_empty = "shell_empty"
 	flags_ammo_behavior = AMMO_ROCKET|AMMO_EXPLOSIVE
-	damage = 50
+	damage = 175
 	penetration = 200
 	max_range = 30
 	shell_speed = 0.75
@@ -2148,8 +2148,8 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	icon_state = "apfds"
 	hud_state = "bigshell_apfds"
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_PASS_THROUGH_TURF|AMMO_PASS_THROUGH_MOVABLE
-	damage = 200
-	penetration = 75
+	damage = 300
+	penetration = 50
 	shell_speed = 7
 	accurate_range = 24
 	accurate_range_min = 6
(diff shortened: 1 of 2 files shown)


Pariah919:
SADAR AP does 340/200 and has way less setup and no min range?

ivanmixo:
I would rather not with this one

SimplySmiley123:
> but like legit tho, 300 damage + 50AP?
> 
> When factoring in armor
> 
> this 1 hits the following castes from full health:
> 
> -Drone -Sentinel -Baneling -Runner -Hunter -Wraith
> 
> and if the caste has 25 damage done to it (like 1-2 shots from any weapons), this now one hits:
> 
> -Drone -Sentinel -Baneling -Runner -Hunter -Wraith -Carrier -Bull -Spitter -Boiler
> 
> youre genuinely saying you want a weapon that 1 taps 10/25 castes? Did you even think about this?
> 
> I created both a [simplified and complex stats chart](https://tgstation13.org/wiki/User:ClosetedSkeleton#Xenomorph_Castes_Basic_Stats) on the wiki for our new players who are interested in seeing how values like xeno health and armor are when compared to eachother.
> 
> I think it would help you a lot and would recommend using it. It is kept up to date

I gotta agree about the scope and range of this, but I don't really fully agree about this selection of xeno castes being an issue if they get killed by it **if** the x8 scope got removed or replaced with something less crazy. 

Drones likely shouldn't be in combat against a stationary emplacement, but supporting other castes meant for breaking through emplacements.
Sentinel maybe? They are tier 1 and are ranged though so, I wouldn't really expect this one to survive much anyway.
Baneling is a suicide xeno caste, if you are already close enough you did your job if you died to it.
Runner is a hit-and-run xeno, wouldn't expect this one to tank any big hitters anyway, they should pick better fights.
Hunter I could see the point, it is a higher tier caste, but at the same time it's still a hit-and-run caste, probably not a xeno that should be worried about dealing with an emplacement but harrassing the squishier and usually not hard-hitting marines at the frontlines.
Wraith is also a xeno that doesn't seem like it's meant to tank anything that hits harder.

Carrier, this one I can see the point, but at the same time it's not a tanky xenomorph meant for tanking from what I seen.
Bull, now this one I can agree on somewhat. You are meant to charge in as one, though at the same time Crusher exists and is just better at this then you are. You're better off harrassing the marines at the frontline.
Spitter, well, same as Sentinel, you're a ranged xeno not a frontliner.
Boiler, I don't know why exacly you would try charging down an emplacement as a Boiler? 

Let tankier xenomorphs tank and break through cades and emplacements, let squishier xenomorphs stay behind them and be support, flank around and generally pick better fights that make more sense for them to pick.

PR #14531, opened 2023-11-24
Chemrette invis fix

TGCMbetcher:
## About The Pull Request

Changes the slots from 25 to 18

## Why It's Good For The Game

Because someone made the chemrette package larger than the cig one but it uses the same sprite changes, when you go over 18 things inside it goes invisible. Theres no real reason that its larger so this pushes it back down.

I havent tested this, beyond failed every time I tried to run it and I dont know what the vscode version of how to host and run the game is, but the change is so minor there is no way its going to break something. Its literally changing slot sizes. Id appreciate if someone told me the updated method of testing things since the warnings that refused to allow it to run were about how I need to use vscode to do it and I have zero idea where to do that. That or open and run it while testing something else.

## Changelog
:cl:
fix: fixed a few things
/:cl:

### DIFF
--- a/code/game/objects/items/storage/fancy.dm
+++ b/code/game/objects/items/storage/fancy.dm
@@ -171,8 +171,8 @@
 	w_class = WEIGHT_CLASS_TINY
 	throwforce = 2
 	flags_equip_slot = ITEM_SLOT_BELT
-	max_storage_space = 25
-	storage_slots = 25
+	max_storage_space = 18
+	storage_slots = 18
 	can_hold = list(
 		/obj/item/clothing/mask/cigarette,
 		/obj/item/tool/lighter,


novaepee:
been two years

Xander3359:
This would be a balance PR, not a fix PR. 

Adjusting the size doesn't solve the problem, it just hides it.

TGCMbetcher:
There are other issues, like how somewhere between #7256 and now the cigs instantly give you their chems but normal cigarettes dont, so normal cigs are actually better to use. No one uses cigs let alone chemrettes so its not really an issue. It looks like neuro no longer wipes the chems in your system from the games Ive played recently so unless someone gives you hyper it has no gameplay difference.

TGCMbetcher:
> This would be a balance PR, not a fix PR.
> 
> Adjusting the size doesn't solve the problem, it just hides it.

I wanted to try to see if adding more cigs in initially would make it so the initial state would be 25 but I cant test it so I didnt want to mess anything up. It says theres an online component needed so they byond testing method no longer works. That I need to go to vscode tester but that says its fine and I dont know how to run the game from vscode. It fixes the bug and isnt perfect but from what I remember it was originally 18 even if I cant find a PR for any changes.

Xander3359:
If you join the TGMC discord we can talk with less delay in between messages.
Normally to compile and run a localhost, you just hit F5 and it should open a local.

The reason why this "bug" exists, is because chemrettes use a hardcoded update_icon_state which adds a ## at the end of the icon_state based on how its contents.
image

The problem is that we only have up to chempacketbox18
image

This is why adding >18 items to the chemrette pack makes the sprite invisible.

TGCMbetcher:
> The problem is that we only have up to chempacketbox18 
I figured that was the issue when I made the bug report, didnt know how to find it and the initial state wanted to make me try adding more things to start with. Sounds like the proper way to do things would be to space those all out until you have 25 different sprites, but it also looks like it was only ever intended to hold 18 so reverting it back to that wouldnt be a bad thing. No idea when it turned to 25, I searched for that but couldnt find it originally. I was just doign a basic fix if someone wants to go into the sprites and change things I can close this, but theres not reason to have 25 if it only starts with 18.
Thanks for the start point for the vscode, turns out I needed to download like 9 different extensions then some extras for those and I got it working.

TGCMbetcher:
I fear that if this isnt merged first there is going to be a conflict so maybe I should just shut it down. The conflict was just a change log thing. Someone wrote a change log for railguns on crash and merged it and suddenly a merge conflict. I did this directly from git the way I used to. If it doesnt pass Ill just close it

Lumipharon (approved):

PR #14539, opened 2023-11-25
Nerfs Laser MG melting mode

yyzsong:
## About The Pull Request
See title. Laser MG now has 0.4 seconds of fire delay with a 0.3 second windup, as well as being semiauto

## Why It's Good For The Game
Each stack of melting does 3 sunder. The laser MG applies two stacks of melting per shot, which translates to 6 sunder per shot, or 18 sunder per second pre nerf, which is pretty nutty for a free gun with free ammo. This PR makes the sunder still be good while nerfing the sunder per shot.

## Changelog

:cl:
balance: Increases laser MG melting mode firedelay and makes it require windup on every shot.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -970,9 +970,9 @@
 /datum/lasrifle/energy_mg_mode/standard/melting
 	rounds_per_shot = 20
 	ammo_datum_type = /datum/ammo/energy/lasgun/marine/autolaser/melting
-	fire_delay = 0.4 SECONDS
+	fire_delay = 0.5 SECONDS
 	fire_sound = 'sound/weapons/guns/fire/laser_rifle_2.ogg'
-	windup_delay = 0.3 SECONDS
+	windup_delay = 0.4 SECONDS
 	fire_mode = GUN_FIREMODE_SEMIAUTO
 	message_to_user = "You set the machine laser's charge mode to melting."
 	radial_icon_state = "laser_heat"
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -970,9 +970,9 @@
 /datum/lasrifle/energy_mg_mode/standard/melting
 	rounds_per_shot = 20
 	ammo_datum_type = /datum/ammo/energy/lasgun/marine/autolaser/melting
-	fire_delay = 0.5 SECONDS
+	fire_delay = 0.4 SECONDS
 	fire_sound = 'sound/weapons/guns/fire/laser_rifle_2.ogg'
-	windup_delay = 0.4 SECONDS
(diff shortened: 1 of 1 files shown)


DeltaFire15:
>This PR makes the sunder still be good while nerfing the sunder per shot.

You might've changed your direction of this PR, since that isn't what this PR modifies
(Also random fact, hivelord jelly protects you from the Melting debuff & purges it if on you when applied)

yyzsong:
> > This PR makes the sunder still be good while nerfing the sunder per shot.
> 
> You might've changed your direction of this PR after writing this, since that isn't what this PR modifies (Also random fact, hivelord jelly protects you from the Melting debuff & purges it if on you when applied)

or i fucked up typing :agony:

TGCMbetcher:
Alright I just played over 3 and a half hours of atleast a 4 hour round where we had these issues. What would have solved it? 3 primodrones and an armor fruit garden. They head 30 armor each come back like every minute and no one wants to use them. I had a garden of like 15 and I had to physically drag people over to them. Drones also heal 5 sunder every 5 seconds. Late game armor shouldnt be an issue. you can plant an armor fruit every 40 seconds as a drone. In 5 minutes you should have a new garden at the new spot that will heal all sunder issues anyone has. The real issue is the other drone didnt want to add to the garden, they added 2 fruits total, and no one wanted to use it at all. Some of the people I drug didnt even eat a flower they just left. 1 flower cures 30 sunder, between that natural healing and 2 salves everyones back to full in like 15 seconds.

Lets compare guns. 
Melting machinegun, 75 damage 18 sunder 20 pen .9 seconds
T12, 125 damage 2.5 sunder 5 pen 1 second
t35 fletch 65 damage 8.5 sunder 15 pen 2 seconds
p-23 150 damage 10.8 sunder 5 pen 1 seconds (twice t12 falloff bad ammo bad pen makes it only good for sunder, loses 5 dps per tile of distance and you need to reload less than 3 seconds of shooting so its only good on paper like the t19)

So if we want to bring it in line to t12 like everything should be youre proposal is "62.5 damage 15 sunder 20 pen 1 second" which is still generous honestly I would make it do 1 stack of melting per second and place the sunder per stack at 4 instead for "60 damage 12 sunder 20 pen .9 seconds" because its HITSCAN and can be shot with while ignoring range because its a sunder weapon like the pistol. What it does is creates a wall the xenos have to cross before they can attack that eliminates all armor. Recovery can be ignored because there are ingame resources to easily deal with that sunder. A 3 nerf and a bit of damage isnt going hurt its role. This brings it in line with the other major free sunder weapon as well, more sunder but it cant aim.

Lumipharon:
It has hard counters, it doesn't work against jelly at all.

Also not sure why you are triple nerfing it with ROF, wind up, AND semiauto.

PR #14547, opened 2023-11-27
removes rds compatibility with mg60

coldironwarrior:
## About The Pull Request

removes rds compatibility with mg60

## Why It's Good For The Game

you can buff to 100% faster firing with bipod and RDS; so this would make using aimmode a bit more of a tradeoff -- you want to shoot past allies, your shots have to be a little slower, etc. you can also still combo bipod with aim order to reach 100% faster firing, this just makes that maybe require a little more strategy for command to make the decision to use aim order for that benefit.

open to editing

## Changelog

:cl:
balance: removed the ability to use the red-dot sight on the MG-60
removal: the MG-60 can no longer take the red-dot sight as an attachment
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1016,7 +1016,7 @@
 	starting_attachment_types = list(/obj/item/attachable/verticalgrip, /obj/item/attachable/reddot)
 
 /obj/item/weapon/gun/rifle/standard_gpmg/machinegunner
-	starting_attachment_types = list(/obj/item/attachable/stock/t60stock, /obj/item/attachable/foldable/bipod, /obj/item/attachable/extended_barrel)
+	starting_attachment_types = list(/obj/item/attachable/stock/t60stock, /obj/item/attachable/foldable/bipod, /obj/item/attachable/magnetic_harness, /obj/item/attachable/extended_barrel)
 
 //-------------------------------------------------------
 //M41AE2 Heavy Pulse Rifle
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -975,7 +975,6 @@
 	default_ammo_type = /obj/item/ammo_magazine/standard_gpmg
 	allowed_ammo_types = list(/obj/item/ammo_magazine/standard_gpmg)
 	attachable_allowed = list(
-		/obj/item/attachable/reddot,
 		/obj/item/attachable/flashlight,
 		/obj/item/attachable/lasersight,
 		/obj/item/attachable/flashlight/under,
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1016,7 +1016,7 @@
 	starting_attachment_types = list(/obj/item/attachable/verticalgrip, /obj/item/attachable/reddot)
 
 /obj/item/weapon/gun/rifle/standard_gpmg/machinegunner
-	starting_attachment_types = list(/obj/item/attachable/stock/t60stock, /obj/item/attachable/foldable/bipod, /obj/item/attachable/reddot, /obj/item/attachable/extended_barrel)
+	starting_attachment_types = list(/obj/item/attachable/stock/t60stock, /obj/item/attachable/foldable/bipod, /obj/item/attachable/extended_barrel)
 
 //-------------------------------------------------------
 //M41AE2 Heavy Pulse Rifle
--- a/code/modules/projectiles/gun_attachables.dm
+++ b/code/modules/projectiles/gun_attachables.dm
@@ -574,11 +574,12 @@ inaccurate. Don't worry if force is ever negative, it won't runtime.
 
 /obj/item/attachable/reddot
 	name = "red-dot sight"
-	desc = "A red-dot sight for short to medium range. Does not have a zoom feature, but does increase weapon accuracy. \nNo drawbacks."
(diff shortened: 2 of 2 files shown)


Strayed-Raven:
ah yes. instead of a gun having its attachment removed from play, we will now go after benefits that an attachment gives for all guns.

Pariah919:
Local man sends RDS back to worthlessness realm.

Mrrpip:
....There's already a short ranged scope. It's callef a miniscope. ._.

Regardless, I say merge it. Then we can have the more useful combo of Scope + Bipod + Focus order become popular.

Lumipharon:
Is RDS of any concern on any gun except the T60? Because uh, there's a better solution in that case.

coldironwarrior:
> Is RDS of any concern on any gun except the T60? Because uh, there's a better solution in that case.

i think it's literally just t60 but i'm very open to changing this to end up just making it incompatible with t60 only; this is an iterant form thing, lmk if making incompat. with t60 would be preferable. i'd like to know what ivan says though since idk if they have any other guns of concern with rds compatibility

Lumipharon:
if its an issue on the t60, then address that. Don't take things out of scope.
Alternatively change it to something different. For example faster aim time or some shit, I unno.

coldironwarrior:
> if its an issue on the t60, then address that. Don't take things out of scope. Alternatively change it to something different. For example faster aim time or some shit, I unno.

will do; this is just what i pr'd since this is what someone requested without anything fancy regarding incompatibility on mg60, but point is taken and i'll edit this.

coldironwarrior:
more work to do on this with the variant attachments, will get back to it in a bit tho

ivanmixo:
mmm super weird to remove it from just the t60, theyll just switch to the 42 lol (and if you remove it from 42 as well then theyll just burst it because that was a necessary addition to that gun)

Lumipharon:
> mmm super weird to remove it from just the t60, theyll just switch to the 42 lol (and if you remove it from 42 as well then theyll just burst it because that was a necessary addition to that gun)

you're trading like 15% aimmode dps for something like 45% nonaimmode dps by switching to a T42 (and I guess the ability to have a scope). If they want that mega specific niche, good for them I guess?

ivanmixo:
-25% off aim mode. Not counting autoburst which probably has even more non aim mode (and aim mode) dps.

Pariah919:
> mmm super weird to remove it from just the t60, theyll just switch to the 42 lol (and if you remove it from 42 as well then theyll just burst it because that was a necessary addition to that gun)

T-42 has the same burst as a T-12 almost papercut, its copy pasted. The only real honest to god noticeable difference is your mag size and if you want to be snide about it, packets make that honestly nigh irrelevant for bursting down walls considering you can hold a ungodly amount of ammo with packets.
If the main issue with gpmg aimballs is the ability to burst down walls/beans a t-12 and a t-42 have the same effective ability to do it in aimmode, considering the main big thing the gpmg has is that +5 pen and bipod+rdsed fire rate.

code/modules/projectiles/guns/rifles.dm
```diff
@@ -1017,7 +1016,7 @@
 	starting_attachment_types = list(/obj/item/attachable/verticalgrip, /obj/item/attachable/reddot)
 
 /obj/item/weapon/gun/rifle/standard_gpmg/machinegunner
-	starting_attachment_types = list(/obj/item/attachable/stock/t60stock, /obj/item/attachable/foldable/bipod, /obj/item/attachable/reddot, /obj/item/attachable/extended_barrel)
+	starting_attachment_types = list(/obj/item/attachable/stock/t60stock, /obj/item/attachable/foldable/bipod, /obj/item/attachable/extended_barrel)
```
Lumipharon, line 1019:
```suggestion
	starting_attachment_types = list(/obj/item/attachable/stock/t60stock, /obj/item/attachable/foldable/bipod, /obj/item/attachable/magnetic_harness, /obj/item/attachable/extended_barrel)
```

TGCMbetcher:
> Is RDS of any concern on any gun except the T60? Because uh, there's a better solution in that case.

Pretty much every rifle would like it more than a mag harness, but the question is it worth 6 mag space. You can tac reload from the lightweight backpack so the answer is yes, yes it is. The only reasons not to use it on a rifle instead is if youre using 2 weapons and cant spare the room in your build, you have no aim mode, or you want to use another attachment. I guess you could be a medic or engineer and be space strapped for other reasons than the weapon, but red dot over mag harness is a no brainer. There is a noticeable difference between with and without it, and using a gyro basically turns aim mode into a wield delay more than anything else.

SimplySmiley123:
> ....There's already a short ranged scope. It's called a miniscope. ._.
> 
> Regardless, I say merge it. Then we can have the more useful combo of Scope + Bipod + Focus order become popular.

And using scopes in this game is a chore, and honestly not worth it with faster firing weapons.
It would genuinely jsut return to magharness 100% of the time again since no other rail attachment besides it and RDS is worth bothering with on weapons that aren't dedicated hard hitting sniper rifles.

ivanmixo:
Wharever we'll deal with the next one when we get to it I suppose

code/modules/projectiles/guns/rifles.dm
```diff
@@ -1017,7 +1016,7 @@
 	starting_attachment_types = list(/obj/item/attachable/verticalgrip, /obj/item/attachable/reddot)
```
Xander3359, line 1016: How was this merged with checks that fail?

chizzy376:
After months of consideration, this PR being merged helped me finally decide that I'll become a moper, thank you Ivan

thatguythere03:
You might as well remove the gun from the game entirely given that you just made it unusable with this PR

PR #14548, opened 2023-11-27
Adds a minimum travelling time to CAS ordnance

lKiseki:
## About The Pull Request
CAS ordnance now has a minimum of 1.5 seconds. Travelling times can not go below this.

## Why It's Good For The Game
Inhuman reactions should not be required to play the game.
Yes, this is a salt PR.
No, I don't care. Take it as you will.

## Changelog
:cl: Lewdcifer
balance: CAS ordnance now has a minimum travelling time of 1.5 seconds.
/:cl:


Pariah919:
Salted crackers very good would recommend

Other than that like why, what can even go below 1.5 seconds?

dopamiin0:
> Salted crackers very good would recommend
> 
> Other than that like why, what can even go below 1.5 seconds?

widowmakers and highvelocity 30mm are both 1 second on low orbit. nothing else can reach this speed.

speaking of which, you should probably reduce the price of those two or otherwise buff them if you're making this change. this kinda, you know. steps on their niche.

Knightfall5:
Makes the already expensive Lasgun completely useless, it's entire gimmick is how fast it hits. Makes widowmaker worse, it's entire gimmick is that it's the fast rocket but doesn't do as much on impact. Makes the HV 30mm ammo pointless, it's entire gimmick is it's more expensive than regular 30mm ammo but hits sooner.

Salt PR with zero thought or effort put into it, CAS already has terrible point income on low orbit to rearm.

Lumipharon:
yeah some ammo is specifically meant to go fast, this is a giga cope pr.
If you want to argue the merits of just removing orbital levels entirely then go for it, but this ain't it.

PR #14552, opened 2023-11-28
Makes standard marine vendors unwrenchable

ivanmixo:
## About The Pull Request

Namely the clothing, armor, mech and youtool vendor.
## Why It's Good For The Game

These should not be something you should be able to bring groundside ever, they bring shipside level supply groundside at no cost.
## Changelog
:cl:
balance: Makes standard marine vendors unwrenchable
/:cl:

### DIFF
--- a/code/game/objects/machinery/vending/marine_vending.dm
+++ b/code/game/objects/machinery/vending/marine_vending.dm
@@ -998,6 +998,7 @@
 	icon_vend = "surplus-vend"
 	icon_deny = "surplus_armor-deny"
 	isshared = TRUE
+	wrenchable = FALSE
 	product_ads = "You are out of uniform, marine! Where is your armor? Don't have any? You expect me to believe that, maggot?;Why wear heavy armor and unable to chase the enemy when you can go light and zoom by your peers?;Thank your armor later when you didn't die!;I remember PAS, do you remember PAS?;Time to paint the rainbow!;So many selections to choose from!"
 	products = list(
 		"Xenonauten" = list(
@@ -1143,6 +1144,7 @@
 	icon_state = "surplus_clothes"
 	icon_vend = "surplus-vend"
 	icon_deny = "surplus_clothes-deny"
+	wrenchable = FALSE
 	isshared = TRUE
 	product_ads = "Be the musician that you parents never approve you of.;You gotta look good when you're in the battlefield.;We have all types of hats here!;What did one hat say to the other on the hiking trip? I'll wait here, you go on ahead;Sometimes, a beret is better than a helmet.;Drip is the priority, marine."
 	products = list(
@@ -1623,6 +1625,7 @@
 	icon_deny = "tool-deny"
 	icon_vend = "tool-vend"
 	isshared = TRUE
+	wrenchable = FALSE
 	products = list(
 		/obj/item/stack/cable_coil = -1,
 		/obj/item/tool/crowbar = -1,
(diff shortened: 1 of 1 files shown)


Lumipharon:
what did the youtool ever do to you?

ivanmixo:
Not exactly a fan of infinite healing for the clankers groundside

EvilMelonTheSecond:
So instead people should vend as much shit as they can and drag it with crates? I don't see the problem of having a vendor that needs power and is hard to drag around being deployed in FOB. Not sure if this will do more than add more prep time outside of very niche cases

ivanmixo:
We should be consistent. One type of vendor being wrenchable while all the others aren't just because it's not as egregious is a bit silly.

EvilMelonTheSecond:
> We should be consistent. One type of vendor being wrenchable while all the others aren't just because it's not as egregious is a bit silly.

That's a weak reason, the lasgun battery charger can move for QoL and youtool is not that different in practice. If you mean its because it has the sprite and title of a vendor those are easy changes to make to differentiate it. Throwing away functionality because of how it looks is not good reasoning.

ivanmixo:
It is. Also the laser recharger thingy is very much meant to be movable, it was added with that in mind. Who in their right mind considers it a shipside vendor? Why would you recharge cells slowly when the weapons vendor is right there to vend fresh cells?

lbnesquik:
How about removing their movement in the alamo or tadpole? That would achieve the same thing, but marines could still shuffle things around.

Lumipharon:
> We should be consistent. One type of vendor being wrenchable while all the others aren't just because it's not as egregious is a bit silly.

I mean, most are still wrenchable, and the tool vendor is groundside as well

chizzy376:
Absolutely not, bad PR.

TellyLsc:
What's the problem with those vendors being dragged to planetside? 
Outside of the LZ FOBs and Tads it's really a pain in the ass to make sure the area APCs don't get slashed in the first place, so I don't buy "dragging it to the frontline and having infinite supply on the spot" bit.

And if the issue is that "the FOB has too many supplies" well isn't that what the FOB is supposed to be? These vendors don't even have the most essential part of FOB defense: ammo; other than welding fuel and TE Powerpacks, which then maybe just move them into the green vendors that are already unwrenchable then?
What else? Pre-filled flare pouches lighting up areas surrounding the FOB? That one isn't even an issue of "infinite flares" like in that other PR from couple weeks ago, it's just an issue of couple guys actually willing to flare areas and keep it lit up for once, I'd know because I was one of those guys and let me tell you: I only use flare pouches because it was a pain in the ass to right click flare packs inside backpacks (which could've carried 12*14 flares mind you instead of the pouch's 28) since you'd have to sprite-hunt the flare pack, you can't just right click the blank space around the sprite inside storage like you can with slot based storages like flare pouch, which REALLY adds up for how much of a chore spamming flare is because god forbid marines have autoflares like xenos have autoweed.
Alright what other thing then? Machete belts? Last time I left a pile of those on FOBs no one really picked them up. Prefilled BKTT reagent pouches? Yeah sure, because infinite BKTT really grinds FOB sieges to a halt by making marines unkillable as RR does. Oh wait, they don't.
Oh wait I get it, it's the protein pack belts isn't it?

I'll throw a bone here, how about making the vendors only movable by clicking on them with powerloaders/making dragging them to be VERY SLOW instead of making them unwrenchable?

PR #14554, opened 2023-11-28
Engineer teleporter rebalance

lKiseki:
## About The Pull Request
The engineer's teleporter is no longer able to teleport carbon mobs, such as marines and xenos.
To compensate, the teleporter is now able to teleport just about anything besides mobs (marines and xenos) and the nuke itself.

## Why It's Good For The Game
Removes a largely exploitable aspect of the teleporter, while giving it a much more varied and accessible niche in compensation.

## Changelog
:cl: Lewdcifer
balance: Engineer teleporter can now teleport anything with the exception of human mobs (marines) and the nuke.
/:cl:

### DIFF
--- a/code/game/objects/structures/teleporter.dm
+++ b/code/game/objects/structures/teleporter.dm
@@ -4,14 +4,9 @@
 	max_integrity = 200
 	resistance_flags = XENO_DAMAGEABLE
 	idle_power_usage = 50
-	///List of all teleportable types
-	var/static/list/teleportable_types = list(
-		/obj/structure/closet,
-		/mob/living/carbon/human,
-		/obj/machinery,
-	)
 	///List of banned teleportable types
 	var/static/list/blacklisted_types = list(
+		/mob/living/carbon/human,
 		/obj/machinery/nuclearbomb
 	)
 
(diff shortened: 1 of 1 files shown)


MiniMeatwad:
A based engineer may use it as a trap to teleport living xenos into a kill zone.

lKiseki:
> A based engineer may use it as a trap to teleport living xenos into a kill zone.

ah fuck me sideways hold on

DeltaFire15:
>Removes a largely exploitable aspect of the teleporter [...]

Uh? Exploitable implies there is an exploit that can be located in there and not that it is being used for the purpose it was added for. The language here seems super icky, at least the branch name got it right.

lKiseki:
> > Removes a largely exploitable aspect of the teleporter [...]
> 
> Uh? Exploitable implies there is an exploit that can be located in there and not that it is being used for the purpose it was added for. The language all over the PR body seems super icky, at least the branch name got it right.

Since you want to get technical:
image
This is referencing the fact that you can use these to bypass traversing maps almost entirely.
Looking forward to seeing you in my next PR!

Pariah919:
Isn't this just kind of worthless
I mean TP is gamebreaking but if logi is so not real as people say why would I ever teleport anything from fob to front? Just drop it at the literal front

Unless you make it so cheap as to basically make it standard issue which is also silly

Lumipharon:
I don't understand what the point of a teleporter would be, after this change.

SimplySmiley123:
I honestly don't think it would be really good for anything after this change? It's problematic right now, yes, but what niche does this fill in really that isn't outclassed already by requisitions already being able to teleport crates directly to any beacon or marine with antenna helmet module?

> Though the more polsci inlcined part of me thinks this is just soft removal which is whatever, it can rot.

And yeah this does honestly sound like soft removal.

lbnesquik:
Can you present an use case for a teleporter that only sends items?

PR #14389, opened 2023-11-05
Gorger Feast is AOE + lower cooldown

Xander3359:
## About The Pull Request
Atomization: https://github.com/tgstation/TerraGov-Marine-Corps/pull/13317
Feast is now AOE + lower cooldown (30 seconds from 180)
## Why It's Good For The Game
Copypaste: "Gorger is really uncomfortable to play in general, these are merely QoL considering what state he is in right now, Gorger cannot sustain effectively neither offensive nor support abilities, some abilities are very rarely used in general, this tries to correct that (very successfully, considering RU experience with it). Probably needs a testmerge to estimate how good/bad these changes are. These changes are present on RU for 3 months without any balance issues."
## Changelog
:cl:
balance: Gorger Feast is AOE
balance: Gorger Feast now 30 seconds (180 seconds before)
/:cl:

PR #14560, opened 2023-11-28
Gives gorger plasma on attack

Xander3359:
## About The Pull Request
Gives gorger plasma on attack
## Why It's Good For The Game
Gorger struggles with gaining blood after using carnage and feast.
This buff aims to allow Gorger to take longer engagements and have some source of blood that doesn't stun himself.
## Changelog
:cl:
balance: Gorger gains blood on hit
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/gorger/castedatum_gorger.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/gorger/castedatum_gorger.dm
@@ -60,6 +60,14 @@
 		/datum/action/xeno_action/activable/devour,
 	)
 
+/datum/xeno_caste/gorger/on_caste_applied(mob/xenomorph)
+	. = ..()
+	xenomorph.AddElement(/datum/element/plasma_on_attack, 1.5)
+
+/datum/xeno_caste/gorger/on_caste_removed(mob/xenomorph)
+	. = ..()
+	xenomorph.RemoveElement(/datum/element/plasma_on_attack, 1.5)
+
 /datum/xeno_caste/gorger/normal
 	upgrade = XENO_UPGRADE_NORMAL
 


lbnesquik:
Practically speaking, how much plasma is that? Is it a good chunk of the gorger plasma reserves?

Arctrooper8:
What happened to the Gorger buff that was definitively liked: https://github.com/tgstation/TerraGov-Marine-Corps/pull/14389

Also I don't think this is a good buff for gorger since it discourages using her abilities, gives them too much blood, and encourages them into just being Rav 2.0

Xander3359:
> What happened to the Gorger buff that was definitively liked: #14389

Maintainers didn't want the buff. Only Ivan had an opinion on it and they were against it.
It was amazing to play with but it seems like we can't have nice things.
> Also I don't think this is a good buff for gorger since it discourages using her abilities, gives them too much blood, and encourages them into just being Rav 2.0

Contrary to popular belief, gorger WANTS a lot of blood. It means you can heal a lot and use your other abilities (feast/primo)
The amount of blood you get might be a bit overtuned, which I'd like to adjust (via TM), but I don't think this steps on the idea that rav has (rip and tear)

imsxz:
gorger does want a lot of blood but it also wants to spend it very quickly because it loses the stun+movespeed on carnage if you have too much. it can become an issue especially because it takes a while to empty out blood using just the overheal, takes over a minute to use 400 blood even if theres enough xenos around to spend it on

Xander3359:
> gorger does want a lot of blood but it also wants to spend it very quickly because it loses the stun+movespeed on carnage if you have too much. it can become an issue especially because it takes a while to empty out blood using just the overheal, takes over a minute to use 400 blood even if theres enough xenos around to spend it on

Feast and the primo ability both exist.
Although the feast ability would be a lot more usable if they didn't stale the buff PR to death.
I want to give him a new primo/rework his current primo, so that could be another good way to use blood.

Xander3359:
Maintainers are clearly not interested in any sort of Gorger buff.

PR #14561, opened 2023-11-28
Remove max client version

MrStonedOne:
We have to co-exist with the other ss13 servers in the overall landscape and making clients downgrade builds conflicts with that. known bad versions can be blacklisted but telling clients not to use newer builds is reckless and puts clients in a hard place where they have to decide between servers that don't work with the older builds and servers that don't work with the newer builds.


yyzsong:
Salt PR?

lbnesquik:
Does this impact the game in any way? More instability or dev work? Or is this just a sort of administrative stance?

PR #14570, opened 2023-11-30
Autoweed exploit fix

Lumipharon:
## About The Pull Request
Autoweed was bypassing a check.
## Why It's Good For The Game
Exploit bad.
## Changelog
:cl:
fix: fixed an exploit with autoweeding
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -125,7 +125,9 @@
 		return
 	if(get_dist(owner, last_weeded_turf) < AUTO_WEEDING_MIN_DIST)
 		return
-	if(!can_use_action(silent = TRUE))
+	if(!can_use_action(TRUE))
+		return
+	if(!can_use_ability(xeno_owner.loc, TRUE))
 		return
 	plant_weeds(owner)
 


coldironwarrior:
i am curious about how this works for the hivemind in the context of the idea that:
autoweed shouldnt work in the FOB, but manual weeding does work in the FOB --> hiveminds are not able to manually weed the FOB --> the loss of autoweeding on the hivemind removes the only way for them to weed the FOB --> this leaves them as the only caste capable of weeding with no way to weed the FOB
all while totally recognizing that allowing manual weeding would be a disaster because there's no risk necessary for that as hivemind can do it by line of sight from outside the FOB. i am wondering if there's not some sort of alternative though between these two that would still allow a weeding caste mechanical ability in this manner that is consistent with other weeding castes.

additionally, i believe this potentially makes areas that TAD has landed and taken off from unweedable for hiveminds, as these similarly require manual weeding from other castes because of shuttle areacoding. what might be a solution for this to prevent there being tad-sized gaps in mazes, as re/mazing potential landing zones is highly important?

**tldr, and to be clear about what the outcome of this PR is supposed to be**: are you intending to make hivemind unable to weed FOB and TAD landing spots in any capacity?

Oscari-pog:
This is not a fix of a exploit, this is a nerf.

Lumipharon:
It only worked because the code bypassed proper checks that it should be checking no matter what.

TiviPlus:
Im not even going to tolerate the idea that this isnt an exploit

PR #14591, opened 2023-12-03
Fix combat log runtime

ivanmixo:
## About The Pull Request

Some of them don't have a user or target, this isn't really an issue
## Why It's Good For The Game

runtime bad
## Changelog
:cl:
fix: Fixed a combat log runtime
/:cl:


vvvv-vvvv (changes requested):
i'd argue that they shouldn't ever be null instead as log_combat is for atom-on-atom attacks

Lumipharon:
curious how this is actually coming up, yeah

PR #14593, opened 2023-12-05
warlock psyblast nerf

Lumipharon:
## About The Pull Request
Psyblast/lance now requires the target to be in screen range to successfully cast.
What this means is if you target a mob, if they get off screen quickly enough the ability will fail to cast.

This doesn't make that much difference to blast consider it already has shorter range, but makes lance less guh to deal with as you get offscreened.

You can of course still jsut aim at a turf to line up the offscreen, but you lose the autotracking.
## Why It's Good For The Game
Less oppressive lance tracking shenanigans.
## Changelog
:cl:
balance: Psylance and Psyblast will now fail to cast if their target moves off screen
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/warlock/abilities_warlock.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/warlock/abilities_warlock.dm
@@ -542,10 +542,7 @@
 
 /datum/action/ability/activable/xeno/psy_blast/use_ability(atom/A)
 	var/mob/living/carbon/xenomorph/xeno_owner = owner
-	//var/turf/target_turf = get_turf(A)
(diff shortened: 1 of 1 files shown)


Strayed-Raven:
Next up, Warlock AOE cannot be quick cast on one single tile/target.

coldironwarrior:
just a quick question, but would this still burn the plasma cost of casting and set the cooldown if it fails to track a target?

Lumipharon:
> just a quick question, but would this still burn the plasma cost of casting and set the cooldown if it fails to track a target?

No, it just fails, same as if you were stunned or something mid cast.

ivanmixo (approved):

PR #14594, opened 2023-12-05
Kill antenna beacon

Lumipharon:
## About The Pull Request
Antenna no longer works as a req beacon.
## Why It's Good For The Game
Antenna is an extremely good, free item. In many cases theres literally no reason to use another module on your helmet, and in the rest of cases you can just put it on a spare helmet in your bag, and its still better than an actual beacon since you're storing extra shit in the helmet itself.
## Changelog
:cl:
balance: Antennas no longer function as req beacons
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -611,7 +611,7 @@
 
 /obj/item/armor_module/module/antenna
 	name = "Antenna helmet module"
-	desc = "Designed for mounting on a modular Helmet. This module is able to provide a readout of the user's coordinates and connect to the shipside supply console."
+	desc = "Designed for mounting on a modular Helmet. This module is able to provide a readout of the user's coordinates."
 	icon = 'icons/mob/modular/modular_armor_modules.dmi'
 	icon_state = "antenna_head"
 	item_state = "antenna_head_a"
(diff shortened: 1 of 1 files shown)


Xander3359:
Rename it to "compass module"

Barnet2:
Beacon on demand is guh, but this makes it useless. Use the cave comms idea.

Lumipharon:
Only reason I didn't nuke it entirely is because people wear it for fashion.

DeltaFire15:
> That takes up the same space as beacon if I remember right and is more an issue of beacon taking up way too much space.

Xander made a good point on this which I also missed when I made this argument, which is that the helmet in your backpack also holds two tiny items, making it more space efficient.

Strayed-Raven:
Just make helmet with antenna become a larger item for weight lol

coldironwarrior:
hero! neat. coords will still be interesting to tinker with.

not for this PR but i'd love to see it lean into comms aspect more if possible. people floated the idea of amplifying radio signal in caves so shallow can be heard clear and deep can be heard at shallow clarity. i'd be interested in re-instating something like the old radiopacks where you can just place calls to other antenna users.

Lumipharon:
> > and in the rest of cases you can just put it on a spare helmet in your bag
> 
> That takes up the same space as beacon if I remember right and is more an issue of beacon taking up way too much space.

A beacon takes up 2 spaces. A helmet takes up 3 space. But for that 1 extra space, you get 2 extra slots of storage, so the effective storage consumed by a helmet with antenna is less than a beacon (unless you have no space at all for a normal sized item)

Someone is welcome to do some comm stuff if they want, but you'll likely have to refactor comms to make it work the way you'd expect it to.

Strayed-Raven:
> How many people are actually carrying around second helmets just to have an antenna? Or lug it around to juggle 2 small item slots? I've seen about two people carry a second helmet ever. It seems like a lot of hassle for very little gain.
> 
> If we go back to needing coords for req then it's just going to be a matter of putting 1 compass in your helmet and making req gameplay more tedious as people have to copy and paste coordinates instead of having a helmet module to make what needs to be sent to who much clearer. Or having to lug around beacons everywhere if you want to play the game, which is... worse.

The nerfs of the many are paved by the few

efzapa:
>no reason to use any other module

then buff the other modules? wut

Strayed-Raven:
Seriously do not get why the easier way of increasing the space the helmet with antenna takes by default is not being taken here.

ivanmixo:
Ill close due to the other pr being merged

PR #14604, opened 2023-12-06
C4 (Plastique Explosives) No Longer Require A Multitool To Disarm

Wisemonster:
## About The Pull Request
C4, or plastique explosives, no longer require a multitool to disarm, instead only requiring to click on the C4 to remove it.
## Why It's Good For The Game
Currently, C4 requires a multitool to disarm, meaning the only job that can reliably disarm the C4 are engies. Considering that both FC and SLs can acquire C4, this means that most of the time if a C4 is misplaced, it is just gonna destroy whatever (or whoever) it was planted on. This at least makes it easier to avoid accidents with C4.
## Changelog
:cl:
balance: C4 (Plastique explosives) no longer requires a multitool to disarm.
/:cl:

### DIFF
--- a/code/game/objects/items/explosives/plastique.dm
+++ b/code/game/objects/items/explosives/plastique.dm
@@ -136,6 +136,7 @@
 		alarm_sounded = FALSE
 		plant_target = null
 		update_icon()
+	return ..()
 
 ///Handles the actual explosion effects
 /obj/item/explosive/plastique/proc/detonate()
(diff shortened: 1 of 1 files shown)


Wisemonster:
If there's a concern that bald marines can disarm (and consiquently acquire) C4, an alternative could be requiring at least 1 level of engineering skill in order to disarm the C4. (Though this would mean adding more "Skill-locked" actions similar to the chem dispenser)

Lumipharon (approved):
eh, ok

PR #14606, opened 2023-12-07
Slightly reduces the cheese level of Wraiths

DeltaFire15:
## About The Pull Request 
Banished objects that are solid (have density) will now push out any living beings on their tile before returning, if possible. This can potentially be used offensively (to force targets off a tile), but less to cheese like the previous state allowed.
## Why It's Good For The Game
Aims to counteract a specific technique relying on ending up on the same tile as certain solid objects. You shouldn't be on that tile, and telegibbing would be very rude to anyone, so this it is.
## Changelog
:cl:
balance: Banished solid objects now push out any living beings on their return tile (if possible) before returning.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/wraith/abilities_wraith.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/wraith/abilities_wraith.dm
@@ -296,6 +296,16 @@ GLOBAL_LIST_INIT(wraith_banish_very_short_duration_list, typecacheof(list(
 	var/turf/return_turf = get_turf(portal)
 	if(!return_turf)
 		return_turf = locate(backup_coordinates[1], backup_coordinates[2], backup_coordinates[3])
+	if(banishment_target.density)
+		var/list/cards = GLOB.cardinals.Copy()
+		for(var/mob/living/displacing in return_turf)
+			if(displacing.stat == DEAD) //no.
+				continue
+			shuffle(cards) //direction should vary.
+			for(var/card AS in cards)
+				if(step(displacing, card))
+					to_chat(displacing, span_warning("A sudden force pushes you away from [return_turf]!"))
+					break
 	banishment_target.resistance_flags = initial(banishment_target.resistance_flags)
 	banishment_target.status_flags = initial(banishment_target.status_flags) //Remove stasis and temp invulerability
 	banishment_target.forceMove(return_turf)


AlexCrimson:
This isnt "slightly reducing" the cheese, its outright killing it. I assume this is done to stop Wraiths banishing turrets, standing on the tile, recalling them, and not being shot at? If so, it isnt a small change. A valid change, sure. Its cheese as hell. Just not a "slight" one.

Arturlang:
Honestly wraith without the cheesing abilities is really not all that useful, what's the point if you can't solo kill a sentry when other castes can? It'd be better to go something like a hunter instead as you'll be participate in combat properly, wraith is really damn  squishy not to mention struggles to deal damage. Wraith's going to be a whole lot less useful on their own as a result, basically requiring someone to support, which as shown by behemoth which requires pretty much the same, leads the caste to be pretty damn unpopular. Being able to backline and kill the sentries while nobody is guarding them makes a wraith useful atleast until you can evolve into t3 to do some actual damage, there's really only one person that I know that stays a wraith. But yes it's a problem with the caste itself if you can't be useful or practical without being able to cheese things.

Strayed-Raven:
> Honestly wraith without the cheesing abilities is really not all that useful, what's the point if you can't solo kill a sentry when other castes can? It'd be better to go something like a hunter instead as you'll be participate in combat properly, wraith is really damn squishy not to mention struggles to deal damage. Wraith's going to be a whole lot less useful on their own as a result, basically requiring someone to support, which as shown by behemoth which requires pretty much the same, leads the caste to be pretty damn unpopular. Being able to backline and kill the sentries while nobody is guarding them makes a wraith useful atleast until you can evolve into t3 to do some actual damage, there's really only one person that I know that stays a wraith. But yes it's a problem with the caste itself if you can't be useful or practical without being able to cheese things. Not to mention that the turret IFF change also made the portal sentry killing thing unusable, thus leading a wraith completely unable to kill sentries without a lot of effort. Lots of other castes can do the same, crusher, defiler, ravager, warlock , though in those cases they're all T3, and require access to the sentry without cades unlike a wraith, which gives marines more warning. Perhaps an alternative would be to make banishing sentries make them send a radio message and disabling the unbanish, perhaps giving it more of a delay before it shows up again, to nerf it without making it unusable?

The issue is how Wraith does the same thing as other castes. if all caste can say it has as an identity is 'cheesy abilities' then it should be removed instead. This keeps Wraith without touching all the other stuff it can do. Like you know. teleporting a mob with you.

Arturlang:
> The issue is how Wraith does the same thing as other castes. if all caste can say it has as an identity is 'cheesy abilities' then it should be removed instead. This keeps Wraith without touching all the other stuff it can do. Like you know. teleporting a mob with you.

I'm confused, all of it's abilities and what they do are pretty unique, what parts of it do the same thing as other castes?

blackdav123:
good change but now im imagining how funny giving them telefrags would be

coldironwarrior:
> Not to mention that the turret IFF change also made the portal sentry killing thing unusable, thus leading a wraith completely unable to kill sentries without a lot of effort.

is that why it's so easy for you to be a wraith, take out FOB turrets, get on alamo, devo, and become shipside shrike so often?

Derrick2Electricboogaloo:
Finally putting an end to this nonsense that makes wraiths so infuriating. Good work!

Strayed-Raven:
> > The issue is how Wraith does the same thing as other castes. if all caste can say it has as an identity is 'cheesy abilities' then it should be removed instead. This keeps Wraith without touching all the other stuff it can do. Like you know. teleporting a mob with you.
> 
> I'm confused, all of it's abilities and what they do are pretty unique, what parts of it do the same thing as other castes?

Your defense of Wraith was, whats the point of it if it can not do what other castes can? The issue I am point out is Wraith abilities completely invalidate a Sentry. More so than what the other castes can with complete impunity solo. Not too sure about Warlock because it has a shield. But on the other hand you can use portals to make a sentry kill itself or invalidate it shooting at you. So it does not need its phasing teleport as a 'kill sentry' button.

ivanmixo (approved):

PR #14614, opened 2023-12-08
Orders Show Job Title of Order Sender

Wisemonster:
## About The Pull Request
Orders will now show what job sent the order, rather than being ambiguous.
Change Me - Debugdalus 12_8_2023 6_48_20 PM
## Why It's Good For The Game
Knowing who sent the orders can influence whether or not you would want to follow em. This makes it clearer, especially in the case of conflicting orders.
## Changelog
:cl:
add: Orders will now show what job sent them in the orders' header.
/:cl:


FosterGrime-Heymynameslouie:
having the order senders name also show up in the thingy majiggy aslso would be very useful too
cool pr spedmeght

Wisemonster:
Sure, but some names are rather long, which would result in chat spam/bloat.

FosterGrime-Heymynameslouie:
> Sure, but some names are rather long, which would result in chat spam/bloat.

would the first name only showing up instead of the full name could possibly mitigate bloat!

TiviPlus (approved):

PR #14616, opened 2023-12-09
Canterbury interruptable takeoff delay

Arturlang:
## About The Pull Request
The canterbury takes 2 minutes to take off, with some UI changes to make some shuttles things more clear.
This can be interrupted by a xeno slashing the APC, and they get a HUD timer just like nuke or SD to inform them that the canterbury is about to evacuvate. The exact numbers can be changed if need be but 2 minutes seemed pretty reasonable.
## Why It's Good For The Game
Being able to silently and with no way to interrupt it, launching the canterbury has been pretty impossible for xenos to stop, I'm thinking these changes would lead to fun sieges while marines are desperately trying to stay alive and evac, while xenos do their utmost to break in.
Also adds indestructible shutters to the indestructible cockpit windows so it's a wee bit harder for a wraith to kill the APC.
## Changelog
:cl:
add: The TGUI shuttle UI shows the shuttle's status a bit better now.
add: The canterbury now uses the TGUI shuttle console, instead of the old HTML one.
add: Xenos get a HUD timer on their screen while the canterbury is trying to take off, to keep them better informed.
balance: The canterbury takes 2 minutes to take off/ignite it's engines, which can be interrupted by disabling power.
balance: The canterbury cockpit now has indestructible shutters for the cockpit windows. To stop wraiths. Unfortunately.
refactor: Xeno hive status datum can now support arbitrary hud timers, instead of each being hardcoded in.
refactor: The ShuttleControl tgui UI is now in typescript
/:cl:


Barnet2:
If the hull windows are made completely intangible (i.e wraith cannot teleport in, lasers cannot shoot out, basically they act as walls but you can see through them) this is a cool idea.

Arturlang:
> If the hull windows are made completely intangible (i.e wraith cannot teleport in, lasers cannot shoot out, basically they act as walls but you can see through them) this is a cool idea.
Engi turrets/Shocked razor/Marines guarding with laser guns counter this pretty well as long as it's not only one that the wraith can banish, plus a stagger or pepperball can trap the wraith very easily, not to mention how fast they die to a spray carbine

Arturlang:
Please testmerge this PR before merging, I faffed about with a lot of shuttle code so there's bound to be something I missed

Arturlang:
Feedback has caused me to give the canterbury windows indestructible shutters. Probably overkill honestly

Mrrpip:
Couldn't you just intentionally disable the APC as marine repeatedly to keep the timer low?

Arturlang:
> Couldn't you just intentionally disable the APC as marine repeatedly to keep the timer low?
You could, though I'm not sure how to stop this.

Lumipharon:
Why is this needed? Evacing is marines surrendering

Technical-l:
> Why is this needed? Evacing is marines surrendering

evaccing also only gives xenos a minor, which is instant and xenos cant really stop. if xenos ever get a major, it's always because the marines are really, REALLY bad and have no-one left to press the evac button.

yyzsong:
As good as killing everyone feels, there's no functional difference between a major and a minor. This only serves to mildly inconvenience marines while preventing xenos from getting into the cockpit; Breaking the Canterbury takes a bit more than two minutes anyways.

Arturlang:
I can change the time if need be, simply I wish it to be possible to interrupt the launch in some way, and have a warning before it happens.

Arturlang:
This also helps when some random marine decides to start a evac without consulting with the rest, allowing marines to stop it by turning off the APC.

Arturlang:
Another option instead of the shutters could be to move the APC to the cockpit inside the center of the canterbury. Should I do this instead?

yyzsong:
I don't feel like any of those are a good solution. At the least, you could lock canterbury to requiring an SL/FC/CMO/Synth ID to takeoff if having random marines launching it is a problem, or you could code a round time requirement for takeoff (Canterbury can't take off before it's been half an hour groundside, etc etc)

Arturlang:
> I don't feel like any of those are a good solution. At the least, you could lock canterbury to requiring an SL/FC/CMO/Synth ID to takeoff if having random marines launching it is a problem, or you could code a round time requirement for takeoff (Canterbury can't take off before it's been half an hour groundside, etc etc)

I'm fairly sure that it's already time limited, but limiting by roles will lead into marines being unable to call it when they really need to due to the lack of those roles, and fixing that is far more complex. Honestly if you can't hold 2 minutes in the canterbury the xenos have likely already breached and are about to wipe you, just that marines cannot fly off with xenos inside as it is interruptable.

InterroLouis:
Do you consider marines evacuating in crash to be a marine win?

efzapa:
this just seems like an excuse to lean toward an "all or nothing" playstyle, which can wind up being toxic 

Crash is already bad enough as a lowpop mode and should ideally be replaced, we don't need command doing all the things they already do AND considering if they need to bail within the next 2 minutes

Lumipharon:
I don't think any of the maints see a benefit for this.
Evacing is a xeno win and ends the round. Delaying evac is literally just delaying the round end when you've already won.

PR #14617, opened 2023-12-09
Fixes spiderlings not continuing their attack

Arturlang:
## About The Pull Request
This fixes spiderlings stopping their attack after one slash, now they'll only stop if the widow is not adjacent to them
## Why It's Good For The Game
TerraGov-Marine-Corps/pull/13713 had broken it by removing a check, without this it's much harder to have them dealing damage
## Changelog
:cl:
fix: Widow's spiderlings will no longer just constantly stop attacking their target, now the widow will have to move away from the spiderlings and only then they will stop
/:cl:


TiviPlus (approved):
https://tenor.com/view/cat-cats-cat-eating-eating-eating-cat-gif-8459125914348971806

PR #14618, opened 2023-12-09
tgui 5.0

vvvv-vvvv:
Brings us up with tg tgui, should be easy to review commit-by-commit. The main changes are:

[The OpenDream TGUI fix (#79244)](https://github.com/tgstation/TerraGov-Marine-Corps/commit/7fbe4b420d70d247de70db50e58400e4453c4f7b)

Modern browsers run the ready script before the DOM is loaded. Also
future proofs us for when Lummox upgrades the web engine.

[Remove context from tgui](https://github.com/tgstation/TerraGov-Marine-Corps/commit/7f2cddc6151602976c7fd0a04ecc1ad8af3b5dcc)

Removes all context from tgui, except in tgui-panel

\> This is a precursor PR for moving to React. Newer React (>16) does not
\> have the `context` parameter, so it needs removed. I want to make
\> converting to React look slightly less intimidating, so this is just
\> short-circuiting context here.

[Port React](https://github.com/tgstation/TerraGov-Marine-Corps/commit/0edb57ac569c99c877b1d8096eb14af81074a818)

\> Moves us from Inferno 7.4.11 to React 18.2. This is a big project!


Lumipharon:
I assume this pr caused this when trying to play internet music. 
image

TellyLsc:
Overwriting a loadout and loading the same loadout gets you the "Error when loading this loadout" message and deletes said loadout with this it seems.

Lumipharon:
image
Ghmme vendors are fucked

Lumipharon:
There's also 3 new bug reports that are related to this pr, but the person didn't ref the TM

Arturlang:
Port https://github.com/tgstation/tgstation/pull/80296 too please

tgui/packages/tgui/constants.ts
```diff
@@ -143,7 +143,7 @@ const GASES = [
     path: '/datum/gas/nitrogen',
     name: 'Nitrogen',
     label: 'N₂',
-    color: 'red',
+    color: 'yellow',
```
jlsnow301, line 146: careful, this does not seem to be react related

jlsnow301:
There seems to be some confusion with file naming convention- some .ts files were renamed into .tsx when they needn't be. Files without ui components are ok as TS/JS (the X means it contains jsx)

[One example](https://github.com/tgstation/TerraGov-Marine-Corps/blob/e6d7ad386b23e111b8650faa970090b268820a70/tgui/packages/tgui/interfaces/MechVendor/data.tsx)

tgui/packages/tgui/constants.ts
vvvv-vvvv: ya i cherrypicked this manually

ivanmixo (changes requested):
The `CargoRequest` interface got broken by the latest commit

novaepee:
for some people, spacebar keybind is not working, @DeltaFire15 found the issue
image

DeltaFire15:
Mhm, apparently something changed in preferences that makes "spacebar" binds get interpreted as " " instead of space. Currently set ones work fine but any new inputs in the kb settings that use space don't. Unsure what exactly is messing up the sanitze or translate there.

PR #14621, opened 2023-12-09
Snatch now spends plasma only on success.

Helg2:
## About The Pull Request
Snatch now spends plasma only on success.
I guess it could be called a balance change, but you're like spending plasma even after you're told that the target has nothing to steal.
## Why It's Good For The Game
Less confusing stuff, which is confusing just to be confusing.
## Changelog
:cl:
fix: Snatch now spends plasma only on success.
/:cl:


Strayed-Raven:
missing an ability should be punished.

Helg2:
Why so?

Al-1ce:
> missing an ability should be punished.

This is an old CM mindset and doesn’t really blend well. **Why?** exactly should this be punished?

ivanmixo (approved):

PR #10151, opened 2022-05-07
I have yet to meet man who can outsmart bullet (The SG Minigun)

Arctrooper8:
## About The Pull Request

image
This is a reopening of MJP's PR #10024
However it is slightly different in certain ways.
The most noticeable difference lies in the SG nomenclature, which now uses an SG (clever I know) instead of the strange T.
The second biggest difference is the fact that the SG-85 has a much much faster firerate (even faster than minigun, it's hilarious) and weaker bullets to compensate this truly suppressive firerate. (I don't know how to record videos please help)
**Of course values are subject to change and discussion, helpful feedback is appreciated.**

### Video
https://imgur.com/a/mwKVjxY

## Why It's Good For The Game

First and foremost: Fun, it's a lot of fun to wield a gun that fires this fast, even if the overall DPS is quite similar to the SG-29. 

Second and more importantly: the SG is supposed to be a backline support gunner that provides suppressive fire against Xenos. The SG-25 does not fit this role well, being an IFF assault rifle with a low magazine size (compared to normal smartguns) and faster usage and movement speed which promotes chasing Xenos. 
The SG-85 fits the role much better due to it's tyrannical firerate and high ammo count, if the SG-85 user attempts to run off on their own they're likely to be stunned and have to spend time reconnecting their minigun and then winding it up again.

The SG-25 supports a lone wolf chasedown style that contrasts with the SG's role.
Whereas the SG-85 encourages cooperation and teamwork whilst providing ample amounts of supporting fire.

TLDR:
### BRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRT
(It's a ton of fun, trust me)

## Changelog
:cl: Arctrooper8 Code: MJP Sprites: Kuro
add: SG-85 handheld gatling gun, an SG minigun with an absurd firerate at the cost of low damage per bullet.
del: T-25 has truly been sent to the shadow realm this time (They will not survive the Brazilians).
balance: SG-29 now has a firerate of .2 instead of .25
/:cl:

PR #9912, opened 2022-04-04
Swaps T-25 and -29 fire rates.

Pariah919:
## About The Pull Request
Swaps Smartrifle and Smart machinegun fire rates. 0.2->.25, and .25->.2

## Why It's Good For The Game
Smartrifle should get into the lower fire rate bracket and the -29 can go back up to the .2 bracket, I think it'll work out longterm, could be a nice test at the very least.

## Changelog
:cl:
balance: T-25 and T-29 fire rates have swapped. .2->.25 and .25->.2 respectively
/:cl:

PR #14627, opened 2023-12-11
SG-29 Fire delay .2->.25

Pariah919:
## About The Pull Request
Moves T-29's fire delay from .2 to .25.
## Why It's Good For The Game
I first tested moving T-29 from heavy rifle fire rate to rifle fire rate in https://github.com/tgstation/TerraGov-Marine-Corps/pull/9912, eventually after closing that PR Arctrooper made https://github.com/tgstation/TerraGov-Marine-Corps/pull/10151 which was a continuation of my PR and included the SG-29 buffs as a side effect.

This was a mistake.

T-29 should not have had its fire rate increased, it should have lesser DPS to compensate for its IFF, higher penetration and mesonnods, overall I tried the swap to see how the gun felt and personally I think it lead to powercreep as people said it 'Felt better' due to being stronger overall and being useable as a mere rifle with IFF rather than as a fire support gun so the powercreep was allowed to go through, generally speaking T-29 should move back down a level of fire rate and be back to its original state instead of having the ability and incentives to be used as just a rifle with IFF instead as of pure fire support.
I feel like if you want rate of fire you should move to the SG minigun, and if you want an all rounder move to the SG-29, if you want long range with a funny side gun, move to the SG DMR.
## Changelog
:cl:
balance: SG-29 now has 0.25 fire rate rather than 0.2.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1227,7 +1227,7 @@
 	starting_attachment_types = list(/obj/item/attachable/stock/sgstock, /obj/item/attachable/sgbarrel)
 	gun_skill_category = SKILL_SMARTGUN //Uses SG skill for the penalties.
 	attachable_offset = list("muzzle_x" = 42, "muzzle_y" = 17,"rail_x" = 15, "rail_y" = 21, "under_x" = 24, "under_y" = 14, "stock_x" = 12, "stock_y" = 13)
-	fire_delay = 0.2 SECONDS
+	fire_delay = 0.25 SECONDS
 	burst_amount = 0
 	accuracy_mult_unwielded = 0.5
 	accuracy_mult = 1.1


RagingLemons1:
machineguns have already been nerfed with the red dot t60 removal, you don't need to take it further. also, comparing the 29 to the sg dmr is a no go, the dmr is ass for hvh and the ammo is a pain in the ass to supply, compared to the 29 or the 85 which can be reqtorio'd

usnpeepoo:
merge this

Lumipharon:
If nothing else, this would make it garbage compared to the minigun

Nuker555:
instead of making the only usable sg worse make the bad ones viable?

ivanmixo:
Don't think this is necessary, but it shouldn't have happened in the first place in a completely unrelated PR. Oh well.

PR #14630, opened 2023-12-12
Skirmish Rifle buff, 30->40 mag cap

Pariah919:
## About The Pull Request
Skirmisher now has 40 magazine capacity.
## Why It's Good For The Game
Makes the skirmirsher a more enticing choice, it currently suffers from terrible ammo economy overall as it has significantly worse magazine capacity compared to other weapons in the same overall class (BR, StormR, AK, etc) so this should help it stand out more.
## Changelog
:cl:
balance: Skirmish Rifle has been buffed to 40 magazine capacity instead of 30.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/rifles.dm
+++ b/code/modules/projectiles/guns/rifles.dm
@@ -1686,7 +1686,7 @@
 	unload_sound = 'sound/weapons/guns/interact/t21_unload.ogg'
 	reload_sound = 'sound/weapons/guns/interact/t21_reload.ogg'
 	caliber = CALIBER_10X25_CASELESS //codex
-	max_shells = 30 //codex
+	max_shells = 40 //codex
 	force = 20
 	default_ammo_type = /obj/item/ammo_magazine/rifle/standard_skirmishrifle
 	allowed_ammo_types = list(/obj/item/ammo_magazine/rifle/standard_skirmishrifle)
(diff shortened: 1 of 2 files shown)


Lumipharon:
Why does this gun even exist?
Compared to a BR it has worse damage, worse AP, worse capacity and identical sunder.
Really the only thing that splits them apart is better accuracy and (slightly) better move speed). I guess you can strap more underbarrels to it though.

The gun is so redundant its not even funny.

Pariah919:
> Why does this gun even exist? Compared to a BR it has worse damage, worse AP, worse capacity and identical sunder. Really the only thing that splits them apart is better accuracy and (slightly) better move speed). I guess you can strap more underbarrels to it though.
> 
> The gun is so redundant its not even funny.

Almost everything has worse damage and AP compared to the battle rifle, its incredibly overtuned (the buff to the bullet and extra delay that happened way back was a huge mistake) even you know that. I don't even think its compareable considering one is full auto focused and the other one is burst fire focused? If you want to say that I might as well delete everything that isn't a BR or a lascarbine in the current stats..

Lumipharon:
Other guns are different enough in other ways. i.e. a t12 has notably faster move speed, where as skirm is almost identical.
Auto vs burst auto isn't really a difference either. You're still holding down m1 on someone with either.

Its not even that the skirm is bad, its just redundant.

FIXTHERADBOXPLEASE:
Don't play marine but automatic fire vs. burst fire is already a huge difference. The delay between bursts gives xenos a lot more options.

ivanmixo (approved):

PR #14632, opened 2023-12-13
Glass turf fix

Lumipharon:
## About The Pull Request
Fixes all types of glass turfs not actually respecting the pass_glass flag.
This means lasers will actually go through them as expected now.

Also fixed a couple other minor related alamo bugs.
The invisible dense turfs under the alamo computer consoles are no longer dense (so you can actually shoot through the front window)
The invisible attach points on alamo and tad (and the object they are linked to) no longer have 50% coverage (so your bullets aren't blocked by invisible funny effect)
## Why It's Good For The Game
Bug fix good.
## Changelog
:cl:
fix: fixed glass turfs not being passable by lasers and such when they should have been
/:cl:

### DIFF
--- a/code/game/objects/structures/dropship_equipment.dm
+++ b/code/game/objects/structures/dropship_equipment.dm
@@ -17,6 +17,7 @@
 	layer = ABOVE_OBJ_LAYER
 	dir = NORTH
 	density = TRUE
+	coverage = 0
 	var/base_category //what kind of equipment this base accepts.
 	var/ship_tag //used to associate the base to a dropship.
 	/// offset in pixels when equipment is attached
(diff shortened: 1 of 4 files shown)


Barnet2:
Shooting through the alamo windows is very :/. Same shtick as Canterbury windows. Also, wouldn't this mean wraith could TP through the Alamo windows?

ivanmixo:
I'm really not a fan of marines being able to shoot through alamo windows. Wraith can't blink in at least but this is stinky still.

lbnesquik:
Alamo windows are a pretty niche hold out. Like, they can't exactly defend much once they are shooting from there.

PR #14639, opened 2023-12-15
Rebinds unique action to C by default (instead of space)

ivanmixo:
## About The Pull Request

Why C? Because it's unused on humans. Also space to jump is more intuitive for newer people I feel like.

Keep in mind if you aren't connecting for the first time your keybinds won't be affected by this!!!
## Why It's Good For The Game

2 different things on the same key by default is a bit questionable.
## Changelog
:cl:
qol: Rebound unique action to C by default (instead of space)
/:cl:

### DIFF
--- a/code/datums/keybinding/human.dm
+++ b/code/datums/keybinding/human.dm
@@ -36,7 +36,7 @@
 	quick_equip_slot = 5
 
 /datum/keybinding/human/unique_action
-	hotkey_keys = list("Space")
+	hotkey_keys = list("C")
 	name = "unique_action"
 	full_name = "Perform unique action"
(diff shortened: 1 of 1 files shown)


blackdav123:
i feel like space will always make more sense because your thumb will be free and you will probably use unique action 3x more than jumping

Derrick2Electricboogaloo:
Already use this, saves me a step if my settings revert. Based

Lumipharon:
This is a default keybind, so space for jump makes sense considering space is jump for, basically every game in the last 25 years.

Looking4bros:
I shall suggest using alt+Q because that's what i do as i thought it'd be funny, and i think it would be even more funny to have it the default keybind.

PR #14645, opened 2023-12-16
Adds linters for duplicate windows and window frames.

Helg2:
## About The Pull Request
It would be usefull to have these, to delete duplicate windows and window frames.
If it doesn't fail, then it doesn't work.
## Why It's Good For The Game
Less duplicates.
## Changelog
Not player facing.


Helg2:
Well, okay, no idea how to make it work,.

PR #14670, opened 2023-12-22
Nerfs scout rifle sunder, buffs penetration

ivanmixo:
## About The Pull Request

Halves (roughly) the sunder on penetration/impact, but doubles the penetration.
## Why It's Good For The Game

Currently the scout rifle does 25 sunder (on non impact, it has more) a second, which isn't fightable unless you want to sacrifice all your armor for one guy. Even a req weapon should be something you can fight. Top it off with the fact this weapon is IFF and you can stack it very well with your fellow marines to delete the armor off any xeno that enters your screen, which just isn't fun or fair to fight, even if the ammo is paid. 

Doubling it's armor penetration means this gun is basically built to take down high armor xenos, which means it shouldn't be useless with the sunder cut in half and will still be able to do it's job of taking down large xenos.
## Changelog
:cl:
balance: The basic scout rifle ammo now deals 5 sunder down from 12, but has double the armor penetration (30, up from 20)
balance: The impact scout rifle ammo now deals 6.5 sunder (down from 12.5), but has double the armor penetration (20, up from 10)
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -734,7 +734,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_SUNDERING
 	accurate_range = 15
 	damage = 40
-	penetration = 40
+	penetration = 30
 	sundering = 5
 	bullet_color = COLOR_SOFT_RED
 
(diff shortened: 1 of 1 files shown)


yyzsong:
> down from 12

minor spelling mistake, close this pr

Pariah919:
W/E on normal, good enough change from my POV.
However this kinda nukes high impact, at least leave it at around 8 or 9, its slowdown effect isn't really a good effect considering it also knocks you back on hit unless you're in extremely specific spots where you can troll people, doubling the penetration of 20 to 40 is really good but 10 to 20 is basically not going to do anything to make up for the massive loss of 50% sunder and how armor averages around like, 35-40. So going from 20 to 40 is ignoring all armor while for HI you're barely changing anything.

1kajowsmol:
merge shit shit already scout rifle is mimir gameplay when i just see everyone order it

Lumipharon:
This turns out to be really strong lol.

Lumipharon:
40 Pen is a real buff to this weapon overall, especially considering sunder is % and pen is flat, 40 pen means you effectively negate all/almost all armour on the majority of xenos.

ivanmixo:
It's supposed to be usable still yeah... The base damage isn't overly high where it having so much AP would be gamebreaking

Lumipharon:
The effective DPS is almost the same as a T60, against high armour castes its actually better

PR #14673, opened 2023-12-22
Boiler globules now visible in stat panel

mrhugo13:
## About The Pull Request
Uses carrier's hugger count code in ``carrier.dm`` as seen here:
bild
and repurposes it for globules and adds it to boiler's ``boiler.dm`` file as seen here:
bild

This PR has been fully tested on a local build with zero (0) noticeable runtimes or errors.
dreamseeker_2023-12-22_12-53-20

## Why It's Good For The Game
From what i can tell the only way to see the amount of globules a boiler is carrying is via either how luminous they are which isn't especially exact or noticing the globlue creation icon has changed ever so slightly (to the point where i only noticed it while testing this PR and never before)

This provides another layer of information the boiler can see for the exact amount of globules they are carrying both corrosive and neurotoxin which is displayed alongside the rest of the exact information most xenos get in their stat panel (like how carriers can see the exact amount of huggers they have stored in *their* stat panel).
## Changelog
:cl: Vondiech
qol: Boilers can now see the exact amount of acid/neuro globules they are carrying in their "Status" panel on the right-hand side of the screen alongside with the rest of their xeno stats.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/boiler.dm
@@ -40,6 +40,15 @@
 	set_light_range_power_color(glow, 4, color)
 
 // ***************************************
+// *********** Life overrides
+// ***************************************
+/mob/living/carbon/xenomorph/boiler/get_status_tab_items()
+	. = ..()
+	. += "Stored corrosive acid globules: [corrosive_ammo]"
+	. += "Stored neurotoxic globules: [neuro_ammo]"
+
+
+// ***************************************
 // *********** Init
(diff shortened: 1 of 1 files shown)


Lumipharon (changes requested):
I don't see the issue with the pips, I have no idea how you've never noticed that lol.
But if you have a problem with that and think numbers are better, you need to kill the pips, having both is redundant.

mrhugo13:
> I don't see the issue with the pips, I have no idea how you've never noticed that lol. But if you have a problem with that and think numbers are better, you need to kill the pips, having both is redundant.

I'll let the pips win out then if such is the case

PR #14675, opened 2023-12-22
Removes the magnum from the crash weapons vendor

Barnet2:
## About The Pull Request
What it says on the tin.
## Why It's Good For The Game
When moving the magnum to the SL vendor during it's release, MJP forgot to take it away from the special crash weapons vendor. This seems like an oversight, lest maintainers have their reason for letting it stay in.
## Changelog
:cl:
fix: Fixes the Magnum being available in Crash outside of the SL vendor.
/:cl:


chizzy376:
Hello I'm a mapper and I think it should stay

PR #14689, opened 2023-12-25
Prevents Hypo Injecting Synths And Combat Robots.

Wisemonster:
## About The Pull Request
Prevents injecting synths and combat robots with hypos.
## Why It's Good For The Game
Attempting to inject synths/robots currently causes the hypo to "inject" 0 units. This makes it clearer to new players that robots can't use chems, and is overall less confusing.
## Changelog
:cl:
fix: Synths/Robots no longer can be injected with hypos.
/:cl:

### DIFF
--- a/code/game/objects/items/reagent_containers/hypospray.dm
+++ b/code/game/objects/items/reagent_containers/hypospray.dm
@@ -70,6 +70,9 @@
 	if(!reagents.total_volume)
 		balloon_alert(user, "Hypospray is Empty.")
 		return
+	if(is_species(A, /datum/species/robot) || is_species(A, /datum/species/synthetic) || is_species(A, /datum/species/early_synthetic))
+		A.balloon_alert(user, "Can't inject (robot)")
+		return
 	if(!A.is_injectable() && !ismob(A))
 		A.balloon_alert(user, "Can't fill.")
 		return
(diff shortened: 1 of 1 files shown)


code/game/objects/items/reagent_containers/hypospray.dm
```diff
@@ -70,6 +70,9 @@
 	if(!reagents.total_volume)
 		balloon_alert(user, "Hypospray is Empty.")
 		return
+	if(is_species(A, /datum/species/robot) || is_species(A, /datum/species/synthetic) || is_species(A, /datum/species/early_synthetic))
```
Lumipharon, line 73: Don't make snowflake like this.
These species have traits you can reference instead.

Lumipharon (changes requested):

code/game/objects/items/reagent_containers/hypospray.dm
Wisemonster: Sure, I can use the IS_SYNTHETIC trait for checking synths, though I'm not sure what trait to use for combat robots.

Lumipharon: robots don't have synthetic afaik.
The more useful trait is the 'don't metabolise chems', whatever its actually called.

Wisemonster: Don't dead players recieve that trait though?

Lumipharon: no, they just stop processing

Wisemonster: Alright, it should now use flags to check instead of species datums.

Lumipharon (approved):

PR #14690, opened 2023-12-25
Adds a new Hiveling minion and reworks nymph

ClosetedSkeletonGH:
## About The Pull Request

Adds new Hiveling minion:

sprites:(hiveling resting weed backstate has changed, see comments)

image

gib animation:

hivelingexplode

Stats/Info:

- 150 health, 0 defense of any kind
- +0.5 speed with 0 weed bonus (a bit slower than a boiler, its pretty slow)
- no crit state, it gibs on hitting 0 health/setting on fire, no body is left
- 350 plasma, 35 plasma gain

- cannot melee ala hivemind, however has no demanifestation/teleportation/ranged weeding abilities
- loses 40 health every few (3 or 4 I think) seconds of being off weeds along with a big danger message. 

*note: weedwalker prevents this, and AI/clientless are immune to it as AI have special behavior that doesn't let them move more than a tile off weeds unless theyre using weedwalker)*

abilities:
- creates walls that start at 85 health (3 knife hits on creation) as opposed to non-hivelord castes normal 150. It still gains health like regular walls and at 3 health every 5 seconds til 350. 

*note: walling is its secondary ability, although I may buff its wall strength depending on how the minyun plays*

- new ability: weedwalker, spreads weeds beneath itself as it moves, creating a pod every 5 tiles (4 weedsteps then a pod). costs a decent bit of plasma but allows you to move and spread weeds in a very cool looking way. also this only spreads normal weed and normal pod for balance reason

video: 

(this may need plasma balancing change as this was with plasma cost reduction maybe? also ignore the minor pain sprite error, its since been fixed)

https://github.com/tgstation/TerraGov-Marine-Corps/assets/120620754/1cb325b1-e6df-42b3-8a36-b1967413b775

- they get the ability to place normal/resting/sticky weed pods of they want via a normal weed but their primary weeding will be weedwalker

Nymph changes:

stats:
- 200 health, fire defense removed
- 750 plasma, 25 plasma regen (75% that of a drone)
- melee damage moved from 8 -> 16
- speed from 0 -> -0.2 (this is faster)

ability changes:

- gets a plasma transfer ability (can transfer 200 plasma or so over a slow period if the other xeno stays still)
- gets a heal ability (heals the same amount as drone/queen/hivemind but costs 500 plasma and has a 20 second cooldown)
- gets 1.0x phero strength (this is barely anything, its 2.5 more defense or -0.06 extra speed or a 1% boost to most regens)

**current issues:**

**weedwalker was being activated and used by AI and it was working well and looked dope but for some reason between then and now it just stopped working**

## Why It's Good For The Game

Right now nymph is very :/ to play as, you cant make a difference at all if you try to fight as its pretty much instant death. I am trying to shift nymph from one of our main support roles to a xeno that. while worse than all current xenos + the other minions, can still fight and has a unique role (primarily as a plasma battery/support minion). Nymph already looked like a replete ant so why not have it be an actual replete ant.

I instead relegated the wall placement and weed spreading into something that entirely specializes into spreading weeds into unweeded areas while having placing walls be its secondary ability. It fits the theme pretty well I think and can be a fun thing for someone who wants to help contribute/build a lil bit of maze while in queue or not wanting to fully contribute while also not wanting to take the full responsibility of a builder

## Changelog

:cl:
add: New Hiveling minion specialized in spreading weeds to unweeded areas
add: Changes Nymph into a balanced/support minion with a plasma transfer
/:cl:


Kuro020 (changes requested):
image
very funny, but no

FosterGrime-Heymynameslouie:
the hive is real

aarrisu:
hive ass obliterated

Helg2:
You could change nymph to lesser drone as well will those changes, like it's on cm's for example.

Kuro020:
> image
> 
> Hiveling bubble butt has been removed because of tyranny and unjust abuse of power

It wasnt removed, its still there, remove it completely

FosterGrime-Heymynameslouie:
HIVE ASS NERFED BEFORE IT WAS EVEN ERGHED HOW COULD THIS BE

PR #14693, opened 2023-12-25
Gives beetle a weaker crest defense + rebalances their values a lil 

ClosetedSkeletonGH:
## About The Pull Request

pt 3/4

Right now beetles have a fake crest defense where they lower their crest and move at a lower speed for no actual extra defense. this fixes that.

The values for their crest defense is lower, being a flat +25 bonus to every armor and slowing down the beetle by +1.0 (or the same as moving in walking speed). this is a bit more punishing than defenders +30 armor and +0.8 speed but this is a minion.

Beetles stats were also changed a lil bit, its health was lowered to 250 from 260 and its armor was changed from 30's to 20's to a flat 25 (meaning crest defense doubles its armor to 50)

Beetles can also be given plasma now. they only have 240 or so plasma but there isnt a point to locking out teamwork

## Why It's Good For The Game

Right now the fake crest defense exists for no real reason other than to look silly, confuse new players, and nod. While this won't get the most use for 75% of the situation beetle players are in, this adds a lil bit more strategy and depth to the beetle in how it allows a player to play.

## Changelog

:cl:
add: Beetle can now use a slightly worse version of crest defense
qol: Beetle can be given plasma now
/:cl:


Lumipharon:
What, where's this old pretend crest toggle you're talking about?

Lumipharon:
Where in code? Are you sure that's not literally just toggling walk?

Lumipharon:
Guh, then why is your desc talking about a fake crest defense, and fixing something?

Also this kind of just nerfs AI beetles... which is most of them.

TiviPlus:
iirc ai doesnt use crest so this feels like a nerf

Arturlang:
Just make the AI use crest shield if they're already next to their victim

PR #14695, opened 2023-12-25
Laser pouch can no longer be used as a magazine for Lasguns or Tesla guns.

Pariah919:
## About The Pull Request
Basically title, you can no longer attach lasguns (and also Tesla) to laspouches.
## Why It's Good For The Game
I originally meant for this to also add the ability to charge mags but then I tested it and remembered you can... Already do that.
The ability to attach your pouch to your gun is absurd because it basically works as a super extended mag, which were removed for a reason.
## Changelog
:cl:
balance: You can no longer attach lasguns nor tesla to lasgun pouches for magazine use.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -276,7 +276,7 @@
 	damage_falloff_mult = 0.25
 	fire_delay = 2
 	default_ammo_type = /obj/item/cell/lasgun/lasrifle
-	allowed_ammo_types = list(/obj/item/cell/lasgun/lasrifle, /obj/item/cell/lasgun/volkite/powerpack/marine, /obj/item/cell/lasgun/lasrifle/recharger)
+	allowed_ammo_types = list(/obj/item/cell/lasgun/lasrifle, /obj/item/cell/lasgun/lasrifle/recharger)
 	/// A list of available modes this gun can switch to
 	var/list/datum/lasrifle/mode_list = list()
 	/// The index of the current mode selected, used for non radial mode switches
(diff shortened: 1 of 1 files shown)


Lumipharon:
Whats the point of the pouch if its not linkable to guns? Just carry a pouch of mags.

The whole item is kinda cringe anyway though.

Pariah919:
> Whats the point of the pouch if its not linkable to guns? Just carry a pouch of mags.
> 
> The whole item is kinda cringe anyway though.

Would you rather I just to convert it to basically being a packet?

ButterChugger:
Wasn't the whole balancing aspect of the las pouch is the moment the marine is knocked over (gas, pain causing an unga to collapse, rouny tackle, defender tail whip, blob impact, queen shriek, king roar, vortex, or any other myriad of methods of CC from FF) to cause the gun to completely disconnect from the las pouch, forcing the unga to pull the gun back out, then manually click the gun to the pouch to reconnect, then get his bearings? Considering the average marine has to lug around an ammo box full of them or be in the general vicinity to a roving ammo box (acid bait) and the fact it takes up a full pocket highlights it's own distinct disadvantages. Seems particularly overkill to me.

FosterGrime-Heymynameslouie:
also las pouch is already a pretty shit item in general, its literally just 3 mags + 1 (for the one in the barrel of the gun attached to it) attached to a hose, you could get the same results and better without getting 2 second stunned from trying to reattach the hose back onto the gun if you just tac reload a rifle with a mag pouch, with what couldve been its only saving grace if it was hand crankable but that got shot down because someone couldnt count

Oscari-pog:
If you are gonna do this, you should make it fit in bags instead of being a pouch - because what you are doing is making a packet. And if it will still only fit in a pocket, it'll be useless.

GuyGuysonF19:
Bad PR i don't really have more to say on this one, it's balanced as is, because it's unusable in a push due to any knock down unlinking it, and it works very well at backline fire support, which is balanced by the fact none of the laser guns have aim mode, so they can't be fully utilised anyway. Your effectively pushing the just use MG meta even further forward, shouldn't be merged imo (note, the backline usefulness nullifies the entire PR, because in a backline support role you are able to just have ammo crates filled with energy cells anyway, making the point of this PR effectively mute)

ButterChugger:
Actually wait a moment, aren't you also doing a PR for plasma gun with the same mechanic of attaching it to a gun to a charge pack (albeit on your back) as a form of balance? Doesn't this fall on the same principle?

TellyLsc:
Turns laser sniper even worse than it already is, the shatter/heat ray mode makes you reload too often without powerpack.

1kajowsmol:
another glorious MJP nerf after vali spear to grace us

ivanmixo (approved):

ivanmixo:
> Whats the point of the pouch if its not linkable to guns? Just carry a pouch of mags.
> 
> The whole item is kinda cringe anyway though.

Carries one more mag + acts as a packet of sorts for laser weapons so it's still better than having a mag pouch of laser cells compared to a mag pouch

Yttriums:
Why even keep lasers in the game if you remove elements of their core identity like this? Just get rid of them and we can use barrel charger if we want hitscans.

PR #14704, opened 2023-12-26
More OB cinematics for shipside

Barnet2:
## About The Pull Request
When the OB is fired, an alarm for those not near the OB will play, their screen will shake, and a flavor message will be sent to their chatbox for even more Flavor™
## Why It's Good For The Game
Cinematics are cool.
## Changelog
:cl:
add: Firing the OB will play an alarm for those outside of hearing distance of the cannon and on ship, shake the ship, and send out a flavor message.
/:cl:


Arctrooper8:
Why not make it so the firing sound with the cool alarms can be heard all across the ship?

Barnet2:
> Why not make it so the firing sound with the cool alarms can be heard all across the ship?

That would require splitting up the audio file. Currently, when the OB is fired, you hear the cool alarm and the cannon actually firing apart of one audio file. If I were to play the alarm standalone for the ship, if you were next to the OB you would hear two alarms, alongside the cannon firing/reloading.

Barnet2:
> Why not make it so the firing sound with the cool alarms can be heard all across the ship?

Nevermind I was just dumb, it isn't as cursed as I thought it would be. Thank God for maintainers knowing how to code very well.

ivanmixo (approved):

PR #14716, opened 2023-12-29
Behemoth 2: Now with more caffeine

lKiseki:
# TESTMERGE FIRST I BEG YOU
At least one day of no sleep was dedicated to this PR. Make of that what you will.

## About The Pull Request
Various large-scale changes intended to drastically change Behemoth's playstyle from a slow defensive role to a more offensive role, as well as fixing various bugs and issues.

Starting with bug fixes, sprites, sounds, etc. and leaving balance for last:
- Fixed a bug where Primal Wrath would no longer consume Wrath instead of plasma while active in some cases. This was caused by #14467, specifically this bit, which should've been in the first proc definition instead: https://github.com/tgstation/TerraGov-Marine-Corps/blob/1351a45f33cd4184627301fbec30238df356079f/code/datums/actions/xeno_action.dm#L58-L67
  - Renamed and relocated the relevant signal to account for this change.
- **Fixed an issue where Landslide could not be cancelled sometimes.** Took me long enough, boys, but it's finally responsive and reliable.
- Adjusted ability cooldown messages to no longer include plasma cost. Damn you, action refactor!
- Adjusted various usages of min() and max() to instead use clamp() since it does the same thing but better.
- Changed Seismic Fracture from an activable to a normal ability.
- Refactored the way AoE attacks are done on Behemoth. A lot of things shared Seismic Fracture stuff, so I figured I might as well go the extra length and make it a global proc that's more adjustable to each case.
- Added an icon.
- Removed an unused sound effect.

## The Balance Stuff
Alright, now here's the balance stuff so you can grill me alive.
- Melee damage reduced from 26 to 23. **Bear in mind that all abilities use melee damage to calculate damage, so this is a damage reduction to most of Behemoth's kit.**
- Maximum health reduced from 750 to 650.
  - Maximum wrath reduced from 750 to 650 to account for this change.
- Movement speed increased from -0.2 to -0.5, weed speed modifier increased from -0.1 to -0.2. 

### Rolling (the ability)
- Wind up increased from 1.8s to 2s.
- Speed gain reduced from 0.4 to 0.35 per tile _(this means Behemoth is now slower while rolling, should be nearly the same as before the above movement speed buffs)_.

### Landslide
- Movement speed increased from 1.7 to 1.5. This is actually much faster than it sounds.

### Earth Riser
- Completely revamped the primordial version of this ability! Now it will drop meteors from god knows where straight to your head. After a decent warning, these will hit the ground, dealing damage and a knockdown to those caught in its range. After a set duration, it will also explode as if it had been hit by Seismic Fracture. Watch out, marines...
  - Primordial version has increased range and a reduced wind up, by the way.
- Reduced plasma cost from 30 to 20.
- Reduced cooldown from 15s to 10s. 

### Seismic Fracture
- Seismic Fracture is no longer a ranged ability, but rather an AoE around the Behemoth.
- Increased attack radius from 2 to 3.

## Why It's Good For The Game
I admittedly low balled Behemoth when it comes to balance, and it is unable to keep up with marines due to how slow it is by comparison. As such, I rebalanced it to hit less hard but be much faster, therefore pushing it into a more offensive frontline role rather than a slow defensive dude who's unsure whether to go hit you or defend. This should better define it in terms of playstyle, and give it that edge it needs to actually keep up with marines.

## Changelog
:cl: Lewdcifer
add: Behemoth 2: Now with more caffeine! Check the relevant PR for the full detailed changelog.
add: Completely revamped Earth Riser's primo. Meteors rain from above!
balance: Behemoth melee damage reduced from 26 to 23. This also affects all ability damage.
balance: Maximum health reduced from 750 to 650, maximum wrath adjusted accordingly.
balance: Movement speed increased from -0.2 to -0.5, weed speed modifier increased from -0.1 to -0.2.
balance: Behemoth roll wind up increased from 1.8s to 2s.
balance: Behemoth roll speed gain reduced from 0.4 to 0.35.
balance: Landslide movement speed increased from 1.7 to 1.5. This is much faster than it sounds!
balance: Earth Riser plasma cost reduced from 30 to 20.
balance: Earth Riser cooldown reduced from 15s to 10s.
balance: Seismic Fracture is no longer ranged, but rather an AoE around the Behemoth.
balance: Seismic Fracture attack radius increased from 2 to 3.
fix: Fixed a bug where Primal Wrath would no longer replace plasma costs with wrath.
fix: Fixed a bug where Landslide couldn't be cancelled sometimes. Now it's responsive and reliable.
spellcheck: Behemoth cooldown messages no longer include the ability's plasma cost.
sounddel: Removed an unused Behemoth sound file.
imageadd: Added an icon for the primordial version of Earth Riser
refactor: Refactored Behemoth AoE attacks.
/:cl:


Kuro020:
FUCK IT, WE BALL

Kuro020 (approved):

Arturlang:
I reccomend TM-ing this so you can still make some changes once you see how balance works out.

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
```diff
@@ -284,24 +266,29 @@
...
 		return TRUE
 	return ..()
 
 /datum/action/ability/activable/xeno/landslide/on_cooldown_finish()
-	current_charges = min(maximum_charges, current_charges + 1)
+	if(ability_active)
+		owner.balloon_alert(owner, "Use [initial(name)] again to cancel")
+		return ..()
+	current_charges = clamp(current_charges+1, 0, maximum_charges)
+	owner.balloon_alert(owner, "[initial(name)] ready[current_charges > 1? " ([current_charges]/[maximum_charges])" : ""]")
```
Lumipharon, line 281:
```suggestion
	owner.balloon_alert(owner, "[initial(name)] ready[current_charges > 1 ? " ([current_charges]/[maximum_charges])" : ""]")
```

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
```diff
@@ -606,7 +590,7 @@
 	var/list/obj/structure/earth_pillar/active_pillars = list()
 
 /datum/action/ability/activable/xeno/earth_riser/on_cooldown_finish()
-	owner.balloon_alert(owner, "[name] ready ([length(active_pillars)]/[maximum_pillars])")
+	owner.balloon_alert(owner, "[initial(name)] ready[maximum_pillars > 1? " ([length(active_pillars)]/[maximum_pillars])" : ""]")
```
Lumipharon, line 593:
```suggestion
	owner.balloon_alert(owner, "[initial(name)] ready[maximum_pillars > 1 ? " ([length(active_pillars)]/[maximum_pillars])" : ""]")
```

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
```diff
@@ -648,32 +632,41 @@
 		return
 	var/turf/owner_turf = get_turf(owner)
 	var/turf/target_turf = get_turf(target)
-	if(!line_of_sight(owner, target, EARTH_RISER_RANGE) || LinkBlocked(owner_turf, target_turf))
+	var/mob/living/carbon/xenomorph/xeno_owner = owner
+	var/datum/action/ability/xeno_action/primal_wrath/primal_wrath_action = xeno_owner.actions_by_path[/datum/action/ability/xeno_action/primal_wrath]
+	if(!line_of_sight(owner, target, primal_wrath_action?.ability_active? EARTH_RISER_ENHANCED_RANGE : EARTH_RISER_RANGE) || LinkBlocked(owner_turf, target_turf))
```
Lumipharon, line 637: I know this was already here, but LinkBlocked is for checking adjacent turfs, not ranged turfs.

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
```diff
@@ -759,37 +747,32 @@
 			pixel_x += 18
 			pixel_y -= 8
 
-/datum/action/ability/activable/xeno/seismic_fracture
+/datum/action/ability/xeno_action/seismic_fracture
```
Lumipharon, line 750: why's this no longer an activatable?

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
```diff
@@ -1443,3 +1370,77 @@
...
+// ***************************************
+#define AREA_ATTACK_DAMAGE_MECHA_MODIFIER 10
+
+/**
+ * Checks for any atoms caught in the attack's range, and applies several effects based on the atom's type.
+ * * turfs_to_attack: The turfs affected by this proc.
+ * * enhanced: Whether this is enhanced or not.
+ * * instant: Whether this is done instantly or not.
+*/
+/proc/behemoth_area_attack(owner, list/turf/affected_turfs, enhanced, damage_multiplier = 1.5, paralyze_duration = 0.8 SECONDS)
```
Lumipharon, line 1386: why does this need to be a global proc?

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
```diff
@@ -1443,3 +1370,77 @@
...
+					var/obj/vehicle/sealed/mecha/affected_mecha = affected_atom
+					affected_mecha.take_damage(attack_damage * AREA_ATTACK_DAMAGE_MECHA_MODIFIER, MELEE)
+					continue
+				if(istype(affected_atom, /obj/structure/reagent_dispensers/fueltank))
+					var/obj/structure/reagent_dispensers/fueltank/affected_tank = affected_atom
+					affected_tank.explode()
+					continue
+
+/// Warns nearby players, in any way or form, of the incoming ability and the range it will affect.
+/proc/do_warning(owner, list/turf/target_turfs, duration)
```
Lumipharon, line 1433: ditto. I know it was already a global but I don't see why, exactly

Lumipharon (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
lKiseki: Because the ability was changed to no longer be targetable. It's now an AoE around the Behemoth, so it might as well just not be an activable anymore.

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
lKiseki: Because this is used across various abilities. I decided to make it a global instead since it was easier to adjust for each case and saved me a lot of ability references.

code/modules/mob/living/carbon/xenomorph/castes/behemoth/abilities_behemoth.dm
lKiseki: Ditto. This proc exclusively was a suggestion from Tivi back when Behemoth was first added.

Arturlang:
Not sure if behemoth will be able to offensive if it cant deal much damage, it already struggles to slash marines due to how big it is

Arturlang:
Also consider making client moving canceling the charge. More intuitive that way

Lumipharon:
ok code seems fine, I unno about bemo balance so we'll just tm and see what happens.

Carnage068:
I appreciate the effort that went into this PR but I just can't agree with it. The loss of 100 health and 3 damage on virtually all of its abilities (which REALLY adds up if you get a successful combo off) just.. Doesn't seem to make up for the increased movement speed.
Not to mention, why was rolling nerfed? Before you had the option of saving your roll in case you were being rushed, but now that you have less health and you're faster (not to mention the windup was increased for some reason?) there's virtually no reason to even consider it. Less decision making, less nuance.
Also, I don't know if I'm the only one, but I miss the ranged Seismic Fracture. It was nice being able to predict where people would move and place it accordingly, but now that it always happens in an AoE around yourself I just.. I don't know. Not to mention, with the lower health pool and nerfing Roll as an escape, getting yourself surrounded in order to optimally use Seismic Crush just feels way too damn risky.

I think you tried to keep the changes too balance-neutral for a caste that was already having issues performing in the first place. I do not think this is what Behemoth needs. Maybe with some tweaks and additional changes, but as things are right now it just feels wrong to me.

Carnage068:
Okay, I've actually went and played a round of it, now. It feels okay-ish, but like, and the Seismic Fracture change is kind of a tradeoff. It's great that you can use it during Landslide, but it feels significantly worse to combo with Rising Earth, and I still miss being able to aim it. I wish there was a way you could have both the aimable, smaller version, and the larger version around yourself that this PR implemented.
The nerfs to roll windup, damage, and health still aren't necessary, imo. It was never the tankiest xeno or the xeno that did the most damage.

imsxz:
havent gotten the chance to play it outside of valhalla but i feel like the speed will make it way easier to play aggressively even with the HP nerf

lKiseki:
Found a bug where a certain Behemoth AoE is bugged. Do not merge yet.

novaepee:
two behemoths are fun to fight against; they are actually a threat in the field assuming they can hold up against the wall of aim mode mg-60

lKiseki:
AoE bug apparently persists for god forsaken reasons. Do not merge.

lKiseki:
Bug no longer present, tested locally. Balance still pending.

shelteredfreak:
Feedback, or more like suggestions
1: Make it so landsliding into a rock stops landslide, mostly I don't end up charging into a crowd of marines on accident.
2: Maybe, as a feature, make destroying your rock pillar while it is being flung (by charge or manual throw) explode like it impacted mid-air. As a bit of a skill requirement for bonus use.
3: Let rocks go over unwired cades. Because like... Not wiring cades deserves punishment....

PR #14747, opened 2024-01-03
Click Code cleanup

Xander3359:
## About The Pull Request
DO NOT MERGE, Please only test-merge since I'm not absolutely sure this won't break clicks in some way.
Makes signals_click.dm
Adds attack_self_alternate() proc + Adds it to autoinjectors updates old signals to the new ones
Cleans up click code on ClickOn() for mob, AI and observer
## Why It's Good For The Game
Code improvement
## Changelog
:cl:
refactor: Click code is now somewhat cleaner and more standardized
code: Adds attack_self_alternate(), this gets called when an item in-hand is right-clicked
fix: You can now right click any hypospray in-hand to change the injection size 
/:cl:


vvvv-vvvv (changes requested):
why are you changing the macros for the signals? why differ from what tg has

Xander3359:
> why are you changing the macros for the signals? why differ from what tg has

Why do we care what TG has?
I think these signals are more descriptive

vvvv-vvvv:
> Why do we care what TG has?

because we are basically a tg downstream and snowflaking something as small as this will make diffs harder to cherrypick

> We already don't have code parity

signals are pretty much one-to-one right now. If you think these new names are more descriptive, just PR it upstream please

Xander3359:
> > Why do we care what TG has?
> 
> because we are basically a tg downstream and snowflaking something as small as this will make diffs harder to cherrypick
> 
> > We already don't have code parity
> 
> signals are pretty much one-to-one right now. If you think these new names are more descriptive, just PR it upstream please

We have never been a TG downstream.
PRing these signals to TG would not in any way change how we operate here.
Nor do I really care about the TG codebase outside of porting some stuff from there.

TiviPlus (approved):

PR #14751, opened 2024-01-04
Adds gun flags to laser gun fire modes, carbine spread can no longer be fired one handed.

Pariah919:
## About The Pull Request
Title, literally.
Carbine spread requires full wield to be fired one handed.
## Why It's Good For The Game
Consistent with other 'spread' style guns. At least with combat shotgun you're dealing with shotgun level spread and such.
## Changelog
:cl:
balance: Lasercarbine spread now requires two hands to be fired like a normal shotgun.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -307,6 +307,8 @@
 	var/windup_delay = 0
 	///codex description
 	var/description = ""
+	///Used to change gun flag features per firemode.
+	var/flags_gun_features
 
 /obj/item/weapon/gun/energy/lasgun/lasrifle/unique_action(mob/user)
 	if(!user)
@@ -344,6 +346,7 @@
 	fire_sound = initial(choice.fire_sound)
 	rounds_per_shot = initial(choice.rounds_per_shot)
 	windup_delay = initial(choice.windup_delay)
+	flags_gun_features = initial(choice.flags_gun_features)
 	SEND_SIGNAL(src, COMSIG_GUN_BURST_SHOTS_TO_FIRE_MODIFIED, burst_amount)
 	SEND_SIGNAL(src, COMSIG_GUN_AUTOFIREDELAY_MODIFIED, fire_delay)
 	SEND_SIGNAL(src, COMSIG_GUN_FIRE_MODE_TOGGLE, initial(choice.fire_mode), user.client)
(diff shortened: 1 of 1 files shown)


usnpeepoo:
Speed merge this

Oscari-pog:
With combat shotty you are also dealing a nerf foam dart amount of damage, while the lascarb deals comical damage

PineappleGuy42:
> With combat shotty you are also dealing a nerf foam dart amount of damage, while the lascarb deals comical damage

I don't mean to be rude, but you're pretty wrong. PBing with Combat Shotty one handed is viable and for the most part (better) than the lasbine. The only real drawback is the recoil but that's not truly a factor due to the recoil having ended by the time you are able to fire the next shot. I feel the lasbine's hate is misplaced really as combat shotty can stun and stagger and has more damage and since lasbine is being used by a majority of people it's being overly targeted.

(I use buckshot though I understand slugs are a favored round I, myself typically only use buckshot and feel as if it'd make the best comparison)

https://github.com/tgstation/TerraGov-Marine-Corps/assets/56173966/0ff3ea94-65c9-476a-98c7-1ec786a3c723

The counter argument
--------------------------
https://github.com/tgstation/TerraGov-Marine-Corps/assets/56173966/b9758763-2a72-4c7f-8756-33eef164aa93

Now I do understand that at range the lasbine is stronger, but typically in the average battle 7/10 times I will be used within PB or a tile away from a xeno and is rarely a chase weapon as other guns are much more effective in that role. (Skirmish rifle, Carbine.)
I feel you could address this issue with damage falloff or more inaccuracy but forcing a already mid-high tier weapon to become a worse SH-35 is not the way in this player's opinion.

blackdav123:
the issue with the lasbine is that it is a melee weapon that deals claymore damage while not having the drawbacks of high slowdown and a long time to hit, and being able to be refired after a stun much faster because of the onehandedness

while the sh35 does more damage it is also a worse melee weapon because it deals knockback which makes running down low health xenos much harder

Lumipharon:
Whats the issue, exactly?
Spreadmode does extremely similar dps to flechette, but flechette does something like x5 more sunder/second.

Notamaniac:
> Whats the issue, exactly? Spreadmode does extremely similar dps to flechette, but flechette does something like x5 more sunder/second.

The issue is that being able to fire one handed allows you to ignore the wield delay and shoot xenoes the instant you get unstunned from an attack (i.e. being hugged, downed from a queen shriek, shrike fling,  pounce ect.)

With shotguns this is annoying for runners, but just about tolerable for other xenoes, who can feasibly survive a PB shot and get away out of range for the second shot to deal not too much damage (especially if the buckshot knocks them away a bit)- for the las carbine it maintains it's damage over medium range (and has no knockback, meaning the follow up shot does more damage as the xeno is closer). This can lead to very quick unexpected deaths when you say, knock a marine down- head to slash them, eat a PB lascarbine shot the instant they get up, try to get away but eat a second shot, then potentially get chased down by follow up shots or shot down by other marines.

Though given the reactions I geninuely don't know if this is a skill issue or not

ivanmixo:
id rather make the spread mode like... actually spread a significant amount to the point it doesnt just hit all of its shots even at a relatively far distance

Pariah919:
> id rather make the spread mode like... actually spread a significant amount to the point it doesnt just hit all of its shots even at a relatively far distance
eh fair

yyzsong:
Coward

PR #14302, opened 2023-10-27
Water decreases smoke lifespan

Redory:
## About The Pull Request

Water splashes will now decrease the current lifespan of smoke effects by 33%.
The smaller the lifespan of the smoke the less effect water will have, since it just shaves off 33% off the current lifespan.

Hydro cannon can now be attached to the following weapons:
- TE Laser Rifle
- TE Laser Carbine
- AR-12
- AR-21
- MG-42
- SH-15
- AR-55

Hydro cannon water tanks now only hold 125 water instead of 200, but are now smaller in size.

(I don't know what happened that made the videos say they are 321 hours long.
I just downloaded OBS and converted the .mkv files into .webm using VLC)

[2023-10-27 18-59-03.webm](https://github.com/tgstation/TerraGov-Marine-Corps/assets/41807890/8b977113-2db6-4886-971a-bf66aa309ea3)

[2023-10-27 19-01-45.webm](https://github.com/tgstation/TerraGov-Marine-Corps/assets/41807890/97b18328-3036-42b3-928c-ec55a32596a8)

Just a random thought i had today and made with like 1.5 hours of work.
## Why It's Good For The Game

I think water could become somewhat more of a support role in dealing with excessive amounts of smoke spam.
I know Mimir exists, but i thought water clearing smoke could become an interesting concept.
I tweaked the water amount and gave it only to specific weapons, so i don't think it's gonna change the balance all too much.

But, hey, at least a couple more muhreens to extinguish you exist when you inevitably run into your own incendiary grenade.

In the videos above i used 4 spray splashes of water and used up A LOT from the water tank, just to clear a small patch of smoke from the acid grenade quicker.
## Changelog
:cl:
balance: Water splashes will now decrease smoke lifespan by 33%.
balance: Hydro water canisters now hold 125 water instead of 200.
balance: Hydro water canisters are now small size instead of normal size.
balance: Hydro cannon can now be attached to: TE Laser Rifle, TE Laser Carbine, AR-12, AR-21, MG-42, SH-15 and AR-55.
/:cl:

PR #14763, opened 2024-01-06
Tanglefoot no longer purges gas, adds a new gas type that purges certain gas types, adds a pepperball ammo for it.

Pariah919:
## About The Pull Request
- Tangle no longer purges gas.
- New gas type, currently only accessible within pepperball guns. It purges Tanglefoot, Xeno gasses, and Sataprine.
- Pepperball ammo for the new gas type. PAC Balls.
## Why It's Good For The Game
Adds a more interesting dynamic to the gas purge difecta, also encourages more aggression in the forms of clearing gas that you can do.
## Changelog
:cl:
add: Added new gas that purges xeno gasses, tanglefoot and satrapine. It is avaliable for use in the pepperball gun as 'PAC Balls'
balance: Removed tanglefoots ability to purge gas.
/:cl:


Kuro020 (approved):

FosterGrime-Heymynameslouie:
an acutal use for the pepper ballers beyond annoying for 3 seconds fellow pee ballers rejoice

code/modules/projectiles/ammo_datums.dm
```diff
@@ -4085,6 +4085,35 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
...
+
+/datum/ammo/bullet/pepperball/purger
+	name = "bursting pepperball"
+	damage = 30
+	flags_ammo_behavior = AMMO_BALLISTIC|AMMO_EXPLOSIVE
+	plasma_drain = 15
+	drain_multiplier = 0
+	/// Smoke type created when projectile detonates.
+	var/datum/effect_system/smoke_spread/smoketype = /datum/effect_system/smoke_spread/purge_gas
+	/// Radius this smoke will encompass on detonation.
```
ivanmixo, line 4097: you forgor

ivanmixo (approved):
I'd be down to try getting rid of root if this turns out well

chizzy376:
But then what's the point of using satrapine if tangle keeps the main part of itself? (Purging plasma)

Loafin34:
i would rather have a dedicated PAC grenade than this. being able to counteract a boiler's gas is important for holding barricades, and having it be limited to a PAC pepperball feels gimmicky at best and a soft-remove at worse. what's the radius like for the PAC balls boiler neutralization zone?

Lumipharon:
This takes gas countering ability away from all the support mechanics and instead gives it to pepperball guns? wuh? why?

DaTruRiZZoB:
good change, though t-foot nades are already removed from marine access, why not just keep the gas purge of t-foot whilst including this? a variety of options is better than 1 singular answer. (plus it can probably be balanced by removing root instead)

ivanmixo:
> This takes gas countering ability away from all the support mechanics and instead gives it to pepperball guns? wuh? why?

What support mechanics (please tm this lumlum)

Lumipharon:
> > This takes gas countering ability away from all the support mechanics and instead gives it to pepperball guns? wuh? why?
> 
> What support mechanics (please tm this lumlum)

Every source of tfoot that presently exists (mortars/cas/ob/rockets). Sure, you could just add new ammo types for those systems for the new antigas gas... but whats the actual reason to split up the support abilities of plasma drain and gas purge into two different types? Tfoot doing both just makes it more generally useful, while gas purge alone isn't particularly useful on any of those delivery systems due to the delay in application.

You can tm if you want though, but I think its a bad idea.

shelteredfreak:
Finally, not getting locked out of Defiler gameplay simply because Req ordered one box of TFoot nades and then marines just cover every single push with TFoot so I can't gas.

Pariah919:
Welp. An attempt was made.

Kuro020:
I think this pr can be interesting, I want to TM it and see how it goes

QualityVan:
My concerns about this are basically a mirror of the reason #14302 got closed. It seems like once this gun's on the field xeno gas becomes pretty much permanently unusable forever, and unlike normal pepperball it's easy enough to just swap between this and a separate main gun when you need to clear smoke.
OTOH, I actually don't remember how limited tangle supply is these days. It was cut down a bunch, I think?

Kuro020:
> My concerns about this are basically a mirror of the reason #14302 got closed. It seems like once this gun's on the field xeno gas becomes pretty much permanently unusable forever, and unlike normal pepperball it's easy enough to just swap between this and a separate main gun when you need to clear smoke. OTOH, I actually don't remember how limited tangle supply is these days. It was cut down a bunch, I think?

valid concerns but then I think TM'ing it and seeing how it plays out is a good idea, if its whack mjp can fix it, if he cant fix it we can kill the pr for good

FosterGrime-Heymynameslouie:
really hope this gets merghed in some way

dont really have any more gas removal tools now, considering tgas got banished to req hell and preplaced smoke no longer stops other gas kinds from spreading which kinda gives assault gas free reign against a marine blob with the exception of people who use mimir2 

having some form, albeit weak counterpart to tgas clearing in pepperball seems like an okay compromise, plus gives pepperball a viable use now!

I think (this also reminds me of the first itteration of pepperball)

Wisemonster:
While you're at it, you could consider adding the PAC ammo type to the underbarrel pepperball gun.

usnpeepoo:
this is seriously way too strong in gas clearing, with the req price being 5 points a mag any long running round can end up with gas castes just getting cucked out of their abilities repeatedly since one person running this deletes any gas near marines.

code/modules/reqs/supplypacks.dm
```diff
@@ -796,6 +796,11 @@ WEAPONS
 	contains = list(/obj/item/ammo_magazine/pistol/standard_pistol/smart_pistol)
 	cost = 10
 
+/datum/supply_packs/weapons/pepperball_purger_ammo
```
ivanmixo, line 799: Why did you give it import gun level ammo pricing? The 20 you get roundstart should realistically be enough for the whole round, but this just makes it dirt cheap to restock on top

code/modules/projectiles/ammo_datums.dm
ivanmixo: you still forgor to remove this random autodoc that isnt doccing anything

ivanmixo (changes requested):

Lumipharon:
This is extremely strong, dirt cheap, and highly targetable/instantly castable. This is hugely better than tangle in every practical way.

Wisemonster:
Yea, currently a single pepperball user can just straight up shut down any kind of gas the xenos use. If there's multiple, then the xenos just don't get to use gas for the round.

Pariah919:
don't agree with making it a greande, so get someone else to do it.

PR #14766, opened 2024-01-06
Fixes missing area on big red office.

Helg2:
## About The Pull Request
Fixes this:
image
## Changelog
:cl:
fix: [Big Red] One of the office variants now doesn't have a hole in the roof.
/:cl:


Lumipharon:
does the area that tile is currently in have another apc? this seems like an intentional (if somewhat dumb) workaround to have an apc for outside be... inside.

Barnet2:
This was intentional by Grayson to have the outside lights be controlled by an indoor APC. I would recommend you talk to him on Discord about it, since there are multiple of these, not just the office one.

Lumipharon:
The current solution is also frankly not good, and is very hacky. There should really be a coded solution to hard set an apc to a particular area thats its not physically in.

Helg2:
Oh well, kek

PR #14757, opened 2024-01-04
Hud icon and overlays cleanup

ivanmixo:
## About The Pull Request

Title. A bunch of this was shitcode and icon states/overlays were handled outside of the relevant procs.

This should fix stamina being invisible when you first spawn in (non existent icon state as the default lol), also some alerts being named weirdly.
## Why It's Good For The Game

This is less shitcode. 
## Changelog
:cl:
refactor: Cleaned up hud icons and overlays
/:cl:

PR #14767, opened 2024-01-07
Port update_appearance, alternative_appearance and ssvisoverlays

ivanmixo:
## About The Pull Request

Yeah. Unshitcodes a bunch of our `update_icon()` procs and makes them handle overlays properly. 

This will need a lot of testing because it touches so much and has potential to break a lot, and I can't possibly test everything myself.

Requires (and includes) https://github.com/tgstation/TerraGov-Marine-Corps/pull/14757 so either merge or close that and TM this instead, please.
## Why It's Good For The Game

This is less shitcode and less shitcode good.
## Changelog
:cl:
code: Ported update_appearance, alternative_appearance and ssvisoverlays
refactor: Refactored almost every instance of update_icon, update_icon_state and update_overlays
/:cl:


ivanmixo:
Need to see about xeno overlays

ivanmixo:
nvm byond moment apparently

code/game/atoms/atom_appearance.dm
```diff
@@ -0,0 +1,112 @@
...
+					managed_overlays = new_overlays
+
+		. |= UPDATE_OVERLAYS
+
+	. |= SEND_SIGNAL(src, COMSIG_ATOM_UPDATED_ICON, updates, .)
+
+/// Updates the icon state of the atom
+/atom/proc/update_icon_state()
+	SHOULD_CALL_PARENT(TRUE)
+	return SEND_SIGNAL(src, COMSIG_ATOM_UPDATE_ICON_STATE)
```
Lumipharon, line 106: While it (probably) doesn't matter so much in practice, is there really a reason to have this require call to parent?
update_icon_state should only ever be called by update_icon, so if you just send the signal in update_icon before calling this proc, we don't need to call parent here.

I only say this because if you have some parent doing whatever checks for its update_icon_state, just to be entirely overridden by a child that does something completely different, it seems extremely redundant.

TiviPlus: it has a signal so you have to

Lumipharon: I think Ivan already mentioned it, but I was saying to just have the signal sent in update_icon (or update_appearance for name/desc). Those SHOULD always call parent, where as update icon state/name etc have zero reason to, once the signal is pulled out. They're only ever called by update_appearance, so the signals will always be sent regardless.

Doing it this way avoids needless code running from parent calls setting stuff that just gets overridden anyway. 
Only update_overlays specifically needs the signal in it.

PR #14772, opened 2024-01-07
Reduces spray flamer nozzle's cone range

lKiseki:
## About The Pull Request
Per title. Before and after shown below:
image

## Why It's Good For The Game
This thing's range is way too oppressive and covers a lot for what it can do. I believe this is a reasonable way of nerfing it.

## Changelog
:cl: Lewdcifer
balance: Spray flamer nozzle's cone range reduced.
/:cl:


Lumipharon:
What is even the point at that rate?

lKiseki:
ok i tried

PR #14775, opened 2024-01-07
Melee special attacks

Lumipharon:
## About The Pull Request
Atomised change so Tivi doesn't burst a vein later.

Adds weapon skills.
A type of activatable ability associated with a melee weapon.
These skills cost stamina to use, and are gated behind having a particular trait (which will be available in campaign in the future)

Adds a sweep attack to the breaching axe, very similar to ravage. Can't stun, and missing anyone lying down (you're swinging at chest level).
Adds a lunging attack for the machete (or more specifically, any kind of sword), a very short dash attack like pounce/charge. Can't stun.

Abilities now use signals for determining what is selected, so they align with jetpack and blink drive. On that note I realised blinkdrive sigs are wack, but I'll make a separate pr to just have jetpack/blinkdrive use ability actions so they're all consistant.

Also fixed a bug where I forgot to unregister a signal on a component axe has.
## Why It's Good For The Game
special attack cool, atomised change gud.
## Changelog
:cl:
balance: Adds weapon skill functionality, for a future campaign release
/:cl:


Lumipharon:
Draft for now as I need to review blinkdrive/jetpack to make them more compatible.

TiviPlus (changes requested):
No reason this cant be rclick/normal click functionality

Lumipharon:
These need to be action buttons.
If there is more than one, right click is no longer going to cut it, and resorting to random alt + click combos or whatever is extremely cringe, and ultimately limited by those param combos that are already used for other things, while also providing no visibility on what abilities you have available, cooldowns, cost, etc.

No xeno players would want all their actions to be converted in the same way, and having these as actions keeps abilities consistant.

ivanmixo (approved):

PR #14787, opened 2024-01-09
Removes hatchling rank, adds prime

Barnet2:
## About The Pull Request
Hatchling rank is no more. It now goes Young > Mature > Elder > Ancient > Prime
## Why It's Good For The Game
The amount of mental anguish I received when I saw an ancient titled xeno with a prime rank icon was very immense. In addition, this will make it so the rank icons line up with how they were before maturity removal.
This should be the last change I have to make to the system. I can live with "Primordial Prime".
## Changelog
:cl:
spellcheck: Hatchling xeno rank is replaced with young, prime is now the top xeno rank
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/king/king.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/king/king.dm
@@ -30,17 +30,17 @@
 	var/prefix = (hive.prefix || xeno_caste.upgrade_name) ? "[hive.prefix][xeno_caste.upgrade_name] " : ""
 	switch(playtime_mins)
 		if(0 to 600)
-			name = prefix + "Hatchling King ([nicknumber])"
-		if(601 to 1500)
 			name = prefix + "Young King ([nicknumber])"
+		if(601 to 1500)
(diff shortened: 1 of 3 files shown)


Dikozda:
Primordial Prime Puppeteer would sound a bit sily

JoeHammad1844:
> Primordial Prime Puppeteer would sound a bit sily

silly is bad even though we are playing spess shooty shoot game

Lumipharon:
Yeah just from a simple clarity point of view "primordial prime" is uh, not ideal.

Barnet2:
> Yeah just from a simple clarity point of view "primordial prime" is uh, not ideal.

I would change the name of Primo itself to Apex, which I tried doing but it was denied. Though primo is the last remnant of "maturity", the xeno rank names took the place of it. I am completely welcome to change it to Apex if the maints agree on it, but Tivi and Ivan said this PR was fine given Prime has a different meaning than Primordial. Will take some getting used to, probably - but it is a fair complaint.

Al-1ce:
Is it still scuffed where if you already have 400 hours in "Xenomorph" it still will label you as young?

PR #14794, opened 2024-01-10
Harvester bicard fix

Lumipharon:
## About The Pull Request
Bicard no longer magically heals dead bodies, or robots.
## Why It's Good For The Game
Exploit bad.
## Changelog
:cl:
fix: Harvester bicard no longer heals dead bodies or robots
/:cl:

### DIFF
--- a/code/datums/components/harvester.dm
+++ b/code/datums/components/harvester.dm
@@ -261,6 +261,10 @@
 		to_chat(user, span_rose("[target] is already dead."))
 		return
 
+	if((target.species.species_flags & NO_CHEM_METABOLIZATION))
+		to_chat(user, span_rose("[target] Cannot process chemicals."))
+		return
+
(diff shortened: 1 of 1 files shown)


Xander3359:
Not healing corpses would be a balance change.
It was intended to keep that feature.

DeltaFire15:
Healing dead bodies is a feature of bicard harvester and not a bug. Do not label it as such if you want to remove it.

For reference, the heal of a defib at med skill 3 is up to 36; 12 of each brute, burn and tox, while bicard heals 25 of brute only. Time-wise, defib is 7 seconds, bicard is ~5. At med skill 0, defib heals 24 total (8 of each), while bicard heals 10 of brute only.
The tradeoff is that harvester has a tendency to be particularely bulky to carry, unless you are using the knife that for some reason exists, in which case you have to carry a bunch of reagent storage instead because it only carries 5u of chems.

Bicard also already doesn't heal robots, it just wastes the charge if you try. Now it gives you a message, but also still wastes the charge.

PR #14797, opened 2024-01-11
Research Outpost is no longer votable if the gamemode is Nuclear War.

Barnet2:
## About The Pull Request
What it says on the tin. If the gamemode selected at the time (or in vote) is Crash, it can still be played. The only way it can be played is if someone votes Nuclear War in the lobby; which shouldn't usually happen, as pop drops most of the time after a vote.
## Why It's Good For The Game
This map is the worst of the worst when it comes to Nuclear War. The other crash maps (Icy Caves, Orion, Desparity) aren't great either, but somewhat(?) work for lowpop NW.
## Changelog
:cl:
del: Research Outpost is no longer votable when Nuclear War is selected.
/:cl:

### DIFF
--- a/code/__DEFINES/__game.dm
+++ b/code/__DEFINES/__game.dm
@@ -14,6 +14,9 @@
 #define MAP_GELIDA_IV "Gelida IV"
 #define MAP_DELTA_STATION "Delta Station"
 #define MAP_OSCAR_OUTPOST "Oscar Outpost"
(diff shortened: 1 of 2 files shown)


RipGrayson (approved):
Research Outpost bad

chizzy376:
9 lifesavers in research outpost medbay stay winning

PR #14806, opened 2024-01-12
move Big Red's ETA APC closer to disk

novaepee:
## About The Pull Request

Remove these two APCs in Big Red's ETA.

image

image

Replace those two APCs in Big Red's ETA to an APC closer to disk.

image

## Why It's Good For The Game

Big Red's APCs are known to be visually near disks so that engineers can keep their sanity when dealing with APC.

This is the only disk I know in Big Red that force marines to be far from APC.
image

## Changelog

:cl:
balance: move Big Red's ETA APC closer to disk
/:cl:


Lumipharon:
just build a new one 4head

chizzy376:
What do you mean far away? It's 8 tiles away, within screen view. I don't think this counts as "far away"

Helg2:
You can just blow up the wall tho?

ivanmixo:
Making a map where the marines already have a very easy time if they commit to disks even easier is very meh in my opinion, especially considering this disk has a high likelyhood of being unflankable due to the way big red modulars work.

Yttriums:
doesn't this also double the area you power by getting the one APC? thats real bad

PR #14816, opened 2024-01-14
Reorganizes xeno structures

TiviPlus:
Each one in its own file instead of lumped into one


ZeWaka:
you should delete the old file?

Lumipharon (approved):

PR #14822, opened 2024-01-15
Bunch of weird runtimes that broke one round then never happened again

ivanmixo:
## About The Pull Request

- Runtime related to target missing in the attack log message

- Runtime related to trying to display messages to clientless ghosts, causing dchat to not work
- Another runtime related to trying to display comms messages to ghosts, causing radio to not work
- Runtime related to a mob somehow calling `keyLoop` without a client being passed in as user
- An arg I forgot in the orphan hivemind onscreen countdown
- Some remote eyes not playing nice with the logic in the sound transferring to the user (because their `parent_cameranet` var gets set to a non `/datum/cameranet` for whatever reason, I've opted to just not let them hear at all until this code is refactored to have one unified var that determines the hud user)
## Why It's Good For The Game
runtime dad
## Changelog
:cl:
fix: Various runtime fixes, see the PR
/:cl:

### DIFF
--- a/code/__HELPERS/logging/attack.dm
+++ b/code/__HELPERS/logging/attack.dm
@@ -33,11 +33,11 @@
 	var/postfix = "[sobject][saddition][hp]"
 
 	var/message = "[what_done] [starget][postfix]"
-	user.log_message(message, LOG_ATTACK, color="red")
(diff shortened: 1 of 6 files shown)


Lumipharon (approved):
Going to tm this now since chat completely eating shit 4noraisin is really annoying.

PR #14826, opened 2024-01-15
Fix flamer back tank disconnecting with magharness

TiviPlus:
## About The Pull Request

Note: also removes .3 second delay for magharness. Since you get stunned anyway this shouldn't do anything?

## Why It's Good For The Game
Best case yay back tank worst case we find out why the delay is there and I redo this

## Changelog
:cl:
del: Removed split second delay on magharness
fix: Fix flamer back tank disconnecting with magharness
/:cl:

### DIFF
--- a/code/datums/components/reequip.dm
+++ b/code/datums/components/reequip.dm
@@ -5,10 +5,8 @@
 
 	///list of SLOT_* defines for equip_in_one_of_slots
 	var/list/slots_to_try
-	///Ticks between the object being dropped and reequipped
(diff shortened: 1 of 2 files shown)


Lumipharon (approved):

srxtreme:
Taking the armor off a corpse with a magharn'd gun will, instead of causing the magharn'd gun to drop, cause it to remain stuck to the corpse's sprite but functionally disappear (not drop, not re-attach itself anywhere else on the corpse). Back slot was filled, so it had nowhere to re-attach itself. 
magharn

PR #14832, opened 2024-01-16
Reduces The Cost Of The Minion Spawner

Wisemonster:
## About The Pull Request
Reduces the cost of the minion spawner from 600 tac points to 400.
## Why It's Good For The Game
Right now this thing is incredibly expensive, especially compared to other tac buildings. This will hopefully make it more in-line with the cost of other tac purchase. I wasn't sure how to price this, so I set it to 400 so if it's bought round-start, there wouldn't be points for other tac structures. If it's still too expensive or too cheap, let me know
## Changelog
:cl:
balance: Reduced the price of the minion spawner from 600 to 400 tac points
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_upgrades.dm
@@ -235,7 +235,7 @@ GLOBAL_LIST_INIT(tier_to_primo_upgrade, list(
 /datum/hive_upgrade/building/spawner
 	name = "Spawner"
 	desc = "Constructs a spawner that generates ai xenos over time"
-	psypoint_cost = 600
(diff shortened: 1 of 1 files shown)


FosterGrime-Heymynameslouie:
not gonna lie it would be funny if it was like, 250 or something

the days of WO on nuke to return would be the day I die happily knowing atleast one of my bullets hit something alive

ivanmixo:
ehhh not overly keen on this

Wisemonster:
If the issue is being able to get it round-start, i could increase the price to 450, which is worth 3 phero towers and prevents it being bought round-start

TiviPlus:
I'm ok with this since while tac points have less to be spent on the cost was effectively increased by 400% with the swap to them

PR #14875, opened 2024-01-21
Removes Mimir 2 modules from HvX

lKiseki:
## About The Pull Request
Per title.

## Why It's Good For The Game
Hard counters acid, very stinky in my eyes. Mimir 1 can stay since it isn't a straight counter.

## Changelog
:cl: Lewdcifer
del: Removes Mimir 2 modules from HvX game modes.
/:cl:

### DIFF
--- a/code/modules/factory/parts.dm
+++ b/code/modules/factory/parts.dm
@@ -441,18 +441,6 @@ GLOBAL_LIST_INIT(module, list(
 	. = ..()
 	recipe = GLOB.module
 
-/obj/item/factory_part/module_mimir2
-	name = "\improper Mark 2 Mimir environmental resistance system"
-	desc = "An unfinished Mark 2 Mimir environmental resistance system module."
-	result = list(
-		/obj/item/armor_module/module/mimir_environment_protection,
-		/obj/item/armor_module/module/mimir_environment_protection/mimir_helmet,
-	)
-
-/obj/item/factory_part/module_mimir2/Initialize(mapload)
-	. = ..()
-	recipe = GLOB.module
-
 /obj/item/factory_part/module_tyr2
 	name = "\improper Mark 2 Tyr armor reinforcement"
 	desc = "An unfinished Mark 2 Tyr armor reinforcement module."
--- a/code/modules/factory/unboxer.dm
+++ b/code/modules/factory/unboxer.dm
@@ -299,12 +299,6 @@
 	refill_type = /obj/item/factory_part/module_valk
 	refill_amount = 10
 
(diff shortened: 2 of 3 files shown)


TellyLsc:
Meh, mimir 2 has always been too marginal of an upgrade over mimir 1 anyway.

Derrick2Electricboogaloo:
> Meh, mimir 2 has always been too marginal of an upgrade over mimir 1 anyway.

Wrong.

FosterGrime-Heymynameslouie:
Screenshot 2024-01-16 104843
THIS MAN IS A PROPHET

zzurge:
are you fr rn

SimplySmiley123:
> Someone commented that they'll start using mimir2 on your tyr2 removal PR and this is your response? I can't see this PR being in good faith 😁

Already on the first one it seems like a hefty "I died to to it" or "I couldn't easly and with zero struggles kill a single person out of the whole unga team" or "I got really unlucky and can't cope with it" case.

Wisemonster:
Ya'll realize that 1 person making a batch of prs doesn't necessarily mean the prs are bad, right? Having a module that is a hard counter, which is really cheap (about 120 points) is really bad, especially when some castes heavily rely on said mechanic (Boiler, or example).

Aroliacue:
> Ya'll realize that 1 person making a batch of prs doesn't necessarily mean the prs are bad, right? Having a module that is a hard counter, which is really cheap (about 120 points) is really bad, especially when some castes heavily rely on said mechanic (Boiler, or example).

It's not 1 person making a bunch of PRs, it's 1 person who deleted their head repository after the runner evasion nerf got merged and has come back with a bunch of these marine nerf PRs (two of which I like conceptually, but the majority are just godawful)

120 points per person, and they have to be willing to overlook other modules they may want to use.

It protects them against gasses (neuro/acid/whatever). It does not protect them against stings and it does not give them visibility in those gas clouds (so defiler/boiler are still blocking LOS). It is a very niche module overall and only counters being inside gas clouds (which can just be straight-up avoided 90% of the time anyway)

Dolth:
Best take would be to slightly nerf mimir but I don't even think mimir is an issue, hardly anyone uses it and it still is a req item that could be used for tyr or valk or whatever else.

Lumipharon:
I don't see the reason to kill mimir 2.
There could be an argument that gas immunity is strong enough for mimir 1 without extra acid armour (so the acid armour on 1 and 2 could be lower, as 90 acid armour is very strong). But total removal of the modular seems eh.

PR #14876, opened 2024-01-21
Nerfs the Vali module's ability availability

lKiseki:
## About The Pull Request
Per title. Changes are as follows;
- Amount of substance required to use the suit's ability increased from 10 > 150.
- Amount of substance drained while the ability is active changed.
  - The previous formula was: `Boost level * (3 + boost level)`, resulting in 4 at x1, and 10 at x2 _(very inconsistent, mind you)_.
  - The new formula is: `10 * boost level`, resulting in 10 at x1 and 20 at x2.
- Amount of substance acquisition changed.
  - The previous formula was: `20 * (weapon attack speed / 11)`, which had varying results depending on your weapon of choice;
    - 14 units per Knife attack.
    - 21 units per Harvester attack. 
    - 43 units per Claymore attack.
  - The new formula is: `10 + (weapon attack speed * 0.5)`, which has varying results depending on your weapon of choice;
    - 14 units per Knife attack _(note it remains unchanged, this is intentional for the knife)_.
    - 16 units per Harvester attack.
    - 22 units per Claymore attack.

## Why It's Good For The Game
We're trying to cut down on just how tanky marines can be, and Vali is a major offender in this party. This PR makes it so that vali users have to put in more work to use their boost, and forces them to actually think about when and where to use it as its availability is reduced.

## Changelog
:cl: Lewdcifer
balance: Vali: amount of substance required to use the suit's ability increased from 10 > 150.
balance: Vali: amount of substance drained while the ability is active changed. Previously, it was 4 at x1 and 10 at x2. This was changed to 10 at x1 and 20 at x2.
balance: Vali: amount of substance acquisition changed. Previously, it was 21 for harvester, and 43 for claymore. This was changed to 16 for harvester, and 22 for claymore. Knife unaffected.
/:cl:

### DIFF
--- a/code/datums/components/chem_booster.dm
+++ b/code/datums/components/chem_booster.dm
@@ -24,7 +24,9 @@
 	var/resource_storage_max = 200
 	///Amount of substance stored currently
 	var/resource_storage_current = 0
-	///Amount required for operation
+	/// Amount of substance required to start operating.
+	var/resource_storage_required = 150
+	/// Amount of substance to drain while operating.
 	var/resource_drain_amount = 10
 	///Actions that the component provides
 	var/list/datum/action/component_actions = list(
@@ -243,7 +245,7 @@
 		if(!COOLDOWN_CHECK(src, chemboost_activation_cooldown))
 			wearer.balloon_alert(wearer, "You need to wait another [COOLDOWN_TIMELEFT(src, chemboost_activation_cooldown)/10] seconds")
 			return
-		if(resource_storage_current < resource_drain_amount)
+		if(resource_storage_current < resource_storage_required)
 			wearer.balloon_alert(wearer, "Insufficient green blood to begin operation")
 			return
 		if(wearer.stat)
@@ -284,7 +286,7 @@
 /datum/component/chem_booster/proc/update_boost(amount)
 	boost_amount = amount
 	wearer?.balloon_alert(wearer, "Enhancement level set to [boost_amount]")
-	resource_drain_amount = boost_amount*(3 + boost_amount)
+	resource_drain_amount *= amount
 
 ///Handles Vali stat boosts and any other potential buffs on activation/deactivation
 /datum/component/chem_booster/proc/setup_bonus_effects()
@@ -378,7 +380,7 @@
 		return
 	if(resource_storage_current >= resource_storage_max)
 		return
-	update_resource(round(20*connected_weapon.attack_speed/11))
+	update_resource(round(10 + connected_weapon.attack_speed * 0.5))
 
 ///Adds or removes resource from the suit. Signal gets sent at every 25% of stored resource
 /datum/component/chem_booster/proc/update_resource(amount)


Looking4bros:
This'd like... force valitards to never necro. And the previous greenblood consumed was formulated that way as a way to show diminishing returns, because that's normal in life. you can only milk a cow so much.

I'd suggest removing all toxin healing from Synaptizine and Dylovene on the boost, and making it slightly toxic in it's use, you could try looking at the defiler chems that do similar things for an estimate on the percentages, and adjust it's toxicity on additional chems, i'd suggest leaving Neuraline as is, as it's already too much for more than maybe ten ticks, before you'd have to pull out entirely or die.

SimplySmiley123:
> Why is it that nobody ever seems to realise the vali healing is there to counteract the fact that you're pressed up against xenos and therefore taking way more damage than any other marines? Your healing can only just keep pace with a single T1 slashing at you (after armor reduction). If you're running into melee with xenos, you're engaging on their terms and take more damage, from more xenos than marines who are using guns do.
> 
> Why would you ever use vali over a shield+gun with this kind of gutted ability to heal? You only need a single runner to pounce the vali while they can't heal and now they've got a frac and need to go shipside because splints fall off instantly and even more instantaneously for valis. With a shield you would have an easier ability to tank and do more damage, at range. This nerf doesn't even make you some kind of glass cannon, it just makes you a detriment to your team (even more than most valis already are since spear removal) that's going to fall over as soon as they get into a fight unless xenos are so kind as to stand still and let you slash them ten times before hitting back.

I think that's the point of these. Not to make it any better, but to gimp it so there's another easy-picking for xeno mains.

chizzy376:
So what's the point of all these PR's? Is it to hugbox xenos until they get a 99% winrate again?
Frankly, I haven't seen any issues come from the availability of modules.

SimplySmiley123:
> So what's the point of all these PR's? Is it to hugbox xenos until they get a 99% winrate again? Frankly, I haven't seen any issues come from the availability of modules.

personally the issue I seen from both sides is literally a players issue of them... Ya know, not cooperating in a team game? especially on xenomorphs side where it matters way more for them to cooperate.

bowlofcereal:
Why are you going on a rampant removal spree dude? I'm used to you making.. "interesting" PRs but this is a little obscene. 3 PRs trying to decimate 3 different parts of loadouts entirely in the span of a singular day.

If you aren't enjoying the game you should probably take a break. I am being serious. It isn't good for your health. There is no bad faith to be had.

Lumipharon:
Not going to close this because I don't play vali enough, but honestly it needs a rework more than just a simple nerf.

SimplySmiley123:
> Not going to close this because I don't play vali enough, but honestly it needs a rework more than just a simple nerf.

It does need a rework, not necesarilly raw nerfs or buffs.

Autisem:
I think the main issue with vali is that(especialy with the knife) the melee playstle is optional. your far better off grinding some blood and using shotgun. While keeping it as a pocket heal.

Autisem:
> And make it much much much much more ass to play with. Mostly because vali itself is pretty mediocre and unless xenomorphs are dumb you're not really gonna live all that long either.

Vali was made to have a melee playstyle, and was only balanced around that. Is hardly about being optimal or not. Thats why its OP if you dont use it in melee

PR #14888, opened 2024-01-22
Even more DNR (& defib) fixes.

DeltaFire15:
## About The Pull Request
So. Turns out that while I did correctly discern & fix the two issues I mentioned in a previous PR..
Disconnecting while ghosted ALSO DNRd you because get_ghost at the moment only returns a ghost if it has a client.
This adjusts that by adding a var to the proc that makes it return the ghost even if it has no client if true.

Additionally adjusts uses of get_ghost where it recognizing disconnected but not DNR ghosts is favorable (mostly defib huds and said DNR case).

Lastly, this also goes into defib code and adjusts conditions for various messages because some were plain wrong and others just have their conditions set kind of badly (it still fails for disconnected targets but should at least give you a different message).

Please TM this first because it is torture to test thoroughly.

## Why It's Good For The Game
Fix man good.

## Changelog

:cl:
fix: Various HUD elements now display revivability more accurately.
fix: Another case that would randomly DNR you has been fixed.
fix: Defib messages & failures should be less likely to display in incorrect situations.
fix: Minimap icons should no longer display as DNR for now fixed cases.
/:cl:

### DIFF
--- a/code/game/data_huds.dm
+++ b/code/game/data_huds.dm
@@ -259,14 +259,19 @@
 		simple_status_hud.icon_state = ""
 		if(stat != DEAD)
 			status_hud.icon_state = "hudsynth"
+		else if(HAS_TRAIT(src, TRAIT_UNDEFIBBABLE))
+			status_hud.icon_state = "hudsynthdnr"
+			return TRUE
 		else
-			if(!client)
-				var/mob/dead/observer/G = get_ghost(FALSE, TRUE)
+			if(!mind)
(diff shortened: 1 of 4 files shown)


ivanmixo (approved):
What a nightmare

DeltaFire15:
>Please TM this first

I am not longer responsible for any consequences if something explodes down the line.

Lumipharon:
Not on first line, did not read

PR #14889, opened 2024-01-22
puppeteers organic bomb base timer reduced, now has the windup

usnpeepoo:
## About The Pull Request

organic bomb base timer before automatically detonating reduced from 15 secs to 5, now has the 1.5sec windup instead of an instant explosion
## Why It's Good For The Game

Very underused ability, at times worse because 3 puppets with fury blessing will almost always be a better option (and reducing your puppets for this ability feels weak). The acid itself isn't that strong either. Couple this with a base 15 sec timer before it explodes by itself without reaching a target (and with minion fuckery) makes this abillity very weak as the puppets will likely just die by the time it reaches its target.. In practice the total time before the puppets explode (without slashing anything) will be 6.5s (5 + 1.5s windup)

## Changelog

:cl:
balance: Puppeteers organic bombs base timer reduced, now uses a windup
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/puppeteer/abilities_puppeteer.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/puppeteer/abilities_puppeteer.dm
@@ -274,7 +274,7 @@
 		return fail_activate()
 	RegisterSignal(victim, COMSIG_XENOMORPH_ATTACK_LIVING, PROC_REF(start_exploding))
 	RegisterSignal(victim, COMSIG_MOB_DEATH, PROC_REF(detonate))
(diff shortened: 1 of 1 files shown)


ivanmixo (commented):
Why'd you change the proc to start exploding?

usnpeepoo:
> Why'd you change the proc to start exploding?

that proc has the 1.5s timer above the puppet and the message that gets put into chat 
it was using detonate() which just exploded it on the spot instantly

ivanmixo:
Think itd be better if it had the alert before it went boom

usnpeepoo:
> Think itd be better if it had the alert before it went boom

ill change that then

ivanmixo (approved):

PR #14898, opened 2024-01-23
Xenomorph resting rework

ivanmixo:
## About The Pull Request

This is essentially a shameless ripoff of CM's resting. You can lay down instantly, but then cannot get up for 2 seconds after. In exchange you get up instantly after the 2 seconds have passed.

I also renamed the proc to be more clear about what it does and put in docs
## Why It's Good For The Game

As it stands right now xenos get punished for not mashing the hell out of their resting key because it is objectively better to do so as it makes you safer from being jumped. This should help make all xenos equal, lazy resting or not
## Changelog
:cl:
balance: Xenomorph resting now blocks you from unresting for 2 seconds, instead of it being a 2 second progress bar to get up
/:cl:

### DIFF
--- a/code/__DEFINES/mobs.dm
+++ b/code/__DEFINES/mobs.dm
@@ -479,6 +479,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 #define XENO_SILO_DETECTION_RANGE 10//How far silos can detect hostiles
 #define XENO_GARGOYLE_DETECTION_COOLDOWN 30 SECONDS
 #define XENO_GARGOYLE_DETECTION_RANGE 10//How far gargoyles can detect hostiles
+#define XENO_RESTING_COOLDOWN 2 SECONDS
 
 #define XENO_HIVEMIND_DETECTION_RANGE 10 //How far out (in tiles) can the hivemind detect hostiles
 #define XENO_HIVEMIND_DETECTION_COOLDOWN 1 MINUTES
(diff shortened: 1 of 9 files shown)


FosterGrime-Heymynameslouie:
> This makes it impossible to punish xenos resting on frontlines unless they do it literally in view of you. The unrest-spam meta is bad but this PR just gives a better version of that to every xeno.
> 
> Why not make resting start at a lower healing speed then ramp up as you're laying down, so unrest spam is no longer the best thing to do?

restint ramp up sounds like a good idea

why would you insta heal a xeno for taking a 2 second nap? that shits like the RR without the cloneloss the fuck?

Autisem:
does this work while staggerd? if not please add that

AusarAtkin:
> This makes it impossible to punish xenos resting on frontlines unless they do it literally in view of you. The unrest-spam meta is bad but this PR just gives a better version of that to every xeno.
> 
> Why not make resting start at a lower healing speed then ramp up as you're laying down, so unrest spam is no longer the best thing to do?

Benos already have a healing cooldown when damaged, there's no real reason to add this on top of that.

TheWiznard:
rip my rest spam when this inevitably is merged

Lumipharon (approved):

Autisem:
> and plasma gain seems completely unaffected. 

unless this got changed it should very much effect it

AusarAtkin:
> > Benos already have a healing cooldown when damaged, there's no real reason to add this on top of that.
> 
> I never notice it. I rest right around corners/walls or in the middle of smoke and still get a decent amount of health back, and plasma gain seems completely unaffected. You can also make it so that resting's heal speed starts slower... and then ramps up to being faster than it currently is. That way you simultaneously kill unrest spam, make it possible to punish frontline resting, and keep overall resting heal speed virtually identical to what it is now.

It depend on the strength of recovery pheromones, currently its hidden stat but Ivan also changing that in #14878 .

PR #14899, opened 2024-01-23
Runner evasion can be cancelled early

XElectricX:
## About The Pull Request

Couple of lines of code that accommodate for cancelling evasion if active. Ability still goes on cooldown on activation regardless of if cancelled or not.

Has a 1 second cooldown before it can be deactivated to prevent accidental deactivations when trying to spam press it in the heat of combat. Could probably be 0.5 seconds if players feel it is too long ingame.

## Why It's Good For The Game

Yes it's a minor buff, but more importantly quality of life feature since evasion disarms the rounies and they must wait out the entire duration.

## Changelog
:cl:
qol: Runners can manually cancel evasion.
balance: Runners no longer disarmed for the entire duration of evasion if they cancel it.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
@@ -120,6 +120,9 @@
 	 * When called by the evasion extension, deactivate will be FALSE so the ability won't turn off by itself
 	 */
 	if(deactivate && evade_active)
+		if(TIMER_COOLDOWN_CHECK(src, COOLDOWN_EVASION_ACTIVATION))
+			owner.balloon_alert(owner, "Accidental deactivation guard!")	//Little message to let the player know and not think it is a bug that their evasion didn't turn off
+			return
 		evasion_deactivate()
 		return
 
@@ -148,6 +151,7 @@
 	RegisterSignal(owner, COMSIG_LIVING_PRE_THROW_IMPACT, PROC_REF(evasion_throw_dodge))
 	GLOB.round_statistics.runner_evasions++
 	SSblackbox.record_feedback("tally", "round_statistics", 1, "runner_evasions")
+	TIMER_COOLDOWN_START(src, COOLDOWN_EVASION_ACTIVATION, 1 SECONDS)
 
 /datum/action/ability/xeno_action/evasion/process()
 	var/mob/living/carbon/xenomorph/runner/runner_owner = owner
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
```diff
@@ -112,7 +113,20 @@
 	action_icon_state = "evasion_[auto_evasion? "on" : "off"]"
 	update_button_icon()
 
-/datum/action/ability/xeno_action/evasion/action_activate()
+/datum/action/ability/xeno_action/evasion/action_activate(deactivate = TRUE)
```
Lumipharon, line 116: This is smelly, Don't do this.

Just make the proc itself check if it's active, and deactive if it is instead of activate.

Lumipharon (changes requested):

code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
```diff
@@ -113,6 +114,16 @@
 	update_button_icon()
 
 /datum/action/ability/xeno_action/evasion/action_activate()
+	//Since both the button and the evasion extension call this proc directly, check if the cooldown timer exists
+	//The evasion extension removes the cooldown before calling this proc again, so use that to differentiate if it was the player trying to cancel
+	if(evade_active && cooldown_timer)
+		if(TIMER_COOLDOWN_CHECK(src, COOLDOWN_EVASION_ACTIVATION))
+			owner.balloon_alert(owner, "Accidental deactivation guard!")	//Little message to let the player know and not think it is a bug that their evasion didn't turn off
```
Lumipharon, line 121: this is extremely wordy for a balloon alert.

Lumipharon (approved):

code/modules/mob/living/carbon/xenomorph/castes/runner/abilities_runner.dm
ivanmixo: its fine, just probably has no reason to exist. don't see why the user needs to be alerted about this in the first place

TiviPlus:
```suggestion
```
?

XElectricX: I guarantee someone will say the evasion cancelling is buggy or laggy because they tried to do it too fast. Informing the user something is not working the way they intended matters.

ivanmixo: Then someone will just tell them they're being a goober, this is clutter

XElectricX: Ok.

PR #14906, opened 2024-01-25
The Great Unification of Modules

Pariah919:
## About The Pull Request
Kuro resprited modules, removed xenonauten having its own variants, basically speaking everything is unified.
asdasdasd
## Why It's Good For The Game
Unification of art in MY TGMC? More likely than you think.
## Changelog
:cl:
spriteadd: New sprites for Jaeger modules.
/:cl:


Kuro020 (approved):

Derrick2Electricboogaloo:
So your armor will not change based on what armor mod you have on?

TellyLsc:
The module sprites are so obtrusive that you can't even see what armour you're wearing anymore.

Kuro020:
> The module sprite overlays on some modules are so obtrusive that you can't even see what armour you're wearing anymore. image The left one is "Xenonauten Medium Mimir" and the right one is "Jaeger EOD Heavy Mimir" and there's no way you could tell that without being told so with how much the module overlay covers the armour. image Another example with Arrowhead, try guessing how many of these are medium, light, or heavy armor.
> 
> Not to mention other sprites such as belt, armor storage, satchel sling, etc. covering it up.
> 
> Also I was hoping if paintable armour modules would be possible? I guess the module type can be discerned by unrecolourable glow in the dark stripes then might as well.

paintable modules can be done, not on this pr tho

also not being able to see which armor it is, its whatever honestly, capes bags and whatever else + the ammount of armor variations we have means you cant tell what level of armor it is anyway, at least now you know what module they have

t-toasted:
Playing with this makes the armors kind of ugly in some cases. The weird shoulder pads on mimir stand out oddly on slim type armors, and for example, the tyr helmet choice is really bad; it just covers your entire faceplate (at least with robot helmet). I appreciate the module visuals being unified now, however, and being able to tell who has what at a glance, but it hinders some of the visual uniqueness of armor and makes it harder to design an outfit when you have to work around mimir for instance.

EaglePhntm:
cool things like open helmet mimir is gone....

PR #14911, opened 2024-01-26
Gives all xenos with acid strong acid during prep

Barnet2:
## About The Pull Request
What it says on the tin. Shares the same checks as quickbuild so podding/tadding should also disable it, alongside shutters of course.
## Why It's Good For The Game
Less prep torture, given Queen is usually the one with it and they have things to do
## Changelog
:cl:
balance: All xenomorphs with acid now have strong acid during prep (Nuclear War)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/abilities.dm
+++ b/code/modules/mob/living/carbon/xenomorph/abilities.dm
@@ -599,6 +599,9 @@
 	use_state_flags = ABILITY_USE_BUCKLED
 
 /datum/action/ability/activable/xeno/corrosive_acid/can_use_ability(atom/A, silent = FALSE, override_flags)
+	var/obj/effect/xenomorph/acid/current_acid_type = acid_type
+	if(SSmonitor.gamestate == SHUTTERS_CLOSED && CHECK_BITFIELD(SSticker.mode?.flags_round_type, MODE_ALLOW_XENO_QUICKBUILD) && SSresinshaping.active)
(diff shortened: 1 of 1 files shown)


TiviPlus (approved):

DeltaFire15:
Not entirely a fan of amping terraforming this way. Weaker acid abilities tend to be a lot cheaper to use, making it objectively superior to go sentinel over boiler to melt rwalls pre-shutters (with this change).

Given that strong acid is only required for reinforced walls, there are already a lot (the majority) of walls that you can melt through without needing a queen or boiler regardless.

Barnet2:
> Not entirely a fan of amping terraforming this way. Weaker acid abilities tend to be a lot cheaper to use, making it objectively superior to go sentinel over boiler to melt rwalls pre-shutters (with this change).
> 
> Given that strong acid is only required for reinforced walls, there are already a lot (the majority) of walls that you can melt through without needing a queen or boiler regardless.

The idea with this is alongside the evo rework where people normally should be going drone and what not. Boiler was extremely niche in this category where people would go it shutters to wait around FOB to melt walls which I don't think is exactly healthy

usnpeepoo:
Will make xeno flanking options more prevalent which means funner gameplay +1

Ragecrabbo:
I think you are underestimating the power of strong acid. My prediction is that xenos will start deleting every wall near disks to make building a new APC pain. Either way this will definately be interesting to see.

Kuro020:
very nice pr

PR #14917, opened 2024-01-27
Xeno AP

Lumipharon:
## About The Pull Request
Adds an AP stat to xenos, for melee attacks.

Xeno castes can now have AP.

Hunter and mantis get 5 AP (Hunter in particular has been kinda powercrept out of usefulness).
Ravager has 15AP but has its damaged reduced from 30 to 25.

This means Rav does slightly more against light armour nerds, but significantly more against the tankiest of marines such as tyr heavy or B18.

Numbers are experimental, may need some testing. It could be added to other castes in the future of course, but I wanted to keep it limited to start with for easier testing.
## Why It's Good For The Game
Gives xenos some way to more effectively counter heavy tankiness short of 'more xenos' or 'stun lock to death'.
## Changelog
:cl:
balance: Ravager, Hunter and Mantis now have some AP on their melee attacks
/:cl:

### DIFF
--- a/code/_onclick/xeno.dm
+++ b/code/_onclick/xeno.dm
@@ -15,7 +15,7 @@
 	SSblackbox.record_feedback("tally", "round_statistics", 1, "xeno_unarmed_attacks")
 
 
-/atom/proc/attack_alien(mob/living/carbon/xenomorph/X, damage_amount = X.xeno_caste.melee_damage, damage_type = BRUTE, damage_flag = "", effects = TRUE, armor_penetration = 0, isrightclick = FALSE)
+/atom/proc/attack_alien(mob/living/carbon/xenomorph/X, damage_amount = X.xeno_caste.melee_damage, damage_type = BRUTE, damage_flag = "", effects = TRUE, armor_penetration = X.xeno_caste.melee_ap, isrightclick = FALSE)
 	return
 
 
--- a/code/datums/gamemodes/campaign/missions/base_rescue.dm
+++ b/code/datums/gamemodes/campaign/missions/base_rescue.dm
@@ -147,7 +147,7 @@
 	else
 		icon_state = "[initial(icon_state)]_open"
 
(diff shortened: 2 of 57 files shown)


bowlofcereal:
i want to see how this plays out, yeah

Kuro020 (approved):
gaming

lordexvar:
utterly opposed to having ap tied to all of a castes attacks rather than abilities. ravager ravage, warrior punch/jab, etc. makes sense to have ap. tying it to all attacks just significantly buffs that caste lazily.

Breadeagle:
I assume this only affects the slash of a xeno, so abilities that do damage like punch, ravage, or flay can have their AP tuned independently? I don't see why this would be a lazy change to any caste given that it'd be adding a balance lever in this case.

Lumipharon:
its specifically normal attacks yes.

MiniMeatwad:
Anything to prevent over-stacking armor modifiers to get like 95% damage reduction.

coldironwarrior:
mostly second what bell said; i dont totally enjoy the move away from distinctly knowing that my armor protects me from most instances of IB and organ damage and if i go above X damage, i can expect to have internal damages, to now, it's...i get hit and regardless of my armor build, maybe i walk away with IB or a damaged liver but there's no real way to know.

would still be interested in seeing how this plays for TM, but i do think it'll make playing a human tedious if you get organ damage on every encounter with a rav or hunter.

also hunter is remarkably useful. whoever's saying it's powercrept out of usefulness hasn't played or hasn't played with patience. it's not going to chain a bunch of high-damage attacks in a crowd and endure like rav.

Arturlang:
May want to consider tying it to abilities instead, though I'm not sure

UltraBigRat:
Even as a xeno main this feels like too much, id much rather deal with heavy armor nerds rather then everyone just wearing light armor because AP bypasses it anyways

zzurge:
the return of the ravager stack

lKiseki:
Removed this testmerge as it was clearly busted in terms of balance. I'd recommend further reducing the slash damage or rethinking who gets AP, because Ravagers and Hunters are shredding marines.

lordexvar:
this merge is fucking hilarious i need to get all rav mains back while this lasts it is insane

MiniMeatwad:
Eh, just brings in hard armor bullshit if you want to work around it, and doesn't really improve gameplay.

PR #14919, opened 2024-01-27
Silo time scaling

Lumipharon:
## About The Pull Request
What it says on the tin. Aka: It only gets worse.

Larva spawn rate now scales with round duration, up to a max of x2 larva rate after 3 hours.
## Why It's Good For The Game
Marines are nominally the attacking faction in HvX, yet they tend to benefit more from prolonged rounds due to req and other factors. This is somewhat compounded by our modern high pop, where rounds can easily be prolonged to allow for large marine respawn waves.

This change creates a level of time pressure on marines, which encourages them to actually be aggressive, and disincentivises prolonged seiges or stalemates.
## Changelog
:cl:
balance: Larva generation slowly scales up over time, to a maximum of x2 speed after 3 hours
/:cl:

### DIFF
--- a/code/controllers/subsystem/silo.dm
+++ b/code/controllers/subsystem/silo.dm
@@ -1,3 +1,8 @@
+///Scaling multiplier for larva respawn rate based on round time
+#define XENO_SPAWN_TIME_SCALING_MAX_AMOUNT 2
+///Time at which scaling multiplier hits its max
+#define XENO_SPAWN_TIME_SCALING_MAX_TIME 2.5 HOURS
+
 SUBSYSTEM_DEF(silo)
 	name = "Silo"
 	wait = 1 MINUTES
@@ -28,6 +33,8 @@ SUBSYSTEM_DEF(silo)
 	current_larva_spawn_rate *= SSticker.mode.silo_scaling
 	//We scale the rate based on the current ratio of humans to xenos
 	current_larva_spawn_rate *= clamp(round((active_humans / active_xenos) / (LARVA_POINTS_REGULAR / xeno_job.job_points_needed), 0.01), 0.5, 1)
+
+	current_larva_spawn_rate *= LERP(1, XENO_SPAWN_TIME_SCALING_MAX_AMOUNT, min((world.time - SSticker?.round_start_time) / XENO_SPAWN_TIME_SCALING_MAX_TIME, 1))
 
 	current_larva_spawn_rate += larva_spawn_rate_temporary_buff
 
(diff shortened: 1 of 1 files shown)


bowlofcereal:
I love a good challenge, bring on the horde

coldironwarrior:
i dont think the correct way to counter marine req snowballing is to make xenos snowball even harder than marines.

if the issue is req point generation over a prolonged round, then maybe what needs to be reworked is...req point generation, that unglamorous terrible system we all talk about changing but never do anything for

code/controllers/subsystem/silo.dm
```diff
@@ -29,6 +34,8 @@ SUBSYSTEM_DEF(silo)
 	//We scale the rate based on the current ratio of humans to xenos
 	current_larva_spawn_rate *= clamp(round((active_humans / active_xenos) / (LARVA_POINTS_REGULAR / xeno_job.job_points_needed), 0.01), 0.5, 1)
 
+	current_larva_spawn_rate *= LERP(1, XENO_SPAWN_TIME_SCALING_MAX_AMOUNT, min((world.time - SSticker?.round_start_time) / XENO_SPAWN_TIME_SCALING_MAX_TIME, 1))
```
coldironwarrior, line 37: regardless of my aforementioned opinion on bigger-picture problems and onto the actual code after looking at this again:
would this indicate that the scaling begins from the start of the round? would there be a way to have it start from 12:18 (minimum deployment time, usually) instead so that there aren't 15 minutes that count into the scaling factor that marines aren't realistically utilizing while groundside?

Kuro020 (approved):

MiniMeatwad:
A dead xenomorph is just more req points, after all

code/controllers/subsystem/silo.dm
ivanmixo: Yeah would probably be good if this ticked only after shutters open.

Breadeagle:
Probably far too much of a xeno buff - this is equivalent to saving and accelerating psy generation that might otherwise be spent on silos. This just isn't an interesting way to change xenos, it just sort of makes them stronger.

DeltaFire15:
This is bad. Some people who have been here for a while may recall a certain quirk of nuke war where it used to have increased silo impact relative to base modes (now all other modes are dead so the modifier is effectively the base despite not being 1).
Well, I sure wonder what happened to that.
Started at x3, got nerfed to x2.5 by #9835 , nerfed to x2 by #10309 , then again nerfed to 1.5x by #13925, which has remained our current final value.

At the moment, you can already scale silo gain during the round. By making more silos. This adds onto that, creating an unreasonably high amount of pressure on marines and negating negatives for dying even further. Or, probably more likely, giving xenos free autobalance because there aren't remotely enough ghosts for all the larva they will be getting.

lKiseki:
I agree with the above two posts. Besides that, I'm here to report that this PR sent larva generation to the afterlife:
image
I'm gonna unTM it until it gets fixed.

TiviPlus:
What deltafire said

Kuro020:
This isnt needed anymore, xenos are doing fine

PR #14920, opened 2024-01-27
Reduce SOM bullet armour

Lumipharon:
## About The Pull Request
SOM bullet armour reduced across the board by 5 (with the exception of lorica module, which remains at parity with tyr 2).
This means SOM armour now has equal bullet armour to marines, but still has 5 worse laser armour.

Also fixed a minor armour bug, lorica had slightly lower slowdown than tyr 2 for some reason.
## Why It's Good For The Game
Makes some projectile weapons a bit more viable against SOM. Variety is good. Lasers are still (generally) better against SOM however.
## Changelog
:cl:
balance: SOM armour no longer has superior bullet armour compared to marines. Still retains a slight weakness to lasers
fix: Lorica armour has equal slowdown to tyr 2
/:cl:

### DIFF
--- a/code/modules/clothing/modular_armor/attachments/modules.dm
+++ b/code/modules/clothing/modular_armor/attachments/modules.dm
@@ -135,8 +135,8 @@
 	icon_state = "lorica_armor"
 	item_state = "lorica_armor_a"
 	attachment_layer = null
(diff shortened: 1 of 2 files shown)


chizzy376:
Does this apply to Campaign or the SOM ert as well?

Lumipharon:
both. there's no stat differences between hvh and hvx

ivanmixo (approved):

PR #10507, opened 2022-06-21
Blood over maximum leaks out

QualityVan:
## About The Pull Request
If a human is above maximum blood volume (about 7% above normal max) the extra blood will fall out. Go for the high score and splash your friends. QC suppresses bleeding as usual but once it runs out you pop. Doesn't apply any actual damage.

## Why It's Good For The Game
Fixes #10488 
Prevents just putting iron/qc in your system and getting to ignore anything that affects blood level.

## Changelog
:cl:
add: If you have too much blood in you and no quickclot, all the extra blood will spill.
/:cl:

PR #14926, opened 2024-01-27
You can no longer go over blood limit, or below 0 blood

ivanmixo:
## About The Pull Request

Title. Added 2 wrappers used for managing the blood volume var, which ensure we don't go over the maximum or the minimum blood level.

## Why It's Good For The Game

There have been some balancing issues when people reach very high blood levels, mainly because of nanites. 

Going below 0 blood is also bad.

## Changelog
:cl:
fix: You can no longer go over the blood limit, or go below 0 blood
/:cl:


Xander3359:
The boring solution, my PR should get merged instead :)

MiniMeatwad:
> The boring solution, my PR should get merged instead :)

Yours is funny and this is a bug fix. They're both good.

Aporhtonoma:
i like xanders solution to bloodmaxing more 
:(

QualityVan:
As I've told you specifically, the qc thing is intended. It's clearly stated on the pr that added spilling, #10507 . If it wasn't, what would the point of the if drip line be and how would the higher thresholds be reachable? IVing a corpse?

code/modules/mob/living/blood.dm
```diff
@@ -14,7 +14,7 @@
 	if(stat != DEAD && bodytemperature >= 170)	//Dead or cryosleep people do not pump the blood.
 		if(blood_volume > BLOOD_VOLUME_MAXIMUM) //Warning: contents under pressure.
 			var/spare_blood = blood_volume - ((BLOOD_VOLUME_MAXIMUM + BLOOD_VOLUME_NORMAL) / 2) //Knock you to the midpoint between max and normal to not spam.
-			if(drip(spare_blood))
+			if(drip(spare_blood, TRUE))
```
QualityVan, line 17: I'd just cut the drip line entirely if you insist on this rather than adding a workaround. If(true) is weird, and there's no case where drip would return false here.

ivanmixo: Returning null is a falsey value, so there is a case.

QualityVan: When would a null return be reached by this call? The TRUE prevents an early return, and if this line is reached blood volume is nonzero. NOBLOOD species aren't reaching it. In stasis, I guess? But then chems aren't processing, and I don't think there's any way to IV someone while they're also in a stasis effect.

ivanmixo: `.` is null by default in `drip`, so returning if theres qc in their system is returning null

QualityVan: Which, like I said, is unreachable in this call because the TRUE argument prevents qc from mattering.

ivanmixo: oh that was your point sorry

MiniMeatwad: Drip function before changes:
```
/mob/living/carbon/proc/drip(amt)

	if(reagents.get_reagent_amount(/datum/reagent/medicine/quickclot)) //Quickclot stops bleeding, magic!
		return

	if(blood_volume)
		blood_volume = max(blood_volume - amt, 0)
		if(isturf(src.loc)) //Blood loss still happens in locker, floor stays clean
			if(amt >= 10)
				add_splatter_floor(loc)
			else
				add_splatter_floor(loc, 1)
		return TRUE
```
It's acknowledging the check for a 'blood_volume'.

efzapa:
gibbing nanite users is pretty funny though ngl

Lumipharon:
I would just add a hardcap to blood tbh. There's zero reason to even let people go above 100% blood to begin with.

code/modules/mob/living/blood.dm
```diff
@@ -54,32 +32,34 @@
...
 				switch(nutrition)
 					if(NUTRITION_OVERFED to INFINITY)
 						adjust_nutrition(-10)
-						blood_volume += 1 // regenerate blood quickly.
+						adjust_blood_volume(1)// regenerate blood quickly.
+
 					if(NUTRITION_HUNGRY to NUTRITION_OVERFED)
 						adjust_nutrition(-5)
-						blood_volume += 0.5 // regenerate blood slowly.
+						adjust_blood_volume(-5) // regenerate blood slowly.
```
QualityVan, line 58: .

code/modules/mob/living/living_health_procs.dm
```diff
@@ -191,6 +191,20 @@
 		return
 	remove_movespeed_modifier(MOVESPEED_ID_DROWSINESS)
 
+///Adjusts the blood volume, with respect to the minimum and maximum values
+/mob/living/proc/adjust_blood_volume(amount)
+	if(!amount)
+		CRASH("adjust_blood_volume called with [isnull(amount) ? "null" : "[amount]"] amount of blood.")
```
QualityVan, line 197: Do you really want a crash out if adjust(0) is called?

code/modules/mob/living/blood.dm
```diff
@@ -11,35 +11,13 @@
...
-						bleed_range = 3
-				if(bleed_range)
-					for(var/turf/canvas in RANGE_TURFS(bleed_range, src))
-						add_splatter_floor(canvas)
-					for(var/mob/canvas in viewers(bleed_range, src))
-						canvas.add_blood(species.blood_color) //Splash zone
-					playsound(loc, 'sound/effects/splat.ogg', 25, TRUE, 7)
-
 	//Effects of bloodloss
+	if(stat != DEAD && bodytemperature >= 170)	//Dead or cryosleep people do not pump the blood.
```
QualityVan, line 15: Early return

QualityVan (changes requested):

code/modules/mob/living/blood.dm
ivanmixo: FUNNY

ivanmixo, line 58:
```suggestion
						adjust_blood_volume(0.5) // regenerate blood slowly.
```

code/modules/mob/living/living_health_procs.dm
ivanmixo: It really shouldn't happen shrug

code/modules/mob/living/blood.dm
Xander3359: "Slowly"

code/modules/mob/living/living_health_procs.dm
ivanmixo: Ill change it to just if its null when Im home

PR #14927, opened 2024-01-27
removes xeno probing

coldironwarrior:
## About The Pull Request

title. barring a true Researcher Rework, i've opted to just removing xeno probing dead corpses. you can still play gacha with digsite materials.

changes out the sprite for the xeno analyzer with the implant case/pad sprite, since the analyzer is still used in interactions with newt (why were we hitting newt with a saw); additionally puts a limit on the number of analyzers vendable from the nanomed because it's serving a different purpose now and 12 people fighting to have newt follow *them* seems like it'd be a headache.

**THE MATHS**
totally optional and not mandatory to keep, but given that this whole thing would provide a nerf to marine incomes, i've opted to also go ahead and re-balance the point values per xeno corpse sold to be a bit more generous, while still (roughly) preserving the ratios from #14118. stuff i took into account for this is namely that each corpse probed guarantees the generation of a 50-credit coin at minimum, and that pushing t4s to 1200 makes them equivalent to some of the really critical req gear. i didn't change the t1 value for reasons of #ratios.

T1 - 150 points (no change)
T2 - 350 points (+50)
T3 - 600 points (+100) [includes shrike]
T4 - 1100 points (+100 points)

ratios
image

## Why It's Good For The Game

people seem to think this assists in req snowballing and i much prefer to fix up this aspect of the game versus trying to power scale either team in response to the req snowball

## Changelog

:cl:
add: removes xeno probing
balance: point adjustments to xeno caste values for req exports
/:cl:

PR #9404, opened 2022-01-25
Removes hard skill lock from research xenomorph analyzer

Verlurk:
## About The Pull Request

Removes hard skill lock from xenomorph analyzer, enabling everyone to use them with a 15 second fumble dependent on your medical skill. No change to excavations.

## Why It's Good For The Game

Hard locks against TGMC ethos, allows a variation of playstyles if nobody's playing researcher. Dying as xenos now has a much higher risk.

## Changelog
:cl:
balance: Xenomorph analyzer is now useable by all, with a delay to reflect lack of skill.
/:cl:

PR #14928, opened 2024-01-27
re-locks probing to surgery skill (revert #9404: Removes hard skill lock from research xenomorph analyzer)

coldironwarrior:
## About The Pull Request

reverts #9404 to make probing xenos locked to researchers, doctors, CMOs, and synths

## Why It's Good For The Game
this is a band-aid that dances around the problem more, but may be better received than #14927. it doesn't change the point scaling, but makes it more limited in who can participate in it.

differentially, removing one of the Primary Things for a specific job in the spirit of "less skill locks", when that thing wasn't replaced for that job just seems...bad. why even have researcher, then, you know?

## Changelog

:cl:
balance: re-locks xeno probing to researchers and those with similar or higher surgery skill
/:cl:


dopamiin0:
this is really dumb and only serves to punish lowpop marines and marines who didnt happen to get a researcher player. researchers are already better at researching, and will virtually always be allowed to. non-researchers researching only realistically happens when there's no researcher, which is already penalized by the action taking way longer AND by taking up someone who likely has something else to be doing. if you think xeno probing itself is bad then argue for that or whatever on that pr but this is such a half baked "compromise" idea that doesnt benefit anyone.

ivanmixo:
Other PR was merg'd

PR #14940, opened 2024-01-28
No lobby wait time by default

XElectricX:
## About The Pull Request

Simply makes the lobby wait time 0 by default. Should have no effect on any live server since the config overrides the default entries.

## Why It's Good For The Game

A giant thorn in my side, and probably other developers. No more having to type out the starting early command.

## Changelog
:cl:
config: The lobby wait time is 0 by default
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -158,7 +158,7 @@ Administrative related.
 /datum/config_entry/flag/looc_enabled
 
 /datum/config_entry/number/lobby_countdown
-	config_entry_value = 180
+	config_entry_value = 0
 
 /datum/config_entry/number/mission_end_countdown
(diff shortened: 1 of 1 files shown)


ivanmixo (changes requested):
This is uhhhh... what if I want to test something in lobby? Bit guh. Also you're editing the fuckin code instead of the actual config txt file what is wrong with you

XElectricX:
> This is uhhhh... what if I want to test something in lobby? Bit guh. Also you're editing the fuckin code instead of the actual config txt file what is wrong with you

It doesn't do anything if you have something set in the config. This is just the default number used if there is no config entry.

ivanmixo:
Edit the txt instead

TiviPlus:
Yea this is useful by default just edit your own txt if you dont need it

PR #14946, opened 2024-01-29
MP-19 no longer fuller autos, now has burst again, has insane akimbo penalties.

Pariah919:
## About The Pull Request
Readds MP-19 Burst fire (5 round burst)
Removes fuller auto.
Adds insane akimbo penalties (10 akimbo additional delay.)

## Why It's Good For The Game
Akimbo mp19 was absolutely stupid but this definitely wasn't the proper way of getting rid of it. Just making the akimbo penalty so absurd you can't actually deal noticeable damage with it should be enough to kill that stupid play style anyway.

## Changelog

:cl:
balance: MP-19 now has burst fire again instead of auto burst, has insanely high akimbo penalties.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/smgs.dm
+++ b/code/modules/projectiles/guns/smgs.dm
@@ -75,14 +75,11 @@
 	aim_slowdown = 0.15
 	movement_acc_penalty_mult = 2
 
-	upper_akimbo_accuracy = 5
-	lower_akimbo_accuracy = 3
-
-	burst_amount = 1
-	autoburst_delay = 0.1 SECONDS
-	autoburst_delay = 0.1 SECONDS //this makes it fuller auto
-	burst_accuracy_bonus = -0.3
-	burst_scatter_mult = 25
+	upper_akimbo_accuracy = 12
+	lower_akimbo_accuracy = 9
+	burst_amount = 5
+	burst_delay = 0.1 SECONDS
+	akimbo_additional_delay = 10 // Literally do not even bother to try
 
 /obj/item/weapon/gun/smg/standard_machinepistol/compact
 	starting_attachment_types = list(/obj/item/attachable/foldable/t19stock, /obj/item/attachable/reddot, /obj/item/attachable/compensator, /obj/item/attachable/lasersight)


Lumipharon:
Its entirely fine as a solo weapon, you just need to actual wield it, and use the proper mods. Adding the higher rate of fire had nothing to do with nerfing akimbo, it was something I'd already discussed long before people started doing it.

srxtreme:
It was not fine as a solo weapon, it was a massive downgrade to previous version, with absurdly increased scatter in exchange for a marginal firerate boost compared to pre-nerf mp19 (10% to be exact). Even with all the attachments reducing scatter it was absolutely awful outside of point blank range at which it could not function due to having no survivability/defensive effects like say, shotguns, revolvers, vali, or absurd DPS like akimbo. It was a huge nerf to a very average weapon.

Kuro020 (approved):
I dont see anybody using this weapon, so trying to revive it like this is ok

BoltedAirlock:
> Its entirely fine as a solo weapon, you just need to actual wield it, and use the proper mods.

the current effective range on the mp-19 if you even dare try to use the autoburst mode  is a grand total of two tiles because it has a x25 scatter angle; using an rcomp and v-grip/gyro and wielding it will not extend the range it can hit anything no matter how hard you try because a x25 scatter angle kneecaps it into being buckshot but worse, and using the normal firemode just makes it a worse smg-90.

Lumipharon:
Yeah that's literally the intention. On normal mode it is the same rate of fire as a p90, only its the size of a pistol. The mp19 is literally a sidearm, p90 isn't.
Going fuller auto means having 50% higher dps than the larger p90, so there is a trade off. Otherwise it's better in every possible way except capacity.

Tommy-211:
its a decent secondary weapon

SimplySmiley123:
> Yeah that's literally the intention. On normal mode it is the same rate of fire as a p90, only its the size of a pistol. The mp19 is literally a sidearm, p90 isn't. Going fuller auto means having 50% higher dps than the larger p90, so there is a trade off. Otherwise it's better in every possible way except capacity.

then frankly maybe put it in the sidearms section of the vendor?
But then I guess it would leave the SMG section nigh empty.

PR #14949, opened 2024-01-29
Buffs tweezer removal time

Barnet2:
## About The Pull Request
Changes removal of each shrap from 10 seconds to 4, keeps 10 second fumble if you have no med skill
## Why It's Good For The Game
Shrap is a pretty guh mechanic, and barring adding CM shrap removal (unga can do it himself), having to wait 10 seconds in the middle of combat per mistake a teammate made is not ideal. Hopefully this makes it way less agonizing after the MG guns you down.
## Changelog
:cl:
balance: Tweezer takes 4 seconds to remove shrap instead of 10
/:cl:

### DIFF
--- a/code/game/objects/items/devices/tweezers.dm
+++ b/code/game/objects/items/devices/tweezers.dm
@@ -9,7 +9,7 @@
 
 /obj/item/tweezers/Initialize(mapload)
 	. = ..()
-	AddElement(/datum/element/shrapnel_removal, 10 SECONDS)
+	AddElement(/datum/element/shrapnel_removal, 4 SECONDS, 10 SECONDS)
 
 /obj/item/tweezers_advanced
 	name = "\improper ESR-12"


QualityVan (commented):
Strongly opposed to this. There's no reason for you to be doing it in the middle of combat in the first place - pull back to a safe distance if you want to treat non-basic damage, or just hold it until you can take a breather. Beyond that, if you want a faster solution then splint the limb. It's temporary, but suppresses however many pieces of shrap there are. If you want a faster and permanent solution, go to surgery.

Lumipharon:
Shrap exists to punish FF, the whole point is to require you to spend time sorting it out.

Barnet2:
> Strongly opposed to this. There's no reason for you to be doing it in the middle of combat in the first place - pull back to a safe distance if you want to treat non-basic damage, or just hold it until you can take a breather. Beyond that, if you want a faster solution then splint the limb. It's temporary, but suppresses however many pieces of shrap there are. If you want a faster and permanent solution, go to surgery.

Suppose I worded it wrong then, apologies. It is very easy to be interrupted by someone shuffling you (which happens very often, even in the backline); it's just compounded mistakes on mistakes. FF is already punishes by giving you damage you can't plan for, alongside shrapnel.

Arctrooper8:
> Shrap exists to punish FF, the whole point is to require you to spend time sorting it out.

Why does the punishment not fall upon the person causing shrap anyways? It's stupid that because someone else messed up you now have to get organ damage and wait around for a corpsman to pull it out.

QualityVan:
'I don't like being shuffled' isn't a reason to cut treatment time by more than half, and shrapnel can't actually cause organ damage.

PR #12756, opened 2023-04-20
Adds a proximity check for sentries to prevent stacking

lKiseki:
## About The Pull Request
Adds a proximity check in code for sentry deployment. If there is another sentry within two tiles of range, you will be unable to deploy it.

## Why It's Good For The Game
Prevents bullshit like this:
image

This, on the other hand, feels more fair for xenos:
image

## Changelog
:cl: Lewdcifer
balance: Sentries can no longer be deployed within two tiles of each other.
/:cl:

PR #14950, opened 2024-01-29
Adds a proximity check for sentries to prevent stacking, part 2: electric boogaloo

lKiseki:
## About The Pull Request
Per title. Revival of #12756.
Sentries cannot be deployed adjacent to one another.

## Why It's Good For The Game
Refer to #12756.

## Changelog
:cl: Lewdcifer
balance: Proximity check for sentries added to prevent stacking. Sentries can no longer be deployed adjacent to one another.
/:cl:

### DIFF
--- a/code/datums/components/deployable_item.dm
+++ b/code/datums/components/deployable_item.dm
@@ -82,6 +82,12 @@
 		if(user.do_actions)
 			user.balloon_alert(user, "You are already doing something!")
 			return
+		if(istype(item_to_deploy, /obj/item/weapon/gun/sentry))
+			for(var/turf/turf_to_check AS in RANGE_TURFS(SENTRY_DEPLOY_RANGE_LIMIT, location))
+				for(var/obj/machinery/deployable/mounted/sentry/deployable_sentry in turf_to_check)
+					if(get_dist_euclide(deployable_sentry, item_to_deploy) >= SENTRY_DEPLOY_RANGE_LIMIT)
+						user.balloon_alert(user, "Another sentry is too close")
+						return
 		user.balloon_alert(user, "You start deploying...")
 		user.setDir(newdir) //Face towards deploy location for ease of deploy.
 		if(!do_after(user, deploy_time, NONE, item_to_deploy, BUSY_ICON_BUILD))
--- a/code/__DEFINES/conflict.dm
+++ b/code/__DEFINES/conflict.dm
@@ -146,6 +146,9 @@
 #define ENERGY_OVERCHARGE_AMMO_COST 80
 #define ENERGY_OVERCHARGE_FIRE_DELAY 10
 
+//Define sentry
+#define SENTRY_DEPLOY_RANGE_LIMIT 1 // SQUARE range within which another sentry cannot be deployed.
+
 //Define stagger damage multipliers
 #define STAGGER_DAMAGE_MULTIPLIER 0.5 //-50% damage dealt by the staggered target after all other mods.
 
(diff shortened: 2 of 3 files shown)


code/datums/components/deployable_item.dm
```diff
@@ -82,6 +82,12 @@
 		if(user.do_actions)
 			user.balloon_alert(user, "You are already doing something!")
 			return
+		if(istype(item_to_deploy, /obj/item/weapon/gun/sentry))
+			for(var/turf/turf_to_check AS in RANGE_TURFS(SENTRY_DEPLOY_RANGE_LIMIT, location))
+				for(var/obj/machinery/deployable/mounted/sentry/deployable_sentry in turf_to_check)
+					if(get_dist_euclide(deployable_sentry, item_to_deploy) >= SENTRY_DEPLOY_RANGE_LIMIT)
```
DeltaFire15, line 88: doesn't this make you unable to deploy a sentry *unless* another is adjacent?

DeltaFire15 (changes requested):

FosterGrime-Heymynameslouie:
is okey at first since this seems like its just 1 tile spacing from eachother, major problems start to arise when you consider small maps, and small holds like miners and canter (I think thats what the crash ship is called), the other sentry just gets unused, or worse, a engineer tries to figure out how to fit all of it in, dying in the process as they are being mauled without sentry cover

plus I whole heartedly believe that this is just a stepping stone pr and 2 months down the line someones ognna make a pr that makes it so that the 1 tile range becomes 5 tile because fob seighe cope

code/datums/components/deployable_item.dm
lKiseki: Doesn't seem to be the case!

Ragecrabbo:
This is unironically a buff since sentry explosions damage other sentries and a single tile does nothing meaningful to prevent sentry stacking. 
Regardless im not a fan of doing balance changes by adding limits like this and i think a better solution would be to make them more vulnerable to explosions or even upgrading the explosion damage "tier" of a sentry blowing up or even nerfing sentry damage outright but giving them some sort of unique debuff that cannot be stacked, thus making it a better strategy to spread sentries out.

SimplySmiley123:
> This is unironically a buff since sentry explosions damage other sentries and a single tile does nothing meaningful to prevent sentry stacking. Regardless im not a fan of doing balance changes by adding limits like this and i think a better solution would be to make them more vulnerable to explosions or upgrading the explosion damage "tier" of a sentry blowing up or even nerfing sentry damage outright but giving them some sort of unique debuff that cannot be stacked, thus making it a better strategy to spread sentries out.

Warlocks already cheese the hell out of sentries they don't need to have it any easier...

RagingLemons1:
evenfunnierthe2ndtime

TellyLsc:
Boilers and banelings don't exist to gas the sentries?

ivanmixo:
Yeah my opinion is basically this is in effect a buff except in one case (tad) since it *forces* proper sentry stacking. Imo either go harder on the distance allowed between them or increase the explosion size when a sentry dies. Do consult kurt beforehand, though.

code/datums/components/deployable_item.dm
Autisem: this code feels kinda hacky in general, could this not be part of the sentry code instead of all deployables? also you can use locate(sentry) to get all sentrys instead of checking every turf for a sentry manualy

PR #14960, opened 2024-01-30
Revert "Fix flamer back tank disconnecting with magharness"

Barnet2:
Reverts tgstation/TerraGov-Marine-Corps#14826
TM only, mag harness still randomly fails without the delay for ~2 weeks. UnTM when it is fixed


Lumipharon (approved):
There is no fix in sight and this is actually a very large bug, so going to merge.
Someone is welcome to try re-pr the original change and fix the issue however.

PR #14986, opened 2024-02-02
Deletes examinate keybind.

Xander3359:
## About The Pull Request
Removes the keybind for "Examinate"
Puts it in our click code chain instead.

## Why It's Good For The Game
Keybinds are weird, they override our click code chain.
So if you have this keybinded to shift (like default), then it fucks with other shift clicks.

## Changelog

:cl:
del: Deletes "Examinate" Keybind
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -402,10 +402,6 @@ if(selected_ability.target_flags & flagname && !istype(A, typepath)){\
 			return FALSE
 		if(COMSIG_MOB_CLICK_HANDLED)
 			return TRUE
-
-	if(examinate(A))
-		return
-
 	return A.ShiftClick(src)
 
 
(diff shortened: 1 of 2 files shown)


TellyLsc:
Why? specialclick can be rebinded while the click code stuff can't, right? What if I want to examine items by holding shift+click on items which is only possible because it's rebindable?

code/_onclick/click.dm
```diff
@@ -402,6 +402,10 @@ if(selected_ability.target_flags & flagname && !istype(A, typepath)){\
 			return FALSE
 		if(COMSIG_MOB_CLICK_HANDLED)
 			return TRUE
+
+	if(examinate(A))
```
ivanmixo, line 406: Put it in `ShiftClick` instead

ivanmixo (changes requested):

ivanmixo (approved):

PR #14994, opened 2024-02-03
Adjust psylance stagger

Lumipharon:
## About The Pull Request
Psylance only staggers for 1 second instead of 4.
## Why It's Good For The Game
4 seconds was never intended, it was a left over from before stagger was refactored to actually count duration accurately.
## Changelog
:cl:
balance: Psylance stagger reduced to 1 second
/:cl:


ivanmixo (approved):
very cool

PR #14996, opened 2024-02-03
Stagger doesn't stack in duration anymore

ivanmixo:
## About The Pull Request

Title. Stuff like this won't happen anymore.

https://github.com/tgstation/TerraGov-Marine-Corps/assets/22431091/1a0d6483-62b2-4d59-8186-37ca5ae7d48d
## Why It's Good For The Game

Stagger stacking like this is problematic due to us having a fair few weapons which fire way faster than the stagger dissipates, which results in xenos being staggered for over 10 seconds after the fight in some cases. I'd rather we make it not stack at all instead of nerfing those weapons to the ground, since it'd take away the point of a fair few of them (not to mention its hard maintainability wise, needing to make sure every gun is below a certain stagger duration to avoid silly stuff)

Also kurt said yes
## Changelog
:cl:
balance: Stagger doesn't stack in duration anymore
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -167,19 +167,19 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	if(iscarbon(victim))
 		var/mob/living/carbon/carbon_victim = victim
 		#if DEBUG_STAGGER_SLOWDOWN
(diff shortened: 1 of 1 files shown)


Kuro020 (approved):

Lumipharon:
This prevents ALL increase in stagger duration. Just make a separate proc like all the other status effects. One that only adds if higher than current duration, and one that adds to current duration.

As this means stagger from other sources don't stack. I.e. human stagger from pain, xeno abilities etc, when it sounds like you have an issue specifically with guns stacking stagger.

efzapa:
# Cannot while staggered

Lumipharon (approved):

PR #15002, opened 2024-02-04
Locks tadpole to leader access, reduces PO slot by one.

Barnet2:
## About The Pull Request
What it says on the tin. Leader access includes squad leaders and FC, and leader access can be given to ID's in CIC.
## Why It's Good For The Game
This aims to remedy two issues:
1. Picking PO when there is already one, gambling what job you are actually doing.
2. Hopefully less Tadpole grief.
Number 1 is fairly self explanatory, if there is only 1 PO slot they are expected to be doing CAS. If someone else steals CAS (which is ALOT rarer than someone taking Tadpole), that is already an admin issue. This fixes one admin issue as well.
Number 2 is a bit more complicated. Tadpole throwing an entire round is inherent to it's flawed design, there is no fixing that. However, PO anecdotally is a hotspot for new players, SL and FC much less so. This hopes to reduce the amount of people with sub 100 hours trying out the tadpole without any relevant game knowledge. If you want a non-leader to pilot the Tadpole, you have to very willingly give them access in CIC which has a degree of trust to it, unlike PO being selected from the get-go. On 30 pop there should be atleast one leader, and if there is none the players already have to be given a captain ID for req and such by the admins which is a separate issue.
## Changelog
:cl:
balance: Removes one Pilot Officer slot, makes the Tadpole only be able to be piloted by those with Leader access (SL, aSL, FC's)
/:cl:

### DIFF
--- a/code/datums/gamemodes/nuclear_war.dm
+++ b/code/datums/gamemodes/nuclear_war.dm
@@ -8,7 +8,7 @@
 		/datum/job/terragov/command/captain = 1,
 		/datum/job/terragov/command/fieldcommander = 1,
 		/datum/job/terragov/command/staffofficer = 4,
-		/datum/job/terragov/command/pilot = 2,
+		/datum/job/terragov/command/pilot = 1,
 		/datum/job/terragov/engineering/chief = 1,
 		/datum/job/terragov/engineering/tech = 2,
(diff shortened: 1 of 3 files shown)


Barnet2:
Maints are okay with splitting the role and giving tad PO a higher timelock, so that will be done instead

PR #15005, opened 2024-02-04
Reduce time dilation from chat window messages

MrStonedOne:
This proc is the top of the list for overhead and cost during high lag moments.

This reduces the amount of associated list scans from 3 `logn` scans and 1 `n` scan to 2 `logn` scans.

`in` is a major code smell because it does everything the harder way.

the final solution to speeding up this proc would be to move all of this to the client datum or the player_info datum if we need it to persist between reconnections (assuming tgmc has the player_info datum)


Xander3359:
The king blesses us with performance PRs.
Praise the almighty

Lumipharon (approved):

PR #15011, opened 2024-02-05
Mine alpha fix

Lumipharon:
## About The Pull Request
Normal claymores had the lower alpha meant only for AT mines.
I am dum and messed up the typepath.
## Why It's Good For The Game
After playing against sneaky claymores in HvH, I actually like them, so would be happy for them to stay. It was an unintended HvX balance change though, so would like thoughts on this.
## Changelog
:cl:
fix: fixed normal claymores being sneakier than intended
/:cl:


ivanmixo (approved):
Transparent stuff is cancer and claymores are already sneaky enough

PR #15030, opened 2024-02-08
Beacon Auto update when changing Z-level instead of turning off

AusarAtkin:
## About The Pull Request
Change how Deployed Beacon on Tadpole work back to #14421.

Now it Auto update without needing to Redeploy it every time tad move, and can be deployed in Tad on whatever Z-Level tad is (but still can only be used groundside). Code thanks to lumi

It filter out non groundside beacon from the supply console list, so no clutter. Code thanks to ivan and electric.
## Why It's Good For The Game
One less thing to think about for Tadpole Pilot and ROs, while maintaining beacon usage only to groundside use.
## Changelog
:cl:
qol: Supply Beacon now Auto Update if Tad are moving across Z-level.
/:cl:

### DIFF
--- a/code/datums/components/beacon.dm
+++ b/code/datums/components/beacon.dm
@@ -83,11 +83,6 @@
 ///Activates the beacon
 /datum/component/beacon/proc/activate(atom/movable/source, mob/user)
 	var/turf/location = get_turf(source)
-	if(!is_ground_level(location.z))
-		to_chat(user, span_warning("You have to be on the planet to use this or it won't transmit."))
-		active = FALSE
-		return FALSE
-
 	var/area/A = get_area(location)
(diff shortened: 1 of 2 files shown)


ivanmixo:
You didnt do as I asked, and this leaves essentially dead signals in the req console. Buh.

AusarAtkin:
> You didnt do as I asked, and this leaves essentially dead signals in the req console. Buh. 

You asked for it to be able to be deployed everywhere for consistency, and still not be able to be used non ground side. 

This is the most i can do, and i can't even do this without Lumi's help.

ivanmixo:
It clutters the list

Lumipharon (approved):

PR #15036, opened 2024-02-08
Removes larva generation from psychic drain and cocoon opening

Barnet2:
## About The Pull Request
What it says on the tin.
## Why It's Good For The Game
_Preface_
Let's get started with what happens when a member on either side of the team dies and has their body secured.
Marine: Drained / Cocoon
- 20 cloneloss (pretty moot)
- If perma or DNR, -1 marine
- Psy points scaling on population, which can be used to purchase a plethora of things
- Larva points (the issue I will speak about)

Xeno: Sold
- Minus 1 xeno
- Req points

With the merge of #14927, I believe now is an appropriate time to look at snowballing on both sides. Though marine requisitions will remain superior to the Blessing tab as of now, psy draining is usually more accessible and has the benefit of **population snowballing,** which I believe is the most powerful snowballing you can get; thus the PR. This is also compounded by the fact xenos have a rather high regeneration rate with silo, alongside being able to spend psy points on getting more xenos (more silos). Primos are also very strong, too. My argument comes down to "there is too much in the basket of drain/cocoon" given the other things it gives. I would enjoy any critiques in the comment as I might just be schizo here.
## Changelog
:cl:
balance: Removed larva generation on psy drain and cocoon
/:cl:

### DIFF
--- a/code/datums/gamemodes/_game_mode.dm
+++ b/code/datums/gamemodes/_game_mode.dm
@@ -444,12 +444,8 @@ GLOBAL_LIST_INIT(bioscan_locations, list(
 		parts += "[GLOB.round_statistics.ravager_rages] number of times Ravagers raged."
 	if(GLOB.round_statistics.hunter_silence_targets)
 		parts += "[GLOB.round_statistics.hunter_silence_targets] number of targets silenced by Hunters."
-	if(GLOB.round_statistics.larva_from_psydrain)
-		parts += "[GLOB.round_statistics.larva_from_psydrain] larvas came from psydrain."
 	if(GLOB.round_statistics.larva_from_silo)
 		parts += "[GLOB.round_statistics.larva_from_silo] larvas came from silos."
-	if(GLOB.round_statistics.larva_from_cocoon)
-		parts += "[GLOB.round_statistics.larva_from_cocoon] larvas came from cocoons."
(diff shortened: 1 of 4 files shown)


blackdav123:
I agree that the snowballing of it should be removed but as it currently stands drains and cocoons are the only way to make a comeback population wise if you lose your silo, because even if you do wipe half of the marines that rushed silo you may still be down a large amount of xenos that will take an hour to be replaced by a new silo.

A compromise that I think would work would be making the larva points given by drains/cocoons scale down to zero with how many xenos you have so that it isnt a win more mechanic and still can serve the role to provide a comeback.

Barnet2:
> I agree that the snowballing of it should be removed but as it currently stands drains and cocoons are the only way to make a comeback population wise if you lose your silo, because even if you do wipe half of the marines that rushed silo you may still be down a large amount of xenos that will take an hour to be replaced by a new silo.
> 
> A compromise that I think would work would be making the larva points given by drains/cocoons scale down to zero with how many xenos you have so that it isnt a win more mechanic and still can serve the role to provide a comeback.

That is actually a good idea. I'll bring it up with Kuro.

Barnet2:
There are some better alternatives to this PR as discussed with maints, so closed.

PR #15051, opened 2024-02-10
adds the SH49 from RU

SomrNetwork:
## About The Pull Request
requested by CVS_Pharmacy
"it has a pretty decent firerate, with a low shell capacity and low damage to balance it out
it has some pretty nice sounds and I think the community would like those"
## Why It's Good For The Game
"an ACTUAL semi auto shotgun that isn't slow as shit would be really nice, and this one is pretty bad in terms of damage and knockback so it's really just a gimmicky weapon and not something that could actually hold it's own ground" - CVS_Pharmacy
## Changelog
:cl:
add: sh49
soundadd: sh46 sounds
/:cl


SomrNetwork:
so thats what code: does

KerbalCitizen:
You should probably write somewhere that you're removing the double barreled shotgun, SH34. 
Also seems pretty spicy with slugs. 300 DPS, 30 AP, 9 sunder per second. Granted it only holds 5 rounds, but still.

SomrNetwork:
> You should probably write somewhere that you're removing the double barreled shotgun, SH34. Also seems pretty spicy with slugs. 180 DPS, 30 AP, 9 sunder per second. Granted it only holds 5 rounds, but still.

the DB isn't removed just moved down on the shotgun list, also I just copy and pasted code from RU :trollface:

KerbalCitizen:
Ah yeah I see the DB now. Editted my numbers in the previous comment since I kept on messing up the math. 0.6 damage (and sunder) multiplier on the gun. 0.3 firedelay, so 10/3 shots per second.

code/game/objects/machinery/vending/marine_vending.dm
```diff
@@ -57,15 +57,16 @@
...
+			/obj/item/weapon/gun/shotgun/combat/shq6 = -1,
 			/obj/item/ammo_magazine/shotgun = -1,
 			/obj/item/ammo_magazine/shotgun/buckshot = -1,
 			/obj/item/ammo_magazine/shotgun/flechette = -1,
 			/obj/item/ammo_magazine/shotgun/tracker = -1,
 			/obj/item/weapon/gun/rifle/standard_autoshotgun = -1,
 			/obj/item/ammo_magazine/rifle/tx15_flechette = -1,
 			/obj/item/ammo_magazine/rifle/tx15_slug = -1,
+			/obj/item/weapon/gun/shotgun/double/marine = -1,
+			/obj/item/storage/holster/belt/ts34/full = -1,
```
Xander3359, line 69: Any particular reason why this is being moved?

SomrNetwork: probably moved for sorting and clutter

SomrNetwork: image
basing off of this

Xander3359: Just because RU makes a change, doesn't mean it also has to be made here.
That's dumb, moving it just for the shits and giggles.

CVSthePharmacy:
I like this pr it is very nice

fe-ed:
:pog:

Pariah919:
What am I even looking at? A shotgun with no stagger/stun (which is whatever really), with 0.03 fire rate (since it isn't set to SECONDS) so it has insane DPS.
What are you even trying to do here? We have the autoshotgun and the combat shotgun already, even if you made this balanced it'll just be a really scuffed double barrel.

CVSthePharmacy:
> What am I even looking at? A shotgun with no stagger/stun (which is whatever really), with 0.03 fire rate (since it isn't set to SECONDS) so it has insane DPS. What are you even trying to do here? We have the autoshotgun and the combat shotgun already, even if you made this balanced it'll just be a really scuffed double barrel.

I knew I'd get one of these... no it really is it's own weapon, the dps would be absolutely shit, I know that because I've personally tried it in RU, and with such a small capacity it's barely even good against a drone, and that's IF you hit all shots

TiviPlus:
Author tried to trick me into thinking this pr was approved by kuro when it was not

CVSthePharmacy:
I was not the author...

PR #15054, opened 2024-02-10
Human v Human melee tweaks and fixes

Lumipharon:
## About The Pull Request
Fixes choking someone to death... not causing any damage.
Fixes weapon miss chance never actually missing.
Fix some runtimes if you miss an attack.

Upgrading your grab now is faster with higher CQC skill.
When punching someone who is prone, you now always do max damage instead of the normal rng damage.
Unarmed attacks now uses the same target zone prob stuff as melee weapons, with the minor exception of it comparing cqc skills of the attacker and defender to modify miss chance.
The effect of this means unarmed attack will generally actually hit the target area you aim for instead of a complete rng shitshow, although some times you can now miss entirely.

Also replaced some one letter var shit.
## Why It's Good For The Game
Fixes and minor cqc buffs to make it mildly less useless
## Changelog
:cl:
balance: Unarmed damage against humans is no longer rng if your victim is prone
balance: CQC skill lowers grab upgrade time
balance: Unarmed attacks against humans uses the same zone targeting as melee weapons
fix: Choking someone to death once again actually causes oxyloss
fix: Melee attack can now actually miss some times as intended (only for humans attacking humans)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -76,8 +76,8 @@
 		return
 
 	if(pulledby?.grab_state >= GRAB_KILL)
-		Losebreath(3)
-		adjustOxyLoss(3)
+		Losebreath(1)
+		adjustOxyLoss(4)
 	else if(losebreath <= 10)
 		adjust_Losebreath(-1) //Since this happens before checking to take/heal oxyloss, a losebreath of 1 or less won't do anything.
 	else
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -78,10 +78,10 @@
 	if(pulledby?.grab_state >= GRAB_KILL)
 		Losebreath(1)
 		adjustOxyLoss(4)
-	else if(losebreath <= 10)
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -75,7 +75,10 @@
...
 		return
 
-	if(losebreath <= 10)
+	if(pulledby?.grab_state >= GRAB_KILL)
+		Losebreath(1)
+		adjustOxyLoss(4)
+	else if(losebreath <= 10)
 		adjust_Losebreath(-1) //Since this happens before checking to take/heal oxyloss, a losebreath of 1 or less won't do anything.
 	else
 		set_Losebreath(10) //Any single hit is functionally capped - to keep someone suffocating, you need continued losebreath applications.
```
QualityVan, line 84:
```suggestion
	if(pulledby?.grab_state >= GRAB_KILL)
		Losebreath(1)
		adjustOxyLoss(4)
	else if(losebreath > 10)
		set_Losebreath(10) //Any single hit is functionally capped - to keep someone suffocating, you need continued losebreath applications.
	else if(losebreath > 0)
		adjust_Losebreath(-1) //Since this happens before checking to take/heal oxyloss, a losebreath of 1 or less won't do anything.
```
I got annoyed by this using extra adjust calls and pushing losebreath to negative lots uselessly.

PR #15056, opened 2024-02-10
Fixes xenos with non-standard wound handling not showing wounds in special states.

Helg2:
## About The Pull Request
It fixes defender, baneling, bull, warrior, crusher and behemoth not showing their wounds in special states, such as crest defense or running.
## Changelog
:cl:
fix: Xenos now show their wound in their special states.
/:cl:


Lumipharon:
To add to this, some xenos simply don't have wound overlays for special states. Running queen for example.

Helg2:
> I always thought this was intentional for defender's fortify. It can trick marines into disengaging or picking the wrong target because they don't know how hurt you are. Maybe someone should make it that way deliberately.

It remains that way, because we have no sprites xd.

Lumipharon (approved):

PR #15059, opened 2024-02-11
Increases BR QB to 2300 from 1800

Barnet2:
## About The Pull Request
Title
## Why It's Good For The Game
BR is a deceptively big map and often basically all of the caves besides silo goes unmazed if colony is mazed, allowing people to just run through it.
## Changelog
:cl:
balance: Big Red now has 2300 quickbuild points instead of 1800
/:cl:

### DIFF
--- a/_maps/bigred_v2.json
+++ b/_maps/bigred_v2.json
@@ -5,7 +5,7 @@
 	"disk_sets": {
 		"basic": 1
 	},
-	"quickbuilds": 1800,
+	"quickbuilds": 2300,
 	"announce_text": "A second generation colony has had a beacon transmitting the same signal, nonstop. Attempts to hail the colony over comms have proved futile. Because the ship was at a nearby drydock, it has been dispatched to figure out what's wrong. TGMC, prepare to deploy!",
 	"traits":[{
 		"weather_sandstorm": true


yyzsong:
xalt pr?

Breadeagle:
I said it in game but I'm unsure if Big Red needs more QB, but I'm quite sure it doesn't need THIS much extra QB. Most maps should probably have some gaps in quickbuild, for better or worse. Daedalus for example, I think has too much quickbuild for what it is, which contributes to the uneven feeling.

Barnet2:
> I said it in game but I'm unsure if Big Red needs more QB, but I'm quite sure it doesn't need THIS much extra QB. Most maps should probably have some gaps in quickbuild, for better or worse. Daedalus for example, I think has too much quickbuild for what it is, which contributes to the uneven feeling.

It is better to have more QB than to have too little. This is because there is usually a critical mass of maze that cannot grow any thicker lest it turns into a brick, which just cucks xenos. Too little is guh as well; leads to uncovered areas. This is partly due to the fault of BR being a modular map so the amount you need to maze varies wildly.

Breadeagle:
Honestly, thinking about it, I think it'd just be better if xenos got enough qb for pretty much all the map but marines started with better consumable wall destruction at the start. Like if cluster could consistently kill resin walls, then give them a bit more OB fuel + shells. The thought kind of came from the realization that it is only really players like Rio or myself who consistently maze paths post-shutters with slow build.

TiviPlus (approved):

PR #15067, opened 2024-02-11
Adds a cooldown to re-resting

yyzsong:
## About The Pull Request
After getting up from rest as a xenomorph, you will not be able to rest again for one second.
this PR dedicated to tgmc round 25705 (salt pr)
## Why It's Good For The Game
The unrest changes in #14898 means that there is no feedback on how long you have before you can get up, so if you need to get up in an emergency you will probably be spamming the 'get up' button, which will result in immediately sitting back down after getting up, which is kind of stinky
## Changelog
:cl:
qol: Adds a cooldown between unresting and then resting again for xenomorphs
/:cl:


code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -475,12 +475,17 @@ Returns TRUE when loc_weeds_type changes. Returns FALSE when it doesn’t change
 	if(!COOLDOWN_CHECK(src, xeno_resting_cooldown))
 		balloon_alert(src, "Cannot get up so soon after resting!")
 		return
+	if(!COOLDOWN_CHECK(src, xeno_unresting_cooldown))
+		balloon_alert(src, "Cannot rest so soon after getting up!")
+		return
```
Lumipharon, line 480: you're just checking both regardless of whether you're resting or not resting?

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -475,12 +475,17 @@ Returns TRUE when loc_weeds_type changes. Returns FALSE when it doesn’t change
...
 	return ..()
 
 /mob/living/carbon/xenomorph/set_resting()
 	. = ..()
 	if(resting)
 		COOLDOWN_START(src, xeno_resting_cooldown, XENO_RESTING_COOLDOWN)
+	if(!resting)
```
Lumipharon, line 487: this is unneeded, you're already checking if you're resting above.

Lumipharon (changes requested):

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
yyzsong: There's not really a reason to check if I'm resting or not

Lumipharon: One is for resting, the othe is for unresting. Checking them both is by definition bad.

code/modules/mob/living/carbon/xenomorph/xenomorph.dm
```diff
@@ -471,16 +471,22 @@ Returns TRUE when loc_weeds_type changes. Returns FALSE when it doesn’t change
...
 	if(!resting_action || !resting_action.can_use_action())
 		return
+	if(resting)
+		if(!COOLDOWN_CHECK(src, xeno_resting_cooldown))
+			balloon_alert(src, "Cannot get up so soon after resting!")
+			return
 
-	if(!COOLDOWN_CHECK(src, xeno_resting_cooldown))
-		balloon_alert(src, "Cannot get up so soon after resting!")
+	else if(!COOLDOWN_CHECK(src, xeno_unresting_cooldown))
```
Lumipharon, line 479:
```suggestion
	if(!COOLDOWN_CHECK(src, xeno_unresting_cooldown))
```

code/__DEFINES/mobs.dm
```diff
@@ -480,6 +480,7 @@ GLOBAL_LIST_INIT(xenoupgradetiers, list(XENO_UPGRADE_BASETYPE, XENO_UPGRADE_INVA
 #define XENO_GARGOYLE_DETECTION_COOLDOWN 30 SECONDS
 #define XENO_GARGOYLE_DETECTION_RANGE 10//How far gargoyles can detect hostiles
 #define XENO_RESTING_COOLDOWN 2 SECONDS
+#define XENO_UNRESTING_COOLDOWN 1 SECONDS
```
Lumipharon, line 483:
```suggestion
#define XENO_UNRESTING_COOLDOWN 0.5 SECONDS
```

Lumipharon: shortening this because 1 second is actually quite long, considering this is just to cover misclicks

Lumipharon (approved):

PR #15071, opened 2024-02-12
Reverts the Alamo Xeno TGUI change

DeltaFire15:
Reverts tgstation/TerraGov-Marine-Corps#15043

Apparently the Alamo UI is broken for xenos rn, and I can't be bothered to fix the UI myself, so until it gets improved, this should work in the interrim.


ivanmixo (approved):
@Xander3359

PR #15075, opened 2024-02-12
Bull (and Crusher) no longer gets stunned by projectiles/rocket sledge while charging

usnpeepoo:
## About The Pull Request

bull no longer gets stunned by projectiles such as buckshot/slugs while charging. This also affects crushers charge but only in the case of being stunned during charges (only thing I can think of is a mid charge rocket sledge hit)

## Why It's Good For The Game

caste is absolute shit because it just dies on the charge back if it overextends even slightly, immensely inferior to crusher and you're better off playing a different caste now that forced evo paths are gone before playing the better xeno (crusher).
Admittedly this is a band aid fix, there's a lot more problems with the caste that would entail a rework of its abilities/where bull can stand as a t2 however this pr should alleviate a piece of bull being terrible and a throw t2 to evolve to

## Changelog

:cl:
balance: Bull (and Crusher) no longer gets stunned by projectiles/rocket sledge while charging
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -92,6 +92,7 @@
 #define TRAIT_STASIS "stasis"//Subject to the stasis effect
 #define ENDURE_TRAIT "endure" //Ravager Endure ability.
 #define RAGE_TRAIT "rage" //Ravager Rage ability.
+#define CHARGE_TRAIT "charge_stun_immunity" //Charge Stun Immunity
 #define UNMANNED_VEHICLE "unmanned"
 #define STEALTH_TRAIT "stealth" //From hunter stealth
 #define REVIVE_TO_CRIT_TRAIT "revive_to_crit"
--- a/code/modules/mob/living/carbon/xenomorph/castes/bull/bull.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/bull/bull.dm
@@ -10,6 +10,7 @@
 	plasma_stored = 200
 	tier = XENO_TIER_TWO
 	upgrade = XENO_UPGRADE_NORMAL
+	mob_size = MOB_SIZE_BIG
 
 	pixel_x = -16
 	pixel_y = -3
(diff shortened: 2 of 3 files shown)


Kuro020:
> Stunning bulls with slugs/maybe buck is the only way to deal with them without them vanishing offscreen at mach 2. Other than spamming aimmode and waiting for the bull to run into a wall of bullets.
> 
> Vali spear with tram used to be the other way to deal with bulls...

the caste is absolute trash beyond trash, removing one of its main counters is something I'm ok with pending a rework

Kuro020 (approved):

TheWiznard:
just make it so bull cant charge if its on fire then in exchange for stun immunity

usnpeepoo:
> merges
> 
> bull rework never happens/is shelved because nobody knows how to balance bull
> 
> 
> 

I'm sure something will be figured out

Lumipharon (changes requested):
changing mob size does more than just give you immunity to some stuns.

Edit: posting here for clarity.
Mob size effects gun accuracy, it changes certain throw interactions, lets you break reinforced miners, effects some shuffle stuff, directly impacts charge interactions, probably some other stuff.

If you are specifically looking to add stun immunity (always/when charging?) mob size is not the right thing.

Lumipharon:
this also applies to crusher (and I assume queen).

usnpeepoo:
> This also applies to crusher, queen and bemo.

crusher/queen are already mob_size_big so stuns don't even affect them to begin with, and with the stun immunity being on charge only it's only really a benefit for bull (queen needs primo, crusher already stun immune, rarely gets hit with sledge mid charge)

Lumipharon:
Mob size big does not actually make you stun immune. SOME stuns specifically check mob size, but not all.

usnpeepoo:
> reopen

not sure whats required of the pr now so im not reopening it

lbnesquik:
Lumi wanted you to use a different mechanism to achieve the outcome you want, because mobsize does more than just what you want. Changing that would make your PR up for merge talks again.

Lumipharon:
this

usnpeepoo:
> this

I'm not using mob_size anymore though 😭 it's adding a trait (anti stun) when they are actively charging and removes it when they stop charging. I was just waiting for a response if it's that's good enough

Lumipharon:
oh, shouda pinged me or something, I didn't see it had been updated

Lumipharon:
As mentioned, the traits need to be specific to bull

PR #15080, opened 2024-02-12
Synthetic critical condition changes

Neocloudy:
## About The Pull Request
Synthetics/Early Synthetics will no longer enter critical condition, but instead, at -30 health, you start to overheat rapidly. I've tried to make it so you can still self repair to delay your death or even stop overheating, but not getting repaired can quickly kill you. 
The numbers may need tuning, but my reason is the burn mods for synths are fucked up real good. Or I'm just not mathing right.
image
If the balance gods want I could just make it so they still enter crit but take a lot of burn damage so you can't be dragged by xenos or stuck in crit. But I want to try this option first.
## Why It's Good For The Game
Synths slumping over the moment they hit -50 health is pretty boring, along with the situations where your body could be hidden by xenos or you could be stuck in crit for a long time.

Synthetics not falling over in crit, in exchange for both taking a lot of burn and also having to repair all that burn if they ever make it out should be more interesting than immediately falling over at -50 health.
## Changelog
:cl:
balance: late synths have a 20% burn reduction instead of 30%
balance: synths/early synths will no longer be unconscious at -50 health
balance: synths/early synths take rapid burn damage at -30 health
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -629,7 +629,7 @@
 
 /datum/species/early_synthetic/handle_unique_behavior(mob/living/carbon/human/H)
 	if(H.health <= -30 && H.stat != DEAD) // Instead of having a critical condition, they overheat and slowly die.
-		H.adjustFireLoss(rand(15, 25)) // This may need tweaks
+		H.adjustFireLoss(rand(18, 28)) // This may need tweaks
 		if(prob(8))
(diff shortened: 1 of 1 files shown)


coldironwarrior:
would like this. i dont even see synths get dragged that often, but there sure is a lot of "sitting in crit forever waiting for someone to notice me to weld me or slash me and put me out of my misery", and this would fix that too.

Kuro020 (approved):

PR #15090, opened 2024-02-13
Removes and moves Ice Colony phoron miners

blackdav123:
## About The Pull Request
This removes 7 of Ice Colony's 16 phoron miners and places 1 new one.
Ice Colony Miners
## Why It's Good For The Game
Ice Colony has the highest number of phoron miners out of all of our maps, beating even Big Red's 13 miners. This map is already marine sided with its numerous open fields and choke points, but it gives marines an extreme snowball advantage just for taking the areas they would be taking to regardless of miners. Unlike Big Red, many of Ice Colony's miners are much harder to siege and many of them also help defend other miners, which creates a compounding problem where marines will have their full force at the southern disks which threatens a silo rush, while 6 phoron miners are ticking away and covering each others flanks up at the open fields near engi. 

My decisions on which miners to specifically remove are mostly opinion based, but I tried to take out the miners that I thought had the least interesting fights over them.
## Changelog
:cl:
balance:  Removed 7 phoron miners from Ice Colony, Added 1.
/:cl:

### DIFF
--- a/_maps/map_files/Ice_Colony_v2/Ice_Colony_v2.dmm
+++ b/_maps/map_files/Ice_Colony_v2/Ice_Colony_v2.dmm
@@ -21354,10 +21354,6 @@
 	dir = 8
 	},
 /area/ice_colony/surface/storage_unit/power)
-"hqp" = (
-/obj/machinery/miner/damaged,
-/turf/open/floor/plating/ground/snow/layer1,
-/area/ice_colony/exterior/surface/valley/southeast)
 "hqr" = (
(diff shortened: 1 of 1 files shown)


ivanmixo:
wonder when someone is gonna do something about the plat that isnt even underground

blackdav123:
> wonder when someone is gonna do something about the plat that isnt even underground

the plat itself is entirely underground but it is pretty close to surface

i may try changing it to a phoron or removing it entirely depending on how ice feels after this pr

PR #15100, opened 2024-02-14
Some recipes which use 1u of Lemoline now add to the final volume

Xander3359:
## About The Pull Request
The following chems now make 1u more per reaction (using lemoline):
- Meralyne
- Dermaline
- Ryetalyn
- Synaptizine
- Hyronalin
## Why It's Good For The Game
It was dumb that mixing 40u of chems would give 30u of reagent.
I really dislike lemolock, but if it remains in, I think it actually adding to the final recipe is fair.
## Changelog
:cl:
balance: Some lemolocked chems have a bit more generous reactions.
/:cl:

### DIFF
--- a/code/modules/reagents/reactions/medical.dm
+++ b/code/modules/reagents/reactions/medical.dm
@@ -22,7 +22,7 @@
 
 /datum/chemical_reaction/dermalime
 	name = "Dermaline"
-	results = list(/datum/reagent/medicine/dermaline = 3)
+	results = list(/datum/reagent/medicine/dermaline = 4)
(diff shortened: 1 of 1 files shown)


Looking4bros:
Yeah, this'd be nice, it'd put Mera/Derma into the magic number threshold of 120 instead of 90 per bottle. Also nice little QoL thing for Rye since it's SO expensive Lemoline wise.

TiviPlus (approved):

TiviPlus:
@Kuro020

PR #15114, opened 2024-02-16
Medkit now fits in corpsman backpack/satchel

Xander3359:
## About The Pull Request
Corpsman backpack/satchel can hold 1 medkit
## Why It's Good For The Game
Medics spawn with a useless medkit
I think it should be possible to bring it with you.
This might require some balance since it's giga slot efficient, but I think it should either be balanced+allowed in, or removed from their essentials kit.
## Changelog
:cl:
balance: Corpsman backpack/satchel can hold 1 medkit
/:cl:

### DIFF
--- a/code/game/objects/items/storage/backpack.dm
+++ b/code/game/objects/items/storage/backpack.dm
@@ -283,6 +283,8 @@
 	item_state = "marinepackm"
 	var/obj/item/cell/high/cell //Starts with a high capacity energy cell.
 	var/icon_skin
+	bypass_w_limit = list(/obj/item/storage/firstaid)
+	storage_type_limits = list(/obj/item/storage/firstaid = 1) //You can only hold 1 medkit
 
 /obj/item/storage/backpack/marine/corpsman/Initialize(mapload, ...)
 	. = ..()


SimplySmiley123:
On one hand, it would be too good if it was space efficient, but on the other hand if it were to take just as much space as what you could put in it, why bother?

Xander3359:
> On one hand, it would be too good if it was space efficient, but on the other hand if it were to take just as much space as what you could put in it, why bother?

Medkits holds 14w worth of space.
So it being maybe w_class of 7 would be reasonable or maybe some value around that much.

Ragecrabbo:
Neat idea. I doubt storage creep will be that much of an issue since you could always just take a few lifesaver belts wherever you go anyways or ask for more supplies from RO.

Arctrooper8:
Make it so you can store the medkit in your suit storage, if for some reason you just really really don't wanna bring a gun.

SimplySmiley123:
> Medkits holds 14w worth of space. So it being maybe w_class of 7 would be reasonable or maybe some value around that much. But unless the medkit gets w_class 14, it will be storage creep, no matter how you look at it.

Again, the problem is with the reasoning for why you would ever bother with putting the medkit in your backpack without even a slight benefit to it, but giving it any benefits for putting in your backpack would get it labeled with storage creep.

Lumipharon:
This is just giga storage creep.
If you want to bring it with you, you have a thing called a hand. The better question is whether there is any reason WHY medics spawn with one.

PR #15128, opened 2024-02-18
Adds xeno manifest to lobby

Barnet2:
## About The Pull Request
image
Only usable in lobby, doesn't use hive status as we don't want lobby viewers to see plasma/health/silo etc, same with marine manifest.
(p.s: i have no clue about html, if you think you can help make it look better please tell me)

## Why It's Good For The Game
Idea from CM, essentially an "weaker" equivalent to the manifest. If marines can see who they are playing with before joining, then so shall xenos.

## Changelog

:cl:
add: Xeno manifest on lobby screen
/:cl:

### DIFF
--- a/code/_onclick/hud/screen_objects/menu_text_objects.dm
+++ b/code/_onclick/hud/screen_objects/menu_text_objects.dm
@@ -112,6 +112,15 @@
 	var/mob/new_player/player = hud.mymob
 	player.view_manifest()
 
+/atom/movable/screen/text/lobby/clickable/xenomanifest
+	maptext = "<span class='maptext' style=font-size:8px>VIEW XENO MANIFEST</span>"
+	icon_state = "manifest"
+
+/atom/movable/screen/text/lobby/clickable/xenomanifest/Click()
+	. = ..()
+	var/mob/new_player/player = hud.mymob
+	player.view_xeno_manifest()
+
 /atom/movable/screen/text/lobby/clickable/background
 	maptext = "<span class='maptext' style=font-size:8px>BACKGROUND</span>"
 	icon_state = "background"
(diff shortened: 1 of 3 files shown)


Naaanii:
'View Hive' or 'View the Xeno Hive' would be a nicer name and label if possible.

lbnesquik:
On one hand, it feels weird because we never had it. On the other, it is only fair.

coldironwarrior:
not opposed to the concept. thinking about the balance of information.

something makes me want more info regarding normal hive numbers too (maybe too much but marines get everybody but AI on manifest but also numbers on each team mean something important...much to think about)
something also makes me a little worried that seeing the hive ruler caste might reveal too much info but i don't actually know how this'd be fixed. if the round is 30 minutes in and i see there's a shrike and two hive leaders marked, i can kind of guess what happened. which is maybe good for potential xeno players to know. and less good for potential marine players to know?

not necessarily "issues" since i think the intent was pretty straightforward, and these are questions that might overcomplicate or occlude that to the disadvantage of a good qol feature. duno. just thinking.

Ragecrabbo:
I feel like this has the unfortunate effect of enabling people who only play when certain hive leaders are on and it might give too much information about current round. Regardless this is a nice piece of QoL

code/datums/datacore.dm
```diff
@@ -118,6 +118,38 @@ GLOBAL_DATUM_INIT(datacore, /datum/datacore, new)
 
 	return dat
 
+/datum/datacore/proc/get_xeno_manifest(monochrome)
```
TiviPlus, line 121: missing autodoc

code/modules/mob/new_player/new_player.dm
```diff
@@ -233,6 +236,13 @@
 	popup.set_content(dat)
 	popup.open(FALSE)
 
+/mob/new_player/proc/view_xeno_manifest()
```
TiviPlus, line 239: ditto

code/_onclick/hud/screen_objects/menu_text_objects.dm
```diff
@@ -112,6 +112,15 @@
 	var/mob/new_player/player = hud.mymob
 	player.view_manifest()
 
+/atom/movable/screen/text/lobby/clickable/xenomanifest
+	maptext = "<span class='maptext' style=font-size:8px>VIEW XENO MANIFEST</span>"
```
TiviPlus, line 116: do nanis suggestion

TiviPlus (changes requested):
get kuro approval

PR #15145, opened 2024-02-19
nanites no longer purge during adaptation

Xander3359:
## About The Pull Request
nanites no longer purge during adaptation
## Why It's Good For The Game
losing 20% blood level and having to sit in medbay because you take 100 damage during the adaptation phase is not fun.
Your options were either to sit in medbay hooked up to 5 IVs, or keep spamming BKTT and restock your injectors in an attempt to heal before they got purged.
## Changelog
:cl:
balance: Nanites no longer purge chems during adaptation phase
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medical.dm
+++ b/code/modules/reagents/reagents/medical.dm
@@ -1303,6 +1303,7 @@
 /datum/reagent/medicalnanites/on_mob_life(mob/living/L, metabolism)
 	switch(current_cycle)
 		if(1 to 75)
+			purge_rate = 0
 			L.take_limb_damage(0.015 * current_cycle * effect_str, 0.015 * current_cycle * effect_str)
 			L.adjustToxLoss(1 * effect_str)
 			L.adjustStaminaLoss(1.5 * effect_str)
@@ -1311,6 +1312,7 @@
 				to_chat(L, span_notice("You feel intense itching!"))
 		if(76)
 			to_chat(L, span_warning("The pain rapidly subsides. Looks like they've adapted to you."))
+			purge_rate = initial(purge_rate)
 		if(77 to INFINITY)
 			if(volume < 30) //smol injection will self-replicate up to 30u using 240u of blood.
 				L.reagents.add_reagent(/datum/reagent/medicalnanites, 0.15)


Arctrooper8:
Why not make it so they just don't deal brute or burn damage during set up? Having it not purge during set up but then purge after is going to be confusing to people who aren't code diving. 
It still does toxin damage (which requires dylo that stops stamina regen) alongside actively draining stamina, and if that's still not enough just make it cause pain during set up.

Lumipharon:
The point of this is specifically to stop/make it difficult to nanite up in the field

Looking4bros:
You have to spam kits, not Chems. Bandages and ointment are enough to keep you from paincritting, and in Medbay you just hack the Blood machine and refill the minor bloodloss.

Xander3359:
> You have to spam kits, not Chems. Bandages and ointment are enough to keep you from paincritting, and in Medbay you just hack the Blood machine and refill the minor bloodloss.

Again that introduces needless tedium.
And kits only do so much since after kitting a limb you will simply take more damage immediately afterwards which removes the treated status.

Ragecrabbo:
> losing 20% blood level and having to sit in medbay because you take 100 damage during the adaptation phase is not fun. 

Usually you have pretty great PRs but my man just eat two isotonic pills after the nanites have finished adjusting to your body. You are making it sound like its a great hassle and takes alot of effort to deal with nanite adaptation phase.

QualityVan (commented):
Pinging @ivanmixo since he added the purging recently. This does seems like an oversight, but otoh the brute/burn damage isn't particularly meaningful in total and you can just have dylo in your system for the tox since it doesn't get purged, and only purging some of the time feels unintuitive.
Also not sure why you mention blood level since this changes nothing for that.

Looking4bros:
> Also not sure why you mention blood level since this changes nothing for that.

It's due to nanites healing the damage of their acclimation, the purging prevents microdosing chems, and he was using BKTT injections which are microdoses in terms of a 5u purge, so you had to restore ~15u of nanites with your blood. 
A working way of healing in it's current state is either bandaids and ointment, or eating a lot of pills, instead of injections. As mentioned in my above comment.

QualityVan:
Tedium declared needful.

PR #15146, opened 2024-02-19
DNR now simply makes you unable to return to your body.

MiniMeatwad:
## About The Pull Request
Now you can revive people who DNR so their mob can be taken by a ghost.

May have unintended consequences if there are any checks that utilize !mind to assume someone is DNR without explicitly checking.

## Why It's Good For The Game
So marines can still be revived by medics even if the player previously playing is no longer willing.

## Changelog
:cl:
code: Mindless or ownerless bodies no longer become undefibbable before the revive timer is up.
/:cl:

### DIFF
--- a/code/game/data_huds.dm
+++ b/code/game/data_huds.dm
@@ -263,14 +263,7 @@
 			status_hud.icon_state = "hudsynthdnr"
 			return TRUE
 		else
-			if(!mind)
-				var/mob/dead/observer/G = get_ghost(TRUE)
-				if(!G)
-					status_hud.icon_state = "hudsynthdnr"
-				else
-					status_hud.icon_state = "hudsynthdead"
-			else
-				status_hud.icon_state = "hudsynthdead"
+			status_hud.icon_state = "hudsynthdead"
 			return TRUE
 		infection_hud.icon_state = "hudsynth" //Xenos can feel synths are not human.
 		return TRUE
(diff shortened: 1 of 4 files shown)


DeltaFire15:
This is kind of bad. DNRing serves a twofold purpose, both removing you from the round, which you modify here, but also marking your body as "this is no longer worth to retrieve because the soul is gone".

With this PR, people who DNR (aka, will not actually have a player playing them when revived and will need to get vacced / give you no gain), still show up as the same marker as a marine that is normally revivable and will be able to fight immediately, making it quite likely you will end up sacrificing life and limb for what turns out to just be a mindless husk, given the minimap markers are the main way people track corpses.

This kind-of occurs currently with people who crash / disconnect, but is exacerbated when it's someone who very definitely does not intend to return to the game.

Thus, if you want to do something like this, I highly recommend adding a third type of marker for the minimap (and potentially healthhud), which would denote the body being soulless (aka formerly DNR), and thus make this condition a consideration visible for rescuers.

MiniMeatwad:
Is this a codersprite challenge?

MiniMeatwad:
> This is kind of bad. DNRing serves a twofold purpose, both removing you from the round, which you modify here, but also marking your body as "this is no longer worth to retrieve because the soul is gone".

This is just questioning player agency, players can already decide this on their own.


> With this PR, people who DNR (aka, will not actually have a player playing them when revived and will need to get vacced / give you no gain), still show up as the same marker as a marine that is normally revivable and will be able to fight immediately, making it quite likely you will end up sacrificing life and limb for what turns out to just be a mindless husk, given the minimap markers are the main way people track corpses.

Ghosts can take SSD mobs, provided they're alive. Medevac or store them somewhere safe I guess. 

As for minimap, eh I dunno, I'll leave it on a maintainer requesting it.
I also noticed the "[t_His] soul has departed" examine text is missing from our codebase, which tells you if the mob is "SSD" (DNR).

MiniMeatwad:
Oh yeah, apparently mind code is strange, so this change actually has a few bugs.

FosterGrime-Heymynameslouie:
could alternatively just make the marker in the tacmap yellow instead of green, so you know the body is revivable but the original soul is gone

TheUSSWatercooler:
I'd say orange, just to make it more noticeable in the mess of map icons.

PR #15166, opened 2024-02-22
Adds a accuracy malus to Puppeteer and Widow minions

Barnet2:
## About The Pull Request
Instead of a 70% chance to hit a selected limb, widow and puppet minions now have a 5% to hit. Accuracy for Widow, Puppeter and all other castes have not been changed.
## Why It's Good For The Game
Widow and Puppeteer can use their minions as disposables to get free chest fracs. With this change they have to actually get into the fray, and even with that the free fractures won't be so free anymore. Their extremely low TTK has not changed, just less chance for more permanent forms of damage in exchange.
## Changelog
:cl:
balance: Widow and Puppeteer minions have a 5% chance to hit their selected limb (chest) instead of a 70% chance.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
@@ -225,6 +225,8 @@
 	var/evolve_min_xenos = 0
 	// How many of this caste may be alive at once
 	var/maximum_active_caste = INFINITY
+	// Accuracy malus, 0 by default. Should NOT go over 70.
+	var/accuracy_malus = 0
 
 ///Add needed component to the xeno
 /datum/xeno_caste/proc/on_caste_applied(mob/xenomorph)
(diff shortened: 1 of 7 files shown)


MiniMeatwad:
Might be more like 24.5%, but yeah

blackdav123:
I think it would make more sense as a 1/6 chance to hit so that the targeted limb is equally likely to be hit instead of less likely.

Kuro020 (approved):

code/modules/mob/living/carbon/xenomorph/xeno_defines.dm
```diff
@@ -225,6 +225,8 @@
 	var/evolve_min_xenos = 0
 	// How many of this caste may be alive at once
 	var/maximum_active_caste = INFINITY
+	// Accuracy malus, 0 by default. Should NOT go over 70.
+	var/accuracy_malus = 0
```
TiviPlus, line 229: if it shouldnt go over 70 why not add a sanity unit test

TiviPlus (changes requested):

PR #15170, opened 2024-02-23
Boiler Root Removal

Barnet2:
## About The Pull Request
Removes being needed to root to fire. Needs #15171 with it
## Why It's Good For The Game
There was actual genuine balance concerns to giving Boiler root, but it made the caste even more unfun to play with. There are other variables we can change about Boiler to keep it fun; the addition of antigas nades is one, we can also look at caste limits and the glob itself. There were many things to try before putting in Root, which I think most can agree is extremely unfun.
## Changelog
:cl:
balance: Boiler no longer has to root to fire
/:cl:


Zangeiti:
we love gas spam in this house

Looking4bros:
Cades are vestigial structures left over from early development.

Woker2004:
As a boiler player, I don't think root should be killed, it's just the rooting/uprooting speed in my opinion

Woker2004:
I do think thar we should buff prae if this gets Merged in my opinion

Wisemonster:
Ya'll realize this will only go through if the pr making anti-gas nades more available/cheaper goes through, right? I'd be glad to see boiler root go simply cause it makes an already easy to catch caste (speed-wise) a joke against any kind of sniper or diving vali user if they are caught with their pants down (rooted).

SimplySmiley123:
> Ya'll realize this will only go through if the pr making anti-gas nades more available/cheaper goes through, right? I'd be glad to see boiler root go simply cause it makes an already easy to catch caste (speed-wise) a joke against any kind of sniper or diving vali user if they are caught with their pants down (rooted).

At the same time barely anyone really wants to deal with even more gas and something besides warlock cheesing any holds.

timothymtorres:
Boiler players switched to baneling so gas worries never really went away.

Ragnuv:
I would only be comfortable with this if the boiler range got cut down hard. Boiler root is hardly a problem when you're using a caste that can borderline outrange tac binocs and almost every gun marines have. If the problem with boiler is getting killed while rooted, just don't go melee range to root?

Kuro020 (approved):

Woker2004:
A better solution would make the unrooting, not rooting, faster

MiniMeatwad:
> I would only be comfortable with this if the boiler range got cut down hard. Boiler root is hardly a problem when you're using a caste that can borderline outrange tac binocs and almost every gun marines have. If the problem with boiler is getting killed while rooted, just don't go melee range to root?

If you get CAS on you and you're rooted, you're dead no questions asked. Death is overwhelmingly punishing for xenos so it's not very fair lol

Maybe just make any damage cancel bombardments and drop the cooldown as if they'd fired anyways if you want to achieve a similar effect to root

code/_globalvars/bitfields.dm
```diff
@@ -605,8 +605,7 @@ GLOBAL_LIST_INIT(bitfields, list(
 		"ABILITY_IGNORE_DEAD_TARGET" = ABILITY_IGNORE_DEAD_TARGET,
 		"ABILITY_IGNORE_SELECTED_ABILITY" = ABILITY_IGNORE_SELECTED_ABILITY,
 		"ABILITY_DO_AFTER_ATTACK" = ABILITY_DO_AFTER_ATTACK,
-		"ABILITY_USE_BURROWED" = ABILITY_USE_BURROWED,
-		"ABILITY_USE_ROOTED" = ABILITY_USE_ROOTED
+		"ABILITY_USE_BURROWED" = ABILITY_USE_BURROWED
```
TiviPlus, line 608: readd comma

TiviPlus (changes requested):
rest lgtm

TiviPlus (approved):

PR #15186, opened 2024-02-25
Adds the Critical sructure and ignore weed removal tag on Gargoyle.

2whatever2:
## About The Pull Request
Adds in the flag that makes it a critical structure, making it visible in pinpointer and able to be tracked on xeno structures list.

The ignore weed removal flag was advised by Xander, saying that other structures had it added due to buggy nature of weeds. with structures(?)
## Why It's Good For The Game
Xenos can track them now, Marines can track them now.
## Changelog
:cl:
add: Gargoyles are now act the same as other structures, and can be tracked with Structure pinpointers and hive tracker.
/:cl:

### DIFF
--- a/code/modules/xenomorph/resin_gargoyle.dm
+++ b/code/modules/xenomorph/resin_gargoyle.dm
@@ -4,6 +4,7 @@
 	icon = 'icons/Xeno/2x2building.dmi'
 	icon_state = "gargoyle"
 	max_integrity = 100
+	xeno_structure_flags = CRITICAL_STRUCTURE|IGNORE_WEED_REMOVAL
 	///Bool if we're currently alerting
 	var/is_alerting = FALSE
 	//cd tracking for the alert


Helg2:
While `IGNORE_WEED_REMOVAL` is understandable, adding `CRITICAL_STRUCTURE` for something that isn't critical or doesn't do something more than just alerting the hive is a bit stupid?
Also you have a minimap icons for the gargoyle iirc?
And the title isn't informative at all.

2whatever2:
You don't have minimap icons for gargoyle.
Critical structure is shared on every structure except acid pool, which is also missing the ignore_weed_removal.
Tivi made both of these things, and I asked them if it was intended for gargoyle in specific that they are not in pinpointer, they stated that they didn't thought of it.

https://discord.com/channels/498569646014857217/500801682058379265/1210690212234141736
image


The main intent of this PR is to let both xenos track it and marines track it, mostly marines with structure pinpointer as it searches for that structure with this flag.
I forgot to specify gargoyle on the title, will edit it.

coldironwarrior:
would this cause them to appear under the hivemind structures at the bottom half of hive status?

this is a good catch for now, but yeh; i'd definitely like to see map icons for them eventually, acid pools given parity regarding the features, and if this DOES appear on the list in hive status, maybe some consolidated way to view them so they don't clog that up.

all ideas for a later date.

Wisemonster:
Do gargoyles alert the xenos when they are destroyed? If not, it might be better to give them a map icon that flashes when they go off, similar to turrets.

TiviPlus (approved):
code ok

PR #15192, opened 2024-02-26
Wraith portal destruction change

Lumipharon:
## About The Pull Request
Blowing up a portal now destroys BOTH portals.
## Why It's Good For The Game
Currently, you throw a nade through a portal, it destroys the other end, which means the wraith can just instantly build a new portal. To actually counter a portal you have to portal your end, which is often unsafe, and its frankly unintuitive that nuking a portal only hurts one end.
## Changelog
:cl:
balance: Blowing up a wraith portal kills both ends of the portal now
/:cl:


ju45he:
Ok but what if you made the wraith, the thing holding portals open with it's mind, require a constant plasma drain to keep the portals open.

TiviPlus (approved):

PR #15193, opened 2024-02-26
Kitting fix

Lumipharon:
## About The Pull Request
Fixes some weird bug with kitting, probably.

No idea why it happens, but sometimes this proc passes null, which runtimes, and may or may not perma curse the person with a stuck do_after.
## Why It's Good For The Game
Bug bad.
## Changelog
:cl:
fix: fixed a runtime with kitting
/:cl:


TiviPlus:
this feels like burying the problem instead of fixing it

PR #15200, opened 2024-02-27
Larva queue fixes 3 : Off by one mystery

DeltaFire15:
## About The Pull Request
Soo after being done with highly important things (monster hunter) & taking a closer look at the entire proc chain around larva spawns, it turns out the larva slot already gets occupied before the larva spawn stuff is reached, which means a 0 in currently available slots is actually fine (only a value below 0 meaning no slots having been available).
That also opens up the question if the initial PR was needed at all given the slot should already be accounted for when it starts to ask, but oh well may aswell just fix this proc at this point.
## Why It's Good For The Game
:)
## Changelog
:cl:
fix: You no longer need two or more stored larvae to spawn (yippie)
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
+++ b/code/modules/mob/living/carbon/xenomorph/hive_datum.dm
@@ -991,7 +991,7 @@ to_chat will check for valid clients itself already so no need to double check f
 		return FALSE
 
 	var/datum/job/xeno_job = SSjob.GetJobType(/datum/job/xenomorph)
-	if((xeno_job.total_positions - xeno_job.current_positions) <= 0)
+	if((xeno_job.total_positions - xeno_job.current_positions) < 0)
 		to_chat(xeno_candidate.mob, span_warning("There are no longer burrowed larvas available."))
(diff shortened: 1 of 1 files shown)


Lumipharon:
Are you positive this is correct? I have 100% spawned in the last round completely fine when there was only 1 larva available. And the original bug let you spawn multiple larva as a single person (or multiple), with only 1 larva available.

DeltaFire15:
> Are you positive this is correct? I have 100% spawned in the last round completely fine when there was only 1 larva available. And the original bug let you spawn multiple larva as a single person (or multiple), with only 1 larva available.

This proc is in the silo section which means anything that is not silo (pre silo build / call) will spawn you in correctly.
And yes. See the parent proc which also carries the check, or the spawn proc even further up the chain which occupies the slot before going into the entire spawn chain this check sits in.

Lumipharon:
I am still spawning fine with 1 larva and a silo deployed

Lumipharon:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/15201

PR #15201, opened 2024-02-27
Larva fix for real

Lumipharon:
## About The Pull Request
Fixes larva queue for real.

Turns out the issue was that in some cases a larva slot was reserved before you actually spawned, and in other cases it didn't, even with the same button.

It now ALWAYS reserves the larva.

Additionally fixed what looks to be another bug where reservation wasn't passed on correctly.

Honestly the whole 'is reserved' thing should be killed, but I can't be bothered checking every proc to see if there is some other admeme or otherwise way of spawning without reserving first, so someone else can do that.
## Why It's Good For The Game
Actually fixes exploit without breaking it in other ways.
## Changelog
:cl:
fix: Fixed a new issue with the larva queue
/:cl:


TiviPlus (approved):
google definition of crazy

PR #15210, opened 2024-02-28
Remove delimb points from campaign

Lumipharon:
## About The Pull Request
After scenes that would go straight to the UN warcrimes tribunal, I've decided its better to simply remove points gain from delimbs entirely.
## Why It's Good For The Game
Chopping up the dead and dying for fun and profit is not ideal.
## Changelog
:cl:
balance: Campaign: Delimbing no longer provides additional credits
/:cl:

### DIFF
--- a/code/datums/personal_statistics.dm
+++ b/code/datums/personal_statistics.dm
@@ -112,7 +112,6 @@ GLOBAL_LIST_EMPTY(personal_statistics_list)
 	var/mission_projectile_damage = 0
 	var/mission_friendly_fire_damage = 0
 	var/mission_melee_damage = 0
-	var/mission_delimbs = 0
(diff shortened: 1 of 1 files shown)


Xander3359:
Why not make it count delimbs only if the victim is living?

Lumipharon:
because then they're just going to do it to living people instead of dead people, which is only marginally better (or worse, depending on your perspective)

Technical-l:
you know getting delimbed in hvh kinda sucks, beheading is funny tho. Just hate to see it when my SL has no arm so no epic SoM beheadings is a fair price to pay

Looking4bros:
Delimbing is like securing the kill, it's its own reward.

lbnesquik:
Not a full 24h

PR #13020, opened 2023-05-19
Railgun nerfs

Lumipharon:
## About The Pull Request
Muh xope etc.

100 sunder weapons are dumb, they'll all go in the bin eventually.
Reduced sunder of HVAP from 100 to 50, and for the IFF one 50 to 20 (but increased its damage)
This is still very significant sunder, it just means a p90 won't completely delete you after the fact.

Also made it fully wielded fire only.
## Why It's Good For The Game
100 to 0 sunder is bad gameplay.
## Changelog
:cl:
balance: reduced sunder of HVAP and IFF railgun rounds
balance: railgun must be fully wielded to fire
/:cl:

PR #15245, opened 2024-03-02
Railgun increased sunder

MiniMeatwad:
## About The Pull Request

More railgun sunder

## Why It's Good For The Game

Seems to be its niche as a weapon, could reduce its damage in exchange
## Changelog
:cl:
balance: 1.42x Railgun sunder
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -1529,7 +1529,7 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	max_range = 14
 	damage = 150
 	penetration = 100
-	sundering = 20
+	sundering = 80
 	bullet_color = COLOR_PULSE_BLUE
 	on_pierce_multiplier = 0.85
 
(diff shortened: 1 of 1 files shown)


coldironwarrior:
#13020

Woker2004:
I will go ahead and state this, on the marine side you can literally hampter a crusher's sunder in a single hit, same with basically every caste as well as the fact you get multiple shots in a mag, now onto a xeno perspective, It simply destroys your sunder in 1 hit, to 50% or below, railgun is, in my opinion, balance pretty well the way it currently is

MiniMeatwad:
I changed its value to be 1.42x what it was previously, in line with crusher's 70% sunder multiplier.

MiniMeatwad:
> I changed its value to be 1.42x what it was previously, in line with crusher's 70% sunder multiplier.
> 
> I dunno about giving it a mechanic to have increased damage against the multiplier, just gonna let Juro Kaeger decide

As for what I found to be SMG pick-off territory, it's usually around 60-70% sunder. Aka most castes have no armor at 60-70% sunder.

MiniMeatwad:
> We should go back to railgun having 1 shot capacity so you can try to bait it out as a xeno instead of just not getting to engage if you want to avoid huge sunder and/or stun/stagger.

This makes sense

ivanmixo:
Not a fan. @Kuro020 cast judgement

MLGTASTICa:
Not really a good change , the gun is fast to reload and can't miss since its hitscan. It effectively removes 30% of a xeno's effective in a single hit , i don't think even 20% is reasonable in first place.(ontop of dealing 150 damage , which  for most xenos is very significant)

Kuro020:
I dont think the railgun needs a buff overall, I see it frequently and it does fine

PR #15252, opened 2024-03-03
fix: map parity ai droid changes

coldironwarrior:
## About The Pull Request

each shipside map now has two combat droids for the AI
PoS had two combat droids
Arachne had a scout and a combat droid
Sulaco had a scout and a combat droid
Theseus had two combat droids

extra gun refills added as necessary to Arachne and Sulaco

## Why It's Good For The Game

many AI players are confused when they start out about why one droid functions differently from others, if they're playing on sulaco for the first time, for example; or they think they need the extra weapon in the storeroom attached to the scout droid, or whatnot.
i'd like to standardize the droids to reduce that.

## Changelog

:cl:
balance: all shipmaps now start with two combat droids
/:cl:

### DIFF
--- a/_maps/map_files/Arachne/TGS_Arachne.dmm
+++ b/_maps/map_files/Arachne/TGS_Arachne.dmm
@@ -6304,6 +6304,7 @@
 "fur" = (
 /obj/structure/rack,
 /obj/item/uav_turret/droid,
+/obj/item/uav_turret/droid,
 /turf/open/floor/mainship/cargo,
 /area/mainship/command/airoom)
 "fuI" = (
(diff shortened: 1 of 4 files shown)


lbnesquik:
Personally, I'd rather ditch the scout droid and have all maps get 2 combat droids. They aren't very useful, but the scout droid is clearly worse. It isn't stealthy enough to avoid getting mauled, and it can't do anything useful, like mark for OB, or railgun, or artillery.

coldironwarrior:
> Personally, I'd rather ditch the scout droid and have all maps get 2 combat droids. They aren't very useful, but the scout droid is clearly worse. It isn't stealthy enough to avoid getting mauled, and it can't do anything useful, like mark for OB, or railgun, or artillery.

yeaaaaa, i think that does make more sense to resolve the issue of confused new players anyways. swapped for two combat droids on each map + 2 spare guns

PR #15253, opened 2024-03-03
fix + patch: re-balances flanks on icy caves LZ2

coldironwarrior:
## About The Pull Request
title.

1. rebalances flanks based on standardizing the use of mats from the fob builder compared between LZ1 and LZ2 (see "Why" for more)
2. fixes some really weird zoning in LZ2
3. did you know the vent pipe into LZ2 can spawn broken and let xenos into the LZ before first drop? fixed via vent + pipe removal.

new lz2 below 
image

## Why It's Good For The Game

makes this lz something you can actually pick without throwing
1. LZ1 requires 40 plasteel and 154 metal to build. LZ2 required 50 plasteel and 245 metal to build. the fob builder, by default, has 100 plasteel and 200 metal. why was a lowpop map requiring more metal to secure the FOB than some highpop maps?
2. the open turf inside the shutters is not, in fact, an enclosed area
3. invalidates the point of having this LZ as a choice.

## Changelog
:cl:
balance: rebalanced flanks and exits on Icy Caves LZ2
fix: fixed zoning on Icy Caves LZ2 and vent + broken piping bugs
/:cl:

### DIFF
--- a/_maps/map_files/icy_caves/icy_caves.dmm
+++ b/_maps/map_files/icy_caves/icy_caves.dmm
@@ -83,9 +83,6 @@
 /turf/open/floor/cult,
 /area/icy_caves/caves/south)
 "at" = (
-/obj/machinery/atmospherics/pipe/simple/green/hidden{
-	dir = 4
-	},
 /obj/structure/cable,
 /turf/open/floor/tile/dark,
 /area/icy_caves/outpost/LZ2)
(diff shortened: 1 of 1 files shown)


RipGrayson (changes requested):
I'm fine with most of this, but the south side of the LZ is practically un-siegable with the changes you added here.

coldironwarrior:
mr rip, totally understanding of what you said, and realistically it probably wouldn't take *that* long to find a nice solution.

im going to close this though not because i disagree but because i dont have any passion whatsoever for Icy Caves and dread opening strongdmm to work on it when i think my time could be spent better elsewhere 😔

PR #15268, opened 2024-03-04
adds antigas grenade pouches to FC and SL vendors

coldironwarrior:
## About The Pull Request

title. four in an explosive pouch; 16 points for the set. no option to buy individuals.

**edit: pricing change, cheaper**
when asking the Barred Net briefly about it, they said req pricing equivalence with other grenades already available would be good but weren't sure about the maths for that; the numbers have told me:

razornades (also available as a 4-pack in SL vendor for 24 points) cost 500 for a set of 15 = **34 points per grenade**
antigas grenades cost 600 for a set of 25 = **24 points per grenade**

for a value of 136 req points for razornades, where 1 SL point is equal to 6 req points
or 96 req points for gas grenades, where 1 SL point is equal to 4 req points

and that in mind, the price for antigas based on that should be 16 SL points. this means an SL could buy 2 packs of these.

that's just how i figured i had overinflated the price and it needed to be cheaper, but isnt argumentation for *why* it needs to be that price specifically, if there's other concerns regarding balance and availability, i'd love to hear.

## Why It's Good For The Game

cool new grenades; these vendors need more Useful High-Impact Support Stuff and the antigas grenades fit the bill

## Changelog
:cl:
balance: antigas grenade sets in FC and SL vendors; 24 points for a set of four grenades
/:cl:


Woker2004:
As FC, while I struggle to use all my points, this is way to many points for 4 grenades, I could simply do more with detpacks

TiviPlus (approved):
code ok

Bluedino1025:
i can just get these from the weapon vendor and save points

MLGTASTICa:
Why is this not put as balance? There is no new feature added.

Jackal-boop:
Probably not needed. The marine vendor gets like 20 something of these? That's enough for the round start

PR #15270, opened 2024-03-04
Trying to disassemble some deployables is a balloon alert instead of to_chat

Xander3359:
## About The Pull Request
Trying to disassemble some deployables is a balloon alert instead of to_chat

## Why It's Good For The Game
Balloon > to_chat

This is also less misleading, since this flag has nothing to do with anchored status.

## Changelog

:cl:
fix: Trying to un-deploy some things no longer gives a misleading message
/:cl:

### DIFF
--- a/code/game/objects/machinery/deployable.dm
+++ b/code/game/objects/machinery/deployable.dm
@@ -59,7 +59,7 @@
 	if(!item)
 		return
 	if(CHECK_BITFIELD(item.flags_item, DEPLOYED_NO_PICKUP))
-		balloon_alert(user, "cannot be disassembled")
+		balloon_alert(user, "Cannot disassemble")
 		return
 	operator?.unset_interaction()
 	SEND_SIGNAL(src, COMSIG_ITEM_UNDEPLOY, user)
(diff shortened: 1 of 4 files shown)


ivanmixo (commented):
ew please uppercase it

code/game/objects/machinery/deployable.dm
```diff
@@ -59,7 +59,7 @@
 	if(!item)
 		return
 	if(CHECK_BITFIELD(item.flags_item, DEPLOYED_NO_PICKUP))
-		to_chat(user, span_notice("The [src] is anchored in place and cannot be disassembled."))
+		balloon_alert(user, "Cannot disassemble")
```
TiviPlus, line 62: you should probably say *why* its not undeplayoable instead

Xander3359: It's undeployable because it has the flag.

Xander3359: "[src] is undeployable because it cannot be undeployed"

TiviPlus: ok but it said anchored before was that lying

Xander3359: Yes.

Xander3359: Wether you anchor the tat or not, you can still NEVER undeploy it. This flag is not added/removed anywhere other than on the define itself.

PR #15272, opened 2024-03-04
Xeno's health and plasma HUDs will now properly update

breadhunt:
## About The Pull Request

See title; expect xeno HUDs to actually be consistent now, rather than only updating after a delay.

## Why It's Good For The Game

More responsiveness for xenos is very good. Health updates should **not** have a significant performance impact since this already updates for the little green/blue bars next to the xeno, this just updates the HUD on the right of the screen. (the little ovals)

## Changelog

:cl:
fix: Xeno health/plasma HUD now update correctly when the xeno is damaged/used an ability
/:cl:

### DIFF
--- a/code/datums/actions/ability_actions.dm
+++ b/code/datums/actions/ability_actions.dm
@@ -273,12 +273,13 @@
 
 //related mob procs
 
-///deducts the cost of using an ability
+///deducts the cost of using an ability & updates plasma HUD accordingly
 /mob/living/carbon/proc/deduct_ability_cost(amount)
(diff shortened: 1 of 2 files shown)


Lumipharon (changes requested):
? I pr'd instant updates for plasma like a week ago.

Is this for the dumb extra bar on the side of the screen? If that has issues than update health specifically, why are you calling the regular update when thats not what you need?

Lumipharon:
Superceded by #15320

PR #15281, opened 2024-03-05
Lowers warlock HP

blackdav123:
## About The Pull Request
Nerfs warlock HP from 375 to 325.
## Why It's Good For The Game
This caste has very high stats despite also being extremely safe. Hopefully this should make a warlock think twice before taking an AP rocket to the face which they would previously not even drop into crit from.
## Changelog
:cl:
balance: Lowers warlock HP from 375 to 325.
/:cl:


DemonOfHatred:
just stop

ju45he:
Why are you shooting AP rockets at the one caste that can reflect them lol

TiviPlus (approved):
kuro pr

yyzsong:
Good PR. It's kind of weird that warlock has more base hp than rav of all things, and I think you could stand to go further with this.

PR #15299, opened 2024-03-06
swaps research rig belt dexalin for isotonic

coldironwarrior:
## About The Pull Request
swaps researcher med rig dexalin bottle for isotonic

## Why It's Good For The Game

other medrigs have iso so this is part consistency; iso is probably more useful since rig belt comes with tricord anyways for oxy damage; and i dont see much sense to this since researchers can just grab dex+ injectors for free from the medvend and use the superior version of the chem for no cost

probably a greater indicator that dex/dex+ need to be consolidated or differentiated more but i didn't have any big ideas on what to do there

## Changelog
:cl:
balance: researchers spawn with isotonic pill bottles instead of dexalin bottles in their belts
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -342,9 +342,10 @@
 	new /obj/item/storage/pill_bottle/imidazoline(src)
 	new /obj/item/storage/pill_bottle/quickclot(src)
 	new /obj/item/storage/pill_bottle/hypervene(src)
+	new /obj/item/bodybag/cryobag(src)
 	new /obj/item/defibrillator(src)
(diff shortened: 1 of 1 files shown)


Arctrooper8:
You should replace what used to be their saw in their belt while you're at it. It serves no actual usage now and anybody wanting to play with newt still has the two in the vendor.

coldironwarrior:
> You should replace what used to be their saw in their belt while you're at it. It serves no actual usage now and anybody wanting to play with newt still has the two in the vendor.

yea, that's a good idea. wasn't sure what to put in there instead so i just swapped it for a cryobag. if anyone has an idea for anything more flavorful, i'm all ears

PR #15301, opened 2024-03-06
Glasses and goggles can be activated properly again

XElectricX:
## About The Pull Request

Welding goggles use `activate()` like every other eye wear. Medhuds should also work again, apparently the `if(QDELETED)` was not needed (yet it worked before I made any changes to it somehow?). A lot of the code is mostly so the action icons properly changed to reflect the state they were in.

## Why It's Good For The Game

Oopsy poopsy broke the glassies.

## Changelog
:cl:
fix: Welding goggles and HUD glasses work again. Night vision module req order spawns the actual module.
/:cl:

### DIFF
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -59,10 +59,13 @@
 
 ///Toggle the functions of the glasses
 /obj/item/clothing/glasses/proc/activate(mob/user)
-	if(activation_sound && deactivation_sound)
-		playsound(get_turf(src), active ? deactivation_sound : activation_sound, 15)
-
 	active = !active
+
+	if(active && activation_sound)
+		playsound(get_turf(src), activation_sound, 15)
+	else if(deactivation_sound)
+		playsound(get_turf(src), deactivation_sound, 15)
+
 	update_icon()	//Found out the hard way this has to be before update_inv_glasses()
 	user?.update_inv_glasses()
 	user?.update_sight()
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -51,10 +51,9 @@
 /obj/item/clothing/glasses/ui_action_click(mob/user, datum/action/item_action/action)
 	//In case someone in the future adds a non-toggle action to a child type
 	if(istype(action, /datum/action/item_action/toggle))
-		var/datum/action/item_action/toggle/toggle = action
-		//Toggling item actions are dumb; it has to be done this way for the icon to properly update
-		toggle.set_toggle(activate(user))
-		return
+		activate(user)
+		//Always return TRUE for toggles so that the UI button icon updates
+		return TRUE
 
 	return activate(user)
 
@@ -64,7 +63,7 @@
 		playsound(get_turf(src), active ? deactivation_sound : activation_sound, 15)
 
 	active = !active
-	update_icon_state()	//Found out the hard way this has to be before update_inv_glasses()
+	update_icon()	//Found out the hard way this has to be before update_inv_glasses()
 	user?.update_inv_glasses()
 	user?.update_sight()
 
@@ -249,14 +248,9 @@
 		activate(usr)
 
 /obj/item/clothing/glasses/welding/activate(mob/user)
-	if(active)
-		DISABLE_BITFIELD(flags_inventory, COVEREYES)
-		DISABLE_BITFIELD(flags_inv_hide, HIDEEYES)
-		DISABLE_BITFIELD(flags_armor_protection, EYES)
-	else
-		ENABLE_BITFIELD(flags_inventory, COVEREYES)
-		ENABLE_BITFIELD(flags_inv_hide, HIDEEYES)
-		ENABLE_BITFIELD(flags_armor_protection, EYES)
+	TOGGLE_BITFIELD(flags_inventory, COVEREYES)
+	TOGGLE_BITFIELD(flags_inv_hide, HIDEEYES)
+	TOGGLE_BITFIELD(flags_armor_protection, EYES)
 
 	eye_protection = active ? 0 : initial(eye_protection)
 
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -63,7 +63,7 @@
 
 	if(active && activation_sound)
 		playsound(get_turf(src), activation_sound, 15)
-	else if(deactivation_sound)
+	else if(!active && deactivation_sound)
 		playsound(get_turf(src), deactivation_sound, 15)
 
 	update_icon()	//Found out the hard way this has to be before update_inv_glasses()
--- a/code/modules/clothing/glasses/glasses.dm
+++ b/code/modules/clothing/glasses/glasses.dm
@@ -248,19 +248,32 @@
 		activate(usr)
 
 /obj/item/clothing/glasses/welding/activate(mob/user)
-	TOGGLE_BITFIELD(flags_inventory, COVEREYES)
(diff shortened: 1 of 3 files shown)


code/modules/clothing/glasses/glasses.dm
```diff
@@ -52,20 +52,21 @@
...
 
 ///Toggle the functions of the glasses
-/obj/item/clothing/glasses/proc/activate(mob/user, silent = FALSE)
-	if(!silent)
+/obj/item/clothing/glasses/proc/activate(mob/user)
+	if(activation_sound && deactivation_sound)
 		playsound(get_turf(src), active ? deactivation_sound : activation_sound, 15)
 
 	active = !active
+	update_icon_state()	//Found out the hard way this has to be before update_inv_glasses()
```
Lumipharon, line 67: don't call this directly

code/modules/clothing/glasses/glasses.dm
```diff
@@ -52,20 +52,21 @@
 	//In case someone in the future adds a non-toggle action to a child type
 	if(istype(action, /datum/action/item_action/toggle))
 		var/datum/action/item_action/toggle/toggle = action
-		toggle.toggled = !activate(user)
+		//Toggling item actions are dumb; it has to be done this way for the icon to properly update
+		toggle.set_toggle(activate(user))
```
Lumipharon, line 56: ? What is happening here? Hitting a toggle action already toggles the button if you're actually returning a value properly.

code/modules/clothing/glasses/glasses.dm
```diff
@@ -232,57 +233,47 @@
...
 
-/obj/item/clothing/glasses/welding/toggle_item_state(mob/user)
-	. = ..()
-	active = !active
-	icon_state = "[initial(icon_state)][!active ? "up" : ""]"
-	if(!active)
-		flip_up()
+/obj/item/clothing/glasses/welding/activate(mob/user)
+	if(active)
+		DISABLE_BITFIELD(flags_inventory, COVEREYES)
```
Lumipharon, line 253: you can toggle bitfields instead of doing if, else, which is consistant with what you're doing elsewhere, but honestly the whole thing is better to have a turn on and turf on proc instead of toggling, I don't know why you changed it, as its less safe and requires extra checks if you want to set to a specific state instead of just calling that proc directly.

code/modules/clothing/glasses/glasses.dm
```diff
@@ -444,7 +435,7 @@
 	///Looping sound to play
 	var/datum/looping_sound/active_sound = /datum/looping_sound/scan_pulse
 	///How loud the looping sound should be
-	var/looping_sound_volume = 15
+	var/looping_sound_volume = 25
```
Lumipharon, line 438: why god, why does it even have a looping sound? Its a passive item.

code/modules/clothing/glasses/glasses.dm
```diff
@@ -52,20 +52,21 @@
...
+	if(activation_sound && deactivation_sound)
 		playsound(get_turf(src), active ? deactivation_sound : activation_sound, 15)
 
 	active = !active
+	update_icon_state()	//Found out the hard way this has to be before update_inv_glasses()
 	user?.update_inv_glasses()
 	user?.update_sight()
-	update_icon_state()
 
 	return active	//For the UI button update
```
Lumipharon, line 71: yeah this is why the button gets fucked

Lumipharon (changes requested):
I am confused by this pr but also frankly by the original pr

code/modules/clothing/glasses/glasses.dm
XElectricX: It was part of the cons of it.

code/modules/clothing/glasses/glasses.dm
XElectricX: I thought it was much better to have one `activate` proc when all you're really doing with glasses is toggling functions. Do you want me to make them all use `turn_on` and `turn_off` procs?

code/modules/clothing/glasses/glasses.dm
XElectricX: I wanted the icon in the button to update but funny thing, the point where the `toggled` state is checked only happens if the return value of `ui_action_whatever` was not false. Which is why it only updated half the time. This way it guarantees it is always updated when `ui_action_whatever` is called.

code/modules/clothing/glasses/glasses.dm
XElectricX: What should I call then? Just `update_icon`?

XElectricX:
> I am confused by this pr but also frankly by the original pr

Why didn't you say anything then. 😄

code/modules/clothing/glasses/glasses.dm
Lumipharon: yes

code/modules/clothing/glasses/glasses.dm
Lumipharon: being really annoying shouldn't be a balancing factor. I'm not going to make a giga 1 shot kill gun with the draw back of breaking your eardrums.

BonkaiB: As I worked on the balancing of it. It isn't played only to you but others around you. To give xenos who have paid attention beforehand to it a slight heads up, it isn't ear breaking levels or shouldn't be nor the soundeffect for it to be grating. The item is about as passive as a scout cloak, you actively have to choose when to use it. It's to prevent complete blind sides without making the item completely silent or taking away another factor

Being "annoying" isn't the con of the item

XElectricX:
Thanks Lumi, your recommendations worked.

Lumipharon:
going to testmerge because the revert is fucked and this seems fine now, but yes I prefer having a toggle off and toggle on proc because A: its consistant with other stuff, and B: It lets you just call toggle on/off directly without having to worry about checks to see what state its currently in.

code/modules/clothing/glasses/glasses.dm
Lumipharon: yes, see below.

code/modules/clothing/glasses/glasses.dm
```diff
@@ -51,21 +51,21 @@
...
 
-	activate(user)
+	return activate(user)
 
 ///Toggle the functions of the glasses
-/obj/item/clothing/glasses/proc/activate(mob/user, silent = FALSE)
-	if(!silent)
+/obj/item/clothing/glasses/proc/activate(mob/user)
+	if(activation_sound && deactivation_sound)
 		playsound(get_turf(src), active ? deactivation_sound : activation_sound, 15)
```
QualityVan, line 63: Missing one sound will prevent the other from playing.

XElectricX: No reason why you'd only have one sound rather than both, but I can change it if you think it's best.

Lumipharon: There's no inherent reason why you'd have to have to have both sounds. Someone could make glasses that only have an activation sound.

code/modules/clothing/glasses/glasses.dm
```diff
@@ -59,10 +59,13 @@
...
 /obj/item/clothing/glasses/proc/activate(mob/user)
-	if(activation_sound && deactivation_sound)
-		playsound(get_turf(src), active ? deactivation_sound : activation_sound, 15)
-
 	active = !active
+
+	if(active && activation_sound)
+		playsound(get_turf(src), activation_sound, 15)
+	else if(deactivation_sound)
+		playsound(get_turf(src), deactivation_sound, 15)
```
Lumipharon, line 67: if you are activating, and have no activation sound, it will play the deactivation sound.

Lumipharon (changes requested):

PR #15311, opened 2024-03-07
Stops non tank users from touching tank

TiviPlus:
## About The Pull Request

stuff like removing guns etc

## Changelog
:cl:
add: restricted things like removing tank gun to tank crew only
/:cl:


Lumipharon (approved):
>didnt say skill issue
Why even bother

PR #15322, opened 2024-03-08
Removes the shoulder mount.

Naaanii:
## About The Pull Request
Title
## Why It's Good For The Game
An attachment that bypasses gun skills (SGs), can be used alongside melee cheese like sledge and also be used with binoc modules (RR) isn't very good design. Would be nicer if someone could address these issues but no one has done so in years so, death it is.
## Changelog
:cl:
del: Removes shoulder mount from research gacha.
/:cl:


aarrisu:
@nutric

AusarAtkin:
Pretty sure you can't use it with vali since it takes up your suit module slot, And since the probing been removed it became even rarer.

UzedPickle:
I agree with your points but I dont think t is problematic enough to warrent a removal. It being a rare neiche item for creative loadout artists that rarely gets used is a fine state to be in, maybe it sees use in 1/100 rounds but thats fine in my view. neiche items shoudnt be removed for being neiche.

Naaanii:
Oh yeah I should update the text.

Lumipharon:
> The module never bypassed any skills. Despite the lack of visible recoil, projectile spread from low skill remains no matter the weapon attached.

I don't really think it needs to be removed, but what he means is it ignores RECOIL. SG guns are entirely unusable by non sgs due to like 20 recoil. Shouldermount has no recoil thus any unga can have a shoulder mounted sg without problems.

SzkieletorNutric:
>any unga can have a shoulder mounted sg without problems

The problem for 'any unga' relies in acquisition and availability.
Others have already expressed just how rare it is to see the module in-game, as well as the removal of xeno probing making a present and active Researcher a prerequisite for it to even have a chance to appear.

Actually getting an SG-29 for the module will only happen if one is lucky to loot a DNR Smartgunner that decided to deploy with it, thus merely replacing the now vacant role.
Otherwise it comes down to REQ to break bank for a whole new smartgun and any additional drums thereafter.

Yes, the module is capable of bypassing the primary malus of an unskilled SG wielder - but imo the prerequisites for someone to actually do it are expensive enough for it to be very limited and fair.

If it is such a taboo all of a sudden and you are incapable of actually adressing it, simply stick the 29 onto the same list of weapons that cannot be mounted, just like the other smartguns, grenade launchers, REQ weapons, etc.

ivanmixo:
Why have this exist as a type at all?

Naaanii:
> Why have this exist as a type at all?

If anyone wants to rework or fix it in the future I guess.

ivanmixo:
That's why github exists. We don't need dead code sticking around.

Naaanii:
> That's why github exists. We don't need dead code sticking around.

Sure, done.

ivanmixo (approved):

Kuro020:
If someone actually wants this shitty thing to exist, fix the bugs, make it respect the stats it ignores right now, and put it in req for an appropriate price, for now having rng op items is terrible so I'm not allowing it to stay

PR #15332, opened 2024-03-10
removes the tgs theseus

coldironwarrior:
## About The Pull Request

rip to the ~~almayer~~ theseus :)
removes the TGS Theseus from being votable, and then also just the game since if you can't vote it in NW, no other gamemode sees it

## Why It's Good For The Game
really old ship, been around for many years. the game has changed a lot since then and the nature of maps is that they're only so flexible for accommodating such changes.

low-pop staff on shipside makes running from midship req to fore OB to aft engis really difficult. the prep rooms for specialists are abysmally tiny. tons of space is given to shipside roles' and command roles' quarters which never see use. medbay pharmacy is terribly small. cas arrangements waste a ton of prep time. nobody knows where the mech bay is.

## Changelog

:cl:
del: TGS Theseus has been retired
/:cl:


yyzsong:
It's so over

Bluedino1025:
just remake it from the ground up with the exception of a single tile

00ze-cyclone:
We’re getting rid of Theseus meanwhile Shitlaco is still plaguing us

Naaanii:
Yeah, I don't see how Theseus is a problem when Sulaco exists.

lbnesquik:
I think it has some practical issues, but only arachne really doesn't have any of those. Walking from POS req to the OB room takes a while too. I can see the specialist rooms being bigger, but I never really had problems getting the gear I need in there. 

And sure, we got fluff rooms for command, but sulaco has an entire botany bay, Arachne has a sports area and some empty unused rooms, POS has a bar, toilets, CL room, and even a chapel. No one's bothered about those, or the fluff atmospherics rooms.

As for the pharmacy issue, I think those aren't too hard to solve. We got a fair chunk of room in the east of medical to eat.
CAS? I'm not sure how you would solve that, except by making ammo not suck to move. Which is honestly annoying on all ships. The whole concept of doing that prep with the ripley isn't fun. 

The generators are a fair remark, but at the same time, you basically don't need to think about fixing all of them. Just use the center ones and you got plenty of power. I've honestly had more issues with power on POS with autodoc marines than on Theseus. The latter one having more reserve generators by default. 

All in all, sorry to say, I think this PR shouldn't be merged.

bowlofcereal:
No

RipGrayson:
Not happening

PR #15335, opened 2024-03-10
removes the tgs sulaco

coldironwarrior:
## About The Pull Request

title

## Why It's Good For The Game

people HATE it!!!! no, same idea as theseus sort of, the lay of the game has changed a lot so there's not always tons of shipside staff to effectively split between two floors and cross-ship duties, for example you have to not only run from one side of the ship to the other as a ship tech to load the OB then do generators, but you also have to go up a ladder.

incidentally the ladders are also kinda super suck for hijack; marines or xenos camp ladders, and xenos has a tougher time if they land on the top floor (maybe?); if xenos land top floor and friendly ERT lands, that ERT will never make it down to the team at SD, etc etc.

also the pharmacy is too tiny and only has one bluespace beaker, this is cruel and unusual

im of the terribly strange opinion as well that people dying to space crush because they went through the disposals system on the bottom floor is CRINGE !

Closes #15380

## Changelog
:cl:
del: retired the TGS Sulaco
/:cl:


74delta:
:ladder:

Barnet2:
After some thought I think players have a point in keeping variety at the cost of map quality, even if I disagree I am but a minority in that. I personally loathe commanding on Sulaco but it is okay to keep around; Arachne is similar to PoS in quality so it definitely can be done to replace Theseus and Sulaco.

coldironwarrior:
guys i saved sulaco, so i can close this now!

PR #15368, opened 2024-03-12
Improves the hypersleep console

lbnesquik:
## About The Pull Request
This PR improves the hypersleep equipment console. It adds some categories for items, and filters them into them.

I spent like 6 hours on this. That was a bit of a pain.

## Why It's Good For The Game
It should reduce lag and improve readability on long rounds with hundreds of items in the vendor. In both cases, it is because we don't just have one giant list of items, but about 7 smaller lists. 

If I still had energy, I would have made the console be able to dump out items by categories, but I'm too tired. 

## Changelog
:cl:
qol: added categories to the hypersleep bay console
/:cl:


Xander3359:
Man, you're really gonna make all this work and not even make it TGUI????

lbnesquik:
> Man, you're really gonna make all this work and not even make it TGUI????

It was supposed to be quick, and then 4 hours in, I didn't want to trash everything.

Xander3359:
> > Man, you're really gonna make all this work and not even make it TGUI????
> 
> It was supposed to be quick, and then 4 hours in, I didn't want to trash everything.

This really should be TGUI.
Doing all this extra effort to keep it as a /Topic...

code/game/objects/machinery/cryopod.dm
```diff
@@ -39,37 +119,96 @@
...
+		GLOB.cryoed_item_list_explosive -= I;
+		GLOB.cryoed_item_list_melee -= I;
+		GLOB.cryoed_item_list_clothing -= I;
+		GLOB.cryoed_item_list_food -= I;
+		GLOB.cryoed_item_list_drugs -= I;
+		GLOB.cryoed_item_list_containers -= I;
+		GLOB.cryoed_item_list_other -= I;
 		CRASH("Deleted or erroneous variable ([I]) called for hypersleep inventory retrivial.")
-	if(!(I in GLOB.cryoed_item_list))
+	if(!I in (GLOB.cryoed_item_list_gun || GLOB.cryoed_item_list_ammo || GLOB.cryoed_item_list_explosive || GLOB.cryoed_item_list_melee || GLOB.cryoed_item_list_clothing || GLOB.cryoed_item_list_food || GLOB.cryoed_item_list_drugs || GLOB.cryoed_item_list_containers || GLOB.cryoed_item_list_other))
```
Xander3359, line 193: You can;t check lists like this.
You have to do !(var in list) x your amount of lists

code/game/objects/machinery/cryopod.dm
```diff
@@ -1,3 +1,12 @@
+#define HS_CNSL_GUN 0
```
Xander3359, line 1: Abbreviations for the sake of abbreviation makes code less readable.
Please just use the full words so it's less insane

code/game/objects/machinery/cryopod.dm
```diff
@@ -1,3 +1,12 @@
...
+#define HS_CNSL_AMMO 1
+#define HS_CNSL_EXPLOSIVE 2
+#define HS_CNSL_MELEE 4
+#define HS_CNSL_CLOTHING 5
+#define HS_CNSL_FOOD 6
+#define HS_CNSL_DRUGS 7
+#define HS_CNSL_CONTAINERS 8
+#define HS_CNSL_OTHER 9
```
Xander3359, line 9: Don't forget to ``#undef`` your defines at the end of the file or else they are global.

Xander3359 (changes requested):
I still really think this should be TGUI instead of all this HREF bullcrap

Naaanii:
Great feature, but Xander's idea for TGUI is better. Iirc TGUI also has a search function like in Valhalla spawning options.

lbnesquik:
> Great feature, but Xander's idea for TGUI is better. Iirc TGUI also has a search function like in Valhalla spawning options.

It would honestly be better, they are right. I just honestly spent +6 hours on it, and I'm not feeling like stumbling around learning TGUI for what I thought would be a quick PR. Maybe in the future, but for how much we use it, it's a straight improvement for users anyway.

code/game/objects/machinery/cryopod.dm
lbnesquik: image
I went for solution B. Didn't notice the nonsensical code.

Lumipharon:
why god

Naaanii:
Pls unstale, its still a neat feature.

TiviPlus:
this code is nasty but should just be tgui but w/e

Xander3359:
Hey so this PR is pretty broken:
The "Dispense All" just doesn't work anymore: https://github.com/tgstation/TerraGov-Marine-Corps/issues/15443 
Clicking on "Drugs" brings me to the "Other" category
"Food" Just doesn't have food (All the MREs/Protein bars end up in "Other)

This generally makes it harder to find something you are looking for too, since instead of doing CTRL F once, you have to do CTRL F for every category to find your item.
And it's not like this reduces the scroll time since it still sends all the unsorted crap to "Other" which holds the more relevant items (Like supply beacons)
image

lbnesquik:
The reason for the MRE's not working is because the roomba tries to unload everything, including MRE's. And MRE's are actually storage packs with food items in them. 

image

PR #15369, opened 2024-03-13
fire support bino tweaks

Lumipharon:
## About The Pull Request
Adjust some targeting code for fire support binos, and cleaned it up a bit.

- You now actually needactually see your target to lase so no more calling cas through walls, and you can block it with smoke etc.
- Lasing is now done against an atom, not turf. This means if you lase a human/mech/etc, THE FIRESUPPORT WILL TARGET THEM IF THEY MOVE. This only lasts until the firesupport is actually called in (i.e. once the lase goes away) so you still have time to run away before the firesupport blasts the spot you were standing in.
- This also means you could cancel the firesupport entirely if you break LOS around a corner or something, so in most cases its probably better to still target a turf, but could be funny for dropping missiles on mechs.

In the future I might add the ability to keep painting a target for guided munitions (like the cruise missile), but not for now.
## Why It's Good For The Game
Funny cas through walls is kinda wack.
Proper ability to lase a specific target.
## Changelog
:cl:
balance: Campaign: Fire support binos require line of sight, but can now target any atom
/:cl:

### DIFF
--- a/code/game/objects/items/firesupport_binoculars.dm
+++ b/code/game/objects/items/firesupport_binoculars.dm
@@ -7,10 +7,11 @@
 	///Faction locks this item if specified
 	var/faction = null
 	///lase effect
-	var/obj/effect/overlay/temp/laser_target/laser
+	var/image/laser_overlay
+	///lasing time
 	var/target_acquisition_delay = 5 SECONDS
 	///Last stored turf targetted by rangefinders
-	var/turf/targetturf
+	var/turf/target_atom
 	///Current mode for support request
 	var/datum/fire_support/mode = null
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
Fire support sucks even with through-wall lazing. I'd say its far better to have it like this , else it just becomes unused again.

Lumipharon:
this is for campaign.

Kuro020 (approved):

MLGTASTICa:
> this is for campaign.

my bad then

Lumipharon:
updated to use byond vision, as its too frustrating with how different true LOS is vs byond LOS.

PR #15372, opened 2024-03-13
Optimizes playsound

TiviPlus:
## About The Pull Request

TM before merge, did not profile because i dont have 60 clients on hand i would test with

## Changelog
:cl:
code: optimized sounds
/:cl:

### DIFF
--- a/code/game/sound.dm
+++ b/code/game/sound.dm
@@ -62,9 +62,11 @@ A good representation is: 'byond applies a volume reduction to the sound every X
 		frequency = GET_RANDOM_FREQ // Same frequency for everybody
 	// Looping through the player list has the added bonus of working for mobs inside containers
 	var/sound/S = sound(get_sfx(soundin))
-	for(var/mob/M AS in GLOB.player_list|GLOB.aiEyes)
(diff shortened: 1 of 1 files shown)


Kapu1178:
You don't need real clients to profile this with. You can lightly modify the code & manually add objects to the spatial grid recursive contents lists nearby.

I think it's unlikely this will be faster than just iterating over all cliented mobs, as spatial grid overhead is not nothing.

TiviPlus:
no but I work on the assumption its faster than 80+ players

Barnet2:
I'll reiterate what I said on Discord;

While this is testmerged I often had instances where audio wouldn't play. Notably, it happened most frequently when I spawned roundstart, when I died and turned into a ghost, and when I spawned in as ERT. They would clear up 15-60 seconds after those instances, where I could hear again. Also saw people in dchat and XOOC talking about it.

Lumipharon:
From when I played this when it was test merged, it seems like an orange_ear is not assigned when a mob spawns, but when I moved far enough from my spawn spot for it to assign a new orange_ear, sound seemed to work fine.

PR #15374, opened 2024-03-13
resprites mimir worn sprite

coldironwarrior:
## About The Pull Request

title. makes it less bulky-looking so it fits better with armors. puts the side views into perspective.
image

## Why It's Good For The Game

people didnt like mimir after it got resprited, and i dont think it looked particularly good either. side sprites in particular were not in perspective compared to the human mob sprite, so you ended up with the filter box sticking way out, and the filter box itself had some issues in looking appropriately box-like. this moves it lower, closer to where the old mimir filter-box-thing was on the mob sprite, and makes it stick out less.

because mimir was also made to look so bulky, it did this thing in the back with having the backside piece of the module on side sprites so far out that on some light armors and style line stuff, the module didn't actually connect to the armor it was attached to, so you had a weird prong
image

lastly the color palette of the filter box being reddish-grey and the green highlights on the module itself means that this looked really bad attached to armors of most color palettes other than green. by making the box just grey to match the rest of the module, the green accents will now look better, perhaps even COOL on armors, like how the hiln/surt red accents look, or how the blue autodoc ones look.

tries to keep an appropriately happy medium of the aesthetic of the current module sets, being more bulky-looking and full-body, while making mimir less busty and an easier fit for more armor styles and colors.

## Changelog
:cl:
imageadd: resprites mimir module worn sprite
/:cl:


Kuro020 (approved):

Lumipharon:
the sideview of the chest bit is darker than it seems like it should be, compared to the front view. And the should green bit has a weird pixel that seems much to bright, again compared to the front view. 
image

PR #15377, opened 2024-03-14
theseus edits

coldironwarrior:
## About The Pull Request

nani said instead of trying to remove theseus, i should just rework it, so i spent four hours doing what i could to make PO lives easier :)

this PR doesn't fix the whole ship, but does some things, mostly in the hangar.

- moves all the cas stuff together
- re-aligns the alamo loading zones
- moves FC bunks to south of hangar, moves synth room there too
- redoes the staff officer area
- very gracefully shoves a spot in for the vehicle bay on the north of the hangar
image
image

## Why It's Good For The Game

ship has really bad flow; people don't know where the synth stuff is
holds a space for new vehicle features
fixes alignment of stuff with alamo since it got re-done

there's a lot of stuff that'll remain suck-y, like moving pods from req, if you're ST juggling OB and gens, or CSE juggling those and overwatch, etc. the woes of having a 250-tile-wide shipmap.

## Changelog
:cl:
qol: re-organizes theseus hangar and staff officer quarters
/:cl:

### DIFF
--- a/_maps/map_files/Theseus/TGS_Theseus.dmm
+++ b/_maps/map_files/Theseus/TGS_Theseus.dmm
@@ -529,10 +529,6 @@
 /turf/open/floor/mainship/mono,
 /area/mainship/hull/starboard_hull)
 "abZ" = (
-/obj/machinery/power/apc/mainship{
-	dir = 4
-	},
-/obj/structure/cable,
 /turf/open/floor/mainship/tcomms,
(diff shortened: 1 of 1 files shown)


lbnesquik:
Good stuff. Although, I just wanna say, there is no consistency with synthetic rooms. It's just down to memory for people to know. 
Also, so sad that people forget the giant conveyors tricks.

Issue #14446, opened 2023-11-14
No FC GHMME Vendor in Valhalla


74delta:
### Description

There is a FC Point vendor but no GHMME, likely someone simply forgot about it.

### Test Merges

```#14258: 'Reverts Hyperscale 3' by lKiseki at commit f1eedc1e4a
#14393: 'Flare Rebalance: Removes infinite flares and adds a flare crate to req' by KrabSpider at commit 0bf22a2d60
#14378: 'Bump rust-g to 3.0.0' by vvvv-vvvv at commit d01a959374
#14261: 'Adds Beginner loadouts.' by Xander3359 at commit a98df94476
#14389: 'Gorger Feast is AOE + lower cooldown' by Xander3359 at commit 22ada19c83```

### Reproduction Steps

Go to Valhalla. 
No GHMME vendor for FC. 

### Screenshots

_No response_

PR #15382, opened 2024-03-14
Add the FCDR vendor

lbnesquik:
## About The Pull Request
Fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/14446

Add the FCDR vendor into the FC Valhalla room. 
Extended the room for it by eating some Rwalls. 

image

## Why It's Good For The Game
Useful

## Changelog
:cl:
add: Added FC vendor to valhalla 
/:cl:


TiviPlus:
doesnt this need a valhalla variant or is everything infinite

lbnesquik:
> doesnt this need a valhalla variant or is everything infinite

Is there? I never go to that place. I'll check it out.

lbnesquik:
Well, the gear vendors aren't infinite. So I'm not sure why they need to be special versions?

RipGrayson:
> Well, the gear vendors aren't infinite. So I'm not sure why they need to be special versions?

The special version needs to use no power and be indestructible so people can't grief it with grenades. I don't know if the vendor you added has either of those traits.

lbnesquik:
> > Well, the gear vendors aren't infinite. So I'm not sure why they need to be special versions?
> 
> The special version needs to use no power and be indestructible so people can't grief it with grenades. I don't know if the vendor you added has either of those traits.

Well I copied the squad leader version. It should be sturdy, but I saw nothing about power.

Helg2:
Can you move the "nosmoking" sign a bit higher so it doesn't stack in the mapper?
Also maybe add a second table to the first one so there's less just empty space?

lbnesquik:
> Can you move the "nosmoking" sign a bit higher so it doesn't stack in the mapper? Also maybe add a second table to the first one so there's less just empty space?

I can't believe you are nitpicking me too. 
Done.

PR #11624, opened 2022-11-25
TGMC Christmas 2023 [IDB IGNORE]

RipGrayson:
## About The Pull Request

a_xenomorph_queen_wearing_a_santa_suit_a

With Thanksgiving officially over, it's time to move onto the next holiday.
Introducing the TGMC Christmas PR, this PR

-Adds festive decorations to all ships and shuttles.
-Adds Christmas food and drink to all ships and shuttles.
-Places Christmas trees throughout the ships.
-Implements gifts with randomly generated presents inside, much like TG already does (for the sake of balance the items are much more curated compared to TG.)
-Allows players the magical chance to meet Santa and his elves face to face.
-Various other surprises and things I can't be bothered to note.

This PR isn't meant to be merged, only TMC'ed. At a later date I _might_ add some of the holiday stuff into the code itself for easier reuse in future Christmases.

**TO ADMINS: Add this TM into the game whenever you think is appropriate for Christmas time, this PR was opened somewhat early so the code could be vetted before December rolls around**

## Why It's Good For The Game

We didn't have Christmas in 2021, I have taken it upon myself to fix this problem. If you don't like this PR you are officially a scrooge who needs to embark on a vaguely allegorical journey to learn the true meaning of Christmas.

## Changelog
:cl:
add: Added Christmas. If you're seeing this in game the PR was merged somehow, whoops.
/:cl:

PR #15431, opened 2024-03-21
removes unused icons in the "unused" folder

coldironwarrior:
## About The Pull Request

title. gets rid of the whole thing. 
files re/moved were:
xmas.dmi (none of these are used and the special MREs are duplicated in the food dmi where they're actually called from)
Marine_Research.dmi (resin jelly's sprite and the sprite for xeno scalps were from here)

- renames "biomass" sprite to "resin jelly"
- re-homes those two in a new dmi for "xeno materials"
- edits the name pathing of any other dmis in the "Xeno" folder that didn't have spacing in the name for /readability/
- renames the acid well file to "acid_well.dmi" since it was "acidpool.dmi" before which can cause confusion with the actual acid pool structure

## Why It's Good For The Game

spring cleaning, baby steps. none of these were used, but moreover i don't think they have the potential to *be* used; they're very old CM-era sprites and not our style.

## Changelog
:cl:
imagedel: deleted the "unused" folder of dmis
/:cl:


Lumipharon:
some of these are definitely used in the chrismas PR's, but I guess they can just be in that pr

coldironwarrior:
i mean, "yes" to your point about putting relevant sprites on the christmas PRs;

but no, actually, not a single one of these is used as per checking the files of #11624. we don't override weeds with snowmen and cocoons with snow angels. we don't override armor with christmas variants, that's a CM thing. we don't override all helmets with santa hats, that's a CM thing. we don't use any of these present icons, the icons we use are called "gift", not "present" and are housed elsewhere. as stated in the PR body, the seasonal MRE sprites we use are duplicated of the ones here called from "food/mre.dmi", not this dmi, so the ones in here are unused.

lbnesquik:
Does clearing icons like this actually save significant bandwidth for anyone? For anyone coding, it just adds extra .git to download when you first clone the repo, no?

TiviPlus (approved):
code ok

PR #15432, opened 2024-03-21
removes unused icons from the "objs" folder

coldironwarrior:
## About The Pull Request

gets through about half the dmis in the "objs" folder that aren't categorized and cleans out unused icons. doesn't do any sanitization of the file management, it's still a disaster.

## Why It's Good For The Game

cleaner repo; i'd hazard to say all the sprites i deleted are also of the sort that would not be coming back
## Changelog
:cl:
imagedel: deletes unused sprites from the "obj" folder
/:cl:


Lumipharon:
I'm pretty certain some of these are used. grill_on at the very least.

coldironwarrior:
i mean, i know i *did* touch it because it's deleted in here but i wasn't aware that i caught the grill in this which was not intentional, i thought the only things i removed out of the kitchen.dmi were the blender and green blood meatspike. i will add the grill back in (obviously!)
but again besides the fat-finger with the grill, no, i checked each of these in VSC manually and if the icon was deleted here, it's because there is no object using it. in the case of the power_cond_white.dmi, i looked to see if that dmi itself was called anywhere, and it wasn't. these are unused.

lbnesquik:
I like the fuel tank and the white cables. Can we keep those? I could see use using that honestly! Like, CAS fueling sucks.

TiviPlus (approved):
Code ok

PR #15450, opened 2024-03-23
Fixes my bad PR regarding hypersleep

lbnesquik:
## About The Pull Request
I didn't test my PR quite enough.

Fix the button drugs not working
Make dispense all work, and work by categories
Make food checking better by using IsFood
## Why It's Good For The Game
Bug fix good
## Changelog
:cl:
fix: fix the hypersleep dispense all not working
fix: fix the drug category not displaying
/:cl:


code/game/objects/machinery/cryopod.dm
```diff
@@ -172,9 +172,34 @@
...
+					dispense_item(I, usr, FALSE)
+			if(STATE_DRUGS)
+				for(var/obj/item/I in GLOB.cryoed_item_list_drugs)
+					dispense_item(I, usr, FALSE)
+			if(STATE_CONTAINERS)
+				for(var/obj/item/I in GLOB.cryoed_item_list_containers)
+					dispense_item(I, usr, FALSE)
+			if(STATE_OTHER)
+				for(var/obj/item/I in GLOB.cryoed_item_list_other)
+					dispense_item(I, usr, FALSE)
```
Lumipharon, line 202: just switch to change what list to dispense from. No point in copypasting the dispense 10 times.

Lumipharon (changes requested):

code/game/objects/machinery/cryopod.dm
```diff
@@ -172,34 +172,29 @@
 
 		visible_message(span_notice("[src] beeps happily as it disgorges the desired objects."))
 
+		var/list/da_list
```
Lumipharon, line 175: cringe var name

lbnesquik: Changed the var name.
Also, I figured having multiple big lists would be better when they get really full of crap. Now that's less of a problem with roomba refills, I think.

Lumipharon (approved):

PR #15451, opened 2024-03-23
Removes an outdoor plat miner from Icy Caves

blackdav123:
## About The Pull Request
Removes the plat miner above medbay from icy caves.
icycaves miner2

## Why It's Good For The Game
This plat miner is very easy to permanently take just by landing tad in the spot that it would already want to be in for pushing last disk. Icy Caves is also a lowpop map so xenos typically wont have many extra resources to devote to sieging an autism tad plat miner fort.
## Changelog
:cl:
balance: Removed outdoor plat miner from Icy Caves
/:cl:

### DIFF
--- a/_maps/map_files/icy_caves/icy_caves.dmm
+++ b/_maps/map_files/icy_caves/icy_caves.dmm
@@ -2115,10 +2115,6 @@
 	dir = 9
 	},
 /area/icy_caves/caves/cavesbrig)
-"mo" = (
-/obj/machinery/miner/damaged/platinum,
-/turf/open/floor/plating/ground/snow/layer1,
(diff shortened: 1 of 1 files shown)


lbnesquik:
This is a balance PR during a balance freeze so it will get nuked, but I think this is a valid change.

Helg2:
Replace with plasma miner?

blackdav123:
> Replace with plasma miner?

considered it but its in a corner that the tad can completely cover so it'll be a pain in the ass for xenos on lowpop regardless of which type. there are also two miners a screen away north and west which are both much easier for xenos to deal with, so i'd rather marines have to get one of those instead of the free tad miner.

PR #15459, opened 2024-03-25
gives job vendors tag sets for item categories; reorganizes so alike items are grouped together

coldironwarrior:
## About The Pull Request

three things for job vendors:

- each specialist job-specific vendor now has a tag set associated with it to group alike items together and provide brief explanation of what an item is
  - for example, claymore mines in the engineering vendor now have an "explosives" tag

- vendor contents lists were reorganized so that everything of a similar category appears together. if you wanted to run mortar as engineer, all the shells are grouped together (as they have been), but the AI targeting module and rangefinders will also be grouped with them

- removes the "recommended" tags from items in the vendor. not a problem on SG; in the case of corpsmen and engineers, they were given to individually useful items, but without much consideration for what might happen if someone were to take multiple "recommended" items together.
  - for example, currently, as engineer, you'd be advised to take sandbags, plasteel, metal, and deployable shields, and nothing else, providing an extremely limited build. as a corpsman, you're recommended to take neuraline, synap, meralyne and dermaline pills, qk+ hypo, peri+ hypo, meraderm as a syringe case, nanoblood hypo, normal hypo, and big hypo; just a lot of hypos, and nothing very cohesive about them together.

one thing for synth vendor:

- changes the magenta storage tag for aquamarine; the magenta was pretty eye-straining and kind of too close in color to the other stuff it was near; all the colors that appear now won't cause that kind of issue again

image
image
image

## Why It's Good For The Game

hopefully this makes vendors easier to navigate and digestible for new players, and gives a better indication of what items might work well together. 

## Changelog
:cl:
add: adds category tags to the specialist vendors
/:cl:


lbnesquik:
Fantastic PR

TiviPlus:
pretty

tgui/packages/tgui/interfaces/MarineSelector.jsx
```diff
@@ -159,13 +159,73 @@ const ItemLine = (props) => {
...
+          {prod_color === 'sg-machinegun' && (
+            <Box inline mr="6px" ml="6px" color="#00FF00">
+              Machine Gun
+            </Box>
+          )}
+          {prod_color === 'sg-smartpistol' && (
+            <Box inline mr="6px" ml="6px" color="#FFE4C4">
+              Smartpistol
+            </Box>
+          )}
```
TiviPlus, line 221: Did you think about using dictionaries instead of this massive copy paste if block?

TiviPlus (changes requested):

tgui/packages/tgui/interfaces/MarineSelector.jsx
coldironwarrior: let me know if what i did looks okay

TiviPlus: What I meant was
name: color then just look it up as needed

Lumipharon:
Organizing them into categories is great, but I feel it may be easier to just have literal categories that item fall into.
Like ATM we have "essentials" and "sg supplies". So rather than have 6 items all with a visible explosive flag, why not just have them sorted into an explosive category?

TiviPlus:
not what i meant but i am beyond caring now

PR #15462, opened 2024-03-26
Removes traumatic shock from organ damage, health analyzer can see organ damage from above 5 instead of 10.

Barnet2:
## About The Pull Request
What it says on the tin.
## Why It's Good For The Game
For traumatic shock:
After doing a bit of research I found out that the only thing organ damage does below threshold was traumatic shock, which essentially seems to be pain. This led to invisible pain, and then past threshold it just stacked with the extremely debilitating effects of a bruised organ (heart, for example, at bruised gives -20 max hp and reduced stamina.) Below the 10 threshold it was pain you could not really deal with besides also getting pain from cloneloss if you took peri+, alongside the reduced max HP.
For the health analyzer: You should be able to see when a organ is getting close to be bruised. I picked above 5 as the most reasonable number to start seeing organ damage at, as fracture RNG can pick from 3 - 5 damage; since the max is 5, this should give you a pre-warning when the next time you take organ damage you may meet the threshold, and thus doctors can assist with peri+ which is meant for lower amounts of organ damage in the first place.
## Changelog
:cl:
balance: Organ damage no longer does traumatic shock
balance: Health Analyzer can see organ damage above 5 instead of 10 and above
/:cl:

### DIFF
--- a/code/game/objects/items/scanners.dm
+++ b/code/game/objects/items/scanners.dm
@@ -253,11 +253,16 @@ REAGENT SCANNER
 		data["implants"] = unknown_implants
 		var/damaged_organs = list()
 		for(var/datum/internal_organ/organ AS in human_patient.internal_organs)
-			if(organ.organ_status == ORGAN_HEALTHY)
+			if(organ.damage <= 0)
 				continue
+			var/organ_health_status = ""
+			if(organ.organ_status == ORGAN_HEALTHY)
+				organ_health_status = "Scarred"
+			else
+				organ_health_status = organ.organ_status == ORGAN_BRUISED ? "Bruised" : "Broken"
 			var/current_organ = list(
 				"name" = organ.name,
-				"status" = organ.organ_status == ORGAN_BRUISED ? "Bruised" : "Broken",
+				"status" = organ_health_status,
 				"damage" = organ.damage
 			)
 			damaged_organs += list(current_organ)
--- a/code/game/objects/items/scanners.dm
+++ b/code/game/objects/items/scanners.dm
@@ -253,7 +253,7 @@ REAGENT SCANNER
 		data["implants"] = unknown_implants
 		var/damaged_organs = list()
 		for(var/datum/internal_organ/organ AS in human_patient.internal_organs)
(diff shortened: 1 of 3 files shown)


Lumipharon:
Why nerf organ damage? The whole point of organ damage is attrition, this just makes it do significantly less.

I never really saw why organ damage was hidden below a threshold though, seems to make no meaningful difference one way or another.

Barnet2:
> Why nerf organ damage? The whole point of organ damage is attrition, this just makes it do significantly less.
> 
> I never really saw why organ damage was hidden below a threshold though, seems to make no meaningful difference one way or another.

An alternative is to tack it onto the threshold hold, i.e if above organ bruised you take pain. Tivi seems to dislike the shock below the threshold, which I do as well.

Lumipharon:
Doing something below the threshold is good because it means organ damage... actually does something before getting to severe levels, but even then its only 15 shock max below threshold which is not a lot on it's own.

Organ damage SHOULD be meaningful, that is the point of attrition, this just makes it easier to ignore. Fighting with severe organ fail is supposed to be a terrible idea because it doesn't just happen easily (you have to repeatedly accrue it from prolonged fighting, or just ignoring broken bones for long periods), and you either get the field treatment of swapping it for cloneloss, or being taken out of the fight for a short while for the medbay fix.

lbnesquik:
I actually kinda like how organ damage is hidden.

Lumipharon:
To elaborate further on why I strongly disagree with this.
Pain as a general mechanic is a form of soft attrition that is managable (but not entirely fixable) by the common unga. Without going into the arguments of why attrition is good and removing yet more of our limited attrition is bad, pain relies on there being a variety of courses for it to accumulate.

No single source of traumatic shock is likely to cause debilitating shock_stage unless you're like almost crit with no pain killers for ages. It relies on all sources of badshit causing traumatic shock that accumulates together, reducing your window to fight without negative shock_stage debuffs from actually taking more damage.

Removing more sources of traumatic shock because they're inconvenient greatly weakens the entire pain mechanic as a concept.

PR #15463, opened 2024-03-26
Warding nerf

Barnet2:
## About The Pull Request
Near halves the armor warding gives, essentially what I did in #14030.
## Why It's Good For The Game
As we learned with hold order, stacking armor values at the higher end can lead to some really wacky damage. For example, with king warding on crusher you could be receiving 3.75dmg from a T12 bullet, which not accounting for sunder (you can heal the in-combat sunder with hivelord beam or just existing, meaning getting shot by the small arms will only increase sunder very slowly) would take over 100 bullets to kill, realistically 70+ with sunder accounted for.
## Changelog
:cl:
balance: Warding pheros have near half effectiveness, from 2.5 armor per aura level to 1.5
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -161,10 +161,10 @@
 
 	if(warding_aura != (received_auras[AURA_XENO_WARDING] || 0))
 		if(warding_aura) //If either the new or old warding is 0, we can skip adjusting armor for it.
-			soft_armor = soft_armor.modifyAllRatings(-warding_aura * 2.5)
+			soft_armor = soft_armor.modifyAllRatings(-warding_aura * 1.25)
 		warding_aura = received_auras[AURA_XENO_WARDING] || 0
 		if(warding_aura)
-			soft_armor = soft_armor.modifyAllRatings(warding_aura * 2.5)
+			soft_armor = soft_armor.modifyAllRatings(warding_aura * 1.25)
 
 	recovery_aura = received_auras[AURA_XENO_RECOVERY] || 0
 
--- a/code/modules/mob/living/carbon/xenomorph/life.dm
+++ b/code/modules/mob/living/carbon/xenomorph/life.dm
@@ -161,10 +161,10 @@
 
 	if(warding_aura != (received_auras[AURA_XENO_WARDING] || 0))
 		if(warding_aura) //If either the new or old warding is 0, we can skip adjusting armor for it.
-			soft_armor = soft_armor.modifyAllRatings(-warding_aura * 1.25)
+			soft_armor = soft_armor.modifyAllRatings(-warding_aura * 1.5)
 		warding_aura = received_auras[AURA_XENO_WARDING] || 0
 		if(warding_aura)
-			soft_armor = soft_armor.modifyAllRatings(warding_aura * 1.25)
+			soft_armor = soft_armor.modifyAllRatings(warding_aura * 1.5)
 
 	recovery_aura = received_auras[AURA_XENO_RECOVERY] || 0
 


Lumipharon:
This ignores the fact that there is only 2 (3 if you include defender) high armour xenos, so this change has a pronounced effect on every other xeno that doesn't have super armour armour resists.

Marine hold order is different because its a flat damage reduction instead of a armour modifier, so what armour a human is wearing doesn't have any bearing on the benefit of the hold order, while xenos base stats have a DRAMATIC effect on the benefit of warding.

Basically, if you have an issue with warding buffing a couple of castes too much, make a change that targets them (or make it a flat dam modifier like hold order I guess), because this change is significantly more far reaching than just those castes.
I don't imagine anyone is complaining about a runner being OP with warding right now.

Barnet2:
> This ignores the fact that there is only 2 (3 if you include defender) high armour xenos, so this change has a pronounced effect on every other xeno that doesn't have super armour armour resists.
> 
> Marine hold order is different because its a flat damage reduction instead of a armour modifier, so what armour a human is wearing doesn't have any bearing on the benefit of the hold order, while xenos base stats have a DRAMATIC effect on the benefit of warding.
> 
> Basically, if you have an issue with warding buffing a couple of castes too much, make a change that targets them (or make it a flat dam modifier like hold order I guess), because this change is significantly more far reaching than just those castes. I don't imagine anyone is complaining about a runner being OP with warding right now.

Even if we take a more average armor xeno with say, 50 armor versus a T12, it is a 23%ish EHP increase for standing around about any xeno. I don't particularly think that pheros/orders should be that game changing, same with recovery pheros but no idea what to do about that dependency.

Naaanii:
This is a pretty massive nerf.

Arturlang:
Okay, let's do the math! King currently grants 15 armor to every stat with his 6 phero strength, but this nerf would reduce it to 7.5,
Even praetorian for example, often using warding has a phero strength of 4.5 grants 11.25 grants armor currently, but would be reduced to just 5.625 more armor, this would practically make warding such a tiny buff that it might be just more useful to save your plasma than even use warding, or instead focus on frenzy. This is a really bad idea, xenos currently don't need any nerfs.

MLGTASTICa:
I believe the proper way to fix this issue is to make armor buffs have a multiplier applied onto them based on how close to 90-95 bullet armor someone's default armor is.

Lumipharon:
> > This ignores the fact that there is only 2 (3 if you include defender) high armour xenos, so this change has a pronounced effect on every other xeno that doesn't have super armour armour resists.
> > Marine hold order is different because its a flat damage reduction instead of a armour modifier, so what armour a human is wearing doesn't have any bearing on the benefit of the hold order, while xenos base stats have a DRAMATIC effect on the benefit of warding.
> > Basically, if you have an issue with warding buffing a couple of castes too much, make a change that targets them (or make it a flat dam modifier like hold order I guess), because this change is significantly more far reaching than just those castes. I don't imagine anyone is complaining about a runner being OP with warding right now.
> 
> Even if we take a more average armor xeno with say, 50 armor versus a T12, it is a 23%ish EHP increase for standing around the xeno horde. I don't particularly think that pheros/orders should be that game changing, same with recovery pheros but no idea what to do about that dependency.

You're implying every source of pheros is king level, and again, my point is xeno castes all having varying armour levels, and sunder is also factored in here, as well as varying AP of weapons. So warding unlike hold order can be negated by weapon choices.

I strongly dislike nerfs that exist to address a specific issue but have MUCH wider impacts, such as this. Its a bad change that has wider repercussions.

As for phero dependency, xeno gameplay is kind of built around the mechanic, so I don't know what you'd want to do with it without gutting xenos and starting from the ground up.

Barnet2:
Feedback taken, will find another way to change Warding.

Naaanii:
> Feedback taken, will find another way to change Warding.

Quick workaround is to just kill King warding and have Xenos use phero towers for max rank pheros. Don't see why King needs better pheros than Queen.

Arturlang:
@Lumipharon Just was a example of max tier warding effects, unless pheremone towers do more? I'm not sure

Arturlang:
Fact is that rifle ammo has very little sundering and penetration as they're designed to

DeltaFire15:
> This ignores the fact that there is only 2 (3 if you include defender) high armour xenos, so this change has a pronounced effect on every other xeno that doesn't have super armour armour resists.

In fact, this only makes warding more powerful. Hold order is a flat damage reduction applied independant from armor, which means it is always apply 5% flat DR / level to incoming damage. Warding on the other hand adds on top of already existing armor, which leads to drastically scaling effective DR the higher the pre-warding armor of the xenomorph was. Is is a *minimum* of 2.5 / mag or 15 at King, but starts to scale to completely wack levels of resilience at higher base armor.

Then, add on top of this the fact human health maximum is static. That of xenomorphs is not, varying between castes, which also leads to drastic variations of EHP on many low-armor castes, such as Gorger and Behemoth, when affected by warding, as opposed to say, a rouny. Let alone that of armored AND high health castes.

Then there's also the bomb armor part, which is currently making devastation explosions as a category laughable for xenomorphs due to almost any level of warding preventing the most powerful part of them, which ironically affects marines through almost any level of protection.

Combine that with near-permanent uptime (minus being set on fire, but almost every source of fire is just as dangerous to marines themselves, especially chasing ones), and you have a hidden almost permanent DR bonus to xenos that most likely flips games.

In closing, warding phero as a concept is super scuffed, given how tanky it makes some xenomorphs. Bear in mind, marines do not have any more free things to stack on top of their DR (only Armor + Hold situationally), while xenos can stack every single one of their DR abilities on top of eachother. Ever fought an armored xeno with king ward + gorger link & shield + queen DR (nobody uses this but it slaps on heavy armor xenos) + drone link + Hivelord HoT? Yeah.

Arturlang:
Warding on it's own is a minor buff, but it helps a lot of xenos pretty easily, @DeltaFire15 your latter example  requires the cooperation of atleast 5 different xenos buffing one xeno, which is highly unlikely to be kept up for long due to how they are used. Point of pheremones is to encourage xenos sticking together and buffing themselves. Health to armor ratio is really not standardized much at all, but xenos do need to actually be tanky, as most castes require you to go up to the marines to do anything, marines have a lot of ranged weaponry that deals a lot of damage, which xenos do not.

TiviPlus:
Meanwhile regen pheros:

TiviPlus (approved):
code ok

PR #15464, opened 2024-03-26
Bull buffs

Barnet2:
## About The Pull Request
**For the love of god do a test merge**
Bull has boiler's zoom ability
Bull is now a big mob (same size as crusher for the purpose of abilities)
## Why It's Good For The Game
For the zoom ability:
Crusher and Bull are interesting because they are both very similar and very different. One main difference is the fact that bull is not affected by razor wire, but Crusher is. Giving zoom to either of these xenos to allow them to make a good decision based on information given has been an idea that has floated in the air for awhile; though I feel like it is better for bull to have and crusher to not. Crusher has the ability to shrug off alot more than Bull, as well as the counter of razor as a surprise. Giving it zoom would sort of ruin the purpose. Bull has many more ways to die instantly when charging, thus I believe the zoom is a good addition.
For mob size:
This was requested by Kuro as a buff, and honestly it might be too big of one but it'll have to be seen with a TM. You can still get staggered but like with most T3's you won't get knocked over and back by slugs. One slugger would make Bull completely useless because they could either deny your charge or just knock you back into the marines repeatedly. Again, has to be TM'd to see if this is way too strong.
## Changelog
:cl:
balance: Bull now has Boiler's zoom ability
balance: Bull is now has a mob size of big
/:cl:

### DIFF
--- a/code/datums/keybinding/xeno.dm
+++ b/code/datums/keybinding/xeno.dm
@@ -110,6 +110,13 @@
 	keybind_signal = COMSIG_XENOABILITY_XENO_SPIT
 	hotkey_keys = list("Q")
 
+/datum/keybinding/xeno/long_range_sight
+	name = "long_range_sight"
+	full_name = "Long Range Sight"
+	description = "Toggles the zoom in."
+	keybind_signal = COMSIG_XENOABILITY_LONG_RANGE_SIGHT
+	hotkey_keys = list("E")
+
 /datum/keybinding/xeno/xenohide
 	name = "xenohide"
 	full_name = "Hide"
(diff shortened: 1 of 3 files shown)


Lumipharon:
A note that charging makes you immune to stagger and slowdown entirely, while big mob makes you immune to all hard stuns (outside of hammer) and knockback from projectiles/grenades.
So mr slugger in your example cannot CC a bull in any way, while it is charging.

Barnet2:
> A note that charging makes you immune to stagger and slowdown entirely, while big mob makes you immune to all hard stuns (outside of hammer) and knockback from projectiles/grenades. So mr slugger in your example cannot CC a bull in any way, while it is charging.

Yeah, it's Kurt requested though I think it may be too strong. We shall see, I suppose

Naaanii:
This is a pretty dumb buff for a charging caste considering it already got a decent buff not too long ago.

lbnesquik:
I like the zoom the most.

Al-1ce:
> Zoom buff


I actually really like this. It feels fitting and more fair, and also makes bull more practically difficult than a runner.


>Size buff


The whole point of bull originally was to be a fast mini-crusher targeted towards infantry. Making it a larger size comes with benefits that aren't fair given bulls evolution rank.

Victor239:
I'd like to see this testmerged because as it stands right now Bull is virtually never picked because of how vulnerable it is.

Kuro020 (approved):

PR #15467, opened 2024-03-26
Boiler nerfs #2 (hopefully final)

Barnet2:
## About The Pull Request
Boiler glob CD increased by 5 seconds (old elder time)
Boiler lance range only extends to their zoom range
## Why It's Good For The Game
These issues still plagued boiler even when it had root, but of course people didn't really play Boiler.
The gas uptime can be quite oppressive hence the CD nerf, should be fine around there given previous Elder boiler
The lance change is a bit more complicated. Lance going offscreen of boiler range seems to hint that it was made to counter snipers, as the snipers cannot see the Boiler. That doesn't exactly make sense since snipers are supposed to at least plink at the boiler. Lance still will remain effective in that it goes past the targeted turf while doing mega, mega damage. Normal glob keeps the 30 range since it cannot shoot offscreen without other xeno aid. The 25 range for lance will be slightly offscreen if they are facing north since north + south are two tiles shorter due to widescreen, but should be alot better than 40 at max.
## Changelog
:cl:
balance: Boiler glob CD increased by 5
balance: Boiler lance has a range of 25, the same distance as it's horizontal zoom range.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
+++ b/code/modules/mob/living/carbon/xenomorph/castes/boiler/castedatum_boiler.dm
@@ -51,7 +51,7 @@
 	// *** Boiler Abilities *** //
 	max_ammo = 7
 	bomb_strength = 1.3 //Multiplier to the effectiveness of the boiler glob.
-	bomb_delay = 27 SECONDS
+	bomb_delay = 32 SECONDS
 
 	acid_spray_duration = 10 SECONDS
 	acid_spray_damage = 16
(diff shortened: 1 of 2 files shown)


Al-1ce:
Might as well remove the spitting ability altogether at this rate.

Jackal-boop:
I think with AG being available roundstart boiler is in an OK place right now

Lumipharon:
The CD nerf is needed, I don't see the issue with offscreening lance though.
As mentioned previously, snipers can see further than boilers, they can shoot at the same range, and marines can you know, move up.
Offscreening is a very specific niche of hammering static positions which is boiler's role to begin with, as this just rewards turtling more.

Kuro020 (approved):

PR #15472, opened 2024-03-26
Buffs ozymelyn purge rate from 5u to 10u

MLGTASTICa:
## About The Pull Request
Recently tried the defiler caste. The ozymelyn smoke was kind of useless since it was never adjusted for the new reagent BKKT pouches.

## Why It's Good For The Game
A marine standing in a smoke of ozemelyn will usually take a 2-3u dose. Ozymelon has a metab rate of 1.5 units , so it would only purge 10 units. Most marines using BKKT will have 40-50 units inside of them , making the effect kind of underwhelming. With these changes it will purge 20 units, putting a decent dent in a marine's bloodstream meds. (but still not enough to really purge their entire bloodstream, that requires reagent slashing) 
This is without taking into consideration wheter or not they have dylovene . If the marines are also using dylovene the effects are preety much halved (since dylo purges ozymelyn at 1u per tick)

## Changelog
:cl:
balance: Ozymelyn purge rate buffed from 5u to 10u
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -622,7 +622,7 @@
 	scannable = TRUE
 	toxpwr = 0 // This is going to do slightly snowflake tox damage.
 	purge_list = list(/datum/reagent/medicine)
-	purge_rate = 5
+	purge_rate = 8
 
 /datum/reagent/toxin/xeno_ozelomelyn/on_mob_life(mob/living/L, metabolism)
 	if(L.getToxLoss() < 40) // if our toxloss is below 40, do 0.75 tox damage.
--- a/code/modules/reagents/reagents/toxin.dm
+++ b/code/modules/reagents/reagents/toxin.dm
@@ -622,7 +622,7 @@
 	scannable = TRUE
 	toxpwr = 0 // This is going to do slightly snowflake tox damage.
 	purge_list = list(/datum/reagent/medicine)
-	purge_rate = 8
+	purge_rate = 10
 
 /datum/reagent/toxin/xeno_ozelomelyn/on_mob_life(mob/living/L, metabolism)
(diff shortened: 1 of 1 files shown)


Xander3359:
What.
Ozy purges all chems, not just 1 chem at a time.
A marine premedding on the maximum amount of BKTT is still using 30u of each chem
And according to your math, this leaves them with 10u of chems left.
I don't disagree with the change, but the body of the PR is somewhat misleading.

AusarAtkin:
You said 5u to 10u, but you changing it to 8u

Lumipharon:
I don't understand the logic here. If you avoid standing in the smoke too long you don't get as severely effected. ok?

MLGTASTICa:
> I don't understand the logic here. If you avoid standing in the smoke too long you don't get as severely effected. ok?

The case i present is the average. Where marines run out of the smoke , even if they were deep enough in it. Unluckier marines that were deep in get at most 2 doses.

AusarAtkin:
Its named Ozelomelyn by the way

Arturlang:
Issue here is that defiler gas doesn't transfer many chemicals due to everyone using gas masks as there's no malus to wearing one, and they're extremely accessible.

TiviPlus:
fix your cl btw

MLGTASTICa:
> fix your cl btw

fixed.

TiviPlus (approved):
code ok

Jackal-boop:
> fix your cl btw

You could be evil and add a gas mask FOV for April fools

Jackal-boop:
> Issue here is that defiler gas doesn't transfer many chemicals due to everyone using gas masks as there's no malus to wearing one, and they're extremely accessible.

You could be evil and add a gas mask FOV for April fools

Kuro020:
I dont really think defiler needs a buff as is

PR #15475, opened 2024-03-27
Welding repair amount scale off engi skill

AusarAtkin:
## About The Pull Request
Welding repair amount scale off engi skill, different skill level should have different effect.
## Why It's Good For The Game
Make use of higher engi skill
## Changelog
:cl:
balance: Welder repair amount scale off Engi skill.
/:cl:

### DIFF
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -322,9 +322,9 @@
 		if(!do_after(user, (fumble_time ? fumble_time : repair_time) * (skill_required - user.skills.getRating(SKILL_ENGINEER)), NONE, src, BUSY_ICON_BUILD))
 			return TRUE
 
-	if(user.skills.getRating(SKILL_ENGINEER) >= SKILL_ENGINEER_ENGI)
+	if(user.skills.getRating(SKILL_ENGINEER) > SKILL_ENGINEER_ENGI)
 		repair_amount *= (1+(0.1*(user.skills.getRating(SKILL_ENGINEER) - SKILL_ENGINEER_ENGI)))
-		
+
 	repair_time *= welder.toolspeed
 	balloon_alert_to_viewers("starting repair...")
 	handle_weldingtool_overlay()
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -322,6 +322,7 @@
 		if(!do_after(user, (fumble_time ? fumble_time : repair_time) * (skill_required - user.skills.getRating(SKILL_ENGINEER)), NONE, src, BUSY_ICON_BUILD))
 			return TRUE
 
+	repair_amount *= (1+(0.1*(user.skills.getRating(SKILL_ENGINEER) - SKILL_ENGINEER_ENGI)))
 	repair_time *= welder.toolspeed
 	balloon_alert_to_viewers("starting repair...")
 	handle_weldingtool_overlay()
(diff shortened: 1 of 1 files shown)


Lumipharon:
As noted in discord, there is no conditional check here. That means this nerfs unskilled users, which has some fairly large repercussions.

AusarAtkin:
> As noted in discord, there is no conditional check here. That means this nerfs unskilled users, which has some fairly large repercussions.

Yes, it will affect non engineer that want to fix cades and vehicles, the other option is to not have negative value at all but that mean it would be just flat buff, i'm fine with either way but i took the neutral route because kuro made me decide. 

Would it be better if it does have a check and it being a flat buff?

Lumipharon:
that would be my preference yes. Low skill is already punished with longer build time, I don't see why they need another debuff on top.

AusarAtkin:
> that would be my preference yes. Low skill is already punished with longer build time, I don't see why they need another debuff on top.

Aighty, gonna update it in a sec

code/game/objects/objs.dm
```diff
@@ -322,6 +322,9 @@
 		if(!do_after(user, (fumble_time ? fumble_time : repair_time) * (skill_required - user.skills.getRating(SKILL_ENGINEER)), NONE, src, BUSY_ICON_BUILD))
 			return TRUE
 
+	if(user.skills.getRating(SKILL_ENGINEER) > SKILL_ENGINEER_ENGI)
+		repair_amount *= (1+(0.1*(user.skills.getRating(SKILL_ENGINEER) - SKILL_ENGINEER_ENGI)))
```
Lumipharon, line 326: why are you referencing the engie level of skill? Ref the skill_required var.

AusarAtkin: Because my intent is to make ENGI (3) as the baseline of 100% repair amount, because how metal and plasteel barricade repair only have skill_required at Metal (1) and Plasteel (2) level of engineer, if i put it at those then it mean mostly Everyone going to have increased repair amount. Or you would prefer it to be like that?

Lumipharon: What

AusarAtkin: Like this? 
        if(user.skills.getRating(SKILL_ENGINEER) > skill_required) 
                 repair_amount *= (1+(0.1*(user.skills.getRating(SKILL_ENGINEER) - skill_required)

Or this

if(user.skills.getRating(SKILL_ENGINEER) > skill_required) 
                 repair_amount *= (1+(0.1*(user.skills.getRating(SKILL_ENGINEER) - SKILL_ENGINEER_ENGI)

Arturlang:
This is going to mean that you can just outrepair a xeno just slashing the cade, won't it? Which was already pretty difficult to do even as a ravager.

code/game/objects/objs.dm
Lumipharon: This is a kuro question since its balance, but both ways have issues. Arbitrarily choosing SKILL_ENGINEER_ENGI though is specifically problematic if something in the future has a higher requirement. If I need level 5 to repair something for example, its unintuitive to repair it for 120% or whatever just for meeting the requiment.

AusarAtkin: Fair, so skill_required + 1 should be fine then?

TiviPlus (approved):
code ok but yea what the others said

PR #15484, opened 2024-03-28
Removes the ability to vote for the current map during groundmap vote

Barnet2:
## About The Pull Request
What it says on the tin.
## Why It's Good For The Game
A common complaint was the same map 3-4 times in a row. This eliminates that.
## Changelog
:cl:
del: Removes the ability to vote for the current map during groundmap vote
/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -273,10 +273,12 @@ SUBSYSTEM_DEF(vote)
 					var/datum/map_config/VM = config.maplist[GROUND_MAP][map]
 					if(!VM.voteweight)
 						continue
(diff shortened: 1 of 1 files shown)


coldironwarrior:
probably a bit more complex but i'd rather see blacklist on voting the same map 3+ times in a row. going back-to-back is okay, usually. 
given our current map rotation, sometimes there really is only one map people want to be playing for whatever the server pop/balance is at the time.

Lumipharon:
Voting the same map 3 times in a row is cringe but... Its literally the players choosing, so I don't see why we'd take away their choice?

Lumipharon:
I still don't particularly agree with taking player's choice away (even if it is very cringe to repeatedly vote the same map) but if other maints want to merge then ogey

PR #15487, opened 2024-03-28
Makes basic med pills and their chem bottles infinite in med vendor and makes analyzers infinite too.

Naaanii:
## About The Pull Request
Title.
## Why It's Good For The Game
Makes little sense that these pill bottles are infinite in the marine med vendor and not in the actual med vendor.
Analyzers are infinite in their own way via Vali so they should probably be infinite in med vendors too.
## Changelog
:cl:
balance: Already infinite pill bottles and chem bottles in marine medivend are now also infinite in med vendors.
balance: Scanners are now infinite in med vendors.
/:cl:


QualityVan (commented):
Two things about this. First, about a third of the PR is missing from the title/desc/CL. Second, 'it's available as a module so it should be infinite as a small item too' is incredibly goofy.

Naaanii:
> Two things about this. First, about a third of the PR is missing from the title/desc/CL. Second, 'it's available as a module so it should be infinite as a small item too' is incredibly goofy.

And having armor modules, which also gives instant scans in infinite supply but not regular scans make it less goofy...?

Kuro020:
Remove analyzers from this pr, leave only pill bottles

Kuro020 (changes requested):
what was said above

Xander3359:
@Naaanii You forgot to update PR title, CL and "Why it's good for the game"

PR #15490, opened 2024-03-28
autodoc code comments on __skills.dm define

coldironwarrior:
## About The Pull Request

title, just cleans up the comments on various skills and autodocs them.

## Why It's Good For The Game
autodocs them for anyone who might be editing skillsets later
puts strings for base skills with the defines for the various levels and values of each
fixes MISINFO about firearms trainings
notes which skills are broken (in effect; powerloader is broken, and pilot isn't used anywhere)

## Changelog
no player-facing changes; this is just documentation fixes


code/__DEFINES/skills.dm
```diff
@@ -24,140 +24,140 @@
...
-#define SKILL_FIREARMS_DEFAULT 1	//marines (allow tactical reloads)
-#define SKILL_FIREARMS_TRAINED 2	//special training
+// firearms skill (general knowledge of guns) (hidden skill)
+// increase or decrase accuracy, recoil, and firing delay of most guns.
+#define SKILL_FIREARMS_UNTRAINED 0	// civilian (Researcher, MD, CL, SO)
+#define SKILL_FIREARMS_DEFAULT 1	// marines (allow tactical reloads)
+#define SKILL_FIREARMS_TRAINED 2	// special training (CO)
 
-///Damage mod for having the weapon specific skill above 0
+// Damage mod for having the weapon specific skill above 0
```
Xander3359, line 33: Don't remove this autodoc

code/__DEFINES/skills.dm
```diff
@@ -24,140 +24,140 @@
...
+
+// melee_weapons skill
+// buff to melee weapon attack damage
 #define SKILL_MELEE_WEAK -1
 #define SKILL_MELEE_DEFAULT 0
 #define SKILL_MELEE_TRAINED 1
 #define SKILL_MELEE_SUPER 2
 
-///The amount of extra damage per melee skill level
+// The amount of extra damage per melee skill level
```
Xander3359, line 146: Don't remove this autodoc

Xander3359 (changes requested):

Lumipharon:
If you're going to edit these, please just make them autodoc

Lumipharon (changes requested):
autodoc has to be above the define. All the comment shit thats in line with every define needs to be moved

Xander3359:
> autodoc has to be above the define. All the comment shit thats in line with every define needs to be moved

I didn't think that had to be autodoced since that doesn't really explain what it does.

Lumipharon:
Autodocing defines isn't mandatory, but autodoc is by default more useful than a normal comment. So if we're going to clean up comments, they might as well be a bit more useful and be autodoc.

coldironwarrior:
> autodoc has to be above the define. All the comment shit thats in line with every define needs to be moved

oki, sorry about this; will fix

code/__DEFINES/skills.dm
```diff
@@ -24,120 +24,200 @@
...
+#define SKILL_HEAVY_WEAPONS_TRAINED 1
+
+/**
+ * Smartgun skill; Increase or decrase accuracy, recoil, and firing delay for smartgun, and whether we can use smartguns at all.
+ * DEFAULT = Big negative so the effects are far worse than pistol/rifle untrained
+ * USE = Can use a smartgun (CO)
+ * TRAINED = Smartgunners
+ * EXPERT = Unused
+ * MASTER = Deathsquad
+ */
```
Lumipharon, line 87: oh yeah, you should probably these uh, accurate.
Increasing skill improves acc, scatter and damage. It does not improve firing speed (although for heavy I think it improved reload speed?

Lumipharon (changes requested):
When I said autodoc I mean like, doc'ing each define instead of the comments on each line. Having a multiline autodoc on the top means you're only giving autodoc for the top value while the rest are blank. Most of these explainations are more relevant for the actual skil TYPE defines. i.e. #define SKILL_MELEE_WEAPONS "melee_weapons"

That said, a lot of these old comments could also probs just be deleted, but your call

code/__DEFINES/skills.dm
coldironwarrior: in gun_system.dm, lines 1736-1751:

```/obj/item/weapon/gun/proc/gun_on_cooldown(mob/user)
	var/added_delay = fire_delay
	if(user)
		if(user.skills.getRating(SKILL_FIREARMS) < SKILL_FIREARMS_DEFAULT)
			added_delay += 3 //untrained humans fire more slowly.
		else
			switch(gun_skill_category)
				if(SKILL_HEAVY_WEAPONS)
					if(fire_delay > 1 SECONDS) //long delay to fire
						added_delay = max(fire_delay - 3 * user.skills.getRating(gun_skill_category), 6)
				if(SKILL_SMARTGUN)
					if(user.skills.getRating(gun_skill_category) < 0)
						added_delay -= 2 * user.skills.getRating(gun_skill_category)
	var/delay = last_fired + added_delay
	if(gun_firemode == GUN_FIREMODE_BURSTFIRE)
		delay += extra_delay```

coldironwarrior:
i did all the clean-up; i ended up re-orging the file because i do thing keeping the info about what levels go to which people is useful, especially if anyone makes new jobs and wants to see who's comparable in skills

unfortunately all the comments won't appear for me on autodoc except for the firearms one? to my best knowledge, they're all formatted identically to that one, so i don't know why it's just that one...

coldironwarrior:
me when checks fail on [checks hand] prison station because of REASON #1: Assertion failed: No weeds found in neighbouring tiles after 5 seconds!

Lumipharon (approved):

PR #15496, opened 2024-03-28
fixes minor things on sulaco

coldironwarrior:
## About The Pull Request

title, as with all big edits, stuff hidden under tables and the like now becomes clear
moves intercoms in prep back onto the wall
moves a camera in the pharmacy that i think overlapped with the chemmaster too much
deletes two emergency closets that won't stick to the walls correctly and look ugly
moves medbay hall sign to not be on a window
makes tanker area a little bigger
adds tank console to tanker area
button tied to tanker area shutters now only works on tanker area to prevent them going into MP area

## Why It's Good For The Game
 oops
also some tankers were doing Bad Things in the mech area, though that's fixed now but juuuuuuuust to be sure they don't go where they don't need to be

## Changelog
:cl:
fix: makes the tanker area on sulaco a bit bigger and more feature complete
fix: shutters to tank workshop now arent on the same button as shutters to mech workshop
/:cl:


TiviPlus (approved):

RipGrayson (changes requested):
carp7
You doubled the intercoms here.

PR #15500, opened 2024-03-29
Campaign mission: Raiding Base

Lumipharon:
## About The Pull Request
New mission, just for TGMC until I have a map for the SOM to use.

TGMC forces are trying to assault a SOM base hidden in TG territory that has been raiding their supply lines.
As the base is underneat a mountain, the TGMC are aiming to deploy a orbital beacon into the base itself. If they can defend it long enough, an orbiting destroyer will launch a precision guided, bunker busting missile to blow up the base, disabling it for good.

SOM forces need to repel the TGMC forces from their base and if necessary, C4 any orbital beacon deployed before target lock is aquired.

Marines have three beacons, so lose if they run out of time, beacons, or manpower.

Removes supply raid maluses (or in the future removes the enemy's ability to play supply raid if they haven't already, but not enabled yet because we have only a small map pool).

Added callback functionality to the deployable item component to add extra checks for deploying (in the case of this pr, we specify a set of areas the beacon can be deployed in, which you can check by examining the beacon).

Fixed a random door not being opaque when it should be. Also fun fact: If a double door dir is specified as 2 instead of 1, it fucks up the opacity and it will only be opaque on one tile. wew.

This mission is designed to fit multitile vehicles, so in the future we may see them here.
## Why It's Good For The Game
New mission gud.
## Changelog
:cl:
add: Campaign: Added a new mission - Raiding Base for TGMC
/:cl:


code/datums/components/deployable_item.dm
```diff
@@ -6,13 +6,17 @@
...
 
-/datum/component/deployable_item/Initialize(_deploy_type, _deploy_time, _undeploy_time)
+/datum/component/deployable_item/Initialize(_deploy_type, _deploy_time, _undeploy_time, _deploy_check_callback)
 	if(!isitem(parent))
 		return COMPONENT_INCOMPATIBLE
 	deploy_type = _deploy_type
 	deploy_time = _deploy_time
 	undeploy_time = _undeploy_time
+	if(_deploy_check_callback)
+		deploy_check_callback = _deploy_check_callback
```
TiviPlus, line 19: why the if though

code/datums/gamemodes/campaign/missions/raiding_base.dm
```diff
@@ -1,97 +1,176 @@
...
+	map_text_broadcast(starting_faction, "We've lost [beacons_destroyed] beacon[beacons_destroyed > 1 ? "s" : null], get it together!", "Overwatch")
+	map_text_broadcast(hostile_faction, "[beacons_destroyed] beacon[beacons_destroyed > 1 ? "s" : null] destroyed, keep up the good work!", "Overwatch")
+
+///Reacts to an OB beacon being successfully triggered
+/datum/campaign_mission/raiding_base/proc/beacon_triggered(datum/source, obj/structure/campaign_objective/destruction_objective/bunker_buster/beacon, activation_delay)
+	SIGNAL_HANDLER
+	deltimer(game_timer) //stops the game from ending if it comes down to the wire
+	addtimer(CALLBACK(src, PROC_REF(beacon_effect), beacon, beacon.loc), activation_delay)
+
+///Makes the bang
```
TiviPlus, line 169: Not a very descriptive autodoc

TiviPlus (changes requested):

code/datums/components/deployable_item.dm
Lumipharon: I have no idea

code/datums/gamemodes/campaign/missions/raiding_base.dm
Lumipharon: updated

RipGrayson (changes requested):
carp6
This is really the only outright mapping error I see here, otherwise approved.

RipGrayson (approved):

PR #15502, opened 2024-03-29
AT-36 ammo price increase and amount decrease

coldironwarrior:
## About The Pull Request

- AT ammo price for the 3-shell kits moves from 20 points to 40 points; or 13.3 points per shells
- AT ammo reqtorio kits moves from 100 points to 200 points; or 7 points per shell
- AT HE shell pack (x3) included 2 HE shells and 1 standard shell; this fixes that and makes them all HE

## Why It's Good For The Game

AT is devastatingly good but refills for it are pretty trivial right now. the work of a single well-placed AT shell is definitely worth way more than 3.3 req points. this keeps the supply pipeline for AT a little more in-parallel to the game impact the weapon has, while still providing decent room for one to do TATconomics and turn a profit off of kills to fund more shells to get more kills, etc.

## Changelog
:cl:
balance: AT-36 reqtorio supply packs now have 200 points 
fix: fixed the AT HE shell pack in req including 2 HE and 1 standard shell instead of 3 HE
/:cl:

### DIFF
--- a/code/modules/factory/unboxer.dm
+++ b/code/modules/factory/unboxer.dm
@@ -484,31 +484,31 @@
 	name = "box of rounded metal plates"
 	desc = "A box with round metal plates inside. Used to refill Unboxers."
 	refill_type = /obj/item/factory_part/atgun_aphe
-	refill_amount = 15
+	refill_amount = 30
 
 /obj/item/factory_refill/atgun_apcr_refill
 	name = "box of rounded metal plates"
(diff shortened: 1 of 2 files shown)


TiviPlus (approved):
Code ok

Kuro020 (changes requested):
These nerfs are way too extreme, double the costs for a start and see how it goes

PR #12505, opened 2023-03-24
Gives CMO Master Surgery and Medical skill

Naaanii:
## About The Pull Request
Title
## Why It's Good For The Game
The chief is and should be the best at their craft.
## Changelog
:cl:
balance: changes CMO surgery and medical skill to Master.
fix: fixes some descriptions in code to be more accurate
/:cl:

PR #15503, opened 2024-03-29
suturing with above expert med skill gives more healing than 30 brute/burn damage

coldironwarrior:
## About The Pull Request

edits the healing of sutures to be 37.5 damage max, with the formula for the healing mutliplier using skill_master as the divisor, instead of skill_expert. tldr: buff to damage healed by CMO suturing

## Why It's Good For The Game

#12505 gave CMO level 5 med and surgery skill; the way this is used in code means that all the surgery steps had their time reduced.
there is a separate formula on suturing as a step to calculate how much suturing heals. the base is 30 damage split between brute and burn, but the formula uses SKILL_SURGERY_EXPERT (4) as the divisor and clamps it at that.
what this means is that those with surgery skill level 4 maximize healing from sutures. those (the CMO) with surgery skill level 5 don't get any additional healing from sutures, which seems odd that a doctor would or should be just as good as the CMO.
this allows healing with sutures by CMOs (and spatial agents?) to provide an additional 7.5 damage healed, for 37.5 total; while no other role sees any reduction or increase in the amount healed from suturing

## Changelog
:cl:
balance: CMOs now heal an additional 7.5 damage per use of sutures
/:cl:

### DIFF
--- a/code/modules/surgery/generic.dm
+++ b/code/modules/surgery/generic.dm
@@ -314,7 +314,7 @@
 	user.visible_message(span_notice("[user] sews some of the wounds on [target]'s [affected.display_name] shut.") , \
 	span_notice("You finish suturing some of the wounds on [target]'s [affected.display_name].") )
 	target.balloon_alert_to_viewers("Success")
-	var/skilled_healing = base_healing * max(user.skills.getPercent(SKILL_SURGERY, SKILL_SURGERY_EXPERT), 0.1)
+	var/skilled_healing = base_healing * (user.skills.getPercent(SKILL_SURGERY, SKILL_SURGERY_EXPERT), 0.1)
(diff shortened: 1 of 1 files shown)


TiviPlus:
why the random number

TiviPlus:
also missing other changes in CL

coldironwarrior:
> why the random number

it's what the math worked out to; if the formula is
var/skilled_healing = base_healing * max(user.skills.getPercent(SKILL_SURGERY, SKILL_SURGERY_EXPERT), 0.1)

then 30*(max(SKILL/4)*0.1) for CMO is 30*(max(5/4)*0.1), so 1.25 times 30, which is 37.5

TiviPlus:
@Kuro020

Kuro020 (approved):

PR #15506, opened 2024-03-29
Increase the AI cargo droid movespeed

lbnesquik:
## About The Pull Request
Reduces the AI cargo droid movespeed delay. 
## Why It's Good For The Game
The AI drone is actually painfully slow to use. This thing moves as slow as a ripley with a PVT using it, without as much functionnality. It just makes sense that something with less capacity moves faster. 
## Changelog
:cl:
qol: Reduced the AI droid movespeed delay
/:cl:

### DIFF
--- a/code/modules/vehicles/unmanned/unmanned_droid.dm
+++ b/code/modules/vehicles/unmanned/unmanned_droid.dm
@@ -131,7 +131,7 @@
 	desc = "A cargo droid, rigged with experimental technology to allow AI control. The claw is not standard and cannot grasp warheads."
 	icon = 'icons/obj/powerloader.dmi'
 	icon_state = "ai_powerloader"
-	move_delay = 7
+	move_delay = 5
 	max_integrity = 550
 	spawn_equipped_type = null
 	unmanned_flags = GIVE_NIGHT_VISION


Lumipharon:
Everyone moves at the same speed in a ripley, so this is just buffing AI to be a better ripley.

TiviPlus (approved):
code ok

TiviPlus:
Also yea what lumi says this feels like creep

lbnesquik:
> Everyone moves at the same speed in a ripley, so this is just buffing AI to be a better ripley.

That would be true if it could do everything a ripley can. If we could move OB shells, sentries, pods or dropship gear, I'd totally be fine to have that speed. But given you can't do that, I don't feel like it's really that comparable.

Kuro020:
> > Everyone moves at the same speed in a ripley, so this is just buffing AI to be a better ripley.
> 
> That would be true if it could do everything a ripley can. If we could move OB shells, sentries, pods or dropship gear, I would absolutely be fine staying at the current movement speed. But given you can't do that, I don't feel like it's really that comparable.
> 
> I would also add that a lot of the AI features over the years are straight buffs and QoL because it started out in such a deep pit of being boring / broken, so creep is one word for it. We are generally moving towards less obstacles for everything, and less tedium, with quick build, less medical fumble, now generators being preset... Not moving at a snail pace when trying to play with a feature seems fine to me.

then make it able to do those things, this is creep

PR #15523, opened 2024-03-30
Silo scaling change

homexp13:
## About The Pull Request

Changed the rate scale formula in larva generation, so ratio of marines to xenos means something

## Why It's Good For The Game

The rate based on the current ratio of humans to xenos is simply ridiculous. for example, if the ratio of marines to xenos is 1 to 3 (1 marine and 3 xenos, is correct), there is still no reduction in the generation of larva .
Also, https://github.com/tgstation/TerraGov-Marine-Corps/pull/13106 says:

> So with this PR, if the ratio is say 2 ungas for every xeno, the larva gen would be about 65% of what it would currently be (in addition to all the extra larva points for murdering enough marines to get to that ratio).

but calculation give different result (I'm too lazy to use a clamp)
image
so this is how it is with this PR
image
should this be a bug?

## Changelog

:cl:
balance: Silo larva gene now decreases more depending on the marine to xenos ratio.
/:cl:

### DIFF
--- a/code/controllers/subsystem/silo.dm
+++ b/code/controllers/subsystem/silo.dm
@@ -27,7 +27,7 @@ SUBSYSTEM_DEF(silo)
 	current_larva_spawn_rate *= SSmonitor.gamestate == SHIPSIDE ? 3 : 1
 	current_larva_spawn_rate *= SSticker.mode.silo_scaling
 	//We scale the rate based on the current ratio of humans to xenos
-	current_larva_spawn_rate *= clamp(round((active_humans / active_xenos) / (LARVA_POINTS_REGULAR / xeno_job.job_points_needed), 0.01), 0.5, 1)
+	current_larva_spawn_rate *= clamp(round((active_humans / active_xenos) * (LARVA_POINTS_REGULAR / xeno_job.job_points_needed), 0.01), 0.5, 1)
 
 	current_larva_spawn_rate += larva_spawn_rate_temporary_buff
 


Notamaniac:
I was going to say that this was going to nerf the larva spawn rate by a factor of 8, but the value is clamped between 0.5 and 1, which was why this probably wasn't noticed- it was constantly stuck at 1.

So while it isn't a massive nerf, xenoes are definetly going to feel it.

code/controllers/subsystem/silo.dm
```diff
@@ -27,7 +27,7 @@ SUBSYSTEM_DEF(silo)
 	current_larva_spawn_rate *= SSmonitor.gamestate == SHIPSIDE ? 3 : 1
 	current_larva_spawn_rate *= SSticker.mode.silo_scaling
 	//We scale the rate based on the current ratio of humans to xenos
-	current_larva_spawn_rate *= clamp(round((active_humans / active_xenos) / (LARVA_POINTS_REGULAR / xeno_job.job_points_needed), 0.01), 0.5, 1)
+	current_larva_spawn_rate *= clamp(round((active_humans / active_xenos) * (LARVA_POINTS_REGULAR / xeno_job.job_points_needed), 0.01), 0.5, 1)
```
BraveMole, line 30:
```suggestion
    var/current_human_to_xeno_ratio = active_humans / active_xenos
    var/optimal_human_to_xeno_ratio = xeno_job.job_points_needed / LARVA_POINTS_REGULAR
	current_larva_spawn_rate *= clamp(current_human_to_xeno_ratio/optimal_human_to_xeno_ratio , 0.5, 1)
```

BraveMole: Excellent catch

BraveMole:
TM despite this being a fix, because the origial PR never did anything

Kuro020 (approved):
let's see how the game looks after this

Kuro020 (changes requested):
This completely fucks xenos sideways, the ratio decrease needs to be uhh, decreased

homexp13:
I think this should be ok

TiviPlus (approved):

PR #15528, opened 2024-03-30
Add extra clone bays to Theseus medical

lbnesquik:
## About The Pull Request
This PR adds extra clone bays to TGS Theseus. It also shuffles the position of some gear and extends the room a little to make space. It also adds an emergency gun cabinet in the corner.

## Why It's Good For The Game
I always liked the fact our ships had unique traits, be it from pods capacity, layout making holding, or deploying easier. However, I figure we can go further in the variations, like having some ships support certain styles of play better. A bit like our seasonals in the weapons vendors. 

That's where this PR comes in. Basically, we get some of those weakass early vatborns for people to play around with, at the cost of having to spend some dosh on biomass to run them all. I do not forsee it being a substantial meta change, but I think it would be noticeable fluff, which changes from the repetition. 

## Changelog
:cl:
add: Added extra clone bays to TGS Theseus 
/:cl:

### DIFF
--- a/_maps/map_files/Theseus/TGS_Theseus.dmm
+++ b/_maps/map_files/Theseus/TGS_Theseus.dmm
@@ -52923,7 +52923,7 @@ pES
 hjE
 bjV
 biu
-wyX
+nSl
 wyX
 nBH
 uEK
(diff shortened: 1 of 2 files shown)


74delta:
based

PR #15465, opened 2024-03-26
The toxin reagent from QC+, Peri+ and Nanoblood shows on scanners

Barnet2:
## About The Pull Request
What it says on the tin.
## Why It's Good For The Game
No reason to really be hidden from the player, just hurts newer players in the long run for no reason at all. 
image
## Changelog
:cl:
qol: The toxin reagent from QC+, Peri+ and Nanoblood now shows on medical scanners.
/:cl:

PR #15326, opened 2024-03-09
Water is now recognizable by medical scanners

lbnesquik:
## About The Pull Request
When you scan someone with water, it should now not say it is an unknown reagent.

Fixes https://github.com/tgstation/TerraGov-Marine-Corps/issues/14537

## Why It's Good For The Game
Water shouldn't be a mystery reagent.

## Changelog
:cl:
fix: water is now scannable by medical scanners
/:cl:

PR #15551, opened 2024-04-02
Make all reagents scannable by default

Xander3359:
## About The Pull Request
Makes all chems scannable by a health scanner by default
## Why It's Good For The Game
As we've seen in:
https://github.com/tgstation/TerraGov-Marine-Corps/pull/15326
https://github.com/tgstation/TerraGov-Marine-Corps/pull/15465
People want to be able to see what chems are circulating through their system.
I don't see any reason why we would want to have invisible chems, so imo they should all be made scannable in a single PR, rather than slowly make every single chem scannable individually.
## Changelog
:cl:
qol: All chems (except neuraline) are scannable by a health scanner
/:cl:

### DIFF
--- a/code/modules/reagents/reagents.dm
+++ b/code/modules/reagents/reagents.dm
@@ -48,7 +48,7 @@ GLOBAL_LIST_INIT(name2reagent, build_name2reagent())
 	/// increases as addiction gets worse
 	var/addiction_stage = 0
 	/// does this show up on health analyzers
-	var/scannable = FALSE
+	var/scannable = TRUE
 	/// if false stops metab in liverless mobs
 	var/self_consuming = FALSE
 	/// List of reagents removed by this chemical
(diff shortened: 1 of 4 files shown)


code/modules/reagents/reagents.dm
```diff
@@ -48,7 +48,7 @@ GLOBAL_LIST_INIT(name2reagent, build_name2reagent())
 	/// increases as addiction gets worse
 	var/addiction_stage = 0
 	/// does this show up on health analyzers
-	var/scannable = FALSE
+	var/scannable = TRUE
```
BraveMole, line 51: then scannable is true for everything; so you need to remove from the code.

But I doubt this will go through walance

BraveMole (changes requested):

code/modules/reagents/reagents.dm
Xander3359: I left it since neuraline seems to want to be non-scannable specifically (For some reason?)
So I left that one alone.
But arguably that one could be allowed to be scanned so idk /shrug

Issue #15535, opened 2024-03-31
Ricochet death tile bug


HanderBoy:
### Description

The ricochet creates duped projectiles when interacting with a solid wall and a obstructing object ex. doors, shutters, resin walls even that creates a stream of 28(I counted the hits multiple times) invisible ricochet projectiles that hit everything that comes in contact with that tile.
The bug is mainly happens because of the buggy behaviour the ricochet sniper rifle has with anything that it can hit and is NOT a wall aditionally if you hit it in the wall in front of you the beam will IGNORE you and pass through to ricochet anything else if the tile behind you also has no space between you it will phase through you AGAIN into an infinite loop until the rico dries of beams.

### Test Merges

Round ID 26381 

### Reproduction Steps

1.Put the las sniper on ricochet mode
2.Sandwich yourself between a wall tile and any sort of non ricocheable tile .shutter doors .normal doors .worked on resin walls once
3.Fire
4.You created a invisible death trap that has a 20% chance to kill you if you step out of it have fun!

https://github.com/tgstation/TerraGov-Marine-Corps/assets/76793032/333df996-8dfc-42ba-b4a5-997f23786306

### Screenshots

Send a video up in reproduction

Notamaniac:
Okay so it triggers a max recursion exception, which irrc is stopping the laser from processing and causing it to hang about, but it technically isn't recursion. From trying to interpret the callstack, it doesn't properly see the door as a dense object, but bounces off it regardless and fires directly from the open turf hence why it doesn't hit us. The amount a shot can bounce around is determined by it's range, and that range var is going down by one everytime it bounces. It's just that the range var is 30, and that's the max level of recursion that can occur before the exception triggers.

So option A : Do a check for airlocks/doors/objects on scan_a_turf, instead of just checking the turfs density
option B : hard limit the number of bounces to a number that's below 30
option C : remove bounce completely

HanderBoy:
> Okay so it triggers a max recursion exception, which irrc is stopping the laser from processing and causing it to hang about, but it technically isn't recursion. From trying to interpret the callstack, it doesn't properly see the door as a dense object, but bounces off it regardless and fires directly from the open turf hence why it doesn't hit us. The amount a shot can bounce around is determined by it's range, and that range var is going down by one everytime it bounces. It's just that the range var is 30, and that's the max level of recursion that can occur before the exception triggers.
> 
[...]

PR #15556, opened 2024-04-03
Fixes the deathtile bug

HanderBoy:
## About The Pull Request

Someone found out why the bug is happening and as I'm no coder or smart enough to describe it here it is https://github.com/tgstation/TerraGov-Marine-Corps/issues/15535#issuecomment-2033282217
But in the issues section of the bug the guy talked about why it happened and it's because it bounces more than 30 times on the object :sob: so put the range so it bounces less times

## Why It's Good For The Game

I fix the godforsaken bug I unearthed.

## Changelog

Fixes the Deathtile bug for the ricochet
Also nerfs the ricochet by reducing it's range by 4 tiles

:cl:
fix: fixed the deathtile bug
balance: reduces lassniper ricochets firemode reflectable tiles by 4
/:cl:


Lumipharon:
Uh, how does this fix anything? If you have more than 4 range left it will still reflect unchanged?
Main issue is that I made it reflect using the orginal projectile type unconditionally, I meant to do that only behind a arg

Lumipharon:
See #15558

PR #15578, opened 2024-04-05
Raid base map tweaks

Lumipharon:
## About The Pull Request
Added some cave tunnels by the west entrance, also probably added a few decals or some shit.
## Why It's Good For The Game
At lower pop (40-50) the current entrances are manageable (but difficult) for TGMC, but at high pop like when we tested at 80-90 people 3 times, its extremely hard to push the chokepoints.

Adding a 4th should help marines split the SOM up more when you have too many people to just throw at the other entrances, while still making it a horrific death trap area that some x-fuel flamer guy is going to have a field day with.
## Changelog
:cl:
balance: Campaign: Adjusted the Raiding Base map slightly
/:cl:


RipGrayson:
This is probably a good use-case for population dependent modulars, but I'm not forcing it.

Lumipharon:
Yeah I hadn't considered that, so I might just do that

PR #15581, opened 2024-04-05
Gun skill tweak

Lumipharon:
## About The Pull Request
Removes skill modifiers for fire delay based on user skill.

Makes skill level effect reload delay for mags that actually have it.

Also cleaned up some autodoc
## Why It's Good For The Game
These modifiers almost never came or could come into practice due to them only being checked when starting fire (i.e. auto/burst fire ignores it anyway) so were effectively pointless other than punishing like the CMO for semi auto firing a gun or something.

Adjusting reload delay is more useful in that it actually will do something, but this should make no particular change for HvX.

In HvH it means skill perks can help with slow loading guns like MG60/RPG/ALF etc.
## Changelog
:cl:
balance: Gun skills no longer effect fire delay for semi auto firing
balance: Gun skills effect reload delay for magazines that have one
/:cl:

### DIFF
--- a/code/modules/projectiles/gun_system.dm
+++ b/code/modules/projectiles/gun_system.dm
@@ -1309,9 +1309,11 @@
 		(!((loc == user) || (master_gun?.loc == user)) || (new_mag.loc != user)))
 			to_chat(user, span_warning("You need to be carrying both [src] and [new_mag] to connect them!"))
 			return FALSE
-		if(get_magazine_reload_delay(new_mag) > 0 && user && !force)
+		var/reload_delay = get_magazine_reload_delay(new_mag)
+		if(reload_delay > 0 && user && !force)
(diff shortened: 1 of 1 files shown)


coldironwarrior:
researchers everywhere, rejoice

Kuro020 (approved):

SzkieletorNutric:
> so were effectively pointless other than punishing like the CMO for semi auto firing a gun or something

This penalty is especially punishing towards roles with 0 gun skill that may end up groundside from the start; i.e. Researchers, extra MDs, and the occasional bold CLs.
This had always effectively cut their damage output in half with things like handguns, machineguns, and most noticeably, shotguns.

As fun as it'd be to be finally rid of this malus, are you quite certain that this is a wise change in the long-term?

PR #15585, opened 2024-04-05
Makes synthetics scannable

jonash07:
## About The Pull Request

Title

## Why It's Good For The Game

I don't see a reason why they shouldn't be, as it can be a great help to corpsmen/medical doctors to revive or repair synthetics.

## Changelog

:cl:
balance: Made synthetics scannable
/:cl:


ivanmixo:
not qol and do not try to pass it off as such

jonash07:
what does it count as then

coldironwarrior:
not into this but from a flavor standpoint, nothing to do with balance or qol; i didn't like robots becoming scannable either, i just like the different experience they require and intuitiveness it cultivated for players to know when and how they were too damaged

QualityVan:
I was gonna close this but apparently combat robots have been scannable for a year? Which is weird, but does make this feel more like parity.

PR #15590, opened 2024-04-06
Removes puppeteer orders ability and scout order, makes attack and recall orders normal abilitys

Dikozda:
## About The Pull Request
Makes the puppeteers recall and attack orders normal abilitys with keybinds, removes the order ability, its radial menu and the scout order, also removes a image list since it will be no longer used anywhere with these changes.
## Why It's Good For The Game
Having to fumble through a radial menu to recall your puppets before they die or trying to send them out to attack is a pain, it is a extra step for abilitys that you would want to use as fast as possible. having them as keybinds makes it so much nicer to use and feel less annoying.

This also means that puppeteers can recall their puppets faster and perhaps lose their puppets less as a result which kinda sounds like a balance thing but honestly it seems kinda minor.

As for removing the scout ability, i have never found a use for this ability and feels like just sending your puppets out to die in random corners of the map and making it also a ability just seemed pointless.

## Changelog
:cl:
del: Removes puppeteer scout order and order radial menu.
qol: Puppeteer attack and recall orders are now normal abilitys and have keybinds.
/:cl:


Dikozda:
can't wait to find out i broke something major that i did not notice in testing

blackdav123:
scout isnt entirely useless as you can use it to keep your puppets in a safe spot while you go fight a marine that would kill your puppets very quickly, like when you just want to farm a little biomass off of a flamer or grenade marine

BraveMole (approved):

PR #15592, opened 2024-04-06
Increases RG 220 wield delay

Yttriums:
## About The Pull Request
ided pr
Increases RG 220 wield delay from 1.2 to 2.0 seconds.
## Why It's Good For The Game

Currently, RG 220 is a very powerful operational weapon which easily farms kills and snowballs the game for marines. 

Because it's wield time is significantly shorter than other weapons of it's size/strength, it's incredibly safe to run as a primary weapon - you can wield it quickly out of stuns and force xenos off of you with a single shot.

This change forces RG220 users to carry a backup weapon or consider their positioning more carefully, similar to other weapons of it's caliber.
## Changelog
:cl:
balance: Increased RG220 wield delay from 1.2 to 2.0 seconds
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/specialist.dm
+++ b/code/modules/projectiles/guns/specialist.dm
@@ -994,7 +994,7 @@ Note that this means that snipers will have a slowdown of 3, due to the scope
 		/obj/item/ammo_magazine/railgun/hvap,
 	)
 	force = 40
-	wield_delay = 2.0 SECONDS
+	wield_delay = 2.4 SECONDS
 	attachable_offset = list("muzzle_x" = 33, "muzzle_y" = 18,"rail_x" = 31, "rail_y" = 23, "under_x" = 19, "under_y" = 14, "stock_x" = 19, "stock_y" = 14)
 	attachable_allowed = list(
 		/obj/item/attachable/magnetic_harness,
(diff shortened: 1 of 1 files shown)


Lumipharon:
This comes back to it being a real poor idea to give it more than one shot per reload.

Yttriums:
> This comes back to it being a real poor idea to give it more than one shot per reload.

Maybe true. Still, marines shouldn't be power cleaning this weapon up to their shoulders in less time than an HPR

Naaanii:
It already has a high fire delay and can't be used really well with other weapons because of said fire delay. (Funny juggle PR). You're better off removing the extra shots per mag.

Yttriums:
> It already has a high fire delay and can't be used really well with other weapons because of said fire delay. (Funny juggle PR). You're better off removing the extra shots per mag.

My point about using a backup weapon has nothing to do with juggling - the point is that you can fire this large, heavy weapon too quickly out of getting stunned. You shouldn't be able to wake up from a stun/knockdown and use this within 1 second - it's a man-sized railgun. You should have to reach for a secondary.

I'm not interesting in nerfing the weapon into the ground - I just want it to play like a heavy weapon.

Kuro020 (changes requested):
2.4 is too much, make it 2.0 and we'll see

lbnesquik:
I think we can give it a try.

TiviPlus (approved):

PR #15609, opened 2024-04-08
Storage opening QOL

BraveMole:
## About The Pull Request
Clicking on a storage with an item that cannot be inserted in said storage will open it

## Why It's Good For The Game
QOL, reduce jungling a bit when trying to reload.

For example, if you thought about opening your ammunition belt before the fight, you can do a tactical reload.
If you forgot to have the storage screen opened, you need to fumble a bit which is clonky. 

## Changelog
:cl:
qol: Clicking on a storage with an item that cannot be inserted in said storage will open it
/:cl:

### DIFF
--- a/code/game/objects/items/storage/storage.dm
+++ b/code/game/objects/items/storage/storage.dm
@@ -559,7 +559,7 @@
 				return do_refill(I, user)
 
 	if(!can_be_inserted(I))
-		open(usr)
+		open(user)
 		return FALSE
 	return handle_item_insertion(I, FALSE, user)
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/storage/storage.dm
```diff
@@ -559,6 +559,7 @@
 				return do_refill(I, user)
 
 	if(!can_be_inserted(I))
+		open(usr)
```
ivanmixo, line 562:
```suggestion
		open(user)
```
x3

Lumipharon (approved):

PR #15613, opened 2024-04-08
Makes lascarbine spread stats more in line with combat shotgun buckshot's

blackdav123:
## About The Pull Request
Increases the projectile scatter, lowers accurate range, adds damage falloff.
## Why It's Good For The Game
Lascarbine spread generally has very few of the drawbacks that typically come with shotguns, most notably being just as effective at range. With this change, spread will no longer be able to hit all three projectiles outside of a ~3 tile range and will have very weak damage at max range, killing a runner at range at about the same speed as the combat shotgun would with buckshot. For people who want to full sprint and rush at xenos this change will not affect them much, but people who want to sit back and snipe with a shotgun should switch to a different firemode.
## Changelog
:cl:
balance: Reduced lascarbine spread mode's effectiveness at range.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammo_datums.dm
+++ b/code/modules/projectiles/ammo_datums.dm
@@ -3096,12 +3096,13 @@ GLOBAL_LIST_INIT(no_sticky_resin, typecacheof(list(/obj/item/clothing/mask/faceh
 	hud_state = "laser_spread"
 	bonus_projectiles_type = /datum/ammo/energy/lasgun/marine/blast/spread
 	bonus_projectiles_amount = 2
-	bonus_projectiles_scatter = 5
+	bonus_projectiles_scatter = 10
 	accuracy_var_low = 9
 	accuracy_var_high = 9
-	accurate_range = 5
+	accurate_range = 3
 	max_range = 8
 	damage = 35
+	damage_falloff = 8
 	penetration = 20
 	sundering = 1
 	hitscan_effect_icon = "pu_laser"


Breadeagle:
Unlike buckshot, scatter doesn't do stagger, knockback, or knockdown. Its more comparable with flechettes, and even then it does less damage and less sunder than flechettes. So, I don't really see the reasoning behind this. It already (imo) wasn't a super powerful fire mode, since the gun has auto burst which is better at literally every range that isn't literally pb and especially not once you can miss 1 of the laser shots from spread.

blackdav123:
> Unlike buckshot, scatter doesn't do stagger, knockback, or knockdown. Its more comparable with flechettes, and even then it does less damage and less sunder than flechettes. So, I don't really see the reasoning behind this. It already (imo) wasn't a super powerful fire mode, since the gun has auto burst which is better at literally every range that isn't literally pb and especially not once you can miss 1 of the laser shots from spread.

unlike buckshot this has 0 recoil and 0 knockback which makes it ideal for rushing down targets which this pr doesnt change. use the damn autoburst if you want to shoot things at range!! shotguns should not be landing every pellet at 5 tile range

Lumipharon:
As above, I have no idea why you're comparing this to buckshot when its the equiv of flechette... which has very tight spread

MLGTASTICa:
> Unlike buckshot, scatter doesn't do stagger, knockback, or knockdown. Its more comparable with flechettes, and even then it does less damage and less sunder than flechettes. So, I don't really see the reasoning behind this. It already (imo) wasn't a super powerful fire mode, since the gun has auto burst which is better at literally every range that isn't literally pb and especially not once you can miss 1 of the laser shots from spread.

100 burst damage with 20% AP and 12 shots, also 12 times easier to reload and more convenient to store ontop of no slowdown. It has crazy Burst potential.

TiviPlus (approved):
Code ok Kuro do walance

Arturlang:
Akimbo carbine rushing is kinda nutty

CVSthePharmacy:
did yall mfs forget that the gimmick of laser weapons is no damage falloff

lbnesquik:
It's a few changes at once. Given how often we tend to have things get wacked hard and never recover, could you remove some of those changes?

PR #15617, opened 2024-04-09
Tad can't land indoors

Lumipharon:
## About The Pull Request
Tad can no longer land in areas with metal roofs. (It can smash through glass roofs just fine though).
## Why It's Good For The Game
The tad is strangely inconsistant with drop pods, and being able to land inside funny buildings just makes tad turtling easier and is generally worse for gameplay.
## Changelog
:cl:
balance: Tadpole can no longer land in metal roofed areas
/:cl:

### DIFF
--- a/code/modules/shuttle/navigation_computer.dm
+++ b/code/modules/shuttle/navigation_computer.dm
@@ -295,7 +295,7 @@
 	if(!T || T.x <= 10 || T.y <= 10 || T.x >= world.maxx - 10 || T.y >= world.maxy - 10)
 		return SHUTTLE_DOCKER_BLOCKED
 	var/area/turf_area = get_area(T)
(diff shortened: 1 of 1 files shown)


Yttriums:
I agree with the drop pod consistency argument, but Tad is almost always easier to kill when it's inside a building. Walls provide cover and limit the turret's sights while vents offer free engagements for defilers and primo sents.

lbnesquik:
WHat does indoor mean in this context? Because prison and other ship type maps might have a slight bad time with that.

Arturlang:
Marine gameplay really centers around the tadpole right now way too much.

TheLeopold:
Turns out these roofs magically stay even if entire building was OB'd to dust.
So TAD is plain useless on most of the map.
Even CAS has more freedom...

PR #15620, opened 2024-04-09
some announcement qol

Neocloudy:
## About The Pull Request
- Orbital bombardment now plays screentext for every human, and changed the content of the message because "get out of danger close" is dumb
  - If possible, I want to make this play over other sources of screentext (Send Orders especially, as it is used almost always before or during an OB) at some point, as a simple narration for the type of an incoming OB is *definitely* more important than whatever the PVT who just unlocked FC has to say. Maybe for a future PR.
- "Send Orders" will now clarify the faction of the sender (EDIT: regardless of TGMC affiliation)
  - `faction_to_acronym` global list for getting faction acronyms to make this possible.

image
image
image
image
## Why It's Good For The Game
- Orbital bombardment playing screentext is a good idea to try and get attention of marines who can't afford to look in chat in the middle of combat
- Being able to tell what faction an announcement is from is good. Also, future coders can use the `faction_to_acronym` global list for something interesting
## Changelog
:cl:
qol: Orbital bombardments now play screentext to humans that includes the type of the OB
qol: Send Orders will now include the sender's faction acronym and job
spellcheck: The content of the orbital bombardment announcement is now "Evacuate the impact zone immediately!" instead of "Get out of danger close!"
code: "faction_to_acronym" global list for getting faction acronyms
/:cl:

### DIFF
--- a/code/modules/screen_alert/command_alert.dm
+++ b/code/modules/screen_alert/command_alert.dm
@@ -78,8 +78,8 @@
 	for(var/mob/faction_receiver in alert_receivers)
 		if(faction_receiver.faction == human_owner.faction || isdead(faction_receiver))
 			var/faction_title = "Command" // the default title, for tgmc announcements
-			if(human_owner.faction != FACTION_TERRAGOV)
-				faction_title = "[GLOB.faction_to_acronym[human_owner.faction]] Command" // if the sender doesn't belong to the TGMC, clarify the faction in the title
+			if(human_owner.faction != FACTION_TERRAGOV) // if the sender doesn't belong to the TGMC, clarify the faction in the title
+				faction_title = GLOB.faction_to_acronym[human_owner.faction] ? GLOB.faction_to_acronym[human_owner.faction] + " Command" : "Unknown Faction" + " Command"
 			faction_receiver.play_screen_text("<span class='maptext' style=font-size:24pt;text-align:center valign='top'><u>[uppertext(faction_title)] ANNOUNCEMENT:</u></span><br>" + text, /atom/movable/screen/text/screen_text/command_order)
 			to_chat(faction_receiver, assemble_alert(
 				title = "[faction_title] Announcement",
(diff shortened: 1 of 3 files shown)


Lumipharon:
why would you use the full name instead of TGMC/SOM/ICC

code/modules/screen_alert/command_alert.dm
```diff
@@ -77,9 +77,12 @@
 		return
 	for(var/mob/faction_receiver in alert_receivers)
 		if(faction_receiver.faction == human_owner.faction || isdead(faction_receiver))
-			faction_receiver.play_screen_text("<span class='maptext' style=font-size:24pt;text-align:center valign='top'><u>COMMAND ANNOUNCEMENT:</u></span><br>" + text, /atom/movable/screen/text/screen_text/command_order)
+			var/faction_title = "Command" // the default title, for tgmc announcements
+			if(human_owner.faction != FACTION_TERRAGOV)
+				faction_title = "[GLOB.faction_to_acronym[human_owner.faction]] Command" // if the sender doesn't belong to the TGMC, clarify the faction in the title
```
Lumipharon, line 82: I wouldn't bother with the exclusion for tgmc tbh.

Lumipharon, line 82: Should probs not assume the faction will always be in the list (even if it should)

Lumipharon (changes requested):

code/modules/screen_alert/command_alert.dm
Neocloudy: how should I implement this? I am basically clueless as to how I would make it check if the leader's faction is in that list or something

Neocloudy: TGMC exclusion is an antique from when I worked on 14957, I want send orders from TGMC leaders to look like a CIC announcement. I'll change this if you really want though

Lumipharon: ternary or something

Lumipharon: its just consistant

Lumipharon (approved):

PR #15639, opened 2024-04-12
removes "intentional" var from emotes, so involuntary emotes have a cooldown applied

Neocloudy:
## About The Pull Request
kills Intentional var on emotes, so now involuntary emotes have a cooldown and involuntarily emotes can't be run a gorillion times in a second. might have to adjust cooldowns for audible emotes for this to work better?
image
## Why It's Good For The Game
Emote spam (from involuntary sources) can be funny/dramatic, but it is very loud or annoying in certain situations, like if you're being gibbed, running through fire or hit by a lot of projectiles, such as buckshot from a hefa grenade or shotgun. Not overloading a player's ears as they scream a fucktillion times seems reasonable.
## Changelog
:cl:
del: removed `intentional` on emotes, making involuntary emotes (such as those from combat) affected by the cooldown, as if they were used intentionally
/:cl:

### DIFF
--- a/code/datums/emotes.dm
+++ b/code/datums/emotes.dm
@@ -40,9 +40,9 @@
 	mob_type_ignore_stat_typecache = typecacheof(mob_type_ignore_stat_typecache)
 
 
-/datum/emote/proc/run_emote(mob/user, params, type_override, intentional = FALSE, prefix)
+/datum/emote/proc/run_emote(mob/user, params, type_override, prefix)
 	. = TRUE
-	if(!can_run_emote(user, TRUE, intentional))
+	if(!can_run_emote(user, TRUE))
 		return FALSE
 	var/msg = select_message_type(user)
 	if(params && message_param)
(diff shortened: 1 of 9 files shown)


Neocloudy:
i hit enter accidentally

Lumipharon:
These are usually/all audio cues that 'badshit be happening', so I don't really see why you'd want to disable them with a cooldown

AwkwardStereo:
screams

This is very funny and I will be sad to see it go.
Don't take it from us, bros.

lbnesquik:
It's worth noting that unlike other spam, this only happens in combat. You kinda have to be dying for that to happen.

Neocloudy:
> These are usually/all audio cues that 'badshit be happening', so I don't really see why you'd want to disable them with a cooldown

Mainly to remove ear-destroying involuntary scream/pain spam in extreme situations. The cooldown seems to be 8 seconds, maybe it could be reduced to be more lenient?

Lumipharon:
I don't really see an issue. Fire specifically may be problematic since it triggers a scream every tick of fire/moving across fire tiles.
But that's an unrelated fix for that.

Neocloudy:
#/15648

PR #15646, opened 2024-04-13
Change FK-88 HE shell description to be less confusing

irRegularGuy646:
## About The Pull Request

changes the FK-88 HE shell's description to say you need to "right click with other hand" to swap rather than simply "activate in hand"

## Why It's Good For The Game

this shit confused the fuck out of me

## Changelog
:cl:
change: updates FK-88 HE shell's description to be less confusing
/:cl:


TiviPlus (changes requested):
firstly shouldnt this be calling parent?
secondly what is this fixing? I see that you're supposed to rclick to swap but this just makes it so you use inhand

irRegularGuy646:
the description for the HE shell says "Activate in hand to swap between unguided and guided modes.". the proc it was calling before needed you to "click it with an empty hand" which was impossible to actually proc since it would just pick it up or swap hands or something. I'm not quite sure what you mean by "calling parent"

Lumipharon:
As mentioned in discord, the right click interaction works fine currently, so this isn't fixing anything.
Also changing this is a balance change since it means you can swap ammo modes while still manning the gun. I (assume) MJP did this to specifically prevent that, although I could be wrong.

irRegularGuy646:
changed PR to make the description less confusing instead of fixing something i thought was broken

Lumipharon (approved):

PR #15661, opened 2024-04-15
Reduces AR-55 req points to 250

Lauwma:
## About The Pull Request

Reduces the AR-55 req point requirement from 525 to 250
## Why It's Good For The Game

AR-55 is a rifle with AR12 stats with worse recoil and 14 less bullets. It loses aim mode if you remove its scope. With a barrel charger it has 0.3 bullet falloff which is not good for a long range weapon. 

Its only pro is that it has a gl54 strapped to it allowing you to shoot airburst shrapnel, fire or tangle. 
These gl54 ammo are limited in vendors and can be bought for a fairly expensive price for the rate at which they are expended.

The HE and razorwire gl54 are no longer available, they were the main problem with that weapon.

I do not believe this is worth 525 req points, not even half ot that, so I'm toning it down to mateba price (A revolver that can stagger a xeno at every shot) 

## Changelog
:cl:
balance: AR-55 req price is now affordable 
/:cl:


Lauwma:
Open to value changes, I put it to 150 given its ammo price already limits how many you can run of these (3 airbust mag + 3 tanglefoot costs over 200)

Lauwma:
> If this gets merged, consider re-adding AR55 to SL vendor

https://github.com/tgstation/TerraGov-Marine-Corps/pull/15662#issue-2244065783

Lauwma:
Though after some more calculations on my excel I thought of a better approach.
I think I'm gonna get gl54 to 200 and ar55 to 250 and introduce gl54 grenades factory

ivanmixo:
Basically the same opinion as the comment on the GL-54 PR, however 525 points *does* seem a bit excessive. Should be in the 300-400 range instead.

Arturlang:
It's not a overpowered gun by any means and it costs way too much to even be considered usable right now

ivanmixo:
It's not, purely because it is essentially 2 guns in one.

Looking4bros:
Two separate guns in one, and neither are particularly easy to bring a lot of ammo for without a box. If you want the most from the GL you'll need the scope, so either just don't drop it or use the belt harness. When i used the GL-54 for a little bit as support, i just bought more incendiary ammunition and carried a shotgun and engaged sparingly with it as to not pack too many shells. 300 points is another 5 Incendiary magazines. :man_shrugging: I rarely use req and don't really use guns that much anymore though.

Naaanii:
This gun is hilariously strong after testing it. I'd make it 500 though because 525 feels too weird.

Lauwma:
In my testing in Valhalla this gun has an higher ttk on most xenos compared to the scout rifle despite being similarly the same price.

In my testing in practice hitting the perfect full damage of the gl54 damage is extremely hard to do, especially offscreen, unless your target is standing still, lowering the effective DPS to something on par with B64. 

For me the strongest thing this weapon currently has is the tanglefoot gas, maybe its ammo price could be toned up a bit, but it's hard to quantify the effectiveness of a small area of tanglefoot because the xeno can just exit it and fight next to it.

Lauwma:
Still convinced it's not worth its price but I'm willing to see how it performs when Leaders issued

Looking4bros:
> For me the strongest thing this weapon currently has is the tanglefoot gas

I've always preferred the incendiary, and using it to harass off screen xenos or igniting escape paths for xenos who just started to push, making them decide to either push through the marines, or escape now over the fire without risking too much damage. With the IFF and Imager function it works great at seeing xenos resting in the distance and setting them on fire to keep them out of the fight for a little longer. Keep in mind, with the right angling it works around corners as well to some degree.

PR #15662, opened 2024-04-15
Adds back ar-55 to leader vendors

Lauwma:
## About The Pull Request

Puts back ar-55 in the leader vendors

## Why It's Good For The Game

ar-55 has been removed from leaders vendors in reaction to razorwire and he gl54 spam. Now that these two ammunitions are gone I think it'd be a good thing to introduce this support weapon to squad leaders and FCs.

"AR-55 is a rifle with AR12 stats with worse recoil and 14 less bullets. It loses aim mode if you remove its scope. With a barrel charger it has 0.3 bullet falloff which is not good for a long range weapon.

Its only pro is that it has a gl54 strapped to it allowing you to shoot airburst shrapnel, fire or tangle.
These gl54 ammo are limited in vendors and can be bought for a fairly expensive price for the rate at which they are expended."

## Changelog
:cl:
balance: Adds the ar-55 back in leaders vendors.
/:cl:


Lauwma:
Closing because conflict bruh

PR #15663, opened 2024-04-15
Changes Gl-54 req price to 200

Lauwma:
## About The Pull Request

Changes gl-54 req price from 300 to 200

## Why It's Good For The Game

Gl-54 is not worth 2 req armor modules or 2 matebas. HE and Razorwire shells are gone.

## Changelog

:cl:
balance: Gl-54 req price is now affordable
/:cl:


Naaanii:
Doesn't it have tfoot mags?

ivanmixo:
Taking it's price down to less than that of an import gun is extremely silly, especially with how stackable these are on aim mode

Lauwma:
My reasoning is that the gun is fairly expensive to shoot while being a mild dps increase from our dedicated sniper rifle SR127. (SR127 ttk on shrike is 7.5 seconds and on GL54 with only perfect its (all pellets go inside the target) it's 6.5 seconds) You can roughly kill one xeno with one full mag of airbust 20 req points. 
Having multiple people spamming this weapon could be a huge point drain. 

Though after some more calculations on my excel I thought of a better approach.
I think I'm gonna get gl54 to 200 and ar55 to 250 and introduce gl54 grenades factory

ivanmixo:
> Having multiple people spamming this weapon could be a huge point drain.

I'm not convinced this means much if it just makes you win. There is next to no counterplay to even one of them being shot at you as a xeno from almost offscreen, let alone multiple. It deals sunder and staggers, which the sniper rifle does not, and this tends to make xenos unable to play the game both short and long term. Overall very unhealthy for the game in my opinion. It also has a bit of NV on top of a strong scope on top of aim mode. 

I completely disagree it should be made more available in any capacity, neither the ammo nor the gun itself. Won't close just yet in case there's going to be a second opinion.

PR #15664, opened 2024-04-15
Makes Bull good again (buffs)

Clay500z:
## About The Pull Request
Baby's first PR. Buffs Bull across the board. Kuro managed to approve the buff in between his Blackstone gooning sessions.

- Gore charge: Now inflicts 3 seconds of Staggerslow (50% less damage and reduced move speed) and 5 seconds of Shatter (armor decreased by a flat 20). Additional PSA: Gore charge does not inflict guaranteed organ damage and never did, its a myth. It only does 2 instant (normal) slashes.

- Headbutt charge: Stun reduced to 2 seconds from 2.5 seconds, and the target is now displaced 1 tile further.

- Plow charge: (Unchanged) Deals 15 damage and displaces the target in a random direction that isn't forward or backwards.

- Primo: Reduced the amount of speed lost from turning to 6, it was previously 8.

## Why It's Good For The Game
Current Bull is complete doodoo water, I don't think anyone will argue with that. It was so bad that the entire tgmc player base somehow Mandela effected themselves into believing that Gore charge does organ damage.

- Gore charge: Only doing 2 slashes of damage and nothing else felt too weak. The shatter allows gore charge to serve as an opener for other Xenos to deal increased damage. The Stagger is there because being able to PB for full damage right after a bull gores your groin off makes no sense, and should increase the survivability of the Bull on successful charges.

- Headbutt charge: Slightly less CBT stun and more braziling potential instead. More fun for both sides.

- Plow charge: Dealing 15 damage is too weak. However, there is currently a bug/feature that lets the bull multi-ram the same person up to three times due to the way mobs bounce off walls, and it's entirely dependent on rng. So technically good but not really because rng, will make a separate PR to address this.

- Primo: Reducing speed by 8 was far too much and rendered the primo as almost completely worthless. For reference, bull starts charging at 5 speed (5 tiles) and reaches max speed at 10. This meant that you'd have to be at max speed when turning to retain a grand total of 2 speed, and the primo does nothing otherwise. Reducing the speed loss to 6 makes the primo actually do something, without enabling the brainrotting back and forth charge spam of the olden days.

## Changelog
:cl:
balance: Gore charge now inflicts stagger and stun
balance: Headbutt charge stun reduced to 2 seconds, and displacement increased by 1 tile
balance: Primordial turning reduces speed by 6 instead of 8
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
+++ b/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
@@ -564,7 +564,7 @@
 		if(CHARGE_CRUSH)
 			Paralyze(CHARGE_SPEED(charge_datum) * 2 SECONDS)
 		if(CHARGE_BULL_HEADBUTT)
-			Paralyze(CHARGE_SPEED(charge_datum) * 2.5 SECONDS)
+			Paralyze(CHARGE_SPEED(charge_datum) * 2 SECONDS)
 
 	if(anchored)
 		charge_datum.do_stop_momentum(FALSE)
@@ -610,7 +610,7 @@
 
 		if(CHARGE_BULL_HEADBUTT)
 			var/fling_dir = charger.a_intent == INTENT_HARM ? charger.dir : REVERSE_DIR(charger.dir)
-			var/fling_dist = min(round(CHARGE_SPEED(charge_datum)) + 1, 3)
+			var/fling_dist = min(round(CHARGE_SPEED(charge_datum)) + 2, 3)
 			var/turf/destination = loc
 			var/turf/temp
 
--- a/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
+++ b/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
@@ -565,6 +565,10 @@
 			Paralyze(CHARGE_SPEED(charge_datum) * 2 SECONDS)
 		if(CHARGE_BULL_HEADBUTT)
 			Paralyze(CHARGE_SPEED(charge_datum) * 2 SECONDS)
+		if(CHARGE_BULL_GORE)
+			target.apply_status_effect(/datum/status_effect/shatter, CHARGE_SPEED(charge_datum) * 5 SECONDS)
+			M.adjust_stagger(CHARGE_SPEED(charge_datum) * 3 SECONDS)
+			M.adjust_slowdown(CHARGE_SPEED(charge_datum) * 3)
 
 	if(anchored)
 		charge_datum.do_stop_momentum(FALSE)
--- a/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
+++ b/code/modules/mob/living/carbon/xenomorph/charge_crush.dm
@@ -96,7 +96,7 @@
 	if(!old_dir || !new_dir || old_dir == new_dir) //Check for null direction from help shuffle signals
 		return
 	if(agile_charge)
-		speed_down(8)
+		speed_down(6)
 		return
 	do_stop_momentum()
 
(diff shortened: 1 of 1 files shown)


Clay500z:
> It would also make gore charge the only one worth using. 2.5 second stun or knockback doesn't compare to liquidating organs and tearing off limbs.

The idea is that gore will be used like 60% of the time as the bread butter simply because it's the one that does the most up front damage. The shatter is just icing for other xenos to lick off.

Meanwhile, stun will be used when teamwork is the plan. 2 second stun + minor Brazil that can setup ganks or further Brazilling seems better than 5 second shatter where the marine can just retreat back into Unga ball.

And plow is for just charging down an Unga line.

We'll see how things play out. If it's overtuned we can reduce the shatter time or just use plain old AP instead. If the headbutt is too weak then we can further reduce stun time and increase displacement.

Clay500z:
>Anyone who gets hit by a bull (which will always be from offscreen) will pretty much immediately get crippled with organ damage and probably a frac on top, and need to go shipside immediately.

Without this buff, you get a frac after getting hit by two gores in a row (Marmor) . With premed, that can go up to thrice in a row. I don't know where you're getting the 'instant organ damage and frac' from. With these changes you'd still probably need to be gored twice in a row to get a frac. 


>Add to that the stagger halving the damage you do to the bull, and the slowdown letting it hit you more easily after the charge

A bull does not simply slash after it's charge without instantly dying. Marines expect to see you after your first charge and are ready to blast you when you show up. The only reason current bull is able to survive at all is because it's immune to stuns. If marines are attentive they'll blast away 3/4 of your health after you charge once, you're a bright red target in the middle of an Unga ball that is hungry to shoot xenos.


I want you to keep in mind that the alternatives to Bull are: 
A xeno designed to get you pregnant and 100% guarantee that you need surgery with no alternatives
A xeno with with two stuns that can go invisible and has base AP on all its slashes
A xeno that can ambush and kill an Marmor marine in 2 seconds flat if used correctly

Clay500z:
>  proposed.gore.buff.vs.heavy.armor.mp4
Aight yeah, just two extra slashes is pretty fast. But I'm still not fully convinced it's going to be a problem in actual gameplay. Bull gets melted fast if it charges someone near an unga ball. Getting in those two slashes could be enough time for an amr to shoot you alongside another marine aimmoding you down. Or even just a PB + aimmode can be enough to kill if you linger around for two slashes. It might have hunters damage potential now, but it doesn't have hunters speed. It can't retreat while zig zagging and dodging bullets, only in a straight line, which matters a lot for surviving doing 2 slashes. 

It can also just... Straight up miss it's charge if the marine just steps 1 tile to the side, so it doesn't seem fair to compare it hunter, which is a lot more difficult to miss with. 

We can TM and test it out, if it's too busted even in practise, I'll change it to do something else ¯\_(ツ)_/¯

SimplySmiley123:
Would definitely go for a TM first, see how it works in your average rounds, several of them even, make the decision after that.

ivanmixo:
Right. This will essentially make gore charge the meta I can just tell you that up front. 

31.5 raw damage against medium armor. 

Why would I ever want to stun the guy when I could just destroy them with gore? The 15 damage the plow charge is not equal damage to the damage gore does to a single marine for below 4 humans, more depending on armor. Xenos can follow up on a shatter far more since it cuts most armor values in almost half than they can a stun.

The primo buff is also guh, I tested values from 3 to 8 when I was originally making the PR and all of them felt like they weren't doing enough because not enough speed was lost. Could be okay though.

In my opinion the debuffs you put on gore here should just be a separate charge type, a pure CC charge would be neater and more balanced.

Clay500z:
So, Ivan informed me that the code didn't actually work the way I thought it worked. I'd thought the gore charge itself would apply Shatter but not benefit from it until the second slash. This was not case, and it does benefit from it. This makes the buff go from "Probably too strong but maybe not," to "Definitely too strong."

I don't think it's worth bothering to TM it now, so I'll be closing this PR and making a new one with a different buff (to avoid confusion from edits).

PR #15672, opened 2024-04-16
Reduces crash xeno starting pop by 1

Barnet2:
## About The Pull Request
What it says on the tin.
## Why It's Good For The Game
3 xenos for 4 marines is uh, a bit too funny. Part of the reason over the months Crash winrates are so skewed.
## Changelog
:cl:
balance: Crash xeno starting pop is 1 instead of 2
/:cl:

### DIFF
--- a/code/datums/gamemodes/crash.dm
+++ b/code/datums/gamemodes/crash.dm
@@ -12,7 +12,7 @@
 		/datum/job/terragov/medical/professor = 1,
 		/datum/job/terragov/silicon/synthetic = 1,
 		/datum/job/terragov/command/fieldcommander = 1,
-		/datum/job/xenomorph = FREE_XENO_AT_START
+		/datum/job/xenomorph = 1
 	)
 	job_points_needed_by_job_type = list(
 		/datum/job/terragov/squad/smartgunner = 20,


TiviPlus (approved):

Naaanii:
Salt PR?

Aidenzack1:
This won’t fix the actual issue crash has right now, it’s not an issue of how many they start with, it’s the fact they come in huge waves.

this just pushes the ball further down, lets marines win for a few more minutes until a spawn wave of four to five xenos happens at once. I’d suggest evening out the spawn waves since getting four xenos all at once is well, very powerful, make it one or two at a time max in shorter intervals

hivemuscle:
image

0 burrowed. Are you sure you want to make it 1vs8+2 shipside?

Barnet2:
> image
> 
> 0 burrowed. Are you sure you want to make it 1vs8+2 shipside?

That is the balance_scales() proc being what it is, which should honestly be touched in another PR. Doing the math that is a bit over 30 larva points in total (if no xenos had died), so it would be 5 versus 9 (including FC, not including the CMO which also gives larva points).

hivemuscle:
> > image
> > 0 burrowed. Are you sure you want to make it 1vs8+2 shipside?
> 
> That is the balance_scales() proc being what it is, which should honestly be touched in another PR. Doing the math that is a bit over 30 larva points in total (if no xenos had died), so it would be 5 versus 9 (including FC, not including the CMO which also gives larva points).

I think it should be touched in the same PR, because lowering xeno start numbers but not doing anything about balance_scales() will not fix, but fuck up balance even more to the marine side. Moreover, the crash gamemode have more significant problems (for example, it is too hard to get major as xeno because marines can turn your win into a minor by a press of a single button) that continue to be ignored.

Lumipharon:
I don't think this is right. At very low pop xenos can only have 2 dudes to start with. If there's issues with scaling, then scaling needs to be adjusted, not this.

PR #15674, opened 2024-04-17
Cleaned up fire sprites

Lumipharon:
## About The Pull Request
Cleaned up fire sprites as they were a fucking mess of 5000 colours.
They are now standardised so its easy to recolour them (such as for pyrogen), although standard fire (and ONLY for the max flame sprite) had 5 colours instead of 4, and I kept it that way as 5 colours didn't look as nice for green/blue since they are transitioning from white to blue/green instead of yellow to red.

image

image

image

The better colours are also much more noticable/better when you see the lighting change around the fire (as fire's own lighting effects its appearance)

image

Tye's pr reminded me to do this, as I don't personally find pure particles the best look, but current sprites were really jank. Blue/green fire in particulary is really shitty currently, pic for reference.

image

## Why It's Good For The Game
Better sprites.
## Changelog
:cl:
imageadd: Updated fire sprites
/:cl:


Gregormack:
> To a degree, won't this slightly hurt readability?

Quite the opposite

Wirpy:
> To a degree, won't this slightly hurt readability?

no, old f ire had like 300+ colours. Its like someone changed colour everytime they painted a new pixel.

Wirpy (approved):

PR #15676, opened 2024-04-18
Stacks of newly crafted items merge to existing stacks in user hands

antropod:
## About The Pull Request

Adds QoL feature of merging stacks of newly crafted items to existing stacks that user holds in their hands.
Makes life of an engineer a bit easier.

## Changelog
:cl:
add: Stacks of newly crafted items merge to existing stacks in user hands
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -235,6 +235,10 @@
 		O.color = color
 	use(R.req_amount * multiplier)
 
+	if(isitemstack(O))
+		var/obj/item/stack/stack = O
+		stack.merge_with_stack_in_hands(user)
+
 	if(QDELETED(O))
 		return //It's a stack and has already been merged
 
@@ -398,6 +402,21 @@
 /// Proc for special actions and radial menus on subtypes. Returning FALSE cancels the recipe menu for a stack.
 /obj/item/stack/proc/select_radial(mob/user)
 	return TRUE
+
+/**
+	Merges stack into the one that user is currently holding in their left or right hand.
+	Returns TRUE if the stack was merged, FALSE otherwise.
+*/
+/obj/item/stack/proc/merge_with_stack_in_hands(mob/user)
+	var/obj/item/stack/stack_in_hands = null
+	if(istype(user.l_hand, merge_type))
+		stack_in_hands = user.l_hand
+	else if(istype(user.r_hand, merge_type))
+		stack_in_hands = user.r_hand
+	if(stack_in_hands && merge(stack_in_hands))
+		return TRUE
+	return FALSE
+
 
 #undef STACK_CHECK_CARDINALS
 #undef STACK_CHECK_ADJACENT


code/modules/mob/inventory.dm
```diff
@@ -128,15 +128,25 @@
...
-/mob/proc/put_in_hands(obj/item/W, del_on_fail = FALSE)
+/mob/proc/put_in_hands(obj/item/W, del_on_fail = FALSE, merge_stacks = TRUE)
 	W.do_pickup_animation(src)
 	if(status_flags & INCORPOREAL) // INCORPOREAL things don't have hands
 		return FALSE
 	if(!W)
 		return FALSE
+	if(merge_stacks && istype(W, /obj/item/stack))
+		var/obj/item/stack/S = W
+		var/obj/item/stack/D = null
```
QualityVan, line 143: Don't use single letter var names like this

code/modules/mob/inventory.dm
```diff
@@ -128,15 +128,25 @@
...
+		if(istype(src.l_hand, S.merge_type))
```
QualityVan, line 144: No need for src., just use the var directly

QualityVan (commented):
The pr intent is good, but I'm shaky on the implementation. Putting stack-specific code in mob put in hands seems not great.

code/modules/mob/inventory.dm
antropod: I thought this is the code style because I saw that everywhere, I'll fix it

antropod:
I fixed style issues on what you pointed at.
On the matter of putting stack specific code into mob... tgstation does that, not that their code is great, just sayin. I'm not very familiar with codebase, so I did what I knew would work.

code/modules/mob/inventory.dm
```diff
@@ -128,15 +128,25 @@
...
 		return FALSE
+	if(merge_stacks && istype(W, /obj/item/stack))
+		var/obj/item/stack/new_stack = W
+		var/obj/item/stack/stack_in_hands = null
+		if(istype(l_hand, new_stack.merge_type))
+			stack_in_hands = l_hand
+		else if(istype(r_hand, new_stack.merge_type))
+			stack_in_hands = r_hand
+		if(stack_in_hands && new_stack.merge(stack_in_hands))
+			return TRUE
```
Lumipharon, line 149: this is very smelly. This should be an item proc, not a mob proc.

Lumipharon (changes requested):

code/modules/mob/inventory.dm
antropod: `user.put_in_hands(stack)`, one could argue it should be a mob proc because it involves mob and mob hands. and what if mob does not have hands? then you check type of a mob like you check type of a stack here.
I don't want to make it super complicated and change a bunch of code just for such simple thing as merging stacks.
KISS

code/modules/mob/inventory.dm
```diff
@@ -128,15 +128,25 @@
...
 	Returns TRUE if it was able to put the thing into one of our hands.
 */
-/mob/proc/put_in_hands(obj/item/W, del_on_fail = FALSE)
+/mob/proc/put_in_hands(obj/item/W, del_on_fail = FALSE, merge_stacks = TRUE)
 	W.do_pickup_animation(src)
 	if(status_flags & INCORPOREAL) // INCORPOREAL things don't have hands
 		return FALSE
 	if(!W)
 		return FALSE
+	if(merge_stacks && istype(W, /obj/item/stack))
```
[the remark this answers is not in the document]
antropod: There are few reasons why I put it here:
1. Mob already has a proc called put_in_hands which selects where to put the item.
2. You have to check if stacks can merge before failing and putting item on the ground.
3. In the proc that handles interaction between mob and item you will always have to typecheck one or the other.
4. I don't want to make new procs just for this simple change. More procs = less readable code.

More about "snowflake" and "smelly" code. Those definitions are subjective, so I don't understand what you mean. You looked at the code and it felt "not right". I don't understand what is wrong with it because I do not have your experience.

I don't like where this conversation is going, I think I spent more energy trying to explain the change than actually making the change. And English is not my first language.
I wanted a simple fix to make people life better, and you instead of helping me do the fix, just point and say "that's wrong", "this is not how you do that". I know. You would have done it different;y. If you don't like it, close the PR.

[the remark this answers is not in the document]
Lumipharon: You've had 3 maints tell you the same thing, we are helping you fix it, we're telling you where to move the code.

Its snowflake because it is stack item specific code, being handled entirely by a mob level proc which is supposed to be very generic. You generally want things handled as internally as possible, its more maintainable, its cleaner, and easier to read.

What happens if we have another pick up interaction with another item type? Do we add more snowflake for that type? What about a 3rd? a 10th? What happens if stack item code changes in the future, or something else shares behavior but isn't the same type?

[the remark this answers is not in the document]
BraveMole: "I don't like where this conversation is going, I think I spent more energy trying to explain the change than actually making the change."

I get your feeling, your code is working. But you said it yourself, you are not familiar with the code base. We are pointing out that here, we want out mob code to be extra generic. Which means the logic for stacking item should be on /obj/item/stack. 

And yes, "snowflake" is very subjective, but this looks to be a consensus among maints.

"I wanted a simple fix to make people life better, and you instead of helping me do the fix, just point and say "that's wrong", "this is not how you do that". I know. You would have done it different;y. If you don't like it, close the PR." 
I think we are trying to help here, by pointing out where we think the code should be. 
Your feature seems interesting, and I would really like not to close this PR; but you also gotta play the open source game, which is to make the maintainer like your code.

Please try to put your logic in stack, ask for help in discord if needs be, and this will be merged with no problems

code/modules/mob/inventory.dm
```diff
@@ -128,6 +128,7 @@
 	Arguments
 	* obj/item/W is the item you are trying to equip
 	* del_on_fail if true will delete the item instead of dropping it to the floor
+	* merge_stacks if true will merge stacks that src is holding in its hands with new ones
```
QualityVan, line 131:
```suggestion
```
Leftover comment from the first version

code/game/objects/items/stacks/stack.dm
```diff
@@ -235,6 +235,11 @@
 		O.color = color
 	use(R.req_amount * multiplier)
 
+	if(isitemstack(O))
+		var/obj/item/stack/stack = O
+		if(stack.merge_with_stack_in_hands(user))
+			return // Stack is merged and original item is deleted
```
QualityVan, line 241: It's a niche case and probably fine anyway since none of the lower code looks relevant, but what about when a partial merge occurs due to the in hand stack being near cap? Say you're holding 45 and create 10 of something with a max 50 stack. Probably matters more for the comment assuming deletion than functionality, honestly.

QualityVan (changes requested):

code/game/objects/items/stacks/stack.dm
antropod: Good point, I'm gonna test it

antropod: Tested with stacks of razor wire assembly (those craft 2 at a time). Seems like nothing breaks, but I changed code and removed early return

QualityVan (approved):
Don't mark comments as resolved yourself, maintainers use those to double check that stuff was properly addressed while reviewing and resolve it then.

PR #15679, opened 2024-04-18
Squad order walance changes

breadhunt:
## About The Pull Request
Makes SL orders **only apply to fellow squad members**
In exchange SL orders have been increased from 3-->4 (same as FC orders)

Can atomise if wanted but I think they're related enough

## Why It's Good For The Game

Right now it's optimal to stick in a single ball and suboptimal to split up which gets kind of stale after a while

Intention is to incentivise using squad systems more (alongside other incoming changes), especially if you're planning on playing together so you can buff each other up

Removal of order barks is just a side thing that nobody I've talked to cares for, clutters comms & confuses new players
## Changelog
:cl:
balance: SL leadership skill 3 --> 4 (same as FC)
balance: SL orders now only affect their squad
del: Removed order barks
/:cl:


coldironwarrior:
not sold on this.
squads aren't always cohesive and sometimes the group of people doing things together are not all part of the same squad. in numerous cases, i've attached myself to a squad on the ground doing things that seemed fun because i didn't have an actual SL and didn't know this until after latejoining and confirming my original join squad. this isn't a dealbreaker since it's just affecting the aura but it sucks to be the least-performant of a team for reasons outside my control because i joined late and no one's in CIC to transfer me

i'm very against the SL leadership buff for orders, though. the only two roles with MASTER level leadership are the captain and the FC, who should be The Best leaders; above squad leaders. making SLs MASTER level leaders circumvents the unique contributions an FC specifically can have, and also takes them out of parallel with **their superior officers** by making SOs weaker than SLs

Helg2:
It's not orders without the actual orders, reflavor it to aura giving nanomachines or something then.

XElectricX:
Good PR but why are you removing flavor from the game? Keep the quotes.

Coastfront:
A wind laboring a beach apart is doomed to see it come to one again; The waves make it so.

By nature of how the game works, it will, in every situation for marines, be explicitly suboptimal to split up more than being within a walks reach from eachother.

As has been mentioned; The weakest squad can be constantly forced to bear the brunt of enemy contact due to how much map freedom xenos have.
As-well; This wouldn't necessarily "Encourage" squads to split up. It would encourage squad leaders to all stay together at all times; And would then simply become a stark buff for the marines.
The only theoretically-feasible way to encourage "4 distinct squads who all split up" is either via outright banning the unga bad (Fucking horrible decision) or adding safeguards against squads simply getting jumped (Which would still only be slight encouragement.)
Unless if you outright granted SL's negative synergy via having an SL cause all of its non-squadmates to become automatically depressed & lose their leadership buffs if within range of an opposing SL; This change simply causes all SL's to sit in the ball, so that everyone has FC level leader-ship whether or not the FC is dead.

This is either a buff having skinned a rebalance; Or it is a foolish (If earnest) attempt to diversify gameplay.
If it is the former; Poor form
If it is the ladder; Decent attempt, poor execution.

SimplySmiley123:
As said above, this won't encourage marines to split up in squads _at all_ mostly because it doesn't do anything to the main factor as to _why_ marines stay together: xenomorphs can and will gang on a smaller squad to wipe them out without much of an effort 
thanks to how xenomorphs are balanced around one large unga ball instead of smaller groups, and go back to the larger/stronger group(s) of marines with a numbers advantage.

Hell I would argue that this will just encourage clumping together even more.

breadhunt:
> not sold on this. squads aren't always cohesive and sometimes the group of people doing things together are not all part of the same squad. in numerous cases, i've attached myself to a squad on the ground doing things that seemed fun because i didn't have an actual SL and didn't know this until after latejoining and confirming my original join squad. this isn't a dealbreaker since it's just affecting the aura but it sucks to be the least-performant of a team for reasons outside my control because i joined late and no one's in CIC to transfer me because my old SL Fragger Jones died/cryo'd
> 
squad tablet exists, I will probably also be adding some kind of option in headset to make it easier, i don't envision this as a problem but I might be missing something
attaching to squads is cool but whenever I SL it's a constant battle trying to keep non-squad members out -- if someone wants to come with us, I usually ask them to join us because it's easier to follow a single blob of purple rather than multicolour vomit

> i'm very against the SL leadership buff for orders, though. the only two roles with MASTER level leadership are the captain and the FC, who should be The Best leaders; above squad leaders. making SLs MASTER level leaders circumvents the unique contributions an FC specifically can have, and also takes them out of parallel with **their superior officers** and shipside counterparts by making SOs weaker than SLs

Cap/FC are still the best leaders because their orders literally apply to everyone. The intention of this PR is to focus SL into a more specialised role of "leading their squad" rather than the current "FC-" that they are
point about SO makes sense though, unsure what to do about that. I still want to give SL something in exchange for nerfing orders however, any ideas?

> Good PR but why are you removing flavor from the game? Keep the quotes.
1) confusing and obtuse to new players
2) clogs up comms
It's been a gripe for me for a while now, if you have any ways of phrasing it that don't give new marines completely conflicting signals (HOLD TIHS AREA HOLD THIS AREA HOLD THI-MOVE MOVE MOVE! NO HOLD YOU FUCKS) i'd be all for keeping it though

> Won't be able to fix this without majorly reworking the game. Marines can't split into multiple groups because the entire xeno team can relocate to whichever group has fewer/weaker marines, wipe them with the team size disparity, then move back to the other marine group(s) in less time than it takes for two cycles of a nuke disk to complete.

this is the plan
I am open to closing the PR until this is done however

> As a result of the game not facilitating multiple squads, you're always going to be mixed in with with multiple other squads in a given area. Orders will apply inconsistently to the marines around you with this PR.

plan is for FC to have this kinda role

> Maybe this could work better if we reduced the number of squads to 2 (people can still make custom squads ingame).

I will probably do this tonight

Lumipharon:
I like the idea of this pr in principal, but as other people have said, there are various reasons why I don't think this would work in isolation.

SimplySmiley123:
> squad tablet exists, I will probably also be adding some kind of option in headset to make it easier, i don't envision this as a problem but I might be missing something attaching to squads is cool but whenever I SL it's a constant battle trying to keep non-squad members out -- if someone wants to come with us, I usually ask them to join us because it's easier to follow a single blob of purple rather than multicolour vomit

Why does the colour matter though, what other marines care about, only thing they care about, is what role someone is when they need that role for something, be it a corpsman to heal them or revive someone's corpse, or an engineer for setting up cades, repairing an APC and whatnot.

Hell at this rate, why _have_ squads at all beyond custom squad gimmicks?

QualityVan:
Agree that this doesn't really accomplish its goal of encouraging splitting. If anything, won't it just encourage all marines to stay in the same squad?

Also agree that order chat spam is annoying. What about just removing the radio chatter from it and leaving the local message? Should be as easy as cutting the ;s from the message list. Only issue with that offhand is overwatch cams.

breadhunt:
> Why does the colour matter though, what other marines care about, only thing they care about, is what role someone is when they need that role for something, be it a corpsman to heal them or revive someone's corpse, or an engineer for setting up cades, repairing an APC and whatnot.

Don't ask me I've just noticed people follow _significantly_ more when it's a singular colour. Something something rule of cool, I guess

> Hell at this rate, why _have_ squads at all beyond custom squad gimmicks?

Main reasons why it's worth the effort:
1. Teamwork is cool and one of the main selling points we have; if you want to play the best top-down shooter out there you probably aren't playing TGMC
2. Squads = tactics which help make the game less stale. It's near impossible to do tactics without separating people into squads and downright impossible to do reliably, trust me we've been trying, it's really fucking difficult.
3. Having teammates you can actually get to know makes bad teammates less frustrating (in theory) because at least you know them. "fuck this game it's awful" vs "Fuck Jerry he's an asshole" kinda thing.

I do completely agree otherwise though, I'm gonna close this PR fornow, when the objective changes are fully cooked I'll reopen & see what people think afterwards

PR #15680, opened 2024-04-18
Revert "gives job vendors tag sets for item categories; reorganizes so alike items are grouped together"

coldironwarrior:
Reverts tgstation/TerraGov-Marine-Corps#15459

overrides the point cost showing now and idk how to fix it


Lumipharon:
Fixed in #15689

PR #8865, opened 2021-11-30
Multiz port

hyper2snyper:
Draft
## Changelog
:cl:
add: Added new mechanics or gameplay changes
add: Added more mechanics or gameplay changes
expansion: Expands content of an existing feature
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:

PR #15681, opened 2024-04-19
TGMC Multi-Z

RipGrayson:
## About The Pull Request

This is a very basic foundation for multi-z, using https://github.com/tgstation/TerraGov-Marine-Corps/pull/8865 as a base. I've applied it to Sulaco for the sake of play testing. Since Sulaco was already using a fake implementation of multi-z, there shouldn't be any differences balance-wise (except for potential uncaught bugs with multi-z itself).
I've also added a map for debugging multi-z, ported from TG itself.

## Why It's Good For The Game

There's a lot of cool things and elements we can do with multi-z, but we have no support for it at the moment. My hope is that with this PR, we can work through implementation issues and then do work on actually making it fit the TGMC gameflow. In the future we can do everything from underground sewers, to proper ventilation systems for xenos to crawl around in, to HvH sniper perches, but we need to have a base implementation ready and bug free before we can add systems for any of those things.

## Changelog
:cl:
add: Added basic support for multi-z.
add: The Sulaco now uses true multi-z.
/:cl:


TiviPlus:
As discussed in discord, while this is an ok idea to test, we should port the up to date version for the sake of maintainability + parity

PR #15708, opened 2024-04-24
Adds silo scaling for xenos when behind

blackdav123:
## About The Pull Request
Currently xenos get a 0.7x up to 1x multiplier depending on the current ratio of xenos to marines (with about 2:5 being the ideal ratio), lowering silo gen if they are above this ratio. This PR raises the cap on this from 1x to 1.3x allowing xenos to get increased larva gen for holding out while outnumbered.
## Why It's Good For The Game
Many rounds especially in lowpop nuke war end up with xenos stuck at about a 1:5 ratio because they can do very little but sit back and watch as marines do disks for an hour while the silo is generating a larva every 17 minutes. 
My goals with this are as follows:

1. Create fewer slow rounds where marines have hardly anything to shoot at.
2. Encourage marines to take more risks on lowpop instead of slowpushing with a 5 to 1 ratio.
3. Give losing xeno teams more chances to get back into the game.

I cant say for certain if this change will do these things (and whether this will be a good or bad thing for highpop) but I do think its worth a testmerge to trial it.
## Changelog
:cl:
balance: Xenos now recieve up to 1.3x larva gen while down in numbers.
/:cl:

### DIFF
--- a/code/controllers/subsystem/silo.dm
+++ b/code/controllers/subsystem/silo.dm
@@ -31,7 +31,7 @@ SUBSYSTEM_DEF(silo)
 	//We scale the rate based on the current ratio of humans to xenos
 	var/current_human_to_xeno_ratio = active_humans / active_xenos
 	var/optimal_human_to_xeno_ratio = xeno_job.job_points_needed / LARVA_POINTS_REGULAR
-	current_larva_spawn_rate *= clamp(current_human_to_xeno_ratio / optimal_human_to_xeno_ratio , 0.7, 1)
+	current_larva_spawn_rate *= clamp(current_human_to_xeno_ratio / optimal_human_to_xeno_ratio , 0.7, 1.3)
 
 	GLOB.round_statistics.larva_from_silo += current_larva_spawn_rate / xeno_job.job_points_needed
 


Barnet2:
Don't know how the numbers will play out but I definitely like the idea.

Lauwma:
My issue with this is that it encourages xenos to delay and drag a marine win longer than it should be. Marines don't get more things when they are behind, they just lose straight up

blackdav123:
> My issue with this is that it encourages xenos to delay and drag a marine win longer than it should be. Marines don't get more things when they are behind, they just lose straight up

I agree with this somewhat but lately we have had a lot of rounds where xenos are putting in 100% of the effort while down to 3 or 4 xenos but they just still cant have any chance to win because marines are sitting behind cades playing for maximum safety. To me these rounds are being delayed but its not at any fault of the xenos, just that we arent giving them any chance to win.

I think both options for marines (playing safe and silo rushing) should come with risks so they have to be more proactive with which they are choosing. If marines fully commit to playing safe then xenos will have a chance to scale back into the game, while if marines choose to silo rush then they risk losing people and letting the xenos back into the game.

blackdav123:
dont think we need this see https://github.com/tgstation/TerraGov-Marine-Corps/pull/15740 for why

